PR #656, opened 2024-02-07
DuckDB DataBackend

phillipleblanc:
Adds a DuckDB backend that supports either a duckdb file or an in-memory DuckDB.

File-based DuckDB
```yaml
from: spice.ai/eth.recent_blocks
name: eth_recent_blocks
acceleration:
  refresh_mode: full
  mode: file
  engine: duckdb
```

Memory DuckDB
```yaml
from: spice.ai/eth.recent_blocks
name: eth_recent_blocks
acceleration:
  refresh_mode: full
  mode: memory
  engine: duckdb
```

Use a specific DuckDB filename (defaults to `{dataset.name}.db`)
```yaml
from: spice.ai/eth.recent_blocks
name: eth_recent_blocks
params:
  duckdb_file: foo.db
acceleration:
  refresh_mode: full
  mode: file
  engine: duckdb
```


crates/runtime/src/databackend.rs
```diff
@@ -1,47 +1,56 @@
...
+            Engine::Arrow => match mode {
+                Mode::Memory => Ok(Box::new(MemTableBackend::new(Arc::clone(ctx), name))),
+                Mode::File => InvalidConfigurationSnafu {
+                    msg: "File mode not supported for Arrow engine".to_string(),
+                }
+                .fail()?,
+            },
+            #[cfg(feature = "duckdb")]
+            Engine::DuckDB => Ok(Box::new(
+                DuckDBBackend::new(Arc::clone(ctx), name, mode.into(), params)
```
mitchdevenport, line 50: Do we want to support memory and file based DB for DuckDB too?

mitchdevenport: Oh I see we're doing that within DuckDB

crates/runtime/src/databackend.rs
```diff
@@ -1,47 +1,56 @@
...
+    ) -> std::result::Result<Box<Self>, Error> {
+        match engine {
+            Engine::Arrow => match mode {
+                Mode::Memory => Ok(Box::new(MemTableBackend::new(Arc::clone(ctx), name))),
+                Mode::File => InvalidConfigurationSnafu {
+                    msg: "File mode not supported for Arrow engine".to_string(),
+                }
+                .fail()?,
+            },
+            #[cfg(feature = "duckdb")]
```
mitchdevenport, line 48: Why does this need to be behind a feature?

crates/runtime/src/databackend/duckdb.rs
```diff
@@ -1,5 +1,274 @@
...
+    duckdb_conn: r2d2::PooledConnection<DuckdbConnectionManager>,
+}
+
+impl DuckDBUpdate {
+    fn update(&mut self) -> Result<()> {
+        match self.update_type {
+            UpdateType::Overwrite => self.create_table()?,
+            UpdateType::Append => {
+                if !self.table_exists() {
+                    self.create_table()?;
```
mitchdevenport: There's a race condition here if two append updates come in simultaneously before the table is created then they will both see that the table doesn't exist and will both attempt to create the table. One will fail and we'll lose that data.

crates/runtime/src/databackend.rs
mitchdevenport: Answered in the learn and grow :)

crates/runtime/src/databackend/duckdb.rs
phillipleblanc: Fixed

PR #657, opened 2024-02-08
Basic OpenTelemetry Server

mitchdevenport:
Adds an implementation of OpenTelemetry MetricsService: 

https://github.com/open-telemetry/opentelemetry-proto/blob/4ca4f0335c63cda7ab31ea7ed70d6553aee14dce/opentelemetry/proto/collector/metrics/v1/metrics_service.proto

For now, just logs the metrics. A follow up PR will add logic to write the metrics to datafusion tables.

### DIFF
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -7,6 +7,7 @@ use app::App;
 use clap::Parser;
 use runtime::config::Config as RuntimeConfig;
 use runtime::datasource::DataSource;
+
 use runtime::{databackend, datasource, Runtime};
 use snafu::prelude::*;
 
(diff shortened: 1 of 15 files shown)


bin/spiced/src/lib.rs
```diff
@@ -111,20 +114,26 @@ pub async fn run(args: Args) -> Result<()> {
...
-                    .context(UnableToAttachDataSourceSnafu {
-                        data_source: source,
-                    })?;
+                unsafe {
+                    MUT_DATA_FUSION
+                        .attach(ds, data_source, databackend::DataBackendType::default())
+                        .context(UnableToAttachDataSourceSnafu {
+                            data_source: source,
+                        })
+                }?;
```
phillipleblanc: Lets find a way to do this that doesn't involve unsafe code

phillipleblanc (approved):

PR #662, opened 2024-02-08
Write metrics received from the OpenTelemetry endpoint to databackends

mitchdevenport:
This PR finishes implementing the OpenTelemetry API to convert received metrics to arrow and write them to datafusion tables.

There is a limitation currently that columns are always written in the order that attributes are received from OpenTelemetry. This order is no consistent so column values are written out of order to datafusion. This PR is already extremely dense so I will fix this in a follow up PR as it is non-trivial.

### DIFF
--- a/crates/runtime/src/opentelemetry.rs
+++ b/crates/runtime/src/opentelemetry.rs
@@ -233,6 +233,31 @@ fn number_data_points_to_record_batch(
     }
 }
 
+macro_rules! handle_attribute {
+    ($columns:expr, $fields:expr, $key:expr, $value:expr, $builder_type:ty, $data_type:expr, $metric:expr) => {{
+        let key_str = $key.as_str();
+        match $columns.get_mut(key_str) {
+            None => {
+                $fields.insert($key.clone(), Field::new(key_str, $data_type, true));
+                let mut builder = <$builder_type>::new();
+                builder.append_value($value);
+                $columns.insert($key.clone(), Box::new(builder));
+            }
+            Some(column) => {
+                if let Some(builder) = column.as_any_mut().downcast_mut::<$builder_type>() {
+                    builder.append_value($value);
+                } else {
+                    tracing::error!(
+                        "Attribute defined multiple times with multiple types: {}.{}",
+                        $metric,
+                        key_str
+                    );
+                }
+            }
+        };
+    }};
+}
+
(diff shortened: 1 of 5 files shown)


crates/runtime/src/opentelemetry.rs
```diff
@@ -69,8 +132,231 @@ impl MetricsService for Service {
...
+    let mut fields: IndexMap<String, Field> = IndexMap::new();
+    let mut columns: IndexMap<String, Box<dyn ArrayBuilder>> = IndexMap::new();
+    for inner_attributes in attributes {
+        for attribute in inner_attributes {
+            let key_str = attribute.key.as_str();
+            match &attribute.value {
+                Some(any_value) => match &any_value.value {
+                    Some(value) => match value {
+                        any_value::Value::StringValue(string_value) => {
+                            match columns.get_mut(key_str) {
```
mitchdevenport: Unfortunately these branches can't be generalized because `new` and `append` are not defined by a trait, they are defined by each builder individually.

mitchdevenport: Per @y-f-u I'm going to look into using macros for this

crates/runtime/src/opentelemetry.rs
```diff
@@ -69,8 +135,279 @@ impl MetricsService for Service {
...
+            .map(|(_, mut v)| v.finish()),
+    );
+
+    match RecordBatch::try_new(Arc::new(Schema::new(fields)), columns) {
+        Ok(record_batch) => Ok(record_batch),
+        Err(e) => Err(e).context(FailedToBuildRecordBatchSnafu),
+    }
+}
+
+macro_rules! insert_attribute {
```
y-f-u: took the crown of first ever macro_rules in spice oss I believe

crates/runtime/src/opentelemetry.rs
```diff
@@ -69,8 +135,255 @@ impl MetricsService for Service {
...
+                        $metric,
+                        key_str
+                    );
+                }
+            }
+        };
+    }};
+}
+
+#[allow(clippy::too_many_lines)]
```
phillipleblanc:
```suggestion
```

Is this still too many lines?

mitchdevenport: nope!

phillipleblanc (approved):

PR #664, opened 2024-02-09
Match incoming OpenTelemetry metrics to existing schema

mitchdevenport:
This PR updates the OpenTelemetry metrics API to match incoming metrics to an existing table schema if present. Specifically:

1. Match order of attribute columns in arrow Records to existing table schema.
2. If an attribute was previously provided on the metric but is not provided on the current metric, null will be inserted in the new row for that attribute's column.
3. Matches value types across metrics of the same name and will error if they do not match.
4. Insert a null value if a value was previously provided but is not provided on the current metric.

This PR also establishes the rule that the very first metric ever received sets the initial schema. So, if the first metric is missing a value then we simply skip that metric because it is not valid for establishing schema.

While the OpenTelemetry endpoint handles arbitrary new columns, none of our databackends do. They all error on insert of data with new columns, so we will need to handle this case for all databackends to support via all ingestion methods.


crates/runtime/src/databackend/duckdb.rs
```diff
@@ -20,6 +20,7 @@ use super::{AddDataResult, DataBackend};
 
 #[derive(Debug, Snafu)]
 pub enum Error {
+    #[snafu(display("DuckDBError: {source}"))]
```
phillipleblanc, line 23: Is this redundant?

crates/runtime/src/opentelemetry.rs
```diff
@@ -254,108 +353,170 @@ macro_rules! append_attribute {
...
 }
 
+fn initialize_attribute_schema(
+    fields: &mut IndexMap<String, Arc<Field>>,
+    columns: &mut IndexMap<String, Box<dyn ArrayBuilder>>,
+    existing_schema: &Option<Schema>,
+) {
+    if let Some(s) = existing_schema {
+        for field in s.fields() {
+            // Skip value and time fields sbecause they are not attributes and are already handled.
```
phillipleblanc:
```suggestion
            // Skip value and time fields because they are not attributes and are already handled.
```

crates/runtime/src/opentelemetry.rs
```diff
@@ -254,108 +353,170 @@ macro_rules! append_attribute {
...
+            if field.name() == VALUE_COLUMN_NAME
+                || field.name() == TIME_UNIX_NANO_COLUMN_NAME
```
phillipleblanc, line 491: Is this column part of the OTel spec? Or something that telegraf is sending us?

phillipleblanc (approved):

crates/runtime/src/opentelemetry.rs
mitchdevenport: This is part of the OTel spec: https://github.com/open-telemetry/opentelemetry-proto/blob/4ca4f0335c63cda7ab31ea7ed70d6553aee14dce/opentelemetry/proto/metrics/v1/metrics.proto#L335

crates/runtime/src/databackend/duckdb.rs
mitchdevenport: No, I discovered that if a display message is not specified with the underlying error then it is not included in the message by default. So we should actually start doing this throughout our code.

PR #667, opened 2024-02-12
Load models during instantiation; make /inference 100x faster

Jeadie:
## Changes
- For a basic NBeats model (the one we showed during demos), we had ~600-700ms latency, now 6-7ms!

### DIFF
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -8,9 +8,9 @@ use clap::Parser;
 use runtime::config::Config as RuntimeConfig;
 use runtime::datasource::DataSource;
 
-use runtime::modelruntime::ModelRuntime;
-use runtime::modelsource::ModelSource;
-use runtime::{databackend, datasource, modelruntime, modelsource, Runtime};
+
+
+use runtime::{databackend, datasource, Runtime};
 use snafu::prelude::*;
 
 #[derive(Debug, Snafu)]
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -98,13 +98,17 @@ pub async fn run(args: Args) -> Result<()> {
                 let path = path.unwrap().clone();
 
                 let model = modelruntime::tract::Tract {
-                    path: path.to_string()
+                    path: path.to_string(),
                 }
                 .load();
-
-                if model.is_ok() {
-                    let result = model.unwrap().run();
-                    tracing::info!("Model loaded: {:?}", result);
+                match model {
+                    Ok(m) => {
+                        let result = m.run();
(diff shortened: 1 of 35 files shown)


bin/spiced/src/lib.rs
```diff
@@ -144,7 +146,23 @@ pub async fn run(args: Args) -> Result<()> {
...
+    let mut model_map = HashMap::with_capacity(app.models.len());
+    for m in &app.models {
+        match Model::load(m) {
+            Ok(in_m) => {
+                model_map.insert(m.name.clone(), in_m);
+            }
+            Err(e) => {
+                tracing::warn!(
+                    "Unable to load runnable model from spicepod {}, error: {}",
+                    m.name.clone(),
```
phillipleblanc, line 158: is this clone necessary?

bin/spiced/src/lib.rs
```diff
@@ -144,7 +146,23 @@ pub async fn run(args: Args) -> Result<()> {
...
+                    e
```
phillipleblanc, line 159: You can insert the error directly in the string, this should give a lint error.

crates/runtime/src/http/v1.rs
```diff
@@ -57,20 +58,21 @@ pub(crate) mod inference {
...
             return Json(InferenceResponse {
                 forecast: Vec::new(),
                 duration_ms: start_time.elapsed().as_millis(),
             });
         }
 
-        let runnable = Model::load(&(model.unwrap())).unwrap();
-        let result: Vec<f32> = runnable
+        let result: Vec<f32> = model
+            .unwrap()
```
phillipleblanc, line 75: This is a lint error

crates/runtime/src/lib.rs
```diff
@@ -57,6 +65,7 @@ impl Runtime {
             self.config.http_bind_address,
             self.app.clone(),
             self.df.clone(),
+            self.models.clone(),
```
phillipleblanc: We'll need to support dynamic updates (the item Sergei is working on) - which cloning doesn't support. We'll need to rethink how to load data in here. Ok for this PR

bin/spiced/src/lib.rs
```diff
@@ -144,7 +146,23 @@ pub async fn run(args: Args) -> Result<()> {
         tracing::info!("Loaded dataset: {}", ds.name);
     }
 
-    let rt: Runtime = Runtime::new(args.runtime, app, df);
+    let mut model_map = HashMap::with_capacity(app.models.len());
```
phillipleblanc, line 149: This model_map should be encapsulated in the `App`, not a separate thing

PR #671, opened 2024-02-13
Fix ONNX inferencing for multivariate inputs.

Jeadie:
## Changes
- Core change is the ` transpose_and_flatten` being applied to the input vector.

### DIFF
--- a/crates/runtime/src/modelruntime/tract.rs
+++ b/crates/runtime/src/modelruntime/tract.rs
@@ -33,7 +33,7 @@ pub struct Model {
     model: Plan,
 }
 
-fn transpose_and_flatten(matrix: Vec<Vec<f64>>) -> Vec<f64> {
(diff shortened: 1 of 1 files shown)


y-f-u (approved):

crates/runtime/src/modelruntime/tract.rs
```diff
@@ -32,6 +33,33 @@ pub struct Model {
...
+
+    let nrows = matrix.len();
+    let ncols = matrix[0].len();
+
+    // Check for uniformity in inner vector lengths
+    for row in matrix {
+        assert!(
+            row.len() == ncols,
+            "All rows must have the same number of columns"
+        );
```
phillipleblanc, line 49: Does this indicate a bug in our code, or the user provided the wrong data? If the latter - we don't want to panic here.

phillipleblanc (approved):

PR #674, opened 2024-02-14
Improve model logging; fix an issue with view attempts to create before plan is ready

y-f-u:
This adds more verbose logging to model part.

Also fix dependent table detection for queries with subquery and with, like
```
      WITH bbb as (
            SELECT
                ROW_NUMBER() OVER (ORDER BY a.number) AS ts,
                2 * ROW_NUMBER() OVER (ORDER BY a.number) + 50 AS y,
                -3 * ROW_NUMBER() OVER (ORDER BY a.number) + 1 AS y2
            FROM
                eth_blocks a
      ), ccc as (
            SELECT
                ROW_NUMBER() OVER (ORDER BY a.number) AS ts,
                2 * ROW_NUMBER() OVER (ORDER BY a.number) + 50 AS y,
                -3 * ROW_NUMBER() OVER (ORDER BY a.number) + 1 AS y2
            FROM
                eth_blocks_2 a
      )
      SELECT nested_0.ts, CAST(nested_0.y as DOUBLE) AS y, CAST(nested_0.y2 as DOUBLE) AS y2
      FROM (
            SELECT * from bbb
            ORDER BY ts DESC
            LIMIT 50
      ) nested_0 inner join ccc on nested_0.ts = ccc.ts;
```

### DIFF
--- a/crates/runtime/src/datafusion.rs
+++ b/crates/runtime/src/datafusion.rs
@@ -225,6 +225,24 @@ impl DataFusion {
                 }
             }
 
+            let mut attempts = 1;
+            loop {
+                if ctx
+                    .state()
+                    .statement_to_plan(statements[0].clone())
+                    .await
+                    .is_ok()
+                {
+                    break;
+                }
+
+                if attempts % 10 == 0 {
+                    tracing::error!("Unable to generate plan for view, retrying...");
+                }
+                attempts += 1;
+                sleep(Duration::from_secs(1)).await;
+            }
+
             let plan = match ctx.state().statement_to_plan(statements[0].clone()).await {
                 Ok(plan) => plan,
                 Err(e) => {
@@ -232,6 +250,7 @@ impl DataFusion {
                     return;
                 }
             };
+
             let view = match ViewTable::try_new(plan, Some(sql.to_string())) {
                 Ok(view) => view,
                 Err(e) => {
--- a/crates/runtime/src/datafusion.rs
+++ b/crates/runtime/src/datafusion.rs
@@ -225,25 +225,6 @@ impl DataFusion {
                 }
             }
 
-            let mut attempts = 1;
-            loop {
-                if ctx
-                    .state()
-                    .statement_to_plan(statements[0].clone())
-                    .await
-                    .is_ok()
-                {
-                    break;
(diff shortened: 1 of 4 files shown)


crates/runtime/src/datafusion.rs
```diff
@@ -225,13 +225,33 @@ impl DataFusion {
...
+                    break;
+                }
+
+                if attempts == 10 {
+                    tracing::error!("Unable to generate plan for view, retrying...");
+                    return;
+                }
+                attempts += 1;
+                sleep(Duration::from_secs(1)).await;
+            }
```
phillipleblanc: This is the wrong fix - we need to understand why the planning errors happen and make the code above this detect that condition explicitly.

phillipleblanc: All of the planning errors we saw before were because the dependent tables didn't exist, but we're checking for that explicitly. What is the new error?

phillipleblanc (changes requested):
No shortcuts

crates/runtime/src/datafusion.rs
Jeadie: For context, this was the spicepod and error
```yaml
kind: Spicepod
name: demo
version: v1beta1
 
datasets:
- from: spice.ai/btc.blocks
  name: btc_blocks
  acceleration:
    refresh_mode: full
- name: test_view
  sql: |
    SELECT ts, CAST(y as DOUBLE) AS y, CAST(y2 as DOUBLE) as y2
    FROM (
          SELECT
              ROW_NUMBER() OVER (ORDER BY a.number) AS ts,
              2 * ROW_NUMBER() OVER (ORDER BY a.number) + 50 AS y,
              -3 * ROW_NUMBER() OVER (ORDER BY a.number) + 1 AS y2
          FROM
              btc_blocks a
          ORDER BY ts DESC
          LIMIT 50
    ) nested_0;
 
models:
  - name: demo
    from: spice.ai/Jeadie/my-first-app/darts@latest
    model_type: onnx
    datasets:
      - test_view
```

And error not logged, `HTTP/1.1 500 Internal Server Error`

phillipleblanc (approved):

PR #679, opened 2024-02-14
Dynamic pods: support for loading new datasets

sgrebnov:
1. Extracted dataset loading logic to re-usable function so it cab used by pods watcher
2. Pods watcher is now async  so it can work with datasets loading and other async functionality
3. Datafusion is shared via RWLock

### DIFF
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -4,16 +4,14 @@ use std::collections::HashMap;
 use std::env;
 use std::net::SocketAddr;
 use std::path::PathBuf;
-use std::sync::Arc;
 
 use app::App;
 use clap::Parser;
 use runtime::config::Config as RuntimeConfig;
-use runtime::dataconnector::DataConnector;
 use runtime::model::Model;
 
 use runtime::podswatcher::PodsWatcher;
-use runtime::{dataconnector, Runtime};
+use runtime::Runtime;
 use snafu::prelude::*;
 
 #[derive(Debug, Snafu)]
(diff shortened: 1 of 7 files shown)


bin/spiced/src/lib.rs
```diff
@@ -99,65 +77,9 @@ pub async fn run(args: Args) -> Result<()> {
...
-                        UnableToAttachDataConnectorSnafu {
-                            data_connector: source,
-                        },
-                    )?;
-                }
-            }
-        }
-
-        tracing::info!("Loaded dataset: {}", ds.name);
+        Runtime::load_dataset(ds, &auth, &mut df)
```
mitchdevenport, line 77: Looks like you have a conflict to resolve, I think this code was changed recently.

sgrebnov: @mitchdevenport - merged the latest from `rust`, thanks!

crates/runtime/src/lib.rs
```diff
@@ -66,7 +105,75 @@ impl Runtime {
         }
     }
 
-    pub async fn start_servers(&self) -> Result<()> {
+    pub async fn load_dataset(
```
mitchdevenport, line 111: nit: should be called `load_datasets`

mitchdevenport: Oh wait, I misread

crates/runtime/src/lib.rs
```diff
@@ -76,35 +183,41 @@ impl Runtime {
...
+        while let Some(new_app) = rx.recv().await {
+            tracing::debug!("Updated pods information: {:?}", new_app);
+            tracing::debug!("Previous pods information: {:?}", current_app);
 
-                    current_app = Arc::new(new_app);
-                }
-            }
-        };
+            // TODO: This is a placeholder to demo/test that datasets could be loaded here
+            // To be replaced with actual implementation using shared ds/datafusion instance
```
mitchdevenport: Why can't we add this now?

sgrebnov: @mitchdevenport  - added

crates/runtime/src/lib.rs
```diff
@@ -36,14 +37,53 @@ pub enum Error {
 
     #[snafu(display("Unable to start OpenTelemetry server"))]
     UnableToStartOpenTelemetryServer { source: opentelemetry::Error },
+
```
mitchdevenport, line 40: All these errors need to explicitly print the underlying source. I just merged a PR that went through and fixed this for every error in the codebase so you just need to make sure you pull in those changes properly.

mitchdevenport (approved):

crates/runtime/src/lib.rs
sgrebnov: Thank you - double checked all errors are propogated

PR #686, opened 2024-02-14
Change dremio login to spice login dremio-cloud

Sevenannn:
**DON'T MERGE THIS PR BEFORE AURASH'S SIGN OFF**


sgrebnov (approved):

bin/spice/pkg/cli/cmd/login.go
```diff
@@ -91,10 +91,10 @@ spice login
 }
 
 var dremioCmd = &cobra.Command{
-	Use:   "dremio",
+	Use:   "dremio-cloud",
 	Short: "Login to a Dremio instance",
 	Example: `
-spice login dremio --username <username> --password <password>
+spice login dremio-cloud --username <username> --password <password>
```
phillipleblanc, line 113: We'll also need to change the data connector selector in the Rust project

digadeesh (changes requested):
Let's hold on this for now.  We're having some discussions on the namespacing and we'd have to update our spicepods again with this namespace change.  Better to make a clear decision on where it should all go and then make the changes.

lukekim:
I think we decided against this rename and will keep Dremio for now. Closing.

PR #689, opened 2024-02-15
Dynamic pods: support for loading new datasets

sgrebnov:
- Extracted dataset loading logic to re-usable function so it cab used by pods watcher
 - Pods watcher is now async so it can work with datasets loading and other async functionality
 - Datafusion is shared via RWLock


crates/runtime/src/lib.rs
```diff
@@ -53,20 +93,90 @@ impl Runtime {
...
-            df: Arc::new(df),
+            auth: Arc::new(auth),
+            df: Arc::new(RwLock::new(df)),
             models: Arc::new(models),
             pods_watcher,
         }
     }
 
-    pub async fn start_servers(&self) -> Result<()> {
+    pub async fn load_dataset(
```
phillipleblanc, line 110: Should this be a mutable self function? Then we don't need to pass the auth providers

crates/runtime/src/podswatcher.rs
```diff
@@ -56,20 +51,25 @@ impl PodsWatcher {
                         }
 
                         if let Ok(app) = App::new(root_path.clone()) {
-                            event_handler(PodsWatcherEvent::PodsUpdated(app));
+                            futures::executor::block_on(async {
```
phillipleblanc: We should execute this in the tokio runtime - we should already be in that context

crates/runtime/src/lib.rs
sgrebnov: load_dataset is used [here](https://github.com/spiceai/spiceai/pull/689/files#diff-c1cad500b94aa50ce0f9f728ca73e6c4be703c15e5e5cc64f5b5644490747fe1R77) before runtime is created. I decided to keep this behavior as-is to minimize the changes and potential impact => thus static function instead of instance method. We can update this to create runtime first and them load datasets

crates/runtime/src/podswatcher.rs
sgrebnov: This is actually context of sync watcher callback. Can we keep this for tomorrow/demo and I'll update it to use mpsc channel and tokio runtime?

sgrebnov: I actually found there is tx.blocking_send so we don't need a futures::executor. Updated

phillipleblanc (approved):

PR #694, opened 2024-02-15
Use auth context from specified orgName/appName, if running login from spicepod root

ewgenius:
Switching to the new `GET /api/spice-cli/auth` endpoint in spice.ai for obtaining the preferred org/app API key when running `spice login` in the root of spicepod.

https://github.com/spiceai/spiceai/assets/827338/aacc75d1-dc90-4983-8dd2-7f4d91cbc710

### DIFF
--- a/bin/spice/pkg/cli/cmd/login.go
+++ b/bin/spice/pkg/cli/cmd/login.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
 	"github.com/spiceai/spiceai/bin/spice/pkg/api"
+	"gopkg.in/yaml.v2"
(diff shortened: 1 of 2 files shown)


bin/spice/pkg/cli/cmd/login.go
```diff
@@ -72,7 +73,22 @@ spice login
 			}
 		}
 
-		user, err := spiceApiClient.GetUser(accessToken)
+		// try reading spicepod.yaml, to check if we have preffered org and app
```
phillipleblanc:
```suggestion
		// try reading spicepod.yaml, to check if we have preferred org and app
```

phillipleblanc (approved):

PR #695, opened 2024-02-15
Rename `/v1/models/[model_name]/predictions` to  `/v1/models/[model_name]/predict` and add `/v1/models/predict` for batching support

mitchdevenport:
This PR modifies the `/v1/models/[model_name]/predictions` method to be named `/v1/models/[model_name]/predict` and adds batching support with a new method `/v1/models/predict`.

### DIFF
--- a/crates/runtime/src/http/routes.rs
+++ b/crates/runtime/src/http/routes.rs
@@ -4,7 +4,7 @@ use crate::{datafusion::DataFusion, model::Model};
 use app::App;
 
 use axum::{
-    routing::{get, Router},
+    routing::{get, post, Router},
     Extension,
 };
 
@@ -18,7 +18,7 @@ pub(crate) fn routes(
     Router::new()
         .route("/health", get(|| async { "ok\n" }))
         .route("/v1/datasets", get(v1::datasets::get))
-        .route("/v1/models/:name/predictions", get(v1::inference::get))
+        .route("/v1/models/predict", post(v1::inference::post))
         .layer(Extension(app))
         .layer(Extension(df))
         .layer(Extension(models))
--- a/crates/runtime/src/http/routes.rs
+++ b/crates/runtime/src/http/routes.rs
@@ -19,6 +19,7 @@ pub(crate) fn routes(
     Router::new()
         .route("/health", get(|| async { "ok\n" }))
         .route("/v1/datasets", get(v1::datasets::get))
+        .route("/v1/models/:name/predict", get(v1::inference::get))
         .route("/v1/models/predict", post(v1::inference::post))
         .layer(Extension(app))
         .layer(Extension(df))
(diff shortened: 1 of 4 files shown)


crates/runtime/src/http/v1.rs
```diff
@@ -52,64 +51,190 @@ pub(crate) mod inference {
...
+                duration_ms: start_time.elapsed().as_millis(),
+                predictions: model_predictions,
+            }),
+        )
+            .into_response()
+    }
+
+    async fn run_inference(
+        app: Arc<App>,
+        df: Arc<DataFusion>,
```
sgrebnov: Should be Arc<RwLock< DataFusion >>. Please merge the latest rust branch

then `df.read().await.` for usage

crates/runtime/src/http/routes.rs
```diff
@@ -18,7 +18,7 @@ pub(crate) fn routes(
     Router::new()
         .route("/health", get(|| async { "ok\n" }))
         .route("/v1/datasets", get(v1::datasets::get))
-        .route("/v1/models/:name/predictions", get(v1::inference::get))
+        .route("/v1/models/predict", post(v1::inference::post))
```
lukekim, line 23: @mitchdevenport don't we want to keep both? The option to batch and the option to use the named API?

crates/runtime/src/http/v1.rs
```diff
@@ -53,64 +53,208 @@ pub(crate) mod inference {
...
+                }
+            }
             Err(e) => {
-                tracing::error!("Unable to run inference: {}", e);
-                (StatusCode::INTERNAL_SERVER_ERROR,).into_response()
+                tracing::error!("Unable to run inference: {e}");
+                PredictResponse {
+                    status: PredictStatus::InternalError,
+                    error_message: Some(e.to_string()),
+                    model_name,
```
lukekim, line 261: @mitchdevenport can we get model version in here?

crates/runtime/src/http/v1.rs
```diff
@@ -53,64 +53,208 @@ pub(crate) mod inference {
...
+                    lookback,
+                    forecast: vec![],
```
lukekim: I would call this `prediction` over `forecast`.

crates/runtime/src/http/v1.rs
mitchdevenport: Yes, though it might be confusing because locally the user can name the model whatever they want.

Should we add the full `from` reference to disambiguate?

`spice.ai/lukekim/demo/models/darts:latest`

mitchdevenport: Actually, I think it makes sense because the user knows what they named it locally, we can always easily add the full reference later

PR #703, opened 2024-02-16
Don't require internet access for the runtime to start

mitchdevenport:
This PR makes the necessary changes to allow the runtime to start without internet access. It also makes the necessary change to allow the runtime to start even if a bad dataset is loaded. To do this, on dataset load failure the load is reattempted a configurable number of times. If it is unsuccessful after the configured retries then it gives up.

### DIFF
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -77,10 +77,7 @@ pub async fn run(args: Args) -> Result<()> {
     let pods_watcher = PodsWatcher::new(current_dir.clone());
 
     let mut rt: Runtime = Runtime::new(args.runtime, app.clone(), df, model_map, pods_watcher);
-
-    for ds in app.datasets.clone() {
-        rt.load_dataset(&ds, &auth);
-    }
+    rt.load_datasets(&auth);
 
     rt.start_servers()
         .await
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -84,7 +84,7 @@ pub async fn run(args: Args) -> Result<()> {
         pods_watcher,
         auth.clone(),
     );
-    rt.load_datasets();
+    rt.load_datasets(&auth);
 
     rt.start_servers()
         .await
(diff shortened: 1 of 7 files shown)


bin/spiced/src/lib.rs
```diff
@@ -68,24 +70,14 @@ pub struct Args {
...
-
+    let app = Arc::new(App::new(current_dir.clone()).context(UnableToConstructSpiceAppSnafu)?);
+    let auth = Arc::new(load_auth_providers());
+    let df = Arc::new(RwLock::new(runtime::datafusion::DataFusion::new()));
     let model_map = load_models(&app, &auth);
-
     let pods_watcher = PodsWatcher::new(current_dir.clone());
 
-    let mut rt: Runtime = Runtime::new(args.runtime, app, df, model_map, pods_watcher);
+    let mut rt: Runtime = Runtime::new(args.runtime, app.clone(), df, model_map, pods_watcher);
```
```suggestion
    let mut rt: Runtime = Runtime::new(args.runtime, Arc::clone(&app), df, model_map, pods_watcher);
```

crates/runtime/src/lib.rs
```diff
@@ -93,49 +95,81 @@ impl Runtime {
...
+        for ds in self.app.datasets.clone() {
+            self.load_dataset(&ds, auth);
+        }
+    }
 
-        tracing::info!("Loaded dataset: {}", &ds.name);
+    pub fn load_dataset(&self, ds: &Dataset, auth: &Arc<auth::AuthProviders>) {
+        let ds = ds.clone();
+        let df = self.df.clone();
+        let auth = auth.clone();
```
phillipleblanc: One reason I prefer using Arc::clone vs .clone() is that for example here - which ones are just cloning a pointer vs cloning the actual object?

mitchdevenport: Ugh I know but it feels so much nice to just call .clone!

Will update throughout the PR :)

lukekim (approved):

phillipleblanc (approved):
I think we want the Kubernetes model of not giving up on trying to load a dataset - but this is a good first step.

PR #706, opened 2024-02-16
Dynamic pods: support for loading new models

sgrebnov:
1. App state and models list are shared via RWLock
2. Watcher detects and loads new models

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -91,7 +91,7 @@ pub struct Runtime {
     pub df: Arc<RwLock<DataFusion>>,
     pub models: Arc<RwLock<HashMap<String, Model>>>,
     pub pods_watcher: podswatcher::PodsWatcher,
-    pub auth: Arc<auth::AuthProviders>,
+    pub auth: Arc<RwLock<auth::AuthProviders>>,
 
     spaced_tracer: Arc<tracers::SpacedTracer>,
 }
@@ -103,7 +103,7 @@ impl Runtime {
         app: Arc<RwLock<app::App>>,
         df: Arc<RwLock<DataFusion>>,
         pods_watcher: podswatcher::PodsWatcher,
-        auth: Arc<auth::AuthProviders>,
+        auth: Arc<RwLock<auth::AuthProviders>>,
     ) -> Self {
         Runtime {
             app,
@@ -116,22 +116,25 @@ impl Runtime {
         }
     }
 
-    pub async fn load_datasets(&self, auth: &Arc<auth::AuthProviders>) {
+    pub async fn load_datasets(&self) {
         for ds in self.app.read().await.datasets.clone() {
-            self.load_dataset(ds, auth);
+            self.load_dataset(ds);
         }
     }
 
-    pub fn load_dataset(&self, ds: Dataset, auth: &Arc<auth::AuthProviders>) {
+    pub fn load_dataset(&self, ds: Dataset) {
         let df = Arc::clone(&self.df);
-        let auth = Arc::clone(auth);
         let spaced_tracer = Arc::clone(&self.spaced_tracer);
+        let shared_auth = Arc::clone(&self.auth);
+
         tokio::spawn(async move {
             loop {
                 if ds.acceleration.is_none() && !ds.is_view() {
                     tracing::warn!("No acceleration specified for dataset: {}", ds.name);
                     break;
                 };
+
+                let auth = shared_auth.read().await;
 
                 let source = ds.source();
                 let source = source.as_str();
@@ -279,19 +282,16 @@ impl Runtime {
         Ok(())
     }
 
-    pub async fn load_models(&self, auth: &Arc<auth::AuthProviders>) {
+    pub async fn load_models(&self) {
         for model in self.app.read().await.models.clone() {
-            self.load_model(&model, auth).await;
+            self.load_model(&model).await;
         }
     }
 
-    pub async fn load_model(
-        &self,
-        m: &spicepod::component::model::Model,
-        auth: &auth::AuthProviders,
-    ) {
+    pub async fn load_model(&self, m: &spicepod::component::model::Model) {
         tracing::info!("Loading model [{}] from {}...", m.name, m.from);
         let mut model_map = self.models.write().await;
+        let auth = self.auth.read().await;
 
         match Model::load(m, auth.get(m.source().as_str())) {
             Ok(in_m) => {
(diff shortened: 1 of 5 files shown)


crates/runtime/src/lib.rs
```diff
@@ -304,14 +333,18 @@ impl Runtime {
...
 
         while let Some(new_app) = rx.recv().await {
+            let current_app = self.app.read().await;
+            let current_datasets = current_app.datasets.clone();
+            let current_models = current_app.models.clone();
+
             tracing::debug!("Updated pods information: {:?}", new_app);
             tracing::debug!("Previous pods information: {:?}", current_app);
 
+            drop(current_app);
```
mitchdevenport: Did you mean to drop the app here?

crates/runtime/src/lib.rs
```diff
@@ -304,14 +333,18 @@ impl Runtime {
...
+
             let mut auth = auth::AuthProviders::default();
```
mitchdevenport: Let's do the same thing with auth. Move it behind an Arc<RwLock<>> and then just update the auth here so auth is updated throughout the runtime instead of just passing it in to the `load()` methods.

crates/runtime/src/lib.rs
sgrebnov: This is intentionally as we would like to update the app later (acquire write lock). Don't get write lock immediately as load/modify functions for datasets and models could be modified in the future and would like to access app state which will result in the deadlock.

mitchdevenport: As you have it structured now this introduces a race condition. If I add 2 new datasets in quick succession then the first will read the apps/models and drop the app. If the 2nd dataset reads the existing apps and models before the 1st sets the new app then it will try to read a non-existent app. Same goes for anything that is trying to read during the window between the old app being dropped and the new one being set.

mitchdevenport: You're basically defeating the purpose of the Arc which keeps a reference count to ensure the app isn't dropped while something still needs it.

mitchdevenport: If you want to ensure the models and datasets aren't changed after this code reads the models/datasets then you need to acquire the write lock instead. That won't cause deadlock as long as you release the lock once you're done updating.

mitchdevenport: I misread this, you're dropping the lock, not the app itself.

crates/runtime/src/lib.rs
```diff
@@ -304,14 +333,18 @@ impl Runtime {
     }
 
     pub async fn start_pods_watcher(&mut self) -> notify::Result<()> {
-        let mut current_app = Arc::clone(&self.app);
-
         let mut rx = self.pods_watcher.watch()?;
 
         while let Some(new_app) = rx.recv().await {
+            let current_app = self.app.read().await;
```
mitchdevenport: This code relies on the app state not changing between reading and writing so we should acquire the write lock here and then release it at the end of the loop iteration.

crates/runtime/src/lib.rs
```diff
@@ -304,14 +333,18 @@ impl Runtime {
...
+            let current_datasets = current_app.datasets.clone();
```
mitchdevenport: Once you acquire the write lock you don't have to clone the data anymore, you can just use the write lock to read below.

crates/runtime/src/lib.rs
sgrebnov: The only concern here is that loading models/databases/other future functionality might want to use current app state. If we acquire write lock immediately, then it will result in a deadlock. As we have single thread/channel responsible for the state change then it should be fine not to lock the app immediately IMO.

When we will be supporting removing datasets//models/etc we will need to lock app before we start doing this (as you are pointing here.)

mitchdevenport: How would acquiring the lock immediately result in deadlock? Just acquire, read, write, then release the lock. I think it's very unlikely that we will continue to have a single thread changing state moving forward so relying on that isn't a good solution.

sgrebnov: I've just tested `while let Some(new_app) = rx.recv().await` for potential race condition with 5 sec async/await delay inside and its executed sequentially so we should be good for the potential case when app states are processed in parallel leading to broken state

mitchdevenport: This particular thread won't race with itself but my concern is other threads changing the state. The reason we have the RwLock is to prevent this, so let's use it properly. Your code works on it's own for now, but it should work properly with other code reading/writing the App too.

crates/runtime/src/lib.rs
sgrebnov: Updated to keep lock longer so we don't need to clone data.

crates/runtime/src/lib.rs
```diff
@@ -304,42 +333,46 @@ impl Runtime {
...
-                );
-            }
-            let auth_arc = Arc::new(auth);
-
-            let existing_dataset_names = current_app
-                .datasets
-                .iter()
-                .map(|ds| ds.name.clone())
-                .collect::<Vec<String>>();
+            *self.auth.write().await = load_auth_providers();
```
sgrebnov, line 345: TODO: update only if we loaded auth providers successfully

mitchdevenport (approved):

crates/runtime/src/lib.rs
```diff
@@ -279,6 +282,32 @@ impl Runtime {
         Ok(())
     }
 
+    pub async fn load_models(&self) {
+        for model in self.app.read().await.models.clone() {
+            self.load_model(&model).await;
+        }
+    }
+
+    pub async fn load_model(&self, m: &spicepod::component::model::Model) {
```
```suggestion
    pub async fn load_model(&self, m: &Model) {
```

crates/runtime/src/lib.rs
```diff
@@ -279,6 +282,32 @@ impl Runtime {
         Ok(())
     }
 
+    pub async fn load_models(&self) {
+        for model in self.app.read().await.models.clone() {
```
phillipleblanc: If we're only passing a reference to a Model, do we need to clone it?

crates/runtime/src/lib.rs
sgrebnov: We actually have here `spicepod::component::model::Model ` and `runtime::model::Model` with similar names and conflicting. Updated to have it as `spicepod::component::model::Model as SpicepodModel;`

```
  pub async fn load_model(&self, m: &SpicepodModel) {
        tracing::info!("Loading model [{}] from {}...", m.name, m.from);
```

crates/runtime/src/lib.rs
sgrebnov: Updated, thank you!
```
pub async fn load_models(&self) {
        for model in &self.app.read().await.models {
            self.load_model(model).await;
        }
    }
```

phillipleblanc (approved):

PR #707, opened 2024-02-17
Async model load and pull

sgrebnov:
When I test model loading locally the following line panics with the following error on my laptop (rust branch w/o changes). Making this logic async fixes this plus makes it more flexible allowing using async functionality moving forward.

[let client = reqwest::blocking::Client::new();](https://github.com/spiceai/spiceai/blob/rust/crates/runtime/src/modelsource/spiceai.rs#L89C9-L89C55)

```
2024-02-16T22:24:54.205714Z  INFO spiced: Deploying model [drive_stats] from spice.ai/lukekim/smart/models/drive_stats:latest...
thread 'main' panicked at /Users/sg/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.36.0/src/runtime/blocking/shutdown.rs:51:21:
**Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.**
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### DIFF
--- a/crates/runtime/src/modelsource.rs
+++ b/crates/runtime/src/modelsource.rs
@@ -1,8 +1,8 @@
 use crate::auth::AuthProvider;
+use async_trait::async_trait;
 use snafu::prelude::*;
 use std::collections::HashMap;
 use std::sync::Arc;
-use tonic::async_trait;
 
 pub mod local;
 pub mod spiceai;
--- a/crates/runtime/src/modelsource.rs
+++ b/crates/runtime/src/modelsource.rs
@@ -2,6 +2,7 @@ use crate::auth::AuthProvider;
 use snafu::prelude::*;
 use std::collections::HashMap;
 use std::sync::Arc;
+use tonic::async_trait;
 
 pub mod local;
 pub mod spiceai;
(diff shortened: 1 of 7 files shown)


y-f-u (approved):

crates/runtime/src/modelsource.rs
```diff
@@ -2,6 +2,7 @@ use crate::auth::AuthProvider;
 use snafu::prelude::*;
 use std::collections::HashMap;
 use std::sync::Arc;
+use tonic::async_trait;
```
phillipleblanc: It's weird that we're using the tonic async_trait here. tonic is a GRPC library - what does that have to do with a model source?

sgrebnov: Updated to import async-trait explicitly from async-trait https://github.com/spiceai/spiceai/pull/707/commits/eb0d8680c761edaeb072a4c888ac44fa0712fcf9

Previously:
tonic::async_trait was  a re-export of [async-trait](https://docs.rs/async-trait) (see https://docs.rs/tonic/latest/tonic/attr.async_trait.html)

async-trait is recommended solution by rust async book: https://rust-lang.github.io/async-book/07_workarounds/05_async_in_traits.html

phillipleblanc (approved):

PR #715, opened 2024-02-20
Refactor DuckDBTableProvider to be a generic SQLTableProvider

mitchdevenport:
This PR updates the existing DuckDBTableProvider for datafusion to be a generic SQLTableProvider that can be reused for multiple SQL databases.


crates/sql_provider_datafusion/src/dbconnectionpool.rs
```diff
@@ -0,0 +1,41 @@
...
+use spicepod::component::dataset::acceleration;
+
+use crate::dbconnection::DbConnection;
+
+pub mod duckdbpool;
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("DuckDBError: {source}"))]
+    DuckDBError { source: duckdb::Error },
```
y-f-u, line 13: I think we have a tendency not exposing implementation details on the trait level

y-f-u: no I am wrong.

crates/sql_provider_datafusion/src/dbconnection.rs
```diff
@@ -0,0 +1,22 @@
+use ::duckdb::arrow::array::RecordBatch;
+use datafusion::{arrow::datatypes::SchemaRef, sql::TableReference};
+use snafu::prelude::*;
+
+pub mod duckdbconn;
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("DuckDBError: {source}"))]
+    DuckDBError { source: duckdb::Error },
```
y-f-u: nit. if we put DuckDBError here, for any new type of implementation, we will need to put a new error type, e.g. PostgresError here even we implement this trait outside the crate.

Maybe wrapped it into a dyn Error or some other generic error type?

mitchdevenport: Good point

crates/sql_provider_datafusion/src/dbconnectionpool.rs
```diff
@@ -0,0 +1,31 @@
...
+        }
+    }
+}
+
+pub trait DbConnectionPool<T: r2d2::ManageConnection, C, P: 'static> {
+    fn new(name: &str, mode: Mode, params: Arc<Option<HashMap<String, String>>>) -> Result<Self>
+    where
+        Self: Sized;
+    fn connect(&self) -> Result<Box<dyn DbConnection<T, P>>>;
+    fn connect_downcast(&self) -> Result<C>;
```
y-f-u: nit. Another way to implement this might be pushing the C type into DbConnection, then have the conn() or a downcast() on DbConnection to return a implementation specific for other queries.

crates/sql_provider_datafusion/src/dbconnection.rs
mitchdevenport: Actually, this caused issues using the crate because it doesn't implement Send. I think it also makes the errors harder to work with

crates/sql_provider_datafusion/src/dbconnectionpool.rs
mitchdevenport: Looks like the idiomatic way to do this is to implement `as_any` and then use `downcast_ref`

crates/sql_provider_datafusion/src/dbconnection.rs
mitchdevenport: got this working, done.

crates/sql_provider_datafusion/src/dbconnectionpool.rs
```diff
@@ -0,0 +1,31 @@
+use std::{collections::HashMap, sync::Arc};
+
+use spicepod::component::dataset::acceleration;
```
phillipleblanc: This is a weird dependency for a generic sql_provider_datafusion crate to have.

phillipleblanc (approved):

crates/sql_provider_datafusion/src/dbconnectionpool.rs
mitchdevenport: Good point, moved into spicepod

PR #716, opened 2024-02-20
Allow dataset to be initialized if there is a table provider for the data source even no acceleration defined

y-f-u:
This is the first step to allow user to do a `mesh` query: even the table is not accelerated in local, the query will be sent down to the data source.

With this change, instead of erroring out when no acceleration detected for dataset, it will check if there is a `TableProvider` defined for the data connector.

`TableProvider` is datafusion concept for extending queries on data not in local. https://arrow.apache.org/datafusion/library-user-guide/custom-table-providers.html

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -162,7 +162,7 @@ impl Runtime {
                 };
 
                 if ds.acceleration.is_none() && !ds.is_view() && !has_table_provider {
-                    tracing::warn!("No acceleration specified for dataset and no table provider found for dataset's connector: {}", ds.name);
+                    tracing::warn!("No acceleration specified for dataset '{}' and no table provider found for dataset's connector '{}'", ds.name, source);
                     break;
                 };
 
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -151,18 +151,11 @@ impl Runtime {
                         }
                     };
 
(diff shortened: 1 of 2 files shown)


crates/runtime/src/lib.rs
```diff
@@ -156,6 +151,21 @@ impl Runtime {
...
+                if data_connector.is_some()
+                    && data_connector
+                        .as_ref()
+                        .is_some_and(|dc| dc.get_table_provider().is_some())
+                {
+                    has_table_provider = true;
+                };
+
+                if ds.acceleration.is_none() && !ds.is_view() && !has_table_provider {
+                    tracing::warn!("No acceleration specified for dataset and no table provider found for dataset's connector: {}", ds.name);
```
phillipleblanc: We shouldn't log anything about a table provider to the user

crates/runtime/src/lib.rs
```diff
@@ -156,6 +151,21 @@ impl Runtime {
...
 
+                let mut has_table_provider = false;
+
+                if data_connector.is_some()
+                    && data_connector
+                        .as_ref()
+                        .is_some_and(|dc| dc.get_table_provider().is_some())
+                {
+                    has_table_provider = true;
+                };
```
phillipleblanc: Let's extract this as a function

phillipleblanc (approved):

PR #720, opened 2024-02-21
Keep Spice Runtime running when no spicepod is provided

Sevenannn:
**DON'T MERGE THIS BEFORE LUKE'S DEMO**

# What
Keep Spice Runtime running when no spicepod is provided at `spice run`. Display instructions to users so that they know `spice init` should be run.

# How
The basic idea here is not to panic at App::new when no spicepod is provided, and mark such app as `None` type. We also don't executed any function logic (e.g. load_datasets() load_models() ) that require a `Some(App)` when no spicepod is provided and hence the `None` type app is presented. 

# Spec
[spice run with no spicepod defined will wait for spicepod to be created](https://github.com/orgs/spicehq/projects/38/views/1?pane=issue&itemId=53676141)

# Key changes
### App
Convert app from type `Arc<RwLock<App>>` to `<Arc<RwLock<Option<App>>>`

### Runtime
load_datasets() load_models() only execute its original logic when `Option<App>` is not `None`
start_servers() keep running whether app exists or not, this is to keep the run time alive regardless of whether spicepod is provided.
start_pods_watcher() will assign the new app to self.app if no spicepod is provided.

### Error Log
Show warning message instead of error when spicepod is not provided and app creation fail for the first time.
Add the `spice init` instruction for `SpicepodNotFound` error
Show user friendly error messages instead of json blob when no spicepod is provided.


bin/spiced/src/lib.rs
```diff
@@ -68,14 +68,19 @@ pub struct Args {
...
     let current_dir = env::current_dir().unwrap_or(PathBuf::from("."));
-    let app = Arc::new(RwLock::new(
-        App::new(current_dir.clone()).context(UnableToConstructSpiceAppSnafu)?,
-    ));
     let auth = Arc::new(RwLock::new(runtime::load_auth_providers()));
     let df = Arc::new(RwLock::new(runtime::datafusion::DataFusion::new()));
     let pods_watcher = PodsWatcher::new(current_dir.clone());
+    let app = match App::new(current_dir.clone()).context(UnableToConstructSpiceAppSnafu) {
+        Ok(app) => Some(Arc::new(RwLock::new(app))),
+        Err(e) => {
```
sgrebnov: Wondering if we can just fallback to empty/default app instead of introducing the Option

crates/runtime/src/lib.rs
```diff
@@ -87,7 +87,7 @@ pub enum Error {
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
 pub struct Runtime {
-    pub app: Arc<RwLock<App>>,
+    pub app: Option<Arc<RwLock<App>>>,
```
phillipleblanc: This should be `Arc<RwLock<Option<App>>>` - putting the Option first defeats the point of having the Arc & RwLock wrappers - since only one object can reference the variable with the Option.

Putting the Option inside means that we have the flexibility to replace the app inside the Arc<RwLock<>> and have that be reflected in all places that have that reference - otherwise what you end up with is several places in code with their view of App, but when its updated in one place, the other places in code that reference the App will still be referencing the old version. Which will be a source of subtle, hard to debug bugs.

bin/spiced/src/lib.rs
phillipleblanc: I like the Option, but as I noted in another comment we need to tweak it

crates/spicepod/src/lib.rs
```diff
@@ -21,7 +21,7 @@ pub enum Error {
         source: component::Error,
         path: PathBuf,
     },
-    #[snafu(display("spicepod.yaml not found in {}", path.display()))]
+    #[snafu(display("spicepod.yaml not found in {}, run `spice init <pod>` to initiate spicepod.yaml", path.display()))]
```
lukekim:
```suggestion
    #[snafu(display("spicepod.yaml not found in {}, run `spice init <name>` to initialize spicepod.yaml", path.display()))]
```

Sevenannn: Updated

bin/spiced/src/lib.rs
Sevenannn: Thanks for the suggestions! I updated the code for tweaking the option, please help take a look if you are available.

crates/runtime/src/lib.rs
Sevenannn: Thanks for the suggestions! I updated to Arc<RwLock<Option<App>>> and the new code is ready for review

phillipleblanc (approved):

bin/spiced/src/lib.rs
phillipleblanc, line 1: Looks good - but let's wait to merge this until after Luke's demo

Sevenannn: Got it, thanks for the review!

digadeesh (approved):

PR #726, opened 2024-02-21
Dynamic spicepods: support for editing and removing datasets

sgrebnov:
Basic implementation to unload removed dataset and re-create a dataset if it has been modified.

Should we introduce some Drop/Shutdown method to DataConnector or DataBackend so we can implement custom cleanup / shutdown logic per each provider(for example remove file when we use file mode)?


crates/runtime/src/datafusion.rs
```diff
@@ -70,7 +76,7 @@ type DatasetAndPublishers = (Arc<Dataset>, PublisherList);
 
 pub struct DataFusion {
     pub ctx: Arc<SessionContext>,
-    tasks: Vec<task::JoinHandle<()>>,
+    tasks: HashMap<String, task::JoinHandle<()>>,
```
sgrebnov: TODO: rename `tasks` so it more descriptive

crates/runtime/src/datafusion.rs
```diff
@@ -170,8 +176,8 @@ impl DataFusion {
         data_connector: Box<dyn DataConnector>,
         publisher: Arc<Box<dyn DataPublisher>>,
     ) -> Result<()> {
-        let table_name = dataset.name.as_str();
-        let table_exists = self.ctx.table_exist(table_name).unwrap_or(false);
+        let table_name = dataset.name.clone();
+        let table_exists = self.ctx.table_exist(table_name.as_str()).unwrap_or(false);
```
phillipleblanc, line 180: we clone the string just to reference it? That doesn't look right.

crates/runtime/src/datafusion.rs
```diff
@@ -193,7 +199,38 @@ impl DataFusion {
...
+    #[must_use]
+    pub fn table_exists(&self, dataset_name: &str) -> bool {
+        self.ctx.table_exist(dataset_name).unwrap_or(false)
+    }
+
+    pub fn remove_table(&mut self, dataset_name: &str) -> Result<()> {
+        let table_exists = self.ctx.table_exist(dataset_name).unwrap_or(false);
+        if !table_exists {
+            return TableDoesNotExistSnafu.fail();
+        }
```
phillipleblanc: I don't think we should error out here - if we want to remove something that doesn't exist, we just return success without doing anything.

phillipleblanc (approved):

crates/runtime/src/datafusion.rs
sgrebnov: Updated.

PR #727, opened 2024-02-21
spice.ai table provider

y-f-u:
NOTE: this still doesn't support `count` and `distinct` etc. Which will work on after this one merged. Given it will be the same logic to enable all table providers we have now

With this change, spice.ai and dremio data source can be queried without acceleration. The query will go to the remote and fetched via flight.

e.g.  with this yaml
```yaml
kind: Spicepod
name: demo
version: v1beta1

datasets:
  name: eth_blocks
  # acceleration:
  #   enabled: true
  #   refresh_mode: full
  name: blocks
  acceleration:
    enabled: true
    refresh_mode: full
- name: test_view
  sql: select number as ts, (number::double / 100) as y, (number::double) / 100 as y2 from eth_blocks order by ts desc limit 100
```

queries like this can be done:
```bash
sql> select number from eth_blocks where number not in (select number from blocks);
+----------+
| number   |
+----------+
| 19281206 |
| 19281205 |
| 19281204 |
| 19281203 |
| 19281201 |
| 19281202 |
| 19281199 |
+----------+

Query took: 0.971788792 seconds
```


crates/flight_datafusion/src/lib.rs
```diff
@@ -0,0 +1,310 @@
...
+        _context: Arc<TaskContext>,
+    ) -> DataFusionResult<SendableRecordBatchStream> {
+        let sql = match self.sql().map_err(to_execution_error) {
+            Ok(sql) => sql,
+            Err(error) => return Err(error),
+        };
+
+        let data = tokio::task::block_in_place(move || {
+            Handle::current().block_on(async {
+                let result = self.client.clone().query(sql.as_str()).await;
```
mitchdevenport: Don't clone the client directly, use an Arc to share it between threads

crates/flight_datafusion/src/lib.rs
```diff
@@ -0,0 +1,330 @@
+#![allow(clippy::missing_errors_doc)]
```
y-f-u, line 1: This file still has duplication with sql_provider_datafusion.

The idea is to wait until MItch's branch merged, then work on another refactoring. As, for flight client, there is no need for the concept the ConnectionPool, also it doesn't need to have the implement DbManager in SQLTable too.

Merging and refactoring now may cause stepping over other's toes. So please ignore the duplication.

crates/flight_datafusion/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+            Arc::clone(&self.client),
+            sql.as_str(),
+            self.schema(),
+        )))
+    }
+}
+
+#[allow(clippy::needless_pass_by_value)]
+fn to_stream(client: Arc<FlightClient>, sql: &str) -> impl Stream<Item = Result<RecordBatch>> {
+    let mut client = client.deref().clone();
```
y-f-u: arc doesn't work well with async

y-f-u: intentional copy

crates/runtime/src/dataconnector/debug.rs
```diff
@@ -82,4 +82,8 @@ impl DataConnector for DebugSource {
           }
         })
     }
+
+    fn has_table_provider(&self) -> bool {
```
phillipleblanc: no need to implement this if the default implementation is the same

crates/runtime/src/datafusion.rs
```diff
@@ -230,6 +233,31 @@ impl DataFusion {
         Ok(())
     }
 
+    #[allow(clippy::needless_pass_by_value)]
+    pub fn attach_remote(
```
```suggestion
    pub fn attach_mesh(
```

crates/flight_datafusion/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+    filters: Vec<Expr>,
+    limit: Option<usize>,
+}
+
+impl FlightSQLExec {
+    fn new(
+        projections: Option<&Vec<usize>>,
+        schema: &SchemaRef,
+        table_reference: &OwnedTableReference,
+        client: Arc<FlightClient>,
```
phillipleblanc: There is no point wrapping a FlightClient in an Arc - because as you've discovered, we need to clone it to be able to query. I would just remove the Arc<>

phillipleblanc: The real way to fix this would be using a connection pooler - should be fairly straightforward using r2d2 - but let's track that separately

crates/flight_datafusion/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+    ArrowFlight { source: FlightError },
+
+    #[snafu(display("Unable to retrieve schema"))]
+    NoSchema,
+}
+
+type Result<T, E = Error> = std::result::Result<T, E>;
+
+pub struct FlightSQLTable {
+    client: Arc<FlightClient>,
```
phillipleblanc:
```suggestion
    client: FlightClient,
```

crates/flight_datafusion/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+    #[snafu(display("Unable to query FlightSQL: {source}"))]
+    ArrowFlight { source: FlightError },
+
+    #[snafu(display("Unable to retrieve schema"))]
+    NoSchema,
+}
+
+type Result<T, E = Error> = std::result::Result<T, E>;
+
+pub struct FlightSQLTable {
```
phillipleblanc: The naming here is weird - the sources you are connecting to aren't FlightSQL endpoints - they are normal Flight endpoints that support SQL queries using the `Cmd` FlightDescriptors - which are two separate things.

Mitch is implementing a real `FlightSQL` provider - so I think we should rename this to `FlightTable`

crates/flight_datafusion/src/lib.rs
phillipleblanc: As mentioned in another comment - we will eventually need the connection pool for FlightClient

crates/flight_datafusion/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+            "SELECT {columns} FROM {table_reference} {where_expr} {limit_expr}",
+            table_reference = self.table_reference,
+        ))
+    }
+}
+
+impl std::fmt::Debug for FlightSQLExec {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        let sql = self.sql().unwrap_or_default();
+        write!(f, "FlightSQLExec sql={sql}")
```
phillipleblanc:
```suggestion
        write!(f, "FlightExec sql={sql}")
```

crates/flight_datafusion/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+}
+
+impl DisplayAs for FlightSQLExec {
+    fn fmt_as(&self, _t: DisplayFormatType, f: &mut fmt::Formatter) -> std::fmt::Result {
+        let sql = self.sql().unwrap_or_default();
+        write!(f, "FlightSQL sql={sql}")
```
phillipleblanc:
```suggestion
        write!(f, "FlightExec sql={sql}")
```

crates/flight_datafusion/src/lib.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+
+    #[allow(clippy::needless_pass_by_value)]
+    fn get_schema<'a>(
+        client: Arc<FlightClient>,
+        table_reference: impl Into<TableReference<'a>>,
+    ) -> Result<SchemaRef> {
+        let mut client = client.deref().clone();
+        tokio::task::block_in_place(move || {
+            Handle::current().block_on(async {
```
phillipleblanc: We shouldn't do this - make `get_schema` and `new` async

phillipleblanc: this means you will also need to make `get_table_provider` and `attach_mesh` async as well.

crates/flight_datafusion/src/lib.rs
y-f-u: sounds good, can be done when refactoring them.

crates/flight_datafusion/src/lib.rs
y-f-u: yeah will do in the refactoring

phillipleblanc (approved):

PR #732, opened 2024-02-22
Change Spicepod DataConnector/ModelSource selector

phillipleblanc:
## What

The Spicepod `from` will now select the DataConnector/ModelSource to load based on the first string before a colon (`:`) rather than a slash (`/`). If there is no colon, then we default to the Spice.ai Data Connector/Model Source.

The Spice.ai Data Connector will also change to accept new formats:

```
spiceai:spice.ai/lukekim/demo/datasets/my_data
spiceai:spice.ai/lukekim/demo/my_data
spiceai:lukekim/demo/datasets/my_data
spiceai:lukekim/demo/my_data
spice.ai/lukekim/demo/datasets/my_data
spice.ai/lukekim/demo/my_data
lukekim/demo/datasets/my_data
lukekim/demo/my_data
```

Would all be equivalent - they select the `my_data` dataset from the Spice.ai App `lukekim/demo`

## Why

The current solution to select the dataconnector is confusing when it isn't a URL, as is the case with `dremio`.

### DIFF
--- a/.github/workflows/build_and_release.yml
+++ b/.github/workflows/build_and_release.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - trunk
+      - rust
       - release-*
     tags:
       - v*
--- a/bin/spice/pkg/api/auth.go
+++ b/bin/spice/pkg/api/auth.go
@@ -1,7 +1,7 @@
 package api
 
 const (
-	AUTH_TYPE_SPICE_AI = "spice.ai"
+	AUTH_TYPE_SPICE_AI = "spiceai"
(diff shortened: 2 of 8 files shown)


.github/workflows/build_and_release.yml
phillipleblanc, line 1: Blocking merge until after Luke's demo is complete.

digadeesh (approved):

.github/workflows/build_and_release.yml
lukekim: @phillipleblanc demo complete and successful!

PR #735, opened 2024-02-22
No hardcoded dremio endpoint

digadeesh:
https://github.com/spicehq/spiceai/issues/100

This is to remove the hardcoded dremio endpoint when you do a spice login dremio.  

now users have to include the dremio endpoint with a -e "endpoint".  It adds the endpont to the ~/.spice/auth file with dremio username and password.  It removes the dataset configuration from datasets.yaml

It has some limitations.  We couldn't have multiple dremios (I don't think we can now atm).  So maybe its better with the dataset configuration.

### DIFF
--- a/bin/spice/pkg/cli/cmd/login.go
+++ b/bin/spice/pkg/cli/cmd/login.go
@@ -19,7 +19,6 @@ const (
 	apiKeyFlag   = "key"
 	usernameFlag = "username"
 	passwordFlag = "password"
-	endpointFlag = "endpoint"
 	charset      = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
 )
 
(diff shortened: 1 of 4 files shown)


bin/spice/pkg/cli/cmd/login.go
```diff
@@ -138,10 +139,22 @@ spice login dremio --username <username> --password <password>
...
+		if endpoint == "" {
+			cmd.Println("No endpoint provided, use --endpoint or -p to provide an endpoint")
+			os.Exit(1)
+		}
+
 		mergeAuthConfig(cmd, api.AUTH_TYPE_DREMIO, &api.Auth{
 			Params: map[string]string{
 				api.AUTH_PARAM_USERNAME: username,
 				api.AUTH_PARAM_PASSWORD: password,
+				api.AUTH_PARAM_ENDPOINT: endpoint,
```
mitchdevenport: I think this should be a property of the dataset instead of a property of the auth, so `spice dataset configure` would ask for an endpoint and then write it to the dataset params. If auth determines the dataset location and not the dataset, then sharing spicepods becomes difficult because users will have to ensure that their auth configurations match.

digadeesh: Yeah agree and its also a much simpler implementation, it turns out.  I've updated the PR.

mitchdevenport (approved):

PR #753, opened 2024-02-29
Initial Postgres support

mitchdevenport:
This adds initial support for Postgres as a data backend in the Spice runtime.


crates/runtime/src/databackend/postgres.rs
```diff
@@ -0,0 +1,227 @@
...
+    create_mutex: &'a Mutex<()>,
+}
+
+impl<'a> PostgresUpdate<'a> {
+    async fn update(&mut self) -> Result<()> {
+        if !self.table_exists().await {
+            self.create_table()?;
+        } else if self.update_type == UpdateType::Overwrite {
+            self.postgres_conn
+                .execute(format!(r#"TRUNCATE TABLE "{}""#, self.name).as_str(), &[])
```
phillipleblanc: We should run this update in a transaction - and truncate breaks out of transactions on Postgres, so it would need to be a `DELETE FROM "{}"`

crates/runtime/src/databackend/postgres.rs
```diff
@@ -0,0 +1,227 @@
...
+            name = self.name
+        );
+        tracing::trace!("{sql}");
+
+        let Ok(row) = self.postgres_conn.conn.query_one(&sql, &[]).await else {
+            return false;
+        };
+
+        let exists: bool = row.get(0);
+        exists
```
phillipleblanc:
```suggestion
        row.get(0)
```

crates/sql_provider_datafusion/src/dbconnection.rs
```diff
@@ -1,20 +1,22 @@
...
-pub trait DbConnection<T, P> {
+pub trait DbConnection<T, P>: Send {
     fn new(conn: T) -> Self
     where
         Self: Sized;
-    fn get_schema(&mut self, table_reference: &TableReference) -> Result<SchemaRef>;
-    fn query_arrow(&mut self, sql: &str, params: &[P]) -> Result<Vec<RecordBatch>>;
-    fn execute(&mut self, sql: &str, params: &[P]) -> Result<u64>;
+    fn get_schema(&self, table_reference: &TableReference) -> Result<SchemaRef>;
+    fn query_arrow(&self, sql: &str, params: &[P]) -> Result<SendableRecordBatchStream>;
```
phillipleblanc, line 18: Can we make these async functions so that we don't need to block the tokio executor?

mitchdevenport: Unfortunately as we've currently structured our code it's not that simple because the DuckDB connection and pool library we're using is not thread safe. Yong is looking into the right thing to do here: https://github.com/spicehq/spiceai/issues/151

crates/arrow_sql_gen/src/statement.rs
```diff
@@ -55,6 +65,181 @@ impl CreateTableBuilder {
...
+    #[must_use]
+    pub fn new(table_name: &str, record_batches: Vec<RecordBatch>) -> Self {
+        Self {
+            table_name: table_name.to_string(),
+            record_batches,
+        }
+    }
+
+    #[allow(clippy::too_many_lines)]
+    pub fn construct_insert_stmt(
```
phillipleblanc, line 83: FYI, no action needed: This will be a useful fallback to databases we support which don't have a more efficient way of inserting data - but for Postgres specifically, we can leverage the [binary protocol](https://www.postgresql.org/docs/current/sql-copy.html). @y-f-u is working on that.

phillipleblanc (dismissed):

crates/arrow_sql_gen/src/statement.rs
y-f-u: there is a copy_into method that create a sink as well as a binary in thing we can utilise.

The implementation will be quite straight forward, just FYI. I mentioned it previous in the ticket, but seems it's removed now. I will create a card and assign it to me.

phillipleblanc (approved):

PR #760, opened 2024-03-02
Fix missing app extension in routes, due to type mismatch

ewgenius:
From changes in #720, app object is passed with type `Arc<RwLock<Option<App>>>`
https://github.com/spiceai/spiceai/blob/63ff3bbd6e0cf5b02efc871a83baabf6ff8a90b1/crates/runtime/src/http/routes.rs#L15

It was still expected as `Arc<RwLock<App>>` in some handlers, which was causing following error:

CleanShot 2024-03-02 at 23 56 54@2x

After the fix:

CleanShot 2024-03-02 at 23 57 17@2x

### DIFF
--- a/crates/runtime/src/http/v1.rs
+++ b/crates/runtime/src/http/v1.rs
@@ -85,19 +85,26 @@ pub(crate) mod datasets {
     }
 
     pub(crate) async fn get(
-        Extension(app): Extension<Arc<RwLock<App>>>,
+        Extension(app): Extension<Arc<RwLock<Option<App>>>>,
         Query(filter): Query<DatasetFilter>,
     ) -> Json<Vec<Dataset>> {
+        let app_lock = app.read().await;
+        let readable_app = match &*app_lock {
+            Some(app) => app,
+            None => {
+                tracing::debug!("App not found");
+                return Json(vec![]);
+            }
+        };
+
         let mut datasets: Vec<Dataset> = match filter.source {
-            Some(source) => app
-                .read()
-                .await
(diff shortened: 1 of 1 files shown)


crates/runtime/src/http/v1.rs
```diff
@@ -85,19 +85,23 @@ pub(crate) mod datasets {
     }
 
     pub(crate) async fn get(
-        Extension(app): Extension<Arc<RwLock<App>>>,
+        Extension(app): Extension<Arc<RwLock<Option<App>>>>,
         Query(filter): Query<DatasetFilter>,
     ) -> Json<Vec<Dataset>> {
+        let app_lock = app.read().await;
+        let Some(readable_app) = &*app_lock else {
+            tracing::debug!("App not found");
```
phillipleblanc:
```suggestion
```

crates/runtime/src/http/v1.rs
```diff
@@ -245,15 +249,28 @@ pub(crate) mod inference {
...
         models: Arc<RwLock<HashMap<String, Model>>>,
         model_name: String,
         lookback: usize,
     ) -> PredictResponse {
         let start_time = Instant::now();
 
-        let readable_app = app.read().await;
+        let app_lock = app.read().await;
+        let Some(readable_app) = &*app_lock else {
+            tracing::debug!("App not found");
```
phillipleblanc:
```suggestion
```

phillipleblanc (dismissed):

y-f-u (approved):

Issue #764, opened 2024-03-05
Failed to add OpenTelemetry data (permission denied)


lukekim:
**Describe the bug**

Received error message:

```bash
2024-03-05T22:55:34.291242Z ERROR runtime::opentelemetry: Failed to add OpenTelemetry data to backend: Unable to publish data to SpiceAI: Unable to query: status: PermissionDenied, message: "Permission denied", details: [], metadata: MetadataMap { headers: {"date": "Tue, 05 Mar 2024 22:55:34 GMT", "content-type": "application/grpc", "strict-transport-security": "max-age=15724800; includeSubDomains", "alt-svc": "h3=\":443\"; ma=86400", "cf-ray": "85fdb1bec98127ad-SEA", "cf-cache-status": "DYNAMIC", "server": "cloudflare"} }
```

**To Reproduce**

```bash
spice init demo
spice add lukekim/demo
spice add lukekim/smart
spice run

# wait 30 seconds, error appears
```

**Expected behavior**

No error.

**Screenshots**

<img width="976" alt="image" src="https://github.com/spiceai/spiceai/assets/80174/4faa5267-3c6d-4043-a0e6-9dcb76abe872">

phillipleblanc:
Closed in #771

PR #771, opened 2024-03-07
Add a better error message when publishing fails to SpiceAI due to a PermissionDenied error

phillipleblanc:
Fixes the bad error message in #764

Previous Error:

`2024-03-07T00:49:29.389197Z ERROR runtime::opentelemetry: Failed to add OpenTelemetry data to backend: Unable to publish data to SpiceAI: Unable to query: status: PermissionDenied, message: "Permission denied", details: [], metadata: MetadataMap { headers: {"date": "Thu, 07 Mar 2024 00:49:29 GMT", "content-type": "application/grpc", "strict-transport-security": "max-age=15724800; includeSubDomains", "alt-svc": "h3=\":443\"; ma=86400", "cf-ray": "860695fe4b8b933d-ICN", "cf-cache-status": "DYNAMIC", "server": "cloudflare"} }`

New Error:

`2024-03-07T01:05:10.914499Z ERROR runtime::dataconnector::spiceai: Permission denied when publishing data to Spice AI at 'phillipleblanc.demo.drive_stats' - does the API Key ('X') match the app that owns the dataset?`


crates/runtime/src/dataconnector/spiceai.rs
```diff
@@ -159,13 +161,18 @@ impl DataPublisher for SpiceAI {
...
         Box::pin(async move {
             tracing::debug!("Adding data to {dataset_path}");
 
-            client
-                .publish(&dataset_path, data_update.data)
-                .await
-                .context(UnableToPublishDataSnafu)?;
+            if let Err(e) = client.publish(&dataset_path, data_update.data).await {
+                if let flight_client::Error::PermissionDenied {} = e {
+                    tracing::error!("Permission denied when publishing data to Spice AI at '{dataset_path}' - does the API Key ('{api_key}') match the app that owns the dataset?");
```
mitchdevenport, line 170: It's not safe to print the API key in the logs, right?

mitchdevenport (approved):

crates/runtime/src/dataconnector/spiceai.rs
phillipleblanc: The increased observability outweighs the downside here. Otherwise people might think "I do have permissions to this dataset" without realizing that the API key the runtime is configured to use is the thing that needs to match, which leads to the confusion in the linked bug.

PR #772, opened 2024-03-07
Properly extract postgres Numeric values into Arrow Decimal128

mitchdevenport:
This PR makes the necessary changes to properly extract postgres Numeric values into Arrow Decimal128 columns.


crates/arrow_sql_gen/src/postgres.rs
```diff
@@ -219,7 +235,7 @@ fn map_column_type_to_data_type(column_type: &Type) -> DataType {
         Type::TEXT => DataType::Utf8,
         Type::BOOL => DataType::Boolean,
         // TODO: Figure out how to handle decimal scale and precision, it isn't specified as a type in postgres types
-        Type::NUMERIC => DataType::Decimal128(38, 10),
+        Type::NUMERIC => DataType::Decimal128(38, 9),
```
phillipleblanc, line 238: Surely there is a way to get this? It's part of the Numeric type.

crates/arrow_sql_gen/src/postgres.rs
```diff
@@ -235,8 +251,95 @@ fn map_column_type_to_array_builder(column_type: &Type) -> Box<dyn ArrayBuilder>
...
+    scale: u16,
+}
+
+impl BigDecimalFromSql {
+    fn to_decimal_128(&self) -> Option<i128> {
+        (&self.inner * 10i128.pow(u32::from(self.scale))).to_i128()
+    }
+}
+
+impl<'a> FromSql<'a> for BigDecimalFromSql {
```
phillipleblanc, line 276: Was this code taken from somewhere? I would prefer to rely on a library that does this for us rather than having it in-house. Or if we do keep it, we should add unit tests to cover it.

mitchdevenport: Nope, I had to dig in and learn about how Postgres represents numeric values myself. I searched pretty far and wide for an out of the box solution and it doesn't exist.

`rust_decimal` implements `FromSql` but it only supports up to 96 bits, not 128 which is what we need for Decimal128: https://docs.rs/crate/rust_decimal/latest

The only implementation I could find of `FromSql` for `BigDecimal` is in Diesel, but to use it we'd have to use Diesel as an ORM throughout our code which would require knowing schemas at compile time, which we don't: https://docs.rs/diesel/latest/diesel/

crates/arrow_sql_gen/src/postgres.rs
mitchdevenport: Unfortunately it's not exposed in postgres `Row` :( The only solution I've found thus far is to query the information schema ourselves, but I'm hoping to find a better solution.

mitchdevenport: We can at least get the `scale` be reusing some of the `FromSql` logic below.

crates/arrow_sql_gen/src/postgres.rs
mitchdevenport: Added 2 basic tests below, need more to cover edge cases. Will add after livesite.

phillipleblanc (approved):

PR #775, opened 2024-03-07
use postgres async queries

y-f-u:
- this remove the block in the postgres query
- also add the capability for future expansion for both async and sync based queries.

### DIFF
--- a/crates/sql_provider_datafusion/src/dbconnection.rs
+++ b/crates/sql_provider_datafusion/src/dbconnection.rs
@@ -3,12 +3,25 @@ use std::{any::Any, sync::Arc};
 use datafusion::{
     arrow::datatypes::SchemaRef, execution::SendableRecordBatchStream, sql::TableReference,
 };
+use snafu::prelude::*;
 
 pub mod duckdbconn;
 pub mod postgresconn;
 
-pub type Error = Box<dyn std::error::Error + Send + Sync>;
-type Result<T, E = Error> = std::result::Result<T, E>;
+pub type GenericError = Box<dyn std::error::Error + Send + Sync>;
+type Result<T, E = GenericError> = std::result::Result<T, E>;
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Unable to downcast connection"))]
+    UnableToDowncastConnection {},
+
+    #[snafu(display("Unable to get schema: {source}"))]
+    UnableToGetSchema { source: GenericError },
+
+    #[snafu(display("Unable to query arrow: {source}"))]
+    UnableToQueryArrow { source: GenericError },
+}
 
 pub trait SyncDbConnection<T, P>: DbConnection<T, P> {
     fn new(conn: T) -> Self
@@ -46,11 +59,14 @@ pub async fn get_schema<T, P>(
     table_reference: &datafusion::sql::TableReference<'_>,
 ) -> Result<Arc<arrow::datatypes::Schema>, Error> {
     let schema = if let Some(conn) = conn.as_sync() {
-        conn.get_schema(table_reference)?
+        conn.get_schema(table_reference)
+            .context(UnableToGetSchemaSnafu)?
     } else if let Some(conn) = conn.as_async() {
-        conn.get_schema(table_reference).await?
+        conn.get_schema(table_reference)
+            .await
+            .context(UnableToGetSchemaSnafu)?
     } else {
-        return Err(Error::from("Unable to downcast connection"));
+        return Err(Error::UnableToDowncastConnection {});
     };
     Ok(schema)
 }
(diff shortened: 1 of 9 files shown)


crates/sql_provider_datafusion/src/dbconnection/postgresconn.rs
```diff
@@ -31,57 +31,67 @@ pub struct PostgresConnection {
...
     ) -> Result<SendableRecordBatchStream> {
-        let rows = tokio::task::block_in_place(move || {
-            Handle::current().block_on(async { self.conn.query(sql, params).await })
-        })
-        .context(QuerySnafu)?;
+        let rows = self.conn.query(sql, params).await.context(QuerySnafu)?;
         let rec = rows_to_arrow(rows.as_slice()).context(ConversionSnafu)?;
         let schema = rec.schema();
         let recs = vec![rec];
         Ok(Box::pin(MemoryStream::try_new(recs, schema, None)?))
```
y-f-u, line 91: this still needs to be modified to use a stream. will address in future PR.

crates/sql_provider_datafusion/src/dbconnection.rs
```diff
@@ -10,14 +10,33 @@ pub mod postgresconn;
...
+    async fn get_schema(&self, table_reference: &TableReference) -> Result<SchemaRef>;
+    async fn query_arrow(&self, sql: &str, params: &[P]) -> Result<SendableRecordBatchStream>;
+    async fn execute(&self, sql: &str, params: &[P]) -> Result<u64>;
+}
+
+pub trait DbConnection<T, P>: Send {
     fn as_any(&self) -> &dyn Any;
     fn as_any_mut(&mut self) -> &mut dyn Any;
+
+    fn as_sync(&self) -> Option<&dyn SyncDbConnection<T, P>> {
```
phillipleblanc, line 49: This feels very clunky. Why not just make the trait always async? Isn't it possible to run sync code in an async context?

y-f-u: Tried the approach. The issue is the duckdb rs is synced and all struct not implemented sync properly.

y-f-u: I couldn't find a way that makes the synchronised pool + duckdb inner structs thread safe. The duckdb connection is implemented in a synchronised way, e.g. InnerConnection. I don't think there is way to patch it.

crates/sql_provider_datafusion/src/lib.rs
```diff
@@ -53,9 +61,18 @@ impl<T, P> SqlTable<T, P> {
...
+        let schema = if let Some(conn) = conn.as_sync() {
+            conn.get_schema(&table_reference)
+                .context(UnableToQueryDbConnectionSnafu)?
+        } else if let Some(conn) = conn.as_async() {
+            conn.get_schema(&table_reference)
+                .await
+                .context(UnableToQueryDbConnectionSnafu)?
+        } else {
+            return Err(Error::UnableToDowncastConnection);
+        };
```
phillipleblanc: Instead of requiring the callers to have to do this - why don't we add a default function on the DbConnection trait that basically does this? Then at least from the caller's perspective, it always looks async.

crates/sql_provider_datafusion/src/dbconnection.rs
phillipleblanc: See my other comment

crates/sql_provider_datafusion/src/dbconnection.rs
```diff
@@ -10,14 +10,60 @@ pub mod postgresconn;
...
+        None
+    }
+    fn as_async(&self) -> Option<&dyn AsyncDbConnection<T, P>> {
+        None
+    }
+}
+
+pub async fn get_schema<T, P>(
```
y-f-u, line 57: putting this into DbConnnection result in a cycle dependency between DbConnection and the Async/Sync.

It's not the best, but still hiding the details from the caller.

crates/sql_provider_datafusion/src/dbconnection.rs
```diff
@@ -10,14 +10,60 @@ pub mod postgresconn;
...
+    conn: Box<dyn DbConnection<T, P>>,
+    table_reference: &datafusion::sql::TableReference<'_>,
+) -> Result<Arc<arrow::datatypes::Schema>, Error> {
+    let schema = if let Some(conn) = conn.as_sync() {
+        conn.get_schema(table_reference)?
+    } else if let Some(conn) = conn.as_async() {
+        conn.get_schema(table_reference).await?
+    } else {
+        return Err(Error::from("Unable to downcast connection"));
```
phillipleblanc: This needs a specific Snafu error enum (same for below)

crates/sql_provider_datafusion/src/dbconnection/postgresconn.rs
```diff
@@ -31,57 +31,67 @@ pub struct PostgresConnection {
...
+    > for PostgresConnection
+{
+    fn as_any(&self) -> &dyn Any {
+        self
+    }
+
+    fn as_any_mut(&mut self) -> &mut dyn Any {
+        self
+    }
+    fn as_async(
```
phillipleblanc:
```suggestion
    }

    fn as_async(
```

crates/sql_provider_datafusion/src/dbconnection/postgresconn.rs
```diff
@@ -31,57 +31,67 @@ pub struct PostgresConnection {
...
-
-    fn as_any(&self) -> &dyn Any {
-        self
-    }
-
-    fn as_any_mut(&mut self) -> &mut dyn Any {
-        self
+    async fn execute(&self, sql: &str, params: &[&'a (dyn ToSql + Sync)]) -> Result<u64> {
+        let result = self.conn.execute(sql, params).await;
+        Ok(result?)
```
phillipleblanc:
```suggestion
        self.conn.execute(sql, params).await?;
```

phillipleblanc (approved):

PR #786, opened 2024-03-08
Set decimal scale and precision from query result

mitchdevenport:
This PR properly sets the scale and precision of arrow Decimal128 values read from a Postgres row.


mitchdevenport:
LGTM, but can't approve since I'm the original author

lukekim (approved):

PR #788, opened 2024-03-08
split record batch into chunks up to 2048 per chunk due to duckdb downstream limitation

y-f-u:
when writing to duckdb backend, only 2048 maximum is allowed to be in the vector. That could be some downstream limitation we can investigate further to unblock this.

### DIFF
--- a/crates/runtime/src/databackend/duckdb.rs
+++ b/crates/runtime/src/databackend/duckdb.rs
@@ -1,4 +1,5 @@
 use std::{
+    cmp,
     collections::HashMap,
     fmt, mem,
     sync::{Arc, PoisonError},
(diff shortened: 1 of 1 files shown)


crates/runtime/src/databackend/duckdb.rs
```diff
@@ -230,6 +242,15 @@ impl<'a> DuckDBUpdate<'a> {
         Ok(())
     }
 
+    fn split_batch(batch: &RecordBatch) -> Vec<RecordBatch> {
+        let mut result = vec![];
+        for offset in (0..=batch.num_rows()).step_by(2048) {
+            let length = cmp::min(2048, batch.num_rows() - offset);
```
ewgenius: maybe worth moving to a constant?

ewgenius (dismissed):

phillipleblanc (approved):

ewgenius (approved):

PR #790, opened 2024-03-11
Add keyring secret store

ewgenius:
Added support for keyring secret store, based on top of https://docs.rs/keyring

This store is optional, and can be enabled by `keyring-secret-store` cargo feature.

To store secret requires a keychain entry named `spiced_secret_<secret_name>` with a key-value map stored as a json string and `spiced` account.

CleanShot 2024-03-11 at 12 14 00@2x


crates/runtime/Cargo.toml
```diff
@@ -52,6 +52,7 @@ arrow_sql_gen = { path = "../arrow_sql_gen", optional = true }
 bb8 = {workspace = true, optional = true}
 bb8-postgres = {workspace = true, optional = true}
 base64 = "0.22.0"
+keyring = "2.3.2"
```
phillipleblanc: Can we make this an optional feature?

phillipleblanc (dismissed):

crates/runtime/Cargo.toml
ewgenius: @phillipleblanc The keyring is a cross-platform library, so I'm not sure if it's feasible to make it optional based on different criteria

phillipleblanc: DuckDB is a cross-platform library, and I made that optional?

ewgenius: Added `keyring-secret-store` feature to runtime

phillipleblanc (approved):

PR #792, opened 2024-03-11
Basic appropriate metrics for SpiceOSS

Jeadie:
## Changes
 - Latency calls for 
     - `do_get`
     - `/v1/sql` 
     - `load_dataset`, per dataset
- For DuckDB updates
    - memory usage
    - memory limit
    - wal size
    - database size

## Spec
Metrics we want: 
- SQL duration
- For data backends
    - Database size
    - WAL Size (if appropriate)
    - Memory usage
    - Latest time latency (i.e. time between getting new data, and it being queryable in the data backend). 
- Per dataset:
        - Register dataset latency
        - Update dataset


crates/runtime/src/databackend/duckdb.rs
```diff
@@ -168,6 +171,7 @@ impl<'a> DuckDBUpdate<'a> {
             self.insert_batch(batch)?;
         }
 
+        self.monitor_duckdb()?;
```
Jeadie, line 174: I don't think this is the best way to, on a frequent schedule, to get metrics about the specific backend

phillipleblanc: No, this should be on a timer that only gets started if the metrics endpoint is enabled.

phillipleblanc: We may want to extend the trait definition to include a function that collects metrics that the runtime calls in the timer loop for each backend.

crates/runtime/src/timing.rs
```diff
@@ -0,0 +1,75 @@
...
+            start: Instant::now(),
+            metric_name,
+            labels
+        }
+    }
+}
+
+impl Drop for TimingGuard {
+    fn drop(&mut self) {
+        metrics::histogram!(self.metric_name, &self.labels).record(self.start.elapsed().as_secs_f64());
```
phillipleblanc, line 73: This should be in milliseconds?

crates/runtime/src/databackend/duckdb.rs
```diff
@@ -246,6 +250,38 @@ impl<'a> DuckDBUpdate<'a> {
...
             .unwrap_or(false)
     }
+
+
+    fn monitor_duckdb(&self) -> Result<()> {
+        self.duckdb_conn.conn.query_row("select database_size, wal_size, memory_usage, memory_limit FROM pragma_database_size();", [], |row| {
+            
+            if let Ok(database_size) = row.get::<usize, String>(0) {
+                if let Ok(bytes) = pg_size_bytes(database_size.as_str()) {
+                    metrics::gauge!("database_size").set(bytes);
```
phillipleblanc: These metrics should be prefixed with `duckdb_`

Jeadie:
https://github.com/spiceai/spiceai/pull/794 suffices for now

PR #793, opened 2024-03-11
Support postgres backend & data connector server, port, username, password, secret config

Sevenannn:
## What
Support reading and using the value of postgres server, port, username, password, secret config specified in spicepod.yaml for both postgres backend and postgres data connector

Example of spicepod.yaml:
```yaml
datasets:
- from: spice.ai/eth.beacon.recent_slots
  name: eth_beacon_recent_slots
  params: # Parameters for data connector
    pg_host: localhost
    pg_port: "5432"
    pg_db: postgres
    pg_user: postgres
    pg_pass_key: my_password_key # Key to postgres password in secret store data. If presented, use the key to find postgres password secret
  acceleration:
    refresh_mode: full
    engine: postgres
    engine_secret: backend_secret_key # Secret key points to the secret data for postgres backend
    params: # Parameters for postgres backend
       pg_host: localhost
       pg_port: "5432"
       pg_db: postgres
       pg_user: postgres
       pg_pass_key: my_password_key # Key to postgres password in secret store data. If presented, use the key to find postgres password secret
```

## How
* Assumptions made on secrets and backend params: https://github.com/orgs/spicehq/projects/23/views/4?pane=issue&itemId=56241757


crates/sql_provider_datafusion/src/dbconnectionpool/postgrespool.rs
```diff
@@ -42,7 +42,30 @@ impl
...
+            if let Some(pg_host) = params.get("pg_host") {
+                host = pg_host;
+            }
+            if let Some(pg_user) = params.get("pg_user") {
+                user = pg_user;
+            }
+            if let Some(pg_db) = params.get("pg_db") {
+                dbname = pg_db;
+            }
+            if let Some(pg_pass) = params.get("pg_pass") {
```
mitchdevenport: We need to support @ewgenius work on secrets for providing the password

Sevenannn: Thanks, will update the changes

crates/runtime/src/databackend.rs
```diff
@@ -69,6 +72,12 @@ impl DataBackendBuilder {
         self
     }
 
+    #[must_use]
+    pub fn secrets(mut self, secrets: Option<Secret>) -> Self {
```
ewgenius: Let's name it `secret` to avoid confusion. The secret is a hashmap item with multiple fields.

crates/runtime/src/databackend/postgres.rs
```diff
@@ -142,6 +144,24 @@ impl PostgresBackend {
...
+    if let Some(params_val) = params.as_ref() {
+        let mut new_params = params_val.clone();
+        if let Some(secrets) = secrets {
+            if let Some(pg_pass_val) = new_params.get("pg_pass") {
+                if let Some(pg_pass_secret) = secrets.get(pg_pass_val) {
+                    new_params.insert("pg_pass".to_string(), pg_pass_secret.to_string());
+                    return Arc::new(Some(new_params));
+                }
+            }
+        }
```
ewgenius: would be better to make early returns

```rust

let Some(secrets) = secrets else {
  params
}

let Some(pg_pass_val) = new_params.get("pg_pass") else {
  params
}

...

```

crates/runtime/src/databackend/postgres.rs
```diff
@@ -142,6 +144,24 @@ impl PostgresBackend {
     }
 }
 
+fn read_pg_config(
```
mitchdevenport: This isn't well named, we're retrieving the password, not reading a config

crates/runtime/src/databackend/postgres.rs
```diff
@@ -142,6 +144,24 @@ impl PostgresBackend {
...
+    params: Arc<Option<HashMap<String, String>>>,
+    secrets: Option<Secret>,
+) -> Arc<Option<HashMap<String, String>>> {
+    if let Some(params_val) = params.as_ref() {
+        let mut new_params = params_val.clone();
+        if let Some(secrets) = secrets {
+            if let Some(pg_pass_val) = new_params.get("pg_pass") {
+                if let Some(pg_pass_secret) = secrets.get(pg_pass_val) {
```
mitchdevenport: If the secret doesn't exist, we should default to assuming we've been provided a raw password

crates/sql_provider_datafusion/src/dbconnectionpool/postgrespool.rs
```diff
@@ -42,7 +42,30 @@ impl
         _mode: Mode,
         _params: Arc<Option<HashMap<String, String>>>,
```
mitchdevenport: Rename to `params` now that it's being used. The `_` prefix is only used to indicate that an argument is intentionally unused.

crates/runtime/src/databackend/postgres.rs
mitchdevenport: Oh I see, we're getting the params. Maybe `read_pg_params` would be better

mitchdevenport: Actually, I think this name is fine now.

crates/runtime/src/databackend/postgres.rs
mitchdevenport: I misunderstood the code, we're already doing this

crates/runtime/src/databackend/postgres.rs
Sevenannn: Make early returns in the new update

crates/runtime/src/databackend.rs
Sevenannn: Name updated to secret in the recent updates

crates/sql_provider_datafusion/src/dbconnectionpool/postgrespool.rs
Sevenannn: Renamed to params in the new commits

crates/runtime/src/databackend/postgres.rs
```diff
@@ -141,6 +143,27 @@ impl PostgresBackend {
...
 }
 
+fn read_pg_config(
+    params: Arc<Option<HashMap<String, String>>>,
+    secrets: Option<Secret>,
+) -> Arc<Option<HashMap<String, String>>> {
+    let Some(params_val) = params.as_ref() else {
+        return params;
+    };
+    let mut new_params = params_val.clone();
```
mitchdevenport: nit: don't instantiate the variable until it's actually needed below

crates/runtime/src/databackend/postgres.rs
```diff
@@ -141,6 +143,27 @@ impl PostgresBackend {
...
+    let Some(secrets) = secrets else {
+        return params;
+    };
+    let Some(pg_pass_val) = new_params.get("pg_pass_key") else {
```
mitchdevenport: Just use params here, we don't need to use new_params until we're actually modifying the data

crates/spicepod/src/component/dataset.rs
```diff
@@ -131,6 +131,28 @@ impl Dataset {
...
+            }
+        }
+
+        None
+    }
+
+    #[must_use]
+    pub fn engine_secret(&self) -> Option<String> {
+        if let Some(acceleration) = &self.acceleration {
+            if let Some(engine_secret) = &acceleration.engine_secret {
```
mitchdevenport: You can just return `acceleration.engine_secret.clone()` instead of unwrapping it and then rewrapping it in an optional.

crates/spicepod/src/component/dataset.rs
```diff
@@ -131,6 +131,28 @@ impl Dataset {
         }
     }
 
+    #[must_use]
+    pub fn acceleration_params(&self) -> Option<HashMap<String, String>> {
+        if let Some(acceleration) = &self.acceleration {
+            if let Some(refresh_interval) = &acceleration.params {
```
mitchdevenport: You can just return `acceleration.params.clone()` instead of unwrapping it and then rewrapping it in an optional.

crates/runtime/src/databackend/postgres.rs
```diff
@@ -1,5 +1,6 @@
 use std::{collections::HashMap, mem, sync::Arc};
 
+use crate::secrets::Secret;
```
mitchdevenport: You need to do a similar set up for the postgres data connector in `crates/runtime/src/dataconnector/postgres.rs`

crates/runtime/src/databackend/postgres.rs
```diff
@@ -141,6 +143,27 @@ impl PostgresBackend {
     }
 }
 
+fn read_pg_config(
```
mitchdevenport: Let's minimize duplication of this logic between the connector and backend as much as possible

crates/spicepod/src/component/dataset.rs
Sevenannn: Thanks, I updated this part and will push the changes in future commit

Sevenannn:
The most recent commit doesn’t really make sense, will update in later commits.

crates/runtime/src/databackend/postgres.rs
Sevenannn: Move the read_pg_config to datafusion.rs, and it is shared by both data backend and data connector

crates/runtime/src/databackend/postgres.rs
Sevenannn: Update the support for data connector

crates/runtime/src/datafusion.rs
```diff
@@ -122,6 +123,7 @@ impl DataFusion {
     pub async fn new_accelerated_backend(
         &self,
         dataset: impl Borrow<Dataset>,
+        secrets_provider: Arc<RwLock<secrets::SecretsProvider>>,
```
ewgenius, line 125: nit: I think we can read backend secret before calling `new_accelerated_backend`, and pass it down instead of whole secrets provider

Sevenannn: Updated, thanks!

crates/runtime/src/datafusion.rs
```diff
@@ -408,3 +415,28 @@ impl Default for DataFusion {
         Self::new()
     }
 }
+
+// TODO: This is a temporary solution since we cannot import secret and define this function in connection pool due to cyclical dependency
+// Ideally we want separate crate for secret and define this function in connection pool
+#[must_use]
+#[allow(clippy::implicit_hasher)]
+pub fn read_pg_config(
```
mitchdevenport: This really doesn't belong here, we need to do this the right way.

I've done the secrets refactor, I'll leave the db_connection_pool crate to you.

https://github.com/spiceai/spiceai/pull/803

Sevenannn (commented):
Reviewed changes made by Mitch. Look good to me, genius changes to remove the cyclical dependencies!

mitchdevenport (commented):
Approving Qianqian's portion of the changes

mitchdevenport (approved):

PR #794, opened 2024-03-12
Add basic timing metrics to SpiceOSS.

Jeadie:
## Changes
 - Endpoint latency metrics
 - Load latency {models, datasets, secrets}
 - Number of components active in the spicepod


crates/runtime/src/lib.rs
```diff
@@ -210,8 +214,15 @@ impl Runtime {
...
-
                 tracing::info!("Loaded dataset: {}", &ds.name);
+                let engine_str = ds.acceleration.clone().map(|acc| {
+                    if acc.enabled {
+                        acc.engine().clone().to_string()
+                    } else {
+                        "None".to_string()
+                    }
+                });
+                metrics::gauge!("datasets/count", "engine" => engine_str.unwrap_or("None".to_string())).increment(1.0);
```
phillipleblanc: Does this work?
```suggestion
                let engine_str = ds.acceleration.as_ref().map_or_else(
                    || "None", // Default value for the None case
                    |acc| if acc.enabled {
                        acc.engine().to_string().as_str() // Transform if Some and enabled
                    } else {
                        "None" // Return "None" if not enabled
                    }
                );
                metrics::gauge!("datasets/count", "engine" => engine_str.unwrap_or("None".to_string())).increment(1.0);
```

crates/runtime/src/lib.rs
```diff
@@ -228,6 +239,15 @@ impl Runtime {
         }
 
         tracing::info!("Unloaded dataset: {}", &ds.name);
+        let engine_str = ds.acceleration.clone().map(|acc| {
```
phillipleblanc: Use `Option::map_or_else` as noted above to reduce cloning

crates/runtime/src/timing.rs
```diff
@@ -0,0 +1,77 @@
...
+            metric_name,
+            labels,
+        }
+    }
+}
+
+impl Drop for TimeMeasurement {
+    fn drop(&mut self) {
+        metrics::histogram!(self.metric_name, &self.labels)
+            .record(self.start.elapsed().as_secs_f64());
```
phillipleblanc: Can we track milliseconds instead? And if so, I would rename the macro to be `measure_scope_ms`

crates/runtime/src/flight.rs
```diff
@@ -167,6 +168,7 @@ impl FlightService for Service {
         &self,
         request: Request<Streaming<FlightData>>,
     ) -> Result<Response<Self::DoPutStream>, Status> {
+        measure_scope!("flight", "operation" => "do_put");
```
phillipleblanc: What does this actually measure? This code returns a stream, so I don't think this is measuring what we think it will measure?

crates/runtime/src/timing.rs
Jeadie: From the docs: 
> The returned value does include the fractional (nanosecond) part of the duration.

So this includes enough sig figs, but also happy to convert to miliseconds

crates/runtime/src/lib.rs
Jeadie: I had issue with borrowing temporary value.

crates/runtime/src/flight.rs
Jeadie: You're right, this will complete recording once its dropped out of scope (i.e. after the streams are initialised, but (at least potentially) before the streams are read).

crates/runtime/src/timing.rs
```diff
@@ -0,0 +1,77 @@
...
+            metric_name,
+            labels,
+        }
+    }
+}
+
+impl Drop for TimeMeasurement {
+    fn drop(&mut self) {
+        metrics::histogram!(self.metric_name, &self.labels)
+            .record(10000_f64 * self.start.elapsed().as_secs_f64());
```
phillipleblanc:
```suggestion
            .record(1000_f64 * self.start.elapsed().as_secs_f64());
```

Shouldn't this be 1000?

phillipleblanc (approved):

PR #795, opened 2024-03-12
Implement Flight SQL server

phillipleblanc:
Implements the Flight SQL commands on top of our Flight server. Refactors the code so that the flight implementation isn't all in one big file.


crates/runtime/src/flight/flightsql/get_tables.rs
```diff
@@ -0,0 +1,86 @@
...
+    Ok(Response::new(
+        Box::pin(record_batches_to_flight_stream(vec![record_batch]))
+            as <Service as FlightService>::DoGetStream,
+    ))
+}
+
+fn table_type_name(table_type: TableType) -> &'static str {
+    match table_type {
+        // from https://github.com/apache/arrow-datafusion/blob/26b8377b0690916deacf401097d688699026b8fb/datafusion/core/src/catalog/information_schema.rs#L284-L288
+        TableType::Base => "BASE TABLE",
```
Jeadie, line 82: I assume we're explicitly not including `SYSTEM TABLE` (as suggested in the .proto) file (sample and link below).
```proto
  /*
   * Specifies a filter of table types which must match.
   * The table types depend on vendor/implementation. It is usually used to separate tables from views or system tables.
   * TABLE, VIEW, and SYSTEM TABLE are commonly supported.
   */
  repeated string table_types = 4;
```

source: https://github.com/apache/arrow/blob/e268b310639b98ccba0d00dbda04451e645e86b7/format/FlightSql.proto#L1235

phillipleblanc: Correct. If that becomes a problem we can add it later

crates/runtime/src/flight/get_flight_info.rs
```diff
@@ -0,0 +1,52 @@
...
+use prost::Message;
+use tonic::{Request, Response, Status};
+
+use super::{flightsql, to_tonic_err, Service};
+
+pub(crate) async fn handle(
+    flight_svc: &Service,
+    request: Request<FlightDescriptor>,
+) -> Result<Response<FlightInfo>, Status> {
+    let Ok(message) = Any::decode(&*request.get_ref().cmd) else {
```
Jeadie, line 14: Is the desired behaviour, if we don't get any valid/known command, return a "raw" flight info response?

phillipleblanc: Yeah, this is how we need to maintain compatibility with our current protocol.

crates/runtime/src/flight/actions.rs
```diff
@@ -0,0 +1,100 @@
...
+                })?;
+            let stmt =
+                prepared_statement_query::do_action_create_prepared_statement(flight_svc, cmd)
+                    .await?;
+            let output = futures::stream::iter(vec![Ok(arrow_flight::Result {
+                body: stmt.as_any().encode_to_vec().into(),
+            })]);
+            Ok(Response::new(Box::pin(output)))
+        }
+        ActionType::ClosePreparedStatement => {
```
Jeadie, line 94: This is essentially a noop because we don't instantiate any state for the `CreatePreparedStatement`?

phillipleblanc: Yep

crates/runtime/src/flight/flightsql/get_schemas.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    Response::new(info)
+}
+
+pub(crate) async fn do_get(
+    flight_svc: &Service,
+    query: sql::CommandGetDbSchemas,
+) -> Result<Response<<Service as FlightService>::DoGetStream>, Status> {
+    let catalog = &query.catalog;
+    tracing::trace!("do_get: {query:?}");
```
Jeadie, line 35: Is tracing like this able to delineate a `do_get` for a `CommandGetDbSchemas` against all other types of `do_get`s?

phillipleblanc: Yeah, the module name is part of the output.

mitchdevenport (approved):

PR #796, opened 2024-03-12
Support for Postgres as a Data Connector and Mesh

mitchdevenport:
This change adds support to use Postgres as a Data Connector and Data Mesh.

### DIFF
--- a/crates/runtime/src/dataconnector/postgres.rs
+++ b/crates/runtime/src/dataconnector/postgres.rs
@@ -1,0 +1,121 @@
+use async_trait::async_trait;
+use bb8_postgres::tokio_postgres::types::ToSql;
+use bb8_postgres::tokio_postgres::NoTls;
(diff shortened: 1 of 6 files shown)


crates/runtime/src/dataconnector/postgres.rs
```diff
@@ -0,0 +1,121 @@
...
+                &'static (dyn ToSql + Sync),
+            > + Send
+            + Sync,
+    >,
+}
+
+#[async_trait]
+impl DataConnector for Postgres {
+    fn new(
+        _secret: Option<Secret>,
```
y-f-u, line 38: assuming this is going to be fixed once the store is integrated?

mitchdevenport: Yes it will be used there. But it's part of the trait definition so we have to include it even if we aren't using it.

y-f-u (approved):

PR #801, opened 2024-03-13
Basic FlightSQL data connector

Jeadie:
## Changes
- Add a new flightSQL data connector to `spiceai`.
- New crate, `flightsql_datafusion`, a Datafusion `TableProvider` for FlightSQL compatible servers.
- Example spicepod dataset
```yaml
datasets:
  - from: flightsql:datafusion.public.eth_recent_blocks
    name: blocks
    params:
      endpoint: http://127.0.0.1:50051
```


Jeadie:
Still need to:
- Create `TableProvider` for FlightSQL
- Test (against Phillips Spice FilghtSQL).

crates/flight_client/src/tls.rs
```diff
@@ -16,6 +16,11 @@ pub enum Error {
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
+/// # Errors
```
Jeadie, line 19: Needed for Clippy linting.

crates/flight_client/src/lib.rs
```diff
@@ -19,7 +19,7 @@ use tonic::transport::Channel;
 use tonic::IntoRequest;
 use tonic::IntoStreamingRequest;
 
-mod tls;
```
Jeadie, line 22: Need `tls::new_tls_flight_channel` to make FlightSQL channel

phillipleblanc (dismissed):

crates/runtime/src/flight.rs
```diff
@@ -125,6 +125,20 @@ impl FlightService for Service {
...
+            .read()
+            .await
+            .ctx
+            .sql(&sql)
+            .await
+            .map_err(to_tonic_err)?;
+        Ok(df.schema().into())
+    }
+
     async fn get_arrow_schema_and_size_sql(
```
phillipleblanc: Let's remove this function

crates/flightsql_datafusion/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+
+    pub async fn get_schema(
+        mut client: FlightSqlServiceClient<Channel>,
+        table_reference: OwnedTableReference,
+    ) -> Result<SchemaRef> {
+        let flight_info = client
+            .get_tables(CommandGetTables {
+                catalog: table_reference
+                    .catalog()
+                    .map(std::string::ToString::to_string),
```
phillipleblanc:
```suggestion
                    .map(ToString::to_string),
```

crates/flightsql_datafusion/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+                db_schema_filter_pattern: table_reference
+                    .schema()
+                    .map(std::string::ToString::to_string),
```
phillipleblanc:
```suggestion
                    .map(ToString::to_string),
```

crates/flightsql_datafusion/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+                {
+                    return Some(table_schema.value(i));
+                }
+            }
+            None
+        });
+
+        match schema_bytz_opt {
+            Some(schema_bytz) => {
+                match Schema::try_from(IpcMessage(tonic::codegen::Bytes::copy_from_slice(
```
phillipleblanc:
```suggestion
                match Schema::try_from(IpcMessage(Bytes::copy_from_slice(
```

crates/flightsql_datafusion/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+                    Self::get_str_from_record_batch(b, i, "catalog_name").unwrap_or_default();
+                let db_schema_name =
+                    Self::get_str_from_record_batch(b, i, "db_schema_name").unwrap_or_default();
+
+                // We got at least one None above, so we can't compare.
+                if table_name.is_empty() || catalog_name.is_empty() || db_schema_name.is_empty() {
+                    continue;
+                }
+
+                // Match even if catalog_name or db_schema_name from table_reference was None.
```
phillipleblanc: This code won't return the table_schema if the table_reference catalog_name or db_schema_name is None. Did you intend to implement this?

crates/flightsql_datafusion/src/lib.rs
```diff
@@ -0,0 +1,465 @@
...
+    mut client: FlightSqlServiceClient<Channel>,
+    query: String,
+) -> Result<Option<FlightRecordBatchStream>, FlightSQLError> {
+    let flight_info = client
+        .clone()
+        .execute(query, None)
+        .await
+        .map_err(|e| FlightSQLError::ArrowFlight { source: e.into() })?;
+
+    if let Some(ep) = flight_info.endpoint.first() {
```
phillipleblanc: I don't think we should assume that all Flight SQL servers will just have a single endpoint. We only have a single endpoint, but this code will break if connected to a different Flight SQL server that does have multiple.

phillipleblanc: This is implemented in dataconnector/flightsql.rs

crates/runtime/src/dataconnector/flightsql.rs
```diff
@@ -0,0 +1,127 @@
...
+    async fn query(
+        client: FlightSqlServiceClient<Channel>,
+        query: String,
+    ) -> Result<Vec<arrow::record_batch::RecordBatch>, Box<dyn std::error::Error>> {
+        let flight_info = client.clone().execute(query, None).await?;
+
+        let mut batches = vec![];
+        for ep in &flight_info.endpoint {
+            if let Some(tkt) = &ep.ticket {
+                match batch_from_ticket(&mut client.clone(), tkt.to_owned()).await {
```
phillipleblanc: This also needs to handle the case where the URL in endpoint isn't None. If so, we need to connect to that new URL to run the do_get.

phillipleblanc: Also we should unify this logic so we're not repeating it - you have a different implementation of `query` in `crates/flightsql_datafusion/src/lib.rs`.

crates/flightsql_datafusion/src/lib.rs
Jeadie: We will now resolve all possible matches. If >1 match, raise error.

phillipleblanc (approved):

PR #802, opened 2024-03-13
Add `v1/status`.

Jeadie:


crates/runtime/src/http/v1.rs
```diff
@@ -67,6 +66,154 @@ pub(crate) mod query {
...
+    pub struct QueryParams {
+        #[serde(default = "default_format")]
+        format: Format,
+    }
+
+    #[derive(Debug, Serialize, Deserialize)]
+    #[serde(rename_all = "lowercase")]
+    pub enum Format {
+        Json,
+        Yaml,
```
lukekim: Let's scope support to just JSON and CSV.

lukekim (dismissed):

Jeadie:
Not implementing for opentelemetry collector as it doesn't satisfies GRPC [health checks](https://grpc.io/docs/guides/health-checking/)

mitchdevenport (changes requested):
I'm not sure that adding a basic health service for Otel is as much work as you're thinking it is. All we need to do is use the pre-compiled protos provided by tonic, attach the service to the same tonic server, and implement a very basic health check. Since we just want to know that the server is up, it could probably just always return `Ok`. To be clear, no GRPC service definition "satisfies GRPC health checks", servers need to implement that themselves.

Also, it looks like for the Flight endpoint you're just connecting to the GRPC server and considering it healthy if you can connect, is there a reason that is sufficient for the Flight endpoint but not the Otel endpoint?

sgrebnov (dismissed):

ewgenius (approved):

mitchdevenport (approved):

PR #803, opened 2024-03-13
Refactor secrets into its own crate

mitchdevenport:
This PR refactors secrets into its own crate so that it can be used throughout spice crates without causing dependency cycles.


Sevenannn (approved):
lgtm

PR #807, opened 2024-03-14
Add `v1/models`.

Jeadie:


crates/runtime/src/http/v1.rs
```diff
@@ -71,29 +100,63 @@ pub(crate) mod datasets {
...
     use tokio::sync::RwLock;
+    use tract_core::tract_data::itertools::Itertools;
+
+    use crate::datafusion::DataFusion;
+
+    use super::{convert_entry_to_csv, Format};
 
     #[derive(Debug, Deserialize)]
-    pub(crate) struct DatasetFilter {
+    pub(crate) struct QueryParams {
```
phillipleblanc: This was specific to Datasets - let's not reuse it.

crates/runtime/src/http/v1.rs
```diff
@@ -103,11 +166,126 @@ pub(crate) mod datasets {
...
+
+    use crate::model::Model;
+
+    use super::Format;
+
+    #[derive(Debug, Deserialize)]
+    pub(crate) struct QueryParams {
+        // TODO: Implement status checking.
+        // #[serde(default)]
+        // _status: bool,
```
phillipleblanc:
```suggestion
```

ewgenius (approved):

PR #830, opened 2024-03-15
Add metrics for FlightSQL server

phillipleblanc:
Adds metrics to calculate number of requests and request durations for the Flight SQL server.

Also improves the `load_dataset` metric to be accurate.

Adds a `TimedStream` wrapper for a stream that will measure how long it takes to drain the stream and record the elapsed time as a metric.

### DIFF
--- a/crates/runtime/src/flight.rs
+++ b/crates/runtime/src/flight.rs
@@ -1,5 +1,6 @@
 use crate::datafusion::DataFusion;
 use crate::dataupdate::DataUpdate;
+use crate::measure_scope_ms;
 use arrow::array::RecordBatch;
 use arrow::datatypes::Schema;
 use arrow::ipc::writer::{DictionaryTracker, IpcDataGenerator};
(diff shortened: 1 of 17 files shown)


crates/runtime/src/flight.rs
```diff
@@ -66,13 +69,16 @@ impl FlightService for Service {
         &self,
         request: Request<FlightDescriptor>,
     ) -> Result<Response<FlightInfo>, Status> {
+        measure_scope_ms!("flight_get_flight_info_request_duration_ms");
```
Jeadie, line 72: We want `measure_scope_ms` in this and only this function?

Jeadie (approved):

crates/runtime/src/flight.rs
phillipleblanc: Yeah, it doesn't work for any of the functions that return streams because it doesn't actually measure the time it takes to do the work and return the data in the stream, just the time it takes to create the stream object and return that - which is always super quick.

GetFlightInfo is one of the exceptions that just returns an object right away and makes sense to track using the macro.

PR #831, opened 2024-03-15
Sqlite DataBackend

mitchdevenport:
This PR adds support for sqlite as a data backend.

Sqlite can be used as the acceleration data backend for a dataset by specifying the `engine` as `sqlite`:

```
datasets:
    name: eth_recent_blocks
    acceleration:
      engine: sqlite
      mode: memory
```

The `sqlite` engine supports both `memory` and `file` acceleration `mode`.


crates/db_connection_pool/src/sqlitepool.rs
```diff
@@ -0,0 +1,68 @@
...
+        name: &str,
+        mode: Mode,
+        params: Arc<Option<HashMap<String, String>>>,
+    ) -> Result<Self> {
+        let file_name = params
+            .as_ref()
+            .as_ref()
+            .and_then(|params| params.get("sqlite_file").cloned())
+            .unwrap_or(name.to_string());
+        let file_name = format!("{file_name}_sqlite.db");
```
phillipleblanc: Why not just use the file name as is?

phillipleblanc: Ah I see. We should move the _sqlite.db into the unwrap or default

phillipleblanc (dismissed):

crates/db_connection_pool/src/sqlitepool.rs
mitchdevenport: good catch

phillipleblanc (approved):

PR #847, opened 2024-03-18
Opentelemetry health GRPC

Jeadie:
## Changes
- Add [HealthCheck GRPC](https://grpc.io/docs/guides/health-checking/) methods for Opentelemetry collector endpoint.
- Add to `v1/status`.


crates/runtime/Cargo.toml
```diff
@@ -23,6 +23,7 @@ arrow-flight.workspace = true
 arrow-ipc = "50.0.0"
 tonic.workspace = true
 tonic_0_9_0 = { version = "0.9.0", package = "tonic", features = ["gzip"] }
+tonic-health = "0.9.0"
```
lukekim, line 26: Should we be using tonic 0.11?

lukekim (approved):

crates/runtime/Cargo.toml
Jeadie: Wanted to keep it inline with what opentelemetry server is using for tonic (which is 0.9 because of - https://github.com/open-telemetry/opentelemetry-rust/issues/1135#issuecomment-1618239407 (if i understand correctly))

PR #850, opened 2024-03-18
Friendlier DataFusion errors

### LINKED ISSUE #837: Friendlier DataFusion errors
**Describe the bug**
User friendly DataFusion errors (not JSON dumps).

**To Reproduce**

```bash
sql> SELECT *
FROM eth.recent_traces trace
JOIN eth.recent_transactions trans ON trace.transaction_hash = trans.hash
ORDER BY trans.block_number DESC
status: Internal, message: "Plan(\"table 'datafusion.eth.recent_traces' not found\")", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 18 Mar 2024 01:32:56 GMT", "content-length": "0"} }
```

**Expected behavior**
A user friendly DataFusion errors (not JSON dump).

**Screenshots**
image

phillipleblanc:
Closes #837

Passes the actionable DataFusion errors directly as an invalid argument GRPC error. In the REPL, we interpret that to provide a better user experience.

Also adds colors to make the output easier to read in the REPL. Adds a `.error` command to retrieve the full technical details of an error.

Uses FlightSQL to make the request.

Screenshot 2024-03-18 at 6 09 40 PM

### DIFF
--- a/crates/flightrepl/src/lib.rs
+++ b/crates/flightrepl/src/lib.rs
@@ -168,7 +168,7 @@ pub async fn run(repl_config: ReplConfig) -> Result<(), Box<dyn std::error::Erro
 fn display_grpc_error(err: &Status) {
     let user_err_msg = match err.code() {
         Code::Ok => return,
-        Code::Unknown | Code::Internal | Code::Unauthenticated | Code::DataLoss | Code::OutOfRange | Code::FailedPrecondition => 
+        Code::Unknown | Code::Internal | Code::Unauthenticated | Code::DataLoss | Code::OutOfRange | Code::FailedPrecondition =>
             "An internal error occurred while processing the query. Show technical details with '.error'"
         ,
         Code::InvalidArgument | Code::AlreadyExists | Code::NotFound => err.message(),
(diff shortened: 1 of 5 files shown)


crates/flightrepl/src/lib.rs
```diff
@@ -60,10 +68,24 @@ pub async fn run(repl_config: ReplConfig) -> Result<(), Box<dyn std::error::Erro
         }
         match line {
             ".exit" | "exit" | "quit" | "q" => break,
+            ".error" => {
```
lukekim, line 71: We should distinguish between user query errors and runtime errors. I think this means query error, so we should label it as such.

lukekim: DuckDB labels their error as `Error: Parser Error`... I'd say the first error is redundant though. So, "Parser Error:" or "Query Error:"

image

lukekim (approved):

crates/flightrepl/src/lib.rs
phillipleblanc: This is for both runtime errors and user query errors. I was intending this to be used as a debugging tool, so we could get access to the raw error without pasting it to the screen by default. I can hide the help text for it if we don't want to expose it.

lukekim (dismissed):

crates/flightrepl/src/lib.rs
phillipleblanc: Renamed to `Query Error`

mitchdevenport (approved):

PR #869, opened 2024-03-19
Setup E2E Test CI workflow

ewgenius:
Added E2E CI workflow, intended to test binaries built from trunk or release.
Test scenarios following quickstart examples from https://github.com/spiceai/spiceai/blob/trunk/README.md


.github/workflows/e2e_test_ci.yml
```diff
@@ -0,0 +1,249 @@
...
+
+on:
+  push:
+    branches:
+      - trunk
+      - release-*
+
+  pull_request:
+    branches:
+      - trunk
```
ewgenius: TODO: disable trigger on PR, this should be run only on push to trunk/release branches, or manually

.github/workflows/e2e_test_ci.yml
```diff
@@ -0,0 +1,341 @@
...
+          mv target/release/spice spice
+          chmod +x spice
+
+      - name: Print versions
+        if: matrix.target_os != 'windows'
+        run: |
+          ./spice version
+          ./spiced --version
+
+      - name: Save spice artifact
```
sgrebnov, line 123: Alternatively we can do `make install` and upload the whole `.spice` folder. This way we don't need to check artifact name (spcied vs spiced.exe for windows) and this will automatically store all necessary artifacts/files (version file, spice folder structure, etc)

.github/workflows/e2e_test_ci.yml
```diff
@@ -0,0 +1,341 @@
...
+  test_quickstart_dremio:
+    name: "E2E Test: Dremio quickstart, using ${{ matrix.target_os }}-${{ matrix.target_arch }}"
+    runs-on: ${{ matrix.runner }}
+    needs: build
+
+    strategy:
+      matrix:
+        include:
+          - name: Linux x64
+            runner: rust
```
sgrebnov: We should test binaries are run on machine w/o rust/go installed  (if we don't plan to build anything)

.github/workflows/e2e_test_ci.yml
ewgenius: This should be as close as possible to release build, to not miss anything

.github/workflows/e2e_test_ci.yml
```diff
@@ -0,0 +1,343 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: build_${{ matrix.target_os }}_${{ matrix.target_arch }}
+          path: ./build
+
+      - name: Install spice
+        run: |
+          chmod +x ./build/spice
+          chmod +x ./build/spiced
+          echo "$HOME/build" >> $GITHUB_PATH
```
ewgenius:
```suggestion
          echo "${{ github.workspace }}/build" >> $GITHUB_PATH
```

.github/workflows/e2e_test_ci.yml
```diff
@@ -0,0 +1,343 @@
...
+          echo "$HOME"
```
ewgenius:
```suggestion
          echo "${{ github.workspace }}"
```

sgrebnov (dismissed):

.github/workflows/e2e_test_ci.yml
```diff
@@ -0,0 +1,343 @@
...
+  test_quickstart_dremio:
+    name: "E2E Test: Dremio quickstart, using ${{ matrix.target_os }}-${{ matrix.target_arch }}"
+    runs-on: ${{ matrix.runner }}
+    needs: build
+
+    strategy:
+      matrix:
+        include:
+          - name: Linux x64
+            runner: rust
```
ewgenius:
```suggestion
            runner: ubuntu-latest
```

.github/workflows/e2e_test_ci.yml
```diff
@@ -0,0 +1,343 @@
...
+        if: matrix.target_os != 'windows'
+        run: |
+          mv target/release/spice spice
+          chmod +x spice
+
+      - name: Print versions
+        if: matrix.target_os != 'windows'
+        run: |
+          ./spice version
+          ./spiced --version
```
ewgenius:
```suggestion
```

.github/workflows/e2e_test_ci.yml
```diff
@@ -0,0 +1,338 @@
...
+        uses: actions/upload-artifact@v4
+        if: matrix.target_os != 'windows'
+        with:
+          name: spiced_${{ matrix.target_os }}_${{ matrix.target_arch }}
+          path: spiced
+
+      - name: Merge spice and spiced in single artifact
+        uses: actions/upload-artifact/merge@v4
+        with:
+          name: build_${{ matrix.target_os }}_${{ matrix.target_arch }}
```
ewgenius:
```suggestion
      - name: Save spice artifact
        uses: actions/upload-artifact@v4
        if: matrix.target_os != 'windows'
        with:
          name: build_${{ matrix.target_os }}_${{ matrix.target_arch }}
          path: |
            spice
            spiced
```

.github/workflows/e2e_test_ci.yml
```diff
@@ -0,0 +1,335 @@
...
+
+      - name: Init spice app
+        run: |
+          spice init test_app
+
+      - name: Configure spice app
+        working-directory: test_app
+        run: |
+          # equal to "spice dataset configure"
+          mkdir -p datasets/eth_recent_blocks
```
sgrebnov, line 284: Let's simulate the same user commands/instructions we have in quickstart:

echo -e "eth_recent_blocks\neth recent logs\nspice.ai/eth.recent_blocks\ny" | spice dataset configure

.github/workflows/e2e_test_ci.yml
ewgenius: good point, switched to `ubuntu-latest`

.github/workflows/e2e_test_ci.yml
```diff
@@ -0,0 +1,325 @@
...
+        uses: ./.github/actions/setup-rust
+        with:
+          os: ${{ matrix.target_os }}
+
+      - name: Restore build cache (macOS)
+        if: matrix.target_os == 'darwin'
+        run: |
+          mkdir -p target
+          if [ -d /Users/spiceai/build/target ]; then
+            rsync -av /Users/spiceai/build/target/ target/
```
lukekim, line 65: Should we use the GH Cache action for these?

ewgenius: That step is executed only only on our `rust` runner which persists cache, so for this case it is faster than GH Cache

sgrebnov (dismissed):

.github/workflows/e2e_test_ci.yml
ewgenius: removed

phillipleblanc (approved):

.github/workflows/e2e_test_ci.yml
ewgenius: @sgrebnov good point, thanks. Let's add that in separate PR, merging this as is now.

Issue #839, opened 2024-03-18
Tonic OutOfRange messagetolarge error


lukekim:
**Describe the bug**
Crash due to message too large.

**To Reproduce**

1. Add `lukekim/demo` spicepod.
2. Run query:

```sql
SELECT *
FROM eth_recent_traces trace
JOIN eth_recent_transactions trans ON trace.transaction_hash = trans.hash
ORDER BY trans.block_number DESC
```

**Expected behavior**
No crash. Query completes.

**Screenshots**
![image](https://github.com/spiceai/spiceai/assets/80174/c55a8f00-5fdf-485b-a2e7-14b92b3759a9)

y-f-u:
There could be an easy fix to increase the tonic encoder/decoder range; however I am not satisfied with the solution as it still will bump into issues when the query is big.

Also for this query, the returning body is too huge to print out efficiently, a paging like `psql` would be a good addition.

Will prepare a PR to address these things.

lukekim:
Increase limit to get data *but* by default only print the first 500 lines to console.

Issue #874, opened 2024-03-19
BUG: REPL takes a long time to print out a large query result 


y-f-u:
**Describe the bug**
When using REPL to retrieve a large query result, the stdout is continuously printing for ages.

**To Reproduce**
- `spice sql`
- `select * from a_real_large_dataset`

**Expected behavior**
The repl should provide a nice way to handle this kind of situation, like `psql` to give user options to do paging, etc option

lukekim:
Will be fixed with #839

PR #875, opened 2024-03-19
make repl encoding/decoding size larger and improve OutOfRange error message

y-f-u:
This improves #839.

It introduces a 100MB encoder/decoder size rather than default 4MB.

Also it prints out an error message to suggest limit/page on the query. 

There will be following fix/improvement for paging the query response properly.


crates/flightrepl/src/lib.rs
```diff
@@ -38,7 +38,13 @@ pub async fn run(repl_config: ReplConfig) -> Result<(), Box<dyn std::error::Erro
     let channel = Channel::from_shared(repl_config.repl_flight_endpoint)?
         .connect()
         .await?;
-    let mut client = FlightServiceClient::new(channel);
+
+    // The encoder/decoder size is limited to be 100MB by default.
+    // This may not be enough for large queries.
+    // In future, paging will be supported to handle the large query output.
+    let mut client = FlightServiceClient::new(channel)
```
mitchdevenport, line 45: Let's allow the user to override this with a flag?

y-f-u: thought about this but decided not to go with this option. Once the `spice sql` is more feature complete to support paging like `psql`, user won't need to worry about this decoder/encoder size. 

The issue is described here: https://github.com/spiceai/spiceai/issues/874 which is sitting between feature/bug.

Even allowing user to config this, it won't be usable as it will flooding the terminal with super long content.

y-f-u: And this PR is only an improvement to current situation which gives no error about what is happening

phillipleblanc: Yeah I'm not sure how useful it is to even have 4MB of data flooding the screen here. Only use-case I think raising this limit makes sense for is if we would save the results to a parquet file or something.

phillipleblanc: But Mitch has a point that the error message says "configured maximum size" but we don't allow them to configure it.

y-f-u: yeah I will modify the error message.

phillipleblanc (approved):

PR #886, opened 2024-03-21
[WIP] Initial work on storing metrics locally.

Jeadie:
## Changes
- Add metrics Recorder, `LocalGaugeRecorder` that allows gauges to be stored locally.
    - TODO: Only store gauges with given prefix
-  Add metrics Recorder, `CompositeRecorder` that aggregates Recorders. Useful to run multiple recorders at global level.

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -4,6 +4,7 @@ use std::borrow::Borrow;
 use std::net::SocketAddr;
 use std::{collections::HashMap, sync::Arc};
 
+use ::metrics::{counter, gauge};
 use app::App;
 use config::Config;
 use model::Model;
@@ -26,6 +27,7 @@ pub mod datapublisher;
 pub mod dataupdate;
 mod flight;
 mod http;
+pub mod metrics;
 pub mod model;
 pub mod modelformat;
 pub mod modelruntime;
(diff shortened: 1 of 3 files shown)


crates/runtime/src/lib.rs
```diff
@@ -4,6 +4,7 @@ use std::borrow::Borrow;
 use std::net::SocketAddr;
```
Jeadie, line 4: Changes in this file are just around the namespacing of `metrics`

crates/runtime/src/metrics.rs
```diff
@@ -0,0 +1,307 @@
...
+impl LocalGaugeRecorder {
+    #[must_use]
+    pub fn new() -> Self {
+        let (operations_sender, mut operations_receiver) = mpsc::channel(100);
+        let gauges: Arc<RwLock<HashMap<Key, f64>>> = Arc::new(RwLock::new(HashMap::new()));
+
+        let recv_gauges: Arc<RwLock<HashMap<Key, f64>>> = gauges.clone();
+        tokio::spawn(async move {
+            while let Some(operation) = operations_receiver.recv().await {
+                println!("Received operation: {operation}. Gauges: {recv_gauges:#?}");
```
phillipleblanc, line 58:
```suggestion
```

Remove or use tracing

Jeadie:
Pausing work in this manner for an easier solution https://github.com/spiceai/spiceai/pull/892

PR #891, opened 2024-03-21
Add E2E tests for local acceleration

sgrebnov:
Adds tests for currently supported local materialization/acceleration engines and modes

Name | Description | Status | Engine Modes
-- | -- | -- | --
arrow | In-Memory Arrow Records | Alpha | memory
duckdb | Embedded DuckDB | Alpha | memory, file
sqlite | Embedded SQLite | Alpha | memory, file
postgres | Attached PostgreSQL | Alpha


ewgenius (dismissed):

.github/workflows/e2e_test_ci.yml
```diff
@@ -352,4 +352,127 @@ jobs:
...
+        run: spice version
+
+      - name: Spice dataset configure
+        working-directory: test_app
+        run: |
+          ENGINE=$(echo '${{ matrix.acceleration.engine }}')
+          MODE=$(echo '${{ matrix.acceleration.mode }}')
+
+          
+          echo "datasets:" >> spicepod.yaml
```
phillipleblanc, line 578: Can we pull the spicepods we use for e2e tests from a folder in the repo? The benefit to that is we have an easy place to go to for example spicepods. i.e. `cat $REPO_ROOT/test/spicepods/acceleration_arrow_memory.yaml > spicepod.yaml`

sgrebnov: Like it 👍 - I'm going merge this and create separate PR for this approach-  will see how to make this approach re-usable so we can rely on it for other scenarios as well (not just acceleration), there might be some discussion around additional meta, for example target dataset name to verify (which is part of spicepod definition and can vary)

ewgenius (dismissed):

ewgenius (approved):

PR #901, opened 2024-03-22
Huggingface models source

ewgenius:
Added support to pull models from huggingface.

Model can be configured using `huggingface` model source, and a link to the model `huggingface.co/spiceai/darts_private:latest`

Added optional `files` field to explicitly specify which files should be downloaded for HF model

```yaml


models:
  - from: huggingface:huggingface.co/spiceai/darts_private:latest
    name: darts_private
    datasets:
      - drive_stats_inferencing
    files:
      - model.onnx

```

It also requires `huggingface` secret with `token` field from HF user access tokens https://huggingface.co/settings/tokens

E.g. using env variables:

```bash
SPICE_SECRET_HUGGINGFACE_TOKEN='<...>' spice run
```

### DIFF
--- a/crates/runtime/src/modelsource/huggingface.rs
+++ b/crates/runtime/src/modelsource/huggingface.rs
@@ -1,3 +1,4 @@
+use super::Error;
 use super::ModelSource;
 use async_trait::async_trait;
 use regex::Regex;
@@ -13,7 +14,7 @@ pub struct Huggingface {}
 impl ModelSource for Huggingface {
     async fn pull(
         &self,
-        _: Secret,
+        secret: Secret,
         params: Arc<Option<HashMap<String, String>>>,
     ) -> super::Result<String> {
         let name = params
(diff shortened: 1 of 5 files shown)


crates/runtime/src/modelsource/huggingface.rs
```diff
@@ -0,0 +1,136 @@
...
+
+            let download_url = format!(
+                "https://huggingface.co/{}/{}/resolve/{}/{}",
+                caps["org"].to_owned(),
+                caps["model"].to_owned(),
+                revision,
+                file,
+            );
+
+            tracing::info!("Downloading model: {}", download_url);
```
phillipleblanc, line 107:
```suggestion
            tracing::debug!("Downloading model: {}", download_url);
```

crates/runtime/src/modelsource/spiceai.rs
```diff
@@ -119,6 +119,11 @@ impl ModelSource for SpiceAI {
         let versioned_path = format!("{local_path}/{version}");
         let file_name = format!("{versioned_path}/model.onnx");
 
+        if std::fs::metadata(file_name.clone()).is_ok() {
+            println!("File already exists: {file_name}, skipping download");
```
phillipleblanc:
```suggestion
            tracing::debug!("File already exists: {file_name}, skipping download");
```

phillipleblanc (approved):

PR #902, opened 2024-03-22
TPC-H benchmark basic test flow

sgrebnov:
Adds TPC-H benchmark basic test flow:
1. Generate test dataset and test queries
1. Load test datasets to local PostgreSQL
1. Run test queries against PostgreSQL directly
1. Run test queries against Spice connected to PostgreSQL

Notes (to be completed separately):
1. Some queries result in "No data returned for query" which should be investigated.
1. Currently Linux only, other platforms should be enabled and tested
1. In case of query error we should run and display information using`.error` (spice sql). Extract this logic to dedicated script and re-use across all tests when final spice sql query is performed
1. Query output/result is not currently processed


.github/workflows/tpch_benchmark.yml
```diff
@@ -0,0 +1,308 @@
...
+        run: psql -h localhost -U postgres -c 'SELECT version();'
+
+      - name: Generate TPC-H Benchmark data
+        run: |
+          git clone https://github.com/gregrahn/tpch-kit.git
+          cd tpch-kit/dbgen
+          make MACHINE=LINUX DATABASE=POSTGRESQL
+          ./dbgen -vf -s 0.1
+
+      - name: Load TPC-H Benchmark data to PostgreSQL
```
lukekim, line 219: For TPC-H specific, I think we'll want to have it in a separate script that we can run locally or through Actions.

sgrebnov: @lukekim - like it, I'll create a script + instructions to generate / run TPC-H  locally and then update the workflow to rely on the script.

sgrebnov: The note above has been addressed:
1. Added https://github.com/spiceai/spiceai/pull/905
2. Updated PR to rely on created script

digadeesh (approved):

PR #920, opened 2024-03-25
Bundle sqlite into the build

phillipleblanc:
This changes the build so that we don't rely on the system installed version of sqlite - we bundle it into the build.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,5 +47,5 @@ async-stream = "0.3.5"
 async-trait = "0.1.77"
 bb8 = "0.8"
 bb8-postgres = "0.8"
-rusqlite = "0.31.0"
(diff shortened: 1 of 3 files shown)


Jeadie (dismissed):

Cargo.toml
```diff
@@ -47,5 +47,5 @@ async-stream = "0.3.5"
 async-trait = "0.1.77"
 bb8 = "0.8"
 bb8-postgres = "0.8"
-rusqlite = "0.31.0"
+rusqlite = { version = "0.31.0", features = ["bundled"] }
```
mitchdevenport, line 50: We can also remove the installation of sqlite in the GH workflows now.

mitchdevenport (dismissed):

PR #928, opened 2024-03-26
Support for postgres connection string

### LINKED ISSUE #867: Bug: PostgreSQL DataConnector should allow parsing a connection string
**Describe the bug**
When signing up for a [PostgreSQL server in Neon](https://neon.tech/), it gives me a connection string that looks like this:
`postgresql://phillipleblanc:**********@ep-tight-bar-32413055.ap-southeast-1.aws.neon.tech/neondb?sslmode=require`

We should allow users to specify this as a secret in addition to individually configuring each piece with a different parameter.

mitchdevenport:
This PR adds support for providing a raw postgres connection string the the Postgres Data Connector and Accelerator.

Closes: #867


crates/db_connection_pool/src/postgrespool.rs
```diff
@@ -38,31 +38,32 @@ impl PostgresConnectionPool {
...
-                dbname = pg_db;
-            }
-            if let Some(pg_pass) = get_pg_pass(params, secret) {
-                connection_string.push_str(format!("password={pg_pass} ").as_str());
-            }
-            if let Some(pg_port) = params.get("pg_port") {
-                connection_string.push_str(format!("port={pg_port} ").as_str());
+            connection_string = String::new();
+
+            if let Some(pg_connection_string) = params.get("pg_connection_string") {
```
phillipleblanc: This should also be configurable via secrets

mitchdevenport: Good call

phillipleblanc (approved):

Issue #849, opened 2024-03-18
Bug: Better error message for 504 Gateway Timeout


phillipleblanc:
**Describe the bug**
When the runtime receives a 504 Gateway Timeout, it spews an unhelpful text dump to the console.

```bash
❯ spice run
Spice.ai runtime starting...
2024-03-18T08:32:35.123881Z  INFO spiced: Metrics listening on 127.0.0.1:9091
2024-03-18T08:32:35.130243Z  INFO runtime::http: Spice Runtime HTTP listening on 127.0.0.1:3000
2024-03-18T08:32:35.130286Z  INFO runtime::flight: Spice Runtime Flight listening on 127.0.0.1:50051
2024-03-18T08:32:35.130517Z  INFO runtime::opentelemetry: Spice Runtime OpenTelemetry listening on 127.0.0.1:50052
2024-03-18T08:32:35.804123Z  INFO runtime: Loaded dataset: smart_attribute_raw_value
2024-03-18T08:34:39.307581Z  INFO runtime::dataconnector: Refreshing data for smart_attribute_raw_value
2024-03-18T08:35:39.827285Z ERROR runtime::dataconnector::flight: Failed to read batch from flight client: Tonic(Status { code: Internal, message: "protocol error: received message with invalid compression flag: 60 (valid flags are 0 and 1) while receiving response with status: 504 Gateway Timeout", source: None })
```

digadeesh:
@Jeadie closing this as it seems its done.

PR #931, opened 2024-03-26
Better error message for Flight errors

Jeadie:
PR for issue: #849

### DIFF
--- a/crates/runtime/src/dataconnector/flight.rs
+++ b/crates/runtime/src/dataconnector/flight.rs
@@ -2,6 +2,7 @@ use std::future::Future;
 use std::pin::Pin;
 
 use arrow::record_batch::RecordBatch;
+use arrow_flight::error::FlightError;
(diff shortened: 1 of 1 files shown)


crates/runtime/src/dataconnector/flight.rs
```diff
@@ -57,3 +58,30 @@ impl Flight {
         })
     }
 }
+
+// Render a user-friendly message for when a FlightError occurs.
+pub fn render_flight_error(error: FlightError, ctx: &str) -> String {
```
phillipleblanc, line 63: We shouldn't include details about the full error in the main logs - that should go behind a debug! trace and the user-friendly error is shown in error! \

See the guide to user facing errors: https://github.com/spiceai/spiceai/blob/trunk/docs/dev/user_facing_errors.md

phillipleblanc (approved):

PR #945, opened 2024-03-26
Postgres login commands

### LINKED ISSUE #863: Bug: No way via CLI to login to a PostgreSQL server
**Describe the bug**
When attempting to use the PostgreSQL data connector with a secret stored in the file secret store - there is no CLI command to generate the correct secret structure, leaving a user unable to actually use the PostgreSQL connector.

**To Reproduce**
Navigate to the docs for PostgreSQL https://docs.spiceai.org/connectors/postgres/

Create a PostgreSQL dataset based on the example given:

```yaml
datasets:
  - from: postgres:path.to.my_dataset
    name: my_dataset
    params:
        pg_host: localhost
        pg_port: 5432
        pg_db: my_database
        pg_user: my_user
        pg_pass_key: my_secret
```

Start the runtime, get an error that `my_secret` doesn't exist.

mitchdevenport:
This PR adds support for `spice login postgres` and `spice login postgres engine` to login to postgres as a Data Connector and Acceleration Engine respectively.

This PR also adds a default engine secret name as `{engine_name}_engine` and a default for postgres secret keys as `{param}` to make logging in a more seamless experience.

Closes #863

### DIFF
--- a/bin/spice/cmd/login.go
+++ b/bin/spice/cmd/login.go
@@ -147,6 +147,36 @@ spice login s3 --access-key <key> --access-secret <secret>
 	}),
 }
 
+var postgresCmd = &cobra.Command{
+	Use:   "postgres",
+	Short: "Login to a Postgres instance",
+	Example: `
+spice login postgres --password <password>
+
+# See more at: https://docs.spiceai.org/
+`,
+	Run: CreateLoginRunFunc(api.AUTH_TYPE_POSTGRES, map[string]string{
+		passwordFlag: fmt.Sprintf("No password provided, use --%s or -p to provide a password", passwordFlag),
+	}, map[string]string{
+		passwordFlag: api.AUTH_PARAM_PG_PASSWORD,
+	}),
+}
+
+var postgresEngineCmd = &cobra.Command{
+	Use:   "engine",
+	Short: "Login to a Postgres instance as an engine",
+	Example: `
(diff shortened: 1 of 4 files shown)


bin/spice/cmd/login.go
```diff
@@ -147,6 +147,36 @@ spice login s3 --access-key <key> --access-secret <secret>
...
+var postgresCmd = &cobra.Command{
+	Use:   "postgres",
+	Short: "Login to a Postgres instance",
+	Example: `
+spice login postgres --password <password>
+
+# See more at: https://docs.spiceai.org/
+`,
+	Run: CreateLoginRunFunc(api.AUTH_TYPE_POSTGRES, map[string]string{
+		passwordFlag: fmt.Sprintf("No password provided, use --%s or -p to provide a password", passwordFlag),
```
lukekim, line 159: If we support -p it seems we should also support -u for username?

mitchdevenport: We don't currently support postgres username in secrets, but we could add that and then add to the CLI. Though it's not entirely necessary since usernames aren't sensitive.

lukekim: Then, i think this is fine for this iteration/PR, but for consistency I would add it. Some organizations do consider usernames to be sensitive.

lukekim (approved):

PR #960, opened 2024-03-27
Fix E2E flaky test - give spice more time for loading dremio dataset

ewgenius:
Fixing flaky dremio test, which fails sometimes


phillipleblanc (approved):
Not a huge fan of tests that rely on sleep to work. Ok for now, but ideally we have a deterministic way to do this.

PR #969, opened 2024-03-27
Improve Spice REPL show tables & row limit

### LINKED ISSUE #839: Tonic OutOfRange messagetolarge error

### LINKED ISSUE #874: BUG: REPL takes a long time to print out a large query result

### LINKED ISSUE #952: Bug: `spice init` creates an empty dir
**Describe the bug**
When running `spice init` the `spicepod.yaml` is created in the current dir and an empty dir is created alongside it.

**To Reproduce**

1. Run `spice init`
2. When prompted provide a `name`
3. Observe that the `spicepod.yaml` is created in the current dir and an empty dir is created alongside it

**Expected behavior**
The command should not create an empty dir.

phillipleblanc:
Improves the Spice REPL to rewrite the query for `show tables` to just list the tables in the `public` schema.

Also limits the output to the terminal to 500 lines.

Also fixes a bug with `spice init` creating an empty dir.

Closes #952
Closes #874 
Closes #839

### DIFF
--- a/crates/flightrepl/src/lib.rs
+++ b/crates/flightrepl/src/lib.rs
@@ -88,7 +88,7 @@ pub async fn run(repl_config: ReplConfig) -> Result<(), Box<dyn std::error::Erro
         if line.is_empty() {
             continue;
         }
-        match line {
+        let line = match line {
             ".exit" | "exit" | "quit" | "q" => break,
             ".error" => {
                 match last_error {
@@ -111,8 +111,11 @@ pub async fn run(repl_config: ReplConfig) -> Result<(), Box<dyn std::error::Erro
                 println!("\nAny other line will be interpreted as a SQL query");
                 continue;
             }
-            _ => {}
-        }
+            "show tables" => {
+                "select table_name from information_schema.tables where table_schema = 'public'"
+            }
+            _ => line,
+        };
 
         let _ = rl.add_history_entry(line);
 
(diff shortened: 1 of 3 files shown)


crates/flightrepl/src/lib.rs
```diff
@@ -187,6 +190,7 @@ pub async fn run(repl_config: ReplConfig) -> Result<(), Box<dyn std::error::Erro
         let df = DataFrame::new(
             ctx.state(),
             LogicalPlanBuilder::scan(UNNAMED_TABLE, provider_as_source(Arc::new(provider)), None)?
+                .limit(0, Some(500))?
```
y-f-u, line 193: we need to display a message saying we are limiting the result by 500

y-f-u:
Also to close https://github.com/spiceai/spiceai/issues/839 properly, based on my test, the query needs to have about ~120M decoding.

The batch we are getting from datafusion stream is 8K, so the size of the records need to smaller than that.

Unless people using a small limit or filter the selected columns; it's hard to get this fix properly.

y-f-u (dismissed):

ewgenius (approved):

PR #974, opened 2024-03-27
SQL View that depends on tables not defined should error out

sgrebnov:
SQL View that depends on tables not defined should error out

image


phillipleblanc (approved):
Approving for release - but this should check the loaded datasets to see if this will ever resolve, not just timing out.

PR #992, opened 2024-03-28
Verify view dependent tables while loading dataset

sgrebnov:
Verify view dependencies and report error if dependent table does not exist.

image

image

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -178,7 +178,7 @@ impl Runtime {
             loop {
                 let secrets_provider = shared_secrets_provider.read().await;
 
-                if !verify_dependent_tables(&ds, &existing_tables, df.clone()).await {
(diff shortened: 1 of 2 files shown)


ewgenius (dismissed):

crates/runtime/src/lib.rs
```diff
@@ -155,24 +155,35 @@ impl Runtime {
...
+        let existing_tables = all_datasets
+            .iter()
+            .map(|d| d.name.clone())
+            .collect::<Vec<String>>();
+
         tokio::spawn(async move {
             loop {
                 let secrets_provider = shared_secrets_provider.read().await;
 
+                if !verify_dependent_tables(&ds, &existing_tables, df.clone()).await {
```
```suggestion
                if !verify_dependent_tables(&ds, &existing_tables, Arc::clone(&df)).await {
```

I find this syntax clearer to understand - we're just cloning a pointer, not the object.

phillipleblanc (dismissed):

ewgenius (approved):

PR #1004, opened 2024-04-01
MySQL support

gloomweaver:


crates/db_connection_pool/src/dbconnection/mysqlconn.rs
```diff
@@ -0,0 +1,87 @@
...
+use mysql_async::{prelude::Queryable, Conn, Params};
+use mysql_common::value::convert::ToValue;
+use snafu::prelude::*;
+
+use super::{AsyncDbConnection, DbConnection, Result};
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("{source}"))]
+    QueryError { source: mysql_async::Error },
```
lukekim, line 35: Will need higher quality/more specific errors, including SyntaxError, ConnectionError, etc.

crates/arrow_sql_gen/src/mysql.rs
```diff
@@ -0,0 +1,265 @@
...
+    let arrow_fields = arrow_fields.into_iter().flatten().collect::<Vec<Field>>();
+    let options = &RecordBatchOptions::new().with_row_count(Some(rows.len()));
+    RecordBatch::try_new_with_options(Arc::new(Schema::new(arrow_fields)), columns, options)
+        .map_err(|err| Error::FailedToBuildRecordBatch { source: err })
+}
+
+fn map_column_to_data_type(column_type: &ColumnType) -> Option<DataType> {
+    match *column_type {
+        ColumnType::MYSQL_TYPE_NULL => Some(DataType::Null),
+        ColumnType::MYSQL_TYPE_BIT => Some(DataType::Boolean),
```
digadeesh: I don't think boolean is the correct conversion for "Bit"

digadeesh: https://dev.mysql.com/doc/refman/8.0/en/bit-type.html

digadeesh (dismissed):
Let's get this in and then iterate on the feedback.  Specifically the connection error messages.  

Let's also just do a second look on the type conversions.

crates/arrow_sql_gen/src/mysql.rs
ewgenius: Yes, I think the best fit option is `UInt64` and we need to convert manually
https://arrow.apache.org/docs/cpp/api/datatype.html#_CPPv4N5arrow4Type4type6UINT64E

ewgenius:
Let's also add a corresponding E2E test job for mysql, similar to postgres https://github.com/spiceai/spiceai/blob/c7a40ad7a46d1a3d3bb3a7b737c4923f6bee9d9e/.github/workflows/e2e_test_ci.yml#L342-L347

gloomweaver:
Will do tests in separate PR

crates/db_connection_pool/src/dbconnection/mysqlconn.rs
gloomweaver: Will address in separate PR, should also improve errors for other dataproviders

crates/arrow_sql_gen/src/mysql.rs
gloomweaver: Handled Bit type properly
CleanShot 2024-04-04 at 1  51 41@2x
If needed can map it to bool-u8-u64 depending on the length

digadeesh (dismissed):

digadeesh (approved):

PR #1015, opened 2024-04-02
show tables; return same as show table

### LINKED ISSUE #1013: `show tables` and `show tables;` show different content
**Describe the bug**
As title described

**To Reproduce**
Type in two commands separately and compare the output.

**Expected behavior**
Output should be same.

y-f-u:
fix #1013

Make show tables; return same as `show tables`

Could make it more complicated using regex, but instead adding another dependency, this will resolve 99% of the time


crates/flightrepl/src/lib.rs
```diff
@@ -109,7 +109,7 @@ pub async fn run(repl_config: ReplConfig) -> Result<(), Box<dyn std::error::Erro
                 println!("\nAny other line will be interpreted as a SQL query");
                 continue;
             }
-            "show tables" => {
+            "show tables" | "show tables;" => {
```
lukekim, line 112: Why not just trim space and ;

lukekim (approved):

crates/flightrepl/src/lib.rs
y-f-u: to escape from rust ownership check, you would need to do multiple lines to get rid of multiple suffix, etc. 

Which is not as nice as this one.

PR #1017, opened 2024-04-02
Verify connection to Dremio during dataset initialization

sgrebnov:
PR adds Dremio connection verification on dataset initialization:
 -  connection verification timeout is set to 30s

### DIFF
--- a/crates/ns_lookup/src/lib.rs
+++ b/crates/ns_lookup/src/lib.rs
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 use std::net::{SocketAddr, TcpStream};
+use std::time::Duration;
 
 use snafu::prelude::*;
 use trust_dns_resolver::{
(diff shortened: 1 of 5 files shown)


crates/ns_lookup/src/lib.rs
```diff
@@ -26,10 +27,47 @@ use trust_dns_resolver::{
 pub enum Error {
     #[snafu(display("Failed to connect to {host}:{port}, are the host and port correct?"))]
     UnableToConnect { host: String, port: u16 },
+
+    #[snafu(display("Failed to parse endpoint {endpoint}: {source}"))]
+    UnableToParseUrl {
+        endpoint: String,
+        source: url::ParseError,
+    },
+
```
digadeesh, line 36: Are these legitimate error states?  no default ports?

sgrebnov: Parsing error - it is unlikely, but still possible due to typos, for example if there is invalid character in the url specified(typo): `grpc://20 .163.171.81`

>Failed to parse endpoint grpc://20 .163.171.81: invalid domain character

sgrebnov: For ports - there is no default port for grpc:// in comparison to http:// and https://.  For gRPC, the use of port 50051 is more of a convention rather than a formal standard.

We use [URL](https://docs.rs/url/latest/url/) crate to parse and to see if there is default port available for schema.

My reasoning here is that we can fallback on some defaults (80 for example) but there is a chance that it will fail later during actual connection and it will be harder to detect the issue and the error message won't be very user friendly.

digadeesh (approved):

PR #1022, opened 2024-04-03
Ensure unit and doc tests compile and run.

rthomas:
Previously, running `$ cargo test --all` would fail with doctest
compilation issues as well as unit test failures. This change ensures
that the doctests compile and the unit test is updated.

The unit test in question (`test_table_insertion_with_list`) was
failing due to the generated SQL having `CAST` statements in it - the
`assert_eq` has been updated.

This change also adds a `test` make target that will run the rust
tests and ensures that this is run as a part of the PR GHA.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -32,13 +32,9 @@ jobs:
           fi
 
       - name: Set up Rust
-        run: |
-          echo "${CARGO_HOME:-$HOME/.cargo}/bin" >> $GITHUB_PATH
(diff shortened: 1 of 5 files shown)


phillipleblanc (approved):
Thanks @rthomas!

phillipleblanc:
The E2E tests are failing because they relied on some secrets that only get passed to workflows for members of the repo. You've found a bug in our process by being the first external contributor! We'll get this fixed, and I'll go ahead and merge this PR now.

Issue #1033, opened 2024-04-04
[Delta Lake Direct] Dataset in Databricks configured with Liquid Clustering returning no results when querying with no limit


digadeesh:
**Describe the bug**
Reported in Discord:

I am trying to query data in Databricks via Spice, and got it to return data when using a limit. However, when I use a where clause I am getting errors. I am also running my machine out of RAM when I attempt to setup acceliration. The dataset I am testing on is a sync of data from Maxmind, where every IP address has a record. The query and error are:

sql> select * from table_name_here where ip = '104.63.88.72';
Error An internal error occurred while processing the query. Show technical details with '.error'
No data returned for query
sql> .error
Status { code: Internal, message: "Arrow error: External error: Parquet error: ParquetObjectReader::get_byte_ranges error: Generic S3 error: request or response body error: operation timed out", source: None }

Using a 200 file, 45.6 GiB dataset on Databricks with liquid clustering enabled.

Using acceleration in-memory leads to OOM error.  File based does sqlite or duckdb also doesn't succeed.

ablyler:
Also, the query above would return a single row based on the data in the table.

y-f-u:
@ablyler did you consistently getting the S3 generic error with where clause?

Underneath we are using delta-rs, so this issue might be related: https://github.com/delta-io/delta-rs/issues/2301.

I am going to test against our own databricks with similar data volume to see if it can reproduced.

ablyler:
@y-f-u yes, it happens consistently. Exposing the timeout setting is a great idea!

Sevenannn:
@ablyler Thanks again for reporting this issue! 

I’ve investigated and couldn’t reproduce it using the latest version (v1.0.0-rc.1). Could you help confirm if it’s still occurring on your end? If it’s resolved, we’d like to close this issue. Please let us know if you’re still encountering any problems, thanks!

ablyler:
I am not sure, I don't have this environment set up at the moment. I say go ahead and close it and we can reopen if it happens again.

Sevenannn:
@ablyler Sounds great! I’ll go ahead and close this issue for now. Please let us know if the same issue happens in the future. Thanks again!

PR #1034, opened 2024-04-05
timeout for databricks

y-f-u:
this will address one of the problem in #1033 .

Default timeout is 30s, which failed on my test - a query with where clause on 30G+ dataset.
```
sql> select * from transactions where hash = '0xeabc0e3694a6e202c360a48b380b4205ee15541b816d66f56da687b3a3a0fd79';
Error An internal error occurred while processing the query. Show technical details with '.error'
No data returned for query
```

Increase it to 120s fixed the issue, result is returned successfully.

By
```
params:
  timeout: 120s
```

```
sql> select * from transactions where hash = '0xeabc0e3694a6e202c360a48b380b4205ee15541b816d66f56da687b3a3a0fd79';



+--------------------------------------------------------------------+-------+-------------------+--------------------------------------------+--------------------------------------------+-------+-------------+-------+-----------------------------+------------------+--------------------------+--------------+----------------+-----------------------+--------------+--------------------------------------------------------------------+-----------------+--------------------------+------------------+-----------------------------+-----------+
| hash                                                               | nonce | transaction_index | from_address                               | to_address                                 | gas   | gas_price   | input | receipt_cumulative_gas_used | receipt_gas_used | receipt_contract_address | receipt_root | receipt_status | block_timestamp       | block_number | block_hash                                                         | max_fee_per_gas | max_priority_fee_per_gas | transaction_type | receipt_effective_gas_price | end_block |
+--------------------------------------------------------------------+-------+-------------------+--------------------------------------------+--------------------------------------------+-------+-------------+-------+-----------------------------+------------------+--------------------------+--------------+----------------+-----------------------+--------------+--------------------------------------------------------------------+-----------------+--------------------------+------------------+-----------------------------+-----------+
| 0xeabc0e3694a6e202c360a48b380b4205ee15541b816d66f56da687b3a3a0fd79 | 18    | 74                | 0xdc4392cfde1d0e467a3d7602bbbc500bfe20304d | 0xdc4392cfde1d0e467a3d7602bbbc500bfe20304d | 46814 | 90340839758 | input | 3950477                     | 21000            |                          |              | 1              | +54048-12-19T15:20:00 | 14100236     | 0x000d2ed73601ebe1cadd19bec68cf2e59a44239bf8ee29e43da801d0d49fa2ed | 110709536502    | 1650000000               | 2                | 90340839758                 | 14441014  |
+--------------------------------------------------------------------+-------+-------------------+--------------------------------------------+--------------------------------------------+-------+-------------+-------+-----------------------------+------------------+--------------------------+--------------+----------------+-----------------------+--------------+--------------------------------------------------------------------+-----------------+--------------------------+------------------+-----------------------------+-----------+

Query took: 1008.800289729 seconds. 1/1 rows displayed.
```

### DIFF
--- a/crates/runtime/src/dataconnector/databricks.rs
+++ b/crates/runtime/src/dataconnector/databricks.rs
@@ -38,6 +38,7 @@ use super::{DataConnector, DataConnectorFactory};
 #[derive(Clone)]
 pub struct Databricks {
     secret: Arc<Option<Secret>>,
+    params: Arc<Option<HashMap<String, String>>>,
 }
 
 impl DataConnectorFactory for Databricks {
@@ -51,6 +52,7 @@ impl DataConnectorFactory for Databricks {
 
         let databricks = Self {
             secret: Arc::new(secret),
+            params: params.clone(),
         };
 
         Box::pin(async move {
@@ -79,10 +81,11 @@ impl DataConnector for Databricks {
     ) -> Pin<Box<dyn Future<Output = Vec<arrow::record_batch::RecordBatch>> + Send>> {
         let dataset = dataset.clone();
         let secret = Arc::clone(&self.secret);
+        let params = Arc::clone(&self.params);
         Box::pin(async move {
             let ctx = SessionContext::new();
 
-            let table_provider = match get_table_provider(&secret, &dataset).await {
+            let table_provider = match get_table_provider(&secret, &params, &dataset).await {
                 Ok(provider) => provider,
                 Err(e) => {
                     tracing::error!("Failed to get table provider: {}", e);
@@ -117,7 +120,7 @@ impl DataConnector for Databricks {
         &self,
         dataset: &Dataset,
     ) -> std::result::Result<Arc<dyn datafusion::datasource::TableProvider>, super::Error> {
-        get_table_provider(&self.secret, dataset).await
+        get_table_provider(&self.secret, &self.params, dataset).await
     }
 
     fn get_data_publisher(&self) -> Option<Box<dyn DataPublisher>> {
@@ -128,7 +131,7 @@ impl DataConnector for Databricks {
 impl DataPublisher for Databricks {
     fn add_data(&self, dataset: Arc<Dataset>, data_update: DataUpdate) -> AddDataResult {
         Box::pin(async move {
-            let delta_table = get_delta_table(&self.secret, &dataset).await?;
+            let delta_table = get_delta_table(&self.secret, &self.params, &dataset).await?;
 
             let _ = DeltaOps(delta_table)
                 .write(data_update.data)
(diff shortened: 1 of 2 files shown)


crates/runtime/src/dataconnector/databricks.rs
```diff
@@ -172,6 +177,15 @@ async fn get_delta_table(
     };
     storage_options.insert(AWS_S3_ALLOW_UNSAFE_RENAME.to_string(), "true".to_string());
 
+    let timeout: Option<String> = params
+        .as_ref() // &Option<HashMap<String, String>>
+        .as_ref() // Option<&HashMap<String, String>>
+        .and_then(|params| params.get("timeout").cloned());
+
+    if let Some(timeout) = timeout {
+        storage_options.insert("timeout".to_string(), timeout);
```
phillipleblanc, line 188: What happens if the timeout value is invalid?

phillipleblanc: Let's increase the default to 60s

phillipleblanc: Do we need this option for S3 as well?

y-f-u: > Let's increase the default to 60s

60s is not the magic number. It's rather good to stick the downstream dependency is, and leave the user to override it. Which follows the same pattern delta-rs using object storage.

> Do we need this option for S3 as well?

No need for s3 so far from the testing.

y-f-u: > failed to parse "60xxljksd" as Duration

Error propagated from delta-rs.

ablyler:
I went ahead and checked out this branch and ran it with the following in the dataset.yaml:

```yaml
params:
  timeout: 300s
```

I got the following when trying to run a query with a where clause (that matches a liquid clustered column):

```
sql> select * from TABLE_NAME where ip = '1.1.1.1';
Error An internal error occurred while processing the query. Show technical details with '.error'
No data returned for query
```

```
sql> .error
Status { code: Internal, message: "Parquet error: Parquet error: AsyncChunkReader::get_bytes error: Generic S3 error: Error after 1 retries in 240.140585625s, max_retries:10, retry_timeout:180s, source:error sending request for url (https://s3.us-east-1.amazonaws.com/BUCKET/CATALOG/__unitystorage/schemas/GUID/tables/GUID/part-GUID.c000.snappy.parquet): operation timed out", source: None }
```

y-f-u:
@ablyler thanks for trying this out. I think the timeout does resolve one issue, but we are likely bumping into another issue described here https://github.com/delta-io/delta-rs/issues/2301.

I will get this partial fix merged. Meanwhile, will investigate the new symptom.

crates/runtime/src/dataconnector/databricks.rs
y-f-u: set the default to 300s

digadeesh (approved):

PR #1037, opened 2024-04-05
MySQL support SSL

gloomweaver:


crates/db_connection_pool/src/mysqlpool.rs
```diff
@@ -80,9 +84,31 @@ impl MySQLConnectionPool {
...
+                                parameter_name: "mysql_sslmode".to_string(),
+                            }
+                            .fail()?;
+                        }
+                    }
+                }
             }
         }
 
+        let default_ssl_opts = SslOpts::default().with_danger_accept_invalid_certs(true);
```
phillipleblanc:
```suggestion
        let insecure_ssl_opts = SslOpts::default().with_danger_accept_invalid_certs(true);
```

crates/db_connection_pool/src/mysqlpool.rs
```diff
@@ -80,9 +84,31 @@ impl MySQLConnectionPool {
...
+        let ssl_opts = match ssl_mode {
+            "disable" => None,
+            "require" => Some(SslOpts::default()),
+            "prefer" => Some(default_ssl_opts),
+            _ => Some(default_ssl_opts),
```
phillipleblanc: Default should be require

crates/db_connection_pool/src/mysqlpool.rs
```diff
@@ -50,6 +53,7 @@ impl MySQLConnectionPool {
         secret: Option<Secret>,
     ) -> Result<Self> {
         let mut connection_string = mysql_async::OptsBuilder::default();
+        let mut ssl_mode = "prefer";
```
phillipleblanc: require

crates/db_connection_pool/src/mysqlpool.rs
gloomweaver: We use prefer as default for postgres, should it also be changed?

phillipleblanc: Yes

gloomweaver: Changed

crates/db_connection_pool/src/mysqlpool.rs
```diff
@@ -100,12 +100,10 @@ impl MySQLConnectionPool {
             }
         }
 
-        let default_ssl_opts = SslOpts::default().with_danger_accept_invalid_certs(true);
         let ssl_opts = match ssl_mode {
             "disable" => None,
```
phillipleblanc: Let's use the format of the MySQL ssl-mode enum: https://dev.mysql.com/doc/refman/8.0/en/connection-options.html#option_general_ssl-mode

i.e.

`DISABLED`
`PREFERRED`

gloomweaver: It can be confusing cause values for postgres dataconnector are different
I think we should be consisted with this

digadeesh: Its actually probably better to be consistent with the source data connector.  Then the user can match directly to  the official documentation rather than guessing what the mapping is.  

Nit - "disabled, preferred, required" instead of disable, require, prefer

ewgenius: > It can be confusing cause values for postgres dataconnector are different

postgres dataconnector follows postgres `sslmode` parameter https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION

mysql connector should be consistent with  mysql `ssl-mode`

gloomweaver: Updated

digadeesh (approved):

PR #1040, opened 2024-04-08
Rearchitect the runtime internals

phillipleblanc:
This is a major overhaul of the internal components for the Spice runtime.

The main changes in this PR are:

Add a new `data_components` crate which houses our implementations of DataFusion `TableProviders` for different data sources. Each component has 4 capabilities:
1. `Read` - This returns a `TableProvider` that can read from the connected source.
2. `ReadWrite` - This returns a `TableProvider` with the method `insert_into` implemented for inserting into the connected source.
3. `Stream` - This returns a `TableProvider` where the `.scan()` method returns an `ExecutionPlan` with `unbounded = true` to represent an infinite stream of batches.
4. `TableProviderFactory` - This creates a table in the connected source and returns a ReadWrite `TableProvider` that points to the newly created table.

A `DataConnector` is now a thin wrapper that returns a `TableProvider` based on the capability required, using the `data_components` crate.

A `DataAccelerator` is now a thin wrapper on a `TableProviderFactory`.

A new `AcceleratedTable` `TableProvider` was created that takes a references to a federated `TableProvider` and an accelerator `TableProvider`. It starts a refresh loop that depending on the settings will read from the federated table and insert it into the accelerated table.


crates/data_components/src/arrow.rs
```diff
@@ -0,0 +1,57 @@
...
+use datafusion::{
+    common::OwnedTableReference,
+    datasource::{MemTable, TableProvider},
+};
+use std::{error::Error, sync::Arc};
+
+use crate::{Read, ReadWrite};
+
+pub struct Arrow {
+    memtable: Arc<dyn TableProvider>,
```
edmondop: What is the requirement for this indirection? I find it suspicious that the struct name is `memtable` while the type is Arc<dyn TableProvider`. Can you please clarify?

crates/data_components/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+use std::{error::Error, sync::Arc};
+
+use async_trait::async_trait;
+use datafusion::{common::OwnedTableReference, datasource::TableProvider};
+
+pub mod arrow;
+pub mod databricks;
+
+#[async_trait]
+pub trait Read {
```
edmondop, line 38: What would be the impact of using Associate types here?
Would something like https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4976ceda80e9e354f4fbe7316f8af3c8 this work?

crates/data_components/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+    async fn table_provider(
+        &self,
+        table_reference: OwnedTableReference,
+    ) -> Result<Arc<dyn TableProvider + 'static>, Box<dyn Error + Send + Sync>>;
```
edmondop, line 42: Are we sure we want to do this? I think this allows underlying errors to propagate up in the call stack. Handling them correctly requires a knowledge of the type of error.

I think it's more robust to have a custom type of error for each provider (using another associate type) ,i.e. 

```
ArrowReadError
DatabricksReadError
```
etc

crates/flight_datafusion/src/lib.rs
```diff
@@ -270,16 +271,20 @@ impl ExecutionPlan for FlightExec {
...
+        );
+
+        Ok(Box::pin(stream_adapter))
     }
 }
 
 #[allow(clippy::needless_pass_by_value)]
-fn to_stream(client: FlightClient, sql: &str) -> impl Stream<Item = Result<RecordBatch>> {
+fn query_to_stream(
+    client: FlightClient,
```
edmondop: is this expected  that the client is then owned by the function? I would have expected the statement at line 288 to be outside, i.e. at invocation site

edmondop: Actually it is at invocation site, maybe you can only have a mut Flightclient at line 285?

crates/runtime/src/accelerated_table.rs
```diff
@@ -0,0 +1,260 @@
...
+                        .insert_into(
+                            &state,
+                            Arc::new(DataUpdateExecutionPlan::new(data_update)),
+                            overwrite,
+                        )
+                        .await
+                    {
+                        Ok(plan) => {
+                            if let Err(e) = collect(plan, ctx.task_ctx()).await {
+                                tracing::error!("Error adding data for {dataset_name}: {e}");
```
edmondop, line 116: I think we would rather return the state of success/failure to the outside rather than swallow it in async code, and let the calling function decide what to do with the error.  one option is to make this an async function that returns an Error for example

```rust
 async fn start_refresh(
        dataset_name: String,
        federated: Arc<dyn TableProvider>,
        refresh_mode: RefreshMode,
        accelerator: Arc<dyn TableProvider>,
    )  -> Option<TableRefreshError>
```

edmondop (commented):
Issued some comments, don't understand the codebase yet, will need to look into it for a bit

crates/data_components/src/arrow.rs
phillipleblanc: Yeah, this was leftover from a refactor I did. I'll rename it to `arrow_provider`.

crates/data_components/src/lib.rs
phillipleblanc: We need to work with trait objects for `TableProvider`s because that is what a lot of the DataFusion APIs accept, and also we want to eventually allow extending Spice using a plugin system, which would require trait objects - so that is why you might see us use trait objects more than is strictly required.

crates/data_components/src/lib.rs
phillipleblanc: This is a great idea - I originally had this as a concrete error type that all providers would have to adhere to - which is inflexible, because you basically have to aggregate all the possibles errors from each provider into one error type, or you have one super-generic error in the concrete instead which doesn't really help. However using associated types is great, and something I didn't think about before.

The only thing I can think of is I'm not sure if this would break the ability to make this a trait object, which as I mentioned above we want to keep. But I'll test it out.

crates/flight_datafusion/src/lib.rs
phillipleblanc: Yeah, I made it a `mut FlightClient` - thanks for catching this.

crates/runtime/src/accelerated_table.rs
phillipleblanc: Yeah, the design here is that we start this refresh on a "background thread" and immediately return the AcceleratedTable to be registered in DataFusion. In that design, there isn't anyone to return the error to - other than to report it to the console/metrics.

I'll keep it like this for this PR, since this has already taken way too much of my time and I want to get it merged. But happy to consider alternative ways to approach this.

crates/data_components/src/lib.rs
phillipleblanc: So it can remain as a trait object, but then we would need to specify the associated type whenever working with the trait object itself, which partially defeats the point. I'll keep as is for now.

lukekim (dismissed):

crates/spicepod/src/component/dataset.rs
```diff
@@ -78,7 +78,7 @@ impl Dataset {
         Dataset {
             from,
             name,
-            mode: None,
+            mode: Mode::default(),
             sql: None,
             sql_ref: None,
             params: Option::default(),
```
ewgenius: not related to changes, that should be just `None`

ewgenius:
```suggestion
            params: None,
```

phillipleblanc: I made this not optional

phillipleblanc: I made this not optional

Jeadie (approved):

PR #1063, opened 2024-04-09
Define AWS Secrets Store provider

sgrebnov:
Adds new provider to retrieve secrets from AWS Secrets Manager. Could be enabled via
```
secrets:
  store: aws_secrets_manager
```

Implementation for `AwsSecretStore` provider will be added separately


crates/spicepod/src/component/secrets.rs
```diff
@@ -43,4 +43,5 @@ pub enum SpiceSecretStore {
     Env,
     Kubernetes,
     Keyring,
+    Aws,
```
phillipleblanc:
```suggestion
    AwsSecretStore,
```

crates/secrets/src/lib.rs
```diff
@@ -80,17 +82,21 @@ pub enum SecretStoreType {
     #[cfg(feature = "keyring-secret-store")]
     Keyring,
     Kubernetes,
+    #[cfg(feature = "aws-secret-store")]
+    AwsSecretStore,
 }
 
 #[must_use]
+#[allow(unreachable_patterns)]
```
digadeesh: I don't think this is right.   You removed the last option so I think its an unreachable pattern now.  I'd rather not suppress these types of warnings unless we really have to.

crates/spicepod/src/component/secrets.rs
phillipleblanc: Probably needs to be serialized as snake case as well

crates/secrets/src/lib.rs
```diff
@@ -14,6 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+#[cfg(feature = "aws-secret-store")]
+pub mod aws;
```
lukekim: Let's make this specific. `aws_secrets_manager`

crates/secrets/src/lib.rs
sgrebnov: Updated, thank you 
```
#[cfg(not(all(feature = "keyring-secret-store", feature = "aws-secrets-manager")))]
 _ => None,
```

crates/spicepod/src/component/secrets.rs
sgrebnov: Updated. Thank you @phillipleblanc !
```
    #[serde(rename = "aws_secrets_manager")]
    AwsSecretsManager,
```

crates/secrets/src/lib.rs
sgrebnov: Updated to `aws_secrets_manager` (file name, struct name, feature flag so we have consistent naming everywhere). Thank you @lukekim!

digadeesh (approved):

PR #1065, opened 2024-04-10
Fix MySQL `TINY` and `DATE` types support

ewgenius:
- **Handle MYSQL_TYPE_DATE**
- **Add Int8 mapping for arrow types**

Fixed missing MySQL `DATE` type, and added int8 (MySQL `TINYINT`) to arrow types mapping

Before fix (mysql table with tinyint and date columns):

CleanShot 2024-04-10 at 17 25 07@2x

CleanShot 2024-04-10 at 17 26 28@2x

After:

CleanShot 2024-04-10 at 17 27 10@2x

### DIFF
--- a/crates/arrow_sql_gen/src/mysql.rs
+++ b/crates/arrow_sql_gen/src/mysql.rs
@@ -297,7 +297,7 @@ pub fn rows_to_arrow(rows: &[Row]) -> Result<RecordBatch> {
                                 }
                             };
 
-                            builder.append_value(Date32Type::from_naive_date(date))
+                            builder.append_value(Date32Type::from_naive_date(date));
                         }
                         None => builder.append_null(),
                     }
--- a/crates/arrow_sql_gen/src/mysql.rs
+++ b/crates/arrow_sql_gen/src/mysql.rs
@@ -281,25 +281,24 @@ pub fn rows_to_arrow(rows: &[Row]) -> Result<RecordBatch> {
                         },
                     )?;
 
-                    match v.clone() {
-                        Some(v) => {
-                            let date = match v {
-                                Value::Date(year, month, day, _, _, _, _) => {
-                                    chrono::NaiveDate::from_ymd_opt(
-                                        i32::from(year),
-                                        u32::from(month),
-                                        u32::from(day),
-                                    )
-                                    .unwrap_or_default()
-                                }
-                                _ => {
-                                    chrono::NaiveDate::from_ymd_opt(1970, 1, 1).unwrap_or_default()
-                                }
-                            };
+                    if let Some(v) = v.clone() {
+                        if let Value::Date(year, month, day, _, _, _, _) = v {
+                            let date = chrono::NaiveDate::from_ymd_opt(
+                                i32::from(year),
+                                u32::from(month),
+                                u32::from(day),
+                            );
 
-                            builder.append_value(Date32Type::from_naive_date(date));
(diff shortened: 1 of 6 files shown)


crates/arrow_sql_gen/src/mysql.rs
```diff
@@ -265,6 +265,43 @@ pub fn rows_to_arrow(rows: &[Row]) -> Result<RecordBatch> {
...
+                                Value::Date(year, month, day, _, _, _, _) => {
+                                    chrono::NaiveDate::from_ymd_opt(
+                                        i32::from(year),
+                                        u32::from(month),
+                                        u32::from(day),
+                                    )
+                                    .unwrap_or_default()
+                                }
+                                _ => {
+                                    chrono::NaiveDate::from_ymd_opt(1970, 1, 1).unwrap_or_default()
```
phillipleblanc: When would this happen? It feels weird to default to 1970 for a date

crates/arrow_sql_gen/src/mysql.rs
```diff
@@ -265,6 +265,43 @@ pub fn rows_to_arrow(rows: &[Row]) -> Result<RecordBatch> {
...
+                    match v.clone() {
+                        Some(v) => {
+                            let date = match v {
+                                Value::Date(year, month, day, _, _, _, _) => {
+                                    chrono::NaiveDate::from_ymd_opt(
+                                        i32::from(year),
+                                        u32::from(month),
+                                        u32::from(day),
+                                    )
+                                    .unwrap_or_default()
```
phillipleblanc: Why is the default value acceptable here?

gloomweaver:
With chrono feature enabled for mysql_async can just do 
```
ColumnType::MYSQL_TYPE_DATE => {
                    let Some(builder) = builder else {
                        return NoBuilderForIndexSnafu { index: i }.fail();
                    };
                    let Some(builder) = builder.as_any_mut().downcast_mut::<Date32Builder>() else {
                        return FailedToDowncastBuilderSnafu {
                            mysql_type: format!("{mysql_type:?}"),
                        }
                        .fail();
                    };
                    let v = row.get_opt::<NaiveDate, usize>(i).transpose().context(
                        FailedToGetRowValueSnafu {
                            mysql_type: ColumnType::MYSQL_TYPE_DATE,
                        },
                    )?;

                    match v {
                        Some(v) => {
                            builder.append_value(Date32Type::from_naive_date(v));
                        }
                        None => builder.append_null(),
                    }
                }
```

crates/arrow_sql_gen/src/mysql.rs
ewgenius: oh, you're right. fixed to set `null` instead

crates/arrow_sql_gen/src/mysql.rs
ewgenius: refactored

phillipleblanc (dismissed):

test/scripts/setup-data-mysql.sql
```diff
@@ -0,0 +1,63 @@
...
+--   NULL,
+--   NULL,
+--   NULL,
+--   NULL,
+--   NULL,
+--   NULL,
+--   NULL,
+--   NULL,
+--   NULL
+-- );
```
ewgenius, line 63: tracked as a bug in #1066

phillipleblanc (approved):

PR #1067, opened 2024-04-10
TEXT and DECIMAL types support

gloomweaver:
- Handle MYSQL_TYPE_BLOB correctly
- Handle MYSQL_TYPE_DECIMAL, MYSQL_TYPE_NEWDECIMAL
- Handle null values properly

Data:
CleanShot 2024-04-10 at 5  06 29@2x

Before fix:
CleanShot 2024-04-10 at 5  08 32@2x

After: 
CleanShot 2024-04-10 at 5  10 58@2x

P.S. 
- mysql driver for rust doesn't have a separate type for TEXT, so text columns are being handled as BLOB. Will create issue in the mysql repo.
- bigdecimal doesn't expose scale, so need to determine it via self-written hack


crates/arrow_sql_gen/src/mysql.rs
```diff
@@ -356,12 +412,29 @@ fn map_column_to_data_type(column_type: ColumnType) -> Option<DataType> {
...
         ColumnType::MYSQL_TYPE_FLOAT => Some(DataType::Float32),
         ColumnType::MYSQL_TYPE_DOUBLE => Some(DataType::Float64),
+        ColumnType::MYSQL_TYPE_DECIMAL | ColumnType::MYSQL_TYPE_NEWDECIMAL => None,
         ColumnType::MYSQL_TYPE_TIMESTAMP => Some(DataType::Timestamp(TimeUnit::Millisecond, None)),
         ColumnType::MYSQL_TYPE_DATE => Some(DataType::Date32),
         ColumnType::MYSQL_TYPE_VARCHAR
         | ColumnType::MYSQL_TYPE_STRING
-        | ColumnType::MYSQL_TYPE_VAR_STRING => Some(DataType::Utf8),
-        ColumnType::MYSQL_TYPE_BLOB => Some(DataType::Binary),
+        | ColumnType::MYSQL_TYPE_VAR_STRING
```
ewgenius, line 439: `BLOB` is a binary type, why treating it as a string?

ewgenius: > mysql driver for rust doesn't have a separate type for TEXT, so text columns are being handled as BLOB. Will create issue in the mysql repo.

oh, I see

ewgenius (commented):
Please update test table and data in `test/scripts/setup-data-mysql.sql` with decimal columns

gloomweaver:
Properly handle null values, updated test data

gloomweaver:
Fixed lint errors

crates/arrow_sql_gen/src/mysql.rs
```diff
@@ -235,6 +239,55 @@ pub fn rows_to_arrow(rows: &[Row]) -> Result<RecordBatch> {
                         i
                     );
                 }
+                ColumnType::MYSQL_TYPE_DECIMAL | ColumnType::MYSQL_TYPE_NEWDECIMAL => {
+                    let val = handle_null_error(row.get_opt::<BigDecimal, usize>(i).transpose())
+                        .context(FailedToGetRowValueSnafu {
+                            mysql_type: ColumnType::MYSQL_TYPE_DECIMAL,
+                        })?;
+
+                    let scale = match &val {
```
phillipleblanc, line 248: Could this be simplified with `.unwrap_or`?

phillipleblanc (dismissed):

crates/arrow_sql_gen/src/mysql.rs
gloomweaver: Can't use unwrap_or or other Option mapping methods, because that would transfer ownership, so better to match it ourselves
Could use clone, but seems unnecessary 
Found out that bigdecimal actually has `fractional_digit_count` method, so using it instead

phillipleblanc (approved):

PR #1075, opened 2024-04-11
Add Api to trigger manual refresh for dataset

sgrebnov:
Adds `v1/datasets/:name/refresh` Api endpoint to trigger manual dataset refresh.

Refresh logic is not implemented yet; `501 Not Implemented` is currently returned.

```
curl -i -X POST 127.0.0.1:3000/v1/datasets/taxi_trips/refresh
```

### DIFF
--- a/crates/runtime/src/http/v1.rs
+++ b/crates/runtime/src/http/v1.rs
@@ -408,7 +408,7 @@ pub(crate) mod datasets {
         let Some(dataset) = readable_app
             .datasets
             .iter()
-            .find(|d| d.name == dataset_name)
+            .find(|d| d.name.to_lowercase() == dataset_name.to_lowercase())
         else {
             return (
                 status::StatusCode::NOT_FOUND,
--- a/crates/runtime/src/http/v1.rs
+++ b/crates/runtime/src/http/v1.rs
@@ -426,7 +426,7 @@ pub(crate) mod datasets {
                 status::StatusCode::BAD_REQUEST,
                 Json(DatasetRefreshResponse {
                     message: format!(
-                        "Dataset {dataset_name} does not have local acceleration enabled"
+                        "Dataset {dataset_name} does not have acceleration enabled"
                     ),
                 }),
             )
--- a/crates/runtime/src/http/v1.rs
+++ b/crates/runtime/src/http/v1.rs
@@ -276,75 +276,12 @@ pub(crate) mod status {
     }
 }
 
-pub(crate) mod dataset_refresh {
-    use std::sync::Arc;
-
-    use app::App;
-    use axum::{
-        extract::Path,
-        http::status,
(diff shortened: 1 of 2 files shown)


crates/runtime/src/http/v1.rs
```diff
@@ -387,6 +388,53 @@ pub(crate) mod datasets {
...
+            return (status::StatusCode::INTERNAL_SERVER_ERROR).into_response();
+        };
+
+        let Some(dataset) = readable_app
+            .datasets
+            .iter()
+            .find(|d| d.name == dataset_name)
+        else {
+            return (
+                status::StatusCode::BAD_REQUEST,
```
digadeesh: Probably 404 is more appropriate here.

sgrebnov: Yeah, good point @digadeesh - thank you. Double checked and GitHub API also returns a 404 when a specific repository, user, or other resource specified in the URL path does not exist.

sgrebnov: Updated 👍

phillipleblanc (dismissed):

crates/runtime/src/http/v1.rs
```diff
@@ -387,6 +388,53 @@ pub(crate) mod datasets {
...
+    ) -> Response {
+        let app_lock = app.read().await;
+        let Some(readable_app) = &*app_lock else {
+            return (status::StatusCode::INTERNAL_SERVER_ERROR).into_response();
+        };
+
+        let Some(dataset) = readable_app
+            .datasets
+            .iter()
+            .find(|d| d.name == dataset_name)
```
lukekim: This should probably be a case-insensitive compare.

crates/runtime/src/http/v1.rs
```diff
@@ -387,6 +388,53 @@ pub(crate) mod datasets {
...
+        };
+
+        let acceleration_enabled = dataset.acceleration.as_ref().is_some_and(|f| f.enabled);
+
+        if !acceleration_enabled {
+            return (
+                status::StatusCode::BAD_REQUEST,
+                Json(DatasetRefreshResponse {
+                    message: format!(
+                        "Dataset {dataset_name} does not have local acceleration enabled"
```
lukekim:
```suggestion
                        "Dataset {dataset_name} does not have acceleration enabled"
```

crates/runtime/src/http/v1.rs
sgrebnov: Updated - thank you!

crates/runtime/src/http/v1.rs
sgrebnov: Updated - thank you!

digadeesh (dismissed):

digadeesh (approved):

PR #1078, opened 2024-04-11
Enable manual table refresh for AcceleratedTable

sgrebnov:
Updates `AcceleratedTable` refresh logic implementation to use mpsc::channel. This enables manual table refresh.
 - channel capacity is set to 1 so that sender is waiting until there is capacity
 - added `trigger_refresh` method as an example of manual refresh

### DIFF
--- a/crates/runtime/src/accelerated_table.rs
+++ b/crates/runtime/src/accelerated_table.rs
@@ -70,6 +70,11 @@ pub(crate) struct AcceleratedTable {
     _scheduled_refreshes_handle: Option<JoinHandle<()>>,
 }
 
+enum AccelerationRefreshMode {
+    Full(Receiver<()>),
+    Append,
+}
+
 impl AcceleratedTable {
     pub async fn new(
         dataset_name: String,
@@ -80,24 +85,25 @@ impl AcceleratedTable {
         object_store: Option<(Url, Arc<dyn ObjectStore + 'static>)>,
     ) -> Self {
         let mut refresh_trigger = None;
-        let mut refresh_trigger_receiver = None;
         let mut scheduled_refreshes_handle: Option<JoinHandle<()>> = None;
 
-        if refresh_mode == RefreshMode::Full {
-            let (trigger, receiver) = mpsc::channel::<()>(1);
-            refresh_trigger_receiver = Some(receiver);
-            refresh_trigger = Some(trigger.clone());
-            scheduled_refreshes_handle =
-                Self::schedule_regular_refreshes(refresh_interval, trigger).await;
+        let acceleration_refresh_mode: AccelerationRefreshMode = match refresh_mode {
+            RefreshMode::Append => AccelerationRefreshMode::Append,
+            RefreshMode::Full => {
+                let (trigger, receiver) = mpsc::channel::<()>(1);
+                refresh_trigger = Some(trigger.clone());
+                scheduled_refreshes_handle =
+                    Self::schedule_regular_refreshes(refresh_interval, trigger).await;
+                AccelerationRefreshMode::Full(receiver)
+            }
         };
 
         let refresh_handle = tokio::spawn(Self::start_refresh(
             dataset_name,
             Arc::clone(&federated),
-            refresh_mode.clone(),
+            acceleration_refresh_mode,
             Arc::clone(&accelerator),
             object_store,
-            refresh_trigger_receiver,
         ));
 
         Self {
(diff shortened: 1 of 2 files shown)


crates/runtime/src/accelerated_table.rs
```diff
@@ -20,6 +20,11 @@ use spicepod::component::dataset::acceleration::RefreshMode;
 use tokio::task::JoinHandle;
 use url::Url;
```
phillipleblanc, line 22:
```suggestion
```

crates/runtime/src/accelerated_table.rs
```diff
@@ -60,38 +71,71 @@ impl AcceleratedTable {
         refresh_interval: Option<Duration>,
         object_store: Option<(Url, Arc<dyn ObjectStore + 'static>)>,
     ) -> Self {
+        let (refresh_trigger, refresh_trigger_receiver) = mpsc::channel::<()>(1);
```
phillipleblanc: I would structure this so that we spawn the task that sends to the channel at this point in `new` only when the refresh mode is `Full`. Then all we pass into the `start_refresh` method is the receiver, not the sender (since it won't need it). We also don't need to pass the interval down anymore. We can also save the handle to the spawned task and abort it at the same time we abort the refresh_handle

And to leverage Rust's type system effectively, we could embed the receiver into a new RefreshType enum that only this struct knows about.

Something like:

```rust
pub enum AcceleratedRefreshMode {
        Full(Receiver<()>),
        Append,
    }
```

So when we match on the refresh mode in `stream_updates`, we get the receiver automatically. And we don't have to pass down another parameter.

crates/runtime/src/accelerated_table.rs
```diff
@@ -60,38 +71,71 @@ impl AcceleratedTable {
...
         }
     }
 
+    #[allow(dead_code)]
+    pub async fn trigger_refresh(&self) -> Result<()> {
+        self.refresh_trigger
+            .send(())
+            .await
+            .context(FailedToTriggerRefreshSnafu)?;
+        Ok(())
```
phillipleblanc: In the new model I specified earlier, `refresh_trigger` would become an Option and this would return an error if it was `None` (i.e trying to refresh a dataset that is configured as Append)

crates/runtime/src/accelerated_table.rs
sgrebnov: I really like this approach and experimented with it today. One thing I noticed is that mpsc::Receiver can't be copied or shared between threads so keeping it as part of `Full(Receiver<()>)` makes logic much more complex and requires Arc<Mutex<>> etc. Especially if you would like to keep AcceleratedRefreshMode as part of AcceleratedTable instance (to use instead of original RefreshMode to process manual refresh request).

Propose to proceed w/o AcceleratedRefreshMode struct as this point.

crates/runtime/src/accelerated_table.rs
sgrebnov: Updated - thank you

crates/runtime/src/accelerated_table.rs
phillipleblanc: The [`tokio::sync::mpsc::Receiver`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html) struct is `Sync` + `Send` - so it should be movable between threads. Its true that it can't be copied - but that is part of the design (multiple producer, *single* consumer). So when constructing the enum you would need to move the value, not copy it.

sgrebnov: Confirmed that the plan is to use AcceleratedRefreshMode instance for refresh loop initialization, no need to keep it as part of main struct, so we have single instance.

crates/runtime/src/accelerated_table.rs
```diff
@@ -73,22 +110,69 @@ impl AcceleratedTable {
             accelerator,
             federated,
             refresh_handle,
+            refresh_mode,
+            refresh_trigger,
+            _scheduled_refreshes_handle: scheduled_refreshes_handle,
```
phillipleblanc: This should be added to the `impl Drop` section to be aborted when the accelerated table is dropped.

crates/runtime/src/accelerated_table.rs
```diff
@@ -73,22 +110,69 @@ impl AcceleratedTable {
...
+        }
+    }
+
+    #[allow(dead_code)]
+    pub async fn trigger_refresh(&self) -> Result<()> {
+        if self.refresh_mode == RefreshMode::Append {
```
phillipleblanc: It seems a bit redundant to save the `refresh_mode` to check it here - couldn't we just move this error into the `None` arm below and not need to store `refresh_mode` on the table?

phillipleblanc: That would mean that not having a `refresh_trigger` implies that it is an Append refresh, which I think makes sense here.

sgrebnov: Updated, thank you (I thought about proper support when there will be more modes, updated as unnecessary at this point)

crates/runtime/src/accelerated_table.rs
sgrebnov: Thank you, updated

phillipleblanc (approved):

Issue #1057, opened 2024-04-09
Enhancement: Trigger manual refresh API


lukekim:
## Goal-State/What/Result

A `POST /v1/datasets/[name]/refresh` API for triggering manual refreshes from external applications.
 - The payload could contain additional refresh filters, like a specific Parquet file in the bucket for S3, or a time-range filter for time-series data.
 - Some safety machinery around refreshes to ensure duplicate/concurrent refreshes, don't happen, a good default of minimum refresh interval per source, etc. 
 - Data rate-limiting options for refreshes. 
 - Fuzzy Refresh: If many instances of Spice are refreshing at once, how do we ensure they all don't DDOS the source. Consider adding fuzziness to the refresh schedule.

Milestones:

M1: Basic dataset refresh
M2: Rate-Limited/Fuzzy Refresh. Moved to separate enhancement: https://github.com/spiceai/spiceai/issues/1102
M3: Filtered refresh. Moved to separate enhancement: https://github.com/spiceai/spiceai/issues/1102

## Why/Purpose

This is important for scenarios when very granular control over when data is refreshed is required:
 - have large dataset so would like to reduce unnecessary refreshes, data transfer over the network and associated cost
 - event based refreshes (for example, when source dataset has been updated)
 - when refresh schedule is hard to define using `refresh_interval`

For example, the team creates a reporting based on data queries every day at 8am. Instead of setting refresh interval for 1 hr to have fresh data I can create simple automation to trigger refresh via Api at 7-30am.

## By When

**Issue/Spec written and reviewed:** 04/10 (M1)

**M1:** 04/11
**M2 / M3:** Moved to separate enhancement: https://github.com/spiceai/spiceai/issues/1102
**Done-Done:** 04/17

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

**M1. Basic dataset refresh**

- `POST /v1/datasets/[name]/refresh` endpoint. 
  - Empty request body
[...]

PR #1083, opened 2024-04-11
Perform actual refresh on `v1/datasets/:name/refresh` Api call

sgrebnov:
Adds logic to trigger dataset refresh on `v1/datasets/:name/refresh` Api call

This implements basic dataset refresh scenarios for M1:  https://github.com/spiceai/spiceai/issues/1057


phillipleblanc (dismissed):

crates/runtime/src/datafusion.rs
```diff
@@ -284,6 +293,30 @@ impl DataFusion {
         Ok(())
     }
 
+    pub async fn trigger_table_refresh(&self, dataset_name: &str) -> Result<()> {
```
lukekim: Weird name. I'd suggest `pub async fn refresh_table(`

ewgenius (dismissed):

crates/runtime/src/datafusion.rs
```diff
@@ -284,6 +293,30 @@ impl DataFusion {
...
+        let table = self
+            .ctx
+            .table_provider(OwnedTableReference::bare(dataset_name.to_string()))
```
lukekim, line 311: Seems like there should be some output to console like `Refresh for dataset <name> started...`

sgrebnov: Confirm there is already a system message when refresh is initiated 👍
```bash
2024-04-12T02:34:01.868025Z  INFO runtime::accelerated_table: Refreshing data for recent_blocks
```

crates/runtime/src/datafusion.rs
sgrebnov: Updated - thanks!

lukekim (approved):

Issue #1062, opened 2024-04-09
Duckdb Data Connector


Sevenannn:
## Goal-State/What/Result

Spice OSS users can access persistent duckdb through the Spice runtime.

**Updates April 11th:**
MotherDuck relies on older version (0.9.2) of DuckDB, which causes issues including missing DuckDB data type, arrow - duckdb::arrow conversion, r2d2 DuckDB compatibility in other parts of the code. Since MotherDuck connector is based on a DuckDB connector, we redefine the milestones to start with implementing a DuckDB connector.

M1: Implement DuckDB data connector (using the version we use now 0.10.1).
M2: Pass E2E test For DuckDB Data Connector.
M3: DuckDB Data Connector Documentations & Customer Notified.

## Why/Purpose

Users of Spice that have data in DuckDB can query their data alongside other data sources through the spice runtime.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** April 9th

- [x] **M1:** April 10th
- [x] **M2:** April 15th
- [x] **M3:** April 15th

**Done-Done:** April 15th

## Done-Done

- [x] The Algorithm
- [x] First-Principles Driven
- [x] Issue reviewed by PM (Luke or Aurash)
- [x] Threat Model / Security Review
- [x] Tests (if applicable)
- [x] Dev
  - [x] Tested
  - [x] Qualitative & Performative Observability (Metrics & Dashboards)
  - [x] Alerting
- [x] Documentation
- [x] Customers notified (if applicable)

### User configuration in spicepod.yaml spec
```yaml
datasets:
  - from: duckdb:database.schema.table
    name: my_dataset
    params:
      open: path/to/duckdb_file.duckdb
```

## The Algorithm
- [x] Every requirement questioned?
  - [x] Duckdb rust library might not support attachment to MotherDuck
  - [x] Should MotherDuck connector be optional when runtime starts?
  - [x] How to shut down MotherDuck connection when not needed. 
- [x] Delete (Scope) any part you can.
  - [x] Duckdb connector, which could be relevant, should be supported in a separate enhancement ticket in the future
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automatatin.

digadeesh:
I think the mode: file doesn't makes sense in this case.  There isn't any other option for their data connector.  We'll just have data file based data connectors.

PR #1085, opened 2024-04-12
DuckDB Data Connector

Sevenannn:
Part of [Motherduck Data Connector](https://github.com/spiceai/spiceai/issues/1062). After discussion we start the ticket with the implementation of DuckDB Data Connector. See details from **Updates April 11th** in the ticket.

Key Changes:
* Remove `schema` field from `DuckDB` structure in [DuckDB data component](https://github.com/spiceai/spiceai/blob/trunk/crates/data_components/src/duckdb.rs) to support `impl ReadWrite for DuckDBTableFactory`
* Add new function `delete_all_table_data` in [DuckDB data component](https://github.com/spiceai/spiceai/blob/trunk/crates/data_components/src/duckdb.rs) to support duckdb overwrite and distinguish from `create_table`
* Added new function `new_with_file_mode` to create DuckDbConnectionPool for DuckDB data connector with File Mode enforced

User Configuration in spicepod.yaml. `duckdb_file` is a required parameter as it specify the duckdb source for data connector.
```yaml
from: duckdb:my_database.my_dataset
name: my_dataset
description: dataset from duckdb
params:
  open: path/to/my/duckdb/file
```


crates/runtime/src/dataconnector/duckdb.rs
```diff
@@ -0,0 +1,87 @@
...
+use std::{collections::HashMap, future::Future};
+
+use super::{DataConnector, DataConnectorFactory};
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Unable to create DuckDB connection pool: {source}"))]
+    UnableToCreateDuckDBConnectionPool { source: db_connection_pool::Error },
+
+    UnableToGetReadProvider {
```
phillipleblanc, line 38:
```suggestion
    #[snafu(display("{source}"))]
    UnableToGetReadProvider {
```

crates/data_components/src/duckdb.rs
```diff
@@ -137,18 +138,13 @@ fn to_datafusion_error(error: Error) -> DataFusionError {
 #[derive(Clone)]
 pub struct DuckDB {
     table_name: String,
-    schema: SchemaRef,
```
phillipleblanc: What is the rationale for removing schema here?

crates/data_components/src/duckdb.rs
```diff
@@ -260,24 +281,44 @@ impl DuckDB {
...
+#[async_trait]
+impl ReadWrite for DuckDBTableFactory {
+    async fn table_provider(
+        &self,
+        table_reference: OwnedTableReference,
+    ) -> Result<Arc<dyn TableProvider + 'static>, Box<dyn std::error::Error + Send + Sync>> {
+        let read_provider = Read::table_provider(self, table_reference.clone()).await?;
+
+        let table_name = table_reference.to_string();
+        let duckdb = DuckDB::new(table_name, Arc::clone(&self.pool));
```
phillipleblanc: The `read_provider` can give you the schema at this point to pass into the `duckdb` instance.

crates/data_components/src/duckdb.rs
Sevenannn: The only place where the schema field is really needed is in the original create_table function, specifically in the overwrite case. However, previously we are mixing the use of creating new table, recreating table, and overwrite old table within the create_table function. 

In creating a new table and recreating table cases, schema exists near create_table. It would make more sense to pass it as create_table input, instead of making it a field for duckdb. (Line 115: https://github.com/spiceai/spiceai/blob/639df26a93a714f32455e16387b7f85bd3ce652d/crates/data_components/src/duckdb.rs#L115)

The overwrite old table is what only requires the schema field. The previous way we do it is by dropping the old table, build an empty record batch with old table schema retrieved from the duckdb field, and recreate the table with the empty record from it. While what we can do is simply delete all of the rows in the old table, which doesn’t need a schema. (Line 119: https://github.com/spiceai/spiceai/blob/639df26a93a714f32455e16387b7f85bd3ce652d/crates/data_components/src/duckdb/write.rs#L119)

phillipleblanc: I think logically it makes sense for this struct to have the schema up front, even if its technically only needed at the create_table level.

But if you think it makes more sense to remove it, then we should make it consistent with the other data components as well.

crates/db_connection_pool/src/duckdbpool.rs
```diff
@@ -68,6 +71,32 @@ impl DuckDbConnectionPool {
...
+    /// * `params` - Data connector parameters for the connection pool.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if there is a problem creating the connection pool.
+    pub fn new_with_file_mode(params: &Arc<Option<HashMap<String, String>>>) -> Result<Self> {
+        let path = params
+            .as_ref()
+            .as_ref()
+            .and_then(|params| params.get("duckdb_file").cloned())
```
digadeesh: I thought we wanted the yaml to be 

params:
  open: path/to/file.db

digadeesh: Per the linked enhancement

crates/data_components/src/duckdb.rs
Sevenannn: Yeah, I think it logically makes sense if the schema is defined and not expected to change per DuckDB initialization. I’ll keep the schema area if so. I was referring to postgres implementation when changing duckdb and postgres doesn't keep a schema up front, so I was not sure whether to keep schema or not for DuckDB.

But I think we should keep the delete_all_table_data added, it makes the rewrite more efficient than the previous create_table cuz we only connected to DuckDB and executed the statement once. 

Would you share your thoughts on these? Thanks!

crates/db_connection_pool/src/duckdbpool.rs
Sevenannn: Updated

crates/data_components/src/duckdb.rs
phillipleblanc: Yes, the `delete_all_table_data` is good 👍

phillipleblanc: And the Postgres one should be changed to match this as well. That was an oversight on my part.

Sevenannn: DuckDB Schema parameter added back

crates/data_components/src/duckdb.rs
Sevenannn: Schema retrieved from read_provider in the updated code

crates/db_connection_pool/src/duckdbpool.rs
```diff
@@ -30,6 +30,9 @@ pub enum Error {
 
     #[snafu(display("ConnectionPoolError: {source}"))]
     ConnectionPoolError { source: r2d2::Error },
+
+    #[snafu(display("Missing open in data connector parameters"))]
```
sgrebnov, line 34:
```suggestion
    #[snafu(display("Missing required parameter:  open"))]
```

sgrebnov:
```suggestion
    #[snafu(display("Missing required parameter: open"))]
```

sgrebnov:
Reviewed CR notes w/ @Sevenannn  and tested. It is good to be merged IMO.

sgrebnov (approved):

PR #1096, opened 2024-04-12
Retention handlers

### LINKED ISSUE #1087: Add eviction/retention handler on accelerated dataset
```yaml
from: spice.ai/eth.recent_blocks
name: eth_recent_blocks
acceleration:
  enabled: true
  refresh_mode: append
  time_column: timestamp
  time_format: unix_seconds // defaults to 'unix_milli'
  retention_period: 30m
  retention_enabled: true
  retention_check_interval: 30m (default)
```

### LINKED ISSUE #1089: Engine - Postgres able to handle data eviction

### LINKED ISSUE #1090: Engine - MemTable able to handle data eviction

### LINKED ISSUE #1091: Engine - SQLite able to handle data eviction

### LINKED ISSUE #1092: Engine - DuckDB able to handle eviction

y-f-u:
add retention handler in accelerate table. Supported for engines, postgres, in-mem, duckdb, sqlite

fix #1087 
fix #1092 
fix #1089 
fix #1090 
fix #1091


y-f-u:
~~convert back to draft to handle more logical plan stuff~~

Had more thought on this one. At this stage, we don't need the logicall plan and physical plan conversion yet. As they will eventually translate into the delete action with filter which we know when retaining the data.

crates/data_components/src/lib.rs
```diff
@@ -59,3 +64,25 @@ pub trait Stream {
         table_reference: OwnedTableReference,
     ) -> Result<Arc<dyn TableProvider + 'static>, Box<dyn Error + Send + Sync>>;
 }
+
+#[async_trait]
+pub trait DeleteTableProvider: TableProvider {
+    async fn delete(
```
phillipleblanc:
```suggestion
    async fn delete_from(
```

I think calling it `delete_from` is clearer

crates/data_components/src/lib.rs
```diff
@@ -59,3 +64,25 @@ pub trait Stream {
...
+        &self,
+        _state: &SessionState,
+        _filters: &[Expr],
+    ) -> datafusion::error::Result<Arc<dyn ExecutionPlan>> {
+        Err(DataFusionError::Plan("Not implemented".to_string()))
+    }
+}
+
+// There is no good way to allow inter trait casting yet as TableProvider is not controlled
+pub fn cast_to_deleteable<'a>(
```
phillipleblanc: Would something like this not work?

https://github.com/datafusion-contrib/datafusion-federation/blob/main/datafusion-federation/src/analyzer.rs#L152

y-f-u: they are same. Both are casting to a concrete implication, then use return type to cast back to dyn trait.

crates/data_components/src/lib.rs
```diff
@@ -66,3 +71,25 @@ pub trait Stream: Send + Sync {
...
+    from: &'a dyn TableProvider,
+) -> Option<&'a (dyn DeleteTableProvider + 'a)> {
+    if let Some(p) = from.as_any().downcast_ref::<MemTable>() {
```
phillipleblanc: Shouldn't this downcast to a `dyn DeleteTableProvider`, not a `MemTable`?

crates/runtime/src/accelerated_table.rs
```diff
@@ -80,6 +81,11 @@ impl AcceleratedTable {
         refresh_mode: RefreshMode,
         refresh_interval: Option<Duration>,
         refresh_sql: Option<String>,
+        time_column: Option<String>,
+        time_format: Option<TimeFormat>,
+        retention_check_interval: Option<Duration>,
+        retention_period: Option<Duration>,
+        retention_enabled: bool,
```
phillipleblanc: Not a fan of all these parameters - can we leverage the type system more effectively here? Maybe something similar to what Sergei did for the refresh mode enum to only pass the data relevant to a particular case.

crates/data_components/src/lib.rs
y-f-u: no rust doesn't support downcast to dyn. It has to be a concrete struct in this case. Or else downcast won't be successful

y-f-u: essentially this will try to downcast to every concrete type of delete table provider, and return when succeed.

phillipleblanc: Right, but what I meant was we could have an `Adaptor` struct which just wraps the `dyn` trait similar to https://github.com/datafusion-contrib/datafusion-federation/blob/main/datafusion-federation/src/analyzer.rs#L152

y-f-u: why it can does that because when the instance is initialized (even with the wrapper), the table source type is already using the trait of source in the table provider. But what we are doing here is convert a tableprovider struct to a new trait type that has a new method.

y-f-u: Above example is not converting one type trait to another type of trait. It is consistent through the caller chain.

What we are doing is convert a trait to another trait; which rust won't allow in downcasting.

phillipleblanc: Right, what they have is:

```rust
pub struct FederatedTableProviderAdaptor {
    pub source: Arc<dyn FederatedTableSource>,
}
```

Downcast to the concrete `FederatedTableProviderAdaptor` and return the embedded `source`. We could do something similar.

y-f-u: even that, the `source` still need to be using the same type of trait. Can't trick rust in this case.

crates/data_components/src/arrow/write.rs
```diff
@@ -264,3 +266,146 @@ impl DataSink for MemSink {
...
+        for vec in df.collect_partitioned().await? {
+            for batch in vec {
+                new_batches[i].push(batch);
+            }
+
+            i = (i + 1) % batches.len();
+        }
+
+        for (target, mut batches) in batches.iter().zip(new_batches.into_iter()) {
+            target.write().await.append(&mut batches);
```
phillipleblanc, line 343: Wouldn't it be more efficient to take the lock once outside the loop?

crates/data_components/src/arrow/write.rs
```diff
@@ -264,3 +266,146 @@ impl DataSink for MemSink {
...
+        execution::context::SessionContext,
+        logical_expr::{cast, col, lit},
+        physical_plan::collect,
+        scalar::ScalarValue,
+    };
+
+    use crate::{arrow::write::MemTable, delete::DeletionTableProvider};
+
+    #[tokio::test]
+    #[allow(clippy::unwrap_used)]
```
phillipleblanc:
```suggestion
```

crates/data_components/src/delete.rs
```diff
@@ -0,0 +1,159 @@
...
+                        "failed to create record batch".to_string(),
+                    ))
+                }
+            })
+        })))
+    }
+}
+
+// There is no good way to allow inter trait casting yet as TableProvider is not controlled
+pub fn cast_to_deleteable<'a>(
```
phillipleblanc:
```suggestion
pub fn get_deletion_provider<'a>(
```

crates/runtime/Cargo.toml
```diff
@@ -76,6 +76,7 @@ data_components = { path = "../data_components" }
 mysql_async = { workspace = true, optional = true }
 postgres-native-tls = { version = "0.5.0", optional = true }
 ns_lookup = { path = "../ns_lookup" }
+datafusion-functions = "=36.0.0"
```
```suggestion
datafusion-functions = "37.0.0"
```

crates/runtime/src/accelerated_table.rs
phillipleblanc: All of these new parameters are only related to retention. Let's wrap them with a struct in an Option<> and then you can also get rid of `retention_enabled`

crates/data_components/src/arrow/write.rs
y-f-u: we need this for test not complaining when run single one. It will treat test fail if a panic happens.

crates/runtime/src/accelerated_table.rs
y-f-u: yeah actually about to move it to use enum.

crates/data_components/src/arrow/write.rs
phillipleblanc: We shouldn't use unwrap even in tests. We can use `.expect` which will do the same thing and doesn't bypass our style guide. It also gives context on why the panic happened, instead of just that it failed.

crates/data_components/src/lib.rs
phillipleblanc: This seems to work: https://github.com/spiceai/spiceai/pull/1145

crates/data_components/src/arrow/write.rs
phillipleblanc: Actually this is a separate lock per partition, so it should be fine as is.

crates/data_components/src/arrow/write.rs
```diff
@@ -264,3 +266,153 @@ impl DataSink for MemSink {
...
+
+        let mut df = DataFrame::new(
+            ctx.state(),
+            LogicalPlanBuilder::scan("?table?", provider_as_source(Arc::new(provider)), None)?
+                .build()?,
+        );
+
+        let mut count = df.clone().count().await?;
+
+        // TODO: check this either AND or OR makes more sense
```
phillipleblanc:
```suggestion
```

crates/data_components/src/duckdb/write.rs
```diff
@@ -173,3 +175,207 @@ impl DisplayAs for DuckDBDataSink {
...
+            .insert_into(&ctx.state(), exec, false)
+            .await
+            .expect("insertion should be successful");
+
+        collect(insertion, ctx.task_ctx())
+            .await
+            .expect("insert successful");
+
+        let delete_table = get_deletion_provider(table.clone())
+            .expect("table should be returned as deletetion provider");
```
phillipleblanc:
```suggestion
            .expect("table should be returned as deletion provider");
```

crates/data_components/src/sqlite/write.rs
```diff
@@ -173,3 +176,176 @@ impl DisplayAs for SqliteDataSink {
...
+
+#[async_trait]
+impl DeletionTableProvider for SqliteTableWriter {
+    async fn delete_from(
+        &self,
+        _state: &SessionState,
+        filters: &[Expr],
+    ) -> datafusion::error::Result<Arc<dyn ExecutionPlan>> {
+        Ok(Arc::new(DeletionExec::new(
+            Arc::new(SqliteDeletionSink::new(self.sqlite.clone(), filters)),
```
phillipleblanc:
```suggestion
            Arc::new(SqliteDeletionSink::new(Arc::clone(&self.sqlite), filters)),
```

crates/data_components/src/sqlite/write.rs
```diff
@@ -173,3 +176,176 @@ impl DisplayAs for SqliteDataSink {
...
+            .insert_into(&ctx.state(), Arc::new(exec), false)
+            .await
+            .expect("insertion should be successful");
+
+        collect(insertion, ctx.task_ctx())
+            .await
+            .expect("insert successful");
+
+        let table =
+            get_deletion_provider(table).expect("table should be returned as deletetion provider");
```
phillipleblanc:
```suggestion
            get_deletion_provider(table).expect("table should be returned as deletion provider");
```

crates/runtime/src/accelerated_table.rs
```diff
@@ -52,6 +60,10 @@ pub enum Error {
 
     #[snafu(display("Manual refresh is not supported for `append` mode"))]
     ManualRefreshIsNotSupported {},
+
+    UnableToGetUnixTimestamp {
```
phillipleblanc: This needs a display - otherwise the error message just shows this name.

crates/runtime/src/accelerated_table.rs
```diff
@@ -155,6 +195,82 @@ impl AcceleratedTable {
...
+        loop {
+            interval_timer.tick().await;
+
+            if let Some(deleted_table_provider) = get_deletion_provider(Arc::clone(&accelerator)) {
+                let ctx = SessionContext::new();
+
+                let start = SystemTime::now() - retention_period;
+
+                let Ok(timestamp) = get_timestamp(start) else {
+                    tracing::error!("[retention] failed to get timestamp");
```
```suggestion
                    tracing::error!("[retention] Failed to get timestamp");
```

crates/runtime/src/accelerated_table.rs
```diff
@@ -155,6 +195,82 @@ impl AcceleratedTable {
...
+                    continue;
+                };
+                let expr = get_expr(&time_column, timestamp, expr_time_format.clone());
+                tracing::info!(
+                    "[retention] Evicting data for {dataset_name} where {time_column} < {}...",
+                    if let Some(value) = chrono::DateTime::from_timestamp(timestamp as i64, 0) {
+                        value.to_rfc3339()
+                    } else {
+                        tracing::warn!("[retention] unable to convert timestamp");
```
```suggestion
                        tracing::warn!("[retention] Unable to convert timestamp");
```

crates/runtime/src/accelerated_table.rs
```diff
@@ -284,11 +400,78 @@ impl AcceleratedTable {
...
+                ExprTimeFormat::UnixTimestamp(ExprUnixTimestamp { scale })
+            }
+            DataType::Timestamp(_, _)
+            | DataType::Date32
+            | DataType::Date64
+            | DataType::Time32(_)
+            | DataType::Time64(_) => ExprTimeFormat::Timestamp,
+            DataType::Utf8 | DataType::LargeUtf8 => ExprTimeFormat::ISO8601,
+            _ => {
+                tracing::warn!("date type is not handled yet: {}", field.1.data_type());
```
```suggestion
                tracing::warn!("Date type is not handled yet: {}", field.1.data_type());
```

crates/sql_provider_datafusion/src/expr.rs
```diff
@@ -23,14 +23,55 @@ pub enum Error {
...
     match expr {
         Expr::BinaryExpr(binary_expr) => {
-            let left = to_sql(&binary_expr.left)?;
-            let right = to_sql(&binary_expr.right)?;
+            let left = to_sql_with_engine(&binary_expr.left, engine)?;
+            let right = to_sql_with_engine(&binary_expr.right, engine)?;
+
+            if let Some(Engine::DuckDB) = engine {
+                // DuckDB doesn't support compare timestamp with timestamptz in 0.10.1 yet. Revisit
+                // in 0.10.2
```
phillipleblanc, line 41: Let's add this as a note to https://github.com/spiceai/spiceai/issues/1141

phillipleblanc (dismissed):

ewgenius (approved):

PR #1110, opened 2024-04-15
Databricks connector based on spark-connect-rs

edmondop:
Reworks the Databricks connector to use [Spark Connect](https://www.databricks.com/blog/2022/07/07/introducing-spark-connect-the-power-of-apache-spark-everywhere.html) as the mechanism to execute queries against the Databricks instance.


phillipleblanc (commented):
Thank you @edmondop!

PR #1115, opened 2024-04-15
Choose between databricks implementations via `format` & `mode` parameters

phillipleblanc:
Chooses to use the previous `databricks` behavior that uses delta tables directly if the `mode` parameter is set to `s3` and the `format` is `deltalake`.


crates/data_components/src/databricks_spark.rs
```diff
@@ -23,12 +23,12 @@ use std::{collections::HashMap, error::Error, sync::Arc};
 use crate::{spark_connect, Read, ReadWrite};
 
 #[derive(Clone)]
-pub struct Databricks {
+pub struct DatabricksSpark {
```
lukekim: DatabricksSparkConnect might be a better name?

lukekim (dismissed):

lukekim (approved):

Issue #1084, opened 2024-04-12
Unloading and then reloading datasets shouldn't clear out data


phillipleblanc:
**Describe the bug**
When a dataset is hot reloaded in the runtime, it is first unregistered from DataFusion and then the new dataset is registered. This can lead to a period of time where the new dataset hasn't finished initializing/loading data and it appears there are 0 rows.

We should wait for the new dataset to finish its initialization before unloading the old dataset and replacing it.

PR #1126, opened 2024-04-16
Test connection when new connection pool created

ewgenius:
Partially for #1084

Added check for connectivity inside connection pools.

Refactored `load_dataset` by splitting int two methods `load_dataset_connector` and `register_loaded_dataset`.

Updated `update_dataset` logic: The process now follows these steps:

1. Load the dataset connector and test the connection.
2. If the load and test are successful, unload the old dataset.
3. Register the updated dataset.

That will minimize dataset downtime and prevent wiping out the loaded dataset in case of a wrong configuration update.

The `test_connection` function has been implemented for PostgreSQL, MySQL, and DuckDB connectors. For other connectors, a default stub used.

https://github.com/spiceai/spiceai/assets/827338/627b8a54-025b-417a-a6da-a0261fb2decf

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -215,14 +215,6 @@ impl Runtime {
                 }
             };
 
-            if let Ok(()) = connector.test_connection().await {
-                tracing::info!("Connection verified for dataset: {}", &ds.name);
-            } else {
-                tracing::error!("Unable to verify connection for dataset: {}", &ds.name);
-                status::update_dataset(&ds.name, status::ComponentStatus::Error);
-                return;
-            }
-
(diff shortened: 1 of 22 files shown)


crates/runtime/src/lib.rs
```diff
@@ -313,9 +374,26 @@ impl Runtime {
     }
 
     pub async fn update_dataset(&self, ds: &Dataset, all_datasets: &[Dataset]) {
```
sgrebnov, line 368: I really like the `test_connection`. Not sure it is the right thing to do to not unload the original dataset when updated definition can't be loaded.

sgrebnov (dismissed):

phillipleblanc (approved):

Issue #856, opened 2024-03-19
Bug: Improve warning message when data connector can't be initialized


phillipleblanc:
**Describe the bug**
The current warning when we can't initialize a dataconnector is too verbose. We should just print what we were trying to do, and the root cause of that - not all of the intermediate layers.

Current message:
`2024-03-19T03:09:21.717423Z  WARN runtime: Failed to get data connector from source for dataset smart_attribute_raw_value, retrying: Unable to initialize data connector spiceai: Unable to create data connector: Unable to connect to server: Unable to connect to endpoint: transport error`

Sevenannn:
This is break down into separate enhancements for each data connector. See details from:
- https://github.com/spiceai/spiceai/issues/1229

PR #1133, opened 2024-04-17
Better Data Connector Initialization Error Message Without Intermediate Layer

Sevenannn:
This is more about a design question: which intermediate layers we want to keep, which also apply to error messages other than Data Connector Initialization Error

- https://github.com/spiceai/spiceai/issues/856

### DIFF
--- a/crates/runtime/src/dataconnector/postgres.rs
+++ b/crates/runtime/src/dataconnector/postgres.rs
@@ -31,7 +31,7 @@ use super::{DataConnector, DataConnectorFactory};
 
 #[derive(Debug, Snafu)]
 pub enum Error {
-    #[snafu(display("Unable to create Postgres connection pool: {source}"))]
+    #[snafu(display("{source}"))]
     UnableToCreatePostgresConnectionPool { source: db_connection_pool::Error },
 
     #[snafu(display("{source}"))]


crates/runtime/src/dataconnector/postgres.rs
```diff
@@ -31,7 +31,7 @@ use super::{DataConnector, DataConnectorFactory};
 
 #[derive(Debug, Snafu)]
 pub enum Error {
-    #[snafu(display("Unable to create Postgres connection pool: {source}"))]
```
phillipleblanc, line 34: We can use the Snafu `transparent` attribute for this: https://docs.rs/snafu/latest/snafu/derive.Snafu.html#delegating-to-the-underlying-error

digadeesh:
If you can point to the error directly without displaying the underlying source error.  We should do that without showing this proprietary errors from each data source because this will look more consistent.

If we can't, then we should have a simple statemnt pointing to the error in our runtime, and then display the source error after.

Something like:

Failed to connect to S3 datasource {name}.  Error from datasource: {S3's error returned to the client}

Sevenannn:
The data connector initialization error message, along with other error message, will be addressed in this ticket
- https://github.com/orgs/spiceai/projects/18/views/1?pane=issue&itemId=60604317
New PR will be opened to refactor all error messages in Spice.

Issue #853, opened 2024-03-18
Enhancement: Spice installation on Windows


sgrebnov:
## Goal-State/What/Result

Latest version of Spice could be installed on Windows via PowerShell (or via [WinGet](https://learn.microsoft.com/en-us/windows/package-manager/)). The experience is similar as the current installation process on Linux and macOS

Milestones:

- M1 Latest Spice version could be installed on Windows via shell command

## Why/Purpose

Windows is the most popular operating system for developers according to the [StackOverflow](https://survey.stackoverflow.co/2023/#section-most-popular-technologies-operating-system) -  46.91% of developers use Windows professionally at work. They are unable to use Spice as there is no Spice installation available for Windows.

## By When

**Issue/Spec written and reviewed:** 04/16

**M1:** 04/17
**Done-Done:** 04/17

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation (installation has been documented in docs, quickstarts refer to docs)
- [x] Samples/Quickstarts/Tutorials (n/a; same samples/quickstarts work on Windows, no dedicated quickstart for Windows is required)

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

1.  I can run the following command to install latest Spice on Windows (arm and intel)
```shell
curl -L "https://raw.githubusercontent.com/spiceai/spiceai/trunk/install/Install.ps1" -o Install.ps1 && Powershell -ExecutionPolicy Bypass -File ./Install.ps1
```
```shell
curl -L "https://win-install.spiceai.org" -o Install.ps1 && Powershell -ExecutionPolicy Bypass -File ./Install.ps1
```
   -  Note: default PowerShell's execution policy is usually configured as "Restricted" that prevents any scripts from running. Temporary overriding execution policy via `-ExecutionPolicy Bypass` is a standard approach for PowerShell scripts execution.
1. spice.exe is installed to `C:\Users\<user-name>\.spice\bin` (similar to Linux/macOS)
1. spice.exe is added to Path
1. Spice runtime works when started as `spice run`

[...]

PR #1134, opened 2024-04-17
Document Windows installation and add test

sgrebnov:
Part of https://github.com/spiceai/spiceai/issues/853:  Document Windows installation and add installation test

Example test run: https://github.com/spiceai/spiceai/actions/runs/8731684944/job/23957488118

Notes:
1. `"https://raw.githubusercontent.com/spiceai/spiceai/trunk/install/Install.ps1"` URL will be replaced with shorter version (similar to `https://install.spiceai.org`that is used for Linux and macOS)

### DIFF
--- a/.github/workflows/e2e_test_release_install.yml
+++ b/.github/workflows/e2e_test_release_install.yml
@@ -27,8 +27,15 @@ jobs:
             runner: macos-12
             target_os: darwin
             target_arch: x86_64
+          - name: Windows x64
+            runner: rust-windows-x64
+            target_os: windows
+            target_arch: x86_64
+            target_arch_go: amd64
+            
     steps:
     - name: system info
+      if : matrix.target_os != 'windows'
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e_test_release_install.yml
```diff
@@ -27,8 +27,15 @@ jobs:
             runner: macos-12
             target_os: darwin
             target_arch: x86_64
+          - name: Windows x64
+            runner: rust-windows-x64
```
ewgenius: @sgrebnov that is our hosted runner, we shouldn't install spice there, let's use `windows-latest` instead

ewgenius:
```suggestion
            runner: windows-latest
```

ewgenius (dismissed):

ewgenius (approved):

PR #1140, opened 2024-04-18
Allow only alphanumeric dataset names when using `spice dataset configure`

### LINKED ISSUE #957: Adding a dataset with dots needs to be quoted or it shows as not existing
**Describe the bug**
Creating a dataset name with dots, i.e. `eth.recent_blocks` will create a dataset with that table name:

```bash
sql> show tables
+---------------+--------------------+-------------------+------------+
| table_catalog | table_schema       | table_name        | table_type |
+---------------+--------------------+-------------------+------------+
| datafusion    | public             | eth.recent_blocks | BASE TABLE |
| datafusion    | information_schema | tables            | VIEW       |
| datafusion    | information_schema | views             | VIEW       |
| datafusion    | information_schema | columns           | VIEW       |
| datafusion    | information_schema | df_settings       | VIEW       |
+---------------+--------------------+-------------------+------------+
```

But querying without quoting will return an error:

```bash
sql> select * from eth.recent_blocks limit 10
Error An internal error occurred while processing the query. Show technical details with '.error'
```

Querying with quotes will work correctly:

```bash
select * from "eth.recent_blocks" limit 10
```

**Expected behavior**
We should warn users when they try to create/load a dataset with a name that has dots.

ewgenius:
closes #957 

CleanShot 2024-04-19 at 12 48 25@2x

CleanShot 2024-04-19 at 12 49 41@2x

### DIFF
--- a/bin/spice/cmd/dataset.go
+++ b/bin/spice/cmd/dataset.go
@@ -81,6 +81,11 @@ spice dataset configure
 			os.Exit(1)
 		}
 
+		if strings.Contains(datasetName, "-") {
+			// warn that dataset name with hyphen should be quoted in queries
+			cmd.Println(aurora.BrightYellow(fmt.Sprintf("Dataset names with hyphens should be quoted in queries:\nSELECT * FROM \"%s\"", datasetName)))
+		}
+
 		cmd.Print("description: ")
 		description, err := reader.ReadString('\n')
 		if err != nil {
--- a/bin/spice/cmd/dataset.go
+++ b/bin/spice/cmd/dataset.go
@@ -83,7 +83,7 @@ spice dataset configure
 
 		if strings.Contains(datasetName, "-") {
 			// warn that dataset name with hyphen should be quoted in queries
-			cmd.Println(aurora.BrightYellow(fmt.Sprintf("Dataset names with hyphens should be quoted in queries:\nSELECT * FROM \"%s\"", datasetName)))
+			cmd.Println(aurora.BrightYellow(fmt.Sprintf("Dataset names with hyphens should be quoted in queries:\ni.e. SELECT * FROM \"%s\"", datasetName)))
 		}
 
 		cmd.Print("description: ")
--- a/bin/spice/cmd/dataset.go
+++ b/bin/spice/cmd/dataset.go
@@ -69,6 +69,11 @@ spice dataset configure
 			datasetName = defaultDatasetName
 		}
 
+		if strings.Contains(datasetName, ".") {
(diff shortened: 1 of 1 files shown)


phillipleblanc (dismissed):

bin/spice/cmd/dataset.go
```diff
@@ -69,6 +69,11 @@ spice dataset configure
 			datasetName = defaultDatasetName
 		}
 
+		if strings.Contains(datasetName, ".") {
```
lukekim: @ewgenius are there other characters we should check for? If so, we should just check for the slice of them here.

ewgenius: I don't see any restrictions in datafusion table names, but I think we should be more strict in our case and allow using only alphanumeric characters with `_` and `-` symbols.

@phillipleblanc what do you think?

ewgenius: and just checked that names with hyphens should be quoted in queries

CleanShot 2024-04-19 at 12 34 24@2x

phillipleblanc: Yeah dashes also need to be escaped. I think we should warn for dashes, not prevent it because dots in SQL table names are probably a mistake, whereas dashes are more common - but they should be aware it needs to be quoted

ewgenius: Added more strict rules:

CleanShot 2024-04-19 at 12 48 25@2x

Also warning about hyphens
CleanShot 2024-04-19 at 12 49 41@2x

bin/spice/cmd/dataset.go
```diff
@@ -69,6 +70,22 @@ spice dataset configure
...
+		}
+
+		if !match {
+			cmd.Println(aurora.BrightRed("Dataset name can only contain letters, numbers, underscores, and hyphens"))
+			os.Exit(1)
+		}
+
+		if strings.Contains(datasetName, "-") {
+			// warn that dataset name with hyphen should be quoted in queries
+			cmd.Println(aurora.BrightYellow(fmt.Sprintf("Dataset names with hyphens should be quoted in queries:\nSELECT * FROM \"%s\"", datasetName)))
```
phillipleblanc:
```suggestion
			cmd.Println(aurora.BrightYellow(fmt.Sprintf("Dataset names with hyphens should be quoted in queries:\ni.e. SELECT * FROM \"%s\"", datasetName)))
```

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #1147, opened 2024-04-20
Hot reload accelerated table on dataset update

### LINKED ISSUE #1084: Unloading and then reloading datasets shouldn't clear out data

ewgenius:
- extracted accelerated table creation into a separate method to call it before deregistering the dataset on update
- accelerated table registered only after it is fully loaded (or in 30 seconds timeout) - when accelerated dataset updated, old table is kept registered until new loaded

https://github.com/spiceai/spiceai/assets/827338/75e163bf-c5e5-457d-b8b6-3367f2091df0


phillipleblanc:
This logic should be encapsulated into the AcceleratedTable itself. When we need to hot reload the source, we should just get a reference to the existing AcceleratedTable and call a function that will manage the swap over to the new source.

crates/runtime/src/datafusion.rs
```diff
@@ -338,6 +355,97 @@ impl DataFusion {
...
+        tracing::debug!("Registering temp accelerated table {dataset:?}");
+
+        let id = uuid::Uuid::new_v4().to_string();
+        let temp_table_name = format!("__{}__{}", &dataset.name, &id[..8]);
+
+        self.ctx
+            .register_table(&temp_table_name, accelerated_table)
+            .context(UnableToRegisterTableToDataFusionSnafu)?;
+
+        tracing::info!("Waiting accelerated table {} to be ready...", dataset.name);
```
lukekim:
```suggestion
        tracing::info!("Verifying accelerated dataset {} readiness...", dataset.name);
```

ewgenius:
redone in #1195

PR #1149, opened 2024-04-21
Handle and trace errors in secret stores

### LINKED ISSUE #1070: SecretStore should not swallow errors while retrieving secrets
Existing trait signature does not allow underline Secret Store implementation to report error while getting a secret. 

https://github.com/spiceai/spiceai/blob/ccf8615eca63f483ae82c6194da28c255a548eb6/crates/secrets/src/lib.rs#L42
```
#[async_trait]
pub trait SecretStore {
    async fn get_secret(&self, secret_name: &str) -> Option<Secret>;
}
```

As a results those errors are ignored/shallowed that makes troubleshooting secrets related errors very complicated. That doesn't allow the caller to also distinguish between whether a secret isn't there, or there was some error getting the secret. We should update Api to allow passing error which should be processed in an centralized way (log ged as WARN for example)

```
async fn get_secret(&self, secret_name: &str) -> Option<Secret> {
        if let Ok(secret) = self.kubernetes_client.get_secret(secret_name).await {
            return Some(Secret::new(secret.clone()));
        }

        None
    }
```

ewgenius:
closes #1070 

Adds error handling and tracing for secret stores. Handles all suitable errors, swallowing them only when the secret is explicitly known to be none. Properly passes errors down.

Examples:

CleanShot 2024-04-22 at 23 59 14@2x

CleanShot 2024-04-22 at 23 59 34@2x

CleanShot 2024-04-23 at 00 00 00@2x

CleanShot 2024-04-23 at 00 02 02@2x

CleanShot 2024-04-23 at 00 01 02@2x

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -93,7 +93,7 @@ pub enum Error {
     #[snafu(display("Unable to load secrets for data connector: {data_connector}"))]
     UnableToLoadDataConnectorSecrets { data_connector: String },
 
-    #[snafu(display("Unable to get secret for data connector: {source}"))]
+    #[snafu(display("Unable to get secret for data connector {data_connector}: {source}"))]
     UnableToGetSecretForDataConnector {
         source: Box<dyn std::error::Error + Send + Sync>,
         data_connector: String,
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -385,7 +385,9 @@ impl Runtime {
         secrets_provider: &secrets::SecretsProvider,
         params: Arc<Option<HashMap<String, String>>>,
     ) -> Result<Arc<dyn DataConnector>> {
-        let secret = secrets_provider.get_secret(source).await;
+        // secret in general case is optional, so we don't fail if we can't get it
+        // Each data connector is responsible for checking the required secrets.
+        let secret = secrets_provider.get_secret(source).await.unwrap_or(None);
 
         match dataconnector::create_new_connector(source, secret, params).await {
             Some(dc) => dc.context(UnableToInitializeDataConnectorSnafu {
@@ -456,7 +458,10 @@ impl Runtime {
             .unwrap_or(format!("{accelerator_engine}_engine").to_lowercase());
 
         let secrets_provider_read_guard = secrets_provider.read().await;
-        let acceleration_secret = secrets_provider_read_guard.get_secret(&secret_key).await;
+        let acceleration_secret = secrets_provider_read_guard
+            .get_secret(&secret_key)
+            .await
+            .unwrap_or(None);
         drop(secrets_provider_read_guard);
 
         dataaccelerator::get_accelerator_engine(&accelerator_engine)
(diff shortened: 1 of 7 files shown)


crates/secrets/src/aws_secrets_manager.rs
```diff
@@ -110,29 +110,29 @@ impl SecretStore for AwsSecretsManager {
...
+                return Ok(None);
             }
             Err(err) => {
                 tracing::warn!(
                     "Failed to get secret {} from AWS Secrets Manager: {}",
                     secret_name,
                     err
                 );
-                return None;
+                return Ok(None);
```
phillipleblanc: Shouldn't this return an `Err`?

crates/secrets/src/keyring.rs
```diff
@@ -42,23 +42,54 @@ impl KeyringSecretStore {
...
+
+        let entry = match Entry::new(entry_key.as_str(), "spiced") {
+            Ok(entry) => entry,
+            Err(err) => {
+                tracing::warn!(
+                    "Failed to get secret entry {} from keyring store, {}",
+                    entry_key,
+                    err
+                );
+                return Ok(None);
```
phillipleblanc, line 67: Shouldn't this return an `Err`?

ewgenius: fixed

crates/secrets/src/aws_secrets_manager.rs
ewgenius: fixed

crates/runtime/src/lib.rs
```diff
@@ -93,6 +93,12 @@ pub enum Error {
     #[snafu(display("Unable to load secrets for data connector: {data_connector}"))]
     UnableToLoadDataConnectorSecrets { data_connector: String },
 
+    #[snafu(display("Unable to get secret for data connector: {source}"))]
```
```suggestion
    #[snafu(display("Unable to get secret for data connector {data_connector}: {source}"))]
```

crates/secrets/src/aws_secrets_manager.rs
```diff
@@ -36,10 +36,17 @@ pub enum Error {
     UnableToVerifyAwsIdentity {
         source: SdkError<GetCallerIdentityError>,
     },
-    #[snafu(display("Unable to parse as JSON: {}", source))]
+
+    #[snafu(display("Unable to parse AWS secret as JSON: {}", source))]
```
phillipleblanc:
```suggestion
    #[snafu(display("Unable to parse AWS secret as JSON: {source}"))]
```

crates/secrets/src/aws_secrets_manager.rs
```diff
@@ -36,10 +36,17 @@ pub enum Error {
...
     UnableToParseJson { source: serde_json::Error },
-    #[snafu(display("Invalid JSON format: JSON object is expected"))]
+
+    #[snafu(display("Invalid AWS secret value: JSON object is expected"))]
     InvalidJsonFormat {},
+
+    #[snafu(display("Unable to get AWS secret: {}", source))]
```
phillipleblanc:
```suggestion
    #[snafu(display("Unable to get AWS secret: {source}"))]
```

phillipleblanc (approved):

PR #1156, opened 2024-04-22
Extend PR check to build with no default features, and each individual feature

### LINKED ISSUE #1111: Enhancement: Add CI checks for `--no-default-features` and enabling each feature separately.
## Goal-State/What/Result

Add CI checks for running `cargo check` with `--no-default-features` and one for each feature we support, to avoid breaking them in the future.

## Why/Purpose

Ensures that users who want a slimmer build of Spice AI can build it.

phillipleblanc:
Extends the PR build to check that building without the default features and with each individual feature turned on works.

Closes #1111

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -63,6 +63,36 @@ jobs:
 
       - run: make test ci
 
+      - name: Build without default features
+        run: cargo check --no-default-features
+
+      - name: Build with only duckdb
+        run: cargo check --no-default-features --features duckdb
+
+      - name: Build with only postgres
+        run: cargo check --no-default-features --features postgres
+
+      - name: Build with only sqlite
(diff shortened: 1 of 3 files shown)


y-f-u (dismissed):

.github/workflows/pr.yml
```diff
@@ -63,6 +63,36 @@ jobs:
 
       - run: make test ci
 
+      - name: Build without default features
+        run: cargo check --no-default-features
+
+      - name: Build with only duckdb
+        run: cargo check --no-default-features --features duckdb
```
y-f-u: can this be in a matrix?

phillipleblanc: Putting it into a matrix will run a separate job per matrix item, whereas in theory this can re-use the existing build cache to go faster.

lukekim (approved):

PR #1165, opened 2024-04-22
Update postgres data connector errors

Sevenannn:
Sample of fixing Postgres data connector errors. All intermediate error messages are remove. Only the top level runtime error and root source of error will be kept.

The error follows the structure:
`{Top Level Runtime Error Message}{Source Error Message}`

### Examples
1.
- Config
```yaml
    name: my_dataset
    acceleration:
      engine: postgres
      engine_secret: pg_backend
      params:
        pg_host: localhost
        pg_port: localhost
        pg_db: my_databsda123ase
        pg_user: my_user
        pg_pass_key: my_secret
```
- From:
`Failed to initialize data connector for dataset my_dataset: Unable to attach data connector spiceai: Unable to create dataset acceleration: Acceleration creation failed: Unable to create table: External error: Unable to create Postgres connection pool: ConnectionPoolError: invalid connection string: invalid value for option `port``
- To:
`Failed to initialize data connector for dataset my_dataset: Unable to attach data connector spiceai: Unable to create dataset acceleration: External error: ConnectionPoolError: invalid connection string: invalid value for option `port`

2.
- Config
```yaml
    name: my_dataset
    acceleration:
      engine: postgres
      engine_secret: pg_backend
      params:
        pg_host: localhost
        pg_port: 5432
        pg_db: my_database
        pg_user: invalid
```
- From:
`Failed to initialize data connector for dataset my_dataset: Unable to attach data connector spiceai: Unable to create dataset acceleration: Acceleration creation failed: Unable to create table: External error: Unable to create Postgres connection pool: ConnectionPoolRunError: Timed out in bb8`
- To:
` Failed to initialize data connector for dataset my_dataset: Unable to attach data connector spiceai: Unable to create dataset acceleration: External error: ConnectionPoolRunError, possible invalid Postgres parameters: Timed out in bb8`

3.
- Config
```yaml
  - from: postgres:eth_recent_blocks
    name: my_dataset
    params:
      pg_host: localhost
      pg_port: localhost
      pg_db: my_databsda123ase
      pg_user: my_user
      pg_pass_key: my_secret
```
- From
`Failed to get data connector from source for dataset my_dataset: Unable to initialize data connector postgres: Unable to create Postgres connection pool: ConnectionPoolError: invalid connection string: invalid value for option `port``
- To
`Failed to get data connector from source for dataset my_dataset: Unable to initialize data connector postgres: ConnectionPoolError: invalid connection string: invalid value for option `port``

4.
- Config
```yaml
  - from: postgres:eth_recent_blocks
    name: my_dataset
    params:
      pg_host: localhost
      pg_port: 5432
      pg_user: invalid
      pg_pass_key: my_secret
```
- From
`Failed to get data connector from source for dataset my_dataset: Unable to initialize data connector postgres: Unable to create Postgres connection pool: ConnectionPoolRunError: Timed out in bb8`
- To
`Failed to get data connector from source for dataset my_dataset: Unable to initialize data connector postgres: ConnectionPoolRunError, possible invalid Postgres parameters: Timed out in bb8`


digadeesh:
What is the top level error message in this case?

Sevenannn:
@digadeesh The top level error is the errors from [runtime](https://github.com/spiceai/spiceai/blob/d3defed3dd1c455f0313cb1772e87194450fd10a/crates/runtime/src/lib.rs). In the above examples, text before {Top Level Runtime Error Message} are top level error messages. Text between {Top Level Runtime Error Message} and {Source Error Message} is source error message.

crates/data_components/src/postgres.rs
```diff
@@ -65,7 +65,7 @@ pub enum Error {
         source: db_connection_pool::dbconnection::GenericError,
     },
 
-    #[snafu(display("Unable to create Postgres connection pool: {source}"))]
+    #[snafu(display("{source}"))]
```
phillipleblanc, line 68: Let's use the Snafu `transparent` attribute instead of this - its more idiomatic: https://docs.rs/snafu/latest/snafu/derive.Snafu.html#delegating-to-the-underlying-error

phillipleblanc, line 68:
```suggestion
    #[snafu(transparent)]
```

Sevenannn: Thanks for the suggestion! I swapped some errors with transparent in the code, and I think there are a few things to consider in whether to use transparent, and I'd love to hear your thoughts on this:
* Transparent disables context information. Error no longer needs .context(mycontextSnafu), but instead automatically gets recognized. Could be hard to debug since the context information is swallowed. (See the changes in postgresconn)
* When adding new errors that have the same type of source to the enum Error, we have to convert transparent back to display to distinguish the context.
* From the documentation: Snafu transparent is useful when aggregating smaller errors that already completely explain the failure. https://github.com/shepmaster/snafu/blob/main/CHANGELOG.md
* Since transparent can only be used on error context where error sources uniquely distinguish context, we will need to mix the use of `transparent` and `display` in existing `enum Error`, which could be a bit messy.

digadeesh:
I think we should rework this a bit.  We want to improve these such that the errors that surface in the runtime are clear and actionable.  In this case, there is some noise reduced, but I don't think we're at the actionable part yet.  Let's move this to draft while we work through it.

crates/data_components/src/postgres.rs
phillipleblanc: Ok, I changed my mind - no more `transparent` :)

phillipleblanc: The context information is what makes Snafu so great, so if we lose that - then its not worth it.

Sevenannn:
New solutions available in [Enhancement: User Friendly Error messages, PostgreSQL data connector #1229](https://github.com/spiceai/spiceai/issues/1229)

PR #1186, opened 2024-04-23
Properly support Spark Connect filter pushdown

phillipleblanc:
Fixes a bug in the Databricks Data Connector where filters were not being properly pushed down, resulting in the equivalent of a `SELECT * FROM big_table` and then trying to filter in-memory, which was causing some timeouts and inefficient queries.

I attempted to use the in-built DataFusion unparser, but ran into an issue with column quoting and I raised a PR to fix it upstream: https://github.com/apache/datafusion/pull/10198

Once that PR lands and a new version is released, we can start removing our own `expr::to_sql` method.


ablyler:
I've confirmed that is fixes the issue I've seen. Thanks for the quick fix!

digadeesh (approved):

PR #1187, opened 2024-04-24
Convert temporal column and refresh period to datafusion expr

### LINKED ISSUE #1175: Generate sql filter based on dataset schema and temporal column and format for refresh period

y-f-u:
Add a timestamp converter to simplify the logic.
Convert the refresh period logic to expr.

There will be following PR to enable the filter when fetching the data.

Fixes #1175 

Edit:
With the merge of #1188 . This allows the period to work properly

```
    name: taxi_trips
    time_column: tpep_pickup_datetime
    acceleration:
      enabled: true
      refresh_interval: 10m
      refresh_period: 1000w
      refresh_sql: |
        SELECT * FROM taxi_trips WHERE passenger_count = 1;
 ```

### DIFF
--- a/crates/runtime/src/accelerated_table.rs
+++ b/crates/runtime/src/accelerated_table.rs
@@ -203,9 +203,8 @@ impl AcceleratedTable {
 
         let mut interval_timer = tokio::time::interval(retention.check_interval);
 
-        let var_name = Some(time_column.clone());
         let Some(timestamp_filter_converter) =
-            TimestampFilterConvert::create(field, retention.time_format, var_name)
+            TimestampFilterConvert::create(field, retention.time_format, Some(time_column.clone()))
         else {
             tracing::error!("[retention] Failed to get the expression time format for {time_column}, check schema and time format");
             return;
--- a/crates/runtime/src/accelerated_table.rs
+++ b/crates/runtime/src/accelerated_table.rs
@@ -79,6 +79,12 @@ enum AccelerationRefreshMode {
     Append,
 }
 
+fn validate_refresh_period(refresh: &Refresh, dataset: &str) {
+    if refresh.period.is_some() && refresh.time_column.is_none() {
+        tracing::warn!("No time_column is provided, refresh_period will be ignored for {dataset}");
+    }
+}
+
 #[allow(clippy::too_many_arguments)]
 impl AcceleratedTable {
     pub async fn new(
@@ -103,6 +109,7 @@ impl AcceleratedTable {
             }
         };
 
+        validate_refresh_period(&refresh, &dataset_name);
         let refresh_handle = tokio::spawn(Self::start_refresh(
             dataset_name.clone(),
             Arc::clone(&federated),
@@ -356,20 +363,10 @@ impl AcceleratedTable {
                     }
                 }
                 AccelerationRefreshMode::Full(receiver) => {
-                    let filter_converter = match (refresh.period, refresh.time_column) {
-                        (None, Some(_) | None) => {
-                            None
-                        },
-                        (Some(_), None) => {
-                            tracing::warn!("[refresh] No time_column is provided, refresh_period will be ignored");
-                            None
-                        }
-                        (Some(_), Some(column)) => {
-                            let schema = federated.schema();
-                            let field = schema.column_with_name(column.as_str());
-                            TimestampFilterConvert::create(field, Some(column), refresh.time_format)
-                        }
-                    };
+                    let schema = federated.schema();
+                    let column = refresh.time_column.clone().unwrap_or_default();
+                    let field = schema.column_with_name(column.as_str());
+                    let filter_converter = TimestampFilterConvert::create(field, refresh.time_column, refresh.time_format);
 
                     let mut refresh_stream = ReceiverStream::new(receiver);
 
--- a/crates/runtime/src/accelerated_table.rs
+++ b/crates/runtime/src/accelerated_table.rs
@@ -2,19 +2,16 @@ use std::time::{SystemTime, SystemTimeError, UNIX_EPOCH};
 use std::{any::Any, sync::Arc, time::Duration};
 
 use arrow::array::UInt64Array;
-use arrow::datatypes::{DataType, SchemaRef};
+use arrow::datatypes::SchemaRef;
 use async_stream::stream;
 use async_trait::async_trait;
 use data_components::delete::get_deletion_provider;
 use datafusion::common::OwnedTableReference;
 use datafusion::error::Result as DataFusionResult;
 use datafusion::execution::context::SessionState;
-use datafusion::logical_expr::{
-    binary_expr, cast, col, lit, Operator, TableProviderFilterPushDown,
-};
+use datafusion::logical_expr::{Operator, TableProviderFilterPushDown};
 use datafusion::physical_plan::union::UnionExec;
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: crates/runtime/src/accelerated_table.rs [hunk 1]
```diff
@@ -400,64 +427,91 @@ impl AcceleratedTable {
...
-                    if time_format == TimeFormat::UnixMillis {
-                        scale = 1000;
+#[derive(Clone, Debug)]
+struct TimestampFilterConvert {
+    time_column: String,
+    time_format: ExprTimeFormat,
+}
+
+#[allow(clippy::needless_pass_by_value)]
+impl TimestampFilterConvert {
```
phillipleblanc: This seems unrelated to the accelerated_table - let's put this in a separate module

crates/runtime/src/accelerated_table.rs
```diff
@@ -400,64 +427,91 @@ impl AcceleratedTable {
...
+    fn create(
+        field: Option<(usize, &arrow::datatypes::Field)>,
```
phillipleblanc: Seems like the index information in the tuple isn't used? `(usize, &arrow::datatypes::Field)` it would be simpler just to pass the arrow field here and strip the index.

crates/runtime/src/accelerated_table.rs
```diff
@@ -372,16 +371,44 @@ impl AcceleratedTable {
...
+
                     let mut refresh_stream = ReceiverStream::new(receiver);
 
                     while refresh_stream.next().await.is_some() {
-                        tracing::info!("Refreshing data for {dataset_name}");
+                        tracing::info!("[refresh] Refreshing data for {dataset_name}");
                         status::update_dataset(&dataset_name, status::ComponentStatus::Refreshing);
                         let timer = TimeMeasurement::new("load_dataset_duration_ms", vec![("dataset", dataset_name.clone())]);
-                        let all_data = match get_all_data(&mut ctx, OwnedTableReference::bare(dataset_name.clone()), Arc::clone(&federated), refresh_sql.clone()).await {
+                        let  filters = match (refresh.period, filter_converter.clone()){
```
```suggestion
                        let filters = match (refresh.period, filter_converter.clone()){
```

crates/runtime/src/accelerated_table.rs
```diff
@@ -372,16 +371,44 @@ impl AcceleratedTable {
...
+                        (Some(_), None) => {
+                            tracing::warn!("[refresh] No time_column is provided, refresh_period will be ignored");
+                            None
+                        }
+                        (Some(_), Some(column)) => {
+                            let schema = federated.schema();
+                            let field = schema.column_with_name(column.as_str());
+                            TimestampFilterConvert::create(field, Some(column), refresh.time_format)
+                        }
+                    };
```
phillipleblanc: This feels unnecessarily complex. Is there a simpler way to achieve what we want?

crates/runtime/src/accelerated_table.rs
```diff
@@ -372,16 +371,44 @@ impl AcceleratedTable {
...
+                                let Ok(timestamp) = get_timestamp(start) else {
+                                    tracing::error!("[refresh] Failed to get timestamp");
+                                    continue;
+                                };
+                                vec![converter.convert(timestamp, Operator::Gt)]
+                            },
+                            _ => vec![],
+                        };
+
+                        let all_data = match get_all_data(&mut ctx, OwnedTableReference::bare(dataset_name.clone()), Arc::clone(&federated), refresh.sql.clone(), filters).await {
```
```suggestion
                        let data = match get_data(&mut ctx, OwnedTableReference::bare(dataset_name.clone()), Arc::clone(&federated), refresh.sql.clone(), filters).await {
```

This isn't technically all data anymore.

crates/runtime/src/accelerated_table.rs
```diff
@@ -400,64 +427,91 @@ impl AcceleratedTable {
...
+        Some(Self {
+            time_column,
+            time_format,
+        })
+    }
+
+    #[allow(clippy::cast_possible_wrap)]
+    fn convert(&self, timestamp: u64, op: Operator) -> Expr {
+        let format = self.time_format.clone();
+        let time_column: &str = &self.time_column.clone();
```
phillipleblanc, line 489:
```suggestion
        let time_column: &str = self.time_column.as_str;
```

crates/runtime/src/accelerated_table.rs
```diff
@@ -400,64 +427,91 @@ impl AcceleratedTable {
...
+        let expr_time_format = format.clone();
+        match expr_time_format {
```
```suggestion
        let time_column: &str = self.time_column.as_str();
        match &self.time_format {
```

crates/runtime/src/accelerated_table.rs
y-f-u: good point.

crates/runtime/src/accelerated_table.rs
y-f-u: it might be. The main complication comes from informative information to the user if they have set `refresh_period` and expect it is working.

crates/runtime/src/accelerated_table.rs
y-f-u: Surprised clippy was accepting that.

crates/runtime/src/accelerated_table.rs
y-f-u: Yeah It is not used. It was mainly for not doing extra logic to split the tuple in the caller spaces as there are two callers. And the tuple kinda make sense being together.

I had the other version too before this and opted for this one.

crates/runtime/src/accelerated_table.rs
phillipleblanc: That type of validation/warning should happen when we are loading the spicepod imo

y-f-u: moved to `create`

sgrebnov (dismissed):

crates/runtime/src/datafusion/filter_converter.rs
```diff
@@ -0,0 +1,104 @@
...
+        Some(Self {
+            time_column,
+            time_format,
+        })
+    }
+
+    #[allow(clippy::cast_possible_wrap)]
+    pub(crate) fn convert(&self, timestamp: u64, op: Operator) -> Expr {
+        let time_column: &str = &self.time_column.clone();
+        match self.time_format.clone() {
```
phillipleblanc:
```suggestion
        match &self.time_format {
```

crates/runtime/src/datafusion/filter_converter.rs
```diff
@@ -0,0 +1,104 @@
...
+
+        Some(Self {
+            time_column,
+            time_format,
+        })
+    }
+
+    #[allow(clippy::cast_possible_wrap)]
+    pub(crate) fn convert(&self, timestamp: u64, op: Operator) -> Expr {
+        let time_column: &str = &self.time_column.clone();
```
phillipleblanc:
```suggestion
        let time_column: &str = self.time_column.as_ref();
```

crates/runtime/src/datafusion/filter_converter.rs
phillipleblanc: I pushed a commit for this

crates/runtime/src/datafusion/filter_converter.rs
phillipleblanc: I pushed a commit for this

phillipleblanc (approved):

PR #1195, opened 2024-04-24
Hot reload accelerated table on dataset update

### LINKED ISSUE #1084: Unloading and then reloading datasets shouldn't clear out data

ewgenius:
closes #1084

Refactored dataset update logic:
For accelerated datasets (only memory mode), spice runtime performs hot reload:

- create new accelerated table and wait until it loaded with data, while keeping old dataset registered in datafusion
- once new table successfully loaded, unregister old dataset and register new one
- in case of failure it falls back to regular reload

https://github.com/spiceai/spiceai/assets/827338/5e3838cb-4be4-44e7-82b7-90ae930fc551

### DIFF
--- a/crates/runtime/src/accelerated_table.rs
+++ b/crates/runtime/src/accelerated_table.rs
@@ -27,8 +27,8 @@ use tokio::task::JoinHandle;
 use tokio::time::interval;
 use url::Url;
 
-use tokio::sync::mpsc;
 use tokio::sync::mpsc::Receiver;
+use tokio::sync::{mpsc, oneshot};
 use tokio_stream::wrappers::ReceiverStream;
 
 use crate::datafusion::{Refresh, Retention};
@@ -68,7 +68,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 // An accelerated table consists of a federated table and a local accelerator.
 //
 // The accelerator must support inserts.
-pub(crate) struct AcceleratedTable {
+pub struct AcceleratedTable {
     accelerator: Arc<dyn TableProvider>,
     federated: Arc<dyn TableProvider>,
     refresh_trigger: Option<mpsc::Sender<()>>,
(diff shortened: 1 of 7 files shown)


digadeesh (dismissed):

crates/runtime/src/accelerated_table.rs
```diff
@@ -68,7 +68,7 @@ pub type Result<T> = std::result::Result<T, Error>;
 // An accelerated table consists of a federated table and a local accelerator.
 //
 // The accelerator must support inserts.
```
phillipleblanc, line 69: Let's update the comment to indicate what the oneshot channel is doing.

crates/runtime/src/datafusion.rs
```diff
@@ -145,8 +174,19 @@ impl DataFusion {
         &mut self,
         dataset: impl Borrow<Dataset>,
         table: Table,
+        accelerated_table: Option<AcceleratedTable>,
```
phillipleblanc: This should be a new enum for `Table`

crates/runtime/src/datafusion/schema.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use async_trait::async_trait;
+use dashmap::DashMap;
+use datafusion::{
+    catalog::schema::SchemaProvider,
+    datasource::TableProvider,
+    error::{DataFusionError, Result},
+};
+
+// Copy of default MemorySchemaProvider, with allowed table register overwrites
```
phillipleblanc:
```suggestion
// Copy of default MemorySchemaProvider that allows `register_table` to atomically overwrite any existing tables
```

crates/runtime/src/lib.rs
```diff
@@ -372,11 +385,31 @@ impl Runtime {
     }
 
     pub async fn update_dataset(&self, ds: &Dataset, all_datasets: &[Dataset]) {
-        tracing::info!("Updating dataset: {}...", &ds.name);
         status::update_dataset(&ds.name, status::ComponentStatus::Refreshing);
         if let Ok(connector) = self.load_dataset_connector(ds, all_datasets).await {
+            if ds.is_file_accelerated() {
+                tracing::warn!("File accelerated datasets doesn't support hot reload, falling back to full dataset reload");
```
```suggestion
                tracing::warn!("File accelerated datasets currently don't support hot reload, falling back to full dataset reload");
```

crates/runtime/src/datafusion.rs
ewgenius: thanks, updated

crates/runtime/src/accelerated_table.rs
ewgenius: Added a comment

phillipleblanc (approved):

PR #1198, opened 2024-04-24
Clickhouse refactor connection code and set secure option

gloomweaver:
- Refactor setting up connection to clickhouse
- Support `clickhouse_secure` parameter for setting up ssl connection
- Tested against clickhouse cloud instance


crates/db_connection_pool/src/clickhousepool.rs
```diff
@@ -109,6 +81,84 @@ impl ClickhouseConnectionPool {
...
+    if let Some(clickhouse_connection_string) = get_secret_or_param(
+        params,
+        secret,
+        "clickhouse_connection_string_key",
+        "clickhouse_connection_string",
+    ) {
+        let mut new_options = Options::from_str(&clickhouse_connection_string)
+            .context(InvalidConnectionStringSnafu)?;
+        if !clickhouse_connection_string.contains("connection_timeout") {
+            // Default timeouf of 500ms is not enough in some cases.
```
digadeesh:
```suggestion
            // Default timeout of 500ms is not enough in some cases.
```

digadeesh: Was 5 seconds picked at random?

crates/db_connection_pool/src/clickhousepool.rs
```diff
@@ -109,6 +81,84 @@ impl ClickhouseConnectionPool {
...
+        }
+        let connection_string = format!(
+            "tcp://{}:{}@{}:{}/{}",
+            connection_opts[0],
+            connection_opts[1],
+            connection_opts[2],
+            connection_opts[3],
+            connection_opts[4]
+        );
+        // Default timeouf of 500ms is not enough
```
digadeesh:
```suggestion
        // Default timeout of 500ms is not enough
```

digadeesh: Also, copy/paste code.  At minimum this should be a constant, but it's a little peculiar to format it to a string, just so you can parse it back into options.  Can we refactor this to be a little cleaner?

crates/db_connection_pool/src/clickhousepool.rs
gloomweaver: Felt like a good default value.
Will change to default specified by clickhouse https://clickhouse.com/docs/en/operations/settings/settings#connect-timeout-receive-timeout-send-timeout
Also will create a fix for clickhouse-rs crate

crates/db_connection_pool/src/clickhousepool.rs
gloomweaver: It's a quirk of the clickhouse-rs crate. Options struct doesn't expose all necessary methods to construct manually.
Will open up a pr to clickhouse-rs to fix those API problems.

crates/db_connection_pool/src/clickhousepool.rs
```diff
@@ -109,6 +92,91 @@ impl ClickhouseConnectionPool {
...
+            .fail()?,
+        }
+        let connection_string = format!(
+            "tcp://{}:{}@{}:{}/{}",
+            connection_opts[0],
+            connection_opts[1],
+            connection_opts[2],
+            connection_opts[3],
+            connection_opts[4]
+        );
```
ewgenius: This doesn't look durable enough, and it is hard to follow which parameter goes in each position. I would prefer to explicitly retrieve each parameter rather than relying on loops.

crates/db_connection_pool/src/clickhousepool.rs
```diff
@@ -42,8 +47,20 @@ pub enum Error {
         source: clickhouse_rs::errors::Error,
     },
 
-    #[snafu(display("Invalid parameter: {parameter_name}"))]
-    InvalidParameterError { parameter_name: String },
+    #[snafu(display("ParametersEmptyError"))]
+    ParametersEmptyError {},
+
+    #[snafu(display("Required parameter was not provided: {parameter_name}"))]
+    NotEnoughParametersForConnection { parameter_name: String },
```
ewgenius:
```suggestion
    #[snafu(display("Missing required parameter: {parameter_name}"))]
    MissingRequiredParameterForConnection { parameter_name: String },
```

crates/db_connection_pool/src/clickhousepool.rs
ewgenius: Something like

```rust

let user = params.get("clickhouse_user")
     .ok_or_else(|| MissingRequiredParameter { name: "clickhouse_user" });

let host = params.get("clickhouse_host")
     .ok_or_else(|| MissingRequiredParameter { name: "clickhouse_host" });

...

let connection_string = format!("tcp://{user}:{host}...");

```

crates/db_connection_pool/src/clickhousepool.rs
```diff
@@ -109,6 +92,91 @@ impl ClickhouseConnectionPool {
...
+    let secure = params
+        .get("clickhouse_secure")
+        .map(|s| s.parse::<bool>())
+        .transpose()
+        .context(InvalidSecureParameterValueSnafu {
+            parameter_name: "clickhouse_secure".to_string(),
+        })?;
+    options = options.secure(secure.unwrap_or(true));
+
+    if let Some(clickhouse_sslrootcert) = params.get("clickhouse_sslrootcert") {
```
ewgenius: Where is `clickhouse_sslrootcert` used? Should it be passed to options?

crates/db_connection_pool/src/clickhousepool.rs
```diff
@@ -109,6 +92,91 @@ impl ClickhouseConnectionPool {
...
+            connection_opts[0],
+            connection_opts[1],
+            connection_opts[2],
+            connection_opts[3],
+            connection_opts[4]
+        );
+        // Default timeout of 500ms is not enough
+        let new_options = Options::from_str(&connection_string)
+            .context(InvalidConnectionStringSnafu)?
+            .connection_timeout(DEFAULT_CONNECTION_TIMEOUT);
```
ewgenius, line 150: I think that should be also configurable through options - `params.clickhouse_connection_timeout`

crates/db_connection_pool/src/clickhousepool.rs
gloomweaver: Fixed

crates/db_connection_pool/src/clickhousepool.rs
gloomweaver: Removed, currently there is a problem this crate.
It doesn't expose the required code.

crates/db_connection_pool/src/clickhousepool.rs
gloomweaver: Added

crates/db_connection_pool/src/clickhousepool.rs
```diff
@@ -42,8 +47,20 @@ pub enum Error {
         source: clickhouse_rs::errors::Error,
     },
 
-    #[snafu(display("Invalid parameter: {parameter_name}"))]
-    InvalidParameterError { parameter_name: String },
+    #[snafu(display("ParametersEmptyError"))]
```
phillipleblanc:
```suggestion
    #[snafu(display("No parameters specified"))]
```

ewgenius (approved):

PR #1204, opened 2024-04-25
Feature: Support ODBC reads using unixodbc

mach-kernel:
###### What?

This implements a connector for the [unixodbc](https://www.unixodbc.org/) driver using the [odbc-api](https://crates.io/crates/odbc-api) and [arrow_odbc](https://docs.rs/arrow-odbc/latest/arrow_odbc/) crates (being 90% of the hard part). You can use a host-configured DSN or connection string ([docs](https://learn.microsoft.com/en-us/sql/integration-services/import-export-data/connect-to-an-odbc-data-source-sql-server-import-and-export-wizard?view=sql-server-ver16)).

<details>
<summary>
Demo
</summary>
  
To query a Databricks SQL warehouse, [configure and install the Simba ODBC driver](https://www.databricks.com/spark/odbc-drivers-download) with this spice pod definition:

```yaml
from: odbc:default.foo_meta
name: foo_meta
mode: read
time_column: created_at
acceleration:
  enabled: true
  engine: arrow
  mode: memory
  refresh_mode: full
  refresh_interval: 5m
  # or even `select * from foo_meta where day > now() - interval '7 days'`
  refresh_sql: |
    select * from foo_meta where tenant_id = 'dstancu'
params:
  mode: odbc
  url: Host=dstancu-is-very-cool.cloud.databricks.com;Port=443;transportMode=http;ssl=1;AuthMech=3;httpPath=/sql/1.0/warehouses/snipsnap;Driver={Simba Spark ODBC Driver};UID=token;PWD=snipsnap;ThriftTransport=2

```

Describe a table:
```sql
sql> describe foo_meta;
+-------------+------------------------------+-------------+
| column_name | data_type                    | is_nullable |
+-------------+------------------------------+-------------+
| tenant_id   | Utf8                         | YES         |
| day         | Date32                       | YES         |
| document_id | Utf8                         | YES         |
| created_at  | Timestamp(Microsecond, None) | YES         |
| updated_at  | Timestamp(Microsecond, None) | YES         |
| tags        | Utf8                         | YES         |
| folders     | Utf8                         | YES         |
+-------------+------------------------------+-------------+
```

Basic query (tiny, 2400 rows, usually takes ~2s wall time):
```sql
Query took: 0.07819675 seconds. 500/2466 rows displayed.
sql> select * from foo_meta where tenant_id = 'dstancu'
```

Agg on materialized arrow set (same story, upto ~seconds sans Arrow):
```sql
sql> select max(created_at) from foo_meta where tenant_id = 'dstancu'
+----------------------------+
| MAX(foo_meta.created_at)  |
+----------------------------+
| 2024-04-24T14:59:15.548080 |
+----------------------------+
Query took: 0.047860917 seconds. 1/1 rows displayed.
```
</details>

###### Building!

Assuming `aarch64` Mac. Install the `unixodbc` driver for your distribution, and validate the installation with pkg-config and odbcinst:

```bash
$ brew install unixodbc

$ pkg-config --libs odbc
-L/opt/homebrew/Cellar/unixodbc/2.3.12/lib -lodbc

$ odbcinst -j
unixODBC 2.3.12
DRIVERS............: /opt/homebrew/etc/odbcinst.ini
SYSTEM DATA SOURCES: /opt/homebrew/etc/odbc.ini
FILE DATA SOURCES..: /opt/homebrew/etc/ODBCDataSources
USER DATA SOURCES..: /Users/dstancu/.odbc.ini
SQLULEN Size.......: 8
SQLLEN Size........: 8
SQLSETPOSIROW Size.: 8
```

Using the `pkg-config` discovered path, override your Mac's `DYLD_LIBRARY_PATH` (`LD_LIBRARY_PATH` for non-Mac):

```bash
DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:/opt/homebrew/Cellar/unixodbc/2.3.12/lib cargo run --bin spiced
```

###### Configuring your ODBC driver

Written in context of the Databricks Simba adapter, but is useful for arbitrary adapters. After installing Simba, it should update `/etc/odbcinst.ini`:

```
[ODBC Drivers]
Simba Spark ODBC Driver = Installed

[Simba Spark ODBC Driver]
Driver = /Library/simba/spark/lib/libsparkodbc_sb64-universal.dylib
```

You will have to move this to the Homebrew config file:

```bash
cat /etc/odbcinst.ini >> /opt/homebrew/etc/odbcinst.ini
```

The search path defaults to your platform's standard prefix (i.e. not Homebrew's `/opt/homebrew`), so you will get an error about not being able to load `libodbcinst.dylib`. We have to update the driver's configuration:

(where the path to the library is in the `pkg-config` provided path):
```bash
echo "ODBCInstLib=/opt/homebrew/Cellar/unixodbc/2.3.12/lib/libodbcinst.dylib" >> /Library/simba/spark/lib/simba.sparkodbc.ini
```

To run the ODBC parameter binding test, you will also need:

```bash
$ brew install sqliteodbc
$ cat << EOF >> /opt/homebrew/etc/odbcinst.ini
[SQLite]
Description=SQLite ODBC Driver
Setup=/opt/homebrew/Cellar/sqliteodbc/0.99991/lib/libsqlite3odbc.so
Driver=/opt/homebrew/Cellar/sqliteodbc/0.99991/lib/libsqlite3odbc.so
Threading=2
EOF
```


crates/db_connection_pool/src/dbconnection/odbcconn.rs
```diff
@@ -0,0 +1,231 @@
...
+    use odbc_api::IntoParameter;
+
+    use crate::odbcpool::ODBCPool;
+    use crate::Result;
+    use std::error::Error;
+    use std::str;
+
+    use super::*;
+
+    // This test crudely validates that parameters are being received by the ODBC driver
```
mach-kernel, line 241: Customize as needed, will figure out the CI piece in a bit.

```
DYLD_LIBRARY_PATH=/opt/homebrew/Cellar/unixodbc/2.3.12/lib cargo test --features odbc --package db_connection_pool --lib -- dbconnection::odbcconn::tests --show-output
```

crates/sql_provider_datafusion/src/lib.rs
```diff
@@ -132,13 +138,13 @@ impl<T, P> TableProvider for SqlTable<T, P> {
...
-        let mut filter_push_down = vec![];
-        for filter in filters {
-            match expr::to_sql(filter) {
-                Ok(_) => filter_push_down.push(TableProviderFilterPushDown::Exact),
-                Err(_) => filter_push_down.push(TableProviderFilterPushDown::Unsupported),
-            }
-        }
+        let filter_push_down: Vec<TableProviderFilterPushDown> = filters
+            .iter()
+            .map(|f| match expr::to_sql_with_engine(f, Some(Engine::ODBC)) {
```
phillipleblanc:
```suggestion
            .map(|f| match expr::to_sql_with_engine(f, self.engine) {
```

crates/data_components/Cargo.toml
```diff
@@ -43,6 +43,8 @@ arrow_sql_gen = { path = "../arrow_sql_gen", optional = true }
 rusqlite = { workspace = true, optional = true }
 tokio-rusqlite = { workspace = true, optional = true }
 spark-connect-rs =  { git = "https://github.com/edmondop/spark-connect-rs.git", rev = "24b8a437af41989bf7594e1606c37f30c134bba2", features = ["tls"], optional = true }
+odbc-api = { version = "7.0.0", optional = true }
+arrow-odbc = { version = "8.3.0", optional = true }
```
gloomweaver: Better to add this dependencies to workspace Cargo.toml, then
`arrow-odbc = { workspace = true, optional = true }`

crates/db_connection_pool/Cargo.toml
```diff
@@ -30,6 +30,9 @@ mysql_async = { workspace = true, optional = true }
 ns_lookup = { path = "../ns_lookup" }
 native-tls = { version = "0.2.11", optional = true }
 postgres-native-tls = { version = "0.5.0", optional = true }
+odbc-api = { version = "7.0.0", optional = true }
+arrow-odbc = { version = "8.3.0", optional = true }
```
gloomweaver: Same here

crates/db_connection_pool/src/dbconnection/odbcconn.rs
```diff
@@ -0,0 +1,231 @@
...
+    ODBCAPIError { source: odbc_api::Error },
+}
+
+pub struct ODBCConnection<'a> {
+    pub conn: Arc<Mutex<Connection<'a>>>,
+}
+
+impl<'a> DbConnection<Connection<'a>, ODBCParameter> for ODBCConnection<'a>
+where
+    'a: 'static,
```
gloomweaver, line 84: Why is `'static` necessary here?

crates/db_connection_pool/src/dbconnection/odbcconn.rs
```diff
@@ -0,0 +1,231 @@
...
+        params: &[ODBCParameter],
+    ) -> Result<SendableRecordBatchStream> {
+        let cxn = self.conn.lock().await;
+        let mut prepared = cxn.prepare(sql)?;
+        let schema = Arc::new(arrow_schema_from(&mut prepared)?);
+        let mut statement = prepared.into_statement();
+
+        bind_parameters(&mut statement, params);
+
+        let cursor = unsafe {
```
gloomweaver, line 137: Can it be done without using `unsafe`?

crates/db_connection_pool/src/dbconnection/odbcconn.rs
```diff
@@ -0,0 +1,231 @@
...
+        Ok(row_count
+            .unwrap()
+            .try_into()
+            .expect("Must obtain row count"))
+    }
+}
+
+fn bind_parameters(statement: &mut StatementImpl, params: &[ODBCParameter]) {
+    for (i, param) in params.iter().enumerate() {
+        unsafe {
```
gloomweaver, line 216: Same here. Why is `unsafe` necessary here?

crates/sql_provider_datafusion/src/lib.rs
mach-kernel: Oops! 😃

crates/db_connection_pool/src/dbconnection/odbcconn.rs
mach-kernel: [odbc-api reference for parameters](https://docs.rs/odbc-api/7.0.0/odbc_api/parameter/index.html)

[`bind_input_parameter`](https://docs.rs/odbc-api/7.0.0/odbc_api/handles/trait.Statement.html#method.bind_input_parameter) is unsafe. The crate has a safe API for doing this (like we are by exposing a safe function that leverages unsafe code) but I cannot use it given the constraints of `AsyncDbConnection` and also `async_trait`.

```rust
pub trait AsyncDbConnection<T, P>: DbConnection<T, P> + Sync {
    ...
    async fn query_arrow(&self, sql: &str, params: &[P]) -> Result<SendableRecordBatchStream>;
    async fn execute(&self, sql: &str, params: &[P]) -> Result<u64>;
}
```

Our input parameters need to be sync to work with `async_trait`:
```rust
// initially: pub type ODBCParameter = Box<dyn InputParameter + Sync>
pub trait ODBCSyncParameter: InputParameter + Sync {...}
```

But this `(params: &[Box<dyn InputParameter + Sync>])` doesn't work for [`execute`](https://docs.rs/odbc-api/7.0.0/odbc_api/struct.Connection.html#method.execute) (the 'safe' API), and I have no way to rebox the ref to cast to `&dyn InputParameter`:

```
the trait bound `[Box<dyn InputParameter + Sync>]: odbc_api::parameter_collection::InputParameterCollection` is not satisfied
the trait `odbc_api::parameter_collection::InputParameterCollection` is implemented for `[T]`
required for `&[Box<dyn InputParameter + Sync>]` to implement `ParameterCollectionRef`
```

The odbc-api [IntoParameter](https://docs.rs/odbc-api/7.0.0/odbc_api/trait.IntoParameter.html) trait wants to consume an instance to lift into `InputParameter` (common pattern for FFI glue for data going out). I think it would be worthwhile to think about passing params by value. It would let us use the safe API directly:

```rust
async fn query_arrow(&self, sql: &str, params: [P]) -> Result<SendableRecordBatchStream>;
async fn execute(&self, sql: &str, params: [P]) -> Result<u64>;
```

I am happy to work with you to figure out a cleaner way around this, I spent a bunch of time here but unfortunately I wasn't able to find a better solution. Let me know if this makes sense!

crates/db_connection_pool/src/dbconnection/odbcconn.rs
mach-kernel: [`CursorImpl::new`](https://docs.rs/odbc-api/7.0.0/odbc_api/struct.CursorImpl.html#method.new) is unsafe. See the other comment for a possible trait tweak that would let us use the safe API.

crates/db_connection_pool/src/dbconnection/odbcconn.rs
mach-kernel: More FFI fun! 😄 

A [connection handle](https://docs.rs/odbc-api/7.0.0/src/odbc_api/handles/connection.rs.html#41) closes over an [environment handle](https://docs.rs/odbc-api/7.0.0/src/odbc_api/handles/environment.rs.html#32) which ends up being a static ref. The ODBC [environment](https://docs.rs/odbc-api/7.0.0/odbc_api/struct.Environment.html) is global and there must be one per process.

Our structs can (and still are) dropped just fine (as they themselves are not static). You can validate this by e.g. doing something like this and triggering a reconnection (fsnotify on saving the dataset yaml seems to do the trick):

```rust
impl<'a> std::ops::Drop for ODBCConnection<'a> {
    fn drop(&mut self) {
        println!("weeeee i'm gettin dropped!");
    }
}
```

mach-kernel:
✨ new! ✨ 

Queried some of our tables that store raw email messages, where one of the columns is the raw binary of the MIME message. Emails can get kind of large. There was a problem with this:

- AFAICT, the Databricks ODBC driver is not reporting accurate col sizes for its result sets (but, this is only an issue with this specific driver)
  - Opened https://github.com/pacman82/arrow-odbc/pull/96 to figure out how to deal with such cases. For now, I settled on choosing `max(driver_metadata_size, user_provided_size)`.
- Added params that pass through to [OdbcReaderBuilder](https://docs.rs/arrow-odbc/latest/arrow_odbc/struct.OdbcReaderBuilder.html) to customize the buffer sizes. 
  - [Saw this PR against odbc-api for an issue with DB2](https://github.com/pacman82/odbc-api/issues/398). So it looks like users will need to be able to toggle quriks from the dataset params.

It looks like this:

```yaml
  from: odbc:default.foo
  name: foo
  mode: read
  time_column: parsed_at
  params:
    odbc_url: ...
    max_binary_size: 16000000
    max_text_size: 16000000
    max_bytes_per_batch: 1000000000
```

crates/db_connection_pool/src/dbconnection/odbcconn.rs
gloomweaver: Thanks for detailed response. Good to know.

crates/db_connection_pool/src/dbconnection/odbcconn.rs
gloomweaver: Thanks for detailed feedback. Makes sense. 
I think we only pass empty vec params to `execute` and `query_arrow` as of now.
Probably can make a change to pass params by value, but needs some more investigating. I will look into it.
Also, would be nice to see comments with justifications of using unsafe in a code.

crates/db_connection_pool/src/odbcpool.rs
```diff
@@ -0,0 +1,88 @@
...
+      // See <https://docs.microsoft.com/en-us/sql/odbc/reference/develop-app/driver-aware-connection-pooling>
+      Environment::set_connection_pooling(AttrConnectionPooling::DriverAware).unwrap();
+      Environment::new().unwrap()
+  };
+}
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Invalid parameter: {parameter_name}"))]
+    InvalidParameterError { parameter_name: String },
```
gloomweaver, line 53:
```suggestion
    InvalidParameterError { parameter_name: String },
    
    #[snafu(display("Missing required parameter: {parameter_name}"))]
    MissingRequiredParameter { parameter_name: String },
```

crates/db_connection_pool/src/odbcpool.rs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+#[async_trait]
+impl<'a> DbConnectionPool<Connection<'a>, ODBCParameter> for ODBCPool
+where
+    'a: 'static,
+{
+    async fn connect(&self) -> Result<Box<ODBCDbConnection<'a>>> {
+        if let Some(params) = self.params.as_ref() {
+            let odbc_url = params.get("odbc_url").expect("Must provide URL");
```
gloomweaver:
```suggestion
            let odbc_url = params
                .get("odbc_url")
                .ok_or(Error::MissingRequiredParameter {
                    parameter_name: "odbc_url".to_string(),
                })?;
```

crates/db_connection_pool/src/dbconnection/odbcconn.rs
```diff
@@ -0,0 +1,271 @@
...
+    let mut builder = OdbcReaderBuilder::new();
+    builder.with_schema(schema.clone());
+
+    let bind_as_usize = |k: &str, f: &mut dyn FnMut(usize) -> ()| {
+        params
+            .as_ref()
+            .and_then(|p| p.get(k).cloned())
+            .and_then(|s| s.parse::<usize>().ok())
+            .into_iter()
+            .for_each(|sz| f(sz));
```
gloomweaver:
```suggestion
            .for_each(f);
```

bin/spiced/Cargo.toml
```diff
@@ -21,7 +21,7 @@ tracing-subscriber.workspace = true
 metrics-exporter-prometheus = "0.13.0"
 
 [features]
-default = ["duckdb", "postgres", "sqlite", "mysql", "flightsql", "databricks", "dremio", "clickhouse"]
+default = ["duckdb", "postgres", "sqlite", "mysql", "flightsql", "databricks", "dremio", "clickhouse", "odbc"]
```
gloomweaver: Do we need to install odbc on our runners to support with as default feature or there is other way?

mach-kernel: This one is a bit tough to get right. I would:

- Make sure the Docker image has `unixodbc` installed, and if possible, any of the useful open source ODBC drivers available on apt/yum/etc
- For a commercial offering, you might have to make an affordance for exposing the ODBC configuration of the runners + ability for users to bring their own ODBC drivers. Popular examples:
  - MS SQL server: https://learn.microsoft.com/en-us/sql/connect/odbc/download-odbc-driver-for-sql-server?view=sql-server-ver16
  - Salesforce: https://www.cdata.com/drivers/salesforce/odbc/
- For a FOSS offering, offer nice docs (I am happy to volunteer to do this, since I need to for our use case) on how to customize the base image to include a user provided ODBC driver + e.g. override the Helm chart image

phillipleblanc (approved):
Thanks @mach-kernel! This is an epic PR.

Issue #1197, opened 2024-04-24
E2E tests: Refactor commonly used steps to actions


gloomweaver:
There are quite a lot repeating steps in CI workflows, need to extract them to separate actions
https://github.com/spiceai/spiceai/pull/1193#discussion_r1577840297

sgrebnov:
Implemented via https://github.com/spiceai/spiceai/pull/1211 and https://github.com/spiceai/spiceai/pull/1205

PR #1205, opened 2024-04-26
E2E: reusable Spice installation

sgrebnov:
Part of https://github.com/spiceai/spiceai/issues/1197
Add re-usable script to install Spice from build artifacts:
 - `spice version` is now part of Spice installation as final verification step
 - Example test run: https://github.com/spiceai/spiceai/actions/runs/8841248299
image


ewgenius (dismissed):

.github/workflows/e2e_test_ci.yml
```diff
@@ -826,15 +793,7 @@ jobs:
...
-          chmod +x ./build/spice
-          chmod +x ./build/spiced
-          mkdir -p "$HOME/.spice/bin"
-          mv ./build/spice "$HOME/.spice/bin"
-          mv ./build/spiced "$HOME/.spice/bin"
-          echo "$HOME/.spice/bin" >> $GITHUB_PATH
-
-      - name: Check spice version
-        run: spice version
+          export SPICE_BUILD_ARTIFACTS_DIR=$PWD/build && ./test/scripts/install_spice.sh
```
digadeesh: Couldn't we just make this an action like this?
https://github.com/spiceai/spiceai/blob/trunk/.github/actions/setup-rust/action.yml

sgrebnov: @digadeesh  -  yeah, @ewgenius  seems to also like Actions more than scripts where I prefer scripts as they are more reusable. As we have more people thinking Actions is a better choice I'll re-work this later today

sgrebnov: Updated

ewgenius (approved):

Issue #1181, opened 2024-04-23
Unable to select timestamp field from Sqlite Data Accelerator


sgrebnov:
**Describe the bug**

There is the following error when query Timestamp field from Sqlite Data Accelerator

>Error receiving data: Decode error: Error decoding ipc RecordBatch: Invalid argument error: Need at least 80 bytes in buffers[0] in array of type Timestamp(Microsecond, None), but got 44

**To Reproduce**

1. Create the following pod
```yaml
version: v1beta1
kind: Spicepod
name: retention-quickstart
datasets:
- from: s3://spiceai-demo-datasets/taxi_trips/2024/
  name: taxi_trips
  description: taxi trips in s3
  acceleration:
    enabled: true
    refresh_interval: 30s
    refresh_mode: full
    engine: sqlite
```
3. Run Spice REPL using `spice sql`
4. Perform query `select tpep_pickup_datetime from taxi_trips limit 10;`
7. See error
>Error receiving data: Decode error: Error decoding ipc RecordBatch: Invalid argument error: Need at least 80 bytes in buffers[0] in array of type Timestamp(Microsecond, None), but got 44

**Additional context**
Can query other fields
```bash
sql> select total_amount from taxi_trips limit 10
+--------------+
| total_amount |
+--------------+
| 27.5         |
| 18.37        |
| 35.28        |
| 15.96        |
| 26.88        |
| 90.94        |
| 51.3         |
| 62.89        |
| 39.85        |
| 12.9         |
+--------------+
```

taxi_trips format
```bash
sql> describe taxi_trips
+-----------------------+------------------------------+-------------+
| column_name           | data_type                    | is_nullable |
+-----------------------+------------------------------+-------------+
| VendorID              | Int32                        | YES         |
| tpep_pickup_datetime  | Timestamp(Microsecond, None) | YES         |
| tpep_dropoff_datetime | Timestamp(Microsecond, None) | YES         |
| passenger_count       | Int64                        | YES         |
| trip_distance         | Float64                      | YES         |
| RatecodeID            | Int64                        | YES         |
| store_and_fwd_flag    | LargeUtf8                    | YES         |
| PULocationID          | Int32                        | YES         |
| DOLocationID          | Int32                        | YES         |
| payment_type          | Int64                        | YES         |
| fare_amount           | Float64                      | YES         |
| extra                 | Float64                      | YES         |
| mta_tax               | Float64                      | YES         |
| tip_amount            | Float64                      | YES         |
[...]

PR #1212, opened 2024-04-27
Add schema types check for query result

sgrebnov:
Part of https://github.com/spiceai/spiceai/issues/1181

Every Data Accelerator is responsible for correct data type mapping. If an accelerator failed to correctly handle some type and returned different/invalid type it is hard to detect and can lead to records decoding issues (for example https://github.com/spiceai/spiceai/issues/1181).

PR adds schema check before returning results to ensure expected schema (data source schema) and records schema returned by query engine match.

Before
image

After
image

Temporary disabled sqlite acceleration from E2E tests as it returns Int64 when schema is Int32 ([test run](https://github.com/spiceai/spiceai/actions/runs/8858326856/job/24326863304#step:13:119))


crates/runtime/src/flight.rs
```diff
@@ -229,6 +233,38 @@ impl Service {
     }
 }
 
+fn verify_schema(
```
phillipleblanc, line 236: Schema already implements [`Eq`](https://docs.rs/arrow/latest/arrow/datatypes/struct.Schema.html#impl-Eq-for-Schema) - so you can just compare two schemas with `==`

i.e. 
```rust
let schema1 = arrow::datatypes::Schema::new(vec![
    arrow::datatypes::Field::new("a", arrow::datatypes::DataType::Int64, false),
    arrow::datatypes::Field::new("b", arrow::datatypes::DataType::Utf8, false),
]);

let schema2 = arrow::datatypes::Schema::new(vec![
    arrow::datatypes::Field::new("a", arrow::datatypes::DataType::Int64, false),
    arrow::datatypes::Field::new("b", arrow::datatypes::DataType::Utf8, false),
]);

assert_eq!(schema1, schema2); // True!
if schema1 == schema2 {
  println!("Also true!");
}
```

crates/runtime/src/flight.rs
```diff
@@ -282,6 +318,19 @@ pub enum Error {
 
     #[snafu(display("Unable to start Flight server: {source}"))]
     UnableToStartFlightServer { source: tonic::transport::Error },
+
+    #[snafu(display("Expected and actual number of fields in the query result don't match: expected {expected}, received {actual}"))]
+    SchemaMismatchNumFields { expected: usize, actual: usize },
+
+    #[snafu(display("Query returned an unexpected data type for column {name}: expected {expected}, received {actual}. Is the column data type supported by the used data accelerator (https://docs.spiceai.org/reference/datatypes)?"))]
```
phillipleblanc:
```suggestion
    #[snafu(display("Query returned an unexpected data type for column {name}: expected {expected}, received {actual}. Is the column data type supported by the data accelerator (https://docs.spiceai.org/reference/datatypes)?"))]
```

crates/runtime/src/flight.rs
phillipleblanc: If we need a more lenient schema comparison, then we can use one of the two methods in https://docs.rs/datafusion/latest/datafusion/common/trait.SchemaExt.html#required-methods
`equivalent_names_and_types` or `logically_equivalent_names_and_types`

crates/runtime/src/flight.rs
sgrebnov: Thank you!

crates/runtime/src/flight.rs
sgrebnov: @phillipleblanc  - thank you. The idea was to provide a very granular error - what specific field does not match instead of generic schema mismatch error. **Please confirm you still prefer to use `equivalent_names_and_types`**.

Thanks for pointing to built-in functions. Looking to them I've also found [check_arrow_schema_type_compatible](https://github.com/apache/datafusion/blob/main/datafusion/common/src/dfschema.rs#L589) but it reports OK/compatible so we can't use it here (. 
```
/// Check to see if fields in 2 Arrow schemas are compatible
    pub fn check_arrow_schema_type_compatible(
        &self,
        arrow_schema: &Schema,
    ) -> Result<()> {
        let self_arrow_schema: Schema = self.into();
        self_arrow_schema
            .fields()
            .iter()
            .zip(arrow_schema.fields().iter())
            .try_for_each(|(l_field, r_field)| {
                if !can_cast_types(r_field.data_type(), l_field.data_type()) {
                    _plan_err!("Column {} (type: {}) is not compatible with column {} (type: {})",
                                r_field.name(),
                                r_field.data_type(),
                                l_field.name(),
                                l_field.data_type())
                } else {
                    Ok(())
                }
            })
    }
```

image

.github/actions/verify-dataset/action.yml
```diff
@@ -39,11 +39,18 @@ runs:
     - name: Verify SQL REPL result count for table ${{ inputs.name }}
       shell: bash
       run: |
-        sql_output=$(echo "select * from ${{ inputs.name }} limit 10;" | spice sql)
-        echo "$sql_output"
+        query="select * from ${{ inputs.name }} limit 10;"
+        echo "Executing query: $query"
+        sql_output=$(spice sql <<EOF
+        $query
+        .error
```
sgrebnov, line 46: This includes internal error (.error) to sql_output. Works even if there is no error - there will be 'No error to display' in this case.

crates/runtime/src/flight.rs
phillipleblanc: Ok that makes sense

phillipleblanc (approved):

PR #1215, opened 2024-04-28
Optimize build time in dev mode

gloomweaver:
Replace `--release` flag with `--profile dev` for faster build speeds with `make install-dev`
Getting two times faster builds when `make install` on my machine
With selecting only features you're currently working on you can additionally get up to two times faster builds when dev with all features


phillipleblanc (approved):
Thanks!

PR #1217, opened 2024-04-29
Support Date32 & Date64 in arrow_sql_gen

Jeadie:
# Changes
 - Support Date32 & Date64 in `arrow_sql_gen` crate.
 - This allows support for these types to postgres, mysql, sqlite etc (assuming the underlying data backend can support it).


crates/arrow_sql_gen/src/statement.rs
```diff
@@ -167,6 +167,30 @@ impl InsertBuilder {
...
                     }
+                    DataType::Date32 => {
+                        let array = column.as_any().downcast_ref::<array::Date32Array>();
+                        if let Some(valid_array) = array {
+                            row_values.push(
+                                match OffsetDateTime::from_unix_timestamp(valid_array.value(row) as i64 * 86_400) {
+                                    Ok(offset_time) => offset_time.date().into(),
+                                    Err(_) => return,
+                                }
+
```
phillipleblanc:
```suggestion
```

crates/arrow_sql_gen/src/statement.rs
```diff
@@ -167,6 +167,30 @@ impl InsertBuilder {
...
+                    }
+                    DataType::Date64 => {
+                        let array = column.as_any().downcast_ref::<array::Date64Array>();
+                        if let Some(valid_array) = array {
+                            row_values.push(
+                                match OffsetDateTime::from_unix_timestamp(valid_array.value(row) as i64 * 86_400) {
+                                    Ok(offset_time) => offset_time.date().into(),
+                                    Err(_) => return,
+                                }
+
```
phillipleblanc:
```suggestion
```

crates/arrow_sql_gen/src/statement.rs
```diff
@@ -167,6 +167,30 @@ impl InsertBuilder {
                             );
                         }
                     }
+                    DataType::Date32 => {
+                        let array = column.as_any().downcast_ref::<array::Date32Array>();
+                        if let Some(valid_array) = array {
+                            row_values.push(
+                                match OffsetDateTime::from_unix_timestamp(valid_array.value(row) as i64 * 86_400) {
+                                    Ok(offset_time) => offset_time.date().into(),
+                                    Err(_) => return,
```
phillipleblanc: It is weird that we are just ignoring this error - I know this was done elsewhere, but I think we should properly return an error and bubble it up to the runtime to be reported. Otherwise we'll silently ignore these and it could lead to subtle hard to debug data quality bugs.

Fine to create an issue and tackle it in a separate PR though.

crates/arrow_sql_gen/src/statement.rs
```diff
@@ -308,33 +317,35 @@ impl InsertBuilder {
                             }
                         }
                     }
-                    _ => unimplemented!(
-                        "Data type mapping not implemented for {}",
-                        column.data_type()
-                    ),
+                    _ => return Result::Err(Error::FailedToCreateInsertStatement { source: format!("Data type mapping not implemented for {}", column.data_type()).into() })
```
phillipleblanc: This should be a separate Snafu error type specific to Unimplemented

crates/arrow_sql_gen/src/statement.rs
```diff
@@ -308,33 +317,35 @@ impl InsertBuilder {
...
+            
+            }
         }
+        Ok(())
     }
 
     #[must_use]
     pub fn build_postgres(self) -> String {
-        self.build(PostgresQueryBuilder)
+        self.build(PostgresQueryBuilder).unwrap_or_default()
```
phillipleblanc: This should also bubble up the error (here and for `build_sqlite` and `build_mysql`)

crates/arrow_sql_gen/src/statement.rs
```diff
@@ -348,9 +359,12 @@ impl InsertBuilder {
             .to_owned();
 
         for record_batch in &self.record_batches {
-            self.construct_insert_stmt(&mut insert_stmt, record_batch);
```
phillipleblanc, line 329: Does the `?` syntax not work here?

crates/arrow_sql_gen/src/statement.rs
Jeadie: I've added a Result to `.build()`, and log an error for now.

crates/arrow_sql_gen/src/statement.rs
phillipleblanc: Fine to create an issue to track and leave this as is

crates/arrow_sql_gen/src/statement.rs
Jeadie: Issue: https://github.com/spiceai/spiceai/issues/1221

crates/arrow_sql_gen/src/statement.rs
Jeadie: It does

crates/arrow_sql_gen/src/statement.rs
Jeadie: Done

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #1214, opened 2024-04-28
Improve clarity and consistency of output messages

lukekim:

PR #1218, opened 2024-04-29
Add E2E tests for acceleration engines types support

sgrebnov:
Motivation/Why?
 - We have a record of issues related to acceleration backends when specific data type can't be saved locally (runtime panic/error) or query failure.
 - This PR attempts to test all possible types support by acceleration backends by creating local DuckDB instance as a data source containing different types (DuckDB was selected as the one easy to setup locally; if we like this approach we can switch to remote s3/parquet file), enable supported accelerations and compare query results (query source DuckDB) vs accelerated backend (DuckDB, Arrow, sqlite, postgreSQL(not currently covered))

Implements Github workflow to test that acceleration backend correctly support common types.
1. Creates test duckdb database and populates it with main Arrow types (isolated table per type as some types cause Spice runtime to crash so all types are isolated in own table)
2. For each type creates spicepod and runs query twice
  - without acceleration
  - with local acceleration
3. Compares results and creates summary table

Example test run (scroll page down to find summary): https://github.com/spiceai/spiceai/actions/runs/8872467039

Sample report for boolean
image
image
image


.github/actions/run-query/action.yml
```diff
@@ -0,0 +1,69 @@
...
+
+        echo "$sql_output"
+
+        if [[ $sql_output != *"No error to display"* ]]; then
+          echo "Query failed to execute."
+        fi
+
+        echo "query-output=$(echo "$sql_output" | base64 -w 0)" >> "$GITHUB_OUTPUT"
+
+        extracted_result=$(echo "$sql_output" | sed -n '/show tables;/,/Query took/p' | sed '1d;$d')
```
phillipleblanc: This will need to take into account https://github.com/spiceai/spiceai/pull/1214

.github/workflows/types_check.yml
```diff
@@ -0,0 +1,500 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.x
+
+      - name: Generate test data
+        run: |
+          pip install pyarrow duckdb
+          python test/arrow-types/create_arrow_types_test_data.py
+          ls -l test/arrow-types/
```
phillipleblanc, line 128: This only should happen for the `duckdb` engine right?

.github/workflows/types_check.yml
```diff
@@ -0,0 +1,500 @@
...
+      - name: Test boolean
+        id: test-boolean
+        continue-on-error: true
+        uses: ./.github/actions/verify-accelerated-query
+        with:
+          dataset-source: |
+            - from: duckdb:arrow_types.t_boolean
+              name: t_boolean
+              params:
+                  open: arrow_types.db
```
phillipleblanc, line 150: It looks like this only works for the `duckdb` engine? I don't quite see the point in the matrix then

.github/actions/verify-accelerated-query/action.yml
```diff
@@ -0,0 +1,77 @@
...
+        spice init app_${{ inputs.table-name }}
+        # copy local databases (if any) to create spicepod folder
+        cp *.db app_${{ inputs.table-name }}
+
+    - name: Create non-accelerated dataset
+      shell: bash
+      working-directory: app_${{ inputs.table-name }}
+      run: |
+        echo "Creating non-accelerated dataset"
+        echo "datasets:" >> spicepod.yaml
```
phillipleblanc, line 37: This doesn't need to change for this PR - but I'm not a huge fan of constructing spicepods manually via string manipulation. It is brittle - if a breaking change is made to the spicepod definition, its hard to find all of these places that would need to be updated. Ideally we could use the CLI or some other tool that manipulates spicepods using our existing spicepod parser.

Let's file an issue to track.

.github/workflows/types_check.yml
```diff
@@ -0,0 +1,500 @@
+name: Test types support
+
+on:
+  workflow_dispatch:
+
+concurrency:
+  # Allow only one workflow per any non-trunk branch.
+  group: ${{ github.workflow }}-${{ github.ref_name }}-${{ github.ref_name == 'trunk' && github.sha || 'any-sha' }}
+  cancel-in-progress: true
```
phillipleblanc: Given this is a workflow_dispatch only action, do we need this concurrency check?

.github/workflows/types_check.yml
```diff
@@ -0,0 +1,500 @@
...
+    runs-on: ${{ matrix.target.runner }}
+    needs: setup-matrix
+    env:
+      GOVER: 1.22.0
+      GOOS: ${{ matrix.target.target_os }}
+      GOARCH: ${{ matrix.target.target_arch_go }}
+
+    strategy:
+      matrix:
+        target: ${{ fromJson(needs.setup-matrix.outputs.matrix) }}
```
phillipleblanc: Why not just inline the matrix here instead of the extra step to create a json and then parse it?

.github/workflows/types_check.yml
```diff
@@ -0,0 +1,500 @@
...
+        uses: actions/upload-artifact@v4
+        if: matrix.target.target_os != 'windows'
+        with:
+          name: build_${{ matrix.target.target_os }}_${{ matrix.target.target_arch }}
+          path: |
+            spice
+            spiced
+
+  test_acceleration:
+    name: 'Arrow types using ${{ matrix.acceleration.engine }} on ${{ matrix.target.name }}'
```
phillipleblanc, line 96: This test can only be executed in GitHub Actions - which I think is fine in general, and matches what we do for the E2E tests. But it would be nice if we could also somehow run a version of it locally.

.github/actions/run-query/action.yml
```diff
@@ -0,0 +1,69 @@
+name: 'Execute Spice REPl query'
```
lukekim: REPL is REPL

.github/workflows/types_check.yml
ewgenius: Matrix can't be dynamic, it has to be calculated before scheduling dependent workflow.
That makes sense if we want this workflow behave same as E2E tests: use  single platform for quick PR runs, and all platforms for manual and on push to trunk. @sgrebnov is that correct?

.github/workflows/types_check.yml
sgrebnov: This local duckdb database is used as a data source for all acceleration engines so we can verify non-accelerated query vs accelerated. Added more details on the idea/approach to the ticket. Sorry that approach was not clarified.

.github/actions/verify-accelerated-query/action.yml
sgrebnov: Yeah, created the following issue: https://github.com/spiceai/spiceai/issues/1226

.github/workflows/types_check.yml
sgrebnov: Forgot to remove after workflow testing, to be updated. Thank you!

.github/actions/run-query/action.yml
sgrebnov: Updated, thank you!

.github/actions/run-query/action.yml
sgrebnov: Updated, thank you!

.github/workflows/types_check.yml
sgrebnov: The same duckdb local database is used as a data source (different tables with different data types) for all acceleration engines so we can verify non-accelerated query vs accelerated. Added more details on the idea/approach to the ticket. Sorry that approach was not clarified.

.github/workflows/types_check.yml
sgrebnov: [Test data generation script ](https://github.com/spiceai/spiceai/blob/sgrebnov/acceleration-types-test/test/arrow-types/create_arrow_types_test_data.py#L68)  contains logic (currently commented) to generated a ready to use spicepod. It can be used for local testing (not automated, but experimenting/testing with specific type). I can enable this and add readme.md with build instructions (few commands) - pls let me know if you think we should do this

sgrebnov: I don't think we need to run those tests on all platforms (single Linux is fine IMO). I'll review and update - thank you @phillipleblanc and @ewgenius. Current matrix logic is copy-paste from other places, need to review.

.github/workflows/types_check.yml
sgrebnov: @ewgenius - this is correct, thanks, I used the same approach as we do for other workflows where we have dynamic behavior (quick PR run vs full tests). For this particular case I don't think we will want to run such tests as part of PR so we actually don't need matrix to be dynamic so I've simplified the code. Thanks. /cc @phillipleblanc  - updated.

https://github.com/spiceai/spiceai/pull/1218/commits/4a7a51e1a69fe28f80f58d0d26cb8b6b4c49f411

test/arrow-types/create_arrow_types_test_data.py
```diff
@@ -0,0 +1,91 @@
...
+if os.path.exists('arrow_types.db'):
+    os.remove('arrow_types.db')
+
+conn = duckdb.connect('arrow_types.db')
+for field in fields:
+    table_name = field.name
+    conn.execute(f"CREATE TABLE arrow_types.{table_name} AS SELECT {field.name} FROM read_parquet('arrow_types.parquet');")
+    print(conn.execute(f"DESCRIBE arrow_types.{table_name};").fetchall())
+
+# Generate the spicepod.yml
```
digadeesh: Do we need this?

test/arrow-types/create_arrow_types_test_data.py
```diff
@@ -0,0 +1,91 @@
+import os
```
digadeesh, line 15: Can you add the apache license comment to this file?

sgrebnov: Updated, thank you!

test/arrow-types/create_arrow_types_test_data.py
sgrebnov: I would keep this so you can create local test setup if needed. This generates ready to use spicepod definition for all types so you can run the same tests locally (via manual SQL REPL query) . Should I remove this or add some additional comment here regarding the purpose and how this could be used?

sgrebnov: Another option is to add extra parameter that will control whether spicepod should be generated or not - for example TYPES_TEST_DATA_GENERATE_SPICEPOD

digadeesh: why not make this a utility that you can run separately then?

sgrebnov: Updated to always generate `spicepod.yml`  that could be used for local testing. This does not affect GH Action execution.

digadeesh (approved):

PR #1244, opened 2024-04-30
Update postgres parameter check

### LINKED ISSUE #1269: Test connection to the postgres using direct `tokio_postgres::connect` (if getting error details from bb8 is too complex)

Sevenannn:
* Update the error handling for pg_port & pg_host check, previous handling method doesn't give any error messages.
* Use tokio-postgres to test connection using user provided parameters before connecting using bb8 pool. This will immediately propagate any error relevant to user parameter misconfiguration, and therefore avoid the waiting on bb8 connection result and the vague bb8 error message.

### DIFF
--- a/crates/db_connection_pool/Cargo.toml
+++ b/crates/db_connection_pool/Cargo.toml
@@ -51,6 +51,7 @@ postgres = [
     "dep:tokio",
     "arrow_sql_gen/postgres",
     "spicepod/postgres",
+    "dep:tokio-postgres",
 ]
 sqlite = ["dep:rusqlite", "dep:tokio-rusqlite", "arrow_sql_gen/sqlite", "spicepod/sqlite"]
 mysql = ["dep:mysql_async", "arrow_sql_gen/mysql", "spicepod/mysql"]
--- a/crates/db_connection_pool/Cargo.toml
+++ b/crates/db_connection_pool/Cargo.toml
@@ -34,6 +34,7 @@ odbc-api = { workspace = true, optional = true }
 arrow-odbc = { workspace = true, optional = true }
 lazy_static = "1.4.0"
 clickhouse-rs = { workspace = true, optional = true }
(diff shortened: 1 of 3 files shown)


crates/db_connection_pool/src/postgrespool.rs
```diff
@@ -110,8 +117,9 @@ impl PostgresConnectionPool {
                 }
                 if let Some(pg_sslmode) = params.get("pg_sslmode") {
                     match pg_sslmode.to_lowercase().as_str() {
-                        "disable" | "require" | "prefer" | "verify-ca" | "verify-full" => {
```
ewgenius, line 113: `verify-ca` and `verify-full` not supported by tokio_postgres, but we simulate it by controlling hostname and cert validation, check this lines
https://github.com/spiceai/spiceai/blob/1bdb613457787c938388948b10ac443df89f5526/crates/db_connection_pool/src/postgrespool.rs#L203-L206

crates/db_connection_pool/Cargo.toml
```diff
@@ -34,9 +34,9 @@ odbc-api = { workspace = true, optional = true }
 arrow-odbc = { workspace = true, optional = true }
 lazy_static = "1.4.0"
 clickhouse-rs = { workspace = true, optional = true }
+tokio-postgres = "0.7.10"
```
ewgenius: We alreaady have a dependency for it in workspace

```suggestion
tokio-postgres = { workspace = true }
```

phillipleblanc: This should also be optional and enabled by the `postgres` feature

ewgenius:
closes #1269

crates/db_connection_pool/Cargo.toml
Sevenannn: updated to optional tokio postures dependency

digadeesh (approved):

PR #1248, opened 2024-04-30
Top level err handler

Sevenannn:
A minimal version of error traversal and retrieval of the last error in error source chain. However we still need a way to distinguish Spice defined error versus external crate errors.

The current solution I’ve tried is bring Spice defined errors into scope and try to downcast error source to Spice errors, but there are 2 problems
* Bringing Spice defined errors into runtime scope will cause cyclical dependency, for example, adding db_connection_pool into Cargo.toml in runtime
* We will have to bring in all Spice defined errors into runtime and downcast for every source in the error chain, which is extremely inefficient.
We probably need a better way to infer whether the error is Spice error or not.


Sevenannn:
The changes is now available in [Define and pass through custom error messages #1267](https://github.com/spiceai/spiceai/pull/1267)

PR #1261, opened 2024-05-01
Record batch casting to fix SQLite data type issues

### LINKED ISSUE #1028: SQLite Data backend issue with Timestamp(microseconds) (needed in `taxi_trips` dataset).
**Describe the bug**
When loading the example taxi_trips data form S3 using the SQLite accelerator, the spice runtime fails and crashes.

**To Reproduce**
1. Edit spicepod.yaml with this configuration:

```yaml
version: v1beta1
kind: Spicepod
name: my_test
datasets:
  name: taxi_trips
  description: taxi trips in s3
  acceleration:
    enabled: true
    engine: sqlite
```
2. Open the sql repl
3. query the taxi_trips table
4. Observe the type conversion error in the spice runtime output:
```bash
thread '<unnamed>' panicked at crates/arrow_sql_gen/src/statement.rs:331:14:
not implemented: Data type mapping not implemented for Date32
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-04-03T05:21:34.784751Z ERROR runtime::datafusion: Error adding data: Failed to update sqlite table: ConnectionClosed
```

**Expected behavior**
A successful query on the taxi_trips data

### LINKED ISSUE #1181: Unable to select timestamp field from Sqlite Data Accelerator

### LINKED ISSUE #1236: Enhancement: Convert accelerated schema to source schema
## Goal-State/What/Result

A mechanism can cast record from accelerator schema into the schema consumed by user.

## Why/Purpose

Databases(accelerator) have different types to Arrow data types; in order to serve records via flights, the record comes from accelerator needs to have the same arrow schema as the one registered in accelerated table provider. Or it will break on any mismatch which is very common now.

## By When

**Done-Done:** May 1st, 2024 AEST

## Done-Done

- [x] The Algorithm
- [x] PM/Design Review
- [x] ~~UX Review~~
- [x] ~~Threat Model / Security Review~~
- [x] Tests
- [x] ~~Documentation~~
- [x] ~~Samples/Quickstarts/Tutorials~~
- [x] Release Notes

## How/MAP

- A record batch utility to help cast the data from 1 schema to another schema with best effort.

- A datafusion idiomatic way to embed above function in the execution flow.

y-f-u:
fixes #1181 
fixes #1028 

fixes #1236

This introduces a RecordBatch helper that `try_cast_to` given schema for an existing record batch.

Also introduce a execution plan that casts the records in more engine primitive type schema to the schema defined in accelerated table.

Note, this is not trying to exhaust all possible casts, all casts need to be implemented on demand.

### DIFF
--- a/crates/arrow_tools/src/record_batch.rs
+++ b/crates/arrow_tools/src/record_batch.rs
@@ -14,8 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use std::{num::TryFromIntError, sync::Arc};
-
 use arrow::{
     array::{
         new_null_array, Array, ArrayRef, Int32Array, Int64Array, LargeStringArray, RecordBatch,
@@ -24,6 +22,7 @@ use arrow::{
     datatypes::{DataType, Field, SchemaRef, TimeUnit},
 };
 use snafu::prelude::*;
+use std::{num::TryFromIntError, sync::Arc};
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
@@ -46,11 +45,11 @@ pub enum Error {
     ConvertInt { source: TryFromIntError },
 }
 
-/// Convert a given record batch into a new record batch with the given schema.
+/// Cast a given record batch into a new record batch with the given schema.
 ///
 /// # Errors
 ///
-/// This function will return an error if the record batch cannot be converted.
+/// This function will return an error if the record batch cannot be cast.
 #[allow(clippy::needless_pass_by_value)]
 pub fn cast_to(record_batch: RecordBatch, schema: SchemaRef) -> Result<RecordBatch> {
     let existing_schema = record_batch.schema();
@@ -74,7 +73,7 @@ pub fn cast_to(record_batch: RecordBatch, schema: SchemaRef) -> Result<RecordBat
             if field.contains(existing_field) {
                 array = Arc::clone(column);
             } else {
-                array = convert_column(Arc::clone(column), existing_field, field)?;
+                array = cast_array(Arc::clone(column), existing_field, field)?;
             }
         }
 
@@ -85,47 +84,52 @@ pub fn cast_to(record_batch: RecordBatch, schema: SchemaRef) -> Result<RecordBat
 }
 
 #[allow(clippy::needless_pass_by_value)]
-fn convert_column(
-    column: Arc<dyn Array>,
+fn cast_array(
+    array: Arc<dyn Array>,
     existing_field: &Field,
     field: &Field,
 ) -> Result<Arc<dyn Array>> {
     let result: Arc<dyn Array> = match (existing_field.data_type(), field.data_type()) {
         (DataType::Int32, DataType::Int64) => {
-            if let Some(array) = column.as_any().downcast_ref::<Int32Array>() {
-                let converted: Vec<Option<i64>> =
-                    array.into_iter().map(|f| f.map(i64::from)).collect();
-                Arc::new(Int64Array::from(converted))
+            if let Some(array) = array.as_any().downcast_ref::<Int32Array>() {
+                Arc::new(Int64Array::from(
+                    array
+                        .into_iter()
+                        .map(|f| f.map(i64::from))
+                        .collect::<Vec<Option<i64>>>(),
+                ))
             } else {
-                Arc::new(new_null_array(field.data_type(), column.len()))
+                Arc::new(new_null_array(field.data_type(), array.len()))
             }
         }
         (DataType::Int64, DataType::Int32) => {
-            if let Some(array) = column.as_any().downcast_ref::<Int64Array>() {
-                let converted: Vec<Option<i32>> = array
-                    .into_iter()
-                    .map(|f| -> Result<Option<i32>> {
-                        match f {
-                            Some(f) => Ok(Some(i32::try_from(f).context(ConvertIntSnafu)?)),
-                            None => Ok(None),
-                        }
-                    })
-                    .collect::<Result<_, _>>()?;
-                Arc::new(Int32Array::from(converted))
+            if let Some(array) = array.as_any().downcast_ref::<Int64Array>() {
+                Arc::new(Int32Array::from(
+                    array
+                        .into_iter()
+                        .map(|f| -> Result<Option<i32>> {
+                            match f {
+                                Some(f) => Ok(Some(i32::try_from(f).context(ConvertIntSnafu)?)),
+                                None => Ok(None),
+                            }
+                        })
+                        .collect::<Result<Vec<Option<i32>>, _>>()?,
+                ))
             } else {
-                Arc::new(new_null_array(field.data_type(), column.len()))
+                Arc::new(new_null_array(field.data_type(), array.len()))
             }
         }
         (DataType::Utf8, DataType::LargeUtf8) => {
-            if let Some(array) = column.as_any().downcast_ref::<StringArray>() {
-                let converted: Vec<Option<&str>> = array.into_iter().collect();
-                Arc::new(LargeStringArray::from(converted))
+            if let Some(array) = array.as_any().downcast_ref::<StringArray>() {
+                Arc::new(LargeStringArray::from(
+                    array.into_iter().collect::<Vec<Option<&str>>>(),
+                ))
             } else {
-                Arc::new(new_null_array(field.data_type(), column.len()))
+                Arc::new(new_null_array(field.data_type(), array.len()))
             }
         }
         (DataType::Utf8, DataType::Timestamp(unit, _)) => {
-            convert_timestamp_column(column, unit, field)?
+            cast_array_from_string_to_timestamp(array, unit, field)?
         }
         (_, _) => ConvertArrowArraySnafu {
             from: existing_field.data_type().clone(),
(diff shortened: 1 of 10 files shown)


crates/arrow_tools/src/record_batch.rs
```diff
@@ -0,0 +1,267 @@
...
+    datatypes::{DataType, Field, SchemaRef, TimeUnit},
+};
+use snafu::prelude::*;
+use std::{num::TryFromIntError, sync::Arc};
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Error converting record batch: {}", source))]
```
phillipleblanc:
```suggestion
    #[snafu(display("Error converting record batch: {source}"))]
```

crates/arrow_tools/src/record_batch.rs
```diff
@@ -0,0 +1,267 @@
...
+    ConvertRecordBatch { source: arrow::error::ArrowError },
+
+    #[snafu(display("Unable to convert array: {}", source))]
```
phillipleblanc:
```suggestion
    #[snafu(display("Unable to convert array: {source}"))]
```

crates/arrow_tools/src/record_batch.rs
```diff
@@ -0,0 +1,267 @@
...
+    ParseTimestampFromString { source: chrono::ParseError },
+
+    #[snafu(display("Unable to convert {} array from {:?} to {:?}", name, from, to))]
```
phillipleblanc:
```suggestion
    #[snafu(display("Unable to convert {name} array from {from:?} to {to:?}"))]
```

crates/arrow_tools/src/record_batch.rs
```diff
@@ -0,0 +1,267 @@
...
+    ConvertArrowArray {
+        from: DataType,
+        to: DataType,
+        name: String,
+    },
+
+    #[snafu(display("Unable to try from int: {}", source))]
```
phillipleblanc:
```suggestion
    #[snafu(display("Unable to try from int: {source}"))]
```

crates/arrow_tools/src/record_batch.rs
```diff
@@ -0,0 +1,267 @@
...
+    ConvertInt { source: TryFromIntError },
+
+    #[snafu(display("Unable to use null for field {field}"))]
+    UnableToUseNullForField { field: String },
```
phillipleblanc:
```suggestion
    #[snafu(display("Field is not nullable: {field}"))]
    FieldNotNullable { field: String },
```

crates/arrow_tools/src/record_batch.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    #[snafu(display("Unable to downcast array {field} to {downcast_to}"))]
+    UnableToDowncast { field: String, downcast_to: String },
+}
+
+/// Cast a given record batch into a new record batch with the given schema.
+///
+/// # Errors
+///
+/// This function will return an error if the record batch cannot be casted.
```
phillipleblanc:
```suggestion
/// This function will return an error if the record batch cannot be cast.
```

crates/arrow_tools/src/record_batch.rs
```diff
@@ -0,0 +1,267 @@
...
+#[allow(clippy::needless_pass_by_value)]
+pub fn try_cast_to(record_batch: RecordBatch, schema: SchemaRef) -> Result<RecordBatch> {
+    let existing_schema = record_batch.schema();
+
+    if schema.contains(&existing_schema) {
```
phillipleblanc, line 47: What are the semantics for `.contains()` and why we're returning a new record batch with the passed in schema instead of just returning it directly? It would be good to add this as a comment given its not obvious.

crates/arrow_tools/src/record_batch.rs
```diff
@@ -0,0 +1,267 @@
...
+            if let (Ok(existing_field), Some(column)) = (
+                record_batch.schema().field_with_name(field.name()),
+                record_batch.column_by_name(field.name()),
+            ) {
+                if field.contains(existing_field) {
+                    Ok(Arc::clone(column))
+                } else {
+                    try_cast_array(Arc::clone(column), existing_field, field)
+                }
+            } else if field.is_nullable() {
```
phillipleblanc, line 69: What are the scenarios in which we want to do this vs just erroring? This basically means the requested schema has a column that doesn't exist in the source schema - which I would expect to be an error case.

crates/arrow_tools/src/record_batch.rs
```diff
@@ -0,0 +1,267 @@
...
+                        .map(|f| f.map(i64::from))
+                        .collect::<Vec<Option<i64>>>(),
+                ))
+            } else {
+                UnableToDowncastSnafu {
+                    field: field.name(),
+                    downcast_to: "Int32Array".to_string(),
+                }
+                .fail()?
+            }
```
phillipleblanc:
```suggestion
            let array = array.as_any().downcast_ref::<Int32Array>().context(
                  UnableToDowncastSnafu {
                      field: field.name(),
                      downcast_to: "Int32Array".to_string(),
                  },
              )?;
              Arc::new(Int64Array::from(
                  array
                      .into_iter()
                      .map(|f| f.map(i64::from))
                      .collect::<Vec<Option<i64>>>(),
              ))
```

I find this a bit nicer.

phillipleblanc (commented):
I think we may be able to use https://github.com/apache/arrow-rs/tree/master/arrow-cast for this instead of writing our own.

crates/arrow_tools/src/record_batch.rs
y-f-u: Technically you can cast a record batch from less columns to more columns if the  new fields are nullable. Which can support schema expansion.

crates/arrow_tools/src/record_batch.rs
```diff
@@ -0,0 +1,267 @@
...
+};
+use snafu::prelude::*;
+use std::{num::TryFromIntError, sync::Arc};
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Error converting record batch: {}", source))]
+    ConvertRecordBatch { source: arrow::error::ArrowError },
```
ewgenius: Nitpick: I think error names should represent the failure
```suggestion
    UnableToConvertRecordBatch { source: arrow::error::ArrowError },
```

crates/arrow_tools/src/record_batch.rs
y-f-u: Is about each field is a superset of the original one. Mostly it’s about nullable, when casting from a non-nullable to a nullable one, the change on schema needs to be considered.

y-f-u: Will add a comment in.

crates/arrow_tools/src/record_batch.rs
y-f-u: good to know.

crates/arrow_tools/src/record_batch.rs
y-f-u: Yeah, rust-analyzer suggested a common prefix is not preferred but I guess I can override the linter rule in this case.

crates/runtime/src/execution_plan/schema_cast.rs
```diff
@@ -0,0 +1,138 @@
...
+    properties: PlanProperties,
+}
+
+impl SchemaCastScanExec {
+    pub fn new(input: Arc<dyn ExecutionPlan>, schema: SchemaRef) -> Self {
+        let eq_properties = input.equivalence_properties().clone();
+        let execution_mode = input.execution_mode();
+        let properties = PlanProperties::new(
+            eq_properties,
+            Partitioning::UnknownPartitioning(1),
```
phillipleblanc: This can match the input's Partitioning scheme

phillipleblanc (dismissed):

digadeesh (dismissed):

sgrebnov (approved):

PR #1271, opened 2024-05-01
Publish ODBC images

mach-kernel:
This change would allow users of the ODBC connector to have simpler Dockerfiles. 

For example, to install the sqlite ODBC driver:

```Dockerfile
FROM spiceai/spiceai:latest-odbc

RUN apt update \
    && apt install --yes libsqliteodbc --no-install-recommends \
    && rm -rf /var/lib/{apt,dpkg,cache,log}
```

It would also add `unixodbc` to all of the image variants. Can add a toggle to avoid installing the extra dependencies, but thought NBD:

```bash
$ container-diff diff daemon://spice-with-odbc spiceai/spiceai:latest

-----Size-----

Image size difference between spice-with-odbc and spiceai/spiceai:latest:
SIZE1         SIZE2
288.2M        271.3M
```


phillipleblanc (commented):
Thanks for the PR! I think it makes sense to have two Docker images, one with `odbc` and one without - but I would actually bucket them into a `full` image and a `slim` image, similar to what `node` does. Then the "main" image can be the full image with odbc and included tools, and the "slim" image can be built without odbc and tools. And over time we can make more decisions about what makes sense to go into a "slim" image.

.github/workflows/spiced_docker.yml
```diff
@@ -49,11 +49,30 @@ jobs:
...
+        id: docker_build_full
+        uses: docker/build-push-action@v5
+        with:
+          context: .
+          file: Dockerfile
+          platforms: linux/amd64,linux/arm64
+          push: ${{ startswith(github.ref, 'refs/tags/v') }}
+          build-args: |
+            REL_VERSION=${{ env.REL_VERSION }}
+            CARGO_FEATURES=odbc
```
sgrebnov, line 68: Please note, I've decided to keep/re-use CARGO_FEATURES parameter to specify what features we would like to include to full build (more control and flexibility). Alternatively we can have dedicated flag to control this

Dockerfile
```diff
@@ -6,30 +6,40 @@ FROM rust:${RUST_VERSION}-slim-bookworm as build
 USER root
 
 RUN apt update \
-    && apt install --yes pkg-config libssl-dev build-essential libsqlite3-dev cmake protobuf-compiler \
+    && apt install --yes pkg-config libssl-dev build-essential libsqlite3-dev cmake protobuf-compiler unixodbc-dev \
```
sgrebnov, line 9: For simplicity I keep `unixodbc-dev` always installed in build image, let me know if we should make it conditional (in theory this might make build slightly faster).

sgrebnov (dismissed):

.github/workflows/spiced_docker.yml
phillipleblanc: I think eventually we will have a `full` flag to control all of the features that go into the full image - but not needed for now.

phillipleblanc (approved):

PR #1272, opened 2024-05-01
Add Snowflake login support (username and password)

sgrebnov:
Implements https://github.com/spiceai/spiceai/issues/1273: login support for default Snowflake auth option: username and password

Note: Snowflake `account` parameter is planned to be part of dataset definition.


digadeesh (approved):
Code LGTM but please don't submit until you have a data connector to go with it.

PR #1278, opened 2024-05-02
Add Snowflake data connector

sgrebnov:
Implements Snowflake data connector basic functionality.

Example configuration:

```
datasets:
- from: snowflake:snowflake_sample_data.tpch_sf1.customer
  name: customer
  params: 
    account: ztyvqyb-vgb35424
    warehouse: COMPUTE_WH
```

Associated PR for login functionality: https://github.com/spiceai/spiceai/pull/1272

### DIFF
--- a/crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
+++ b/crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
@@ -36,11 +36,11 @@ pub enum Error {
     #[snafu(display("Not implemented"))]
     NotImplemented,
 
-    #[snafu(display("Unable to retrieve schema"))]
-    UnableToRetrieveSchema,
+    #[snafu(display("Unable to retrieve schema: {reason}"))]
+    UnableToRetrieveSchema { reason: String },
 
-    #[snafu(display("Unexpected query response, expected Arrow, got JSON"))]
-    UnexpectedResponse,
+    #[snafu(display("Unexpected query response, expected Arrow, got JSON: {json}"))]
+    UnexpectedResponse { json: String },
 
     #[snafu(display("Error executing query: {source}"))]
     SnowflakeQueryError {
@@ -86,7 +86,14 @@ impl<'a> AsyncDbConnection<Arc<SnowflakeApi>, &'a (dyn Sync)> for SnowflakeConne
                 let schema = record_batches[0].schema();
                 return Ok(Arc::clone(&schema));
             }
-            _ => UnableToRetrieveSchemaSnafu.fail()?,
+            snowflake_api::QueryResult::Empty => UnableToRetrieveSchemaSnafu {
+                reason: "query to get table schema returned no result".to_string(),
+            }
+            .fail()?,
+            snowflake_api::QueryResult::Json(json) => UnexpectedResponseSnafu {
+                json: json.to_string(),
+            }
+            .fail()?,
         }
     }
 
@@ -101,6 +108,18 @@ impl<'a> AsyncDbConnection<Arc<SnowflakeApi>, &'a (dyn Sync)> for SnowflakeConne
 
         match res {
             snowflake_api::QueryResult::Arrow(record_batches) => {
+                tracing::trace!(
+                    "Snowflake data connector: received {} record batches",
+                    record_batches.len()
+                );
+
+                if record_batches.is_empty() {
+                    return Ok(Box::pin(RecordBatchStreamAdapter::new(
+                        Arc::new(Schema::empty()),
+                        stream::empty(),
+                    )));
+                }
+
                 let schema = record_batches[0].schema();
                 return Ok(Box::pin(MemoryStream::try_new(
                     record_batches,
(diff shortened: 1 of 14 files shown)


crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
```diff
@@ -0,0 +1,127 @@
...
+        _: &[&'a (dyn Sync)],
+    ) -> Result<SendableRecordBatchStream> {
+        let sql = sql.to_string();
+
+        let res = self.api.exec(&sql).await.context(SnowflakeQuerySnafu)?;
+
+        match res {
+            snowflake_api::QueryResult::Arrow(record_batches) => {
+                let schema = record_batches[0].schema();
+                return Ok(Box::pin(MemoryStream::try_new(
```
y-f-u, line 124: is this streaming?

crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
```diff
@@ -0,0 +1,127 @@
...
+            }
+            snowflake_api::QueryResult::Empty => {
+                return Ok(Box::pin(RecordBatchStreamAdapter::new(
+                    Arc::new(Schema::empty()),
+                    stream::empty(),
+                )));
+            }
+            snowflake_api::QueryResult::Json(json) => {
+                tracing::debug!("Unexpected response from Snowflake query: {json}");
+                UnexpectedResponseSnafu.fail()?
```
y-f-u: passing json into the UnexpectedResponseSnafu so it has a more detailed error message rather than just `debug!`

crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
```diff
@@ -0,0 +1,127 @@
...
+            .exec(format!("SELECT * FROM {table} limit 1").as_str())
+            .await
+            .context(SnowflakeQuerySnafu)?;
+
+        match res {
+            snowflake_api::QueryResult::Arrow(record_batches) => {
+                let schema = record_batches[0].schema();
+                return Ok(Arc::clone(&schema));
+            }
+            _ => UnableToRetrieveSchemaSnafu.fail()?,
```
y-f-u: do we miss any information here why it fails?

crates/runtime/src/dataconnector/snowflake.rs
```diff
@@ -0,0 +1,92 @@
...
+use std::pin::Pin;
+use std::sync::Arc;
+use std::{collections::HashMap, future::Future};
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Unable to create flight client: {source}"))]
+    UnableToCreateFlightClient { source: flight_client::Error },
+
+    #[snafu(display("{source}"))]
```
y-f-u, line 37: Unable to get read provider: {source}

crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
sgrebnov: Data is returned from Snowflake post API as base64 encoded response which is then converted to Vec<RecordBatch> so the data/response is not a stream.

`MemoryStream` we create implements streaming. We use the same approach for most data connectors.

crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
sgrebnov: Updated, thank you!

crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
sgrebnov: Updated to include more details. Thank you!

```rust
match res {
      snowflake_api::QueryResult::Arrow(record_batches) => {
          let schema = record_batches[0].schema();
          return Ok(Arc::clone(&schema));
      }
      snowflake_api::QueryResult::Empty => UnableToRetrieveSchemaSnafu {
          reason: "query to get table schema returned no result".to_string(),
      }
      .fail()?,
      snowflake_api::QueryResult::Json(json) => UnexpectedResponseSnafu {
          json: json.to_string(),
      }
      .fail()?,
  }
```

crates/runtime/src/dataconnector/snowflake.rs
sgrebnov: Propose to currently keep as-is as this is what we do for all data connectors to make errors more user friendly. I'll test this and update if needed. Thank you.

crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
y-f-u: yeah, we are going to need to convert them to streaming approach for most of them.

y-f-u: Had a discussion with Sergei, seems the underlying snowflake-rs is pretty limited for now. Okay to go as it is.

crates/runtime/src/dataconnector/snowflake.rs
y-f-u: With a little bit more info like other display will help debug too. I think the user friendly error may happen on another interception layer.

y-f-u (approved):

PR #1285, opened 2024-05-02
Refactor duckdb to properly set access_mode for connection

### LINKED ISSUE #1264: The DuckDb file connection failed to establish due to an existing open write connection
**Describe the bug**

When connecting to duckdb database, which has open write connection, spice fails to load:

> 2024-05-01T06:30:13.025926Z  WARN runtime: Failed to get data connector from source for dataset data: Unable to initialize data connector duckdb: Unable to create DuckDB connection pool: DuckDBError: IO Error: Could not set lock on file "data.db": Conflicting lock is held in /opt/homebrew/Cellar/duckdb/0.10.1/bin/duckdb (PID 80599) by user evgenii. See also https://duckdb.org/docs/connect/concurrency

**To Reproduce**

- prepare persisted duck db dataset

```shell
duckdb data.db # open db
```

In duckdb shell: Load httpfs and create table from s3
```
INSTALL httpfs;
LOAD httpfs;
create table data as select * from read_parquet('s3://spiceai-demo-datasets/cleaned_sales_data.parquet') limit 100;
```

- configure spicepod with duckdb dataset

```yaml
version: v1beta1
kind: Spicepod
name: duckdb-connector
datasets:
  - from: duckdb:data
    name: data
    description: data from duckdb
    params:
      open: data.db
```

- keeping duckdb repl connected, start spice

CleanShot 2024-05-01 at 15 27 51@2x

**Expected behavior**
Spice should connect to duckdb in readonly mode

ewgenius:
closes #1264 

### Refactored DuckDB connection factory to utilize access_mode:

Note about concurrency - https://duckdb.org/docs/connect/concurrency
> One process can both read and write to the database.
Multiple processes can read from the database, but no processes can write ([access_mode = 'READ_ONLY'](https://duckdb.org/docs/configuration/overview#configuration-reference)).

- connector uses `read_only` connection
- accelerator uses `read_write` connection

### DIFF
--- a/crates/db_connection_pool/src/duckdbpool.rs
+++ b/crates/db_connection_pool/src/duckdbpool.rs
@@ -30,6 +30,9 @@ pub enum Error {
 
     #[snafu(display("ConnectionPoolError: {source}"))]
     ConnectionPoolError { source: r2d2::Error },
+
+    #[snafu(display("Unable to connect to DuckDB: {source}"))]
+    UnableToConnect { source: duckdb::Error },
 }
 
 pub struct DuckDbConnectionPool {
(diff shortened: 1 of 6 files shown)


digadeesh (dismissed):

crates/db_connection_pool/src/duckdbpool.rs
```diff
@@ -120,10 +89,14 @@ impl DbConnectionPool<r2d2::PooledConnection<DuckdbConnectionManager>, &'static
...
 
-fn get_duckdb_file(name: &str, params: &Arc<Option<HashMap<String, String>>>) -> String {
-    params
-        .as_ref()
-        .as_ref()
-        .and_then(|params| params.get("duckdb_file").cloned())
-        .unwrap_or(format!("{name}.db"))
+fn get_config(access_mode: &AccessMode) -> Result<duckdb::Config> {
+    let config = duckdb::Config::default()
+        .access_mode(match access_mode {
```
Sevenannn, line 127: Why matching instead of directly passing access_mode into duckdb::Config::default().access_mode() here

ewgenius: `access_mode` passed as reference (and in other places above), and can't be dereferenced here

CleanShot 2024-05-03 at 22 19 26@2x

I've tried different ways, and just matching over the reference is the most convenient way in this case

digadeesh (approved):

Issue #1263, opened 2024-05-01
Stopping spice and restarting it with file based accelerator gives a weird warning and then repl doesn't work.  


digadeesh:
**Describe the bug**
Cxreating a duckdb acclerated dataset using the file based option doesn't work after you restart spice.  The dataset won't load after restarting and the REPL will return errors.

Spice Runtime Warning:
2024-05-01T00:29:58.215237Z WARN runtime: Failed to initialize data connector for dataset sample_table: Unable to attach data connector clickhouse: Unable to create dataset acceleration: Acceleration creation failed: Unable to create table: External error: Unable to create duckdb table: Catalog Error: Table with name "sample_table" already exists!

REPL won't start.  Error message:
Flight REPL error: transport error

**To Reproduce**
1. Create a dataset that has a duckdb acceleration enabled, and set it to use the file based mode.  
2. Start spice to load the dataset
3. Observe that the duckdb .db file is created in the specified directory
4. Stop Spice
5. Start Spice
6. Observe that the runtime can't load the dataset, and the REPL returns errors (errors above in description.

**Expected behavior**
Dataset loads and refreshes.  Until the dataset is fully refreshed, the SQL repl would query the contents in the .db file.

digadeesh:
fixed with https://github.com/spiceai/spiceai/pull/1290

PR #1290, opened 2024-05-03
Add if not exists clause to create statement on when creating a table using duckdb acceleration.

digadeesh:
In the case that we're using file based duckdb acceleration, we get an error if we start spice, stop it, and restart, because the table already exists.

For other persistent stores, postgres and sqlite, we check for existence.  If there is a change in schema in the data connector, it'll still error out.  But thats OK since a schema change would probably be a situation where you'd want the user to think and take action before erasing existing data.


digadeesh:
Fixes: https://github.com/spiceai/spiceai/issues/1263

phillipleblanc (approved):

Issue #1270, opened 2024-05-01
Wrap and handle postgres error messages and pass through the context


ewgenius:
- [ ] Misconfigured host or port name, misconfigured username/password:: #1292
- [ ] Misconfigured dataset name in "from": #1295
- [ ] pg_sslmode set to enable:

PR #1292, opened 2024-05-03
Postgres data connector - handle invalid host/port and username/password errors

ewgenius:
Part of #1270 

- Misconfigured host or port name
CleanShot 2024-05-06 at 12 45 27@2x

- Misconfigured username/password
CleanShot 2024-05-06 at 12 47 34@2x

### DIFF
--- a/crates/db_connection_pool/src/postgrespool.rs
+++ b/crates/db_connection_pool/src/postgrespool.rs
@@ -50,7 +50,7 @@ pub enum Error {
     #[snafu(display("Invalid parameter: {parameter_name}"))]
     InvalidParameterError { parameter_name: String },
 
-    #[snafu(display("Cannot connect to PostgreSQL on {host}:{port}."))]
+    #[snafu(display("Cannot connect to PostgreSQL on {host}:{port}. Ensure that the host and port are correctly configured, and that the host is reachable."))]
     InvalidHostOrPortError {
         source: ns_lookup::Error,
         host: String,
--- a/crates/db_connection_pool/src/postgrespool.rs
+++ b/crates/db_connection_pool/src/postgrespool.rs
@@ -240,10 +240,6 @@ async fn verify_postgres_config(config: &Config) -> Result<()> {
             if let Host::Tcp(host) = host {
                 verify_ns_lookup_and_tcp_connect(host, *port)
                     .await
-                    .map_err(|err| {
-                        tracing::debug!("Failed to connect to host: {}", err);
-                        err
-                    })
                     .context(InvalidHostOrPortSnafu { host, port: *port })?;
             }
         }
--- a/crates/db_connection_pool/src/postgrespool.rs
+++ b/crates/db_connection_pool/src/postgrespool.rs
@@ -85,7 +85,6 @@ impl PostgresConnectionPool {
     /// # Errors
     ///
     /// Returns an error if there is a problem creating the connection pool.
-    #[allow(clippy::too_many_lines)]
     pub async fn new(
         params: Arc<Option<HashMap<String, String>>>,
         secret: Option<Secret>,
@@ -241,7 +240,7 @@ async fn test_postgres_connection(
                 }
             }
 
-            return Err(Box::new(Error::PostgresConnectionError { source: err }));
+            Err(Box::new(Error::PostgresConnectionError { source: err }))
         }
     }
 }
--- a/crates/db_connection_pool/src/postgrespool.rs
+++ b/crates/db_connection_pool/src/postgrespool.rs
@@ -68,6 +68,11 @@ pub enum Error {
 
     #[snafu(display("Postgres connection error: {source}"))]
     PostgresConnectionError { source: tokio_postgres::Error },
+
+    #[snafu(display(
+        "Authentication failed. Ensure that the username and password are correctly configured."
+    ))]
+    InvalidUsernameOrPassword { source: tokio_postgres::Error },
 }
 
 pub struct PostgresConnectionPool {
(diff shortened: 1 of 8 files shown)


crates/runtime/src/dataconnector.rs
```diff
@@ -138,10 +142,32 @@ pub async fn create_new_connector(
...
+                    #[cfg(feature = "postgres")]
+                    if let Some(postgrespool::Error::InvalidHostOrPortError {
+                        source: _,
+                        host,
+                        port,
+                    }) = err_source.downcast_ref::<postgrespool::Error>()
+                    {
+                        return Some(Err(Box::new(Error::UnableToCreateDataConnector {
+                                message: format!(
+                                    "Cannot connect to PostgreSQL data connector on {host}:{port}. Ensure that the host and port are correclty configured in the spicepod, and that the host is reachable"
```
phillipleblanc:
```suggestion
                                    "Cannot connect to PostgreSQL data connector on {host}:{port}. Ensure that the host and port are correclty configured in the spicepod, and that the host is reachable."
```

crates/runtime/src/dataconnector.rs
```diff
@@ -91,6 +92,9 @@ pub enum Error {
     UnableToFilterDataFrame {
         source: datafusion::error::DataFusionError,
     },
+
+    #[snafu(display("{message}"))]
+    UnableToCreateDataConnector { message: String },
```
phillipleblanc: If the only usage of this error is to report that we couldn't connect - I think it would be better to bring the error message into this, rename it to something specific and only pass the two properties needed to display in the error (host/port)

ewgenius: Yes, but this is a generic error for all data connectors, and in general case it can have only common prefix, but reason will be different in each case

phillipleblanc (dismissed):

crates/db_connection_pool/src/postgrespool.rs
```diff
@@ -79,6 +80,7 @@ impl PostgresConnectionPool {
     /// # Errors
     ///
     /// Returns an error if there is a problem creating the connection pool.
+    #[allow(clippy::too_many_lines)]
```
phillipleblanc: 🤨

crates/db_connection_pool/src/postgrespool.rs
```diff
@@ -157,7 +159,11 @@ impl PostgresConnectionPool {
                 if let Host::Tcp(host) = host {
                     verify_ns_lookup_and_tcp_connect(host, *port)
                         .await
-                        .context(InvalidHostOrPortSnafu)?;
+                        .map_err(|err| {
+                            tracing::debug!("Failed to connect to host: {}", err);
```
Sevenannn: Do we need tracing here? it doesn't print any log and the InvalidHostOrPortSnafu will already show the error in user logs

crates/runtime/src/dataconnector.rs
```diff
@@ -138,10 +142,32 @@ pub async fn create_new_connector(
...
+                    #[cfg(feature = "postgres")]
+                    if let Some(postgrespool::Error::InvalidHostOrPortError {
+                        source: _,
+                        host,
+                        port,
+                    }) = err_source.downcast_ref::<postgrespool::Error>()
+                    {
+                        return Some(Err(Box::new(Error::UnableToCreateDataConnector {
+                                message: format!(
+                                    "Cannot connect to PostgreSQL data connector on {host}:{port}. Ensure that the host and port are correclty configured in the spicepod, and that the host is reachable."
```
phillipleblanc:
```suggestion
                                    "Cannot connect to PostgreSQL data connector on {host}:{port}. Ensure that the host and port are correctly configured in the spicepod, and that the host is reachable."
```

phillipleblanc (dismissed):

crates/db_connection_pool/src/postgrespool.rs
ewgenius: we use [default 100 lines config](https://doc.rust-lang.org/nightly/clippy/lint_configuration.html#too-many-lines-threshold), and allow that in several places. Maybe makes sense to extend the limit?

crates/db_connection_pool/src/postgrespool.rs
ewgenius: you're right, it is useless here, removed

crates/runtime/src/dataconnector.rs
```diff
@@ -138,10 +142,32 @@ pub async fn create_new_connector(
...
-    match connector_factory {
-        Some(factory) => Some(factory(secret, params).await),
-        None => None,
+    if let Some(create_future) = connector_factory {
+        match create_future(secret, params).await {
+            Ok(connector) => return Some(Ok(connector)),
+            Err(err) => {
+                if let Some(err_source) = err.source() {
+                    #[cfg(feature = "postgres")]
+                    if let Some(postgrespool::Error::InvalidHostOrPortError {
```
y-f-u: Just wonder if this level of details can be pushed down into the postgres.rs.

crates/db_connection_pool/src/postgrespool.rs
phillipleblanc: I think we should follow this lint. The other places that have allowed it should be reworked.

ewgenius: will be partially refactored in #1300

crates/runtime/src/dataconnector.rs
phillipleblanc: Ah yeah, this is a bad pattern to set imo. We should have the dataconnector abstraction be unaware of the specific implementations.

crates/db_connection_pool/src/postgrespool.rs
ewgenius: refactored

crates/runtime/src/dataconnector.rs
ewgenius: @phillipleblanc I don't think we should to put spicepod specific errors into pool implementation

ewgenius: updated error messages to be out of spicepod scope, and moved outside of dataconnector

phillipleblanc (approved):

PR #1300, opened 2024-05-05
Fix unsupported SSL mode options for PostgreSQL connection string

ewgenius:
Removed invalid ssl_mode from postgres connection string.

Spice use `"disable" | "require" | "prefer" | "verify-ca" | "verify-full"` ssl modes, but underlying tokio-postgres supports only `disable`, `require` and `prefer`.

Supported mode passed to the connection string here:
https://github.com/spiceai/spiceai/blob/02801593ff48c1b9af8a6380c065d143b63eed40/crates/db_connection_pool/src/postgrespool.rs#L144-L151

Also updated the case with `pg_connection_string` param, to extract sslmode and sslrootcert to process separately.

### DIFF
--- a/crates/db_connection_pool/src/postgrespool.rs
+++ b/crates/db_connection_pool/src/postgrespool.rs
@@ -160,7 +160,7 @@ impl PostgresConnectionPool {
 
         connection_string.push_str(format!("sslmode={mode} ").as_str());
         let config = Config::from_str(connection_string.as_str()).context(ConnectionPoolSnafu)?;
-        verify_posstgres_config(&config).await?;
+        verify_postgres_config(&config).await?;
 
         let mut certs: Option<Vec<Certificate>> = None;
 
--- a/crates/db_connection_pool/src/postgrespool.rs
+++ b/crates/db_connection_pool/src/postgrespool.rs
@@ -232,7 +232,7 @@ async fn test_postgres_connection(
     Ok(())
 }
 
-async fn verify_posstgres_config(config: &Config) -> Result<()> {
+async fn verify_postgres_config(config: &Config) -> Result<()> {
     for host in config.get_hosts() {
         for port in config.get_ports() {
             if let Host::Tcp(host) = host {
--- a/crates/db_connection_pool/src/postgrespool.rs
+++ b/crates/db_connection_pool/src/postgrespool.rs
@@ -79,7 +79,6 @@ impl PostgresConnectionPool {
     /// # Errors
     ///
     /// Returns an error if there is a problem creating the connection pool.
-    #[allow(clippy::too_many_lines)]
     pub async fn new(
         params: Arc<Option<HashMap<String, String>>>,
         secret: Option<Secret>,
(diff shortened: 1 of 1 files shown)


crates/db_connection_pool/src/postgrespool.rs
```diff
@@ -119,7 +119,6 @@ impl PostgresConnectionPool {
                     match pg_sslmode.to_lowercase().as_str() {
                         "disable" | "require" | "prefer" | "verify-ca" | "verify-full" => {
                             ssl_mode = pg_sslmode.as_str();
-                            connection_string.push_str(format!("sslmode={ssl_mode} ").as_str());
```
y-f-u, line 122: it's not related to this line. But if connection_string has ssl_mode defined, that could be a double configuration for the ssl in the final connection_string?

ewgenius: It was the case before this fix, but now the only place where sslmode added to the connection string is set is here

https://github.com/spiceai/spiceai/blob/02801593ff48c1b9af8a6380c065d143b63eed40/crates/db_connection_pool/src/postgrespool.rs#L151

ewgenius: Updated the case with `pg_connection_string` to parse and extract sslmode and sslrootcert parameters, to process sll connection properly

y-f-u (dismissed):

crates/db_connection_pool/src/postgrespool.rs
```diff
@@ -96,7 +97,31 @@ impl PostgresConnectionPool {
                 "pg_connection_string_key",
                 "pg_connection_string",
             ) {
-                connection_string.push_str(pg_connection_string.as_str());
+                // parse and extract ssl_mode and ssl_rootcert_path from connection string
+                let str = pg_connection_string.clone();
```
phillipleblanc: This would be cleaner if we refactored this to a separate method (and will also fix the clippy lint warning)

crates/db_connection_pool/src/postgrespool.rs
```diff
@@ -79,12 +79,13 @@ impl PostgresConnectionPool {
     /// # Errors
     ///
     /// Returns an error if there is a problem creating the connection pool.
+    #[allow(clippy::too_many_lines)]
```
phillipleblanc:
```suggestion
```

crates/db_connection_pool/src/postgrespool.rs
ewgenius: refactored

y-f-u (dismissed):

phillipleblanc (dismissed):

crates/db_connection_pool/src/postgrespool.rs
```diff
@@ -198,6 +194,58 @@ impl PostgresConnectionPool {
...
+    connector: MakeTlsConnector,
+) -> Result<()> {
+    let (_, _) = tokio_postgres::connect(connection_string, connector)
+        .await
+        .context(PostgresConnectionSnafu)?;
+
+    Ok(())
+}
+
+async fn verify_posstgres_config(config: &Config) -> Result<()> {
```
phillipleblanc:
```suggestion
async fn verify_postgres_config(config: &Config) -> Result<()> {
```

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #1301, opened 2024-05-05
Check only the latest published release to avoid installing pre-release versions

### LINKED ISSUE #1238: Installation script is failing, during new release publish process
**Describe the bug**

Running `curl https://install.spiceai.org | /bin/bash` fails if new release is in process of publishing:

```
Run curl https://install.spiceai.org/ | /bin/bash
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  70[5](https://github.com/spiceai/spice.js/actions/runs/8887208084/job/24402540770#step:10:6)5  100  7055    0     0  28747      0 --:--:-- --:--:-- --:--:-- 28795
Getting the latest Spice.ai CLI...
Installing Spice CLI v0.12.0-alpha ...
Downloading https://github.com/spiceai/spiceai/releases/download/v0.12.0-alpha/spice_linux_x8[6](https://github.com/spiceai/spice.js/actions/runs/8887208084/job/24402540770#step:10:7)_64.tar.gz ...
ERROR: version not found 
Failed to install Spice CLI
For support, see https://docs.spiceai.org
```

This is because we're publishing pre-release tag before building and uploading the binaries, and install scripts just gets most recent tag
https://github.com/spiceai/spiceai/blob/eec359d2dfbe3ffdd7c1032647c69840a201d2d6/install/install.sh#L95-L106

**Expected behavior**

Spice installation should install mosst recent **published** release at any time.

ewgenius:
Using [`releases/latest`](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release) to get only most recent **published**, non-prerelease version

Currently install script gets all available releases, including pre-releases, which breaks installation flow during the time when we publish new release (pre-release and tag published in github, but binaries not ready yet)

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -93,7 +93,7 @@ checkJqInstalled() {
 }
 
 getLatestRelease() {
-    local spiceReleaseUrl="https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/releases"
+    local spiceReleaseUrl="https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/releases/latest"
     local latest_release=""
 
     if [ "$SPICE_HTTP_REQUEST_CLI" == "curl" ]; then
--- a/install/install-spiced.sh
+++ b/install/install-spiced.sh
@@ -91,7 +91,7 @@ checkJqInstalled() {
 }
 
 getLatestRelease() {
-    local spiceReleaseUrl="https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/releases"
+    local spiceReleaseUrl="https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/releases/latest"
     local latest_release=""
 
     if [ "$SPICE_HTTP_REQUEST_CLI" == "curl" ]; then


y-f-u (approved):

install/install.sh
```diff
@@ -93,7 +93,7 @@ checkJqInstalled() {
 }
 
 getLatestRelease() {
-    local spiceReleaseUrl="https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/releases"
+    local spiceReleaseUrl="https://api.github.com/repos/${GITHUB_ORG}/${GITHUB_REPO}/releases/latest"
     local latest_release=""
```
phillipleblanc, line 119: Does the logic below to set `latest_release` still work with this change? `/releases` returns an array and `/releases/latest` returns just an object.

ewgenius: Yes, I've checked that - code below just extracts first `tag_name` with assertions from the output, it doesn't require it to be an array

Example:

CleanShot 2024-05-06 at 11 24 26@2x

ewgenius: For spiceai repo

CleanShot 2024-05-06 at 11 29 08@2x

phillipleblanc (approved):

PR #1302, opened 2024-05-05
Use the `GITHUB_TOKEN` environment variable in the installation script, if available, to avoid rate limiting in CI workflows

ewgenius:
Updated calls to github api from installation scripts, to use GITHUB_TOKEN env variable when it is set.

test run - https://github.com/spiceai/spiceai/actions/runs/8958831027

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -92,14 +92,36 @@ checkJqInstalled() {
     fi
 }
 
+function gh_curl() {
+    if [ -z "$GITHUB_TOKEN" ]
+    then
+        curl \
+            $@
+    else
+        curl -H "Authorization: token $GITHUB_TOKEN" \
+            $@
+    fi
+}
+
+function gh_wget() {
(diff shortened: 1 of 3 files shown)


install/install.sh
```diff
@@ -92,14 +92,36 @@ checkJqInstalled() {
...
+        curl \
+            $@
+    else
+        curl -H "Authorization: token $GITHUB_TOKEN" \
+            $@
+    fi
+}
+
+function gh_wget() {
+    if [ -n "$GITHUB_TOKEN" ]
```
y-f-u: above using -z, this is using -n. Probably just use same for consistency.

y-f-u (dismissed):

install/install.sh
ewgenius: ah, right

ewgenius: fixed

phillipleblanc (approved):

lukekim (approved):

Issue #1305, opened 2024-05-06
Snowflake key-pair auth support


sgrebnov:

sgrebnov:
Implemented and documented

PR #1314, opened 2024-05-06
Add support for Snowflake key-pair authentication

sgrebnov:
Part of https://github.com/spiceai/spiceai/issues/1305

PR adds support for [Snowflake key-pair auth](https://docs.snowflake.com/en/user-guide/key-pair-auth) via `snowflake_auth_type: keypair`and `snowflake_private_key_path: <path>` params. Both unencrypted and encrypted (via passphrase) keys are supported.

``` yaml
version: v1beta1
kind: Spicepod
name: test
datasets:
- from: snowflake:snowflake_sample_data.tpch_sf1.lineitem
  name: lineitem
  params:
    snowflake_auth_type: keypair
    snowflake_private_key_path: /Users/name/rsa_key.p8
    snowflake_private_key_passphrase: 12345
```

Remaining steps for full key-pair authentication support:
1.  Done ~Add support for encrypted private keys (`passphrase/password` parameter and logic to decrypt the key)~
2. Add key-pair support to Spice CLI (key-pair login for Snowflake)


crates/db_connection_pool/src/snowflakepool.rs
```diff
@@ -124,6 +113,74 @@ impl SnowflakeConnectionPool {
...
+        .context(MissingRequiredSecretSnafu { name: "username" })?;
+    let account = get_param(params, secret, "account")
+        .context(MissingRequiredSecretSnafu { name: "account" })?;
+    // account identifier can be in <orgname.account_name> format but API requires it as <orgname-account_name>
+    let account = account.replace('.', "-");
+
+    let warehouse = get_param(params, secret, "snowflake_warehouse");
+    let role = get_param(params, secret, "snowflake_role");
+    let auth_type = get_param(params, secret, "snowflake_auth_type")
+        .unwrap_or_else(|| "auth_snowflake".to_string())
```
sgrebnov: `auth_snowflake` for default username/password auth was selected to be consistent with Snowflake naming: it is called `snowflake` in `snowsql` tool, `go driver`, ADBC driver, etc

phillipleblanc: Then it should just be `snowflake` then, not `auth_snowflake`?

crates/db_connection_pool/src/snowflakepool.rs
```diff
@@ -124,6 +113,74 @@ impl SnowflakeConnectionPool {
...
+            &username,
+            role.as_deref(),
+            &password,
+        )
+        .context(UnableToConnectSnafu)?;
+
+        return Ok(api);
+    }
+
+    if auth_type == "auth_keypair" {
```
phillipleblanc:
```suggestion
    if auth_type == "keypair" {
```

crates/db_connection_pool/src/snowflakepool.rs
sgrebnov: @phillipleblanc - thank you - I'll update to just `snowflake` (it is what `snowsql` is using). Snowflake ADBC driver naming is `auth_snowflake`, but just `snowflake` is better IMO (same for `keypair`)
image
image

crates/db_connection_pool/src/snowflakepool.rs
sgrebnov: Updated - thank you

crates/db_connection_pool/src/snowflakepool.rs
sgrebnov: Updated - thank you

crates/db_connection_pool/src/snowflakepool.rs
```diff
@@ -124,6 +123,86 @@ impl SnowflakeConnectionPool {
...
+            warehouse.as_deref(),
+            None,
+            None,
+            &username,
+            role.as_deref(),
+            &password,
+        )
+        .context(UnableToConnectSnafu)?;
+
+        return Ok(api);
```
phillipleblanc: I would refactor this and the block after `if auth_type == "keypair"` into separate methods.

phillipleblanc (dismissed):

crates/db_connection_pool/src/snowflakepool.rs
sgrebnov: Refactored to have separate methods - `init_snowflake_api_with_password_auth` and `init_snowflake_api_with_keypair_auth`

phillipleblanc (approved):

PR #1326, opened 2024-05-07
using data fusion listing table to refactor primitive object store implementation of s3 connector

### LINKED ISSUE #1243: FallbackOnZeroResults doesn't work when the federated source is an S3 bucket
**Describe the bug**
This Spicepod will produce an error:

```yaml
version: v1beta1
kind: Spicepod
name: fallback_demo
datasets:
  name: taxi_trips
  acceleration:
    enabled: true
    refresh_sql: |
      SELECT * FROM taxi_trips WHERE passenger_count = 1;
    on_zero_results: use_source
```

with the query: `select tpep_pickup_datetime, passenger_count from taxi_trips where passenger_count = 9;`

### LINKED ISSUE #1322: Enhancement: Use listing table to replace object store for S3 connector

y-f-u:
fixes #1243 

Also fixes #1322 

This uses ListingTable and object store registry from data fusion to create a more unified architecture for all connectors we have.

This also opens up the possibility to easily extend s3 connector to other cloud store and other formats.


crates/runtime/src/dataconnector/s3.rs
```diff
@@ -93,74 +97,69 @@ impl DataConnectorFactory for S3 {
...
-                };
-                if let Some(secret) = secret.get("secret") {
-                    s3_builder = s3_builder.with_secret_access_key(secret);
-                };
-            } else {
-                s3_builder = s3_builder.with_skip_signature(true);
+impl S3 {
+    fn get_object_store_url(&self, dataset: &Dataset) -> AnyErrorResult<Url> {
+        let mut fragments = vec![];
+        let mut query = form_urlencoded::Serializer::new(String::new());
```
y-f-u: use encoded url form query in the fragment to make sure data can be passed through correctly + not polluting the listing table to misuse the query parameters.

crates/runtime/src/lib.rs
```diff
@@ -54,6 +54,7 @@ pub mod model;
 pub mod modelformat;
 pub mod modelruntime;
 pub mod modelsource;
+pub mod object_store_registry;
```
y-f-u, line 57: we have different type of naming convention, e.g. "accelerated_table" or "modelruntime". underscore is clearer imo

crates/runtime/src/dataconnector/s3.rs
phillipleblanc:
```suggestion
        let mut fragment_builder = form_urlencoded::Serializer::new(String::new());
```

This is really a fragment builder

crates/runtime/src/object_store_registry.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+impl SpiceObjectStoreRegistry {
+    #[must_use]
+    pub fn new() -> Self {
+        SpiceObjectStoreRegistry::default()
+    }
+
+    #[allow(clippy::unused_self)]
+    fn get_feature_store(&self, url: &Url) -> datafusion::error::Result<Arc<dyn ObjectStore>> {
```
phillipleblanc:
```suggestion
    fn get_feature_store(url: &Url) -> datafusion::error::Result<Arc<dyn ObjectStore>> {
```

And then to call it, use `Self::get_feature_store(url)?;` below

phillipleblanc (dismissed):
Much cleaner!

crates/runtime/src/object_store_registry.rs
y-f-u: I left it so that in future we can pass in some context here - E.g. configuration stuff.

I guess we can refactor it to Self:: and change it when it is needed.

phillipleblanc: Yeah - I think that is cleaner for now (refactor to use Self::)

ewgenius (approved):

PR #1335, opened 2024-05-08
Move model code into separate, feature-flagged crate

Jeadie:
## Changes
- New feature-flagged crate `model_components`, flagged on `models`. 
- Takes model code from `runtime` crate. 
- New make target `make install-with-models`
- Reduces binary size from 152 mb -> 138 mb
- When not built with `models`, model HTTP endpoints aren't supported (e.g. `GET v1/models`). 
- Note: No functional changes are expected, except for:
    - Infer model format from model path (i.e. the path downloaded from ModelSource).
    - Infer what runtime to use for model format


crates/models/src/model.rs
```diff
@@ -0,0 +1,95 @@
+/*
```
Jeadie: This is mainly the same as `crates/runtime/model.rs`.

crates/runtime/src/model.rs
```diff
@@ -14,137 +14,38 @@ See the License for the specific language governing permissions and
...
+use crate::DataFusion;
 
-    for source in &sources {
-        if from.starts_with(source) {
-            match from.find(':') {
-                Some(index) => return from[index + 1..].to_string(),
-                None => return from.to_string(),
-            }
-        }
+pub async fn run(
```
Jeadie: Convenient function for `runtime` crate so that the `models` crate can continue to be ignorant of Datafusion.

crates/runtime/Cargo.toml
```diff
@@ -38,6 +38,7 @@ tract-onnx = "0.21.0"
 ndarray = "0.15.3"
 duckdb = { workspace = true, features = ["bundled", "r2d2", "vtab", "vtab-arrow"], optional = true }
 sql_provider_datafusion = { path = "../sql_provider_datafusion", optional = true }
+models = { path = "../models"}
```
```suggestion
models = { path = "../models", optional = true}
```

crates/runtime/Cargo.toml
```diff
@@ -109,3 +110,4 @@ spark = ["data_components/spark_connect"]
 dremio = []
 odbc = ["db_connection_pool/odbc", "data_components/odbc", "dep:odbc-api"]
 snowflake = ["dep:snowflake-api", "db_connection_pool/snowflake", "data_components/snowflake"]
+models = []
```
```suggestion
models = ["dep:models"]
```

Makefile
```diff
@@ -78,6 +78,10 @@ install: build
 	install -m 755 target/release/spice ~/.spice/bin/spice
 	install -m 755 target/release/spiced ~/.spice/bin/spiced
 
+.PHONY: install-full
```
sgrebnov: Better to have task name with `models` in name (`install-with-models`?). otherwise it can confuse as it is not actually a full/all features installation

crates/runtime/src/http/v1.rs
```diff
@@ -557,11 +557,10 @@ pub(crate) mod models {
         Extension,
     };
     use csv::Writer;
+    use models::model::Model;
```
ewgenius: shouldn't it be under a feature?

ewgenius: and in other places as well

Jeadie: yes sorry, i undrafted early

phillipleblanc (approved):

PR #1336, opened 2024-05-08
Select columns from Spark Dataframe according to projected_schema

### LINKED ISSUE #1307: Selecting a subset of columns returns a schema mismatch error in Spark/Databricks Data Connector
**Describe the bug**
Selecting a subset of columns from a table sourced from Databricks/Spark will produce the following error:

`select log_index, transaction_hash, transaction_index, address, data, block_timestamp, block_hash, block_number from eth_logs;`

`Status { code: Internal, message: "Expected and actual number of fields in the query result don't match: expected 8, received 9", source: None }`

**To Reproduce**
Login to a Sparks cluster with `spice login spark --spark_remote <remote>`. For a Databricks Spark cluster, this can be of the format `sc://{databricks_host}:443/;user_id=spice.ai;session_id={Uuid::v4()};token={token};x-databricks-cluster-id={cluster_id}`

Use a spicepod that connects to a dataset and then select a subset of the columns from that dataset.

`spicepod.yaml`
```yaml
version: v1beta1
kind: Spicepod
name: spark

datasets:
- from: spark:eth.default.logs
  name: eth_logs
```

**Expected behavior**
The query is successful and returns only the selected columns.

Sevenannn:
- Close https://github.com/spiceai/spiceai/issues/1307

image

### DIFF
--- a/crates/data_components/src/spark_connect.rs
+++ b/crates/data_components/src/spark_connect.rs
@@ -25,6 +25,7 @@ use datafusion::{
 };
 use futures::Stream;
 use spark_connect_rs::{
+    functions as F,
     spark::{data_type, DataType},
     DataFrame, SparkSession, SparkSessionBuilder,
 };
@@ -343,12 +344,19 @@ impl ExecutionPlan for SparkConnectExecutionPlan {
         tracing::trace!("projected_schema {:#?}", self.projected_schema);
         tracing::trace!("sql columns {:#?}", columns);
         tracing::trace!("filters {:#?}", self.filters);
+        let filtered_columns = self
+            .projected_schema
+            .fields()
+            .iter()
+            .map(|f| F::col(f.name()))
+            .collect::<Vec<_>>();
         let df = self
             .filters
             .iter()
             .fold(self.dataframe.clone(), |df, filter| {
                 df.filter(filter.as_str())
-            });
+            })
+            .select(filtered_columns);
         let df = match self.limit {
             Some(limit) => df.limit(limit),
             None => df,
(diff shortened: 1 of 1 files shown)


crates/data_components/src/spark_connect.rs
```diff
@@ -343,12 +344,19 @@ impl ExecutionPlan for SparkConnectExecutionPlan {
         tracing::trace!("projected_schema {:#?}", self.projected_schema);
         tracing::trace!("sql columns {:#?}", columns);
         tracing::trace!("filters {:#?}", self.filters);
+        let filtered_columns = self
```
phillipleblanc: So after reviewing - the `columns` above aren't being used by anything. Let's remove them, so we have something that looks like:

```rust
use spark_connect_rs::{
    functions::col,
    spark::{data_type, DataType},
    DataFrame, SparkSession, SparkSessionBuilder,
};
...
        let filtered_columns = self
            .projected_schema
            .fields()
            .iter()
            .map(|f| col(f.name()))
            .collect::<Vec<_>>();
        tracing::trace!("projected_schema {:#?}", self.projected_schema);
        tracing::trace!("sql columns {:#?}", filtered_columns);
        tracing::trace!("filters {:#?}", self.filters);
        let df = self
            .filters
            .iter()
            .fold(self.dataframe.clone(), |df, filter| {
                df.filter(filter.as_str())
            })
            .select(filtered_columns);
```

Sevenannn: That sounds good, let me change to that

phillipleblanc (approved):

PR #1341, opened 2024-05-09
fix spark connect list

### LINKED ISSUE #1258: List data types return an error in Spark/Databricks Data Connector
**Describe the bug**
Running this query on a list data type produces an error:

```sql
sql> select topics from logs limit 10
Error receiving data: Decode error: Error decoding ipc RecordBatch: Invalid argument error: First offset 210 of List(Field { name: "", data_type: Utf8, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }) is larger than values length 10
```

**To Reproduce**
Connect to a dataset that has a list column (i.e. eth.logs) and do the above query on it.

**Expected behavior**
The query returns successfully

y-f-u:
This fixes #1258 

Spark connect list is working now

Screenshot 2024-05-09 at 10 10 51

### DIFF
--- a/crates/data_components/src/spark_connect.rs
+++ b/crates/data_components/src/spark_connect.rs
@@ -83,7 +83,7 @@ async fn get_table_provider(
     let dataframe = spark_session.table(table_reference.table())?;
     let schema = dataframe.clone().schema().await?;
     let arrow_schema = datatype_as_arrow_schema(schema)?;
-    Ok(Arc::new(SparkConnectTablePovider {
+    Ok(Arc::new(SparkConnectTableProvider {
         dataframe,
         schema: arrow_schema,
     }))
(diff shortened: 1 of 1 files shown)


crates/data_components/src/spark_connect.rs
```diff
@@ -127,8 +127,8 @@ fn arrow_field_datatype_from_spark_connect_field_datatype(
                 Some(data_type) => {
                     let arrow_inner_type =
                         arrow_field_datatype_from_spark_connect_field_datatype(Some(*data_type))?;
-                    // Very smelly
-                    let field = Field::new("", arrow_inner_type, false);
+                    // The default name for the field is element
```
phillipleblanc: I don't think this is a general purpose solution, see https://discord.com/channels/885562378132000778/1166447479609376850/1237526038700101644

phillipleblanc: > Have any of you used DataFusion in databricks reading from an existing delta lake table that was written using pyspark? I'm trying to do so and getting annoying errors about the schemas not matching. ArrowException: Invalid argument error: column types must match schema types, expected List(Field { name: "item", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }) but found List(Field { name: "element", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }) at column index 10. It looks to me like there's a difference in how the data are stored. It's a list of strings and the "name" of the field is "item" vs "element".

y-f-u: it's actually from spark-connect-rs, which seems to be always elements from what I can tell. But realise this is not needed at all.

phillipleblanc (approved):
Nice! Love removing unnecessary code.

PR #1344, opened 2024-05-09
s3 csv support to validate the listing table extensibility

y-f-u:
- **support csv file format with basic options**

```
version: v1beta1
kind: Spicepod
name: spiceai
datasets:
  - from: s3://spiceai-demo-datasets/taxi_trips_csv/2024/
    name: taxi_trips
    acceleration:
      enabled: true
    params:
      file_format: csv
```

```
show tables; -- list available tables
sql> select count(1) from taxi_trips;
+-----------------+
| COUNT(Int64(1)) |
+-----------------+
| 2964624         |
+-----------------+

Time: 0.018100417 seconds. 1 rows.
```


crates/runtime/src/dataconnector/s3.rs
```diff
@@ -124,6 +129,52 @@ impl S3 {
 
         Ok(s3_url)
     }
+
+    fn get_file_format_and_extension(&self) -> AnyErrorResult<(Arc<dyn FileFormat>, String)> {
+        let params = &self.params;
+        match params.get("file_format") {
```
phillipleblanc: This code is very nested - is there a way to improve it? I think just creating separate functions for handling the CSV and Parquet match arms would be enough.

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #1355, opened 2024-05-09
FTP connector

gloomweaver:
Can check using this config
```
from: ftp://eu-central-1.sftpcloud.io/taxi_trips/2024/
name: taxi_trips
params:
  ftp_port: 21
  ftp_user: 054e99c4b547435282d3f9a2bb0b9143
  ftp_pass: ysrqM6BIfAsIR0glZPl4uyUqAJuB5KMc
  file_format: csv
description: eth recent blocks
acceleration:
  enabled: true
  refresh_mode: full
  refresh_check_interval: 300s
```


crates/runtime/src/dataconnector/ftp.rs
```diff
@@ -0,0 +1,189 @@
...
+                Arc::new(ParquetFormat::default()),
+                extension.unwrap_or(".parquet".to_string()),
+            )),
+            Some(format) => Err(Error::UnsupportedFileFormat {
+                format: format.to_string(),
+            }
+            .into()),
+        }
+    }
+}
```
y-f-u: this should be same as S3 one?

gloomweaver: Created new trait ListingTableConnector with auto implementation of DataConnector

crates/runtime/src/dataconnector.rs
```diff
@@ -278,3 +305,114 @@ pub async fn get_data(
...
 }
+
+pub trait ListingTableConnector: DataConnector {
+    fn as_any(&self) -> &dyn Any;
+
+    fn get_object_store_url(&self, dataset: &Dataset) -> AnyErrorResult<Url>;
+
+    fn get_params(&self) -> &HashMap<String, String>;
+
+    fn get_file_format_and_extension(&self) -> AnyErrorResult<(Arc<dyn FileFormat>, String)> {
```
y-f-u: This doesn't feel like an interface required to expose on this trait level.

gloomweaver: it's used in the read_provider, so seems reasonable to include it in the trait

crates/runtime/src/dataconnector.rs
```diff
@@ -278,3 +305,114 @@ pub async fn get_data(
 
     Ok((table_provider.schema(), batches))
 }
+
+pub trait ListingTableConnector: DataConnector {
```
y-f-u, line 309: Can't we just have a `pub Struct ListingTableConnector` rather than trait that has a method `get_object_store_url` which based on the object_store_type to decide how to construct the url?

Given the most abstraction is done in object_store level, I don't think we need to have another level of abstraction using the trait here.

y-f-u: unless there are some really big advantage this trait can provide.

gloomweaver: I think it's more readable this way and more inline with how we implement other data connectors

crates/runtime/src/dataconnector.rs
```diff
@@ -136,6 +152,15 @@ pub enum DataConnectorError {
         dataset_name: String,
         table_name: String,
     },
+
+    #[snafu(display("Unsupported file format: {format}"))]
+    UnsupportedFileFormat { format: String },
+
+    #[snafu(display("Unsupported compression type for CSV"))]
```
phillipleblanc, line 159: If this is specific to CSV, should we rename it to `UnsupportedCsvCompressionType`?

crates/runtime/src/dataconnector.rs
```diff
@@ -278,3 +305,114 @@ pub async fn get_data(
 
     Ok((table_provider.schema(), batches))
 }
+
+pub trait ListingTableConnector: DataConnector {
+    fn as_any(&self) -> &dyn Any;
+
+    fn get_object_store_url(&self, dataset: &Dataset) -> AnyErrorResult<Url>;
```
phillipleblanc, line 312:
```suggestion
    fn get_object_store_url(&self, dataset: &Dataset) -> DataConnectorResult<Url>;
```

These should return DataConnectorResult

phillipleblanc: Here and for `get_file_format_and_extension` and `get_csv_format`

crates/runtime/src/dataconnector.rs
```diff
@@ -278,3 +305,114 @@ pub async fn get_data(
...
+                    FileCompressionType::from_str(compression_type)
+                        .context(UnsupportedCompressionTypeSnafu { compression_type })?,
+                )
+                .with_delimiter(delimiter),
+        ))
+    }
+}
+
+#[async_trait]
+impl<T: ListingTableConnector + Display> DataConnector for T {
```
phillipleblanc, line 358: I really like this approach

crates/runtime/src/dataconnector.rs
gloomweaver: Changed

crates/runtime/src/dataconnector.rs
gloomweaver: `get_object_store_url` should return `AnyErrorResult` cause there we return errors which are defined for specific DataConnector
`get_file_format_and_extension`, `get_csv_format` changed to `DataConnectorResult`

phillipleblanc (approved):

PR #1360, opened 2024-05-10
Incremental append based on time column

### LINKED ISSUE #1357: Enhancement: Incremental append mode based on temporal difference between acceleration and source
## Goal-State/What/Result

A batch-based acceleration appending mechanism to only fetch the new diffs based on temporal column from the source to the acceleration.

m1 - function done
m2 - doc done

## Why/Purpose

To enable a "pseudo-append" mode for time-series data and improve the efficiency of refresh data fetches to the source data provider as requested by users.

## By When

**Issue/Spec written and reviewed:** May 13 AEST 

**M1:** May 13 AEST (stretch, same for below)
**M2:** May 13 AEST 
**Done-Done:** May 13 AEST

## Done-Done

- [x] The Algorithm
- [ ] PM/Design Review
- [ ] UX Review
- [ ] Tests
- [ ] Documentation
- [ ] ~~Samples/Quickstarts/Tutorials~~
- [ ] Release Notes

## The Algorithm

- [x] Simplify.

## Spec

No UX change. When time_column is specified along with append data mode, it will use the incremental append mode to fetch the new data.
```
datasets:
  - from: databricks:my_dataset
    name: accelerated_dataset
    time_column: created_at
    acceleration:
      enabled: true
      refresh_mode: append
      refresh_check_interval: 10m
      refresh_sql: |
        SELECT * FROM accelerated_dataset WHERE city = 'Seattle'
```

## How/MAP

- Move refresh into its own mod so it can be easier to add new refresh related function.
- Allow a new refresh mode - append batch (name TBD)
- Use latest temporal value to fetch data from the source when refresh mode is `append_batch`.
- Docs

### Implementation (issues)

- Refactor refresh into its own mode from accelerated table
- Add a new refresh mode `append_batch` (TBD)
- Fetch data from source table provider using latest temporal value constraint
- Modify the docs

y-f-u:
First iteration of the incremental append

- add more tests to cover timestamp conversion
- refactoring refresh logic into refresher of accelerated tables
- add test on refresh on accelerated table
- introduce a new incremental append mechanism

Limitation:
- incremental append mechanism only and will work when append mode + time column is set. We are potential moving current append (streaming) into a changes mode.
- ~~duplication handling. This is a common problem for key-based incremental append. Some vendor (stitch) uses Greater/Equal operation to fetch new dates which prefer data duplication over incompleteness. And right now the code is doing so. Further enhancement can build more comprehensive record batch comparison to reduce the data duplication.~~ Opted for Gt as it supports nano second.
- it also does not respect refresh_data_window yet.


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -0,0 +1,683 @@
...
+        let refresh = self.refresh.read().await;
+        let filter_converter = self.get_filter_converter(&refresh);
+
+        tracing::info!("[refresh] Loading data for dataset {dataset_name}");
+        status::update_dataset(dataset_name.as_str(), status::ComponentStatus::Refreshing);
+        let refresh = refresh.clone();
+        let mut filters = vec![];
+        if let Some(converter) = filter_converter.as_ref() {
+            if let Some(timestamp) = overwrite_timestamp_in_nano {
+                filters.push(converter.convert(timestamp, Operator::GtEq));
```
phillipleblanc: Shouldn't this be Gt, not GtEq?

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -0,0 +1,683 @@
...
+        let ctx = SessionContext::new();
+
+        let mut ready_sender = Some(ready_sender);
+
+        loop {
+            let future_result = stream.next().await;
+
+            match future_result {
+                Some(data_update) => {
+                    let Ok(data_update) = data_update else {
```
Jeadie, line 109: I think the inverse is clearer (i.e. continue on `Err(e)`).

crates/runtime/src/accelerated_table/refresh.rs
y-f-u: No, as it could have new data with same timestamp.

> duplication handling. This is a common problem for key-based incremental append. Some vendor (stitch) uses Greater/Equal operation to fetch new dates which prefer data duplication over incompleteness. And right now the code is doing so. Further enhancement can build more comprehensive record batch comparison to reduce the data duplication.

Listed in the description.

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -0,0 +1,683 @@
...
+
+    async fn stream_updates(
+        &self,
+        acceleration_refresh_mode: AccelerationRefreshMode,
+    ) -> BoxStream<'_, super::Result<DataUpdate>> {
+        let time_column = self.refresh.read().await.time_column.clone();
+
+        match acceleration_refresh_mode {
+            AccelerationRefreshMode::Append(receiver) => {
+                if let (Some(receiver), Some(_)) = (receiver, time_column) {
```
phillipleblanc, line 167: This is a bit of a smell - if there isn't a receiver or time column we just ignore it? I think it's fine for now, but I'd want to see this refactored to cover all cases, or make it such that this can't happen

phillipleblanc (dismissed):

Jeadie (dismissed):

crates/runtime/src/accelerated_table/refresh.rs
phillipleblanc: Sure but there could also be new data with a lower timestamp - this doesn't guarantee us data completeness anyway

crates/runtime/src/accelerated_table/refresh.rs
y-f-u: Yeah, one reason I was leaving it so is we are going to have more change about OnExchange which is the OG append. So this code may need further changes.

crates/runtime/src/accelerated_table/refresh.rs
y-f-u: yeah, it is tricky one. We can tune it later or even expose the configure for this.

crates/runtime/src/accelerated_table/refresh.rs
y-f-u: The reason of leaving so is we will need to do more to mutate the data if it's needed (remove duplication within the new data, inbetween new and old data).

Jeadie (dismissed):

crates/runtime/src/accelerated_table.rs
```diff
@@ -1,47 +1,37 @@
-use std::time::{SystemTime, SystemTimeError, UNIX_EPOCH};
```
y-f-u, line 1: Block for merge for now. Saw an issue when doing the final testing.

phillipleblanc (dismissed):

crates/runtime/src/accelerated_table/refresh.rs
y-f-u: Opt for Gt

sgrebnov (approved):

phillipleblanc (approved):

PR #1361, opened 2024-05-10
Store runtime metrics in `spice.runtime.metrics` table

ewgenius:
Added internal metrics table.

- connects to prometheus metrics endpoint and scrapes metrics each 30 seconds
- stores in `spice.runtime.metrics` table
- scraper started if spiced is run with arguments `metrics` and `spice-cloud-metrics`

```
spiced --metrics 127.0.0.1:9000 --spice-cloud-connect
```

CleanShot 2024-05-13 at 12 50 45@2x

CleanShot 2024-05-13 at 12 52 09@2x

CleanShot 2024-05-13 at 12 51 27@2x

CleanShot 2024-05-13 at 12 53 23@2x


crates/runtime/src/spice_metrics_recorder.rs
```diff
@@ -0,0 +1,231 @@
...
+    pub fn new() -> Self {
+        let fields = vec![
+            Field::new(
+                "timestamp",
+                DataType::Timestamp(TimeUnit::Second, None),
+                false,
+            ),
+            Field::new("metric", DataType::Utf8, false),
+            Field::new("value", DataType::Float64, false),
+            Field::new("labels", DataType::Utf8, false),
```
Jeadie: I think it could be good to have labels as a DataType::Map or something where its more SQL-queryable. I don't know if its possible since i think the values in Map/Dict type must be consistent. alternatively have UTF8 as JSON.

crates/runtime/src/lib.rs
```diff
@@ -662,7 +684,29 @@ impl Runtime {
...
+
+        self.df
+            .write()
+            .await
+            .register_runtime_table("metrics".to_string(), recorder.table())
+            .context(UnableToRegisterMetricsTableSnafu)?;
+
+        recorder.start();
+
+        tracing::info!("Runtime metrics available in `spice.runtime.metrics` table");
```
```suggestion
        tracing::info!("Runtime metrics available at `spice.runtime.metrics`");
```

crates/runtime/src/lib.rs
```diff
@@ -21,13 +21,17 @@ use std::collections::HashSet;
...
+use ::datafusion::error::DataFusionError;
 use ::datafusion::sql::parser::{self, DFParser};
 use ::datafusion::sql::sqlparser::ast::{SetExpr, TableFactor};
 use ::datafusion::sql::sqlparser::dialect::PostgreSqlDialect;
 use ::datafusion::sql::sqlparser::{self, ast};
 use accelerated_table::AcceleratedTable;
 use app::App;
 use config::Config;
+#[cfg(feature = "spice-metrics")]
+use metrics::SetRecorderError;
```
phillipleblanc, line 32: Not a huge fan of all these `#[cfg(feature = "spice-metrics")]` everywhere. Can we put most of the logic in a separate module that is conditionally compiled in? The module can live in this file, but just for better separation of the cfg.

ewgenius: removed feature flag, that will be handled by spiced argument

crates/runtime/src/spice_metrics_recorder.rs
ewgenius: Arrow doesn't support non strict types, so I'll leave `labels` as string for now. I'm planning to extend it with `tags` field, which will hold some known tags like `dataset` and `engine`.

phillipleblanc (approved):

Issue #1362, opened 2024-05-10
[Spike] Approach to patch YAML file


sgrebnov:
1. Approach1: Use existing deserialization mehacnizm `serde_yaml` to patch yaml: [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d032e70077029c42fcbc1a35682f2ebd)
   - robust, can re-use existing models
   - does not preserve comments
1. Approach2: Manual String Manipulation: error prone, won't work well for more complex scenarios
1. Approach3: Mixed approach: 
   - step 1: Use serde_yaml to create patched definition for a dataset -> serialize patched definition to a string
   - step 2: Manual string manipulation (regexp) to update specific dataset only.

Decision is to proceed with Approach 1 and investigate effort to add own implementation/support for comments. Other approaches are not very robust and might lead to different issues hard to catch/troubleshoot

Issue #1354, opened 2024-05-09
Enhancement: Update Refresh SQL through an API


digadeesh:
## Goal-State/What/Result

A user can update the refresh sql on their accelerator through an API to make acceleration of data ephemeral or on demand based on their business needs.

Milestones:

- M1 Refresh SQL via API (Non-Persistent)

## Why/Purpose

Users have large datasets and can't accelerate all of the data into a local accelerator.  They want to specify the refresh sql on demand to load different data to accelerate as their request patterns change.

## By When

v0.13.0-alpha

**Issue/Spec written and reviewed:** 05/09

**M1:** 05/09
...
**Done-Done:** 05/13

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### M1 Refresh SQL via API (Non-Persistent)

An accelerator API can be exposed to modify the configuration of the refresh sql.  At this time the only configuration that can be changed is refresh sql.

PATCH /datasets/:name/acceleration

{
    refresh_sql: "Select * from Foo limit 100;"
}

 - On a successful POST, the response is 200 OK . The change does not automatically trigger data refresh.  Instead on the next refresh (defined as refresh interval), the new data would be refreshed. 
 - Should return 404 NOT_FOUND if database does not exist (`:name` parameter)
 - Should return 500 INTERNAL_SERVER_ERROR if unable to apply new acceleration parameters (runtime)

Note: the user can refresh their data immediately by calling the refresh API.

## How/MAP
- https://github.com/spiceai/spiceai/pull/1356
- [x] Document Refresh SQL update via API

### Implementation
- https://github.com/spiceai/spiceai/pull/1356
- https://github.com/spiceai/docs/pull/237

PR #1365, opened 2024-05-12
Save updated through API Refresh SQL to manifest

sgrebnov:
Part of https://github.com/spiceai/spiceai/issues/1354

PR adds logic to save updated via API acceleration params (`refresh_sql`) to the spicepod manifest.

**Limitations**:
- The `serde-yaml` library currently in use [does not support retaining comments](https://github.com/dtolnay/serde-yaml/issues/145#issuecomment-559201090), which results in the loss of all comments during YAML serialization. Other alternatives were reviewed (https://github.com/spiceai/spiceai/issues/1362) as well, none of them are robust.
 -  Current implementation presumes that the dataset is defined in the primary 'spicepod.yaml'. PR to track dataset definition actual location will be sent next (support of datasets defined via dataset ref/separate file and in dependency spicepod).

### DIFF
--- a/crates/runtime/src/datafusion.rs
+++ b/crates/runtime/src/datafusion.rs
@@ -38,6 +38,7 @@ use datafusion::sql::sqlparser;
 use datafusion::sql::sqlparser::dialect::PostgreSqlDialect;
 use secrets::Secret;
 use snafu::prelude::*;
+use spicepod::component::dataset::acceleration::Acceleration;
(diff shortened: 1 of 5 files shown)


phillipleblanc:
This will not work in Kubernetes. I don't think trying to update the spicepod itself is the right approach here. It also doesn't work if you've checked in your spicepods to Git, as it will start to diverge.

If we need to start storing local state (for these types of overrides) we should be explicit about it.

PR #1387, opened 2024-05-15
fix dataset refreshing status

y-f-u:
This fixes the dataset refreshing status.

Also added a test to prevent regression.


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -145,8 +153,16 @@ impl Refresher {
                             } else if let Some(sender) = ready_sender.take() {
                                 sender.send(()).ok();
                             };
+                            status::update_dataset(
+                                dataset_name.as_str(),
+                                status::ComponentStatus::Ready,
+                            );
```
phillipleblanc: This pattern is repeated a couple of times, maybe its worth extracting out a common function that both sends the `ready_sender` and updates the dataset.

phillipleblanc (approved):

PR #1388, opened 2024-05-15
clickhouse friendly error

y-f-u:
Mimic the postgres way to include:
- host or port error
- username or password error

### DIFF
--- a/crates/db_connection_pool/src/clickhousepool.rs
+++ b/crates/db_connection_pool/src/clickhousepool.rs
@@ -49,6 +49,13 @@ pub enum Error {
         source: clickhouse_rs::errors::Error,
     },
 
+    #[snafu(display(
+        "Authentication failed. Ensure that the username and password are correctly configured."
+    ))]
+    InvalidUsernameOrPasswordError {
+        source: clickhouse_rs::errors::Error,
+    },
+
     #[snafu(display("Cannot connect to ClickHouse on {host}:{port}. Ensure that the host and port are correctly configured, and that the host is reachable."))]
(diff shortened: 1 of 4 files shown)


crates/db_connection_pool/src/clickhousepool.rs
```diff
@@ -179,9 +203,34 @@ fn get_config_from_params(
...
+                        Err(Error::InvalidUsernameOrPasswordError {
+                            source: server_error.into(),
+                        })
+                    } else {
+                        Err(Error::ConnectionPoolRunError {
+                            source: server_error.into(),
+                        })
+                    }
+                }
+            },
```
gloomweaver, line 232:
```suggestion
            Err(e) => match e {
                clickhouse_rs::errors::Error::Server(server_error) => {
                    if server_error.code == 516 {
                        Err(Error::InvalidUsernameOrPasswordError {
                            source: server_error.into(),
                        })
                    } else {
                        Err(Error::ConnectionPoolRunError {
                            source: server_error.into(),
                        })
                    }
                }
                _ => {
                    Err(Error::ConnectionPoolRunError { source: e })
                }
            },
```

ewgenius (approved):

digadeesh (approved):

crates/db_connection_pool/src/clickhousepool.rs
y-f-u: The reason to keep so is to make sure if the downstream updated, we will get a compile error or something to tell us there is new error pops up.

PR #1393, opened 2024-05-16
use df logical plan to query latest timestamp when refreshing incrementally

y-f-u:
- use df logical plan to query latest timestamp from accelerator in refresher
- add a timestamp test for incremental appending.


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -338,6 +333,19 @@ impl Refresher {
...
 
+    #[allow(clippy::needless_pass_by_value)]
+    fn get_df(&self, ctx: SessionContext, column: &str) -> Result<DataFrame, DataFusionError> {
+        let expr = cast(
+            col(column),
+            DataType::Timestamp(arrow::datatypes::TimeUnit::Nanosecond, None),
+        )
+        .alias("a");
+        ctx.read_table(Arc::clone(&self.accelerator))?
+            .select(vec![expr.clone()])?
```
phillipleblanc:
```suggestion
            .select(vec![expr])?
```

phillipleblanc (approved):

PR #1402, opened 2024-05-16
Refactor datafusion.write_data to use table reference

ewgenius:
Changed `is_writable` and `write_data` methods to use `TableReference` for custom schemas support.

### DIFF
--- a/crates/runtime/src/datafusion.rs
+++ b/crates/runtime/src/datafusion.rs
@@ -323,7 +323,7 @@ impl DataFusion {
                     .context(UnableToGetTableSnafu)?
                     .ok_or_else(|| {
                         TableMissingSnafu {
-                            schema: "runtime".to_string(),
+                            schema: schema_name.to_string(),
                             table: table_name.to_string(),
                         }
                         .build()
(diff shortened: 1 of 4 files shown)


crates/runtime/src/datafusion.rs
```diff
@@ -276,16 +299,40 @@ impl DataFusion {
...
+
+        if let Some(schema_name) = table_reference.schema() {
+            if let Some(schema) = self.schema(schema_name) {
+                let table_provider = schema
+                    .table(table_name)
+                    .await
+                    .context(UnableToGetTableSnafu)?
+                    .ok_or_else(|| {
+                        TableMissingSnafu {
+                            schema: "runtime".to_string(),
```
```suggestion
                            schema: schema_name.to_string(),
```

crates/runtime/src/datafusion.rs
```diff
@@ -294,6 +341,23 @@ impl DataFusion {
...
+        data_update: DataUpdate,
+    ) -> Result<()> {
+        if !self.is_writable(&table_reference) {
+            TableNotWritableSnafu {
+                table_name: table_reference.to_string(),
+            }
+            .fail()?;
+        }
+
+        let table_provider = self.get_table_provider(table_reference.clone()).await?;
```
phillipleblanc: Maybe this can take a reference so we don't have to clone?

phillipleblanc (dismissed):

crates/runtime/src/datafusion.rs
ewgenius: fixed, thanks

phillipleblanc (approved):

PR #1415, opened 2024-05-17
Propagate object store error

gloomweaver:
Before:
CleanShot 2024-05-17 at 8  39 21@2x
After:
CleanShot 2024-05-17 at 9  35 29@2x


crates/runtime/src/dataconnector.rs
```diff
@@ -393,35 +392,31 @@ impl<T: ListingTableConnector + Display> DataConnector for T {
 
         let url = self
             .get_object_store_url(dataset)
```
ewgenius: `get_object_store_url` needs to be changed to return correct `DataConnectorError`. Returning wrapped errors is last resort if we can't detect it.

crates/runtime/src/dataconnector.rs
```diff
@@ -393,35 +392,31 @@ impl<T: ListingTableConnector + Display> DataConnector for T {
...
-                message: "Unable to parse URL",
             })?;
 
         let table_path =
             ListingTableUrl::parse(url)
                 .boxed()
-                .context(InvalidConfigurationSnafu {
+                .context(ListingTableInternalSnafu {
                     dataconnector: format!("{__self}"),
-                    message: "Unable to parse URL",
```
ewgenius, line 406: What is the actual error returned from `ListingTableUrl` in this case?

crates/runtime/src/dataconnector.rs
```diff
@@ -393,35 +392,31 @@ impl<T: ListingTableConnector + Display> DataConnector for T {
...
                 })?;
 
         let (file_format, extension) = self
             .get_file_format_and_extension()
             .map_err(Into::into)
-            .context(InvalidConfigurationSnafu {
+            .context(ListingTableInternalSnafu {
```
ewgenius: `get_file_format_and_extension` already returns DataConnectorError, why we need map and context here?

crates/runtime/src/dataconnector.rs
```diff
@@ -391,48 +404,35 @@ impl<T: ListingTableConnector + Display> DataConnector for T {
...
         let resolved_schema = options
             .infer_schema(&ctx.state(), &table_path)
             .await
             .boxed()
-            .context(InvalidConfigurationSnafu {
-                dataconnector: format!("{__self}"),
-                message: "Unable to infer files schema",
+            .context(UnableToConnectInternalSnafu {
+                dataconnector: format!("{self}"),
             })?;
```
phillipleblanc, line 424: Is there a way we can inspect the error returned by `infer_schema` to see if its a username/password error? Then we could return a specific `DataConnectorError::UnableToConnectInvalidUsernameOrPassword`, instead of the more generic one.

gloomweaver: We can downcast Boxed errors to specific Error and than treat it accordingly

ewgenius: > We can downcast Boxed errors to specific Error and than treat it accordingly

That should be done inside ftp connector, and mapped to `DataConnectorError`

phillipleblanc (approved):

Issue #1358, opened 2024-05-09
Enhancement: LRU cache for query results


lukekim:
## Goal-State/What/Result

In-memory LRU caching of query results

Example:

```yaml

runtime:
  results_cache:
    enabled: true
    cache_max_size: 128MiB
    eviction_policy: lru 
    item_ttl: 1s

datasets:
   ...
```

## Why/Purpose

To provide accelerated performance for federated and fallback queries (in conjunction with [behavior on zero results](https://docs.spiceai.org/data-accelerators#behavior-on-zero-results)) and to prevent many redundant queries within a small time period.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** 05/20
...
**Done-Done:** 05/23

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

- [x] Cache configuration
```
runtime:
  results_cache:
    enabled: true
    eviction_policy: lru 
    cache_max_size: 128mb
    item_ttl: 1s
```
- `enabled` - optional, true by default (if there is a non-empty results_cache section defined)
- `cache_max_size` - optional, maximum cache size. Default is 128mb
- `item_ttl` - optional, cache entry expiration time, 1 second by default. Duration format follows existing conventions in Spice
- `eviction_policy` - optional, specifies the eviction policy to be used when the cached data reaches the `cache_max_size`. The default and currently the only supported policy is 'lru'.
- [x] New telemetry events are added: 
  - results_cache_request_count
  - results_cache_hit_count
  - results_cache_size
  - results_cache_item_count
  - results_cache_max_size
- [x] New cache header is added to Spice http query response: `X-Cache: HIT/MISS` (no etag, age for simplicity). Exists only when cache is enabled.
- [x] Caching is enabled by default
- [x] #1522

Out of scope for LRU cache alpha
- Header/option to bypass Cache
- Manual cache eviction (API/CLI)
[...]

PR #1456, opened 2024-05-21
Add telemetry for in-memory caching

sgrebnov:

Adds the following metrics:
- results_cache_request_count
- results_cache_hit_count
- results_cache_size
- results_cache_item_count
- results_cache_max_size

### DIFF
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -58,7 +58,7 @@ pub struct CachedQueryResult {
 pub trait QueryResultCache {
     async fn get(&self, plan: &LogicalPlan) -> Result<Option<CachedQueryResult>>;
     async fn put(&self, plan: &LogicalPlan, result: CachedQueryResult) -> Result<()>;
-    fn size(&self) -> u64;
+    fn size_bytes(&self) -> u64;
     fn item_count(&self) -> u64;
 }
 
@@ -92,6 +92,9 @@ impl QueryResultCacheProvider {
             ttl,
             metrics_reported_last_time: AtomicU64::new(0),
         };
+
+        #[allow(clippy::cast_precision_loss)]
+        metrics::gauge!("results_cache_max_size").set(cache_max_size as f64);
 
         Ok(cache_provider)
     }
@@ -127,8 +130,6 @@ impl QueryResultCacheProvider {
             self.metrics_reported_last_time
                 .store(now_seconds, Ordering::Relaxed);
             #[allow(clippy::cast_precision_loss)]
-            metrics::gauge!("results_cache_max_size").set(self.max_size() as f64);
-            #[allow(clippy::cast_precision_loss)]
             metrics::gauge!("results_cache_size").set(self.size() as f64);
             #[allow(clippy::cast_precision_loss)]
             metrics::gauge!("results_cache_item_count").set(self.item_count() as f64);
@@ -142,7 +143,7 @@ impl QueryResultCacheProvider {
 
     #[must_use]
     pub fn size(&self) -> u64 {
-        self.cache.size()
+        self.cache.size_bytes()
     }
 
     #[must_use]
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -107,12 +107,12 @@ impl QueryResultCacheProvider {
     }
 
     #[must_use]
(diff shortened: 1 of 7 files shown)


crates/runtime/src/datafusion.rs
```diff
@@ -289,6 +290,7 @@ impl DataFusion {
                     .context(UnableToCreateMemoryStreamSnafu)?,
                 ));
             }
+            metrics::counter!("results_cache_miss").increment(1);
```
digadeesh: Do we really need a miss statistic?  We have the request count already, right?

sgrebnov: Updated

crates/runtime/src/lib.rs
```diff
@@ -838,10 +838,26 @@ impl Runtime {
...
+        // cache metrics reporting
+        let _handle = tokio::spawn(async move {
+            loop {
+                #[allow(clippy::cast_precision_loss)]
+                metrics::gauge!("results_cache_max_size").set(cache_copy.max_size() as f64);
+                #[allow(clippy::cast_precision_loss)]
+                metrics::gauge!("results_cache_size").set(cache_copy.size() as f64);
+                #[allow(clippy::cast_precision_loss)]
+                metrics::gauge!("results_cache_item_count").set(cache_copy.item_count() as f64);
+                sleep(Duration::from_secs(10)).await;
```
lukekim: This is weird. Not sure we want a hardcoded 10s loop here?

crates/runtime/src/lib.rs
```diff
@@ -838,10 +838,26 @@ impl Runtime {
 
         tracing::info!("Initialized query results cache; {cache_provider}");
 
+        let shared_cache_provider = Arc::new(cache_provider);
+        let cache_copy = Arc::clone(&shared_cache_provider);
```
lukekim: This is poor naming. It's not a copy of the cache, but an additional pointer to it.

crates/runtime/src/lib.rs
sgrebnov: Thank you - moved metrics reporting after cache write.

crates/runtime/src/lib.rs
sgrebnov: This is now resolved. Thank you

digadeesh (dismissed):

crates/cache/src/lib.rs
```diff
@@ -94,20 +96,47 @@ impl QueryResultCacheProvider {
...
 
     /// # Errors
     ///
     /// Will return `Err` if method fails to access the cache
     pub async fn put(&self, plan: &LogicalPlan, result: CachedQueryResult) -> Result<()> {
-        self.cache.put(plan, result).await
+        let res = self.cache.put(plan, result).await;
+
+        #[allow(clippy::cast_precision_loss)]
+        metrics::gauge!("results_cache_max_size").set(self.max_size() as f64);
```
sgrebnov: `results_cache_max_size` is a constant value but periodically reporting this metric is useful for charts/etc

lukekim (dismissed):

crates/cache/src/lib.rs
```diff
@@ -54,12 +58,15 @@ pub struct CachedQueryResult {
 pub trait QueryResultCache {
     async fn get(&self, plan: &LogicalPlan) -> Result<Option<CachedQueryResult>>;
     async fn put(&self, plan: &LogicalPlan, result: CachedQueryResult) -> Result<()>;
+    fn size(&self) -> u64;
```
phillipleblanc:
```suggestion
    fn size_bytes(&self) -> u64;
```

crates/cache/src/lib.rs
phillipleblanc: Even if we report it once, it will still show up in the Prometheus metrics every time - so reporting it multiple times doesn't really buy us anything. I would move this to the `QueryResultCacheProvider::new()` function.

crates/cache/src/lib.rs
sgrebnov: Updated, thank you

crates/cache/src/lib.rs
sgrebnov: Moved to `QueryResultCacheProvider::new()` - thank you!

phillipleblanc (approved):

PR #1472, opened 2024-05-22
Add X-Cache header for http queries

sgrebnov:

Add `X-Cache` header to indicate whether query result is from results cache (`hit`) or not(`miss`). Applied only when results cache is enabled.

### DIFF
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -45,21 +45,26 @@ pub enum Error {
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-pub enum ResultCacheStatus {
-    Cached,
-    NotCached,
-    CacheDisabled,
+pub enum RetrievalResult {
+    Hit,
+    Miss,
 }
 
 pub struct QueryResult {
     pub data: SendableRecordBatchStream,
-    pub cache_status: ResultCacheStatus,
+    pub cache_retrieval_result: Option<RetrievalResult>,
 }
 
 impl QueryResult {
     #[must_use]
-    pub fn new(data: SendableRecordBatchStream, cache_status: ResultCacheStatus) -> Self {
-        QueryResult { data, cache_status }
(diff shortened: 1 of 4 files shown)


digadeesh (dismissed):

digadeesh (dismissed):

crates/cache/src/lib.rs
```diff
@@ -44,6 +45,24 @@ pub enum Error {
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
+pub enum ResultCacheStatus {
+    Cached,
```
lukekim, line 49: Let's call this Hit/Miss

crates/cache/src/lib.rs
```diff
@@ -44,6 +45,24 @@ pub enum Error {
...
+    NotCached,
+    CacheDisabled,
+}
```
lukekim:
```suggestion
pub enum RetrievalResult {
    Hit,
    Miss
}
```

Don't put disabled in here.

crates/runtime/src/datafusion.rs
```diff
@@ -267,7 +269,7 @@ impl DataFusion {
         &self,
         sql: &str,
         restricted_sql_options: Option<SQLOptions>,
-    ) -> Result<SendableRecordBatchStream> {
+    ) -> Result<QueryResult> {
```
lukekim, line 270: Make this Optional. If it is missing, cache is disabled.

sgrebnov: Like the proposed approach with Option< RetrievalResult>. Updated

crates/cache/src/lib.rs
sgrebnov: Like the proposed approach with Option< RetrievalResult>. Updated

lukekim (approved):

phillipleblanc (approved):

Issue #1318, opened 2024-05-06
Enhancement: Improved federated query pushdown


phillipleblanc:
## Goal-State/What/Result

Queries involving a single table in a federated query engine pushes through the entire query to that engine, and no work is done in DataFusion to the results. Logically the query text will just be passed from DataFusion to the Query Engine directly.

More complex queries involving multiple federated tables and other sources will send as much of the computation to be done by each remote query engine as possible. "As much as possible" will be a best effort approach for now and will be improved over time as we get more concrete use-cases, but should cover the majority of the most common ones.

Specifically, that means enabling support for sorting (i.e. TopK), aggregations and joins. As an example, these queries will all be pushed down properly:
- `SELECT * FROM table ORDER BY col LIMIT 10` (i.e. TopK)
- `SELECT max(number) FROM table` (aggregation)
- `SELECT max(number), service FROM table GROUP BY service` (aggregation)
- `SELECT t.id, t.name, r.description FROM table1 t JOIN table2 r ON t.id = r.id` (joins between two tables in the federated source)

## Why/Purpose

This will be needed to enable more efficient queries that perform aggregations, sorts and joins between two tables in the same federated source.

## By When

**Issue/Spec written and reviewed:** May 8, 2024

**M1:** May 10, 2024
**M2:** May 15, 2024
**M3:** May, 17, 2024
...
**Done-Done:** <Target Date>

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

[...]

PR #1477, opened 2024-05-22
Add basic integration test for MySQL federation

phillipleblanc:
Part of #1318

Adds a basic MySQL integration test that uses Docker to start a MySQL server.

Also adds a new Docker module for starting a Docker container programmatically. Use it to start a local MySQL server for the integration test.

This has exposed an issue with MySQL support, when deparsing from a DataFusion LogicalPlan back to SQL for MySQL:

The following query:
`SELECT * FROM line ORDER BY line.l_orderkey DESC LIMIT 10`

gets deparsed to:
`SELECT "line"."l_orderkey", "line"."l_partkey", "line"."l_suppkey", "line"."l_linenumber", "line"."l_quantity", "line"."l_extendedprice", "line"."l_discount", "line"."l_tax", "line"."l_returnflag", "line"."l_linestatus", "line"."l_shipdate", "line"."l_commitdate", "line"."l_receiptdate", "line"."l_shipinstruct", "line"."l_shipmode", "line"."l_comment" FROM "lineitem" AS "line" ORDER BY "line"."l_orderkey" DESC NULLS FIRST LIMIT 10`

Which returns the following error:
`Error: "query `SELECT * FROM line ORDER BY line.l_orderkey DESC LIMIT 10` to results: Execution error: Unable to query arrow: Server error: `ERROR 42000 (1064): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'NULLS FIRST LIMIT 10' at line 1'"`


digadeesh (approved):

crates/runtime/tests/integration.rs
```diff
@@ -14,5 +14,39 @@ See the License for the specific language governing permissions and
...
+    };
+
+    let subscriber = tracing_subscriber::FmtSubscriber::builder()
+        .with_env_filter(filter)
+        .with_ansi(true)
+        .finish();
+    let _ = tracing::subscriber::set_global_default(subscriber);
+}
+
+async fn get_tpch_lineitem() -> Result<Vec<RecordBatch>, anyhow::Error> {
```
sgrebnov, line 41: @phillipleblanc  - where we use `get_tpch_lineitem`?

phillipleblanc: It's used in the MySQL integration test

sgrebnov: Oh, missed the part inside "Load diff". Now it clear how it works.

PR #1494, opened 2024-05-23
Dataset Configure CLI - Add `file_format` prompt for s3 and ftp datasets, if no extension detected

ewgenius:
When using `spice dataset configure`
- if dataset is from s3, ftp or sftp, and path don't have file extension, cli will prompt user for `file_format`

https://github.com/spiceai/spiceai/assets/827338/385caa61-2aaa-4803-8307-f5f37fff8db3

### DIFF
--- a/bin/spice/cmd/dataset.go
+++ b/bin/spice/cmd/dataset.go
@@ -116,6 +116,31 @@ spice dataset configure
 			params["endpoint"] = endpoint
 		}
 
+		if datasetPrefix == spec.DATA_SOURCE_S3 || datasetPrefix == spec.DATA_SOURCE_FTP || datasetPrefix == spec.DATA_SOURCE_SFTP {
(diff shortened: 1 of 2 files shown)


bin/spice/cmd/dataset.go
```diff
@@ -116,6 +116,31 @@ spice dataset configure
 			params["endpoint"] = endpoint
 		}
 
+		if datasetPrefix == spec.DATA_SOURCE_S3 || datasetPrefix == spec.DATA_SOURCE_FTP || datasetPrefix == spec.DATA_SOURCE_SFTP {
```
lukekim, line 119: My guess is we don't need this check, but we'll see.

ewgenius: Do you mean we don't need to check for dataset source here? File format applicable only for ftp and s3 connectors

lukekim (approved):

PR #1498, opened 2024-05-24
Cache invalidation for accelerated tables

sgrebnov:

Remove cached results for accelerated dataset after data is updated (refresh and retention logic).

### DIFF
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -115,11 +115,11 @@ pub async fn run(args: Args) -> Result<()> {
 
     rt.load_secrets().await;
 
+    rt.init_results_cache().await;
+
     rt.load_datasets().await;
 
     rt.load_models().await;
-
-    rt.init_results_cache().await;
 
     if args.spice_cloud_connect {
         if let Err(err) = rt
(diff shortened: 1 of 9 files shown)


crates/runtime/src/http/v1.rs
```diff
@@ -166,12 +166,12 @@ pub(crate) mod query {
 
         match is_data_from_cache {
             Some(true) => {
-                if let Ok(value) = "HIT".parse() {
+                if let Ok(value) = "Hit from spiceai".parse() {
```
sgrebnov, line 169: This adds more clarity and matches format of industry leading systems, for example  Amazon CloudFront

```
< HTTP/1.1 200 OK< Content-Type: text/html
< Content-Length: 437
...
< Server: AmazonS3
< Age: 14
< X-Cache: Hit from cloudfront
```

bin/spiced/src/lib.rs
```diff
@@ -115,12 +115,12 @@ pub async fn run(args: Args) -> Result<()> {
 
     rt.load_secrets().await;
 
+    rt.init_results_cache().await;
```
sgrebnov, line 118: Cache provider should be set before datasets initialization => moved up.

phillipleblanc (dismissed):

lukekim (approved):

PR #1501, opened 2024-05-24
Store history of queries in 'spice.runtime.query_history'

Jeadie:
# Changes
- New internal table 'spice.runtime.query_history' with schema

```
sql> describe spice.runtime.query_history
+-------------------+-----------------------------+-------------+
| column_name       | data_type                   | is_nullable |
+-------------------+-----------------------------+-------------+
| query_id          | Utf8                        | NO          |
| schema            | Utf8                        | NO          |
| sql               | Utf8                        | NO          |
| nsql_query        | Utf8                        | YES         |
| start_time        | Timestamp(Nanosecond, None) | NO          |
| end_time          | Timestamp(Nanosecond, None) | NO          |
| execution_time    | UInt64                      | YES         |
| rows_produced     | UInt64                      | NO          |
| results_cache_hit | Boolean                     | NO          |
+-------------------+-----------------------------+-------------+
```
- Track SQL history for `v1/sql` and `v1/nsql`
- Create a `QueryHistory` object that stores `query_history` properties during request lifetime. On `drop`, write to datafusion 
- Note: Long term solution will be to write this as an extension.


bin/spiced/src/lib.rs
```diff
@@ -124,6 +124,21 @@ pub async fn run(args: Args) -> Result<()> {
         rt.load_llms().await;
     }
 
+    rt.init_results_cache().await;
```
y-f-u: is this line relevant here?

crates/runtime/src/http/v1.rs
```diff
@@ -992,6 +1025,8 @@ pub(crate) mod nsql {
         Extension(nsql_models): Extension<Arc<RwLock<LLMModelStore>>>,
         Json(payload): Json<Request>,
     ) -> Response {
+        let mut q_trace = QueryHistory::default().results_cache_hit(false); // TODO use query cache
```
y-f-u: This might need to be in an issue with a link here if it's planned to be fixed later.

crates/runtime/src/http/v1.rs
```diff
@@ -1040,9 +1077,20 @@ pub(crate) mod nsql {
...
             Ok(Some(model_sql_query)) => {
                 let cleaned_query = clean_model_based_sql(&model_sql_query);
+                q_trace = q_trace.sql(cleaned_query.clone());
                 tracing::trace!("Running query:\n{cleaned_query}");
-                match readable_df.ctx.sql(&cleaned_query).await {
-                    Ok(result) => dataframe_to_response(result).await,
+
+                q_trace = q_trace.start_time(SystemTime::now());
+                let result = readable_df.ctx.sql(&cleaned_query).await;
+                q_trace = q_trace.end_time(SystemTime::now());
```
y-f-u: it this supposed to trace the dataframe with sql time? or intended for the dataframe_tro_response time.

crates/runtime/src/query_history.rs
```diff
@@ -0,0 +1,355 @@
...
+            retention,
+        )
+        .await
+        .boxed()
+        .context(UnableToRegisterTableSnafu),
+    }
+}
+
+#[must_use]
+pub fn table_schema() -> Schema {
```
y-f-u: is this required to be public?

bin/spiced/src/lib.rs
ewgenius: this is duplicate from line 118, must be from conflict resolving

ewgenius:
```suggestion
```

bin/spiced/src/lib.rs
```diff
@@ -124,6 +124,21 @@ pub async fn run(args: Args) -> Result<()> {
...
+            rt.df
+                .write()
+                .await
+                .register_runtime_table(query_history::DEFAULT_QUERY_HISTORY_TABLE, table)
+                .context(UnableToCreateBackendSnafu)?;
+        }
+        Err(err) => {
+            tracing::warn!("Creating internal query history table: {err}");
+        }
+    };
```
ewgenius, line 128: I think that can be moved as a function inside Runtime

crates/runtime/src/query_history.rs
```diff
@@ -0,0 +1,355 @@
...
+                Arc::new(UInt64Array::from(vec![self.execution_time])),
+                Arc::new(UInt64Array::from(vec![self.rows_produced])),
+                Arc::new(BooleanArray::from(vec![self.results_cache_hit])),
+            ],
+        )
+        .boxed()
+        .context(UnableToCreateRowSnafu)
+    }
+
+    pub fn validate(&self) -> Result<(), Error> {
```
y-f-u, line 256: this should return all errors rather than the error in order?

crates/runtime/src/query_history.rs
```diff
@@ -0,0 +1,355 @@
...
+    accelerated_table::{refresh::Refresh, AcceleratedTable, Retention},
+    datafusion::DataFusion,
+    dataupdate::DataUpdate,
+    internal_table::{create_internal_accelerated_table, create_synced_internal_accelerated_table},
+};
+
+pub const DEFAULT_QUERY_HISTORY_TABLE: &str = "query_history";
+
+pub async fn instantiate_query_history_table(
+    secrets_provider: &Arc<RwLock<secrets::SecretsProvider>>,
```
y-f-u: doesn't feel like this needs the entire secrets_provider passing here

crates/runtime/src/lib.rs
```diff
@@ -172,6 +173,9 @@ pub enum Error {
     #[snafu(display("Unable to start local metrics: {source}"))]
     UnableToStartLocalMetrics { source: spice_metrics::Error },
 
+    #[snafu(display("Unable to track query history: {source}"))]
+    UnableToStartQueryHistory { source: query_history::Error },
```
```suggestion
    UnableToTrackQueryHistory { source: query_history::Error },
```

crates/runtime/src/http/v1.rs
Jeadie: Not really, moreso just a note to myself now the cache is ready.

crates/runtime/src/http/v1.rs
Jeadie: ahh it should track both 1. `readable_df.ctx.sql` and 2.  `dataframe_to_response`. Essentially to give as close as an analogy to the standard `POST v1/sql` endpoint.

crates/runtime/src/query_history.rs
```diff
@@ -0,0 +1,313 @@
...
+    .context(UnableToRegisterTableSnafu)
+}
+
+#[must_use]
+fn table_schema() -> Schema {
+    Schema::new(vec![
+        Field::new("query_id", DataType::Utf8, false),
+        Field::new("schema", DataType::Utf8, false),
+        Field::new("sql", DataType::Utf8, false),
+        Field::new("nsql_query", DataType::Utf8, true),
```
lukekim:
```suggestion
        Field::new("nsql", DataType::Utf8, true),
```

crates/runtime/src/query_history.rs
```diff
@@ -0,0 +1,313 @@
...
+    #[must_use]
+    pub fn sql(mut self, sql: String) -> Self {
+        self.sql = Some(sql);
+        self
+    }
+
+    #[must_use]
+    pub fn nsql_query(mut self, nsql_query: String) -> Self {
+        self.nsql_query = Some(nsql_query);
+        self
```
lukekim:
```suggestion
    pub fn nsql_query(mut self, nsql: String) -> Self {
        self.nsql_query = Some(nsql);
        self
```

crates/runtime/src/query_history.rs
```diff
@@ -0,0 +1,313 @@
...
+pub const DEFAULT_QUERY_HISTORY_TABLE: &str = "query_history";
+
+pub async fn instantiate_query_history_table() -> Result<Arc<AcceleratedTable>, Error> {
+    let time_column = Some("start_time".to_string());
+    let time_format = Some(TimeFormat::UnixSeconds);
+
+    let retention = Retention::new(
+        time_column.clone(),
+        time_format,
+        Some(Duration::from_secs(3 * 24 * 60 * 60)), // 3 days
```
lukekim: Let's make this 1 day by default.
```suggestion
        Some(Duration::from_secs(1 * 24 * 60 * 60)), // 1 day
```

lukekim (approved):

Issue #1159, opened 2024-04-22
Enhancement: MotherDuck Data Connector


digadeesh:
## Goal-State/What/Result

MotherDuck can be used as a data connector with the Spice OSS

Milestones:
- M1 MotherDuck Data Connector Alpha

## Why/Purpose

MotherDuck is a cloud based, serverless compute option integrated with DuckDB.  Spice runtime users can seamlessly leverage their MotherDuck compute services in Spice.

## By When

**Issue/Spec written and reviewed:** 05/14

**M1:** TBD
**Done-Done:** TBD

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [ ] UX Review
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Documentation
- [ ] Samples/Quickstarts/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### M1 MotherDuck Data Connector Alpha
- [x] MotherDuck Data Connector can be defined in Spicepod.

```
- from: motherduck:database.schema.table
  name: my_table
  params: 
    motherduck_token: <token>
```
- [ ] Pushdowns are correctly supported for MD federated queries

- [ ] MotherDuck token based login support is added to Spice CLI:

```
spice login motherduck <token>
```

Token from secrets is used in this case.
```
- from: motherduck:database.schema.table
  name: my_table
```

- [ ] ~Support for MotherDuck OAuth flow to retrieve access token: `spice login motherduck`~

- [x] `link-dead-code` compile option is added to default build in order to get motherduck extension working. This will increase target build size from 153mb to 249mb (+63%) - the decision is that this is ok and we would like to proceed with this approach until we have proper support for motherduck by duckdb-rs.  We compile `slim` Docker version w/o this flag and don't support Motherduck Data Connector in slim version

## How/MAP
- [x] Implement MotherDuck Data Connector
- [ ] Implement `spice login motherduck <token>` support (token-based auth)
- [ ] ~Implement `spice login motherduck` (OAuth flow)~
- [ ] Quickstart to query data from MotherDuck' pre-defined `sample_data.nyc.taxi` dataset
- [ ] MotherDuck Data Connector documentation

[...]

PR #1502, opened 2024-05-24
Add MotherDuck Data Connector

sgrebnov:
Part of https://github.com/spiceai/spiceai/issues/1159

Add Data Connector for MotherDuck
- `motherduck` extension requires `link-dead-code` build flag to work. The flag is not correctly populated using `build.rs` and also can't be conditional enabled/disabled for `motherduck` feature only in `.cargo/config.toml`(config does not support feature-based flags) so always applied
- slim docker container image overrides this flag manually using `-C link-dead-code=n`

```
- from: motherduck:sample_data.nyc.taxi
  name: taxi
  params:
    motherduck_token: <token>
```


.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[build]
```
sgrebnov, line 3: Moved `config.toml` to repo root so that flags are applied when you do `cargo build` from repo root or from VSCode. Still applies as well when you use `cargo` in `bin/spiced`

crates/runtime/src/dataconnector/motherduck.rs
```diff
@@ -0,0 +1,106 @@
...
+    secret: &Option<Secret>,
+    param_name: &str,
+) -> Option<String> {
+    return get_secret_or_param(
+        params.as_ref(),
+        secret,
+        &format!("{param_name}_key"),
+        param_name,
+    );
+}
```
phillipleblanc:
```suggestion
fn get_param(
    params: &HashMap<String, String>,
    secret: &Option<Secret>,
    param_name: &str,
) -> Option<String> {
    get_secret_or_param(params, secret, &format!("{param_name}_key"), param_name)
}
```

sgrebnov: Updated - thank you

digadeesh (approved):

PR #1503, opened 2024-05-24
Separate runtime Dataset from spicepod Dataset

phillipleblanc:
This is a refactoring/removing tech debt PR that:
1. Separates the concerns of the spicepod crate and the runtime crate with respect to the `Dataset` struct. Previously we had the spicepod crate handle both the serialization/deserialization - but it also encoded the runtime logic/validation. The spicepod crate should only be concerned with how to parse a spicepod from an external representation. The semantics of interpreting the Spicepod should live in the runtime crate.
2. Changes the parameters for a dataset from `Option<HashMap<String, String>>` to just a `HashMap<String, String>`. It was unnecessary and made the code more complicated than it needed to be.
3. Removes unnecessary feature flags from the spicepod crate

This is a pre-req PR for changing the `runtime::component::dataset::Dataset.name` type from `String` to `TableReference`. At which point we can support dataset names with schemas (i.e. `eth.recent_blocks`) which will lazily register the schema into the default catalog. The validation of parsing the name will be encoded into the new Runtime Dataset struct.

### DIFF
--- a/crates/runtime/src/dataconnector/databricks.rs
+++ b/crates/runtime/src/dataconnector/databricks.rs
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+use crate::component::dataset::Dataset;
 use async_trait::async_trait;
 use data_components::databricks_delta::DatabricksDelta;
 use data_components::databricks_spark::DatabricksSparkConnect;
@@ -22,7 +23,6 @@ use datafusion::datasource::TableProvider;
 use datafusion::sql::TableReference;
 use secrets::Secret;
 use snafu::prelude::*;
-use spicepod::component::dataset::Dataset;
 use std::any::Any;
 use std::pin::Pin;
 use std::sync::Arc;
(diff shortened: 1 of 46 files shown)


crates/runtime/src/lib.rs
```diff
@@ -224,13 +232,38 @@ impl Runtime {
         }
     }
 
+    fn datasets_iter(app: &App) -> impl Iterator<Item = Result<Dataset>> + '_ {
+        app.datasets.iter().cloned().map(Dataset::try_from)
+    }
+
+    /// Returns a list of valid datasets from the given App, skipping any that fail to parse and logging an error for them.
+    fn get_valid_datasets(app: &App, log_failures: bool) -> Vec<Dataset> {
```
phillipleblanc, line 240: This is a new function that tries to parse the datasets from the spicepod into the new representation. If it fails for any reason, it skips it and returns only the valid ones. It also logs an error depending on who called it.

crates/runtime/src/dataconnector/databricks.rs
```diff
@@ -111,7 +104,10 @@ impl Databricks {
                     }
                 };
             }
-            let Some(cluster_id) = ref_params.and_then(|p| p.get("databricks-cluster-id")) else {
+            let ((Some(cluster_id), _) | (_, Some(cluster_id))) = (
+                params.get("databricks_cluster_id"),
+                params.get("databricks-cluster-id"),
```
phillipleblanc, line 109: I also changed this to accept either `databricks-cluster-id` or `databricks_cluster_id`. We can update the docs after the next release to use the underscore method.

ewgenius (approved):

PR #1514, opened 2024-05-26
Add unit tests for results caching utils

sgrebnov:
Add tests for cache utility functionality that retrieves table names from logical plan.

### DIFF
--- a/crates/cache/src/utils.rs
+++ b/crates/cache/src/utils.rs
@@ -93,32 +93,27 @@ fn collect_table_names(plan: &LogicalPlan, table_names: &mut HashSet<String>) {
 mod tests {
     use super::*;
     use arrow::datatypes::{DataType, Field, Schema};
-    use datafusion::common::{plan_err, Result};
-    use datafusion::config::ConfigOptions;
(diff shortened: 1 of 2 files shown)


crates/cache/src/utils.rs
```diff
@@ -88,3 +88,191 @@ fn collect_table_names(plan: &LogicalPlan, table_names: &mut HashSet<String>) {
...
+    #[tokio::test]
+    async fn test_collect_table_names_system_query_describe() {
+        let sql = "describe customer";
+        let logical_plan = parse_sql_to_logical_plan(sql).await;
+
+        let table_names = get_logical_plan_input_tables(&logical_plan);
+
+        let expected: HashSet<String> = vec![].into_iter().collect();
+        assert_eq!(table_names, expected);
+    }
```
y-f-u, line 122: nit. Most of the tests are repetitive pattern, you can use a for-loop to put all tests into one to have a table-based test style.

y-f-u (approved):

crates/cache/src/utils.rs
sgrebnov: Thank you Yong!

PR #1528, opened 2024-05-27
Don't cache system queries

sgrebnov:
PR disables cache for system queries (no input tables or there is `information_schema.tables` table).
Includes integration test


crates/runtime/src/datafusion.rs
```diff
@@ -274,22 +277,28 @@ impl DataFusion {
...
-                    MemoryStream::try_new(
-                        cached_result.records.to_vec(),
-                        cached_result.schema,
-                        None,
-                    )
-                    .context(UnableToCreateMemoryStreamSnafu)?,
-                );
+        let query_input_tables = get_logical_plan_input_tables(&plan);
+        let cache_enabled_for_query = !(query_input_tables.is_empty()
+            || query_input_tables.contains("information_schema.tables"));
```
lukekim, line 282: This string compare is likely to be expensive in the critical path. Isn't there any better way to detect this?

y-f-u (approved):

crates/runtime/src/datafusion.rs
sgrebnov: Thank you, I'll update this.

Issue #1389, opened 2024-05-15
Store query history in `spice.runtime.query_history` table


Jeadie:
Create a new `spice.runtime.query_history` prefixed table that holds the history of jobs ran by the runtime. Columns

- query_id
- schema
- sql
- text (nullable, present for NSQL queries)
- start_time
- end_time
- execution_time
- rows_produced
- results_cache_hit

See #1361 for comparison.

Jeadie:
Need to think more about schema

Jeadie:
Or in `INFORMATION_SCHEMA`

phillipleblanc:
@Jeadie to close this as its done.

PR #1549, opened 2024-05-29
Query History support for Flight API

sgrebnov:
Part of https://github.com/spiceai/spiceai/issues/1389
- Original QueryHistory model is now `Query` so we can extend it further (add `cache_allowed`, etc). It also contains functionality to perform a query.
- `query_history` table creation and record writing is still in `QueryHistory`
- `attach_query_context_to_stream` wrapper for streaming response is added to collect num returned records and save query details when stream is finished
- Common `sql_to_http_response` functionality for both `sql` and `nql` http handlers is added

Remaining:
- Correctly populate and save query context in case of execution error (`query_with_cache` and `attach_query_context_to_stream`)
- Review if data insertion can be accelerated w/o fully construction plan/etc. For example pre-creating insertion plan once and using it to populate new data (or inserting data to accelerated table directly).

Overall performance increased from `2_584.9414` to `7_405.9370` RPS for HTTP

Original
image

Updated
image


crates/runtime/src/datafusion.rs
```diff
@@ -297,7 +300,13 @@ impl DataFusion {
                         .context(UnableToCreateMemoryStreamSnafu)?,
                     );
 
-                    return Ok(QueryResult::new(record_batch_stream, Some(true)));
+                    return Ok(QueryResult::new(
+                        attach_query_context_to_stream(
```
sgrebnov: We can potentially finish context here as we know number of records and schema and don't wrap cached response with context to finish it once stream is completed but benchmarking does not show any performance gain so keeping unified logic here.

crates/runtime/src/query_context.rs
```diff
@@ -0,0 +1,160 @@
...
+
+pub struct QueryContext {
+    pub query_id: uuid::Uuid,
+    pub schema: Option<Arc<Schema>>,
+    pub sql: Option<String>,
+    pub nsql: Option<String>,
+    pub start_time: SystemTime,
+    pub end_time: Option<SystemTime>,
+    pub execution_time: Option<u64>,
+    pub rows_produced: Option<u64>,
```
lukekim: Instead of using `Option<u64|bool>` for all these, I would purely use integers/enums. E.g.

`results_cache_hit`

- `-1` not set
- `0` miss
- `1` hit

`rows_produced`

- `-1` not set
- etc.

sgrebnov: Like it 👍 propose to add this change as a separate iteration/PR to reduce num of changes introduced. I'll improve this as next step including query `error` indication (don't currently have this).

phillipleblanc (approved):

PR #1559, opened 2024-05-30
put load_datasets as the latest step along with start_servers

### LINKED ISSUE #1525: Spice should load datasets concurrently instead of sequentially
**Describe the bug**
Currently when Spice starts up, it loads datasets sequentially. One of the steps to loading a dataset is that is queries the remote source to check the connection and fetch the schema. This can be slow, and if there are many datasets it can cause the startup to take a while.

**To Reproduce**
Use this spicepod, run spice and observe the slow startup

```yaml
version: v1beta1
kind: Spicepod
name: spiceai

datasets:
- from: eth.recent_blocks
  name: blocks
- from: eth.recent_transactions
  name: tx
- from: eth.recent_logs
  name: eth.logs
- from: eth.recent_traces
  name: eth.recent_traces
- from: eth.beacon.slots
  name: eth_beacon.slots
```

**Expected behavior**
Each dataset is loaded concurrently, improving the startup time.

y-f-u:
Load datasets when starting the other components. As dataset loading is slower than the other components (http, etc), the log will appear nicer.

In future we can put a probe onto other components status then start loading the datasets if this becomes a strong dependency.

```
2024-05-30T05:06:44.426718Z  INFO runtime: Initialized results cache; max size: 128.00 MiB, item ttl: 1s
2024-05-30T05:06:44.432705Z  INFO runtime::flight: Spice Runtime Flight listening on 127.0.0.1:50051
2024-05-30T05:06:44.433478Z  INFO runtime::opentelemetry: Spice Runtime OpenTelemetry listening on 127.0.0.1:50052
2024-05-30T05:06:44.436567Z  INFO runtime::http: Spice Runtime HTTP listening on 127.0.0.1:3000
2024-05-30T05:06:46.842116Z  INFO runtime: Registered dataset blocks
2024-05-30T05:06:46.957034Z  INFO runtime: Registered dataset tx
2024-05-30T05:06:47.499614Z  INFO runtime: Registered dataset eth.logs
2024-05-30T05:06:47.746849Z  INFO runtime: Registered dataset eth_beacon.slots
2024-05-30T05:06:47.786891Z  INFO runtime: Registered dataset eth.recent_traces
2024-05-30T05:06:49.089736Z  INFO runtime: Registered dataset taxi_trips_1
```


bin/spiced/src/lib.rs
```diff
@@ -134,9 +132,10 @@ pub async fn run(args: Args) -> Result<()> {
         tracing::warn!("{err}");
     }
 
-    rt.start_servers(args.runtime, args.metrics)
-        .await
-        .context(UnableToStartServersSnafu)?;
+    let result = futures::join!(
+        rt.load_datasets(),
+        rt.start_servers(args.runtime, args.metrics)
```
lukekim: We should start_servers and then load datasets.

```suggestion
        rt.start_servers(args.runtime, args.metrics),
        rt.load_datasets()
```

bin/spiced/src/lib.rs
```diff
@@ -134,9 +132,10 @@ pub async fn run(args: Args) -> Result<()> {
...
+    );
 
-    Ok(())
+    result.1.context(UnableToStartServersSnafu)
```
```suggestion
    result.0.context(UnableToStartServersSnafu)
```

phillipleblanc (dismissed):

Jeadie (dismissed):

crates/runtime/src/datafusion.rs
```diff
@@ -540,11 +546,19 @@ impl DataFusion {
...
         accelerated_table_builder.zero_results_action(acceleration_settings.on_zero_results);
 
-        accelerated_table_builder.cache_provider(self.cache_provider.clone());
+        accelerated_table_builder.cache_provider(self.cache_provider());
 
         Ok(accelerated_table_builder.build().await)
     }
 
+    pub fn cache_provider(&self) -> Option<Arc<QueryResultsCacheProvider>> {
+        let Ok(provider) = self.cache_provider.read() else {
```
sgrebnov, line 556: @y-f-u  - can this impact performance as we are introducing new lock that we need to acquire during each query

y-f-u: yeah, good point. it will have some impact as it is still atomic.

y-f-u: I did a quick test on the read() performance. (machine spec may differ this) it takes about 60ms for 1 million operation.

y-f-u: also this is not going to worsen the performance comparing to the previous release, as we removed the df.read() which is essentially the same. Also this won't cause write starvation as the write is only used for initialising which happened the once.

The benefit of doing this way is to init/mutate the cache provider after other components start, that can be reviewed later if this particular read() becomes performance matter.

phillipleblanc (approved):

PR #1562, opened 2024-05-30
`object_store` table provider for UTF8 data formats

Jeadie:
## Changes
- New TableProvider `ObjectStoreTextTable` that supports text data from any `object_store` compatible source. Schema is constant
```rust
    Schema::new(vec![
        Field::new("location", DataType::Utf8, false),
        Field::new("content", DataType::Utf8, false),
    ])
```
- For any `object_store` compatible source, if it isn't a known tabular format (e.g. `json`, `parquet`), it will consider it as text data. NOTE: for this unstructured data, a `file_format` must be set. 

- Share common code from existing `ObjectStoreMetadataTable` and `ObjectStoreTextTable` into new common object `ObjectStoreContext`.

### Other changes
 - `crates/data_components/src/metadata/` becomes `crates/data_components/src/object/`.


crates/runtime/src/dataconnector.rs
```diff
@@ -338,66 +340,60 @@ pub trait ListingTableConnector: DataConnector {
...
+        dataset: &Dataset,
+    ) -> DataConnectorResult<Arc<dyn TableProvider>>
+    where
+        Self: Display,
+    {
+        let store_url: Url = self.get_object_store_url(dataset)?;
+        let store = self.get_object_store(dataset)?;
+        let (_, extension) = self.get_file_format_and_extension(dataset)?;
+
+        let table = ObjectStoreMetadataTable::try_new(store, &store_url, Some(extension)).context(
```
phillipleblanc: Let's also add a comment (similar to the others) explaining why this can't error.

crates/data_components/src/object/text.rs
```diff
@@ -0,0 +1,263 @@
...
+        filters: &[&Expr],
+    ) -> DataFusionResult<Vec<TableProviderFilterPushDown>> {
+        Ok(vec![
+            TableProviderFilterPushDown::Unsupported;
+            filters.len()
+        ])
+    }
+}
+
+pub struct ObjectStoreRawExec {
```
phillipleblanc:
```suggestion
pub struct ObjectStoreTextExec {
```

crates/runtime/src/dataconnector.rs
Jeadie: This can error?

phillipleblanc (approved):

Issue #1283, opened 2024-05-02
Enhancement: Improved telemetry & dashboard


phillipleblanc:
## Goal-State/What/Result

Spice is instrumented to provide the following metrics:

- Query execution time (p50,p75,p90,p95,p99)
- Refresh fails (by dataset)
- Offline datasets (connection/auth errors - either during or after startup)
- Failing queries (with errors)
- Accelerated table last updated (by dataset)
- p50,p75,p90,p95,p99 for all duration metrics.

And this is visualized on the Grafana dashboard.

## Why/Purpose

Unblock production usage of Spice

## By When

**Issue/Spec written and reviewed:** May 31, 2024

**Done-Done:** Thu, June 6, 2024

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] ~Tests~
- [x] ~Documentation~
- [x] ~Samples/Quickstarts/Tutorials~
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

Alignment should be to OTEL spec where possible: https://github.com/open-telemetry/opentelemetry-specification

### - [x] `Query duration`
  **name**: `query_duration_seconds`
  **value**: The total duration that the query took (including queuing, execution, streaming results - until the last chunk of data was returned), measured in floating seconds (i.e. `1.15`).
  **percentile**: p50, p75, p90, p95, p99
  **dimensions**:
    `tags`: [accelerated, cache-hit] -> Indicates if this query involved a dataset that was accelerated.
    `datasets`: [blocks, logs] -> Comma separated list of datasets that were involved in the query
    `protocol: http | flight | etc` -> Which API was used to make the request (HTTP or Flight)

Visualization:
 - `SQL Query Engine` panel with chart showing queries duration aggregated by protocol (`http`, `flight`, etc) and for selected `percentile`
 - Percentile format is `p90`, etc
 
![image](https://github.com/spiceai/spiceai/assets/981580/82fca8f6-0cd1-4d00-a4fa-824ef6e769ab)

### - [x] `Failing queries`
  **name**: `query_failures`
  **value**: Count of query failures
  **dimensions**:
[...]

PR #1575, opened 2024-06-02
Add `query_duration_seconds` and `query_failures` metrics

sgrebnov:

1. Added the following metrics
 - `query_duration_seconds`
 - `query_failures`
   - Don't currently include `accelerated` tag as part of tags dimension (will be added separately)
2. Updated query execution logic to handle failures and correctly report metrics / update query history.
3. Added the following fields to`runtime.query_history` (inspired by [snowflake query_history](https://docs.snowflake.com/en/sql-reference/account-usage/query_history))
- `execution_status` - 0 means success, -1 failure. Int8 selected to enable support for more granular statuses - `cancelled`, etc in the future
- `error_message`
4. `execution_time` now uses Float32


crates/runtime/src/datafusion/query.rs
```diff
@@ -64,112 +83,173 @@ pub struct Query {
...
-            if let Some(execution_time) = self
-                .end_time
-                .and_then(|end_time| end_time.duration_since(self.start_time).ok())
-                .and_then(|duration| u64::try_from(duration.as_millis()).ok())
-            {
-                self.execution_time = Some(execution_time);
+            self.execution_time = Some(duration.as_secs_f32());
+        }
+
+        self.protocol.to_string();
```
phillipleblanc:
```suggestion
```

crates/runtime/src/datafusion/query/builder.rs
```diff
@@ -40,6 +42,7 @@ impl QueryBuilder {
             query_id: Uuid::new_v4(),
             nsql: None,
             restricted_sql_options: None,
+            protocol: Protocol::Http,
```
phillipleblanc: We shouldn't assume this is HTTP - this should be a required parameter.

phillipleblanc (dismissed):

crates/runtime/src/datafusion/query/builder.rs
sgrebnov: Updated, thank you!

crates/runtime/src/datafusion/query.rs
sgrebnov: Thank you

phillipleblanc (approved):

PR #1577, opened 2024-06-02
Adds benchmarking tests for Spice

### LINKED ISSUE #1379: Performance testing benchmarks
Part of #1318 

Collect some performance benchmarks for the before/after of federation.

phillipleblanc:
Closes #1379

Adds benchmarking tests around the TPCH datasets. Unfortunately all of the "official" benchmark queries fail currently, so I added some simpler ones to kickstart the benchmarking.

Uploads the results from running the benchmark suite to the dataset `spiceai.tests.oss_benchmarks` when run from CI.


.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,28 @@
+---
+    name: benchmark tests
+    
+    on:
+      workflow_dispatch:
+    
+    jobs:
+      build:
+        name: Benchmark Tests
+        runs-on: ubuntu-latest-16-cores
```
lukekim: Do we need to run on larges?

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,28 @@
+---
+    name: benchmark tests
```
lukekim, line 2: Should we consolidate or reconcile with the tcp-h workflow?

lukekim (dismissed):

.github/workflows/benchmarks.yml
phillipleblanc: The tpch workflow should be removed now. I'll remove it

.github/workflows/benchmarks.yml
phillipleblanc: My thinking is that the shared runners will have more variance due to other workloads potentially running at the same time. The larger runners might be more consistent, but I'm not actually sure.

crates/runtime/benches/bench.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    let mut min_iter_duration_ms = i64::MAX;
+    let mut max_iter_duration_ms = i64::MIN;
+
+    for _ in 0..benchmark_results.iterations() {
+        let start_iter_time = get_current_unix_ms();
+        let _ = rt
+            .df
+            .ctx
+            .sql(query)
```
sgrebnov, line 78: Should we use QueryBuilder w/ disabled cache to detect potential performance regression while we are adding more features to query?  The difference between two is that we have extra logic in QueryBuilder to collect table names, check schema, have extra stream wrappers to collect num records, save query history, detect whether query is accelerated or not, etc. Or even have two runs - direct query and via QueryBuilder.

sgrebnov (approved):

crates/runtime/benches/bench.rs
lukekim: We should have both cached and non-cached runs.

phillipleblanc: I'll leave it as is for now.

PR #1585, opened 2024-06-03
initial idea for 'POST v1/assist'

Jeadie:
## Changes
- New HTTP API `POST /v1/assist` with payload
```json
{
  "text":  "What gets used to find the relevant data",
  "from": ["public.my_data", "another_table_name"]
}
```

#### Other Changes
- Add support for `array_distance(FixedSizeList(float32, *), List(Float64))`. This supports the brute force vector search against a constant `Vec<f32>`.
```sql
ORDER BY array_distance(col_embedding, [1.0, 2.0, 3.0,])
LIMIT {n}
```


Jeadie:
Depends on https://github.com/spiceai/spiceai/pull/1601

crates/runtime/src/http/v1.rs
```diff
@@ -1178,3 +1178,271 @@ pub(crate) mod embed {
...
+        embeddings: Arc<RwLock<EmbeddingModelStore>>,
+    ) -> Result<HashMap<String, Vec<f32>>, Box<dyn std::error::Error>> {
+        let mut embedded_inputs: HashMap<String, Vec<f32>> = HashMap::new();
+        for (name, model) in embeddings
+            .read()
+            .await
+            .iter()
+            .filter(|(model_name, _)| embeddings_to_run.contains(model_name))
+        {
+            match model
```
y-f-u, line 1231: a large part of the body should be a method in EmbeddingModelStore

exposing read(), write() here feels like exposing too much thing in this layer

phillipleblanc (approved):

crates/runtime/src/http/v1.rs
Jeadie: See https://github.com/spiceai/spiceai/issues/1617

PR #1589, opened 2024-06-03
Add accelerated datasets refresh metrics

sgrebnov:

Adds the following accelerated dataset refresh metrics:

-  **name**: `datasets_acceleration_refresh_errors`
  **value**: Count of times we tried to load/refresh an accelerated dataset and failed 
  **dimensions**:
    `dataset`: str -> The dataset that failed to load
- **name**: `datasets_acceleration_last_refresh_time` (TIMESTAMP)
  **value**: A gauge with the timestamp in Unix seconds since the last time a successful refresh of an accelerated dataset occurred.
  **dimensions**:
    `dataset`: str -> The dataset in question
    `refresh_sql`: str -> The refresh SQL used (if any configured)


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -130,8 +131,13 @@ impl Refresher {
 
             match future_result {
                 Some(result) => {
-                    let Ok((start_time, data_update)) = result else {
-                        continue;
+                    let (start_time, data_update) = match result {
+                        Ok((start_time, data_update)) => (start_time, data_update),
+                        Err(e) => {
+                            tracing::debug!("Error getting update for dataset {dataset_name}: {e}");
```
sgrebnov, line 136: Writing as debug instead of error to not duplicate the error (there is already error from connection stream)

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -113,6 +113,7 @@ impl Refresher {
         self
     }
 
+    #[allow(clippy::too_many_lines)]
```
y-f-u: is that possible to split this methods into multiple smaller ones?

I think we should try to avoid these outlier if no obvious reason.

sgrebnov: Updated: https://github.com/spiceai/spiceai/pull/1589/commits/64b8d3a6b434218e47cccc1b1fcc7024b4617aeb

digadeesh (approved):

PR #1598, opened 2024-06-03
Add query duration, failures and accelerated dataset metrics to Grafana dashboard

sgrebnov:
Adds `Query Duration`, `Query Failures` charts

image

Adds `Time Since Last Refresh` and `Refresh Failures` charts for Accelerated Dataset

image

Remaining steps:
1. Quantile-> Percentile
2. Unavailable Federated Datasets chart


monitoring/grafana-dashboard.json
```diff
@@ -438,6 +438,316 @@
...
+          },
+          "editorMode": "code",
+          "expr": "time()-datasets_acceleration_last_refresh_time{pod=~\"${pods}\",dataset!~\"runtime.*\"}",
+          "instant": false,
+          "legendFormat": "{{dataset}}",
+          "range": true,
+          "refId": "A"
+        }
+      ],
+      "title": "Last Refresh",
```
lukekim:
```suggestion
      "title": "Refresh Duration",
```

(unless this is the time since last refresh in which case the title should be that)

lukekim: I'd make this a column (vertical bar) chart.

monitoring/grafana-dashboard.json
```diff
@@ -438,6 +438,316 @@
         "x": 0,
         "y": 16
       },
+      "id": 26,
+      "panels": [],
+      "title": "Queries",
```
lukekim: Better title and add proper descriptions.

monitoring/grafana-dashboard.json
sgrebnov: Confirm this is the time since last refresh

monitoring/grafana-dashboard.json
sgrebnov: Updated to `SQL Queries`

monitoring/grafana-dashboard.json
sgrebnov: To confirm - we don't want to change this to vertical bar if the `Last Refresh` is the time since last refresh / not refresh duration. Could you please confirm.

digadeesh (approved):

PR #1601, opened 2024-06-03
Scalar UDF `array_distance` as euclidean distance between Float32[]

Jeadie:
## Changes
 - New `ScalarUDFImpl for ArrayDistance` for Float32[] data type.
 - Should allow us the following within Datafusion

 ```sql
SELECT * 
FROM tbl
order by array_distance(col_embedding, [1.0, 2.0, 3.0])
limit 10
```

### CODE REVIEW

File: crates/runtime/src/embeddings/array_distance.rs [hunk 1]
```diff
@@ -0,0 +1,210 @@
...
+        let args = ColumnarValue::values_to_arrays(args)?;
+
+        let v1: Vec<Float32Array> = as_fixed_size_list_array(&args[0])?
+            .iter()
+            .map(|v| {
+                v.ok_or_else(|| DataFusionError::Internal("no null entries allowed".into()))
+                    .and_then(|vv| {
+                        Arc::clone(&vv)
+                            .as_any()
+                            .downcast_ref::<Float32Array>()
```
Jeadie: This `.downcast_ref::<Float32Array>()` is failing for reasons I don't understand.

crates/runtime/src/embeddings/array_distance.rs
```diff
@@ -0,0 +1,189 @@
...
+            }
+        }
+    }
+
+    // Basic implementation of the Euclidean distance between two arrays
+    fn invoke(&self, args: &[ColumnarValue]) -> DataFusionResult<ColumnarValue> {
+        let args = ColumnarValue::values_to_arrays(args)?;
+
+        let v1: Vec<Float32Array> = Self::to_float32_array(&args[0])?;
+        let v2: Vec<Float32Array> = Self::to_float32_array(&args[1])?;
```
y-f-u, line 116: if you casting the array to use arrow computing to some type that contains other type range, this UDF can work more broadly. Just in case you need it in the future.

Jeadie: Ultimately yes, but currently embeddings also only support Float32. Allowing more numeric types will be a feature in both.

y-f-u (approved):
reckon some of the function can be redone using arrow computing native ops.

Discussed with @Jeadie , this can be left for future improvement.

PR #1619, opened 2024-06-04
openai server trait and move HTTP endpoints to `crates/runtime/src/http/v1/`

Jeadie:
## Changes
 - Create a trait `openai::Server` that allows models to be openAI compatible.
 - Move endpoints from `crates/runtime/src/http/v1.rs` to `crates/runtime/src/http/v1/*.rs`

#### Other changes
 - Create a new HashMap in `Runtime` that contains parsed OpenAI compatible models from `llms` spicepod. This raises the issue that having OpenAI compatible servers is incompatible with our separation of `llms` and `embeddings` in the spicepod.yaml
 - Create `LocalCreateEmbeddingRequest` since the `async_openai::types::CreateEmbeddingRequest` doesn't support deserialisation. #1623 tracks fixing this issue.


crates/llms/src/openai/mod.rs
```diff
@@ -27,8 +31,11 @@ use async_openai::{
...
 use futures::future::try_join_all;
 use serde_json::Value;
+use server::Server;
 use snafu::ResultExt;
 
+pub mod server;
+
 const MAX_COMPLETION_TOKENS: u16 = 1024_u16; // Avoid accidentally using infinite tokens. Should think about this more.
 
 pub(crate) const GPT3_5_TURBO_INSTRUCT: &str = "gpt-3.5-turbo";
```
lukekim, line 42: Let's also add GPT-4o.

crates/runtime/src/http/routes.rs
```diff
@@ -67,6 +67,8 @@ pub(crate) fn routes(
             .route("/v1/predict", post(v1::inference::post))
             .route("/v1/nsql", post(v1::nsql::post))
             .route("/v1/embed", post(v1::embed::post))
+            .route("v1/oai/:name/chat/completions", post(v1::oai::chat))
```
lukekim: Let's make this purely compatible.

```suggestion
            .route("/v1/chat/completions", post(v1::oai::chat))
```

crates/runtime/src/http/routes.rs
```diff
@@ -67,6 +67,8 @@ pub(crate) fn routes(
...
+            .route("v1/oai/:name/embeddings", post(v1::oai::embeddings))
```
lukekim:
```suggestion
            .route("/v1/embeddings", post(v1::oai::embeddings))
```

crates/runtime/src/http/v1/oai.rs
```diff
@@ -0,0 +1,43 @@
+/*
```
lukekim: Let's not have a `oai.rs`. This should just be `chat.rs` and `embeddings.rs`

lukekim (dismissed):

crates/llms/src/openai/mod.rs
Jeadie: we'll be able to specify which to use in the model config.

crates/runtime/src/http/routes.rs
Jeadie: I think we want to have a `:name` parameter so that we enable future compatibilies. It'll just mean setting the endpoint to be `http://localhost:300/oai/my_llm_name/`. But you are right, it'd have to be switched around.

ewgenius (approved):

ewgenius (approved):

PR #1622, opened 2024-06-04
Add first-class support for views

phillipleblanc:
Now supports a new top-level `views` node in a spicepod that allows specifying views that should be created.

Example spicepod:

```yaml
datasets:
- from: duckdb:my_table
  name: my_dataset
  params:
    open: duck.db

views:
- name: my_view
  sql: |
    SELECT * FROM my_dataset

- name: my_view_ref
  sql_ref: my_view.sql
```


lukekim (dismissed):

ewgenius (dismissed):

crates/runtime/src/lib.rs
```diff
@@ -310,7 +314,28 @@ impl Runtime {
...
+        app.views
+            .iter()
+            .cloned()
+            .map(View::try_from)
+            .zip(&app.views)
+            .filter_map(|(view, spicepod_view)| match view {
+                Ok(view) => Some(view),
+                Err(e) => {
+                    if log_failures {
+                        //status::update_view(&spicepod_view.name, status::ComponentStatus::Error);
```
digadeesh: Let's remove this unless we plan to do something with it very near term.

digadeesh (dismissed):

crates/runtime/src/component/view.rs
```diff
@@ -0,0 +1,95 @@
...
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct View {
+    pub name: TableReference,
+    /// Inline SQL that describes a view.
+    sql: Option<String>,
+    /// Reference to a SQL file that describes a view.
+    sql_ref: Option<String>,
```
lukekim: To make it clear you cannot have both at the same time, you could have a single String and a flag/enum field for the type of String.

crates/runtime/src/lib.rs
```diff
@@ -141,6 +142,9 @@ pub enum Error {
     #[snafu(display("Unable to create view: {reason}"))]
     UnableToCreateView { reason: String },
 
+    #[snafu(display("Need to specify the view SQL string for {name} inline with `sql: SELECT * FROM...` or as a file reference with `sql_ref: my_view.sql`"))]
```
lukekim:
```suggestion
    #[snafu(display("Specify the SQL string for view {name} using either `sql: SELECT * FROM...` inline or as a file reference with `sql_ref: my_view.sql`"))]
```

crates/runtime/src/lib.rs
```diff
@@ -365,33 +398,39 @@ impl Runtime {
...
-        let secrets_provider = shared_secrets_provider.read().await;
-
-        if !verify_dependent_tables(&ds, &existing_tables) {
-            status::update_dataset(&ds.name, status::ComponentStatus::Error);
-            metrics::counter!("datasets_load_error").increment(1);
-            return UnableToLoadDatasetConnectorSnafu {
-                dataset: ds.name.clone(),
+        if !verify_dependent_tables(view, &existing_tables) {
+            return UnableToCreateViewSnafu {
+                reason: "One or more of the tables in the view's SQL statement do not exist"
```
lukekim:
```suggestion
                reason: "One or more tables in the view's SQL statement do not exist."
```

lukekim (dismissed):

crates/runtime/src/component/view.rs
phillipleblanc: I reworked this so that we fail during the spicepod conversion if they don't specify one.

lukekim (approved):

PR #1629, opened 2024-06-05
Add `/v1/ready` API that returns 200 when all datasets have loaded

### LINKED ISSUE #1588: /ready API that only returns 200 when the datasets are initially loaded
This API tells if the Spice is ready for all datasets after they are being loaded, so it can be used for use cases like k8s readiness/startup probes.

phillipleblanc:
Closes #1588

Adds a new HTTP API `/v1/ready` that returns 200 OK when all datasets have finished initial loading, otherwise it return 523 Not Ready. This makes it suitable for use in a K8s readiness probe to ensure traffic doesn't go to this instance until it has finished loading.

### DIFF
--- a/crates/runtime/src/datafusion.rs
+++ b/crates/runtime/src/datafusion.rs
@@ -16,7 +16,7 @@ limitations under the License.
 
 use std::borrow::Borrow;
 use std::collections::HashSet;
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Mutex, RwLock};
 use std::time::Duration;
 
 use crate::accelerated_table::{refresh::Refresh, AcceleratedTable, Retention};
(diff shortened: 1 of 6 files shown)


lukekim (approved):

crates/runtime/src/datafusion.rs
```diff
@@ -183,7 +184,10 @@ pub enum Table {
 pub struct DataFusion {
     pub ctx: Arc<SessionContext>,
     data_writers: RwLock<HashSet<TableReference>>,
-    pub cache_provider: RwLock<Option<Arc<QueryResultsCacheProvider>>>,
+    cache_provider: RwLock<Option<Arc<QueryResultsCacheProvider>>>,
+
+    /// Has the initial load of the data been completed? It is the responsibility of the caller to call `mark_initial_load_complete` when the initial load is complete.
+    initial_load_complete: Mutex<bool>,
```
y-f-u, line 190: I think this is good enough to start with. But later we might need to introduce more complex state tracking.

y-f-u (approved):

PR #1639, opened 2024-06-06
Add `err_code` to `query_failures` metric

sgrebnov:

Add `err_code` dimension to `query_failures` metric:
- SyntaxError - sql is syntactically incorrect
- QueryPlanningError - error while planning of the query (missing table or column) or invalid operation
- QueryExecutionError - error when reading data from an object_store  or Data Connector failure while executing query
- InternalError - internal or other error

Telemetry `err_code`  is string instead of code (Int) so it is more user friendly and easy to use when it is displayed on the dashboard

### DIFF
--- a/crates/runtime/src/datafusion/query.rs
+++ b/crates/runtime/src/datafusion/query.rs
@@ -27,7 +27,7 @@ use datafusion::{
     execution::{context::SQLOptions, SendableRecordBatchStream},
     physical_plan::{memory::MemoryStream, stream::RecordBatchStreamAdapter},
 };
-use error_code::{error_code_from_datafusion_error, ErrorCode};
+use error_code::ErrorCode;
 use snafu::Snafu;
 use tokio::time::Instant;
 use uuid::Uuid;
@@ -116,7 +116,7 @@ impl Query {
         let plan = match session.create_logical_plan(&ctx.sql).await {
             Ok(plan) => plan,
             Err(e) => {
-                let error_code = error_code_from_datafusion_error(&e);
+                let error_code = ErrorCode::from(&e);
                 handle_error!(ctx, error_code, e, UnableToExecuteQuery)
             }
         };
@@ -167,7 +167,7 @@ impl Query {
         let df = match ctx.df.ctx.execute_logical_plan(plan).await {
             Ok(df) => df,
             Err(e) => {
-                let error_code = error_code_from_datafusion_error(&e);
+                let error_code = ErrorCode::from(&e);
                 handle_error!(ctx, error_code, e, UnableToExecuteQuery)
             }
         };
@@ -177,7 +177,7 @@ impl Query {
         let res_stream: SendableRecordBatchStream = match df.execute_stream().await {
             Ok(stream) => stream,
             Err(e) => {
-                let error_code = error_code_from_datafusion_error(&e);
+                let error_code = ErrorCode::from(&e);
                 handle_error!(ctx, error_code, e, UnableToExecuteQuery)
             }
         };
--- a/crates/runtime/src/datafusion/query.rs
+++ b/crates/runtime/src/datafusion/query.rs
@@ -216,6 +216,10 @@ impl Query {
     }
 
     pub async fn finish_with_error(mut self, error_message: String, error_code: ErrorCode) {
+        tracing::debug!(
+            "Query '{}' finished with error: {error_message}; code: {error_code}",
+            self.sql
+        );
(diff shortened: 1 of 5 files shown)


crates/runtime/src/datafusion/query/error_code.rs
```diff
@@ -0,0 +1,60 @@
...
+        match self {
+            ErrorCode::SyntaxError => write!(f, "SyntaxError"),
+            ErrorCode::QueryPlanningError => write!(f, "QueryPlanningError"),
+            ErrorCode::DataReadError => write!(f, "DataReadError"),
+            ErrorCode::UnexpectedError => write!(f, "UnexpectedError"),
+        }
+    }
+}
+
+impl From<&ErrorCode> for i8 {
```
sgrebnov, line 38: This is used to make `execution_status` result more granular in case of failure - specific code instead of generic -1

crates/runtime/src/datafusion/query/error_code.rs
```diff
@@ -0,0 +1,62 @@
...
+
+#[must_use]
+#[allow(clippy::module_name_repetitions)]
+pub fn error_code_from_datafusion_error(error: &DataFusionError) -> ErrorCode {
+    match error {
+        DataFusionError::SQL(..) => ErrorCode::SyntaxError,
+        DataFusionError::Plan(..) | DataFusionError::SchemaError(..) => {
+            ErrorCode::QueryPlanningError
+        }
+        DataFusionError::ObjectStore(..) | DataFusionError::External(..) => {
```
sgrebnov: - connectivity issue when using Spice Cloud Platform data connector or Dremio is reported as `External`
- `External` is also reported when PostgreSQL and similar connectors are unable to complete query (get data)
- s3 data connector report connectivity issue as `ObjectStore`

crates/runtime/src/datafusion/query.rs
```diff
@@ -290,20 +310,21 @@ fn attach_query_context_to_stream(
...
         while let Some(batch_result) = stream.next().await {
 
             match &batch_result {
                 Ok(batch) => {
                     num_records += batch.num_rows() as u64;
+                    yield batch_result
                 }
                 Err(e) => {
-                    ctx = ctx.error(e.to_string());
+                    ctx.finish_with_error(e.to_string(), ErrorCode::DataReadError).await;
```
sgrebnov: Found out that finalization code below is not called if there is an error so we need to finish query here.

```
ctx
            .schema(schema_copy)
            .rows_produced(num_records)
            .finish()
            .await;
```

crates/runtime/src/datafusion/query.rs
```diff
@@ -290,20 +310,24 @@ fn attach_query_context_to_stream(
...
+                    yield batch_result
                 }
                 Err(e) => {
-                    ctx = ctx.error(e.to_string());
+                    ctx
+                    .schema(schema_copy)
+                    .rows_produced(num_records)
+                    .finish_with_error(e.to_string(), ErrorCode::DataReadError).await;
+                    yield batch_result;
+                    return;
```
y-f-u, line 332: is this return intentional?

crates/runtime/src/datafusion/query/error_code.rs
```diff
@@ -0,0 +1,62 @@
...
+            ErrorCode::QueryPlanningError => -20,
+            ErrorCode::DataReadError => -30,
+            ErrorCode::UnexpectedError => -120,
+        }
+    }
+}
+
+#[must_use]
+#[allow(clippy::module_name_repetitions)]
+pub fn error_code_from_datafusion_error(error: &DataFusionError) -> ErrorCode {
```
y-f-u: is that possible to do a From trait implementation?

crates/runtime/src/datafusion/query.rs
y-f-u: does it change existing behaviour?

y-f-u (dismissed):
make sure the `return` is intentional change, otherwise looks good.

crates/runtime/src/datafusion/query.rs
sgrebnov: Confirm this is intentional.

Returning Err as record batch finalizes the stream and after while section is not executed. This is used to tell rust that we are not going to use `ctx` anymore and can safely consume it (`ctx` is also used below to finish query for happy path).

y-f-u (approved):

PR #1643, opened 2024-06-07
GraphQL pagination, limit pushdown and refactor

gloomweaver:
GraphQL pagination, limit pushdown and refactor
Limitation:
- No support for streaming, filter pushdown (need to fetch whole dataset if filter applied, regardless of the specified limit)
- Pagination works only for top level resource so this won't be paginated:
```yaml
query: |
    {
      viewer {
        starredRepositories(first: 100) {
          nodes {
            name
            watchers (first: 1) {
              nodes {
                name
              }
              pageInfo {
                endCursor
                hasNextPage
              }
            }
          }
        }
      }
    }
```
Example of the paginated query: 
```yaml
from: graphql:https://api.github.com/graphql
name: repos
params:
  auth_token: <token>
  json_path: data.viewer.starredRepositories.nodes
  query: |
    {
      viewer {
        starredRepositories(first: 100) {
          nodes {
            name
            watchers (first: 1) {
              nodes {
                name
              }

            }
          }

        }
      }
      pageInfo {
        endCursor
        hasNextPage
      }
    }
description: eth recent blocks
acceleration:
  enabled: true
  refresh_mode: full
  refresh_check_interval: 30m
```
The pagination happens sequentially, can't make any optimizations for that
Supports only cursor based pagination
Will look into streaming the results


crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -132,8 +134,41 @@ impl GraphQLClient {
...
         &self,
         schema: Option<SchemaRef>,
-    ) -> Result<(Vec<Vec<RecordBatch>>, SchemaRef)> {
-        let body = format!(r#"{{"query": {}}}"#, json!(self.query));
+        limit: Option<usize>,
+        cursor: Option<String>,
+    ) -> Result<(Vec<RecordBatch>, SchemaRef, Option<String>, usize)> {
+        let count_pattern = format!(r#"\(.*first:.*?(\d+).*\)"#,);
+        let regex_count = Regex::new(&count_pattern).unwrap();
+        let captures = regex_count.captures(&self.query);
```
ewgenius: This can be done in client initialization

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -187,9 +221,45 @@ impl GraphQLClient {
...
+        let next_cursor = match (&self.paginated_resource_name, limit_reached) {
+            (Some(resource), false) => {
+                let page_info = {
+                    // trim path to paginated resource
+                    let pattern = format!(r"^(.*?{resource})");
+                    let regex = Regex::new(pattern.as_str()).unwrap();
+                    let captures = regex.captures(&pointer);
+                    let path = captures.map_or(None, |c| {
+                        c.get(1).map(|m| m.as_str().to_owned() + "/pageInfo")
+                    });
```
ewgenius: This also can be done on client initialization, as it depends only on json_path and paginated_resource_name

crates/runtime/src/dataconnector/graphql/client.rs
```diff
@@ -0,0 +1,258 @@
...
+        request = request_with_auth(request, &self.auth);
+
+        let response = request.send().await.context(ReqwestInternalSnafu)?;
+        let status = response.status();
+        let response: serde_json::Value = response.json().await.context(ReqwestInternalSnafu)?;
+
+        handle_http_error(status, &response)?;
+        handle_graphql_query_error(&response, &query)?;
+
+        let pointer = format!("/{}", self.json_path.replace('.', "/"));
```
gloomweaver: Can be computed on init

crates/runtime/src/dataconnector/graphql/client.rs
```diff
@@ -0,0 +1,258 @@
...
+
+fn request_with_auth(request_builder: RequestBuilder, auth: &Option<Auth>) -> RequestBuilder {
+    match &auth {
+        Some(Auth::Basic(user, pass)) => request_builder.basic_auth(user, pass.clone()),
+        Some(Auth::Bearer(token)) => request_builder.bearer_auth(token),
+        _ => request_builder,
+    }
+}
+
+fn handle_http_error(status: StatusCode, response: &Value) -> Result<()> {
```
gloomweaver: Add unit tests

crates/runtime/src/dataconnector/graphql/client.rs
```diff
@@ -0,0 +1,258 @@
...
+            message: response["message"]
+                .as_str()
+                .unwrap_or("No message provided")
+                .to_string(),
+        });
+    }
+    Ok(())
+}
+
+fn handle_graphql_query_error(response: &Value, query: &str) -> Result<()> {
```
gloomweaver: Add unit tests

crates/runtime/src/dataconnector/graphql/pagination.rs
```diff
@@ -0,0 +1,147 @@
...
+                        })
+                    }
+                    _ => None,
+                }
+            }
+            None => None,
+        }
+    }
+
+    pub fn apply(
```
gloomweaver: Add unit tests

crates/runtime/src/dataconnector/graphql/pagination.rs
```diff
@@ -0,0 +1,147 @@
...
+                    )
+                    .as_str(),
+                );
+                (new_query.to_string(), limit_reached)
+            }
+            _ => (query.to_string(), false),
+        }
+    }
+
+    pub fn get_next_cursor_from_response(
```
gloomweaver: Add unit tests

crates/runtime/src/dataconnector/graphql/mod.rs
ewgenius, line 1: https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames

CleanShot 2024-06-10 at 00 21 34@2x

crates/runtime/src/dataconnector/graphql/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+            .get("json_path")
+            .ok_or("`json_path` not found in params".into())
+            .context(super::InvalidConfigurationSnafu {
+                dataconnector: "GraphQL",
+                message: "`json_path` not found in params",
+            })?
+            .to_owned();
+        let pointer = format!("/{}", json_path.replace('.', "/"));
+
+        let pagination_parameters = PaginationParameters::parse(&query, &pointer);
```
ewgenius: Why changing from `.` to `/`? Can't we just use `json_path` as is for parsing?

ewgenius: Or can we just use that format in spicepod?

gloomweaver: serde_json uses this format for access
And dot notation is more clean for a user

crates/runtime/src/dataconnector/graphql/mod.rs
gloomweaver: Yeah, I agree, but it was too cramped for a single file module
Didn't want to make significant structure changes with moving code into multiple crates/folders, so decided that this approach is good enough in scope of this pr

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -105,93 +182,66 @@ struct GraphQLClient {
...
-                    .split(" at [")
-                    .next()
-                    .unwrap_or_default()
-                    .to_string(),
-                line,
-                column,
-                query: format_query_with_context(&self.query, line, column),
-            });
-        }
+        handle_http_error(status, &response)?;
```
sgrebnov, line 225: Propose to rename to `check_for_http_error` +`check_for_graphql_query_error`.

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -219,13 +269,168 @@ impl GraphQLClient {
...
+    async fn scan(
+        &self,
+        state: &SessionState,
+        projection: Option<&Vec<usize>>,
+        filters: &[Expr],
+        limit: Option<usize>,
+    ) -> datafusion::error::Result<Arc<dyn ExecutionPlan>> {
+        let res = self
+            .client
+            .execute_paginated(Arc::clone(&self.schema), limit)
```
sgrebnov, line 406: Please double check it is safe to pass limit here w/o filter. Let's say "select users where name == 'Kirill' limit 1". Does execute return first/single person due to limit 1 (for example 'Anton') and then we apply where after that?

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -274,83 +482,269 @@ impl GraphQL {
...
+mod tests {
+    use reqwest::StatusCode;
 
-    fn schema(&self) -> Arc<Schema> {
-        Arc::clone(&self.schema)
+    use super::handle_http_error;
+    use super::PaginationParameters;
+
+    #[test]
+    fn test_pagination_parse() {
```
sgrebnov, line 530: Kudos for adding tests ❤️

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -79,20 +78,98 @@ Please verify the syntax of your GraphQL query."#
...
-        params: Arc<HashMap<String, String>>,
-    ) -> Pin<Box<dyn Future<Output = super::NewDataConnectorResult> + Send>> {
-        Box::pin(async move {
-            let graphql = Self { secret, params };
-            Ok(Arc::new(graphql) as Arc<dyn DataConnector>)
-        })
+impl PaginationParameters {
+    fn parse(query: &str, pointer: &str) -> Option<Self> {
+        let pagination_pattern = r"(?xsm)(\w*)\s*\([^)]*first:\s*(\d+).*\).*\{.*pageInfo.*\{.*hasNextPage.*endCursor.*?\}.*?\}";
+        let regex = unsafe { Regex::new(pagination_pattern).unwrap_unchecked() };
```
sgrebnov: Why we are using `unsafe` + `unwrap_unchecked`? Because `pagination_pattern` is hard-coded string and we know it works?

ewgenius: Let's use `let regex = Regex::new(pagination_pattern).expect("Invalid regex pagination pattern")` instead. It will still panic in case of an error but with more details.

On the topic of graceful/panic error handling I like [this answer](https://stackoverflow.com/a/39478185). TLDR:
- if error is recoverable, we should handle it gracefully
- if error is a bug, it is acceptable to panic

In this case 

I think we need to put some notes about `unsafe` and `unwrap` usage in our guidelines cc @phillipleblanc

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -79,20 +78,98 @@ Please verify the syntax of your GraphQL query."#
...
+    }
+
+    fn get_next_cursor_from_response(
+        &self,
+        response: &Value,
+        limit_reached: bool,
+    ) -> Option<String> {
+        if limit_reached {
+            return None;
+        }
```
ewgenius: That feels a bit weird - if limit reached I'd expect that `get_next_cursor_from_response` not called at all. Can we make it pure from the response, and check for limit in place?

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -219,13 +269,168 @@ impl GraphQLClient {
...
+        limit: Option<usize>,
+    ) -> Result<Vec<Vec<RecordBatch>>> {
+        let (first_batch, _, mut next_cursor) =
+            self.execute(Some(Arc::clone(&schema)), limit, None).await?;
+        let mut res = vec![first_batch];
+
+        while let Some(next_cursor_val) = next_cursor {
+            let (next_batch, _, new_cursor) = self
+                .execute(
+                    Some(Arc::clone(&schema)),
```
ewgenius: Let's move schema cloning outside of the loop

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -219,13 +269,168 @@ impl GraphQLClient {
...
+                    limit.map(|l| l - self.pagination_parameters.as_ref().map_or(0, |x| x.count)),
```
ewgenius: nitpick: I think it is worth moving next limit calculation in a separate var

crates/runtime/src/dataconnector/graphql.rs
ewgenius: @sgrebnov that is good catch

gloomweaver: Query: 
CleanShot 2024-06-10 at 11  34 15@2x
Requests:
CleanShot 2024-06-10 at 11  34 40@2x
Datafusion doesn't pass limit down, when filters are provided

crates/runtime/src/dataconnector/graphql.rs
gloomweaver: Don't quite understand. We can't do it with arc currently.

crates/runtime/src/dataconnector/graphql.rs
gloomweaver: Fixed

crates/runtime/src/dataconnector/graphql.rs
gloomweaver: Used panic! instead of unwrap_unchecked

crates/runtime/src/dataconnector/graphql.rs
gloomweaver: Added a test for this

crates/runtime/src/dataconnector/graphql.rs
ewgenius: Ah, you're right. It is not a reference here

ewgenius (approved):

PR #1647, opened 2024-06-07
Change default http port to 8080

digadeesh:


sgrebnov (approved):
We should update all quick starts and docs as well.

digadeesh:
Holding on this PR for the time being.  We'll introduce it after V0.14

PR #1648, opened 2024-06-07
Support accelerated embedding tables in `v1/assist`

Jeadie:
## Changes
 - `POST v1/assist` supports tables with accelerations (with embeddings).
 - Can manually configure primary keys to return with `v1/assist` text result, e.g.
    - Primary keys are also inferred from table, when possible. 
```yaml
datasets:
  - from: file://Users/jeadie/Github/data-platform/docs/decisions
    name: decisions
    params:
      file_format: md
    acceleration:
      enabled: true
    embeddings:
      - column: content
        use: oai
        column_pk:
          - id
```

### DIFF
--- a/crates/runtime/src/accelerated_table.rs
+++ b/crates/runtime/src/accelerated_table.rs
@@ -275,7 +275,7 @@ impl AcceleratedTable {
     }
 
     #[must_use] pub fn get_federated_table(&self) -> Arc<dyn TableProvider> {
-        Arc::clone(&self.accelerator)
+        Arc::clone(&self.federated)
     }
 
     pub async fn update_refresh_sql(&self, refresh_sql: Option<String>) -> Result<()> {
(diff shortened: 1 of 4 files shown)


crates/runtime/src/accelerated_table.rs
```diff
@@ -274,6 +274,10 @@ impl AcceleratedTable {
         Ok(())
     }
 
+    #[must_use] pub fn get_federated_table(&self) -> Arc<dyn TableProvider> {
```
phillipleblanc: Should this return `self.federated`?

Jeadie: Yes, sorry I didn't make a good comment. Essentially this was a draft hacky implementation (using spicepod details instead of just the `TableProvider` to ensure we had it this week). Solution needed this amongst other things.

phillipleblanc (approved):

PR #1650, opened 2024-06-08
Federated datasets availability monitoring

sgrebnov:

Implements periodic availability verification for Federated datasets (every 10 minutes per dataset) and report the `datasets_unavailable_time` metric (including dataset name and last availability time). If a dataset is detected as unavailable, verification is scheduled every 1 minute until the dataset becomes available.

Notes:
1. The `DatasetsHealthMonitor` naming is used instead of `DatasetsAvailabilityMonitor` to allow for a broader scope in the future, beyond just connectivity tests. Currently, it already ensures that parameters (tokens/auth) are still valid by performing a synthetic query.
1. The test for the federated dataset type is performed inside `DatasetsHealthMonitor` so that the current health check functionality can be extended to include accelerated dataset sources (if we decide to).

Next steps:
1. Use `query_history` to check if there have been successful requests for a table to update availability information without performing a synthetic request.
2. Exponential delay for connectivity tests for unavailable datasets.


crates/runtime/src/datasets_health_monitor.rs
```diff
@@ -0,0 +1,229 @@
...
+}
+
+pub enum AvailabilityVerificationResult {
+    Available,
+    Unavailable(SystemTime, String),
+}
+
+pub struct DatasetsHealthMonitor {
+    df_ctx: Arc<SessionContext>,
+    monitored_datasets: Arc<Mutex<HashMap<String, Arc<DatasetAvailabilityInfo>>>>,
```
gloomweaver, line 69: Isn't it better to use RwLock here?

sgrebnov: RwLock is better than there are many readers, in this case we mostly write/mutate `monitored_datasets` so Mutex is appropriate here.

digadeesh (approved):

PR #1653, opened 2024-06-10
Support stream in `v1/assist`

Jeadie:
## Changes
 - If `?stream=true` for route `POST /v1/assist/`. Returns data via server-side events (SSE), in the following order
     - 1: JSON describing the rows of data used from each source (structure is the same as the `from`  field in `v1/assist`)
     - 2 -> n: Plaintext streamed (i.e. the `text` field, but in message chunks). 

- New optional trait in `Chat`. Default method utilises existing `run` method.
```rust
async fn stream<'a>(&mut self, prompt: String) -> Result<Pin<Box<dyn Stream<Item = Result<Option<String>>> + Send>>>;
```


crates/llms/src/chat/mod.rs
```diff
@@ -127,6 +131,74 @@ pub fn message_to_content(message: &ChatCompletionRequestMessage) -> String {
...
+                        role: Some(Role::System),
+                        function_call: None,
+                    },
+                    index: i,
+                    finish_reason: None,
+                    logprobs: None,
+                };
+
+            yield Ok(CreateChatCompletionStreamResponse {
+                id: "42".to_string(),
```
sgrebnov: Please see if it is worth to document "42" usage here

sgrebnov (dismissed):

crates/llms/src/chat/mod.rs
```diff
@@ -127,6 +131,74 @@ pub fn message_to_content(message: &ChatCompletionRequestMessage) -> String {
...
+        &mut self,
+        prompt: String,
+    ) -> Result<Pin<Box<dyn Stream<Item = Result<Option<String>>> + Send>>> {
+        let resp = self.run(prompt).await;
+        Ok(Box::pin(stream! {
+            yield resp
+        }))
+    }
+
+    #[allow(deprecated)]
```
ewgenius, line 143: Can't we use `tool_call` instead of deprecated `function_call`?

ewgenius: Ah, it just have to be passed

ewgenius (dismissed):

phillipleblanc (dismissed):

digadeesh (approved):

PR #1658, opened 2024-06-10
Validate table constraints (Primary Keys/Unique Index) on accelerated tables

### LINKED ISSUE #1659: Detect when an incoming batch of data would violate the uniqueness constraints specified by the table's primary keys or unique indexes

phillipleblanc:
Closes #1659

Validate table constraints on new data arriving into the accelerated table.

This can be used to provide a good error message when we receive a batch of data that we know cannot be applied to the existing table, because the incoming data itself violates the constraints of the table.

The only safe thing to do in this case is report an error - this is the same behavior as PostgreSQL - even if the user has specified `on_conflict` behavior. The `on_conflict` behavior specifies what happens when new data conflicts with existing data in the table, not when the new data conflicts with itself.


crates/data_components/src/duckdb/write.rs
```diff
@@ -113,9 +120,7 @@ impl DataSink for DuckDBDataSink {
         data: SendableRecordBatchStream,
         _context: &Arc<TaskContext>,
     ) -> datafusion::common::Result<u64> {
-        let mut num_rows = 0;
-
-        let mut db_conn = self.duckdb.connect().await.map_err(to_datafusion_error)?;
+        let mut db_conn = self.duckdb.connect_sync().map_err(to_datafusion_error)?;
```
gloomweaver, line 123: Why using sync method is necessary here?

crates/data_components/src/util/constraints.rs
```diff
@@ -0,0 +1,183 @@
...
+    let mut futures = Vec::new();
+    for constraint in &**constraints {
+        let fut = validate_batch_with_constraint(batches.to_vec(), constraint.clone());
+        futures.push(fut);
+    }
+
+    let results = future::join_all(futures).await;
+    for result in results {
+        result?;
+    }
```
gloomweaver:
```suggestion
    let results = future::try_join_all(futures).await?;
```

crates/data_components/src/util/constraints.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    Ok(())
+}
+
+#[tracing::instrument(level = "debug", skip(batches))]
+async fn validate_batch_with_constraint(
+    batches: Vec<RecordBatch>,
+    constraint: Constraint,
+) -> Result<()> {
+    let unique_cols = match constraint {
```
gloomweaver, line 67: Probably can use if let here

crates/data_components/src/util/constraints.rs
```diff
@@ -0,0 +1,183 @@
...
+        Constraint::PrimaryKey(cols) | Constraint::Unique(cols) => cols,
+    };
+
+    let schema = batches[0].schema();
```
gloomweaver, line 71: Is it guaranteed to be not empty?
Maybe worth using `get` here

gloomweaver (dismissed):

crates/data_components/src/duckdb/write.rs
phillipleblanc: It's a setup for my next PR

crates/data_components/src/util/constraints.rs
phillipleblanc: The if let syntax is more verbose

crates/data_components/src/util/constraints.rs
phillipleblanc: Yes, its guaranteed to not be empty - that is checked at the parent function

ewgenius (approved):

PR #1678, opened 2024-06-11
Generate unique runtime instance name, and store in `runtime.metrics` table

ewgenius:
Generates unique instance name on spice runtime start, using name from spicepod (defaults to `spice`) and 8 characters uuid hash.

Added `instance` column to runtime.metrics table

CleanShot 2024-06-11 at 23 33 40@2x

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -204,7 +204,7 @@ pub type EmbeddingModelStore = HashMap<String, RwLock<Box<dyn Embed>>>;
 
 #[derive(Clone)]
 pub struct Runtime {
-    pub instance_name: String,
+    instance_name: String,
     pub app: Arc<RwLock<Option<App>>>,
     pub df: Arc<DataFusion>,
     pub models: Arc<RwLock<HashMap<String, Model>>>,
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -55,6 +55,7 @@ use tokio::sync::oneshot::error::RecvError;
 use tokio::sync::RwLock;
 use tokio::time::sleep;
 pub use util::shutdown_signal;
+use uuid::Uuid;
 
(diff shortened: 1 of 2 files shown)


sgrebnov (dismissed):

crates/runtime/benches/setup/mod.rs
```diff
@@ -18,7 +18,7 @@ pub(crate) async fn setup_benchmark(
 
     let app = build_app(upload_results_dataset);
 
-    let rt = Runtime::new(Some(app), Arc::new(vec![])).await;
+    let rt = Runtime::new("spice".to_string(), Some(app), Arc::new(vec![])).await;
```
sgrebnov: Nit: we can have separate method to specify the name and default to "spice" or auto-generated uuid. This way we don't need to update all tests/benches.


let rt = Runtime::new()
rt.with_name(...)

gloomweaver (dismissed):

crates/runtime/benches/setup/mod.rs
ewgenius: Thanks for pointing out 👍. Moved instance name generation inside runtime, we don't actually need to modify it outside.

crates/runtime/src/lib.rs
```diff
@@ -203,6 +204,7 @@ pub type EmbeddingModelStore = HashMap<String, RwLock<Box<dyn Embed>>>;
 
 #[derive(Clone)]
 pub struct Runtime {
+    pub instance_name: String,
```
phillipleblanc: This doesn't need to be public right?

phillipleblanc (dismissed):

crates/runtime/src/lib.rs
ewgenius: Yes, changed to private

phillipleblanc (approved):

PR #1682, opened 2024-06-12
Atomic updates for DuckDB tables with constraints

### LINKED ISSUE #1652: Handle DuckDB limitation on uniqueness violations in transactions
DuckDB has a limitation it calls [Over-Eager Unique Constraint Checking](https://duckdb.org/docs/sql/indexes#over-eager-unique-constraint-checking) which prevents us from being able to delete and insert in the same transaction.

phillipleblanc:
Closes #1652

DuckDB has a limitation which prevents it from being able to delete from a table and insert into that table if the table has any constraints defined. https://duckdb.org/docs/sql/indexes#over-eager-unique-constraint-checking

We work around this by creating a new table that is a copy of the existing table with no data (but same schema, indexes and constraints). We then insert into this new table, and then finally we drop the old table, drop the indexes of the copied table (since DuckDB doesn't support renaming tables that have indexes), rename the copied table to the original table name, and recreate the indexes. And it does this all within a transaction, so any error will not leave the table in an inconsistent state.

We can do this if the update is an overwrite, if its an update then we still create the new table and insert into it. But then we insert from the new table into the original table. We currently don't handle conflict violations, but that will come in a follow up PR.

This is enabled by moving the logic for creating the table into a new struct `TableCreator`. This struct is responsible for managing the lifecycle of a table, including the advanced functionality to create the copy, merge the table back in, create/drop indexes, etc.


crates/data_components/src/duckdb/write.rs
```diff
@@ -142,26 +142,41 @@ impl DataSink for DuckDBDataSink {
...
 }
 
 impl DuckDBDataSink {
-    fn new(duckdb: Arc<DuckDB>, overwrite: bool) -> Self {
+    pub(crate) fn new(duckdb: Arc<DuckDB>, overwrite: bool) -> Self {
         Self { duckdb, overwrite }
     }
 
-    fn try_write_all(
+    /// If there are constraints on the `DuckDB` table, we need to create an empty copy of the target table, write to that table copy and then depending on
```
Jeadie, line 163: Good to have https://duckdb.org/docs/sql/indexes#over-eager-unique-constraint-checking referenced here.

crates/data_components/src/duckdb/write.rs
```diff
@@ -170,11 +185,52 @@ impl DuckDBDataSink {
...
+        };
+
         if self.overwrite {
-            tracing::debug!("Deleting all data from table.");
+            insert_table_creator
+                .replace_table(tx, orig_table_creator)
+                .map_err(to_datafusion_error)?;
+        } else {
+            insert_table
+                .insert_table_into(tx, &self.duckdb)
```
Jeadie, line 211: Nice to have temporary comment mentioning issue with on conflicts.

crates/data_components/src/duckdb/creator.rs
```diff
@@ -0,0 +1,438 @@
...
+        &self,
+        transaction: &Transaction<'_>,
+        columns: Vec<&str>,
+        unique: bool,
+    ) -> super::Result<()> {
+        let mut index_builder = IndexBuilder::new(&self.table_name, columns);
+        if unique {
+            index_builder = index_builder.unique();
+        }
+        let sql = index_builder.build_postgres();
```
Jeadie, line 293: build postgres?

phillipleblanc: Sea-Query doesn't support DuckDB as a target - but the INDEX creation syntax is similar enough that it doesn't really matter which variant is used.

ewgenius (approved):

Issue #1440, opened 2024-05-21
Enhancement: Improve dataset registration output


lukekim:
## Goal-State/What/Result

Dataset registration output includes caching and acceleration information about the dataset.

## Why/Purpose

To help the user confirm the configuration of each dataset at runtime to avoid misconfiguration issues.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** 06/13
**Done-Done:** 06/14

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## Spec

1. Should include `from` spec information.
1. Should include information whether results cache is enabled
1. For accelerated datasets should include acceleration with engine informational text: `with acceleration(<engine>, <optional refresh interval>, <optional retention check interval>, <optional information about fallback on source>)`
 - If refresh is on: include refresh check interval information as `30s refresh`
 - If retention is on: include retention check interval information as `1hr retention`
 - If non-default on_empty_result fallback add "fallback on source on empty result"
 - for file mode use `acceleration (sqlite:file)` format

Examples:

```bash
Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), results cache enabled.
Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), acceleration (duckdb), results cache enabled.
Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), acceleration (sqlite:file, 30s refresh).
Dataset nation registered (spiceai:tpch.nation), acceleration (sqlite, 30s refresh, 1hr retention).
Dataset nation registered (spiceai:tpch.nation), acceleration (sqlite, fallback on source on empty result).
```

## How/MAP
 - [x] Update registration output
 - [x] Update [README.md](https://github.com/spiceai/spiceai/blob/trunk/README.md) with updated registration output

### Implementation (issues)
- https://github.com/spiceai/spiceai/pull/1692
- https://github.com/spiceai/spiceai/pull/1694

phillipleblanc:
@sgrebnov set a target date, then spec out the issue. Add examples of what we expect the logs to look like after this is done.

[...]

PR #1692, opened 2024-06-14
Improve dataset registration output

sgrebnov:
Implements https://github.com/spiceai/spiceai/issues/1440 

Improves dataset registration output. Examples:

```bash
Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), results cache enabled.
Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), acceleration (duckdb), results cache enabled.
Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), acceleration (sqlite:file, 30s refresh).
Dataset nation registered (spiceai:tpch.nation), acceleration (sqlite, 30s refresh, 1hr retention).
Dataset nation registered (spiceai:tpch.nation), acceleration (sqlite, fallback on source on empty result).
```

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -54,6 +54,7 @@ use spicepod::component::model::Model as SpicepodModel;
 use tokio::sync::oneshot::error::RecvError;
 use tokio::sync::RwLock;
 use tokio::time::sleep;
+use tracing_util::dataset_registered_trace;
 pub use util::shutdown_signal;
 use uuid::Uuid;
 
@@ -80,6 +81,7 @@ pub mod spice_metrics;
 pub mod status;
 pub mod timing;
 pub(crate) mod tracers;
+mod tracing_util;
 
 pub mod datasets_health_monitor;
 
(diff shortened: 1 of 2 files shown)


crates/runtime/src/lib.rs
```diff
@@ -80,6 +81,7 @@ pub mod spice_metrics;
 pub mod status;
 pub mod timing;
 pub(crate) mod tracers;
+mod tracing_util;
```
y-f-u, line 84: is this line needed?

crates/runtime/src/tracing_util.rs
```diff
@@ -0,0 +1,116 @@
...
+        info.push_str(":file");
+    }
+
+    if let Some(refresh_interval) = &acceleration.refresh_check_interval {
+        info.push_str(&format!(", {refresh_interval} refresh"));
+    }
+    if let Some(retention_check_interval) = &acceleration.retention_check_interval {
+        if acceleration.retention_check_enabled {
+            info.push_str(&format!(", {retention_check_interval} retention"));
+        }
```
y-f-u, line 60: this seems missing the incremental append info?

crates/runtime/src/lib.rs
sgrebnov: Confirm this is required to include `dataset_registered_trace`

y-f-u: right, I see, it's in the same package.

crates/runtime/src/tracing_util.rs
sgrebnov: Added, thank you!

y-f-u (approved):

PR #1706, opened 2024-06-17
refresh with append overlap

y-f-u:
This adds a parameters `refresh_append_overlap` duration to indicate the incremental append will try to fetch `max_timestamp_from_acceleration - refresh_append_overlap` from the source to cover any late arrival data.

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh.rs
+++ b/crates/runtime/src/accelerated_table/refresh.rs
@@ -352,7 +352,7 @@ impl Refresher {
                     "append_dataset_duration_ms",
                     vec![("dataset", dataset_name.to_string())],
                 );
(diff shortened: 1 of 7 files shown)


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -362,9 +372,18 @@ impl Refresher {
         }
     }
 
+    async fn refresh_append_overlap(&self) -> u128 {
```
phillipleblanc: It would be clearer that this returns nanoseconds if it was either added as a comment to the function, or the function name changed to `refresh_append_overlap_nanos`

phillipleblanc (approved):

PR #1707, opened 2024-06-17
Use `spiceai/async-openai` to solve `Deserialize` issue in `v1/embed`

### LINKED ISSUE #1623: Use internal dependency of `async-openai`
## Overview
-  We use [async-openai](https://crates.io/crates/async-openai) to use OpenAI as a backend, but also to make our endpoints OpenAI compatible.
- Some of the types in `async-openai` have issues doing the latter. For example, `CreateEmbeddingRequest` does not support deserialisation and there cannot be used in the server. 
- We should fork this dependency to `spiceai/async-openai` and fix the relevant tech debt from https://github.com/spiceai/spiceai/pull/1619

Jeadie:
Closes #1623

## Changes
 - Use https://github.com/spiceai/async-openai/tree/spiceai
 - Remove need for local equivalent of Embedding structs in `v1/assist`.
 - Resolve 2 change in nomenclature due to newer version of `async-openai`.


crates/secrets/src/aws_secrets_manager.rs
```diff
@@ -75,7 +75,7 @@ impl AwsSecretsManager {
     /// - If the AWS configuration cannot be loaded.
     /// - If the call to STS `get_caller_identity` fails, which might be due to invalid or expired AWS credentials.
     pub async fn init(self) -> Result<()> {
-        let config = aws_config::defaults(BehaviorVersion::v2023_11_09())
```
Jeadie, line 78: I believe this diff is a no-op

crates/llms/Cargo.toml
```diff
@@ -15,7 +15,7 @@ snafu.workspace = true
 
 dirs = "5.0.1"
 serde_json.workspace = true
-async-openai = "0.21.0"
+async-openai = { git = "https://github.com/spiceai/async-openai", rev = "48173bdee3d3be04dcc579b3211662e359b72734" }
```
phillipleblanc: Make this a workspace dependency so we don't need to change it in two places.

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #1727, opened 2024-06-19
default timeunit to be seconds when time column is a numeric column

y-f-u:
default time unit to be seconds when time column is numeric.

Also exhaust the time unit rather than using _ to prevent this happening again.

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh.rs
+++ b/crates/runtime/src/accelerated_table/refresh.rs
@@ -441,10 +441,10 @@ impl Refresher {
                 Some(TimeFormat::UnixMillis) => {
                     value *= 1_000_000;
                 }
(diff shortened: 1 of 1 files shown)


phillipleblanc (dismissed):

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -441,10 +441,10 @@ impl Refresher {
                 Some(TimeFormat::UnixMillis) => {
                     value *= 1_000_000;
                 }
-                Some(TimeFormat::UnixSeconds) => {
+                Some(TimeFormat::UnixSeconds) | None => {
```
phillipleblanc, line 444: Let's add a test for the default being seconds

phillipleblanc (approved):

PR #1733, opened 2024-06-19
Support Struct in PostgreSQL accelerator

### LINKED ISSUE #1732: Add support for Structs to PostgreSQL accelerator
Trying to accelerate a dataset into a PostgreSQL accelerator that has a Struct arrow type will cause the runtime to panic with the following message:

`thread 'main' panicked at crates/arrow_sql_gen/src/statement.rs:608:14:
not implemented: Data type mapping not implemented for Struct([Field { name: "id", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "login", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, Field { name: "name", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }])
note: run with RUST_BACKTRACE=1 environment variable to display a backtrace
exit status 101`

phillipleblanc:
Closes #1732 

This PR adds support for nested Struct types accelerated into the PostgreSQL engine.

It works by using PostgreSQL composite types: https://www.postgresql.org/docs/current/rowtypes.html

This PR also fixes a bug with the `time_column` being case-insensitive, when it should preserve case.


crates/arrow_sql_gen/src/postgres/composite.rs
```diff
@@ -0,0 +1,254 @@
...
+        source: Box<dyn std::error::Error + Sync + Send>,
+    },
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+/// A `PostgreSQL` composite type.
+/// Fields of a type can be accessed using `CompositeType::get` and `CompositeType::try_get` methods.
+///
+/// Adapted from <https://github.com/sfackler/rust-postgres/pull/565>
```
phillipleblanc, line 49: I took this entire file (with some small modifications to make lint happy) from this PR linked here.

ewgenius (approved):

crates/runtime/src/http/v1/nsql.rs
```diff
@@ -82,7 +82,7 @@ pub(crate) async fn post(
         match df.get_arrow_schema(t).await {
             Ok(schm) => {
                 let c = CreateTableBuilder::new(Arc::new(schm), format!("public.{t}").as_str());
-                table_create_stms.push(c.build_postgres());
+                table_create_stms.push(c.build_sqlite());
```
ewgenius, line 85: Is it broken with postgres now?

phillipleblanc: It requires the `postgres` feature to be enabled since that function is now feature gated - which doesn't make sense to require that the nsql code takes a dependency on that feature.

PR #1741, opened 2024-06-20
Streaming in v1/chat/completion

Jeadie:
## Changes
 - For `v1/chat/completion` use `chat_request` and `chat_stream` based on the request's `stream: bool` parameter.
 - Use `is_streaming: true` for `MistralLlama` and implement `chat_stream`.


crates/llms/src/chat/mistral.rs
```diff
@@ -25,13 +27,16 @@ use mistralrs::{
...
+use tokio::sync::{
+    mpsc::{channel, Receiver, Sender},
+    RwLock,
+};
 
 pub struct MistralLlama {
     pipeline: Arc<MistralRs>,
     tx: Sender<MistralRsponse>,
-    rx: Receiver<MistralRsponse>,
+    rx: Arc<RwLock<Receiver<MistralRsponse>>>,
```
Jeadie: Must add Arc to handle `stream!`. Must add `RwLock` because `stream!` needs mutable access.

crates/runtime/src/http/v1/chat.rs
```diff
@@ -32,10 +38,45 @@ pub(crate) async fn post(
...
     match llms.read().await.get(&model_id) {
-        Some(model) => match model.write().await.chat_request(req).await {
-            Ok(response) => Json(response).into_response(),
-            Err(_) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
-        },
+        Some(model) => {
+            if req.stream.unwrap_or_default() {
+                match model.write().await.chat_stream(req).await {
+                    Ok(strm) => create_sse_response(strm, time::Duration::from_secs(30)),
+                    Err(_) => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
```
phillipleblanc: We should at least debug log the errors here

crates/llms/src/chat/mistral.rs
phillipleblanc: Instead of keeping a single channel for all requests, let's remove the channels from the struct members and create a new channel pair on each request - which will keep it isolated for multiple requests.

Jeadie:
~~Seems to be an issue with per-request channels that is blocking any response. PR not yet ready~~
 ^ I believe issue was with model. Neither `stream=true` or `stream=false` worked. Worked on larger model.

crates/llms/src/chat/mistral.rs
```diff
@@ -224,53 +223,103 @@ impl MistralLlama {
...
+                        yield Err(ChatError::FailedToRunModel {
+                            source: err_msg,
+                        })
+                    },
+                    MistralResponse::CompletionDone(cr) => {
+                        yield Ok(Some(cr.choices[0].text.clone()));
+                        break;
+                    },
+                    MistralResponse::Done(_) => {
+                        unreachable!()
```
phillipleblanc, line 303: Why is this unreachable?

crates/llms/src/chat/mistral.rs
```diff
@@ -224,53 +223,103 @@ impl MistralLlama {
...
+                    },
+                }
+            }
+        })))
+    }
+
     async fn run(&mut self, prompt: String) -> Result<Option<String>> {
-        let r = self.to_request(prompt);
+        let (snd, mut rcv) = channel::<MistralResponse>(10_000);
+        tracing::debug!("Sending request to pipeline");
```
phillipleblanc: nit: I would keep this at a `trace` level, since its only purpose is to help us trace the flow and doesn't contain any extra information from variables that would otherwise be useful for debugging.

phillipleblanc (dismissed):

crates/llms/src/chat/mistral.rs
Jeadie: Added comment.

ewgenius (approved):

phillipleblanc (approved):

PR #1742, opened 2024-06-20
Support DuckDB function in DuckDB datasets

Jeadie:
## Changes
 - Support using DuckDB functions as datasets in spicepod. For example
```yaml
datasets:
   - from: duckdb:read_parquet('cleaned_sales_data.parquet')
     name: duc
     params:
       open: ./local.db
```
-  Create a DuckDB variations of `SQLTable` and `SQLExec` (`DuckDBTable` and `DuckSqlExec`) that will add a CTE to table scans. For example, the above becomes:
```sql
WITH read_parquet_cleaned_sales_dataparquet__view AS (
    SELECT * FROM read_parquet('cleaned_sales_data.parquet')
)
SELECT * FROM read_parquet_cleaned_sales_dataparquet__view
```


crates/data_components/src/duckdb.rs
```diff
@@ -460,3 +474,29 @@ impl ReadWrite for DuckDBTableFactory {
...
+fn create_table_function_view_name(table_reference: &TableReference) -> TableReference {
+    let tbl_ref_view = [
+        table_reference.catalog(),
+        table_reference.schema(),
+        Some(&flatten_table_function_name(table_reference)),
+    ]
+    .iter()
+    .flatten()
+    .collect_vec()
+    .iter()
```
phillipleblanc: Is the collect_vec().iter() actually required?

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #1743, opened 2024-06-20
rewrite refresh append update dedup logic using arrow comparators

y-f-u:
DataFusion is not supporting struct comparison yet as it's calling cmp package. [[Support comparison operators on nested data types (Struct, List, ..) · Issue #10856 · apache/datafusion](https://github.com/apache/datafusion/issues/10856)](https://github.com/apache/datafusion/issues/10856)

The latest arrow 52 has implemented comparators which works well to detect nested struct object.

Before this change, datafusion will failed on "not distinct from" when doing `except`. Now as it's going to use arrow native, the issue should be fixed.

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh.rs
+++ b/crates/runtime/src/accelerated_table/refresh.rs
@@ -497,13 +497,10 @@ impl Refresher {
         };
 
         let update_data = update.clone();
-        let update_data = update_data.data.first();
-
-        let Some(update_data) = update_data else {
+        let Some(update_data) = update_data.data.first() else {
             return Ok(update);
         };
 
-        let update_struct_array = StructArray::from(update_data.clone());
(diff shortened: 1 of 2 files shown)


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -495,32 +496,57 @@ impl Refresher {
...
+            predicates.push(not_matched);
+        }
+
+        filter_record_batch(update_data, &predicates.into())
+            .map(|data| DataUpdate {
+                schema: update.schema,
+                data: vec![data],
+                update_type: update.update_type,
+            })
+            .context(super::FailedToFilterUpdatesSnafu)
```
y-f-u, line 549: DO NOT MERGE YET

lukekim (approved):

Issue #1712, opened 2024-06-17
Enhancement: retry for accelerated dataset refresh


sgrebnov:
## Goal-State/What/Result

Configurable retry logic for accelerated dataset refresh 

```yaml
- from: s3://spiceai-demo-datasets/taxi_trips/2024/
  ...
  acceleration:
    refresh_retry_enabled: true,
    refresh_retry_max_attempts: 10,
    ...
```

## Why/Purpose

To allow enterprise usage, Spice must retry refresh on transient errors (temporary connectivity issues, compute warehouse goes idle, etc.). This should be configurable with the possibility to enable/disable retries.

## By When

**Issue/Spec written and reviewed:** 06/18
**Done-Done:** 06/24

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] ~Samples/Quickstarts/Tutorials~
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### Configuration

The following parameters are added to `dataset.acceleration`

```yaml
- from: s3://spiceai-demo-datasets/taxi_trips/2024/
  ...
  acceleration:
    refresh_retry_enabled: true,
    refresh_retry_max_attempts: 10,
    ...
```
- `refresh_retry_enabled`: bool  - whether retry logic is enabled. Default is true. 
- `refresh_retry_max_attempts` - maximum retry attempts to perform (if specified). default is undefined (unlimited attempts)

### Retry logic 

1. Based on Fibonacci backoff  ~exponential backoff~ starting from 1 second with a maximum 5-minute delay between retries.
1. Should detect and not retry on permanent issues related to invalid user input, for example, refresh SQL.
1. Retries continue until successfully completed, `refresh_retry_max_attempts` reached (if specified), or another refresh is triggered manually. In the latter case, retries are stopped, and the refresh starts from the beginning, including resetting the retry delay and attempt number.
1. After a completed refresh (successfully or `refresh_retry_max_attempts` reached), the next refresh is planned using refresh_check_interval (if defined).
1. The same unified retry strategy is used for both Append and Full refresh.
[...]

PR #1744, opened 2024-06-20
Split refresh into coordination and execution layers

sgrebnov:
Part of https://github.com/spiceai/spiceai/issues/1712:
Split refresh logic into coordination and execution layers:
1. The refresh execution task is now a non-blocking operation for refresh coordination logic and can be canceled or dropped by initiating a new refresh.
1. The refresh task notifies the refresh coordinator upon completion of the refresh, whether successful or not.


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -120,583 +127,115 @@ impl Refresher {
         self
     }
 
-    pub(crate) async fn start(
-        &self,
-        acceleration_refresh_mode: AccelerationRefreshMode,
+    fn start_streamed_append(
```
phillipleblanc:
```suggestion
    fn start_streaming_append(
```

phillipleblanc: Also let's put non-public functions below public ones.

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -0,0 +1,652 @@
...
+    ) -> Self {
+        Self {
+            dataset_name,
+            federated,
+            refresh,
+            accelerator,
+        }
+    }
+
+    pub async fn start_streamed_append(
```
phillipleblanc:
```suggestion
    pub async fn start_streaming_append(
```

crates/runtime/src/accelerated_table/refresh_task_runner.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use datafusion::{datasource::TableProvider, sql::TableReference};
+
+use super::refresh::Refresh;
+
+#[derive(Debug)]
+pub enum RefreshTaskResult {
+    Success,
+    Error,
+}
```
phillipleblanc: We don't need to define our own Result type, we can just use `std::result::Result` with `()` unit types for the Ok() and Err() variants

crates/runtime/src/accelerated_table/refresh_task_runner.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+}
+
+pub struct RefreshTaskRunner {
+    dataset_name: TableReference,
+    federated: Arc<dyn TableProvider>,
+    refresh: Arc<RwLock<Refresh>>,
+    accelerator: Arc<dyn TableProvider>,
+    task_thread: Option<JoinHandle<()>>,
```
phillipleblanc: We shouldn't call these threads - they aren't. Let's just use the term `task` as Tokio does. So we could rename this to `runner_task`

crates/runtime/src/accelerated_table/refresh_task_runner.rs
```diff
@@ -0,0 +1,131 @@
...
+            let mut task_completion: Option<BoxFuture<super::Result<()>>> = None;
+
+            loop {
+                if let Some(task) = task_completion.take() {
+                    select! {
+                        res = task => {
+                            match res {
+                                Ok(()) => {
+                                    tracing::debug!("Refresh task successfully completed for dataset {dataset_name}");
+                                    _ = callback_sender.send(RefreshTaskResult::Success).await;
```
phillipleblanc: We should debug log if sending the message on the channel fails.

crates/runtime/src/accelerated_table/refresh_task_runner.rs
```diff
@@ -0,0 +1,131 @@
...
+                                },
+                                Err(err) => {
+                                    tracing::debug!("Refresh task for dataset {dataset_name} failed with error: {}", err);
```
phillipleblanc:
```suggestion
                                    tracing::debug!("Refresh task for dataset {dataset_name} failed with error: {err}");
```

crates/runtime/src/accelerated_table/refresh_task_runner.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        (task_sender, callback_receiver)
+    }
+
+    pub fn stop(&mut self) {
+        if let Some(task_thread) = &self.task_thread {
+            task_thread.abort();
+            self.task_thread = None;
+        }
+    }
```
phillipleblanc: This could also be part of a drop implementation, and the parent `drop` just calls drop on the task runner.

crates/runtime/src/accelerated_table/refresh_task_runner.rs
```diff
@@ -0,0 +1,131 @@
...
+            dataset_name,
+            federated,
+            refresh,
+            accelerator,
+            task_thread: None,
+        }
+    }
+
+    pub fn start(&mut self) -> (Sender<()>, Receiver<RefreshTaskResult>) {
+        let (task_sender, mut task_receiver) = mpsc::channel::<()>(1);
```
phillipleblanc: Lets rename these to be more descriptive.

`task_sender` -> `start_refresh`
`task_receiver` -> `on_start_refresh`

crates/runtime/src/accelerated_table/refresh_task_runner.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        let (callback_sender, callback_receiver) = mpsc::channel::<RefreshTaskResult>(1);
```
phillipleblanc: Lets rename these to be more descriptive:

`callback_sender` -> `notify_refresh_complete`
`callback_receiver` -> `on_refresh_complete`

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -120,583 +127,115 @@ impl Refresher {
...
-                    }
-                    None => {
-                        tracing::warn!("Append stream ended for dataset {dataset_name}");
-                        break;
-                    },
-                }
-            }
-        }
-    }
+        let (task_sender, mut task_callback) = self.refresh_task_runner.start();
```
phillipleblanc:
```suggestion
        let (start_refresh, mut on_refresh_complete) = self.refresh_task_runner.start();
```

crates/runtime/src/accelerated_table/refresh_task_runner.rs
sgrebnov: Updated, thank you!

crates/runtime/src/accelerated_table/refresh_task_runner.rs
sgrebnov: Updated - thank you

crates/runtime/src/accelerated_table/refresh_task_runner.rs
sgrebnov: Updated - thank you

crates/runtime/src/accelerated_table/refresh_task_runner.rs
sgrebnov: Updated - thank you

crates/runtime/src/accelerated_table/refresh_task_runner.rs
sgrebnov: Updated - thank you

crates/runtime/src/accelerated_table/refresh.rs
sgrebnov: Update 👍

crates/runtime/src/accelerated_table/refresh_task_runner.rs
```diff
@@ -0,0 +1,123 @@
...
+        Self {
+            dataset_name,
+            federated,
+            refresh,
+            accelerator,
+            task: None,
+        }
+    }
+
+    pub fn start(&mut self) -> (Sender<()>, Receiver<super::Result<()>>) {
```
phillipleblanc, line 57: Let's add an `assert!()` that `self.task` is `None`

phillipleblanc (approved):

crates/runtime/src/accelerated_table/refresh_task_runner.rs
sgrebnov: Confirm I'll add this as part of next PR - thanks!

PR #1749, opened 2024-06-21
Tweak refresh scheduling logic

sgrebnov:
Tweak Refresh Scheduling Logic:
- Move the scheduling logic from AcceleratedTable to Refresh.
- Apply the next refresh interval only after the current refresh has completed.
- Reset the refresh interval upon manual refresh.

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh.rs
+++ b/crates/runtime/src/accelerated_table/refresh.rs
@@ -158,7 +158,7 @@ impl Refresher {
 
         tokio::spawn(async move {
             // first refresh is on start, thus duration is 0
-            let mut next_scheduled_refresh_timer = Some(Box::pin(sleep(Duration::from_secs(0))));
+            let mut next_scheduled_refresh_timer = Some(sleep(Duration::from_secs(0)));
 
             loop {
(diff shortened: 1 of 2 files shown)


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -152,9 +154,26 @@ impl Refresher {
...
         let cache_provider = self.cache_provider.clone();
 
+        let refresh_check_interval = self.refresh.read().await.check_interval;
+
         tokio::spawn(async move {
+            // first refresh is on start, thus duration is 0
+            let mut next_scheduled_refresh_timer = Some(sleep(Duration::from_secs(0)));
+
             loop {
+                let scheduled_refresh_fut: BoxFuture<()> = match next_scheduled_refresh_timer.take()
```
digadeesh: nit: just call it schedule_refresh_future.

digadeesh (dismissed):

phillipleblanc (dismissed):

crates/runtime/src/accelerated_table/refresh.rs
sgrebnov: Updated, thank you

digadeesh (approved):

PR #1750, opened 2024-06-21
Fix error when GraphQL error locations is missing

### LINKED ISSUE #1714: GraphQL connector panics and crashes runtime if query has missing `}`
**Describe the bug**

GraphQL connector panics and crashes runtime if query has missing `}`

**To Reproduce**

use spicepod
```yaml
- from: graphql:https://api.github.com/graphql
  name: stars
  params:
    auth_token_key: token
    json_path: data.repository.stargazers.edges
    query: |
      {
        repository(name: "spiceai", owner: "spiceai") {
          id
          name
          stargazers(first: 100) {
            edges {
              node {
                id
                name
                login
              }
            }
            pageInfo {
              hasNextPage
              endCursor
            }
          }
        }
    # }
```

Starting runtime fails with panic

CleanShot 2024-06-18 at 08 45 56@2x

**Expected behavior**

GraphQL connector handles invalid queries

phungleson:
## Issue

In the official GraphQL specification, the locations attribute is a required field in error responses. However, not all GraphQL implementations strictly adhere to this requirement. 

For example, when dealing with malformed GraphQL queries, such as the one described in  https://github.com/spiceai/spiceai/issues/1714, GitHub API returns an error response without including the locations attribute.

```json
{
  "message": "Unexpected end of document"
}
```

When locations attribute is missing, the code in `format_query_with_context` will fail, specifically this line as `0 - 1 = -1` which is out of bound for `usize`.

https://github.com/spiceai/spiceai/blob/a0b4686e9829f0d75e99742c76f515bef3441c1f/crates/runtime/src/dataconnector/graphql.rs#L353

This error will cause server to crash, so this potentially closes https://github.com/spiceai/spiceai/issues/1714.

image

## Solution

The proposed solution is to avoid formatting query when locations attribute is missing.

### DIFF
--- a/crates/runtime/src/dataconnector/graphql.rs
+++ b/crates/runtime/src/dataconnector/graphql.rs
@@ -323,27 +323,41 @@ fn handle_http_error(status: StatusCode, response: &Value) -> Result<()> {
 
 fn handle_graphql_query_error(response: &Value, query: &str) -> Result<()> {
     let graphql_error = &response["errors"][0];
+
     if !graphql_error.is_null() {
-        let line = usize::try_from(graphql_error["locations"][0]["line"].as_u64().unwrap_or(0))
-            .unwrap_or_default();
-        let column = usize::try_from(
-            graphql_error["locations"][0]["column"]
-                .as_u64()
-                .unwrap_or(0),
-        )
-        .unwrap_or_default();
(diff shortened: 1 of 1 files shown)


lukekim:
Amazing, thank you, @phungleson !

phillipleblanc (approved):
This is great, thank you!

PR #1762, opened 2024-06-24
Enable retries for accelerated table refresh

sgrebnov:
Part of https://github.com/spiceai/spiceai/issues/1712

Enable retry policy for accelerated table refresh and add integration tests.

Moved mysql docker functionality to common module so it can be re-used between federation and refresh retries integration tests.


crates/runtime/tests/refresh_retry/mysql.rs
```diff
@@ -0,0 +1,195 @@
...
+use mysql_async::{prelude::Queryable, Params, Row};
+use runtime::{
+    accelerated_table::{refresh_task::RefreshTask, AcceleratedTable},
+    Runtime,
+};
+use spicepod::component::dataset::acceleration::Acceleration;
+use tokio::time;
+use tracing::instrument;
+
+const MYSQL_DOCKER_CONTAINER: &str = "runtime-integration-test-frefresh-retry-mysql";
```
phillipleblanc:
```suggestion
const MYSQL_DOCKER_CONTAINER: &str = "runtime-integration-test-refresh-retry-mysql";
```

crates/runtime/tests/utils/mysql.rs
phillipleblanc, line 1: Let's not put this into a `utils` module - follow the postgres pattern and have a `mysql` module with a `common` submodule that this goes into.

sgrebnov: Updated so it is now `mysql` module with `common` submodule.

phillipleblanc (approved):

PR #1787, opened 2024-06-25
fix(bin/spice): Implement custom Unmarshaller for DatasetOrReference

### LINKED ISSUE #910: Bug: `spice add` and `spice dataset configure` might break non-empty spicepod.yaml
**Describe the bug**
When running commands that update spicepod.yaml, if spicepod is not empty (containing dependencies or datasets), it might break.

**To Reproduce**

1. Setup new spicepod
2. Add some non-trivial configuration with single inline dataset

CleanShot 2024-03-24 at 23 52 52@2x

3. Add dependency or configure new dataset, and see that existing configuration is invalid now

CleanShot 2024-03-24 at 23 52 59@2x

**Expected behavior**

`spice add ...` and `spice dataset configure` keeps any valid configuration in spicepod

**Additional context**

That problem is caused because we have 2 separate spicepod spec implementations: in spice runtime in rust (which is correct) and in spice cli in golang (which doesn't support inline components) ([source](https://github.com/spiceai/spiceai/blob/c0adb680c0afd0a722c4b1d53f0d916b43013142/bin/spice/pkg/spec/spicepod.go#L9-L11))

Ideally, we should replace all CLI commands that update spicepod.yaml with a Rust implementation that uses a single spicepod definition specification.

peasee:
## ❓ What type of PR is this?

- [ ] ♻️ Refactor
- [ ] ✨ Feature
- [x] 🐛 Bug Fix
- [ ] 👷 Optimization
- [ ] 📝 Documentation Update
- [ ] 🔖 Release
- [ ] 🚩 Other

## 🗣 Description

See #910 for more context.

`SpicepodSpec` statically set the type of `Datasets` to `*Reference`, when `Datasets` could be either a `Reference` or a `DatasetSpec`.

This PR resolves the issue by implementing a custom unmarshaller and marshaller with a struct to provide `DatasetOrReference` in `Datasets`.

Also updates the Go YAML package to `v3`, which is required for implementing the custom marshalling in the way I have.

## 🔨 Related Issues

* Fixes #910 

## 🤔 Concerns

* ~Are there other locations or types in the `Spicepod` spec that need similar custom unmarshalling logic?~
* ~Do we need to implement a custom unmarshaller for other script types (e.g. JSON, CSV, etc)?~
* ~Does this need tests (outside of running locally and seeing the results)?~
* ~Should the `panic`'s inside the `DatasetOrReference` implementations be replaced with something else?~

### DIFF
--- a/bin/spice/cmd/dataset.go
+++ b/bin/spice/cmd/dataset.go
@@ -200,7 +200,7 @@ spice dataset configure
 
 		var datasetReferenced bool
 		for _, dataset := range spicePod.Datasets {
-			if dataset.Reference != (spec.Reference{}) && dataset.Reference.Ref == dirPath {
+			if dataset.Reference.Ref == dirPath {
 				datasetReferenced = true
 				break
 			}
--- a/bin/spice/cmd/dataset.go
+++ b/bin/spice/cmd/dataset.go
@@ -26,9 +26,10 @@ import (
 	"time"
 
 	"github.com/logrusorgru/aurora"
+	"github.com/mitchellh/mapstructure"
 	"github.com/spf13/cobra"
 	"github.com/spiceai/spiceai/bin/spice/pkg/spec"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
 )
 
 var datasetCmd = &cobra.Command{
@@ -200,16 +201,30 @@ spice dataset configure
 
 		var datasetReferenced bool
 		for _, dataset := range spicePod.Datasets {
-			if dataset.Ref == dirPath {
+			var dataset_reference spec.Reference
+			err := mapstructure.Decode(dataset, &dataset_reference)
+			if err != nil {
+				panic(err)
+			}
+
+			if dataset_reference.Ref == dirPath {
 				datasetReferenced = true
 				break
 			}
 		}
 
 		if !datasetReferenced {
-			spicePod.Datasets = append(spicePod.Datasets, &spec.Reference{
+			dataset_reference := &spec.Reference{
 				Ref: dirPath,
-			})
+			}
+
+			var result map[string]interface{}
+			err := mapstructure.Decode(dataset_reference, &result)
+			if err != nil {
+				panic(err)
+			}
+
+			spicePod.Datasets = append(spicePod.Datasets, result)
 			spicepodBytes, err = yaml.Marshal(spicePod)
 			if err != nil {
 				cmd.Println(err)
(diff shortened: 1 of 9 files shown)


bin/spice/pkg/spec/spicepod.go
```diff
@@ -16,17 +16,73 @@ limitations under the License.
...
+	Metadata     map[string]string    `json:"metadata,omitempty" csv:"metadata" yaml:"metadata,omitempty"`
+	Datasets     []DatasetOrReference `json:"datasets,omitempty" csv:"datasets" yaml:"datasets,omitempty"`
+	Functions    []*Reference         `json:"functions,omitempty" csv:"functions" yaml:"functions,omitempty"`
+	Models       []*Reference         `json:"models,omitempty" csv:"models" yaml:"models,omitempty"`
+	Dependencies []string             `json:"dependencies,omitempty" csv:"dependencies" yaml:"dependencies,omitempty"`
+	Secrets      Secrets              `json:"secrets,omitempty" csv:"secrets" yaml:"secrets,omitempty"`
+}
+
+type DatasetOrReference struct {
+	Dataset   DatasetSpec
```
y-f-u: This might be a generic struct which handles the Models too? Not sure how the unmarshal works within Generics though.

bin/spice/cmd/dataset.go
```diff
@@ -200,16 +200,16 @@ spice dataset configure
 
 		var datasetReferenced bool
 		for _, dataset := range spicePod.Datasets {
-			if dataset.Ref == dirPath {
+			if dataset.Reference != (spec.Reference{}) && dataset.Reference.Ref == dirPath {
```
y-f-u: this line could be a method in dataset?

peasee: I had naively thought Go would complain about `Reference` being an empty value, thanks for the pointer!

bin/spice/pkg/spec/spicepod.go
peasee: I think if we largely agree this is the way forward for reading the YAML in Go, I can probably refactor to give it similar naming to Rust (e.g. calling it `Component` instead of `DatasetOrReference`)

peasee: @y-f-u I've refactored to rename to `Component` and moved it out to another file to make it easier to add more components. I've not added `Model` for now though, as I don't have enough knowledge to know which properties to use to make the unmarshal detect a `Model` vs a `Dataset`.

y-f-u (approved):
Looks good!

PR #1789, opened 2024-06-25
Support local and HF embedding models

Jeadie:
## Changes
 - Support for `embeddings` from Huggingface or local filesystem.
```yaml

embeddings:
  - name: from_hf
    from: huggingface:huggingface.co/sentence-transformers/all-MiniLM-L6-v2

  - name: from_local
    from: file://Users/jeadie/Github/spiceai/model.safetensors
    params:
      config_path: /Users/jeadie/Github/spiceai/config.json
      tokenizer_path: /Users/jeadie/Downloads/tokenizer.json

```
- Using [spiceai/text-embeddings-inference](https://github.com/spiceai/text-embeddings-inference)
    - Does not yet use any of the server, GRPC or queueing.


crates/llms/src/embeddings/candle/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+        let model_root = link_files_into_tmp_dir(files)?;
+        Self::try_new(&model_root, DType::F32)
+    }
+
+    pub fn from_hf(model_id: &str, revision: Option<&str>) -> Result<Self> {
+        let model_root = download_hf_artifacts(model_id, revision)?;
+        Self::try_new(&model_root, DType::F32)
+    }
+
+    /// Attemmpt to create a new `CandleEmbedding` instance. Requires all model artifacts to be within a single folder.
```
phillipleblanc:
```suggestion
    /// Attempt to create a new `CandleEmbedding` instance. Requires all model artifacts to be within a single folder.
```

crates/llms/src/embeddings/candle/mod.rs
```diff
@@ -0,0 +1,238 @@
...
+///
+/// ```
+///
+fn link_files_into_tmp_dir(files: HashMap<String, &Path>) -> Result<PathBuf> {
+    let temp_dir = tempdir()
+        .boxed()
+        .context(FailedToInstantiateEmbeddingModelSnafu)?;
+
+    for (name, file) in files {
+        symlink(file, &temp_dir.path().join(name))
```
phillipleblanc, line 237: This won't work on Windows

Jeadie: Using  `std::os::windows::fs::symlink_file` based on OS `cfg!` has limitations [too](https://doc.rust-lang.org/std/os/windows/fs/fn.symlink_file.html#limitations).

phillipleblanc (approved):

Issue #1785, opened 2024-06-25
Enhancement: Real-time acceleration updates via Debezium CDC 


phillipleblanc:
## Goal-State/What/Result

A new Debezium DataConnector that requires a non-Arrow acceleration engine with `refresh_mode: changes` will subscribe to changes in real-time (including the original snapshot) from a Kafka topic and apply those changes to the accelerated table to keep it updated in real-time with the original table.

## Why/Purpose

This is the holy grail of functionality we want to enable - a no compromises solution that combines both high-performance accelerated queries with highly efficient real-time delta updates.

## By When

**Issue/Spec written and reviewed:** 06/24
**Done-Done:** By `v0.15-alpha`, 06/27

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

```yaml
datasets:
- from: debezium:my_debezium_kakfa_topic
  name: cdc_table
  params:
    debezium_transport: kafka
    debezium_message_format: json # avro in the future
    # Most kafka consumers use `bootstrap.servers` as the parameter name 
    # https://docs.confluent.io/platform/current/installation/configuration/consumer-configs.html#bootstrap-servers
    kafka_bootstrap_servers: broker1:9092,broker2:9092,broker3:9092
  acceleration:
    enabled: true # acceleration is required for the debezium connector, it is an error to not specify an acceleration
    engine: duckdb # arrow isn't supported
    refresh_mode: changes # optional, default is `changes` - this is required to be set to `changes`, any other value is an error.
    mode: file # persistence is recommended to not have to rebuild the table each time Spice starts
    primary_key: id # We can discover the primary key of the table from the Debezium messages - but it can be optionally specified here to override the default.
    on_conflict:
      id: upsert # this on_conflict section is always set to `upsert` for the `primary_key` - so it is optionally to set here. If a user manually specifies something else - that is an error.
```

[...]

PR #1819, opened 2024-06-28
Initial working E2E of Debezium CDC

phillipleblanc:
## 🗣 Description

This wires up all of the pieces to have a fully working E2E experience of consuming changes and upserting them to the accelerated table.

This fully supports the `at-least-once` semantics we need to guarantee data consistency.

Other notable changes:
- Adds a `cdc` module which defines some generic traits and structs for working with change data that doesn't depend on Kafka or Debezium. This sets us up to be able to consume any changes that matches that format.
- Infers the primary keys from the Debezium Key message
- Adds a `ChangeBatch` struct that wraps a `RecordBatch` with schema validation and helper methods for extracting out the pieces needed to consume a change.
- Adds logic for turning a Debezium `ChangeEvent` message into a `RecordBatch` compatible with `ChangeBatch`
- Adds a method on the DataConnector to get a `ChangesStream` if the `refresh_mode` is `Changes`
- Logic for passing the `ChangesStream` to the `AcceleratedTable` and the logic for starting the stream and consuming the changes, and marking them as processed once we've validated no errors inserting to the table.
- Added logging when processing a change to show the primary key of the data, to make it easier to understand which change happened.

Some limitations that will be addressed in upcoming PRs:
- Right now, the spicepod needs to set the primary key and upsert `on_conflict` behavior to match what we receive from Debezium. We can auto-configure this to make it less error-prone.
- This will create an unused consumer group because the runtime calls `read_provider` twice, one to test the dataset connectivity and once to actually create the table.
- This still doesn't persist the consumer group id, so it will create a new one on startup, replaying all changes.

## 🔨 Related Issues

Part of #1785

### DIFF
--- a/crates/data_components/src/debezium/arrow/changes.rs
+++ b/crates/data_components/src/debezium/arrow/changes.rs
@@ -29,7 +29,7 @@ use arrow::{
 use snafu::prelude::*;
 
 /// Converts a `ChangeEvent` into a `ChangeBatch`
-pub fn to_record_batch(
+pub fn to_change_batch(
     table_schema: &SchemaRef,
     primary_key: &[String],
     change: &ChangeEvent,
--- a/crates/data_components/src/debezium/arrow/changes.rs
+++ b/crates/data_components/src/debezium/arrow/changes.rs
@@ -1,0 +1,100 @@
+/*
+Copyright 2024 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+use std::sync::Arc;
+
+use crate::{
+    arrow::struct_builder::StructBuilder,
+    debezium::{
+        arrow::downcast_builder,
+        change_event::{ChangeEvent, Op},
+    },
+};
+use arrow::{
+    array::{ArrayBuilder, ListBuilder, RecordBatch, StringBuilder},
+    datatypes::{DataType, Field, Schema, SchemaRef},
+};
+use snafu::prelude::*;
+
+/// Converts a `ChangeEvent` into an Arrow `RecordBatch`
+pub fn to_record_batch(
+    table_schema: &SchemaRef,
+    primary_key: &[String],
+    change: &ChangeEvent,
+) -> super::Result<RecordBatch> {
+    let schema = changes_schema(table_schema);
+
+    let mut struct_builder = StructBuilder::from_fields(schema.fields().clone(), 1);
+
+    struct_builder.append(true);
+
+    for (idx, field) in schema.fields().iter().enumerate() {
+        let field_builder = struct_builder.field_builder_array(idx);
+        match field.name().as_str() {
+            "op" => {
+                let str_builder = downcast_builder::<StringBuilder>(field_builder)?;
+                str_builder.append_value(change.payload.op.to_string());
+            }
+            "primary_key" => {
+                let list_builder =
+                    downcast_builder::<ListBuilder<Box<dyn ArrayBuilder>>>(field_builder)?;
+                if primary_key.is_empty() {
+                    list_builder.append(false);
+                } else {
+                    let str_builder = downcast_builder::<StringBuilder>(list_builder.values())?;
+                    for key in primary_key {
+                        str_builder.append_value(key);
+                    }
+                    list_builder.append(true);
+                }
+            }
+            "data" => {
+                let change_data = match change.payload.op {
+                    Op::Delete => change
+                        .payload
+                        .before
+                        .clone()
+                        .context(super::DeleteOpWithoutBeforeFieldSnafu)?,
+                    _ => change.payload.after.clone(),
+                };
+
+                let data_struct_builder = downcast_builder::<StructBuilder>(field_builder)?;
+
+                super::append_value_to_struct_builder(change_data, data_struct_builder)?;
+            }
+            _ => unreachable!("Unexpected field in changes schema"),
+        }
+    }
+
+    let struct_array = struct_builder.finish();
+    Ok(struct_array.into())
+}
+
+#[must_use]
+pub fn changes_schema(table_schema: &Schema) -> Schema {
+    Schema::new(vec![
+        Field::new("op", DataType::Utf8, false),
+        Field::new(
+            "primary_key",
+            DataType::List(Arc::new(Field::new("item", DataType::Utf8, false))),
+            true,
+        ),
+        Field::new(
+            "data",
+            DataType::Struct(table_schema.fields().clone()),
+            true,
+        ),
+    ])
+}
--- a/crates/data_components/src/debezium/arrow/changes.rs
+++ b/crates/data_components/src/debezium/arrow/changes.rs
@@ -13,10 +13,10 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-use std::sync::Arc;
 
 use crate::{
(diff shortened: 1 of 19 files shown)


digadeesh (approved):

crates/data_components/src/debezium/arrow/changes.rs
```diff
@@ -0,0 +1,91 @@
...
+    },
+};
+use arrow::{
+    array::{ArrayBuilder, ListBuilder, RecordBatch, StringBuilder},
+    datatypes::SchemaRef,
+};
+use snafu::prelude::*;
+
+/// Converts a `ChangeEvent` into a `ChangeBatch`
+pub fn to_record_batch(
```
y-f-u: to_change_batch?

crates/data_components/src/debezium/change_event.rs
```diff
@@ -21,7 +21,29 @@ use std::{
...
+    pub fn from_bytes(bytz: &[u8]) -> Result<Self, serde_json::Error> {
+        serde_json::from_slice(bytz)
+    }
+
+    #[must_use]
+    pub fn get_primary_key(&self) -> Vec<String> {
+        self.schema
+            .fields
+            .iter()
+            .filter_map(|field| field.field.clone())
```
y-f-u, line 65: this is bit confusing without knowing the data construct here, may add a comment of example here what the schema fields can be like?

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -0,0 +1,247 @@
...
+                    let delete_plan = deletion_provider
+                        .delete_from(&session_state, &delete_where_exprs)
+                        .await
+                        .context(crate::accelerated_table::FailedToWriteDataSnafu)?;
+
+                    collect(delete_plan, ctx.task_ctx())
+                        .await
+                        .context(crate::accelerated_table::FailedToWriteDataSnafu)?;
+                }
+                "c" | "u" | "r" => {
```
ewgenius: Why `read` event handled as an upsertion?

phillipleblanc: `read` is part of the initial snapshot that Debezium ingests into Kafka. https://debezium.io/documentation/reference/2.7/connectors/postgresql.html#postgresql-snapshots

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -0,0 +1,247 @@
...
+            DataType::Utf8 => {
+                let key_col = key_col.as_any().downcast_ref::<StringArray>().context(
+                    crate::accelerated_table::ArrayDataTypeMismatchSnafu {
+                        field_name: key.to_string(),
+                        expected_data_type: "String".to_string(),
+                        schema: Arc::clone(&data_schema),
+                    },
+                )?;
+                Ok((key_col.value(0).to_string(), lit(key_col.value(0))))
+            }
```
y-f-u, line 246: nit, can use macro above?

crates/data_components/src/debezium/change_event.rs
phillipleblanc: Added an example JSON payload

y-f-u (approved):

crates/runtime/src/accelerated_table/refresh_task/changes.rs
phillipleblanc: switched to use a macro

crates/runtime/src/embeddings/connector.rs
```diff
@@ -98,17 +97,6 @@ impl DataConnector for EmbeddingConnector {
         }
     }
 
-    async fn stream_provider(
```
Jeadie, line 101: Why did we remove this?

phillipleblanc: It wasn't used, and this changes mode supersedes it.

crates/runtime/src/accelerated_table.rs
```diff
@@ -91,6 +93,27 @@ pub enum Error {
...
+    #[snafu(display("The accelerated table does not support delete operations"))]
+    AcceleratedTableDoesntSupportDelete {},
+
+    #[snafu(display("Expected schema to have field '{field_name}' schema={schema}"))]
+    ExpectedSchemaToHaveField {
+        field_name: String,
+        schema: SchemaRef,
+    },
+
+    #[snafu(display("Expected field in schema '{field_name}' to have type '{expected_data_type}' schema={schema}"))]
```
Jeadie, line 106: > `schema={schema}"`
Could make very large error messages for common tables (eg. tpch).

phillipleblanc: This technically should never happen, so I wanted to provide as much debugging context as possible if it does.

crates/runtime/src/datafusion.rs
```diff
@@ -170,6 +172,9 @@ pub enum Error {
 
     #[snafu(display("Unable to get the lock of data writers"))]
     UnableToLockDataWriters {},
+
+    #[snafu(display("The schema returned by the data connector for refresh_mode: changes does not contain a data field"))]
```
Jeadie: nit:  ` for 'refresh_mode: changes' does not`

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -0,0 +1,239 @@
...
+use data_components::delete::get_deletion_provider;
+use datafusion::logical_expr::lit;
+use datafusion::logical_expr::{col, Expr};
+use datafusion::{execution::context::SessionContext, physical_plan::collect};
+use futures::StreamExt;
+use snafu::{OptionExt, ResultExt};
+use std::sync::Arc;
+use tokio::sync::oneshot;
+
+macro_rules! extract_primary_key {
```
Jeadie, line 49: Docs/example could be helpful.

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -0,0 +1,239 @@
...
+                    let delete_plan = deletion_provider
+                        .delete_from(&session_state, &delete_where_exprs)
+                        .await
+                        .context(crate::accelerated_table::FailedToWriteDataSnafu)?;
+
+                    collect(delete_plan, ctx.task_ctx())
+                        .await
+                        .context(crate::accelerated_table::FailedToWriteDataSnafu)?;
+                }
+                "c" | "u" | "r" => {
```
Jeadie: could be good to have these as a string enum with readable names but single-character string representations.

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -0,0 +1,239 @@
...
+        let (primary_key_idx, field) = data_schema.column_with_name(key).ok_or_else(|| {
+            crate::accelerated_table::ExpectedSchemaToHaveFieldSnafu {
+                field_name: key.to_string(),
+                schema: Arc::clone(&data_schema),
+            }
+            .build()
+        })?;
+
+        let key_col = data.column(primary_key_idx);
+        match field.data_type() {
```
Jeadie, line 237: Remind me why primary keys in this context are limited to just `{Int32, Int64, Utf8}`?

Jeadie: Okay, its a limitation (see `PrimaryKeyTypeNotYetSupportedSnafu `).

Jeadie (approved):

crates/runtime/src/accelerated_table.rs
Jeadie: Good point

crates/runtime/src/accelerated_table/refresh_task/changes.rs
phillipleblanc: It covers the most common types. We can get feedback on additional types we need to support once its out.

crates/runtime/src/accelerated_table/refresh_task/changes.rs
phillipleblanc: Changed to use an enum

crates/runtime/src/accelerated_table/refresh_task/changes.rs
phillipleblanc: Added an example

PR #1827, opened 2024-06-29
Persist metadata for Debezium connector to gracefully survive restarts (DuckDB/Sqlite)

phillipleblanc:
## 🗣 Description

Creates a new metadata table `spice_sys_metadata` in the accelerator engine with a `dataset` column and a `metadata` column which stores JSON.

Implemented for DuckDB and Sqlite, the next PR will implement it for Postgres.

Also fixes a bug where we missed the first message due to not properly rewinding the offsets for the topic after doing the schema inference.

Also adds a warning when using the Debezium connector with a non-file based acceleration setting that this will cause all changes to be replayed on restart.

## 🔨 Related Issues

Part of #1785

### DIFF
--- a/crates/runtime/src/dataaccelerator/metadata.rs
+++ b/crates/runtime/src/dataaccelerator/metadata.rs
@@ -120,7 +120,7 @@ async fn get_metadata_provider(
             .await?,
         )),
         #[cfg(not(feature = "sqlite"))]
-        Engine::DuckDB => Err("Spice wasn't build with Sqlite support enabled".into()),
+        Engine::Sqlite => Err("Spice wasn't build with Sqlite support enabled".into()),
         Engine::PostgreSQL => todo!(),
         Engine::Arrow => Err("Arrow acceleration not supported for metadata".into()),
     }
--- a/crates/runtime/src/dataaccelerator/metadata.rs
+++ b/crates/runtime/src/dataaccelerator/metadata.rs
@@ -16,29 +16,30 @@ limitations under the License.
 
 //! Store and retrieve Spice metadata in durable accelerator tables
 //!
-//! The metadata table will be a new table `__spice_metadata` with two columns:
+//! The metadata table will be a new table `spice_sys_metadata` with two columns:
 //! - `dataset` (PRIMARY KEY, TEXT): The dataset the metadata entry corresponds to
 //! - `metadata` (TEXT): The metadata entry in JSON format
 
 use std::{path::Path, sync::Arc};
 
-use arrow::datatypes::SchemaRef;
 use data_components::duckdb::DuckDB;
 use duckdb::AccessMode;
 use serde::de::DeserializeOwned;
+use tokio_rusqlite::Connection;
 
 use crate::component::dataset::{acceleration::Engine, Dataset};
 use crate::dataaccelerator::DuckDBAccelerator;
-use db_connection_pool::{dbconnection::SyncDbConnection, duckdbpool::DuckDbConnectionPool};
+use db_connection_pool::duckdbpool::DuckDbConnectionPool;
 
 use super::get_accelerator_engine;
+use super::sqlite::SqliteAccelerator;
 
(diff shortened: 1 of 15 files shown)


crates/runtime/src/dataconnector/debezium.rs
```diff
@@ -197,3 +196,104 @@ impl DataConnector for Debezium {
...
+
+async fn set_metadata_to_accelerator(
+    dataset: &Dataset,
+    metadata: &DebeziumKafkaMetadata,
+) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let accelerated_metadata = AcceleratedMetadata::new_create_if_not_exists(dataset).await?;
+    accelerated_metadata.set_metadata(metadata).await
+}
+
+async fn infer_metadata_from_kafka(
```
Jeadie: I don't think this is a great name. I wouldn't expect `infer_metadata_from_kafka` to read from the topic and then restart the kafka topic.

crates/runtime/src/dataaccelerator/metadata.rs
```diff
@@ -0,0 +1,127 @@
...
+pub trait AcceleratedMetadataProvider {
+    async fn get_metadata(&self, dataset: &str) -> Option<String>;
+    async fn set_metadata(
+        &self,
+        dataset: &str,
+        metadata: &str,
+    ) -> Result<(), Box<dyn std::error::Error + Send + Sync>>;
+}
+
+async fn get_metadata_provider(
```
Jeadie, line 92: Should this be under `impl AcceleratedMetadata {` possibly as `pub(crate)`.

crates/runtime/src/dataaccelerator/metadata.rs
```diff
@@ -0,0 +1,127 @@
...
+        Engine::DuckDB => Err("Spice wasn't build with DuckDB support enabled".into()),
+        #[cfg(feature = "sqlite")]
+        Engine::Sqlite => Ok(Box::new(
+            crate::dataaccelerator::metadata::sqlite::AcceleratedMetadataSqlite::try_new(
+                dataset,
+                create_if_file_not_exists,
+            )
+            .await?,
+        )),
+        #[cfg(not(feature = "sqlite"))]
```
Jeadie, line 122: I like this. This is a pattern that I should be using for when some model flags aren't set.

crates/runtime/src/dataaccelerator/metadata/duckdb.rs
```diff
@@ -0,0 +1,116 @@
...
+        metadata: &str,
+    ) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+        let mut db_conn = Arc::clone(&self.pool)
+            .connect_sync()
+            .map_err(|e| e.to_string())?;
+        let duckdb_conn = DuckDB::duckdb_conn(&mut db_conn)
+            .map_err(|e| e.to_string())?
+            .get_underlying_conn_mut();
+
+        let create_if_not_exists = format!(
```
Jeadie, line 95: I wonder if this whole schema format should be in `crates/runtime/src/dataaccelerator/metadata.rs`

Jeadie (approved):

crates/runtime/src/dataaccelerator/metadata/duckdb.rs
phillipleblanc: Technically there could be variations in how the table is constructed depending on the underlying provider, which is why I didn't centralize it.

crates/runtime/src/dataaccelerator/metadata.rs
phillipleblanc: This is an implementation detail of `AcceleratedMetadata::new` - I don't think it should be public.

crates/runtime/src/dataconnector/debezium.rs
phillipleblanc: I changed it to `get_metadata_from_kafka` - which has a nice parallel to `get_metadata_from_accelerator`

Issue #1797, opened 2024-06-26
Refactor refresh's data update to use stream of batch record.


y-f-u:

sgrebnov:
Implemented

PR #1833, opened 2024-07-01
Streaming approach to write refresh data to accelerated table

sgrebnov:
## 🗣 Description

Implements a streaming approach to write refresh data to an accelerated table.

This PR is targeted at the feature branch to reduce potential issues with the regular Monday release.

## 🔨 Related Issues


crates/runtime/src/accelerated_table/refresh_execution_plan.rs
```diff
@@ -0,0 +1,116 @@
...
+};
+use futures::StreamExt;
+use tokio::sync::Mutex;
+
+use super::DataFusionResult;
+
+use async_stream::stream;
+
+pub struct RefreshExecutionPlan {
+    record_batch_stream: Arc<Mutex<SendableRecordBatchStream>>,
```
sgrebnov: ExecutionPlan is required to be sync => Mutex is used here.

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -182,42 +189,118 @@ impl RefreshTask {
...
+
+        let is_data_receiving_err: Arc<RwLock<bool>> = Arc::new(RwLock::new(false));
+        let is_data_receiving_err_ref_copy = Arc::clone(&is_data_receiving_err);
+
+        let dataset_name_copy = dataset_name.clone();
+
+        let schema = Arc::clone(&streaming_data_update.schema);
+
+        let observed_record_batch_stream = RecordBatchStreamAdapter::new(Arc::clone(&schema), {
+            stream! {
```
phillipleblanc: Let's use stream::unfold here instead of `stream!`

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -57,6 +58,12 @@ use datafusion::{
 use datafusion::{execution::context::SessionContext, physical_plan::collect};
 
 use super::refresh::Refresh;
+
+#[derive(Debug, Clone, Default)]
+struct RefreshStat {
+    pub num_rows: usize,
+    pub memory_size: usize,
+}
```
phillipleblanc, line 68:
```suggestion
}

```

crates/runtime/src/accelerated_table/refresh_task.rs
sgrebnov: Updated, thank you!

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -175,9 +182,9 @@ impl RefreshTask {
 
         let dataset_name = self.dataset_name.clone();
 
-        let mode = self.refresh.read().await.mode.clone();
+        let mode = self.refresh.read().await.mode;
 
-        let timer = TimeMeasurement::new(
+        let _ = TimeMeasurement::new(
```
phillipleblanc: Rust will drop the value immediately if you do this - it needs to at least be assigned to a variable so that when it goes out of scope it will be dropped.

crates/runtime/src/dataupdate.rs
```diff
@@ -75,6 +78,24 @@ impl StreamingDataUpdate {
...
+        )?) as SendableRecordBatchStream;
+        Ok(Self {
+            schema,
+            data,
+            update_type: data_update.update_type,
+        })
+    }
+}
+
 pub struct DataUpdateExecutionPlan {
```
phillipleblanc, line 99: Is anything still using this?

phillipleblanc (approved):

crates/runtime/src/dataupdate.rs
sgrebnov: Yes, it it is still used in [datafusion.write_data](https://github.com/spiceai/spiceai/blob/0839f73a8e646aae38fea02c7dd9cbd577c098d9/crates/runtime/src/datafusion.rs#L432) (used by `oss_benchmarks`, `query_history`, `opentelemetry`, `spice_metrics`) and also in [changes stream](https://github.com/spiceai/spiceai/blob/0839f73a8e646aae38fea02c7dd9cbd577c098d9/crates/runtime/src/accelerated_table/refresh_task/changes.rs#L168).

Do we want to update all such places so we have single method to write data to accelerator?

phillipleblanc: Yes, but it can be in a separate PR

sgrebnov: Noted  👍   - confirm this will be updated next.

crates/runtime/src/accelerated_table/refresh_task.rs
sgrebnov: Updated- thank you

PR #1840, opened 2024-07-01
Add `"cmake-build"` feature to `rdkafka` for windows

Jeadie:
## 🗣 Description
- From [rdkafka-sys](https://docs.rs/crate/rdkafka-sys/latest), _"Windows is only supported when using the CMake build system via the cmake-build Cargo feature."_. This can be enabled via `rdkafka` with
```toml
[dependencies]
rdkafka = { version = "0.25", features = ["cmake-build"] }
```
- Only want to add in Cmake features for windows

## 🔨 Related Issues
- Github action to test build on windows: https://github.com/spiceai/spiceai/actions/runs/9752485392/job/26916063493#step:11:23

### DIFF
--- a/crates/data_components/Cargo.toml
+++ b/crates/data_components/Cargo.toml
@@ -49,6 +49,10 @@ odbc-api = { workspace = true, optional = true }
 postgres-native-tls = { version = "0.5.0", optional = true }
 r2d2 = { workspace = true, optional = true }
 rdkafka = { version = "0.36.2", optional = true }
+
+[target.'cfg(windows)'.dependencies]
+rdkafka = { version = "0.25", features = ["cmake-build"] }
+
 regex = "1.10.4"
 reqwest = { version = "0.11.24", features = ["json"] }
 rusqlite = { workspace = true, optional = true }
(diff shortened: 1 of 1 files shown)


crates/data_components/Cargo.toml
```diff
@@ -49,6 +49,10 @@ odbc-api = { workspace = true, optional = true }
 postgres-native-tls = { version = "0.5.0", optional = true }
 r2d2 = { workspace = true, optional = true }
 rdkafka = { version = "0.36.2", optional = true }
+
+[target.'cfg(windows)'.dependencies]
+rdkafka = { version = "0.36.2", features = ["cmake-build"] }
```
ewgenius: shouldn't it keep `optional = true`?

crates/data_components/Cargo.toml
```diff
@@ -49,6 +49,10 @@ odbc-api = { workspace = true, optional = true }
 postgres-native-tls = { version = "0.5.0", optional = true }
 r2d2 = { workspace = true, optional = true }
 rdkafka = { version = "0.36.2", optional = true }
+
+[target.'cfg(windows)'.dependencies]
```
sgrebnov: Should this still be optional?
rdkafka = { version = "0.36.2", features = ["cmake-build"], optional=true }

crates/data_components/Cargo.toml
Jeadie: Yes!

crates/data_components/Cargo.toml
Jeadie: Yes

phillipleblanc (approved):

PR #1849, opened 2024-07-02
Use in-memory DB pool for DuckDB functions

Jeadie:
## 🗣 Description
- For the DuckDBFactory, create both an in-memory DB pool and a file pool (conditional on if an `open` param exists). 
- For DuckDB functions, use in-memory
- Avoids having to pass a `param.open` for DuckDB functions.
- Leads the way to reusing the `DBConnectionPool` for in-memory datasets (i.e. functions). 

### Testing
Against spicepod
```yaml
  - from: duckdb:read_csv_auto('https://docs.google.com/spreadsheets/export?format=csv&id=1Q6-rPG7F_ETHEKNG03YWBdk9CYb3QnSyYIKGVP2KC74&gid=0')
    name: gsheet

  - from: duckdb:read_parquet('cleaned_sales_data.parquet')
    name: duc
    params:
      open: tpch.db

  - from: duckdb:region
    name: region
    params:
      open: tpch.db

  - from: duckdb:embeded_cleaned_sales_data
    name: sales_data
    params:
      open: local.db
```

## 🔨 Related Issues
- https://github.com/spiceai/spiceai/pull/1742

## 🤔 Concerns
- Complexity


crates/runtime/src/dataconnector/duckdb.rs
```diff
@@ -89,12 +122,21 @@ impl DataConnector for DuckDB {
         &self,
         dataset: &Dataset,
     ) -> super::DataConnectorResult<Arc<dyn TableProvider>> {
-        Ok(
```
phillipleblanc, line 92: I think it would be cleaner to keep this like it was `duckdb_factory` and just resolve the correct one to use in the `DataConnectorFactory::create`

phillipleblanc (approved):

PR #1850, opened 2024-07-02
Streaming refresh: logic to retry on data retrieval errors

sgrebnov:
## 🗣 Description

Adds logic for 
1. Data components to detect and report data retrieval retriable errors
2. Refresh functionality to unwrap/detect transient data retrieval errors and perform retry.

Currently includes single `DataRetrievalError` error. This could be extended by adding other errors that should be retried, for example writing data to PostgreSQL (as acceleration engine).

PR is targeting to refresh refactoring feature branch

## 🔨 Related Issues

### DIFF
--- a/crates/data_components/src/util/retriable_error.rs
+++ b/crates/data_components/src/util/retriable_error.rs
@@ -18,7 +18,7 @@ use datafusion::error::DataFusionError;
 use snafu::Snafu;
 
 #[derive(Debug, Snafu)]
-pub enum TransientError {
+pub enum RetriableError {
     #[snafu(display("{source}"))]
     DataRetrievalError {
         source: datafusion::error::DataFusionError,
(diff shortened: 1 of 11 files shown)


crates/data_components/src/util/transient_error.rs
```diff
@@ -0,0 +1,56 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use datafusion::error::DataFusionError;
+use snafu::Snafu;
+
+#[derive(Debug, Snafu)]
+pub enum TransientError {
+    #[snafu(display("{source}"))]
```
sgrebnov: No need for additional text as it is just a wrapper around original DF error.

crates/data_components/src/util/retriable_error.rs
```diff
@@ -0,0 +1,60 @@
...
+pub fn is_df_retriable_error(err: &DataFusionError) -> bool {
+    match err {
+        DataFusionError::External(err) => return err.downcast_ref::<RetriableError>().is_some(),
+        DataFusionError::Context(_, err) => is_df_retriable_error(err.as_ref()),
+        _ => false,
+    }
+}
+
+#[must_use]
+pub fn detect_retriable_data_retrieval_error(err: DataFusionError) -> DataFusionError {
```
phillipleblanc:
```suggestion
pub fn check_and_mark_retriable_error(err: DataFusionError) -> DataFusionError {
```

I think this more clearly conveys what the function is trying to do. It would also be good to add function documentation on what this function does and what it considers to be a retriable error.

crates/data_components/src/util/retriable_error.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Debug, Snafu)]
+pub enum RetriableError {
+    #[snafu(display("{source}"))]
+    DataRetrievalError {
+        source: datafusion::error::DataFusionError,
+    },
+}
+
+#[must_use]
+pub fn is_df_retriable_error(err: &DataFusionError) -> bool {
```
phillipleblanc:
```suggestion
pub fn is_retriable_error(err: &DataFusionError) -> bool {
```

phillipleblanc (approved):

crates/data_components/src/util/retriable_error.rs
sgrebnov: Updated 👍, thank you

PR #1851, opened 2024-07-02
Remove DataUpdateExecutionPlan in favor of StreamingDataUpdateExecutionPlan

sgrebnov:
## 🗣 Description

Remove `DataUpdateExecutionPlan` in favor of `StreamingDataUpdateExecutionPlan`

The PR is targeting to refresh refactoring feature branch.

## 🔨 Related Issues

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task/changes.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task/changes.rs
@@ -16,7 +16,7 @@ limitations under the License.
 
 use super::RefreshTask;
 use crate::{
-    dataupdate::{DataUpdate, DataUpdateExecutionPlan, UpdateType},
+    dataupdate::{DataUpdate, UpdateType},
     status,
 };
 use arrow::array::{Int32Array, Int64Array, RecordBatch, StringArray};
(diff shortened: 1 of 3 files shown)


crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -157,31 +157,21 @@ impl RefreshTask {
...
-                            })),
-                            false,
-                        )
-                        .await
-                        .context(crate::accelerated_table::FailedToWriteDataSnafu)?;
-
-                    collect(insert_plan, ctx.task_ctx())
-                        .await
-                        .context(crate::accelerated_table::FailedToWriteDataSnafu)?;
+                    self.write_data_update(
```
phillipleblanc: This does a lot of unnecessary work that will slow down the CDC stream - since we only get changes one row at a time. Things like calculating the size of the update and marking it as ready. Let's just create the streaming plan here and insert it into the table directly.

sgrebnov: Updated 👍

phillipleblanc (approved):

Issue #1855, opened 2024-07-03
Bug: SQLite acceleration from an ODBC SQLite connection fails


peasee:
**Describe the bug**

With SQLite acceleration configured using an ODBC SQLite connection source, `spiced` outputs errors on startup for failed SQL queries:

```
2024-07-03T00:41:20.809107Z DEBUG runtime::datasets_health_monitor: Starting datasets availability monitoring
2024-07-03T00:41:20.809361Z  INFO runtime::opentelemetry: Spice Runtime OpenTelemetry listening on 127.0.0.1:50052
2024-07-03T00:41:20.810888Z  INFO runtime::http: Spice Runtime HTTP listening on 127.0.0.1:3000
2024-07-03T00:41:20.810918Z  INFO runtime::flight: Spice Runtime Flight listening on 127.0.0.1:50051
2024-07-03T00:41:20.811153Z  INFO runtime: Initialized results cache; max size: 128.00 MiB, item ttl: 1s
2024-07-03T00:41:20.811693Z DEBUG runtime::accelerated_table::refresh: Starting scheduled refresh
2024-07-03T00:41:20.811755Z DEBUG runtime::accelerated_table::refresh_task: Loading data for dataset runtime.query_history
2024-07-03T00:41:20.812200Z DEBUG runtime::accelerated_table: [retention] Evicting data for runtime.query_history where start_time < 2024-07-02T00:41:20+00:00...
2024-07-03T00:41:20.812228Z DEBUG runtime::accelerated_table: [retention] Expr BinaryExpr(BinaryExpr { left: Column(Column { relation: None, name: "start_time" }), op: Lt, right: Literal(TimestampMillisecond(1719880880812, None)) })
[...]

PR #1862, opened 2024-07-03
feat: Optimize ODBC for streaming results

peasee:
## 🗣 Description

Updates ODBC to stream results out using an inner async thread with `tokio::spawn`.

At the beginning of the operation, it clones some of the state required for the connection and query operation so that the inner async thread can take ownership of it. This also required importing [`dyn-clone`](https://github.com/dtolnay/dyn-clone) so we can clone dynamic trait objects (which we use in `ODBCParameter`), because dynamic traits cannot be cloned with `std::clone::Clone`.

The footprint of this data should be rather low making it safe to clone, as it's only connection information, parameters and an SQL query string.

The inner async thread then uses some [flume](https://github.com/zesterer/flume) channels to send schema and record batch data back to the main thread.

The batch channel is used to construct a `Send` stream, which is used for the resulting `RecordBatchStreamAdapter`.

#### Alternative 1 (ext)

`odbc-api` exposes a number of different methods that could be suitable for async streaming like we're trying to do in this PR, including:

* [`.execute_polling`](https://docs.rs/odbc-api/latest/odbc_api/struct.Connection.html#method.execute_polling)
* [`.into_cursor`](https://docs.rs/odbc-api/latest/odbc_api/struct.Connection.html#method.into_cursor)
* [`.into_concurrent`](https://docs.rs/arrow-odbc/latest/arrow_odbc/struct.OdbcReader.html#method.into_concurrent)

Nearly all of these options, and any combination of them, failed for some reason or another due to [`Stmt`](https://docs.rs/odbc-sys/0.24.0/odbc_sys/enum.Stmt.html) and [`Dbc`](https://docs.rs/odbc-sys/0.24.0/odbc_sys/enum.Dbc.html) not implementing either `Send`, `Sync` or both.

A sub-alternative to this option was forking the `odbc-*` repos and deriving `Send` and `Sync` on these `Dbc` and `Stmt` enums, as they appear to be empty enums.

However, the handles that are used for the CFFI bindings are raw pointers like `*mut Dbc` to create the type [`HDbc`](https://docs.rs/odbc-sys/0.24.0/odbc_sys/type.HDbc.html), and the connection objects that were originating the `Dbc` and `Stmt` errors used the handles not the direct types. Because these were raw pointers, implementing `Send` and `Sync` was not viable.

#### Alternative 2

Bypass the compiler checks by forcing the types to be `Send` and `Sync`, ironically [through a code sample from one of the owners of `odbc-api`](https://github.com/pacman82/force-send-sync/blob/main/src/lib.rs).

I didn't feel comfortable bypassing the compiler checks, even if it happened to be perfectly "safe". I was adamant there was another solution that didn't require fooling the compiler like this.

## 🔨 Related Issues

* Part of #1778

## 🕹️ Testing

* [x] ODBC connector with Databricks, using DuckDB acceleration :heavy_check_mark:
    * `tpch.customers`, `tpch.orders` both load their full table sets - but cannot load simultaneously in the same `spicepod.yml`. See #1864 for details.
* [x] ODBC connector with Databricks, with acceleration disabled :white_check_mark: 
* [x] ODBC connector with Databricks, using SQLite acceleration :white_check_mark: 
* [x] ODBC connector with SQLite, with acceleration disabled :white_check_mark: 
* [x] ODBC connector with SQLite, with DuckDB or SQLite acceleration enabled ❌
    * See #1855 for details.
* [ ] ODBC connector with Dremio, with acceleration disabled
* [ ] ODBC connector with Dremio, with acceleration enabled
* [ ] ODBC connector with MS SQL, with acceleration disabled
* [ ] ODBC connector with MS SQL, with acceleration enabled


crates/db_connection_pool/src/dbconnection/odbcconn.rs
```diff
@@ -137,26 +145,90 @@ where
...
-        let mut statement = prepared.into_statement();
-
-        bind_parameters(&mut statement, params)?;
-
-        // StatementImpl<'_>::execute is unsafe, CursorImpl<_>::new is unsafe
-        let cursor = unsafe {
-            statement.execute().unwrap();
-            CursorImpl::new(statement.as_stmt_ref())
+        // prepare some flume channels to communicate query results back from the thread
+        let (batch_tx, batch_rx): (
```
phillipleblanc, line 160: It seems like we could directly use streams for this, instead of channels.

Jeadie:
Is the reason we're using `flume` because _"Sender and Receiver both implement Send + Sync + Clone"_?

phillipleblanc (dismissed):

peasee:
> Is the reason we're using `flume` because _"Sender and Receiver both implement Send + Sync + Clone"_?

This is tangentially correct. We don't use tokio streams, because we need to `.await` inside the thread when we're using the Sender. `.await`'ing causes the same `Send` / `Sync` errors I'm trying to avoid from `odbc-api`.

crates/runtime/Cargo.toml
```diff
@@ -123,7 +123,7 @@ clickhouse = [
 ]
 databricks = ["data_components/databricks"]
 debezium = ["data_components/debezium"]
-default = ["keyring-secret-store", "aws-secrets-manager"]
+default = ["keyring-secret-store", "aws-secrets-manager", "odbc"]
```
```suggestion
default = ["keyring-secret-store", "aws-secrets-manager"]
```

y-f-u (dismissed):

phillipleblanc (dismissed):

y-f-u (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

Issue #1802, opened 2024-06-26
Enhancement: GraphQL Data Connector Unnesting


lukekim:
## Goal-State/What/Result

GraphQL data connector automatically unnests nested types to a specified nest depth with friendly columns names (the leaf name).

Parameter should be specified like:

```yaml
# Fetch stargazers of the spiceai repository every 1 hour, materialize locally
  - from: graphql:https://api.github.com/graphql
    name: stargazers
    params:
      json_path: data.repository.stargazers.edges
      unnest_depth: 1
```

An `unnest_depth` of undefined or 0, will result in the existing behavior (no unnesting).

Example:

Before column data: `{id: MDQ6VXNlcjcwNzIw, login: ashtom, name: Thomas Dohmke}`
After columns: `| MDQ6VXNlcjcwNzIw           | ashtom                        | Thomas Dohmke                             |`

Before column name: `unnest(stargazers.node).name`
After column name: `name`

## Why/Purpose

Significantly better developer UX. Without this feature, nested types are stored as a JSON blob in single column with a non-descriptive column name, such as `node`. The `unnest` function [can be used](https://docs.spiceai.org/data-connectors/graphql#piping-array-into-rows) however that is another step and still results in unfriendly colum names.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

Basic specification of the enhancement.

## How/MAP

A clear and concise plan of how this should be implemented.

### Implementation (issues)

- 
-

PR #1876, opened 2024-07-05
fix: Add support for varying duplicate columns behavior in GraphQL unnesting

peasee:
## 🗣 Description

Adds a new parameter for GraphQL data connectors, `unnest_duplicate_columns`.

This is a string parameter field, which accepts the values `error`, `overwrite`, and `rename` each identifying different duplicate column handling behavior:

* `error`: The default behavior when the parameter isn't specified. If a duplicate column is identified during unnesting, the data connector errors
* `overwrite`: Duplicate columns overwrite any existing column with the same name. The last visited column is the data that remains after unnesting.
* `rename`: Duplicate columns are renamed if they are identified. For example, two objects each with a sub-field of `name` would be renamed as unnested fields `name_1` and `name_2`.

_Misc:_ Also includes some contributor guide updates after I reinstalled my OS and tried using the guide.

## 🔨 Related Issues

* Part of #1802

### DIFF
--- a/crates/runtime/src/dataconnector/graphql.rs
+++ b/crates/runtime/src/dataconnector/graphql.rs
@@ -38,7 +38,7 @@ use snafu::{ResultExt, Snafu};
 use std::{any::Any, collections::HashMap, future::Future, io::Cursor, pin::Pin, sync::Arc};
 use url::Url;
 
-use super::{DataConnector, DataConnectorFactory};
+use super::{DataConnector, DataConnectorError, DataConnectorFactory};
 
 #[derive(Debug, Snafu)]
 pub enum Error {
@@ -79,8 +79,16 @@ Please verify the syntax of your GraphQL query."#
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
 #[derive(Debug, PartialEq, Eq)]
+enum OverwriteBehavior {
+    Error,
+    Overwrite,
+    Rename,
+}
+
+#[derive(Debug)]
 struct UnnestParameters {
     depth: usize,
+    overwrite_behavior: OverwriteBehavior,
 }
 
 #[derive(Debug, PartialEq, Eq)]
@@ -212,6 +220,7 @@ fn unnest_json_object(unnest_parameters: &UnnestParameters, object: &Value) -> R
     let mut new_objects = Vec::new();
     if let Value::Object(obj) = object {
         let mut new_object = obj.clone();
+        let mut key_counter: HashMap<String, usize> = HashMap::new();
 
         // setup some loop controls
         let mut depth_counter = 0;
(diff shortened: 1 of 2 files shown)


crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -245,7 +254,47 @@ fn unnest_json_object(unnest_parameters: &UnnestParameters, object: &Value) -> R
 
             // add the staged additions back to the root object
             for (key, value) in additions {
-                new_object.insert(key, value);
+                let new_key = match unnest_parameters.overwrite_behavior {
+                    OverwriteBehavior::Error => {
+                        if new_object.contains_key(&key) {
+                            return Err(Error::InvalidObjectAccess {
+                                message: format!(
+                                    "Column {key} already exists in the object and overwrite is set to error."
```
phillipleblanc:
```suggestion
                                    "Column {key} already exists in the object"
```

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -245,7 +254,47 @@ fn unnest_json_object(unnest_parameters: &UnnestParameters, object: &Value) -> R
...
+                        let mut new_key = key.clone();
+                        let counter = key_counter
+                            .entry(key.clone())
+                            .and_modify(|e| *e += 1)
+                            .or_insert(1);
+
+                        if new_object.contains_key(&key) {
+                            let Some(existing_value) = new_object.remove(&key) else {
+                                return Err(Error::InvalidObjectAccess {
+                                    message: format!("Column {key} already exists in the object and overwrite is set to rename, but the key does not exist."),
```
phillipleblanc:
```suggestion
                                    message: format!("Column {key} already exists in the object, but the key does not exist."),
```

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -245,7 +254,47 @@ fn unnest_json_object(unnest_parameters: &UnnestParameters, object: &Value) -> R
 
             // add the staged additions back to the root object
             for (key, value) in additions {
-                new_object.insert(key, value);
+                let new_key = match unnest_parameters.overwrite_behavior {
```
phillipleblanc: Recommend refactoring this into a separate method.

crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -245,7 +254,47 @@ fn unnest_json_object(unnest_parameters: &UnnestParameters, object: &Value) -> R
...
+                            .or_insert(1);
+
+                        if new_object.contains_key(&key) {
+                            let Some(existing_value) = new_object.remove(&key) else {
+                                return Err(Error::InvalidObjectAccess {
+                                    message: format!("Column {key} already exists in the object and overwrite is set to rename, but the key does not exist."),
+                                });
+                            };
+
+                            new_object.insert(format!("{key}_{counter}"), existing_value);
```
phillipleblanc: In theory, `counter` should always be 1 here right?

peasee: That's correct, could probably rewrite to just set it to 1 in the format

phillipleblanc (approved):

Issue #1696, opened 2024-06-14
Enhancement: Move Databricks Delta implementation from delta-rs to delta-kernel-rs


digadeesh:
## Goal-State/What/Result

The Spice Databricks Deltalake mode uses [`delta-kernel-rs`](https://github.com/delta-incubator/delta-kernel-rs) instead of `delta-rs`. Spice will also support connecting to any object store that Databricks supports (i.e. AWS S3, Azure Blob, GCS), not just S3.

## Why/Purpose

Connecting directly to object store to perform queries on Deltalake tables will be the most performant way to query Databricks data stored in Deltalake tables. Unfortunately the current implementation of this approach bypasses the Unity Catalog, which is a non-starter for many enterprises that would like to control data governance via a centralized catalog. Fortunately, those customers have an alternative in Spice by using Spark Connect or ODBC. The downside is a loss in performance relative to what can be achieved with a direct Deltalake connection.

The `delta-kernel-rs` incubation project is a lightweight implementation of the Deltalake protocol and allows Spice to remove usage of `delta-rs` which is a burden on maintainability since the Spice maintainers often need to manage [a fork](https://github.com/spiceai/delta-rs/tree/spiceai-39) to ensure alignment on DataFusion versions. `delta-rs` provides many more features than Spice actually uses, so a smaller binary is possible by removing `delta-rs` as well.

The downside of replacing `delta-rs` with `delta-kernel-rs` is that the kernel project is still in alpha and may have bugs that the `delta-rs` project doesn't. Another downside is that the kernel project only (currently?) supports querying, not updating delta tables, whereas the `delta-rs` project supports both.

## By When

**Issue/Spec written and reviewed:** July 2, 2024
**Done-Done:** July 4, 2024

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

The current YAML for specifying a direct connection is by setting a `mode` option to `s3`:

```yaml
params:
[...]

PR #1878, opened 2024-07-05
Replace `delta-rs` with `delta-kernel-rs` and add new `delta` data connector.

phillipleblanc:
## 🗣 Description

Replace the deltalake table provider from `delta-rs` with our own implementation based on `delta-kernel-rs`. This also changes the `mode` from `s3` to `delta` and removes the `format` parameter, as based on the spec in #1696

Also add a new `delta` data connector that can be passed a path to a delta table either locally or in object storage.

Example usage of the new `delta` connector.
```yaml
datasets:
  - from: delta:/Users/phillip/data/logs_delta
    name: logs
  - from: delta:s3://databricks-workspace-stack-473ec-bucket/unity-catalog/8429759719709712/__unitystorage/catalogs/fa937541-ca24-4f62-a93b-088db0fb40af/tables/60f9be16-e156-4868-b7c6-0a1660939413/
    name: traces
```

Example usage of the modified `databricks` connector, which logically is the same as the 2nd one above.
```yaml
  - from: databricks:spiceai_sandbox.default.traces
    name: traces
    params:
      mode: delta
      endpoint: https://{databricks_workspace}.cloud.databricks.com/
```

This dataset would call the Databricks Unity Catalog API to resolve the S3 address, and from this point they both work the same.

The `delta-kernel-rs` project will load a snapshot from the Delta table, which is a log of the relevant parquet files and their deletion vectors (used to mark a row as "deleted" without rewriting the parquet file). This PR hooks into the callback that visits each active Parquet file in the current snapshot, and we use the built-in DataFusion `ParquetExec` to read the parquet file. `delta-kernel-rs` has its own built-in Parquet reader, but by leveraging the DataFusion one we automatically get all of the optimizations that have been made by the DataFusion community, plus it natively supports the filter pushdown of DataFusion (i.e. only select certain columns, apply predicates to filter out reading row groups that can't contain any data we are interested in, etc).

I still need to do some more testing to make sure we aren't regressing anything we get from `delta-rs` - but given that we're just reading parquet files, it seems fairly straightforward.

I also looked into the Unity Catalog temporary credentials support - and its currently in private preview. I applied to have our workspace enabled - but we can't rely on that being enabled for everyone right now.

## 🔨 Related Issues

Part of #1696

## 🤔 Concerns

I need to test the deletion vector support more carefully. If we do prune out row groups from the `ParquetExec`, that probably will make the deletion vector mis-aligned. If so, then we'll need some way to make them aligned again. An alternative is to convert the DataFusion expressions into the ones expected by the `delta-kernel-rs` parquet reader so that it stays aligned.


crates/runtime/src/dataconnector.rs
```diff
@@ -246,6 +248,8 @@ pub async fn register_all() {
     register_connector_factory("localhost", localhost::LocalhostConnector::create).await;
     #[cfg(feature = "databricks")]
     register_connector_factory("databricks", databricks::Databricks::create).await;
+    #[cfg(feature = "databricks")]
```
ewgenius, line 251: maybe worth a separate `delta` feature?

ewgenius (dismissed):

digadeesh (dismissed):

peasee (dismissed):

y-f-u (approved):

lukekim (approved):
Might want to add more output and logging in a future PR.

peasee (approved):

crates/runtime/src/dataconnector.rs
phillipleblanc: Will handle in a future PR

PR #1891, opened 2024-07-08
Handle deletion vectors for `delta` tables

phillipleblanc:
## 🗣 Description

Handles the deletion vectors for `delta` tables properly. Instead of applying the deletion vectors after the fact, which can be inefficient for large tables - we convert the deletion vector into a `ParquetAccessPlan` which the `ParquetExec` execution plan will use to skip reading parts of the file. If lots of rows are deleted, this can result in entire row groups that we can skip - which will speed up the execution.

I initially implemented this as part of the execution phase - but once I realized I could use `ParquetExec` directly, then I moved the logic that walks each parquet file and applies the deletion vector to the planning phase. Then we just return the `ParquetExec` directly and datafusion takes care of the rest.

## 🔨 Related Issues

Part of #1696


lukekim (approved):

crates/data_components/src/delta.rs
```diff
@@ -476,82 +338,163 @@ fn handle_scan_file(
...
+
+    let total_rows = parquet_metadata
+        .row_groups()
+        .iter()
+        .map(RowGroupMetaData::num_rows)
+        .sum::<i64>();
+
+    let selection_vector = get_full_selection_vector(&selection_vector, total_rows as usize);
+
+    // Create a ParquetAccessPlan that will be used to skip rows based on the selection vector
```
alamb, line 438: Nice!

PR #1905, opened 2024-07-09
fix: Detect and error on duplicate component names on spiced (re)load

### LINKED ISSUE #1772: Loading more than one dataset with the same name leads to unexpected behavior.
It is possible to define more than one dataset with the same name, when runtime already started, and runtime doesn't warn about this - it unloads existing and loads newly added dataset

peasee:
## 🗣 Description

* Explicitly checks and errors on duplicate component `name`s during `spiced` load or reload

## 🔨 Related Issues

* Closes #1772


crates/runtime/src/lib.rs
```diff
@@ -360,13 +360,32 @@ impl Runtime {
...
         app.views
             .iter()
             .cloned()
             .map(View::try_from)
             .zip(&app.views)
             .filter_map(|(view, spicepod_view)| match view {
-                Ok(view) => Some(view),
+                Ok(view) => {
+                    // only load this view if the name isn't used by an existing dataset
+                    if datasets.contains(&view.name) {
```
phillipleblanc, line 376:
```suggestion
                    if !datasets.contains(&view.name) {
```
Shouldn't this be inverted?

phillipleblanc (approved):

Issue #1810, opened 2024-06-27
Enhancement: Databricks/UC/Spice.ai Catalog Providers


digadeesh:
## Goal-State/What/Result

Spice OSS can be used to connect to catalog providers to query all federated tables inside of a catalog. Initial catalog providers supported will be Databricks (via [Databricks-hosted Unity Catalog](https://www.databricks.com/product/unity-catalog)), [Unity Catalog OSS](https://github.com/unitycatalog/unitycatalog/tree/main) and Spice.ai.

## Why/Purpose

Configuring datasets individually becomes tedious beyond a handful. Integrating with catalog providers makes it easier to explore data and test it with Spice.

## By When

**Issue/Spec written and reviewed:** Tue, July 9
**Done-Done:** Wed, July 10

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

- A new `catalogs` top-level configuration will be added to the Spicepod.
- The `from` field specifies the catalog provider with an optional colon (`:`) and an identifier the catalog provider can use to determine which catalog is requested.
- The `name` field specifies the catalog registered in spice, and any schema/tables are registered automatically. (i.e. registering a catalog with `my_catalog` automatically registers any tables as `my_catalog.auto_schema.auto_table`). The `spice` name is reserved and will be rejected.
- A `params` field specifies any options needed for the catalog provider connection.
- A `dataset_params` field specifies any options needed for connecting to the tables in the catalog provider.
- An `include` field is an array of regex filters that are OR'ed together to filter down the list of tables
- Only federated queries will be supported, Accelerations need to be specified one dataset at a time in `datasets`.

### Unity Catalog Provider
The Unity Catalog Provider is invoked via `unity_catalog`. The colon `:` is required, and the identifier is the full URL to the [`getCatalog`](https://github.com/unitycatalog/unitycatalog/blob/main/api/Apis/CatalogsApi.md#getcatalog) Rest API call that returns the catalog details.

[...]

PR #1925, opened 2024-07-10
Implement `spice.ai` `CatalogProvider`

phillipleblanc:
## 🗣 Description

Implements a Catalog Provider for the hosted Spice.ai data platform. By performing a `spice login` and then creating a Spicepod with:

```yaml
version: v1beta1
kind: Spicepod
name: catalog_test

catalogs:
  - from: spiceai
    name: spiceai
```

Spice will now load all of the tables under the `spiceai` namespace and make them available for federated query.

Here is a screenshot of DBeaver connected to my local Spice runtime with the spiceai catalog loaded:
Screenshot 2024-07-10 at 5 22 51 PM

### ⚠️ Breaking Change
I chose to put the logic that implements the Catalog Provider into the `spice_cloud` extension. Normally this extension is only loaded if the user adds an `extension` field to their yaml like this:

```yaml
version: v1beta1
kind: Spicepod
name: catalog

extensions:
  spice_cloud:
    enabled: true
    params:
       <optional params>
```

But I've allowed that extension to "autoload" if its required and it isn't explicitly loaded.

Normally when the `spice_cloud` extension is loaded, it will register a metrics table to start replicating to the Spice.ai cloud platform. I didn't want that behavior by default, so I made a breaking change to only enable the metrics if `metrics: "true"` is specified in the params. i.e.

```yaml
extensions:
  spice_cloud:
    enabled: true
    params:
       metrics: "true"
```

### Details
- Added ` .with_autoload_extensions` to the Runtime builder to specify which extensions we want to autoload if not present. I've added the `spice_cloud` extension to this list.
- In order to prevent the default schema inference behavior, I added a `.with_schema` method to the runtime `Dataset` struct. I modified the `Read` and `ReadWrite` traits to accept an optional schema. If a schema is specified, then we can skip the schema inference step (which saves a lot of time if we're loading a lot of tables)
- I added a `catalog_provider` method to the `DataConnector` trait. It returns None for all connectors except SpiceAI right now.
- I noticed the `ReadWrite` trait was implemented for Databricks - this isn't correct as we don't support writing to Databricks via Spark Connect, so while I was in the code I decided to remove it.

## 🔨 Related Issues

Part of #1810

## 🤔 Concerns

The normal way we do schema inference was too slow - so I took the limited schema information returned by the `https://data.spiceai.io/v1/schemas` API and used that to construct Arrow schemas. This isn't perfect, since there isn't enough information to reconstruct the full type information - but it works well enough in practice. Once we implement FlightSQL support on the Spice AI cloud side, then we could remove that and replace it with a FlightSQL-based schema inference.

### DIFF
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -100,12 +100,10 @@ pub async fn run(args: Args) -> Result<()> {
 
     let mut extension_factories: Vec<Box<dyn ExtensionFactory>> = vec![];
 
-    if cfg!(feature = "spice-cloud") {
-        if let Some(app) = &app {
-            if let Some(manifest) = app.extensions.get("spice_cloud") {
-                let spice_extension_factory = SpiceExtensionFactory::new(manifest.clone());
-                extension_factories.push(Box::new(spice_extension_factory));
-            }
+    if let Some(app) = &app {
+        if let Some(manifest) = app.extensions.get("spice_cloud") {
+            let spice_extension_factory = SpiceExtensionFactory::new(manifest.clone());
+            extension_factories.push(Box::new(spice_extension_factory));
         }
     }
 
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -16,6 +16,7 @@ limitations under the License.
 
 #![allow(clippy::missing_errors_doc)]
 
+use std::collections::HashMap;
 use std::env;
 use std::net::SocketAddr;
 use std::path::PathBuf;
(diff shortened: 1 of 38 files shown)


bin/spiced/src/lib.rs
```diff
@@ -110,7 +111,13 @@ pub async fn run(args: Args) -> Result<()> {
 
     let rt: Runtime = Runtime::builder()
         .with_app_opt(app)
+        // User configured extensions
         .with_extensions(extension_factories)
+        // Extensions that will be auto-loaded if not explicitly loaded and requested by a component
+        .with_autoload_extensions(HashMap::from([(
+            "spice_cloud".to_string(),
```
ewgenius, line 116: that should be under the feature flag `"spice-cloud"`

phillipleblanc: Removed

digadeesh (dismissed):

ewgenius (dismissed):

crates/spice_cloud/src/schema.rs
```diff
@@ -0,0 +1,98 @@
...
+        self
+    }
+
+    /// Retrieves the list of available table names in this schema.
+    fn table_names(&self) -> Vec<String> {
+        self.tables.keys().cloned().collect()
+    }
+
+    /// Retrieves a specific table from the schema by name, if it exists,
+    /// otherwise returns `None`.
```
y-f-u: nit. Returns Ok(None).

ewgenius (approved):

ewgenius (approved):

peasee (approved):

PR #1930, opened 2024-07-10
fix: GraphQL Data Connector - Change json path to json pointer

### LINKED ISSUE #1858: GraphQL `json_path` is not really a JSON path
**Describe the bug**
Attempting to use more advanced JSON Path syntax does not work, because its not actually JSON Path. This is not well documented and misleading.

It isn't actually JSON Path, but a JSON Pointer as defined by [RFC 6901 - JavaScript Object Notation (JSON) Pointer (ietf.org)](https://datatracker.ietf.org/doc/html/rfc6901) - which is very different and much more limited.

We should rename `json_path` to `json_pointer` and provide better documentation on how what syntax is allowed and common examples on how to use it.

digadeesh:
## 🗣 Description

With proper documentation, this closes: #1858

This is a**breaking change**

Our json_path in the graphql data connector is really a json pointer, but it follows the format / syntax of a json path.  This change makes it a json pointer and requires the syntax of json pointer. 

Tested with multiple GraphQL api's and with / without unnest depth.


crates/runtime/src/dataconnector/graphql.rs
```diff
@@ -576,18 +576,17 @@ impl GraphQL {
                 message: "Invalid URL in dataset `from` definition",
             },
         )?;
-        let json_path = self
+        let json_pointer = self
             .params
-            .get("json_path")
-            .ok_or("`json_path` not found in params".into())
+            .get("json_pointer")
+            .ok_or("`json_pointer` not found in params".into())
```
y-f-u, line 582: Since this is a breaking change, some tests should've been broken if there is any? If not, it might be good to consider adding some.

digadeesh: Yes I fixed the test.  Updated.

crates/runtime/tests/graphql/mod.rs
```diff
@@ -304,7 +304,7 @@ async fn test_graphql_pagination() -> Result<(), String> {
             &format!("http://{addr}/graphql"),
             "test_graphql",
             "query { paginatedUsers(first: 2) { users { id name posts { id title content } } pageInfo { hasNextPage endCursor } } }",
-            "data.paginatedUsers.users",
+            "/data/paginatedUsers/users",
```
y-f-u, line 307: do we implement escape rule for json pointer?

y-f-u: a quick ai search says the json pointer and json path use different escape rule, which might be worthy to check, and potential add some tests if that is considered important.

digadeesh: The code actually previously just converted the path to a json pointer.  I'll double check the escaping.

digadeesh: We used Value.pointer() from serde_json Value.  It does all of the escaping and conforms to the RFC.  Since we're not adding additional logic here it doesn't make senses to add testing around this.  

https://docs.rs/serde_json/latest/serde_json/enum.Value.html

sgrebnov (approved):

Issue #1931, opened 2024-07-10
Enhancement: Enable ODBC query push down for table joins


peasee:
REMEMBER, BE **SMART**!

S: Specific
M: Measurable
A: Achievable
R: Relevant
T: Time-Bound

## Goal-State/What/Result

ODBC queries from the same ODBC source have query push down enabled for table joins.

## Why/Purpose

To offload query compute to the database instead of performing full table scans and performing queries on the client, reducing memory usage and improving query performance.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** 2024/07/12
**Done-Done:** 2024/07/16

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

* Enable `JoinPushDown` on ODBC connections by changing the value to `JoinPushDown::AllowedFor(_)`
* Hash the ODBC connection string to use like `JoinPushDown::AllowedFor(hashed(self.connection_string))`

Example psuedo-code:
```rust
JoinPushDown::AllowedFor(hashed(self.connection_string))
```

Example `EXPLAIN` with hashing enabled:
```
| physical_plan | SchemaCastScanExec                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     |
|               |   RepartitionExec: partitioning=RoundRobinBatch(20), input_partitions=1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
|               |     VirtualExecutionPlan name=odbc compute_context=269952280446484108
```

The `compute_context=269952280446484108` is the hashed connection string.

## How/MAP

[...]

PR #1954, opened 2024-07-12
feat: Enable ODBC JoinPushDown with hashed connection string

peasee:
## 🗣 Description

Enables `JoinPushDown` for ODBC connectors, using the hashed connection string from the dataset as the comparison.

This allows ODBC connections from the same source to use federated joins, preventing full table scans and loading table contents into memory first.

## 🔨 Related Issues

* Part of #1931


crates/db_connection_pool/src/odbcpool.rs
```diff
@@ -70,9 +81,13 @@ impl ODBCPool {
             .map(Secret::expose_secret)
             .map(ToString::to_string)
             .context(MissingConnectionStringSnafu)?;
+
+        let connection_id = hash_string(&connection_string);
```
phillipleblanc, line 87: Let's add a comment for why we're hashing here

phillipleblanc (approved):

PR #1962, opened 2024-07-14
Fix bug: arrow acceleration reports zero results during refresh

### LINKED ISSUE #1959: Bug: arrow acceleration reports zero results during refresh
**Describe the bug**
When I use `spiceai/quickstart` and run my query between `Loading data for dataset taxi_trips` and `Loaded 2,964,624 rows ..` I get empty result

```
select count(*) from taxi_trips
+----------+
| COUNT(*) |
+----------+
| 0        |
+----------+

Time: 0.003906875 seconds. 1 rows (cached).
```

**To Reproduce**

1. Follow Spice quick start: https://github.com/spiceai/spiceai?tab=readme-ov-file#%EF%B8%8F-quickstart-local-machine
2. Wait until data is loaded and run test query 
```
select count(*) from taxi_trips
+----------+
| COUNT(*) |
+----------+
| 2964624  |
+----------+

Time: 0.007432792 seconds. 1 rows.
```
4. Wait for accelerated table refresh started (`Loading data for dataset taxi_trips`) and repeat the query
```
select count(*) from taxi_trips
+----------+
| COUNT(*) |
+----------+
| 0        |
+----------+

Time: 0.00604075 seconds. 1 rows.
```

**Expected behavior**
Query should return correct result

**Additional context**
No such issue when using `duckdb` acceleration engine

sgrebnov:
## 🗣 Description

Fixes https://github.com/spiceai/spiceai/issues/1959:
- Update Arrow write logic to clear exiting data only after data update is fully collected
- Make update operation atomic: clear + write during partitions lock

## 🔨 Related Issues

[Bug: arrow acceleration reports zero results during refresh](https://github.com/spiceai/spiceai/issues/1959):

### DIFF
--- a/crates/data_components/src/arrow/write.rs
+++ b/crates/data_components/src/arrow/write.rs
@@ -239,14 +239,6 @@ impl DataSink for MemSink {
     ) -> Result<u64> {
         let num_partitions = self.batches.len();
 
-        if self.overwrite {
-            for partition in &self.batches {
-                let mut partition_vec = partition.write().await;
-                partition_vec.clear();
-                drop(partition_vec);
-            }
-        }
-
         // buffer up the data round robin style into num_partitions
 
         let mut new_batches = vec![vec![]; num_partitions];
@@ -263,10 +255,14 @@ impl DataSink for MemSink {
             i = (i + 1) % num_partitions;
         }
 
-        // write the outputs into the batches
-        for (target, mut batches) in self.batches.iter().zip(new_batches.into_iter()) {
-            // Append all the new batches in one go to minimize locking overhead
-            target.write().await.append(&mut batches);
+        let mut writable_targets: Vec<_> =
+            futures::future::join_all(self.batches.iter().map(|target| target.write())).await;
+
+        for (target, mut batches) in writable_targets.iter_mut().zip(new_batches.into_iter()) {
+            if self.overwrite {
+                target.clear();
+            }
+            target.append(&mut batches);
         }
 
         Ok(row_count as u64)


crates/data_components/src/arrow/write.rs
```diff
@@ -263,10 +255,14 @@ impl DataSink for MemSink {
             i = (i + 1) % num_partitions;
         }
 
-        // write the outputs into the batches
-        for (target, mut batches) in self.batches.iter().zip(new_batches.into_iter()) {
-            // Append all the new batches in one go to minimize locking overhead
-            target.write().await.append(&mut batches);
+        let mut writable_targets: Vec<_> =
```
sgrebnov, line 258: We currently use a single partition (`self.batches`) only for data acceleration. However, locking at the partition level could potentially lead to issues where some partitions have already been updated while others have not, resulting in an invalid state during a refresh when we perform queries simultaneously. To prevent such cases, this PR locks all partitions first before writing/updating them. Note: At present, there is no difference from the previously used approach since we only use a single partition, but locking and updating all partitions at once seems to be the right approach and prevent potential issues in the future when we use multiple partitions.

Scan code is below

```rust
 async fn scan(
        &self,
        _state: &SessionState,
        projection: Option<&Vec<usize>>,
        _filters: &[Expr],
        _limit: Option<usize>,
    ) -> Result<Arc<dyn ExecutionPlan>> {
        let mut partitions = vec![];
        for arc_inner_vec in &self.batches {
            let inner_vec = arc_inner_vec.read().await;
            partitions.push(inner_vec.clone());
        }
        Ok(Arc::new(MemoryExec::try_new(
            &partitions,
            self.schema(),
            projection.cloned(),
        )?))
    }
```

phillipleblanc (approved):

PR #1990, opened 2024-07-16
Spice cli `spice chat` command, to interact with deployed spiced instance in spice.ai cloud

ewgenius:
Experimental cli feature, hidden under cli feature fag `cloud`.

Run `SPICE_CLI_FEATURES="cloud" make install-cli-dev` to enable and build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,10 @@ all: build
 build-cli:
 	make -C bin/spice
 
+.PHONY: build-cli-dev
+build-cli-dev:
+	export DEV=true; make -C bin/spice
+
 .PHONY: build-runtime
 build-runtime:
 	make -C bin/spiced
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -8,6 +8,10 @@ all: build
 build-cli:
 	make -C bin/spice
 
+.PHONY: build-cli-dev
+build-cli-dev:
+	export DEV=true; make -C bin/spice
```
lukekim, line 13: Doesn't this also need `SPICED_NON_DEFAULT_FEATURES="models"`?

bin/spice/cmd/cloud.go
```diff
@@ -0,0 +1,142 @@
...
+			if strings.ToLower(message) == "exit" {
+				fmt.Println("Goodbye!")
+				break
+			}
+
+			done <- true
+
+			for _, char := range message {
+				cmd.Printf("%c", char)
+				time.Sleep(10 * time.Millisecond)
```
lukekim: Let it fly.

```suggestion
```

lukekim (dismissed):

Makefile
ewgenius: that is only for cli build

lukekim (approved):

Issue #1989, opened 2024-07-16
Enhancement: Add parameter for ODBC SQL dialect


peasee:
REMEMBER, BE **SMART**!

S: Specific
M: Measurable
A: Achievable
R: Relevant
T: Time-Bound

## Goal-State/What/Result

A parameter for ODBC connections should be added to specify the SQL dialect of the ODBC connection.

Currently, we assume one single and default dialect for every ODBC connection. Because of the generic nature of ODBC though, some connections may not support this dialect and instead support a different dialect (like PostgreSQL vs MySQL for specifics like `INTERVAL`).

## Why/Purpose

To better support more ODBC connection types, providing the user with an escape hatch to change dialect if their connection type does not the default dialect.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** 2024-07-16 PST
**Done-Done:** 2024-07-19 PST

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

1. Attempt automating selecting a dialect if none are specified, based on the contents of the connection string.
2. If automated dialect detection is not possible, fallback to the `DefaultDialect`.
3. Provide a parameter, `sql_dialect`, to allow a user to specify a dialect for a connection and use it if specified.

### User-provided dialect

By providing a new parameter, users can supply their own dialect through a set of enum values. For example, `sql_dialect: postgresql`, `sql_dialect: mysql`, etc.

### Automated dialect detection

Dialect could potentially be detected from the `odbc_connection_string` for a connection, by performing string matching against the selected `Driver=` as it would be the only guaranteed key value to exist.

For example, if the value of `Driver` was `{PostgreSQL Unicode}` or `psqlodbcw.so` the runtime could set the dialect to `PostgreSqlDialect`.

This removes the need for users to manually select a different `sql_dialect` for well-known drivers that are known to use a different dialect from the default.

### Join Push Down Behavior

[...]

PR #1997, opened 2024-07-17
feat: Autodetect ODBC dialect

peasee:
## 🗣 Description

* Adds a mechanism to automatically detect and change the SQL dialect for an ODBC connection, based on the contents of the `Driver=` portion of the connection string.

An additional PR will follow, adding the ability to set a `sql_dialect` parameter to override the auto-detected one.

## 🔨 Related Issues

* Part of #1989


crates/data_components/src/odbc.rs
```diff
@@ -81,11 +88,18 @@ where
     ) -> Result<Arc<dyn TableProvider + 'static>, Box<dyn std::error::Error + Send + Sync>> {
         let pool = Arc::clone(&self.pool);
         let dyn_pool: Arc<ODBCDbConnectionPool<'a>> = pool;
-        let table_provider = Arc::new(
-            SqlTable::new("odbc", &dyn_pool, table_reference, Some(Engine::ODBC))
-                .await
-                .map_err(|e| Box::new(e) as Box<dyn std::error::Error + Send + Sync>)?,
-        );
+
+        let table = SqlTable::new("odbc", &dyn_pool, table_reference, Some(Engine::ODBC))
```
phillipleblanc, line 92: Does the `Engine::ODBC` still make sense to have given the different dialects?

crates/runtime/src/dataconnector/odbc.rs
```diff
@@ -47,6 +53,17 @@ where
...
 }
 
+fn match_driver_to_dialect(driver: &str) -> Option<Arc<dyn Dialect + Send + Sync>> {
+    match driver {
+        _ if driver.contains("mysql") => Some(Arc::new(MySqlDialect {})), // odbcinst.ini profile name
+        _ if driver.contains("libmyodbc") => Some(Arc::new(MySqlDialect {})), // library filename
+        _ if driver.contains("postgres") => Some(Arc::new(PostgreSqlDialect {})), // odbcinst.ini profile name
+        _ if driver.contains("psqlodbc") => Some(Arc::new(PostgreSqlDialect {})), // library filename
+        _ if driver.contains("sqlite") => Some(Arc::new(SqliteDialect {})), // profile and library name
+        _ => None,
```
phillipleblanc: Let's add a `tracing::debug!` if it is None to say we couldn't auto-detect the dialect for the driver, and include the driver string.

crates/data_components/src/odbc.rs
peasee: It could still make sense, but I think the only other option we could supply here that would also make sense would be `None`.

crates/runtime/src/dataconnector/odbc.rs
```diff
@@ -47,6 +53,20 @@ where
     odbc_factory: ODBCTableFactory<'a>,
 }
 
+fn match_driver_to_dialect(driver: &str) -> Option<Arc<dyn Dialect + Send + Sync>> {
+    match driver {
+        _ if driver.contains("mysql") => Some(Arc::new(MySqlDialect {})), // odbcinst.ini profile name
+        _ if driver.contains("libmyodbc") => Some(Arc::new(MySqlDialect {})), // library filename
+        _ if driver.contains("postgres") => Some(Arc::new(PostgreSqlDialect {})), // odbcinst.ini profile name
+        _ if driver.contains("psqlodbc") => Some(Arc::new(PostgreSqlDialect {})), // library filename
+        _ if driver.contains("sqlite") => Some(Arc::new(SqliteDialect {})), // profile and library name
```
y-f-u: it might complicate things, but kinda nice to have the driver type as enums then we can have something like
match driver.try_into() {
  OBDCDriver::MySQL
  }

y-f-u (dismissed):

crates/runtime/src/dataconnector/odbc.rs
```diff
@@ -47,6 +53,20 @@ where
...
+        _ => {
+            tracing::debug!("No dialect detected for driver: {}", driver);
```
Sevenannn: Might be better to return an error and handle it in the caller below

Sevenannn (dismissed):

crates/data_components/src/odbc.rs
phillipleblanc: Right, I mean - what does this actually change, and is this really a dialect not an engine that we should be passing?

crates/runtime/src/dataconnector/odbc.rs
peasee: https://github.com/spiceai/spiceai/pull/2008

crates/data_components/src/odbc.rs
peasee: It changes the behavior of `to_sql_with_engine` for some expressions. E.g. for columns: https://github.com/datafusion-contrib/datafusion-table-providers/blob/02081887de7b64dfd816ef86716f8b0853ae3196/src/sql/sql_provider_datafusion/expr.rs#L48

Do we use `to_sql_with_engine`? If we bypass it in favor of something else then we can probably supply `None`.

phillipleblanc: It shouldn't be used - its a fallback in case federation isn't enabled, but we always enable federation now. At some point we should remove our own expr_to_sql implementation and just rely on the DataFusion one, in which case this gets removed anyway.

phillipleblanc (dismissed):

sgrebnov (dismissed):

phillipleblanc (approved):

Issue #1701, opened 2024-06-15
Enhancement: Load secrets from multiple secret stores & secrets UX refresh


phillipleblanc:
## Goal-State/What/Result

Simple, easy-to-understand secrets UX that allows specifying which store a secret should be loaded from.

## Why/Purpose

The current experience for configuring secrets is hard to understand and the code to support it is more complicated than it needs to be. We also want to give flexibility to load secrets from environment variables + another secret store.

## By When

**Issue/Spec written and reviewed:** July 15, 2024
**Done-Done:** July 19, 2024

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

The current UX for configuring secrets relies on a lot of "magic" that requires understanding that secrets are namespaced based on the connector that is used. The default file we use is a TOML file instead of a more familiar `.env` most developers have used before. It isn't clear from the spicepod YAML if a secret is needed or not, that feedback only comes at runtime and is hard to discover.

This enhancement will make the secrets experience much simpler and easier to understand. It will be explicit, so its easy to see when secrets are required and where they come from. It will also make the code simpler by removing concepts that are no longer required.

The secrets implementation will be changed as follows:
- The "secret" namespace will be removed. All secrets will live in a flat namespace that isn't separated by connector.
- The method of specifying `<secret_name>-key` in the `params` for telling Spice which secret to use will be removed.
- Multiple secret stores can be specified at the top of a Spicepod. The `file` and `env` secrets stores will be combined into a single `env` secret store that reads from the environment, but also natively knows how to read from a `.env` file. The `.env` file is the [standard format developers](https://www.npmjs.com/package/dotenv#%EF%B8%8F-usage) expect. We will no longer have a global auth file, i.e. `~/.spice/auth` will no longer be referenced. Only project-specific `.env` files.

```yaml
secrets:
  - from: env
    name: env
    params:
[...]

PR #2005, opened 2024-07-17
Support loading from `.env` files

phillipleblanc:
## 🗣 Description

Improves the Env secret store to load `.env` files. By default it will look for `.env` or `.env.local` files in the working directory of the spice runtime. An optional `file_path` param can be specified to point to a custom `.env` file.

Manually load from `.env.production`
```yaml
secrets:
  - from: env
    name: env
    params:
      file_path: .env.production
```

Load the default `.env` and `.env.local` files.
```yaml
secrets:
  - from: env
    name: env
```

Also tweaks the secret loading logic to fallback to a Spice-prefixed key if the key that was specified didn't exist. This will be useful in the next PR when I add support for auto-loading required secrets if the user didn't specify it in their params. i.e. if we need a `pg_password` then we will automatically search for an environment variable called `PG_PASSWORD` or `SPICE_PG_PASSWORD`

## 🔨 Related Issues

Part of #1701

## 🤔 Concerns

By default the `dotenvy` crate will load the contents of the `.env` file into the process environment variables. This is usually the desired behavior, but that would mean if you defined multiple `env` secrets that each loaded a different `.env` file, then they would all be added to the global process namespace, and could be accessed from any of the `env` defined secrets. I think that is a fine tradeoff for now since that seems like an edge case, and would probably still result in the behavior the user wanted anyway.

### DIFF
--- a/crates/runtime/src/secrets/stores/env.rs
+++ b/crates/runtime/src/secrets/stores/env.rs
@@ -19,7 +19,7 @@ use secrecy::SecretString;
 
 use crate::secrets::SecretStore;
 
-const ENV_SECRET_PREFIX: &str = "SPICE_SECRET_";
+const ENV_SECRET_PREFIX: &str = "SPICE_";
 
 pub struct EnvSecretStore {}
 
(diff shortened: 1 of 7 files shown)


lukekim (approved):

crates/runtime/src/secrets/stores/env.rs
```diff
@@ -14,25 +14,73 @@ See the License for the specific language governing permissions and
...
+            };
+        }
+        if let Err(err) = dotenvy::from_filename(".env.local") {
+            if matches!(err, dotenvy::Error::LineParse(_, _)) {
+                tracing::warn!(".env.local: {err}");
+            }
+        };
+        if let Err(err) = dotenvy::from_filename(".env") {
+            if matches!(err, dotenvy::Error::LineParse(_, _)) {
+                tracing::warn!(".env: {err}");
```
digadeesh, line 81: I think we should be more verbose with these warnings.

ewgenius (approved):

crates/runtime/src/secrets/stores/env.rs
phillipleblanc: The error text from `dotenvy` is quite good for this error:

`Error parsing line: '{}', error at line index: {}`

PR #2008, opened 2024-07-18
feat: Add sql_dialect parameter for overriding ODBC dialect

peasee:
## 🗣 Description

* Adds a parameter, `sql_dialect`, to override the detected dialect for an ODBC connection
* Refactors driver detection to use an enum and `From` implementations for cleaner, more reusable code.

## 🔨 Related Issues

* Part of #1989

### DIFF
--- a/crates/runtime/src/dataconnector/odbc.rs
+++ b/crates/runtime/src/dataconnector/odbc.rs
@@ -21,6 +21,7 @@ use data_components::odbc::ODBCTableFactory;
 use data_components::Read;
 use datafusion::datasource::TableProvider;
 use datafusion::sql::unparser::dialect::{Dialect, MySqlDialect, PostgreSqlDialect, SqliteDialect};
+use datafusion_table_providers::sql::db_connection_pool::dbconnection::GenericError;
 use db_connection_pool::dbconnection::odbcconn::ODBCDbConnectionPool;
 use db_connection_pool::odbcpool::ODBCPool;
 use secrecy::ExposeSecret;
@@ -40,6 +41,8 @@ pub enum Error {
     },
     #[snafu(display("A required ODBC parameter is missing: {param}"))]
     MissingParameter { param: String },
+    #[snafu(display("An ODBC parameter is configured incorrectly: {param}. {msg}"))]
+    InvalidParameter { param: String, msg: String },
     #[snafu(display("No ODBC driver was specified in the connection string"))]
     NoDriverSpecified,
 }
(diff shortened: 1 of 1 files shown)


crates/runtime/src/dataconnector/odbc.rs
```diff
@@ -85,20 +109,33 @@ where
...
-                .to_lowercase();
-
-            let driver = driver
-                .split(';')
-                .find(|s| s.starts_with("driver="))
-                .context(NoDriverSpecifiedSnafu)?;
-
-            let dialect = match_driver_to_dialect(driver);
+            let dialect = if let Some(sql_dialect) = params.get("sql_dialect") {
+                let driver: ODBCDriver = ODBCDriver::from(sql_dialect.expose_secret().as_str());
```
phillipleblanc: This is overloading the sql dialect with the ODBC driver - it would be cleaner to have them separate and be able to convert from the sql dialect to the ODBC driver. i.e. I could imagine we want to support a "generic" dialect, but that wouldn't ever be a ODBC driver

y-f-u (approved):

crates/runtime/src/dataconnector/odbc.rs
```diff
@@ -86,7 +95,23 @@ impl From<ODBCDriver> for Option<Arc<dyn Dialect + Send + Sync>> {
             ODBCDriver::MySql => Some(Arc::new(MySqlDialect {})),
             ODBCDriver::PostgreSql => Some(Arc::new(PostgreSqlDialect {})),
             ODBCDriver::Sqlite => Some(Arc::new(SqliteDialect {})),
-            ODBCDriver::Unknown => None,
+            ODBCDriver::Unknown => Some(Arc::new(DefaultDialect {})),
+        }
+    }
+}
+
+impl TryFrom<SQLDialectParam> for Option<Arc<dyn Dialect + Send + Sync>> {
```
y-f-u, line 103: in the spiceai fork, the Dialect should be already Send + Sync?

peasee: it's because of the bounds included in the `with_dialect` function in `datafusion-table-providers`: https://github.com/datafusion-contrib/datafusion-table-providers/blob/ab9457dc8aec7ebc392e22efe4176f11b47d369a/src/sql/sql_provider_datafusion/mod.rs#L107

You're correct though, that `Dialect` has `Dialect: Send + Sync` so it should already implicitly be Send + Sync.

PR #2009, opened 2024-07-18
Add Postgres Data Connector TPCH Benchmark Tests

Sevenannn:
## 🗣 Description

* Refactor bench functions into traits, so that each data connector can reuse the logic for setting up benchmark tests.
* Add Postgres benchmark test

## 🔨 Related Issues

- https://github.com/spiceai/spiceai/issues/1993


.github/workflows/benchmarks.yml
```diff
@@ -1,28 +1,32 @@
...
+        run: |
+          mkdir -p ~/.spice
+          echo "[spiceai]" > ~/.spice/auth
+          echo "  [spiceai.params]" >> ~/.spice/auth
+          echo '    key = "${{ secrets.SPICE_SECRET_SPICEAI_BENCHMARK_KEY }}"' >> ~/.spice/auth
+
+      - run: cargo bench -p runtime
+        env:
+          UPLOAD_RESULTS_DATASET: 'spiceai.tests.oss_benchmarks'
+          PG_BENCHMARK_PG_HOST: ${{ secrets.SPICE_SECRET_PG_BENCHMARK_PG_HOST }}
```
sgrebnov: `PG_BENCHMARK_PG_HOST` - Please check if PG_ prefix is still required as there is ongoing work to simplify naming. We should align to new naming so we don't need to rework this

crates/runtime/benches/bench_spicecloud/mod.rs
```diff
@@ -12,44 +13,58 @@ pub(crate) async fn run(
         super::run_query_and_record_result(rt, benchmark_results, query_name, query).await?;
     }
 
+    // std::thread::sleep(Duration::from_secs(10));
```
sgrebnov:
```suggestion
```

crates/runtime/benches/bench_spicecloud/mod.rs
```diff
@@ -12,44 +13,58 @@ pub(crate) async fn run(
...
...\"tpch\\\".\\\"lineitem\\\".\\\"l_tax\\\", \\\"tpch\\\".\\\"lineitem\\\".\\\"l_returnflag\\\", \\\"tpch\\\".\\\"lineitem\\\".\\\"l_linestatus\\\", \\\"tpch\\\".\\\"lineitem\\\".\\\"l_shipdate\\\", \\\"tpch\\\".\\\"lineitem\\\".\\\"l_commitdate\\\", \\\"tpch\\\".\\\"lineitem\\\".\\\"l_receiptdate\\\", \\\"tpch\\\".\\\"lineitem\\\".\\\"l_shipinstruct\\\", \\\"tpch\\\".\\\"lineitem\\\".\\\"l_shipmode\\\", \\\"tpch\\\".\\\"lineitem\\\".\\\"l_comment\\\" FROM \\\"tpch\\\".\\\"lineitem\\\" WHERE ((\\\"tpch\\\".\\\"lineitem\\\".\\\"l_orderkey\\\" = \\\"tpch\\\".\\\"orders\\\".\\\"o_orderkey\\\") AND (\\\"tpch\\\".\\\"lineitem\\\".\\\"l_commitdate\\\" < \\\"tpch\\\".\\\"lineitem\\\".\\\"l_receiptdate\\\")))) GROUP BY \\\"tpch\\\".\\\"orders\\\".\\\"o_orderpriority\\\" ORDER BY \\\"tpch\\\".\\\"orders\\\".\\\"o_orderpriority\\\" ASC NULLS LAST\", details: [], metadata: MetadataMap { headers: {\"date\": \"Thu, 11 Jul 2024 10:46:10 GMT\", \"content-type\": \"application/grpc\", \"strict-transport-security\": \"max-age=15724800; includeSubDomains\", \"alt-svc\": \"h3=\\\":443\\\"; ma=86400\", \"cf-ray\": \"8a18354d2b5f5ac0-MEL\", \"cf-cache-status\": \"DYNAMIC\", \"server\": \"cloudflare\"} }"
```
sgrebnov, line 26: Other queries should be updated as well so you can just uncomment them

crates/runtime/benches/setup/mod.rs
```diff
@@ -132,3 +117,104 @@ fn get_branch_name() -> String {
...
+
+    fn make_rw_dataset(&self, path: &str, name: &str) -> Dataset {
+        let mut ds = Dataset::new(format!("spiceai:{path}"), name.to_string());
+        ds.mode = Mode::ReadWrite;
+        ds.replication = Some(Replication { enabled: true });
+        ds
+    }
+}
+
+struct PostgresBenchAppBuilder {}
```
sgrebnov: Better to move under `bench_postgres`

crates/runtime/benches/setup/mod.rs
```diff
@@ -132,3 +117,104 @@ fn get_branch_name() -> String {
...
+    }
+
+    fn make_dataset(&self, path: &str, name: &str) -> Dataset {
+        let mut dataset = Dataset::new(format!("postgres:{path}"), name.to_string());
+        dataset.params = Self::get_postgres_params();
+        dataset
+    }
+
+    fn make_rw_dataset(&self, path: &str, name: &str) -> Dataset {
+        let mut dataset = Dataset::new(format!("postgres:{path}"), name.to_string());
```
sgrebnov: I think we should consolidate all bench results in single place, we should not use data connector to store results IMO. Should be spiceai.

Sevenannn: Resolved in new commit

crates/runtime/benches/setup/mod.rs
Sevenannn: Fixed in new commit

crates/runtime/benches/bench_spicecloud/mod.rs
Sevenannn: Fixed in new commit

.github/workflows/benchmarks.yml
Sevenannn: Will confirm and update this after the PG_ prefix changes is included into trunk

Sevenannn: Confirmed that PG prefix is still needed in spice pod configuration, will keep it as it is

crates/runtime/benches/bench_postgres/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+        ("tpch_q12", include_str!("../tpch_queries/tpch_q12.sql")),
+        ("tpch_q13", include_str!("../tpch_queries/tpch_q13.sql")),
+        ("tpch_q14", include_str!("../tpch_queries/tpch_q14.sql")),
+        // tpch_q15 has a view creation which we don't support by design
+        ("tpch_q16", include_str!("../tpch_queries/tpch_q16.sql")),
+        ("tpch_q17", include_str!("../tpch_queries/tpch_q17.sql")),
+        ("tpch_q18", include_str!("../tpch_queries/tpch_q18.sql")),
+        ("tpch_q19", include_str!("../tpch_queries/tpch_q19.sql")),
+        // ("tpch_q20", include_str!("../tpch_queries/tpch_q20.sql")),
+        // ("tpch_q21", include_str!("../tpch_queries/tpch_q21.sql")),
```
phillipleblanc: Add a comment with the error for why this couldn't complete

crates/runtime/benches/bench_postgres/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+            "tpch_simple_q5",
+            include_str!("../tpch_queries/tpch_simple_q5.sql"),
+        ),
+    ]
+}
+
+pub struct PostgresBenchAppBuilder {}
+
+impl PostgresBenchAppBuilder {
+    fn get_postgres_params() -> Option<Params> {
```
phillipleblanc: Make this function return just a `Params` and wrap it in Some() from the caller

crates/runtime/benches/setup/mod.rs
```diff
@@ -128,3 +121,18 @@ fn get_branch_name() -> String {
...
         )
 }
+
+pub trait BenchAppBuilder {
+    fn build_app(&self, upload_results_dataset: &Option<String>) -> App;
+    fn make_dataset(&self, path: &str, name: &str) -> Dataset;
+
+    fn make_rw_dataset(&self, path: &str, name: &str, dataconnector: DataConnector) -> Dataset {
+        let mut ds = Dataset::new(
+            format!("spiceai:{}_{}", path, dataconnector),
```
phillipleblanc:
```suggestion
            format!("spiceai:{path}_{dataconnector}"),
```

crates/runtime/benches/setup/mod.rs
```diff
@@ -128,3 +121,18 @@ fn get_branch_name() -> String {
             |output| String::from_utf8_lossy(&output.stdout).trim().to_string(),
         )
 }
+
+pub trait BenchAppBuilder {
+    fn build_app(&self, upload_results_dataset: &Option<String>) -> App;
+    fn make_dataset(&self, path: &str, name: &str) -> Dataset;
+
+    fn make_rw_dataset(&self, path: &str, name: &str, dataconnector: DataConnector) -> Dataset {
```
phillipleblanc: Let's not have a separate table per data connector - we can add a new column to the benchmark results for the data connector.

crates/runtime/benches/bench_postgres/mod.rs
Sevenannn: Updated in the new commit

crates/runtime/benches/bench_postgres/mod.rs
Sevenannn: All queries are uncommented now

crates/runtime/benches/bench.rs
```diff
@@ -39,11 +39,14 @@ use crate::results::Status;
...
 mod setup;
 
+#[cfg(feature = "postgres")]
+mod bench_postgres;
 mod bench_s3;
 #[cfg(feature = "spark")]
 mod bench_spark;
 mod bench_spicecloud;
 
+#[allow(unreachable_patterns)]
```
sgrebnov: @Sevenannn - why we are adding `allow(unreachable_patterns)]` here?

Makefile
```diff
@@ -49,7 +49,7 @@ test-integration-without-spiceai-dataset:
 
 .PHONY: test-bench
 test-bench:
-	@cargo bench -p runtime
+	@cargo bench -p runtime --features postgres
```
sgrebnov: Propose to be consistent with spark and enable bench for non-default connectors in workflow action (postgres seems already enabled): https://github.com/spiceai/spiceai/blob/trunk/.github/workflows/benchmarks.yml#L26

sgrebnov (approved):

phillipleblanc (approved):

PR #2010, opened 2024-07-18
Switch order to look for `SPICE_` prefixed keys first

phillipleblanc:
## 🗣 Description

Switch the order that the secret stores look for keys - by default they will look for a `SPICE_` prefixed key, and then fallback to the explicit key. The rationale is that the `SPICE_` prefix is more specific, and thus should take precedence.

## 🔨 Related Issues

Part of #1701

### DIFF
--- a/crates/runtime/src/secrets/stores/aws_secrets_manager.rs
+++ b/crates/runtime/src/secrets/stores/aws_secrets_manager.rs
@@ -124,13 +124,13 @@ impl SecretStore for AwsSecretsManager {
             }
         };
 
+        let prefixed_key = format!("{SPICE_KEY_PREFIX}{key}");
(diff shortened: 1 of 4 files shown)


lukekim (approved):

peasee (approved):
Are `SPICE_KEY_PREFIX`, `KEYRING_SECRET_PREFIX` and `ENV_SECRET_PREFIX` all the same thing?

phillipleblanc:
> Are `SPICE_KEY_PREFIX`, `KEYRING_SECRET_PREFIX` and `ENV_SECRET_PREFIX` all the same thing?

They are logically the same - but I kept them separate so they could have secret store specific differences, i.e. `SPICE_` for `ENV_SECRET_PREFIX` and `spice_` for the others.

PR #2013, opened 2024-07-18
Autoload secrets that Data Connectors have indicated are needed

phillipleblanc:
## 🗣 Description

Data Connectors now need to specify which secrets it might need as part of the `autoload_secrets()` function on the `DataConnectorFactory`. This list is used to try to autoload those secrets if they exist from the defined secret stores.

Each Data Connector now also needs to specify a `prefix` that is used to namespace the parameters and secrets it needs. Any parameter which does not start with that prefix is filtered out (with a warning log) and not sent to the data connector.

This PR also depends on changes in https://github.com/datafusion-contrib/datafusion-table-providers/pull/16

## 🔨 Related Issues

Part of #1701

## 🤔 Concerns

This PR does not yet handle embedding or models, as they don't follow the same model as data connectors. I think we'll need to have them follow the same model of a registry with a factory so we can enable the same semantics.

### DIFF
--- a/crates/runtime/src/dataconnector.rs
+++ b/crates/runtime/src/dataconnector.rs
@@ -17,6 +17,7 @@ limitations under the License.
 use crate::component::catalog::Catalog;
 use crate::component::dataset::acceleration::RefreshMode;
 use crate::component::dataset::Dataset;
+use crate::secrets::Secrets;
 use crate::Runtime;
 use arrow::datatypes::SchemaRef;
 use async_trait::async_trait;
@@ -49,7 +50,7 @@ use std::fmt::Display;
 use std::pin::Pin;
 use std::str::FromStr;
 use std::sync::Arc;
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, RwLock};
 use url::Url;
 
 use std::future::Future;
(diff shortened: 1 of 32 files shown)


crates/runtime/src/dataconnector.rs
```diff
@@ -247,61 +239,108 @@ pub async fn register_connector_factory(
...
     params: HashMap<String, SecretString>,
+    secrets: Arc<RwLock<Secrets>>,
 ) -> Option<AnyErrorResult<Arc<dyn DataConnector>>> {
     let guard = DATA_CONNECTOR_FACTORY_REGISTRY.lock().await;
 
     let connector_factory = guard.get(name);
 
     match connector_factory {
-        Some(factory) => Some(factory(params).await),
+        Some(factory) => {
```
phillipleblanc, line 249: This is the main change to filter out the prefixed parameters and autoload the secrets

crates/runtime/src/dataconnector.rs
```diff
@@ -247,61 +239,108 @@ pub async fn register_connector_factory(
...
+            let result = factory.create(params).await;
+            Some(result)
+        }
         None => None,
     }
 }
 
 pub async fn register_all() {
-    register_connector_factory("localhost", localhost::LocalhostConnector::create).await;
+    register_connector_factory("localhost", localhost::LocalhostConnectorFactory::new_arc()).await;
```
phillipleblanc, line 277: Instead of registering a function that returns a data connector, we now register an `Arc<dyn DataConnectorFactory>` that has methods to create a data connector, but also get the prefix and autoload secrets

lukekim (approved):

PR #2016, opened 2024-07-18
Update ROADMAP.md to include MySQL data connector in Beta

digadeesh:
## 🗣 Description

Add MySQL data connector as Beta, and then Stable at v1.0.


peasee (commented):
Should we include MySQL in the dot-points for RC and Stable as well?

lukekim (approved):

PR #2025, opened 2024-07-19
[WIP] /v1/chat/completions with data tools

ewgenius:


crates/runtime/src/http/v1/chat.rs
```diff
@@ -27,21 +32,145 @@ use axum::{
...
     Extension(llms): Extension<Arc<RwLock<LLMModelStore>>>,
+    Extension(df): Extension<Arc<DataFusion>>,
     Json(req): Json<CreateChatCompletionRequest>,
 ) -> Response {
     let model_id = req.model.clone();
+    let mut req = req.clone();
+
+    if let Ok(system_message) = ChatCompletionRequestSystemMessageArgs::default()
+        .content(
+            r"You are an AI assistant built on top of the Spice runtime, which provides a unified SQL interface to materialize, accelerate, and query data from any database, data warehouse, or data lake. Your primary functionalities include:
```
phillipleblanc, line 53: We shouldn't hard-code the system message in the API code

PR #2028, opened 2024-07-19
Separate data connector parameters into `connector` and `runtime` categories

phillipleblanc:
## 🗣 Description

All data connector parameters are now required to be declared up front in the `parameters()` trait function on the `DataConnectorFactory`. Attempting to load a parameter from the newly created `Parameters` object which wasn't declared will result in a panic. The `Parameters` object also now can provide the user-friendly parameter name for use in error messages. The `ParameterSpec` object that declares the parameters can set whether its a secret that should be autoloaded, a default value, whether its required, a description, a help link and some examples. Not all of these are being at the moment, but the idea is we can start generating good help documentation when the user misconfigures something.

If the user specifies a parameter that isn't one of the parameters we've declared it will result in a warning that it is being ignored.

Overall I'm pretty happy with where this PR landed and it will make updating docs much easier in the future. It also puts Data Connector authors on tighter rails that guide them to the pit of success.

## 🔨 Related Issues

Part of #1701


digadeesh (dismissed):

Sevenannn (dismissed):
lgtm! I was wondering that after this change, if we are adding any new data connectors / new parameters to existing data connectors, do we always expose the parameter values only when we use them?

phillipleblanc:
> lgtm! I was wondering that after this change, if we are adding any new data connectors / new parameters to existing data connectors, do we always expose the parameter values only when we use them?

Yeah the idea is to keep it as a SecretString for as long as possible and expose it when it's needed as a string.

digadeesh (dismissed):

lukekim (dismissed):

lukekim (approved):

lukekim (approved):

PR #2041, opened 2024-07-22
enable spark connect federated query

y-f-u:
## 🗣 Description

This adds federated query capability into spark_connect connector.

## 🔨 Related Issues

n/a

## 🤔 Concerns

n/a

### DIFF
--- a/crates/data_components/src/spark_connect/federation.rs
+++ b/crates/data_components/src/spark_connect/federation.rs
@@ -1,0 +1,137 @@
+/*
+Copyright 2024 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


crates/data_components/src/spark_connect/federation.rs
```diff
@@ -0,0 +1,121 @@
...
+    query: String,
+) -> impl Stream<Item = DataFusionResult<RecordBatch>> {
+    let session = Arc::clone(&session);
+
+    stream! {
+        let data = session
+            .sql(&query)
+            .await
+            .map_err(map_error_to_datafusion_err)?
+            .collect()
```
y-f-u, line 129: This uses the existing way to collect data which is not streaming yet, will address this in future.

phillipleblanc (approved):

PR #2070, opened 2024-07-23
Workflow to build MySQL image with TPCH benchmark data

sgrebnov:
## 🗣 Description

Create workflow to build custom MySQL image with TPCH benchmark data. 
- Triggered manually.
- Image is saved to repo Container Registry

To be used for MySQL benchmark testing. This can also be used to simplify integration tests logic that [manually downloads TPCH data/parquet and provision test table](https://github.com/spiceai/spiceai/blob/32fe9e31aaee41d9ed8bdad7921f82057a3c6063/crates/runtime/tests/refresh_retry/mysql.rs#L56)

##  Concerns
2. For GH Actions testing purpose we can have GH Action to provision MySQL instance on the fly instead of having Docker image (similar to `Initialize TPCH data` step).
1. Instead of single Docker image we can produce separate images with different TPCH scale factor. Current approach was selected so we can mix different scale factors and use it for local development and testing


Sevenannn (approved):

Sevenannn:
lgtm. I think this approach makes sense in the cost & time saving perspective, compared to spinning up the database in gh action on the fly everytime the test is run. We might need a brief doc on running docker container locally using image from registry if using this approach

PR #2095, opened 2024-07-24
Add cfg flag to properly support build w / wo feature in benchmark tests

Sevenannn:
## 🗣 Description

Add cfg flags so the bench test can successfully run with `cargo bench -p runtime --features postgres` or `cargo bench -p runtime --features spark`

### DIFF
--- a/crates/runtime/benches/setup/mod.rs
+++ b/crates/runtime/benches/setup/mod.rs
@@ -14,7 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use crate::{bench_mysql, bench_postgres, bench_s3, results::BenchmarkResultsBuilder};
+#[cfg(feature = "mysql")]
+use crate::bench_mysql;
+#[cfg(feature = "postgres")]
+use crate::bench_postgres;
+use crate::{bench_s3, results::BenchmarkResultsBuilder};
 use app::{App, AppBuilder};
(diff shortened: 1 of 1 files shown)


crates/runtime/benches/setup/mod.rs
```diff
@@ -80,7 +84,9 @@ fn build_app(upload_results_dataset: &Option<String>, connector: &str) -> App {
             .with_dataset(make_spark_dataset("samples.tpch.region", "region"))
             .with_dataset(make_spark_dataset("samples.tpch.supplier", "supplier")),
         "s3" => bench_s3::build_app(app_builder),
+        #[cfg(feature = "postgres")]
```
digadeesh, line 87: should it be added for spark, s3, spice.ai, etc?

sgrebnov (approved):

crates/runtime/benches/setup/mod.rs
sgrebnov: I think this is mysql and postgres only as they use `build_app` from feature protected files (below). spark is feature protected by defines datasets creation directly in this file so works fine.
```
#[cfg(feature = "mysql")]
mod bench_mysql;
```

Sevenannn: Yes, Sergei explains what's happened here

PR #2101, opened 2024-07-24
Add workflow to build Postgres Docker image using tpch data

Sevenannn:
## 🗣 Description

Create workflow to build custom Postgres image with TPCH benchmark data.

- Triggered manually.
- Image is saved to repo Container Registry

The image will be used for Postgres benchmark testing, which will mitigate the network latency / extra cost of using AWS rds for tests.


.github/workflows/spice_postgres_bench_docker.yml
```diff
@@ -0,0 +1,62 @@
+name: spice-postgres-bench docker
+
+on:
+  push:
+    branches:
+      - qianqian/docker-tpch-postgres
+  pull_request:
+    branches:
+      - qianqian/docker-tpch-postgres
```
phillipleblanc:
```suggestion
      - trunk
  pull_request:
    branches:
      - trunk
```

.github/workflows/spice_postgres_bench_docker.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Initialize TPCH data
+        working-directory: ./test/tpch
+        run: |
+          make tpch-init
+          export PGPASSWORD='postgres'
+          for sf in 0.01 0.05 1; do
+            DB_NAME=tpch_sf${sf//./_}
+            DBGEN_SCALE=$sf make tpch-gen
+            DB_HOST=127.0.0.1 DB_PORT=5432 DB_USER=postgres DB_NAME=$DB_NAME make pg-init
```
phillipleblanc, line 36: This is starting a Postgres server, generating the TPCH data and then dumping the TPCH data from Postgres. Is there a way to just generate the SQL directly without having to stage it in Postgres first?

test/tpch/postgres-bench/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM postgres:latest
+
+COPY postgres-entrypoint.sh /usr/local/bin/postgres-entrypoint.sh
+RUN chmod +x /usr/local/bin/postgres-entrypoint.sh
+
+COPY backup_*.sql /sql_files/
+
+ENTRYPOINT ["/usr/local/bin/postgres-entrypoint.sh"]
```
phillipleblanc, line 8: Instead of doing the work to import the TPCH files everytime we start the postgres server, can't we bake in an image that already has the databases loaded so we just pull it down and its ready to go?

lukekim (changes requested):
Need a higher standard of PR description.

.github/workflows/spice_postgres_bench_docker.yml
Sevenannn: To create a Postgres Docker image that has pre-populated database would require a SQL dump that contains all the commands of populating data. To get these SQL of populating data would require a staging database with popultaed data, and use pg_dump over the staged database to get SQL dump.

test/tpch/postgres-bench/Dockerfile
Sevenannn: Nope, the only way to get populated data in an image is to have sql dump with base image. That's why Postgres and MySQL are both using this approach.

.github/workflows/spice_postgres_bench_docker.yml
Sevenannn: Will change to workflow dispatch, this workflow is not supposed to run once we confirm the image contains tpch data we want

lukekim (approved):

Issue #2071, opened 2024-07-23
Enhancement: Support HTTPS and GRPC+TLS on all Spice endpoints


phillipleblanc:
## Goal-State/What/Result

All Spice endpoints can be secured with encryption in-transit via TLS encryption. This will enable HTTPS on the `8090` HTTP endpoint, the `9090` metrics endpoint, and GRPC+TLS on the two GRPC ports, `50051` for Flight and `50052` for OpenTelemetry.

## Why/Purpose

Required for security and compliance of most companies. Also follows our secure by default principle.

## By When

**Issue/Spec written and reviewed:** Tue, July 23, 2024
**Done-Done:** Fri, July 26, 2024

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

Spice needs a `certificate` and `key` which correspond to the certificate and private key used to encrypt the TLS connections. This can be provided via command line arguments, i.e.:

Specify with file references:
`spiced --tls-certificate-file ./cert.pem --tls-key-file ./key.pem`

Or inline with:
```bash
TLS_CERT=$(cat /path/to/cert.pem)
TLS_KEY=$(cat /path/to/key.pem)
spiced --tls-certificate $TLS_CERT --tls-key $TLS_KEY
```

Or via runtime parameters in the Spicepod. Specify file references with:

```yaml
runtime:
  tls:
    certificate_file: ./cert.pem
    key_file: ./key.pem
```

Or reference the key inline:

```yaml
runtime:
  tls:
    certificate: |
      -----BEGIN CERTIFICATE-----
      ...
      -----END CERTIFICATE-----
    key: |
      -----BEGIN PRIVATE KEY-----
      ...
      -----END PRIVATE KEY-----
```

Secrets replacement will also work, enabling:
```yaml
runtime:
  tls:
    certificate: ${secrets:tls_cert}
    key: ${secrets:tls_key}
```

[...]

PR #2108, opened 2024-07-25
Enable TLS for metrics endpoint

phillipleblanc:
## 🗣 Description

Enables the metrics endpoint (`https://localhost:9090`) to work over HTTPS via a new set of CLI args to `spiced`:

With certificates on the filesystem
`spiced --metrics 127.0.0.1:9090 --tls --tls-certificate-path /path/to/cert.pem --tls-key-path /path/to/key.pem`

With certificates specified inline
```bash
export TLS_CERT=$(cat path/to/cert.pem)
export TLS_KEY=$(cat path/to/key.pem)
spiced --metrics 127.0.0.1:9090 --tls --tls-certificate $TLS_CERT --tls-key $TLS_KEY
```

This PR only enables TLS for the metrics endpoint, the other endpoints will be follow up PRs. Enabling TLS via a Spicepod runtime configuration will be a follow-on PR as well.

## 🔨 Related Issues

Part of #2071


sgrebnov (approved):

crates/runtime/src/embeddings/array_distance.rs
```diff
@@ -51,9 +48,10 @@ impl Default for ArrayDistance {
 /// For two [`DataType::FixedSizeList`], the inputs must have the same length, and have compatible
 /// inner types. Compatible inner types are
 ///   - Both inputs are [`DataType::Float16`], [`DataType::Float32`], or [`DataType::Float64`].
```
Sevenannn, line 50: Float16 should be removed from comment if we don't support it in these functions anymore

Sevenannn (approved):

PR #2109, opened 2024-07-25
Enable TLS for HTTP endpoint

phillipleblanc:
## 🗣 Description

Changes the HTTP port to be secured with TLS when configured. `axum::serve` does not support TLS and the official recommendation is to use hyper directly. There is a community crate [`axum_server`](https://crates.io/crates/axum-server) that easily supports TLS serving with axum, and I initially used that for the implementation - but it was behind on the rustls version and looking closely, it seems to not be as well maintained. That meant we would have needed to fork it to update the version and maintain that fork.

However, in the end the approach I took in #2108 ended up being reusable here. The only tricky part was figuring how how to translate an axum (Tower) service into a `hyper` service. Fortunately the `hyper_util` crate had a handy conversion.

## 🔨 Related Issues

Part of #2071


bin/spiced/src/lib.rs
```diff
@@ -140,3 +176,46 @@ pub async fn run(args: Args, metrics_handle: Option<PrometheusHandle>) -> Result
...
+
+    let cert_reader: &mut dyn io::BufRead =
+        match (&args.tls_certificate_file, &args.tls_certificate) {
+            (Some(cert_path), _) => &mut load_file(Path::new(cert_path))?,
+            (_, Some(cert)) => &mut cert.as_bytes(),
+            (None, None) => return Err("TLS certificate is required (--tls-certificate)".into()),
+        };
+    let key_reader: &mut dyn io::BufRead = match (&args.tls_key_file, &args.tls_key) {
+        (Some(key_path), _) => &mut load_file(Path::new(key_path))?,
+        (_, Some(key)) => &mut key.as_bytes(),
```
peasee: **question:** What loading precedence occurs here, does it follow the position of the match arm? e.g. file ahead of CLI arg, if a user supplies both arguments?

crates/runtime/src/metrics_server/mod.rs
```diff
@@ -64,31 +68,58 @@ where
         let stream = match listener.accept().await {
             Ok((stream, _)) => stream,
             Err(e) => {
-                tracing::warn!(
+                tracing::debug!(
```
peasee: **question:** Should this log be `debug!`?

peasee (dismissed):
🚢

bin/spiced/src/lib.rs
phillipleblanc: Yeah it tries each match arm in order

crates/runtime/src/metrics_server/mod.rs
phillipleblanc: I don't think so. It was fairly noisy when I tried connecting incorrectly (I.e HTTP instead of HTTPS). Axum serve didn't report them so I opted to hide it behind a debug

peasee (approved):

digadeesh (approved):

PR #2118, opened 2024-07-26
Support configuring TLS in the spicepod

phillipleblanc:
## 🗣 Description

Support configuring TLS via the Spicepod runtime configuration. This is lower precedence than specifying on the command line.

```yaml
runtime:
  tls:
    certificate: |
      -----BEGIN CERTIFICATE-----
     ...
      -----END CERTIFICATE-----
    key_file: ./certs/spiced_key.pem
```

Secrets replacement also works:

```yaml
runtime:
  tls:
    certificate: ${env:TLS_CERT}
    key: ${env:TLS_KEY}
```

I've moved the secrets loading from `load_components` to load as part of building the Runtime object. Now that both the server startup and component loading need secrets, it makes sense to just have it always available in the Runtime.

I also moved the code for loading the TLS configuration to its own module in `spiced`.

## 🔨 Related Issues

Part of #2071

### DIFF
--- a/bin/spiced/src/tls.rs
+++ b/bin/spiced/src/tls.rs
@@ -1,0 +1,119 @@
+/*
+Copyright 2024 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use std::fs::File;
+use std::io::{self, Read};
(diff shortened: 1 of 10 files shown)


bin/spiced/src/tls.rs
```diff
@@ -0,0 +1,119 @@
...
+    let key_bytes: Vec<u8> = match (&args.tls_key_file, &args.tls_key, app_key_bytes) {
+        (Some(key_path), _, _) => load_file(Path::new(key_path))?,
+        (_, Some(key), _) => key.as_bytes().to_vec(),
+        (_, _, Some(key)) => key,
+        (None, None, None) => return Err("TLS key is required (--tls-key)".into()),
+    };
+
+    let tls_config = TlsConfig::try_new(cert_bytes, key_bytes)?;
+
+    tracing::info!("All endpoints secured with TLS");
```
y-f-u: nit. Probably the trace should be in start_servers?

y-f-u (dismissed):

bin/spiced/src/tls.rs
phillipleblanc: Moved

digadeesh (approved):

PR #2120, opened 2024-07-26
Add delta bench

y-f-u:
## 🗣 Description

- add delta bench
- reorg the benchmarks job a bit so it can run selected features.


peasee (dismissed):

crates/runtime/Cargo.toml
```diff
@@ -155,6 +155,8 @@ sqlite = [
   "db_connection_pool/sqlite",
   "data_components/sqlite",
 ]
+spiceai-bench = []
+s3-bench = []
```
phillipleblanc: It feels weird to have this as a feature on the main runtime

y-f-u: It probably can be better using prefix pattern like bench-spiceai, bench-s3 with some comments to indicate these features are only used internally for bench/test.
More idiomatic way could be using multiple bench target to associate features and test entry point together which could be preferred, however, this makes it tricky for display_bench to include all bench tests.

phillipleblanc (approved):

PR #2135, opened 2024-07-29
Use forked dotenvy to disable variable substitution

### LINKED ISSUE #2098: Secrets with special characters are not loading from file
**Describe the bug**
Secrets appear to be failing when loading from a file.  Secrets have "$#[$~?>" characters in it.  It seems one is not being escaped properly in file based .env secrets.  

**To Reproduce**

```yaml
version: v1beta1
kind: Spicepod
name: postgres-demo


secrets:
  - from: env
    name: env
    params:
      file_path: .env

datasets:
  - from: postgres:customer
    name: customer
    acceleration:
      enabled: true
    params:
      pg_host: <redacted host>
      pg_user: postgres
      pg_db: tpch
      pg_pass: ${env:PG_PASS}
      pg_sslmode: require
```

Use the following spicepod yaml as guidance, host name and secret were removed, but use the special characters in the .env file in the secrets.

**Expected behavior**
secrets load and can be used in the connection to the dataset.

**spice, spiced, OS info**
- OS info: `uname -a`

yes

**Additional context**
The postgres connector doesn't complain about the secrets, but instead says authentication failed.  The same secrets can be loaded in environment variables and the dataset loads successfully.

phillipleblanc:
## 🗣 Description

Brings in https://github.com/spiceai/dotenvy/pull/1 to fix #2098

## 🔨 Related Issues

Closes #2098

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3406,8 +3406,7 @@ checksum = "fea41bba32d969b513997752735605054bc0dfa92b4c56bf1189f2e174be7a10"
 [[package]]
 name = "dotenvy"
 version = "0.15.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1aaf95b3e5c8f23aa320147307562d361db0ae0d51242340f558153b4eb2439b"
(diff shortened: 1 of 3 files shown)


Jeadie (approved):
Didn't know about `[patch.crates-io]`. Could be very useful for models work.

PR #2138, opened 2024-07-29
Avoid more unnecessary allocations in the query path

### LINKED ISSUE #1534: Code review all critical query code paths to ensure highest possible query performance
## Goal-State/What/Result

All critical query code paths are code reviewed for performance concerns and any issues resolved.

## Why/Purpose

Query performance is critical to the core value proposition of Spice.

## By When

For beta.

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Tests
- [x] Documentation
- [x] Release Notes

## The Algorithm

- [x] Simplify.

## Spec

Basic specification of the enhancement.

## How/MAP

- [x] Review the call tree structure.
- [x] Review uses of arg passing (ref/val), stack vs heap, etc. 
- [x] Review uses of `.clone()` in the critical path.
- [x] Review all loops and recursion.

### Implementation (issues)

- 
-

phillipleblanc:
## 🗣 Description

Another pass on the query path to remove unnecessary allocations and improve performance. The major win from this PR is avoiding the clone of the `LogicalPlan`, which is potentially expensive. One of the only reasons we cloned the `LogicalPlan` previously was so we could pass it to the hash function to get a `u64`. We can just calculate that hash up front and avoid an expensive clone.

I reworked the input tables scanner to work directly with `TableReference`s instead of `String`. This also avoid some more allocations, and it was what we had originally - which meant another unnecessary allocation to convert from TableReference -> String. The only place where we "needed" a string was in an error message, which I changed to use the `table` name from the TableReference directly, which is a cheap clone of a pointer rather than a clone of a string.

I also noticed that the `cache` crate had implicit knowledge about which schemas we didn't want to cache (`runtime` and `information_schema`) - that is specific to the Spice runtime, and thus it doesn't belong in that generic cache crate. I've reworked the constructor to take a reference to a slice of `str` to use when filtering.

## 🔨 Related Issues

Closes #1534

### DIFF
--- a/crates/runtime/src/datafusion/query.rs
+++ b/crates/runtime/src/datafusion/query.rs
@@ -18,10 +18,7 @@ use std::{cell::LazyCell, sync::Arc};
 
 use arrow::datatypes::{Schema, SchemaRef};
 use arrow_tools::schema::verify_schema;
-use cache::{
-    cache_is_enabled_for_plan, get_logical_plan_input_tables, to_cached_record_batch_stream,
-    QueryResult,
-};
+use cache::{get_logical_plan_input_tables, to_cached_record_batch_stream, QueryResult};
 use datafusion::{
     error::DataFusionError,
     execution::{context::SQLOptions, SendableRecordBatchStream},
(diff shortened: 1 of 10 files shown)


crates/runtime/src/datafusion/query.rs
```diff
@@ -167,8 +168,6 @@ impl Query {
 
         tracker = tracker.datasets(Arc::new(get_logical_plan_input_tables(&plan)));
 
-        let plan_copy = plan.clone();
```
phillipleblanc, line 170: The main win from this PR is avoiding this potentially expensive clone in the query hot path

lukekim (approved):

PR #2156, opened 2024-07-31
fix hanging spiced when odbc loading data and received a cancel signal

### LINKED ISSUE #2083: Bug: Runtime does not abort (Ctrl+C) while accelerating dataset
**Describe the bug**

While accelerating datasets, if you attempt to exit the runtime with a quit command (Ctrl+C) the runtime will not immediately exit.

The runtime will print `Goodbye!`, but will continue running until acceleration of the dataset that was in-progress completes before actually exiting.

This also happens when registering the federated datasets.

**To Reproduce**

1. Configure a large dataset with acceleration that takes some time to accelerate
2. Start the runtime
3. Try to quit the runtime (Ctrl+C)
4. Observe the runtime will not quit

**Expected behavior**

The runtime should correctly shutdown on kill commands, and quit regardless of in-progress operations.

Yes

y-f-u:
## 🗣 Description

`spawn_blocking` will prevent the tokio runtime to shutdown as
> When you shut down the executor, it will wait indefinitely for all blocking operations to finish.

Alternatively, a force shutdown can be done however that will potential leave things in a bad state. Opt for using `CancellationToken` from tokio to coordinate between a dedicated ctrl-c listening task and the data fetch task.

## 🔨 Related Issues

Fixes #2083

### DIFF
--- a/crates/db_connection_pool/src/dbconnection/odbcconn.rs
+++ b/crates/db_connection_pool/src/dbconnection/odbcconn.rs
@@ -48,6 +48,7 @@ use snafu::Snafu;
 use tokio::runtime::Handle;
 
 use odbc_api::Connection;
+use tokio::signal;
 use tokio::sync::mpsc::Sender;
 
 type Result<T, E = GenericError> = std::result::Result<T, E>;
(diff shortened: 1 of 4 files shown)


crates/db_connection_pool/src/dbconnection/odbcconn.rs
```diff
@@ -173,6 +174,26 @@ where
...
+
+        tokio::spawn(async move {
+            let ctrl_c = async {
+                let signal_result = signal::ctrl_c().await;
+                if let Err(err) = signal_result {
+                    tracing::error!(
+                        "Failed to listen to shutdown signal in ODBC data fetching: {err}"
+                    );
+                } else {
+                    cancellation_token.clone().cancel();
```
phillipleblanc: Move this cancel outside of the else - even if listening fails, we still want to cancel

crates/db_connection_pool/src/dbconnection/odbcconn.rs
```diff
@@ -173,6 +174,26 @@ where
         let params = params.iter().map(dyn_clone::clone).collect::<Vec<_>>();
         let secrets = Arc::clone(&self.params);
 
+        let cancellation_token = tokio_util::sync::CancellationToken::new();
+        let cloned_token = cancellation_token.clone();
+
+        tokio::spawn(async move {
+            let ctrl_c = async {
+                let signal_result = signal::ctrl_c().await;
```
phillipleblanc: Does this also handle SIGTERM? That is the signal K8s sends for graceful shutdown

y-f-u: it seems spiced doesn't handle sigterm gracefully now. Created a separate issue to track this https://github.com/spiceai/spiceai/issues/2159

phillipleblanc (dismissed):

sgrebnov (approved):

PR #2161, opened 2024-07-31
Benchmark: data connector tests should continue on query error (s3)

sgrebnov:
## 🗣 Description

PR updates `run_query_and_record_result` to save failed query result instead of failing and allow data connector to continue and complete all test queries. This is useful to identify all failed queries for a data connector when some of the queries fail.

The change currently applied to S3 data connector and will be extended to other data connectors if current PR is applied.

### DIFF
--- a/crates/runtime/benches/bench.rs
+++ b/crates/runtime/benches/bench.rs
@@ -187,17 +187,16 @@ async fn run_query_and_record_result(
     let mut min_iter_duration_ms = i64::MAX;
     let mut max_iter_duration_ms = i64::MIN;
 
+    let mut query_err: Option<String> = None;
+
+    let mut completed_iterations = 0;
+
     for _ in 0..benchmark_results.iterations() {
+        completed_iterations += 1;
+
         let start_iter_time = get_current_unix_ms();
-        let _ = rt
-            .datafusion()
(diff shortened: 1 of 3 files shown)


phillipleblanc (approved):

Sevenannn (approved):

crates/runtime/benches/bench.rs
```diff
@@ -187,17 +187,16 @@ async fn run_query_and_record_result(
     let mut min_iter_duration_ms = i64::MAX;
     let mut max_iter_duration_ms = i64::MIN;
 
+    let mut query_err: Option<String> = None;
+
+    let mut completed_iterations = 0;
+
     for _ in 0..benchmark_results.iterations() {
+        completed_iterations += 1;
```
Sevenannn, line 195: It would be better to put `completed_iterations += 1` after `if let Err(e) = res {
            query_err = Some(e);
            break;
        }`. If a query failed in the first iteration and the loop breaks, we probably don't want to mark it to have 1 completed iteration

sgrebnov: Discussed with @Sevenannn / will keep existing implementation as failed query is actually a completed attempt/test.

Issue #1960, opened 2024-07-13
Enhancement: Spice AI Data Connector Append Stream Support


Sevenannn:
REMEMBER, BE **SMART**!

S: Specific
M: Measurable
A: Achievable
R: Relevant
T: Time-Bound

## Goal-State/What/Result

Support the append refresh mode without time_column using the append data stream in Spice AI Data Connector.

## Why/Purpose

The spice ai data connector append refresh mode without time_column is not subscribing to the spiced and fetching new data into accelerator and need to be fixed.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** 2024/7/10
**Done-Done:** TBD

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

* Subscribing to the change data stream using the DoExchange method in the Spice AI data connector. 
* Route dataset with append refresh mode and without time_column to use the acceleration logic of change refresh mode.
* Configure server DoExchange handler to send a ChangeEnvelope format data back to client.

## How/MAP

A clear and concise plan of how this should be implemented.

### Implementation (issues)

- https://github.com/spiceai/spiceai/pull/1935
- https://github.com/spiceai/spiceai/pull/1958
- https://github.com/spiceai/spiceai/issues/1961

digadeesh:
@Sevenannn and @y-f-u  to sync on the append logic.

phillipleblanc:
Goal: Done-Done by Wednesday.

PR #2167, opened 2024-08-01
Localhost -> Sink; Fix Sink connector to not require schema via `CREATE TABLE...` and infer on first write

phillipleblanc:
## 🗣 Description

Renames the undocumented `localhost` data connector to `sink` to more accurately reflect its job. Change the behavior of the `sink` connector to not require a schema to be created via `CREATE TABLE...` and instead infer the schema on the first write. This requires keeping a list of the sink datasets that have been registered but are still waiting for the first write to infer the schema.

Once we get a write for that dataset, then register the table with the schema. This change also adds a `RefreshMode::Disabled` to the sink connector and sets it as the default refresh mode.

This significantly improves the UX of the sink connector and fixes it to its previous behavior before the major rewrite about 3 months ago.

This is the main issue blocking #1960, as that functionality depends on this connector working correctly - and setting the schema via CREATE TABLE statements was a hack.

## 🔨 Related Issues

Part of #1960

## 🤔 Concerns

The sink datasets will not be queryable via the SQL interface until the first write. We could register it with the placeholder schema, but that feels worse than just waiting to register it.

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task.rs
@@ -493,7 +493,7 @@ impl RefreshTask {
                 refresh.sql.clone(),
                 match refresh.mode {
                     RefreshMode::Disabled => {
-                        unreachable!("refresh is not be called for disabled acceleration")
+                        unreachable!("Refresh cannot be called when acceleration is disabled")
                     }
                     RefreshMode::Full => UpdateType::Overwrite,
                     RefreshMode::Append => UpdateType::Append,
--- a/crates/runtime/src/accelerated_table/refresh_task.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task.rs
@@ -203,7 +203,7 @@ impl RefreshTask {
 
         let get_data_update_result = match mode {
             RefreshMode::Disabled => {
-                unreachable!("refresh is not be called for disabled acceleration")
+                unreachable!("Refresh cannot be called when acceleration is disabled")
             }
             RefreshMode::Full => self.get_full_update().await,
             RefreshMode::Append => self.get_incremental_append_update().await,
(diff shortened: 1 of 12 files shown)


crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -189,6 +189,9 @@ impl RefreshTask {
 
         let _timer = TimeMeasurement::new(
             match mode {
+                RefreshMode::Disabled => {
+                    unreachable!("refresh is not be called for disabled acceleration")
```
lukekim:
```suggestion
                    unreachable!("Refresh cannot be called when acceleration is disabled")
```

lukekim (dismissed):

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -199,6 +202,9 @@ impl RefreshTask {
         let start_time = SystemTime::now();
 
         let get_data_update_result = match mode {
+            RefreshMode::Disabled => {
+                unreachable!("refresh is not be called for disabled acceleration")
```
peasee:
```suggestion
                unreachable!("Refresh cannot be called when acceleration is disabled")
```

**nit:** Copies the error suggestion from above.

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -486,6 +492,9 @@ impl RefreshTask {
             (
                 refresh.sql.clone(),
                 match refresh.mode {
+                    RefreshMode::Disabled => {
+                        unreachable!("refresh is not be called for disabled acceleration")
```
peasee:
```suggestion
                        unreachable!("Refresh cannot be called when acceleration is disabled")
```

**nit:** Copies the error suggestion from above.

peasee (dismissed):

peasee (approved):

PR #2176, opened 2024-08-02
Remove embedding columns from `retrieved_primary_keys` in v1/search

Jeadie:
## 🗣 Description
 - For a `tbl` with primary keys `a, b, c` and an embedded column `e`, the SQL used to retrieve 3 rows with embeddings similar to `[1.0, 2.0, 3.0]`
 ```sql
SELECT a, b, c, e FROM tbl ORDER BY array_distance(e_embedding, [1.0, 2.0, 3.0]) LIMIT 3
 ```
- For `v1/search`, the response payload looks like 
```json
{
    "entries": {
        "tbl": [ "foo", "bar", "fizzzz"] 
    },
    "retrieved_public_keys":  {
         "tbl": [
            {"a": "pk_value_1a", "b": "pk_value_2a", "c": "pk_value_3a", "e": "foo"},
            {"a": "pk_value_1b", "b": "pk_value_2b", "c": "pk_value_3b", "e": "bar"},
            {"a": "pk_value_1c", "b": "pk_value_2c", "c": "pk_value_3c", "e": "fizzzz"}
        ] 
     }
}
```
- The above response is incorrect. In `entries.retrieved_public_keys.[*]`, we should not return embedded columns, in this example `"e"` (unless `"e"` is also apart of the primary key). 
- This PR fixes it.


crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -115,8 +115,9 @@ impl VectorSearch {
 
         for (tbl, search_vectors) in per_table_embeddings {
             tracing::debug!("Running vector search for table {:#?}", tbl.clone());
+            let primary_keys = table_primary_keys.get(&tbl).cloned().unwrap_or(vec![]);
 
-            // Only support one embedding column per table.
+             // Only support one embedding column per table.
```
phillipleblanc:
```suggestion
            // Only support one embedding column per table.
```

crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -174,9 +175,30 @@ impl VectorSearch {
 
                     let outtt: Vec<String> =
                         outt.iter().flat_map(std::clone::Clone::clone).collect();
+                    /// Retrieve the column data for just the primary keys
+                    /// This could be empty or also include the embedding column.
+                    let public_key_data = batch
```
phillipleblanc: Why is this called `public_key_data`? Shouldn't it be called `primary_key_data`?

crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -89,6 +90,89 @@ impl VectorSearch {
         }
     }
 
+    /// Perform a single SQL query vector search.
+    async fn individual_search(
```
Jeadie, line 94: Refactored out of `.search(` for function length reasons.

phillipleblanc (dismissed):

phillipleblanc (approved):

lukekim (approved):

PR #2179, opened 2024-08-02
Benchmark: Discard first measurement of sql query, disable result caching

Sevenannn:
## 🗣 Description

Updates to ensure fair sql query results in benchmark tests;
* Discard the first measurement of sql query. This is to discard the abnormal results caused by warm up: e.g. starting up spark cluster, establish new connection
* Disable results caching in benchmark tests. This ensures the later iterations of query don't directly get results from cache.

### DIFF
--- a/crates/runtime/benches/bench.rs
+++ b/crates/runtime/benches/bench.rs
@@ -185,6 +185,10 @@ async fn run_query_and_record_result(
     query_name: &str,
     query: &str,
 ) -> Result<(), String> {
+    // Additional round of query run before recording results.
+    // To discard the abnormal results caused by: establishing initial connection / spark cluster startup time
+    let _ = run_query(rt, connector, query_name, query).await;
+
     tracing::info!("Running query `{connector}` `{query_name}`...");
     let start_time = get_current_unix_ms();
 
--- a/crates/runtime/benches/setup/mod.rs
+++ b/crates/runtime/benches/setup/mod.rs
@@ -71,13 +71,6 @@ fn build_app(
     connector: &str,
     acceleration: Option<&Acceleration>,
 ) -> App {
-    // Disable result caching in benchmark test
-    let results_cache = ResultsCache {
-        enabled: false,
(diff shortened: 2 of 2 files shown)


crates/runtime/benches/setup/mod.rs
```diff
@@ -70,7 +71,15 @@ fn build_app(
     connector: &str,
     acceleration: Option<&Acceleration>,
 ) -> App {
-    let mut app_builder = AppBuilder::new("runtime_benchmark_test");
+    // Disable result caching in benchmark test
+    let results_cache = ResultsCache {
```
sgrebnov: Like it. Just a note that we currently run queries directly using datafusion.sql functionality instead of `query` api so results cache is not used. If we can update this as well it will be great.

crates/runtime/benches/bench.rs
```diff
@@ -185,6 +185,10 @@ async fn run_query_and_record_result(
     query_name: &str,
     query: &str,
 ) -> Result<(), String> {
+    // Additional round of query run before recording results.
```
sgrebnov, line 189: I would recommend to move this to Spark run method (if it is Spark related only)
```
pub(crate) async fn run(
    rt: &mut Runtime,
    benchmark_results: &mut BenchmarkResultsBuilder,
) -> Result<(), String> {
    let test_queries = get_test_queries();


    // wake up query

    for (query_name, query) in test_queries {
        super::run_query_and_record_result(rt, benchmark_results, "spark", query_name, query)
            .await?;
    }

    Ok(())
}
```

Sevenannn: It’s not just about spark, we want to discard first warm up run for all connectors

crates/runtime/benches/setup/mod.rs
Sevenannn: Got it, will remove this update

sgrebnov (approved):

PR #2180, opened 2024-08-03
Enable message chains and tool calls for local LLMs

Jeadie:
## 🗣 Description
- Enable OpenAI compatible [tool use](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools) for local LLMs.
- Depends on new update in mistral.rs,  https://github.com/EricLBuehler/mistral.rs/pull/649
- A big change in the code, is going from mistral.rs's `RequestMessage::Completion` to `RequestMessage::Chat`. This is needed for multi part inputs, and for actor based messages (i.e. system prompts and tool calls). 
- Quality of models and tool use still WIP. There may be an issue in how/what mistral.rs uses HF).


Jeadie:
Need to investigate more, I think models like `from: huggingface:huggingface.co/meta-llama/Meta-Llama-3.1-8B` don't come with the expected chat templating OOTB. 

```
WARN runtime: e="Unable to load LLM from spicepod hf, error: Failed to run LLM health check: Failed to run the LLM chat model: Received messages for a model which does not have a chat template. Either use a different model or pass a single string as the prompt"
```

~~We could use a `params.chat_template` but currently mistral.rs only supports building LLMs with chat_templates from a file.~~
May not be entirely true in recent mistral.rs update.

phillipleblanc (approved):

PR #2196, opened 2024-08-05
Add index and resource tuning for postgres ghcr image to support postgres benchmark in sf1

Sevenannn:
## 🗣 Description

* Add index and tune resource to speed up tpch scale factor 1 queries in postgres
* Use scale factor 1 for postgres tpch benchmark tests


peasee (commented):
**question:** Shouldn't we re-use the indexes that are created from [`tpch_index.sql`](https://github.com/spiceai/spiceai/blob/trunk/test/tpch/tpch_index.sql), like we do for MySQL?

Sevenannn:
@peasee Different query engines optimizers plans the query differently, the same query may have completely different plan in different query engines, for example, mysql and postgres, therefore the index to speed up queries are not necessarily the same. For example, the 
`CREATE INDEX idx_part_p_partkey ON part (p_partkey);` index, which is critical to speed up postgres tpch q20, is not added in mysql index. That's why the index file is not reused.

phillipleblanc (approved):

PR #2236, opened 2024-08-08
WIP: Add types integration test for accelerator hardening

Sevenannn:
## 🗣 Description

Add integration test to cover arrow type -> postgres type -> arrow type round trip conversion, in order to make sure every type in accelerator beta hardening is covered.

The arrow record generation methods can be used for integration tests of other accelerator

WIP:
- [x] Initial test framework added
- [ ] Support all arrow types covered in beta hardening 
- [ ] Migrate the test as unit test instead of integration test


Sevenannn:
Closed as migrated to datafusion table provider

PR #2242, opened 2024-08-08
feat: Set SQLite DB path to .spice/data

### LINKED ISSUE #2239: Enhancement: Set SQLite accelerator default db path to `~/.spice/data`
## Spec

Collect the user's home directory with `dirs::home_dir()`, and append `.spice/data` to the path. Prefix the regular database filename with the directory path, if a filename override is not specified.

This still allows users full control over the file location if they desire, e.g. with `sqlite_file: ./mydata/something.db`.

## How/MAP

* Add a parameter to accelerator table providers for a base directory path (e.g. `SqliteTableProviderFactory::new().db_base_folder()`)
* Append the base folder path to the file path for the default file path, e.g. to the `SqliteTableProviderFactory::sqlite_file_path()`
* Supply the user's home directory to `.db_base_folder()` during accelerator setup inside the Spice Runtime
* Ensure the base spice directory path exists to allow db file creation

This removes Spice specifics from the `datafusion-table-providers` crate which providers the table factory, while still providing the ability to change the root provider factory folder for the runtime.

peasee:
## 🗣 Description

* Set SQLite default DB path to `~/.spice/data`

## 🔨 Related Issues

* Closes #2239


phillipleblanc (dismissed):

crates/runtime/src/dataaccelerator.rs
```diff
@@ -260,6 +263,9 @@ pub async fn create_accelerator_table(
     acceleration_settings: &acceleration::Acceleration,
     secrets: Arc<RwLock<Secrets>>,
 ) -> Result<Arc<dyn TableProvider>> {
+    make_spice_data_directory()
+        .map_err(|err| Error::AccelerationCreationFailed { source: err.into() })?;
```
phillipleblanc: I don't think we want to do this unless we know we need to store something there. I can imagine a situation of Spice running in production where creating random folders could cause issues because of permissions issues.

crates/runtime/src/dataaccelerator/sqlite.rs
```diff
@@ -89,6 +96,11 @@ impl DataAccelerator for SqliteAccelerator {
         &self,
         cmd: &CreateExternalTable,
     ) -> Result<Arc<dyn TableProvider>, Box<dyn std::error::Error + Send + Sync>> {
+        if !cmd.options.contains_key("sqlite_file") {
```
phillipleblanc: Isn't the prefix getting stripped?

phillipleblanc (approved):

PR #2250, opened 2024-08-09
Improve tool quality and add vector search tool

Jeadie:
## 🗣 Description
- Create runtime tools `DocumentSimilarityTool`, `TableSchemaTool` and `ListDatasetsTool`.
- Improve readability of code running local tools
- Add recursive loop to `chat_request`. Now a model can call itself for more than just one tool use.

### DIFF
--- a/crates/runtime/src/embeddings/vector_search.rs
+++ b/crates/runtime/src/embeddings/vector_search.rs
@@ -107,16 +107,17 @@ impl VectorSearch {
         tbl: &TableReference,
         primary_keys: &[String],
         embedding: Vec<f32>,
-        embedding_column: String,
+        embedding_column: &str,
         n: usize,
     ) -> Result<(Vec<std::string::String>, Vec<arrow::array::RecordBatch>)> {
+        let projection = if primary_keys.is_empty() {
+            embedding_column.to_string()
+        } else {
+            [primary_keys.join(", "), embedding_column.to_string()].join(", ")
+        };
+
         let query = format!(
-            "SELECT {} FROM {} ORDER BY array_distance({}_embedding, {:?}) LIMIT {}",
-            [primary_keys.join(", "), embedding_column.clone()].join(", "),
-            tbl,
-            embedding_column,
-            embedding,
-            n
+            "SELECT {projection} FROM {tbl} ORDER BY array_distance({embedding_column}_embedding, {embedding:?}) LIMIT {n}"
         );
 
         let batches = self
@@ -241,7 +242,7 @@ impl VectorSearch {
                             &tbl,
                             &primary_keys,
                             embedding.clone(),
-                            embedding_column,
+                            &embedding_column,
                             n,
                         )
                         .await?;
--- a/crates/runtime/src/tool_use.rs
+++ b/crates/runtime/src/tool_use.rs
@@ -110,6 +110,33 @@ pub trait SpiceModelTool: Sync + Send {
 
 pub struct DocumentSimilarityTool {}
 
+static DOCUMENT_SIMILARITY_PARAMETERS: Lazy<Value> = Lazy::new(|| {
+    serde_json::json!({
+        "type": "object",
+        "properties": {
+            "text": {
+                "type": "string",
+                "description": "The text to search documents for similarity",
+            },
+            "from": {
+                "type": "array",
+                "items": {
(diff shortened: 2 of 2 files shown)


crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -107,16 +107,17 @@ impl VectorSearch {
         tbl: &TableReference,
         primary_keys: &[String],
         embedding: Vec<f32>,
-        embedding_column: String,
+        embedding_column: &str,
```
Jeadie, line 110: Issue in creating the projection when no `primary_keys` explicitly set

crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -107,16 +107,17 @@ impl VectorSearch {
...
+        };
+
         let query = format!(
-            "SELECT {} FROM {} ORDER BY array_distance({}_embedding, {:?}) LIMIT {}",
-            [primary_keys.join(", "), embedding_column.clone()].join(", "),
-            tbl,
-            embedding_column,
-            embedding,
-            n
+            "SELECT {projection} FROM {tbl} ORDER BY array_distance({embedding_column}_embedding, {embedding:?}) LIMIT {n}"
```
phillipleblanc, line 120: Should we implement `Display` on `embedding`? It feels weird to rely on the debug representation.

crates/runtime/src/tool_use.rs
```diff
@@ -106,7 +108,100 @@ pub trait SpiceModelTool: Sync + Send {
...
+#[derive(Deserialize)]
+struct DocumentSimilarityToolArgs {
+    text: String,
+    from: Vec<String>,
+    limit: Option<usize>,
+}
+
+#[async_trait]
+impl SpiceModelTool for DocumentSimilarityTool {
+    fn name(&self) -> String {
```
phillipleblanc:
```suggestion
    fn name(&self) -> &'static str {
```

This isn't ever generated at runtime right?

crates/runtime/src/tool_use.rs
```diff
@@ -106,7 +108,100 @@ pub trait SpiceModelTool: Sync + Send {
...
+        "document_similarity".to_string()
+    }
+
+    fn description(&self) -> Option<String> {
```
phillipleblanc:
```suggestion
    fn description(&self) -> Option<&'static str> {
```

crates/runtime/src/tool_use.rs
```diff
@@ -106,7 +108,100 @@ pub trait SpiceModelTool: Sync + Send {
...
+        Some("Search and retrieve documents from available datasets".to_string())
+    }
+
+    fn parameters(&self) -> Option<Value> {
```
phillipleblanc, line 157: Memo this, doesn't need to be generated each time.

crates/runtime/src/tool_use.rs
```diff
@@ -135,20 +230,111 @@ impl SpiceModelTool for ListTablesTool {
...
+        "table_schema".to_string()
+    }
+
+    fn description(&self) -> Option<String> {
+        Some("Retrieve the schema of all available SQL tables".to_string())
+    }
+
+    fn parameters(&self) -> Option<Value> {
+        Some(serde_json::json!({
+            "type": "object",
```
phillipleblanc: Memo this

crates/runtime/src/tool_use.rs
```diff
@@ -228,15 +420,92 @@ impl ToolUsingChat {
...
+                        tracing::error!("Error calling tool: {}", e);
+                        Value::String(format!("Error calling tool {}", t.name()))
+                    }
+                }
+            }
             None => Value::Null,
         }
     }
+
+    ///
```
phillipleblanc:
```suggestion
```

crates/runtime/src/tool_use.rs
Jeadie: Not yet, so happy to change

crates/runtime/src/embeddings/vector_search.rs
Jeadie: We can't implement it on `Vec<f32>` can we? We could make a helper function i suppose.

phillipleblanc (approved):

PR #2257, opened 2024-08-09
Track nested spans into Zipkin

phillipleblanc:
## 🗣 Description

Instruments the `ai_completion` calls with proper nested span tracking, with optional Zipkin tracing configuration.

The next step will be to configure the OTel traces exporter to go directly into the `task_history` table and remove the existing TaskSpan code.

### Testing

Run zipkin locally:
`docker run -d -p 9411:9411 openzipkin/zipkin`

Add this Spicepod config:

```yaml
runtime:
  tracing:
    zipkin_enabled: true
    zipkin_endpoint: http://localhost:9411/api/v2/spans
```

### DIFF
--- a/crates/runtime/src/http/v1/assist.rs
+++ b/crates/runtime/src/http/v1/assist.rs
@@ -32,7 +32,6 @@ use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::{collections::HashMap, sync::Arc};
 use tokio::sync::RwLock;
-use tracing::Instrument;
 
 use futures::StreamExt;
 
--- a/crates/runtime/src/http/v1/assist.rs
+++ b/crates/runtime/src/http/v1/assist.rs
@@ -32,6 +32,7 @@ use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::{collections::HashMap, sync::Arc};
 use tokio::sync::RwLock;
+use tracing::Instrument;
 
 use futures::StreamExt;
 
@@ -240,8 +241,11 @@ pub(crate) async fn post(
     )
     .label("tables".to_string(), format!("{input_tables:?}"));
 
+    let tspan = tracing::span!(target: "task_history", tracing::Level::INFO, "vector_search", input_text = %payload.text);
+
     let relevant_data = match vs
         .search(payload.text.clone(), input_tables, RetrievalLimit::TopN(3))
+        .instrument(tspan)
         .await
     {
         Ok(relevant_data) => {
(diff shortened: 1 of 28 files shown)


ewgenius (approved):

crates/runtime/src/http/v1/assist.rs
```diff
@@ -240,8 +241,11 @@ pub(crate) async fn post(
     )
     .label("tables".to_string(), format!("{input_tables:?}"));
 
+    let tspan = tracing::span!(target: "task_history", tracing::Level::INFO, "vector_search", input_text = %payload.text);
+
     let relevant_data = match vs
         .search(payload.text.clone(), input_tables, RetrievalLimit::TopN(3))
+        .instrument(tspan)
```
Jeadie: We already trace practically the same span (both with name `"vector_search"`), in `vs.search`. 
https://github.com/spiceai/spiceai/pull/2257/files#diff-2099bcaabb894d8accf6f48632574ffa7633d2c2b9c4d56383a84b7e1e8c7a4fR195

crates/runtime/src/http/v1/chat.rs
```diff
@@ -50,51 +51,62 @@ pub(crate) async fn post(
...
-                match model.write().await.chat_request(req).await {
-                    Ok(response) => {
-                        let preview = response
-                            .choices
-                            .first()
-                            .map(|s| serde_json::to_string(s).unwrap_or_default())
-                            .unwrap_or_default();
-                        span.truncated_output_text(Arc::from(preview)).finish();
-                        Json(response).into_response()
+    let tspan = tracing::span!(target: "task_history", tracing::Level::INFO, "ai_completion", input_text = %serde_json::to_string(&req).unwrap_or_default());
```
Jeadie: I don't know if it's possible, but it would be convenient if the `"ai_completion"` span was produced within the `Box<dyn Chat>`'s `.chat_request` & `.chat_stream` functions. We call this method in many places (e.g. recursively in `ToolUsingChat`, and right now we'd have to keep track of all of them, and keep parameters in sync.

crates/runtime/src/model/tool_use.rs
```diff
@@ -641,153 +655,158 @@ fn make_a_stream(
...
+                                    if let Err(e) = sender_clone.send(Err(e)).await {
+                                        tracing::error!("Error sending error: {}", e);
+                                    }
+                                    return;
+                                }
+                            };
+
+                            let mut new_req = req.clone();
+                            new_req.messages.clone_from(&new_messages);
+                            match model.chat_stream(new_req).await {
```
Jeadie, line 788: Follow on from above, this is an example. Even though it's within an `ai_completion`, I think we would want to refer to this as another `ai_completion` subspan.

Jeadie: Similar question above, on L630, with `self.chat_request(new_req).await`.

Issue #2234, opened 2024-08-08
Enhancement: Arrow Acceleration Beta Hardening


y-f-u:
## Goal-State/What/Result

Through Spice, be able to accelerate up to 20GB of data into in-memory Arrow

## Why/Purpose

To ensure that the arrow acceleration engine meets the beta hardening criteria for acceleration engines.

## By When

**Issue/Spec written and reviewed:** August 8th, 2024
**Done-Done:** August 12th, 2024

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

Basic specification of the enhancement.

### Feature set complete
- [x] ~~Streaming when accelerating/loading~~
- [x] ~~Index~~
- [x] ~~On conflict behaviors - Drop/Upsert~~
- [x] ~~Common/core arrow data type supports: Null, all Int types, all Float types, Time32, Time64, Timestamp (with/without TZ), Date32, Date64, Duration, Interval, Binary/LargeBinary/FixedSizeBinary, Utf8/LargeUtf8, List/FixedSizeList/LargeList, Struct, Decimal128/Decimal256~~
- [x] No known major bugs/issues

### Test Coverage
- [x] End-to-end test to cover accelerating TPCH-SF1 data from S3 and benchmarking TPCH queries (official and simple).
- [x] ~~Integration tests to cover accelerating data from S3 parquet, MySQL, Postgres with arrow types: Null, all Int types, all Float types, Time32, Time64, Timestamp (with/without TZ), Date32, Date64, Duration, Interval, Binary/LargeBinary/FixedSizeBinary, Utf8/LargeUtf8, List/FixedSizeList/LargeList, Struct, Decimal128/Decimal256~~
- [x] ~~Integration tests to cover "On Conflict" behaviors.~~
- [x] A reproducible test script that can load a variety of datasets into accelerators in both file/memory mode.
    - [x] Script can load TPCH-SF10
    - [x] Script can load TPCDS-SF10
[...]

PR #2262, opened 2024-08-12
run queries bash script

y-f-u:
## 🗣 Description

Add couple script to setup spicepod with tpch/tpcds datasets, also add a script to run queries on the datasets.

## 🔨 Related Issues

part of #2234

### DIFF
--- a/test/scripts/run-queries.bash
+++ b/test/scripts/run-queries.bash
@@ -1,0 +1,45 @@
+#!/bin/bash
+set -e
+
+clean_up () {
+    ARG=$?
+		rm runqueries.tmp.txt
+    exit $ARG
+}
+
+if [ -z "$1" ]; then
+		echo "No query folder provided. Aborting."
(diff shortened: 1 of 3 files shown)


peasee:
This seems like it duplicates some work from [https://github.com/spiceai/spiceai/blob/trunk/test/tpch/Readme.md#run-tcp-h-queries](https://github.com/spiceai/spiceai/blob/trunk/test/tpch/Readme.md#run-tcp-h-queries). Should we merge these scripts with those from that existing `Makefile`?

y-f-u:
> This seems like it duplicates some work from https://github.com/spiceai/spiceai/blob/trunk/test/tpch/Readme.md#run-tcp-h-queries. Should we merge these scripts with those from that existing `Makefile`?

yes, let's merge these scripts then update the readme separately

peasee (approved):

PR #2263, opened 2024-08-12
Add primary keys to response of `DocumentSimilarityTool`

Jeadie:
## 🗣 Description
- For the `DocumentSimilarityTool`, this pull request adds primary keys to the response.
- This is in an attempt to improve the ability for tool using models to reference where/how they got their data from.

### DIFF
--- a/crates/runtime/src/embeddings/vector_search.rs
+++ b/crates/runtime/src/embeddings/vector_search.rs
@@ -20,8 +20,10 @@ use app::App;
 use arrow::array::{RecordBatch, StringArray};
 use arrow::error::ArrowError;
 use async_openai::types::EmbeddingInput;
+use datafusion::common::utils::quote_identifier;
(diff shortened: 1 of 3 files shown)


crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -110,10 +112,19 @@ impl VectorSearch {
         embedding_column: &str,
         n: usize,
     ) -> Result<(Vec<std::string::String>, Vec<arrow::array::RecordBatch>)> {
+        // Need to handle quoting projection columns manually (when needed).
```
Jeadie, line 115: Bug occurs when there are primary keys that would have needed quoting. 
For example 
```sql
SELECT cveId, description_embedding FROM...
```

Needs to be 
```sql
SELECT "cveId", desciption_embedding FROM...
```

phillipleblanc (approved):

PR #2281, opened 2024-08-13
Use UTC time instead of machine time in recording benchmark test results

Sevenannn:
## 🗣 Description

Use UTC time instead of machine time in benchmark test results. This will standardize the time zone to be UTC in query results, and avoid confusion in time caused by runner's time zone shift


Sevenannn:
Closed as not needed

PR #2283, opened 2024-08-13
feat: Add a reference to originating App for a Dataset

### LINKED ISSUE #2273: Enhancement: Accelerators are aware of other accelerators
## Goal-State/What/Result

Accelerators have a mechanism to be aware of other configured accelerators in the Spice runtime.

## Why/Purpose

Specifically for SQLite accelerators, to be aware of other SQLite accelerators to attach them together for acceleration federation.

Accelerators being aware of other accelerators could also be a useful pattern for other acceleration engines.

## By When

**Issue/Spec written and reviewed:** 13th Aug, 2024
**Done-Done:** 14th Aug, 2024

## Done-Done

- [x] The Algorithm
- [x] PM/Design Review
- [x] ~UX Review~
- [x] ~Tests~
- [x] ~Documentation~
- [x] ~Samples/Quickstarts/Tutorials~
- [x] ~Release Notes~

## The Algorithm

- [x] Simplify.

## Spec

`Dataset` has a new property referencing the `App` it originates from.

This allows accelerators to construct related accelerated datasets if they need too, from the originating `App` specification.

## How/MAP

1. Create a `.with_app()` method on the `runtime::Dataset`
2. In `Runtime::datasets_iter` update the `try_from` to add `with_app` to add a reference for the `Runtime`'s current `App` to the `Dataset`.

### Implementation (issues)

- 
-

peasee:
## 🗣 Description

* Adds a property for the originating `App` of `Dataset`

This allows datasets to identify where they came from, which is helpful for the SQLite accelerator to find any other SQLite accelerators that are in the same app.

## 🔨 Related Issues

* Closes #2273

### DIFF
--- a/crates/app/src/lib.rs
+++ b/crates/app/src/lib.rs
@@ -34,7 +34,7 @@ use spicepod::{
     Spicepod,
 };
 
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, PartialEq)]
 pub struct App {
     pub name: String,
 
--- a/crates/app/src/lib.rs
+++ b/crates/app/src/lib.rs
@@ -34,7 +34,7 @@ use spicepod::{
     Spicepod,
 };
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
 pub struct App {
     pub name: String,
 
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -486,7 +486,12 @@ impl Runtime {
     }
 
     fn datasets_iter(app: &App) -> impl Iterator<Item = Result<Dataset>> + '_ {
-        app.datasets.iter().cloned().map(Dataset::try_from)
+        let app_ref = Arc::new(app.clone());
+        app.datasets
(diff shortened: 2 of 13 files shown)


crates/app/src/lib.rs
```diff
@@ -34,7 +34,7 @@ use spicepod::{
     Spicepod,
 };
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Clone)]
```
phillipleblanc: Why does this need to be clone?

crates/runtime/src/lib.rs
```diff
@@ -486,7 +486,12 @@ impl Runtime {
     }
 
     fn datasets_iter(app: &App) -> impl Iterator<Item = Result<Dataset>> + '_ {
-        app.datasets.iter().cloned().map(Dataset::try_from)
+        let app_ref = Arc::new(app.clone());
```
phillipleblanc: We shouldn't clone the app here, this should take an `Arc<App>` to begin with.

phillipleblanc (approved):

Issue #2280, opened 2024-08-13
Enhancement: SQLite inter-table acceleration federation


peasee:
## Goal-State/What/Result

SQLite acceleration should support federation of queries, include cross-table `JOIN` and cross-table `count()`, etc.

## Why/Purpose

Federation improves accelerated query performance and lowers memory usage by utilising the native SQLite query engine for.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** 14th Aug, 2024
**Done-Done:** 15th Aug, 2024

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

Basic specification of the enhancement.

## How/MAP

A clear and concise plan of how this should be implemented.

### Implementation (issues)

- 
-

PR #2293, opened 2024-08-15
feat: Add `DataAccelerator::init()` for SQLite acceleration federation

### LINKED ISSUE #2237: Enhancement: Use ATTACH DATABASE for SQLite Federated Accelerators
## Goal-State/What/Result

The Spice Runtime uses [`ATTACH DATABASE`](https://www.sqlite.org/lang_attach.html) to make all SQLite accelerated connectors available in each SQLite reader connection.

## Why/Purpose

This will provide the functionality for federated accelerator querying across SQLite accelerated connectors.

## By When

**Issue/Spec written and reviewed:** Aug 15th, 2024
**Done-Done:** Aug 16th, 2024

## Done-Done

- [x] The Algorithm
- [x] PM/Design Review
- [x] ~UX Review~
- [x] Tests
- [x] ~Documentation~
- [x] ~Samples/Quickstarts/Tutorials~
- [x] ~Release Notes~

## The Algorithm

- [x] Simplify.

## Spec

SQLite accelerators attach databases using a supplied list of database files on-disk.

## How/MAP

1. Update `datafusion-table-providers` to accept an option in `SqliteTableProviderFactory` which consumes a semicolon-separated string of database disk paths. E.g. `./path/one.sqlite;./path/two.sqlite;./path/three.sqlite`
2. Read the semicolon-separated list and attach the databases on read-mode `SqliteConnectionPool` using `ATTACH DATABASE`
3. Add `Dataset` as a function argument to `DataAccelerator::create_external_table()`, so accelerators can identify the originating dataset (and app) for the acceleration request
4. For SQLite, using the originating App collect a list of all other SQLite acceleration engines. Collect their database file locations, and set it as the option parameter used by the `SqliteTableProviderFactory`

peasee:
## 🗣 Description

* Adds originating dataset as an option for an accelerator in `create_external_table()`. This is used to source the parent App, and in turn find related SQLite accelerated datasets to attach.
* Creates a new `DataAccelerator::init()` trait method, to initialize accelerators before registering datasets.
* Implements `DataAccelerator::init()` for the SQLite accelerator, to create an empty SQLite database on-disk. This allows the SQLite accelerator to attach SQLite databases on dataset registration, as the files wouldn't exist before-hand.
* Implements logic in `SqliteAccelerator::create_external_table()` to attach SQLite databases together for SQLite accelerator federation.

## 🔨 Related Issues

* Closes #2237 
* Part of #2280


crates/runtime/src/dataaccelerator/sqlite.rs
```diff
@@ -63,7 +74,11 @@ impl SqliteAccelerator {
         if !dataset.is_file_accelerated() {
             return None;
         }
-        let acceleration = dataset.acceleration.as_ref()?;
+        let mut acceleration = dataset.acceleration.clone()?;
```
Sevenannn: Maybe consider cloning the acceleration param instead of the whole acceleration object here. What's finally used in `sql_file_path` function is acceleration param, and cloning the whole acceleartion object can be expensive.

y-f-u:
Concept looks good to me.

crates/runtime/src/lib.rs
```diff
@@ -592,6 +598,53 @@ impl Runtime {
...
+    /// Initialize datasets configured with accelerators before registering the datasets.
+    /// This ensures that the required resources for acceleration are available before registration,
+    /// which is important for acceleration federation for some acceleration engines (e.g. `SQLite`).
+    async fn initialize_accelerators(&self, datasets: &[Arc<Dataset>]) {
+        let spaced_tracer = Arc::clone(&self.spaced_tracer);
+
+        for ds in datasets {
+            if let Some(acceleration) = &ds.acceleration {
+                let retry_strategy = FibonacciBackoffBuilder::new().max_retries(None).build();
+                let _ = retry(retry_strategy, || async {
```
phillipleblanc: I don't think a retry policy makes sense here?

crates/runtime/src/dataaccelerator/sqlite.rs
```diff
@@ -88,18 +105,68 @@ impl DataAccelerator for SqliteAccelerator {
...
+                        } else {
+                            self.sqlite_file_path(other_dataset)
+                        }
+                    })
+                    .collect::<Vec<String>>()
+            })
+        }) {
+            cmd.options
+                .insert("attach_databases".to_string(), attach_databases.join(";"));
+        }
```
phillipleblanc, line 166: It feels weird to have this Sqlite specific code in the generic DataAccelerator path. Would a middleware style approach work better? (i.e. when registering a DataAccelerator, it can opt to register a middleware handle that runs here and has the option to modify `cmd`)

phillipleblanc (dismissed):

sgrebnov (dismissed):

crates/runtime/src/lib.rs
```diff
@@ -600,6 +650,11 @@ impl Runtime {
 
         let valid_datasets = Self::get_valid_datasets(app, LogErrors(true));
         let mut futures = vec![];
+
+        if !self.initialize_accelerators(&valid_datasets).await {
```
phillipleblanc: This isn't a friendly user experience to just bail if one of the datasets fails initialization - this function could return the list of valid datasets that didn't have any errors, and we continue initializing those.

phillipleblanc (approved):

Issue #2182, opened 2024-08-04
Enhancement: Federated Accelerated Tables for DuckDB


phillipleblanc:
## Goal-State/What/Result

For duckdb accelerator, accelerated tables have federation push-down enabled. Accelerated tables that accelerate into the same file/DB instance have federation join push down enabled.

## Why/Purpose

Tables which are accelerated do not get the benefits of federation push down to reduce the computation done in DataFusion. This is especially apparent when accelerating tables into an accelerator duckdb engine file mode. This negatively impacts the performance of accelerated tables.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

Basic specification of the enhancement.

## How/MAP

A clear and concise plan of how this should be implemented.

### Implementation (issues)

- [x] federation in a single duckdb acceleration table
- [x] #2313
- [x] Document
- [x] toggle to disable the federation on acceleration

lukekim:
- @Sevenannn re-purpose this issue for common type conversion across accelerators.
- @peasee split into intra and inter DB federation for SQLite.
- @Sevenannn split into intra and inter DB federation for PostgreSQL.
- @y-f-u split into intra and inter DB federation for DuckDB.

y-f-u:
@Sevenannn rewrote the common type conversion into #2285 . I am repurposing this for duckdb federated accelerated tables.

y-f-u:
Inter table is pending on the sqlite change of attach/init.

y-f-u:
actually federation is done. Follow-up regressions are raise, so close this in favour of those.

peasee:
This is done-done with regressions

PR #2299, opened 2024-08-16
Intra-table federation query on duckdb accelerated table

y-f-u:
## 🗣 Description

Intra table federation for accelerated table.

## 🔨 Related Issues

part of #2182 

image

## Concerns

- when there is a use_source action on zero result, the federation will be disabled to use original logic. This can be enhanced introducing a virtual execution to handle this case.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ clickhouse-rs = { git = "https://github.com/spiceai/clickhouse-rs.git", tag = "0
 datafusion = "41.0.0"
 datafusion-federation = "0.1"
 datafusion-federation-sql = { git = "https://github.com/spiceai/datafusion-federation.git", rev = "ae9da45f8b90451119f3ca032b75ad82a28d7547" }
-datafusion-table-providers = { git = "https://github.com/datafusion-contrib/datafusion-table-providers.git", rev = "a726db96439dc7916b8a2ce21c444074a6590235" }
+datafusion-table-providers = { git = "https://github.com/datafusion-contrib/datafusion-table-providers.git", rev = "6a383441d962cf8a0fe027f95a0b1dffb66ee414" }
 dirs = "5.0.1"
 dotenvy = "0.15"
 duckdb = "1.0.0"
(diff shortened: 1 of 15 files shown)


Cargo.toml
```diff
@@ -116,3 +116,6 @@ arrow-odbc = { git = "https://github.com/spiceai/arrow-odbc.git", rev = "24ecbdf
 
 # Tracking Issue: https://github.com/allan2/dotenvy/issues/113
 dotenvy = { git = "https://github.com/spiceai/dotenvy.git", rev = "e5cef1871b08003198949dfe2da988633eaad78f" }
+
+[patch."https://github.com/datafusion-contrib/datafusion-table-providers.git"]
+datafusion-table-providers = { path = "../../datafusion-contrib/datafusion-table-providers" }
```
y-f-u: organising now

y-f-u:
tests are failing, working on a fix.

crates/data_components/src/poly.rs
```diff
@@ -0,0 +1,111 @@
...
+};
+use datafusion_federation::{FederatedTableProviderAdaptor, FederationProvider};
+
+use crate::delete::{get_deletion_provider, DeletionTableProvider};
+
+#[derive(Clone)]
+pub struct PolyTableProvider {
+    write: Arc<dyn TableProvider>,
+    delete: Arc<dyn TableProvider>,
+    fed: Arc<dyn TableProvider>,
```
phillipleblanc: Why not just store the values with their concrete types?

y-f-u: This is meant to be handling all types of accelerator. The only thing can be concrete is delete which can use the adaptor. Will change it so.

y-f-u:
The test is broken for a known issue from sqlite. Which @peasee is looking at for a fix.

crates/runtime/src/datafusion.rs
```diff
@@ -818,9 +826,35 @@ impl DataFusion {
                     table_name: dataset_name.to_string(),
                 })?;
         }
+
         Ok(())
     }
 
+    pub async fn get_accelerated_table_provider(
```
y-f-u, line 833: this method doesn't return a &AcceleratedTable due to lifetime issue that `table` can't be used in returned reference.

Let me know if any better way to deal with this.

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #2316, opened 2024-08-19
Add gRPC action to trigger accelerated dataset refresh

sgrebnov:
## 🗣 Description

Adds support for triggering an accelerated dataset refresh via gRPC. This will be an alternative method for refreshing datasets used by Spice SDKs.

Tests will be added next.

## 🔨 Related Issues

https://github.com/spiceai/spice-java/issues/15

## 🤔 Concerns

1. Considered adding single `SpiceRuntimeCommand` action type (command_type: int, params: json encoded string) to handle all runtime commands in the future (simplify adding new commands). Selected current approach as it is more explicit (params) and easier to use/integrate with.
2. PR does not include .proto file that is usually used to generate gRPC client code.

### DIFF
--- a/crates/runtime/src/flight/actions/datasets.rs
+++ b/crates/runtime/src/flight/actions/datasets.rs
@@ -1,0 +1,51 @@
+/*
+Copyright 2024 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use tonic::Status;
+
+use crate::{datafusion::Error, flight::Service};
+
+#[derive(Clone, PartialEq, ::prost::Message)]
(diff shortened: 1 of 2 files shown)


peasee (approved):

crates/runtime/src/flight/actions/datasets.rs
```diff
@@ -0,0 +1,51 @@
...
+
+pub(crate) async fn do_action_accelerated_dataset_refresh(
+    flight_svc: &Service,
+    cmd: ActionAcceleratedDatasetRefreshRequest,
+) -> Result<(), Status> {
+    tracing::trace!("do_action_accelerated_dataset_refresh: {cmd:?}");
+
+    let dataset_name = &cmd.dataset_name;
+
+    match flight_svc.datafusion.refresh_table(&cmd.dataset_name).await {
```
phillipleblanc, line 35: This should probably take a `TableReference` not a string, since this wouldn't work for datasets configured with a schema in the name.

phillipleblanc: And then that would mean we would need to call TableReference::parse_str on `cmd.dataset_name`

sgrebnov: Verified names with schema support - this works correct as `table_provider` method accept `table_ref: impl Into<TableReference>` and there is special helper that automatically parses string to TableReference so manual string to TableReference conversion is not required. Also tested to make sure it works, will include table name with schema in tests as well
```
impl<'a> From<&'a str> for TableReference {
    fn from(s: &'a str) -> Self {
        Self::parse_str(s)
    }
}
...
```

phillipleblanc (approved):

PR #2327, opened 2024-08-20
add `disable_query_push_down` option to acceleration settings

y-f-u:
## 🗣 Description

Add a new `disable_query_push_down` option to the acceleration settings which allows users to escape from the default federated query on accelerator.

UI:

```
    acceleration:
      engine: postgres
      enabled: true
      params:
        disable_query_push_down: true
```

## 🔨 Related Issues

part of #2182 

## 🤔 Concerns

This pattern may eventually lead to a group of "support" keys in the acceleration settings. However, we are not implementing that struct yet until we see another use case emerge.

### DIFF
--- a/crates/runtime/src/component/dataset/acceleration.rs
+++ b/crates/runtime/src/component/dataset/acceleration.rs
@@ -244,7 +244,7 @@ pub struct Acceleration {
 
     pub on_conflict: HashMap<ColumnReference, OnConflictBehavior>,
 
-    pub disable_federation: bool,
+    pub disable_query_push_down: bool,
 }
 
 impl TryFrom<spicepod_acceleration::Acceleration> for Acceleration {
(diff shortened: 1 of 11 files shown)


crates/runtime/src/component/dataset/acceleration.rs
```diff
@@ -206,6 +206,7 @@ impl Display for OnConflictBehavior {
 }
 
 #[derive(Debug, Clone, PartialEq)]
+#[allow(clippy::struct_excessive_bools)]
```
y-f-u, line 208: allow this explicitly to prioritize the clarity of the configuration struct.

crates/spicepod/src/component/dataset.rs
```diff
@@ -236,6 +236,7 @@ pub mod acceleration {
 
     #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
     #[cfg_attr(feature = "schemars", derive(JsonSchema))]
+    #[allow(clippy::struct_excessive_bools)]
```
y-f-u: allow this explicitly to prioritize the clarity of the configuration struct.

phillipleblanc (approved):

Issue #1715, opened 2024-06-17
Enhancement: Anonymous and optional usage telemetry


lukekim:
## Goal-State/What/Result

Opt-out, anonymous, privacy-preserving usage telemetry of Spice is collected, including versions, component usage, and runtime performance KPIs.

## Why/Purpose

To understand Spice usage so that future versions can be improved.

## By When

**Issue/Spec written and reviewed:** Tue, Jul 30
**Done-Done:** v0.17.2-beta (Mon, Aug 26)

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### Scope

While the spec makes provisions for the Spice SDKs to also make use of the Usage Telemetry APIs, this spec only focuses on the metrics for `spiced`.

### Enablement

Anonymous usage telemetry is enabled by default, but supports opt-out.

The usage telemetry components can be disabled at compile-time by removing the `telemetry` feature flag, which will remove all usage telemetry code from being compiled into the final binary.

It can also be disabled at runtime via a `spiced` command line argument `--telemetry-enabled false`.

Disabling via the `runtime` section in `spicepod.yml` is possible with:

```yaml
runtime:
  telemetry:
    enabled: false
```

### Anonymous Instance ID

An `instance_id` is populated for each telemetry metric. This `instance_id` will be a SHA256 hash of the current hostname + the name of the root spicepod. 
`instance_id = SHA256(hostname + spicepod.name)`

This allows aggregating and de-duplicating telemetry metrics by unique instances. The hash protects privacy by making it impossible get the original hostname/spicepod name inputs.

### Usage Metrics

All usage metrics are batched, aggregated and sent once an hour, except `ping` which is every 5 minutes. The metrics will be kept in Arrow format (see below for schema) and sent via Flight.

- Ping (`ping`)
  - Dimensions: N/A
- Bytes scanned (`bytes_scanned`)
  - Description: Total bytes scanned by DataFusion when serving queries during the interval period.
[...]

PR #2333, opened 2024-08-21
Add query telemetry metrics

### LINKED ISSUE #2218: Implement `bytes_returned` telemetry

### LINKED ISSUE #2219: Implement `query_duration_ms` telemetry

### LINKED ISSUE #2221: Implement P90/95/99 query time telemetry

### LINKED ISSUE #2222: Implement P90/95/99 query execution time telemetry

phillipleblanc:
## 🗣 Description

Implements the following telemetry metrics:

- `bytes_returned`: The number of bytes returned by the query execution to the client.
- `query_duration`: The total amount of time spent planning and executing queries, in ms.
- `query_execution_duration`: The total amount of time spent only executing queries, in ms. This is 0 for cached queries.

Note: In line with OTel best practices, I'm not encoding the unit in the metric name, that is a dedicated field in the metric payload. (Unlike Prometheus, which typically would have the unit in the metric name.)

The `bytes_scanned` metric will come in a separate PR, as it will require rewriting the LogicalPlan to insert a tracking layer for each TableScan to measure the number of bytes actually scanned.

## 🔨 Related Issues

Closes #2218
Closes #2219
Closes #2221
Closes #2222
Part of #1715

## 🤔 Concerns

The spec (#1715) called for having the `data_connector` and `accelerator` as dimensions for the query related metrics. There can be multiple data connectors/accelerators involved in a query, and I don't think it makes sense to try to encode them as a delimiter separated string (i.e. comma separated), as that wouldn't be very useful for actually calculating which connectors/accelerators are in use.

We should probably have separate metric(s) that are emitted for each instance we detect of a data connector/accelerator being used in a query, and is separate from the other query metrics.


crates/telemetry/src/lib.rs
```diff
@@ -15,18 +15,86 @@ limitations under the License.
...
         .with_description("Number of queries run.")
         .with_unit("queries")
         .init()
 });
+
+pub fn track_query_count() {
+    QUERY_COUNT.add(1, &[]);
+}
+
+static BYTES_SCANNED: LazyLock<Counter<u64>> = LazyLock::new(|| {
```
Sevenannn, line 44: Learning Question: Could you explain why the LazyLock is used here? It looks like there are several nested layers of LazyLock within LazyLock::new(|| METER). Is it for hierarchical lazy initialization?

phillipleblanc: Yeah, its so that each metric is only initialized when its first used, and the first metric to be used will also lazily initialize the METER itself.

sgrebnov (approved):

PR #2337, opened 2024-08-22
Propagate errors through all `task_history` nested spans

phillipleblanc:
## 🗣 Description

Correctly propagates errors to each individual task_history span.

### DIFF
--- a/crates/runtime/src/model/tool_use.rs
+++ b/crates/runtime/src/model/tool_use.rs
@@ -146,25 +146,38 @@ impl SpiceModelTool for DocumentSimilarityTool {
         rt: Arc<Runtime>,
     ) -> Result<Value, Box<dyn std::error::Error + Send + Sync>> {
         let span = tracing::span!(target: "task_history", tracing::Level::INFO, "tool_use::document_similarity", tool = self.name(), input = arg);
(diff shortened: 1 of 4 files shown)


crates/runtime/src/model/tool_use.rs
```diff
@@ -146,25 +146,38 @@ impl SpiceModelTool for DocumentSimilarityTool {
...
     ) -> Result<Value, Box<dyn std::error::Error + Send + Sync>> {
         let span = tracing::span!(target: "task_history", tracing::Level::INFO, "tool_use::document_similarity", tool = self.name(), input = arg);
-        let mut req: SearchRequest = serde_json::from_str(arg)?;
 
-        let vs = VectorSearch::new(
-            rt.datafusion(),
-            Arc::clone(&rt.embeds),
-            parse_explicit_primary_keys(Arc::clone(&rt.app)).await,
-        );
+        let tool_use_result = async {
```
phillipleblanc, line 150: The basic pattern is to wrap the code which can produce an error in a nested async block, instrument that block (which removes the need to instrument each call individually), and if the return value was an error, add the error to the current span

ewgenius (approved):

PR #2353, opened 2024-08-23
Add dataset details to tool using LLMs; early check tables in vector search

Jeadie:
## 🗣 Description
 - Add details about datasets in the system prompt for users.
 - For vector search, early check if tables are in runtime.


crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -40,8 +40,8 @@ use snafu::prelude::*;
 
 #[derive(Debug, Snafu)]
 pub enum Error {
-    #[snafu(display("Data source {} does not exist", data_source))]
```
Jeadie, line 43: This is just moving datasets from `String` -> `TableReference`, where appropriate.

phillipleblanc (approved):

PR #2366, opened 2024-08-26
nql parameter 'use' -> 'model'

Jeadie:
## 🗣 Description
Update `/v1/nql` parameter from `use` to `model`.

### DIFF
--- a/crates/runtime/src/http/v1/nsql.rs
+++ b/crates/runtime/src/http/v1/nsql.rs
@@ -54,7 +54,7 @@ fn clean_model_based_sql(input: &str) -> String {
 pub struct Request {
     pub query: String,
 
-    #[serde(rename = "model", default = "default_model")]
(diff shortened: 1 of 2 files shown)


crates/runtime/src/http/v1/nsql.rs
```diff
@@ -54,7 +54,7 @@ fn clean_model_based_sql(input: &str) -> String {
 pub struct Request {
     pub query: String,
 
-    #[serde(rename = "use", default = "default_model")]
+    #[serde(rename = "model", default = "default_model")]
```
phillipleblanc: This probably doesn't need a rename directive anymore right?

phillipleblanc (dismissed):

phillipleblanc (approved):

phillipleblanc (approved):

PR #2377, opened 2024-08-26
Install mysql with specified version in test ci

Sevenannn:


Sevenannn:
Close as no longer needed

PR #2378, opened 2024-08-26
Make E2E tests pass and add trace msg when runtime installation is required

sgrebnov:
## 🗣 Description

1. Disable MySQL quick start E2E
2. Disable DuckDB file mode acceleration E2E
3. Add trace msg when runtime installation is required

### DIFF
--- a/.github/workflows/e2e_test_ci.yml
+++ b/.github/workflows/e2e_test_ci.yml
@@ -365,7 +365,7 @@ jobs:
         if: matrix.target.target_os == 'linux'
         run: |
           sudo apt-get update
-          sudo apt-get install -y mysql-server
+          sudo apt-get install -y mysql-server=8.4*
           sudo systemctl start mysql.service
           sleep 5
           mysql -uroot -proot -e "CREATE USER test_user@localhost IDENTIFIED BY 'password';"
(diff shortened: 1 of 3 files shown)


bin/spice/pkg/runtime/runtime.go
```diff
@@ -41,6 +41,7 @@ func EnsureInstalled(flavor string) (bool, error) {
 	shouldInstall := false
 	var upgradeVersion string
 	if installRequired := rtcontext.IsRuntimeInstallRequired(); installRequired {
+		fmt.Println("Runtime installation is required")
```
lukekim:
```suggestion
		fmt.Println("Spice runtime installation required")
```

lukekim (dismissed):

Jeadie (approved):

.github/workflows/e2e_test_ci.yml
```diff
@@ -621,12 +622,14 @@ jobs:
...
         run: |
-          killall spice
+          killall spice || true
           cat spice.log
 
   test_local_acceleration:
     name: 'Test acceleration on ${{ matrix.target.name }} using ${{ matrix.acceleration.engine }}'
     runs-on: ${{ matrix.target.runner }}
+    env:
+      HOME: ${{ github.workspace }}
```
Sevenannn, line 632: Why is the `env: HOME: ${{ github.workspace }}` is used instead of `HOME=$GITHUB_WORKSPACE` here

sgrebnov: Moved on job level so Spice runtime (spiced) is installed/placed to the right folder (where spice cli will expect it to find). GITHUB_WORKSPACE was added due to write permission issue. We will be able to remove this as we've switched where spice data folder is located (not at HOME anymore)

Issue #2384, opened 2024-08-27
Enhancement: Github Connector


Jeadie:
## Goal-State/What/Result

Let the spice runtime retrieve data from Github APIs. 

## Why/Purpose

Several users have requested this data source. 

## By When

**Issue/Spec written and reviewed:** 27th August
**M1 - Basic functionality:** Sept 2
**Done-Done:** Sept 2

**Note**: created separate enhancement placeholder issue for filtering pushdowns support: https://github.com/spiceai/spiceai/issues/2453

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

#### Spicepod

```yaml
datasets:
  - from: github:github.com/spiceai/spiceai/files/v0.17.2-beta
    name: spiceai.files
    params:
      github_token: ${secrets:GITHUB_TOKEN}
      include:  **/*.json; **/*.yaml. # everything by default  
   acceleration:
      enabled: true # `content` column is added to the table with the file/blob content.
             
  - from: github:github.com/spiceai/spiceai/issues
    name: spiceai.issues
    params:
      github_token: ${secrets:GITHUB_TOKEN}
      
  - from: github:github.com/spiceai/spiceai/pulls
    name: spiceai.pulls
    params:
      github_token: ${secrets:GITHUB_TOKEN}

  - from: github:github.com/spiceai/spiceai/commits
    name: spiceai.commits
    params:
      github_token: ${secrets:GITHUB_TOKEN}
```

Note, Github REST API naming
```
https://api.github.com/repos/spiceai/spiceai/contents/?ref=trunk
https://api.github.com/repos/spiceai/spiceai/git/trees/v0.17.2-beta?recursive=1
```

#### Table schemas

```
sql> describe spiceai.files
+--------------+-----------+-------------+
| column_name  | data_type | is_nullable |
+--------------+-----------+-------------+
| name         | Utf8      | YES         |
| path         | Utf8      | YES         |
[...]

PR #2400, opened 2024-08-28
Add `content` column to GitHub Connector when dataset is accelerated

sgrebnov:
## 🗣 Description

Add `content` column to GitHub Connector when dataset is accelerated

## 🔨 Related Issues

https://github.com/spiceai/spiceai/issues/2384

## 🤔 Concerns

`download_url` array can be used to fetch the content of the file, but this involves downcasting and using additional itertool as Array does not support chunking by default. This can be optimized later if needed.

### DIFF
--- a/crates/data_components/src/github.rs
+++ b/crates/data_components/src/github.rs
@@ -14,12 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 use async_trait::async_trait;
+use futures::future;
 use globset::GlobSet;
 use snafu::{ResultExt, Snafu};
 
 use crate::arrow::write::MemTable;
 use arrow::{
-    array::{RecordBatch, StringBuilder, UInt64Builder},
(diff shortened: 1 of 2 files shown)


Sevenannn (approved):

crates/data_components/src/github.rs
```diff
@@ -63,19 +65,35 @@ impl GithubFilesTableProvider {
...
             Field::new("size", DataType::UInt64, true),
             Field::new("sha", DataType::Utf8, true),
             Field::new("mode", DataType::Utf8, true),
             Field::new("url", DataType::Utf8, true),
-        ]));
+            Field::new("download_url", DataType::Utf8, true),
+        ];
+
+        if fetch_content {
+            fields.push(Field::new("content", DataType::Utf8, true));
```
phillipleblanc, line 81: I don't think it makes sense to change the schema based on whether its accelerated or not.

PR #2401, opened 2024-08-28
Fix repeated base table data in acceleration with embeddings

Sevenannn:
## 🗣 Description

When accelerated table is configured with embeddings and refresh_append_overlap, the data within the `(max(time_column) - refresh_append_overlap)` is refreshed from data source within every `refresh_check_interval`. This is the expected behavior.

Embeddings will be created for the refresh data fetched. Since embeddings is not deterministic, so even when data in accelerated table is same with refreshed data, when different embeddings is for refreshed data, the (same base table + new embedding) data will be inserted into accelerated table, causing unwanted base table data duplication.

This PR includes the following changes:
* Add a method to access base table schema in `EmbeddingTable`
* Introduce the `filter_schema` in refresh task, which will be used in determining which columns to be compared in refresh task.
* Only compare the columns that's within the base table schema, i.e., exclude `*_embedding` column created by the embedding table when comparing refreshing data with existing data.
* Introduce a new struct `BaseSchema` in `runtime::datafusion::schema`. This is a helper structure that abstracts methods of fetching schema for given table providers.


crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -751,18 +759,48 @@ impl RefreshTask {
 fn filter_records(
     update_data: &RecordBatch,
     existing_records: &Vec<RecordBatch>,
+    base_table_schema: Option<SchemaRef>,
```
sgrebnov: Could be better just `filter_schema` or `comparison_schema`

sgrebnov (dismissed):

sgrebnov (dismissed):

ewgenius (dismissed):

Jeadie (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #2433, opened 2024-08-29
fix: Only attach successfully initialized accelerators

### LINKED ISSUE #2368: Bug: Attached accelerators fail when an accelerator fails initialization
**Describe the bug**

Take the follow spicepod as an example:

```yaml
version: v1beta1
kind: Spicepod
name: spice_app
datasets:
- from: duckdb:customer
  name: customer
  acceleration:
    enabled: true
    engine: sqlite
    mode: file
    params:
        sqlite_file: ./customer.notsqlite
  params:
    duckdb_open: ./tpch.duckdb
- from: duckdb:region
  name: region
  acceleration:
    enabled: true
    engine: sqlite
    mode: file
  params:
    duckdb_open: ./tpch.duckdb
- from: duckdb:part
  name: part
  acceleration:
    enabled: true
    engine: sqlite
    mode: file
  params:
    duckdb_open: ./tpch.duckdb
```

An error is induced by using `.notsqlite`, which isn't a valid SQLite database extension.

In this example, the error from `.notsqlite` causes the database to fail initialization. However, the database isn't removed from the related SQLite accelerators during their table registration causing these connectors to fail as well.

**To Reproduce**

1. Configure a spicepod with DuckDB or SQLite accelerators
2. Configure one of the accelerators incorrectly to induce an error
3. See behavior

**Expected behavior**

The failing connector/accelerator should be excluded from attachment to the remaining connectors/accelerators.

peasee:
## 🗣 Description

* Creates new trait methods on `DataAccelerator` to determine if an accelerator is ready and successfully setup: `is_initialized`, `file_path`, `has_existing_file`
* Creates new `Dataset` method, `is_accelerator_initialized`
* Uses the new methods to create a new `Runtime` method, `get_initialized_datasets`
* Updates SQLite and DuckDB database attachments to use `Runtime::get_initialized_datasets` to only attach successfully initialized accelerators

## 🔨 Related Issues

* Closes #2368


crates/runtime/src/dataaccelerator/duckdb.rs
```diff
@@ -123,6 +126,19 @@ impl DataAccelerator for DuckDBAccelerator {
...
+        self.duckdb_file_path(dataset)
+    }
+
+    fn is_initialized(&self, dataset: &Dataset) -> bool {
+        if !dataset.is_file_accelerated() {
+            return true; // memory mode DuckDB is always initialized
+        }
+
+        // otherwise, we're initialized if the file exists
+        self.has_existing_file(dataset)
```
Sevenannn, line 139: Is `has_existing_file` sufficient for judging whether accelerator is initialized? Both DuckDB and SQLite init has another step of initializing pool after validating the file path in `init` function, where accelerator could fail to be initialized as well.

peasee: Initializing the pool is just a shortcut for creating the file. We do nothing with the pool, it gets dropped immediately. `init` also validates the file extension is correct, so if the file exists then we can assume initialization was successful.

phillipleblanc (approved):

PR #2435, opened 2024-08-29
Fix tool calling with Groq (and potentially other tool-enabled models)

phillipleblanc:
## 🗣 Description

Fixes a bug where the input to a tool call was duplicated, leading to errors when parsing the arguments.

This was initially developed against the OpenAI model, which likes to stream the arguments for the tool call chunks bit by bit, i.e.:

Screenshot 2024-08-29 at 3 23 49 PM

Because of this, we need to build up the argument list piece by piece as we receive it from the model. The problem was that we initialized the state with the argument list of the first message, and then immediately appended the first message again. This worked for the OpenAI model, because the first message always contained an empty argument list. However, the Groq model does not do this, and so we ended up with the argument list being duplicated causing the function call to fail.

The fix is to not set the argument list from the first message when initializing the state, but instead to set it to an empty string. Then the logic to append the current chunk takes care of both the first message and subsequent ones.

Here is what Groq looks like, the first message contains the complete argument list:

Screenshot 2024-08-29 at 3 26 07 PM


.vscode/tasks.json
```diff
@@ -0,0 +1,15 @@
+{
```
Jeadie, line 1: Nooooo

Jeadie (approved):

PR #2450, opened 2024-09-01
Enable `ArrayDistance` to handle arrays of vectors

Jeadie:
## 🗣 Description
 - Currently `ArrayDistance` can only handle vectors (i.e. FixedSizeListArray[float32/64], Large/ListArray[float32/64]). This PR extends the functionality of `ArrayDistance` to handle arrays of vectors.
 - This will enable us to process distance calculations for tables with multiple vectors per row (e.g. when spiced starts to handle vector chunks). 
 - Note: `ArrayDistance` not at the state to be a broadly adopted. Specifically tested for `VectorSearch`.


crates/runtime/src/embeddings/array_distance.rs
```diff
@@ -15,21 +15,294 @@ limitations under the License.
...
+                            "{} is not a valid inner scalar type for VectorBatches",
+                            f2.data_type()
+                        ))
+                    }
+                }
+                DataType::Float32 | DataType::Float64 => Ok(VectorBatchType::FixedVectors {
+                    size: *count, // for flat FixedSizeList, inner elements are numerics so size=count.
+                    scalar_type: VectorScalarType::try_from(f1.data_type())?,
+                }),
+                _ => unimplemented!(),
```
phillipleblanc: Lets return an error instead of panicking

crates/runtime/src/embeddings/array_distance.rs
```diff
@@ -15,21 +15,294 @@ limitations under the License.
...
+                        Err(format!(
+                            "{} is not a valid inner scalar type for VectorBatches",
+                            f2.data_type()
+                        ))
+                    }
+                }
+                DataType::Float32 | DataType::Float64 => Ok(VectorBatchType::ListVectors {
+                    scalar_type: VectorScalarType::try_from(f1.data_type())?,
+                }),
+                _ => unimplemented!(),
```
phillipleblanc: Return an error

crates/runtime/src/embeddings/array_distance.rs
```diff
@@ -15,21 +15,294 @@ limitations under the License.
...
+                DataType::FixedSizeList(f2, size) => {
+                    if Self::allowed_scalar_type(f2.data_type()) {
+                        Ok(VectorBatchType::ListOfFixedVectors {
+                            size: *size,
+                            scalar_type: VectorScalarType::try_from(f2.data_type())?,
+                        })
+                    } else {
+                        Err(format!(
+                            "{} is not a valid inner scalar type for VectorBatches",
+                            f2.data_type()
```
phillipleblanc: Let's use Snafu errors instead of Strings

PR #2452, opened 2024-09-02
GitHub connector: convert `labels` and `hashes` to primitive arrays

### LINKED ISSUE #2447: Github Data Connector: flatten `labels` and `hashes` column data as it is a list of structs with single field
Must be array/list of primitive type (string)

sgrebnov:
## 🗣 Description

`labels` and `hashes` columns are currently list of structs with single field. PR converts them to List of strings to simplify usage. 

PR extends `GraphQLTableProvider` to allow providing custom transformation function for raw graphql result.

Note: `test_to_primitive_type_list` does not copy/duplicate data, it re-uses original array data.

Before:

```shell
| labels          | List(Field { name: "item", data_type: Struct([Field { name: "name", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }]), nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} })      
```

After

```shell  
| labels          | List(Field { name: "name", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} })      

sql> select title, url, labels from issues limit 10
+----------------------------------------------------------------------------+------------------------------------------------+--------------------+
| title                                                                      | url                                            | labels             |
+----------------------------------------------------------------------------+------------------------------------------------+--------------------+
| e2e test for metal context                                                 | https://github.com/spiceai/spiceai/issues/296  | []                 |
| Matrix publish is flaky                                                    | https://github.com/spiceai/spiceai/issues/298  | [kind/bug]         |
| Support inferencing on multiple training runs                              | https://github.com/spiceai/spiceai/issues/297  | [kind/enhancement] |
| SpiceRack registry client is not setting User-Agent                        | https://github.com/spiceai/spiceai/issues/300  | [kind/bug]         |
| Releases should only be created with a combination of release tag + branch | https://github.com/spiceai/spiceai/issues/299  | [kind/bug]         |
| Basic polling data grid in dashboard                                       | https://github.com/spiceai/spiceai/issues/306  | [kind/enhancement] |
| Extend Spicepod definition (spec) with `results_cache` object              | https://github.com/spiceai/spiceai/issues/1422 | []                 |
| Implement LRU cache provider (size-based eviction + TTL)                   | https://github.com/spiceai/spiceai/issues/1423 | []                 |
| Enhancement: Specify Primary Key/Indexes for Acceleration Engines          | https://github.com/spiceai/spiceai/issues/1412 | [kind/enhancement] |
| Implement caching for HTTP sql requests                                    | https://github.com/spiceai/spiceai/issues/1424 | []                 |
+----------------------------------------------------------------------------+------------------------------------------------+--------------------+

Time: 0.023711792 seconds. 10 rows.  
```

## 🔨 Related Issues

Closes https://github.com/spiceai/spiceai/issues/2447

### DIFF
--- a/crates/arrow_tools/src/record_batch.rs
+++ b/crates/arrow_tools/src/record_batch.rs
@@ -143,6 +143,7 @@ mod test {
     use arrow::{
         array::{Int32Array, StringArray},
         datatypes::{DataType, Field, Schema, TimeUnit},
+        json::ReaderBuilder,
     };
 
     use super::*;
@@ -183,5 +184,74 @@ mod test {
     fn test_string_to_timestamp_conversion() {
         let result = try_cast_to(batch_input(), to_schema()).expect("converted");
         assert_eq!(3, result.num_rows());
+    }
+
+    fn parse_json_to_batch(json_data: &str, schema: SchemaRef) -> RecordBatch {
+        let reader = ReaderBuilder::new(schema)
+            .build(std::io::Cursor::new(json_data))
+            .expect("Failed to create JSON reader");
+
+        reader
+            .into_iter()
+            .next()
+            .expect("Expected a record batch")
+            .expect("Failed to read record batch")
+    }
+
+    #[test]
+    fn test_to_primitive_type_list() {
+        let input_batch_json_data = r#"
+            {"labels": [{"id": 1}, {"id": 2}]}
+            {"labels": null}
+            {"labels": null}
+            {"labels": null}
+            {"labels": [{"id": 3}{"id": null}]}
+            {"labels": null}
+            "#;
+
+        let input_batch = parse_json_to_batch(
+            input_batch_json_data,
+            Arc::new(Schema::new(vec![Field::new(
+                "labels",
+                DataType::List(Arc::new(Field::new(
+                    "struct",
+                    DataType::Struct(vec![Field::new("id", DataType::Int32, true)].into()),
+                    true,
+                ))),
+                true,
+            )])),
+        );
+
+        let expected_list_json_data = r#"
+            {"labels": [1, 2]}
+            {"labels": null}
+            {"labels": null}
+            {"labels": null}
+            {"labels": [3, null]}
+            {"labels": null}
+            "#;
+
+        let expected_list_batch = parse_json_to_batch(
+            expected_list_json_data,
+            Arc::new(Schema::new(vec![Field::new(
+                "labels",
+                DataType::List(Arc::new(Field::new("id", DataType::Int32, true))),
+                true,
+            )])),
+        );
+
+        let (processed_array, processed_field) = to_primitive_type_list(
+            input_batch.column(0),
+            &Arc::new(input_batch.schema().field(0).clone()),
+        )
+        .expect("to_primitive_type_list should succeed");
+
+        let processed_batch = RecordBatch::try_new(
+            Arc::new(Schema::new(vec![processed_field])),
+            vec![processed_array],
+        )
+        .expect("should create new record batch");
+
+        assert_eq!(expected_list_batch, processed_batch);
     }
 }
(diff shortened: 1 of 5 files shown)


crates/arrow_tools/src/record_batch.rs
```diff
@@ -80,12 +82,68 @@ pub fn try_cast_to(record_batch: RecordBatch, schema: SchemaRef) -> Result<Recor
     RecordBatch::try_new(schema, cols).context(UnableToConvertRecordBatchSnafu)
 }
 
+/// Flattens a list of struct types with a single field into a list of primitive types.
+/// The struct field must be a primitive type.
+///
```
Jeadie, line 88: nit: I'd make it more explicit that for `List[Struct]` with multiple keys, all except the first will be ignored.

crates/data_components/src/graphql/provider.rs
```diff
@@ -27,19 +27,39 @@ use datafusion::{
...
-    schema: SchemaRef,
+    gql_schema: SchemaRef,
+    table_schema: SchemaRef,
+    transform_fn: Option<TransformFn>,
 }
 
 impl GraphQLTableProvider {
-    pub async fn new(client: GraphQLClient) -> Result<Self> {
-        let (_, schema, _) = client.execute(None, None, None).await?;
+    pub async fn new(client: GraphQLClient, transform_fn: Option<TransformFn>) -> Result<Self> {
```
phillipleblanc: Let's change this to not require `transform_fn` but instead is set with a builder-like fn, i.e. `fn with_schema_transform(mut self, ...`

crates/data_components/src/graphql/provider.rs
```diff
@@ -27,19 +27,39 @@ use datafusion::{
 };
 use std::{any::Any, sync::Arc};
 
-use super::client::GraphQLClient;
 use super::Result;
+use super::{client::GraphQLClient, ResultTransformSnafu};
+
+pub type TransformFn = Arc<
+    dyn Fn(&RecordBatch) -> Result<RecordBatch, Box<dyn std::error::Error + Send + Sync>>
```
phillipleblanc: Can this just be a function pointer instead? https://doc.rust-lang.org/reference/types/function-pointer.html

It doesn't need to be a closure since the function itself doesn't close over any environment variables.

crates/data_components/src/graphql/provider.rs
```diff
@@ -64,14 +84,26 @@ impl TableProvider for GraphQLTableProvider {
...
+        let mut res = self
             .client
-            .execute_paginated(Arc::clone(&self.schema), limit)
+            .execute_paginated(Arc::clone(&self.gql_schema), limit)
             .await
             .boxed()
             .map_err(DataFusionError::External)?;
 
-        let table = MemTable::try_new(Arc::clone(&self.schema), res)?;
+        if let Some(transform_fn) = &self.transform_fn {
```
Jeadie, line 110: I don't know if it's needed (because it doesn't hurt performance that much), but i feel like it should be possible to only perform certain transform_fn if the projection contains certain keys. Probably unnecessary,

Jeadie (dismissed):

crates/data_components/src/graphql/provider.rs
phillipleblanc: Here is an example of using a function pointer in Spice: https://github.com/spiceai/spiceai/blob/trunk/crates/data_components/src/unity_catalog/provider.rs#L41

crates/arrow_tools/src/record_batch.rs
ewgenius: Added, also covered with test

crates/data_components/src/graphql/provider.rs
ewgenius: Changed to function pointer

crates/data_components/src/graphql/provider.rs
ewgenius: Refactored to use builder pattern

crates/data_components/src/graphql/provider.rs
ewgenius: That means we'll need to define list of fields, right?

phillipleblanc (approved):

crates/data_components/src/graphql/provider.rs
ewgenius: Merging as is for now, we'll check possible improvements separately

Issue #2147, opened 2024-07-30
Enhancement: Native schema inference for Data Connectors


phillipleblanc:
## Goal-State/What/Result

Each Data Connector uses the source's native mechanism to retrieve the table schema.

## Why/Purpose

The current approach most data connectors have taken is to do a `SELECT * FROM {table} LIMIT 1` and infer the schema from the returned results. As #2061 has demonstrated, this cannot be relied upon.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** Oct 18

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

Basic specification of the enhancement.

## How/MAP

A clear and concise plan of how this should be implemented.

### Implementation (issues)

- [x] #2460
- [x] #2461
- [x] #2462
- [x] #2463
- [x] #2464
- [x] #2465

Sevenannn:
Target EOW: Done-done

phillipleblanc:
Target EOW: Done-Done

phillipleblanc:
This is done!

PR #2493, opened 2024-09-03
Native schema inference for snowflake (and support timezone_tz, better numeric support)

### LINKED ISSUE #2465: Snowflake Native Schema Inference

phillipleblanc:
## 🗣 Description

Adds native schema inference for Snowflake using the query `SHOW COLUMNS IN <TABLE>`. Unlike normal data queries that return Arrow, this returns results from the Snowflake API as JSON.

I also realized when working on this PR that we had inconsistent support for numeric types. Now we treat all numeric types as Decimal128 - whereas previously it was inferring the type based on the query `SELECT * FROM <TABLE> LIMIT 1` which the Snowflake Arrow API had "variable" encoding for numeric types, i.e. if all of the data in the Arrow batch can fit within an Int16, then the type it returned was an Int16, etc. This meant that the types could change between queries, which we haven't seen but was a possibility. Now its consistent.

I also realized in testing that we didn't support the `TIMESTAMP_TZ` snowflake type. This PR adds support for that, and changes the `TIMESTAMP_NTZ` type to use Nanosecond timestamp arrays, which is what Snowflake already encodes timestamp values as.

## 🔨 Related Issues

Closes #2465 
Part of #2147

### DIFF
--- a/crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
+++ b/crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
@@ -112,59 +112,11 @@ impl<'a> AsyncDbConnection<Arc<SnowflakeApi>, &'a (dyn Sync)> for SnowflakeConne
 
         match res {
             snowflake_api::QueryResult::Json(resp) => {
-                let columns: Vec<Vec<serde_json::Value>> = resp
-                    .value
-                    .as_array()
-                    .ok_or_else(|| dbconnection::Error::UnableToGetSchema {
-                        source: "Response is not an array".to_string().into(),
-                    })?
-                    .iter()
-                    .map(|column| {
-                        column
-                            .as_array()
-                            .ok_or_else(|| dbconnection::Error::UnableToGetSchema {
-                                source: "Column data is not an array".to_string().into(),
-                            })
-                            .cloned()
-                    })
-                    .collect::<Result<Vec<_>, _>>()?;
-
-                let mut fields = Vec::new();
-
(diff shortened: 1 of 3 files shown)


crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
```diff
@@ -141,6 +191,7 @@ impl<'a> AsyncDbConnection<Arc<SnowflakeApi>, &'a (dyn Sync)> for SnowflakeConne
 
         let mut transformed_stream = stream.map(|batch| {
             batch.and_then(|batch| {
+                println!("batch schema: {:?}", batch.schema());
```
peasee:
```suggestion
```

peasee (dismissed):

peasee (approved):

crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
```diff
@@ -337,12 +361,94 @@ where
...
+        Some(t) => Err(Error::UnableToRetrieveSchema {
+            reason: format!("Unsupported Snowflake data type: {t}"),
+        }),
+        None => Err(Error::UnableToRetrieveSchema {
+            reason: "Missing data type".to_string(),
+        }),
+    }
+}
+
+fn parse_schema_from_json(resp: &serde_json::Value) -> Result<SchemaRef, Error> {
```
sgrebnov, line 397: Nit: might be good to have example response/JSON to simplify review

PR #2502, opened 2024-09-03
Add stargazers to GitHub data connector

lukekim:
## 🗣 Description

Adds `/stargazers` support.

## 🤔 Concerns

Stargazers has many fields. How many should we include?

### DIFF
--- a/crates/runtime/src/dataconnector/github.rs
+++ b/crates/runtime/src/dataconnector/github.rs
@@ -213,6 +213,46 @@ impl GithubTableArgs for IssueTableArgs {
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/runtime/src/dataconnector/github.rs
```diff
@@ -213,6 +213,46 @@ impl GithubTableArgs for IssueTableArgs {
...
+                            starred_at: starredAt
+                            node {{
+                                login
+                                name
+                                avatar_url: avatarUrl
+                                bio
+                                location
+                                company
+                                email
+                                x_username: twitterUsername
```
sgrebnov, line 238: nice :)

sgrebnov (approved):

PR #2510, opened 2024-09-04
Add refresh jitter to refreshing dataset acceleration

Jeadie:
## 🗣 Description
 - New `datasets.acceleration` parameters:
     -  refresh_jitter_enabled (bool): Whether to add a random jitter to whenever the dataset acceleration is refreshed. Default false.  
     -  refresh_jitter_max (string): An upper bound to the additional jitter duration added to `refresh_check_interval`. Default, if `refresh_jitter_enabled: true` is 10% of `refresh_check_interval`. 
 - When `refresh_check_interval` is set and `refresh_jitter_enabled: true`, the acceleration refresh job will run at `refresh_check_interval + rand(0, refresh_jitter_max)`. 
- For `Refresh::new`, due to number of `Option<T> ` parameters, move to using a builder pattern to make it easier to use.

## 🔨 Related Issues
 #2491


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -69,12 +71,13 @@ pub struct Refresh {
 }
 
 impl Refresh {
-    #[allow(clippy::needless_pass_by_value)]
+    #[allow(clippy::needless_pass_by_value, clippy::too_many_arguments)]
```
phillipleblanc: I think we should take this opportunity to turn this into a builder pattern given most of the arguments are optional. i.e. `Refresh::new(RefreshMode::Full).with_max_jitter(max_jitter)`

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -466,7 +488,7 @@ mod tests {
             MemTable::try_new(schema, vec![vec![batch]]).expect("mem table should be created"),
         ) as Arc<dyn TableProvider>;
 
-        let refresh = Refresh::new(None, None, None, None, RefreshMode::Full, None, None);
+        let refresh = Refresh::new(None, None, None, None, None, RefreshMode::Full, None, None);
```
sgrebnov: Nit: we can add `with_jitter` fn so we don't need to pass None everywhere / tests

sgrebnov (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #2515, opened 2024-09-04
Add refresh options to `POST /v1/datasets/:name/acceleration/refresh`

Jeadie:
### API Changes
 - For refreshes triggered manually, we can now set overrides for `refresh_sql` and `refresh_mode`.
 - In HTTP `POST /v1/datasets/:name/acceleration/refresh`, there is a new payload
 ```
 {
  "refresh_mode":  "(optional) <string>",  // Possibilities same as `acceleration.params.refresh_mode`, e.g. full, append. Default to the dataset's `refresh_mode`
  "refresh_sql": "(optional) <String>", //  Default to the dataset's `refresh_sql`
 }
 ```

### Refactoring
 - To achieve the above, we had to rearchitect `/crates/runtime/src/accelerated_table/`. The core change is that `RefreshTask` no longer tracks the accelerated table's `refresh: Arc<RwLock<Refresh>>` (which can change during the table's lifecycle: 1. manually, 2. on Spicepod changes).  This is best seen in the diff
 ```diff
 - pub async fn run(&self) -> super::Result<()>
+ pub async fn run(&self, request: Refresh) -> super::Result<()>;
```

## 🔨 Related Issues
 - #2492

### CODE REVIEW

File: crates/runtime/src/accelerated_table/refresh_task.rs [hunk 1]
```diff
@@ -86,13 +85,12 @@ impl RefreshTask {
...
         dataset_name: TableReference,
         federated: Arc<dyn TableProvider>,
-        refresh: Arc<RwLock<Refresh>>,
         accelerator: Arc<dyn TableProvider>,
     ) -> Self {
         Self {
             dataset_name,
             federated,
-            refresh,
+            // refresh,
```
phillipleblanc:
```suggestion
```

phillipleblanc (dismissed):

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -187,55 +186,65 @@ impl RefreshTask {
...
     }
 
-    async fn run_once(&self) -> Result<(), RetryError<super::Error>> {
-        self.mark_dataset_status(status::ComponentStatus::Refreshing)
-            .await;
-
-        let dataset_name = self.dataset_name.clone();
-
-        let mode = self.refresh.read().await.mode;
+    async fn run_once(&self, request: &Refresh) -> Result<(), RetryError<super::Error>> {
```
sgrebnov: Might be better to keep `refresh` param name fro consistency with other places?

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -100,16 +105,22 @@ impl RefreshTask {
                             }
                         }
                         Err(e) => {
-                            self.mark_dataset_status(status::ComponentStatus::Error)
-                                .await;
+                            Self::mark_dataset_status(
```
sgrebnov, line 108: We are not using sql value from above as it can be changed dynamically? https://github.com/spiceai/spiceai/pull/2515/files#diff-9fb43c580ce020358c24d68b4e431efbee2d5583820210dd3c06d499b985cb74R70

sgrebnov (dismissed):

crates/runtime/src/accelerated_table/refresh_task.rs
Jeadie: Thought I got them all. Thanks.

crates/runtime/src/accelerated_table/refresh_task.rs
Jeadie: I will. Initially I had it as `request` when the default `Refresh` was still in `RefreshTask.refresh`.

crates/runtime/src/accelerated_table/refresh_task/changes.rs
Jeadie: That was my initial thought yes. I could see us moving it to L82 (at the start of `Ok(change_envelope) => {`), so that we write the correct SQL to error metrics if an error occurs.

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -187,55 +185,65 @@ impl RefreshTask {
...
 
-        let get_data_update_result = match mode {
+        let get_data_update_result = match request.mode {
             RefreshMode::Disabled => {
                 unreachable!("Refresh cannot be called when acceleration is disabled")
             }
-            RefreshMode::Full => self.get_full_update().await,
-            RefreshMode::Append => self.get_incremental_append_update().await,
+            RefreshMode::Full => {
+                self.get_full_or_incremental_append_update(request, None)
```
Sevenannn: Should we use the name `refresh` instead of `request` here to keep consistent with the definition? Same for the  `get_full_or_incremental_append_update` below as well ` pub async fn get_full_or_incremental_append_update(
        &self,
        refresh: &Refresh,
        overwrite_timestamp_in_nano: Option<u128>,
    )`

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #2546, opened 2024-09-06
Fix to manual refresh

Jeadie:
## 🗣 Description
Possible issue with manual data accelerator refresh. Only occurs for `refresh_mode='append'`. The `refresh_sql` provided by the user is getting an additional `WHERE {time_column} > {latest_timestamp}` clause. This is causing zero results to be returned. When the user specifies a manual refresh, they do not expect this additional predicate.

Example 
- For `refresh_sql`: `"select * from test where territory == 'EMEA'"` (spicepod has refresh sql `territory != 'EMEA'`).
- Appends zero results.

```
dataset_name: Bare { table: "test" }, sql=Some(
    "select * from test where territory == 'EMEA'",
), filters=[
    BinaryExpr(
        BinaryExpr {
            left: Cast(
                Cast {
                    expr: Column(
                        Column {
                            relation: None,
                            name: "order_date",
                        },
                    ),
                    data_type: Timestamp(
                        Nanosecond,
                        None,
                    ),
                },
            ),
            op: Gt,
            right: Literal(
                TimestampNanosecond(1117497600000000000, None),
            ),
        },
    ),
]
```

## 🔨 Related Issues
 - Original PR to add `refresh_sql` to manual refresh: https://github.com/spiceai/spiceai/pull/2515

## 🤔 Concerns
 - Implementation is a possible fix. Wanted to raise issue first to confirm desired behaviour. Fix may not be optimal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -28,12 +28,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
-name = "adler2"
-version = "2.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
-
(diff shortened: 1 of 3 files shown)


Jeadie:
feature, not bug

Issue #2481, opened 2024-09-02
Enhancement: `/ready` API only reports 200 when initial dataset load is successful


lukekim:
## Goal-State/What/Result

`/ready` API only reports 200 when initial dataset/models load is successful.

## Why/Purpose

When running in production, the system (Kubernetes normally) needs to know when spice daemon is ready for requests, where ready includes 100% data integrity and consistency.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** 9/10/24
**Done-Done:** 9/10/24

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [ ] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [ ] Documentation
- [ ] Samples/Quickstarts/Tutorials
- [ ] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## How/MAP

A clear and concise plan of how this should be implemented.

### Implementation (issues)

- 
-

sgrebnov:
Planned outcome for this week: done-done

PR #2600, opened 2024-09-10
Tweak `/ready` to only report ready when components have all reported Ready

phillipleblanc:
## 🗣 Description

Updates the `/ready` API to only report ready when all of the components have loaded and reported ready. For accelerated datasets, they won't report ready until their initial refresh has finished, except in the case where there was previous acceleration data from a previous run, in which case it marks as ready immediately.

We already had the notion of a ComponentStatus, but it was static across the process. This change ties it to a struct on the runtime (this allows multple runtime instances to exist at once and have the correct status for each component within, this is useful for tests).

## 🔨 Related Issues

Part of #2481


crates/runtime/src/dataaccelerator.rs
```diff
@@ -323,6 +324,12 @@ pub async fn create_accelerator_table(
                 msg: format!("Unknown engine: {engine}"),
             })?;
 
+    // If we already have an existing file, it means there is data from a previous acceleration and we don't need
+    // to wait for the first refresh to complete to mark it ready.
+    if let Some(dataset) = dataset.filter(|d| accelerator.has_existing_file(d)) {
+        runtime_status.update_dataset(&dataset.name, status::ComponentStatus::Ready);
+    }
```
phillipleblanc: This is the check for an existing acceleration file

crates/runtime/src/datafusion/initial_load.rs
```diff
@@ -1,38 +0,0 @@
...
-
-use std::sync::MutexGuard;
-
-use crate::DataFusion;
-
-impl DataFusion {
-    pub fn mark_initial_load_complete(&self) {
-        let mut guard = self.get_initial_load_guard();
-        *guard = true;
-    }
```
phillipleblanc, line 25: All of this code goes away 🎉

crates/runtime/src/http/v1/ready.rs
```diff
@@ -24,7 +24,7 @@ use axum::{
 };
 
 pub(crate) async fn get(Extension(df): Extension<Arc<DataFusion>>) -> Response {
-    if df.is_initial_load_complete() {
+    if df.runtime_status().is_ready() {
```
phillipleblanc, line 27: This is main change for the `/ready` API. The rest of this PR is scaffolding to get here.

crates/runtime/tests/acceleration/on_conflict.rs
```diff
@@ -182,16 +184,7 @@ INSERT INTO event_logs (event_name, event_timestamp) VALUES
...
-    dataset_ready_check(&rt, "SELECT * FROM pg_on_conflict_upsert LIMIT 1").await;
-    dataset_ready_check(&rt, "SELECT * FROM pg_on_conflict_drop LIMIT 1").await;
-    dataset_ready_check(&rt, "SELECT * FROM duckdb_mem_on_conflict_drop LIMIT 1").await;
-    dataset_ready_check(&rt, "SELECT * FROM duckdb_mem_on_conflict_upsert LIMIT 1").await;
-    dataset_ready_check(&rt, "SELECT * FROM duckdb_file_on_conflict_upsert LIMIT 1").await;
-    dataset_ready_check(&rt, "SELECT * FROM duckdb_file_on_conflict_drop LIMIT 1").await;
-    dataset_ready_check(&rt, "SELECT * FROM sql_mem_on_conflict_drop LIMIT 1").await;
-    dataset_ready_check(&rt, "SELECT * FROM sql_mem_on_conflict_upsert LIMIT 1").await;
-    dataset_ready_check(&rt, "SELECT * FROM sql_file_on_conflict_upsert LIMIT 1").await;
-    dataset_ready_check(&rt, "SELECT * FROM sql_file_on_conflict_drop LIMIT 1").await;
```
phillipleblanc, line 194: 🎉

crates/runtime/src/lib.rs
```diff
@@ -647,13 +645,21 @@ impl Runtime {
...
                         continue;
                     }
                 };
 
+                // If we already have an existing file, it means there is data from a previous acceleration and we don't need
+                // to wait for the first refresh to complete to mark it ready.
+                if accelerator.has_existing_file(ds) {
+                    self.status
+                        .update_dataset(&ds.name, status::ComponentStatus::Ready);
+                }
```
phillipleblanc, line 661: This is the check for an existing acceleration file

crates/runtime/src/status.rs
```diff
@@ -43,49 +47,126 @@ impl Display for ComponentStatus {
...
+
+        if statuses.is_empty() {
+            return false; // No components registered yet
+        }
+
+        // Check if all registered components have been ready at least once
+        statuses
+            .keys()
+            .all(|component| ever_ready.contains(component))
+    }
```
phillipleblanc, line 171: This is the logic to check if the runtime is ready

sgrebnov (approved):

crates/runtime/src/status.rs
sgrebnov: Learning purpose: what will be behavior for dynamically added components (adding dataset to yaml while Spice is running) - will it be added to components and the status will be changed to not ready until new dataset/component is loaded?

lukekim (approved):

Sevenannn (approved):

crates/runtime/src/lib.rs
lukekim: @phillipleblanc what if there's an existing file but the previous initial load did not complete? It seems like there needs to be an additional check that the file exists and at least the initial load completed. You could read this from task history/metrics, but it would be evicted. Perhaps one way is to store a semaphore such as `.spice/data/ready.sem`.

crates/runtime/src/status.rs
```diff
@@ -43,49 +47,126 @@ impl Display for ComponentStatus {
...
+        };
+
+        if statuses.is_empty() {
+            return false; // No components registered yet
+        }
+
+        // Check if all registered components have been ready at least once
+        statuses
+            .keys()
+            .all(|component| ever_ready.contains(component))
```
Sevenannn, line 170: Learning question: Does it mean that as long as the component is marked as `ready` once, it will pass the is_ready() check regardless of the component current status, e.g. error, refreshing, etc. Is there any methods to check the readiness of individual components that are keys in statuses?

crates/runtime/src/status.rs
phillipleblanc: That shouldn't affect the `/ready` API response, as adding components at runtime is mainly a local convenience. And even if we did want to support loading components dynamically as a production scenario, loading a new component shouldn't cause the API to break for existing components that have already initialized. The `/ready` API is mainly for making sure that on initial startup we indicate when all of the components defined in the spicepod are ready to serve requests, and its safe to put the pod into the load balancer

crates/runtime/src/status.rs
phillipleblanc: That is correct. We don't have a need for that, but it could be added here if we wanted it later.

crates/runtime/src/lib.rs
phillipleblanc: I will handle this in a follow-on PR.

PR #2603, opened 2024-09-10
Allow for prefix to be kept in internal Parameters

Jeadie:
## Changes 
 - For data connector parameters, if the parameter starts with the connector's prefix, it is stripped from the parameters before being used.
   - The parameter mapping internally is `format: /tmp/documents`.

```yaml
name: my_files
from: file:/tmp/documents
params:
  file_format: docx
```
 - There are some use cases where we want the prefix to be retained, i.e. `file_format: docx` (currently, the user would have to specify `params.file_file_format`).
   - To allow for this, we first check if the data connector is expecting the unprefixed version (e.g. `format`), before checking for (e.g. `file_format`).


crates/runtime/src/parameters.rs
```diff
@@ -26,14 +26,20 @@ impl Parameters {
         let mut params: Vec<(String, SecretString)> = params
             .into_iter()
             .filter_map(|(key, value)| {
-                let mut unprefixed_key = key.as_str();
+                let mut key_to_use = key.as_str();
```
phillipleblanc: Let's add some tests - this logic is getting complicated

Jeadie:
~~Possible issue with secrets~~

Fixed.

phillipleblanc (approved):

PR #2608, opened 2024-09-11
WIP: Split benchmark tests into separate target for each connector / accelerator

Sevenannn:
## 🗣 Description

Split the test for connector / accelerator into separate targets inside the bench,


Sevenannn:
- Closed as new implementation is available in https://github.com/spiceai/spiceai/pull/2626

Issue #2605, opened 2024-09-10
Benchmark test shouldn't crash when snapshot test fails


Sevenannn:

Sevenannn:
- Close by: https://github.com/spiceai/spiceai/pull/2613

PR #2613, opened 2024-09-11
Snapshot tests shouldn't crash the Spice benchmark test

Sevenannn:
## 🗣 Description

This PR addresses the following issues for the snapshot test added to the benchmark test
* Snapshot test failure won't cause the benchmark process to panic
* Write snapshot test failure results back to spice benchmark result dataset
* Propagate both query failure and snapshot failure in benchmark test logs

After this change, snapshot test failure will look like the following:
image


Sevenannn:
- Close https://github.com/spiceai/spiceai/issues/2605

crates/runtime/benches/bench.rs
```diff
@@ -293,17 +312,30 @@ async fn record_explain_plan(
...
+            insta::assert_snapshot!(format!("{connector}_{query_name}_explain"), explain_plan);
+        });
+        if result.is_err() {
+            assertion_err = Some(format!("Snapshort assertion failed for {connector}, {query_name}"));
+        }
     });
+
+    if let Some(assertion_err) = assertion_err {
+        return Err(assertion_err);
+    }
```
phillipleblanc, line 394:
```suggestion
    assertion_err.ok_or_else(Err)?;
```

Does this work? (I haven't tested it)

crates/runtime/benches/bench.rs
```diff
@@ -243,8 +247,23 @@ async fn run_query_and_record_result(
...
+    if query_err.is_some() {
+        return Err(format!("Query Error: {}", query_err.unwrap_or_default()));
+    }
+
+    if snapshot_err.is_some() {
+        return Err(format!(
+            "Snapshot Test Error: {}",
+            snapshot_err.unwrap_or_default()
+        ));
     }
```
phillipleblanc:
```suggestion
    match (query_err, snapshot_err) {
    (Some(query), Some(snapshot)) => {
        return Err(format!("Query Error: {}; Snapshot Test Error: {}", query, snapshot));
    }
    (Some(query), None) => {
        return Err(format!("Query Error: {}", query));
    }
    (None, Some(snapshot)) => {
        return Err(format!("Snapshot Test Error: {}", snapshot));
    }
    (None, None) => {}
  }
```

crates/runtime/benches/bench.rs
Sevenannn: `ok_or_else` only works as follows
```
    assertion_err
        .clone()
        .ok_or_else(|| assertion_err.unwrap_or_default())?;
```

phillipleblanc (approved):

PR #2614, opened 2024-09-11
Utilize runtime.status().is_ready() to check acceleration dataset readiness in benchmark test

Sevenannn:
## 🗣 Description

Utilize runtime.status().is_ready() to check acceleration dataset readiness in the benchmark test, instead of waiting for 150s for every all acceleration datasets to load for every accelerator (e.g. sqlite takes 2min+, but arrow only takes 30s), thus improving the benchmark efficiency.

### DIFF
--- a/crates/runtime/benches/setup/mod.rs
+++ b/crates/runtime/benches/setup/mod.rs
@@ -106,7 +106,7 @@ where
         if f().await {
             return true;
         }
-        tokio::time::sleep(Duration::from_millis(100)).await;
+        tokio::time::sleep(Duration::from_secs(1)).await;
     }
     false
 }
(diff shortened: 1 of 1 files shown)


lukekim (dismissed):

crates/runtime/benches/setup/mod.rs
```diff
@@ -73,17 +79,38 @@ pub(crate) async fn setup_benchmark(
...
+where
+    F: FnMut() -> Fut,
+    Fut: Future<Output = bool>,
+{
+    let start = std::time::Instant::now();
+    while start.elapsed() < max_wait {
+        if f().await {
+            return true;
+        }
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
lukekim: I would make this 1 second.

phillipleblanc (dismissed):

Jeadie (dismissed):

crates/runtime/benches/setup/mod.rs
Sevenannn: Updated

phillipleblanc (approved):

PR #2630, opened 2024-09-12
Improve `/ready` to only mark a dataset ready iff the initial refresh completed

### LINKED ISSUE #2481: Enhancement: `/ready` API only reports 200 when initial dataset load is successful

### LINKED ISSUE #2612: `/ready` API doesn't mark accelerator ready if previous run failed to finish loading
Part of #2481

phillipleblanc:
## 🗣 Description

Improves the `/ready` API to mark an accelerated dataset as ready iff the initial refresh has already completed for all accelerated datasets. Previously this only supported file-based accelerator engines (DuckDB/SQLite), and Postgres was not included. The previous support was also limited in that it didn't know if the file contained a full copy of the dataset or if it was interrupted during load.

For file-based accelerators and the Postgres accelerator it now knows when a previous run loaded the data via a new `spice_sys_dataset_checkpoint` table that will now be created alongside the acceleration table data to store the state of whether a refresh has ever completed for a given dataset (i.e. checkpointed). A similar table (`spice_sys_metadata`) was previously created for the Debezium CDC connector to store state related to the Kafka connection, but now that table will be renamed to `spice_sys_debezium_kafka` with columns specific to its scenario instead of a generic metadata column with JSON.

The `spice_sys_metadata` table is now deprecated, which represents a breaking change from previous releases for users of the Debezium CDC connector.

## 🔨 Related Issues

Closes #2612 
Closes #2481


Jeadie:
>  iff the initial refresh has already completed, for all accelerator engines. 

Clarification, if two datasets, `d_a` and `d_b`, are in the same accelerator engine, will `d_a` be ready if its dataset is loaded into the accelerator, but `d_b` is not?

crates/runtime/src/dataaccelerator/metadata.rs
```diff
@@ -16,9 +16,10 @@ limitations under the License.
 
 //! Store and retrieve Spice metadata in durable accelerator tables
 //!
-//! The metadata table will be a new table `spice_sys_metadata` with two columns:
+//! The metadata table will be a new table `spice_sys_metadata` with three columns:
 //! - `dataset` (PRIMARY KEY, TEXT): The dataset the metadata entry corresponds to
-//! - `metadata` (TEXT): The metadata entry in JSON format
+//! - `key` (PRIMARY KEY, TEXT): The key of the metadata entry
```
Jeadie: I would add this to the explanation for `key`, _"allow multiple metadata values to be associated with a dataset. "_.

Jeadie (dismissed):

phillipleblanc:
> > iff the initial refresh has already completed, for all accelerator engines.
> 
> Clarification, if two datasets, `d_a` and `d_b`, are in the same accelerator engine, will `d_a` be ready if its dataset is loaded into the accelerator, but `d_b` is not?

No. It should be `iff the initial refresh has already completed for all accelerated datasets`. It doesn't matter how they are split up across engines.

crates/runtime/src/dataaccelerator/metadata.rs
```diff
@@ -27,6 +28,11 @@ use crate::component::dataset::{acceleration::Engine, Dataset};
 pub const METADATA_TABLE_NAME: &str = "spice_sys_metadata";
 pub const METADATA_DATASET_COLUMN: &str = "dataset";
 pub const METADATA_METADATA_COLUMN: &str = "metadata";
```
lukekim: It's weird this is called `metadata`. Isn't the entire table metadata? Maybe just `data` or `value` so you have `key` + `data` | `value`. Basically, this is a K/V store for each accelerator.

lukekim:
Should we hide this table by default from `show tables`?

crates/runtime/src/accelerated_table.rs
```diff
@@ -247,6 +250,18 @@ impl Builder {
         self
     }
 
+    /// Set the metadata provider for the accelerated table
```
```suggestion
    /// Set the checkpointer for the accelerated table
```

crates/runtime/src/accelerated_table.rs
```diff
@@ -247,6 +250,18 @@ impl Builder {
...
+    pub fn checkpointer(&mut self, checkpointer: DatasetCheckpoint) -> &mut Self {
+        self.checkpointer = Some(checkpointer);
+        self
+    }
+
+    /// Set the metadata provider for the accelerated table
```
```suggestion
    /// Set the checkpointer for the accelerated table
```

lukekim (approved):

PR #2641, opened 2024-09-13
Add 'RefreshOverrides::max_jitter' to 'POST /v1/datasets/:name/acceleration/refresh'

Jeadie:
## 🗣 Description
 - This PR adds the `Refresh::max_jitter` field into `RefreshOverrides::max_jitter` so that it can be set in `POST /v1/datasets/:name/acceleration/refresh` request payload.

```shell
curl -i -XPOST 127.0.0.1:8090/v1/datasets/my_dataset/acceleration/refresh \
  --data '{
    "refresh_mode": "append",
    "refresh_sql": "SELECT * FROM my_dataset",
    "max_jitter": "1s"
  }'
```


crates/runtime/src/accelerated_table/refresh_task_runner.rs
```diff
@@ -107,27 +107,16 @@ impl RefreshTaskRunner {
                                 }
                             }
                         },
-
```
Jeadie, line 110: Not explicitly related to this PR, but avoids having the additional `if let Some(overrides_opt) = overrides_msg`.

phillipleblanc (dismissed):

lukekim (approved):

lukekim (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

Issue #2476, opened 2024-09-02
QA: Comprehensive test coverage for spill-to-disk and re-hydration Accelerator scenarios


lukekim:
## Goal-State/What/Result

Rigorous test coverage that provides confidence in the matrix of spill-to-disk and re-hydration scenarios for DuckDB and SQLite accelerators.

## Why/Purpose

Using a persistent file for spill-to-disk and re-hydration is the most common configuration of Spice and should have very good test coverage to ensure there are no regressions.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] ~UX Review~ (not applicable)
- [x] Threat Model / Security Review
- [x] Tests
- [x] ~Documentation~ (not applicable)
- [x] ~Samples/Quickstarts/Tutorials~ (not applicable)
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### Scenarios

- [x] DuckDB auto-create file restart with re-hydration
- [x] DuckDB specified file restart with re-hydration
- [x] SQLite auto-create file restart with re-hydration
- [x] SQLite specified file restart with re-hydration
- [x] Restart with re-hydration after updated acceleration/app configuration: Index / PK

## How/MAP

Implement restart with re-hydration integration test (repeat 3 times)
1. Retrieve the number of rows using a native `MySQL` connection to use as a baseline.
1. Start Spice, retrieve row count and loaded items after acceleration is completed, and compare with the baseline.
1. Restart the runtime and ensure the loaded items remain consistent immediately after the runtime is loaded.
1. Simulate federated dataset access issue after the runtime is restarted, ensure query result remain consistent.

Note: add test to restart app during refresh ?
Note: do we need multiple datasets (federated mode)?

### Implementation (issues)
- [x] #2640
- https://github.com/datafusion-contrib/datafusion-table-providers/pull/101
- https://github.com/spiceai/spiceai/pull/2658

sgrebnov:
Target outcome: Done-done by next release (09/18)

sgrebnov:
[...]

PR #2643, opened 2024-09-13
Add basic integration test for Spice spill-to-disk and re-hydration scenario

sgrebnov:
## 🗣 Description

Add basic integration test for Spice spill-to-disk and re-hydration scenario

Next steps:
 - Simulate federated dataset access issue after restart, ensure query results remain consistent. Temporary removed as it didn't work as expeced (zero results returned after components loading)
 - More variations: duckdb, specifying file path, updated configuration

## 🔨 Related Issues

Part of https://github.com/spiceai/spiceai/issues/2476

## 🤔 Concerns

1. Don't currently verify initially loaded data against federated source using native connection. We can use sample data and  snapshotting for this. Will also investigate if there is a way to produce dataset data sha/md5/checksum which could be used instead.


peasee (dismissed):

crates/runtime/tests/rehydration/mod.rs
```diff
@@ -0,0 +1,199 @@
...
+    acceleration::{Acceleration, Mode},
+    Dataset,
+};
+use tracing::instrument;
+
+/// Validates spill-to-disk and rehydration functionality by simulating runtime restarts
+/// and checking data consistency.
+///
+/// 1. Retrieve num of rows using a native MySQL connection to use as a baseline
+/// 2. Start Spice, retrieve row count and top 100 items after acceleration is completed, compare with baseline
```
Sevenannn: Is the top 100 items retrieval going to be included in this PR

Sevenannn (dismissed):

crates/runtime/tests/rehydration/mod.rs
sgrebnov: Good catch - updated 👍

Sevenannn (approved):

phillipleblanc (approved):

PR #2646, opened 2024-09-13
Define GitHub `issues` data connector schema upfront

sgrebnov:
## 🗣 Description

PR fixes `Unable to get data from connector: External error: External error: Json error: whilst decoding field 'milestone_id': expected null got "MI_kwDOFWSkxs4AbBhg"` issue and similar when using `issues` table with GitHub data connector. Previously target schema was inferred from first  GrapgQL response that could have data for some columns empty resulting in invalid schema and errors in consequent queries. PRs updates this to have target schema upfront

**Breaking changes**:
1. Type of `closed_at`, `updated_at`, `created_at` has been changed from Utf8 to Timestamp

## 🤔 Concerns

1. `fn get_graphql_values(&self)` become large - to be improved along with adding the same schema upfront definition for other GH Connector tables:  `stargezers`, `pulls`, `commits`
1. `comments` was not flattened due to GraphQL limitation preventing to make author (login) available on required level. This PR does not change previous behavior just pointing that this should be improved.

### DIFF
--- a/crates/runtime/src/dataconnector/github.rs
+++ b/crates/runtime/src/dataconnector/github.rs
@@ -284,7 +284,6 @@ impl GithubTableArgs for IssueTableArgs {
             query.into(),
             "/data/repository/issues/nodes".into(),
             2,
-            //None,
             Some(gql_schema),
         )
     }
--- a/crates/runtime/src/dataconnector/github.rs
+++ b/crates/runtime/src/dataconnector/github.rs
@@ -16,7 +16,7 @@ limitations under the License.
 
 use crate::component::dataset::Dataset;
 use arrow::array::{Array, RecordBatch};
-use arrow_schema::{DataType, Field, Schema};
+use arrow_schema::{DataType, Field, Schema, SchemaRef};
 use async_trait::async_trait;
 use data_components::{
     github::{GithubFilesTableProvider, GithubRestClient},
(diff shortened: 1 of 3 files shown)


crates/runtime/src/dataconnector/github.rs
```diff
@@ -47,7 +47,8 @@ pub trait GithubTableArgs: Send + Sync {
     ///   1. The GraphQL query string
     ///   2. The JSON pointer to the data in the response
     ///   3. The depth to unnest the data
-    fn get_graphql_values(&self) -> (GraphQLQuery, JSONPointer, UnnestDepth);
+    ///   4. The GraphQL schema of the response data
+    fn get_graphql_values(&self) -> (GraphQLQuery, JSONPointer, UnnestDepth, Option<SchemaRef>);
```
peasee, line 51: this is turning into quite a long tuple type. maybe we should think about creating a custom return struct/type?

crates/runtime/src/dataconnector/github.rs
```diff
@@ -209,7 +213,80 @@ impl GithubTableArgs for IssueTableArgs {
             name = self.repo
         );
 
-        (query.into(), "/data/repository/issues/nodes".into(), 1)
+        let gql_schema = Arc::new(Schema::new(vec![
```
peasee, line 216: this schema looks static, and is quite long? we should move it into a separate function, like `IssueTableArgs::schema()`

crates/runtime/src/dataconnector/github.rs
```diff
@@ -209,7 +213,80 @@ impl GithubTableArgs for IssueTableArgs {
...
+                ))),
+                true,
+            ),
+        ]));
+
+        (
+            query.into(),
+            "/data/repository/issues/nodes".into(),
+            2,
+            //None,
```
peasee:
```suggestion
```

crates/runtime/src/dataconnector/github.rs
sgrebnov: Yeah 👍, plan to improve this when adding schema for the next GH connector table

crates/runtime/src/dataconnector/github.rs
sgrebnov: Yeah 👍, plan to improve this when adding schema for the next GH connector table

peasee (approved):

PR #2655, opened 2024-09-14
Enable Warn/Error traces from dependency components

sgrebnov:
## 🗣 Description

We don't currently show trace messages from dependency components, for example, `datafusion-table-providers`. PR enables important (Warnings and Errors) messages to be displayed.

## 🤔 Concerns

It might be better to include traces specifically for `datafusion-table-providers` only?


lukekim (approved):

phillipleblanc (approved):
I don't think generally we want to show any messages from dependencies - anything important to show should be bubbled up to the runtime as an Err which we would then format and show in one of the components we control.

PR #2684, opened 2024-09-16
Trace query result while running benchmark tests

sgrebnov:
## 🗣 Description

Trace query result while running benchmark tests (only first 10 rows from the first query iteration)


Sevenannn (dismissed):

Sevenannn (dismissed):

crates/runtime/benches/bench.rs
```diff
@@ -284,9 +284,34 @@ async fn run_query_and_record_result(
...
+                let num_rows = records
+                    .iter()
+                    .map(arrow::array::RecordBatch::num_rows)
+                    .sum::<usize>();
+
+                // Show the first 30 records of the result
+                let limited_records = records
+                    .iter()
+                    .take(1)
+                    .map(|x| x.slice(0, x.num_rows().min(30)))
```
lukekim:
```suggestion
                    .map(|x| x.slice(0, x.num_rows().min(10)))
```

10 is probably enough.

lukekim (dismissed):

crates/runtime/benches/bench.rs
sgrebnov: Updated. thank you

Sevenannn (approved):

PR #2685, opened 2024-09-16
Better spicepod parsing error messages

Sevenannn:
## 🗣 Description

A more user-friendly error message when the user configure spicepod.yaml that's not conforming to json schema.
* Add custom handling for the errors that specify `line` and `column` message
* For other errors, pass the original error

Note: `serde::Error` variations are private implementation, so the string pattern matching method with regex is adopted here.

Before:
image

After:
image


phillipleblanc:
This is closed in favor of https://github.com/spiceai/spiceai/issues/2691

PR #2717, opened 2024-09-17
Early check EmbeddingConnector if embedding models do not exist

Jeadie:
### Current
 - When spicepod dataset is configured with embedding column, if embedding model does not exist, or the spice runtime wasn't built with `--features models`, then a cryptic error will be returned at query time
```shell
 sql> select * from tbl
Query Error Arrow error: Invalid argument error: number of columns(25) must match number of fields(26) in schema
 ```

### Now
 - Check existence of embedding models for relevant columns when we initialise the dataset. 
     - This also helps when users have misconfigured either the embedding model or the dataset's embedding column.

### CODE REVIEW

File: crates/runtime/src/embeddings/connector.rs [hunk 1]
```diff
@@ -61,6 +62,18 @@ impl EmbeddingConnector {
             .map(|e| (e.column.clone(), e.model.clone()))
             .collect::<HashMap<_, _>>();
 
+        // Early check if embedding models are available.
+        for (column, model) in &embed_columns {
+            if !self.embedding_models.read().await.contains_key(model) {
+                return Err(DataConnectorError::InvalidConfigurationNoSource {
+                    dataconnector: "File".to_string(),
+                    message: format!(
+                    "Dataset '{}' expects to use embedding model '{model}' to embed column '{column}', but the model is not found.\nEither '{model}' is not defined in Spicepod (as an 'embeddings'), or the runtime does not have model support. For the latter, either: \n  1) `spice install ai` \n  2) Build spiced binary with flag `--features models`.",
```
lukekim: I think we should separate the checks so we don't have an ambiguous (and verbose) error message. We can detect if models is compiled in pretty easily. If it passes that check, then give the other error.

Jeadie: oh yes, good point. I completely forgot about `if cfg!(feature = `

sgrebnov (approved):

PR #2730, opened 2024-09-17
Check pageInfo is correct at initialisation of GraphQL connector

Jeadie:
## 🗣 Description
 - For GraphQL connector, if user specifies pagination, check if the `PageInfo` field is correct given the pagination direction. Example errors
 ```shell
WARN data_components::graphql::client: GraphQL query has pagination specified (last: 100), 
but invalid pagination fields: 'pageInfo' field needs both 'hasPreviousPage' and 'startCursor' 
for backward pagination.

WARN data_components::graphql::client: GraphQL query has pagination specified (first: 100), 
but invalid pagination fields: 'pageInfo' field needs both 'hasNextPage' and 'endCursor' 
for forward pagination.
 ```

### DIFF
--- a/crates/data_components/src/graphql/client.rs
+++ b/crates/data_components/src/graphql/client.rs
@@ -30,6 +30,7 @@ use serde::{Deserialize, Serialize};
 use serde_json::{json, Map, Value};
 use snafu::ResultExt;
 use std::{cmp::min, io::Cursor, sync::Arc};
+
 use url::Url;
 
 pub enum Auth {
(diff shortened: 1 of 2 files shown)


crates/data_components/src/graphql/client.rs
```diff
@@ -327,23 +375,28 @@ impl PaginationParameters {
...
+
+                        // Check [`PaginationArgument`] and `pageInfo` fields are consistent.
+                        if let Err(e) = pag_args.validate_page_info(field) {
+                            tracing::warn!("GraphQL query has pagination specified ({pag_args}), but invalid pagination fields: {e}");
+                            return (None, None);
                         }
+
+                        return (
+                            Some(PaginationParameters {
+                                resource_name: format!("{:?}", parent_field.name).replace('"', ""),
```
phillipleblanc: Why is this replace needed?

Jeadie: The `{:?}` outputs `'"field_name"'`, not `'field_name'`. There's no way I can find to render the `Text` easily otherwise.

peasee (dismissed):

crates/data_components/src/graphql/client.rs
phillipleblanc: I pushed up a commit that fixes it

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #2735, opened 2024-09-18
feat: use task history for dataset availability checks

### LINKED ISSUE #1662: Use `task_history` table for datasets availability monitoring
Enhancement for [Improved telemetry & dashboard](https://github.com/spiceai/spiceai/issues/1283): 

## Goal-State/What/Result

Extend the `runtime.query_history` table with information about datasets used in queries. Use this information to update the availability status for federated queries.

## Why/Purpose

To limit unnecessary synthetic requests (availability tests) to federated datasets and to provide a general improvement that allows filtering and aggregating query information by dataset.

## Spec
- `datasets`  column is added to `runtime.query_history` table (Utf8 Array; Not nullable). Existing schema is listed below

| column_name       | data_type                                                                                                       | is_nullable |
|-------------------|-----------------------------------------------------------------------------------------------------------------|-------------|
| query_id          | Utf8                                                                                                            | NO          |
| schema            | Utf8                                                                                                            | YES         |
| sql               | Utf8                                                                                                            | NO          |
| nsql              | Utf8                                                                                                            | YES         |
| start_time        | Timestamp(Nanosecond, None)                                                                                     | NO          |
| end_time          | Timestamp(Nanosecond, None)                                                                                     | NO          |
| execution_time    | Float32                                                                                                         | NO          |
| execution_status  | Int8                                                                                                            | NO          |
| rows_produced     | UInt64                                                                                                          | NO          |
| **queried_datasets**  | **List(Field { name: "item", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} })** | **YES**         |
| results_cache_hit | Boolean                                                                                                         | NO          |
| error_message     | Utf8                                                                                                            | YES         |

- `DatasetsHealthMonitor` is updated to first check the most recent successful query not from cache for a dataset to update availability:
  - If there is a recent query within the last 10 minutes, the dataset is considered available, and this information is used to update the internal `last_available_time` information.
  - If there is no recent query within the last 10 minutes, a fallback on a synthetic query is used.
  - A logical plan builder must be used to construct a query against the saved instance of the `query_history` TableProvider for optimal performance.

Note: Query information already contains a list of datasets used for telemetry purposes, so we do not increase query complexity/time.

peasee:
## 🗣 Description

* Uses the contents of `task_history` to determine datasets that have been successfully accessed in the previous 10 minutes, and excludes them from the dataset health check.

This saves on unnecessary queries to federated sources if the Runtime is in active use.

## 🔨 Related Issues

* Closes #1662

## 🤔 Concerns

* I've excluded adding an integration test, because `task_history` requires a fully-fledged Runtime instance with task spans, telemetry, etc which we currently don't have bootstrapping for in integration tests.


crates/runtime/src/datasets_health_monitor.rs
```diff
@@ -123,6 +127,63 @@ impl DatasetsHealthMonitor {
...
+                    .into_iter()
+                    .flatten()
+                    .collect::<Vec<_>>(),
+                _ => {
+                    unreachable!("Unexpected data type from task_history query result");
+                }
+            })
+            .flat_map(|datasets| datasets.split(',').collect::<Vec<_>>())
+            .map(std::string::ToString::to_string)
+            .collect::<Vec<_>>();
```
Sevenannn: Is it possible to directly do a `.collect::<HashSet<_>>()` here? so we don't need to iterate the Vec<String> once more to craete the hashset of datasets_with_recent_activity?

crates/runtime/src/datasets_health_monitor.rs
```diff
@@ -123,6 +127,57 @@ impl DatasetsHealthMonitor {
...
+    pub async fn get_recently_accessed_datasets(
+        df_ctx: Arc<SessionContext>,
+    ) -> Result<Arc<HashSet<String>>> {
+        let query = format!(
+            "
+SELECT labels.datasets AS datasets
+FROM runtime.task_history
+WHERE labels.datasets IS NOT NULL
+AND NOW() < end_time + INTERVAL '{DATASET_UNAVAILABLE_THRESHOLD_MINUTES}' MINUTE
+AND labels.tags NOT LIKE '%error%'"
```
phillipleblanc: Tags were removed in the previous PR

phillipleblanc (approved):

PR #2739, opened 2024-09-19
Add `spice search` CLI command

lukekim:
## 🗣 Description

Add `spice search` command.
Refactor utils.

## Screenshots

image


Jeadie:
I wonder if instead of solely having a numbered list for each search result, we should also show the primary keys, or the extra attributes.

phillipleblanc (dismissed):

sgrebnov (dismissed):

Jeadie (approved):

PR #2749, opened 2024-09-19
Resolve primary keys for datasets with catalog or schema

Jeadie:
## 🗣 Description
 - Issue in how we resolve `datasets[*].embeddings[*].column_pk` for datasets with non-default catalog or schema.
 - Also fix and return a score for `v1/search`.

### DIFF
--- a/crates/runtime/src/embeddings/vector_search.rs
+++ b/crates/runtime/src/embeddings/vector_search.rs
@@ -100,6 +100,8 @@ impl Display for RetrievalLimit {
     }
 }
 
+static VECTOR_DISTANCE_COLUMN_NAME: &str = "dist";
+
 #[derive(Debug, Clone, JsonSchema, Serialize, Deserialize)]
 #[serde(rename_all = "lowercase")]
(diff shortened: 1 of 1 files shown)


crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -260,7 +285,7 @@ impl VectorSearch {
         let where_str = where_cond.map_or_else(String::new, |cond| format!("WHERE ({cond})"));
 
         let query = format!(
-            "SELECT {projection_str} FROM {tbl} {where_str} ORDER BY array_distance({embedding_column}_embedding, {embedding:?}) LIMIT {n}"
+            "SELECT {projection_str}, sqrt(array_distance({embedding_column}_embedding, {embedding:?})) as {VECTOR_DISTANCE_COLUMN_NAME} FROM {tbl} {where_str} ORDER BY dist LIMIT {n}"
```
Jeadie, line 288: `sqrt()` needed to make distance a true L2 measure.

phillipleblanc (approved):

PR #2759, opened 2024-09-20
Utilize DataConnectorError for MySQL Data Connector Errors

Sevenannn:
## 🗣 Description

Refatcor the MySQL Data Connector Errors to wrap MySQL errors in common DataConnectorError, in order to keep consistency with other data connector errors and for less verbose messages.

- Relevant changes for updating verifying user configurable parameters in Datafusion table provider: https://github.com/datafusion-contrib/datafusion-table-providers/pull/113

Before:
Host invalid:
image
Port Invalid:
image
Database invalid:
image
Username/ Password invalid:
image
SSL mode invalid:
image

After
Port / Host Invalid:
image
Database invalid:
image
User / Password invalid:
image
SSL mode invalid:
image

### DIFF
--- a/crates/runtime/src/dataconnector/mysql.rs
+++ b/crates/runtime/src/dataconnector/mysql.rs
@@ -19,7 +19,7 @@ use async_trait::async_trait;
 use data_components::Read;
 use datafusion::datasource::TableProvider;
 use datafusion_table_providers::mysql::MySQLTableFactory;
-use datafusion_table_providers::sql::db_connection_pool::mysqlpool::MySQLConnectionPool;
+use datafusion_table_providers::sql::db_connection_pool::mysqlpool::{self, MySQLConnectionPool};
 use datafusion_table_providers::sql::db_connection_pool::{
     DbConnectionPool, Error as DbConnectionPoolError,
 };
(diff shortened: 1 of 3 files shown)


crates/runtime/src/dataconnector/mysql.rs
```diff
@@ -80,12 +80,40 @@ impl DataConnectorFactory for MySQLFactory {
...
-            > = Arc::new(
-                MySQLConnectionPool::new(params.to_secret_map())
-                    .await
-                    .context(UnableToCreateMySQLConnectionPoolSnafu)?,
-            );
-
+            > = match MySQLConnectionPool::new(params.to_secret_map()).await {
+                Ok(pool) => Arc::new(pool),
+                Err(error) => match error {
+                    mysqlpool::Error::InvalidUsernameOrPassword { .. } => {
```
sgrebnov, line 86: If mysqlpool::Error::InvalidUsernameOrPassword is not always the same and contain details I would recommend to keep it for troubleshooting reason, otherwise it could be hard to identify the problem. For example if env key is configured incorrectly resulting in empty password

sgrebnov (approved):

crates/runtime/src/dataconnector/mysql.rs
sgrebnov: Can be updated/improved next IMO

Sevenannn: Yeah let me verify whether different wrong configuration of username / password would result in different error messages. IIRC the error messages are the same for wrong password / empty password / wrong username, etc.

Issue #2688, opened 2024-09-16
Enhancement: Microsoft SQL Server (MSSQL) Data Connector


phillipleblanc:
## Goal-State/What/Result

I can use Spice to query and accelerate datasets stored on Microsoft SQL Server.

## Why/Purpose

There is a large community of MS SQL users attempting to integrate with Spice, but the lack of a dedicated connector and no support in the ODBC connector creates a significant barrier.

## By When

**Issue/Spec written and reviewed:** 09/19
**Done-Done:** 09/23
## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### Definition

`mssql:` prefix is used to define MS SQL Server connector

```yaml
datasets:
  - from: mssql:saleslt.customer
    name: customer
    params: 
      mssql_connection_string: ${ secrets:MSSQL_CONNECTION_STRING }
```

### Auth / Configuration

Only `SQL Login` is supported. Other authentication methods such as `Windows Authentication` and `Microsoft Entra ID` are out of scope.

**mssql_connection_string** -  a complete ADO connection string, which could be copied from the MSSQL Server configuration page or manually constructed. `Windows Auth` and `Micrsooft Entra ID` are out of scope. 

**mssql_host** - the hostname or IP address of the Microsoft SQL Server instance. Required if there is no connection_string provided.
**mssql_port** - the host port of the Microsoft SQL Server instance. Optional. Default is `1433`
**mssql_username** - the username for SQL authentication.
**mssql_password** - the password associated with the specified SQL user
**mssql_database** - the name of the specific database to connect. 
**mssql_encrypt** -  whether encryption is required for the connection (true or false). Default is true. Connection string uses true/false
**mssql_trust_server_certificate** - determines whether the server certificate is trusted without validation when encryption is enabled. Default is false. Connection string uses true/false

Example connection string
[...]

PR #2773, opened 2024-09-20
Support for standalone config parameters for MS SQL

sgrebnov:
## 🗣 Description

Support for standalone config parameters in addition to `connection_string` for MSSQL

Configuring connection via standalone parameters is much better UX when you don't have connection string provided so you must construct it yourself. Tools like Azure Data Studio do the same - you can provide pre-configured connection string vs set each param individually.

## 🔨 Related Issues

Part of https://github.com/spiceai/spiceai/issues/2688

### DIFF
--- a/crates/runtime/src/dataconnector/mssql.rs
+++ b/crates/runtime/src/dataconnector/mssql.rs
@@ -51,9 +51,9 @@ pub enum Error {
 const PARAMETERS: &[ParameterSpec] = &[
     ParameterSpec::connector("connection_string").secret(),
     ParameterSpec::connector("username").secret(),
-    ParameterSpec::connector("password"),
-    ParameterSpec::connector("host").secret(),
-    ParameterSpec::connector("port").secret(),
+    ParameterSpec::connector("password").secret(),
+    ParameterSpec::connector("host"),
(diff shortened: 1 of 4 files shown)


Sevenannn (dismissed):

crates/runtime/src/dataconnector/mssql.rs
```diff
@@ -33,16 +35,29 @@ use super::{
...
+    InvalidParamValueError { parameter: String, reason: String },
 }
 
-const PARAMETERS: &[ParameterSpec] = &[ParameterSpec::connector("connection_string")
-    .secret()
-    .required()];
+const PARAMETERS: &[ParameterSpec] = &[
+    ParameterSpec::connector("connection_string").secret(),
+    ParameterSpec::connector("username").secret(),
+    ParameterSpec::connector("password"),
```
Sevenannn: Should the password be secret as well?

sgrebnov: Good catch - thank you, updated

Sevenannn (dismissed):

Sevenannn (approved):

PR #2783, opened 2024-09-21
feat: Add captured_output option for task_history

### LINKED ISSUE #2756: Task: Support `captured_output` in `task_history` spicepod config

peasee:
## 🗣 Description

* Updates `TaskHistory` spicepod config to be more robust, specifying default values all of the time. This is because the configuration values for `task_history` always have defaults, and are always used.
* Fixes a bug in enabling/disabling `task_history` log spans (boolean if was the wrong way around).
* Adds `captured_output` parameter, renames `truncated_output` trace field to `captured_output` and implements output processing for either truncation or none.

## 🔨 Related Issues

* Closes #2756 

## 🤔 Concerns

* Output processing at the span level allows for truncation of any message, but the downside is the row/record "truncated output" still only produces 3 rows as the input text. If those 3 rows are longer than the truncation length, the JSON will still get truncated.


crates/runtime/src/task_history/otel_exporter.rs
```diff
@@ -39,14 +41,124 @@ impl Debug for TaskHistoryExporter {
...
+        match self.captured_output {
+            TaskHistoryCapturedOutput::None => "".into(),
+            TaskHistoryCapturedOutput::Truncated => {
+                let output_str = output.as_ref();
+                if output_str.len() > 1000 {
+                    // TODO: set a config value for truncation length?
+                    output_str[..1000].into()
+                } else {
+                    output
+                }
```
phillipleblanc: Let's keep the current behavior of having each component decide how to truncate, and not apply a global truncation limit.

crates/runtime/src/task_history.rs
```diff
@@ -50,7 +50,7 @@ pub(crate) struct TaskSpan {
 
     pub(crate) task: Arc<str>,
     pub(crate) input: Arc<str>,
-    pub(crate) truncated_output: Option<Arc<str>>,
+    pub(crate) captured_output: Option<Arc<str>>,
```
phillipleblanc, line 53: cc @ewgenius this will need a corresponding change in the cloud

phillipleblanc (dismissed):

crates/runtime/src/task_history.rs
ewgenius: I’ll make sure this is handled

phillipleblanc (approved):

PR #2784, opened 2024-09-21
feat: Add retention options for task_history config

### LINKED ISSUE #2754: Task: Support `retention_period` in `task_history` spicepod config

### LINKED ISSUE #2755: Task: Support `retention_check_interval` in `task_history` spicepod config

peasee:
## 🗣 Description

* Added some more default wrappers for `task_history` config values, because serde doesn't respect the `Default` derive
* Added config options for `retention_period` and `retention_check_interval`
* Use `fundu` to parse duration as seconds from the config input
* Update `TaskHistory` to accept configurable retention check and interval as seconds, powered from the config values.

## 🔨 Related Issues

* Closes #2755
* Closes #2754


peasee:
https://github.com/spiceai/spiceai/pull/2783 should get merged first so this target can reset to `trunk`

Cargo.toml
```diff
@@ -70,6 +70,7 @@ fundu = "2.0.1"
 futures = "0.3.30"
 globset = "0.4.15"
 graph-rs-sdk = { git = "https://github.com/spiceai/graph-rs-sdk", rev = "f8703df260146b313461029d41c4a021306832b8" }
+humantime = "2.1.0"
```
phillipleblanc: We already use `fundu` for this purpose - I'd prefer to reuse that if possible

crates/spicepod/src/component/runtime.rs
```diff
@@ -102,14 +106,33 @@ pub struct TelemetryConfig {
...
+    "truncated".to_string()
+}
+
+fn default_retention_period() -> String {
+    "1d".to_string()
+}
+
+fn default_retention_check_interval() -> String {
+    "30m".to_string()
 }
```
phillipleblanc: nit: These can use an `Arc<str>` and we can just have a single static instance that gets reference cloned, instead of allocating each time these functions are called. see: https://github.com/spiceai/spiceai/blob/trunk/crates/db_connection_pool/src/dbconnection/snowflakeconn.rs#L72

phillipleblanc: Or even better, could these be `&'static str`?

Cargo.toml
phillipleblanc: Oh I see we're already using `humantime` for something else - we should just pick one.

crates/runtime/src/task_history.rs
```diff
@@ -37,6 +37,8 @@ use crate::accelerated_table::{AcceleratedTable, Retention};
 pub mod otel_exporter;
 
 pub const DEFAULT_TASK_HISTORY_TABLE: &str = "task_history";
+pub const DEFAULT_TASK_HISTORY_RETENTION_PERIOD_SECS: u64 = 24 * 60 * 60; // 1 day
```
lukekim: Let's reduce the default to 8 hours.

```suggestion
pub const DEFAULT_TASK_HISTORY_RETENTION_PERIOD_SECS: u64 = 8 * 60 * 60; // 8 hours
```

crates/runtime/src/task_history.rs
```diff
@@ -37,6 +37,8 @@ use crate::accelerated_table::{AcceleratedTable, Retention};
...
+pub const DEFAULT_TASK_HISTORY_RETENTION_CHECK_INTERVAL_SECS: u64 = 30 * 60; // 30 minutes
```
lukekim: Let's reduce this default.

```suggestion
pub const DEFAULT_TASK_HISTORY_RETENTION_CHECK_INTERVAL_SECS: u64 = 16 * 60; // 15 minutes
```

crates/spicepod/src/component/runtime.rs
```diff
@@ -121,14 +144,37 @@ pub enum TaskHistoryCapturedOutput {
...
+            .into()),
         }
     }
+
+    fn retention_value_as_secs(value: &str) -> Result<u64, Box<dyn Error + Send + Sync>> {
+        let duration = humantime::parse_duration(value).map_err(|e| e.to_string())?;
+
+        if duration.as_secs() < TASK_HISTORY_RETENTION_MINIMUM {
+            return Err(format!(
+                "Task history retention period must be at least {TASK_HISTORY_RETENTION_MINIMUM} seconds",
```
lukekim:
```suggestion
                "Task history retention period must be at least {TASK_HISTORY_RETENTION_MINIMUM} seconds. To disable task history, set the property `enabled: false`.",
```

crates/spicepod/src/component/runtime.rs
```diff
@@ -102,14 +106,33 @@ pub struct TelemetryConfig {
...
+fn default_retention_check_interval() -> String {
+    "30m".to_string()
 }
 
 impl Default for TaskHistory {
     fn default() -> Self {
         Self {
             enabled: true,
             captured_output: "truncated".to_string(),
+            retention_period: "1d".to_string(),
```
lukekim:
```suggestion
            retention_period: "8h".to_string(),
```

crates/spicepod/src/component/runtime.rs
```diff
@@ -102,14 +106,33 @@ pub struct TelemetryConfig {
...
+            retention_check_interval: "30m".to_string(),
```
lukekim:
```suggestion
            retention_check_interval: "15m".to_string(),
```

lukekim (dismissed):

Cargo.toml
peasee: I haven't removed `humantime` in this PR, but I've changed to using `fundu` instead because it seems to have more usage in the repo.

phillipleblanc (approved):

PR #2789, opened 2024-09-22
More types support for MS SQL data connector

sgrebnov:
## 🗣 Description

PR adds support for the following types: `DATETIMEOFFSET`, `SMALLMONEY`, `XML`, `BINARY` to MS SQL data connector.

## 🔨 Related Issues

Part of https://github.com/spiceai/spiceai/issues/2688

Limitation that `geography` type is not supported and will be ignored must be added to MS SQL data connector documentation

## 🤔 Concerns

An alternative approach to handling unsupported geography types could be to fail the dataset registration. However, ignoring unsupported columns with a warning provides a better user experience (UX) in my opinion, allowing users to continue working with the dataset.

### DIFF
--- a/crates/data_components/src/mssql/convert.rs
+++ b/crates/data_components/src/mssql/convert.rs
@@ -137,7 +137,23 @@ pub(crate) fn rows_to_arrow(rows: &[Row], schema: &SchemaRef) -> super::Result<R
                         None => builder.append_null(),
                     }
                 }
-
+                ColumnType::Datetime2 => {
+                    let Some(builder) = builder
+                        .as_any_mut()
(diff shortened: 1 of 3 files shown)


crates/data_components/src/mssql/convert.rs
```diff
@@ -215,11 +220,36 @@ pub(crate) fn rows_to_arrow(rows: &[Row], schema: &SchemaRef) -> super::Result<R
                         None => builder.append_null(),
                     }
                 }
+                ColumnType::Xml => {
```
eadgbear, line 239: This is fantastic, thank you. XML column types are pretty frequently used

eadgbear (dismissed):

eadgbear (approved):

PR #2795, opened 2024-09-23
MS SQL connector should ignore all unsupported types

sgrebnov:
## 🗣 Description

Enhance the MS SQL connector to handle unsupported types more gracefully by logging a warning and continuing with dataset registration.

This is potential improvement in case of very rare / exotic types that are not supported by connector. All core data types (Transact-SQL) are supported: https://learn.microsoft.com/en-us/sql/t-sql/data-types/data-types-transact-sql


eadgbear (approved):
LGTM

Issue #2639, opened 2024-09-13
Enhancement: Replace `query_history` with `task_history`


peasee:
## Goal-State/What/Result

The `query_history` table should be removed and its usage replaced with `task_history`

## Why/Purpose

To consolidate metadata history tables.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** 16th Sep, 2024
**Done-Done:** 18th Sep, 2024

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [ ] Tests
- [ ] Documentation
- [ ] Samples/Quickstarts/Tutorials
- [ ] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

`query_history.rs` should be deleted, and any usage of `query_history` updated to `task_history`.

## How/MAP

1. Migrate `trace_query` from `query_history.rs` to `tracker.rs`
2. Delete `query_history.rs`
3. Update `QueryTracker` in `tracker.rs` to call `trace_query`
4. Validate that queries are traced in `task_history`, including failed queries
5. Update logged labels to include `rows_produced` and `results_cache_hit`

### Implementation (issues)

- 
-

### QA

- [ ] Query history appears in `task_history`, including failing queries
- [ ] Query history and metrics continue to report in Grafana

PR #2796, opened 2024-09-23
Queries that cannot be planned show up in task history

### LINKED ISSUE #2742: Task: Queries with missing tables appear in `task_history`
Running a query in REPL for a table that doesn't exist is not populated in `task_history` as a failed query.

Part of  #2639

phillipleblanc:
## 🗣 Description

Queries that cannot be planned (i.e. due to referencing tables that don't exist) will now show up in the task_history table as an error.

Performing a query over Flight is a two-stage process, the first stage is calling `GetFlightInfo` which returns the schema of the query. The second stage is calling `DoGet` which performs the actual query. If the schema couldn't get retrieved because the query couldn't be planned, then the error should still be logged to task_history properly.

## 🔨 Related Issues

Closes #2742
Part of #2639

### DIFF
--- a/crates/runtime/src/datafusion/query/tracker.rs
+++ b/crates/runtime/src/datafusion/query/tracker.rs
@@ -130,6 +130,11 @@ fn trace_query(query_tracker: &QueryTracker, truncated_output: &str) {
     if let Some(error_code) = &query_tracker.error_code {
         tracing::info!(target: "task_history", error_code = %error_code, "labels");
     }
+
+    if let Some(error_message) = &query_tracker.error_message {
+        tracing::error!(target: "task_history", "{error_message}");
+    }
+
     if let Some(query_execution_duration_secs) = &query_tracker.query_execution_duration_secs {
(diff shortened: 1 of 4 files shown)


crates/runtime/src/datafusion/query/tracker.rs
```diff
@@ -130,6 +130,11 @@ fn trace_query(query_tracker: &QueryTracker, truncated_output: &str) {
     if let Some(error_code) = &query_tracker.error_code {
         tracing::info!(target: "task_history", error_code = %error_code, "labels");
     }
+
+    if let Some(error_message) = &query_tracker.error_message {
+        tracing::error!(target: "task_history", "{error_message}");
+    }
```
peasee, line 136: What does this look like in the task history? does it have an associated column?

peasee (approved):

crates/runtime/src/datafusion/query/tracker.rs
phillipleblanc: It shows up in the `error_message` column

PR #2823, opened 2024-09-25
Use User message for health check

Jeadie:
Some models do not like consecutive system messages.


Jeadie:
Raises a good point. If a user provides a system message, we may consider concatting them into one message

sgrebnov (dismissed):

lukekim (approved):

Issue #2301, opened 2024-08-16
Verbose mode for Spice CLI


edmondop:
I am trying to perform the followig command from the spiceAI [documentation](https://docs.spiceai.org/getting-started):

```sh
spice add spiceai/quickstart
Getting Spicepod spiceai/quickstart ...
an error occurred fetching Spicepod 'spiceai/quickstart'
```

However, the information is not sufficient to perform any troubleshoot. I suspect it might be a proxy problem, but I cannot really investigate it easily. I noticed that the CLI commands do not have a `-v` flag that can be used to enable more verbose logging, for example [here](https://github.com/spiceai/spiceai/blob/576d91cc28318d7d34626e00f438335f725785b1/bin/spice/pkg/registry/spicerack.go#L77)

Jeadie:
For both `-v` and `-vv` flags to both `spice` and `spiced`.

For `spiced`, we will alter the `EnvFilter`. Will override environment variable `SPICED_LOG`.
 - `-v`: `task_history=DEBUG,spiced=DEBUG,runtime=DEBUG,secrets=DEBUG,data_components=DEBUG,cache=DEBUG,extensions=DEBUG,spice_cloud=DEBUG,WARN` 
 - `-vv`: `task_history=TRACE,spiced=TRACE,runtime=TRACE,secrets=TRACE,data_components=TRACE,cache=TRACE,extensions=TRACE,spice_cloud=TRACE,WARN` 

For `spice`, will only be used for `spice run` (pass through).

Jeadie:
To increase verbosity of other `spice ...` commands, will require additional work to add underlying telemetry calls.

Jeadie:
We decided on 
- `-v` -> `--verbose`
- `-vv` -> `--very-verbose`

Jeadie:
This is now supported. We will need continued effort to improve the amount of logging at these lower levels

PR #2831, opened 2024-09-26
Add verbosity flags for spiced, spice: `-v`, `-vv`, `--verbose`, `--very-verbose`.

Jeadie:
## 🗣 Description
 - Add support for specifying log verbosity as flags `--very-verbose` `--verbose` and shorthand. 
 - In spiced, this overrides the `EnvFilter` used by tracing to set the log level. These flags take precedence over the `SPICED_LOG` environment variable that currently dictates `EnvFilter`. 
 - In spice, verbosity flags are just passed to spice. 
 - Also, I made `-h` a global flag in spice.

### Examples
```shell
>> spiced --verbose

>> spice run -vv

// --very-verbose overrides the SPICED_LOG
>> SPICED_LOG="spiced=WARN,runtime=WARN,data_components=WARN,cache=WARN" spice run --very-verbose
```

## 🔨 Related Issues
 - #2301 

## 🤔 Concerns
 - Currently -v, -vv in spice are only used to pass into spiced. In spice, our current logging does not support log levels (using https://pkg.go.dev/log). We would have to migrate to https://pkg.go.dev/log/slog@go1.23.1 or similar.

### DIFF
--- a/bin/spice/cmd/add.go
+++ b/bin/spice/cmd/add.go
@@ -124,6 +124,5 @@ spice add spiceai/quickstart
 }
 
 func init() {
-	addCmd.Flags().BoolP("help", "h", false, "Print this help message")
 	RootCmd.AddCommand(addCmd)
 }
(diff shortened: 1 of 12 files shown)


phillipleblanc (approved):
Very clean, thanks!

PR #2845, opened 2024-09-27
fix: Use GitHub search for issues in GraphQL

### LINKED ISSUE #2834: Update the `issues` query to use search with no filter parameters

### LINKED ISSUE #2839: Bug: GraphQL ignores other parameters in pagination field
Take this GraphQL query as an example:

```
search(query:"repo:spiceai/spiceai", first:100, type:issue)
```

It consumes a pagination parameter, as well as some other parameters. This example uses GitHub's GraphQL [search](https://docs.github.com/en/search-github/searching-on-github) function.

The GraphQL connector re-writes this block as:

```
search (first: 100)
```

This causes the query to be invalid because of the dropped parameters. The connector should instead retain any existing parameters in the field.

peasee:
## 🗣 Description

* Uses [search](https://docs.github.com/en/graphql/reference/queries#search) in the GraphQL query for `issues`, as a per-requisite for filter push down.
* Updates the pagination argument parsing to support parsing and retaining parameters that aren't related to pagination.

## 🔨 Related Issues

* Closes #2839 
* Closes #2834


crates/data_components/src/graphql/client.rs
```diff
@@ -428,14 +451,27 @@ impl PaginationParameters {
         let regex =
             Regex::new(&pattern).unwrap_or_else(|_| panic!("Invalid regex query resource pattern"));
 
+        let other_arguments = self
+            .other_arguments
+            .iter()
+            .map(|arg| format!("{}: {}", arg.name, arg.value))
```
Jeadie: nit:
 - Move `format!("{}: {}", arg.name, arg.value` to `FieldArgument`.
 - Move `.collect::<Vec<String>>().join(", ")` to ` PaginationParameters`.

crates/data_components/src/graphql/client.rs
```diff
@@ -914,8 +952,53 @@ mod tests {
...
+                "#,
+                expected: (
+                    Some(PaginationParameters {
+                        resource_name: "paginatedUsers".to_owned(),
+                        pagination_argument: super::PaginationArgument::First(2),
                         page_info_path: Some("/paginatedUsers/pageInfo".to_owned()),
+                        other_arguments: vec![
+                            super::FieldArgument {
+                                name: "some_field".to_owned(),
+                                value: "\"value\"".to_owned(),
```
Jeadie: Do we want to be persisting these as `"\"value\""`, or `"value"`?

peasee: `\"` unwraps to `"` anyway. I should probably use `r#""` to make it clearer.

phillipleblanc (approved):

Issue #2719, opened 2024-09-17
Enhancement: Embedding and Vector Search Chunking


Jeadie:
## Goal-State/What/Result

Compute, compare and store several embeddings per row, per column for a dataset by intelligently chunking the content. 

## Why/Purpose

Chunking long data to create several embedding is a common strategy to improve vector search quality. 

## By When

**Issue/Spec written and reviewed:** 25th September
**Done-Done:** v1.0-rc.2

## Done-Done

- [ ] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] UX Review
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Documentation
- [ ] Samples/Quickstarts/Tutorials
- [ ] Release Notes

## The Algorithm

- [ ] Every requirement questioned?
- [ ] Delete (Scope) any part you can.
- [ ] Simplify.
- [ ] Break down into smaller iterations/milestones.
- [ ] Opportunities for automation.

### Spicepod UX
 - User's should be able to automatically enable chunking of the embedding column for their dataset.  
     - Fortunate rust library https://github.com/benbrandt/text-splitter  
 - Sane defaults for predominately text based chunking, but then can infer better defaults from the dataset.
     - E.g. if `params.file_format` provides a hint, there are better chunkers in [benbrandt/text-splitter](https://github.com/benbrandt/text-splitter)

Default options:
```yaml
datasets: 
  - from: github:github.com/spiceai/spiceai/files/trunk
    name: docs
    params:
      include: docs/**/*.md
    embeddings:
      - column: content
        use: embedding_model_name
        chunking:
          enabled: true
```

Overrides/schema
```yaml
embeddings:
  - column: string
    use: string
   chunking:
     enabled: bool # Optional, default false
     desired_chunk_size: usize # Optional, default to maximum (number of tokens)
     overlap_size: usize # Optional, default 0. (number of tokens)
     trim: bool, # optional, default true
```

### SQL UX
 - when `.datasets[*].embeddings[*].chunking.enabled == true`, the `{embedding_column}_embedding` will be of type `List(FixedSizeList(Float))` (currently, and if false, `FixedSizeList(Float)`). 
 - The `embedding_column` will not be chunked. Instead we need auxiliary data structure. Either 
[...]

PR #2854, opened 2024-09-27
Support chunking dataset embeddings

Jeadie:
## 🗣 Description
 - Support the ability to specify an embedding column should have its data chunked.
 - Can be used in `POST /v1/search`. 
 - See #2719 for UX.
 - This PR **does not** include (coming in a follow up):
     -  Embedding model specific chunkers/sizers

## 🔨 Related Issues
 - #2719 

 - TODO

## 🤔 Concerns
 - Performance is still very bad.


crates/spicepod/src/component/embeddings.rs
```diff
@@ -137,6 +137,22 @@ impl Display for EmbeddingPrefix {
...
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[cfg_attr(feature = "schemars", derive(JsonSchema))]
+pub struct EmbeddingChunkConfig {
+    #[serde(default)]
+    pub enabled: bool,
+
+    #[serde(default)]
+    pub desired_chunk_size: usize,
```
lukekim:
```suggestion
    pub chunk_target_size: usize,
```

peasee (approved):

Issue #2850, opened 2024-09-27
Bug: GitHub/GraphQL connector does not respect LIMIT


peasee:
## Describe the bug

When running a query with a LIMIT in the GitHub/GraphQL connector, it does not respect it and instead returns all results:

```
sql> select count(*) from spiceai.issues
+----------+
| count(*) |
+----------+
| 931      |
+----------+

Time: 17.888681966 seconds. 1 rows.
sql> select count(*) from spiceai.issues limit 200
+----------+
| count(*) |
+----------+
| 931      |
+----------+

Time: 17.182362023 seconds. 1 rows.
```

## To Reproduce

1. Create a spicepod using the GitHub issues connector
2. Run a query with a limit
4. Observe unlimited rows

## Expected behavior

The rows should be limited as expected

### Spicepod

```yaml
version: v1beta1
kind: Spicepod
name: spicepod-graphql-spicehq-decisions

datasets:
  - from: github:github.com/spiceai/spiceai/issues
    name: spiceai.issues
    params:
      github_token: ${secrets:GITHUB_TOKEN}
```

**Have you tried this on the latest `trunk` branch?**

- [x] Yes
- [ ] No

**If possible, run `spiced` with DEBUG log level**
By setting the environment variable `SPICED_LOG="task_history=DEBUG,spiced=DEBUG,runtime=DEBUG,secrets=DEBUG,data_components=DEBUG,cache=DEBUG,extensions=DEBUG,spice_cloud=DEBUG"`

Jeadie:
Using 
```yaml
datasets: 
  - name: me
    from: graphql:https://api.github.com/graphql
    params:
      graphql_auth_token: ${env:GITHUB_TOKEN}
      unnest_depth: 1
      graphql_query: |
        {
          repository(owner: "jeadie", name: "my-first-app") {
            defaultBranchRef {
                target {
                    ... on Commit {
                        history(first: 100) {
                            pageInfo {
                                hasNextPage
                                endCursor
                            }
                            nodes {
                                message
                                message_head_line: messageHeadline
                                message_body: messageBody
                                sha: oid
                                additions
                                deletions
                                id
                                committed_date: committedDate
                                authorName: author {
                                    author_name: name
                                }
                                authorEmail: author {
                                    author_email: email
                                }
                            }
                        }
                    }
                }
            }
          }
        }
```

Running
```shell
[...]

PR #2867, opened 2024-09-30
Change `BytesProcessedRule` to be an optimizer rather than an analyzer rule

phillipleblanc:
## 🗣 Description

DataFusion has two kinds of rules that can rewrite LogicalPlans, `AnalyzerRule`s and `OptimizerRules`. `AnalyzerRule`s can be used to change the semantic meaning of the query, while `OptimizerRule`s are not allowed to change the semantic meaning of the query, but to compute it in some "better" way.

The rule `PushDownLimit` is responsible for pushing down the limit to a `TableScan` node, but it doesn't support the ability to push down limits beyond extension nodes. #2850 showed that this caused an issue for us, since AnalyzerRules always run before OptimizerRules - and we inserted the `BytesProcessedNode` extension into the plan, and thus it wasn't able to properly push down the limit.

By making the rule to add the `BytesProcessedNode` an `OptimizerRule`, we can add it to the end of the list which will ensure the `PushDownLimit` node has a chance to run first, which means by the time the `BytesProcessedRule` runs, the limit is already applied.

## 🔨 Related Issues

#2850

### DIFF
--- a/crates/runtime/src/datafusion/extension/bytes_processed.rs
+++ b/crates/runtime/src/datafusion/extension/bytes_processed.rs
@@ -21,11 +21,10 @@ use datafusion::{
         tree_node::{Transformed, TreeNode},
         DFSchemaRef,
     },
-    config::ConfigOptions,
     error::Result,
     execution::{SendableRecordBatchStream, TaskContext},
     logical_expr::{Extension, LogicalPlan, UserDefinedLogicalNodeCore},
-    optimizer::AnalyzerRule,
+    optimizer::{OptimizerConfig, OptimizerRule},
     physical_plan::{
         stream::RecordBatchStreamAdapter, DisplayAs, DisplayFormatType, ExecutionPlan,
     },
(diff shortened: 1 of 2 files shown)


Jeadie (dismissed):

sgrebnov (dismissed):

crates/runtime/src/datafusion/extension/bytes_processed.rs
```diff
@@ -71,18 +90,24 @@ impl AnalyzerRule for BytesProcessedAnalyzerRule {
                     Ok(Transformed::no(LogicalPlan::Extension(extension)))
                 }
             }
+            LogicalPlan::TableScan(table_scan) => {
+                let bytes_processed = BytesProcessedNode::new(LogicalPlan::TableScan(table_scan));
+                let ext_node = Extension {
```
Jeadie, line 94: We should comment the rationale/need for this, for posterity.

Jeadie (dismissed):

PR #2869, opened 2024-09-30
Don't run E2E or PR tests on documentation

Jeadie:
## 🗣 Description
 - Certain PRs, especially those during a release, don't touch the code. We should not run E2E or PR tests on these PRs.
 - For the required checks, we run the workflow and skip jobs. This should still pass the status check on pull requests, see [status checks](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#checks). 
 
 > _Note: The Checks tab only gets populated for pull requests if you set up checks, not commit statuses, for the repository._


peasee (dismissed):

peasee (commented):
Actually, we should be careful with this. The PR tests are required tests, so if they don't run the PRs will get locked and can't merge without admin override.

peasee (approved):

phillipleblanc (approved):

Issue #2453, opened 2024-09-02
Enhancement: filter predicate push down support for GraphQL GitHub Connector


sgrebnov:
## Goal-State/What/Result

 Support filter predicate push down for GraphQL-based Github Connector tables: pull requests, issues, commits

## Why/Purpose

To reduce the amount of data transferred in large queries, and improve response times.

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** 26th Sep, 2024
**Done-Done:** 11th October, 2024

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

GitHub connector queries should support injecting filter parameters into the GitHub GraphQL search.

## How/MAP

GitHub connector queries should first be updated to use [search](https://docs.github.com/en/graphql/reference/queries#search).

Once updated, columns that support filter expressions for exact values and values greater/less than should be identified. For example, in issues the `title`, `body` and `author` columns support exact matches while `updated_at` supports both exact and range based matches.

For filters targeting these columns, the search value in the GitHub GraphQL query should be updated to reflect the filter. For example, `WHERE author = 'peasee'` should apply a filter which results in a GraphQL search parameter like `author:peasee`.

For queries with filters that explicitly match single columns (e.g. `WHERE author = 'peasee' AND state = 'closed'`) we should pass `Exact` filter push down as it should return the exact results from the query (`author:peasee state:closed`).

For queries with complex filters, or queries that match multiple values for a column (e.g `WHERE assignee IN ('peasee', 'phillipleblanc')`) we should pass `Inexact` filter push down as the query may not return the exact results. This SQL query would be rewritten as search parameters like `assignee:peasee assignee:phillipleblance`, because assignee search parameters don't support multiple values. This can return issues with one or both assignees, which requires the `Inexact` filter parameter.

### Implementation (issues)

- [x] #2833
- [x] #2834
- [x] #2836
- [x] #2837
- [x] #2838 
[...]

PR #2897, opened 2024-09-30
fix: Remove login column rename in pulls and update Optimizer

peasee:
## 🗣 Description

* Removes a breaking change for renaming `login` to `author` in the GitHub pulls table
* Because of the column name difference, adds logic to remap column names to search column names

## 🔨 Related Issues

* Part of #2453

### DIFF
--- a/crates/data_components/src/graphql/client.rs
+++ b/crates/data_components/src/graphql/client.rs
@@ -705,7 +705,6 @@ impl GraphQLClient {
         cursor: Option<String>,
     ) -> Result<GraphQLQueryResult> {
         let query_string = query.to_string(limit, cursor);
-        println!("{query_string}");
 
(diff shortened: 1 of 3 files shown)


crates/data_components/src/graphql/client.rs
```diff
@@ -705,6 +705,7 @@ impl GraphQLClient {
         cursor: Option<String>,
     ) -> Result<GraphQLQueryResult> {
         let query_string = query.to_string(limit, cursor);
+        println!("{query_string}");
```
lukekim: Is this expected?

lukekim (dismissed):

crates/data_components/src/graphql/client.rs
peasee: nope, oops! removed them 😄

Jeadie (approved):

Issue #2885, opened 2024-09-30
Bug: dynamically enabling DuckDB acceleration fails with `DuckDBError: IO Error: Cannot open file`


sgrebnov:
## Describe the bug

Follow DuckDB Accelerator Quickstart:https://github.com/spiceai/quickstarts/blob/trunk/duckdb/accelerator/README.md

1. start spice on non-accelerated dataset
1. update acceleration params
1. got an error in logs:

>2024-09-30T18:14:06.689910Z  INFO runtime: Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), results cache enabled.
2024-09-30T18:14:16.300819Z  INFO runtime: Unloaded dataset taxi_trips
2024-09-30T18:14:17.580574Z  WARN runtime: Unable to attach data connector s3: Unable to create dataset acceleration: Acceleration creation failed: Unable to create table: External error: DbConnectionPoolError: DuckDBError: IO Error: Cannot open file "< truncated ... >/accelerator-duckdb/.spice/data/accelerated_duckdb.db": No such file or directory

Sevenannn:
- Close by: https://github.com/spiceai/spiceai/pull/2923

PR #2923, opened 2024-10-01
Initialize file based accelerators on dataset reload

Sevenannn:
## 🗣 Description

When dynamically enabling file based acceleration, the file based accelerator need to be initiated before registering dataset, to ensure the dataset path exist and checkpoints are set.

This solves the path not exist problem for dynamically enabling sqlite & duckdb file-based acceleration

## 🔨 Related Issues

- https://github.com/spiceai/spiceai/issues/2885


crates/runtime/src/lib.rs
```diff
@@ -1046,6 +1046,15 @@ impl Runtime {
 
                 self.remove_dataset(&ds).await;
 
+                // Initialize file mode accelerator when reloading with file mode acceleration
+                // Fail when the length of successfully initiated dataset is not 1
+                if ds.is_file_accelerated() {
+                    let datasets = self.initialize_accelerators(&[Arc::clone(&ds)]).await;
+                    if datasets.len() != 1 {
```
peasee:
```suggestion
                    if datasets.is_empty() {
```

**nit:** The vec can only ever be 1 or 0 in this case. `.is_empty()` is more succinct syntax.

peasee (dismissed):

sgrebnov (approved):

PR #2932, opened 2024-10-02
Use Cosine Similarity in vector search

Jeadie:
## 🗣 Description
 - Copying similar to Datafusion's in-built [array_distance](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#array-distance), define and use cosine similarity for vector search. 
 - Cosine similarity better for
     1. High dimensionality
     2. Invariant to magnitide.
 - Structure of UDF kepts as close to `array_distance` so we can possibly upstream.


crates/runtime/src/embeddings/cosine_distance.rs
```diff
@@ -0,0 +1,259 @@
...
+        let mut result = Vec::new();
+        for arg_type in arg_types {
+            match arg_type {
+                List(_) | LargeList(_) | FixedSizeList(_, _) => {
+                    result.push(coerced_fixed_size_list_to_list(arg_type));
+                }
+                _ => {
+                    return exec_err!(
+                    "The cosine_distance function can only accept List/LargeList/FixedSizeList."
+                )
```
phillipleblanc:
```suggestion
                        "The cosine_distance function can only accept List/LargeList/FixedSizeList."
                    )
```

phillipleblanc (dismissed):

crates/runtime/src/embeddings/cosine_distance.rs
Jeadie: Odd `cargo fmt` would be okay with this.

lukekim (approved):

PR #2940, opened 2024-10-03
Don't assume first vector or content will be non-null/zero

### LINKED ISSUE #2933: Bug with refresh_sql acceleration and chunked embedding column
```yaml
datasets:
  - from: github:github.com/apache/datafusion/pulls
    name: datafusion_pulls
    description: Datafusion docs at https://datafusion.apache.org
    metadata:
      instructions: |
        This dataset provides information on Apache Datafusion recent pull requests.
    params:
      github_token: ${secrets:GITHUB_TOKEN}
    acceleration:
      enabled: true
      refresh_check_interval: 15m
      refresh_jitter_enabled: true
      refresh_jitter_max: 1m
      refresh_sql: |
        SELECT * FROM datafusion_pulls LIMIT 300
    embeddings:
      - column: body
        use: openai_embeddings
        chunking:
          enabled: true
          target_chunk_size: 1024
```

Jeadie:
## 🗣 Description
 - Possible issues occur for some edge cases:
     - If length of embedding column of last row is empty, we'd get an index out of bound when computing length.
     - Don't use vector data to find embedding length. Use `.size()` from `&dyn Embed`.

## 🔨 Related Issues
 - Fixes #2933


crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -205,13 +210,14 @@ fn construct_record_batch(
     embedding_cols: &HashMap<String, ArrayRef>,
 ) -> Result<RecordBatch, ArrowError> {
     let cols: Vec<ArrayRef> = projected_schema
-        .flattened_fields()
```
Jeadie, line 208: Somehow this only caused bugs in some chunking setups/datasets.

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -365,7 +374,11 @@ async fn get_vectors_with_chunker(
         values.append_slice(&inner); // I believe this is a clone under the hood.
 
         // Get the length of the last chunk
-        let last_chunk_length = embedded_data.as_slice()[curr + chunks_in_row - 1].len();
```
Jeadie, line 368: if `curr + chunks_in_row == 0` this `[curr + chunks_in_row - 1]` will be -18446744073709551615 and an index out of range.

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -365,7 +374,11 @@ async fn get_vectors_with_chunker(
...
+        let last_chunk_length = embedded_data
+            .as_slice()
+            .get(curr + chunks_in_row - 1)
```
eadgbear, line 379: There's no chance that a chunk would be empty, right? Not sure if this would screw up the math

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -183,17 +183,22 @@ fn to_sendable_stream(
...
                         Err(e) => {
+                            tracing::debug!("Error when getting embedding columns: {:?}", e);
                             yield Err(DataFusionError::Internal(e.to_string()));
 
                         },
                     };
                 },
-                Err(e) => yield Err(e),
+                Err(e) => {
+                    tracing::debug!("Error in underlying base stream");
```
sgrebnov:
```suggestion
                    tracing::debug!("Error in underlying base stream: {e:?}");
```

crates/runtime/src/embeddings/execution_plan.rs
sgrebnov: This looks very dangerous :)

Jeadie: The last chunk could be 0, but its not an issue. The reason we calculate it is that we cannot infer the last index from the starting offsets. Take this example
```rust
// Raw content: "this is good code"
// Offsets (from `StringArray`): ['this', ' is', ' good', ' code']
// Offsets (from `StringArray`): [0, 4, 7, 12]
assert_eq!(chunk_offsets_to_col_values(
    &[0, 4, 7, 12], 5),
    [[0, 4], [4, 7], [7, 12], [12, 17]]
)
```
Without the last length `3`, we cannot know what the second value in last offset pair `[12, 15]`  should be.

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -365,7 +374,11 @@ async fn get_vectors_with_chunker(
         values.append_slice(&inner); // I believe this is a clone under the hood.
 
         // Get the length of the last chunk
-        let last_chunk_length = embedded_data.as_slice()[curr + chunks_in_row - 1].len();
+        let last_chunk_length = embedded_data
```
phillipleblanc, line 377: I'd like to see some tests to cover this logic. I can see that the `.get` method on a slice will return `None` if it goes out of bounds: https://doc.rust-lang.org/std/primitive.slice.html#method.get so it should not cause a panic, but would return 0. Is it valid for `last_chunk_length` to be 0 in this context?

Jeadie: Is okay for it to be zero, see [above](https://github.com/spiceai/spiceai/pull/2940#discussion_r1786948207).

phillipleblanc (approved):

PR #2943, opened 2024-10-03
Add Markdown aware chunker when `params.file_format: md`.

Jeadie:
## 🗣 Description
 - When using chunking, instead of preserving semantics based on plaintext, for markdown files, use markdown semantics.
 - This is practically implemented in the choice of `text_splitter` provided to `EmbeddingTable`. Based on `params.file_format`, use markdown formatter where appropriate.
 - This gives future viability to other formats, e.g. code.

### DIFF
--- a/crates/llms/src/chunking/mod.rs
+++ b/crates/llms/src/chunking/mod.rs
@@ -11,12 +11,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use std::sync::Arc;
-
 use text_splitter::{ChunkCapacity, ChunkConfig};
 
 #[derive(Debug, Clone)]
-pub struct ChunkingConfig {
+pub struct ChunkingConfig<'a> {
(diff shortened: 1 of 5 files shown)


crates/llms/src/chunking/mod.rs
```diff
@@ -38,37 +38,38 @@ pub trait Chunker: Sync + Send {
     }
 }
```
Jeadie, line 40: The shared trait behind these `text_splitter::Splitter` is parameterised as follows
```rust
pub trait Splitter<Sizer>
where
    Sizer: ChunkSizer,
{
    type Level: SemanticLevel;
```

Implementations of `Splitter` do not share a `SemanticLevel` in common, making it hard to use
```rust
pub struct CharacterSplittingChunker {
    splitter: Arc<dyn text_splitter::Splitter>
}
```

phillipleblanc (approved):

PR #2951, opened 2024-10-03
Update OpenAI client and use new request fields

Jeadie:
## 🗣 Description
 - Support new request parameters released by OpenAI.
 - Log `request.metadata` into `runtime.task_history`.

## 🔨 Related Issues
 - Relevant API changes from https://github.com/spiceai/async-openai/pull/1/

## 🤔 Concerns
 - New fields are optional/backwards compatible.

### DIFF
--- a/crates/runtime/src/model/chat.rs
+++ b/crates/runtime/src/model/chat.rs
@@ -357,6 +357,10 @@ impl Chat for ChatWrapper {
         let req = self.prepare_req(req)?;
         let span = tracing::span!(target: "task_history", tracing::Level::INFO, "ai_completion", stream=false, model = %req.model, input = %serde_json::to_string(&req).unwrap_or_default(), "labels");
 
+        if let Some(metadata) = &req.metadata {
+            tracing::info!(target: "task_history", metadata = %metadata, "labels");
+        }
+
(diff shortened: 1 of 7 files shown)


crates/runtime/src/model/chat.rs
```diff
@@ -351,6 +357,10 @@ impl Chat for ChatWrapper {
         let req = self.prepare_req(req)?;
         let span = tracing::span!(target: "task_history", tracing::Level::INFO, "ai_completion", stream=false, model = %req.model, input = %serde_json::to_string(&req).unwrap_or_default(), "labels");
 
+        if let Some(metadata) = &req.metadata {
+            tracing::info!(target: "task_history", metadata = %metadata, "labels");
```
phillipleblanc, line 361: Nice!

phillipleblanc: Should we also add `store` if it exists?

phillipleblanc (approved):

crates/runtime/src/model/chat.rs
Jeadie: I don't think its as interesting. `metadata` was explicitly added to help filter chat completions in OpenAI dashboards.

PR #2952, opened 2024-10-03
Add error message when spicepod has embeddings or models without '--features models'

Jeadie:
## 🗣 Description
 - If users build/install `spiced` without model support, log error when starting runtime.

### Example
```shell
>> spiced
2024-10-03T23:15:05.192334Z ERROR runtime: Cannot load embedding models without the 'models' feature enabled. To enable model support, either:
  1) `spice install ai`
  2) Build spiced binary with flag `--features models`.
```


crates/runtime/src/lib.rs
```diff
@@ -450,8 +450,11 @@ impl Runtime {
     pub async fn load_components(&self) {
         self.start_extensions().await;
 
-        #[cfg(feature = "models")]
-        self.load_embeddings().await; // Must be loaded before datasets
+        if cfg!(feature = "models") {
+            self.load_embeddings().await; // Must be loaded before datasets
+        } else {
+            tracing::error!("Cannot load embedding models without the 'models' feature enabled. {ENABLE_MODEL_SUPPORT_MESSAGE}");
```
phillipleblanc: Won't this always log an error when models isn't enabled?

crates/runtime/src/lib.rs
```diff
@@ -466,6 +469,8 @@ impl Runtime {
 
         if cfg!(feature = "models") {
             futures.push(Box::pin(self.load_models()));
+        } else {
+            tracing::error!("Cannot load models without the 'models' feature enabled. {ENABLE_MODEL_SUPPORT_MESSAGE}");
```
phillipleblanc: Similarly, this will always log. We should only log if the user attempts to use a functionality that needs the `models` feature.

phillipleblanc (approved):

PR #3026, opened 2024-10-08
docs: Add RC connector criteria

peasee:
## 🗣 Description

* Adds criteria definition for release candidate level connectors.


docs/criteria/definitions.md
```diff
@@ -0,0 +1,43 @@
...
+
+## RC
+
+Acronym for "Release Candidate". Identifies a version that is eligible for general/stable release.
+
+## Major Bug
+
+A major bug is classified as a bug that:
+
+- Renders the accelerator completely inoperable (i.e. all queries on the accelerator fail, loading fails, etc), or;
```
phillipleblanc: This isn't specific to accelerators anymore

docs/criteria/connectors.md/rc.md
```diff
@@ -0,0 +1,101 @@
...
+
+#### [Relational Connectors](../definitions.md)
+
+- [ ] Data is streamed when reading/performing queries from this connector
+- [ ] The connector supports full federation within a single dataset (e.g. `select * from my_dataset`)
+- [ ] The connector supports federation push down across multiple datasets within the same connection source (e.g. `select * from first_dataset, second_dataset`)
+
+#### [Non-Relational Connectors](../definitions.md)
+
+- [ ] The connector supports some level of filter federation within a single dataset (e.g. `select * from my_dataset where id = 1`)
```
phillipleblanc: `some level` isn't precise enough. What guidelines should be used to determine what level is good enough?

eadgbear (dismissed):

docs/criteria/connectors.md/rc.md
```diff
@@ -0,0 +1,110 @@
...
+#### General
+
+- [ ] Integration tests to cover simple use cases based on the connector type, e.g. a Spicepod sourcing a file from an FTP server, reading latest GitHub issues, etc.
+
+#### TPC-H
+
+- [ ] End-to-end test to cover connecting to TPC-H SF1 for the connector type and benchmarking TPC-H queries (official and simple).
+  - [ ] Connectors should run all queries with no [Major Bugs](../definitions.md).
+- [ ] A test script exists that can load TPC-H data at the [designated scale factor](#test-coverage) into this connector.
+- [ ] The connector can load TPC-H at the [designated scale factor](#test-coverage), and can run all queries with no [Major Bugs](../definitions.md).
```
Sevenannn: Do we need to specify the correctness requirement for TPCH / TPCDS benchmarks

phillipleblanc (dismissed):

ewgenius (dismissed):

sgrebnov (dismissed):

docs/criteria/connectors/rc.md
```diff
@@ -0,0 +1,143 @@
...
+## RC Release Criteria
+
+Some connectors impose different testing requirements, as by their nature they support different feature requirements.
+
+Some connectors are unable to support TPC derived test packages due to their nature (e.g. GitHub Issues).
+These connectors are exempt from running the TPC derived test packages, and rely instead on their general integration test.
+
+This table defines the required features and/or tests for each connector:
+
+| Connector     | [TPC-H Derived Tests (Scale Factor)](#tpc-h) | [TPC-DS Derived Tests (Scale Factor)](#tpc-ds) | [Federation](#federation) | [Data Correctness](#data-correctness) | [Streaming](#streaming) | [Native Schema Inference](#schema-inference) |
```
Jeadie, line 40:
```suggestion
| Connector     | [TPC-H Derived Tests (Scale Factor)](#tpc-h) | [TPC-DS Derived Tests (Scale Factor)](#tpc-ds) | [Federation](#federation) | [TPC Data Correctness](#data-correctness) | [Streaming](#streaming) | [Native Schema Inference](#schema-inference) |
```

docs/criteria/connectors/rc.md
```diff
@@ -0,0 +1,143 @@
...
+
+### Conditional Criteria
+
+The following features/tests are dependent on the required features/tests for the specified connector, from the [Connector criteria table](#rc-release-criteria)
+
+#### Streaming
+
+- [ ] Data is streamed when reading/performing queries from this connector.
+
+#### Schema Inference
```
Jeadie, line 89: I think an important alternate condition for this (for some connectors with a _"fixed"_ schema), is that schema inference is determined natively OR the schema requires no inference.

docs/criteria/connectors/rc.md
peasee: I would prefer to leave that heading as "Data Correctness". Data Correctness could be expanded to more things than just TPC in the future, and the intention is that although it runs on TPC it tests for data correctness as a whole.

docs/criteria/connectors/rc.md
```diff
@@ -0,0 +1,143 @@
...
+This table defines the required features and/or tests for each connector:
+
+| Connector     | [TPC-H Derived Tests (Scale Factor)](#tpc-h) | [TPC-DS Derived Tests (Scale Factor)](#tpc-ds) | [Federation](#federation) | [Data Correctness](#data-correctness) | [Streaming](#streaming) | [Native Schema Inference](#schema-inference) |
+| ------------- | - | - | - | - | - | - |
+| Clickhouse    | ✅ (100) | ✅ (100) | ✅ | ✅ | ✅ | ✅ |
+| Databricks    | ✅ (100) | ✅ (100) | ✅ | ✅ | ✅ | ✅ |
+| Delta Lake    | ✅ (100) | ✅ (100) | ✅ | ✅ | ✅ | ✅ |
+| Dremio        | ✅ (100) | ✅ (100) | ✅ | ✅ | ✅ | ✅ |
+| File          | ✅ (1) | ✅ (0.5) | ❌ | ❌ | ❌ | ❌ |
+| FTP/SFTP      | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
```
Jeadie, line 47: I think some of these support streaming, in the approximate definition of _"not all rows of data needs to be retrieved before being provided to a downstream component of the system reading the data"_ (still separate from whether the connector supports `LIMIT` pushdown). In this case, I think the following support streaming:
 - File/S3/Azure BlobFS: For document table formats, each file is streamed individually, and the _"list metadata"_ is [streamed](https://github.com/spiceai/spiceai/blob/6534930313a0fe6cba1e4fa96bdbfcdc4259b529/crates/data_components/src/object/text.rs#L265). For tabular data (e.g), it is determined per file format, but generally i believe they also stream the rows within a file, e.g. [csv](https://github.com/apache/datafusion/blob/1b10c9f89eac127507fe7a137ff7c40534f7ca9a/datafusion/core/src/datasource/file_format/csv.rs#L145-L150). 
 - Github/Sharepoint: As document table only, they are similar to above. They stream the metadata, and stream each row (e.g. a full sharepoint file).

docs/criteria/connectors/rc.md
Jeadie: Fair point. My idea primarily was to avoid current confusion (i.e. data correctness currently only means TPC data correct).

Jeadie (approved):

eadgbear (approved):

PR #3032, opened 2024-10-08
Use 'Accept' header for `/v1/sql/` and `/v1/nsql`

Jeadie:
## 🗣 Description
 - Introduce the use of the 'Accept' header for `/v1/sql/` and `/v1/nsql` endpoints to allow clients to specify the format of the response data.
 - Default format is still JSON.

### Example
```shell
>>> curl -XPOST "http://localhost:8090/v1/sql" -H 'Accept: text/plain' \
    --data "select status, quarter, month from fall limit 5;"
+---------+---------+-------+
| status  | quarter | month |
+---------+---------+-------+
| Shipped | 1       | 2     |
| Shipped | 2       | 5     |
| Shipped | 3       | 7     |
| Shipped | 3       | 8     |
| Shipped | 4       | 10    |
+---------+---------+-------+%

>>> curl -XPOST "http://localhost:8090/v1/nsql" -H 'Content-Type: application/json' \
    -H 'Accept: text/csv' \
    --data '{"model": "me", "query": "How many sales did i make?"}' 
total_sales
2823
```


crates/llms/src/openai/mod.rs
```diff
@@ -87,13 +87,19 @@ impl Openai {
             model,
         }
     }
+
+    /// Returns true if the `OpenAI` compatible model supports [structured outputs](https://platform.openai.com/docs/guides/structured-outputs/).
```
Jeadie, line 91: Found bug, only OpenAI models gpt-4o and later are suppored.

crates/runtime/src/http/v1/mod.rs
```diff
@@ -52,6 +54,33 @@ pub enum Format {
...
 }
 
+#[derive(Default, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+/// The various formats that the Arrow data can be converted and returned from HTTP requests.
+pub enum ArrowFormat {
+    #[default]
+    Json,
+    Csv,
+    Shell,
```
peasee:
```suggestion
    Text,
```

**nit:** `text/plain` is `Text`, not `Shell`?

crates/runtime/src/http/v1/mod.rs
```diff
@@ -130,5 +157,61 @@ pub async fn sql_to_http_response(df: Arc<DataFusion>, sql: &str, nsql: Option<&
...
+    (StatusCode::OK, headers, body).into_response()
+}
+
+/// Converts a vector of `RecordBatch` to a JSON string.
+fn arrow_to_json(data: &[RecordBatch]) -> Result<String, Box<dyn std::error::Error>> {
+    let buf = Vec::new();
+    let mut writer = arrow_json::ArrayWriter::new(buf);
+
+    if let Err(e) = writer.write_batches(data.iter().collect::<Vec<&RecordBatch>>().as_slice()) {
+        tracing::debug!("Error converting results to JSON: {e}");
```
peasee: **question:** Why debug print these if we're returning the error anyway?

crates/runtime/src/http/v1/mod.rs
```diff
@@ -130,5 +157,61 @@ pub async fn sql_to_http_response(df: Arc<DataFusion>, sql: &str, nsql: Option<&
...
+        Err(e) => {
+            tracing::debug!("Error converting CSV buffer to string: {e}");
+            Err(Box::new(e))
+        }
+    }
+}
+
+/// Converts a vector of `RecordBatch` to a pretty formatted string.
+/// This is equivalent to [`datafusion::dataframe::DataFrame::show`].
+fn arrow_to_shell(data: &[RecordBatch]) -> Result<String, Box<dyn std::error::Error>> {
```
peasee:
```suggestion
fn arrow_to_text(data: &[RecordBatch]) -> Result<String, Box<dyn std::error::Error>> {
```

crates/llms/src/openai/mod.rs
```diff
@@ -87,13 +87,19 @@ impl Openai {
...
+    /// This is only supported for GPT-4o models from `OpenAI` (i.e not any other compatible servers).
+    fn supports_structured_output(&self) -> bool {
+        self.client.config().api_base() == OPENAI_API_BASE && self.model.starts_with("gpt-4o")
```
phillipleblanc, line 94: It seems like this should come from the model parameters, not be hard coded. It means that we can't be flexible to support new models without recompiling.

crates/runtime/src/http/v1/mod.rs
Jeadie: decided to go for `Plain` since Csv is `text/csv`

crates/llms/src/openai/mod.rs
Jeadie: This is true. I contemplated using /v1/models to get this, but it doesn't have the details. Right now, only Openai and no other providers supports structured output. 

I would really like users to be unaware of these details.

phillipleblanc (approved):

PR #3056, opened 2024-10-08
refactor: Use is modifier for checking GitHub state filter

peasee:
## 🗣 Description

* Changes the query parameter for filter push down on state to `is:merged`, `is:closed` instead of `state:closed`. `state:closed` returns both closed and merged PRs, and does not support `state:merged`, while `is:merged` returns only merged PRs. A slight push down performance improvement.

### DIFF
--- a/crates/runtime/src/dataconnector/github/mod.rs
+++ b/crates/runtime/src/dataconnector/github/mod.rs
@@ -647,16 +647,7 @@ pub(crate) fn filter_pushdown(expr: &Expr) -> FilterPushdownResult {
             let value = match value {
                 ScalarValue::Utf8(Some(v)) => {
                     if column.name == "state" {
-                        // "state" in GitHub search is odd
-                        // it returns values for CLOSED, MERGED and OPEN
-                        // but you can only search with either closed or open (in lowercase as well)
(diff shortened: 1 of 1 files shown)


crates/runtime/src/dataconnector/github/mod.rs
```diff
@@ -703,6 +694,9 @@ pub(crate) fn filter_pushdown(expr: &Expr) -> FilterPushdownResult {
             let parameter = match column_name {
                 "title" => format!("{value} in:title"),
                 "body" => format!("{value} in:body"),
+                "state" => format!("is:{value}"), // is:merged, is:closed, is:open provides more granular results than state:closed
```
lukekim, line 697: Should these be uppercase? GitHub states are like `OPEN` `CLOSED` etc.

lukekim (approved):

crates/runtime/src/dataconnector/github/mod.rs
peasee: no, search uses lowercase values for modifiers like these. That's something I've experienced working on this, the state value is uppercase but the search value is lowercase.

Issue #3030, opened 2024-10-08
Connector RC Release: MySQL


peasee:
### All Connectors

These requirements are imposed on every connector, regardless of the selected features/tests in the above table.

- [x] [Core Arrow Data Types](../definitions.md) are supported
- [x] For exceptions where a [Core Arrow Data Type](../definitions.md) is not supported, print a warning and ignore the column instead of crashing or failing the query.
- [x] All known [Major Bugs](../definitions.md) are resolved

#### Documentation

- [x] Documentation includes all information and steps for a user to set up the connector.
- [x] Documentation includes all known issues/limitations for the connector.
- [x] Documentation includes any exceptions made to allow this connector to reach RC quality (e.g. if a particular data type cannot be supported by the connector).
- [x] The connector has an easy to follow quickstart.

#### Testing

- [x] Integration tests to cover simple use cases based on the connector type, e.g. a Spicepod sourcing a file from an FTP server, reading latest GitHub issues, etc.

### Conditional Criteria

The following features/tests are dependent on the required features/tests for the specified connector, from the [Connector criteria table](#rc-release-criteria)

#### Streaming

- [x] Data is streamed when reading/performing queries from this connector.

#### Schema Inference

Support for schema inference in a connector is classified as:

- ✅: Full coverage. The connector supports native schema inference.
- ⚠️: Partial coverage. The connector does not support native schema inference, but it requires no inference (e.g. static schema, etc).
- ❌: No coverage. The connector does not support native schema inference, and typically infers schema from the first row of results.

##### Full Coverage

- [x] The schema for data returned from queries is determined using a native CLI/library method provided by the connector.

#### Federation

Support for federation in a connector is classified as:

- ✅: Full coverage. The connector supports full federation and query push down.
- ⚠️: Partial coverage. The connector supports partial federation and query push down.
- ❌: No coverage. The connector does not support federation or query push down.

##### Full Coverage

- [x] The connector supports full federation within a single dataset (e.g. `select * from my_dataset`)
- [x] The connector supports federation push down across multiple datasets within the same connection source (e.g. `select * from first_dataset, second_dataset`)

#### Test Coverage

Indexes are not required for test coverage, but can be introduced if required for tests to pass (e.g. due to performance characteristics, etc).

##### TPC-H

- [x] End-to-end test to cover connecting to TPC-H SF1 for the connector type and benchmarking TPC-H queries (official and simple).
[...]

PR #3070, opened 2024-10-10
docs: Document stack overflow workaround, add helper script

peasee:
## 🗣 Description

* Documents a limitation with stack overflows, encountered when using the MySQL connector. This does not appear to be infinite recursion - confirmed by the fact that a minor stack size bump resolves the error. It appears to be caused by a large number of on-the-stack calls from the MySQL library when streaming results.
* Adds a simple test helper script to run a single query in the REPL.

## 🔨 Related Issues

* Part of #3030

## 🤔 Concerns

* I'm not sure if we should add a way to supply this parameter from `spice run` rather than running `spiced` directly? Alternatively, adding a default override to always make it 8MB because that appears to be the default on some other kernels.

### DIFF
--- a/crates/runtime/benches/queries/README.md
+++ b/crates/runtime/benches/queries/README.md
@@ -38,3 +38,27 @@ SELECT (now() + INTERVAL '30 days');
 | ------------------------ | ------------------------ |
 | [q14.sql](tpcds/q14.sql) | [q23.sql](tpcds/q23.sql) |
 | [q24.sql](tpcds/q24.sql) | [q39.sql](tpcds/q39.sql) |
+
+### Runtime worker has overflowed its stack
+
+**Limitation**: On some platforms (e.g. Linux kernel 6.9.3), the Runtime will encounter a stack overflow when running certain queries.
+
+**Solution**: Increase the stack size when running `spiced`, with `RUST_MIN_STACK=8388608 spiced` to set an 8MB minimum stack size.
+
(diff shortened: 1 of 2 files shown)


Sevenannn:
Is the helper script similar to the [run-queries.bash](https://github.com/spiceai/spiceai/blob/trunk/test/scripts/run-queries.bash) that's already in the test/scripts folder

peasee:
> Is the helper script similar to the [run-queries.bash](https://github.com/spiceai/spiceai/blob/trunk/test/scripts/run-queries.bash) that's already in the test/scripts folder

It's basically the same, but runs only a single query. Often, I don't want to run the full query set, but just want to re-run a single query without having to open REPL, copy the query, paste, etc.

phillipleblanc (approved):

sgrebnov:
> > Is the helper script similar to the [run-queries.bash](https://github.com/spiceai/spiceai/blob/trunk/test/scripts/run-queries.bash) that's already in the test/scripts folder
> 
> It's basically the same, but runs only a single query. Often, I don't want to run the full query set, but just want to re-run a single query without having to open REPL, copy the query, paste, etc.

Can we just support additional parameter instead of duplicating code? Example case - I wanted to improve this functionality to automatically do `.error` to display actual error when query has failed. I will need to update both files now. It seems most functionality is duplicated so we should add param support  instead of creating separate script IMO.

PR #3073, opened 2024-10-10
Automatically infer partitions for hive-style partitioned files for object store based connectors

phillipleblanc:
## 🗣 Description

Automatically infer the partitions for hive-style partitioned files, i.e.:

`/data/sales/year=2023/month=10/day=09/part-00000.parquet`

This applies to all object store based connectors, including S3 and file.

This PR also removes the runtime parameter `sql_query_keep_partition_by_columns` added in #3065 since it only modified the behavior when DataFusion wrote parquet files, not when it read them.

I've added a new dataset parameter `hive_infer_partitions` (defaults to true) that will control whether the runtime attempts to automatically infer the hive partitions.

Example:
```yaml
datasets:
  - from: s3://spiceai-public-datasets/hive_partitioned_data/
    name: hive_data_s3
    params:
      file_format: parquet
      hive_infer_partitions: true
```

This needs a documentation update for the new `hive_infer_partitions` parameter.


eadgbear (approved):
Would this cover other `object_store` file formats such as CSV and JSON as well? Just for documentation purposes

PR #3123, opened 2024-10-16
Adding telemetry to HTTP and Flight endpoints

### LINKED ISSUE #2714: task: Runtime supports `x-spice-user-agent` header values for telemetry
## What

The Runtime supports reading `x-spice-user-agent` header values from the HTTP and Flight API requests.

The Runtime should take these header values and transform them into telemetry dimensions for:
- `user-agent`: The originating SDK user agent (e.g. `gospice`, `spicepy`, etc). If malformed or an invalid value, defaults to the originating protocol (e.g. `HTTP`, `Flight`).
- `user-agent-version`: The SDK version supplied with the SDK agent. For example, a header value of `gospice 0.1.0 (Pop!_OS 22.04 LTS)` would have a `user-agent-version: 0.1.0`. Malformed or invalid values default to `unknown`.
- `user-agent-os`: The OS the SDK is running on. For example, a header value of `gospice 0.1.0 (Pop!_OS 22.04 LTS)` would have a `user-agent-os: Pop!_OS 22.04 LTS`

These dimensions would then be sent over flight to the OSS telemetry.

eadgbear:
## 🗣 Description

- User agent, version and OS are now propagated to the metrics as labels

## 🔨 Related Issues

- Closes #2714

## 🤔 Concerns

- In error situations (or when the User Agent isn't provided) the values default to `unknown` for all 3 fields.


Sevenannn (dismissed):

crates/runtime/src/user_agent_util.rs
```diff
@@ -0,0 +1,40 @@
+use regex::Regex;
+use std::sync::LazyLock;
+
+static UA_REGEX: LazyLock<Result<Regex, regex::Error>> =
+    LazyLock::new(|| Regex::new(r"^(\S+) (\S+) \((.*?)\)$"));
```
peasee:
```suggestion
    LazyLock::new(|| Regex::new(r"^(spice-rs|spice-java|spice.js|spicepy|gospice) v?(\d\.?)+(\-\w+)? \((.*?)\)$"));
```

could we use a stricter regex here? suggestion isn't fully fleshed, but e.g. specifying the specific values for each SDK user-agent value, checking version string format, etc.

peasee: I guess we'd need to solve for the capture groups differently if we make the regex stricter though

crates/runtime/src/flight.rs
```diff
@@ -241,6 +253,19 @@ impl Service {
 
         Ok((flights_stream.boxed(), query_result.from_cache))
     }
+
+    fn get_user_agent<T>(req: &Request<T>) -> [KeyValue; 3] {
```
phillipleblanc:
```suggestion
    fn get_request_labels<T>(req: &Request<T>) -> [KeyValue; 3] {
```

phillipleblanc: If we wanted to extend the labels we extract from a request, it should go in this function - so naming it appropriately up front seems wise.

crates/runtime/src/user_agent_util.rs
```diff
@@ -0,0 +1,40 @@
...
+
+pub fn extract_user_agent(ua_header: Option<&str>) -> (String, String, String) {
+    let ua_header = ua_header.unwrap_or("");
+    if let Ok(regex) = &*UA_REGEX {
+        if let Some(ua_parts) = regex.captures(ua_header) {
```
phillipleblanc: Does this need to be a regex? We control all of the possible inputs here, would a simple string split work? Regexes can add non-trivial perf cost, and this is in the critical path of requests.

crates/runtime/tests/metrics/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+        .with_reader(prometheus_exporter)
+        .with_reader(periodic_reader)
+        .build();
+    global::set_meter_provider(provider);
+
+    Ok(())
+}
+
+#[tokio::test]
+#[allow(clippy::clone_on_ref_ptr)]
```
phillipleblanc:
```suggestion
```

crates/runtime/tests/metrics/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    let rt = Runtime::builder()
+        .with_metrics_server_opt(Some(socket_addr), registry.clone())
+        .with_app(app)
+        .build()
+        .await;
+    tracing::info!("Starting runtime");
+    let rt_arc = Arc::new(rt);
+
+    let rt_arc_ref = rt_arc.clone();
```
phillipleblanc:
```suggestion
    let rt_arc_ref = Arc::clone(&rt_arc);
```

crates/runtime/src/user_agent_util.rs
peasee: someone could pass a wrong header if they use `curl` or something else. I think a regex makes sense, because we can match for format and expected values.

phillipleblanc: Sure, but someone using curl isn't what we care about. We want to know our SDK usage.

eadgbear:
Converting this to draft while I rework how we're capturing the user-agent. This will be done as part of the next release.

eadgbear:
Added a few changes:

* We're now less strict with the user agent. It'll now support anything split by spaces, so `gospice 0.1.0 (Pop!_OS 22.04 LTS)` will be split into `gospice`, `0.1.0` and `(Pop!_OS 22.04 LTS)` but `gospice 0.1.0` will be `unknown` and `unknown`. 
* Anonymous metrics now report two new fields:
  * `service.build` : either `dev` or `release`
  * `service.internal`: Controlled by `SPICEAI_INTERNAL_BUILD` environment variable. Will allow us to exclude internal benchmarks from certain dashboards

crates/runtime/src/lib.rs
```diff
@@ -102,6 +102,7 @@ pub mod tls;
 pub mod tools;
 pub(crate) mod tracers;
 mod tracing_util;
+mod user_agent_util;
```
phillipleblanc: We should have a `util` module with `tracing` and `user_agent` sub-modules rather than multiple top-level ones.

phillipleblanc (commented):
I think this will be good to go once the module restructure is done.

eadgbear:
Closing for a proper design process

PR #3124, opened 2024-10-16
docs: Update ROADMAP.md and release criterias

peasee:
## 🗣 Description

* Updates the roadmap to be in line with current expectations for the next RC releases in preparation for Stable.
* Updates the RC connector criteria to add the missing DuckDB connector, and add points about comparing TPC query success rate to Datafusion.
* Backports the Beta connector criteria as a subset of the criteria we previously used to release Beta connectors.
* Adds an RC accelerator criteria.
* Stubs out Alpha criteria for both connectors and accelerators.
* For release levels Beta and RC, explicitly calls out that the previous criteria must continue to pass (e.g. for RC, Beta criteria must still pass).


docs/ROADMAP.md
```diff
@@ -46,19 +87,46 @@ If you have a feature request or suggestion, please [get in touch](https://githu
...
+| `sqlite`     | Stable |
+| `postgresql` | Stable |
+
+### v1.0-stable Catalog Provider Support
+
+| Name         | Status |
+| ------------ | ------ |
+| `unity_catalog`| Stable |
+| `spice.ai`     | Stable |
+| `databricks`   | Beta   |
```
Sevenannn, line 119: Do we have a document for catalog provider stable / beta quality

docs/criteria/connectors/rc.md
```diff
@@ -37,33 +37,37 @@ These connectors are exempt from running the TPC derived test packages, and rely
...
+| Spark                   | ✅ (1) | ✅ (1) | ✅ | ✅ | ✅ | ✅ |
 
 ### All Connectors
 
 These requirements are imposed on every connector, regardless of the selected features/tests in the above table.
 
-- [ ] [Core Arrow Data Types](../definitions.md) are supported
+The RC release criteria expand on and require that all [Beta release criteria](./beta.md) continue to pass for the connector.
+
+- [ ] All [Beta release criteria](./beta.md) pass for this connector.
```
sgrebnov, line 69: > All [Beta release criteria](./beta.md) pass for this connector.

I found this hard to use as diff between Beta and RC is unclear - what specific point of Beta must additionally be satisfied? It looks like RC includes everything from Beta already or I miss something?

docs/criteria/connectors/rc.md
```diff
@@ -136,6 +140,7 @@ Indexes are not required for test coverage, but can be introduced if required fo
 - [ ] A test script exists that can load TPC-H data at the [designated scale factor](#rc-release-criteria) into this connector.
 - [ ] The connector can load TPC-H at the [designated scale factor](#rc-release-criteria), and can run all queries with no [Major Bugs](../definitions.md).
 - [ ] All [Minor Bugs](../definitions.md) for TPC-H are raised as issues.
+- [ ] TPC-H queries pass with a success rate equal or greater than TPC-H execution on Datafusion.
```
sgrebnov: It is good to elaborate more on  `success rate`  -  does it mean percentage of passed queries? for ex Postgres fails on Q1 and Q2, but works for Q3 and Q4 and Arrow fails for Q3 and Q4 and passes Q1 and Q2 - does this case satisfies `success rate` criteria?

docs/criteria/connectors/rc.md
```diff
@@ -37,33 +37,37 @@ These connectors are exempt from running the TPC derived test packages, and rely
...
 
 These requirements are imposed on every connector, regardless of the selected features/tests in the above table.
 
-- [ ] [Core Arrow Data Types](../definitions.md) are supported
+The RC release criteria expand on and require that all [Beta release criteria](./beta.md) continue to pass for the connector.
+
+- [ ] All [Beta release criteria](./beta.md) pass for this connector.
+- [ ] [Core Arrow Data Types](../definitions.md) are supported.
 - [ ] For exceptions where a [Core Arrow Data Type](../definitions.md) is not supported, print a warning and ignore the column instead of crashing or failing the query.
 - [ ] All known [Major Bugs](../definitions.md) are resolved
```
sgrebnov:
```suggestion
- [ ] Documentation includes Native Type to Arrow Type mapping (if applicable)
- [ ] All known [Major Bugs](../definitions.md) are resolved
```

docs/criteria/connectors/rc.md
```diff
@@ -37,33 +37,37 @@ These connectors are exempt from running the TPC derived test packages, and rely
...
 
 ### All Connectors
 
 These requirements are imposed on every connector, regardless of the selected features/tests in the above table.
 
-- [ ] [Core Arrow Data Types](../definitions.md) are supported
+The RC release criteria expand on and require that all [Beta release criteria](./beta.md) continue to pass for the connector.
+
+- [ ] All [Beta release criteria](./beta.md) pass for this connector.
+- [ ] [Core Arrow Data Types](../definitions.md) are supported.
```
sgrebnov: Should be around native types support and not Arrow type IMO. Arrow Type is just an internal representation.

docs/ROADMAP.md
peasee: There is currently no criteria for either models or catalog providers. It will need to be created.

docs/criteria/connectors/rc.md
peasee: RC should include everything from Beta and extend on it. Beta:

* includes only TPC-H, while RC is TPC-H and TPC-DS
* does not include native schema inference, data correctness tests, and only partial coverage for federation
* excludes the requirement to print warnings instead of failing on unsuported column types

Some segments are repeated in RC. This is intended, because they should explicitly be re-tested. For example, Beta tests TPC-H at scale 1, while RC tests at 100. Documentation should continue to be updated in RC, because there may be new exceptions from RC that need to be documented. There are a few points that can be excluded from RC because they exist in Beta that I've updated to move across.

docs/criteria/connectors/rc.md
peasee: I'm not sure I agree. We support native types through Arrow types. If we can't support a native type, we can't support the Arrow type either. For accelerators, the focus is on Arrow conversion back to native (which for Arrow, is still Arrow) - accelerators don't know about the connector's native type the data originated from.

I think adding a documentation table for any mappings from native types makes sense, but I don't think we should change the supported data types.

docs/criteria/connectors/rc.md
peasee: I've changed this to call out comparing against the individual query executions on Datafusion

lukekim (dismissed):

docs/criteria/connectors/rc.md
sgrebnov: This is Connector specific (not Accelerator):

I see, thank you for the details. I still think that if this is about Native types rather than Arrow types, it will be more useful and provide better guidance for testing and understanding what it means for people using this data connector. For example:

MS SQL Connector supports (and we tested) the following types, which makes more sense:
```
UNIQUEIDENTIFIER
NVARCHAR
NTEXT
..
```
than saying MS SQL Connector supports the following Arrow types:
```
Utf8
```

phillipleblanc (approved):

PR #3126, opened 2024-10-17
deps: Update datafusion version

### LINKED ISSUE #1952: Bug: MySQL error `Every derived table must have its own alias`
**Describe the bug**

When running TPCH query 13 with MySQL, the following rewritten query is generated and is failed to execute with the following error 'Every derived table must have its own alias'

```sql
SELECT
	`c_orders`.`c_count`,
	COUNT(1) AS `custdist`
FROM
	(
	SELECT
		`customer`.`c_custkey` AS `c_custkey`,
		`COUNT(orders.o_orderkey)` AS `c_count`
	FROM
		(
		SELECT
			`customer`.`c_custkey`,
			COUNT(`orders`.`o_orderkey`)
		FROM
			`customer`
		LEFT JOIN `orders` ON
			((`customer`.`c_custkey` = `orders`.`o_custkey`)
				AND `orders`.`o_comment` NOT LIKE CAST('%special%requests%' AS CHAR))
		GROUP BY
			`customer`.`c_custkey`)) AS `c_orders`
GROUP BY
	`c_orders`.`c_count`
ORDER BY
	`custdist` DESC,
	`c_orders`.`c_count` DESC
```

Note: `CAST('%special%requests%' AS TEXT))` was manually replaced with `CAST('%special%requests%' AS CHAR))` due to another but;

peasee:
## 🗣 Description

* Updates the datafusion version to bring in the latest updates.

## 🔨 Related Issues

* Closes #1952

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2904,7 +2904,7 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "42.0.0"
-source = "git+https://github.com/spiceai/datafusion.git?rev=4e34dad74a63fd3ead3f33bae55c6916887b1439#4e34dad74a63fd3ead3f33bae55c6916887b1439"
+source = "git+https://github.com/spiceai/datafusion.git?rev=774f1ebff31f117658f29e0329f488dd2f337bce#774f1ebff31f117658f29e0329f488dd2f337bce"
 dependencies = [
  "ahash 0.8.11",
  "arrow",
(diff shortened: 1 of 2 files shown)


sgrebnov (approved):

sgrebnov:
The datafusion version has been updated to a later version alrady so this update is no longer valid/required.

sgrebnov (changes requested):
The datafusion version has been updated to a later version already so this update is no longer valid/required.

PR #3166, opened 2024-10-21
Fix task_history output for sql, add output to table_schema & list_datasets tool

phillipleblanc:
## 🗣 Description

Fixes the task_history output for SQL queries, add the output for the tools `table_schema` and `list_datasets` and use a fork of `arrow-rs` that implements `Decimal128` and `Decimal256` types to ensure we provide output for SQL queries which return decimal values.

Also fixes an issue where if the zipkin endpoint wasn't available, it would prevent the task_history table from being populated due to the opentelemetry exporter timing out.

## 🔨 Related Issues

arrow-rs PR: https://github.com/apache/arrow-rs/pull/6606

N/A

### DIFF
--- a/bin/spiced/src/tracing.rs
+++ b/bin/spiced/src/tracing.rs
@@ -26,7 +26,9 @@ use opentelemetry_sdk::{
     trace::{Config, TracerProvider},
     Resource,
 };
+use reqwest::Client;
 use runtime::{datafusion::DataFusion, task_history};
+use std::time::Duration;
 use tracing::Subscriber;
 use tracing_subscriber::{filter, fmt, layer::Layer, prelude::*, registry::LookupSpan, EnvFilter};
 
@@ -65,7 +67,7 @@ impl From<LogVerbosity> for EnvFilter {
     }
 }
 
-pub(crate) fn init_tracing(
+pub(crate) async fn init_tracing(
     app: &Option<Arc<App>>,
     config: Option<&TracingConfig>,
     df: Arc<DataFusion>,
@@ -91,7 +93,7 @@ pub(crate) fn init_tracing(
 
     let subscriber = tracing_subscriber::registry()
         .with(filter)
-        .with(datafusion_task_history_tracing(df, app, config)?)
+        .with(datafusion_task_history_tracing(df, app, config).await?)
         .with(
             fmt::layer()
                 .with_ansi(true)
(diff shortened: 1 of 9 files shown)


crates/runtime/src/datafusion/query/tracker.rs
```diff
@@ -156,5 +156,5 @@ fn trace_query(query_tracker: &QueryTracker, truncated_output: &str) {
         .collect::<Vec<String>>()
         .join(",");
     tracing::info!(target: "task_history", protocol = ?query_tracker.protocol, datasets = datasets_str, "labels");
-    tracing::info!(target: "task_history", truncated_output = %truncated_output);
+    tracing::info!(target: "task_history", captured_output = %captured_output);
```
phillipleblanc, line 159: This is the change that fixes the output for sql

crates/runtime/src/tools/builtin/table_schema.rs
```diff
@@ -97,10 +97,19 @@ impl SpiceModelTool for TableSchemaTool {
...
                 .boxed()?;
 
-            let v = serde_json::value::to_value(schema).boxed()?;
+            let schema_value = serde_json::value::to_value(schema).boxed()?;
 
-            table_schemas.push(v);
+            let table_schema = serde_json::json!({
+                "table": t,
+                "schema": schema_value
+            });
```
phillipleblanc, line 105: We weren't including the table name with the schema information returned, so it was probably confusing the model when multiple table schemas were returned previously.

bin/spiced/src/tracing.rs
```diff
@@ -173,11 +182,23 @@ fn zipkin_task_history_otel_exporter(
...
+            .with_http_client(Client::new())
             .init_exporter()?,
     )))
 }
 
+async fn is_zipkin_endpoint_reachable(endpoint: &str) -> bool {
+    let client = Client::new();
+    let timeout = Duration::from_secs(5);
+
+    let url = format!("{endpoint}?serviceName=test");
```
phillipleblanc, line 194: This will return an empty array and is a good test to ensure we are connected to a valid zipkin endpoint: https://zipkin.io/zipkin-api/#/default/get_spans

ewgenius (approved):

Jeadie (approved):

Issue #3100, opened 2024-10-14
Enhancement: Separate Tokio runtimes for query/AI computation/acceleration


phillipleblanc:
## Goal-State/What/Result

Have separate Tokio runtimes for query/AI computation (inference, embeddings, etc) and acceleration refreshes. The goal is that all queries should stay fast no matter what else the runtime is doing.

## Why/Purpose

This is required for delivering a good user experience when the runtime is also busy with CPU-bound tasks, like calculating embeddings with local models.

## By When

_A target date for done-done completion of the entire enhancement._

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [ ] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Samples/Quickstarts/Tutorials
- [ ] Release Notes

## The Algorithm

- [ ] Every requirement questioned?
- [ ] Delete (Scope) any part you can.
- [ ] Simplify.
- [ ] Break down into smaller iterations/milestones.
- [ ] Opportunities for automation.

## Spec

_Basic specification of the enhancement._

## How/Plan

_A clear and concise plan of how this should be implemented._

### Implementation (Sub Issues)

- Sub Issue 1
- Sub Issue 2

### QA

- Sub Issue 1
- Sub Issue 2

Sevenannn:
EOW Target: Done-Done

phillipleblanc:
I believe this might be relevant:

https://docs.rs/tokio/1.14.0/tokio/#cpu-bound-tasks-and-blocking-code

> It is also possible to create an extra Tokio runtime dedicated to CPU-bound tasks, but if you do this, you should be careful that the extra runtime runs only CPU-bound tasks, as IO-bound tasks on that runtime will behave poorly.

phillipleblanc:
Also relevant: https://www.influxdata.com/blog/using-rustlangs-async-tokio-runtime-for-cpu-bound-tasks/

PR #3170, opened 2024-10-21
Calculate embeddings in a separate blocking thread

phillipleblanc:
## 🗣 Description

Calculating embeddings using candle locally on the Tokio runtime threads causes other async tasks to be starved, causing responsiveness issues in the runtime. Run the embeddings using `tokio::task::spawn_blocking` in a separate thread and yield control of the current task back to the Tokio runtime so it can work on other tasks.

## 🔨 Related Issues

Part of #3100

N/A

## 🤔 Concerns

This doesn't separate out the task onto a separate runtime - but blocking work shouldn't be done in any tokio runtime. This is a good first step towards that goal.

### DIFF
--- a/crates/llms/src/embeddings/candle/mod.rs
+++ b/crates/llms/src/embeddings/candle/mod.rs
@@ -146,16 +146,19 @@ impl CandleEmbedding {
         encoding.get_ids().iter().take_while(|&&x| x != 0).count()
     }
 
-    fn encode(&self, inp: String) -> Result<Encoding> {
-        self.tok
-            .encode::<String>(inp, true)
+    fn encode(tok: &Arc<Tokenizer>, inp: String) -> Result<Encoding> {
+        tok.encode::<String>(inp, true)
             .context(FailedToPrepareInputSnafu)
     }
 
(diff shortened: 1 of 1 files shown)


ewgenius (approved):

crates/llms/src/embeddings/candle/mod.rs
```diff
@@ -237,7 +239,15 @@ impl Embed for CandleEmbedding {
     }
 
     async fn embed(&self, input: EmbeddingInput) -> Result<Vec<Vec<f32>>> {
-        self._embed(input).map(|(embeddings, _usage)| embeddings)
+        // This needs to be blocking to avoid starving other async tasks in the runtime, since this blocks for long periods of time.
```
Jeadie, line 242: Is there a reason not to perform the blocking in `Self::_embed`?

Jeadie (approved):

crates/llms/src/embeddings/candle/mod.rs
phillipleblanc: I chose to keep `Self::_embed` as a non-async function and treat the whole function as blocking

PR #3174, opened 2024-10-21
normalise embeddings by default; allow pooling params.

Jeadie:


Jeadie:
Not needed, #3199 instead.

PR #3188, opened 2024-10-21
Improve CLI error output when REPL can't connect to the Flight endpoint

eadgbear:
## 🗣 Description

We were outputting the wrong address when in an error condition when trying to connect to the Flight endpoint. This small fix improves that.


crates/flightrepl/src/lib.rs
```diff
@@ -153,19 +153,20 @@ pub async fn run(repl_config: ReplConfig) -> Result<(), Box<dyn std::error::Erro
...
+            .await
     };
 
     // Set up the Flight client
-    let spice_endpoint = repl_config.http_endpoint.clone();
-    let channel = channel.map_err(|_err| {
+    let channel = channel.map_err(|err| {
         Box::<dyn Error>::from(format!(
-            "Unable to connect to spiced at {spice_endpoint}. Is it running?"
+            "Unable to connect to spiced at {repl_flight_endpoint}. Is it running? ({err})"
```
phillipleblanc: I would keep the error itself under a debug log, not by default

eadgbear: Flightrepl currently has no tracing or logging of any kind, I'll have to add all of that in

eadgbear:
Flightrepl crate currently doesn't have any form of tracing, so that would have to be added before this can be marked as completed.

lukekim (approved):

PR #3206, opened 2024-10-23
Split spiced_docker build across architectures

phillipleblanc:
## 🗣 Description

Splits the `spiced_docker` build to run on runners native to the architecture for each layer, and then combine the layers at the end before publishing.

Example run to publish `v0.19.2-beta`: [spiced_docker · spiceai/spiceai@3257f3b](https://github.com/spiceai/spiceai/actions/runs/11483049529/job/31959047271)

### DIFF
--- a/.github/workflows/build_and_release.yml
+++ b/.github/workflows/build_and_release.yml
@@ -22,7 +22,7 @@ jobs:
       matrix:
         include:
           - name: Linux x64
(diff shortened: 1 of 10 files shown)


peasee (approved):
Wow, runs much faster now that we're not cross-compiling `arm` builds!

PR #3229, opened 2024-10-25
Enable overlap_size correctly

Jeadie:
## 🗣 Description
 - We weren't applying the overlap_size for chunking datasets :( 
 - To fix this, we also need to update the `Embed` trait (it is now possible that the users parameters are inconsistent (overlap > desired)). 
 ```diff
- fn chunker(&self, cfg: &ChunkingConfig) -> Option<Arc<dyn Chunker>> {
+ fn chunker(&self, cfg: &ChunkingConfig) -> Result<Arc<dyn Chunker>> {
```

### CODE REVIEW

File: crates/llms/src/chunking/mod.rs [hunk 1]
```diff
@@ -162,7 +167,7 @@ mod tests {
                 "target_chunk_size",
                 ": 3",
                 "overlap_size:",
-                "1"
```
Jeadie, line 165: This is evidence our overlap is now being applied :)

lukekim (dismissed):

lukekim (dismissed):

Jeadie:
This needs a fix to the `-offsets` to take into account that the end index won't simply be the start index of the next chunk.

phillipleblanc (dismissed):

lukekim (approved):

PR #3242, opened 2024-10-25
Better S3 error message

Sevenannn:
## 🗣 Description

Display more actionable S3 error messages when query fail due to client timeout.

Before:
image

After:
image
image
Full error is still retained in 

## 🤔 Concerns

The S3 error matching is added in the `sql_to_flight_stream`. This is because the error `Arrow error: External error: External: Generic S3 error: error decoding response body; code: QueryExecutionError` is emitted at https://github.com/spiceai/datafusion/blob/bfd815622f1fe2c84d6fab32596b83ffbe52a84a/datafusion/core/src/datasource/physical_plan/parquet/opener.rs#L206 ,and it would be inappropriate to add customized error handling logic in datafusion / object store, like what we did for `databaseconn` in Spice AI repo and Datafusion Table Provider repo.

### DIFF
--- a/crates/runtime/src/flight.rs
+++ b/crates/runtime/src/flight.rs
@@ -163,7 +163,10 @@ impl Service {
     ) -> Result<Schema, Status> {
         let query = QueryBuilder::new(sql, datafusion, protocol).build();
 
-        let schema = query.get_schema().await.map_err(handle_datafusion_error)?;
+        let schema = query
+            .get_schema()
+            .await
+            .map_err(handle_get_schema_datafusion_error)?;
         Ok(schema)
     }
 
@@ -211,19 +214,7 @@ impl Service {
                             flights.push(flight_batch.into());
                             Ok(flights)
                         }
-                        Err(e) => match e {
-                            DataFusionError::External(e) => {
-                                if let Some(execution_error) =
-                                    e.to_string().strip_prefix("Execution error: ")
-                                {
-                                    return Err(Status::invalid_argument(
-                                        execution_error.to_string(),
-                                    ));
-                                }
-                                Err(Status::invalid_argument(e.to_string()))
-                            }
-                            _ => Err(Status::invalid_argument(e.to_string())),
-                        },
+                        Err(e) => Err(handle_sql_datafusion_error(e)),
                     }
                 }
             })
@@ -263,7 +254,7 @@ where
 }
 
 #[allow(clippy::needless_pass_by_value)]
-fn handle_datafusion_error(e: DataFusionError) -> Status {
+fn handle_get_schema_datafusion_error(e: DataFusionError) -> Status {
     match e {
         DataFusionError::Plan(err_msg) | DataFusionError::Execution(err_msg) => {
             Status::invalid_argument(err_msg)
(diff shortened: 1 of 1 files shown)


crates/runtime/src/flight.rs
```diff
@@ -283,6 +274,34 @@ fn handle_datafusion_error(e: DataFusionError) -> Status {
...
+                external_error.to_string().strip_prefix("Execution error: ")
+            {
+                return Status::invalid_argument(execution_error.to_string());
+            }
+            Status::invalid_argument(e.to_string())
+        }
+        DataFusionError::ArrowError(ref arrow_error, _) => {
+            // The S3 query error is managed here to avoid
+            // introducing custom downcast logic within Datafusion or Object Store.
+            let error_message = format!("{arrow_error:?}");
```
phillipleblanc, line 291: Can you put an example of what this is supposed to be matching on?

crates/runtime/src/flight.rs
```diff
@@ -283,6 +274,34 @@ fn handle_datafusion_error(e: DataFusionError) -> Status {
...
+            let s3_error_indicator = "store: \"S3\",";
```
phillipleblanc, line 292: Putting S3 specific error handling logic in the Flight server code doesn't make sense.

crates/runtime/src/flight.rs
```diff
@@ -283,6 +274,34 @@ fn handle_datafusion_error(e: DataFusionError) -> Status {
...
+
+            if let Some(error_details) = error_message.split(s3_error_indicator).nth(1) {
+                if error_details.contains("TimedOut") && error_details.contains("Decode") {
+                    return Status::invalid_argument("Timed out while fetching data from S3. Please increase the `client_timeout` parameter.");
```
phillipleblanc, line 296:
```suggestion
                    return Status::invalid_argument("Timed out while fetching data from S3.");
```

This error message is being returned to the flight client, telling the user making the flight request to increase the client_timeout isn't going to help.

crates/runtime/src/flight.rs
```diff
@@ -283,6 +274,34 @@ fn handle_datafusion_error(e: DataFusionError) -> Status {
...
+            Status::invalid_argument(e.to_string())
+        }
+        DataFusionError::ArrowError(ref arrow_error, _) => {
+            // The S3 query error is managed here to avoid
+            // introducing custom downcast logic within Datafusion or Object Store.
+            let error_message = format!("{arrow_error:?}");
+            let s3_error_indicator = "store: \"S3\",";
+
+            if let Some(error_details) = error_message.split(s3_error_indicator).nth(1) {
+                if error_details.contains("TimedOut") && error_details.contains("Decode") {
```
phillipleblanc, line 295: Is there no concrete error type we can match against?

crates/runtime/src/flight.rs
Sevenannn: The error pattern matched is: ArrowError(ExternalError(ObjectStore(Generic { store: "S3", source: reqwest::Error { kind: Decode, source: reqwest::Error { kind: Body, source: TimedOut } } })), None)

crates/runtime/src/flight.rs
Sevenannn: Yeah I've thought about that, the tricky thing here is that neither Listing Table nor Object store is something defined in our codebase, so it's hard to implement a S3 specific matching like how we handle the postgres error in postgresconn or github error in GithubRestClient. Do you have any suggestions in terms of avoiding to put this in flight server (since the flightserver has no knowledge of the S3 connector). Thanks!

crates/runtime/src/flight.rs
Sevenannn: The matching will be DatafusionError::ArrowError -> DatafusionError::ExternalError -> ObjectStoreError::Generic -> reqwest::Error -> reqwest::Error. I wonder if the exact matching of this chain of error types would be too complicated

phillipleblanc: I think in general matching against a complicated error chain is preferable to matching against strings.

Sevenannn:
Closed in favor of https://github.com/spiceai/spiceai/pull/3304

PR #3245, opened 2024-10-26
Enable parquet pushdown in Spice

Sevenannn:
## 🗣 Description

Enable parquet pushdown in Spice. This will fix the tpcds query 4, 11, 74, incorrectness issue where filter don't get applied in the non-pushdown mode.


crates/runtime/src/dataconnector/listing/connector.rs
```diff
@@ -135,10 +137,23 @@ pub trait ListingTableConnector: DataConnector {
...
                 extension.unwrap_or(".csv".to_string()),
             )),
-            Some("parquet") => Ok((
-                Some(Arc::new(ParquetFormat::default())),
-                extension.unwrap_or(".parquet".to_string()),
-            )),
+            Some("parquet") => {
+                let mut table_parquet_options = TableParquetOptions::new();
+                table_parquet_options
+                    .set("pushdown_filters", "true")
```
phillipleblanc: A simpler option might be to just set the `datafusion.execution.parquet.pushdown_filters ` option on the session context: https://datafusion.apache.org/user-guide/configs.html

Sevenannn: It won't work. I've tried that in `DataFusionBuilder`. It's because the datafusion ParquetExec take the filter pushdown configuration from the `ParquetFormat` options instead of session context, so it needs to be set in the `TableParquetOptions ` for `ParquetFormat`: https://github.com/apache/datafusion/blob/444a673682a4823b98695568ecdab1087b5b6a60/datafusion/core/src/datasource/file_format/parquet.rs#L409

crates/runtime/src/dataconnector/listing/connector.rs
```diff
@@ -149,8 +164,19 @@ pub trait ListingTableConnector: DataConnector {
                         ));
                     }
                     if ext.eq_ignore_ascii_case("parquet") {
+                        let mut table_parquet_options = TableParquetOptions::new();
```
Jeadie: nit: duplicate code to above, could make into function.

Jeadie (dismissed):

Jeadie (approved):

PR #3251, opened 2024-10-27
Replace `query_duration_seconds` and `http_requests_duration_seconds` with `milliseconds` metrics

### LINKED ISSUE #3231: OTEL Metrics: Update seconds based telemetry to be ms

sgrebnov:
## 🗣 Description

Following [OTEL recommendations for http metrics](https://opentelemetry.io/docs/specs/semconv/http/http-metrics/), this PR replaces `http_requests_duration_seconds` with milliseconds based `http_requests_duration_ms`

`query_duration_seconds` metric has been removed in favor of existing `request_duration` (milliseconds), which was renamed to `query_duration_ms`

**Breaking changes:**
- `http_requests_duration_seconds` metric is replaced with `http_requests_duration` (milliseconds)
- existing `query_duration` (milliseconds) metric has been renamed to `query_duration_ms`
- `query_duration_seconds` metric has been removed in favor metric above - `query_duration_ms` (milliseconds)

## 🔨 Related Issues

Closes https://github.com/spiceai/spiceai/issues/3231


phillipleblanc:
To prevent a breaking change on the upcoming release, we need to duplicate the metrics instead of renaming them. We'll revisit all metric names in a future release and make a breaking change for all of them at once.

crates/runtime/src/http/routes.rs
```diff
@@ -102,7 +102,7 @@ async fn track_metrics(req: Request<Body>, next: Next) -> impl IntoResponse {
 
     let response = next.run(req).await;
 
-    let latency = start.elapsed().as_secs_f64();
+    let latency_ms = start.elapsed().as_secs_f64() * 1000.0;
```
lukekim, line 105: Is there no as_millis here?

lukekim (approved):

crates/runtime/src/http/routes.rs
sgrebnov: Confirm, I've created separate work item to address this as a common issue/pattern that will be done this week / separate PR.

PR #3252, opened 2024-10-27
For NSQL, remove arbitrary CREATE table statements, use `table_schema` tool instead.

Jeadie:
## 🗣 Description
 - Currently, we add  Postgres table creation statements in the system prompt of a text to SQL command. We already have a tool for retrieving table schemas, we should use that instead. This is a known part of the system for users, and will be visible in our telemetry/tracing. 

 - Fix tracing for `task='nsql'` so that `tool_use::table_schema`, `tool_use::sample_data` and `sql_query` have complete parent relationships. 

## 🔨 Related Issues
 - Quickstart with example: https://github.com/spiceai/quickstarts/pull/221


crates/runtime/src/tools/builtin/table_schema.rs
```diff
@@ -44,6 +62,71 @@ impl TableSchemaTool {
             description,
         }
     }
+
+    pub async fn get_schema(
```
Jeadie, line 66: This refactor allows us to use `TableSchemaTool` in the code, when we only have a `Datafusion` object, and still get the telemetry.

crates/runtime/src/tools/builtin/table_schema.rs
```diff
@@ -44,6 +62,71 @@ impl TableSchemaTool {
...
+        for t in &req.tables {
+            let schema = df
+                .get_arrow_schema(t)
+                .instrument(span.clone())
+                .await
+                .boxed()?;
+
+            let schema_value = serde_json::value::to_value(schema).boxed()?;
+
+            let table_schema = serde_json::json!({
```
peasee, line 83: why not make an intermediary struct, like `TableSchemaResponse`? This would avoid needing to JSON serialise every element, when we JSON serialise again on L91 anyway. I'm mainly concerned about performance of many JSON serialises here.

phillipleblanc: It's not quite serializing it to a JSON string, but creating a `serde_json::Value` struct instead. The stringification only happens on line 91

phillipleblanc (approved):

peasee (approved):

Issue #2916, opened 2024-10-01
Enhancement: Handle GitHub rate limiting in GitHub and GraphQL data connectors


ewgenius:
## Goal-State/What/Result

Runtime should detect when GitHub requests are rate-limited for accelerated datasets and pause data refresh based on the reset timeout

https://docs.github.com/en/graphql/overview/rate-limits-and-node-limits-for-the-graphql-api

## Why/Purpose

If the runtime continuously hits the GitHub API rate limits, the user token or host IP address might be blocked.

Currently, when using GitHub or the GraphQL/GitHub connector, the runtime does not handle rate limiting:

```console
2024-10-01T15:09:35.171427Z  WARN runtime::accelerated_table::refresh_task: Failed to load data for dataset github.pulls_spiceai: Unable to get data from connector: External error: External error: The API failed with status code 403 Forbidden; Please check if provided credentials have the necessary permissions.
```

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [ ] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] UX Review
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Documentation
- [ ] Samples/Quickstarts/Tutorials
- [ ] Release Notes

## The Algorithm

- [ ] Every requirement questioned?
- [ ] Delete (Scope) any part you can.
- [ ] Simplify.
- [ ] Break down into smaller iterations/milestones.
- [ ] Opportunities for automation.

## Spec

Basic specification of the enhancement.

## How/MAP

A clear and concise plan of how this should be implemented.

### Implementation (issues)

- 
-

### QA

eadgbear:
This issue should be focused on global rate limiting across all connectors and across Spice instances in Kubernetes/Spice Operator

phillipleblanc:
This was closed by https://github.com/spiceai/spiceai/pull/3610

PR #3258, opened 2024-10-28
fix: Expose GitHub ratelimit errors

### LINKED ISSUE #3257: GitHub GraphQL does not report rate limit errors

peasee:
## 🗣 Description

* Changes `GraphQLOptimizer` to `GraphQLContext`, because it's a useful carrier of other GraphQL related configuration.
* Adds `error_checker` to the `GraphQLContext`, allowing consumers to middle-man the headers and response to return custom errors
* Implements `error_checker` for GitHub, checking for errors relating to rate limits.

This ensures rate limit errors are actually made visible, instead of generic `403 Forbidden` errors.

## 🔨 Related Issues

* Closes #3257

### DIFF
--- a/crates/data_components/src/graphql/provider.rs
+++ b/crates/data_components/src/graphql/provider.rs
@@ -27,7 +27,7 @@ use datafusion::{
 };
 use std::{any::Any, sync::Arc};
 
-use super::{client::GraphQLClient, GraphQLOptimizer, ResultTransformSnafu};
+use super::{client::GraphQLClient, GraphQLContext, ResultTransformSnafu};
 use super::{client::GraphQLQuery, Result};
 
 pub type TransformFn =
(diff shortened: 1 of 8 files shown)


crates/data_components/src/graphql/provider.rs
```diff
@@ -56,7 +56,7 @@ impl GraphQLTableProviderBuilder {
     }
 
     #[must_use]
-    pub fn with_optimizer(mut self, optimizer: Arc<dyn GraphQLOptimizer>) -> Self {
+    pub fn with_optimizer(mut self, optimizer: Arc<dyn GraphQLContext>) -> Self {
```
phillipleblanc: Change to context everywhere vs optimizer

phillipleblanc (approved):

sgrebnov (approved):

Jeadie:
Is it possible to use the rate limiting errors from the client to rate limit the provider so that a `.scan(` can be successful? or is it not worth it if all GH datasets aren't using the same underlying client connection?

peasee:
> Is it possible to use the rate limiting errors from the client to rate limit the provider so that a `.scan(` can be successful? or is it not worth it if all GH datasets aren't using the same underlying client connection?

This PR is not interested in retrying rate limits. Rate limit retries will be addressed in https://github.com/spiceai/spiceai/issues/2916

Issue #3185, opened 2024-10-21
Enhancement: Unified refresh


lukekim:
## Goal-State/What/Result

Refreshes for two or more data connectors with the same acceleration parameters should be unified (only one refresh happens).

## Why/Purpose

To enable the in-memory primary, file backup acceleration fallback pattern.

## By When

_A target date for done-done completion of the entire enhancement._

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

E.g.

```yaml
datasets:
  - from: s3://blah/blah
     name: primary
     refresh_interval: 5m
  - from: s3://blah/blah
     name: secondary
     refresh_interval: 5m
```

Should only cause one refresh every 5 mins, but both datasets to have the refreshed data.

## How/Plan

_A clear and concise plan of how this should be implemented._

### Implementation (Sub Issues)

- Sub Issue 1
- Sub Issue 2

### QA

- Sub Issue 1
- Sub Issue 2

phillipleblanc:
Target: Done by Tue

PR #3280, opened 2024-10-29
Accelerated datasets can fallback to federated source while loading

### LINKED ISSUE #3250: Accelerated datasets can fallback to federated source if not ready
Add a config option on acceleration parameters: `fallback_on_load` (or something) that allows the dataset to fallback to the federated source while the initial acceleration load is happening.

Also mark the dataset as "ready" for the readiness endpoint immediately if this is set.

phillipleblanc:
## 🗣 Description

This PR introduces a new option to the control the behavior of accelerated tables during their initial data load. The previous behavior was to have the `/v1/ready` API return `Not Ready` until all datasets were fully loaded. This is still the default behavior with `ready_state: on_load`. However, now when a query is executed against an accelerated table that is still loading data configured with `ready_state: on_load`, an error is returned indicating the acceleration is still initializing. Previously this would have returned zero results.

A new option `ready_state: on_registration` has been added to allow the `/v1/ready` API to return `Ready` immediately after the accelerated table is registered, and any queries on the accelerated table during the initial load to fallback to the federated source directly.

Example Spicepod:

```yaml
datasets:
  - from: s3://my_bucket/my_dataset/
    name: my_dataset
    params:
      ...
    acceleration:
      enabled: true
      ready_state: on_registration # Will mark the dataset as ready immediately and queries on this table will fallback to the S3 source directly until the initial acceleration is complete.
  - from: s3://my_bucket/my_dataset2/
    name: my_dataset2
    params:
      ...
    acceleration:
      enabled: true
      ready_state: on_load # Will mark the dataset as ready only after the initial acceleration is complete. Queries against the dataset will return an error before the load has completed.
```

The most important changes include adding the `ReadyState` enum, updating the `AcceleratedTable` and `Refresher` structs, and modifying related methods to incorporate this new behavior.

## 🔨 Related Issues

Closes #3250
Part of #3185

This will require docs changes to document the new acceleration behavior.

## 🤔 Concerns

It might make sense to have the accelerated table always query the federated source while the initial load is in progress, and have this new option only affect the readiness API.


sgrebnov (dismissed):

Jeadie:
> It might make sense to have the accelerated table always query the federated source while the initial load is in progress, and have this new option only affect the readiness API.

I think this is reasonable.

bin/spiced/src/lib.rs
```diff
@@ -226,7 +226,9 @@ pub async fn run(args: Args) -> Result<()> {
 
     let cloned_rt = rt.clone();
     let server_thread =
-        tokio::spawn(async move { cloned_rt.start_servers(args.runtime, tls_config).await });
+        tokio::spawn(
```
Jeadie, line 229: Why did we have to pin this now?

Jeadie (dismissed):

bin/spiced/src/lib.rs
phillipleblanc: I was getting this lint error: https://rust-lang.github.io/rust-clippy/master/index.html#large_futures

Jeadie: I just got it too 😆

sgrebnov (approved):

PR #3316, opened 2024-10-31
initial work on xAI, blocked on it not being OpenAI compatible.

Jeadie:


Jeadie:
Not priority and not much work done on this.

PR #3324, opened 2024-11-01
fix: Allow only ODBC profiles

peasee:
## 🗣 Description

* Prevents ODBC from loading drivers specified as direct files, only permitting ODBC profiles (e.g. from `odbcinst.ini`)

https://github.com/spiceai/docs/issues/577


crates/model_components/src/modelsource/mod.rs
```diff
@@ -119,9 +119,10 @@ impl FromStr for ModelSourceType {
     }
 }
 
-pub fn ensure_model_path(name: &str) -> Result<String> {
+pub fn ensure_model_path(rnd_dir: &str, name: &str) -> Result<String> {
     let mut model_path = dirs::home_dir().context(UnableToFindHomeDirSnafu)?;
     model_path.push(".spice/models");
+    model_path.push(rnd_dir);
```
phillipleblanc: Why not just generate the UUID here?

crates/runtime/src/dataconnector/odbc.rs
```diff
@@ -187,7 +187,19 @@ impl DataConnectorFactory for ODBCFactory {
                     .find(|s| s.starts_with("driver="))
                     .context(NoDriverSpecifiedSnafu)?;
 
-                Ok(ODBCDriver::from(driver).into())
+                // explicitly check if the user has tried to specify a file path
+                if driver
```
phillipleblanc: Can you add some tests to cover this?

crates/model_components/src/modelsource/mod.rs
```diff
@@ -122,6 +123,7 @@ impl FromStr for ModelSourceType {
 pub fn ensure_model_path(name: &str) -> Result<String> {
     let mut model_path = dirs::home_dir().context(UnableToFindHomeDirSnafu)?;
     model_path.push(".spice/models");
+    model_path.push(Uuid::new_v4().simple().to_string());
```
sgrebnov: Will this change start re-downloading models every Spice restart as target path for a model is unique for every start now?

Sevenannn: Why model uniqueness is needed in this case?

phillipleblanc (approved):

PR #3334, opened 2024-11-01
Anthropic model regex and fix tool parsing aggregation bug

Jeadie:


crates/llms/src/anthropic/types_stream.rs
```diff
@@ -84,21 +85,23 @@ impl ContentBlock {
                 refusal: None,
                 role: None,
             },
-            ContentBlock::ToolUse(tool_use) => ChatCompletionStreamResponseDelta {
```
Jeadie, line 87: This is just refactor
```diff
- ContentBlock::ToolUse(tool_use) => ChatCompletionStreamResponseDelta {
+ ContentBlock::ToolUse(ContentBlockToolUse { id, name, .. }) => { ChatCompletionStreamResponseDelta {
```

crates/llms/src/anthropic/types_stream.rs
```diff
@@ -123,7 +126,7 @@ impl Delta {
     pub fn into_completion(
         self,
         role: &Option<MessageRole>,
-        tool_content: Option<(i32, ContentBlockToolUse)>,
```
Jeadie, line 126: Bug meant that many tool chunks weren't all being aggregated. Anthropic only ever has index=0 (can't do multiple streams in parallel).

crates/runtime/src/model/tool_use.rs
```diff
@@ -312,6 +312,9 @@ impl ToolUsingChat {
             req
         } else {
             runtime_tools.extend(req.tools.clone().unwrap_or_default());
+            // Ensure function names are unique. Tool-use recursion sometimes creates duplicates.
+            runtime_tools.sort_by(|a, b| a.function.name.cmp(&b.function.name));
```
Jeadie: In some recursions, tools would be duplicated. Anthropic doesn't like this (Openai doesn't care).

crates/llms/src/anthropic/types.rs
```diff
@@ -296,83 +296,44 @@ pub struct MetadataParam {
...
-            | AnthropicModelVariant::Claude3Haiku20240307
-            | AnthropicModelVariant::Claude21
-            | AnthropicModelVariant::Claude20
-            | AnthropicModelVariant::ClaudeInstant12 => 4096,
-        }
+// Combined pattern that matches all three formats:
+// 1. Anthropic API: claude-3-5-sonnet-20241022 or claude-3-5-sonnet-latest
+// 2. AWS Bedrock: anthropic.claude-3-5-sonnet-20241022-v2:0
+// 3. GCP Vertex AI: claude-3-5-sonnet-v2@20241022
+pub(crate) static ANTHROPIC_REGEX: &str = r"(?x) # Enable verbose mode
```
phillipleblanc, line 303: Is there a way to interrogate the anthropic endpoint for the list of models it supports? i.e. a custom anthropic-compatible API might not support all of the official ones. And if a new model is added that doesn't match our existing regex, then we need to release a new version before people can use it.

phillipleblanc (approved):
One comment about possible getting the list of models from the API instead of using a regex.

crates/llms/src/anthropic/types.rs
Jeadie: I do not believe any such API exists. The only provider I've seen is xAI.

PR #3335, opened 2024-11-02
Allow only valid WHERE conditions in vector searches

phillipleblanc:
## 🗣 Description

Validate that the HTTP API and LLM Models can only send valid WHERE expressions as inputs to the vector search component.


crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -140,7 +182,7 @@ impl SearchRequest {
         text: String,
         datasets: Option<Vec<String>>,
         limit: usize,
-        where_cond: Option<String>,
+        where_cond: Option<Expr>,
```
Jeadie, line 189: nit: add docstring that we expect this `where_cond` to be validated/sanitised. And `TryFrom` should be used otherwise.

Jeadie (dismissed):

Jeadie (approved):

Issue #3317, opened 2024-10-31
Enhancement: Semantic model (Columnar granularity)


lukekim:
\*REMEMBER, BE **SMART**!

S: Specific
M: Measurable
A: Achievable
R: Relevant
T: Time-Bound\*

## Goal-State/What/Result

Users should be able to express column level semantics about their datasets, and these details should be available to LLMs.

## Why/Purpose

Improve the ability for models to understand the semantic of the datasets they can access.

## By When

**Issue/Spec written and reviewed:** 1st November.
**Done-Done:** 5th November.

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### Spicepod DX changes
Datasets in spicepod have optional column-level (each column also optional) configuration
```yaml
datasets:
- from: file:cleaned_sales_data.parquet
  name: sales
  columns:
    - name: order_number
    - name: msrp
      description: |
        Manufacturer's Suggested Retail Price (MSRP) is the price that a product's manufacturer recommends it be sold for at the point of sale.
    - name: address_line1
      embeddings: # `column` and `column_pk` are no longer allowed.
        - use: hf_minilm
          chunking:
            enabled: true
            target_chunk_size: 256
            overlap_size: 32
          row_id: order_number
```

Additional details:
 - `datasets[].indexes` are unaffected.
 - `datasets[].embeddings` are unaffected. When provided, will be aggregrated with `datasets[].columns[].embeddings`. With v1 launch of spicepod specification, `datasets[].embeddings` to be removed.
 - Add a pseudo-PK (name TBD, e.g. `row_id`) for `datasets[].columns[].embeddings` to denote what is unique. 

### Tool changes
- Tool `table_schema` will have additional, optional parameter `output: enum` (default `full`, others `minimal`). When set, the tool will return additional metadata from `datasets[].columns[]` in the output.
- Tool `table_schema` will mention that semantic information is available via `output` parameter.

[...]

PR #3336, opened 2024-11-02
Add column field to dataset spicepod component

Jeadie:
## 🗣 Description
 - Add a new field in dataset spicepod component, `column`. Currently this will have the following:
```yaml
datasets:
- from: file:cleaned_sales_data.parquet
  name: sales
  columns:
    - name: address_line1
      description: |
        The first line of the address.
      embeddings: # `column` and `column_pk` are no longer allowed.
        - from: hf_minilm
          chunking:
            enabled: true
            target_chunk_size: 256
            overlap_size: 32
          row_id: order_number
```
 - Like `datasets[].embeddings`, does not verify the columns exist in the dataset.
 - `datasets[].columns` is optional
 - Handle unifying both `datasets[].embeddings` and `datasets[].columns.embeddings` when making `EmbeddingConnector`.
 - Update `"table_schema"` tool for new input parameter `output: ["full" "minimal"]`. Default `full`.
   - On `full`, include the `description` from `datasets[].columns` in the output. 

## 🔨 Related Issues
 - https://github.com/spiceai/spiceai/issues/3317

 - https://github.com/spiceai/docs/pull/621
 - https://github.com/spiceai/quickstarts/pull/227


crates/spicepod/src/component/dataset.rs
```diff
@@ -366,3 +369,148 @@ pub mod replication {
...
+                        .map(ToString::to_string)
+                        .ok_or_else(|| D::Error::custom("Expected a string"))
+                })
+                .collect::<Result<Vec<String>, D::Error>>()
+                .map(Some),
+            _ => Err(D::Error::custom("Invalid format for row_id")),
+        }
+    }
+
+    #[allow(clippy::unwrap_used)]
```
phillipleblanc:
```suggestion
```

Use `.expect` in tests

crates/spicepod/src/component/dataset.rs
```diff
@@ -366,3 +369,148 @@ pub mod replication {
...
+    // ```yaml
+    // row_id: foo
+    //
+    // row_id: foo, bar
+    //
+    // row_id:
+    //  - foo
+    //  - bar
+    // ```
+    fn deserialize_row_ids<'de, D>(deserializer: D) -> Result<Option<Vec<String>>, D::Error>
```
phillipleblanc, line 443: One of the issues with custom deserialization functions is that the errors they produce are sometimes not helpful. What errors occur here if the format isn't valid? Maybe that could be part of the tests below.

Jeadie: Good idea.

phillipleblanc (approved):

PR #3342, opened 2024-11-04
Remove need for `params.model_type` for most HF LLMs

Jeadie:
## 🗣 Description
 - Currently all LLMs from HF need `params.model_type` even though most often, this can be found in the `config.json` that will be downloaded during load. This PR removes this need.
 - Also keeps HF_token a `Secret` until the last possible moment.

## 🤔 Concerns
- Before this PR, if no `params.model_type` was provided, and no additional files specified, HF models would fail to load because the runtime wouldn't be able to determine if it was a LLM or ML model. This PR favours LLMs, and therefore unless explicit ML config is specified, it will determine it to be a HF LLM.


crates/llms/src/chat/mod.rs
```diff
@@ -560,36 +561,21 @@ pub trait Chat: Sync + Send {
...
-    {
-        Err(Error::FailedToRunModel {
-            source: "No Chat model feature enabled".into(),
-        })
-    }
+    mistral::MistralLlama::from_hf(
+        model_id,
+        model_type.as_deref(),
+        hf_token_literal,
+        // TODO: Support HF models with non-standard paths.
```
lukekim: Hygiene.

Jeadie: https://github.com/spiceai/spiceai/issues/3377

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #3384, opened 2024-11-05
Add `Extension<Runtime>` to HTTP routes to simplify tooling in NSQL. 

Jeadie:
## 🗣 Description
 - Standardise how we create chat messages from explicit tool calls (i.e. how we can explicitly add an LLM calling and receiving the results of a tool in `req.messages`). This is done in the function `create_tool_use_messages` and currently used in `nsql.rs`.
 - As a consequence, remove some clippy issues with `#[allow(clippy::too_many_arguments)]`.

### DIFF
--- a/crates/runtime/src/http/v1/nsql.rs
+++ b/crates/runtime/src/http/v1/nsql.rs
@@ -15,25 +15,21 @@ limitations under the License.
 */
 use crate::{
     datafusion::DataFusion,
-    http::v1::sql_to_http_response,
+    http::v1::{sql_to_http_response, ArrowFormat},
     model::LLMModelStore,
-    tools::builtin::{
-        sample::{
-            distinct::DistinctColumnsParams, random::RandomSampleParams, tool::SampleDataTool,
-            SampleTableParams,
+    tools::{
+        builtin::{
+            sample::{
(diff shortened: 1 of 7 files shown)


crates/runtime/src/lib.rs
```diff
@@ -362,7 +362,7 @@ impl Runtime {
     ///
     /// It is recommended to start the servers in parallel to loading the Runtime components to speed up startup.
     pub async fn start_servers(
-        &self,
```
Jeadie, line 365: Not sure if this is axiomatic rust

crates/runtime/src/http/v1/nsql.rs
```diff
@@ -232,41 +167,37 @@ pub(crate) async fn post(
         vec![]
     };
 
-    let sql_query_result = match llms.read().await.get(&payload.model) {
```
Jeadie, line 235: Rest of file is just moving from `match` to `let.. else { return...`

crates/runtime/src/lib.rs
phillipleblanc: Totally fine

phillipleblanc (approved):

Issue #2670, opened 2024-09-16
Enhancement: API Key auth


phillipleblanc:
## Goal-State/What/Result

All requests to the HTTP APIs require a header `X-API-Token` or query parameter `api_key` to be set to the value of an API key configured by the Spice runtime with:

```yaml
runtime:  
  auth:    
    api-key:      
      enabled: true      
      keys:        
        - HARDCODED_TOKEN
        - ${ secrets: api_key_1 }
        - ${ secrets: api_key_2 }
```

All requests to the Flight APIs require basic authentication with the password set to the value of the API above.

## Why/Purpose

Provides basic security

## By When

A target date for done-done completion of the entire enhancement.

**Issue/Spec written and reviewed:** Tue, Nov 5, 2024
**Done-Done:** Wed, Nov 6, 2024

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Spec

A set of new traits `HttpAuth`, `FlightBasicAuth` and `GrpcAuth` will be created that contains the methods that should be implemented to inspect the request (from HTTP, gRPC and Flight) and provide a verdict on whether the request should be allowed/denied. Initially there will not be any AuthZ decisions, just whether or not to allow the request.

```rust
enum AuthVerdict {
  Allow,
  Deny
}

trait HttpAuth {
  // Receive the entire HTTP request object and return a verdict on whether to allow/deny it
  fn http(req: HttpRequest) -> Result<AuthVerdict> {}
}

trait FlightBasicAuth {
  // Receive the username/password for Flight basic auth and return a verdict
  fn flight_basic(username: String, password: String) -> Result<AuthVerdict> {}
}

trait GrpcAuth {
  // Receive the entire gRPC request object and return a verdict
  fn grpc(req: GrpcRequest) -> Result<AuthVerdict> {}
}
```

[...]

PR #3397, opened 2024-11-06
Implement traits & plumbing for pluggable HTTP Auth

### LINKED ISSUE #3379: Create `runtime-auth` crate and traits for HttpAuth, FlightBasicAuth, and GrpcAuth
A trait called `EndpointAuth` is defined which contains the three methods that need to be implemented to handle auth at the API layer.

One method will be for handling requests at the Flight layer.
One method will be for handling requests at the gRPC layer (OpenTelemetry).
One method will be for handling requests at the HTTP layer.

phillipleblanc:
## 🗣 Description

This PR introduces the main traits and plumbing that enables pluggable authentication for the Spice runtime endpoints.

`runtime-auth` - A new crate where the bulk of the new authentication logic will live.

`EndpointAuth` - New struct that is constructed from an App object that will parse the configuration to determine which auth providers to enable, if any. This isn't wired up and always returns `None` for all auth providers.

`AuthVerdict` - An enum that can be either `Allow` or `Deny`. Returned by the implementation of the auth providers on their auth decision.

`HttpAuth` - A new trait that auth providers can implement that receives a request object and returns an `AuthVerdict`. Configured on the HTTP endpoints as a middleware.

`FlightBasicAuth` - A new trait that auth providers can implement that will plug into the Flight BasicAuthenticate flow to authenticate requests. This just receives a username/password and returns an `AuthVerdict`.

`GrpcAuth` - A new trait that auth providers can implement that will plug into the OpenTelemetry gRPC API to authenticate requests. This could also be used on the Flight endpoints if `FlightBasicAuth` isn't secure enough, but that is out of scope for this enhancement.

`ApiKeyAuth` - A simple implementation of `HttpAuth` that receives a list of API keys and verifies that the `X-API-Key` header is present on the request and contains one of the provided API keys. This isn't yet wired up anywhere.

## 🔨 Related Issues

Part of #2670
Closes #3379

This will require documentation, tracked in #2670

## 🤔 Concerns

Instead of returning `None` and optionally configuring auth on endpoints, I could implement a no-op Auth provider that just allows everything and the middleware always configures Auth. But I think that would be unnecessary runtime overhead if auth isn't required.

### DIFF
--- a/crates/runtime-auth/src/traits.rs
+++ b/crates/runtime-auth/src/traits.rs
@@ -30,3 +30,21 @@ pub trait HttpAuth {
     /// This function will return an error if the validator can't validate the request.
     fn http(&self, request: &http::request::Parts) -> Result<AuthVerdict, Error>;
 }
+
+pub trait FlightBasicAuth {
+    // Receive the username/password for Flight basic auth and return a verdict
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the validator can't validate the request.
+    fn flight_basic(&self, username: String, password: String) -> Result<AuthVerdict, Error>;
+}
+
+pub trait GrpcAuth {
+    // Receive the entire gRPC request object and return a verdict
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the validator can't validate the request.
+    fn grpc(&self, req: &http::request::Parts) -> Result<AuthVerdict, Error>;
+}
--- a/crates/runtime-auth/src/traits.rs
+++ b/crates/runtime-auth/src/traits.rs
@@ -1,0 +1,42 @@
+/*
+Copyright 2024 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use crate::error::Error;
+use axum::http;
+
+pub enum AuthVerdict {
+    Allow,
+    Deny,
+}
+
+pub trait HttpAuth {
+    /// Receive the entire HTTP request object and return a verdict on whether to allow/deny it
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the validator can't validate the request.
+    fn http(&self, request: &http::request::Parts) -> Result<AuthVerdict, Error>;
+}
+
+// pub trait FlightBasicAuth {
+//     // Receive the username/password for Flight basic auth and return a verdict
+//     fn flight_basic(username: String, password: String) -> Result<AuthVerdict> {}
+// }
+
+// pub trait GrpcAuth {
+//     // Receive the entire gRPC request object and return a verdict
+//     fn grpc(req: GrpcRequest) -> Result<AuthVerdict> {}
+// }
--- a/crates/runtime-auth/src/traits.rs
+++ b/crates/runtime-auth/src/traits.rs
@@ -30,13 +30,3 @@ pub trait HttpAuth {
     /// This function will return an error if the validator can't validate the request.
     fn http(&self, request: &http::request::Parts) -> Result<AuthVerdict, Error>;
 }
-
-// pub trait FlightBasicAuth {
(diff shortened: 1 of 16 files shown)


crates/runtime-auth/src/traits.rs
```diff
@@ -0,0 +1,50 @@
...
+    fn flight_basic(&self, username: String, password: String) -> Result<AuthVerdict, Error>;
+}
+
+pub trait GrpcAuth {
+    // Receive the entire gRPC request object and return a verdict
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the validator can't validate the request.
+    fn grpc(&self, req: &http::request::Parts) -> Result<AuthVerdict, Error>;
```
phillipleblanc, line 49: This signature might need to change once I wire it up, but its fine for this PR.

crates/runtime/src/http/auth.rs
```diff
@@ -0,0 +1,106 @@
...
+    auth_verifier: Arc<dyn HttpAuth + Send + Sync>,
+}
+
+impl AuthLayer {
+    pub fn new(auth_verifier: Arc<dyn HttpAuth + Send + Sync>) -> Self {
+        Self { auth_verifier }
+    }
+}
+
+impl<S> Layer<S> for AuthLayer {
```
phillipleblanc, line 39: This is a tower Layer, which you can think of as a factory for building tower Services. In this case, we are just wrapping the given tower Service with our `AuthMiddleware` Service implementation

crates/runtime/src/http/auth.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let auth_verifier = Arc::clone(&self.auth_verifier);
+        let mut inner = self.inner.clone();
+
+        Box::pin(async move {
+            let (parts, body) = req.into_parts();
+
+            match auth_verifier.http(&parts) {
+                Ok(AuthVerdict::Allow) => inner.call(Request::from_parts(parts, body)).await,
```
phillipleblanc, line 77: If we want to allow this request, we just return a future from calling the inner tower Service.

crates/runtime/src/http/auth.rs
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+fn unauthorized_response() -> Response<Body> {
+    match Response::builder()
+        .status(StatusCode::UNAUTHORIZED)
+        .body(Body::from("Unauthorized"))
+    {
+        Ok(response) => response,
+        Err(e) => panic!("Failed to build unauthorized response: {e}"),
```
phillipleblanc, line 94: This should never happen and we can't return an error here because axum middlewares should be infallible. The correct way to handle an error is to return an error response, but since this IS the error response, there isn't much we can do if it can't be constructed. I will test this in future PRs to ensure it doesn't happen.

crates/runtime/src/http/routes.rs
```diff
@@ -55,12 +55,10 @@ pub(crate) fn routes(
             "/v1/datasets/:name/acceleration",
             patch(v1::datasets::acceleration),
         )
-        .route("/v1/spicepods", get(v1::spicepods::get))
-        .route("/v1/ready", get(v1::ready::get))
-        .route_layer(middleware::from_fn(track_metrics));
```
phillipleblanc, line 60: The way that layers work is they don't apply to routes that are defined after them, so this was a subtle bug in that metrics weren't being properly tracked for the model endpoints below.

crates/runtime/src/http/routes.rs
```diff
@@ -74,13 +72,26 @@ pub(crate) fn routes(
...
         .layer(Extension(Arc::clone(&rt.app)))
         .layer(Extension(Arc::clone(&rt.df)))
         .layer(Extension(Arc::clone(rt)))
         .layer(Extension(rt.metrics_endpoint))
         .layer(Extension(config));
-    router
+
+    // If we have an auth layer, add it to the authenticated router
+    if let Some(auth_layer) = auth_layer {
+        authenticated_router = authenticated_router.route_layer(auth_layer);
```
phillipleblanc, line 84: We use `route_layer` here instead of `layer` because we only want this to run on defined routes. Otherwise it would run on routes that didn't exist and potentially return a `401 Unauthorized` instead of a `404 Not Found`.

crates/runtime/src/http/routes.rs
```diff
@@ -74,13 +72,26 @@ pub(crate) fn routes(
...
+    }
+
+    let unauthenticated_router = Router::new()
```
phillipleblanc, line 87: We separate the unauthenticated routes here and only pass in the runtime object that powers the ready endpoint.

ewgenius (approved):

PR #3398, opened 2024-11-06
Add allow_http parameter for S3 data connector

Sevenannn:
## 🗣 Description

Add `allow_http` parameter for S3 data connector. This parameter configures the object store http client to allow for HTTP protocol. This allows for connection to `s3_endpoint` with HTTP protocol, e.g. `http://my-minio-server`

The `allow_http` parameter is [defaulted to false](https://docs.rs/object_store/latest/object_store/struct.ClientOptions.html#method.with_allow_http), and won't affect any existing user of S3 connector until explicitly specify the `allow_http` parameter to be true.

## 🤔 Concerns

Should the `allow_http` parameter be prefixed with `s3_allow_http`. We don't prefix `client_timeout` parameter, which is also a object store client configuration in S3 connector. But all object store client configuration in [abfs](https://docs.spiceai.org/components/data-connectors/abfs#params) are prefixed with `abfs`. Consistency needs to be determined here.

### DIFF
--- a/crates/runtime/src/dataconnector/s3.rs
+++ b/crates/runtime/src/dataconnector/s3.rs
@@ -82,7 +82,7 @@ const PARAMETERS: &[ParameterSpec] = &[
     ParameterSpec::runtime("client_timeout")
         .description("The timeout setting for S3 client."),
     ParameterSpec::runtime("allow_http")
-        .description("Allow HTTP protocal for S3 endpoint."),
+        .description("Allow HTTP protocol for S3 endpoint."),
 
     // Common listing table parameters
     ParameterSpec::runtime("file_format"),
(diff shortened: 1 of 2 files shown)


sgrebnov (dismissed):

crates/runtime/src/dataconnector/s3.rs
```diff
@@ -81,6 +81,8 @@ const PARAMETERS: &[ParameterSpec] = &[
     ParameterSpec::connector("auth").description("Configures the authentication method for S3. Supported methods are: public (i.e. no auth), iam_role, key.").secret(),
     ParameterSpec::runtime("client_timeout")
         .description("The timeout setting for S3 client."),
+    ParameterSpec::runtime("allow_http")
+        .description("Allow HTTP protocal for S3 endpoint."),
```
lukekim: Spelling

lukekim (dismissed):

sgrebnov (approved):

Issue #3009, opened 2024-10-08
Enhancement: AI/ML integration/E2E tests on CI


lukekim:
## Goal-State/What/Result

The following AI/LLM functionality is covered by integration and end-to-end (E2E) tests, allowing us to release and update it with high confidence, while preventing and detecting bugs and regressions:
  - AI/LLM HTTP Api
  - Task History and Metrics when using AI/LLM
  - Chat functionality (OpenAI,  Anthropic, supported Local Models / Huggingface)
  - Embeddings and vector search functionality
  - Defining AI/LLM via Spicepod

Out of scope:
 - Covering traditional ML functionality by tests is not a priority (out of scope)
 - No additional reporting/alerting/dashboarding is in scope - test are run on PR and as part of release process.

## Why/Purpose

Ensuring that AI/LLM functionality works correct is absolutely crucial. Currently, this is done mostly manually, which is costly and can lead to undetected bugs. Comprehensive AI/ML testing ensures high-quality AI/LLM functionality, allowing us to detect issues earlier, release with high confidence, and move quickly.

## By When

**Issue/Spec written and reviewed:** 11/05
M1 Integration tests for core LLM functionality - 11/19
M2 E2E tests for core LLM functionality - 11/20
**Done-Done:**  11/20

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes (not required)

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### M1 Integration tests include core LLM functionality
 - [x] HTTP Api: search, nsql, chat, embeddings
 - [x] Local models downloading/initialization (huggingface)
 - [x] Model inference (chat)
    - [x] OpenAI
    - [x] HF
    - [x] Anthropic
    - [x] Supported local model types
 - [x] Tools: ListDatasetsTool, SqlTool, TableSchemaTool, DocumentSimilarityTool
 - [x] Embeddings: creation/chunking/search
 - [x] Task History when using AI/LLM
[...]

PR #3402, opened 2024-11-06
Add integration tests for OpenAI NSQL functionality

sgrebnov:
## 🗣 Description

PR adds `integration_models` tests that will include all tests related to AI/LLM functionality and adds integration test for `NSQL` functionality based on openai model.

`make test-integration-models` to run

## 🔨 Related Issues

No

## 🤔 Concerns

Tests verify that the actual result for an NSQL query is correct (i.e., the row count is calculated correct). Since it is LLM-based and not deterministic, it may be unstable. As tests are not currently enabled for PRs, we can assess stability and decide if improvements are necessary.

Tests don't include case with `sample_data_enabled: true` as it is currently failing/investigating.


crates/runtime/tests/models/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+use reqwest::Client;
+use serde_json::json;
+use spicepod::component::{
+    dataset::{acceleration::Acceleration, Dataset},
+    params::Params,
+};
+
+mod openai;
+
+fn get_taxi_trips_dataset() -> Dataset {
```
Jeadie, line 26: Question about this and integration tests in general: Why don't we define models/views/datasets etc from YAML strings?

crates/runtime/tests/models/openai.rs
```diff
@@ -0,0 +1,129 @@
...
+    .await?
+    .contains(":10}"));
+
+    tracing::info!("/v1/nsql: Ensure error when invalid dataset name is provided");
+    assert!(send_nsql_request(
+        base_url.as_str(),
+        "how many records in taxi_trips dataset?",
+        Some("nql"),
+        Some(false),
+        Some(vec!["dummy".to_string()]),
```
Jeadie: nit: change `"dummy" -> "dataset_not_in_spice"` to make it instantly understandable.

crates/runtime/tests/models/mod.rs
sgrebnov: This is just a more convenient and easy to maintain way to define such components. YAML is covered by E2E tests

crates/runtime/tests/models/openai.rs
sgrebnov: Updated - thank you

crates/runtime/tests/models/openai.rs
```diff
@@ -0,0 +1,129 @@
...
+    tracing::info!("/v1/nsql: Ensure default request succeeds");
+    assert!(send_nsql_request(
+        base_url.as_str(),
+        "how many records in taxi_trips dataset?",
+        None,
+        Some(false),
+        None,
+    )
+    .await?
+    .contains(":10}"));
```
sgrebnov, line 84: While working on `search` functionality tests I've improved this to return json and verify that json returned contains single record with value 10. The improvement will be submitted as part of next PR.

phillipleblanc (approved):

crates/runtime/tests/models/mod.rs
sgrebnov: Jack - I plan to merge this and address this via separate PR if needed.

PR #3405, opened 2024-11-07
Implement optional api-key auth for the HTTP endpoint

### LINKED ISSUE #3380: Implement `APITokenAuth` for `HttpAuth`
Implement the `APITokenAuth` struct for the trait `EndpointAuth`

### LINKED ISSUE #3381: Spicepod runtime configuration can be used to enable API Token Auth
Enables on the following endpoints:

- HTTP
- OpenTelemetry
- Flight

phillipleblanc:
## 🗣 Description

Implements an optional `api-key` authentication for the HTTP endpoint. A spicepod configured as:

```yaml
runtime:
  auth:
    api-key:
      enabled: true
      keys:
        - ${ env:API_KEY }
```

Will only accept requests on the HTTP endpoint that contain a header `X-API-Key` with a value that matches one of the keys configured in the spicepod.

i.e., if the value of `API_KEY` was `foobar`:
`curl -H "x-api-key: foobar" -XPOST -i http://localhost:8090/v1/sql -d 'SELECT 1'`

Output:
```bash
HTTP/1.1 200 OK
content-type: text/plain; charset=utf-8
x-cache: Miss from spiceai
content-length: 16
date: Thu, 07 Nov 2024 01:23:50 GMT

[{"Int64(1)":1}]
```

Otherwise for invalid keys or requests without the API key header:
`curl -XPOST -i http://localhost:8090/v1/sql -d 'SELECT 1'`

Output:
```bash
HTTP/1.1 401 Unauthorized
content-length: 12
date: Thu, 07 Nov 2024 01:26:04 GMT

Unauthorized
```

## 🔨 Related Issues

Closes #3380
Closes #3381
Part of #2670

Documentation tracked as part of #2670

## 🤔 Concerns

I expect I'll need to refactor how the EndpointAuth struct constructs the individual auth traits, but I'll handle that in a future PR. Also our CLI needs to be updated to support calling the runtime with proper auth.


Sevenannn (dismissed):

crates/runtime/src/auth/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+};
+use runtime_auth::{api_key::ApiKeyAuth, FlightBasicAuth, GrpcAuth, HttpAuth};
+use secrecy::ExposeSecret;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+pub struct EndpointAuth {
+    pub http_auth: Option<Arc<dyn HttpAuth + Send + Sync>>,
+    pub flight_basic_auth: Option<Arc<dyn FlightBasicAuth + Send + Sync>>,
+    pub grpc_auth: Option<Arc<dyn GrpcAuth + Send + Sync>>,
```
Sevenannn, line 29: Is the api key based auth going to be implemented for flight & grpc as well?

phillipleblanc: Yes, incremental PRs

crates/runtime/src/auth/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+    app: &App,
+) -> Option<Arc<dyn HttpAuth + Send + Sync>> {
+    let api_key_auth = app
+        .runtime
+        .auth
+        .as_ref()
+        .and_then(|auth| auth.api_key.as_ref())?;
+
+    let secrets = secrets.read().await;
+    let mut keys = Vec::with_capacity(api_key_auth.keys.len());
```
peasee, line 77: given this auth check runs on every request, is it possible to make it a e.g. `HashSet` instead of `Vec`?

phillipleblanc: For small numbers of elements, iterating over an array is much faster than computing a hash

peasee (approved):

PR #3410, opened 2024-11-07
Add integration tests for Search API (OpenAI and HF models)

sgrebnov:
## 🗣 Description
Add integration tests for Search API (OpenAI and HF models)

## 🔨 Related Issues

## 🤔 Concerns

Was unable to get NSQL functionality working with any of the small local models (models are not supported due to lack of `tool` functionality or unable to generate valid SQL to execute) as a result there is a test to download and initialize the `meta-llama/Llama-3.2-1B-Instruct` local model, this test will be improved to test model can be actually invoked with Spice `chat` functionality


Sevenannn:
lgtm, just wondering what are the small local models that doesn't work well to get NSQL functionality, and is this something can be solved in our code

Sevenannn (approved):

PR #3412, opened 2024-11-07
Implement optional api-key auth for the Flight/FlightSQL endpoint

### LINKED ISSUE #3382: Implement `APITokenAuth` for `FlightBasicAuth`

phillipleblanc:
## 🗣 Description

Implements API key based authentication for the Flight & FlightSQL endpoints in Spice. Also adds an `--api-key` parameter to all commands on the CLI that connect to the runtime, to allow usage when API keys are configured.

With a spicepod configured like this, it will protect both the HTTP and Flight endpoints:

```yaml
runtime:
  auth:
    api-key:
      enabled: true
      keys:
        - ${ env:API_KEY }
```

Using the CLI
```bash
❯ spice status
2024/11/07 17:29:48 ERROR getting spiced status error="error fetching runtime information: Unauthorized"
~/c/a/arrow-rs k8s:kind-kind ❯❯❯ spice status --api-key foobar

NAME          ENDPOINT        STATUS
http          127.0.0.1:8090  Ready
flight        127.0.0.1:50051 Ready
metrics       127.0.0.1:9090  Ready
opentelemetry 127.0.0.1:50052 Ready
```

Using the SQL REPL:

```bash
❯ spice sql
sql> select 1;
Authentication Error Access denied. Invalid credentials.

❯ spice sql --api-key foobar
sql> select 1;
+----------+
| Int64(1) |
+----------+
| 1        |
+----------+

Time: 0.007247375 seconds. 1 rows.
```

## 🔨 Related Issues

Closes #3382
Part of #2670

Documentation is tracked by #2670

## 🤔 Concerns

I kept the behavior that we have on Spice.ai for the authentication with Flight Basic Auth. This also matches what Influx does for their authentication support, as far as I can tell. I will test with our existing SDKs/DBeaver before merging this PR.


crates/runtime/src/flight/auth.rs
```diff
@@ -0,0 +1,225 @@
...
+use tonic::{
+    metadata::{MetadataMap, GRPC_CONTENT_TYPE},
+    Status,
+};
+use tower::{Layer, Service};
+
+/// The handshake request for Flight basic auth is a base64 encoded token that is calculated like:
+///
+/// ```rust,no_run
+/// let val = BASE64_STANDARD.encode(format!("{username}:{password}"));
```
eadgbear, line 36: Are we enabling Basic Auth anywhere in the project?

phillipleblanc: Basic Auth is a fairly standard way for clients to connect to FlightSQL endpoints. Its how our SDKs work for API key auth, and how the FlightSQL JDBC driver works https://arrow.apache.org/docs/java/flight_sql_jdbc_driver.html

sgrebnov (approved):

crates/runtime-auth/src/api_key/mod.rs
```diff
@@ -46,6 +46,24 @@ impl HttpAuth for ApiKeyAuth {
...
 
+impl FlightBasicAuth for ApiKeyAuth {
+    fn validate(&self, _username: &str, password: &str) -> Result<String, Error> {
+        if self.api_keys.iter().any(|key| key == password) {
+            Ok(password.to_string())
+        } else {
+            Err(Error::InvalidCredentials)
+        }
+    }
+
```
Sevenannn, line 57: For learning purpose: Are password and bearer token in `ApiKeyAuth` both the same value as api_key? Why the 2 different notions (password, bearer token) are both used here?

Sevenannn (approved):

bin/spice/cmd/catalogs.go
```diff
@@ -51,5 +56,6 @@ spice catalogs
 
 func init() {
 	catalogsCmd.Flags().String("tls-root-certificate-file", "", "The path to the root certificate file used to verify the Spice.ai runtime server certificate")
+	catalogsCmd.Flags().String("api-key", "", "The API key to use for authentication")
```
Jeadie, line 59: You can use `RootCmd.PersistentFlags().String("api-key", "", "The API key to use for authentication")` to avoid defining the flag for each CLI command?

Jeadie:
This has implications for our SDKs, right?

Jeadie (approved):

bin/spice/cmd/catalogs.go
phillipleblanc: Not all of the CLI commands make sense to have an API key though

phillipleblanc:
> This has implications for our SDKs, right?

Yes, but because the Spice.ai Cloud Platform uses the same auth that is being defined here, they already all support it. They will need some tweaking to ensure it works well with the local runtime, but its mostly cosmetic changes required.

crates/runtime-auth/src/api_key/mod.rs
phillipleblanc: Yes, this is how our auth works in the Spice.ai Cloud Platform. It allows us to conform to the Flight Basic Auth specification, which can technically be different (i.e. correct user + password, get issued a token) - however for API key auth we don't want to have to issue and keep track of tokens.

PR #3416, opened 2024-11-08
Add integration tests for Embeddings API (OpenAI and HF models)

sgrebnov:
## 🗣 Description

Add integration tests for Embeddings API (OpenAI and HF models)

## 🔨 Related Issues


crates/runtime/tests/models/openai.rs
```diff
@@ -195,6 +200,88 @@ async fn openai_search_test() -> Result<(), anyhow::Error> {
...
+            Some("float"),
+            None,
+            None,
+        ),
+        (
+            EmbeddingInput::StringArray(vec![
+                "The food was delicious".to_string(),
+                "and the waiter...".to_string(),
+            ]),
+            // Some("base64"), Error: 'When encoding_format is base64, use Embeddings::create_base64'
```
sgrebnov, line 233: @Jeadie - could you please confirm `base64` is the right use case for embeddings API so I file an issue and address it

crates/runtime/tests/models/hf.rs
```diff
@@ -142,6 +144,94 @@ async fn huggingface_model_download_test() -> Result<(), anyhow::Error> {
...
+            "hf_minilm",
+        ))
+        .with_embedding(get_huggingface_embeddings("intfloat/e5-small-v2", "hf_e5"))
+        .build();
+
+    let http_port = rand::thread_rng().gen_range(50000..60000);
+
+    tracing::debug!("Running Spice runtime with http port: {http_port}");
+
+    let api_config = Config::new().with_http_bind_address(SocketAddr::new(LOCALHOST, http_port));
```
phillipleblanc: You'll need to specify all of the ports - otherwise the ones you don't specify get their default values, which will conflict in other tests.

crates/runtime/tests/models/openai.rs
```diff
@@ -195,6 +200,88 @@ async fn openai_search_test() -> Result<(), anyhow::Error> {
...
+            Some("text-embedding-3-small"),
+            "openai_embeddings",
+        ))
+        .build();
+
+    let http_port = rand::thread_rng().gen_range(50000..60000);
+
+    tracing::debug!("Running Spice runtime with http port: {http_port}");
+
+    let api_config = Config::new().with_http_bind_address(SocketAddr::new(LOCALHOST, http_port));
```
phillipleblanc: You'll need to specify all of the ports - otherwise the ones you don't specify get their default values, which will conflict in other tests.

phillipleblanc (dismissed):
Looks good to me, please get @Jeadie's review before merging.

crates/runtime/tests/models/mod.rs
```diff
@@ -164,3 +165,59 @@ fn normalize_search_response(mut json: Value) -> String {
 
     json.to_string()
 }
+
+fn normalize_embeddings_response(mut json: Value) -> String {
```
Jeadie, line 195: This probably needs a docstring. If i understand what's happening, we're replacing the actual embedding vector, with its length?

sgrebnov: Documented - thank you!

crates/runtime/tests/models/hf.rs
sgrebnov: Good point - thank you. Added helper function. Plan to update other integration tests to use this helper function as separate PR.

```
fn create_test_api_bindings_config() -> Config {
    let mut rng = rand::thread_rng();
    let http_port: u16 = rng.gen_range(50000..60000);
    let flight_port: u16 = http_port + 1;
    let otel_port: u16 = http_port + 2;
    let metrics_port: u16 = http_port + 3;

    let localhost: IpAddr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));

    let api_config = Config::new()
        .with_http_bind_address(SocketAddr::new(localhost, http_port))
        .with_flight_bind_address(SocketAddr::new(localhost, flight_port))
        .with_open_telemetry_bind_address(SocketAddr::new(localhost, otel_port));

    tracing::debug!(
        "Created api bindings configuration: http: {http_port}, flight: {flight_port}, otel: {otel_port}, metrics: {metrics_port}"
    );

    api_config
}
```

crates/runtime/tests/models/openai.rs
sgrebnov: Updated, thank you

Sevenannn (approved):

PR #3418, opened 2024-11-08
Support base64 embedding format

Jeadie:
## 🗣 Description
 - Support ` "encoding_format": "base64"` for `/v1/embeddings`. 

## 🔨 Related Issues
- Depends on https://github.com/spiceai/async-openai/pull/4


Cargo.toml
```diff
@@ -1,27 +1,27 @@
 [workspace]
 default-members = ["bin/spiced"]
 members = [
-  "bin/spiced/",
```
sgrebnov, line 4: I don't see any new references, so assume it is just a formatting change, also scanned the changes and they look good.

sgrebnov (approved):

sgrebnov (dismissed):

crates/llms/src/openai/embed.rs
```diff
@@ -39,9 +39,13 @@ impl Embed for Openai {
         &self,
         req: CreateEmbeddingRequest,
     ) -> Result<CreateEmbeddingResponse, OpenAIError> {
+        let outer_model = req.model.clone();
         let mut inner_req = req.clone();
         inner_req.model.clone_from(&self.model);
-        self.client.embeddings().create(inner_req).await
+        let mut resp = self.client.embeddings().create(inner_req).await?;
+
+        resp.model = outer_model;
```
Sevenannn, line 47: Changes lgtm, just wondering why the model needs to be cloned and explicitly set here

Sevenannn (dismissed):

crates/llms/src/openai/embed.rs
Jeadie: This is because the response we get back from openai sets the model to the internal model (e.g. `gpt-4o`), but we want the response to contain the name of the model as defined in the spicepod. See the lifecycle of the request/response here.

```
req { model: "spicepod.model.name"}  # Request provided from user

req { model: "spicepod.model.from.split(":")[1]}  # We override model_name in request to OpenAI specific model

resp { model: "spicepod.model.from.split(":")[1]}  # Response contains OpenAI specific model

resp { model:  "spicepod.model.name"}  # We override back to user model name
```

sgrebnov (approved):

PR #3428, opened 2024-11-10
Have views update on `--pods-watcher-enabled`

Jeadie:
## 🗣 Description
 - Using `spiced --pods-watcher-enabled` update views if they are changed.
 - Also, have `view` component statuses (like datasets, models, etc). 
 - When updating views, handle dependencies with views (topological ordering of the DAG). 
 - Refactor `Runtime::start_pods_watcher` for function length.

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -1653,7 +1653,7 @@ impl Runtime {
 
         // Get ordering of views to load, including those unchanged but with dependencies that have changed
         // If we can't determine the order, we'll just load the views in the order they are in the app
-        let afffected_views_in_order_of_dependencies = match valid_views
+        let affected_views_in_order_of_dependencies = match valid_views
             .iter()
             .map(|v| {
                 let Some(statement) =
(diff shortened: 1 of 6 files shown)


crates/runtime/src/lib.rs
```diff
@@ -1565,6 +1593,149 @@ impl Runtime {
         Ok(())
     }
 
+    async fn apply_model_diff(&self, current_app: &Arc<App>, new_app: &Arc<App>) {
```
Jeadie, line 1596: Except for `apply_view_diff`, this is just refactored out of [start_pods_watcher](https://github.com/spiceai/spiceai/pull/3428/files#diff-c6d4f276260e77036ce0adaf946b9ee6e55b1af1640342035121d4b254090df8R1739).

crates/runtime/src/lib.rs
```diff
@@ -1565,6 +1593,153 @@ impl Runtime {
...
+                };
+                self.status
+                    .update_view(&view.name, status::ComponentStatus::Disabled);
+                self.remove_view(&view);
+            }
+        }
+
+        // Get ordering of views to load, including those unchanged but with dependencies that have changed
+        // If we can't determine the order, we'll just load the views in the order they are in the app
+        let afffected_views_in_order_of_dependencies = match valid_views
```
```suggestion
        let affected_views_in_order_of_dependencies = match valid_views
```

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #3430, opened 2024-11-11
feat: Add invalid type action handling for DuckDB

### LINKED ISSUE #3391: Enhancement: Enable DuckDB to warn for unsupported data types
## Goal-State/What/Result

DuckDB should have a configuration option that enables them to warn for column data types in tables that are unsupported.

## Why/Purpose

To improve robustness of the DuckDB connector/accelerator for queries on tables that contain unsupported data types.

## By When

**Issue/Spec written and reviewed:** 6th Nov, 2024
**Done-Done:** 8th Nov, 2024

## Done-Done

- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [ ] Tests
- [ ] Documentation
- [ ] Samples/Quickstarts/Tutorials
- [ ] Release Notes

## The Algorithm

- [x] Simplify.

## Specification

`get_schema()` should be updated to detect for unsupported column data types from tables.

In the default configuration, unsupported column data types should result in an error.

By using a configuration option in the spicepod, users can override this behavior and choose to silence the error. In this case, it will produce a warning log instead.

### UX Changes

* A new spicepod configuration parameter is introduced for the DuckDB connector and accelerator as a Dataset parameter.

This parameter configures the behavior of silencing unsupported column data types:

```
datasets:
  - from: duckdb:foobar
    name: my_table
    invalid_type_action: error # warn, ignore
    params:
        duckdb_file: myfile.db
```

The default value of `invalid_type_action` is `error`.

When set to:

* `error`: The connector will return an error when encountering a data type that is unsupported, and the table will fail to load
* `warn`: The connector will log a warning when encountering a data type that is unsupported. The table will load, and exclude the column with the unsupported data type.
* `ignore`: The connector will log no output when encountering a data type that is unsupported. The table will load, and exclude the column with the unsupported data type.

When set on a connector that does not support the parameter, the Runtime should error.

### Security Review

* This enhancement does not have any security considerations.

## How/Plan

`get_schema()` is updated in the `TableProvider` for DuckDB to detect and handle unsupported column types per the configured spicepod parameter (or default).

The result of `result.get_schema()` is checked before returning to detect any data types that are unsupported. The configuration option needs to be passed through to the `DuckDBTableProviderFactory`, which in turn passes the parameter to the connection factory.

The dataset parameter is added like `on_invalid_types: Option<OnInvalidType>`. Connectors that do not support the parameter should validate if the parameter has been set like `on_invalid_types.is_some()`, and return an error if it has been set.

peasee:
## 🗣 Description

* Updates `DataConnectorFactory::create()` to accept the new `DataConnectorParams`, a wrapper around parameters used to setup data connectors
* Adds `DataConnectorFactory::supports_invalid_type_action()` for a data connector to identify if it supports this configuration option. Defaults to `false`.
* Adds `invalid_type_action` as a dataset-level parameter, including a `From` implementation to convert it to the `datafusion_table_providers::InvalidTypeAction`.
* Updates the DuckDB connector factory to pass through the spicepod parameter to the DuckDB action handle, to configure how unsupported data types are handled.

## 🔨 Related Issues

* Closes #3391 
* Part of #3347 

Needs a spicepod reference update/parameters update for DuckDB?


peasee:
Benchmark run: https://github.com/spiceai/spiceai/actions/runs/11772848316

crates/spicepod/Cargo.toml
```diff
@@ -16,6 +16,7 @@ serde_yaml.workspace = true
 serde-value = "0.7.0"
 snafu.workspace = true
 tracing.workspace = true
+datafusion-table-providers.workspace = true
```
phillipleblanc: I don't think this dependency makes sense

Sevenannn:
Is the `invalid_type_action` parameter only for connector? Since it's part of `DataConnectorParameter`. I wonder how we would distinguish a `invalid_type_action` for data connector and data accelerator, since it's currently a dataset level parameter.

peasee:
> Is the `invalid_type_action` parameter only for connector? Since it's part of `DataConnectorParameter`. I wonder how we would distinguish a `invalid_type_action` for data connector and data accelerator, since it's currently a dataset level parameter.

The parameter is currently only valid for data connectors.

phillipleblanc (approved):

PR #3433, opened 2024-11-11
Add integration tests for chat completion API (HF and OpenAI)

sgrebnov:
## 🗣 Description

PR adds integration tests for chat completion API (HF and OpenAI):
 - test http completion api
 - instantiating model manually and performing `CreateChatCompletionRequest` - tests  `_test_chat_messages` will be extended with more scenarios via creating chat models with different settings using `ToolUsingChat`  w/o restarting runtime and re-initiating the model.
 - verification that expected tools have been called and traced to `task_history` for openai
 
## 🔨 Related Issues

## 🤔 Concerns

HF models used for testing (`microsoft/Phi-3-mini-4k-instruct` and `Llama-3.2-1B-Instruct`) don't support tools as a result models can't provide the expected response so Hugingface test verifies that chat completion is completed only w/o verification of tools called (as it is done for openai). This will be improved once appropriate model is identified.

`test_chat_completion` and `test_chat_messages` look similar; the difference is that `test_chat_completion` verifies   chat completion http api and `test_chat_messages` constructs model manually and will be extended with more scenarios verifying different tools usage and messages format w/o restarting the runtime.


crates/runtime/Cargo.toml
```diff
@@ -130,6 +130,8 @@ async-graphql-axum = "7.0.5"
 bollard = "0.16.1"
 spice-cloud = { path = "../spice_cloud" }
 tracing-subscriber.workspace = true
+tracing-opentelemetry.workspace = true
+opentelemetry_sdk = { workspace = true, features = ["rt-tokio-current-thread"] }
```
sgrebnov, line 140: Feature is required for [TokioCurrentThread](https://docs.rs/opentelemetry_sdk/0.24.1/opentelemetry_sdk/runtime/struct.TokioCurrentThread.html#). With Tokio runtime tests hang similar to https://github.com/open-telemetry/opentelemetry-rust/issues/868

crates/runtime/src/model/mod.rs
```diff
@@ -26,6 +26,7 @@ mod tool_use;
 
 pub use chat::{try_to_chat_model, LLMModelStore};
 pub use embed::{try_to_embedding, EmbeddingModelStore};
+pub use tool_use::ToolUsingChat;
```
sgrebnov, line 29: Made public so it could be used for chat model creation without the need to re-instantiate the underlying model and restart runtime configuration.

crates/runtime/tests/models/hf.rs
```diff
@@ -222,6 +234,106 @@ async fn huggingface_test_embeddings() -> Result<(), anyhow::Error> {
...
+            ("system".to_string(), "You are an assistant that responds to queries by providing only the requested data values without extra explanation.".to_string()),
+            ("user".to_string(), "Provide the total number of records in the taxi_trips dataset. If known, return a single numeric value.".to_string()),
+        ],
+        "hf_model",
+        false,
+    ).await?;
+
+    // Result verification is disabled due to issue below: model does not use tools and can't provide the expected response.
+    // https://github.com/spiceai/spiceai/issues/3426
+    // assert_eq!(response["choices"][0]["message"]["content"].as_str(), Some("10"));
```
phillipleblanc: Should we assert on a snapshot with the content replaced, like the below test?

phillipleblanc (dismissed):

crates/runtime/tests/models/hf.rs
sgrebnov: That makes sense - thank you, added snapshost based verification step for both HF and OpenAI:
https://github.com/spiceai/spiceai/pull/3433/commits/73e30f953e322b14c293f6c8f20d1d31c12491c9

crates/runtime/tests/models/openai.rs
```diff
@@ -262,6 +268,134 @@ async fn openai_test_embeddings() -> Result<(), anyhow::Error> {
...
+                    .into(),ChatCompletionRequestUserMessageArgs::default()
+                    .content("Provide the total number of records in the taxi trips dataset. If known, return a single numeric value.".to_string())
+                    .build()?
+                    .into()])
+                .build()?;
+
+    let task_start_time = std::time::SystemTime::now();
+    let response = tool_model.chat_request(req).await?;
+
+    insta::assert_snapshot!("chat_1_choices", format!("{:?}", response.choices));
```
eadgbear, line 372: Is the possibility of hallucination/different response values going to cause problems with `insta`?

eadgbear (approved):

crates/runtime/tests/models/openai.rs
sgrebnov: This is possible but the probability should be very low, I'll be checking this to see how stable it is and improve if needed, so far it worked well for me.

PR #3446, opened 2024-11-11
Use spiceai-large-runners to build benchmark binary

Sevenannn:
## 🗣 Description

Use runner group `spiceai-large-runners` to speed up building benchmark binary.
This will reduce the benchmark bianry build time to ~10 minutes.
Successful run: https://github.com/spiceai/spiceai/actions/runs/11785897780

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -66,6 +66,9 @@ jobs:
       - name: Install Protoc
         uses: arduino/setup-protoc@v3
 
+      - name: Install Required Packages
+        uses: sudo apt install build-essential
+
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -66,6 +66,9 @@ jobs:
       - name: Install Protoc
         uses: arduino/setup-protoc@v3
 
+      - name: Set up cc
```
lukekim: What is a CC? Make this self-documenting.

Jeadie (approved):

phillipleblanc (approved):

PR #3458, opened 2024-11-12
Tool use and model name for local models

Jeadie:
## 🗣 Description
 - Ensure the model name from spicepod is returned in the LLM response.
 - Pass all chat/completion stream request parameters to model pipeline. This requires a refactor of functions. 
 - Ensure we return `tool_calls` finish reason when appropriate. 
 - Fix `created` response. It should be seconds, but mistralrs uses milliseconds.


crates/llms/src/chat/mistral.rs
```diff
@@ -356,38 +386,44 @@ impl MistralLlama {
...
+    /// Process a single `stream=false` request to generate `OpenAi` chat completion.
+    fn post_process_req(resp: MistralResponse) -> Result<ChatCompletionResponse> {
+        match resp {
+            MistralResponse::Done(mut resp) => {
+                // mistralrs does not return "tool_calls" as a finish_reason correctly (like OpenAI spec).
+                // This is a workaround to set it correctly.
+                resp.choices.iter_mut().for_each(|c| {
+                    if c.finish_reason == "stop" && !c.message.tool_calls.is_empty() {
+                        c.finish_reason = "tool_calls".to_string();
+                    }
```
Sevenannn, line 408: Is the `finish_reason==stop` and `!tool_calls.is_empty()` sufficient to prove that the `finish_reason` is actually tool_call, could the finish reason something else (e.g. content filter) when both requirements are meet in this case.

crates/llms/src/chat/mistral.rs
```diff
@@ -340,14 +343,41 @@ impl MistralLlama {
         })
     }
 
-    async fn run_internal(
+    /// Prepares and sends a [`CreateChatCompletionRequest`] to the model pipeline.
+    #[allow(deprecated, clippy::cast_possible_truncation)]
```
sgrebnov: @Jeadie  - could you please add a note why `deprecated` api is used,

crates/llms/src/chat/mistral.rs
```diff
@@ -340,14 +343,41 @@ impl MistralLlama {
...
+                Stop::String(s) => mistralrs::StopTokens::Seqs(vec![s]),
+                Stop::StringArray(s) => mistralrs::StopTokens::Seqs(s),
+            }),
+            max_len: req.max_completion_tokens.map(|x| x as usize),
+            // logits_bias: req.logit_bias,
+            logits_bias: None,
+            n_choices: req.n.unwrap_or(1) as usize,
+            dry_params: None,
+        };
+        let (snd, rcv) = channel::<MistralResponse>(10_000);
```
sgrebnov: (tx, rx) is a more common naming pattern

crates/llms/src/chat/mistral.rs
```diff
@@ -407,141 +443,96 @@ impl Chat for MistralLlama {
...
+                r#type: None,
+                param: None,
+                code: None,
+            }));
         };
 
-        let resp = self
-            .run_internal(messages, tools, tool_choice, Some(p))
-            .await
+        let z = Self::post_process_req(resp)
```
sgrebnov: Naming: `z` is hard to read/understand the purpose, please see if there could be a better alternative

sgrebnov (dismissed):

crates/llms/src/chat/mistral.rs
```diff
@@ -340,14 +343,41 @@ impl MistralLlama {
...
+            min_p: None,
+            top_n_logprobs: req.top_logprobs.unwrap_or_default().into(),
+            frequency_penalty: req.frequency_penalty,
+            presence_penalty: req.presence_penalty,
+            stop_toks: req.stop.map(|s| match s {
+                Stop::String(s) => mistralrs::StopTokens::Seqs(vec![s]),
+                Stop::StringArray(s) => mistralrs::StopTokens::Seqs(s),
+            }),
+            max_len: req.max_completion_tokens.map(|x| x as usize),
+            // logits_bias: req.logit_bias,
```
phillipleblanc:
```suggestion
```

crates/llms/src/chat/mistral.rs
```diff
@@ -562,6 +553,102 @@ impl Chat for MistralLlama {
     }
 }
 
+fn stream_from_response(
+    mut rcv: Receiver<MistralResponse>,
+) -> Pin<Box<dyn Stream<Item = Result<CreateChatCompletionStreamResponse, OpenAIError>> + Send>> {
+    Pin::from(Box::new(stream! {
```
phillipleblanc:
```suggestion
    Box::pin(stream! {
```

crates/llms/src/chat/mistral.rs
Jeadie: no, ` !tool_calls.is_empty(` is pretty definitive in my experience. There are other stop reasons for these cases `ContentFilter` or `Stop` if `max_completion_tokens` is reached.

sgrebnov (approved):

PR #3461, opened 2024-11-12
Fix `TableReference` quoting for MySQL

Jeadie:
## 🗣 Description
 - `TableReference::parse_str(` does not handle non-default quotations. MySQL has non-default quotations. 

## 🔨 Related Issues
 - Need https://github.com/datafusion-contrib/datafusion-table-providers/pull/166

### DIFF
--- a/crates/runtime/src/component/dataset.rs
+++ b/crates/runtime/src/component/dataset.rs
@@ -356,7 +356,6 @@ impl Dataset {
     /// For [`Dataset`]s where the path in the `from` field is a [`TableReference`], parse and return the [`TableReference`].
     ///
     ///
-    #[must_use]
     pub fn parse_path(
         &self,
         case_sensitive: bool,
--- a/crates/runtime/src/component/dataset.rs
+++ b/crates/runtime/src/component/dataset.rs
@@ -357,7 +357,7 @@ impl Dataset {
     ///
     ///
     #[must_use]
-    pub fn table_reference_path(
+    pub fn parse_path(
         &self,
         case_sensitive: bool,
         dialect: Option<&dyn Dialect>,
--- a/crates/runtime/src/component/dataset.rs
+++ b/crates/runtime/src/component/dataset.rs
@@ -17,7 +17,13 @@ limitations under the License.
 use acceleration::Engine;
 use app::App;
 use arrow::datatypes::SchemaRef;
-use datafusion::sql::TableReference;
+use datafusion::sql::{
+    sqlparser::{
+        dialect::{Dialect, GenericDialect},
+        parser::{Parser, ParserError},
+    },
+    TableReference,
+};
 use datafusion_table_providers::util::column_reference;
 use snafu::prelude::*;
 use spicepod::component::{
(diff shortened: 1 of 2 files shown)


crates/runtime/src/component/dataset.rs
```diff
@@ -344,6 +353,56 @@ impl Dataset {
...
+    ///
+    ///
+    #[must_use]
+    pub fn table_reference_path(
+        &self,
+        case_sensitive: bool,
+        dialect: Option<&dyn Dialect>,
+    ) -> Result<TableReference> {
+        // Manually parse the table reference to avoid case folding.
+        if case_sensitive {
```
sgrebnov, line 366: Might be better to re-use Datafusion implementation?

https://github.com/apache/datafusion/blob/main/datafusion/common/src/utils/mod.rs#L276

```
pub(crate) fn parse_identifiers_normalized(s: &str, ignore_case: bool) -> Vec<String> {
    parse_identifiers(s)
        .unwrap_or_default()
        .into_iter()
        .map(|id| match id.quote_style {
            Some(_) => id.value,
            None if ignore_case => id.value,
            _ => id.value.to_ascii_lowercase(),
        })
        .collect::<Vec<_>>()
}
```

crates/runtime/src/component/dataset.rs
```diff
@@ -344,6 +353,56 @@ impl Dataset {
...
+            let path_str = self.path();
+            let dialect = dialect.unwrap_or(&GenericDialect {});
```
sgrebnov, line 368: Can we make this DF method public and re-use it instead of adding own implementation?

https://github.com/apache/datafusion/blob/main/datafusion/common/src/utils/mod.rs#L276

```
pub(crate) fn parse_identifiers(s: &str) -> Result<Vec<Ident>> {
    let dialect = GenericDialect;
    let mut parser = Parser::new(&dialect).try_with_sql(s)?;
    let idents = parser.parse_multipart_identifier()?;
    Ok(idents)
}

pub(crate) fn parse_identifiers_normalized(s: &str, ignore_case: bool) -> Vec<String> {
    parse_identifiers(s)
        .unwrap_or_default()
        .into_iter()
        .map(|id| match id.quote_style {
            Some(_) => id.value,
            None if ignore_case => id.value,
            _ => id.value.to_ascii_lowercase(),
        })
        .collect::<Vec<_>>()
}
```

crates/runtime/src/component/dataset.rs
```diff
@@ -344,6 +353,56 @@ impl Dataset {
         }
     }
 
+    /// For [`Dataset`]s where the path in the `from` field is a [`TableReference`], parse and return the [`TableReference`].
+    ///
+    ///
+    #[must_use]
+    pub fn table_reference_path(
```
sgrebnov: Implementation LGTM and [is aligned to DF implementation](https://github.com/apache/datafusion/blob/main/datafusion/common/src/utils/mod.rs#L276). Not sure this should be part of Dataset component - might be better to add as utility function so it can be re-used

sgrebnov (dismissed):

crates/runtime/src/component/dataset.rs
sgrebnov:
```suggestion
    pub fn parse_path(
```

Jeadie: I did think of that, but 1. there's not a great place to put it, 2. I don't know how encouraging we should be able it being used alot (if it's needed alot it might be a sign of structural problems).

sgrebnov (dismissed):

sgrebnov (approved):

PR #3479, opened 2024-11-13
Update benchmark workflow to create PR for updating snapshot

Sevenannn:
## 🗣 Description

Add a choice to update snapshots for benchmarks in benchmark workflow.

Benchmarks can be updated by choosing `always` in the following choice
image

When choosing `always`, benchmark test will be run with INSTA_UPDATE=always, changes will be uploaded to branch `bench-snapshot-update/${{ github.run_id }}` and PRs will be created to update snapshots for the benchmark

Successful run using a subset of benchmarks can be viewed here: https://github.com/spiceai/spiceai/pull/3480

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -356,7 +356,7 @@ jobs:
           git config --global user.email 'bench-snapshot-update-bot@spice.ai'
           git checkout -b bench-snapshot-update/${GITHUB_RUN_ID}
           git add '*.snap'
-          git commit -m "Update benchmark snapshots for ${{ matrix.name }} and ${{ matrix.bench }}""
+          git commit -m "Update benchmark snapshots for ${{ matrix.name }} and ${{ matrix.bench }}"
           git pull --rebase origin bench-snapshot-update/${{ github.run_id }}
           git push origin bench-snapshot-update/${GITHUB_RUN_ID}
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -387,3 +410,21 @@ jobs:
           CLICKBENCH_S3_ENDPOINT: ${{ secrets.CLICKBENCH_S3_ENDPOINT}}
           CLICKBENCH_S3_KEY: ${{ secrets.CLICKBENCH_S3_KEY}}
           CLICKBENCH_S3_SECRET: ${{ secrets.CLICKBENCH_S3_SECRET}}
+
+  create-pr:
+    runs-on: ubuntu-latest
+    needs: run-database-bench
+    if: always()
```
phillipleblanc: Should only create the PR when `github.event.inputs.update_snapshots == 'always'`?

Sevenannn: Nice catch! Let me update that

phillipleblanc (approved):

.github/workflows/benchmarks.yml
Sevenannn: Updated to `always() && github.event.inputs.update_snapshots == 'always'`, since we want the `create-pr` to run regardless of failure in `run-database-bench` and only run when `github.event.inputs.update_snapshots == 'always'`

Issue #3393, opened 2024-11-06
Enhancement: Add an option to return SQL query instead of query results in /v1/nsql


ewgenius:
`/v1/nsql` always returns query results. In some cases (e.g. for code completion), it would be useful to have only a generated query.

Jeadie:
This can, or similar, likely be achieved via a standard chat/completion where, if we know which dataset they are using, tell the LLM to sample before completing text to SQL (and with a hardcoded "system prompt" in the user message). For example:

```shell
curl http://127.0.0.1:8090/v1/chat/completions \
-H "Content-Type: application/json" \
-d '{
  "model": "nql",  "tool_choice": {"type": "function", "function": {"name": "table_schema"}},
  "messages": [
    {
      "role": "user",
      "content": ""Task: {task_description}. Instruction: Return only valid SQL code, nothing additional. Columns with capitals must be quoted. For tables with schemas and catalogs '"catalog"."schema"."table"' not '"catalog.schema.table"'."}
  ]}'
```

PR #3487, opened 2024-11-13
Option to return sql from `v1/nsql`

Jeadie:
## 🗣 Description
 - Option to return SQL from `v1/nsql` instead of running the SQL query generated. 

### Example
```shell
>> curl http://localhost:8090/v1/nsql -XPOST \
   -H 'Accept: application/sql' \
   -d '{"query": "write valid SQL that is also ASCII art"
}'

SELECT
'##### ### ###  #######\n#/     \#\ #/  \##/#/ \ #\n#     ###  #   ########.  ## #' AS "SQL_ART"
FROM "sales"
WHERE "order_number" = 10107
```
 

## 🔨 Related Issues
 - #3393 

 - [x] We don't document `v1/nsql`.


crates/runtime/src/http/v1/nsql.rs
```diff
@@ -200,6 +210,12 @@ pub(crate) async fn post(
     match sql_gen.parse_response(resp) {
         Ok(Some(model_sql_query)) => {
             let cleaned_query = clean_model_based_sql(&model_sql_query);
+
+            if return_sql_only(&accept) {
+                tracing::trace!("Not running query, returning SQL:\n{cleaned_query}");
```
lukekim:
```suggestion
                tracing::trace!("Not running query, requested SQL only:\n{cleaned_query}");
```

lukekim (dismissed):

crates/runtime/src/http/v1/nsql.rs
```diff
@@ -119,6 +122,13 @@ fn default_model() -> String {
     "nql".to_string()
 }
 
+/// Checks if the request is asking to only generate SQL.
+fn return_sql_only(accept: &Option<TypedHeader<Accept>>) -> bool {
+    accept
+        .as_ref()
+        .is_some_and(|a| accept_header_types(a).contains(&"application/sql".to_string()))
```
phillipleblanc, line 129: TIL that `application/sql` is a valid mime type: https://www.rfc-editor.org/rfc/rfc6922.html

phillipleblanc (approved):

crates/runtime/src/http/v1/nsql.rs
Jeadie: Yes, TIL too

PR #3500, opened 2024-11-14
Add vector search support for Utf8View type columns

sgrebnov:
## 🗣 Description

Add vector search support for Utf8View type columns

### DIFF
--- a/crates/runtime/src/embeddings/vector_search.rs
+++ b/crates/runtime/src/embeddings/vector_search.rs
@@ -17,7 +17,7 @@ limitations under the License.
 use std::{collections::HashMap, fmt::Display, sync::Arc};
 
 use app::App;
-use arrow::array::{RecordBatch, StringArray};
+use arrow::array::{RecordBatch, StringArray, StringViewArray};
 use arrow::error::ArrowError;
 use arrow::util::pretty::pretty_format_batches;
 use arrow_schema::{Schema, SchemaRef};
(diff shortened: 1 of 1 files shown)


crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -345,7 +345,13 @@ impl VectorSearchTableResult {
...
             .iter()
             .flat_map(|v| {
-                if let Some(col) = v.column(0).as_any().downcast_ref::<StringArray>() {
+                let embedded_column = v.column(0);
+                if let Some(col) = embedded_column.as_any().downcast_ref::<StringArray>() {
+                    col.iter()
+                        .map(|v| v.unwrap_or_default().to_string())
+                        .collect::<Vec<String>>()
+                } else if let Some(col) = embedded_column.as_any().downcast_ref::<StringViewArray>()
+                {
```
lukekim, line 354: Weird styling. No prettier?

crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -345,7 +345,13 @@ impl VectorSearchTableResult {
         let result = embedding_records
             .iter()
             .flat_map(|v| {
-                if let Some(col) = v.column(0).as_any().downcast_ref::<StringArray>() {
+                let embedded_column = v.column(0);
```
lukekim, line 348: Could this have a test?

lukekim (approved):

crates/runtime/src/embeddings/vector_search.rs
sgrebnov: Confirm I'll add a test, thx 👍

crates/runtime/src/embeddings/vector_search.rs
sgrebnov: Going to merge this as-is and review how to refactor this- this pattern is used in several places in vector search module. 

```
col.iter()
  .map(|v| v.unwrap_or_default().to_string())
  .collect::<Vec<String>>()
```

I also see that there are places where we default to empty vec instead of failling / reporting an error due to unexpected search result - will be improved.

```
if let Some(col) = v.as_any().downcast_ref::<arrow::array::Float64Array>() {
      col.iter().collect::<Vec<Option<f64>>>()
  } else {
      vec![]
  }
```

PR #3514, opened 2024-11-16
Update Datafusion Table Provider patch to fix MySQL refresh append mode

Sevenannn:
## 🗣 Description

Update Datafusion Table Provider patch to include the following changes:
- https://github.com/datafusion-contrib/datafusion-table-providers/pull/174
- https://github.com/datafusion-contrib/datafusion-table-providers/pull/173 

By implementing this change, all MySQL Timestamp types will have UTC timestamp value within Spice, keeping a consistency with Datafusion & Spice operating in UTC timezone by default.

 ## 🔨 Related Issues

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3524,7 +3524,7 @@ dependencies = [
 [[package]]
 name = "datafusion-table-providers"
 version = "0.1.0"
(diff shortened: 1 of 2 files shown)


eadgbear:
Do we have to document any of these changes so the users know how TIMESTAMPs should be handled?

Sevenannn:
@slyons Yes this needs to be documented in docs repo. I'll create a separate PR for that

sgrebnov (approved):

Issue #3450, opened 2024-11-11
OTEL: finalize metrics names and units


sgrebnov:
- [x] https://github.com/spiceai/spiceai/pull/3516
- [x] https://github.com/spiceai/spiceai/pull/3515
- [x] https://github.com/spiceai/spiceai/pull/3518
- [x] https://github.com/spiceai/spiceai/pull/3520
- [x] https://github.com/spiceai/spiceai/pull/3528
- [x] https://github.com/spiceai/spiceai/pull/3529
- [x] https://github.com/spiceai/spiceai/pull/3530

| Before v1.0-rc.1                                           | Unit     | v1.0-rc.1                           |
|------------------------------------------------------------|----------|-------------------------------------|
| catalogs_load_error                                        | counter  | catalog_load_errors                 |
| catalogs_status                                            | -        | catalog_load_state                      |
| datasets_acceleration_append_duration_ms                   | ms       | dataset_acceleration_refresh_duration_ms {mode: append }   |
| datasets_acceleration_last_refresh_time                    | s->ms        | dataset_acceleration_last_refresh_time_ms |
| datasets_acceleration_load_duration_ms                     | ms       | dataset_acceleration_refresh_duration_ms {mode: full}    |
| datasets_acceleration_refresh_error                        | counter  | dataset_acceleration_refresh_errors |
| datasets_count                                             | -        | dataset_active_count (`_loaded_count`?)      |
| datasets_load_error                                        | counter  | dataset_load_errors                 |
| datasets_status                                            | -        | dataset_load_state (alternative `_ready_state`?)                      |
| datasets_unavailable_time                                  | s->ms        | dataset_unavailable_time_ms     |
| embeddings_count                                           | -        | embeddings_active_count       |
| embeddings_load_error                                      | counter  | embeddings_load_errors               |
| embeddings_status                                          | -        | embeddings_load_state                    |
| flight_do_action_duration_ms                               | ms       | flight_request_duration_ms                  |
| flight_do_action_requests                                  | counter  | flight_requests                     |
| flight_do_exchange_data_updates_sent                       | counter  | flight_requests                     |
| flight_do_exchange_requests                                | counter  | flight_requests                     |
| flight_do_get_get_primary_keys_duration_ms                 | ms       | flight_request_duration_ms                  |
| flight_do_get_get_catalogs_duration_ms                     | ms       | flight_request_duration_ms                  |
[...]

PR #3516, opened 2024-11-17
Add Metrics naming principles and guidelines

sgrebnov:
## 🗣 Description

Add Metrics naming principles and guidelines

## 🔨 Related Issues

Part of https://github.com/spiceai/spiceai/issues/3450

### DIFF
--- a/docs/dev/metrics.md
+++ b/docs/dev/metrics.md
@@ -11,7 +11,7 @@
   - Metric domain or namespace:  refers to a logical grouping or category that organizes related metrics. It typically indicates the component, system, or protocol that the metric belongs to and helps avoid naming conflicts and provides context for interpreting the metric.
 - Label (or Dimension): Key-value pairs that provide additional context to a metric.
 - Counter-Based Metrics: Metrics that represent a cumulative count of events or occurrences.
-- UpDownCounter-Based Metrics: Counter-Based Metrics that can increase or decrease over time, unlike regular counters which only increase. They are used for values that fluctuate, such as current active connections or in-progress operations.
+- Gauge-Based Metrics: Counter-Based Metrics that can increase or decrease over time, unlike regular counters which only increase. They are used for values that fluctuate, such as current active connections or in-progress operations.
 - Unit-Based Metrics: Metrics that represent a specific unit of measurement, like duration or data size.
 
 ## Metric Naming Template
--- a/docs/dev/metrics.md
+++ b/docs/dev/metrics.md
@@ -80,7 +80,7 @@ The following guidelines apply the above principles, focusing on Developer Exper
 8. **UpDownCounter Naming**: Use `_count` suffix and pluralization only when recording discrete instances of countable items, e.g., `db_client_active_connections_count`.
    - **Why**: Clearly distinguishing counter-based metrics improves interpretability, which enhances the Developer Experience by making data analysis more intuitive.
 
-9. **Include Units in Name for Unit-Based Metrics**: Specify the unit (e.g., `_ms` for milliseconds, `_bytes` for data size) within metric names where applicable. Units should adhere to UCUM standards for consistent scientific notation.
+9. **Include Units in Name for Unit-Based Metrics**: Specify the unit (e.g., `_ms` for milliseconds, `_bytes` for data size) within metric names where applicable. Units should adhere to UCUM standards for consistent scientific notation where applicable. The `_bytes` suffix is chosen in favor of the UCUM-recommended `By` as it is more intuitive and readable, prioritizing the Developer Experience (DX) principle.
    - **Why**: Including UCUM-adherent units in names prevents confusion over metric scale, especially when integrating with external tools, and reinforces Clarity and Developer Experience by providing universally understood units.
 
 10. **Use Default Units for Common Metrics**:
--- a/docs/dev/metrics.md
+++ b/docs/dev/metrics.md
@@ -1,0 +1,108 @@
+# Metrics Naming
+
+## TL;DR
+
+**Metric Naming Guide**: Prioritize Developer Experience (DX) with intuitive, readable names that follow consistent conventions. Start with a domain prefix, use snake_case, avoid plurals in names (except counters), include units where relevant, and use labels for variations. Align with Prometheus and OpenTelemetry standards, and adhere to UCUM for units.
+
+## Definitions
+
+- Metric: is a measurement used to track the state and behavior of a system component. Metrics represent the current status (e.g., readiness or availability) or describe specific operations (e.g., request counts, cache hits, latencies, failures) to provide insight into system health, performance, and workload.
+- Metric Name: The identifier for the metric, typically composed of a domain or namespace prefix followed by the metric's specific name.
+  - Metric domain or namespace:  refers to a logical grouping or category that organizes related metrics. It typically indicates the component, system, or protocol that the metric belongs to and helps avoid naming conflicts and provides context for interpreting the metric.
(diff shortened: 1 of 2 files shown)


docs/dev/metrics.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Definitions
+
+- Metric: is a measurement used to track the state and behavior of a system component. Metrics represent the current status (e.g., readiness or availability) or describe specific operations (e.g., request counts, cache hits, latencies, failures) to provide insight into system health, performance, and workload.
+- Metric Name: The identifier for the metric, typically composed of a domain or namespace prefix followed by the metric's specific name.
+  - Metric domain or namespace:  refers to a logical grouping or category that organizes related metrics. It typically indicates the component, system, or protocol that the metric belongs to and helps avoid naming conflicts and provides context for interpreting the metric.
+- Label (or Dimension): Key-value pairs that provide additional context to a metric.
+- Counter-Based Metrics: Metrics that represent a cumulative count of events or occurrences.
+- UpDownCounter-Based Metrics: Counter-Based Metrics that can increase or decrease over time, unlike regular counters which only increase. They are used for values that fluctuate, such as current active connections or in-progress operations.
```
phillipleblanc:
```suggestion
- Gauge-Based Metrics: Counter-Based Metrics that can increase or decrease over time, unlike regular counters which only increase. They are used for values that fluctuate, such as current active connections or in-progress operations.
```

Gauge is the OTel terminology for this, which I think more accurately describes it: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#gauge

docs/dev/metrics.md
```diff
@@ -0,0 +1,108 @@
...
+   - **Why**: Redundancy adds unnecessary complexity and reduces readability. Shorter, more focused names enhance understanding and ease of use (Developer Experience principle).
+
+7. **Counter Naming Conventions**: Avoid `_total` suffix in counter names e.g., `flight_requests` instead of `flight_requests_total`.   - **Why**: Standardized counter naming aligns with Prometheus conventions, reducing confusion and supporting Developer Experience by avoiding unnecessary naming variation.
+
+8. **UpDownCounter Naming**: Use `_count` suffix and pluralization only when recording discrete instances of countable items, e.g., `db_client_active_connections_count`.
+   - **Why**: Clearly distinguishing counter-based metrics improves interpretability, which enhances the Developer Experience by making data analysis more intuitive.
+
+9. **Include Units in Name for Unit-Based Metrics**: Specify the unit (e.g., `_ms` for milliseconds, `_bytes` for data size) within metric names where applicable. Units should adhere to UCUM standards for consistent scientific notation.
```
phillipleblanc: Might be worth adding a note that this goes against the OTel recommendation, but we are choosing to go against that for prometheus metric compatibility.

phillipleblanc: Also UCUM states that `_by` should be the suffix for bytes, not `_bytes` - is that an explicit decision?

docs/dev/metrics.md
```diff
@@ -0,0 +1,108 @@
...
+- [UCUM Standard for Units](https://ucum.org/trac)
+
+### Notable Differences
+
+- **From OpenTelemetry**:
+  - Uses underscores (`_`) for domain separation instead of dots.
+  - Explicitly includes unit suffixes (e.g., `query_execution_duration_ms` instead of `query.operation.duration`).
+- **From Prometheus**:
+  - Does not use `_total` suffix for counters.
+  - Default units favor milliseconds (`ms`) over Prometheus’s preferred base unit of seconds.
```
phillipleblanc, line 108: Add we use `_bytes` instead of `_by`

phillipleblanc (dismissed):

docs/dev/metrics.md
sgrebnov: Updated, thank you!

docs/dev/metrics.md
sgrebnov: `_bytes` suffix proposed based on `Prioritize Developer Experience (DX)` principle as `_by` is not very intuitive to use in metric name.

Note: `with_unit("By")` is proposed to be used in the same time. While not very consistent, it should provide a good developer experience (DX) when using the metric and ensure the metric unit adheres to the UCUM standard.

sgrebnov: Added the following note:

>The `_bytes` suffix is chosen in favor of the UCUM-recommended `By` as it is more intuitive and readable, prioritizing the Developer Experience (DX) principle.

docs/dev/metrics.md
sgrebnov: Just to confirm: this is to add for OTEL difference as Prometheus is using `bytes`? Going to add this:
```
- Uses the `_bytes` unit suffix instead of the UCUM-recommended `By` unit
```

docs/dev/metrics.md
```diff
@@ -0,0 +1,108 @@
...
+## Metric Naming Template
+
+- **Format**: `<domain>_<metric>_<unit>` (for most metrics) or `<domain>_<metric_plural>` (for counters)
+  - **Example**: `flight_request_duration_ms` or `results_cache_hits`
+  - **Components**:
+    - **Domain**: Functional area, logical group or protocol, e.g., `flight`, `results_cache`
+    - **Metric**: Specific metric name, e.g., `request_duration`, `hits`
+    - **Unit**: Unit of measurement, e.g., `ms` for milliseconds, using UCUM standard units where applicable. 
+
+- Note: <metric_plural> includes the `_count` suffix when used for UpDownCounters to record discrete instances of countable items, e.g. `db_client_active_connections_count`.
```
phillipleblanc:
```suggestion
- Note: <metric_plural> includes the `_count` suffix when used for Gauges to record discrete instances of countable items, e.g. `db_client_active_connections_count`.
```

docs/dev/metrics.md
```diff
@@ -0,0 +1,108 @@
...
+   - **Why**: Domain prefixes provide immediate context, allowing developers to quickly understand a metric’s scope and relevance, supporting both Clarity and Developer Experience.
+
+6. **Avoid Redundancy in Naming**: Avoid repeating terms that are implied by the context or domain, making metric names shorter and more readable. Example: `results_cache_hits` instead of `results_cache_cache_hits`, `query_duration_ms` instead of `query_query_duration_ms`.
+   - **Why**: Redundancy adds unnecessary complexity and reduces readability. Shorter, more focused names enhance understanding and ease of use (Developer Experience principle).
+
+7. **Counter Naming Conventions**: Avoid `_total` suffix in counter names e.g., `flight_requests` instead of `flight_requests_total`.   - **Why**: Standardized counter naming aligns with Prometheus conventions, reducing confusion and supporting Developer Experience by avoiding unnecessary naming variation.
+
+8. **UpDownCounter Naming**: Use `_count` suffix and pluralization only when recording discrete instances of countable items, e.g., `db_client_active_connections_count`.
```
phillipleblanc:
```suggestion
8. **Gauge Naming**: Use `_count` suffix and pluralization only when recording discrete instances of countable items, e.g., `db_client_active_connections_count`.
```

docs/dev/metrics.md
phillipleblanc: I think its fine to leave out with the above note

docs/dev/metrics.md
sgrebnov: Thank you

docs/dev/metrics.md
sgrebnov: Thank you

phillipleblanc (approved):

PR #3519, opened 2024-11-17
Tokenizer for OpenAI embedding models for accurate chunking

Jeadie:
## 🗣 Description
Because embedding models use different tokenizers, when a user wants to use a parent `spiced` as an OpenAI compatible embedding model (or any, non-spice OpenAI compatible model), and use it locally to chunk a dataset column locally, the child `spiced` needs the right tokenizer to size the chunks accurately. 

This PR introduces a way to do this for `from: openai:...` embedding models:
  1. Provide the tokenizer as a URL/file. 
  2. Reference a `tokenizer.json` from a HF repo. 

```yaml
embeddings:
  - from: openai:hf_minilm
    name: separate_tokenizer
    files:
        # URL from HF website
      - path: https://huggingface.co/meta-llama/Llama-3.2-1B-Instruct/blob/main/tokenizer.json
        # HF Fsspec format
      - path: hf://models/sentence-transformers/all-MiniLM-L6-v2/tokenizer.json
        # Local Filesystem, same as before this PR
      - path: models/embed/BAAI/bge-small-en-v1.5/tokenizer.json
        # Providing secrets to files 
      - path: hf://meta-llama/Llama-3.2-1B-Instruct/tokenizer.json
        params:
          hf_token: ${ secrets:HF_TOKEN }

 - [ ] Update embedding reference for `openai` type embedding component.

### CODE REVIEW

File: crates/llms/src/embeddings/candle/util.rs [hunk 1]
```diff
@@ -107,19 +105,58 @@ pub(crate) fn download_hf_artifacts(
...
+    file: &str,
+    hf_token: Option<&str>,
+) -> Result<PathBuf, Box<dyn std::error::Error + Send + Sync>> {
+    let api = ApiBuilder::new()
+        .with_progress(false)
+        .with_token(hf_token.map(ToString::to_string))
+        .build()
+        .boxed()?;
+
+    let repo_type = match repo_type_opt {
```
sgrebnov, line 124: Better to warn/error if invalid repo_type specified instead of always defaulting to Model  - fro example typo `dattasets`

crates/llms/src/embeddings/candle/util.rs
```diff
@@ -107,19 +105,58 @@ pub(crate) fn download_hf_artifacts(
...
+    let repo_url = api_repo.url("");
+
+    tracing::trace!("Downloading 'config.json' for {}", repo_url);
     api_repo
         .get("config.json")
         .boxed()
         .context(FailedToInstantiateEmbeddingModelSnafu)?;
 
-    tracing::trace!("Downloading 'tokenizer.json' for {}", repo.url());
+    tracing::trace!("Downloading 'tokenizer.json' for {}", repo_url);
```
sgrebnov, line 153:
```suggestion
    tracing::trace!("Downloading 'tokenizer.json' for {repo_url}");
```

sgrebnov (approved):
👍 also run models and LLMs tests on this branch (green)

crates/runtime/src/model/embed.rs
```diff
@@ -13,20 +13,164 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+        _ => {
+            // Need to add `file://` for file paths
+            let final_url = match PathBuf::from_str(url).map(|p| p.canonicalize()) {
+                Ok(Ok(ref p)) if p.exists() => {
+                    format!("file://{}", p.to_string_lossy())
+                }
+                _ => url.to_string(),
+            };
+            let url = Url::parse(final_url.as_str()).boxed()?;
+            let (store, path) = object_store::parse_url(&url).boxed()?;
```
Sevenannn, line 126: We could use a custom error here to indicate invalid file path and provide user an actionable feedback

Sevenannn (approved):

PR #3520, opened 2024-11-17
Update Results Cache metrics according to metrics naming guidelines

sgrebnov:
## 🗣 Description

Update Results Cache according to [Metrics naming principles and guidelines](https://github.com/spiceai/spiceai/pull/3516)

Dashboards will be updated/tested separately

**This is a Breaking Change**, changes are tracked as part of https://github.com/spiceai/spiceai/issues/3450

## 🔨 Related Issues

Part of https://github.com/spiceai/spiceai/issues/3450

## Concenrs
`results_cache_items` might be updated from counter to u64_gauge and `results_cache_items_count` name to better represent target intent of tracking number of items currently stored in cache instead of aggregating number of items in cache

### DIFF
--- a/crates/cache/src/metrics.rs
+++ b/crates/cache/src/metrics.rs
@@ -53,9 +53,9 @@ pub(crate) static HITS: LazyLock<Counter<u64>> = LazyLock::new(|| {
         .init()
 });
 
-pub(crate) static ITEMS: LazyLock<Counter<u64>> = LazyLock::new(|| {
+pub(crate) static ITEMS: LazyLock<Gauge<u64>> = LazyLock::new(|| {
     METER
-        .u64_counter("results_cache_items")
-        .with_description("Number of items in the cache.")
+        .u64_gauge("results_cache_items_count")
+        .with_description("Number of items currently in the cache.")
         .init()
 });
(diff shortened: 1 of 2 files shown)


crates/cache/src/metrics.rs
```diff
@@ -23,39 +23,39 @@ use opentelemetry::{
...
         .init()
 });
 
-pub(crate) static ITEM_COUNT: LazyLock<Counter<u64>> = LazyLock::new(|| {
+pub(crate) static ITEMS: LazyLock<Counter<u64>> = LazyLock::new(|| {
     METER
-        .u64_counter("results_cache_item_count")
-        .with_description("Number of items currently in the cache.")
+        .u64_counter("results_cache_items")
+        .with_description("Number of items in the cache.")
```
phillipleblanc: If this is a counter, then this description should really be "Number of items ever in the cache"?

phillipleblanc: But given the previously comment, it seems like it should be a gauge of the number of items currently in the cache, not a counter.

sgrebnov: Thank you - will change to gauge as it makes more sense

sgrebnov: Updated

phillipleblanc (approved):

Issue #3544, opened 2024-11-19
A Postgres dataset that doesn't initialize will prevent other datasets from loading.


phillipleblanc:
## Describe the bug

Adding several datasets to a Spicepod and then having one Postgres table that takes a while to initialize will prevent the other datasets from starting up. The Spice runtime appears to be hanging. Once the dataset is registered, it will then immediately register the other datasets.

## To Reproduce

_Steps to reproduce the behavior:_

2. Click on '...'
3. Scroll down to '...'

## Expected behavior

_A clear and concise description of what you expected to happen._

### Spicepod

_Add the relevant `spicepod.yml` section._

### Output of `describe table`

_Add any relevant `describe table` output here._

### Output of `explain query`

_Add any relevant `explain query` output here._

E.g. `explain select 1`

```bash
+---------------+--------------------------------------+
| plan_type     | plan                                 |
+---------------+--------------------------------------+
| logical_plan  | Projection: Int64(1)                 |
|               |   EmptyRelation                      |
| physical_plan | ProjectionExec: expr=[1 as Int64(1)] |
|               |   PlaceholderRowExec                 |
|               |                                      |
+---------------+--------------------------------------+
```

### spice, spiced, OS info

- spice version: `spice version`
- spiced version: `spiced --version`
- OS info: `uname -a`

**Have you tried this on the latest `trunk` branch?**

- [ ] Yes
- [ ] No

**If possible, run `spiced` with DEBUG log level**
By setting the environment variable `SPICED_LOG="task_history=DEBUG,spiced=DEBUG,runtime=DEBUG,secrets=DEBUG,data_components=DEBUG,cache=DEBUG,extensions=DEBUG,spice_cloud=DEBUG"`

PR #3566, opened 2024-11-19
Load components in parallel, not concurrently

phillipleblanc:
## 🗣 Description

Rewrite `load_components` to load each component in parallel, not concurrently on the same thread. There is a similar problem when loading datasets, which is the root cause of https://github.com/spiceai/spiceai/issues/3544. I will tackle the dataset loading issue in a separate PR.

Using join_all or `buffered_unordered` will drive all of the linked futures on the same thread. If one is misbehaved and doesn't yield control, it can starve the other tasks from making progress.

## 🔨 Related Issues

Part of #3544

N/A

## 🤔 Concerns

Slightly less readable than the previous approach, but this way is also efficient in that we aren't allocating a Vec

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -443,22 +443,53 @@ impl Runtime {
     pub async fn load_components(&self) {
         self.start_extensions().await;
 
-        self.load_embeddings().await; // Must be loaded before datasets
+        // Must be loaded before datasets
+        self.load_embeddings().await;
 
-        let mut futures: Vec<Pin<Box<dyn Future<Output = ()>>>> = vec![
-            Box::pin(async {
-                if let Err(err) = self.init_task_history().await {
+        // Spawn each component load in its own task to run in parallel
+        let task_history = tokio::spawn({
+            let self_clone = self.clone();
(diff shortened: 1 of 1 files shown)


ewgenius (dismissed):

crates/runtime/src/lib.rs
```diff
@@ -443,22 +443,53 @@ impl Runtime {
...
+            async move {
+                self_clone.load_models().await;
+            }
+        });
+
+        // Wait for all tasks to complete
+        let load_result = tokio::try_join!(task_history, results_cache, datasets, catalogs, models);
+
+        if let Err(err) = load_result {
+            tracing::error!("Could not start the Spice runtime: {err}");
```
eadgbear, line 488: Should we report which subsystem failed to start?

phillipleblanc: There is already logging at the inner layer for that. This is the error if the JoinHandle returns an error, which would be an internal Tokio issue - which is highly unlikely in practice.

Jeadie (dismissed):

lukekim (approved):

PR #3572, opened 2024-11-19
Run AI integration tests as part of CI

sgrebnov:
## 🗣 Description

Add workflow to run OpenAI integration tests as part of CI.

Requires `SPICE_SECRET_OPENAI_API_KEY` secret to be added to the repository (done).

## 🔨 Related Issues

## 🤔 Concerns

Workflow does not include running Huggingface tests Github Runners don't support GPU (Cuda, Metal) hardware acceleration so tests run extremely slow and fail. This will be improved next.

### DIFF
--- a/.github/workflows/integration_models.yml
+++ b/.github/workflows/integration_models.yml
@@ -1,0 +1,93 @@
+---
+name: integration tests (models)
+
+on:
+  pull_request:
+    branches:
+      - trunk
+      - release-*
+      - feature-*
+    paths-ignore:
+      - 'docs/**'
+      - 'README.md'
+      - 'version.txt'
+      - 'acknowledgements.md'
+
+  workflow_dispatch:
+
+concurrency:
+  # Allow only one workflow per any non-trunk branch.
+  group: ${{ github.workflow }}-${{ github.ref_name }}-${{ github.ref_name == 'trunk' && github.sha || 'any-sha' }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build Test Binary
+    runs-on: spiceai-runners
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Rust
+        uses: ./.github/actions/setup-rust
+        with:
+          os: 'linux'
+
+      - name: Set up make
+        uses: ./.github/actions/setup-make
+
+      - name: Set up cc
+        uses: ./.github/actions/setup-cc
+
+      # Build the test binary without running tests
+      - name: Build AI integration test binary
+        run: |
+          TEST_BINARY_PATH=$(cargo test -p runtime --test integration_models --features models --no-run --message-format=json | jq -r 'select(.reason == "compiler-artifact" and (.target.kind | contains(["test"])) and .executable != null) | .executable')
+          cp $TEST_BINARY_PATH ./ai_integration_test
+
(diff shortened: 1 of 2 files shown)


.github/workflows/integration_models.yml
```diff
@@ -0,0 +1,93 @@
...
+
+      - name: Set up Open.ai API Key
+        run: |
+          echo 'SPICE_OPENAI_API_KEY="${{ secrets.SPICE_SECRET_OPENAI_API_KEY }}"' > .env
+
+      - name: Run integration test
+        env:
+          SPICE_SECRET_OPENAI_API_KEY: ${{ secrets.SPICE_SECRET_OPENAI_API_KEY }}
+        run: |
+          if [ -n "$SPICE_SECRET_OPENAI_API_KEY" ]; then
```
sgrebnov, line 88: Testing for `SPICE_SECRET_OPENAI_API_KEY` here instead of `Set up Open.ai API Key` so this can be changed later to run only local tests if `SPICE_SECRET_OPENAI_API_KEY` is not defined.

.github/workflows/integration_models.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Set up make
+        uses: ./.github/actions/setup-make
+
+      - name: Set up cc
+        uses: ./.github/actions/setup-cc
+
+      # Build the test binary without running tests
+      - name: Build AI integration test binary
+        run: |
+          TEST_BINARY_PATH=$(cargo test -p runtime --test integration_models --features models --no-run --message-format=json | jq -r 'select(.reason == "compiler-artifact" and (.target.kind | contains(["test"])) and .executable != null) | .executable')
```
Jeadie, line 44: We should make this a Makefile target/command.

.github/workflows/integration_models.yml
```diff
@@ -0,0 +1,93 @@
...
+          cp $TEST_BINARY_PATH ./ai_integration_test
+
+      # Upload the test binary as an artifact
+      - name: Upload test binary
+        uses: actions/upload-artifact@v4
+        with:
+          name: ai-integration-test-binary
```
Jeadie, line 51: Doesn't doing it this way imply that we expect to run the integration tests on the same binary? I feel like if a binary with the same SHA256 exists, we shouldn't rerun it?

.github/workflows/integration_models.yml
Jeadie, line 1: I am slightly worried this will churn OpenAI credits.

sgrebnov: Updated to run on push and manually only

.github/workflows/integration_models.yml
sgrebnov: We have the following Make command, the arguments are slightly different (--nocapture, --no-run --message-format=json). I think this way it is more visually easy to understand the command instead of wrapping it in make file - I'll review how to re-use the same/single make command and handle the args difference

```make
.PHONY: test-integration-models
test-integration-models:
	@cargo test -p runtime --test integration_models --features models -- --nocapture

.PHONY: test-integration-models-without-openai
test-integration-models-without-openai:
	@cargo test -p runtime --test integration_models --features models -- --nocapture --skip openai_test
```

Sevenannn (approved):

PR #3576, opened 2024-11-20
Make LLM integration tests more extensible.

Jeadie:
## 🗣 Description
 - Make adding more tests to a particular test setup more extensible by
   1. Using a `TestCase` struct to define each test
   2. Snapshot testing to ensure correct output (in this case `v1/nsql`) and to verify properties of the expected tasks.

Note: Currently only changed OpenAI testing [tests/models/openai.rs](https://github.com/spiceai/spiceai/compare/trunk...jeadie/24-11-19/snapshot?expand=1#diff-03045e056979f2a71965a2dd7fd5de17313cd746ca68e51400dda723fdd0461b).

A follow up PR will continue making the integration tests more dependent on snapshots.


crates/runtime/tests/models/snapshots/integration_models__models__openai__basic_tasks.snap
```diff
@@ -0,0 +1,18 @@
...
+[
+  {
+    "input": "how many records (as 'total_records') are in taxi_trips dataset?",
+    "task": "nsql"
+  },
+  {
+    "input": "{\"tables\":[\"spice.public.taxi_trips\"],\"output\":\"full\"}",
+    "task": "tool_use::table_schema"
+  },
+  {
```
sgrebnov: Can there be a case when model is trying few queries before getting it to work so this snapshot can vary?

Jeadie: Yes, but if this arises in future cases, we can augment the test case to decide what SQL(s) to verify properties of the task_history table.

crates/runtime/tests/models/mod.rs
```diff
@@ -332,6 +336,45 @@ async fn send_chat_completions_request(
     Ok(response)
 }
 
+/// Generic function to send a POST request, returning the response as a String.
+async fn http_post(
```
sgrebnov, line 337: Propose to move to mod.rs so it can be shared between different model providers / tests

sgrebnov (dismissed):

crates/runtime/tests/models/openai.rs
```diff
@@ -450,33 +523,43 @@ async fn verify_sql_query_chat_completion(
...
+        "chat_2_sql_tasks",
+        sql_to_display(
+            &rt,
+            format!(
+                r#"SELECT task, count(1)
+                FROM runtime.task_history
+                WHERE start_time >= '{}'
+                AND task in ('tool_use::list_datasets', 'tool_use::sql', 'tool_use::sql_query')
+                GROUP BY task;
+                ORDER BY task
```
sgrebnov: ### 
```suggestion
                GROUP BY task
                ORDER BY task;
```

crates/runtime/tests/models/openai.rs
```diff
@@ -450,33 +523,43 @@ async fn verify_sql_query_chat_completion(
...
-    assert!(
-        tasks.iter().any(|t| { t.0 == "tool_use::sql" }),
-        "Expected 'tool_use::sql' task to be executed"
-    );
-    assert!(
-        tasks.iter().any(|t| { t.0 == "sql_query" }),
-        "Expected 'sql_query' task to be executed"
+    // Verify Task History
+    insta::assert_snapshot!(
+        "chat_2_sql_tasks",
```
sgrebnov:
```suggestion
        "chat_1_sql_tasks",
```

sgrebnov (approved):

PR #3580, opened 2024-11-20
fix: Infer if the specified file format is correct in object store

### LINKED ISSUE #3556: File connector returns No Results instead of error if invalid format is specified or unsupported file is specified
## Describe the bug

File connector returns No Results instead of error if invalid format is specified or unsupported file is specified

- from: file:.env
    name: customer
    params:
      file_format: parquet

### LINKED ISSUE #3557: S3 returns `No results.` when specifying an invalid `file_format`
## Describe the bug

When using the S3 connector, specifying a `file_format` that is invalid for the files being queried returns `No results.` instead of returning an error.

## To Reproduce

1. Configure a dataset using a `file_format` parameter that is different to the actual stored files
2. Observe behavior

## Expected behavior

The Runtime should return an error when encountering a file in the specified path that differs from the specified `file_format`.

### Spicepod

```yaml
version: v1beta1
kind: Spicepod
name: spicepod
datasets:
    name: taxi_trips
    params:
      file_format: csv
```

- [x] Yes
- [ ] No

peasee:
## 🗣 Description

* Updates the `read_provider` for the `ListingTableConnector` to infer if the specified `file_format` is correct

This is because if a user specifies the wrong `file_format`, like `csv` when the path contains `parquet` files, the Runtime will simply return `No results.` instead of erroring. Now, the Runtime will return a more informative error detailing any identified extensions and a link to the documentation for available file formats.

When files with extensions are present:
```bash
2024-11-20T03:05:12.650761Z  WARN runtime::init::dataset: Invalid configuration for s3. Failed to find any files matching the extension '.csv'.
Is your `file_format` parameter correct? Spice found the following file extensions: '.parquet'.
For further information, visit: https://docs.spiceai.org/components/data-connectors#object-store-file-formats
```

When files with no extensions are present (or folders, like how S3 represents):
```bash
2024-11-20T03:03:51.670986Z  WARN runtime::init::dataset: Invalid configuration for s3. Failed to find any files matching the extension '.parquet'.
Spice could not find any files with extensions at the specified path. Check the path and try again.
```

Additionally, because we do not currently support schema evolution when new files appear at a location, also checks if there are any files to begin with. If there are no files, the `ListingTableConnector` would normally create a table with an empty schema which provides results with no columns if files are then created. This PR changes this behavior to instead outright error when there are no files:

```bash
2024-11-20T02:57:57.422333Z  WARN runtime::init::dataset: Invalid configuration for file. Failed to find any files at the specified path. Please check the path and try again.
```

## 🔨 Related Issues

* Closes #3557
* Closes #3556

### DIFF
--- a/crates/runtime/src/dataconnector/listing/connector.rs
+++ b/crates/runtime/src/dataconnector/listing/connector.rs
@@ -17,6 +17,7 @@ limitations under the License.
 use crate::accelerated_table::AcceleratedTable;
 use crate::component::dataset::Dataset;
 use crate::dataconnector::DataConnector;
+use crate::dataconnector::DataConnectorError;
 use crate::dataconnector::DataConnectorResult;
 use crate::parameters::Parameters;
 use async_trait::async_trait;
@@ -34,9 +35,11 @@ use datafusion::datasource::listing::{
 use datafusion::datasource::TableProvider;
 use datafusion::execution::config::SessionConfig;
 use datafusion::execution::context::SessionContext;
+use futures::TryStreamExt;
 use object_store::ObjectStore;
 use snafu::prelude::*;
 use std::any::Any;
+use std::collections::HashSet;
 use std::fmt::Display;
 use std::str::FromStr;
 use std::sync::Arc;
@@ -319,6 +322,16 @@ impl<T: ListingTableConnector + Display> DataConnector for T {
                 })?)
             }
             Some(file_format) => {
+                let object_store = self.get_object_store(dataset)?;
+                check_for_files_and_extensions(
+                    format!("{self}"),
+                    &extension,
+                    table_path.clone(),
+                    &ctx,
+                    &object_store,
+                )
+                .await?;
+
                 let mut options = ListingOptions::new(file_format).with_file_extension(&extension);
 
                 let resolved_schema = options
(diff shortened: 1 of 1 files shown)


crates/runtime/src/dataconnector/listing/connector.rs
```diff
@@ -387,6 +400,72 @@ impl<T: ListingTableConnector + Display> DataConnector for T {
...
+
+        let display_extensions = extensions
+            .iter()
+            .map(|e| format!("'{e}'"))
+            .collect::<Vec<_>>()
+            .join(", ");
+
+        return Err(DataConnectorError::InvalidConfigurationNoSource {
+            dataconnector: dataconnector.clone(),
+            message: format!("Failed to find any files matching the extension '{extension}'.\nIs your `file_format` parameter correct? Spice found the following file extensions: {display_extensions}.\nFor further information, visit: https://docs.spiceai.org/components/data-connectors#object-store-file-formats")
```
Jeadie, line 462: nit: Could be worth being its own error type.

Jeadie (approved):

crates/runtime/src/dataconnector/listing/connector.rs
peasee: Is there a non-object-store connector that could use this? This error applies to every object store connector, which includes `file`, `s3`, `abfs` and `ftp/sftp`. I'm not sure there's another connector that would use this.

PR #3582, opened 2024-11-20
fix: Handle invalid S3 auth and region better

### LINKED ISSUE #3553: S3 returns a generic timeout error when invalid region or auth mode is specified
## Describe the bug

When specifying an invalid region, or using `s3_auth: iam_role` when no IAM role is loaded, returns a generic timeout error instead of a more specific error:

```
2024-11-18T23:39:25.337062Z  WARN runtime::init::dataset: Cannot connect to s3. Object Store error: Generic S3 error: Error performing list request: Error after 10 retries in 3.836037419s, max_retries:10, retry_timeout:180s, source:error sending request for url (https://s3.not-a-real-region.amazonaws.com/spiceai-demo-datasets?list-type=2&prefix=taxi_trips%2F2024%2F)
```

## To Reproduce

1. Configure an S3 dataset like the supplied spicepod
2. See error

## Expected behavior

The Runtime should return a more specific and actionable error. For example, for the invalid region it should return an error about invalid region. For `s3_auth: iam_role` without a loaded IAM role it should return a permissions error.

### Spicepod

```yaml
version: v1beta1
kind: Spicepod
name: spicepod
datasets:
    name: taxi_trips
    params:
      file_format: parquet
      s3_auth: iam_role
      s3_region: not-a-region
```

- [x] Yes
- [ ] No

peasee:
## 🗣 Description

* Updates `s3_region` validation to check against a hard-coded list of regions, avoiding generic time out errors from trying to access an invalid region.

Regions list is hard coded here, because they're unlikely to change (with the exception of brand new regions) and there isn't a great alternative in the Rust SDK to retrieving the regions.

When failing due to an incorrect region, the error looks like:
```bash
2024-11-20T05:06:38.362997Z  WARN runtime::init::dataset: Error initializing dataset taxi_trips. The 's3_region' parameter must be a valid AWS region code, but 'not-a-real-region' was provided.
For further information, visit: https://docs.spiceai.org/components/data-connectors/s3#params
```

* Adds error handling interception for implementations of `ListingTableConnector`, to override the return error for `object_store` errors

S3 implements this interceptor to handle `Generic` errors when `s3_auth: iam_role`. This is because, when using `s3_auth: iam_role` in a non-IAM role environment, object store will return a generic timeout when trying to reach the AWS instance profile manager to get a token: `http://169.254.169.254/latest/api/token`.

We make an educated guess with the error that it's due to the IAM role environment configuration, with an error that looks like:
```bash
2024-11-20T05:05:56.986759Z  WARN runtime::init::dataset: Invalid configuration for s3. Failed to authenticate using an IAM role.
Are you sure you're running in an environment with an IAM role?
{object store error}
For further information, visit: https://docs.spiceai.org/components/data-connectors/s3#auth
```

With an inner error visible like:
```bash
2024-11-20T05:05:56.986759Z  WARN runtime::init::dataset: Invalid configuration for s3. Failed to authenticate using an IAM role.
Are you sure you're running in an environment with an IAM role?
Error after 10 retries in 16.509493296s, max_retries:10, retry_timeout:180s, source:error sending request for url (http://169.254.169.254/latest/api/token)
For further information, visit: https://docs.spiceai.org/components/data-connectors/s3#auth
```

## 🔨 Related Issues

* Closes #3553

### DIFF
--- a/crates/runtime/src/dataconnector/s3.rs
+++ b/crates/runtime/src/dataconnector/s3.rs
@@ -16,8 +16,8 @@ limitations under the License.
 
 use super::{
     listing::{self, ListingTableConnector},
-    DataConnector, DataConnectorFactory, DataConnectorParams, DataConnectorResult, ParameterSpec,
-    Parameters,
+    DataConnector, DataConnectorError, DataConnectorFactory, DataConnectorParams,
+    DataConnectorResult, ParameterSpec, Parameters,
 };
 
 use crate::component::dataset::Dataset;
@@ -30,6 +30,42 @@ use std::pin::Pin;
 use std::string::String;
 use std::sync::Arc;
 use url::Url;
+
+// https://docs.aws.amazon.com/general/latest/gr/rande.html
+pub const AWS_REGIONS: [&str; 32] = [
+    "us-east-1",
+    "us-east-2",
+    "us-west-1",
+    "us-west-2",
+    "af-south-1",
+    "ap-east-1",
+    "ap-south-1",
+    "ap-south-2",
+    "ap-northeast-1",
+    "ap-northeast-2",
+    "ap-northeast-3",
+    "ap-southeast-1",
+    "ap-southeast-2",
+    "ap-southeast-3",
+    "ap-southeast-4",
+    "ap-southeast-5",
+    "ca-central-1",
+    "ca-west-1",
+    "eu-central-1",
+    "eu-central-2",
+    "eu-west-1",
+    "eu-west-2",
+    "eu-west-3",
+    "eu-south-1",
+    "eu-south-2",
+    "eu-north-1",
+    "sa-east-1",
+    "il-central-1",
+    "me-south-1",
+    "me-central-1",
+    "us-gov-east-1",
+    "us-gov-west-1",
+];
 
 #[derive(Debug, Snafu)]
 pub enum Error {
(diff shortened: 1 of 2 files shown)


crates/runtime/src/dataconnector/s3.rs
```diff
@@ -131,6 +177,15 @@ impl DataConnectorFactory for S3Factory {
                 }
             }
 
+            if let Some(region) = params.parameters.get("region").expose().ok() {
+                if !AWS_REGIONS.contains(&region) {
```
phillipleblanc: I don't think we should hard-fail here. This means that if/when AWS adds a new region, customers can't use the S3 connector until we release a new version that includes the new region. This should be a warning, but still allow it through.

sgrebnov: Can region be defined upper case / need to lowercase the region while checking?

peasee: good call, region cannot be uppercase. I've added a new error type to warn when we automatically convert it to lower case:

```bash
2024-11-20T05:52:48.718016Z  WARN runtime::dataconnector::s3: The `s3_region` parameter requires a lowercase AWS region code, but 'US-EAST-1' was provided.
Spice will automatically convert the region code to lowercase.
For further information, visit: https://docs.spiceai.org/components/data-connectors/s3#params
```

phillipleblanc (approved):

PR #3590, opened 2024-11-20
`spice search` to warn if dataset is not ready and won't be included in search

sgrebnov:
## 🗣 Description

Add back logic to warn if dataset is not ready and won't be included in search removed in [spice search to default to only datasets with embeddings](https://github.com/spiceai/spiceai/pull/3588)

Also improve error message if there are no datasets available for search

## 🤔 Concerns

~Converted to DRAFT as We should warn only if dataset has embeddings defined (improve `api.GetDatasetsWithStatus` to include can_search_documents ). Plan to improve this otherwise such message for a dataset w/o embeddings can confuse.~

### DIFF
--- a/bin/spice/cmd/search.go
+++ b/bin/spice/cmd/search.go
@@ -66,7 +66,11 @@ spice search --cloud
 
 		for _, dataset := range datasets {
 			if dataset.Status != api.Ready.String() && dataset.Status != api.Refreshing.String() {
-				slog.Warn(fmt.Sprintf("Dataset %s is not ready (%s) and will not be included in search.", dataset.Name, dataset.Status))
+				// warn only if vector_search is supported by the dataset
+				prop_val, _ := dataset.GetPropertyValue("vector_search")
+				if prop_val == "supported" {
+					slog.Warn(fmt.Sprintf("Dataset %s is not ready (%s) at the moment and will be excluded from the search if it is not ready when the search is performed.", dataset.Name, dataset.Status))
+				}
 			}
 		}
 
--- a/bin/spice/cmd/search.go
+++ b/bin/spice/cmd/search.go
@@ -69,7 +69,7 @@ spice search --cloud
 				// warn only if vector_search is supported by the dataset
 				prop_val, _ := dataset.GetPropertyValue("vector_search")
 				if prop_val == "supported" {
-					slog.Warn(fmt.Sprintf("Dataset %s is not ready (%s) at the moment and will be excluded from the search if it is not ready when the search is performed.", dataset.Name, dataset.Status))
+					slog.Warn(fmt.Sprintf("Dataset %s is not ready (%s) and will be excluded from the search.", dataset.Name, dataset.Status))
 				}
 			}
 		}
--- a/bin/spice/cmd/search.go
+++ b/bin/spice/cmd/search.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/peterh/liner"
 	"github.com/spf13/cobra"
+	"github.com/spiceai/spiceai/bin/spice/pkg/api"
 	"github.com/spiceai/spiceai/bin/spice/pkg/context"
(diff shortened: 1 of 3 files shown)


bin/spice/cmd/search.go
```diff
@@ -58,6 +59,21 @@ spice search --cloud
...
+		if err != nil {
+			slog.Error("could not list datasets", "error", err)
+		}
+
+		for _, dataset := range datasets {
+			if dataset.Status != api.Ready.String() && dataset.Status != api.Refreshing.String() {
+				// warn only if vector_search is supported by the dataset
+				prop_val, _ := dataset.GetPropertyValue("vector_search")
+				if prop_val == "supported" {
+					slog.Warn(fmt.Sprintf("Dataset %s is not ready (%s) at the moment and will be excluded from the search if it is not ready when the search is performed.", dataset.Name, dataset.Status))
```
phillipleblanc:
```suggestion
					slog.Warn(fmt.Sprintf("Dataset %s is not ready (%s) and will be excluded from the search.", dataset.Name, dataset.Status))
```

bin/spice/pkg/api/dataset.go
```diff
@@ -17,9 +17,18 @@ limitations under the License.
...
+	Status              string                 `json:"status,omitempty" csv:"status,omitempty" yaml:"status,omitempty"`
+	Properties          map[string]interface{} `json:"properties,omitempty"`
+}
+
+func (ds *Dataset) GetPropertyValue(key string) (interface{}, bool) {
+	if ds.Properties == nil {
+		return nil, false
+	}
+	value, exists := ds.Properties[key]
+	return value, exists
```
phillipleblanc:
```suggestion
	value, ok := ds.Properties[key]
	return value, ok
```

phillipleblanc: Slightly more idiomatic Go

crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -51,6 +51,12 @@ pub enum Error {
     #[snafu(display("Data sources [{}] does not exist", data_source.iter().map(TableReference::to_quoted_string).join(", ")))]
     DataSourcesNotFound { data_source: Vec<TableReference> },
 
+    #[snafu(display("Failed to retrive table {} data source", data_source.to_quoted_string()))]
```
phillipleblanc:
```suggestion
    #[snafu(display("Failed to retrieve table {} data source", data_source.to_quoted_string()))]
```

I don't think this error message conforms with the new guidelines, @peasee to weigh in.

bin/spice/cmd/search.go
sgrebnov: @phillipleblanc - this check is performed only once when the tool is started. As we don't collect datasets to pass to the search API now, there could be cases where a dataset is NOT ready at the start of the search but becomes ready and is included later when we work with the app

Phillip - could you please confirm the message should be updated.

crates/runtime/src/embeddings/vector_search.rs
sgrebnov: Updated error message.

> Failed to find table '{table_name}'. An internal error occurred during vector search.\nPlease report a bug on GitHub: https://github.com/spiceai/spiceai/issues"

bin/spice/cmd/search.go
phillipleblanc: I don't think the addition words add anything to the message that the dataset is not ready? I would expect that when it becomes ready, it will be included in the search - but as of the moment the message is sent, it wouldn't be included, correct?

sgrebnov: >I would expect that when it becomes ready, it will be included in the search - but as of the moment the message is sent, it wouldn't be included, correct?

This is correct

phillipleblanc (approved):

PR #3599, opened 2024-11-21
Add E2E test for spice search and chat functionality (OpenAI)

sgrebnov:
## 🗣 Description

Add E2E tests for `spice search` and `spice chat` functionality (OpenA) using [expect](https://core.tcl-lang.org/expect/index) (simple tool for automating interactive applications)
 - `s3://spiceai-public-datasets/tpcds/catalog_page/` is used to test `Utf8View` type  records with NULL values
 - `github:github.com/spiceai/spiceai/issues` covers `Utf8` type and real content characteristics (length variety, etc.)

Example test run: https://github.com/spiceai/spiceai/actions/runs/11950463815

## 🔨 Related Issues

## 🤔 Concerns

1. OpenAI chat returns stable result for queries like `How many issues stored in database?` so test can include specific number instead of testing that chat request just completed.

### DIFF
--- a/.github/workflows/e2e_test_ci_models.yml
+++ b/.github/workflows/e2e_test_ci_models.yml
@@ -39,13 +39,19 @@ jobs:
                 target_os: "linux",
                 target_arch: "x86_64",
                 target_arch_go: "amd64"
+              }, {
+                name: "macOS aarch64 (Apple Silicon)",
+                builder: "macos-14",
+                runner: "macos-14",
+                target_os: "darwin",
+                target_arch: "aarch64",
+                target_arch_go: "arm64"
               }
             ];
 
             return context.eventName === 'pull_request' ? matrix.slice(0, 1) : matrix;
 
   build:
-    # if: false
     name: Build ${{ matrix.target.name }} binaries
     runs-on: ${{ matrix.target.builder }}
     needs: setup-matrix
@@ -187,6 +193,11 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y expect
+
+      - name: Install expect (macOS)
+        if: matrix.target.target_os == 'darwin'
+        run: |
+          brew install expect
 
       - name: Test vector search
         run: |
--- a/.github/workflows/e2e_test_ci_models.yml
+++ b/.github/workflows/e2e_test_ci_models.yml
@@ -1,0 +1,205 @@
+name: E2E Test CI (models)
+
+on:
+  push:
+    branches:
+      - trunk
+      - release-*
+    paths-ignore:
+      - 'docs/**'
+      - 'README.md'
+      - 'version.txt'
+      - 'acknowledgements.md'
+
+  workflow_dispatch:
+
+concurrency:
+  # Allow only one workflow per any non-trunk branch.
+  group: ${{ github.workflow }}-${{ github.ref_name }}-${{ github.ref_name == 'trunk' && github.sha || 'any-sha' }}
+  cancel-in-progress: true
+
+jobs:
+  setup-matrix:
+    name: Setup strategy matrix
+    runs-on: spiceai-runners
+    outputs:
+      matrix: ${{ steps.setup-matrix.outputs.result }}
+
+    steps:
+      - name: Set up matrix
+        uses: actions/github-script@v7
+        id: setup-matrix
+        with:
+          script: |
+            const matrix = [
+              {
+                name: "Linux x64",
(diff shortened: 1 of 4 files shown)


test/models/spicepod.yml
```diff
@@ -0,0 +1,76 @@
...
+    metadata:
+      instructions: Always provide reference links. Use the primary key as `issue_id` to generate reference links.
+      reference_url_template: https://github.com/spiceai/spiceai/issues/<issue_id>
+    params:
+      github_token: ${secrets:GITHUB_TOKEN}
+      github_query_mode: "search"
+    acceleration:
+      enabled: true
+      refresh_retry_max_attempts: 3
+      refresh_sql: "SELECT * FROM issues WHERE created_at='2024-11-09T22:47:45'"
```
sgrebnov, line 24: Embeddings can't currently handle NULL records, filter will be updated to include more records once it is fixed

test/models/spicepod.yml
```diff
@@ -0,0 +1,76 @@
...
+          - id
+    
+  - from: s3://spiceai-public-datasets/tpcds/catalog_page/
+    name: catalog_page
+    description: Represents an e-commerce platform's catalog page metadata
+    params:
+      file_format: parquet
+    acceleration:
+      enabled: true
+      refresh_sql: SELECT * FROM catalog_page limit 1;
```
sgrebnov, line 38: Embeddings can't currently handle NULL records, filter will be updated to include more records once it is fixed

test/models/chat_01.exp
```diff
@@ -0,0 +1,115 @@
...
+
+    expect {
+        " ERROR " {
+            # allow stream to complete
+            sleep 2
+            send_user "Error detected\n"
+            exit 1
+        }
+        "chat> " {
+            # end of model response
```
Sevenannn, line 91: Does these test only verify whether the chat is successfully initiated by the question, but doesn't verify the correctness of the result?

sgrebnov: Correct, in first iteration it is proposed to just verify that model responds to the query (the api works). If response is reliable we will then add check for specific response.

.github/workflows/e2e_test_ci_models.yml
```diff
@@ -0,0 +1,205 @@
...
+                name: "Linux x64",
+                builder: "spiceai-runners",
+                runner: "ubuntu-latest",
+                target_os: "linux",
+                target_arch: "x86_64",
+                target_arch_go: "amd64"
+              }
+            ];
+
+            return context.eventName === 'pull_request' ? matrix.slice(0, 1) : matrix;
```
phillipleblanc: If we aren't using the dynamic matrix, I'd prefer to just define the matrix inline on the build job

.github/workflows/e2e_test_ci_models.yml
```diff
@@ -0,0 +1,205 @@
...
+
+  build:
+    # if: false
```
phillipleblanc:
```suggestion
```

.github/workflows/e2e_test_ci_models.yml
```diff
@@ -0,0 +1,205 @@
...
+          path: |
+            spice
+            spiced
+
+  test_openai_model:
+    name: 'openai model (${{ matrix.target.target_os }}-${{ matrix.target.target_arch }})'
+    runs-on: ${{ matrix.target.runner }}
+    needs:
+      - build
+      - setup-matrix
```
phillipleblanc: Add timeout-minutes to each job

phillipleblanc (dismissed):

.github/workflows/e2e_test_ci_models.yml
sgrebnov: Updated, thank you!

.github/workflows/e2e_test_ci_models.yml
sgrebnov: Added 30m for build and 10m for test execution, thank you!

.github/workflows/e2e_test_ci_models.yml
sgrebnov: Thank you

.github/workflows/e2e_test_ci_models.yml
sgrebnov: Example test run: https://github.com/spiceai/spiceai/actions/runs/11950463815

Sevenannn (approved):

PR #3600, opened 2024-11-21
Handle empty embeds by allowing for nulls

Jeadie:
## 🗣 Description
 - Since switching to complete text-embedding-inference solution, inputs cannot be empty/null. 
 - This PR handles empty/null inputs in a dataset embedding column by avoiding providing these inputs to the embedding models. 
 - The output format for embedding columns are:
     - Standard embedding column: null embedding element in column
     - Chunked embedding column: empty array for both `_embedding` and `_offset`. 

#### Example (Standard)
```
>> select b, b is null, b_embedding is null, b_embedding from test;
+---------------------------+----------------+--------------------------+------------------------------+
| b                         | test.b IS NULL | test.b_embedding IS NULL |          b_embedding         |
+---------------------------+----------------+--------------------------+------------------------------+
| hello world               | false          | false                    | [0.030892028, 0.13509002,... |
|                           | true           | true                     |                              |
|                           | true           | true                     |                              |
| hello worldhello world... | false          | false                    | [0.06095122,  0.0672081,...  |
+---------------------------+----------------+--------------------------+------------------------------+
```

#### Example (Chunked)
```
>> select b, b is null, array_length(b_embedding) from test;
+------------------------------------------------------------------------------------------+----------------+--------------------------------+
| b                                                                                        | test.b IS NULL | array_length(test.b_embedding) |
+------------------------------------------------------------------------------------------+----------------+--------------------------------+
| hello world                                                                              | false          | 1                              |
|                                                                                          | true           | 0                              |
|                                                                                          | true           | 0                              |
| hello worldhello worldhello worldhello worldhello worldhello worldhello worldhello world | false          | 7                              |
+------------------------------------------------------------------------------------------+----------------+--------------------------------+
```

## 🔨 Related Issues
 - https://github.com/spiceai/spiceai/pull/3199 

 - [ ] Update requirement of inheriting embedding dataset columns.

### DIFF
--- a/crates/runtime/src/embeddings/execution_plan.rs
+++ b/crates/runtime/src/embeddings/execution_plan.rs
@@ -342,7 +342,7 @@ async fn get_vectors(
     arr: impl Iterator<Item = Option<&str>>,
     model: &dyn Embed,
 ) -> Result<FixedSizeListArray, Box<dyn std::error::Error + Send + Sync>> {
-    // Filter out nulls or empty strings before caling [`Embed::embed`].
+    // Filter out nulls or empty strings before calling [`Embed::embed`].
     let (null_pairs, values): (Vec<_>, Vec<_>) = arr
         .enumerate()
         .partition(|(_, o)| o.is_none() || o.is_some_and(str::is_empty));
--- a/crates/runtime/src/embeddings/execution_plan.rs
+++ b/crates/runtime/src/embeddings/execution_plan.rs
@@ -322,7 +322,7 @@ async fn get_vectors(
     // Filter out nulls or empty strings before caling [`Embed::embed`].
     let (null_pairs, values): (Vec<_>, Vec<_>) = arr
         .enumerate()
-        .partition(|(_, o)| o.is_none() || o.is_some_and(|v| v.is_empty()));
+        .partition(|(_, o)| o.is_none() || o.is_some_and(str::is_empty));
     let nulls: Vec<usize> = null_pairs.into_iter().map(|(i, _)| i).collect();
 
     let column: Vec<String> = values
@@ -340,7 +340,8 @@ async fn get_vectors(
         ),
         vector_length as i32,
         embedded_data.len() + nulls.len(),
-    );
+    )
+    .with_field(Arc::new(Field::new("item", DataType::Float32, false)));
 
     let mut i: usize = 0;
     for vector in embedded_data {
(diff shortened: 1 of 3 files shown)


crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -314,24 +314,46 @@ pub(crate) async fn compute_additional_embedding_columns(
...
+    let mut builder = FixedSizeListBuilder::with_capacity(
+        PrimitiveBuilder::<Float32Type>::with_capacity(
+            (embedded_data.len() + nulls.len()) * vector_length,
+        ),
+        vector_length as i32,
+        embedded_data.len() + nulls.len(),
+    );
+
+    let mut i: usize = 0;
+    for vector in embedded_data {
```
sgrebnov, line 374: Interesting approach could be constructing resultant array by providing offsets and logical nulls
https://github.com/spiceai/spiceai/blob/67879865616df4197290d9ad2d7a9be5c09d3300/crates/arrow_tools/src/record_batch.rs#L124

Jeadie: They key reason we can't do this is because we split out empty/null inputs, from the remaining and only pass non-empty/null inputs to the model (`let embedded_data = model.embed(EmbeddingInput::StringArray(column)).await?;`). We then need to recombine it, we don't have a contiguous block of data.

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -314,24 +314,77 @@ pub(crate) async fn compute_additional_embedding_columns(
...
+///                 [`FixedSizeListArray`]
+/// ```
+#[allow(clippy::cast_possible_truncation, clippy::cast_possible_wrap)]
 async fn get_vectors(
     arr: impl Iterator<Item = Option<&str>>,
     model: &dyn Embed,
 ) -> Result<FixedSizeListArray, Box<dyn std::error::Error + Send + Sync>> {
-    let field = Arc::new(Field::new("item", DataType::Float32, false));
-    let column: Vec<String> = arr.filter_map(|s| s.map(ToString::to_string)).collect();
+    // Filter out nulls or empty strings before caling [`Embed::embed`].
```
phillipleblanc:
```suggestion
    // Filter out nulls or empty strings before calling [`Embed::embed`].
```

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -314,24 +314,77 @@ pub(crate) async fn compute_additional_embedding_columns(
...
+///                   | [0.1, 0.2] | <------------------+  |
+///                   | None       |                       |
+///                   | None       |                       |
+///                   | [0.8, 0.9] | <---------------------+
+///                   +------------+
+///
+///                 [`FixedSizeListArray`]
+/// ```
+#[allow(clippy::cast_possible_truncation, clippy::cast_possible_wrap)]
 async fn get_vectors(
```
phillipleblanc, line 341: This functions needs some unit tests to cover it.

Jeadie: Good point, helped find an edge case too!

phillipleblanc (approved):

crates/runtime/src/embeddings/execution_plan.rs
sgrebnov: Confirm this is to recombine: the idea is to just specify nulls/offests instead of `append`/`append_nulls`

sgrebnov: ~I also think we should cover this by unit test (if not yet) to ensure this is not broken at some point~. I see it is already covered.

sgrebnov (approved):

PR #3609, opened 2024-11-21
Adding custom User-Agent parameters to chat, nsql and flightrepl

eadgbear:
## 🗣 Description

This change ensures that all CLI commands that might execute queries (`chat`, `nsql`, `sql`) include the ability to specify a custom `User-Agent` string. It also ensures that `chat` and `nsql` send a `User-Agent` string along with all requests.

- https://github.com/spiceai/docs/issues/649


bin/spice/cmd/chat.go
```diff
@@ -78,6 +79,13 @@ spice chat --model <model> --cloud
 			rtcontext.SetApiKey(apiKey)
 		}
 
+		userAgent, _ := cmd.Flags().GetString(userAgentKeyFlag)
```
sgrebnov, line 90: @slyons  - is my understanding correct that the target use case for this is to provide custom agent for E2E tests so we correctly handle such usage? Any other example where we would like to override default user agent?

eadgbear: That's *our* use case for it, but it's a generic enough use case since the `User-Agent` will be exposed through the metrics

bin/spice/pkg/context/context.go
```diff
@@ -322,10 +324,23 @@ func (c *RuntimeContext) SetApiKey(apiKey string) {
...
+
+func (c *RuntimeContext) SetUserAgentClient(client string) {
+	c.userAgent = util.GetSpiceUserAgent(client)
+}
+
 func (c *RuntimeContext) GetHeaders() http.Header {
 	headers := make(http.Header)
 	headers.Set("Content-Type", "application/json")
 
+
```
```suggestion
```

bin/spice/pkg/context/context.go
```diff
@@ -50,13 +50,15 @@ type RuntimeContext struct {
 	isCloud         bool
 	httpClient      *http.Client
 	apiKey          string
+	userAgent	    string
```
phillipleblanc: Match the indentation with the other params (the linter should do this automatically)

crates/runtime/tests/endpoint_auth/flight.rs
```diff
@@ -79,11 +79,17 @@ async fn test_flight_auth() -> Result<(), anyhow::Error> {
         .expect("to connect to flight endpoint");
     let client = FlightServiceClient::new(channel);
 
-    let result =
-        flightrepl::get_records(client.clone(), "SELECT 1", Some(&"valid".to_string())).await;
+    let result = flightrepl::get_records(
+        client.clone(),
+        "SELECT 1",
+        Some(&"valid".to_string()),
+        "test_flight_auth/1.0",
```
phillipleblanc: This version should use the Cargo.toml version from the `env!("CARGO_PKG_VERSION")` macro

crates/runtime/tests/endpoint_auth/flight.rs
```diff
@@ -79,11 +79,17 @@ async fn test_flight_auth() -> Result<(), anyhow::Error> {
...
+    )
+    .await;
     assert!(result.is_ok());
 
-    let Err(e) = flightrepl::get_records(client, "SELECT 1", None).await else {
+    let Err(e) = flightrepl::get_records(client, "SELECT 1", None, "test_flight_auth/1.0").await
```
phillipleblanc: This version should use the Cargo.toml version from the `env!("CARGO_PKG_VERSION")` macro

phillipleblanc (approved):

Issue #3616, opened 2024-11-22
Enhancement: The REPL identifies it is waiting for input on multi-lines


peasee:
## Goal-State/What/Result

The Spice REPL provides clear feedback when it is expecting more user input (e.g. a user has not ended their query with `;`).

## Why/Purpose

To improve the experience of the REPL by ensuring user's are aware the REPL is waiting for their input instead of providing no feedback.

## By When

_A target date for done-done completion of the entire enhancement._

**Issue/Spec written and reviewed:** 22nd Nov, 2024
**Done-Done:** 23rd Nov, 2024

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

The REPL prefixes the input line to signify it is waiting for user input, similar to MySQL:

```bash
mysql>
    ->
```

This proposal would update the Spice REPL like:
```bash
sql>
  ->
```

### Security Review

* There is no security concern with this enhancement

## How/Plan

* Remove the `ReplHelper` to manage newlines externally

This is because `rustyline` blocks the `stdin`/`stdout` while waiting for additional input, and `ValidationResult` provides no mechanism of adding additional prefixes/prompts on newlines.

* Add a new input manager loop to check for the end of a newline

Using a completely custom input manager with a loop, we can manage the state of `stdin`/`stdout` by preventing `rustyline` from operating over multiple lines. `rustyline` will still be used in each individual line.

On new lines that aren't the first line, a prompt of `->` should be used instead of `sql>`

PR #3617, opened 2024-11-22
feat: Make REPL identify it's waiting on a new line

peasee:
## 🗣 Description

* Updates the REPL to make it identify when it is waiting on more input with `->`

Example:
```bash
sql> explain
  -> select * from
  -> taxi_trips
  -> limit 5;
```

## 🔨 Related Issues

* Part of #3616

* Needs a pass on quickstarts/samples to make sure none need updating from this change.

### DIFF
--- a/crates/flightrepl/src/lib.rs
+++ b/crates/flightrepl/src/lib.rs
@@ -38,8 +38,8 @@ use llms::chat::LlmRuntime;
 use prost::Message;
 use reqwest::Client;
 use rustyline::error::ReadlineError;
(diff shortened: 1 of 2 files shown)


phillipleblanc (approved):

Sevenannn (approved):
lgtm! Just quick question on the show_tables.exp, is this designed for local verification purpose, instead of being part of the test routine?

peasee:
> lgtm! Just quick question on the show_tables.exp, is this designed for local verification purpose, instead of being part of the test routine?

yes, it's just local testing for now!

PR #3677, opened 2024-11-25
Improve github connector error

Sevenannn:
## 🗣 Description

Improve github connector errors according to error handling guidelines: https://github.com/spiceai/spiceai/blob/trunk/docs/dev/error_handling.md

### DIFF
--- a/crates/data_components/src/token_provider.rs
+++ b/crates/data_components/src/token_provider.rs
@@ -6,7 +6,7 @@ use snafu::prelude::*;
 
 #[derive(Debug, Snafu)]
 pub enum Error {
-    #[snafu(display("Unable to get token: {source}"))]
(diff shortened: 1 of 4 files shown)


crates/data_components/src/token_provider.rs
```diff
@@ -6,7 +6,7 @@ use snafu::prelude::*;
 
 #[derive(Debug, Snafu)]
 pub enum Error {
-    #[snafu(display("Unable to get token: {source}"))]
+    #[snafu(display("Failed to get token.\n{source}"))]
```
sgrebnov, line 9: ~We might add an action to verify configuration and link to connector docs~

I see it is a wrapper around existing error: UnableToGenerateJWTSnafu

sgrebnov (approved):

Issue #3517, opened 2024-11-17
AI/LLM integration tests include test for locally stored model


sgrebnov:

sgrebnov:
Implemented by https://github.com/spiceai/spiceai/pull/3690

PR #3688, opened 2024-11-26
Add local model test to LLMs integration tests

sgrebnov:
## 🗣 Description

Add integration test for local model to LLMs testing

## 🔨 Related Issues

https://github.com/spiceai/spiceai/issues/3517

## 🤔 Concerns

Local model inference is very slow and can't be run on GH Hosted runners so temporarily disabled in CI.

### DIFF
--- a/crates/llms/tests/llms/create.rs
+++ b/crates/llms/tests/llms/create.rs
@@ -14,7 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use std::{path::PathBuf, sync::Arc};
+use std::{
+    path::{Path, PathBuf},
+    sync::Arc,
+};
 
 use anyhow::Context;
 use async_openai::error::OpenAIError;
@@ -41,7 +44,7 @@ pub(crate) fn create_anthropic(model_id: Option<&str>) -> Result<Arc<dyn Chat>, 
 }
 
 pub(crate) async fn create_local() -> Result<Arc<dyn Chat>, anyhow::Error> {
-    let temp_dir = local_model_dir("Phi-3-mini-4k-instruct").await;
+    let temp_dir = local_model_dir("Phi-3-mini-4k-instruct").await?;
 
     download_hf_model_artifacts(
         "microsoft/Phi-3-mini-4k-instruct",
@@ -56,8 +59,7 @@ pub(crate) async fn create_local() -> Result<Arc<dyn Chat>, anyhow::Error> {
         ],
         &temp_dir,
     )
-    .await
-    .expect("Failed to download test model artifacts");
+    .await?;
 
     let model_weights = [
         temp_dir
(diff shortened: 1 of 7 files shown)


crates/llms/tests/llms/create.rs
```diff
@@ -36,3 +42,96 @@ pub(crate) fn create_anthropic(model_id: Option<&str>) -> Result<Arc<dyn Chat>,
 
     Ok(Arc::new(model))
 }
+
+pub(crate) async fn create_local() -> Result<Arc<dyn Chat>, anyhow::Error> {
```
Jeadie, line 46:
```suggestion
pub(crate) async fn create_local(model_id: &str) -> Result<Arc<dyn Chat>, anyhow::Error> {
```

crates/llms/tests/llms/create.rs
```diff
@@ -36,3 +42,96 @@ pub(crate) fn create_anthropic(model_id: Option<&str>) -> Result<Arc<dyn Chat>,
...
+        temp_dir.join("tokenizer.json").to_str(),
+        temp_dir.join("tokenizer_config.json").to_str(),
+        None,
+    )
+    .map_err(anyhow::Error::from)?;
+    Ok(Arc::from(model))
+}
+
+/// Creates a directory for the specified model under `.spice/test_models`.
+pub async fn local_model_dir(model_name: &str) -> Result<PathBuf, anyhow::Error> {
```
Jeadie, line 89: I don't think we need to move them to `.spice/test_models`? Loading them from `~/.cache/huggingface/...` should be simpler?

crates/llms/tests/llms/mod.rs
```diff
@@ -196,24 +190,68 @@ async fn run_test_case(
...
+            continue;
         }
+        run_test_case(ts, model_name, Arc::clone(&model)).await?;
     }
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_anthropic_model() {
```
Jeadie, line 218: Why do we need a test per model?

Jeadie: Models can be filtered out before the model is loaded, see: https://github.com/spiceai/spiceai/pull/3690/files#diff-4162e2a84adbf9bc3bf9d99f2c9a12fdf04b440211a4dd46a02d0bd48936ae9cR155

crates/llms/tests/llms/create.rs
sgrebnov: As local model requires many parameters and we have hardcoded params only for one (currently support only single local model / model id) I think this is fine to have it as-is, it can be added later if we are be adding more. Please let me know if I miss anything and we can specify different model id for the currently used local model params

crates/llms/tests/llms/create.rs
sgrebnov: Good point - this was added to make the test more tailored to the target use case and ensure this scenario (custom folder) is supported.

crates/llms/tests/llms/mod.rs
sgrebnov: This allows control what tests to run, for example we don't currently support and don't want to run local models and HF tests on GH Hosted runners. This approach also allows to skip OpenAI or Anthropic tests if API KEY is not defined - better DX when running tests locally

sgrebnov: I prefer current approach as
1. it follows standard approach to run/skip tests
2. better DX as you can run/debug specific model tests directly from IDE  w/o configuring additional env variables

sgrebnov: Single test will work better if we don't want tests to run in parallel - for example, can local model test and HF test potentially conflict?

sgrebnov:
This was combined to the following PR, closing
https://github.com/spiceai/spiceai/pull/3690

Issue #3689, opened 2024-11-26
Postgres accelerator query 46 degradation after datafusion 43 upgrade


Sevenannn:
Postgres accelerator query 46 has performance regression after datafusion upgrade, the inexact filters will cause queries with duplicated filter conditions, caused by inexact filter pushdown.

For example, the filter below:
```
|                                                            |                         Filter: date_dim.d_dow IN ([Int64(6), Int64(0)]) AND date_dim.d_year IN ([Int64(1999), Int64(1999) + Int64(1), Int64(1999) + Int64(2)])                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
|                                                            |                           TableScan: date_dim projection=[d_date_sk, d_year, d_dow], partial_filters=[date_dim.d_dow IN ([Int64(6), Int64(0)]), date_dim.d_year IN ([Int64(1999), Int64(1999) + Int64(1), Int64(1999) + Int64(2)])]                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
```
Will be rewritten to duplicated condition
```
(((date_dim.d_dow IN (6, 0) AND date_dim.d_year IN (1999, (1999 + 1), (1999 + 2))) AND date_dim.d_dow IN (6, 0)) AND date_dim.d_year IN (1999, (1999 + 1), (1999 + 2))))
```

Note that this duplicated rewrite also affects other sql based accelerator.

PR #3694, opened 2024-11-26
Determine support_filter_pushdown based on Accelerator federated reader & ZeroResultsAction

Sevenannn:
## 🗣 Description

Determine support_filter_pushdown based on Accelerator federated reader & ZeroResultsAction.

When `ZeroResultsAction` is `ReturnEmpty`, the query won't fall back to the source. Therefore the query plan could be based on the filter push down rule of the federated accelerator, which is determined by the `supports_filter_pushdown` of the `PolyTableProvider.fed`.

After this change, the filter pushdown will be determined by the federated reader of PolyTableProvider when  `ZeroResultsAction` is `ReturnEmpty`. For arrow accelerator it will be unsupported, since arrow accelerator `MemTable` doesn't implement `supports_filter_pushdown`. For DuckDB / Postgres / SQLite tables it will be based on the `supports_filter_pushdown` of `SqlTable`, which is the same rule that Postgres/ DuckDB /etc. federated connectors are following. 

Successful benchmark test run, Postgres accelerator time reduced from 2.22 hours to 30 minutes. No regression on query results is observed. Noted that TPCH snapshots for arrow accelerator needs to be updated due to the plan change (Pushdown changed from inexact -> unsupported )https://github.com/spiceai/spiceai/actions/runs/12041194445

## 🔨 Related Issues

- https://github.com/spiceai/spiceai/issues/3689

## 🤔 Concerns

### DIFF
--- a/crates/runtime/src/accelerated_table.rs
+++ b/crates/runtime/src/accelerated_table.rs
@@ -666,7 +666,15 @@ impl TableProvider for AcceleratedTable {
         &self,
         filters: &[&Expr],
     ) -> DataFusionResult<Vec<TableProviderFilterPushDown>> {
-        Ok(vec![TableProviderFilterPushDown::Inexact; filters.len()])
+        match self.zero_results_action {
+            // Filter pushdown for all accelerators are currently exact
+            ZeroResultsAction::ReturnEmpty => {
+                Ok(vec![TableProviderFilterPushDown::Exact; filters.len()])
+            }
+            ZeroResultsAction::UseSource => {
+                Ok(vec![TableProviderFilterPushDown::Inexact; filters.len()])
+            }
+        }
     }
 
     async fn scan(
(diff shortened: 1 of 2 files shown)


crates/runtime/src/accelerated_table.rs
```diff
@@ -666,7 +666,13 @@ impl TableProvider for AcceleratedTable {
         &self,
         filters: &[&Expr],
     ) -> DataFusionResult<Vec<TableProviderFilterPushDown>> {
-        Ok(vec![TableProviderFilterPushDown::Inexact; filters.len()])
+        match self.zero_results_action {
+            // Filter pushdown for all accelerators are currently exact
```
sgrebnov:
```suggestion
```

sgrebnov (approved):

PR #3697, opened 2024-11-27
Fix acceleration in append mode with refresh_sql specified

sgrebnov:
## 🗣 Description

Fix acceleration when `append` mode is used with `refresh_sql`

Example test run: https://github.com/spiceai/spiceai/actions/runs/12046323524

## 🔨 Related Issues

N/A

N/A

## 🤔 Concerns

N/A


.github/workflows/e2e_test_ci.yml
```diff
@@ -516,6 +519,10 @@ jobs:
         run: |
           killall spice || true
           cat spice.log
+          if grep -iE "(failed|error)" spice.log; then
```
sgrebnov, line 565: Initial load can be successful and main verification logic will pass but refreshes can error so we check logs to identify any errors/failures

peasee (commented):
Could we make these new tests, rather than editing existing tests? e.g. E2E coverage with and without `append`, rather than just with `append`.

test/scripts/setup-data-mysql.sql
```diff
@@ -106,4 +88,26 @@ INSERT INTO test_mysql_table (
   NULL,
   NULL,
   NULL
+),
+-- Third dynamically generated test row with realistic values
```
sgrebnov:
```suggestion
-- Third test row
```

test/scripts/setup-data-postgresql.sql
```diff
@@ -72,11 +75,33 @@ INSERT INTO test_postgresql_table (
...
   NULL,
   NULL,
   NULL,
   NULL,
   NULL,
   NULL
-);
\ No newline at end of file
+),
+-- Third dynamically generated test row with realistic values
```
sgrebnov:
```suggestion
-- Third test row
```

.github/workflows/e2e_test_ci.yml
```diff
@@ -472,6 +474,43 @@ jobs:
           if [[ -n "$ENGINE" ]]; then
             echo "    acceleration:" >> spicepod.yaml
             echo "      enabled: true" >> spicepod.yaml
+            echo "      refresh_sql: SELECT * FROM test_postgresql_table WHERE int2_column is NULL or int2_column != 10" >> spicepod.yaml
```
sgrebnov, line 477: Learned today that `int2_column != 10` will also filter out records where value is NULL (native engines behavior)

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -514,23 +514,19 @@ impl RefreshTask {
...
             .limit(0, Some(1))
     }
 
-    async fn accelerator_df(
+    fn accelerator_df(
         &self,
-        ctx: SessionContext,
-        sql_opt: Option<&str>,
+        ctx: &SessionContext,
+        _sql_opt: Option<&str>,
```
Jeadie: Why keep `_sql_opt` as variable if unused?

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -550,10 +546,9 @@ impl RefreshTask {
 
         let existing_records = self
             .accelerator_df(
-                self.refresh_df_context(Arc::clone(&federated_provider)),
+                &self.refresh_df_context(Arc::clone(&federated_provider)),
```
Jeadie: Doesn't this change imply:
1. There are no existing records, 
2. There is no overlap between existing and the new records
3. There is no need to `filter_records` [below](https://github.com/spiceai/spiceai/pull/3697/files#diff-90f336ae367603afbc75d3643de47bae6a7f49eadbdc061aeaac5d1d70b4433eR569). 

IIRC, The need to diff current/new records is to handle timestamp jitter in the underlying systems that may have been missed in the previous append refresh.

Jeadie: Not super confident on this however.

crates/runtime/src/accelerated_table/refresh_task.rs
sgrebnov: Updated, thank you

crates/runtime/src/accelerated_table/refresh_task.rs
Jeadie: No, was mistaken.

Jeadie (approved):

ewgenius (approved):

phillipleblanc (approved):

lukekim (approved):

Issue #3702, opened 2024-11-27
No refresh and no errors when `time_*` keys needed but not specified.


Jeadie:
## Describe the bug
For an append accelerated dataset, if `time_column` and `time_format` are not specified, there are no logs signs that the dataset won't refresh on `refresh_check_interval`.

## To Reproduce
`spiced -v`

Logs erroring:
```
2024-11-27T03:35:12.880575Z DEBUG runtime::accelerated_table::refresh_task::changes: Inserting data row for transac
2024-11-27T03:35:12.880618Z DEBUG runtime::accelerated_table::refresh_task::changes: Inserting data row for transac
```

### Spicepod
```yaml
version: v1beta1
kind: Spicepod
name: ai-test-spicepod

datasets:
  - from: spice.ai/eth.recent_transactions
    name: transac
#   time_column: block_timestamp
#    time_format: unix_seconds
    params:
      spiceai_api_key: ${secrets:SPICE_SPICEAI_API_KEY}
    acceleration:
      enabled: true
      refresh_mode: append
      refresh_sql: "SELECT * FROM transac"
      refresh_check_interval: 1s
```

Sevenannn:
Yeah we should add something to user indicating that the `refresh_check_interval` won't be used when the connector is doing 'true refresh' - refresh without time_column specified.

Sevenannn:
Investigated the issue - this has been fixed by https://github.com/spiceai/spiceai/pull/3703, dataset won't be loaded and warning message is shown when user configure append dataset without `time_column` specified.

```bash
2025-03-05T02:30:55.689118Z  WARN runtime::init::dataset: Failed to setup the dataset transac (s3).
Acceleration mode `append` requires `time_column` parameter for source s3://spiceai-demo-datasets/taxi_trips/2024/.
Configure `time_column` parameter and try again.
For details, visit: https://spiceai.org/docs/reference/spicepod/datasets#time_column
```

PR #3703, opened 2024-11-27
Update error message when dataset isn't configured with time_column in append refresh

Sevenannn:
## 🗣 Description

A minor update to error message when dataset doesn't support append refresh mode without time column. Before the error message is misleading as it indicates the connector engine doesn't support refresh mode, whereas it's only because `time_column` is not configured.

Before:
image

After:
image


crates/runtime/src/datafusion.rs
```diff
@@ -197,8 +197,10 @@ pub enum Error {
     #[snafu(display("{source}"))]
     InvalidTimeColumnTimeFormat { source: refresh::Error },
 
-    #[snafu(display("Acceleration mode {mode} not supported for dataset from source {from}"))]
-    UnsupportedAccelerationMode { mode: String, from: String },
+    #[snafu(display(
+         "Acceleration mode `append` requires `time_column` parameter for source {from}.\nConfigure `time_column` parameter with a valid value.\nFor details, visit: https://docs.spiceai.org/reference/spicepod/datasets#time_column"
```
sgrebnov:
```suggestion
         "Acceleration mode `append` requires `time_column` parameter for source {from}.\nConfigure `time_column` parameter and try again.\nFor details, visit: https://docs.spiceai.org/reference/spicepod/datasets#time_column"
```

sgrebnov (dismissed):

crates/runtime/src/datafusion.rs
```diff
@@ -197,8 +197,10 @@ pub enum Error {
...
+    ))]
+    UnsupportedAppendAcceleration { from: String },
```
sgrebnov:
```suggestion
    AppendRequiresTimeColumn { from: String },
```

Jeadie:
Does this also fix https://github.com/spiceai/spiceai/issues/3702

Sevenannn:
https://github.com/spiceai/spiceai/issues/3702 is a different error fixing issue, spiceai connector supports append refresh without time column. We should provide users a warning (or something equivalent) when they are doing refresh without time column, telling user the `time_column` param won't be used in this case.

sgrebnov (approved):

PR #3714, opened 2024-11-27
Initialize accelerator for datasets dynamically included

Sevenannn:
## 🗣 Description

- Initialize accelerator for all valid datasets in `apply_dataset_diff`

## 🔨 Related Issues

This will fix the failure in https://docs.spiceai.org/features/federated-queries, where `spice add spiceai/fed-demo` causes warning of not finding SQLite database file in runtime log.


ewgenius (dismissed):

crates/runtime/src/init/dataset.rs
```diff
@@ -604,11 +605,16 @@ impl Runtime {
             if let Some(current_ds) = existing_datasets.iter().find(|d| d.name == ds.name) {
                 if ds != *current_ds {
                     self.update_dataset(ds).await;
+                } else if ds.is_file_accelerated() {
+                    let datasets = self.initialize_accelerators(&[Arc::clone(&ds)]).await;
```
phillipleblanc: Can't we call `initialize_accelerators for all `valid_datasets` before iterating over them one-by-one?

Jeadie (approved):

PR #3727, opened 2024-11-30
Fix NSQL models integration tests for HF

sgrebnov:
## 🗣 Description

PR fixes NSQL models integration tests for HF:
1. Re-use `run_nsql_test` logic from OpenAI tests
2. Improve system prompt to make nsql more robust:  added *don't wrap in ```* part.
3. 2. Add missing snapshots

Example test run: https://github.com/sgrebnov/spiceai/actions/runs/12099993195/job/33738154362

## 🔨 Related Issues

N/A

N/A

## 🤔 Concerns

N/A


crates/runtime/tests/models/hf.rs
```diff
@@ -118,142 +117,104 @@ async fn huggingface_test_search() -> Result<(), anyhow::Error> {
...
+                    },
+                ];
+
+                for ts in test_cases {
+                    run_nsql_test(http_base_url.as_str(), &ts, &trace_provider).await?;
+                }
 
-            insta::assert_snapshot!("nsql_sample_data_task_sql_query", sql_query_task.1);
+                tracing::info!("/v1/nsql: Ensure error when invalid dataset name is provided");
+                assert!(send_nsql_request(
```
Jeadie: Why can't these be test cases like above?

crates/runtime/tests/models/openai.rs
```diff
@@ -153,22 +90,22 @@ mod nsql {
 
             let test_cases = [
                 TestCase {
-                    name: "basic",
```
Jeadie, line 156: Another way to have the snapshots prefixed with openai is to change the module from `nsql` [here](https://github.com/spiceai/spiceai/pull/3727/files?diff=split&w=0#diff-03045e056979f2a71965a2dd7fd5de17313cd746ca68e51400dda723fdd0461bL49) and have it as openai.

crates/runtime/tests/models/hf.rs
Jeadie: It's because we expect an error? I wonder if we can account for that in `run_nsql_test` by not erroring on failed response [here](https://github.com/spiceai/spiceai/pull/3727/files?diff=split&w=0#diff-a56a01a97a0e053ada4281ce00dcfab87c5af4aa04bf3c0b58771e0730366168R72).

sgrebnov: Thank you - very good suggestion, improved

crates/runtime/tests/models/openai.rs
sgrebnov: Thank you Jack - I think we should prefer the same module name so we can get rid of duplicate snapshots in the future. Can we expect the same response and tasks for both OpenAI and HF so we can actually share the snapshots?

Jeadie: good idea yeah, there's definitely many tasks we could

Jeadie (approved):

Issue #3348, opened 2024-11-04
Enhancement: Data Connector Release Candidate: GraphQL


phillipleblanc:
## Goal-State/What/Result

The GraphQL Data Connector meets or exceeds the [release candidate criteria](https://github.com/spiceai/spiceai/blob/trunk/docs/criteria/connectors/rc.md).

## Why/Purpose

Part of our stated Roadmap: https://github.com/spiceai/spiceai/blob/trunk/docs/ROADMAP.md

## By When

_A target date for done-done completion of the entire enhancement._

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### All Connectors

These requirements are imposed on every connector, regardless of the selected features/tests in the above table.

The RC release criteria expand on and require that all [Beta release criteria](./beta.md) continue to pass for the connector.

- [x] All [Beta release criteria](./beta.md) pass for this connector.
- [x] For exceptions where a [Core Connector Data Type](../definitions.md) is not supported, print a warning and ignore the column instead of crashing or failing the query.
- [x] All known [Major Bugs](../definitions.md) are resolved.

#### Documentation

Documentation criteria should be re-checked on every release, to ensure the documentation is still accurate for the connector.

- [x] Documentation includes all information and steps for a user to set up the connector.
- [x] Documentation includes any mappings from [Core Connector Data Types](../definitions.md) to [Core Arrow Data Types](../definitions.md) types.
- [x] Documentation includes all known issues/limitations for the connector.
- [x] Documentation includes any exceptions made to allow this connector to reach RC quality (e.g. if a particular data type cannot be supported by the connector).
- [x] The connector has an easy to follow quickstart.
- [x] The connector status is updated in the table of connectors in [spiceai/docs](https://github.com/spiceai/docs).

### Conditional Criteria

[...]

PR #3756, opened 2024-12-03
test: Add GraphQL integration tests from external sources

peasee:
# 🗣 Description

* Adds some integration tests for GraphQL using external data sources

We should test using public sources to make sure in practice, the GraphQL connector works with other installations.

## 🔨 Related Issues

* Part of #3348

### DIFF
--- a/crates/runtime/tests/graphql/mod.rs
+++ b/crates/runtime/tests/graphql/mod.rs
@@ -447,7 +447,7 @@ async fn test_graphql_tcgdex() -> Result<(), String> {
             }
 
             let queries: QueryTests = vec![(
-                "SELECT * FROM tcgdex",
+                "SELECT * FROM tcgdex LIMIT 5",
                 "select_tcgdex",
                 Some(Box::new(|result_batches| {
                     for batch in result_batches {
@@ -533,7 +533,7 @@ async fn test_graphql_swapi() -> Result<(), String> {
             }
 
             let queries: QueryTests = vec![(
-                "SELECT * FROM swapi",
+                "SELECT * FROM swapi LIMIT 5",
                 "select_swapi",
                 Some(Box::new(|result_batches| {
                     for batch in result_batches {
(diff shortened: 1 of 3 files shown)


crates/runtime/tests/graphql/mod.rs
```diff
@@ -377,3 +393,173 @@ async fn test_graphql_pagination() -> Result<(), String> {
...
+
+            tokio::select! {
+                () = tokio::time::sleep(std::time::Duration::from_secs(10)) => {
+                    return Err("Timed out waiting for datasets to load".to_string());
+                }
+                () = rt.load_components() => {}
+            }
+
+            let queries: QueryTests = vec![(
+                "SELECT * FROM swapi",
```
sgrebnov: Nit: we might add LIMIT to speedup tests if query may return many rows

crates/runtime/tests/graphql/mod.rs
```diff
@@ -377,3 +393,173 @@ async fn test_graphql_pagination() -> Result<(), String> {
...
+
+            tokio::select! {
+                () = tokio::time::sleep(std::time::Duration::from_secs(10)) => {
+                    return Err("Timed out waiting for datasets to load".to_string());
+                }
+                () = rt.load_components() => {}
+            }
+
+            let queries: QueryTests = vec![(
+                "SELECT * FROM tcgdex",
```
sgrebnov: Nit: we might add LIMIT to speedup tests if query may return many rows

sgrebnov (dismissed):

crates/runtime/tests/graphql/mod.rs
```diff
@@ -209,12 +209,26 @@ async fn start_server() -> Result<(tokio::sync::oneshot::Sender<()>, SocketAddr)
...
     let mut dataset = Dataset::new(format!("graphql:{path}"), name.to_string());
-    let params = HashMap::from([
-        ("json_pointer".to_string(), json_pointer.to_string()),
-        ("graphql_query".to_string(), query.to_string()),
-    ]);
+    let params = if let Some(unnest_depth) = unnest_depth {
+        HashMap::from([
+            ("json_pointer".to_string(), json_pointer.to_string()),
+            ("graphql_query".to_string(), query.to_string()),
+            ("unnest_depth".to_string(), unnest_depth.to_string()),
```
sgrebnov: Might be better to make params mut and add unnest_depth separately to avoid duplicate code.

sgrebnov (approved):

PR #3765, opened 2024-12-03
Accelerate a subset of columns from source dataset in Refresh SQL

### LINKED ISSUE #3760: Enhancement: Accelerate a subset of columns from source dataset
## Goal-State/What/Result

Spice can accelerate only a subset of the full column set from a remote source into an acceleration.

## Why/Purpose

Reduce the amount of data transfer needed if only a subset of information is needed for an application to work.

## By When

**Issue/Spec written and reviewed:** Dec 3, 2024
**Done-Done:** Dec 3, 2024

## Done-Done

- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Release Notes prepared

## The Algorithm

- [x] Simplify.

## Specification

The `refresh_sql` parameter in the acceleration section of the dataset can be used to specify a subset of columns from the source schema, and only those columns will be included in the acceleration and fetched from the remote source.

Only bare column references will be supported in the SELECT. Computed columns, aliases or any other expressions are not allowed.

Changing the columns selected from the Refresh SQL after startup is not allowed.

### Security Review

Reviewed, no impact.

## How/Plan

The validation of the refresh sql will produce a new schema that is used instead of the federated source schema.

## Release Notes

Accelerate a subset of columns: Spice now supports accelerating a subset of the columns from a federated source by specifying the columns in the [Refresh SQL](https://docs.spiceai.org/components/data-accelerators/data-refresh#refresh-sql).

Example Spicepod.yaml configuration:
```yaml
datasets:
    name: taxi_trips
    params:
      file_format: parquet
    acceleration:
      refresh_sql: SELECT tpep_pickup_datetime, tpep_dropoff_datetime, trip_distance, total_amount FROM taxi_trips
```

phillipleblanc:
# 🗣 Description

Enables specifiying column references in the Refresh SQL to only accelerate a subset of columns from a source dataset.

i.e. this now works:

```yaml
datasets:
    name: taxi_trips
    description: taxi trips in s3
    params:
      file_format: parquet
    acceleration:
      enabled: true
      refresh_sql: SELECT tpep_pickup_datetime, tpep_dropoff_datetime, trip_distance, total_amount FROM taxi_trips
```

This was fairly straightforward to implement, the bulk of the PR handling various error cases that can come up when parsing the Refresh SQL and validating it against the source schema. We also don't allow changing the schema after the dataset is registered.

## 🔨 Related Issues

Closes #3760

## 📄 Documentation Updates

We can remove the limitation of Refresh SQL not being able to specify column references.
https://github.com/spiceai/docs/pull/664

### DIFF
--- a/crates/runtime/src/datafusion/refresh_sql.rs
+++ b/crates/runtime/src/datafusion/refresh_sql.rs
@@ -95,13 +95,6 @@ pub fn validate_refresh_sql(
                     }
                 );
                 ensure!(
-                    query.limit.is_none(),
-                    UnexpectedExpressionSnafu {
-                        expr: "LIMIT",
-                        expected_table,
-                    }
-                );
-                ensure!(
                     query.offset.is_none(),
                     UnexpectedExpressionSnafu {
                         expr: "OFFSET",
--- a/crates/runtime/src/datafusion/refresh_sql.rs
+++ b/crates/runtime/src/datafusion/refresh_sql.rs
@@ -238,3 +238,104 @@ fn validate_select_columns(
 
     Ok(Arc::new(Schema::new(fields)))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use datafusion::arrow::datatypes::{DataType, Field, Schema};
+
+    fn create_test_schema() -> Arc<Schema> {
+        Arc::new(Schema::new(vec![
+            Field::new("id", DataType::Int64, false),
+            Field::new("name", DataType::Utf8, false),
+            Field::new("value", DataType::Float64, true),
+        ]))
+    }
+
+    #[test]
+    fn test_valid_select_all() -> Result<()> {
+        let schema = create_test_schema();
+        let table = TableReference::parse_str("test_table");
+        let sql = "SELECT * FROM test_table";
+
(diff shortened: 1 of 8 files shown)


crates/runtime/src/datafusion/refresh_sql.rs
```diff
@@ -55,49 +86,377 @@ pub fn validate_refresh_sql(expected_table: TableReference, refresh_sql: &str) -
...
+                                expected_table,
+                            }
+                        );
+                        ensure!(
+                            select.qualify.is_none(),
+                            UnexpectedExpressionSnafu {
+                                expr: "QUALIFY",
+                                expected_table,
+                            }
+                        );
```
peasee: could we make this a `macro_rules!` pattern? it's repeated quite a lot in similar ways

peasee (dismissed):

crates/runtime/src/datafusion/refresh_sql.rs
phillipleblanc: Addressed

peasee (approved):

PR #3777, opened 2024-12-05
fix: Check ODBC parameters are positive integers

### LINKED ISSUE #3561: ODBC errors when setting `odbc_max_num_rows_per_batch` to 0
## Describe the bug

ODBC errors or has returns no results when specifying some parameter values as `0`:

* `odbc_max_bytes_per_batch` - returns `No results.`
* `odbc_max_num_rows_per_batch` - Panics on the inner thread task
* `odbc_max_text_size` and `odbc_max_binary_size` when set to `0` makes no observable change in my testing

## To Reproduce

1. Configure an odbc dataset with parameter values set to `0`
2. Observe behavior

## Expected behavior

The Runtime should return an invalid configuration error when setting parameter values to `0`, as they are not supported.

### Spicepod

```yaml
version: v1beta1
kind: Spicepod
name: spicepod
datasets:
  - from: odbc:table
    name: my_table
    params:
      odbc_connection_string: odbc_connection_string
      odbc_max_bytes_per_batch: 0
      odbc_max_num_rows_per_batch: 0
      odbc_max_text_size: 0
      odbc_max_binary_size: 0
```

- [x] Yes
- [ ] No

peasee:
# 🗣 Description

* Adds validation to some ODBC parameters that are expected to be positive integers.

## 🔨 Related Issues

* Closes #3561

### DIFF
--- a/crates/runtime/src/dataconnector/odbc.rs
+++ b/crates/runtime/src/dataconnector/odbc.rs
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 use crate::component::dataset::Dataset;
+use crate::parameters::Parameters;
 use async_trait::async_trait;
 use data_components::odbc::ODBCTableFactory;
 use data_components::Read;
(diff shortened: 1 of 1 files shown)


crates/runtime/src/dataconnector/odbc.rs
```diff
@@ -176,12 +176,54 @@ fn driver_is_file(driver: &str) -> bool {
...
+}
+
 impl DataConnectorFactory for ODBCFactory {
     fn create(
         &self,
         params: DataConnectorParams,
     ) -> Pin<Box<dyn Future<Output = super::NewDataConnectorResult> + Send>> {
         Box::pin(async move {
+            if let Some(binary_size) = params.parameters.get("max_binary_size").expose().ok() {
+                parameter_is_integer!(binary_size);
```
phillipleblanc: I'm not a big fan of macros that obscure the control flow - I would prefer the macro just evaluate to a Result that can be returned directly here.

crates/runtime/src/dataconnector/odbc.rs
```diff
@@ -176,12 +177,37 @@ fn driver_is_file(driver: &str) -> bool {
         .is_some()
 }
 
+fn parameter_is_integer(parameters: &Parameters, param: &str) -> Result<()> {
+    if let Some(value) = parameters.get("max_binary_size").expose().ok() {
```
phillipleblanc:
```suggestion
    if let Some(value) = parameters.get(param).expose().ok() {
```

phillipleblanc (approved):

Sevenannn (approved):

PR #3791, opened 2024-12-06
Enable `trace_id` & `parent_span_id` overrides for `v1/chat/completion`

Jeadie:
# 🗣 Description
 - Enable the ability to override the `trace_id` & `parent_span_id` for any `runtime.task_history` trace. This is useful for when a trace id wants to be known before it's eventually written to an exporter.
 - `parent_span_id` is also overridable so that HTTP endpoints can honour [traceparent](https://www.w3.org/TR/trace-context/#traceparent-header) HTTP header. 

How to do this in code: 
```rust
tracing::info!(target: "task_history", parent: &span.clone(), trace_id = %trace_id, parent_id = %parent_span_id);

// or
let headers: HeaderMap = ...
override_task_history_with_traceparent(&span, &headers);
 ```

 As an example use, in `v1/chat/completion` now supports traceparent Header.
```shell
curlie http://localhost:8090/v1/chat/completions \
  -H "traceparent: 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01" \
  -d '{
    "model": "oai",
    "messages":[
            {
                "content": "Write me a poem",
                "role": "user"
            }
        ],
    "store": true
}'
```

- We also ensure that the override is [w3 compliant](https://www.w3.org/TR/trace-context/#trace-id).

### DIFF
--- a/crates/runtime/src/http/v1/chat.rs
+++ b/crates/runtime/src/http/v1/chat.rs
@@ -41,6 +41,12 @@ pub(crate) async fn post(
     let span = tracing::span!(target: "task_history", tracing::Level::INFO, "ai_chat", input = %serde_json::to_string(&req).unwrap_or_default());
     span.in_scope(|| tracing::info!(target: "task_history", model = %req.model, "labels"));
 
+    if let Some(ref metadata) = req.metadata {
+        if let Some(serde_json::Value::String(trace_id)) = metadata.get("trace_id") {
+            tracing::info!(target: "task_history", parent: &span.clone(), trace_id = %trace_id);
+        }
+    };
+
     let span_clone = span.clone();
     async move {
         let model_id = req.model.clone();
--- a/crates/runtime/src/http/v1/chat.rs
+++ b/crates/runtime/src/http/v1/chat.rs
@@ -17,10 +17,11 @@ limitations under the License.
 use core::time;
 use std::{convert::Infallible, sync::Arc, time::Duration};
 
+use crate::{http::traceparent::override_task_history_with_traceparent, model::LLMModelStore};
 use async_openai::types::{ChatCompletionResponseStream, CreateChatCompletionRequest};
 use async_stream::stream;
 use axum::{
-    http::StatusCode,
+    http::{HeaderMap, StatusCode},
     response::{
         sse::{Event, KeepAlive, Sse},
         IntoResponse, Response,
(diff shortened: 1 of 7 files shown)


crates/runtime/src/task_history/otel_exporter.rs
```diff
@@ -147,10 +170,27 @@ impl TaskHistoryExporter {
...
+                    .as_ref()
+                    .map(|new_trace| (Arc::clone(&span.trace_id), Arc::clone(new_trace)))
+            })
+            .collect();
+
+        for span in &mut spans {
+            if let Some(new_trace_id) = overrides.get(&span.trace_id) {
+                span.trace_id = Arc::clone(new_trace_id);
+            }
+        }
```
phillipleblanc, line 199: Why are we doing a double iteration over the spans here? Couldn't we just directly apply the override with a single loop?

crates/runtime/src/http/v1/chat.rs
```diff
@@ -41,6 +41,12 @@ pub(crate) async fn post(
     let span = tracing::span!(target: "task_history", tracing::Level::INFO, "ai_chat", input = %serde_json::to_string(&req).unwrap_or_default());
     span.in_scope(|| tracing::info!(target: "task_history", model = %req.model, "labels"));
 
+    if let Some(ref metadata) = req.metadata {
+        if let Some(serde_json::Value::String(trace_id)) = metadata.get("trace_id") {
```
phillipleblanc: Shouldn't this come from a `traceparent` header instead? https://www.w3.org/TR/trace-context/#traceparent-header

Jeadie: yeah that's a good point

crates/runtime/src/task_history/otel_exporter.rs
Jeadie: From my understanding of how we accumulate the spans, if 2 or more spans are in a trace, only 1 of them may have the override. We then need to apply the override to the remaining >1 spans.

crates/runtime/src/http/v1/chat.rs
Jeadie: I have added this as a `traceparent` header. This means i've also added parent span id propagation.

phillipleblanc (approved):

Issue #3608, opened 2024-11-21
Use MinIO for S3 TPCH / TPCDS tests


Sevenannn:

Sevenannn:
- Closed by: https://github.com/spiceai/spiceai/pull/3794

PR #3794, opened 2024-12-06
Use MinIO instead of S3 for benchmark tests

Sevenannn:
# 🗣 Description

Use MinIO as source for the following tests in benchmark:
* S3 Tests
* All accelerator tests
* File connector
* DuckDB tests

This will help lower the s3 cost, and potentially decrease network latency by using the s3 endpoint hosted inside spice infrastructure.

Successful benchmark run: 

**Note:** There will be slight query plan changes due to the parquet file path change and results change for file connector due to the data source change (using parquet data originally in `spice-demo-dataset` bucket instead of the ones in `spice-public-dataset`). The snapshots updates will be addressed in another PR.

## 🔨 Related Issues

- https://github.com/spiceai/spiceai/issues/3608

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/runtime/benches/bench_object_store/s3.rs
+++ b/crates/runtime/benches/bench_object_store/s3.rs
@@ -23,151 +23,163 @@ pub fn build_app(app_builder: AppBuilder, bench_name: &str) -> Result<AppBuilder
     match bench_name {
         "tpch" => Ok(app_builder
             .with_dataset(make_dataset(
-                "benchmarks/tpch/customer/",
+                "benchmarks/tpch_sf1/customer/",
                 "customer",
                 bench_name,
             ))
             .with_dataset(make_dataset(
-                "benchmarks/tpch/lineitem/",
+                "benchmarks/tpch_sf1/lineitem/",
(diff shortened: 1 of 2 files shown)


sgrebnov (approved):

crates/runtime/benches/bench_object_store/s3.rs
```diff
@@ -22,325 +22,195 @@ use spicepod::component::{dataset::Dataset, params::Params};
...
-                "spiceai-demo-datasets/tpch/region/",
-                "region",
-                bench_name,
-            ))
-            .with_dataset(make_dataset(
-                "spiceai-demo-datasets/tpch/supplier/",
-                "supplier",
-                bench_name,
-            ))),
+            .with_dataset(make_dataset("benchmarks/tpch_sf1/customer/", "customer"))
```
sgrebnov, line 25: Might be good to have s3 root as a constant so I can update it back to `spiceai-public-datasets/` if I want to run/debug tests locally

phillipleblanc (approved):

Issue #3351, opened 2024-11-04
Enhancement: Data Connector Release Candidate: ODBC Data Connector - Athena, Databricks


phillipleblanc:
## Goal-State/What/Result

The ODBC Data Connector for Athena and Databricks meets or exceeds the [release candidate criteria](https://github.com/spiceai/spiceai/blob/trunk/docs/criteria/connectors/rc.md).

## Why/Purpose

Part of our stated Roadmap: https://github.com/spiceai/spiceai/blob/trunk/docs/ROADMAP.md

## By When

_A target date for done-done completion of the entire enhancement._

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Samples/Quickstarts/Tutorials
- [ ] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### All Connectors

These requirements are imposed on every connector, regardless of the selected features/tests in the above table.

The RC release criteria expand on and require that all [Beta release criteria](./beta.md) continue to pass for the connector.

- [ ] All [Beta release criteria](./beta.md) pass for this connector.
- [ ] For exceptions where a [Core Connector Data Type](../definitions.md) is not supported, print a warning and ignore the column instead of crashing or failing the query.
- [ ] All known [Major Bugs](../definitions.md) are resolved.

#### Documentation

Documentation criteria should be re-checked on every release, to ensure the documentation is still accurate for the connector.

- [ ] Documentation includes all information and steps for a user to set up the connector.
- [ ] Documentation includes any mappings from [Core Connector Data Types](../definitions.md) to [Core Arrow Data Types](../definitions.md) types.
- [ ] Documentation includes all known issues/limitations for the connector.
- [ ] Documentation includes any exceptions made to allow this connector to reach RC quality (e.g. if a particular data type cannot be supported by the connector).
- [ ] The connector has an easy to follow quickstart.
[...]

PR #3825, opened 2024-12-11
feat: add Databricks ODBC TPCDS benchmark

peasee:
# 🗣 Description

* Adds the TPCDS benchmark for ODBC Databricks.

## 🔨 Related Issues

* Part of #3351


crates/runtime/benches/utils/mod.rs
```diff
@@ -24,7 +24,7 @@ pub(crate) fn init_tracing(trace_config: Option<&str>) {
         (_, Some(log)) => EnvFilter::new(log),
         (Some(level), None) => EnvFilter::new(level),
         _ => EnvFilter::new(
-            "runtime=TRACE,datafusion-federation=TRACE,datafusion-federation-sql=TRACE,bench=TRACE,INFO",
+            "runtime=TRACE,datafusion-federation=TRACE,datafusion-federation-sql=TRACE,bench=TRACE",
```
phillipleblanc: Why was this removed?

phillipleblanc (dismissed):

crates/runtime/benches/utils/mod.rs
peasee: `INFO` was enabling `task_history` log spans, which was causing a lot of logs in the benchmark runs making it hard to debug. I've updated with a new log span copying the default from `spiced` but disabling `task_history` - I'm testing the output of this new span now to see what it looks like

phillipleblanc (approved):

Issue #3684, opened 2024-11-26
Enhancement: Data Connector Beta: MS SQL


peasee:
## Goal-State/What/Result

The MS SQL Data Connector meets or exceeds the [beta criteria](https://github.com/spiceai/spiceai/blob/trunk/docs/criteria/connectors/beta.md).

## Why/Purpose

Part of our stated Roadmap: https://github.com/spiceai/spiceai/blob/trunk/docs/ROADMAP.md

## By When

_A target date for done-done completion of the entire enhancement._

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### All Connectors

The Beta release criteria expand on and require that all [Alpha release criteria](./alpha.md) continue to pass for the connector.

- [x] All [Alpha release criteria](./alpha.md) pass for this connector.
- [x] [Core Connector Data Types](../definitions.md) are supported.

#### Documentation

Documentation criteria should be re-checked on every release, to ensure the documentation is still accurate for the connector.

- [x] Documentation includes all information and steps for a user to set up the connector.
- [x] Documentation includes all known issues/limitations for the connector.
- [x] Documentation includes any exceptions made to allow this connector to reach Beta quality (e.g. if a particular data type cannot be supported by the connector).
- [x] The connector has an easy to follow quickstart.
- [x] The connector status is updated in the table of connectors in [spiceai/docs](https://github.com/spiceai/docs).

#### UX

- [x] All of the connector's error messages follow the [error handling guidelines](../../dev/error_handling.md)

#### Testing

- [x] Integration tests to cover simple use cases based on the connector type, e.g. a Spicepod sourcing a file from an FTP server, reading latest GitHub issues, etc.

### Conditional Criteria

The following features/tests are dependent on the required features/tests for the specified connector, from the [Connector criteria table](#beta-release-criteria)

[...]

PR #3833, opened 2024-12-11
feat: add MS SQL TPCH benchmark

peasee:
# 🗣 Description

* Adds the TPCH benchmark for MS SQL
* Adds benchmark image building actions/scripts for MS SQL

## 🔨 Related Issues

* Part of #3684


crates/runtime/benches/bench_mssql/tpch_q12.sql
```diff
@@ -0,0 +1,30 @@
+select
+    l_shipmode,
+    sum(case
+            when trim(o_orderpriority) = '1-URGENT'
```
peasee, line 4: The use of `trim()` should be able to be fixed in the arrow converter for MS SQL, by trimming where the source type is `CHAR`. I'll address this in another PR.

TL;DR - `SELECT *` with a `CHAR` in MS SQL returns space-padded values up to the length of the char. However, when MS SQL performs `WHERE` comparisons against `CHAR` columns, it strips the extra padded spaces during the comparison.

This creates problems when we encounter filters that are not pushed down, because the values are returned with padded spaces but the Arrow memory execution doesn't know/expect that.

test/tpc-bench/mssql-tpch/data/.gitkeep
Sevenannn, line 1: Should the git keep file be removed?

test/tpc-bench/Makefile
```diff
@@ -238,4 +238,25 @@ file-tpch-load:
...
+	sqlcmd -U $(DB_USER) -P $(DB_PASS) -S $(DB_HOST) -d master -Q "IF NOT EXISTS (SELECT name FROM master.dbo.sysdatabases WHERE name = N'$(DB_NAME)') CREATE DATABASE $(DB_NAME);"
+	sqlcmd -U $(DB_USER) -P $(DB_PASS) -S $(DB_HOST) -d $(DB_NAME) -i ./mssql-tpch/up.sql
+	@echo "Database $(DB_NAME) has been successfuly created or updated."
+
+mssql-load:
+	@for table in region nation customer supplier part partsupp orders lineitem; do \
+		cp tpch-kit/dbgen/$$table.tbl ./mssql-tpch/data/$$table.tbl; \
+	done
+	chmod 755 ./mssql-tpch/data
+	sqlcmd -U $(DB_USER) -P $(DB_PASS) -S $(DB_HOST) -d $(DB_NAME) -i ./mssql-tpch/insert.sql
```
Sevenannn, line 253: The `dss.ddl` in TPCH kit can be reused here instead of creating a separate SQL file for initiating table

Sevenannn (dismissed):

test/tpc-bench/Makefile
peasee: I elected not to use it, because it doesn't create primary or foreign keys. `dss.ri` would be troublesome to reuse without some re-writing, and both files need to have `GO` added to the end of them. I settled on making a new file instead.

phillipleblanc (approved):

Issue #3803, opened 2024-12-07
Embeddings creation on `metal` is slower than MTEB benchmark


sgrebnov:
## Describe the bug

1. Run `QuoraRetrievalHardNegatives` with `sentence-transformers/all-MiniLM-L6-v2` using MTEB cli  (1m 43s)
```
rm -fr  ~/.cache/huggingface/datasets/
mteb run -m sentence-transformers/all-MiniLM-L6-v2 -t QuoraRetrievalHardNegatives
```

1. Run Spice benchmark and observe index_time 17m

```
cargo bench -p runtime --bench vector_search --features models,metal
```

#### MTEB
```
mteb run -m sentence-transformers/all-MiniLM-L6-v2 -t QuoraRetrievalHardNegatives
INFO:mteb.cli:Running with parameters: Namespace(model='sentence-transformers/all-MiniLM-L6-v2', task_types=None, categories=None, tasks=['QuoraRetrievalHardNegatives'], languages=None, benchmarks=None, device=None, output_folder='results', verbosity=2, co2_tracker=False, eval_splits=None, model_revision=None, batch_size=None, overwrite=False, save_predictions=False, func=<function run at 0x311adc220>)
WARNING:mteb.model_meta:Loader not specified for model sentence-transformers/all-MiniLM-L6-v2, loading using sentence transformers.
INFO:mteb.evaluation.MTEB:

## Evaluating 1 tasks:
───────────────────────────────────────────────────────────────────────────────────────────── Selected tasks  ──────────────────────────────────────────────────────────────────────────────────────────────
Retrieval
    - QuoraRetrievalHardNegatives, s2s


INFO:mteb.evaluation.MTEB:

********************** Evaluating QuoraRetrievalHardNegatives **********************
INFO:mteb.evaluation.MTEB:Running evaluation for missing splits: ['test']
INFO:mteb.abstasks.AbsTaskRetrieval:Loading Corpus...
hf://datasets/mteb/QuoraRetrieval_test_top_250_only_w_correct-v2@907a33577e9506221d3ba20f5a851b7c3f8dc6d3/README.md not found in cache or force_download set to True, downloading to /Users/sg/.cache/huggingface/datasets/downloads/229d4ea40899dc2430f96e2bca39d2725c607a438dce103f2014dd0cdc94180e.incomplete
[...]

PR #3834, opened 2024-12-11
Add Github Workflow to run Search Benchmark

### LINKED ISSUE #3800: Search benchmark: scheduled run via GitHub workflow

sgrebnov:
# 🗣 Description

Add Github Workflow to run Search Benchmark: manually and every Saturday at 10 AM UTC (2AM PT)

Example benchmark run: https://github.com/spiceai/spiceai/actions/runs/12285418904 (note: miniml score is low as it runs only on the subset of data - see note below)

## 🔨 Related Issues

Closes https://github.com/spiceai/spiceai/issues/3800

## 🤔 Concerns

Local embeddings are slow so currently limited to only `20000` test records. There is the following item to investigate/address this: https://github.com/spiceai/spiceai/issues/3803. We can update workflow to run tests on self-hosted MacOS runner with Metal acceleration support which is faster, but https://github.com/spiceai/spiceai/issues/3803 should be completed first.


Sevenannn (approved):
lgtm. Where do we currently track the search bench result? Is it through a read-write spiceai dataset like connector / accelerator bench?

sgrebnov:
> lgtm. Where do we currently track the search bench result? Is it through a read-write spiceai dataset like connector / accelerator bench?

Yeah - we don't currently write results but working on approach to write them into `spiceai.benchamarks.oss_search_benchmarks` dataset. This will be added next.

PR #3852, opened 2024-12-13
fix: trim CHAR in mssql

peasee:
# 🗣 Description

* Trims the padded spaces at the end of `CHAR` data types

MS SQL returns space-padded `CHAR` values, but trims the spaces when performing pushed-down filters that use `CHAR` columns. For filters that cannot be pushed down, they no longer work effectively as `text   ` is not equal to `text`.

* Updates the benchmarks to use the regular q12 and q19 queries, instead of the workarounds with `trim()`.

## 🔨 Related Issues

* Part of #3684


crates/data_components/src/mssql/convert.rs
```diff
@@ -278,7 +278,13 @@ pub(crate) fn rows_to_arrow(rows: &[Row], schema: &SchemaRef) -> super::Result<R
                     };
                     let v = row.get::<&str, usize>(i);
                     match v {
-                        Some(v) => builder.append_value(v),
+                        Some(v) => {
+                            if matches!(mssql_type, ColumnType::BigChar | ColumnType::NChar) {
+                                builder.append_value(v.trim_end());
```
phillipleblanc: Add a comment for why we do this

phillipleblanc (dismissed):

Sevenannn (dismissed):

phillipleblanc (approved):

PR #3900, opened 2024-12-17
Ensure eval dataset input and ouput of correct length

Jeadie:
# 🗣 Description

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/runtime/src/model/eval/mod.rs
+++ b/crates/runtime/src/model/eval/mod.rs
@@ -18,10 +18,7 @@ use std::{collections::HashMap, sync::Arc};
 use tokio::sync::RwLock;
 
 use arrow_schema::ArrowError;
-use async_openai::{
-    error::{ApiError, OpenAIError},
-    types::CreateChatCompletionRequest,
-};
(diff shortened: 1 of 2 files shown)


crates/runtime/src/model/eval/mod.rs
```diff
@@ -193,6 +213,22 @@ async fn run_eval(
         vec![]
     };
 
+    if actual.len() != ideal.len() {
```
phillipleblanc: The Snafu `ensure` macro can make this a bit nicer: https://docs.rs/snafu/latest/snafu/macro.ensure.html

phillipleblanc (dismissed):

phillipleblanc (approved):

crates/runtime/src/model/eval/mod.rs
```diff
@@ -58,6 +58,12 @@ pub enum Error {
         source: Box<dyn std::error::Error + Send + Sync>,
```
russellwmy, line 58: non-block, maybe  create trait to generalize error.

PR #3906, opened 2024-12-17
Ensure trace_id is overridden for prior written spans

Jeadie:
# 🗣 Description
Limitation in prior implementation of overriding `trace_id` in `runtime.task_history`. Only applies the trace_id override to spans in the same batch being exported. Below is an example of the effect (for large traces). 

```
+----------------------------------+------------------+---------------------+
| trace_id                         | span_id          | task                |
+----------------------------------+------------------+---------------------+
| 4006b51208dbcd4dd9e2711e9370a743 | 6d2dd0bfb1b9d769 | eval_run            |  <- `eval_run` is always
| 31cf86e5572330ac71bc787e10600d81 | 7749fce6f2e00baa | sql_query           |---+   exported last.
| 31cf86e5572330ac71bc787e10600d81 | 8d32124df669bc05 | ai_completion       |   |
| 31cf86e5572330ac71bc787e10600d81 | 568ee8b20203b641 | ai_completion       |   | These are in same trace 
| 31cf86e5572330ac71bc787e10600d81 | 46279bb639ade86d | ai_completion       |   | 
| 31cf86e5572330ac71bc787e10600d81 | 002d92285acca2b5 | ai_completion       |   |
| 31cf86e5572330ac71bc787e10600d81 | 3de9a06e21f4d54b | ai_completion       |---+
| 4006b51208dbcd4dd9e2711e9370a743 | 5aceb0da1846095e | ai_completion       |--+
| 4006b51208dbcd4dd9e2711e9370a743 | 1546f815c2425bff | ai_completion       |  | They should be with these.
| 4006b51208dbcd4dd9e2711e9370a743 | 2865cf8e77fb6915 | ai_completion       |  |
| 4006b51208dbcd4dd9e2711e9370a743 | 53d2883765d9cdd5 | ai_completion       |  |
| 4006b51208dbcd4dd9e2711e9370a743 | 21634592ca92fdb1 | sql_query           |--+
+----------------------------------+------------------+---------------------+
```

- This PR fixes it. 

## 🤔 Concerns
 - Adding a primary key to `runtime.task_history` just for this feature.

### DIFF
--- a/crates/runtime/src/task_history.rs
+++ b/crates/runtime/src/task_history.rs
@@ -16,15 +16,17 @@ limitations under the License.
 
 use crate::accelerated_table::refresh::Refresh;
 use crate::datafusion::DataFusion;
-use crate::dataupdate::DataUpdate;
+use crate::dataupdate::{DataUpdate, UpdateType};
 use crate::internal_table::create_internal_accelerated_table;
 use crate::status;
 use crate::{component::dataset::acceleration::Acceleration, datafusion::SPICE_RUNTIME_SCHEMA};
 use crate::{component::dataset::TimeFormat, secrets::Secrets};
-use arrow::array::{ArrayBuilder, MapBuilder, RecordBatch, StringBuilder};
+use arrow::array::{ArrayBuilder, MapBuilder, RecordBatch, StringArray, StringBuilder};
 use arrow::datatypes::{DataType, Field, Schema, TimeUnit};
+use arrow_schema::ArrowError;
 use data_components::arrow::struct_builder::StructBuilder;
 use datafusion::sql::TableReference;
+use futures::TryStreamExt;
 use snafu::prelude::*;
 use snafu::{ResultExt, Snafu};
 use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


crates/runtime/src/task_history.rs
```diff
@@ -169,6 +180,66 @@ impl TaskSpan {
...
+            .collect::<Result<_, ArrowError>>()
+            .boxed()
+            .context(UnableToUpdateTracesSnafu)?;
+
+        df.write_data(
+            &TableReference::partial(SPICE_RUNTIME_SCHEMA, DEFAULT_TASK_HISTORY_TABLE),
+            DataUpdate {
+                schema: Arc::new(Self::table_schema()),
+                data: overriden,
+                update_type: UpdateType::Changes,
```
Sevenannn, line 236: lgtm, just a question for learning purposes, is having `span_id` as primary key needed for the changes update here, locating rows with trace_id override changes and make updates?

Sevenannn (approved):

crates/runtime/src/task_history.rs
Jeadie: Yes, we don't need it for the initial read (i.e. `WHERE trace_id='{}'`), but is needed so that when we insert these rows back in, we don't get duplicates, but rather the original rows are overwritten.

PR #3920, opened 2024-12-18
Add analyzer rule to override Spice internal UDFs (`cosine_distance`)

### LINKED ISSUE #3796: Bug: search with duckdb acceleration fails with 'Scalar Function with name cosine_distance does not exist' error
## Describe the bug

When a dataset with embeddings uses duckdb acceleration there is the following error during search:

>2024-12-07T19:21:00.261118Z ERROR task_history: Error occurred interacting with datafusion: Execution error: Failed to execute query.
Catalog Error: Scalar Function with name cosine_distance does not exist!
Did you mean "cos"?
LINE 1: ...M (SELECT "subq"."_id", "subq"."text", cosine_distance("subq"."text_embedding"...
                                                  ^

Note: assume the same error when using sqlite and postgres acceleration as `cosine_distance` is Spice-defined UDF and should be correctly translated to other engines (DuckDB natively supports this via `array_cosine_distance` https://duckdb.org/docs/sql/functions/array.html#array_cosine_distancearray1-array2)

sgrebnov:
# 🗣 Description

Add `spiceai_udf_override_rule` analyzer rule to override Spice internal UDFs to match the signature of the target execution engine. Currently supports overriding `cosine_distance` for DuckDB only.

Pending the following PRs:
 - datafusion: [Enable downcasting support for Dialect trait](https://github.com/spiceai/datafusion/pull/67)
 - datafusion-federation: [Enable downcasting for FederationProvider and expose Dialect information ](https://github.com/spiceai/datafusion-federation/pull/27)
 - datafusion-table-providers: to be updated after datafusion-federation is merged

## 🔨 Related Issues

Fixes https://github.com/spiceai/spiceai/issues/3796

## 🤔 Concerns

 - Need to add tests

### DIFF
--- a/crates/runtime/src/datafusion/builder.rs
+++ b/crates/runtime/src/datafusion/builder.rs
@@ -36,7 +36,10 @@ use datafusion::{
 use datafusion_federation::FederationAnalyzerRule;
 use tokio::sync::RwLock as TokioRwLock;
 
-use crate::{embeddings, object_store_registry::default_runtime_env, status};
+use crate::{
+    datafusion::spice_udf_override::SpiceUDFsOverride, embeddings,
+    object_store_registry::default_runtime_env, status,
+};
 
 use super::{
     extension::{bytes_processed::BytesProcessedOptimizerRule, SpiceQueryPlanner},
@@ -170,6 +173,7 @@ fn get_analyzer_rules() -> Vec<Arc<dyn AnalyzerRule + Send + Sync>> {
     vec![
         Arc::new(InlineTableScan::new()),
         Arc::new(ExpandWildcardRule::new()),
+        Arc::new(SpiceUDFsOverride::new()),
         Arc::new(FederationAnalyzerRule::new()),
         // The rest of these rules are run after the federation analyzer since they only affect internal DataFusion execution.
         Arc::new(ResolveGroupingFunction::new()),
--- a/crates/runtime/src/datafusion/builder.rs
+++ b/crates/runtime/src/datafusion/builder.rs
@@ -103,6 +103,7 @@ impl DataFusionBuilder {
             .with_query_planner(Arc::new(SpiceQueryPlanner::new()))
             .with_runtime_env(default_runtime_env())
             .with_analyzer_rules(get_analyzer_rules())
+            .with_optimizer_rule(Arc::new(SpiceUDFsOverride::new()))
             .build();
 
         if let Err(e) = datafusion_functions_json::register_all(&mut state) {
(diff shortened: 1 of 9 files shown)


crates/runtime/src/datafusion/builder.rs
```diff
@@ -170,6 +173,7 @@ fn get_analyzer_rules() -> Vec<Arc<dyn AnalyzerRule + Send + Sync>> {
     vec![
         Arc::new(InlineTableScan::new()),
         Arc::new(ExpandWildcardRule::new()),
+        Arc::new(SpiceUDFsOverride::new()),
```
sgrebnov: New rule is added before FederationAnalyzerRule so it is universal and could be used with and without federation (same logic w/o additional implementation specifically for FederationAnalyzerRule)

crates/runtime/src/datafusion/spice_udf_override.rs
```diff
@@ -0,0 +1,273 @@
...
+    let mut modified = false;
+    let exprs: Vec<Expr> = exprs
+        .into_iter()
+        .map(|expr| {
+            // recursive `transform` is required as UDF can be nested in the expression tree
+            expr.transform(|expr| {
+                if let Expr::ScalarFunction(scalar_fn) = expr {
+                    let udf_fn_name = scalar_fn.name();
+                    match udf_fn_name {
+                        "cosine_distance" if is_duckdb_accelerator_provider(input)? => {
```
phillipleblanc: This only makes sense within the context of federation - otherwise the Spice UDF can be used normally by DataFusion. It probably makes sense to add this as an optimizer rule that we run on a sub-plan carved out by federation, since we know what the federated source is at that point. And that avoids this complicated logic to try to guess what the relevant engine is.

sgrebnov: @phillipleblanc  - updated, thank you

crates/runtime/src/datafusion/spice_udf_override.rs
```diff
@@ -0,0 +1,290 @@
...
+            }
+        }
+        _ => Ok(Transformed::no(plan)),
+    }
+}
+
+fn override_spice_udf_in_exprs(
+    input: &LogicalPlan,
+    exprs: Vec<Expr>,
+) -> DataFusionResult<(Vec<Expr>, bool)> {
```
phillipleblanc, line 116: Instead of returning a bool to indicate modified, could we reuse the Transformed::yes/no here?

PR #3926, opened 2024-12-18
`fuzzy_match`, `json_match`, `includes` scorer

Jeadie:
# 🗣 Description
Three new scorers:
 - Fuzzy match: Inclusive match (i.e. `actual in ideal || ideal in actual`), once the two terms have been normalised. 
 - Json match: Match string output as deserialised serde_json::Value. Also has some normalisation (e.g. 1.0 == 1). 
     - `"{\n   \"key\":\"value\"\n   }\n" == "{ \"key\": \"value\" }"`
     - `"{ \"key2\": \"foo\", \"key1\": \"bar\" }" == "{ \"key1\": \"bar\", \"key2\": \"foo\" }"`
 - Includes: True if the actual response includes the expected response. 
     - Actual: "No, the user ...", 
     - deal: "No"

## ⛓️‍💥 Breaking Change


crates/runtime/src/model/eval/scorer/fuzzy_match.rs
```diff
@@ -0,0 +1,299 @@
...
+            let normalized_i = Self::normalize(i);
+
+            if normalized_a.is_empty() || normalized_i.is_empty() {
+                normalized_a == normalized_i
+            } else {
+                normalized_a.contains(&normalized_i) || normalized_i.contains(&normalized_a)
+            }
+        });
+
+        // TODO: Should consider using F1 score instead of boolean accuracy.
```
phillipleblanc: Let's create an issue to track and link it here.

phillipleblanc (dismissed):

phillipleblanc (dismissed):

sgrebnov (dismissed):

phillipleblanc (approved):

PR #3940, opened 2024-12-20
Implement `GET /v1/namespaces/{namespace}/tables/{table}` API

### LINKED ISSUE #3917: Implement `HEAD /v1/iceberg/namespaces/{namespace}/tables/{table}` API
Adapted from: https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L1129

Check if a table exists within a given namespace. The response does not contain a body.

### LINKED ISSUE #3918: Implement `GET /v1/iceberg/namespaces/{namespace}/tables/{table}` API
Adapted from: https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L904

Loads the configuration and metadata for a table. The useful part for us is that part of the metadata contains the schema to use and the `location` parameter that will be used to differentiate this from an Iceberg table.

The `metadata.location` will return the `spice.ai/<dataset_id>` identifier that can be used by another Spice instance to load the dataset as a reference to the current Spice instance rather than as an Iceberg table.

The `metadata.properties.spiceai_endpoint` will return the endpoint that can be used in the consuming Spice instance as the value of `spiceai_endpoint` in the dataset params. This value will come from the runtime config `runtime.endpoint.flight`. If not set, a default value of `http://<hostname>:<flight_port>` will be used.

Example Response:
```yaml=
{
  "metadata": {
      "format-version": 2 # Required by the Iceberg spec
      "table-uuid": "C1056009-C78E-4C00-87E0-4347E8F2E603" # Required by the spec
      "location": "spice.ai/schema.my_foo_table", # Returns the identifier to the current dataset as a spice.ai reference
      "properties": {
        "spiceai_endpoint": "https://my_external_api.foocorp.com:50051"
      }
      "schemas": [ # Array because Iceberg supports schema evolution
        {
          "schema-id": 0,
          "type": "struct",
          "fields": [
            {
              "id": 0,
              "name": "foobar_col",
              "type": "string",
              "required": false
            }
          ]
        }
      ]
  }
}
```

The format for the schema is dictated by https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L2115

phillipleblanc:
# 🗣 Description

Implements the `GET /v1/namespaces/{namespace}/tables/{table}` API which loads the schema from a table and returns it in the Iceberg format.

I initially implemented the conversion of Arrow -> Iceberg schema manually, but I realized the `iceberg-rust` crate already does this, so I've switched it over to that implementation.

## 🔨 Related Issues

Closes #3917
Closes #3918


crates/runtime/src/http/v1/iceberg/tables.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    let Some(table) = datafusion.get_table(&table_reference).await else {
+        return status::StatusCode::NOT_FOUND.into_response();
+    };
+
+    let arrow_schema = table.schema();
+    let iceberg_schema = match Schema::try_from(arrow_schema.as_ref()) {
+        Ok(schema) => schema,
+        Err(e) => {
+            tracing::debug!("Error converting arrow schema to iceberg schema: {e}");
```
sgrebnov: Nit: adding table name to log might be useful for troubleshooting

sgrebnov (dismissed):

peasee (approved):

ewgenius (approved):

Issue #3953, opened 2024-12-22
Add Clickbench benchmark tests for SQLite accelerator


peasee:
Add Clickbench benchmark tests for SQLite accelerator using s3 as a source

PR #3958, opened 2024-12-23
feat: Source ClickBench queries from ClickBench

peasee:
# 🗣 Description

* Creates a script to source the ClickBench queries from the ClickBench repo, because the queries we have are not the same
* Adds overrides for SQLite to remove clickbench query 29, as SQLite does not support it.

## 🔨 Related Issues

* Part of #3953


crates/runtime/benches/queries/README.md
```diff
@@ -427,3 +427,16 @@ select  cast(amc as FLOAT)/cast(pmc as FLOAT) am_pm_ratio
...
+## ClickBench
+
+### Attribution Notice
+
+The [ClickBench benchmark queries](./clickbench) are used under the [Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International License (CC BY-NC-SA 4.0)](https://creativecommons.org/licenses/by-nc-sa/4.0/).
+
+- **Original Author(s):** [ClickBench Authors](https://github.com/ClickHouse/ClickBench)
+- **Source Repository:** [https://github.com/ClickBench/ClickBench](https://github.com/ClickHouse/ClickBench)
+- **License:** [CC BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/)
+- **Modifications:** This version use a subset of database sources, and makes the following modifications:
```
peasee:
```suggestion
- **Modifications:** This version uses a subset of database sources, and makes the following modifications:
```

peasee:
Need to update the script to make all the query column names wrapped in quotes `"`

phillipleblanc (approved):

Sevenannn:
Did you test if the changed queries would work for DuckDB & Arrow? I added modifications to the original Clickbench query so that it could properly work with parquet data and Postgres syntax requirements, (see changes here: https://github.com/spiceai/spiceai/pull/3389), since Spice SQL is essentially a Postgres dialect SQL.

peasee:
> Did you test if the changed queries would work for DuckDB & Arrow? I added modifications to the original Clickbench query so that it could properly work with parquet data and Postgres syntax requirements, (see changes here: #3389), since Spice SQL is essentially a Postgres dialect SQL.

Ideally, we can source the DuckDB queries from the ClickHouse benchmark and use those for DuckDB. However, those queries use mixed case column names that are unquoted which results in query errors because of ident normalization applied from our datafusion engine (and we cannot disable ident normalization).

PR #4002, opened 2024-12-26
Make Flight DoPut atomic and commit write only on successful stream completion

sgrebnov:
# 🗣 Description

Switch DoPut write mechanism to use `StreamingDataUpdate` which commits changes only at the successful end of stream preventing partial writes when DoPut stream fails or we terminate the stream due to max rows or timeout policy.

Tests include:
 - happy path (stream is successfully completed and data is written)
 - simulating transition error when client stream unexpectedly closed and no data is written

## ⛓️‍💥 Breaking Change


crates/runtime/src/flight/do_put.rs
```diff
@@ -16,32 +16,40 @@ limitations under the License.
...
-) -> Option<Arc<Sender<DataUpdate>>> {
-    let channel_map_read = channel_map.read().await;
-    if channel_map_read.contains_key(path) {
-        let channel = channel_map_read.get(path)?;
-        Some(Arc::clone(channel))
-    } else {
-        None
-    }
-}
+// async fn get_sender_channel(
```
sgrebnov: Investigating if this is still required and non-streaming update should be kept as-is specifically for this case.

phillipleblanc (approved):

Issue #3872, opened 2024-12-16
Enhancement: Iceberg Catalog Connector Beta


phillipleblanc:
## Goal-State/What/Result

A new Iceberg Catalog Connector is implemented in Spice OSS that loads Iceberg/Spice.ai tables from the specified Iceberg REST Catalog API service

```yaml
catalogs:
  - from: iceberg:https://my_iceberg_host.com/v1/namespaces # Path to the `/v1/{prefix}/namespaces` API
    name: icey
    include:
      - "*.my_table"
    params:
      iceberg_bearer_token: ${ secrets:ICEBERG_BEARER_TOKEN }
```

The connector will recursively load all of the tables from all of the namespaces in the Iceberg catalog API.

To only load tables from a sub-namespace, specify the namespace identifier:

```yaml
catalogs:
  - from: iceberg:https://my_iceberg_host.com/v1/namespaces/top_namespace%1Fsecond_level_namespace # Only load tables under the `top_namespace.second_level_namespace` sub-namespace.
    name: icey
    include:
      - "*.my_table"
    params:
      iceberg_bearer_token: ${ secrets:ICEBERG_BEARER_TOKEN }
```

> [!NOTE]
> The `0x1F` as a separator is a bit ugly, but its what the official Iceberg spec requires - and I think its better to stick to that than trying to support a nicer syntax here. You can copy-paste this URL and it will work.

## Why/Purpose

Iceberg is an industry standard catalog API and table format.

## By When

**Issue/Spec written and reviewed:** Dec 16, 2024
**Done-Done:** To be scheduled

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Samples/Quickstarts/Tutorials
- [ ] Release Notes prepared

## The Algorithm

- [ ] Every requirement questioned?
- [ ] Delete (Scope) any part you can.
- [ ] Simplify.
- [ ] Break down into smaller iterations/milestones.
- [ ] Opportunities for automation.

#### Merging namespaces

DataFusion doesn't support arbitrarily nested namespaces, limiting us to a 3-level hierarchy of `catalog.schema.table`.

[...]

PR #4003, opened 2024-12-26
Create a `CatalogConnector` abstraction

phillipleblanc:
# 🗣 Description

This large refactoring PR adds the concept of a `CatalogConnector` that is separate from a `DataConnector` and moves over two existing CatalogConnectors (UnityCatalog/Databricks) over to this new abstraction.

This is work necessary to prepare for #3872 to ensure that the right catalog abstractions are in place.

Additionally, we need catalogs to be able to periodically refresh their included tables to reflect changes from the upstream catalog provider. This PR defines the `RefreshableCatalogProvider` trait with a single method `refresh` that all catalog providers supported in Spice must support. Currently this is hard-coded to refresh every 60 seconds.

## 🔨 Related Issues

Ground work preparing for #3872

## 📄 Documentation Updates

N/A

## 🤔 Concerns

Hard-coding the refresh interval for now

### DIFF
--- a/crates/data_components/src/unity_catalog/provider.rs
+++ b/crates/data_components/src/unity_catalog/provider.rs
@@ -23,15 +23,19 @@ use datafusion::{
 };
 use globset::GlobSet;
 use snafu::prelude::*;
-use std::{any::Any, collections::HashMap, sync::Arc};
+use std::{
+    any::Any,
+    collections::HashMap,
+    sync::{Arc, RwLock},
+};
 
-use crate::Read;
+use crate::{Read, RefreshableCatalogProvider};
 
 use super::{CatalogId, Result, UCSchema, UCTable, UnityCatalog};
 
 #[derive(Debug)]
 pub struct UnityCatalogProvider {
-    schemas: HashMap<String, Arc<dyn SchemaProvider>>,
+    schemas: HashMap<String, Arc<UnityCatalogSchemaProvider>>,
 }
 
 impl UnityCatalogProvider {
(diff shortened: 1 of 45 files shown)


crates/data_components/src/unity_catalog/provider.rs
```diff
@@ -90,13 +91,41 @@ impl CatalogProvider for UnityCatalogProvider {
...
+            .get(name)
+            .cloned()
+            .map(|s| s as Arc<dyn SchemaProvider>)
+    }
+}
+
+#[async_trait]
+impl RefreshableCatalogProvider for UnityCatalogProvider {
+    async fn refresh(&self) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+        let refresh_futures = self.schemas.values().map(|schema| schema.refresh());
```
sgrebnov: Nit: we might want to add limit on num of schemas refreshed in parallel (if there could be a lot of parallel requests)

sgrebnov (dismissed):

crates/runtime/src/catalogconnector/mod.rs
```diff
@@ -0,0 +1,243 @@
...
+    fn schema_names(&self) -> Vec<String> {
+        self.inner.schema_names()
+    }
+
+    fn schema(&self, name: &str) -> Option<Arc<dyn datafusion::catalog::SchemaProvider>> {
+        self.inner.schema(name)
+    }
+}
+
+impl Drop for RefreshingCatalogProvider {
```
sgrebnov, line 237: 👍

PR #4010, opened 2024-12-27
Allow Flight DoPut only for write api-keys

sgrebnov:
# 🗣 Description

PR gives ability to define auth api keys as read-only and read-write and updates Flight middleware to propagate auth user information based on api key and allow Flight DoPut when principal information exists (authenticated) and has write permissions. To propagate auth information `AuthRequestContext` and `AuthPrincipal` traits were introduced and `RequestContext` struct was extended to implement `AuthRequestContext` trait and logic to get/set auth principal.

```yaml
runtime:
  auth:
    api_key:
      keys:
        - 123456       # Default/implicit Read-Only
        - abcdef:ro    # Read-only
        - 098765:rw  # Read-write
```

## ⛓️‍💥 Breaking Change

Flight DoPut now requires auth enabled and write access key. Previously Flight DoPut allowed w/o auth enabled and worked for spiceai and flightsql connectors only. Note - Flight DoPut was not documented and we don't have recipes for it.


crates/runtime-auth/src/layer/flight.rs
```diff
@@ -113,6 +113,8 @@ pub struct BasicAuthMiddleware<S> {
     auth_verifier: Option<Arc<dyn FlightBasicAuth + Send + Sync>>,
 }
 
+//type BoxFuture<'a, T> = Pin<Box<dyn std::future::Future<Output = T> + Send + 'a>>;
+
```
phillipleblanc:
```suggestion
```

crates/runtime/src/flight/do_put.rs
```diff
@@ -33,15 +34,26 @@ use async_stream::stream;
...
     flight_svc: &Service,
     request: Request<Streaming<FlightData>>,
 ) -> Result<Response<<Service as FlightService>::DoPutStream>, Status> {
+    match RequestContext::current(crate::request::AsyncMarker::new().await).auth_principal() {
+            Some(principal) => {
+                if !principal.groups().iter().any(|group| *group == "write" || *group == "read_write") {
+                    return Err(Status::permission_denied("Write access denied. Verify that authentication key used has write access and try again."));
+                }
+            },
+            None => return Err(Status::unauthenticated("Flight DoPut requires authentication enabled.\nFor auth details, visit https://docs.spiceai.org/api/auth")),
```
phillipleblanc:
```suggestion
            None => return Err(Status::unauthenticated("Flight DoPut requires authentication.\nFor auth details, visit https://docs.spiceai.org/api/auth")),
```

phillipleblanc (dismissed):
Very clean, thanks @sgrebnov. Just a couple minor comments.

phillipleblanc (approved):

Issue #3994, opened 2024-12-24
Dremio query error: ROW/RANGE not allowed with RANK, DENSE_RANK or ROW_NUMBER functions


Sevenannn:
Affect TPCDS query 36 44 47 49 57 67 86

PR #4012, opened 2024-12-27
Implement window_func_support_window_frame from DremioDialect

Sevenannn:
NOTE: DO NOT merge this before https://github.com/spiceai/datafusion is merged and patch is updated

# 🗣 Description

Implement window_func_support_window_frame from DremioDialect to return false when the window frame indicates no window frame specified in original SQL Query, since dremio RANK, DENSE_RANK or ROW_NUMBER window functions doesn't support window frame. Also implement it for SUM window function, as the datafusion `preceding None following None` over sum window function is syntatically equal to. dremio sum window function without window frame

## 🔨 Related Issues

- https://github.com/spiceai/spiceai/issues/3994

## ⛓️‍💥 Breaking Change


crates/runtime/src/dataconnector/dremio.rs
```diff
@@ -82,6 +82,13 @@ impl Dialect for DremioDialect {
     ) -> datafusion::sql::sqlparser::ast::DataType {
         datafusion::sql::sqlparser::ast::DataType::Timestamp(None, TimezoneInfo::None)
     }
+
+    fn window_func_support_window_frame(&self, func_name: &str) -> bool {
+        if matches!(func_name, "rank" | "row_number" | "dense_rank") {
```
Jeadie:
```suggestion
        !matches!(func_name, "rank" | "row_number" | "dense_rank")
```
Straight return?

phillipleblanc (approved):

PR #4084, opened 2024-12-31
Fix OpenAI Models Integration tests

phillipleblanc:
# 🗣 Description

Makes the testing for OpenAI integration model tests more consistent by adding a system prompt to fix the table quoting, and also retrieve the task history input as a JSON value and pretty printing it after sorting the keys.


lukekim (dismissed):

crates/runtime/tests/models/openai.rs
```diff
@@ -67,6 +67,7 @@ mod nsql {
                     primary_keys: None,
                     chunking: None,
                 }];
+            //taxi_trips_with_embeddings.metadata.insert("instructions".to_string(), "Do not put double quotes around `spice.public.taxi_trips`.".to_string().into());
```
phillipleblanc:
```suggestion
```

PR #4116, opened 2025-01-02
Set unified max encoding and decoding message size for all flight client configurations across runtime

ewgenius:
Use a single encode/decode message size of 100 MiB for all flight client and server configurations

### DIFF
--- a/crates/data_components/src/flightsql.rs
+++ b/crates/data_components/src/flightsql.rs
@@ -21,13 +21,16 @@ use arrow::{
 use async_stream::stream;
 use async_trait::async_trait;
 use datafusion_table_providers::sql::sql_provider_datafusion::expr;
-use flight_client::tls::new_tls_flight_channel;
+use flight_client::{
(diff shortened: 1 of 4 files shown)


crates/data_components/src/flightsql.rs
```diff
@@ -516,6 +524,8 @@ fn query_to_stream(
     mut client: FlightSqlServiceClient<Channel>,
     sql: &str,
 ) -> impl Stream<Item = DataFusionResult<RecordBatch>> {
+    println!("[flightsq] query_to_stream: {}", sql);
```
sgrebnov:
```suggestion
```

lukekim (dismissed):

crates/data_components/src/flightsql.rs
```diff
@@ -516,6 +524,7 @@ fn query_to_stream(
     mut client: FlightSqlServiceClient<Channel>,
     sql: &str,
 ) -> impl Stream<Item = DataFusionResult<RecordBatch>> {
+
```
ewgenius:
```suggestion
```

phillipleblanc (approved):

PR #4117, opened 2025-01-02
Enforce Flight DoPut policies: Rate Limiting, Read Timeout, and Max Records per Batch

sgrebnov:
# 🗣 Description

Add the following  policies for Flight DoPut (write):
 - rate limiting: max 100 write requests per minute
 - max 30s timeout to read next record batch to write
 - max allowed records per record batch: 32_768  (4x DataFusion default records per batch `datafusion.execution.batch_size` is 8192 - default batch size while creating new batches)

Note: Ratelimiting is based on own Tower Middleware as [existing tower rate limiting layer](https://docs.rs/tower/latest/tower/limit/rate/struct.RateLimitLayer.html) and [tower_governon](https://crates.io/crates/tower_governor) don't support configuring rate limiting to be applied to specific path only (DoPut only). Creating wrapper around RateLimitLayer is also not possible as [it will panic when called if poll_ready is based on wrapped layer](https://github.com/tower-rs/tower/blob/master/tower/src/limit/rate/service.rs#L116). Used as a reference implementations instead.

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

1. Can enforced limits potentially impact telemetry reporting, should be increased/updated? 🤔 
1. We might also add policy on max allowed record batch size to write, for example InfluxDB
https://docs.influxdata.com/influxdb/cloud/account-management/limits/#global-limits

Write request limits:
50 MB maximum HTTP request batch size (compressed or uncompressed–defined in the Content-Encoding header)
250 MB maximum HTTP request batch size after decompression


crates/runtime/Cargo.toml
```diff
@@ -98,6 +98,7 @@ prost = { version = "0.13.1", default-features = false, features = [
     "prost-derive",
 ] }
 rand = "0.8.5"
+ratelimit = "0.10.0"
```
phillipleblanc: Have you considered https://docs.rs/governor/0.8.0/governor/ instead?

crates/runtime/src/flight/do_put.rs
```diff
@@ -168,9 +185,28 @@ pub(crate) async fn handle(
...
 
-    let timed_stream = TimedStream::new(response_stream, move || start);
+async fn handle_record_batch(
+    batch: RecordBatch,
+    batch_tx: &Sender<Result<RecordBatch, DataFusionError>>,
+) -> Result<PutResult, Status> {
+    tracing::trace!("Received batch with {} rows", batch.num_rows());
 
-    Ok(Response::new(Box::pin(timed_stream)))
+    if batch.num_rows() > 32_768 {
```
phillipleblanc, line 210: Add why this number as a comment

crates/runtime/src/rate_limits.rs
```diff
@@ -0,0 +1,55 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use std::time::Duration;
+
+pub struct RateLimit {
```
phillipleblanc: Move this to the `flight` module

crates/runtime/Cargo.toml
sgrebnov: Thank you @phillipleblanc - despite both crates are very popular I found `governon` crate is a better choice so updated PR to switch from `ratelimit` to `governon`. https://github.com/spiceai/spiceai/pull/4117/commits/1b100cb239674674482efbfe0cb9cf4d06dc067b

phillipleblanc (dismissed):

crates/runtime/src/rate_limits.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+}
+
+impl Default for RateLimits {
+    fn default() -> Self {
+        Self {
+            // Allow 100 Flight DoPut requests every 60 seconds by default
+            #[allow(clippy::expect_used)]
+            flight_write_limit: Quota::per_minute(
+                NonZeroU32::new(100)
```
phillipleblanc: Use the `unreachable!` macro instead of `.expect`

sgrebnov: Updated - thank you

crates/runtime/src/flight/do_put.rs
sgrebnov: Added the following comment, thank you

```rust
// 32,768 is four times the default batch size in DataFusion (`datafusion.execution.batch_size`), which defaults to 8,192.
if batch.num_rows() > 32_768 {
    return Err(Status::invalid_argument(format!(
        "The provided batch contains too many rows. Maximum allowed: {allowed}, received: {received}.",
        allowed = 32_768,
        received = batch.num_rows()
    )));
}
```

crates/runtime/src/rate_limits.rs
sgrebnov: My thinking is that we will be able to extend the `RateLimits` instance with HTTP and other non-flight limits in the future, so I propose keeping `RateLimits` in `runtime` instead of making it flight only and moving it to the `flight` module.

crates/runtime/src/rate_limits.rs
```diff
@@ -0,0 +1,47 @@
...
+        self.flight_write_limit = rate_limit;
+        self
+    }
+}
+
+impl Default for RateLimits {
+    fn default() -> Self {
+        Self {
+            // Allow 100 Flight DoPut requests every 60 seconds by default
+            #[allow(clippy::expect_used)]
```
phillipleblanc:
```suggestion
```

crates/runtime/src/rate_limits.rs
phillipleblanc: I disagree, this is basically just a small wrapper around the `governor` crate that is currently specific to flight. This is an example of premature optimization combined with adding unnecessary abstractions (i.e. complexity). I'd either move this as a module under `flight` or just remove it entirely and use the governor directly in the middleware.

crates/runtime/src/rate_limits.rs
sgrebnov: Thank you

crates/runtime/src/rate_limits.rs
sgrebnov: @phillipleblanc  - I see, thank you! Moved `RateLimits` to `flight` module, this required to modify accessibility for the module. The main reason of adding RateLimits was ability to override default behavior for tests so they don't need to wait for the full 60s before limit is reset (plus allow less requests to reduce size of snapshot file / less records), I think we can completely remove the RateLimits configuration and allow tests to run longer or remove part that verifies that request can successfully complete again after some time as we test `governon` crate behavior here which I don't think we should. Should I remove RateLimits configuration completely?

phillipleblanc: I think its fine to keep it.

phillipleblanc (approved):

Issue #4114, opened 2025-01-02
Run benchmarks on the testoperator


peasee:
## Goal-State/What/Result

Benchmark tests should be run on the `testoperator` tool, and continue to report benchmark results

## Why/Purpose

To sunset the benchmark runner, in favor of the more robust `testoperator` test framework

## By When

**Issue/Spec written and reviewed:** _Target Date_
**Done-Done:** _Target Date_

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

* Metrics from the testoperator tests are optionally written via OTel to the metrics receiver
* Benchmarks are migrated from the benchmarks runner to individual spicepods for each benchmark
* A new append-mode benchmark command is added, to support append-mode data loading tests

Depends on `spiced` supporting write datasets.

## How/Plan

1. The test operator optionally writes benchmark metrics via OTel to the metrics receiver

2. Benchmarks are migrated from the benchmarks runner to individual spicepods

All current benchmark tests in the benchmarks runner are migrated to individual spicepods instead, including tests with complex behavior like append-mode tests. Append-mode tests receive an additional command for their specialty data loading logic, `testoperator run append ...`.

3. A trait for append mode data loading is created and implemented for append-mode tests

```rust
pub trait AppendableDatasource {
    fn prepare(&mut self) -> Result<()>;
    fn load(&mut self) -> Result<JoinHandle<...>>;
}
```

Anything that implements an `AppendableDatasource` can be used as a data source for an append mode test. As part of scope for this task though, it is implemented only for a `FileAppendableDatasource` for the file connector. This replicates existing behavior on the benchmarks runner.

peasee:
Spark has some failing TPCDS queries that need exclusions:

https://github.com/spiceai/spiceai/actions/runs/14188008526/job/39746694490

phillipleblanc:
Close this issue and open a new one for metrics

Issue #4102, opened 2025-01-02
Run an append-mode TPCH DuckDB acceleration test in benchmarks


peasee:
Run an append-mode TPCH DuckDB acceleration test in benchmarks.

peasee:
Something doesn't seem right. Append keeps re-inserting only a specific timestamp:

```
D select l_created_at, count(*) from lineitem group by l_created_at;
┌─────────────────────────┬──────────────┐
│      l_created_at       │ count_star() │
│        timestamp        │    int64     │
├─────────────────────────┼──────────────┤
│ 2025-01-02 13:17:10.367 │       600572 │
│ 2025-01-02 13:18:11.792 │      6593367 │
└─────────────────────────┴──────────────┘
```

While the source data has a distributed number of rows for each timestamp as it's loaded:

```
D select l_created_at, count(*) from lineitem group by l_created_at;
┌─────────────────────────┬──────────────┐
│      l_created_at       │ count_star() │
│        timestamp        │    int64     │
├─────────────────────────┼──────────────┤
│ 2025-01-02 13:19:14.002 │       600124 │
│ 2025-01-02 13:17:10.367 │       600572 │
│ 2025-01-02 13:20:16.29  │       599647 │
│ 2025-01-02 13:18:11.792 │       599397 │
│ 2025-01-02 13:25:27.901 │       599715 │
│ 2025-01-02 13:23:23.15  │       599301 │
│ 2025-01-02 13:24:25.554 │       600504 │
│ 2025-01-02 13:22:20.737 │       601365 │
│ 2025-01-02 13:21:18.633 │       599931 │
│ 2025-01-02 13:26:29.978 │       600659 │
├─────────────────────────┴──────────────┤
│ 10 rows                      2 columns │
└────────────────────────────────────────┘
```

PR #4120, opened 2025-01-03
feat: Add the file connector as an appendable benchmark connector

### LINKED ISSUE #4101: Run a delayed Parquet source load in benchmarks
Using Parquet, load a source database over a period of time. This will support using Parquet as a file connector to test append-mode accelerators.

peasee:
# 🗣 Description

* Adds support to use the `file` connector as an appendable connector source - loading data into the connector (parquet file) over a period of time. Loads the connector source over 30 minutes at scale factor 10 with a static scale factor/load count/load interval.

Uses DuckDB as the generator for simplicity - it has the functions we need to easily generate datasets, add a timestamp column, and export the tables as parquet.

* Adds support for using `duckdb` in an append-mode accelerator benchmark. Validates that the accelerator finishes loading within 1 hour, to test the performance of the accelerator over time.

The benchmark can be run like `bench --bench -a duckdb -m file -b tpch --append-mode true`.

## 🔨 Related Issues

* Closes #4101 

## 🤔 Concerns

* I've avoided making sweeping changes to the structure of the benchmark runners to support these benchmarks. We should focus effort instead on superseding them with #4114 in future.
* Currently does not run as part of the benchmark workflow - will add in future PRs as part of #4102

### DIFF
--- a/crates/runtime/benches/bench.rs
+++ b/crates/runtime/benches/bench.rs
@@ -27,6 +27,7 @@ limitations under the License.
 
 use std::panic;
 use std::sync::Arc;
+use std::time::Duration;
 
 #[cfg(feature = "postgres")]
 use crate::bench_postgres::get_postgres_params;
@@ -40,8 +41,9 @@ use futures::TryStreamExt;
 use results::BenchmarkResultsBuilder;
 use runtime::request::{Protocol, RequestContext, UserAgent};
 use runtime::{dataupdate::DataUpdate, Runtime};
-use spicepod::component::dataset::acceleration::{self, Acceleration, Mode};
+use spicepod::component::dataset::acceleration::{self, Acceleration, Mode, RefreshMode};
 use spicepod::component::params::Params;
+use tract_core::num_traits::ToPrimitive;
 
 mod results;
 mod setup;
@@ -88,6 +90,9 @@ struct BenchArgs {
     /// Set the acceleration mode for accelerator
     #[arg(short, long)]
     mode: Option<String>,
+
+    #[arg(long)]
+    append_mode: Option<bool>,
 
     /// Set the benchmark to run: TPCH / TPCDS
     #[arg(short, long, default_value = "tpch")]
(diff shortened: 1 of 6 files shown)


crates/runtime/benches/bench_duckdb/mod.rs
```diff
@@ -280,3 +285,105 @@ fn get_tpcds_test_queries() -> Vec<(&'static str, &'static str)> {
...
+
+    tokio::spawn(async move {
+        let dest_db_file = format!("./{bench_name}.db");
+        let dest_conn = Connection::open(&dest_db_file).map_err(|e| e.to_string())?;
+
+        for i in 0..load_count {
+            println!("Loading data for {bench_name} benchmark suite, iteration {i}");
+            match bench_name.as_str() {
+                "tpch" => {
+                    let batch_sql = if i == 0 {
```
peasee, line 307: I'll improve the generation of this SQL in follow-up PR(s).

crates/runtime/benches/bench.rs
```diff
@@ -179,19 +182,24 @@ async fn bench_main() -> Result<(), String> {
                 _ => return Err(format!("Invalid mode parameter for {accelerator} accelerator")),
             };
 
-            let acceleration = create_acceleration(accelerator, mode, args.bench_name.as_ref());
+            let append_mode = args.append_mode.unwrap_or(false);
```
phillipleblanc: It would be clearer if this was an enum, instead of a bool. Then in the match statement it can match on `RefreshMode::Append, "tpch"` which is much clearer. Probably can reuse the existing enum

crates/runtime/benches/bench.rs
```diff
@@ -314,18 +396,35 @@ async fn run_accelerator_bench(
     Ok(())
 }
 
-fn create_acceleration(engine: &str, mode: acceleration::Mode, bench_name: &str) -> Acceleration {
-    let params: Option<Params> = match engine {
+fn create_acceleration(
+    engine: &str,
+    mode: acceleration::Mode,
+    bench_name: &str,
+    append: bool,
```
phillipleblanc: Use the enum here too

crates/runtime/benches/bench_duckdb/mod.rs
```diff
@@ -57,6 +61,7 @@ pub(crate) async fn run(
     Ok(())
 }
 
+#[allow(clippy::too_many_lines)]
```
phillipleblanc: This wasnt here before?

peasee: Yeah, this can get removed. I had it temporarily from when I was still using duckdb as the connector source.

phillipleblanc (approved):

PR #4146, opened 2025-01-05
Promote Spicepod version to V1

phillipleblanc:
# 🗣 Description

Updates the default version of the Spicepod from `v1beta1` to `v1` in preparation for our 1.0 release.

The previous version `v1beta` is still accepted, but will no longer be referenced in our docs or generated by default.

## 🔨 Related Issues

N/A

## ⛓️‍💥 Breaking Change

It isn't technically a breaking change - but it is a significant one.

## 📄 Documentation Updates

https://github.com/spiceai/docs/pull/715
https://github.com/spiceai/cookbook/pull/20

### DIFF
--- a/bin/spice/pkg/spec/spicepod_test.go
+++ b/bin/spice/pkg/spec/spicepod_test.go
@@ -57,7 +57,7 @@ func convertMap(i interface{}) interface{} {
 
 func TestSpicepodSpec_UnmarshalYAML_KnownFields(t *testing.T) {
 	yamlText := `
-version: v1beta1
+version: v1
 kind: Spicepod
 name: test-pod
 datasets:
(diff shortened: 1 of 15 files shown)


bin/spice/pkg/spec/spicepod_test.go
```diff
@@ -127,8 +127,8 @@ nested_unknown:
 	}
 
 	// Verify known fields
-	if spicePod.Version != "v1beta1" {
-		t.Errorf("Expected version v1beta1, got %s", spicePod.Version)
+	if spicePod.Version != "v1" {
```
ewgenius, line 130: I think we should keep it backwards compatible and allow v1beta1

bin/spice/pkg/spec/spicepod_test.go
```diff
@@ -204,8 +204,8 @@ nested_unknown:
 	result = convertMap(result).(map[string]interface{})
 
 	// Check known fields
-	if result["version"] != "v1beta1" {
-		t.Errorf("Expected version v1beta1, got %v", result["version"])
+	if result["version"] != "v1" {
```
ewgenius, line 207: same here

ewgenius (approved):

bin/spice/pkg/spec/spicepod_test.go
phillipleblanc: it does, this is for the test where we write `v1`

bin/spice/pkg/spec/spicepod_test.go
phillipleblanc: it does, this is for the test where we write v1

sgrebnov (approved):

PR #4154, opened 2025-01-05
Fix `platform_option` variable in `build_and_release.yml`.

Jeadie:
- Fix `workflow_dispatch` options to a single entry in the matrix.
- Working in: https://github.com/spiceai/spiceai/actions/runs/12623912033


.github/workflows/build_and_release.yml
```diff
@@ -9,17 +9,16 @@ on:
 
   workflow_dispatch:
     inputs:
-      desired_os:
-        description: "Which OS do you want to run on? (Default is ALL)"
+      desired_option:
```
lukekim: "desired" sounds weird. Maybe we should just call it `platform`?

peasee (approved):

PR #4156, opened 2025-01-06
Fix model name in `xAI`; add to `runtime/llms` tests

Jeadie:
# 🗣 Description
 - xAI was not using the model name from the `from: xai:model_name`. 
 - Add xAI model type to `runtime/llms` tests.
 
 **Note**: Need to add env variable `SPICE_XAI_API_KEY` to GH actions CI.

### DIFF
--- a/crates/llms/src/embeddings/candle/util.rs
+++ b/crates/llms/src/embeddings/candle/util.rs
@@ -192,14 +192,15 @@ pub(crate) fn max_seq_length_from_st_config(
 ///
 /// ```rust
 /// use std::collections::HashMap;
-/// use std::path::Path;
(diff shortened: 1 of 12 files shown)


crates/llms/src/embeddings/candle/util.rs
```diff
@@ -192,14 +192,15 @@ pub(crate) fn max_seq_length_from_st_config(
 ///
 /// ```rust
 /// use std::collections::HashMap;
-/// use std::path::Path;
```
Jeadie, line 195: Fix doc tests.

phillipleblanc (approved):

PR #4157, opened 2025-01-06
Add `test-framework` crate to contain all common benchmark, E2E, integration testing logic.

phillipleblanc:
# 🗣 Description

Adds a new `test-framework` crate which will house all of our common testing logic that is currently spread across multiple test crates, with no easy way to share logic between them.

There are two main pieces that I've moved here, in preparation for adding the throughput tests needed for most of our connectors to go into 1.0 Stable.
1) Test utilities for starting/controlling a `spiced` instance so that we can execute the tests against real binaries instead of just portions of the binaries built directly into the test executable.
2) The TPCH, TPCDS and Clickbench test queries are moved into here, with common logic for applying overrides depending on the test context. This was needed for the throughput tests, but the queries were scattered across each benchmark test module - making it difficult to reuse them.

I've also started a new tool called `testoperator` where I envision the new throughput tests actually being run from. It should intentionally be kept lightweight, and mostly just a driver of the `test-framework`. It currently starts a spice instance, does a test query and then shuts down.

I've approached this such that any type of testing we might want to do (i.e. AI evals, models, etc) could use this new framework.

## 🔨 Related Issues

No direct issues are related to this, but this is foundational for all of the tests needed going into Stable.

## 📄 Documentation Updates

N/A

## 🤔 Concerns

I haven't moved all of the queries over to the test-framework yet - some of the benchmarks do some complex overriding that I wasn't sure how to handle elegantly. Once we get some more confidence in the test-framework, I expect we will move more items into it.


peasee (dismissed):
Nice, I like this.

peasee (approved):

PR #4158, opened 2025-01-06
Ensure non-nullity of primary keys in `MemTable`; check validity of initial data.

Jeadie:
# 🗣 Description
- Ensure primary keys are not null (as per [docs]( https://github.com/apache/datafusion/blob/3f4297f505887accb5e5c2afc149a03de6631f15/datafusion/common/src/functional_dependencies.rs#L31)).
- For `MemTable` ensure the initial `RecordBatch`s satisfy the primary key uniqueness constraints specified (currently these are only checked during insert).

### DIFF
--- a/crates/data_components/src/arrow/write.rs
+++ b/crates/data_components/src/arrow/write.rs
@@ -98,10 +98,44 @@ impl MemTable {
         })
     }
 
-    #[must_use]
-    pub fn with_constraints(mut self, constraints: Constraints) -> Self {
+    pub async fn try_with_constraints(mut self, constraints: Constraints) -> Result<Self> {
+        self.ensure_batches_satisfy_constraints(&constraints)
+            .await?;
         self.constraints = constraints;
-        self
+        Ok(self)
+    }
+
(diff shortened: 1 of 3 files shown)


crates/data_components/src/arrow/write.rs
```diff
@@ -306,7 +438,12 @@ fn filter_existing(
 
         let mut keep_row_builder = BooleanBuilder::with_capacity(keys.len());
         for k in keys {
-            keep_row_builder.append_value(!overwriting_primary_keys.contains(&k));
+            // We check non-nullity of primary keys at insertion, so we can safely assume non-null here.
+            if let Some(k) = k {
+                keep_row_builder.append_value(!overwriting_primary_keys.contains(&k));
+            } else {
+                tracing::trace!("Primary keys in `MemSink` record batch contain null. This should be impossible!");
```
phillipleblanc: Should we panic here with `unreachable!` if this happens? It sounds like it would be a programmer bug that we should catch during development.

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #4171, opened 2025-01-06
Upgrade dependencies

phillipleblanc:
# 🗣 Description

## ⛓️‍💥 Breaking Change


Cargo.lock
phillipleblanc, line 1: Hold for after release

PR #4172, opened 2025-01-06
Introduce `supports_transactional_write` DataConnector flag for `read_write` mode

sgrebnov:
# 🗣 Description

Introduce `supports_transactional_write` `DataConnector` flag indicating that concurrent writes for `read_write` table provider are safe. Motivation: not all DataConnectors can fully support concurrent/transactional writes that can lead to unpredicted state and unwanted behavior, PR adds warning message if write is configured for a DataConnector w/o explicitly marked transactional write support. 

Marked `spice.ai` and `postgres` as supporting transactional writes. List of other DataConnectors implementing `read_write_provider`:
 - `dremio`: requires additional review before setting `supports_transactional_write`
 - `memory`: based on implementation review write is performed after collecting all records and based on special rw lock which guarantees concurrent writes don't conflict (next write awaits for in-progress one to be fully completed). `supports_transactional_write` will be set via  separate PR after running test to confirm this behavior.

Alternative approach considered was detecting supports_transactional_write capability as part of Flight DoPut handler and rejecting write if DataConnector does  not have `supports_transactional_write` but found the following limitations with such approach:
 - For some DataConnectors (for example Flight) support is based on underlaying server implementation (Flight Server / DoPut handler implementation) so not always could be set for a DataConnector thus warning instead of full rejection is preferred approach IMO.
 - Detecting DataConnector source based on  dyn TableProvider is not always possible for Federated Tables and requires using dataset definition to reconstruct DataConnector which brings additional overhead so it is proposed to use warning message during writable Table creation instead of adding verification into Flight DoPut. 

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/runtime/src/dataconnector.rs
+++ b/crates/runtime/src/dataconnector.rs
@@ -400,6 +400,10 @@ pub trait DataConnector: Send + Sync {
         None
     }
 
+    fn supports_transactional_write(&self) -> bool {
+        false
+    }
+
     async fn metadata_provider(
         &self,
         _dataset: &Dataset,
--- a/crates/runtime/src/dataconnector/postgres.rs
+++ b/crates/runtime/src/dataconnector/postgres.rs
@@ -142,6 +142,10 @@ impl DataConnector for Postgres {
         self
     }
 
+    fn supports_transactional_write(&self) -> bool {
+        true
+    }
+
     #[cfg(feature = "postgres-write")]
     async fn read_write_provider(
         &self,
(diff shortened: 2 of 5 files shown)


crates/runtime/src/dataconnector.rs
```diff
@@ -400,6 +400,10 @@ pub trait DataConnector: Send + Sync {
         None
     }
 
+    fn supports_transactional_write(&self) -> bool {
```
phillipleblanc, line 403: I don't think we should add this flag, we should just not allow data connectors to become read/write if they don't support this behavior.

PR #4175, opened 2025-01-06
Update spice cli to use `GH_TOKEN` or `GITHUB_TOKEN` env variables when calling releases api

### LINKED ISSUE #4086: Fix GitHub rate limiting in E2E Release Installation CI workflow
Unauthorized GH calls are restricted to 60 per hour max. For some reason MacOS runners always have more API calls used, and probably other requests are coming from this ip in other jobs

Image

ewgenius:
closes #4086 

When checking for latest available release, Spice CLI will use `GH_TOKEN` or `GITHUB_TOKEN` (in the order of precedence) to authenticate GitHub requests. That is needed to fix E2E Test Release Installation which is gets rate limited.

### DIFF
--- a/bin/spice/pkg/github/github.go
+++ b/bin/spice/pkg/github/github.go
@@ -28,8 +28,9 @@ import (
 )
 
 type GitHubClient struct {
-	Owner string
-	Repo  string
+	Owner       string
+	Repo        string
+	accessToken string
 }
 
 func NewGitHubClientFromPath(path string) (*GitHubClient, error) {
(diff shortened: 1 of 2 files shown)


bin/spice/pkg/github/github.go
```diff
@@ -47,8 +48,9 @@ func NewGitHubClientFromPath(path string) (*GitHubClient, error) {
 
 func NewGitHubClient(owner string, repo string) *GitHubClient {
 	return &GitHubClient{
-		Owner: owner,
-		Repo:  repo,
+		Owner:       owner,
+		Repo:        repo,
+		accessToken: os.Getenv("SPICE_CLI_GITHUB_TOKEN"),
```
lukekim: Let's just call this `SPICE_GITHUB_TOKEN`. We will also need to support disabling this.

ewgenius: reworked to use GH_TOKEN or GITHUB_TOKEN as most commonly used

phillipleblanc (approved):

Issue #4124, opened 2025-01-03
Run a TPCH throughput and load test on the Arrow accelerator


peasee:
Run a TPCH throughput and load test on the Arrow accelerator

PR #4214, opened 2025-01-08
feat: Add throughput test example

peasee:
# 🗣 Description

* Adds an example throughput test for the file connector using the TPCH dataset
* Adds a `README.md` for running the test operator
* Adds calculation of the [Throughput Metric](https://github.com/spiceai/spiceai/blob/trunk/docs/criteria/definitions.md#throughput-metric)
* Updates throughput test workers to return a `BTreeMap` of query times by query, to support calculating 99%/1% query times in future updates

## 🔨 Related Issues

* Part of #4124 

## 🤔 Concerns

* I've placed the example spicepod in `benchmarks/`. We should consider if this is the location we want to store all of our test spicepods, but the commitment cost is low anyway because the `testoperator` can run with any defined spicepod path.


benchmarks/file_tpch.yaml
phillipleblanc, line 1: Let's put this at `test/spicepods/file_tpch.yaml`

phillipleblanc (approved):

PR #4227, opened 2025-01-09
feat: Add test framework metrics collection

peasee:
# 🗣 Description

* Refactors metrics and statistics calculation out of `ThroughputTest` into a new `StatisticsCollector` trait
* Adds a new `MetricsCollector` trait to build metrics over query sets from tests
* Adds an `ExtendedMetrics` trait for queries to optionally include additional metric data than what is calculated by default. For example, for throughput tests it might include the proportion of queries that take longer than the baseline 99th percentile, memory usage, etc.
* Implements `MetricsCollector<NoExtendedMetrics>` for `ThroughputTest<Completed>` as an example of a metrics collection
* Displays metrics using `DataFrame::show()` at the end of a test run for nicer metrics output to console.

Future PRs will submit these metrics to Grafana/writing to a data store.

## 🔨 Related Issues

* Part of #4124

### DIFF
--- a/crates/test-framework/src/metrics/mod.rs
+++ b/crates/test-framework/src/metrics/mod.rs
@@ -1,0 +1,483 @@
+/*
+Copyright 2024-2025 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


crates/test-framework/src/metrics/mod.rs
```diff
@@ -0,0 +1,483 @@
...
+        let records = self.build()?;
+        let ctx = SessionContext::new();
+        let provider = MemTable::try_new(schema, vec![records])?;
+        let df = DataFrame::new(
+            ctx.state(),
+            LogicalPlanBuilder::scan(UNNAMED_TABLE, provider_as_source(Arc::new(provider)), None)?
+                .build()?,
+        );
+
+        df.show().await?;
```
phillipleblanc: There is a utility function in Arrow that we can use to print the batches without needing to pull in DataFusion: https://docs.rs/arrow/latest/arrow/util/pretty/fn.print_batches.html

phillipleblanc (approved):

PR #4245, opened 2025-01-10
Run benchmark test with the new test framework

Sevenannn:
# 🗣 Description

A minimum impelemntation to run benchmark test with the new test framework. The benchmark test could be run with 
`cargo run -p testoperator run bench `. Bench statistics collection are not included in this PR

This is not needed for catalog connectors with source connector, but might be useful for running iceberg catalog e2e.

## ⛓️‍💥 Breaking Change


crates/test-framework/src/spicetest/mod.rs
```diff
@@ -71,24 +71,26 @@ impl TestState for Completed {}
 
 /// A throughput test is a test that runs a set of queries in a loop until a condition is met
 /// The test queries can also be run in parallel, each with the same end condition.
-pub struct ThroughputTest<S: TestState> {
+pub struct SpiceTest<S: TestState> {
     name: String,
     spiced_instance: SpicedInstance,
     query_count: usize,
     parallel_count: usize,
+    iterations: usize,
```
peasee: iteration count is retained in `EndCondition::QuerySetCompleted`

tools/testoperator/src/tests/bench/mod.rs
```diff
@@ -0,0 +1,67 @@
...
+    spiced_instance
+        .wait_for_ready(Duration::from_secs(10))
+        .await?;
+
+    // baseline run
+    println!("Running benchmark test");
+    let benchmark_test = SpiceTest::new(app.name.clone(), spiced_instance)
+        .with_query_set(queries.clone())
+        .with_end_condition(EndCondition::QuerySetCompleted(1))
+        .with_iterations(5)
```
peasee:
```suggestion
        .with_end_condition(EndCondition::QuerySetCompleted(5))
```

peasee (changes requested):

crates/test-framework/src/spicetest/mod.rs
Sevenannn: I'd like some feedback regrading doing the iterations in benchmark test. I saw different benchmarks having different approaches to run query iteration, in Clickbench it runs each queries for the number of iterations before proceeding to the next query, which is the approach I'm taking here, and in the old benchmark test, whereas in some other tpc benchmark it runs the whole query set for certain iterations, which is the approach you're suggesting here. I wonder which approach would be better for our usage.

peasee: The only difference in each is the application of `results_cache`. Running query sets in iterations rather than queries in iterations would result in fewer cache hits (too much time elapsed between executions).

Unless we're specifically testing for the behavior of results cache, there is no difference and we should not duplicate code.

tools/testoperator/src/tests/bench/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+pub(crate) async fn run(args: &TestArgs) -> anyhow::Result<()> {
+    let query_set = QuerySet::from(args.query_set.clone());
+    let query_overrides = args.query_overrides.clone().map(QueryOverrides::from);
+    let queries = query_set.get_queries(query_overrides);
+
+    let (app, start_request) = get_app_and_start_request(args)?;
+    let mut spiced_instance = SpicedInstance::start(start_request).await?;
+
+    spiced_instance
+        .wait_for_ready(Duration::from_secs(10))
```
peasee:
```suggestion
        .wait_for_ready(Duration::from_secs(args.ready_wait.unwrap_or(30) as u64))
```

There is a parameter here we should use.

peasee (approved):

PR #4297, opened 2025-01-13
Fix Bug: No field named body_embedding when do vector search with refresh sql containing subset of columns

### LINKED ISSUE #4258: Bug: No field named body_embedding when do vector search with refresh sql containing subset of columns
## Describe the bug

There is the following search error when refresh sql contains subset of columns, for example `body` column is included in refresh sql and  it contains embeddings

> search> test
2025/01/12 06:04:26 ERROR search failed error="Error occurred interacting with datafusion: Failed to execute query: Schema error: No field named body_embedding. Valid fields are spice.public.issues.id, spice.public.issues.number, spice.public.issues.title, spice.public.issues.body."

## To Reproduce

```yaml
version: v1beta1
kind: Spicepod
name: test

embeddings:
  - name: hf_minilm
    from: huggingface:huggingface.co/sentence-transformers/all-MiniLM-L6-v2

datasets:
  - from: github:github.com/spiceai/spiceai/issues
    name: issues
    description: GitHub Issues for the Spice.ai project (github.com/spiceai/spiceai)
    params:
      github_token: ${secrets:GITHUB_TOKEN}
      github_query_mode: "search"
    acceleration:
      enabled: true
      refresh_sql: "SELECT id, number, title, body FROM issues WHERE created_at='2024-11-09T22:47:45'"
    embeddings:
      - column: body
        use: hf_minilm
```

sgrebnov:
# 🗣 Description

When `refresh_sql` contains a subset of columns with embeddings, Spice internal embedding fields are not included automatically in the dataset schema and are not fetched/calculated during the refresh from the `EmbeddingTable` resulting in broken vector similarity search functionality. This PR implements the logic to:

1. Detect when the source schema contains automatically created embedding fields for a column included in `refresh_sql` and add them.
2. When getting refresh data automatically include Spice internal embedding columns in the logical plan.

Note: The `Projection` node is updated/mutated in-place without additional copies. The logic is applied for data refresh only.

Other Alternatives Considered

1. **Warn user if `refresh_sql` contains a subset of columns with embeddings**: This would require users to manually include Spice internal columns in the refresh SQL. This was rejected as it does not provide the best DX; embedding columns are Spice internal and should be added automatically.
  
2. **Inject Spice internal columns directly into `refresh_sql`**: This alternative involved injecting additional Spice internal columns directly into `refresh_sql` so that data refresh works out of the box, without injecting Spice internal columns into the plan. The preferred original approach was chosen to preserve the original `refresh_sql` in traces, logs, and telemetry, as updating the `refresh_sql` could cause confusion.

## 🔨 Related Issues

Closes https://github.com/spiceai/spiceai/issues/4258

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/arrow_tools/src/schema.rs
+++ b/crates/arrow_tools/src/schema.rs
@@ -136,3 +136,63 @@ pub fn schema_meta_get_computed_columns(
         None
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use datafusion::arrow::datatypes::{DataType, Field, Schema};
+
+    fn create_test_schema_with_embeddings() -> Schema {
+        Schema::new(vec![
+            Field::new("id", DataType::Int64, false),
+            Field::new("name", DataType::Utf8, false),
+            Field::new("value", DataType::Float64, true),
+            Field::new(
+                "name_embedding",
+                DataType::List(Arc::new(Field::new(
+                    "item",
+                    DataType::FixedSizeList(
+                        Arc::new(Field::new("item", DataType::Float32, false)),
+                        1536,
+                    ),
+                    false,
+                ))),
+                false,
+            ),
+            Field::new(
+                "name_offset",
+                DataType::List(Arc::new(Field::new(
+                    "item",
+                    DataType::FixedSizeList(
+                        Arc::new(Field::new("item", DataType::Int32, false)),
+                        2,
+                    ),
+                    false,
+                ))),
+                false,
+            ),
+        ])
+    }
+
+    #[test]
+    fn test_computed_columns_meta() {
+        let mut schema = create_test_schema_with_embeddings();
+
+        let mut computed_columns_meta = HashMap::new();
+        computed_columns_meta.insert(
+            "name".to_string(),
+            vec!["name_embedding".to_string(), "name_offset".to_string()],
+        );
+
+        // Set metadata
+        set_computed_columns_meta(&mut schema, &computed_columns_meta);
+
+        // Retrieve computed columns metadata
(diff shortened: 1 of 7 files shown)


crates/arrow_tools/src/schema.rs
```diff
@@ -91,3 +93,46 @@ pub fn expand_views_schema(schema: &Schema) -> Schema {
...
+        set_computed_columns_meta_for_base_column(schema, base_column, computed_columns);
+    }
+}
+
+pub fn set_computed_columns_meta_for_base_column(
+    schema: &mut Schema,
+    base_column: &str,
+    computed_columns: &[String],
+) {
+    schema.metadata.insert(
```
sgrebnov, line 111: An alternative approach considered was using field metadata directly, e.g., adding an `"is_computed": "base_col_name"` key-value pair. However, fields cannot be mutated as they are `Arc` references and do not implement `DerefMut`.

crates/arrow_tools/src/schema.rs
```diff
@@ -91,3 +93,46 @@ pub fn expand_views_schema(schema: &Schema) -> Schema {
...
+        format!("{base_column}_computed_columns"),
+        computed_columns.join(","),
+    );
+}
+
+#[must_use]
+pub fn schema_meta_get_computed_columns(
```
phillipleblanc, line 118: These functions could use some tests.

phillipleblanc (dismissed):

crates/arrow_tools/src/schema.rs
sgrebnov: Added test, thank you!

phillipleblanc (approved):

PR #4311, opened 2025-01-14
feat: Update load test criteria, testoperator updates

peasee:
# 🗣 Description

* Updates the load test criteria to introduce the concept of the stop-light system for timing measurements (green, yellow, red)
* Updates the automatic pass to less than 1000ms of the 99th percentile
* Updates printing output to use progress bars rather than printing a line for each individual query, to limit log spam on outputs
* Print memory usage on every kind of test
* Print metrics on benchmark tests
* Moved common environment export code into a common path

### DIFF
--- a/.github/actions/setup-testoperator/action.yml
+++ b/.github/actions/setup-testoperator/action.yml
@@ -32,6 +32,20 @@ runs:
         sudo mkdir -p /opt/spiced/data
         sudo chown runner:runner -R /opt/spiced/data
 
+    - name: Download DuckDB TPCH Data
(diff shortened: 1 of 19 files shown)


phillipleblanc (dismissed):

peasee:
I added a fallback for terminal output while running in actions, because progress bars weren't rendering there.

Prints only from the first worker thread, as threads should be close enough that we don't need to spam log all of them.

phillipleblanc (approved):

Sevenannn (approved):

Issue #4027, opened 2024-12-30
Enhancement: DuckDB Data Accelerator 1.0 Stable


phillipleblanc:
## Goal-State/What/Result

The DuckDB accelerator is released at Stable quality.

## Why/Purpose

Part of our stated roadmap: https://github.com/spiceai/spiceai/blob/trunk/docs/ROADMAP.md

## By When

**Issue/Spec written and reviewed:** 14th Jan
**Done-Done:** 17th Jan

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Stable Release Criteria

The Stable release criteria expand on and require that all [RC release criteria](./rc.md) continue to pass for the accelerator.

- [x] All [RC release criteria](./rc.md) pass for this accelerator.

### Test Coverage

Stable quality accelerators should be able to run test packages derived from the following:

- [TPC-H](https://www.tpc.org/TPC-H/)
- [TPC-DS](https://www.tpc.org/TPC-DS/)
- [ClickBench](https://github.com/ClickHouse/ClickBench)

Indexes are not required for test coverage, but can be introduced if required for tests to pass (e.g. due to performance characteristics, etc).

#### TPC-H

- [x] An end-to-end test is created simulating an append mode load of the TPC-H dataset at scale factor 10 in all [Access Modes](../definitions.md). The data load should complete to the source within 30 minutes, using 5 minute refresh intervals on the accelerator.
  - An accelerator that takes longer than 1 hour to load the dataset is considered a test fail.
- [x] An end-to-end test is created using a refresh SQL that excludes results from the accelerator for the TPC-H dataset at scale factor 1 in all [Access Modes](../definitions.md). TPC-H queries should induce a zero results action, with `on_zero_results: use_source` behavior.
  - The test validates that for queries that would return results from the source that are excluded in the accelerator, that they correctly return results through the use of `on_zero_results`.
[...]

PR #4319, opened 2025-01-14
feat: add testoperator data consistency

peasee:
# 🗣 Description

* Changes the log level for on zero results fallback to `DEBUG`
* Adds a new test operator command, `data-consistency`, and associated test framework changes.
    * A data consistency run operates over 2 supplied spicepods. Both spicepods are created in separate apps, one after the other, with a benchmark test run on each. Row counts for each query are then compared, with a test failure for non-matching row counts.
    * The test also fails if query executions within the same app run don't match - e.g. the 2 workers in a single test got different query results.

## 🔨 Related Issues

* Part of #4027

### DIFF
--- a/tools/testoperator/src/commands/mod.rs
+++ b/tools/testoperator/src/commands/mod.rs
@@ -33,9 +33,10 @@ pub enum TestCommands {
     Throughput(TestArgs),
     Load(TestArgs),
     Bench(TestArgs),
+    DataConsistency(DataConsistencyArgs),
 }
 
-#[derive(Parser)]
+#[derive(Parser, Debug, Clone)]
 pub struct TestArgs {
     /// Path to the spicepod.yaml file
     #[arg(short('p'), long)]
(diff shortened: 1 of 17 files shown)


tools/testoperator/src/commands/mod.rs
```diff
@@ -98,7 +99,7 @@ impl From<QuerySetArg> for QuerySet {
         match arg {
             QuerySetArg::Tpch => QuerySet::Tpch,
             QuerySetArg::Tpcds => QuerySet::Tpcds,
-            QuerySetArg::ClickBench => QuerySet::ClickBench,
+            QuerySetArg::Clickbench => QuerySet::ClickBench,
```
phillipleblanc:
```suggestion
            QuerySetArg::Clickbench => QuerySet::Clickbench,
```

phillipleblanc (dismissed):

Sevenannn:
Is it possible to also compare results on top of row count? e.g. compare first 10 rows of records in the result and check if they are the same.

peasee:
> Is it possible to also compare results on top of row count? e.g. compare first 10 rows of records in the result and check if they are the same.

Yes, it would be possible. I will not be adding it in this pull request though.

Sevenannn (approved):

PR #4355, opened 2025-01-16
New test operator: HTTP consistency

Jeadie:
# 🗣 Description
 - New test for test operator; HTTP consistency. This checks that the latency of various HTTP components (e.g. models/embeddings) does not degrade after persistent use.


crates/test-framework/Cargo.toml
```diff
@@ -9,25 +9,30 @@ rust-version.workspace = true
...
+tokio.workspace = true
+tonic.workspace = true
+uuid = { workspace = true, features = ["v4"] }
 
 [target.'cfg(not(windows))'.dependencies]
 nix = { version = "0.29", features = ["signal"] }
 
 [features]
+default = ["models"]
+models = ["dep:async-openai"]
```
phillipleblanc: I don't think we need to feature flag the test operator - that just adds extra complexity for no real gain

crates/test-framework/src/spicetest/consistency/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /// Sends a request to the component and returns the duration of the request.
+    /// Payload may be the entire HTTP request body, or a portion of it (dependent of the component).
+    pub async fn send_request(&self, client: &Client, payload: &str) -> anyhow::Result<Duration> {
+        let c = OpenAIClient::with_config(OpenAIConfig::default().with_api_base(self.api_base()))
+            .with_http_client(client.clone())
+            .clone();
+
+        let start_time = SystemTime::now();
```
phillipleblanc: Its better to use `Instant::now()` for tracking timings, its a monotonic clock vs the system clock.

crates/test-framework/src/spicetest/consistency/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+                        Err(_) => CreateChatCompletionRequestArgs::default()
+                            .model(model.clone())
+                            .messages(vec![ChatCompletionRequestMessage::User(
+                                ChatCompletionRequestUserMessageArgs::default()
+                                    .content(payload.to_string())
+                                    .build()
+                                    .expect("failed to build user message"),
+                            )])
+                            .build()
+                            .expect("failed to build model request"),
```
phillipleblanc: Just return the error with `?` - this will fail linting

crates/test-framework/src/spicetest/consistency/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+        Self {
+            start_time: None,
+            spiced_instance,
+            config,
+            state: ConsistencyState::NotStarted,
+        }
+    }
+
+    pub async fn start(self) -> anyhow::Result<ConsistencySpiceTest> {
+        if !matches!(self.state, ConsistencyState::NotStarted) {
```
phillipleblanc, line 157: If you look at the pattern of the SpiceTest, it is parameterized on the state (i.e. `struct SpiceTest<S: TestState>`) which means you can define functions to only exist when its in a particular state, which uses the type system to enforce this check at compile time rather than at runtime

This is an example of the TypeState pattern: https://cliffle.com/blog/rust-typestate/

crates/test-framework/src/spicetest/consistency/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    /// The HTTP component, within the Spiced instance, to test.
+    pub component: HttpConsistencyComponent,
+}
+
+/// A component within the Spiced instance to test for consistency.
+///
+/// This component must be accessible over HTTP.
+#[derive(Clone)]
+pub enum HttpConsistencyComponent {
```
phillipleblanc, line 64: Might be cleaner to make this a trait with separate implementations? Otherwise it might get unwieldy with more than a couple of variants

phillipleblanc (approved):

PR #4374, opened 2025-01-17
Embedding support for xAI

Sevenannn:
# 🗣 Description

## ⛓️‍💥 Breaking Change


Sevenannn:
- Closed as we don't have access to xAI embedding models. This could be reopened / reused when needed

PR #4407, opened 2025-01-20
Spice CLI: fallback to CPU build for unsupported GPU Compute Capability

sgrebnov:
# 🗣 Description

Spice CLI: fallback to CPU build for unsupported GPU Compute Capability

## ⛓️‍💥 Breaking Change

### DIFF
--- a/bin/spice/pkg/github/runtime_release.go
+++ b/bin/spice/pkg/github/runtime_release.go
@@ -132,7 +132,7 @@ func get_ai_accelerator() (string, bool) {
 		}
 
 		if !checkCudaVersionSupported(*version) {
-			slog.Warn(fmt.Sprintf("GPU version is not supported. Spice will run but without GPU acceleration. Detected GPU version: '%s'", *version))
+			slog.Warn(fmt.Sprintf("Spice detected a GPU, but the GPU version (%s) is not supported for model acceleration. Spice will fallback to using the CPU to run local models, which may impact performance.", *version))
 			return "", false
 		}
 
--- a/bin/spice/pkg/github/runtime_release.go
+++ b/bin/spice/pkg/github/runtime_release.go
@@ -97,6 +97,18 @@ func getRustArch() string {
 	return runtime.GOARCH
 }
 
+// GPU versions that are supported via dedicated CUDA builds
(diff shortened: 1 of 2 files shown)


bin/spice/pkg/github/runtime_release.go
```diff
@@ -114,9 +126,18 @@ func get_ai_accelerator() (string, bool) {
...
 		}
-		if version != nil {
-			return "cuda_" + *version, true
+
+		if version == nil {
+			return "", false
+		}
+
+		if !checkCudaVersionSupported(*version) {
+			slog.Warn(fmt.Sprintf("The GPU Compute Capability '%s' is not supported by Spice GPU builds. Supported versions are: %v. Falling back to CPU.",
```
phillipleblanc: This is to technical of an error message. Let's make sure to follow the error guidelines. Most people don't know what a compute capability is, let's just say that the GPU isn't supported.

sgrebnov: Thank you, updated to

>GPU version is not supported. Spice will run but without GPU acceleration. Detected GPU version: _version_

bin/spice/pkg/github/runtime_release.go
```diff
@@ -114,9 +126,17 @@ func get_ai_accelerator() (string, bool) {
...
 		}
-		if version != nil {
-			return "cuda_" + *version, true
+
+		if version == nil {
+			return "", false
+		}
+
+		if !checkCudaVersionSupported(*version) {
+			slog.Warn(fmt.Sprintf("GPU version is not supported. Spice will run but without GPU acceleration. Detected GPU version: '%s'", *version))
```
sgrebnov:
```suggestion
			slog.Warn(fmt.Sprintf("GPU version ('%s') is not supported. Spice will run but without GPU acceleration.", *version))
```
Could be shorter, original version adds technical details to the end of the message which is more user friendly IMO

phillipleblanc: I would word it like:
`Spice detected a GPU, but the GPU version (%s) is not supported for model acceleration. Spice will fallback to using the CPU to run local models, which may impact performance.`

sgrebnov: Thank you

phillipleblanc (approved):

PR #4409, opened 2025-01-20
Fix streaming for Openai and Anthropic

Jeadie:
# 🗣 Description

## Issue with OpenAI
Now, usage isn't added to the last non-empty stream chunk that has the `finish_reason: stop`. Rather, its returned in an empty payload (i.e. `.choices.length() ==0`) at the end. 
```jsonl
{"id":"chatcmpl-ArZlM3uuEuwDzGvGjaetc4GeaWDca","choices":[{"index":0,"delta":{"content":null,"refusal":null},"finish_reason":"stop","logprobs":null}],"created":1737332220,"model":"openai-with-spice","service_tier":"default","system_fingerprint":"fp_50cad350e4","object":"chat.completion.chunk"}

{"id":"chatcmpl-ArZlM3uuEuwDzGvGjaetc4GeaWDca","choices":[],"created":1737332220,"model":"openai-with-spice","service_tier":"default","system_fingerprint":"fp_50cad350e4","object":"chat.completion.chunk","usage":{"prompt_tokens":9,"completion_tokens":9,"total_tokens":18,"prompt_tokens_details":{"audio_tokens":0,"cached_tokens":0},"completion_tokens_details":{"accepted_prediction_tokens":0,"audio_tokens":0,"reasoning_tokens":0,"rejected_prediction_tokens":0}}}
```
For tool using models, we would not process the last stream packet. Now we handle it explicitly. 

## Issue with Anthropic
We were not closing the stream gracefully on an SSE event with type `message_stop`. Anthropic started returning an error in the SSE when this happened. We now gracefully handle this by returning early when we receive this stream end SSE.

### DIFF
--- a/crates/llms/src/anthropic/types_stream.rs
+++ b/crates/llms/src/anthropic/types_stream.rs
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 #![allow(deprecated)] // `function_call` argument is deprecated but no builder pattern alternative is available.
+use super::types::{MessageRole, StopReason, Usage};
 use async_openai::{
     error::{ApiError, OpenAIError},
     types::{
@@ -23,12 +24,11 @@ use async_openai::{
     },
 };
 use futures::{Stream, StreamExt};
+use reqwest_eventsource::Error as SseError;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::{collections::HashMap, fmt, pin::Pin, sync::Arc, time::SystemTime};
 use tokio::sync::Mutex;
-
-use super::types::{MessageRole, StopReason, Usage};
 
 #[derive(Debug, Deserialize, Serialize)]
 #[serde(tag = "type")]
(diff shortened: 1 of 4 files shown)


ewgenius (dismissed):

crates/llms/src/anthropic/types_stream.rs
```diff
@@ -393,18 +384,23 @@ pub fn transform_stream(
...
                             code: None,
-                        }))
+                        })))
                     }
                 }
             }
+        })
+        // Because we don't early exit on [`MessageCreateStreamResponse::MessageStop`], we need to handle stream end explicitly, otherwise we will infinite loop on the stream.
+        .take_while(|item| {
+            let keep_going = !matches!(item, Err(OpenAIError::ApiError(ApiError { message, .. })) if message == "Stream ended");
```
phillipleblanc: This seems brittle? Where does this get emitted?

crates/runtime/src/model/tool_use.rs
```diff
@@ -680,6 +680,15 @@ fn make_a_stream(
                             }
                         }
                     }
+
+                    // When there are no [`ChatChoiceStream`]s, but the model has usage, send the response (with not choices).
```
sgrebnov: ### 
```suggestion
                    // When there are no [`ChatChoiceStream`]s, but the model has usage, send the response (with no choices).
```

sgrebnov (dismissed):

phillipleblanc (dismissed):

sgrebnov (approved):

PR #4417, opened 2025-01-20
Handle OpenAI API Errors

### LINKED ISSUE #4413: Using structured outputs without having `json` in the prompt causes a 500 in Spice
## Describe the bug

Making a request to `http://localhost:8090/v1/chat/completions` and asking for structured output via the `response_format` parameter will return a 500 if the prompt does not contain the word "json". An error is logged in the runtime with the following:

`ERROR runtime::http::v1::chat: Error from v1/chat: invalid_request_error: 'messages' must contain the word 'json' in some form, to use 'response_format' of type 'json_object'. (param: messages)`

Spice should handle this and return the 400 to the user, not a 500

## To Reproduce

Configure a Spicepod with an openai model that supports structured outputs:

```yaml
models:
  - from: openai:gpt-4o-mini
    name: mini
    params:
      openai_api_key: ${ secrets:OPENAI_API_KEY }
      tools: auto
```

Make a request to the model and ask for a structured output without including the word JSON:

```bash
curl -v -L 'http://localhost:8090/v1/chat/completions' \
             -H 'Content-Type: application/json' \
             -H 'Accept: application/json' \
             -d '{
                     "messages": [
                       {
                         "content": "You only talk like a pirate",
                         "role": "system"
                       },
                       {
                         "content": "What can you do?",
                         "role": "user"
                       }
                     ],
                     "model": "mini",
                     "response_format": { "type": "json_object" }
                   }' 
```

## Expected behavior

The 400 is returned to the user, not a 500 with an error logged in the runtime

### Spicepod

### Output of `explain query`

E.g. `explain select 1`

```bash
```

### spice, spiced, OS info

- OS info: `uname -a`

- [ ] Yes
- [ ] No

ewgenius:
# 🗣 Description

Handles OpenAI API errors, and pass error details and status in v1/chat/completions response

CleanShot 2025-01-20 at 16 03 36@2x
CleanShot 2025-01-20 at 16 58 46@2x
CleanShot 2025-01-20 at 16 58 58@2x

## 🔨 Related Issues

closes #4413

### DIFF
--- a/crates/runtime/src/http/v1/chat.rs
+++ b/crates/runtime/src/http/v1/chat.rs
@@ -227,6 +227,8 @@ impl OpenaiErrorEvent {
     }
 }
 
+/// Converts OpenAI errors to HTTP responses
+/// Preserve the original OpenAI error structure to maintain compatibility with OpenAI's documentation
 #[must_use]
 pub fn openai_error_to_response(e: OpenAIError) -> Response {
     match e {
--- a/crates/runtime/src/http/v1/chat.rs
+++ b/crates/runtime/src/http/v1/chat.rs
@@ -227,8 +227,8 @@ impl OpenaiErrorEvent {
     }
 }
 
-/// Converts OpenAI errors to HTTP responses
-/// Preserve the original OpenAI error structure to maintain compatibility with OpenAI's documentation
+/// Converts `OpenAI` errors to HTTP responses
+/// Preserve the original `OpenAI` error structure to maintain compatibility with `OpenAI` documentation
 #[must_use]
 pub fn openai_error_to_response(e: OpenAIError) -> Response {
     match e {
--- a/crates/runtime/src/http/v1/chat.rs
+++ b/crates/runtime/src/http/v1/chat.rs
@@ -20,7 +20,10 @@ use std::{convert::Infallible, sync::Arc, time::Duration};
 use crate::{http::traceparent::override_task_history_with_traceparent, model::LLMModelStore};
 #[cfg(feature = "openapi")]
 use async_openai::types::CreateChatCompletionResponse;
-use async_openai::types::{ChatCompletionResponseStream, CreateChatCompletionRequest};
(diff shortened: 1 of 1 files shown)


crates/runtime/src/http/v1/chat.rs
```diff
@@ -221,3 +226,29 @@ impl OpenaiErrorEvent {
         }
     }
 }
+
+#[must_use]
+pub fn openai_error_to_response(e: OpenAIError) -> Response {
+    match e {
+        OpenAIError::InvalidArgument(_) => (StatusCode::BAD_REQUEST, e.to_string()).into_response(),
+        OpenAIError::ApiError(api_error) => {
+            let error_response = serde_json::json!({
```
Jeadie, line 237: Why doesn't `error_response` doesn't naturally serialise as this `json!` does

Jeadie (dismissed):

crates/runtime/src/http/v1/chat.rs
```diff
@@ -221,3 +226,29 @@ impl OpenaiErrorEvent {
         }
     }
 }
+
+#[must_use]
+pub fn openai_error_to_response(e: OpenAIError) -> Response {
```
Jeadie, line 233: Worth making a note that we should keep the error body untouched for OpenAI, so that users can use OpenAI error documentation

phillipleblanc (dismissed):

crates/runtime/src/http/v1/chat.rs
ewgenius: It can be serialized, but the type will be `Json<ApiError>` and there is no `into_response` implementation for this.
So have to construct json value directly.

crates/runtime/src/http/v1/chat.rs
ewgenius: Done

sgrebnov (approved):

PR #4418, opened 2025-01-20
Metrics for LLMs+ embeddings

Jeadie:
# 🗣 Description

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/runtime/src/embeddings/metrics.rs
+++ b/crates/runtime/src/embeddings/metrics.rs
@@ -23,21 +23,21 @@ use opentelemetry::{
     Key, KeyValue, Value,
 };
 
-static METER: LazyLock<Meter> = LazyLock::new(|| global::meter("embedding"));
+static METER: LazyLock<Meter> = LazyLock::new(|| global::meter("embeddings"));
 
 pub(crate) static EMBEDDING_REQUESTS: LazyLock<Counter<u64>> =
-    LazyLock::new(|| METER.u64_counter("embedding_requests").build());
+    LazyLock::new(|| METER.u64_counter("embeddings_requests").build());
 
 pub(crate) static FAILURES: LazyLock<Counter<u64>> = LazyLock::new(|| {
     METER
-        .u64_counter("embedding_failures")
+        .u64_counter("embeddings_failures")
         .with_description("Number of embedding failures.")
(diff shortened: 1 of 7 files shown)


phillipleblanc (dismissed):

crates/runtime/src/embeddings/metrics.rs
```diff
@@ -0,0 +1,94 @@
...
+pub(crate) static FAILURES: LazyLock<Counter<u64>> = LazyLock::new(|| {
+    METER
+        .u64_counter("embedding_failures")
+        .with_description("Number of embedding failures.")
+        .build()
+});
+
+pub(crate) static EMBEDDING_INTERNAL_DURATION_MS: LazyLock<Histogram<f64>> = LazyLock::new(|| {
+    METER
+        .f64_histogram("embedding_internal_request_duration_ms")
```
sgrebnov:
```suggestion
        .f64_histogram("embeddings_internal_request_duration_ms")
```

To be consistent with `embeddings_load_errors`/etc where `embeddings` is in plural

crates/runtime/src/embeddings/metrics.rs
```diff
@@ -0,0 +1,94 @@
...
+};
+
+static METER: LazyLock<Meter> = LazyLock::new(|| global::meter("embedding"));
+
+pub(crate) static EMBEDDING_REQUESTS: LazyLock<Counter<u64>> =
+    LazyLock::new(|| METER.u64_counter("embedding_requests").build());
+
+pub(crate) static FAILURES: LazyLock<Counter<u64>> = LazyLock::new(|| {
+    METER
+        .u64_counter("embedding_failures")
```
sgrebnov:
```suggestion
        .u64_counter("embeddings_failures")
```

For consistency with other embeddings metrics, for example `embeddings_load_errors `

crates/runtime/src/embeddings/metrics.rs
Jeadie: I do think we should possibly change `embeddings_load_errors` to `embedding_load_errors`. But since we already have it as a plural I will follow convention. We can change all afterwards.

phillipleblanc (approved):

PR #4454, opened 2025-01-21
fix: Use specific model errors where available

peasee:
# 🗣 Description

* Adds a function to map a boxed error, based on the error as a string, to a known `llms::chat::Error`

### DIFF
--- a/crates/llms/src/chat/mod.rs
+++ b/crates/llms/src/chat/mod.rs
@@ -111,6 +111,9 @@ pub enum Error {
 
     #[snafu(display("Failed to find tensors for the model.\nExpected tensors with a file extension of: {extensions}.\nVerify the model is correctly configured, and try again."))]
     ModelMissingTensors { extensions: String },
+
+    #[snafu(display("Failed to load a file specified for the model.\nCould not find the file: {file_url}.\nVerify the `files` parameters for the model, and try again."))]
+    ModelFileMissing { file_url: String },
 }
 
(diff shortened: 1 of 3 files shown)


phillipleblanc (dismissed):

ewgenius (dismissed):

peasee:
New error for missing files from the `files` parameter, if we can identify them from huggingface:

```
2025-01-21T04:14:01.731055Z  WARN runtime::init::model: Failed to load LLM: llama3.
Failed to load a file specified for the model.
Could not find the file: https://huggingface.co/spiceai/darts/resolve/latest/tokenizer.json.
Verify the `files` parameters for the model, and try again.
```

phillipleblanc (dismissed):

crates/llms/src/chat/mod.rs
```diff
@@ -93,10 +95,58 @@ pub enum Error {
 
     #[snafu(display("The specified model, '{from}', does not support executing the task '{task}'.\nSelect a different model or task, and try again."))]
     UnsupportedTaskForModel { from: String, task: String },
+
+    #[snafu(display("Failed to find tensors for the model.\nExpected tensors with a file extension of: {extensions}.\nVerify the model is correctly configured, and try again."))]
```
Jeadie: They're more commonly called weights rather than tensors. That's also the file type we use in the spicepod :https://spiceai.org/docs/reference/spicepod/models#files

phillipleblanc (approved):

PR #4457, opened 2025-01-21
docs: Promote HF and File model components

peasee:
# 🗣 Description

* Promotes HF and Local model provider to Beta
* Promotes HF and Local embedding provider to Release Candidate


Jeadie (approved):

Jeadie:
Thank you! I forgot we showed component statuses here too

PR #4490, opened 2025-01-23
Use upgraded spice cli when performing runtime upgrade in spice upgrade

Sevenannn:
# 🗣 Description

In spice upgrade command, reload spice cli when new cli has been downloaded and replaced the old one, and use the new version of cli to perform runtime upgrade.

For Unix:
1. Spice CLI already at the latest version && No runtime upgrade required: no CLI upgrade, no runtime upgrade
2. Spice CLI already at the latest version && runtime upgrade required: no CLI upgrade, finish the runtime upgrade with the newest cli
3. Spice CLI at a outdated version / force update: CLI will upgrade, new CLI will replace old CLI to finish upgrade, in new CLI run it will be the same as scenario (1) / (2) depending on the runtime upgrade status

For Windows:
1. Spice CLI already at the latest version && No runtime upgrade required: no CLI upgrade, no runtime upgrade
2. Spice CLI already at the latest version && runtime upgrade required: no CLI upgrade, finish the runtime upgrade with the newest cli
3. Spice CLI at an outdated version / force update: Upgrade CLI, Prompt user to rerun upgrade, and the old cli shut down. The second run will be exactly the same as scenario (1) / (2)

Special for Windows:
Windows spice upgrade cleans up any spice temporary directory created by spice upgrade after the cli upgrade on each run.

## ⛓️‍💥 Breaking Change


bin/spice/cmd/upgrade.go
```diff
@@ -45,103 +48,143 @@ spice upgrade
...
+		slog.Error("upgrading the spice binary", "error", err)
+		return
+	}
+
+	releaseFilePath := filepath.Join(spiceBinDir, constants.SpiceCliFilename)
+
+	// On Windows, it is not possible to overwrite a binary while it's running.
+	// However, it can be moved/renamed making it possible to save new release with the original name.
+	if util.IsWindows() {
+		runningCliTempLocation := filepath.Join(spiceBinDir, constants.SpiceCliFilename+".bak")
```
lukekim: Don't call this .bak. Copy the binary to the Window temporary directory, under a temp path, and run it from there. That way it will get cleaned up. Also, clean it up on exit.

bin/spice/cmd/upgrade.go
```diff
@@ -32,6 +33,8 @@ import (
 	"github.com/spiceai/spiceai/bin/spice/pkg/version"
 )
 
+const reloadEnvVar = "SPICE_CLI_RELOADED"
```
lukekim: Put in constants.

bin/spice/cmd/upgrade.go
```diff
@@ -45,103 +48,203 @@ spice upgrade
...
+		if os.Getenv(constants.SpiceUpgradeReloadEnv) != "true" {
+			// Run CLI upgrade
+			if !upgradeCli(force, rtcontext) {
+				// Exit if CLI upgrade fail
+				return
+			}
+		}
+
+		// Cleanup old binaries on windows
+		if runtime.GOOS == "windows" {
```
sgrebnov, line 67: Nit: we can call this for all platforms so the logic is consistent and perform clean up on Windows only (what we already check/do)

```
func cleanupOldBinaries() {
	if !util.IsWindows() {
		return
	}
```

bin/spice/cmd/upgrade.go
```diff
@@ -45,103 +48,203 @@ spice upgrade
 			return
 		}
 
-		slog.Info("Checking for latest Spice CLI release...")
```
sgrebnov, line 48: Is removing upgrade tracing intentional?

```
slog.Info("Checking for latest Spice CLI release...")

slog.Info(fmt.Sprintf("Using the latest version %s. No upgrade required.", release.TagName))
```

bin/spice/pkg/constants/constants.go
```diff
@@ -21,8 +21,10 @@ import (
 )
 
 const (
-	DotSpice               = ".spice"
-	SpicePodsDirectoryName = "spicepods"
+	DotSpice                  = ".spice"
+	SpicePodsDirectoryName    = "spicepods"
+	SpiceUpgradeReloadEnv     = "SPICE_CLI_RESTARTED"
```
sgrebnov:
```suggestion
	SpiceUpgradeReloadEnv     = "SPICE_CLI_RESTARTED_AFTER_UPGRADE"
```

sgrebnov (dismissed):

sgrebnov (approved):

PR #4495, opened 2025-01-24
Run LLM integration tests on Macs; add running local models. 

Jeadie:
# 🗣 Description
 - Run LLM tests on Mac runners so that we can test local models. 
 - Also run Beta functionality testing for embeddings.


.github/actions/run-llm-integration-tests/action.yml
```diff
@@ -0,0 +1,49 @@
+name: Run Integration Tests
+description: Runs `/crates/llms/tests` integration tests.
+inputs:
+  model_allowlist:
+    description: "Comma-separated list of allowed models for the integration test."
```
sgrebnov: We are using single quotes `'` for quoting in YAML. Please verify the formatter settings.

.github/actions/run-llm-integration-tests/action.yml
```diff
@@ -0,0 +1,49 @@
+name: Run Integration Tests
```
sgrebnov:
```suggestion
name: Run LLMs Integration Tests
```

.github/workflows/integration_llms.yml
```diff
@@ -10,6 +10,16 @@ on:
       - "crates/llms/**"
 
   workflow_dispatch:
+    inputs:
+      run:
+        description: "Which groups of models to run"
+        required: true
+        default: "all"
+        type: choice
+        options:
```
sgrebnov, line 19: Jack, could you pelase eloborate more on the motivation to use `networking` instead of standalone providers -   `HuggingFace`, `OpenAI`, `Anthropic`, `xAI`? I found `networked` term/logic is a little but hard to follow.

.github/workflows/integration_models.yml
```diff
@@ -8,23 +8,23 @@ on:
...
-      - 'version.txt'
-      - '.schema/**'
-      - '.vscode/**'
-      - 'deploy/**'
-      - 'install/**'
-      - 'media/**'
-      - 'monitoring/**'
-      - 'acknowledgements.md'
-      - 'Dockerfile*'
+      - "**/*.md"
```
sgrebnov: I think we should continue using `'` for quotes

.github/actions/run-llm-integration-tests/action.yml
```diff
@@ -0,0 +1,49 @@
...
+description: Runs `/crates/llms/tests` integration tests.
```
sgrebnov: Not sure we require additional action to run LLM integration tests, we can just extend main workflow functionality IMO:
- always build with metal (like we currently do?)
- matrix step with target provider filtering vs all
- test step only runs tests belong to this model provider. So HuggingFace, OpenAI, Anthropic, xAI will run in parallel

sgrebnov: Provider configuration can also include target - linux (cloud hosted) vs spiceai-macos (local model that requires metal)

.github/workflows/integration_llms.yml
```diff
@@ -47,54 +51,34 @@ jobs:
...
       - uses: actions/checkout@v4
-
-      - name: Set up Rust
-        uses: ./.github/actions/setup-rust
+      - name: Run hosted integration tests
+        uses: ./.github/actions/run-llm-integration-tests
         with:
-          os: "linux"
+          model_allowlist: "local_phi3,hf_phi3"
+          networked: "false"
```
Sevenannn: For learning purpose: What's the difference between `test-hosted` and `test-networked` in terms of models running in these test. Might have understand this wrong, should anthropic,openai,xai belong to the hosted model instead?

sgrebnov: I think the terminology here is
- hosted - local/self-hosted
- networked- cloud-hosted

.github/workflows/integration_llms.yml
lukekim: +1 not sure what networked means?

.github/workflows/integration_llms.yml
```diff
@@ -47,54 +51,34 @@ jobs:
           path: ./llms_integration_test
           retention-days: 1
 
-  test:
-    name: LLMs Integration Tests
+  test-hosted:
```
lukekim:
```suggestion
  test-self-hosted:
```

.github/workflows/integration_llms.yml
lukekim: It should be `hosted` and `self-hosted`

.github/workflows/integration_models.yml
lukekim: @Jeadie please fix your prettier to conform to the codebase so we are not changing back and forth.

lukekim (approved):

PR #4556, opened 2025-01-29
Add HTTP consistency & overhead to `testoperator dispatch` tool 

Jeadie:
# 🗣 Description
 - Add HTTP consistency to `testoperator dispatch ...`.
 - Add HTTP overhead to `testoperator dispatch ...`.
 - Add WORKFLOW_COMMIT env variable to dispatch tool.
   -  Also add to `.github/workflows/testoperator_dispatch.yml`.
 - Fix HTTP consistency/overhead GH actions so that there are <= 10 inputs. Do not include inputs with default we wouldn't often change. 
 - Add dispatch tests into `tools/testoperator/dispatch/{beta, rc}` that are criteria levels. 
 - Add `GH_TOKEN`, `SPICED_COMMIT`, and `WORKFLOW_COMMIT` as env variables configurable in CLI.

### DIFF
--- a/tools/testoperator/src/commands/dispatch/mod.rs
+++ b/tools/testoperator/src/commands/dispatch/mod.rs
@@ -100,6 +100,17 @@ pub async fn dispatch(args: DispatchArgs) -> Result<()> {
                 payload["spiced_commit"] = Value::String(args.spiced_commit.clone());
                 payload
             }
+            (
+                TestType::HttpOverhead,
+                DispatchTests {
+                    http_overhead: Some(overhead),
+                    ..
+                },
+            ) => {
+                let mut payload = serde_json::json!(overhead.clone());
+                payload["spiced_commit"] = Value::String(args.spiced_commit.clone());
+                payload
+            }
             (TestType::Benchmark, _) => {
                 println!("Test file {path:#?} does not contain a benchmark test");
                 continue;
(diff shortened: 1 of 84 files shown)


Sevenannn (dismissed):

tools/testoperator/src/commands/dispatch/mod.rs
```diff
@@ -55,42 +53,83 @@ pub async fn dispatch(args: DispatchArgs) -> Result<()> {
...
+            }
+            (
+                TestType::HttpConsistency,
+                DispatchTests {
+                    http_consistency: Some(consistency),
+                    ..
+                },
+            ) => {
+                let mut payload = serde_json::json!(consistency.clone());
+                payload["spiced_commit"] = Value::String(args.spiced_commit.clone());
```
peasee: this probably needs a separate workflow arguments struct with `spiced_commit` as a property, like Load/Bench/ThroughputWorkflowArgs

```rust
struct HttpConsistencyWorkflowArgs {
    http_args: ...,
    spiced_commit: ...
}
```

tools/testoperator/src/commands/dispatch/mod.rs
```diff
@@ -55,42 +53,83 @@ pub async fn dispatch(args: DispatchArgs) -> Result<()> {
...
+            }
+            (
+                TestType::HttpOverhead,
+                DispatchTests {
+                    http_overhead: Some(overhead),
+                    ..
+                },
+            ) => {
+                let mut payload = serde_json::json!(overhead.clone());
+                payload["spiced_commit"] = Value::String(args.spiced_commit.clone());
```
peasee: ditto

tools/testoperator/src/commands/dispatch/mod.rs
Jeadie: Or remove all `*workflowArg` structs by adding `payload["spiced_commit"]` further down?

peasee: I think operating on plain JSON would generally lower the quality of this tooling.

With dedicated structs, it's clear what arguments are being sent to what workflows, without hiding that side-effect in the command as a JSON operation.

Jeadie: Good principle. This is why i wanted to add some of the env variables we were using, into the Clap args struct.

peasee (approved):

PR #4558, opened 2025-01-29
Fix huggingface embedding errors

Sevenannn:
# 🗣 Description

Fix huggingface embedding errors

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/llms/src/embeddings/mod.rs
+++ b/crates/llms/src/embeddings/mod.rs
@@ -44,7 +44,7 @@ pub enum Error {
         source: Box<dyn std::error::Error + Send + Sync>,
     },
 
-    #[snafu(display("Invalid value for `pooling` parameter: {value}.\nSupported pooling value include: `cls`, `mean`, `splade`, `last_token`. Verify `pooling` parameter and try again."))]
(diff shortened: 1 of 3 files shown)


crates/llms/src/embeddings/mod.rs
```diff
@@ -44,6 +44,9 @@ pub enum Error {
         source: Box<dyn std::error::Error + Send + Sync>,
     },
 
+    #[snafu(display("Invalid value for `pooling` parameter: {value}.\nSupported pooling value include: `cls`, `mean`, `splade`, `last_token`. Verify `pooling` parameter and try again."))]
```
peasee:
```suggestion
    #[snafu(display("Invalid value for `pooling` parameter: {value}.\nSpecify a valid pooling value of `cls`, `mean`, `splade`, or `last_token`."))]
```

We can combine the last 2 sentences into a single action. `Allowed values are: {}. Check it`, vs `Check the value is: {}`

peasee: Is there a details page we can also link to in docs?

Jeadie (approved):

crates/llms/src/embeddings/mod.rs
Jeadie: There aren't great docs. Mainly https://www.sbert.net/docs/package_reference/sentence_transformer/models.html#sentence_transformers.models.Pooling

PR #4560, opened 2025-01-29
Add federated source information to refresh errors

### LINKED ISSUE #4552: Data refresh error must include data connector information
## Describe the bug

Current behavior/error is shown below, similar to dataset registration errors, refresh error should include data connector information

>2025-01-27T18:51:52.494087Z ERROR runtime::accelerated_table::refresh_task: Failed to refresh dataset users: Failed to get data from the connector.

sgrebnov:
# 🗣 Description

Propagate federated source information to refresh logic and include it to refresh errors:

>2025-01-29T04:05:32.656513Z  WARN runtime::accelerated_table::refresh_task: Failed to load data for dataset taxi_trips (s3): Failed to get data from the connector.
...

## 🔨 Related Issues

Closes  #4552 

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

1. We have federated TableProvider but it does not expose source name information so `federated_source` was added on top of `federated` field.
1. Alternative approach considered was replacing `dataset_name` (`TableReference`) with full `Dataset` instance that can be used to retrieve both name and source but in several places like tests and internal logic we construct Refresh/AcceleratedTable w/o using dataset, passing source parameter only worked better.

### DIFF
--- a/crates/runtime/src/accelerated_table.rs
+++ b/crates/runtime/src/accelerated_table.rs
@@ -205,6 +205,7 @@ pub struct Builder {
     runtime_status: Arc<status::RuntimeStatus>,
     dataset_name: TableReference,
     federated: Arc<FederatedTable>,
+    federated_source: String,
     accelerator: Arc<dyn TableProvider>,
     refresh: refresh::Refresh,
     retention: Option<Retention>,
@@ -224,6 +225,7 @@ impl Builder {
         runtime_status: Arc<status::RuntimeStatus>,
         dataset_name: TableReference,
         federated: Arc<FederatedTable>,
+        federated_source: String,
         accelerator: Arc<dyn TableProvider>,
         refresh: refresh::Refresh,
     ) -> Self {
@@ -231,6 +233,7 @@ impl Builder {
             runtime_status,
             dataset_name,
             federated,
+            federated_source,
             accelerator,
             refresh,
             retention: None,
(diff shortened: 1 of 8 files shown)


peasee (commented):
Could we reuse [`ConnectorComponent`](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/dataconnector.rs#L483) instead of storing `federated_source` as `String`?

E.g. update to use `federated_source: ConnectorComponent`

It's already supplied as part of connector parameters, so it would be good if we can reuse the `Display` traits from that.

lukekim (approved):

Sevenannn (approved):

sgrebnov:
Yeah, as mentioned in PR description/concerns, one of approaches considered was using full Dataset reference which can be represented as `ConnectorComponent::Dataset`. So that it can be used to retrieve both name and source (we can also benefit from implemented Display trait) but in several places like tests and internal logic we construct Refresh/AcceleratedTable w/o using Dataset so I found that using pure source parameter worked better.

> Could we reuse [`ConnectorComponent`](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/dataconnector.rs#L483) instead of storing `federated_source` as `String`?
> 
> E.g. update to use `federated_source: ConnectorComponent`
> 
> It's already supplied as part of connector parameters, so it would be good if we can reuse the `Display` traits from that.

PR #4577, opened 2025-01-30
fix: Retry flaky tests

peasee:
# 🗣 Description

* Adds a `.scope_retry` to `RequestContext`, to retry the provided function a certain number of times until completion or failure
* Applies it to the flaky DuckDB and MySQL federation integration tests.

### DIFF
--- a/crates/runtime/src/request/context.rs
+++ b/crates/runtime/src/request/context.rs
@@ -101,6 +101,27 @@ impl RequestContext {
         REQUEST_CONTEXT.scope(self, f).await
     }
 
+    /// Retries the provided future from the closure `r` times until it fails or succeeds.
+    pub async fn scope_retry<F, Fut, T, E>(self: Arc<Self>, r: u16, f: F) -> Fut::Output
+    where
+        F: Fn() -> Fut,
+        Fut: Future<Output = Result<T, E>>,
+    {
+        let mut try_count = 0;
+        loop {
+            let fut = f();
+            match REQUEST_CONTEXT.scope(Arc::clone(&self), fut).await {
+                Ok(result) => return Ok(result),
+                Err(e) => {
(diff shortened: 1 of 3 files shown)


sgrebnov (dismissed):
@peasee - do we understand why those tests are flaky, do we need to improve underlaying logic instead?

Retry logic looks great but can we add some clarity on flakiness before, otherwise we can hide potential issues

peasee:
> @peasee - do we understand why those tests are flaky, do we need to improve underlaying logic instead?
> 
> Retry logic looks great but can we add some clarity on flakiness before, otherwise we can hide potential issues

They're both timing related - sometimes, MySQL takes too long to get ready, and the DuckDB test is a race condition running a query while loading because we forcibly clear accelerators and re-register them without waiting long enough for acceleration to finish again: https://github.com/spiceai/spiceai/pull/4577/files#diff-0bd7cdb0540a7773f191a561739e75edd9dba402f6e1b568aa11e8900befc828L95-L97

peasee:
> > @peasee - do we understand why those tests are flaky, do we need to improve underlaying logic instead?
> > Retry logic looks great but can we add some clarity on flakiness before, otherwise we can hide potential issues
> 
> They're both timing related - sometimes, MySQL takes too long to get ready, and the DuckDB test is a race condition running a query while loading because we forcibly clear accelerators and re-register them without waiting long enough for acceleration to finish again: https://github.com/spiceai/spiceai/pull/4577/files#diff-0bd7cdb0540a7773f191a561739e75edd9dba402f6e1b568aa11e8900befc828L95-L97

To extend on this, I've also updated increasing the wait timers for these 2 tests. I've left the retry behavior there in case the waits still aren't long enough.

sgrebnov (approved):

Issue #4035, opened 2024-12-30
Enhancement: Spice.ai Cloud Platform Data Connector 1.0 Stable


phillipleblanc:
## Goal-State/What/Result

The Spice Cloud Data Connector is released at [Stable quality](https://github.com/spiceai/spiceai/blob/trunk/docs/criteria/connectors/stable.md).

## Why/Purpose

As part of our stated roadmap.

## By When

**Issue/Spec written and reviewed:** 30th Jan, 2025
**Done-Done:** 14th July, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### All Connectors

These requirements are imposed on every connector, regardless of the selected features/tests in the above table.

The RC release criteria expand on and require that all [Beta release criteria](./beta.md) continue to pass for the connector.

- [x] All [Beta release criteria](./beta.md) pass for this connector.
- [x] For exceptions where a [Core Connector Data Type](../definitions.md) is not supported, print a warning and ignore the column instead of crashing or failing the query.
- [x] ~~All known [Major Bugs](../definitions.md) are resolved.~~

#### Documentation

Documentation criteria should be re-checked on every release, to ensure the documentation is still accurate for the connector.

- [x] Documentation includes all information and steps for a user to set up the connector.
- [x] Documentation includes any mappings from [Core Connector Data Types](../definitions.md) to [Core Arrow Data Types](../definitions.md) types.
- [x] Documentation includes all known issues/limitations for the connector.
- [x] Documentation includes any exceptions made to allow this connector to reach RC quality (e.g. if a particular data type cannot be supported by the connector).
- [x] The connector has an easy to follow cookbook recipe.
- [x] The connector status is updated in the table of connectors in [spiceai/docs](https://github.com/spiceai/docs).

### Conditional Criteria

The following features/tests are dependent on the required features/tests for the specified connector, from the [Connector criteria table](#rc-release-criteria)

[...]

PR #4579, opened 2025-01-30
feat: add schema inference for the Spice.ai Data Connector

peasee:
# 🗣 Description

* Adds native schema inference for the Spice.ai Data Connector, re-using the capabilities from the Catalog Connector to retrieve schema from the REST API.
* If a `NamespaceIdent` or `TableIdent` can't be inferred from the dataset path, schema inference falls back to using data-query schema mode.
* Prefers the schema from `dataset.schema()`, if one exists. This maintains any upstream schemas, like those supplied by the Catalog Connector.

## 🔨 Related Issues

* Part of #4035

### DIFF
--- a/crates/runtime/src/dataconnector/spiceai.rs
+++ b/crates/runtime/src/dataconnector/spiceai.rs
@@ -297,7 +297,7 @@ impl DataConnector for SpiceAI {
                     })?,
             ),
             (None, None) => {
-                tracing::warn!("Could not retrieve schema in advance for Spice.ai dataset.\nSchema will be retrieved on first read.");
+                tracing::warn!("Could not retrieve schema in advance for Spice.ai dataset.\nSchema will be retrieved by querying the dataset.");
                 None
             }
         };
--- a/crates/runtime/src/dataconnector/spiceai.rs
+++ b/crates/runtime/src/dataconnector/spiceai.rs
@@ -288,7 +288,7 @@ impl DataConnector for SpiceAI {
         })?;
 
         let schema = match (dataset.schema(), table_ident) {
-            (Some(schema), Some(_) | None) => Some(schema),
+            (Some(schema), _) => Some(schema),
             (None, Some(table_ident)) => Some(
                 self.catalog
                     .get_table_schema(&table_ident)
@@ -543,6 +543,7 @@ fn get_dataset_ident(dataset_path: &SpiceAIDatasetPath) -> Result<Option<TableId
 #[cfg(test)]
 mod tests {
     use super::*;
+    use datafusion::sql::TableReference;
 
     #[test]
     #[allow(clippy::too_many_lines)]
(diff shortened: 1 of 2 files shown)


crates/runtime/src/dataconnector/spiceai.rs
```diff
@@ -144,6 +159,21 @@ impl SpiceAIFactory {
...
+
+    #[must_use]
+    fn create_rest_catalog_client(params: &Parameters) -> RestCatalog {
+        let mut props = HashMap::new();
+        if let ExposedParamLookup::Present(api_key) = params.get("api_key").expose() {
+            props.insert("token".to_string(), api_key.to_string());
+        };
+
+        let catalog_config = RestCatalogConfig::builder()
+            .uri("https://data.spiceai.io".to_string())
```
Sevenannn: Spice ai connector supports endpoint parameter and the endpoint could be something other than "https://data.spiceai.io", we could make this an uwrap or default for endpoint param

Sevenannn:
I wonder if this is going to be more efficient, under this change each spiceai connector would now contains a rest catalog component, which could be extra overhead?

peasee:
> I wonder if this is going to be more efficient, under this change each spiceai connector would now contains a rest catalog component, which could be extra overhead?

I'm not sure I understand how it is extra overhead. The catalog component isn't used unless called, which we only call once to collect dataset schema?

crates/runtime/src/dataconnector/spiceai.rs
peasee: The endpoint parameter is the flight endpoint, not the HTTP endpoint.

crates/runtime/src/dataconnector/spiceai.rs
```diff
@@ -237,9 +279,39 @@ impl DataConnector for SpiceAI {
...
+        let table_ident = get_dataset_ident(&dataset_path).map_err(|err| {
+            DataConnectorError::UnableToGetReadProvider {
+                dataconnector: "spice.ai".to_string(),
+                connector_component: ConnectorComponent::from(dataset),
+                source: err.into(),
+            }
+        })?;
+
+        let schema = match (dataset.schema(), table_ident) {
+            (Some(schema), Some(_) | None) => Some(schema),
```
sgrebnov: Not sure 
```suggestion
            (Some(schema), _) => Some(schema),
```

crates/runtime/src/dataconnector/spiceai.rs
```diff
@@ -417,6 +489,57 @@ impl CommitChange for SpiceAIChangeCommiter {
     }
 }
 
+fn get_dataset_ident(dataset_path: &SpiceAIDatasetPath) -> Result<Option<TableIdent>, Error> {
```
sgrebnov, line 492: Do we need some unit tests for this logic?

sgrebnov (dismissed):

crates/runtime/src/dataconnector/spiceai.rs
peasee: `_` would match every capture group, but I still want to make `Some(ident)` in a different arm 😃

sgrebnov (dismissed):

crates/runtime/src/dataconnector/spiceai.rs
```diff
@@ -237,9 +279,39 @@ impl DataConnector for SpiceAI {
...
+                        connector_component: ConnectorComponent::from(dataset),
+                        source: Error::UnableToGetSchema {
+                            table: table_ident.name.to_string(),
+                            source: err.into(),
+                        }
+                        .into(),
+                    })?,
+            ),
+            (None, None) => {
+                tracing::warn!("Could not retrieve schema in advance for Spice.ai dataset.\nSchema will be retrieved by querying the dataset.");
```
phillipleblanc: Let's mark this down to a debug trace

phillipleblanc (dismissed):
Looks good once UX is reviewed

phillipleblanc (approved):

PR #4581, opened 2025-01-30
Test running test on macos runner

Sevenannn:
# 🗣 Description

## ⛓️‍💥 Breaking Change


Sevenannn:
Temporarily close since spiceai-macos runners cannot download aritifacts as spiceai-runners, will need alternative ways of storing mac artifacts to be used in model tests.

PR #4601, opened 2025-02-03
Improve UX of downloading GGUF from HF

Jeadie:
# 🗣 Description
 - Add progress bar to `spiced` output when downloading GGUF model from HF. This requires loading GGUF HF models in the similar manner as other HF models (currently we load in a entirely separate function). 
 - Also add log for when we have an explicit HF token provided by user.

### DIFF
--- a/crates/llms/src/chat/mistral.rs
+++ b/crates/llms/src/chat/mistral.rs
@@ -304,38 +304,38 @@ impl MistralLlama {
     ) -> Result<Self> {
         let model_parts: Vec<&str> = model_id.split(':').collect();
 
-        let loader: Result<Box<dyn Loader>> = match gguf_filename {
-            // Loading the GGUF directly (as if it is a quantized model, although it need not be quantized).
-            Some(gguf) => Ok(GGUFLoaderBuilder::new(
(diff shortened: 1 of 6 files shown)


crates/llms/src/chat/mistral.rs
```diff
@@ -300,32 +300,49 @@ impl MistralLlama {
...
 
-        let builder = NormalLoaderBuilder::new(
-            mistralrs::NormalSpecificConfig::default(),
-            None,
-            None,
-            Some(model_parts[0].to_string()),
-        );
         let device = Self::get_device();
         let token_source = hf_token_literal.map_or(TokenSource::CacheToken, |secret| {
+            tracing::info!("Explicitly given a Huggingface token. Using instead of any system defaults (if any).");
```
peasee:
```suggestion
            tracing::debug!("A HuggingFace token was specified in parameters. The specified token will be used instead of any system/environment defaults.");
```

nit: this could be a debug log. wording.

peasee (dismissed):

phillipleblanc (dismissed):

peasee (approved):

PR #4648, opened 2025-02-04
Run `build_and_release_cuda.yml` when `crates/llms/Cargo.toml` changes

Jeadie:
# 🗣 Description
This will help ensure downstream changes do not break CUDA related builds.

### DIFF
--- a/.github/workflows/build_and_release_cuda.yml
+++ b/.github/workflows/build_and_release_cuda.yml
@@ -1,6 +1,11 @@
 name: build_and_release_cuda
 
 on:
+  pull_request:
+    branches:
+      - trunk
+    paths:
+      - "crates/llms/Cargo.toml"
   push:
     tags:
       - v*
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_release_cuda.yml
```diff
@@ -1,6 +1,11 @@
 name: build_and_release_cuda
 
 on:
+  pull_request:
```
sgrebnov, line 4: We might filter compute_cap in this case and run only single compute_cap

sgrebnov (dismissed):

.github/workflows/build_and_release_cuda.yml
Jeadie: I thought of that, wasn't sure how?

sgrebnov: There is filtering example here

https://github.com/spiceai/spiceai/pull/4648/files#diff-f6c5bf35bb6ac7ed7eafdef419821dd828e5c80446f5f1c55318ecc4b9c30163R144

peasee (approved):

phillipleblanc (approved):

Issue #4521, opened 2025-01-27
Enhancement: Perplexity.ai Connector


lukekim:
## Goal-State/What/Result
Perplexity can be used as both 
1. An LLM component (i.e. defined in a spicepod under `models`
2. A search engine tool that provides summary and citations for a given question. The paradigm for Perplexity as a search engine tool should be extensible to similar components (e.g. exa.ai, tavily.com, DuckDuckGo). 

## Why/Purpose
Perplexity is a useful model LLM interactions heavily relying on public context (citations), but may not be the only or best search engine for a given use case/task .

## By When

**Issue/Spec written and reviewed:** 4th February
**Done-Done:** 7th February

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [ ] Release Notes prepared

#### Perplexity model connector,
 - via Sonar [API](https://docs.perplexity.ai/api-reference/chat-completions).
   - The [response citations](https://docs.perplexity.ai/api-reference/chat-completions#response-citations) should be added as internal tool calls in the tracing (i.e. `runtime.task_history`).

```yaml
models
 - from: perplexity:sonar # default is sonar
   name: perp_llm
   params:
     perplexity_api_key: ${ secrets:SPICE_PERPLEXITY_API_KEY }
     perplexity_search_recency_filter: month
```
- `params.perplexity_*` provide model overrides, including non-openai compatible params (e.g. `search_recency_filter`).

#### A standard interface for search engine tools
Possible input Schema
```json
{
  "query": String,
  "limit": uint,
}
```

Response Schema
```json
{ 
  "summary": String | optional,
  "results": [
    {
      "url": String,
      "title": String | optional,
      "type": Webpage | Image,
      "content": String | optional
    }
 ] 
}
```
- This enhancement will only include perplexity as a search engine. 

--- 

Spicepod
```yaml
tools:

 - from: search:perp_llm # assumes `model` with `name: perp_llm` is defined. 
 - from: search:perplexity
    name: perp_tool
    params:
      model: sonar
      perplexity_api_key: ${ secrets:SPICE_PERPLEXITY_API_KEY }
      perplexity_search_recency_filter: month
[...]

PR #4673, opened 2025-02-05
Perplexity Sonar LLM component.

Jeadie:
# 🗣 Description
- New LLM model: `perplexity`, powered by [sonar](https://docs.perplexity.ai/).
```yaml
models:
  - name: webs
    from: perplexity:sonar
    params:
      perplexity_auth_token: ${ secrets:SPICE_PERPLEXITY_AUTH_TOKEN }
      # Non openai perplexity parameters can be added here (prefixed by `perplexity_`).
      perplexity_search_recency_filter: hour
      # Still supports openai parameters.
      openai_temperature: 0.3141595
```

Two refactors to improve LLM more generally:
 - Config for hosted LLMs. Anthropic, and now perplexity, use `HostedModelConfig` (not `AnthropicConfig`).
 - Refactor OpenAI parameter overrides.
   - Moved `ChatWrapper` (the struct handling overrides) to separate file.
   - Emit warning error if the conversion of the override from string to the field's type does not succeed.
   - At startup, check if overrides would work (i.e. override parameters in a default request, check for errors). 

## 🔨 Related Issues
- #4521 

## 📄 Documentation Updates
 - [x] https://github.com/spiceai/docs/pull/838

## Concerns
 - citations are not emitted via LLM. Can be returned from tool (in progress).


crates/runtime/src/model/chat.rs
```diff
@@ -325,252 +342,3 @@ fn file(
...
-                }
-                None => Some(ChatCompletionStreamOptions {
-                    include_usage: true,
-                }),
-            };
-        }
-        req
-    }
-
-    /// For [`None`] valued fields in a [`CreateChatCompletionRequest`], if the chat model has non-`None` defaults, use those instead.
```
Jeadie, line 393: Moved this to `crates/runtime/src/model/wrapper.rs`.

phillipleblanc (approved):

PR #4689, opened 2025-02-06
Add GH Workflow to run Text to SQL benchmark

sgrebnov:
# 🗣 Description

Add GH Workflow to run Text to SQL evals:
  -  models: `gpt-4o-mini` (openai) and `llama3-3B-Instruct` (HF local)
  - evaluation result is traced directly to console, writing results to a database will be added next

Example test run: https://github.com/spiceai/spiceai/actions/runs/13187353898/job/36813014829

## ⛓️‍💥 Breaking Change


.github/workflows/ai_benchmark_nsql.yml
```diff
@@ -0,0 +1,154 @@
...
+          chmod +x ./spiced
+
+      - name: Run tpch_nsql evals
+        env:
+          SPICE_OPENAI_API_KEY: ${{ secrets.SPICE_SECRET_OPENAI_API_KEY }}
+          SPICE_HF_TOKEN: ${{ secrets.SPICE_SECRET_HF_TOKEN }}
+        run: |
+          ./testoperator run evals \
+            -s $GITHUB_WORKSPACE/spiced \
+            --eval tpch_nsql \
```
Jeadie: I think this GH action could be easily broadened to input both a model option, and an eval task (instead of hardcoded `tpch_nsql`).

sgrebnov: Simplified so that model and eval are optional and the testoperator will default to spicepod definition (use first found model and eval if not specified). GH workflow has been updated so that more eval configurations could be added (combination of spicepod + data folder)

Sevenannn (approved):

PR #4708, opened 2025-02-09
Move `spice.ai` benchmark tests to testoperator

phillipleblanc:
# 🗣 Description

Implements the plumbing necessary to start running the `spice.ai` TPCH/TPCDS benchmark tests via the testoperator.

It is working locally for me, and I've generated the snapshots - the next piece to get this fully working is to integrate it into CI and the cron dispatch.

I needed to make several changes to how the testoperator benchmark test works and emits metrics - namely aligning to the existing benchmark schema, and adding a `spiceai` r/w dataset to write the final benchmark metrics.

One other "breaking" change for the test operator is that the dataset tests now require a "--connector-name" parameter to indicate which connector is being tested. This was used in the metrics that were emitted.

The other main change to the testoperator is that queries executed via the `SpiceTest` runner will now execute the same query from a query set back-to-back for the `EndCondition::QuerySetCompleted` - i.e. running with `EndCondition::QuerySetCompleted(5)` runs each query from the set in a row (Q1 Q1 Q1 Q2 Q2 Q2) instead of (Q1 Q2 Q1 Q2). This aligns with the behavior of the existing benchmarks.

## 🔨 Related Issues

Part of #4114


crates/test-framework/src/snapshot/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+limitations under the License.
+*/
+
+use std::panic;
+
+use flight_client::FlightClient;
+
+use crate::flight::query_to_batches;
+
+pub async fn record_explain_plan(
```
peasee, line 23: I think some snapshots also snapshot the 10 rows of result contents, which we'll need to add

peasee (approved):

crates/test-framework/src/metrics/mod.rs
```diff
@@ -46,38 +55,53 @@ pub enum QueryStatus {
...
-    pub run_count: usize,
+    pub started_at: usize,
+    pub finished_at: usize,
+    pub min_duration_ms: i64,
+    pub max_duration_ms: i64,
+    pub iterations: usize,
+    pub median_duration_ms: i64,
+    pub percentile_99_duration_ms: i64,
+    pub percentile_95_duration_ms: i64,
+    pub percentile_90_duration_ms: i64,
```
Sevenannn, line 75: Do we want to switch to use the percentile metrics for benchmarks at some point in the future if we intend to run more iterations for benchmark? I recall some discussion on the max / min duration not being the best metric for benchmark, but it won't make any difference as for now since there's only 5 iterations and the percentiles 99 95 90 will simply be the max value.

PR #4717, opened 2025-02-10
Fix bug: Ensure refresh only retrieves data within the retention period

sgrebnov:
# 🗣 Description

Currently, initial refresh fetches all data even if retention period is defined. This PR fixes the bug by ensuring that the refresh only retrieves data within the retention period

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

If both `refresh_data_window` and `retention_period` are defined, we technically need only data within min of the two. Current implementation always use `refresh_data_window` if defined as it is explicit configuration for refresh defined by user.

### DIFF
--- a/crates/runtime/src/datafusion.rs
+++ b/crates/runtime/src/datafusion.rs
@@ -799,7 +799,7 @@ impl DataFusion {
             refresh = refresh.append_overlap(append_overlap);
         }
 
-        // we must not fetch data longer than explicitly set refresh data window or retention period
+        // we must not fetch data older than the explicitly set refresh data window or retention period
         let refresh_data_window = dataset.refresh_data_window().or(dataset.retention_period());
 
         if let Some(refresh_data_window) = refresh_data_window {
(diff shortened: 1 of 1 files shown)


crates/runtime/src/datafusion.rs
```diff
@@ -798,8 +798,12 @@ impl DataFusion {
...
             refresh = refresh.append_overlap(append_overlap);
         }
-        if let Some(refresh_data_window) = dataset.refresh_data_window() {
-            refresh = refresh.period(refresh_data_window);
+
+        // we must not fetch data longer than explicitly set refresh data window or retention period
+        let refresh_data_window = dataset.refresh_data_window().or(dataset.retention_period());
+
+        if let Some(refresh_data_windows) = refresh_data_window {
+            refresh = refresh.period(refresh_data_windows);
```
```suggestion
        if let Some(refresh_data_window) = refresh_data_window {
            refresh = refresh.period(refresh_data_window);
```

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #4719, opened 2025-02-10
Implement `unsupported_type_action: string` for Postgres JSONB support

### LINKED ISSUE #4714: Enhancement: `unsupported_type_action: string` to attempt to convert invalid types to Strings (i.e. JSONB)
## Goal-State/What/Result
Deprecate the existing `invalid_type_action` and rename to `unsupported_type_action` to better reflect what it is doing. If both `invalid_type_action` and `unsupported_type_action` are set, prefer `unsupported_type_action`. Remove all documentation for `invalid_type_action`.

Add a new `string` action to the existing `unsupported_type_action` enum that will attempt to convert unsupported types to strings, specifically to support Postgres JSONB columns until Apache Arrow Variant type support is improved. This will at least allow users with JSONB data to be able to query it via a string representation rather than failing or ignoring it.

```yaml
datasets:
  - from: postgres:my_table
    name: my_table
    unsupported_type_action: string
    params:
      pg_host: <>
```

## Why/Purpose
Currently, JSONB columns from Postgres cannot be properly decoded into Apache Arrow Variant types due to limited support. Adding a `string` conversion action provides an immediate solution to handle this data type while maintaining compatibility and avoids making this a "default" that would require a breaking change once we properly support this in the future.

## By When
**Issue/Spec written and reviewed:** February 10, 2025
**Done-Done:** February 10, 2025

## Done-Done
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [ ] Tests
- [ ] Documentation
- [ ] Release Notes prepared

## The Algorithm
  - Verified need for string conversion vs. waiting for Arrow Variant support
  - Confirmed JSONB is primary use case
- [x] Delete (Scope) any part you can
  - Focused solely on string conversion, no additional type conversions
  - Focus only on Postgres
- [x] Simplify
  - Single conversion path to string
  - No complex type inference
- [x] Break down into smaller iterations/milestones
  1. Implement basic String action
  2. Add JSONB specific handling
  3. Add tests and documentation
- [x] Opportunities for automation
  - Add automated tests for type conversion
  - Create benchmarks for performance testing

## Specification
1. Add `String` as a new variant to the `InvalidTypeAction` enum
2. For the connectors that support `InvalidTypeAction` (currently DuckDB/Postgres), add support for the String. For DuckDB, map `String` to the `Error` variant for now (since we won't actually support converting any unknown types to String).
3. Ensure that the Postgres type conversion code correctly converts JSONB to String when this flag is enabled.

### Security Review
N/A

## How/Plan
1. Add `String` variant to `InvalidTypeAction`
2. Implement string conversion logic
3. Add JSONB specific handling
4. Update tests and documentation
5. Add performance benchmarks
6. Create cookbook examples

### Implementation (Sub-Issues)
- [ ] Add String variant to InvalidTypeAction enum
- [ ] Implement string conversion logic
- [ ] Add JSONB specific handling
- [ ] Test with `parse_json`
- [ ] Update type conversion pipeline
- [ ] Add documentation and examples

### QA
- [ ] Unit tests for string conversion
- [ ] Integration tests with Postgres JSONB
- [ ] Performance benchmarks
- [ ] Documentation review

## Release Notes
New InvalidTypeAction - String Conversion: Added a new `string` action to `invalid_type_action` that converts invalid types to string representation, particularly useful for handling Postgres JSONB columns until full Apache Arrow Variant type support is available.

Example configuration:
```yaml
datasets:
  - from: postgres:my_table
    name: my_table
    invalid_type_action: string
    params:
      pg_host: <>
```

This allows handling of JSONB columns by converting their content to string representation rather than raising errors or warnings.

phillipleblanc:
# 🗣 Description

Depends on:
- https://github.com/datafusion-contrib/datafusion-table-providers/pull/234
- https://github.com/datafusion-contrib/datafusion-table-providers/pull/235

Deprecates `invalid_type_action` in favor of `unsupported_type_action` which supports all previous values (`error`, `warn`, `ignore`) and adds support for a new `unsupported_type_action: string`. This new value will attempt to convert invalid types to strings, which currently only works for JSONB columns in Postgres.

```yaml
datasets:
  - from: postgres:table_with_jsonb
    name: test
    unsupported_type_action: string
    params:
      pg_host: <>
```

The old `invalid_type_action` will continue to be supported in Spicepod definitions, but they will log a warning if used:

i.e.:

```yaml
datasets:
  - from: postgres:table_with_jsonb
    name: test
    invalid_type_action: ignore
```

Output:
```console
2025-02-10T07:40:21.050061Z  WARN spicepod::component::dataset: test: `dataset.invalid_type_action` is deprecated, use `dataset.unsupported_type_action` instead
```

## 🔨 Related Issues

Closes #4714

## ⛓️‍💥 Breaking Change

Not a breaking change, but `invalid_type_action` is now deprecated.

## 📄 Documentation Updates

https://github.com/spiceai/docs/pull/857

## 🤔 Concerns

Unfortunately, while the JSONB columns are correctly being returned as strings, attempting to use any of the functions from `datafusion-functions-json` will fail because the federation planner is trying to federate those DataFusion specific function calls, which do not exist in Postgres.

~~I'm looking into if this is an easy fix.~~
Unfortunately not, I've created this issue to track what needs to be done: https://github.com/spiceai/spiceai/issues/4720

i.e.:

```bash
sql> select * from test;
+---------------+----+
| the_jsonb_col | id |
+---------------+----+
| {"foo":"bar"} | 1  |
+---------------+----+

Time: 0.019214708 seconds. 1 rows.
sql> select the_jsonb_col->'foo' from test;
Query Error: Failed to execute query.
Query execution failed.
db error: ERROR: function json_get(jsonb, unknown) does not exist
HINT: No function matches the given name and argument types. You might need to add explicit type casts.
For details, refer to the PostgreSQL manual: https://www.postgresql.org/docs/17/index.html
```


crates/runtime/src/dataconnector/github/github_app_token_provider.rs
```diff
@@ -285,94 +285,4 @@ mod tests {
             .expect("Failed to get token");
         assert_eq!(token, "token_2");
     }
-
```
phillipleblanc, line 288: This was a flaky test, I've removed it.

peasee (approved):

PR #4745, opened 2025-02-11
Update SECURITY.md with 1.0.3

phillipleblanc:
# 🗣 Description

Updates `SECURITY.md` with `1.0.3` and also adds in the release notes to trunk.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -8,6 +8,7 @@ In the latest major version, the last 2 minor versions are supported for securit
 
 | Version        | Supported          |
 | -------------- | ------------------ |
+| 1.0.3          | :white_check_mark: |
 | 1.0.2          | :white_check_mark: |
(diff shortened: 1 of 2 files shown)


ewgenius (approved):

SECURITY.md
```diff
@@ -8,6 +8,7 @@ In the latest major version, the last 2 minor versions are supported for securit
```
Sevenannn, line 8: Do we need to update the number of versions supported for security updates?

Sevenannn (approved):

Jeadie (approved):

SECURITY.md
phillipleblanc: Minor versions is still accurate - this is still patch versions, once we release `1.1` we can probably say all `1.0.x` are covered.

Issue #3919, opened 2024-12-18
`spice version` command suggests update and displays same output even after running `spice upgrade`


Advayp:
## Describe the bug

After running `spice upgrade`, which should update the Spice CLI to the latest version, `spice version` still suggests that a new version is available.

`spice upgrade` doesn't appear to be functioning properly, as `spice version` isn't changing its output after what seems to be a successful upgrade. This is either a problem with `spice upgrade` not functioning correctly or `spice version` misreporting the current version of the CLI/Runtime.

## To Reproduce

1. Install an older version of the Spice CLI. The version of Spice CLI I first saw the bug in was `v0.18.3-beta.`
2. Run `spice version`, ensuring it suggests an update using `spice upgrade.`
3. Run `spice upgrade.`
4. Run `spice version` again, noting that the output hasn't changed from Step 2.

## Expected behavior

`spice upgrade`: Should update the end-user's version of Spice CLI to the latest one available.
`spice version`: Should accurately report the end user's current version of Spice CLI.

## Screenshots

Initial run of `spice version`:
![Image](https://github.com/user-attachments/assets/00e46d00-56da-4c0d-9abe-7115a30148d1)

Running `spice upgrade,` which indicates that it has upgraded to the latest version of Spice *successfully*:
![Image](https://github.com/user-attachments/assets/0d4829a4-107d-4f75-88e9-755bb295fabd)

Running `spice version` again, which has the same output from earlier: 
![Image](https://github.com/user-attachments/assets/57daf801-a329-4e4b-994f-999106170ff4)

All three were run back to back like so:
![Image](https://github.com/user-attachments/assets/fa4005e9-23bf-452f-a8c2-c0d2d0b34d35)

PR #4746, opened 2025-02-11
Prompt user to upgrade through brew / do another clean install when spice is installed through homebrew / at non-standard path

Sevenannn:
# 🗣 Description

The `spice upgrade` command only works for spice directly downloaded through `curl https://install.spiceai.org | /bin/bash`. When spice is downloaded through homebrew, running `spice upgrade` will upgrade the binary in the `spicehomedir/bin`, but not the brew binary, causing issue as if an upgrade doesn't upgrade CLI at all (see the screenshot below)

image

This PR implemented a mechanism to check whether the cli is downloaded from homebrew. If so, it prompts user to upgrade through brew instead of running spice command.

After:
- Spice downloaded from homebrew:
image

- Spice downloaded at a non-standard path:
image

## 🔨 Related Issues

https://github.com/spiceai/spiceai/issues/3919

## ⛓️‍💥 Breaking Change


bin/spice/pkg/context/context.go
```diff
@@ -394,6 +394,19 @@ func (c *RuntimeContext) SetHttpEndpoint(endpoint string) {
 	c.httpEndpoint = endpoint
 }
 
+func (c *RuntimeContext) IsDownloadedFromHomebrew() bool {
```
```suggestion
func (c *RuntimeContext) WasInstalledFromHomebrew() bool {
```

bin/spice/pkg/context/context.go
```diff
@@ -394,6 +394,19 @@ func (c *RuntimeContext) SetHttpEndpoint(endpoint string) {
...
+	executableDir, err := os.Executable()
+	if err != nil {
+		slog.Error("getting path", "error", err)
+		return false
+	}
+
+	spiceBinDir := filepath.Join(c.SpiceRuntimeDir(), "bin")
+	releaseFilePath := filepath.Join(spiceBinDir, constants.SpiceCliFilename)
+
+	return executableDir != releaseFilePath
```
phillipleblanc: This should also check that `homebrew` is in the path - otherwise this would incorrectly report it was installed from homebrew when it could have just been installed into a non-standard path.

phillipleblanc: If we detect a non-standard path that isn't homebrew, the message should be more generic, i.e. "Spice cannot upgrade because the executable is in a non-standard location: {location}."

bin/spice/pkg/context/context.go
lukekim: I would do `IsHomebrewInstall`

Sevenannn: Update to another function that detects the brew / standard / nonstandard enum

phillipleblanc (approved):

PR #4761, opened 2025-02-12
Always download spice runtime version matched with spice cli version

### LINKED ISSUE #4762: Spice CLI shouldn't download a runtime version higher than the cli version

Sevenannn:
# 🗣 Description

When downloading runtime when runtime is previously not installed, download a runtime version matched with cli version instead of the latest version. When upgrading runtime, upgrade to runtime version matched with cli version.

This PR eliminates all path for downloading the latest runtime version, only keep the path for downloading the matching runtime version. The runtime would only upgrade when it has a version not matching with the cli version, not when it's not the latest version. This ensures the runtime version would always match the cli version.

After:
- spice run
image
- spice install
image
- spice upgrade (no upgrade when matching with cli version)
image

The behavior of downloading matching version is verified in e2e test: https://github.com/spiceai/spiceai/actions/runs/13326091615/job/37219679263
## 🔨 Related Issues

Brew spice behavior tested in e2e:
- Upgrade cli: https://github.com/spiceai/spiceai/actions/runs/13326517397/job/37220915988
- Upgrade runtime: https://github.com/spiceai/spiceai/actions/runs/13326584566/job/37221115722

- Close: https://github.com/spiceai/spiceai/issues/4762

## ⛓️‍💥 Breaking Change


bin/spice/pkg/context/context.go
```diff
@@ -234,6 +234,39 @@ func (c *RuntimeContext) IsRuntimeInstallRequired() bool {
 	return errors.Is(err, os.ErrNotExist)
 }
 
+func (c *RuntimeContext) InstallMatchingRuntime(cliVersion string, flavor constants.Flavor, allowAccelerator bool) error {
```
phillipleblanc: This is basically a duplicate of `InstallOrUpgradeRuntime`, with the only difference being the function that gets the release. This can be refactored to have the common code isolated. But also, in what scenario would we not want to get the runtime version that matches the CLI?

Sevenannn: Good point. In fact I can't think of a scenario where we wouldn't want a runtime version matches the CLI. we should eliminate the download latest runtime functions if the assumption is that runtime downloaded should always match the cli version.

bin/spice/cmd/upgrade.go
```diff
@@ -101,7 +101,7 @@ spice upgrade
 		}
 
 		if runtimeUpgradeRequired == "" {
-			slog.Info(fmt.Sprintf("Using version %s. No upgrade required.", currentVersion))
+			slog.Info(fmt.Sprintf("Using version %s matching with CLI version %s. No upgrade required.", currentVersion, version.Version()))
```
phillipleblanc: I don't think this change is required

bin/spice/cmd/upgrade.go
```diff
@@ -188,7 +188,7 @@ func upgradeCli(force bool, rtcontext *context.RuntimeContext) bool {
 	assetName := github.GetAssetName(constants.SpiceCliFilename)
 	spiceBinDir := filepath.Join(rtcontext.SpiceRuntimeDir(), "bin")
 
-	slog.Info("Upgrading the Spice.ai CLI ...")
+	slog.Info(fmt.Sprintf("Found latest CLI version %s, upgrading the Spice.ai CLI ...", release.TagName))
```
phillipleblanc:
```suggestion
	slog.Info(fmt.Sprintf("Found version %s, upgrading the Spice.ai CLI ...", release.TagName))
```

bin/spice/pkg/github/release.go
```diff
@@ -76,6 +76,28 @@ func (r *RepoRelease) HasAsset(assetName string) bool {
...
+		return nil, err
+	}
+
+	for _, release := range githubRepoReleases {
+		if release.TagName == version {
+			return &release, nil
+		}
+	}
+
+	return nil, fmt.Errorf("No Matching release is founded for version %s", version)
```
phillipleblanc:
```suggestion
	return nil, fmt.Errorf("No matching release is found for version %s", version)
```

bin/spice/pkg/context/context.go
```diff
@@ -266,6 +266,7 @@ func (c *RuntimeContext) InstallOrUpgradeRuntime(flavor constants.Flavor, allowA
 	slog.Info(fmt.Sprintf("Spice runtime installed into %s successfully.\n", c.spiceBinDir))
 
 	return nil
+
```
```suggestion
```

phillipleblanc (dismissed):

sgrebnov (dismissed):

sgrebnov (approved):

phillipleblanc (approved):

Issue #4758, opened 2025-02-12
Enhancement: Support `Cache-Control: no-cache` to invalidate results cache


phillipleblanc:
## Goal-State/What/Result

The `Cache-Control: no-cache` header can be applied to either the Flight or HTTP Query APIs for a specific query, which will cause the runtime to re-execute that query, even if it already has a hit in the results cache. This allows the user to ensure a specific query will always fetch the latest results.

This does not affect whether a query uses an acceleration or the federated source, only the results cache.

The response from the runtime will not set a `Cache-Control` header, this is only valid on the request to the runtime.

HTTP Example:
```bash
curl -H "cache-control: no-cache" -XPOST -i http://localhost:8090/v1/sql -d 'SELECT 1'
```

Flight Example:
```rust
let descriptor = FlightDescriptor::new_cmd("SELECT 1");
let mut req = tonic::Request::new(descriptor);
req.metadata_mut()
            .insert("cache-control", "no-cache");
flight_client.do_get(req);
```

## Why/Purpose

There are use-cases where its deseriable to have more control over whether a specific query is cached, i.e. when the application knows new data was just added.

## By When

**Issue/Spec written and reviewed:** Feb 12, 2025
**Done-Done:** Feb 12, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [ ] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

When the `Cache-Control: no-cache` header is present, it will ignore the results cache for the given query. The result will still be stored in the results cache, so subsequent queries would retrieve the results of the request that specified `Cache-Control: no-cache`.
This matches the behavior of the standard HTTP header: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

We will not support any other directives for `Cache-Control`, only `no-cache`. Another interesting one that we might want to support eventually is `no-store` which tells the server not to store the results in the cache - but that is out of scope.

### Security Review

[...]

PR #4763, opened 2025-02-12
Implement `Cache-Control: no-cache` to bypass results cache.

phillipleblanc:
# 🗣 Description

Implements a new feature that supports reading `Cache-Control: no-cache` headers on the SQL query APIs to control whether the results cache is considered for the execution of a given query.

I've also added a `--cache-control` CLI argument to `spice sql` that allows specifying this behavior via CLI as well.

`spice sql --cache-control cache` (Default)
`spice sql --cache-control no-cache`

Via HTTP Endpoint:

```bash
curl -H "cache-control: no-cache" -XPOST -i http://localhost:8090/v1/sql -d 'SELECT 1'
```

## Refactoring

I've taken the liberty of refactoring the cache logic in the main query path to make it easier to understand.

Also instead of storing the cache status as an `Option<bool>`, I've refactored it to use an explicit enum with the states we expect:

```rust
pub enum QueryCacheStatus {
    CacheNotChecked,
    CacheHit,
    CacheMiss,
}
```

## 🔨 Related Issues

Part of #4758 

## 📄 Documentation Updates

Needs a docs update for explaining how to use this new feature.

## 🤔 Concerns

N/A

### DIFF
--- a/crates/runtime/src/datafusion/query.rs
+++ b/crates/runtime/src/datafusion/query.rs
@@ -148,9 +148,15 @@ impl Query {
         ))
     }
 
-    fn should_cache_results(df: &DataFusion, plan: &LogicalPlan) -> bool {
-        df.cache_provider()
-            .is_some_and(|provider| provider.cache_is_enabled_for_plan(plan))
+    fn should_cache_results(
+        df: &DataFusion,
+        plan: &LogicalPlan,
+        cache_status: QueryCacheStatus,
+    ) -> (bool, QueryCacheStatus) {
+        match df.cache_provider() {
+            Some(provider) if provider.cache_is_enabled_for_plan(plan) => (true, cache_status),
+            _ => (false, QueryCacheStatus::CacheNotChecked),
+        }
     }
 
     fn wrap_stream_with_cache(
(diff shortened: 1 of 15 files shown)


crates/runtime/src/datafusion/query.rs
```diff
@@ -92,7 +96,76 @@ macro_rules! handle_error {
...
+        plan: &LogicalPlan,
+    ) -> CacheResult {
+        let Some(cache_provider) = df.cache_provider() else {
+            return CacheResult::MissOrSkipped(tracker, QueryCacheStatus::CacheNotChecked);
+        };
+
+        // If the user requested no caching, skip the cache lookup
+        if matches!(ctx.cache_control(), CacheControl::NoCache) {
+            return CacheResult::MissOrSkipped(tracker, QueryCacheStatus::CacheNotChecked);
+        }
```
phillipleblanc, line 119: This the main logic of the PR - the rest is infrastructure/refactoring.

crates/runtime/src/request/context.rs
```diff
@@ -192,6 +200,7 @@ impl RequestContextBuilder {
             UserAgentCollection::Full => UserAgent::from_headers(headers),
             UserAgentCollection::Disabled => UserAgent::Absent,
         };
+        self.cache_control = CacheControl::from_headers(headers);
```
phillipleblanc, line 203: We're already checking the headers for every request to get things like the `baggage` and `user_agent` - so it was fairly simple to plug in and add a check for the cache control here - and we automatically get it everywhere.

Sevenannn (dismissed):

lukekim (approved):

PR #4769, opened 2025-02-13
feat: Add IMAP access token parameter

### LINKED ISSUE #4752: Add IMAP access token parameter
Add an `imap_access_token` parameter, used for supplying an OAuth access token. When an `imap_access_token` is set, it implicitly sets `imap_auth_mode` to `oauth`.

`imap_access_token` is otherwise an alias for `imap_password`.

peasee:
# 🗣 Description

* Removes `auth_mode` parameter.
* Adds `access_token` parameter. The `password` parameter uses `plain` auth mode, while `access_token` uses `oauth` mode.

## 🔨 Related Issues

* Closes #4752

### DIFF
--- a/crates/runtime/src/dataconnector/imap.rs
+++ b/crates/runtime/src/dataconnector/imap.rs
@@ -43,11 +43,9 @@ const PARAMETERS: &[ParameterSpec] = &[
         .secret()
         .description("The username to use for the IMAP connection"),
     ParameterSpec::connector("password")
-        .required()
         .secret()
         .description("The password to use for the IMAP connection"),
     ParameterSpec::connector("access_token")
-        .required()
         .secret()
         .description("The OAuth access token to use for the IMAP connection"),
     ParameterSpec::connector("host").description("The IMAP server host to connect to"),
--- a/crates/runtime/src/dataconnector/imap.rs
+++ b/crates/runtime/src/dataconnector/imap.rs
@@ -17,11 +17,12 @@ limitations under the License.
 use crate::component::dataset::Dataset;
 use async_trait::async_trait;
 use data_components::imap::{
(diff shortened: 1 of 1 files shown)


crates/runtime/src/dataconnector/imap.rs
```diff
@@ -180,6 +186,34 @@ impl ImapFactory {
...
+        username: &SecretString,
+        password: Option<&SecretString>,
+        access_token: Option<&SecretString>,
+    ) -> Result<ImapAuthMode, Box<dyn std::error::Error + Send + Sync>> {
+        match (password, access_token) {
+            (Some(_), Some(_)) => Err(DataConnectorError::InvalidConfigurationSourceOnly {
+                dataconnector: "imap".to_string(),
+                connector_component: connector_component.clone(),
+                source: Error::PasswordAndAccessTokenError.into(),
+            }
```
Sevenannn, line 199: Would it be better to default to use either password / access token to connect and warn user that both params are provided instead of error and refuse connection? Since user do provide credentials to access, it's just they provide 2 different credentials at the same time.

peasee: Both are entirely different methods of authentication, and should not be used together. There is no way to tell which authentication method is valid.

crates/runtime/src/dataconnector/imap.rs
```diff
@@ -45,6 +46,10 @@ const PARAMETERS: &[ParameterSpec] = &[
         .required()
         .secret()
         .description("The password to use for the IMAP connection"),
+    ParameterSpec::connector("access_token")
+        .required()
```
Sevenannn: If only one of password / access_token is expected, should they both be not-required?

Sevenannn (approved):

Issue #4710, opened 2025-02-09
Improve `spice trace` formatting


Jeadie:
Changes
1. Use the table format, with header names (like `spice datasets`). Keep the tree all in the first column. 
1. Remove the ( ) around duration No longer required with headers.
1. Remove the [ ] around task IDs. Too noisy.
1. Add status on the tree. E.g.
    ```
     ├── [2a8d117769157c8d] ✅ (    0.02ms) tool_use::list_datasets`
     ├── [2a8d117769157c8d] ❌ (    0.02ms) tool_use::list_datasets`
    ```
1. Add a component name after the task name. E.g. 0.2ms tool_use::websearch perp and 1c7cceecf4bdd3b0 9230.25ms  ai_completion gpt-4o
1. Add input data after the component name if --include-input  is supplied. E.g. 1c7cceecf4bdd3b0 9230.25ms  ai_completion gpt-4o What datasets do I have access to?
1. Add output data after the component name if --include-output  is supplied. E.g. 1c7cceecf4bdd3b0 9230.25ms  ai_completion gpt-4o What datasets do I have access to? blah balh

peasee:
For long inputs or outputs, include an option to truncate them. By default, truncation is disabled. Use `--truncate` with `--include-input` or `--include-output` to truncate them if they are too long.

E.g. with only `--include-input`:

```console
[
some, really, long, embedding, column, from, a, vector, search
]
```

with `--truncate --include-input`:

```console
[some, really, long...
```

PR #4771, opened 2025-02-13
feat: Add truncate command for spice trace

peasee:
# 🗣 Description

* Fixes a bug for displaying output when both `--include-input` and `--include-output` were supplied, where the output wasn't being de-referenced and was displaying a pointer address.
* Updated the input/output when value is empty to display `<empty>` on the line, instead of nothing (empty string)
* Added `--truncate`, defaulted to 80 length. Can be used as `--truncate` or `--truncate={length}` for an arbitrary truncation length.

Example output:

```console
TREE                   STATUS DURATION   TASK                    INPUT                                                                                                         OUTPUT  
5122013bb0035d84       ✅       908.08ms ai_chat                 {"messages":[{"role":"user","content":"Hello, how are you?"}],"model":"openai","... (53 characters omitted)   <empty> 
  ├── 7374cf7f07b37915 ✅         0.33ms tool_use::list_datasets <empty>                                                                                                       <empty> 
  └── a8a33044a9081926 ✅       905.79ms ai_completion           {"messages":[{"role":"assistant","tool_calls":[{"id":"initial_list_datasets","ty... (4223 characters omitted) <empty>
```

## 🔨 Related Issues

* Part of #4710

### DIFF
--- a/bin/spice/cmd/trace.go
+++ b/bin/spice/cmd/trace.go
@@ -155,7 +155,8 @@ func ToRowInterface(treePrefix string, t *taskhistory.TaskHistory, includeInput 
 		if len(t.Input) == 0 {
 			t.Input = "<empty>"
 		} else if truncateLength > 0 && len(t.Input) > truncateLength {
-			t.Input = t.Input[:truncateLength] + "..."
+			originalLength := len(t.Input)
+			t.Input = t.Input[:truncateLength] + "... " + fmt.Sprintf("(%d characters omitted)", originalLength-truncateLength)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


bin/spice/cmd/trace.go
```diff
@@ -148,17 +151,32 @@ func ToRowInterface(treePrefix string, t *taskhistory.TaskHistory, includeInput
...
+			t.Input = t.Input[:truncateLength] + "..."
+		}
+	}
+
+	var output string
+	if t.CapturedOutput != nil {
+		if len(*t.CapturedOutput) == 0 {
+			output = "<empty>"
+		} else if truncateLength > 0 && len(*t.CapturedOutput) > truncateLength {
+			output = (*t.CapturedOutput)[:truncateLength] + "..."
```
Jeadie: i wonder if we could have like `... (345 characters omitted)` or equivalent.

Jeadie (dismissed):

Sevenannn (approved):

PR #4776, opened 2025-02-13
Mark trunk builds as unstable

phillipleblanc:
# 🗣 Description

Small change to our release process to indicate that only released binaries should be considered stable, a given trunk commit may have regressions and should be used with caution.


peasee (dismissed):

Sevenannn (dismissed):

sgrebnov (dismissed):

Cargo.toml
```diff
@@ -41,7 +41,7 @@ homepage = "https://spice.ai"
 license = "Apache-2.0"
 repository = "https://github.com/spiceai/spiceai"
 rust-version = "1.84"
-version = "1.1.0"
+version = "1.1.0-unstable"
```
sgrebnov, line 44: Alternatively this can be marked as development version - `1.1.0-dev`

peasee (approved):

sgrebnov (approved):

Cargo.toml
phillipleblanc: Anyone can build from trunk, not just for development purposes. The connotation is that doing so is possible, but its inherently `unstable` compared to using a released binary.

PR #4784, opened 2025-02-13
Add support for a `time_partition_column`

phillipleblanc:
# 🗣 Description

Adds support for two new dataset parameters: `time_partition_column` and `time_partition_format`.

These behave similarly to the existing `time_column` and `time_format` parameters, but are used to specify a column and format that partitions the dataset by time. This is useful to allow appending by a fine-grained `timestamp` column, but the dataset is physically partitioned into a larger time bucket (like a `date`).

Example:
```yaml
datasets:
  - from: delta_lake://my_delta_table
    name: test
    time_column: created_at # A fine-grained timestamp 
    time_format: timestamp
    time_partition_column: day # Data is physically partitioned by `day`
    time_partition_format: date
```

## 🔨 Related Issues

N/A

## 📄 Documentation Updates

This will require docs changes

## 🤔 Concerns

This isn't the ideal approach to solve this problem. Ideally we would be able to tell Spice how columns are related to each other, which would allow them to "substitute" for each other.


sgrebnov (approved):
We should also add unit or E2E tests IMO

sgrebnov (dismissed):

sgrebnov (approved):

Sevenannn (approved):

PR #4802, opened 2025-02-14
Do no return underlying content of chunked embedding column by default during `tool_use::document_similarity`

Jeadie:
# 🗣 Description
### Bug
 - tool_use::document_similarity uses pretty arrow table formatting to return search results to LLMs. See [VectorSearchTableResult::to_pretty](https://github.com/spiceai/spiceai/blob/ef28cd820c02cbb1ea22bff366040b80f88e931b/crates/runtime/src/embeddings/vector_search.rs#L449-L451).
 - Currently, for chunked datasets, we still `SELECT` all of `[primary keys, underlying embedding column, additional columns]` from the chunked SQL query [see](https://github.com/spiceai/spiceai/blob/ef28cd820c02cbb1ea22bff366040b80f88e931b/crates/runtime/src/embeddings/vector_search.rs#L782). This means we received the underlying embedding column from the database.
 - This resulted in the [underlying record batch](https://github.com/spiceai/spiceai/blob/ef28cd820c02cbb1ea22bff366040b80f88e931b/crates/runtime/src/embeddings/vector_search.rs#L439) being used by `VectorSearchTableResult::to_pretty` to still have the original column, and was therefore printed in conjunction with the chunked column (e.g. `content` and `content_chunk`). 

### Fix
 - Explicitly remove the underlying embedding column from the chunked SQL projection. Only keep if it is explicitly requested (i.e. via `additional_columns`). 

## 📄 Documentation Updates
 - This is already the expected behavior, as per the [docs](https://spiceai.org/docs/features/search/vector-search#document-retrieval). 
 
     > _When performing searches on datasets with chunking enabled, Spice returns the most relevant chunk for each match. To retrieve the full content of a column, include the embedding column in the additional_columns list._

### DIFF
--- a/crates/runtime/src/embeddings/vector_search.rs
+++ b/crates/runtime/src/embeddings/vector_search.rs
@@ -695,6 +695,16 @@ impl VectorSearch {
                 .collect_vec()
         };
 
+        let final_projection_str = if projection.is_empty() {
+            String::new()
+        } else {
+            format!(
+                "{},",
+                // `t.` refers to the table name alias in SQL below.
+                projection.iter().map(|s| format!("t.{s}")).join(", ")
+            )
+        };
+
         format!(
             "WITH distances as (
                 SELECT
@@ -721,7 +731,7 @@ impl VectorSearch {
             )
             SELECT
                 substring(t.{embed_col}, rd.offset[1], rd.offset[2] - rd.offset[1]) AS {embed_col}_chunk,
-                {projection_str},
+                {projection_str}
                 rd.{VECTOR_DISTANCE_COLUMN_NAME}
             FROM ranked_docs rd
             JOIN {table_name} t ON {join_on_conditions}",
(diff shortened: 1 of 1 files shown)


sgrebnov (approved):

phillipleblanc (approved):

crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -721,17 +731,15 @@ impl VectorSearch {
             )
             SELECT
                 substring(t.{embed_col}, rd.offset[1], rd.offset[2] - rd.offset[1]) AS {embed_col}_chunk,
-                {projection_str},
+                {projection_str}
```
lukekim, line 734: Are you sure you meant to remove this trailing comma?

Jeadie: The comma is accounted for in https://github.com/spiceai/spiceai/pull/4802/files#diff-2099bcaabb894d8accf6f48632574ffa7633d2c2b9c4d56383a84b7e1e8c7a4fR702

The SQL in vector search is likely getting complicated enough to use an unparser.

PR #4804, opened 2025-02-14
fix: Return BAD_REQUEST when not embeddings are configured

peasee:
# 🗣 Description

* Makes vector search endpoint return a `BAD_REQUEST` instead of an `INTERNAL_SERVER_ERROR` when a search is requested with no datasets with embeddings available.

### DIFF
--- a/crates/runtime/src/embeddings/vector_search.rs
+++ b/crates/runtime/src/embeddings/vector_search.rs
@@ -833,7 +833,6 @@ impl VectorSearch {
         })
     }
 
-    #[allow(clippy::too_many_lines)]
     pub async fn search(&self, req: &SearchRequest) -> Result<VectorSearchResult> {
         let SearchRequest {
             text: query,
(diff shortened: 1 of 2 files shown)


lukekim (dismissed):

Sevenannn (dismissed):

crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -833,6 +833,7 @@ impl VectorSearch {
         })
     }
 
+    #[allow(clippy::too_many_lines)]
```
phillipleblanc: ?

phillipleblanc (dismissed):

crates/runtime/src/embeddings/vector_search.rs
peasee: 🤫

phillipleblanc (approved):

Issue #4015, opened 2024-12-28
Enhancement: Extension support for tools and `terminal` tool


Jeadie:
## Goal-State/What/Result
Enable users to define tools in an extensible, runtime manner. One of which is a shell terminal, and shell command tool(s).

## Why/Purpose
Long-tail of user tooling is needed to empower AI apps for users. Requiring code contributions, or users to build them into the binary is too demanding, uninviting & poor for rapid experimentation.

## By When
**Issue/Spec written and reviewed:** 11th February
**Done-Done:** 17th February

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

## Specification
- [MCP](modelcontextprotocol.io) is an emerging industry standard for, amongst other things, interoperability in tooling.
- Spice could handle interacting with tools via MCP. Two network protocols:
  - [ ] HTTP SSE
  - [ ] stdio: This may require spiced running the tool as process and managing the stdio.
- Spice should also provide MCP compatible endpoints for its tools. This helps both integrating with users apps, but also enables chaining spiced instances (i.e. spiced A can use tools from spiced B over MCP).

- For a list of example MCP servers, see [modelcontextprotocol/servers](https://github.com/modelcontextprotocol/servers).

<img width="1056" alt="Image" src="https://github.com/user-attachments/assets/d6cb5228-90c1-470b-aa09-634a3d455392" />

### Spicepod.yml
The first thing to note is that defining tools via MCP servers (stdio or SSE), means each tool
actually contains a list of tools. Therefore a defined tool becomes a tool group (i.e. like `auto` or `memory`).

Comparing to [Claude Desktop configuration](https://modelcontextprotocol.io/quickstart/user), an example like [puppeteer](https://github.com/modelcontextprotocol/servers/tree/main/src/puppeteer#npx):
```json
{
  "mcpServers": {
    "puppeteer": {
      "command": "npx",
      "args": ["-y", "@modelcontextprotocol/server-puppeteer"]
    }
  }
}
```
Becomes
```yaml
tools:
  name: puppeteer
  from: mcp:npx
  params:
    mcp_args:
      - "-y"
[...]

PR #4808, opened 2025-02-14
Add support for MCP tools 

Jeadie:
# 🗣 Description
Two associated, significant changes are included in this PR:
 1. Allow `spiced` to connect to, or run MCP tool servers (only run MCP servers using stdio transport).
 2. Extend the ability of `tools` defined in the spicepod to define and load multiple tools. For example
     ```yaml
     tools:
       - name: only_one_tool
          from: builtin:sql
       - name: defines_many_tools # suppose there are tools foo, bar, fizz
          from: mcp:https://somewhere.com
     ```

## Load multiple tools
Internally, tool catalogs describe tooling that provides > 1 component. Up until now, these have been internal grouping of tools (e.g. memory, auto), whereby the underlying set of tools are also available as individual tools. This PR introduces tool catalogs where this is not the case. As such, the UX to interact with them is slightly different. 

In the API, the `POST /v1/tools/:name` endpoint extends to support both tool catalog name and tool name:
 - `POST /v1/tools/only_one_tool`
 - `POST /v1/tools/defines_many_tools/foo`

In the spicepod, tool catalogs can be provided to language models either as per current catalogs, or specifying individual tools.
```yaml
models:
  - name: agi
    from: file:agi.safetensors
    params:
      tools:  only_one_tool, defines_many_tools:foo

  - name: agi_pt2
    from: file:agi.safetensors2
    params:
      tools: defines_many_tools # all tools within `defines_many_tools` are available.
```

## 🔨 Related Issues
- #4015 

## 📄 Documentation Updates
 - [ ] Document MCP tools
 - [ ] Document tool catalog UX
 - [ ] Document changes to `POST /v1/tools/*name`.


crates/runtime/src/tools/mcp/catalog.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+        .context(UnderlyingTransportSnafu)?;
+
+        client
+            .write()
+            .await
+            .initialize(
+                ClientInfo {
+                    name: "spiced".to_string(),
+                    version: "1.0".to_string(),
```
Sevenannn: Does the MCP tool client version need to stay the same as spiced version? Or is it separate?

phillipleblanc (approved):

Issue #4749, opened 2025-02-11
Refactor param handling in model components


Jeadie:
**Is your feature request related to a problem? Please describe.**
Spice runtime does not verify the correctness of `models[*].params.*` from spicepod.yaml. Currently the runtime will attempt to construct the model component using the provided `params` and only if an error occurs do we propagate it to the user. This is both:
 1. An expensive feedback cycle (e.g. an error occurs after downloading model weights).
 2. An vector for poor DX back to the user when parameters are incorrect, invalid or missing.

**Describe the solution you'd like**
See how data components handle params by being explicit. E.g. [DuckDB](https://github.com/spiceai/spiceai/blob/9d07bce21d9d3ca1904d7ff7afa89cbbcda5b69c/crates/runtime/src/dataconnector/flightsql.rs#L68-L72). Helper struct/utilities [here](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/parameters.rs). 

**Describe alternatives you've considered**
The current implementation is a feasible working alternative. Improving error message handling in the current paradigm is possible, but requires handling many issues from underlying causes instead of validating parameters explicitly first.

Sevenannn:
- Model parameter spec has been explicitly defined as part of the pr https://github.com/spiceai/spiceai/pull/6199

PR #4833, opened 2025-02-18
Refactor Parameter Handling

Advayp:
# 🗣 Description

- Refactored parameter handling to work generally by removing any coupling to specific components. 
- For parameters that are prefixed, like for those in connectors and accelerators, I raised the prefixing configuration API. Instead of specifying which parameters need to be prefixed via a dedicated `ParameterType`, you can just call `unset_prefix()` on the parameter spec instead. I did this because, based on what I saw, the `ParameterType` enum was _only_ being used to handle prefixing, and nothing else.
- Also, I'm aware this PR seems pretty big, but most of the changes are from migrating from the old parameter handling to the way I implemented

## 🔨 Related Issues
- Related to, but doesn't close, #4749


crates/runtime/src/model/chat.rs
```diff
@@ -212,6 +212,17 @@ fn openai(
     let org_id = extract_secret!(params, "openai_org_id");
     let project_id = extract_secret!(params, "openai_project_id");
 
+    match api_key {
+        None => {
+            return Err(LlmError::InvalidParamError {
+                param: "openai_api_key".to_string(),
```
lukekim: API key is not always required.

crates/runtime/src/parameters.rs
```diff
@@ -315,13 +315,13 @@ pub struct ParameterSpec {
...
     pub examples: &'static [&'static str],
-    pub r#type: ParameterType,
     pub deprecation_message: Option<&'static str>,
+    pub is_prefixed: bool,
 }
 
 impl ParameterSpec {
     #[must_use]
-    pub const fn connector(name: &'static str) -> Self {
+    pub const fn new(name: &'static str) -> Self {
```
phillipleblanc: Let's keep the distinction of a `runtime` parameter vs a `component` parameter, where a runtime parameter is unprefixed, but a component parameter isn't.

Then the distinction between them would be:

```
pub enum ParameterType {
    /// A parameter which tells Spice how to configure the underlying component, and is usually passed directly to the underlying component configuration.
    ///
    /// These parameters are automatically prefixed with the component's prefix.
    ///
    /// # Examples
    ///
    /// In Postgres, `host` is a Component parameter and would be auto-prefixed with `pg_`.
    #[default]
    Component,

    /// Other parameters which control how the runtime interacts with the component, but does
    /// not affect the actual component configuration.
    ///
    /// These parameters are not prefixed with the component's prefix.
    ///
    /// # Examples
    ///
    /// In Databricks, the `mode` parameter is used to select which connection to use, and thus is
    /// not a component parameter.
    Runtime,
}
```

phillipleblanc: So instead of ParameterSpec::new() it would be ParameterSpec::component()

crates/runtime/src/parameters.rs
```diff
@@ -492,7 +415,7 @@ mod test {
         // key with prefix, parameter does not expect prefix.
         assert_eq!(
             Parameters::validate_and_format_key(
-                &[ParameterSpec::runtime("endpoint")], // deliberately `runtime` not `connector`.
+                &[ParameterSpec::new("endpoint").unset_prefix()], // deliberately `runtime` not `connector`.
```
phillipleblanc:
```suggestion
                &[ParameterSpec::new("endpoint").unset_prefix()], // deliberately `runtime` not `component`.
```

phillipleblanc (dismissed):

phillipleblanc (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

phillipleblanc:
I've brought these changes into a branch and created a draft PR to run the checks - if they all pass, I'll force merge this PR: https://github.com/spiceai/spiceai/pull/4896

PR #4835, opened 2025-02-19
Fix 'actual" and "output" columns in `eval.results`. 

Jeadie:
# 🗣 Description
 - Previously we were writing the expect and actual model responses in the opposite columns in the `eval.results` table.
 - Change naming to be clear which is from model, vs from eval dataset.
 
 ## Breaking Change
```diff
sql> describe eval.results;
  +-------------+-------------------------+-------------+
  | column_name | data_type               | is_nullable |
  +-------------+-------------------------+-------------+
  | run_id      | Utf8                    | NO          |
  | created_at  | Timestamp(Second, None) | NO          |
  | input       | Utf8                    | NO          |
- | output      | Utf8                    | NO          |
+ | expected    | Utf8                    | NO          |
  | actual      | Utf8                    | NO          |
  | scorer      | Utf8                    | NO          |
  | value       | Float32                 | NO          | 
  +-------------+-------------------------+-------------+
```

Docs:
- https://github.com/spiceai/cookbook/pull/92

### DIFF
--- a/crates/runtime/src/model/eval/mod.rs
+++ b/crates/runtime/src/model/eval/mod.rs
@@ -269,21 +269,21 @@ async fn write_results(
     run_id: &EvalRunId,
     df: Arc<DataFusion>,
     input: &[DatasetInput],
-    output: &[DatasetOutput],
+    actual: &[DatasetOutput],
     expected: &[DatasetOutput],
     scores: &HashMap<String, Vec<f32>>,
 ) -> Result<()> {
     let mut bldr = ResultBuilder::new();
     for i in 0..input.len() {
         let input = &input[i];
-        let output = &output[i];
+        let actual = &actual[i];
         let expected = &expected[i];
         for (name, score) in scores {
             bldr.append(
                 run_id,
                 chrono::Utc::now(),
                 input,
-                output,
+                actual,
                 expected,
                 name,
                 score[i],
--- a/crates/runtime/src/model/eval/mod.rs
+++ b/crates/runtime/src/model/eval/mod.rs
@@ -283,8 +283,8 @@ async fn write_results(
                 run_id,
                 chrono::Utc::now(),
                 input,
-                expected,
                 actual,
+                expected,
                 name,
                 score[i],
             )?;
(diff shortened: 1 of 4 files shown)


crates/runtime/src/model/eval/result.rs
```diff
@@ -87,7 +87,7 @@ pub(super) struct ResultBuilder {
     run_id: StringBuilder,
     created_at: TimestampSecondBuilder,
     input: StringBuilder,
-    output: StringBuilder,
+    output: StringBuilder, // expected output given the `input`.
```
lukekim:
```suggestion
    expected: StringBuilder, // expected output given the `input`.
```

crates/runtime/src/model/eval/result.rs
```diff
@@ -112,15 +112,15 @@ impl ResultBuilder {
...
         actual: &DatasetOutput,
+        expected: &DatasetOutput,
         scorer: &str,
         value: f32,
     ) -> Result<()> {
         self.run_id.append_value(id);
         self.created_at.append_value(created_at.timestamp());
         self.input.append_value(input.try_serialize()?);
-        self.output.append_value(output.try_serialize()?);
+        self.output.append_value(expected.try_serialize()?);
```
lukekim:
```suggestion
        self.expected.append_value(expected.try_serialize()?);
```

crates/runtime/src/model/eval/mod.rs
```diff
@@ -269,21 +269,21 @@ async fn write_results(
...
+        let actual = &actual[i];
         let expected = &expected[i];
         for (name, score) in scores {
             bldr.append(
                 run_id,
                 chrono::Utc::now(),
                 input,
-                output,
+                actual,
                 expected,
```
phillipleblanc, line 287: This order is `input`, `actual`, `expected` - but the field order in the other file is `input`, `expected`, `actual`

Jeadie: Fixed

crates/runtime/src/model/eval/result.rs
```diff
@@ -112,15 +112,15 @@ impl ResultBuilder {
         id: &EvalRunId,
         created_at: DateTime<Utc>,
         input: &DatasetInput,
-        output: &DatasetOutput,
         actual: &DatasetOutput,
+        expected: &DatasetOutput,
```
phillipleblanc, line 116: This order also needs to be updated?

Sevenannn (approved):

PR #4868, opened 2025-02-25
Fixing Go installation in the setup script for Linux Arm64

sergey-shandar:
# 🗣 Description

Several issues were fixed in CONTRIBUTING.MD:

1. Allow changing the platform (e.g. ARM64) when installing Go.
2. `spiceai init` creates a directory. So the correct sequence should be
   ```sh
   spice init test-app
   cd test-app
   spice run
   ```
3. Update GO_VERSION to 1.24.0.

## ⛓️‍💥 Breaking Change

## 📄 Documentation Updates

Fix for the [./CONTRIBUTING.md](https://github.com/spiceai/spiceai/blob/trunk/CONTRIBUTING.md).

## 🤔 Concerns

Perhaps we should use `sudo apt install golang-go` instead? Unless we need a specific version of Go.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,7 +114,6 @@ export PATH="$PATH:$HOME/.spice/bin"
 
 # Initialize and run a test app to ensure everything is working
 cd ../
-mkdir test-app
 spice init test-app
 cd test-app
 spice run
@@ -164,7 +163,6 @@ export PATH="$PATH:$HOME/.spice/bin"
 
 # Initialize and run a test app to ensure everything is working
 cd ../
-mkdir test-app
 spice init test-app
 cd test-app
 spice run
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -128,12 +128,13 @@ sudo apt update
 sudo apt install build-essential curl openssl libssl-dev pkg-config protobuf-compiler cmake
 
 # Install Go
-export GO_VERSION="1.23.4"
+export GO_VERSION="1.24.0"
+export ARCH="arm64" # See https://golang.org/dl/ for available architectures
 rm -rf /tmp/spice
 mkdir -p /tmp/spice
 cd /tmp/spice
-wget https://go.dev/dl/go$GO_VERSION.linux-amd64.tar.gz
-tar xvfz go$GO_VERSION.linux-amd64.tar.gz
+wget https://go.dev/dl/go$GO_VERSION.linux-$ARCH.tar.gz
+tar xvfz go$GO_VERSION.linux-$ARCH.tar.gz
 sudo mv ./go /usr/local/go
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -128,12 +127,13 @@ sudo apt update
 sudo apt install build-essential curl openssl libssl-dev pkg-config protobuf-compiler cmake
 
 # Install Go
-export GO_VERSION="1.23.4"
+export GO_VERSION="1.24.0"
+export ARCH="amd64" # E.g. arm64. See https://golang.org/dl/ for available architectures
```
phillipleblanc: Would be nice if we could auto-detect this somehow.

phillipleblanc (dismissed):

lukekim (dismissed):

phillipleblanc (dismissed):

CONTRIBUTING.md
sergey-shandar: Done. The script detects only Amd64 and Arm64 for now.

CONTRIBUTING.md
```diff
@@ -126,11 +126,9 @@ spice run
 sudo apt update
 sudo apt install build-essential curl openssl libssl-dev pkg-config protobuf-compiler cmake
 
-# Detect the machine's architecture
-ARCH=$(uname -m)
-
-# Map uname output to Go's naming convention
-case "$ARCH" in
+# Install Go
+ARCH=$(uname -m) # Detect the machine's architecture
```
phillipleblanc, line 130: `uname -m` returns `arm64` for me on macOS 🤔

phillipleblanc (approved):

CONTRIBUTING.md
sergey-shandar:
```
sergey@warm:~/sergey-spiceai$ uname -m
aarch64
```

sergey-shandar: It looks for Debian (Ubuntu) it should be `aarch64`: https://wiki.debian.org/ArchitectureSpecificsMemo

sergey-shandar: Found the answer here https://stackoverflow.com/questions/31851611/differences-between-arm64-and-aarch64 :

>It seems that ARM64 was created by Apple and AARCH64 by the others, most notably GNU/GCC guys.

>After some googling I found this link:

>[The LLVM 64-bit ARM64/AArch64 Back-Ends Have Merged](https://www.phoronix.com/news/MTY5ODk)

>So it makes sense, iPad calls itself ARM64, as Apple is using LLVM, and Edge uses AARCH64, as Android is using GNU GCC toolchain.

phillipleblanc: TIL! Given this is for the Linux section, I think we can leave it

PR #4879, opened 2025-02-26
Gracefully shutdown the runtime and deallocate static resources

sgrebnov:
# 🗣 Description

Properly shut down the runtime and deallocate static resources: this ensures that open connections used to accelerate datasets locally are properly closed and that the runtime is shut down gracefully.

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

DataFusion shutdown currently performs de-registration/cleanup of accelerated tables only. This will be updated to de-register all tables.

### DIFF
--- a/crates/runtime/src/catalogconnector/mod.rs
+++ b/crates/runtime/src/catalogconnector/mod.rs
@@ -138,6 +138,11 @@ pub async fn register_all() {
     );
 }
 
+pub async fn unregister_all() {
+    let mut registry = CATALOG_CONNECTOR_FACTORY_REGISTRY.lock().await;
+    registry.clear();
+}
+
 pub(crate) struct CatalogConnectorFactory {
     connector_factory: fn(ConnectorParams) -> Arc<dyn CatalogConnector>,
     prefix: &'static str,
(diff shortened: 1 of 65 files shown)


crates/runtime/src/catalogconnector/mod.rs
```diff
@@ -138,6 +138,11 @@ pub async fn register_all() {
     );
 }
 
+pub async fn unregister_all() {
+    let mut registry = CATALOG_CONNECTOR_FACTORY_REGISTRY.lock().await;
+    registry.clear();
```
Sevenannn, line 143: For learning purpose: Why cleaning the static registry here could ensure that connections are correctly dropped

Sevenannn (dismissed):

crates/runtime/src/catalogconnector/mod.rs
sgrebnov: Some static resources include instances of TableProviderFactories that might contain references to created connections, like `DuckDBTableProviderFactory.instances` used to track already created connection pools to avoid duplicate pool creation.

```
pub struct DuckDBTableProviderFactory {
    access_mode: AccessMode,
    instances: Arc<Mutex<HashMap<DbInstanceKey, DuckDbConnectionPool>>>,
    unsupported_type_action: UnsupportedTypeAction,
    dialect: Arc<dyn Dialect>,
}
```
Factories might have references to other resources that needs to be properly dropped and require manual clean up as they are stored in static item and not dropped automatically.

Sevenannn (approved):

rileyseaburg:
@sgrebnov does this pr handle the Ctrl+c issue as well? Just want to make sure I'm not adding unneeded code

rileyseaburg:
nvm disregard just tested it

Issue #4760, opened 2025-02-12
Bug: OpenAI snapshot tests are not robust


peasee:
## Describe the bug

The OpenAI snapshot tests are not robust, resulting on test failures for changes in a snapshot that are not indicative of a failure.

## To Reproduce

_Steps to reproduce the behavior:_

1. See transient test failures: https://github.com/spiceai/spiceai/actions/runs/13278441153/job/37072276216?pr=4759

## Expected behavior

_A clear and concise description of what you expected to happen._

Tests with snapshot changes that are functionally the same, or not significant enough to warrant a test failure, should produce a test pass.

In the example linked above, the contents of the vector search text changed and the ordering of object columns changed. The text changed from `vehicles` to `Find information about vehicles` - I would expect both of these texts to return the same results, and thus be functionally the same.

Jeadie:
Issue, see [example](https://github.com/spiceai/spiceai/actions/runs/13278441153/job/37072276216?pr=4759) appears to be key ordering in JSON.

PR #4885, opened 2025-02-26
Shutdown Runtime when user presses `Ctrl + C`

### LINKED ISSUE #3409: Bug: unable to terminate Spice with Ctrl-C while model is being downloaded
## Describe the bug

Unable to terminate Spice with Ctrl-C while model is being downloaded

Image

rileyseaburg:
# 🗣 Description

This pull request implements a more robust Ctrl+C signal handling mechanism for the `spiced` binary. The changes ensure that the application can gracefully shut down on the first Ctrl+C press, warn the user on the second press, and forcefully exit on the third press. This is achieved by using the `ctrlc` crate to handle the signal and track the number of Ctrl+C presses.

## 🔨 Related Issues

This pull request addresses the issue where the application does not stop properly when Ctrl+C is pressed.

#3409

## ⛓️‍💥 Breaking Change

*None

## 📄 Documentation Updates

No documentation updates are required for this change.

## 🤔 Concerns

There are no particular concerns about this pull request. The changes have been tested and verified to work as expected.

Tested with Ctrl+c three times.
image


rileyseaburg:
I'm adding support for the graceful shutdown

bin/spiced/src/main.rs
```diff
@@ -48,10 +53,45 @@ fn main() {
         return;
     }
 
-    if let Err(err) = tokio_runtime.block_on(start_runtime(args)) {
-        spiced::in_tracing_context(|| {
-            tracing::error!("{err}");
-        });
+    // Register a global Ctrl+C handler that forcibly exits after repeated presses
+    ctrlc::set_handler(move || {
```
sgrebnov: @rileyseaburg - thank you for PR - looks great, few notes

1. The logging is too verbose IMO. I propose to trace message only once (first Ctrl-C)
```
2025-02-26T23:32:01.262957Z  INFO spiced: Shutting down the runtime... Press Ctrl+C again to force exit.
2025-02-26T23:32:01.263939Z  INFO runtime: Goodbye!
```

```
spice run
...
2025-02-26T23:31:56.999020Z  INFO runtime::init::model: Model [local_model] deployed, ready for inferencing
^C2025-02-26T23:32:01.262957Z  INFO spiced: Received Ctrl+C, shutting down gracefully...
2025-02-26T23:32:01.263060Z  WARN spiced: Received Ctrl+C again, waiting for graceful shutdown to complete...
2025-02-26T23:32:01.263067Z  WARN spiced: Press Ctrl+C once more to force exit
2025-02-26T23:32:01.263939Z  INFO runtime: Goodbye!
2025-02-26T23:32:01.264120Z  INFO spiced: Runtime shut down successfully
``` 

1. I also hit Ctrl-C only once (logs are above) but there is ` Received Ctrl+C again,` message. Tested on macos 15.3.1 / Warp terminal

rileyseaburg: got it, let me fix that, i was focusing so much on getting it to acutally kill the process, I'll fine tune that

sgrebnov: Note - tracing like this is fine/very useful for debugging, it should just be traced as debug, not info

```
tracing::debug!("Shutdown signal received, stopping runtime");
```

rileyseaburg: Got it. Thank you for the feedback!

rileyseaburg:
Simplified the exit process.
Added minimal debug tracing

bin/spiced/src/main.rs
```diff
@@ -48,11 +51,13 @@ fn main() {
...
-            tracing::error!("{err}");
-        });
-    }
+    // Register a global Ctrl+C handler that initiates a shutdown
+    ctrlc::set_handler(move || {
+        
+            tracing::debug!("Shutdown signal received, stopping runtime.");
+            std::process::exit(130);
+    
+    }).expect("Error setting Ctrl+C handler");
```
phillipleblanc: Please add proper error handling here.

bin/spiced/src/main.rs
```diff
@@ -17,11 +17,14 @@ limitations under the License.
 use clap::Parser;
 use opentelemetry::global;
 use rustls::crypto::{self, CryptoProvider};
+use std::sync::atomic::{AtomicBool, Ordering};
 use tokio::runtime::Runtime;
 
 #[global_allocator]
 static ALLOC: snmalloc_rs::SnMalloc = snmalloc_rs::SnMalloc;
 
+static SHUTDOWN_REQUESTED: AtomicBool = AtomicBool::new(false);
```
phillipleblanc: When does this get set to true?

phillipleblanc:
Please update the PR title

bin/spiced/Cargo.toml
```diff
@@ -11,6 +11,7 @@ version.workspace = true
 [dependencies]
 app = { path = "../../crates/app" }
 clap = { workspace = true, features = ["derive"] }
+ctrlc = "3.4.5"
```
phillipleblanc, line 14: Why do we need an extra crate for this?

rileyseaburg: Couldn't get the runtime to recognize the ctrl+c without it, tried it for like 2 days

bin/spiced/src/main.rs
rileyseaburg: artifact that was left over for a more robust shtudown, I'll remove it.

bin/spiced/src/main.rs
rileyseaburg: just updated this.

bin/spiced/src/main.rs
```diff
@@ -48,6 +48,20 @@ fn main() {
         return;
     }
 
+    // Register a global Ctrl+C handler that initiates a shutdown
+    if let Err(err) = ctrlc::set_handler(move || {
+        spiced::in_tracing_context(|| {
+            tracing::debug!("Ctrl+C received, shutting down");
+        });
+        std::process::exit(130);
```
phillipleblanc: This is an immediate unclean shutdown when the user presses ctrl-c - this is not what should happen. When ctrl-c is pressed, the runtime should perform a graceful shutdown and cancel any in-progress tasks (like downloading models) to allow the runtime to exit cleanly.

phillipleblanc (changes requested):
The runtime should perform a graceful shutdown on ctrl-c - force killing the process doesn't give the runtime a chance to clean up any resources it is using.

rileyseaburg:
> The runtime should perform a graceful shutdown on ctrl-c - force killing the process doesn't give the runtime a chance to clean up any resources it is using.

Predefined behavior makes me think this issues is a feature request and not a bug. 

Nevertheless I'll add the requested graceful cleanup process.

rileyseaburg:
Closing so people don't keep getting emails,when I push

bin/spiced/src/main.rs
rileyseaburg: Handled in latest commit, please verify this is what you want @phillipleblanc

rileyseaburg:
@phillipleblanc please let me know if this is the type of behavior you're looking for. 

I had to reach into various parts of the runtime to ensure the temporary directories have been cleaned up.

rileyseaburg:
The required checks passed, so I think this is finally ready for review

Jeadie:
If this is to just address HF model download #3409 , then we can do that in our fork of `mistralrs`. This PR https://github.com/spiceai/mistral.rs/pull/24 should resolve the issue. 

Otherwise, if Ctrl+C isn't handled more broadly, that is something we will have to address here.

rileyseaburg:
Not much direction came except this is a bug that needs to be worked on, it's not really just a mistral issue. For example I was testing it with llama.

peasee:
> The required checks passed, so I think this is finally ready for review

Although the E2E tests aren't required, it looks like this PR has introduced some bug with Runtime readiness. Most of the E2E  tests are failing because the runtime is failing to report "ready" (`http://localhost:8090/v1/ready`). 

Could you take a look?

rileyseaburg:
I'm not sure why the OpenAI one is failing, looks like it's only one failing test in that group as well

```
test result: FAILED. 3 passed; 1 failed; 1 ignored; 0 measured; 8 filtered out; finished in 6.12s

```

rileyseaburg:
> > The required checks passed, so I think this is finally ready for review
> 
> Although the E2E tests aren't required, it looks like this PR has introduced some bug with Runtime readiness. Most of the E2E tests are failing because the runtime is failing to report "ready" (`http://localhost:8090/v1/ready`).
> 
> Could you take a look?
@peasee 
Addressed this, please confirm the issue is resolved

peasee:
> I'm not sure why the OpenAI one is failing, looks like it's only one failing test in that group as well
> 
> ```
> test result: FAILED. 3 passed; 1 failed; 1 ignored; 0 measured; 8 filtered out; finished in 6.12s
> ```

Thanks! The OpenAI tests are a bit flaky, there's a tracking issue in #4760. I'm going to test this locally for some ctrl+c scenarios.

rileyseaburg:
> > I'm not sure why the OpenAI one is failing, looks like it's only one failing test in that group as well
> > ```
> > test result: FAILED. 3 passed; 1 failed; 1 ignored; 0 measured; 8 filtered out; finished in 6.12s
> > ```
> 
> Thanks! The OpenAI tests are a bit flaky, there's a tracking issue in #4760. I'm going to test this locally for some ctrl+c scenarios.

For me it shuts down about 10 seconds after ctrl+c. I think it depends on what you're downloading though

crates/runtime/src/lib.rs
```diff
@@ -46,6 +49,19 @@ use tls::TlsConfig;
...
 use tools::factory::default_available_catalogs;
 use tools::{catalog::SpiceToolCatalog, Tooling};
+
+// Global shutdown flag that can be set from other crates
+pub static SHUTDOWN_IN_PROGRESS: AtomicBool = AtomicBool::new(false);
+
+/// Sets the global shutdown flag to indicate shutdown is in progress.
+/// This allows components across the application to check if they should stop processing.
+pub fn set_shutdown_in_progress(value: bool) {
+    SHUTDOWN_IN_PROGRESS.store(value, Ordering::SeqCst);
```
sgrebnov, line 59: I think it is better to track this via `runtime.status` field that can expose this information (it seems to be available where SHUTDOWN_IN_PROGRESS is currently used to terminate model loading), let's consolidate two PRs - I have not seen this when worked on shutdown as well based on other work item - graceful shutdown of db connections.

https://github.com/spiceai/spiceai/pull/4917/files#diff-f1f8c385c3e7ac464ae7a46530049494764373926ee5744df15062489ddeb80eR218

bin/spiced/src/main.rs
```diff
@@ -49,9 +60,85 @@ fn main() {
         return;
     }
 
-    if let Err(err) = tokio_runtime.block_on(start_runtime(args)) {
+    // Create a broadcast channel to signal shutdown across all threads
+    let (shutdown_tx, _) = broadcast::channel::<()>(32); // Increase capacity to ensure all components can receive signals
```
sgrebnov, line 64: I'm under impression is that capacity is number of messages here, not max receivers

bin/spiced/src/main.rs
```diff
@@ -49,9 +60,85 @@ fn main() {
...
+                }
+                Err(e) => {
+                    // Don't show this as an error since it's expected during some shutdown scenarios
+                    tracing::debug!("Note: No active receivers for shutdown signal: {}", e);
+                }
+            }
+
+            // Always run global cleanup operations during shutdown
+            // Use a closure to contain cleanup tasks that need to be performed
+            let cleanup_resources = || {
```
sgrebnov, line 99: Riley, could you please elaborate why it is required

bin/spiced/src/main.rs
```diff
@@ -60,11 +147,14 @@ fn main() {
...
 }
 
-async fn start_runtime(args: spiced::Args) -> Result<(), Box<dyn std::error::Error>> {
+async fn start_runtime(
+    args: spiced::Args,
+    shutdown_rx: broadcast::Receiver<()>,
+) -> Result<(), Box<dyn std::error::Error>> {
     spiced::in_tracing_context(|| {
-        tracing::info!("Starting runtime {version}", version = get_version_string());
+        tracing::debug!("Starting runtime {version}", version = get_version_string());
```
sgrebnov, line 155: We should not change this IMO

crates/llms/src/embeddings/candle/tei.rs
```diff
@@ -304,3 +306,38 @@ impl Embed for TeiEmbed {
         ))
     }
 }
+
+impl Drop for TeiEmbed {
+    fn drop(&mut self) {
+        // Clean up the model directory when the embedding model is dropped
```
sgrebnov, line 312: Riley, could you please elaborate why it is required - I was under impression is that it is good that we keep model directory so we don't need to re-download files again. Is there a case where files are not properly downloaded so we need to clean them up?

crates/runtime/src/init/model.rs
```diff
@@ -52,24 +54,76 @@ pub enum Error {
...
         if let Some(app) = app_lock.as_ref() {
+            let mut models = Vec::new();
             for model in &app.models {
+                models.push(model);
+            }
+
+            for model in models {
+                // Check for cancellation signal between models
+                // Add a small yield point to be responsive to shutdown signals
+                task::yield_now().await;
```
sgrebnov, line 77: Responsive means to be able to catch Ctrl+C? As it seems there is next step that should correctly stop models loading

bin/spiced/src/main.rs
```diff
@@ -17,12 +17,23 @@ limitations under the License.
 use clap::Parser;
 use opentelemetry::global;
 use rustls::crypto::{self, CryptoProvider};
+use std::{
```
sgrebnov, line 20: I think we should align on the overall approach, propose the following - @phillipleblanc WDYT?
1. Spiced main.rs should be responsible only for robust handling Ctrl+C
 - first Ctrl+C - signal runtime to initiate termination (approach for rt.close to be called), should not  have own logic to control global termination flags, etc - this is Runtime responsibility IMO
 - ? multiple Ctrl-C - force termination
 - We should remove `std::thread::sleep(std::time::Duration::from_secs(10));` to force terminate runtime IMO
 2. `Runtime` does its best to terminate everything, can manipulate by special flags (global vs just runtime.status.is_shutdown), channels, etc

crates/runtime/src/lib.rs
rileyseaburg: Yeah I saw you guys started adding extra flags in. 

I can pull in your changes and go from there.

crates/runtime/src/init/model.rs
rileyseaburg: Without this the model runtime seems to data race and ignore the shutdown signal.

crates/llms/src/embeddings/candle/tei.rs
rileyseaburg: From what I gathered during download it's actually storing into a temporary PathBuf which might not always store directly to a model directory if a download is interrupted. 

Additionally we need to cleanup partially downloaded resources in order to gracefully handle a shutdown.

bin/spiced/src/main.rs
rileyseaburg: Good catch that was accidental changed in a find and replace

bin/spiced/src/main.rs
rileyseaburg: @phillipleblanc requested the cleanup of resources. 

https://github.com/spiceai/spiceai/pull/4885#pullrequestreview-2647572347

rileyseaburg: @sgrebnov id you can point me to where we are already cleaning up resources, I'll call that method on shutdown signal.

bin/spiced/src/main.rs
rileyseaburg: Some of this I had already adjusted per @phillipleblanc recommendations.

I'll hold off until you all decide the way you expect this system to perform.

PR #4893, opened 2025-02-27
Add `workers` component

lukekim:
# 🗣 Description

Adds new concept of a `worker` of which the first type is a `ModelWorker`. Workers will be used for coordinating work across the Spicepod (aka AI agent). In the future, there can be other types of workers too, imagine `PythonWorker` etc. for running code.

Example Spicepod.yml:

```yaml
workers:
  - from: models:gpt-4o
    role: researcher
    name: barry_the_researcher
```

Which will create a Researcher worker using the gpt-4o model.

## ⛓️‍💥 Breaking Change

None.

## 📄 Documentation Updates

Will be needed once wired up.

## 🤔 Concerns

This PR adds the worker concept, but it is yet to be used.


phillipleblanc (dismissed):

crates/runtime/src/http/v1/workers.rs
```diff
@@ -0,0 +1,153 @@
...
+    let workers = match app.read().await.as_ref() {
+        Some(a) => a
+            .workers
+            .iter()
+            .map(|w| Worker {
+                from: w.from.clone(),
+                name: w.name.clone(),
+                role: w.role.clone(),
+            })
+            .collect::<Vec<Worker>>(),
```
kczimm: It might be easier to `#[derive(Clone)]` for `Worker` and just `workers.iter().cloned().collect()` here.

crates/runtime/src/init/worker.rs
```diff
@@ -0,0 +1,173 @@
...
+        }
+
+        // Remove workers that are no longer in the app
+        for worker in &current_app.workers {
+            if !new_app.workers.iter().any(|w| w.name == worker.name) {
+                self.status
+                    .update_worker(&worker.name, status::ComponentStatus::Disabled);
+                self.remove_worker(worker).await;
+            }
+        }
```
kczimm: If workers load models (or do any large memory allocation), do we want to remove workers before adding new ones to avoid OOM?

crates/runtime/src/status.rs
```diff
@@ -139,6 +139,27 @@ impl RuntimeStatus {
         metrics::views::STATUS.record(status as u64, &[KeyValue::new("view", view_name)]);
     }
 
+    /// Update the status of a worker
+    pub fn update_worker(&self, name: &str, status: ComponentStatus) {
+        let mut components = match self.statuses.write() {
+            Ok(guard) => guard,
+            Err(poisoned) => poisoned.into_inner(),
```
kczimm, line 161: Any thoughts on migrating to the `parking_lot` crate to avoid dealing with lock poisoning?

kczimm (dismissed):

crates/runtime/src/status.rs
phillipleblanc: I think it makes sense. I initially was wary about adding extra dependencies to our codebase - which is incremental maintenance overhead. But it shouldn't affect compilation times since DataFusion already depends on it.

crates/runtime/src/init/worker.rs
Jeadie: It's not clear that workers would be responsible for load/unloading models instead of pointing to existing/predefined models. Either way, no harm in swapping order of operations.

crates/runtime/src/status.rs
Jeadie: Will deal with separately in https://github.com/spiceai/spiceai/issues/5499

Jeadie:
Proceeded by https://github.com/spiceai/spiceai/pull/5513. Unneeded now

Issue #4888, opened 2025-02-27
Upon runtime termination, components must be shut down in the correct sequence


sgrebnov:
1. Shutdown initiated (Ctrl-C)
2. HTTP API including /ready, /status, /sql turns 503 except /health which continue 200
3. Components shutdown
   - Step1: Flight, OpenTelemetry (metrics ingestion) servers graceful shutdown: wait until in-progress requests are completed, stop serving new requests
   - Step2: Datasets, Models, etc shutdown (currently Runtime.shutdown / Datafusion.shutdown)
5. HTTP API complete shutdown
6. Metrics (9090) endpoint shutdown
7. All resources clean up (termination)

sgrebnov:
Expected workflow
```
1. Shutdown initiated
2. /ready API turns 503
3. [optional, configurable wait period to allow connection drain]
4. internal components shutdown (except HTTP API)
5. /health API turns 503
6. HTTP API shutsdown
7. Runtime shutsdown
```

Issue #4814, opened 2025-02-15
Bug: DuckDB acceleration `Binder Error: Failure while replaying WAL file` error


sgrebnov:
## Describe the bug

>Binder Error: Failure while replaying WAL file "/Users/sg/spice/spiceai/.spice/data/accelerated_duckdb.db.wal": Referenced table "taxi_trips_spice_99a9ab1a" not found!

## To Reproduce

```yaml
- from: s3://spiceai-demo-datasets/taxi_trips/2024/
  name: taxi_trips
  time_column: tpep_pickup_datetime
  params:
    file_format: parquet
  acceleration:
    enabled: true
    engine: duckdb
    mode: file
    refresh_sql: "SELECT * FROM taxi_trips limit 1"
    primary_key: tpep_pickup_datetime
    indexes:
      tpep_pickup_datetime: enabled
```

sgrebnov:
Fixed by https://github.com/spiceai/spiceai/pull/4879

sgrebnov:
Using new test I was able to repro this again. If I don't do requests (use oha), WAL is correctly written, but after the workload I can see the same behavior
https://github.com/spiceai/spiceai/pull/4891

PR #4912, opened 2025-03-04
Fix Datafusion resources deallocation during shutdown

sgrebnov:
# 🗣 Description

`SpiceMetricsExporter` has reference to Datafusion and is used by global `SdkMeterProvider` when metrics are enabled. Even if we add clean up logic for global meter at the end, and the  `SdkMeterProvider` and all it is internal structures are dropped, already created metrics which are static in our case will still have references to `OtelArrowExporter`/`Datafusion` preventing  Datafusion to gracefully shutdown and clean up resources including closing open connections.

PR updates `SpiceMetricsExporter` to use Weak reference to Datafusion so it can be dropped, when export is performed the SpiceMetricsExporter checks that Datafusion is still available by creating temporary strong reference using `upgrade`.

PR also resets global meter provider at the end of runtime termination to clean up SdkMeterProvider instance as global meter use static field to keep reference to meter provider which is not dropped during termination.

## 🔨 Related Issues

Fixes
 - https://github.com/spiceai/spiceai/issues/4814

## ⛓️‍💥 Breaking Change

## 🤔 Concerns
1. Will add more debug tracing (when Datafusion is dropping, etc as part of next step: https://github.com/spiceai/spiceai/issues/4888 )
1. Alternative approach considered was implementing static metrics clean up during termination (or special utility/abstraction around metrics), this approach does not seem very robust as requires tracking all metrics and ensure all new metrics are manually cleaned up during termination. Note: we might combine both.

Example metric

```
 pub(crate) static TOOLS_METER: LazyLock<Meter> = LazyLock::new(|| global::meter("tool"));

    pub(crate) static COUNT: LazyLock<UpDownCounter<i64>> = LazyLock::new(|| {
        TOOLS_METER
            .i64_up_down_counter("tool_active_count")
            .with_description("Number of currently loaded LLM tools.")
            .build()
    });
```

### DIFF
--- a/crates/runtime/src/spice_metrics.rs
+++ b/crates/runtime/src/spice_metrics.rs
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 use std::time::Duration;
 
 use arrow::array::RecordBatch;
@@ -44,12 +44,14 @@ pub enum Error {
 }
 
 pub struct SpiceMetricsExporter {
-    datafusion: Arc<DataFusion>,
+    datafusion: Weak<DataFusion>,
 }
 
 impl SpiceMetricsExporter {
-    pub fn new(datafusion: Arc<DataFusion>) -> Self {
-        SpiceMetricsExporter { datafusion }
+    pub fn new(datafusion: &Arc<DataFusion>) -> Self {
+        SpiceMetricsExporter {
+            datafusion: Arc::downgrade(datafusion),
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/runtime/src/spice_metrics.rs
```diff
@@ -62,8 +64,13 @@ impl otel_arrow::ArrowExporter for SpiceMetricsExporter {
             update_type: UpdateType::Append,
         };
 
-        self.datafusion
-            .write_data(&get_metrics_table_reference(), data_update)
+        let Some(df) = self.datafusion.upgrade() else {
+            return Err(MetricError::Other(
+                "DataFusion is not available".to_string(),
```
lukekim: Need a more informative error

sgrebnov: Thank you - update to 
```rust
let Some(df) = self.datafusion.upgrade() else {
    // this should never happen as the exporter must be shutdown before the DataFusion instance is dropped
    return Err(MetricError::Other(
        "Failed to export metrics as the DataFusion instance has already been dropped.\nReport an issue on GitHub: https://github.com/spiceai/spiceai/issues".to_string(),
    ));
};
```

Sevenannn (approved):

phillipleblanc (approved):

PR #4917, opened 2025-03-05
Introduce runtime shutdown state

sgrebnov:
# 🗣 Description

Introduce a runtime shutdown state that can be used to gracefully stop in-progress processes such as dataset loading, etc.

Note: As all servers are stopped before runtime termination, and there is no way to access the runtime state or query the runtime, so I don't update component statuses to termination. Only the overall ready state is changed to "Not Ready."

## 🔨 Related Issues

Implements https://github.com/spiceai/spiceai/issues/4888

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

The approach does not include emitting telemetry for components with the ShuttingDown state. This can be added if needed.

### DIFF
--- a/crates/runtime/src/init/dataset.rs
+++ b/crates/runtime/src/init/dataset.rs
@@ -215,6 +215,11 @@ impl Runtime {
             let connector = match self.load_dataset_connector(Arc::clone(&ds)).await {
                 Ok(connector) => connector,
                 Err(err) => {
+                    if self.status.is_shutdown() {
+                        // should not retry or trace error if runtime is shutting down
+                        return Err(RetryError::permanent(err));
+                    }
+
                     let ds_name = &ds.name;
                     self.status
                         .update_dataset(ds_name, status::ComponentStatus::Error);
@@ -228,6 +233,10 @@ impl Runtime {
                 .register_loaded_dataset(Arc::clone(&ds), connector, None)
                 .await
             {
+                if self.status.is_shutdown() {
+                    // should not retry if runtime is shutting down
+                    return Err(RetryError::permanent(err));
+                }
                 return Err(RetryError::transient(err));
             };
 
--- a/crates/runtime/src/status.rs
+++ b/crates/runtime/src/status.rs
@@ -17,7 +17,10 @@ limitations under the License.
 use std::{
     collections::{HashMap, HashSet},
     fmt::Display,
-    sync::{Arc, RwLock},
+    sync::{
+        atomic::{AtomicBool, Ordering},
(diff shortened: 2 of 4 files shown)


crates/runtime/src/status.rs
```diff
@@ -44,6 +47,9 @@ pub enum ComponentStatus {
 
     /// The component is in the process of refreshing its state
     Refreshing,
+
+    /// The component is in the process of shutting down
+    ShuttingDown,
```
sgrebnov, line 52: Note: this state is not currently used

crates/runtime/src/init/dataset.rs
```diff
@@ -215,6 +215,11 @@ impl Runtime {
             let connector = match self.load_dataset_connector(Arc::clone(&ds)).await {
                 Ok(connector) => connector,
                 Err(err) => {
+                    if self.status.is_shutdown() {
+                        // should not retry or trace error if runtime is shutting down
+                        return Err(RetryError::permanent(err));
```
Sevenannn, line 220: Does other components needs the shutdown detection for breaking out retry loop as well, e.g. catalog

crates/runtime/src/status.rs
```diff
@@ -202,4 +222,9 @@ impl RuntimeStatus {
             })
             .collect()
     }
+
+    /// Sets the runtime to the shutting down state.
+    pub fn mark_shutdown(&self) {
+        self.is_shutdown.store(true, Ordering::Relaxed);
```
Sevenannn, line 228: Do we need a stronger ordering here to guarantee the memory ordering?
Reference: https://doc.rust-lang.org/nomicon/atomics.html
`Relaxed operations are appropriate for things that you definitely want to happen, but don't particularly otherwise care about. For instance, incrementing a counter can be safely done by multiple threads using a relaxed fetch_add if you're not using the counter to synchronize any other accesses.`

crates/runtime/src/init/dataset.rs
sgrebnov: Yeah, good catch - Propose to merge this version and incrementally detect/update other places (I'll do first iteration tomorrow)

crates/runtime/src/status.rs
sgrebnov: The Ordering::Relaxed is fine here IMO as it is a simple boolean check across threads with no need for synchronizing other memory; it is also aligned with what we use for similar cases (like initial_load_completed) and also how opentelemetry implements is_shutdown primitive internally

Sevenannn (approved):

PR #4920, opened 2025-03-06
run spiceai benchmark test using test operator

Sevenannn:
# 🗣 Description

Run spiceai tpch / tpcds tests using test operator.

This PR is based upon changes made in: https://github.com/spiceai/spiceai/pull/4708. Along with minor fixes to make the benchmark compile & not panicking & updates in snapshots.

Also updated to use spicepod name to replace `connector_name` in metrics collection.

Command to run bench
```
cargo run -p testoperator run bench -s path/to/spiced -p path/to/spicecloud.yaml --query-set tpcds --ready-wait 90
```

Example result
```
+--------------------------------------+---------------+---------------+----------------+--------+-----------------+-----------------+------------+-----------------+--------------------------------+-----------------+--------------------+---------------------------+---------------------------+---------------------------+
| run_id                               | started_at    | finished_at   | query_name     | status | min_duration_ms | max_duration_ms | iterations | commit_sha      | branch_name                    | name            | median_duration_ms | percentile_99_duration_ms | percentile_95_duration_ms | percentile_90_duration_ms |
+--------------------------------------+---------------+---------------+----------------+--------+-----------------+-----------------+------------+-----------------+--------------------------------+-----------------+--------------------+---------------------------+---------------------------+---------------------------+
| f7fd737a-5ab1-479e-b52e-e607b51a03df | 1741248411342 | 1741248412750 | tpch_q1        | passed | 1               | 2               | 5          | c632cbb57-dirty | qianqian/spiceai-test-operator | spicecloud_tpch | 1                  | 2                         | 2                         | 2                         |
| f7fd737a-5ab1-479e-b52e-e607b51a03df | 1741248425069 | 1741248432132 | tpch_q10       | passed | 15              | 17              | 5          | c632cbb57-dirty | qianqian/spiceai-test-operator | spicecloud_tpch | 17                 | 17                        | 17                        | 17                        |
```

## 🔨 Related Issues

Part of: https://github.com/spiceai/spiceai/issues/4114

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

Instead of moving all snapshots across we could keep the copy in the previous snapshot directory, and delete previous snapshot when the migration of test operator is finished.


crates/test-framework/src/spicetest/datasets/worker.rs
```diff
@@ -67,98 +77,297 @@ impl SpiceTestQueryWorker {
...
+        &self,
+        query: &(&'static str, &'static str),
+        query_durations: &mut BTreeMap<String, Vec<Duration>>,
+        row_counts: &mut BTreeMap<String, Vec<usize>>,
+        results_snapshot: bool,
+    ) -> Result<()> {
+        let query_start = Instant::now();
+        let result_stream = self.flight_client.query(query.1).await?;
+
+        let records = match result_stream.try_collect::<Vec<_>>().await {
```
peasee: This will introduce large memory usage, by collecting the entire result set into the testoperator memory. If we're only collecting 10 rows, we can watch the stream for the first 10 then only collect the row counts afterwards.

test/spicepods/tpcds/spicecloud.yaml
```diff
@@ -1,6 +1,11 @@
 version: v1
 kind: Spicepod
 name: spicecloud_tpcds
+
+runtime:
+  results_cache:
+    enabled: false
+
```
peasee:
```suggestion
```

test/spicepods/tpch/spicecloud.yaml
```diff
@@ -1,6 +1,11 @@
 version: v1
 kind: Spicepod
 name: spicecloud_tpch
+
+runtime:
+  results_cache:
+    enabled: false
+
```
peasee:
```suggestion
```

Results cache should be left to match the remaining spicepods. We will address results cache usage by adding `cache-control` headers to the flight requests in another PR.

peasee:
There are snapshots in 2 different paths for explain plan and results snapshots? We should consolidate these into one parent. Maybe like `test/snapshots/explain/*.snap` or `test/snapshots/results/*.snap`

crates/test-framework/src/spicetest/datasets/worker.rs
Sevenannn: This is fine IMO - The benchmark should finish the whole query execution, which means it should collect finish running query by collecting all results returned, similar to the behavior in spice sql. If we only collect the 10 rows from stream, we might hide issues relevant to pulling full results from the stream when running the benchmark.

peasee: we're collecting the entire stream immediately into a `Vec<_>`, which allocates everything. AFAIK it is not optimized away by the compiler.

Sevenannn: I agree that collecting the whole stream is not optimal for memory. However the point of benchmark is to similate the real-world workload, and the queries are finished with whole results being collected when the user use spice. For example, when user call [v1/sql endpoint](https://github.com/spiceai/spiceai/blob/b01ee3f9efdebb8c916f6f8d9a69768f2e20606f/crates/runtime/src/http/v1/mod.rs#L143). Additionally the memory usage of the benchmark query results is minimal.

peasee: This is a simple fix to avoid collecting all of the record batches. We still iterate the entire stream to retrieve every batch, but on the client end we simply drop the batch after receiving it. We have no concern for how other users use the batches once received.

I have raised a fix: https://github.com/spiceai/spiceai/pull/4921

This avoids:
* Collecting all record batches into memory
* Iterating the list of batches after collection to collect row counts
* Iterating the list of batches again to collect a subset of rows for snapshotting

The memory usage can also be high, depending on the scale factor being used and queries being run. Consider queries with no `LIMIT`, you risk returning a many-million result set across 8 parallel workers.

Sevenannn: I see what you mean here - Watch the whole stream while only collect the first 10 rows makes sense 👍

Sevenannn: Left minor comment on your PR.

peasee:
Before this merges, I'd like to test the behavior of snapshots using a built `testoperator` binary from the root of the workspace to emulate how it will run in Actions. We might need to make some updates to the action to set `INSTA_WORKSPACE_ROOT` or similar.

Sevenannn:
@peasee I found some issues with the snapshot testing and metrics calculation. Fixes included in this PR: https://github.com/spiceai/spiceai/pull/4935. You can directly merge it and do testing.

peasee (approved):

PR #4921, opened 2025-03-06
fix: Improve row collection loop

peasee:
# 🗣 Description

* Avoids collecting the entire set of record batches into memory in testoperator workers
* Avoids re-looping to collect row counts
* Avoids re-looping to collect a subset of rows for snapshotting

### DIFF
--- a/crates/test-framework/src/spicetest/datasets/worker.rs
+++ b/crates/test-framework/src/spicetest/datasets/worker.rs
@@ -322,7 +322,9 @@ impl SpiceTestQueryWorker {
                             batch.num_rows()
                         };
 
-                        limited_records.push(batch.slice(0, end));
+                        for i in 0..end {
+                            limited_records.push(batch.slice(i, 1));
+                        }
                     }
                 }
             }
--- a/crates/test-framework/src/spicetest/datasets/worker.rs
+++ b/crates/test-framework/src/spicetest/datasets/worker.rs
@@ -21,7 +21,6 @@ use std::{
 };
 
 use anyhow::Result;
-use arrow::array::RecordBatch;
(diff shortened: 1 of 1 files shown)


crates/test-framework/src/spicetest/datasets/worker.rs
```diff
@@ -296,24 +295,40 @@ impl SpiceTestQueryWorker {
...
+                        self.id, query.0, e
+                    );
+                    query_durations.entry(query.0.to_string()).or_default();
+                    return Err(e.into());
+                }
+                Ok(Some(batch)) => {
+                    row_count += batch.num_rows();
+
+                    if limited_records.len() < 10 {
+                        let required_rows = 10 - limited_records.len();
```
Sevenannn, line 318: We could use `try_chunks` and then `into_inner` here to avoid manual calculation. https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_chunks

peasee: I'm not entirely sure I understand the suggestion.

Sevenannn: i mean we could use try_chunks(10) plus .await to collect the first 10 rows of record, and replay the rest of the stream by converting chunked stream to regular stream with into_inner

peasee: That seems very complex, along with iterating a stream on every batch to determine if we've reached our required display row count. This method uses `.len()` on purpose, because it's relatively cheap to run on every received batch.

Sevenannn (approved):

PR #4936, opened 2025-03-07
Convert sequential search code block to parallel async

### LINKED ISSUE #4881: Search across datasets in parallel, not sequentially.
Currently [`v1/search`](https://github.com/spiceai/spiceai/blob/49b80a4e1fc14dafa5c3d4c525210d6973395a49/crates/runtime/src/http/v1/search.rs) and the tool [`document_similarity`](https://github.com/spiceai/spiceai/blob/49b80a4e1fc14dafa5c3d4c525210d6973395a49/crates/runtime/src/tools/builtin/document_similarity.rs#L31) both support searching across multiple datasets (or views). However when this is done, each dataset is queried & processed sequentially, for example like so (for 3 datasets):
1. Compute embedding for query 
2. Use embedding vector to search dataset A
3. Use embedding vector to search dataset B.
4. Use embedding vector to search dataset C.
5. Combine results from across datasets into output payload.

Step 1 must be done first, but there is no reason steps 2, 3 & 4 cannot be done in parallel. They are handled sequentially [here](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/embeddings/vector_search.rs#L902).

Garamda:
# 🗣 Description

Convert current sequential search code to asynchronous parallel one.

## 🔨 Related Issues

Closes #4881

### DIFF
--- a/crates/runtime/src/embeddings/vector_search.rs
+++ b/crates/runtime/src/embeddings/vector_search.rs
@@ -896,60 +896,75 @@ impl VectorSearch {
                 .get_primary_keys_with_overrides(&self.explicit_primary_keys, &tables)
                 .await?;
 
-            let mut response: VectorSearchResult = HashMap::new();
-            let keywords = keywords.iter().map(std::string::String::as_str).collect::<Vec<&str>>();
+            let keywords = keywords.iter().map(String::as_str).collect::<Vec<&str>>();
 
-            for (tbl, search_vectors) in per_table_embeddings {
-                tracing::debug!("Running vector search for table {:#?}", tbl);
+            let search_futures = per_table_embeddings.into_iter().map(|(tbl, search_vectors)| {
+                let keywords = keywords.clone();
                 let primary_keys = table_primary_keys.get(&tbl).map_or(&[] as &[String], |v| v.as_slice());
 
-                // Only support one embedding column per table.
-                let table_provider = self
-                    .df
-                    .get_table(&tbl)
-                    .await
-                    .ok_or(Error::DataSourcesNotFound {
-                        data_source: vec![tbl.clone()],
-                    })?;
+                async move {
+                    tracing::debug!("Running vector search for table {:#?}", tbl);
 
-                let Some(embedding_table) = get_embedding_table(&table_provider).await else {
-                    return Err(Error::CannotVectorSearchDataset {
-                        data_source: tbl.clone()
-                    });
-                };
+                    // Only support one embedding column per table.
+                    let table_provider = self
+                        .df
+                        .get_table(&tbl)
+                        .await
+                        .ok_or(Error::DataSourcesNotFound {
+                            data_source: vec![tbl.clone()],
+                        })?;
 
-                let embedding_columns = embedding_table.get_embedding_columns();
-                let Some(embedding_column) = embedding_columns.first() else {
-                    return Err(Error::NoEmbeddingColumns {
-                        data_source: tbl.clone(),
-                    });
-                };
+                    let Some(embedding_table) = get_embedding_table(&table_provider).await else {
+                        return Err(Error::CannotVectorSearchDataset {
+                            data_source: tbl.clone()
+                        });
+                    };
 
-                if search_vectors.len() != 1 {
-                    return Err(Error::IncorrectNumberOfEmbeddingColumns {
-                        data_source: tbl.clone(),
-                        num_embeddings: search_vectors.len(),
-                    });
-                }
-                match search_vectors.first() {
-                    None => unreachable!(),
-                    Some(embedding) => {
-                        let result = self
-                            .individual_search(
-                                &tbl,
-                                embedding.clone(),
-                                primary_keys,
-                                embedding_column,
-                                embedding_table.is_chunked(embedding_column),
-                                additional_columns,
-                                where_cond.as_ref(),
-                                &keywords,
-                                *limit,
-                            )
-                            .await?;
-                        response.insert(tbl.clone(), result);
+                    let embedding_columns = embedding_table.get_embedding_columns();
+                    let Some(embedding_column) = embedding_columns.first() else {
+                        return Err(Error::NoEmbeddingColumns {
+                            data_source: tbl.clone(),
+                        });
+                    };
+
+                    if search_vectors.len() != 1 {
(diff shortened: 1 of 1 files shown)


crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -896,60 +896,75 @@ impl VectorSearch {
...
+                    };
+
+                    if search_vectors.len() != 1 {
+                        return Err(Error::IncorrectNumberOfEmbeddingColumns {
+                            data_source: tbl.clone(),
+                            num_embeddings: search_vectors.len(),
+                        });
+                    }
+                    match search_vectors.first() {
+                        None => unreachable!(),
```
peasee:
```suggestion
                    match search_vectors.first() {
                        None => return Err(Error::IncorrectNumberOfEmbeddingColumns {
                            data_source: tbl.clone(),
                            num_embeddings: search_vectors.len(),
                        }),
```

I know it comes from the existing code, but this length check is redundant - we can just make it the reachable action for a `None` from `.first()`. There might need to be some work for ownership with `search_vectors`.

crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -896,60 +896,75 @@ impl VectorSearch {
...
-                        data_source: tbl.clone(),
-                        num_embeddings: search_vectors.len(),
-                    });
+            let mut response: VectorSearchResult = HashMap::new();
+            for result in results {
+                match result {
+                    Ok((tbl, result)) => {
+                        response.insert(tbl, result);
+                    },
+                    Err(e) => return Err(e),
```
peasee:
```suggestion
            let response: VectorSearchResult = results.into_iter().collect();
```

crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -896,60 +896,75 @@ impl VectorSearch {
...
+                }
+            }).collect::<Vec<_>>();
 
-                let embedding_columns = embedding_table.get_embedding_columns();
-                let Some(embedding_column) = embedding_columns.first() else {
-                    return Err(Error::NoEmbeddingColumns {
-                        data_source: tbl.clone(),
-                    });
-                };
+            let results = futures::future::join_all(search_futures).await;
```
peasee:
```suggestion
            let results = futures::future::try_join_all(search_futures).await?;
```

Could we use `try_join_all` here? It avoids needing to make a manual loop to pull out the errors later.

peasee (commented):
Some small comments, but otherwise looking good!

crates/runtime/src/embeddings/vector_search.rs
Garamda: Thank you for your suggestion!
I have a question before applying it.

```
  if search_vectors.len() != 1 {
      return Err(Error::IncorrectNumberOfEmbeddingColumns {
          data_source: tbl.clone(),
          num_embeddings: search_vectors.len(),
      });
  }
```

Here, `search_vectors.len()` is not allowed not only to be 0, but also 2,3,4... and so on.
But if we removed this checking, then here we can not handle edge case like below

```
  #[snafu(display("Only one embedding column per table currently supported. Table: {} has {num_embeddings} embeddings", data_source.to_string()))]
  IncorrectNumberOfEmbeddingColumns {
      data_source: TableReference,
      num_embeddings: usize,
  },
```

This says "Only one embedding column per table currently supported."
Is it okay not to check it anymore?

crates/runtime/src/embeddings/vector_search.rs
Garamda: Oh, the error message shows

"Trait `FromIterator<Result<(TableReference, VectorSearchTableResult), Error>>` is not implemented for `VectorSearchResult` [E0277]"

Garamda: Oh nevermind! I have just checked the another comment you made below. Combined with that, the code works and becomes better :)

crates/runtime/src/embeddings/vector_search.rs
peasee: Ah, yes you're right my bad. In that case, let's just add a note to the unreachable to document this 😄 

```suggestion
                    if search_vectors.len() != 1 {
                        return Err(Error::IncorrectNumberOfEmbeddingColumns {
                            data_source: tbl.clone(),
                            num_embeddings: search_vectors.len(),
                        });
                    }
                    match search_vectors.first() {
                        None => unreachable!("Vector search can only be performed on one embedding column, and the column exists"),
```

Garamda: I have applied your suggestions!

peasee:
Looks like E2E tests are failing because it can't access the secrets across the fork 😞 I'm going to clone locally and do some testing

peasee (approved):
Thank you for your work! Ran a couple queries on small-medium datasets, reduced search times from 2s to ~0.8s 🎉

Garamda:
@peasee Thank you for sharing the result of performance improvement test!

Is the small-medium datasets your private data? Or is it publicly accessible?
If it's the latter, would you also let me know where it is?

I can use it to test through multiple datasets more efficiently next time 👍

peasee:
> @peasee Thank you for sharing the result of performance improvement test!
> 
> Is the small-medium datasets your private data? Or is it publicly accessible? If it's the latter, would you also let me know where it is?
> 
> I can use it to test through multiple datasets more efficiently next time 👍

They're public datasets! I tested with both TPCH ([generated with DuckDB](https://duckdb.org/docs/stable/extensions/tpch.html), connecting with the DuckDB connector), and the Spice.AI Docs datasets which are available at `spice.ai/spiceai/docs/datasets/spice.spiceai.cookbook` and `spice.ai/spiceai/docs/datasets/spice.spiceai.docs` using the Spice.AI connector (you'll need an API key from [spice.ai](https://spice.ai))

The Spice.AI datasets use [passthrough embeddings](https://spiceai.org/docs/components/embeddings#passthrough-embeddings) with the `openai:text-embedding-3-small` embedding model. The DuckDB TPCH dataset you can generate your own embeddings with your model of choice!

Garamda:
@peasee I appreciate your detailed explanation! I will utilize those from now on :)

PR #4938, opened 2025-03-08
fix: Throughput metric calculation

peasee:
# 🗣 Description

Fixes some bugs with the throughput metric calculation:

1. Uses the test duration for the time factor in the calculation. The previous method used the sum of query durations, which isn't representative of wall clock time because these queries can run in parallel across the worker threads.
2. Adjusts the QpH based on the number of query sets completed. The original QpH calculation relies on the fact that only one round of queries is completed. As a result, if 2 rounds are completed the time factor increases with no other adjustment - resulting in an incorrect (lower) QpH.

## 🔨 Related Issues

* Part of #4114

### DIFF
--- a/crates/test-framework/src/spicetest/datasets/mod.rs
+++ b/crates/test-framework/src/spicetest/datasets/mod.rs
@@ -288,16 +288,6 @@ impl SpiceTest<Completed> {
         self.state.query_durations.values().flatten().copied().sum()
     }
 
-    pub fn get_average_total_queries_executed(&self) -> Result<u32> {
-        // total query count across all workers
-        let total_query_count = self.state.query_durations.values().flatten().count();
-
-        // average query count per worker
-        Ok(u32::try_from(
-            total_query_count / self.state.parallel_count,
-        )?)
-    }
-
     #[must_use]
     pub fn get_test_duration(&self) -> Duration {
         self.state.test_duration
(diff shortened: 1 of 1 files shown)


crates/test-framework/src/spicetest/datasets/mod.rs
```diff
@@ -284,6 +288,16 @@ impl SpiceTest<Completed> {
         self.state.query_durations.values().flatten().copied().sum()
     }
 
+    pub fn get_average_total_queries_executed(&self) -> Result<u32> {
+        // total query count across all workers
+        let total_query_count = self.state.query_durations.values().flatten().count();
+
+        // average query count per worker
+        Ok(u32::try_from(
+            total_query_count / self.state.parallel_count,
```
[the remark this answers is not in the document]
peasee: The state machine checks the parallel count is nonzero before proceeding to this state.

lukekim: Can this be a percentile? Really want to avoid averages. It sets a bad precedent.

crates/test-framework/src/spicetest/datasets/mod.rs
```diff
@@ -292,9 +296,22 @@ impl SpiceTest<Completed> {
...
+        // e.g. query set count of 5, means our calculated QpH needs to be times 5 because we ran 5 query sets
+        // this adjusts for a longer test duration as a result of running multiple query sets, which would otherwise reduce the QpH
+        let end_condition_modifier = match self.state.end_condition {
+            EndCondition::Duration(_) => {
+                return Err(anyhow::anyhow!("Throughput metric calculation for duration-based tests is not supported. Use a QuerySetCompleted test instead."))
+            }
+            EndCondition::QuerySetCompleted(count) => f64::from(u32::try_from(count)?),
+        };
+
+        Ok(lhs / self.state.test_duration.as_secs_f64() * scale * end_condition_modifier)
```
phillipleblanc: nit: add parenthesis to make it clear that the multiplication happens after the division.

phillipleblanc (dismissed):

Sevenannn (approved):

PR #4942, opened 2025-03-09
Improve servers shutdown sequence during runtime termination

sgrebnov:
# 🗣 Description

PR implements `spawn_cancellable_task` functionality to spawn a Tokio task and return `CancellableTaskHandle` that allows external control over a task's execution, supporting both graceful cancellation via the provided [`CancellationToken`] and forced termination. 

Runtime server initialization has been updated to create and register each background task using `CancellableTaskHandle`, with `Runtime.shutdown` performing shutdown in the following order:

```
1. Shutdown initiated (Ctrl-C)
2. HTTP API including /ready, /status, /sql turns 503 except /health which continue 200
3. Components shutdown
   - Step1: Flight, OpenTelemetry (metrics ingestion) servers graceful shutdown: wait until in-progress requests are completed, stop serving new requests
   - Step2: Datasets, Models, etc shutdown (currently Runtime.shutdown / Datafusion.shutdown)
5. HTTP API complete shutdown
6. Metrics (9090) endpoint shutdown
7. All resources clean up (termination)
```

Also, Tonic based API (Flight, OpenTelemetry) has been updated with support of `CancellationToken` for more graceful shutdown.

## 🔨 Related Issues

Implements https://github.com/spiceai/spiceai/issues/4888

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -15,6 +15,7 @@ limitations under the License.
 */
 #![allow(clippy::missing_errors_doc)]
 
+use std::future::Future;
 use std::net::SocketAddr;
 use std::time::Duration;
 use std::{collections::HashMap, sync::Arc};
@@ -278,6 +279,12 @@ pub enum Error {
     ComponentError { source: component::Error },
 }
 
+const HTTP_SERVER: &str = "http_server";
+const METRICS_SERVER: &str = "metrics_server";
+const FLIGHT_SERVER: &str = "flight_server";
+const OPENTELEMETRY_SERVER: &str = "opentelemetry_server";
+const PODS_WATCHER: &str = "pods_watcher";
+
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
 #[derive(Clone, Copy)]
@@ -306,7 +313,7 @@ pub struct Runtime {
 
     status: Arc<status::RuntimeStatus>,
 
-    running_components: Arc<RwLock<HashMap<String, ManagedTaskHandle>>>,
+    server_components: Arc<RwLock<HashMap<String, ManagedTaskHandle>>>,
 }
 
 impl Runtime {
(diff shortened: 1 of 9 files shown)


crates/runtime/src/lib.rs
```diff
@@ -431,44 +486,36 @@ impl Runtime {
...
-                    futures::future::pending().await
-                }
-            } => {
-                pods_watcher_res.context(UnableToInitializePodsWatcherSnafu)
-            },
-            () = shutdown_signal() => {
-                tracing::info!("Goodbye!");
-                Ok(())
-            },
+        // Start Spicepod watcher
```
sgrebnov, line 489: The Spicepod watcher is not technically a server component. To minimize the scope and complexity of changes, the original behavior is being retained. This can be improved next.

crates/runtime/src/lib.rs
```diff
@@ -377,48 +396,84 @@ impl Runtime {
...
-            Arc::clone(&self),
-            config.clone().into(),
-            tls_config.clone(),
-            http_auth,
-        ));
-
-        // Spawn the metrics server in the background
+        let self_ref = Arc::clone(&self);
+
+        let http_future = self
```
sgrebnov, line 405: Axum supports `with_graceful_shutdown` that should be added similar to Tonic servers (Flight, OpenTelemetry). Will add next

crates/runtime/src/managed_task.rs
```diff
@@ -0,0 +1,108 @@
...
+    let (notify_abort_task, on_abort_task) = oneshot::channel();
+    let (notify_task_completed, on_task_completed) = oneshot::channel();
+
+    let task_handle = ManagedTaskHandle {
+        notify_abort_task,
+        cancellation_token: task_cancellation,
+        on_task_completed,
+    };
+
+    let handle: JoinHandle<Result<(), Error>> = tokio::task::spawn(async move {
```
sgrebnov: `JoinHandle` can't be shared and awaited at the same time, so we create a future based on it and use channels to abort execution and signal when the task is completed.

crates/runtime/src/managed_task.rs
```diff
@@ -0,0 +1,108 @@
...
+use crate::{Error, FailedToExecuteTaskSnafu};
+
+/// A handle for managing the lifecycle of a spawned task.
+///
+/// Allows external control over a task's execution, supporting
+/// both graceful cancellation via provided [`CancellationToken`] and forced termination.
+///
+/// - If a [`CancellationToken`] is provided, it enables graceful shutdown.
+/// - If the task does not exit within the allowed time after termination request, it is forcefully aborted.
+pub(crate) struct ManagedTaskHandle {
```
phillipleblanc: This isn't really a managed task - its a cancellable task. There is no management of the task other than its ability to be cancelled

crates/runtime/src/managed_task.rs
```diff
@@ -0,0 +1,108 @@
...
+            _ = self.on_task_completed => {}
+        };
+    }
+}
+
+/// Spawns a managed task with termination support.
+///
+/// Returns a future that resolves when the task completes or is canceled,
+/// along with a [`ManagedTaskHandle`] for external task control.
+pub(crate) fn spawn_managed_task<F>(
```
phillipleblanc: Add some tests in this file to ensure the shutdown/abort behavior is correct

crates/runtime/src/managed_task.rs
```diff
@@ -0,0 +1,108 @@
...
+/// - If a [`CancellationToken`] is provided, it enables graceful shutdown.
+/// - If the task does not exit within the allowed time after termination request, it is forcefully aborted.
+pub(crate) struct ManagedTaskHandle {
+    notify_abort_task: oneshot::Sender<()>,
+    cancellation_token: Option<CancellationToken>,
+    on_task_completed: oneshot::Receiver<()>,
+}
+
+impl ManagedTaskHandle {
+    pub async fn shutdown(mut self, timeout: Duration) {
```
phillipleblanc: If this is a cancellable task, then this should probably be named `cancel` and not `shutdown`

crates/runtime/src/lib.rs
```diff
@@ -610,15 +658,72 @@ impl Runtime {
...
+        tracing::info!("Shutting down runtime...");
         self.status.mark_shutdown();
+
+        // shutdown all running components except the HTTP and Metrics servers
+        let mut running_components = self.server_components.write().await;
+
+        // HTTP and METRICS servers must be shutdown last
+        let (to_shutdown, to_shutdown_last) = running_components
+            .drain()
+            .partition::<Vec<_>, _>(|(name, _)| *name != HTTP_SERVER && *name != METRICS_SERVER);
```
```suggestion
        // HTTP and METRICS servers must be shutdown last
        let mut first_shutdown_group = Vec::new();
        let mut last_shutdown_group = Vec::new();
        
        for (name, handle) in running_components.drain() {
            match name.as_str() {
                HTTP_SERVER | METRICS_SERVER => last_shutdown_group.push((name, handle)),
                _ => first_shutdown_group.push((name, handle)),
            }
        }
```

I think this is slightly cleaner/easier to understand and allows us to create more groups in the future if we need it.

crates/runtime/src/lib.rs
```diff
@@ -610,15 +658,72 @@ impl Runtime {
...
+
+        join_all(shutdown_futures).await;
+
         // Clean up DataFusion first as there could be datasets loading and accessing registries below.
         self.df.shutdown().await;
         dataconnector::unregister_all().await;
         catalogconnector::unregister_all().await;
         dataaccelerator::unregister_all().await;
         tools::factory::unregister_all_factories().await;
         document_parse::unregister_all().await;
```
phillipleblanc, line 699: Not for this PR, but I think we should move to the model where all static registries are tied to the runtime object instead of being static. This code essentially prevents us from having multiple runtimes defined, which is useful for testing.

crates/runtime/src/managed_task.rs
sgrebnov: Renamed to CancellableTaskHandle, thank you

crates/runtime/src/managed_task.rs
sgrebnov: Updated, thank you

crates/runtime/src/lib.rs
sgrebnov: Thank you, updated

crates/runtime/src/managed_task.rs
sgrebnov: Added 6 unit tests 👍

phillipleblanc (approved):

PR #4949, opened 2025-03-10
feat: Add appendable data sources for the testoperator

peasee:
# 🗣 Description

* Adds `run append` to run an append-mode test in the `testoperator`
* Adds support for a `AppendableSource`, currently only implementing for the File connector
* Moves the root `spiced_instance` into an `Option<_>` on the `SpiceTest` - append needs to setup data before spiced is running, and this change helps accomplish that by not requiring a spiced instance while the state is `NotStarted` for an append test.
* Adds a `SpiceTest` for append-mode tests, because they'll require some different end-of-test checks - there might be opportunity for code de-dupe which I'll look into with follow-up PRs.

Follow-up PRs will implement TPCDS and ClickBench source loading, as well as validation that the append succeeded at the end of the test.

## 🔨 Related Issues

* Part of #4929
* Part of #4114


crates/test-framework/src/spicetest/append/sources/file.rs
```diff
@@ -0,0 +1,140 @@
...
+#[async_trait]
+impl AppendableSource for FileAppendableSource {
+    async fn setup(&self, config: &AppendConfig) -> Result<()> {
+        if std::fs::exists(&self.dest_db_file)? {
+            std::fs::remove_file(&self.dest_db_file)?;
+        }
+
+        for TableWithColumn { name, .. } in &self.tables {
+            let parquet_path = config.temp_directory.join(format!("{name}.parquet"));
+            if std::fs::exists(&parquet_path)? {
```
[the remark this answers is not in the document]
peasee: `std::fs::exists` is a `Result<bool>` 🤔

crates/test-framework/Cargo.toml
```diff
@@ -14,6 +14,12 @@ app = { path = "../app" }
 arrow = { workspace = true, features = ["prettyprint"] }
 async-openai.workspace = true
 clap.workspace = true
+duckdb = { workspace = true, features = [
```
phillipleblanc, line 17: This will significantly increase the amount of time it takes to build the testoperator binary - is this really required?

crates/test-framework/src/queries/mod.rs
```diff
@@ -26,6 +29,18 @@ pub enum QuerySet {
     Clickbench,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct TableWithColumn {
```
phillipleblanc: This isn't very descriptive - TableWithTimeColumn?

phillipleblanc (dismissed):

crates/test-framework/Cargo.toml
peasee: Yeah, I had a similar concern. We could pull in `tpch-dbgen` manually, clone the repo and run shell commands to generate in steps. I ran some builds, and DuckDB increased builds from 4m 34s to 7m 52s.

In some ways, I'm okay with the build time trade-off for the reduction in complexity. Using DuckDB gives us the benefits of:
1. Easily adding a new time column to existing data
2. Easily generating TPCH/DS data without downloading the associated CLI tool and running shell commands
3. Easily partitioning existing parquet files (e.g. for ClickBench)
4. Easily generating performant parquet files vs CSV from TPCH/DS tools (which also helps enforce correct column types)

TPCH and TPCDS benefit from 1, 2 and 4. ClickBench benefits from 1 and 3.

phillipleblanc (approved):

PR #4956, opened 2025-03-10
fix: Redundant results snapshotting

peasee:
# 🗣 Description

* Fixes redundant results snapshotting, which runs on every query iteration on every query worker.
* Makes snapshotting only occur from the first worker, to reduce lock poisons.
* Adds spicepods for cache scenarios

## 🔨 Related Issues

* Part of #4114

### DIFF
--- a/crates/test-framework/src/spicetest/datasets/worker.rs
+++ b/crates/test-framework/src/spicetest/datasets/worker.rs
@@ -160,7 +160,8 @@ impl SpiceTestQueryWorker {
 
                         let snapshot_results = self
                             .results_snapshot_predicate
-                            .is_some_and(|predicate| predicate(query.0));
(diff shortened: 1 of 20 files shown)


phillipleblanc (dismissed):

crates/test-framework/src/spicetest/datasets/worker.rs
```diff
@@ -153,14 +153,17 @@ impl SpiceTestQueryWorker {
                 }
                 EndCondition::QuerySetCompleted(target_count) => {
                     // For QuerySetCompleted, run each query target_count times before moving to next
+                    // let start = SystemTime::now();
```
phillipleblanc:
```suggestion
```

phillipleblanc (approved):

PR #4988, opened 2025-03-11
`tool` crate and other tool improvements.

Jeadie:
# 🗣 Description
See individual commits for groupings. Rough changes:
 - Make crate `tools`, and provide `Runtime`object on initialisation to tools.
 - Retry failure handling for connecting to MCP clients.
 - Use [modelcontextprotocol/rust-sdk](https://github.com/modelcontextprotocol/rust-sdk) (in short term, use PR patch to account for @ewgenius change in [spiceai/goose](https://github.com/spiceai/goose/commit/5bacbba75925d76cbda53421115151d9097bed17)).


crates/runtime/Cargo.toml
```diff
@@ -76,8 +76,8 @@ keyring = { version = "3.6.2", features = [
 ], optional = true }
 llms = { path = "../llms" }
 logos = "0.15.0"
-mcp-client = { git = "https://github.com/block/goose", package = "mcp-client", optional = true }
-mcp-core = { git = "https://github.com/block/goose", package = "mcp-core", optional = true }
+mcp-client = { git = "https://github.com/Jeadie/mcp-rust-sdk.git", rev = "060cdd2d9e23d129894160e26c66cc82b3fe37eb", package = "mcp-client", optional = true }
+mcp-core = { git = "https://github.com/Jeadie/mcp-rust-sdk.git", rev = "060cdd2d9e23d129894160e26c66cc82b3fe37eb", package = "mcp-core", optional = true }
```
phillipleblanc: Make a PR into https://github.com/spiceai/modelcontextprotocol-rust-sdk/tree/spiceai and raise an issue on yourself to track merging the upstream change and reverting back to the upstream crate.

crates/runtime/src/tools/mcp/catalog.rs
```diff
@@ -23,29 +23,38 @@ use mcp_client::{
...
 };
 
+const HEARTBEAT_INTERVAL_SECONDS: u64 = 30; // 30 seconds
+
 pub(crate) struct McpToolCatalog {
     client: Arc<RwLock<Box<dyn McpClientTrait>>>,
 
     /// User defined name & description, not from underlying MCP.
     name: String,
+    heartbeat_task: tokio::task::JoinHandle<()>,
```
phillipleblanc: This can use the graceful CancelleableTask that was added recently in https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/cancellable_task.rs

crates/runtime/src/tools/mcp/catalog.rs
```diff
@@ -41,12 +42,21 @@ pub(crate) struct McpToolCatalog {
...
+
+    // Must be `Option` to allow taking a mutable reference to the `heartbeat_task` field in the `Drop` implementation.
+    heartbeat_task: Option<CancellableTaskHandle>,
 }
 
 impl Drop for McpToolCatalog {
     fn drop(&mut self) {
-        self.heartbeat_task.abort();
+        if let Some(heartbeat) = self.heartbeat_task.take() {
+            tokio::spawn(heartbeat.cancel(Duration::from_secs(15)));
```
phillipleblanc: This will panic if the tokio runtime is shutdown - which there is a good chance it might be when this is called in a shutdown routine.

Ideally we don't need to do any async work here - is there a way we can just send a cancel signal without waiting for the response?

Jeadie: Created `cancel_without_wait`.

phillipleblanc (dismissed):

Jeadie:
Favouring #5121

PR #5010, opened 2025-03-12
Improve Vector Search performance for large content w/o primary key defined

### LINKED ISSUE #5000: Bug: Vector search crashes when used with a large dataset
## Describe the bug

Loaded 114 rows (199.01 MiB) for dataset my_data in 16m 29s 615ms.

>Spice.ai OSS CLI v1.1.0-unstable-build.59f266ed2
search> test
2025/03/11 14:43:33 ERROR search failed error="Error occurred interacting with datafusion: Execution error: Failed to execute DuckDB query: Query execution failed.\nError code 1: Unknown error code\nFor details, refer to the DuckDB manual: https://duckdb.org/docs/"

## To Reproduce

1. Embed dataset with large content
2. Perform search using `spice search`
3. Search is failed with error

```
2025/03/11 14:43:33 ERROR search failed error="Error occurred interacting with datafusion: Execution error: Failed to execute DuckDB query: Query execution failed.\nError code 1: Unknown error code\nFor details, refer to the DuckDB manual: https://duckdb.org/docs/"
```

sgrebnov:
# 🗣 Description

Existing search logic unnests search content based on the number of chunks corresponding to each record if there is no primary key specified as part of embedding definition. As a result, each document appears in search results 100–200 times (num chunks per document), leading to significantly increased resource allocation. Content column is also used to match search results (`FROM ranked_docs rd JOIN my_data t ON rd.content = t.content`) with the final record which is not very efficient as content could be large.

```sql
WITH distances as (
    SELECT
        content, <-- Content is duplicated for each offset due to unnest
        unnest(content_offset) AS offset,
        array_cosine_distance(unnest(content_embedding), [-0.009877119, .. -0.019920357]::FLOAT[1536]) AS dist
    FROM my_data    
),
ranks as (
    SELECT
        content,
        distances.offset,
        distances.dist,
        ROW_NUMBER() OVER (PARTITION BY (content) ORDER BY distances.dist ASC) AS chunk_rank
    FROM distances
),
ranked_docs as (
    select content, ranks.dist, ranks.offset
    from ranks
    WHERE chunk_rank = 1
    ORDER by dist ASC
    LIMIT 10
)
SELECT
    substring(t.content, rd.offset[1], rd.offset[2] - rd.offset[1]) AS content_chunk,
    rd.dist
FROM ranked_docs rd
JOIN my_data t ON rd.content = t.content <-- This is slow
```

This PR proposes using an additional temporary CTE with a surrogate ID based on the row number that can be used as a unique identifier to avoid content duplication and content based matching.

With this change, a search that previously crashed after approximately 30 seconds now executes in 200ms.
- Fixes https://github.com/spiceai/spiceai/issues/5000

## 🔨 Related Issues

Fixes https://github.com/spiceai/spiceai/issues/5000

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

1. Vector search logic becomes more complex and not very easy to read, we might think how to refactor / improve this.
2. If there are additional columns defined we must include them into temporary CTE scan now, previously additional columns were fetched based on final join with original row (we can't match original row anymore as we operate using new ID (or we need to fallback to original content matching or embeddings matching) - I was thinking if we can use the same generated ID for final matching so we don't need to fetch content initially and also include extra columns but I don't think this will be robust, at least for Arrow optimizations / partitioning and parallel execution)

### DIFF
--- a/crates/runtime/src/embeddings/vector_search.rs
+++ b/crates/runtime/src/embeddings/vector_search.rs
@@ -127,8 +127,11 @@ impl Display for RetrievalLimit {
     }
 }
 
+// Distance column name for the vector search query.
 static VECTOR_DISTANCE_COLUMN_NAME: &str = "dist";
+// Surrogate unique identifier name to use when no primary keys are provided.
 static VSS_TEMP_GEN_ID_COLUMN: &str = "vss_temp_gen_id";
+// Temporary table name to provide surrogate unique id for vector search query when no primary keys are provided.
 static VSS_TEMP_TABLE_NAME: &str = "vss_temp_table";
 
 #[derive(Debug, Clone, JsonSchema, Serialize, Deserialize)]
--- a/crates/runtime/src/embeddings/vector_search.rs
+++ b/crates/runtime/src/embeddings/vector_search.rs
@@ -128,6 +128,8 @@ impl Display for RetrievalLimit {
 }
 
 static VECTOR_DISTANCE_COLUMN_NAME: &str = "dist";
+static VSS_TEMP_GEN_ID_COLUMN: &str = "vss_temp_gen_id";
+static VSS_TEMP_TABLE_NAME: &str = "vss_temp_table";
 
 #[derive(Debug, Clone, JsonSchema, Serialize, Deserialize)]
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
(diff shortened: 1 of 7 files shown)


crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -128,6 +128,8 @@ impl Display for RetrievalLimit {
 }
 
 static VECTOR_DISTANCE_COLUMN_NAME: &str = "dist";
+static VSS_TEMP_GEN_ID_COLUMN: &str = "vss_temp_gen_id";
+static VSS_TEMP_TABLE_NAME: &str = "vss_temp_table";
```
phillipleblanc: nit add comments explaining what these constants represent

phillipleblanc (dismissed):

crates/runtime/src/embeddings/vector_search.rs
sgrebnov: Added - thank you!

break-the-build (approved):

PR #5018, opened 2025-03-13
feat: Update testoperator workflows for automatic snapshot updates

peasee:
# 🗣 Description

* Updates the workflows for testoperator to generate snapshots automatically, and raise a PR.
* Also supports snapshot updates through dispatcher (to bulk-update snapshots).
* Feature flags out the append test from `test-framework` to reduce most build times.

## 🔨 Related Issues

* Part of #4981


crates/test-framework/Cargo.toml
```diff
@@ -41,3 +41,28 @@ uuid = { workspace = true, features = ["v4"] }
 
 [target.'cfg(not(windows))'.dependencies]
 nix = { version = "0.29", features = ["signal"] }
+
+[features]
+anonymous_telemetry = []
```
phillipleblanc, line 46: Why are all these required?

peasee: Similar to why I added them for `testoperator`. When you add any feature to a `Cargo.toml`, `rust-analyzer` tries applying all of your default features from `settings.json` to the crate. Because I added an append feature, I added the rest of these to make `rust-analyzer` continue to work because you can't apply member-specific overrides in a workspace 😢

phillipleblanc (approved):

PR #5024, opened 2025-03-13
Include test for primary key passed as additional column in Vector Search

sgrebnov:
# 🗣 Description

Cover case when primary key is duplicated  via additional column in Vector Search

### DIFF
--- a/crates/runtime/tests/models/hf.rs
+++ b/crates/runtime/tests/models/hf.rs
@@ -297,7 +297,7 @@ mod search {
                         body: json!({
                             "text": "friends",
                             "datasets": ["catalog_page_with_chunking"],
-                            "additional_columns": ["cp_department", "cp_description"],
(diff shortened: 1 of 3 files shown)


crates/runtime/tests/models/hf.rs
```diff
@@ -297,7 +297,7 @@ mod search {
                         body: json!({
                             "text": "friends",
                             "datasets": ["catalog_page_with_chunking"],
-                            "additional_columns": ["cp_department", "cp_description"],
+                            "additional_columns": ["cp_catalog_page_sk", "cp_department", "cp_description"],
```
lukekim, line 300: pk?

lukekim (approved):

crates/runtime/tests/models/hf.rs
sgrebnov: it is actually the TPCS naming, confirm it is a `cp_catalog_page_sk` / correct

PR #5041, opened 2025-03-14
Include local files in generated Spicepod package

sgrebnov:
# 🗣 Description

Adds support for including local files in the generated Spicepod package. If a dataset refers to a local folder, all files from the folder are included.

Note: `HttpStore` can't be used to list files from the GitHub repo folder, so we use a separate GitHub REST client to fetch repo metadata and list files.

## ⛓️‍💥 Breaking Change


crates/runtime/src/objectstore/github.rs
```diff
@@ -111,10 +126,51 @@ impl ObjectStore for GitHubRawObjectStore {
 
     fn list(
```
phillipleblanc, line 127: Can we include an integration test for this?

phillipleblanc (dismissed):

lukekim (approved):

crates/runtime/src/objectstore/github.rs
sgrebnov: Thank you Phillip - confirm we have a test for `store.list` function (included into existing `test_get_opts`):

```rust
#[cfg(test)]
mod tests {
    use super::*;
    use futures::StreamExt;

    #[tokio::test]
    async fn test_get_opts() {
        let store = GitHubRawObjectStore::try_new("spiceai", "spiceai", "refs/heads/trunk", None)
            .expect("failed to create store");
        let result = store
            .get_opts(&Path::from("README.md"), GetOptions::default())
            .await
            .expect("failed to get README");
        println!("{result:?}");

        let files: Vec<_> = store
            .list(Some(&Path::from("docs/release_notes/rc")))
            .collect::<Vec<_>>()
            .await;
        println!("{files:?}");
        assert!(!files.is_empty());
    }
}
```

PR #5051, opened 2025-03-16
fix `v1/models` payload on CLI side

Jeadie:
# 🗣 Description
 - Difference in expected payload between spiced and spice CLI for `v1/models`.
 - Causes error in `spice chat` when model is not preselected.
    ```shell
    >>> spice chat
    Spice.ai OSS CLI v1.0.4 (https://spiceai.org/docs/cli)
    2025/03/17 08:26:20 ERROR could not list models error="error decoding response: json: cannot unmarshal object into Go value of type []api.Model"
    ```


Jeadie:
We should include this in any 1.0.X

Jeadie:
Not an issue. Issue with incompatible versions of spiced and spice CLI

PR #5083, opened 2025-03-17
imap provider: attachments / headers / content-section parsing.

### LINKED ISSUE #4605: List email attachment filenames
List the attachments as a list of the attachments filenames, if any

### LINKED ISSUE #4791: MIME-encoded content is decoded automatically
MIME-encoded content is decoded automatically

### LINKED ISSUE #4792: Email headers are separated from email content
Email headers are separated from email content

### LINKED ISSUE #4793: Email attachments can be downloaded when accelerated
Email attachments can be downloaded when accelerated

benrussell:
# 🗣 Description

Extend imap provider with MIME parsing for attachments and content-sections.

resolves #4605 - "List email attachment filenames"

resolves #4791 ? - HTML/Text content sections are extracted

resolves #4792 - Add support for parsing email headers.

resolves #4793 - parse attachments blobs

## 📄 Documentation Updates

Adds new fields to the schema

## 🤔 Concerns

Uses the mail-parser crate: https://crates.io/crates/mail-parser
Found a bug with the attachment name decoding for the upstream crate mail-parser.
Will need to patch and submit for upstream crate.
Added a unit test that relies on the broken behaviour for now.

MIME attachment rules allow for an edge case where an attachment does not require either a mime_type field OR a filename.
Have dealt with this by assigning the default mime type of "application/octet-stream".
This mime-type is used to prevent a user-agent from attempting to display binary data as text to the end user.


crates/data_components/src/imap/attachments.rs
```diff
@@ -0,0 +1,367 @@
...
+                    ret.push(
+                        AttachmentInfo{
+                            filename: None,
+                            mime_type: Some(t.clone()),
+                        }
+                    )
+                },
+                (None,None) => {
+                    //panic!("no filename and no mime type.");
+                    //FIXME: This is a legal MIME state. How to handle for Spice?
```
peasee: I think for the case where there is no filename, regardless of MIME type, we could introduce a new connector parameter to control whether we log a warning or error like `invalid_attachment_action: warn`.

It could accept values similar to the existing [`unsupported_type_action`](https://spiceai.org/docs/reference/spicepod/datasets#unsupported_type_action) (`warn`, `error`, `ignore`), but it would be a [connector parameter](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/dataconnector/imap.rs#L41) instead of a dataset parameter and behavior would be basically the same.

I think for most cases a warning would suffice, but allowing an error here could be useful.

crates/data_components/src/imap/attachments.rs
```diff
@@ -0,0 +1,367 @@
...
+                }else{
+                    //unable to extract mime type
+                    mime_types.push(None);
+                }
+
+                //FIXME: This might be a good place to deal with None,None case where there's no filename AND no mime_type
+
+            } //loop attachments for email
+
+        } //can we parse this email?
```
peasee: this could be a good place to add an `else {}` and handle based on the `invalid_attachment_action`

crates/data_components/src/imap/provider.rs
```diff
@@ -112,6 +116,21 @@ impl TableProvider for ImapTableProvider {
...
+                DataType::List(Arc::new(Field::new("item", DataType::Utf8, true))),
+            true)
+            );
+
+            fields.push(
+                Field::new(
+                    "attachment_mime_type",
+                    DataType::List(Arc::new(Field::new("item", DataType::Utf8, true))),
+                    true)
+            );
```
peasee: I think we should use a struct field here, like `attachments`. This also avoids us needing to keep separate vecs for filenames and mime type, so should help simplify some of the parsing where we compare filename vs mime vec length.

E.g.

```
Field::new(
           "attachments",
           DataType::List(Arc::new(Field::new_list_field(
               DataType::Struct(Fields::from(vec![
                   Field::new("filename", DataType::Utf8, true),
                   Field::new("mime_type", DataType::Utf8, true),
               ])),
               true,
           ))),
           true,
       )
```

crates/data_components/src/imap/attachments.rs
```diff
@@ -0,0 +1,367 @@
...
+    pub filename: Option<String>,
+}
+
+
+#[derive(Debug)]
+#[derive(PartialEq)]
+pub struct AttachmentRecords{
+    pub mime_types: Vec<Option<String>>,
+    pub filenames: Vec<Option<String>>,
+}
```
peasee: if we take a struct as our field, this gets simpler

```
pub struct AttachmentRecords {
    pub attachments: Vec<AttachmentInfo>,
}
```

crates/data_components/src/imap/mod.rs
```diff
@@ -125,6 +131,12 @@ impl ImapTableProvider {
 
         if self.fetch_content {
             fields.push(Arc::new(StringArray::from(bodies)));
+
+            //FIXME: not sure if this is the correct place to add our records to the fields vec
```
peasee: 💯 I think we would put attachment bodies here, but because it's just metadata at the moment IMO we could include it all the time. Even when we're getting attachment bodies, I think we should still support attachment metadata when `self.fetch_content == false`.

peasee (commented):
Thanks for working on this, I'm liking it! A couple comments around error handling and the field type(s).

crates/data_components/src/imap/attachments.rs
benrussell: Have added notes to review the Email MIME RFC. 

IIRC filename= param for attachments is a "suggested default" for mail clients.

The RFC may also define that bare attachment records such as this are plain text. 

If the above is the case perhaps a default to untitled.txt is a useful path.


Connector param as suggested may be the best path for a Spice provider.

crates/data_components/src/imap/attachments.rs
benrussell: Code LOC affected has changed with refactor. Tagged with FIXME's so it's easy to find in latest patch.

crates/data_components/src/imap/provider.rs
benrussell: Nice. I wasn't aware that this is possible in Arrow.

I have refactored appropriately and created a new `fn build_listarray_for_attachments(...)` that hopefully does the right thing.

I'm not up to speed with creating Arrow data structs so I've tagged the new fn with a FIXME and appropriate notes.

crates/data_components/src/imap/attachments.rs
benrussell: I had started with this but split it into the two vectors and a tuple integration for easier integration with the existing Arrow examples.

As you can appreciate, it took extra work to zip() the two Vecs back together in the AttachmentRecords.collate() fn which is now dead and gone.

All good.

crates/data_components/src/imap/mod.rs
benrussell: Have made some changes to suit this feedback also.

Have left a note as I don't believe the bodies data is available without a fetch_content (at least it's not as Rust vars) so that will need some consideration.

benrussell:
Added code to parse and optionally store attachment blobs as decoded files.

crates/data_components/src/imap/attachments.rs
benrussell: Defaulted mime_type to application/octet-stream.

Have not added connector param for raising warning.

Have left marker in code so it can be addressed if it's still wanted.

crates/data_components/src/imap/mod.rs
```diff
@@ -74,6 +82,105 @@ fn build_listarray_for_strings(values: Vec<Option<Vec<Option<String>>>>) -> List
...
+    );
+
+    let mut list_builder = ListBuilder::new(struct_builder);
+
+    for attachment_list in values {
+        if let Some(attachments) = attachment_list {
+            for attachment in attachments {
+                let struct_builder = list_builder.values();
+
+                //FIXME: the expect() msgs in this code seem like they could be improved? (attachments)
```
peasee: Yes, we typically don't merge code with `.expect()` unless used in tests. We should make some new errors and wrap these instead with `.context()`. The signature of this function would need to change to `Result<ListArray, Error>`

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+        let _content_records = ContentSectionRecords::try_from( &message ).unwrap();
+        //println!("\ncontent_records = {:#?}", content_records);
+
+
+        let attachment_records = AttachmentRecords::try_from( &message ).unwrap();
+
+        let extracted_set = attachment_records.attachments;
+        // for rec in &collated{
+        //     println!("-rec: {:?}", rec);
+        // }
```
phillipleblanc:
```suggestion
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+    // is used as the first component of the reconstructed long filename.
+
+    println!("\nutf8 check");
+
+    let eml_msg = MessageParser::default().parse(&raw_email);
+
+    if let Some(message) = eml_msg {
+        //println!("message html_body_count: {}", message.html_body_count());
+        let _content_records = ContentSectionRecords::try_from( &message ).unwrap();
+        //println!("\ncontent_records = {:#?}", content_records);
```
phillipleblanc:
```suggestion
        let _content_records = ContentSectionRecords::try_from( &message ).expect("Failed to parse message");
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+    let raw_email = include_str!("test_data/no_attachments.txt").to_string(); // Load an email file
+
+    println!("\nno attachments");
+
+    let eml_msg = MessageParser::default().parse(&raw_email);
+
+    if let Some(message) = eml_msg {
+        //println!("message html_body_count: {}", message.html_body_count());
+        let _content_records = ContentSectionRecords::try_from( &message ).unwrap();
+        //println!("\ncontent_records = {:#?}", content_records);
```
phillipleblanc:
```suggestion
        let _content_records = ContentSectionRecords::try_from(&message).expect("Failed to parse message");
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+    let raw_email = include_str!("test_data/three_attachments.txt").to_string(); // Load an email file
+
+    println!("\nthree attachments");
+
+    let eml_msg = MessageParser::default().parse(&raw_email);
+
+    if let Some(message) = eml_msg {
+        //println!("message html_body_count: {}", message.html_body_count());
+        let _content_records = ContentSectionRecords::try_from( &message ).unwrap();
+        //println!("\ncontent_records = {:#?}", content_records);
```
phillipleblanc:
```suggestion
        let _content_records = ContentSectionRecords::try_from( &message ).expect("Failed to parse message");
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+        let attachment_records = AttachmentRecords::try_from(&message).unwrap();
+
+        let extracted_set = attachment_records.attachments;
+        for rec in &extracted_set {
+            let blob = rec.blob.clone().unwrap();
+
+            let blob_str = if let Ok(blob) = String::from_utf8(blob) {
+                blob
+            } else {
+                format!("[{}] Could not decode into utf8 for display", rec.mime_type.clone().unwrap())
```
phillipleblanc:
```suggestion
                format!("[{}] Could not decode into utf8 for display", rec.mime_type.clone().expect("mime type not set"))
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+            };
+            println!("-rec: {:?}", &blob_str);
+            println!("");
+        }
+
+
+        //let image_bytes = include_bytes!("test_data/Mark - Orange on White.png");
```
phillipleblanc:
```suggestion
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+    let raw_email = include_str!("test_data/malformed_headers.txt").to_string(); // Load an email file
+
+    println!("\nmalformed headers");
+
+    let eml_msg = MessageParser::default().parse(&raw_email);
+
+    if let Some(message) = eml_msg {
+        //println!("message html_body_count: {}", message.html_body_count());
+        let _content_records = ContentSectionRecords::try_from( &message ).unwrap();
+        //println!("\ncontent_records = {:#?}", content_records);
```
phillipleblanc:
```suggestion
        let _content_records = ContentSectionRecords::try_from(&message).expect("Failed to parse message");
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+
+
+        let attachment_records = AttachmentRecords::try_from( &message ).unwrap();
```
phillipleblanc:
```suggestion
        let attachment_records = AttachmentRecords::try_from( &message ).expect("Failed to parse attachments");
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    let eml_msg = MessageParser::default().parse(&raw_email);
+
+    if let Some(message) = eml_msg {
+        //println!("message html_body_count: {}", message.html_body_count());
+        let _content_records = ContentSectionRecords::try_from( &message ).unwrap();
+        //println!("\ncontent_records = {:#?}", content_records);
+
+
+        let attachment_records = AttachmentRecords::try_from(&message).unwrap();
```
phillipleblanc:
```suggestion
        let attachment_records = AttachmentRecords::try_from(&message).expect("Failed to parse attachments");
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        let extracted_set = attachment_records.attachments;
+        for rec in &extracted_set {
+            let blob = rec.blob.clone().unwrap();
```
phillipleblanc:
```suggestion
            let blob = rec.blob.clone().expect("blob not set");
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    let eml_msg = MessageParser::default().parse(&raw_email);
+
+    if let Some(message) = eml_msg {
+        //println!("message html_body_count: {}", message.html_body_count());
+        let _content_records = ContentSectionRecords::try_from( &message ).unwrap();
+        //println!("\ncontent_records = {:#?}", content_records);
+
+
+        let attachment_records = AttachmentRecords::try_from(&message).unwrap();
```
phillipleblanc:
```suggestion
        let attachment_records = AttachmentRecords::try_from(&message).expect("Failed to parse attachments");
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        let extracted_set = attachment_records.attachments;
+        // for rec in &collated{
+        //     println!("-rec: {:?}", rec);
+        // }
```
phillipleblanc:
```suggestion
```

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+            },
+            AttachmentInfo {
+                filename: None,
+                mime_type: Some("application/pdf".to_string()),
+                blob: None
+            },
+            // AttachmentInfo{
+            //     filename: None,
+            //     mime_type: None,
+            // },
```
phillipleblanc:
```suggestion
```

crates/data_components/src/imap/mod.rs
benrussell: precedent:

https://github.com/spiceai/spiceai/blob/trunk/crates/db_connection_pool/src/dbconnection/snowflakeconn.rs#L708

https://github.com/spiceai/spiceai/blob/trunk/crates/db_connection_pool/src/dbconnection/snowflakeconn.rs#L739


I went looking for existing Spice OSS code to create a suitable Arrow Structure record type when refac to compound attachment struct records.

Found the snowflakeconn.rs example.

Would not normally submit with loose code paths.

fn's with .expect in PR have been FIXME tagged for further review since creation as I know they're a pain point.

crates/data_components/src/imap/provider.rs
```diff
@@ -152,26 +198,32 @@ impl TableProvider for ImapTableProvider {
...
+
             let body = if self.fetch_content {
-                message.body().as_ref().map(|v| decode(v))
+                message.body().as_ref().map(|v| decode_string(v))
             } else {
                 None
             };
 
+
+            let mime_extract = MimeExtract::try_from(&body).ok().unwrap();
```
phillipleblanc: This will need to return a DataFusion error instead of unwrapping.

crates/data_components/src/imap/mod.rs
```diff
@@ -74,6 +82,105 @@ fn build_listarray_for_strings(values: Vec<Option<Vec<Option<String>>>>) -> List
...
 
+
+//FIXME: The following fn were written with help from GPT.
+// They seem plausible but I am not versed in the construction of Arrow suitable structures.
+// I had a look at snowflakeconn.rs - seems similar.
+// - build_listarray_for_content_sections(..)
+// - build_listarray_for_attachments(..)
+
+
+
```
phillipleblanc:
```suggestion
```

Thanks for the note!

crates/data_components/src/imap/mod.rs
```diff
@@ -74,6 +82,105 @@ fn build_listarray_for_strings(values: Vec<Option<Vec<Option<String>>>>) -> List
...
+fn build_listarray_for_content_sections(values: Vec<Option<Vec<ContentSectionInfo>>>) -> ListArray {
```
phillipleblanc: This function should return a Result<ListArray> and instead of panicking below, return an error that is handled higher up.

phillipleblanc: Asking GPT to create some tests for this code would be useful as well.

crates/data_components/src/imap/mime.rs
```diff
@@ -0,0 +1,244 @@
...
+
+#[derive(Debug)]
+#[derive(PartialEq)]
+pub struct AttachmentRecords{
+    pub attachments: Vec<AttachmentInfo>,
+}
+
+
+impl TryFrom<&Message<'_>> for AttachmentRecords{
+    type Error = String;
```
phillipleblanc: We should define a new Snafu Error enum and use that here.

crates/data_components/src/imap/mime.rs
```diff
@@ -0,0 +1,244 @@
...
+
+#[derive(Debug)]
+#[derive(PartialEq)]
+pub struct ContentSectionRecords {
+    pub sections: Vec<ContentSectionInfo>,
+}
+
+
+impl TryFrom<&Message<'_>> for ContentSectionRecords {
+    type Error = String;
```
phillipleblanc: We should define a new Snafu Error enum and use that here.

crates/data_components/src/imap/tests.rs
```diff
@@ -0,0 +1,221 @@
...
+#[cfg(test)]
+use mail_parser::MessageParser;
+use super::mime::{AttachmentInfo, AttachmentRecords,ContentSectionRecords};
+
+
+
+
+
+#[test]
+fn eml_utf8() {
```
phillipleblanc: I would move these tests to the bottom of the mime.rs file instead of having them in their own file.

crates/data_components/src/imap/mod.rs
phillipleblanc: As part of our style guide, we don't allow the usage of `.expect` or `.unwrap` in production code - we allow `.expect` only in tests. Those two examples you found are helper methods in a test module. A more idiomatic example you can reference would be: https://github.com/spiceai/spiceai/blob/trunk/crates/db_connection_pool/src/dbconnection/snowflakeconn.rs#L257

crates/data_components/src/imap/tests.rs
benrussell: Updated this to reflect the new default mime_type assignment of "application/octet-stream".

This triggered a fail on my cargo test but I didn't see it before I'd pushed and requested review.

crates/data_components/src/imap/provider.rs
benrussell: MimeExtract can always return a valid struct. 
Worst case is None vals for member vars. 
Removes need for DataFusion error wrap.

Have refactored to use MimeExtract::from(...)

https://github.com/spiceai/spiceai/pull/5083/commits/a44621a75b7d3a18bca32f305323e59841845841

crates/data_components/src/imap/mime.rs
benrussell: Done.

https://github.com/spiceai/spiceai/pull/5083/commits/e15738f8dfd374adda1e45681206afbeaf48bada

crates/data_components/src/imap/mime.rs
benrussell: Done.

https://github.com/spiceai/spiceai/pull/5083/commits/e15738f8dfd374adda1e45681206afbeaf48bada

crates/data_components/src/imap/mod.rs
benrussell: Fixed.

crates/data_components/src/imap/mod.rs
benrussell: return type, data, and handling has been upgraded.

crates/data_components/src/imap/tests.rs
benrussell: merged tests.rs into mime.rs

prefixed test function signatures with "test_"

crates/data_components/src/imap/mod.rs
benrussell: Created tests with o3-mini-high - found and fixed a bug with the Arrow struct def code.

benrussell:
Clippy passes.

crates/data_components/src/imap/mod.rs
```diff
@@ -57,6 +66,19 @@ pub enum Error {
     InvalidSSLMode { ssl_mode: String },
     #[snafu(display("An invalid authentication mode was provided: {auth_mode}"))]
     InvalidAuthMode { auth_mode: String },
+
+    //FIXME: Not happy with error descriptions.
```
benrussell, line 70: Are these error descriptions adequate / correct?

crates/data_components/src/imap/mod.rs
```diff
@@ -74,6 +96,113 @@ fn build_listarray_for_strings(values: Vec<Option<Vec<Option<String>>>>) -> List
     builder.finish()
 }
 
+//FIXME: I don't really like the overload/shadow of Result here but GPT told me it was an accepted Rust idiom.
```
benrussell, line 99: Please clarify which Error idiom is prefered in the Spice codebase so I can make it consistent. (1 of 2)

crates/data_components/src/imap/mime.rs
```diff
@@ -0,0 +1,413 @@
...
+                })
+            })
+            .unwrap_or(MimeExtract {
+                attachments: None,
+                content_sections: None,
+            })
+    }
+}
+
+//FIXME: Apparently overload/shadow of "Error" is normal Rust idiomatic code.
```
benrussell, line 68: Please clarify which Error idiom is prefered in the Spice codebase so I can make it consistent. (2 of 2)

crates/data_components/src/imap/mime.rs
```diff
@@ -0,0 +1,412 @@
...
```
benrussell, line 68: Please clarify which Error idiom is prefered in the Spice codebase so I can make it consistent. (2 of 2)

crates/data_components/src/imap/mod.rs
```diff
@@ -57,6 +66,19 @@ pub enum Error {
...
+        field_number: i32,
+        field_name: String,
+    },
+
+    #[snafu(display("Should return a field builder: {field_number}/{field_name}"))]
+    FailedToBuildListArrayForContentSections {
+        field_number: i32,
+        field_name: String,
+    },
 }
```
benrussell, line 82: Are these error descriptions adequate / correct?

crates/data_components/src/imap/mod.rs
```diff
@@ -74,6 +96,113 @@ fn build_listarray_for_strings(values: Vec<Option<Vec<Option<String>>>>) -> List
     builder.finish()
 }
 
+//FIXME: I don't really like the overload/shadow of Result here but GPT told me it was an accepted Rust idiom.
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
```
benrussell, line 101: Please clarify which Error idiom is prefered in the Spice codebase so I can make it consistent. (1 of 2)

crates/data_components/src/imap/mime.rs
```diff
@@ -0,0 +1,412 @@
...
+}
+
+#[derive(Debug, PartialEq)]
+pub struct AttachmentRecords {
+    pub attachments: Vec<AttachmentInfo>,
+}
+
+impl TryFrom<&Message<'_>> for AttachmentRecords {
+    type Error = MimeExtractError;
+
```
benrussell, line 97: Please clarify which Error idiom is prefered in the Spice codebase so I can make it consistent. (2 of 2)

Issue #5084, opened 2025-03-17
Enhancement: Remove redundant refreshes on initial acceleration load for accelerations with persistent data


sgrebnov:
## Goal-State/What/Result

There are no redundant refreshes on initial acceleration load for accelerations with persistent data.

Ability to always refresh on start is maintained:

```yaml
acceleration:
  refresh_on_startup: always | auto (default)
```

- **Scenario 1**: Acceleration with persistence (DuckDB), and default refresh configuration.
 - A: Previously refreshed -> No refresh.
 - B: Not previously refreshed (but configured to use persistence) -> Refresh on startup.

- **Scenario 2**: Acceleration with persistence (DuckDB), refresh interval configured.
  - A: On load, interval already passed -> Refresh on startup.
  - B: On load, interval pending -> No refresh, but next refresh schedule is maintained based on previous checkpoint.

## Why/Purpose

Redundant refreshes impose unnecessary load on data sources, increasing network traffic, latency, and resource usage.

## By When

**Issue/Spec written and reviewed:**  03/17
**Done-Done:** 03/19

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

1. Spicepod acceleration supports `refresh_on_startup` configuration parameter (YAML)

```yaml
acceleration:
  refresh_on_startup: always | auto (default)
```

1. Spice implements the following refresh behavior when application starts:

| Scenario | refresh_on_startup | Refresh Interval Configured | Condition                             | Expected Outcome                                             |
|----------|--------------------|-----------------------------|---------------------------------------|--------------------------------------------------------------|
| 1A       | auto               | No                          | Previously refreshed                  | No refresh on startup                                        |
| 1A       | always             | No                          | Previously refreshed                  | Refresh on startup                                           |
| 1B       | auto               | No                          | Not previously refreshed              | Refresh on startup                                           |
[...]

PR #5086, opened 2025-03-17
Add `refresh_on_startup` Spicepod configuration param

sgrebnov:
# 🗣 Description

Introduce `refresh_on_startup ` Spicepod configuration param to control accelerated data refresh during application start: 

```yaml
acceleration:
  refresh_on_startup: always | auto (default)
```

I've also created a new trait for the DatasetCheckpoint called `DatasetCheckpointer` and changed all references to use that:

```rust
#[async_trait]
pub trait DatasetCheckpointer: Send + Sync {
    async fn exists(&self) -> bool;
    async fn checkpoint(&self, schema: &SchemaRef) -> Result<()>;
    async fn get_schema(&self) -> Result<Option<SchemaRef>>;
}
```

This makes it easy to test the refresh logic is correct from a unit test.

The `should_refresh` method can be extended in the future to take into account the time of the last checkpoint and the refresh check interval to see if we should initiate a new refresh immediately or not.

## 🔨 Related Issues

Part of 
- https://github.com/spiceai/spiceai/issues/5084

## ⛓️‍💥 Breaking Change

This is technically a breaking change in behavior for file mode accelerated datasets - however it should not affect functionality as the extra refreshes this will filter out were unnecessary to begin with.

## 📄 Documentation Updates

New docs for the `refresh_on_startup` parameter.


phillipleblanc (dismissed):

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -234,6 +234,31 @@ impl Refresh {
...
+            RefreshMode::Append | RefreshMode::Changes => return true,
+            RefreshMode::Disabled => return false,
+        };
+
+        // If there is no previous checkpoint, we need to start a refresh.
+        if !has_previous_checkpoint {
+            return true;
+        }
+
+        // If we have a previous checkpoint and the refresh interval is set, then we need to refresh - otherwise we don't.
```
lukekim: This is a misleading comment. We should note this is temporary behavior, in the future, we need to check the interval time against the current time to determine if we need to refresh.

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -428,12 +463,28 @@ impl Refresher {
...
+        {
+            let refresh = self.refresh.read().await;
+
+            // If the table already has an existing acceleration and the refresh options wouldn't start a new refresh,
+            // we can exit early.
+            if self.refresh_on_startup == RefreshOnStartup::Auto
+                && !refresh.should_refresh(self.checkpointer.clone()).await
+            {
+                tracing::debug!(
+                    "Refresh not started for {} - an existing acceleration is available",
```
lukekim:
```suggestion
                    "Skipped refresh for {}: existing acceleration is available",
```

lukekim (dismissed):

phillipleblanc (approved):

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -234,6 +234,33 @@ impl Refresh {
...
+        last_checkpoint: Option<Arc<dyn DatasetCheckpointer>>,
+    ) -> bool {
+        // If there is no checkpoint, we need to start a refresh.
+        let Some(last_checkpoint) = last_checkpoint else {
+            return true;
+        };
+
+        let has_previous_checkpoint = match self.mode {
+            RefreshMode::Full => last_checkpoint.exists().await,
+            // Append and Changes modes are always refreshed since they stream changes from the source table.
```
Sevenannn, line 249:
```suggestion
            // Append and Changes modes should always refresh since they stream changes from the source table.
```

Sevenannn (approved):

PR #5090, opened 2025-03-18
fix: Acceleration federation test

peasee:
# 🗣 Description

* My last fix (#5070) wasn't quite right. This time, we check the query result to determine if the query error is a result of acceleration not being ready - this test is for federation of the acceleration; not of the connecting source! That makes `ready_state` inapplicable for this test.

### DIFF
--- a/crates/runtime/tests/acceleration/query_push_down.rs
+++ b/crates/runtime/tests/acceleration/query_push_down.rs
@@ -14,22 +14,17 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use std::time::Duration;
-
 use app::AppBuilder;
 use datafusion::assert_batches_eq;
 use futures::StreamExt;
 use futures::TryStreamExt;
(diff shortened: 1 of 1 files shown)


phillipleblanc (dismissed):

crates/runtime/tests/acceleration/query_push_down.rs
```diff
@@ -154,13 +152,32 @@ async fn acceleration_with_and_without_federation() -> Result<(), anyhow::Error>
 
             assert!(
                 wait_until_true(Duration::from_secs(30), || async {
-                    let mut query_result = rt
+                    let mut query_result = match rt
```
sgrebnov: Can we just use `wait_untill_ready` here ?

And then
1. Test query
2. Test query plan

Sevenannn:
Will `runtime_ready_check` help in this case, running before the query against accelerated table and ensure the tables are already accelerated at the time of querying

sgrebnov (approved):

PR #5096, opened 2025-03-18
fix: Make timings more robust for vector search perf

peasee:
# 🗣 Description

* Makes the test for vector search request parameter parsing performance more robust:
    * Increase test iteration count from 3 to 5
    * Add a small wait between parsing requests, to limit resource contention
    * Increased time threshold from 1ms to 1.25ms


crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -1478,7 +1482,7 @@ pub(crate) mod tests {
         let average_time = total_time / (timings.len() as u32);
         let average_time_ns = average_time.as_nanos();
         assert!(
-            average_time_ns < 1_000_000,
+            average_time_ns < 1_250_000,
```
phillipleblanc: I think experience has shown that these types of tests are not reliable enough - is there another way we can test this?

lukekim (dismissed):

crates/runtime/src/embeddings/vector_search.rs
```diff
@@ -1411,29 +1414,25 @@ pub(crate) mod tests {
         assert!(result.is_err());
     }
 
-    #[test]
-    fn test_performance_of_column_parsing() {
+    #[tokio::test]
+    async fn test_performance_of_column_parsing() {
```
phillipleblanc: This is still not reliable enough, it failed in a check on this PR: https://github.com/spiceai/spiceai/actions/runs/14002451832/job/39211332410

peasee: I've made an update to use the statistical set of measurements to remove measurement outliers (outside the IQR), and also added a modifier based on system load for the target timing measurement.

phillipleblanc (dismissed):

peasee:
This still isn't robust enough

Issue #5059, opened 2025-03-17
Fix file snapshotting path comparisons in explain plan


peasee:
https://github.com/spiceai/spiceai/actions/runs/13889723940/job/38859537642

Example snapshot change:

```console
    13       │-|               |           ParquetExec: file_groups={24 groups: [[tmp/.tmpJ1DebA/data/orders.parquet:0..2311466], [tmp/.tmpJ1DebA/data/orders.parquet:2311466..4622932], [tmp/.tmpJ1DebA/data/orders.parquet:4622932..6934398], [tmp/.tmpJ1DebA/data/orders.parquet:6934398..9245864], [tmp/.tmpJ1DebA/data/orders.parquet:9245864..11557330], ...]}, projection=[o_orderkey], limit=10 |
         13 │+|               |           ParquetExec: file_groups={24 groups: [[tmp/.tmpJGzOJs/data/orders.parquet:0..2311466], [tmp/.tmpJGzOJs/data/orders.parquet:2311466..4622932], [tmp/.tmpJGzOJs/data/orders.parquet:4622932..6934398], [tmp/.tmpJGzOJs/data/orders.parquet:6934398..9245864], [tmp/.tmpJGzOJs/data/orders.parquet:9245864..11557330], ...]}, projection=[o_orderkey], limit=10 |
```

Replace the temporary data directory with nothing in the explain plan text? `tmp/.tmp/JGzOJs/data/` -> `data/` which would now be a comparable path.

PR #5103, opened 2025-03-18
Use fixed data directory for test operator

Sevenannn:
# 🗣 Description

Utilize the insta filter to get reliable snapshots for file data connectors, the temporary file paths will be converted to a consistent /data/some_file.parquet

file path is persistent after this change: https://github.com/spiceai/spiceai/actions/runs/13921456269/job/38955484474

## 🔨 Related Issues

- https://github.com/spiceai/spiceai/issues/5059

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

### DIFF
--- a/crates/test-framework/src/spiced/mod.rs
+++ b/crates/test-framework/src/spiced/mod.rs
@@ -24,29 +24,43 @@ use anyhow::Result;
 use flight_client::{Credentials, FlightClient};
 use spicepod::spec::SpicepodDefinition;
 use sysinfo::{Pid, ProcessesToUpdate, System};
-use tempfile::TempDir;
 
 use crate::utils::wait_until_true;
 
 pub struct SpicedInstance {
     child: Child,
-    tempdir: TempDir,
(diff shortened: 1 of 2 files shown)


crates/test-framework/src/spiced/mod.rs
```diff
@@ -24,29 +24,43 @@ use anyhow::Result;
...
     data_dir: Option<PathBuf>,
     prepared: bool,
 }
 
 impl StartRequest {
     pub fn new(spiced_path: PathBuf, spicepod: SpicepodDefinition) -> Result<Self> {
+        let tempdir = {
+            #[cfg(not(target_os = "windows"))]
+            {
+                let path = PathBuf::from("/tmp/spice_test_operator");
```
phillipleblanc: Instead of hard coding a path, let's try to use filters: https://insta.rs/docs/filters/

peasee (commented):
Can we add some unit tests to test that regex?

phillipleblanc (approved):

PR #5116, opened 2025-03-19
Properly handle '/' for tool names.

Jeadie:
# 🗣 Description
 - Openai standard needs tools to have names comply with `^[a-zA-Z0-9_-]+$`.
 - Especially in MCP, many tools have `/`.
 - We use an encoding scheme if `/` are present. 
   - `foo/bar_baz` -> `foo_bar__baz`.

### DIFF
--- a/crates/runtime/src/model/tool_use.rs
+++ b/crates/runtime/src/model/tool_use.rs
@@ -77,7 +77,7 @@ impl ToolUsingChat {
                 r#type: ChatCompletionToolType::Function,
                 function: FunctionObject {
                     strict: t.strict(),
-                    name: t.name().to_string(),
+                    name: encode_tool_name(t.name().to_string().as_str()),
                     description: t.description().map(|d| d.to_string()),
                     parameters: t.parameters(),
                 },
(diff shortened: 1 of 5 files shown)


ewgenius (approved):

crates/runtime/src/model/tool_use.rs
```diff
@@ -695,3 +692,12 @@ fn make_a_stream(
     );
     Box::pin(CustomStream { receiver }) as ChatCompletionResponseStream
 }
+
+// OpenAI tools must satisfy '^[a-zA-Z0-9_-]+$'. Commonly external tools may have '/' in their name.
+fn encode_tool_name(name: &str) -> String {
+    if name.contains('/') {
+        name.replace('_', "__").replace('/', "_")
```
phillipleblanc, line 699: This feels extremely weird to me - why can't we support tools with `/`?

Jeadie: An inherent limitation in Openai and some other providers. Tool names must satisfy `^[a-zA-Z0-9_-]+$`.

PR #5120, opened 2025-03-20
Use retry logic when loading tools

Jeadie:
# 🗣 Description

Retry logic for tools loading

### DIFF
--- a/crates/runtime/src/init/tool.rs
+++ b/crates/runtime/src/init/tool.rs
@@ -14,17 +14,19 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, sync::Arc, time::Duration};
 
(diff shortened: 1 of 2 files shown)


phillipleblanc (dismissed):

crates/runtime/src/init/tool.rs
```diff
@@ -37,16 +39,25 @@ impl Runtime {
             }
         }
 
+        let mut spawned_tasks = vec![];
+        let cloned_self = self.clone();
```
Sevenannn, line 43: Would the cloning of `Runtime` be expensive?

ewgenius (approved):

crates/runtime/src/init/tool.rs
Jeadie: Not really. Runtime is essentially a collection of `Arc<T>` so this clone is essentially just n `Arc<T>` clones.

Issue #5113, opened 2025-03-19
Enhancement: Model graded Eval scorers.


Jeadie:
## Goal-State/What/Result
Spice supports running LLMs, evaluating LLMs, and provides several functions to the score the results of an eval run. Currently, users can not use an LLM to judge/score another LLMs eval ouput.

This enhancement is to add LLM-as-a-judge scoring natively, through spicepod model components.

## Why/Purpose
LLM scoring, or LLM-as-a-judge scorers (also called judge models in this enhancement) are a powerful way to measure certain characteristics of a LLM response that may be hard to codify in strict logic. They are also an effective way to capture the long tail of characteristics users want to measure.

## By When

**Issue/Spec written and reviewed:** 20th March
**Done-Done:** _Target Date_

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

### Requirements
 - Judge models will be defined as a model spicepod component.
 - There must be a mechanism to provide data from the eval dataset (minimally input, output, actual; but further to arbitrary columns in the dataset) into the model. Tangibly, this could be:
   - Templating of the system prompt. There is the open question for how this data would be provided via `chat/completions`. Current prototypes use the `metadata` field of the request payload.
      ```yaml
      models:
        - name: openai-with-spice
          from: openai:gpt-4o-2024-08-06
          params:
            parameterized_prompt: enabled # disabled
            system_prompt: |
              # Task
              Given an input prompt, you must evaluate the provided haiku and how well it describes {{ student.theme }}
              <poet>{{ input }}</poet>
              <research_artifacts>{{ actual }}</research_artifacts>
      ```
 - Need mechanism in `chat/completion` to provide data to model.
 - Need UX to know when to template `system_prompt` with data, vs current implementation.

### Implementation
 - Use `metadata` key in request payload for templating data. Any metadata provided can be used by templating.
 - Use `params.parameterized_prompt: enabled # disabled` to determine if templating should be attempted.
   - Security posture: No templating by default. 
 - Can check at startup if LLM has template variables in `system_prompt`. If so, when request metadata has these fields, template.
[...]

PR #5123, opened 2025-03-21
Enable 'parameterized_prompt' in LLMs

Jeadie:
# 🗣 Description
Enable LLM system prompt to be templated, in Jinja syntax, per-request (i.e. at `chat/completion` request time).

## Example
Spicepod
```yaml
models:
  - name: openai-with-spice
    from: openai:gpt-4o-2024-08-06
    params:
      parameterized_prompt: enabled
      spice_tools: auto
      system_prompt: |
        Foo: {{ bar }}
        IS: {{ another_one }}
```

Variables for templating are provided via `.metadata`.
```shell
curl -XPOST http://localhost:8090/v1/chat/completions \
  -d '{
    "model": "openai-with-spice",
    "messages": [
        {
          "role": "user",
          "content": "Hello?"
        }
    ],
    "metadata": {
      "bar": "BBBBBAAAAAAAA",
      "another_one": " we Americans have distanced ourselves from the making of our everyday surrounding"
    }
}'
```

These two results show 1. `ai_chat` task with metadata in input , then 2. `ai_completion` task with no metadata (since `.store` field is not set `true`), but the system prompt has been templated.
```shell
>> curlie -X POST "http://localhost:8090/v1/sql" \
  --data "select input from runtime.task_history where trace_id='b5a1459676a204859fd9597a49b03ffd';" \
  | jq -r '.[].input' | jq .
```
```json
{
  "messages": [
    {
      "role": "user",
      "content": "Hello?"
    }
  ],
  "model": "openai-with-spice",
  "metadata": {
    "bar": "BBBBBAAAAAAAA",
    "another_one": " we Americans have distanced ourselves from the making of our everyday surrounding"
  }
}
{
  "messages": [
    {
      "role": "system",
      "content": "Foo: BBBBBAAAAAAAA\nIS:  we Americans have distanced ourselves from the making of our everyday surrounding\n"
    },
    {
      "role": "user",
      "content": "Hello?"
    }
  ],
  "model": "openai-with-spice"
}
```

## 🔨 Related Issues
#5113 

## 📄 Documentation Updates
- [ ] https://github.com/spiceai/docs/pull/907


crates/runtime/Cargo.toml
```diff
@@ -121,6 +121,7 @@ ssh2 = { workspace = true, optional = true }
 strsim = { version = "0.11.1", optional = true }
 suppaftp = { workspace = true, optional = true }
 telemetry = { path = "../telemetry" }
+tera = "1.17"
```
phillipleblanc: This can be an optional dependency enabled by `models`

crates/runtime/src/model/wrapper.rs
```diff
@@ -374,3 +431,127 @@ impl<S> Drop for TracedChatCompletionStream<S> {
...
+    let mut t = Tera::default();
+    if t.add_raw_template("system_prompt", prompt).is_err() {
+        return false;
+    };
+
+    let Ok(tt) = t.get_template("system_prompt") else {
+        return false;
+    };
+    let result = has_variables_in_ast(&tt.ast);
+    tracing::debug!("Early check if system prompt='{prompt:?}' is template has result='{result}'. Not gating prompts.");
```
phillipleblanc: So we're not actually returning if the system prompt has variables? The function name is misleading then.

crates/runtime/src/model/wrapper.rs
```diff
@@ -374,3 +431,127 @@ impl<S> Drop for TracedChatCompletionStream<S> {
...
+    true
+}
+
+fn has_variables_in_ast(ast: &[Node]) -> bool {
```
phillipleblanc: These ast walking functions can live in their own sub-module.

phillipleblanc (dismissed):

crates/runtime/src/model/wrapper/mod.rs
```diff
@@ -374,3 +431,31 @@ impl<S> Drop for TracedChatCompletionStream<S> {
...
+    if t.add_raw_template("system_prompt", prompt).is_err() {
+        return false;
+    };
+
+    let Ok(tt) = t.get_template("system_prompt") else {
+        return false;
+    };
+    let result = ast::has_variables_in_ast(&tt.ast);
+    tracing::debug!("Early check if system prompt='{prompt:?}' is template has result='{result}'. Not gating prompts.");
+    true
```
Sevenannn: should result be returned here instead of true?

phillipleblanc: Yeah I'm not sure what Jack had in mind there.

phillipleblanc (approved):

PR #5124, opened 2025-03-21
Improve performance of schema inference for object-store data connectors

sgrebnov:
# 🗣 Description

Improve the performance of schema inference by using only the last modified file instead of all files. The last modified file is used (rather than the first) to correctly handle potential schema evolution, as the most recent file is more likely to contain the most up-to-date schema.

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/runtime/src/dataconnector.rs
+++ b/crates/runtime/src/dataconnector.rs
@@ -235,7 +235,7 @@ pub enum DataConnectorError {
     },
 
     #[snafu(display(
-        "Failed to load the {connector_component} ({dataconnector}).\nFailed to detect a table schema.\nReport a bug on GitHub (https://github.com/spiceai/spiceai/issues) and reference the error: {source}"
+        "Failed to load the {connector_component} ({dataconnector}).\nFailed to infer the table schema.\nReport a bug on GitHub (https://github.com/spiceai/spiceai/issues) and reference the error: {source}"
     ))]
     UnableToGetSchemaInternal {
         dataconnector: String,
--- a/crates/runtime/src/dataconnector.rs
+++ b/crates/runtime/src/dataconnector.rs
@@ -235,6 +235,15 @@ pub enum DataConnectorError {
     },
 
     #[snafu(display(
+        "Failed to load the {connector_component} ({dataconnector}).\nFailed to detect a table schema.\nReport a bug on GitHub (https://github.com/spiceai/spiceai/issues) and reference the error: {source}"
+    ))]
+    UnableToGetSchemaInternal {
+        dataconnector: String,
+        connector_component: ConnectorComponent,
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
+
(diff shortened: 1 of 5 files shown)


crates/runtime/src/dataconnector/listing/connector.rs
```diff
@@ -489,6 +498,23 @@ impl<T: ListingTableConnector + Display> DataConnector for T {
     }
 }
 
+fn to_listing_table_url(
+    original_url: &Url,
+    path: &Path,
+    dataset: &Dataset,
+    dataconnector: &str,
+) -> DataConnectorResult<ListingTableUrl> {
```
sgrebnov: TODO: Is there a better/right way to construct ListingTableUrl?

crates/runtime/src/dataconnector.rs
```diff
@@ -234,6 +234,15 @@ pub enum DataConnectorError {
         source: Box<dyn std::error::Error + Send + Sync>,
     },
 
+    #[snafu(display(
+        "Failed to load the {connector_component} ({dataconnector}).\nFailed to detect a table schema.\nReport a bug on GitHub (https://github.com/spiceai/spiceai/issues) and reference the error: {source}"
```
lukekim:
```suggestion
        "Failed to load the {connector_component} ({dataconnector}).\nFailed to infer the table schema.\nReport a bug on GitHub (https://github.com/spiceai/spiceai/issues) and reference the error: {source}"
```

crates/runtime/src/dataconnector/listing/connector.rs
```diff
@@ -331,6 +333,148 @@ pub trait ListingTableConnector: DataConnector {
...
+        extension: &str,
+    ) -> DataConnectorResult<Arc<dyn TableProvider>>
+    where
+        Self: Display,
+    {
+        let content_formatter =
+            document_parse::get_parser_factory(extension)
+                .await
+                .map(|factory| {
+                    // TODO: add opts.
```
sgrebnov, line 350: This is original version TODO

crates/runtime/src/dataconnector/listing/connector.rs
```diff
@@ -331,6 +333,148 @@ pub trait ListingTableConnector: DataConnector {
...
+            },
+        )?;
+
+        let object_store = self.get_object_store(dataset)?;
+
+        let ctx: SessionContext = Self::get_session_context();
+
+        // Get the last modified object for the provided ObjectStore to infer the schema.
+        // Report an error if no files matching required extension are found.
+        let last_modified_or_added = get_last_modified(
```
sgrebnov, line 395: We can use `table_path.is_collection()` to detect whether the path is a folder or a file, and avoid the logic for getting the last modified file. I'm keeping this logic as it does not add overhead (just returns the same path/file) and makes the code more readable by reducing if/else branching.

phillipleblanc (approved):

PR #5125, opened 2025-03-21
Use embedding models and LLMs are eval scorers. 

Jeadie:
# 🗣 Description
 - Two new types of eval scorers:
   - Use embedding model to compute embedded distance as a similarity score.
   - Use LLM as a judge, where the LLM is expected to return JSON with at least a `.score` field.
 - At `spiced` initialisation, for each LLM and embedding model, load as this associated scorer. 
 - In this PR, needed to change trait signature for scoring
    ```diff
    #[async_trait]
    pub trait Scorer: Sync + Send {
        async fn score(
            &self,
            input: &DatasetInput,
            actual: &DatasetOutput,
            ideal: &DatasetOutput,
    -      ) -> f32;
    +      ) -> Result<f32>;
    ```
## 🔨 Related Issues
#5113 
 - Need #5123 for LLM graded scorer to be effective.

## 📄 Documentation Updates
 - [ ] https://github.com/spiceai/docs/pull/908


crates/runtime/src/model/eval/scorer/embed.rs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        let distance = actual_vector
+            .iter()
+            .zip(ideal_vector.iter())
+            .map(|(a, b)| (a - b).powi(2))
+            .sum::<f32>()
+            .sqrt();
+
+        Ok(1.0 / (1.0 + distance))
```
Sevenannn, line 101: Do we have plans to include other embedding comparison metrics (e.g. consine similarity) as the score, l2 distance generally increases with vector dimensionality and the magnitude difference of embeddings may vary if the embedding model don't do normalization.

Jeadie: That is a good point. we should actually use cosine similarity by default. All that will need to change for this is to normalise by both vector's norms.

crates/runtime/src/model/eval/dataset.rs
```diff
@@ -58,6 +58,24 @@ pub enum DatasetOutput {
     AssistantResponse(String),
 }
 
+impl std::fmt::Display for DatasetInput {
```
phillipleblanc: nit: should this be debug instead? Display using a Debug internally is a smell.

crates/runtime/src/model/eval/scorer/embed.rs
```diff
@@ -0,0 +1,121 @@
...
+pub struct EmbedScorer {
+    models: Arc<RwLock<EmbeddingModelStore>>,
+    model_name: String,
+
+    /// Provide domain context for improved embedding comparison.
+    prefix: Option<String>,
+}
+
+impl EmbedScorer {
+    pub fn new(models: Arc<RwLock<EmbeddingModelStore>>, model_name: String) -> Self {
```
phillipleblanc: Why pass in a model store and a model name - why not just pass the model itself?

crates/runtime/src/model/eval/scorer/embed.rs
```diff
@@ -0,0 +1,121 @@
...
+#[async_trait]
+impl Scorer for EmbedScorer {
+    async fn score(
+        &self,
+        input: &DatasetInput,
+        actual: &DatasetOutput,
+        ideal: &DatasetOutput,
+    ) -> super::Result<f32> {
+        let prefix = self.prefix.as_deref().unwrap_or_default();
+        let actual_input = format!("{prefix}{}", extract_text(actual),);
```
phillipleblanc:
```suggestion
        let actual_input = format!("{prefix}{}", extract_text(actual));
```

crates/runtime/src/model/eval/scorer/embed.rs
```diff
@@ -0,0 +1,121 @@
...
+            .zip(ideal_vector.iter())
+            .map(|(a, b)| (a - b).powi(2))
+            .sum::<f32>()
+            .sqrt();
+
+        Ok(1.0 / (1.0 + distance))
+    }
+
+    fn metrics(&self, scores: &[f32]) -> Vec<(String, f32)> {
+        vec![("mean".to_string(), mean(scores))]
```
phillipleblanc, line 105: What is this `metrics` used for? Also `mean` as a metric is generally not useful.

crates/runtime/src/model/eval/scorer/mod.rs
```diff
@@ -17,10 +17,15 @@ limitations under the License.
 use std::{collections::HashMap, sync::Arc};
 
 use async_trait::async_trait;
+use snafu::Snafu;
 use tokio::sync::RwLock;
 
 use super::{DatasetInput, DatasetOutput};
 
+#[cfg(feature = "models")]
```
phillipleblanc, line 25: Should this top-level module be feature flagged? Feels weird to have so many sub-modules feature flagged.

crates/runtime/src/model/eval/scorer/modelgraded.rs
```diff
@@ -0,0 +1,156 @@
...
+/// [`ModelGradedScorer`] scores based on the score provided in response by the language model. The model response is expected to be JSON, with at least a `score` field of float or integer type. Additional fields are allowed, but ignored.
+///
+/// The [`DatasetInput`] and both [`DatasetOutput`]s are provided to the [`Chat`] model via request metadata (i.e. [`CreateChatCompletionRequest`]'s metadata field]).
+pub struct ModelGradedScorer {
+    models: Arc<RwLock<LLMModelStore>>,
+    model_name: String,
+}
+
+impl ModelGradedScorer {
+    pub fn new(models: Arc<RwLock<LLMModelStore>>, model_name: String) -> Self {
```
phillipleblanc: Again, it feels weird to pass a collection of models and a name that indexes into that collection - why not just pass the model itself?

phillipleblanc (approved):

PR #5040, opened 2025-03-14
Improvements for evals

Jeadie:
# 🗣 Description
- Improve `spice eval ` CLI ux for when eval isn't found
- Add error logging to `POST v1/evals`.
- Ensure datasets in evals exist
- Add tracing component `eval_step` 
   ```bash
   >> spice trace eval_run
      Spice.ai OSS CLI v1.1.0-unstable-build.ce7f338fb-dev
        
      TREE                     STATUS DURATION   TASK
      33510c2b2b511e2b         ✅     10873.96ms eval_run
        ├── 409b6addaa3bb060   ✅       291.23ms sql_query
        ├── dda73ede1ca68688   ✅       730.96ms eval_step
        │ ├── fc1f0ff45f8926a8 ✅         0.69ms tool_use::list_datasets
        │ └── d7ab44726f559d13 ✅       727.34ms ai_completion
        ├── b93a96d01e783c09   ✅       984.65ms eval_step
        │ ├── 72562afa31854635 ✅         0.52ms tool_use::list_datasets
        │ └── 834c0f7582002403 ✅       982.20ms ai_completion
  ```
 - Add Levenshtein scorer for evals.

PR #5129, opened 2025-03-21
Eval errors aren't warning or debug logged in Runtime logs 

Sevenannn:
# 🗣 Description

Add debug loggings for errors running eval, otherwise some eval errors are not properly exposed in logs.

## 🔨 Related Issues

https://github.com/spiceai/spiceai/issues/5127

## ⛓️‍💥 Breaking Change


crates/runtime/src/http/v1/eval.rs
```diff
@@ -103,6 +103,7 @@ pub(crate) async fn post(
 
     let evals = rt.evals.read().await;
     let Some(eval) = evals.iter().find(|e| e.name == eval_name) else {
+        tracing::error!("Failed to run the eval `{eval_name}`. The eval does not exist.\nVerify the configuration and try again.");
```
phillipleblanc, line 106: Logging errors here because a model doesn't exist is a bad idea:

- **Noise in Logs**: Logging every instance of these errors floods the logs with entries that don’t represent actual system failures or unexpected behavior. For example, if a user queries an API for a resource they haven’t configured—like a missing endpoint or an uninitialized database table—this is a predictable outcome, not a program flaw. Excessive logging of such cases makes it harder to spot genuine issues, like server crashes or unhandled exceptions, because the signal gets buried in the noise.
- **User Errors Aren’t System Errors**: Misconfiguration or calling an API incorrectly is a user-side issue, not a malfunction of the program. Logging these as errors implies something went wrong with the system, which misleads developers or ops teams reviewing the logs. Instead, these cases are better handled with appropriate API responses—like a 400 Bad Request or 404 Not Found—letting the user know it’s their input or setup that’s off, without cluttering the error log.
- **Alert Fatigue**: If logs feed into monitoring systems that trigger alerts, logging user-induced errors can desensitize teams to real problems. Imagine getting paged at 3 a.m. because a user typo’d an API parameter—over time, this erodes trust in the alerting system and slows response to critical incidents.
- **Performance Overhead**: Writing to logs, especially at scale, introduces overhead. Disk I/O, network calls to a logging service, or even memory usage for buffering logs can add up. When the error is a routine, user-driven scenario (e.g., a 404 for a nonexistent resource), logging it repeatedly wastes resources on something that doesn’t need debugging or monitoring.

A better approach is to treat these as expected conditions: return a clear, user-facing error message via the API (e.g., “Resource not found—ensure it’s configured”) and reserve logging for unexpected, system-level failures (e.g., database connection drops). If tracking user errors is needed for analytics or debugging, consider a separate, lightweight mechanism—like a metrics counter—rather than full error logs. This keeps the system lean, focused, and easier to maintain.

phillipleblanc (changes requested):

Sevenannn:
Closing this issue as user-facing errors is returned when calling the api, which is fixed in this PR: https://github.com/spiceai/spiceai/pull/5040/files#diff-e2e6beee4754d828a9d8a6d950d3a3726a74bd063f02a8f4843bbdb9f110f411

PR #5161, opened 2025-03-27
Add github token to macos test release download tasks

Sevenannn:
# 🗣 Description

Add github token to macos tes release tasks, otherwise it would be rate limited due to the same ip address of macos runners and fail

Successful run: 
https://github.com/spiceai/spiceai/actions/runs/14105321097

## ⛓️‍💥 Breaking Change


phillipleblanc (approved):

lukekim (commented):
👍

Issue #2871, opened 2024-09-30
Automatic NSQL retries in `/v1/nsql`


Jeadie:
To mimic our ability for LLMs to repeatedly retry tools based on errors, the `/v1/nsql`, which does not use this same paradigm, should retry internally. 

If possible, improve the structured output to increase the likelihood of valid SQL in the response. Currently we just inforce JSON like this
```json
{
  "sql": "SELECT ..."
}
```

PR #5169, opened 2025-03-28
Add automatic retry for NSQL queries

sgrebnov:
# 🗣 Description

This PR introduces `QueryGenerationContext` which keeps track of previously generated and failed queries and is used to generate prompts for the next NSQL attempt.

If the generated NSQL query fails, the retry mechanism will make another attempt (up to 3 times) to generate and execute the query.

## 🔨 Related Issues

Implements 
- https://github.com/spiceai/spiceai/issues/2871

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

We might want to increase default num of retry attempts (currently 3 only) and make it configurable.

### DIFF
--- a/crates/llms/src/chat/nsql/mod.rs
+++ b/crates/llms/src/chat/nsql/mod.rs
@@ -20,12 +20,33 @@ pub mod default;
 pub(crate) mod json;
 pub(crate) mod structured_output;
 
+#[derive(Default)]
+pub struct QueryGenerationContext {
+    pub failed_attempts: Vec<FailedAttempt>,
+}
+
+pub struct FailedAttempt {
+    pub attempted_query: String,
+    pub error_message: String,
+}
+
+impl FailedAttempt {
+    #[must_use]
+    pub fn new(attempted_query: String, error_message: String) -> Self {
+        Self {
+            attempted_query,
+            error_message,
+        }
+    }
+}
+
 /// Additional methods (beyond [`super::Chat`]), whereby a model can provide improved results for SQL code generation.
 pub trait SqlGeneration: Sync + Send {
     fn create_request_for_query(
         &self,
         model_id: &str,
         query: &str,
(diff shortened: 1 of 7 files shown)


crates/runtime/src/http/v1/nsql.rs
```diff
@@ -190,6 +193,7 @@ LIMIT 5
         )))
     )
 ))]
+
 pub(crate) async fn post(
```
sgrebnov, line 197: TODO: refactor to reduce handler size (num lines)

crates/runtime/src/http/v1/nsql.rs
```diff
@@ -190,6 +193,7 @@ LIMIT 5
         )))
     )
 ))]
+#[allow(clippy::too_many_lines)]
```
sgrebnov, line 196: Did attempt to refactor but was unable to come up with good approach to decouple/simplify method logic. Most logic is converting errors to response message.

crates/llms/src/chat/nsql/mod.rs
```diff
@@ -36,10 +57,28 @@ pub trait SqlGeneration: Sync + Send {
...
+
+    if !ctx.failed_attempts.is_empty() {
+        let failed_atttempts_str = format!("\nUse incorrectly written SQL queries and associated errors below to ensure that the new query avoids repeating the same mistakes or generating identical queries:\n\n{}", failed_attempts_formatted(&ctx.failed_attempts));
+        prompt.push_str(&failed_atttempts_str);
+    }
+
+    prompt
+}
+
+fn failed_attempts_formatted(attempts: &Vec<FailedAttempt>) -> String {
```
sgrebnov, line 73: We should see if table representation works better, keep this version for simplicity and performance consideration (less payload and faster conversion)

phillipleblanc (approved):

Issue #5031, opened 2025-03-14
Make the `snowflake_sf1` TPCH benchmark run on testoperator


peasee:

peasee:
```console
 2025-03-14T04:23:27.430833Z  WARN runtime::init::dataset: Error initializing dataset lineitem. Failed to authenticate with Snowflake.
Verify your credentials, and try again.
Failed to authenticate. Error code: 390913. Message: Your free trial has ended and all of your virtual warehouses have been suspended. Add billing information in the Snowflake web UI to continue using the full set of Snowflake features.
```

PR #5190, opened 2025-03-31
fix: Testoperator DuckDB, SQLite, Postgres, Spicecloud

### LINKED ISSUE #5036: Make TPCDS `file_sqlite` benchmark run on the testoperator

### LINKED ISSUE #5206: Update MySQL TPCH query exclusions
Update MySQL TPCH exclusions for simple Q7 because of unsupported column renaming in table alias (`AS c(key)`).

peasee:
# 🗣 Description

For the testoperator:
* Fixes the TPCDS benchmarks for the DuckDB and SQLite accelerators
* Fixes the TPCDS Spice.AI connector benchmark
* Adds the Postgres accelerator for TPCH, TPCDS and ClickBench
* Adds some additional logging to the testoperator for pre-runs, explain plans and when total query set count is less than 5 to increase debugging verbosity (had issues debugging long running queries due to missing indexes).
* Removes Snowflake from the automatic testoperator schedule. It's not currently part of the existing benchmarks, so it can be added later as part of #5031

Notes:
* SQLite is set at TPCDS SF0.01, the same as the current benchmarks. It fails to run at 0.1 still.
* Postgres is set at 5 million rows for ClickBench vs 10 million for the existing benchmarks. This is because the existing benchmarks use an in-memory postgres instance. The testoperator uses a persistent benchmarking cluster, which is significantly slower.

## 🔨 Related Issues

* Closes #5036
* Closes #5206
* Part of #4114


crates/test-framework/src/snapshot/snapshots/explain/test_framework__snapshot__file_duckdb_file_tpcds_q19_explain.snap
```diff
@@ -25,6 +25,6 @@ description: "Query: tpcds_q19"
...
...item.i_brand ASC NULLS LAST, item.i_brand_id ASC NULLS LAST, item.i_manufact_id ASC NULLS LAST, item.i_manufact ASC NULLS LAST LIMIT 100 rewritten_sql=SELECT "item"."i_brand_id" AS "brand_id", "item"."i_brand" AS "brand", "item"."i_manufact_id", "item"."i_manufact", sum("store_sales"."ss_ext_sales_price") AS "ext_price" FROM "date_dim" JOIN "store_sales" ON (("date_dim"."d_date_sk" = "store_sales"."ss_sold_date_sk") AND (("date_dim"."d_moy" = 11) AND ("date_dim"."d_year" = 1999))) JOIN "item" ON (("store_sales"."ss_item_sk" = "item"."i_item_sk") AND ("item"."i_manager_id" = 8)) JOIN "customer" ON ("store_sales"."ss_customer_sk" = "customer"."c_customer_sk") JOIN "customer_address" ON ("customer"."c_current_addr_sk" = "customer_address"."ca_address_sk") JOIN "store" ON ((substr("customer_address"."ca_zip", 1, 5) <> substr("store"."s_zip", 1, 5)) AND ("store_sales"."ss_store_sk" = "store"."s_store_sk")) GROUP BY "item"."i_brand", "item"."i_brand_id", "item"."i_manufact_id", "item"."i_manufact" ORDER BY "ext_price" DESC NULLS FIRST, "item"."i_brand" ASC NULLS LAST, "item"."i_brand_id" ASC NULLS LAST, "item"."i_manufact_id" ASC NULLS LAST, "item"."i_manufact" ASC NULLS LAST LIMIT 100 |
```
phillipleblanc: This tmp folder is going to change on every run - you will need to use an insta filter, i.e. https://insta.rs/docs/filters/

We have an example of this in the codebase: https://github.com/spiceai/spiceai/blob/8324af0624966597f1c294c5653564a077201c8e/crates/runtime/benches/bench.rs#L713

peasee: I had assumed we were applying this filter all the time for the `.tmp` folder. I'll update to incorporate and regenerate.

peasee:
- [x] Update dispatch overrides for Spicecloud
- [x] Increase wait time for postgres-based acceleration dispatches

phillipleblanc (approved):

PR #5205, opened 2025-03-31
Add verification for `schema_source_path`

sgrebnov:
# 🗣 Description

Add logic to list files from the schema_source_path (but no more than SCHEMA_SOURCE_PATH_FILE_SCAN_LIMIT | 10_000) and identify the first file that matches the target extension. If no file matching the extension is found, report a warning or error (depending on whether we scanned all files or only the first SCHEMA_SOURCE_PATH_FILE_SCAN_LIMIT). This likely indicates that the schema_source_path parameter is invalid.

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/runtime/src/dataconnector/listing/connector.rs
+++ b/crates/runtime/src/dataconnector/listing/connector.rs
@@ -58,6 +58,9 @@ use crate::object_store_registry::default_runtime_env;
 use super::infer::infer_partitions_with_types;
 use super::DelimitedFormat;
 
+/// Maximum number of files to scan when validating that the schema source path contains objects with the expected extension.
+const SCHEMA_SOURCE_PATH_FILE_SCAN_LIMIT: usize = 10_000;
+
 #[async_trait]
 pub trait ListingTableConnector: DataConnector {
     fn as_any(&self) -> &dyn Any;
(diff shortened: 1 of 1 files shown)


crates/runtime/src/dataconnector/listing/connector.rs
```diff
@@ -684,6 +697,65 @@ async fn get_last_modified(
...
+            })?
+    {
+        if let Some(ext) = file.location.extension() {
+            if format!(".{ext}") == extension {
+                return Ok(());
+            }
+        };
+
+        scanned_files += 1;
+        if scanned_files > SCHEMA_SOURCE_PATH_FILE_SCAN_LIMIT {
```
sgrebnov, line 742: Concern here is that if `schema_source_path` containing too many files then schema inference will take a lot of time. Should we report this as well? Try to list first N files and if num of files larger than warn

crates/runtime/src/dataconnector/listing/connector.rs
```diff
@@ -684,6 +697,65 @@ async fn get_last_modified(
     }
 }
 
+async fn verify_schema_source_path(
```
sgrebnov, line 700: Alternative approach would be to re-use existing `get_last_modified` fn and pass `schema_source_path`, the only difference that we would like to exit early (no need to list everything), plus limit files to scan. Based on `schema_source_path` purpose I don't expect schema_source_path to point to the location with too many objects, so we can re-use get_last_modified

phillipleblanc (approved):

PR #5207, opened 2025-04-01
feat: Add testoperator search workflow

### LINKED ISSUE #3006: Enhancement: Search with keyword filtering
## Goal-State/What/Result

Vector search operations support optional keyword filtering, by performing a vector search on the resulting filtered lexical search.

## Why/Purpose

To improve the accuracy of vector searching, especially for LLM tool use, by pre-filtering results before attempting vector searching.

Because the result set used in the vector search will be smaller, performance will also improve.

## By When

_A target date for done-done completion of the entire enhancement._

**Issue/Spec written and reviewed:** 3rd Feb, 2025
**Done-Done:** 14th Feb, 2025

## Done-Done

- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Release Notes

## The Algorithm

- [x] Simplify.

## Spec

Models provide optional keywords to vector search tool use calls, which are used to pre-filter records for performing vector searches. The tool parameter is updated with a `keywords` parameter to accept an optional array of keywords.

The HTTP API will also be updated to support an optional `keywords` parameter, to override any model generated keywords during vector search:

- `keywords`: An array of keywords to perform a lexical search on the embeddings columns before performing a vector search

The keywords are applied in a logical `OR` to each embedding column across the dataset. For example, consider the following dataset embeddings configuration:

```yaml
datasets:
  - from: file:sales_data.parquet
    name: sales
    columns:
      - name: country
        description: The customer's country
        embeddings:
          - from: xl_embed
```

Searching this dataset with keywords like `["Australia", "South Korea"]` results in a logical filter expression like `WHERE (country LIKE '%Australia%' OR country LIKE '%South Korea%')`. The resulting rows of the search with this filter will be supplied for vector searching.

Keywords can be supplied in API requests like:

```json
{
"text": "string",
"keywords": ["one", "two", "three"],
...
}
```

Performance of vector searching before and after keywords should be measured, to identify any improvement or regression to performance as a result of keywords. Any performance considerations should be noted in the vector search documentation.

Support for inverted indexes will be implemented in a follow-up enhancement: https://github.com/spiceai/spiceai/issues/4594

## How/Plan

* The schema definition for vector search tool use is updated to allow a model to provide optional keywords for lexical pre-filtering before performing vector search over the filtered result set
* The vector search tool use implements lexical pre-filtering, performing a vector search on the filtered result set
* An optional keywords parameter is added to the HTTP API as an escape-hatch to override any keywords a model generates
* Compare performance of vector search before and after keywords, by running repeatable vector searches with and without keywords

### LINKED ISSUE #4767: Update search benchmarks to compare keyword filtering
Update search benchmarks to compare keyword filtering

peasee:
# 🗣 Description

* Adds a workflow for running vector search benchmarks via the `testoperator`

## 🔨 Related Issues

* Closes #4767 
* Closes #3006


.github/workflows/testoperator_run_search.yml
```diff
@@ -0,0 +1,63 @@
...
+
+jobs:
+  run-bench:
+    name: Run vector search tests
+    runs-on: ${{ github.event.inputs.runner_type }}
+    timeout-minutes: 600
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install MinIO
```
lukekim, line 31: Can we extract MinIO into it's own re-usable Action? Or, is there already one to do this?

peasee: It should be possible. It's used a lot in the other testoperator workflows too, so once this merges I will re-work to add a reusable action.

phillipleblanc (approved):

PR #5243, opened 2025-04-02
Add Support for AWS Session Token in S3 Data Connector

### LINKED ISSUE #5226: Bug: s3 data connector does not support session token
## Describe the bug

When using SSO / temp s3 access key during development, I can't specify `AWS_SESSION_TOKEN `

kczimm:
# 🗣 Description

This PR enhances the S3 data connector in Spice AI to support temporary AWS credentials enabling compatibility with AWS SSO and STS-based authentication. Key changes include:

**Added session_token Parameter:** Introduced `s3_session_token` as an optional parameter for the s3_auth=key method, allowing users to provide a session token alongside `s3_key` and `s3_secret` for temporary credentials.

**Flexible Validation:** Updated the key auth method to require only `s3_key` and `s3_secret`, with `s3_session_token` being optional. This preserves compatibility with static IAM credentials while adding support for temporary credentials.

**Improved Parameter Handling:** Refactored auth validation to use a loop for `key`, `secret`, and `session_token`, ensuring consistent error messages and reducing code duplication.

**Code Consistency:** Moved the "s3" string to a static `PREFIX` constant and updated references for better maintainability.

**Opt-out of Parameter Auto Loading:** Disabled auto-loading for the `s3_session_token` parameter to prevent implicity use when multiple S3 datasets are used.

## 🔨 Related Issues

Addresses #5226 

## ⛓️‍💥 Breaking Change

## 📄 Documentation Updates

* [ ] needs to be documented in the S3 data connector param list.

### DIFF
--- a/crates/runtime/src/dataconnector/s3.rs
+++ b/crates/runtime/src/dataconnector/s3.rs
@@ -23,7 +23,6 @@ use super::{
 use crate::parameters::ParamLookup;
 use crate::{component::dataset::Dataset, dataconnector::listing::LISTING_TABLE_PARAMETERS};
 
-use secrecy::ExposeSecret;
 use snafu::prelude::*;
 use std::any::Any;
 use std::clone::Clone;
@@ -136,7 +135,7 @@ static PARAMETERS: LazyLock<Vec<ParameterSpec>> = LazyLock::new(|| {
             ParameterSpec::component("endpoint").secret(),
             ParameterSpec::component("key").secret(),
             ParameterSpec::component("secret").secret(),
-            ParameterSpec::component("session_token").secret(),
+            ParameterSpec::component("session_token").secret().disable_auto_load(),
             ParameterSpec::component("auth")
                 .description("Configures the authentication method for S3. Supported methods are: public (i.e. no auth), iam_role, key.")
                 .secret(),
@@ -232,16 +231,6 @@ impl DataConnectorFactory for S3Factory {
                             return Err(Box::new(e) as Box<dyn std::error::Error + Send + Sync>);
                         }
                     }
-
-                    // Session token is optional; log if present
-                    if let ParamLookup::Present(secret) = params.parameters.get("session_token") {
-                        tracing::info!(
-                            "Using temporary credentials with session token for S3 auth."
-                        );
-
-                        // Set environment variable because session token cannot be used in URL fragment
-                        std::env::set_var("AWS_SESSION_TOKEN", secret.expose_secret());
-                    }
                 }
                 Some(auth) => {
                     return Err(Box::new(Error::UnsupportedAuthenticationMethod {
@@ -307,7 +296,7 @@ impl ListingTableConnector for S3 {
                 "client_timeout",
                 "allow_http",
                 "auth",
-                // "session_token" cannot be sent via URL fragments
+                "session_token",
             ],
         )));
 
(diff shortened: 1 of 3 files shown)


crates/runtime/src/dataconnector/s3.rs
```diff
@@ -208,31 +212,33 @@ impl DataConnectorFactory for S3Factory {
...
-                    }
-                    if matches!(params.parameters.get("secret"), ParamLookup::Present(_)) {
-                        // The 's3_secret' parameter cannot be set unless the `s3_auth` parameter is set to 'key'.
-                        return Err(Box::new(Error::InvalidAuthParameterCombination {
-                            parameter: "s3_secret".to_string(),
-                            auth: "key".to_string(),
-                        })
-                            as Box<dyn std::error::Error + Send + Sync>);
+                    // These parameters cannot be set unless the `s3_auth` parameter is set to 'key'.
+                    for param in ["key", "secret", "session_token"] {
```
sgrebnov, line 215: Nice 👍

sgrebnov (dismissed):

crates/runtime/src/dataconnector/s3.rs
```diff
@@ -208,31 +212,35 @@ impl DataConnectorFactory for S3Factory {
...
                     }
-                    if matches!(params.parameters.get("secret"), ParamLookup::Absent(_)) {
-                        return Err(Box::new(Error::NoAccessSecret)
-                            as Box<dyn std::error::Error + Send + Sync>);
+
+                    // Session token is optional; log if present
+                    if let ParamLookup::Present(secret) = params.parameters.get("session_token") {
+                        tracing::info!(
+                            "Using temporary credentials with session token for S3 auth."
+                        );
```
phillipleblanc:
```suggestion
```

I don't think we need to log this

crates/runtime/src/dataconnector/s3.rs
```diff
@@ -208,31 +212,35 @@ impl DataConnectorFactory for S3Factory {
...
+
+                        // Set environment variable because session token cannot be used in URL fragment
+                        std::env::set_var("AWS_SESSION_TOKEN", secret.expose_secret());
```
phillipleblanc: This shouldn't set an env var and instead should set it on the fragment, and set it on the object store instance from the fragment in `object_store_registry.rs`

kczimm:
Disabled auto loading for session token for spicepods like the following:

```yaml
datasets:
  - from: s3://private-bucket/data.csv
    name: data_sso
    params:
      file_format: csv
      s3_region: us-east-2
      s3_auth: key
      s3_key: ${ env:S3_KEY } 
      s3_secret: ${ env:S3_SECRET } 
      s3_session_token: ${ env:S3_SESSION_TOKEN } 

  - from: s3://private-bucket/data.csv
    name: data_iam
    params:
      file_format: csv
      s3_region: us-east-2
      s3_auth: key
      s3_key: ${ env:S3_KEY_2 } 
      s3_secret: ${ env:S3_SECRET_2 } 
```

I imagine we don't want the session token to be auto loaded for the `data_iam` dataset which is using IAM.

sgrebnov (approved):

phillipleblanc (approved):
Thanks! We'll also need to document this new option in https://github.com/spiceai/docs

Issue #5218, opened 2025-04-01
Integration test `test_acceleration_duckdb_single_instance` is not robust


Sevenannn:

Sevenannn:
Related: https://github.com/spiceai/spiceai/issues/5282

Sevenannn:
- Closed by: https://github.com/spiceai/spiceai/pull/5318

PR #5257, opened 2025-04-03
Enforce runtime graceful shutdown in integration tests

Sevenannn:
# 🗣 Description

Use `rt.shutdown()` before`drop(rt)` to enforce a runtime graceful shutdown in integration tests. This helps prevent issues caused by unclean shutdown of runtime.

Also added to use `ACCELERATION_MUTEX` for several tests that need to use shared status of acceleration registry, to avoid flaky tests caused by sharing the registry.

## 🔨 Related Issues

- #5218 
- #5233

## ⛓️‍💥 Breaking Change


Sevenannn:
This is closed because the Runtime static registry being shared by tests is likely the true cause of the bug. Following improvements will be tracked in this issue: https://github.com/spiceai/spiceai/pull/5257

Issue #5251, opened 2025-04-03
Enhancement: Results Cache Key Type (`sql` / `plan`)


phillipleblanc:
## Goal-State/What/Result

A new runtime parameter is added to the runtime:

```yaml
runtime:
  results_cache:
    cache_key_type: sql
```

**runtime.results_cache.cache_key_type**

Optional. Controls the cache key used for storing query results into the results cache.

### Supported Values

- **`plan` (Default)** – Uses the intermediate DataFusion LogicalPlan generated from the SQL text as the cache key. Ensures that minor non-semantic differences in queries can reuse the same query results cache.
- **`sql`** – Uses the raw SQL text as the cache key. Can provide a small performance improvement (~1-2ms) for queries that often hit the cache compared to `plan`.

## Why/Purpose

Allows customers the ability to tradeoff raw performance for cached queries at the cost of being more sensitive to minor query differences.

## By When

**Issue/Spec written and reviewed:** Apr 3, 2025
**Done-Done:** Apr 3, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

The request context module will receive this parameter on startup. For each request that comes in, we decide how to cache it via an enum `CacheControl`. This is defaulted to `CacheControl::Cache`, but can be overridden by the user setting the `Cache-Control` HTTP header. This change would extend the internal enum `CacheControl` to have an option for the `CacheKeyType`. To ensure that we could potentially re-use this for non-SQL results cache context, I won't use `SQL` in the CacheKeyType, instead opting for the terms `Default` and `Raw`. `Default` would map to `cache_key_type: plan` and `Raw` maps to `cache_key_type: sql`.

```rust
pub enum CacheKeyType {
    // i.e. cache_key_type: plan
    #[default]
    Default,
    // i.e. cache_key_type: sql
    Raw,
}

pub enum CacheControl {
    Cache(CacheKeyType),
[...]

PR #5265, opened 2025-04-03
Implements support for `runtime.results_cache.cache_key_type`

### LINKED ISSUE #5253: Implement `runtime.results_cache.cache_key_type`

phillipleblanc:
# 🗣 Description

Implements support for a new runtime parameter `runtime.results_cache.cache_key_type`.

By default, we will cache SQL query results based on the LogicalPlan they generate. This maps to:

```yaml
runtime:
  results_cache:
    cache_key_type: plan
```

This PR adds the ability for SQL query results to be cached based on the raw SQL string instead, bypassing the step to convert to a LogicalPlan first. This maps to:

```yaml
runtime:
  results_cache:
    cache_key_type: sql
```

I've added a new type `CacheKey` which can be either a `LogicalPlan` or a `String`:

```rust
pub enum CacheKey<'a> {
    LogicalPlan(&'a LogicalPlan),
    String(&'a str),
}
```

The corresponding interface for the `QueryResultCache` trait has been updated to use `CacheKey` instead of `LogicalPlan`:

```rust
pub trait QueryResultCache {
    async fn get<'a>(&self, key: CacheKey<'a>) -> Result<Option<CachedQueryResult>>;
    async fn put<'a>(&self, key: CacheKey<'a>, result: CachedQueryResult) -> Result<()>;
    async fn put_raw_key(&self, raw_key: RawCacheKey, result: CachedQueryResult) -> Result<()>;
    async fn invalidate_for_table(&self, table_name: TableReference) -> Result<()>;
    fn size_bytes(&self) -> u64;
    fn item_count(&self) -> u64;
}
```

The new `RawCacheKey` is added to represent the already-hashed value (so that we don't need to re-hash the key).

The `RequestContext` now updates the `CacheControl` to have the CacheKey it should use, i.e.:

```rust
pub enum CacheControl {
    Cache(CacheKeyType),
    NoCache,
}

pub enum CacheKeyType {
    /// Use the server's default logic (e.g., `LogicalPlan` hash).
    #[default]
    Default,
    /// Use the raw input (e.g., unparsed SQL string) as the cache key.
    Raw,
}
```

The `CacheKeyType` is currently only controllable via the runtime parameter `runtime.results_cache.cache_key_type`, but we can easily extend this in the future to pull from a header and use that as a cache key instead.

The logic to get the cached query results was quite messy, so I spent some time cleaning it up and ensuring that it would integrate cleanly with this new change. I moved most of the caching specific logic to `datafusion::query::cache`. The logic is convoluted, but I've tried to keep it as clean as possible while also being as performant as possible.

## 🔨 Related Issues

Closes #5253
Part of #5251

## 📄 Documentation Updates

Will require a documentation update for `runtime.results_cache.cache_key_type`.


crates/runtime/src/datafusion.rs
```diff
@@ -836,7 +836,7 @@ impl DataFusion {
 
         accelerated_table_builder.ready_state(dataset.ready_state);
 
-        accelerated_table_builder.cache_provider(self.cache_provider());
+        // accelerated_table_builder.cache_provider(self.cache_provider());
```
```suggestion
        accelerated_table_builder.cache_provider(self.cache_provider());
```

crates/runtime/src/datafusion/query/cache.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Returns a `LogicalPlan` if the result is not cached and needs to be executed, otherwise returns a cached `QueryResult`.
+    pub(super) async fn get_plan_or_cached(
+        df: &DataFusion,
+        session: &SessionState,
+        request_context: Arc<RequestContext>,
+        sql: &str,
+        tracker: QueryTracker,
+    ) -> super::Result<PlanOrCached> {
+        // Try to get cached results first from sql
+        let (tracker, cache_status) = match Self::try_get_cached_result(
```
sgrebnov, line 75: @phillipleblanc - could you please elaborate why we always try both CacheKey::String and CacheKey::LogicalPlan to retrieve cached result?

As we know what caching mechanism we will use for this request to store cached result (String vs LogicalPlan) can we just use the same for checking if there is cached value?

sgrebnov (dismissed):

crates/runtime/src/datafusion/query/cache.rs
phillipleblanc: Only one would possibly return results - having the check in `try_get_cached_result` makes it easier to follow the logic in this function.

sgrebnov: Got it 👍. Assume try_get_cached_result is cheap so we can call its underlying logic even if we can technically avoid this.

crates/runtime/src/datafusion/query/cache.rs
```diff
@@ -0,0 +1,386 @@
...
+                    result.results_cache_status,
+                    QueryResultsCacheStatus::CacheHit
+                );
+            }
+            Err(e) => panic!("Expected PlanOrCached::Cached, got {e:?}"),
+            Ok(PlanOrCached::Plan(_, _, _)) => panic!("Expected PlanOrCached::Cached, got Plan"),
+        }
+    }
+
+    // #[tokio::test]
```
Sevenannn: Are these tests supposed to be removed?

phillipleblanc: I'm still working on them

crates/cache/src/lib.rs
```diff
@@ -87,6 +87,30 @@ pub enum QueryResultsCacheStatus {
...
+            }
+            CacheKey::String(sql) => {
+                sql.hash(&mut hasher);
+                RawCacheKey(hasher.finish())
+            }
+        }
+    }
+}
+
+pub struct RawCacheKey(u64);
```
kczimm: Since `LogicalPlan` derives `Hash` can we simplify this by:

```rust
use std::hash::{Hash, Hasher, DefaultHasher};

#[derive(Hash)]
pub enum CacheKey<'a> {
    LogicalPlan(&'a LogicalPlan),
    String(&'a str),
}

impl CacheKey<'_> {
    #[must_use]
    pub fn as_raw_key(&self) -> RawCacheKey {
        let mut hasher = DefaultHasher::new();
        self.hash(&mut hasher);  // Directly hash the CacheKey
        RawCacheKey(hasher.finish())
    }
}
```

allowing future variants to be added without needing to update `as_raw_key`?

kczimm: Do we envision future functionality for `RawCacheKey` being added or is this merely an alias? If just an alias,

```rust
pub type RawCacheKey = u64;
```

will help us avoid the need to add `key.0` everywhere to use it.

phillipleblanc: I think the additional type safety we get from it is worth it being a newtype.

sgrebnov (approved):

PR #5284, opened 2025-04-04
Stablize the test `verify_similarity_search_chat_completion`

Sevenannn:
# 🗣 Description

Further stablize document similarity input by checking whether `journalist` and `vehicle` appear as part of the substring of the input instead of using a static snapshot.

This solves the issue that the input could be either `journalists` and `vehicles` or `journalist` and `vehicle` which causes an unstable snapshot.

Integration test (Model) passed 10/10.

## ⛓️‍💥 Breaking Change


crates/runtime/tests/models/openai.rs
```diff
@@ -560,9 +560,9 @@ async fn verify_similarity_search_chat_completion(
 
     for mut value in task_input {
         if let serde_json::Value::Object(ref mut map) = value {
-            // Remove the unstable "text" and "datasets" field
-            map.remove("text");
-            map.remove("datasets");
+            for (_, val) in map.iter_mut() {
+                *val = serde_json::Value::String("dummy_value".to_string());
```
phillipleblanc: Doesn't this somewhat defeat the point of the test?

Sevenannn: That makes sense. I updated the approach to do assertions on whether the field contains certain substring value that would appear in different result variations model give. e.g. `vehicle`, which covers `vehicle`, `vehicles`, or other longer phrases containing `vehicle`, this will be more robust than snapshotting.

Sevenannn:
~Found one more unstable test case: https://github.com/spiceai/spiceai/actions/runs/14346555634/job/40217491772?pr=5284. Will also add the fix for this test in the PR.~
Verified that this is occasionally caused by model not generating correct sql query as per request. Since the goal of the test is to validate whether model is able to generate SQL and successfully query, occasional failure due to model generating wrong query is acceptable

crates/runtime/tests/models/openai.rs
```diff
@@ -672,13 +672,31 @@ async fn verify_similarity_search_chat_completion(
...
-    insta::assert_snapshot!(
-        "chat_2_document_similarity_tasks",
-        serde_json::to_string_pretty(&combined_value).expect("Failed to serialize task_input")
+    assert!(
+        found_journalist,
+        "Expected to find 'journalists' in keywords"
+    );
+    assert!(
+        found_vehicle,
+        "Expected to find a keyword containing 'vehicle'"
```
phillipleblanc: Put the actual inputs into the error message as well, so that its clear to understand why the test fails if it does.

phillipleblanc (dismissed):

phillipleblanc (approved):

Issue #4960, opened 2025-03-10
Bug:`load_components` continue to loading  components during shutdown


sgrebnov:
## Describe the bug

Spawned in `load_components` tasks continue to run even when returned future is dropped and runtime is shutting down. 

```
^C2025-03-10T07:02:52.527617Z  INFO runtime: Shutting down runtime...
2025-03-10T07:03:01.647910Z  INFO runtime::init::dataset: Dataset lineitem registered (duckdb:lineitem), acceleration (duckdb:file, 5s refresh).
2025-03-10T07:03:01.649301Z  INFO runtime::accelerated_table::refresh_task: Loading data for dataset lineitem
2025-03-10T07:03:01.682434Z  INFO runtime::init::dataset: Dataset orders registered (duckdb:orders), acceleration (duckdb:file, 5s refresh).
2025-03-10T07:03:01.683848Z  INFO runtime::accelerated_table::refresh_task: Loading data for dataset orders
2025-03-10T07:03:01.718008Z  INFO runtime::init::dataset: Dataset part registered (duckdb:part), acceleration (duckdb:file, 5s refresh).
2025-03-10T07:03:01.718043Z  INFO runtime: Goodbye!
2025-03-10T07:03:01.719017Z  INFO runtime::accelerated_table::refresh_task: Loading data for dataset part
```

sgrebnov:
Fixed by https://github.com/spiceai/spiceai/pull/5306

PR #5306, opened 2025-04-07
Stop `load_components` during runtime shutdown

sgrebnov:
# 🗣 Description

`load_components` uses Tokio tasks to asynchronously load components, which continue running even if `load_components` is dropped or no longer awaited. This PR improves the behavior by sending a cancellation signal during runtime shutdown to abort all in-progress Tokio tasks.

Also moves `Goodbye` message trace to the very last program statement.

## 🔨 Related Issues

Fixes
- https://github.com/spiceai/spiceai/issues/4960
- https://github.com/spiceai/spiceai/issues/5093

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

The components load task is currently added to `server_components`, which is probably a runtime async task rather than a component. We might want to rename `server_components` so it reflects that it includes not only components, but also other jobs/tasks — or use a separate field specifically for the components load.


crates/runtime/src/lib.rs
```diff
@@ -645,17 +653,48 @@ impl Runtime {
...
+
+        // Signal that the load must be canceled if the runtime is shut down before the components are loaded
+        let cancel_loading = CancellationToken::new();
 
-        if let Err(err) = load_result {
-            tracing::error!("Could not start the Spice runtime: {err}");
+        // Wait for all components to load returning the first error
+        // or canceling spawned tokio tasks if the runtime is shutting down
+        let load_result = self
+            .start_server_component(
```
phillipleblanc: I think calling this a `server_component` is no longer accurate. Perhaps a better name is `runtime_task`?

phillipleblanc (dismissed):

crates/runtime/src/lib.rs
sgrebnov: Updated: https://github.com/spiceai/spiceai/pull/5306/commits/dc414742d05292b4990cfeb88ac757f80e382fd0

kczimm (dismissed):

kczimm (approved):

Issue #4932, opened 2025-03-07
Enhancement: Update testoperator dataset tests to validate against the query set


peasee:
## Goal-State/What/Result

Testoperator dataset tests (like `bench`, `throughput`, etc) which use query sets with verifiable results, validate the results against the expected results.

## Why/Purpose

To ensure the data returned from queries is valid.

## By When

**Issue/Spec written and reviewed:** Apr 3rd, 2025
**Done-Done:** Apr 10th, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

Query sets which contain validation for query results, like TPCH, perform this validation on the results of queries when performing `bench`, `throughput` and `load` tests.

The validation behavior should be off by default, and enabled with the `--validate` flag when running the testoperator test.

### Security Review

* There is no security impact with these changes

## How/Plan

The `QuerySet` enum is updated with an implementation of `QuerySet::validate()`. The return type of `get_tpch_test_queries` and other functions is also updated to return a `Vec<Query>`:

```rust
impl QuerySet {
    fn validate(&self, query: &Query, results: Vec<RecordBatch>) -> Result<QueryValidationResult> {
        match self {
            QuerySet::Tpch => tpch_query_validation(),
            ....
        }
    }
}

struct Query {
    name: String,
    query: String,
    overridden: bool
}

enum QueryValidationResult {
    Pass,
    Fail
}
```

Queries which fail validation return a `Fail`.

Queries which pass validation return a `Pass`.

For example, validation based on TPCH would retrieve expected row results from CSVs included at compile time. These CSVs contain the expected row results for a scale factor 1 query, and when using the same TPCH seed should expect identical results across TPCH generations.

Validation at scale factors higher than 1 are not supported.

peasee:
- [x] DuckDB Connector
- [x] DuckDB File Accelerator
- [x] DuckDB Memory Accelerator
[...]

PR #5311, opened 2025-04-07
feat: Add initial testoperator query validation

peasee:
# 🗣 Description

* Adds an initial `validation` sub-module as part of the `queries` module in the `test-framework`
* Loads expected answers for the TPCH queries from a CSV of TPCH answer results
* Updates some TPCH queries that require limiting to include the expected limit. For example, Query 18 does not contain a `LIMIT` in the SQL but the business question from the specification defines: `The Large Volume Customer Query finds a list of the top 100 customers ...`
* Adds schema and row count validation as part of the initial validation for TPCH queries
* Updates queries returned by the `queries` module to be represented by a new `Query` struct

## 🔨 Related Issues

* Part of #4932

## 🤔 Concerns

* I've been having some struggles getting schemas to be equivalent using the CSV answer source. Any brainstorming about alternative storage formats that would be easy to load without including many (if any) additional dependencies would be helpful. Alternatively, I've been considering just statically defining the expected schema.


crates/test-framework/Cargo.toml
```diff
@@ -14,6 +14,7 @@ app = { path = "../app" }
 arrow = { workspace = true, features = ["prettyprint"] }
 async-openai.workspace = true
 clap.workspace = true
+datafusion.workspace = true
```
phillipleblanc: What is this needed for? This will add a bunch of extra time to the compilation of testoperator (and anything else that uses test-framework)

peasee: I was using one of the datafusion functions during testing, this isn't needed anymore. I'll remove it

phillipleblanc (approved):

PR #5318, opened 2025-04-08
Refactor accelerator registry as part of runtime.

Sevenannn:
# 🗣 Description

Remove the static data accelerator registry `DATA_ACCELERATOR_ENGINES`, and refactor it as part of runtime.

**Key Changes:**

### New Sturct
`AcceleratorEngineRegistry`, containing a single field representing accelerator registry. `get_accelerator_engine`, `register_accelerator_engine`, `register_all`, `unregister_all`, `create_accelerator_table` are refactored as method in `AcceleratorEngineRegistry`.

The AcceleratorEngineRegistry is initiated in `RuntimeBuilder::new()`. It will then be used to build runtime and Datafusion.

### In Runtime
New field `Arc<AcceleratorEngineRegistry>` is added to Runtime Struct, the Runtime method `accelerator_registry()` is used to get a reference for various tasks that have access to runtime struct and requires the accelerator registry.

### In Datafusion
New field `Arc<AcceleratorEngineRegistry>` is added to Datafusion Struct, the Datafusion method `accelerator_registry()` is used to get a reference for various tasks that have access to runtime struct and requires the accelerator registry.

Removed the `with_runtime_status` and `with_datafusion` methods in DatafusionBuilder, which was merely used in integration tests and test CI for custom datafusion configuration. It is replaced by the `RuntimeBuilder::with_datafusion_configuration`, which takes a call back to apply customized configuration on `Datafusion` in tests.

 ### In Dataset
`Arc<Runtime>` is added as a new field in dataset. This is used for various tasks that requires access to the runtime `AcceleratorEngineRegistry`, including creating DuckDB, SQLite, Debezium tables.

The `runtime` and `app` become **required** field for `Dataset`, which means the runtime Dataset cannot be directly try_from spice pod dataset.

Create new struct `DatasetBuilder`, which implements `TryFrom` for spicepod Dataset, and builds a runtime `Dataset` when `runtime` and `app` are provided.

### In runtime intergation test
Remove `ACCELERATION_MUTEX` as the accelerator tests in integration tests no longer share the same registry

## ⛓️‍💥 Breaking Change


crates/runtime/src/builder.rs
```diff
@@ -125,15 +128,24 @@ impl RuntimeBuilder {
         self
     }
 
+    pub fn with_accelerator_registry(mut self, accelerator_registry: AcceleratorRegistry) -> Self {
```
phillipleblanc: Why is this needed? I would expect this to be an internal concern of the Runtime, not something people would construct externally and pass in.

crates/runtime/src/dataaccelerator/sqlite.rs
```diff
@@ -163,11 +168,11 @@ impl SqliteAccelerator {
...
-impl Default for SqliteAccelerator {
-    fn default() -> Self {
-        Self::new()
-    }
-}
+// impl Default for SqliteAccelerator {
+//     fn default() -> Self {
+//         Self::new()
+//     }
+// }
```
phillipleblanc:
```suggestion
```

crates/runtime/src/dataconnector/debezium.rs
```diff
@@ -62,11 +62,18 @@ pub enum Error {
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-#[derive(Debug)]
 pub struct Debezium {
     kafka_config: KafkaConfig,
 }
 
+impl std::fmt::Debug for Debezium {
```
phillipleblanc: Why was this change required?

crates/runtime/src/dataconnector/debezium.rs
```diff
@@ -145,9 +152,16 @@ impl Debezium {
     }
 }
 
-#[derive(Default, Debug, Copy, Clone)]
+#[derive(Default, Clone)]
 pub struct DebeziumFactory {}
 
+// For the DebeziumFactory struct
```
phillipleblanc: Why was this change required?

crates/runtime/src/init/dataset.rs
```diff
@@ -151,8 +150,14 @@ impl Runtime {
     }
 
     /// Returns a list of valid datasets from the given App, skipping any that fail to parse and logging an error for them.
-    pub(crate) fn get_valid_datasets(app: &Arc<App>, log_errors: LogErrors) -> Vec<Arc<Dataset>> {
-        Self::datasets_iter(app)
+    pub(crate) fn get_valid_datasets(
+        &self,
```
phillipleblanc:
```suggestion
        self: Arc<Self>,
```

We should just make this change an arc if we're going to need to clone it anyway.

crates/runtime/src/init/dataset.rs
```diff
@@ -167,12 +172,15 @@ impl Runtime {
             .collect()
     }
 
-    fn datasets_iter(app: &Arc<App>) -> impl Iterator<Item = Result<Dataset>> + '_ {
+    // TODO: with rt
```
phillipleblanc:
```suggestion
```

crates/runtime/src/component/dataset.rs
```diff
@@ -579,9 +591,15 @@ impl Dataset {
     }
 
     #[must_use]
-    pub async fn is_accelerator_initialized(&self) -> bool {
+    pub async fn is_accelerator_initialized(
+        &self,
+        accelerator_engine_registry: AcceleratorEngineRegistry,
```
phillipleblanc: We don't need to pass in the registry here, we can get it from the runtime instance on self.

crates/runtime/src/component/dataset.rs
```diff
@@ -220,6 +218,7 @@ pub struct Dataset {
     pub unsupported_type_action: Option<UnsupportedTypeAction>,
     pub ready_state: ReadyState,
     pub metrics: Metrics,
+    pub runtime: Option<Arc<Runtime>>,
```
phillipleblanc: I don't think it makes sense for a runtime dataset component to not ever have the Runtime reference, i.e. this shouldn't be Option<Arc<Runtime>>, but just Arc<Runtime>. I think that is also true for app.

What that means is that we wouldn't be able to construct a runtime Dataset directly from just a spicepod dataset anymore. We could have a new struct `DatasetBuilder` that implements TryFrom for the spicepod dataset, but in order to build it you need to supply a `runtime` and `app` reference.

So then the flow for constructing it would look like:

app.datasets
            .clone()
            .into_iter()
            .map(DatasetBuilder::try_from)
            .map(move |ds| {
                ds.map(|ds| ds.build(ds, Arc::clone(app), Arc::clone(&self)))
            })

crates/runtime/src/builder.rs
```diff
@@ -40,9 +44,10 @@ pub struct RuntimeBuilder {
     datasets_health_monitor_enabled: bool,
     metrics_endpoint: Option<SocketAddr>,
     prometheus_registry: Option<prometheus::Registry>,
-    datafusion: Option<Arc<DataFusion>>,
     runtime_status: Option<Arc<status::RuntimeStatus>>,
     rate_limits: Option<Arc<RateLimits>>,
+    accelerator_engine_registry: AcceleratorEngineRegistry,
+    datafusion_configuration: Option<DatafusionConfigurationCallback>,
```
phillipleblanc:
```suggestion
    datafusion_configuration_fn: Option<DatafusionConfigurationCallback>,
```

crates/runtime/src/builder.rs
```diff
@@ -115,8 +121,12 @@ impl RuntimeBuilder {
...
     }
 
-    pub fn with_datafusion(mut self, datafusion: Arc<DataFusion>) -> Self {
-        self.datafusion = Some(datafusion);
+    /// Used to configure `DataFusion` in integration tests & test CI
+    pub fn with_datafusion_configuration(
+        mut self,
+        callback: DatafusionConfigurationCallback,
+    ) -> Self {
+        self.datafusion_configuration = Some(callback);
```
phillipleblanc:
```suggestion
    /// Used to configure `DataFusion` in integration tests & test CI
    #[cfg(test)]
    pub fn with_datafusion_configuration(
        mut self,
        callback: DatafusionConfigurationCallback,
    ) -> Self {
        self.datafusion_configuration = Some(callback);
```

To make it explicit

crates/runtime/src/lib.rs
```diff
@@ -346,6 +348,7 @@ pub struct Runtime {
     status: Arc<status::RuntimeStatus>,
 
     runtime_tasks: Arc<RwLock<HashMap<String, CancellableTaskHandle>>>,
+    accelerator_engine_registry: AcceleratorEngineRegistry,
```
phillipleblanc: Shouldn't this be an Arc<>?

crates/runtime/src/dataaccelerator.rs
```diff
@@ -450,12 +466,13 @@ mod test {
...
     async fn test_file_mode_duckdb_creation() {
+        use crate::builder::RuntimeBuilder;
         use std::{fs, path::Path};
 
         let path = "./abc-duckdb.db".to_string();
         let params = HashMap::from([("duckdb_file".to_string(), path.clone())]);
-
-        register_all().await;
+        let runtime = Arc::new(RuntimeBuilder::new().build().await);
+        runtime.accelerator_engine_registry().register_all().await;
```
phillipleblanc: Shouldn't building the runtime already register all of the accelerators?

crates/runtime/src/dataaccelerator.rs
```diff
@@ -483,12 +502,13 @@ mod test {
...
     async fn test_file_mode_sqlite_creation() {
+        use crate::builder::RuntimeBuilder;
         use std::{fs, path::Path};
 
         let path = "./abc-sqlite.db".to_string();
         let params = HashMap::from([("sqlite_file".to_string(), path.clone())]);
-
-        register_all().await;
+        let runtime = Arc::new(RuntimeBuilder::new().build().await);
+        runtime.accelerator_engine_registry().register_all().await;
```
phillipleblanc:
```suggestion
```

crates/runtime/src/dataaccelerator.rs
```diff
@@ -517,15 +539,16 @@ mod test {
...
         use crate::make_spice_data_directory;
+        use std::{fs, path::Path};
 
         let spice_data_dir = crate::spice_data_base_path();
         make_spice_data_directory().expect("spice data directory created");
         let path = format!("{spice_data_dir}/abc_sqlite.db");
 
-        register_all().await;
+        let runtime = Arc::new(RuntimeBuilder::new().build().await);
+        runtime.accelerator_engine_registry().register_all().await;
```
phillipleblanc:
```suggestion
```

crates/runtime/src/dataaccelerator.rs
```diff
@@ -75,41 +74,166 @@ pub enum Error {
...
+    ) {
+        let mut registry = self.accelerator_engine_registry.lock().await;
+        registry.insert(name, accelerator_engine);
+    }
 
-pub async fn unregister_all() {
-    let mut registry = DATA_ACCELERATOR_ENGINES.lock().await;
-    registry.clear();
-}
+    pub async fn register_all(&self) {
```
phillipleblanc: I don't think we want to make this public, it should really be internal logic for the runtime builder. Perhaps a pub(crate)?

crates/runtime/src/datafusion.rs
```diff
@@ -259,6 +260,7 @@ pub struct DataFusion {
     cache_provider: RwLock<Option<Arc<QueryResultsCacheProvider>>>,
 
     pending_sink_tables: TokioRwLock<Vec<PendingSinkRegistration>>,
+    accelerator_engine_registry: AcceleratorEngineRegistry,
```
```suggestion
    accelerator_engine_registry: Arc<AcceleratorEngineRegistry>,
```

crates/runtime/src/lib.rs
```diff
@@ -379,6 +382,11 @@ impl Runtime {
         Arc::clone(&self.app)
     }
 
+    #[must_use]
+    pub fn accelerator_engine_registry(&self) -> AcceleratorEngineRegistry {
```
```suggestion
    pub(crate) fn accelerator_engine_registry(&self) -> AcceleratorEngineRegistry {
```

I don't think we need to expose this? Could we make this only visible to tests as well (I'm not sure what else requires this?)

crates/runtime/src/lib.rs
```diff
@@ -867,6 +875,12 @@ impl Runtime {
     }
 }
 
+impl std::fmt::Debug for Runtime {
```
phillipleblanc: This isn't a very useful Debug impl - why is this needed? I would say let's remove it or make it more useful.

crates/runtime/tests/integration.rs
```diff
@@ -78,13 +75,30 @@ mod utils;
...
+//         .config_mut()
+//         .options_mut()
+//         .execution
+//         .target_partitions = 3;
+//     let new_ctx = SessionContext::new_with_state(new_state);
+
+//     // Replace the old context with the modified one
+//     df.ctx = new_ctx.into();
+//     Arc::new(df)
+// }
```
phillipleblanc:
```suggestion
```

crates/runtime/tests/integration.rs
```diff
@@ -78,13 +75,30 @@ mod utils;
...
+
+fn configure_test_datafusion(df: &mut runtime::datafusion::DataFusion) {
```
phillipleblanc, line 80: Add back the comment from the original get_test_datafusion on why we do this.

crates/runtime/tests/integration.rs
```diff
@@ -78,13 +75,30 @@ mod utils;
...
     let mut new_state = df.ctx.state();
```
phillipleblanc: Its unnecessary to create a new state and then replace the datafusion context - all this needs to do is get a mutable reference to the config, and the runtime builder just calls this on its initial construction. That way we don't construct the session context twice.

crates/runtime/src/builder.rs
```diff
@@ -32,6 +34,8 @@ use crate::{
     tools, tracers,
 };
 
+type DatafusionConfigurationCallback = Box<dyn FnOnce(&mut DataFusion)>;
```
phillipleblanc: This could be a function pointer instead of a Box<dyn FnOnce>, which would remove a heap allocation.

https://doc.rust-lang.org/reference/types/function-pointer.html

See usage in Spice: https://github.com/spiceai/spiceai/blob/trunk/crates/data_components/src/unity_catalog/provider.rs#L47

crates/runtime/tests/acceleration/mod.rs
```diff
@@ -14,12 +14,13 @@ See the License for the specific language governing permissions and
...
-
 use runtime::{
-    component::dataset::Dataset, dataaccelerator::spice_sys::dataset_checkpoint::DatasetCheckpoint,
+    Runtime, component::dataset::Dataset,
+    dataaccelerator::spice_sys::dataset_checkpoint::DatasetCheckpoint,
 };
 use spicepod::{component::dataset::acceleration::Mode, param::Params};
+use std::sync::Arc;
+use std::sync::LazyLock;
 use tokio::sync::Mutex;
```
phillipleblanc: We shouldn't need the `ACCELERATION_MUTEX` anymore - please remove it.

crates/runtime/tests/integration.rs
```diff
@@ -78,24 +74,17 @@ mod utils;
...
-/// 1) Sets the number of `target_partitions` to 3, by default its the number of CPU cores available.
-fn get_test_datafusion(status: Arc<status::RuntimeStatus>) -> Arc<DataFusion> {
-    let mut df = DataFusion::builder(status).build();
-
-    // Set the target partitions to 3 to make RepartitionExec show consistent partitioning across machines with different CPU counts.
-    let mut new_state = df.ctx.state();
-    new_state
+// /// Gets a test `DataFusion` to make test results reproducible across all machines.
+// ///
+// /// 1) Sets the number of `target_partitions` to 3, by default its the number of CPU cores available.
```
phillipleblanc:
```suggestion
/// Modifies the `DataFusion` configuration to make test results reproducible across all machines.
///
/// 1) Sets the number of `target_partitions` to 3, by default its the number of CPU cores available.
```

crates/runtime/src/lib.rs
Sevenannn: It's needed in spicecloud crate when using `create_synced_internal_accelerated_table`

crates/runtime/src/component/dataset.rs
Sevenannn: Updated to make both ruuntime and app non-optional for runtime Dataset

crates/runtime/src/builder.rs
Sevenannn: The #[cfg(test)] won't apply to the integration tests in the tests/ directory, it only works for unit tests within the src/. Integration tests are considered a separate crate and will require this function as public.

crates/runtime/src/catalogconnector/mod.rs
```diff
@@ -69,6 +69,9 @@ pub enum Error {
         connector_component: ConnectorComponent,
         source: Box<dyn std::error::Error + Send + Sync>,
     },
+
+    #[snafu(display("Failed to get app from the runtime"))]
```
peasee: can this error be more descriptive? this is a potentially user-facing error in the SCP catalog connector?

crates/runtime/src/component/dataset.rs
```diff
@@ -85,6 +83,9 @@ pub enum Error {
 
     #[snafu(display("Error parsing `from` path {path} as table reference: {source}"))]
     UnableToParseTableReferenceFromPath { path: String, source: ParserError },
+
+    #[snafu(display("Unable to build Dataset: {missing_component} missing in DatasetBuilder"))]
```
peasee: If I am a user - what does this mean? How do/Can I resolve this? This sounds like an internal error that would need a bug submission.

crates/runtime/src/dataaccelerator.rs
```diff
@@ -75,41 +74,166 @@ pub enum Error {
...
+                msg: format!("Unknown engine: {engine}"),
+            }
+        })?;
+
+        if let Err(e) = acceleration_settings.validate_indexes(&schema) {
+            InvalidConfigurationSnafu {
+                msg: format!("{e}"),
+            }
+            .fail()?;
+        };
```
peasee, line 149: while we're making this refactor, can we take advantage of some new errors here? We can use `#[snafu(transparent)]` to passthrough these errors, instead of capturing them in a `format!()`. For example:

```rust
acceleration_settings.validate_indexes(&schema)?;
```

With a new error enum here:

```rust
    #[snafu(transparent)]
    DatasetComponentFailure {
        source: crate::component::dataset::Error,
    },
```

crates/runtime/src/dataaccelerator.rs
```diff
@@ -75,41 +74,166 @@ pub enum Error {
...
+                    .on_conflict(OnConflict::Upsert(ColumnReference::new(primary_keys)));
+            }
+        }
+
+        if let Some(on_conflict) =
+            acceleration_settings
+                .on_conflict()
+                .map_err(|e| Error::InvalidConfiguration {
+                    msg: format!("on_conflict invalid: {e}"),
+                })?
```
peasee, line 209: with a new transparent error this could change to:

```suggestion
                .on_conflict()?
```

crates/runtime/src/dataaccelerator.rs
```diff
@@ -75,41 +74,166 @@ pub enum Error {
...
+                }
+            }
+            Ok(None) => {}
+            Err(e) => {
+                InvalidConfigurationSnafu {
+                    msg: format!("{e}"),
+                }
+                .fail()?;
+            }
+        }
```
peasee, line 227:
```suggestion
        if let Some(constraints) = acceleration_settings.table_constraints(Arc::clone(&schema))? {
            if !constraints.is_empty() {
                external_table_builder = external_table_builder.constraints(constraints);
            }
        }
```

the transparent error will also simplify this match

crates/runtime/src/internal_table.rs
```diff
@@ -62,19 +63,42 @@ pub enum Error {
     UnableToBuildAcceleratedTable {
         source: AcceleratedTableBuilderError,
     },
+
+    #[snafu(display("Missing App From Runtime"))]
+    MissingAppFromRuntime {},
```
peasee:
```suggestion
```

This error is unused?

crates/spice_cloud/src/lib.rs
```diff
@@ -275,8 +281,22 @@ async fn get_spiceai_table_provider(
...
     cloud_dataset_path: &str,
     secrets: Arc<RwLock<Secrets>>,
+    runtime: Arc<Runtime>,
 ) -> Result<Arc<dyn TableProvider>, Error> {
-    let mut dataset = Dataset::try_new(cloud_dataset_path.to_string(), name)
+    let app_ref = runtime.app();
+    let app_lock = app_ref.read().await;
+    let Some(app) = app_lock.as_ref() else {
+        return Err(Error::UnableToCreateDataConnector {
+            source: "Missing App From Runtime".into(),
```
peasee, line 292: This source isn't returned in the error enum, so it will just print `"Unable to create data connector"`. Can this be turned into a more informative error? Probably another bug report error.

peasee (commented):
Some comments around error handling and messaging.

crates/runtime/src/dataconnector/debezium.rs
```diff
@@ -145,7 +145,7 @@ impl Debezium {
     }
 }
 
-#[derive(Default, Debug, Copy, Clone)]
+#[derive(Default, Debug, Clone)]
```
phillipleblanc: Why remove the `Copy`?

crates/runtime/src/init/dataset.rs
```diff
@@ -167,12 +170,22 @@ impl Runtime {
...
+    fn datasets_iter(self: Arc<Self>, app: &Arc<App>) -> impl Iterator<Item = Result<Dataset>> {
         app.datasets
             .clone()
             .into_iter()
-            .map(Dataset::try_from)
-            .map(move |ds| ds.map(|ds| Dataset::with_app(ds, Arc::clone(app))))
+            .map(DatasetBuilder::try_from)
+            .map(move |ds_builder_result| {
+                ds_builder_result.and_then(|ds_builder| {
+                    let dataset = ds_builder.name.to_string();
```
phillipleblanc:
```suggestion
                    let dataset_name = ds_builder.name.to_string();
```

crates/runtime/src/init/dataset.rs
```diff
@@ -649,8 +666,20 @@ impl Runtime {
         // Remove datasets that are no longer in the app
         for ds in &current_app.datasets {
             if !new_app.datasets.iter().any(|d| d.name == ds.name) {
-                let ds = match Dataset::try_from(ds.clone()) {
-                    Ok(ds) => ds,
+                let ds = match DatasetBuilder::try_from(ds.clone()) {
```
phillipleblanc: It feels a little weird that we need to construct a new dataset object to be able to remove it from being registered. Could we update `self.remove_dataset` to instead just take the dataset name?

crates/runtime/src/dataaccelerator.rs
```diff
@@ -75,41 +74,166 @@ pub enum Error {
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-static DATA_ACCELERATOR_ENGINES: LazyLock<Mutex<HashMap<Engine, Arc<dyn DataAccelerator>>>> =
-    LazyLock::new(|| Mutex::new(HashMap::new()));
+#[derive(Default, Clone)]
+pub struct AcceleratorEngineRegistry {
+    pub accelerator_engine_registry: Arc<Mutex<HashMap<Engine, Arc<dyn DataAccelerator>>>>,
```
phillipleblanc: We should probably make this an `RwLock` rather than a `Mutex` since we want to be able to read from it concurrently.

crates/runtime/src/lib.rs
```diff
@@ -297,6 +299,12 @@ pub enum Error {
     #[snafu(display("Unable to create directory: {source}"))]
     UnableToCreateDirectory { source: std::io::Error },
 
+    #[snafu(display("Unable to build dataset: {dataset}: {source}"))]
+    UnableToBuildDataset {
+        dataset: String,
+        source: Box<dyn std::error::Error + Send + Sync>,
```
phillipleblanc: Isn't this the Snafu error type from component::dataset?

crates/runtime/src/lib.rs
```diff
@@ -662,12 +678,15 @@ impl Runtime {
                     self_clone.load_eval_scorer().await;
                     let () = self_clone.verify_evals().await;
                     let an_eval_exists = app_lock.as_ref().is_some_and(|app| !app.evals.is_empty());
-                    if !an_eval_exists {
+                    if an_eval_exists {
+                        drop(app_lock);
+                        if let Err(err) = self_clone.load_eval_tables().await {
+                            tracing::warn!("Creating internal eval run table: {err}");
```
phillipleblanc: While we're refactoring, let's improve this message

crates/runtime/src/dataaccelerator.rs
Sevenannn: I'm not sure [snafu(transparent)] is what we want here, the `InvalidConfiguration` error variant display information other than the underlying error so it's not a simple pass through. Usage from trunk: https://github.com/spiceai/spiceai/blob/082076e8ff118a63dc1b8e029a17eda3ae81ad55/crates/runtime/src/dataaccelerator.rs#L61

phillipleblanc (approved):
Thanks for making this change!

Issue #4910, opened 2025-03-04
Enhancement: Support parameterized queries in the runtime


peasee:
## Goal-State/What/Result

Parameterized queries are supported via the Flight SQL and HTTP, like `SELECT $1`.

## Why/Purpose

Parameterized queries are a standard practice for supplying input parameters to queries. We should support the use of this standard practice.

## By When

**Issue/Spec written and reviewed:** March 4th, 2025
**Done-Done:** March 10th, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials
- [ ] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

**HTTP Endpoint**
```
POST /v1/sql
Headers: Content-Type: application/json
Body: {"query": "SELECT $1, $2", "params": [1, 2]}
```

The Spice Runtime supports parameterized/prepared queries, like `SELECT $1`, through the standard Flight interface.

### Security Review

* Parameterized/prepared queries are considered as more secure than non-prepared queries, as they prevent SQL injection attacks.

## How/Plan

* The Runtime is updated to DF 44, to [add this bugfix](https://github.com/apache/datafusion/pull/13632) for parameterized queries
* `do_put.rs` is updated to support bind parameters as part of the query interface
* A simple Python script is used to validate that the server supports

PR #5322, opened 2025-04-08
Support parameterized queries via HTTP API

kczimm:
# 🗣 Description

Add support for parameterized queries

## 🔨 Related Issues

Addresses #4910 

## ⛓️‍💥 Breaking Change

## 📄 Documentation Updates

Needs doc updates for each API

### DIFF
--- a/crates/runtime/src/datafusion/query.rs
+++ b/crates/runtime/src/datafusion/query.rs
@@ -207,7 +207,7 @@ impl Query {
                     e,
                     SchemaMismatch
                     )
-                }                
+                }
             }
 
             let final_stream = if cache_manager.should_cache_results() {
--- a/crates/runtime/src/datafusion/query.rs
+++ b/crates/runtime/src/datafusion/query.rs
@@ -24,6 +24,7 @@ use arrow::{
 use arrow_tools::schema::verify_schema;
 use cache::PlanOrCached;
 use datafusion::{
+    common::ParamValues,
     error::DataFusionError,
     execution::{context::SQLOptions, SendableRecordBatchStream},
     physical_plan::stream::RecordBatchStreamAdapter,
@@ -83,6 +84,7 @@ thread_local! {
 pub struct Query {
     df: Arc<crate::datafusion::DataFusion>,
     sql: Arc<str>,
+    params: Option<ParamValues>,
     tracker: QueryTracker,
 }
 
@@ -126,6 +128,13 @@ impl Query {
             {
                 PlanOrCached::Plan(plan, tracker, cache_manager) => (plan, tracker, cache_manager),
                 PlanOrCached::Cached(query_result) => return Ok(query_result),
+            };
+
+            let plan = if let Some(params) = ctx.params {
+                plan.with_param_values(params)
+                    .map_err(|e| Error::UnableToExecuteQuery { source: e })?
+            } else {
+                plan
             };
 
             if let Err(e) =
--- a/crates/runtime/src/datafusion/query.rs
+++ b/crates/runtime/src/datafusion/query.rs
@@ -197,14 +197,18 @@ impl Query {
             let res_schema = res_stream.schema();
 
             if let Err(e) = verify_schema(df_schema.fields(), res_schema.fields()) {
-                handle_error!(
+                // See https://github.com/apache/datafusion/blob/main/datafusion/expr/src/expr_schema.rs#L213
+                // Unspecified placeholder type is null and so parameters will not match schema
+                // If we get a mistmatch datatype and the actual is Null we just allow it
+                if matches!(&e, arrow_tools::schema::Error::SchemaMismatchDataType { actual, .. } if actual == "Null") {                handle_error!(
                     tracker,
                     &request_context,
                     ErrorCode::InternalError,
                     e,
                     SchemaMismatch
-                )
-            };
+                    )
+                }                
+            }
 
             let final_stream = if cache_manager.should_cache_results() {
                 Self::wrap_stream_with_cache(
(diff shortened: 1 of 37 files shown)


crates/runtime/src/datafusion/query.rs
```diff
@@ -128,6 +130,13 @@ impl Query {
                 PlanOrCached::Cached(query_result) => return Ok(query_result),
```
phillipleblanc, line 130: How does the results cache interact with parameterized queries?

crates/runtime/src/datafusion/query.rs
```diff
@@ -188,14 +197,18 @@ impl Query {
             let res_schema = res_stream.schema();
 
             if let Err(e) = verify_schema(df_schema.fields(), res_schema.fields()) {
-                handle_error!(
+                // See https://github.com/apache/datafusion/blob/main/datafusion/expr/src/expr_schema.rs#L213
+                // Unspecified placeholder type is null and so parameters will not match schema
+                // If we get a mistmatch datatype and the actual is Null we just allow it
+                if matches!(&e, arrow_tools::schema::Error::SchemaMismatchDataType { actual, .. } if actual == "Null") {                handle_error!(
```
phillipleblanc, line 203: We should only do this if the plan has a placeholder node, and ideally we verify that the null comes from that placeholder

crates/runtime/src/http/v1/query.rs
```diff
@@ -102,18 +105,379 @@ use super::{sql_to_http_response, ArrowFormat};
...
+)]
+pub(crate) async fn prepared(
+    Extension(df): Extension<Arc<DataFusion>>,
+    accept: Option<TypedHeader<Accept>>,
+    Json(PreparedStatement { sql, params }): Json<PreparedStatement>,
+) -> Response {
+    let params = match convert_json_to_param_values(params) {
+        Ok(params) => params,
+        Err(e) => {
+            tracing::debug!("Error reading params: {e}");
```
phillipleblanc, line 206: I don't think we should have a log that could spam the output that is user-controlled. That could be used by an attacker to attempt to hide their activity.

crates/runtime/src/http/v1/query.rs
```diff
@@ -102,18 +105,379 @@ use super::{sql_to_http_response, ArrowFormat};
...
+///
+/// # Supported JSON Formats
+/// - Array: Converted to positional parameters (ParamValues::Positional).
+/// - Object: Converted to named parameters (ParamValues::Named).
+/// - Primitive types (null, bool, number, string) are supported within arrays or objects.
+///
+/// # Errors
+/// - Returns an error if the top-level JSON is not an array or object.
+/// - Returns an error if any value cannot be converted to a ScalarValue.
+fn convert_json_to_param_values(json: Value) -> Result<ParamValues, String> {
```
phillipleblanc, line 236: Let's use a Snafu error instead of a String error, see https://github.com/spiceai/spiceai/blob/trunk/docs/dev/style_guide.md and https://github.com/spiceai/spiceai/blob/trunk/docs/dev/error_handling.md

crates/runtime/src/http/v1/query.rs
```diff
@@ -102,18 +105,379 @@ use super::{sql_to_http_response, ArrowFormat};
...
+                    .map_err(|e| format!("failed to convert value for key '{}': {}", key, e))?;
+                map.insert(key, scalar);
+            }
+            Ok(ParamValues::Map(map))
+        }
+        _ => Err("params must be a JSON array or object".to_string()),
+    }
+}
+
+/// Helper function to convert a single serde_json::Value to a ScalarValue.
```
phillipleblanc, line 261:
```suggestion
/// Helper function to convert a single serde_json::Value to a DataFusion ScalarValue.
```

crates/runtime/src/http/v1/query.rs
```diff
@@ -102,18 +105,379 @@ use super::{sql_to_http_response, ArrowFormat};
...
+/// # Returns
+/// * `Result<ScalarValue, String>` - The converted ScalarValue or an error message.
+///
+/// # Supported JSON Types
+/// - Null -> ScalarValue::Utf8(None)
+/// - Bool -> ScalarValue::Boolean(Some(bool))
+/// - Number -> ScalarValue::Int64(Some(i64)) if integer, else ScalarValue::Float64(Some(f64))
+/// - String -> ScalarValue::Utf8(Some(String))
+/// - Arrays and objects return an error (handled at a higher level).
+fn json_to_scalar(json: &Value) -> Result<ScalarValue, String> {
```
phillipleblanc, line 275: This is probably useful outside of the HTTP module context. Let's find a more general place for this to live. I would also double check that this functionality doesn't already exist in the DataFusion crate somewhere - it seems like a common thing to want to do?

phillipleblanc: This should also use Snafu errors

crates/runtime/src/http/routes.rs
```diff
@@ -135,7 +135,8 @@ pub(crate) fn routes(
     cors_config: &CorsConfig,
 ) -> Router {
     let mut authenticated_router = Router::new()
-        .route("/v1/sql", post(v1::query::post))
+        .route("/v1/sql", post(v1::query::sql))
+        .route("/v1/sql/prepared", post(v1::query::prepared))
```
phillipleblanc, line 139: I wouldn't have this be a separate API - UX should be reviewed by Luke, but I'd prefer this was combined with the `/v1/sql` API and distinguished using the `Content-Type` header.

crates/runtime/tests/s3/mod.rs
```diff
@@ -126,6 +126,60 @@ async fn s3_federation() -> Result<(), anyhow::Error> {
         .await
 }
 
+#[tokio::test]
+async fn s3_pdfs() -> Result<(), anyhow::Error> {
```
phillipleblanc, line 130: Looks like you will need to merge from trunk? I thought this code was already merged.

kczimm:
Closing as it's pretty outdated. Will start a new one.

PR #5362, opened 2025-04-14
Convert DataFusion filters to Delta Kernel predicates

phillipleblanc:
# 🗣 Description

Converts the DataFusion push-down filters into Delta Kernel predicates, which enables the Delta Kernel scan to optimize which files it considers for inclusion in the scan. In combination with https://github.com/delta-io/delta-kernel-rs/pull/713 - this enables partition pruning to happen in the scan itself, which should improve the performance of scans on larger Delta tables.

## 🔨 Related Issues

N/A


ewgenius (dismissed):

crates/data_components/src/delta_lake.rs
```diff
@@ -719,6 +738,278 @@ async fn get_parquet_access_plan(
...
+    for filter in filters {
+        if let Some(expr) = to_delta_kernel_expr(filter) {
+            exprs.push(expr);
+        }
+    }
+
+    if exprs.is_empty() {
+        None
+    } else if exprs.len() == 1 {
+        let expr = exprs.into_iter().next()?;
```
peasee:
```suggestion
        let expr = exprs.pop()?;
```

nit: Not sure why we're turning this into an iter when we can just pop the only item out?

peasee (dismissed):

peasee (approved):

Sevenannn (approved):

sgrebnov (approved):

Issue #5329, opened 2025-04-09
Enhancement: `accept: application/vnd.spiceai.sql.v1+json` support for `/sql` and `/nsql` API


sgrebnov:
When using the header:

```
accept: application/vnd.spiceai.sql.v1+json
```

The result **must be a JSON object** containing the following fields:

- `schema`
- `data`
- `row_count`
- `sql` *(only for nsql responses)*

### Example:

```json
{
  "row_count": 2,
  "schema": [
    {
      "name": "customer_id",
      "type": "string"
    },
    {
      "name": "total_sales",
      "type": "integer"
    }
  ],
  "data": [
    {
      "customer_id": "12345",
      "total_sales": 150000
    },
    {
      "customer_id": "67890",
      "total_sales": 125000
    }
  ],
  "sql": "SELECT customer_id, total_sales FROM sales_summary LIMIT 2;"
}
```

PR #5365, opened 2025-04-14
Extend `application/vnd.spiceai.sql.v1+json` with `schema` and `row_count` fields

sgrebnov:
# 🗣 Description

Add `schema` and `row_count` fields when using `application/vnd.spiceai.sql.v1+json` format.

```json
{
  "row_count": 2,
  "schema": {
    "fields": [
      {
        "name": "customer_id",
        "data_type": "Utf8",
        "nullable": false,
        "dict_id": 0,
        "dict_is_ordered": false,
        "metadata": {}
      },
      {
        "name": "total_sales",
        "data_type": "Int64",
        "nullable": false,
        "dict_id": 0,
        "dict_is_ordered": false,
        "metadata": {}
      }
    ],
    "metadata": {}
  },
  "data": [
...
```

## 🔨 Related Issues

Closes 
- https://github.com/spiceai/spiceai/issues/5329

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

1. The built-in Arrow JSON format for schema representation can be more complex to use compared to [JSON Schema](https://json-schema.org/understanding-json-schema/reference/type#footnotes), which supports only a limited set of types. In contrast, the Arrow schema format provides more detailed and granular type definitions, making it more closely aligned with Arrow's internal type system. For example, JSON Schema does not distinguish between integers and floating-point numbers—they are both represented as the `number` type:

> [JSON Schema](https://json-schema.org/understanding-json-schema/reference/type#footnotes):  
> JSON does not have separate types for integer and floating-point.

1.  `metadata` part `schema` does not seems to be very valuable, keeping it for consistency with Arrow Schema JSON format, we might want to remove metadata and flatten fielads so it is 
```json

{
  "row_count": 2,
  "schema": [
      {
        "name": "customer_id",
        "data_type": "Utf8",
        "nullable": false,
        "dict_id": 0,
        "dict_is_ordered": false,
        "metadata": {}
      },
      {
        "name": "total_sales",
        "data_type": "Int64",
        "nullable": false,
        "dict_id": 0,
        "dict_is_ordered": false,
        "metadata": {}
      }
    ],
...
```

### DIFF
--- a/crates/runtime/src/http/v1/mod.rs
+++ b/crates/runtime/src/http/v1/mod.rs
@@ -287,7 +287,7 @@ fn arrow_to_plain(
     pretty_format_batches(data).map(|d| format!("{d}")).boxed()
 }
 
-/// Converts a vector of `RecordBatch` to a application/vnd.spiceai.v1+json format
+/// Converts a vector of `RecordBatch` to an application/vnd.spiceai.sql.v1+json format
 fn arrow_to_vnd_sql_json_v1(
     data: &[RecordBatch],
     meta: ResponseMetadata,
--- a/crates/runtime/src/http/v1/mod.rs
+++ b/crates/runtime/src/http/v1/mod.rs
@@ -101,7 +101,7 @@ pub enum ResponseMimeType {
     Json,
     Csv,
     Plain,
-    VndJsonV1,
+    VndSqlJsonV1,
 }
 
 /// Represents additional metadata to produce a response, such as the SQL query used, etc.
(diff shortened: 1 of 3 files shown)


crates/runtime/src/http/v1/mod.rs
```diff
@@ -288,7 +288,7 @@ fn arrow_to_plain(
 }
 
 /// Converts a vector of `RecordBatch` to a application/vnd.spiceai.v1+json format
```
phillipleblanc:
```suggestion
/// Converts a vector of `RecordBatch` to an application/vnd.spiceai.sql.v1+json format
```

phillipleblanc (dismissed):

crates/runtime/src/http/v1/mod.rs
sgrebnov: Thank you!

lukekim (approved):

PR #5290, opened 2025-04-06
Implement Component Metrics system + MySQL connection pool metrics

phillipleblanc:
# 🗣 Description

> [!NOTE]  
> Most of the changed files are due to implementing the `Debug` trait on the data connectors. The other largest source of changed files is a refactoring of the Spicepod crate to pull non-components out of the `components` module. (i.e. `params` & `extension`).

A new `metrics` top-level node is added to each component that defines a list of component-specific metrics that can be enabled. Similar to parameters, each component defines a set of metrics that it exposes. By default these metrics are hidden from the Spice metrics. However, by including them in in the `metrics` list, they will be exposed in the Prometheus metrics.

Example definition:
```yaml
datasets:
  - from: mysql:my_table
    name: my_table
    metrics:
      - name: connection_count
        enabled: true
    params:
      mysql_host: localhost
      mysql_tcp_port: 3306
      mysql_user: root
      mysql_pass: ${ secrets:MYSQL_PASS }
```

Initially these component metrics are implemented for MySQL for exposing the following connection pool metrics:

```rust
    MetricSpec::new("connection_count", MetricType::ObservableGaugeU64)
        .description("Gauge of active connections to the database server"),
    MetricSpec::new("connections_in_pool", MetricType::ObservableGaugeU64)
        .description("Gauge of active connections that are idling in the pool"),
    MetricSpec::new("active_wait_requests", MetricType::ObservableGaugeU64).description(
        "Gauge of requests that are waiting for a connection to be returned to the pool",
    ),
    MetricSpec::new("create_failed", MetricType::ObservableCounterU64)
        .description("Counter of connections that failed to be created"),
    MetricSpec::new(
        "discarded_superfluous_connection",
        MetricType::ObservableCounterU64,
    )
        .description(
            "Counter of connections that were closed because there were already enough idle connections in the pool",
        ),
    MetricSpec::new("discarded_unestablished_connection", MetricType::ObservableCounterU64)
        .description(
            "Counter of connections that were closed because they could not be established",
        ),
    MetricSpec::new("dirty_connection_return", MetricType::ObservableCounterU64)
        .description(
            "Counter of connections that were returned to the pool but were dirty (ie. open transactions, pending queries, etc)",
        ),
    MetricSpec::new("discarded_expired_connection", MetricType::ObservableCounterU64)
        .description(
            "Counter of connections that were discarded because they were expired by the pool constraints (i.e. TTL expired)",
        ),
[...]

PR #5393, opened 2025-04-15
Track `ai_inferences_count` with used tools flag. Extensible runtime request context. 

ewgenius:
# 🗣 Description

Added the `ai_inferences_count` metric, which includes the `tools_used: bool` dimension. This metric tracks the number of AI inferences made by the runtime, with the `tools_used` dimension indicating whether **runtime** tools were utilized (Only built-in and MCP tools counted, client-defined tools are excluded).

If multiple tools are called during a single AI inference call (v1/chat/completions), it is still counted as one ai_inferences_count.

> [!NOTE]  
> `list_datasets` tool excluded from counting to track fair numbers, as it is always called if tools are enabled

Tool use tracking is managed through the request context extension `ModelContextExtension`, which allows to handle recursive chat completion calls and streaming. The model extension is injected only for the `/v1/chat/completions` API handler.

---

Request extensions example:

```rs

// create new extension
#[derive(Clone)]
pub struct MyContextExtension {
    <...>
}

impl MyContextExtension {
    <...>
}

// get current context and insert extension instance
let context = RequestContext::current(AsyncMarker::new().await);
context.insert_extension(MyContextExtension::new()).await;

// override scope with extended context
context
    .scope(async move {
        let mut inner_service = inner;
        inner_service.call(req).await
    })
```

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/runtime/src/metrics/telemetry/mod.rs
+++ b/crates/runtime/src/metrics/telemetry/mod.rs
@@ -90,3 +90,16 @@ pub fn track_query_execution_duration(duration: Duration, dimensions: &[KeyValue
     telemetry::track_query_execution_duration(duration, dimensions);
     QUERY_EXECUTION_DURATION_MS.record(duration.as_secs_f64() * 1000.0, dimensions);
 }
+
+static AI_INFERENCES_COUNT: LazyLock<Counter<u64>> = LazyLock::new(|| {
+    TELEMETRY_METER
+        .u64_counter("ai_inferences_count")
+        .with_description("Number of AI Inferences")
+        .with_unit("inferences")
+        .build()
+});
+
+pub fn track_ai_inferences_count(dimensions: &[KeyValue]) {
+    telemetry::track_ai_inferences_count(dimensions);
+    AI_INFERENCES_COUNT.add(1, dimensions);
+}
--- a/crates/runtime/src/metrics/telemetry/mod.rs
+++ b/crates/runtime/src/metrics/telemetry/mod.rs
@@ -99,7 +99,7 @@ static AI_INFERENCES_COUNT: LazyLock<Counter<u64>> = LazyLock::new(|| {
         .build()
 });
 
-pub fn track_ai_inferences_count(dimensions: &[KeyValue]) {
+pub fn track_ai_inferences_with_spice_count(dimensions: &[KeyValue]) {
     telemetry::track_ai_inferences_count(dimensions);
     AI_INFERENCES_COUNT.add(1, dimensions);
 }
--- a/crates/runtime/src/metrics/telemetry/mod.rs
+++ b/crates/runtime/src/metrics/telemetry/mod.rs
@@ -94,7 +94,7 @@ pub fn track_query_execution_duration(duration: Duration, dimensions: &[KeyValue
 static AI_INFERENCES_COUNT: LazyLock<Counter<u64>> = LazyLock::new(|| {
     TELEMETRY_METER
         .u64_counter("ai_inferences_with_spice_count")
-        .with_description("Number of AI Inferences")
+        .with_description("AI Inferences with Spice count")
         .with_unit("inferences")
         .build()
 });
--- a/crates/runtime/src/metrics/telemetry/mod.rs
+++ b/crates/runtime/src/metrics/telemetry/mod.rs
@@ -93,7 +93,7 @@ pub fn track_query_execution_duration(duration: Duration, dimensions: &[KeyValue
 
 static AI_INFERENCES_COUNT: LazyLock<Counter<u64>> = LazyLock::new(|| {
     TELEMETRY_METER
-        .u64_counter("ai_inferences_count")
+        .u64_counter("ai_inferences_with_spice_count")
         .with_description("Number of AI Inferences")
         .with_unit("inferences")
         .build()
--- a/crates/runtime/src/metrics/telemetry/mod.rs
+++ b/crates/runtime/src/metrics/telemetry/mod.rs
@@ -91,7 +91,7 @@ pub fn track_query_execution_duration(duration: Duration, dimensions: &[KeyValue
     QUERY_EXECUTION_DURATION_MS.record(duration.as_secs_f64() * 1000.0, dimensions);
 }
 
-static AI_INFERENCES_COUNT: LazyLock<Counter<u64>> = LazyLock::new(|| {
+static AI_INFERENCES_WITH_SPICE_COUNT: LazyLock<Counter<u64>> = LazyLock::new(|| {
     TELEMETRY_METER
         .u64_counter("ai_inferences_with_spice_count")
         .with_description("AI Inferences with Spice count")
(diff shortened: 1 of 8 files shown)


crates/telemetry/src/lib.rs
```diff
@@ -90,3 +90,15 @@ static QUERY_EXECUTION_DURATION_MS: LazyLock<Histogram<f64>> = LazyLock::new(||
...
     QUERY_EXECUTION_DURATION_MS.record(duration.as_secs_f64() * 1000.0, dimensions);
 }
+
+static AI_INFERENCES_COUNT: LazyLock<Counter<u64>> = LazyLock::new(|| {
+    METER
+        .u64_counter("ai_inferences_count")
+        .with_description("Number of AI Inferences")
+        .with_unit("inferences")
+        .build()
+});
```
kczimm: Why are we tracking this here and in `crates/runtime/src/metrics/telemetry/mod.rs`?

ewgenius: runtime::metrics::telemetry is essentially a wrapper for the telemetry crate
That was added in this PR
- https://github.com/spiceai/spiceai/pull/5290

crates/runtime/src/model/tool_use.rs
```diff
@@ -223,6 +224,20 @@ impl ToolUsingChat {
...
+            // exclude "list_datasets" tool from counting, since it is used on every AI inference call
+            let used_tools = spiced_tools
+                .iter()
+                .any(|t| t.function.name != "list_datasets");
+
+            if used_tools {
+                let context = RequestContext::current(AsyncMarker::new().await);
+                if let Some(model_context) = context.extension::<ModelContextExtension>().await {
+                    model_context.set_tools_used(true);
+                }
```
ewgenius: consider adding a panic in dev mode, to make sure we include it in all places where it is supposed to be included

crates/runtime/src/model/model_context.rs
```diff
@@ -0,0 +1,107 @@
...
+        Box::pin(async move {
+            let context = RequestContext::current(AsyncMarker::new().await);
+            context.insert_extension(ModelContextExtension::new()).await;
+
+            context
+                .scope(async move {
+                    let mut inner_service = inner;
+                    inner_service.call(req).await
+                })
+                .await
```
phillipleblanc: This extra scoping shouldn't be necessary

crates/runtime/src/model/tool_use.rs
```diff
@@ -303,6 +318,7 @@ impl ToolUsingChat {
     async fn chat_stream_inner(
         &self,
         req: CreateChatCompletionRequest,
+        is_root_call: bool,
```
phillipleblanc: As discussed, let's use a wrapping stream that emits the metric once the stream finishes.

crates/telemetry/src/lib.rs
phillipleblanc: We only submit a subset of our normal runtime metrics as anonymous telemetry metrics. We explicitly define both so that we are intentional about which metrics are submitted as anonymous ones vs which ones are runtime metrics only available locally.

Jeadie:
> This metric tracks the number of AI inferences made by the runtime, with the tools_used dimension indicating whether runtime tools were utilized (Only built-in and MCP tools counted, client-defined tools are excluded).

Clarifying on this point. We should be tracking all tool calls where the runtime is responsible for calling the tool and rerunning the LLM (and I believe we are already doing so). Currently this is only built-in (and tools derived from this) and MCP tools

crates/runtime/src/model/tool_use.rs
```diff
@@ -198,7 +199,7 @@ impl ToolUsingChat {
                 .into();
 
         let mut tool_and_response_content = vec![];
-        for t in spiced_tools {
+        for t in spiced_tools.clone() {
```
Jeadie, line 202: doesn't need cloning?

crates/runtime/src/metrics/telemetry/mod.rs
```diff
@@ -90,3 +90,16 @@ pub fn track_query_execution_duration(duration: Duration, dimensions: &[KeyValue
     telemetry::track_query_execution_duration(duration, dimensions);
     QUERY_EXECUTION_DURATION_MS.record(duration.as_secs_f64() * 1000.0, dimensions);
 }
+
+static AI_INFERENCES_COUNT: LazyLock<Counter<u64>> = LazyLock::new(|| {
+    TELEMETRY_METER
+        .u64_counter("ai_inferences_count")
```
lukekim:
```suggestion
        .u64_counter("ai_inferences_with_spice_count")
```

crates/runtime/src/metrics/telemetry/mod.rs
```diff
@@ -90,3 +90,16 @@ pub fn track_query_execution_duration(duration: Duration, dimensions: &[KeyValue
...
+        .with_description("Number of AI Inferences")
```
lukekim:
```suggestion
        .with_description("AI Inferences with Spice count")
```

crates/runtime/src/metrics/telemetry/mod.rs
```diff
@@ -90,3 +90,16 @@ pub fn track_query_execution_duration(duration: Duration, dimensions: &[KeyValue
...
+        .with_unit("inferences")
+        .build()
+});
+
+pub fn track_ai_inferences_count(dimensions: &[KeyValue]) {
```
lukekim:
```suggestion
pub fn track_ai_inferences_with_spice_count(dimensions: &[KeyValue]) {
```

crates/telemetry/src/lib.rs
```diff
@@ -90,3 +90,15 @@ static QUERY_EXECUTION_DURATION_MS: LazyLock<Histogram<f64>> = LazyLock::new(||
 pub fn track_query_execution_duration(duration: Duration, dimensions: &[KeyValue]) {
     QUERY_EXECUTION_DURATION_MS.record(duration.as_secs_f64() * 1000.0, dimensions);
 }
+
+static AI_INFERENCES_COUNT: LazyLock<Counter<u64>> = LazyLock::new(|| {
+    METER
+        .u64_counter("ai_inferences_count")
```
lukekim: Update as above.

crates/runtime/src/model/tool_use.rs
ewgenius: Added

crates/runtime/src/model/model_context.rs
ewgenius: Removed

crates/runtime/src/request/context.rs
```diff
@@ -28,12 +30,14 @@ use spicepod::component::runtime::UserAgentCollection;
 
 use super::{CacheControl, CacheKeyType, Protocol, UserAgent, baggage};
 
+type Extensions = HashMap<TypeId, Arc<dyn Any + Send + Sync>>;
+
 pub struct RequestContext {
-    // Use an AtomicU8 to allow updating the protocol without locking
```
phillipleblanc, line 32: Keep this comment?

crates/runtime/src/model/tool_use.rs
ewgenius: It is accessed below to filter out `list_datasets`, so have to clone

crates/runtime/src/model/tool_use.rs
ewgenius: resolved

crates/runtime/src/request/context.rs
ewgenius: Added back

crates/telemetry/src/lib.rs
ewgenius: updated

phillipleblanc (dismissed):

phillipleblanc (approved):

Sevenannn (approved):

PR #5397, opened 2025-04-16
feat: Write OTel metrics for testoperator

### LINKED ISSUE #5359: Create an OTel builder in testoperator to submit benchmark metrics
Create an OTel builder in testoperator to submit benchmark metrics

peasee:
# 🗣 Description

* Sets up metrics submission for testoperator, using a new `Telemetry` struct in `test-framework`.
* Uses the similar implementation from anonymous telemetry for specifying a meter provider, allowing a no-op meter provider to be specified when metrics submission is desired to be disabled.
* Updates `Credentials` in `flight_client` to use a `SecretString` for the API key/password instead of a plain string
* Updates `AnonymouseTelemetryExporter` to `TelemetryExporter`. This now has a builder, which optionally takes an API key to create an authenticated telemetry exporter
* Uses the authenticated telemetry exporter in testoperator, when `SPICEAI_API_KEY` is set
* Refactors some of the testoperator code to remove the setup/management of writing to the "results" dataset, now that we'll be using OTel for that instead.
* Adds `--metrics` as a flag for the testoperator. By default, metrics submissions is disabled but can be enabled by specifying the `--metrics` flag.

## 🔨 Related Issues

* Closes #5359


tools/testoperator/src/telemetry.rs
peasee, line 1: I think this file is a candidate for code de-dup with the `telemetry` crate. I'll look into this in follow-up PRs.

tools/testoperator/src/telemetry.rs
```diff
@@ -0,0 +1,112 @@
...
+
+pub(crate) static METER: LazyLock<Meter> =
+    LazyLock::new(|| METER_PROVIDER.meter("benchmarks_telemetry"));
+
+pub(crate) fn setup(resource: Resource) {
+    // TODO: setup an exporter with API key
+    let provider = SdkMeterProvider::builder().with_resource(resource).build();
+
+    if METER_PROVIDER_ONCE.set(Arc::new(provider)).is_err() {
+        println!("Testoperator metrics are disabled");
```
Sevenannn: Does the METER_PROVIDER_ONCE set failure always means that test operator metrics are disabled, or it could be caused by other factors?

peasee: Because it is a `OnceLock`, if it is already set (e.g. setting it to a `NoopProvider` to prevent recording metrics) then this function will error. `METRIC_PROVIDER_ONCE` will only ever be set externally if metrics collection is being disabled - but this is not implemented in this PR yet.

crates/test-framework/src/metrics/mod.rs
```diff
@@ -400,24 +407,24 @@ impl<T: ExtendedMetrics, R: ExtendedMetrics> QueryMetrics<T, R> {
...
+            Field::new("min_duration_ms", DataType::UInt64, false),
+            Field::new("max_duration_ms", DataType::UInt64, false),
+            Field::new("iterations", DataType::UInt64, false),
             Field::new("commit_sha", DataType::Utf8, false),
             Field::new("branch_name", DataType::Utf8, false),
         ];
 
         base_fields.extend(extended_fields);
 
         base_fields.extend(vec![
```
phillipleblanc: We can put the extended fields at the end - I had them split up like this to match the existing metrics table format, but since we don't need to do that anymore we can add them at the end.

tools/testoperator/src/commands/bench/mod.rs
```diff
@@ -81,17 +73,47 @@ pub(crate) async fn run(args: &DatasetTestArgs) -> anyhow::Result<RowCounts> {
...
-        put_batches(&mut flight_client, TEST_RESULTS_DATASET, records).await?;
+    let commit_sha = metrics.commit_sha.clone();
+    let spiced_version = metrics.spiced_version.clone();
+    let app_name = app.name.clone();
+    let benchmark_resource = Resource::new(vec![
+        KeyValue::new("service.name", "testoperator"),
+        KeyValue::new("benchmark.name", app_name.clone()),
+        KeyValue::new("benchmark.spiced_version", spiced_version.clone()),
+        KeyValue::new("benchmark.query_set", query_set.to_string()),
+        KeyValue::new("benchmark.commit_sha", commit_sha.clone()),
```
phillipleblanc:
```suggestion
        KeyValue::new("benchmark.spiced_commit_sha", commit_sha.clone()),
```

tools/testoperator/src/commands/bench/mod.rs
```diff
@@ -81,17 +73,47 @@ pub(crate) async fn run(args: &DatasetTestArgs) -> anyhow::Result<RowCounts> {
...
+    ]);
+
+    crate::telemetry::setup(benchmark_resource);
+
+    for query in &metrics.metrics {
+        let query_name = query.query_name.clone();
+        let row_count = row_counts.get(&query_name).unwrap_or(&0);
+        let attributes = vec![
+            KeyValue::new("query_name", query_name),
+            KeyValue::new("commit_sha", commit_sha.clone()),
```
phillipleblanc:
```suggestion
            KeyValue::new("spiced_commit_sha", commit_sha.clone()),
```

tools/testoperator/src/telemetry.rs
```diff
@@ -0,0 +1,112 @@
...
+*/
+
+use std::sync::{Arc, LazyLock, OnceLock};
+
+use opentelemetry::metrics::{Gauge, Meter, MeterProvider};
+
+use opentelemetry_sdk::{Resource, metrics::SdkMeterProvider};
+use telemetry::noop::NoopMeterProvider;
+
+pub(crate) static METER_PROVIDER_ONCE: OnceLock<Arc<dyn MeterProvider + Send + Sync>> =
```
phillipleblanc: I would put this code in the `test-framework` crate, not in the `testoperator` binary directly. That way we could reuse it for other more specialized binaries in the future if we needed to.

phillipleblanc: I would also split the meter/exporter creation from the individiual metrics themselves.

crates/test-framework/src/telemetry/mod.rs
```diff
@@ -0,0 +1,109 @@
...
+        };
+
+        Self {
+            reader,
+            resource: resource.clone(),
+            setup,
+        }
+    }
+
+    pub async fn read(&self, api_key: Option<&str>) -> Result<()> {
```
phillipleblanc:
```suggestion
    pub async fn emit(&self, api_key: Option<&str>) -> Result<()> {
```

Perhaps a little clearer?

phillipleblanc: Also should the API key be part of the `Telemetry` struct? Seems weird that we need to specify it each time.

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #5402, opened 2025-04-16
Feat/testoperator otel exporter

peasee:
# 🗣 Description

* Updates `Credentials` in `flight_client` to use a `SecretString` for the API key/password instead of a plain string
* Updates `AnonymouseTelemetryExporter` to `TelemetryExporter`. This now has a builder, which optionally takes an API key to create an authenticated telemetry exporter
* Uses the authenticated telemetry exporter in testoperator, when `SPICEAI_API_KEY` is set
* Refactors some of the telemetry setup code to move back into the `telemetry` crate, reducing code duplication. This required updating some scopes in `telemetry` to `pub` instead of `pub(crate)`.
* Refactors some of the testoperator code to remove the setup/management of writing to the "results" dataset, now that we'll be using OTel for that instead.

## 🔨 Related Issues

* Part of #5359


crates/telemetry/src/exporter.rs
```diff
@@ -14,41 +14,87 @@ See the License for the specific language governing permissions and
...
 use arrow::array::RecordBatch;
 use async_trait::async_trait;
 use flight_client::{Credentials, FlightClient};
 use opentelemetry_sdk::metrics::MetricError;
+use secrecy::SecretString;
 
-#[derive(Debug, Clone)]
-pub struct AnonymousTelemetryExporter {
-    flight_client: Option<FlightClient>,
+const ENDPOINT_CONST: &str = "https://telemetry.spiceai.io";
```
phillipleblanc: This should still live in the anonymous module

crates/telemetry/src/exporter.rs
```diff
@@ -14,41 +14,87 @@ See the License for the specific language governing permissions and
...
+
+pub static ENDPOINT: LazyLock<Arc<str>> = LazyLock::new(|| {
```
phillipleblanc: Same for this - the telemetry builder should require an endpoint as input. The telemetry.spiceai.io endpoint should only be used for anonymous telemetry

Issue #5291, opened 2025-04-06
Support Llama 4 model family


lukekim:
## By When

**Issue/Spec written and reviewed:** _Target Date_
**Done-Done:** _Target Date_

## Done-Done

- [ ] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials
- [ ] Release Notes prepared

## The Algorithm

- [ ] Every requirement questioned?
- [ ] Delete (Scope) any part you can.
- [ ] Simplify.
- [ ] Break down into smaller iterations/milestones.
- [ ] Opportunities for automation.

lukekim:
Requires: https://github.com/EricLBuehler/mistral.rs/pull/1268

EricLBuehler:
Hi @lukekim! Saw the PR was mentioned here - full Llama 4 support was just merged!

lukekim:
> Hi @lukekim! Saw the PR was mentioned here - full Llama 4 support was just merged!

Thank you!

lukekim:
Thanks for the report! This appears to be fixed in trunk by 9b2858a2c1 (https://github.com/spiceai/spiceai/commit/null). Closing for now — please reopen if you can still reproduce on a release that includes this commit.

PR #5404, opened 2025-04-16
Upgrade mistral.rs

Jeadie:
# 🗣 Description
 - This PR updates the [mistral.rs](https://github.com/EricLBuehler/mistral.rs) dependency primarily to support LLama4.
 - To support Llama4, we will need to support vision models, this PR will not implement that yet.
## 🔨 Related Issues
 - Needs https://github.com/spiceai/mistral.rs/pull/27
 - #5291


bin/spiced/src/lib.rs
```diff
@@ -376,7 +376,7 @@ fn apply_overrides(
                 }
                 .fail();
             }
-        };
```
Jeadie: Many syntax changes came from the new rustc version and running 
```bash
cargo clippy --all-targets --allow-dirty --features aws-secrets-manager,keyring-secret-store,models,odbc,release --workspace --fix
```

phillipleblanc: Are you running 1.86? If you pin to 1.85 these shouldn't pop up. I'm ok to fix these now, but it should come in a separate PR - not one that is tied to the Llama 4 upgrade.

phillipleblanc (dismissed):

peasee (approved):

Issue #5394, opened 2025-04-15
Bug: `Underlying language model produced no score in response` error


sgrebnov:
## Describe the bug

There is the following error when I use model scorer

```bash
curl -XPOST "http://localhost:8090/v1/evals/financebench" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "gpt-4o"
  }'
[{"id":"ef86182525936e6dbd4d5a46bcd222f5","created_at":"2025-04-15T20:10:18","dataset":"financebench.evals","model":"gpt-4o","status":"Failed","error_message":"An error occured whilst scoring the results of the eval run. Scorer failed: judge - Could not score case [input=UserInput(\"What is the FY2019 fixed asset turnover ratio for Activision Blizzard? Fixed asset turnover ratio is defined as: FY2019 revenue / (average PP&E between FY2018 and FY2019). Round your answer to two decimal places. Base your judgments on the information provided primarily in the statement of income and the statement of financial position.\"), actual=AssistantResponse(\"I could not retrieve specific financial data for Activision Blizzard for FY2019, including revenue and PP&E values. Therefore, I am unable to calculate the fixed asset turnover ratio for FY2019.\"), expected=AssistantResponse(\"24.26\")]. Error: Underlying language model produced no score in response","scorers":["judge"],"metrics":{}}]%  
```

## To Reproduce

```
- name: judge
    from: openai:gpt-4o
    params:
      openai_api_key: ${ secrets:SPICE_OPENAI_API_KEY }
      parameterized_prompt: enabled
      system_prompt: |
        You are a financial domain expert acting as a **scorer** to evaluate the **correctness and quality** of a generated response from a Retrieval-Augmented Generation (RAG) system.

        You are provided with:
          - A **User Question**.
          - A **Reference Answer**, which is the correct and complete response.
          - A **Generated Answer**, produced by the RAG model.

        Your evaluation should focus on:
          1. **Correctness**: Does the generated answer contain accurate factual information based on the question and context?
          2. **Completeness**: Does it address all relevant aspects of the question?
          3. **Relevance**: Is the information included pertinent and within scope?
          4. **Faithfulness**: Is the content grounded in the retrieved documents and avoids hallucination or fabrication?
          5. **Clarity**: Is the reasoning clear and coherent (when applicable)?

        Specific evaluation guidance:
          - For **numerical answers**, check precision, rounding, and if values match the correct source.
[...]

PR #5405, opened 2025-04-16
Retry model graded scorer once on successful, empty response

Jeadie:
# 🗣 Description
 - A single erroneous LLM judge response can derail a full `v1/evals` run. Often this is because of invalid JSON (even with structured output applied). 
 - This PR adds a single retry when this occurs.

## 🔨 Related Issues
#5394

### DIFF
--- a/crates/runtime/src/model/eval/scorer/modelgraded.rs
+++ b/crates/runtime/src/model/eval/scorer/modelgraded.rs
@@ -64,6 +64,41 @@ impl ModelGradedScorer {
             .build()
     }
 
+    /// Attempt to call a model graded LLM scorer and parse the `score` JSON key.
+    /// Returns `Ok(None)` if the model call was successful, but no `score` was found in response (score must also be a valid number).
+    async fn attempt_score(
+        &self,
+        req: &CreateChatCompletionRequest,
+    ) -> super::Result<Option<Number>> {
+        let response =
+            self.model
+                .chat_request(req.clone())
+                .await
+                .map_err(|e| Error::ErrorScoringCase {
+                    input: input.clone(),
+                    actual: actual.clone(),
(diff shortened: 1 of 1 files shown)


crates/runtime/src/model/eval/scorer/modelgraded.rs
```diff
@@ -93,51 +128,35 @@ impl Scorer for ModelGradedScorer {
...
-            });
-        };
-        let Ok(Some(Value::Number(score))) =
-            serde_json::from_str::<Value>(content.as_str()).map(|v| v.get("score").cloned())
-        else {
-            return Err(Error::ErrorScoringCase {
+        let mut score = self.attempt_score(&req).await?;
+
+        // Retry once for when LLM scorer was successfully called, but `score` key was not returned.
+        if score.is_none() {
```
sgrebnov, line 134: We can add debug tracing here

sgrebnov (dismissed):

phillipleblanc (approved):

sgrebnov (approved):

crates/runtime/src/model/eval/scorer/modelgraded.rs
```diff
@@ -60,10 +60,37 @@ impl ModelGradedScorer {
                     .build()?
                     .into(),
             ])
-            .store(true)
```
sgrebnov, line 63: @Jeadie - why we remove `store` here? I was under impression that it is required for OpenAI parametrized input we use.

Jeadie: No, it isn't used in parameterised input. For parameterised input, if `store=true`, we don't evict the metadata fields. In the case of evals, we know we want to remove the metadata fields so we should have `store=false` (which is default).

Issue #5143, opened 2025-03-24
spiceai connector `grpc: received message larger than max` error


sgrebnov:
## Describe the bug

Unable to retrieve large spiceai dataset

>External error: Execution error: Query execution failed.
Tonic error: status: ResourceExhausted, message: "DoGet recv error: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (150018393 vs. 104857600)", details: [], metadata: MetadataMap { headers: {} }
Verify the configuration and try again.
Ensure the dataset configuration is valid, and try again.

## To Reproduce

```yaml
  - from: spice.ai/spiceai/financebench/datasets/spice.financebench.data
    name: financebench.data
    description: Financial information
    params:
      spiceai_api_key: ${ secrets:SPICEAI_API_KEY }
      spiceai_endpoint: https://flight.spiceai.io
    acceleration:
      enabled: true
```

sgrebnov:
Size is configured by the following params:
- https://github.com/spiceai/spiceai/pull/4116/files

MAX_ENCODING_MESSAGE_SIZE
MAX_DECODING_MESSAGE_SIZE

PR #5407, opened 2025-04-16
Allow `flightsql` and `spiceai` connectors to override flight max message size

sgrebnov:
# 🗣 Description

This PR introduces a new `runtime.flight.max_message_size` parameter that allows `flightsql` and `spiceai` connectors to override the default values (100 MB) for Flight's `max_encoding_message_size` and `max_decoding_message_size`, enabling support for receiving larger Flight messages.

```yaml
runtime:
  flight:
    max_message_size: 512MiB 
``` 

Notes: 
 1. support for Dremio connector is not included in this PR, can be added the same way as support for `spiceai`.
 2. Parameter is not currently applied to Flight Server configuration

## 🔨 Related Issues

Fixes 
- https://github.com/spiceai/spiceai/issues/5143

## ⛓️‍💥 Breaking Change

## 📄 Documentation Updates

Support for new parameter must be documented

## 🤔 Concerns

1. Passing additional reference to the app configuration extends scope of `ConnectorParams`, but I was unable to come up with alternative good approach to propagate runtime configuration to connectors.
2. `max_message_size` is runtime / not dataset-level configuration - this is due to plans to reflect this setting to Flight Server as well so there is single configuration.

### DIFF
--- a/crates/flight_client/src/lib.rs
+++ b/crates/flight_client/src/lib.rs
@@ -244,6 +244,20 @@ impl FlightClient {
         self
     }
 
+    /// Overrides default maximum message size for encoding and decoding.
+    #[must_use]
+    pub fn with_max_message_size(
+        mut self,
+        max_encoding_message_size: usize,
+        max_decoding_message_size: usize,
+    ) -> Self {
+        self.flight_client = self
+            .flight_client
+            .max_encoding_message_size(max_encoding_message_size)
+            .max_decoding_message_size(max_decoding_message_size);
+        self
+    }
+
     /// Queries the flight service for the schema of the path.
     ///
     /// # Arguments
--- a/crates/runtime/src/dataconnector/spiceai.rs
+++ b/crates/runtime/src/dataconnector/spiceai.rs
@@ -72,6 +72,14 @@ pub enum Error {
         value: Arc<str>,
         source: InvalidMetadataValue,
     },
+
+    #[snafu(display(
+        "Failed to apply parameter '{parameter}': {source}. Ensure the value is valid and retry.\nFor details, visit: https://spiceai.org/docs/components/data-connectors/spiceai#parameters"
+    ))]
+    InvalidParameterValue {
+        parameter: String,
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
 }
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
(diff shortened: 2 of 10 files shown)


crates/flight_client/src/lib.rs
```diff
@@ -244,6 +244,20 @@ impl FlightClient {
         self
     }
 
+    /// Overrides default maximum message size for encoding and decoding.
+    #[must_use]
+    pub fn with_max_message_size(
```
sgrebnov, line 249: We use single configuration, made I made it separate values for setting flexibility

crates/spicepod/src/component/runtime.rs
```diff
@@ -56,6 +56,9 @@ pub struct Runtime {
     #[serde(default, skip_serializing_if = "is_default")]
     pub cors: CorsConfig,
 
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub flight: Option<Flight>,
```
sgrebnov, line 60: Alternatively we can make it non-optional to simplify usage, current version is used to align with similar properties implementation, like `auth`

crates/runtime/src/init/dataset.rs
```diff
@@ -197,6 +197,7 @@ impl Runtime {
 
         let source = ds.source();
         let params = ConnectorParamsBuilder::new(source.into(), (&ds).into())
+            .with_app(Arc::clone(&ds.app))
```
ewgenius: I don't think it's necessary to pass the app explicitly here - it should be possibel to get app reference from dataset inside `ConnectorParamsBuilder.build` method from dataset component, and pass it down to the connector params.

ewgenius:
```rs
let (params, prefix, parameters, app) = match &self.component {
    ConnectorComponent::Catalog(catalog) => {
        // ...

        (
            get_params_with_secrets(Arc::clone(&secrets), &catalog.params).await,
            factory.prefix(),
            factory.parameters(),
            None,
        )
    }
    ConnectorComponent::Dataset(dataset) => {
        // ...

        (
            params,
            factory.prefix(),
            factory.parameters(),
            Some(Arc::clone(&dataset.app)),
        )
    }
};
```

crates/runtime/src/dataconnector/spiceai.rs
ewgenius, line 1: Note - `SpiceAIFactory` also used by Spice AI Cloud catalog connector, and in theory it might work out of the box, but it is worth checking.

crates/runtime/src/init/dataset.rs
sgrebnov: Updated, thank you! Next iteration will be adding app property to the Catalog (similar to Datasets) and changing from `pub(crate) app: Option<Arc<App>>` to `pub(crate) app: Arc<App>` to simplify usage.  Will keep this change for the next iteration/PR to simplify review as this will require changes in Catalog creation/parsing logic (using builder pattern similar to Datasets)

crates/runtime/src/dataconnector/spiceai.rs
sgrebnov: Good point, thx!! This will be done next - 
https://github.com/spiceai/spiceai/pull/5407#discussion_r2047543763

lukekim (approved):

PR #5409, opened 2025-04-16
Track used_tools in ai_inferences_with_spice_count as number

ewgenius:
# 🗣 Description

Track `used_tools` in `ai_inferences_with_spice_count` as number

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/runtime/src/model/model_context.rs
+++ b/crates/runtime/src/model/model_context.rs
@@ -16,7 +16,7 @@ limitations under the License.
 
 use std::sync::{
     Arc,
-    atomic::{AtomicBool, Ordering},
+    atomic::{AtomicU16, Ordering},
 };
 
 use axum::body::Body;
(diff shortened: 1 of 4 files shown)


phillipleblanc (dismissed):

Sevenannn (dismissed):

crates/runtime/src/model/model_context.rs
```diff
@@ -110,7 +110,7 @@ pub fn track_ai_inferences_count(context: &Arc<RequestContext>) {
     if let Some(model_context) = context.extension::<ModelContextExtension>() {
```
Sevenannn, line 110: The comments of this function need to be updated to reflect change from bool -> number

ewgenius: thanks, fixed

Sevenannn (approved):

PR #5413, opened 2025-04-17
fix: Testoperator uses an exact API key for benchmark metric submission

peasee:
# 🗣 Description

* Adds credential variant `Exact` for the flight client, allowing exact control over the `authorization` header
* Uses the `Credentials::Exact` variant for submitting benchmark metrics, as it uses a static key
* Creates a new `Token` to wrap a resultant token from a handshake exchange. This also updates the token to be stored as a `SecretString` instead of a `String`, to improve security of the flight client.
* Updates the expected benchmark API key path to `SPICEAI_BENCHMARK_METRICS_KEY`

### DIFF
--- a/crates/flight_client/src/lib.rs
+++ b/crates/flight_client/src/lib.rs
@@ -15,6 +15,7 @@ limitations under the License.
 */
 
 use std::borrow::Cow;
+use std::fmt::Display;
 use std::sync::Arc;
 use std::task::Poll;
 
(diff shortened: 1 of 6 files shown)


crates/flight_client/src/lib.rs
```diff
@@ -175,11 +176,43 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
...
         username: Arc<str>,
         password: Arc<SecretString>,
     },
+    // Anonymous access
     Anonymous,
+    // An explicitly exact token
+    Exact {
+        token: Arc<SecretString>,
+        bearer: bool, // whether this token is a Bearer token, or if it is set to the 'authorization' header verbatim
+    },
```
phillipleblanc:
```suggestion
    Bearer {
        token: Arc<SecretString>,
        prefix: bool, // whether this token requires the 'Bearer ' prefix, or if it is set to the 'authorization' header verbatim
    },
```

crates/test-framework/src/metrics/mod.rs
```diff
@@ -671,12 +671,12 @@ pub trait MetricCollector<T: ExtendedMetrics, R: ExtendedMetrics> {
             started_at: usize::try_from(
                 self.start_time()
                     .duration_since(SystemTime::UNIX_EPOCH)?
-                    .as_secs(),
+                    .as_millis(),
```
Jeadie, line 674: Why switching seconds -> milliseconds?

peasee: https://github.com/spiceai/spiceai/blob/trunk/docs/dev/metrics.md#guidelines

phillipleblanc (approved):

PR #5420, opened 2025-04-17
Parameterized queries via Flight SQL API

### LINKED ISSUE #5379: Support parameterized queries via Flight SQL API
## Goal-State/What/Result

Parametrized queries are supported via the Flight SQL API.

## Why/Purpose

Parameterized queries are a standard practice for supplying input parameters to queries. We should support the use of this standard practice.

## By When

**Issue/Spec written and reviewed:** 4/14/25
**Done-Done:** 4/15/25

## Done-Done

- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Tests
- [ ] Documentation
- [ ] Release Notes prepared

## The Algorithm

- [x] Simplify.

## Specification

The runtime will be enhanced to support prepared statements, enabling clients to create, bind parameters to, execute, and close parameterized SQL queries efficiently and securely. This feature will improve query performance by reusing plans, enhance security by preventing SQL injection, and ensure compatibility with Flight SQL clients.

## How/Implementation Plan

Follow [example from datafusion](https://github.com/datafusion-contrib/datafusion-flight-sql-server/blob/main/datafusion-flight-sql-server/src/service.rs)

## QA Plan

Test via:

```python
from adbc_driver_flightsql import DatabaseOptions
from adbc_driver_flightsql.dbapi import connect

with connect("grpc://127.0.0.1:50051") as conn, conn.cursor() as cur:
    cur.execute("SELECT $1 + 1 AS the_answer", parameters=(41,))
    table = cur.fetch_arrow_table()
    assert table["the_answer"][0].as_py() == 42, "Expected the_answer to be 42"
    conn.close()
```

Ensure that existing authentication mechanisms continue to work properly.

kczimm:
# 🗣 Description

Adds support for parameterized queries via the Flight SQL API.

## 🔨 Related Issues

#5379 

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

We are caching the prepared statements because Flight SQL involves multiple round trip messages to create the prepared statement, bind parameters, and execute it. Should we remove the prepared statement from the cache after execution? I'm not sure.


kczimm:
Tested using

```python
from adbc_driver_flightsql import DatabaseOptions
from adbc_driver_flightsql.dbapi import connect

with connect(
    "grpc://127.0.0.1:50051",
    db_kwargs={
        DatabaseOptions.AUTHORIZATION_HEADER.value: "Bearer 123",
        DatabaseOptions.TLS_SKIP_VERIFY.value: "true",
    }
) as conn:
    with conn.cursor() as cur:
        cur.execute("SELECT $1 + 1 AS the_answer", parameters=(41,))
        table = cur.fetch_arrow_table()
        print(table)
        cur.execute("SELECT $1, $2", parameters=(1, 2))
        table = cur.fetch_arrow_table()
        print(table)
        cur.execute("SELECT 1 + 1 AS the_answer")
        table = cur.fetch_arrow_table()
        print(table)

    conn.close()
 ```
 
 ```shell
 pyarrow.Table
the_answer: int64 not null
----
the_answer: [[42]]
pyarrow.Table
$1: int64 not null
$2: int64 not null
----
$1: [[1]]
$2: [[2]]
pyarrow.Table
the_answer: int64 not null
----
the_answer: [[2]]
 ```

**Comment by kczimm** [comment]
I'll test with more variations of prepared statements.

**Comment by kczimm** [comment]
I'll test with more variations of prepared statements.

File: crates/arrow_tools/src/schema.rs [hunk 1]
```diff
@@ -67,6 +67,10 @@ pub fn verify_schema(
         let a_data_type = a.data_type();
         let b_data_type = b.data_type();
 
+        if is_null_placeholder(a) || is_null_placeholder(b) {
```
phillipleblanc, line 73: nit: Add a comment why we skip verifying the field for a null placeholder

crates/runtime/src/datafusion/param_utils.rs
```diff
@@ -0,0 +1,269 @@
...
+///
+/// # Supported JSON Formats
+/// - Array: Converted to positional parameters (ParamValues::Positional).
+/// - Object: Converted to named parameters (ParamValues::Named).
+/// - Primitive types (null, bool, number, string) are supported within arrays or objects.
+///
+/// # Errors
+/// - Returns an error if the top-level JSON is not an array or object.
+/// - Returns an error if any value cannot be converted to a ScalarValue.
+pub fn convert_json_to_param_values(json: Value) -> Result<ParamValues, String> {
```
phillipleblanc: This should be a Snafu error type instead of a String error type. ref: https://github.com/spiceai/spiceai/blob/trunk/docs/dev/style_guide.md#all-errors-use-snafu-functionality

crates/runtime/src/datafusion/param_utils.rs
```diff
@@ -0,0 +1,269 @@
...
+/// # Returns
+/// * `Result<ScalarValue, String>` - The converted ScalarValue or an error message.
+///
+/// # Supported JSON Types
+/// - Null -> ScalarValue::Utf8(None)
+/// - Bool -> ScalarValue::Boolean(Some(bool))
+/// - Number -> ScalarValue::Int64(Some(i64)) if integer, else ScalarValue::Float64(Some(f64))
+/// - String -> ScalarValue::Utf8(Some(String))
+/// - Arrays and objects return an error (handled at a higher level).
+fn json_to_scalar(json: &Value) -> Result<ScalarValue, String> {
```
phillipleblanc: This should be a Snafu error type instead of a String error type

crates/runtime/src/flight/flightsql/prepared_statement_query.rs
```diff
@@ -86,8 +110,6 @@ pub(crate) async fn get_flight_info(
 
     let info = FlightInfo::new()
         .with_endpoint(endpoint)
-        .try_with_schema(&arrow_schema)
-        .map_err(to_tonic_err)?
         .with_descriptor(fd);
```
phillipleblanc, line 97: Why was this removed?

crates/runtime/src/flight/flightsql/prepared_statement_query.rs
```diff
@@ -102,19 +124,153 @@ pub(crate) async fn do_get(
...
+        Box::pin(Service::sql_to_flight_stream(datafusion, &sql, parameters)).await?;
+    let timed_output = TimedStream::new(output, move || start);
+
+    let mut response =
+        Response::new(Box::pin(timed_output) as <Service as FlightService>::DoGetStream);
+    attach_cache_metadata(&mut response, from_cache);
+    Ok(response)
+}
+
+pub(crate) async fn do_put_query(
```
phillipleblanc, line 132: nit: Add a comment that explains this is what "binds" the parameters to the prepared statement.

crates/runtime/src/flight/flightsql/prepared_statement_query.rs
```diff
@@ -102,19 +124,153 @@ pub(crate) async fn do_get(
...
+    if total_rows > 1 {
+        return Err(Status::invalid_argument(
+            "parameters should contain a single row",
+        ));
+    }
+
+    let parameters = if parameters.is_empty() {
+        None
+    } else {
+        Some(parameters.as_bytes())
```
phillipleblanc: nit: Why not call decode_param_values in here and have decode_param_values be:

fn decode_param_values(
    parameters: &[u8],
)

instead of 

fn decode_param_values(
    parameters: Option<&[u8]>,
)

crates/runtime/src/flight/flightsql/prepared_statement_query.rs
```diff
@@ -102,19 +124,153 @@ pub(crate) async fn do_get(
...
+            let schema = decoder.schema();
+            let batches = decoder.into_iter().collect::<Result<Vec<_>, _>>()?;
+            let batch = concat_batches(&schema, batches.iter())?;
+            Ok(record_to_param_values(&batch)?)
+        })
+        .transpose()
+}
+
+// Converts a record batch with a single row into ParamValues
+fn record_to_param_values(batch: &RecordBatch) -> Result<ParamValues, DataFusionError> {
```
phillipleblanc: This seems like a useful utility to have beyond the flightsql code - perhaps we can move it to the arrow_tools crate? This function could also benefit from some tests.

crates/runtime/src/flight/flightsql/prepared_statement_query.rs
```diff
@@ -102,19 +124,153 @@ pub(crate) async fn do_get(
...
+        String::from_utf8(query.prepared_statement_handle.to_vec()).map_err(error_to_status)?;
+
+    flight_svc
+        .prepared_statements
+        .write()
+        .await
+        .entry(handle)
+        .and_modify(|stmt| stmt.parameters = parameters);
+
+    Ok(Response::new(Box::pin(empty())))
```
phillipleblanc: I don't think making the FlightSQL server stateful is the correct approach. The `handle` here is the parameterized SQL query, i.e. `SELECT * FROM table WHERE foo = $1`.

If multiple concurrent requests come in with the exact same parameterized query, but bind to different parameters - then we would return incorrect results. The first query would come in and bind to the handle `SELECT * FROM table WHERE foo = $1` and then the next query comes in right after and overwrites the parameters for that handle with new parameters. When the first query calls `DoGet` - it is getting the results from the second query, not its own.

The `handle` we return is an arbitrary sequence of bytes - we can use that to encode the `PreparedStatement` struct you defined and are currently storing in `flight_svc.prepared_statements`. The advantage of that approach is now the server becomes stateless again - the client now has the state of its "bound query" in its `handle`.

So in the initial `do_action_create_prepared_statement` we return a `PreparedStatement` with the parameters as None (as is currently implemented). Then when this `do_put_query` is called to bind the parameters, we can return an updated handle with the `PreparedStatement` filled in with the correct values. And finally in the `do_get`, we just deserialize the handle into a `PreparedStatement` and execute with that.

This is also mentioned in the [FlightSQL spec](https://arrow.apache.org/docs/format/FlightSql.html) as a valid approach:

> CommandPreparedStatementQuery
> Execute a previously created prepared statement and get the results.
>
> When used with DoPut: binds parameter values to the prepared statement. The server may optionally provide an updated handle in the response. Updating the handle allows the client to supply all state required to execute the query in an ActionPreparedStatementExecute message. For example, stateless servers can encode the bound parameter values into the new handle, and the client will send that new handle with parameters back to the server.
>
> Note that a handle returned from a DoPut call with CommandPreparedStatementQuery can itself be passed to a subsequent DoPut call with CommandPreparedStatementQuery to bind a new set of parameters. The subsequent call itself may return an updated handle which again should be used for subsequent requests.

crates/runtime/src/flight/flightsql/prepared_statement_query.rs
kczimm: I noticed that the schema at this point says the field is nullable. The schema returned in the `DoGet` has the field as not nullable. I tested the same client code against DuckDB on the particular query and it returned not nullable. So I concluded that this schema here is incorrect and found that removing it fixed the issue. Do you think we need to fix the schema calculation here and leave this part in?

```shell
adbc_driver_manager.OperationalError: UNKNOWN: [FlightSQL] endpoint 0 returned inconsistent schema: expected schema:
  fields: 1
    - the_answer: type=int64, nullable but got schema:
  fields: 1
    - the_answer: type=int64
 ```

File: crates/runtime/src/flight/flightsql/prepared_statement_query.rs (hunk 8 above)
Comment (kczimm) [reply crates/runtime/src/flight/flightsql/prepared_statement_query.rs]: So here I think I run into a consequence of implementing the `FlightService` trait instead of the `FlightSqlService` trait. Specifically, the `FlightService::do_put` method returns a `Self::DoPutStream` which is a different type than what I need to return from the `do_put_query` function here if I return an updated handle to make the server stateless. I need to instead return a `DoPutPreparedStatementResult` per the spec. Unless I'm missing something this is just a limitation of the `FlightService` trait.

Comment (kczimm) [reply crates/runtime/src/flight/flightsql/prepared_statement_query.rs]: If I try to return the serialized prepared statement in the bytes of the `PutResult` message instead of the `DoPutPreparedStatementResult` message, I get an `invalid wire-format data` message, which makes sense.

File: crates/runtime/src/datafusion/param_utils.rs [hunk 9]
```diff
@@ -0,0 +1,269 @@
...
+/// # Errors
+/// - Returns an error if the top-level JSON is not an array or object.
+/// - Returns an error if any value cannot be converted to a ScalarValue.
+pub fn convert_json_to_param_values(json: Value) -> Result<ParamValues, String> {
+    match json {
+        Value::Array(arr) => {
+            let mut vec = Vec::with_capacity(arr.len());
+            for (i, val) in arr.into_iter().enumerate() {
+                let scalar = json_to_scalar(&val).map_err(|e| {
+                    format!("failed to convert array element at index {}: {}", i, e)
```
sgrebnov:
```suggestion
                    format!("failed to convert array element at index {i}: {e}")
```

crates/runtime/src/datafusion/param_utils.rs
```diff
@@ -0,0 +1,269 @@
...
+                })?;
+                vec.push(scalar);
+            }
+            Ok(ParamValues::List(vec))
+        }
+        Value::Object(obj) => {
+            let mut map = HashMap::new();
+            for (key, val) in obj {
+                let scalar = json_to_scalar(&val)
+                    .map_err(|e| format!("failed to convert value for key '{}': {}", key, e))?;
```
sgrebnov:
```suggestion
                    .map_err(|e| format!("failed to convert value for key '{key}': {e}"))?;
```

crates/runtime/src/datafusion/param_utils.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use datafusion::scalar::ScalarValue;
+    use serde_json::json;
+    use std::collections::HashMap;
+
+    fn scalar_to_string(scalar: &ScalarValue) -> String {
```
sgrebnov: I think it could be just  `scalar.to_string()`, no need for manual conversion to string

crates/runtime/src/datafusion/param_utils.rs
```diff
@@ -0,0 +1,269 @@
...
+                    .iter()
+                    .map(|(k, v)| format!("\"{}\": {}", k, scalar_to_string(v)))
+                    .collect();
+                items.sort(); // Ensure consistent ordering
+                format!("Map({{{}}})", items.join(", "))
+            }
+        }
+    }
+
+    fn assert_eq_param_values(a: ParamValues, b: ParamValues) {
```
sgrebnov: We can avoid string manipulations / using custom serialization logic to compare ParamValues
```
fn assert_eq_param_values(a: ParamValues, b: ParamValues) {
        match (&a, &b) {
            (ParamValues::Map(map_a), ParamValues::Map(map_b)) => {
                assert_eq!(map_a, map_b);
            }
            (ParamValues::List(vec_a), ParamValues::List(vec_b)) => {
                assert_eq!(vec_a, vec_b);
            }
            _ => {
                panic!(
                    "ParamValues are different types: {a:?} and {b:?}"
                );
            }
        }
    }
```

crates/runtime/src/flight/do_put.rs
```diff
@@ -69,16 +73,27 @@ pub(crate) async fn handle(
...
         return Err(Status::invalid_argument("No flight descriptor provided"));
     };
+
+    if let Ok(message) = Any::decode(&*fd.cmd) {
+        match Command::try_from(message).map_err(|e| Status::internal(format!("{e:?}")))? {
+            Command::CommandPreparedStatementQuery(query) => {
+                return prepared_statement_query::do_put_query(flight_svc, query, streaming_flight)
+                    .await;
+            }
+            c => return Err(Status::unimplemented(&format!("{c:?}"))),
```
sgrebnov: Please add a note why we always return error here and not executing `do_put` further

crates/runtime/src/flight/flightsql/prepared_statement_query.rs
```diff
@@ -102,19 +124,153 @@ pub(crate) async fn do_get(
...
+    attach_cache_metadata(&mut response, from_cache);
+    Ok(response)
+}
+
+pub(crate) async fn do_put_query(
+    flight_svc: &Service,
+    query: CommandPreparedStatementQuery,
+    streaming_flight: Peekable<Streaming<FlightData>>,
+) -> Result<Response<<Service as FlightService>::DoPutStream>, Status> {
+    let streaming_flight = streaming_flight
```
sgrebnov, line 137: Let's add metrics::track_flight_request

crates/runtime/src/datafusion/param_utils.rs
```diff
@@ -0,0 +1,277 @@
...
+/// - Bool -> ScalarValue::Boolean(Some(bool))
+/// - Number -> ScalarValue::Int64(Some(i64)) if integer, else ScalarValue::Float64(Some(f64))
+/// - String -> ScalarValue::Utf8(Some(String))
+/// - Arrays and objects return an error (handled at a higher level).
+fn json_to_scalar(json: &Value) -> Result<ScalarValue, Error> {
+    match json {
+        Value::Null => Ok(ScalarValue::Utf8(None)),
+        Value::Bool(b) => Ok(ScalarValue::Boolean(Some(*b))),
+        Value::Number(n) => {
+            if let Some(i) = n.as_i64() {
```
sgrebnov: should we try `if let Some(u) = n.as_u64()` first?

https://github.com/influxdata/influxdb3_core/blob/main/iox_query_params/src/params.rs#L539

kczimm: This is a good find, it appears they are basically doing the same thing. What is the advantage of doing `if let Some(u) = n.as_u64()` first? Is it the most probable data type and can short circuit evaluation in most cases?

crates/runtime/src/flight/flightsql/prepared_statement_query.rs
phillipleblanc: We should understand why the schema is different. Also if we're not going to return the schema here, then we don't need to calculate it above.

crates/runtime/src/flight/flightsql/prepared_statement_query.rs
phillipleblanc: That doesn't make sense to me - The FlightSqlService trait must eventually return a DoPutStream, because that is what the underlying protocol needs. How does the FlightSqlService implementation handle that?

kczimm: > The FlightSqlService trait must eventually return a DoPutStream

I don't think this is true. See https://arrow.apache.org/docs/format/FlightSql.html#id3. The client sends a `DoPut(CommandPreparedStatementQuery)` and the server should respond with `DoPutPreparedStatementResult{handle}`. Prepared statements are only a thing in the extended Flight SQL protocol. The `FlightService` trait only allows you to return `DoPutStream` from `do_put`, as you say, https://docs.rs/arrow-flight/latest/arrow_flight/flight_service_server/trait.FlightService.html#tymethod.do_put. I got away with it in the stateful server because the return message is optional, and I chose not to send anything back.

phillipleblanc: The FlightSqlService is just syntactic sugar over the FlightSql server: https://github.com/apache/arrow-rs/blob/main/arrow-flight/src/sql/server.rs#L569

phillipleblanc: It looks like this is how they are transforming it to a DoPutStream:
https://github.com/apache/arrow-rs/blob/main/arrow-flight/src/sql/server.rs#L734C13-L742C14

```rust
Command::CommandPreparedStatementQuery(command) => {
                let result = self
                    .do_put_prepared_statement_query(command, request)
                    .await?;
                let output = futures::stream::iter(vec![Ok(PutResult {
                    app_metadata: result.encode_to_vec().into(),
                })]);
                Ok(Response::new(Box::pin(output)))
            }
```

kczimm: I stand corrected. Great find.

crates/runtime/src/datafusion/param_utils.rs
sgrebnov: I think the difference will be in how positive numbers will be treated - I believe (based on current code) we will parse them always as Int64 and InfluxDB as UInt64? Honestly I don't know the real impact/what should be the right behavior but my assumption is that Influx probably know what they are doing

kczimm:
Testing with JDBI/JDBC in Java/Scala is held up with `?` versus `$1` placeholder syntax. I think this PR is still valid though. @phillipleblanc?

crates/arrow_tools/src/schema.rs
```diff
@@ -80,6 +87,10 @@ pub fn verify_schema(
     Ok(())
 }
 
+fn is_null_placeholder(field: &Arc<Field>) -> bool {
+    field.name().starts_with('$') && field.data_type() == &DataType::Null
```
phillipleblanc: Should this also include `?`

crates/arrow_tools/src/record_batch.rs
```diff
@@ -208,6 +209,47 @@ pub fn truncate_numeric_column_length(
...
+
+    let mut is_list = true;
+    for col_index in 0..batch.num_columns() {
+        let array = batch.column(col_index);
+        let scalar = ScalarValue::try_from_array(array, 0)?;
+        let name = batch
+            .schema_ref()
+            .field(col_index)
+            .name()
+            .trim_start_matches('$')
```
phillipleblanc, line 227: Should this also include `?`

crates/runtime/src/flight/flightsql/prepared_statement_query.rs
```diff
@@ -46,37 +60,30 @@ pub(crate) async fn do_action_create_prepared_statement(
 
     let schema_bytes = Service::serialize_schema(&arrow_schema)?;
 
+    let stmt = PreparedStatement {
+        query: statement.query.clone(),
+        parameters: vec![],
+    };
+
+    let handle = serde_json::to_vec(&stmt).map_err(error_to_status)?;
```
phillipleblanc: Using JSON for this is inefficient - especially since the binary parameter data will need to be base64 encoded. Could we try with https://docs.rs/postcard as the serialization protocol? It should be more efficient than JSON.

phillipleblanc (approved):

PR #5423, opened 2025-04-17
Enable task name override for non-static span names; add dynamic MCP tool task names

Jeadie:
# 🗣 Description
 - `tracing::Span` require `&'static str` for the span name. For task history, the span name is used as the task name. 
 - This PR enables span attributes, specifically `task_override` to be used as the task name if it is present.
 - This is useful for MCP tools where we don't know the task name at compile time. This PR also updates `tool_use::mcp` to `tool_use::{mcp server spicepod name}/{mcp tool name}` (e.g. `tool_use::fs/list_directories`). 
 
 ### Example
 ```
>>> spice trace ai_chat
TREE                   STATUS DURATION   TASK
f068afe967857377       ✅     11899.12ms ai_chat
  ├── 6e362bd54c5af12e ✅     11896.64ms ai_completion
  ├── aa233ab5ad58ceb0 ✅         6.54ms tool_use::fs/list_directory
  └── 63d7d7b3b14ef970 ✅     10304.54ms ai_completion
```

## 📄 Documentation Updates
 - [ ] Update docs for the presence of `tool_use::*` task names (from MCP tools)

### DIFF
--- a/crates/runtime/src/tools/mcp/tool.rs
+++ b/crates/runtime/src/tools/mcp/tool.rs
@@ -31,11 +31,22 @@ use super::{McpProxy, Result};
 pub struct McpToolWrapper {
     client: Arc<RwLock<Box<dyn McpClientTrait>>>,
     spec: McpTool,
+
+    /// Spicepod defined name, not from underlying MCP.
+    server_name: String,
 }
 
 impl McpToolWrapper {
-    pub fn new(client: Arc<RwLock<Box<dyn McpClientTrait>>>, spec: McpTool) -> Self {
-        Self { client, spec }
+    pub fn new(
+        client: Arc<RwLock<Box<dyn McpClientTrait>>>,
+        spec: McpTool,
+        server_name: String,
+    ) -> Self {
+        Self {
+            client,
+            spec,
+            server_name,
+        }
     }
 
     #[must_use]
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/runtime/src/tools/mcp/tool.rs [hunk 1]
```diff
@@ -68,6 +79,10 @@ impl SpiceModelTool for McpToolWrapper {
         _rt: Arc<Runtime>,
     ) -> Result<Value, Box<dyn std::error::Error + Send + Sync>> {
         let span: Span = tracing::span!(target: "task_history", tracing::Level::INFO, "tool_use::mcp", tool = self.name().to_string(), input = arg);
+        span.in_scope(
+            || tracing::info!(target: "task_history", task_override = %format!("tool_use::{}/{}", self.server_name, self.spec.name), "labels"),
```
lukekim, line 83: I wonder if this needs to be `{}-{}`, to be consistent with VS Code tool names.

Jeadie: The main reason that `/` was decided (more broadly than just the task name used in observability) was to 
1. Allow UX to still use `-` in either MCP name, or the underlying tools, without confusion. 
2. Natural fit for `v1/tools/{tool_name}` (e.g. `v1/tools/mcp_server/underlying_tool`).  

The use of `/` does cause some issues, like you mentioned, consistency with VSCode. Also, some model providers (e.g. OpenAI) disallow `/` and these must be [mapped](https://github.com/spiceai/spiceai/blob/aa2b7299ff3c9e44c4f0677016e54958c4ed0815/crates/runtime/src/model/tool_use.rs#L716). 

In saying this `-`, although often allowed, is not commonly used.

Jeadie: We're going to merge this PR with current semantics. If we change `/` will involve separate PR.

sgrebnov (approved):

PR #5427, opened 2025-04-18
Handle second Ctrl-C to force runtime termination

### LINKED ISSUE #5396: Implement force termination support via second Ctrl-C

sgrebnov:
# 🗣 Description

Implements logic to cancel runtime grace termination in case of user pressed Ctrl-C to force terminate the runtime.

Note: force termination is considered dev mode currently and triggered by Ctrl-C only.

## 🔨 Related Issues

Closes https://github.com/spiceai/spiceai/issues/5396

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

Current implementation only cancel runtime grace termination and does not guarantee immediate termination. Alternative approach could be moving handler to spiced and terminate the process.

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -23,6 +23,7 @@ use std::time::Duration;
 use std::{collections::HashMap, sync::Arc};
 use tokio::task::JoinHandle;
 use tokio::time::Instant;
+use util::force_shutdown_signal;
 
 use crate::dataaccelerator::AcceleratorEngineRegistry;
 use crate::{
@@ -314,6 +315,9 @@ pub enum Error {
 
     #[snafu(display("Initialization has been cancelled"))]
     ComponentsInitializationCancelled,
+
+    #[snafu(display("Force shutdown requested"))]
+    ForceTerminated,
 }
 
 const HTTP_SERVER: &str = "http_server";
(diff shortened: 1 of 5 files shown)


crates/util/Cargo.toml
```diff
@@ -14,3 +14,4 @@ humantime = "2.1.0"
 rand = "0.9.0"
 tokio = { workspace = true }
 tracing = { workspace = true }
+ctrlc = "3.4"
```
sgrebnov, line 17: Similar to [this comment](https://github.com/spiceai/spiceai/pull/4885/files#r1972830562)
1. I found this to work better to listen for the second Ctrl-C signal. Existing solution still reports another shutdown signal even if I use it after substantial delay like 1s. This is related to how systems signal queues are implemented, I found that the channel must be drained otherwise could still be termination signal (tried `shutdown_signal` and `tokio::signal::ctrl_c().await`)
1. Taking into account this is more of dev mode, this also allows listening only for Ctrl-C and avoid forcing shutdown on other signals.

Sevenannn (dismissed):

crates/runtime/src/lib.rs
```diff
@@ -546,10 +550,20 @@ impl Runtime {
 
         // Shutdown signal
         let shutdown_signal_future = async {
-            shutdown_signal().await;
-            tracing::debug!("Shutdown signal received.");
-            self.shutdown().await;
-            Ok(())
+            let graceful_shutdown = async {
+                shutdown_signal().await;
+                tracing::debug!("Shutdown signal received.");
```
ewgenius: Maybe inform about force shutdown option?

```suggestion
                tracing::debug!("Shutdown signal received. Press Ctrl-C again to force exit.");
```

ewgenius (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #5449, opened 2025-04-21
Enable parallel data sampling for NSQL

### LINKED ISSUE #5440: Bug: NSQL sample queries must be concurrent
## Describe the bug

When sample data is enabled for nsql request, 
1- distinct sample queries run sequentially for each column
2- sampling for each table is done  sequentially

resulting in very slow execution.

sgrebnov:
# 🗣 Description

1. Run data sampling tools (DistinctColumns, RandomSample) in parallel (max 10 concurrent tool calls).
1. Run separate queries (column data) of DistinctColumns tool in parallel (max 5 columns in parallel)

## 🔨 Related Issues

Fixes https://github.com/spiceai/spiceai/issues/5440

## ⛓️‍💥 Breaking Change

### DIFF
--- a/crates/runtime/src/http/v1/nsql.rs
+++ b/crates/runtime/src/http/v1/nsql.rs
@@ -38,6 +38,7 @@ use axum::{
 };
 use axum_extra::TypedHeader;
 use datafusion::sql::TableReference;
+use futures::future::try_join_all;
 use headers_accept::Accept;
 
 use itertools::Itertools;
@@ -52,6 +53,9 @@ use super::accept_header_types;
 
 // Default number of retries for NSQL queries if the generated query fails to execute
 const DEFAULT_NSQL_RETRIES: u8 = 10;
+
+// Maximum number of parallel tables for sampling
+const DATA_SAMPLING_MAX_PARALLEL_TABLES: usize = 5;
 
 fn clean_model_based_sql(input: &str) -> String {
     let no_dashes = match input.strip_prefix("--") {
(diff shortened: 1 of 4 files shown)


crates/runtime/src/http/v1/nsql.rs
```diff
@@ -76,29 +80,40 @@ async fn sample_messages(
...
-                Arc::clone(&rt),
-                &SampleDataTool::new(method.clone()),
-                format!("sample-{method:?}").as_str(),
-                &params,
-            )
-            .instrument(Span::current())
-            .await?;
-            messages.extend(msgs);
-        }
+    for datasets_chunk in sample_from.chunks(DATA_SAMPLING_MAX_PARALLEL_TABLES) {
```
phillipleblanc: I think it would be better to use the same mechanism we use for parallel dataset loading. This implementation will do 5 requests in parallel, but then it waits for all 5 of those to finish before moving onto the next 5. Ideally as soon as one completes we can start processing the next one so that we always have 5 in progress. This design would wait for the slowest task in the chunk to finish before moving on.

That code is roughly:

```rust
let semaphore = Arc::new(Semaphore::new(parallel_num));

let mut dataset_futures = HashMap::new();
let future: Pin<Box<dyn Future<Output = ()> + Send>> =
                Box::pin(async move { cloned_self.load_dataset(ds_clone).await })
                    as Pin<Box<dyn Future<Output = ()> + Send>>;
            dataset_futures.insert(ds.name.clone(), future);

for (ds, dataset_load_future) in dataset_futures {
            let semaphore = Arc::clone(&semaphore);
            let handle = tokio::spawn(async move {
                let Ok(_guard) = semaphore.acquire().await else {
                    unreachable!("Semaphore is never closed.");
                };
                tracing::info!("Initializing dataset {ds}");
                dataset_load_future.await;
            });
            spawned_tasks.push(handle);
        }

        let _ = join_all(spawned_tasks).await;
```

phillipleblanc: Another simpler mechanism would be to turn each of the futures into a stream and use `buffer_unordered` to start working on each item in parallel and accomplishes the same goal. https://docs.rs/futures/latest/futures/stream/struct.BufferUnordered.html

sgrebnov: I'll update - thank you

phillipleblanc: futures::stream::iter(futures)
            .buffer_unordered(max_concurrent)
            .try_collect::<Vec<_>>()
            .await?;

sgrebnov: The only concern with unordered sampling results is snapshot testing we use, will review

sgrebnov: Updated, thank you

phillipleblanc (approved):

PR #5463, opened 2025-04-21
Add support for `?` placeholder syntax in parameterized queries

### LINKED ISSUE #5453: Parameterized query placeholder `?` not supported
## Goal-State/What/Result

Support `?` placeholder syntax for parameterized queries.

## Why/Purpose

Parameterized queries using the `Postgres` syntax `$1`, `$2`, etc. are supported with #5420 but not `?` syntax. The flight-sql-jdbc-driver uses `?` syntax for positional args and converts named args to positional args using `?`. This means that parameterized queries are ultimatley **not** supported using JDBC.

## By When

**Issue/Spec written and reviewed:** 4/21/25
**Done-Done:** 4/22/25

## Done-Done

- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Tests
- [ ] Documentation
- [ ] Release Notes prepared

## The Algorithm

- [x] Simplify.

## How/Implementation Plan

Convert the MySQL style query into a Postgres style query and parse as normal

kczimm:
# 🗣 Description

- Convert parameterized query into Postgres dialect.
- Caculate and return the parameter schema; assuming UInt64 for any placeholder datatype that DF could not infer

## 🔨 Related Issues

#5453 

## ⛓️‍💥 Breaking Change

## 🤔 Concerns

- Robustness of conversion. I've added many tests. Ideally we will parse the SQL and convert the placeholder names in the AST. We need to be able to run queries from the AST and not just from the raw SQL which warrants some refactoring.
- Assumption of UInt64 datatype for placeholder datatypes that could not be inferred is troublesome. This is a mitigation to https://github.com/apache/arrow-java/issues/153 where the datatypes of each parameter must be known explicitly and communicated to the JDBC driver.


crates/runtime/src/datafusion/query.rs
```diff
@@ -277,6 +282,28 @@ impl Query {
     }
 }
 
+fn parameter_schema_for_plan(plan: &LogicalPlan) -> Result<Option<Schema>, DataFusionError> {
+    let parameters = plan
+        .get_parameter_types()?
+        .into_iter()
+        .map(|(name, dt)| {
```
sgrebnov, line 290: Could you please add a note when this can happen and why it is safe to proceed with this assumption

crates/runtime/src/datafusion/query.rs
```diff
@@ -246,7 +248,7 @@ impl Query {
             .finish_with_error(request_context, error_message, error_code);
     }
 
-    pub async fn get_schema(self) -> Result<Schema, DataFusionError> {
+    pub async fn get_schema(self) -> Result<(Schema, Option<Schema>), DataFusionError> {
```
phillipleblanc, line 252: Add a comment explaining what the return types mean

crates/runtime/src/datafusion/query.rs
kczimm: I don't feel confident this is a great assumption in all cases. From the examples I've tested so far, this only occurs for LIMIT placeholders.

phillipleblanc (dismissed):

phillipleblanc (approved):

Issue #5416, opened 2025-04-17
Enhancement: Support `spice chat <message>` command


Sevenannn:
## Goal-State/What/Result

Enable one-shot chat interactions with Spice Cloud for the spice-qa workflow in spiceai/cookbook use case, including:

– Support `spice chat <message>` so that users can send a message and receive a synchronous response without opening an interactive REPL
– Add a `--temperature <value>` option to let users fine‑tune the model’s creativity and randomness on spice chat request.

After the enhancement. user can use spice chat CLI to obtain chat response from SCP in a synchronous manner, the example request look like below:

```
spice chat --api-key <my-scp-app-api-key> --endpoint <my-spice-endpoint> --model my_cool_model --temperature 0.1 "Hello I'm Alice"
Hello Alice what can I do for you
# Spice CLI quit
```

## Why/Purpose

To enable quick, one‑shot chat calls—ideal for scripts, CI/CD pipelines—by eliminating REPL startup and custom temperature settings.

## By When

**Issue/Spec written and reviewed:** April 17th 2025
**Done-Done:** April 21st 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

**Chat Request**
When <message> argument is supplied, the chat request body will be formulated with `"stream": true,`

**Chat Response Display**
User can receive chat response without entering a REPL. As soon as the OpenAI stream yields each delta, write it directly to STDOUT
```
spice chat message "hello I'm Alice"
(at time 1)H
(at time 2)He
(at time 3)Hey
```
Upon stream completion, exit normally (exit code 0).

**Model Selection**
If `--model <model>` is provided, its value is sent as <model>in the request.
If `--model <model>` is not provided
- Single Model Available: Automatically use that model.
- Multiple Models Available: Prompt the user with an interactive chooser:

**Error Handling**
[...]

PR #5476, opened 2025-04-22
Better user message for `spice chat --model`

Sevenannn:
## 📝 Summary

Update the user facing message when `--model <model> flag is supplied in `spice chat`

### When `<model>` is not configured in spicepod:
```
2025/04/28 11:03:55 ERROR model bar does not exist — configured models: openai, openai1, foo
```

### When `<model>` is configured in spicepod but not ready:
```
2025/04/28 10:56:57 ERROR model foo is not ready — try again when ready
```

Test CI is validate and `spice chat <message>` works as expected: https://github.com/spiceai/spiceai/actions/runs/14714639181

## 🔗 Related

- Part of https://github.com/spiceai/spiceai/issues/5416


bin/spice/cmd/chat.go
```diff
@@ -194,7 +195,9 @@ spice chat --model <model> "What is Spice.ai?"
 		}
 
 		availableModels := []string{}
+		configuredModels := []string{}
```
lukekim: Why do you need another slice when you have `models.Data` available?

Sevenannn: The configured models needs a way to be stored and used later in the error message `slog.Error(fmt.Sprintf("model %s is not ready — try again later or use an available model: %s",model, strings.Join(availableModels, ", “)))`. 

I can update to loop through the model.Id of models.Data to store unavailable models only when user supply a model that’s not configured in spicepod. Could you provide suggestion on a better approach?

phillipleblanc: You could replace it with something like this:

```go
found := false
for _, m := range models {
  if m.Id == model.Id {
	  found = true
	  break
  }
}

if !found {
  // Extract just the IDs for logging
  ids := make([]string, len(models))
  for i, m := range models {
	  ids[i] = m.Id
}

slog.Error(
	"model does not exist",
	"model_id", model.Id,
	"configured_models", strings.Join(ids, ", "),
)
os.Exit(1)
}
```

And then we can also remove the `golang.org/x/exp/slices` import

bin/spice/cmd/chat.go
```diff
@@ -225,17 +225,36 @@ spice chat --model <model> "What is Spice.ai?"
 			cmd.Printf("Using model: %s\n", selectedModel)
 			model = selectedModel
 		} else {
-			modelIsAvailable := false
-			for _, availableModel := range availableModels {
-				if availableModel == model {
-					modelIsAvailable = true
+			is_configured := false
```
lukekim: Poor name. 1 uses diffferent casing style. 2. what is configured???

bin/spice/cmd/chat.go
```diff
@@ -225,17 +225,36 @@ spice chat --model <model> "What is Spice.ai?"
...
+					ids[i] = m.Id
+				}
+
+				slog.Error(fmt.Sprintf("model %s does not exist — configured models: %s",
+					model, strings.Join(ids, ", ")))
+
+				os.Exit(1)
+			}
+
+			is_available := false
```
lukekim: Same comment.

bin/spice/cmd/chat.go
```diff
@@ -225,17 +225,36 @@ spice chat --model <model> "What is Spice.ai?"
...
+			for _, m := range availableModels {
+				if m == model {
+					is_available = true
+					break
+				}
+			}
+			if !is_available {
+				slog.Error(fmt.Sprintf("model %s is not ready — try again later or use an available model: %s",
```
lukekim:
```suggestion
				slog.Error(fmt.Sprintf("model %s is not ready — try again when model is ready: %s",
```

bin/spice/cmd/chat.go
lukekim: Use a consistent name. In this case it's testing for readiness, so use "ready" not "available" which can mean something else.

phillipleblanc (approved):

sgrebnov (approved):

PR #5481, opened 2025-04-22
Allow parameter queries with no parameters

kczimm:
## 📝 Summary

Allow parameter queries with no parameters.

### DIFF
--- a/crates/runtime/src/flight/flightsql/prepared_statement_query.rs
+++ b/crates/runtime/src/flight/flightsql/prepared_statement_query.rs
@@ -71,9 +71,11 @@ pub(crate) async fn do_action_create_prepared_statement(
             .map_err(to_tonic_err)?;
 
     let dataset_schema = Service::serialize_schema(&dataset_schema)?;
-    let parameter_schema = Service::serialize_schema(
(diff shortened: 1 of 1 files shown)


crates/runtime/src/flight/flightsql/prepared_statement_query.rs
```diff
@@ -71,9 +71,11 @@ pub(crate) async fn do_action_create_prepared_statement(
             .map_err(to_tonic_err)?;
 
     let dataset_schema = Service::serialize_schema(&dataset_schema)?;
-    let parameter_schema = Service::serialize_schema(
-        &parameter_schema.ok_or(Status::internal("no parameter schema"))?,
-    )?;
+    let parameter_schema = if let Some(schema) = &parameter_schema {
+        Service::serialize_schema(schema)?
+    } else {
+        Default::default()
```
phillipleblanc: I'm pretty sure this will be a lint failure - it needs to have the concrete type instead of `Default`, so it would be `MyStruct::default()` - that makes it more obvious what is actually being defaulted.

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #5487, opened 2025-04-22
Cache logical plans of SQL statements

### LINKED ISSUE #5460: Cache prepared statements
Parameterized queries are logically planned before parameters are bound. We can cache the logical plans to avoid the first stage of planning for prepared statements.

kczimm:
## 📝 Summary

Caches the logical plans of SQL statements to improve performance and reduce the time taken to create or get a logical plan from the query.

## 🔗 Related

#5460 

## 👀 Notes for Reviewers

- This is beneficial for all SQL queries not just prepared statements
- Do we need a mechanism to clear the cache like we do with the results cache? Obviously, this cache is substantially smaller in memory footprint but still might be something we want.


crates/runtime/src/datafusion.rs
```diff
@@ -1334,6 +1339,28 @@ impl DataFusion {
...
         }
     }
+
+    /// Create or get a logical plan from the query
+    async fn get_or_create_logical_plan(
+        &self,
+        session: &SessionState,
+        sql: &str,
+    ) -> Result<LogicalPlan, DataFusionError> {
+        let cached_plans = self.cached_plans.read().await;
```
lukekim: @kczimm This is in the critical query path. We should use a lockfree data structure if possible.

kczimm:
I'm working on tests

crates/runtime/src/datafusion.rs
```diff
@@ -1334,6 +1340,27 @@ impl DataFusion {
...
+    ) -> Result<LogicalPlan, DataFusionError> {
+        if let Some(plan) = self.cached_plans.get(sql).await {
+            tracing::trace!("using cached plan for {sql}");
+            return Ok(plan);
+        }
+
+        let plan = session.create_logical_plan(sql).await?;
+
+        tracing::trace!("caching plan for {sql}");
+        self.cached_plans
```
phillipleblanc: This is in the critical query path - cloning a potentially large SQL string and the plan itself is a large price to pay.

crates/runtime/src/datafusion/builder.rs
```diff
@@ -172,12 +173,15 @@ impl DataFusionBuilder {
 
         ctx.register_catalog(SPICE_DEFAULT_CATALOG, Arc::new(catalog));
 
+        const DEFAULT_CACHED_PLANS_MAX_CAPACITY: u64 = 100;
```
lukekim: Seems low... maybe 256 or 512?

crates/runtime/src/datafusion/builder.rs
```diff
@@ -172,12 +173,15 @@ impl DataFusionBuilder {
...
+
         DataFusion {
             runtime_status: self.status,
             ctx: Arc::new(ctx),
             data_writers: RwLock::new(HashSet::new()),
             cache_provider: RwLock::new(self.cache_provider),
             pending_sink_tables: TokioRwLock::new(Vec::new()),
+            cached_plans: Cache::new(DEFAULT_CACHED_PLANS_MAX_CAPACITY),
```
lukekim: Is there a default TTL we want to add here too? 1h?

crates/runtime/src/datafusion.rs
lukekim: Can't we just store the pointer to the plan? Box?

crates/runtime/src/datafusion.rs
```diff
@@ -1334,6 +1340,27 @@ impl DataFusion {
             }
         }
     }
+
+    /// Create or get a logical plan from the query
+    async fn get_or_create_logical_plan(
```
lukekim, line 1345: We must already have a function to do this from results cache for `cache_key_type: plan`? Can we not just use that almost directly?

crates/runtime/src/datafusion.rs
kczimm: We can hash the SQL string. The problem with Arc'ing the Logical Plan is that every time we call `Query::run` we ultimately call `datafusion::dataframe::Dataframe::new` which takes the `plan` by value so we cannot avoid a clone of the plan. Binding parameters via `LogicalPlan::with_param_values` also consumes the plan.

crates/runtime/src/datafusion.rs
kczimm: It looks like this

```rust
    async fn try_get_cached_result(
        df: &DataFusion,
        request_context: Arc<RequestContext>,
        mut tracker: QueryTracker,
        key: CacheKey<'_>,
    ) -> super::Result<CacheResult>
```

I believe it requires a bit of refactor if we want to make that generic for getting cached results and getting cached plans. Not sure it's worth it versus just adding ~20LOC.

kczimm:
I did some benchmarking on my laptop.

The time it takes to clone a 10,000 ASCII character String (representing a large SQL query):

image

The time it takes to hash a 10,000 ASCII character String (representing a large SQL query):

image

The time it takes to do a lookup in a `moka::Cache` if you store 512, 10,000 ASCII character Strings (representing a large SQL query). **DIVIDE THE RESULT BY 512 TO GET AVERAGE PER QUERY LOOKUP**

image

The time it takes to do a lookup in a `moka::Cache` if you store 512, u64 (representing the hash of a SQL query). **DIVIDE THE RESULT BY 512 TO GET AVERAGE PER QUERY LOOKUP**

image

I believe one approach is to store the SQL query Strings as keys in the map. You may have to clone then in or out and you have to do a String lookup. Clone + String Lookup on my machine is ~2 microseconds. The other approach is to hash the Strings, store the u64 as the key and do lookups on u64. Hash + u64 lookup on my machine is ~3.6 microseconds.

Storing Strings is more memory intensive but if we assume 10,000 ASCII character queries are large and limit to say 512 queries in the cache, the keys cost ~5MB of memory.

lukekim (dismissed):

kczimm:
I had some benchmarking in a comment but one of the experiments was flawed so I removed the graphs as to not mislead anyone.

kczimm:
Benchmarking TLDR:

- Cloning and lookup of SQL query (P99 on my MBP): ~2.41 microseconds
- Hashing SQL query and lookup of u64 (P99 on my MBP): ~2.14 microseconds

It was an interesting investigation but no revelation.

lukekim (approved):

PR #5489, opened 2025-04-22
feat: Record memory usage as a metric

peasee:
## 📝 Summary

* Records the memory usage during a benchmark test as metrics, for median and maximum memory usage.
* Updates the memory usage recorder to check every second, instead of every 5 seconds - to better capture memory usage for fast running tests.

### DIFF
--- a/tools/testoperator/src/commands/bench/mod.rs
+++ b/tools/testoperator/src/commands/bench/mod.rs
@@ -72,7 +72,7 @@ pub(crate) async fn run(args: &DatasetTestArgs) -> anyhow::Result<RowCounts> {
     let metrics: QueryMetrics<_, NoExtendedMetrics> = test.collect(TestType::Benchmark)?;
     let test_succeeded = test.succeeded();
     let mut spiced_instance = test.end()?;
-    let (max_memory, _) = observe_memory(memory_token, memory_readings).await?;
+    let (max_memory, median_memory) = observe_memory(memory_token, memory_readings).await?;
 
     let commit_sha = metrics.commit_sha.clone();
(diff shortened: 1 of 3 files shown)


sgrebnov (approved):

tools/testoperator/src/commands/bench/mod.rs
```diff
@@ -115,6 +115,8 @@ pub(crate) async fn run(args: &DatasetTestArgs) -> anyhow::Result<RowCounts> {
 
     crate::metrics::TEST_DURATION
         .record((metrics.finished_at - metrics.started_at).try_into()?, &[]);
+    crate::metrics::MAX_MEMORY_USAGE.record(max_memory * 1024.0, &[]);
```
lukekim, line 118:
```suggestion
    crate::metrics::PEAK_MEMORY_USAGE.record(peak_memory * 1024.0, &[]);
```

Can we call this peak? Max generally refers to allocated, while peak to used.

tools/testoperator/src/metrics.rs
```diff
@@ -98,3 +98,19 @@ pub static TEST_DURATION: LazyLock<Gauge<u64>> = LazyLock::new(|| {
         .with_unit("ms")
         .build()
 });
+
+pub static MAX_MEMORY_USAGE: LazyLock<Gauge<f64>> = LazyLock::new(|| {
+    METER
+        .f64_gauge("max_memory_usage_mb")
```
lukekim, line 104:
```suggestion
        .f64_gauge("peak_memory_usage_mb")
```

PR #5513, opened 2025-04-23
LLM router via `worker` spicepod component

Jeadie:
## 📝 Summary
 - Add a new spicepod concept `workers` that helps organise how to orchestrate multiple spicepod components to commit work. 
 - Add an initial `workers` component type: routers. Routers provide load-balancer like mechanisms for language models. Currently this supports: falling back on an LLM error, or routing between LLMs in a round-robin, or weighted percentage basis.
 - New CLI command `spice workers`

----

Spicepod YAML format: 
```yaml
workers:
  - name: round-robin
    description: |
      Call models 'foo' & 'bar' in round robin.
    models:
      - from: foo
      - from: bar
  - name: fallback
    description: |
      Call 'bar'. On error, call 'foo'. Failing that 'baz'.
    models:
      - from: foo
        order: 2
      - from: bar
        order: 1
      - from: baz
        order: 3
  - name: weighted
    description: |
      Routes 80% of traffic to 'foo'.
    models:
      - from: foo
        weight: 4
      - from: bar
        weight: 1

models:
  - name: foo
    from: openai:gpt-4o-mini
    params:
  - name: bar
    from: huggingface.co/meta-llama/Llama-3.3-70B-Instruct
  - name: baz
    from: file:gemma-2b.gguf
```

With router based workers, they can be called like any other LLM.
```bash
curl http://localhost:8090/v1/chat/completions \
  -d '{
    "model": "fallback",
    "messages": [{ "role": "user", "content": "Tell me a joke"}]
  }'
```

Example trace from `fallback`.
```bash
>>> spice trace ai_chat
Spice.ai OSS CLI v1.1.2

TREE                   STATUS DURATION   TASK
a1800e6025634f86       ✅       908.04ms ai_chat
  ├── a2bfdde38896673f 🚫       290.18ms ai_completion ## Errored 
  └── 1815c64f373faa3e ✅       615.74ms ai_completion
```

Spice CLI
```bash
>> spice workers
Spice.ai OSS CLI v1.2.0-unstable-build.86d7cfc6b

NAME        DESCRIPTION                                           MODELS
round-robin Call models 'foo' & 'bar' in round robin.             [map[from:foo] map[from:bar]]

fallback    Call 'bar'. On error, call 'foo'. Failing that 'baz'. [map[from:foo order:2] map[from:bar order:1] map[from:baz order:3]]

weighted    Routes 80% of traffic to 'foo'.                       [map[from:foo order:4] map[from:bar order:1]]
```

## 📚 Docs
 - https://github.com/spiceai/docs/pull/959


crates/runtime/src/http/v1/workers.rs
```diff
@@ -0,0 +1,165 @@
...
+            ..
+        } = value;
+
+        Worker {
+            name: name.clone(),
+            description: description.clone(),
+            models: models
+                .iter()
+                .filter_map(|m| serde_json::to_value(m).ok())
+                .collect(), // This cannot fail as it has already been serialized.
```
phillipleblanc: I'm not sure what this comment means

crates/runtime/src/metrics/workers/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+limitations under the License.
+*/
+
+use super::{Histogram, LazyLock, Meter, UpDownCounter, global};
+
+pub(crate) static WORKERS_METER: LazyLock<Meter> = LazyLock::new(|| global::meter("worker"));
+
+pub(crate) static COUNT: LazyLock<UpDownCounter<i64>> = LazyLock::new(|| {
+    WORKERS_METER
+        .i64_up_down_counter("spice.workers")
```
phillipleblanc: These metric names need to conform to our metric naming guidelines: https://github.com/spiceai/spiceai/blob/trunk/docs/dev/metrics.md

crates/workers/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    /// Load the worker's resources
+    async fn load(&self, _params: Arc<HashMap<String, SecretString>>) -> Result<()> {
+        Ok(())
+    }
+}
+
+/// Registry for managing workers
+pub struct WorkerRegistry {
+    workers: HashMap<String, Box<dyn SpiceWorker>>,
```
phillipleblanc:
```suggestion
    workers: HashMap<String, Arc<dyn SpiceWorker>>,
```

crates/workers/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+    /// Creates a new empty worker registry
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            workers: HashMap::new(),
+        }
+    }
+
+    /// Add a worker to the registry
+    pub fn add(&mut self, worker: Box<dyn SpiceWorker>) {
```
phillipleblanc:
```suggestion
    pub fn add(&mut self, worker: Arc<dyn SpiceWorker>) {
```

crates/workers/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    /// Get all workers in the registry
+    #[must_use]
+    pub fn all(&self) -> Vec<(&String, &dyn SpiceWorker)> {
+        self.workers.iter().map(|(k, v)| (k, v.as_ref())).collect()
+    }
+
+    /// Remove a worker by name
+    pub fn remove(&mut self, name: &str) -> Option<Box<dyn SpiceWorker>> {
```
phillipleblanc:
```suggestion
    pub fn remove(&mut self, name: &str) -> Option<Arc<dyn SpiceWorker>> {
```

crates/workers/src/router.rs
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    fn as_sql(&self) -> Option<&dyn SqlGeneration> {
+        None
+    }
+}
+
+/// Assumes all elements of `cfg` are [`worker::RouterConfig::Weighted`].
+/// Panics if no elements of `cfg` are [`worker::RouterConfig::Weighted`].
+fn select_from_weighted(cfg: &[worker::RouterConfig]) -> String {
```
phillipleblanc: Add tests for this file

crates/workers/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn params(&self) -> &HashMap<String, Value>;
+
+    /// Load the worker's resources
+    async fn load(&self, _params: Arc<HashMap<String, SecretString>>) -> Result<()> {
+        Ok(())
+    }
+}
+
+/// Registry for managing workers
+pub struct WorkerRegistry {
```
phillipleblanc: This doesn't appear to be used yet? Intentional?

crates/runtime/src/init/worker.rs
```diff
@@ -0,0 +1,102 @@
...
+        let worker = RouterModel::new(cfg.name.clone(), cfg.models.clone(), Arc::clone(&self.llms));
+
+        let mut llm_registry = self.llms.write().await;
+        llm_registry.insert(cfg.name.clone(), Arc::new(worker));
+
+        tracing::info!("Worker [{}] loaded, ready for use", cfg.name);
+        metrics::workers::COUNT.add(1, &[KeyValue::new("worker", cfg.name.clone())]);
+        self.status
+            .update_worker(&cfg.name, status::ComponentStatus::Ready);
+    }
```
kczimm: If we don't need the lock anymore, we should drop it explicitly rather than wait for it to go out of scope since we later call into a function, `update_worker`.

crates/runtime/src/init/worker.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    pub(crate) async fn apply_worker_diff(
+        &self,
+        current_app: &Arc<app::App>,
+        new_app: &Arc<app::App>,
+    ) {
+        // Remove workers that are no longer in the app
+        for worker in &current_app.workers {
+            if !new_app.workers.iter().any(|w| w.name == worker.name) {
```
kczimm, line 83: What happens if a user specifies multiple workers with the same `name`?

crates/workers/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+    pub fn remove(&mut self, name: &str) -> Option<Box<dyn SpiceWorker>> {
+        self.workers.remove(name)
+    }
+}
+
+impl Default for WorkerRegistry {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
kczimm: You should be able to `#[derive(Default)]`

crates/workers/src/router.rs
```diff
@@ -0,0 +1,253 @@
...
+    let weighted: Vec<(String, u32)> = cfg
+        .iter()
+        .filter_map(|c| {
+            if let worker::RouterConfig::Weighted { from, weight } = c {
+                Some((from.clone(), *weight))
```
kczimm: I believe the `clone` is unnecessary

crates/workers/src/router.rs
```diff
@@ -0,0 +1,253 @@
...
+        .collect::<Vec<_>>();
+
+    let index = if let Ok(dist) = WeightedIndex::new(weighted.iter().map(|(_, w)| w)) {
+        let mut rng = rng();
+        dist.sample(&mut rng)
+    } else {
+        0
+    };
+
+    weighted[index].0.clone()
```
kczimm: I don't think we are gauranteed that `weighted` has elements.

crates/runtime/src/http/v1/workers.rs
Jeadie: Sorry on wrong line, i've improved comment too.

crates/workers/src/lib.rs
Jeadie: Correct, I removed it initially in #4893 but it must have been dragged in.

crates/workers/src/router.rs
Jeadie: Where/how it is currently used yes we are (comments atop the function), but may as well make it return Option<String> instead.

crates/runtime/src/init/worker.rs
Jeadie: Right now the `Runtime` relies on names being unique (e.g. `HashMap`s keyed by `name` String). This section of code shouldn't cause issue. If there is multiple with the same name, we'd remove it from HashMaps several times, then the last worker with a given name would be selected (i.e. last added into HashMap).

crates/spicepod/src/component/worker.rs
```diff
@@ -0,0 +1,122 @@
...
+        ));
+    };
+
+    let strategy: RoutingStrategy = first.into();
+
+    if configs.iter().any(|c| RoutingStrategy::from(c) != strategy) {
+        return Err(de::Error::custom(
+            "All `.models` must be the same format, but are not.",
+        ));
+    }
```
peasee, line 119:
```suggestion
    if configs.iter().any(|c| c != first) {
        return Err(de::Error::custom(
            "All `.models` must be the same format, but are not.",
        ));
    }
```

Couldn't we replace this with an `Eq` implementation on `RouterConfig`? I'm not exactly sure why the `RoutingStrategy` enum exists when `RouterConfig` is already an enum.

peasee: `// If there is at least one element, check they are all the same variant.` ok, this makes more sense now.

phillipleblanc (approved):

PR #5518, opened 2025-04-23
Fix results cache by SQL with prepared statements

kczimm:
## 📝 Summary

When results cache type is SQL, we cache using the SQL query and its parameters.


Jeadie (dismissed):

sgrebnov (dismissed):

crates/runtime/src/datafusion/query/cache.rs
```diff
@@ -102,9 +102,9 @@ impl Query {
         };
 
         // Use the logical plan with parameter values for caching and lookup
-        let plan = match parameters {
+        let plan = match &parameters {
             Some(param_values) => plan
-                .with_param_values(param_values)
+                .with_param_values(param_values.clone())
```
phillipleblanc: Is this a true clone or just a reference counter increase? If the previous lets refactor this so that we're passing pointers to the value instead of cloning it

crates/runtime/src/datafusion/query/cache.rs
```diff
@@ -396,4 +398,150 @@ mod tests {
             })
             .await;
     }
+
+    #[tokio::test]
+    #[allow(clippy::too_many_lines)]
```
phillipleblanc:
```suggestion
```

crates/runtime/src/datafusion/query/cache.rs
```diff
@@ -396,4 +398,150 @@ mod tests {
...
+                assert_eq!(
+                    result.results_cache_status,
+                    QueryResultsCacheStatus::CacheMiss
+                );
+            })
+            .await;
+    }
+
+    #[tokio::test]
+    #[allow(clippy::too_many_lines)]
```
phillipleblanc:
```suggestion
```

phillipleblanc (approved):

Issue #5490, opened 2025-04-22
Enhancement: Accelerated views support


sgrebnov:
## Goal-State/What/Result

Spice supports views local acceleration (materialization) using the familiar `acceleration` configuration.

```yaml
views:
  - name: rankings
    sql: |
      WITH a AS (
        SELECT products.id, SUM(count) AS count
        FROM orders
        INNER JOIN products ON orders.product_id = products.id
        GROUP BY products.id
      )
      SELECT name, count
      FROM products
      LEFT JOIN a ON products.id = a.id
      ORDER BY count DESC;
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
      refresh_check_interval: 2h
```

## Why/Purpose

Materialized (locally accelerated) views significantly enhance performance, especially in environments with complex or heavy queries and large datasets. This is a common need among clients seeking faster, more efficient data access, and it provides familiar, industry-standard functionality that clients expect to use. 

## By When

**Issue/Spec written and reviewed:** 04/22
**M1. PoC** 04/23
**Done-Done:** 04/28

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
  - Developer experience first: re-use familiar acceleration configuration from datasets 
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] ~Cookbook Recipes/Tutorials~ (to be implemented as part of https://github.com/spiceai/spiceai/issues/5497)
- [x] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification 

Accelerated views begin initialization after their dependent datasets. They first wait for each referenced dataset to report a ready state before starting initialization and loading:
  1. Step1: All datasets are registered
  2. Step2: Views start initialization
    - wait for dependent datasets are registered / fail if there are invalid references
    - non-accelerated view: report ready state immediately
    - accelerated view: wait for dependent datasets report ready, perform initial data load, report ready

- [x] Spicepod definition supports `acceleration` section for views

```yaml
views:
  - name: rankings
    sql: |
      ...
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
[...]

PR #5526, opened 2025-04-24
Apply Spice advanced acceleration logic and params support to accelerated views 

sgrebnov:
## 📝 Summary

Apply common acceleration logic for both Datasets and Views by introducing `AccelerationSource` abstraction Data Accelerators can operate by. This apply Spice advanced acceleration logic and params support to accelerated views:
 - shared default acceleration file (`accelerated_duckdb.db`) for full federation
 - temp_directory

Note: this is first iteration to add most important functionality/logic support, the following item was created to make datasets attachment logic to include views to have full federation when accelerated datasets and views define custom acceleration file path.
- https://github.com/spiceai/spiceai/issues/5530

## 🔗 Related

Part of https://github.com/spiceai/spiceai/issues/5490

## 👀 Notes for Reviewers

Some `DataAccelerator` methods still use Dataset as param instead of `AccelerationSource`, this is to be addressed in follow-up PR to enabling checkpointing support for accelerated views.

```
fn is_initialized(&self, _dataset: &Dataset) -> bool {
..
}
```

### DIFF
--- a/crates/runtime/src/component/view.rs
+++ b/crates/runtime/src/component/view.rs
@@ -14,58 +14,55 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+use app::App;
 use datafusion::sql::TableReference;
 use serde_json::Value;
 use snafu::prelude::*;
 use spicepod::component::view as spicepod_view;
-use std::{collections::HashMap, fs};
+use std::{collections::HashMap, fs, sync::Arc};
+
+use crate::{Runtime, dataaccelerator::AccelerationSource};
 
 use super::{
-    dataset::{Dataset, acceleration},
+    dataset::{
+        Dataset,
+        acceleration::{self, Acceleration},
+    },
     validate_identifier,
 };
 use spicepod::semantic::Column;
 
 /// [`View`] is the internal representation of the [`spicepod_view::View`] spicepod component.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Clone)]
 pub struct View {
     pub name: TableReference,
     pub sql: String,
     pub metadata: HashMap<String, Value>,
     pub columns: Vec<Column>,
     pub acceleration: Option<acceleration::Acceleration>,
+    pub runtime: Arc<Runtime>,
+    pub app: Arc<App>,
 }
 
-impl TryFrom<spicepod_view::View> for View {
-    type Error = crate::Error;
+impl PartialEq for View {
+    fn eq(&self, other: &Self) -> bool {
+        self.name == other.name
+            && self.sql == other.sql
+            && self.metadata == other.metadata
+            && self.columns == other.columns
+            && self.acceleration == other.acceleration
+    }
+}
 
-    fn try_from(view: spicepod_view::View) -> Result<Self, Self::Error> {
-        validate_identifier(&view.name).context(crate::ComponentSnafu)?;
-
-        let table_reference = Dataset::parse_table_reference(&view.name)?;
-
-        let sql = if let Some(view_sql) = &view.sql {
-            view_sql.to_string()
-        } else if let Some(sql_ref) = &view.sql_ref {
-            Self::load_sql_ref(sql_ref)?
-        } else {
-            return Err(crate::Error::NeedToSpecifySQLView {
-                name: table_reference.to_string(),
-            });
-        };
-
-        let acceleration = view
-            .acceleration
-            .map(acceleration::Acceleration::try_from)
-            .transpose()?;
-
-        Ok(View {
-            name: table_reference,
-            sql,
-            metadata: view.metadata,
-            columns: view.columns,
-            acceleration,
-        })
+impl std::fmt::Debug for View {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("View")
+            .field("name", &self.name)
+            .field("sql", &self.sql)
+            .field("metadata", &self.metadata)
+            .field("columns", &self.columns)
+            .field("acceleration", &self.acceleration)
+            .finish_non_exhaustive()
     }
 }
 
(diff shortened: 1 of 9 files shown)


crates/runtime/src/component/view.rs
```diff
@@ -59,29 +116,89 @@ impl TryFrom<spicepod_view::View> for View {
...
+    }
+
+    pub fn build(self) -> crate::Result<View> {
+        let runtime = self.runtime.context(crate::UnableToCreateViewSnafu {
+            reason: "Runtime is not set.\nAn unexpected error occurred. Report a bug to request support: https://github.com/spiceai/spiceai/issues".to_string()
+        })?;
+
+        let app = self.app.context(crate::UnableToCreateViewSnafu {
+            reason: "App is not set.\nAn unexpected error occurred. Report a bug to request support: https://github.com/spiceai/spiceai/issues".to_string()
+        })?;
```
kczimm: There are a couple ways that we can avoid these errors at the syntax level. There is a state pattern we can use that looks like this:

```rust
use std::marker::PhantomData;

struct Runtime;
struct App;

struct View {
    sql: String,
    runtime: Runtime,
    app: App,
}

struct ViewBuilder<State> {
    pub sql: String,
    pub runtime: Option<Runtime>,
    pub app: Option<App>,
    _state: PhantomData<State>,
}

// ViewBuilder states
struct StateInit;
struct StateApp;
struct StateRuntime;

impl ViewBuilder<StateInit> {
    pub fn new<S: ToString>(sql: S) -> ViewBuilder<StateInit> {
        Self {
            sql: sql.to_string(),
            runtime: None,
            app: None,
            _state: PhantomData,
        }
    }
}

impl ViewBuilder<StateInit> {
    #[must_use]
    pub fn with_runtime(self, runtime: Runtime) -> ViewBuilder<StateRuntime> {
        ViewBuilder {
            sql: self.sql.to_string(),
            runtime: Some(runtime),
            app: None,
            _state: PhantomData,
        }
    }
}

impl ViewBuilder<StateRuntime> {
    #[must_use]
    pub fn with_app(self, app: App) -> ViewBuilder<StateApp> {
        ViewBuilder {
            sql: self.sql.to_string(),
            runtime: self.runtime,
            app: Some(app),
            _state: PhantomData,
        }
    }
}

impl ViewBuilder<StateApp> {
    #[must_use]
    pub fn build(self) -> View {
        View {
            sql: self.sql,
            runtime: self.runtime.expect("runtime set"),
            app: self.app.expect("app set"),
        }
    }
}

fn main() {
    let app = App;
    let runtime = Runtime;

    let sql = "SELECT 1";

    // The state structs enforce that build cannot be called without first
    // callling `with_runtime` and `with_app` first (in that order, here).
    let builder: View = ViewBuilder::new(sql)
        .with_runtime(runtime)
        .with_app(app)
        .build();

    // This does not compile
    // let builder: View = ViewBuilder::new(sql).build();
}
```

It's verbose but uses the type system to gaurantee we cannot build without the app or runtime. Another way is to simply do the following:

```rust
pub fn build_with(self, app: Arc<App>, runtime: Arc<Runtime>)
```

sgrebnov: I like the `build_with` approach ❤️

kczimm (dismissed):

peasee (dismissed):

crates/runtime/src/dataaccelerator/duckdb.rs
```diff
@@ -291,32 +291,32 @@ impl DataAccelerator for DuckDBAccelerator {
...
                 // then use the shared DuckDB file `accelerated_duckdb.db`
                 if !cmd.options.contains_key("open") {
-                    let duckdb_file = self.duckdb_file_path(this_dataset)?;
+                    let duckdb_file = self.duckdb_file_path(source)?;
                     cmd.options.insert("open".to_string(), duckdb_file);
                 }
 
-                let datasets = Arc::clone(&this_dataset.runtime)
-                    .get_initialized_datasets(&this_dataset.app, crate::LogErrors(false))
+                let datasets: Vec<Arc<Dataset>> = Arc::clone(&source.runtime())
```
phillipleblanc, line 316: I'm guessing this will come in a later PR, but we should have an abstraction that returns AccelerationSources here instead of Dataset

crates/runtime/src/dataaccelerator.rs
```diff
@@ -439,6 +439,25 @@ impl AcceleratorExternalTableBuilder {
     }
 }
 
+/// Represents acceleration source component, such as a dataset or a view.
+/// Provides additional information about the source, such as its name and associated runtime information.
+pub trait AccelerationSource: Sync {
```
sgrebnov, line 444: Alternatively this can be a struct/wrapper around dataset/view to provide unified functionality

crates/runtime/src/init/view.rs
```diff
@@ -83,34 +93,29 @@ impl Runtime {
...
-            view.sql.clone(),
-            view.acceleration.clone(),
-            secrets,
-        )
-        .context(UnableToAttachViewSnafu)
-        .inspect_err(|_| {
-            self.status
-                .update_view(&view.name, status::ComponentStatus::Error);
-        })?;
+        df.register_view(view.clone(), secrets)
```
phillipleblanc: The view clone is relatively expensive because the SQL that represent the view is stored directly. Can we store the SQL as an Arc<str>?

crates/runtime/src/component/view.rs
```diff
@@ -14,29 +14,84 @@ See the License for the specific language governing permissions and
...
+        Dataset,
+        acceleration::{self, Acceleration},
+    },
     validate_identifier,
 };
 use spicepod::semantic::Column;
 
 /// [`View`] is the internal representation of the [`spicepod_view::View`] spicepod component.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Clone)]
```
phillipleblanc: This is relatively expensive to Clone - let's either make it inexpensive (i.e. wrap internal components with Arc) or work with an Arc<View> directly

crates/runtime/src/dataaccelerator/duckdb.rs
sgrebnov: Confirm this will be added as part of this work item
 - https://github.com/spiceai/spiceai/issues/5530

crates/runtime/src/init/view.rs
sgrebnov: Updated: updated View.sql from String to Arc<str>

crates/runtime/src/component/view.rs
sgrebnov: Thank you - updated to operate by Arc<View> to avoid cloning

phillipleblanc (approved):

PR #5528, opened 2025-04-24
Support switching the memory allocator Spice uses via `alloc-*` features.

phillipleblanc:
## 📝 Summary

Spice uses the `snmalloc` memory allocator, which is optimized for speed and reducing memory fragmentation. However, it comes with a tradeoff of holding onto too much memory after loading a large dataset. Allow users to specify the memory allocator that works best for their workload by specifying one of the below Cargo features:

- `alloc-jemalloc`: Uses [`tikv-jemallocator`](https://crates.io/crates/tikv-jemallocator)
- `alloc-mimalloc`: Uses [`mimalloc`](https://crates.io/crates/mimalloc)
- `alloc-system`: Uses the default system allocator

Some light performance benchmarking on a medium sized (~6GB parquet/27GB uncompressed) dataset shows that `snmalloc` (the default allocator Spice uses) is the fastest, but also keeps more than double the memory around after the dataset loads compared to the system allocator, which is the most efficient at releasing memory back to the OS.

- `snmalloc`: Loaded 75,000,000 rows (27.28 GiB) for dataset log in 32s 902ms.
- `jemalloc`: Loaded 75,000,000 rows (27.28 GiB) for dataset log in 46s 231ms.
- `system`: Loaded 75,000,000 rows (27.28 GiB) for dataset log in 51s 483ms.
- `mimalloc`: Loaded 75,000,000 rows (27.28 GiB) for dataset log in 58s 290ms.

The following table shows the memory usage that Spice kept around AFTER the above data finished loading. Logically all memory related to the load has been released, but different allocators have different strategies for returning memory to the OS.

| Name                        | Memory |
|-----------------------------|--------|
| snmalloc | 6.4GiB |
| mimalloc | 8.1GiB |
| jemalloc | 3.8GiB |
| system         | 2.8GiB |

## Docker YAML anchors

I've wanted to template the Docker builds a bit more so that it would reduce the chance of making a mistake when adding 6 more Docker images to build for each architecture (so 12 in total). I first attempted to use YAML Anchors, but then I discovered that GitHub Actions doesn't support it ([yet](https://github.com/actions/runner/issues/1182)).

I've create the concept of `workflow_templates` which are just GitHub action workflow if they supported YAML anchors, and then I ran it through a tool `yq` that unparsed it into full YAML without anchors. I hope that makes it easier to review and modify in the future, and once the YAML anchor support does land in GitHub proper we can remove the templates entirely and just replace their generated counterpart.

## 📚 Docs

We should have a section in docs that guides people around which allocator they should use if the default one doesn't serve their needs.

## 👀 Notes for Reviewers

No need to review the generated spiced_docker.yml and spiced_docker_nightly.yml files - just review the templates they are generated from. I will callout with a comment the relevant sections in the template files.


.github/workflow_templates/spiced_docker.yml
```diff
@@ -0,0 +1,430 @@
...
+      - name: models-mimalloc
+        run: |
+          echo "IMAGE_TAG=models-mimalloc" >> $GITHUB_ENV
+          echo "CARGO_FEATURES=release,models,alloc-mimalloc" >> $GITHUB_ENV
+      - *build-arm64-image
+      - name: models-sysalloc
+        run: |
+          echo "IMAGE_TAG=models-sysalloc" >> $GITHUB_ENV
+          echo "CARGO_FEATURES=release,models,alloc-system" >> $GITHUB_ENV
+      - *build-arm64-image
```
phillipleblanc, line 206: This is the relevant section to review. AMD64 also follows this pattern, as well as the nightly workflow.

.github/workflow_templates/spiced_docker.yml
```diff
@@ -0,0 +1,430 @@
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Import images and create manifests
+        env:
+          REL_VERSION: ${{ needs.setup.outputs.rel_version }}
+          PUBLISH: ${{ startsWith(github.ref, 'refs/tags/v') || inputs.publish == 'true' }}
+        run: |
+          echo "REL_VERSION=${REL_VERSION}"
+          echo "PUBLISH=${PUBLISH}"
+          variants=("default" "odbc" "models" "jemalloc" "mimalloc" "sysalloc" "models-jemalloc" "models-mimalloc" "models-sysalloc")
```
phillipleblanc, line 309: Updated the variants to include the 6 new images.

.github/workflow_templates/spiced_docker_nightly.yml
```diff
@@ -0,0 +1,316 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Import images and create manifests
+        env:
+          REL_VERSION: ${{ needs.setup.outputs.rel_version }}
+          PUBLISH: ${{ github.event_name == 'schedule' || inputs.publish }}
+        run: |
+          echo "REL_VERSION=${REL_VERSION}"
+          echo "PUBLISH=${PUBLISH}"
+          variants=("models" "models-jemalloc" "models-mimalloc" "models-sysalloc")
```
phillipleblanc, line 264: Additional variants for nightly images

bin/spiced/src/main.rs
```diff
@@ -63,7 +94,14 @@ fn main() {
 
 async fn start_runtime(args: spiced::Args) -> Result<(), Box<dyn std::error::Error>> {
     spiced::in_tracing_context(|| {
-        tracing::info!("Starting runtime {version}", version = get_version_string());
+        if let Some(allocator_name) = get_allocator_name() {
+            tracing::info!(
+                "Starting runtime {version} (allocator: {allocator_name})",
```
phillipleblanc, line 99: I added this log to indicate the allocator being used, only if its not the default one.

ewgenius (approved):

PR #5533, opened 2025-04-25
Enable DatasetCheckpoint logic for accelerated views

sgrebnov:
## 📝 Summary

Similar to datasets, accelerated views should rely on the checkpointer to avoid unnecessary load and become ready faster if the data has already been loaded.

## 🔗 Related

Part of https://github.com/spiceai/spiceai/issues/5490

## 👀 Notes for Reviewers

`DatasetCheckpoint` struct and checkpoint table name are still based on `dataset` term (`const CHECKPOINT_TABLE_NAME: &str = "spice_sys_dataset_checkpoint";`), decided to keep this logic/naming as-is


crates/runtime/src/datafusion.rs
```diff
@@ -1360,20 +1360,31 @@ impl DataFusion {
...
 
+        // Detect if data for view was already loaded so we don't need to wait for the first refresh to complete to mark it as ready.
+        let mut initial_load_complete = false;
+        if let Ok(checkpoint) = DatasetCheckpoint::try_new(view).await {
+            if checkpoint.exists().await {
+                initial_load_complete = true;
+            }
+        }
+
         // TODO Extend and refactor (move to helper functionality that can be re-used with datasets logic)
```
sgrebnov, line 1371: Have a work item to complete this TODO - it seems there is a room to share the same logic when applying acceleration params when creating Dataset and View and refactor this.

phillipleblanc (approved):

PR #5536, opened 2025-04-25
feat: Add Runtime memory limit parameter

peasee:
## 📝 Summary

* Adds a `memory_limit` parameter to the runtime configuration:

```yaml
runtime:
  memory_limit: 4gb # or 4096mb or 4096kb
```

* When specified, configures a [`FairSpillPool`](https://github.com/apache/datafusion/blob/main/datafusion/execution/src/memory_pool/pool.rs#L136) with the specified memory limit. When the memory limit is reached, DataFusion spills the query operation to disk for supported query operations. Notably though, a number of operations are not supported like:
  * HashJoin - has a [tracking issue](https://github.com/apache/datafusion/issues/12952)
  * ExternalSorterMerge - no tracking issue, but there was a previous issue to implement spilling [for SortMergeJoin](https://github.com/apache/datafusion/issues/9359)
  * RepartitionMerge - [this issue](https://github.com/apache/datafusion/issues/7858) suggests it's supposed to be supported for spilling, but it might rely on the support from `HashJoin`.

* Passes through the `runtime.temp_directory` parameter to DataFusion's `DiskManager`, to specify the directory DataFusion should use to spill to disk.


crates/runtime/src/builder.rs
```diff
@@ -232,3 +244,26 @@ impl Default for RuntimeBuilder {
         Self::new()
     }
 }
+
+fn parse_memory_limit(memory_limit: Option<String>) -> Option<u64> {
+    let memory_limit = memory_limit?;
+    let original_memory_limit = memory_limit.clone();
+    let memory_limit = memory_limit.to_uppercase();
+    let memory_limit = if let Some(limit) = memory_limit.strip_suffix("GB") {
```
Jeadie: there isn't a library for this we've been using?

Jeadie (dismissed):
Needs documentation (possibly with those limitations mentioned)

crates/runtime/src/builder.rs
lukekim: Yes, we should use the library.

crates/runtime/src/datafusion/builder.rs
```diff
@@ -209,6 +233,52 @@ fn get_analyzer_rules() -> Vec<Arc<dyn AnalyzerRule + Send + Sync>> {
...
+        DiskManagerConfig::new()
+    };
+
+    let memory_pool: Arc<dyn MemoryPool> = if let Some(limit) = memory_limit {
+        let Ok(limit) = limit.try_into() else {
+            panic!("Memory limit could not fit into a usize");
+        };
+
+        let Some(topn) = NonZeroUsize::new(5) else {
+            panic!("Memory pool TopN must be greater than 0");
```
phillipleblanc: panic! -> unreachable

crates/runtime/src/datafusion/builder.rs
```diff
@@ -209,6 +233,52 @@ fn get_analyzer_rules() -> Vec<Arc<dyn AnalyzerRule + Send + Sync>> {
...
+) -> Arc<RuntimeEnv> {
+    let disk_manager = if let Some(directory) = temp_directory {
+        DiskManagerConfig::new_specified(vec![directory.into()])
+    } else {
+        DiskManagerConfig::new()
+    };
+
+    let memory_pool: Arc<dyn MemoryPool> = if let Some(limit) = memory_limit {
+        let Ok(limit) = limit.try_into() else {
+            panic!("Memory limit could not fit into a usize");
```
phillipleblanc: I think this should default to the max value if it can't fit, with an warning trace.

crates/runtime/src/builder.rs
```diff
@@ -232,3 +244,22 @@ impl Default for RuntimeBuilder {
         Self::new()
     }
 }
+
+fn parse_memory_limit(memory_limit: Option<String>) -> Option<u64> {
```
phillipleblanc, line 248: Add tests for this method

phillipleblanc (approved):

PR #5540, opened 2025-04-25
Move rate limit check to after parameterized query binding

### LINKED ISSUE #5538: Rate limiting of Flight DoPut should not affect prepared statements
Currently, we apply a rate limiting middleware for the Flight interface on the DoPut call to limit write access. This is now used for prepared statements which should not be limited.

phillipleblanc:
## 📝 Summary

Moves the rate limit check for `DoPut` requests to after the parameterized query binding. Previously all `DoPut` requests were for ingesting data, which we did want to enforce a rate limit on. However, parameterized query binding shouldn't be be subject to the same rate limit that was have for ingesting data.

This PR accomplishes that goal by moving the rate limit check from the middleware handler - which has no way to know whether the request needs to be rate limited or not - to the DoPut handler itself. It does this by encoding the rate limiting check into a callback function that gets added to the request extension. In the DoPut handler, if the request is not for parameterized query binding, we retrieve the rate limit check function from the request extension and call it to enforce the rate limit.

## 🔗 Related

Closes #5538


ewgenius (approved):

crates/runtime/src/flight/do_put.rs
```diff
@@ -48,12 +48,19 @@ use crate::{
...
-use super::{Service, flightsql::prepared_statement_query, metrics};
+use super::{
+    Service, flightsql::prepared_statement_query, metrics, middleware::RateLimiterExtension,
+};
 
 pub(crate) async fn handle(
     flight_svc: &Service,
     request: Request<Streaming<FlightData>>,
 ) -> Result<Response<<Service as FlightService>::DoPutStream>, Status> {
+    let rate_limit_check_fn = request
```
Jeadie, line 59: Does this need to be here? Or can be below on L85?

crates/runtime/src/flight/middleware.rs
```diff
@@ -160,41 +179,45 @@ where
...
-            let retry_after_secs = wait_time
-                .wait_time_from(wait_time.earliest_possible())
-                .as_secs();
+        let rate_limiter = Arc::clone(&self.rate_limiter);
 
-            tracing::trace!("Request rate-limited, must retry after {retry_after_secs} seconds.",);
+        // Create a rate limit check that the downstream service can use to check if the request should be rate limited.
+        // We don't directly check the rate limit here because not all DoPut requests should be rate limited
+        // (i.e. binding parameters to parameterized queries).
+        let check: Arc<RateLimitCheckFn> = Arc::new(move || {
```
Jeadie, line 193: Does this checker get applied anywhere except doPut? How/where does it do that?

Jeadie (approved):

crates/runtime/src/flight/middleware.rs
phillipleblanc: It's currently only for DoPut, that is checked by the URI above

crates/runtime/src/flight/do_put.rs
phillipleblanc: It needs to be here because the request object gets consumed by the into_inner call on the next line

PR #5541, opened 2025-04-25
Add aws_allow_http parameter for delta lake connector

Sevenannn:
## 📝 Summary

Add `delta_lake_aws_allow_http` parameter for delta lake connector. This unlocks the ability to connect to a s3 compatible storage with http scheme endpoint, e.g. MinIO endpoint.

The aws_allow_http conforms to the delta lake storage option spec, see reference: https://github.com/delta-io/delta-rs/blob/82661b85e1512398a469d55549230424b8fc689c/docs/integrations/object-storage/s3-like.md?plain=1#L73

This won't affect any existing delta lake connector users as it is an optional parameter. When not supplied, "aws_allow_http" won't be passed in as a storage option

UX:
```
  - from: delta_lake:s3://minio/table
    name: delta_lake_table
    params:
      delta_lake_aws_access_key_id: <my-id>
      delta_lake_aws_secret_access_key: <my-key>
      delta_lake_aws_endpoint: http://localhost:9000
      delta_lake_aws_allow_http: true
```

## 📚 Docs

Docs progress tracked in ticket: https://github.com/spiceai/spiceai/issues/5561


crates/runtime/src/dataconnector/delta_lake.rs
```diff
@@ -76,6 +76,9 @@ const PARAMETERS: &[ParameterSpec] = &[
     ParameterSpec::component("aws_endpoint")
         .description("The AWS endpoint to use for S3 storage.")
         .secret(),
+    ParameterSpec::component("aws_allow_http")
```
Jeadie, line 79: I assume these params just get passed down to https://github.com/apache/arrow-rs-object-store/blob/main/src/aws/builder.rs#L405 ?

Jeadie (dismissed):

peasee:
Could we also add this for the Databricks Delta parameters? https://spiceai.org/docs/components/data-connectors/databricks#aws-s3

crates/runtime/src/dataconnector/delta_lake.rs
phillipleblanc: That is correct

phillipleblanc (approved):

peasee (approved):

PR #5542, opened 2025-04-25
Databricks Service Principle M2M Auth support for dataset connector, spark connect and delta lake modes

### LINKED ISSUE #5468: Support Databricks Service Principal credentials and token refresh for data and catalog connectors

ewgenius:
## 📝 Summary

Concept:

Added optional `databricks_client_id` and `databricks_client_secret` params for Databricks dataset connector. (Either `databricks_token` or `databricks_client_id` and `databricks_client_secret` can be used for authentication)

When client_id and client_secret are provided, the connector will use the `DatabricksM2MTokenProvider` to manage the token lifecycle: new or existing instance of token provider pulled from internal registry, if two datasets configured with the same client_id, single token provider is used for both. (Will be extended to Databricks catalog catalog and model provider).

Service Principal credentials used to generate access token accroding to the [Databricks M2M flow](https://docs.databricks.com/aws/en/dev-tools/auth/oauth-m2m). By default each access token TTL is 1 hour, and `DatabricksM2MTokenProvider` will schedule token refresh 5 minutes before expiration.

* Refactored Databricks-related data connectors into a single `databricks` module
* Added `subscribe` support to the `TokenProvider` trait for token refresh notifications.
* Added support for dynamic token updates in `DatabricksSparkConnect` with a new `set_token` method and integration with the token provider. Updated `SparkConnect` to use `RwLock` for session management, enabling thread-safe token updates.
* Added `DatabricksM2MTokenProvider` for machine-to-machine token management, including token refresh logic and a shared registry for reusing token providers.


crates/data_components/src/databricks/auth.rs
```diff
@@ -0,0 +1,234 @@
...
+
+        // spawn background refresh loop
+        let cloned_client_id = client_id.clone();
+        let cloned_endpoint = endpoint.clone();
+        let cloned_tx = tx.clone();
+
+        let secret = client_secret.clone();
+
+        let handle = tokio::spawn(async move {
+            // schedule the first refresh at 90% of `expires_in`
```
ewgenius, line 97: not sure about 90%, considering to use static 10 minutes

crates/data_components/src/databricks/auth.rs
```diff
@@ -0,0 +1,234 @@
...
+                        tracing::debug!("M2M token refreshed; expires in {}", expires_in);
+                        let _ = cloned_tx.send(access_token.clone());
+                        #[allow(clippy::cast_precision_loss)]
+                        {
+                            next_wait = Duration::from_secs_f64(expires_in as f64);
+                        }
+                    }
+                    Err(e) => {
+                        tracing::error!("M2M token refresh failed: {}", e);
+                        // back‑off 60s on error
```
ewgenius, line 127: should use exponential back-off

ewgenius: probalby max_retry needed

crates/runtime/src/dataconnector/databricks/mod.rs
```diff
@@ -72,23 +83,72 @@ impl std::fmt::Debug for Databricks {
...
+            }
+            (None, None, Some(_)) => {
+                return MissingParameterSnafu {
+                    parameter: "databricks_client_id".to_string(),
+                }
+                .fail();
+            }
+            (Some(_), Some(_), Some(_)) => {
+                return InvalidConfigurationSnafu {
+                    message: "Cannot use `databricks_token`, `databricks_client_id` and `databricks_client_secret` together".to_string(),
```
phillipleblanc: Tell them what they should do to fix it.

crates/data_components/src/databricks/auth.rs
```diff
@@ -0,0 +1,234 @@
...
+            .field("client_id", &self.client_id)
+            .field("tx", &"<watch::Sender>")
+            .field("rx", &"<watch::Receiver>")
+            .field("_handle", &"<JoinHandle>")
+            .finish()
+    }
+}
+
+impl DatabricksM2MTokenProvider {
+    async fn new(
```
phillipleblanc:
```suggestion
    async fn try_new(
```

PR #5555, opened 2025-04-28
Use spiceai-macos for lint

lukekim:


sgrebnov (dismissed):

ewgenius (dismissed):

phillipleblanc (dismissed):

Jeadie (dismissed):

.github/workflows/pr.yml
```diff
@@ -165,16 +168,18 @@ jobs:
           fi
```
Sevenannn, line 168: This step needs a `if: runner.os != 'macOS'` condition

PR #5557, opened 2025-04-28
Run test operator every day and current and previous commits

lukekim:


.github/workflows/testoperator_dispatch.yml
```diff
@@ -33,7 +33,7 @@ on:
         description: 'Validate results?'
         required: false
         type: boolean
-        default: false
+        default: true
```
peasee:
```suggestion
        default: false
```

This is an escape hatch until validation is enabled for everything. Enabling this now will likely make a lot of tests start failing.

Validation is enabled on a spicepod-by-spicepod basis. E.g: https://github.com/spiceai/spiceai/blob/trunk/tools/testoperator/dispatch/tpch/sf1/federated/duckdb.yaml

peasee (changes requested):

.github/workflows/testoperator_dispatch.yml
```diff
@@ -44,28 +44,38 @@ jobs:
   dispatch-tests:
     name: Dispatch tests
     runs-on: spiceai-runners
+    concurrency:
+      group: testoperator-dispatch
+      cancel-in-progress: false
```
peasee: I don't think we should enable concurrency like this. It's completely valid for more than one to be running at a time, scheduling under different paths. I think a more relevant concurrency group would be:

```suggestion
    concurrency:
      group: testoperator-dispatch-${{ github.events.inputs.test_files_path || 'scheduled' }
      cancel-in-progress: false
```

.github/workflows/testoperator_dispatch.yml
```diff
@@ -77,7 +87,7 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SPICED_COMMIT: ${{ steps.setup-spiced.outputs.SPICED_COMMIT }}
-          WORKFLOW_COMMIT: ${{ github.ref }}
+          WORKFLOW_COMMIT: ${{ matrix.commit }}
```
peasee, line 94:
```suggestion
          WORKFLOW_COMMIT: ${{ github.ref }}
```

This needs to remain the ref of the executing dispatch workflow. Changing this to spiced commit will make dispatch execute on e.g. `trunk`, but the benchmark executes on a spiced commit HEAD.

Often, spiced commit will also differ from a fix branch where only changes to testoperator have been made. This also needs to remain `github.ref` for these cases.

phillipleblanc: I think for simplicity we should make the `matrix.commit` the same commit as the testoperator and the spiced under test. Trying to support a different testoperator version doesn't work in practice (i.e. snapshots being updated). Concretely this means we need to backport any relevant testoperator changes to the branch under test. Once we've gotten testoperator to be reliable, it shouldn't have too many changes.

.github/workflows/testoperator_dispatch.yml
```diff
@@ -44,28 +45,41 @@ jobs:
...
     runs-on: spiceai-runners
+    concurrency:
+      group: testoperator-dispatch-${{ github.event.inputs.test_files_path || 'scheduled' }}
+      cancel-in-progress: false
+    strategy:
+      matrix:
+        commit:
+          - ${{ github.event.inputs.spiced_commit || '' }}
+          - '00e9d7d671ca9dd28eac1e78ba1253b1ae33fe76' # release/1.1
+        exclude:
```
peasee, line 56: nice, this is neat

peasee (approved):

phillipleblanc (approved):

PR #5567, opened 2025-04-28
fix: Split scheduled dispatch into a separate job

peasee:
## 📝 Summary

* Splits the testoperator dispatch workflow into 2 jobs, one for triggered from input and one for from schedule. This makes it easier to manage each job type, instead of trying to manage the matrix within the same job


phillipleblanc (approved):

phillipleblanc:
Force merging this PR because it only affects the github workflow, not any spice runtime changes.

PR #5651, opened 2025-05-01
Add TokenProviderRegistry to the runtime

ewgenius:
## 📝 Summary

Added a new `subscribe` method to the `TokenProvider` trait, allowing providers to optionally return a `watch::Receiver` for token updates. The default implementation returns `None`.

Added a `TokenProviderRegistry` to manage and share instances of token providers among runtime components. Main goal is to allow having multiple connectors (e.g. Databricks data connector, catalog connector and model provider) to share a single token provider instance.

Registry not used anywhere yet, will be applied for #5468

Usage example:

```rs
let token_provider = rt.token_provider_registry
    .get_or_create_provider(format!("databricks_m2m_{}", client_id.to_string()), || async {
        DatabricksM2MTokenProvider::try_new(
            endpoint.to_string(),
            client_id.to_string(),
            client_secret.clone(),
        )
        .await
    })
    .await?
```

## 🔗 Related

- #5468
- decoupled from #5542

### DIFF
--- a/crates/runtime/src/registry/token_provider.rs
+++ b/crates/runtime/src/registry/token_provider.rs
@@ -38,7 +38,6 @@ impl TokenProviderRegistry {
     /// # Arguments
     ///
     /// * `key` - The key to use for the token provider.
-    /// * `provider_type` - The type of token provider to create.
     /// * `factory` - The factory function to use to create the token provider.
     ///
     /// # Returns
@@ -47,7 +46,6 @@ impl TokenProviderRegistry {
     pub async fn get_or_create_provider<P, E, F, Fut>(
         &self,
         key: String,
-        provider_type: &str,
         factory: F,
     ) -> Result<Arc<dyn TokenProvider>, E>
     where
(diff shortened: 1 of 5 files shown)


crates/runtime/src/registry/token_provider.rs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        {
+            let registry = self.token_provider_registry.read().await;
+            if let Some(provider) = registry.get(&key) {
+                tracing::debug!("Using existing token provider for key: {key}",);
+                return Ok(Arc::clone(provider));
+            }
+        }
+
+        let mut registry = self.token_provider_registry.write().await;
```
phillipleblanc, line 64: Move this registry guard to right before the insert - otherwise it could block other instances if the factory takes a long time to initialize.

phillipleblanc: So you will need to do a read guard to get the instance and return if exists, but then drop the read guard and call the factory and then take the write guard.

phillipleblanc (approved):

PR #5654, opened 2025-05-01
Databricks data_components refactor. Databricks Spark connect - add set_token method and writable spark session

ewgenius:
## 📝 Summary

Refactor the Databricks data_components module. Update spark-connect-rs, move the Spark session inside RwLock, and add `set_token` to enable token refresh.

## 🔗 Related

decoupled from #5542


crates/data_components/src/spark_connect.rs
```diff
@@ -65,7 +66,9 @@ impl SparkConnect {
     }
 
     pub async fn from_connection(connection: &str) -> Result<Self, Box<dyn Error + Send + Sync>> {
-        let session = Arc::new(SparkSessionBuilder::remote(connection)?.build().await?);
+        let session = SparkSessionBuilder::remote(connection)?.build().await?;
```
Jeadie, line 69: nit: Reuse of variable name `session` for two different types.

Jeadie (approved):

phillipleblanc (approved):

crates/data_components/src/spark_connect.rs
ewgenius: According to rust book, variable shadowing is a correct way https://doc.rust-lang.org/stable/book/ch03-01-variables-and-mutability.html#shadowing

PR #5659, opened 2025-05-01
Databricks M2M auth for spark connect data connector

ewgenius:
## 📝 Summary

Implemented the Databricks M2M authentication flow using a token provider. Updated the Databricks data connector in Spark mode to support service principal credentials - `databricks_client_id` and `databricks_client_secret`.

Every dataset that uses the Databricks connector in Spark mode will create or subscribe to an existing M2M token provider. The initial access token is acquired during the provider's initialization. The token provider is responsible for refreshing tokens, and each Spark data connector subscribes to the provider to receive updated values.

Updated FibonacciBackoff to expose the public method `next_duration` for use without the retry crate.

Prefixed Databricks env vars in integeration tests, to prevent autoloading in runtime and having conflicting configuration with both token and SP credentials used.

## 🔗 Related

part of #5468


crates/runtime/tests/databricks_delta/mod.rs
```diff
@@ -37,11 +37,11 @@ fn get_params() -> Params {
         vec![
             (
                 "databricks_endpoint".to_string(),
-                "${ env:DATABRICKS_HOST }".to_string(),
+                "${ env:TEST_DATABRICKSCKS_HOST }".to_string(),
```
Jeadie: what does _"databricks cks host"_ mean?

crates/data_components/src/databricks/auth.rs
```diff
@@ -0,0 +1,194 @@
...
+    ))]
+    UnableToGetToken { source: Box<dyn std::error::Error> },
+}
+
+#[derive(Clone)]
+pub struct DatabricksM2MTokenProvider {
+    endpoint: String,
+    client_id: String,
+
+    #[allow(dead_code)]
```
Jeadie: Is this still `dead_code` now that they get used in `impl TokenProvider for DatabricksM2MTokenProvider`?

crates/data_components/src/databricks/auth.rs
```diff
@@ -0,0 +1,194 @@
...
+    tx: watch::Sender<String>,
+    #[allow(dead_code)]
+    rx: watch::Receiver<String>,
+
+    _handle: Arc<JoinHandle<()>>,
```
Jeadie, line 46: What is the reason we need `_handle`?

crates/data_components/src/databricks/spark_connect.rs
```diff
@@ -17,30 +17,60 @@ limitations under the License.
...
         })
     }
+
+    pub async fn new_m2m(
+        endpoint: String,
+        cluster_id: String,
+        databricks_use_ssl: bool,
+        token_provider: Arc<dyn TokenProvider>,
+    ) -> Result<Self, Box<dyn std::error::Error + Send + Sync>> {
+        let token = token_provider.get_token().await?;
```
Jeadie, line 53: Could reduce code duplication of L55-60 via:

```rust
pub async fn new_m2m(
    endpoint: String,
    cluster_id: String,
    databricks_use_ssl: bool,
    token_provider: Arc<dyn TokenProvider>,
) -> Result<Self, Box<dyn std::error::Error + Send + Sync>> {
  let token = token_provider.get_token().await?;
  let result = self.new(endpoint, cluster_id, token, databricks_use_ssl)?;
...
```

Jeadie (commented):
Do we want any changes to `spice login databricks...`?

crates/runtime/tests/databricks_delta/mod.rs
ewgenius: oh, thanks fixed. that was a typo

crates/data_components/src/databricks/auth.rs
ewgenius: It is needed to hold token refresh async task

crates/data_components/src/databricks/auth.rs
ewgenius: Right, that was from old implementation, not needed anymore

crates/data_components/src/databricks/spark_connect.rs
ewgenius: Thanks!

ewgenius:
> Do we want any changes to spice login databricks...?

We'll need, but only after new auth supported for all connectors and modes

crates/data_components/src/databricks/auth.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    tx: watch::Sender<String>,
+    rx: watch::Receiver<String>,
+
+    _handle: Arc<JoinHandle<()>>,
+}
+
+impl fmt::Debug for DatabricksM2MTokenProvider {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("M2mTokenProvider")
```
phillipleblanc:
```suggestion
        f.debug_struct("DatabricksM2MTokenProvider")
```

crates/data_components/src/databricks/auth.rs
```diff
@@ -0,0 +1,192 @@
...
+        let (tx, rx) = watch::channel(access_token.clone());
+
+        // spawn background refresh loop
+        let cloned_client_id = client_id.clone();
+        let cloned_endpoint = endpoint.clone();
+        let cloned_tx = tx.clone();
+
+        let secret = client_secret.clone();
+
+        let handle = tokio::spawn(async move {
```
phillipleblanc, line 86: Would it make sense to use CancellableTaskHandle here? https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/cancellable_task.rs

crates/data_components/src/databricks/auth.rs
```diff
@@ -0,0 +1,192 @@
...
+                    secret.clone(),
+                )
+                .await
+                {
+                    Ok(TokenResponse {
+                        access_token,
+                        expires_in,
+                        ..
+                    }) => {
+                        tracing::info!("M2M token refreshed; expires in {}", expires_in);
```
phillipleblanc: Maybe bump this down to a debug log

phillipleblanc: I think we should also make this a metric

crates/data_components/src/databricks/spark_connect.rs
```diff
@@ -17,30 +17,55 @@ limitations under the License.
...
         let connection = format!(
             "sc://{endpoint}:443/;use_ssl={databricks_use_ssl};user_id=spice.ai;session_id={session_id};token={token};x-databricks-cluster-id={cluster_id}"
         );
         Ok(Self {
-            spark_connect: SparkConnect::from_connection(connection.as_str()).await?,
+            spark_connect: Arc::new(SparkConnect::from_connection(connection.as_str()).await?),
         })
     }
+
+    pub async fn new_m2m(
```
phillipleblanc, line 47: This is no longer specific to m2m, maybe call it `new_token_provider`?

Also maybe we have `new` be a wrapper around `new_token_provider` which just constructs a static provider.

crates/runtime/src/dataconnector/databricks.rs
```diff
@@ -94,40 +115,140 @@ impl Databricks {
...
-                    read_provider: Arc::new(databricks_spark.clone()),
-                })
             }
             _ => Err(Error::InvalidMode {
                 value: mode.to_string(),
             }),
         }
     }
 
+    pub fn build_auth_credentials(params: &Parameters) -> Result<AuthCredentials> {
```
phillipleblanc, line 148: Add unit tests for this function

crates/data_components/src/databricks/auth.rs
ewgenius: Yes, will do in a separate PR

crates/data_components/src/databricks/auth.rs
ewgenius: Will do in a separate PR

ewgenius: updated to use debug

crates/runtime/src/dataconnector/databricks.rs
```diff
@@ -94,40 +115,140 @@ impl Databricks {
...
+            }
+            (Some(_), Some(_), Some(_)) => {
+                InvalidConfigurationSnafu {
+                    message: "Choose either `databricks_token` or `databricks_client_id` and `databricks_client_secret`".to_string(),
+                }
+                .fail()
+            }
+            _ => {
+                InvalidConfigurationSnafu {
+                    message: format!("Invalid authentication configuration: {params:?}"),
```
Sevenannn: `params` contains other parameters that are not related to authentication configuration, which could be misleading. Consider adding an actionable message on fixing the invalid configuration
```suggestion
                    message: format!("Invalid authentication configuration. Choose either `databricks_token` or `databricks_client_id` and `databricks_client_secret`"),
```

crates/data_components/src/databricks/spark_connect.rs
ewgenius: new_m2m case is special for spark_connect, because we need to subscribe it to updates and pass down to spark session

crates/runtime/src/dataconnector/databricks.rs
ewgenius: added

phillipleblanc (approved):

Issue #5657, opened 2025-05-01
Enhancement: Metadata columns for object-store based data connectors


phillipleblanc:
## Goal-State/What/Result

Configure object-store based data connectors (i.e. s3, file) to expose metadata columns as queryable columns in the dataset.

The metadata columns are enabled by setting their name to `enabled` in the `metadata` section. The following metadata columns are available:
- `location`
- `last_modified`
- `size`

```yaml
datasets:
  - from: s3://spiceai-public-datasets/table1
    name: table1
    params:
      file_format: parquet
    metadata:
      location: enabled
      last_modified: enabled
      size: enabled
```

## Why/Purpose

Allows efficient SQL queries on specific files based on their metadata.

## By When

**Issue/Spec written and reviewed:** May 1, 2025
**Done-Done:** May 1, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

Use the above Spicepod mapping to enable the underlying metadata columns on the listing table provider configuration.

### Security Review

No security impact.

## QA Plan

Integration tests

phillipleblanc:
Decided to close this with the current UX and leave it undocumented.

PR #5661, opened 2025-05-01
Support metadata columns for object-store based data connectors

phillipleblanc:
## 📝 Summary

Configure object-store based data connectors (i.e. s3, file) to expose metadata columns as queryable columns in the dataset.

The metadata columns are enabled by setting their name to `enabled` in the `metadata` section. The following metadata columns are available:

- `location`
- `last_modified`
- `size`

```yaml
datasets:
  - from: s3://spiceai-public-datasets/table1
    name: table1
    params:
      file_format: parquet
    metadata:
      location: enabled
      last_modified: enabled
      size: enabled
```

## 🔗 Related

Part of #5657

## 📚 Docs

Will be required


crates/runtime/src/component/dataset/metadata.rs
```diff
@@ -0,0 +1,545 @@
...
+        if self.metadata_column_enabled(MetadataColumn::Size.name(), schema) {
+            columns.push(MetadataColumn::Size);
+        }
+
+        if columns.is_empty() {
+            None
+        } else {
+            Some(columns)
+        }
+    }
```
kczimm, line 53: Nit: If the distinction between empty and non-existent `Vec` is not important, this may be cleaner without any performance difference. You won't have to `unwrap` in your tests below.

- https://stackoverflow.com/questions/76708968/is-it-better-to-return-an-optionvec-or-just-an-empty-vec
- https://www.reddit.com/r/rust/comments/9neuby/empty_vec_or_optionvec_none_and_why/

```suggestion
    pub fn listing_table_metadata_columns(&self, schema: &Schema) -> Vec<MetadataColumn> {
        let needs_last_modified = self.needs_last_modified(schema);
        
        // Handle the common case where no metadata columns are enabled
        if !needs_last_modified && self.metadata.is_empty() {
            return vec![];
        }

        let mut columns = Vec::new();
        
        if self.metadata_column_enabled(MetadataColumn::LastModified.name(), schema)
            || needs_last_modified
        {
            columns.push(MetadataColumn::LastModified);
        }

        if self.metadata_column_enabled(MetadataColumn::Location.name(), schema) {
            columns.push(MetadataColumn::Location);
        }

        if self.metadata_column_enabled(MetadataColumn::Size.name(), schema) {
            columns.push(MetadataColumn::Size);
        }

        columns
    }
```

kczimm (dismissed):

crates/runtime/src/component/dataset/metadata.rs
phillipleblanc: Yeah, I was trying to avoid the allocation for the Vec if it wasn't required

phillipleblanc: But you're right that it probably doesn't make much difference

kczimm: Empty `Vec`s do not allocate, IIRC.

Jeadie (dismissed):

Sevenannn (dismissed):

Sevenannn (approved):

Issue #5658, opened 2025-05-01
Bug: accelerated view initial data load does not use federation


sgrebnov:
## Describe the bug

When view is accelerated sql does not use federation and instead does separate table scans.

This is due to general dataset refresh does not use/support federation.

PR #5677, opened 2025-05-02
Enable federated data refresh support for accelerated views

sgrebnov:
## 📝 Summary

Enable federation logic support for accelerated views by default that can be disabled using `disable_federation` param. 
1. We had `disable_query_push_down` undocumented parameter (applied for dataset access query federation) that was generalized to `disable_federation` to cover all federation logic (data refresh/load and queries). `disable_query_push_down` is still supported for backward capability.
2. Datasets load does not involve federation due to a single source, but with this change, the same unparsing mechanism that is used for federated queries will now apply to refresh SQL.

```yaml
views:
  - name: complex_vw
    sql: |
      ...
    acceleration:
      params: 
        disable_federation: true
```

Before 
```
2025-05-02T09:24:49.619891Z  INFO runtime::accelerated_table::refresh_task: Loaded 9,033,323 rows (920.50 GiB) for view test_view in 8m 39s 285ms.
```

After 
```
2025-05-02T06:29:52.585976Z  INFO runtime::accelerated_table::refresh_task: Loaded 9,033,323 rows (7.25 GiB) for view test_view in 2m 48s 800ms.
```

## 🔗 Related

Fixes 
- https://github.com/spiceai/spiceai/issues/5658

## 👀 Notes for Reviewers

1. The change must be covered by tests (to be added to this or next PR)

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task.rs
@@ -22,6 +22,7 @@ use arrow::{
 use arrow_schema::SchemaRef;
 use async_stream::stream;
 use datafusion::datasource::TableType;
+use datafusion::execution::SessionStateBuilder;
 use datafusion::logical_expr::dml::InsertOp;
 use datafusion_table_providers::util::retriable_error::{
     check_and_mark_retriable_error, is_retriable_error,
@@ -33,8 +34,9 @@ use tracing::{Instrument, Span};
 use util::fibonacci_backoff::FibonacciBackoffBuilder;
 use util::{RetryError, retry};
 
-use crate::datafusion::builder::get_df_default_config;
+use crate::datafusion::builder::{get_analyzer_rules, get_df_default_config};
 use crate::datafusion::error::{SpiceExternalError, find_datafusion_root, get_spice_df_error};
+use crate::datafusion::extension::SpiceQueryPlanner;
 use crate::datafusion::is_spice_internal_dataset;
 use crate::datafusion::schema::BaseSchema;
 use crate::federated_table::FederatedTable;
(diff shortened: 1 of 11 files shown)


crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -506,8 +515,23 @@ impl RefreshTask {
     }
 
     fn refresh_df_context(&self, federated_provider: Arc<dyn TableProvider>) -> SessionContext {
-        let ctx =
-            SessionContext::new_with_config_rt(get_df_default_config(), default_runtime_env());
+        let ctx = if self.disable_refresh_federation {
```
phillipleblanc: I'm not sure I understand why this needs to be a configuration parameter. Why not just always use the federated query planner?

sgrebnov: Combined params - thank you

crates/runtime/src/component/dataset/acceleration.rs
```diff
@@ -358,9 +358,9 @@ impl TryFrom<spicepod_acceleration::Acceleration> for Acceleration {
 
         let mut params = acceleration.params.clone();
 
-        let disable_query_push_down = match params
+        let disable_federation = match params
             .as_mut()
-            .and_then(|x| x.data.remove("disable_query_push_down"))
```
sgrebnov, line 363: `disable_query_push_down` is not documented param we added here: https://github.com/spiceai/spiceai/pull/2327

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -506,8 +516,23 @@ impl RefreshTask {
     }
 
     fn refresh_df_context(&self, federated_provider: Arc<dyn TableProvider>) -> SessionContext {
-        let ctx =
-            SessionContext::new_with_config_rt(get_df_default_config(), default_runtime_env());
+        let ctx = if self.disable_federation {
```
sgrebnov, line 519: This is the main change that unlocks federation for data refresh

```
.with_query_planner(Arc::new(SpiceQueryPlanner::new()))
.with_analyzer_rules(get_analyzer_rules())
```

kczimm (approved):

PR #5682, opened 2025-05-03
Update UX to disable acceleration federation

sgrebnov:
## 📝 Summary

Replace `acceleration.disable_federation: bool` with `acceleration.federation: enabled/disabled`.

## 🚨 Breaking Changes

The `disable_query_push_down` support is removed it is not a documented feature, does not follow preferred UX and is fully replaced by `acceleration.federation`

Not marking as a breaking change as it was not a documented feature


crates/runtime/src/component/dataset/acceleration.rs
```diff
@@ -439,3 +429,58 @@ impl Default for Acceleration {
         }
     }
 }
+
+/// Returns true if the `query_federation` parameter is set to "disabled".
+fn parse_is_query_federation_disabled(params: &mut Option<Params>) -> Result<bool, crate::Error> {
```
Sevenannn, line 434: Should the function name be `parse_query_federation` instead? Since the param is `parse_query_federation` instead of `is_query_federation_disabled`

sgrebnov: As function returns bool whether federation is disabled, it is easier to understand and use this function name to my taste, for `parse_query_federation` I would expect an enum Suppored/Disabled.

phillipleblanc (approved):

lukekim (approved):

PR #5717, opened 2025-05-06
Pass the default user agent string to the Databricks Spark, Delta, and Unity clients

ewgenius:
## 📝 Summary

Updated spark-connect-rs, to support user agent in connection string https://github.com/spiceai/spark-connect-rs/pull/9
Updated Databricks data and catalog connectors, to pass down defautl runtime user agent string `SpiceAI_OSS/<version>`

## 🔗 Related

Part of #5469

### DIFF
--- a/crates/runtime/src/catalogconnector/unity_catalog.rs
+++ b/crates/runtime/src/catalogconnector/unity_catalog.rs
@@ -132,7 +132,7 @@ impl CatalogConnector for UnityCatalog {
             Arc::new(StaticTokenProvider::new(token.clone())) as Arc<dyn TokenProvider>
         });
 
-        let client = Arc::new(UnityCatalogClient::new(endpoint, token_provider));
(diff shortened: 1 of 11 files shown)


crates/runtime/src/catalogconnector/unity_catalog.rs
```diff
@@ -132,7 +132,7 @@ impl CatalogConnector for UnityCatalog {
             Arc::new(StaticTokenProvider::new(token.clone())) as Arc<dyn TokenProvider>
         });
 
-        let client = Arc::new(UnityCatalogClient::new(endpoint, token_provider));
+        let client = Arc::new(UnityCatalogClient::new(endpoint, token_provider, None));
```
ewgenius: Detect if target connection is databricks

phillipleblanc (approved):

PR #5742, opened 2025-05-07
Remove HTTP auth from LLM config and simplify Databricks models logic by using static headers

Jeadie:
## 📝 Summary
- HTTP basic auth is not used by any current provider anymore. It will be removed. 
- Use `&'static str` for databricks header to simplify things.

### DIFF
--- a/crates/llms/src/config.rs
+++ b/crates/llms/src/config.rs
@@ -16,15 +16,10 @@ limitations under the License.
 #![allow(clippy::missing_errors_doc)]
 
 use async_openai::config::Config;
-use reqwest::header::{
-    AUTHORIZATION, CONTENT_TYPE, HeaderMap, HeaderName, HeaderValue, InvalidHeaderValue,
-};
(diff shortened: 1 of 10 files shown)


crates/llms/src/config.rs
```diff
@@ -194,7 +161,7 @@ impl Config for HostedModelConfig {
             Some(GenericAuthMechanism::BearerToken(prov) | GenericAuthMechanism::ApiKey(prov)) => {
                 Arc::new(SecretString::from(prov.get_token()))
             }
-            _ => Arc::new(DUMMY_API_KEY.clone()),
+            _ => Arc::new(SecretString::from(String::new())),
```
Sevenannn, line 166: nit: why SecretString::from(String::new() is used instead of DUMMY_API_KEY here

Sevenannn (dismissed):

crates/llms/src/config.rs
Jeadie: very similar, but `DUMMY_API_KEY` was LazyLock. This code path doesn't really happen in our codebase.

phillipleblanc (approved):

PR #5747, opened 2025-05-08
Databricks U2M Token Provider support

ewgenius:
## 📝 Summary

Sample spicepod

```yaml
  - from: databricks:spiceai_sandbox.default.messages
    name: messages
    params:
      databricks_endpoint: ${secrets:DATABRICKS_HOST}
      databricks_cluster_id: ${secrets:DATABRICKS_CLUSTER_ID}
      databricks_client_id: ${secrets:DATABRICKS_SP_CLIENT_ID} # Databricks oauth client id
      databricks_token: ${secrets:DATABRICKS_PAT} # Initial access token
```


crates/runtime/src/http/routes.rs
```diff
@@ -58,6 +58,8 @@ use axum::{
     routing::{Router, get, post},
 };
 use runtime_auth::layer::http::AuthLayer;
+#[cfg(feature = "databricks")]
```
phillipleblanc: I don't think this needs to be gated by the databricks feature?

crates/runtime/src/http/routes.rs
```diff
@@ -220,6 +222,14 @@ pub(crate) fn routes(
         .layer(Extension(rt.metrics_endpoint))
         .layer(Extension(config));
 
+    #[cfg(feature = "databricks")]
+    {
+        authenticated_router = authenticated_router.route_layer(middleware::from_fn_with_state(
```
phillipleblanc, line 225: Similarly, I don't think this needs to be gated by the databricks feature

crates/runtime/src/http/routes.rs
```diff
@@ -335,3 +345,33 @@ async fn check_shutdown(
 
     next.run(req).await
 }
+
+#[cfg(feature = "databricks")]
+async fn databticks_u2m_middleware(
```
phillipleblanc: Instead of defining a static function here, we should have a trait for token provider middleware we want to run and the databricks middleware adds itself to the middleware registry. The route creation would then add all token provider middlewares from the registry.

crates/runtime/src/model/chat.rs
```diff
@@ -301,6 +301,28 @@ async fn databricks(
...
         }
+        (Some(token),  Some(client_id), None) => {
+            let token_provider = DatabricksU2MTokenProvider::new(
+                endpoint.to_string(),
+                client_id.to_string(),
+                token.into()
+            )
+            .map_err(|e| LlmError::FailedToLoadModel {
+                source: Box::from(format!(
+                    "Could not retrieve M2M tokens from Databricks. Error: {e}"
```
phillipleblanc:
```suggestion
                    "Could not retrieve U2M tokens from Databricks. Error: {e}"
```

crates/runtime/src/http/routes.rs
```diff
@@ -220,6 +221,13 @@ pub(crate) fn routes(
         .layer(Extension(rt.metrics_endpoint))
         .layer(Extension(config));
 
+    {
+        authenticated_router = authenticated_router.route_layer(middleware::from_fn_with_state(
+            rt.token_provider_registry(),
+            databticks_u2m_middleware,
```
phillipleblanc:
```suggestion
            databricks_u2m_middleware,
```

crates/runtime/src/http/routes.rs
```diff
@@ -335,3 +343,27 @@ async fn check_shutdown(
 
     next.run(req).await
 }
+
+async fn databticks_u2m_middleware(
+    State(token_provider_registry): State<Arc<TokenProviderRegistry>>,
```
phillipleblanc:
```suggestion

async fn databricks_u2m_middleware(
    State(token_provider_registry): State<Arc<TokenProviderRegistry>>,
```

crates/runtime/src/http/routes.rs
phillipleblanc: I will take a look at this.

phillipleblanc (approved):

PR #5748, opened 2025-05-08
show available tables from catalogs in list_datasets tool

Jeadie:


crates/runtime/src/tools/builtin/list_datasets.rs
```diff
@@ -1,3 +1,4 @@
+use arrow::array::{Array, RecordBatch, StringArray};
```
phillipleblanc: Move to after copyright header

crates/runtime/src/tools/builtin/list_datasets.rs
```diff
@@ -124,6 +133,63 @@ pub async fn get_table_elements(
...
+    _opt_include: Option<&[String]>,
+) -> Result<Vec<ListDatasetElement>, Box<dyn std::error::Error + Send + Sync>> {
+    let Some(ref app) = *rt.app.read().await else {
+        return Ok(vec![]);
+    };
+
+    let catalogs: Vec<_> = app.catalogs.iter().map(|c| c.name.clone()).collect();
+
+    let batches = rt.datafusion().query_builder(format!(
+        "SELECT table_catalog, table_schema, table_name FROM information_schema.tables WHERE table_schema != 'information_schema' AND table_catalog in ({});",
```
phillipleblanc: we can just ask datafusion itself for the catalogs, no need for a SQL query

phillipleblanc: i.e. https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/datafusion.rs#L336

Jeadie: Nice, I didn't know this.

phillipleblanc (approved):

PR #5763, opened 2025-05-09
Control the parallelism of acceleration refresh datasets with `runtime.dataset_load_parallelism`

### LINKED ISSUE #5669: Bug: accelerated refresh does not respect `runtime.dataset_load_parallelism`
## Describe the bug

If I have 20+ accelerated datasets, initial data load for all of them happens in the same time and there is no parameter I can use to control this

phillipleblanc:
## 📝 Summary

The existing runtime parameter `runtime.dataset_load_parallelism` is now used to control the parallelism of accelerated table refreshes as well.

This was implemented by creating a Semaphore set to have a number of permits equal to `runtime.dataset_load_parallelism` if provided. Then whenever an accelerated dataset wants to perform a refresh, it needs to first get a permit from the semaphore. If there are no permits left (i.e. we are already at the maximum number of parallel accelerated refreshes) then it will wait until a permit becomes available (i.e. when a refresh finishes).

I did some small refactoring to the accelerated table structs to use a builder pattern to make this more ergonomic to use.

## 🔗 Related

- Closes #5669


ewgenius (dismissed):

crates/runtime/src/builder.rs
```diff
@@ -153,13 +153,24 @@ impl RuntimeBuilder {
             .as_ref()
             .and_then(|app| app.runtime.temp_directory.clone());
 
-        let mut df = DataFusion::builder(
+        let dataset_parallelism = self
+            .app
+            .as_ref()
+            .and_then(|app| app.runtime.dataset_load_parallelism);
+        tracing::info!("dataset_parallelism: {dataset_parallelism:?}");
```
sgrebnov:
```suggestion
```

sgrebnov (approved):

Sevenannn (approved):

Jeadie (approved):

PR #5771, opened 2025-05-11
Fix LLMs calling nested MCP tools

Jeadie:
#### LLM calling MCP tool bug
- [`self.call_tool`](https://github.com/spiceai/spiceai/blob/bdf0dedd1e6165c49b7e5e9c052be9253ab1b123/crates/runtime/src/model/tool_use.rs#L145) fails to find MCP encoded tool names. This results in all 
  - This was introduced in #5116 
  - This resulted in the tool not being called and a `null` response returned.
     ```shell
      Ran tools, and retrieved responses: [
        (ChatCompletionMessageToolCall {
            id: "call_IeaIhvu8Qdw7ah91ZvWNu6c7",
            type: Function,
            function: FunctionCall {
                  name: "spice__mcp_fs_list__directory",
                  arguments: "{\"path\":\"/Users/jeadie/Github/cookbook/mcp\"}" 
            }
      }, Null)] # This is the null response.
     ```

#### Improve nested MCP names
When tools are provided to model with non-default catalog, they will be provided with the catalog name prefixed. For example:
```yaml
models:
  - name: openai-with-spice
    from: openai:gpt-4o-mini
    params:
      tools: spice_mcp
```
Will produce:
```json
["spice_mcp/random_sample", "spice_mcp/list_datasets", "spice_mcp/document_similarity", "spice_mcp/sample_distinct_columns", "spice_mcp/sql", "spice_mcp/store_memory", "spice_mcp/get_readiness", "spice_mcp/fs/read_file", "spice_mcp/fs/read_multiple_files", "spice_mcp/fs/write_file", "spice_mcp/fs/edit_file", "spice_mcp/fs/create_directory", "spice_mcp/fs/list_directory", "spice_mcp/fs/directory_tree", "spice_mcp/fs/move_file", "spice_mcp/fs/search_files", "spice_mcp/fs/get_file_info", "spice_mcp/fs/list_allowed_directories", "spice_mcp/load_memory", "spice_mcp/table_schema", "spice_mcp/top_n_sample"]
```

### DIFF
--- a/crates/runtime/src/tools/mod.rs
+++ b/crates/runtime/src/tools/mod.rs
@@ -51,7 +51,7 @@ impl Tooling {
                 c.all()
                     .await
                     .iter()
-                    .map(|t| with_name(&t, format!("{catalog_name}/{}", t.name()).as_str()))
+                    .map(|t| with_name(t, format!("{catalog_name}/{}", t.name()).as_str()))
                     .collect()
             }
         }
--- a/crates/runtime/src/tools/mod.rs
+++ b/crates/runtime/src/tools/mod.rs
@@ -15,8 +15,9 @@ limitations under the License.
 */
 
 use catalog::SpiceToolCatalog;
+use factory::default_catalog_names;
 use std::{borrow::Cow, sync::Arc};
-use tools::SpiceModelTool;
+use tools::{SpiceModelTool, rename::with_name};
 
 pub mod builtin;
 pub mod catalog;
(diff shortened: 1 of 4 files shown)


crates/runtime/src/tools/mod.rs
```diff
@@ -40,7 +41,19 @@ impl Tooling {
     pub async fn tools(&self) -> Vec<Arc<dyn SpiceModelTool>> {
         match self {
             Tooling::Tool(t) => vec![Arc::clone(t)],
-            Tooling::Catalog(c) => c.all().await,
+            Tooling::Catalog(c) => {
```
phillipleblanc, line 44: Let's add a test for this.

phillipleblanc (approved):

PR #5780, opened 2025-05-12
Refactor `token_providers` crate into `token_provider` crate with generic token provider and registry types and `token_providers` runtime module with implementations

ewgenius:
## 📝 Summary

Refactored `token_providers` crate into two parts:
- `token_provider` crate to keep generic TokenProvider and registry types
- `token_providers` runtime module with databricks and github app token provider implementations

That will be needed later, since DatabricksU2MTokenProvider will depend on runtime request context.


crates/runtime/src/token_providers/databricks.rs
```diff
@@ -232,6 +231,7 @@ impl TokenProvider for DatabricksU2MTokenProvider {
 
 impl DatabricksU2MTokenProvider {
     #[allow(clippy::needless_pass_by_value)]
+    #[allow(clippy::unnecessary_wraps)]
     pub fn new(endpoint: String, client_id: String, token: SecretString) -> Result<Self> {
         let (tx, rx) = watch::channel(token.clone().expose_secret().to_string());
```
phillipleblanc:
```suggestion
    pub fn new(endpoint: String, client_id: String, token: SecretString) -> Self {
        let (tx, rx) = watch::channel(token.clone().expose_secret().to_string());

```

phillipleblanc (dismissed):

phillipleblanc (approved):

kczimm (approved):

PR #5807, opened 2025-05-13
Add DataFusion request context extension to http routes

ewgenius:
## 📝 Summary

Adds `Arc<DataFusion>` reference to request context through extension, and overrides it down the http routes stack.
Will be used later to update datafusion context by incoming headers.

## 🔗 Related

part of #5804


crates/runtime/src/http/routes.rs
```diff
@@ -215,8 +217,6 @@ pub(crate) fn routes(
     }
 
     authenticated_router = authenticated_router
-        .layer(Extension(Arc::clone(&rt.app)))
-        .layer(Extension(Arc::clone(&rt.df)))
```
ewgenius, line 219: Both the app and df references are injected after merging with the unauthenticated router, so there is no need to duplicate them.

crates/runtime/src/datafusion/datafusion_context.rs
phillipleblanc, line 1: Rename to `request_context_extension.rs` - the datafusion part is implied

crates/runtime/src/http/routes.rs
```diff
@@ -242,13 +242,28 @@ pub(crate) fn routes(
...
         .merge(authenticated_router)
         .route_layer(middleware::from_fn_with_state(rt.status(), check_shutdown))
+        .route_layer(middleware::from_fn(with_datafusion_context))
         .route_layer(middleware::from_fn(track_metrics))
         .layer(Extension(Arc::clone(&rt.app)))
+        .layer(Extension(Arc::clone(&rt.df)))
         .layer(cors_layer(cors_config))
 }
 
+async fn with_datafusion_context(mut req: Request<Body>, next: Next) -> impl IntoResponse {
```
phillipleblanc: I don't see a benefit to adding this to the `req` extensions since its already added to the request context.

crates/runtime/src/http/routes.rs
```diff
@@ -242,13 +242,28 @@ pub(crate) fn routes(
     unauthenticated_router
         .merge(authenticated_router)
         .route_layer(middleware::from_fn_with_state(rt.status(), check_shutdown))
+        .route_layer(middleware::from_fn(with_datafusion_context))
         .route_layer(middleware::from_fn(track_metrics))
         .layer(Extension(Arc::clone(&rt.app)))
+        .layer(Extension(Arc::clone(&rt.df)))
```
phillipleblanc: I want to force the code that relies on a `DataFusion` instance to always get it from the request context, so we should add this layer only to the `track_metrics` middleware layer, which will remove it from all other handlers.

Something like:

```rust
let track_metrics_with_df = middleware::from_fn(track_metrics)
        .layer(Extension(Arc::clone(&rt.df)));
..

.route_layer(track_metrics_with_df)
```

phillipleblanc: Otherwise, we're now in a place where we can get it either from the current Extension layer or from the request context.

crates/runtime/src/http/routes.rs
phillipleblanc: That would give us 3 ways to get access to the DataFusion context, which is confusing. We should just have one way.

crates/runtime/src/http/routes.rs
ewgenius: updated

crates/runtime/src/http/routes.rs
ewgenius: removed

crates/runtime/src/datafusion/request_context_extension.rs
```diff
@@ -0,0 +1,48 @@
...
+///
+/// Panics if the datafusion extension is not found in the request context.
+/// It is supposed to be set on all routes.
+pub fn get_current_datafusion(context: &Arc<RequestContext>) -> Arc<DataFusion> {
+    if let Some(df) = context.extension::<DataFusionContextExtension>() {
+        df.datafusion()
+    } else {
+        panic!("DataFusionContextExtension not found");
+    }
+}
```
phillipleblanc:
```suggestion
/// # Panics
///
/// Panics if the datafusion extension is not found in the request context.
pub fn get_current_datafusion(context: &Arc<RequestContext>) -> Arc<DataFusion> {
    if let Some(df) = context.extension::<DataFusionContextExtension>() {
        df.datafusion()
    } else {
        unreachable!("DataFusionContextExtension is expected to always be available");
    }
}
```

crates/runtime/src/http/routes.rs
```diff
@@ -242,12 +242,16 @@ pub(crate) fn routes(
...
+        .route_layer(middleware::from_fn_with_state(
+            Arc::clone(&rt.df),
+            track_metrics,
+        ))
         .layer(Extension(Arc::clone(&rt.app)))
         .layer(cors_layer(cors_config))
 }
 
 async fn track_metrics(
+    State(df): State<Arc<DataFusion>>,
```
phillipleblanc, line 254: nice!

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #5808, opened 2025-05-13
Move sandboxing logic to Dockerfile

phillipleblanc:
## 📝 Summary

In #4411 I implemented sandboxing logic to ensure that Spice was run as the `nobody` user in a chroot sandbox. It accomplished that sandboxing by placing a script that created a slim chroot environment, and then executing Spice in that chroot.

However, this meant that in order for Spice to create the chroot, it needed to be started with elevated privileges (as creating a chroot requires a privileged user), before it dropped into the chroot.

Kubernetes and Docker already provide good tooling to accomplish the same sandboxing that I did earlier, but at build time when constructing the Docker image itself, and at runtime by Kubernetes by applying a security context that enforces the image is not run as root.

This is accomplished by moving the sandboxing logic into an intermediate Docker build stage that finds all of the dependencies that Spice needs, and then the final layer is a `scratch` image that copies only exactly what Spice needs to run.

## 🔗 Related

- #4411

## 👀 Notes for Reviewers

This will have the side-effect that `exec`ing into the Docker image won't work, as there won't be a shell in the image anymore. The upside is that the image should be smaller, as we're not including any of the default Linux distro utilities.


lukekim (dismissed):

phillipleblanc:
I'm testing this in Kubernetes before merging

ewgenius (approved):

PR #5830, opened 2025-05-14
Add support for Databricks SQL Statement Execution API

### LINKED ISSUE #5790: Databricks Connector SQL API initial support

kczimm:
## 📝 Summary

Can query Databricks SQL warehouse using SQL Statement Execution API.

## 🔗 Related

- #5790 

## 📚 Docs

Updates needed:
- https://spiceai.org/docs/components/data-connectors/databricks
- https://github.com/spiceai/cookbook/tree/trunk/databricks

## 👀 Notes for Reviewers

I'm unsure about the JoinPushDown


crates/runtime/src/dataconnector/databricks.rs
```diff
@@ -339,6 +296,10 @@ const PARAMETERS: &[ParameterSpec] = &[
         .required()
         .secret()
         .description("The endpoint of the Databricks instance."),
+    ParameterSpec::component("sql_warehouse_id")
+        .required()
+        .secret()
+        .description("The ID of the SQL Warehouse. Only valid when mode is sql_warehouse"),
```
lukekim:
```suggestion
        .description("The SQL Warehouse ID to use when 'mode' is set to 'sql_warehouse'"),
```

lukekim: Below for 'mode':

"The execution mode for running queries: 'spark_connect', 'delta_lake', or 'sql_warehouse'."

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,448 @@
...
+
+        let response = self.execute_request(&token, &payload).await?;
+        schema_from_json(&response)
+    }
+
+    fn create_schema_payload(&self, table: &TableReference, sql: &str) -> Value {
+        json!({
+            "warehouse_id": self.sql_warehouse_id,
+            "catalog": table.catalog().unwrap_or("spiceai"),
+            "schema": table.schema().unwrap_or("public"),
```
Sevenannn: Does it make sense to default to spice catalog / schema here? The spice catalog / schema is different from databricks catalog / schema.

kczimm: Ah thanks, this was left in from my early development phase when I didn't know any better and didn't have error handling. I'll fix this.

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,451 @@
...
+
+    #[snafu(display("Failed to create table provider: {source}"))]
+    TableProviderCreationFailed { source: DataFusionError },
+
+    #[snafu(display("Failed to initialize SQL table: {source}"))]
+    SqlTableInitializationFailed {
+        source: datafusion_table_providers::sql::sql_provider_datafusion::Error,
+    },
+
+    #[snafu(display("A fully-qualified path for is required: {reason}"))]
```
phillipleblanc: grammar

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,451 @@
...
+                InvalidJsonArraySnafu {
+                    field: "external_links",
+                }
+                .build()
+            })?;
+
+        links
+            .iter()
+            .map(|link| {
+                serde_json::from_value(link.clone()).context(DeserializeExternalLinkFailedSnafu)
```
phillipleblanc: Could this be `into_iter` and then we avoid the clone?

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,451 @@
...
+            })
+    }
+
+    async fn query_arrow(
+        &self,
+        sql: &str,
+        _: &[&'a dyn Sync],
+        _projected_schema: Option<SchemaRef>,
+    ) -> Result<SendableRecordBatchStream, Box<dyn std::error::Error + Send + Sync>> {
+        let sql = sql.replace('\"', "");
```
phillipleblanc: Why is this necessary?

kczimm: image
I've noticed this API gets quoted schema names and symbols and Databricks does not like it.

crates/data_components/src/databricks/sql_warehouse.rs
kczimm: Not quite because of `Value::as_array` but I can do this:

```rust
    fn extract_external_links(mut response: Value) -> Result<Vec<ExternalLink>, Error> {
        let links = response
            .get_mut("result")
            .and_then(|result| result.get_mut("external_links").map(|v| v.take()))
            .ok_or_else(|| {
                MissingJsonFieldSnafu {
                    field: "result.external_links",
                }
                .build()
            })?;

        let Value::Array(links) = links else {
            return Err(Error::InvalidJsonArray {
                field: "external_links".into(),
            });
        };

        links
            .into_iter()
            .map(|link| serde_json::from_value(link).context(DeserializeExternalLinkFailedSnafu))
            .collect()
    }
```

crates/data_components/src/databricks/sql_warehouse.rs
phillipleblanc: We should handle this by using a Dialect option to disable the quoting, not by replacing it here. See https://github.com/apache/datafusion/blob/main/datafusion/sql/src/unparser/dialect.rs#L50

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,568 @@
...
+                return Err(Error::UnsupportedType {
+                    ty: type_name.to_string(),
+                });
+            }
+        } else {
+            (10, 0)
+        };
+
+        return Ok(DataType::Decimal128(precision, scale));
+    }
```
phillipleblanc:
```suggestion
    if type_name.to_uppercase().starts_with("DECIMAL") {
        let (precision, scale) = match type_name.split_once('(') {
            Some((_, params)) => {
                let params = params.trim_end_matches(')').trim();
                let parts: Vec<_> = params.split(',').map(str::trim).collect();
                if parts.len() != 2 {
                    return Err(Error::UnsupportedType { ty: type_name.to_string() });
                }
                let precision: u8 = parts[0].parse().map_err(|_| Error::UnsupportedType {
                    ty: type_name.to_string(),
                })?;
                let scale: i8 = parts[1].parse().map_err(|_| Error::UnsupportedType {
                    ty: type_name.to_string(),
                })?;
                (precision, scale)
            }
            None => (10, 0),
        };
        return Ok(DataType::Decimal128(precision, scale))
    }
```

I think this is slightly cleaner.

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,634 @@
...
+
+    // Fetch the arrow data at the external links, repeating for each chunk
+    async fn fetch_external_links(
+        &self,
+        result_object: Value,
+    ) -> Result<SendableRecordBatchStream, Error> {
+        let mut streams = Vec::new();
+        let token = self.token_provider.get_token();
+        let mut external_link = Self::extract_external_links(result_object)?;
+        loop {
```
sgrebnov: We should update this into real streamed approach as it seems right now we collect all batches first and start streaming only after that

This can be done next

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,634 @@
...
+                        .json()
+                        .await
+                        .context(JsonParsingFailedSnafu)?;
+                    external_link = Self::extract_external_links(result_object)?;
+                }
+                None => break,
+            }
+        }
+
+        let mut combined_stream = stream::select_all(streams);
```
sgrebnov: I think this will go away with fully streamed approach, but does this approach guarantee the right order of RecordBatches?

```
/// The stream will yield items as they become available on the underlying
/// streams internally, in the order they become available.
```

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,634 @@
...
+        let mut external_link = Self::extract_external_links(result_object)?;
+        loop {
+            tracing::trace!(
+                "Fetching chunk {} from {}",
+                external_link.chunk_index,
+                external_link.external_link
+            );
+            let bytes = self.fetch_chunk_data(&external_link.external_link).await?;
+            let batches = Self::read_arrow_batches(bytes)?;
+            streams.push(futures::stream::iter(batches.into_iter().map(Ok)));
```
phillipleblanc: This is buffering all of the batches into memory before returning the stream. It should return the batches into the stream as it is processing. So this logic to fetch the data should be in the stream itself.

crates/data_components/src/databricks/sql_warehouse.rs
phillipleblanc: The order of the batches matters, so we need to make sure the first batch is returned first, etc

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,655 @@
...
+    fn create_schema_payload(&self, table: &TableReference, sql: &str) -> Result<Value, Error> {
+        Ok(json!({
+            "warehouse_id": self.sql_warehouse_id,
+            "catalog": table.catalog().ok_or_else(|| Error::FullyQualifiedPath{ reason: "missing catalog".into() })?,
+            "schema": table.schema().ok_or_else(|| Error::FullyQualifiedPath{ reason: "missing schema".into() })?,
+            "statement": sql,
+        }))
+    }
+
+    async fn execute_request(&self, token: &str, payload: &Value) -> Result<Value, Error> {
```
Jeadie, line 177: Should we `.error_for_status()` here too?

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,655 @@
...
+
+        Ok(SqlWarehouseApi::fetch_external_links(Arc::clone(&self.api), result_object).await?)
+    }
+
+    async fn execute(
+        &self,
+        _query: &str,
+        _: &[&'a dyn Sync],
+    ) -> Result<u64, Box<dyn std::error::Error + Send + Sync>> {
+        Ok(NotImplementedSnafu.fail()?)
```
Jeadie, line 526: Is this expected?

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,655 @@
...
+        Box<dyn DbConnection<Arc<SqlWarehouseApi>, &'static dyn Sync>>,
+        Box<dyn std::error::Error + Send + Sync>,
+    > {
+        Ok(Box::new(SqlWarehouseConnection {
+            api: Arc::clone(&self.api),
+        }))
+    }
+
+    fn join_push_down(&self) -> JoinPushDown {
+        JoinPushDown::Disallow
```
phillipleblanc, line 129: This can come in a follow-up, but I think we should allow join push down here for the same endpoint and warehouse id?

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,655 @@
...
+    // Fetch the arrow data at the external links, repeating for each chunk
+    async fn fetch_external_links(
+        self: Arc<Self>,
+        result_object: Value,
+    ) -> Result<SendableRecordBatchStream, Error> {
+        let token = self.token_provider.get_token();
+        let initial_external_link = Self::extract_external_links(result_object)?;
+
+        let token = token.to_string();
+        let stream = stream::unfold(Some(initial_external_link), move |current_link| {
```
phillipleblanc, line 202: I like this use of unfold!

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -0,0 +1,655 @@
...
+            .await
+            .map_err(|source| dbconnection::Error::UnableToGetSchema {
+                source: Box::new(source),
+            })
+    }
+
+    async fn query_arrow(
+        &self,
+        sql: &str,
+        _: &[&'a dyn Sync],
```
phillipleblanc, line 499: Does the Databricks SQL warehouse API support parameterized queries?

phillipleblanc (dismissed):

crates/data_components/src/databricks/sql_warehouse.rs
kczimm: yes it does! I don't fully understand yet how we would use it here yet. It would be trivial if `AsyncDbConnection::query_arrow` had arguments for it here.

kczimm: I didn't realize that the `AsyncDbConnection::execute` takes parameters. I'll have to look into this more.

kczimm: Yes it does! I can see how we can add that in a follow-up.

crates/data_components/src/databricks/sql_warehouse.rs
kczimm: I was using the snowflake implementation as a reference which also leaves this API unimplemented. We can implement this if we want write operations in a follow-up.

sgrebnov (dismissed):

sgrebnov (approved):

Issue #5832, opened 2025-05-14
Enhancement: Results cache supports alternative hashing algorithms


peasee:
## Goal-State/What/Result

Results cache supports specifying alternative hashing algorithms.

## Why/Purpose

For improved performance of alternative hashing algorithms, like `ahash` - which can be 8-10x faster than the default `SipHash 1-3` used in `HashMap`.

These alternatives are provided as options, not defaults, as they could incur trade-offs that need to be acknowledged before switching.

## By When

**Issue/Spec written and reviewed:** 16th May, 2025
**Done-Done:** 15th May, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

Alternative hashing algorithms can be specified in the `results_cache` runtime configuration:

```yaml
results_cache:
    hashing_algorithm: ahash | siphash # default = siphash
```

### Security Review

* Only hashing algorithms which have evidence of high security should be implemented. For example, `ahash` utilises a keyed hashing algorithm with random key data injected for resistance against DoS attacks.

## How/Implementation Plan

1. `lru_cache` uses `build_with_hasher()` instead of `build()` to build the cache with the specified hasher.
2. `ahash` is implemented as an alternative hasher to be used by 1.

## QA Plan

1. Benchmark tests are created using the `ahash` hashing algorithm, to track and compare the performance of the algorithm.

## Release Notes

* **Improved Results Cache Performance & Hashing Algorithm:** In addition to performance optimizations for results caching, the Runtime now supports specifying an alternative hashing algorithm, `ahash`. The `siphash` algorithm is used by default, which can be changed by specifying a Runtime parameter:

```yaml
runtime:
  results_cache:
    hashing_algorithm: ahash
```

[...]

PR #5838, opened 2025-05-15
fix: Prevent excessively copying RawCacheKey

peasee:
## 📝 Summary

1. Remove the `Clone` and `Copy` implementations from `RawCacheKey`. This is a newtype that wraps a `u64` - the `u64` can be accessed directly through a borrow on `&self`, as `u64` itself implements `Copy`. This prevents a situation where a `RawCacheKey` is copied, only for the underlying `u64` to also be copied out of the newtype (e.g. when the newtype is hashed, etc).
    * Also includes associated updates to trait bounds for `QueryResultCache` and `QueryResultsCacheProvider` to accept borrowed values of `RawCacheKey` instead of owned values.
2. Prevent calculating the `RawCacheKey` inside `get_or_create_logical_plan` - if a cache miss occurs, the `RawCacheKey` will already be computed, so there is no reason to re-compute.
3. Updates the logical plan cache to use keys of `u64`, instead of keys of `RawCacheKey`. This aligns with existing behavior for the `LruCache`, which uses the `u64` value from the `RawCacheKey`.

Performing testing with `oha`, some headline metrics are observed:

* Before:
    * `cache_key_type: plan` - 153,438 req/s, P99 @ 0.0011
    * `cache_key_type: sql` - 160,441 req/s, P99 @ 0.0010
* After:
    * `cache_key_type: plan` - 184,764 req/s, P99 @ 0.0009
    * `cache_key_type: sql` - 199,520 req/s, P99 @ 0.0009

Translates to a roughly 20% improvement in query throughput, and reduction in P99 query latency.

## 🔗 Related

* Part of #5832

### DIFF
--- a/crates/runtime/src/datafusion/query/cache.rs
+++ b/crates/runtime/src/datafusion/query/cache.rs
@@ -44,11 +44,11 @@ pub(super) enum PlanOrCached {
 
 pub(super) struct RequestCacheManager {
     pub(super) cache_status: QueryResultsCacheStatus,
-    pub(super) raw_cache_key: Option<RawCacheKey>,
+    pub(super) raw_cache_key: RawCacheKey,
 }
 
 impl RequestCacheManager {
-    fn new(cache_status: QueryResultsCacheStatus, raw_cache_key: Option<RawCacheKey>) -> Self {
+    fn new(cache_status: QueryResultsCacheStatus, raw_cache_key: RawCacheKey) -> Self {
         Self {
             cache_status,
             raw_cache_key,
@@ -147,7 +147,7 @@ impl Query {
         Ok(PlanOrCached::Plan(
             plan,
             tracker,
-            RequestCacheManager::new(cache_status, Some(raw_cache_key)),
+            RequestCacheManager::new(cache_status, raw_cache_key),
         ))
     }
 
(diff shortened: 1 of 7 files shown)


crates/runtime/src/datafusion/query/cache.rs
```diff
@@ -133,23 +139,23 @@ impl Query {
...
+        let raw_cache_key = plan_cache_key.unwrap_or(sql_raw_cache_key);
 
         let cache_status = Self::should_cache_results(df, &plan, cache_status);
         tracker = tracker.results_cache_hit(false);
 
         Ok(PlanOrCached::Plan(
             plan,
             tracker,
-            RequestCacheManager::new(cache_status, raw_cache_key),
+            RequestCacheManager::new(cache_status, Some(raw_cache_key)),
```
phillipleblanc: Is this always Some() now?

phillipleblanc (dismissed):

crates/runtime/src/datafusion/query/cache.rs
peasee: Yeah - I'll refactor to take a concrete value. We had a test to denote a disabled cache manager, but it doesn't look like we use it anywhere else (e.g. forcibly specifying `None`).

phillipleblanc (approved):

PR #5852, opened 2025-05-16
fix: Disable the query tracker when task history is disabled

peasee:
## 📝 Summary

* Updates the `QueryTracker` on `Query` to be an `Option<QueryTracker>`
* When `task_history.enabled == false`, disables the `Option<QueryTracker>` by setting it to `None`

Benchmark results of an aggregation with low compute (`SELECT COUNT(*) FROM customer`) when task history is disabled:

* **Before:** 208k req/s, 0.0018 secs P99
* **After:** 228k req/s (+9.6%), 0.0016 secs P99 (-11%)

## 🔗 Related

* Part of #5832

### DIFF
--- a/crates/runtime/src/datafusion/mod.rs
+++ b/crates/runtime/src/datafusion/mod.rs
@@ -273,6 +273,7 @@ pub struct DataFusion {
     accelerator_engine_registry: Arc<AcceleratorEngineRegistry>,
     // Controls the parallelism of accelerated table refreshes
     acceleration_refresh_semaphore: Option<Arc<Semaphore>>,
+    pub task_history_enabled: bool,
 }
 
 impl std::fmt::Debug for DataFusion {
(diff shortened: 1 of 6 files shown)


crates/runtime/src/datafusion/mod.rs
```diff
@@ -273,6 +273,7 @@ pub struct DataFusion {
     accelerator_engine_registry: Arc<AcceleratorEngineRegistry>,
     // Controls the parallelism of accelerated table refreshes
     acceleration_refresh_semaphore: Option<Arc<Semaphore>>,
+    pub task_history_enabled: bool,
```
phillipleblanc: does this need to be pub?

peasee: It can be `pub(crate)`

phillipleblanc (dismissed):

phillipleblanc (approved):

Issue #5738, opened 2025-05-07
Enhancement: Support intercepting the Databricks U2M OAuth access token from API requests


ewgenius:
## Goal-State/What/Result

When the Databricks user access token and OAuth client ID are included in the headers of incoming HTTP and flight API requests, they are used in the related Databricks components, including datasets, catalogs, embeddings, and models.

## Why/Purpose

To ensure that the current user context is passed through for Databricks components configured with U2M authentication credentials.

## By When

**Issue/Spec written and reviewed:** May 7
**Done-Done:** June 10

## Principles

- Secure by default
- Developer experience first

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials
- [x] Release Notes prepared

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

When using the U2M OAuth flow, the runtime securely passes the user authentication context to the configured components for each individual user making requests.
It is expected that the user authentication context is provided with each API request to the runtime.

**spicepod configuration**

Databricks datasets, catalogs, embeddings, and models use the U2M OAuth flow when the following required parameters are specified:
-  `databricks_endpoint`: The Databricks endpoint.
-  `databricks_client_id`: The Client ID of the OAuth application configured in Databricks.

Overall authentication modes for Databricks:
- `databricks_token`: Uses a Personal Access Token (PAT).
- `databricks_client_id` + `databricks_client_secret`: Uses the machine-to-machine (M2M) flow.
- Only `databricks_client_id`: Uses the user-to-machine (U2M) flow and expects headers to be set.

Any other combination is invalid, and the runtime will display an error to the user:
> Error initializing dataset reviews. Invalid authentication configuration. Choose either `databricks_token` or `databricks_client_id`.

Databricks datasets with U2M auth do not support acceleration, and the runtime will warn the user about an invalid configuration:

> Error initializing dataset reviews. Invalid configuration: Cannot accelerate a dataset that requires user-scoped credentials.

**components loading**

[...]

PR #5858, opened 2025-05-16
Overried health checks for Databricks models in U2M auth mode

ewgenius:
## 📝 Summary

For Databricks models and embeddings `health` skips actual check and returns Ok(()), if model ocnfigured in U2M auth mode, which allows to load cmponents without checking connectivity.

## 🔗 Related

Part of #5738

### DIFF
--- a/crates/llms/src/databricks.rs
+++ b/crates/llms/src/databricks.rs
@@ -100,7 +100,6 @@ impl Chat for Databricks {
     }
 
     fn needs_health_check(&self) -> bool {
-        tracing::info!("needs_health_check, {:?}", self.needs_health_check);
         self.needs_health_check
     }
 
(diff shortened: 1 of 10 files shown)


Jeadie (commented):
This can be simplified to just databricks by overriding the health() method for Databricks to be an early return

crates/llms/src/databricks.rs
```diff
@@ -42,6 +42,7 @@ use crate::{
 pub struct Databricks {
     pub model: String,
     client: Client<HostedModelConfig>,
+    needs_health_check: bool,
```
phillipleblanc: nit: I would make this an enum like `HealthCheck::Required` and `HealthCheck::Skip` to make it more obvious what the booleans are when passing into the function.

phillipleblanc (dismissed):

Jeadie (dismissed):

crates/llms/src/databricks.rs
ewgenius: updated

phillipleblanc (approved):

Issue #3020, opened 2024-10-08
Enhancement: Results caching for prompts


lukekim:
\*REMEMBER, BE **SMART**!

S: Specific
M: Measurable
A: Achievable
R: Relevant
T: Time-Bound\*

## Goal-State/What/Result

_A clear and concise description of what the enhancement is and the target goal-state._

## Why/Purpose

_Clear and concise answer to why this should be done now._

## By When

_A target date for done-done completion of the entire enhancement._

**Issue/Spec written and reviewed:** <Target Date>
**Done-Done:** <Target Date>

## Done-Done

- [ ] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Samples/Quickstarts/Tutorials
- [ ] Release Notes

## The Algorithm

- [ ] Every requirement questioned?
- [ ] Delete (Scope) any part you can.
- [ ] Simplify.
- [ ] Break down into smaller iterations/milestones.
- [ ] Opportunities for automation.

## Spec

_Basic specification of the enhancement._

## How/Plan

_A clear and concise plan of how this should be implemented._

### Implementation (Sub Issues)

- Sub Issue 1
- Sub Issue 2

### QA

- Sub Issue 1
- Sub Issue 2

Issue #3018, opened 2024-10-08
Enhancement: Results caching for search


lukekim:
## Goal-State/What/Result

Search requests against tables with specific search terms and keywords are cached.

## Why/Purpose

To improve performance of search, and reduce cost by limiting search requests for the same content.

## By When

**Issue/Spec written and reviewed:** 20th May, 2025
**Done-Done:** 29th May, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

The runtime caching specifications are updated for search to use properties under `caching` to separate out different components that can be cached:

```yaml
runtime:
  caching:
    embeddings:
      enabled: true
      max_size: 128mb
      item_ttl: 5s
      eviction_policy: lru
      hashing_algorithm: siphash
    search_results:
      enabled: true
      max_size: 128mb
      item_ttl: 5s
      eviction_policy: lru
      hashing_algorithm: siphash
```

This allows for fine-grained control of each cached component. For example, a user can use a particular hashing algorithm on one component, and a different on another. A user can assign a larger cache size to one component, and a smaller size to another.

A property key of `max_size` instead of `cache_max_size` is used, removing `cache_` as the application of the property implies it is a cache as it is under a `caching` property.

#### Results Cache Deprecation

`runtime.results_cache` will be marked for deprecation after this enhancement. It will instead be replaced with `runtime.caching.sql_results`. Values in `runtime.results_cache` will still be accepted, but will log a warning for deprecation.

Values specified in `runtime.caching.sql_results` will override those from `runtime.results_cache`.

### Security Review

* Users should consider whether their embedded content is sensitive, and whether this data should be cached in memory. For these use-cases, users should disable the results cache.
* Malicious users could perform cache denial-of-service attacks, where they request excessive search terms or content which causes cache evictions of legitimate-use cache entries.
[...]

PR #5859, opened 2025-05-16
feat: Refactor caching, support hashing algorithms

peasee:
## 📝 Summary

* Refactors caching to consolidate `cache_provider` and `cached_plans` into `caching.results` and `caching.plans`. This will help expand caching into embeddings and models in future updates, like #3020 and #3018.
* Creates a new concept of a `CacheProvider` to wrap various caching mechanisms, like the `SimpleCache` that the plans cache uses - `SimpleCache` uses basic TTL and maximum entry count, while `LruCache` retains the existing behavior of operating on item byte weight with a least-recently-used eviction policy.
* Updates `LruCache` and `SimpleCache` with more generics, to simplify their reuse across different value and hash types - allowing them to be used interchangeable with e.g. `LogicalPlan` and `CachedQueryResult`.
* Moves the initialization of `caching.plans` into the `results_cache::init` - this probably needs to be further refactored into a dedicated initialization for logical plan caching, with its own associated configuration. This aligns with the existing initialization behavior of the results cache provider.
* Adds `results_cache.hashing_algorithm` with supported options of `ahash` and `siphash`.
* Updates `CacheKey::as_raw_key` to accept a `Hasher`, typically provided by the `CacheProvider::hasher()`

In `sql` cache key type, differences between `ahash` and `siphash` are usually negligble. The improvements with `ahash` are more obvious when larger SQL queries are used, especially in `plan` cache key type - which usually produces larger (in bytes) hashes than their plain SQL alternatives.

Benchmarking with `oha -c 50` using TPCH Q17, as `plan` cache key type:

* **Before:** 188k req/s, 0.0010 secs P99
* **After:** 223k req/s (+18%), 0.0008 secs P99 (-20%)

## 🔗 Related

* Part of #5832


peasee:
I still have some work to tie up before this is ready for review:

* [x] Update tests for `LruCache`, `SimpleCache`, `CacheProvider`, etc.
* [x] Update dependencies to make `ahash` a workspace level dependency.
* [x] Run benchmarks with validation to ensure these changes haven't impacted the correctness of results - especially when running as `ahash`.
* [ ] ~~Documentation!~~ documentation can merge after this PR

peasee:
Benchmarks ran with validation to confirm updates - no snapshot changes, validation passed successfully:
* [MySQL](https://github.com/spiceai/spiceai/actions/runs/15102671949)
* [PostgreSQL](https://github.com/spiceai/spiceai/actions/runs/15102635090)
* [S3](https://github.com/spiceai/spiceai/actions/runs/15102631998)
* [DuckDB](https://github.com/spiceai/spiceai/actions/runs/15102605748)

crates/cache/src/lib.rs
```diff
@@ -64,10 +68,49 @@ pub enum Error {
...
 
+#[derive(Debug)]
+pub enum CacheType {
+    Results,
+    Plans,
+}
+
+#[derive(Default)]
+pub struct Caching {
+    pub results: RwLock<Option<Arc<QueryResultsCacheProvider>>>,
```
phillipleblanc, line 105: Why is this RwLock needed?

crates/cache/src/lru_cache.rs
```diff
@@ -14,33 +14,57 @@ See the License for the specific language governing permissions and
...
+    cache: Cache<u64, V, T>,
+    hashing_algorithm: HashingAlgorithm,
 }
 
-impl LruCache {
-    pub fn new(cache_max_size: u64, ttl: Duration) -> Self {
-        let cache: Cache<u64, CachedQueryResult> = Cache::builder()
+pub trait Sizeable {
+    fn get_memory_size(&self) -> usize;
+}
```
phillipleblanc: This trait should probably live at the root?

crates/cache/src/lru_cache.rs
```diff
@@ -14,33 +14,57 @@ See the License for the specific language governing permissions and
...
+
+impl<V: Sizeable + Clone + Send + Sync + 'static, T: BuildHasher + Clone + Send + Sync + 'static>
+    LruCache<V, T>
+{
+    pub fn new(
+        cache_max_size: u64,
+        ttl: Duration,
+        hasher: T,
+        hashing_algorithm: HashingAlgorithm,
```
phillipleblanc: Isn't T a directly computable given `HashingAlgorithm`?

crates/runtime/src/datafusion/mod.rs
```diff
@@ -282,7 +279,7 @@ impl std::fmt::Debug for DataFusion {
             .field("runtime_status", &self.runtime_status)
             .field("data_writers", &self.data_writers)
             .field("accelerated_tables", &self.accelerated_tables)
-            .field("cache_provider", &self.cache_provider)
+            .field("cache_provider", &self.caching.results)
```
phillipleblanc: This should be `.field("caching")` now?

crates/cache/src/lib.rs
```diff
@@ -64,10 +68,49 @@ pub enum Error {
...
+    FailedToInvalidateCacheGeneric { source: moka::PredicateError },
 }
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
+#[derive(Debug)]
+pub enum CacheType {
+    Results,
+    Plans,
+}
```
phillipleblanc: This doesn't seem to be used anywhere?

crates/cache/src/lib.rs
phillipleblanc: If possible, I'd prefer this to be a builder where once its built it can't be changed.

crates/cache/src/lru_cache.rs
peasee: Yes, but `T` is a generic that the caller can specify. Hence, we cannot build it inside `LruCache<V, T>::new()` because the caller might specify a `T` where `T != HashingAlgorithm`.

peasee: I ended up refactoring this to rely on `BuildHasher` to make a `Hasher` instance instead.

crates/cache/src/lib.rs
peasee: I've replicated the `RwLock` from existing behavior. I've created a tracking issue: https://github.com/spiceai/spiceai/issues/5874

phillipleblanc (approved):

PR #5876, opened 2025-05-19
Define `SearchGeneration` paradigm & use in Vector Search .

Jeadie:
## 📝 Summary
 -  New crate `search` with `SearchGeneration` paradigm.
 - New module `crates/runtime/src/search`  in `runtime`. Moves logic out of `crates/runtime/src/embeddings/vector_search.rs`. 

## 🔗 Related
- #5844 

## 👀 Notes for Reviewers
 - Code in `crates/runtime/src/search/request.rs`, `crates/runtime/src/search/types.rs` and `crates/runtime/src/search/util.rs` is moved from `crates/runtime/src/embeddings/vector_search.rs` without any logic.


crates/runtime/src/search/request.rs
```diff
@@ -0,0 +1,528 @@
...
+        let result = SearchRequest::parse_keywords(&keywords);
+        assert!(result.is_ok());
+
+        // Test empty keyword
+        let keywords = vec![String::new()];
+        let result = SearchRequest::parse_keywords(&keywords);
+        assert!(result.is_ok());
+
+        // Test escaping keyword
+        let keywords = vec!["\"); DROP TABLE testing;".to_string()];
```
peasee: This test is failing, it's returning `Ok(_)` instead of `Err(_)`.

I think it's due to the format change of `ILIKE \"%{}%\"` to `ILIKE '{}'`.

We should investigate to make sure the `DROP TABLE` isn't being permitted. Alternatively, update this test to attempt escaping the `'` instead of `"`.

Jeadie: Looked at this separately. I agree, just need to update to a new valid SQL injection attack.

peasee (dismissed):
Some CI failures, but PR content LGTM

phillipleblanc (approved):

PR #5916, opened 2025-05-20
Load request context extensions on every flight incoming call

ewgenius:
## 📝 Summary

Renamed `run_extensions` to `load_extensions` for clarity. 

Load request context extensions on every incoming call to the flight API. Same way it is done for http 
https://github.com/spiceai/spiceai/blob/daa91f7bbd9da737802faa783110ad55310d1795/crates/runtime/src/http/routes.rs#L282-L287

### DIFF
--- a/crates/runtime/src/flight/middleware/request_context.rs
+++ b/crates/runtime/src/flight/middleware/request_context.rs
@@ -98,7 +98,7 @@ where
             );
 
         Box::pin(async move {
-            request_context.run_extensions().await;
+            request_context.load_extensions().await;
             request_context.scope(inner.call(req)).await
         })
     }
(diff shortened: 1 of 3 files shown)


crates/runtime/src/flight/middleware/request_context.rs
```diff
@@ -97,6 +97,9 @@ where
                 Arc::clone(&request_context) as Arc<dyn AuthRequestContext + Send + Sync>
             );
 
-        Box::pin(async move { request_context.scope(inner.call(req)).await })
+        Box::pin(async move {
+            request_context.run_extensions().await;
```
lukekim: Are these extensions or middleware?

phillipleblanc: Renaming this to `run_middleware` would work.

ewgenius: The term "middleware" doesn't fit here - extension don't have access to the request, and can't modify it. The idea is that some extensions may need to be loaded after being added to the context.

Initially that was `load_extensions`, renamed back for clarity.

phillipleblanc (approved):

PR #5929, opened 2025-05-20
Make `spill_to_disk_and_rehydration` test more robust

sgrebnov:
## 📝 Summary

1. Limit number of running Docker containers to 3 to prevent containers stopped due to OOM (`"OOMKilled": true`)
2. Don't use default acceleration folder (`.spice/data/accelerated_duckdb.db`) to avoid conflicts with other tests using the same default folder.
3. Unified mysql container naming based on port to avoid naming conflicts (for example recently added `mysql_character_set_results_test` conflicted with `mysql_integration_test`)

This should also improve `mysql_refresh_retries` (I previously saw the test failing time to time)

### DIFF
--- a/crates/runtime/tests/rehydration/mod.rs
+++ b/crates/runtime/tests/rehydration/mod.rs
@@ -63,13 +63,9 @@ async fn spill_to_disk_and_rehydration() -> Result<(), anyhow::Error> {
 
         let config = vec![
             #[cfg(feature = "duckdb")]
-            ("duckdb", None),
-            #[cfg(feature = "duckdb")]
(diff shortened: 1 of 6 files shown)


crates/runtime/tests/rehydration/mod.rs
```diff
@@ -65,13 +63,9 @@ async fn spill_to_disk_and_rehydration() -> Result<(), anyhow::Error> {
 
         let config = vec![
             #[cfg(feature = "duckdb")]
-            ("duckdb", None),
```
sgrebnov, line 68: Avoid using default acceleration folder to prevent conflicts with other tests.

phillipleblanc (approved):

PR #5932, opened 2025-05-20
Fix deferred loading for datasets with embeddings

ewgenius:
## 📝 Summary

If dataset is configured with embeddings, its connector is wrapped with `EmbeddingConnector`. If source data connector is deferred (e.g. Databricks with U2M auth), embedding wrapping should be applied only when the connector is ready.

## 🔗 Related

part of #5738

### DIFF
--- a/crates/runtime/src/init/dataset.rs
+++ b/crates/runtime/src/init/dataset.rs
@@ -577,19 +577,19 @@ impl Runtime {
             return Ok(Arc::new(LocalPodConnector::new(Arc::clone(&self.df))));
         }
 
-        let mut data_connector = match dataconnector::create_new_connector(source, params).await {
-            Some(dc) => dc.context(UnableToInitializeDataConnectorSnafu {})?,
-            None => {
+        let mut data_connector =
+            if let Some(dc) = dataconnector::create_new_connector(source, params).await {
+                dc.context(UnableToInitializeDataConnectorSnafu {})?
+            } else {
                 if source == ODBC_DATACONNECTOR {
                     return Err(OdbcNotInstalledSnafu.build());
-                } else {
-                    return Err(UnknownDataConnectorSnafu {
-                        data_connector: source,
-                    }
-                    .build());
                 }
-            }
-        };
+
+                return Err(UnknownDataConnectorSnafu {
+                    data_connector: source,
+                }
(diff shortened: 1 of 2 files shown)


crates/runtime/src/init/dataset.rs
```diff
@@ -215,7 +215,15 @@ impl Runtime {
             Ok(data_connector) => {
                 // if connector marked as deferred, replace it with stub DeferredConnector
                 if data_connector.deferred_load() {
-                    Arc::new(DeferredConnector::new(data_connector)) as Arc<dyn DataConnector>
+                    // If the dataset has embeddings, datafusion needs to load the wrapped connector when it is ready
+                    if ds.has_embeddings() {
```
phillipleblanc: A better interface would be to shift responsibility of wrapping the data connector to the dataset itself. So the dataset component would have a method like `ds.get_data_connector(data_connector) -> Arc<dyn DataConnector>` that would internally apply the EmbeddingConnector if required. That would also have prevented this bug in the first place - since it would have been more obvious that this logic needed to be applied.

As it is, this is turning into spaghetti code. :(

crates/runtime/src/init/dataset.rs
```diff
@@ -209,17 +209,11 @@ impl Runtime {
             .context(UnableToInitializeDataConnectorSnafu)?;
 
         let data_connector: Arc<dyn DataConnector> = match self
-            .get_dataconnector_from_source(source, params)
+            .get_dataconnector_from_dataset(Arc::clone(&ds), params)
+            // .get_dataconnector_from_source(source, params)
```
phillipleblanc:
```suggestion
```

phillipleblanc (dismissed):

crates/runtime/src/init/dataset.rs
ewgenius: Refactored to have a single method to create data connector and apply wrappers when needed. I'm not entierely sure if it would be beneficial to move it in the scope of `Dataset` type.

crates/runtime/src/init/dataset.rs
ewgenius: already removed

phillipleblanc (approved):

PR #5943, opened 2025-05-21
Define `CandidateAggregation` trait and implement RRF for multi column vector search. 

Jeadie:
## 📝 Summary
- Support multi-column vector search by defining a `CandidateAggregation` pattern and implementing Rank reciprocal fusion. 
- Simply add embeddings to extra columns

```yaml
datasets:
  - name: dataset
    from: file:cleaned_sales_data.parquet
    acceleration:
      enabled: true
    columns:
      - name: customer_name
        embeddings:
          - from: hf_minilm
            row_id:
              - order_line_number
      - name: product_line
        embeddings:
          - from: openai
      - name: address_line1
        embeddings:
          - from: hf_minilm
```

## 📚 Docs
 - [ ] TODO

## 👀 Questions for Reviewers
 - [ ] Current UX of `POST /v1/search` returns a single value per match `.matches[].value`. This is a natural UX for vector search on a single column (even when the data is chunked), but does not work for multi-column vector search. Current implementation selects one of the matched values.


crates/search/src/generation.rs
```diff
@@ -0,0 +1,55 @@
+/*
```
Jeadie, line 1: Unchanged in this PR. Moved from `crates/search/mod.rs`.

crates/runtime/tests/models/snapshots/integration_models__models__search__mulit_column_no_pks_basic_error_response.snap
```diff
@@ -0,0 +1,6 @@
+---
+source: crates/runtime/tests/models/search.rs
+expression: e.to_string()
+snapshot_kind: text
+---
+HTTP error: 500 Internal Server Error - Error occurred aggregating candidate search results: Cannot aggregate candidates from multiple sources (e.g. embedded columns of a dataset) when the dataset has no primary key.
```
phillipleblanc: Is this supposed to be a 500 error?

phillipleblanc (dismissed):

crates/runtime/tests/models/snapshots/integration_models__models__search__mulit_column_no_pks_basic_error_response.snap
Jeadie: Yes, I updated the testing to be able to snapshot expected errors. This is us ensuring we give a nice error if a user tries to do multi-column vector search without the required primary keys (stated or inferred).

phillipleblanc: Sure, but I don't think we want a 500 Internal Server Error to be an expected error type. This should be more of a 400 level error.

crates/runtime/tests/models/search.rs
```diff
@@ -0,0 +1,320 @@
...
+            runtime_ready_check(&rt).await;
+            for ts in test_cases {
+                run_search_test(http_base_url.as_str(), &ts).await?;
+            }
+            Ok(())
+        })
+        .await
+}
+
+#[ignore = "Non-deterministic order of search results makes snapshot testing unreliable"]
```
kczimm, line 165: can you sort them for the test?

crates/runtime/tests/models/snapshots/integration_models__models__search__mulit_column_no_pks_basic_error_response.snap
Jeadie: Good point. I can update this.

crates/runtime/tests/models/search.rs
Jeadie: The non-determinism is from the current implementation selecting one of the arbitrary `value` columns as the final value. This non-determinism cannot be solved in the tests. In the following PR, we will have a different UX for search results that can be sorted.

phillipleblanc (approved):

PR #5965, opened 2025-05-23
Read Iceberg tables from Glue Catalog Connector

### LINKED ISSUE #5724: Glue Catalog Connector for Iceberg tables

kczimm:
## 📝 Summary

Creates a Glue Catalog Connector that can read Iceberg tables.

## 🔗 Related

Closes:
- #5724


crates/runtime/src/catalogconnector/glue.rs
```diff
@@ -0,0 +1,522 @@
...
+}
+
+impl fmt::Debug for GlueCatalogProvider {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("GlueCatalogProvider")
+            .finish_non_exhaustive()
+    }
+}
+
+struct Inner {
```
Jeadie: Should this be called just `Inner`?

crates/runtime/src/catalogconnector/glue.rs
```diff
@@ -0,0 +1,522 @@
...
+                            DataFusionError::External(Box::new(Error::CreateIcebergTableProvider {
+                                source: e,
+                            }))
+                        })?;
+
+                    Ok(Some(Arc::new(table_provider)))
+                }
+                TableType::Unsupported => Ok(None),
+            }
+        } else {
```
Jeadie: nit: 
```rust
let Some(table) = self
  .inner
  .databases
  .get(&self.schema)
  .and_then(|tables| tables.iter().find(|t| t.name() == name)) else {
    tracing::error!("Glue schema name not in databases");
    return Ok(None);
}
match TableType::from(table) {...
```

crates/runtime/src/catalogconnector/glue.rs
```diff
@@ -0,0 +1,522 @@
...
+            tracing::error!("Glue schema name not in databases");
+            Ok(None)
+        }
+    }
+}
+
+// copy from iceberg-catalog-glue internals
```
Jeadie, line 236: nit: Add a link

crates/runtime/src/catalogconnector/glue.rs
```diff
@@ -0,0 +1,522 @@
...
+        } else {
+            None
+        }
+    }
+}
+
+#[async_trait]
+impl RefreshableCatalogProvider for GlueCatalogProvider {
+    async fn refresh(&self) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+        // TODO
```
Jeadie: nit: Github issue number

crates/runtime/src/catalogconnector/glue.rs
```diff
@@ -0,0 +1,522 @@
...
+        .secret(),
+    ParameterSpec::component("glue_aws_access_key_id")
+        .description("The AWS access key ID to use for Glue.")
+        .secret(),
+    ParameterSpec::component("glue_aws_secret_access_key")
+        .description("The AWS secret access key to use for Glue.")
+        .secret(),
+    ParameterSpec::component("glue_aws_session_token")
+        .description("The AWS session token to use for Glue.")
+        .secret(),
```
peasee: If this is a `glue` catalog connector, I don't think these parameters are right? As `::component`, they'll prefix with the catalog name and become `glue_glue_xxx`.

kczimm: It appears we explicitly use the prefix for catalogs?

- https://github.com/spiceai/cookbook/tree/trunk/catalogs/unity_catalog#aws-s3
- https://github.com/spiceai/cookbook/blob/trunk/catalogs/iceberg/iceberg-catalog-recipe/spicepod.yaml#L9
- https://github.com/spiceai/cookbook/tree/trunk/catalogs/databricks#step-3-add-the-databricks-unity-catalog-connector-to-spicepodyaml

kczimm: It happened to work because I was asking for the parameters by the same name elsewhere. But you're right. I've renamed the parameters and the names I use to look them up by removing the `glue` prefix.

crates/runtime/src/catalogconnector/glue.rs
```diff
@@ -0,0 +1,518 @@
...
+    }
+
+    async fn table(&self, name: &str) -> DFResult<Option<Arc<dyn TableProvider>>> {
+        let Some(table) = self
+            .inner
+            .databases
+            .get(&self.schema)
+            .and_then(|tables| tables.iter().find(|t| t.name() == name))
+        else {
+            tracing::error!("Glue schema name not in databases");
```
sgrebnov: Nit: add schema name to error to simplify troubleshooting

Sevenannn: Should this error be glue table not in the databases instead?

sgrebnov (dismissed):

crates/runtime/src/catalogconnector/glue.rs
kczimm: Yes good call. There is a mapping of terminology between Spice convention (schema.table) and Glue (database.table) and I'm mixing them in the code.

sgrebnov (dismissed):

sgrebnov (approved):

PR #5967, opened 2025-05-24
Alias the `random()` function to `rand()`

phillipleblanc:
## 📝 Summary

Aliases the DataFusion `random()` function to `rand()` to allow it to be used in federated push down to databases that support `rand()`, i.e. MySQL

### DIFF
--- a/crates/runtime/src/datafusion/udf.rs
+++ b/crates/runtime/src/datafusion/udf.rs
@@ -27,6 +27,8 @@ use datafusion::{
 
 use super::error::find_datafusion_root;
 
+pub mod alias;
+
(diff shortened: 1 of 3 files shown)


sgrebnov (approved):

crates/runtime/src/datafusion/udf.rs
```diff
@@ -27,6 +27,8 @@ use datafusion::{
 
 use super::error::find_datafusion_root;
 
+pub mod alias;
```
sgrebnov, line 30: Would be great to add unit or integration test (could be added separately)

PR #5971, opened 2025-05-25
Enable TPCH query results validation for Testoperator Dispatch

sgrebnov:
## 📝 Summary

Add option to enable TPCH query results validation.
 - TPCDS and ClickBench does not support validation currently so can't be enabled
 - Not all TPCH configurations pass validation so the option is not enabled by default

### DIFF
--- a/.github/workflows/testoperator_dispatch.yml
+++ b/.github/workflows/testoperator_dispatch.yml
@@ -128,7 +128,8 @@ jobs:
         run: |
           testoperator dispatch ./tools/testoperator/dispatch/tpch/sf1 \
             --workflow ${{ github.event.inputs.workflow_type }} \
-            --update-snapshots ${{ github.event.inputs.update_snapshots }}
+            --update-snapshots ${{ github.event.inputs.update_snapshots }} \
+            --validate
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/testoperator_dispatch.yml
```diff
@@ -128,7 +133,8 @@ jobs:
         run: |
           testoperator dispatch ./tools/testoperator/dispatch/tpch/sf1 \
             --workflow ${{ github.event.inputs.workflow_type }} \
-            --update-snapshots ${{ github.event.inputs.update_snapshots }}
+            --update-snapshots ${{ github.event.inputs.update_snapshots }} \
+            --validate ${{ github.event.inputs.validate_results }}
```
peasee, line 137: I don't think we should apply this here. Validation should be enabled at the dispatch-file level. If enabled here, the onus is now that the spicepod/component under test must pass validation before it can have an automated test - which is not always true for e.g. Alpha or Beta status components that may require a scheduled test without any specific guarantee on automated validation.

sgrebnov: Yeah, it seems there is a bug in dispatch that does not actually respect the `validate_results` setting, this should be fixed there

Issue #5556, opened 2025-04-28
Rearchitect worker paradigm to be broader than routing models.


Jeadie:
Workers in the Spice runtime represent configurable units of compute. There are several types of workers, of which, the type will fufill one or more possible traits (e.g. being callable like an LLM, is a trait).

**Definitions**
 - Worker: A spice component that encapsulates and organises configurable unit(s) of compute.
 - Worker Compute: A standalone spice component, self sufficient (i.e. can be used in its own right; has its own traits)
   - e.g: An LLM, a tool, code-running compute.
 - Worker Type: Determines how the worker can be configured, and what subset of compute traits are applicable.
   - e.g:
 - Compute Traits: Define how a worker component can be interacted with. All workers implement a worker trait that allows for basic introspection (e.g. CRUD of workers).
   - e.g. being callable like an LLM,

TODO: use retryable as trait for > worker types (separate for routing).

### YAML UX
```yaml
workers:
 - name: <string>
   type: <string>
   description: <String>
   params: <map[string]string; needed to instantiate the worker, independent of trait-level configuration>
   trait_x: <map[string]string; (optional) additional configuration of `worker_type` to implement `trait_x`. Specification should depend on `trait_x`, not `worker_type`.
   trait_y: <see `trait_x`>
```

#### YAML example: LLM Load-balancing
```yaml
workers:
  - name: a_b_llms
    type: load_balance
    description: |
      Ensure 80% of requests go to 'foo' and 20% go to 'bar'.
    load_balance:
      routing:
        - model: foo
          weight: 4
        - model: bar
          weight: 1
    retry: # retry is a trait implementable by any worker.
      max_attempts: 3
      back_off: fibonacci
      initial_backoff: 3s
```

#### YAML example: LLM postprocess (in code)
```yaml
# Assume some code running spice component definition.
code:
  - name: pythonner
    from: wasm:internal # Run code internally in Spice WASM runner.

workers:
  - name: email_sender
    type: postprocess # This type runs: `LLM -> code`.
    description: |
      This worker implements:
        - Chat completion API (output of code function returned in response payload).
    retry:
      # This will retry the whole flow.
      # To retry either the python code, or the LLM model separately, add retry to their workers.
      max_attempts: 1
      back_off: fibonacci
      initial_backoff: 3s
    model:
      # This section could configure anything pertaining to how `email_sender` acts as an LLM.
    postprocess:
      model: a_b_llms
      runner: pythonner
[...]

PR #5973, opened 2025-05-25
Initial worker concept

Jeadie:
## 📝 Summary
This is initial work towards makings spicepod workers a first class citizen in the Spice runtime. The spicepod UX for this is in https://github.com/spiceai/spiceai/issues/5556. 

This PR changes the spicepod worker UX as per the provided specification, and implements the `load_balance` trait for workers (i.e. `workers[].type: load_balance`). 

## 🔗 Related
#5556 

## 🚨 Breaking Changes
 - Changes to `workers` spicepod component YAML specification.
 - Changes to `v1/workers` API response payload.

## 📚 Docs
 -  https://github.com/spiceai/docs/pull/1015


peasee (commented):
Can you please update the PR summary?

Jeadie:
TODO
 - [x] No hardcoded `models` field in workers. 
 - [x] Move parameters for `load_balance` under `load_balance`spicepod field. 
 - [x] Don't default `.workers.type` to `load_balance`.

crates/runtime/src/http/v1/models.rs
```diff
@@ -50,7 +50,7 @@ pub(crate) struct OpenAIModelResponse {
     data: Vec<OpenAIModel>,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
```
peasee, line 53: If we make this `Clone`, can we update `OpenAIModel` to store `Arc<str>` instead of `String`? I'd imagine these values do not get updated often.

crates/runtime/src/http/v1/workers.rs
```diff
@@ -43,37 +41,25 @@ pub struct WorkersQueryParams {
...
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 #[serde(rename_all = "snake_case")]
-pub(crate) struct Worker {
+pub(crate) struct WorkerResponseItem {
     name: String,
     description: Option<String>,
-
-    /// Use generic to avoid using `utoipa::ToSchema` in `spicepod` crate.
-    models: Vec<serde_json::Value>,
+    r#type: String,
```
peasee, line 52: what is `type`? The type of worker, or the type of response? It is ambiguous.

crates/runtime/src/http/v1/workers.rs
```diff
@@ -43,37 +41,25 @@ pub struct WorkersQueryParams {
...
-    data: Vec<Worker>,
+    data: Vec<WorkerResponseItem>,
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 #[serde(rename_all = "snake_case")]
-pub(crate) struct Worker {
+pub(crate) struct WorkerResponseItem {
     name: String,
```
peasee, line 50: These are also probably candidates for `Arc<str>` if they do not update frequently.

crates/runtime/src/http/v1/workers.rs
```diff
@@ -119,33 +107,28 @@ models:gpt-4o,gpt-4o-writer,writer
         )))
     )
 ))]
-
+#[allow(clippy::type_complexity)]
```
peasee: let's move the type to a `type` instead of permitting this rule

crates/runtime/src/init/worker.rs
```diff
@@ -45,12 +44,24 @@ impl Runtime {
 
         tracing::info!("Loading worker [{}]...", cfg.name);
 
-        // Currently, only worker paradigm is [`RouterModel`].
-        let worker = RouterModel::new(cfg.name.clone(), cfg.models.clone(), Arc::clone(&self.llms));
+        let worker = match try_construct_worker(&cfg.r#type, cfg, self) {
+            Ok(worker) => worker,
+            Err(e) => {
+                tracing::error!("Failed to load worker [{}]: {}", cfg.name, e);
```
peasee:
```suggestion
                tracing::error!("Failed to load worker [{}]: {e}", cfg.name);
```

crates/runtime/src/status.rs
```diff
@@ -230,31 +230,34 @@ impl RuntimeStatus {
...
     /// Returns the status of all registered datasets.
     #[must_use]
     pub fn get_dataset_statuses(&self) -> HashMap<TableReference, ComponentStatus> {
+        self.get_statuses_of_prefix("dataset:")
+    }
+
+    /// Returns the status of all registered workers.
+    #[must_use]
+    pub fn get_worker_statuses(&self) -> HashMap<String, ComponentStatus> {
+        self.get_statuses_of_prefix("worker:")
```
peasee, line 245: Nice!

crates/runtime/src/worker.rs
```diff
@@ -0,0 +1,94 @@
...
+                Arc::new(model),
+                worker.description.clone(),
+            )))
+        }
+    }
+}
+
+#[async_trait]
+pub trait Worker: Send + Sync {
+    fn name(&self) -> Cow<'_, str>;
```
peasee, line 60: Does this need to be a `Cow`? Similar to other comments, I wouldn't imagine this gets updated so the `on-write` behavior seems redundant.

crates/runtime/src/http/v1/models.rs
Jeadie: `OpenAIModel` is created just before being returned in a HTTP response. It cannot be `Arc<str>` because the trait `ToSchema` is not implemented for `Arc<str>`.

crates/runtime/src/http/v1/workers.rs
Jeadie: As above.

crates/runtime/src/http/v1/workers.rs
Jeadie: `type` is the field within the worker spicepod component.

crates/runtime/src/http/v1/workers.rs
Jeadie: type is a field within the worker spicepod component, like `datasets[].from`

crates/runtime/src/worker.rs
Jeadie: Standardise with `trait SpiceModelTool`

crates/runtime/src/http/v1/workers.rs
```diff
@@ -109,7 +108,7 @@ fallback,\"Attempts bar first, then foo, then baz if previous models fail.\",loa
 ))]
 #[allow(clippy::type_complexity)]
```
peasee:
```suggestion
```

peasee (dismissed):

peasee (approved):

Issue #6007, opened 2025-05-27
Remove the `runtime.results_cache` parameter


peasee:
The `runtime.results_cache` has been deprecated, and is due to be removed.

PR #6008, opened 2025-05-27
refactor: Deprecate results_cache for caching.sql_results

### LINKED ISSUE #5996: Add `runtime.caching` block with `sql_results`
`sql_results` deprecates `runtime.results_cache`.

peasee:
## 📝 Summary

* Deprecates the `runtime.results_cache` parameter in favor of the new `runtime.caching.sql_results`
* Future tracking task for parameter removal: #6007

## 🔗 Related

* Closes #5996
* Part of #3018

## 📚 Docs

* Caching documentation will require updating, and a note for deprecation.

### DIFF
--- a/crates/runtime/src/builder.rs
+++ b/crates/runtime/src/builder.rs
@@ -31,6 +31,7 @@ use crate::{
     tracers,
 };
 use app::App;
+use spicepod::component::runtime::Caching;
 use token_provider::registry::TokenProviderRegistry;
 use tokio::sync::{Mutex, RwLock};
 
@@ -137,6 +138,7 @@ impl RuntimeBuilder {
         self
     }
 
+    #[allow(clippy::too_many_lines)]
     pub async fn build(self) -> Runtime {
         self.accelerator_engine_registry.register_all().await;
         dataconnector::register_all().await;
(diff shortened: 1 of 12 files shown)


crates/runtime/src/builder.rs
```diff
@@ -279,16 +297,19 @@ fn parse_memory_limit(memory_limit: Option<String>) -> Option<u64> {
         .map(|v| v.get_adjusted_unit(byte_unit::Unit::B).get_value() as u64);
 
     if memory_limit.is_none() {
-        tracing::warn!(
```
Jeadie, line 282: Why did we have to wrap these `tracing::warn!` in `crate::in_tracing_context(|| {`? (and subsequently why we moved from `spiced::in_tracing_context` to `runtime::in_tracing_context`?

Jeadie (approved):

crates/runtime/src/builder.rs
peasee: Tracing isn't initialised until the runtime is built as it relies on a datafusion instance - so within the runtime builder itself, tracing calls will print nowhere. I moved "in_tracing_context" into runtime to make it easier to access.

crates/spicepod/src/component/runtime.rs
```diff
@@ -27,8 +27,10 @@ const TASK_HISTORY_RETENTION_MINIMUM: u64 = 60; // 1 minute
 #[cfg_attr(feature = "schemars", derive(JsonSchema))]
 #[serde(deny_unknown_fields)]
 pub struct Runtime {
+    #[serde(default, skip_serializing_if = "is_default_or_none")]
+    pub results_cache: Option<ResultsCache>,
```
Sevenannn, line 31: For learning purpose: why serialization is skipped here when it is default

Sevenannn (approved):

crates/spicepod/src/component/runtime.rs
kczimm: My understanding is the following: Default for an `Option` is `Option::None`. We are saying that we would rather not take the time/space to serialize null or the default values. If we go to deserialize and we do not see this field specified, we can recreate the default value without the need to transmit.

kczimm (approved):

PR #6009, opened 2025-05-27
fix: Ensure snapshots on different scale factors are retained

peasee:
## 📝 Summary

* Ensures that snapshots for different scale factors are stored separately.

### DIFF
--- a/crates/test-framework/src/snapshot/mod.rs
+++ b/crates/test-framework/src/snapshot/mod.rs
@@ -28,6 +28,7 @@ pub async fn record_explain_plan(
     client: &FlightClient,
     name: &str,
     query: &Query,
+    scale_factor: f64,
 ) -> anyhow::Result<()> {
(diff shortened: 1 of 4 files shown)


crates/test-framework/src/snapshot/mod.rs
```diff
@@ -62,8 +63,14 @@ pub async fn record_explain_plan(
             (r#"grouping\((?:store|"store")\.(?:s_state|s_county|"s_state"|"s_county")\),\s*grouping\((?:store|"store")\.(?:s_state|s_county|"s_state"|"s_county")\)"#, "<GROUPING_PAIR>")
         ],
     }, {
+        let snapshot_name = if (scale_factor - 1.0).abs() < f64::EPSILON {
```
sgrebnov, line 66: Nice 👍

sgrebnov (approved):

PR #6013, opened 2025-05-27
Fix models benchmark results reporting

sgrebnov:
## 📝 Summary

First record batch can actually be empty, so we should find the first non-empty record batch to get the evaluation metrics.

### DIFF
--- a/tools/testoperator/src/commands/evals/mod.rs
+++ b/tools/testoperator/src/commands/evals/mod.rs
@@ -68,10 +68,10 @@ pub struct EvalMetrics {
 
 impl EvalMetrics {
     pub fn from_record_batch(batch: &[RecordBatch]) -> anyhow::Result<Self> {
-        if batch.is_empty() || batch[0].num_rows() == 0 {
-            return Err(anyhow::anyhow!("No evaluation metrics found"));
-        }
-        let record = &batch[0];
+        let record = batch
+            .iter()
+            .find(|b| b.num_rows() > 0)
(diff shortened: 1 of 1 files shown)


Sevenannn (dismissed):

tools/testoperator/src/commands/evals/mod.rs
```diff
@@ -68,10 +68,8 @@ pub struct EvalMetrics {
 
 impl EvalMetrics {
     pub fn from_record_batch(batch: &[RecordBatch]) -> anyhow::Result<Self> {
-        if batch.is_empty() || batch[0].num_rows() == 0 {
-            return Err(anyhow::anyhow!("No evaluation metrics found"));
-        }
-        let record = &batch[0];
+        let record = batch.iter().find(|b| b.num_rows() > 0)
```
Jeadie: Shouldn't we collect data from all `RecordBatch` in `&[RecordBatch]`? Not just the first non-empty `RecordBatch`

sgrebnov: We take first one here as there is only a single max row with aggregated results expected, so there should be only one batch with one record.

peasee (dismissed):

peasee:
It got removed from queue for a linting issue. ~~I don't know how it passed in the PR CI :thinking:~~ It passed in the PR because it got skipped :thinking:

Sevenannn (approved):

Issue #5683, opened 2025-05-03
Enhancement: cron based refresh schedule support


sgrebnov:
## Goal-State/What/Result

Accelerated datasets can be refreshed via a cron schedule, which sequentially refreshes datasets.

Workers also support refreshes via cron schedule, and execute based on the defined prompt parameter.

## Why/Purpose

To minimize the load impact of refreshing datasets by scheduling at off-peak hours, and to schedule automatic model updates on refreshed data.

## By When

**Issue/Spec written and reviewed:** 20th May, 2025
**Done-Done:** 26th May, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

* The spicepod specification for `datasets.acceleration` is updated to include `refresh_cron` which specifies a cron schedule to execute a scheduled refresh:

```yaml
datasets:
  - name: my_dataset
    from: s3://my-bucket/my_file.parquet
    acceleration:
      refresh_cron: 0 0 * * * # Daily refresh at midnight
```

* The spicepod specification for `datasets.acceleration` is updated to include `refresh_cron` which specifies a cron schedule to execute a scheduled refresh:

```yaml
datasets:
  - name: my_dataset
    from: s3://my-bucket/my_file.parquet
    acceleration:
      refresh_cron: 0 0 * * * # Daily refresh at midnight
```

Refreshes initiated by cron schedules respect the `dataset_refresh_parallelism`. For datasets containing the same refresh schedule, they are queued to refresh ordered by dataset name.

`refresh_cron` is mutually exclusive with `refresh_check_interval` - only one of each parameter can exist on the same acceleration, but not both. Dataset validation should error when both parameters exist.

* The spicepod specification for `workers` is updated to include a `cron` parameter, and a `params.prompt` parameter. When triggered via a schedule, the worker will execute with the specified prompt:

```yaml
[...]

PR #6016, opened 2025-05-28
feat: Add refresh_cron acceleration parameter, start scheduler on table load

### LINKED ISSUE #5920: Add scheduler configuration for accelerated datasets to spicepod

### LINKED ISSUE #5922: Initialise a scheduler on Runtime startup
Initialise a scheduler on Runtime startup, when schedules are configured.

peasee:
## 📝 Summary

* Adds the `refresh_cron` acceleration parameter for specifying a cron schedule. Currently does not parse the schedule, as it depends on #6005 
* Adds support for starting a `Scheduler` on runtime startup. Schedulers are stored in a vec on the runtime, to support adding schedules for other components as well (like future workers schedules, in #5921)
* Updates the `df.register_table()` function to return a `oneshot::Receiver<()>` to signal when the underlying table is ready. At the moment, this is only passed up from the `create_accelerated_table` call. In #6015 this will be updated to include a signal for when an accelerated table is loaded from a checkpoint.
* Refactors the `Scheduler` to support dynamically adding new schedules and triggers to a running scheduler. This refactor also made the code slightly cleaner/more separated for initialising new schedules and their associated channels.

## 🔗 Related

* Closes #5922
* Closes #5920
* Part of #5683

## 📚 Docs

* Will need to add docs for the `refresh_cron` parameter, and the supported cron expression formats. Also, needs a note for exclusivity of `refresh_cron` and `refresh_check_interval` parameters.

### DIFF
--- a/crates/runtime/src/component/dataset/mod.rs
+++ b/crates/runtime/src/component/dataset/mod.rs
@@ -97,7 +97,7 @@ pub enum Error {
     },
 
     #[snafu(display(
-        "Both a 'refresh_cron' and 'refresh_check_interval' were specified.\nOnly one of these options can be specified at a time.\nFor details, visit: https://spiceai.org/docs/features/data-acceleration/data-refresh"
+        "Both a 'refresh_cron' and 'refresh_check_interval' were specified.\nOnly one of these options can be specified for a given dataset.\nFor details, visit: https://spiceai.org/docs/features/data-acceleration/data-refresh"
     ))]
     MultipleRefreshExpressionSpecified,
 }
--- a/crates/runtime/src/component/dataset/mod.rs
+++ b/crates/runtime/src/component/dataset/mod.rs
@@ -95,6 +95,11 @@ pub enum Error {
         dataset: String,
         missing_component: String,
     },
+
+    #[snafu(display(
+        "Both a 'refresh_cron' and 'refresh_check_interval' were specified.\nOnly one of these options can be specified at a time.\nFor details, visit: https://spiceai.org/docs/features/data-acceleration/data-refresh"
+    ))]
+    MultipleRefreshExpressionSpecified,
 }
 
 pub type Result<T> = std::result::Result<T, Error>;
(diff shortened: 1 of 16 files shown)


crates/runtime/src/component/dataset/mod.rs
```diff
@@ -95,6 +95,11 @@ pub enum Error {
         dataset: String,
         missing_component: String,
     },
+
+    #[snafu(display(
+        "Both a 'refresh_cron' and 'refresh_check_interval' were specified.\nOnly one of these options can be specified at a time.\nFor details, visit: https://spiceai.org/docs/features/data-acceleration/data-refresh"
```
Jeadie:
```suggestion
        "Both a 'refresh_cron' and 'refresh_check_interval' were specified.\nOnly one of these options can be specified for a given table.\nFor details, visit: https://spiceai.org/docs/features/data-acceleration/data-refresh"
```

crates/runtime/src/init/scheduler.rs
```diff
@@ -0,0 +1,92 @@
...
+
+const REFRESH_SCHEDULER_NAME: &str = "refresh_scheduler";
+
+impl Runtime {
+    pub async fn create_dataset_scheduler_if_required(
+        self: Arc<Self>,
+        dataset: Arc<Dataset>,
+    ) -> Result<()> {
+        // TODO: Actually schedule the refresh task for cron - https://github.com/spiceai/spiceai/issues/6015
+        let Some(_refresh_cron) = dataset.refresh_cron() else {
```
Jeadie:
```suggestion
        if dataset.refresh_cron().is_none() {
```

crates/runtime/src/init/scheduler.rs
```diff
@@ -0,0 +1,92 @@
...
+        tracing::debug!("Creating dataset scheduler for dataset: {}", dataset.name);
+        let scheduler_lock = Arc::clone(&self.schedulers);
+        let mut schedulers = scheduler_lock.write().await;
+        let dataset_name = dataset.name.to_string().into();
+        for scheduler in schedulers.iter() {
+            if scheduler
+                .schedules()
+                .await
+                .iter()
+                .any(|schedule| schedule.name() == dataset_name)
```
Jeadie: Shouldn't we only check the `REFRESH_SCHEDULER_NAME` scheduler to see if the dataset schedule exists.

Jeadie (commented):
If we're going to have one scheduler per spice component, it could be worth using a HashMap or struct instead of a Vec<Scheduler>

Jeadie (approved):

PR #6020, opened 2025-05-28
feat: Add search cache configuration and results wrapper

### LINKED ISSUE #5998: Create a search results cache wrapper
Like a `CachedSearchResult`

peasee:
## 📝 Summary

* Implements the cache configuration and wrappers to support a search results cache
  * Although the cache is setup on Runtime startup, it is not currently setup to cache anything/return from cache. This will be implemented in #6019
* Refactors the spicepod module to separate out the cache configuration into its own file
* Refactors the cache module to separate out the different result types into their own files, as well as cache keys

## 🔗 Related

* Closes #5998
* Part of #5997
* Part of #3018

## 📚 Docs

* Docs updates to include `search_results` in cache configuration tracked in https://github.com/spiceai/docs/issues/1018


crates/cache/src/result/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+pub mod query;
+pub mod search;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CacheStatus {
+    // The request was not eligible for caching, and thus the results cache was not checked.
```
phillipleblanc:
```suggestion
    // The request was not eligible for caching, and thus the cache was not checked.
```

crates/runtime/src/http/v1/mod.rs
```diff
@@ -236,7 +236,7 @@ pub async fn to_http_response(
     (StatusCode::OK, headers, body).into_response()
 }
 
-fn attach_cache_headers(headers: &mut HeaderMap, results_cache_status: QueryResultsCacheStatus) {
+fn attach_cache_headers(headers: &mut HeaderMap, results_cache_status: CacheStatus) {
```
phillipleblanc, line 239: Do we need to add additional headers for other cache types?

crates/runtime/src/search/request.rs
```diff
@@ -1,3 +1,4 @@
+use cache::key::SearchKey;
```
phillipleblanc: Move import

phillipleblanc (dismissed):

crates/runtime/src/http/v1/mod.rs
peasee: Good question. I had not considered this, and will need to update the spec. Tracking in https://github.com/spiceai/spiceai/issues/6028, I would like to continue to merge this without additional headers.

peasee:
Changes introduced by search highlighting has changed some of the struct names. I will need to update before this PR is ready.

peasee:
I have updated to use an `AggregationResult`. Connecting this to the actual cache will be completed in #6019

crates/runtime/src/http/v1/mod.rs
phillipleblanc: I don't think we would have separate headers for requesting bypass, only headers that indicate what happened.

phillipleblanc (approved):

Issue #3015, opened 2024-10-08
Enhancement: Full-text Search


lukekim:
## Goal-State/What/Result
Provide full text search candiates, scored by BM25, through the `v1/search` HTTP endpoint. 

## Why/Purpose
Vector search produces search candidates from embedding-based semantic meaning, but can be less precise in circumstances where specific lexical matches can be important (e.g. Country, project name, etc (or partial match)). 

Providing candidates from a full text search experience can provide a more robust and effective out of the box experience for hybrid search.

Improvements in effectiveness can be measured in:
 - NDCG@10 on MTEB search benchmarks: https://github.com/spiceai/spiceai/actions/workflows/benchmarks_search.yml
 - Eval scores for search-based LLM testing, specifically Financebench: https://github.com/spiceai/spiceai/actions/workflows/financebench.yml

## By When

**Issue/Spec written and reviewed:** 14th May 2025
**Done-Done:** 9th June 2025

## Done-Done

- [x] [First Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Threat Model / Security Review
- [x] Tests
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Samples/Quickstarts/Tutorials
- [x] Release Notes

## Spec
Note: See #3016 for details about how candidate generation will be 1. abstracted in code, 2. useable in the `v1/search` experience.

These specification detail how full text search will be integrated for a given dataset.

### Spicepod DX
Columns can be used in full text search by enabling `full_text_search` on a per-column basis.
```yaml
datasets:
 - name: dataset
   from: file:cleaned_sales_data.parquet
   columns:
     - name: product_line
       full_text_search:
         enabled: true
```

This can be used in combination with vector search, by specifying an embedding model for a column.
```yaml
datasets:
 - name: dataset
   from: file:cleaned_sales_data.parquet
   columns:
     - name: product_line
       full_text_search:
         enabled: true
       embeddings:
         - from: hf_minilm

embeddings:
  - name: hf_minilm
    from: huggingface:huggingface.co/sentence-transformers/all-MiniLM-L6-v2
```

Or across different columns.
```yaml
datasets:
 - name: dataset
   from: file:cleaned_sales_data.parquet
   columns:
     - name: customer_name
       full_text_search:
         enabled: true
     - name: product_line
       embeddings:
[...]

PR #6044, opened 2025-05-29
Define SearchPipeline and use in `runtime/vector_search.rs`.

Jeadie:
## 📝 Summary
Remove logic from vector search in runtime to the `search` crate to handle calling `CandidateGeneration` implementations and combing them via a `CandidateAggregation`. 

This looks like
```rust
pub struct SearchPipeline {
    generators: Vec<Box<dyn CandidateGeneration>>,
    aggregator: Box<dyn CandidateAggregation>,
}
impl SearchPipeline {
    pub async fn run(
        &self,
        query: String,
        opt_filters: Vec<Expr>,
        addition_projection: Vec<Expr>,
        primary_keys: Vec<String>,
        keywords: Vec<String>,
        limit: usize,
    ) -> std::result::Result<AggregationResult, Error> {
```

- [Code Link](https://github.com/spiceai/spiceai/blob/jeadie/25-05-28/search-combine/crates/search/src/pipeline.rs#L43-L46)

## 🔗 Related
This will be needed for #3015

### DIFF
--- a/crates/runtime/src/search/vector_search.rs
+++ b/crates/runtime/src/search/vector_search.rs
@@ -170,7 +170,7 @@ impl VectorSearch {
                         );
                     };
 
-                    let agg_result = SearchPipeline::new(generators, Box::new(ReciprocalRankFusion)).run(
+                    let agg_result = SearchPipeline::new(generators, ReciprocalRankFusion).run(
                          query.clone(),
                          where_cond.as_ref().map(|e| vec![e.clone()]).unwrap_or_default(),
                           additional_columns.iter().map(|s| Expr::Identifier(Ident::new(s))).collect(),
--- a/crates/runtime/src/search/vector_search.rs
+++ b/crates/runtime/src/search/vector_search.rs
@@ -18,16 +18,23 @@ use std::{collections::HashMap, sync::Arc};
 
 use super::request::SearchRequest;
 use super::util::{get_embedding_table, user_tables_with_embeddings};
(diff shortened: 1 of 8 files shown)


crates/runtime/src/search/vector_search.rs
```diff
@@ -189,34 +153,37 @@ impl VectorSearch {
...
-                                // *limit
-                            ).await?
+                            ).await?)
                         );
-                    }
-
-                    Ok((tbl.clone(), VectorSearchGenerationTableResult{data: results, primary_keys: primary_keys.to_vec()}))
+                    };
+
+                    let agg_result = SearchPipeline::new(generators, Box::new(ReciprocalRankFusion)).run(
```
phillipleblanc: Is this Box allocation on every request really necessary?

crates/search/src/pipeline.rs
```diff
@@ -0,0 +1,204 @@
...
+///
+/// Also validates keywords against being SQL injections.
+fn prepare_keywords(keywords: &[String], column: &str) -> Result<Vec<Expr>, Error> {
+    keywords
+        .iter()
+        .map(|k| validate_keyword_to_ilike(k, column))
+        .collect::<Result<Vec<Expr>, Error>>()
+}
+
+fn validate_keyword_to_ilike(k: &str, target_column: &str) -> Result<Expr, Error> {
```
phillipleblanc: Let's add unit tests for this.

peasee (approved):

PR #6045, opened 2025-05-29
Upgrade to delta_kernel 0.11

phillipleblanc:
## 📝 Summary

Upgrades `delta_kernel_rs` to 0.11

Changelog: https://github.com/delta-io/delta-kernel-rs/blob/main/CHANGELOG.md


crates/data_components/src/delta_lake.rs
```diff
@@ -741,12 +739,6 @@ async fn get_parquet_access_plan(
...
 }
 
-#[derive(Debug, PartialEq)]
-enum DeltaBinaryOperator {
-    BinaryOperator(BinaryOperator),
-    JunctionOperator(JunctionOperator),
-}
-
 /// Convert a `DataFusion` filter expression to a `delta_kernel` expression
 #[allow(clippy::too_many_lines)]
```
kczimm, line 743: Can this be removed now?

crates/data_components/src/delta_lake.rs
```diff
@@ -819,32 +803,35 @@ fn to_delta_kernel_expr(expr: &Expr) -> Option<Expression> {
     }
 }
 
-fn to_delta_kernel_binary_op(op: Operator) -> Option<DeltaBinaryOperator> {
+fn into_predicate(expr: Expression) -> Option<Predicate> {
```
kczimm, line 806: nit: consider refactoring this to `TryFrom`

```rust
impl TryFrom<Expression> for Predicate {
    type Error = ();

    fn try_from(expr: Expression) -> Result<Self, Self::Error> {
        if let Expression::Predicate(predicate) = expr {
            Ok(*predicate)
        } else {
            Err(())
        }
    }
}
```

kczimm (approved):

crates/data_components/src/delta_lake.rs
phillipleblanc: It's not really an error if its not a predicate, so it feels weird to have to return one. I'll keep the existing approach for now.

crates/data_components/src/delta_lake.rs
phillipleblanc: No, it still needs to be here.

PR #6050, opened 2025-05-29
refactor: Replace refresh oneshot with notify

peasee:
## 📝 Summary

* Replaces the `oneshot` channel which signals dataset refresh with a `sync::Notify` instead:
  * This allows multiple listeners to be notified when a refresh completes
  * Updates the trigger refresh to optionally return this notification channel, so an downstream consumer can decide whether to wait for completion after their triggered refresh
  * Uses this behavior for the Dataset Refresh `ScheduledTask` to ensure that the schedule waits for completion of a refresh before notifying completion.

## 🔗 Related

* Part of #5683


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -454,6 +455,7 @@ impl Refresher {
             initial_load_completed: Arc::new(AtomicBool::new(false)),
             disable_federation: false,
             semaphore: None,
+            notifier: None,
```
Jeadie: Not immediately obvious this notifies on completion. Either a comment or rename the variable to `on_completion_nofication` or similar.

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -922,21 +932,21 @@ mod tests {
...
 
         trigger
             .send(None)
             .await
             .expect("trigger sent correctly to refresh");
 
-        timeout(Duration::from_secs(2), async move {
-            is_ready.await.expect("data is received");
-        })
+        timeout(
```
Jeadie, line 958: I know this isn't new to this PR, but this 2s timeout is to ensure the refresh was received appropriately?

Jeadie (dismissed):

crates/runtime/src/accelerated_table/refresh.rs
peasee: That the refresh was received, completed and notified up the chain that it completed.

crates/runtime/src/scheduling/mod.rs
```diff
@@ -38,20 +38,27 @@ impl From<Arc<Dataset>> for DatasetRefreshTask {
...
+                .await
+            {
+                Ok(notifier) => {
+                    if let Some(notifier) = notifier {
+                        notifier.notified().await;
+                    }
+                }
+                Err(e) => {
+                    // Handle the error
+                    todo!("Handle when refresh fails: {e}");
```
phillipleblanc: Return a proper error instead of panicking

phillipleblanc (approved):

PR #6052, opened 2025-05-30
task: Switch GH runners in PR and testoperator

peasee:
## 📝 Summary

* Updates the PR tests to use [nextest](https://nexte.st/docs/running/). This helps control retries, and also identifies flaky tests in the test output. It automatically adds an exponential delay based on the nextest configuration file to retry flaky tests.
* Updates the PR test workflow to run `make build-cli ...` instead of `make build ...`. `make build` triggers a build of the Runtime in a different profile, which is not reused by the test build. This reduced test times in PRs by ~50%.
* Updates the integration tests (non-models) to use nextest
* Updates the PR checks to use a `lint` profile, instead of building in `release` mode on PRs
* Updates the testoperator to only permit running on the `spiceai-dev-*` runners
  * Also makes all the required updates to the test spicepods for running in the dev environment (different connection secrets, file locations, etc)
* Updates the snapshots for the updated dev environment
* Updates integration tests and check all features to use SCCache when available to reduce build times
* Updates PR builds to use the dev runners.


.github/workflows/pr.yml
```diff
@@ -13,28 +13,29 @@ on:
   workflow_dispatch:
     inputs:
       profile_option:
-        description: "Which Rust build profile to use? (Default is release)"
+        description: "Which Rust build profile to use? (Default is ci)"
         required: false
         type: choice
         options:
+          - "ci"
```
lukekim: Wait, why do we need a new profile?

Cargo.toml
```diff
@@ -175,6 +175,12 @@ x509-certificate = "0.23.1"
 inherits = "release"
 lto = true
 
+[profile.ci]
+inherits = "dev"
```
lukekim, line 180: I think we should resist adding more profiles. We should keep bechmarks/tests as close to release as possible.

peasee: This profile is specifically for PR check workflows, to optimise build time and sccache hits. Release profiles are still used for benchmarks. I will rename it from `ci` to `lint`.

peasee:
* ~~[ ] mysql-duckdb[file] tpch 100 needs a lower duckdb memory_limit~~
* ~~[ ] dremio tpch and tpcds needs setting up~~

I will address these in follow-up PRs.

phillipleblanc (approved):

PR #6053, opened 2025-05-30
Enhance Arrow Flight DoPut operation tracing

sgrebnov:
## 📝 Summary

Simplify DoPut troubleshooting by enhancing Arrow Flight DoPut operation tracing

### DIFF
--- a/crates/runtime/src/flight/do_put.rs
+++ b/crates/runtime/src/flight/do_put.rs
@@ -166,6 +166,7 @@ fn create_response_stream(
     first_message: &FlightData,
 ) -> impl futures::Stream<Item = Result<PutResult, Status>> + use<> {
     let dictionaries_by_id = Arc::new(HashMap::new());
+    tracing::debug!("Starting writing data to path: {path}");
 
     // Sometimes the first message only contains the schema and no data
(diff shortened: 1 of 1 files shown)


crates/runtime/src/flight/do_put.rs
```diff
@@ -166,6 +166,7 @@ fn create_response_stream(
     first_message: &FlightData,
 ) -> impl futures::Stream<Item = Result<PutResult, Status>> + use<> {
     let dictionaries_by_id = Arc::new(HashMap::new());
+    tracing::debug!("Starting writing data to path: {path}");
```
lukekim: Message could be more descriptive.

lukekim (dismissed):

crates/runtime/src/flight/do_put.rs
sgrebnov: Information about the Flight method is included by default as part of the tracing (`DEBUG runtime::flight::do_put: `prefix), so updated the message from `Starting writing data to path: {path}` to `Starting writing data into dataset: {path}`

Jeadie (approved):

Issue #5697, opened 2025-05-05
Bug: Iceberg Table Provider does not reflect new data after append on Iceberg table


phillipleblanc:
## Describe the bug

Start Spice with a dataset that loads an Iceberg table:

```yaml
datasets:
  - from: iceberg:https://glue.ap-northeast-2.amazonaws.com/iceberg/v1/catalogs/211125479522/namespaces/security_logs/tables/audit_log
    name: audit_log
    params:
      iceberg_s3_region: ap-northeast-2
      iceberg_s3_access_key_id: ${ secrets:AWS_ACCESS_KEY_ID }
      iceberg_s3_secret_access_key: ${ secrets:AWS_SECRET_ACCESS_KEY }
      iceberg_s3_role_arn: ${ secrets:AWS_ROLE_ARN }
```

Load additional data into the Iceberg table. Try querying for that new data in Spice and see it doesn't exist. Restart Spice and observe the data is now present.

PR #6054, opened 2025-05-30
Glue catalog connector for hive style parquet

### LINKED ISSUE #5725: Glue Catalog Connector for Hive style Parquet

kczimm:
## 📝 Summary

- Adds the ability to read Hive-style parquet tables from Glue catalog connector
- Moves `ConnectorParams` to a new module.
- Move AWS parameter validation to a separate module for reuse
- Moves AWS config loading to a separate module for reuse with DynamoDB

## 📚 Docs

Docs needed


crates/runtime/src/dataconnector/parameters/aws.rs
```diff
@@ -0,0 +1,242 @@
...
+use aws_credential_types::Credentials;
+use snafu::prelude::*;
+use tonic::async_trait;
+
+use crate::parameters::{ParamLookup, Parameters};
+
+use super::ConnectorParams;
+
+// https://docs.aws.amazon.com/general/latest/gr/rande.html
+pub const AWS_REGIONS: [&str; 32] = [
```
Jeadie, line 27: Is there no similar constant in one of our aws dependencies?

crates/runtime/src/dataconnector/parameters/aws.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    #[snafu(display("Missing access secret"))]
+    NoAccessSecret,
+
+    #[snafu(display("Unsupported authentication method: {method}"))]
+    UnsupportedAuthenticationMethod { method: String },
+}
+
+#[async_trait]
+pub(crate) trait Validator {
```
Jeadie: This trait definition feels like it should be in `crates/runtime/src/dataconnector/parameters.rs`?

Jeadie: And possibly worth a docstring stating that `.validate` is also responsible for mutating `params` appropriately.

crates/runtime/src/dataconnector/parameters/aws.rs
Jeadie: I now realise this is just moved from `crates/runtime/src/dataconnector/s3.rs`.

crates/runtime/src/catalogconnector/glue.rs
```diff
@@ -14,46 +14,67 @@ See the License for the specific language governing permissions and
...
+    .collect()
+});
+
+static VALIDATORS: LazyLock<Vec<Box<dyn Validator + Send + Sync + 'static>>> =
+    LazyLock::new(|| vec![Box::new(RegionValidator), Box::new(AuthValidator)]);
 
 #[derive(Debug, Snafu)]
 pub enum Error {
     #[snafu(display("Failed to get Glue databases: {source}"))]
-    GetDatabases { source: SdkError<GetDatabasesError> },
```
Jeadie, line 53: Is there a reason we're using full imports (e.g. `aws_sdk_glue::error::SdkError`) over importing at top of file?

crates/runtime/src/catalogconnector/glue/provider.rs
```diff
@@ -0,0 +1,475 @@
...
+/// A schema provider for a specific Glue database, providing table metadata.
+pub struct GlueSchemaProvider {
+    database: String,
+    state: Arc<GlueCatalogState>,
+}
+
+impl fmt::Debug for GlueSchemaProvider {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("GlueSchemaProvider")
+            .field("database", &self.database)
```
Jeadie: Could be worth `impl fmt::Debug for GlueCatalogState` and then using here

crates/runtime/src/catalogconnector/glue/provider.rs
```diff
@@ -0,0 +1,475 @@
...
+        for validator in super::VALIDATORS.iter() {
+            validator
+                .validate(&mut parameters)
+                .await
+                .context(super::ParameterValidationSnafu)?;
+        }
+
+        parameters
+            .parameters
+            .insert("file_format".to_string(), "parquet".into());
```
Jeadie, line 78: This isn't a limitation of AWS glue, but rather us constraining ourselves to hive style parquet, correct?

Jeadie (dismissed):

crates/runtime/src/dataconnector/parameters.rs
```diff
@@ -0,0 +1,131 @@
...
+
+pub(crate) mod aws;
+
+#[derive(Clone)]
+pub struct ConnectorParams {
+    pub(crate) parameters: Parameters,
+    pub(crate) unsupported_type_action: Option<UnsupportedTypeAction>,
+    pub(crate) component: ConnectorComponent,
+    pub(crate) app: Option<Arc<App>>,
+    #[allow(dead_code)] // will be utilized in https://github.com/spiceai/spiceai/pull/5542
```
phillipleblanc: Can we either remove this dead_code directive or remove the runtime here if its still not used?

crates/runtime/src/dataconnector/parameters/aws.rs
```diff
@@ -0,0 +1,242 @@
...
+    async fn validate(&self, params: &mut ConnectorParams) -> Result<(), Error>;
+}
+
+pub(crate) struct EndpointValidator;
```
phillipleblanc:
```suggestion
pub(crate) struct S3EndpointValidator;
```

crates/runtime/src/dataconnector/parameters/aws.rs
```diff
@@ -0,0 +1,242 @@
...
+                return Err(Error::UnsupportedAuthenticationMethod {
+                    method: auth.to_string(),
+                });
+            }
+        }
+        Ok(())
+    }
+}
+
+pub async fn load_config(
```
phillipleblanc, line 191: This is a nice refactor

crates/runtime/src/catalogconnector/glue/provider.rs
```diff
@@ -0,0 +1,475 @@
...
+        table: &Table,
+    ) -> DFResult<Option<Arc<dyn TableProvider>>> {
+        let mut props = Vec::new();
+        if let Some(provider) = self.state.config.credentials_provider() {
+            let creds = provider
+                .provide_credentials()
+                .await
+                .map_err(|e| DataFusionError::External(Box::new(e)))?;
+            props.push((S3_ACCESS_KEY_ID, creds.access_key_id().to_string()));
+            props.push((S3_SECRET_ACCESS_KEY, creds.secret_access_key().to_string()));
```
phillipleblanc: If the creds are temporary, then this will stop working once they expire.

phillipleblanc: Ideally we let the underlying provider have the same inputs as the glue connector itself, i.e. only if an explicit key_id/secret_access_key is provided then we pass that

crates/runtime/src/catalogconnector/glue/provider.rs
```diff
@@ -0,0 +1,475 @@
...
+        let table = IcebergTable::builder()
+            .file_io(file_io)
+            .metadata(metadata)
+            .identifier(identifier)
+            .build()
+            .map_err(|e| {
+                DataFusionError::External(Box::new(super::Error::BuildIcebergTable { source: e }))
+            })?;
+
+        let table_provider = IcebergTableProvider::try_new_from_table(table)
```
phillipleblanc: This needs to use the variant of the IcebergTableProvider that takes in a catalog, otherwise it will not get updates as the table itself is updated.

phillipleblanc: See https://github.com/spiceai/spiceai/issues/5697

crates/runtime/src/catalogconnector/glue.rs
```diff
@@ -14,46 +14,67 @@ See the License for the specific language governing permissions and
...
+
+pub static PREFIX: &str = "glue";
+
+pub static PARAMETERS: LazyLock<Vec<ParameterSpec>> = LazyLock::new(|| {
+    vec![
+        ParameterSpec::component("region")
+            .description("The AWS region for Glue operations")
+            .secret(),
+        ParameterSpec::component("key")
+            .description("The AWS access key ID for Glue authentication")
```
phillipleblanc: I don't think we should have separate auth parameters for Glue and S3. We explicitly decided against that for the Iceberg connector that uses Glue, with the assumption that most deployments will configure the same auth to connect to both Glue and S3.

crates/runtime/src/catalogconnector/glue/provider.rs
```diff
@@ -0,0 +1,475 @@
...
+                .unwrap_or_default()
+                .into_iter()
+                .filter(|t| {
+                    !matches!(TableType::from(t), TableType::Unsupported)
+                        && is_included(catalog.include.as_ref(), &db.name, t.name())
+                })
+                .collect::<Vec<_>>();
+
+            if !table_names.is_empty() {
+                databases.insert(db.name, table_names);
```
Sevenannn: Is it by design that a schema with no tables won't be registered in spice?

crates/runtime/src/catalogconnector/glue/provider.rs
kczimm: `file_format` is required early for `ListingConnector` which the S3 connector uses. We can change the file format when we create `TableProvider`s if we need to.

crates/runtime/src/catalogconnector/glue/provider.rs
kczimm: I need to create a `GlueCatalogConfig`, https://rust.iceberg.apache.org/api/iceberg_catalog_glue/struct.GlueCatalogConfig.html#method.builder. It requires specifying a `warehouse`. It's not clear what value this should be. Do you know? My best guess is it's where the metadata for the catalog are stored but I don't know how to find that and I don't see why this would need to be explicitly set. Any insight here would be appreciated.

crates/runtime/src/catalogconnector/glue/provider.rs
kczimm: I think this should be allowed. Thanks!

crates/runtime/src/catalogconnector/glue/provider.rs
phillipleblanc: You can set it to the location of one of the tables, its used as the base path for creating new tables, which we aren't doing.

crates/runtime/src/catalogconnector/glue/provider.rs
```diff
@@ -0,0 +1,425 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("GlueCatalogProvider")
+            .finish_non_exhaustive()
+    }
+}
+
+/// A schema provider for a specific Glue database, providing table metadata.
+#[derive(Debug)]
+pub struct GlueSchemaProvider {
+    database: String,
```
sgrebnov: Is `database` here a name  - 
It seems we have `type DatabaseName = String;` - we should use `DatabaseName` instead of String vs remove extra type IMO.

crates/runtime/src/catalogconnector/glue/provider.rs
sgrebnov: Should this be checking if there is file_format already provided and default to `parquet` if not?

crates/runtime/src/catalogconnector/glue/provider.rs
```diff
@@ -0,0 +1,425 @@
...
+                tracing::debug!("skipping database {}", &db.name);
+                continue;
+            }
+
+            let get_tables_output = glue
+                .get_tables()
+                .database_name(&db.name)
+                .send()
+                .await
+                .context(super::GetTablesSnafu)?;
```
sgrebnov: Will error contain specific db.name failed?  If not we should pass db.name to super::GetTablesSnafu) to simplify future troubleshooting

crates/runtime/src/catalogconnector/glue/state.rs
```diff
@@ -0,0 +1,50 @@
...
+*/
+
+use super::DatabaseName;
+use crate::{Runtime, dataconnector::parameters::ConnectorParams};
+use aws_sdk_glue::types::Table;
+use std::collections::HashMap;
+use std::fmt;
+use std::sync::Arc;
+
+pub struct GlueCatalogState {
```
sgrebnov, line 24: Not sure `State` is the right term here as it is not changed (??) - more of GlueCatalogInner or GlueCatalogContext

sgrebnov (dismissed):

crates/runtime/src/catalogconnector/glue/state.rs
kczimm: Funny you should say that. I had it as Inner and got feedback to change it. But I'm working on the refresh method for `RefreshableCatalogProvider` and that will update the state. Otherwise, I agree with you.

sgrebnov (approved):

PR #6056, opened 2025-05-31
Add initial support for Cloud Platform management to export task_history records

sgrebnov:
## 📝 Summary

Adds `management` extension that exports `task_history` records to the Spice.ai Cloud Platform using Arrow Flight endpoint and `spice.ai` data connector. 
1. Records are exported every 5 seconds, and the last successful export time is tracked to avoid sending duplicate records. Since there can be a delay between when an event occurs and when it is written into the `task_history` table (tracing/telemetry exporting), an additional 5-second buffer is used as a window to include new records (based on `end_time`). This can be further improved by operating on the maximum known `end_time` instead of system time, which would eliminate the need for the the buffer. Note: task_history write is not atomic and includes additional spans rewrite step after main flush so small extra window must still be used to ensure final version of the records is exported (alternative approach is to add additional delay before new records can be send - upper time limit).
1. `start_runtime_task` is used for graceful shutdown and to ensure that most recent records are written. Alternative approach could be adding additional `on_shutdown` method support to Extension API

```
extensions:
  management:
    enabled: true
    params:
      api_key: ... # Spice.ai Cloud Platform api-key
      flight_endpoint: https://dev-flight.spiceai.io
```

## 🔗 Related

Part of https://github.com/spiceai/spiceai/issues/5269

## 👀 Notes for Reviewers

1. The following implementation alternatives were considered as well
 - Use locally accelerated version of remote table with `replication` enabled (original metrics implementation). Downside of this approach is that this adds unnecessary synchronization overhead with remote table version and is unnecessary.
 - Creating accelerated table pointing to local task history as source and using remote table provider as accelerator. This allows to configure append acceleration based on local task_history `end_time` column out of the box. Disadvantages of this approach are: append implementation is based on retrieving last know max time for time column, as a result every append will first perform cloud platform query to calculate max known last time and only do write after that. This also provides less control over refresh/write, for example final refresh/flush when application is shutdown. This will also add additional complexity when we add `instance_id` part so local history and remote history tables won't match so we will need to add additional wrapper/logic.
3. Plan to review if new cron functionality could be re-used instead of creating managed tokio spawned task.

### DIFF
--- a/crates/spice_cloud/src/lib.rs
+++ b/crates/spice_cloud/src/lib.rs
@@ -14,34 +14,36 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use std::{sync::Arc, time::Duration};
+use std::{
+    collections::HashMap,
+    sync::Arc,
+    time::{Duration, SystemTime},
+};
 
 use async_trait::async_trait;
-use datafusion::{datasource::TableProvider, sql::TableReference};
-use serde::{Deserialize, Serialize, de::DeserializeOwned};
-use serde_json::json;
+use chrono::{DateTime, Utc};
+use datafusion::{arrow::array::RecordBatch, datasource::TableProvider};
+use futures::TryStreamExt;
 use snafu::{ResultExt, prelude::*};
 
 use runtime::{
     Runtime,
-    accelerated_table::{
-        AcceleratedTable, AcceleratedTableBuilderError, Retention, refresh::Refresh,
-    },
-    component::dataset::{
-        Mode, TimeFormat,
-        acceleration::{Acceleration, RefreshMode},
-        builder::DatasetBuilder,
-        replication::Replication,
-    },
-    dataaccelerator::{self, AcceleratorEngineRegistry},
+    accelerated_table::AcceleratedTableBuilderError,
+    component::dataset::{Mode, builder::DatasetBuilder},
+    dataaccelerator::{self},
     dataconnector::{ConnectorParamsBuilder, DataConnectorError, create_new_connector},
+    datafusion::DataFusion,
+    dataupdate::{DataUpdate, UpdateType},
     extension::{Error as ExtensionError, Extension, ExtensionFactory, ExtensionManifest, Result},
-    federated_table::FederatedTable,
-    secrets::{ExposeSecret, Secrets},
-    spice_metrics::get_metrics_table_reference,
-    status,
+    secrets::Secrets,
 };
 use tokio::sync::RwLock;
+use tokio_util::sync::CancellationToken;
+
+const TASK_HISTORY_SINK_REMOTE_TABLE: &str = "runtime.task_history";
+const TASK_HISTORY_SINK_TABLE: &str = "runtime.scp_task_history_sink";
+const DEFAULT_EXPORT_INTERVAL_SECS: u64 = 5;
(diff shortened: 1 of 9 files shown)


crates/spice_cloud/src/lib.rs
```diff
@@ -66,175 +68,184 @@ pub enum Error {
...
     }
 
-    async fn connect(&self) -> Result<SpiceCloudConnectResponse, Error> {
-        self.post_json("/v1/connect", json!({})).await
+    async fn export_task_history_records(df: &Arc<DataFusion>, since: SystemTime) -> bool {
+        // We fetch records and then insert them into the sink table separatly as
+        // Insert Into is not allowed via runtime settings, so we can't do 'INSERT INTO .. SELECT * FROM ..'
+        let data = match get_task_history_records(df, since).await {
+            Ok(records) => records,
+            Err(e) => {
```
sgrebnov, line 141: Task history table could be not accelerated/accelerated at this point, this should NOT be traced

crates/spice_cloud/src/lib.rs
```diff
@@ -325,63 +337,57 @@ async fn get_spiceai_table_provider(
...
-struct SpiceCloudConnectResponse {
-    org_name: String,
-    app_name: String,
-    metrics_dataset_name: String,
+async fn get_task_history_records(
+    df: &Arc<DataFusion>,
+    since: SystemTime,
+) -> Result<Vec<RecordBatch>, Error> {
+    let sql = format!(
+        "SELECT * FROM runtime.task_history WHERE end_time >= '{}'",
```
sgrebnov: Will switch to table manipulation - scan + filter. Will also review if we can combine SELECT and INSERT INTO

Jeadie:
> Plan to review if new cron functionality could be re-used instead of creating managed tokio spawned task.

I think this is true. But we should also include in this, considering how we want to change/improve `Runtime.runtime_tasks` to best use the new generic cron functionality.

crates/spice_cloud/src/lib.rs
```diff
@@ -66,175 +79,190 @@ pub enum Error {
...
-            return false;
-        }
-
+    fn get_flight_url(&self) -> String {
         self.manifest
             .params
-            .get("metrics")
-            .map_or_else(|| false, |v| v.eq_ignore_ascii_case("true"))
+            .get("flight_endpoint")
+            .unwrap_or(&"https://flight.spiceai.io".to_string())
```
Jeadie, line 115: Should this be like the Spice.ai data connector?
```rust
let default_flight_url: Arc<str> = if cfg!(feature = "dev") {
        "https://dev-flight.spiceai.io".into()
    } else {
        "https://flight.spiceai.io".into()
    };
```

[source](https://github.com/spiceai/spiceai/blob/726e86526d38e1c7f0b51aef66a7452be81a33c1/crates/runtime/src/dataconnector/spiceai.rs#L173-L177)

Jeadie (dismissed):

crates/spice_cloud/src/lib.rs
sgrebnov: Thank you, to be updated to match default behavior

sgrebnov:
Closed in favor of 
https://github.com/spiceai/spiceai/pull/6089

PR #6069, opened 2025-06-02
feat: Add a workers integration test

peasee:
## 📝 Summary

* Adds an integration test for a worker configured with a cron schedule
* While adding this integration test, identified a bug in the way workers are registered for ready status preventing the runtime from ever reporting ready if there was a worker defined.

## 🔗 Related

* Part of #5683


sgrebnov (approved):

phillipleblanc (approved):

crates/runtime/tests/models/mod.rs
```diff
@@ -189,7 +189,7 @@ fn get_small_clickbench_dataset(name: &str) -> Dataset {
     dataset
 }
 
-fn get_tpcds_dataset(
+pub fn get_tpcds_dataset(
```
sgrebnov, line 192: Nit: might be good to move into some utils/general purpose module from models if we are using it across different components

sgrebnov: Pls ignore - noted that this version is with embeddings - models/search specific so should stay here.

PR #6071, opened 2025-06-02
Fix DuckDB acceleration `ORDER BY rand()` and `ORDER BY NULL`

phillipleblanc:
## 📝 Summary

Fixes DuckDB acceleration to accept `ORDER BY rand()` (rewritten to `ORDER BY random()`) and `ORDER BY NULL` (ignored via the DuckDB setting `SET order_by_non_integer_literal = true`).

- This PR depends on https://github.com/datafusion-contrib/datafusion-table-providers/pull/362

DuckDB configuration has two flavors, GLOBAL or LOCAL. The current DuckDB settings were all global settings (i.e. `memory_limit`, `temp_directory`, `preserve_insertion_order`). It is sufficient for global settings to only set them once and have them persist across all connections.

For local configuration settings, they need to be applied on each connection. This required some plumbing in the table-providers repo, but the end result is that all we need to do now is set the proper `scope` on a `DuckDBSetting` trait object and it will do the correct thing.


Cargo.toml
```diff
@@ -184,7 +184,7 @@ datafusion-expr = { git = "https://github.com/spiceai/datafusion.git", rev = "b5
 datafusion-datasource = { git = "https://github.com/spiceai/datafusion.git", rev = "b5c62f29d2c70c5331ff50015b67b5e1cafcd578" } # spiceai-47
 
 datafusion-federation = { git = "https://github.com/spiceai/datafusion-federation.git", rev = "9db74a4b360df6be1bb554c59a474a2fd4bfb7e9" } # spiceai-47
-datafusion-table-providers = { git = "https://github.com/datafusion-contrib/datafusion-table-providers.git", rev = "d53510efd7b9ffb8d7f87e63df33ca0a5756e234" } # spiceai
+datafusion-table-providers = { git = "https://github.com/datafusion-contrib/datafusion-table-providers.git", rev = "b55edf560d369607567fed27f7027fa375c18c05" } # phillip/250602-order-by-null
```
phillipleblanc: Will update once https://github.com/datafusion-contrib/datafusion-table-providers/pull/362 merges

ewgenius (approved):

PR #6081, opened 2025-06-02
Retry on transient flight query failure

### LINKED ISSUE #4700: Bug: Spice.ai Data Connector drops connections during high load
## Describe the bug

The Spice.ai Data Connector drops connections during the TPCDS load test: https://github.com/spiceai/spiceai/actions/runs/15366171460/job/43239358295, failing with the following errors:
`FAIL - Worker 0 - Query 'tpcds_q40' failed: Tonic error: status: InvalidArgument, message: "grpc-status header missing, mapped from HTTP status code 400", details: [], metadata: MetadataMap { headers: {} }`
`FAIL - Worker 0 - Query 'tpcds_q13' failed: Tonic error: status: InvalidArgument, message: "operation was canceled", details: [], metadata: MetadataMap { headers: {} }`
`FAIL - Worker 3 - Query 'tpcds_q33' failed: Tonic error: status: InvalidArgument, message: "Failed to connect to Flight server: Handshake failed.\nAn internal error occurred.\nh2 protocol error: http2 error\ntransport error", details: [], metadata: MetadataMap { headers: {} }`
`FAIL - Worker 3 - Query 'tpch_q19' failed: Tonic error: status: InvalidArgument, message: "h2 protocol error: http2 error", details: [], metadata: MetadataMap { headers: {} }`

## To Reproduce

1. Run TPCDS queries with the Spice.ai Data Connector in load test
2. Observe dropped connections starting from a few minutes after test started,

## Expected behavior

Connections should not drop during load.

### Spicepod

### Output of `explain query`

E.g. `explain select 1`

```bash
```

### spice, spiced, OS info

- OS info: `uname -a`

- [ ] Yes
- [ ] No

Sevenannn:
## 📝 Summary

Retry on transient query failure for Spice Cloud Data Connector in flight client.

Spice cloud data connector is subject to the `http2_max_requests` limit in cloud platform, which is defaulted to 1000. The `http2_max_requests` specifies the number of requests can be made through a http2 connection before that connection is closed. For a `FlightClient::query`, it performs 3 requests (handshake, getflightinfo, doget), which means an existing connection will be dropped after ~300 queries. Although any new RPC call will initiate a new connection after the old connection completely failed as per implementation of tonic, rpc calls at the ending of old connection may suffer race condition and fail.

This PR implements retry on `FlightClient::query` and resolves the transient query error caused by connection getting dropped due to server restrictions.

Before - Connections constantly getting dropped a few minutes after test started
https://github.com/spiceai/spiceai/actions/runs/15366171460/job/43239358295
After - Errors related to connection getting dropped no longer observed
https://github.com/spiceai/spiceai/actions/runs/15366440617/job/43325781125

## 🔗 Related

- Close: https://github.com/spiceai/spiceai/issues/4700

## 👀 Notes for Reviewers

- This change could potentially be made only for Spice Cloud Connector instead of applying to all components using the `FlightClient`.
- There are separate errors detected, including `502 bad gateway`, `received message with invalid compression flag: 101` after this bug is fixed: https://github.com/spiceai/spiceai/actions/runs/15366440617/attempts/2. I've triggered additional run to confirm if the new issue is due to some transient factor (e.g. spice cloud service redeployed) and will report / fix it in separate ticket if it persist


crates/flight_client/src/lib.rs
```diff
@@ -387,6 +389,24 @@ impl FlightClient {
...
     pub async fn query(&self, query: &str) -> Result<FlightRecordBatchStream> {
+        // Server may configured with a http2_max_requests limit
+        // Which can causes transient error due to connection being closed and reopened.
+        // Use a retry strategy with exponential backoff to handle transient errors.
+        let retry_strategy = FibonacciBackoffBuilder::new().max_retries(Some(5)).build();
+
+        retry(retry_strategy, || async {
+            match self.query_internal(query).await {
+                Ok(stream) => Ok(stream),
+                Err(e) => Err(RetryError::transient(e)),
```
peasee: Every query error may not be transient. If a query error is due to a legitimate incorrect query, a user now has to wait for 5 retries for the response?

We should ensure we only retry when the error is due to a connection error.

crates/flight_client/src/lib.rs
```diff
@@ -387,6 +389,24 @@ impl FlightClient {
...
+            }
+        })
+        .await
+        .map_err(|e| Error::UnableToQuery {
+            source: Box::new(e),
+        })
```
peasee:
```suggestion
```

Why is this error getting mapped? `query_internal` returns a `Result<>` we can use as-is

crates/flight_client/src/lib.rs
```diff
@@ -636,3 +639,12 @@ fn map_tonic_error_to_message(e: tonic::Status) -> Error {
         source: e.message().into(),
     }
 }
+
+fn is_transient_error(error: &Error) -> bool {
```
phillipleblanc, line 643: Is there an underlying error type that we can downcast to that would make this more robust?

crates/flight_client/src/lib.rs
Sevenannn: Thank you, I updated to match the transient error based on the error string given that there's no specific error code / variant to match on

crates/flight_client/src/lib.rs
peasee: > Is there an underlying error type that we can downcast to that would make this more robust?

There's an example of this in the `test-framework` that could be useful here: https://github.com/spiceai/spiceai/blob/trunk/crates/test-framework/src/spicetest/datasets/worker.rs#L367-L371

Sevenannn: I've updated to match on a concrete error type when possible - some of the trasient error belongs to the `UnableToQuery` variant and cannot be matched without matching on substring. See the failure messages from the following. Open to feedbacks for better ways than matching on error strings.
`FAIL - Worker 0 - Query 'tpcds_q40' failed: Tonic error: status: InvalidArgument, message: "grpc-status header missing, mapped from HTTP status code 400", details: [], metadata: MetadataMap { headers: {} }`
`FAIL - Worker 0 - Query 'tpcds_q13' failed: Tonic error: status: InvalidArgument, message: "operation was canceled", details: [], metadata: MetadataMap { headers: {} }`
`FAIL - Worker 3 - Query 'tpch_q19' failed: Tonic error: status: InvalidArgument, message: "h2 protocol error: http2 error", details: [], metadata: MetadataMap { headers: {} }`

crates/flight_client/src/lib.rs
```diff
@@ -387,6 +389,27 @@ impl FlightClient {
     ///
     /// Returns an error if the query fails.
     pub async fn query(&self, query: &str) -> Result<FlightRecordBatchStream> {
+        // Server may configured with a http2_max_requests limit
+        // Which can causes transient error due to connection being closed and reopened.
```
kczimm, line 393:
```suggestion
        // Server may be configured with a http2_max_requests limit
        // Which can cause a transient error due to the connection being closed and reopened.
```

crates/flight_client/src/lib.rs
kczimm: I think we should change to:

```rust
    #[snafu(display("Query execution failed.\n{source}"))]
    UnableToQuery {
        status: tonic::Status,
    },
```

then I think you should be able to `matches!(status.code, Code::Cancelled | Code::Internal | ...)`

Sevenannn:
- Closed in favor of the new approach of emitting a retry error

PR #6082, opened 2025-06-02
Revert: fix: Use HTTPS ubuntu sources

Sevenannn:
## 📝 Summary

The setup-cc action is failing due to runners get 403 for accessing some https source in the setup cc action. This PR fix the issue by reverting the changes of [using https source](https://github.com/spiceai/spiceai/pull/6034) and use http source instead.
```
E: Failed to fetch https://archive.ubuntu.com/ubuntu/pool/main/p/pkg-config/pkg-config_0.29.2-1ubuntu3_amd64.deb  403  Forbidden [IP: 185.125.190.83 443]
Fetched 85.8 MB in 5s (18.9 MB/s)
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```
https://github.com/spiceai/spiceai/actions/runs/15400546703/job/43331864334?pr=6081
https://github.com/spiceai/spiceai/actions/runs/15401007890/job/43333362744?pr=6082


.github/actions/setup-cc/action.yml
```diff
@@ -8,4 +8,4 @@ runs:
       shell: bash
       run: |
         sudo apt-get update
-        sudo apt-get install -y build-essential cmake pkg-config libssl-dev unixodbc unixodbc-dev protobuf-compiler
\ No newline at end of file
+        sudo apt-get install --fix-missing -y build-essential cmake pkg-config libssl-dev unixodbc unixodbc-dev protobuf-compiler
```
sgrebnov: @Sevenannn  - could you please specify why `--fix-missing` is required or is it optional/nth? Do we also know why we get 403 for https? Note: if we think there could be similar case in the future we can rename current action to be more generic/flexible - like `set-linux-repo-source` so we can change/control http vs https in the future.

Sevenannn: Thank you for reporting, I've updated this pr to explicitly revert previous change to use https source. The 403 is likely caused by outdated package indexes in https ubuntu source, which is not in our control

.github/actions/update-http-sources/action.yml
peasee, line 1: We should keep this action in case we need it again (if Ubuntu goes down again).

peasee (commented):
`features.yml`, `testoperator_dispatch.yml` and `testoperator_run_bench.yml` also need this removed.

.github/actions/update-http-sources/action.yml
sgrebnov: +1 to keep this: https://github.com/spiceai/spiceai/pull/6082#discussion_r2122219818

Sevenannn: Thanks, I added back the action and added a note for the action usage

peasee (approved):

PR #6089, opened 2025-06-02
Add initial support for Spice Cloud Platform management

sgrebnov:
## 📝 Summary

PR adds initial support for Spice.ai Cloud Platform management, enabling observability based on task history records for Spice instances running outside the Spice.ai Cloud Platform. This allows monitoring and troubleshooting capabilities for on-premises and cluster environments.

```yaml
management:
  enabled: true
  params:
    api_key: ... # Spice.ai Cloud Platform api-key
    flight_endpoint: https://dev-flight.spiceai.io
```

1. Records are exported every 5 seconds and include data from the past 3 days.
2. If the management module fails to initialize or start, an error message is logged, but it does not block the runtime from starting.

## 🔗 Related

Part of https://github.com/spiceai/spiceai/issues/5269

## 👀 Notes for Reviewers

We might consider moving functionality to own crate/module. Keeping directly in runtime currently for simplicity.

### DIFF
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -808,8 +808,8 @@ impl Runtime {
             .as_ref()
             .and_then(|app| app.management.as_ref())
         {
-            if let Err(err) = management::init_cloud_management(Arc::clone(&self), cfg).await {
-                tracing::error!("Failed to initialize Spice Cloud management: {err}");
+            if let Err(err) = management::init_management(Arc::clone(&self), cfg).await {
+                tracing::error!("Failed to initialize management of the Spice runtime: {err}");
             }
         }
 
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -88,6 +88,7 @@ pub mod flight;
 mod http;
 mod init;
 pub mod internal_table;
+mod management;
 mod metrics;
 mod metrics_server;
 pub mod model;
@@ -799,6 +800,18 @@ impl Runtime {
                 }
             }
         });
+
+        if let Some(cfg) = self
+            .app
+            .read()
+            .await
+            .as_ref()
+            .and_then(|app| app.management.as_ref())
+        {
+            if let Err(err) = management::init_cloud_management(Arc::clone(&self), cfg).await {
+                tracing::error!("Failed to initialize Spice Cloud management: {err}");
+            }
+        }
 
         let components = vec![task_history, datasets, catalogs, models_and_evals];
 
--- a/crates/runtime/src/management/mod.rs
+++ b/crates/runtime/src/management/mod.rs
@@ -1,0 +1,372 @@
+/*
+Copyright 2024-2025 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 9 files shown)


crates/runtime/src/lib.rs
```diff
@@ -800,6 +801,18 @@ impl Runtime {
...
         });
 
+        if let Some(cfg) = self
+            .app
+            .read()
+            .await
+            .as_ref()
+            .and_then(|app| app.management.as_ref())
+        {
+            if let Err(err) = management::init_cloud_management(Arc::clone(&self), cfg).await {
```
```suggestion
            if let Err(err) = management::init_management(Arc::clone(&self), cfg).await {
```

crates/runtime/src/lib.rs
```diff
@@ -800,6 +801,18 @@ impl Runtime {
...
+                tracing::error!("Failed to initialize Spice Cloud management: {err}");
```
```suggestion
                tracing::error!("Failed to initialize management of the Spice runtime: {err}");
```

crates/runtime/src/management/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+    datasource::DefaultTableSource,
+    error::DataFusionError,
+    execution::SessionStateBuilder,
+    logical_expr::LogicalPlanBuilder,
+    prelude::{DataFrame, SessionContext, col, lit},
+    sql::TableReference,
+};
+use secrecy::{ExposeSecret, SecretBox, SecretString};
+use snafu::{OptionExt, ResultExt, Snafu};
+use spicepod::component::management::Management as spicepod_management;
```
phillipleblanc:
```suggestion
use spicepod::component::management::Management as SpicepodManagement;
```

crates/runtime/src/management/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+}
+
+/// Spice Cloud Management
+///
+/// Implements runtime management and observability for Spice instances running
+/// outside of the Spice Cloud Platform, enabling monitoring and troubleshooting capabilities
+/// across on-premises and cluster environments.
+pub(crate) struct Management {
+    runtime: Arc<Runtime>,
+    api_key: String,
```
phillipleblanc:
```suggestion
    api_key: SecretString,
```

crates/runtime/src/management/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+        drop(app_lock);
+
+        self.init_task_history_sink_table().await?;
+
+        let cancellation_token = CancellationToken::new();
+        let df = self.runtime.datafusion();
+
+        let _task = self
+            .runtime
+            .start_runtime_task(
```
phillipleblanc, line 146: Would this be a good fit for this? https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/cancellable_task.rs

phillipleblanc (dismissed):

crates/runtime/src/lib.rs
sgrebnov: Updated, thank you

crates/runtime/src/management/mod.rs
sgrebnov: Updated, thank you

crates/runtime/src/management/mod.rs
sgrebnov: Updated, thank you

crates/runtime/src/management/mod.rs
sgrebnov: Yeah 👍 — `start_runtime_task` is based on the `spawn_cancellable_task` functionality, with additional logic to automatically call cancellation when the runtime is shutting down. This way, the task spawned here is gracefully shut down, including a finalization/cancellation signal that is used to perform the final export.

crates/runtime/src/lib.rs
sgrebnov: Updated, thank you

Jeadie (approved):

Issue #6036, opened 2025-05-29
Enhancement: Worker supports an SQL worker action


peasee:
## Goal-State/What/Result

A `sql` worker type is supported, which executes the specified SQL.

## Why/Purpose

To enable scheduled actions that perform SQL queries.

## By When

**Issue/Spec written and reviewed:** May 29th, 2025
**Done-Done:** May 30th, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

An `sql` worker type is supported, which supports a `sql:` parameter and `cron:` parameter to schedule the SQL at particular times:

```yaml
workers:
  - name: my_worker
    cron: 0 * * * *
    sql: "SELECT * FROM lineitem"
```

Additionally, the `type:` parameter for workers is to be removed. The type should be inferred by the available parameters. For example, `sql:` infers that the worker is an SQL worker. `load_balance:` infers a model load balancing/routing worker.

### Security Review

* Specifying an SQL query to be executed on a schedule inside the `spicepod.yml` could expose the system to data exfiltration.
  * Users should ensure strict access controls to the `spicepod.yml` file, preventing access from unauthorised users.
  * The Spice Runtime should run with minimum access, as a user account with low privileges and minimal access to files.

## How/Implementation Plan

* The `Worker` spicepod model is updated to include the `sql` parameter.
* An SQL worker is scheduled when specified, and runs the specified SQL on a cron schedule using the existing scheduler pattern.

## QA Plan

* An integration test is added to perform an SQL query on a cron expression.

## Release Notes

* **SQL Worker Type:** The Runtime now supports specifying `sql` workers, to perform automated SQL queries on a cron schedule:

```yaml
workers:
  - name: my_worker
    cron: 0 * * * *
    sql: "SELECT * FROM lineitem"
```

For more details, refer to the [Workers documentation](#).

PR #6093, opened 2025-06-03
feat: Add SQL worker action

peasee:
## 📝 Summary

* Adds the `sql:` property for workers, to specify an SQL worker action
* Updates the scheduler creation for workers to move the schedule parameters into the `Worker` trait, as `scheduler_parameters()`. A `Worker` that returns `None` is a worker that cannot be scheduled.
* Adds an integration test for the SQL worker action
* Adds the scheduled action for SQL workers, which executes the specified SQL only when the runtime is ready (e.g. to avoid scheduling while datasets are accelerating).

## 🔗 Related

* Part of #6036

## 📚 Docs

* Worker documentation will require updates with this new worker action


crates/runtime/src/scheduling/worker/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+
+            if !runtime.status.is_ready() {
+                tracing::debug!("Runtime is not ready, skipping worker SQL task execution for worker: {worker_name}");
+                return Ok(());
+            }
+
+            let workers_lock = Arc::clone(&runtime.workers);
+            let workers = workers_lock.read().await;
+            // we don't actually need the worker, but validate that it still exists
+            if workers.get(&worker_name.to_string()).is_none() {
```
Jeadie:
```suggestion
            if !workers.contains_key(&worker_name.to_string()) {
```

crates/runtime/src/scheduling/worker/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+                tracing::debug!("Worker not found for ScheduledTask: {worker_name}");
+                return Ok(());
+            }
+
+            tracing::debug!("Executing worker SQL task for worker: {worker_name}, SQL: {sql}");
+
+            let df = runtime.datafusion();
+            if let Err(e) = run_sql(df, &sql, None).await {
```
Jeadie, line 148: Should we add a  captured_output for the result of this, otherwise it might be tough on observability.

crates/runtime/src/worker.rs
```diff
@@ -105,4 +198,46 @@ impl Worker for LoadBalanceWorker {
...
+
+    fn schedule_parameters(&self) -> Option<WorkerScheduleParameters> {
+        self.schedule_parameters.clone()
+    }
+}
+
+pub struct SQLWorker {
+    name: String,
+    description: Option<String>,
+    schedule_parameters: Option<WorkerScheduleParameters>,
```
Jeadie, line 212: We might want (possibly not now), a middleware style pattern to add worker scheduling to any worker (similar to how we process tools for arbitrary LLMs).

crates/runtime/src/worker.rs
```diff
@@ -105,4 +198,46 @@ impl Worker for LoadBalanceWorker {
...
+    fn name(&self) -> Cow<'_, str> {
+        self.name.clone().into()
+    }
+
+    fn description(&self) -> Option<Cow<'_, str>> {
+        self.description.as_ref().map(Into::into)
+    }
+
+    fn as_model(self: Arc<Self>) -> Option<Arc<dyn Chat>> {
+        None // SQLWorker does not implement Chat
```
Jeadie: Add as default function to `Worker`?

crates/runtime/tests/workers/mod.rs
```diff
@@ -95,7 +108,14 @@ async fn test_worker_with_cron() -> Result<(), anyhow::Error> {
 
             runtime_ready_check(&rt).await;
 
-            tokio::time::sleep(std::time::Duration::from_secs(40)).await; // wait for the cron job to run at least once
+            let second_now = chrono::Utc::now().second();
+            let wait = if second_now < 30 {
```
Jeadie: What't the minimum duration this could be?

phillipleblanc (dismissed):

crates/runtime/tests/workers/mod.rs
peasee: it could technically be zero, if the current second is the 30th or 0th second. This calculates the shortest distance to either the 30th or 0th so the test waits the smallest amount (and prevents a 2nd cron run by waiting too long)

crates/runtime/src/worker.rs
peasee: Yeah, I had a similar thought. I'm not quite sure what they would look like yet though, which is why I've landed on this solution.

crates/runtime/src/scheduling/worker/mod.rs
peasee: The output is captured by the underlying `sql_query` call. There's an example of this in the snapshot: https://github.com/spiceai/spiceai/pull/6093/files#diff-5ae4e1681777752d02e443daf00949e1ded70b0c2924c9cc78f713a5000582cfR10

If outputs are being captured, they'll show up in the trace with the `sql_query` associated call.

phillipleblanc (dismissed):

crates/runtime/src/scheduling/worker/mod.rs
Jeadie: there's something about having it at the top level that might be eventually easier for users (i.e. get all `scheduler::worker` and then look at their output over time), but is not worth the duplication for now.

Jeadie (dismissed):

phillipleblanc (approved):

PR #6102, opened 2025-06-03
test: Add multi-dataset cron test

peasee:
## 📝 Summary

* Adds an integration test for a multi-dataset cron test

## 🔗 Related

* Part of #5683


crates/runtime/tests/acceleration/cron.rs
```diff
@@ -132,7 +133,9 @@ async fn test_cron_schedule_creates() -> Result<(), anyhow::Error> {
                 .expect("append to file");
 
             // wait 30 seconds for at least one run of the cron job
-            tokio::time::sleep(std::time::Duration::from_secs(30)).await;
+            let second_now = chrono::Utc::now().second();
+            let wait = (second_now % 30) + 10; // wait for the next 30th second, and wait 10 seconds for the job to succeed
+            tokio::time::sleep(std::time::Duration::from_secs(u64::from(wait))).await;
```
Sevenannn: Is there any other utils that can be used to determine cron job start / succeed here? Wait for certain time is what causes some flaky integration tests in the past.

Jeadie (dismissed):

crates/runtime/tests/acceleration/cron.rs
peasee: The nature of this test is linked to time (e.g. every 30th second), so it must wait.

phillipleblanc (approved):

PR #6104, opened 2025-06-03
Avoid partial writes by FlightTableWriter

sgrebnov:
## 📝 Summary

Existing implementation processes each batch as a separate Flight DoPut operation which finalize/commit write even if some later write operation fails. This PR improves this to handle write via single/atomic operation and added streaming do_put implementation


kczimm (dismissed):

Sevenannn (dismissed):

crates/data_components/src/flight/write.rs
```diff
@@ -135,14 +135,21 @@ impl DataSink for FlightDataSink {
...
-        let mut flight_client = self.flight_client.clone();
+        let mut batches = Vec::new();
 
+        // We collect all batches to publish them in one go, as each publishing operation is a separate
+        // Flight DoPut invocation, including finalizing the write and rate limiting.
+        // This makes the write operation atomic, meaning that if it fails, no data is published.
         while let Some(batch) = data.next().await {
             let batch = batch.map_err(check_and_mark_retriable_error)?;
             num_rows += batch.num_rows() as u64;
+            batches.push(batch);
```
Sevenannn: Just to confirm - Is the batches size not expected to be very huge and wouldn't consume much memory?

crates/data_components/src/flight/write.rs
```diff
@@ -135,14 +135,21 @@ impl DataSink for FlightDataSink {
...
+        }
 
+        if !batches.is_empty() {
+            let mut flight_client = self.flight_client.clone();
             flight_client
-                .publish(&format!("{}", self.table_reference), vec![batch])
+                .publish(&format!("{}", self.table_reference), batches)
```
phillipleblanc: `publish` is a method we created, the underlying DoPut supports streaming itself. We can pass a stream instead of Vec<RecordBatch> and create a new `publish_streaming` that handles the stream writing.

sgrebnov: Will re-work to have proper streaming support. Closing this PR

crates/data_components/src/flight/write.rs
sgrebnov: This is correct, we aim to small data that we can handle in-memory

phillipleblanc (approved):

PR #6110, opened 2025-06-04
Improve the ListingTableURL logging

phillipleblanc:
## 📝 Summary

Improves the logging of the ListingTableURL

### DIFF
--- a/crates/runtime/src/dataconnector/listing/connector.rs
+++ b/crates/runtime/src/dataconnector/listing/connector.rs
@@ -409,7 +409,7 @@ pub trait ListingTableConnector: DataConnector {
         let (schema_infer_url, schema_infer_meta) =
             if let Some(url) = dataset.params.get("schema_source_path") {
                 let url = self.get_object_store_url(dataset, Some(url))?;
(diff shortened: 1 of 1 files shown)


crates/runtime/src/dataconnector/listing/connector.rs
```diff
@@ -765,16 +769,47 @@ fn to_listing_table_url(
...
             connector_component: ConnectorComponent::from(dataset),
         },
-    )
+    )?;
+
+    Ok(SensitiveListingTableUrl::new(sensitive_url, new_url))
+}
+
+fn sanitize_url(mut url: Url) -> Url {
+    url.set_fragment(None);
```
Jeadie, line 787: Fragment is the only place we'd expect secrets?

Jeadie (approved):

crates/runtime/src/dataconnector/listing/connector.rs
phillipleblanc: Yes, we add them to the fragment for the object store registry to extract

sgrebnov (approved):

PR #6123, opened 2025-06-04
Emit retry error on flight transient connection failure

### LINKED ISSUE #6120: task: Emit connection reset error on transient failure for spice cloud connector

Sevenannn:
## 📝 Summary

Define a new error variant `ConnectionReset` in spice_client error indicating retryable connection reset. Emit this  retry error on all transient connection failure for flight requests.

## 🔗 Related

- Close: https://github.com/spiceai/spiceai/issues/6120
- Part of: https://github.com/spiceai/spiceai/issues/4035

Client retry will be implemented in incremental change as part of https://github.com/spiceai/spiceai/issues/6121 and https://github.com/spiceai/spiceai/issues/6122

### DIFF
--- a/crates/flight_client/src/lib.rs
+++ b/crates/flight_client/src/lib.rs
@@ -172,6 +172,9 @@ pub enum Error {
         "No endpoints found. Ensure the endpoint is configured and the server is running."
     ))]
     NoEndpointsFound,
+
+    #[snafu(display("Connection is reset by the server. Please retry the request.\n{source}"))]
+    ConnectionReset { source: TonicStatusError },
 }
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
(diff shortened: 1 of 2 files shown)


crates/flight_client/src/lib.rs
```diff
@@ -629,7 +641,25 @@ impl FlightClient {
...
+    }
     Error::UnableToQuery {
         source: e.message().into(),
     }
 }
+
+pub fn is_connection_reset_error(error: &tonic::Status) -> bool {
+    if error.code() == tonic::Code::Internal {
+        let error_message = error.message().to_lowercase();
+        if error_message.contains("operation was canceled")
```
kczimm, line 657: is there a typo in the `tonic::Status` message or your `contains` invocation. "operation was cancelled"

kczimm: Nevermind, I guess one way is British english. I did not know that.

kczimm (approved):

crates/flight_client/src/lib.rs
```diff
@@ -629,7 +641,25 @@ impl FlightClient {
...
+        return Error::ConnectionReset {
+            source: TonicStatusError::from(e),
+        };
+    }
     Error::UnableToQuery {
         source: e.message().into(),
     }
 }
+
+pub fn is_connection_reset_error(error: &tonic::Status) -> bool {
```
peasee, line 654: Can we add an integration test for this, to test the behavior by specifying an invalid endpoint (e.g. to simulate if the endpoint were down) and confirming that the connection is not retried as a connection reset error?

Sevenannn: The invalid endpoint won't be able to reach the rpc calls, it would already fail at https://github.com/spiceai/spiceai/blob/afc3927bf140333afff8bc9ad4cb9960bc0ce62b/crates/runtime/src/dataconnector/spiceai.rs#L188, I don't think integration test is necessary for this specific case

PR #6136, opened 2025-06-05
Use spice-rs in test operator and retry on connection reset error

### LINKED ISSUE #6121: task: use spice-rs in test operator

Sevenannn:
## 📝 Summary

This PR includes changes to use spice-rs sdk inside the test framework, and changes to retry on connection reset error emitted by spice cloud connector
- Replace flight client with spice-rs client for query in test framework & test operator. Each `SpiceTestQueryWorker` contains a `spice_client: Arc<SpiceClient>`.
- In test-framework / test operator, retry the query stream consumption in `execute_query` and `query_to_batches` when the connection reset error is encountered. `is_transient_error` function is used to determine the connection reset error
- For retryable spice error, add `spiceai-retryable` metadata in tonic error status for easy matching & handling of retryable error from client side.

TPCDS Load test run at designated concurrency without issue after this change: https://github.com/spiceai/spiceai/actions/runs/15497392622/job/43637240095

Changes have been validated on a full benchmark dispatch:
https://github.com/spiceai/spiceai/actions/runs/15501496471 and a throughput run: https://github.com/spiceai/spiceai/actions/runs/15570591024

## 🔗 Related

- spice-rs change: https://github.com/spiceai/spice-rs/pull/51
- Close: https://github.com/spiceai/spiceai/issues/6121

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,7 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
-spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "1ab50ce213e090b9dff13f50499294c371618d61" }
+spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "5272a8f43002993a30144cac7175dfc95400d287" }
 
 [profile.release-lto]
 inherits = "release"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,7 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
-spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "1579a895fa467a8d9f8eafd48219f9768e48524d" }
+spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "4e1b0394aeba7b03d61edfcad27562045c450b02" }
 
 [profile.release-lto]
 inherits = "release"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,6 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
+spiceai = { path = "../spice-rs" }
 
 [profile.release-lto]
 inherits = "release"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,6 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
+# TODO: Switch to official crate once spice-rs is released
 spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "4e1b0394aeba7b03d61edfcad27562045c450b02" }
 
 [profile.release-lto]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,7 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
-spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "3d0e679a98ca51d8df3fa593f3db8f3f7f2b2eae" }
+spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "df953882dbb54e4c0d713fd87ed6ddf077406250" }
 
 [profile.release-lto]
 inherits = "release"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,7 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
-spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "5272a8f43002993a30144cac7175dfc95400d287" }
+spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "3d0e679a98ca51d8df3fa593f3db8f3f7f2b2eae" }
 
 [profile.release-lto]
 inherits = "release"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,7 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
-spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "df953882dbb54e4c0d713fd87ed6ddf077406250" }
+spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "1579a895fa467a8d9f8eafd48219f9768e48524d" }
 
 [profile.release-lto]
 inherits = "release"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,7 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
-spiceai = { path = "../spice-rs" }
+spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "6fd5d663151949836eb510326854bbf7c67efbf1" }
 
 [profile.release-lto]
 inherits = "release"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -171,7 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
-spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "6fd5d663151949836eb510326854bbf7c67efbf1" }
+spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "1ab50ce213e090b9dff13f50499294c371618d61" }
 
 [profile.release-lto]
 inherits = "release"
--- a/crates/data_components/src/flight.rs
+++ b/crates/data_components/src/flight.rs
@@ -524,7 +524,7 @@ fn query_to_stream(
                     match batch {
                         Ok(batch) => yield Ok(batch),
                         Err(error) => {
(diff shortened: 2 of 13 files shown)


crates/test-framework/src/spicetest/datasets/mod.rs
```diff
@@ -175,32 +177,33 @@ impl SpiceTest<NotStarted> {
...
-                if let Some(multi) = &multi {
-                    worker.with_progress_bar(multi.add(self.get_new_progress_bar()))
-                } else {
-                    worker
-                }
-            })
-            .map(SpiceTestQueryWorker::start)
-            .collect();
+        let mut query_workers = Vec::with_capacity(self.state.parallel_count);
+        for id in 0..self.state.parallel_count {
```
peasee: why can't we continue to use the `(0..self.state.parallel_count)` pattern?

crates/test-framework/src/spicetest/datasets/worker.rs
```diff
@@ -415,11 +475,6 @@ impl SpiceTestQueryWorker {
             match batch {
                 Ok(None) => break,
                 Err(e) => {
-                    eprintln!(
```
phillipleblanc, line 418: why is this removed?

crates/test-framework/src/spicetest/datasets/worker.rs
```diff
@@ -495,20 +550,12 @@ impl SpiceTestQueryWorker {
         }
 
         let duration = query_start.elapsed();
-        query_durations
```
phillipleblanc, line 498: why is this removed?

crates/test-framework/src/spicetest/datasets/worker.rs
```diff
@@ -401,11 +405,67 @@ impl SpiceTestQueryWorker {
...
+    async fn execute_query_internal(
+        &self,
+        query: &Query,
+        results_snapshot: bool,
+        validate: bool,
+    ) -> Result<(Duration, usize)> {
         let query_start = Instant::now();
-        let mut result_stream = self
-            .flight_client
+        let mut spice_client = self.spice_client.lock().await;
```
phillipleblanc: I don't think it makes sense to take a lock to make a query?

crates/test-framework/src/spicetest/datasets/worker.rs
Sevenannn: The error println is moved to the `execute_query`, it prints out the failure when the retry fail / the error is not retryable

crates/test-framework/src/spicetest/datasets/worker.rs
Sevenannn: Similar to above, the write to `query_durations` is moved to `execute_query` to write only when query succeed / query retry succeed.

crates/test-framework/src/spicetest/datasets/worker.rs
Sevenannn: The `query` method of spice-rs client takes a mutable reference of the `Client`, and the `&mut Client` won't be properly captured inside the retry closure. I think it make sense to do some refactor inside the spice-rs SDK so that a `&mut Client` is not required for query. I'll make a change in the SDK and remove the locks inside the test operator.

Jeadie (dismissed):

peasee (changes requested):
Has there been a successful testoperator run with this SDK? This should not merge until a full dispatch run has been completed for all of bench, throughput and load tests.

Sevenannn:
@peasee thank you for calling this out, I still have some changes in progress for this pr and will make sure to run all the tests before it's merged

crates/test-framework/src/spicetest/datasets/mod.rs
Sevenannn: The spice client doesn't implement a clone and to get_spiced is async and requires error handling. Is there any specific benefit of using (0..self.state.parallel_count) instead of loop?

peasee: Why doesn't the spice-rs client implement clone? Our `FlightClient` is `Clone`, and does not require `&mut self`. It sounds like the spice-rs client might need some improvements before we use it here.

phillipleblanc: Yeah, we shouldn't require taking a lock to make a query. The SDK should work similarly to our FlightClient

Sevenannn: Thanks for the feedback, I've already incorporated changes to not taking a mut self for query in this commit: https://github.com/spiceai/spiceai/pull/6136/commits/2e78f9020471c6a85b475ea98d6f8f1d84a05d00. Sorry for not updating the summary promptly. I can make the change to make it a clone as well, but I'd like to understand the specific benefit of making it clone and using a `(0..self.state.parallel_count)`, would you help clarify?

sgrebnov (dismissed):

Cargo.toml
```diff
@@ -171,6 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
+spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "df953882dbb54e4c0d713fd87ed6ddf077406250" }
```
sgrebnov: TODO: this needs to be updated before merging

crates/test-framework/src/spicetest/datasets/mod.rs
peasee: It avoids needing a mutable vec, and is generally more concise/declarative. In this case, there are no other trade-offs for either (in my opinion) - it is a styling recommendation.

Cargo.toml
```diff
@@ -171,6 +171,7 @@ tracing-opentelemetry = "0.28"
 tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 uuid = "1"
 x509-certificate = "0.23.1"
+spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "4e1b0394aeba7b03d61edfcad27562045c450b02" }
```
sgrebnov, line 175: Please add TODO to replace with official crates version after release

crates/data_components/src/flight.rs
```diff
@@ -538,32 +538,39 @@ fn query_to_stream(
 fn to_execution_error(e: Error) -> DataFusionError {
     match e {
         Error::Flight { source } => match source {
-            flight_client::Error::UnableToQuery { source } => {
-                DataFusionError::Execution(format!("{source}"))
+            flight_client::Error::UnableToQuery {
+                source: source_error,
+            } => DataFusionError::Execution(format!("{source_error}")),
+            flight_client::Error::ConnectionReset { source: _ } => {
```
peasee:
```suggestion
            flight_client::Error::UnableToQuery {
                source,
            } => DataFusionError::Execution(format!("{source}")),
            flight_client::Error::ConnectionReset { .. } => {
```

nit: This can be more concise - there is no reason to specify `source: _` when we can just drop the struct. There is no reason to override `source: source_error` when we can overload `source` for the scope of that match.

crates/data_components/src/flight.rs
```diff
@@ -538,32 +538,39 @@ fn query_to_stream(
...
+                return DataFusionError::External(Box::new(FlightClientError::ConnectionReset {
+                    source: TonicStatusError::from(source_owned),
+                }));
             }
+            DataFusionError::Execution(format!(
+                "{}",
+                Error::ArrowFlight {
+                    source: Box::new(error),
+                }
+            ))
```
peasee:
```suggestion
            DataFusionError::Execution(
                Error::ArrowFlight {
                    source: Box::new(error),
                }
                .to_string(),
            )
```

nit: There is no need to use `format!()` when the error already implements `Display`.

crates/data_components/src/flight.rs
```diff
@@ -538,32 +538,39 @@ fn query_to_stream(
...
+        _ => DataFusionError::Execution(format!(
+            "{}",
             Error::ArrowFlight {
                 source: Box::new(error),
             }
-        }
-        _ => Error::ArrowFlight {
-            source: Box::new(error),
-        },
+        )),
```
peasee: as above

crates/flight_client/src/lib.rs
```diff
@@ -652,14 +652,19 @@ fn map_tonic_error_to_message(e: tonic::Status) -> Error {
...
-            || error_message.contains("grpc-status header missing")
-        {
-            return true;
+    match error.code() {
+        tonic::Code::Internal | tonic::Code::Cancelled => {
+            let error_message = error.message().to_lowercase();
+            if error_message.contains("operation was canceled")
+                || error_message.contains("http2 error")
+                || error_message.contains("grpc-status header missing")
+                || error_message.contains("received message with invalid compression flag")
```
peasee, line 661: this compression flag error seems like it could be more than a connection reset?

crates/test-framework/src/spicetest/datasets/worker.rs
```diff
@@ -401,11 +431,70 @@ impl SpiceTestQueryWorker {
...
+            .spice_client
+            .query_with_params(
+                &query.sql,
+                query
+                    .get_parameters_batch()
+                    .transpose()
+                    .map_err(QueryError::nonretryable)?,
+            )
+            .await
+            .map_err(|e| QueryError::nonretryable(anyhow!(e)))?;
```
peasee, line 497: if a connection reset error occurs here, wouldn't it become nonretryable?

crates/test-framework/src/spicetest/datasets/worker.rs
```diff
@@ -476,7 +568,9 @@ impl SpiceTestQueryWorker {
                 format!("{name}_{query_name}_sf{}", self.scale_factor)
             };
 
-            let records_pretty = arrow::util::pretty::pretty_format_batches(&limited_records)?;
+            let records_pretty = arrow::util::pretty::pretty_format_batches(&limited_records)
+                .map_err(QueryError::nonretryable)?
+                .to_string();
```
peasee: why is this getting `.to_string()` now?

crates/test-framework/src/spicetest/datasets/worker.rs
Sevenannn: This part of retry is currently handled by spice-rs SDK currently so it's not retried in test operator.

crates/flight_client/src/lib.rs
Sevenannn: The invalid compression flag indicates a invalid grpc payload. In cases where the connection is not suddenly cut there won't be a won't be a corrupted grpc payload, assuming that all of the intermediaries passes valid grpc response, see original error here: `FAIL - Worker 2 - Query 'tpcds_q52' failed: Tonic error: status: InvalidArgument, message: "Query execution failed.\nTonic error: status: Internal, message: \"protocol error: received message with invalid compression flag: 101 (valid flags are 0 and 1) while receiving response with status: 502 Bad Gateway\", details: [], metadata: MetadataMap { headers: {} }\nVerify the configuration and try again.", details: [], metadata: MetadataMap { headers: {} } `

crates/test-framework/src/spicetest/datasets/worker.rs
Sevenannn: nice catch, updated

peasee (approved):

PR #6137, opened 2025-06-05
fix: Invalidate search cache on refresh

peasee:
## 📝 Summary

* Ensures the search cache invalidates on a dataset refresh
* Updates the way invalidation is handled through a `CacheProvider`:
  * `TableInvalidator` has been removed. This required a reference to the underlying `LruCache`, which `CacheProvider` could not provide.
  * Instead, `invalidate_for_table` is now a member of `CacheProvider` and has a generic implementation on `V: AsTableRefs`
  * Introduces `AsTableRefs` for cache values, which returns the list of table references involved in the value. It is the responsibility of the value to ensure it reports the correct tables for invalidation.
* Updates acceleration tasks to take a reference of `Arc<Caching>` instead of `Arc<QueryResultsCacheProvider>` so that it can invalidate more than just the one cache.

## 🔗 Related

* Part of #3018


crates/cache/src/result/search.rs
```diff
@@ -56,6 +56,12 @@ pub struct CachedSearchResult {
     pub results: Arc<HashMap<TableReference, CachedAggregationResult>>,
 }
 
+impl AsTableRefs for CachedSearchResult {
+    fn as_table_refs(&self) -> Arc<HashSet<TableReference>> {
```
sgrebnov, line 61: Can we return an iterator instead to avoid unnecessary new data allocation?

```
trait AsTableRefs<'a> {
    type Iter: Iterator<Item = &'a TableReference>;
    fn table_refs(&'a self) -> Self::Iter;
}
``

peasee: An iterator would probably result in more allocations as it would clone each inner element. I've updated instead to also store `input_tables` for search cache results, similar to query cache results, so the output is just a single `Arc::clone`.

sgrebnov (approved):

PR #6140, opened 2025-06-05
Update Spice Cloud Platform management UX

sgrebnov:
## 📝 Summary

Update Spice Cloud Platform management UX:

1. Internal sink dataset is now under dedicated `scp` schema: `runtime.task_history_cloud` -> `scp.task_history`
2. `api_key` has been moved from `params` directly to `management`
3. Target flight endpoint override param renamed to `data_endpoint`
```yaml
management:
  enabled: true
  api_key: ${ secrets:SPICEAI_API_KEY}
  params:
    data_endpoint: https://dev-flight.spiceai.io
```

### DIFF
--- a/crates/runtime/src/datafusion/builder.rs
+++ b/crates/runtime/src/datafusion/builder.rs
@@ -21,7 +21,8 @@ use std::{
 };
 
 use crate::{
-    dataaccelerator::AcceleratorEngineRegistry, object_store_registry::SpiceObjectStoreRegistry,
+    dataaccelerator::AcceleratorEngineRegistry, datafusion::SPICE_SCP_SCHEMA,
+    object_store_registry::SpiceObjectStoreRegistry,
 };
 use cache::Caching;
 use datafusion::{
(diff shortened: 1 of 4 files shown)


crates/runtime/src/datafusion/builder.rs
```diff
@@ -213,7 +214,14 @@ impl DataFusionBuilder {
...
             Err(e) => {
-                panic!("Unable to register spice runtime schema: {e}");
+                panic!("Unable to register spice metadata schema: {e}");
+            }
+        }
+
+        match catalog.register_schema(SPICE_SCP_SCHEMA, Arc::new(SpiceSchemaProvider::new())) {
+            Ok(_) => {}
+            Err(e) => {
+                panic!("Unable to register spice cloud platform schema: {e}");
```
peasee, line 224: Is this `unreachable!()`, or an actual panic?

peasee (approved):

crates/runtime/src/datafusion/builder.rs
sgrebnov: It is `panic!` for consistency with other schema registration

PR #6149, opened 2025-06-08
Improve parsing of full text search query 

Jeadie:
## 📝 Summary
 - Currently, full text search will match against documents the contain the full query string. 
 - This PR updates it so that it score against documents that contain any of the terms in the query string. Scoring will still be relative to term frequency etc. 
- Example
  - Before: "foo and bar" -> Match documents that contain the string "foo and bar".
  - After: "foo and bar" -> Match documents that contain at least one of "foo", "and", "bar".


peasee (commented):
Should this be a configuration parameter? `text_search_mode: full_match | any_match`

peasee (approved):

PR #6156, opened 2025-06-09
Support multiple columns in FTS

Jeadie:
## 📝 Summary
 - Support > 1 full text search column
   ```yaml
     datasets:
      - name: d1
        from: file:cleaned_sales_data.parquet
        acceleration:
          enabled: true
        columns:
          - name: product_line
            full_text_search:
              enabled: true
              row_id:
                - order_line_number
                - order_number
          - name: address_line1
            full_text_search:
              enabled: true
          ```

- Fixed some bugs,
  - For Reciprocal rank fusion, if there are >1 input set of candidates, but only <=1 are non-empty, we must return early. 
  - Handle `limit` in `PostApplyCandidateGeneration`.


crates/runtime/tests/models/search.rs
```diff
@@ -554,15 +554,24 @@ async fn test_text_search_multiple_columns() -> Result<(), anyhow::Error> {
             SearchTestCase {
                 name: "multi_text_column_basic",
                 body: json!({
-                    "text": "general",
+                    "text": "In general basic",
```
Jeadie, line 557: More specific query so that identical scored results do not break consistency of testing.

kczimm (approved):

PR #6182, opened 2025-06-10
Jeadie/25 06 10/finance

Jeadie:
## 📝 Summary
 - Add full text search and hybrid search for financebench

## 🔗 Related
 - #3015 

## 👀 Notes for Reviewers
 - Once we add spicepod dependencies for `test-operator`, we shall not need three duplicate spicepod.yaml files.


test/financebench/spicepod_fts_openai.yaml
peasee, line 1: These should go into `test/spicepods`? There is already `test/spicepods/models`. Maybe there needs to be sub-directories under `test/spicepods/models` for the various model-based test types, like `test/spicepods/models/fts`, etc.

We should also update the [naming convention](https://github.com/spiceai/spiceai/blob/trunk/test/spicepods/README.md) to account for model-based spicepods.

peasee (approved):

PR #6191, opened 2025-06-10
Use arrow pretty print util instead of init dataframe / logical plan in `display_records`

### LINKED ISSUE #4219: Dremio TPCDS SF5 Query 64 failed with error: SQL REPL Error: Schema error: Schema contains duplicate qualified field name "?table?".syear
**Note**: This is a SQL REPL error, scoped out of 1.0 Stable milestone

Sevenannn:
## 📝 Summary

The `display_records` used to create a logical plan when displaying results in the REPL, which was unecessary and causing results with duplicated name failed to be displayed. This PR simplifies the result display by directly limit the number of rows to be displayed for given result batches.

The new approach iterates through the full results and collect the first 500 rows to be displayed. The result is then displayed through `pretty_format_batches `. This keeps parity with the `df.display()` approach -  The `df.display()` uses `pretty::print_batches(&results)` for printing results `Vec<RecordBatch>`, which is using the `println!("{}", pretty_format_batches(results)?);`  underlying.

Unit tests with snapshot assertion are added to ensure the correct record batch operation and display for the updated `display_records` function. The `pretty_format_batches` result is returned in `display_records` to facilitate the test snapshotting.

The UX of SQL REPL doesn't change, it displays the truncated / non-truncated results as before:
* truncated:
```SQL
sql> select * from catalog_page;
```
image

* non-truncated
```SQL
select * from catalog_page limit 10;
```
image

## 🔗 Related

- Close: https://github.com/spiceai/spiceai/issues/4219


peasee (commented):
The CLI used to display if it truncated results, and the full count. Is this still displayed?

You should include a before and after example of the REPL in use

Sevenannn:
@peasee There isn't any change from the UX perspective, user see exactly the same results from the sql repl. I added notes and screenshots in the pr description so that it's easier to understand by reviewers

peasee (approved):

PR #6192, opened 2025-06-10
fix: return empty stream when no results for Databricks SQL Warehouse

kczimm:
## 📝 Summary

Fixes a bug that causes an error when no results are returned.

### DIFF
--- a/crates/data_components/src/databricks/sql_warehouse.rs
+++ b/crates/data_components/src/databricks/sql_warehouse.rs
@@ -203,8 +203,16 @@ impl SqlWarehouseApi {
         let token = self.token_provider.get_token();
         let initial_external_link = Self::extract_external_links(result_object)?;
 
+        // If no external link, return an empty stream
+        if initial_external_link.is_none() {
+            return Ok(Box::pin(RecordBatchStreamAdapter::new(
+                Arc::new(Schema::empty()),
+                Box::pin(stream::empty()),
+            )));
+        }
+
         let token = token.to_string();
-        let stream = stream::unfold(Some(initial_external_link), move |current_link| {
+        let stream = stream::unfold(initial_external_link, move |current_link| {
             let api = Arc::clone(&self);
             let token = token.clone();
             async move {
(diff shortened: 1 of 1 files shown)


sgrebnov (approved):

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -203,8 +203,16 @@ impl SqlWarehouseApi {
         let token = self.token_provider.get_token();
         let initial_external_link = Self::extract_external_links(result_object)?;
 
+        // If no external link, return an empty stream
+        if initial_external_link.is_none() {
+            return Ok(Box::pin(RecordBatchStreamAdapter::new(
+                Arc::new(Schema::empty()),
+                Box::pin(stream::empty()),
```
Sevenannn, line 210: When no results is returned for a query, is it expected to have a schema (e.g. some queries lead to 0 rows, but still have a schema)?

kczimm: I don't believe it matters. I've tested with the following queries:

Join with empty subquery:
```sql
SELECT r.r_regionkey, r.r_name
FROM region r
LEFT JOIN (
    SELECT r_regionkey
    FROM region
    WHERE r_name = 'NO_SUCH_REGION'
) sub ON r.r_regionkey = sub.r_regionkey
WHERE sub.r_regionkey IS NULL;
```

aggregation with empty subquery:
```sql
SELECT COUNT(*)
FROM region
WHERE r_regionkey IN (
    SELECT r_regionkey
    FROM region
    WHERE r_name = 'NO_SUCH_REGION'
);
```

nested subquery:
```sql
SELECT r_regionkey
FROM region
WHERE r_name IN (
    SELECT r_name
    FROM region
    WHERE r_regionkey IN (
        SELECT r_regionkey
        FROM region
        WHERE r_name = 'NO_SUCH_REGION'
    )
);
```

I was concerned subqueries wouldn't work but it appears that Datafusion has no problem with empty results having no schema.

PR #6200, opened 2025-06-11
Add bucket UDF for partitioning

### LINKED ISSUE #6183: Implement Scalar UDFs for Partitioning

kczimm:
## 📝 Summary

Adds the `bucket` UDF which partitions a column into a fixed number of buckets based on a hash of the column value.

## 🔗 Related

Part of: #6183


crates/runtime/src/datafusion/udf/bucket.rs
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, arg_types: &[DataType]) -> Result<DataType, DataFusionError> {
+        if arg_types.len() != 2 {
+            return Err(DataFusionError::Plan(format!(
+                "Expected exactly two arguments, got {}",
```
Sevenannn: `InvalidArgumentCount` can be reused here

crates/runtime/src/datafusion/udf/bucket.rs
```diff
@@ -0,0 +1,321 @@
...
+fn compute_bucket_array(array: &ArrayRef, num_buckets: i64) -> Result<Int32Array, DataFusionError> {
+    let mut hashes = vec![0; array.len()];
+    let random_state = RandomState::new();
+    let capacity = array.len();
+    create_hashes(&[Arc::clone(array)], &random_state, &mut hashes)?;
+
+    let mut builder = Int32Array::builder(capacity);
+    for (i, hash) in hashes.iter().enumerate().take(capacity) {
+        let bucket = if array.is_null(i) {
+            NULL_BUCKET
```
Sevenannn: Is the actual number of bucket `num_buckets + 1` since there's a extra `NULL_BUCKET`

kczimm: Good observation. I changed the behavior of NULLs to conform to how the bucket transform in Iceberg works. NULL values result in NULL.

crates/runtime/src/datafusion/udf/bucket.rs
```diff
@@ -0,0 +1,357 @@
...
+    let random_state = RandomState::new();
+    create_hashes(&[array], &random_state, &mut hashes)?;
+    Ok(ScalarValue::Int32(Some(
+        u64::try_from(num_buckets)
+            .and_then(|n| i32::try_from(hashes[0] % n))
+            .context(BucketLargerThanTypeSnafu)?,
+    )))
+}
+
+fn compute_bucket_array(array: &ArrayRef, num_buckets: i64) -> Result<Int32Array, DataFusionError> {
```
phillipleblanc: This is going to be slow because it processes each array item sequentially. It would be better to compute the hashes into an Arrow Array and then use one of the vectorized arrow compute kernels to compute the bucket positioning.

https://docs.rs/arrow/latest/arrow/compute/index.html

If there isn't a ready-made compute kernel we can use, then we should mimic the vectorized computation those kernels internally provide.

phillipleblanc (dismissed):

crates/runtime/src/datafusion/udf/bucket.rs
```diff
@@ -0,0 +1,464 @@
...
+    ColumnarValue, ScalarFunctionArgs, ScalarUDFImpl, Signature, Volatility,
+};
+use datafusion::scalar::ScalarValue;
+use snafu::{ResultExt as _, Snafu};
+
+/// Maximum number of buckets, chosen to support large-scale partitioning while preventing excessive memory usage.
+const MAX_NUM_BUCKETS: i64 = 1_000_000;
+
+/// Fixed seed for deterministic hashing.
+const HASH_SEED: usize = 0x53_50_49_43_45;
```
phillipleblanc: Can we also snapshot the value we get in the test? It wouldn't be good for this value to change across versions.

kczimm: excellent idea!

phillipleblanc (approved):

Issue #6214, opened 2025-06-12
Bug: Full text search index does not use accelerated table for filter conditions


Jeadie:
## Describe the bug
Given this spicepod

```yaml
datasets:
  - from: github:github.com/spiceai/docs/pulls
    name: doc.pulls
    params:
      github_token: ${secrets:GITHUB_TOKEN}
    acceleration:
      enabled: true
    columns:
      - name: title
        full_text_search:
          enabled: true
          row_id:
            - id
      - name: body
        full_text_search:
          enabled: true

embeddings:
  - name: hf_minilm
    from: huggingface:huggingface.co/sentence-transformers/all-MiniLM-L6-v2
```

Running the following is slow because the full text search CandidateGeneration is not using the accelerated table, and has to fall back to Github API.
```shell
>>> curl -XPOST http://localhost:8090/v1/search \
    -H "Content-Type: application/json" \
    -d '{
        "datasets": ["doc.pulls"],
        "text": "Glue data",
        "limit": 3,
        "where": "comments_count>1"
```

PR #6215, opened 2025-06-12
Don't use base table for full text search post apply vector search

Jeadie:
## 🔗 Related
 Solves #6214

### DIFF
--- a/crates/runtime/src/search/full_text/table.rs
+++ b/crates/runtime/src/search/full_text/table.rs
@@ -114,7 +114,7 @@ impl TableWithFullText {
 
     /// Construct a new [`TableWithFullText`] with an updated [`TableProvider`].
     ///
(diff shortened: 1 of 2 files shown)


crates/runtime/src/search/full_text/table.rs
```diff
@@ -112,6 +112,19 @@ impl TableWithFullText {
         Arc::clone(&self.base_table)
     }
 
+    /// Construct a new [`TableWithFullText`] with an updated [`TableProvider`].
+    ///
+    /// No Checks are done to confirm compatibility between the current index and the provider [`TableProvider`].
```
phillipleblanc:
```suggestion
    /// No Checks are done to confirm compatibility between the current index and the provided [`TableProvider`].
```

phillipleblanc (dismissed):

peasee (approved):

PR #6222, opened 2025-06-12
Paginate results from FullTextSearch

Jeadie:
## 📝 Summary
Currently `FullTextSearch` searches for its entire `limit`. Since `FullTextSearch` is often augmented with `PostApplyCandidateGeneration`, it's limit is often artificially high (to accomodate its filter pushdown limitations). 

This PR paginates results back in its  `SendableRecordBatchStream` to reduce how many search results `FullTextSearch` has to traverse, and search documents it has to retrieve.

### DIFF
--- a/crates/search/src/generation/post_apply.rs
+++ b/crates/search/src/generation/post_apply.rs
@@ -261,11 +261,11 @@ impl CandidateGeneration for PostApplyCandidateGeneration {
             .collect::<Vec<_>>();
 
         let need_post_apply = !unapplied_filters.is_empty() || !unapplied_projection.is_empty();
-        let underlying_limit = if need_post_apply {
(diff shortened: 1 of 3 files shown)


crates/search/src/generation/post_apply.rs
```diff
@@ -261,11 +261,11 @@ impl CandidateGeneration for PostApplyCandidateGeneration {
             .collect::<Vec<_>>();
 
         let need_post_apply = !unapplied_filters.is_empty() || !unapplied_projection.is_empty();
-        let underlying_limit = if need_post_apply {
+        let underlying_limit = if unapplied_filters.is_empty() {
```
Jeadie, line 264: This is a separate improvement I found. We only need a higher limit iff there are unapplied filters.

peasee (approved):

PR #6229, opened 2025-06-13
fix: query schema for is_nullable

kczimm:
## 📝 Summary

Query the schema for the `is_nullable` column to determine if a field is nullable in SQL Warehouse.


crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -163,14 +163,16 @@ impl SqlWarehouseApi {
 
     async fn get_schema(&self, table: &TableReference) -> Result<SchemaRef, Error> {
```
sgrebnov, line 164: Do we need to update some of the integration tests to include `describe table` snapshot based verification  for nullability?

sgrebnov (approved):

Sevenannn (approved):

PR #6233, opened 2025-06-14
Add Remote Spicepod support

phillipleblanc:
# Add Remote Spicepod Support (Object Store Integration)

## Summary

This PR adds support for loading Spicepods from remote locations using object stores, enabling users to reference Spicepods from cloud storage services like S3, Google Cloud Storage, and Azure Blob Storage.

It also adds a positional parameter on `spiced` to allow specifying a Spicepod on the filesystem that isn't in the current directory.

`spiced s3://my_bucket/spicepod.yaml`
`spiced /path/to/spicepod.yaml`
`spice run s3://my_bucket/spicepod.yaml`

(see more usage examples below)

### 🎯 Core Features

**Command Line Interface**
- Adds a positional argument to `spiced` binary to specify custom Spicepod path/URL
- Supports both local filesystem paths and remote URLs (s3://, gs://, azure://, etc.)
- Defaults to current directory if not specified

**Object Store Integration**
- Added `ObjectStoreFilesystem` implementation for reading from cloud storage
- Integrated `object_store` crate with AWS, GCP, and Azure features enabled
- Added URL parsing and validation for remote Spicepod locations

**Support interactive auth options for S3**
- Adds a new crate `object_store_aws_sdk` which implements the `CredentialProvider` trait in `object_store` and leverages the Rust AWS SDK's credential provider that follows the full credential provider chain as outlined in: https://docs.aws.amazon.com/sdkref/latest/guide/standardized-credentials.html#credentialProviderChain

### 🔄 API Changes

**Async Transformation**
- Converted `Spicepod::load()` and related methods to async functions
- Updated `AppBuilder::build_from_filesystem_path()` to be async and renamed to `AppBuilder::build_from_path()`
- Made `resolve_component_references()` async with concurrent processing
- Added `ReadablePath` and `ReadableYaml` traits with async methods

## Usage Examples

```bash
# Local Spicepod
spiced ./my-spicepod-dir
spiced ./my-spicepod-dir/spicepod.yaml

# S3 Spicepod
spiced s3://my-bucket/spicepods/production/
spiced s3://my-bucket/spicepods/production/spicepod.yaml
spice run s3://my-bucket/spicepods/production/
spice run s3://my-bucket/spicepods/production/spicepod.yaml

# Google Cloud Storage
spiced gs://my-bucket/spicepods/staging/

# Azure Blob Storage
spiced abfs://my-account/container/spicepods/

# HTTPS
spiced https://host.com/spicepod.yaml
```

## Security

When a remote spicepod is specified, the Spice instance loads the spicepod at startup and does not allow changes to the spicepod while the instance is running (i.e. pods watcher is disabled). Only HTTPS services are allowed to ensure tampering is not possible.

If an attacker is able to maliciously change the contents of the Spicepod at the remote location and force the Spice instance to restart, then the attacker could load any arbitrary Spicepod.


bin/spiced/src/lib.rs
```diff
@@ -215,7 +223,7 @@ pub async fn run(args: Args) -> Result<()> {
         .with_metrics_server_opt(args.metrics, prometheus_registry.clone());
 
     if args.pods_watcher_enabled {
-        let pods_watcher = PodsWatcher::new(current_dir.clone());
+        let pods_watcher = PodsWatcher::new(spicepod_path.clone());
```
ewgenius, line 227: How podswatcher will work with remote paths?

ewgenius (dismissed):

bin/spiced/src/lib.rs
phillipleblanc: podswatcher is a local only experience

ewgenius: Shouldn't it skip remote paths in that case?

ewgenius (dismissed):

bin/spiced/src/lib.rs
phillipleblanc: Updated to skip

crates/spicepod/src/reader.rs
```diff
@@ -27,49 +29,100 @@ pub enum Error {
...
+
+impl ObjectStoreFilesystem {
+    pub fn new(store: Arc<DynObjectStore>) -> Self {
+        Self { store }
+    }
+}
+
+#[async_trait]
+impl ReadablePath for ObjectStoreFilesystem {
+    async fn open(&self, path: PathBuf) -> Result<Box<dyn io::Read + Send + Sync>> {
```
Sevenannn, line 104: For learning purpose: How is the file path for dependent spicepods handled in the remote spicepod case

sgrebnov (dismissed):

bin/spice/cmd/run.go
```diff
@@ -33,6 +33,7 @@ var runCmd = &cobra.Command{
 	Short: "Run Spice.ai - starts the Spice.ai runtime, installing if necessary",
 	Example: `
 spice run
+spice run /path/to/spicepod.yaml
```
Jeadie, line 39: nit: add a remote path example.

crates/spicepod/src/reader.rs
phillipleblanc: It should work the same way, we use the path to the spicepod as the root and any dependent spicepods are resolved relative to that root and fetched using the object store API.

Jeadie (dismissed):
Currently we're using `object_store_aws_sdk::from_s3_url(&url)` for spicepods in AWS, which is getting its credential from the environment (i.e. `S3CredentialProvider::from_env()`). Does this support all credential chains (e.g. SAML authentication) natively, or will we add support for that?

phillipleblanc:
> Currently we're using `object_store_aws_sdk::from_s3_url(&url)` for spicepods in AWS, which is getting its credential from the environment (i.e. `S3CredentialProvider::from_env()`). Does this support all credential chains (e.g. SAML authentication) natively, or will we add support for that?

Yes, that supports all credential chains supported by the AWS SDK.

Sevenannn (dismissed):

peasee (dismissed):

peasee (approved):

PR #6254, opened 2025-06-17
For Spice.ai connectors, do not default to dev SCP for dev builds

Jeadie:
## 📝 Summary
For dev builds (`cfg!(feature="dev")`), Spice.AI connectors (catalog and dataconnector) currently point to dev by default. We no longer want this.

### DIFF
--- a/crates/runtime/src/catalogconnector/spice_cloud.rs
+++ b/crates/runtime/src/catalogconnector/spice_cloud.rs
@@ -117,7 +117,7 @@ impl SpiceCloudPlatformCatalog {
             .params
             .get("http_endpoint")
             .expose()
-            .unwrap_or_else(|_| "https://spiceai.io".into());
+            .unwrap_or_else(|_| "https://data.spiceai.io");
 
         let mut props = HashMap::new();
         if let ExposedParamLookup::Present(api_key) = self.params.get("api_key").expose() {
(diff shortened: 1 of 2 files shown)


crates/runtime/src/catalogconnector/spice_cloud.rs
```diff
@@ -117,7 +117,13 @@ impl SpiceCloudPlatformCatalog {
             .params
             .get("http_endpoint")
             .expose()
-            .unwrap_or_else(|_| "https://data.spiceai.io");
+            .unwrap_or_else(|_| {
+                if cfg!(feature = "dev") {
+                    "https://dev.spiceai.io".into()
```
phillipleblanc:
```suggestion
                    "https://dev-data.spiceai.io".into()
```

phillipleblanc:
I actually don't think we should tie the `dev` build of `spiceai` to the DEV endpoint of SCP. It should always point to PROD and must be overridden explicitly.

Jeadie:
> I actually don't think we should tie the dev build of spiceai to the DEV endpoint of SCP. It should always point to PROD and must be overridden explicitly.

https://github.com/spiceai/spiceai/pull/6254#issuecomment-2982147086

I will make a separate PR to remove this from flight endpoints (i.e. this mentioned [above](https://github.com/spiceai/spiceai/blob/79cecfd4c1e1dd4e3a2337d5912ebae77a3e8b18/crates/runtime/src/dataconnector/spiceai.rs#L173-L177)).

phillipleblanc (approved):

PR #6278, opened 2025-06-18
Add truncate UDF

### LINKED ISSUE #6183: Implement Scalar UDFs for Partitioning

kczimm:
Adds a new `truncate` scalar UDF.

### Key Features
- **Functionality**: The `truncate(value, width)` function truncates numeric types (`Int8`, `Int16`, `Int32`, `Int64`, `UInt8`, `UInt16`, `UInt32`, `UInt64`, `Decimal128`, `Decimal256`) to the nearest multiple of width and truncates `Utf8` and `Binary` data to the specified `width` length.
- **Signature**: Accepts two arguments: an `Int64` width and a supported value type.
- **Performance**: Uses Arrow's binary kernel for efficient array operations.
- **Safety**: Enforces a `MAX_TRUNCATE_WIDTH` constant to prevent overflow or excessive memory usage.

### Usage Example
```sql
SELECT truncate(50, CAST(10.65 AS DECIMAL(10, 2))) AS truncated; -- Returns 10.50
SELECT truncate(3, 'iceberg') AS truncated; -- Returns 'ice'
```

### DIFF
--- a/crates/runtime/src/datafusion/udf/truncate.rs
+++ b/crates/runtime/src/datafusion/udf/truncate.rs
@@ -17,7 +17,8 @@ limitations under the License.
 use std::sync::Arc;
 
 use arrow::array::{
-    ArrayRef, Decimal128Array, Decimal256Array, Int32Array, Int64Array, StringArray,
+    ArrayRef, Decimal128Array, Decimal256Array, Int8Array, Int16Array, Int32Array, Int64Array,
+    StringArray, UInt8Array, UInt16Array, UInt32Array, UInt64Array,
 };
 use arrow::compute::binary;
 use arrow::compute::kernels::substring::{substring, substring_by_char};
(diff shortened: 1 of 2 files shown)


sgrebnov (approved):

crates/runtime/src/datafusion/udf/truncate.rs
```diff
@@ -0,0 +1,722 @@
...
+        .into()),
+    }
+}
+
+fn truncate_numeric<V, W>(v: V, w: W) -> Result<V, TruncateError>
+where
+    V: Num + Copy + TryFrom<W, Error = TryFromIntError>,
+{
+    let w = V::try_from(w).context(WidthCastingFailedSnafu)?;
+    Ok(v - (((v % w) + w) % w))
```
phillipleblanc, line 245: perhaps refactor the core truncate expression into a function that can be used both here and in the binary array calculation.

phillipleblanc (approved):

crates/runtime/src/datafusion/udf/truncate.rs
kczimm: I know it occurs in several places. I tried for a while to make the truncation logic generic over the Rust primitive types and the Arrow native types with no luck.

PR #6286, opened 2025-06-20
Create a table provider from full text search index + query

Jeadie:
## 📝 Summary
 - A given full text search index and a query is enough to define a [`TableProvider`]. 
 - This PR introduces this ability, for use in future work where an index needs to be used further within datafusion.

Structs are renamed to make the distinction clearer:
 - `FullTextSearch` -> `FullTextSearchIndex`: A search index that can be used in search.
 - `FullTextSearchTable`: A table that can be used in the database, for a given search index and query.
 - `FullTextSearchExec`: The physical execution of a `FullTextSearchTable` (i.e. has given projection/filters/limit). 

## 🔗 Related
Preparation for #6240

### DIFF
--- a/crates/search/src/generation/text_search/table.rs
+++ b/crates/search/src/generation/text_search/table.rs
@@ -1,0 +1,119 @@
+/*
+Copyright 2024-2025 The Spice.ai OSS Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+     https://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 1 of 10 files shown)


peasee (approved):

crates/search/src/generation/text_search/table.rs
```diff
@@ -0,0 +1,121 @@
...
+                arrow::datatypes::DataType::Float64,
+                false,
+            )])
+            .collect::<Vec<_>>();
+
+        Arc::new(Schema::new(fields))
+    }
+
+    fn constraints(&self) -> Option<&Constraints> {
+        // TODO primary keys
```
Sevenannn, line 89: nit: link to the issue tracking the primary key implementation

Sevenannn (approved):

Sevenannn:
Question on how this works. Is a new `TableProvider` created for each user search query, or is the search query expected to be the same all the time.

Jeadie:
https://github.com/spiceai/spiceai/pull/6286#issuecomment-2992655983 

Search query will be different each time (for each call to UDTF). But once we have a query, we will need a TableProvider interface to use it with other datasets in the spice runtime (#6280 has much of this).

PR #6289, opened 2025-06-20
Better Glue connector error messages

kczimm:
## 📝 Summary

Improves the error messages to adhere to the [error handling guidelines](https://github.com/spiceai/spiceai/blob/trunk/docs/dev/error_handling.md).

### DIFF
--- a/crates/runtime/src/dataconnector/glue.rs
+++ b/crates/runtime/src/dataconnector/glue.rs
@@ -51,10 +51,12 @@ static PREFIX: &str = "glue";
 #[derive(Debug, Snafu)]
 enum Error {
     #[snafu(display(
-        "Could not retrieve table '{table}' from database '{database}'. Verify that both the database and table exist and are accessible."
+        "Cannot retrieve table '{table}' from Glue database '{database}'.\nVerify that both the database and table exist and are accessible.\nFor help with AWS Glue configuration, visit: https://docs.spiceai.org/components/data-connectors/glue"
     ))]
     GetTable { database: String, table: String },
-    #[snafu(display("Unable to load the AWS configuration.\n{source}"))]
+    #[snafu(display(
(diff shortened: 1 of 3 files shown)


crates/runtime/src/dataconnector/glue.rs
```diff
@@ -49,13 +50,51 @@ use super::{
...
+    #[snafu(display(
+        "Cannot retrieve input format for table '{table}'.\nEnsure the table is correctly configured in AWS Glue.\nFor help, visit: https://docs.spiceai.org/components/data-connectors/glue"
+    ))]
+    MissingInputFormat { table: String },
+    #[snafu(display(
+        "The input format {input_format} for table '{table}' is not supported.\nFor help, visit: https://docs.spiceai.org/components/data-connectors/glue"
+    ))]
+    InvalidInputFormat { input_format: String, table: String },
+    #[snafu(display(
+        "No storage descriptor found for table '{table}'.\nEnsure the table is correctly configured in AWS Glue.\nFor help, visit: https://docs.spiceai.org/components/data-connectors/glue"
```
Sevenannn, line 91: Is the storage descriptor a user configurable parameter? If so, we should have it listed in the documentation: https://spiceai.org/docs/components/data-connectors/glue

crates/runtime/src/dataconnector/glue.rs
```diff
@@ -49,13 +50,51 @@ use super::{
...
+    ))]
+    MissingStorageDescriptor { table: String },
+    #[snafu(display(
+        "No storage location specified for table '{table}'.\nEnsure the table has a valid S3 location in AWS Glue.\nFor help, visit: https://docs.spiceai.org/components/data-connectors/glue"
+    ))]
```
Sevenannn, line 96: Same as above

crates/runtime/src/dataconnector/glue.rs
Jeadie: Storage descriptor comes from the Glue SDK. Example input format `org.apache.hadoop.hive.ql.io.parquet.MapredParquetInputFormat`.

Jeadie (approved):

phillipleblanc (approved):

PR #6296, opened 2025-06-21
fix: Warn about no configured datasets if no datasets and catalogs are present

### LINKED ISSUE #5982: Runtime warns about no datasets configured even when there are catalogs
## Describe the bug

Regardless of the configured catalogs in the spiecpod, the runtime will display a warning if there are no datasets.

```
2025-05-26T01:58:06.439526Z  INFO runtime::init::dataset: No datasets were configured. If this is unexpected, check the Spicepod configuration.
2025-05-26T01:58:06.440052Z  INFO runtime::init::catalog: Registering catalog 'db_uc' for databricks
2025-05-26T01:58:23.927907Z  INFO runtime::init::catalog: Registered catalog 'db_uc' with 2 schemas and 32 tables
2025-05-26T01:58:24.032010Z  INFO runtime: All components are loaded. Spice runtime is ready!
```

## To Reproduce

Spicepod with non empty catalog and 0 datasets

## Expected behavior

If Spicepod has no datasets but has configured catalog connectors, it should suppress the warning.

### Spicepod

### Output of `explain query`

E.g. `explain select 1`

```bash
```

### spice, spiced, OS info

- OS info: `uname -a`

- [ ] Yes
- [ ] No

Advayp:
## 📝 Summary

- Warn about no configured datasets only when no datasets *and* no catalogs are present

## 🔗 Related

- Closes #5982 

## Output

`spicepod.yaml`:
```yaml
version: v1
kind: Spicepod
name: spice-catalog-demo
catalogs:
    - from: spice.ai/spiceai/tpch
      name: scp
      include:
          - tpch.part*
          - tpch.supplier
    - from: spice.ai/spiceai/quickstart
      name: quickstart
```

Before:
image

After:
image

### DIFF
--- a/crates/runtime/src/init/dataset.rs
+++ b/crates/runtime/src/init/dataset.rs
@@ -67,12 +67,8 @@ impl Runtime {
         };
 
         let valid_datasets = Arc::clone(&self).get_valid_datasets(app, LogErrors(true));
-        let valid_catalogs = Arc::clone(&self).get_valid_catalogs(app, LogErrors(true));
 
-        if valid_datasets.is_empty() && valid_catalogs.is_empty() {
-            tracing::info!(
-                "No datasets were configured. If this is unexpected, check the Spicepod configuration."
-            );
+        if valid_datasets.is_empty() {
             return;
         }
 
--- a/crates/runtime/src/init/dataset.rs
+++ b/crates/runtime/src/init/dataset.rs
@@ -67,8 +67,9 @@ impl Runtime {
         };
 
         let valid_datasets = Arc::clone(&self).get_valid_datasets(app, LogErrors(true));
+        let valid_catalogs = Arc::clone(&self).get_valid_catalogs(app, LogErrors(true));
 
-        if valid_datasets.is_empty() {
+        if valid_datasets.is_empty() && valid_catalogs.is_empty() {
             tracing::info!(
                 "No datasets were configured. If this is unexpected, check the Spicepod configuration."
             );
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -868,7 +868,7 @@ impl Runtime {
                                     Arc::clone(&self).get_valid_catalogs(app, LogErrors(false));
                                 if valid_datasets.is_empty() && valid_catalogs.is_empty() {
                                     tracing::info!(
-                                        "No datasets or catalogs were configured. If this is unexpected, please check the configuration of your Spicepod."
+                                        "No datasets or catalogs were configured. If this is unexpected, check the Spicepod configuration."
                                     );
                                 }
                             }
--- a/crates/runtime/src/lib.rs
+++ b/crates/runtime/src/lib.rs
@@ -863,9 +863,9 @@ impl Runtime {
                         if status.is_ready() {
                             if let Some(app) = self.app.read().await.as_ref() {
                                 let valid_datasets =
-                                    Arc::clone(&self).get_valid_datasets(app, LogErrors(true));
+                                    Arc::clone(&self).get_valid_datasets(app, LogErrors(false));
                                 let valid_catalogs =
-                                    Arc::clone(&self).get_valid_catalogs(app, LogErrors(true));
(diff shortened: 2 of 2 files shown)


crates/runtime/src/init/dataset.rs
```diff
@@ -67,8 +67,9 @@ impl Runtime {
         };
 
         let valid_datasets = Arc::clone(&self).get_valid_datasets(app, LogErrors(true));
+        let valid_catalogs = Arc::clone(&self).get_valid_catalogs(app, LogErrors(true));
 
-        if valid_datasets.is_empty() {
+        if valid_datasets.is_empty() && valid_catalogs.is_empty() {
```
peasee: It might be a better pattern to move this check out of `load_datasets` and into [`load_components`](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/lib.rs#L733)? It seems weird that we check for valid catalogs inside `load_datasets()`

crates/runtime/src/init/dataset.rs
```diff
@@ -67,8 +67,9 @@ impl Runtime {
...
             tracing::info!(
                 "No datasets were configured. If this is unexpected, check the Spicepod configuration."
```
peasee: This message probably needs to change if it is gated behind both empty datasets and catalogs.

crates/runtime/src/init/dataset.rs
peasee: actually, we would still need the check just for datasets to exit early. But the check that generates a message could move out.

crates/runtime/src/lib.rs
```diff
@@ -813,6 +813,16 @@ impl Runtime {
             }
         }
 
+        if let Some(app) = self.app.read().await.as_ref() {
```
Advayp: I'm not sure if checking this in the middle of loading the components will limit concurrency or block for too long. I don't believe it should since all locks acquired by tasks in this method seem to be read locks, but again I'm not 100% sure.

peasee: I would imagine the check would occur [here](https://github.com/spiceai/spiceai/pull/6296/files#diff-c6d4f276260e77036ce0adaf946b9ee6e55b1af1640342035121d4b254090df8L863). It could report the missing dataset/catalog message before Runtime reports ready. If the check occurs there, it should limit/avoid lock contention.

IIRC status will still `is_ready()` when there are no components loaded.

peasee (dismissed):
Looks good, thanks! Could we get a before/after of the log output in the PR description?

Advayp:
> Looks good, thanks! Could we get a before/after of the log output in the PR description?

Yup, I'll go and add that now.

Advayp:
Done @peasee !

crates/runtime/src/lib.rs
```diff
@@ -861,6 +861,17 @@ impl Runtime {
                             break;
                         }
                         if status.is_ready() {
+                            if let Some(app) = self.app.read().await.as_ref() {
+                                let valid_datasets =
+                                    Arc::clone(&self).get_valid_datasets(app, LogErrors(true));
```
phillipleblanc: I don't think we need to log errors here?

Advayp: You're right, I'll fix that.

crates/runtime/src/lib.rs
```diff
@@ -861,6 +861,17 @@ impl Runtime {
...
                         }
                         if status.is_ready() {
+                            if let Some(app) = self.app.read().await.as_ref() {
+                                let valid_datasets =
+                                    Arc::clone(&self).get_valid_datasets(app, LogErrors(false));
+                                let valid_catalogs =
+                                    Arc::clone(&self).get_valid_catalogs(app, LogErrors(false));
+                                if valid_datasets.is_empty() && valid_catalogs.is_empty() {
+                                    tracing::info!(
+                                        "No datasets or catalogs were configured. If this is unexpected, please check the configuration of your Spicepod."
```
```suggestion
                                        "No datasets or catalogs were configured. If this is unexpected, check the Spicepod configuration."
```

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #6298, opened 2025-06-22
Add partitioning models 

### LINKED ISSUE #6184: Partitioning Data Structures and Models

kczimm:
## 📝 Summary

This PR introduces support for partitioning in the `acceleration` section of the spicepod configuration.

- Added `partition_by: Vec<String>` to store partitioning expressions, with a method `partition_by_expressions` to convert them to DataFusion `Expr`s.
- Modified the `DataAccelerator` trait to accept `partition_by: Vec<Expr>` in `create_external_table`. Updated implementations (`arrow.rs`, `duckdb.rs`, `postgres.rs`, `sqlite.rs`) to handle this new parameter.
- Introduced a `partition` module in the `runtime` crate with a `PartitionTableProvider`, `PartitionCreator`, and `PartitionedTableInsert` for handling partitioned tables. These are currently incomplete.
- Introduced an `expression` module that converts the `partition_by` parameter into a list of `Expr`s and validates them using various criteria.

## 🔗 Related

Closes:
- #6184


crates/runtime/src/dataaccelerator/arrow.rs
```diff
@@ -63,6 +63,7 @@ impl DataAccelerator for ArrowAccelerator {
         &self,
         cmd: CreateExternalTable,
         _source: Option<&dyn AccelerationSource>,
+        _partition_by: Vec<Expr>,
```
phillipleblanc: I would have this return an error for now if its provided but not supported/used.

crates/runtime/src/dataaccelerator/duckdb.rs
```diff
@@ -365,26 +368,33 @@ impl DataAccelerator for DuckDBAccelerator {
...
-        let table_provider = TableProviderFactory::create(&self.duckdb_factory, &ctx.state(), &cmd)
-            .await
-            .context(UnableToCreateTableSnafu)
-            .boxed()?;
-
-        let Some(duckdb_writer) = table_provider.as_any().downcast_ref::<DuckDBTableWriter>()
-        else {
-            unreachable!("DuckDBTableWriter should be returned from DuckDBTableProviderFactory")
+        let table_provider: Arc<dyn TableProvider> = if partition_by.is_empty() {
+            let ctx = SessionContext::new();
```
phillipleblanc: Won't we need this logic to create the PolyTableProvider in the PartitionTableProvider?

crates/runtime/src/partition/mod.rs
phillipleblanc, line 1: Let's make this a separate crate called `runtime-table-partition`

crates/runtime/src/partition/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+    common::Constraints,
+    datasource::TableType,
+    error::DataFusionError,
+    execution::{SendableRecordBatchStream, TaskContext},
+    logical_expr::CreateExternalTable,
+    physical_plan::{DisplayAs, DisplayFormatType, ExecutionPlan, PlanProperties},
+    prelude::Expr,
+};
+
+// [`datafusion::scalar::ScalarValue`] does not necessary traits for Hash key.
```
phillipleblanc: grammar

crates/runtime/src/partition/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+type ScalarValueString = String;
+
+#[derive(Debug)]
+pub struct PartitionTableProvider {
+    schema: SchemaRef,
+    _partition_by: Vec<Expr>,
+    _partitions: HashMap<ScalarValueString, Arc<dyn TableProvider>>,
```
phillipleblanc: The partition key can also be numerics right?

crates/runtime/src/partition/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+    _creator: PartitionCreator,
+}
+
+#[derive(Debug)]
+struct PartitionCreator {}
+
+impl PartitionTableProvider {
+    #[must_use]
+    pub fn new(_partition_by: Vec<Expr>, _cmd: &CreateExternalTable) -> Self {
```
phillipleblanc: I don't think `CreateExternalTable` is the right interface here. As mentioned in another comment, we need to create the PolyTableProvider properly which this wouldn't. I think this should be a lambda that takes a `location` and then creates a `dyn TableProvider`. Or maybe this should be the PartitionCreator itself that is passed in? That might make more sense, if you want to create a PartitionTableProvider, you need to provide the struct that knows how to create partitions.

phillipleblanc: And I would see if you can make the `PartitionCreator` itself a trait object instead of a struct.

crates/runtime/src/partition/mod.rs
phillipleblanc: The runtime crate is large enough as is, let's take every opportunity to let the compiler work in parallel.

phillipleblanc: It also means we can have separate integration tests for just this crate.

crates/runtime/src/component/dataset/acceleration.rs
```diff
@@ -300,6 +307,17 @@ impl Acceleration {
...
     }
+
+    pub fn partition_by_expressions(
+        &self,
+        ctx: &SessionContext,
+        df_schema: &DFSchema,
+    ) -> Result<Vec<Expr>, DataFusionError> {
+        self.partition_by
+            .iter()
+            .map(|sql| ctx.parse_sql_expr(sql, df_schema))
```
phillipleblanc: This will need to enforce the restrictions on partition expressions from the spec, i.e. exactly one column reference, etc.

crates/runtime/src/partition/mod.rs
kczimm: Yes, my garbled comment about the key was originally intended to say the following: ScalarValue does not implement the necessary traits to be used as a key in the HashMap, so ScalarValueString was meant to be the stringified version of ScalarValue which can be used as the key.

crates/runtime/src/component/dataset/acceleration.rs
kczimm: added an `expression` module that builds validation criteria and does this conversion.

crates/runtime/src/dataaccelerator/duckdb.rs
kczimm: yes. I'll work on the integration of these new types into duckdb in a follow up PR.

sgrebnov (dismissed):

crates/runtime-table-partition/src/expression.rs
```diff
@@ -0,0 +1,259 @@
...
+    /// # Errors
+    /// Returns an error if the validation failed or cannot complete.
+    fn validate(&self, expr: &Expr, schema: &DFSchema) -> ValidationResult;
+}
+
+pub struct PartitionCriteria;
+
+impl Criterion for PartitionCriteria {
+    fn validate(&self, expr: &Expr, schema: &DFSchema) -> ValidationResult {
+        let criteria: Vec<&dyn Criterion> = vec![
```
sgrebnov: Nice 👍

Sevenannn (dismissed):

crates/runtime-table-partition/src/expression.rs
```diff
@@ -0,0 +1,259 @@
...
+struct SingleColumnCriterion;
+
+impl Criterion for SingleColumnCriterion {
+    fn validate(&self, expr: &Expr, _schema: &DFSchema) -> ValidationResult {
+        let num_columns = expr.column_refs().len();
+        ensure!(
+            num_columns == 1,
+            CriterionFailedSnafu {
+                expr: expr.to_string(),
+                message: format!("Expression references {num_columns}, expected exactly 1")
```
phillipleblanc:
```suggestion
                message: format!("Expression references {num_columns} columns, expected exactly 1")
```

crates/runtime-table-partition/src/expression.rs
```diff
@@ -0,0 +1,259 @@
...
+        let expr = case(col("a"))
+            .when(
+                regexp_match(col("a"), lit("^DATAFUSION(-cli)*"), None),
+                lit("datafusion"),
+            )
+            .otherwise(lit("other"))
+            .expect("expression created");
+        assert!(criterion.validate(&expr, &schema).is_ok());
+
+        // Invalid: date_trunc('month', date)
```
phillipleblanc: I would expect this to work - is it just out of scope for this implementation?

crates/runtime-table-partition/src/expression.rs
```diff
@@ -0,0 +1,259 @@
...
+    logical_expr::ExprSchemable,
+    prelude::{Expr, SessionContext},
+};
+use snafu::prelude::*;
+
+#[derive(Debug, Snafu)]
+pub enum ValidationError {
+    #[snafu(display("Failed to determine data type: {source}"))]
+    DataTypeError { source: DataFusionError },
+    #[snafu(display("Expression {expr} does not meet criterion: {message}"))]
```
phillipleblanc: We should list all of the criteria we expect the expression to match here. Its frustrating to be told something is wrong and then fix that one thing and find out something else is wrong.

crates/runtime-table-partition/src/provider.rs
```diff
@@ -0,0 +1,96 @@
...
+    ///
+    /// # Errors
+    /// Returns an error if partition inferencing fails.
+    pub async fn new(
+        creator: Arc<dyn PartitionCreator>,
+        partition_by: Vec<Expr>,
+        schema: SchemaRef,
+    ) -> Result<Self, Error> {
+        let partitions = creator
+            .infer_existing_partitions()
```
phillipleblanc, line 76: We should have validation here that the inferred partitions actually match the valid output of the expression. Ideally the actual possible values, but at least the data type.

phillipleblanc (dismissed):

crates/runtime-table-partition/src/expression.rs
kczimm: I allowed DataType::Timestamp now.

crates/runtime-table-partition/src/expression.rs
kczimm: Should look like this when printed with Display:

```
Expression Int32(42) does not meet the criteria: Expression references 0 columns, expected exactly 1

Expression Criteria:
- data type must be a String, Number, Boolean or Timestamp
- expression must reference a single column
- expression must not contain Alias, OuterReferenceColumn, Unnest, WindowFunction,
  AggregateFunction, Exists, InSubquery, ScalarSubquery, Placeholder, or
  GroupingSet
```

phillipleblanc (approved):

PR #6317, opened 2025-06-25
Surface table not found errors immediately

### LINKED ISSUE #6315: Table not found error misclassified as internal when running `spice sql`
## Describe the bug

When running `spice sql`, executing a query on a table that does not exist produces an error that is incorrectly classified as internal. Moreover, this error isn't surfaced immediately to the user, which is problematic because the error is related to the query the user entered.

## To Reproduce

1. Start the Spice runtime
2. Run `spice sql` in a separate terminal
3. Execute a query involving a table that does not exist
4. Observe that a vague error message is provided that indicates the error is internal
5. Run `.error` to retrieve the details of the error, noting that it is not internal as it involves the user's query

## Expected behavior

The error should be surfaced immediately, like so:
```
sql> select * from non_existent_table;
Error during planning: table 'spice.public.non_existent_table' not found
```

### Spicepod

```yaml
version: v1
kind: Spicepod
name: example
```

### spice, spiced, OS info

- OS info: `uname -a`

- [x] Yes
- [ ] No

## Screenshots

Image

Advayp:
## 📝 Summary

- In `spice sql`, if a user entered a table name that does not exist, the error was misclassified as an internal error.
- This PR surfaces the table not found error immediately

## 🔗 Related
- Closes #6315 

## 👀 Notes for Reviewers
- Based on what I saw, I believe the root cause of the misclassification is within Arrow Flight (not in this repo)
- I found a workaround, but I'm unsure if there's a better way of checking whether or not the error is a table not found error
- Any suggestions here are appreciated

## Output

Before:
image

After:
image

### DIFF
--- a/crates/flightrepl/src/lib.rs
+++ b/crates/flightrepl/src/lib.rs
@@ -545,10 +545,48 @@ fn lines_need_truncation(lines: &[&str]) -> bool {
     lines.iter().any(|line| line.len() > 120)
 }
 
+fn is_table_not_found_error(err: &Status) -> (bool, Option<String>) {
+    let message = err.message();
+    if message.contains("table '") && message.contains("' not found") {
+        // Find the start of the table name after "table '"
+        if let Some(start) = message.find("table '") {
+            let start_pos = start + "table '".len();
+            // Find the end of the table name at the next single quote
+            if let Some(end_pos) = message[start_pos..].find('\'') {
+                let table_name = &message[start_pos..start_pos + end_pos];
+                return (true, Some(table_name.to_string()));
+            }
+        }
+        // Found the pattern but couldn't extract the table name
+        (true, None)
+    } else {
+        (false, None)
+    }
+}
+
 fn display_grpc_error(err: &Status) {
     let (error_type, user_err_msg) = match err.code() {
         Code::Ok => return,
-        Code::Unknown | Code::Internal | Code::DataLoss | Code::FailedPrecondition => (
+        Code::Internal => {
+            let (is_table_not_found, table_name) = is_table_not_found_error(err);
+            if is_table_not_found {
+                let message = if let Some(name) = table_name {
+                    format!(
+                        "The table '{name}' does not exist. Check that the table name was entered correctly or verify the configuration in the Spicepod."
+                    )
+                } else {
+                    "The specified table does not exist. Check that the table name was entered correctly or verify the configuration in the Spicepod.".to_string()
+                };
+                ("Table Not Found", message)
+            } else {
+                (
+                    "Internal Error",
+                    "An unexpected internal error occurred. Execute '.error' for details."
+                        .to_string(),
+                )
+            }
+        }
+        Code::Unknown | Code::DataLoss | Code::FailedPrecondition => (
             "Internal Error",
             "An unexpected internal error occurred. Execute '.error' for details.".to_string(),
         ),
--- a/crates/flightrepl/src/lib.rs
+++ b/crates/flightrepl/src/lib.rs
@@ -567,25 +567,10 @@ fn is_table_not_found_error(err: &Status) -> (bool, Option<String>) {
 fn display_grpc_error(err: &Status) {
     let (error_type, user_err_msg) = match err.code() {
         Code::Ok => return,
-        Code::Internal => {
-            let (is_table_not_found, table_name) = is_table_not_found_error(err);
-            if is_table_not_found {
-                let message = if let Some(name) = table_name {
-                    format!(
-                        "The table '{name}' does not exist. Check that the table name was entered correctly or verify the configuration in the Spicepod."
-                    )
-                } else {
-                    "The specified table does not exist. Check that the table name was entered correctly or verify the configuration in the Spicepod.".to_string()
-                };
-                ("Table Not Found", message)
-            } else {
-                (
-                    "Internal Error",
-                    "An unexpected internal error occurred. Execute '.error' for details."
-                        .to_string(),
-                )
-            }
-        }
+        Code::Internal => (
+            "Internal Error",
+            "An unexpected internal error occurred. Execute '.error' for details.".to_string(),
+        ),
         Code::Unknown | Code::DataLoss | Code::FailedPrecondition => (
             "Internal Error",
             "An unexpected internal error occurred. Execute '.error' for details.".to_string(),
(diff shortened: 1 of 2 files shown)


crates/flightrepl/src/lib.rs
```diff
@@ -545,10 +545,48 @@ fn lines_need_truncation(lines: &[&str]) -> bool {
     lines.iter().any(|line| line.len() > 120)
 }
 
+fn is_table_not_found_error(err: &Status) -> (bool, Option<String>) {
+    let message = err.message();
+    if message.contains("table '") && message.contains("' not found") {
+        // Find the start of the table name after "table '"
+        if let Some(start) = message.find("table '") {
```
kczimm: It's not great that the type of error is essentially extracted from a String but it may be the best we can do, at least here. Perhaps we can bring in the regex crate from the workspace into this crate and get a more robust match.

Also, when would we match "table" and "not found" but not get a table name out of it?

crates/flightrepl/src/lib.rs
```diff
@@ -545,10 +545,48 @@ fn lines_need_truncation(lines: &[&str]) -> bool {
...
+                    "The specified table does not exist. Check that the table name was entered correctly or verify the configuration in the Spicepod.".to_string()
+                };
+                ("Table Not Found", message)
+            } else {
+                (
+                    "Internal Error",
+                    "An unexpected internal error occurred. Execute '.error' for details."
+                        .to_string(),
+                )
+            }
```
kczimm: There is quite a bit of nesting matches/if statements here. [Match guards](https://doc.rust-lang.org/rust-by-example/flow_control/match/guard.html) might be useful.

kczimm (commented):
It's possible we can refactor both functions and use a more expressive type to convert Status::code() and Status::message() into a new type with variants that when printed yield the output we have here from `error_type` and `user_err_msg`. 

```rust
enum GrpcError {
    TableNotFound { 
        table_name: String // assuming we can always get a table name
    },
    Internal,
    Query,
    Cancelled,
    Aborted,
    Timeout,
    Authentication,
    Authorization,
    ResultLimit,
    ResourceLimit, 
    Unsupported,
}

impl From<&Status> for GrpcError {
   fn from(err: &Status) -> Self {
      ...
   }
}

impl fmt::Display for GrpcError {
    ...
}
```

This is probably more idiomatic. Then we can actually remove the `display_grpc_error` function and just convert the `Status` and print it.

Curious what your thoughts are here.

crates/flightrepl/src/lib.rs
Advayp: Yeah, regex would be ideal here.

> Also, when would we match "table" and "not found" but not get a table name out of it?

I was thinking of the case the error message changes and "not found" comes before "table" instead of after (like my function is assuming). Though, of course, when the error message changes this function no longer works. I think I'll modify the function to operate under the assumption that "not found" comes after "table." In the (hopefully rare) case the error message does change, we will need to rewrite this function.

Advayp:
> It's possible we can refactor both functions and use a more expressive type to convert Status::code() and Status::message() into a new type with variants that when printed yield the output we have here from `error_type` and `user_err_msg`.
> 
> ```rust
> enum GrpcError {
>     TableNotFound { 
>         table_name: String // assuming we can always get a table name
>     },
>     Internal,
>     Query,
>     Cancelled,
>     Aborted,
>     Timeout,
>     Authentication,
>     Authorization,
>     ResultLimit,
>     ResourceLimit, 
>     Unsupported,
> }
> 
> impl From<&Status> for GrpcError {
>    fn from(err: &Status) -> Self {
>       ...
>    }
> }
> 
> impl fmt::Display for GrpcError {
>     ...
> }
> ```
> 
> This is probably more idiomatic. Then we can actually remove the `display_grpc_error` function and just convert the `Status` and print it.
> 
> Curious what your thoughts are here.

I like that idea. I think it might be possible to generalize this further into a `FlightReplError` enum. This is because our error handling when making a flight query is the following:
```rust
Err(FlightError::Tonic(status)) => {
                display_grpc_error(&status);
                last_error = Some(*status);
            }
Err(e) => { // e is of type FlightError
    println!(
        "Unexpected Flight Error {}",
        Colour::Red.paint(e.to_string())
    );
}
```
Instead of just implementing `From<&Status>` we can implement `From<&FlightError>` and not have to separate these two match arms. 

When I code this up, I'll use Snafu and follow Spice's error handling guidelines, which will take care of the `Display` trait.

Let me know what you think.

kczimm:
> Instead of just implementing `From<&Status>` we can implement `From<&FlightError>` and not have to separate these two match arms.

Great idea!

crates/flightrepl/src/lib.rs
```diff
@@ -545,77 +686,16 @@ fn lines_need_truncation(lines: &[&str]) -> bool {
...
-        ),
-    };
-
-    println!(
-        "{} {user_err_msg}",
-        Colour::Red.paint(format!("{error_type}:"))
-    );
+fn is_table_not_found_error(err: &Status) -> (bool, String) {
+    let message = err.message();
+    if message.contains("table '") && message.contains("' not found") {
```
lukekim: We shouldn't be special casing this. We should have better logic for categorizing errors between user, dependency, and system, etc. and always showing all user errors directly into the REPL.

@phillipleblanc @peasee

Jeadie: The tonic status error roughly looks like this

```rust
FlightError::Tonic(tonic::Status{
    code: Code,
    message: String,
    details: Bytes,
    metadata: MetadataMap,
    source: Option<Arc<dyn Error + Send + Sync + 'static>>
})
```

We could utilise the `MetadataMap` with spice.ai specific properties. This would resolve it, but would require some refactoring of our flight service.

phillipleblanc: The REPL currently decides whether something is a user error or an internal error based on the `code`. A table not existing should be an INVALID_ARGUMENT error (https://grpc.io/docs/guides/status-codes/) which the REPL already knows how to display.

The underlying issue then is that we aren't returning INVALID_ARGUMENT for tables not existing from the Spice flight service.

crates/runtime/src/flight/mod.rs
```diff
@@ -316,7 +316,32 @@ fn handle_datafusion_error(e: DataFusionError) -> Status {
...
+            // Since DataFusionError doesn't implement Clone, we can't extract it from Arc
+            // Just treat it as a generic error
+            Status::internal(format!("Shared DataFusion error: {source}"))
+        }
+        DataFusionError::Collection(sources) => {
+            let first_error = sources.into_iter().next();
+            if let Some(first_error) = first_error {
+                handle_datafusion_error(first_error)
+            } else {
+                unreachable!("DataFusionError::Collection should always contain at least one error")
```
phillipleblanc: Let's not panic here if this happens, just return an internal error

peasee (approved):

Sevenannn (approved):

phillipleblanc (approved):

PR #6320, opened 2025-06-25
[WIP] Add initial Spicepod configuration for an error-message-checking model 

Advayp:
## 📝 Summary
- See title.
- **Motivation**: We want to have consistent and strong error messages across the codebase. 
-  One way to do so is through Spice itself, creating an app on SCP, connecting it to this repo, and integrating a model provider to evaluate errors.
- This PR creates the spicepod necessary for that.


spicepod.yaml
```diff
@@ -0,0 +1,35 @@
...
+      params:
+          include: '**/*.rs'
+
+models:
+    - from: openai:gpt-4.1
+      name: error_checker
+      description: Ensures errors across the repo are consistent
+      params:
+          tools: auto
+          endpoint: https://api.openai.com/v1
```
Jeadie: `endpoint` not needed when using OpenAI.

Sevenannn:
I'm not sure if we want a spicepod.yaml defined in the root of the spiceai repo. Can you add more description on why this is needed and how it would be used for later tasks?

Issue #6314, opened 2025-06-25
Enhancement: Oracle Data Connector


sgrebnov:
## Goal-State/What/Result

Spice can be used to query and accelerate datasets stored in Oracle Database:
 - Oracle On-Prem hosted Databases
 - Oracle Cloud User-Managed Databases
 - Oracle Cloud Autonomous Databases (Serverless)

## Why/Purpose

There is a large community of Oracle users attempting to integrate with Spice, but the lack of a dedicated connector creates a significant barrier.

## By When

**Issue/Spec written and reviewed:** 06/24
**Done-Done:** 07/14

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

`oracle:` prefix is used to define Oracle Database data connector

```yaml
# On-premise Oracle instance
datasets:
  - from: oracle:"SH"."PRODUCTS"
    name: sh.products
    params: 
      oracle_host: 127.0.0.1
      oracle_username: scott
      oracle_password: tiger
```

### Auth / Configuration

Support login based authentication only. `OS Authentication` is out of scope for alpha release.

**oracle_host** - the hostname or IP address of the Oracle Database instance. Requried
**oracle_port** - the host port of the Microsoft SQL Server instance. Optional. Default is `1521`
**oracle_username** - the username. Required
**oracle_password** - the password associated with the specified SQL user
**oracle_service_name** - the Oracle Database service name to connect to. Optional. Default to `XEPDB1`
Not supported by Spice, as **oracle_service_name** is modern and recommended configuration, supported since Oracle v9 (2001)
**oracle_connection_string** - the raw connection string (TNS alias from tnsnames.ora for mTLS/Wallet or Easy Connect strings for TLS connection). Note: oracle python library name it as `dsn`
[...]

PR #6321, opened 2025-06-26
Add initial support for Oracle Data Connector

sgrebnov:
## 📝 Summary

PR adds initial support for Oracle Data Connector based on [Oracle](https://crates.io/crates/oracle) rust crate. Oracle has own specific SQL Dialect that is not supported by DataFusion unparser so `OracleTableProvider` implementation is based on own ExecutionPlan. First version does not have support of optimizations / pushdowns that will be added next. To enable this functionality, the `oracle` feature must be enabled, as it is not included by default.

```yaml
datasets:
  - from: oracle:"SALES_DATA"
    name: sales_data
    params: 
      oracle_host: localhost
      oracle_username: system
      oracle_password: ${ SPICE_ORACLE_PASSWORD }
```

**Supported:**
1. Configuration params:
    `oracle_host` - the hostname or IP address of the Oracle Database instance. Requried
    `oracle_port` - the host port of the Oracle Database instance. Optional. Default is `1521`
    `oracle_username` - the username. Required
    `oracle_password` - the password associated with the specified SQL user
    `oracle_service_name` - the Oracle Database service name to connect to. Optional. Default to `XEPDB1`
2. Connection based on connection pool
3. Streaming query interface
4. Limited basic types and native schema inference
5. NO pushdowns are currently supported

**The following will be implemented next:**
1. Support for all core Oracle Data Types
2. Pushdown support for LIMIT and WHERE predicates
3. Tests

## 🔗 Related


Cargo.toml
```diff
@@ -71,8 +71,9 @@ aws-config = "1.6.1"
 aws-credential-types = "1.2"
 axum = { version = "0.8", features = ["macros"] }
 base64 = "0.22.1"
-bb8 = "0.8"
-bb8-postgres = "0.8"
+bb8 = "0.9"
```
sgrebnov, line 75: `bb8` has been upgraded to be the same version with `bb8-oracle` crate

crates/data_components/src/mssql/connection_manager.rs
```diff
@@ -51,15 +51,23 @@ impl bb8::ManageConnection for SqlServerConnectionManager {
     type Connection = Client<Compat<TcpStream>>;
     type Error = tiberius::error::Error;
 
-    async fn connect(&self) -> Result<Self::Connection, Self::Error> {
-        let tcp = TcpStream::connect(&self.config.get_addr()).await?;
-        tcp.set_nodelay(true)?;
-        Client::connect(self.config.clone(), tcp.compat_write()).await
+    fn connect(&self) -> impl Future<Output = Result<Self::Connection, Self::Error>> + Send {
```
sgrebnov, line 54: related to bb8 0.8->0.9 upgrade (api changed)

crates/data_components/src/oracle/convert.rs
```diff
@@ -0,0 +1,217 @@
...
+    datatypes::{DataType, SchemaRef},
+};
+use bigdecimal::BigDecimal;
+use oracle::Row;
+use snafu::OptionExt;
+use snafu::ResultExt;
+
+/// Oracle Built-in Data Types
+/// `<https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-7B72E154-677A-4342-A1EA-C74C1EA928E6>`
+pub(crate) fn map_oracle_type_to_arrow_type(
```
sgrebnov, line 35: This is initial implementation that will be extended and revised next, including type conversion tests

crates/data_components/src/oracle/convert.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    for row in rows {
+        for (idx, field) in schema.fields.iter().enumerate() {
+            let builder = &mut arrow_columns_builders[idx];
+
+            let Some(col) = row.column_info().get(idx) else {
+                return Err(super::Error::NoColumnForIndex { index: idx });
+            };
+            let native_type = col.oracle_type();
```
sgrebnov, line 100: This is initial implementation, remaining types will be added next

crates/data_components/src/oracle/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        tracing::trace!("Retrieved dataset {table} schema: {fields:?}");
+
+        Ok(Arc::new(Schema::new(fields)))
+    }
+}
+
+#[async_trait]
+impl TableProvider for OracleTableProvider {
```
sgrebnov, line 192: `supports_filters_pushdown` implementation functionality will be added next

crates/data_components/src/oracle/execution_plan.rs
```diff
@@ -0,0 +1,211 @@
...
+                EquivalenceProperties::new(projected_schema),
+                Partitioning::UnknownPartitioning(1),
+                EmissionType::Incremental,
+                Boundedness::Bounded,
+            ),
+        })
+    }
+
+    #[allow(clippy::unnecessary_wraps)]
+    pub fn sql(&self) -> DataFusionResult<String> {
```
sgrebnov, line 79: To be extended with filter expressions (where clause) and limit support next, including tests

crates/data_components/src/oracle/connection.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+#[derive(Debug, Clone)]
+pub struct OracleConnectionParams {
+    pub host: String,
+    pub port: u16,
+    pub username: String,
+    pub password: String,
+    pub service_name: String,
+}
```
peasee, line 46:
```suggestion
pub struct OracleConnectionParams {
    pub host: Arc<str>,
    pub port: u16,
    pub username: Arc<str>,
    pub password: Arc<str>,
    pub service_name: Arc<str>,
}
```

nit: if these don't change after they are built, can they be `Arc<str>`?

sgrebnov: Cool, I'll consider this approach, thank you!

bin/spiced/Cargo.toml
```diff
@@ -88,6 +88,7 @@ models = ["runtime/models"]
 mssql = ["runtime/mssql"]
 mysql = ["runtime/mysql"]
 odbc = ["runtime/odbc"]
+oracle = ["runtime/oracle"]
```
phillipleblanc, line 91: Can you add the `oracle` feature to the check all features workflow test?

crates/data_components/src/oracle/connection.rs
phillipleblanc: If it isn't beneficial, then can we remove the `Clone` derive? If we're planning to clone this config, then we should use Arc<str>

crates/data_components/src/oracle/execution_plan.rs
```diff
@@ -0,0 +1,211 @@
...
+    pool: Arc<OracleConnectionPool>,
+    sql: String,
+    projected_schema: SchemaRef,
+) -> DataFusionResult<SendableRecordBatchStream> {
+    let schema = Arc::clone(&projected_schema);
+
+    let conn = pool.get().await.map_err(to_datafusion_err)?;
+
+    let query_res = conn
+        .query(&sql, &[])
```
phillipleblanc, line 164: Is this a streaming query?

crates/data_components/src/oracle/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+            schema,
+            table: table.clone(),
+        })
+    }
+
+    pub async fn get_schema(
+        conn: Arc<OracleConnectionPool>,
+        table: &TableReference,
+    ) -> Result<SchemaRef> {
+        let table_name = table.table();
```
phillipleblanc, line 123: Should we provide an error if `catalog` is set since Oracle doesn't support it.

crates/data_components/src/oracle/execution_plan.rs
sgrebnov: My understanding that it is a streamed-based implementation for query - there is iterator that fetches new records  by requests - Lazy Fetching with Buffer Management
https://github.com/kubo/rust-oracle/blob/master/src/statement.rs#L434C42-L434C52

```rust
impl<T> Iterator for ResultSet<'_, T>
where
    T: RowValue,
{
    type Item = Result<T>;

    fn next(&mut self) -> Option<Self::Item> {
        self.stmt_mut()
            .next()
            .map(|row_result| row_result.and_then(|row| row.get_as::<T>()))
    }
}
```

```rust
fn try_next(&mut self) -> Result<Option<&Row>> {
        let index = self.shared_buffer_row_index.load(Ordering::Relaxed);
        let last_index = self.last_buffer_row_index;
        if index + 1 < last_index {
            self.shared_buffer_row_index
                .store(index + 1, Ordering::Relaxed);
            Ok(Some(self.row.as_ref().unwrap()))
        } else if self.more_rows && self.fetch_rows()? {
            Ok(Some(self.row.as_ref().unwrap()))
        } else {
            Ok(None)
        }
    }
```

crates/data_components/src/oracle/mod.rs
sgrebnov: Added the following check, thank you
```
if table.catalog().is_some() {
            return Err(Error::SchemaRetrievalCatalogsUnsupported {
                table: table.to_string(),
            });
        }
```

crates/data_components/src/oracle/connection.rs
sgrebnov: Removed, thank you

bin/spiced/Cargo.toml
sgrebnov: Added, thank you

phillipleblanc (approved):

PR #6323, opened 2025-06-27
Oracle connector: Support all major Oracle data types

sgrebnov:
## 📝 Summary

Add support for all main Oracle types except `INTERVAL YEAR` and `INTERVAL DAY`
 - https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-1BABC478-FB47-4962-9B0C-8B8BD059E733

`TIMESTAMP WITH LOCAL TIME ZONE`  relies on connection session time zone information to return converted  timestamp information (similar to MySQL). Spice defaults to UTC when no time zone information is specified so Oracle connection pool has been updated to set UTC time zone for Spice connections.

## 🔗 Related


crates/data_components/src/oracle/connection.rs
```diff
@@ -104,6 +125,10 @@ pub async fn connect(params: &OracleConnectionParams) -> super::Result<OracleCon
     let manager = OracleConnectionManager::from_connector(connector);
 
     let pool = bb8::Pool::builder()
+        // Spice default Timezone is UTC. Set it to correctly handle TIMESTAMP WITH LOCAL TIME ZONE data types
```
peasee: Is this a documented default? Cron schedules evaluate to the system time of the Runtime.

sgrebnov: Updated so it is more clear that we are talking about datetime/timestamp here: `Spice uses UTC timezone for timestamp data. Set preferred timezone for automated datatype conversion to correctly handle TIMESTAMP WITH LOCAL TIME ZONE data types`

https://spiceai.org/docs/reference/timestamp
>In Spice all timestamps are represented as an integer value denoting the number of seconds that have passed since the Unix epoch **in UTC time.**

This is similar to MySQL
[Handle all MySQL Timestamp types as UTC time](https://github.com/datafusion-contrib/datafusion-table-providers/pull/174)

phillipleblanc (dismissed):

lukekim (approved):

Issue #6325, opened 2025-06-28
Oracle Connector: support for Datetime filter predicates pushdown


sgrebnov:
## Describe the bug

Filter predicate pushdown is disabled for Datetime/Timestamp columns because Oracle requires a specific format for datetime literals to cast them correctly. Currently, the expression unparser cannot generate valid Oracle timestamp expressions.

>SELECT * FROM "TEST_TABLE" WHERE ("VAL_TIMESTAMP" >= CAST('2024-06-27 00:00:00' AS TIMESTAMP))

>ORA-01843: not a valid month
https://docs.oracle.com/error-help/db/ora-01843/01843. 00000 -  "An invalid month was specified."
*Cause:    You specified a date with an invalid month.
*Action:   Enter a valid month value using either the correctly spelled full
           month name or valid short month code.

/ Oracle requires a specific format for datetime literals to correctly cast them to timestamps.
                        // Currently, the expression unparser cannot handle timestamps, resulting in
                        // an `ORA-01843: not a valid month` error.

## To Reproduce

```
- from: oracle:"TEST_TABLE"
    name: table1
    params: 
      oracle_host: localhost
      oracle_username: ${secrets:SPICE_ORACLE_USER}
      oracle_password: ${secrets:SPICE_ORACLE_PASSWORD}
```

```
sql> describe table1;
+------------------------+------------------------------------+-------------+
| column_name            | data_type                          | is_nullable |
+------------------------+------------------------------------+-------------+
| ID                     | Decimal128(38, 0)                  | YES         |
| VAL_NUMBER             | Decimal128(10, 2)                  | YES         |
| VAL_DECIMAL            | Decimal128(18, 4)                  | YES         |
| VAL_FLOAT              | Float32                            | YES         |
| VAL_DOUBLE             | Float64                            | YES         |
| VAL_CHAR               | Utf8                               | YES         |
| VAL_NCHAR              | Utf8                               | YES         |
| VAL_VARCHAR2           | Utf8                               | YES         |
| VAL_NVARCHAR2          | Utf8                               | YES         |
| VAL_CLOB               | LargeUtf8                          | YES         |
| VAL_NCLOB              | LargeUtf8                          | YES         |
| VAL_DATE               | Date32                             | YES         |
| VAL_TIMESTAMP          | Timestamp(Nanosecond, None)        | YES         |
| VAL_TIMESTAMP_TZ       | Timestamp(Nanosecond, Some("UTC")) | YES         |
[...]

PR #6326, opened 2025-06-28
Oracle connector: support filter predicate pushdown

sgrebnov:
## 📝 Summary

Add support for `LIMIT` and filter predicates to `OracleTableProvider`. This enables more efficient federated queries and data loading when using acceleration with `refresh_sql` with predicates.

```
- from: oracle:"TEST_TABLE"
    name: table1
    params: 
      oracle_host: localhost
      oracle_username: ${secrets:SPICE_ORACLE_USER}
      oracle_password: ${secrets:SPICE_ORACLE_PASSWORD}
    acceleration:
      enabled: true
      refresh_sql: |
        SELECT * FROM table1
        WHERE "VAL_NUMBER" = 123.45
          AND "VAL_FLOAT" > 2.8
          AND "VAL_DECIMAL" < 10000
          AND "VAL_VARCHAR2" = 'ghi'
          AND "VAL_BOOL" = 'Y'
        limit 10;
```
This above is translated into single SQL query to fetch the data:

```sql
SELECT .. FROM "TEST_TABLE" WHERE ("VAL_NUMBER" = 123.45) AND ("VAL_FLOAT" > 2.8) AND ("VAL_DECIMAL" < 10000.0000) AND ("VAL_VARCHAR2" = 'ghi') AND ("VAL_BOOL" = 'Y') FETCH FIRST 10 ROWS ONLY
```

Note: Datetime filters are not currently supported (not pushed down) and tracked as the following work item.
- https://github.com/spiceai/spiceai/issues/6325

## 🔗 Related

### DIFF
--- a/crates/data_components/src/oracle/mod.rs
+++ b/crates/data_components/src/oracle/mod.rs
@@ -18,10 +18,12 @@ use async_trait::async_trait;
 use bigdecimal::BigDecimal;
 use snafu::{ResultExt, Snafu};
 
-use arrow::datatypes::{Field, Schema, SchemaRef};
+use arrow::datatypes::{DataType, Field, Schema, SchemaRef};
 use datafusion::{
     catalog::Session,
     datasource::{TableProvider, TableType},
+    error::DataFusionError,
+    logical_expr::{Operator, TableProviderFilterPushDown},
     physical_plan::ExecutionPlan,
     prelude::Expr,
     sql::TableReference,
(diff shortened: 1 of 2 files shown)


crates/data_components/src/oracle/mod.rs
```diff
@@ -219,3 +255,27 @@ impl TableProvider for OracleTableProvider {
         )?))
     }
 }
+
+fn is_time_related_expr(expr: &Expr) -> bool {
+    match expr {
+        Expr::Cast(cast) => {
+            matches!(
+                cast.data_type,
+                DataType::Time32(_) | DataType::Time64(_) | DataType::Timestamp(_, _)
```
peasee: should this include `Date32`, `Date64`, `Duration` or `Interval`?

sgrebnov: @peasee - missed `Date` type, which is required (not supported properly), included now - thank you. `Duration` type is not supported in Oracle data schema / can be in `filters` expr.

peasee (approved):

PR #6327, opened 2025-06-28
Improve type conversion and add integration tests for the Oracle connector

### LINKED ISSUE #6361: Oracle: Bug `NUMBER` w/o precision and scale is converted to Decimal with zero scale
## Describe the bug

According to documentation it should be treated as decimal with high scale
https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/Data-Types.html#GUID-9401BC04-81C4-4CD5-99E7-C5E25C83F608

>The absence of precision and scale designators specifies the maximum range and precision for an Oracle number.

sgrebnov:
## 📝 Summary

PR improves Oracle types conversion:

- Fixes `TIMESTAMP WITH LOCAL TIME ZONE` type. Although we use Timestamp with time zone (UTC) to represent this type, we can't rely on the `chrono::DateTime<FixedOffset>` logic, as offset information is not applicable or valid in this case. Since we enforce UTC time zone in the connection session, the conversion happens on the server side and the received value is already in UTC.
- Fixes `NUMBER` without scale and precision defined. The absence of precision and scale designators specifies the maximum range and precision for an Oracle number (as per documentation), so we default to `Decimal128(38, 20)`, similar to [Postgres](https://github.com/datafusion-contrib/datafusion-table-providers/blob/b85b68544241c6811415fc0b78f04a0cc67e81d6/core/src/sql/arrow_sql_gen/postgres/schema.rs#L199).
- Use Arrow `Int64` for `NUMBER` with zero scale. Oracle does not have a dedicated type for integer values, and integers are converted to `NUMBER` with zero scale when added to the database. For performance and memory considerations, it is much more efficient to use the Arrow integer type instead of Arrow `Decimal128` for such cases.

PR adds Integration tests for Oracle type conversion and predicates pushdown based on `gvenzl/oracle-free:latest` - more lightweight and developer friendly version of oracle docker image ([the same is used by rust-oracle create for testing](https://github.com/kubo/rust-oracle/commit/caf3c0d06b654e79192e88bbaf516e599982af05))

Example test run: https://github.com/spiceai/spiceai/actions/runs/16035471009/job/45246959649?pr=6327

>PASS [  46.584s] runtime::integration oracle::oracle_integration_test

## 🔗 Related

Fixes https://github.com/spiceai/spiceai/issues/6361


crates/runtime/tests/docker/mod.rs
```diff
@@ -145,7 +146,10 @@ pub struct ContainerRunner<'a> {
 }
 
 impl<'a> ContainerRunner<'a> {
-    pub async fn run(self) -> Result<RunningContainer<'a>, anyhow::Error> {
+    pub async fn run(
+        self,
+        timeout: Option<Duration>,
```
sgrebnov: Oracle container initialization time is ~50s, added  optional timeout param to extend initialization time for Oracle to 2m to ensure tests are robust

.github/workflows/integration.yml
```diff
@@ -126,6 +126,15 @@ jobs:
         with:
           os: 'linux'
 
+      - name: Install Oracle ODPI-C
```
sgrebnov, line 129: This is required for Oracle Connector to work. To be part of Oracle documentation

Installation takes ~5 seconds, simplified version of 
https://github.com/kubo/rust-oracle/blob/master/.github/workflows/run-tests.yml#L49

crates/runtime/tests/docker/mod.rs
```diff
@@ -171,7 +175,7 @@ impl<'a> ContainerRunner<'a> {
                 format!("{container_port}/tcp"),
                 Some(vec![PortBinding {
                     host_ip: Some("127.0.0.1".to_string()),
-                    host_port: Some(format!("{host_port}/tcp")),
+                    host_port: Some(format!("{host_port}")),
```
sgrebnov, line 178: Oracle docker container was not exposing port when using `/tcp` suffix (tested in CI and locally - same issue), per documentation protocol is required for container_port only, also tested all other docker containers binds ports correctly with this change

crates/runtime/tests/dynamodb/mod.rs
```diff
@@ -52,18 +52,6 @@ async fn dynamodb_federated() -> Result<(), anyhow::Error> {
                 () = cloned_rt.load_components() => {}
             }
 
-            let mut query_result = rt
```
sgrebnov, line 55: Unrelated: this logic was perviously moved into `run_and_snapshot_query` but the code has not been removed.

crates/data_components/src/oracle/convert.rs
```diff
@@ -242,7 +243,9 @@ pub(crate) fn rows_to_arrow(rows: &[Row], schema: &SchemaRef) -> super::Result<R
                         }
                     );
                 }
-                DataType::Timestamp(TimeUnit::Second, None) => {
+                // If TIMESTAMP WITH LOCAL TIME ZONE or TIMESTAMP WITHOUT TIME ZONE (seconds precision)
+                (DataType::Timestamp(TimeUnit::Second, _), OracleType::TimestampLTZ(_))
```
sgrebnov, line 267: Integration tests helped to identify invalid conversion of TIMESTAMP WITH LOCAL TIME ZONE type. Fixed

Despite we use Timestamp with time zone to represent this time, we can't rely on the `chrono::DateTime<FixedOffset>` logic as offset information is not applicable/valid in this case.

crates/runtime/tests/oracle/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+                &rt,
+                "describe test_tbl",
+                "schema",
+            )
+            .await?;
+
+            run_and_snapshot_query(
+                &rt,
+                r#"select 
+                    round("ID") as ID,
```
sgrebnov, line 229: We round ID as it has very high scale and fractional part might fluctuate, for example I see it returned as `1.0000000000006389`. Im going to double check to see if there could be something how we initialize test number as well

crates/runtime/tests/oracle/snapshots/integration__oracle__filters_pushdown_query_plan.snap
```diff
@@ -0,0 +1,15 @@
...
+| physical_plan | GlobalLimitExec: skip=0, fetch=1                                                                                                                                                                                                                                                                                                                                               |
+|               |   BytesProcessedExec                                                                                                                                                                                                                                                                                                                                                           |
+|               |     OracleExec sql=SELECT "VAL_NUMBER", "VAL_DECIMAL", "VAL_FLOAT" FROM "TEST_TABLE" WHERE ("VAL_NUMBER" = 123.45) AND (CAST("VAL_FLOAT" AS FLOAT) > 2.8) AND ("VAL_INTEGER" >= 123456789012345678) AND ("VAL_DECIMAL" < 10000.0000) AND ("VAL_VARCHAR2" = 'ghi') AND ("VAL_BOOL" = 'Y')                                                                                       |
```
sgrebnov, line 13: This confirms predicates are pushed down

crates/runtime/tests/docker/mod.rs
```diff
@@ -200,6 +211,7 @@ impl<'a> ContainerRunner<'a> {
             env: Some(env_vars_str),
             host_config,
             healthcheck: self.healthcheck,
+            exposed_ports,
```
sgrebnov, line 214: W/o `exposed_ports` configuration matching ports bindings  I was unable to get port bindings to work for Oracle container - not sure why it is special (. This fixed connection / ports binding issue

crates/data_components/src/oracle/convert.rs
```diff
@@ -68,11 +70,18 @@ pub(crate) fn map_oracle_type_to_arrow_type(
         // Oracle types below max size is 32767 bytes
         "ROWID" | "CHAR" | "NCHAR" | "VARCHAR2" | "NVARCHAR2" | "LONG" => Some(DataType::Utf8),
         "CLOB" | "NCLOB" => Some(DataType::LargeUtf8),
-        "NUMBER" | "DECIMAL" => {
-            // In Oracle, default precision and scale are (38, 0).
-            let precision = precision.unwrap_or(38);
-            let scale = scale.unwrap_or(0);
-            Some(DataType::Decimal128(precision, scale))
+        "NUMBER" => {
```
Sevenannn, line 73: Does the "DECIMAL" datatype still need to be handled? It's removed from the matching branch

sgrebnov: No, DECIMAL can't actually be here - despite you can use "DECIMAL" to define column and as SQL type it is stored as NUMERIC type. We have `VAL_DECIMAL" DECIMAL(18,4),` as part of the integration test to verify such behavior.

It is also not in the `Table 2-1 Built-In Data Type Summary`
https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-7B72E154-677A-4342-A1EA-C74C1EA928E6

crates/runtime/tests/oracle/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+            let cloned_rt = Arc::new(rt.clone());
+
+            // Set a timeout for the test
+            tokio::select! {
+                () = tokio::time::sleep(std::time::Duration::from_secs(60)) => {
+                    return Err(anyhow::Error::msg("Timed out waiting for datasets to load"));
+                }
+                () = cloned_rt.load_components() => {}
+            }
+
```
Sevenannn, line 216: A `runtime_ready_check` can be added here to ensure datasets are ready in spice before query

sgrebnov: Added - thank you

Sevenannn (approved):

Issue #5872, opened 2025-05-19
Enhancement: Initial support for `partition_by` acceleration expressions (DuckDB)


phillipleblanc:
## Goal-State/What/Result

Spice can be configured to partition acceleration data using `partition_by` expressions defined in the dataset acceleration settings. Data inserted into the acceleration by Spice will automatically partition the data by the given expressions. Queries on the accelerated data will prune out partitions based on the query predicates.

Multiple `partition_by` expressions can be added to the spicepod, but initially, only one will be supported.

Two new UDFs are registered in Spice that are useful in `partition_by` expressions:
  - `bucket(num_buckets, col)` - Partitions a column into a fixed number of buckets based on a hash of the column value.
  - `truncate(width, col)` - Truncates a column to a given width specified by `col - (col % width)`. i.e. `truncate(10, 101) = 100`, `truncate(10, 100) = 100`, `truncate(10, 99) = 90`. This works the same way as the Iceberg `truncate` partition transform.

Partition expression examples:

- `partition_by: region` - Uses the `region` column directly from the table to partition.
- `partition_by: a > 5` - Partition into two partitions: `a > 5` and `a <= 5`.
- `partition_by: bucket(10, a)` - Partition into 10 partitions based on the hash of the value of `a`.
- `partition_by: a % 10` - Partition into 10 partitions based on the remainder of the value of `a` divided by 10.
- `partition_by: CASE WHEN a ~* '^DATAFUSION(-cli)*' THEN 'datafusion' ELSE 'other' END` - Partition into two partitions: `datafusion` and `other`.
- `partition_by: date_trunc('month', date)` - Partition the `date` column by year and month.
- `partition_by: truncate(1000, sales_volume)` - Partition sales volume into 1000-unit buckets. (i.e. `1234 -> 1000`, `2235 -> 2000`, etc.)

Spicepod example:

```yaml
datasets:
  - from: s3://my_bucket/some_large_table/
    name: my_table
    params:
      file_format: parquet
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
      partition_by: 
        - bucket(100, account_id)
```

## Why/Purpose

Required for Spice to be able to accelerate large quantities of data that it currently can't handle. It also allows for faster query execution when the predicate of a query can filter out many partitions.

## By When

**Issue/Spec written and reviewed:** 18th June 2025
**Done-Done:** 27th June 2025

## Done-Done

[...]

PR #6338, opened 2025-06-29
Add a PartitionedDuckDB Accelerator

### LINKED ISSUE #6185: Implement PartitionCreator

### LINKED ISSUE #6186: Implement PartitionedTableProvider::scan

### LINKED ISSUE #6187: Implement PartitionedTableInsert

kczimm:
## 📝 Summary

- Introduce a `PartitionedDuckDBAccelerator` that is used when DuckDB file mode and partition_by expressions are specified.
- Implemented partitioning in `PartitionTableProvider` with `insert_into` for data partitioning.
- Created `DuckDBPartitionCreator` for creating and inferring DuckDB partitions.
- Refactored `create_table_provider` in duckdb for reusable table provider creation.
- Introduced `EngineVariant` to support partitioned and non-partitioned accelerator engines.
- Updated `AcceleratorEngineRegistry` to use `EngineVariant` for engine registration.
- Implemented PartitionInsertExec execution plan for streaming, non-blocking partitioned data insertion.
- Added a generic ConnectionPool to the Partition, PartitionTableProvider, and PartitionCreator so that pools can be retrieved for checkpointing purposes.
- Added checkpointing for PartitionedDuckDb which checkpoints each partition eliminating the need for a full refresh.
Implement TableProvider::scan for PartitionTableProvider with filter push down.

## 🔗 Related

Part of:
- #5872 

Closes:
- #6185 
- #6368 
- #6187


kczimm:
Changing to draft because I found an issue with certain expressions.

crates/runtime-table-partition/src/provider.rs
```diff
@@ -84,10 +100,12 @@ impl PartitionTableProvider {
             .collect::<Result<HashMap<_, _>, _>>()
             .context(ValidatingExpressionsSnafu)?;
 
+        let partitions = RwLock::new(partitions);
```
sgrebnov: Do we plan to add special logic to determine that `partition_by` has changed and existing partitions are no longer valid?

crates/runtime-table-partition/src/provider.rs
```diff
@@ -118,6 +136,136 @@ impl TableProvider for PartitionTableProvider {
...
+        if execution_plans.is_empty() {
+            let mem_exec = DataSourceExec::new(Arc::new(MemorySourceConfig::try_new(
+                &[vec![]],
+                Arc::clone(&self.schema),
+                None,
+            )?));
+            Ok(Arc::new(mem_exec))
+        } else if execution_plans.len() == 1 {
+            #[allow(clippy::unwrap_used)]
+            Ok(execution_plans.into_iter().next().unwrap())
```
sgrebnov: Should we replace unwrap here with error?

crates/runtime-table-partition/src/provider.rs
```diff
@@ -118,6 +136,136 @@ impl TableProvider for PartitionTableProvider {
...
+                let mem_exec = DataSourceExec::new(Arc::new(MemorySourceConfig::try_new(
+                    &[vec![batch]],
+                    Arc::clone(&self.schema),
+                    None,
+                )?));
+
+                let plan = table_provider
+                    .insert_into(state, Arc::new(mem_exec), insert_op)
+                    .await?;
+                execution_plans.push(plan);
```
sgrebnov: @kczimm - do we have any special final commit/rollback logic for partitioned insert - let's say one table provider failed to insert but other completed, how we control this?

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,308 @@
...
+        "partitioned_duckdb"
+    }
+
+    fn valid_file_extensions(&self) -> Vec<&'static str> {
+        self.base_accelerator.valid_file_extensions()
+    }
+
+    fn file_path(&self, _source: &dyn AccelerationSource) -> Result<String, DataAcceleratorError> {
+        Ok(String::new())
+    }
```
peasee: Why is this an empty string?

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    async fn init(
```
peasee, line 223: This should probably replicate the existing DuckDB accelerator init

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,308 @@
...
+        if let Some(acceleration_settings) = source.acceleration() {
+            ensure!(
+                matches!(acceleration_settings.mode, Mode::File),
+                FileModeOnlySnafu
+            );
+        }
+        Ok(())
+    }
+
+    async fn create_external_table(
```
peasee, line 236: Will partitioned DuckDB accelerators support cross-database federation? This is missing the similar setup from the existing DuckDB accelerator.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+impl Default for PartitionedDuckDBAccelerator {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[async_trait]
+impl DataAccelerator for PartitionedDuckDBAccelerator {
```
peasee, line 198: This is missing an `is_initialized` implementation.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,308 @@
...
+    async fn create_partition(
+        &self,
+        partition_value: ScalarValue,
+    ) -> Result<Arc<dyn TableProvider>, creator::Error> {
+        let mut cmd = self.cmd.clone();
+        let partition_value_str = encode_scalar_value(&partition_value)
+            .map_err(|e| creator::Error::CreatePartition { source: e.into() })?;
+
+        let table = cmd.name.table();
+        let data_path = spice_data_base_path();
```
peasee: What happens if someone specifies a `duckdb_file` parameter that is not in this folder location? The partitioned accelerator just forces the files into this folder?

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,308 @@
...
+            return Ok(vec![]);
+        }
+
+        let mut dir_entries = tokio::fs::read_dir(&data_path)
+            .await
+            .map_err(|e| creator::Error::InferringPartitions { source: e.into() })?;
+
+        let mut partitions = Vec::new();
+
+        let valid_extensions = ["db"];
```
peasee: should this be a method on the `DuckDBPartitionCreator`, like `DuckDBPartitionCreator::valid_extensions`?

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,308 @@
...
+        Ok(table_provider)
+    }
+
+    async fn infer_existing_partitions(&self) -> Result<Vec<Partition>, creator::Error> {
+        let data_path = spice_data_base_path();
+        let table = self.cmd.name.table();
+        let data_path = Path::new(&data_path).join(table);
+
+        if !data_path.is_dir() {
+            create_dir(data_path)
```
peasee: `create_dir_all` should probably be used, in case there are missing parts of the tree

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,308 @@
...
+                    .map_err(|e| creator::Error::InferringPartitions { source: e })?;
+
+                partitions.push(Partition {
+                    partition_value,
+                    table_provider,
+                });
+            }
+        }
+
+        tracing::info!(
```
peasee: should this be a trace?

crates/runtime-table-partition/src/provider.rs
```diff
@@ -118,6 +136,136 @@ impl TableProvider for PartitionTableProvider {
...
+            Ok(execution_plans.into_iter().next().unwrap())
+        } else {
+            Ok(Arc::new(UnionExec::new(execution_plans)))
+        }
     }
 }
+
+fn group_by_partition(
+    expr: &Expr,
+    batches: &[RecordBatch],
```
Sevenannn: nit: This function can take `batches: &RecordBatch` as input, the caller of the function seems to be only wrapping single batch in a list `let partition_groups = group_by_partition(expr, &[batch], &df_schema)?;`

crates/runtime/src/dataaccelerator/duckdb.rs
```diff
@@ -410,6 +379,34 @@ impl DataAccelerator for DuckDBAccelerator {
...
+) -> Result<Arc<dyn TableProvider>, Box<dyn std::error::Error + Send + Sync>> {
+    let ctx = SessionContext::new();
+    let table_provider = duckdb_factory
+        .create(&ctx.state(), cmd)
+        .await
+        .context(UnableToCreateTableSnafu)
+        .boxed()?;
+
+    let Some(duckdb_writer) = table_provider.as_any().downcast_ref::<DuckDBTableWriter>() else {
+        unreachable!("DuckDBTableWriter should be returned from DuckDBTableProviderFactory")
```
Sevenannn, line 417: Instead of panicking this could use an internal error and ask user to report bug (if the issue is truly unexpected under normal user usage)

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,308 @@
...
+            }
+        }
+
+        let mut table_provider_guard = self.table_provider.lock().await;
+        ensure!(table_provider_guard.is_none(), SingleTableSnafu);
+
+        let schema = Arc::new(cmd.schema.as_arrow().clone());
+        let creator = Arc::new(DuckDBPartitionCreator::new(cmd));
+        let table_provider =
+            Arc::new(PartitionTableProvider::new(creator, partition_by, schema).await?);
```
Sevenannn, line 252: Would the partitioned DuckDB instances be attached to each other and allowing for federated pushdown?

crates/runtime-table-partition/src/provider.rs
phillipleblanc: That is out of scope for this enhancement, its a similar problem to detecting if the user has changed the refresh_sql for an existing acceleration. Its possible for them to do that, but we don't do anything special to handle it.

crates/runtime-table-partition/src/provider.rs
phillipleblanc: While I don't think we can ever completely eliminate that risk in the current design, we can minimize its impact by only calling `insert_into` once on each of the partitions and keeping that stream alive until the end, and then closing them all at once. The current implementation will commit each batch, which is dangerous.

phillipleblanc: So then the logic would need to keep a map of partition values -> insert streams and just sending data to the correct stream as it determines it, and then close all of the streams at once.

crates/runtime-table-partition/src/provider.rs
```diff
@@ -118,6 +136,136 @@ impl TableProvider for PartitionTableProvider {
...
+            "PartitionedTableProvider::scan not implemented yet".to_string(),
+        ))
+    }
+
+    async fn insert_into(
+        &self,
+        state: &dyn Session,
+        input: Arc<dyn ExecutionPlan>,
+        insert_op: InsertOp,
+    ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
```
phillipleblanc, line 144: The insert_into method itself shouldn't do the computation - it should return an ExecutionPlan which represents the computation graph. If you look at the current implementations, you'll see that they just return a `DataSinkExec` which actually handles the insertion.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
kczimm: There is no one file path but one for each partition. This function is only internally used (within this trait) in the DuckDBAccelerator, for example, but is never used in this implementation.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
kczimm: The existing init creates the DuckDB file and creates the tables. In this implementation, the files cannot be created until the partition values are known when data are inserted.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
kczimm: Added

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
kczimm: Currently, the implementation ignores this parameter because there will likely be many files. It would be better to specify a directory (I imagine) but that's not implemented in this PR. I'll discuss the desired UX with Luke.

crates/runtime-table-partition/src/provider.rs
kczimm: Function does not exist in newest version.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
kczimm: This implementation has separate pools for each partition but I think that having them attached with a single connection would be an improvement.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
kczimm: this implementation uses a separate db connection pool for each partition but I believe an improvement would be to have a single connection pool with each partition file attached.

crates/runtime-table-partition/src/creator.rs
```diff
@@ -36,15 +36,23 @@ pub enum Error {
 
 #[async_trait]
 pub trait PartitionCreator: Debug + Send + Sync {
+    /// Connection pool to the partitioned data source
+    type ConnectionPool;
```
phillipleblanc: As mentioned on the other PR, I don't see why we need this

crates/runtime-table-partition/src/insert.rs
```diff
@@ -14,62 +14,305 @@ See the License for the specific language governing permissions and
...
+        let creator = Arc::clone(&self.creator);
+        let partition_providers = Arc::clone(&self.partitions);
+        let input = Arc::clone(&self.input);
+        let task_ctx = Arc::clone(&context);
+
+        let mut partition_senders: HashMap<String, Sender<RecordBatch>> = HashMap::new();
+
+        let mut stream = execute_stream(input, task_ctx)?;
+
+        block_in_place(move || {
```
phillipleblanc: As mentioned in the other PR, this shouldn't block the thread calling execute, any async work that needs to be done should be done in the returned stream (which is async and expected to do work)

phillipleblanc: You can use the `RecordBatchReceiverStream` to spawn an async task and return the batches that are returned from that, here is an example implementation: https://github.com/spiceai/spiceai/blob/trunk/crates/data_components/src/dynamodb/provider.rs#L331

Docs: https://docs.rs/datafusion/latest/datafusion/physical_plan/stream/struct.RecordBatchReceiverStreamBuilder.html

crates/runtime-table-partition/src/insert.rs
```diff
@@ -14,62 +14,305 @@ See the License for the specific language governing permissions and
...
 
-impl DisplayAs for PartitionedTableInsert {
-    fn fmt_as(&self, _t: DisplayFormatType, _f: &mut fmt::Formatter) -> fmt::Result {
-        // TODO
-        Ok(())
+impl LazyBatchGenerator for BatchGenerator {
+    fn generate_next_batch(&mut self) -> Result<Option<RecordBatch>, DataFusionError> {
+        block_in_place(|| Handle::current().block_on(async { Ok(self.rx.recv().await) }))
     }
 }
```
phillipleblanc:
```suggestion
```

I'm not sure why this is needed?

crates/runtime-table-partition/src/provider.rs
```diff
@@ -40,61 +46,85 @@ pub enum Error {
...
+            partition_by,
+            partitions,
             schema,
         })
     }
+
+    /// Get `ConnectionPool`s for each partition.
+    ///
+    /// # Errors
+    pub async fn get_shared_pools(&self) -> Vec<Arc<ConnectionPool>> {
```
phillipleblanc: I'm not a fan of having this here. I'm fine if we need to store extensions on the partitions and partition table provider to enable this scenario, but connection pools are a very specific implementation detail that won't generalize.

Here is an example of what I mean by extensions: https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/request/context.rs#L171

crates/runtime-table-partition/src/provider.rs
```diff
@@ -111,13 +141,148 @@ impl TableProvider for PartitionTableProvider {
...
+            Arc::clone(&self.schema),
+        )))
+    }
+}
+
+/// Determine whether a partition should be pruned from the scan plan based on
+/// the query `filters`, the expression that the partition was created from,
+/// `partition_by`, and the `partition_value` produced by the `partition_by`
+/// `Expr` for this particular partition.
+fn prune_partition(filters: &[Expr], partition_by: &Expr, partition_value: &ScalarValue) -> bool {
```
phillipleblanc: Doesn't need to be in this PR, but I'm guessing as this function gets more generalized that it will need to be in a separate module with its own set of tests.

crates/runtime/src/dataaccelerator/duckdb.rs
phillipleblanc: Yeah this is definitely something that could happen via bugs :)

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,377 @@
...
+        Self {
+            base_accelerator: DuckDBAccelerator::new(),
+            table_provider: Mutex::new(None),
+            is_initialized: AtomicBool::new(false),
+        }
+    }
+
+    pub(crate) async fn get_shared_pools(&self) -> Vec<Arc<DuckDbConnectionPool>> {
+        if let Some(provider) = self.table_provider.lock().await.as_ref() {
+            provider.get_shared_pools().await
```
phillipleblanc: This is where the caller would get the extensions it expects (in this case a connection pool)

crates/runtime-table-partition/src/creator.rs
kczimm: Checkpointing needs a connection pool via `runtime::dataaccelerator::spice_sys::dataset_checkpoint`. The connection pool will come from the PartitionCreator, presumably.

crates/runtime-table-partition/src/provider.rs
kczimm: Again, the generic `ConnectionPool` was to accomplish checkpointing using the established checkpointing patterns in `runtime::dataaccelerator::spice_sys::dataset_checkpoint`. `AccelerationConnection` is an enum with variants, which are obviously known statically. `PartitionCreator` is a trait object for generic code. We needed a way to get an `AccelerationConnection` from the `PartitionCreator`. I'll see how I can remove this `ConnectionPool` and still get checkpointing to work.

crates/runtime-table-partition/src/insert.rs
kczimm: An `impl LazyBatchGenerator` type is required to use the `LazyMemExec` plan which I thought was appropriate for this case. We need a way to stream the newly created RecordBatches via an `ExecutionPlan` into each partition's `insert_into`. If there is a better way, I'm happy to learn about it.

crates/runtime-table-partition/src/provider.rs
```diff
@@ -111,13 +141,148 @@ impl TableProvider for PartitionTableProvider {
...
+    for filter in filters {
```
phillipleblanc: Is there a card I missed that is tracking expanding this support to cover all expressions types?

phillipleblanc: Also this is assuming that the partition by expression is a literal column - this should actually compute the expression in terms of the query filter.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
peasee: You don't need a single connection pool to attach all of the databases. The [existing database attachment method](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/dataaccelerator/duckdb.rs#L322-L323) for example, allows cross-database federation using distinct connection pools. It may become more complex with partitioned insertion though, but it should be no different than the existing accelerators adding more data/acceleration refresh.

I think this will be required for larger dataset sizes (SF1000, 3000, etc) as loading all of the partitions into memory for the final scan will likely be a bottleneck.

Will partitioned datasets support cross-dataset federation with non-partitioned datasets? E.g. a federated join from non-partitioned A with partitioned B.

kczimm: Thanks for elaborating. I need to think more carefully about this. I shied away from this due to DuckDB being unable to write to multiple attached databases within a single transaction, https://duckdb.org/docs/stable/sql/statements/attach.html#transactional-semantics. However, I understand what you're saying how this may be practically required for the large datasets.

crates/runtime-table-partition/src/insert.rs
kczimm: The work is now done in the stream as it should be. The trouble with `RecordBatchReceiverStream` for this is that we need to stream RecordBatches into an ExecutionPlan that gets passed into the DuckDB TableProvider::insert_into. That is what the LazyMemExec was for but unfortunately, it had a sync interface. I've opted for a new custom ExecutionPlan that will send a stream of RecordBatches into the DuckDB TableProvider insert_into without any invocations of `block_in_place`.

crates/runtime-table-partition/src/insert.rs
kczimm: Like mentioned in a different comment, this approach was taken because we cannot pass a stream of RecordBatches to DuckDB TableProvider::insert_into, but need an `ExecutionPlan`. I've created a custom one that avoids the use of `block_in_place` and the need to `impl LazyBatchGenerator`.

crates/runtime-table-partition/src/provider.rs
kczimm: I've removed all instances of the associated type or generic parameter `ConnectionPool`. Instead, the `PartitionedDuckDBAccelerator` provides the pool directly for a single checkpointing file/table that is used once for all partitions.

crates/runtime-table-partition/src/provider.rs
kczimm: I've added an issue to address making this more generic #6406.

crates/runtime-table-partition/src/provider.rs
kczimm: I've added an issue to address making this more generic https://github.com/spiceai/spiceai/issues/6406.

crates/runtime-table-partition/src/creator.rs
kczimm: I've removed all instances of this associated type and address checkpointing from `PartitionedDuckDBAccelerator` directly.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
kczimm: I've added an issue specifically for addressing/supporting this: #6408

crates/runtime/src/dataaccelerator/mod.rs
```diff
@@ -79,9 +83,15 @@ pub enum Error {
 
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Hash)]
+pub struct EngineVariant {
```
phillipleblanc: Instead of creating this EngineVariant, why not just extend the existing `Engine` enum to have a PartitionedDuckDB variant?

```rust
pub enum Engine {
    #[default]
    Arrow,
    DuckDB,
    PartitionedDuckDB,
    Sqlite,
    PostgreSQL,
    Void,
}
```

phillipleblanc: I think ultimately this needs to be a trait instead of an enum, otherwise it won't be extensible - but that is a conversation for another time.

crates/runtime/src/dataaccelerator/mod.rs
```diff
@@ -92,9 +102,17 @@ impl AcceleratorEngineRegistry {
         }
     }
 
-    pub async fn get_accelerator_engine(&self, engine: Engine) -> Option<Arc<dyn DataAccelerator>> {
+    pub async fn get_accelerator_engine(
+        &self,
```
phillipleblanc: Instead of changing this signature, you could add an associated function on `Engine` that returns an `Engine` from the settings, i.e.:

```rust
impl Engine {
    fn from_acceleration_settings(acceleration_settings: &acceleration::Acceleration) -> Self {...}
}

Comment (phillipleblanc) [reply crates/runtime/src/dataaccelerator/mod.rs]: or perhaps add this to the component::dataset when its created from the spicepod spec.

File: crates/runtime/src/dataaccelerator/partitioned_duckdb.rs [hunk 25]
```diff
@@ -0,0 +1,386 @@
...
+            duckdb_factory: create_factory(),
+        }
+    }
+
+    /// Returns an existing `DuckDB` connection pool for the given dataset, or creates a new one if it doesn't exist.
+    pub async fn get_shared_pool(
+        &self,
+        source: &dyn AccelerationSource,
+    ) -> Result<Arc<DuckDbConnectionPool>> {
+        let data_path = spice_data_base_path();
```
phillipleblanc: Do we support the option to change where they place the partitioned files? I would expect so, and then this checkpoint DB should be placed next to the data partitions.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,386 @@
...
+
+#[derive(Debug)]
+pub(crate) struct DuckDBPartitionCreator {
+    cmd: CreateExternalTable,
+    duckdb_factory: DuckDBTableProviderFactory,
+    partition_dir: PathBuf,
+}
+
+impl DuckDBPartitionCreator {
+    pub(crate) fn new(cmd: CreateExternalTable) -> Self {
```
phillipleblanc: There is a location field in `CreateExternalTable` that we could use for overriding the default location. I think the regular DuckDB accelerator uses an option called `duckdb_file`, but `location` is a bit more idiomatic.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,386 @@
...
+
+        let partition = Partition {
+            partition_value,
+            table_provider,
+        };
+
+        Ok(partition)
+    }
+
+    async fn infer_existing_partitions(&self) -> Result<Vec<Partition>, creator::Error> {
```
phillipleblanc, line 319: As discussed, this could potentially come from the checkpoint/metadata DB itself.

phillipleblanc: But that can come in a separate issue/PR

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -0,0 +1,386 @@
...
+    partition_dir: &Path,
+    cmd: &mut CreateExternalTable,
+    partition_value: &ScalarValue,
+) -> Result<String, filename::Error> {
+    let partition_value_str = encode_scalar_value(partition_value)?;
+
+    let duckdb_file = format!("{partition_value_str}.db");
+    let duckdb_path = partition_dir.join(&duckdb_file);
+    let duckdb_path = duckdb_path.display().to_string();
+    cmd.options.insert("open".to_string(), duckdb_path.clone());
```
phillipleblanc, line 412: Ah yeah, this is where we pass the location to the underlying DuckDB instance - it would be more idiomatic if this actually used `cmd.location` parameter instead, but that probably would require a table-providers update, so out of scope for this change.

phillipleblanc: We could at least support `cmd.location` for the partitioned table provider, even if we need to specify `cmd.options["open"]` for each individual DuckDB partition

crates/runtime-table-partition/src/provider.rs
```diff
@@ -111,13 +128,63 @@ impl TableProvider for PartitionTableProvider {
         TableType::Base
     }
 
+    fn supports_filters_pushdown(
+        &self,
+        filters: &[&Expr],
+    ) -> Result<Vec<TableProviderFilterPushDown>, DataFusionError> {
+        Ok(vec![TableProviderFilterPushDown::Exact; filters.len()])
```
phillipleblanc: I think this might be a dangerous default - we're basically claiming that any filter DataFusion gives us we'll be able to handle. This is probably true for DuckDB, but not all possible future implementations. We may need to add a `supports_filters_pushdown` method on the partition creator - since it will have some more information on what it can support or not.

crates/runtime-table-partition/src/insert.rs
```diff
@@ -14,62 +14,498 @@ See the License for the specific language governing permissions and
...
+    }
+}
+
+impl DisplayAs for PartitionerExec {
+    fn fmt_as(
+        &self,
+        _t: datafusion::physical_plan::DisplayFormatType,
+        f: &mut std::fmt::Formatter,
+    ) -> std::fmt::Result {
+        write!(f, "{}", self.name())
```
phillipleblanc: This Display will show up in Explain plans when you do `EXPLAIN ...` - I think we can add some more useful info here that would help us debug issues, i.e. the partition expression

crates/runtime-table-partition/src/insert.rs
```diff
@@ -14,62 +14,498 @@ See the License for the specific language governing permissions and
...
+
+struct PartitionInsertExec {
+    rx: Mutex<Option<Receiver<RecordBatch>>>,
+    schema: SchemaRef,
     properties: PlanProperties,
-    projected_schema: SchemaRef,
 }
 
-impl ExecutionPlan for PartitionedTableInsert {
+impl PartitionInsertExec {
```
phillipleblanc: Consider renaming this to `PartitionInputExec` - it took me a few minutes to realize this isn't actually doing the insert for a partition, it exists to adapt a tokio::sync::mcsp Receiver channel into the input for a partition's `insert_into`.

crates/runtime-table-partition/src/insert.rs
```diff
@@ -14,62 +14,498 @@ See the License for the specific language governing permissions and
...
+                    for (partition_key, (partition_value, batch)) in batches {
+                        let tx = if let Some(tx) = partition_senders.get(&partition_key) {
+                            tx.clone()
+                        } else {
+                            // spawn the insertion task for this partition
+                            let (tx, rx) = channel(10);
+                            partition_senders.insert(partition_key.clone(), tx.clone());
+
+                            let partition = creator
+                                .create_partition(partition_value)
```
phillipleblanc: Shouldn't we check the partition_providers list first to see if we already have the partition?

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
kczimm: If I follow how we are doing things in the DuckDB accelerator, I need the partition directory to be derivable from a `&dyn AccelerationSource`. I've added a function that does this.

crates/runtime/src/dataaccelerator/partitioned_duckdb.rs
```diff
@@ -358,6 +361,13 @@ impl PartitionCreator for DuckDBPartitionCreator {
         );
         Ok(partitions)
     }
+
+    fn supports_filters_pushdown(
+        &self,
+        filters: &[&Expr],
+    ) -> Result<Vec<TableProviderFilterPushDown>, DataFusionError> {
+        Ok(vec![TableProviderFilterPushDown::Exact; filters.len()])
```
phillipleblanc: We should first run the `Expr` through `expr_to_sql` to make sure we can properly unparse it: https://docs.rs/datafusion/latest/datafusion/sql/unparser/fn.expr_to_sql.html

If we can, then its exact. If we can't then its unsupported.

phillipleblanc (dismissed):
We should have some integration/E2E tests for this - but if you want to track that in another issue, then I'm fine to merge this after fixing the `supports_filter_pushdown` method to call expr_to_sql

kczimm:
> We should have some integration/E2E tests for this - but if you want to track that in another issue, then I'm fine to merge this after fixing the `supports_filter_pushdown` method to call expr_to_sql

#6386

phillipleblanc (approved):

PR #6362, opened 2025-07-02
Fix projection pushdown issues for document based file connector

### LINKED ISSUE #4882: Incorrect projection pushdown for document based File Connector
## Describe the bug
When using file connector locally, when a user asks for a projection with a single column, it will incorrectly only return the `location` column (i.e. the first column) regardless of what is specified.

## To Reproduce
```shell
sql> select location from docs;
+------------------------------------------------------------------------------------+
| location                                                                           |
+------------------------------------------------------------------------------------+
| Users/jeadie/Github/spiceai/docs/decisions/001-use-snmalloc-as-global-allocator.md |
| Users/jeadie/Github/spiceai/docs/decisions/003-duration-ms.md                      |
| Users/jeadie/Github/spiceai/docs/decisions/002-default-ports.md                    |
+------------------------------------------------------------------------------------+

sql> select content from docs; # returning location, not content
+------------------------------------------------------------------------------------+
| content                                                                            |
+------------------------------------------------------------------------------------+
| Users/jeadie/Github/spiceai/docs/decisions/001-use-snmalloc-as-global-allocator.md |
| Users/jeadie/Github/spiceai/docs/decisions/003-duration-ms.md                      |
| Users/jeadie/Github/spiceai/docs/decisions/002-default-ports.md                    |
+------------------------------------------------------------------------------------+

sql> select length(content), location from docs; # When both are in projection, content is returned (and used) correctly.
+--------------------------------+------------------------------------------------------------------------------------+
| character_length(docs.content) | location                                                                           |
+--------------------------------+------------------------------------------------------------------------------------+
| 3025                           | Users/jeadie/Github/spiceai/docs/decisions/003-duration-ms.md                      |
| 2443                           | Users/jeadie/Github/spiceai/docs/decisions/001-use-snmalloc-as-global-allocator.md |
| 3579                           | Users/jeadie/Github/spiceai/docs/decisions/002-default-ports.md                    |
+--------------------------------+------------------------------------------------------------------------------------+
```

## Runtime Details
Happens on latest build and prior releases.

### Spicepod
Can use `docs/` from this repo to reproduce.

```yaml
datasets:
  - from: file:docs/decisions/
    name: docs
    params:
      file_format: md
```

## Additional context
 - Does not occur for structured files (e.g. CSVs). 
 - Code related to document object store tables is [here](https://github.com/spiceai/spiceai/blob/trunk/crates/data_components/src/object/text.rs).

Advayp:
## 📝 Summary

- Fixes projection pushdown issues in the document-based file connector
- For local file connectors, when a user asks for a projection with a single column (e.g., `content`), the `location` column is incorrectly always returned. This PR fixes the issue by ensuring the record batch returned from a scan always matches the requested schema.

## 🔗 Related

- Closes #4882

### DIFF
--- a/crates/data_components/src/object/text.rs
+++ b/crates/data_components/src/object/text.rs
@@ -81,24 +81,19 @@ impl ObjectStoreTextTable {
         ])
     }
 
-    fn to_record_batch(
-        meta_list: &[ObjectMeta],
-        raw: &[Bytes],
-        formatter: Option<&Arc<dyn DocumentParser>>,
-    ) -> Result<RecordBatch, ArrowError> {
-        if meta_list.len() != raw.len() {
-            return Err(ArrowError::ParseError("Length mismatch".to_string()));
-        }
-
-        let schema = Self::table_schema();
-
-        let location_array: ArrayRef = Arc::new(StringArray::from(
+    fn get_location_value(meta_list: &[ObjectMeta]) -> ArrayRef {
+        Arc::new(StringArray::from(
             meta_list
                 .iter()
                 .map(|meta| meta.location.to_string())
                 .collect::<Vec<_>>(),
-        ));
+        ))
+    }
 
+    fn get_content_value(
+        raw: &[Bytes],
+        formatter: Option<&Arc<dyn DocumentParser>>,
+    ) -> Result<ArrayRef, ArrowError> {
         let utf8_strings: Result<Vec<_>, ArrowError> = raw
             .iter()
             .enumerate()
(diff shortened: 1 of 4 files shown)


Advayp:
TODO: Add unit tests

crates/data_components/src/object/text.rs
```diff
@@ -81,24 +81,20 @@ impl ObjectStoreTextTable {
...
-        formatter: Option<&Arc<dyn DocumentParser>>,
-    ) -> Result<RecordBatch, ArrowError> {
-        if meta_list.len() != raw.len() {
-            return Err(ArrowError::ParseError("Length mismatch".to_string()));
-        }
-
-        let schema = Self::table_schema();
-
-        let location_array: ArrayRef = Arc::new(StringArray::from(
+    fn get_location_value(meta_list: &[ObjectMeta]) -> ArrayRef {
```
lukekim, line 84: @Advayp let's make sure we have a snapshot test before and after, so we can see the explain plan is correct for this.

phillipleblanc: Yes, let's add a test for this.

Advayp: The output of the plan depends on the location of the test runner. So, instead of making it a snapshot, I added a plan check to ensure the projection gets pushed down to the Table Scan.

crates/data_components/src/object/text.rs
```diff
@@ -237,7 +270,12 @@ impl ExecutionPlan for ObjectStoreTextExec {
     ) -> DataFusionResult<SendableRecordBatchStream> {
         Ok(Box::pin(RecordBatchStreamAdapter::new(
             self.schema(),
-            to_sendable_stream(self.ctx.clone(), self.formatter.clone(), self.limit), // TODO get prefix from filters
+            to_sendable_stream(
+                self.ctx.clone(),
+                self.formatter.clone(),
+                self.limit,
+                self.schema(),
+            ), // TODO get prefix from filter
```
kczimm: Let's not leave the TODO in here to get lost. Replace it with a comment referencing the GitHub issue number or complete the work (probably the former to keep PRs concise).

kczimm: Unless it is no longer relevant.

kczimm (approved):

sgrebnov (approved):

PR #6368, opened 2025-07-03
Partitioning TableProvider::scan and streaming ExecutionPlan

kczimm:
## 📝 Summary

- Implemented `PartitionInsertExec` execution plan for streaming, non-blocking partitioned data insertion.
- Added a generic `ConnectionPool` to the `Partition`, `PartitionTableProvider`, and `PartitionCreator` so that pools can be retrieved for checkpointing purposes.
- Added checkpointing for PartitionedDuckDb which checkpoints each partition eliminating the need for a full refresh.
- Implement `TableProvider::scan` for `PartitionTableProvider` with filter push down.
## 🔗 Related

Closes:
- #6186 
## 👀 Notes for Reviewers

- Checkpoints are stored in each partitioned database. May be better approach to make a table at the root which is used only for checkpointing the entire dataset.
- A separate pool is created for each duckdb file and insert_into involves separate transactions to each.
- Filters are pushed down to the partitions, but partition pruning only works for column equality.

### DIFF
--- a/crates/runtime-table-partition/src/creator.rs
+++ b/crates/runtime-table-partition/src/creator.rs
@@ -14,10 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use std::{fmt::Debug, sync::Arc};
+use std::fmt::Debug;
 
 use async_trait::async_trait;
-use datafusion::{catalog::TableProvider, scalar::ScalarValue};
+use datafusion::scalar::ScalarValue;
 use snafu::prelude::*;
 
 use crate::Partition;
@@ -36,6 +36,9 @@ pub enum Error {
 
 #[async_trait]
 pub trait PartitionCreator: Debug + Send + Sync {
+    /// Connection pool to the partitioned data source
+    type ConnectionPool;
+
     /// Create a new [`Partition`] using the `partition_value`.
     ///
     /// # Errors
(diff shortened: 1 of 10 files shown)


crates/runtime-table-partition/src/insert.rs
```diff
@@ -0,0 +1,318 @@
...
+        let creator = Arc::clone(&self.creator);
+        let partition_providers = Arc::clone(&self.partitions);
+        let input = Arc::clone(&self.input);
+        let task_ctx = Arc::clone(&context);
+
+        let mut partition_senders: HashMap<String, Sender<RecordBatch>> = HashMap::new();
+
+        let mut stream = execute_stream(input, task_ctx)?;
+
+        block_in_place(move || {
```
Sevenannn, line 186: Is the block_in_place going to starve the tokio threads if multiple partition insertion tasks are happening at the same time? Since the block_in_place is going to block the current thread

kczimm: It's true it will block the thread but it won't block the executor. You raise a good point that if we have more partitions than threads in the pool we may have a problem. I'll try to test this. Thanks for pointing this out!

crates/runtime-table-partition/src/creator.rs
```diff
@@ -36,18 +36,23 @@ pub enum Error {
 
 #[async_trait]
 pub trait PartitionCreator: Debug + Send + Sync {
+    /// Connection pool to the partitioned data source
+    type ConnectionPool;
```
phillipleblanc, line 40: This is fairly specific to something that requires a connection pool - this wouldn't make sense for something like Iceberg partitions. Why do we need this?

crates/runtime-table-partition/src/insert.rs
phillipleblanc, line 186: We shouldn't do this, we should do any async work in the stream that is returned.

crates/runtime-table-partition/src/creator.rs
kczimm: We need to be able to provide a connection pool for checkpointing.

PR #6373, opened 2025-07-05
Small refactoring in duckdb data accelerator

krinart:
## 📝 Summary

Small refactoring in duckdb DataAccelerator which affects two methods:

- get_shared_pool: broken into build_file_pool and build_memory_pool
- get_num_accelerating_datasets: makes it easier to understand

Doesn't add any functionality, just makes existing code slightly more readable. 

Feel free to reject for any reason

## 🔗 Related

N/A

## 📚 Docs

N/A

## 👀 Notes for Reviewers

N/A


crates/runtime/src/dataaccelerator/duckdb.rs
```diff
@@ -131,82 +131,95 @@ impl DuckDBAccelerator {
...
-                    .context(AccelerationCreationFailedSnafu)?
-            }
-            (Err(e), Mode::File) => {
-                return Err(Error::InvalidConfiguration {
-                    detail: Arc::from(e.to_string()),
-                });
-            }
-        };
+        match acceleration.mode {
+            Mode::File => self.build_file_pool(source).await,
```
Jeadie, line 139: Instead of moving to new functions, you could strip most of the shared logic out. For example
```rust
pub async fn get_shared_pool(
        &self,
        source: &dyn AccelerationSource,
    ) -> Result<DuckDbConnectionPool> {
        let duckdb_file = self.duckdb_file_path(source);

        let acceleration = source.acceleration().context(AccelerationNotEnabledSnafu {
            dataset: source.name().to_string(),
        })?;

        let num_accelerating_datasets = self.get_num_accelerating_datasets(
            duckdb_file.as_deref().ok(),
            &source.app(),
            source.runtime(),
        );
        let max_size = Self::get_max_size(num_accelerating_datasets);

        // The only difference is now the choice of pool builder.
        let mut pool_builder = match (duckdb_file, acceleration.mode) {
            (Ok(duckdb_file), Mode::File) => DuckDbConnectionPoolBuilder::file(&duckdb_file),
            (_, Mode::Memory) => DuckDbConnectionPoolBuilder::memory(),
            (Err(e), Mode::File) => {
                return Err(Error::InvalidConfiguration {
                    detail: Arc::from(e.to_string()),
                });
            }
        };
        pool_builder = pool_builder
            .with_max_size(Some(max_size))
            .with_min_idle(Some(DEFAULT_MIN_IDLE_CONNECTIONS));
        self.duckdb_factory
            .get_or_init_instance_with_builder(pool_builder)
            .await
            .boxed()
            .context(AccelerationCreationFailedSnafu)
    }
```

phillipleblanc:
Feel free to re-open once you've incorporated Jack's feedback, thanks!

PR #6381, opened 2025-07-07
`vector_search` UDTF and related changes

Jeadie:
## 📝 Summary
- New UDTF, similar to `text_search`, but for vector embedding based columns/datasets.
```sql
SELECT my_primary_key, score, something_else
FROM vector_search(my_dataset, 'search query')
WHERE any_normal_sql_filter IS NOT NULL
ORDER BY score
DESC LIMIT 1
```

 - Important points:
   - No support for chunked columns.
   - For datasets with >1 embedding column, no aggregation is done. Must be called with additional (optional) third param `vector_search(my_dataset, 'search query', chosen_column)`. 

## 📚 Docs
- [ ] SQL reference docs
- [ ] Docs in _"Search feature"_ describing new UX.

### DIFF
--- a/crates/runtime/src/embeddings/udtf.rs
+++ b/crates/runtime/src/embeddings/udtf.rs
@@ -14,7 +14,7 @@ limitations under the License.
 //!
 //! `vector_search(tbl`: `TableReference`, query: &str, col: Option<str>, limit: Option<usize>, `include_score`: Option<bool>)
 //!
-//! - tbl: Table to perform full text search upon. If the table does not support it (i.e. no index), and empty table is returned.
+//! - tbl: Table to perform full text search upon. If the table does not support it (i.e. no index), an empty table is returned.
 //! - query: Query to perform full text search against.
 //! - col: If provided, use this column to compare vector search results against.
 //! - limit:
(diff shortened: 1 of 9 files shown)


crates/runtime/src/embeddings/udtf.rs
```diff
@@ -0,0 +1,421 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+//! A user-defined table function (UDTF) for performing vector search on a preexisting table that has an embedding configured on at least one of its columns.
+//!
+//! `vector_search(tbl`: `TableReference`, query: &str, col: Option<str>, limit: Option<usize>, `include_score`: Option<bool>)
+//!
+//! - tbl: Table to perform full text search upon. If the table does not support it (i.e. no index), and empty table is returned.
```
phillipleblanc:
```suggestion
//! - tbl: Table to perform full text search upon. If the table does not support it (i.e. no index), an empty table is returned.
```

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #6383, opened 2025-07-07
Regex Update to disallow hyphens dataset names

### LINKED ISSUE #5743: `spice configure dataset` allows specifying an invalid dataset name
## Describe the bug

When using `spice configure dataset`, the prompts allow specifying a dataset name that is invalid. For example, `taxi-trips` can be input as the dataset name and the dataset is configured "successfully" - but the Runtime will fail to start with this dataset, because a dataset name with a hyphen is not permitted.

## To Reproduce

1. Use `spice configure dataset` to configure a dataset with an invalid name
2. See Runtime startup error

## Expected behavior

`spice configure dataset` should not allow specifying an invalid dataset name.

varunguleriaCodes:
## 📝 Summary

Fixes #5743 , Updated the regex to disallow hyphens in dataset names, updated the error message accordingly, and removed the warning message.
## 🔗 Related
image


lukekim:
@varunguleriaCodes thanks for the contribution. What was the driver of this PR? Hyphens are expected be valid if the name is quoted. Are you seeing that conflict with something?

varunguleriaCodes:
Hi @lukekim ,  I haven't noticed this conflict. It  related to one of the issues mentioned - the Postgres syntax restriction that disallows hyphens in table names, as pointed here: https://github.com/spiceai/spiceai/issues/5743#issuecomment-3043227468

lukekim:
@varunguleriaCodes got it, thanks! This would be a breaking change, so our policy would be to warn on hyphen, then disallow from v2.0. Would you mind just changing this to a warning?

varunguleriaCodes:
Hi @lukekim sure, does this warning message look correct? 
" Dataset names containing hyphens (-) are deprecated and will no longer be supported starting with version 2.0."

lukekim:
@varunguleriaCodes looks reasonable, thanks!

PR #6415, opened 2025-07-09
Provide error message when partition by expression changes

### LINKED ISSUE #6414: Provide error message when partition_by expression changes

kczimm:
## 📝 Summary

- Provided error context when partition writing fails for one or more partitions
- Detect when partition_by expressions are modified and fail with a descriptive error
  - Added partition_by argument to the trait methods in `PartitionCreator` for creating and inferring partitions
  - Encoded a hash of the partition_by expressions into the partition filenames

## 🔗 Related

Closes:
- #6414

## 👀 Notes for Reviewers

There is an opporunity to store this (and other) metadata in the checkpointing database file but I thought this would suffice for now.

### DIFF
--- a/crates/runtime-table-partition/src/creator.rs
+++ b/crates/runtime-table-partition/src/creator.rs
@@ -35,6 +35,10 @@ pub enum Error {
     CreatePartition { source: StdError },
     #[snafu(display("Failed to infer accelerated partitions: {source}"))]
     InferringPartitions { source: StdError },
+    #[snafu(display(
+        "The 'partition_by' expressions are different from the expressions used to create the existing partition files. Revert the 'partition_by' expressions, delete the partition files, or change the location the partition files are stored to create new partitions."
+    ))]
+    PartitionByExpressionsChanged,
 }
 
 #[async_trait]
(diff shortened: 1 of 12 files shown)


crates/runtime-table-partition/src/creator.rs
```diff
@@ -35,21 +35,34 @@ pub enum Error {
     CreatePartition { source: StdError },
     #[snafu(display("Failed to infer accelerated partitions: {source}"))]
     InferringPartitions { source: StdError },
+    #[snafu(display(
+        "The 'partition_by' expressions are different from the expressions used to create the existing partition files. Revert the 'partition_by' expressions, delete the partition files, or change the location the partition files are stored to create new partitions."
+    ))]
+    PartitionByExpressionsChanged,
```
Jeadie, line 41: nit: I think we can make this `...Revert the 'partition_by' expressions to '{prior_expression}...', no?

Jeadie (dismissed):

crates/runtime-table-partition/src/creator/filename.rs
```diff
@@ -27,40 +27,65 @@ limitations under the License.
...
+#[derive(Serialize, Deserialize)]
+struct SerializablePair {
+    scalar: SupportedScalarValue,
+    exprs_hash: String, // Store hash of Vec<Expr> Debug string
+}
+
+/// Hashes the Debug string of a [`Vec<Expr>`] using `DefaultHasher`.
+#[must_use]
+pub fn hash_exprs(exprs: &[Expr]) -> String {
+    let exprs_str = format!("{exprs:?}");
```
phillipleblanc: I'm not sure that this is guaranteed to be stable across version upgrades. It would be unfortunate if a DataFusion upgrade broke existing partitioned tables.

crates/runtime-table-partition/src/creator.rs
kczimm: This will be great when we move to store metadata in the checkpointing database. For now, I'm just storing the hash of the partition by expressions in the filename, so the original expressions are not recoverable.

crates/runtime-table-partition/src/creator/filename.rs
kczimm: I've instead hashed the original Strings from the config file. Let me know if you think that is problematic. I was trying to avoid building out the metadata handling in the checkpointing database now.

phillipleblanc (approved):

crates/runtime-table-partition/src/creator/filename.rs
phillipleblanc: Looks good, thanks!

PR #6423, opened 2025-07-11
prune partitions for exact and in list with and without UDFs

### LINKED ISSUE #6406: Improve partition pruning

kczimm:
## 📝 Summary

- Enhanced `prune_partition` to handle UDFs in `partition_by` expressions.
- Improved pruning logic to skip filters with columns not in `partition_by`.
- Added support for `IN` and `NOT IN` list expressions in pruning.

## 🔗 Related

Closes:
- #6406


kczimm:
IN lists get rewritten into multiple nested binary operations when the list size is less than 4. I need to make sure to cover those cases, as well, so I'm reverting back to draft.

crates/runtime-table-partition/src/provider/pruning.rs
```diff
@@ -15,96 +15,513 @@ limitations under the License.
...
+            continue;
+        }
+
+        match filter {
+            Expr::BinaryExpr(BinaryExpr {
+                left,
+                op: Operator::Eq,
+                right,
+            }) => {
+                if let (Expr::Column(_), Expr::Literal(lit)) = (left.as_ref(), right.as_ref()) {
```
phillipleblanc: We should also support if the left expression has the literal and the right has the column

crates/runtime-table-partition/src/provider/pruning.rs
```diff
@@ -15,96 +15,513 @@ limitations under the License.
...
+                    } else {
+                        Err(DataFusionError::Plan(
+                            "Invalid argument: expected column or literal".into(),
+                        ))
+                    }
+                })
+                .collect::<Result<Vec<_>, _>>()?;
+            Ok(call(func.as_ref(), args)? == *partition_value)
         }
+        _ => Ok(filter_value == partition_value),
```
phillipleblanc: This shouldn't assume that the expression at this point is a column reference - it should explicitly check for that and otherwise return false

crates/runtime-table-partition/src/provider/pruning.rs
```diff
@@ -15,96 +15,513 @@ limitations under the License.
...
+    Ok(false)
+}
+
+fn filter_or_udf_value_matches(
+    column: &Expr,
+    partition_by: &Expr,
+    partition_value: &ScalarValue,
+    filter_value: &ScalarValue,
+) -> Result<bool, DataFusionError> {
+    match partition_by {
```
phillipleblanc: Wouldn't a more general approach be to walk the expression and replace the column reference with the partition value literal, and then execute the expression and see if the output matches the partition value?

i.e. something like:

```rust
fn replace_column_with_literal(partition_by_expr: Expr, literal: Expr, col: Column) -> Result<Expr> {
    partition_by_expr.transform(|e| {
        Ok(match e {
            Expr::Column(expr_col) => Transformed::yes(literal), // Also check if `expr_col` matches `col`, otherwise return an error.
            _ => Transformed::no(e),
        })
    }).map(|transformed| transformed.data)
}
```

Then you can execute the Expr directly - assuming you've properly verified it already contains the partition column in the filter, otherwise it doesn't apply. And if the output of executing that expression matches the partition_value, then you know we should include this partition.

crates/runtime-table-partition/src/provider/pruning.rs
```diff
@@ -15,96 +15,535 @@ limitations under the License.
...
+                Expr::Column(expr_col) if expr_col == *col => {
+                    Transformed::yes(Expr::Literal(filter_value.clone()))
+                }
+                _ => Transformed::no(e),
+            })
+        })
+        .map_err(|e| DataFusionError::Plan(format!("Failed to transform expression: {e}")))?
+        .data;
+
+    let result = match transformed_expr {
```
phillipleblanc, line 204: This still isn't executing the expression - this won't handle more advanced partition_by expressions like `CASE ... WHEN... THEN...`. I'll merge this for release, but I would want to see a general expression evaluator being used here.

phillipleblanc (approved):

PR #6426, opened 2025-07-12
Enable session time zone override for MySQL

sgrebnov:
## 📝 Summary

PR adds support for `mysql_time_zone` parameter that can be used to override default session time zone (UTC):

```
- from: mysql:my_table
  name: my_table
  params:
    ...
    mysql_time_zone: "system" / "local_system" / "+02:00" / "America/Los_Angeles"
```

Use `system` to refer to the MySQL server host’s OS time zone, and `local_system` to use the local runtime OS time zone.

**Note:** To use the `"America/Los_Angeles"` format, a corresponding record must exist in the `mysql.time_zone_name` table on the server.


crates/runtime/tests/mysql/mod.rs
```diff
@@ -395,3 +398,147 @@ async fn mysql_character_set_results_test() -> Result<(), String> {
...
+                    .await
+                    .map_err(RetryError::transient)
+            })
+            .await
+            .map_err(|e| {
+                tracing::error!("Failed to initialize MySQL timezone database: {e}");
+                e.to_string()
+            })?;
+
+            let mut ds_system = make_mysql_dataset("tz_test", "tz_system_tbl", MYSQL_PORT3, false);
```
sgrebnov, line 429: Test does not include `local_system` as result will vary on different runners

crates/runtime/src/dataconnector/mysql.rs
```diff
@@ -88,6 +88,8 @@ const PARAMETERS: &[ParameterSpec] = &[
     ParameterSpec::component("pool_max")
         .description("The maximum number of connections to allow in the pool.")
         .default("100"),
+    ParameterSpec::component("time_zone")
+        .description("The time zone to use for the connection. Default is '+00:00' (UTC)."),
```
Jeadie, line 92: Nit: add `local_system` and `system` to description.

Jeadie (approved):

Issue #6412, opened 2025-07-09
Refactor `Embed` trait to return Spice specific errors, not OpenAIError


phillipleblanc:
The Embed trait should return our own Snafu errors that we can downcast specific variants of, not the OpenAi Error: https://github.com/spiceai/spiceai/blob/trunk/crates/llms/src/bedrock/embed/mod.rs#L206C42-L206C53

PR #6427, opened 2025-07-12
Refactor Embed trait to return Spice specific errors instead of OpenAI error

kumarlokesh:
## 📝 Summary

**Error Type Unification**:
   - Replaced `OpenAIError` with `EmbeddingError` across all `Embed` trait implementations
   - Removed redundant error type definitions and consolidated error handling
   - Added proper error conversions where needed

## 🔗 Related

Addresses #6412


crates/llms/src/bedrock/embed/mod.rs
```diff
@@ -203,10 +203,14 @@ where
     async fn embed_request(
         &self,
         req: CreateEmbeddingRequest,
-    ) -> Result<CreateEmbeddingResponse, OpenAIError> {
+    ) -> EmbedResult<CreateEmbeddingResponse> {
```
phillipleblanc, line 224: Thanks for the PR! It looks like we're returning an OpenAIError on line 90 as well - ideally we would clean up all instances of this with this PR.

Jeadie:
Closing in favour of #6748

PR #6468, opened 2025-07-16
Amazon S3 Vectors support

lukekim:


crates/llms/src/bedrock/embed/mod.rs
```diff
@@ -241,10 +355,52 @@ where
...
+                )
+            }),
+        )
+    }
+}
+
+impl Default for BedrockRateLimitConfig {
+    fn default() -> Self {
+        Self {
+            requests_per_minute_limit: 5_000,
```
sgrebnov: Propose to change this to 1_500. Based on recent experiments limit higher that ~1_500 leads to rate limiting/throttling for standard AWS account w/o elevated limits. Note: we can add this change after merging this PR.

sgrebnov (dismissed):

crates/llms/src/bedrock/embed/mod.rs
lukekim:
```suggestion
            requests_per_minute_limit: 1_500,
```

sgrebnov (approved):

PR #6469, opened 2025-07-16
Add integration tests for S3 Vectors filters pushdown

sgrebnov:
## 📝 Summary

1. Add integration test for S3 Vectors filter predicate pushdown to S3 Vectors store (query) based on S3 JSON array data and local embeddings model.
1. Enable S3 Vectors integration tests in CI
1. Prepare Bedrock embeddings tests to run in CI via `extended_tests` feature

>S3VectorsQueryExec: filter={$and:[{attempt_count:{$gt:1}},{package_weight_kg:{$gt:5.0}},{message.status:{$eq:"FAILED"}}]} limit=100

## 👀 Notes for Reviewers

Note: model for existing s3_vectors basic test changed from Bedrock to Local so snapshot has been updated


crates/runtime/tests/models/snapshots/integration_models__models__s3_vectors__search__basic.snap
```diff
@@ -2,11 +2,11 @@
...
-+-----------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+---------------------------------------+
-| 19        | Patient                                                                                                                                                                                                 | 1.0                                   |
-| 7         | New                                                                                                                                                                                                     | 0.26                                  |
-| 12        | Theories take various bottles. Lucky, modern matters keep sessions; special sides see machines; still spatial activities disguise at a characteristics. Long, structural patients used to ask           | 0.22                                  |
```
Sevenannn, line 10: Is the search score going to change across search runs? Or a fixed score is expected

sgrebnov: This is due to new model - bedrock -> hf_minilm and changing round from 2 decimals to single 1 to be more robust

crates/runtime/tests/models/s3_vectors.rs
```diff
@@ -29,74 +35,136 @@ mod search {
...
+                "hf_minilm",
+            ))
             .build();
 
         let rt = start_app(app).await?;
 
         run_and_snapshot_query(
             &rt,
-            "SELECT i_item_sk, i_item_desc, round(score, 2) FROM vector_search(item, 'Patient') where i_item_sk > 5 order by score desc LIMIT 4;",
+            "SELECT i_item_sk, i_item_desc, round(score, 1) FROM vector_search(item, 'Patient') where i_item_sk > 5 order by score desc LIMIT 4;",
```
sgrebnov, line 79: This is added to make test more robust, ensure rounded score is the same across test runs

crates/runtime/tests/models/s3_vectors.rs
```diff
@@ -29,74 +35,136 @@ mod search {
...
+            &rt,
+            r#"
+            SELECT 
+              "message.body",
+              attempt_count, "message.status",
+              package_weight_kg,
+              round(score, 1)
+            FROM vector_search(delivery, 'wrong location')
+            WHERE attempt_count > 1 AND package_weight_kg > 5.0 AND "message.status"='FAILED'
+            ORDER BY package_weight_kg desc, score DESC
```
sgrebnov, line 135: We order on package_weight_kg and not on score as we have multiple items with similar content so score and order of records may fluctuate across test runs

Jeadie: nit, should be comment for the future

crates/runtime/tests/models/s3_vectors.rs
```diff
@@ -163,3 +231,63 @@ mod search {
...
+            ("client_timeout".to_string(), "120s".to_string()),
+        ]
+        .into_iter()
+        .collect(),
+    ));
+    dataset.acceleration = Some(Acceleration {
+        enabled: true,
+        refresh_sql: Some(
+            refresh_sql
+                .unwrap_or(&format!("SELECT * FROM {ds_name}"))
```
Jeadie, line 263: Why is this needed `"SELECT * FROM {ds_name}"))`? Isn't this the default?

Jeadie (approved):

crates/runtime/tests/models/s3_vectors.rs
sgrebnov: Confirm this is default option, just to test that refresh_sql can be used with s3 vectors and also to support some specific refresh sql

Issue #6418, opened 2025-07-10
Enhancement: Evict data from an acceleration based on `retention_sql`


phillipleblanc:
## Goal-State/What/Result

A new acceleration parameter `retention_sql` is supported that when combined with `retention_check_enabled` uses the filter predicates defined in `retention_sql` to delete the data. Takes the form `DELETE FROM <table> WHERE <predicates>`

```yaml
datasets:
  - from: s3://my-bucket/my-dataset/
    name: my_dataset
    params: ...
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
      retention_check_enabled: true
      retention_check_interval: 5m
      retention_sql: DELETE FROM my_dataset WHERE deleted = 'true'
```

## Why/Purpose

There are scenarios where it would be useful to evict data in an acceleration based on a predicate.

A motivating example is to "delete" rows from an append-only acceleration by marking the row as soft-deleted and then pruning the soft-deleted rows.

## By When

**Issue/Spec written and reviewed:** _Target Date_
**Done-Done:** _Target Date_

## Done-Done

- [ ] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Telemetry / Metrics / Task History
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials

## The Algorithm

- [ ] Every requirement questioned?
- [ ] Delete (Scope) any part you can.
- [ ] Simplify.
- [ ] Break down into smaller iterations/milestones.
- [ ] Opportunities for automation.

PR #6488, opened 2025-07-18
Add support for `retention_sql` acceleration param

sgrebnov:
## 📝 Summary

This PR adds support for the `retention_sql` parameter for acceleration, which defines a custom filter to evict data from the accelerator. 

When both `retention_sql` and `retention_period` are defined, `retention_sql` takes precedence. Unlike `retention_period`, `retention_sql` does not require a `time_column`

## 🔗 Related

Implements basic functionality for
- https://github.com/spiceai/spiceai/issues/6418

## 👀 Notes for Reviewers

Next steps:
1. Add integration tests (for both retention modes)
2. Add support for simultaneous `retention_sql` and `retention_period` defined
3. Review if `retention_sql` must be applied during refresh to avoid fetching unnecessary data similar to retention period (can this break CDC / append modes with with upsert param ??)

### DIFF
--- a/crates/runtime/src/accelerated_table/mod.rs
+++ b/crates/runtime/src/accelerated_table/mod.rs
@@ -14,7 +14,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use std::time::SystemTime;
 use std::{any::Any, sync::Arc, time::Duration};
 
 use crate::component::dataset::acceleration::{RefreshMode, RefreshOnStartup, ZeroResultsAction};
@@ -24,23 +23,20 @@ use crate::datafusion::error::SpiceExternalError;
 use crate::datafusion::is_spice_internal_dataset;
 use crate::federated_table::FederatedTable;
 use crate::status;
-use arrow::array::UInt64Array;
 use arrow::datatypes::SchemaRef;
 use arrow::error::ArrowError;
 use async_trait::async_trait;
 use cache::Caching;
 use data_components::cdc::ChangesStream;
-use data_components::delete::get_deletion_provider;
 use datafusion::catalog::Session;
 use datafusion::error::{DataFusionError, Result as DataFusionResult};
+use datafusion::logical_expr::TableProviderFilterPushDown;
 use datafusion::logical_expr::dml::InsertOp;
-use datafusion::logical_expr::{Operator, TableProviderFilterPushDown};
+use datafusion::physical_plan::ExecutionPlan;
 use datafusion::physical_plan::union::UnionExec;
-use datafusion::physical_plan::{ExecutionPlan, collect};
 use datafusion::sql::TableReference;
 use datafusion::{
     datasource::{TableProvider, TableType},
-    execution::context::SessionContext,
     logical_expr::Expr,
 };
 use opentelemetry::KeyValue;
@@ -50,7 +46,6 @@ use synchronized_table::SynchronizedTable;
 use tokio::sync::{Notify, RwLock, Semaphore, mpsc};
 use tokio::task::JoinHandle;
 
-use crate::datafusion::filter_converter::TimestampFilterConvert;
 use crate::execution_plan::TableScanParams;
 use crate::execution_plan::fallback_on_zero_results::FallbackOnZeroResultsScanExec;
 use crate::execution_plan::schema_cast::SchemaCastScanExec;
@@ -62,6 +57,7 @@ mod metrics;
 pub mod refresh;
 pub mod refresh_task;
 mod refresh_task_runner;
+mod retention;
 mod sink;
 mod synchronized_table;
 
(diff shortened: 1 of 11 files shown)


crates/runtime/src/accelerated_table/mod.rs
```diff
@@ -575,131 +571,6 @@ impl AcceleratedTable {
 
         Ok(())
     }
-
-    #[allow(clippy::cast_possible_wrap)]
-    #[allow(clippy::cast_possible_truncation)]
-    #[allow(clippy::too_many_lines)]
-    async fn start_retention_check(
```
sgrebnov, line 582: Moved into own `retention.rs` module

crates/runtime/src/accelerated_table/retention.rs
```diff
@@ -0,0 +1,223 @@
...
+
+            let Some(retention_period) = retention.period else {
+                tracing::error!(
+                    "[retention] The `retention_period` parameter must be specified for retention"
+                );
+                return;
+            };
+
+            let Some(converter) =
+                create_timestamp_filter_converter(&retention, &accelerator, time_column)
```
sgrebnov, line 76: This is to keep original logic, the following method returns Option<>

```rust
TimestampFilterConvert::create(
        field.cloned(),
        Some(time_column.to_string()),
        retention.time_format,
        partition_field.cloned(),
        retention.time_partition_column.clone(),
        retention.time_partition_format,
    )
```

crates/runtime/src/accelerated_table/retention.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        loop {
+            interval_timer.tick().await;
+
+            if let Some(deleted_table_provider) = get_deletion_provider(Arc::clone(&accelerator)) {
+                let expr = match &retention_filter {
+                    DataRetentionFilter::DeleteExpr(expr) => {
+                        if is_spice_internal_dataset(&dataset_name) {
+                            tracing::trace!(
+                                "[retention] Evicting data for {dataset_name} with using retention sql expression"
```
sgrebnov, line 97: This can be improved this next to include the actual sql next

crates/runtime/src/accelerated_table/retention.rs
```diff
@@ -0,0 +1,223 @@
...
+                            );
+                        }
+                        expr.clone()
+                    }
+                    DataRetentionFilter::TimeColumn {
+                        converter,
+                        time_column,
+                        retention_period,
+                    } => {
+                        let start = SystemTime::now() - *retention_period;
```
sgrebnov, line 111: This logic is copied as-is from the original implementation.

crates/runtime/src/accelerated_table/retention.rs
```diff
@@ -0,0 +1,223 @@
...
+                    }
+                }
+            } else {
+                tracing::error!("[retention] Accelerated table does not support delete");
+            }
+        }
+    }
+}
+
+fn create_timestamp_filter_converter(
```
sgrebnov, line 198: This logic is copied as-is from the original implementation. Just wrapped in helper function

crates/runtime/src/accelerated_table/retention.rs
```diff
@@ -0,0 +1,223 @@
...
+                converter,
+                time_column: time_column.clone(),
+                retention_period,
+            }
+        };
+
+        let mut interval_timer = tokio::time::interval(retention.check_interval);
+
+        loop {
+            interval_timer.tick().await;
```
peasee, line 90: nit because I know this is copied from the original.

could we use the `scheduler` crate for this? refresh interval will be moved there at some point as well.

crates/spice_cloud/src/lib.rs
```diff
@@ -174,6 +174,7 @@ impl SpiceExtension {
             Some(Duration::from_secs(1800)), // delete metrics older then 30 minutes
             Some(Duration::from_secs(300)),  // run retention every 5 minutes
             true,
+            None,
```
peasee, line 177: there is a lot of `None`'s in this initializer now. Maybe it's time to switch it to a builder?

peasee (approved):

crates/runtime/src/accelerated_table/mod.rs
```diff
@@ -819,15 +690,17 @@ impl TableProvider for AcceleratedTable {
...
-    pub(crate) time_column: String,
+    pub(crate) time_column: Option<String>,
     pub(crate) time_format: Option<TimeFormat>,
     pub(crate) time_partition_column: Option<String>,
     pub(crate) time_partition_format: Option<TimeFormat>,
-    pub(crate) period: Duration,
+    pub(crate) delete_expr: Option<Expr>,
+    pub(crate) period: Option<Duration>,
     pub(crate) check_interval: Duration,
 }
```
phillipleblanc, line 700: To simplify this, I would have an enum for RetentionMode with 3 variants that groups the related params.

enum RetentionMode {
  Time {
    period: Duration,
    time_column: String
    time_format: Option<String>
  }
  Expression {
    delete_expr: Expr
  }
  TimeAndExpression {
    period: Duration,
    time_column: String
    time_format: Option<String>
    delete_expr: Expr
  }
}

crates/runtime/src/accelerated_table/retention.rs
```diff
@@ -0,0 +1,223 @@
...
+    accelerated_table::{Retention, refresh},
+    datafusion::{
+        builder::get_df_default_config, filter_converter::TimestampFilterConvert,
+        is_spice_internal_dataset,
+    },
+    object_store_registry::default_runtime_env,
+};
+
+enum DataRetentionFilter {
+    DeleteExpr(Expr),
```
phillipleblanc, line 40: Yeah, basically a version of this for the Retention struct.

phillipleblanc (approved):

crates/runtime/src/accelerated_table/retention.rs
sgrebnov: Thank you, I'll improve this in next PR

crates/spice_cloud/src/lib.rs
sgrebnov: Thank you, I'll improve this in next PR

crates/runtime/src/accelerated_table/mod.rs
sgrebnov: Yeah, thank you, I'll improve this in next PR

PR #6532, opened 2025-07-21
S3: default to public access if no auth is provided

sgrebnov:

### DIFF
--- a/crates/runtime/src/dataconnector/s3.rs
+++ b/crates/runtime/src/dataconnector/s3.rs
@@ -161,8 +161,14 @@ impl DataConnectorFactory for S3Factory {
                 validator.validate(&mut params).await?;
             }
 
-            // Initialize the AWS SDK and make it available.
-            let _ = object_store_aws_sdk::initialize_sdk_config().await;
+            // `initialize_sdk_config` emits a warning if the credentials provider cannot be initialized
+            // so we skip it if the auth method is public.
+            match params.parameters.get("auth").expose().ok() {
+                None | Some("public") => (),
+                _ => {
+                    let _ = object_store_aws_sdk::initialize_sdk_config().await;
+                }
+            }
 
             let s3 = S3 {
                 params: params.parameters,
--- a/crates/runtime/src/object_store_registry/mod.rs
+++ b/crates/runtime/src/object_store_registry/mod.rs
@@ -101,7 +101,11 @@ impl SpiceObjectStoreRegistry {
                     s3_builder = s3_builder.with_skip_signature(true);
                     load_credentials_from_environment = false;
                 }
-                None => {}
+                None => {
+                    // Default to public if no auth is provided
+                    s3_builder = s3_builder.with_skip_signature(true);
+                    load_credentials_from_environment = false;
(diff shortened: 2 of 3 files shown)


crates/runtime/src/dataconnector/s3.rs
```diff
@@ -161,8 +161,14 @@ impl DataConnectorFactory for S3Factory {
                 validator.validate(&mut params).await?;
             }
 
-            // Initialize the AWS SDK and make it available.
-            let _ = object_store_aws_sdk::initialize_sdk_config().await;
+            // `initialize_sdk_config` emits a warning if the credentials provider cannot be initialized
+            // so we skip it if the auth method is public.
+            match params.parameters.get("auth").expose().ok() {
```
sgrebnov, line 166: I'm under impression is that the same could also happen for `auth: key` where credentials are explicitly provided, we might just want to eliminate tracing for `aws_config::profile::credentials:` but it can be useful so I'm keeping this as a quick workaround for s3 public and we might want to improve this further

kczimm (dismissed):

ewgenius (dismissed):

ewgenius (approved):

crates/runtime/tests/aws_sdk/spicepods/aws_sdk_verify.yaml
```diff
@@ -14,6 +14,7 @@ datasets:
   - from: s3://spiceai-seoul-glue/parquet_taxi_trips_2023/
     name: private_taxi
     params:
+      s3_auth: iam_role
```
phillipleblanc, line 17:
```suggestion
```

This should work without specifying an explicit `s3_auth` - which is consistent with all of our other AWS components

crates/runtime/src/object_store_registry/mod.rs
```diff
@@ -101,7 +101,11 @@ impl SpiceObjectStoreRegistry {
                     s3_builder = s3_builder.with_skip_signature(true);
                     load_credentials_from_environment = false;
                 }
-                None => {}
+                None => {
+                    // Default to public if no auth is provided
+                    s3_builder = s3_builder.with_skip_signature(true);
+                    load_credentials_from_environment = false;
+                }
```
phillipleblanc, line 108: I don't think we should default to public if auth isn't specified - we should ensure public works if they don't specify it, but we should attempt to use credentials if they exist.

crates/runtime/src/dataconnector/s3.rs
```diff
@@ -161,8 +161,14 @@ impl DataConnectorFactory for S3Factory {
                 validator.validate(&mut params).await?;
             }
 
-            // Initialize the AWS SDK and make it available.
-            let _ = object_store_aws_sdk::initialize_sdk_config().await;
+            // `initialize_sdk_config` emits a warning if the credentials provider cannot be initialized
```
phillipleblanc, line 164: We shouldn't do this - we can suppress the warning message from loading the SDK if it can't be loaded instead.

phillipleblanc (approved):

PR #6577, opened 2025-07-23
Initial snapshot tests for 'VectorQueryTableProvider'

Jeadie:
## 📝 Summary
 - This PR adds tests to `VectorQueryTableProvider` to ensure the correct physical plans are being generated.
 - We use a fake vector index structure to ensure this doesn't require proper AWS S3 vector integration.


crates/runtime/src/embeddings/index/query_table.rs
```diff
@@ -297,7 +299,7 @@ impl VectorQueryTableProvider {
         ];
 
         query_table_projection_exprs.extend(metadata_columns_to_exprs(
-            self.vector_index.metadata_columns(),
+            &self.vector_index.metadata_columns(),
```
phillipleblanc: Why did you make `metadata_columns()` return an owned clone when you just reference it?

crates/runtime/src/embeddings/index/query_table.rs
```diff
@@ -555,3 +557,462 @@ fn metadata_columns_to_exprs(metadata_columns: &MetadataColumns) -> Vec<Expr> {
...
+                    Arc::new(self.schema.clone()),
+                    vec![vec![one_row_default_record_batch_for_schema(&Arc::new(
+                        self.schema.clone(),
+                    ))]],
+                )
+                .unwrap(),
+            ))
+        }
+
+        fn metadata_columns(&self) -> MetadataColumns {
```
phillipleblanc: You could compute these metadata column on init and then you don't need to return an owned clone

Jeadie: Yeah I suppose. Am lazy

phillipleblanc (approved):

PR #6579, opened 2025-07-23
Client-supplied cache keys (Spice-Cache-Key)

### LINKED ISSUE #6413: Enhancement: X-Spice-Cache-Key Header Support for Query Caching
## Goal-State/What/Result

Support for HTTP ETag and `X-Spice-Cache-Key` headers in the HTTP and Arrow Flight SQL query APIs to for finer-grained client-side caching control.

- ~~**ETag**: Server generates an ETag for query results, returned in the response header. Clients include the ETag in subsequent requests via `If-None-Match` to check if results are unchanged, receiving a `304 Not Modified` response if valid.~~
  - Scoped out: to provide an ETag against the result set we'd need to hash something in every row (and spool the whole result set). Best left as a user concern.

- **X-Spice-Cache-Key**: Clients can send a custom cache key in the `X-Spice-Cache-Key` header, which the server uses to index cached results, allowing clients to align cache keys with application-specific logic.

## Why/Purpose

Requested feature from customers to improve cache hit rates and reduce redundant query execution.

## By When

**Issue/Spec written and reviewed:** 07/23
**Functional demo:** 07/25
**Done-Done:** 07/30

## Done-Done

- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Tests
- [ ] Documentation

## The Algorithm

- [x] Simplify.

### ETag Support

[scoped out]

### X-Spice-Cache-Key Support
* Endpoints: HTTP `/v1/sql`, `/v1/search`, and Arrow Flight SQL APIs.
* Behavior:
    * Client sends X-Spice-Cache-Key: "<client-key>" in request header.
    * Server uses <client-key> as the cache key instead of query plan or SQL string.
    * If no X-Spice-Cache-Key is provided, falls back to default key (plan or SQL).
    * Response includes Vary: X-Spice-Cache-Key to ensure caches respect the header.
* Configuration:
    * Enabled by default when runtime.caching.results_cache.enabled: true.
* Validation:
    * Server sanitizes X-Spice-Cache-Key to prevent injection.
      * Must match `^([\w-]{1,128})$` (allowing 128 char keys to support e.g. sha512)
    * If an invalid key is provided, emit a warning and fall back to the default caching behavior

### Security Review

- `X-Spice-Cache-Key` has an explicit spec and is sanitized before query

## How/Implementation Plan

n/a

## QA Plan

- Manual testing of /v1/search and /v1/sql
- Integration testing of header behaviors

## Release Notes

**User cache keys for SQL and search endpoints**: Spice now supports user-provided cache keys for the `/v1/search` and `/v1/sql` endpoints. This allows user control over caching semantically equivalent result sets derived by different queries.

Example:

If further customization of caching behavior is desired, the `X-Spice-Cache-Key` header may be set with a custom cache key. Consider the following scenario when querying for [Scihub Journal Access Data](https://www.kaggle.com/datasets/dagsdags02/scihub-users-2017?select=scihub_2_2017.parquet). 

> Find all journals for a specific (user_id, timestamp, city) tuple

```
sql> select * from scihub_journals_accessed where user_id = '1851400' and date_trunc('DAY', timestamp) = '2017-02-16' and city = 'New York';
+---------------------+------------------------------+---------------+---------+---------------+----------+------------+-------------+
| timestamp           | doi                          | ip_identifier | user_id | country       | city     | longitude  | latitude    |
+---------------------+------------------------------+---------------+---------+---------------+----------+------------+-------------+
| 2017-02-16 09:55:06 | 10.1155/2012/650929          | 1000856       | 1851400 | United States | New York | 40.7830603 | -73.9712488 |
| 2017-02-16 09:58:42 | 10.1186/s13071-015-0822-5    | 1000856       | 1851400 | United States | New York | 40.7830603 | -73.9712488 |
| 2017-02-16 10:08:30 | 10.1586/14787210.2013.844066 | 1000856       | 1851400 | United States | New York | 40.7830603 | -73.9712488 |
+---------------------+------------------------------+---------------+---------+---------------+----------+------------+-------------+
```

It might be possible to find the user based on their `ip_identifier` for that location instead (i.e., in the case that the query is made from differing contexts). A semantically equivalent query can be:

> Find all journals for a specific (user, timestamp, ip_identifier) tuple

```
sql> select * from scihub_journals_accessed where user_id = '1851400' and date_trunc('DAY', timestamp) = '2017-02-16' and ip_identifier = 1000856;
+---------------------+------------------------------+---------------+---------+---------------+----------+------------+-------------+
| timestamp           | doi                          | ip_identifier | user_id | country       | city     | longitude  | latitude    |
+---------------------+------------------------------+---------------+---------+---------------+----------+------------+-------------+
| 2017-02-16 09:55:06 | 10.1155/2012/650929          | 1000856       | 1851400 | United States | New York | 40.7830603 | -73.9712488 |
| 2017-02-16 09:58:42 | 10.1186/s13071-015-0822-5    | 1000856       | 1851400 | United States | New York | 40.7830603 | -73.9712488 |
| 2017-02-16 10:08:30 | 10.1586/14787210.2013.844066 | 1000856       | 1851400 | United States | New York | 40.7830603 | -73.9712488 |
+---------------------+------------------------------+---------------+---------+---------------+----------+------------+-------------+
```

In cases such as these, it is possible to set `X-Spice-Cache-Key`:

```bash
$ curl -vvS -XPOST http://localhost:8090/v1/sql \
-H"x-spice-cache-key: 1851400_20170216_north_america" \
-d "select * from scihub_journals_accessed
    where user_id = '1851400'
      and date_trunc('DAY', timestamp) = '2017-02-16'
      and city = 'New York';"

...
< HTTP/1.1 200 OK
< content-type: application/json
< x-cache: Hit from spiceai
< results-cache-status: HIT
< vary: X-Spice-Cache-Key
< vary: origin, access-control-request-method, access-control-request-headers
< content-length: 604
< date: Wed, 23 Jul 2025 20:26:12 GMT
<
{
[
  {
    "timestamp": "2017-02-16 09:55:06",
    "doi": "10.1155/2012/650929",
    "ip_identifier": 1000856,
    "user_id": 1851400,
    "country": "United States",
    "city": "New York",
    "longitude": 40.7830603,
    "latitude": -73.9712488
  },
...
```

Be mindful that setting `X-Spice-Cache-Key` means **you are responsible for asserting the semantic equivalence of your queries**. Consider the scenario below where the same cache key is provided for a different query. This is expected behavior:

```bash
$ curl -vvS -XPOST http://localhost:8090/v1/sql \
-H"x-spice-cache-key: 1851400_20170216_north_america" \
-d "select 1,2,3"
...
< HTTP/1.1 200 OK
< content-type: application/json
< x-cache: Hit from spiceai
< results-cache-status: HIT
< vary: X-Spice-Cache-Key
< vary: origin, access-control-request-method, access-control-request-headers
< content-length: 604
< date: Wed, 23 Jul 2025 20:26:12 GMT
<
{
[
  {
    "timestamp": "2017-02-16 09:55:06",
    "doi": "10.1155/2012/650929",
    "ip_identifier": 1000856,
    "user_id": 1851400,
    "country": "United States",
    "city": "New York",
    "longitude": 40.7830603,
    "latitude": -73.9712488
  },
...
```

mach-kernel:
## 📝 Summary

Users may provide an `X-Spice-Cache-Key` header value as an alternative to plan and SQL string based caching. There are no changes to developer UX w.r.t. configuration. If a valid** cache key is provided, then it is used for result caching. Invalid (or not-provided) keys fall back to the spicepod's configured behavior.

## TODOs
- [x] Bind user cache key in necessary contexts
- [x] Sanitize cache key + warn if invalid
- [x] Implement for `datafusion::query::cache` (...`/v1/sql`)
- [x] Implement for `/v1/search`
- [x] `Vary` header
- [x] Update docs
  - [x] Basic narrative
  - [x] Docs repo PR
- [x] Add to `e2e_test_ci`

## 🔗 Related

Resolves #6413 

## 👀 Notes for Reviewers

- New `CacheKey::User` variant
- New `CacheResponse` struct (feel free to suggest a better name) to encapsulate commonly used context in `datafusion::query::cache`
  - DRY up some logic around e.g. `try_get_cached_result` handling
  - Simplifies `CacheResult` enum (seems we always want to have a `QueryTracker` on hand)
- Allows cache keys with alphanumeric values (plus `'_'` `'-'` up to 128 chars (supports bigger hashes e.g. sha512)
- Add `Vary: X-Spice-Cache-Key` to headers
  - ...there may be an existing `Vary` header in the response (i.e. you may see two lines with a Vary response). This is [allowed for headers that have list values](https://stackoverflow.com/a/60060332)

## 📚 Docs

https://github.com/spiceai/docs/pull/1076


crates/runtime/src/request/cache_control.rs
```diff
@@ -20,13 +20,15 @@ use app::App;
...
-#[derive(Debug, Clone, Copy, Default)]
+#[derive(Debug, Clone, Copy, Default, Eq, PartialEq)]
 pub enum CacheKeyType {
     /// Use the server's default logic (e.g., `LogicalPlan` hash).
     #[default]
     Default,
     /// Use the raw input (e.g., unparsed SQL string) as the cache key.
     Raw,
+    /// Use a user-provided cache key via the X-Spice-Cache-Key header.
+    User,
```
lukekim: Let's call this `Client` or `ClientSupplied` as our "users" are generally machines (apps/services/agents).

crates/runtime/src/request/context.rs
```diff
@@ -329,20 +357,80 @@ impl RequestContextBuilder {
...
             cache_control,
+            user_cache_key,
             dimensions,
             auth_principal: OnceLock::new(),
             extensions: RwLock::new(self.extensions),
         }
     }
+
+    fn sanitize_cache_key(key: String) -> Option<String> {
+        if USER_CACHE_KEY_REGEX.is_match(&key) {
```
lukekim: Will regex in the critical query path meaningfully affect performance?

crates/runtime/src/request/cache_control.rs
mach-kernel: OK -- I like `ClientSupplied`!

crates/runtime/src/request/context.rs
mach-kernel: No, the regexp is compiled once and the expression itself has anchors + match limits

crates/runtime/src/datafusion/query/cache.rs
```diff
@@ -166,33 +197,39 @@ impl Query {
...
-                ));
+                return Ok(CacheResponse::from(
+                    CacheResult::MissOrSkipped,
+                    CacheStatus::CacheBypass,
+                )
+                .with_query_tracker(tracker));
             }
             _ => {
-                return Ok((CacheResult::WrongCacheKeyType(tracker), None));
+                return Ok(CacheResponse::from(
```
phillipleblanc, line 234: You will also need to pass the query tracker here too right?

crates/runtime/src/http/v1/mod.rs
```diff
@@ -262,6 +271,11 @@ fn attach_cache_headers(headers: &mut HeaderMap, results_cache_status: CacheStat
     {
         headers.insert("Results-Cache-Status", val);
     }
+
+    // Tell CDN entry is unique per user cache key
+    if user_key_specified {
+        headers.insert("Vary", HeaderValue::from_static("X-Spice-Cache-Key"));
```
phillipleblanc:
```suggestion
        headers.insert("Vary", HeaderValue::from_static("Spice-Cache-Key"));
```

https://www.rfc-editor.org/rfc/rfc6648

crates/runtime/src/http/v1/search.rs
```diff
@@ -163,6 +167,11 @@ pub(crate) async fn post(
                     headers.insert("Search-Results-Cache-Status", val);
                 }
 
+                // Tell CDN entry is unique per user cache key
+                if request_context.user_cache_key().is_some() {
+                    headers.insert("Vary", HeaderValue::from_static("X-Spice-Cache-Key"));
```
phillipleblanc:
```suggestion
                    headers.insert("Vary", HeaderValue::from_static("Spice-Cache-Key"));
```

crates/runtime/src/request/cache_control.rs
```diff
@@ -72,17 +74,13 @@ impl CacheControl {
...
     pub fn from_headers(headers: &HeaderMap) -> Self {
         // This will be updated later if the runtime parameter `runtime.results_cache.cache_key_type` is present.
-        let cache_key_type = CacheKeyType::Default;
-
-        let Some(cache_control) = headers.get(CACHE_CONTROL) else {
-            return Self::Cache(cache_key_type);
-        };
-        let Ok(cache_control_str) = cache_control.to_str() else {
-            return Self::Cache(cache_key_type);
+        let cache_key_type = match headers.get("x-spice-cache-key") {
```
phillipleblanc:
```suggestion
        let cache_key_type = match headers.get("spice-cache-key") {
```

crates/runtime/src/request/context.rs
```diff
@@ -257,6 +271,14 @@ impl RequestContextBuilder {
             UserAgentCollection::Disabled => UserAgent::Absent,
         };
         self.cache_control = CacheControl::from_headers(headers);
+        self.user_cache_key = match self.cache_control {
+            CacheControl::Cache(CacheKeyType::ClientSupplied) => headers
+                .get("x-spice-cache-key")
```
phillipleblanc:
```suggestion
                .get("spice-cache-key")
```

crates/runtime/src/request/context.rs
```diff
@@ -329,20 +357,80 @@ impl RequestContextBuilder {
...
             extensions: RwLock::new(self.extensions),
         }
     }
+
+    fn sanitize_cache_key(key: String) -> Option<String> {
+        if USER_CACHE_KEY_REGEX.is_match(&key) {
+            Some(key)
+        } else {
+            warn!(
+                "X-Spice-Cache-Key provided for request ({}) is invalid. A valid cache key is \
```
phillipleblanc:
```suggestion
                "Spice-Cache-Key provided for request ({}) is invalid. A valid cache key is \
```

crates/runtime/src/request/context.rs
```diff
@@ -329,20 +357,80 @@ impl RequestContextBuilder {
...
             auth_principal: OnceLock::new(),
             extensions: RwLock::new(self.extensions),
         }
     }
+
+    fn sanitize_cache_key(key: String) -> Option<String> {
+        if USER_CACHE_KEY_REGEX.is_match(&key) {
+            Some(key)
+        } else {
+            warn!(
```
phillipleblanc: I don't think its a good idea to warn here - we could add this to the task history, but this gives a potential attacker an avenue to spam the logs, which could be used to cover up other malicious activity or make it harder to understand what Spice is doing.

crates/runtime/src/request/context.rs
```diff
@@ -329,20 +357,80 @@ impl RequestContextBuilder {
...
+    use crate::request::{CacheKeyType, Protocol, RequestContextBuilder};
+    use http::{HeaderMap, HeaderValue};
+
+    #[test]
+    fn test_bind_user_cache_key() {
+        let mut headers = HeaderMap::new();
+        headers.append("cache-control", HeaderValue::from_static("cache"));
+
+        // Test user-provided cache key
+        headers.append("x-spice-cache-key", HeaderValue::from_static("foo"));
```
phillipleblanc:
```suggestion
        headers.append("spice-cache-key", HeaderValue::from_static("foo"));
```

crates/runtime/src/request/context.rs
```diff
@@ -329,20 +357,80 @@ impl RequestContextBuilder {
...
+
+        assert_eq!(
+            ctx_happy_path.cache_control,
+            CacheControl::Cache(CacheKeyType::ClientSupplied)
+        );
+        assert_eq!(ctx_happy_path.user_cache_key, Some(String::from("foo")));
+
+        // Test invalid user cache key falling back to default behavior
+        headers.remove("x-spice-cache-key");
+        headers.append("x-spice-cache-key", HeaderValue::from_static("foo$$"));
```
phillipleblanc:
```suggestion
        headers.remove("spice-cache-key");
        headers.append("spice-cache-key", HeaderValue::from_static("foo$$"));
```

crates/runtime/src/request/context.rs
```diff
@@ -51,6 +54,10 @@ tokio::task_local! {
 static INTERNAL_REQUEST_CONTEXT: LazyLock<Arc<RequestContext>> =
     LazyLock::new(|| Arc::new(RequestContext::builder(Protocol::Internal).build()));
 
+#[allow(clippy::expect_used)]
+static USER_CACHE_KEY_REGEX: LazyLock<Regex> =
+    LazyLock::new(|| Regex::new(r"^([\w-]{1,128})$").expect("Must compile regex"));
```
Jeadie: nit: we generally would use `unreachable!` for something like this.

crates/runtime/src/datafusion/query/cache.rs
mach-kernel: Yes, good catch

crates/runtime/src/request/context.rs
mach-kernel: TIL, thanks!

crates/runtime/src/datafusion/query/cache.rs
```diff
@@ -80,28 +105,34 @@ impl Query {
         parameters: Option<ParamValues>,
         tracker: Option<QueryTracker>,
     ) -> super::Result<PlanOrCached> {
-        // Try to get cached results first from sql
         let sql_cache_key = CacheKey::Query(sql, parameters.as_ref());
-        let (tracker, cache_status, sql_raw_cache_key) = match Self::try_get_cached_result(
+        let sql_or_user_cache_key = match request_context.client_supplied_cache_key() {
+            Some(user_key) => CacheKey::ClientSupplied(user_key),
+            _ => sql_cache_key,
```
sgrebnov, line 111: Should we move `CacheKey::Query(sql, parameters.as_ref());` from line 108 here?

sgrebnov: I see - there is special logic where  sql_cache_key is required: `// Always use CacheKey::Query when checking the plan cache`

sgrebnov (approved):

phillipleblanc (approved):

PR #6582, opened 2025-07-23
Improved partition pruning

kczimm:
## 📝 Summary
- Added support for inequality operators (>, >=, <, <=) in prune_partition.
- Added `evaluate_function_filter` for scalar function filters (e.g., date_trunc).
- Improved `evaluate_expr` to handle `Case` and `BinaryExpr` for complex expressions.

## 📚 Docs

Need to update pruning limitations in the docs.


crates/runtime-table-partition/src/provider.rs
```diff
@@ -36,7 +36,7 @@ use crate::{
     insert::PartitionerExec,
 };
 
-mod pruning;
+pub mod pruning;
```
peasee, line 39:
```suggestion
pub(crate) mod pruning;
```

nit: This looks unused outside of this crate?

crates/runtime-table-partition/src/provider/pruning.rs
```diff
@@ -102,44 +104,59 @@ pub(crate) fn prune_partition(
...
+                                let is_match = filter_or_udf_value_matches(
+                                    &col_name,
+                                    partition_by,
+                                    partition_value,
+                                    &lit,
+                                )?;
+                                if is_match {
+                                    return Ok(true);
+                                }
+                            }
```
peasee: could this be compacted like:

```rust
let (col_name, literals) = if let Some((col_name, literals) = collect_conditions(...) {
  (col_name, literals)
} else if let Some((col_name, literals) = collect_conditions(...) {
  (col_name, literals)
} else {
  continue
};

for lit in literals {
...
}
```

peasee: oh, I see there's `any_matches` vs `is_match`. I guess they do need to be separate

peasee (dismissed):

crates/runtime-table-partition/src/provider/pruning.rs
```diff
@@ -25,12 +31,16 @@ use datafusion::{
...
 
-/// Collects equality conditions from nested OR expressions, ensuring they are on the same column.
-fn collect_or_equalities(expr: &Expr) -> Option<(String, Vec<ScalarValue>)> {
+/// Collects conditions (equalities or inequalities) from nested expressions for a given operator.
+fn collect_conditions(
+    expr: &Expr,
+    combining_op: Operator,
+    condition_op: Operator,
+) -> Option<(Column, Vec<ScalarValue>)> {
     match expr {
```
phillipleblanc, line 44: If you use `expr.apply` it will handle the recursion for you, and you can match on all of the operators you care about. The benefit is that this will automatically support more complex expressions.

crates/runtime-table-partition/src/provider/pruning.rs
```diff
@@ -237,20 +325,36 @@ fn call(f: &ScalarUDF, args: Vec<ScalarValue>) -> Result<ScalarValue, DataFusion
...
     Ok(bucket_value)
 }
 
-#[cfg(test)]
-mod tests {
-    use std::sync::Arc;
+/// Used for testing only (unit and integration)
+#[must_use]
+#[allow(clippy::expect_used, clippy::missing_panics_doc)]
+pub fn timestamp_nanos(datetime: &str) -> i64 {
```
phillipleblanc: Move this to the test module?

crates/runtime-table-partition/tests/partition_table_provider.rs
```diff
@@ -28,9 +31,10 @@ use datafusion::logical_expr::expr::ScalarFunction;
 use datafusion::logical_expr::{ColumnarValue, ScalarUDF, TableProviderFilterPushDown};
 use datafusion::physical_expr::create_physical_expr;
 use datafusion::physical_plan::{DisplayAs, ExecutionPlan, PlanProperties};
-use datafusion::prelude::*;
 use datafusion::scalar::ScalarValue;
+use datafusion::{arrow, prelude::*};
 use runtime_datafusion_udfs::{bucket, truncate};
+use runtime_table_partition::provider::pruning::timestamp_nanos;
```
phillipleblanc:
```suggestion
use runtime_table_partition::provider::pruning::tests::timestamp_nanos;
```

crates/runtime-table-partition/src/provider/pruning.rs
```diff
@@ -170,59 +187,130 @@ pub(crate) fn prune_partition(
...
+        .collect::<Result<Vec<_>, _>>()?;
+
+    let result = call(func, evaluated_args)?;
+    Ok(&result == partition_value)
+}
 
-    let result = match transformed_expr {
-        Expr::Literal(lit) => lit,
+/// Evaluates an expression to a scalar value.
+fn evaluate_expr(expr: &Expr) -> Result<ScalarValue, DataFusionError> {
```
phillipleblanc: I think we can use the built-in DataFusion expression simplifier for this: https://docs.rs/datafusion-optimizer/latest/datafusion_optimizer/simplify_expressions/expr_simplifier/struct.ExprSimplifier.html

And we don't even need to rewrite the expression to transform the column references to literals, we can use [with_guarantees](https://docs.rs/datafusion-optimizer/latest/datafusion_optimizer/simplify_expressions/expr_simplifier/struct.ExprSimplifier.html#method.with_guarantees)

phillipleblanc: After simplification, if we are left with Expr::Literal(ScalarValue) - then we can use that directly.

crates/runtime-table-partition/src/provider.rs
kczimm: I actually use a function in here for the integration tests.

crates/runtime-table-partition/src/provider/pruning.rs
kczimm: The issue is using it from the integration test. Integration tests are compiled as their own crates and the crate it is testing does not get compiled with `#[cfg(test)]` and so the entire tests module is configured out in this case.

crates/runtime-table-partition/src/provider/pruning.rs
kczimm: I understand how `Expr::apply` will apply a function to the entire expression tree, but I don't see how I would use it effectively here. I recurse specifically down the `BinaryExpr` branches to recombine. It's different than just being interested in visiting each `Expr`. What do you think? Perhaps, I'm just misunderstanding.

phillipleblanc: Yeah it may not be a great fit here, you explicitly are doing bottom up recursion whereas apply does top-down recursion. The main benefit is it would traverse through expressions to get to any inner BinaryExpr, but I'm guessing you probably wouldn't be interested in those anyway since you would only care about the top-level binary expr

(i.e. a CASE statement with a BinaryExpr inside of it)

crates/runtime-table-partition/src/provider/pruning.rs
```diff
@@ -41,58 +56,136 @@ fn collect_or_equalities(expr: &Expr) -> Option<(String, Vec<ScalarValue>)> {
...
+    partition_by: &Expr,
+    col: &Column,
+    filter_value: &ScalarValue,
+    schema: &Schema,
+) -> Result<ScalarValue, DataFusionError> {
+    let transformed_expr = partition_by
+        .clone()
+        .transform(|e| {
+            Ok(match e {
+                Expr::Column(expr_col) if expr_col.name() == col.name() => {
```
phillipleblanc, line 80: I don't think you need to do this transform anymore since you are passing the guarantees?

crates/runtime-table-partition/src/provider/pruning.rs
```diff
@@ -41,58 +56,136 @@ fn collect_or_equalities(expr: &Expr) -> Option<(String, Vec<ScalarValue>)> {
...
+    let simplifier = ExprSimplifier::new(context).with_guarantees(guarantees.clone());
+
+    let simplified_expr = simplifier.simplify(expr.clone())?;
+
+    if let Expr::Literal(lit) = simplified_expr {
+        return Ok(lit);
+    }
+
+    // Fallback to direct evaluation if simplification doesn't yield a literal
+    // An example of this occurs in `test_prune_partition_case` because regex_match
```
phillipleblanc, line 117: That is unfortunate :( I was hoping the expression simplifier would be smart enough to evaluate UDFs

crates/runtime-table-partition/src/provider/pruning.rs
phillipleblanc: Then I would just copy this code into the integration test as well - I don't think it makes sense to include this in the final binary.

phillipleblanc (dismissed):

crates/runtime-table-partition/src/provider/pruning.rs
kczimm: Fair

crates/runtime-table-partition/src/provider/pruning.rs
kczimm: Unfortunately, if I don't do this transform, my test case for the modulo operator does not work. https://docs.rs/datafusion-expr-common/46.0.1/src/datafusion_expr_common/interval_arithmetic.rs.html#966

phillipleblanc (approved):

PR #6589, opened 2025-07-24
Flight REPL autocomplete

krinart:
## 📝 Summary

Implements autocomplete for flight REPL. 

Supports:
* keywords
* functions
* tables
* columns

Autocomplete

## 🔗 Related

N/A

## 🚨 Breaking Changes

N/A

## 📚 Docs

N/A


crates/flightrepl/src/completer.rs
```diff
@@ -0,0 +1,502 @@
...
+
+    // DML operations
+    "insert", "into", "values", "update", "set", "delete",
+
+    // DDL operations
+    "create", "table", "alter", "drop", "index", "view", "database", "schema",
+];
+
+#[rustfmt::skip]
+const FUNCTION_NAMES: &[&str] = &[
```
mach-kernel: Keeping parity with this const as DataFusion and Spice changes could be tough. It would be cool if the autocomplete could ask Spice via query what UDFs it supports:

- The `SessionContext` function registry can give you a list of supported names: [`udfs()`](https://docs.rs/datafusion/latest/datafusion/execution/context/struct.SessionContext.html#method.udfs)
- You could expose a "reflective" table such as `spice_functions` as a UDTF: [(simple udtf example)](https://github.com/apache/datafusion/blob/main/datafusion-examples/examples/simple_udtf.rs).

Then on REPL boot you could obtain this list via a query, and never have to worry about updating anything!

kczimm: This is a good point. We have recently [added UDFs](https://github.com/spiceai/spiceai/issues/6183) and are likely to add more in the future.

crates/flightrepl/src/completer.rs
```diff
@@ -0,0 +1,502 @@
...
+use rustyline::Context;
+use rustyline::completion::{Completer, Pair};
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::{RwLock, oneshot};
+use tokio::time::interval;
+use tonic::transport::Channel;
+
+#[rustfmt::skip]
+const SQL_KEYWORDS: &[&str] = &[
```
kczimm: Datafusion has a [list of keywords](https://docs.rs/datafusion/latest/datafusion/logical_expr/sqlparser/dialect/keywords/constant.ALL_KEYWORDS.html) from it's SQL parser.

crates/flightrepl/src/lib.rs
```diff
@@ -147,8 +153,32 @@ impl ConditionalEventHandler for KeyEventHandler {
...
-#[derive(Completer, Helper, Hinter, Validator)]
-struct EditorHelper;
+#[derive(Helper, Hinter, Validator)]
+struct EditorHelper {
+    schema_cache: Arc<RwLock<SchemaCache>>,
+    flight_client: Option<FlightServiceClient<Channel>>,
+    api_key: Option<String>,
+    user_agent: String,
+    refresh_task_handle: Option<JoinHandle<()>>,
+    shutdown_sender: Option<oneshot::Sender<()>>,
```
kczimm, line 162: Might want to `impl Drop` and send the shutdown signal.

crates/flightrepl/src/completer.rs
mach-kernel: Along these lines, I think it would be easier to use the [sqlparser](https://docs.rs/sqlparser/latest/sqlparser/) crate to incrementally parse the query as the user enters it. You would be able to transform an assertion like this: 
https://github.com/spiceai/spiceai/pull/6589/files#diff-aba255444129f489045ef42af55102c0d3166a1edf4a629c0cc38cb73f469992R141-R143

To one that folds over the AST "so-far" and realizes "oh, there's a select, but no table ident!".

krinart:
Awesome, thank you guys for the great feedback!

crates/flightrepl/src/completer.rs
krinart: Thanks, this one is done.  The only remaining question is how to properly register this udtf.

One way is to simply add this code:
```
let ctx = &self.datafusion().ctx;
ctx.register_udtf("list_udfs", Arc::new(ListUDFTableFunc::new(ctx.clone())));
```

somewhere during intialization, for example as part of `load_components` method. But that's kinda ugly.

Another way is to utilize some existing mechanism, for example extensions. Even though my understanding extensions are supposed to be configured from the outside, maybe I could extend [this code](https://github.com/spiceai/spiceai/blob/trunk/bin/spiced/src/lib.rs#L196-L206) to always add extension which will initialize this udtf? 

Or maybe there's some better. Please let me know what you think

crates/flightrepl/src/lib.rs
krinart: Done

crates/flightrepl/src/completer.rs
krinart: I tried to play with this idea, but the problem is it only works when we have a completely valid query. 

For example it fails to process these strings, where current implementation properly handles them
```
SELE
SELECT
SELECT * FROM table1 inner 
SELECT * FROM table1 inner jo
SELECT * FROM table1 inner join 
```

So even example ` there's a select, but no table ident` won't work.

Given this limitation I feel like it won't add much benefit. I think the current implementation is a good balance between user experience (handles most cases) and complexity (not too complex).

peasee:
@krinart looks like there are quite a few linting failures. Could you please check them out?

crates/flightrepl/src/completer.rs
```diff
@@ -0,0 +1,552 @@
...
+                }
+            } else {
+                // Check if the last word is "join" (could be after WHERE clause)
+                before_cursor.trim().ends_with("join")
+            };
+
+        if should_suggest_only_tables {
+            // Only suggest tables
+            for table in &cache.tables {
+                if table.to_lowercase().starts_with(&word_lower) {
```
peasee: `.to_lowercase()` seems to be on the hot path for `table` and `column` checks. Could we pre-compute the lowercase value as well, and store it with the original value in some struct/tuple? Maybe something like:

```rust
struct StringValue {
  lower: Arc<str>,
  original: Arc<str>
}
```

I guess there could be a concern if there are a lot of keywords/tables that we're storing duplicate values in memory. If the memory size is relatively small though, I think this might be worth the tradeoff.

crates/flightrepl/src/completer.rs
```diff
@@ -0,0 +1,552 @@
...
+}
+
+fn extract_word(line: &str, pos: usize) -> (usize, &str) {
+    let pos = pos.min(line.len());
+    let chars: Vec<char> = line.chars().collect();
+
+    // Find start of current word
+    let mut start = pos;
+    while start > 0 {
+        let ch = chars[start - 1];
```
peasee: I'm not sure if linting is returning an error for this or not, but I think making this `unreachable!()` would be preferred:

```rust
let Some(ch) = chars.get(start - 1) else {
  unreachable!("Start position should always be equal to the length of the characters");
};
```

ditto for the other array accesses

crates/flightrepl/src/completer.rs
```diff
@@ -0,0 +1,552 @@
...
+            user_agent: "test".to_string(),
+            refresh_task_handle: None,
+            shutdown_sender: None,
+        }
+    }
+
+    fn get_completions(helper: &EditorHelper, line: &str, pos: usize) -> Vec<String> {
+        let history = MemHistory::new();
+        let ctx = Context::new(&history);
+        let result = helper.complete(line, pos, &ctx).unwrap();
```
peasee:
```suggestion
        let result = helper.complete(line, pos, &ctx).expect("Should complete suggestion");
```

Prefer `.expect()` instead of `.unwrap()`

crates/flightrepl/src/completer.rs
krinart: Thanks, done

crates/flightrepl/src/completer.rs
krinart: Thanks, done

crates/flightrepl/src/completer.rs
krinart: Great point. Done

crates/flightrepl/src/completer.rs
```diff
@@ -0,0 +1,570 @@
...
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::{RwLock, oneshot};
+use tokio::time::interval;
+use tonic::transport::Channel;
+
+#[derive(Debug, Clone)]
+struct StringValue {
+    #[allow(dead_code)]
+    original: Arc<str>,
```
peasee:
```suggestion
    _original: Arc<str>,
```

nit: prefixing with underscore should work?

peasee (dismissed):

crates/flightrepl/src/completer.rs
krinart: Done

crates/flightrepl/src/completer.rs
Jeadie: I agree, if we can't get the benefits of `sqlparser`, using `datafusion::sql::sqlparser::keywords::ALL_KEYWORDS` may be the most we can do.

crates/flightrepl/src/completer.rs
Jeadie: One approach here to keep at least the names in sync (between runtime and flightrepl) is:
1. Define a function in `runtime` crate that provides all runtime-specific UDFs name (e.g. `available_runtime_udf_names`). 
2. Provide to flightrepl via `spiced` binary, [here](https://github.com/spiceai/spiceai/blob/b063058d74960c5f9caff38312d5144625e12df8/bin/spiced/src/main.rs#L77)
    ```rust
    if let Err(e) = tokio_runtime.block_on(flightrepl::run(args.repl_config), runtime::available_runtime_udf_names()) {
    ```

Jeadie: And then make sure to have comments on the relevant `runtime` functions so that when other people add UDFs, they know to update `available_runtime_udf_names`. We do similar [here](https://github.com/spiceai/spiceai/blob/b063058d74960c5f9caff38312d5144625e12df8/crates/runtime/src/datafusion/udf.rs#L22).

krinart: Awesome! Thanks, will do!

krinart: Current approach registers **udtf** which allows following sql code:
```
select * from list_udfs()
```

And this sql returns all currently registered **udfs** - which is used by autocomplete logic in flightrepl. This way there's no need to make any updates where people register new udfs.

Unless I misunderstood your comment, in the solution you suggested, it would be required to perform additional work to make sure autocomplete is aware of all the currently registered udfs when people register new udfs.

peasee (approved):

PR #6595, opened 2025-07-25
Fix retention filter when both retention_sql and period are set

sgrebnov:
## 📝 Summary

Allow `refresh_sql` and time-based filters to be applied independently, combining them into single binary expression using  `OR` before passing into `delete_from` operation.

This fixes retention filtering when both retention_sql and period are set for DuckDB, SQLite, and Postgres acceleration.


ewgenius (dismissed):

crates/runtime/tests/retention/mod.rs
```diff
@@ -53,7 +76,17 @@ async fn test_retention_sql() -> Result<(), anyhow::Error> {
...
+                    // keep only ALGERIA, ARGENTINA and CANADA
+                    "DELETE FROM nation WHERE n_nationkey >= 5 OR n_name NOT LIKE '%A'",
+                ))
+                .with_dataset(make_s3_dataset(
+                    "spiceai-public-datasets/taxi_small_samples/taxi_sample.parquet",
+                    "taxi_trips",
+                    "duckdb",
+                    "DELETE FROM taxi_trips WHERE VendorID != 2 OR Airport_fee != 1.75",
+                    Some("tpep_pickup_datetime"),
+                    Some("1000000000w"), // Some large retention period to ensure data is not fitlered out by time
```
phillipleblanc, line 89:
```suggestion
                    Some("1000000000w"), // Some large retention period to ensure data is not filtered out by time
```

phillipleblanc (dismissed):

crates/runtime/src/accelerated_table/retention.rs
```diff
@@ -102,15 +105,23 @@ impl super::AcceleratedTable {
                     }
                 }
 
-                tracing::trace!("[retention] Exprs {exprs:?}");
+                // Combine all expressions into a single OR expression as time and SQL expressions are applied independently
+                let Some(expr) = exprs.into_iter().reduce(Expr::or) else {
```
Jeadie, line 109: TIL: `exprs.into_iter().reduce(Expr::or)`

crates/runtime/src/accelerated_table/retention.rs
```diff
@@ -102,15 +105,23 @@ impl super::AcceleratedTable {
...
+                tracing::trace!("[retention] Expr {expr:?}");
 
                 let ctx = SessionContext::new_with_config_rt(
                     get_df_default_config(),
                     default_runtime_env(),
                 );
 
                 let plan = deleted_table_provider
-                    .delete_from(&ctx.state(), &exprs)
+                    .delete_from(&ctx.state(), &[expr])
```
Jeadie, line 124: Just to learn, the filters in `DeletionTableProvider` are supposed to be used as an AND  (i.e. only delete data that return true for all binary expression in `&[Expr]`)? 
```rust
#[async_trait]
pub trait DeletionTableProvider: TableProvider {
    async fn delete_from(
        &self,
        _state: &dyn Session,
        _filters: &[Expr],
    ) -> datafusion::error::Result<Arc<dyn ExecutionPlan>>;
}
```

Jeadie (dismissed):

crates/runtime/src/accelerated_table/retention.rs
sgrebnov: @Jeadie  - this is correct, thank you for review
> filters in DeletionTableProvider are supposed to be used as an AND (i.e. only delete data that return true for all binary expression in &[Expr])

phillipleblanc (approved):

PR #6606, opened 2025-07-28
allow column config to specify vector size for federated queries that do not use array types with a fixed length

phillipleblanc:


mach-kernel:
I broke this spike up into two PRs for each separate change:
- #6610 
- #6612

PR #6610, opened 2025-07-28
Embedding column config to support client-specified vector sizes

mach-kernel:
## 📝 Summary

When federating queries externally, projections may not come back with Arrow `FixedSizeList`. For example, Spark's `ArrayType` is variable length (there is no fixed length equivalent in Spark).

## 📚 Docs

https://github.com/spiceai/docs/pull/1080


crates/runtime/src/embeddings/table.rs
```diff
@@ -129,9 +130,10 @@ impl EmbeddingTable {
 
                 let Some(vector_length) =
                     Self::embedding_size_from_base_table(&column, &base_schema)
+                        .or(config.vector_size.and_then(|sz| i32::try_from(sz).ok()))
                 else {
                     tracing::warn!(
-                        "Column '{}' has embeddings in base table, but the vector length could not be determined. Ignoring column.",
+                        "Column '{}' has embeddings in base table, but the vector length could not be determined from schema. Ignoring column. Provide a value for the vector_size key in the column's embedding configuration.",
                         column
```
kczimm:
```suggestion
                        "Column '{column}' has embeddings in base table, but the vector length could not be determined from schema. Ignoring column. Provide a value for the 'vector_size' key in the column's embedding configuration."
```

kczimm: nit

kczimm (dismissed):

kczimm (approved):

PR #6614, opened 2025-07-28
chore: upgrade to Rust 1.87.0

### LINKED ISSUE #6587: Enhancement: Upgrade to Rust 1.87.0
## Goal-State/What/Result

Upgrade Spice to use Rust 1.87.0 https://blog.rust-lang.org/2025/05/15/Rust-1.87.0/

## Why/Purpose

Rust 1.88 is released now: https://blog.rust-lang.org/2025/06/26/Rust-1.88.0/
Our policy of staying one version behind means we can now upgrade to 1.87.

## By When

**Done-Done:** _Target Date_

## Done-Done

- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Tests
- [ ] Documentation

## The Algorithm

- [ ] Simplify.

kczimm:
## 📝 Summary

- Upgrades to Rust 1.87.0
- Fixes new clippy lints

## 🔗 Related

Closes:
- #6587


bin/spiced/src/lib.rs
```diff
@@ -274,8 +276,7 @@ pub async fn run(args: Args) -> Result<()> {
 
     let result = match server_thread.await {
         // Don't treat force terminated as an error
-        Ok(Err(runtime::Error::ForceTerminated)) => Ok(()),
-        Ok(ok) => ok.context(UnableToStartServersSnafu),
```
Advayp, line 278: Why was this line removed? The new version doesn't add the context to the error

bin/spiced/src/lib.rs
```diff
@@ -51,13 +51,13 @@ mod tls;
 #[derive(Debug, Snafu)]
 pub enum Error {
     #[snafu(display("Unable to construct spice app: {source}"))]
-    UnableToConstructSpiceApp { source: app::Error },
+    UnableToConstructSpiceApp { source: Box<app::Error> },
```
Advayp, line 54: Was boxing this error because of clippy?

kczimm:
Most of the lints were with regard to the size of the enum variants being too large. In some cases, I ignored it and others I boxed the inner type to reduce the size of the type.

crates/runtime/src/datafusion/retention_sql.rs
```diff
@@ -172,6 +174,7 @@ fn validate_table_name(
     Ok(())
 }
 
+#[allow(clippy::result_large_err)]
```
Advayp, line 177: I noticed we're allowing this a lot. Is it possible to disable clippy's check for it entirely? Also, it seems like we're addressing this lint in certain places (by boxing certain errors), but allowing it in others. Is there a particular reason for this?

lukekim (dismissed):

bin/spiced/src/lib.rs
kczimm: Wow great catch. I don't know how that got removed. I tried to automate some of these lint fixes with clippy's `--fix`. Not sure when this occurred. Thanks!

crates/runtime/src/datafusion/retention_sql.rs
kczimm: Fixing some of these errors appeared to be more involved. It's sort of a rabbit hole. You have an enum variant that is particularly large so it makes your entire enum type large. If you go to the variant's inner type you see that it has field or variants that are large and you keep recursing. In an ideal world, you optimize the size all the way down. I did this when it appeared to be a shallow recursion but ones that seemed deeper I allowed here. It might be a nice reminder to leave these so that when you're dealing with these Error types, we can refactor incrementally. Or so that was the thought.

bin/spiced/src/lib.rs
kczimm: Yeah, https://rust-lang.github.io/rust-clippy/stable/index.html#/result_large_err

Advayp (approved):

phillipleblanc (approved):

PR #6629, opened 2025-07-29
Fixes for Databricks OpenAI compatibility

Jeadie:
## 📝 Summary
 - Fixes for Databricks OpenAI compatibility, specifically to ensure `databricks-llama-4-maverick` works.
 - Note: Since Maverick does not support multi-turn tool usage, it cannot be used with `list_datasets` (or `auto`), since Spice will default use this tool before the initial tool call.


mach-kernel (approved):
Maybe we can file a bug report with Databricks about this?

PR #6635, opened 2025-07-29
Fix multiple query error being classified as an internal error

### LINKED ISSUE #6621: Bug: Multiple statements in a query should not be an internal error
## Describe the bug

Running multiple statements in a query returns "Internal Error"

```bash
sql> select "review_id", "review_date", "customer_id", "review_headline", "review_body" from spice.public.realtime_reviews where lower("review_body") like '%disappointed%' or lower("review_body") like '%frustrated%' or lower("review_body") like '%unhappy%' or lower("review_body") like '%not happy%' or lower("review_body") like '%regret%' or lower("review_body") like '%poor%' or lower("review_body") like '%bad%' or lower("review_body") like '%does not work%' or lower("review_body") like '%broke%' or lower("review_body") like '%broken%' or lower("review_body") like '%defective%' order by "review_date" desc limit 10;SELECT *
FROM vector_search(realtime_reviews, 'disappointed customer negative sentiment', 10)
WHERE review_body ILIKE '%disappointed%' OR review_headline ILIKE '%disappointed%'
ORDER BY score DESC;
Internal Error: An unexpected internal error occurred. Execute '.error' for details.
sql> .error
An internal error occurred.
This feature is not implemented: The context currently only supports a single SQL statement
sql> 
```

## To Reproduce

2. Click on '...'
3. Scroll down to '...'

## Expected behavior

### Spicepod

### Output of `explain query`

E.g. `explain select 1`

```bash
```

### spice, spiced, OS info

- OS info: `uname -a`

- [ ] Yes
- [ ] No

Advayp:
## 📝 Summary
- See title
- Necessary as it improves UX and fixes a bug
image

## 🔗 Related

- Closes #6621

### DIFF
--- a/crates/runtime/src/flight/mod.rs
+++ b/crates/runtime/src/flight/mod.rs
@@ -330,8 +330,10 @@ fn handle_datafusion_error(e: DataFusionError) -> Status {
                 Status::internal("Several DataFusion errors occurred, but no details available")
             }
         }
+        DataFusionError::NotImplemented(message) => {
+            Status::invalid_argument(format!("Feature not implemented: {message}"))
+        }
         DataFusionError::Internal(_)
-        | DataFusionError::NotImplemented(_)
         | DataFusionError::ArrowError(..)
         | DataFusionError::IoError(_)
         | DataFusionError::ObjectStore(_)
--- a/crates/runtime/src/flight/mod.rs
+++ b/crates/runtime/src/flight/mod.rs
@@ -331,7 +331,7 @@ fn handle_datafusion_error(e: DataFusionError) -> Status {
             }
         }
         DataFusionError::NotImplemented(message) => {
-            Status::invalid_argument(format!("Feature not implemented: {message}"))
+            Status::invalid_argument(format!("Unsupported Query. {message}"))
         }
         DataFusionError::Internal(_)
         | DataFusionError::ArrowError(..)


sgrebnov (dismissed):

kczimm (dismissed):

crates/runtime/src/flight/mod.rs
```diff
@@ -330,8 +330,10 @@ fn handle_datafusion_error(e: DataFusionError) -> Status {
                 Status::internal("Several DataFusion errors occurred, but no details available")
             }
         }
+        DataFusionError::NotImplemented(message) => {
+            Status::invalid_argument(format!("Feature not implemented: {message}"))
```
lukekim: Do we need the `Feature not implemented`? What's some examples of the messages that come through? Might be nice to have some examples in a comment.

lukekim: Also, feature not implemented is an internal message to DF consumers, not friendly for end users. Should be something like `X is not supported: `

Advayp: > Do we need the `Feature not implemented`? What's some examples of the messages that come through? Might be nice to have some examples in a comment.

@lukekim, They vary, some include in the message itself saying an action is unsupported, like:
```
Unsupported ast node Function { lateral: true, name: ObjectName([Identifier(Ident { value: "unnest", quote_style: None, span: Span(Location(6,20)..Location(6,26)) })]), args: [Unnamed(Expr(Identifier(Ident { value: "review_comments", quote_style: None, span: Span(Location(6,27)..Location(6,42)) })))], alias: Some(TableAlias { name: Ident { value: "comment", quote_style: None, span: Span(Location(6,47)..Location(6,54)) }, columns: [] }) } in create_relation
```
Some are less clear about where the issue lies, like:
```
The context currently only supports a single SQL statement
```

I think we should prepend it with `Unsupported Query.`, just to be clear about where the error lies.

lukekim (approved):

PR #6656, opened 2025-07-31
Add retry logic for OpenAI embeddings creation

### LINKED ISSUE #6653: Bug: OpenAI embeddings creation should have retry mechanism
## Describe the bug

When runtime is unable to create embeddings during data load, dataset load is restarted. Instead, we should have a retry mechanism. Example for OpenAI

```
2025-07-30T16:57:31.513091Z  INFO runtime::accelerated_table::refresh_task: Dataset corpus received 2,000 records
2025-07-30T16:57:51.930008Z  WARN runtime::accelerated_table::refresh_task: Failed to load data for dataset corpus (file): Failed to get data from the connector.
External error: External error: Failed to create embedding. server_error: The server had an error processing your request. Sorry about that! You can retry your request, or contact us through our help center at help.openai.com if you keep seeing this error. (Please include the request ID 836d0693-fdb2-4080-bddc-1a809dd1e388 in your email.).
Ensure the dataset configuration is valid, and try again.
2025-07-30T16:57:53.034038Z  INFO runtime::accelerated_table::refresh_task: Loading data for dataset corpus
```

sgrebnov:
## 📝 Summary

Add a default retry mechanism (10 retries) for OpenAI embedding API requests, similar to the Bedrock embedding model. This helps prevent unexpected transient errors from terminating data loading and forcing the dataset load to restart from scratch, as described in #6653.

Note:

1. A more generic approach to retrying at a higher level was also considered. However, since embedding batches are typically split into smaller chunks and processed through multiple API calls, retrying at the individual call level is more efficient (including cost) than retrying the entire batch (often several thousand strings). For example, Bedrock processes a single string per API call, so retrying at a higher level there would be much less efficient.

1. The OpenAI client already has built‑in retry logic for 429 (rate limit) errors. We may consider removing explicit retries for 429 to avoid duplicate handling.
https://github.com/64bit/async-openai/commit/7a2e179d49ff48cf669397adb0b184b1bcd9ca18#diff-b790731ef802a641db0145417d16ccfdae47b4747f4157760138596e6bd9554fR169

1. Recent experiments show most transient errors are unexpected server errors without a specific code, e.g.: 
> server_error: The server had an error processing your request.  Sorry about that! You can retry your request, or contact us through our help center at help.openai.com if you keep seeing this error.

## 🔗 Related

Fixes https://github.com/spiceai/spiceai/issues/6653

### DIFF
--- a/crates/llms/src/openai/embed.rs
+++ b/crates/llms/src/openai/embed.rs
@@ -16,8 +16,11 @@ limitations under the License.
 #![allow(clippy::missing_errors_doc)]
 use async_openai::config::Config;
 use bytes::Bytes;
+use reqwest::StatusCode;
 use std::fmt::Debug;
 use std::sync::Arc;
+use util::fibonacci_backoff::{FibonacciBackoff, FibonacciBackoffBuilder};
+use util::{RetryError, retry};
 
 use crate::chunking::{
     ArcSizer, Chunker, ChunkingConfig, RecursiveSplittingChunker, TokenizerWrapper,
@@ -41,6 +44,10 @@ pub(crate) const TEXT_EMBED_3_SMALL: &str = "text-embedding-3-small";
 
 pub const DEFAULT_EMBEDDING_MODEL: &str = TEXT_EMBED_3_SMALL;
 
+fn default_retry_strategy() -> FibonacciBackoff {
+    FibonacciBackoffBuilder::new().max_retries(Some(10)).build()
+}
+
 /// Embedding implementation for `OpenAI` compatible embedding models.
 ///
 /// For non-OpenAI models, a [`Tokenizer`] can be provided to correctly size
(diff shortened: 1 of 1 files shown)


crates/llms/src/openai/embed.rs
```diff
@@ -41,13 +44,19 @@ pub(crate) const TEXT_EMBED_3_SMALL: &str = "text-embedding-3-small";
 
 pub const DEFAULT_EMBEDDING_MODEL: &str = TEXT_EMBED_3_SMALL;
 
+fn default_retry_strategy() -> FibonacciBackoff {
+    FibonacciBackoffBuilder::new().max_retries(Some(10)).build()
```
peasee, line 48: nit: Would be nice if this was configurable.

peasee (approved):

crates/llms/src/openai/embed.rs
sgrebnov: Yeah, we need to revise this logic across all providers and make it more standardized and configurable(below).  We might extend our checklist to include those into release quality criteria. This is  min first iteration to fix immediate problem where we see transient server errors
 - retry
 - max concurrent requests
 - max requests per minute (rate limiting)

PR #6657, opened 2025-07-31
Make models E2E chat test more robust

sgrebnov:
## 📝 Summary

Accept dataset name instead of requiring full schema. Example failed test run where model returned dataset name instead of full schema:
https://github.com/spiceai/spiceai/actions/runs/16637935687/job/47083342875?pr=6594

### DIFF
--- a/test/models/chat_01.exp
+++ b/test/models/chat_01.exp
@@ -15,16 +15,15 @@ proc verify_datasets_access {} {
     send "What datasets you have access to\r"
 
     expect {
-        "spice.public.taxi_trips" {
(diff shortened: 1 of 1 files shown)


test/models/chat_01.exp
```diff
@@ -15,16 +15,15 @@ proc verify_datasets_access {} {
     send "What datasets you have access to\r"
 
     expect {
-        "spice.public.taxi_trips" {
+        -re "(?i)taxi_trips|taxi trips" {
```
sgrebnov, line 18: This is case insensitive regexp

peasee (approved):

PR #6660, opened 2025-07-31
fix: Hadoop integration test

peasee:
## 📝 Summary

* Updates the Hadoop integration test to run using a pre-existing Hadoop Catalog in MinIO. The file-based catalog is disabled unless the `test_hadoop_catalog_docker` feature is enabled, which uses `docker compose` to create the catalog locally.

Successful run: https://github.com/spiceai/spiceai/actions/runs/16664711742

### DIFF
--- a/.github/workflows/integration_data_components.yaml
+++ b/.github/workflows/integration_data_components.yaml
@@ -42,8 +42,8 @@ env:
   CONTAINER_REGISTRY: spiceaitestimages.azurecr.io/
 
 jobs:
-  build:
-    name: Build Data Components Test Binary
+  test:
+    name: Test Data Components (Hadoop Catalog)
     runs-on: spiceai-dev-runners
     steps:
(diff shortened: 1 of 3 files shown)


lukekim (approved):

phillipleblanc (approved):

.github/workflows/integration_data_components.yaml
```diff
@@ -60,77 +60,13 @@ jobs:
 
       - name: Set up cc
         uses: ./.github/actions/setup-cc
-
```
Jeadie, line 63: Why are we not splitting build/test?

peasee: there is no reason to split it for this test. The other tests split so it can run on `ubuntu-latest`, but because this test needs MinIO we run it on `spiceai-dev-runners` anyway.

PR #6672, opened 2025-08-02
Limit Concurrent Requests for GitHub

Advayp:
## 📝 Summary
- See title
- Necessary as it helps prevent rate limiting in most situations
- Can configure via the following:
```yaml
runtime:
    params:
        github_max_concurrent_connections: 5 # Defaults to 10
```

## 🔗 Related
- Related to #6662 

## 📚 Docs
- Docs will need to be updated

### DIFF
--- a/crates/runtime/src/dataconnector/github/mod.rs
+++ b/crates/runtime/src/dataconnector/github/mod.rs
@@ -73,6 +73,8 @@ mod stargazers;
 static GITHUB_CONCURRENCY_LIMITS: LazyLock<Mutex<HashMap<String, Arc<Semaphore>>>> =
     LazyLock::new(|| Mutex::new(HashMap::new()));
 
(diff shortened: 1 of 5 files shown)


crates/runtime/src/dataconnector/github/mod.rs
```diff
@@ -394,31 +401,60 @@ impl DataConnectorFactory for GithubFactory {
...
+        let max_concurrent_connections = params
+            .app
+            .and_then(|app| {
+                app.runtime
+                    .params
+                    .get("github_max_concurrent_connections")
+                    .cloned()
+            })
+            .and_then(|value| value.parse::<usize>().ok())
+            .unwrap_or(10);
```
Jeadie: nit: Make this a static constant at top of file.

Jeadie (dismissed):

Jeadie (approved):

Issue #6340, opened 2025-06-30
Enhancement: AWS Bedrock Models Provider


lukekim:
## Goal-State/What/Result
Configure and run Amazon Bedrock Nova models in the Spice runtime.

## Why/Purpose
Users of the Spice runtime can currently use Amazon bedrock as an embedding provider. Enabling support for language models increases this scope of integration.

## By When

**Issue/Spec written and reviewed:** 29th July 2025
**Done-Done:** 31st July 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## Specification

Use same Amazon specific parameters as [embedding models](https://spiceai.org/docs/components/embeddings/bedrock#aws-parameters).

For [guardrails](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_GuardrailConfiguration.html)

- `bedrock_guardrail_identifier` (String): The identifier for the guardrail.
  - Length in [0, 2048] inclusive.
  - Pattern: `(([a-z0-9]+)|(arn:aws(-[^:]+)?:bedrock:[a-z0-9-]{1,20}:[0-9]{12}:guardrail/[a-z0-9]+))`.
- `bedrock_guardrail_version` (String): The version of the guardrail. 
  - Pattern: `(([1-9][0-9]{0,7})|(DRAFT))`
- `bedrock_trace`: The trace behavior for the guardrail.
  - Valid Values: `enabled` | `disabled` | `enabled_full` default `disabled`.

For [OpenAI overrides](https://spiceai.org/docs/features/large-language-models/parameter_overrides), the following subet: `maxTokens`, `temperature`, `topP`, `topK`, `stopSequences` will be supported. Support will be come by supporting these values in the request payload.

Model Ids:
 - `us.amazon.nova-lite-v1:0`
 - `us.amazon.nova-micro-v1:0`
 - `us.amazon.nova-premier-v1:0`
 - `us.amazon.nova-pro-v1:0`

```yaml
models:
  - from: bedrock:us.amazon.nova-lite-v1:0
    name: novash
    params:
[...]

PR #6673, opened 2025-08-02
Amazon Bedrock, specifically Nova models

Jeadie:
## 📝 Summary
 - Add new LLM provider, Amazon Bedrock. Support models that are compatible with Converse API. 
 - Primarily for, tested against, Amazon Nova. 
 
 Requires two new Github Secrets `SPICE_SECRET_BEDROCK_ACCESS_KEY` & `SPICE_SECRET_BEDROCK_SECRET_KEY`. Long term AWS keys for testing that have Bedrock permissions. 
 
## 🔗 Related
#6340 

## 📚 Docs
Docs as per enhancement


.github/workflows/integration_llms.yml
```diff
@@ -132,6 +134,23 @@ jobs:
               echo "Error: Perplexity API key is not defined."
               exit 1
             fi
+            if [ -z "$SPICE_PERPLEXITY_AUTH_TOKEN" ]; then
+              echo "Error: Perplexity API key is not defined."
```
phillipleblanc: nit: include the name of the env var it expects in the error message

crates/llms/src/bedrock/chat/mod.rs
```diff
@@ -0,0 +1,619 @@
...
+impl BedrockConverse {
+    #[must_use]
+    pub fn new(client: Arc<BedrockClient>, model_id: String) -> Self {
+        Self { client, model_id }
+    }
+
+    /// Alter the  [`CreateChatCompletionRequest`] in a way that Bedrock understands.
+    /// Instead of performing these changes in the underlying conversion (i.e. within Converse
+    /// APIs), we do it here for consistency with other model providers.
+    fn alter_request(&self, mut req: CreateChatCompletionRequest) -> CreateChatCompletionRequest {
```
phillipleblanc, line 83: These conversion functions should be unit tested

crates/llms/src/bedrock/chat/mod.rs
```diff
@@ -0,0 +1,619 @@
...
+        // Individual packets in the stream may need information from previous packets to be converted correctly to a  [`CreateChatCompletionStreamResponse`]. This struct track all the necessary information.
+        #[derive(Clone, Default)]
+        struct StreamState {
+            id: String,
+            role: Option<Role>,
+            content_block_index_to_tool_details: HashMap<i32, ToolUseBlockStart>,
+            content_block_index_to_delta_idx: HashMap<i32, u32>,
+        }
+
+        let s = stream! {
```
phillipleblanc: Instead of using the `stream!` macro, using `unfold` here (https://docs.rs/futures/latest/futures/stream/fn.unfold.html) and keeping the input stream as part of the `StreamState` might be a more natural fit. And has the benefit of being natural Rust instead of macro Rust.

phillipleblanc: I'm not a huge fan of large code blocks in `stream!` macros - it breaks code completion for rust-analyzer

crates/llms/src/bedrock/chat/mod.rs
```diff
@@ -0,0 +1,619 @@
...
+                                ) {
+                                    Ok(s) => yield Ok(s),
+                                    Err(e) => {yield Err(e); break}
+                                }
+                            },
+                            ConverseStreamOutputPacket::ContentBlockStop(_) => {
+
+                            },
+                            unknown => {
+                                yield Err(to_api_error(format!("Unknown event from Bedrock stream: {unknown:?}")));
```
phillipleblanc: `ConverseStreamOutputPacket` doesn't `impl Display`?

crates/llms/src/bedrock/chat/util.rs
```diff
@@ -0,0 +1,284 @@
...
+        total_tokens: *total_tokens as u32,
+        prompt_tokens_details: cache_read_input_tokens.map(|t| PromptTokensDetails {
+            cached_tokens: Some(t as u32),
+            audio_tokens: None,
+        }),
+        completion_tokens_details: None,
+    }
+}
+
+pub(super) fn value_to_document(value: serde_json::Value) -> Document {
```
phillipleblanc, line 241: I'm pretty sure we have very similar code for S3 vectors - might be a candidate for reuse?

crates/llms/src/bedrock/mod.rs
```diff
@@ -62,48 +79,130 @@ impl BedrockClient {
...
+                Err(e) => Err(RetryError::permanent(
+                    Box::new(e) as Box<dyn std::error::Error + Send + Sync>
+                )),
+            }}
         })
         .await
-        .boxed()?;
+    }
+
+    pub(crate) async fn do_the_thing<O, Fut, F>(
```
phillipleblanc:
```suggestion
    pub(crate) async fn rate_limit_request_with_retry<O, Fut, F>(
```

crates/runtime/src/model/params/bedrock.rs
```diff
@@ -0,0 +1,33 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use super::{COMMON_MODEL_PARAMETERS_WITH_DEPRECATED, PARAM_WITH_DEPRE_LEN, concat_arrays};
+use crate::parameters::ParameterSpec;
+
+pub(crate) const BEDROCK_PARAMETERS: [ParameterSpec; 3] = [
+    ParameterSpec::runtime("aws_secret_access_key")
```
phillipleblanc: To keep parity with our other AWS components (i.e. https://spiceai.org/docs/components/data-connectors/dynamodb) these should be `component` parameters, not runtime parameters.

crates/runtime/src/model/util.rs
```diff
@@ -0,0 +1,84 @@
...
+
+pub(super) async fn create_bedrock_client(
+    params: &HashMap<String, SecretString>,
+    credential_provider_name: &'static str,
+) -> Result<BedrockClient, Box<dyn std::error::Error + Send + Sync>> {
+    // Build AWS config
+    let mut config_builder = aws_config::defaults(aws_config::BehaviorVersion::latest());
+
+    // Set region if provided
+    if let Some(region) = extract_secret!(params, "aws_region") {
```
phillipleblanc, line 38: Note that this won't need to change even if you make the parameter a component parameter (i.e. prefixed)

crates/runtime/src/model/util.rs
```diff
@@ -0,0 +1,84 @@
...
+    if let Some(profile) = extract_secret!(params, "aws_profile") {
+        config_builder = config_builder.profile_name(profile);
+    }
+
+    // Set access key and secret key if provided
+    if let (Some(access_key), Some(secret_key)) = (
+        extract_secret!(params, "aws_access_key_id"),
+        extract_secret!(params, "aws_secret_access_key"),
+    ) {
+        let session_token = extract_secret!(params, "aws_session_token");
```
phillipleblanc, line 52: This needs to be added to the list of parameters

crates/runtime/src/model/params/bedrock.rs
Jeadie: we would need to change it for embedding models too

crates/llms/src/bedrock/chat/util.rs
Jeadie: https://github.com/spiceai/spiceai/issues/6676

crates/llms/src/bedrock/chat/mod.rs
Jeadie: Just `Debug`

crates/llms/src/bedrock/chat/mod.rs
Jeadie: IIRC we had an issue with using `unfold` with the anthropic streaming (which is very similar). I will try a simple stream map

crates/runtime/src/model/params/bedrock.rs
phillipleblanc: Let's keep it as is then

crates/llms/src/bedrock/chat/mod.rs
phillipleblanc: Curious what those issues were? I'm not sure if a map will work here with the stream state - it isn't a pure function of the inputs

crates/runtime/src/model/params/bedrock.rs
Jeadie: Offline discussion. Decided to not change.

crates/llms/src/bedrock/chat/mod.rs
Jeadie: Just implemented and realised why we cannot.  `unfold`  doesn't permit an iteration of the `FnMut` to return no data into the stream (no data signals end of the stream). Will reimplement with `filter_map`.

phillipleblanc: wouldn't you just keep looping then in the unfold implementation until you can emit an item?

Jeadie: unfold does not take a stream. It take an initial state and produces a stream. The original stream is used as apart of the state.

crates/llms/src/bedrock/chat/mod.rs
```diff
@@ -0,0 +1,674 @@
...
+    ) -> ChatCompletionResponseStream {
+        // Individual packets in the stream may need information from previous packets to be converted correctly to a  [`CreateChatCompletionStreamResponse`]. This struct track all the necessary information.
+        #[derive(Default)]
+        struct StreamState {
+            id: String,
+            model: String,
+            role: Option<Role>,
+            content_block_index_to_tool_details: HashMap<i32, ToolUseBlockStart>,
+            content_block_index_to_delta_idx: HashMap<i32, u32>,
+            // input_stream: EventReceiver<ConverseStreamOutputPacket, ConverseStreamOutputError>,
```
phillipleblanc:
```suggestion
```

crates/llms/src/bedrock/chat/mod.rs
Jeadie: Will add in followup to get this into v1.5.2

peasee (approved):

PR #6674, opened 2025-08-03
Make 'runtime-object-store' crate

Jeadie:
## 📝 Summary
 - Reduce how much code is in `runtime` crate.


crates/object-store-extra/src/registry/mod.rs
phillipleblanc, line 1: Call the new crate `runtime-object-store` - that is clearer than `object-store-extra`.

peasee (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #6680, opened 2025-08-04
feat: Add Hadoop `metadata_path` parameter

### LINKED ISSUE #6669: Define a Hadoop metadata file override parameter
If we're unable to infer the correct metadata file, the catalog may fail to load. A parameter should be introduced to allow specifying an exact metadata file location like `metadata_path`.

For example:
```yaml
from: iceberg:s3://some/warehouse/namespace/table
params:
  metadata_path: s3://some/warehouse/namespace/table/metadata/my-metadata-file.json
```

peasee:
## 📝 Summary

* Adds the Hadoop `metadata_path` parameter for the Iceberg data connector.

## 🔗 Related

* Closes #6669

### DIFF
--- a/crates/runtime/src/dataconnector/iceberg.rs
+++ b/crates/runtime/src/dataconnector/iceberg.rs
@@ -31,14 +31,15 @@ use secrecy::ExposeSecret;
 use super::DataConnectorFactory;
 use crate::{
     catalogconnector::iceberg::{
-        get_rest_catalog_config, map_param_name_to_iceberg_prop, parse_hadoop_table_url,
-        parse_table_url, verify_s3_endpoint,
+        ICEBERG_PARAM_LEN, get_rest_catalog_config, map_param_name_to_iceberg_prop,
+        parse_hadoop_table_url, parse_table_url, verify_s3_endpoint,
     },
     component::dataset::Dataset,
     dataconnector::{
         ConnectorComponent, ConnectorParams, DataConnector, DataConnectorError as Error,
         parameters::aws::load_config,
     },
+    model::params::concat_arrays,
     parameters::{ParameterSpec, Parameters},
 };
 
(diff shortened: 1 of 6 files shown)


crates/runtime/src/dataconnector/iceberg.rs
```diff
@@ -57,6 +57,12 @@ impl IcebergDataConnectorFactory {
     }
 }
 
+pub(crate) const PARAMETERS: &[ParameterSpec] = &[
+    // Hadoop options
+    ParameterSpec::runtime("metadata_path")
+        .description("The path including scheme to the metadata file for the Hadoop table. Must specify a path to a `.json` file. For example, `s3a://my-bucket/warehouse/namespace/table/metadata/v1.metadata.json`")
```
phillipleblanc, line 65: We don't support `s3a` links right?

crates/runtime/src/dataconnector/iceberg.rs
```diff
@@ -79,7 +85,12 @@ impl DataConnectorFactory for IcebergDataConnectorFactory {
     }
 
     fn parameters(&self) -> &'static [ParameterSpec] {
-        crate::catalogconnector::iceberg::PARAMETERS
+        // Concat the data connector specific parameters with the catalog parameters
+        let iceberg_params = PARAMETERS.to_vec();
+        let catalog_params = crate::catalogconnector::iceberg::PARAMETERS.to_vec();
+        let mut params = iceberg_params;
+        params.extend(catalog_params);
+        Box::leak(Box::new(params)) // Leak the memory to return a static slice
```
phillipleblanc: The whole point of this being a static slice is that its known at compile time - surely there is a way to have it return a static slice here? Maybe with something similar to what was done for models? https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/model/params/mod.rs#L47

crates/runtime/src/dataconnector/iceberg.rs
peasee: We do. You can `from: iceberg:s3a://` or `from: iceberg:s3://`

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #6694, opened 2025-08-05
Fix AWS auth issue

Advayp:
## 📝 Summary
- Fixes an issue where AWS credentials aren't reused, and are re-initialized on every query

### DIFF
--- a/crates/object-store-aws-sdk/src/credential_provider.rs
+++ b/crates/object-store-aws-sdk/src/credential_provider.rs
@@ -18,18 +18,29 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use aws_config::{BehaviorVersion, SdkConfig};
-use aws_credential_types::provider::ProvideCredentials;
+use aws_credential_types::{Credentials, provider::ProvideCredentials};
 use object_store::{CredentialProvider, aws::AwsCredential};
 
 #[derive(Debug)]
 pub struct S3CredentialProvider {
-    credentials: aws_credential_types::provider::SharedCredentialsProvider,
+    credentials_provider: aws_credential_types::provider::SharedCredentialsProvider,
+    credentials: Option<Credentials>,
 }
 
 impl S3CredentialProvider {
     #[must_use]
     pub fn new(credentials: aws_credential_types::provider::SharedCredentialsProvider) -> Self {
-        Self { credentials }
+        Self {
+            credentials_provider: credentials,
+            credentials: None,
+        }
+    }
+
+    #[must_use]
+    pub fn try_new(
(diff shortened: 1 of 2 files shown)


crates/object-store-aws-sdk/src/credential_provider.rs
```diff
@@ -58,7 +69,34 @@ impl S3CredentialProvider {
...
+        }
+
+        let handle =
+            tokio::runtime::Handle::try_current().map_err(|e| object_store::Error::Generic {
+                store: "S3",
+                source: Box::new(e),
+            })?;
+
+        let creds = tokio::task::block_in_place(|| {
+            handle.block_on(self.credentials_provider.provide_credentials())
```
Advayp, line 90: I didn't make this async because several synchronous functions indirectly call this function.

crates/object-store-aws-sdk/src/credential_provider.rs
```diff
@@ -58,7 +69,34 @@ impl S3CredentialProvider {
...
         Ok(Self {
-            credentials: credentials.clone(),
+            credentials_provider: credentials_provider.clone(),
+            credentials: None,
+        }
+        .load_credentials()?)
+    }
+
+    fn load_credentials(self) -> object_store::Result<Self> {
+        if self.credentials.is_some() {
```
phillipleblanc, line 79: This needs to handle expiration

crates/object-store-aws-sdk/src/credential_provider.rs
```diff
@@ -58,7 +69,34 @@ impl S3CredentialProvider {
...
+            return Ok(self);
+        }
+
+        let handle =
```
phillipleblanc, line 83: We shouldn't use block in place - this needs to be redesigned to avoid it

crates/object-store-aws-sdk/src/credential_provider.rs
```diff
@@ -18,18 +18,29 @@ use std::sync::Arc;
...
 use async_trait::async_trait;
 use aws_config::{BehaviorVersion, SdkConfig};
-use aws_credential_types::provider::ProvideCredentials;
+use aws_credential_types::{Credentials, provider::ProvideCredentials};
 use object_store::{CredentialProvider, aws::AwsCredential};
 
 #[derive(Debug)]
 pub struct S3CredentialProvider {
-    credentials: aws_credential_types::provider::SharedCredentialsProvider,
+    credentials_provider: aws_credential_types::provider::SharedCredentialsProvider,
```
phillipleblanc, line 26: The credential provider doesn't do caching?

phillipleblanc: I would prefer to use the native SDK caching than implement our own

PR #6701, opened 2025-08-08
test: Add VectorScanTableProvider snapshot tests

### LINKED ISSUE #6552: Snapshot testing for search UDTF tables
Both `vector_search` and `text_search` UDTFs use complex logic to derive a physical plan. Several optimisations are included to minimise projections, filters and joins where possible. This logic should be tested using snapshot testing to avoid regression in future refactors.

peasee:
## 📝 Summary

* Replicates the `VectorQueryTableProvider` snapshot tests, but for `VectorScanTableProvider`. The format of the queries change, as there is no score - but it instead snapshots against the generated embedding column, `body_embedding`

## 🔗 Related

* Closes #6552


crates/runtime/src/embeddings/index/scan_table.rs
```diff
@@ -306,3 +306,470 @@ impl TableProvider for VectorScanTableProvider {
...
+        sql::TableReference,
+    };
+    use search::generation::util::append_fields;
+    use snafu::ResultExt;
+
+    use crate::embeddings::index::{VectorIndex, VectorScanTableProvider};
+
+    /// This is just a [`MemTable`] that pretends it can support all filter pushdowns.
+    /// This is useful for testing explain plans.
+    #[derive(Debug)]
```
Jeadie: nit: Maybe its worth (if its even possible) having this in `crates/runtime/src/embeddings/index/mod.rs::tests`?

Jeadie (dismissed):

phillipleblanc (approved):

PR #6707, opened 2025-08-10
'stream_options.include_usage' issue for gpt5

Jeadie:


Jeadie:
Transient bug when GPT-5 series was initially released.

Issue #6711, opened 2025-08-11
S3Vectors fails to search with arrow conversion error


peasee:
## Describe the bug

A simple example of using S3 vectors to perform a search fails with a conversion error:

```console
Shared DataFusion error: Arrow error: Json error: whilst decoding field 'data': expected [ got nullReceived only partial JSON payload from QueryVectors
```

It appears that the S3 vectors API is returning `null` values for the `data` field, when we previously expected them to exist.

The AWS documentation defines this field as optional: https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_QueryOutputVector.html#AmazonS3-Type-S3VectorBuckets_QueryOutputVector-data

This matches the smithy models for the API.

However, in the S3 vectors documentation it states that:

> You can optionally include the computed distance (between the query vector and each vector in the response), the vector data [...]

This seems to suggest the `data` field can be optionally included, but there is no documented parameter.

## To Reproduce

_Steps to reproduce the behavior:_

1. Setup a spicepod with S3 vectors
2. Try to run a search
3. Observe error

## Expected behavior

S3 vectors should search successfully.

### Spicepod

```yaml
version: v1
kind: Spicepod
name: testing
datasets:
  - from: file:./test.csv
    name: test
    vectors:
      enabled: true
      engine: s3_vectors
      params:
        s3_vectors_bucket: peasee-indexes
        s3_vectors_aws_access_key_id: ${ secrets:AWS_ACCESS_KEY_ID }
        s3_vectors_aws_secret_access_key: ${ secrets:AWS_SECRET_ACCESS_KEY }
        s3_vectors_aws_region: us-west-2
        s3_vectors_aws_session_token: ${ secrets:AWS_SESSION_TOKEN }
    
    columns:
      - name: 'body'
        embeddings:
          - from: text_large
            row_id: 'title'
embeddings:
  - name: text_large
    from: openai:text-embedding-3-large
    params:
      openai_api_key: ${ secrets:SPICE_OPENAI_API_KEY }
```

## Additional context

I installed `awscurl` to run some simple curl checks, to see if there was an undocumented parameter. I generated a vector file and stored it in `vector.json` with the required dimension, then performed:

```bash
awscurl \
  --service s3vectors \
  --region us-west-2 \
  --profile default \
  -X POST "https://s3vectors.us-west-2.api.aws/QueryVectors" \
  -H "Content-Type: application/json" \
[...]

PR #6712, opened 2025-08-11
fix: Set S3 vectors default limit to 30, add more tracing

peasee:
## 📝 Summary

* The `topK` limit in the [AWS documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-vectors-limitations.html) is 30. When no `LIMIT` is defined, we supply `i32::MAX`. Updates to instead supply a default of `30`.
* Adds some additional debug tracing for Runtime debugging.

### DIFF
--- a/crates/data_components/src/s3_vectors/query_provider.rs
+++ b/crates/data_components/src/s3_vectors/query_provider.rs
@@ -286,8 +286,6 @@ async fn query_vector_stream(
 ) -> DataFusionResult<()> {
     let start = std::time::Instant::now();
 
-    tracing::debug!("QueryVector expected schema: {schema:?}");
-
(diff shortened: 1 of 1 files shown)


crates/data_components/src/s3_vectors/query_provider.rs
```diff
@@ -286,6 +286,8 @@ async fn query_vector_stream(
 ) -> DataFusionResult<()> {
     let start = std::time::Instant::now();
 
+    tracing::debug!("QueryVector expected schema: {schema:?}");
```
Jeadie: Are these going to valuable after #6711

phillipleblanc (approved):

PR #6714, opened 2025-08-11
feat: Avoid joining for VectorScanTableProvider if the index is sufficient

### LINKED ISSUE #6553: Optimization: Avoid joins in `VectorScanTableProvider`
Several optimisation can be added [`VectorScanTableProvider`](https://github.com/spiceai/spiceai/blob/84a57412da63842335a0ce92cb2fd3e5441d92c9/crates/runtime/src/embeddings/index/scan_table.rs#L45) to avoid additional work.
 - For columns in the projection, use the vector index if they're present.
 - If all columns in the projection, and required filters only reference columns in the vector index, do not `JOIN ON` the underlying table. 

These improvements are already in [`VectorQueryTableProvider`](https://github.com/spiceai/spiceai/blob/84a57412da63842335a0ce92cb2fd3e5441d92c9/crates/runtime/src/embeddings/index/query_table.rs#L55)

peasee:
## 📝 Summary

* Refactors some of the common functions from `VectorQueryTableProvider` in identifying if the underlying index is sufficient out into module level functions
* Adds support for `VectorScanTableProvider` to avoid joining if the index is sufficient to complete the query without joining

## 🔗 Related

* Closes #6553


crates/runtime/src/embeddings/index/snapshots/runtime__embeddings__index__tests__scan_table_basic.snap
```diff
@@ -2,19 +2,15 @@
...
-| plan_type     | plan                                                                                                           |
-+---------------+----------------------------------------------------------------------------------------------------------------+
-| logical_plan  | Sort: my_vectored_table.pk DESC NULLS FIRST, fetch=5                                                           |
-|               |   TableScan: my_vectored_table projection=[pk, body_embedding]                                                 |
-| physical_plan | SortPreservingMergeExec: [pk@0 DESC], fetch=5                                                                  |
-|               |   CoalesceBatchesExec: target_batch_size=8192, fetch=5                                                         |
-|               |     SortExec: TopK(fetch=5), expr=[pk@0 DESC], preserve_partitioning=[true]                                    |
-|               |       CoalesceBatchesExec: target_batch_size=8192                                                              |
-|               |         HashJoinExec: mode=CollectLeft, join_type=Left, on=[(pk@0, pk@1)], projection=[pk@0, body_embedding@1] |
```
peasee, line 14: HashJoin is removed from a query which can use the index, as expected

crates/runtime/src/embeddings/index/snapshots/runtime__embeddings__index__tests__scan_table_join_for_filter.snap
```diff
@@ -1,24 +1,26 @@
...
+|               |     Filter: my_vectored_table.another_column != Utf8("something")                                                                      |
+|               |       TableScan: my_vectored_table projection=[pk, another_column, body_embedding]                                                     |
+| physical_plan | SortPreservingMergeExec: [pk@0 DESC], fetch=5                                                                                          |
+|               |   SortExec: TopK(fetch=5), expr=[pk@0 DESC], preserve_partitioning=[true]                                                              |
+|               |     CoalesceBatchesExec: target_batch_size=8192                                                                                        |
+|               |       FilterExec: another_column@1 != something, projection=[pk@0, body_embedding@2]                                                   |
+|               |         ProjectionExec: expr=[pk@2 as pk, another_column@0 as another_column, body_embedding@1 as body_embedding]                      |
```
peasee, line 17: Joins are still retained when the index is not sufficient.

New projections in these plans originate from the updated underlying join generation (which matches the query table provider method).

crates/runtime/src/embeddings/index/snapshots/runtime__embeddings__index__tests__scan_table_join_for_projection.snap
```diff
@@ -1,20 +1,22 @@
 ---
 source: crates/runtime/src/embeddings/index/mod.rs
+assertion_line: 476
```
Jeadie, line 3: what's this snapshot for now?

Am i right in saying this was from the old part of this diff https://github.com/spiceai/spiceai/pull/6714/files#diff-82a945e65f7f9996abccad0124c5148de85fccae496e41c92c76448c8f033b81R501

Just don't want excess snapshots laying around.

peasee: I made a mistake when I created the snapshots originally, there were 2 tests writing to the same snapshot name. This snapshot is still relevant and being used by the non-metadata test.

Jeadie (approved):

Issue #6554, opened 2025-07-22
Enhancement: Multiple-Column Parameter Keys


Jeadie:
## Goal-State/What/Result

Vector indexes should support datasets that have > 1 column as the primary key.

## Why/Purpose
Many datasets have more than one column as the primary key. Specifically, many retrieval contexts introduce an additional column to uniquely identify sub-parts of a row (e.g. splitting a document into chunks).

## By When

**Issue/Spec written and reviewed:** Wednesday 13th August
**Done-Done:** Sunday 25thh August

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## Specification
Currently a [`arrow::compute::cast`](https://docs.rs/arrow/latest/arrow/compute/fn.cast_with_options.html) is used to convert primary keys to strings (S3 vector only has single-value string primary keys). We cannot simply use a `StructArray` and then convert to a `StringArray`, since `arrow::compute::cast` doesn't support to/from `StructArray`.

To reduce the complexity of primary keys in relation to indexes, the following changes will only be applied for datasets with more than one column. The current simpler method (`arrow::compute::cast` to string), will be used when there is a single column primary key.

A standard/protocol for de/serialization. Possibly either:
1. [`arrow-json`](https://docs.rs/arrow-json)
1. [`arrow-ipc`](https://docs.rs/arrow-ipc)

`arrow-json` is required to enable defining deserialization within logical plans (see [below](#index-reading)). Given the serialised format will be of a single row, and the schema is known and doesn't need to be inferred, `arrow-ipc` would likely be no more performant (i.e. serialised size, de/serialisation speed) than `arrow-json`.

**Fundamentally**, the specification for this enhancement is to use `arrow-json` format and [datafusion-functions-json](https://github.com/datafusion-contrib/datafusion-functions-json) to store multi-column primary keys in vector indexes that only support single string keys (whilst using a basic type-cast for single-column primary keys). 

### Index Writing
[...]

PR #6757, opened 2025-08-14
Make `VectorScanTableProvider` and `VectorQueryTableProvider` support multi-column primary keys

Jeadie:
## 📝 Summary
 - Makes `VectorScanTableProvider` and `VectorQueryTableProvider` support multi-column primary keys. Note: `S3Vector`, the only vector index available still does not support multi-column primary keys. That will come in a follow up PR. 
 - To achieve this, we make `VectorIndex` more generic by encapsulating the expected types/logic in [`VectorIndex::query_table_provider`] and [`VectorIndex::list_table_provider`]. 

## 🔗 Related
- Needed for #6554


crates/runtime/src/embeddings/index/mod.rs
```diff
@@ -90,18 +153,51 @@ impl VectorIndex for S3Vector {
...
+    ///   2. Rename [`S3_VECTOR_EMBEDDING_NAME`] appropriately
     fn list_table_provider(&self) -> Arc<dyn TableProvider> {
-        Arc::new(S3VectorsListTable::from(self.index.table.clone()))
+        let Some((pk_name, pk_data_type)) = self
+            .primary_fields()
+            .iter()
+            .next()
+            .map(|f| (f.name().clone(), f.data_type().clone()))
+        else {
+            panic!("mahhhhh")
```
phillipleblanc: fix

Jeadie: Sorry, I was meant to set this as draft.

crates/runtime/src/embeddings/index/mod.rs
```diff
@@ -78,6 +112,31 @@ impl S3Vector {
         let model = model_lock.get(&self.cfg.model_name)?;
         Some(Arc::clone(model))
     }
+
+    pub async fn query_vector(
```
phillipleblanc: query_vector feels a little weird as a name - its not actually querying anything? its just calculating the embedding.

crates/runtime/src/embeddings/index/mod.rs
```diff
@@ -78,6 +112,31 @@ impl S3Vector {
...
+            return Err(Box::from(format!(
+                "Vector index requires '{}' embedding model, but is not available.",
+                self.cfg.model_name
+            )));
+        };
+        let mut resp = embedding_model
+            .embed(EmbeddingInput::String(query.to_string()))
+            .await
+            .boxed()?;
+        let Some(query_vector) = resp.pop() else {
```
phillipleblanc: We should also have an error case that covers if there is more than one vector in the embedding (i.e. resp.pop().len() should be 0)

crates/runtime/src/embeddings/index/mod.rs
```diff
@@ -110,28 +206,51 @@ impl VectorIndex for S3Vector {
...
+                    Operator::Minus,
+                    Box::new(Expr::Column(Column::new_unqualified("distance"))),
+                )),
+                None::<TableReference>,
+                SEARCH_SCORE_COLUMN_NAME,
+            )),
+        ];
+        projection.extend(metadata_columns_to_exprs(self.metadata_columns()));
+
+        // TODO: Restructure [`S3VectorsQueryTable`] to take an async function (probably a trait)
```
phillipleblanc: Link to tracking issue.

crates/runtime/src/embeddings/index/scan_table.rs
```diff
@@ -263,25 +174,76 @@ impl TableProvider for VectorScanTableProvider {
...
                 .table_provider
                 .scan(state, projection, filters, limit)
                 .await;
-        };
+        }
+
+        // Do not use [`VectorIndex::embedded_column`] from index.
+        // [`VectorScanTableProvider`] is used to populate RecordBatch in
+        // [`runtime_datafusion_index::Index::compute_index`]. On initial indexing into the index,
+        // we would get NULL values from [`VectorIndex`], we must use underlying instead.
```
phillipleblanc, line 196: I'm planning to fix this very soon as part of #6518 - do we still need this logic then?

crates/runtime/src/embeddings/index/mod.rs
Jeadie: In my head `query_vector` is an adjective + noun (i.e. the query, in vector form), not a `verb + noun`. Now that its a standalone function, perhaps `verb + noun` is the default intuition.

crates/runtime/src/embeddings/index/scan_table.rs
Jeadie: No the issue here is separate. Imagine a dataset with a vector index like this
```yaml
 columns:
    - name: body
      embeddings:
        - from: my-model
      metadata:
        vectors: non-filterable
```

Consider a query like `SELECT body, score FROM vector_search(...`
 - If the vector index is populated and we use it to get `body`, we return data fine.
 -  If the vector index is not populated and we use it, than `body` will be Null for all rows. 

When we initial index a dataset into a vector index, we would be dealing with the latter case. This means no data would be indexed and written to the vector index.

One possible solution could be to use `COALESCE(vector_index.body, base_table.body)` but I don't think DF can figure out to avoid a scan of base_table if `vector_index.body` exists.

phillipleblanc (approved):

PR #6762, opened 2025-08-14
Pin cargo-deny to 0.18.3

kczimm:
## 📝 Summary

Licence check CI fails because the latest `cargo-deny`[ requires Rust 1.88.0](https://github.com/EmbarkStudios/cargo-deny/blame/main/Cargo.toml#L18).


kczimm:
Rust forbidden license check passed as expected.

Advayp (approved):

PR #6768, opened 2025-08-15
fix: Run enforce pulls with spice on pull_request_target

peasee:
## 📝 Summary

* Updates the event from `pull_request` to `pull_request_target`

### DIFF
--- a/.github/workflows/enforce-pulls-with-spice.yml
+++ b/.github/workflows/enforce-pulls-with-spice.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: spiceai-macos
     steps:
       - uses: spiceai/pulls-with-spice-action@v1.0.6
-        if: github.event_name == 'pull_request'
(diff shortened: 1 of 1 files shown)


.github/workflows/enforce-pulls-with-spice.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: spiceai-macos
     steps:
       - uses: spiceai/pulls-with-spice-action@v1.0.6
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request_target'
```
phillipleblanc, line 19: should it be both?

peasee: the workflow only runs on `pull_request_target`, I don't see why it would need to be both.

phillipleblanc (approved):

Issue #6743, opened 2025-08-13
Enhancement: Add OpenAI embeddings and models request concurrency control


peasee:
## Goal-State/What/Result

The OpenAI Model and Embedding connector supports controlling the maximum number of concurrent embeddings requests.

The Bedrock Model and Embedding connector rate limiting configuration is updated to use a shared crate.

## Why/Purpose

To improve the performance of local acceleration embeddings, and to improve the reliability of embedding requests.

## By When

**Issue/Spec written and reviewed:** 14th Aug, 2025
**Done-Done:** 15th Aug, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

A component parameter `openai_usage_tier` is introduced to the OpenAI Embeddings connector:

```yaml
embeddings:
  - from: openai
    params:
      openai_usage_tier: tier1 | tier2 | tier3 | tier4 | tier5 | free
```

Setting a usage tier updates the maximum concurrent requests

The usage tier parameter defaults to `tier1`.

### Maximum concurrent requests

For each tier, the maximum concurrent requests are:

* `free`: 1
* `tier1`: 35
* `tier2`, `tier3`: 60
* `tier4`, `tier5`: 125

These parameters are set based on the tier's requests per minute limit for [`text-embedding-3-small`](https://platform.openai.com/docs/models/text-embedding-3-small). With the assumption that most embedding requests take around 1 second to complete, for example setting the default to 35 for tier 1 ensures that users do not exceed the rate limits in a default configuration.

### Max Batch Size & Batch Size Bytes

**Specific for OpenAI.**

Batch size is controlled with a constant value set to `2048` (rows). To artificially enforce more concurrent requests, this batch size should be reduced to 256 (rows).

[...]

PR #6773, opened 2025-08-15
feat: Migrate bedrock rate limiter

### LINKED ISSUE #6750: Migrate bedrock config to runtime rate limiter

peasee:
## 📝 Summary

* Migrates the bedrock rate limiter to using a `RateController` instead
* Adds some more logic to a `Permit` from a `RateController`, which required refactoring slightly how permits were issued with semaphore. This logic lets an existing permit re-check against the quotas while retaining their in-progress request (for example, re-checking quotas before retying a failed request).
* Fixes the concurrency for bedrock embeddings. The comment stated up to 5 requests were sent in parallel, but it looked like it was only ever 1 (sequential)? Now, it is up to the rate limiter to decide concurrency. This results in the following performance improvement:

**Before:**
```console
2025-08-15T08:04:17.955706Z  INFO runtime::accelerated_table::refresh_task: Loaded 1,000 rows (1.03 MiB) for dataset customers in 4m 21s 862ms.
```

**After:**
```console
2025-08-15T07:58:53.625147Z  INFO runtime::accelerated_table::refresh_task: Loaded 1,000 rows (1.03 MiB) for dataset customers in 20s 299ms.
```

## 🔗 Related

* Closes #6750 
* Part of #6743


crates/runtime-rate-control/src/lib.rs
```diff
@@ -115,29 +115,50 @@ pub struct RateController {
 
 #[derive(Debug)]
 #[allow(dead_code)] // The inner permit is never used, but it needs to be retained for the lifetime of the permit so it can be correctly dropped.
```
phillipleblanc: This is still needed?

crates/runtime-rate-control/src/lib.rs
```diff
@@ -115,29 +115,50 @@ pub struct RateController {
...
+pub struct Permit {
+    permit: Option<OwnedSemaphorePermit>,
+    rate_controller: Arc<RateController>,
+}
+
+impl Permit {
+    /// Re-check the quotas from an existing permit.
+    /// For example, a request was permitted but has failed and needs to be retried.
+    /// The caller retains their permit, but needs to ensure the rate limiters are still ready.
+    pub async fn until_ready(&self) {
```
phillipleblanc, line 134: Can we add a test for this condition

crates/runtime-rate-control/src/lib.rs
peasee: yeah. the comment on the dead code is still valid, and probably will be valid forever.

I could make it `_permit` and remove the `dead_code`, but then the test code complains that I'm accessing an underscored property :disappointed: either way, a lint rule will need to be allowed.

peasee: actually, I can get around it by manually implementing `Drop`. That way, the field is used during the drop instead of letting it auto-implement `Drop`.

phillipleblanc (approved):

PR #6778, opened 2025-08-18
fix: Move search telemetry calls in UDTF to scan

peasee:
## 📝 Summary

* Moves the telemetry tracking calls for search into their UDTF table provider `.scan()` functions, instead of in the sync `TableFunctionImpl::call` which dropped the request context in a sync scope.

### DIFF
--- a/crates/runtime/src/search/full_text/udtf.rs
+++ b/crates/runtime/src/search/full_text/udtf.rs
@@ -27,13 +27,17 @@ limitations under the License.
 //! The schema of the resultant table will be: `schema(tbl) ∪ {score}`, where:
 //!  - `score` (f32): The similarity score of the row with the request `query`.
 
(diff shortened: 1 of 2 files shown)


crates/runtime/src/search/full_text/udtf.rs
```diff
@@ -223,12 +218,47 @@ impl TableFunctionImpl for TextSearchTableFunc {
...
         }))
     }
 }
+
+#[derive(Debug)]
+struct TextSearchIndexProviderWrapper {
+    inner: Arc<TextSearchIndexProvider>,
+}
+
+#[async_trait::async_trait]
```
phillipleblanc, line 242: You will want to implement all optional methods as well

phillipleblanc (approved):

PR #6775, opened 2025-08-16
Support multiple column primary keys for S3 vectors.

Jeadie:
## 📝 Summary
 - Enable multi-column primary keys for S3 vectors.
 - Behvaviour for single-column unchanged. 

### Example
For a multi-column primary key (note, primary key types are Utf8 and Int64)
```yaml
    - name: '_source.body'
      embeddings:
        - from: bedrock_titan
          row_id:
            - '_id'
            - '_version'
```
Decoded into appropriate columns
```shell
sql> select _id, _version, score  from vector_search(mdr_data, 'hello') order by score desc  limit 4;
+------------------------------------+----------+--------------------+
| _id                                | _version | score              |
+------------------------------------+----------+--------------------+
| SMc7062591bac245ecb4fffe4446f76a2c | 4        | 0.7346948981285095 |
| SM3fe6d2e8c49a46ea8d0e52c8f0db6219 | 2        | 0.7288013696670532 |
+------------------------------------+----------+--------------------+
```
With explain plan as expected
```shell
sql> explain select _id, _version, score  from vector_search(mdr_data, 'hello') order by score desc  limit 4;
+---------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| plan_type     | plan                                                                                                                                                               |
+---------------+--------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| logical_plan  | Sort: vector_search().score DESC NULLS FIRST, fetch=4                                                                                                              |
|               |   BytesProcessedNode                                                                                                                                               |
|               |     TableScan: vector_search() projection=[_id, _version, score]                                                                                                   |
| physical_plan | SortPreservingMergeExec: [score@2 DESC], fetch=4                                                                                                                   |
|               |   SortExec: TopK(fetch=4), expr=[score@2 DESC], preserve_partitioning=[true]                                                                                       |
|               |     BytesProcessedExec                                                                                                                                             |
|               |       BytesProcessedExec                                                                                                                                           |
|               |         ProjectionExec: expr=[union_extract(json_get(key@0, _id), str) as _id, union_extract(json_get(key@0, _version), int) as _version, 1 - distance@1 as score] |
|               |           RepartitionExec: partitioning=RoundRobinBatch(10), input_partitions=1                                                                                    |
|               |             BytesProcessedExec                                                                                                                                     |
|               |               S3VectorsQueryExec: limit=100                                                                                                                        |
|               |                                                                                                                                                                    |
[...]

PR #6782, opened 2025-08-18
Handle errors in vector ingestion pipeline

### LINKED ISSUE #6509: If indexing returns an error, the acceleration attempt is aborted
Currently if indexing fails, the acceleration will continue leading to a situation where we can't be sure the indexing has completely properly when the acceleration finishes.

Wire up errors such that if an indexing job fails, it causes the acceleration load to fail as well.

phillipleblanc:
## 📝 Summary

If the vectors ingestion pipeline has an error (i.e. error computing the embedding, or an error writing to the vector store) then the acceleration attempt should also return an error, which will prompt a retry at the acceleration layer.

This PR also modifies the signature of `compute_index` to return a `Vec<RecordBatch>` in preparation for #6518

## 🔗 Related

- Closes #6509


crates/runtime-datafusion-index/src/lib.rs
```diff
@@ -30,7 +31,11 @@ pub trait Index: Debug + Send + Sync + 'static {
     /// Columns that are required for the index to be computed.
     fn required_columns(&self) -> Vec<String>;
 
-    async fn compute_index(&self, _batches: Vec<RecordBatch>) {}
+    /// Compute the index - if the index data is represented in the batch itself (i.e. a vector
+    /// "*_embedding" column) then modify the provided batches to include the computed column.
+    async fn compute_index(&self, batches: Vec<RecordBatch>) -> Result<Vec<RecordBatch>> {
```
phillipleblanc, line 36: This currently isn't being consumed anywhere - it will be in the next PR that addresses #6518

crates/runtime/src/embeddings/index/s3.rs
```diff
@@ -184,53 +221,57 @@ pub async fn write(index: &S3VectorIndex, cfg: &IndexEmbeddingConfig, record: &R
...
+            len: values.len(),
+        }
+        .fail();
     }
 
     // Currently, we only support when there is one primary key column.
     if primary_keys.len() > 1 {
         tracing::debug!("primary_keys.len() > 1");
-        return;
+        return Ok(());
```
sgrebnov, line 246: Should invalid/unsupported primary keys configurations treated as errors as well?

sgrebnov (approved):

crates/runtime/src/embeddings/index/s3.rs
phillipleblanc: Jack is working on adding multiple primary keys in https://github.com/spiceai/spiceai/pull/6775

PR #6792, opened 2025-08-18
Pass embeddings calculated in `compute_index` to the acceleration

### LINKED ISSUE #6351: Integration tests for indexing analyzer rule

### LINKED ISSUE #6518: Embeddings that are generated are always added to acceleration

phillipleblanc:
## 📝 Summary

Embeddings that are calculated in the `compute_index` function will store the calculated embeddings in the `*_embedding` column if it exists.

It will skip calculating embeddings for rows that already have an embedding value from the upstream provider.

## 🔗 Related

- Closes #6518
- Closes #6351


crates/runtime/src/embeddings/index/s3.rs
```diff
@@ -271,7 +291,72 @@ pub async fn write(
...
+        let mut columns: Vec<Arc<dyn Array>> = (0..record.num_columns())
+            .map(|i| Arc::clone(record.column(i)))
+            .collect();
+
+        // Create new embedding array
+        let embedding_array = create_embedding_array(embedding_vectors)?;
+        columns[idx] = embedding_array;
+
+        // Create new RecordBatch with updated column
+        RecordBatch::try_new(schema, columns).context(CannotUpdateEmbeddingColumnSnafu)
```
peasee, line 322: will `try_new` with a vec of arc columns cause a copy? I'm guessing because they're in an arc it won't, but not positive.

peasee (approved):

crates/runtime/src/embeddings/index/s3.rs
phillipleblanc: It will clone the Arc, so we'll have multiple references to the same column data - but if the previous RecordBatch is dropped, then those references will drop as well, so it behaves how we want.

Jeadie (approved):

Issue #6759, opened 2025-08-14
Implement OpenAI Responses API endpoint


kczimm:
- Create new API endpoint at `/v1/responses`
- Implement the handler for both streaming and non-streaming responses.
- Add route to handler
- Implement `Responses` trait for OpenAI model provider

kczimm:
Streaming support for Responses API is WIP in [async-openai](https://github.com/64bit/async-openai/pull/405) crate.

PR #6798, opened 2025-08-18
Introduce a `Responses` trait and LLM registry for model providers that support the OpenAI Responses API

### LINKED ISSUE #6758: Generalize AI model capabilities
- Introduce `Model` and `Responses` traits
- Refactor `LLMModelStore` to store `Arc<dyn Model>`
- Update existing chat endpoint to use the new `model.as_chat()` method and update existing model providers to implement `Model`

### LINKED ISSUE #6828: Introduce and Populate a registry for models that support the Responses API
- [ ] Create a `Responses` trait
- [ ] Create a registry that will store `Arc<dyn Responses` and populate it when initializing chat models.

Advayp:
## 📝 Summary
This PR:
- Introduces the `Responses` trait, which model providers that support the OpenAI Responses API will need to implement
- Implements this trait for the OpenAI model provider
- Creates a `Responses` registry for storing models with providers that support the OpenAI Responses API. This registry will be used in the `/v1/responses` endpoint to determine whether or not a given model is supported.

## 🔗 Related
- Closes #6828 
- Related to #6759


crates/llms/src/model.rs
```diff
@@ -0,0 +1,41 @@
...
+    #[snafu(display("The model does not support chat completions"))]
+    NoChatSupport,
+
+    #[snafu(display("Failure within model provider: {source}"))]
+    ModelProvider {
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
+}
+
+pub trait Model: Send + Sync {
```
phillipleblanc: I'm not sure why this is required - in what situations do we not know which model type we want from the beginning? i.e. why not just have separate registries for chat and responses models?

Advayp: I'm not exactly sure why @kczimm had wanted it done this way, but my understanding is that it prevents a good amount of the duplication that the chat/responses registries would have.

Also, some of the usages of `Chat` changed to `Model` would I think need a way to support both the `Chat` and `Responses` traits, which the `Model` trait lets us do.

phillipleblanc: I'm still struggling to understand why we would need to support either Chat or Responses - I expect we would know up front which one we need. Is there code that you can point to that would do some fallback if Chat isn't supported or Responses is (or vice versa)? If not then I don't think we should do this.

kczimm: It's not required, it was just a design choice. This seemed fitting conceptually. We have the concept of models in the spicepod. We hold a collection of models via the `LLMModelStore`. They have specific capabilities that this expresses. If you're working with models, the concepts of Chat and Responses shouldn't be foreign. For future code development, adding the Responses functionality to a model provider just involves providing a custom implementation of the trait. However, if you have a separate registry, you would need another piece of logic that knows to add the model to the additional registry. I personally believe it is a larger mental lift to look at two (or more with future APIs) registries, `chat_models: Vec<Arc<dyn Chat>>` and `responses_models: Vec<Arc<dyn Responses>>` and realize that your single model could be in both (or more) registries.

But you're right, it's not required. It just made more sense to me. Both approaches ultimately function equivalently.

kczimm: > why we would need to support either Chat or Responses 

I don't think this trait indicates that a model would support either/or, rather, can support none, some, or all.

crates/runtime/src/model/responses.rs
Advayp, line 1: I chose to duplicate this from `crates/runtime/src/model/chat.rs` because the enhancement scopes support for the Responses API to only the OpenAI model provider. In the future, as we integrate with other providers' support for the API, we can refactor this.

Advayp, line 1: I chose to duplicate this from crates/runtime/src/model/chat.rs because the enhancement scopes support for the Responses API to only the OpenAI model provider. In the future, as we integrate with other providers' support for the API, we can refactor this.

crates/llms/src/responses.rs
```diff
@@ -0,0 +1,93 @@
...
+use snafu::prelude::*;
+use tracing_futures::Instrument;
+
+use crate::chat::nsql::SqlGeneration;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub))]
+pub enum Error {
+    #[snafu(display(
+        "Failed to load the model.\nAn error occurred: {source}\nReport a bug on GitHub: https://github.com/spiceai/spiceai/issues"
```
phillipleblanc:
```suggestion
        "Failed to load the model: {source}"
```

crates/llms/src/responses.rs
```diff
@@ -0,0 +1,93 @@
...
+    ))]
+    FailedToLoadModel {
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
+    #[snafu(display(
+        "Failed to run the model.\nAn error occurred: {source}\nReport a bug on GitHub: https://github.com/spiceai/spiceai/issues"
```
phillipleblanc:
```suggestion
        "Failed to invoke the model: {source}"
```

crates/llms/src/responses.rs
```diff
@@ -0,0 +1,93 @@
...
+    ))]
+    FailedToRunModel {
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
+    #[snafu(display(
+        "Failed to check the status of the model. An error occurred: {source}. Verify the model configuration and try again."
```
phillipleblanc:
```suggestion
        "Failed to invoke the model: {source}. Verify the model configuration and try again."
```

crates/llms/src/responses.rs
```diff
@@ -0,0 +1,93 @@
...
+    async fn health(&self) -> Result<()>;
+    async fn stream<'a>(
+        &self,
+        prompt: String,
+    ) -> Result<Pin<Box<dyn Stream<Item = Result<Option<String>>> + Send>>> {
+        let resp = self.run(prompt).await;
+        Ok(Box::pin(stream! { yield resp }))
+    }
+    async fn responses_stream(&self, req: CreateResponse) -> Result<ResponseStream, OpenAIError>;
+    async fn responses_request(&self, req: CreateResponse) -> Result<Response, OpenAIError>;
```
phillipleblanc: This should be our own error type, not OpenAIError

crates/llms/src/responses.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    async fn health(&self) -> Result<()>;
+    async fn stream<'a>(
+        &self,
+        prompt: String,
+    ) -> Result<Pin<Box<dyn Stream<Item = Result<Option<String>>> + Send>>> {
+        let resp = self.run(prompt).await;
+        Ok(Box::pin(stream! { yield resp }))
+    }
```
phillipleblanc:
```suggestion
   fn stream(self: Arc<Self>, prompt: impl Into<String>)
    -> Result<Pin<Box<dyn Stream<Item = Result<Option<String>>> + Send>>> {
    let prompt = prompt.into();
    Box::pin(stream::once(async move { self.run(prompt).await }))
}
```

1. The function that returns the stream doesn't need to be async if you change `self` to be `Arc<Self>` then you can move it into the async closure without lifetime issues.
2. `stream::once` is more idiomatic than using `stream!`
3. The stream function itself doesn't need to return an error, the error can be returned in the stream itself (i.e. Item = Result<..>)

crates/runtime/src/init/llm.rs
```diff
@@ -17,30 +17,72 @@ limitations under the License.
...
+};
 use secrecy::SecretString;
+use serde_json::Value;
 use snafu::ResultExt;
-use spicepod::component::model::Model as SpicepodModel;
+use spicepod::component::model::{Model as SpicepodModel, ModelSource};
+
+static DEFAULT_OPENAI_ENDPOINT: &str = "https://api.openai.com/v1";
+
+fn supports_responses_api(spicepod_model: &SpicepodModel) -> bool {
```
phillipleblanc: Instead of hardcoding this (which isn't flexible) - let's just try to make a health request and see if it works. That way once more providers start to support this API we won't need to release a new version of Spice

crates/runtime/src/init/llm.rs
```diff
@@ -17,30 +17,72 @@ limitations under the License.
...
+}
 
 impl Runtime {
     /// Loads a specific LLM from the spicepod. If an error occurs, no retry attempt is made.
     pub(crate) async fn load_llm(
         &self,
         m: SpicepodModel,
         params: HashMap<String, SecretString>,
-    ) -> Result<Arc<dyn Chat>> {
+    ) -> Result<(Arc<dyn Chat>, Option<Arc<dyn Responses>>)> {
```
phillipleblanc: Are there model providers that would support only the responses endpoint and not the chat completion one? I would make both of them optional (and return an error if both are None) and handle that in the calling code

crates/runtime/src/model/responses.rs
```diff
@@ -0,0 +1,126 @@
...
+    params: &Parameters,
+) -> Result<Arc<dyn Responses>, LlmError> {
+    let model_id = component.get_model_id();
+    let prefix = component.get_source().ok_or(LlmError::UnknownModelSource {
+        from: component.from.clone(),
+    })?;
+
+    let model = match prefix {
+        ModelSource::OpenAi => openai(model_id, params),
+        _ => Err(LlmError::ResponsesNotSupported {
```
phillipleblanc, line 75: Again, let's not impose this restriction in code - let's just discover whether its supported at runtime.

crates/runtime/src/lib.rs
```diff
@@ -417,7 +418,9 @@ pub struct Runtime {
     app: Arc<RwLock<Option<Arc<App>>>>,
     df: Arc<DataFusion>,
     models: Arc<RwLock<HashMap<String, Model>>>,
-    llms: Arc<RwLock<LLMModelStore>>,
+    chat_llms: Arc<RwLock<LLMModelStore>>,
```
```suggestion
    chat_llms: Arc<RwLock<LLMChatCompletionsModelStore>>,
```

crates/runtime/src/init/llm.rs
Jeadie: Currently no, but still logically more apt

crates/llms/src/chat/mod.rs
```diff
@@ -162,6 +162,11 @@ pub enum Error {
         model: String,
         source: Box<dyn std::error::Error + Send + Sync>,
     },
+
+    #[snafu(display(
+        "Model '{from}' does not support the OpenAI Responses API. Change the model provider to 'openai' to use the Responses API or use the Chat Completions API."
+    ))]
+    ResponsesNotSupported { from: String },
```
Jeadie: Does this work?
```suggestion
    ResponsesNotSupported { from: ModelSource },
```

crates/runtime/src/builder.rs
```diff
@@ -242,7 +242,8 @@ impl RuntimeBuilder {
             app: Arc::new(RwLock::new(self.app)),
             df,
             models: Arc::new(RwLock::new(HashMap::new())),
-            llms: Arc::new(RwLock::new(HashMap::new())),
+            chat_llms: Arc::new(RwLock::new(HashMap::new())),
```
Jeadie: I think `completions` not `chat` is the more canoncial phrase (i.e. we should use `completion_llms`). For example response APIs are still used in chats.

Jeadie (commented):
Possibly out of scope for this PR (and possibly #6759 ), but there is the following for completions which we may need/want an equivalent of:
1. `impl Chat for ChatWrapper` ([source](https://github.com/spiceai/spiceai/blob/6246b0d3cd09c1ad234baaec2a6648e6b5bbbe98/crates/runtime/src/model/wrapper/mod.rs#L276)) which provides additional utilty for each completion model (metrics, default overrides, tracing)
2. `impl Chat for ToolUsingChat `([source](https://github.com/spiceai/spiceai/blob/6246b0d3cd09c1ad234baaec2a6648e6b5bbbe98/crates/runtime/src/model/tool_use.rs#L393)) which provides internal tool use for tools callable from within the spice runtime, local or MCP (i.e. user -> LLM -> request spice tool called -> spice runs tool -> LLM -> response back to use)

crates/llms/src/responses.rs
Advayp: This runs into the lifetime issue of `Self` not living long enough

crates/runtime/src/model/responses.rs
phillipleblanc: Ok, this can stay

crates/llms/src/responses.rs
```diff
@@ -31,24 +26,34 @@ use crate::chat::nsql::SqlGeneration;
...
+    #[snafu(display("Failed to retrieve stream response from OpenAI: {source}."))]
+    StreamOpenAIError {
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
+    #[snafu(display("Failed to retrieve response from OpenAI: {source}."))]
+    ResponseOpenAIError {
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
+    #[snafu(display(
+        "An internal error occurred. Report a bug at https://github.com/spiceai/spiceai/issues."
```
phillipleblanc: I think we should still include the inner error details in the error message, otherwise there isn't much to go on for a bug report.

crates/llms/src/openai/responses.rs
```diff
@@ -74,9 +84,17 @@ impl<C: Config + Send + Sync + Clone> Responses for Openai<C> {
             .rate_controller
             .acquire()
             .await
-            .map_err(|e| OpenAIError::StreamError(e.to_string()))?;
+            .map_err(|e| OpenAIError::StreamError(e.to_string()))
```
phillipleblanc: Why is this map_err required?

crates/llms/src/responses.rs
```diff
@@ -31,24 +26,34 @@ use crate::chat::nsql::SqlGeneration;
...
     },
     #[snafu(display(
-        "Failed to check the status of the model. An error occurred: {source}. Verify the model configuration and try again."
+        "Failed to invoke the model: {source}. Verify the model configuration and try again."
     ))]
     HealthCheckError {
         source: Box<dyn std::error::Error + Send + Sync>,
     },
+    #[snafu(display("Failed to retrieve stream response from OpenAI: {source}."))]
+    StreamOpenAIError {
```
phillipleblanc: This is a generic error struct - having specific errors for concrete implementations is a code smell

phillipleblanc (approved):

Jeadie (approved):

PR #6811, opened 2025-08-19
fix: Allow v1/search with only FTS indexes

peasee:
## 📝 Summary

* Updates the vector search struct to allow performing a search when only columns with full text indexes are defined (no embeddings).
* Adds an integration test to validate as such. It would previously work if you supplied the dataset names, but now it works without having to specify a dataset name!

## 👀 Notes for Reviewers

* The description of the `v1/search` route probably needs updating, given we (for some time) now search across both FTS and Vector search

### DIFF
--- a/crates/runtime/src/http/v1/search.rs
+++ b/crates/runtime/src/http/v1/search.rs
@@ -186,7 +186,7 @@ pub(crate) async fn post(
         },
         Err(e) => {
             let error_type = match e {
-                VectorSearchError::NoTablesWithEmbeddingsFound {}
+                VectorSearchError::NoTablesWithSearchFound {}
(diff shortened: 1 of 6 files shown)


phillipleblanc (commented):
It is a code smell that the full text search relies on the vector search module to work - do we have work tracking to decouple these?

peasee:
> It is a code smell that the full text search relies on the vector search module to work - do we have work tracking to decouple these?

It was my understanding that it would be renamed/replaced as part of https://github.com/spiceai/spiceai/issues/6243. It is `VectorSearch` mostly only in name now. E.g. a `search()` UDTF would invoke the search pipeline directly, instead of via `VectorSearch`.

phillipleblanc (approved):

PR #6830, opened 2025-08-19
Add non-streaming and streaming support for OpenAI Responses API endpoint

### LINKED ISSUE #6759: Implement OpenAI Responses API endpoint

Advayp:
## 📝 Summary
- This PR adds support for a non-streaming and streaming endpoint compatible with the OpenAI Responses API

## 🔗 Related
- Closes #6759


crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,193 @@
...
+    );
+    span.in_scope(|| tracing::info!(target: "task_history", model = %req.model, "labels"));
+
+    override_task_history_with_traceparent(&span.clone(), &headers);
+
+    let span_clone = span.clone();
+    async move {
+        let model_id = req.model.clone();
+        let stream = req.stream.unwrap_or(false);
+        match llms.read().await.get(&model_id) {
```
kczimm: This may not practically be an issue if we never write new models to the collection after we start but, in general, this holds the lock during the entire match arm which does the model inferencing. I would say it would be better practice to clone the reference counted model out of the collection first so that the lock scope does not include the inferencing. Something like:

```rust
let Some(model) = llms.read().await.get(&model_id).cloned() else {
    return (StatusCode::NOT_FOUND, format!("model '{model_id}' not found")).into_response();
};

// read lock is now out of scope and compiler can drop it
```

crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,193 @@
...
+use std::convert::Infallible;
+use std::time::Duration;
+use tokio::sync::RwLock;
+use tracing::{Instrument, Span};
+
+use crate::model::LLMResponsesModelStore;
+use llms::responses::Responses;
+fn extract_text(resp: &OpenAIResponse) -> Option<String> {
+    resp.output
+        .first()
```
kczimm: Does `Response.output` only ever have one [OutputContent](https://docs.rs/async-openai/latest/async_openai/types/responses/enum.OutputContent.html)?

crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,193 @@
...
+                }
+            }
+            None => (StatusCode::NOT_FOUND, format!("model '{model_id}' not found")).into_response(),
+        }
+    }
+    .instrument(span)
+    .await
+}
+
+// Copied from chat.rs for error event formatting
```
kczimm: Why copy? Can we reuse by moving it to a common location? What makes sense to do, if you can imagine OpenAI adding new APIs in the future? They had completions and deprecated it in favor of chat. Now they added responses. There will be more.

crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,193 @@
...
+        };
+        tracing::info!(target: "task_history", parent: &span, captured_output = %captured_output);
+        if let Some(id) = id {
+            tracing::info!(target: "task_history", parent: &span, id = %id, "labels");
+        }
+        drop(span);
+    };
+
+    Sse::new(Box::pin(sse_stream))
+        .keep_alive(KeepAlive::new().interval(Duration::from_secs(30)))
```
kczimm: Do we have the same keep alive in chat? Can we put this magic number in a `const` somewhere? A common place if used in chat also would be nice.

crates/runtime/src/http/v1/responses.rs
Advayp: Based on our current usage, I believe so.

kczimm (dismissed):
LGTM

crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,259 @@
...
+            } else {
+                None
+            }
+        })
+        .join("\n")
+}
+
+#[cfg_attr(feature = "openapi", utoipa::path(
+    post,
+    path = "/v1/chat/completions",
```
phillipleblanc: This path needs to be updated - please also review the rest of this API documentation to see if it should be updated - it looks like this was just copied from the chat completions API.

crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,259 @@
...
+    headers: HeaderMap,
+    Json(req): Json<CreateResponse>,
+) -> Response {
+    let span = tracing::span!(
+        target: "task_history",
+        tracing::Level::INFO,
+        "ai_chat",
+        input = %serde_json::to_string(&req).unwrap_or_default()
+    );
+    span.in_scope(|| tracing::info!(target: "task_history", model = %req.model, "labels"));
```
phillipleblanc: Can we also add that this is the responses API to the labels?

crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,259 @@
...
+                    if !message.is_empty() {
+                        tracing::info!(target: "task_history", parent: &span_clone, captured_output = %message);
+                    }
+                    tracing::info!(target: "task_history", parent: &span_clone,  id = %response.id, "labels");
+
+                    Json(response).into_response()
+                }
+                Err(e) => {
+                    tracing::error!(target: "task_history", parent: &span_clone, "{e}");
+                    tracing::error!("Error from v1/responses: {e}");
```
phillipleblanc: We shouldn't allow API calls to pollute the error message log in Spice - that would allow an attacker to spam the logs with invalid requests to potentially hide other activity.

crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,259 @@
...
+async fn create_response_sse_response(
+    model: Arc<dyn Responses>,
+    req: CreateResponse,
+    span: Span,
+) -> Response {
+    let mut strm: ResponseStream = match model.responses_stream(req).await {
+        Ok(stream) => stream,
+        Err(e) => {
+            tracing::error!(target: "task_history", parent: &span, "{e}");
+            tracing::error!("Error from v1/responses: {e}");
```
phillipleblanc:
```suggestion
```

crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,259 @@
...
+            return openai_error_to_response(e);
+        }
+    };
+
+    let sse_stream = stream! {
```
phillipleblanc: Rework this to not use the `stream!` macro - I think its fine for small amounts of code to use the `stream!` macro, but it breaks rust-analyzer from being able to provide useful hints while typing, and so I would avoid for large code blocks.

crates/runtime/tests/models/openai.rs
```diff
@@ -483,6 +488,153 @@ async fn openai_test_chat_messages() -> Result<(), anyhow::Error> {
...
+            let openai_client = OpenAIClient::with_config(openai_config);
+            let request = CreateResponseArgs::default()
+                .model("openai_model")
+                .input("Copy exactly what I say: The quick brown fox jumps over the lazy dog")
+                .build()?;
+
+            let response = openai_client.responses().create(request).await?;
+            let text = extract_text(&response);
+            assert!(
+                text.is_some_and(|t| t.contains("The quick brown fox jumps over the lazy dog")),
```
phillipleblanc: How reliable is this test going to be? We've seen issues with having models follow directions, so we should really be asserting that our code is correct and passing/receiving the correct information - not that the model is doing what we tell it to.

crates/runtime/tests/models/openai.rs
```diff
@@ -483,6 +488,153 @@ async fn openai_test_chat_messages() -> Result<(), anyhow::Error> {
...
+                    Err(e) => {
+                        eprintln!("{e:#?}");
+                        // When a stream ends, it returns Err(OpenAIError::StreamError("Stream ended"))
+                        // Without this, the stream will never end
+                        break;
+                    }
+                }
+            }
+
+            assert!(final_response.contains("The quick brown fox jumps over the lazy dog"));
```
phillipleblanc: Same feedback as the other test - let's assert that our code worked correctly rather than the model worked correctly.

phillipleblanc (dismissed):

crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,271 @@
...
+    let sse_stream = {
+        let span_clone = span.clone();
+        futures::stream::unfold(
+            (strm, String::new(), None::<u64>, span_clone),
+            move |(mut strm, mut captured_output, mut id, span)| async move {
+                if let Some(msg) = strm.next().instrument(span.clone()).await {
+                    match msg {
+                        Ok(response_event) => {
+                            let should_break = match &response_event {
+                                ResponseEvent::ResponseOutputTextDelta(delta) => {
```
Jeadie, line 208: nit: You can destructure these further if you like. e.g. 
```rust
  ResponseEvent::ResponseOutputTextDelta(ResponseOutputTextDelta { delta, .. }) => {
      captured_output.push_str(&delta);
      false
  }
  ResponseEvent::ResponseCompleted(ResponseCompleted {
      sequence_number, ..
  }) => {
      if id.is_none() {
          id = Some(sequence_number);
      }
      true
  }
```

crates/runtime/src/http/v1/responses.rs
```diff
@@ -0,0 +1,271 @@
...
+                                    captured_output.push_str(&delta.delta);
+                                    false
+                                }
+                                ResponseEvent::ResponseCompleted(resp) => {
```
Jeadie: One benefit of the above is you can join some cases
```rust
ResponseEvent::ResponseCompleted(ResponseCompleted {
    sequence_number, ..
})
| ResponseEvent::ResponseIncomplete(ResponseIncomplete {
    sequence_number, ..
}) => {
    if id.is_none() {
        id = Some(sequence_number);
    }
    true
}
```

crates/runtime/tests/models/openai.rs
```diff
@@ -483,6 +488,158 @@ async fn openai_test_chat_messages() -> Result<(), anyhow::Error> {
         .await
 }
 
+fn extract_text(response: &OpenAIResponse) -> Option<String> {
+    response
+        .output
+        .first()
+        .and_then(|out| {
```
Jeadie, line 495: You can use `?` in functions that return `Option<T>` to immediately return None (similar to how it is used for `Result<T, E>`). That + destructuring
```rust 
fn extract_text(response: &async_openai::types::responses::Response) -> Option<String> {
    let OutputContent::Message(OutputMessage { content, .. }) = response.output.first()? else {
        return None;
    };
    match content.first()? {
        Content::OutputText(output_text) => Some(output_text.text.clone()),
        Content::Refusal(_) => None,
    }
}
```

Jeadie (approved):

PR #6836, opened 2025-08-20
fix: Include FTS search column if it is the PK

### LINKED ISSUE #6784: Full text search fails using the same column as the row ID
## Describe the bug

I have a dataset of files from GitHub that I want to search against the `path`. The `path` field for file entries is the primary key, so is also defined as the `row_id` like:

```yaml
- name: path
        full_text_search:
          enabled: true
          row_id: path
```

Trying to text search with this configuration results in an error:

```
sql> explain select name, path, score from text_search(spiceai.docs, 'duckdb') order by score desc;
+---------------------+----------------------------------------------------------------------+
| plan_type           | plan                                                                 |
+---------------------+----------------------------------------------------------------------+
| logical_plan        | Sort: text_search().score DESC NULLS FIRST                           |
|                     |   BytesProcessedNode                                                 |
|                     |     TableScan: text_search() projection=[name, path, score]          |
| physical_plan_error | type_coercion                                                        |
|                     | caused by                                                            |
|                     | Schema error: No field named path. Valid fields are base_table.name. |
+---------------------+----------------------------------------------------------------------+
```

## To Reproduce

2. Click on '...'
3. Scroll down to '...'

## Expected behavior

### Spicepod

### Output of `explain query`

E.g. `explain select 1`

```bash
```

### spice, spiced, OS info

- OS info: `uname -a`

- [ ] Yes
- [ ] No

peasee:
## 📝 Summary

* Ensures that text searches via the UDTF include the search column in the projection, if the search column is the `row_id`/index primary key - as the column is used to later join the table.

## 🔗 Related

* Closes #6784


crates/search/src/generation/text_search/index.rs
```diff
@@ -327,6 +327,11 @@ impl FullTextDatabaseIndex {
         let schema = schema_builder.build();
         Ok(Arc::new(RwLock::new(tantivy::Index::create_in_ram(schema))))
     }
+
+    #[must_use]
+    pub fn column_is_pk(&self, column: &str) -> bool {
+        self.primary_key.len() == 1 && self.primary_key.first().map(String::as_str) == Some(column)
```
Jeadie: Why `self.primary_key.len() == 1`?

peasee: `TextSearchIndexProvider.column` is a single string that cannot represent a composite key. `FullTextDatabase.primary_key` is a `Vec<String>` which could represent a composite key. `.contains()` is not sufficient, as it could match a part of a composite key, so we must check the length.

Jeadie: But for the below change, if `TextSearchIndexProvider.column` is in `FullTextDatabase.primary_key`, it will be removed from the projection, and then we can't `retains the column for the later table join operations`?

peasee: I think I misunderstood what the `column` of the `TextSearchIndexProvider` was. I thought it was part of the PK, but it is the search column - the correct solution is that if the search column appears as any part of the primary key, the column should be retained. That makes more sense now, I'll stick with `.contains()`.

Jeadie (approved):

PR #6846, opened 2025-08-20
Model2Vec embedding model support

### LINKED ISSUE #6651: Enhancement: Support Model2Vec static embedding models
## Goal-State/What/Result

- Expose the ability to use Model2Vec static embedding models via [model2vec-rs](https://github.com/MinishLab/model2vec-rs).
  - model2vec-rs can load pretrained models from HuggingFace (nice DX)
  - We can provide guidance on using [Model2Vec](https://github.com/MinishLab/model2vec) to distill an arbitrary sentence-transformers model

## Why/Purpose

- Static embedding models are several orders of magnitude faster than transformer-based ones (with the tradeoff of superior contextual understanding). A sentence of ~20 tokens can take ~100ms to embed with transformers, but <1ms with a static model.
  - [See MTEB benchmarks](https://github.com/MinishLab/model2vec/blob/main/results/README.md#mteb-results)
- Developers wishing to index a big N dataset, or do embeddings on a high throughput application, 100ms per sentence can be untenable.
  - i.e., this would greatly empower the vector search use case

## By When

**Issue/Spec written and reviewed:** 08/19/25
**Done-Done:** 08/25/25

## Done-Done

- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [ ] Tests
- [ ] Documentation

## The Algorithm

- [x] Simplify.

## Specification

Model parameters:

  | Parameter               | Type               | Description                                                                   |
  |-------------------------|--------------------|-------------------------------------------------------------------------------|
  | hf_token                | String (optional)  | Optional Hugging Face authentication token (private models)                   |
  | normalize               | Boolean (optional) | Whether to normalize embeddings (defaults to model's `config.json`)           |
  | subfolder               | String (optional)  | When using a huggingface model, it may reside in a subfolder of the repo      |
  | parallelism             | Number (optional)  | How many parallel threads to use for embedding. Defaults to system CPU count. |
  | embed_max_token_length  | Number (optional)  | Maximum token length for embedding operations                                 |
  | embed_custom_batch_size | Number (optional)  | Custom batch size override for embedding operations                           |

Spicepod usage example:

```yaml
embeddings:
  - from: model2vec:minishlab/potion-base-8M
    name: potion_only_8_cores
    params:
        parallelism: 8
```

### Security Review

- Review how serialized model data is retrieved + loaded into the runtime, otherwise no security concerns

## How/Implementation Plan

- `from` key in embeddings is a "URI scheme-style" trigger for different model providers, so register `model2vec` as another

Scoped out `embed` UDF to: #6847

## QA Plan

- Test ability to load spicepod defined models
- E2E vector search test

## Release Notes

**Model2Vec embeddings**: Expanded support for static model2vec embeddings in Spicepod definitions, for scenarios requiring low-ms latency during encoding.

```yaml
embeddings:
  - from: model2vec:minishlab/potion-base-8M
    name: potion
  - from: model2vec:path/to/my/local/model
    name: local
```

mach-kernel:
## 📝 Summary

- Adds support for `Model2Vec` embeddings
- `Embed` interface new functions: `embed_sync`, `supports_sync_embeddings`
- `EmbeddingTableExec`
  - Detects if model supports sync embeddings, if so, parallelizes embed with `rayon` crate

## 🔗 Related

Closes #6651

## 📚 Docs

https://github.com/spiceai/docs/pull/1118

Parameters:
  | Parameter               | Type               | Description                                                                   |
  |-------------------------|--------------------|-------------------------------------------------------------------------------|
  | hf_token                | String (optional)  | Optional Hugging Face authentication token (private models)                   |
  | normalize               | Boolean (optional) | Whether to normalize embeddings (defaults to model's `config.json`)           |
  | subfolder               | String (optional)  | When using a huggingface model, it may reside in a subfolder of the repo      |
  | parallelism             | Number (optional)  | How many parallel threads to use for embedding. Defaults to system CPU count. |
  | embed_max_token_length  | Number (optional)  | Maximum token length for embedding operations                                 |
  | embed_custom_batch_size | Number (optional)  | Custom batch size override for embedding operations                           |
```yaml
embeddings:
  - from: model2vec:minishlab/potion-base-8M
    name: potion_only_8_cores
    params:
        parallelism: 8
  - from: model2vec:path/to/my/local/model
    name: local
```


crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -536,10 +630,26 @@ async fn get_vectors_with_chunker(
     Ok((vectors, content_offsets))
 }
 
+fn build_embedding_pool(
```
mach-kernel, line 636: Two thoughts on this problem (what should encapsulate the parallelism):
- Encapsulate the parallelism in the `Embed` facade
- Execution parallelism is an "exec" level concern

The former is a clean design but handling of `null` rows gets in the way, and it would only be applicable to fast in-process embedding pipelines. To do it that way, we'd have to evolve the interface to accept an embedding request of `Vec<Option<String>> -> Vec<Option<Vec<f32>>>` for an `EmbeddingInput::OptionalStringArray(_)`.

The latter guided me in putting this here, though it required some extra methods on `Embed` to allow us to know the configured parallelism and whether or not the model was a candidate for par exec.

Wanted to bring this up, in case someone sees a nicer way

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -400,6 +407,75 @@ pub(super) async fn get_vectors(
...
+            } else {
+                builder.values().append_nulls(vector_length as usize);
+                builder.append(false);
+            }
+        }
+    } else {
+        let embeds: Vec<_> = pool.install(|| {
+            inputs
+                .into_par_iter()
+                .chunks(32)
```
mach-kernel, line 461: Kevin suggested that we should try to embed in batches for performance reasons. There appears to be a small perf improvement when doing some small chunking (hence the magic number).

I decided to only opt in if the input array has no null/blank values (otherwise it gets annoying, we have to chunk around the blanks, etc)

cc @kczimm

crates/runtime/src/embeddings/execution_plan.rs
mach-kernel: Aside / execution parallelism: Spark has `spark.task.cpus` as a blanket configuration for all tasks. It would seem worthwhile to expose that kind of configuration eventually (for the whole runtime) if/when our execs can leverage SMP. Doing it in the scope of this branch seems like a big lift, but also wanted to call it out.

crates/spicepod/src/component/embeddings.rs
```diff
@@ -37,7 +38,7 @@ pub struct Embeddings {
     pub files: Vec<ModelFile>,
 
     #[serde(default, skip_serializing_if = "HashMap::is_empty")]
-    pub params: HashMap<String, String>,
+    pub params: HashMap<String, Value>,
```
mach-kernel, line 41: I really wanted a number

crates/llms/src/model2vec.rs
```diff
@@ -0,0 +1,202 @@
...
+                "for it does not hurt me".to_string(),
+            ]))
+            .await;
+
+        assert!(embed_sentences.is_ok());
+
+        let embed_sentences = embed_sentences.expect("Must embed sentences");
+        assert_eq!(embed_sentences.len(), 2);
+        for embedded_sentence in embed_sentences {
+            assert_eq!(embedded_sentence.len(), 256);
```
phillipleblanc, line 191: Given these are static models, it should be possible to snapshot their output right?

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -290,7 +292,12 @@ pub(crate) async fn compute_additional_embedding_columns(
 
             Arc::new(vectors) as ArrayRef
         } else {
-            let fixed_size_array = get_vectors(arr_iter, &**model, cfg.vector_size).await?;
+            let fixed_size_array = if model.supports_sync_embeddings() {
+                get_vectors_in_process(arr_iter, &**model, cfg.vector_size)?
```
phillipleblanc: Given that this will potentially block the thread, we may need to run this in a [tokio::task::spawn_blocking](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html)

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -400,6 +407,75 @@ pub(super) async fn get_vectors(
...
+    vector_length: i32,
+) -> Result<FixedSizeListArray, Box<dyn std::error::Error + Send + Sync>> {
+    let inputs: Vec<Option<String>> = arr.map(|o| o.map(ToString::to_string)).collect();
+
+    let mut builder = FixedSizeListBuilder::with_capacity(
+        PrimitiveBuilder::<Float32Type>::with_capacity(inputs.len() * (vector_length as usize)),
+        vector_length,
+        inputs.len(),
+    )
+    .with_field(Arc::new(Field::new("item", DataType::Float32, true)));
```
phillipleblanc, line 432:
```suggestion
    .with_field(Arc::new(Field::new("item", DataType::Float32, false)));
```

The inner floats shouldn't be nullable I believe

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -400,6 +407,75 @@ pub(super) async fn get_vectors(
...
+                })
+                .collect()
+        });
+
+        for embed in embeds {
+            if let Some(embedding) = embed {
+                builder.values().append_slice(&embedding[0]);
+                builder.append(true);
+            } else {
+                builder.values().append_nulls(vector_length as usize);
```
phillipleblanc, line 454:
```suggestion
```

You shouldn't need to append nulls to the values array, marking the list array as null will set the correct offsets such that the next non-null list item will start from the correct offset, i.e.:

```markdown
+---------------------+---------------------------+
| List Validity Bitmap| Values Array (Float32)    |
+---------------------+---------------------------+
| [true, false, true] | [1.0, 2.0, 3.0,           |  <- Offset 0: N=0
|                     |  4.0, 5.0, 6.0]           |  <- Offset 3: N=2
+---------------------+---------------------------+
```

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -443,15 +523,27 @@ async fn get_vectors_with_chunker(
         .unzip();
 
     let (chunk_offsets, chunks): (Vec<_>, Vec<_>) = chunks_in_row.into_iter().flatten().unzip();
-
-    let embedded_data = model
-        .embed(EmbeddingInput::StringArray(chunks.clone()))
-        .await
-        .boxed()?;
+    let embedded_data: Vec<Vec<f32>> = if model.supports_sync_embeddings() {
+        let pool = build_embedding_pool(model.parallelism())?;
```
phillipleblanc, line 527: Similarly, this will also likely need a `tokio::task::spawn_blocking`

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -536,10 +628,34 @@ async fn get_vectors_with_chunker(
...
 }
 
+fn build_embedding_pool(
+    model_parallelism: Option<usize>,
+) -> Result<ThreadPool, Box<dyn std::error::Error + Send + Sync>> {
+    let sys_parallelism = thread::available_parallelism();
+
+    let parallelism = match (model_parallelism, sys_parallelism) {
+        (Some(0), Ok(p)) => p.get(),
+        (Some(0), _) => unreachable!("Must determine system parallelism"),
```
phillipleblanc: Should we pick a default parallelism and log a warning if this happens? Looking at the documentation for `thread::available_parallelism` it seems like it could fail for a variety of reasons

crates/runtime/src/embeddings/execution_plan.rs
mach-kernel: Ahh, yes, we're calling this in an async function and shouldn't just block in place. Doing!

crates/runtime/src/embeddings/execution_plan.rs
mach-kernel: I tried, however this assertion tripped:
https://github.com/apache/arrow-rs/blob/a620957bc98b7aa14faec10635bb798932f00bf9/arrow-array/src/builder/fixed_size_list_builder.rs#L174-L180

crates/runtime/src/embeddings/execution_plan.rs
mach-kernel: Sure! I chose a parallelism of 2 (around the 'worst case' scenario of only having a single core).

crates/runtime/src/embeddings/execution_plan.rs
phillipleblanc: ah maybe fixed sized lists work differently?

crates/runtime/src/embeddings/execution_plan.rs
mach-kernel: I agree with you, but it seems the DF schema at this point has it set as `nullable: true`. To expedite the release, I've captured this in an issue and will revert this change:
https://github.com/spiceai/spiceai/issues/6911

crates/llms/src/model2vec.rs
```diff
@@ -0,0 +1,206 @@
...
+            self.embed_custom_batch_size.unwrap_or(1024),
+        ))
+    }
+
+    fn supports_sync_embeddings(&self) -> bool {
+        true
+    }
+
+    fn parallelism(&self) -> Option<usize> {
+        self.parallelism.or(Some(0))
```
Jeadie: Why is `Some(0)` and not `Some(1)` the correct default?

crates/runtime/src/embeddings/execution_plan.rs
```diff
@@ -229,11 +231,11 @@ pub(crate) fn construct_record_batch(
...
 /// The additional columns returned here should match those specified in [`super::table::EmbeddingTable::embedding_fields`]
-pub(crate) async fn compute_additional_embedding_columns(
-    rb: &RecordBatch,
-    embedded_columns: &HashMap<String, EmbeddingColumnConfig>,
+pub(crate) async fn compute_additional_embedding_columns<'a>(
+    rb: &'a RecordBatch,
+    embedded_columns: &'a HashMap<String, EmbeddingColumnConfig>,
     embedding_models: Arc<RwLock<EmbeddingModelStore>>,
-) -> Result<HashMap<String, ArrayRef>, Box<dyn std::error::Error + Send + Sync>> {
+) -> Result<HashMap<String, ArrayRef>, Box<dyn std::error::Error + Send + Sync + 'a>> {
```
Jeadie: Is this more precise/explicit, or a functionally different?

crates/runtime/src/embeddings/execution_plan.rs
Jeadie: One way you might be able to batch an array with nulls/empty is to partition based on this and reconstruct afterwards. We do something similar here

https://github.com/spiceai/spiceai/blob/d2c6a651025cde22bb8c6d9f3699a919c489709a/crates/runtime/src/embeddings/execution_plan.rs#L348-L356

May not be worth it

Jeadie: I now realise you do it below: `crates/runtime/src/embeddings/execution_plan.rs:L532-L541`

crates/llms/src/model2vec.rs
Jeadie: It's so you can know when to use `thread::available_parallelism()` below

Jeadie: I don't know if `Some(0)` or `None` is the better value for "use default thread-available parallelism`

mach-kernel: The reasoning is a bit shallow, I wanted to do something like:
- None means unsupported
- `Some(0)` means supported but not set by the user
- `Some(_)` means supported + value set by user

https://github.com/spiceai/spiceai/blob/25058846e03f606ec568cd32544206fcb9825d97/crates/runtime/src/embeddings/execution_plan.rs#L642-L660

I can move the `available_parallelism` default into the model2vec impl, or use an enum to illustrate the support/lack thereof?

Jeadie (dismissed):

crates/llms/src/model2vec.rs
Jeadie: In my head `supports_sync_embeddings` was sufficient for knowing if they supported parallelism. I assume a synchronous embedding model (in the future) may not support parallelism?

mach-kernel: OH! Yeah I'm cool with `supports_sync_embeddings` gating opt-in to parallelism. Which means we can use `None`. Changing!!

Yeah / sync models should generally be CPU par'd

crates/runtime/src/embeddings/execution_plan.rs
mach-kernel: `rustc` yelled at me for something I did with `spawn_task` but I backed it out and it looks like we no longer need it, thank!

sgrebnov (approved):

PR #6852, opened 2025-08-21
Show correct limit for EXPLAIN plans in `S3VectorsQueryExec`

Jeadie:
## 📝 Summary
Currently, if `S3VectorsQueryExec` needs to truncate the `LIMIT` to 30, it is not reflected in the explain plan. This PR changes that so EXPLAIN plans are more useful for debugging.


crates/data_components/src/s3_vectors/query_provider.rs
```diff
@@ -204,6 +204,16 @@ impl S3VectorsQueryExec {
             Boundedness::Bounded,
         );
 
+        let limit: i32 = if limit > 30 {
+            // https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-vectors-limitations.html
+            tracing::debug!(
```
lukekim: Maybe this should be a verbose trace uses can use too.

phillipleblanc (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #6865, opened 2025-08-21
fix: Update search test case structure

### LINKED ISSUE #6819: Models search integration tests do not fail on errors
The models search integration tests do not fail when there are test errors/failures from SQL tests.

peasee:
## 📝 Summary

* Updates the search tests to use a more defined `SearchTestCase` struct for HTTP and SQL tests
* Adds support for defining whether a test should fail/pass, and whether a test should be skipped
* Updates test handling so that if a test is not supposed to fail, it now returns an actual test failure instead of silencing the error with an error snapshot

A large number of tests are still skipped/disabled which will be addressed in:
* #6815
* #6816
* #6817
* #6818
* #6820
* #6821

* Also makes a minor update to the `CONTRIBUTING.md` to match the rules defined in `make lint-rust`

## 🔗 Related

* Closes #6819


crates/runtime/tests/models/hf.rs
```diff
@@ -224,95 +224,114 @@ mod search {
             vec![
                 SearchTestCase {
                     name: "hf_basic",
-                    body: json!({
+                    body: SearchTestType::Http(json!({
```
Jeadie: You should make a helper function for this since most will be `should_fail: false, skip: false` and itll push the happy path for tests.

Jeadie (approved):

Issue #6779, opened 2025-08-18
Enhancement: Kafka Data Connector


phillipleblanc:
## Goal-State/What/Result

An [alpha-grade](https://github.com/spiceai/spiceai/blob/trunk/docs/criteria/connectors/alpha.md) Kafka Data Connector, that enables direct acceleration of `refresh_mode: append` data from Kafka topics.

```yaml
- from: kafka:some_kafka_topic
   name: data
   acceleration: 
     enabled: true
     refresh_mode: append
   params: 
     kafka_bootstrap_servers: server1:9092
```

## Why/Purpose

Kafka is one of the most widely adopted event streaming platforms, powering real-time data pipelines and analytics across many industries. Supporting Kafka as a data source allows Spice to integrate seamlessly with existing customer infrastructure without requiring additional middleware or connectors, and is a frequently requested capability by customers.

## By When

**Issue/Spec written and reviewed:** 08/20
**M1. Basic support** 08/25
**M2. Cooperative mode** 08/27
**Done-Done:** 09/01

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### M1. Basic support

The connector is defined using the `kafka:` prefix. When configuring the connector, you must specify `refresh_mode: append`.

```yaml
- from: kafka:orders_events
  name: orders
  acceleration: 
    enabled: true
    refresh_mode: append
  params: 
    kafka_bootstrap_servers: server:9092
```

- Upon startup, Spice fetches all messages for the topic using a uniquely generated consumer group. Multiple Spice instances connected to the same topic will each read the data independently.
- If a persistent acceleration engine is used, data is fetched starting from the last processed record, without requiring a restart from the beginning.
- Schema is inferred based on the first available topic message (JSON). Note: this will be improved later by adding support for a configurable `kafka_schema_infer_max_samples` or similar.

[...]

PR #6870, opened 2025-08-21
Kafka Data Connector: persist consumer between restarts

### LINKED ISSUE #6858: Kafka: consumer group persistence between restarts

sgrebnov:
## 📝 Summary

To avoid generating a new consumer group ID and re-reading data from the beginning (as happens without persisted state), this PR implements logic similar to the Debezium/CDC connector. The Kafka Data Connector now stores metadata—such as the generated consumer group ID, schema, and topic information—in `spice_sys_kafka`. On restart, Spice will check if a consumer group ID already exists and reuse it instead of generating a new one.

This is only applicable for file based acceleration and currently implemented/supported for `duckdb` only

## 🔗 Related

Part of https://github.com/spiceai/spiceai/issues/6779

Closes https://github.com/spiceai/spiceai/issues/6858


crates/runtime/src/dataaccelerator/spice_sys/kafka/duckdb.rs
```diff
@@ -0,0 +1,204 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use super::{KAFKA_TABLE_NAME, KafkaMetadata, KafkaSys, Result};
+use datafusion_table_providers::sql::db_connection_pool::duckdbpool::DuckDbConnectionPool;
+use std::sync::Arc;
+
+impl KafkaSys {
```
Jeadie, line 21: Question (somewhat related) about implementing per-dataaccelerator SQL functions. Could we use the Datafusion SQL API to have a single `CREATE TABLE IF NOT EXISTS....` function that could then use the `sqlparser`s underneath to convert to each underlying data accelerator SQL dialect (i.e. avoid needing `.../spice_sys/kafka/duckdb.rs`, `.../spice_sys/kafka/postgres.rs`, etc)?

Jeadie (approved):

crates/runtime/src/dataaccelerator/spice_sys/kafka/duckdb.rs
phillipleblanc: DataFusion doesn't support full DDL statements (including CREATE TABLE), so it wouldn't be representable in a LogicalPlan and thus unable to be unparsed into a different dialect.

phillipleblanc (approved):

PR #6871, opened 2025-08-21
Add override parameter to force support for the Responses API

### LINKED ISSUE #6849: Add override for Responses API via a parameter

Advayp:
## 📝 Summary
- See title
- By default, Spice will only add a model to the Responses API registry if it is from OpenAI and OpenAI's API only. However, in certain cases, custom endpoints supporting the Responses API may be available (e.g, like Spice's `/v1/responses`). In these cases, the default behavior will fail. This PR adds support for a parameter that forces the specified model into the Responses API registry, assuming the health succeeds.

Sample Spicepod:
```yaml
models:
    - from: openai:my-model
      name: my-model
      params:
          responses_api: enabled # enabled by default for OpenAI and Azure OpenAI, disabled for every other provider
          endpoint: https://endpoint.responses-api-compatible.com/v1
```

## 🔗 Related
- Closes #6849

### DIFF
--- a/crates/runtime/src/init/llm.rs
+++ b/crates/runtime/src/init/llm.rs
@@ -33,6 +33,15 @@ use spicepod::component::model::{Model as SpicepodModel, ModelSource};
 static DEFAULT_OPENAI_ENDPOINT: &str = "https://api.openai.com/v1";
 
 fn supports_responses_api(spicepod_model: &SpicepodModel) -> bool {
+    if *spicepod_model
+        .params
+        .get("supports_responses")
+        .unwrap_or(&Value::Bool(false))
+        == Value::Bool(true)
+    {
+        return true;
+    }
+
     if spicepod_model.get_source() != Some(ModelSource::OpenAi) {
         return false;
     }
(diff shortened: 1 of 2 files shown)


crates/runtime/src/init/llm.rs
```diff
@@ -33,6 +33,11 @@ use spicepod::component::model::{Model as SpicepodModel, ModelSource};
 static DEFAULT_OPENAI_ENDPOINT: &str = "https://api.openai.com/v1";
 
 fn supports_responses_api(spicepod_model: &SpicepodModel) -> bool {
+    let responses_api_override = spicepod_model.params.get("responses_api");
+    if let Some(value) = responses_api_override {
+        return value == &Value::String("enabled".to_string());
```
lukekim: Should be a case-insensitive comparison, right?

lukekim (dismissed):

lukekim (dismissed):

lukekim (dismissed):

Jeadie (approved):

Issue #6666, opened 2025-08-01
Enhancement: OpenAI Model Provider Responses API Support


lukekim:
## Goal-State/What/Result

A `/v1/responses` endpoint in Spice's HTTP server supporting the exact same interface as [Open AI's Responses API](https://platform.openai.com/docs/api-reference/responses) with integration with Spice's functionality for tools.

## Why/Purpose

This is a requested feature and the OpenAI Responses API is OpenAI's preferred way of interacting with its API, thus adhering to Spice's principle of aligning to industry standards.

## By When

**Issue/Spec written and reviewed:** August 14, 2025
**Done-Done:**  August 28, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [ ] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [ ] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

This enhancement adds support for the OpenAI Responses API (`/v1/responses`) to the Spice OSS runtime. This allows users to interact with models using the latest interface from OpenAI. This implementation will create a new endpoint in Spice's HTTP server with path `/v1/responses`. The structure of each request to and response from this endpoint will be the exact same as that present in [Open AI's Responses API](https://platform.openai.com/docs/api-reference/responses). This route will return a 404 error if the model specified in the request was not found or if it does not support the responses API, and 200 on success.

`spice chat` will also support using the OpenAI Responses API. To do this, use the `--responses` flag.
Example:
```bash
spice chat --responses # Use the `/v1/responses` endpoint for all completions instead of `/v1/chat/completions`
```

The Responses API supports a specific set of tools not available from the Chat Completions API. These will be specified like so:
```yaml
models:
    - name: test
      from: openai:gpt-4.1
      params:
          openai_api_key: ${ secrets:SPICE_OPENAI_API_KEY }
          tools: sql, list_datasets
          openai_responses_tools: web_search # 'local_shell', 'code_interpreter', or 'web_search'
```
[...]

PR #6886, opened 2025-08-22
Add test to ensure tools are injected correctly in the Responses API

Advayp:
## 📝 Summary
- See title

## 🔗 Related
- Related to #6666

### DIFF
--- a/crates/runtime/tests/models/openai.rs
+++ b/crates/runtime/tests/models/openai.rs
@@ -701,7 +701,16 @@ async fn openai_responses_api_tools() -> Result<(), anyhow::Error> {
                 .input("Tell me about the movie Ocean's Eleven")
                 .build()?;
 
-            let response = openai_client.responses().create(request).await?;
+            let responses_client = openai_client.responses();
+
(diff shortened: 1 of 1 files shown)


crates/runtime/tests/models/openai.rs
```diff
@@ -638,6 +641,129 @@ async fn openai_responses_api_streaming() -> Result<(), anyhow::Error> {
...
+    );
+    model
+        .params
+        .insert("tools".into(), serde_json::Value::String("auto".into()));
+    model
+}
+
+#[tokio::test]
+async fn openai_responses_api_tools() -> Result<(), anyhow::Error> {
+    let _tracing = init_tracing(None);
```
sgrebnov, line 662: Pls verify this is consistent with other tests (tracing config)

sgrebnov (approved):

crates/runtime/tests/models/openai.rs
Advayp: It's consistent with the other tests for OpenAI

PR #6893, opened 2025-08-22
Add tracing and system prompt support for the Responses API

### LINKED ISSUE #6890: Telemetry and System Prompt Injection for Responses API

Advayp:
## 📝 Summary
- See title
- Necessary as it's part of an enhancement
- This code is directly copied from `ChatWrapper`, but adapted to work with the Responses API

## 🔗 Related
- Closes #6890

### DIFF
--- a/crates/runtime/src/model/metrics.rs
+++ b/crates/runtime/src/model/metrics.rs
@@ -18,6 +18,7 @@ use std::{sync::LazyLock, time::Duration};
 
 use async_openai::types::{
     ChatCompletionNamedToolChoice, ChatCompletionToolChoiceOption, CreateChatCompletionRequest,
+    responses::CreateResponse,
 };
 use opentelemetry::{
     Key, KeyValue, StringValue, Value, global,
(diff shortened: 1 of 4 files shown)


crates/runtime/src/model/metrics.rs
```diff
@@ -91,6 +92,38 @@ pub(crate) fn request_labels(req: &CreateChatCompletionRequest) -> Vec<KeyValue>
...
+            Value::I64(req.tools.as_deref().unwrap_or_default().len() as i64),
+        ),
+        KeyValue::new(Key::new("model"), Value::String(req.model.clone().into())),
+        KeyValue::new(Key::new("responses_api"), Value::Bool(true)),
+    ];
+
+    if let Some(ref metadata) = req.metadata {
+        labels.push(KeyValue::new(
+            Key::new("metadata"),
+            Value::String(format!("{metadata:?}").into()),
```
Jeadie, line 113: I think we could pass the entire `metadata` here
```rust
labels.push(KeyValue::new(Key::new("metadata"), metadata.clone())
```

I see we provide a string in `crates/runtime/src/model/metrics.rs` too. I don't recall why

Jeadie (approved):

crates/runtime/src/model/metrics.rs
Advayp: The value in the key-value pair has to be of type `opentelemetry::Value`, which has the following variants:
```rust
/// The value part of attribute [KeyValue] pairs.
#[non_exhaustive]
#[derive(Clone, Debug, PartialEq)]
pub enum Value {
    /// bool values
    Bool(bool),
    /// i64 values
    I64(i64),
    /// f64 values
    F64(f64),
    /// String values
    String(StringValue),
    /// Array of homogeneous values
    Array(Array),
}
```
To adhere to this, we pass `metadata` in as a `Value::String` as that's the closest variant to it.

PR #6894, opened 2025-08-23
Support using the Responses API from `spice chat`

### LINKED ISSUE #6801: Add `--responses` flag to `spice chat` to support using the Responses API from `spice chat`

Advayp:
## 📝 Summary
- See title
- Necessary as it's part of an enhancement and improves the UX of interacting with the OpenAI Responses API
- Streaming is not supported yet, as the Responses API doesn't support streaming with tools yet
- Tests for the `/v1/models` route will come in a separate PR

Notable UX changes:
- To use the Responses API from `spice chat`, use the `--responses` flag. E.g.: `spice chat --responses`
- `/v1/models` supports an additional query parameter `metadata_fields`, which controls what fields to include for each model in addition to the ones included by default. Currently, the only field you can include is `supports_responses_api`

## 🔗 Related
- Closes #6801 

## 📚 Docs
-Docs PR: https://github.com/spiceai/docs/pull/1111


bin/spice/cmd/chat.go
```diff
@@ -243,7 +298,84 @@ spice chat --model <model> "What is Spice.ai?"
 			}
 		}
 
-		getChatResponse := func(messages []Message, useSpinner bool) ([]Message, error) {
+		// Handler for Responses API - handles non-streaming responses
+		handleResponsesAPI := func(messages []Message, useSpinner bool) ([]Message, error) {
```
Jeadie: `handleResponsesApi` and `handleChatCompletions` should be extracted as functions now that we have two.

Jeadie: Likely theres some additional state (i.e. more than just `messages []Message, useSpinner bool`) mutated internally

bin/spice/cmd/chat.go
```diff
@@ -108,6 +108,52 @@ type ChatCompletion struct {
...
+	Role    string                 `json:"role"`
+	Content []ResponseContentBlock `json:"content"`
+}
+
+type ResponseContentBlock struct {
+	Type string `json:"type"`
+	Text string `json:"text"`
+}
+
+type ResponseUsage struct {
```
Jeadie, line 139: Is this same as chat usage? If so, we won't need `generateResponsesUsageMessage`

crates/runtime/src/http/v1/models.rs
```diff
@@ -40,6 +44,10 @@ pub struct ModelsQueryParams {
     /// If true, includes the status of each model in the response.
     #[serde(default)]
     pub status: bool,
+
+    /// A comma-separated list of metadata fields to include in the response (e.g., `supports_responses_api`)
+    #[serde(default)]
+    pub metadata_fields: String,
```
Jeadie, line 50: Is this the desired UX? (e.g. `v1/models?status=true&metadata_fields=supports_responses_api`). I don't know if it looks great for `metadata_fields > 2`.

Advayp: Yeah Luke and I discussed this. I can get him to reconfirm that this is what he wants.

bin/spice/cmd/chat.go
Advayp: No, this includes some additional fields and some fields with different names.

Jeadie (approved):

PR #6903, opened 2025-08-25
Create comprehensive S3vectors test

Jeadie:
## 📝 Summary
 - Add tests for s3 vector search, as per existing vector search tests. Test specifically with metadata columns and composite primary keys.


crates/runtime/tests/models/search.rs
```diff
@@ -127,7 +129,10 @@ pub async fn run_search_test(
         return Ok(());
     }
 
-    let resp = serde_json::from_str(&resp?).context("Failed to parse HTTP response")?;
+    let resp = anyhow::Context::context(
+        serde_json::from_str(&resp?),
+        "Failed to parse HTTP response",
+    )?;
```
kczimm: Aren't these the same?

Jeadie: Yes. At one point in time, this PR was using two different `context` functions and needed to delineate.

crates/runtime/tests/models/search.rs
```diff
@@ -287,13 +289,32 @@ async fn start_app(app: App) -> Result<Config, anyhow::Error> {
 pub(crate) async fn run_search(
     app: App,
     test_cases: Vec<SearchTestCase>,
+) -> Result<(), anyhow::Error> {
+    run_search_w_explain(app, test_cases, false).await
+}
+
+// if `explain_sql`, for any [`SearchTestCase`] that is [`SearchTestType::Sql`], a snapshot will be taken of the associated explain query.
+pub(crate) async fn run_search_w_explain(
```
Advayp, line 297:
```suggestion
pub(crate) async fn run_search_with_explain(
```
Rename to this?

peasee (approved):

PR #6910, opened 2025-08-25
Better error message when credentials aren't loaded for S3 Vectors

### LINKED ISSUE #6733: S3Vectors error message unclear on incorrect credentials
## Describe the bug

When the S3Vectors component is configured with incorrect credentials, the error message returned is not specifc enough:

2025-08-12T02:17:46.160879Z  WARN runtime::init::dataset: Cannot connect to the dataset pulls (github).
Failed to get bucket from S3 Vectors. unhandled error

## To Reproduce

Run the S3 Vectors cookbook without specifying the credentials: https://github.com/spiceai/cookbook/tree/trunk/vectors/s3

phillipleblanc:
## 📝 Summary

Prevents the S3 Vectors client from retrying when the credentials aren't loading (which leads to a poor experience of waiting for the Spice instance to do something, just to finally error out).

Also fixes the error message to actually indicate what the error is.

## 🔗 Related

- Closes #6733

### DIFF
--- a/crates/runtime/src/embeddings/index/retry_client.rs
+++ b/crates/runtime/src/embeddings/index/retry_client.rs
@@ -14,7 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+use std::error::Error;
+
 use async_trait::async_trait;
+use aws_credential_types::provider::error::CredentialsError;
 use s3_vectors::{
     Client, CreateIndexError, CreateIndexInput, CreateIndexOutput, CreateVectorBucketError,
     CreateVectorBucketInput, CreateVectorBucketOutput, DeleteIndexError, DeleteIndexInput,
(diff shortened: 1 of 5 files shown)


crates/runtime/src/embeddings/index/retry_client.rs
```diff
@@ -114,7 +117,21 @@ impl S3Vectors for S3VectorRetryClient {
...
+                            .is_some_and(|ce| {
+                                matches!(ce, CredentialsError::CredentialsNotLoaded(_))
+                            });
+
+                        if credentials_not_loaded {
+                            Err(RetryError::permanent(e))
+                        } else {
+                            Err(RetryError::transient(e))
+                        }
+                    }
```
kczimm, line 134: Looks like we copy/paste this 16 times. Maybe a macro would be beneficial here.

kczimm (approved):

crates/runtime/src/embeddings/index/retry_client.rs
phillipleblanc: I'm not a huge fan of macros in general - its a much nicer UX imo to be able to debug to the exact line where something happens.

For this specific case its probably fine, but this file is mostly boilerplate code anyway.

lukekim (approved):

PR #6912, opened 2025-08-25
Refactor how models supporting the Responses API are loaded

Advayp:
## 📝 Summary
- Refactors the loading of models supporting the Responses API to be opt-in. The loading stage fails completely if the Chat completions health check fails, but only warns when the Responses API health check fails.

### DIFF
--- a/crates/runtime/src/init/model.rs
+++ b/crates/runtime/src/init/model.rs
@@ -121,27 +121,19 @@ impl Runtime {
         tracing::trace!("Model type for {} is {:#?}", m.name, model_type.clone());
         let result: Result<(), Error> = match model_type {
             Some(ModelType::Llm) => match self.load_llm(m.clone(), params).await {
-                Ok((Some(l), Some(responses_model))) => {
+                Ok((completions_model, Some(responses_model))) => {
                     let mut llm_map = self.completion_llms.write().await;
-                    llm_map.insert(m.name.clone(), l);
+                    llm_map.insert(m.name.clone(), completions_model);
                     drop(llm_map);
                     let mut responses_llm_map = self.responses_llms.write().await;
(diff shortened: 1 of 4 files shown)


crates/runtime/src/init/model.rs
```diff
@@ -121,27 +121,19 @@ impl Runtime {
         tracing::trace!("Model type for {} is {:#?}", m.name, model_type.clone());
         let result: Result<(), Error> = match model_type {
             Some(ModelType::Llm) => match self.load_llm(m.clone(), params).await {
-                Ok((Some(l), Some(responses_model))) => {
```
Jeadie, line 124: Now that only one of these is `Option<T>`, we can avoid duplicate logic without `match`. Something like: 
```rust
 Some(ModelType::Llm) => {
                let (completions_model, responses_model_opt) = self
                    .load_llm(m.clone(), params)
                    .await
                    .boxed()
                    .map_err(|e| Error::FailedToLoadLLM {
                        name: m.name.clone(),
                        source: Box::new(e),
                    })?;
                let mut llm_map = self.completion_llms.write().await;
                llm_map.insert(m.name.clone(), completions_model);
                drop(llm_map);

                if let Some(responses_llms) = responses_model_opt {
                    let mut responses_llm_map = self.responses_llms.write().await;
                    responses_llm_map.insert(m.name.clone(), responses_llms);
                }
                Ok(())
            }
```

Advayp: The `?` causes the error to propagate all the way up to `load_model`, which doesn't return a result. I could refactor the `match`ing on the model type into its own function that returns a result, but given that it's only used here I'm not sure how helpful that would be.

Jeadie (approved):

Issue #5309, opened 2025-04-07
Bug: `refresh_mode: changes` prevents Spice runtime from becoming Ready


sgrebnov:
## Describe the bug

When I follow the CDC Debezium example, the Spice runtime always reports as "Not Ready" because the CDC dataset is always in a "Refreshing" status.

```bash
curl http://localhost:8090/v1/ready
not ready%  
```

## To Reproduce

1. Follow `cdc-debezium` instructions: https://github.com/spiceai/cookbook/blob/trunk/cdc-debezium/README.md
2. Verify runtime status:

```bash
curl http://localhost:8090/v1/ready
not ready%  
```

phillipleblanc:
Duplicate of #5201

PR #6914, opened 2025-08-25
Mark append-stream and CDC datasets as ready after first message

sgrebnov:
## 📝 Summary

In streaming append mode (for example Kafka connector), a dataset should be marked `Ready` as soon as the first message is successfully received. This ensures that the Spice runtime can successfully set the status to `Ready`, indicating that querying may begin.

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task/changes.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task/changes.rs
@@ -16,7 +16,6 @@ limitations under the License.
 
 use super::RefreshTask;
 use crate::accelerated_table::refresh::Refresh;
-use crate::component::dataset::acceleration::RefreshMode;
 use crate::datafusion::error::find_datafusion_root;
 use crate::{dataupdate::StreamingDataUpdateExecutionPlan, status};
 use arrow::array::{Int32Array, Int64Array, RecordBatch, StringArray};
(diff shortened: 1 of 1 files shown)


crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -72,6 +73,11 @@ impl RefreshTask {
     ) -> crate::accelerated_table::Result<()> {
         let dataset_name = self.dataset_name.clone();
         let sql = refresh.read().await.sql.clone();
+
+        // The DataConnector's `append_stream(&mut self, append_stream: ChangesStream)` implementation is based on this same `start_changes_stream`
```
sgrebnov:
```suggestion
        // The DataConnector's `append_stream(&mut self, append_stream: ChangesStream)` implementation is based on the same `start_changes_stream`
```

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -88,6 +94,12 @@ impl RefreshTask {
                             }
                             initial_load_completed.store(true, Ordering::Relaxed);
 
+                            // In Append mode, mark the dataset as ready immediately after the first message is received
+                            if is_append_mode_stream {
+                                self.update_component_status(status::ComponentStatus::Ready)
```
phillipleblanc: I think we should probably do this for changes as well to mitigate https://github.com/spiceai/spiceai/issues/5309

sgrebnov: Updated 👍

https://github.com/spiceai/spiceai/pull/6914/commits/1b388746f6297341fb396232157047300de12dc3

phillipleblanc (approved):

PR #6934, opened 2025-08-26
[Debezium] Support for VariableScaleDecimal

### LINKED ISSUE #6264: Bug. `io.debezium.data.VariableScaleDecimal` type is not supported
## Describe the bug

There is the following error when using cdc with Postgres numeric type

>2025-06-18T02:16:58.279058Z  WARN runtime::init::dataset: Failed to load the dataset lineitem (debezium).
Debezium field type struct not supported yet

With additional logging added I see the following type info 

```rust
_ => {
            println!("Field details: field_type={:?}, optional={:?}, name={:?}, field={:?}, version={:?}, parameters={:?}",
                field.field_type,
                field.optional,
                field.name,
                field.field,
                field.version,
                field.parameters
            );
            return DebeziumFieldNotSupportedSnafu {
                field_type: field.field_type.clone(),
            }
            .fail();
        }
```

>Field details: field_type="struct", optional=true, name=Some("io.debezium.data.VariableScaleDecimal"), field=Some("l_quantity"), version=Some(1), parameters=None

Note: there is a workaround to avoid this - 
"decimal.handling.mode": "double"

## To Reproduce

```bash
                                      Table "public.lineitem"
     Column      |         Type          | Collation | Nullable |               Default                
-----------------+-----------------------+-----------+----------+--------------------------------------
 id              | integer               |           | not null | nextval('lineitem_id_seq'::regclass)
 l_orderkey      | integer               |           |          | 
 l_partkey       | integer               |           |          | 
 l_suppkey       | integer               |           |          | 
 l_linenumber    | integer               |           |          | 
 l_quantity      | numeric               |           |          | 
 l_extendedprice | numeric               |           |          | 
 l_discount      | numeric               |           |          | 
 l_tax           | numeric               |           |          | 
 l_returnflag    | character(1)          |           |          | 
 l_linestatus    | character(1)          |           |          | 
 l_shipdate      | date                  |           |          | 
 l_commitdate    | date                  |           |          | 
 l_receiptdate   | date                  |           |          | 
 l_shipinstruct  | character varying(25) |           |          | 
 l_shipmode      | character varying(10) |           |          | 
 l_comment       | character varying(44) |           |          | 
Indexes:
    "lineitem_pkey" PRIMARY KEY, btree (id)
Publications:
    "dbz_publication"
```

krinart:
## 📝 Summary

In certain cases (for example when postgres uses `numeric` type which has no predefined precision/scale), Debezium uses `VariableScaleDecimal` type. 

However arrow needs a pre-defined scale in order to create a schema.

This requires:
* Having a default Arrow datatype: We will use `DataType::Decimal128(38, 20)` to make it consistent withe oracle/postgres default values
* Special logic to handle value which was added in this PR

Additionally I've included a better error message for `CDC message key is missing` error: now it mentions that the most likely cause for the error is a specific table not having a primary key

## 🔗 Related
https://github.com/spiceai/spiceai/issues/6264

## 📚 Docs

Needs update [Debezium Connector Docs](https://github.com/spiceai/docs/blob/trunk/website/docs/components/data-connectors/debezium.md) to mention that for `VariableScaleDecimal` we will use `DataType::Decimal128(38, 20)`

### DIFF
--- a/crates/data_components/src/debezium/arrow.rs
+++ b/crates/data_components/src/debezium/arrow.rs
@@ -30,10 +30,10 @@ use arrow::{
 };
 use base64::prelude::*;
 use chrono::{DateTime, NaiveTime, Timelike, Utc};
-use snafu::prelude::*;
-use std::sync::Arc;
 use serde_json::Value as Json;
+use snafu::prelude::*;
 use std::cmp::Ordering::*;
+use std::sync::Arc;
 
 pub mod changes;
 
@@ -199,9 +199,7 @@ fn append_field_value_to_builder(
         }
         DataType::Decimal128(_, scale) => {
             let decimal_builder = downcast_builder::<Decimal128Builder>(builder)?;
-            decimal_builder.append_value(
-                convert_json_to_decimal(field_value, *scale)?,
-            );
+            decimal_builder.append_value(convert_json_to_decimal(field_value, *scale)?);
         }
         DataType::Timestamp(unit, time_zone) => match (unit, time_zone) {
             (TimeUnit::Microsecond, None) => {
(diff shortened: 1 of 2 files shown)


crates/data_components/src/debezium/arrow.rs
```diff
@@ -427,6 +427,84 @@ fn convert_string_to_decimal(field_value: &str) -> Result<i128> {
...
+    Some(acc)
+}
+
+fn rescale_i128(unscaled: i128, src_scale: i8, dst_scale: i8) -> Result<i128> {
+    match src_scale.cmp(&dst_scale) {
+        Equal => Ok(unscaled),
+        Less => {
+            let diff = dst_scale - src_scale;
+            let mul = pow10_i128(diff).context(InvalidDecimalJsonSnafu {
+                reason: format!("overflow computing 10^{diff}"),
```
phillipleblanc: nit: instead of having a single error with multiple static reasons, just create multiple error variants.

krinart: Done

phillipleblanc (approved):

Issue #6964, opened 2025-08-28
`regexp_match` returns a non-array value in DuckDB


peasee:
## Describe the bug

_A clear and concise description of what the bug is._

It is not well documented, but `regexp_match` although matching only the first value - actually returns an array with the single string value.

`regexp_extract` in DuckDB instead returns a string value, not an array with a string value.

This results in a decoder error when attempting to run queries:

```console
sql> select regexp_match(target, 'he') from testing;
Internal Error: Unexpected internal error. Use '.error' for details.
sql> .error
An internal error occurred. Unable to convert record batch: Cast error: Failed to decode value: Json error: Encountered unexpected 'h' whilst parsing value
```

## To Reproduce

_Steps to reproduce the behavior:_

1. Run a `regexp_match` on a DuckDB source.
2. See error

## Expected behavior

_A clear and concise description of what you expected to happen._

The DuckDB function should re-map the string value to an array with a string value.

peasee:
Wrapping `regexp_extract` in `array_value` results in a different error:

```console
sql> select regexp_match(target, 'he') from testing;
Internal Error: Unexpected internal error. Use '.error' for details.
sql> .error
An internal error occurred. Unable to convert record batch: Invalid argument error: column types must match schema types, expected List(Field { name: "item", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }) but found List(Field { name: "", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }) at column index 0
```

The schema is there, but it's not generating a column name from DuckDB ☹️

Issue #6963, opened 2025-08-28
DuckDB core dumps `regexp_like` when flags of `c` or `m` are supplied


peasee:
## Describe the bug

_A clear and concise description of what the bug is._

Run a query with a `regexp_like`, using a flag of either `m` or `c`. Observe DuckDB core dumping the query.

```console
spiced: /home/owner/spiceai/bin/spiced/../../target/debug/build/libduckdb-sys-7e23453aca535df6/out/duckdb/src/optimizer/rule/empty_needle_removal.cpp:28: virtual duckdb::unique_ptr<duckdb::Expression> duckdb::EmptyNeedleRemovalRule::Apply(duckdb::LogicalOperator&, duckdb::vector<std::reference_wrapper<duckdb::Expression> >&, bool&, bool): Assertion `root.children.size() == 2' failed.
```

```sql
select * from testing where regexp_like(target, 'hello', 'm');
```

`testing` is a table with a single text column `target`, loaded with 2 rows: `hello`, and `world`.

## To Reproduce

_Steps to reproduce the behavior:_

1. Create a DuckDB spicepod with a simple database source
2. Run a regular expression query with a `c` or `m` flag
3. See error

## Expected behavior

_A clear and concise description of what you expected to happen._

The query should execute successfully.

## Additional context

The `i` flag works, but `c` and `m` do not. The core dump behavior cannot be observed in the plain `duckdb` CLI.

kczimm:
There is a debug assert on that line in the cpp code. If you build spice in release mode, it works fine.

```sql
sql> select * from testing where regexp_like(target, 'hello', 'c');
+--------+
| target |
+--------+
| hello  |
+--------+

Time: 0.012041166 seconds. 1 rows.
sql> select * from testing where regexp_like(target, 'hello', 'm');
+--------+
| target |
+--------+
| hello  |
+--------+

Time: 0.004736958 seconds. 1 rows.
```

Issue #6781, opened 2025-08-18
Enhancement: Support pushdown of `regexp_` functions to DuckDB


phillipleblanc:
## Goal-State/What/Result

`regexp*` functions work with the DuckDB accelerator & connector, and are transparently pushed down to DuckDB.

## Why/Purpose

The following DataFusion `regexp_*` functions are currently not pushed down to DuckDB via federation, because the [functions](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#regular-expression-functions) don't exist in DuckDB, which causes SQL errors when DuckDB is used as an accelerator/connector:

- [regexp-count](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#regexp-count)
- [regexp-instr](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#regexp-instr)
- [regexp-like](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#regexp-like)
- [regexp-match](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#regexp-match)
- [regexp-replace](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#regexp-replace)

## By When

**Issue/Spec written and reviewed:** 28th Aug, 2025
**Done-Done:** ~~29th Aug, 2025~~

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

DuckDB has its own set of regular expression functions: https://duckdb.org/docs/stable/sql/functions/regular_expressions.html

For each DataFusion regexp function, match it to the relevant DuckDB one (or omit with an explanation if there doesn't exist a possible mapping).

### Security Review

* Regular expressions can cause high compute usage. Users should always ensure only authorized users have access to executing SQL queries, to prevent unauthorized users executing malicious regular expressions that cause denial of service.

## How/Implementation Plan

[...]

PR #6966, opened 2025-08-28
feat: Add initial DuckDB regexp pushdown support

peasee:
## 📝 Summary

* Adds initial support for `regexp_like`, `regexp_count` and `regexp_replace` to DuckDB
* Also adds `regexp_match`, but there is a bug preventing this from functioning: #6964

* There is a bug with some flags for regex functions, but if not used the functions work as expected. Ref: #6963

## 🔗 Related

* Part of #6781


crates/runtime/tests/duckdb/mod.rs
```diff
@@ -240,3 +240,113 @@ async fn duckdb_order_by_special_cases() -> Result<(), String> {
...
+    test_request_context()
+        .scope(async {
+            let sample_csv_contents = include_str!("../test_data/regions.csv");
+            // Write the sample file to a temporary directory
+            let temp_dir = std::env::temp_dir().join("spiced_test_data_order_by");
+            std::fs::create_dir_all(&temp_dir).expect("failed to create temp dir");
+            let sample_csv_path = temp_dir.join("regions.csv");
+            std::fs::write(&sample_csv_path, sample_csv_contents)
+                .expect("failed to write sample file");
+            defer! {
```
phillipleblanc: nit: You can use NamedTempFile (I added it as a dev dependency in my most recent PR) and avoid some boilerplate here. https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#method.new

phillipleblanc (dismissed):

mach-kernel (dismissed):

crates/runtime/tests/duckdb/mod.rs
```diff
@@ -240,3 +240,113 @@ async fn duckdb_order_by_special_cases() -> Result<(), String> {
...
+                std::fs::remove_dir_all(&temp_dir).expect("failed to remove temp dir");
+            }
+
+            let app = AppBuilder::new("duckdb_regexp_test")
+                .with_dataset(make_duckdb_acceleration_dataset(
```
Jeadie, line 268: Could we get a non-DuckDB dataset and compare the result set between the two to ensure we don't get differences in behaviour?

crates/runtime/src/datafusion/dialect/duckdb.rs
```diff
@@ -126,6 +133,216 @@ pub(crate) fn rand_to_random(
...
+    }
+
+    fn postprocess_function(&self, mut ast_fn: ast::Expr) -> datafusion_expr::sqlparser::ast::Expr {
+        match self {
+            DuckDBRegexpFunction::Match => {
+                // DuckDB ``regexp_extract`` returns a plain string
+                // DataFusion ``regexp_match`` returns an array with a single string value
+                // wrap the output of the DuckDB function with ``array_value(arg1, ...)``
+                // https://github.com/spiceai/spiceai/issues/6964
+                ast_fn = ast::Expr::Function(Function {
```
Jeadie: nit: use function
```rust
fn wrapz(ast_fn: ast::Expr, function_name: &str) -> {
    ast::Expr::Function(Function {
        name: ObjectName(vec![ast::ObjectNamePart::Identifier(Ident::new(function_name))]),
        args: ast::FunctionArguments::List(ast::FunctionArgumentList {
            duplicate_treatment: None,
            args: vec![FunctionArg::Unnamed(FunctionArgExpr::Expr(ast_fn))],
            clauses: vec![],
        }),
        filter: None,
        null_treatment: None,
        over: None,
        within_group: vec![],
        parameters: ast::FunctionArguments::None,
        uses_odbc_syntax: false,
    });
}
```

crates/runtime/src/datafusion/dialect/duckdb.rs
```diff
@@ -126,6 +133,216 @@ pub(crate) fn rand_to_random(
...
+            DuckDBRegexpFunction::Count => REGEXP_COUNT_NAME,
+        }
+    }
+
+    /// Maps an input function to an underlying function, whose underlying function accepts the same arguments as the input function
+    /// For example, ``DataFusion``'s ``regexp_like`` -> ``DuckDB``'s ``regexp_matches``
+    pub(super) fn to_datafusion_function(
+        &self,
+        flags_position: usize,
+    ) -> impl Fn(
```
Jeadie, line 281: Could be worth returning in a `Box<T>` so that the return type can be `ScalarFnToSqlHandler`.

peasee: Even if it returned as a `Box`, we would still need an `as ScalarFnToSqlHandler` to coalesce it to the same type so we can store it in the vec with the other functions. We already do `as ScalarFnToSqlHandler` anyway, so I don't think boxing it beforehand would make any difference.

sgrebnov (approved):

Issue #6847, opened 2025-08-20
Enhancement: `embed` as UDF


mach-kernel:
## Goal-State/What/Result

- Expose a 'pure' `embed` UDF: `embed('name_in_spicepod', 'query')`
  - Benchmarking different models is going to be valuable, and this would allow us to test "nothing but" the embedding xform in a pure way (e.g. externally over SQL line)

## Why/Purpose

- Observe transformation performance: runtime and process telemetry for embedding transforms against precanned dataset
- Allow for relevancy testing

## By When

**Issue/Spec written and reviewed:** _Target Date_
**Done-Done:** _Target Date_

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

#### New embed UDF

Introduced the embed() scalar function for generating vector embeddings directly within SQL queries, supporting both
single text strings and batch processing with arrays.

Features:
- On-demand embedding generation: Create embeddings for any text during query execution
- Batch processing: Process arrays of text strings in a single function call
- Model flexibility: Use any configured embedding model in your Spicepod

Usage:
```sql
-- Single text embedding
SELECT embed('hello world', 'potion_2m');

-- Batch text embeddings
SELECT embed(['hey', 'there', 'sunshine'], 'potion_2m');
```

Integration: The function works seamlessly with existing embedding configurations and can be used for ad-hoc comparisons, similarity calculations, and dynamic embedding generation without pre-computing data.

PR #6967, opened 2025-08-28
`embed` UDF

mach-kernel:
## 📝 Summary

```
sql> show functions like 'embed';
+---------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------+----------------------------------------------------------------------------------------------------------------------+---------------+-----------------------------------------------------------------+-------------------------+
| function_name | return_type                                                                                                                                                                                                                                                                                                          | parameters         | parameter_types                                                                                                      | function_type | description                                                     | syntax_example          |
+---------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------+----------------------------------------------------------------------------------------------------------------------+---------------+-----------------------------------------------------------------+-------------------------+
| embed         | List(LogicalField { name: "item", logical_type: LogicalType(Native(Float32), Float32), nullable: true })                                                                                                                                                                                                             | [text, model_name] | [String, String]                                                                                                     | SCALAR        | Generates embeddings for text using a specified embedding model | embed(text, model_name) |
| embed         | List(LogicalField { name: "item", logical_type: LogicalType(Native(List(LogicalField { name: "embedding", logical_type: LogicalType(Native(Float32), Float32), nullable: true })), List(LogicalField { name: "embedding", logical_type: LogicalType(Native(Float32), Float32), nullable: true })), nullable: true }) | [text, model_name] | [List(LogicalField { name: "sentence", logical_type: LogicalType(Native(String), String), nullable: true }), String] | SCALAR        | Generates embeddings for text using a specified embedding model | embed(text, model_name) |
+---------------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+--------------------+----------------------------------------------------------------------------------------------------------------------+---------------+-----------------------------------------------------------------+-------------------------+

Time: 0.058756416 seconds. 2 rows.
```

```
sql> select embed('hello world', 'potion_2m');
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| embed(Utf8("hello world"),Utf8("potion_2m"))                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  |
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| [-0.06458018, -0.15089768, -0.27086204, 0.15954897, 0.036854513, 0.2674567, 0.08301116, -0.23883809, 0.18415618, 0.17872564, -0.13338013, 0.04583888, -0.060064204, 0.02837727, 0.12276495, 0.2563953, 0.22441089, 0.18676485, -0.022696462, -0.115517, 0.26406193, -0.038585376, 0.06903323, -0.110280804, -0.22781704, 0.13314919, -0.0445487, 0.052346263, -0.1582148, -0.025428219, 0.18443377, -0.06647159, 0.15869765, 0.043279745, 0.04295943, 0.089315996, 0.06454378, 0.13101783, 0.09101013, 0.04526148, -0.08371773, -0.12598702, -0.01366845, -0.053469446, -0.13809437, -0.004018365, 0.031965364, -0.043048225, -0.035024468, 0.02014134, -0.043770872, 0.028623944, 0.0027316613, 0.028872555, 0.069406845, 0.18640453, -0.016100796, 0.081500605, 0.041663166, 0.19700463, 0.022928907, -0.068717815, 0.06768266, 0.14915158] |
+-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+

Time: 0.018293625 seconds. 1 rows.
```

```
sql> select array_length(embed(['at', 'contoso', 'we', 'never', NULL, 'up', 'our', 'schema'], 'potion_2m'));
+---------------------------------------------------------------------------------------------------------------------------------------------------+
| array_length(embed(make_array(Utf8("at"),Utf8("contoso"),Utf8("we"),Utf8("never"),NULL,Utf8("up"),Utf8("our"),Utf8("schema")),Utf8("potion_2m"))) |
+---------------------------------------------------------------------------------------------------------------------------------------------------+
| 8                                                                                                                                                 |
+---------------------------------------------------------------------------------------------------------------------------------------------------+

Time: 0.011794 seconds. 1 rows.
```

<img width="919" height="502" alt="image" src="https://github.com/user-attachments/assets/4c306f44-6902-455f-b4c4-b94264740587" />


## 🔗 Related

Delivers UDF only from #6847

## 📚 Docs

https://github.com/spiceai/docs/pull/1143


crates/runtime-datafusion-udfs/src/embed.rs
```diff
@@ -0,0 +1,414 @@
...
+
+pub static SIGNATURE: LazyLock<Signature> = LazyLock::new(|| {
+    Signature::one_of(
+        // In order of least likely to auto-coerce, via logical_expr docs for OneOf:
+        // > Coercion is attempted to match the signatures in order,
+        // > and stops after the first success, if any.
+        vec![
+            // embed(make_array(a, b, c), model_name)
+            TypeSignature::Exact(vec![
+                DataType::List(Arc::new(Field::new("sentence", DataType::Utf8, true))),
```
Jeadie, line 62: This will type coerce for `DataType::FixedSizedList(LargeUTF8...`?

sgrebnov (dismissed):

phillipleblanc (dismissed):

crates/runtime-datafusion-udfs/src/embed.rs
mach-kernel: Thanks for pointing this out. Looks like it won't: I added some tests and tried to support every combination of list/string.

Re this note: apparently if you invert the order of the type signatures the UDF breaks because DF will apparently cast your list or array into a `"[string, literal]"` 🤷‍♀️

crates/runtime-datafusion-udfs/src/embed.rs
```diff
@@ -0,0 +1,511 @@
...
+        large_list_utf8.values().append_value("hello");
+        large_list_utf8.values().append_value("world");
+        large_list_utf8.append(true);
+
+        let mut large_list_large_utf8 = LargeListBuilder::new(LargeStringBuilder::new());
+        large_list_large_utf8.values().append_value("hello");
+        large_list_large_utf8.values().append_value("world");
+        large_list_large_utf8.append(true);
+
+        let similar_values = vec![
```
mach-kernel, line 469: cc @Jeadie

crates/llms/src/embeddings/mod.rs
```diff
@@ -138,11 +136,8 @@ fn encode_embedding(format: &EncodingFormat, array: Vec<f32>) -> EmbeddingVector
 pub trait Embed: Debug + Sync + Send {
     async fn embed(&self, input: EmbeddingInput) -> Result<Vec<Vec<f32>>>;
 
-    fn embed_sync(&self, _input: EmbeddingInput) -> Result<Vec<Vec<f32>>> {
-        let name = any::type_name::<Self>();
-        Err(UnsupportedSyncInvocation {
-            model: name.to_string(),
-        })
+    fn embed_sync(&self, input: EmbeddingInput) -> Result<Vec<Vec<f32>>> {
```
mach-kernel, line 139: I wanted to be able to support async embedding models via the UDF as well. I think that this isn't too different from how it works today (other than I/O herding benefits for an entire batch instead of row-wise blocking)?

When DF 49 lands, we can revert to `UnsupportedSyncInvocation` and have the UDF call the async facade.

`xl_embed` below is an OpenAI embedding model.

```
sql> select array_length(embed(['at', 'contoso', 'we', 'never', NULL, 'up', 'our', 'schema'], 'potion_8m')) as embedding_count;
+-----------------+
| embedding_count |
+-----------------+
| 8               |
+-----------------+

Time: 0.015892 seconds. 1 rows.
sql> select array_length(embed(['at', 'contoso', 'we', 'never', NULL, 'up', 'our', 'schema'], 'xl_embed')) as embedding_count;
+-----------------+
| embedding_count |
+-----------------+
| 8               |
+-----------------+

Time: 3.368131208 seconds. 1 rows.
```

kczimm: I expected potion to be faster than that. I remember running batched embeds using intfloat/e5-small-v2 is 12ms but now I can't remember if that was GPU accelerated or not.

crates/runtime/src/builder.rs
```diff
@@ -278,6 +278,11 @@ impl RuntimeBuilder {
         }
         rt.extensions = Arc::new(RwLock::new(extensions));
 
+        // TODO: embed UDF needs access to the EmbeddingModelStore singleton, so it lives here
+        rt.df
+            .ctx
+            .register_udf(runtime_datafusion_udfs::embed::Embed::new(rt.embeds()).into());
+
```
kczimm: We have a `register_udfs` function where we catch all register our custom UDFs. You may be interested in doing this there. `runtime::datafusion::udf::register_udfs`

crates/runtime-datafusion-udfs/src/embed.rs
```diff
@@ -0,0 +1,511 @@
...
+}
+
+impl Embed {
+    #[must_use]
+    pub fn new(model_store: Arc<RwLock<EmbeddingModelStore>>) -> Self {
+        Self { model_store }
+    }
+
+    fn embed_single(
+        model: &Arc<dyn llms::embeddings::Embed>,
```
kczimm:
```suggestion
        model: &dyn llms::embeddings::Embed,
```

nit: more flexible signature

crates/runtime-datafusion-udfs/src/embed.rs
```diff
@@ -0,0 +1,511 @@
...
+        );
+
+        builder.values().append_slice(&embedding[0]);
+        builder.append(true);
+
+        Ok(ColumnarValue::Array(Arc::new(builder.finish())))
+    }
+
+    fn embed_multiple<'a>(
+        model: &Arc<dyn llms::embeddings::Embed>,
```
kczimm:
```suggestion
        model: &dyn llms::embeddings::Embed,
```

nit: more flexible signature

crates/runtime-datafusion-udfs/src/embed.rs
```diff
@@ -0,0 +1,511 @@
...
+    fn embed_single(
+        model: &Arc<dyn llms::embeddings::Embed>,
+        sentence: &str,
+    ) -> DataFusionResult<ColumnarValue> {
+        let embedding = model
+            .embed_sync(EmbeddingInput::String(sentence.to_owned()))
+            .map_err(|e| DataFusionError::External(Box::new(e)))?;
+        let vector_size = match embedding.first() {
+            Some(embedding) => embedding.len(),
+            _ => unreachable!("Should have at least one embedding"),
```
kczimm, line 106: Just to confirm, this should be truly unreachable, correct? Else we should avoid the panic.

crates/runtime/src/builder.rs
mach-kernel: I need access to either our runtime struct or the model registry in `register_udfs`. Do you see a clean way to do that?

crates/runtime-datafusion-udfs/src/embed.rs
mach-kernel: Correct

crates/runtime/src/builder.rs
kczimm: Oh right. I see we register other UDFs at the top of this function. This is probably the more appropriate place to register UDFs rather than where we do now so that UDFs can have access to the `Runtime` and `DataFusion` instances, if needed. There is something nice about having them all registered in a single spot.

mach-kernel: It looks like `register_udf` is only called by the builder, so this should be simple to do!

crates/runtime/src/datafusion/udf.rs
```diff
@@ -16,60 +16,14 @@ limitations under the License.
...
+pub fn register_udfs(runtime: &crate::Runtime) {
+    let ctx = &runtime.df.ctx;
     ctx.register_udf(alias::ScalarUDFAlias::new(Arc::new(RandomFunc::default()), "rand").into());
     ctx.register_udf(bucket::Bucket::new().into());
     ctx.register_udf(cosine_distance::CosineDistance::new().into());
     ctx.register_udf(truncate::Truncate::new().into());
-}
-
-#[cfg(test)]
-mod tests {
```
mach-kernel, line 31: ???

kczimm (approved):

PR #6969, opened 2025-08-28
Add `schema_inference_sample_count` param support for Kafka connector

### LINKED ISSUE #6860: Kafka: more resilient schema inference via configurable kafka_schema_infer_num_samples

sgrebnov:
## 📝 Summary

PR introduces a new parameter, `kafka_schema_infer_num_samples`, which allows configuring the number of Kafka messages sampled during schema inference. Increasing the sample size can improve the robustness and reliability of inferred schemas, especially in cases where data contains optional fields or varying structures.

```yaml
  - from: kafka:orders_events
    name: orders
    params:
      ...
      kafka_bootstrap_servers: localhost:9092
      kafka_security_protocol: PLAINTEXT
      schema_inference_sample_count: 10
    acceleration:
      ..
```

## 🔗 Related

Closes https://github.com/spiceai/spiceai/issues/6860


crates/runtime/src/dataconnector/kafka.rs
```diff
@@ -45,62 +45,68 @@ pub enum Error {
...
+    schema_infer_num_samples: Option<usize>,
 }
 
 impl Kafka {
     #[allow(clippy::needless_pass_by_value)]
     pub fn new(params: Parameters) -> Result<Self> {
         let kafka_config = KafkaConfig {
             brokers: params
-                .get("kafka_bootstrap_servers")
+                .get("bootstrap_servers")
```
sgrebnov, line 68: As params belong to the kafka data connector, params definition changed from `ParameterSpec::runtime` to  `ParameterSpec::component` that automatically includes connector name. Thus  there are NO changes from DX/UX perspective

crates/runtime/src/dataconnector/kafka.rs
```diff
@@ -355,36 +391,44 @@ async fn bootstrap_new_kafka_consumer(
...
+    // this is useful when some of the fields could be optional and use 'null'
+    let mut sample_values = Vec::with_capacity(schema_inference_sample_count);
+
+    for _ in 0..schema_inference_sample_count {
+        match kafka_consumer
+            .next_json::<serde_json::Value, serde_json::Value>()
+            .await
+        {
+            Ok(Some(msg)) => sample_values.push(msg),
+            Ok(None) => {
```
Jeadie, line 406: If I set `schema_inference_sample_count=100`, but the topic has not yet had 100 messages, we expect to fail?

Jeadie: Or `Ok(None)` just occurs if topic/connection is closed?

crates/runtime/tests/kafka/mod.rs
```diff
@@ -39,19 +39,34 @@ async fn kafka_sasl_connect_test() -> anyhow::Result<()> {
...
 
-            let ds = make_kafka_dataset("orders", "kafka_orders", KAFKA_PORT);
+            // Load test data for orders representing a more complex schema inference case where
+            // the first messages do not have all fields present and some contain nulls
+            let orders_schema_infer: Vec<serde_json::Value> =
+                serde_json::from_str(include_str!("./test_data/orders_schema_infer.json"))?;
+            send_messages_to_kafka(&producer, "schema_infer_test", &orders_schema_infer).await?;
+
+            let ds = make_kafka_dataset("orders", "kafka_orders", KAFKA_PORT, None);
+            let mut options = std::collections::HashMap::new();
```
Jeadie, line 58: nit: You can oneline these if you want

```suggestion
            let options: std::collections::HashMap = [("kafka_schema_infer_num_samples".to_string(), "3".to_string())].into();
```
```

crates/runtime/src/dataconnector/kafka.rs
sgrebnov: Good point @Jeadie, my understanding is that `Ok(None)` can only be reported when connection/topic is closed. Otherwise consumer/connector just waits for more data available to complete sampling / schema inference

sgrebnov: Tested/confirmed that connector will be waiting or enough sampling messages to confirm schema inference and dataset registration

Jeadie (approved):

PR #6970, opened 2025-08-28
Consistent snapshots for mongodb

krinart:
## 📝 Summary

Make mongo snapshots consistent across runs. It was done in this https://github.com/datafusion-contrib/datafusion-table-providers/pull/432 PR by sorting columns alpabetically.

This PR:
* Bump datafusion-table-providers revision
* Uses `select *` in tests instead of hardcoded columns
* Adds mongo to integration tests in github workflows


crates/test-framework/src/snapshot/snapshots/explain/test_framework__snapshot__mongodb-arrow_tpch_q10_explain.snap
```diff
@@ -8,65 +8,65 @@ description: "Query: tpch_q10"
...
 |               |                                       BytesProcessedExec                                                                                                                                                                                                                                                                 |
 |               |                                         SchemaCastScanExec                                                                                                                                                                                                                                                               |
-|               |                                           DataSourceExec: partitions=16, partition_sizes=[3, 3, 3, 3, 3, 3, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2]                                                                                                                                                                                |
```
Jeadie, line 50: Why does this change affect the number of `partition_sizes` we expect? `Select * FROM ...` isn't adding any new columns to the projection

krinart: Good question. Maybe it's due to datafusion version update? 

I was checking snapshots multiple times locally and they succeeded, so now we know they are deterministic.

sgrebnov: Can this be due to snapshots were generated on local dev machine so needs to be re-generated using this branch via Benchmark GH Action? Num partitions depends on available CPU/cores AFAIK

peasee: Yes, snapshots will need to be re-run on the actions runners as they have different CPU count for partitions.

PR #6978, opened 2025-08-29
Add xxhash support for search/sql results

krinart:
## 📝 Summary

Implemented XXH3|XXH32|XXH64|XXH128 using [`twox-hash`](https://docs.rs/twox-hash/latest/twox_hash/) behind a `xx-hash` feature flag which is disabled by default.

Benchmarking results on 2M queries [TPCH Query #17](https://github.com/spiceai/spiceai/blob/trunk/crates/test-framework/src/queries/tpch/q17.sql) on a scale factor 1:

Screenshot 2025-08-29 at 10 35 40 AM

Screenshot 2025-08-29 at 10 35 33 AM

  
Testing command:
```bash
oha -c 15 -m POST \ 
    -H 'Content-Type: text/plain' \ 
    -H 'Accept: application/json' \
    -d 'select sum(l_extendedprice) / 7.0 as avg_yearly from lineitem, part where p_partkey = l_partkey and p_brand = '\''Brand#23'\'' and p_container = '\''MED BOX'\'' and l_quantity < (select 0.2 * avg(l_quantity) from lineitem where l_partkey = p_partkey);' \
    http://localhost:8090/v1/sql \
    -n 2000000
```

Spicepod configuration:
```yaml
version: v1
kind: Spicepod
name: duckdb-acceleration-qs

datasets:
  - from: duckdb:customer
    name: customer
    params:
      duckdb_open: ./data/tpch.db
    acceleration:
      enabled: true
      engine: arrow
  - from: duckdb:lineitem
    name: lineitem
    params:
      duckdb_open: ./data/tpch.db
    acceleration:
      enabled: true
      engine: arrow
  - from: duckdb:part
    name: part
    params:
      duckdb_open: ./data/tpch.db
    acceleration:
      enabled: true
      engine: arrow
runtime:
  caching:
    sql_results:
      hashing_algorithm: siphash|ahash|xx_hash
      cache_key_type: plan|sql
```


mach-kernel (dismissed):

crates/spicepod/Cargo.toml
```diff
@@ -24,7 +24,9 @@ snafu.workspace = true
 tokio.workspace = true
 tracing.workspace = true
 url.workspace = true
+twox-hash = { workspace = true, optional = true }
```
phillipleblanc: I don't think we should feature flag the spicepod crate - it should ideally represent all possible inputs and then we handle those inputs appropriately in the runtime crates.

crates/spicepod/src/component/caching.rs
```diff
@@ -35,6 +35,8 @@ pub enum HashingAlgorithm {
     #[default]
     Siphash,
     Ahash,
+    #[cfg(feature = "xx-hash")]
```
lukekim: As discussed, this should include `xxh3`, `xxh128`, `xxh32`, `xxh64`

crates/spicepod/Cargo.toml
peasee: :heavy_plus_sign: I don't understand why this dependency even needs to be here? It's not used in `spicepod`.

crates/cache/src/lru_cache.rs
```diff
@@ -100,6 +100,30 @@ pub fn build_from_config<
             ttl,
             ahash::RandomState::default(),
         )),
+        #[cfg(feature = "xx-hash")]
+        HashingAlgorithm::XxHash3 => Arc::new(SimpleCache::new(
+            cache_max_size,
+            ttl,
+            twox_hash::xxh3::RandomHashBuilder64::default(),
+        )),
```
peasee: Could be a candidate for a macro. E.g. `runtime_cache!(SimpleCache, HashingAlgorithm::XxHash3, twox_hash::xxh3::RandomHashBuilder64::default()`

crates/cache/src/lib.rs
```diff
@@ -226,6 +226,30 @@ impl QueryResultsCacheProvider {
                     ttl,
                     std::hash::RandomState::default(),
                 )),
+                #[cfg(feature = "xx-hash")]
+                HashingAlgorithm::XxHash3 => Arc::new(SimpleCache::new(
```
peasee: Why is this using `SimpleCache`? Query results should use `LruCache`.

crates/cache/src/lru_cache.rs
peasee: also, this should use `LruCache`.

krinart:
Benchmarking results on 2M queries [TPCH Query #17](https://github.com/spiceai/spiceai/blob/trunk/crates/test-framework/src/queries/tpch/q17.sql) on a scale factor 5:

Screenshot 2025-09-08 at 12 42 43 PM

Screenshot 2025-09-08 at 12 44 56 PM



Spicepod (I switched from arrow acceleration to no acceleration):
```yaml
version: v1
kind: Spicepod
name: text-hashing

datasets:
  - from: duckdb:customer
    name: customer
    params:
      duckdb_open: ./data_sf5/tpch.db
  - from: duckdb:lineitem
    name: lineitem
    params:
      duckdb_open: ./data_sf5/tpch.db
  - from: duckdb:part
    name: part
    params:
      duckdb_open: ./data_sf5/tpch.db
runtime:
  caching:
    sql_results:
      hashing_algorithm: siphash|ahash|xxh3|xxh32|xxh64|xxh128
      cache_key_type: plan|sql
```

Testing command (same as before):
```bash
oha -c 15 -m POST \
  -H 'Content-Type: text/plain' \
  -H 'Accept: application/json' \
  -d 'select sum(l_extendedprice) / 7.0 as avg_yearly from lineitem, part where p_partkey = l_partkey and p_brand = '\''Brand#23'\'' and p_container = '\''MED BOX'\'' and l_quantity < (select 0.2 * avg(l_quantity) from lineitem where l_partkey = p_partkey);' \
  http://localhost:8090/v1/sql \
  -n 2000000
```

crates/spicepod/Cargo.toml
krinart: @peasee You are right, the `thox-hash` is not needed here, thanks!

@phillipleblanc I tried to do it the way you suggested, but I ran into an issue. 

We have 3 places where these enum values are used:
* `QueryResultsCacheProvider::try_new` (returns `Result`)
* `LruCache::build_from_config` (returns `Result`)
* `Runtime::init_caching` (returns `Arc<Caching>`, caller method returns `Runtime`)

While for first two cases I can add something like this:
```rust
_ => return UnsupportedCachingAlgorithmValueSnafu.fail(),
```

Since the last one doesn't return `Result`, it's unclear how we should handle invalid values. 

Additionally, having logic I implemented initially gives us out-of-the-box validation with proper error message mentioning valid values in case when one of xxHash values is used without xx-hash feature enable: 
```bash
WARN spiced: Unable to load spicepod data/test-app-duckdb/: Unable to parse spicepod.yaml: runtime.caching.sql_results.hashing_algorithm: unknown variant `xx_hash3`, expected `siphash` or `ahash` at line 23 column 26
```

Although there's in fact one downside to hide these values behind feature flags - they will be invisible by default to `spicepodschema`. However this can easily be addressed by adding `xx-hash` feature to `schemars` list here https://github.com/spiceai/spiceai/blob/c066878200d76371e050c6b46b891ed1802e96b3/crates/spicepod/Cargo.toml#L28-L31

crates/cache/src/lib.rs
krinart: Thanks!

crates/cache/src/lru_cache.rs
krinart: Changed to `LruCache`, thanks!

Regarding macro, here's what I came up with:
```rust
runtime_cache!(LruCache,
    cache_max_size,
    ttl,
    twox_hash::xxh3::RandomHashBuilder64::default()
)
```

Not sure I understand the benefit, it seems like it will only saves us not having to write `Arc::new`, right?

peasee: You could drop the `cache_max_size` and `ttl` as macro inputs, and just reference them directly in the macro as we know they'll exist. I would also exclude the `::default()` call in the macro input as the macro could also make that call. Then, you end up with:

```rust
runtime_cache!(LruCache, RandomHashBuilder64)
```

Also, I had suggested including `HashingAlgorithm::XxHash3` because I think we could have the macro return the entire enum variant arm like:

```rust
macro_rules! runtime_cache {
 ($algo:expr) => {
    $algo => {
      // .. everything else in the match arm ..
    }
  }
}
```

crates/spicepod/Cargo.toml
peasee: I think I understand. The trouble is that the spicepod has all enum variants, but if the feature is disabled the underlying runtime crate won't match on them. For `init_caching` we could do something similar to what we already do when we encounter an error in caching init:

```rust
Err(e) => {
                    in_tracing_context(|| {
                        tracing::error!("Failed to initialize results cache: {e}");
                    });
                }
```

We could use this to do something like `_ => in_tracing_context(|| tracing::error!("Invalid hashing_algorithm ..."))`

crates/cache/src/lru_cache.rs
krinart: It seems like Rust macros expand at the token level, and we can't use a macro to generate match arms like that:
```rust

#[macro_export]
macro_rules! runtime_cache {
    ($algo:path, $cache_type:ident, $hash_builder:path) => {
        $algo => {
            use std::sync::Arc;
            Arc::new($cache_type::new(
                cache_max_size,
                ttl,
                <$hash_builder>::default(),
            ))
        }
    };
}

match config.hashing_algorithm {
    runtime_cache!(HashingAlgorithm::Ahash, LruCache, ahash::RandomState),
    runtime_cache!(HashingAlgorithm::Siphash, LruCache, std::hash::RandomState),
    runtime_cache!(HashingAlgorithm::XxHash3, LruCache, twox_hash::xxh3::RandomHashBuilder64),
    runtime_cache!(HashingAlgorithm::XxHash32, LruCache, twox_hash::RandomXxHashBuilder32),
    runtime_cache!(HashingAlgorithm::XxHash64, LruCache, twox_hash::RandomXxHashBuilder64),
    runtime_cache!(HashingAlgorithm::XxHash128, LruCache, twox_hash::RandomHashBuilder128),
}
```

I get error: 
```bash
error: expected identifier, found `=>`
   --> crates/cache/src/lib.rs:201:16
    |
201 |         $algo => {
    |                ^^ expected identifier
...
235 |                 runtime_cache!(HashingAlgorithm::Ahash, LruCache, ahash::RandomState),
    |                 ---------------------------------------------------------------------
    |                 |              |
    |                 |              while parsing the fields for this pattern
    |                 in this macro invocation
    |
    = note: this error originates in the macro `runtime_cache` (in Nightly builds, run with -Z macro-backtrace for more info)

error: macro expansion ignores `{` and any tokens following
   --> crates/cache/src/lib.rs:201:20
    |
201 |         $algo => {
    |                    ^
...
235 |                 runtime_cache!(HashingAlgorithm::Ahash, LruCache, ahash::RandomState),
    |                 --------------------------------------------------------------------- caused by the macro expansion here
    |
    = note: the usage of `runtime_cache!` is likely invalid in pattern context

error: `match` arm with no body
   --> crates/cache/src/lib.rs:235:17
    |
235 |                 runtime_cache!(HashingAlgorithm::Ahash, LruCache, ahash::RandomState),
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: add a body after the pattern
    |
235 |                 runtime_cache!(HashingAlgorithm::Ahash, LruCache, ahash::RandomState) => { todo!() },
    |                                                                                       ++++++++++++++
```

I spent couple hours trying to make it work. In the end I did it slightly differently.

Even though I think from user perspective the original implementation is better, because xxHash values without enabled `xx-hash` feature shows this error: 
```bash
ERROR runtime::init::caching: Failed to initialize results cache: Invalid hashing algorithm. Please refer to the documentation for supported algorithms: https://spiceai.org/docs/features/caching#choosing-a-hashing_algorithm
```

Whereas any other value will show much more user friendly message:
```bash
WARN spiced: Unable to load spicepod data/test-app-duckdb/: Unable to parse spicepod.yaml: runtime.caching.sql_results.hashing_algorithm: unknown variant `xx_hash321`, expected one of `siphash`, `ahash`, `xx_hash3`, `xx_hash32`, `xx_hash64`, `xx_hash128` at line 23 column 26
```

crates/spicepod/Cargo.toml
krinart: Moving conversation here https://github.com/spiceai/spiceai/pull/6978#discussion_r2331947841

phillipleblanc (approved):

PR #6979, opened 2025-08-29
Add integration tests for models endpoint

### LINKED ISSUE #6956: Regression in `v1/models`
## Describe the bug
```
>>> curl "http://localhost:8090/v1/models?status=true"
Invalid metadata key: %
>>> curl "http://localhost:8090/v1/models"
Invalid metadata key: %
```

Affects `spice nsql`
```
>>> spice nsql
2025/08/27 11:22:40 ERROR listing spiced models error="error decoding response: invalid character 'I' looking for beginning of value"
```

### spice, spiced, OS info

- spice version: `v1.6.0`
- spiced version: `v1.6.0`
- OS info: `uname -a`: `Darwin Jacks-MBP-2.localdomain 24.6.0 Darwin Kernel Version 24.6.0: Mon Jul 14 11:30:29 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6000 arm64`

Advayp:
## 🔗 Related
- Closes #6956

### DIFF
--- a/crates/runtime/tests/models/models_http_endpoint.rs
+++ b/crates/runtime/tests/models/models_http_endpoint.rs
@@ -1,0 +1,351 @@
+use anyhow::Error;
+use app::AppBuilder;
+use reqwest::Client;
+use runtime::{Runtime, auth::EndpointAuth, status::ComponentStatus};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use std::sync::Arc;
+
+use crate::{
+    init_tracing,
+    models::{create_api_bindings_config, openai::get_openai_model},
+    utils::{runtime_ready_check, test_request_context, verify_env_secret_exists},
+};
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
+#[serde(rename_all = "snake_case")]
+pub(crate) struct OpenAIModel {
+    id: String,
+
+    object: String,
+
+    owned_by: String,
+
+    datasets: Option<Vec<String>>,
+
+    status: Option<ComponentStatus>,
+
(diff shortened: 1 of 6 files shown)


crates/runtime/tests/models/models_http_endpoint.rs
```diff
@@ -0,0 +1,348 @@
...
+                .await
+                .map_err(anyhow::Error::from)?;
+
+            assert!(
+                response.status().is_success(),
+                "Expected 200 OK, got {}",
+                response.status()
+            );
+
+            let models_response: OpenAIModelResponse =
```
Jeadie, line 100: We prefer to use snapshots for this. You can use a [insta::assert_json_snapshot](https://docs.rs/insta/latest/insta/macro.assert_json_snapshot.html) (which is used throughout our codebase). Also you will need to use [redactions](https://insta.rs/docs/redactions/) so that the snapshots don't depend on the parts that will change between test runs.

crates/runtime/tests/models/models_http_endpoint.rs
```diff
@@ -0,0 +1,348 @@
...
+                );
+            }
+            Ok(())
+        })
+        .await
+}
+
+#[tokio::test]
+async fn test_models_http_endpoint_status_no_metadata() -> Result<(), anyhow::Error> {
+    let _tracing = init_tracing(None);
```
Jeadie: Because spinning up `spiced` instances is the longest time in some of these tests, we prefer running many tests cases for a given configuration. For example, in our search tests 
https://github.com/spiceai/spiceai/blob/716493c792dedc3ff1433896e9754688f137922b/crates/runtime/tests/models/s3_vectors.rs#L57

Then, as a mental model, a single snapshot is often a sub-test.

sgrebnov (approved):

PR #6984, opened 2025-08-29
Enable parameters to enforce the value is one of several options

Jeadie:


crates/runtime/src/parameters.rs
```diff
@@ -16,9 +16,10 @@ limitations under the License.
 
 use std::{collections::HashMap, fmt::Display, sync::Arc};
 
-use secrecy::SecretString;
+use secrecy::{ExposeSecret, SecretString};
 use snafu::prelude::*;
 use tokio::sync::RwLock;
+use tract_core::tract_data::itertools::Itertools;
```
phillipleblanc: nit: Don't use Itertools from tract_core

phillipleblanc (approved):

PR #6986, opened 2025-08-30
Add s3_vectors index support for CDC and Append streams

sgrebnov:
## 📝 Summary

PR implements index / s3_vectors support for CDC and Append streams

## 👀 Notes for Reviewers

1. `update_embedding_column_in_batch` is updated so that the index can return a batch with pre-computed embeddings. This avoids redundant re-calculation of embeddings and simplifies the logic.
2. There is the following logic that verifies that `compute_index` has the same schema that could be affected but in the general acceleration case, the original batch already includes the embedding column, so the schema is not impacted. We are keeping the schema check as-is for now. If a real use case arises where the schema changes, we can improve the validation to ensure that the original schema is a subset of the returned schema.

https://github.com/spiceai/spiceai/blob/cd36204a00ad73b2b2fe4e0438d531c34ee5826e/crates/runtime-datafusion-index/src/analyzer/index_table_scan.rs#L403-L422

4. CDC delete events: further improvement is needed so that delete events avoid triggering embedding calculation and instead remove vectors directly from the index. Currently this will results in error
>ERROR runtime::accelerated_table::refresh_task::changes: Changes stream error for cdc: External error: External error: Cannot create embedding array: no valid embeddings found to determine dimension


crates/runtime/src/embeddings/index/s3/write.rs
```diff
@@ -385,37 +385,39 @@ async fn embed_column(
 
 /// Update the embedding column in the `RecordBatch` with the computed embeddings.
 fn update_embedding_column_in_batch(
-    record: RecordBatch,
+    record: &RecordBatch,
```
phillipleblanc, line 388: Why change this to a reference? The point of having this be an owned record is that it is a pipeline that consumes the record batch and produces a new one that should be its replacement.

phillipleblanc (approved):

crates/runtime/src/embeddings/index/s3/write.rs
sgrebnov: I see 👍 - main motivation was to fix clippy warnings, understand the motivation and will update

PR #7002, opened 2025-09-01
feat(trace): add --sql flag to output SQL query for spice trace

Jeadie:

### DIFF
--- a/bin/spice/cmd/trace.go
+++ b/bin/spice/cmd/trace.go
@@ -28,20 +28,23 @@ import (
 )
 
 var (
-	// The id of the trace to provide
-	id string
+// The id of the trace to provide
+id string
 
-	// The trace_id of the trace to provide
-	trace_id string
+// The trace_id of the trace to provide
+trace_id string
 
-	// The include input flag
-	include_input bool
+// The include input flag
+include_input bool
 
-	// The include output flag
-	include_output bool
+// The include output flag
+include_output bool
 
-	// The truncation length
-	truncateLength int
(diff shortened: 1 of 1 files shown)


bin/spice/cmd/trace.go
```diff
@@ -76,6 +79,9 @@ $ spice trace ai_chat --id chatcmpl-At6ZmDE8iAYRPeuQLA0FLlWxGKNnM
 
 Include the input and truncate to 120 characters (default is 80).
 $ spice trace ai_chat --include-input --truncate=120
+
+# returns the SQL query for the trace
+$ spice trace ai_chat --sql
```
lukekim: @Jeadie can we make this `--output=sql`? Then in the future, we can do `--output=table` or `--output=csv`

bin/spice/cmd/trace.go
```diff
@@ -76,6 +79,12 @@ $ spice trace ai_chat --id chatcmpl-At6ZmDE8iAYRPeuQLA0FLlWxGKNnM
 
 Include the input and truncate to 120 characters (default is 80).
 $ spice trace ai_chat --include-input --truncate=120
+
+# returns the SQL query for the trace
+$ spice trace ai_chat --output=sql
+
+# (future) returns the trace as CSV
+$ spice trace ai_chat --output=csv
```
phillipleblanc:
```suggestion
```

We shouldn't document things that don't exist

lukekim (approved):

PR #7017, opened 2025-09-02
Fix for nested VectorScanTableProvider

krinart:
## 📝 Summary

For each index in a dataset, we currently create a nested `VectorScanTableProvider`, which unless properly unpacked will result in failed s3 vectors initialization in case of 2 and more indexes.

Not sure about the proper fix, maybe we should not create nested `VectorScanTableProvider`s. But in order to unblock Friday's demo I updated logic of unpacking them.

### DIFF
--- a/crates/runtime/src/embeddings/connector.rs
+++ b/crates/runtime/src/embeddings/connector.rs
@@ -379,7 +379,7 @@ impl DataConnector for EmbeddingConnector {
         {
             let indexed_table = Arc::new(indexed_table);
             let underlying_federated_table =
-                underlying_federated_table_for_indexed_table(&indexed_table)?;
+                underlying_federated_table_for_indexed_table(&table_provider)?;
 
             let stream = self
                 .inner_connector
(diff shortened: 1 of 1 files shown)


Jeadie:
I think this might be a duplicate of the issue addressed in #7004

krinart:
> I think this might be a duplicate of the issue addressed in #7004

I tested it in trunk without these changes - I see the same issue. So I think this change is still needed.

crates/runtime/src/embeddings/connector.rs
```diff
@@ -453,16 +453,26 @@ impl DataConnector for EmbeddingConnector {
...
+    src_table_provider: &Arc<dyn TableProvider>,
 ) -> Option<Arc<FederatedTable>> {
     #[cfg(feature = "s3_vectors")]
     {
-        let vector_scan = indexed_table
-            .underlying
+        if let Some(vector_scan) = src_table_provider
             .as_any()
-            .downcast_ref::<super::index::VectorScanTableProvider>()?;
+            .downcast_ref::<super::index::VectorScanTableProvider>()
```
phillipleblanc, line 462: Instead of doing this, we should merge indexes together in the same provider when we're creating them.

phillipleblanc: i.e. in FullTextConnector::wrap_table, we should detect if the incoming table is already an IndexedTableProvider and inject it in there (same for the vector)

phillipleblanc: This can come in a next PR

sgrebnov (approved):

PR #7024, opened 2025-09-02
Use file-based fts index

Jeadie:
## 📝 Summary
 - Full text search indexes for datasets to write index data to disk. 

### UX change
```yaml
datasets:
  - from: file:./data/quora_duplicate_questions_50000.tsv
    name: questions
    acceleration:
      enabled: true
    columns:
      - name: question1
        full_text_search:
          enabled: true
          row_id: id
          index_store: memory # Is default

  - from: file:./data/quora_duplicate_questions_50000.tsv
    name: questions_file
    acceleration:
      enabled: true
    columns:
      - name: question1
        full_text_search:
          enabled: true
          row_id: id
          index_store: file 
      - name: question2
        full_text_search:
          enabled: true
          row_id: id
          index_store: memory # Will warn, but both indexes will be on file.
```

## 📚 Docs
- [ ] Docs for spicepod reference


crates/runtime/src/component/dataset/mod.rs
```diff
@@ -677,6 +746,13 @@ impl Dataset {
     }
 }
 
+///
```
phillipleblanc:
```suggestion
```

crates/spicepod/src/semantic.rs
```diff
@@ -152,6 +152,8 @@ pub struct FullTextSearchConfig {
         skip_serializing_if = "Option::is_none"
     )]
     pub row_ids: Option<Vec<String>>,
+
+    pub mode: Mode,
```
phillipleblanc: We'll need to get this UX reviewed - I would also think we'd probably want a way to control where the FTS index lives on disk.

crates/spicepod/src/semantic.rs
```diff
@@ -22,7 +22,7 @@ use schemars::JsonSchema;
 use serde::{Deserialize, Serialize, de::Error};
 use serde_json::Value;
 
-use crate::component::embeddings::EmbeddingChunkConfig;
+use crate::{acceleration::Mode, component::embeddings::EmbeddingChunkConfig};
```
phillipleblanc: We should move this out of acceleration since we are reusing it here.

Jeadie:
Change from `datasets[].columns[].full_text_search.mode` -> `datasets[].columns[].full_text_search.index_store`

crates/runtime/src/component/dataset/mod.rs
```diff
@@ -641,6 +647,99 @@ impl Dataset {
...
+                    row_ids,
+                    index_store,
+                    index_directory,
+                }) = &c.full_text_search
+                else {
+                    return None;
+                };
+
+                if index_store.is_some_and(|is| is == IndexStore::Memory) && index_directory.is_some() {
+                    tracing::warn!("Dataset '{}' has in-memory full text search configured for column '{}', but has 'index_directory' set", self.name, c.name);
```
phillipleblanc:
```suggestion
                    tracing::warn!("Dataset '{}' column '{}' has `index_store: memory` but also sets `index_directory`. These options are mutually exclusive. Defaulting to `index_store: file`.", self.name, c.name);
```

crates/runtime/src/search/full_text/connector.rs
```diff
@@ -47,34 +55,59 @@ impl FullTextConnector {
...
+            index_path,
+            search_fields,
+            primary_key,
+        }) = dataset.full_text_search_config()
+        else {
+            return Err(DataConnectorError::InvalidConfigurationNoSource {
+                dataconnector: dataset.source().to_string(),
+                connector_component: dataset.into(),
+                message: format!(
+                    "Attempted to add full text search functionality to '{}', but configuration not available",
```
phillipleblanc, line 69: This shouldn't ever happen right? (in theory)

crates/runtime/src/component/dataset/mod.rs
phillipleblanc: It wasn't clear to me if the behavior was to use in-memory mode or file-mode. Either way we should make it clear in the warning message.

phillipleblanc: Actually looking at the code this should be `index_store`, not `mode` - the PR description should also be updated.

phillipleblanc (dismissed):

crates/runtime/src/component/dataset/mod.rs
Jeadie: Updated description and this comment. Default is memory.

crates/runtime/src/search/full_text/connector.rs
Jeadie: Correct

phillipleblanc (approved):

PR #7025, opened 2025-09-03
feat: Add HTTP query concurrency support to testoperator

peasee:
## 📝 Summary

* Adds support for concurrent HTTP and Flight queries via `testoperator` with `--http-clients`. The default behaviour is that, when specified, the `testoperator` will run requests simultaneously with Flight queries at the same `--concurrency` level. E.g:
  * A `--concurrency 8` with `--http-clients` will result in 16 effective concurrent queries, due to the additional HTTP queries.

* Adds a fix for `testoperator run vector-search` to not re-download the MTEB files if they already exist.
* Adds a `--no-error` option for load tests, to ignore errors and continue returning exit code 0 at the end of a test.


mach-kernel (dismissed):
A lockless concurrent map could help with all the mutex babysitting:

https://github.com/jonhoo/flurry

phillipleblanc (approved):

PR #7034, opened 2025-09-04
Kafka/Debezium: make poll timeout error friendlier

sgrebnov:
## 📝 Summary

It is safe to warn as rdkafka will re-establish connection/re-join the group and will continue. Still keep as warning instead of fully ignoring to inform about potential problem (slow consumer performance or connectivity).  We might also add corresponding param to increase poll interval if needed.

>Kafka poll interval exceeded for dataset '{dataset_name}': connection lost or consumer too slow. Retrying.

image

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task/changes.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task/changes.rs
@@ -138,6 +138,15 @@ impl RefreshTask {
                     }
                 }
                 Err(e) => {
+                    // rdkafka will automatically recover from poll timeout errors, so we can continue.
+                    // warning is logged to inform the user about potential performance or connectivity issues.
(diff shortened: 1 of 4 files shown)


crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -138,6 +138,15 @@ impl RefreshTask {
                     }
                 }
                 Err(e) => {
+                    // rdkafka will automatically recover from poll timeout errors, so we can continue.
+                    // warning is logged to inform the user about potential performance or connectivity issues.
+                    if data_components::kafka::is_max_poll_interval_err(&e) {
+                        tracing::warn!(
```
phillipleblanc:
```suggestion
                    if data_components::kafka::is_max_poll_interval_err(&e) {
                        tracing::debug!(
```

PR #7039, opened 2025-09-04
Only call 'list_datasets' once, after initial system/user messages

Jeadie:
## 📝 Summary
 - For LLMs with the `list_datasets` tool available (i.e. also `tools: auto`), we will call the `list_datasets` and add it to the v1/chat/completion messages internally. This is done as it helps LLMs navigate the spice runtime (i.e. know what tables it can perform SQL, search or just get table schema). 
 - Currently these two messages (assistant message requesting `"list_datasets"` and tool message with result of `"list_datasets"` call) get prepend before all messages. If the completion message list already has a `"list_datasets"`, a duplicate will be added.
 - This PR addresses that. Only a single `"list_datasets"` is ever included in the completion messages list. These two messages will be after the initial system and developer messages as well as any user message immediately after system/developer messages. 
 
#### Example: Standard flow
User message:
```json
[
    {"role":"user","content":"Hello there"},
]
```

With this PR, it becomes:
```json
[
    {"role":"system","content":"Be nice and polite"},
    {"role":"user","content":"Hello there"},
    {"role":"assistant","tool_calls":[{"id":"initial_list_datasets","type":"function","function":{"name":"list_datasets","arguments":""}}
    {"role":"tool","content":"[{\"table\":\"spice.public.s3tables_questions\",\"can_search_documents\":true,\"description\":null,\"metadata\":{}}]","tool_call_id":"initial_list_datasets"}
]
```

Currently it would be:
```json
[
    {"role":"assistant","tool_calls":[{"id":"initial_list_datasets","type":"function","function":{"name":"list_datasets","arguments":""}}
    {"role":"tool","content":"[{\"table\":\"spice.public.s3tables_questions\",\"can_search_documents\":true,\"description\":null,\"metadata\":{}}]","tool_call_id":"initial_list_datasets"}
    {"role":"system","content":"Be nice and polite"},
    {"role":"user","content":"Hello there"},
]
```


crates/runtime/src/model/tool_use.rs
```diff
@@ -542,6 +543,50 @@ pub fn combine_opt_u32(a: Option<u32>, b: Option<u32>) -> Option<u32> {
     }
 }
 
+// Ensure that `tool_messages` have been added to `messages` after all initial developer/system messages and after initial user messages (i.e. not including user messages after assistant messages).
+fn insert_initial_tools(
```
phillipleblanc, line 547: Let's add some unit tests for this.

phillipleblanc (approved):

PR #7051, opened 2025-09-05
fix: Avoid redundant search embedding

peasee:
## 📝 Summary

* Avoids embedding the search column twice during a search pipeline operation


peasee:
re-opening this with a cherry-pick

PR #7052, opened 2025-09-05
fix: Remove keywords from AI document search (#7043)

peasee:
## 📝 Summary

* Removes keywords from the document similarity tool, to avoid AI models supplying keywords that tend to filter out too many results
  * This was originally introduced before full text search for AI models to gain "full text like" abilities - but now that we have full text search, this is less necessary and tends to produce worse search results.


sgrebnov (dismissed):

crates/runtime/src/search/request.rs
```diff
@@ -74,19 +74,11 @@ pub struct SearchRequestHTTPJson {
 pub struct SearchRequestAIJson {
```
phillipleblanc: Seems like `SearchRequestAIJson` isn't required anymore? can replace with just `SearchRequestBaseJson`

phillipleblanc (commented):
This change makes sense to me, but I think we should get feedback from @Jeadie before merging.

Jeadie (approved):

PR #7053, opened 2025-09-05
fix: Avoid redundant search embedding

peasee:
## 📝 Summary

* Avoids embedding the search column twice during a search pipeline operation

### DIFF
--- a/crates/runtime/src/search/candidate/vector.rs
+++ b/crates/runtime/src/search/candidate/vector.rs
@@ -353,13 +353,13 @@ impl CandidateGeneration for VectorGeneration {
         addition_projection: &[&Expr],
         limit: usize,
     ) -> search::generation::Result<SendableRecordBatchStream> {
-        let embedding = self
-            .embed_query(query.as_str())
-            .await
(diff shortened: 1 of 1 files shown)


crates/runtime/src/search/candidate/vector.rs
```diff
@@ -353,13 +353,13 @@ impl CandidateGeneration for VectorGeneration {
...
-            .await
-            .boxed()
-            .map_err(|e| SearchGenerationError::InternalError { source: e })?;
-
         let query = if self.is_chunked {
+            let embedding = self
+                .embed_query(query.as_str())
+                .await
+                .boxed()
+                .map_err(|e| SearchGenerationError::InternalError { source: e })?;
```
Advayp, line 361: I think we can use `.context` here instead of map_err

peasee: No, because `SearchGenerationError` originates from a different crate so we cannot export the e.g. `InternalSnafu`

sgrebnov (approved):

PR #7084, opened 2025-09-08
Kafka/Debezium: make common errors user-friendlier

sgrebnov:

### DIFF
--- a/crates/data_components/src/kafka.rs
+++ b/crates/data_components/src/kafka.rs
@@ -452,24 +452,3 @@ impl TableProvider for Kafka {
         Ok(Arc::new(EmptyExec::new(Arc::clone(&self.schema))) as Arc<dyn ExecutionPlan>)
     }
 }
-
-/// Returns `true` if the provided [`StreamError`] represents a Kafka poll interval error.
-/// This error occurs when the application's poll interval exceeds the maximum allowed by Kafka.
-/// It is generally nonfatal and often indicates that the consumer should retry or continue polling.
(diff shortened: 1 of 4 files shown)


crates/data_components/src/kafka.rs
```diff
@@ -455,3 +452,24 @@ impl TableProvider for Kafka {
...
+        return false;
+    };
+
+    matches!(
+        source,
+        rdkafka::error::KafkaError::MessageConsumption(
+            rdkafka::types::RDKafkaErrorCode::PollExceeded
+        )
+    )
+}
```
kczimm: nit

```suggestion
pub fn is_max_poll_interval_err(err: &cdc::StreamError) -> bool {
matches!(
        err,
        cdc::StreamError::Kafka(
            Error::UnableToReceiveMessage {
                source:
                    rdkafka::error::KafkaError::MessageConsumption(
                        rdkafka::types::RDKafkaErrorCode::PollExceeded
                    )
            }
        )
    )
}
```

kczimm (dismissed):

crates/data_components/src/kafka.rs
sgrebnov: Thank you @kczimm  - I've realized that this is no more used anymore and came from original PR for poll timeout below so I've just removed this logic
https://github.com/spiceai/spiceai/pull/7034

kczimm (dismissed):

krinart (approved):

PR #7090, opened 2025-09-09
Search: RRF UDTF

### LINKED ISSUE #7058: Enhancement: RRF UDTF (+ rank boosting + recency boosting)
## Goal-State/What/Result

Expose a helper UDTF for RRF queries. Currently, `trunk` has an RRF implementation that supports this:
```sql
SELECT id, title, content, fused_score
FROM rrf(
  vector_search(documents, 'machine learning algorithms'),
  text_search(documents, 'neural networks deep learning', content),
  id,    -- optional join key for optimal performance
  60.0   -- optional smoothing factor
);
```

---

#### Expand

With current support, add ability to:
- Support per-subquery boosting
- Support recency ranking

#### Option 1

Define boost parameters in Spicepod:

```yaml
datasets:
  - from: file://wiki_a2.parquet
    name: wiki_a
    acceleration:
      enabled: true
    columns:
     - name: text
       full_text_search:
          enabled: true
          rank_boost: 10
          recency_boost: 1
          recency_boost_column: foo
          row_id:
            - id
       vector_search:
          rank_boost: 5
          recency_boost: 20
          recency_boost_column: foo
       embeddings:
         - from: potion_2m
```

Pros: Easy to configure
Cons: Cannot decide on boosting parameters at invocation time (what if boost specification is contingent on input query, for the same index?)

#### Option 2

Use structs in invocation with named parameters

```sql
SELECT id, title, content, fused_score
FROM rrf(
        named_struct('query', vector_search(documents, 'machine learning algorithms'), 'rank_boost', 10, 'recency_boost', 20, 'recency_column', created_at),
        text_search(documents, 'neural networks deep learning', content),
        id,    -- optional join key for optimal performance
        60.0   -- optional smoothing factor
     );
```

Pros: per invocation query customization
Cons: Need some changes to invoke `named_struct` in the UDF to materialize arguments

This works:

```sql
sql> select named_struct('a', 5, 'b', 10);
+------------------------------------------------------+
| named_struct(Utf8("a"),Int64(5),Utf8("b"),Int64(10)) |
+------------------------------------------------------+
| {a: 5, b: 10}                                        |
+------------------------------------------------------+
```

This does not (yet):

```sql
sql> SELECT id, title, content, fused_score
FROM rrf(
    named_struct('query', vector_search(documents, 'machine learning algorithms'), 'boost', 10),
    text_search(documents, 'neural networks deep learning', content),
    id,    -- optional join key for optimal performance
    60.0   -- optional smoothing factor
 );
Query Error: Table function 'named_struct' not found
```

#### Option 3

Ordered parameter lists after queries, positional:

```sql
SELECT id, title, content, fused_score
FROM rrf(
    vector_search(documents, 'machine learning algorithms'),
    text_search(documents, 'neural networks deep learning', content),
    id,     -- optional join key for optimal performance
    60.0    -- optional smoothing factor,
    [10, 5] -- rank boosting args,
    [1, 20] -- recency boosting args
 );
```

Pros: Same as 2
Cons: Positionally dependent DDL is confusing for users

#### Option 4

Named argument expressions in the subquery search calls (via @lukekim):

```sql
SELECT id, title, content, fused_score
FROM rrf(
        vector_search(documents, 'machine learning algorithms', rank_boost => 10, recency_boost => 20, recency_column => created_at)),
        text_search(documents, 'neural networks deep learning', content),
        id,    -- optional join key for optimal performance
        60.0   -- optional smoothing factor
     );
```

Pros: Same as 2 with easier syntax + instantly know which call a boost belongs to
Cons: Relies on Scalar UDF hack (so RRF will have to lift these args out of the subquery calls, and will be the only usage of them, where otherwise they won't exist for `vector_search` and `text_search`).

---

Option 4 seems to be clearest + provide the most flexibility. Option 1 is clear and easy to use but requires review of search-related Spicepod semantics.

## Why/Purpose

- Ease of use vs manually fusing subqueries + declaring score computations.

## By When

**Done-Done:** _Target Date_

## Done-Done

- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Tests
- [ ] Documentation

## The Algorithm

- [x] Simplify.

## QA Plan

- Test queries against several federated providers to avoid column resolution bugs
- Test reranking + reordering of search subqueries

#### New Reciprocal Rank Fusion (RRF) UDTF

Added support for Reciprocal Rank Fusion (RRF) as a user-defined table function, enabling sophisticated hybrid search
capabilities that combine multiple search methods for improved relevance ranking.

Features:
- Multi-search fusion: Combine results from vector_search, text_search, and other search UDTFs in a single query
- Performance conscious: Optional user-specified join key for better performance
- Automatic joining: Falls back to on-the-fly JOIN key computation when no explicit key is provided

Usage:
```sql
SELECT id, title, content, fused_score
FROM rrf(
  vector_search(documents, 'machine learning algorithms'),
  text_search(documents, 'neural networks deep learning', content),
  id,    -- optional join key for optimal performance
  60.0   -- optional smoothing factor
)
WHERE fused_score > 0.01
ORDER BY fused_score DESC;
```

mach-kernel:
#### Usage

Implements the following UDTF. Its arguments _must_ be other UDTF calls:

```sql
select * from rrf(text_search(answer, 'pointers'), vector_search(answer, 'pointers'));
```

It is varadic, so an arbitrary number of vector search and text search calls can be fused into one result set:

```sql
select * from rrf(text_search(answer, 'pointers'), vector_search(answer, 'pointers'), text_search(answer, 'c++'));
```

An optional smoothing parameter (typically `k`) can be provided after the queries:

```sql
select * from rrf(text_search(answer, 'pointers'), vector_search(answer, 'pointers'), text_search(answer, 'c++'), 50);
```

An optional join key parameter can be provided after the queries and/or smoothing parameter:

```sql
select * from rrf(text_search(answer, 'pointers'), vector_search(answer, 'pointers'), text_search(answer, 'c++'), 50, id);
```

In the case that the join key is missing, a useful error is surfaced:

```sql
sql> select id, title, fused_score from rrf(text_search(wiki_a_potion, 'apple computer'), vector_search(wiki_a_potion, 'apple computer'), missing) order by fused_score desc;
Query Error: rrf: Cannot resolve column missing when fusing results
```

image

EXPLAIN SELECT...

image

## 📚 Docs

https://github.com/spiceai/docs/pull/1142

## 👀 Notes for Reviewers

- All subqueries make an assumption about the presence of a `score` column.
- Rows are hashed with MD5 to generate a JOIN key for results (if user does not provide their own)
- Due to DF limitations, scalar stubs are registered for `vector_search` and `text_search` to prevent the planner from erasing the argument expressions to the `rrf` call.
  - Try removing the scalar stubs and inspecting the plan before reaching out with a question about this -- it'll save everyone a lot of time.

#### Future work

DataFusion needs first-class support for UDTF invocations at the logical plan layer, outside of being a "special" table scan. This will enable us to remove the "scalar stub" hack, and also let us implement this as a "pure" macro. For example, we _cannot_ support this kind of invocation today without changes to DF:

```sql
with query_a as (...),
query_b as (...),
rrf(query_a, query_b)
```

This is something that we will need to support, for things like users BYO materialized ranked result sets not via UDTF invocation. CTEs are commonly used.


crates/runtime/src/search/rrf.rs
```diff
@@ -0,0 +1,571 @@
...
+            .iter()
+            .map(|sqlexpr| format!("select * from {}", sqlexpr.to_string()))
+            .collect::<Vec<_>>();
+
+        let search_df_futures: Vec<_> = search_df_queries
+            .iter()
+            .map(|sql| self.session_context.sql(sql))
+            .collect();
+
+        let search_dfs: Vec<DataFrame> = task::block_in_place(move || {
```
mach-kernel: I personally think it's fine to use `block_in_place`. An alternative would be to decorate DF to this struct as a boxed and pinned future. But I think there is a lot of send/sync headache with doing that vs blocking to evaluate (i.e. you're going to do so anyway)

crates/runtime/src/search/rrf.rs
```diff
@@ -0,0 +1,571 @@
...
+                    "stub".to_string(),
+                ))))
+            }),
+        );
+
+        Expr::ScalarFunction(ScalarFunction::new_udf(Arc::new(stub_udf), vec![]))
+    }
+
+    #[test]
+    fn test_parse_argument_exprs() {
```
mach-kernel, line 573: This tests that we reify our `rrf(...)` calls correctly

crates/runtime/src/search/rrf.rs
```diff
@@ -0,0 +1,571 @@
...
+        rt.df
+            .ctx
+            .register_table("foo", test_table)
+            .expect("Failed to register foo table");
+
+        register_udfs(&rt);
+        Ok(rt)
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
```
mach-kernel, line 530: This tests rrf with multiple subqueries. It only uses `vector_search` calls since the test setup is easier. We need `multi_thread` so we can block during planning.

crates/runtime/src/search/rrf.rs
```diff
@@ -0,0 +1,571 @@
...
+        let rank_expr = row_number()
+            .order_by(vec![col("score").sort(false, false)])
+            .build()?
+            .alias("rank");
+
+        df.window(vec![rank_expr])
+    }
+
+    // Create an internal row ID by hashing all pieces of the row
+    fn with_rrf_rowid(df: DataFrame) -> Result<DataFrame> {
```
mach-kernel, line 312: This makes an internal JOIN key we later drop from the final projection. Columns here are sorted by name, so that if projections have the same columns they'll be hashed in the same order.

crates/runtime/src/search/rrf.rs
```diff
@@ -0,0 +1,571 @@
...
+            // The original RRF paper uses 60 as its default smoothing parameter
+            60.0
+        };
+
+        // Unparse UDTF invocations
+        let unparser = Unparser::new(&DefaultDialect {});
+        let search_udtf_exprs: Vec<SqlExpr> = args
+            .iter()
+            .filter_map(|expr| match expr {
+                e @ Expr::ScalarFunction(_) => Some(unparser.expr_to_sql(&e)),
```
mach-kernel: This is the "scalar stub hack". Try removing the scalar stubs. You'll see that your UDTF then gets no argument exprs.

crates/runtime/src/search/rrf.rs
mach-kernel: Benchmarks:
```sql
sql> select count(*) from wiki_a_potion;
+----------+
| count(*) |
+----------+
| 10000    |
+----------+
```

Hash on:

```sql
select id, title, fused_score from rrf(text_search(wiki_a_potion, 'apple computer'), vector_search(wiki_a_potion, 'apple computer')) order by fused_score desc;

Time: 0.771619709 seconds. 500/10000 rows displayed.
```

Existing key same query:

```
Time: 0.354399125 seconds. 500/10000 rows displayed.
```

I think this warrants a BYO PK argument. I'll noodle on that.

mach-kernel: OK sorted. You can now provide a column ref in the argument to opt out of the auto-hashing.

Jeadie:
> sql> select id, title, fused_score from rrf(text_search(wiki_a_potion, 'apple computer'), vector_search(wiki_a_potion, 'apple computer'), missing) order by fused_score desc;
Query Error: rrf: Cannot resolve column missing when fusing results

This looks old. Am I correct in saying it will now be?
```
rrf: Cannot resolve column title when fusing results
```

From: 
https://github.com/spiceai/spiceai/pull/7090/files#diff-c82c4b20cb66f40e802a67690cb1974ddf11a9399be77640140790a51f06bef8R291

mach-kernel:
> > sql> select id, title, fused_score from rrf(text_search(wiki_a_potion, 'apple computer'), vector_search(wiki_a_potion, 'apple computer'), missing) order by fused_score desc;
> > Query Error: rrf: Cannot resolve column missing when fusing results
> 
> This looks old. Am I correct in saying it will now be?
> 
> ```
> rrf: Cannot resolve column title when fusing results
> ```
> 
> From: https://github.com/spiceai/spiceai/pull/7090/files#diff-c82c4b20cb66f40e802a67690cb1974ddf11a9399be77640140790a51f06bef8R291

End of the arg list!

sql> select id, title, fused_score from rrf(text_search(wiki_a_potion, 'apple computer'), vector_search(wiki_a_potion, 'apple computer'), **missing**) order by fused_score desc;

crates/runtime/src/embeddings/udtf.rs
```diff
@@ -354,6 +364,29 @@ impl TableFunctionImpl for VectorSearchTableFunc {
     }
 }
 
+/// This is a stub implementation, so that we can nest UDTF function invocations
+impl ScalarUDFImpl for VectorSearchTableFunc {
```
Jeadie, line 365: nit: You could use `datafusion_expr::create_udf` macro, but then `scalar_invocation_error` would only be on `invoke_with_args` and not `return_type`.

crates/runtime/src/search/rrf.rs
```diff
@@ -0,0 +1,624 @@
...
+use std::any::Any;
+use std::fmt::Debug;
+use std::sync::{Arc, LazyLock};
+use tokio::runtime::Handle;
+use tokio::task;
+
+pub static RRF_UDF_NAME: &str = "rrf";
+pub static DOCUMENTATION: LazyLock<Documentation> = LazyLock::new(|| Documentation {
+    doc_section: DocSection::default(),
+    description: "Merge and re-rank several search queries into one result set".to_string(),
```
Jeadie:
```suggestion
    description: "Merge and rank several search queries into a single result set solely considering the order and not score of the input search queries".to_string(),
```

crates/runtime/src/search/rrf.rs
Jeadie: What stops us from parsing args as `SqlExpr::Function(...` And then from that finding the associated UDTFs, then constructing a `LogicalPlan` from these? That could avoid the whole async issue, but also allow IIRC `ViewTable`s could be optimised in logical planner?

Jeadie: Otherwise aren't we doing these block_in_place at planning, and not execution?

crates/runtime/src/search/rrf.rs
```diff
@@ -0,0 +1,624 @@
...
+
+    fn documentation(&self) -> Option<&Documentation> {
+        Some(&*DOCUMENTATION)
+    }
+}
+
+impl TableFunctionImpl for ReciprocalRankFusion {
+    fn call(&self, args: &[Expr]) -> Result<Arc<dyn TableProvider>> {
+        let rrf_args = ReciprocalRankFusionArgs::from_udtf_exprs(args)?;
+        let rerank_and_fuse_df = self.rerank_and_fuse_df(&rrf_args)?;
```
phillipleblanc, line 361: This will unfortunately end up blocking a Tokio executor thread. While the top-level `statement_to_plan` is async, the `query.statement_to_plan(statement)` that ends up calling this is not async. https://github.com/apache/datafusion/blob/main/datafusion/core/src/execution/session_state.rs#L474

One option would be to replace the UDTF call with a stub that defers the actual computation until the query is executed. (i.e. it logically encapsulates the work that `self.rerank_and_fuse_df(&rrf_args)?` is doing, but just executes it when ExecutionPlan.execute is called at query time)

We could also try the optimizer route that does the (admittedly ugly) hack of searching for a TableScan with the expected name and replacing them with the UDTF calls of the inner functions.

mach-kernel: Big thanks to @Jeadie for:

> And then from that finding the associated UDTFs, then constructing a LogicalPlan from these?

I found that you can look up a UDTF from `SessionContext` which allowed me to go into a DataFrame without any need for `async`!

---

DF seems to require `TableProvider::schema` is respected as the output of this tablescan node. If you check out `7b96c00b74469e2dd3aadd3e6d25742da985e04d` from this branch, there should be an `AnalyzerRule` (analyzer rules can change schemas in logical plans, optimizers cannot?) you can try.

I wanted to use the DataFrame DSL to generate the query plan (it reads easier), but I saw that even with replacing the scan, I would get obscure "schema mismatch" errors during execution.

crates/runtime/src/search/rrf.rs
mach-kernel: > What stops us from parsing args as `SqlExpr::Function(...` And then from that finding the associated UDTFs, then constructing a `LogicalPlan` from these? That could avoid the whole async issue, but also allow IIRC `ViewTable`s could be optimised in logical planner?

I would much prefer this, too.

With the UDTF API, we are already dealing with `datafusion::logical_expr::Expr` nodes. So the SQL parsing is already done. AFAICT the default `SqlToRel` implementation expects UDTF arguments to be scalar values so by the time we get DF logical plan nodes, the arguments are dropped (without scalar stub registration).

To make this work we'd have to: preserve nested UDTF calls in the AST, preserve subquery nodes (for the `WITH query as...` case), where each of those would have subplans. I think that means a `SqlToRel` extension but it isn't straightforward to me at the moment.

mach-kernel: > Otherwise aren't we doing these block_in_place at planning, and not execution?

Yes, this is bad. I went digging and found `SessionContext::table_function`, which gave a non async path to making a DF from the UDTF argument nodes.

phillipleblanc (dismissed):

crates/runtime/src/search/rrf.rs
```diff
@@ -0,0 +1,631 @@
...
+    fn scalar_stub_error<T>() -> Result<T, DataFusionError> {
+        exec_err!(
+            "{RRF_UDF_NAME} is a table function with a scalar stub. Please call as a table function."
+        )
+    }
+
+    fn empty_df(&self) -> DataFrame {
+        let schema = Schema::new(vec![Field::new("dummy", DataType::Int64, false)]);
+
+        let empty_batch = RecordBatch::new_empty(Arc::new(schema));
```
sgrebnov: Can this be just `RecordBatch::new_empty(Arc::new(Schema::empty()));`? Do we need `dummy` field/column

crates/runtime/src/search/rrf.rs
```diff
@@ -0,0 +1,631 @@
...
+    fn with_rrf_rowid(df: DataFrame) -> Result<DataFrame> {
+        let bin_columns: Vec<Expr> = df
+            .schema()
+            .columns()
+            .iter()
+            .sorted_by_key(|c| c.name())
+            // Don't hash embeddings or scores
+            .filter_map(|c| match c.name() {
+                "score" => None,
+                name if name.contains("embedding") => None,
```
sgrebnov: Might be more accurate and robust to use `ends_with("_embedding")`

sgrebnov (dismissed):

sgrebnov:
Nit: we might add some debug tracing to simplify future troubleshooting

phillipleblanc (dismissed):

crates/runtime/src/search/rrf.rs
mach-kernel: Makes sense!

crates/runtime/src/search/rrf.rs
mach-kernel: Oh yeah, sure can

sgrebnov (approved):

Issue #7092, opened 2025-09-09
SF100 benchmarks are failing with `Unable to open path` error


sgrebnov:
## Describe the bug

https://github.com/spiceai/spiceai/actions/runs/17581001220/job/49937125095
Unable to open path ./test/spicepods/tpch/sf100.0/accelerated/mysql-duckdb[file].yaml: No such file or directory (os error 2)

https://github.com/spiceai/spiceai/actions/runs/17581001220/job/49937125095
Unable to open path ./test/spicepods/tpch/sf100.0/accelerated/indexes/mysql-duckdb[file]-indexes.yaml: No such file or directory (os error 2)

PR #7098, opened 2025-09-09
Fix SF100 benchmark tests dispatch

sgrebnov:
## 📝 Summary

The `scale_factor` parameter is currently represented and read as a decimal, while paths and logic expect it to be an integer (defined as `100` in yaml configuration).
This PR updates the serialization logic so that `scale_factor` values without a fractional part are converted to integers, fixing the mismatch.

## 🔗 Related

Fixes 
 - https://github.com/spiceai/spiceai/issues/7092

### DIFF
--- a/tools/testoperator/src/args/dispatch.rs
+++ b/tools/testoperator/src/args/dispatch.rs
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 use clap::{ArgAction, Parser, ValueEnum};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 use std::path::PathBuf;
 use test_framework::{TestType, queries::QuerySet};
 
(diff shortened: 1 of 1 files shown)


krinart (dismissed):

tools/testoperator/src/args/dispatch.rs
```diff
@@ -102,10 +102,32 @@ pub struct BenchArgs {
...
+    #[serde(
+        skip_serializing_if = "Option::is_none",
+        serialize_with = "serialize_scale_factor"
+    )]
     pub scale_factor: Option<f64>,
 }
 
+#[allow(clippy::cast_possible_truncation)]
+#[allow(clippy::ref_option)]
+fn serialize_scale_factor<S>(x: &Option<f64>, s: S) -> Result<S::Ok, S::Error>
```
sgrebnov, line 114: Serializer requires this to be `&Option<f64>` instead of `Option<&f64>` so we need `clippy::ref_option`

krinart (approved):

phillipleblanc (approved):

PR #7103, opened 2025-09-10
Initial changes for SearchIndex

Jeadie:
## 📝 Summary
 - Standardise `VectorQueryTableProvider` and `TextSearchIndexProvider` into a common `SearchQueryProvider` pattern.
 - This requires moving from a `VectorIndex` to a `SearchIndex` trait to define these indexes.


crates/search/src/provider.rs
```diff
@@ -0,0 +1,481 @@
...
+            }
+        };
+
+        let columns_requested: HashSet<String> = source_schema
+            .fields()
+            .iter()
+            .map(|f| f.name().clone())
+            .collect();
+
+        let has_all_columns = search_index_columns.is_superset(&columns_requested);
```
sgrebnov, line 98: Should we return early w/o calculating `handleable_filters` if `has_all_columns` is false?

sgrebnov (dismissed):

crates/search/src/provider.rs
Jeadie: Smart

crates/runtime/src/embeddings/index/s3/index.rs
```diff
@@ -62,28 +61,52 @@ pub struct S3Vector {
...
     pub fn new(
         table: S3VectorsTable,
         embedded_column: String,
         primary_key: Vec<Field>,
         metadata_columns: MetadataColumns,
         model_name: String,
+        dimension: usize,
         embedding_models: Arc<RwLock<EmbeddingModelStore>>,
     ) -> Self {
+        // Build complete metadata columns including the embedding column as non-filterable
```
phillipleblanc, line 87: Why is this necessary? it seems weird to call it a metadata column when its just the main data for the vector index?

crates/runtime/src/embeddings/index/s3/index.rs
```diff
@@ -197,6 +197,40 @@ impl VectorIndex for S3Vector {
     }
 }
 
+impl VectorIndex for S3Vector {
+    /// Use a [`S3VectorsListTable`] and then:
+    ///   1. Convert the primary key to its appropriate name and data type
+    ///   2. Rename [`S3_VECTOR_EMBEDDING_NAME`] appropriately
+    fn list_table_provider(
+        &self,
+    ) -> Result<Arc<dyn TableProvider>, Box<dyn std::error::Error + Send + Sync>> {
```
phillipleblanc: This pattern makes me believe we should really be returning a LogicalPlan here, not a TableProvider. The LogicalPlan could encode the projection without needing to wrap the TableProvider

crates/search/src/provider.rs
```diff
@@ -0,0 +1,483 @@
...
+        _state: &dyn Session,
+        projection: Option<&Vec<usize>>,
+        filters: &[Expr],
+        limit: Option<usize>,
+    ) -> datafusion::error::Result<Arc<dyn ExecutionPlan>> {
+        let primary_key_fields = self.search_index.primary_fields();
+        // Check primary key constraints
+        if primary_key_fields.is_empty() {
+            return Err(DataFusionError::Execution(
+                "The search index was created successfully without a primary key.\n\
```
phillipleblanc:
```suggestion
                "The search index was created without a primary key.\n\
```

crates/runtime/src/embeddings/index/s3/index.rs
Jeadie: As it relates to a more generic `SearchIndex`, the embedding vector is metadata (i.e. FTS and other retrieval mechanisms don't have this concept).

crates/runtime/src/embeddings/index/s3/index.rs
Jeadie: That could be a better design. I had not yet figured out how to reconcile (eventually) `EmbeddingTable` (i.e. embeddings JIT / accelerated) with the pattern we have for index-based vector search. `LogicalPlan` makes more sense here.

crates/runtime/src/datafusion/mod.rs
```diff
@@ -1014,6 +1014,7 @@ impl DataFusion {
         }
 
         if refresh_mode == RefreshMode::Changes {
+            tracing::error!("Hello headie");
```
phillipleblanc: hello

crates/runtime/src/embeddings/udtf.rs
```diff
@@ -285,14 +283,18 @@ impl VectorSearchTableFunc {
         args: &VectorSearchTableFuncArgs,
     ) -> Result<Option<Arc<dyn TableProvider>>, DataFusionError> {
         use crate::{embeddings::index::s3::S3Vector, search::util::find_index_in_table_provider};
+        use search::{index::SearchIndex, provider::SearchQueryProvider};
```
phillipleblanc: nit, move these `use` statements to the top of the file

crates/search/src/index.rs
```diff
@@ -0,0 +1,61 @@
...
+/// [`SearchIndex::query_table_provider`], or to reduce the need for joining the [`TableProvider`]s
+///  of the search index and underlying table.
+#[async_trait]
+pub trait SearchIndex: std::fmt::Debug + Send + Sync + 'static {
+    /// The name of the column, in the underlying table, of the column for which search is performed against.
+    /// For vector indexes, this is the column that gets embedded. For FTS indexes, this is the text column being searched.
+    fn search_column(&self) -> String;
+
+    /// All [`Field`]s that define a primary key between the underlying table and the [`SearchIndex`].
+    ///
```
phillipleblanc:
```suggestion
```

crates/search/src/index.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// For FTS indexes, this may return empty metadata columns.
+    fn metadata_columns(&self) -> &MetadataColumns;
+
+    /// Update the index based on a [`RecordBatch`] from the underlying table.
+    async fn write(
+        &self,
+        record: RecordBatch,
+    ) -> Result<RecordBatch, Box<dyn std::error::Error + Send + Sync>>;
+
+    /// A [`TableProvider`] containing the [`SearchIndex::primary_fields`], additional metadata
```
phillipleblanc, line 50: Should this also return a LogicalPlan to allow implementors the flexibility to use projection nodes for adding these fields?

crates/search/src/provider.rs
```diff
@@ -0,0 +1,485 @@
...
+            Some(idx) => {
+                let projected = self
+                    .schema()
+                    .project(idx)
+                    .map_err(|e| DataFusionError::ArrowError(Box::new(e), None))?;
+                Arc::new(projected)
+            }
+        };
+
+        let final_proj = LogicalPlan::Projection(Projection::new_from_schema(
```
phillipleblanc, line 451: Instead of implementing a nested TableProvider, would just representing this computation as a projection in a LogicalPlan be cleaner?

phillipleblanc (dismissed):
I think this is looking good - I have a few nits and some general architecture questions we can ponder.

phillipleblanc (approved):

PR #7105, opened 2025-09-10
Fix `ORDER BY`: (`BytesProcessedExec` to avoid pruning ordered execs during physical optimization)

mach-kernel:
## 📝 Summary

When `BytesProcessedExec` is at the top of a query branch with children enforcing order (e.g. `SortExec`), the physical optimizer would erase order maintaining nodes for being redundant, since `BytesProcessedExec` does not `maintains_input_order`.

This PR adds a test with the simplest case of this bug (it doesn't matter that the erasure can happen to other ordered nodes, it matters that it happens), asserts that sort/order ops do not get erased during optimization, and implements `BytesProcesedExec::maintains_input_order` so that the optimizer no longer prunes sort ops.


crates/runtime/src/datafusion/extension/bytes_processed.rs
```diff
@@ -40,11 +44,51 @@ use std::{
...
+impl Stream for BytesProcessedStream {
+    type Item = Result<RecordBatch, DataFusionError>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        match self.inner.poll_next_unpin(cx) {
+            Poll::Ready(Some(Ok(batch))) => {
+                self.bytes_processed += batch.get_array_memory_size() as u64;
+                Poll::Ready(Some(Ok(batch)))
+            }
+            error @ Poll::Ready(Some(Err(_))) => error,
```
kczimm: is this matched for explicit handling of the error, even though the other match arm does the same thing?

crates/runtime/src/datafusion/extension/bytes_processed.rs
```diff
@@ -285,27 +333,106 @@ impl ExecutionPlan for BytesProcessedExec {
...
+        // Fold over the default rules to apply the same optimizations DF would at runtime
+        let optimized = optimizer
+            .rules
+            .iter()
+            .fold(Arc::clone(&final_plan), |plan, rule| {
+                rule.optimize(plan, &config).unwrap()
+            });
+
+        // No semantic eq implemented, so this is the easiest way to compare plans
+        assert_eq!(
```
mach-kernel, line 429: Run me on trunk

crates/runtime/src/datafusion/extension/bytes_processed.rs
```diff
@@ -258,6 +302,10 @@ impl ExecutionPlan for BytesProcessedExec {
         vec![false]
     }
 
+    fn maintains_input_order(&self) -> Vec<bool> {
+        vec![true; self.children().len()]
+    }
+
```
kczimm, line 307: Just for my understanding, this is the crux of the issue, right?

crates/runtime/src/datafusion/extension/bytes_processed.rs
```diff
@@ -285,27 +333,106 @@ impl ExecutionPlan for BytesProcessedExec {
...
-                        yield Err(e);
-                    }
-                }
-            }
-            crate::metrics::telemetry::track_bytes_processed(
-                bytes_processed,
-                &request_context.to_dimensions(),
-            );
-        };
+        let bytes_processed_stream = BytesProcessedStream::new(stream, request_context);
```
kczimm, line 335: It might be a bit more terse to just use `fold`:

```rust
use futures::stream::{StreamExt, Stream};

let bytes_processed_stream = {
    let request_context_clone = Arc::clone(&request_context);
    stream.fold(0u64, |bytes_processed, result| async move {
        match result {
            Ok(batch) => bytes_processed + batch.get_array_memory_size() as u64,
            Err(_) => bytes_processed,
        }
    })
    .into_stream()
    .map(move |bytes_processed| {
        crate::metrics::telemetry::track_bytes_processed(
            bytes_processed,
            &request_context_clone.to_dimensions(),
        );
        None
    })
    .chain(stream.map(|result| match result {
        Ok(batch) => Ok(batch),
        Err(e) => Err(e),
    }))
};
```

kczimm (dismissed):

crates/runtime/src/datafusion/extension/bytes_processed.rs
```diff
@@ -285,27 +333,106 @@ impl ExecutionPlan for BytesProcessedExec {
...
+        let batch = RecordBatch::try_new(
+            Arc::clone(&schema),
+            vec![Arc::new(Int64Array::from_iter_values(0i64..10000))],
+        )?;
+
+        Ok(Arc::new(MemTable::try_new(schema, vec![vec![batch]])?))
+    }
+
+    #[allow(clippy::similar_names)]
+    #[allow(clippy::unwrap_used)]
```
phillipleblanc:
```suggestion
```

I'd prefer to avoid using unwrap in the codebase, even for tests and to prefer `.expect()`

phillipleblanc (dismissed):

crates/runtime/src/datafusion/extension/bytes_processed.rs
```diff
@@ -258,6 +302,10 @@ impl ExecutionPlan for BytesProcessedExec {
         vec![false]
     }
 
+    fn maintains_input_order(&self) -> Vec<bool> {
```
sgrebnov, line 304: ❤️

sgrebnov (dismissed):

crates/runtime/src/datafusion/extension/bytes_processed.rs
mach-kernel: This is neat! Does this page the stream twice?

crates/runtime/src/datafusion/extension/bytes_processed.rs
mach-kernel: I initially thought that an Err would terminate the stream (so I had a call to emit bytes processed at this point). I don't think that's the case. Anyway! Will simplify!

crates/runtime/tests/metadata/snapshots/integration__metadata__explain_met_all_location_filtered.snap
```diff
@@ -8,8 +8,8 @@ expression: explain_met_all_location_filtered
...
 |               |     TableScan: met_all projection=[id, value, year, month, day, last_modified, location, size], full_filters=[met_all.location = Utf8("s3://spiceai-public-datasets/hive_partitioned_data/year=2023/month=2/day=2/data_1.parquet")] |
-| physical_plan | SortExec: expr=[id@0 ASC NULLS LAST, location@6 ASC NULLS LAST], preserve_partitioning=[false]                                                                                                                                      |
-|               |   BytesProcessedExec                                                                                                                                                                                                                |
+| physical_plan | BytesProcessedExec                                                                                                                                                                                                                  |
```
mach-kernel, line 11: The table scan materializes as sortexec + datasource exec as in the unit test.

crates/runtime/src/datafusion/extension/bytes_processed.rs
mach-kernel: Done!

crates/runtime/src/datafusion/builder.rs
```diff
@@ -51,18 +51,7 @@ use super::{
...
-    accelerator_engine_registry: Arc<AcceleratorEngineRegistry>,
-    memory_limit: Option<u64>,
-    temp_directory: Option<String>,
-    accelerated_refresh_semaphore: Option<Arc<Semaphore>>,
-    task_history_enabled: bool,
-    caching: Option<Arc<Caching>>,
-}
-
-pub(crate) fn get_df_default_config() -> SessionConfig {
+pub static DEFAULT_DATAFUSION_CONFIG: LazyLock<RwLock<SessionConfig>> = LazyLock::new(|| {
```
mach-kernel, line 54: https://github.com/spiceai/spiceai/issues/7118

crates/runtime/tests/integration_models.rs
```diff
@@ -34,14 +36,18 @@ pub(crate) const DEFAULT_TRACING_MODELS: Option<&str> = Some(
...
-    let state = df.ctx.state_ref();
-    let mut state_lock = state.write();
-    state_lock
-        .config_mut()
-        .options_mut()
-        .execution
-        .target_partitions = 3;
+/// 2) Disables coalesce batches and repartition joins for terser plans.
+fn configure_test_datafusion() {
+    match DEFAULT_DATAFUSION_CONFIG.write() {
```
phillipleblanc, line 41: Is the idea of modifying the global static because some test paths execute using SessionContexts that aren't derived from `runtime::datafusion::DataFusion`?

phillipleblanc (dismissed):

crates/runtime/tests/integration_models.rs
mach-kernel: Correct, but it's not limited to test scenarios (this happens when running spiced too)

mach-kernel:
Notes about snapshot changes:

- Generally we want to make assertions on the simplest possible plan for a given query.
- A lot of `CoalesceBatchesExec` have been pruned from plans with the new configuration. This I think will also have the side effect of reducing test runtimes in aggregate.
- Partition count now verbatim matches the configured `target_partitions` number.

Jeadie (approved):

PR #7108, opened 2025-09-11
Fix spiced_docker workflows for new actions/download-artifact@v5 behavior

phillipleblanc:
## 📝 Summary

`actions/download-artifact@v5` has a breaking change in how artifacts that are downloaded by pattern are placed onto the filesystem. The `v4` behavior was to place each artifact its its own sub-folder with the artifact in that subfolder, i.e.:

```yaml
      - name: Download all artifacts
        uses: actions/download-artifact@v5
        with:
          path: /tmp
          pattern: images-*
```

With the artifacts `images-amd64` and `images-arm64` that both contained a single `images-amd64.tar` and `images-arm64.tar` file would be placed onto the filesystem like:

```bash
images-amd64/images-amd64.tar
images-arm64/images-arm64.tar
```

Now they are placed onto the filesystem at the root:

```bash
images-amd64.tar
images-arm64.tar
```

Ref: https://github.com/actions/download-artifact/releases/tag/v5.0.0

This PR also includes a small change to make the MongoDB integration tests more resilient in CI by waiting longer for the container to start and become healthy.

### DIFF
--- a/.github/workflows/spiced_docker.yml
+++ b/.github/workflows/spiced_docker.yml
@@ -277,8 +277,6 @@ jobs:
       - name: Verify artifacts
         run: |
           ls -l /tmp
-          ls -l /tmp/images-amd64
-          ls -l /tmp/images-arm64
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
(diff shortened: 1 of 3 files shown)


Jeadie (approved):

phillipleblanc:
Related to: https://github.com/actions/download-artifact/issues/419

sgrebnov (approved):

lukekim (approved):

Issue #7109, opened 2025-09-11
Zero-vector can stop the ingestion to S3 Vectors


phillipleblanc:
## Describe the bug

When a vector with zero norm (i.e. [0, 0, 0, 0, ...]) is sent to S3 Vectors via PutVectors, an error is returned:

`2025-09-11T03:29:47.236681Z  WARN  runtime::accelerated_table::refresh_task: Failed to load data for dataset test (postgres): Failed to get data from the connector. External error: External error: Cannot write to 's3_vector_index' index: Failed to write vectors to S3 Vectors. ValidationException: Invalid record for key '111111': cosine distance does not support vectors with zero norm Ensure the dataset configuration is valid, and try again.`

PR #7110, opened 2025-09-11
Add filter for zero vectors before writing to S3 Vectors

phillipleblanc:
## 📝 Summary

Filters out zero vectors (i.e. where each dimension is `0.0`) before writing to S3 Vectors, as AWS will return a fatal error that prevents Spice from completing the acceleration.

I'm not sure where we would get a zero vector from - if the embedding model generates this or whether its a further bug in Spice, but at least with this PR we won't completely break the acceleration when it happens.

## 🔗 Related

- Mitigates #7109


ewgenius (dismissed):

mach-kernel (approved):
Failing the request seems a bit much on their end, but LGTM

PR #7122, opened 2025-09-11
Enable unified traceparent override support for HTTP API

sgrebnov:
## 📝 Summary

Generalize traceparent override logic and apply it to NSQL and SQL API.

## 👀 Notes for Reviewers

Support for SQL via Flight API will be added next

### DIFF
--- a/crates/runtime/src/http/mod.rs
+++ b/crates/runtime/src/http/mod.rs
@@ -39,7 +39,7 @@ use crate::{
 pub use routes::get_api_doc;
 mod metrics;
 mod routes;
-mod traceparent;
+pub mod traceparent;
 
 pub mod v1;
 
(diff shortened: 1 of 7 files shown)


crates/runtime/src/http/mod.rs
```diff
@@ -39,7 +39,7 @@ use crate::{
 pub use routes::get_api_doc;
 mod metrics;
 mod routes;
-mod traceparent;
+pub mod traceparent;
```
Jeadie, line 42: nit

```suggestion
pub(crate) mod traceparent;
```

Jeadie (approved):

phillipleblanc (approved):

crates/runtime/src/http/mod.rs
sgrebnov: I'll update this next - thank you

PR #7142, opened 2025-09-14
Tune retry settings for Kafka/Debezium connectors

### LINKED ISSUE #7086: Kafka/Debezium Data Connectors use incremental retry when reading messages

sgrebnov:
## 📝 Summary

Tune `librdkafka` client backoff to reduce rapid retries during outages, lowering the risk of hammering or overwhelming brokers, DNS, or other infrastructure during failover. The change increases the base interval from `100ms` to `1000ms` and raises the maximum backoff from `1s` to `30s`, providing a smoother retry cadence and reducing trace volume in the output

https://docs.confluent.io/platform/current/installation/configuration/producer-configs.html

## 🔗 Related

Closes https://github.com/spiceai/spiceai/issues/7086

### DIFF
--- a/crates/data_components/src/kafka.rs
+++ b/crates/data_components/src/kafka.rs
@@ -265,6 +265,11 @@ impl KafkaConsumer {
         config
             .set("group.id", group_id.clone())
             .set("bootstrap.servers", &kafka_config.brokers)
+            .set("retry.backoff.ms", "1000")
+            .set("retry.backoff.max.ms", "30000")
+            .set("reconnect.backoff.ms", "1000")
+            .set("reconnect.backoff.max.ms", "30000")
+            .set("debug", "broker,cgrp,fetch")
             // For new consumer groups, start reading at the beginning of the topic
             .set("auto.offset.reset", "smallest")
             // Commit offsets automatically


crates/data_components/src/kafka.rs
```diff
@@ -265,6 +265,11 @@ impl KafkaConsumer {
         config
             .set("group.id", group_id.clone())
             .set("bootstrap.servers", &kafka_config.brokers)
+            .set("retry.backoff.ms", "1000")
```
phillipleblanc, line 268: In addition to this, should we allow tuning these via parameters on the dataset config?

phillipleblanc (approved):

crates/data_components/src/kafka.rs
sgrebnov: Yeah, I like it - created the following item for this:
 
- https://github.com/spiceai/spiceai/issues/7164

Make Kafka Consumer Important Settings configurable:
 - retry setting such us `retry.backoff.ms`, etc
 - poll and threads behavior: https://learn.conduktor.io/kafka/kafka-consumer-important-settings-poll-and-internal-threads-behavior/

PR #7166, opened 2025-09-15
Flatten nested CoalesceBatchesExec nodes

### LINKED ISSUE #7147: Duplicate coalesce batch
In upgrade to [Datafusion 49](https://github.com/spiceai/spiceai/issues/7078), duplication of `CoalesceBatchesExec` has started for embedding table. 

https://github.com/spiceai/spiceai/blob/235e7f14d81bf586d754514f3f66536509268209/crates/runtime/src/embeddings/index/snapshots/runtime__embeddings__index__tests__scan_table_join_for_projection.snap#L13-L14

Discussed [here](https://github.com/spiceai/spiceai/pull/7101#discussion_r2347663340).

mach-kernel:
## 📝 Summary

In DF, creating a physical plan is coupled to optimizing it:
https://github.com/apache/datafusion/blob/685f5d6c07868c8e47d6ffd11fb8134ca6aa405d/datafusion/core/src/physical_planner.rs#L196

When running EXPLAIN, the physical plan is optimized before display. But we also create a physical plan in our table providers beforehand, so it's possible to do more than one physical planning optimization pass. The default CoalesceBatchesExec optimizer blindly decorates itself as the parent of nodes matching its predicate:
https://github.com/apache/datafusion/blob/685f5d6c07868c8e47d6ffd11fb8134ca6aa405d/datafusion/physical-optimizer/src/coalesce_batches.rs#L63-L66

This PR implements an optimizer that flattens repeated `CoalesceBatchExec` nodes.

Resolves #7147


crates/runtime/src/embeddings/index/snapshots/runtime__embeddings__index__tests__scan_table_join_for_projection_metadata_multiple_pk.snap
```diff
@@ -1,7 +1,6 @@
 ---
 source: crates/runtime/src/embeddings/index/mod.rs
 expression: "format!(\"{}\", pretty::pretty_format_batches(&col).map_err(|e| e.to_string())?)"
-snapshot_kind: text
```
mach-kernel, line 4: This plan cannot be flattened because the `CoalescePartitionExec`s have different parameters (collapse the rest of the file to see).

crates/runtime/src/datafusion/physical_optimizer/flatten_coalesce.rs
```diff
@@ -0,0 +1,133 @@
...
+use datafusion::physical_plan::ExecutionPlan;
+use datafusion::physical_plan::coalesce_batches::CoalesceBatchesExec;
+use std::sync::Arc;
+
+#[derive(Debug)]
+pub struct FlattenCoalesce {}
+
+impl FlattenCoalesce {
+    fn coalesce_batches_exec_node(node: &Arc<dyn ExecutionPlan>) -> Option<CoalesceBatchesExec> {
+        node.as_any().downcast_ref::<CoalesceBatchesExec>().cloned()
```
mach-kernel, line 29: Does anyone know how to do `Arc<dyn ExecutionPlan> -> Arc<CoalesceBatchesExec> ` without cloning? `into_any_arc` won't work because of unsatisfied trait bounds.

crates/runtime/src/datafusion/builder.rs
```diff
@@ -257,6 +260,14 @@ pub fn get_analyzer_rules() -> Vec<Arc<dyn AnalyzerRule + Send + Sync>> {
     ]
 }
 
+/// Custom Spice physical optimizer rules
+#[must_use]
+pub fn get_physical_optimizer_rules() -> Vec<Arc<dyn PhysicalOptimizerRule + Send + Sync>> {
+    let mut rules = PhysicalOptimizer::new().rules;
```
mach-kernel, line 266: Unless you explicitly do this, calling `with_physical_optimizer_rules` isn't "with defaults + ours".

crates/runtime/tests/refresh_sql/refresh_max_timestamp_df.rs
peasee, line 1: bad merge?

phillipleblanc (commented):
I don't think this is the right solution - it is basically working around a bug in DataFusion. I checked and there is an existing DataFusion bug that describes this behavior we're seeing: https://github.com/apache/datafusion/issues/15938 - so I think we should shelve this PR for now.

mach-kernel:
> I checked and there is an existing DataFusion bug that describes this behavior we're seeing: https://github.com/apache/datafusion/issues/15938

I read through the thread for the bug you linked; their failure mechanism is different from ours. They're looking at how the ordering of the optimizer rules affect the resultant plan. This is not the case here.

> It is basically working around a bug in DataFusion

I am having trouble with this part. Pull `mach/duplicate-coalesce-repeat-opt` and look at https://github.com/spiceai/spiceai/commit/9798bee32d54d18ffa439d3325eb5ac8954a2bab#diff-f942e4cf6295830cef4c4a2777f531ea242f4d4398164f7e5f6d673b71e07b2cR146-R176. Running this gives you:

```
Optimization pass 0
CoalesceBatchesExec: target_batch_size=8192
  HashJoinExec: mode=Partitioned, join_type=Full, on=[(foo@0, foo@0)]
    EmptyExec
    EmptyExec

Optimization pass 1
CoalesceBatchesExec: target_batch_size=8192
  CoalesceBatchesExec: target_batch_size=8192
    HashJoinExec: mode=Partitioned, join_type=Full, on=[(foo@0, foo@0)]
      EmptyExec
      EmptyExec

Optimization pass 2
CoalesceBatchesExec: target_batch_size=8192
  CoalesceBatchesExec: target_batch_size=8192
    CoalesceBatchesExec: target_batch_size=8192
      HashJoinExec: mode=Partitioned, join_type=Full, on=[(foo@0, foo@0)]
        EmptyExec
        EmptyExec
```

> I don't think this is the right solution

I agree in a "big picture" way: we should generate the plans correctly without postprocessing. But I really don't think I can say this is a DataFusion bug; the physical optimizer's defined behavior is to work this way. 

For example we could "fix" this (in DataFusion) the same way you fixed `BytesProcessedNode` to avoid wrapping itself. But I am sure that this would have occurred to the DF folks?

Our code happens to be invoking the physical optimizer more than once. The alternative route would be to roll our own `impl PhysicalPlanner` that yields unoptimized physical plans where we let DF run the optimization pipeline (or we remember to run it). However that's a more involved change.

This approach was chosen because of its severability: once we have the time for a more mature fix, this gets easily rm'd and pulled out of the ruleset. The decision remains yours and the team's, but I think that the extra stages have a performance impact and so it isn't an awful idea in the meantime.

Issue #7100, opened 2025-09-10
Enhancement: CDC Streams process changes in batches


phillipleblanc:
## Goal-State/What/Result

CDC Streams process changes in batches rather than one at a time, bringing significant speedups for processing changes.

## Why/Purpose

Processing a lot of changes (i.e. initial changes) is untolerably slow without this.

## By When

**Issue/Spec written and reviewed:** _Target Date_
**Done-Done:** _Target Date_

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Telemetry / Metrics / Task History
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

Initial attempts at this are here:
- https://github.com/spiceai/spiceai/pull/7021
- https://github.com/spiceai/spiceai/pull/7031
- https://github.com/spiceai/spiceai/pull/7032
- https://github.com/spiceai/spiceai/pull/7037
- https://github.com/spiceai/spiceai/pull/7038

To make these production ready, we need to:
- Allow configuration of the batch size (right now its hard-coded to 10k)
- Instead of the `ChangesStream` returning one item at a time, it should use the `chunk_timeout` helper on `tokio-stream` to batch together X changes into a single RecordBatch, along with a combined ChangeCommit that will fan out the commit to all of the included changes.
- Test whether committing the last change removes the need to commit the previous ones - if so, then encode that optimization into the batch stream by replacing the ChangeBatch to be a RecordBatch with X items, and the ChangeCommit to be the last one of the X items.
- Instead of de-duplicating records and favoring the last one, we should split up batches such that a row with the same primary key never appears more than once in a batch.
  - This will ensure that all changes are visible and none are skipped, which may be important for some connectors/changes. We may consider an optimization to "throw away" previous versions of a row if they appear multiple times in the same batch, but there is no guarantee from the interface that the latest version is sufficient.
- Add unit/integration tests
- Add metrics/task history entries
[...]

PR #7168, opened 2025-09-15
Speed up full text indexing/embeddings computation on change streams

sgrebnov:
## 📝 Summary

Propagate into trunk
 - https://github.com/spiceai/spiceai/pull/7021/

### DIFF
--- a/crates/runtime/src/embeddings/connector.rs
+++ b/crates/runtime/src/embeddings/connector.rs
@@ -14,25 +14,22 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 use crate::accelerated_table::AcceleratedTable;
+use crate::changes::embed_change_envelope;
+use crate::changes::flatten_change_envelope_stream;
+use crate::changes::index_change_envelope;
 use crate::component::ComponentInitialization;
 use crate::component::dataset::Dataset;
 use crate::component::metrics::MetricsProvider;
 use crate::dataconnector::DataConnector;
 use crate::dataconnector::DataConnectorError;
 use crate::dataconnector::DataConnectorResult;
-use crate::embeddings::execution_plan::compute_additional_embedding_columns;
-use crate::embeddings::execution_plan::construct_record_batch;
 use crate::federated_table::FederatedTable;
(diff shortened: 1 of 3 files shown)


crates/runtime/src/embeddings/connector.rs
```diff
@@ -424,13 +359,14 @@ impl DataConnector for EmbeddingConnector {
...
 
-            let stream = self
-                .inner_connector
-                .append_stream(underlying_federated_table)?
-                .then(move |item| Self::index_change_envelope(item, Arc::clone(&indexed_table)))
-                .boxed();
+            let stream = Box::pin(
+                self.inner_connector
+                    .append_stream(underlying_federated_table)?
+                    .chunks_timeout(100, Duration::from_secs(2))
```
peasee, line 365: should we add these as parameters for change stream connectors?

peasee (approved):

phillipleblanc (commented):
We should not bring in this change as is, there is a lot of extra work that needs to be done tracked here: https://github.com/spiceai/spiceai/issues/7100

> To make these production ready, we need to:
> - Allow configuration of the batch size (right now its hard-coded to 10k)
> - Instead of the `ChangesStream` returning one item at a time, it should use the `chunk_timeout` helper on `tokio-stream` to batch together X changes into a single RecordBatch, along with a combined ChangeCommit that will fan out the commit to all of the included changes.
> - Test whether committing the last change removes the need to commit the previous ones - if so, then encode that optimization into the batch stream by replacing the ChangeBatch to be a RecordBatch with X items, and the ChangeCommit to be the last one of the X items.
> - Instead of de-duplicating records and favoring the last one, we should split up batches such that a row with the same primary key never appears more than once in a batch.
>   - This will ensure that all changes are visible and none are skipped, which may be important for some connectors/changes. We may consider an optimization to "throw away" previous versions of a row if they appear multiple times in the same batch, but there is no guarantee from the interface that the latest version is sufficient.
> - Add unit/integration tests
> - Add metrics/task history entries
> - Documentation for the batch size parameter

phillipleblanc (changes requested):
This should be done as part of https://github.com/spiceai/spiceai/issues/7100

PR #7170, opened 2025-09-15
Proper handling/mapping for `ThrottlingException` during embedding calls

krinart:
## 📝 Summary

Previously `ThrottlingException` would be mapped to `Internal` error in flight API, which would result in it being displayed as an internal error in repl:

Following changes were made:
* A new `FailedToCreateEmbeddingThrottling` error was introduced
* Proper throttling handling for bedrock/openai
* Mapping `FailedToCreateEmbeddingThrottling` to `unavailable` tonic error

New UX:
Screenshot 2025-09-15 at 4 12 03 PM

## 🔗 Related

https://github.com/spiceai/spiceai/issues/6546

## 🚨 Breaking Changes

N/A

## 📚 Docs

N/A

## 👀 Notes for Reviewers

N/A


crates/llms/src/embeddings/mod.rs
```diff
@@ -46,6 +46,11 @@ pub enum Error {
         source: Box<dyn std::error::Error + Send + Sync>,
     },
 
+    #[snafu(display("Failed to create embedding due to throttling. {source}"))]
+    FailedToCreateEmbeddingThrottling {
```
peasee:
```suggestion
    #[snafu(display("Embedding rate limit exceeded. {source}. Try again later."))]
    RateLimited {
```

crates/llms/src/openai/embed.rs
```diff
@@ -247,6 +252,19 @@ fn is_retriable_error(err: &OpenAIError) -> bool {
     }
 }
 
+fn is_throttling_error(err: &OpenAIError) -> bool {
+    match err {
+        OpenAIError::ApiError(api_err) => {
+            // Supported error codes: https://platform.openai.com/docs/guides/error-codes/api-errors
+            matches!(api_err.code.as_deref(), Some("429" | "503"))
```
lukekim, line 259: 503 isn't an throttling error, that's a OpenAI service down/issue, right?

peasee: OpenAI can return a 503 for "Slow down" - `503 - Slow Down: Cause: A sudden increase in your request rate is impacting service reliability.`, but we should probably match the error message because 503 could also be service unavailable.

krinart: Sorry, not sure what should be done here. The current message is `"Embedding rate limit exceeded. {source}. Try again later."`. Are you saying it should be updated to include `service unavailable`?

peasee: OpenAI can return an error code 503 that either means slow down, or that OpenAI as a service is unavailable. We should make sure we're not silencing the 503 when OpenAI is down.

krinart: Looking at API errors here https://platform.openai.com/docs/guides/error-codes/api-errors, there are 2 cases for 503:

Screenshot 2025-09-15 at 6 45 32 PM

Where do you see 503 indicating service unavailable?

crates/llms/src/embeddings/mod.rs
```diff
@@ -46,6 +46,11 @@ pub enum Error {
         source: Box<dyn std::error::Error + Send + Sync>,
     },
 
+    #[snafu(display("Embedding rate limit exceeded. {source}. Try again later."))]
```
phillipleblanc: `Try again later` is a weird error message to display - the user isn't trying again, Spice will. Perhaps something like "Spice will retry this request" (assuming we do actually retry it)

crates/llms/src/openai/embed.rs
peasee: The first item in that screenshot. `Cause: Our servers are experiencing high traffic`. a-la [HTTP 503 Service Unavailable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503)

crates/llms/src/embeddings/mod.rs
peasee: I thought this error is propagated up after the retry strategy exceeds its retry count?

lukekim: Regardless, we need to make the error better because the user can't try again themselves. I think we can just remove the "Try again later." and leave it at `Embedding rate limit exceeded. {source}.`

crates/llms/src/openai/embed.rs
krinart: Just to make sure, my understanding is that the real 503 will be returned as `OpenAIError::Reqwest`, not `OpenAIError::ApiError`. 

We handle it here: https://github.com/spiceai/spiceai/blob/69bc6e53cf5233cccb78f598f0cc45fbd8ca367a/crates/llms/src/openai/embed.rs#L241

So I do think it is safe to return these two cases of `OpenAIError::ApiError` as a `RateLimited` error. 

Please let me know if I'm missing something.

crates/llms/src/embeddings/mod.rs
krinart: Thank you, done!

crates/llms/src/openai/embed.rs
peasee: Is there some way we can prove that assumption?

phillipleblanc (dismissed):

crates/llms/src/openai/embed.rs
krinart: Good point, I don't think we can. 

So what's the recommendation here? Do you think we should never treat 503 as a throttling error?

peasee: My concern is the error codes page states at the top:

> This guide includes an overview on error codes you might see from both the [API](https://platform.openai.com/docs/introduction) and our [official Python library](https://platform.openai.com/docs/libraries#python-library).

This suggests to me a non-throttling 503 could originate from **either** `OpenAIError::Reqwest` if a load balancer is unavailable or `OpenAIError::ApiError` if it originates directly from their API.

Their Python API also treats [`RateLimitedError` as a 429](https://github.com/openai/openai-python/blob/main/src/openai/_exceptions.py#L128). I think we should not treat 503 as a throttling error because of this.

krinart: Done, thanks!

sgrebnov (approved):

Issue #7107, opened 2025-09-10
`spice trace` fails to parse timestamp


krinart:
## Describe the bug

```bash
> spice trace ai_chat
Spice.ai OSS CLI v1.7.0-unstable-build.9a1effe32
2025/09/10 15:49:57 ERROR SQL query to 'task_history' failed error="error parsing response from spiced: parsing time \"2025-09-10T22:49:53.077161Z\": extra text: \"Z\""
Error: failed to retrieve events from runtime.
```

## To Reproduce

1. Build spice from trunk
2. Execute sql or chat command
3. Run `spice trace ai_chat/sql_query`

## Expected behavior

`spice trace` shows most recent trace

### spice, spiced, OS info

- spice version: v1.7.0-unstable
- spiced version: v1.7.0-unstable
- OS info: Darwin

**Have you tried this on the latest `trunk` branch?**

- [x] Yes
- [ ] No

phillipleblanc:
This is very likely related to https://github.com/spiceai/spiceai/pull/7056

PR #7173, opened 2025-09-16
Fix timestamp parsing for spice trace

krinart:
## 📝 Summary

Add UTC timezone to timestamp parsing for Spice CLI

## 🔗 Related

https://github.com/spiceai/spiceai/issues/7107

### DIFF
--- a/bin/spice/pkg/taskhistory/taskhistory.go
+++ b/bin/spice/pkg/taskhistory/taskhistory.go
@@ -90,7 +90,7 @@ func (tMs *TimeWithMilliSeconds) UnmarshalJSON(b []byte) error {
 	s := string(b)
 	s = s[1 : len(s)-1]
 
-	layout := "2006-01-02T15:04:05.999999"
+	layout := "2006-01-02T15:04:05.999999Z"
 
 	t, err := time.Parse(layout, s)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


bin/spice/pkg/taskhistory/taskhistory.go
```diff
@@ -90,15 +90,23 @@ func (tMs *TimeWithMilliSeconds) UnmarshalJSON(b []byte) error {
...
+
+    for _, layout := range layouts {
+        t, err = time.Parse(layout, s)
+        if err == nil {
+            *tMs = TimeWithMilliSeconds(t)
+            return nil
+        }
+    }
+
+	return fmt.Errorf("cannot parse time: %s", s)
```
lukekim: Seems like we should log the error?

lukekim:
```suggestion
	return fmt.Errorf("cannot parse time %s: %w", s, err)
```

krinart: Good catch. Although I changed to `fmt.Errorf("cannot parse time: %w", err)`. 

Otherwise we would get:
```bash
error parsing response from spiced: cannot parse time 2025-09-16T18:40:50.843578Z: parsing time \"2025-09-16T18:40:50.843578Z\": extra text: \"Z\"
```
Which seems slightly excessive.

Now it's:
```bash
error parsing response from spiced: cannot parse time: parsing time \"2025-09-16T18:41:29.471243Z\": extra text: \"Z\"
```

(The error is an example, the current implementation handles this value properly)

lukekim (approved):

PR #7178, opened 2025-09-16
Kafka: support specifying a target consumer group ID

sgrebnov:
## 📝 Summary

PR adds support for specifying a target consumer group id for Kafka/Debezium connectors via `kafka_consumer_group_id` param. This enabled multiple Spice instances to read data from the same topic/group in cooperative mode or using specific group id

```yaml
- from: kafka:orders_events
    name: orders
    params:
      ...
      kafka_consumer_group_id: spice-orders-group-1
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
      refresh_mode: append
```

## 🔗 Related

Implements
- https://github.com/spiceai/spiceai/issues/6898 

## 👀 Notes for Reviewers

Despite cooperate mode via shared `kafka_consumer_group_id` might NOT be useful for Debezium as parallel processing of CDC might impact correctness/order, decided to still enable `kafka_consumer_group_id` support as it could be used for naming as well

### DIFF
--- a/crates/data_components/src/kafka.rs
+++ b/crates/data_components/src/kafka.rs
@@ -111,6 +111,7 @@ pub struct KafkaConfig {
     pub ssl_ca_location: Option<String>,
     pub enable_ssl_certificate_verification: bool,
     pub ssl_endpoint_identification_algorithm: SslIdentification,
+    pub consumer_group_id: Option<String>,
     pub metrics_store: Option<Arc<KafkaMetrics>>,
 }
 
@@ -134,6 +135,7 @@ impl std::fmt::Debug for KafkaConfig {
                 "ssl_endpoint_identification_algorithm",
                 &self.ssl_endpoint_identification_algorithm,
             )
+            .field("consumer_group_id", &self.consumer_group_id)
             .field(
                 "metrics_store",
                 &self.metrics_store.as_ref().map(|_| "Some(KafkaMetrics)"),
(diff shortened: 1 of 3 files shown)


crates/data_components/src/kafka.rs
```diff
@@ -190,8 +192,8 @@ impl rdkafka::ClientContext for KafkaConsumerContext {
 
         for topic in statistics.topics.values() {
             for partition in topic.partitions.values() {
-                // Skip internal partitions (partition id -1)
-                if partition.partition >= 0 {
+                // Skip internal partitions (partition id -1), and only consider partitions with known lag (-1 means unknown)
+                if partition.partition >= 0 && partition.consumer_lag >= 0 {
```
sgrebnov, line 196: This is unrelated improvement. I found out that during rebalancing `consumer_lag` can actually be -1 indicating that this information is not yet available

crates/runtime/src/dataconnector/kafka.rs
```diff
@@ -357,6 +364,20 @@ async fn init_kafka_consumer(
         }
     );
 
+    if let Some(ref group_id) = kafka_config.consumer_group_id {
+        ensure!(
+            group_id == &metadata.consumer_group_id,
+            super::InvalidConfigurationNoSourceSnafu {
```
krinart, line 370: This error type doesn't seem to be fitting this use case

krinart (approved):

crates/runtime/src/dataconnector/kafka.rs
sgrebnov: Yeah, this can confuse - we have `InvalidConfiguration`, `InvalidConfigurationSourceOnly`, `InvalidConfigurationNoSource`, I propose to improve this separately.

```
 InvalidConfiguration {
        dataconnector: String,
        connector_component: ConnectorComponent,
        message: String,
    },

    #[snafu(display(
        "Cannot setup the {connector_component} ({dataconnector}) with an invalid configuration. {source}"
    ))]
    InvalidConfigurationSourceOnly {
        dataconnector: String,
        connector_component: ConnectorComponent,
    },

    #[snafu(display(
        "Cannot setup the {connector_component} ({dataconnector}) with an invalid configuration. {message}"
    ))]
    InvalidConfigurationNoSource {
        dataconnector: String,
        connector_component: ConnectorComponent,
        message: String,
    },
```

PR #7183, opened 2025-09-16
SpiceExtensionPlanner/BytesProcessedNode to account for latent projection pushdown

mach-kernel:
## 📝 Summary

When using the RRF UDTF with some federated sources, it's possible to get a schema mismatch error on query:

```
sql> SELECT id, description, fused_score
FROM rrf(
        text_search(sources_data, 'drones'),
        vector_search(sources_data, 'drones')
     )
ORDER BY fused_score DESC
LIMIT 5;
Internal Error: Unexpected internal error. Use '.error' for details.
sql> .error
Client specified an invalid argument. Extension planner for BytesProcessedNode created an ExecutionPlan with mismatched schema. LogicalPlan schema: DFSchema { inner: Schema { fields...
```

This adds a check in `SpiceExtensionPlanner` to detect these situations and optionally pass through the logical plan schema (with the correct projection) to `BytesProcessedExec`.


crates/runtime/src/datafusion/extension/mod.rs
```diff
@@ -79,15 +79,29 @@ impl ExtensionPlanner for SpiceExtensionPlanner {
     ) -> Result<Option<Arc<dyn ExecutionPlan>>> {
         // bytes_processed Extension
         let bytes_processed_node = node.as_any().downcast_ref::<BytesProcessedNode>();
-        if bytes_processed_node.is_some() {
+
+        if let Some(bytes_processed_node) = bytes_processed_node {
```
mach-kernel, line 83: I tried to make a test that pushes the query via `ctx.sql` but I cannot reproduce the problem by just mocking the scenario with `MemTable`. I'll keep trying for a bit.

phillipleblanc (approved):

Issue #7035, opened 2025-09-04
Time skew causes cache results metrics to subtract with overflow


peasee:
## Describe the bug

_A clear and concise description of what the bug is._

```console
thread 'tokio-runtime-worker' panicked at crates/cache/src/lru_cache.rs:175:12:
attempt to subtract with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

E.g:

```rust
let now_seconds = current_time_secs();
        if now_seconds - self.metrics_last_reported_time.load(Ordering::Relaxed) >= 5
```

If `now_seconds` has skewed behind `metrics_last_reported_time`, a subtract with overflow will occur.

## To Reproduce

_Steps to reproduce the behavior:_

2. Click on '...'
3. Scroll down to '...'

## Expected behavior

_A clear and concise description of what you expected to happen._

### Spicepod

_Add the relevant `spicepod.yml` section._

### Output of `describe table`

_Add any relevant `describe table` output here._

### Output of `explain query`

_Add any relevant `explain query` output here._

E.g. `explain select 1`

```bash
+---------------+--------------------------------------+
| plan_type     | plan                                 |
+---------------+--------------------------------------+
| logical_plan  | Projection: Int64(1)                 |
|               |   EmptyRelation                      |
| physical_plan | ProjectionExec: expr=[1 as Int64(1)] |
|               |   PlaceholderRowExec                 |
|               |                                      |
+---------------+--------------------------------------+
```

### spice, spiced, OS info

- spice version: `spice version`
- spiced version: `spiced --version`
- OS info: `uname -a`

**Have you tried this on the latest `trunk` branch?**

- [ ] Yes
- [ ] No

**If possible, run `spiced` with DEBUG log level**
By setting the environment variable `SPICED_LOG="task_history=DEBUG,spiced=DEBUG,runtime=DEBUG,secrets=DEBUG,data_components=DEBUG,cache=DEBUG,extensions=DEBUG,spice_cloud=DEBUG"`

PR #7190, opened 2025-09-16
Prevent time skew in `LruCache` by using `Instant`

krinart:
## 📝 Summary

Previously we were using `SystemTime` which may result in time skew. To avoid this we are now using `Instant` which is monotonically increasing.

## 🔗 Related

https://github.com/spiceai/spiceai/issues/7035


crates/cache/src/lru_cache.rs
```diff
@@ -160,24 +164,24 @@ impl<
     async fn put_raw_key(&self, key: &u64, value: V) {
         self.cache.insert(*key, value).await;
 
-        let now_seconds = current_time_secs();
-        if now_seconds - self.metrics_last_reported_time.load(Ordering::Relaxed) >= 5 {
-            self.metrics_last_reported_time
-                .store(now_seconds, Ordering::Relaxed);
+        let now = Instant::now();
+        let mut last_report = self.metrics_last_reported_time.lock().await;
```
peasee, line 198: we should benchmark this. the lock might introduce a larger performance impact than an atomic number. e.g. how would this compare to just correctly doing:

```rust
if let Some(delta_seconds) = now_seconds.checked_sub(self.metrics_last_reported_time.load(Ordering::Relaxed)) {}
```

krinart: Good point! Also I feel like there's an issue with the current implemention: shouldn't we use `compare_exchange` ito guarantee exclusive update? 

Should I run benchmarks against current implementation, or with `compare_exchange`?

peasee: Sure, let's compare all 3. If `compare_exchange` and the existing implementation are within margins of error, we could just use `compare_exchange`. If the `Mutex` is also within margin of error, we could probably just stick with it as `Instant` is better than `SystemTime`.

Let's also use the default 1 second TTL, so that we're exercising the cache-put path for this change.

krinart: Benchmarking results:

Screenshot 2025-09-17 at 12 39 53 PM

Screenshot 2025-09-17 at 12 34 29 PM


Source code: https://gist.github.com/krinart/d0773c0e6266eccc72982156ae131c57

Command:
```bash
oha -c 8 -n 1000000 -m POST http://127.0.0.1:7777/atomic_store
oha -c 8 -n 1000000 -m POST http://127.0.0.1:7777/atomic_cas
oha -c 8 -n 1000000 -m POST http://127.0.0.1:7777/async_mutex
```


Oberall it seems like there is no significant difference in performance between the two approaches: in case of 1 and 2 threads instant+mutext was slightly worse, in case of 4 and 8 threads it was the other way around.

However, from a correctness perspective, I think instant + mutex is preferable: `Instant` is designed to provide monotonically increasing time, which avoids potential issues with system clock drift or non-monotonic time sources.

peasee: I agree let's stick with instant and mutex, thanks for doing these benchmarks

peasee (dismissed):

krinart:
Hey @peasee - I made minor chage, now `metrics_last_reported_time` is `Mutex<Option<Instant>>` and is initialized as `Mutex::new(None)`.

The alternative was to return `Option` from `LruCache::new` which I kinda hate (because we would need to do `Instant.sub(Duration)` which may or may not succeed).

Please take a look when you have a chance. Thank you!

phillipleblanc (dismissed):

krinart:
More benchmarking with the fourth option: channel+worker. I also extended code to have:
* A proper `moka` cache with 20 keys that are updated randomly
* `opentelemetry` metrics
* Add fourth option with channel+worker

Code: https://gist.github.com/krinart/c98ee7e824adc350fcf3efcf2affc9b4

# 
**Notes:** Threads on the chart represent number of threads in the runtime

### Running 8 oha threads:

Screenshot 2025-09-18 at 2 27 54 PM

Screenshot 2025-09-18 at 2 27 25 PM

### Running 128 oha threads:

Screenshot 2025-09-18 at 2 27 54 PM

Screenshot 2025-09-18 at 2 27 25 PM

### Previous charts combined together:

Screenshot 2025-09-18 at 2 31 43 PM

Screenshot 2025-09-18 at 2 31 56 PM

# Learnings:
1. When using 8 oha threads, the difference between 4 approaches is insignificant.
2. With 128 oha threads, mutex-based approach consistently shows 50% throughput of other 3 approaches - which disqualifies this approach

The atomic+store approach should be disqualified since it does not guarantee exclusive updates which may result in duplicate updates.

Which leaves 2 options:

|                 | Pros | Cons |
|----------|----------|----------|
| atomic+cas    | Simplicity   | * Time drift mitigated by using  `saturating_sub`  <br> * Not using the `Instant` which is designed specifically for cases like this |
| channel+worked    | Not having time drift issues   | Extra code to deal with channel/worker   |

krinart:
@peasee @phillipleblanc Can I ask you to take a look at the benchmarks?

Recommendation from @lukekim was to use atomic+cas for now. And maybe in the future use channel+worker approach, potentially having a single worker for all telemetry metrics.

Issue #6180, opened 2025-06-10
Enhancement: Results caching for embeddings


peasee:
## Goal-State/What/Result

Embeddings requests against the same embedding model with the same content are cached.

## Why/Purpose

To improve the performance of embeddings, reducing costs by limiting embedding requests for the same content.

## By When

**Issue/Spec written and reviewed:** 16th Sep, 2025
**Done-Done:** 18th Sep, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

A `embeddings` section under `caching` is created:

```yaml
runtime:
  caching:
    embeddings:
      enabled: true
      max_size: 128mb
      item_ttl: 5s
      eviction_policy: lru
      hashing_algorithm: siphash
```

### Security Review

* Users should consider whether their embedded content is sensitive, and whether this data should be cached in memory. For these use-cases, users should disable the results cache.
* Malicious users could perform cache denial-of-service attacks, where they request excessive search terms or content which causes cache evictions of legitimate-use cache entries.
  * Users should ensure they use API keys to limit access to authorized users, preventing such requests from malicious users.

## How/Implementation Plan

* Embeddings cache is added to the `Caching` struct
* Embedding models take a reference to the `Caching` struct and return results from the cache if they exist
* Embedding models store results in the cache on completion. Records are stored per-item, so batched requests will create multiple cache entries (one for each item in the batch).

## QA Plan

* [x] Integration tests are added to test the efficiency of embeddings caching

## Release Notes

* **Caching Support for Embedding Requests:** The Runtime now supports caching embedding requests, improving the performance and reducing cost of embedding content and search requests.

Example `spicepod.yml`:

```yaml
runtime:
  caching:
    embeddings:
      enabled: true
      max_size: 128mb
      item_ttl: 5s
```

[...]

PR #7194, opened 2025-09-17
feat: Add initial embeddings cache

### LINKED ISSUE #6109: Caching metrics for search & embeddings
The SQL results cache has the following existing metrics:

* `results_cache_max_size_bytes`
* `results_cache_requests`
* `results_cache_hits`
* `results_cache_items_count`
* `results_cache_size_bytes`

Caching for search and embeddings should following a similar pattern to emit the same metrics:

**Search:**

* `search_results_cache_max_size_bytes`
* `search_results_cache_requests`
* `search_results_cache_hits`
* `search_results_cache_items_count`
* `search_results_cache_size_bytes`

**Embeddings:**

* `embeddings_cache_max_size_bytes`
* `embeddings_cache_requests`
* `embeddings_cache_hits`
* `embeddings_cache_items_count`
* `embeddings_cache_size_bytes`

In a future version (v2.0), the results cache metrics should be updated to include the relevant prefix for `sql_` to become `sql_results_cache_max_size_bytes`. This aligns with the caching parameter naming for `sql_results` vs `search_results`.

peasee:
## 📝 Summary

* Adds an embeddings cache provider, and cache wrapper for `CreateEmbeddingResponse` or raw vectors
* Adds a `CacheConfig` for `runtime.caching.embeddings`
* Adds metrics for `CachedEmbeddingResult`
* Integrates caching into the `Embed` trait, and updates `OpenaiEmbed` to use it
  * Caching is implemented within each specific implementation, so caching can cache items that are batched/chunked/split. As from the spec:
  > Records are stored per-item, so batched requests will create multiple cache entries (one for each item in the batch).
* Implements `Display` for `CacheProvider`, so other cache providers can print their max size and TTL similar to SQL results cache

## 🔗 Related

* Part of #6180
* Closes #6109


crates/cache/src/lib.rs
```diff
@@ -77,24 +79,34 @@ pub trait Sizeable {
...
+            .sum()
+    }
+}
+
 pub trait HashProvider {
     fn hasher(&self) -> Box<dyn Hasher>;
 }
 
 /// Trait for types that can be converted to a set of table references.
 pub trait AsTableRefs {
```
phillipleblanc: Should we just not implement this for the new embedding cache? Seems weird to make this change.

crates/cache/src/lru_cache.rs
```diff
@@ -44,6 +45,22 @@ pub struct LruCache<
     hasher: T,
     max_size: u64,
     metrics_last_reported_time: AtomicU64,
+    ttl: Duration,
+}
+
+impl<
+    V: Sizeable + AsTableRefs + CacheMetrics + Clone + Send + Sync + 'static,
```
phillipleblanc: This doesn't strictly require `AsTableRefs` right?

phillipleblanc (approved):

Issue #6746, opened 2025-08-13
Enhancement: Search Query Partitioning


lukekim:
## Goal-State/What/Result

Queries through `vector_search` and `text_search` can be partitioned/sharded across multiple nodes.

## By When

**Issue/Spec written and reviewed:** 
**Done-Done:** 

## Done-Done

- [ ] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Telemetry / Metrics / Task History
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials

## The Algorithm

- [ ] Every requirement questioned?
- [ ] Delete (Scope) any part you can.
- [ ] Simplify.
- [ ] Break down into smaller iterations/milestones.
- [ ] Opportunities for automation.

## Specification

- Support `partition_by:` for S3 Vectors.
- Partitioned writes/reads according to S3 Vectors limits (E.g. index limits, rate-limiting, etc.).

lukekim:
Thanks for the report! This appears to be fixed by spiceai/spiceai#7313 (merged 2025-09-17). Closing for now — please reopen if you can still reproduce on a release that includes this commit.

PR #7198, opened 2025-09-17
Create and parse partitioned S3 vector index names

### LINKED ISSUE #7185: S3 Vector partition naming

kczimm:
## 📝 Summary

Defines logic for generating and parsing partitioned AWS S3 vector index names according to a particular scheme specified in #6746.

## 🔗 Related

Closes:
- #7185


peasee (dismissed):

crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -0,0 +1,231 @@
...
+        assert_eq!(result.column_name, "test-col");
+        assert_eq!(
+            result.partition_value_hash.len(),
+            PARTITION_VALUE_MAX_LENGTH
+        );
+        assert_eq!(result.partition_by_hash.len(), PARTITION_BY_MAX_LENGTH);
+    }
+
+    #[test]
+    fn from_index_name_valid() {
```
phillipleblanc: Let's add a test where the index name and column name has a `-`, i.e. `test-index-test-col-12345-abcde`

crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -0,0 +1,231 @@
...
+        partition_by: &[Expr],
+    ) -> Result<Self, Error> {
+        let index_name = truncate(&sanitize(index_name), INDEX_NAME_MAX_LENGTH);
+        let column_name = truncate(&sanitize(column_name), COLUMN_NAME_MAX_LENGTH);
+        let partition_value_hash = truncate(
+            &hash_to_hex(&partition_value.to_string()),
+            PARTITION_VALUE_MAX_LENGTH,
+        );
+        let partition_by_hash = truncate(
+            &hash_to_hex(&format!("{partition_by:?}")),
```
phillipleblanc: Is the debug representataion of an Expression guaranteed to be stable across DF versions? If not, we may want to built our own stable representation so that upgrading Spice won't cause us to lose the association to existing indexes.

crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -0,0 +1,231 @@
...
+            self.index_name.clone(),
+            self.column_name.clone(),
+            self.partition_value_hash.clone(),
+            self.partition_by_hash.clone(),
+        ]
+        .join(PARTS_SEPARATOR)
+    }
+
+    pub fn from_index_name(index_name: &str) -> Result<Self, Error> {
+        let parts: Vec<&str> = index_name.split('-').collect();
```
phillipleblanc: I think this will need to handle the case where the index_name/column_name have dashes

phillipleblanc: Either that, or sanitize the index_name/column_name to use something other than a dash if they have one.

crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -0,0 +1,231 @@
...
+        let index_name = "test_index";
+        let column_name = "test_col";
+        let partition_value = ScalarValue::from("value");
+        let partition_by = vec![col("col1")];
+
+        let result =
+            PartitionedIndexName::new(index_name, column_name, &partition_value, &partition_by)
+                .expect("new");
+
+        assert_eq!(result.index_name, "test-index");
```
phillipleblanc, line 381: One of the side effects of the sanitization is that the index name is different from what is passed in - is that going to be a problem in the calling code?

phillipleblanc: Or maybe a simpler solution is instead of sanitizing, just require that the index_name/column_name not have  `-` or `_` in them. That might work if we control the `index_name`/`column_name` - but if that comes from user input where then we would need to maintain a mapping in the calling code.

crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -0,0 +1,231 @@
...
+const INDEX_NAME_MAX_LENGTH: usize = 25;
+const COLUMN_NAME_MAX_LENGTH: usize = 25;
+const PARTITION_VALUE_MAX_LENGTH: usize = 5;
+const PARTITION_BY_MAX_LENGTH: usize = 5;
+
+static PARTS_SEPARATOR: &str = "-";
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Expected 4 parts in index name between hypens but got {num_parts}"))]
```
lukekim:
```suggestion
    #[snafu(display("Expected exactly 4 parts in index name separated by hyphens, but found {num_parts}"))]
```

crates/data_components/src/s3_vectors/partition.rs
kczimm: you really woudn't be able to tell which part (index or column) contains the hyphen in the case where there is an extra hyphen. I think a different separator would be better. I think a `.` would work: https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-vectors-indexes.html#s3-vectors-indexes-naming and it is valid for filenames in filesystems.

crates/data_components/src/s3_vectors/partition.rs
kczimm: I don't think we can practically reject column names with '_' at least. It is so pervasive.

crates/data_components/src/s3_vectors/partition.rs
kczimm: Well the Debug implementation is [derived](https://docs.rs/datafusion-expr/50.0.0/src/datafusion_expr/expr.rs.html#279). If the Expression variant changes then it would not be backwards compatible. For example, [this change](https://github.com/apache/datafusion/commit/e4166b368a611387f4b7b1eeef98548cf593ddf0) would have broke partitions three months ago. I don't like that we have to do it but perhaps our own stable representation is the way to go here.

kczimm: I'm going to pause and think harder on this because it's really quite the undertaking. Expr variants have many other types inside of them and we would be committing to making a stable representation for every type that appears in every Expr variant. I don't think we want to own this code and deal with the maintenance of it.

kczimm: I think it will be better to unparse each expression into SQL which should be stable and then just join the SQL together before hashing. I can add snapshots for common expressions so we can catch issues in DF upgrades.

crates/data_components/src/s3_vectors/partition.rs
phillipleblanc: Yeah, so I guess that means we need to keep a mapping in the calling code?

crates/data_components/src/s3_vectors/partition.rs
phillipleblanc: We could just build a mapping of the subset of expressions we want to support in our own code that we could use here. i.e. SpiceExpr that only contains the few common expressions we expect to support with a conversion to/from the DF one.

crates/data_components/src/s3_vectors/partition.rs
kczimm: I don't think so. The comparison in `belongs_with` happens after sanitizing.

Although, we should just error if the index name contain a hyphen because they just can't name indexes with hyphens due to AWS restrictions.

crates/data_components/src/s3_vectors/partition.rs
kczimm: If we agree SQL is stable then this is unnecessary. It would just be code we have to write and maintain. If SQL wasn't stable I would agree we have to do this as I believed earlier. Is there a good reason not to just use the unparsed SQL?

crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -0,0 +1,342 @@
...
+fn to_stable_string(exprs: &[Expr]) -> Result<String, DataFusionError> {
+    Ok(exprs
+        .iter()
+        .map(|expr| expr_to_sql(expr).map(|e| e.to_string()))
+        .collect::<Result<Vec<String>, DataFusionError>>()?
+        .join(PARTS_SEPARATOR))
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
```
phillipleblanc:
```suggestion
```

We shouldn't use unwrap in tests, use `expect` instead.

crates/data_components/src/s3_vectors/partition.rs
phillipleblanc: In an ideal world it should be, but the `expr_to_sql` code isn't guaranteed to return the same SQL for a given expression. i.e. this PR very likely would have broken this logic: https://github.com/apache/datafusion/pull/11186

kczimm: Hmm I understand that the SQL is not stable now. That's unfortunate. I'll work on the subset of expressions.

crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -0,0 +1,459 @@
...
+static PARTS_SEPARATOR: &str = ".";
+
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display(
+        "Expected exactly 4 parts in index name separated by hyphens, but found {num_parts}"
+    ))]
+    IncorrectNumPartsInName { num_parts: usize },
+    #[snafu(display("Failed to unparse a partition_by expression: {source}"))]
+    UnparsingExpression { source: DataFusionError },
```
phillipleblanc: nit: Update this error message, `unparse` doesn't mean anything to users.

phillipleblanc (dismissed):

crates/data_components/src/s3_vectors/partition.rs
lukekim: Not a nit. We should always maintain the highest bar for any output to a user.

crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -0,0 +1,459 @@
...
+fn to_stable_string(exprs: &[Expr]) -> Result<String, DataFusionError> {
+    Ok(exprs
+        .iter()
+        .map(stable_expr_string)
+        .collect::<Result<Vec<String>, DataFusionError>>()?
+        .join(PARTS_SEPARATOR))
+}
+
+#[allow(clippy::too_many_lines)]
+fn stable_expr_string(expr: &Expr) -> Result<String, DataFusionError> {
```
phillipleblanc: Should just return our own Error instead of a DataFusion error here, then you get rid of the current awkward error message.

crates/data_components/src/s3_vectors/partition.rs
mach-kernel: I don't know what to suggest as an alternative here. I like that DF has the `exec_err!`, `plan_err!` and `internal_err!` macros: when thinking of errors I try to bucket the failure into one of these three types.

So to that end this would most closely be an `exec_err!`. I think that it being unparsing specific is OK. Broken execution failures need to be `grepable` -- this isn't an error the user will interpret and then do something to the spicepod to fix. They'll open an issue. We will need to know what's wrong. If we write something like "failed to stabilize partition expression representation" I think it'll be less clear.

Can the advice here be:

- Add the "This is a bug in Spice AI pls file issue" suffix we see in other messages as a CTA to file an issue.

What do we think?

kczimm: I just decided to make it perfect.

mach-kernel (approved):

crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -0,0 +1,512 @@
...
+        );
+
+        if self.index_name != index_name {
+            BelongsWith::DifferentDataset
+        } else if self.column_name_hash != column_name_hash {
+            BelongsWith::DifferentColumn
+        } else if self.partition_by_hash != partition_by_hash {
+            BelongsWith::DifferentPartitionByExpressions
+        } else {
+            BelongsWith::ThisDataset
```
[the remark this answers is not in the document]
kczimm: the original name couldn't have been created with that partition by expression then so they won't match any way. it will be ok

mach-kernel (approved):

PR #7199, opened 2025-09-17
Create PutVectorsSink 

kczimm:
## 📝 Summary

Creates a new `DataSink` for writing vectors into AWS S3 vectors. Current write implementation does not use DataFusion's abstractions like `TableProviders`, `DataSinks`, `DataSinkExecs`, or `ExecutionPlans`. In order to reuse the `PartitionedTableProvider`, a `TableProvider` is needed for reading/writing to/from S3 vectors. A new one will need to be built where the write side uses PutVectors API and the read side uses the QueryVectors API. This addresses the former.

## 🔗 Related

Needed for #6746.


crates/data_components/src/s3_vectors/put_vectors_sink.rs
```diff
@@ -0,0 +1,428 @@
+/*
+Copyright 2024-2025 The Spice.ai OSS Authors
```
phillipleblanc:
```suggestion
Copyright 2025 The Spice.ai OSS Authors
```

nit: for new files the copyright header should just be the current year

crates/data_components/src/s3_vectors/put_vectors_sink.rs
```diff
@@ -0,0 +1,428 @@
...
+            .build()
+            .context(BuildInputSnafu)?;
+
+        put_input_vectors.push(put_input_vector);
+    }
+
+    Ok(put_input_vectors)
+}
+
+#[allow(clippy::unwrap_used)]
```
phillipleblanc: We shouldn't use unwrap in production code - if you are confident this can never happen, then we should use the `unreachable!` macro with the reason why it can't be reached.

phillipleblanc: But in this instance, I'd prefer to have this just return a Result instead

crates/data_components/src/s3_vectors/put_vectors_sink.rs
```diff
@@ -0,0 +1,428 @@
...
+            Document::Number(Number::Float(arr.value(row)))
+        }
+        DataType::Boolean => {
+            let arr = col
+                .as_any()
+                .downcast_ref::<arrow::array::BooleanArray>()
+                .unwrap(); // UNWRAP: we checked the datatype
+            Document::Bool(arr.value(row))
+        }
+        _ => unimplemented!(),
```
phillipleblanc, line 267: This should return a Result instead of panicking

crates/data_components/src/s3_vectors/put_vectors_sink.rs
```diff
@@ -0,0 +1,428 @@
...
+}
+
+impl From<Error> for DataFusionError {
+    fn from(value: Error) -> Self {
+        DataFusionError::Execution(value.to_string())
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::unwrap_used)]
```
phillipleblanc:
```suggestion
#[allow(clippy::expect_used)]
```

We allow expect in tests, but not unwrap

crates/data_components/src/s3_vectors/put_vectors_sink.rs
```diff
@@ -0,0 +1,428 @@
...
+            }
+
+            list_builder.append(true);
+        }
+
+        list_builder.finish()
+    }
+
+    #[test]
+    #[allow(clippy::unwrap_used)]
```
phillipleblanc:
```suggestion
```

crates/data_components/src/s3_vectors/put_vectors_sink.rs
```diff
@@ -0,0 +1,428 @@
...
+
+                count += chunk.len();
+            }
+        }
+
+        Ok(count as _)
+    }
+}
+
+#[allow(clippy::result_large_err)]
```
phillipleblanc: If we box the PutVectorsError then we won't need this clippy allow

crates/data_components/src/s3_vectors/put_vectors_sink.rs
```diff
@@ -0,0 +1,442 @@
...
+        .ok_or_else(|| Error::MissingColumn { name: name.clone() })?
+        .as_any()
+        .downcast_ref::<arrow::array::ListArray>()
+        .ok_or_else(|| Error::ColumnTypeMismatch {
+            name,
+            expected: "ListArray".to_string(),
+        })?;
+
+    let schema = record_batch.schema();
+    let fields = schema
```
mach-kernel, line 154: It boils down to the same logic, but you could tag columns in the schema as metadata only and then collect those (if you wanted to decouple column name assertions in this specific stage)

crates/data_components/src/s3_vectors/put_vectors_sink.rs
```diff
@@ -0,0 +1,442 @@
...
+            .as_any()
+            .downcast_ref::<arrow::array::Float32Array>()
+            .ok_or_else(|| Error::ColumnTypeMismatch {
+                name: format!("data[{row}]"),
+                expected: "Float32Array".to_string(),
+            })?
+            .values()
+            .to_vec();
+
+        if vector.iter().any(|&x| x.is_nan() || x.is_infinite()) {
```
mach-kernel, line 179: Warn that you're dropping?

mach-kernel (dismissed):
Two small comments, but otherwise makes sense

crates/data_components/src/s3_vectors/put_vectors_sink.rs
kczimm: I do like this idea. I think if we do this here we should do this everywhere. That should be a separate PR, IMHO.

mach-kernel (approved):

PR #7200, opened 2025-09-17
Fix projection mismatch issues with RRF calling vector search / text search 

mach-kernel:
## 📝 Summary

- rrf::scan was always emitting the whole projection, where the planner would fix by adding the right proj/limit
- Deduped schema from join in vector_search/text_search would have a set of unqualified fields. Fine if there are no duplicates, but can fail when querying some federated sources (this case, FlightExec via the SpiceAI connector).

### DIFF
--- a/crates/runtime/src/embeddings/index/scan_table.rs
+++ b/crates/runtime/src/embeddings/index/scan_table.rs
@@ -32,7 +32,7 @@ use datafusion::{
     scalar::ScalarValue,
     sql::TableReference,
 };
-use datafusion_expr::{SubqueryAlias, col};
+use datafusion_expr::SubqueryAlias;
 
 use crate::{
     embedding_col,
--- a/crates/runtime/src/embeddings/index/scan_table.rs
+++ b/crates/runtime/src/embeddings/index/scan_table.rs
@@ -293,8 +293,10 @@ impl TableProvider for VectorScanTableProvider {
                         && tbl.is_some_and(|t| *t == TableReference::parse_str("vector_index")))
                 })
                 .map(|(tbl, field_ref)| match tbl {
-                    Some(table_ref) => col(format!("{}.{}", table_ref.table(), field_ref.name())),
-                    None => col(field_ref.name()),
+                    Some(table_ref) => {
+                        Expr::Column(Column::new(Some(table_ref.clone()), field_ref.name()))
+                    }
+                    None => Expr::Column(Column::new(None::<TableReference>, field_ref.name())),
                 })
                 .collect();
 
--- a/crates/runtime/src/embeddings/index/scan_table.rs
+++ b/crates/runtime/src/embeddings/index/scan_table.rs
@@ -32,7 +32,7 @@ use datafusion::{
     scalar::ScalarValue,
     sql::TableReference,
 };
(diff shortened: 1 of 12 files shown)


crates/runtime/src/embeddings/index/scan_table.rs
phillipleblanc, line 1: Let's add some tests for RRF to the existing search tests: https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/tests/models/search.rs#L618

crates/runtime/src/embeddings/index/scan_table.rs
```diff
@@ -285,22 +285,23 @@ impl TableProvider for VectorScanTableProvider {
...
-            )?;
+            let deduped_join_proj_exprs: Vec<_> = join
+                .schema()
+                .iter()
+                .filter(|(tbl, f)| {
+                    !(primary_key_column_names.contains(f.name())
+                        && tbl.is_some_and(|t| *t == TableReference::parse_str("vector_index")))
+                })
+                .map(|(tbl, field_ref)| match tbl {
+                    Some(table_ref) => {
```
mach-kernel, line 296: The difference here is that column binding happens during logical planning instead of immediately with the provided schema.

I spun out #7206 re this change.

crates/runtime/tests/models/search.rs
```diff
@@ -341,7 +341,7 @@ pub(crate) async fn run_search_w_explain(
                                 format!("{}_error_response", ts.name),
                                 err.to_string()
                             );
-                            return Ok(());
+                            continue;
```
mach-kernel, line 344: Don't worry, I do stuff like this all the time. I will update the snapshots for the cases that did not run.

crates/runtime/tests/models/search.rs
```diff
@@ -983,7 +1052,7 @@ async fn test_text_search_multiple_columns() -> Result<(), anyhow::Error> {
             SearchTestCase::new(
                 "multi_text_column_sql_text_search_basic_question",
                 SearchTestType::Sql("SELECT id, question, trunc(score, 3) FROM text_search(qs, 'angles', question) order by score desc LIMIT 4"),
-            ),
+            ).skip(),
```
mach-kernel, line 1055: These are broken because: 
- `search_column` chooses the first search field: https://github.com/spiceai/spiceai/blob/trunk/crates/search/src/generation/text_search/index.rs#L374-L380
- We assume the search provider to give us all metadata columns: https://github.com/spiceai/spiceai/blob/trunk/crates/search/src/provider.rs#L123C33-L132

I think this patch is outside the scope of this PR. cc @Jeadie

kczimm: Let's make sure we have an issue to remind us to come back and fix these.

kczimm (approved):

crates/runtime/tests/models/search.rs
mach-kernel: https://github.com/spiceai/spiceai/issues/7207

crates/runtime/src/embeddings/index/scan_table.rs
mach-kernel: Added

PR #7201, opened 2025-09-17
Make S3vector a FixedSizeListArray

Jeadie:
## 📝 Summary
 - Use the dimension return when finding or creating a S3 vector index to construct a `FixedSizeListArray` type (instead of a more generic `ListArray`. 
- Also, S3vector does not allow empty/null vector data, so remove nullability of `FixedSizeListArray` elements. 
- We cannot convert JSON (from S3 vectors) into `FixedSizeListArray` directly, so a manual `arrow::compute::cast` is required.

### DIFF
--- a/crates/data_components/src/s3_vectors/vector_table.rs
+++ b/crates/data_components/src/s3_vectors/vector_table.rs
@@ -339,7 +339,7 @@ impl S3VectorsTable {
                         S3_VECTOR_EMBEDDING_NAME,
                         Field::new("item", DataType::Float32, false),
                         embedding_dimension,
-                        true,
+                        false,
                     )),
                     Arc::new(Field::new(
                         S3_VECTOR_PRIMARY_KEY_NAME,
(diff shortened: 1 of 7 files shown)


crates/data_components/src/s3_vectors/vector_table.rs
```diff
@@ -450,3 +447,106 @@ impl S3VectorsTable {
...
+                }
+                _ => Arc::clone(f),
+            }
+        })
+        .collect();
+
+    (Arc::new(Schema::new(fields)), len)
+}
+
+pub(super) fn replace_column_in_record(
```
phillipleblanc: Potentially a candidate for moving to `crates/arrow_tools`

crates/data_components/src/s3_vectors/vector_table.rs
```diff
@@ -450,3 +447,106 @@ impl S3VectorsTable {
...
+        .collect::<Vec<_>>();
+
+    RecordBatch::try_new(schema.into(), columns)
+}
+
+pub(super) async fn send_vector_data(
+    tx: &Sender<Result<RecordBatch, DataFusionError>>,
+    rb: RecordBatch,
+    vector_size: i32,
+) -> () {
```
phillipleblanc:
```suggestion
) {
```

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #7260, opened 2025-09-22
Add index name and remove duplicate records string to S3 Vectors log

lukekim:
This pull request makes a small update to the logging statement in the `S3VectorsTable` implementation. The change improves the log output by including the `index_name` in the message and clarifying the records count field.

* Improved logging: The info log now displays the `index_name` and uses a named field for the records count, making log messages clearer and more informative. (`crates/data_components/src/s3_vectors/vector_table.rs`)


sgrebnov (dismissed):

crates/data_components/src/s3_vectors/vector_table.rs
```diff
@@ -439,8 +439,9 @@ impl S3VectorsTable {
         }
 
         tracing::info!(
-            "S3 Vectors Index updated; records={} records, duration={duration:?}",
-            vectors.len(),
+            "S3 Vectors Index {:?} updated; records={records}, duration={duration:?}",
```
peasee:
```suggestion
            "S3 Vectors Index {} updated; records={records}, duration={duration:?}",
```

info traces should not include debug output. I guess the same applies for `duration:?` but I think duration has a pretty print for `:?` compared to other outputs.

Jeadie: Yes, and it may be `None` even when there is a valid index.

Jeadie: We should `impl Display for S3VectorIdentifier` and then use that

phillipleblanc (approved):

PR #7273, opened 2025-09-23
Update naming of `tool_use::document_similarity` and `vector_search` spans

Jeadie:
## 📝 Summary
Update naming conventions for:
 - `tool_use::document_similarity` -> `tool_use::search`
 - `vector_search` -> `search` (tracing span name, internal naming/filepaths; not vector search UDTF and similar). 

This includes related changes:
 - New property in `GET v1/datasets`: `search` returning un/supported if the table can be searched (text, vector or both). 

## 🚨 Breaking Changes
 - [ ] No `document_similarity` tool. Replaced with `search` tool.
 - [ ] No `spice trace tool_use::document_similarity`. Replaced with `spice trace tool_use::search`
 - [ ] No `vector_search` span. Replaced with `search` span. 
   - [ ] TODO: add text_search, vector_search subspans. 

## 📚 Docs
 - https://github.com/spiceai/cookbook/pull/278
 - https://github.com/spiceai/docs/pull/1159


crates/runtime/src/tools/builtin/catalog.rs
```diff
@@ -29,10 +29,10 @@ use crate::{
 
 use super::{
     SpiceModelTool,
-    document_similarity::DocumentSimilarityTool,
     get_readiness::GetReadinessTool,
     list_datasets::ListDatasetsTool,
     sample::{SampleTableMethod, tool::SampleDataTool},
+    search::DocumentSimilarityTool,
```
lukekim: It seems we should also rename the tool to `SearchTool`?

crates/runtime/src/tools/builtin/catalog.rs
```diff
@@ -77,9 +77,7 @@ impl BuiltinToolCatalog {
             (_, Some(desc)) => desc, // Use provided description if available
             ("websearch", None) => "Search the web for information",
             ("get_readiness", None) => "Get the readiness status of the Spice.ai runtime",
-            ("document_similarity", None) => {
-                "Search Spice.ai datasets using Vector Similarity Search (VSS)"
-            }
+            ("search", None) => "Search Spice.ai datasets using Vector Similarity Search (VSS)",
```
lukekim: Description needs to be updated to include full-text-search (or dynamically based on what is enabled in the Spicepod?)

crates/runtime/src/request/table_provider.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+impl<F: Fn(&[KeyValue]) + Send + Sync> std::fmt::Debug for DimensionTrackedTableProvider<F> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.inner.fmt(f)
+    }
+}
+
+#[async_trait::async_trait]
+impl<F: Fn(&[KeyValue]) + Send + Sync> TableProvider for DimensionTrackedTableProvider<F> {
```
phillipleblanc: I'm hesitant to add yet another wrapped table provider - why can't we just add the tracking specifically to each table provider that needs it?

phillipleblanc: Also if we did proceed with this approach, we need to make sure to implement all of the TableProvider methods, not just the required ones.

crates/runtime/src/tools/builtin/catalog.rs
```diff
@@ -77,9 +77,7 @@ impl BuiltinToolCatalog {
             (_, Some(desc)) => desc, // Use provided description if available
             ("websearch", None) => "Search the web for information",
             ("get_readiness", None) => "Get the readiness status of the Spice.ai runtime",
-            ("document_similarity", None) => {
-                "Search Spice.ai datasets using Vector Similarity Search (VSS)"
-            }
+            ("search", None) => "Search datasets available & searchable in Spice.ai runtime",
```
phillipleblanc: This description needs to be updated.

crates/runtime/src/request/table_provider.rs
Jeadie: The key issue is that we want to track a specific metric at `TableProvider::scan`, but at that time, we don't know which metric we should emit (i.e. FTS or vector search). I suppose we could emit this metric at the construction of the `TableProvider`, if that sufficiently equivalent?

crates/runtime/src/search/candidate/vector.rs
```diff
@@ -1,3 +1,4 @@
+use crate::request::{AsyncMarker, RequestContext};
```
phillipleblanc: move to be with other imports

phillipleblanc (dismissed):
I think this is a much nicer approach!

phillipleblanc (dismissed):

peasee (approved):

Issue #7277, opened 2025-09-23
Enhancement: Initial support of S3 Vector index partitioning


kczimm:
## Goal-State/What/Result

S3 Vector indexes support `partition_by` expression using the `bucket(num_buckets, col)` expression for hash partitioning.

## Why/Purpose

Partitioned indexes in S3 Vectors enables more efficient query and higher ingestion rates due to index-level write throughput limitations.

## By When

**Issue/Spec written and reviewed:** September 23, 2025
**Done-Done:** October 2nd, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Telemetry / Metrics / Task History
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

**Out of scope/limitations**

Proper error messages will be displayed to the user when the following conditions are not met.

- Only a single expression in the `partition_by` list will be supported
- Only the `bucket(num_buckets, col)` expression will be supported. 

### Partition Naming

**Limitations**
- Maximum index name length is [63 characters](https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_CreateIndex.html#API_S3VectorBuckets_CreateIndex_RequestSyntax). 
- underscores are not supported in the index name

**Requirements**
- An index name shall include:
  1. **prefix**: identifies the dataset the index belongs to
  2. **column name**: column name identifies to dataset column that the embeddings were generated from in the index
  3. **partition value**: the value of the partition_by expression for this partition
  4. **partition_by**: expressions associated with the partition value
- Prefix and column name shall be human readable

Existing partitions will be discovered and inferred based on the index name. Discovered partitions must match the prefix, column, and partition_by expression or the name will be logged and ignored.

Partition name format:
```
{prefix[:25]}.{column[:25]}.{hex(hash(partition_value))[:5]}.{hex(hash(partition_by))[:5]}
```

Which is `25+25+5+5+3 = 63`.

[...]

PR #7279, opened 2025-09-23
Add partition_by expression to spicepod for vector stores

### LINKED ISSUE #7159: Add `partition_by` expression to `vectors`

kczimm:
## 📝 Summary

- Add `partition_by` to spicepod under `vector`
- Convert the `partition_by: Vec<String>` into `partition_by: Vec<Expr>` by validating that it uses the `bucket` UDF and references a column from the dataset schema.

## 🔗 Related

Closes:
- #7159 

Needed for:
- #7277


crates/runtime/src/embeddings/connector.rs
```diff
@@ -186,9 +198,27 @@ impl EmbeddingConnector {
         match vector_store.engine.as_deref() {
             #[cfg(feature = "s3_vectors")]
             Some("s3" | "s3_vectors") => {
+                use datafusion::prelude::SessionContext;
```
phillipleblanc, line 201: nit: move to the top of the file

crates/runtime/src/embeddings/connector.rs
```diff
@@ -293,6 +324,54 @@ impl EmbeddingConnector {
     }
 }
 
+fn get_and_validate_partition_by(
+    df_schema: &DFSchema,
+    vector_store: &VectorStore,
+    ctx: &SessionContext,
+) -> Result<Vec<Expr>, ValidationError> {
+    // Expression must use the bucket UDF with a column in the dataset
+    struct BucketCriterion;
```
phillipleblanc, line 333: Is there a benefit to defining this struct inline this function vs defining it outside of the function?

phillipleblanc (approved):

kczimm:
Incorporated into #7313. The code in this PR has changed enough when incorporated that I think it can just be reviewed there as some of the feedback no longer applies.

PR #7283, opened 2025-09-23
Make search test more robust

krinart:
## 📝 Summary

Make test which checks query duration with and without cache more robust

### DIFF
--- a/crates/runtime/tests/models/search.rs
+++ b/crates/runtime/tests/models/search.rs
@@ -1233,7 +1233,7 @@ async fn test_search_with_cache() -> Result<(), anyhow::Error> {
                 })),
             ), None, false).await?;
             let duration_cached = start.elapsed();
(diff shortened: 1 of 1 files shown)


sgrebnov (approved):

Jeadie (commented):
Do we know if this is because search is getting faster, or caching is getting slower?

peasee:
> Do we know if this is because search is getting faster, or caching is getting slower?

➕ . I would expect a more "robust" fix for this is a sliding threshold based on how fast the original query was. E.g. 50ms without cache probably won't show as significant of an improvement.

We should also consider if the test itself needs modifying to make it run a significantly slower search, so that the difference is larger to begin with.

lukekim (approved):

krinart:
I created a ticket to improve this test https://github.com/spiceai/spiceai/issues/7285

For now I am going to merge this PR as it will improve developer experience (we had to re-launch this test multiple times during the relese)

PR #7286, opened 2025-09-24
fix: Use median cached response duration for test search cache

### LINKED ISSUE #7285: Improve `test_search_with_cache`
Current implementation compares query duration with and without cache, which sometimes makes this test flaky.

See more details here https://github.com/spiceai/spiceai/pull/7283

One idea that came up in the context of that PR:

> a sliding threshold based on how fast the original query was. E.g. 50ms without cache probably won't show as significant of an improvement.

> We should also consider if the test itself needs modifying to make it run a significantly slower search, so that the difference is larger to begin with.

peasee:
## 📝 Summary

* Uses `potion-base-32M` and increases results limit to 50 to induce a longer uncached search duration
* Calculates cached search duration based on the median from 10 requests to avoid environmental conditions

## 🔗 Related

* Closes #7285

### DIFF
--- a/crates/runtime/tests/models/search.rs
+++ b/crates/runtime/tests/models/search.rs
@@ -1237,7 +1237,7 @@ async fn test_search_with_cache() -> Result<(), anyhow::Error> {
                 let duration_cached = start.elapsed();
                 measured_cache_times.push(duration_cached);
             }
-            
+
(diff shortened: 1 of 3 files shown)


Jeadie (approved):

krinart:
Thanks!

PR #7287, opened 2025-09-24
Fix JOIN level filters not having columns in schema

Jeadie:
## 📝 Summary
- If a `TableProvider` supports a filter pushdown exactly, DataFusion will not provide the column in the projection. When it is not support, DataFusion will have this column in the projection, and add a `Filter` node above.
```sql
-- Without filter pushdown
Projection: text_search(...).id, text_search(...).question1, text_search(...).score
    Filter: text_search(...).is_duplicate = Int64(0)
        TableScan: text_search(...) projection=[id, question1, is_duplicate, score]

-- With
TableScan: text_search(...) projection=[id, question1, score], full_filters=[text_search(...).is_duplicate = Int64(0)]
```

- When `SearchQueryProvider` requires to join on a base table, and has a filter that cannot be pushed into the search index, but can be pushed into the base table, and the column is not in the search index (i.e. primary key or metadata) the filters atop the associated `JOIN` node expect these columns in the `JOIN`'s schema. If they are not, an error like this occurs 
```
type_coercion: caused by: Schema error: No field named is_duplicate. Valid fields are search_index.id, search_index.score, base_table.id.
```

### DIFF
--- a/crates/runtime/src/search/full_text/mod.rs
+++ b/crates/runtime/src/search/full_text/mod.rs
@@ -24,7 +24,6 @@ use search::generation::{
     CandidateGeneration, post_apply::PostApplyCandidateGeneration,
     text_search::index::FullTextDatabaseIndex,
 };
-use snafu::ResultExt;
 
 use runtime_object_store::registry::SpiceObjectStoreRegistry;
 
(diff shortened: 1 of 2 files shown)


mach-kernel (approved):
Makes sense. Nice.

kczimm (approved):

PR #7291, opened 2025-09-24
use SessionContext::new_empty in RRF

kczimm:
## 📝 Summary

Just uses `SessionContext::new_empty` in RRF rather than creating an empty DataFrame manually.


mach-kernel (approved):
😸

PR #7293, opened 2025-09-24
Use rust:1.89-slim-bookworm for build, more places to bump rust version

sgrebnov:
## 📝 Summary

This does not affect Rust version we use for build as it is controlled by toolchain, this is just an optimization to be consistent and fix rust version initially installed (so we don't need to install 1.88 and then automatically do 1.89)

### DIFF
--- a/.github/actions/setup-rust/action.yml
+++ b/.github/actions/setup-rust/action.yml
@@ -36,12 +36,12 @@ runs:
       if: inputs.os != 'windows'
       shell: bash
       run: |
-        rustup toolchain install 1.88
-        rustup default 1.88
+        rustup toolchain install 1.89
+        rustup default 1.89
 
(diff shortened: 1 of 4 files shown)


.github/actions/setup-rust/action.yml
```diff
@@ -36,12 +36,12 @@ runs:
       if: inputs.os != 'windows'
       shell: bash
       run: |
-        rustup toolchain install 1.88
-        rustup default 1.88
+        rustup toolchain install 1.89
+        rustup default 1.89
 
     - name: Install latest Rust stable toolchain (Windows)
```
sgrebnov, line 42: We can technically remove this step and the required toolchain will be installed automatically

krinart (approved):

kczimm (approved):

lukekim (approved):

PR #7295, opened 2025-09-25
chore: remove unused Dataset methods

kczimm:
## 📝 Summary

The `Dataset.schema` field was never set and the `Dataset::with_schema` method was never called. Removed the field and methods and downstream effects.


crates/runtime/src/dataconnector/spiceai.rs
```diff
@@ -311,13 +309,12 @@ impl DataConnector for SpiceAI {
...
         };
 
-        let read_write_result =
-            ReadWrite::table_provider(&flight_factory, table_reference, dataset.schema())
-                .await
-                .context(super::UnableToGetReadWriteProviderSnafu {
-                    dataconnector: "spice.ai",
-                    connector_component: ConnectorComponent::from(dataset),
-                });
+        let read_write_result = ReadWrite::table_provider(&flight_factory, table_reference, None)
```
lukekim: 100% sure this is not used for the spice.ai connector?

kczimm: the `schema` field is not set anywhere in our codebase. I changed `pub` to `pub(crate)` for `Dataset::with_schema` and rust linted it as dead_code I removed the `Dataset::with_schema` and it still compiles. I grepped to codebase for `.with_schema(` and made sure that the calling struct was never an instance of `Dataset`. This simply isn't used. I found this out because I needed the schema of a Dataset and it was `None`, to my surprise.

crates/runtime/src/dataconnector/clickhouse.rs
```diff
@@ -233,16 +233,14 @@ impl DataConnector for Clickhouse {
         &self,
         dataset: &Dataset,
     ) -> super::DataConnectorResult<Arc<dyn TableProvider>> {
-        Ok(Read::table_provider(
-            &self.clickhouse_factory,
-            dataset.path().into(),
-            dataset.schema(),
+        Ok(
+            Read::table_provider(&self.clickhouse_factory, dataset.path().into(), None)
```
Jeadie: Should we update `ReadWrite::table_provider` since it looks like they all provide `schema: None`.

kczimm: Yeah I'll delete it all.

mach-kernel (approved):

sgrebnov (approved):

Issue #7149, opened 2025-09-15
Enhancement: INSERT INTO Iceberg Table


phillipleblanc:
## Goal-State/What/Result

The following syntax will create or append to an existing Iceberg table:

```sql
INSERT INTO { iceberg_table_name }
SELECT * FROM { table_name | query }
```

Or using VALUES syntax:

```sql
INSERT INTO { iceberg_table_name } (column1, column2, ...)
VALUES (value1, value2, ...)
```

The simplest possible INSERT INTO is (assuming `iceberg_table` already exists as an Iceberg dataset in Spice):

`INSERT INTO iceberg_table SELECT * FROM existing_table`
OR
`INSERT INTO ice.some_schema.some_table SELECT * FROM existing_table`

INSERT INTO VALUES examples:
`INSERT INTO iceberg_table VALUES (1, 'John', 100.0)`
`INSERT INTO ice.some_schema.some_table (id, name, amount) VALUES (1, 'John', 100.0), (2, 'Jane', 200.0)`

This syntax is only allowed for targets that are Iceberg datasets/catalogs with `access: read_write` set:

```yaml
catalogs:
  - from: iceberg:https://glue.ap-northeast-3.amazonaws.com/iceberg/v1/catalogs/111111/namespaces
    name: ice
    access: read_write

datasets:
  - from: iceberg:https://iceberg-catalog-host.com/v1/namespaces/my_namespace/tables/my_table
    name: iceberg_table
    access: read_write
```

> [!NOTE]
> `access: read_write` already exists for datasets (but not catalogs) as an undocumented syntax from our early experiments with replicating writes back to datasets. I'm proposing to re-use that UX here.

A query can be used instead of a simple SELECT:

```sql
INSERT INTO ice.my_schema.my_table
SELECT * FROM some_dataset WHERE city = 'Seattle'
```

Or direct value insertion:

```sql
INSERT INTO ice.my_schema.my_table (id, city, population)
VALUES (1, 'Seattle', 750000), (2, 'Portland', 650000)
```

## Why/Purpose

Enables Spice to be used to simplify customer architectures by allowing Spice to act as the writer for Iceberg datasets, instead of requiring Spark or equivalent.

## By When

**Issue/Spec written and reviewed:**  09/24
**Done-Done:** 10/01

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
[...]

PR #7304, opened 2025-09-26
Enable  DML `Update` SQL operations for datasets configured as  `access: read_write`

sgrebnov:
## 📝 Summary

1. Rename `dataset.mode` to `dataset.access` (with mode still supported for backward compatibility) 
2. Enable SQL `INSERT INTO` SQL for datasets configured with `access: read_write` mode. This mode must be supported by underlying TableProvider, otherwise `Failed to setup the dataset  ... Table <table-name> was marked as read_write, but the underlying provider only supports reads.` error will be returned. 
3. Only `INSERT` is allowed, other DML and DDL operations are still prohibited, inserting into Spice internal tables via SQL is prohibited as well.

## 🔗 Related

Part of 
- https://github.com/spiceai/spiceai/issues/7149

### DIFF
--- a/crates/runtime/src/component/dataset/mod.rs
+++ b/crates/runtime/src/component/dataset/mod.rs
@@ -551,7 +551,7 @@ impl Dataset {
     }
 
     #[must_use]
-    pub fn mode(&self) -> AccessMode {
+    pub fn access(&self) -> AccessMode {
         self.access
     }
 
--- a/crates/runtime/src/component/dataset/mod.rs
+++ b/crates/runtime/src/component/dataset/mod.rs
@@ -117,17 +117,17 @@ pub enum Error {
 pub type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Debug, Clone, Copy, PartialEq, Default)]
-pub enum Mode {
+pub enum AccessMode {
     #[default]
     Read,
     ReadWrite,
 }
 
-impl From<spicepod_dataset::Mode> for Mode {
-    fn from(mode: spicepod_dataset::Mode) -> Self {
+impl From<spicepod_dataset::AccessMode> for AccessMode {
+    fn from(mode: spicepod_dataset::AccessMode) -> Self {
         match mode {
-            spicepod_dataset::Mode::Read => Mode::Read,
-            spicepod_dataset::Mode::ReadWrite => Mode::ReadWrite,
+            spicepod_dataset::AccessMode::Read => AccessMode::Read,
+            spicepod_dataset::AccessMode::ReadWrite => AccessMode::ReadWrite,
         }
     }
 }
@@ -258,7 +258,7 @@ impl Display for CheckAvailability {
 pub struct Dataset {
     pub from: String,
     pub name: TableReference,
-    pub mode: Mode,
+    pub access: AccessMode,
     pub params: HashMap<String, String>,
     pub metadata: HashMap<String, String>,
     pub columns: Vec<Column>,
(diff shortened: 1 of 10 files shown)


crates/runtime/src/datafusion/sql_validator.rs
```diff
@@ -0,0 +1,311 @@
...
+/// Reads (SELECT queries) are allowed on all tables.
+/// INSERT operations are only allowed on datasets that are configured as writable,
+/// and are not allowed on internal Spice tables.
+/// DDL, DML (other than allowed INSERT), COPY, and Statement operations are not permitted.
+///
+/// # Returns
+/// * `Ok(())` if the plan is valid
+/// * `Err(DataFusionError)` if the plan contains invalid operations, such as DDL, DML on read-only datasets,
+///   or INSERT on internal datasets
+pub fn validate_sql_query_operations(
```
sgrebnov, line 38: This is reference implementation (previously used) by DataFusion:

https://github.com/apache/datafusion/blob/c1d6f343161ad7a3a9a372469fafd529d0085ada/datafusion/core/src/execution/context/mod.rs#L1912C1-L1933C2

```rust
impl<'n> TreeNodeVisitor<'n> for BadPlanVisitor<'_> {
    type Node = LogicalPlan;

    fn f_down(&mut self, node: &'n Self::Node) -> Result<TreeNodeRecursion> {
        match node {
            LogicalPlan::Ddl(ddl) if !self.options.allow_ddl => {
                plan_err!("DDL not supported: {}", ddl.name())
            }
            LogicalPlan::Dml(dml) if !self.options.allow_dml => {
                plan_err!("DML not supported: {}", dml.op)
            }
            LogicalPlan::Copy(_) if !self.options.allow_dml => {
                plan_err!("DML not supported: COPY")
            }
            LogicalPlan::Statement(stmt) if !self.options.allow_statements => {
                plan_err!("Statement not supported: {}", stmt.name())
            }
            _ => Ok(TreeNodeRecursion::Continue),
        }
    }
}
```

crates/runtime/src/component/dataset/mod.rs
```diff
@@ -551,8 +551,8 @@ impl Dataset {
     }
 
     #[must_use]
-    pub fn mode(&self) -> Mode {
-        self.mode
+    pub fn mode(&self) -> AccessMode {
```
phillipleblanc:
```suggestion
    pub fn access(&self) -> AccessMode {
```

crates/runtime/src/datafusion/mod.rs
```diff
@@ -505,7 +506,7 @@ impl DataFusion {
             }
         };
 
-        if matches!(dataset_mode, Mode::ReadWrite) {
+        if matches!(dataset_mode, AccessMode::ReadWrite) {
```
phillipleblanc:
```suggestion
        if matches!(dataset_access_mode, AccessMode::ReadWrite) {
```

crates/runtime/src/datafusion/sql_validator.rs
```diff
@@ -0,0 +1,345 @@
...
+                        "INSERT operations are not allowed on Spice system dataset '{}'.",
+                        dml.table_name
+                    );
+                }
+
+                if df.is_writable(&dml.table_name) {
+                    Ok(TreeNodeRecursion::Continue)
+                } else {
+                plan_err!(
+                    "INSERT operations are not allowed on read-only dataset '{}'. Verify the dataset is configured as writable and try again.",
```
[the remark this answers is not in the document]
phillipleblanc: I think this is a good suggestion

phillipleblanc (dismissed):

crates/runtime/src/component/dataset/mod.rs
sgrebnov: Updated, thank you

crates/runtime/src/datafusion/mod.rs
sgrebnov: Updated, thank you

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #7311, opened 2025-09-26
For `vector_search`, use a default limit of 1000 if no limit specified

lukekim:
This pull request updates the logic for determining the row limit in vector search operations and ensures consistent handling of limit values. The main changes are focused on improving default behavior and type consistency for the limit parameter.

Limit handling improvements:

* The `limit_to_use` method in `VectorSearchUDTFProvider` now always returns a `usize` instead of an `Option<usize>`, and provides a default value of `1000` when no limit is specified. This ensures a sensible default and simplifies downstream usage.
* The `fetch` field assignment in the `TableProvider` implementation is updated to wrap the result of `limit_to_use` in `Some`, matching the expected type and preventing possible errors.


mach-kernel (dismissed):
Probably a good idea

krinart (approved):

PR #7313, opened 2025-09-26
Read and write partitioned S3 indexes

### LINKED ISSUE #7292: Write partitioned indexes

kczimm:
## 📝 Summary

Adds `partition_by` to vector engine s3 enabling partitioned index writes and reads supported by `vector_search`. 

## 🔗 Related

Closes:
- #7292 

## 👀 Notes for Reviewers

No partition pruning is implemented and full partition union scans are performed.


crates/data_components/src/s3_vectors/query_provider.rs
```diff
@@ -146,38 +156,137 @@ impl TableProvider for S3VectorsQueryTable {
...
+
+        let (_, Some(bucket_name), Some(index_name)) = self.table.idx.index_identifier_variables()
+        else {
+            return exec_err!("No bucket name or index name for bucket query");
+        };
+
+        let list_indexes_output = self
+            .table
+            .client
+            .list_indexes(
```
kczimm, line 215: We have to discover the partitioned indexes every scan here because QueryTableProvider is created anew each invocation of `vector_search`.

crates/runtime-table-partition/src/insert.rs
```diff
@@ -405,7 +405,11 @@ impl DisplayAs for PartitionInputExec {
 /// Evaluate the `physical_expr` for each row in `batch`. A partition batch is
 /// created for each unique value produced by evaluating the expression
 /// containing the rows that produced that unique partition value.
-fn partition_batch(
+///
+/// # Errors
+/// Returns an error when the expressions cannot be evaluated, the batch cannot
+/// be partitioned, Arrays cannot be created or the batch cannot be filtered.
+pub fn partition_batch(
```
kczimm, line 412: Because we aren't reusing `PartitionedTableProvider` yet, this partitioning logic has to be exposed.

crates/runtime/src/embeddings/index/s3/index.rs
```diff
@@ -132,7 +146,68 @@ impl SearchIndex for S3Vector {
...
+                let partitions = partition_batch(&record, physical_expr.as_ref())?;
+
+                for (partition_value, partition_record) in partitions.into_values() {
+                    // change the index name to a partition name
+                    let id = match &self.table.idx {
+                        S3VectorIdentifier::IndexArn(_) => {
+                            tracing::debug!(
+                                "Partitioning is not supported when index ARN is provided. Please provide the bucket and index name instead."
+                            );
+                            return write::write(self, &self.table, record).await.boxed();
```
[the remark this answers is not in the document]
kczimm: The only processing is the partitioning itself, which is not needed in this case.

crates/runtime/src/embeddings/index/s3/index.rs
```diff
@@ -132,7 +146,68 @@ impl SearchIndex for S3Vector {
...
+                        DataFusionError::Execution(
+                            "S3 vector index could not be read or created".to_string(),
+                        )
+                    })?;
+
+                    write::write(self, &table, partition_record).await.boxed()?;
+                }
+            }
+            None => {
+                return write::write(self, &self.table, record).await.boxed();
```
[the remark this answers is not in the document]
kczimm: No

crates/data_components/src/s3_vectors/list_provider.rs
```diff
@@ -95,12 +108,116 @@ impl TableProvider for S3VectorsListTable {
...
+        state: &dyn Session,
         projection: Option<&Vec<usize>>,
-        _filters: &[Expr],
+        filters: &[Expr],
         limit: Option<usize>,
     ) -> DataFusionResult<Arc<dyn ExecutionPlan>> {
-        Ok(Arc::new(S3VectorsListExec::new(self, projection, limit)) as Arc<dyn ExecutionPlan>)
+        if self.partition_by.is_empty() {
+            return Ok(
+                Arc::new(S3VectorsListExec::new(self, projection, limit)) as Arc<dyn ExecutionPlan>
```
sgrebnov, line 118: Learning purpose - does it mean basic `S3VectorsListExec` does not support `filters`? as we are not passing/using them here

crates/data_components/src/s3_vectors/query_provider.rs
```diff
@@ -146,38 +156,137 @@ impl TableProvider for S3VectorsQueryTable {
...
-        Ok(Arc::new(S3VectorsQueryExec::new(
-            self,
-            projection,
-            limit,
-            query_vector,
-            filters.to_vec(),
-        )) as Arc<dyn ExecutionPlan>)
+        if self.partition_by.is_empty() {
+            #[allow(clippy::cast_possible_wrap)]
+            let limit: i64 = match limit {
```
sgrebnov: nit: this can be moved into helper function to make main logic more compact

crates/data_components/src/s3_vectors/query_provider.rs
```diff
@@ -146,38 +156,137 @@ impl TableProvider for S3VectorsQueryTable {
             .collect())
     }
 
+    #[allow(clippy::too_many_lines)]
```
sgrebnov: Is there an opportunity for some refactoring to get rid of #[allow(clippy::too_many_lines)]?

crates/runtime/src/embeddings/connector.rs
```diff
@@ -181,6 +181,7 @@ impl EmbeddingConnector {
         Ok(Arc::new(embedding_table) as Arc<dyn TableProvider>)
     }
 
+    #[allow(clippy::too_many_lines)]
```
sgrebnov: Can we simplify wrap_table_as_index to avoid #[allow(clippy::too_many_lines)]?

crates/data_components/src/s3_vectors/list_provider.rs
```diff
@@ -95,12 +108,116 @@ impl TableProvider for S3VectorsListTable {
 
     async fn scan(
```
sgrebnov, line 109: Should we add some unit tests to cover Partitioned Plan generation (plan snapshot)?

crates/data_components/src/s3_vectors/list_provider.rs
kczimm: This is correct. The current `S3VectorsListExec` does not support filtering.

crates/data_components/src/s3_vectors/list_provider.rs
```diff
@@ -95,12 +108,116 @@ impl TableProvider for S3VectorsListTable {
...
+                client: Arc::clone(&self.table.client),
+                schema: Arc::clone(&self.table.schema),
+                constraints: self.table.constraints.clone(),
+                idx: index_table_identifier,
+                dimension: self.table.dimension,
+                columns: self.table.columns.clone(),
+                distance_metric: self.table.distance_metric.clone(),
+            };
+
+            let list_table = S3VectorsListTable::new(index_table, self.column_name.clone(), vec![]);
```
Jeadie, line 213: The fact that we make `S3VectorsListTable` internally makes me think we should split `S3VectorsListTable` from a `S3VectorsListPartitionedTable`, since
1. Most of the logic is in `S3VectorsTable`
2. When we recursively call `S3VectorsListTable::scan` on L206, that means we make an additional `list_indexes` per partition index too.

crates/data_components/src/s3_vectors/list_provider.rs
```diff
@@ -32,35 +33,47 @@ use arrow::{
...
 use s3_vectors_metadata_filter::document_to_json_map;
 use snafu::ResultExt;
 use tokio::sync::mpsc::Sender;
 
 /// An S3 Vector index that implements a [`TableProvider`] as a list records operation.
 #[derive(Debug, Clone)]
-pub struct S3VectorsListTable(S3VectorsTable);
+pub struct S3VectorsListTable {
+    table: S3VectorsTable,
+    column_name: String,
```
Jeadie, line 65: This is the search column to use in hashing the index name to ensure we're using the appropriate partition? Is having this column explicitly needed. Often the default index name has this search field.

crates/data_components/src/s3_vectors/vector_table.rs
```diff
@@ -58,6 +58,10 @@ pub struct S3VectorsTable {
     pub schema: SchemaRef,
 
     pub(super) constraints: Constraints,
+
+    pub dimension: i64,
```
Jeadie, line 62: I'm not sure where this `dimension` gets used?

crates/runtime/src/embeddings/index/s3/index.rs
```diff
@@ -132,7 +146,68 @@ impl SearchIndex for S3Vector {
...
+                                partition_record.num_rows(),
+                            );
+                            S3VectorIdentifier::Index {
+                                bucket_name: bucket_name.clone(),
+                                index_name,
+                            }
+                        }
+                    };
+
+                    let table = S3VectorsTable::try_create_new_table(
```
Jeadie, line 188: I think if we used `.try_new_table(` and not `.try_create_new_table(`, we wouldn't need to store `dimensions`, `columns` or `distance_metric`?

kczimm:
This is really a stop gap solution to partitioning. We should be using the `PartitionedTableProvider` but as far as I can tell, we don't have an `S3VectorsTable` `TableProvider` that uses `PutVectors` on the `insert_into` and either `ListVectors` or `QueryVectors` based on the filter expressions in `scan`. However, I believe this probably requires a significant redesign of vectors because the read (two different kinds) and write paths are separated and we create new instances of `S3VectorsQueryTable` `TableProvider` for every invocation of `vector_search` rather than registering them in the `SessionContext`.

Jeadie (commented):
Not for this PR, but one thing worth contemplating is how we plan to support partitioned vector indexes when we have more than just S3Vectors (which is soon ™). We don't want to enforce knowing about partitioning on every new index creator.

sgrebnov (dismissed):

crates/data_components/src/s3_vectors/list_provider.rs
```diff
@@ -95,12 +108,116 @@ impl TableProvider for S3VectorsListTable {
...
+
+        let (_, Some(bucket_name), Some(index_name)) = self.table.idx.index_identifier_variables()
+        else {
+            return exec_err!("No bucket name or index name for bucket query");
+        };
+
+        let list_indexes_output = self
+            .table
+            .client
+            .list_indexes(
```
phillipleblanc, line 134: This should use the `prefix` parameter so that we aren't enumerating the entire vector bucket on every query. https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_ListIndexes.html

phillipleblanc: If there isn't a common prefix that allows us to only get the indexes that are part of this partitioned set, then we should probably redesign the index naming so that we can.

crates/data_components/src/s3_vectors/list_provider.rs
```diff
@@ -95,12 +108,116 @@ impl TableProvider for S3VectorsListTable {
...
+            .indexes()
+            .iter()
+            .filter_map(|idx| {
+                let Ok(partitioned_index_name) =
+                    PartitionedIndexName::from_index_name(idx.index_name())
+                else {
+                    return None;
+                };
+
+                if matches!(
```
phillipleblanc, line 162: If we got the prefix output correct, then in theory this should never filter anything out, so might be worth a debug log if we do end up filtering out an index that matches the prefix filter but not this one.

crates/data_components/src/s3_vectors/list_provider.rs
phillipleblanc: Yeah, I think having a separate partitioned S3 vectors provider that creates the non-partitioned provider per index is a cleaner separation. Though for point 2, even in the current implementation it wouldn't do an additional `list_indexes` call since the partitions list is empty.

crates/data_components/src/s3_vectors/list_provider.rs
```diff
@@ -110,12 +227,6 @@ impl std::fmt::Debug for S3VectorsListExec {
     }
 }
 
-impl AsRef<S3VectorsTable> for S3VectorsListTable {
```
phillipleblanc, line 113: Why this change?

crates/data_components/src/s3_vectors/query_provider.rs
```diff
@@ -148,36 +185,115 @@ impl TableProvider for S3VectorsQueryTable {
...
+        let (_, Some(bucket_name), Some(index_name)) = self.table.idx.index_identifier_variables()
+        else {
+            return exec_err!("No bucket name or index name for bucket query");
+        };
+
+        let list_indexes_output = self
+            .table
+            .client
+            .list_indexes(
+                ListIndexesInput::builder()
```
phillipleblanc, line 216: similar to my previous comment, this should use prefix filtering

phillipleblanc: this logic is almost exactly the same as the ListTable one except for the underlying struct it creates - maybe worth a generic function or a function with a callback that creates the `S3VectorsQueryTable`/`S3VectorsListTable`

crates/data_components/src/s3_vectors/list_provider.rs
kczimm: Prior to this PR, `S3VectorsListTable` was just a wrapper for `S3VectorsTable`. It has since changed from a tuple struct to a named field struct. Of course, we could still have this `impl` but I think it's better to be explicit rather than the code appear magical. It's clearer to the reader to see `self.table` and know that we are referring to the `S3VectorsListTable` rather than seeing `&self` and it may refer to `S3VectorsTable` or it may be referring to `S3VectorsListTable`.

crates/data_components/src/s3_vectors/list_provider.rs
kczimm: > This is the search column to use in hashing the index name to ensure we're using the appropriate partition?

Yes.

> Often the default index name has this search field.

Is this the default we generate? Because we need the column name to uniquely identify partitions that belong.

I guess, is there something wrong with storing the column name here? Is the suggestion that it come from somewhere else to make this struct smaller?

crates/data_components/src/s3_vectors/list_provider.rs
kczimm: Thanks for pointing out this feature of the API. I didn't know it was there.

crates/data_components/src/s3_vectors/vector_table.rs
kczimm: it is used to create other `S3VectorsTable`s

crates/runtime/src/embeddings/index/s3/index.rs
kczimm: But we need to create the index if it doesn't exist. `try_create_new_table` does that. `try_new_table` does not.

kczimm:
> Not for this PR, but one thing worth contemplating is how we plan to support partitioned vector indexes when we have more than just S3Vectors (which is soon ™). We don't want to enforce knowing about partitioning on every new index creator.

https://github.com/spiceai/spiceai/pull/7313#issuecomment-3363042724

sgrebnov (approved):

crates/data_components/src/s3_vectors/list_provider.rs
kczimm: Can follow up in a separate PR if necessary.

crates/runtime/src/embeddings/index/s3/index.rs
Jeadie: Sorry, yes correct.

crates/data_components/src/s3_vectors/list_provider.rs
Jeadie: Nothing wrong, mainly just a point that is less important prior to my confusion around needing variables for `try_create_new_table`.

Issue #6555, opened 2025-07-22
Enhancement: Support chunking in vector engines and UDTFs


Jeadie:
## Goal-State/What/Result

`vector_search` UDTF supports datasets with chunked embedding columns. For this, it is required then that vector engines (e.g. S3 vector) can properly handle chunked data too.

## Why/Purpose

Make UDTFs and vector engines feature complete with current HTTP `v1/search` APIs.

## By When

**Issue/Spec written and reviewed:** 10th September
**Done-Done:** 15th September

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [ ] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

### Design
 - Underlying vector indexes should be unaware of chunking. (i.e. `impl VectorIndex` should not be effected).
 - Unify search query pattern of `VectorQueryTableProvider` and `TextSearchIndexProvider`
 - New index middleware to convert table index `([pk1, pk2,...], search_field) -> ([pk1, pk2,..., chunk_id], chunked_search_field)`. 

#### Indexing
 - `impl Index` needs to be responsible for chunking incoming `RecordBatch`s, and then passing into underlying `Index`.
 
#### Querying
 - Both `SearchQueryProvider` and `VectorScanTableProvider` must handle chunking during `JOIN ON` operation.
   - For `SearchQueryProvider` change is 
     - Handle extra internal primary key column (chunk id)
     - Handle search field appropriately (i.e. know when index has full search field as metadata and when it doesnt). 
   -  `VectorScanTableProvider` 
     - Handle extra internal primary key column (chunk id).
     - Handle constructing array of vectors (i.e. aggregate vectors)
 - Schema of `VectorScanTableProvider` as per [using existing embeddings](https://spiceai.org/docs/features/search/vector-search#using-existing-embeddings).
 - `vector_search` UDTF returns additional optional column `match`. Schema of `VectorQueryTableProvider` changes accordingly.
   - For non-chunked, `match` will not be present
   - For chunked, will contain the most relevant chunk from the returned row.
   - For chunked, content offsets stored as non-fiterable in index.

### Security Review
[...]

PR #7314, opened 2025-09-27
Use logical plan in `SearchQueryProvider`. 

### LINKED ISSUE #7242: Refactor `SearchQueryProvider` to be simple join between LogicalPlan/TableProvider
Currently `SearchQueryProvider` performs significant logic upon a `SearchIndex` and `TableProvider`. This causes the base `table_provider` to determine the additional schema that is present in the `SearchQueryProvider::schema`. We should move to an approach where `SearchQueryProvider` is simply joining the base `table_provider` with the `SearchIndex:query_provider` logical plan.

Now
```rust
pub struct SearchQueryProvider {
    pub search_index: Arc<dyn SearchIndex>,
    pub table_provider: Arc<dyn TableProvider>,
    pub query: String,
    pub pre_limit: Option<usize>,
}
```

Future
```rust
pub struct SearchQueryProvider {
    pub search_index: Arc<LogicalPlan>,
    pub table_provider: Arc<dyn TableProvider>,
}
```

This would unblock providing the additional columns for chunked, vector index-based datasets in regular table scans (i.e. `_embeddings`, `_offsets`). i.e. Currently since the base table determines the schema of `SearchQueryProvider`, the base table cannot also provide nested data (i.e. list of vectors for a row) without significant incompatibilities with other search patterns in `SearchQueryProvider`.

Jeadie:
## 📝 Summary
 - The core change in this PR is that during the search query path we use a premade `LogicalPlan` instead of providing a full SearchIndex. This seems like a minor change, but it allows us to have the [`TableProvider::schema`] use _and only use_ the schema of the search index's query logical plan. To achieve this, we also:
   - A new `ComputeQueryVector` that allows us to delay the computation of the query vector for vector indexes until the `TableProvider::scan` operation (for `S3VectorsQueryTable`). 
   - Change the `SearchIndex::query_table_provider` trait to both 1. Be non-async, 2. Return a `LogicalPlan` and not a `TableProvider`. 

## 🔗 Related
 - Resolved #7242
 - Needed for #6555


crates/runtime/tests/models/snapshots/integration_models__models__search__s3vectors_basic_with_where_response.snap
```diff
@@ -25,6 +25,26 @@ snapshot_kind: text
         "id": 6
       },
       "score": 0.82
+    },
```
Jeadie, line 28: This is an interesting find. There were two reasons that caused this to previous miss some data
1. We would not push down filters to the base table that the base table could not support
2. We applied the `limit` in the HTTP request payload as the pre limit even if there were filters. 

**Part 1**
This PR has set `supports_filter_pushdown` to Exact for all filters (in `crates/search/src/provider.rs`). Previously we only pushed down filters for the base table that the base table could handle. What this meant, is that for filters it could not, they were only applied post join (i.e. in DF, but not in the SearchQueryProvider TableProvider). We could not push them down. 

Because of this, any filter on the base table would be applied post join (importantly after we have a prelimit filter). Now we handle all filters, and in this case can still push down this filter (even if we still have to reapply post-join). See this section of an equivalent SQL explain plan [integration_models__models__search__s3vectors_basic_vector_search_sql_filters_explain.snap](https://github.com/spiceai/spiceai/pull/7314/files#diff-69edf888b7c040bb357f28d5d93b3b5968679b276e463991f8df25bed7f99039L30-R31)

**Part 2** 
This was a simple bug on the HTTP input parameters (i.e. no affecting SQL UDTFs). See this [commit](https://github.com/spiceai/spiceai/pull/7314/commits/202a7aede6072676ff7e4915534ef1b16871938d) in this PR:

crates/runtime/tests/models/snapshots/integration_models__models__s3_vectors__search__filters_pushdown_explain.snap
```diff
@@ -10,23 +10,21 @@ snapshot_kind: text
...
 |               |       BytesProcessedNode                                                                                                                                                                                                                                                                                                                                                                                        |
-|               |         TableScan: vector_search(delivery, Utf8("wrong location")) projection=[message.body, message.status, attempt_count, package_weight_kg, score], full_filters=[vector_search(delivery, Utf8("wrong location")).attempt_count > Int64(1), vector_search(delivery, Utf8("wrong location")).package_weight_kg > Float64(5), vector_search(delivery, Utf8("wrong location")).message.status = Utf8("FAILED")] |
```
Jeadie, line 13: The primary difference in these snapshots should be
1. The ordering off projection column names (now alphabetical)
2. In the logical plan, no filter above the table scan (it'll be a full_filter property of the table scan).

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #7318, opened 2025-09-28
Refactor 'VectorScanTableProvider' to use just 'VectorIndex::list_table_provider'

Jeadie:
## 📝 Summary
 - Similar to #7314, refactor `VectorScanTableProvider` to be a simple JOIN on the base table and `VectorIndex::list_table_provider`.
 - In this case, we prioritise the base table, and augment it anything extra in `VectorIndex::list_table_provider'. We also push down filters and avoid joins. 
 - Additionally, we must augment the `ChunkedSearchIndex` `write` function to reconstruct the columns we need to write to the underlying accelerator (i.e. `{}_embedding` and `{}_offset`). 

## 🔗 Related
 #6555


crates/search/src/chunking.rs
```diff
@@ -361,13 +416,13 @@ impl SearchIndex for ChunkedSearchIndex {
 }
 
 fn to_offset_array(x: &[Vec<(usize, usize)>], nullable: bool) -> FixedSizeListArray {
-    let mut builder = FixedSizeListBuilder::new(UInt64Builder::new(), 2)
```
Jeadie, line 350: This is just a bug (i,e, using Uint64, not Int32), it did't affect the writing to S3, but now that we are reading them, we get schema mismatches.

crates/search/src/chunking.rs
```diff
@@ -390,41 +445,46 @@ impl std::fmt::Debug for ChunkedVectorIndex {
 
 #[async_trait]
 impl VectorIndex for ChunkedVectorIndex {
-    fn list_table_provider(&self) -> Result<LogicalPlan, Box<dyn std::error::Error + Send + Sync>> {
+    fn list_table_provider(&self) -> Result<LogicalPlan, DataFusionError> {
```
Jeadie, line 441: Since we couldn't adequately test this, only issue was 
1. We needed to alias columns.
2. Don't group by our special chunking primary key

crates/runtime/src/embeddings/index/scan_table.rs
```diff
@@ -14,136 +14,143 @@ See the License for the specific language governing permissions and
...
     scalar::ScalarValue,
     sql::TableReference,
 };
-use datafusion_expr::SubqueryAlias;
+use datafusion_expr::{SubqueryAlias, col};
 
-use crate::{embedding_col, embeddings::index::search_index_table_is_sufficient};
-use search::{generation::util::append_fields, index::VectorIndex};
+use search::index::VectorIndex;
+use tract_core::tract_data::itertools::Itertools;
```
phillipleblanc: Don't use itertools from tract_core

crates/runtime/src/embeddings/index/scan_table.rs
```diff
@@ -14,136 +14,143 @@ See the License for the specific language governing permissions and
...
     }
 
-    /// For a projection relative to [`VectorScanTableProvider`], check if the embedding column is being requested.
-    fn need_vector_column(&self, projection: Option<&Vec<usize>>) -> bool {
-        let Some(proj) = projection else {
-            return true; // None projection -> "SELECT *".
+    async fn apply_proj_and_filter(
+        &self,
+        state: &dyn Session,
+        lp: Arc<LogicalPlan>,
```
phillipleblanc:
```suggestion
        input: Arc<LogicalPlan>,
```

crates/runtime/src/embeddings/index/scan_table.rs
```diff
@@ -14,136 +14,143 @@ See the License for the specific language governing permissions and
...
+            projection
+                .iter()
+                .sorted_unstable()
+                .map(|p| Expr::Column(Column::new_unqualified(p.clone())))
+                .collect(),
+            filtered,
+        )?);
 
-        proj.contains(&idx)
+        state.create_physical_plan(&lp).await
```
phillipleblanc: Is there a reason we need to create the physical plan here instead of just returning a LogicalPlan and doing the physical plan creation once?

phillipleblanc: We know there is a bug with running the physical optimizer more than once, so we should aim to only run it once.

crates/runtime/src/embeddings/index/scan_table.rs
```diff
@@ -153,17 +160,28 @@ impl TableProvider for VectorScanTableProvider {
...
+            }
+        }
+
+        let mut fields = fields_map.values().cloned().collect::<Vec<_>>();
+        fields.sort_unstable();
+        Arc::new(Schema::new(fields))
     }
 
     fn constraints(&self) -> Option<&Constraints> {
+        // TODO use self.primary_key
```
phillipleblanc: Is there a tracking issue for this?

crates/runtime/src/embeddings/index/scan_table.rs
```diff
@@ -179,145 +197,149 @@ impl TableProvider for VectorScanTableProvider {
...
+            TableReference::parse_str("base_table"),
+            Arc::new(DefaultTableSource::new(Arc::clone(&self.table_provider))),
+            None,
+            vec![],
+            None,
+        )?);
+
+        // Only include fields from index that aren't in base table (including metadata), except primary key.
+        let index_lp = LogicalPlan::SubqueryAlias(SubqueryAlias::try_new(
+            Arc::new(LogicalPlan::Projection(Projection::try_new(
```
phillipleblanc, line 239: DataFrame API might be more ergonomic to use here?

crates/runtime/src/embeddings/index/scan_table.rs
Jeadie: This function is intended to be for early exit, but can move.

crates/runtime/src/embeddings/index/scan_table.rs
Jeadie: This isn't needed

crates/runtime/src/embeddings/index/scan_table.rs
Jeadie: I will give this a try (for this and SearchQueryProvider) in a separate PR.

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #7330, opened 2025-09-28
Enable DML (INSERT INTO) operations for catalogs configured as `access:read_write`

sgrebnov:
## 📝 Summary

PR adds support for `access` mode for catalogs (previously supported for Datasets only) and enables `INSERT INTO` SQL for catalogs  configured with `access: read_write` mode. The mode must be supported by underlying TableProvider.

## 🔗 Related

Part of https://github.com/spiceai/spiceai/issues/7149

Implements
- https://github.com/spiceai/spiceai/issues/7321

### DIFF
--- a/crates/runtime/src/component/catalog.rs
+++ b/crates/runtime/src/component/catalog.rs
@@ -21,7 +21,7 @@ use spicepod::{component::catalog as spicepod_catalog, param::Params};
 use std::{collections::HashMap, sync::Arc};
 
 use super::{find_first_delimiter, validate_identifier};
-use crate::Runtime;
+use crate::{Runtime, component::dataset::AccessMode};
 
 #[derive(Debug, Snafu)]
 pub enum Error {
@@ -42,6 +42,7 @@ pub struct Catalog {
     pub catalog_id: Option<String>,
     pub from: String,
     pub name: String,
+    pub access: AccessMode,
     pub(crate) orig_include: Vec<String>,
     pub include: Option<GlobSet>,
     pub params: HashMap<String, String>,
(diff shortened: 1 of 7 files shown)


crates/runtime/src/component/catalog.rs
```diff
@@ -21,7 +21,7 @@ use spicepod::{component::catalog as spicepod_catalog, param::Params};
 use std::{collections::HashMap, sync::Arc};
 
 use super::{find_first_delimiter, validate_identifier};
-use crate::Runtime;
+use crate::{Runtime, component::dataset::AccessMode};
```
phillipleblanc, line 24: Given this is now common across datasets and catalogs, let's pull it out into a common module.

crates/runtime/src/datafusion/sql_validator.rs
```diff
@@ -55,6 +55,17 @@ pub fn validate_sql_query_operations(
                     );
                 }
 
+                // Check if attempting to write into a catalog. The default catalog name indicates writing to a table registered as a dataset, not via a catalog.
+                if let Some(catalog) = dml.table_name.catalog() && catalog != super::SPICE_DEFAULT_CATALOG {
```
sgrebnov, line 59: There is no simple way to determine whether the target table refers to a dataset or is part of a catalog definition. To simplify the implementation, and since a dataset cannot be registered under a custom catalog name, we use the catalog name to decide whether to verify dataset access mode or catalog configuration access mode.

In theory, a catalog could be defined as spice, which would not work under this approach. We also assume that multiple catalogs with the same name cannot be defined.

phillipleblanc (approved):

Jeadie (approved):

crates/runtime/src/component/catalog.rs
sgrebnov: Good point, thank you, confirm I'll do this as part of next PR with additional integration tests for catalog support

PR #7339, opened 2025-09-29
RRF: Fix ident resolution for struct fields, autohashed join key for varying types

mach-kernel:
## 📝 Summary

Issues:

- With a customer dataset, @lukekim observed:
  - Columns with names like `some_struct.foo` would fail to resolve
  - `join_key` had to be specified, otherwise structs, lists, etc would fail the auto row ID generator coercions

Fixes:

- `Column::new` with a table reference works better, does not require manual quoting even for cased examples
- A new UDF `digest_many` was introduced to simplify RRF row ID generation at call site, and encapsulate all of "into str" coercion for the hashing rows.


crates/runtime-datafusion-udfs/src/digest_many.rs
```diff
@@ -0,0 +1,235 @@
...
+        ]
+        .into_iter()
+        .enumerate()
+        .map(|(i, e)| e.alias(format!("c{i}")))
+        .collect::<Vec<_>>();
+
+        // All supported core Datafusion hash functions
+        let hash_functions = ["md5", "sha224", "sha256", "sha384", "sha512"];
+
+        let hash_exprs = hash_functions
```
mach-kernel, line 214: This is the new UDF and the best illustration of how to use it

crates/runtime/src/search/rrf.rs
```diff
@@ -106,6 +104,9 @@ macro_rules! col_qualified {
     ($column:expr) => {
         Expr::Column(Column::from_qualified_name_ignore_case($column))
     };
+    ($table:expr, $column:expr) => {
```
mach-kernel, line 107: When there exists anything that is `Into<TableReference>`, `col_qualified!` now uses this to correctly lift identifiers into the expr

crates/runtime/src/search/rrf.rs
```diff
@@ -421,11 +424,8 @@ impl ReciprocalRankFusion {
                     None
                 } else {
                     Some(
-                        first_value(
-                            col_qualified!(&cname),
-                            vec![col("fused_score").sort(false, false)],
-                        )
-                        .alias(&cname),
+                        first_value(ident(&cname), vec![col("fused_score").sort(false, false)])
```
mach-kernel, line 427: Ident fix: can use `ident` directly here, as it will be unique via the previous coalesce select expr

crates/runtime/src/search/rrf.rs
```diff
@@ -946,6 +948,34 @@ mod tests {
         Ok(ExitCode::SUCCESS)
     }
 
+    #[tokio::test(flavor = "multi_thread")]
+    async fn test_fuse_queries_auto_hash_and_special_idents() -> Result<ExitCode> {
```
mach-kernel, line 952: Fails on trunk

crates/runtime/src/search/rrf.rs
```diff
@@ -489,13 +489,12 @@ impl ReciprocalRankFusion {
 
     // Given a DF with overlapping unqualified names (as produced by JOIN), where column values
     // are equivalent, return the first (arbitrary) qualified name.
-    fn first_qualified_field(df: &DataFrame, name: &str) -> Result<String> {
+    fn first_qualified_field(df: &DataFrame, name: &str) -> Result<(TableReference, String)> {
```
mach-kernel, line 492: This now returns a `(ref, name)` tuple instead of rasterizing it to a string to make handling table references easier with `Column`.

kczimm (dismissed):

crates/runtime-datafusion-udfs/src/digest_many.rs
```diff
@@ -0,0 +1,235 @@
...
+            .map(|c| match c {
+                ColumnarValue::Array(array) => array
+                    .to_data()
+                    .buffers()
+                    .iter()
+                    .flat_map(|b| b.iter().map(|b| format!("{b:02X}")))
+                    .collect::<String>(),
+                ColumnarValue::Scalar(scalar) => format!("{scalar}"),
+            })
+            .collect::<String>();
```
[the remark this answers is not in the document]
mach-kernel: Done!

crates/runtime-datafusion-udfs/src/digest_many.rs
```diff
@@ -0,0 +1,235 @@
...
+        // Collect variadic args into one string for hashing
+        let hash_me: String = args
+            .into_iter()
+            .map(|c| match c {
+                ColumnarValue::Array(array) => array
+                    .to_data()
+                    .buffers()
+                    .iter()
+                    .flat_map(|b| b.iter().map(|b| format!("{b:02X}")))
+                    .collect::<String>(),
```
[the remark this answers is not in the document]
mach-kernel: In the future, I think we would just hash the slices directly against the underlying hasher. But not as easy to do as this in the moment pre-release.

kczimm (dismissed):

crates/runtime-datafusion-udfs/src/digest_many.rs
```diff
@@ -0,0 +1,240 @@
...
+            exec_err!("{DIGEST_UDF_NAME}: cannot determine return type")
+        }
+    }
+
+    fn invoke_with_args(&self, scalar_args: ScalarFunctionArgs) -> DataFusionResult<ColumnarValue> {
+        let mut args = scalar_args.args;
+        let hash_fn = Self::concrete_hash_function(args.pop())?;
+
+        // Collect variadic args into one string for hashing
+        let mut hash_me = String::with_capacity(32 * args.len());
```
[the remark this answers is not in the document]
mach-kernel: Would rather expand more than overallocate

kczimm (dismissed):

sgrebnov (dismissed):

kczimm (approved):

Issue #7184, opened 2025-09-16
Enhancement: Expose max data timestamp metrics for Iceberg-backed datasets


lukekim:
## Goal-State/What/Result

Expose Prometheus metrics that report:

- The maximum data timestamp present in the source Iceberg table
- The maximum data timestamp known to the Spice acceleration
- Computed ingestion lag (difference in seconds)

These enable clear, dashboardable visibility into ingestion freshness/lag for Iceberg-backed datasets.

## Why/Purpose

- Operators need to know how far Spice is behind the latest data in Iceberg
- Today this is derivable via SQL but is cumbersome and not standardized across dashboards
- Spice already computes these timestamps during refresh; surfacing them is low-risk, high-value
- Enables SLOs/alerts on freshness for production workloads

## By When

**Issue/Spec written and reviewed:** September 24, 2025
**Done-Done:** September 29, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

For any accelerated dataset, if `time_column` is defined, we will emit four metrics after acceleration refresh for `full`/`append` modes:
```rust
pub(crate) static MAX_TIMESTAMP_BEFORE_REFRESH_MS: LazyLock<Gauge<u64>> = LazyLock::new(|| {
    METER
        .u64_gauge("dataset_acceleration_max_timestamp_before_refresh_ms")
        .with_description("Maximum value of the dataset's time_column before the refresh operation, in milliseconds.")
        .build()
});

pub(crate) static MAX_TIMESTAMP_AFTER_REFRESH_MS: LazyLock<Gauge<u64>> = LazyLock::new(|| {
    METER
        .u64_gauge("dataset_acceleration_max_timestamp_after_refresh_ms")
        .with_description("Maximum value of the dataset's time_column after the refresh operation, in milliseconds.")
        .build()
});

[...]

PR #7347, opened 2025-09-29
Update example datadog config to include Refresh/Ingestion Lag

krinart:
## 📝 Summary

Update example datadog config to include Refresh/Ingestion Lag

spice_datadog_dashboard

## 🔗 Related

https://github.com/spiceai/spiceai/issues/7184

### DIFF
--- a/monitoring/datadog-dashboard.json
+++ b/monitoring/datadog-dashboard.json
@@ -206,7 +206,7 @@
     {
       "id": 5713842188321988,
       "definition": {
-        "title": "Ingestion/Refresh Lag",
+        "title": "Ingestion + Refresh Lag",
         "title_size": "16",
         "title_align": "left",
         "show_legend": true,
@@ -243,6 +243,7 @@
                     "unit_name": "second"
                   }
                 },
+                "alias": "Ingestion Lag",
                 "formula": "query1 / 1000"
               },
               {
@@ -252,6 +253,7 @@
                     "unit_name": "second"
                   }
                 },
+                "alias": "Refresh Lag",
                 "formula": "query4 / 1000"
               }
             ],
--- a/monitoring/datadog-dashboard.json
+++ b/monitoring/datadog-dashboard.json
@@ -202,6 +202,78 @@
         "width": 5,
         "height": 3
       }
+    },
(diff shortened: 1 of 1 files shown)


monitoring/datadog-dashboard.json
```diff
@@ -202,6 +202,78 @@
         "width": 5,
         "height": 3
       }
+    },
+    {
+      "id": 5713842188321988,
+      "definition": {
+        "title": "Ingestion/Refresh Lag",
```
lukekim: Confusing. Which is it? Also, let's add a description.

monitoring/datadog-dashboard.json
```diff
@@ -202,6 +202,78 @@
...
+        "legend_layout": "auto",
+        "legend_columns": [
+          "avg",
+          "min",
+          "max",
+          "value",
+          "sum"
+        ],
+        "time": {},
+        "type": "timeseries",
```
lukekim, line 222: Series needs better names:

image

monitoring/datadog-dashboard.json
```diff
@@ -202,6 +202,78 @@
...
+        "requests": [
+          {
+            "response_format": "timeseries",
+            "queries": [
+              {
+                "name": "query1",
```
lukekim, line 228: Shall we name the queries?

monitoring/datadog-dashboard.json
krinart: Maybe we can do "Ingestion + Refresh Lag", since my understanding was we wanted to wanted to display both on the same graph?

Or we can have two separate graphs.

monitoring/datadog-dashboard.json
krinart: Done

monitoring/datadog-dashboard.json
krinart: Done

monitoring/datadog-dashboard.json
krinart: Or did you mean values used in js? For other dashboards we use the same `query1` name, should I update all of them?

PR #7355, opened 2025-09-30
Add support for writing into Glue (Iceberg) tables and catalogs

sgrebnov:
## 📝 Summary

PR enables writes (`INSERT INTO`) into Glue Iceberg tables and adds integration tests for writes: Iceberg Catalog, Glue Table, Glue Catalog.

Also moves `AccessMode` from Datasets to be a common module for Catalogs and Datasets.

## 🔗 Related

Implements 
 - https://github.com/spiceai/spiceai/issues/7322


crates/runtime/src/dataconnector/glue.rs
```diff
@@ -108,72 +108,8 @@ impl GlueDataConnector {
     pub fn new(params: Parameters) -> Self {
         Self { params }
     }
-}
-
-impl GlueDataConnector {
```
sgrebnov, line 113: There are no changes here - I've just moved `read_provider` code into `create_table_provider` so it can be shared between read and read_write providers

phillipleblanc (approved):

PR #7373, opened 2025-09-30
Move S3Vector to 'search' crate

Jeadie:
## 📝 Summary
- Move many S3vector code out of `runtime` into `search` crate.


crates/llms/src/embeddings/mod.rs
```diff
@@ -142,6 +142,25 @@ fn encode_embedding(format: &EncodingFormat, array: Vec<f32>) -> EmbeddingVector
 pub trait Embed: Debug + Sync + Send {
     async fn embed(&self, input: EmbeddingInput) -> Result<Vec<Vec<f32>>>;
 
+    /// Implementation so that upstream do not need [`async_openai`] types.
+    async fn embed_str(&self, input: &str) -> Result<Vec<f32>> {
```
phillipleblanc: Why not just define our own EmbeddingInput enum and keep the same interface but remove our dependency on async_openai completely?

phillipleblanc (dismissed):

crates/llms/src/embeddings/mod.rs
Jeadie: fair

Jeadie: Tried implementing this. Gets tricky because of two compatibilties:
1. Cache keys (cache crate), and the fact we also cache on `CreateEmbeddingRequest`.
2. Providing to `CreateEmbeddingRequest`.

Jeadie: Actually, I found a better method. I just reexport from `llm` crate.

sgrebnov (dismissed):

peasee (dismissed):

phillipleblanc (approved):

PR #7378, opened 2025-09-30
Make table allowlist in SQL tool

Jeadie:


crates/runtime/src/datafusion/query/builder.rs
```diff
@@ -28,6 +29,7 @@ pub struct QueryBuilder<'a> {
     df: Arc<DataFusion>,
     sql: &'a str,
     parameters: Option<ParamValues>,
+    table_allowlist: Option<GlobSet>,
```
lukekim, line 32: We'll need a better name for this I think.

lukekim: Maybe `allowed_datasets` or `target_datasets` (I think I prefer `target_datasets`)

PR #7382, opened 2025-09-30
Fix embed logs

Jeadie:
## 📝 Summary
- #7333 did not resolve all issues since `TaskEmbed::embedding_input_cache_key` uses the inner model's `Embed::model_name`. 
- This PR adds names for local and hugging models specifically


crates/llms/src/embeddings/candle/tei.rs
```diff
@@ -54,6 +54,7 @@ pub struct TeiEmbed {
 
     // Shared embeddings cache
     cache: Option<Arc<dyn CacheProvider<CachedEmbeddingResult> + Send + Sync>>,
+    cache_model_id: Option<String>, // Used for unique key in `cache`.
```
phillipleblanc, line 57: could just use `model_name` here

phillipleblanc (approved):

PR #7383, opened 2025-10-01
[bug] Fix ai UDF bug of mismatched column length

lukekim:
This pull request updates the AI user-defined function (UDF) implementation in `ai.rs` to fix handling of scalar and array arguments, ensuring correct behavior when mixing columnar and scalar values as function arguments. It also adds comprehensive tests to verify correct handling of these cases and prevent errors related to mismatched array lengths in record batches.

**Argument handling improvements:**

* Refactored the AI UDF to handle cases where the message argument is a column (array) and the model name is a scalar, ensuring only the message argument is expanded to match the number of rows, while the model name remains a scalar. This fixes errors like "all columns in a record batch must have the same length" when mixing array and scalar arguments.
* Removed the unused import of `take_function_args` to reflect the simplified argument handling logic.

**Testing enhancements:**

* Added new async tests to cover all combinations of scalar and columnar arguments, including:
  - Columnar message with scalar model name.
  - Scalar message and scalar model name expanded to multiple rows.
  - Mixed cases with arrays and scalars, including handling of nulls and expansion of scalars to match row counts.
  These tests ensure the UDF behaves correctly and robustly in varied query scenarios.


crates/runtime-datafusion-udfs/src/ai.rs
```diff
@@ -1005,6 +997,199 @@ mod tests {
...
+            result.is_ok(),
+            "invoke_async_with_args should handle columnar message + scalar model: {:?}",
+            result.err()
+        );
+
+        let response_array = result.unwrap();
+        let string_array = response_array
+            .as_any()
+            .downcast_ref::<arrow::array::StringArray>()
+            .unwrap();
```
phillipleblanc: lint: change to expect("some message")

phillipleblanc (dismissed):

phillipleblanc (approved):

Issue #7364, opened 2025-09-30
On Spice startup, wire up the SnapshotManager.download_snapshot for datasets with snapshots enabled


phillipleblanc:
Only if the acceleration doesn't already exist at the named file location.

Issue #7361, opened 2025-09-30
`runtime_acceleration::SnapshotManager` downloads latest snapshot


phillipleblanc:
The SnapshotManager will have the functionality to create the connection to the object store snapshot location and find the latest snapshot based on the hive-style partitioned date, and then download it to a named local file location.

It will:
- List the snapshots location to build up a list of the hive-style partitioned files and filter to the current dataset.
- Identify the latest snapshot based on timestamp.
- Open the local file path for reading (truncating if already exists)
- Start a Get request to that path, start reading the byte stream
- Run the bytes through a SHA256 hasher
- Pipe the bytes into the opened local file (never keeping the entire dataset in memory)
- After the write is finished, calculate the final SHA256 hash. Download the associated `.sha256` file and compare the checksums. If they fail, then proceed according to the failure behavior.

PR #7386, opened 2025-10-01
Add new `runtime_acceleration` crate; create `SnapshotManager`; implement `SnapshotManager::download_latest_snapshot`

phillipleblanc:
## 📝 Summary

Adds a new `runtime_acceleration` crate which houses the new `SnapshotManager` which will be responsible for downloading/creating new snapshots.

This PR does an initial implementation of the `download_latest_snapshot` function, but it will need to be improved to support hive-styled partitioned files and SHA256 checksum validation. I will handle this in a follow-on PR.

This is not yet wired up, so it won't work. That will also come in a follow-on PR (ref: #7364)

## 🔗 Related

- Partially implements #7361 - as noted above, still need to implement hive-style partitioned snapshot files and SHS256 checksum validation.


crates/runtime-acceleration/src/dataset_checkpoint.rs
```diff
@@ -0,0 +1,41 @@
...
+#[async_trait]
+pub trait DatasetCheckpointer: Send + Sync {
+    async fn exists(&self) -> bool;
+    async fn checkpoint(&self, schema: &SchemaRef) -> Result<()>;
+    async fn get_schema(&self) -> Result<Option<SchemaRef>>;
+    async fn last_checkpoint_time(&self) -> Result<Option<SystemTime>>;
+}
+
+type CheckpointerFuture =
+    Pin<Box<dyn Future<Output = Result<Arc<dyn DatasetCheckpointer>>> + Send>>;
```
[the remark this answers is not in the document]
phillipleblanc: my linter disagrees

crates/runtime-acceleration/src/snapshot/mod.rs
```diff
@@ -0,0 +1,566 @@
...
+            Err(e) => {
+                tracing::error!(
+                    "Failed to parse snapshot location URL: {snapshot_location}, error: {e}"
+                );
+                return None;
+            }
+        };
+
+        let s3_region = snapshot_config
+            .as_ref()
```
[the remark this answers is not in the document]
phillipleblanc: no

crates/runtime-acceleration/src/snapshot/behavior.rs
```diff
@@ -0,0 +1,133 @@
...
+
+use std::sync::Arc;
+
+use spicepod::acceleration as spicepod_acceleration;
+use spicepod::component::snapshot::Snapshots;
+
+#[cfg(feature = "snapshots")]
+const SNAPSHOTS_ENABLED: bool = true;
+#[cfg(not(feature = "snapshots"))]
+const SNAPSHOTS_ENABLED: bool = false;
```
phillipleblanc, line 22: This is feature-gated and will be disabled by default for OSS builds.

crates/runtime-acceleration/src/snapshot/behavior.rs
```diff
@@ -0,0 +1,133 @@
...
+
+/// The behavior of snapshots for individual accelerated datasets.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub enum SnapshotBehavior {
+    /// Snapshots are disabled (default).
+    #[default]
+    Disabled,
+    /// Enable both creating and bootstrapping from snapshots.
+    Enabled(Arc<Snapshots>),
```
phillipleblanc, line 31: Putting the global configuration here makes it nicer for building the snapshot manager later. Instead of having to stitch together the global config with the dataset config, we only need the dataset config. This also centralizes any validation logic that I will do later.

crates/runtime-acceleration/src/snapshot/mod.rs
```diff
@@ -0,0 +1,566 @@
...
+}
+
+/// Manages snapshots for a specific accelerated dataset.
+pub struct SnapshotManager {
+    dataset_name: String,
+    snapshots_location: object_store::path::Path,
+    local_path: PathBuf,
+    object_store: Box<dyn ObjectStore>,
+    bootstrap_failure_behavior: BootstrapOnFailureBehavior,
+    checkpointer_factory: DatasetCheckpointerFactory,
```
phillipleblanc, line 84: We need a factory here because the acceleration files doesn't yet exist. We will use the factory to get a dataset checkpoint once the file is downloaded to validate we can read it. Also TBD is checksum validation to ensure the bytes are correct.

crates/runtime-acceleration/src/snapshot/mod.rs
```diff
@@ -0,0 +1,566 @@
...
+            Ok(url) => url,
+            Err(e) => {
+                tracing::error!(
+                    "Failed to parse snapshot location URL: {snapshot_location}, error: {e}"
+                );
+                return None;
+            }
+        };
+
+        let s3_region = snapshot_config
```
phillipleblanc, line 122: I'll handle more S3 auth parameters in a follow-on PR.

crates/runtime-acceleration/src/snapshot/behavior.rs
```diff
@@ -0,0 +1,133 @@
...
+
+        SnapshotBehavior::BootstrapOnly(snapshots)
+    }
+
+    #[must_use]
+    pub fn create_only(snapshots: Arc<Snapshots>) -> Self {
+        // Snapshot support must be compiled in for snapshot creation to be possible.
+        if !SNAPSHOTS_ENABLED {
+            tracing::trace!(
+                "Snapshot bootstrapping is not enabled because snapshot support is not compiled in."
```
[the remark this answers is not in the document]
phillipleblanc: I will take this change with my next update

peasee (approved):

PR #7387, opened 2025-10-01
Make `ai` UDF a models only feature

lukekim:
This pull request introduces a new optional "models" feature for the `runtime-datafusion-udfs` crate and integrates it with the main runtime. The main impact is that AI/model-related functionality is now conditionally compiled and only included if the "models" feature is enabled, improving modularity and reducing unnecessary dependencies when this functionality is not needed.

Feature flag integration:

* Added a new `[features]` section with a `models` feature in `runtime-datafusion-udfs/Cargo.toml` to allow conditional compilation of model-related code.
* Updated the `models` feature in `runtime/Cargo.toml` to depend on `runtime-datafusion-udfs/models`, ensuring the feature flag propagates correctly through the workspace.

Conditional compilation of AI functionality:

* Wrapped the `ai` module in `runtime-datafusion-udfs/src/lib.rs` with a `#[cfg(feature = "models")]` attribute so it is only included when the feature is enabled.
* Updated imports in `runtime/src/datafusion/udf.rs` to conditionally import the `ai` module based on the "models" feature.
* Modified the `register_udfs` function in `runtime/src/datafusion/udf.rs` to only register the AI UDF if the "models" feature is enabled.


phillipleblanc (approved):

crates/runtime-datafusion-udfs/Cargo.toml
```diff
@@ -27,3 +27,6 @@ once_cell = "1.21.3"
 async-stream.workspace = true
 insta.workspace = true
 tracing-test = "0.2"
+
+[features]
+models = []
```
lukekim, line 32: I am not sure if this makes all UDFs a models only?

phillipleblanc: No, it just means that is a feature that can be enabled - it does nothing on its own.

PR #7394, opened 2025-10-01
Implement cache invalidation for DML (INSERT INTO) operations

sgrebnov:
## 📝 Summary

PR adds automatic cache invalidation for write operations (INSERT INTO) in DataFusion using custom optimizer rule `cache_invalidation_optimizer_rule`:
 - `CacheInvalidationOptimizerRule`: DataFusion optimizer rule that detects DML operations and injects cache invalidation logic (`CacheInvalidationNode`)
 - `CacheInvalidationNode`: Logical plan node wrapping write operations with table metadata to create `CacheInvalidationExec` during execution.
 - `CacheInvalidationExec`: Physical execution plan that invalidates cache after successful write completion
 - `CacheInvalidationExtensionPlanner`: Converts logical cache invalidation nodes into physical plans

Usage:

When caching is enabled, write operations automatically invalidate relevant cached query results. This ensures cache consistency without requiring manual cache management or application-level invalidation logic.

```sql
-- Cache is populated
SELECT * FROM table; -- Result cached

-- Write operation automatically invalidates cache
INSERT INTO table VALUES (1, 'test', 42.0);

-- Next query rebuilds cache
SELECT * FROM table; -- Cache miss due to invalidation, fresh data is retrieved
```

## 🔗 Related

Closes 
 - https://github.com/spiceai/spiceai/issues/7323


Cargo.toml
```diff
@@ -246,12 +246,11 @@ rusqlite = { git = "https://github.com/spiceai/rusqlite.git", rev = "97054b6af72
 # Tracking Issue: https://github.com/allan2/dotenvy/issues/113
 dotenvy = { git = "https://github.com/spiceai/dotenvy.git", rev = "e5cef1871b08003198949dfe2da988633eaad78f" }
 
-
-iceberg = { git = "https://github.com/spiceai/iceberg-rust.git", rev = "15e2fc97f27580b753c10c21e3ab5fa02d0bc4cc" }              # spiceai-0.7.0.rc1
-iceberg-catalog-glue = { git = "https://github.com/spiceai/iceberg-rust.git", rev = "15e2fc97f27580b753c10c21e3ab5fa02d0bc4cc" } # spiceai-0.7.0.rc1
-iceberg-catalog-rest = { git = "https://github.com/spiceai/iceberg-rust.git", rev = "15e2fc97f27580b753c10c21e3ab5fa02d0bc4cc" } # spiceai-0.7.0.rc1
-iceberg-datafusion = { git = "https://github.com/spiceai/iceberg-rust.git", rev = "15e2fc97f27580b753c10c21e3ab5fa02d0bc4cc" }   # spiceai-0.7.0.rc1
-iceberg_test_utils = { git = "https://github.com/spiceai/iceberg-rust.git", rev = "15e2fc97f27580b753c10c21e3ab5fa02d0bc4cc" }   # spiceai-0.7.0.rc1
+iceberg = { git = "https://github.com/spiceai/iceberg-rust.git", rev = "cdb23210224c2a3b2a8f0c82635d393df86ada6d" }              # spiceai-0.7.0.rc1
```
sgrebnov: This brings fix for iceberg-rust:
https://github.com/spiceai/iceberg-rust/pull/22

crates/runtime/src/datafusion/extension/cache_invalidation.rs
```diff
@@ -0,0 +1,537 @@
...
+                    Ok(batch) => yield Ok(batch),
+                    Err(e) => {
+                        exec_failed = true;
+                        yield Err(e);
+                    }
+                }
+            }
+
+            // If wrapped input_exec completed without error, invalidate the cache
+            if !exec_failed {
```
sgrebnov: We might improve this further to check for 0 rows modifies / empty result and skip invalidation in this case, but propose to start from first robust implementation until we have more data/context/specific need to skip invalidation for zero updates.

crates/runtime/src/datafusion/extension/cache_invalidation.rs
```diff
@@ -0,0 +1,537 @@
+/*
+Copyright 2024-2025 The Spice.ai OSS Authors
```
phillipleblanc:
```suggestion
Copyright 2025 The Spice.ai OSS Authors
```

crates/runtime/src/datafusion/extension/cache_invalidation.rs
```diff
@@ -0,0 +1,537 @@
...
+    }
+}
+
+impl Debug for CacheInvalidationNode {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        UserDefinedLogicalNodeCore::fmt_for_explain(self, f)
+    }
+}
+
+impl UserDefinedLogicalNodeCore for CacheInvalidationNode {
```
phillipleblanc, line 151: double check that you've implemented all optional trait methods

crates/runtime/src/datafusion/extension/cache_invalidation.rs
```diff
@@ -0,0 +1,537 @@
...
+                write!(f, "CacheInvalidationExec: table={}", self.table)
+            }
+            DisplayFormatType::TreeRender => {
+                write!(f, "CacheInvalidationExec")
+            }
+        }
+    }
+}
+
+impl ExecutionPlan for CacheInvalidationExec {
```
phillipleblanc: Please double check that you've implemented all trat methods (where it makes sense to)

phillipleblanc: Instead of re-implementing the same wrapping exec node multiple times, it may be useful to just implement a PassthroughExec that takes a closure that computes some side effect we want.

sgrebnov: Thank you, confirm I've reviewed and added missing methods (where is makes sense to)

crates/runtime/src/datafusion/extension/cache_invalidation.rs
sgrebnov: Thank you, confirm I've reviewed and added missing methods (where is makes sense to)

crates/runtime/src/datafusion/extension/cache_invalidation.rs
sgrebnov: I like the idea and I gave this approach a try today, one concern with this approach I have is that we can't easily keep additional properties as part of the base exec plan and include as part of plan display, for example table_reference which makes sense in cache invalidation plan only, other plans might have some additional info/details to print.
```
 │-| physical_plan | CacheInvalidationExec: table=test_table                                           |
 │+| physical_plan | CacheInvalidationExec  
``` 

@phillipleblanc  - below is example implementation and usage to review and make a decision if we want this to be generic/easy to use but less flexible - please let me know your thoughts.

Usage

```rust
// Creates Physical [`ExecutionPlan`] that wraps a write operation and invalidates cache after successful completion.
fn cache_invalidation_exec(
    input: Arc<dyn ExecutionPlan>,
    table: TableReference,
    caching: Weak<Caching>,
) -> Arc<dyn ExecutionPlan> {
    let exec = move |input_exec: &Arc<dyn ExecutionPlan>, partition, ctx| {
        let schema = input_exec.schema();
        let input_stream = input_exec.execute(partition, ctx)?;
        let table = table.clone();
        let caching = caching.clone();

        let s = stream! {
            let mut input = input_stream;
            let mut ok = true;
            while let Some(item) = input.next().await {
                match item {
                    Ok(b) => yield Ok(b),
                    Err(e) => { ok = false; yield Err(e); }
                }
            }
            if ok {
                if let Some(cache) = caching.upgrade() {
                    let _ = cache.invalidate_for_table(table.clone());
                }
            }
        };
        Ok(Box::pin(RecordBatchStreamAdapter::new(schema, s)) as SendableRecordBatchStream)
    };

    Arc::new(
        PassThruExec::new(input, "CacheInvalidationExec", exec)
            .with_input_partitioning(datafusion::physical_plan::Distribution::SinglePartition)
    )
}
```

Example `PassThruExec` implementation:


```rust
pub struct PassThruExec<F>
where
    F: Fn(&Arc<dyn ExecutionPlan>, usize, Arc<TaskContext>)
        -> Result<SendableRecordBatchStream, DataFusionError>
        + Send
        + Sync
        + Clone
        + 'static,
{
    input_exec: Arc<dyn ExecutionPlan>,
    name: &'static str,
    exec: F,
    required_input_distribution: Distribution,
}

impl<F> PassThruExec<F>
where
    F: Fn(&Arc<dyn ExecutionPlan>, usize, Arc<TaskContext>)
        -> Result<SendableRecordBatchStream, DataFusionError>
        + Send
        + Sync
        + Clone
        + 'static,
{
    pub fn new(input_exec: Arc<dyn ExecutionPlan>, name: &'static str, exec: F) -> Self {
        Self {
            input_exec,
            name,
            exec,
            required_input_distribution: Distribution::UnspecifiedDistribution,
        }
    }

    /// Override input partitioning distribution.
    pub fn with_input_partitioning(mut self, dist: Distribution) -> Self {
        self.required_input_distribution = dist;
        self
    }
}

impl<F> DisplayAs for PassThruExec<F>
where
    F: Fn(&Arc<dyn ExecutionPlan>, usize, Arc<TaskContext>)
        -> Result<SendableRecordBatchStream, DataFusionError>
        + Send
        + Sync
        + Clone
        + 'static,
{
    fn fmt_as(&self, t: DisplayFormatType, f: &mut fmt::Formatter) -> fmt::Result {
        match t {
            DisplayFormatType::Default | DisplayFormatType::Verbose =>
                write!(f, "{}", self.name),
            DisplayFormatType::TreeRender => write!(f, "{}", self.name),
        }
    }
}

impl<F> std::fmt::Debug for PassThruExec<F>
where
    F: Fn(&Arc<dyn ExecutionPlan>, usize, Arc<TaskContext>)
        -> Result<SendableRecordBatchStream, DataFusionError>
        + Send
        + Sync
        + Clone
        + 'static,
{
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        f.debug_struct(self.name)
            .field("required_input_distribution", &self.required_input_distribution)
            .finish()
    }
}

impl<F> ExecutionPlan for PassThruExec<F>
where
    F: Fn(&Arc<dyn ExecutionPlan>, usize, Arc<TaskContext>)
        -> Result<SendableRecordBatchStream, DataFusionError>
        + Send
        + Sync
        + Clone
        + 'static,
{
    fn name(&self) -> &'static str { self.name }
    fn as_any(&self) -> &dyn Any { self }
    fn properties(&self) -> &PlanProperties { self.input_exec.properties() }
    fn children(&self) -> Vec<&Arc<dyn ExecutionPlan>> { vec![&self.input_exec] }
    fn with_new_children(
        self: Arc<Self>,
        children: Vec<Arc<dyn ExecutionPlan>>,
    ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
        if children.len() != 1 {
            return Err(DataFusionError::Internal(
                format!("{} requires exactly one child", self.name),
            ));
        }

        let input_exec = children.into_iter().next().unwrap();

        Ok(Arc::new(Self {
            input_exec,
            name: self.name,
            exec: self.exec.clone(),
            required_input_distribution: self.required_input_distribution.clone(),
        }))
    }

    fn required_input_distribution(&self) -> Vec<Distribution> {
        vec![self.required_input_distribution.clone()]
    }
    fn benefits_from_input_partitioning(&self) -> Vec<bool> { vec![false] }
    fn maintains_input_order(&self) -> Vec<bool> { vec![true] }
    fn required_input_ordering(&self) -> Vec<Option<OrderingRequirements>> { vec![None] }

    fn execute(
        &self,
        partition: usize,
        ctx: Arc<TaskContext>,
    ) -> Result<SendableRecordBatchStream, DataFusionError> {
        (self.exec)(&self.input_exec, partition, ctx)
    }
}
```

sgrebnov: Updated implementation based on re-usable `PassThruExec` implementation, will update `BytesProcessedExec` to use `PassThruExec` next.

crates/runtime/src/datafusion/extension/cache_invalidation.rs
```diff
@@ -0,0 +1,490 @@
...
+    /// This is used for projection push-down when `DataFusion` has determined that
+    /// only a subset of the output columns of this node are needed by its parents.
+    /// This API is used to tell `DataFusion` which, if any, of the input columns are no longer
+    /// needed.
+    ///
+    /// Return `None`, the default, if this information can not be determined.
+    /// Returns `Some(_)` with the column indices for each child of this node that are
+    /// needed to compute `output_columns`
+    fn necessary_children_exprs(&self, _output_columns: &[usize]) -> Option<Vec<Vec<usize>>> {
+        // Default implementation
```
phillipleblanc: Shouldn't we say that all of the input columns are required to compute the output columns?

crates/runtime/src/datafusion/extension/cache_invalidation.rs
```diff
@@ -0,0 +1,490 @@
...
+    /// Returns `true` if a limit can be safely pushed down through this
+    /// `UserDefinedLogicalNode` node.
+    ///
+    /// If this method returns `true`, and the query plan contains a limit at
+    /// the output of this node, `DataFusion` will push the limit to the input
+    /// of this node.
+    fn supports_limit_pushdown(&self) -> bool {
+        // Disallow limit push-down (default implementation).
+        // The node wraps write operations which produce a single output row.
+        false
```
phillipleblanc: We should allow this right?

crates/runtime/src/datafusion/extension/cache_invalidation.rs
```diff
@@ -0,0 +1,490 @@
...
+        Ok(Box::pin(RecordBatchStreamAdapter::new(schema, s)) as SendableRecordBatchStream)
+    };
+
+    let table_fmt_fn = table.clone();
+    let display_fmt_fn = move |t: DisplayFormatType, f: &mut fmt::Formatter| match t {
+        DisplayFormatType::Default | DisplayFormatType::Verbose => {
+            write!(f, "CacheInvalidationExec: table={table_fmt_fn}")
+        }
+        DisplayFormatType::TreeRender => {
+            write!(f, "CacheInvalidationExec")
```
phillipleblanc: maybe for the tree render we can put the table on the second line?

phillipleblanc (dismissed):

crates/runtime/src/datafusion/extension/cache_invalidation.rs
sgrebnov: Updated, thank you. My thinking was that  the node wraps write operations which produces a single output row, but  I agree enabling limit_pushdown is more correct 👍

crates/runtime/src/datafusion/extension/cache_invalidation.rs
sgrebnov: Updated, thank you. My thinking was that the node wraps write operations that returns hardcoded output schema (count) so does not require any columns to compute output schema

crates/runtime/src/datafusion/extension/cache_invalidation.rs
sgrebnov: Good point, I actually found that we should NOT print `CacheInvalidationExec ` as it will be duplicated, looking to example DF implementations we should print only additional properties for `TreeRender`, updated

https://github.com/apache/datafusion/blob/71d741504a9795343c722a0b9a4bfa214c3f6d26/datafusion/physical-plan/src/async_func.rs#L121

```rust
DisplayFormatType::TreeRender => {
                writeln!(f, "format=async_expr")?;
                writeln!(f, "async_expr={exprs}")?;
                Ok(())
            }
```

```
sql> explain format tree insert into task_history_sink select * from runtime.task_history limit 1;
+---------------+-------------------------------+
| plan_type     | plan                          |
+---------------+-------------------------------+
| physical_plan | ┌───────────────────────────┐ |
|               | │   CacheInvalidationExec   │ |
|               | │    --------------------   │ |
|               | │           table:          │ |
|               | │     task_history_sink     │ |
|               | └─────────────┬─────────────┘ |
|               | ┌─────────────┴─────────────┐ |
|               | │     IcebergCommitExec     │ |
|               | │    --------------------   │ |
|               | │ IcebergCommitExec: table: │ |
|               | │   team_app.task_history   │ |
|               | └─────────────┬─────────────┘ |
|               | ┌─────────────┴─────────────┐ |
|               | │   CoalescePartitionsExec  │ |
|               | └─────────────┬─────────────┘ |
|               | ┌─────────────┴─────────────┐ |
|               | │      IcebergWriteExec     │ |
|               | │    --------------------   │ |
|               | │  IcebergWriteExec: table: │ |
|               | │   team_app.task_history   │ |
|               | └─────────────┬─────────────┘ |
|               | ┌─────────────┴─────────────┐ |
|               | │       ProjectionExec      │ |
|               | │    --------------------   │ |
|               | │      captured_output:     │ |
|               | │      captured_output      │ |
|               | │                           │ |
|               | │         end_time:         │ |
|               | │ CAST(end_time AS Timestamp│ |
|               | │    (Microsecond, None))   │ |
|               | │                           │ |
|               | │       error_message:      │ |
|               | │       error_message       │ |
|               | │                           │ |
|               | │   execution_duration_ms:  │ |
|               | │   execution_duration_ms   │ |
|               | │                           │ |
|               | │        input: input       │ |
|               | │                           │ |
|               | │          labels:          │ |
|               | │ CAST(labels AS Map(Field {│ |
|               | │     name: "key_value",    │ |
|               | │     data_type: Struct(    │ |
|               | │   [Field { name: "key",   │ |
|               | │      data_type: Utf8,     │ |
|               | │      nullable: false,     │ |
|               | │         dict_id: 0,       │ |
|               | │       dict_is_ordered     │ |
|               | │    : false, metadata: {   │ |
|               | │    "PARQUET:field_id":    │ |
|               | │  "12"} }, Field { name:   │ |
|               | │    "value", data_type:    │ |
|               | │    Utf8, nullable: true   │ |
|               | │            ...            │ |
|               | └─────────────┬─────────────┘ |
|               | ┌─────────────┴─────────────┐ |
|               | │      RepartitionExec      │ |
|               | │    --------------------   │ |
|               | │ partition_count(in->out): │ |
|               | │          1 -> 14          │ |
|               | │                           │ |
|               | │    partitioning_scheme:   │ |
|               | │    RoundRobinBatch(14)    │ |
|               | └─────────────┬─────────────┘ |
|               | ┌─────────────┴─────────────┐ |
|               | │      GlobalLimitExec      │ |
|               | │    --------------------   │ |
|               | │          limit: 1         │ |
|               | │          skip: 0          │ |
|               | └─────────────┬─────────────┘ |
|               | ┌─────────────┴─────────────┐ |
|               | │     BytesProcessedExec    │ |
|               | │    --------------------   │ |
|               | │     BytesProcessedExec    │ |
|               | └─────────────┬─────────────┘ |
|               | ┌─────────────┴─────────────┐ |
|               | │     SchemaCastScanExec    │ |
|               | │    --------------------   │ |
|               | │     SchemaCastScanExec    │ |
|               | └─────────────┬─────────────┘ |
|               | ┌─────────────┴─────────────┐ |
|               | │       DataSourceExec      │ |
|               | │    --------------------   │ |
|               | │        bytes: 99072       │ |
|               | │       format: memory      │ |
|               | │          rows: 8          │ |
|               | └───────────────────────────┘ |
|               |                               |
+---------------+-------------------------------+

Time: 0.028555667 seconds. 1 rows.
```

sgrebnov: I'll update `SchemaCastScanExec `, `BytesProcessedExec `, etc

phillipleblanc (approved):

PR #7409, opened 2025-10-03
Move snapshot download to DataAccelerator::init()

phillipleblanc:
## 📝 Summary

Moves the snapshot download to happen in the DataAccelerator::init() function. This fixes a bug where the init() function would eagerly create the acceleration file before the previous snapshot download code was invoked.

I had considered putting it here initially, but decided against it because it means that resolving snapshots would prevent other datasets from initializing - but doing it here does simplify some later code and waiting for the snapshots to resolve first is probably desirable behavior anyway.

Part of #7361


ewgenius (dismissed):

phillipleblanc:
Superseded by https://github.com/spiceai/spiceai/pull/7410

PR #7410, opened 2025-10-03
Write new snapshots at the end of an accelerated refresh.

### LINKED ISSUE #7363: At the end of a full/append refresh, flush the WAL log and call SnapshotManager.create_new_snapshot

phillipleblanc:
## 📝 Summary

If snapshots are enabled for writing, at the end of an accelerated refresh Spice will upload the local acceleration as a new snapshot to the snapshot bucket. The next restart of Spice will download that snapshot.

## 🔗 Related

- Closes #7363


crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -690,28 +693,42 @@ impl RefreshTask {
     }
 
     fn refresh_df_context(&self, federated_provider: Arc<dyn TableProvider>) -> SessionContext {
-        let ctx = if self.disable_federation {
-            SessionContext::new_with_config_rt(get_df_default_config(), default_runtime_env())
+        let state_builder = SessionStateBuilder::new()
```
phillipleblanc, line 696: Drive by refactoring for when I thought I would implement the snapshot creating as part of the DF execution plan via optimizer rule. I like this refactoring though, so I decided to keep it.

crates/runtime/src/dataaccelerator/spice_sys/dataset_checkpoint/duckdb.rs
```diff
@@ -108,6 +108,13 @@ impl DatasetCheckpoint {
             .execute(&upsert, [&self.dataset_name, &schema_json])
             .map_err(Error::external)?;
 
+        if self.snapshot_behavior.create_enabled() {
```
phillipleblanc, line 111: We only do a DuckDB checkpoint when snapshots are enabled, because we've already verified there will be no other transactions because only the acceleration is writing to this file, and there is only one dataset.

crates/runtime-acceleration/src/snapshot/mod.rs
```diff
@@ -347,27 +378,37 @@ impl SnapshotManager {
     pub async fn download_latest_snapshot(
         &self,
     ) -> Result<Option<SchemaRef>, SnapshotDownloadError> {
+        let checkpointer_factory = Arc::clone(
+            self.checkpointer_factory
+                .as_ref()
+                .context(CheckpointerFactoryNotSetSnafu)?,
```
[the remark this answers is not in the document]
phillipleblanc: snafu errors are descriptive

crates/runtime/src/dataaccelerator/snapshots.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use crate::{
+    component::dataset::acceleration::Acceleration,
+    dataaccelerator::{
+        AccelerationSource,
+        spice_sys::{OpenOption, dataset_checkpoint::DatasetCheckpoint},
+    },
+};
+
+pub(super) async fn download_snapshot_if_needed(
```
phillipleblanc, line 32: Centralized the logic to download a snapshot to reuse across duckdb & sqlite

mach-kernel (approved):

PR #7412, opened 2025-10-04
S3 Vectors Table Provider

kczimm:


crates/s3_vectors/src/table.rs
```diff
@@ -0,0 +1,268 @@
...
+    }
+}
+
+fn contains_cosine_similarity_udf(expr: &Expr) -> Option<VectorData> {
+    if let Expr::ScalarFunction(ScalarFunction { func, args }) = expr {
+        if func.name() == "cosine_similarity" && args.len() == 2 {
+            if let Expr::Literal(ScalarValue::List(array), _) = &args[1] {
+                if let Some(list_array) = array.as_any().downcast_ref::<ListArray>() {
+                    let values = list_array.values();
+                    if let Some(float32_array) = values.as_any().downcast_ref::<Float32Array>() {
```
mach-kernel, line 150: pro-tip: Use `cast_to` with your desired target type to avoid coercions between e.g. f32 f64 and all the variants there

crates/s3_vectors/src/table.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    async fn scan(
+        &self,
+        _state: &dyn Session,
+        _projection: Option<&Vec<usize>>,
+        filters: &[Expr],
+        limit: Option<usize>,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        let mut source = None;
+        for filter in filters {
```
mach-kernel, line 88: Do we support multiple predicates? Either way is fine: clear error if we don't, or join if we do?

crates/s3_vectors/src/table.rs
```diff
@@ -0,0 +1,268 @@
...
+            f,
+            "{NAME} bucket={} index={}",
+            self.index.vector_bucket_name(),
+            self.index.index_name
+        )
+    }
+}
+
+#[async_trait]
+impl TableProvider for S3VectorsTable {
```
mach-kernel, line 67: Concise and clear, nice

mach-kernel (commented):
I know it's still a draft, but wanted to take a look. This will be a great improvement to our S3V pattern.

PR #7452, opened 2025-10-07
fix: partition name validation

kczimm:
## 📝 Summary

Fix validation of S3 vector index names when using partitions. Previously disallowed hyphens but need to disallow periods, as they are used for metadata delimiter.

### DIFF
--- a/crates/data_components/src/s3_vectors/partition.rs
+++ b/crates/data_components/src/s3_vectors/partition.rs
@@ -55,8 +55,8 @@ pub enum Error {
     IncorrectNumPartsInName { num_parts: usize },
     #[snafu(display("The 'partition_by' expression, {expr}, is not supported"))]
     UnsupportedPartitionByExpression { expr: Box<Expr> },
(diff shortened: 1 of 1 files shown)


crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -55,10 +55,10 @@ pub enum Error {
     IncorrectNumPartsInName { num_parts: usize },
     #[snafu(display("The 'partition_by' expression, {expr}, is not supported"))]
     UnsupportedPartitionByExpression { expr: Box<Expr> },
-    #[snafu(display("Index names cannot contain hyphens"))]
-    InvalidIndexNameHyphen,
+    #[snafu(display("Index names cannot contain periods when using 'partition_by'"))]
```
phillipleblanc: Let's also add the index name into the error message

lukekim (dismissed):

sgrebnov (approved):

Issue #7276, opened 2025-09-23
Return an error when trying to use FTS without acceleration


krinart:
FTS requires acceleration.

Currently when acceleration is not enabled, we simply return no results when doing FTS.

For improved UX we should return an error instead.

Example configuration which should result in an error:
```yaml
datasets:
  - from: duckdb:region
    name: region
    params:
      duckdb_open: tpch.db
    columns:
      - name: r_comment
        full_text_search:
          enabled: true
          row_id: r_regionkey
```

Additionally we should update docs to mention it explicitly (currently we just add acceleration to the example spicepod config)

PR #7458, opened 2025-10-07
Show error if FTS is misconfigured for datasets/views

krinart:
## 📝 Summary

If user tries to enable FTS without enabled acceleration, following error will be shown:
```bash
2025-10-07T16:30:10.960482Z ERROR runtime::init::dataset: Dataset region: acceleration is required for full text search. Ensure the dataset has an acceleration configuration, and try again. For details, visit: https://spiceai.org/docs/reference/spicepod/datasets#acceleration
```

If user tries to define a view with enabled FTS, following error will be displayed:
```
2025-10-07T18:29:32.246462Z ERROR runtime::init::view: Unable to load view foo: Full text search is not supported for views.
```

**Important**: In both cases dataset/view will not be registered

## 🔗 Related

https://github.com/spiceai/spiceai/issues/7276

## 🚨 Breaking Changes

Previously both cases would result in dataset/view successfully registered. With this change error will be displayed and they won’t be registered. 

## 📚 Docs

N/A

## 👀 Notes for Reviewers

* I've added methods `validate_dataset`/`validate_view` which are called at the beginning of `load_dataset`/`load_view` respectively, so that in the future we could extend them with more validation.
* I added `#[allow(clippy::result_large_err)]` because clippy was complaining about too big `runtime::Error`


crates/runtime/src/init/view.rs
```diff
@@ -41,7 +41,8 @@ impl Runtime {
             let runtime = Arc::clone(&self);
             let secrets = runtime.secrets();
             if let Err(e) = runtime.load_view(&view, secrets) {
-                tracing::error!("Unable to load view: {e}");
+                let view_name = &view.name;
+                tracing::error!("Unable to load view {view_name}: {e}");
```
krinart, line 45: I've added `view_name` to the error message

crates/runtime/src/component/view.rs
```diff
@@ -141,6 +141,13 @@ impl View {
 
         false
     }
+
+    #[must_use]
+    pub fn has_full_text_column(&self) -> bool {
```
sgrebnov, line 146:
```suggestion
    pub fn has_full_text_search_column(&self) -> bool {
```

sgrebnov: or  `has_full_text_search`/`is_full_text_search_enabled`

crates/runtime/tests/models/ai_udf.rs
```diff
@@ -255,7 +255,7 @@ async fn test_ai_udf_left_truncate() -> Result<(), anyhow::Error> {
             let provider_names = ["OpenAI (gpt-4o-mini)", "xAI (grok-4)", "Anthropic (claude-haiku)"];
             for (idx, result) in results.iter().enumerate() {
                 assert!(
-                    !result.is_empty() && result.len() <= 25,
+                    !result.is_empty() && result.chars().count() <= 25,
```
sgrebnov, line 260: Is there a specific case/example why we change this?

sgrebnov (approved):

crates/runtime/src/component/view.rs
krinart: I agree, but this is to keep it consistent with `Dataset` https://github.com/spiceai/spiceai/blob/21498b7ae72210bd59dd0bd53b4395450133be6d/crates/runtime/src/component/dataset/mod.rs#L613

crates/runtime/tests/models/ai_udf.rs
krinart: Yes, it was failing with this error:
```
OpenAI (gpt-4o-mini) result should be non-empty and <= 25 chars, got: 'I don’t have access to ex'
```

As you can see there's 25 characters and yet it failed. Which I assume is because `result.len()` was returning number of bytes

sgrebnov (dismissed):

.github/workflows/integration_models.yml
```diff
@@ -315,6 +315,34 @@ jobs:
         run: |
           INSTA_WORKSPACE_ROOT="${PWD}" CARGO_MANIFEST_DIR="${PWD}" ./integration_test/ai_integration_test --nocapture local_embeddings_beta_requirements
 
+      - name: Push snapshots to branch
```
Jeadie: You've accidentally dragged in https://github.com/spiceai/spiceai/pull/7458/commits/220d33858b5bddff9c22d99678cf6efdbf227864

sgrebnov (approved):

PR #7493, opened 2025-10-09
Improve DuckDB partition naming

### LINKED ISSUE #7188: Improve DuckDB file partition naming
~Follow format specified in #6958~

Follow hive-style partition names

kczimm:
## 📝 Summary

- Refactor the S3 vector partition naming into a generic partition naming module.
- Specialize for S3 vectors from the generic.
- Add specialization for DuckDB.

## 🔗 Related

Closes:
- #7188


crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -14,69 +14,36 @@ See the License for the specific language governing permissions and
...
-    prelude::Expr,
-    scalar::ScalarValue,
+use datafusion::{prelude::Expr, scalar::ScalarValue};
+use runtime_table_partition::naming::{
+    BelongsWith, PartitionNameConfig, PartitionedName, Result, hash_to_hex, truncate,
 };
-use snafu::prelude::*;
-use twox_hash::XxHash64;
-
-const HASH_SEED: u64 = 7;
```
kczimm, line 25: Most of the logic in this file got moved to `naming`.

crates/data_components/src/s3_vectors/partition.rs
```diff
@@ -14,69 +14,36 @@ See the License for the specific language governing permissions and
...
+    inner: PartitionedName,
 }
 
-#[derive(Debug, PartialEq, Eq, Clone, Copy)]
-pub enum BelongsWith {
-    ThisDataset,
-    DifferentDataset,
-    DifferentColumn,
-    DifferentPartitionByExpressions,
+fn make_prefix(index_name: &str, column_name: &str) -> String {
```
kczimm, line 42: S3 vectors is unique because we also care about the column.

crates/runtime-table-partition/src/snapshots/runtime_table_partition__naming__tests__partition_by_stability-2.snap
```diff
@@ -0,0 +1,5 @@
+---
```
kczimm, line 1: The snapshots are just moved

kczimm:
I'm currently adapting this for DuckDB. We didn't use the `PartitionedTableProvider` or `PartitionCreator` for S3 vectors. The `PartitionCreator::infer_partitions` is meant to be able to extract the `partition_value` from the partition but the S3 vector index naming hashed it so it cannot be recovered. I'm wondering how best to deal with this. I honestly forgot about this need, again, because we weren't using `PartitionCreator` and because we couldn't fit `ScalarValues` in within the S3 index naming limitations. Perhaps the checkpoint.db can be used here to store a catalog of partitions like we have previously discussed. This would allow using the new naming conventions and extracting the partition values from the database instead of the partition filename since it cannot be recovered.

Issue #7500, opened 2025-10-09
Add client timeout for S3vectors


Jeadie:
Equivalent to, and for similar reasons to, the`client_timeout` in S3 data connector [params](https://spiceai.org/docs/components/data-connectors/s3#params), add a client_timeout param to [s3vectors](https://spiceai.org/docs/components/vectors/s3_vectors)

PR #7501, opened 2025-10-10
Add 'client_timeout' for s3 vector

Jeadie:
## 📝 Summary
 - Add `client_timeout` to s3vector.

## 🔗 Related
#7500


phillipleblanc (dismissed):

lukekim (dismissed):

phillipleblanc (dismissed):

crates/runtime/src/embeddings/index/s3/retry_client.rs
```diff
@@ -91,18 +92,7 @@ impl S3Vectors for S3VectorRetryClient {
...
-                .set_vector_bucket_name(input.vector_bucket_name.clone())
-                .set_index_name(input.index_name.clone())
-                .set_data_type(input.data_type.clone())
-                .set_dimension(input.dimension)
-                .set_distance_metric(input.distance_metric.clone())
-                .set_metadata_configuration(input.metadata_configuration.clone())
-                .send()
-                .await
-            {
+            match self.client.create_index(input.clone()).await {
```
phillipleblanc, line 95: Why do we need to clone the input? We aren't referencing it after we pass it into the function.

Jeadie: IIRC because the this is wrapped by a retry strategy, so `input` is moved otherwise. Will double check

Jeadie: Confirmed.

peasee (approved):

PR #7548, opened 2025-10-14
FTS index with nonfilterable metadata; search field not metadata by default.

Jeadie:
## 📝 Summary
- Full text search indexes will store non-filterable metadata. This includes the search field which is now (i.e. breaking change), not stored by default. 

#### Example 
```yaml
datasets:
  - from: file:/Users/jeadie/Github/spiceai/data/quora_duplicate_questions_1000.tsv
    name: questions
    acceleration:
      enabled: true
      engine: duckdb
    columns:
      - name: is_duplicate
        metadata:
          vectors: non-filterable
      - name: question1
        full_text_search:
          enabled: true
          row_id: id
      - name: question2
        full_text_search:
          enabled: true
        metadata:
          vectors: non-filterable
```
 - `question1` will not be stored in FTS index, `question2` will be.

## 🚨 Breaking Changes
 - Change in defaults. The content of the search field will not be stored in the index. Must be added as non-filterable metadata. 

## 👀 Notes for Reviewers
 - This PR reuses the metadata semantics from vector search indexes (i.e. `.datasets[].columns[].metadata.vectors`.


crates/search/src/index/mod.rs
```diff
@@ -41,10 +41,6 @@ pub trait SearchIndex: Index + std::fmt::Debug + Send + Sync + 'static {
     /// All [`Field`]s that define a primary key between the underlying table and the [`SearchIndex`].
     fn primary_fields(&self) -> Vec<Field>;
 
-    /// The additional columns available in the [`SearchIndex`].
-    /// For FTS indexes, this may return empty metadata columns.
-    fn metadata_columns(&self) -> &MetadataColumns;
```
Jeadie, line 46: Since we refactored how we create `SearchQueryProvider` this was no longer needed.

crates/search/src/generation/text_search/mod.rs
Jeadie, line 285: The reason we can delete alot from this is that `CandidateGeneration::search` is nearly an out dated interface. At least in FTS we were always asking for `search_field` content. Now if we don't have it, DataFusion and `SearchQueryProvider` will do all the necessary base table joins (this is also true for HTTP endpoint since it uses UDTF now too).

lukekim (dismissed):

ewgenius (approved):

PR #7559, opened 2025-10-15
Return `TableProvider` from `CandidateGeneration::search`. 

Jeadie:
## 📝 Summary
 - Significant change to `CandidateGeneration::search` interface. 
     ```diff
     - async fn search(
     -   &self,
     -   query: String,
     -   opt_filters: &[&Expr],
     -   addition_projection: &[&Expr],
     -   limit: usize,
    - ) -> Result<SendableRecordBatchStream>;
    + fn search(&self, query: String) -> DataFusionResult<Arc<dyn TableProvider>>;
    ```
 - The `SearchPipeline<A>` now handles projection, filtering, etc atop these `TableProvider` (instead, currently, where it handles `RecordBatch` streams).  
 - The impact of this is that datasets with non-indexed, chunked embedding columns, can now use the `vector_search` UDTF (Can simply use the `TableProvider` from `ChunkedNonIndexVectorGeneration::search`.


crates/runtime/src/search/request.rs
```diff
@@ -299,84 +297,6 @@ impl SearchRequest {
             })
             .collect::<Result<Vec<sqlparser::ast::Ident>>>()
     }
-
-    pub fn validate_keyword_to_ilike(k: &str, target_column: &str) -> Result<Expr> {
```
Jeadie, line 303: Interestingly, this was never used and Clippy didn't realise. There is an equivalent function in the `search` crate.

crates/runtime/src/builder.rs
```diff
@@ -206,16 +204,6 @@ impl RuntimeBuilder {
 
         let df = Arc::new(df);
 
-        // UDFs that require a weak reference to the DataFusion instance defined here.
```
Jeadie, line 215: These were registered twice (also in `crates/runtime/src/datafusion/udf.rs`).

crates/runtime/src/datafusion/udf.rs
```diff
@@ -38,10 +39,16 @@ pub fn register_udfs(runtime: &crate::Runtime) {
         Arc::new(TextSearchTableFunc::new(Arc::downgrade(&runtime.df))),
     );
 
-    ctx.register_udf(VectorSearchTableFunc::new(Arc::downgrade(&runtime.df)).into());
+    let explicit_pks = parse_explicit_primary_keys(runtime.app()).await;
```
phillipleblanc, line 42: Why is this needed? This will make it harder if we want to dynamically update datasets or add new ones at runtime right? (We don't really handle that today, but we might need to)

crates/runtime/src/embeddings/udtf.rs
```diff
@@ -375,10 +386,34 @@ impl TableFunctionImpl for VectorSearchTableFunc {
...
-            )));
+            let state = df.ctx.state();
+            let Some(embed_udf) = state.scalar_functions().get(EMBED_UDF_NAME) else {
+                return Err(DataFusionError::Plan(format!(
+                    "'{VECTOR_SEARCH_UDTF_NAME}()' requires missing UDF: '{EMBED_UDF_NAME}'",
+                )));
+            };
+
+            // Unsafe: worse case is metric without dimensions.
+            let dimensions = unsafe { RequestContext::current_sync().to_dimensions() };
```
phillipleblanc, line 397: Did you verify we get metrics in the normal case?

crates/runtime/src/search/candidate/vector.rs
```diff
@@ -329,100 +225,93 @@ impl ChunkedNonIndexVectorGeneration {
...
+                        col("score"),
+                    ],
+                ]
+                .concat(),
+            )?;
 
-    /// Whether additional columns of the underlying source can also be retrieved during generation.
-    fn supports_columns(&self, _projection: &[&Expr]) -> Result<Vec<bool>, SearchGenerationError> {
-        Ok(vec![])
+        Ok(Arc::new(ViewTable::new(plan.build()?, None)))
```
phillipleblanc, line 305: Would it make sense to just return the logical plan/dataframe here?

crates/search/src/pipeline.rs
```diff
@@ -49,20 +59,34 @@ impl Error {
     }
 }
 
+/// [`QueryEngine`] defines the minimal interface needed to execute [`LogicalPlan`].
+///
+/// This allows extensibility beyond [`SessionContext::execute_logical_plan`] then [`DataFrame::execute_stream`].
+#[async_trait::async_trait]
+pub trait QueryEngine: Send + Sync {
+    async fn run(&self, plan: LogicalPlan) -> Result<SendableRecordBatchStream, DataFusionError>;
```
phillipleblanc, line 67: What is the motivation for this?

phillipleblanc (dismissed):

crates/runtime/src/datafusion/udf.rs
Jeadie: Same semantic as HTTP v1/search. This is the current mechanism in which primary keys defined in the `spicepod.yaml` get provided to search operations.

crates/runtime/src/embeddings/udtf.rs
Jeadie: Yep

crates/runtime/src/search/candidate/vector.rs
Jeadie: There are four implementations, three of which are naturally `TableProviders`. This is the odd one out

crates/search/src/pipeline.rs
Jeadie: This is `search` crate, which cannot have dependence on `runtime` crate (which has `DataFusion` struct).

phillipleblanc (approved):

PR #7567, opened 2025-10-16
Use 'location' as primary key for document tables

Jeadie:
## 📝 Summary
 - All document tables have the following schema  (plus possible metadata columns), but `location` was not being registered as a primary key.
 https://github.com/spiceai/spiceai/blob/8f6982a30d494027fd9b635f6febf189e915efb8/crates/data_components/src/object/text.rs#L83-L89
- This change means that the `row_id` spicepod.yaml field for full text search or embeddings is now unnecessary for document tables.
- This PR also has a bug about how FTS indexes register primary key from table constraints.

### DIFF
--- a/crates/search/src/generation/text_search/index.rs
+++ b/crates/search/src/generation/text_search/index.rs
@@ -134,9 +134,9 @@ impl FullTextDatabaseIndex {
     ) -> Result<Vec<String>, super::Error> {
         // Use 'primary_key_override', fallback to underlying in table.
         let pks = match (primary_key_override, get_primary_keys(inner).await) {
-            (Some(pks), _) if !pks.is_empty() => pks,
-            (_, Ok(pks)) if !pks.is_empty() => pks,
-            (None, Err(e)) => {
+            // LHS takes precedence.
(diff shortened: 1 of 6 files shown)


lukekim (dismissed):

crates/search/src/generation/text_search/index.rs
```diff
@@ -134,17 +134,12 @@ impl FullTextDatabaseIndex {
     ) -> Result<Vec<String>, super::Error> {
         // Use 'primary_key_override', fallback to underlying in table.
         let pks = match (primary_key_override, get_primary_keys(inner).await) {
-            (Some(pks), _) => pks,
-            (None, Ok(pks)) => {
```
kczimm, line 138: What happened to this case? It looks like it's an error now?

kczimm (approved):

phillipleblanc (approved):

crates/search/src/generation/text_search/index.rs
Jeadie: `(None, Ok(pks))` is now split based on `pks.is_empty()`. Essentially we avoiding the internal conditional by putting it as a match conditional.

PR #7571, opened 2025-10-17
Extend query-related metrics

krinart:
## 📝 Summary

Adds following metrics:
* `query_returned_rows` - Histogram (we already calculated it, just created metric and emission)
* `query_produced_spills` - Counter (extracted from `physical_plan`)
* `query_spilled_bytes` - Counter (extracted from `physical_plan`)
* `query_spilled_rows` - Counter (extracted from `physical_plan`)

## 🔗 Related

https://github.com/spiceai/spiceai/issues/7578

## 🚨 Breaking Changes

N/A

## 📚 Docs

https://github.com/spiceai/docs/pull/1192 (`query_active_count` will be added in a different PR)

## 👀 Notes for Reviewers

N/A


crates/runtime/src/datafusion/query.rs
```diff
@@ -211,11 +212,15 @@ impl Query {
...
+            let physical_plan = session.create_physical_plan(&plan).await.unwrap();
 
-            let df_schema: SchemaRef = Arc::clone(df.schema().inner());
+            let stream = async {
+                let physical_plan = session.create_physical_plan(&plan).await?;
+                execute_stream(Arc::clone(&physical_plan), task_ctx)
+            };
 
-            let res_stream: SendableRecordBatchStream = match df.execute_stream().await {
+            let res_stream: SendableRecordBatchStream = match stream.await {
```
krinart: This work is needed to extract `physical_plan`. And once we have it, we don't really need `DataFrame` - so it was completely removed.

crates/runtime/src/datafusion/query.rs
```diff
@@ -211,11 +212,24 @@ impl Query {
...
+                    let error_code = ErrorCode::from(&e);
+                    handle_error!(
+                        tracker,
+                        &request_context,
+                        error_code,
+                        e,
+                        UnableToExecuteQuery
+                    )
+                }
+            };
```
krinart, line 228: This work is needed to extract `physical_plan`. And once we have it, we don't really need `DataFrame` - so it was completely removed.

crates/runtime/src/datafusion/query.rs
```diff
@@ -454,6 +476,54 @@ fn attach_query_tracker_to_stream(
     ))
 }
 
+#[must_use]
+/// Attaches logic to a stream which emits metrics from a physical plan.
+fn attach_physical_plan_metrics_to_stream(
```
phillipleblanc, line 481: Can this use the PassThruExec that Sergei made?

crates/telemetry/Cargo.toml
```diff
@@ -21,6 +21,7 @@ snafu.workspace = true
 sha2 = "0.10.8"
 tokio.workspace = true
 tracing.workspace = true
+once_cell = "1.21.3"
```
phillipleblanc: Why is this required?

krinart: Thanks!

crates/runtime/src/datafusion/query.rs
krinart: Here's the best I could do:
```rust
let request_context_clone = Arc::clone(&request_context);
let exec = move |input_exec: &Arc<dyn ExecutionPlan>, partition, ctx| {
    let schema = input_exec.schema();
    let input_stream = input_exec.execute(partition, ctx)?;

    let final_stream = attach_physical_plan_metrics_to_stream(
        input_stream,
        input_exec,
        Arc::clone(&request_context_clone),
    );

    Ok(Box::pin(RecordBatchStreamAdapter::new(schema, final_stream)) as SendableRecordBatchStream)
};

let final_physical_plan = Arc::new(
    PassThruExec::new(physical_plan, "CacheInvalidationExec", exec)
        .with_input_partitioning(datafusion::physical_plan::Distribution::SinglePartition)
);
```

Unfortunately we still need to use `attach_physical_plan_metrics_to_stream` function.

phillipleblanc (approved):

PR #7572, opened 2025-10-17
`ObjectMeta` filter pushdown for `ObjectStoreTextTable`

Jeadie:
## 📝 Summary
 - Enable filter pushdown for all `ObjectMeta` filters in `ObjectStoreTextTable`
 - Important for `refresh_mode: append` since the refresh_sql will have filter `WHERE last_modified > 'some_time'`
 
 ### Example 
**Before**
 ```
 sql> explain select location, length(content) from docx where log(size) > 6 ;
 +---------------+-----------------------------------------------------------------------------------------------------------------+
| plan_type     | plan                                                                                                            |
+---------------+-----------------------------------------------------------------------------------------------------------------+
| logical_plan  | Projection: docx.location, character_length(docx.content)                                                       |
|               |   BytesProcessedNode                                                                                            |
|               |     TableScan: docx projection=[location, content], full_filters=[log(CAST(docx.size AS Float32)) > Float32(6)] |
+---------------+-----------------------------------------------------------------------------------------------------------------+
```

**Now**
```
sql> explain select location, length(content) from docx where log(size) > 6 ;
+---------------+--------------------------------------------------------------------------------------------------------------+
| plan_type     | plan                                                                                                         |
+---------------+--------------------------------------------------------------------------------------------------------------+
| logical_plan  | Projection: docx.location, character_length(docx.content)                                                    |
|               |   BytesProcessedNode                                                                                         |
|               |     Filter: log(CAST(docx.size AS Float32)) > Float32(6)                                                     |
|               |       TableScan: docx projection=[location, content, size]                                                   |
+---------------+-----------------------------------------------------------------------------------------------------------------+
```

### DIFF
--- a/crates/data_components/src/object/filter.rs
+++ b/crates/data_components/src/object/filter.rs
@@ -1,5 +1,5 @@
 /*
-Copyright 2024-2025 The Spice.ai OSS Authors
+Copyright 2025 The Spice.ai OSS Authors
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: crates/data_components/src/object/filter.rs [hunk 1]
```diff
@@ -0,0 +1,123 @@
+/*
+Copyright 2024-2025 The Spice.ai OSS Authors
```
kczimm:
```suggestion
Copyright 2025 The Spice.ai OSS Authors
```

crates/data_components/src/object/filter.rs
```diff
@@ -0,0 +1,123 @@
...
+        Field::new("location", DataType::Utf8, false),
+        Field::new(
+            "last_modified",
+            DataType::Timestamp(TimeUnit::Millisecond, Some("UTC".into())),
+            false,
+        ),
+        Field::new("size", DataType::UInt64, false),
+        Field::new("e_tag", DataType::Utf8, true),
+        Field::new("version", DataType::Utf8, true),
+    ]));
```
kczimm: This is created anew each invocation. Perhaps this could be a `LazyLock`?

crates/data_components/src/object/filter.rs
```diff
@@ -0,0 +1,123 @@
...
+        schema,
+        vec![
+            Arc::new(location_array) as ArrayRef,
+            Arc::new(last_modified_array) as ArrayRef,
+            Arc::new(size_array) as ArrayRef,
+            Arc::new(e_tag_array) as ArrayRef,
+            Arc::new(version_array) as ArrayRef,
+        ],
+    )
+    .expect("Failed to create RecordBatch")
```
kczimm: We should return the `Result`.

crates/data_components/src/object/filter.rs
```diff
@@ -0,0 +1,123 @@
...
+    while let Some(batch_result) = stream.next().await {
+        let rb = batch_result?;
+        if let Some(location_array) = rb.column(0).as_any().downcast_ref::<StringArray>() {
+            for loc_opt in location_array {
+                if let Some(loc) = loc_opt {
+                    valid_locations.insert(loc.to_string());
+                }
+            }
+        };
+    }
```
kczimm: I think using `SessionContext::create_physical_expr` and then evaluating the newly created `RecordBatch` directly with the `PhysicalExpr` using `PhysicalExpr::evaluate` will be more efficient by avoiding the overhead of creating a full execution plan and stream.

crates/data_components/src/object/filter.rs
```diff
@@ -0,0 +1,123 @@
...
+*/
+
+use arrow_array::RecordBatch;
+use datafusion::error::DataFusionError;
+use datafusion::prelude::{Expr, SessionContext};
+use futures::StreamExt;
+use object_store::ObjectMeta;
+
+/// Filters [`ObjectMeta`]s that satisfy all provided `filter`s.
+pub async fn filter_object_meta(
```
kczimm: Some unit tests of this function will be helpful.

crates/data_components/src/object/text.rs
```diff
@@ -224,16 +227,24 @@ impl TableProvider for ObjectStoreTextTable {
...
-        ])
+        Ok(filters
+            .iter()
+            .map(|f| {
+                if f.column_refs()
+                    .contains(&Column::from_qualified_name("content"))
+                {
+                    TableProviderFilterPushDown::Unsupported
+                } else {
+                    TableProviderFilterPushDown::Exact
```
kczimm: Just for my own understanding, are we sure that the filtering we push down is Exact in this case and that additional filtering by DF would be unnecessary?

crates/data_components/src/object/text.rs
```diff
@@ -330,40 +342,43 @@ impl ObjectStoreTextExec {
 pub(crate) fn to_sendable_stream(
     ctx: ObjectStoreContext,
     formatter: Option<Arc<dyn DocumentParser>>,
+    filters: Vec<Expr>,
     limit: Option<usize>,
     schema: SchemaRef,
 ) -> impl Stream<Item = DataFusionResult<RecordBatch>> + 'static {
     stream! {
-        let mut object_stream = ctx.store.list(ctx.prefix.clone().map(Path::from).as_ref());
+        let mut object_stream = ctx.store.list(ctx.prefix.clone().map(Path::from).as_ref()).chunks(128);
```
kczimm: How is 128 determined?

crates/data_components/src/object/text.rs
Jeadie: Good point, technically there are some niche ones that could be invalid (since we use `SessionContext::default`).

Will change to inexact.

Jeadie: For example
```sql
SELECT 
  location, content
FROM docx
WHERE array_length(embed(location, 'openai_embeddings')) > 1 ;
```

phillipleblanc (approved):

Issue #7089, opened 2025-09-09
Enhancement: Use separate Tokio runtime for runtime HTTP server


phillipleblanc:
## Goal-State/What/Result

The runtime HTTP server uses a separate Tokio runtime instance that isn't shared with the main query processing Tokio runtime.

## Why/Purpose

DataFusion, by default, plans and executes all operations (both CPU and IO) on the same thread pool. This can cause issues with tail latencies that cause Kubernetes to kill pods due to health checks not returning. By running on a separate runtime instance, HTTP request latencies will stay fast and won't lead to large query processing causing Kubernetes to kill Spice because the HTTP `/health` endpoint isn't responding.

## By When

**Issue/Spec written and reviewed:** _Target Date_
**Done-Done:** _Target Date_

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

Ref: https://github.com/apache/datafusion/blob/main/datafusion-examples/examples/thread_pools.rs#L18
Ref: https://thenewstack.io/using-rustlangs-async-tokio-runtime-for-cpu-bound-tasks/
Ref: https://docs.rs/datafusion/latest/datafusion/index.html#thread-scheduling-cpu--io-thread-pools-and-tokio-runtimes

This issue is related to, but distinct from, https://github.com/spiceai/spiceai/issues/3100

## Release Notes

**Dedicated Query Thread Pool**: Query execution and accelerated refreshes can now run on their own dedicated thread pool, separate from the HTTP server. This prevents heavy query workloads from slowing down API responses, keeping health checks fast and avoiding unnecessary Kubernetes pod restarts under load.

This feature is opt-in for this release and will become enabled by default (opt-out) in the v1.9 release. To enable it now, add the following spicepod.yaml configuration:

```yaml
runtime:
  params:
    dedicated_thread_pool: sql_engine
```

[...]

PR #7586, opened 2025-10-18
Run Datafusion queries on a separate Tokio runtime

phillipleblanc:
## 📝 Summary

All Spice DataFusion queries will now run on a separate Tokio runtime.

I want to merge this in as the default for now so that we can get some testing on it, and I will create a PR that can disable spawning the queries on the separate runtime before the next release so that if it does end up breaking, there is a way to disable it.

## 🔗 Related

- Part of #7089


lukekim (dismissed):

crates/runtime/src/datafusion/query.rs
```diff
@@ -116,9 +118,68 @@ impl Query {
...
+                }
+            }
+        };
+
+        let driver_handle = runtime_handle.spawn(driver_task);
+        tokio::spawn(async move {
+            if let Err(err) = driver_handle.await {
+                tracing::error!("Query driver task failed: {err}");
+            }
+        });
```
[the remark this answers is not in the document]
phillipleblanc: Fixed

sgrebnov (dismissed):

crates/runtime/src/datafusion/query.rs
```diff
@@ -116,9 +124,63 @@ impl Query {
...
+                }
+                Err(err) => {
+                    if result_tx.send(Err(err)).is_err() {
+                        tracing::debug!("Failed to send query result; receiver dropped");
+                    }
+                }
+            }
+        };
+
+        let driver_handle = runtime_handle.spawn(driver_task);
```
krinart: Will this not make us lose current `request_context` as it will run on a different thread pool? Shouldn't we wrap executing this task with current `request_context`'s scope?

crates/runtime/src/datafusion/query.rs
```diff
@@ -416,6 +478,46 @@ fn parameter_schema_for_plan(plan: &LogicalPlan) -> Result<Option<Schema>, DataF
...
+    }
+}
+
+impl Stream for DriverStream {
+    type Item = Result<RecordBatch, DataFusionError>;
+
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let this = self.get_mut();
+        Pin::new(&mut this.receiver).poll_next(cx)
+    }
```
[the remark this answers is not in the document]
phillipleblanc: Fixed

crates/runtime/src/datafusion/query.rs
```diff
@@ -130,9 +139,64 @@ impl Query {
...
+                Ok(query_result) => {
+                    let schema = query_result.data.schema();
+                    let cache_status = query_result.cache_status;
+
+                    if result_tx.send(Ok((cache_status, schema))).is_err() {
+                        return;
+                    }
+
+                    let mut stream = query_result.data;
+                    while let Some(batch) = Arc::clone(&request_context).scope(stream.next()).await
```
krinart: My understanding is that while stream execution will run under `request_context`,  `driver_task`  itself will run under default `RequestContext`.

Maybe:
```rust
let driver_handle = runtime_handle.spawn(Arc::clone(&request_context).scope(driver_task));
```
?

lukekim (dismissed):

crates/runtime/src/datafusion/query.rs
phillipleblanc: The driver task doesn't need to be wrapped, because we're passing the request context in where its needed.

krinart (dismissed):

lukekim (dismissed):

lukekim (dismissed):

ewgenius (approved):

PR #7589, opened 2025-10-18
Emit `query_active_count` metric

### LINKED ISSUE #7579: Add `queries_active` metric

krinart:
## 📝 Summary

* Adds logic to emit `query_active_count`
  *  Uses OTEL `UpDownCounter`
  * Uses guard to make sure we always decrement and always with the same dimensions

## 🔗 Related

Closes https://github.com/spiceai/spiceai/issues/7579

## 📚 Docs

https://github.com/spiceai/docs/pull/1192


crates/runtime/src/flight/mod.rs
```diff
@@ -217,13 +218,15 @@ impl Service {
...
+        let request_context_clone = Arc::clone(&RequestContext::current(AsyncMarker::new().await));
+
         let data_stream = query_result.data;
         let flights_stream = try_stream! {
             yield schema_flight_data;
 
             futures::pin_mut!(data_stream); // needed to use `.next()` on stream
 
-            while let Some(batch_result) = data_stream.next().await {
+            while let Some(batch_result) = Arc::clone(&request_context_clone).scope(data_stream.next()).await {
```
krinart: `data_stream.next()` may be placed on a different tokio task - in this case `RequestContext` will be lost without this change

crates/runtime/src/datafusion/query.rs
```diff
@@ -476,12 +482,69 @@ fn attach_query_tracker_to_stream(
...
+    }
+}
+
+fn attach_query_active_guard_to_stream(
+    mut stream: SendableRecordBatchStream,
+    request_context: Arc<RequestContext>,
+    span: Span,
+) -> SendableRecordBatchStream {
+    let schema = stream.schema();
+    let updated_stream = stream! {
```
phillipleblanc: Rewrite this to not use `stream!`

crates/runtime-request-context/src/context.rs
```diff
@@ -44,6 +46,7 @@ pub struct RequestContext {
     auth_principal: OnceLock<AuthPrincipalRef>,
     extensions: RwLock<Extensions>,
     trace_parent: Option<TraceParent>,
+    nested_query_level: AtomicI16,
```
phillipleblanc, line 54: Why is this required? It doesn't seem to be used anywhere

crates/runtime-request-context/src/context.rs
```diff
@@ -171,6 +174,11 @@ impl RequestContext {
         dimensions
     }
 
+    #[must_use]
+    pub fn to_protocol_dimensions(&self) -> Vec<KeyValue> {
```
phillipleblanc: This will allocate on every query, there are a limited number of possible values here, can we statically allocate them on boot and just return static references to the appropriate dimensions?

crates/runtime-request-context/src/context.rs
krinart: I've added two methods to `RequestContext`:
```rust
    pub fn entered_top_level_query(&self) -> bool {
        self.nested_query_level.fetch_add(1, Ordering::Relaxed) == 0
    }

    pub fn exited_top_level_query(&self) -> bool {
        self.nested_query_level.fetch_add(-1, Ordering::Relaxed) == 1
    }
```

They are used by `QueryActiveGuard`:
* `entered_top_level_query` on creation
* `exited_top_level_query` on drop

phillipleblanc: Right, but are they actually consumed for anything? It seems like we could remove those functions without affecting anything.

phillipleblanc: Ah I see, its for controlling when the metric is emitted

phillipleblanc (approved):

PR #7591, opened 2025-10-19
Properly set RequestContext for stream execution in Flight

krinart:
## 📝 Summary

While we were setting `RequestContext` in flight, we were only setting it for building a stream, not for executing a stream.

Which resulted in a `INTERNAL_REQUEST_CONTEXT` context returned from `RequestContext::current()` while stream is actually being executed.

This in turn had other implications: 
* Inside the stream execution, `protocol` would always be `Internal`
* Our query related metrics would be visible as `Internal` when executed from flight


crates/runtime-request-context/src/context.rs
```diff
@@ -143,6 +144,19 @@ impl RequestContext {
         REQUEST_CONTEXT.scope(self, f).await
     }
 
+    /// Wraps provided stream with the current request context.
+    pub fn scope_stream<S>(self: Arc<Self>, stream: S) -> impl Stream<Item = S::Item>
+    where
+        S: Stream,
+    {
+        stream! {
```
phillipleblanc: Rework this to not use the `stream!` macro.

krinart: Any specific reason? 
Just trying to understand so that I come up with a proper solution.

phillipleblanc: It breaks the use of rust-analyzer hints in the IDE, and generally makes it harder to debug issues that occur inside of the macro.

krinart: Done!

phillipleblanc (dismissed):

sgrebnov (approved):

Issue #7574, opened 2025-10-17
Query and List operations for virtual S3 indexes


kczimm:
- Update query operations to search across spill indexes
- Update list operations to aggregate from spill indexes

PR #7613, opened 2025-10-20
S3 Vector index spilling

### LINKED ISSUE #7573: Virtual S3 vector index spilling logic
- Implement spill index naming/detection logic
- Update `S3VectorIdentifier` for virtual indexes
- Add capacity detection and retry logic in put operations

kczimm:
## 📝 Summary

This PR detectors a ServiceQuota error from the S3V service and creates a spill index and continues writing vectors to the latest spill index. It also modifies the List and Query providers to query across spill indexes. There are tests to verify functionality. 

## 🔗 Related

Closes:
- #7573 
- #7574


crates/s3_vectors/src/mock.rs
```diff
@@ -201,9 +292,43 @@ impl S3Vectors for MockClient {
...
+                .build()
+                .map_err(|_| SdkError::construction_failure("build"))?;
+
+            return Err(SdkError::ServiceError(
+                ServiceError::builder()
+                    .source(PutVectorsError::ServiceQuotaExceededException(
+                        service_error,
+                    ))
+                    .raw(HttpResponse::new(
+                        StatusCode::try_from(402)
```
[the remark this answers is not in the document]
kczimm: https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_PutVectors.html

crates/data_components/src/s3_vectors/list_provider.rs
```diff
@@ -77,6 +80,193 @@ impl S3VectorsListTable {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
+async fn create_spill_plan(
```
sgrebnov, line 85: Nit: we can have brief info about purpose/logic as part of functions description

sgrebnov (dismissed):

crates/data_components/src/s3_vectors/vector_table.rs
```diff
@@ -476,19 +514,123 @@ impl S3VectorsTable {
...
+                    match SpillIndex::parse(index_name) {
+                        Ok(Some(spill)) => (bucket_name.clone(), spill.base_name),
+                        _ => return Ok(None), // Invalid spill index format
+                    }
+                } else {
+                    // This is the main index
+                    (bucket_name.clone(), index_name.clone())
+                }
+            }
+            S3VectorIdentifier::IndexArn(_) => {
```
phillipleblanc: Maybe we should just remove support for IndexArn?

crates/data_components/src/s3_vectors/vector_table.rs
```diff
@@ -476,19 +514,123 @@ impl S3VectorsTable {
...
+            }
+            let spill_name = format!("{base_index_name}.{spill_num:02}");
+            match self.create_spill_index(&bucket_name, &spill_name).await {
+                Ok(()) => {
+                    return Ok(Some(S3VectorIdentifier::Index {
+                        bucket_name: bucket_name.clone(),
+                        index_name: spill_name,
+                    }));
+                }
+                Err(_) => {
```
phillipleblanc: Is the idea that if the spill index is already created we should skip it? It might be better to explicitly check for the Conflict in that case. Or if the index is already created, we could try to use it?

crates/data_components/src/s3_vectors/vector_table.rs
```diff
@@ -48,7 +52,7 @@ use tokio::sync::mpsc::Sender;
 /// An S3 Vector index.
 #[derive(Clone)]
 pub struct S3VectorsTable {
-    pub idx: S3VectorIdentifier,
+    pub idx: Arc<Mutex<S3VectorIdentifier>>,
```
phillipleblanc: Making this a RWLock might be better?

crates/data_components/src/s3_vectors/spill.rs
```diff
@@ -0,0 +1,274 @@
...
+
+        Ok(Some(Self {
+            base_name,
+            sequence,
+        }))
+    }
+
+    /// Checks if an index name represents a spill index.
+    #[must_use]
+    pub fn is_spill_index(index_name: &str) -> bool {
```
phillipleblanc: this potentially will be a false positive if the base name has a `.` right? is that an issue?

crates/data_components/src/s3_vectors/query_provider.rs
```diff
@@ -190,121 +383,80 @@ impl TableProvider for S3VectorsQueryTable {
...
             .compute_vector
             .compute_vector(self.query.as_str())
             .await
             .map_err(DataFusionError::External)?;
 
-        if self.partition_by.is_empty() {
-            return Ok(self.no_partitioning_exec(projection, filters, limit, query_vector));
-        }
+        let (_, bucket_name, index_name) = {
+            let idx = match self.table.idx.lock() {
```
phillipleblanc: Here we are taking a lock on `table.idx` just to get `index_name` which `create_spill_plan_query` then immediately parses to get the base name again so that it can run a list and find all of the related indexes.

It seems we only really need the mutex in the write path, whereas in the read path we don't really care what the current spill index is. How about we keep track of the spill index separately in a Mutex/RWLock and then keep the identifier static. That way we wouldn't need to take the lock (twice!) for every read request.

crates/data_components/src/s3_vectors/list_provider.rs
```diff
@@ -77,6 +80,193 @@ impl S3VectorsListTable {
...
+) -> DataFusionResult<Option<Arc<dyn ExecutionPlan>>> {
+    // Use the base name (without spill suffix) as prefix to get all related indexes
+    let base_name = if let Ok(Some(spill)) = SpillIndex::parse(index_name) {
+        spill.base_name
+    } else {
+        index_name.to_string()
+    };
+
+    let list_indexes_output = client
+        .list_indexes(
```
phillipleblanc: this list indexes code is repeated for the query provider as well, potentially a candidate for refactoring

crates/data_components/src/s3_vectors/list_provider.rs
```diff
@@ -113,115 +303,56 @@ impl TableProvider for S3VectorsListTable {
...
-                    );
-                    None
-                }
-            })
-            .collect();
-
-        if index_names.is_empty() {
-            return Ok(Arc::new(EmptyExec::new(project_schema(
-                &self.schema(),
+        if let (Some(bucket_name), Some(index_name)) = (bucket_name, index_name)
```
phillipleblanc, line 251: it isn't clear to me how spilling is enabled? it seems like this code path is always taken, in which case the partitioning is never applied.

crates/data_components/src/s3_vectors/vector_table.rs
kczimm: It *does* seem like it is of little actual use. It's easy to do but will touch a lot of lines. I think it will be best to do in a follow-up PR. Are we concerned about backwards compatibility here though?

crates/data_components/src/s3_vectors/list_provider.rs
kczimm: We check on the line below (as part of the conditional) if we can get a spill plan. It lists the indexes with the base name as prefix and checks if there are spill variants. If spills are found it returns a union plan otherwise it proceeds to partitioning logic.

crates/data_components/src/s3_vectors/spill.rs
kczimm: it splits on `.` treating only the last part as the potential sequence number and joins the preceding parts as the base name. I've written a bunch of tests for this.

crates/data_components/src/s3_vectors/query_provider.rs
kczimm: I separated the table and spill index. used an atomic integer for the spill index for performance.

sgrebnov (dismissed):

crates/data_components/src/s3_vectors/vector_table.rs
phillipleblanc: Yeah this can be done in a follow-up. We can still technically support it, but we can parse out the bucket/index names from the ARN instead of treating it as a separate identifier

crates/data_components/src/s3_vectors/spill.rs
phillipleblanc: Right, but this is assuming we will always have a spill index in the index name - is that always the case?

crates/data_components/src/s3_vectors/list_provider.rs
phillipleblanc: Does that imply partitioning cannot work with spilling? How is spilling enabled?

crates/data_components/src/s3_vectors/spill.rs
```diff
@@ -0,0 +1,254 @@
...
+limitations under the License.
+*/
+
+use snafu::prelude::*;
+
+/// The separator used between the base index name and spill sequence number.
+const SPILL_SEPARATOR: &str = ".";
+
+/// Maximum sequence number for spill indexes (00-99).
+pub const MAX_SPILL_SEQUENCE: u8 = 99;
```
Jeadie, line 23: Is this a hard constraint?

Jeadie (dismissed):

crates/data_components/src/s3_vectors/spill.rs
kczimm: No just a practical one. current limit is 50M vectors / index so spilling enables, in this case, 5B vectors per virtual index. We are encoding index metadata in the index name which makes a lot of things easier. The name length is restricted to 64 characters. The index partitioning + spilling enables much more than 5B vectors per dataset. Perhaps once S3V becomes GA or in the future, the 50M limit could be increased and then the limit for virtual indexes will naturally increase, also. So 100 spill indexes seems reasonable. If we need, we can increase this in the future.

crates/data_components/src/s3_vectors/spill.rs
kczimm: Refactored out.

crates/data_components/src/s3_vectors/list_provider.rs
kczimm: We recurse into `S3VectorsListTable::scan` for each partition passing no expressions:

```rust
            let list_table =
                S3VectorsListTable::new(index_table, table.column_name.clone(), vec![]);

            let index_plan = list_table.scan(state, projection, filters, limit).await?;
            index_plans.push(index_plan);
```

krinart (approved):

sgrebnov (approved):

PR #7666, opened 2025-10-23
'runtime-datafusion' crate for runtime related DataFusion components

Jeadie:
## 📝 Summary
 - Make a new crate 'runtime-datafusion' that has Datafusion components relevant to Spice (i.e. likely not upstreamable), or depend on other crates. (e.g. `runtime-request-context`).


crates/runtime-datafusion/src/extension/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#[async_trait]
+impl QueryPlanner for ExtensionPlanQueryPlanner {
+    async fn create_physical_plan(
+        &self,
+        logical_plan: &LogicalPlan,
+        session_state: &SessionState,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        let physical_planner =
+            DefaultPhysicalPlanner::with_extension_planners(self.extension_planners.clone());
```
phillipleblanc: create the DefaultPhysicalPlanner in the `with_extension_planners` constructor? That way we don't need to initialize it on every query? Can use a OnceLock to defer the initialization in the case where no extension planners are provided.

phillipleblanc (dismissed):

crates/runtime-datafusion/src/extension/mod.rs
Jeadie: I will move to constructor, but i don't think we need `OnceLock` yet since we always initialise with extensions.

Jeadie:
Interesting issue in integ tests with `BytesProcessedExec`.

```
thread 'acceleration::refresh::refresh_modes::test_acceleration_refresh_sqlite_full_variant' panicked at crates/runtime-datafusion/src/extension/bytes_processed.rs:401:13:
  The request context was not provided to BytesProcessedExec, report a bug at https://github.com/spiceai/spiceai/issues
```

peasee (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #7689, opened 2025-10-24
Spawn `object_store` IO tasks on the original Tokio runtime

### LINKED ISSUE #7677: Run object_store IO on original Tokio runtime

phillipleblanc:
## 📝 Summary

When using a separate CPU runtime for DataFusion execution, we still want IO-bound tasks to be executed on the main Tokio runtime with all other IO.

This PR wires that up using the `SpawnedReqwestConnector` in the `object_store` crate that was designed for that purpose. There is a bunch of associated wiring to pipe the IO runtime everywhere it needs to be.

## 🔗 Related

- Closes #7677

### DIFF
--- a/crates/aws-sdk-credential-bridge/src/lib.rs
+++ b/crates/aws-sdk-credential-bridge/src/lib.rs
@@ -124,6 +124,7 @@ pub fn from_s3_url_and_config(
     url: &url::Url,
     region: Option<String>,
     sdk_config: &SdkConfig,
+    io_runtime: Handle,
 ) -> Result<Box<dyn ObjectStore>> {
     if url.scheme() != "s3" {
         return Err(Error::NotAnS3Url {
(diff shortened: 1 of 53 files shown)


crates/aws-sdk-credential-bridge/src/lib.rs
```diff
@@ -98,7 +98,9 @@ pub async fn from_s3_url(url: &url::Url, region: Option<String>) -> Result<Box<d
     }
 
     let bucket_name = get_bucket_name(url)?;
-    let mut builder = AmazonS3Builder::from_env().with_bucket_name(bucket_name);
+    let mut builder = AmazonS3Builder::from_env()
+        .with_bucket_name(bucket_name)
+        .with_http_connector(SpawnedReqwestConnector::new(Handle::current()));
```
mach-kernel, line 103: Can/should this fn now take a runtime handle? Maybe option if can't provide everywhere?

mach-kernel (dismissed):
Makes sense to me. Only general question is whether the usages that default to `Handle::current` have a not-frustrating way of using the handle bound to the DF builder

crates/aws-sdk-credential-bridge/src/lib.rs
phillipleblanc: I added this to make it easy to audit the places we could do a SpawnedReqwest. This particular usage happens when resolving the spicepod itself from S3, so its before we have the runtime loaded/built.

phillipleblanc:
> Makes sense to me. Only general question is whether the usages that default to `Handle::current` have a not-frustrating way of using the handle bound to the DF builder

All of the places that can use it are using it, the only places that use Handle::current are in tests and for retrieving the Spicepod itself from S3, which happens before the runtime is loaded.

mach-kernel (approved):

PR #7695, opened 2025-10-24
Add streaming option to /nsql endpoint

kczimm:
## 📝 Summary

Adds a `stream` payload parameter to the `/nsql` endpoint.

When used, it streams the response with keep alive events, useful for long running queries.

Example (query is made artificially longer than the keep alive interval):

```bash
curl -XPOST "http://localhost:8090/v1/nsql" \
       -H "Content-Type: application/json" -N \
       -d '{
             query": "Which vendors have made the most trips in 2024?", "stream": true
         }'
: nsql still in progress

data: [{"VendorID":2,"TripCount":2234617},{"VendorID":1,"TripCount":729732},{"VendorID":6,"TripCount":260}]
```


crates/runtime/src/http/v1/mod.rs
```diff
@@ -224,7 +225,9 @@ pub async fn to_http_response(
     cache_status: CacheStatus,
     format: ResponseMimeType,
     meta: ResponseMetadata,
-) -> Response {
+) -> (StatusCode, HeaderMap, String) {
```
kczimm, line 228: This refactor makes it easy to get the data from the nsql handler rather than having to extract it from the Response body

sgrebnov (dismissed):

crates/runtime/src/http/v1/nsql.rs
```diff
@@ -57,6 +61,9 @@ const DEFAULT_NSQL_RETRIES: u8 = 10;
 // Maximum number of concurrent sampling tools executions for NSQL
 const DATA_SAMPLING_MAX_CONCURRENT: usize = 10;
 
+// NSQL streaming keep alive interval in seconds
+const NSQL_STREAM_KEEP_ALIVE: u64 = 10;
```
lukekim:
```suggestion
const NSQL_STREAM_KEEP_ALIVE: u64 = 30;
```

lukekim (dismissed):

mach-kernel (dismissed):

sgrebnov (approved):

Issue #7722, opened 2025-10-27
Enhancement: Search on views


Jeadie:
## Goal-State/What/Result
All search paradigms can be defined on any spicepod defined view.

## Why/Purpose
Views are commonly used to prestructure or simplify one or more underlying physical database tables. Users use Spice to provide search capabilities to their apps and AI systems, and this should extend to structured data broader than just `datasets`.

## By When
**Issue/Spec written and reviewed:** 27th October
**Done-Done** 30th October
  - Support full-text search: 28th October
  - Support index-based vector search: 29th October
  - Support vector search: 30th October

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials

## Specification
There are two separate paradigms in which `views` could be searchable akin to the current `datasets`
  1. Explicitly define search atop view columns.
  2. Propagating searchable columns from a view's underlying dataset(s).

The interface and system design of these two approaches can be considered separately. Currently, this enhancement will outline only 1.

### Search atop views
Like datasets, views can be accelerated, and the view table can have column-level configuration (i.e. `.views[].columns[*]`).

#### Interface
From a `spicepod.yaml` UX, configuring search, full-text or vector, precomputed, just-in-time or indexed, should be identical to the equivalent configuration on a `dataset`.

Similarly, for vector-engine based search, a view must be accelerated (as to provide data to be ingested by the vector engine).

The only change in spicepod.yaml definition for this approach is the top level `vectors` field. The details of which will be identical to the dataset [equivalent](https://spiceai.org/docs/reference/spicepod/datasets#vectors).

#### Implementation & Architecture
**Index-based search**
`datasets` that have indexes are wrapped by `IndexedTableProvider` at startup. For any acceleration refresh (i.e. within a `RefreshTask`), the session will use the `IndexTableScanOptimizerRule` which will, for any `IndexedTableProvider`, add a logical extension node above the table scan. The node, `IndexTableScanNode`, will then receive data from the underlying dataset, and provide it to each search index so as to let it ingested appropriately (before being subsequently inserted into the acceleration).

[...]

PR #7724, opened 2025-10-27
Simplify `AcceleratedTable`

Jeadie:
## 📝 Summary
- Whilst working on #7722, I noticed that `impl Accelerated Table` had several superflous functions (`get_federation_provider_for_accelerator` and `create_federated_table_provider`).

### DIFF
--- a/crates/runtime/src/accelerated_table/federation/mod.rs
+++ b/crates/runtime/src/accelerated_table/federation/mod.rs
@@ -21,66 +21,51 @@ use crate::component::dataset::acceleration::ZeroResultsAction;
 use data_components::poly::PolyTableProvider;
 use datafusion::datasource::TableProvider;
 use datafusion_federation::{
-    FederatedTableProviderAdaptor, FederatedTableSource, FederationProvider, sql::SQLTableSource,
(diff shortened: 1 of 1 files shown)


phillipleblanc (approved):

crates/runtime/src/accelerated_table/federation/mod.rs
```diff
@@ -21,66 +21,51 @@ use crate::component::dataset::acceleration::ZeroResultsAction;
...
-            .and_then(|source| {
-                source
-                    .as_any()
-                    .downcast_ref::<SQLTableSource>()
-                    .map(SQLTableSource::table_reference)
-            })?;
-
         let fed_provider = Arc::new(AcceleratedTableFederationProvider::new(
             enabled,
-            accelerated_table_federation_provider.map(|x| x as Arc<dyn FederationProvider>),
```
Jeadie, line 60: Don't need to map anymore. `accelerated_table_federation_provider` was an `Option`, but we had still required it to exist when computing `remote_table_name` above. 

Early returning when `accelerated_table_federation_provider.is_none()` makes this cleaner.

PR #7733, opened 2025-10-27
Full-text search on views

Jeadie:
## 📝 Summary
Enable full text search on views.
```yaml
views:
  - name: foo
    sql: |
      select * from q1
    acceleration:
      enabled: true
    columns:
      - name: question1
        full_text_search:
          enabled: true
          row_id: id

datasets:
  - from: file:/Users/jeadie/Github/spiceai/data/quora_duplicate_questions_1000.tsv
    name: q1
```

## 🔗 Related
 - Apart of #7722


crates/runtime/src/component/column.rs
```diff
@@ -0,0 +1,114 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+use spicepod::semantic::{Column, FullTextSearchConfig, IndexStore};
+
+use crate::component::dataset::FullTextSearchDatasetConfig;
+
+#[allow(clippy::type_complexity)] // From a two-part `.unzip()`.
+#[must_use]
+pub fn full_text_search_config(
```
Jeadie, line 25: This function is logically equivalent to the removed `Dataset::full_text_search_config` with input parameters changed.

crates/runtime/src/search/full_text/connector.rs
```diff
@@ -59,7 +61,7 @@ impl FullTextConnector {
             index_path,
             search_fields,
             primary_key,
-        }) = dataset.full_text_search_config()
+        }) = full_text_search_config(&dataset.columns, &dataset.name)
```
phillipleblanc: Can this function be removed now?

crates/runtime/tests/models/mod.rs
```diff
@@ -217,17 +229,38 @@ pub fn get_mega_science_dataset(
...
+//  - An arbitrary but unique `id` integer column is added.
+pub fn get_mega_science_view(
+    spice_name: Option<&str>,
+    question_column: Option<spicepod::semantic::Column>,
+    answer_column: Option<spicepod::semantic::Column>,
+) -> (Dataset, View) {
+    let mut dataset = mega_science_dataset(Some("mega_science_ds"), false);
+
+    let mut v = View::new(spice_name.unwrap_or("megascience").to_string());
+    v.sql = Some("SELECT * FROM mega_science_ds".to_string());
```
phillipleblanc: Maybe worth considering a slightly more complicated view?

phillipleblanc (dismissed):

crates/runtime/tests/models/mod.rs
lukekim: @Jeadie please test views with joins and unions (the two primary use-cases for views).

crates/runtime/src/datafusion/mod.rs
```diff
@@ -1931,8 +1917,9 @@ async fn wait_until_dependent_tables_are_ready(
         .collect::<Vec<_>>();
 
     let _ = retry(retry_strategy, || async {
-        let statuses = runtime_status
-            .get_dataset_statuses()
+        let mut table_statuses = runtime_status.get_dataset_statuses();
```
Jeadie, line 1920: `wait_until_dependent_tables_are_ready` (used in `register_view`) can now wait for other views to be ready (i.e. if a view has a dependent, accelerated view ).

phillipleblanc (approved):

PR #7783, opened 2025-10-28
1.8.3 post-release housekeeping

mach-kernel:
See diff


spicepod.yml
```diff
@@ -1,58 +0,0 @@
-version: v1
```
sgrebnov, line 1: Please double check it is safe to delete this

sgrebnov (dismissed):

spicepod.yml
mach-kernel: Accidental

lukekim (approved):

PR #7785, opened 2025-10-28
Run retention_sql on refresh commit for DuckDB

lukekim:
Runs the `retention_sql` against the DuckDB acceleration before the final commit in an accelerated refresh. This ensures that rows that should be deleted are never visible in the acceleration after a refresh completes.


crates/runtime/src/datafusion/retention_sql.rs
```diff
@@ -30,6 +30,13 @@ use tokio::runtime::Handle;
 
 use crate::datafusion::builder::get_df_default_config;
 use runtime_object_store::registry::default_runtime_env;
+
+#[derive(Clone, Debug)]
+pub struct ParsedRetentionSql {
```
Jeadie, line 35: nit: docs

Jeadie (dismissed):

phillipleblanc (approved):

Issue #7806, opened 2025-10-29
Cayenne does not push down filter predicates


peasee:
## Describe the bug

_A clear and concise description of what the bug is._

Pepper does not push down filter predicates:

```
FilterExec: r_name@1 = ASIA                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
|               |                               SchemaCastScanExec                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
|               |                                 RepartitionExec: partitioning=RoundRobinBatch(22), input_partitions=1                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
|               |                                   DataSourceExec: file_groups={1 group: [[home/owner/spiceai/test/.spice/data/region/4/019a3226-ec6f-7993-833b-bacf30235f36/rkm3egXDhmvU8WrT_0.vortex]]}, projection=[r_regionkey, r_name], file_type=vortex
```

Compare this to a Parquet reader:

```
DataSourceExec: file_groups={1 group: [[home/owner/spiceai/test/tpch_sf1/region.parquet]]}, projection=[r_regionkey, r_name], file_type=parquet, predicate=r_name@1 = ASIA, pruning_predicate=r_name_null_count@2 != row_count@3 AND r_name_min@0 <= ASIA AND ASIA <= r_name_max@1, required_guarantees=[r_name in (ASIA)]
```

## To Reproduce

_Steps to reproduce the behavior:_

2. Click on '...'
3. Scroll down to '...'

## Expected behavior

_A clear and concise description of what you expected to happen._

## Runtime Details

### Spicepod

_Add the relevant `spicepod.yml` section._

### Output of `describe table`

_Add any relevant `describe table` output here._

### Output of `explain query`

_Add any relevant `explain query` output here._

E.g. `explain select 1`

```bash
+---------------+--------------------------------------+
| plan_type     | plan                                 |
+---------------+--------------------------------------+
| logical_plan  | Projection: Int64(1)                 |
|               |   EmptyRelation                      |
| physical_plan | ProjectionExec: expr=[1 as Int64(1)] |
|               |   PlaceholderRowExec                 |
|               |                                      |
+---------------+--------------------------------------+
```

### spice, spiced, OS info

- spice version: `spice version`
[...]

PR #7828, opened 2025-10-31
fix: Improve join reordering by ensuring `JoinSelection` is applied

peasee:
## 📝 Summary

The outcome of this PR is that the `JoinSelection` optimization is applied more frequently and reliably. This is achieved through:

1. Ensuring `ListingOptions` supply the configured `SessionOptions`, or set the default `SessionOptions` which enable collecting file statistics. `JoinSelection` uses `ExecutionPlan` statistics to determine if the plan would benefit from re-ordering.
2. Refactoring `BytesProcessed` to use a physical optimizer, instead of being inserted into the logical plan. This also resolves the duplicate `BytesProcessedExec` nodes from appearing in the plan, as a leaf will only ever produce one `BytesProcessedExec` before jumping the tree
3. Ensuring both `BytesProcessed` and `SchemaCastScanExec` passthrough their statistics and metrics functions into the underlying `ExecutionPlan` which they wrap. This ensures statistics are returned correctly, which `JoinSelection` requires to function.

These improvements can significantly increase the performance of `ListingTable`-based datasets, like plain S3/File connectors, or the Pepper accelerator.

**File (parquet)**

* Before (TPCH SF1, Q5): 0.92 seconds
* After (TPCH SF1, Q5): 0.74 seconds - 19% improvement

**Pepper**

* Before (`pepper` with TPCH SF1, custom query): 0.56 seconds
* After (`pepper` with TPCH SF1, custom query): 0.43 seconds - 23% improvement

Custom query:

```
select count(1)
from lineitem, orders, customer, nation
where
l_orderkey = o_orderkey
and o_custkey = c_custkey
and c_nationkey = n_nationkey
and n_name = 'JAPAN'
and c_mktsegment = 'BUILDING';
```

* Before (`pepper` with TPCH SF1, Q5): 0.47 seconds
* After (`pepper` with TPCH SF1, Q5): 0.48 seconds

No improvement here, but this is partially expected because Pepper does not yet support dynamic filters: #7806 

## 🔗 Related

* We really need to do #7827. There are various other execution plans which contain wraps or some kind that would benefit from this.


crates/runtime-datafusion/src/extension/bytes_processed.rs
```diff
@@ -20,13 +20,12 @@ limitations under the License.
 //! Adds telemetry to leaf nodes (i.e. `TableScans`) to track the number of bytes scanned during query execution.
 use arrow::datatypes::SchemaRef;
 use arrow::record_batch::RecordBatch;
-use async_trait::async_trait;
+use datafusion::common::tree_node::TransformedResult;
```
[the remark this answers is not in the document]
peasee: This is used.

crates/runtime-datafusion/src/extension/bytes_processed.rs
```diff
@@ -160,7 +145,8 @@ impl Stream for BytesProcessedStream {
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
         match self.inner.poll_next_unpin(cx) {
             Poll::Ready(Some(Ok(batch))) => {
-                self.bytes_processed += batch.get_array_memory_size() as u64;
+                let array_size: usize = batch.get_array_memory_size();
```
peasee, line 148: my rust-analyzer kept thinking `get_array_memory_size()` returned a `RecordBatch` for some reason...

Cargo.toml
```diff
@@ -288,7 +288,7 @@ datafusion-sql = { git = "https://github.com/spiceai/datafusion.git", rev = "41d
 
 datafusion-table-providers = { git = "https://github.com/datafusion-contrib/datafusion-table-providers.git", rev = "7d4fa6d36b464a72afea88e7d3644e143c2535ad" } # spiceai
 
-datafusion-federation = { git = "https://github.com/spiceai/datafusion-federation.git", rev = "d6e7ebd853463acafaf8132dced23c98c45f7947" } # spiceai-50
+datafusion-federation = { git = "https://github.com/spiceai/datafusion-federation.git", rev = "338e177cfe385f6fc759ec4bdc2c65d4d9807083" } # spiceai-50
```
phillipleblanc: Is this still required?

phillipleblanc (dismissed):

Cargo.toml
peasee: no, it is not

crates/runtime-datafusion/src/extension/bytes_processed.rs
```diff
@@ -36,97 +35,83 @@ use datafusion::physical_plan::filter_pushdown::{
...
+impl std::fmt::Debug for BytesProcessedPhysicalOptimizer {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("BytesProcessedPhysicalOptimizer").finish()
     }
 }
 
-#[async_trait]
-impl ExtensionPlanner for BytesProcessedExtensionPlanner {
-    async fn plan_extension(
+impl PhysicalOptimizerRule for BytesProcessedPhysicalOptimizer {
```
Jeadie, line 75: nit: could also move to `datafusion-optimizer-rules`.

Jeadie (approved):

PR #7852, opened 2025-11-02
Improve model components sanitation

lukekim:


crates/data_components/src/databricks/delta.rs
```diff
@@ -93,7 +93,8 @@ impl DatabricksDelta {
         let uc_client = UnityCatalog::new(
             self.endpoint.clone(),
             Some(Arc::clone(&self.token_provider)),
-        );
+        )
+        .map_err(Box::<dyn std::error::Error + Send + Sync>::from)?;
```
Jeadie:
```suggestion
        .boxed()?
```

crates/data_components/src/unity_catalog.rs
```diff
@@ -113,12 +114,18 @@ impl UnityCatalog {
             };
         }
 
-        Self {
+        let client = reqwest::Client::builder()
```
Jeadie, line 117: Should these be spicepod params? (i.e. we have `client_timeout`)

PR #7854, opened 2025-11-02
Sanitization and security checks

lukekim:
This pull request introduces several important security and reliability improvements, particularly around model file handling, as well as some dependency and infrastructure updates. The most significant changes are the addition of strict path sanitization and validation when handling model downloads, ensuring that model files and directories cannot be misused for path traversal or other attacks. There are also updates to dependencies and Dockerfile user management.

**Security and Model File Handling Improvements:**

* Model file and directory names are now strictly validated and sanitized to prevent path traversal and ensure files stay within the intended model directory. This includes:
  - Adding new error types (`InvalidModelName`, `InvalidModelFilePath`) and logic in `ensure_model_path` to reject absolute paths or parent directory segments in model names. [[1]](diffhunk://#diff-5b991ccb3a04a7c0e7ed1a8c6b253dbde21c68fe561384a31c2d688263f7fa6bR68-R77) [[2]](diffhunk://#diff-5b991ccb3a04a7c0e7ed1a8c6b253dbde21c68fe561384a31c2d688263f7fa6bL144-R175)
  - Sanitizing and validating revision, version, and file names in both `huggingface` and `spiceai` model sources, and introducing a utility function (`resolve_model_file_path`) to ensure resolved paths never escape the model root directory. Comprehensive tests are included. [[1]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022L51-R67) [[2]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022L85-R122) [[3]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022R132-L121) [[4]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022L139-R287) [[5]](diffhunk://#diff-5b991ccb3a04a7c0e7ed1a8c6b253dbde21c68fe561384a31c2d688263f7fa6bR24) [[6]](diffhunk://#diff-416b429d966e02a0d65bc528206f2c85277812b7dad4cabdb41f80f4ca1c36a2L98-R107) [[7]](diffhunk://#diff-416b429d966e02a0d65bc528206f2c85277812b7dad4cabdb41f80f4ca1c36a2L137-R154)
  - Ensuring model file downloads are validated as non-empty before writing, and all file writes are performed in a blocking task for safety. [[1]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022L139-R287) [[2]](diffhunk://#diff-416b429d966e02a0d65bc528206f2c85277812b7dad4cabdb41f80f4ca1c36a2L151-R188)

**Dependency Updates:**

* Upgraded `odbc-api` dependency from version 13 to 14 in both the workspace and `db_connection_pool` crate. [[1]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L179-R177) [[2]](diffhunk://#diff-107b5c09484457920310077d4839b0b33af4705e144d0f4a436176e5aae8722fL22-R22)
* Added new dependencies `tokio` and `util` to `model_components`.
* Added the `limit` feature to the `tower-http` dependency.

**Infrastructure and Dockerfile Changes:**

* The Dockerfile now creates and uses a dedicated `spiced` user and group for running the application, improving sandboxing and security.

**Other Notable Improvements:**

* Improved error handling and code clarity in embedding and Databricks Delta code paths. [[1]](diffhunk://#diff-eeb662cbc7256872ac199bc7b8eaa8469b9e47bb016777061a9beaca2ed53c8cL121-R136) [[2]](diffhunk://#diff-3fd73cfd866a946db5161050ed8f606cb6e2f3fb45a94b957b3cb74fed8ec88bL97-R97)
* Minor code hygiene: added `#[allow(clippy::too_many_lines)]` where needed, and improved imports for path handling. [[1]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022R25-R33) [[2]](diffhunk://#diff-416b429d966e02a0d65bc528206f2c85277812b7dad4cabdb41f80f4ca1c36a2R28) [[3]](diffhunk://#diff-416b429d966e02a0d65bc528206f2c85277812b7dad4cabdb41f80f4ca1c36a2R37)

**Summary of Most Important Changes:**

**Security and Model File Handling**
- Strict validation and sanitization of model and file names to prevent path traversal and ensure files stay within the model directory. New error types and comprehensive path resolution logic with tests. [[1]](diffhunk://#diff-5b991ccb3a04a7c0e7ed1a8c6b253dbde21c68fe561384a31c2d688263f7fa6bR68-R77) [[2]](diffhunk://#diff-5b991ccb3a04a7c0e7ed1a8c6b253dbde21c68fe561384a31c2d688263f7fa6bL144-R175) [[3]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022L51-R67) [[4]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022L85-R122) [[5]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022R132-L121) [[6]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022L139-R287) [[7]](diffhunk://#diff-5b991ccb3a04a7c0e7ed1a8c6b253dbde21c68fe561384a31c2d688263f7fa6bR24) [[8]](diffhunk://#diff-416b429d966e02a0d65bc528206f2c85277812b7dad4cabdb41f80f4ca1c36a2L98-R107) [[9]](diffhunk://#diff-416b429d966e02a0d65bc528206f2c85277812b7dad4cabdb41f80f4ca1c36a2L137-R154)
- Model downloads now validate non-empty content and use blocking tasks for file writes to improve reliability and prevent partial/corrupt files. [[1]](diffhunk://#diff-90bb55b396b30e33ea37dafd061149b75958490b1b34687c4641acc958049022L139-R287) [[2]](diffhunk://#diff-416b429d966e02a0d65bc528206f2c85277812b7dad4cabdb41f80f4ca1c36a2L151-R188)

**Dependency Updates**
- Upgraded `odbc-api` to version 14 and added new dependencies (`tokio`, `util`), and enabled the `limit` feature for `tower-http`. [[1]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L179-R177) [[2]](diffhunk://#diff-107b5c09484457920310077d4839b0b33af4705e144d0f4a436176e5aae8722fL22-R22) [[3]](diffhunk://#diff-18f2b3b4937e2aaed97297022c75318e9ae51b076543558e3a9cdc32d4ce7d01R27-R31) [[4]](diffhunk://#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542L243-R241)

**Infrastructure**
- Dockerfile updated to use a dedicated `spiced` user for improved sandboxing and security.


phillipleblanc (changes requested):
Thanks for the security-focused cleanup! I found a few blockers that we should address before merging:\n\n1.  –  is set to the parent of the model directory (). Because  only errors once it would pop *past* , any entry with two  segments (e.g. ) still escapes the current model folder and drops files into sibling models. That contradicts the new "files must stay within the model directory" guarantee and allows a malicious repo to overwrite other cached models. Can we clamp the boundary to the actual model directory (pass  to ) so downloads can never walk into other models?\n2.  – the inlined ODPI-C install script no longer installs  and unconditionally runs . On ubuntu-latest (currently 22.04) that path does not exist, so the step fails before we set . Even on 24.04 the t64 library only appears after installing . Could we restore the apt install and guard the symlink like the deleted composite action did so the workflow runs again?\n3.  – the new tool-name validation rejects , but catalog tools are intentionally exposed as  (see ). Every catalog-backed tool now hits "invalid characters" before we look it up, effectively disabling catalogs. Could we permit  (or split the catalog/tool components before validating)?

phillipleblanc (dismissed):
Thanks for the security-focused cleanup! I spotted a one issue we should address before merging:

1. crates/model_components/src/modelsource/huggingface.rs lines 66-138 – root_dir is currently set to the parent of the model directory (~/.spice/models). Because resolve_model_file_path only errors once it would pop past root_dir, any files entry with two .. segments (for example ../../other-model/payload.bin) still escapes the current model folder and drops files into sibling models. That contradicts the new guarantee that files must stay within the model directory and lets a malicious repo overwrite other cached models. Can we clamp the boundary to the actual model directory (local_path.as_path()) so downloads cannot walk into other models?

phillipleblanc (dismissed):

phillipleblanc (dismissed):

sgrebnov (dismissed):

crates/data_components/src/object/mod.rs
```diff
@@ -45,7 +45,32 @@ impl ObjectStoreContext {
     ) -> Result<Self, Box<dyn std::error::Error + Send + Sync>> {
         let (prefix, filename_regex_opt) = parse_prefix_and_regex(url, extension)?;
         let filename_regex = filename_regex_opt
-            .map(|regex| Regex::new(&regex).boxed())
+            .map(|regex| {
```
mach-kernel, line 48: Is this for globbing? If so you can use the glob crate, which already has some decent amount of validations for acceptable patterns:

https://docs.rs/glob/latest/glob/struct.Pattern.html

crates/data_components/src/object/mod.rs
```diff
@@ -45,7 +45,32 @@ impl ObjectStoreContext {
...
+                const MAX_REGEX_LENGTH: usize = 1000;
+                if regex.len() > MAX_REGEX_LENGTH {
+                    return Err(format!(
+                        "Regex pattern too long ({} chars). Maximum allowed: {MAX_REGEX_LENGTH}",
+                        regex.len()
+                    )
+                    .into());
+                }
+
+                // Additional ReDoS protection: detect dangerous patterns
```
mach-kernel: If this is for glob support, use the glob crate's pattern struct: https://docs.rs/glob/latest/glob/struct.Pattern.html

It already has some validations for acceptable patterns which should be more restrictive than for an arbitrary regex

crates/runtime-datafusion-udfs/src/embed.rs
```diff
@@ -118,17 +118,22 @@ impl Embed {
         let embedding = model
             .embed_sync(EmbeddingInput::String(sentence.to_owned()))
             .map_err(|e| DataFusionError::External(Box::new(e)))?;
-        let vector_size = match embedding.first() {
-            Some(embedding) => embedding.len(),
-            _ => unreachable!("Should have at least one embedding"),
-        };
+
+        let first_embedding = embedding.first().ok_or_else(|| {
+            DataFusionError::Internal(
```
mach-kernel: `ExecutionError` maybe? We expect there to be one but there isn't.

crates/db_connection_pool/Cargo.toml
```diff
@@ -19,7 +19,7 @@ datafusion.workspace = true
 datafusion-table-providers = { workspace = true }
 dyn-clone = { version = "1.0.17", optional = true }
 futures.workspace = true
-odbc-api = { workspace = true, optional = true }
+odbc-api = { version = "14", optional = true }
```
Jeadie: Base `cargo.toml` is on [version 14](https://github.com/spiceai/spiceai/blob/trunk/Cargo.toml#L185). We should keep as `odbc-api.workspace = true`

crates/model_components/src/modelsource/huggingface.rs
```diff
@@ -22,13 +22,15 @@ use snafu::prelude::*;
...
 use std::collections::HashMap;
 use std::io::Cursor;
+use std::path::{Component, Path, PathBuf};
 use std::sync::Arc;
 use std::time::Duration;
 
 pub struct Huggingface {}
 
 #[async_trait]
 impl ModelSource for Huggingface {
```
Jeadie, line 32: Also this is just for machine learning models BTW (like ONNX).

Dockerfile
```diff
@@ -115,20 +115,25 @@ RUN echo 'nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin' > /spice_s
...
+RUN chown -R 999:999 /spice_sandbox/.duckdb
 
 # Create HuggingFace cache directory in sandbox
 RUN mkdir -p /spice_sandbox/.cache/huggingface/hub
-RUN chown -R 65534:65534 /spice_sandbox/.cache
+RUN chown -R 999:999 /spice_sandbox/.cache
 RUN chmod -R 755 /spice_sandbox/.cache
 
+# Create spiced user and group in sandbox
+RUN echo 'spiced:x:999:999::/app:/usr/sbin/nologin' >> /spice_sandbox/etc/passwd && \
```
Jeadie: Do we need both a `nobody` and a `spiced` user?

crates/data_components/src/object/mod.rs
lukekim: We use glob crate for filtering, but this is validation for URLs, so I don't think we can use it.

sgrebnov (dismissed):

sgrebnov (dismissed):

crates/model_components/src/modelsource/huggingface.rs
```diff
@@ -136,15 +161,132 @@ impl ModelSource for Huggingface {
...
+            Component::CurDir => {}
+            Component::Normal(segment) => candidate.push(segment),
+            Component::ParentDir => {
+                if candidate == root_dir || !candidate.pop() {
+                    return Err(super::InvalidModelFilePathSnafu {
+                        path: file.to_string(),
+                    }
+                    .build());
+                }
+            }
```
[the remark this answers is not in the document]
phillipleblanc: No—this one is off.

  - candidate starts at base_dir (a child of root_dir). On ParentDir, we currently error if
    candidate == root_dir (would escape the model root) or if candidate.pop() fails (already at
    filesystem root). That’s the correct guard to block traversal above the model root and to
    catch popping past /.
  - After the loop we also ensure!(candidate.starts_with(root_dir)), so any path that somehow
    slipped outside the root is rejected.
  - The suggested candidate < root_dir comparison isn’t a valid way to test containment, and
    moving the check after pop would allow one pop to escape the model root before we detect
    it.

  So the existing logic already handles the edge cases (root-level pop fails, or trying to go
  above the model root). No change needed.

phillipleblanc (approved):

PR #7896, opened 2025-11-03
fix: Make integration run with no relevant changes, disable makefile targets on PR

peasee:
## 📝 Summary

* Updates the integration test workflow to ensure it runs when there are no changes as the workflows are now required checks for PRs
* Stops makefile targets from running on PRs and only runs it on push to trunk/release. Adds a step to the endgame to verify the workflow.


peasee:
Successful run: https://github.com/spiceai/spiceai/actions/runs/19052943153

lukekim (approved):

PR #7910, opened 2025-11-04
Memoize S3 vectors ListIndex API call with configurable TTL

### LINKED ISSUE #7646: Memoize ListIndexes on query and list vectors
- Memoize ListIndexes with configurable TTL to avoid listing indexes on every scan call.
- Minimum of 5 seconds to avoid excessive API chatter.

A configurable cache TTL can be used to optimize query performance.

```yaml
datasets:
  - from: spice.ai:dataset.with.embeddings
    name: my_dataset
    vectors:
      enabled: true
      engine: s3_vectors
      params:
        s3_vectors_bucket: my-s3-vector-bucket
        s3_vectors_index_poll_interval: 5m
```

kczimm:
## 📝 Summary

- Caches the response of the S3 Vectors ListIndex API
- Invalidates the cache on CreateIndex.
- Adds a parameter `s3_vectors_index_poll_interval`

## 🔗 Related

Closes #7646 

## 📚 Docs

https://github.com/spiceai/spiceai/issues/7693


crates/runtime/src/embeddings/index/s3/client.rs
```diff
@@ -33,12 +33,36 @@ use s3_vectors::{
 
 use crate::timing::TimeMeasurement;
 
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::RwLock;
+use tokio::time::Instant;
+
+const TTL_DURATION_MINIMUM: Duration = Duration::from_secs(5);
```
Jeadie, line 42: I know this is mentioned in #7646  as:
> Minimum of 5 seconds to avoid excessive API chatter.

But I do wonder whether we want to explicitly block a user (who may with good intentions be happy with the tradeoff), from setting <5s.

crates/s3_vectors/src/lib.rs
```diff
@@ -56,13 +56,254 @@ pub use aws_sdk_s3vectors::{
...
+        Self {
+            client: aws_sdk_s3vectors::Client::new(config),
+        }
+    }
+}
+
 pub mod mock;
 
+#[async_trait]
+impl S3Vectors for Client {
```
Jeadie, line 82: This is a separate refactor or required for this memoization?

crates/runtime/src/embeddings/index/s3/client.rs
```diff
@@ -33,12 +33,36 @@ use s3_vectors::{
...
+
 pub struct S3VectorClient {
-    client: Client,
+    client: Arc<dyn S3Vectors + Send + Sync>,
+    list_indexes_cache: RwLock<HashMap<String, (ListIndexesOutput, Instant)>>,
```
Jeadie, line 46: Could use `use cache::SimpleCache`?

kczimm: Cool, I'll take a look.

crates/s3_vectors/src/lib.rs
kczimm: It's how I can use dependency injection for testing using the mock client.

crates/runtime/src/embeddings/index/s3/client.rs
kczimm: They could omit the config parameter and not have any caching? I understand what you're saying though.

crates/runtime/src/embeddings/index/s3/client.rs
kczimm: Looks like it could be used. I'd have to add selective cache invalidation per key (bucket in this case). I'm also using `tokio::time::Instant` so that I can use `pause`, `resume`, and `advance` for efficient timing related testing. I'd have to refactor testing to sleep to ensure I'm on a specific side of the TTL duration. Maybe it's worth a refactor at some point when SimpleCache can be enhanced but would prefer that be in a separate PR.

crates/runtime/src/embeddings/index/s3/client.rs
Jeadie: I can just imagine legitimately wanting a `1s` TTL for example.

Jeadie (approved):

PR #7934, opened 2025-11-05
fix: Use correct minio secret in build binary push

peasee:
## 📝 Summary

* The linux build uses `-dev-runners` so they need the `TEST_*` secrets
* There is no need for the second upload step because they upload to dev anyway

### DIFF
--- a/.github/workflows/build_and_release.yml
+++ b/.github/workflows/build_and_release.yml
@@ -7,46 +7,46 @@ on:
     tags:
       - v*
     paths-ignore:
-      - '**/*.md'
-      - 'docs/**'
-      - 'README.md'
-      - 'Makefile'
-      - 'CONTRIBUTING.md'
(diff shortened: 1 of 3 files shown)


.github/workflows/build_and_release.yml
```diff
@@ -7,46 +7,46 @@ on:
...
-      - 'version.txt'
-      - '.schema/**'
-      - '.vscode/**'
-      - 'deploy/**'
-      - 'install/**'
-      - 'media/**'
-      - 'monitoring/**'
-      - 'acknowledgements.md'
-      - 'Dockerfile*'
+      - "**/*.md"
```
lukekim: Single quote! (see `prettier.rc`)

peasee: I added a `.prettierrc`

ewgenius (approved):

PR #7943, opened 2025-11-06
Fix AWS SDK credential cache retry handling

phillipleblanc:
Fixes an issue where a failure to load the SDK due to a transient issue would prevent Spice from recovering and ever being able to load credentials.


mach-kernel (dismissed):

sgrebnov (dismissed):

kczimm (dismissed):

sgrebnov (dismissed):

crates/data_components/src/delta_lake.rs
```diff
@@ -118,7 +120,8 @@ pub struct DeltaTable {
 }
 
 impl DeltaTable {
-    pub fn from(
+    #[allow(clippy::unused_async)]
```
phillipleblanc: why make this async then?

Cargo.toml
```diff
@@ -227,6 +229,7 @@ tiberius = { version = "0.12.3", default-features = false, features = [
     "rustls",
     "chrono",
 ] }
+tokenizers = "0.21.2"
```
phillipleblanc, line 232: where is this used?

phillipleblanc: looks like a merge issue

lukekim (approved):

PR #7952, opened 2025-11-06
Run integration tests with `AWS_EC2_METADATA_DISABLED` flag

sgrebnov:
## 📝 Summary

Some tests attempted to access the AWS default metadata host (`169.254.169.254`) and failed due to an incompatible response. Since we are not using AWS metadata, disable its usage in tests.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -285,6 +285,7 @@ jobs:
       - name: Run integration test
         if: needs.check_changes.outputs.relevant_changes == 'true'
         env:
+          AWS_EC2_METADATA_DISABLED: true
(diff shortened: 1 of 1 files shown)


peasee (approved):

.github/workflows/integration.yml
```diff
@@ -285,6 +285,7 @@ jobs:
       - name: Run integration test
         if: needs.check_changes.outputs.relevant_changes == 'true'
         env:
+          AWS_EC2_METADATA_DISABLED: true
```
sgrebnov, line 288: We technically require this for `Run integration test` only but I've included the same into `Run AWS SDK integration test` and `Run AWS SDK credential bridge integration test` for consistency

phillipleblanc (approved):

PR #7959, opened 2025-11-07
Infer RRF `join_key` from `TableProvider::constraints` and implement `SearchQueryProvider::constraints`.

Jeadie:
## 📝 Summary
 - Currently the [RRF UDTF](https://spiceai.org/docs/reference/sql/search#reciprocal-rank-fusion-rrf) uses an explicit `join_key` parameter, or will construct an implicit unique identifier based on an MD5 hash of the row, i.e.  `digest_many(col1, col2, .., colN, 'md5')`. 
 - This PR attempts to infer a `join_key` from all table inputs to `rrf`. If all tables share a common primary key, it will be used as the `join_key`. 
 - This PR implements `SearchQueryProvider::constraints` so that all index-based search UDTFs can have their `join_key` inferred.


mach-kernel (approved):
Nicely done

PR #7963, opened 2025-11-07
ResultsCache Cache-Control `stale-while-revalidate`

lukekim:
This pull request introduces support for "stale-while-revalidate" cache control for query results, allowing stale cached data to be served while asynchronously refreshing the cache entry in the background. It also adds tracking of cache uptime and the timestamp of when each cache entry was stored. These changes improve cache flexibility, observability, and adherence to cache control directives, while maintaining backward compatibility and robust error handling.

**Cache control and stale-while-revalidate support:**

* Added a new `CacheControl::CacheWithStaleWhileRevalidate` variant to support serving stale cache entries while refreshing them in the background; includes parsing logic and helper methods for extracting cache key type and stale-while-revalidate duration from HTTP headers. [[1]](diffhunk://#diff-3ba3e9a17a1bea8512de1b5df6e6d714483423c1317a02296fb895e54adca283R62-R102) [[2]](diffhunk://#diff-3ba3e9a17a1bea8512de1b5df6e6d714483423c1317a02296fb895e54adca283L83-R168)
* Extended tests to cover stale-while-revalidate behavior and helper methods, ensuring correct parsing and handling of cache control directives. [[1]](diffhunk://#diff-3ba3e9a17a1bea8512de1b5df6e6d714483423c1317a02296fb895e54adca283R251-R289) [[2]](diffhunk://#diff-3ba3e9a17a1bea8512de1b5df6e6d714483423c1317a02296fb895e54adca283R299-R335)

**Cache entry timestamp and uptime tracking:**

* Updated cache providers (`SimpleCache`, `LruCache`) to track uptime in seconds and expose it via a new trait method `uptime_secs`; cache entries now store their creation offset in seconds for TTL and staleness checks. [[1]](diffhunk://#diff-17e3f218124a1fbc59b445e6d6885a41124ee52a2c5c248256c205b6c10e3343R114) [[2]](diffhunk://#diff-1f71cbb7a0df017de50a2a256bdd9b4743be3d23681d3c50b3616b438eac4bfeR41) [[3]](diffhunk://#diff-1f71cbb7a0df017de50a2a256bdd9b4743be3d23681d3c50b3616b438eac4bfeR83) [[4]](diffhunk://#diff-1f71cbb7a0df017de50a2a256bdd9b4743be3d23681d3c50b3616b438eac4bfeR135-R140) [[5]](diffhunk://#diff-c89da1341e0e65dcd5594953cead7aff510fd7322d5a847c80f8bdca0213e9ceR256-R261) [[6]](diffhunk://#diff-7a744a9751a351841c82c9cc0f8ab770bc74c7a7eb796a39aa2aa4e31ae87546R41-R73)
* Modified cache insertion logic to record the cache provider's uptime when storing a new entry, enabling precise age calculations for staleness.

**Error handling and configuration validation:**

* Added a new error variant to prevent use of stale-while-revalidate unless the cache key type is set to "sql (raw)", with a clear diagnostic message for misconfiguration.

**Internal API and test adjustments:**

* Refactored construction of `CachedQueryResult` in tests and internal code to use the new timestamp-aware constructor. [[1]](diffhunk://#diff-c89da1341e0e65dcd5594953cead7aff510fd7322d5a847c80f8bdca0213e9ceL308-R320) [[2]](diffhunk://#diff-1f71cbb7a0df017de50a2a256bdd9b4743be3d23681d3c50b3616b438eac4bfeL182-R196)

**Request context and cache key handling:**

* Updated request context builder to correctly propagate cache key type and stale-while-revalidate duration, including fallback logic for client-supplied keys.
* Ensured cache key selection logic in query cache supports the new cache control variant. [[1]](diffhunk://#diff-a1f9ce15739f587eb1d09930eeddcaf30c830009b4e5db50e259061839f6435fL179-R183) [[2]](diffhunk://#diff-a1f9ce15739f587eb1d09930eeddcaf30c830009b4e5db50e259061839f6435fR206)

These changes provide more flexible caching strategies, improve cache observability, and ensure robust handling of cache control directives.


crates/cache/src/lib.rs
```diff
@@ -414,6 +415,16 @@ impl QueryResultsCacheProvider {
         self.cache.item_count()
     }
 
+    #[must_use]
+    pub fn ttl(&self) -> std::time::Duration {
+        self.ttl
+    }
+
+    #[must_use]
+    pub fn uptime_secs(&self) -> u32 {
```
peasee: This might be more representative as a `Duration` to avoid conversions to numbers. The new cache control variant is already stored as a duration:

`CacheWithStaleWhileRevalidate(CacheKeyType, Duration)`

There is no reason then to have this `uptime` trait method, as you could store the `Instant::now()` directly on the `CachedQueryResult`. `CachedQueryResult` would update to use `CachedQueryResult::is_stale(&self, ttl: Duration, now: Instant)`. `CachedQueryResult::cached_at_offset_secs` would be updated to `CachedQueryResult::cached_at` which is an `Instant` because `Instant::now()` is relative to the time the application started.

peasee (changes requested):

peasee (dismissed):

peasee (approved):

crates/runtime/src/datafusion/query/cache.rs
```diff
@@ -283,6 +349,107 @@ impl Query {
...
+                super::QueryBuilder::new(&sql, df).build()
+            };
+            let result = background_context
+                .scope(async move { query.run().await })
+                .await;
+
+            match result {
+                Ok(query_result) => {
+                    // Drain the stream to ensure the query executes and cache is populated
+                    match query_result.data.try_collect::<Vec<_>>().await {
```
sgrebnov, line 431: Nit: this collects all batches in memory where we can just read and drop batches, for example:
```rust
​async fn drain_query_stream(mut stream: SendableRecordBatchStream) -> datafusion::error::Result<()> {
    use futures::StreamExt;
    while let Some(batch) = stream.next().await {
        batch?; // propagate errors early
    }
    Ok(())
}
```
```rust
​if let Err(e) = drain_query_stream(query_result.data).await {
    tracing::debug!("Background revalidation failed during collection: {e}");
}
```

PR #7992, opened 2025-11-10
Optimize UDF push-down analyzer rule

lukekim:
This pull request introduces benchmarking for the `DuplicateLogicalPlanNode` analyzer rule in the `runtime-datafusion` crate, optimizes how JSON function names are cached for UDF configuration, and adds minor dependency and module updates. The most significant changes are the addition of criterion-based benchmarks to measure the performance of duplicate plan node detection, and the optimization of JSON function name retrieval to avoid repeated session context creation.

**Benchmarks for analyzer rule:**
* Added a new benchmark suite in `crates/runtime-datafusion/benches/duplicate_plan_node.rs` using Criterion to measure the performance of the `DuplicateLogicalPlanNode` rule under various scenarios (scaling, duplicates, and no duplicates).
* Registered the benchmark in `Cargo.toml` and added `criterion` as a dev dependency for benchmarking support.

**Performance optimization for UDF configuration:**
* Introduced a static `Lazy<Vec<String>>` (`JSON_FUNCTIONS`) in `crates/runtime/src/datafusion/udf.rs` to cache the list of JSON function names, preventing repeated creation of `SessionContext` objects and improving startup efficiency.
* Updated `deny_spice_specific_functions()` to use the cached `JSON_FUNCTIONS` instead of recalculating the list each time, removing the old `json_functions()` helper.

**Module and dependency updates:**
* Added a new `analyzer_rule` module to `crates/runtime-datafusion/src/lib.rs` for better code organization.
* Minor import update to use `once_cell::sync::Lazy` for static initialization.

### DIFF
--- a/crates/runtime/src/datafusion/udf.rs
+++ b/crates/runtime/src/datafusion/udf.rs
@@ -84,7 +84,7 @@ pub async fn register_udfs(runtime: &crate::Runtime) {
     ctx.register_udf(INSTANCE.clone());
 }
 
-/// Cached list of JSON function names to avoid repeated SessionContext creation.
+/// Cached list of JSON function names to avoid repeated `SessionContext` creation.
 /// This is initialized once at startup and reused for all accelerator configurations.
 static JSON_FUNCTIONS: Lazy<Vec<String>> = Lazy::new(|| {
     let mut ctx = SessionContext::new();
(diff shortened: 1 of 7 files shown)


sgrebnov (dismissed):

sgrebnov (dismissed):

sgrebnov (dismissed):

crates/runtime/src/datafusion/udf.rs
```diff
@@ -25,6 +25,7 @@ use crate::search::util::parse_explicit_primary_keys;
 use datafusion::functions::math::random::RandomFunc;
 use datafusion::prelude::SessionContext;
 use datafusion_table_providers::util::supported_functions::{FunctionRestriction, FunctionSupport};
+use once_cell::sync::Lazy;
```
phillipleblanc, line 28: Use the built-in `LazyLock` that was stabilized in Rust 1.80: https://doc.rust-lang.org/std/sync/struct.LazyLock.html

Issue #7641, opened 2025-10-21
Physical plans should include complete object store URLs


mach-kernel:
We load our params into fragments:

https://github.com/spiceai/spiceai/blob/8a04c2977666c802ce04e6458d5d58029c3a0c1c/crates/runtime/src/dataconnector/listing/mod.rs#L89-L100

Those fragments are then used for params passed into `object_store` depending on the URI scheme:
https://github.com/spiceai/spiceai/blob/31cff84e5ddfc757c91e9d4a1616869f167e8e16/crates/runtime-object-store/src/registry/mod.rs#L61-L63

Make sure that `PartitionedFile` entries contain the whole URL including fragments, so that reads do not break on reification on the other side.

Issue #7675, opened 2025-10-23
Ensure TLS intra-cluster communication


mach-kernel:
Execution plans contain sensitive information like data lake paths, in some cases credentials, etc. If TLS is disabled (e.g. local deployment), emit a loud and visible warning in the console.

It might be worth panicking if TLS is not enabled, but the bind address is `0.0.0.0`.

mach-kernel:
This is 50% of the way there. We decided to punt this a bit down the line, but make sure it's there for the beta release. So far:

- https://github.com/spiceai/spiceai/tree/mach/cluster-obj-store-with-tls
  - Working E2E Ballista cluster RPC with a `spiced` API key
  - WIP: Sharing TLS configuration with Ballista
- https://github.com/spiceai/datafusion-ballista/pull/3
  - Changes to allow consuming scheduler RPC with interceptors
  - WIP: Consuming external TLS configuration

This should be easy to take across the line when we can pick it up next.

PR #8009, opened 2025-11-11
Distributed query: Object store initialization across scheduler/executor gap, misc bugfixes & improvements

### LINKED ISSUE #7672: Use protobuf serde for boot RPC and Spice custom execution nodes
Currently this code uses serde_json, but it would be nice to follow the same pattern as DataFusion (for the plan nodes):

https://github.com/apache/datafusion/blob/main/datafusion/proto/src/physical_plan/mod.rs#L255

Like this we can have ser/de defined with each plan node and keep the codec simple.

---

For the `App` struct RPC, it would be nice to have a struct representing the bare-minimum serialized content necessary to put up an executor. We don't use all of the fields. This would be a nice way to obviate that.

mach-kernel:
## 📝 Summary

- New flight RPC action `ExpandSecret`, which is used by executors to resolve the value of a secret from a scheduler
- RPC protobuf definitions for custom plan nodes and the above
- `DeltaTable::create_parquet_exec` now ships the base table scheme + bucket for its `ObjectStoreUrl`
- If trying to start in clustered mode without a TLS configuration, a panic happens
  - Users must acknowledge the message and pass `--allow-insecure-connections`
  - If not starting in cluster mode, this is ignored
- Misc codec cleanup, moving things out of the runtime crate where possible

Closes: #7672, #7641 

## 📚 Docs

WIP

## 👀 Notes for Reviewers

See PR annotations


crates/data_components/src/delta_lake.rs
```diff
@@ -261,8 +266,16 @@ impl DeltaTable {
             .with_parquet_file_reader_factory(Arc::clone(parquet_file_reader_factory))
             .with_predicate(Arc::clone(physical_expr));
 
+        // Matches keying used by `ObjectStoreRegistry::get_url_key`
+        let object_store_url = ObjectStoreUrl::parse(format!(
```
mach-kernel, line 275: Make the Delta `ObjectStoreUrl` bindable across ser/de gap by specifying the base URL/scheme

crates/datafusion-optimizer-rules/src/physical_plan/cluster/distribute_file_scan.rs
```diff
@@ -226,16 +226,37 @@ impl PhysicalOptimizerRule for DistributeFileScanOptimizer {
...
+            let maybe_file_scan = concrete!(p, DataSourceExec)
                 .and_then(|d| concrete!(d.data_source(), FileScanConfig));
 
             let Some(file_scan_config) = maybe_file_scan else {
-                return Ok(Transformed::no(plan));
+                return Ok(Transformed::no(p));
             };
 
+            // Only repartition sufficiently large LIMIT scans
+            // TODO: statistics + check upstream projections for transforms
```
mach-kernel, line 244: https://github.com/spiceai/spiceai/issues/8075

crates/datafusion-optimizer-rules/src/physical_plan/cluster/ensure_supported_file_scan.rs
```diff
@@ -0,0 +1,75 @@
...
+use datafusion::physical_plan::ExecutionPlan;
+use datafusion_datasource::file_scan_config::FileScanConfig;
+use datafusion_datasource::memory::MemorySourceConfig;
+use datafusion_datasource::source::DataSourceExec;
+use std::sync::Arc;
+
+/// An optimizer to sanity check `DataSourceExec` encapsulate the kinds of plans
+/// we can distribute
+#[derive(Debug, Clone)]
+pub struct EnsureSupportedFileScan {}
```
mach-kernel, line 15: This used to live in a codec, moved to an optimizer. Exists to provide a nice error message (vs the failure to serialize).

crates/runtime-proto/proto/spice.proto
```diff
@@ -0,0 +1,24 @@
+syntax = "proto3";
+package spice.protobuf;
+
+option java_multiple_files = true;
+option java_package = "com.spiceai";
+option java_outer_classname = "SpiceProto";
+
```
mach-kernel, line 7: Protos for our RPC needs + plan nodes. I chose to represent nodes opaquely as `bytes` to avoid maintaining parity with e.g. a `datafusion.proto`.

crates/runtime-secrets/src/stores/scheduler_rpc.rs
```diff
@@ -0,0 +1,57 @@
+use crate::{AnyErrorResult, SecretStore};
+use async_trait::async_trait;
+use prost::{Message, bytes};
+use runtime_proto::{ExecutorExpandSecretRequest, ExecutorExpandSecretResponse};
+use secrecy::SecretString;
+
+/// Used by cluster mode to resolve secrets declared in the scheduler
+/// via flight RPC
```
mach-kernel, line 8: This is a new secrets store used exclusively on executors to expand secrets. It purposefully does not cache them locally to avoid situations like: AWS secret is rotated, a new value can be polled from API, but isn't because the local cache is stale.

crates/runtime/src/cluster/datafusion/codec/spice_physical_codec.rs
```diff
@@ -0,0 +1,105 @@
...
+use datafusion_expr::registry::FunctionRegistry;
+use datafusion_proto::generated::datafusion_common;
+use datafusion_proto::physical_plan::PhysicalExtensionCodec;
+use prost::Message;
+use runtime_datafusion::execution_plan::schema_cast::SchemaCastScanExec;
+use runtime_datafusion::extension::bytes_processed::BytesProcessedExec;
+use runtime_proto::{BytesProcessedExecNode, SchemaCastScanExecNode};
+use std::fmt::Debug;
+use std::sync::Arc;
+
```
mach-kernel, line 17: Simplified with new protos. Also in DF50+, our extension is only called if there is no match on a DataFusion native node (so AFAICT we don't even have to fall back, although, doing it for sanity).

crates/runtime/src/flight/actions.rs
```diff
@@ -143,6 +159,94 @@ pub(crate) async fn do_action(
             let result = arrow_flight::Result::new(bs);
             futures::stream::iter(vec![Ok(result)])
         }
+        ActionType::ExpandSecret => {
```
mach-kernel, line 162: New flight action to serve a secret to an executor:
- Emits task history entry
- Only retrieves secrets context IF the executor is a registered member of the cluster

crates/runtime/src/flight/mod.rs
```diff
@@ -466,13 +470,25 @@ pub async fn start(
             .context(UnableToConfigureTlsSnafu)?;
     }
 
+    #[cfg(feature = "cluster")]
+    if tls_config.is_none()
```
mach-kernel, line 474: If starting in cluster mode without a TLS configuration, we panic and show the user this message. They have to acknowledge and pass an extra flag to run the flight service without TLS.

crates/runtime/src/config.rs
```diff
@@ -120,6 +120,14 @@ pub struct ClusterConfig {
         action
     )]
     pub scheduler_url: Url,
+
+    #[arg(
```
mach-kernel, line 124: New flag

crates/runtime/src/cluster/mod.rs
```diff
@@ -279,10 +300,80 @@ async fn executor_bind_app(
...
     rt.load_embeddings().await;
     Arc::clone(rt).load_models().await;
     Arc::clone(rt).load_tools().await;
 
     Ok(())
 }
+
+/// Traverses dataset definitions and reifies `ListingTableUrl`s, triggering object store
+/// registration for each.
+async fn executor_bind_object_stores(rt: Arc<Runtime>) -> crate::Result<()> {
```
mach-kernel, line 316: Eagerly register object stores before any queries run.

crates/runtime-parameters/src/lib.rs
```diff
@@ -265,6 +272,21 @@ impl Parameters {
     pub fn prefix(&self) -> &'static str {
         self.prefix
     }
+
+    /// `delta_lake` manually initializes an object store not via the registry, and it uses
+    /// `aws_` prefixed parameters. But after serialization, `ParquetExec` et al. have S3 URI schemes
+    /// so they initialize using the registry.
+    pub fn canonicalize_s3_fragments(&mut self) {
```
mach-kernel, line 279: This was a bit frustrating (in that it took me a while to understand what was actually happening). Why not use the same keys as `object_store` verbatim? `s3_key` vs `aws_access_key_id`? 

https://docs.rs/object_store/latest/object_store/aws/enum.AmazonS3ConfigKey.html

2c: It feels confusing that e.g. Delta Lake is `delta_lake_aws_region` here, but plain S3 is `s3_region`. And connectors have a mix of prefixed and unprefixed arguments. `aws_region` should universally represent the same thing across connectors, with a stable identifier.

crates/runtime/src/flight/actions.rs
```diff
@@ -90,17 +97,26 @@ pub(crate) async fn list() -> Response<<Service as FlightService>::ListActionsSt
             Response Message: app::App serialized as JSON bytes"
                 .into(),
     };
+    let expand_secret_action_type = FlightActionType {
```
peasee, line 100: I think this should be opt-in with a spicepod parameter, in addition to requiring TLS connections between scheduler/executor.

E.g something like

```
cluster:
  secret_expansion: enabled | disabled # default: disabled
```

We should also require that the API is secured with an API key to enable this, and validate that this route does not return any secrets unless the API key is correct. Could look something like all the schedulers/executors having the same API key definition in `runtime.auth.api_key`

mach-kernel: I agree about the API key but disagree about the toggle: you can't really use the distributed query functionality without enabling this. Most "biz" users are going to want to hit a private object store.

I am going to add an additional panic if there is no API key set and all the validations around it.

mach-kernel:
Moving back to WIP: need to make Ballista changes to pass through the API key + TLS configuration.

mach-kernel:
Synced up with @lukekim and decided to scaffold this into 2 PRs. This will go out with `--allow-insecure-connections` and the existing keyless, TLS-less RPC. 

API key + TLS is ~50% done and captured in: https://github.com/spiceai/spiceai/issues/7675

lukekim (approved):

PR #8051, opened 2025-11-13
Use 'engine: duckdb' for search tests on views

Jeadie:


Jeadie:
Superflous after #8066 is complete

PR #8057, opened 2025-11-13
Skip logging and return OK() on error during shutdown

krinart:
## 📝 Summary

We have a flaky test which relies on no error logs during shutdown. 

If error in acceleration happens during shutdown - return Ok() and skip logging

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task.rs
@@ -368,8 +368,12 @@ impl RefreshTask {
         let streaming_data_update = match get_data_update_result {
             Ok(data_update) => data_update,
             Err(e) => {
-                self.log_refresh_error(inner_err_from_retry_ref(&e), refresh.sql.as_deref())
-                    .await;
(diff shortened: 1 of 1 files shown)


crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -368,8 +368,12 @@ impl RefreshTask {
...
             Err(e) => {
-                self.log_refresh_error(inner_err_from_retry_ref(&e), refresh.sql.as_deref())
-                    .await;
+                // During runtime shutdown, refresh tasks are canceled resulting in acceleration error.
+                // This is expected and should not be logged as an error.
+                if !self.runtime_status.is_shutdown() {
+                    self.log_refresh_error(inner_err_from_retry_ref(&e), refresh.sql.as_deref())
+                        .await;
+                }
                 return Err(e);
```
lukekim, line 378: Why would we even return the error here at all? Shouldn't we just early return Ok if is shutdown?

krinart: Updated in both places for consistency

lukekim (approved):

PR #8066, opened 2025-11-14
Initial parameterisation of Search integration tests

Jeadie:
## 📝 Summary
 - New runtime `search` test module that allows for easy extension of tests:
   - To variations of a spicepod: acceleration, embedding models & vector engine.
   - To new test datasets: column configuration & definitions, and test cases.
 - For each configuration, a separate `#[tokio::test]`is run with a `spiced` spun up, and test cases ran.
 - Custom `build.rs` logic to generate independent `#[tokio::test]` at compile time. 

## 👀 Notes for Reviewers
 - Does not yet cover all tests in `crates/runtime/tests/models`.
 - Commented out test configurations will be added in subsequent PRs. 
 - Currently these tests are flagged behind `--feature extended_tests`.


crates/runtime/tests/search/mod.rs
```diff
@@ -0,0 +1,502 @@
...
+    };
+}
+
+async fn megascience_search_test_case(slug: &'static str) {
+    let mut app = AppBuilder::new(slug);
+    let cfg = SearchSpicepodConfiguration::from_str(
+        slug,
+        &MEGA_SCIENCE_COLUMN_CONFIGS,
+        &MEGA_SCIENCE_TABLES,
+    )
```
peasee: Why a complex `build.rs` with random `.yaml` files? I feel like this is very susceptible to becoming technical debt and making mistakes in defining tests vs specifying something like:

```rust
enum Accelerator {
  Arrow,
  DuckDB
}

enum Mode {
  Memory,
  File
}

let cfg = SearchSpicepodConfiguration::default()
  .with_engine(Accelerator::Arrow)
  .with_mode(Mode::Memory)
  .with_...()
```

Which you could then configure with `rstest` cases like:

`#[rstest::case(Accelerator::Arrow, Mode::Memory)]`

This is more idiomatic than parsing a string value based on the `-` hyphen delimiters to try and parse test configuration.

Some of these parameters could be reused, e.g. using the already defined `Mode` from runtime components.

peasee (changes requested):

crates/runtime/tests/search/mod.rs
```diff
@@ -0,0 +1,502 @@
...
+    .expect("could not initialise configuration");
+
+    for emb in cfg
+        .embedding_models_used(&EMBEDDING_MODEL_OPTIONS)
+        .expect("could not find embedding models")
+    {
+        app = app.with_embedding(emb);
```
peasee: If we were to use an option like my suggestion, this could become:

```rust
enum Embedding {
  OpenAI,
  Model2Vec
}

impl Embedding {
  fn to_app_embedding(&self) -> _ {
    // return the embedding to configure app
  }
}

app = app.with_embedding(emb.to_app_embedding());
```

You could also define scenarios with multiple items like `#[rstest::case(vec![Embedding::OpenAI, Embedding::Model2Vec])]`

crates/runtime/tests/search/mod.rs
```diff
@@ -0,0 +1,547 @@
...
+                Embeddings::new("model2vec:minishlab/potion-base-2M", "hf_minilm")
+            }
+        }
+    }
+}
+
+#[rstest]
+#[tokio::test]
+async fn test_megascience_permutations(
+    #[values(VectorEngineOptions::NoVectorEngine)] vector_engine: VectorEngineOptions,
```
peasee, line 236: nice!

peasee (dismissed):
Some CI failures, but the rewrite LGTM

peasee (approved):

Issue #8150, opened 2025-11-18
SpiceCloud Connector: TPCDS - Correlated scalar subquery can only be used in Projection, Filter, Aggregate plan nodes


ewgenius:
## Describe the bug

When running TPCDS benchmark, query fails with 
> sql> .error
Client specified an invalid argument. DoGet recv error: rpc error: code = InvalidArgument desc = Correlated scalar subquery can only be used in Projection, Filter, Aggregate plan nodes

On following test queries:

Q32:

```sql
select  sum(cs_ext_discount_amt)  as "excess discount amount"
from
   catalog_sales
   ,item
   ,date_dim
where
i_manufact_id = 283
and i_item_sk = cs_item_sk
and d_date between '1999-02-22' and
        (cast('1999-02-22' as date) + INTERVAL '90 days')
and d_date_sk = cs_sold_date_sk
and cs_ext_discount_amt
     > (
         select
            1.3 * avg(cs_ext_discount_amt)
         from
            catalog_sales
           ,date_dim
         where
              cs_item_sk = i_item_sk
          and d_date between '1999-02-22' and
                             (cast('1999-02-22' as date) + INTERVAL '90 days')
          and d_date_sk = cs_sold_date_sk
      )
 LIMIT 100;
```

Q92:
```sql

```

## To Reproduce

_Steps to reproduce the behavior:_

2. Click on '...'
3. Scroll down to '...'

## Expected behavior

_A clear and concise description of what you expected to happen._

### Spicepod

_Add the relevant `spicepod.yml` section._

### Output of `describe table`

_Add any relevant `describe table` output here._

### Output of `explain query`

_Add any relevant `explain query` output here._

E.g. `explain select 1`

```bash
+---------------+--------------------------------------+
| plan_type     | plan                                 |
+---------------+--------------------------------------+
| logical_plan  | Projection: Int64(1)                 |
|               |   EmptyRelation                      |
| physical_plan | ProjectionExec: expr=[1 as Int64(1)] |
|               |   PlaceholderRowExec                 |
|               |                                      |
+---------------+--------------------------------------+
```

### spice, spiced, OS info

- spice version: `spice version`
- spiced version: `spiced --version`
- OS info: `uname -a`

**Have you tried this on the latest `trunk` branch?**

- [ ] Yes
- [ ] No

**If possible, run `spiced` with DEBUG log level**
[...]

PR #8152, opened 2025-11-18
Update TPCDS overrides for spice cloud connector

ewgenius:
## 📝 Summary

This pull request updates the list of TPC-DS test queries that are skipped for the `Spicecloud` override in the test framework. The changes primarily add several queries to the exclusion list, addressing specific issues and cases where queries return empty results.

**Test query exclusions for Spicecloud:**

* Added queries `32` and `92` to the exclusion list due to issue #8150.
* Added queries `29`, `37`, `41`, `44`, `54`, and `58` to the exclusion list because they return empty results.

### DIFF
--- a/crates/test-framework/src/queries/mod.rs
+++ b/crates/test-framework/src/queries/mod.rs
@@ -723,7 +723,9 @@ pub fn get_tpcds_test_queries(overrides: Option<QueryOverrides>) -> Vec<Query> {
         Some(QueryOverrides::Spicecloud) => remove_tpcds_query!(
             queries, 8,  // https://github.com/spiceai/spiceai/issues/4668
             38, // https://github.com/spiceai/spiceai/issues/4667
-            87  // https://github.com/spiceai/spiceai/issues/4667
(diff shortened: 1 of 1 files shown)


crates/test-framework/src/queries/mod.rs
```diff
@@ -723,7 +723,11 @@ pub fn get_tpcds_test_queries(overrides: Option<QueryOverrides>) -> Vec<Query> {
         Some(QueryOverrides::Spicecloud) => remove_tpcds_query!(
             queries, 8,  // https://github.com/spiceai/spiceai/issues/4668
             38, // https://github.com/spiceai/spiceai/issues/4667
-            87  // https://github.com/spiceai/spiceai/issues/4667
+            87, // https://github.com/spiceai/spiceai/issues/4667
+            32,
+            92, // correlated scalar subquery not supported - https://github.com/spiceai/spiceai/issues/8150
+            29, 37, 41, 44, 54,
```
sgrebnov, line 729: Empty result for those queries is expected, they will be remove from zero result verification

https://github.com/spiceai/spiceai/pull/8149

PR #8211, opened 2025-11-19
feat: Support running queries only in testoperator

peasee:
## 📝 Summary

* Supports a mode of testoperator to only run queries and not control the lifetime of the spiced instance.


crates/test-framework/src/spiced/mod.rs
```diff
@@ -49,9 +49,9 @@ impl Display for SpicedVersion {
 }
 
 pub struct SpicedInstance {
```
sgrebnov: Should we update SpicedInstance usage so it is Option< SpicedInstance > instead?

Or it could be SpicedInstance with inner field of type/enum: controlled (or owned) vs uncontrolled/locally running with different behavior? Or even a trait and two different implementations

phillipleblanc (approved):

PR #8229, opened 2025-11-20
fix: Apply assertion clippy in CI/Makefile only

peasee:
## 📝 Summary

* Using these in `clippy.toml` is inconvenient because there is no way for analyzer plugins (like rust analyzer) to determine whether the lint applies to test or non test code. Move the rules instead to `.config/clippy.toml` and apply a config directory path to `lint-rust` for this path. This ensures that CI and locally run `make lint-rust` apply the assertion lint, while analyzer plugins do not.


.ci/clippy.toml
phillipleblanc, line 1: I think its a little confusing to have both `clippy.toml` and `.config/clippy.toml`

peasee: `.ci/clippy.toml`? not sure. it's annoying because you can't define `-Ddisallowed_macros` in the CLI, it must be defined in config.

phillipleblanc (approved):

PR #8235, opened 2025-11-20
[DynamoDB] Timestamp Filter Pushdown

krinart:
## 📝 Summary

Implement timestamp filter pushdown for DynamoDB, which is especially useful for `append` acceleration refresh.

Since DynamoDB doesn't have native type for datetime, typically timestamps are stored as strings. Which also means we need to know the format/timezone to format timestamps.

This PR adds:
* `time_format` parameter for DynamoDB Connector - will be used for both parsing/formatting
* Go-style datetime formatting in `utils.time_format` module
* By default iso8601 format is used
* Unit/integration tests

## 🚨 Breaking Changes
None

## 📚 Docs

https://github.com/spiceai/docs/pull/1244


crates/data_components/src/dynamodb/table_schema.rs
```diff
@@ -55,6 +58,10 @@ impl DynamoDBTableSchema {
         &self.table_schema
     }
 
+    pub fn time_format(&self) -> String {
```
phillipleblanc: Could this be an `Arc<str>` instead to avoid unnecessary string cloning?

crates/util/src/time_format.rs
```diff
@@ -0,0 +1,651 @@
+use chrono::{DateTime, FixedOffset, NaiveDateTime};
```
phillipleblanc, line 16: copyright header

phillipleblanc (dismissed):

crates/data_components/src/dynamodb/table_schema.rs
krinart: Done

sgrebnov (approved):

PR #8240, opened 2025-11-21
fix: Pushdown dynamic filters to partition scans

peasee:
## 📝 Summary

* Uses a custom `PartitionUnionExec` which attempts to push down filters to the partition children. This allows dynamic filters to be pushed down to partitioned scans (like for the Cayenne accelerator).


crates/runtime-table-partition/src/provider.rs
```diff
@@ -297,6 +311,174 @@ impl TableProvider for PartitionTableProvider {
...
+
+    fn fetch(&self) -> Option<usize> {
+        None
+    }
+
+    fn cardinality_effect(&self) -> CardinalityEffect {
+        self.inner_union.cardinality_effect()
+    }
+
+    fn try_swapping_with_projection(
```
phillipleblanc, line 441: couldn't we in theory push down projections to each child?

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #8282, opened 2025-11-25
Avoid calling S3Vector `list_vectors` (or equivalent) when indexing into `VectorIndex`s

### LINKED ISSUE #7404: Distinguish between indexing and querying table scans on search index.
## Describe the bug

For vector indexes that have all columns in the vector index (e.g. document tables), we cannot fallback to using only the vector index (at querytime), because:
1. At (initial) scan time, the index is empty.
2. indexing and querying use the same `VectorScanTableProvider::scan`. 

See:
https://github.com/spiceai/spiceai/blob/2763756f2547ef1e41fa24d4dbe09ba0c445a5d4/crates/runtime/src/embeddings/index/scan_table.rs#L215-L216

### LINKED ISSUE #7831: accelerated_refresh with embedding projections triggers expensive list_vectors API calls
## Describe the bug

When a dataset has S3 vector embeddings configured and the accelerated_refresh task includes embedding columns in its SQL query projection, it triggers expensive S3 list_vectors API calls on every refresh cycle.

It goes something like this:
- Refresh queries select all columns including embeddings e.g. `SELECT ..., col_embedding, ... FROM table`
- The presence of the embedding column in the projection causes `VectorScanTableProvider` to perform a JOIN with the vector index list
- This executes `S3VectorsListExec` which calls `list_vectors` API to pull all of the vectors out of the index on every refresh.

## To Reproduce

2. Click on '...'
3. Scroll down to '...'

## Expected behavior

### Spicepod

### Output of `explain query`

E.g. `explain select 1`

```bash
```

### spice, spiced, OS info

- OS info: `uname -a`

- [ ] Yes
- [ ] No

Jeadie:
## 📝 Summary
The core change is an `OptimizerRule` `AvoidDerivedVectorColumnOnIndexRule` which removes the need to retrieve embedding columns when indexing (i.e. `AcceleratedRefresh`) S3vector tables (i.e. `VectorScanTableProvider`). Currently, each index update requires a full scan of an S3vector index (true for a new `spiced` using an existing vector index). 

To achieve this, other core changes include:
 - Move `VectorScanTableProvider` into `search` crate.
 - Add a `derived_columns` function for `VectorIndex` with default implementation. 

## 🔗 Related
 - Resolves #7831 
 - Resolves #7404


phillipleblanc (dismissed):
Very nice!

phillipleblanc (dismissed):

phillipleblanc (approved):

Issue #7880, opened 2025-11-03
Enhancement: DynamoDB Streams (Beta)


krinart:
## Goal-State/What/Result

Add support for DynamoDB Streams in order to have a locally accelerated in-sync copy of DynamoDB table.

Requirements:
* Up to single-digit second lag
* No changes should be lost
* No traveling back in time once dataset is reported as `Ready`
* Report current lag behind the source table
* Support existing snapshotting mechanism
* Dataset size: 5GB
* Stream throughput: 100 updates/second

Out of scope (future improvements):
* Resumable table bootstrapping
* Dead letter queue
* Schema evolution

## Why/Purpose

Current use cases must choose between `full` or `append` refresh modes which are unable to provide single-digit second lag.

## By When

**Issue/Spec written and reviewed:** Nov 13, 2025
**Done-Done:** Dec 1, 2025

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks (https://github.com/spiceai/spiceai/issues/8521)
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

### Table Bootstrapping

Since DynamoDB Streams have 24 hours data retention (as such we cannot rely on having the full table history in the stream), we will need to implement table bootstrapping as a combination of table `Scan` + subscribing to the changes in DynamoDB Stream in the following order:

1. Find currently open shards, and for each shard store `StartingSequenceNumber`
2. Extract and ingest current table by using `Scan` request
3. Subscribe to the stream using `AT_SEQUENCE_NUMBER` Iterator using sequence numbers from step 1 (this way no buffering is done, and iterator expiration is not an issue)

Callouts:
* Some changes from stream will be older than the rows extracted using `Scan` request (since we are readings open shards from the beginning), which means that dataset may temporarily go backwards. This is mitigated by dataset not reporting as `Ready` until acceptable lag is achieved
[...]

PR #8283, opened 2025-11-25
Add `dynamodb-streams` crate

krinart:
## 📝 Summary

Added `dynamodb-streams` create which provides logic to consume dynamodb streams as a tokio stream.

Example usage:
```rust
   let client = Client::builder(config, table_name).build();

    // Retrieve checkpoint first    
    let checkpoint = client
        .latest_global_checkpoint()
        .await
        .expect("global checkpoint");
    
    // Do some stuff, for example bootstrap table using `Scan` request

    // Subsribe to stream using this checkpoint any time later (within 24 hours).
    let mut stream = client
        .stream_from_checkpoint(checkpoint)
        .await
        .expect("stream");
    
   while let Some(batch) = stream.next().await {
        // batch.shard_id - shard ID
        // batch.records - records
        // batch.checkpoint - checkpoint to commit progress
    }
```

Overal design is pretty simple: 
* `Client` initializes and spawns `DynamodbStreamProducer` and returns `DynamodbStream`. 
* `DynamodbStreamProducer` writes to channel which `DynamodbStream` is reading from. 
* `SDKClient` provides functionaly to make API requests to AWS.

At the heart of this library is `StreamState` which manages shard lifecycle. It does not make any API calls whicih makes it easy to test. To make sure the logic is correct, there's 40 tests for `StreamState` that cover all possible use cases.

## 🔗 Related

## 🚨 Breaking Changes

N/A

## 📚 Docs

No public API


crates/dynamodb-streams/src/checkpoint.rs
```diff
@@ -0,0 +1,39 @@
+use std::collections::{HashMap, HashSet};
```
phillipleblanc, line 16: nit: add copyright header for 2025

crates/dynamodb-streams/src/checkpoint.rs
```diff
@@ -0,0 +1,39 @@
...
+use std::time::SystemTime;
+
+#[derive(Clone, Debug)]
+pub struct ShardCheckpoint {
```
phillipleblanc, line 29: Can you add a small comment explaining what a shard checkpoint is?

crates/dynamodb-streams/src/checkpoint.rs
```diff
@@ -0,0 +1,39 @@
...
+    pub sequence_number: String,
+    pub parent_id: Option<String>, // Root shards don't have parents
+    pub updated_at: SystemTime,
+    pub position: CheckpointPosition,
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum CheckpointPosition {
```
phillipleblanc, line 41: Can you add a comment on what the difference between these two are.

crates/dynamodb-streams/src/checkpoint.rs
```diff
@@ -0,0 +1,39 @@
...
+    At,
+    After,
+}
+
+#[derive(Clone, Debug)]
+pub struct GlobalCheckpoint {
```
phillipleblanc, line 52: Add a comment that explains what a global checkpoint is

crates/dynamodb-streams/src/client.rs
```diff
@@ -0,0 +1,170 @@
+use crate::checkpoint::{CheckpointPosition, GlobalCheckpoint, ShardCheckpoint};
```
phillipleblanc, line 16: nit: add copyright header for 2025

crates/dynamodb-streams/src/client.rs
```diff
@@ -0,0 +1,170 @@
...
+            sender: tx,
+            client: Arc::clone(&self.sdk_client),
+            retry_strategy: RetryBackoffBuilder::new()
+                .method(BackoffMethod::Fibonacci)
+                .max_retries(Some(3))
+                .build(),
+        };
+
+        tokio::spawn(async move {
+            producer.streaming().await;
```
phillipleblanc, line 131: double check the behavior if the producer panics or stops streaming early

crates/dynamodb-streams/src/lib.rs
```diff
@@ -0,0 +1,31 @@
+use crate::stream_state::RecordBatch;
```
phillipleblanc: nit: add copyright header for 2025

crates/dynamodb-streams/src/stream.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub stream_arn: String,
+    pub state: StreamState,
+    pub interval: Option<Duration>,
+    pub sender: mpsc::Sender<StreamResult>,
+    pub client: Arc<SDKClient>,
+    pub retry_strategy: RetryBackoff,
+}
+
+impl DynamodbStreamProducer {
+    async fn iterate(&mut self) -> Result<Vec<RecordBatch>> {
```
phillipleblanc: maybe call this `collect` to align with what DataFusion calls their stream collector?

crates/dynamodb-streams/src/stream.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        // 4. Initialize shards that require iterators
+        self.initialize_shards_iterators().await;
+
+        Ok(batches)
+    }
+
+    fn handle_failed_shard(&mut self, shard_id: &str, error: &Error) {
+        // TODO: Finalize logic
```
phillipleblanc:
```suggestion
```

Create an issue to track any outstanding work

crates/dynamodb-streams/src/stream.rs
```diff
@@ -0,0 +1,169 @@
...
+                )
+                .await
+            {
+                Ok(iterator) => {
+                    if let Some(iterator) = iterator {
+                        self.state.mark_active(shard_id, iterator);
+                    }
+                }
+                Err(e) => {
+                    tracing::warn!("Failed to initialize shard {}: {}", shard_id, e);
```
phillipleblanc, line 123: This shouldn't return an error?

crates/dynamodb-streams/src/stream.rs
```diff
@@ -0,0 +1,169 @@
...
+                    } else {
+                        tracing::error!("Iteration failed after exhausting retries: {}", e);
+                        return Err(e);
+                    }
+                }
+            }
+        }
+    }
+
+    pub async fn streaming(mut self) {
```
phillipleblanc, line 148: This isn't really streaming? It still collects the entire shard into memory before sending the batches, for a true streaming we need a way to yield batches.

crates/dynamodb-streams/src/stream_state.rs
```diff
@@ -0,0 +1,1963 @@
+use crate::checkpoint::{CheckpointPosition, GlobalCheckpoint, ShardCheckpoint};
```
phillipleblanc, line 16: nit: add copyright header for 2025

crates/dynamodb-streams/src/stream_state.rs
```diff
@@ -0,0 +1,1963 @@
...
+use crate::client_sdk::{ApiShard, SDKClient};
+use aws_sdk_dynamodbstreams::types::{Record, ShardIteratorType};
+use std::collections::HashMap;
+use std::sync::Arc;
+use std::time::SystemTime;
+
+#[derive(Debug)]
+pub struct RecordBatch {
```
phillipleblanc: Given this is the same name as the Arrow struct, I'd avoid using this name

crates/dynamodb-streams/src/stream.rs
```diff
@@ -0,0 +1,169 @@
...
+            }
+        }
+
+        // 3. Discover new shards
+        if let Ok(shards) = self.client.get_all_shards(&self.stream_arn).await {
+            self.state.add_discovered(shards);
+        }
+
+        // 4. Initialize shards that require iterators
+        self.initialize_shards_iterators().await;
```
Jeadie: If we fail at 3, then the next time we call `.iterate()`, will we need to  `initialize_shards_iterators` before we can do `1. Poll active shards`?

crates/dynamodb-streams/src/stream.rs
```diff
@@ -0,0 +1,169 @@
...
+                .client
+                .get_shard_iterator(
+                    &self.stream_arn,
+                    &shard_id,
+                    &ShardIteratorType::TrimHorizon,
+                    None,
+                )
+                .await
+            {
+                Ok(iterator) => {
```
Jeadie, line 117:
```suggestion
                Ok(Some(iterator)) => {
```

crates/dynamodb-streams/src/stream.rs
Jeadie: The affect of missing step 4 (if 3 fails) is that initializing shards do not get 
```rust
self.state.mark_active(shard_id, iterator);
```

crates/dynamodb-streams/src/stream.rs
Jeadie: It's streaming in the sense that it gets all data from active shards, and prepares any newly created shards for the next function call. It doesn't collect the entire shard since the call to `get_records` is from the previous offset and will only return records up to what's available.

But it would be nice to have a more stream-like `get_records` AWS SDK method.

crates/dynamodb-streams/src/stream_state.rs
```diff
@@ -0,0 +1,1963 @@
...
+    /// Add discovered shards, returns shard IDs that need initialization
+    pub fn add_discovered(&mut self, shards: Vec<ApiShard>) {
+        for shard in shards {
+            let shard_id = shard.shard_id.clone();
+
+            if self.active.contains_key(&shard_id)
+                || self.pending.contains_key(&shard_id)
+                || self.initializing.contains_key(&shard_id)
+            {
+                continue;
```
Jeadie, line 159: This is an early exit if ApiShard exists?

crates/dynamodb-streams/src/stream_state.rs
```diff
@@ -0,0 +1,1963 @@
...
+            }
+
+            let blocked = shard.parent_shard_id.clone().is_some_and(|p| {
```
Jeadie, line 166: Worth explaining what/how a shard becomes `blocked` from DynamoDB perspective

crates/dynamodb-streams/src/stream_state.rs
```diff
@@ -0,0 +1,1963 @@
...
+
+#[derive(Debug)]
+pub struct RecordBatch {
+    pub shard_id: String,
+    pub records: Vec<Record>,
+    pub checkpoint: ShardCheckpoint,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct ActiveShard {
```
Jeadie, line 30: Might be cleaner to do 
```rust
pub enum ShardState {
    Active {
        pub shard_id: String,
        pub parent_shard_id: Option<String>,
        pub iterator: String,
    },
    Pending {
        pub shard_id: String,
        pub parent_shard_id: Option<String>,
    },
    Initializing {
        pub shard_id: String,
        pub parent_shard_id: Option<String>,
    },
}
```
That way we don't need three HashMaps that we remove and reinsert when state transitions occur.

Jeadie: e.g. `StreamState::reinitialize_shard`, `StreamState::mark_active`

crates/dynamodb-streams/src/stream_state.rs
```diff
@@ -0,0 +1,1963 @@
...
+        ApiShard {
+            shard_id: id.to_string(),
+            parent_shard_id: parent.map(std::string::ToString::to_string),
+            starting_sequence_number: None,
+            ending_sequence_number: ending_seq.map(std::string::ToString::to_string),
+        }
+    }
+
+    // ========================================
+    // Tests for handle_poll_result()
```
Jeadie: nit: use submodules
```rust
mod tests {
  mod handle_poll_result {
    #[test]
    fn test_shard_not_found() {

```

crates/dynamodb-streams/src/stream_state.rs
```diff
@@ -0,0 +1,1963 @@
...
+
+        let result = state.handle_poll_result(
+            "shard-1",
+            Some("new-iter".to_string()),
+            vec![create_record("123")],
+        );
+
+        assert!(result.is_some());
+        let batch = result.expect("Expected batch to be Some");
+        assert_eq!(batch.shard_id, "shard-1");
```
Jeadie: nit: `insta::snapshot` might be cleaner. You can do inline snapshots for small snapshots too. e.g.

https://github.com/spiceai/spiceai/blob/287ad81296caef06f138dbbd8103ff4793a8f1db/crates/runtime/src/search/request.rs#L331-L334

Jeadie (commented):
This is very nice :)

crates/dynamodb-streams/src/stream_state.rs
krinart: Correct. It means shard we already know about was returned.

crates/dynamodb-streams/src/stream_state.rs
krinart: Let's do it later :)

crates/dynamodb-streams/src/stream_state.rs
krinart: Let's do it after demo

crates/dynamodb-streams/src/stream.rs
krinart: I would need to add another arm for `None`, which in the current implementation I can simply ignore as it means shard has been exhausted

crates/dynamodb-streams/src/stream.rs
krinart: Added TODO and a link to issue

crates/dynamodb-streams/src/client.rs
krinart: Added TODO and link to the issue

krinart:
I changed logic to return a stream of `DynamoDBStreamBatch` which contains records from all shards and a global checkpoint.

As to the channel, I would leave it since it allows for parallelization between producer and consumer. If we want to have a lazy stream, we can set channel capacity to 1 using `.buffer()` method.

Having a truly lazy stream would mean producer is idle until it's polled by the consumer.

Jeadie (dismissed):

crates/dynamodb-streams/src/client.rs
```diff
@@ -112,6 +127,8 @@ impl Client {
         };
 
         tokio::spawn(async move {
+            // TODO - double check the behavior if the producer panics or stops streaming early
+            // https://github.com/spiceai/spiceai/issues/8074
```
phillipleblanc:
```suggestion
```

Don't include TODO comments in the source code, they have a tendency to not get cleaned up. Let's just track with the issue.

crates/dynamodb-streams/src/stream.rs
```diff
@@ -104,17 +121,18 @@ impl DynamodbStreamProducer {
                     }
                 }
                 Err(e) => {
+                    // TODO: Handle failures - https://github.com/spiceai/spiceai/issues/8074
```
phillipleblanc:
```suggestion
```

Same as above, don't include TODO in code

phillipleblanc (dismissed):

Jeadie (approved):

PR #8285, opened 2025-11-25
Update threat model to v1.9.1

phillipleblanc:
## Summary
- add Cayenne accelerator, MCP endpoint, and Iceberg write support to threat model
- add new STRIDE threats covering Iceberg writes, cluster auth, LLM prompt injection, telemetry leakage, and model gateway keys

image

### DIFF
--- a/docs/threat_models/v1.9.1.json
+++ b/docs/threat_models/v1.9.1.json
@@ -1,7 +1,7 @@
 {
   "version": "2.2.0",
   "summary": {
-    "title": "Spice.ai OSS v1.10",
+    "title": "Spice.ai OSS v1.9.1",
     "owner": "Phillip LeBlanc",
     "description": "Spice is a portable, open-source runtime for fast, last-mile SQL query, search, and AI inference.",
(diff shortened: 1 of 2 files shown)


docs/threat_models/v1.9.1.json
```diff
@@ -0,0 +1,1392 @@
...
+                "strokeDasharray": null
+              }
+            },
+            "visible": true,
+            "shape": "process",
+            "zIndex": 19,
+            "id": "0f2b61f2-3f68-46e3-a9e8-0646b8d9c071",
+            "data": {
+              "type": "tm.Process",
+              "name": "Ballista Scheduler",
```
Jeadie, line 1218: Could Ballista scheduler be over an `untrusted network` too?

phillipleblanc: It could be yes, but we wouldn't recommend that deployment until we've implemented tighter cross-cluster comms

ewgenius (approved):

PR #8295, opened 2025-11-26
Pass IndexedTableProvider down in 'changes_stream' and 'append_stream'

Jeadie:
## 📝 Summary
 - Issue when FTS and S3vector index on same table for `refresh_mode: changes`. FTS was removing the `IndexedTableProvider` expected by `EmbeddingConnector`.

### DIFF
--- a/crates/runtime/src/embeddings/connector.rs
+++ b/crates/runtime/src/embeddings/connector.rs
@@ -23,6 +23,7 @@ use crate::dataconnector::{DataConnector, DataConnectorError, DataConnectorResul
 use crate::embeddings::execution_plan::{
     compute_additional_embedding_columns, construct_record_batch,
 };
+use crate::embeddings::index::VectorScanTableProvider;
 use crate::embeddings::index::table::wrap_table_as_index;
 use crate::federated_table::FederatedTable;
 use crate::model::ENABLE_MODEL_SUPPORT_MESSAGE;
@@ -34,6 +35,7 @@ use datafusion::datasource::TableProvider;
 use futures::StreamExt;
 use itertools::Itertools;
 use runtime_datafusion_index::IndexedTableProvider;
+use search::generation::text_search::index::FullTextDatabaseIndex;
 use spicepod::component::embeddings::ColumnEmbeddingConfig;
 
 use std::any::Any;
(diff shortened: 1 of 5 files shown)


Jeadie:
- [x] TODO: recheck with FTS and `refresh_mode: changes`, no vector engine.

crates/runtime/src/embeddings/index/table.rs
```diff
@@ -104,7 +104,14 @@ async fn wrap_table_as_index_s3(
                 .map(|embed| (c.name.clone(), embed.clone()))
         })
         .collect();
-    let mut provider = IndexedTableProvider::new(Arc::clone(&inner_table_provider));
+    let mut provider = if let Some(indexed) = inner_table_provider
```
Jeadie, line 107: This ended up being unnecessary, but it does future-proof if another connector wraps `EmbeddingConnector` and also makes an `IndexedTableProvider` (or if we swap the order of the `FullTextConnector`).

crates/runtime/src/embeddings/connector.rs
```diff
@@ -262,24 +273,34 @@ impl DataConnector for EmbeddingConnector {
...
-
-        let stream = self
-            .inner_connector
-            .changes_stream(underlying_federated_table)?
-            .then(move |item| Self::embed_change_envelope(item, Arc::clone(&embedding_table)))
-            .boxed();
-
-        Some(stream)
+        // `VectorScanTableProvider` is generally wrapped by a `IndexedTableProvider` (as above), but in the case both [`Self`] and the [`FullTextConnector`] exist, the latter will unwrap the `IndexedTableProvider` first. It will correctly handle indexing vector indexes as that point.
+        } else if let Some(vector_scan) = table_provider
```
phillipleblanc, line 278: This logic is pretty ugly - not your fault, its a consequence of the current design. I think its fine for now, but we should have a think about what a better way to do this would be.

phillipleblanc: i.e. the embedding connector shouldn't have to know about a vector scan provider

phillipleblanc (dismissed):

crates/runtime/src/embeddings/connector.rs
Jeadie: I agree, this is partially true with the `EmbeddingTable` too

krinart (approved):

PR #8306, opened 2025-11-26
Return computed embeddings from index for partitioned S3Vectors

Jeadie:
## 📝 Summary
**Fix for the following bug**
With spicepod
```yaml
datasets:
  - from: file:./data/mega-science-small.jsonl
    name: mega_science_ds
    acceleration:
      enabled: true
    vectors:
      enabled: true
      engine: s3_vectors
      partition_by:
        - bucket(3, id)
      params:
        s3_vectors_bucket: team-app-journals-jeadie2
    columns:
      - name: question
        embeddings:
          - from: potion_base_8m
            row_id: id
```
Run
```sql
select id, question_embedding from mega_science_ds limit 10;
```
```
+----+--------------------+
| id | question_embedding |
+----+--------------------+
| 6  |                    |
| 15 |                    |
| 23 |                    |
| 27 |                    |
| 28 |                    |
| 36 |                    |
| 45 |                    |
| 63 |                    |
| 64 |                    |
| 65 |                    |
+----+--------------------+

Time: 0.007326209 seconds. 10 rows.
```

Note: This did not affect vector_search i.e.
```sql
--- non-null `question_embedding`
select id, question_embedding from vector_search(mega_science_ds, 'hello') limit 10;
```

### DIFF
--- a/crates/search/src/index/s3_vectors/mod.rs
+++ b/crates/search/src/index/s3_vectors/mod.rs
@@ -17,6 +17,7 @@ limitations under the License.
 use std::{any::Any, sync::Arc};
 
 use arrow::array::RecordBatch;
+use arrow::compute::concat_batches;
 use arrow_schema::{DataType, Field};
 use async_trait::async_trait;
 use data_components::s3_vectors::compute_query::{CachedQueryVector, ComputeQueryVector};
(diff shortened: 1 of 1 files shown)


crates/search/src/index/s3_vectors/mod.rs
```diff
@@ -176,19 +187,21 @@ impl SearchIndex for S3Vector {
...
-                return write::write(self, &self.table, record, self.batch_write_rows)
-                    .await
-                    .boxed();
-            }
+            None => write::write(self, &self.table, record, self.batch_write_rows)
+                .await
+                .boxed(),
         }
-
-        Ok(record)
```
Jeadie, line 191: This was the source of the issue. Unless we early exited (like the non-partitioned case), we returned the input data (which does not have the calculated embeddings). The output of this function is what gets written to an accelerator (and thus with `*_embedding` as null).

mach-kernel (dismissed):

krinart (approved):

PR #8345, opened 2025-11-29
Persistent checkpoints for DynamoDB Streams

### LINKED ISSUE #8032: Support stream checkpoints: commit + restore from snapshots
Important:
* Find most recent shards (mark closed shards?)
* Prune old shards?

krinart:
## 📝 Summary

1. `DynamoDBSys` implementation with support for postgres, turso, duckdb and sqlite
2. On start check if there's an existing checkpoint, if yes - use it
3. Commit checkpoint after each batch is processed

## 🔗 Related

Closes https://github.com/spiceai/spiceai/issues/8032

## 🚨 Breaking Changes

None

## 📚 Docs


crates/runtime/src/dataconnector/dynamodb.rs
```diff
@@ -247,58 +256,208 @@ impl DataConnector for DynamoDB {
...
                             })
-                            .filter_map(|opt| async move { opt })
-                            .flatten()
-                        )
+                        });
+
+                    // Attach changes stream from initial checkpoint to bootstrap stream
+                    Some(
+                        bootstrap_stream
+                            .chain(
```
sgrebnov, line 280: Can we move checkpoint first save to chain so we have the same logic further and don't need additional case and param

sgrebnov (dismissed):

crates/runtime/src/dataconnector/dynamodb.rs
krinart: Will do in a follow up PR. Thanks!

sgrebnov (approved):

PR #8359, opened 2025-12-01
Batching for CDC

### LINKED ISSUE #7100: Enhancement: CDC Streams process changes in batches

krinart:
## 📝 Summary

1. Divide each ChangeBatch into smaller sub-batches by creating a new sub-batch whenever:
  * An upsert operation is followed by a delete operation
  * A delete operation is followed by an upsert operation
  * The primary key of the next operation already exists in the current sub-batch
2. Tests

Future work: 

## 🔗 Related

Closes https://github.com/spiceai/spiceai/issues/7100

## 🚨 Breaking Changes

N/A

## 📚 Docs

No publicly visible changes


crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -154,134 +158,279 @@ impl RefreshTask {
...
+    if change_batch.record.num_rows() == 0 {
+        return vec![];
+    }
+
+    let mut sub_batches = Vec::new();
+    let mut current_batch_indices = Vec::new();
+    let mut current_op_type: Option<ChangeOperationType> = None;
+    let mut seen_primary_keys: HashSet<String> = HashSet::new();
+
+    for row_id in 0..change_batch.record.num_rows() {
```
sgrebnov, line 370: Please double check if we use optimal approach relying on metadata/indices instead of cloning actual data

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -154,134 +158,279 @@ impl RefreshTask {
...
+    {
+        sub_batches.push((op_type, current_batch_indices));
+    }
+
+    sub_batches
+}
+
+// Used to group batch changes into sub-batches
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ChangeOperationType {
```
sgrebnov, line 418: Please double check additional type is required and we can't use `ChangeOperation` + own equality check function

sgrebnov (approved):

crates/runtime/src/accelerated_table/refresh_task/changes.rs
krinart: Will do in a follow up PR. Thanks!

PR #8374, opened 2025-12-02
Add v1.9.2 threat model

mach-kernel:
## 📝 Summary

This attempts to add some things that were not previously represented in the 1.9x threat model.

#### High level components
- HuggingFace as an external store
- Local model storage as an explicit store
- Model execution as an explicit process with mention of libraries we use

#### Data flows
- Model I/O (from HF, to disk, etc)
- Ballista I/O (shuffles, RPC)
- Catalog I/O

#### Threats
- Model inference DoS
- Model data poisoning (either via DNS hijacking of HF, or local model hijacking -- generally "supply chain" attack surface)
- Shuffle data poisoning and interception

image

### DIFF
--- a/docs/threat_models/v1.9.2.json
+++ b/docs/threat_models/v1.9.2.json
@@ -1,7 +1,7 @@
 {
   "version": "2.5.0",
   "summary": {
-    "title": "Spice.ai OSS v1.10.0",
+    "title": "Spice.ai OSS v1.9.2 Threat Model",
     "owner": "Phillip LeBlanc",
     "description": "Spice is a portable, open-source runtime for fast, last-mile SQL query, search, and AI inference with embedded model execution.",
     "id": 0
@@ -469,8 +469,8 @@
             "connector": "smooth",
             "data": {
               "type": "tm.Flow",
-              "name": "Accelerator I/O (DuckDB/SQLite/Vortex)",
-              "description": "File formats: .duckdb, .db, .vortex, Parquet, Arrow IPC",
+              "name": "Accelerator I/O (DuckDB)",
+              "description": "File formats: .duckdb, Parquet, Arrow IPC",
               "outOfScope": false,
               "isTrustBoundary": false,
               "reasonOutOfScope": "",
@@ -482,7 +482,7 @@
               "threats": []
             },
             "labels": [
-              "Accelerator I/O (DuckDB/SQLite/Vortex)"
+              "Accelerator I/O (DuckDB)"
             ],
             "id": "25b3a98a-0e28-4d2b-bf74-bede05ce95eb",
             "source": {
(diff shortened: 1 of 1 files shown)


docs/threat_models/v1.9.2.json
```diff
@@ -0,0 +1,2947 @@
+{
+  "version": "2.5.0",
+  "summary": {
+    "title": "Spice.ai OSS v1.10.0",
```
phillipleblanc:
```suggestion
    "title": "Spice.ai OSS v1.9.2",
```

docs/threat_models/v1.9.2.json
phillipleblanc, line 1: I propose that we switch to a markdown textual threat model with the following content:

THREAT MODEL

Any entity that can send requests to a Spice instance is a potential threat actor. The main risk is Spice doing *more* than the runtime explicitly configured it to do.

In our model, the `spicepod.yaml` is the **Root of Trust**. The runtime's job is to faithfully execute that configuration and enforce the boundaries it defines. If the runtime fails to enforce those boundaries, that is a vulnerability in Spice.

Since Spice acts as a middleware or "sidecar," the threat actors are primarily defined by their network position relative to Spice.

- **The "Lateral" Attacker**: An attacker who has compromised a different service in the same Kubernetes cluster or VPC. They have network access to Spice’s ports but no filesystem access. They want to use Spice as a proxy to reach upstream databases (Postgres/MySQL) that are firewalled off from the rest of the network, or to steal the credentials Spice holds in memory. Without an API key, they should not be able to perform any actions with Spice or leak any information.
- **The Malicious Client**: A valid user or service authorized to query Spice (has an API key), but wants to do more than allowed. They are trying to escalate their privilege beyond what they are supposed to be able to do in Spice. They have access to `Dataset A` but are crafting SQL inputs to try and read `Dataset B` (which exists in the source but not exposed to them) or write data. This is probably our most important threat actor, they must be tightly constrained by the configuration.

Spice is responsible for protecting:
- **Upstream credentials**: The connection strings, passwords, or keys Spice uses to talk to backend databases (e.g., Postgres, Dremio, Snowflake).
- **Upstream Data Integrity**: The data sitting in the backend. (Spice must not accidentally allow deletion or modification).
- **Data Confidentiality (Scope)**: Ensuring data from `Table A` is not leaked when querying `Table B`.

The threat surface is the **Network API** (HTTP, Flight, OpenTelemetry) and the **SQL Engine Logic**.

The main vectors for exploitation are:
- **Isolation breakout ("table hopping")**: An attacker sends a query for a valid dataset, but uses SQL injection or engine flaws to reference a dataset not exposed in the `spicepod.yaml`, or uses a "Pass-through" vulnerability to execute commands directly on the backend.
- **Read-Only Bypass (Integrity Violation)**: An attacker constructs a query that results in a state change in the backend database.
- **Authentication Bypass (API Surface)**: If `spicepod.yaml` mandates an API Key (`api_key_auth`), the runtime must enforce it globally.
- **Information Leakage via Errors**: Spice encounters an error (e.g., connection failure to Postgres) and returns the raw error message to the client. Example: User runs a bad query. Spice responds: `Error connecting to postgres://admin:SuperSecretPassword@10.0.0.5....`
- **Function-Based SSRF (Server-Side Request Forgery)**: Even if the config is static, SQL engines often support functions that reach out to the network (e.g., `http_get`, `load_extension`). An attacker submits a SQL query that uses a built-in function to make Spice initiate a network connection to an internal server.

**Trust assumptions:**
- The host/K8s environment and `spicepod.yaml` are controlled by a trusted operator.
- Attackers **do not** have direct filesystem or ConfigMap edit access. **An attacker with filesystem access to modify `spicepod.yaml` is already considered to have full control of the environment.** If you can edit files on the Spice host, you almost certainly have network-level access as well. In Kubernetes, modifying the ConfigMap that supplies `spicepod.yaml` requires elevated permissions to the K8s API, which likewise implies full control over the deployment environment.
- Denial-of-service via expensive queries is treated as “expected database behavior” and handled by infra controls, not Spice itself.

Jeadie: `Information Leakage via Errors: Spice`: This also would include leakage via metrics APIs or Otel collectors (and would be on diagram if exist).

phillipleblanc: We could use a mermaid diagram for a visual, but I think the text format conveys the information it needs to.

phillipleblanc:
Closing this in favor of https://github.com/spiceai/spiceai/pull/8400

PR #8390, opened 2025-12-03
New crate `google-genai` 

Jeadie:
## 📝 Summary
 - Generic Gemini AI API wrapper. No integration with `crates/llms` or spicepod LLMs/embeddings.


crates/google-genai/examples/embeddings.rs
```diff
@@ -0,0 +1,56 @@
...
+    ];
+
+    println!("Generating embeddings for {} texts...\n", texts.len());
+
+    for (i, text) in texts.iter().enumerate() {
+        let request = EmbedContentRequest::new(Content::user(*text))
+            .with_output_dimensionality(768)
+            .with_task_type(TaskType::RetrievalQuery);
+
+        let response = client.embed_content("text-embedding-004", request).await?;
```
phillipleblanc, line 45: probably also useful to time how long this takes

crates/google-genai/src/client.rs
```diff
@@ -0,0 +1,75 @@
...
+*/
+
+//! Google Generative AI API client
+#![allow(clippy::missing_errors_doc)]
+
+use crate::error::{Error, HttpSnafu, Result};
+use reqwest::header::{CONTENT_TYPE, HeaderMap, HeaderValue};
+use snafu::ResultExt;
+
+const BASE_URL: &str = "https://generativelanguage.googleapis.com/v1beta";
```
phillipleblanc, line 24: They only have a v1beta API?

crates/google-genai/src/client.rs
```diff
@@ -0,0 +1,75 @@
...
+const API_KEY_HEADER: &str = "x-goog-api-key";
```
phillipleblanc, line 25: looks like google doesn't pay attention to RFC standards :)

crates/google-genai/src/embeddings.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use crate::client::Client;
+use crate::error::{HttpSnafu, Result};
+use crate::types::Content;
+use reqwest::header::HeaderMap;
+use serde::{Deserialize, Serialize};
+use snafu::ResultExt;
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
```
phillipleblanc, line 27: why  camelCase here?

phillipleblanc: ah, is that what the Google API expects?

phillipleblanc (dismissed):

crates/google-genai/examples/embeddings.rs
Jeadie: They will be traced in  `impl Chat` and `impl Embed`

crates/google-genai/src/client.rs
Jeadie: I think they might have a gRPC at v1, but their documentation is on `v1beta`. E.g: https://ai.google.dev/api/generate-content#method:-models.generatecontent

crates/google-genai/src/embeddings.rs
Jeadie: Yes

phillipleblanc (approved):

PR #8406, opened 2025-12-04
SWR caching: serve rotten data instead of error

mach-kernel:
## 📝 Summary

If we have rotten data, serve that to the user instead of surfacing an error when trying to `fetch_from_source`. Note that this does the fetch _first_, and only surfaces data or failure _after_. So the user is responsible for configuring how quickly they time out + how many times to retry (i.e. if they want a speedy response in the failure case).

## 👀 UX considerations

- Errors let you know something is wrong. Do we want a parameter for this behavior?

#### Reproduce

```python
from fastapi import FastAPI
from datetime import datetime
import uvicorn

app = FastAPI()

@app.get("/")
async def root():
    return {"now": datetime.now()}

if __name__ == "__main__":
    uvicorn.run(
        "server:app",
        host="0.0.0.0",
        port=8000,
        reload=True
    )
```

```yaml
version: v1beta1
kind: Spicepod
name: swr-test
runtime:
  caching:
    sql_results:
      enabled: false

datasets:
  - from: http://localhost:8000/
    name: test_data
    params:
      allowed_request_paths: '/*'
      client_timeout: 1
      connect_timeout: 1
      max_retries: 1
    acceleration:
      enabled: true
      refresh_mode: caching
      engine: duckdb
      mode: memory
      refresh_check_interval: 5s
      params:
        duckdb_file: swr_test.db
        caching_ttl: 5s
        caching_stale_while_revalidate_ttl: 30s
      refresh_sql: SELECT * FROM test_data
```

### DIFF
--- a/crates/runtime/src/accelerated_table/caching.rs
+++ b/crates/runtime/src/accelerated_table/caching.rs
@@ -604,6 +604,7 @@ impl CacheRefreshHelper {
         fallback_schema: SchemaRef,
         is_rotten: bool,
         accelerator_mutex: Arc<Mutex<()>>,
+        cached_batches: Option<Vec<RecordBatch>>,
     ) -> SendableRecordBatchStream {
         match Self::fetch_from_source(&federated, dataset_name, filters, limit).await {
             Ok(batches) if !batches.is_empty() => {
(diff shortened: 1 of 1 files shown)


sgrebnov (approved):

crates/runtime/src/accelerated_table/caching.rs
```diff
@@ -668,11 +669,27 @@ impl CacheRefreshHelper {
...
                     e
                 );
-                let error_stream = RecordBatchStreamAdapter::new(
-                    fallback_schema,
-                    futures::stream::once(async move { Err(e) }),
-                );
-                Box::pin(error_stream)
+
+                match cached_batches {
+                    Some(rotten_batches) if is_rotten && !rotten_batches.is_empty() => {
```
sgrebnov, line 674: `rotten_batches.is_empty()` - can there be the case that empty cached batches is actual valid response (no results) and we should return it?

crates/runtime/src/accelerated_table/caching.rs
```diff
@@ -668,11 +669,27 @@ impl CacheRefreshHelper {
                     dataset_name,
                     e
                 );
-                let error_stream = RecordBatchStreamAdapter::new(
-                    fallback_schema,
-                    futures::stream::once(async move { Err(e) }),
-                );
-                Box::pin(error_stream)
+
+                match cached_batches {
```
phillipleblanc, line 673: we shouldn't do this, if the items are expired/rotten then they don't logically exist and we should return an error.

PR #8407, opened 2025-12-04
Periodic snapshots for append/changes streams

### LINKED ISSUE #8034: [DDB Streams] Support for periodic snapshotting

krinart:
## 📝 Summary

1. Added support for `snapshots_trigger_batches` param for `append`/`changes` streams:
```yaml
datasets:
  - from: dynamodb:viktor_test_streams_e2e
    name: viktor_test_streams_e2e
    acceleration:
      enabled: true
      refresh_mode: changes
      snapshots: enabled
      params:
        snapshots_trigger_batches: X
```

2. Validation with errors

3. After processing X batches, snapshot will be created

## 🔗 Related

Closes https://github.com/spiceai/spiceai/issues/8034

## 🚨 Breaking Changes

N/A
## 📚 Docs


crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -468,6 +468,7 @@ pub struct Refresher {
     synchronize_with: Option<SynchronizedTable>,
     snapshot_behavior: SnapshotBehavior,
     snapshot_local_path: Option<PathBuf>,
+    snapshot_trigger_batches: Option<u8>,
```
sgrebnov, line 471: `snapshot_trigger_num_batches` might be more easy to understand and read?

sgrebnov: I see that `snapshot_trigger_batches` is actual UX so this should match

crates/runtime/src/accelerated_table/mod.rs
```diff
@@ -288,6 +288,7 @@ pub struct Builder {
     initial_load_complete: bool,
     snapshot_behavior: SnapshotBehavior,
     snapshot_local_path: Option<PathBuf>,
+    snapshot_trigger_batches: Option<u8>,
```
sgrebnov, line 291: `batches` - is it DynamoDB term? Otherwise it will be hard to specify for the user.  Assume this is not Arrow Batch as users won't be able to operate by it / don't know what is this. Might be `rows/records/num_changes`?

sgrebnov (approved):

PR #8408, opened 2025-12-04
Watermarks support for DynamoDB

krinart:


Cargo.toml
```diff
@@ -168,6 +168,7 @@ graphql-parser = "0.4.0"
 hf-hub = { version = "0.4", features = ["tokio"] }
 http = "1.3.1"
 httpdate = "1.0"
+humantime = "2.1.0"
```
krinart, line 171: Not a new dependency, moved from `crates/utils`

PR #8417, opened 2025-12-05
Watermarks support for DynamoDB

### LINKED ISSUE #8031: [DDB Streams] Support for watermarks + lag tracking

krinart:
## 📝 Summary

1. Watermark calculation in `dynamodb-streams:
  * Each shard has: 
      * `last_produced_at: Option<SystemTime>` # time when it emitted message(s) last time, until then `None`
      * `current_watermark: Option<SystemTime>` # current max event time seen in this shard
  * When building a batch:
      * Find non-idle shards: (`now() - last_produced_at < idle_timeout`)
      * Find `min(current_watermark)` for non-idle shards
2. Added `is_dataset_ready` to `ChangeEnvelope`
  * Always `true` for kafka/debezium
  * For DynamoDB hardcoded as `10 seconds` (Will update in the follow up PR based on UX review)

## 🔗 Related

Closes https://github.com/spiceai/spiceai/issues/8031

## 📚 Docs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -168,6 +168,7 @@ graphql-parser = "0.4.0"
 hf-hub = { version = "0.4", features = ["tokio"] }
 http = "1.3.1"
 httpdate = "1.0"
+humantime = "2.1.0"
 iceberg = "0.7.0"
 iceberg-catalog-glue = "0.7.0"
 iceberg-catalog-rest = "0.7.0"
--- a/crates/data_components/src/cdc.rs
+++ b/crates/data_components/src/cdc.rs
@@ -91,7 +91,11 @@ pub struct ChangeEnvelope {
 
 impl ChangeEnvelope {
     #[must_use]
-    pub fn new(change_committer: Box<dyn CommitChange + Send>, change_batch: ChangeBatch, is_dataset_ready: bool) -> Self {
+    pub fn new(
+        change_committer: Box<dyn CommitChange + Send>,
+        change_batch: ChangeBatch,
+        is_dataset_ready: bool,
+    ) -> Self {
         Self {
             change_committer,
             change_batch,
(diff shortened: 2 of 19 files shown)


Cargo.toml
```diff
@@ -168,6 +168,7 @@ graphql-parser = "0.4.0"
 hf-hub = { version = "0.4", features = ["tokio"] }
 http = "1.3.1"
 httpdate = "1.0"
+humantime = "2.1.0"
```
krinart, line 171: Not a new dependency, moved from `crates/utils`

crates/data_components/src/cdc.rs
```diff
@@ -86,14 +86,20 @@ pub trait CommitChange {
 pub struct ChangeEnvelope {
     change_committer: Box<dyn CommitChange + Send>,
     pub change_batch: ChangeBatch,
+    is_dataset_ready: bool,
```
sgrebnov, line 89: Can this be oneshot channel or similar passed separately (second argument) so we don't need to modify ChangeEnvelope format? If we wrap this into some struct it could be easily extended with additional control/system signals/etc

crates/dynamodb-streams/src/stream.rs
```diff
@@ -163,20 +162,96 @@ impl DynamodbStreamProducer {
...
+            shard_result.last_checkpoint.clone(),
+        );
+
+        // Check eligibility for watermark
+        let is_eligible = match shard_result.outcome {
+            // Shards that produced records and those that failed are always eligible
+            PollOutcome::Records { .. } | PollOutcome::Failed => true,
+
+            // Shards that produced no records are NOT eligible if they have been idle for longer than the idle_timeout
+            PollOutcome::Empty => {
```
sgrebnov, line 197: Nit: review if PollOutcome::Empty actually means this is shard end so we don't need 'idle timeout' logic

sgrebnov (approved):

crates/dynamodb-streams/src/stream.rs
krinart: Will look into it, thanks!

PR #8424, opened 2025-12-06
DynamoDB Stream Follow Up

krinart:
## 📝 Summary

1. Removed shards that returned no records from watermark calculation
2. Moved persisting bootstrap checkpoint into a separate step
3. Added metrics
4. Fixed a bug of not skipping shards that are ancestors of active shards during discovery
5. Set default size of channel buffer to 1
6. Made default polling interval 0 instead of 200 (waiting is done by channel blocking)

## 🔗 Related

## 📚 Docs

### DIFF
--- a/crates/data_components/src/dynamodb/provider.rs
+++ b/crates/data_components/src/dynamodb/provider.rs
@@ -64,7 +64,6 @@ use std::collections::HashSet;
 use std::pin::Pin;
 use std::time::{Duration, SystemTime};
 use std::{any::Any, collections::HashMap, fmt, sync::Arc};
-use std::num::NonZeroUsize;
 
 #[derive(Debug, Clone)]
 pub struct DynamoDBTableProvider {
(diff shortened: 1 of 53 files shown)


crates/data_components/src/dynamodb/provider.rs
```diff
@@ -91,12 +99,17 @@ impl DynamoDBTableProvider {
         config_partitions: Option<usize>,
         stream_poll_interval_ms: u64,
         time_format: String,
+        acceptable_lag: Duration,
+        metrics_collector: Arc<MetricsCollector>,
     ) -> Result<Self, Error> {
         let db_client = Arc::new(DbClient::new(&sdk_config));
+        #[expect(unsafe_code)]
+        let buffer_size = unsafe { NonZeroUsize::new_unchecked(1) };
```
phillipleblanc: Instead of using unsafe, just use the normal safe construction and use the unreachable! macro to handle the unwrapping

PR #8433, opened 2025-12-07
Follow-up improvements and bug fixes

krinart:
## 📝 Summary

1. Removed shards that returned no records from watermark calculation
2. Moved persisting bootstrap checkpoint into a separate step
4. Fixed a bug of not skipping shards that are ancestors of active shards during discovery
5. Set default size of channel buffer to 1
6. Made default polling interval 0 instead of 200 (waiting is done by channel blocking)
7. Fixed a bug of not ignoring ancestor after its child has been removed from active

## 🔗 Related

## 📚 Docs

### DIFF
--- a/crates/dynamodb-streams/src/client.rs
+++ b/crates/dynamodb-streams/src/client.rs
@@ -104,12 +104,15 @@ impl Client {
             .sdk_client
             .get_stream_arn(self.table_name.clone())
             .await?;
+
         let state = initialize_state_from_checkpoint(
             stream_arn.clone(),
             &checkpoint,
             Arc::clone(&self.sdk_client),
         )
         .await?;
+
+        tracing::debug!("Stream initialized from checkpoint: {:#?}", state);
 
         let (tx, rx) = mpsc::channel(self.buffer);
 
@@ -125,7 +128,6 @@ impl Client {
             sender: tx,
             client: Arc::clone(&self.sdk_client),
             retry_strategy,
-            idle_timeout: None,
         };
 
         tokio::spawn(async move {
--- a/crates/dynamodb-streams/src/stream.rs
+++ b/crates/dynamodb-streams/src/stream.rs
@@ -22,6 +22,7 @@ use futures::{Stream, future::join_all};
 use std::collections::HashMap;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::sync::atomic::Ordering;
 use std::task::{Context, Poll};
 use std::time::SystemTime;
 use tokio::{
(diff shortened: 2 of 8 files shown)


crates/dynamodb-streams/src/client.rs
```diff
@@ -104,13 +104,16 @@ impl Client {
...
             .await?;
+
         let state = initialize_state_from_checkpoint(
             stream_arn.clone(),
             &checkpoint,
             Arc::clone(&self.sdk_client),
         )
         .await?;
 
+        tracing::debug!("Stream initialized from checkpoint: {:#?}", state);
```
sgrebnov, line 115: Pls see if we can include any additional stream/database identifiers (name)

crates/dynamodb-streams/src/stream.rs
```diff
@@ -162,7 +156,12 @@ impl DynamodbStreamProducer {
 
                     if had_transient_error {
                         // Transient error occurred during collection - apply backoff
-                        if let Some(duration) = backoff.next_backoff() {
+                        if let Some(mut duration) = backoff.next_backoff() {
+                            // Avoid sleeping for more than 60 seconds
+                            if duration > Duration::from_secs(60) {
```
sgrebnov, line 161: 1. Should this be duration = Max(60s, backoff.next_backoff())? Reseting back to 1s (reset) seems strange here
1. Pls also double check if there is max_backoff_duration already supported as a param

crates/dynamodb-streams/src/stream_state.rs
```diff
@@ -277,6 +272,9 @@ impl StreamState {
                 blocked
             );
 
+            tracing::debug!("Current state: {:#?}", self);
```
sgrebnov, line 275: Please double check this tracing is intended/required or should use trace instead of debug

crates/runtime/src/dataconnector/dynamodb.rs
```diff
@@ -85,12 +84,15 @@ const PARAMETERS: &[ParameterSpec] = &[
     ParameterSpec::runtime("scan_segments")
         .description("Number of segments. 'auto' by default.")
         .default(SEGMENTS_AUTO_STR),
-    ParameterSpec::runtime("stream_poll_interval_ms")
+    ParameterSpec::runtime("scan_interval")
```
sgrebnov, line 87: Please double check 0 scan interval is safe approach

sgrebnov (approved):

crates/dynamodb-streams/src/stream.rs
krinart: This one is intended. Since this is a stream, we don't want to sleep for too long. There's no max_backoff_duration

crates/dynamodb-streams/src/stream_state.rs
krinart: It is for now. It only prints when new shrad is discovered, which reconfigures the state. After it runs in production for some time, I will remove it.

PR #8439, opened 2025-12-08
fix: Disable Cayenne HashJoin rewriter optimizer

peasee:
## 📝 Summary

* Removes the optimizer rule that replaces `HashJoin`s for Cayenne with `ExactLeftAccumulator`s, as they are not sufficiently stable.


phillipleblanc (dismissed):

crates/runtime/tests/acceleration/snapshots/integration__acceleration__cron__test_append_cron_schedule_after_one.snap
phillipleblanc, line 1: why would this change?

peasee: I think the integration test failed, then it pushed the failed snapshot. Need to revert

phillipleblanc (approved):

PR #8469, opened 2025-12-08
Add snapshot creation logging

krinart:
## 📝 Summary

Add INFO log:
```
Snapshots for dataset ddb will be created every 300 batch updates
```

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh.rs
+++ b/crates/runtime/src/accelerated_table/refresh.rs
@@ -926,6 +926,10 @@ fn create_snapshot_callback(
 ) -> Option<SnapshotCallback> {
     let threshold = snapshots_trigger_threshold.unwrap_or(300i64);
 
+    tracing::info!(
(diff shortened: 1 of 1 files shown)


sgrebnov (dismissed):

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -926,6 +926,10 @@ fn create_snapshot_callback(
 ) -> Option<SnapshotCallback> {
```
sgrebnov, line 926: Nit: `create_snapshot_callback` name is very generic and might confuse, might be better to name it a little bit more specific?

create_throttled_snapshot_callback — Emphasizes the rate-limiting nature
create_batch_throttled_snapshot_trigger — Describes what triggers the throttling
build_throttled_checkpoint_callback — Highlights the checkpoint+snapshot behavior

sgrebnov (approved):

PR #8474, opened 2025-12-08
DuckDB agg pushdown: gate behind accelerator parameter

mach-kernel:
## 📝 Summary

To enable DuckDB agg pushdown, you must now specify `aggregate_pushdown_optimization: enabled` as shown below:
```yaml
version: v1
kind: Spicepod
name: my_spicepod

datasets:
  - from: file://test_data.parquet
    name: tdata
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
      params:
        query_federation: disabled
        aggregate_pushdown_optimization: enabled # disabled
```

## 🚨 Breaking Changes

n/a, not released yet

### DIFF
--- a/crates/runtime/src/dataaccelerator/duckdb.rs
+++ b/crates/runtime/src/dataaccelerator/duckdb.rs
@@ -69,6 +69,8 @@ pub(crate) mod settings;
 
 pub(crate) const DEFAULT_MIN_IDLE_CONNECTIONS: u32 = 10;
 pub(crate) const SPICE_ACCELERATOR_METADATA_KEY: &str = "spice.accelerator";
+pub(crate) const SPICE_OPT_DUCKDB_AGG_PUSHDOWN_KEY: &str =
+    "spice.optimizer.duckdb_aggregate_pushdown";
 
 use super::upsert_dedup;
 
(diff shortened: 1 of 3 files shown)


crates/runtime/src/dataaccelerator/duckdb.rs
```diff
@@ -301,6 +303,7 @@ const PARAMETERS: &[ParameterSpec] = &[
     ),
     ParameterSpec::runtime("on_refresh_recompute_statistics"),
     ParameterSpec::runtime("partitioned_write_buffer"),
+    ParameterSpec::runtime("aggregate_pushdown_optimization"),
```
peasee, line 306:
```suggestion
    ParameterSpec::runtime("enable_aggregate_pushdown"),
```

crates/runtime/src/dataaccelerator/duckdb.rs
```diff
@@ -549,6 +552,18 @@ pub(crate) async fn create_table_provider(
         "duckdb".to_string(),
     );
 
+    if cmd
+        .options
+        .get("aggregate_pushdown_optimization")
```
peasee, line 557:
```suggestion
        .get("enable_aggregate_pushdown")
```

peasee (dismissed):

crates/runtime/src/dataaccelerator/duckdb.rs
phillipleblanc: We try to avoid boolean parameters where possible - this should be `duckdb_aggregate_pushdown: enabled # disabled`

peasee (approved):

PR #8484, opened 2025-12-09
Use `Datafusion::Plan` over `Datafusion::Internal` for user-facing search errors.

Jeadie:
## 📝 Summary
- Prefer `Datafusion::Plan` over `Datafusion::Internal` when issue is user facing.
- For using facing messages using `Datafusion::Internal`, the inner message should not have a full stop.
```rust
DataFusionError::Internal(ref desc) => Cow::Owned(format!(
    "{desc}.\nThis issue was likely caused by a bug in DataFusion's code. \
    Please help us to resolve this by filing a bug report in our issue tracker: \
    https://github.com/apache/datafusion/issues"
)),
```
[source](https://github.com/apache/datafusion/blob/83736efc4ad8865019b0809ac9d87e63eabbe0a8/datafusion/common/src/error.rs#L575-L579).

### DIFF
--- a/crates/runtime/src/embeddings/udtf.rs
+++ b/crates/runtime/src/embeddings/udtf.rs
@@ -116,7 +116,7 @@ impl VectorSearchTableFuncArgs {
         match (self.column.as_deref(), cfg) {
             (Some(col), Some(cfg)) => Ok((col.to_string(), cfg)),
             (Some(col), None) => Err(DataFusionError::Internal(format!(
-                "User function 'vector_search' is called on table '{}' that does not have a embedding index on '{col}' column. Index is on column(s): {}.",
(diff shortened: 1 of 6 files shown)


crates/runtime/src/embeddings/udtf.rs
```diff
@@ -116,7 +116,7 @@ impl VectorSearchTableFuncArgs {
         match (self.column.as_deref(), cfg) {
             (Some(col), Some(cfg)) => Ok((col.to_string(), cfg)),
             (Some(col), None) => Err(DataFusionError::Internal(format!(
```
phillipleblanc: I don't think we should be using an internal datafusion error in our code

phillipleblanc (approved):

PR #8520, opened 2025-12-10
DDB Streams Integration Test + Memory Acceleration + Improved Warning

krinart:
## 📝 Summary

1. Added integration test for DynamoDB Streams by using dockerized dynamodb
  * New `endpoint_url` for DynamoDB Connector which allows to point it to local instance
2. Added support for in-memory acceleration (feature parity with kafka/debezium)
3. Added warning for cases:
  * `refresh_mode: changes` with disabled acceleration: `DynamoDB dataset X is configured for changes stream, but acceleration is disabled. Enable acceleration to use DynamoDB Streams`
  * In-memory acceleration: `Dataset X is not file-accelerated. DynamoDB Streams checkpoints will not be persisted.`
  * No `new_image`: `No new_image in DynamoDB Streams record. Make sure stream is configured with either NewImage or NewAndOldImages. https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html#Streams.Enabling`

## 🔗 Related

### DIFF
--- a/crates/runtime/src/dataconnector/dynamodb.rs
+++ b/crates/runtime/src/dataconnector/dynamodb.rs
@@ -100,7 +100,7 @@ const PARAMETERS: &[ParameterSpec] = &[
         .description("When using Streams, once tables reaches this lag, it will be reported as Ready")
         .default("2s"),
     ParameterSpec::runtime("endpoint_url")
-        .description("")
+        .description("Custom endpoint URL for DynamoDB. Useful for development and testing.")
 ];
 
 impl DataConnectorFactory for DynamoDBFactory {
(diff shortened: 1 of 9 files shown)


crates/runtime/src/dataconnector/dynamodb.rs
```diff
@@ -288,7 +306,12 @@ impl DataConnector for DynamoDB {
                 let dataset_name_3 = dataset_name.clone();
                 let dataset_name_4 = dataset_name.clone();
                 let dynamodb = Arc::new(dynamodb_ref.clone());
-                let dynamodb_sys = initialize_dynamodb_sys(&dataset).await?;
+                let dynamodb_sys = Arc::new(if dataset.is_file_accelerated() {
+                    initialize_dynamodb_sys(&dataset).await
+                } else {
+                    tracing::warn!("Dataset {dataset_name} is not file-accelerated. DynamoDB Streams checkpoints will not be persisted.");
```
Jeadie, line 312: Confirming: There is no restriction in using memory based accelerators, just a limitation on the persistence of checkpoints?

Jeadie (approved):

crates/runtime/src/dataconnector/dynamodb.rs
krinart: Correct

PR #8553, opened 2025-12-14
Add cayenne data accelerator

Jeadie:
## 📝 Summary
 - Add cayenne to search tests. 
 - Currently cannot be supported as vortex file paths are non-deterministic.


peasee (commented):
Could we reuse [`CAYENNE_PATH_FILTER_PATTERN`](https://github.com/spiceai/spiceai/blob/trunk/crates/test-framework/src/snapshot/mod.rs#L22) regex?

Jeadie:
> Could we reuse [`CAYENNE_PATH_FILTER_PATTERN`](https://github.com/spiceai/spiceai/blob/trunk/crates/test-framework/src/snapshot/mod.rs#L22) regex?

I think we'd have to expand it. If i understand right, this will only match `/data/**/*.vortex`. We need more broad than this since we currently set it as `cayenne_file_path: .spice/data/cayenne_acceleration_{unique_id}/`.

phillipleblanc (approved):

PR #8608, opened 2025-12-17
Docs: Distributed query ADR

mach-kernel:
## 📝 Summary

See diff


docs/decisions/004-distributed-query-framework.md
```diff
@@ -0,0 +1,76 @@
...
+## Context
+
+Spice aims to be a best-in-class query engine used by enterprises to quickly enable search and inference workloads. To tackle the "big data" warehouses frequently found in big orgs, Spice should be able to scale its query execution capability past a single process.
+
+## Assumptions
+
+1. Spice will need to process queries that exceed the memory and compute capacity of a single process.
+2. Users will run analytical queries and inference workloads that may be long-running and require fault tolerance.
+3. The distributed query framework should integrate well with DataFusion, Spice's existing query engine.
+4. The distributed framework will need to support custom Spice features like UDFs, metrics, and telemetry.
```
lukekim:
```suggestion
4. The distributed framework will need to support custom Spice features like acceleration, search, UDFs, metrics, and telemetry.
```

docs/decisions/004-distributed-query-framework.md
```diff
@@ -0,0 +1,82 @@
...
+    - Spice is already great at low-latency "edge duties" in single process mode with acceleration.
+- "Batteries included"
+  - Ballista has mature concepts for:
+    - Execution model: scheduler (query planning, task management, exec coordination), executor (workers)
+      - Scheduler's work queue model also allows dynamic resizing of executor pool while a query is running
+    - Metrics collection
+    - Shuffle service
+    - Arrow/RPC services around the above
+    - PySpark-like client support (important for DX/UX)
+  - `datafusion-distributed` has a much simpler model: any machine is a worker, a basic flight service consumes execution plans, and it is up to you to implement discovery, the concept of participating in a cluster, etc.
```
peasee: it's not clear from this wording whether this is a negative or positive for `datafusion-distributed`. Because it appears in the `Why` for Ballista, I am guessing this is a negative, but it could be clearer

docs/decisions/004-distributed-query-framework.md
```diff
@@ -0,0 +1,82 @@
+# DR: Use Apache Ballista as Spice's distributed query framework
```
peasee, line 1:
```suggestion
# DR-004: Use Apache Ballista as Spice's distributed query framework
```

docs/decisions/004-distributed-query-framework.md
mach-kernel: Well, it's meant to read neutrally, but I added some color as to why this could be a negative.

docs/decisions/004-distributed-query-framework.md
```diff
@@ -0,0 +1,81 @@
...
+    - Spice is already great at low-latency "edge duties" in single process mode with acceleration.
+- "Batteries included"
+  - Ballista has mature concepts for:
+    - Execution model: scheduler (query planning, task management, exec coordination), executor (workers)
+      - Scheduler's work queue model also allows dynamic resizing of executor pool while a query is running
+    - Metrics collection
+    - Shuffle service
+    - Arrow/RPC services around the above
+    - PySpark-like client support (important for DX/UX)
+  - `datafusion-distributed` has a much simpler model: any machine is a worker, a basic flight service consumes execution plans, and it is up to you to implement discovery, the concept of participating in a cluster, etc. It is more of a "box of tools" than something ready-to-use. Building the basics is time-consuming.
```
phillipleblanc, line 125: I think we realistically will need to customize a bunch of these to be able to have HA scheduler nodes, which Ballista doesn't natively support. In my initial design work for HA schedulers, I've already handled discovery, participating in a cluster, etc.

Given the ballista maintainers' approach so far to accepting our changes, we'd end up maintaining our own Ballista fork to implement our requirements. At what point should we just take the good parts of ballista and implement them in our own code on top of `datafusion-distributed` (which is a lot more flexible)?

phillipleblanc: The shuffle service is arguably the biggest thing we would be giving up from Ballista. But implementing that is a one-time cost vs the ongoing maintenance of a fork.

docs/decisions/004-distributed-query-framework.md
```diff
@@ -0,0 +1,81 @@
+# DR: Use Apache Ballista as Spice's distributed query framework
+
+## Context
+
+Spice aims to be a best-in-class query engine used by enterprises to quickly enable search and inference workloads. To tackle the "big data" warehouses frequently found in enterprise organizations, Spice should be able to scale its query execution capability past a single process.
+
+## Assumptions
+
+1. Spice will need to process queries that exceed the memory and compute capacity of a single process.
+2. Users will run analytical queries and inference workloads that may be long-running and require fault tolerance.
```
phillipleblanc, line 10: If fault-tolerance is a goal (which I believe it should be), then Spice will need an async API for submitting queries. All of the current APIs are synchronous - the Spice node keeps the connection open until the results are ready and then serves the results. Fault tolerance should include the scheduler layer as well, which is a broader concern than the framework we use for executing distributed queries (and should probably be its own DR).

docs/decisions/004-distributed-query-framework.md
phillipleblanc: I actually think using Ballista is the correct approach for single-node scheduler setups. I think the big open question is whether the changes we'd need to make for Ballista to work well for an HA setup would be welcome upstream or not.

lukekim (approved):

Issue #8559, opened 2025-12-15
Enhancement: Distributed Query: Stateless, HA cluster schedulers (active-active)


phillipleblanc:
## Goal-State/What/Result

In cluster mode, scheduler nodes must be highly available and horizontally scalable (no per-scheduler addressing). A scheduler pod may disappear/reappear at any time; in-flight queries coordinated by that scheduler should be resumable by either the same scheduler or another one. Active-active schedulers (no leader/failover).

## Why/Purpose

Required for production deployments of distributed query.

## By When

**Issue/Spec written and reviewed:** 12/16/2025
**Done-Done:** TBD

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Telemetry / Metrics / Task History
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials

## The Algorithm

- [ ] Every requirement questioned?
- [ ] Delete (Scope) any part you can.
- [ ] Simplify.
- [ ] Break down into smaller iterations/milestones.
- [ ] Opportunities for automation.

### Generalized scenarios:
- Ingestion
	- Ingest/index for Full Text/Vector Search at high rate (from any data connector)
		- Elastic scale for rate of ingestion
		- Vectorization (embedding generation)
		- Accelerated indexes
		- Persisted indexes
			- S3 Vector ingestion
	- Ingestion is continuous
- Acceleration
	- Snapshotting
- Distributed query
	- Distributed search
	- Elastic scale for rate of query
- "Workflow"
	- Ability to chain actions together
- Spice to spice
	- single node Spice -> Spice cluster schedulers
	- single node Spice -> single node Spice

### Hard Requirements:
- High availability - needs to tolerate machines going away at any time.
	- Async API
	- Intermediate shuffle to persistent storage
- Needs to be elastic
	- Need to be able to resize cluster while its running
- ACID
	- Durable (don't lose data)
	- Integrity (data can't be wrong)
	- Eventually consistent (at least once execution)
	- Atomic
- Coordinators are load-balanced
- Resumability of both jobs and stages
- Spicepod is static across one deployment
- Performance goals
	- Query performance
	- Search performance
[...]

PR #8616, opened 2025-12-18
Docs: Ballista extension ADR

mach-kernel:
## 📝 Summary

See diff

### DIFF
--- a/docs/decisions/005-ballista-extensions.md
+++ b/docs/decisions/005-ballista-extensions.md
@@ -14,6 +14,7 @@ Stock Apache Ballista has several gaps that prevent seamless integration with Sp
 These limitations would require users to manually replicate schema definitions and function registrations across all clients, undermining the developer experience. Additionally, Spice's multi-tenant architecture requires secure cluster communication that stock Ballista does not provide.
 
 Related decisions:
+
 * [DR-004: Use Apache Ballista as Spice's distributed query framework](./004-distributed-query-framework.md)
 
 ## Assumptions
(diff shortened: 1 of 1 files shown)


docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,141 @@
+# DR: Extend Apache Ballista with Remote Catalog, UDF Sync, and Cluster Security
```
peasee, line 1:
```suggestion
# DR-005: Extend Apache Ballista with Remote Catalog, UDF Sync, and Cluster Security
```

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,141 @@
...
+
+Stock Apache Ballista has several gaps that prevent seamless integration with Spice:
+
+1. Clients cannot discover or query tables registered only on the scheduler without explicit registration on the client side.
+2. Custom UDFs defined on the scheduler are not available for client-side query planning.
+3. No built-in support for TLS communication between cluster components or API key authentication.
+4. No programmatic way to detect when executors are ready to process work.
+
+These limitations would require users to manually replicate schema definitions and function registrations across all clients, undermining the developer experience. Additionally, Spice's multi-tenant architecture requires secure cluster communication that stock Ballista does not provide.
+
```
peasee, line 15:
```suggestion

Related decisions:
* [DR-004: Use Apache Ballista as Spice's distributed query framework](./004-distributed-query-framework.md)

```

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,141 @@
...
+- **Developer experience first**
+- **Simplicity**
+- **Security by default**
+
+## Decision
+
+Spice will extend Apache Ballista with the following capabilities:
+
+1. **Remote Catalog Synchronization** - Clients automatically fetch and populate catalog metadata from the scheduler, creating stub table providers for remote tables.
```
peasee: It doesn't look like the selected options for catalog+UDF sync appear in the options list above?

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,141 @@
...
+2. **Remote UDF Synchronization** - Clients automatically fetch UDF metadata (signatures, return types, documentation) from the scheduler for query planning.
+3. **Cluster Security Extensions** - Support for TLS communication between cluster components and gRPC metadata interceptors for API key authentication.
+4. **Executor Readiness Signaling** - Executors signal readiness through an optional oneshot channel after completing their first poll loop iteration.
```
peasee: This selected option does not appear in the list of options

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Executor Readiness Signaling
+
+- Allows Spice runtime to coordinate executor lifecycle without brittle polling.
+- One-time signal via oneshot channel is non-intrusive and doesn't affect hot path performance.
+- Useful for testing and deployment orchestration where waiting for "ready" state is necessary.
+
+**Why not**:
+
+### Fork Maintenance Burden
```
peasee: This layout is confusing, bold text between actual headers. Laying out `Why not`/`Why` as their own headers would be clearer:

```
### Why not

#### Fork maintenance burden
```

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,141 @@
...
+
+- Maintaining a fork increases complexity during DataFusion/Ballista upgrades.
+  - **Mitigation**: Keep extensions minimal and isolated. Track upstream changes. Contribute extensions back to Apache Ballista when mature.
+
+### Catalog Sync Performance
```
peasee: Let's use the same headers for `Why not` as for `Why`? otherwise it is not immediately obvious which points apply to which option

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,141 @@
...
+
+- Automatic catalog population on every client connection could be expensive for large catalogs.
+  - **Mitigation**: Accept this tradeoff for developer experience. Future optimization could add caching or lazy population if needed.
+
+### Alternative: External Catalog Service
```
peasee: These alternatives should appear in the options considered, not here?

docs/decisions/005-ballista-extensions.md
mach-kernel: I think it is option 2

docs/decisions/005-ballista-extensions.md
mach-kernel: I believe this is option 2 for the readiness section

docs/decisions/005-ballista-extensions.md
mach-kernel: I agree with you and made the change. I used the template from the other ADRs.

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
+# DR-005: Extend Apache Ballista with Remote Catalog, UDF Sync, and Cluster Security
+
+## Context
+
+Following the decision to adopt Apache Ballista as Spice's distributed query framework (DR-004), Spice needs to integrate Ballista's scheduler-executor model with its existing features including metrics, telemetry, custom UDFs, search capabilities, and security requirements.
+
+Stock Apache Ballista has several gaps that prevent seamless integration with Spice:
+
+1. Clients cannot discover or query tables registered only on the scheduler without explicit registration on the client side.
```
phillipleblanc, line 9: I don't understand what a "client" is in this respect. Clients are already talking to the scheduler nodes for the queries, which already has the tables registered by definition.

Any client (or client SDK) we ship would already talk to the schedulers

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+2. Custom UDFs defined on the scheduler are not available for client-side query planning.
```
phillipleblanc, line 10: Same comment as above - I don't think this is an issue for us.

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+3. No built-in support for TLS communication between cluster components or API key authentication.
```
phillipleblanc, line 11: I would expect we could get this upstreamed.

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+4. No programmatic way to detect when executors are ready to process work.
+
+These limitations would require users to manually replicate schema definitions and function registrations across all clients, undermining the developer experience. Additionally, Spice's multi-tenant architecture requires secure cluster communication that stock Ballista does not provide.
```
phillipleblanc, line 14: I don't think this sentence applies give the above comments.

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+## Context
+
+Following the decision to adopt Apache Ballista as Spice's distributed query framework (DR-004), Spice needs to integrate Ballista's scheduler-executor model with its existing features including metrics, telemetry, custom UDFs, search capabilities, and security requirements.
+
+Stock Apache Ballista has several gaps that prevent seamless integration with Spice:
+
+1. Clients cannot discover or query tables registered only on the scheduler without explicit registration on the client side.
+2. Custom UDFs defined on the scheduler are not available for client-side query planning.
+3. No built-in support for TLS communication between cluster components or API key authentication.
+4. No programmatic way to detect when executors are ready to process work.
```
phillipleblanc, line 12: I don't think we need ballista to implement this at a framework level, we could easily just add this to our own gRPC service.

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+Related decisions:
+* [DR-004: Use Apache Ballista as Spice's distributed query framework](./004-distributed-query-framework.md)
+
+## Assumptions
+
+1. Spice will maintain a fork of Apache Ballista to implement these extensions until they can be upstreamed.
+2. The scheduler will be the source of truth for catalog metadata and UDF definitions in a Ballista cluster.
+3. Cluster components (scheduler, executors, clients) need secure communication for production deployments.
+4. Extensions should be backward compatible and not require changes to the core Ballista execution model.
+5. Client-side planning with remote metadata should feel transparent to users.
```
phillipleblanc: I don't think we want to support "client-side planning". The planning is the responsibility of the scheduler node.

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+## Options
+
+These options describe solutions for each high level problem.
+
+### Remote Catalog and UDF Synchronization
+
+1. Require users to manually register tables and UDFs on both scheduler and clients.
+2. Implement catalog and UDF metadata synchronization from scheduler to clients.
+3. Use a shared external catalog service (e.g., Hive Metastore, AWS Glue).
+4. External catalog service: Hive Metastore or AWS Glue would provide catalog synchronization without custom code. However, this would require a custom client build.
```
phillipleblanc: This section goes away (see above comments)

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+- **Developer experience first**
+- **Simplicity**
+- **Security by default**
+
+## Decision
+
+Spice will extend Apache Ballista with the following capabilities:
+
+1. **Remote Catalog Synchronization** - Clients automatically fetch and populate catalog metadata from the scheduler, creating stub table providers for remote tables.
+2. **Remote UDF Synchronization** - Clients automatically fetch UDF metadata (signatures, return types, documentation) from the scheduler for query planning.
```
phillipleblanc:
```suggestion
```

Clients don't do planning

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+3. **Cluster Security Extensions** - Support for TLS communication between cluster components and gRPC metadata interceptors for API key authentication.
+4. **Executor Readiness Signaling** - Executors signal readiness through an optional oneshot channel after completing their first poll loop iteration.
```
phillipleblanc: Relying on communication from an executor sending a signal to a scheduler limits the possible design solutions for HA schedulers (as each scheduler needs to reliably get that signal somehow).

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+
+##### Remote Catalog and UDF Synchronization
+
+- Eliminates manual schema and UDF registration on clients, dramatically improving developer experience.
+- Enables `SHOW TABLES` and catalog discovery queries to work seamlessly from any client.
+- Clients can plan queries against the full catalog without pre-registration.
+- `RemoteTableProvider` and `RemoteScalarUDF` stubs provide clean separation: they participate in query planning but defer execution to the scheduler, ensuring actual data access happens on the cluster.
+- Serialization via protobuf ensures efficient metadata transfer and compatibility with Ballista's existing RPC mechanisms.
+- Users can write queries in notebooks or applications without needing intimate knowledge of the cluster's registered tables.
+- A single build of a Ballista client (after these changes) would remain compatible with subsequent releases unless there is a significant change in RPC schema/workflow.
```
phillipleblanc:
```suggestion
```

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+
+##### Cluster Security Extensions
+
+- Production deployments require secure communication between cluster components.
+- Spice's multi-tenant architecture requires API key authentication for cluster access.
```
phillipleblanc:
```suggestion
- Spice's multi-tenant architecture requires mTLS for cluster access.
```

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+- The endpoint customization pattern (`BallistaConfigGrpcEndpoint`) provides flexibility for TLS configuration without hardcoding specific certificate paths.
+- gRPC metadata interceptors (`BallistaGrpcMetadataInterceptor`) enable arbitrary header injection for authentication and tracing.
+- Making security optional (via builder methods) maintains backward compatibility and allows incremental adoption.
+- The extensions integrate cleanly with tonic's `Endpoint` and `Interceptor` traits, following Rust ecosystem patterns.
+
+##### Executor Readiness Signaling
+
+- Allows Spice runtime to coordinate executor lifecycle without brittle polling.
```
phillipleblanc: I think a better design that more naturally allows for multi-scheduler configurations is for each scheduler to check the health of the executors it knows about.

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+
+##### Fork Maintenance Burden
+
+- Maintaining a fork increases complexity during DataFusion/Ballista upgrades.
+  - **Mitigation**: Keep extensions minimal and isolated. Track upstream changes. Contribute extensions back to Apache Ballista when mature.
+
+##### Catalog Sync Performance
+
+- Automatic catalog population on every client connection could be expensive for large catalogs.
+  - **Mitigation**: Accept this tradeoff for developer experience. Future optimization could add caching or lazy population if needed.
```
phillipleblanc:
```suggestion
```

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+
+
+## Consequences
+
+- Spice will maintain a fork of Apache Ballista at `github.com/spiceai/datafusion-ballista` with the extensions.
+- The Ballista fork will need to be kept in sync with upstream Apache Ballista releases, particularly during DataFusion upgrades.
+- New RPC endpoints will be added to Ballista's protobuf definitions:
+  - `GetCatalog` - Returns catalog, schema, and table metadata with Arrow schemas.
+  - `GetFunctions` - Returns UDF signatures, return types, and documentation.
```
phillipleblanc:
```suggestion
```

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+- `SessionContextExt` in the Ballista client will expose new builder methods:
+  - `remote_with_catalog()` - Connects to scheduler and populates catalog.
+  - `populate_functions_from_scheduler()` - Fetches and registers remote UDFs.
```
phillipleblanc:
```suggestion
```

docs/decisions/005-ballista-extensions.md
```diff
@@ -0,0 +1,136 @@
...
+  - `with_ballista_grpc_metadata()` - Configures metadata headers for authentication.
+  - `with_ballista_override_create_grpc_client_endpoint()` - Customizes gRPC endpoints for TLS.
+- `SchedulerConfig` and `ExecutorConfig` will expose endpoint customization hooks for TLS configuration.
+- `RemoteTableProvider` and `RemoteScalarUDF` will be introduced as stub implementations that error if execution is attempted locally (they must be executed on the cluster).
```
phillipleblanc:
```suggestion
```

docs/decisions/005-ballista-extensions.md
phillipleblanc: Client here means an external code that is capable of doing its own query planning and then submitting a fully-formed plan to a scheduler node to oversee the execution. I think we should assume we keep the boundary as SQL and require the scheduler nodes to do their own query planning.

docs/decisions/005-ballista-extensions.md
Jeadie: Do we have an ADR for HA? I feel like it would be valuable to review both of these simulatenously

phillipleblanc: No, that is still under design, but the current spec is here: https://github.com/spiceai/spiceai/issues/8559

docs/decisions/005-ballista-extensions.md
```diff
@@ -22,115 +23,156 @@ Related decisions:
...
 
 ### Executor Readiness
 
-1. Poll executor endpoints until they respond successfully.
-2. Add explicit readiness signaling through a channel or callback.
-3. Use existing gRPC health check protocols.
+1. **Scheduler-side polling**: Scheduler polls executor health endpoints until they respond. Compatible with HA scheduler designs.
+2. **Executor-to-scheduler signal**: Executors signal readiness through a channel or callback after first poll loop. Simpler but limits HA scheduler designs.
+3. **gRPC health check protocol**: Use existing gRPC health check protocols (standard, well-supported).
+4. **Kubernetes readiness probes**: Rely on K8s orchestration for readiness; Spice runtime doesn't manage executor lifecycle directly.
```
phillipleblanc, line 56: This isn't a valid option in this context - the schedulers need to know which executors are alive so they don't assign work to dead executor nodes.

docs/decisions/005-ballista-extensions.md
```diff
@@ -22,115 +23,156 @@ Related decisions:
...
-- Users can write queries in notebooks or applications without needing intimate knowledge of the cluster's registered tables.
-- A single build of a Ballista client (after these changes) would remain compatible with subsequent releases unless there is a significant change in RPC schema/workflow.
+* **Phase 1 (SQL-only)**: Enables catalog discovery via `SHOW TABLES`, `DESCRIBE TABLE`, and function introspection from any client.
+* **Phase 2 (Future)**: Required for DataFrame API - clients using DataFrame operations (e.g., `df.filter().select()`) will perform local query planning and need catalog metadata to resolve table schemas.
+* **Phase 2 (Future)**: Required for Substrait support - clients submitting pre-planned Substrait queries will need UDF signatures to build valid plans.
+* Eliminates manual schema and UDF registration on clients, dramatically improving developer experience.
+* Clients can discover and explore the full catalog without pre-registration.
+* `RemoteTableProvider` and `RemoteScalarUDF` stubs provide clean separation: they participate in query planning but defer execution to the scheduler, ensuring actual data access happens on the cluster.
```
phillipleblanc, line 88: This feels like an implementation detail that shouldn't be in the decision record

docs/decisions/005-ballista-extensions.md
```diff
@@ -22,115 +23,156 @@ Related decisions:
...
+#### mTLS Complexity
 
-- Automatic catalog population on every client connection could be expensive for large catalogs.
-  - **Mitigation**: Accept this tradeoff for developer experience. Future optimization could add caching or lazy population if needed.
+* mTLS requires certificate management infrastructure (CA, cert rotation, distribution).
+  * **Mitigation**: Provide documentation and tooling for certificate generation. Consider integration with cert-manager in Kubernetes environments.
+
+#### Executor-to-Scheduler Signaling (rejected)
+
+* Original design had executors signal readiness to scheduler via oneshot channel.
```
phillipleblanc, line 128: It wasn't just readiness, it is a registration protocol that is built into Ballista.

The registration API's purpose was to let the scheduler know about which executors exist - if we remove this then we need to design another way for the schedulers to discover which executors are participating in the cluster, in a way that can handle dynamically resizing executor pools.

Once the schedulers know which executors exist, then they can manage their own health pings to those executors.

docs/decisions/005-ballista-extensions.md
```diff
@@ -22,115 +23,156 @@ Related decisions:
...
-- The executor's `poll_loop` function will accept an optional `oneshot::Sender` to signal readiness after the first successful poll.
-- Integration testing will need to cover TLS-enabled cluster scenarios and remote metadata synchronization.
-- Documentation will be needed for configuring TLS and API key authentication in distributed Spice deployments.
-- The development team will need processes for managing the fork, tracking upstream changes, and potentially upstreaming features.
+### Fork Management
+
+* Spice will maintain a fork of Apache Ballista at `github.com/spiceai/datafusion-ballista` with the extensions.
+* The Ballista fork will need to be kept in sync with upstream Apache Ballista releases, particularly during DataFusion upgrades.
+* The development team will need processes for managing the fork, tracking upstream changes, and potentially upstreaming features.
+* TLS/mTLS extensions are likely candidates for upstreaming to Apache Ballista.
```
phillipleblanc, line 140: We'll also need changes to the executor registration, which right now for ballista is hard-coded to call back to a single scheduler

lukekim (approved):

PR #8652, opened 2025-12-23
Serialize acceleration snapshots with refresh writes

phillipleblanc:
## Summary
- serialize snapshot creation with accelerator writes using the shared mutex - this prevents a potential race condition brought up by @krinart where the acceleration could be written to in between the checkpoint and the snapshot creation.
- add an integration test that asserts snapshots wait on the mutex

### DIFF
--- a/crates/runtime/src/accelerated_table/caching.rs
+++ b/crates/runtime/src/accelerated_table/caching.rs
@@ -180,7 +180,7 @@ impl CacheRefreshHelper {
         accelerator: Arc<dyn TableProvider>,
         dataset_name: &str,
         ttl: Duration,
-        accelerator_mutex: Arc<Mutex<()>>,
+        accelerator_write_mutex: Arc<Mutex<()>>,
     ) -> DataFusionResult<usize> {
         let ctx = SessionContext::new();
         let state = ctx.state();
@@ -238,7 +238,7 @@ impl CacheRefreshHelper {
             let federated = Arc::clone(&federated);
             let accelerator = Arc::clone(&accelerator);
             let dataset_name = dataset_name.to_string();
-            let accelerator_mutex = Arc::clone(&accelerator_mutex);
+            let accelerator_write_mutex = Arc::clone(&accelerator_write_mutex);
 
             async move {
                 tracing::debug!(
@@ -257,7 +257,7 @@ impl CacheRefreshHelper {
                 let refreshed_rows: usize = batches.iter().map(RecordBatch::num_rows).sum();
 
                 // Acquire the mutex to protect accelerator operations
-                let lock_guard = accelerator_mutex.lock().await;
+                let lock_guard = accelerator_write_mutex.lock().await;
 
                 // Upsert this specific cache entry - removes rows matching the filters
                 // and adds the new data, preserving other cache entries.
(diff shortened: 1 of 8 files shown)


Jeadie (dismissed):

crates/runtime/src/accelerated_table/caching.rs
```diff
@@ -978,7 +978,7 @@ pub struct CachingAccelerationScanExec {
     projection: Option<Vec<usize>>,
     limit: Option<usize>,
     /// Mutex to protect concurrent access to the accelerator during cache operations
```
krinart:
```suggestion
    /// Mutex to protect concurrent access to the accelerator during cache/snapshot operations
```

crates/runtime/src/accelerated_table/mod.rs
```diff
@@ -232,7 +232,7 @@ pub struct AcceleratedTable {
     cache_stale_if_error: bool,
     io_runtime: Handle,
     /// Mutex to protect concurrent cache operations (insert, upsert) to the accelerator
```
krinart:
```suggestion
    /// Mutex to protect concurrent access to the accelerator during cache/snapshot operations
```

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -481,7 +481,7 @@ pub struct Refresher {
     resource_monitor: Option<crate::resource_monitor::ResourceMonitor>,
     /// Mutex to protect concurrent cache operations (insert, upsert) to the accelerator.
```
krinart:
```suggestion
    /// Mutex to protect concurrent access to the accelerator during cache/snapshot operations
```

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -117,7 +117,7 @@ pub struct RefreshTaskBuilder {
     io_runtime: Handle,
     resource_monitor: Option<crate::resource_monitor::ResourceMonitor>,
     /// Mutex to protect concurrent access to the accelerator during cache operations.
```
krinart:
```suggestion
    /// Mutex to protect concurrent access to the accelerator during cache/snapshot operations.
```

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -259,7 +259,7 @@ pub struct RefreshTask {
...
```
krinart:
```suggestion
    /// Mutex to protect concurrent access to the accelerator during cache/snapshot operations.
```

crates/runtime/src/accelerated_table/refresh_task_runner.rs
```diff
@@ -54,7 +54,7 @@ pub struct RefreshTaskRunnerBuilder {
     resource_monitor: Option<crate::resource_monitor::ResourceMonitor>,
     /// Mutex to protect concurrent cache operations (insert, upsert) to the accelerator.
```
krinart:
```suggestion
    /// Mutex to protect concurrent access to the accelerator during cache/snapshot operations.
```

PR #8662, opened 2025-12-23
fix: Use a GitHub rate controller per auth context

peasee:
## 📝 Summary

* Builds a GitHub rate controller for each auth context (e.g. static token, app/installation instance, etc)
* Adds a simple `dyn_hash` method on `TokenProvider` as `Hash` is not dyn-compatible

### DIFF
--- a/crates/runtime/src/dataconnector/github/mod.rs
+++ b/crates/runtime/src/dataconnector/github/mod.rs
@@ -62,7 +62,7 @@ use std::num::NonZeroU32;
 use std::sync::LazyLock;
 use std::{any::Any, future::Future, pin::Pin, str::FromStr, sync::Arc, time::Duration};
 use token_provider::{StaticTokenProvider, TokenProvider};
-use tokio::sync::{Mutex, Semaphore};
+use tokio::sync::{Mutex, RwLock, Semaphore};
 use url::Url;
 
 use super::{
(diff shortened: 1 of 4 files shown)


crates/runtime/src/dataconnector/github/mod.rs
```diff
@@ -83,7 +83,20 @@ mod workflows;
 static GITHUB_CONCURRENCY_LIMITS: LazyLock<Mutex<HashMap<String, Arc<Semaphore>>>> =
     LazyLock::new(|| Mutex::new(HashMap::new()));
 
-static GLOBAL_GITHUB_RATE_CONTROLLER: LazyLock<Arc<RateController>> = LazyLock::new(|| {
+static GITHUB_AUTH_CONTEXT_RATE_CONTROLLERS: LazyLock<
```
phillipleblanc, line 86: I realize we were already doing this previously, so feel free to ignore and file an issue instead.

However, using a static here for "global" state has bitten us in the past previously, especially when it came to testing multiple instances of the code in parallel. It would be better if we could tie this state back to the Runtime instance it is associated with.

phillipleblanc: It also means that in any dynamic reload scenario (of which we have none today, but could in the future) we might have leftover rate controllers for invalid contexts.

phillipleblanc (approved):

crates/runtime/src/dataconnector/github/mod.rs
peasee: This could result in some broader data connector changes for Runtime, so I've filed a tracking issue.

PR #8706, opened 2025-12-29
Add support for abfss

krinart:
## 📝 Summary

Make sure we properly support `abfss` along with `abfs` url prefix for object storage


crates/runtime/src/dataconnector/abfss.rs
```diff
@@ -0,0 +1,4 @@
+use crate::dataconnector::abfs::AzureBlobFSFactory;
+use crate::register_data_connector;
+
+register_data_connector!("abfss", AzureBlobFSFactory);
```
phillipleblanc: You can just put this in the same file as `abfs`, similar to how we register both `http` and `https` in the same file

krinart: Thanks, done

phillipleblanc (approved):

PR #8718, opened 2025-12-30
refactor: Remove cluster feature flag

phillipleblanc:
## 📝 Summary

The `cluster` Cargo feature flag is default and having it makes some of the logic (especially the flight servers) more complex than necessary.


bin/spiced/src/lib.rs
```diff
@@ -263,6 +259,9 @@ pub async fn run(args: Args) -> Result<()> {
         Err(_) => {
             // No explicit role specified, silently continue in standalone mode
         }
+    #[cfg(feature = "cluster")]
```
ewgenius: Shouldn't it be removed?

phillipleblanc: bad merge, will fix

ewgenius (approved):

lukekim (approved):

PR #8721, opened 2025-12-30
DR-006: High Availability Distributed Query with Stateless Schedulers

lukekim:


docs/decisions/006-ha-distributed-query.md
```diff
@@ -0,0 +1,267 @@
...
+
+Stock Apache Ballista uses a single-scheduler model where the scheduler is a single point of failure. If the scheduler process dies:
+
+1. All in-flight queries are lost with no recovery mechanism.
+2. Clients must manually reconnect and resubmit queries.
+3. Executor state and task progress are orphaned.
+
+For production deployments, Spice requires:
+
+* **Active-active schedulers**: Multiple scheduler nodes behind a load balancer with no leader election or failover delays.
```
phillipleblanc:
```suggestion
* **Active-active schedulers**: Multiple scheduler nodes behind a load balancer/virtual IP with no leader election or failover delays.
```

docs/decisions/006-ha-distributed-query.md
```diff
@@ -0,0 +1,267 @@
...
+* **Stateless schedulers**: Any scheduler can handle any request; scheduler pods may disappear/reappear at any time.
+* **Resilient query execution**: In-flight queries should fail gracefully when a scheduler dies, with clients able to resubmit to any scheduler.
```
phillipleblanc, line 21: Shouldn't this be that in-flight queries are automatically retried? They shouldn't fail from the perspective of the client.

Having the in-flight queries fail is easier to implement and might simplify some of the job state we need to do, but I don't think that is what we want

docs/decisions/006-ha-distributed-query.md
```diff
@@ -0,0 +1,267 @@
...
+* **Elastic scaling**: Ability to resize the cluster (add/remove executors) while queries are running.
+
+Related decisions:
+
+* [DR-004: Use Apache Ballista as Spice's distributed query framework](./004-distributed-query-framework.md)
+* [DR-005: Extend Apache Ballista with Remote Catalog, UDF Sync, and Cluster Security](./005-ballista-extensions.md)
+
+## Assumptions
+
+1. Schedulers are deployed behind a load balancer; clients connect to a single scheduler endpoint.
```
phillipleblanc:
```suggestion
1. Schedulers are deployed behind a load balancer/virtual IP; clients connect to a single scheduler endpoint.
```

docs/decisions/006-ha-distributed-query.md
```diff
@@ -0,0 +1,267 @@
...
+2. The cluster runs on Kubernetes; executor discovery leverages K8s EndpointSlices or DNS.
```
phillipleblanc, line 32: Based on some of the feedback we've gotten, it looks like we'll need to support cluster mode without the operator available? One option for executor discovery could be to ask the operator, but if we need this to work without the operator then we need an alternative (i.e. schedulers ask K8s directly or use DNS)

docs/decisions/006-ha-distributed-query.md
```diff
@@ -0,0 +1,267 @@
...
+3. Spicepod configuration is static within a deployment; dynamic spicepod updates are out of scope.
+4. Spice is object-store native/first. Object stores like S3 with conditional writes are the preferred mechanism for shared state and shuffle data persistence.
```
phillipleblanc, line 34: Shared state makes sense to store in object-store - shuffle state will require a ballista change. One of the ballista assumptions at the moment is that shuffle state is tied to an executor, so we'll need to tweak that (Shuffle partitions are represented by PartitionLocation that includes executor_meta [host/port/id] plus a local path)

phillipleblanc: The upside is that if an executor goes away, it doesn't take any shuffle state with it that we would need to retry.

docs/decisions/006-ha-distributed-query.md
```diff
@@ -0,0 +1,267 @@
...
+5. When shared scheduler state is required (e.g., for query resumption), S3 with conditional writes (PutObject with `If-None-Match`) provides distributed coordination without external infrastructure.
+6. Network partitions between schedulers and executors should cause affected queries to fail rather than hang indefinitely.
+7. Queries are submitted via an async API; synchronous query APIs are not suitable for HA (see DR-004 consequences).
+8. Initial scope targets graceful failure and resubmission; automatic query resumption by a different scheduler is a future enhancement.
+9. Executors are stateless workers; all durable state lives in object storage.
```
phillipleblanc:
```suggestion
9. Executors are stateless workers; all durable/intermediate result state lives in object storage.
```

Shuffle state isn't technically durable, so we should call it out here

phillipleblanc (dismissed):

sgrebnov (dismissed):

phillipleblanc (approved):

PR #8739, opened 2026-01-02
Add FTP, NFS, & SMB TPCH SF1 spicepods

lukekim:
This pull request introduces support for running network-attached storage (NAS) benchmark tests, specifically targeting SMB (Server Message Block) protocol, as part of the CI workflow. It includes new workflow definitions, updates to test dispatch logic, and the addition of a new federated SMB test configuration.

**Key changes:**

**CI/CD Workflow Enhancements:**

* Added a new workflow file, `.github/workflows/testoperator_run_nas.yml`, to define and run NAS benchmark tests via workflow dispatch, supporting configurable parameters such as the Spicepod file, commit, and validation options.
* Updated `.github/workflows/testoperator_dispatch.yml` to support the new `nas` workflow type, including logic to dispatch NAS tests both on schedule and via workflow input. [[1]](diffhunk://#diff-8167c43a6ca2fc8578931a778138b5b373b6207c16f599679a80d90c39a2230fR25) [[2]](diffhunk://#diff-8167c43a6ca2fc8578931a778138b5b373b6207c16f599679a80d90c39a2230fR150-R157) [[3]](diffhunk://#diff-8167c43a6ca2fc8578931a778138b5b373b6207c16f599679a80d90c39a2230fR266-R276)

**Test Configuration:**

* Added `tools/testoperator/dispatch/nas/smb[parquet].yaml` to define the NAS test parameters for SMB, including the path to the federated Spicepod and enabling result validation. ([tools/testoperator/dispatch/nas/smb[parquet].yamlR1-R4](diffhunk://#diff-1f29599b8e5ed3511ed0c7d83556fd5662fe59ad1d4b17b141e850ad78975166R1-R4))
* Introduced a new federated SMB Spicepod configuration at `test/spicepods/tpch/sf1/federated/smb[parquet].yaml` for use in the NAS benchmarks, specifying datasets and SMB connection parameters. ([test/spicepods/tpch/sf1/federated/smb[parquet].yamlR1-R31](diffhunk://#diff-415657c0695185e50c3d8dd9c8f1ed696ccaf23099deb2315c06be312e3f7769R1-R31))


.github/workflows/testoperator_dispatch.yml
```diff
@@ -147,6 +148,14 @@ jobs:
           SPICED_COMMIT: ${{ steps.setup-spiced.outputs.SPICED_COMMIT }}
           WORKFLOW_COMMIT: ${{ matrix.branch }}
 
+      - name: Dispatch Testoperator - Scheduled NAS Bench
```
phillipleblanc: Shouldn't this just be part of the existing benchmark dispatch?

tools/testoperator/dispatch/tpch/sf1/federated/smb[parquet].yaml
```diff
@@ -0,0 +1,4 @@
+tests:
```
Comment (Jeadie) [inline tools/testoperator/dispatch/tpch/sf1/federated/smb[parquet].yaml:1]: This should be in `tools/testoperator/dispatch/tpch/sf1/federated/`. Then we don't need new step (`Dispatch Testoperator - Scheduled NAS Bench`) in `.github/workflows/testoperator_dispatch.yml`

.github/workflows/testoperator_dispatch.yml
Jeadie: The key reason is that we don't support `runner_type:  spiceai-macos` on our current `testoperator_run_*.yaml`. But I believe we could just support this in the associated github workflows.

lukekim: Yeah, I need to run this on the Macs.

.github/workflows/testoperator_dispatch.yml
```diff
@@ -256,6 +265,16 @@ jobs:
           SPICED_COMMIT: ${{ steps.setup-spiced.outputs.SPICED_COMMIT }}
           WORKFLOW_COMMIT: ${{ github.event.ref }}
 
+      - name: Dispatch Testoperator - ${{ github.event.inputs.workflow_type }} - NAS
+        if: ${{ github.event.inputs.workflow_type == 'nas' }}
```
Jeadie: You shouldn't need this conditional. As long as only nas workflows are in `tools/testoperator/dispatch/nas/smb[parquet].yaml`, then when `workflow_type != 'nas'`, no runs will be dispatched. 

For example, this is why the text-to-sql below doesn't need an `.if` either

Jeadie (approved):

PR #8746, opened 2026-01-02
feat: Add scheduler-driven executor discovery via DNS SRV

phillipleblanc:
## 📝 Summary

Implement scheduler-driven executor discovery for HA stateless schedulers in cluster mode. Instead of executors registering themselves with schedulers, schedulers now discover executors by querying DNS SRV records (typically from Kubernetes headless services).

UX Change: New CLI flag for schedulers: `--executor-discovery-dns <dns>`

```rust
    /// The DNS SRV record name for discovering executors (scheduler mode only).
    /// When set, the scheduler will periodically query this DNS SRV record to discover
    /// executor endpoints and automatically register them.
    ///
    /// This is typically a Kubernetes headless service name, e.g.:
    ///   my-executor.default.svc.cluster.local
    ///
    /// The SRV query returns executor hostnames and ports, which are then used to
    /// call the `DescribeExecutor` RPC on each discovered executor.
    #[arg(long = "executor-discovery-dns", value_name = "EXECUTOR_DISCOVERY_DNS")]
    pub executor_discovery_dns: Option<String>,
```

Changes:
- Add ExecutorService gRPC with DescribeExecutor RPC (spice.proto) for schedulers to query executor registration info (id, host, port, task_slots)
- Add --executor-discovery-dns CLI parameter for schedulers to specify the DNS SRV record name to query for executor discovery
- Add DNS SRV lookup support to ns_lookup crate (lookup_srv function)
- Implement executor discovery loop with Fibonacci backoff (1s -> 30s) that:
  - Runs immediately on startup
  - Queries DNS SRV records periodically
  - Calls DescribeExecutor on discovered endpoints
  - Registers new executors with the scheduler
  - Resets backoff when new executors are found
- Use advertise address as executor_id (e.g., 'hostname:port') instead of UUID for easier identification
- Executors still connect to scheduler via --scheduler-address to pull Spicepod

This enables running multiple stateless scheduler replicas behind a load balancer, with each scheduler independently discovering the same set of executors.

## 🔗 Related

Part of #8559

### DIFF
--- a/crates/runtime/src/cluster/discovery.rs
+++ b/crates/runtime/src/cluster/discovery.rs
@@ -1,0 +1,324 @@
+/*
+Copyright 2025 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//! Executor discovery for scheduler-driven cluster mode.
+//!
+//! This module implements DNS SRV-based executor discovery, allowing schedulers to
+//! automatically discover and register executors from Kubernetes headless services.
+
+use std::collections::HashSet;
+use std::sync::Arc;
+use std::time::Duration;
+
+use ballista_core::serde::scheduler::{ExecutorData, ExecutorMetadata};
+use ns_lookup::{SrvRecord, lookup_srv};
+use runtime_proto::DescribeExecutorRequest;
+use runtime_proto::executor_service_client::ExecutorServiceClient;
+use tokio_util::sync::CancellationToken;
+use tonic::transport::{ClientTlsConfig, Endpoint};
+use util::fibonacci_backoff::{Backoff, FibonacciBackoffBuilder};
+
+use crate::Runtime;
+
+/// Maximum interval between executor discovery attempts (steady state).
+const MAX_DISCOVERY_INTERVAL: Duration = Duration::from_secs(30);
+
+/// Timeout for connecting to an executor during discovery.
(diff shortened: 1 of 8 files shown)


crates/runtime/src/cluster/service.rs
```diff
@@ -117,3 +121,66 @@ impl ClusterService for ClusterServiceImpl {
...
+
+        let Some(ref executor) = *executor_guard else {
+            return Err(Status::unavailable(
+                "Executor registration not yet available",
+            ));
+        };
+
+        let registration = &executor.metadata;
+
+        let task_slots = registration
```
sgrebnov, line 163: Is it expected that `registration.specification` can be None and we should default to 0 slots? We might want to add a warning in this case if this is not expected

crates/runtime/src/cluster/discovery.rs
```diff
@@ -0,0 +1,324 @@
...
+        Ok(ch) => ch,
+        Err(e) => {
+            tracing::debug!("Failed to connect to {}: {}", endpoint_url, e);
+            return Ok(None);
+        }
+    };
+
+    let mut client = ExecutorServiceClient::new(channel);
+
+    let response = match client.describe_executor(DescribeExecutorRequest {}).await {
```
sgrebnov, line 296: Nit: we might add a timeout here

sgrebnov (approved):

crates/runtime/src/cluster/discovery.rs
```diff
@@ -0,0 +1,324 @@
...
+async fn discover_and_register_executors(
+    rt: &Runtime,
+    srv_name: &str,
+    client_tls_config: Option<&ClientTlsConfig>,
+    tls_enabled: bool,
+    known_executors: &mut HashSet<String>,
+) -> crate::Result<usize> {
+    tracing::debug!("Performing SRV lookup for: {}", srv_name);
+
+    let srv_records = match lookup_srv(srv_name).await {
```
sgrebnov, line 137: Nit: we might add few retries here, otherwise we will need to wait for the next cycle (up to 30s)

Jeadie:
What's the local & non-K8 experience like for this? IIRC We're not only adding DNS SRV executor discovery, but removing executors ability to self-register. My main worry is that this now all becomes burdensome for all non-k8 experiences.

phillipleblanc:
Luke and I decided on a different approach for this.

PR #8748, opened 2026-01-02
Switch executor scheduling to push-staged with scheduler polling and shared executor gRPC service

phillipleblanc:
## 📝 Summary

- Depends on: https://github.com/spiceai/spiceai/pull/8746

Switches the Ballista scheduling mode from pull based to push based, which is more compatible with an HA scheduler configuration.

- Brings in the following Ballista change: https://github.com/spiceai/datafusion-ballista/pull/5

With this change multiple schedulers can now push work to executors independently.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -337,9 +337,9 @@ datafusion-table-providers = { git = "https://github.com/datafusion-contrib/data
 
 datafusion-federation = { git = "https://github.com/spiceai/datafusion-federation.git", rev = "afcbd1ce99703f1322d176fc1b99745f647234d0" } # spiceai:spiceai-50
 
-ballista-core = { git = "https://github.com/spiceai/datafusion-ballista", rev = "9c17685db62a1511ecf2e90aa72b19feabdea1cb" }           # spiceai:spiceai-50
-ballista-executor = { git = "https://github.com/spiceai/datafusion-ballista", rev = "9c17685db62a1511ecf2e90aa72b19feabdea1cb" }   # spiceai:spiceai-50
-ballista-scheduler = { git = "https://github.com/spiceai/datafusion-ballista", rev = "9c17685db62a1511ecf2e90aa72b19feabdea1cb" } # spiceai:spiceai-50
+ballista-core = { git = "https://github.com/spiceai/datafusion-ballista", rev = "f13a56adea886e60c438585883b1fe404e5666e3" }           # spiceai:spiceai-50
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -337,9 +337,9 @@ datafusion-table-providers = { git = "https://github.com/datafusion-contrib/data
 
 datafusion-federation = { git = "https://github.com/spiceai/datafusion-federation.git", rev = "afcbd1ce99703f1322d176fc1b99745f647234d0" } # spiceai:spiceai-50
 
-ballista-core = { git = "https://github.com/spiceai/datafusion-ballista", rev = "9c17685db62a1511ecf2e90aa72b19feabdea1cb" }           # spiceai:spiceai-50
-ballista-executor = { git = "https://github.com/spiceai/datafusion-ballista", rev = "9c17685db62a1511ecf2e90aa72b19feabdea1cb" }   # spiceai:spiceai-50
-ballista-scheduler = { git = "https://github.com/spiceai/datafusion-ballista", rev = "9c17685db62a1511ecf2e90aa72b19feabdea1cb" } # spiceai:spiceai-50
+ballista-core = { git = "https://github.com/spiceai/datafusion-ballista", rev = "f13a56adea886e60c438585883b1fe404e5666e3" }           # spiceai:spiceai-50
```
phillipleblanc, line 340: Wait for https://github.com/spiceai/datafusion-ballista/pull/5 to merge and then update this commit

crates/runtime/src/cluster/polling.rs
```diff
@@ -0,0 +1,178 @@
...
+
+        let executors = match scheduler.state.executor_manager.get_executor_state().await {
+            Ok(executors) => executors,
+            Err(e) => {
+                tracing::warn!("Executor polling: failed to list executors: {e}");
+                continue;
+            }
+        };
+
+        for (metadata, _last_seen) in executors {
```
sgrebnov, line 70: Nit: we might want to poll executors concurrently + add timeout to `client.clone().poll_executor`

sgrebnov, line 70: Removed executors seem to still in the clients, do we need something like this after `let executors = ..`?

clients.retain(|id, _| executors.iter().any(|(m, _)| &m.id == id));

sgrebnov (approved):

phillipleblanc:
Luke and I decided on a different approach for this

PR #8750, opened 2026-01-02
Implements executor-owned leasing of task slots

phillipleblanc:
## 📝 Summary

In the upstream Ballista model, the scheduler is responsible for owning the state of how busy the executors are. In an HA model, a given scheduler doesn't have the complete picture of how busy an executor is. Add RPCs to allow them to lease task slots and enforce them on the executor side.

Implements executor-side lease management for task slot reservation in cluster mode. This enables schedulers to reserve capacity on executors before dispatching tasks, preventing over-subscription and improving load balancing.

Key components:
- LeaseManager: Executor-owned state for tracking reserved slots with TTL-based expiration, idempotent reservations via request_id, and automatic cleanup of expired leases
- ExecutorService gRPC service: Exposes DescribeExecutor and PollExecutor RPCs for scheduler-driven discovery and status polling
- ExecutorLeaseService gRPC service: Exposes ReserveSlots, RenewLease, ReleaseLease, and GetCapacity RPCs for slot reservation
- LeaseValidatingExecutorGrpc: Wrapper that validates lease ownership before task execution

Design properties:

- Executor is the source of truth for capacity
- Leases expire automatically if not renewed
- Partial grants allowed (request 10 slots, get 5)
- Slots tracked at two levels: reserved (leased) vs in-use (executing)

### DIFF
--- a/crates/runtime-proto/proto/spice.proto
+++ b/crates/runtime-proto/proto/spice.proto
@@ -27,6 +27,19 @@ service ExecutorService {
     rpc PollExecutor(PollExecutorRequest) returns (PollExecutorResponse);
 }
 
+// Executor lease service for task slot reservation.
+// Exposed by executors so schedulers can reserve capacity before dispatching tasks.
+service ExecutorLeaseService {
+    // Reserve task slots on this executor. Idempotent via request_id.
+    rpc ReserveSlots(ReserveSlotsRequest) returns (ReserveSlotsResponse);
+    // Renew an existing lease to extend its TTL.
+    rpc RenewLease(RenewLeaseRequest) returns (RenewLeaseResponse);
+    // Release a lease, returning slots to the available pool.
+    rpc ReleaseLease(ReleaseLeaseRequest) returns (ReleaseLeaseResponse);
+    // Get the executor's current capacity (total and available slots).
+    rpc GetCapacity(GetCapacityRequest) returns (GetCapacityResponse);
+}
+
 // Request message for DescribeExecutor RPC.
 message DescribeExecutorRequest {}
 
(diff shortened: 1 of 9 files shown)


crates/runtime-proto/proto/spice.proto
```diff
@@ -87,6 +100,83 @@ message ExpandSecretResponse {
...
+    // Whether the renewal was successful.
+    bool renewed = 1;
+    // New TTL in milliseconds (from now).
+    uint64 ttl_ms = 2;
+    // Reason for rejection if not renewed.
+    string rejection_reason = 3;
+}
+
+// Request message for ReleaseLease RPC.
+message ReleaseLeaseRequest {
```
Jeadie, line 150: Is there any value in being able to release some lease capacity and not just binary 0/100%?

crates/runtime/src/datafusion/builder.rs
```diff
@@ -378,6 +378,7 @@ impl DataFusionBuilder {
             scheduler_server: RwLock::new(None),
             executor: RwLock::new(None),
             executor_grpc: RwLock::new(None),
+            lease_manager: RwLock::new(None),
```
Jeadie, line 381: I wonder if we should group all multi-node fields into a new struct? (i.e. scheduler_server, executor, executor_grpc, lease_manager).

crates/runtime/src/cluster/lease.rs
```diff
@@ -0,0 +1,823 @@
...
+use uuid::Uuid;
+
+/// TTL for `request_id` idempotency cache entries (60 seconds).
+const REQUEST_ID_CACHE_TTL: Duration = Duration::from_secs(60);
+
+/// Default lease TTL if not specified (30 seconds).
+const DEFAULT_LEASE_TTL: Duration = Duration::from_secs(30);
+
+/// Maximum allowed lease TTL (5 minutes).
+const MAX_LEASE_TTL: Duration = Duration::from_secs(300);
```
Jeadie, line 41: `5m` doesn't seem very long for long-running analytics? I assume the default behaviour is we expect schedulers to renew leases during a running task?

crates/runtime/src/cluster/lease.rs
```diff
@@ -0,0 +1,823 @@
...
+        // Validate and clamp TTL
+        let ttl = if ttl_ms == 0 {
+            DEFAULT_LEASE_TTL
+        } else {
+            Duration::from_millis(ttl_ms).clamp(MIN_LEASE_TTL, MAX_LEASE_TTL)
+        };
+
+        // Check available capacity
+        let available = state.available_slots();
+        if slots == 0 {
```
Jeadie, line 385: might be worth nooping and  not failing on slots==0. If im a scheduler and have determined slots per executor and blindly send requests, it will be easier if I don't have to think about slots==0.

crates/runtime/src/cluster/lease.rs
phillipleblanc: Yeah, they should renew. 5m is honestly too long - if a scheduler dies we would want to reclaim that capacity faster. But we can start with this for now and iterate as we see how it works.

crates/runtime/src/datafusion/builder.rs
phillipleblanc: Yeah that makes sense

crates/runtime-proto/proto/spice.proto
phillipleblanc: I don't think so, at least not initially.

Issue #8755, opened 2026-01-03
TPC-H SF100 benchmarks fail due to Q11 query produces 0 rows


sgrebnov:
## Describe the bug

Sf100 benchmarks are failing due to TPC-H SF100 Q11 query produces 0 rows. 

This is expected as we have `[FRACTION]` parameter hardcoded to `0.0001`. See similar issue created in DuckDB (there is also link to TPC-H spec that instruct parameter to be `FRACTION is chosen as 0.0001 / SF`)
https://github.com/duckdb/duckdb/issues/17965

Also, same result when running the same query using official Snowflake SF100 dataset

<img width="686" height="691" alt="Image" src="https://github.com/user-attachments/assets/22b09362-64c3-4823-9d25-eb4966892495" />

PR #8756, opened 2026-01-03
Benchmarks: make row count validation skip logic configurable by scale factor, query set, and overrides

sgrebnov:
## 📝 Summary

When running TPC-H benchmarks at scale factors other than SF=1, certain queries (e.g., `tpch_q11`) legitimately return 0 rows because their HAVING clause thresholds are calibrated for SF=1. This caused false benchmark failures for SF100.

PR centralized mechanism for determining which queries should skip row count validation based on scale factor, query set, and overrides

## 🔗 Related

Fixes 
- https://github.com/spiceai/spiceai/issues/8755


crates/test-framework/src/queries/mod.rs
```diff
@@ -493,6 +493,39 @@ impl QuerySet {
             _ => Ok(None),
         }
     }
+
+    /// Returns query names that should be skipped for row count validation.
+    #[must_use]
+    pub fn get_row_count_validation_skip_queries(
+        &self,
+        _overrides: Option<QueryOverrides>,
```
sgrebnov, line 501: `_overrides` will be used next for TPCDS  and specific query overrides

krinart (approved):

PR #8789, opened 2026-01-05
Fix `table_allowlist` for table sampling and NSQL

Jeadie:
## 📝 Summary
 - Prior work, #8714, did not implement table allowlisting for sampling tools and on `POST v1/nsql`. 
 - This PR ensures `models.datasets` affects these two functions. For NSQL this means across all: table schema tool calls, table sample tool calls and underlying query operations.


sgrebnov (approved):

crates/runtime/src/http/v1/nsql.rs
```diff
@@ -440,3 +505,40 @@ pub(crate) async fn handle_nsql_query(
         }
     }
 }
+
+/// Construct a [`ResolvedTableAwareAllowlist`] based on the `App`'s `model.datasets`.
+async fn table_allowlist(
```
sgrebnov, line 510: Nit: we can add some unit + integration tests for this logic

lukekim (approved):

crates/runtime/src/http/v1/nsql.rs
Jeadie: We have similar testing in `runtime_datafusion::allowlist`

PR #8852, opened 2026-01-09
feat: Initial HA schedulers distributed query implementation

phillipleblanc:
## Summary

Multiple schedulers may now participate in a distributed query cluster.
In this mode, schedulers need a new configuration in `runtime` to
specify the shared object store state:

```yaml
runtime:
  scheduler:
    state_location: s3://spiceai-phillip-cluster-state-seoul/251214-distributed-query/
    params:
      s3_key: ${secrets:S3_KEY}
      s3_secret: ${secrets:S3_SECRET}
      s3_region: ap-northeast-2
```

When that is configured, the schedulers will write their own advertise
address into `$STATE_LOCATION/schedulers/<adversise_address>.json`

The schedulers also have a new gRPC method: `GetSchedulers` that
executors will call on startup to discover the scheduler list, and then
establish separate poll loop routines against each scheduler
independently.

As schedulers are added/removed to the cluster, the schedulers will
discover this information based on the object store state and the
executors will poll on `GetSchedulers` to get the updated list and
adjust their own work polling.

The work poll loops by the executors are intentionally simple for now,
future work will happen to allow them to share task slots so that they
don't ask for more work than they can handle.


crates/runtime-acceleration/src/snapshot/mod.rs
```diff
@@ -1326,103 +1325,25 @@ static S3_PARAMETERS: LazyLock<Vec<ParameterSpec>> = LazyLock::new(|| {
...
-                );
-            }
-            Err(err) => {
-                tracing::warn!("Unable to initialize AWS credentials for snapshot store: {err}");
-            }
-        }
-    }
-
-    Ok(Box::new(s3_builder.build()?))
+    S3ObjectStoreBuilder::from_url(snapshots_url, io_runtime)
```
phillipleblanc, line 1340: This logic was duplicated between the new scheduler object store and the acceleration snapshots object store configuration, so I consolidated them as part of this PR.

crates/runtime/src/cluster/mod.rs
```diff
@@ -617,14 +860,60 @@ pub async fn initialize_cluster_executor(
...
+        loop {
+            refresh.tick().await;
+            if let Some(addresses) = fetch_scheduler_membership(
+                &scheduler_url_for_manager,
+                client_tls_config_for_manager.clone(),
+            )
+            .await
+            {
+                if addresses.is_empty() {
+                    tracing::warn!(
```
peasee, line 900: Isn't this unreachable? `get_schedulers` always returns at least itself.

Maybe not throw a specific panic, but add a comment here why we're not trusting the response.

crates/runtime/src/cluster/scheduler_registry.rs
```diff
@@ -0,0 +1,526 @@
...
+                    self.delete_record().await;
+                    break;
+                }
+                _ = heartbeat.tick() => {
+                    if let Err(err) = self.heartbeat().await {
+                        tracing::warn!("Scheduler heartbeat failed: {err}");
+                    }
+                }
+                _ = discovery.tick() => {
+                    if let Err(err) = self.refresh_peers().await {
```
peasee, line 219: could discovery block a heartbeat from occurring for longer than the heartbeat interval? a slow object store/connection causing heartbeat for this scheduler to expire.

crates/runtime/src/cluster/mod.rs
phillipleblanc: Correct, just defensive coding.

crates/runtime/src/cluster/scheduler_registry.rs
phillipleblanc: Potentially - I think it is fine for now as if we can't connect to object store we wouldn't be able to update our heartbeat anyway.

peasee (approved):

lukekim (approved):

PR #8867, opened 2026-01-10
Upgrade S3 Vectors SDK and improve test robustness

lukekim:
This pull request improves the handling and testing of vector data in S3 vector queries, updates dependencies, and modernizes test configurations for embedding models. The main changes include refactoring how vector data is merged and serialized, enhancing test coverage for vector output scenarios, updating the S3 vector SDK version, and switching test embedding models to a more consistent configuration.

### S3 Vector Query Refactor and Serialization

* Refactored the merging of vector data in `query_vector_stream` to avoid mutable state and ensure that vector data is only included in results when present, simplifying the logic and improving reliability. The serialization now passes vector data explicitly to `to_flat_value`, which was updated to accept an optional vector and include it only when available. [[1]](diffhunk://#diff-566cc3ffc52b21b88049eb3da43b2ec81475f3d210cd409b10100be0d477a3ffL397-R416) [[2]](diffhunk://#diff-566cc3ffc52b21b88049eb3da43b2ec81475f3d210cd409b10100be0d477a3ffL429-R437) [[3]](diffhunk://#diff-566cc3ffc52b21b88049eb3da43b2ec81475f3d210cd409b10100be0d477a3ffL463-L466)

### Test Coverage and Reliability Improvements

* Added comprehensive unit tests for `to_flat_value` to verify correct serialization of vector data, absence of data, and flattening of metadata fields. These tests ensure robust handling of various output scenarios. [[1]](diffhunk://#diff-566cc3ffc52b21b88049eb3da43b2ec81475f3d210cd409b10100be0d477a3ffR512) [[2]](diffhunk://#diff-566cc3ffc52b21b88049eb3da43b2ec81475f3d210cd409b10100be0d477a3ffR1048-R1151)
* Improved test reliability by using a wider range for random index names to reduce collision risk in parallel test runs.
* Increased test timeout for component loading to 120 seconds, reducing spurious failures in slow environments.

### Dependency Update

* Upgraded `aws-sdk-s3vectors` from version `1.11` to `1.17.0` to ensure compatibility and include recent bug fixes and features.

### Embedding Model Configuration Modernization

[...]

PR #8871, opened 2026-01-11
`setup-make` before `make lint-go`.

Jeadie:
Regression from #8868


phillipleblanc (commented):
This change will also be part of https://github.com/spiceai/spiceai/pull/8867/changes

Jeadie:
> This change will also be part of https://github.com/spiceai/spiceai/pull/8867/changes

True. We should use `uses: ./.github/actions/setup-make` in it

PR #9005, opened 2026-01-20
Optimize PR workflow

lukekim:
This pull request updates the feature check workflow in `.github/workflows/features.yml` to improve efficiency and coverage by splitting feature checks into two matrix jobs and making the exclusion of features and openapi checks configurable per job. The changes streamline CI for feature testing and make it easier to manage feature combinations.

**Workflow matrix and feature check improvements:**

* The `Features Check` job is now split into two matrix entries (`1/2` and `2/2`), each with its own set of excluded features and openapi check configuration, enabling parallel and more targeted feature validation.
* The feature exclusion list for the cargo check step now uses the matrix-defined `exclude_features` variable, making the workflow more flexible and maintainable.

**Conditional openapi check:**

* The openapi feature check step now runs conditionally based on the matrix configuration (`check_openapi`), ensuring it only executes when needed for each job.


.github/workflows/build_and_release.yml
```diff
@@ -136,6 +135,19 @@ jobs:
...
 
+  setup-features:
+    name: Setup features matrix
+    runs-on: spiceai-macos
+    outputs:
+      matrix: ${{ steps.generate-matrix.outputs.matrix }}
+      num_groups: ${{ steps.generate-matrix.outputs.num_groups }}
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
+      - uses: ./.github/actions/setup-rust
```
phillipleblanc, line 146: This setup rust appears to not be required?

.github/workflows/features.yml
```diff
@@ -25,73 +25,123 @@ env:
...
         with:
           fetch-depth: 1
 
-      - name: Set up Rust
-        uses: ./.github/actions/setup-rust
+      - name: Generate feature matrix
+        id: set-matrix
+        run: |
+          # Get all features from spiced package, excluding the ones we don't want to check
+          EXCLUDED="${{ env.EXCLUDED_FEATURES }}"
```
phillipleblanc: Shouldn't this also use the python script instead of re-implementing the logic in bash?

.github/workflows/pr.yml
```diff
@@ -269,15 +215,23 @@ jobs:
           INSTA_UPDATE: ${{ github.event.inputs.update_snapshots || 'no' }}
 
       - name: Build testoperator
-        if: needs.check_changes.outputs.relevant_changes == 'true'
+        if: needs.setup.outputs.relevant_changes == 'true'
         run: cargo build -p testoperator --all-features --locked
 
+      # Step 3: Check all feature combinations (leverages build cache from above)
+      - name: Check features
```
phillipleblanc, line 228: If we are going to run this in the pr check, shouldn't we remove the separate `features` workflow?

phillipleblanc (approved):

PR #9025, opened 2026-01-21
Add more tests for HTTP caching with columns selection

sgrebnov:
## 📝 Summary

Extends existing integration tests to include more cases for HTTP caching with columns selection

### DIFF
--- a/crates/runtime/src/accelerated_table/mod.rs
+++ b/crates/runtime/src/accelerated_table/mod.rs
@@ -1010,10 +1010,10 @@ impl TableProvider for AcceleratedTable {
             }
         }
 
-        // For caching mode, extend projection to include fetched_at for freshness checking if needed.
+        // For caching mode with filters, extend projection to include fetched_at for freshness checking if needed.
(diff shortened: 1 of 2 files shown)


crates/runtime/src/accelerated_table/mod.rs
```diff
@@ -1010,10 +1010,10 @@ impl TableProvider for AcceleratedTable {
             }
         }
 
-        // For caching mode, extend projection to include fetched_at for freshness checking if needed.
+        // For caching mode with filters, extend projection to include fetched_at for freshness checking if needed.
         // Added columns will be automatically stripped by `SchemaCastScanExec`, similar to
         // fallback-to-source on cache miss where results return all columns.
-        let extended_projection = if is_caching_mode {
+        let extended_projection = if is_caching_mode && !filters.is_empty() {
```
sgrebnov, line 1016: There is no need for `fetched_at` if there are no filters => returning data from accelerator w/o http query or expiration validation

lukekim (approved):

PR #9063, opened 2026-01-22
Remove models variant (now default) & Windows builds (use WSL)

lukekim:
- Make PostgreSQL acceleration optional


.github/workflows/spiced_docker_nightly.yml
```diff
@@ -143,20 +143,25 @@ jobs:
             REL_VERSION=${{ needs.setup.outputs.rel_version }}
             CARGO_FEATURES=${{ env.CARGO_FEATURES }}
             RUST_PROFILE=${{ env.RUST_PROFILE }}
-      - name: models-jemalloc
+      - name: data
         run: |
-          echo "IMAGE_TAG=models-jemalloc" >> $GITHUB_ENV
-          echo "CARGO_FEATURES=release,models,alloc-jemalloc" >> $GITHUB_ENV
+          echo "IMAGE_TAG=data" >> $GITHUB_ENV
+          echo "CARGO_FEATURES=release,postgres-accel" >> $GITHUB_ENV
```
phillipleblanc: why is postgres-accel here?

phillipleblanc: Also because `models` is in the default features now, this will still build with the models feature. You will need to specify `--no-default-features` and re-add all features except the models one.

bin/spiced/Cargo.toml
```diff
@@ -106,6 +107,7 @@ odbc = ["runtime/odbc"]
 oracle = ["runtime/oracle"]
 pingora = ["runtime/pingora"]
 postgres = ["runtime/postgres"]
+postgres-accel = ["runtime/postgres-accel"]
```
phillipleblanc, line 110: I don't understand why we need this now?

krinart (approved):

PR #9123, opened 2026-01-27
fix: Use the same vortex dependency as ballista

peasee:
## 📝 Summary

* Specifies our vortex dependency based on the branch source
* Specifies a vortex patch to ensure all dependencies use the same vortex source in the tree

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -301,11 +301,11 @@ url = { version = "2.5", features = ["serde"] }
 util = { path = "crates/util" }
 uuid = "1"
 # We can't use patches for vortex, because their tagged release version .tomls have a version of 0.1.0 not the actual release version
-vortex = { git = "https://github.com/spiceai/vortex", branch = "spiceai-51" }
-vortex-array = { git = "https://github.com/spiceai/vortex", branch = "spiceai-51" }
-vortex-datafusion = { git = "https://github.com/spiceai/vortex", branch = "spiceai-51" }
-vortex-session = { git = "https://github.com/spiceai/vortex", branch = "spiceai-51" }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -301,11 +301,11 @@ url = { version = "2.5", features = ["serde"] }
 util = { path = "crates/util" }
 uuid = "1"
 # We can't use patches for vortex, because their tagged release version .tomls have a version of 0.1.0 not the actual release version
-vortex = { git = "https://github.com/spiceai/vortex.git", rev = "4e9896bc9e080dc5a133e34099cf68277a8782a7" } # branch: spiceai-51
-vortex-array = { git = "https://github.com/spiceai/vortex.git", rev = "4e9896bc9e080dc5a133e34099cf68277a8782a7" } # branch: spiceai-51
-vortex-datafusion = { git = "https://github.com/spiceai/vortex.git", rev = "4e9896bc9e080dc5a133e34099cf68277a8782a7" } # branch: spiceai-51
-vortex-session = { git = "https://github.com/spiceai/vortex.git", rev = "4e9896bc9e080dc5a133e34099cf68277a8782a7" } # branch: spiceai-51
-vortex-utils = { git = "https://github.com/spiceai/vortex.git", rev = "4e9896bc9e080dc5a133e34099cf68277a8782a7" } # branch: spiceai-51
+vortex = { git = "https://github.com/spiceai/vortex", branch = "spiceai-51" }
```
phillipleblanc: We've avoided using branch based revs here because they wouldn't get updated.

peasee: This is what `datafusion-ballista` uses

peasee: I updated the patch and main here back to `rev =`

phillipleblanc (approved):

PR #9141, opened 2026-01-27
S3Vector client level tracing and simplify retry

Jeadie:
## 📝 Summary
 - Add basic task_history tracing for s3vector client.
 - Standardise retry logic
 - Pass S3Vector inputs by reference until base client.


crates/s3_vectors/src/lib.rs
```diff
@@ -82,129 +82,129 @@ pub mod mock;
 impl S3Vectors for Client {
     async fn create_index(
         &self,
-        input: CreateIndexInput,
+        input: &CreateIndexInput,
```
phillipleblanc, line 85: Why this change to take a reference and then immediately clone it?

phillipleblanc (approved):

crates/s3_vectors/src/lib.rs
Jeadie: So that all other _"middleware"_ implentations of `S3Vectors` can avoid `.clone()` when they pass to their inner `S3Vectors`.

Issue #9103, opened 2026-01-26
Cayenne `file[parquet]-cayenne[file]-append_upsert.yaml` invalid row count


sgrebnov:
## Describe the bug

This might indicate Cayenne upsert does not work correct.

https://github.com/spiceai/spiceai/actions/runs/21355992848/job/61463428043

>Error: Table row counts do not match expected values
Table customer has 255000 rows, expected 150000
Table lineitem has 10202337 rows, expected 6001215
Table nation has 41 rows, expected 25
Table orders has 2550000 rows, expected 1500000
Table part has 340000 rows, expected 200000
Table partsupp has 1360000 rows, expected 800000
Table region has 7 rows, expected 5
Table supplier has 17000 rows, expected 10000

sgrebnov:
Can also be reproed using the following example

```
id,name,value,updated_at
1,alpha,100,2026-06-20T10:00:00+00:00
2,beta,200,2026-06-20T10:00:00+00:00
3,gamma,300,2026-05-20T10:00:00+00:00
```

```
datasets:
  - from: file:./data.csv
    name: test_data
    time_column: updated_at
    time_format: timestamp
    acceleration:
      enabled: true
      engine: cayenne
      mode: file
      refresh_mode: append
      refresh_check_interval: 3s
      primary_key: id
      on_conflict:
        id: upsert
```

```
sql> select * from test_data;
+----+-------+-------+---------------------+
| id | name  | value | updated_at          |
+----+-------+-------+---------------------+
| 1  | alpha | 100   | 2026-04-20T10:00:00 |
| 2  | beta  | 200   | 2026-04-20T10:00:00 |
| 3  | gamma | 300   | 2026-04-20T10:00:00 |
+----+-------+-------+---------------------+

Time: 0.002926875 seconds. 3 rows.
sql> select * from test_data;
+----+-------+-------+---------------------+
| id | name  | value | updated_at          |
+----+-------+-------+---------------------+
| 1  | alpha | 100   | 2026-04-20T10:00:00 |
| 2  | beta  | 200   | 2026-04-20T10:00:00 |
| 3  | gamma | 300   | 2026-04-20T10:00:00 |
| 1  | alpha | 100   | 2026-06-20T10:00:00 |
| 2  | beta  | 200   | 2026-06-20T10:00:00 |
| 3  | gamma | 300   | 2026-05-20T10:00:00 |
+----+-------+-------+---------------------+

```

sgrebnov:
[...]

PR #9152, opened 2026-01-27
fix(cayenne): Fix upsert with pending deletions causing duplicate PKs

sgrebnov:
## 📝 Summary

Fixes bug #9103 where upsert operations would create duplicate primary keys when pending deletions existed. 
1. **Bypassed upsert validation** - When `has_pending_deletions=true`, the code path skipped `validate_on_conflict()` entirely, writing incoming rows without checking for existing PKs.

2. **Incorrect visibility check** - `load_existing_keyset()` only checked the deletion cache, not the insert_records cache. This caused re-inserted PKs (from prior upserts) to be incorrectly filtered out.

## Solution

- Added `validate_on_conflict()` call in the `has_pending_deletions` path
- Fixed visibility logic to check both deletion and insert_records caches using sequence-based ordering
- Extracted shared `is_pk_visible_i64()` and `is_pk_visible_row_key()` helper functions
- Consolidated duplicated code into `prepare_stream_for_insert()` helper

## 🔗 Related

Fixes
 - https://github.com/spiceai/spiceai/issues/9103

### DIFF
--- a/crates/cayenne/src/provider/table.rs
+++ b/crates/cayenne/src/provider/table.rs
@@ -23,8 +23,8 @@ use super::constants::{
     DEFAULT_DATA_FILE_ID, DELETION_CACHE_LOCK_POISONED, LISTING_TABLE_LOCK_POISONED,
 };
 use super::delete::{
-    read_deletion_vectors, CayenneDeletionSink, DeletionFilterExec, Int64PkDeletionFilterExec,
-    KeyBasedDeletionFilterExec,
+    is_pk_visible_i64, is_pk_visible_row_key, read_deletion_vectors, CayenneDeletionSink,
+    DeletionFilterExec, Int64PkDeletionFilterExec, KeyBasedDeletionFilterExec,
 };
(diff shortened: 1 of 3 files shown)


crates/cayenne/src/provider/table.rs
```diff
@@ -2101,6 +2139,57 @@ impl CayenneTableProvider {
...
 
+    /// Prepare an incoming stream for insert by validating on_conflict constraints.
+    ///
+    /// If a primary key is configured, this method:
+    /// 1. Loads existing keys from the table (respecting deletion visibility)
+    /// 2. Validates incoming rows against on_conflict behavior (drop/upsert)
+    /// 3. Returns a prepared stream with conflicts resolved and deletion specs
+    ///
+    /// If no primary key is configured, returns the stream unchanged with empty deletion specs.
+    async fn prepare_stream_for_insert(
```
sgrebnov, line 2150: Refactoring  so this logic can be re-used

phillipleblanc (approved):

Issue #9105, opened 2026-01-26
Cayenne `cayenne[file]-append_retention_sql.yaml` `Client specified an invalid argument` error


sgrebnov:
## Describe the bug

https://github.com/spiceai/spiceai/actions/runs/21355921090/job/61463198170

>Client specified an invalid argument
>Error: Table row counts do not match expected values

sgrebnov:
Row count issue:

Position-based deletions (used for tables without a primary key) track deleted rows by their **position offset** within the scan. However, **file scan order is non-deterministic** in DataFusion/Vortex, causing the deletion bitmap to delete **wrong rows** when files are scanned in a different order.

### Root Cause

The position-based deletion strategy assumes stable row ordering:

```
┌─────────────────────────────────────────────────────────────────────────────┐
│  POSITION-BASED DELETION ASSUMPTION                                          │
├─────────────────────────────────────────────────────────────────────────────┤
│                                                                             │
│  Retention delete:                                                          │
│    Scans files: [file_A.vortex, file_B.vortex, file_C.vortex]              │
│    Position 0-1000 = file_A rows                                           │
│    Position 1001-2000 = file_B rows                                        │
│    Position 2001-3000 = file_C rows                                        │
│    Deletes positions: [500, 501, 502, ...]  (rows matching filter)         │
│                                                                             │
│  Compaction scan (DIFFERENT file order):                                    │
│    Scans files: [file_B.vortex, file_C.vortex, file_A.vortex]  ← REORDERED │
│    Position 0-1000 = file_B rows   ← WRONG!                                │
│    Position 1001-2000 = file_C rows                                        │
│    Position 2001-3000 = file_A rows                                        │
│    Filters out positions: [500, 501, 502, ...]  ← DELETES WRONG ROWS!      │
│                                                                             │
└─────────────────────────────────────────────────────────────────────────────┘
```

Issue #9104, opened 2026-01-26
Cayenne `cayenne[file]-append_retention_period.yaml` `Client specified an invalid argument` error


sgrebnov:
## Describe the bug

https://github.com/spiceai/spiceai/actions/runs/21355884357

>Client specified an invalid argument

>Error: Table row counts do not match expected values

PR #9182, opened 2026-01-28
fix(cayenne): Fix data loss by preserving protected snapshots during cleanup

sgrebnov:
## 📝 Summary

Snapshot cleanup was deleting protected snapshots, causing data loss and query failures. Protected snapshots contain data written after deletions (upserts) and must be preserved until compaction.

**Symptoms:**
- Missing rows after retention eviction (e.g., lineitem: 4.5M rows instead of 6M)

## Fix

Modified `cleanup_old_snapshots_blocking` and `cleanup_old_snapshots_s3` to preserve protected snapshots alongside the current snapshot during cleanup.

refresh_deletion_cache() previously used load_deletion_vectors() which only supports position-based deletions, causing errors for tables with primary keys. Now uses load_deletion_vectors_all() to support all deletion strategies.

## 🔗 Related

Part of 
 - https://github.com/spiceai/spiceai/issues/9105
 - https://github.com/spiceai/spiceai/issues/9104


crates/cayenne/src/provider/table.rs
```diff
@@ -3871,65 +3990,6 @@ impl CayenneTableProvider {
...
-    /// severely degrade query performance and indicate poor table health. Compaction removes
-    /// deleted rows and clears deletion vectors.
-    ///
-    /// # Performance Notes
-    ///
-    /// - Queries metastore once via catalog
-    /// - Reads deletion vector files in a blocking task
-    /// - Result is cached in the table provider to avoid repeated queries on every scan
-    /// - `RoaringBitmap` provides 50-90% memory savings vs `HashSet` for sparse deletions
-    async fn load_deletion_vectors(
```
sgrebnov, line 3893: Removed in favor of `load_deletion_vectors_all` that already exists and used

lukekim (approved):

Issue #9203, opened 2026-01-29
In distributed query, Handle schedulers updating accelerated table partition metadata.


Jeadie:

Jeadie:
Done in #9378

PR #9222, opened 2026-01-30
New `object_store_occ`crate and use in SchedulerRegistryRunner 

Jeadie:
## 📝 Summary
 - Standardise how we do optimistic concurrency control on object store  with new crate (object_store_occ). This uses conditional writes via ETags If-Match headers to allow for concurrent writes without coordination. 
 - Use this new crate to simplify how `SchedulerRegistryRunner` store/retrieve/write `SchedulerRecord`. 

## 🔗 Related
 - Prerequisites of #9203  (`SchedulerRegistryRunner` is unrelated to this ticket, but can take advantage of this new crate).


crates/object_store_occ/src/state.rs
```diff
@@ -0,0 +1,575 @@
...
+    }
+
+    /// Insert a new object or update existing with OCC.
+    ///
+    /// If insert fails due to existing object, tries update. If update fails due to not found,
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if serialization fails or the object store operation fails.
+    pub async fn insert_or_update(&self, key: &str, value: &T) -> Result<WriteResult<T>> {
```
phillipleblanc, line 195: This seems not that useful of a function to have - if we try to insert and its not there, its unlikely we have the latest version?

phillipleblanc (dismissed):

crates/object_store_occ/src/state.rs
```diff
@@ -0,0 +1,575 @@
...
+}
+
+/// Manages typed objects in an object store with optimistic concurrency control.
+///
+/// `ObjectState<T>` provides a type-safe interface for storing and retrieving
+/// serializable structs with automatic conflict detection via `ETag`s.
+pub struct ObjectState<T> {
+    store: Arc<dyn ObjectStore>,
+    prefix: String,
+    cache: RwLock<HashMap<String, CachedEntry<T>>>,
```
peasee, line 76: use Moka or something instead of a HashMap for proper TTL/size/eviction?

peasee: I think moka would also remove the lock

crates/object_store_occ/src/state.rs
Jeadie: Correct, but if insert is not there, we will get the new version before attempting update. This is not yet used, but I suspected we'd want this convience for accelerated table partitions. If not will remove.

crates/object_store_occ/src/state.rs
Jeadie: Currently we don't need TTL/size/eviction. Will refactor if/when needed

peasee (approved):

PR #9244, opened 2026-02-01
Simplify retention filter expressions before pushdown

sgrebnov:
## 📝 Summary

Retention failed on Cayenne-accelerated tables because the filter converter always cast timestamps to nanoseconds, regardless of the column's actual time unit. Vortex cannot cast between timestamp types with different time units:

```
Failed to cast array of dtype ext(vortex.timestamp, i64, ExtMetadata([1, 3, 0, 85, 84, 67]))?
to vortex.cast(ext(vortex.timestamp, i64, ExtMetadata([0, 0, 0]))?):
No compute kernel to cast array vortex.ext
```

**Solution**

Apply DataFusion's `ExprSimplifier` to retention filters before deletion. This folds casts, e.g. `l_created_at < CAST( ... AS Timestamp(ns, "UTC"))` -> `l_created_at < TimestampMicrosecond(1769949296123456, Some("UTC"))`)

Result: `l_created_at < TimestampMicrosecond(...)` — a simple comparison matching  Vortex timestamp column precision.

**Other Benefits**
- Faster filter evaluation - no additional conversion
- Better predicate pushdown - storage engines can use indexes/stats directly

## 👀 Notes for Reviewers

We can move `simplify_expr` directly to `TimestampFilterConvert` so it is applied to both refresh and retention automatically, plus it will be automatically applied in unit tests so we can verify all main cases.


crates/runtime/src/accelerated_table/retention.rs
```diff
@@ -117,7 +117,19 @@ impl super::AcceleratedTable {
...
                 };
 
-                tracing::trace!("[retention] Expr {expr:?}");
+                tracing::trace!("[retention] Expr before simplification: {expr:?}");
+
+                let expr = match simplify_expr(expr, &accelerator.schema()) {
+                    Ok(simplified) => simplified,
+                    Err(e) => {
+                        tracing::error!(
+                            "[retention] Failed to simplify retention filter for {dataset_name}: {e}"
```
Jeadie: Improve for end user
```suggestion
                            "[retention] Upon checking retention policy for table '{dataset_name}', an error occurred when attempting to simplify the relevant retention expression '{expr:?}'. Error: {e}"
```

Jeadie (dismissed):

crates/runtime/src/datafusion/expr_utils.rs
```diff
@@ -0,0 +1,137 @@
+/*
+Copyright 2024-2026 The Spice.ai OSS Authors
```
phillipleblanc:
```suggestion
Copyright 2026 The Spice.ai OSS Authors
```

phillipleblanc (dismissed):

Jeadie (approved):

PR #9251, opened 2026-02-02
From v1/sql in scheduler, use accelerated partitions from executor.

Jeadie:
## 📝 Summary
 - New gRPC method for Cluster service: `AllocateInitialPartitions`. 
   ```proto
    // Get an initial allocation of accelerated table partitions.
    // Called by executors when they first register to get initial Accelerated Table partitions.
    // Allows schedulers to subsequently query executors over FlightSQL for accelerated table data.
    rpc AllocateInitialPartitions(AllocateInitialPartitionsRequest) returns (AllocateInitialPartitionsResponse);
   ```
 - New analyzer rule `PartitionedTableScanRewrite`.  For given tables, rewrite `TableScan` as the UNION ALL of a number of partitioned tables. Used in distributed query for schedulers that have accelerated table. Instead of running locally, send to executors (whom have the partition). Like this (note partition filters here are trivial. This PR currently has empty partitions instead for correctness). 
  ```
  explain select * from qs limit 3;
  +---------------+---------------------------------------------------------------------------------------------------------------------------------+
  |   plan_type   |                                                               plan                                                              |
  |    varchar    |                                                             varchar                                                             |
  +---------------+---------------------------------------------------------------------------------------------------------------------------------+
  | logical_plan  | Limit: skip=0, fetch=3                                                                                                          |
  |               |   Union                                                                                                                         |
  |               |     Limit: skip=0, fetch=3                                                                                                      |
  |               |       TableScan: qs projection=[id, question, answer, subject, reference_answer, source], full_filters=[Boolean(true)], fetch=3 |
  |               |     Limit: skip=0, fetch=3                                                                                                      |
  |               |       TableScan: qs projection=[id, question, answer, subject, reference_answer, source], full_filters=[Boolean(true)], fetch=3 |
  | physical_plan | CoalescePartitionsExec: fetch=3                                                                                                 |
  |               |   UnionExec                                                                                                                     |
  |               |     CooperativeExec                                                                                                             |
  |               |       BytesProcessedExec                                                                                                        |
  |               |         FlightSqlExec sql=SELECT id, question, answer, subject, reference_answer, source FROM qs WHERE true LIMIT 3             |
  |               |     CooperativeExec                                                                                                             |
  |               |       BytesProcessedExec                                                                                                        |
  |               |         FlightSqlExec sql=SELECT id, question, answer, subject, reference_answer, source FROM qs WHERE true LIMIT 3             |
  |               |                                                                                                                                 |
  +---------------+---------------------------------------------------------------------------------------------------------------------------------+
  ```
 - Also some refactors to distributed logic in preparing`Runtime` to make it more readable. 
 - Add arbitrary Analyzer rules in `DataFusion::builder`. 
 - 
## 🔗 Related
- Work for #9202


crates/runtime/src/cluster/service.rs
```diff
@@ -529,6 +542,78 @@ impl ClusterService for ClusterServiceImpl {
...
+async fn create_executor_flight_client(
+    endpoint: &str,
+) -> Result<FlightSqlClient, tonic::transport::Error> {
+    let executor_address = if endpoint.starts_with("http://") || endpoint.starts_with("https://") {
+        endpoint.to_string()
+    } else {
+        format!("http://{endpoint}")
+    };
+
+    // TODO support for mTLS certificates per executor.
```
phillipleblanc, line 606: Will this break clusters that have mTLS configured?

phillipleblanc (dismissed):

crates/runtime/src/cluster/service.rs
Jeadie: No, `connect_lazy` means the flightsql will only be connected lazily. This would be just when accelerated tables get called with `v1/sql`. But since we are currently assigning no partitions, accelerated tables will be `EmptyExec`s.

phillipleblanc (dismissed):

phillipleblanc (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #9254, opened 2026-02-02
refactor: Update `Query` to support submitting async queries

### LINKED ISSUE #9224: Update `Datafusion::Query` to submit distributed jobs directly to ballista
Update `Datafusion::Query` to call `scheduler.submit_job` instead of execution the distributed exec plan. Also update the `JobExecutor` to then use `Query::run_distributed` (or whatever the implementation will be named) to execute the distributed queries instead of calling submit job from `JobExecutor`

peasee:
## 📝 Summary

* Adds `Query::submit_distributed` to retrieve an async `QueryHandle`. The job result is not immediately returned, which differs from `run_distributed` which plans, executes and returns the results in the same function call.
* Introduces `QueryHandle` to better handle async queries, and moves most of the result streaming functionality from Ballista into this module.
* Updates `JobExecutor` to use `Query::submit_distributed` instead of submitting and managing the job polling itself.
* Adds parameter support back to async query by ensuring it is supplied to the `QueryBuilder`, making use of all the pre-existing parameterized query support in `Query`

## 🔗 Related

* Closes #9224


crates/runtime/src/datafusion/query/handle.rs
```diff
@@ -0,0 +1,614 @@
...
+
+/// Internal state of a query handle.
+enum QueryHandleState {
+    /// Query was submitted to Ballista and is being executed.
+    Running {
+        /// Reference to the Ballista scheduler server.
+        scheduler: Arc<SchedulerServer<LogicalPlanNode, PhysicalPlanNode>>,
+    },
+    /// Query results were retrieved from cache.
+    Cached {
```
peasee, line 119: This cache state is directly from the SQL results cache - which is to say that the query results are not stored in cache for the job store retrieval from async query API, but they are cached for uploading to the job store for retrieval from async query API.

A future optimization could be to:

1. Ensure job store chunk retrievals are also cached in the SQL results cache, and the cache is checked for each chunk first; and/or,
2. Check if an identical logical plan has been stored in the job store before, and if it has - there is no need to upload this new data again (subject to some TTL, etc).

Neither of these are blocking.

crates/runtime/src/datafusion/query.rs
phillipleblanc: We don't need this method anymore

phillipleblanc (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #9318, opened 2026-02-05
Snapshots Improvements

krinart:
## 📝 Summary

1. Fix Cayenne snapshots - properly map directories after archive extraction
2. Schedule `time_interval` based task based on last snapshot creation time 
3. When returning list of snapshots in `get_snapshot_summary` - only returns snapshots that actually exist
4. Add `compact_metadata` - remove metadata records of snapshots that no longer exist. Performed after snapshot creation in the process of updating metadata
5. Add proper extensions to snapshot files:
    a. Duckdb: `.duckdb`
    b. Cayenne - `.cayenne`
    c. Sqlite - `.sqlite`
    d. Turso - `.turso`

### DIFF
--- a/crates/runtime-acceleration/src/snapshot/mod.rs
+++ b/crates/runtime-acceleration/src/snapshot/mod.rs
@@ -2076,16 +2076,23 @@ impl SnapshotManager {
         let current_snapshot_id = dataset_metadata.current_snapshot_id;
         let dataset_engine = dataset_metadata.engine.clone();
 
-        // Check which snapshots actually exist in the object store (with bounded concurrency)
-        let existence_results: HashMap<u64, bool> =
-            futures::stream::iter(dataset_metadata.snapshots.iter().map(|entry| {
+        // Check which snapshots actually exist in the object store (with bounded concurrency).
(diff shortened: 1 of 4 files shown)


crates/runtime-acceleration/src/snapshot/mod.rs
```diff
@@ -2067,6 +2139,64 @@ impl SnapshotManager {
...
     }
 
+    /// Checks if a snapshot exists in the object store.
+    async fn snapshot_exists(&self, snapshot_uri: &str) -> bool {
+        let Ok(object_path) = self.snapshot_uri_to_object_path(snapshot_uri) else {
+            return false;
+        };
+
+        // Use head() for a lightweight existence check without downloading content
+        self.object_store.head(&object_path).await.is_ok()
```
sgrebnov, line 2149: Do we need retry here to ensure some temp networking problem affects the logic?

sgrebnov (approved):

crates/runtime-acceleration/src/snapshot/mod.rs
krinart: I will create a separate PR with retries for snapshots

PR #9325, opened 2026-02-06
Distributed v1/sql query: Schedulers coordinate assigning partitions for executors, and for use at query time. 

Jeadie:
## 📝 Summary
 - Scheduler nodes setup accelerated table partition metadata on startup.
   - Ensure all tables have `partitions/<tbl_name>.json` files. 
   - On new executor registration, finds current partitions (i.e. already assigned) and assigns more if needed. 
   - Write to object store before returning executor partitions via`AllocateInitialPartitions`. See `crates/runtime/src/cluster/partition/`.
 - Executor nodes store and use accelerated table partitions they receive from scheduler. Update `refresh_sql` accordingly. 

## 🔗 Related
#9203 

## Improvements
 - [ ] How refresh SQL stored in acceleration. Currently stored as `String`. Want to manipulate and validate SQL more robustly (i.e. add/remove filters, validate user-provide SQL).
 - [ ] Typing of partition value maps. Currently `HashMap<String, String>`. 
 - [ ] Improve when we discover acceleration table partitions. Currently we do all tables at startup, but this means waiting/sleeping for tables to be registered (note: not the acceleration). We could/should do after each individual table gets registered.


crates/runtime-proto/proto/spice.proto
```diff
@@ -122,7 +122,7 @@ message AllocateInitialPartitionsRequest {
 }
 
 message StringArray {
-  repeated string items = 1;
+  repeated bytes items = 1;
```
phillipleblanc, line 125: Why this change? Should this be called ByteArray instead?

crates/runtime/src/cluster/partition/manager.rs
```diff
@@ -0,0 +1,245 @@
...
+    #[snafu(display("Partition {partition} not found in table {table}"))]
+    PartitionNotFound { table: String, partition: String },
+
+    #[snafu(display("Concurrent modification detected for table {table}"))]
+    ConcurrentModification { table: String },
+}
+
+pub type Result<T, E = Error> = std::result::Result<T, E>;
+
+static PARTITION_PREFIX: &str = "partitions/";
```
phillipleblanc: I think this should be `accelerations/partitions/` - it isn't a general partition thing, and we can put any other acceleration specific metadata in `accelerations/`

crates/runtime-proto/proto/spice.proto
Jeadie: WIll rename. Changed so that we can use `datafusion_proto::bytes::Serializeable`.

phillipleblanc (approved):

PR #9350, opened 2026-02-09
Improve cayenne to handle upserts more robust during runtime restarts

sgrebnov:
## 📝 Summary

After a runtime restart, Cayenne lost data from earlier upsert rounds. `load_protected_snapshots` computed a single global `max_delete_seq` from ALL deletion vectors and filtered out snapshots with `seq <= max_delete_seq`. With multiple upsert rounds, each round's deletions raised the global max, causing earlier snapshots to be incorrectly dropped — their rows became invisible after restart.

### DIFF
--- a/crates/cayenne/src/cayenne_catalog.rs
+++ b/crates/cayenne/src/cayenne_catalog.rs
@@ -845,7 +845,9 @@ impl MetadataCatalog for CayenneCatalog {
         // Order matters for crash safety:
         // 1. Clear delete files first - they reference the old snapshot's data
         // 2. Clear insert records - they correspond to the cleared delete files
-        // 3. Update snapshot pointer - commits the new snapshot as active
+        // 3. Clear snapshot sequences - protected snapshots are no longer needed
+        //    after compaction since all data is merged into the new snapshot
+        // 4. Update snapshot pointer - commits the new snapshot as active
         //
         // If interrupted between these, the old snapshot remains active with
         // no delete files, which is safe (just loses the pending deletions,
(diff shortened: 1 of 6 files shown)


crates/cayenne/tests/retention_test.rs
```diff
@@ -108,15 +108,15 @@ async fn test_retention_filters_apply_on_insert_impl(
         let row_id_array = batch
             .column(0)
             .as_any()
-            .downcast_ref::<Int64Array>()
+            .downcast_ref::<UInt64Array>()
```
sgrebnov, line 111: The test was failing and `UInt64Array` is the right type for row-based id. Not quite sure how the test previously worked/passed

phillipleblanc (dismissed):

crates/cayenne/src/cayenne_catalog.rs
```diff
@@ -854,6 +856,7 @@ impl MetadataCatalog for CayenneCatalog {
             "BEGIN TRANSACTION; \
              DELETE FROM cayenne_delete_file WHERE table_id = {table_id}; \
              DELETE FROM cayenne_insert_record WHERE table_id = {table_id}; \
+             DELETE FROM cayenne_snapshot_sequence WHERE table_id = {table_id}; \
```
sgrebnov, line 859: There's currently no real scenario where snapshot sequence entries exist when commit_compaction runs as it is only for full overwrite scenario and a single dataset doesn't mix Full and Append refresh modes. The DELETE is defensive/future-proofing improvement

krinart (approved):

Issue #8938, opened 2026-01-16
Enhancement: Distributed Query: Async HTTP API


phillipleblanc:
## Goal-State/What/Result

An async query API that enables long-running queries to execute in a distributed cluster via Ballista, with query status and results retrievable asynchronously via HTTP endpoints.

## Why/Purpose

To enable highly-available distributed query across multiple schedulers - an async query API is required to coordinate results and query execution across multiple schedulers.

## By When

**Issue/Spec written and reviewed:** Jan 22nd, 2026
**Done-Done:** Feb 10th, 2026

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Specification

The Runtime will include a gated set of HTTP endpoints for asynchronous query APIs. When cluster mode is not enabled, the endpoints shall return a HTTP error 503 (Service Unavailable)

When queries are submitted via the async API:

* They will be validated and submitted to the Ballista scheduler for distributed execution
* Query results will be stored in the shared object store location upon completion
* Query results will be retrieved from the shared object store for HTTP responses
* Query results will be removed from the shared object store location after a defined TTL

The Rust SDK should support running queries in async mode, and should be used to provide async query support to the `spice query` CLI command.

### HTTP API

| Endpoint | Description |
|----------|--------|
| `POST /v1/queries` | Submits to Ballista scheduler instead of local execution |
| `GET /v1/queries/{id}/status` | Returns Ballista job status |
| `GET /v1/queries/{id}/results` | Returns results from object store |
| `POST /v1/queries/{id}/cancel` | Propagates cancellation to Ballista |

### Out of Scope

* Recovering queries from failed schedulers - the scheduler that receives the async query will execute it. A future enhancement could support multi-scheduler query failover.

### Security Review

[...]

PR #9355, opened 2026-02-10
fix: Provide a more informative 404 on async API when result has no chunks

peasee:
## 📝 Summary

* When a query has no rows, it has no chunks. Previously, retrieving the chunk would just return an uninformative error about a missing chunk - now, it returns a better error that no chunks are available because there were no rows from the query.

## 🔗 Related

* Part of #8938

### DIFF
--- a/crates/runtime/src/flight/async_actions.rs
+++ b/crates/runtime/src/flight/async_actions.rs
@@ -327,7 +327,10 @@ pub async fn handle_get_async_query_result(body: &[u8]) -> Result<Vec<u8>, Statu
     let batches = executor
         .get_chunk(request.query_id.as_str(), request.chunk_index)
         .await
-        .map_err(|e| Status::internal(format!("Failed to get result chunk: {e}")))?;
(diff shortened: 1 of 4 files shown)


phillipleblanc (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

crates/runtime/src/flight/async_actions.rs
```diff
@@ -327,7 +327,14 @@ pub async fn handle_get_async_query_result(body: &[u8]) -> Result<Vec<u8>, Statu
     let batches = executor
         .get_chunk(request.query_id.as_str(), request.chunk_index)
         .await
-        .map_err(|e| Status::internal(format!("Failed to get result chunk: {e}")))?;
+        .map_err(|e| match &e {
```
Jeadie, line 330: nit: this logic could be function on `crate::jobs::Error` crate.

Jeadie (approved):

Issue #8777, opened 2026-01-05
Enhancement: Cayenne Data Accelerator to RC


phillipleblanc:
## Goal-State/What/Result

The Cayenne Data Accelerator is released at RC quality

## Why/Purpose

To provide a well-tested, high performance, highly scalable accelerator

## By When

**Issue/Spec written and reviewed:** Jan 30th, 2026
**Done-Done:** Feb 5th, 2026

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [x] PM/Design Review
- [x] DX/UX Review
- [x] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [x] Tests
- [x] Telemetry / Metrics / Task History
- [x] Performance / Benchmarks
- [x] Documentation
- [x] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## RC Release Criteria

The RC release criteria expand on and require that all [Beta release criteria](./beta.md) continue to pass for the accelerator.

- [x] All [Beta release criteria](./beta.md) pass for this accelerator.

### Test Coverage

RC quality accelerators should be able to run test packages derived from the following:

- [TPC-H](https://www.tpc.org/TPC-H/)
- [TPC-DS](https://www.tpc.org/TPC-DS/)
- [ClickBench](https://github.com/ClickHouse/ClickBench)

Indexes are not required for test coverage, but can be introduced if required for tests to pass (e.g. due to performance characteristics, etc).

#### TPC-H

- [x] The accelerator can load TPC-H SF10 in all [Access Modes](../definitions.md), and can run all queries with no [Major or Minor Bugs](../definitions.md).
- [x] The accelerator can load TPC-H SF100 in either [file or database mode](../definitions.md), and can run all queries with no [Major or Minor Bugs](../definitions.md).

#### TPC-DS

- [x] The accelerator can load TPC-DS SF10 in all [Access Modes](../definitions.md), and can run all queries with no [Major or Minor Bugs](../definitions.md).
- [x] The accelerator can load TPC-DS SF100 in either [file or database mode](../definitions.md), and can run all queries with no [Major or Minor Bugs](../definitions.md).

#### ClickBench

[...]

Issue #8536, opened 2025-12-11
Make Cayenne metadata database connection management more robust


peasee:
More connections into a separate pool object, similar to the connection pooling strategies from table-providers: https://github.com/datafusion-contrib/datafusion-table-providers/blob/main/core/src/sql/db_connection_pool/duckdbpool.rs

This would allow for more fine-grained concurrency controls by allowing us to potentially maintain locking/transaction outside of the database - which would help prevent race conditions, and allow the safer use of in-memory stores like Arrow.

peasee:
not required for now

PR #9373, opened 2026-02-11
fix: Return better error for Cayenne metadata schema mismatch

peasee:
## 📝 Summary

* Returns a more user friendly error when a Cayenne metadata schema is different from the expected schema, instead of an undetermined SQL error for invalid column selection

## 🔗 Related

* Part of #8777

### DIFF
--- a/crates/cayenne/src/metastore.rs
+++ b/crates/cayenne/src/metastore.rs
@@ -30,6 +30,133 @@ use std::fmt::Display;
 use super::catalog::CatalogResult;
 use async_trait::async_trait;
 
+/// Expected column definitions for a metadata table.
+///
+/// Used by [`validate_existing_schema`] to compare the actual schema of an existing
+/// metadata table against the expected schema. Only column names are checked — types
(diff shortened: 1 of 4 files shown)


crates/cayenne/src/metastore.rs
```diff
@@ -30,6 +30,133 @@ use std::fmt::Display;
...
+    pub name: &'static str,
+    /// The ordered list of expected column names.
+    pub columns: &'static [&'static str],
+}
+
+/// Expected schema definitions for all Cayenne metadata tables.
+///
+/// These must be kept in sync with the DDL constants in `sqlite.rs` and `turso.rs`.
+/// When the schema changes, update both the DDL constants **and** these definitions.
+pub const EXPECTED_TABLES: &[ExpectedTable] = &[
```
peasee, line 51: it would be nice if Cayenne used an ORM like Diesel or something so we could pull this table information from that. That would be a useful improve as part of https://github.com/spiceai/spiceai/issues/8536

phillipleblanc (approved):

PR #9378, opened 2026-02-11
Create `PartitionManagementTask` for scheduler to update accelerated table partition assignments

Jeadie:
## 📝 Summary
 - New background task `PartitionManagementTask`.
   - Add new `acceleration/partitions/<table>.json`.
   - Add and remove  partitions from `acceleration/partitions/<table>.json` based on source SQL table. 
     - On removal, notify executors assigned.
   - Assign unassigned partitions to executors. 

#### New `spicepod.yaml` UX
New UX to configure cadence and logic of how the scheduler's background task assigns partitions. 
```yaml
runtime:
  scheduler:
    state_location: s3://my-metadata-lives-here/
    partition_management:
      interval: 30s
      max_assignments_per_cycle: 100
      max_partitions_per_executor: 1000
      discovery_timeout: 60s
```


sgrebnov (dismissed):

crates/runtime/src/cluster/partition/discovery.rs
```diff
@@ -0,0 +1,169 @@
...
+    };
+
+    let ctx = SessionContext::new_with_state(
+        SessionStateBuilder::new_from_existing(df.ctx.state()).build(),
+    );
+
+    // Register with random UUID. Even if [`SessionContext`] is a clone, it will evict it from the broader DF context.
+    let temp_table =
+        TableReference::parse_str(uuid::Uuid::new_v4().as_simple().to_string().as_str());
+    ctx.register_table(temp_table.clone(), acc.table_provider().await)
```
phillipleblanc: you can use the dataframe API here

crates/runtime/src/cluster/partition/discovery.rs
```diff
@@ -0,0 +1,169 @@
...
+            .as_str(),
+        )
+        .await
+        .boxed()
+        .context(PartitionDiscoverySnafu {
+            table: table_name.clone(),
+        });
+
+    // Deregister the temporary table after query execution to avoid polluting the catalog. This is done regardless of query success or failure.
+    let _ = ctx.deregister_table(temp_table);
```
phillipleblanc: This feels awkward? do we need this if we use the dataframe API and dont register the temp table?

crates/runtime/src/cluster/partition/discovery.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    df_result?
+        .collect()
+        .await
+        .boxed()
+        .context(PartitionDiscoverySnafu { table: table_name })
+}
+
+/// Wait for the [`TableReference`] to be registered in Runtime.
+async fn wait_for_table(table: &TableReference, df: &Arc<DataFusion>) -> bool {
```
phillipleblanc: This could use the wait mechanism in RuntimeStatus to wait for when a dataset becomes ready

crates/runtime/src/cluster/partition/scheduler_task.rs
```diff
@@ -0,0 +1,1150 @@
...
+                }) => {
+                    retries += 1;
+                    if retries >= max_retries {
+                        return Err(Error::MaxRetriesExceeded {
+                            table: table.to_string(),
+                            partition: format!("{} partitions", partition_values.len()),
+                            retries,
+                        });
+                    }
+                    tokio::time::sleep(Duration::from_millis(100 * 2_u64.pow(retries))).await;
```
phillipleblanc: This should use the fibonacci backoff helper

crates/runtime/src/cluster/partition/scheduler_task.rs
```diff
@@ -0,0 +1,1150 @@
...
```
phillipleblanc: This should use the fibonacci backoff helper

crates/runtime/src/cluster/partition/scheduler_task.rs
```diff
@@ -0,0 +1,1150 @@
...
+                            partition: format!("{partition_value:?}"),
+                            retries,
+                        });
+                    }
+                    tracing::debug!(
+                        table = %table,
+                        partition = ?partition_value,
+                        "Concurrent modification detected, retrying"
+                    );
+                    tokio::time::sleep(Duration::from_millis(100 * 2_u64.pow(retries))).await;
```
phillipleblanc: This should use the fibonacci backoff helper

crates/runtime/src/cluster/service.rs
```diff
@@ -562,14 +562,21 @@ impl ClusterService for ClusterServiceImpl {
         &self,
         request: Request<AllocateInitialPartitionsRequest>,
     ) -> Result<Response<AllocateInitialPartitionsResponse>, Status> {
-        let AllocateInitialPartitionsRequest { executor_id } = request.into_inner();
+        let AllocateInitialPartitionsRequest { executor_url } = request.into_inner();
+
+        // Current standard is to have executor id be without scheme.
```
phillipleblanc, line 584: Why this change?

crates/spicepod/src/component/runtime.rs
```diff
@@ -664,6 +664,54 @@ pub struct Scheduler {
     /// Optional object store params for the shared cluster state.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub params: Option<Params>,
+
+    /// Partition management configuration
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub partition_management: Option<PartitionManagement>,
```
phillipleblanc, line 670: This needs to be UX reviewed, please add an example in the PR description of what the parameters are.

crates/runtime/src/cluster/service.rs
Jeadie: The provided executor id is the full URL required to generate a Flight connection. Elsewhere and now in the partition metadata state, the id for a executor node is only the `host:port`. Wanted to make that distinction (i.e. that `executor_url=https://127.0.0.1:12345` not `executor_id=127.0.0.1:12345`).

crates/runtime/src/cluster/partition/discovery.rs
Jeadie: I tried finding something like this. Thought to use RuntimeStatus, but didn't see `RuntimeStatus::wait_for_dataset_ready` existed.

crates/runtime/src/cluster/partition/discovery.rs
Jeadie: Did not know we could use DataFrame API and avoid registering in context. Changing to this.

sgrebnov (approved):

PR #9387, opened 2026-02-12
feat(cayenne): File-based retention deletes for PK tables

sgrebnov:
## 📝 Summary

Extends file-based retention to support PK-based tables (Int64Pk, RowConverterBased). Previously, file-based retention only worked for PositionBased (no PK) tables — PK tables always fell through to the slower deletion vector path.

### Changes

- Remove `is_position_based()` gate in `file_based_deletes_preferred()` — PK tables now use file-level deletes when the filter matches retention pattern
- Scan protected snapshot directories during file-based retention — `FileBasedDeletionSink` iterates main table + all protected snapshot listing tables
- Clean up emptied protected snapshots after file deletion — clears catalog `cayenne_snapshot_sequence`, removes from in-memory `protected_snapshots` map, deletes empty snapshot data directory
- `build_protected_snapshot_listing_tables()` returns `Vec<(String, Arc<ListingTable>)>` shared by both file-based and DV-based delete paths

Write lock coordination is handled by the runtime's existing `accelerator_write_mutex`, which serializes refresh inserts and retention deletes.

### Tests

6 new PK-based retention tests covering main-only, protected snapshots, emptied snapshots, fresh snapshot preservation, upsert-after-delete, and multiple protected snapshots.

## 👀 Notes for Reviewers

There is also separate work item created as further improvement for PK with upserts (nice to have):
- https://github.com/spiceai/spiceai/issues/9388

### DIFF
--- a/crates/cayenne/src/provider/delete/sink/file_based.rs
+++ b/crates/cayenne/src/provider/delete/sink/file_based.rs
@@ -55,7 +55,7 @@ pub(crate) struct FileBasedDeletionResult {
 }
 
 /// Result of scanning a single listing table for retention-eligible files.
-struct DeletionScanResult {
+struct DeletionCheckScanResult {
     /// Files whose `max(retention_col) < threshold`.
     files: Vec<(ObjectMeta, Option<usize>)>,
     /// `true` when every file in the table matched — the table is fully expired.
@@ -82,8 +82,8 @@ pub struct FileBasedDeletionSink {
     /// Main listing table to enumerate files and collect per-file statistics.
     listing_table: Arc<RwLock<Arc<ListingTable>>>,
     /// Protected snapshot listing tables keyed by snapshot ID (PK-based strategies only).
-    /// Empty for position-based tables.
-    protected_snapshot_tables: Vec<(String, Arc<ListingTable>)>,
+    /// `None` for position-based tables.
+    protected_snapshot_tables: Option<Vec<(String, Arc<ListingTable>)>>,
     /// The retention filter expression (e.g., `event_time < literal`).
     filter: Expr,
     /// Table name for logging.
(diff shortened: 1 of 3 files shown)


crates/cayenne/src/provider/delete/sink/file_based.rs
```diff
@@ -34,49 +41,98 @@ use datafusion_catalog::TableProvider;
...
+pub(crate) struct FileBasedDeletionResult {
+    /// Total number of rows deleted across all files.
+    pub total_deleted_rows: u64,
+    /// IDs of protected snapshots that were fully emptied
+    /// (all files expired and deleted).
+    pub emptied_snapshot_ids: Vec<String>,
+}
+
+/// Result of scanning a single listing table for retention-eligible files.
+struct DeletionScanResult {
```
Jeadie: Nit
```suggestion
struct DeletionCheckScanResult {
```

crates/cayenne/src/provider/delete/sink/file_based.rs
```diff
@@ -34,49 +41,98 @@ use datafusion_catalog::TableProvider;
...
 }
 
 impl FileBasedDeletionSink {
     /// Create a new file-based deletion sink.
+    ///
+    /// # Arguments
+    ///
+    /// * `listing_table` - Main listing table for the current snapshot.
+    /// * `protected_snapshot_tables` - Protected snapshot listing tables keyed
+    ///   by snapshot ID. Pass empty vec for position-based tables.
```
Jeadie: > Pass empty vec for position-based tables.

If you want to do it like this, i feel like `protected_snapshot_tables: Option<Vec<(String, Arc<ListingTable>)>>,` where `None for position-based tables.`. 

We could want position-based tables when there are currently no Protected snapshot listing tables?

crates/cayenne/src/provider/delete/sink/file_based.rs
sgrebnov: Updated - thank you

crates/cayenne/src/provider/delete/sink/file_based.rs
sgrebnov: Yeah, I like the `Option<Vec<(String, Arc<ListingTable>)>>` approach, updated.

In term of File-based deletes position-based tables are simplified versions of PK-based tables that can have existing deletions vectors and additional snapshot folders due to upsert functionality. Position-based tables  has no PK-defined so can't have upserts

Jeadie (approved):

PR #9401, opened 2026-02-13
Improvements for snapshot schema comparison

krinart:
## 📝 Summary

1. Only compare `fields`, ignoring `metadata`
2. Improved error message, including details of what fields changed (list of fields plus types)

### DIFF
--- a/crates/runtime-acceleration/src/snapshot/mod.rs
+++ b/crates/runtime-acceleration/src/snapshot/mod.rs
@@ -14,6 +14,7 @@ limitations under the License.
 //! Supports loading and saving snapshots of accelerated database files to and from object storage.
 
 use arrow_schema::{Schema, SchemaRef};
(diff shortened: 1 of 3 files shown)


crates/runtime-acceleration/src/snapshot/mod.rs
```diff
@@ -2581,6 +2584,61 @@ async fn build_s3_parameters(
     }
 }
 
+/// Compares two schemas and returns a human-readable description of the differences.
+fn describe_schema_differences(expected: &SchemaRef, actual: &SchemaRef) -> String {
```
phillipleblanc: There are existing functions in the `arrow_tools` crate that can be reused instead of generating a new function here. Look at `arrow_tools::schema::verify_schema` and `arrow_tools::schema::schema_difference`

sgrebnov (approved):

PR #9403, opened 2026-02-13
Refactor(Cayenne): Replace CatalogError and string based errors with Snafu errors

sgrebnov:
## 📝 Summary

Introduces a structured `Error` with SNAFU for Cayenne table provider operations, replacing ~65 generic `CatalogError` usages across provider code that made no catalog calls.

**Error variants:** `Catalog`, `DataFusion` (transparent), `Arrow` (transparent), `IoError` (transparent), `ObjectStore`, `Vortex`, `DataValidation`, `LockPoisoned`, `TaskPanicked`, `UrlParse`, `Internal`, `InvalidChildrenCount`

**Changes:**
- `table.rs`: ~28 internal functions converted from `CatalogResult` → `Result<_, Error>`; 14 trait-facing functions keep `CatalogResult`
- `vector_io.rs`, `deletion_strategy.rs`, `file_based.rs`: fully migrated off `CatalogError`
- `sink.rs`, `delete/sink.rs`, `position_based.rs`: migrated to typed variants
- `constants.rs`: lock name strings unified via constants
- `mod.rs`: `From` bridges for `CatalogError`↔`Error`↔`DataFusionError` cross-boundary calls


crates/cayenne/src/provider/delete/sink.rs
```diff
@@ -550,7 +604,13 @@ impl CayenneDeletionSink {
             *guard = Arc::new(updated_map);
         }
 
-        let deleted_count = convert_to_u64_box(new_deletion_count, "deleted row count")?;
+        let deleted_count =
+            convert_to_u64_box(new_deletion_count, "deleted row count").map_err(|e| {
+                Error::Internal {
```
Jeadie, line 609: We use`Internal` for a lot of deletion errors. Maybe we should get a deletion specific I've so users know the error is deletion specific

crates/cayenne/src/provider/delete/sink.rs
```diff
@@ -607,10 +663,14 @@ impl DeletionSink for CayenneDeletionSink {
         }
 
         if self.filters.is_empty() {
-            return self.delete_all_rows_from_tables(&ctx, &all_tables).await;
+            return self
+                .delete_all_rows_from_tables(&ctx, &all_tables)
+                .await
+                .map_err(|e| Box::new(e) as Box<dyn std::error::Error + Send + Sync>);
```
Jeadie, line 669: Nit `.boxed()`

crates/cayenne/src/provider/delete/vector_io.rs
```diff
@@ -421,87 +424,65 @@ pub fn detect_deletion_type_and_read(
...
         vec![
             Arc::new(row_id_array) as Arc<dyn Array>,
             Arc::new(deleted_at_array),
         ],
-    )
-    .map_err(|err| CatalogError::InvalidOperation {
-        message: "Failed to build position-based deletion batch.".to_string(),
-        source: Box::new(err),
-    })
+    )?)
```
Jeadie, line 439: Nit: using `Ok` and `?` is unnecessary

crates/cayenne/src/provider/mod.rs
```diff
@@ -65,19 +65,118 @@ pub use vortex_format::{attach_deletion_vectors_to_config, DeletionFilteringVort
...
 #[derive(Debug, Snafu)]
+#[expect(missing_docs)]
 pub enum Error {
-    /// Invalid number of children provided to execution plan
+    /// Catalog operation failed (DB commit, sequence increment, partition ops).
+    #[snafu(display("{source}"))]
+    Catalog { source: CatalogError },
+
+    /// `DataFusion` plan/execution error (scan, sort, insert, listing table ops).
+    #[snafu(transparent)]
```
Jeadie, line 83: Neat

Jeadie (dismissed):

crates/cayenne/src/provider/delete/sink.rs
sgrebnov: Updated - thank you

crates/cayenne/src/provider/delete/vector_io.rs
sgrebnov: Yeah, I don't like it as well - the ? automatically invokes `.into()` to convert from ArrowError to Error. Alternative we can use here `.map_err(Into)` so it is more understandable - thanks  - I'll update

 `Result<RecordBatch, Error>`, found `Result<RecordBatch, ArrowError>`

crates/cayenne/src/provider/delete/sink.rs
sgrebnov: Yeah - my expectation that there will be still deletion error as part of DeleteSink and this one is just internal reason, I'll review how to improve so we have those errors more actionable, but most of them is really something unexpected/internal - like previously saved format is incorrect so we can't convert it back, some unexpected state, etc

krinart (approved):

Issue #8521, opened 2025-12-10
Enhancement: Benchmarks for append/change streams


krinart:
## Goal-State/What/Result

Establish canonical benchmarks for streaming-powered data acceleration (Kafka, Debezium, DynamoDB Streams) that measure ingestion performance and data correctness, comparable to existing TPCH/TPCDS/Clickbench benchmarks for federated and batch-accelerated datasets.

**Primary Metrics:**
- **Ingestion Performance**: Ingestion throughput (records/second)
- **Query Performance Under Load**: Query latency (p50, p95, p99) and success rate during ingestion
- **Data Correctness**: Executing TPCH queries after ingestion and comparing to expected results

---

## Why/Purpose

For federated and batch-accelerated datasets (`full`/`append` modes), we have established canonical benchmarks (TPCH, TPCDS, Clickbench) that provide:

- Performance measurement via query execution duration
- Correctness verification via result validation and query plan comparison

However, **no equivalent solution exists for streaming-powered acceleration** (Kafka, Debezium, DynamoDB Streams). Without streaming benchmarks, we lack:

- **Regression Detection** — No visibility into ingestion performance changes across releases
- **Correctness Guarantees** — No automated verification that CDC pipelines produce accurate final state
- **Capacity Planning Data** — No baseline metrics for customer deployment sizing

This gap increases risk of undetected regressions and limits confidence in streaming acceleration reliability.

## By When

**Issue/Spec written and reviewed:** January 9, 2026
**Done-Done:** January 14, 2026

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Telemetry / Metrics / Task History
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials

## The Algorithm

- [x] Every requirement questioned?
- [x] Delete (Scope) any part you can.
- [x] Simplify.
- [x] Break down into smaller iterations/milestones.
- [x] Opportunities for automation.

## Design Principles

| Principle | Description |
|-----------|-------------|
| **Isolation** | Spice runtime under benchmark must be isolated from other load |
[...]

PR #9405, opened 2026-02-13
Refactor and simplify streaming benchmarks

krinart:
## 📝 Summary

Two Streaming jobs:
* `streaming-bench`: Measures performance of ingestion including while creating snapshots
* `streaming-correctness`: Runs 3 rouns of ingestion with mutations and verifies the results are as expected after each
* Support for both in `testoperator dispatch`

## 🔗 Related

https://github.com/spiceai/spiceai/issues/8521


crates/test-framework/src/queries/tpch/q18.sql
```diff
@@ -4,7 +4,7 @@ select
     o_orderkey,
     o_orderdate,
     o_totalprice,
-    sum(l_quantity)
```
sgrebnov, line 7: @krinart  - why are we changing this? If this is required for a specific connector, it should be done via overrides, IMO.

crates/test-framework/src/queries/validation/mod.rs
```diff
@@ -494,8 +504,12 @@ pub fn validate_tpch_query(
     };
 
     if !equivalent_schemas(&expected_schema, &actual_schema) {
-        println!("expected_schema: {expected_schema:?}");
-        println!("actual_schema: {actual_schema:?}");
+        if let Some(diff) = schema_difference(&expected_schema, &actual_schema) {
```
sgrebnov, line 507: If equivalent_schemas = false, when can we have None for schema_difference()? I would expect the schema diff to always exist when equivalent_schemas is false.

Or is this for additional safety/defense, where we don’t expect the second case to actually happen?

crates/test-framework/src/queries/tpch/q18.sql
krinart: In streaming we use same queries for spice and duckdb to validate data correctnes (duckdb returns the expected result). This change has no effect for spice, it only has effect for duckdb (changes default column name to match what spice returns).

So not sure it warrants an override since it has no effect on Spice.

crates/test-framework/src/queries/validation/mod.rs
krinart: Since `schema_difference` difference returns `Option<String>`, we need to handle the case of `None`. The only alternative is to have `panic`.

Or am I missing something here?

phillipleblanc:
@krinart what is the status of this PR?

krinart:
@phillipleblanc I'm finalizing it

sgrebnov (approved):

PR #9486, opened 2026-02-24
query: reconcile execution stream nullability with logical plan schema

### LINKED ISSUE #9482: FlightSQL schema nullability mismatch on aggregate queries (GetFlightInfo vs DoGet)
## Summary

Aggregate queries like `SELECT MAX(col) FROM table` produce a **nullability mismatch** between the schema reported by FlightSQL `GetFlightInfo` and the schema in the `DoGet` data stream. `GetFlightInfo` correctly reports the aggregate output as nullable (since `MAX` over empty input is NULL), but `DoGet` returns a non-nullable schema. ADBC clients detect this inconsistency and reject the response.

## Steps to Reproduce

1. Create a spicepod with a Cayenne-accelerated dataset (result cache disabled to isolate this bug):

```yaml
version: v1beta1
kind: Spicepod
name: repro

runtime:
  results_cache:
    enabled: false

datasets:
  - from: file:data.csv
    name: test_data
    params:
      file_format: csv
    acceleration:
      enabled: true
      engine: cayenne
      mode: file
      refresh_mode: full
      refresh_check_interval: 10s
```

With `data.csv`:
```csv
id,name,value
1,alice,100
2,bob,200
3,charlie,300
```

2. Start `spiced` and wait for the dataset to load.

3. Query via ADBC FlightSQL:

```python
import adbc_driver_flightsql.dbapi as flightsql

conn = flightsql.connect("grpc://127.0.0.1:50051")
cursor = conn.cursor()
cursor.execute("SELECT MAX(id) FROM test_data")
t = cursor.fetch_arrow_table()  # ERROR: inconsistent schema
```

## Expected Behavior

The query succeeds and returns a single row with the max value. Both `GetFlightInfo` and `DoGet` should agree on the schema nullability.

## Actual Behavior

Fails 100% of the time:

```
UNKNOWN: [FlightSQL] endpoint 0 returned inconsistent schema: expected schema:
  fields: 1
    - max(test_data.id): type=int64, nullable
but got schema:
  fields: 1
    - max(test_data.id): type=int64
```

The `GetFlightInfo` schema says nullable; the `DoGet` schema says non-nullable.

Also reproduces with aliases: `SELECT MAX(value) as max_val FROM test_data` fails with the same pattern.

Queries that are **not** affected:
- `SELECT MAX(id) FROM test_data WHERE 1=0` (empty input — works correctly)
- `SELECT COUNT(*) FROM test_data` (COUNT is non-nullable in both paths)
- `SELECT MIN(id), MAX(id), AVG(value) FROM test_data` (multi-aggregate — works)
- `SELECT name, SUM(value) FROM test_data GROUP BY name HAVING SUM(value) > 100` (works)

## Root Cause

The FlightSQL `GetFlightInfo` handler computes the schema from the **logical plan**, which correctly marks `MAX(int64)` as nullable (since MAX over empty input returns NULL). The `DoGet` handler executes the query and uses the schema from the **physical execution output**, where the Cayenne accelerator produces a non-nullable schema for the aggregate result batch. The two schemas disagree on nullability.

## Impact

This affects any ADBC/FlightSQL client issuing simple `MAX()` aggregate queries against Cayenne-accelerated tables. In SpiceBench, this blocks `SELECT MAX(__created_at)` queries used for end-to-end latency measurement.

phillipleblanc:
## Problem

FlightSQL schema nullability mismatch on aggregate queries (`GetFlightInfo` vs `DoGet`). When ADBC clients execute `SELECT MAX(col) FROM table` against Cayenne-accelerated datasets, `GetFlightInfo` reports the aggregate output as nullable (correct), but `DoGet` returns a non-nullable schema, causing clients to reject the response.

## Root Cause

1. Cayenne's `CayenneTable::statistics()` reports exact column-level stats (min/max) from Vortex `ListingTable`
2. DataFusion's `AggregateStatistics` physical optimizer sees `SELECT MAX(id)` with `Precision::Exact` stats
3. It replaces `AggregateExec` with `ProjectionExec(expressions::lit(max_value))`
4. `Literal::new(ScalarValue::Int64(Some(42)))` creates `Field::new("lit", Int64, value.is_null())` → `nullable=false`
5. Output schema becomes non-nullable, contradicting logical plan's `nullable=true` (from `Max::is_nullable() -> true`)

This is not Cayenne-specific — it affects any table provider reporting exact column statistics.

## Fix

Reconcile the execution result stream with the logical plan schema in `Query::run_internal()`, right after the existing `verify_schema` check. Fields that the logical plan marks as nullable but the execution reports as non-nullable are widened to nullable, and each batch is adjusted via `try_cast_to` (the same mechanism `SchemaCastScanExec` uses).

A non-nullable array is always valid as a nullable array (it just has no nulls), so this widening is safe.

## Tests

Added 6 unit tests for `reconcile_stream_nullability`:
- Widening non-nullable to nullable (core case)
- No-op when schemas already match (both nullable)
- No-op when both non-nullable
- Mixed fields (some need widening, some don't)
- Field count mismatch (pass through unchanged)
- Empty stream with schema reconciliation

Fixes #9482

### DIFF
--- a/crates/runtime/src/datafusion/query.rs
+++ b/crates/runtime/src/datafusion/query.rs
@@ -722,7 +722,9 @@ impl Query {
             // and DDL plans (CREATE TABLE/DROP TABLE) as their logical plan schema may
             // differ from the actual execution result (DDL plans may be rewritten by
             // analyzer rules into extension nodes with different output schemas)
-            if !matches!(&*plan, LogicalPlan::Statement(_) | LogicalPlan::Ddl(_)) {
+            let res_stream = if matches!(&*plan, LogicalPlan::Statement(_) | LogicalPlan::Ddl(_)) {
+                res_stream
+            } else {
                 let plan_schema = Arc::clone(plan.schema().inner());
                 let res_schema = res_stream.schema();
 
@@ -735,7 +737,15 @@ impl Query {
                         SchemaMismatch
                     )
                 }
-            }
+                // The AggregateStatistics physical optimizer may replace an
+                // AggregateExec with a ProjectionExec containing a literal
+                // value, which changes the output nullability (literals report
+                // nullable = value.is_null()).  Reconcile the execution result
+                // schema with the logical plan schema so downstream consumers
+                // (e.g. FlightSQL GetFlightInfo vs DoGet) see consistent
+                // nullability.
+                reconcile_stream_nullability(res_stream, &plan_schema)
+            };
 
             let final_stream = if cache_manager.should_cache_results() {
                 Self::wrap_stream_with_cache(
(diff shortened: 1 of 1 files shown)


sgrebnov (dismissed):

crates/runtime/src/datafusion/query.rs
```diff
@@ -1092,6 +1102,70 @@ pub fn write_to_json_string(
...
+/// schema says they should be nullable, and wraps the stream so every batch
+/// conforms.  This uses the same `try_cast_to` mechanism as `SchemaCastScanExec`.
+fn reconcile_stream_nullability(
+    stream: SendableRecordBatchStream,
+    plan_schema: &Arc<Schema>,
+) -> SendableRecordBatchStream {
+    let exec_schema = stream.schema();
+
+    if exec_schema.fields().len() != plan_schema.fields().len() {
+        return stream;
```
Jeadie, line 1129: nit: this should be a warning or similar?

crates/runtime/src/datafusion/query.rs
```diff
@@ -1092,6 +1102,70 @@ pub fn write_to_json_string(
...
+    }
+
+    // Build a reconciled schema: widen to nullable where the logical plan says
+    // so, but keep data types and metadata from the execution schema.
+    let reconciled_fields: Vec<Field> = exec_schema
+        .fields()
+        .iter()
+        .zip(plan_schema.fields())
+        .map(|(exec_field, plan_field)| {
+            if plan_field.is_nullable() && !exec_field.is_nullable() {
```
Jeadie, line 1151: I don't think `exec_field.is_nullable()`and `!plan_field.is_nullable()`  is possible, correct?

Jeadie (dismissed):

crates/runtime/src/datafusion/query.rs
phillipleblanc: Correct

Jeadie (approved):

PR #9491, opened 2026-02-24
cayenne: Staged append writes to prevent partial writes and data loss on stream error

### LINKED ISSUE #9380: Cayenne: Partial append writes not rolled back on stream error — causes permanent data loss
## Describe the bug

When a data source stream fails mid-download during an append refresh, already-written Vortex file chunks persist on disk in the active snapshot directory. Because append mode uses `max(time_column)` from the accelerator as the watermark for the next fetch, the watermark advances based on the partial data. Subsequent appends only request rows newer than this watermark, so the missing rows from the failed download are never re-fetched — resulting in permanent data loss.

Suggested fix:
Write append chunks to a staging snapshot directory first. Only make them visible (rename into the active snapshot or swap the listing table) after ALL chunks succeed. On error, delete the staging directory. This ensures the watermark never advances past successfully-committed data.

sgrebnov:
## 📝 Summary

Fixes #9380

Append writes chunk data directly into the active snapshot directory. On stream error, partial Vortex files persist. The next refresh reads `MAX(time_column)` from orphaned files, advancing the watermark past data never fully written — partial writes pollute the snapshot and cause permanent data loss.

### Two-steps write

PR changes append writes to first write into fixed `_staging/` directory. On success, move files into the current snapshot (`rename()` is a metadata-only operation, no data bytes are copied). On error, delete staging. This prevents partial writes, also fixes partial writes are visible immediately during append.

Also fixes `write_chunk_to_listing_table` creating a bare `SessionContext::new()` per chunk without registering the object store (broken for S3). Now creates one context and shares `Arc<SessionState>` across parallel chunk writes.

### Staging WAL

A lightweight write-ahead log (`_staging/_wal.json`) is written after data files are staged but before they are moved to the active snapshot. It records the table name, target snapshot, list of staged files, and a timestamp.

**Lifecycle:**
1. Data files written to `_staging/`
2. WAL written and fsynced
3. Files moved to snapshot (WAL is skipped during move)
4. WAL removed on success

### Considerations

- **WAL timing**: An alternative is to write the WAL only after the first file is moved, but there is still a window between the move and a crash where no WAL exists. The current approach — write WAL before any move begins — is more robust and simpler.
- **Staging directory vs. direct writes**: Writing directly to the target snapshot with a WAL would make partial writes immediately visible to queries via the listing provider and require immediate resolution. The current approach writes to `_staging/` first and creates the WAL only after all data has been successfully downloaded. On local FS, rename is a metadata-only operation (no file copy), so the move phase is fast.
- **Manual resolution**: If the WAL is detected on startup, the table is blocked until the operator resolves the inconsistency. Future improvement: automatic WAL cleanup/recovery (data is preserved in `_staging` folder)

## 🔗 Related

Fixes 
- https://github.com/spiceai/spiceai/issues/9380


crates/cayenne/src/provider/sink.rs
```diff
@@ -254,12 +255,39 @@ impl CayenneDataSink {
...
+                            "Failed to clean staging dir after write error for table {}: {cleanup_err}",
+                            self.table.table_name(),
+                        );
+                    }
+                    return Err(e);
+                }
+            };
+
+            // Step 3: Move files from staging into current snapshot (atomic on local FS)
+            self.table.move_files_to_current_snapshot().await?;
```
phillipleblanc, line 298: Is this also a metadata only operation for s3 express one zone?

crates/cayenne/src/provider/table.rs
```diff
@@ -706,6 +709,178 @@ impl CayenneTableProvider {
...
+            let file_type = entry.file_type().await?;
+            if !file_type.is_file() {
+                continue;
+            }
+
+            let file_name = entry.file_name();
+            let src = staging_dir.join(&file_name);
+            let dst = target_dir.join(&file_name);
+
+            tokio::fs::rename(&src, &dst).await?;
```
phillipleblanc, line 811: If these operations fail, then we could also be left in an inconsistent state right? Maybe we should have some concept of a WAL/operation log that we use to indicate which operations we are going to perform that we write and then delete afterwards. It could be used to know if we aborted in the middle of this operation - which would mean we are in an inconsistent state.

phillipleblanc: Initially we could just use the presence of an operation log to indicate something went wrong, and require the user to take a manual action (i.e. delete the acceleration directory and retry). In the future we could replay the operation log to find out which operation failed and attempt to recover.

crates/cayenne/tests/staged_append_test.rs
```diff
@@ -0,0 +1,423 @@
...
+    );
+
+    // Verify: staging dir is clean
+    assert_staging_empty(&staging_dir(&table));
+
+    Ok(())
+}
+
+// ============================================================================
+// Test 3: Self-healing — leftover files in _staging/ cleaned on next append
```
phillipleblanc, line 143: We should have a test that verifies if the rename from staging fails that we either automatically recover or throw an inconsistent error

crates/cayenne/src/provider/sink.rs
sgrebnov: For S3 (including S3 Express One Zone), this will perform a copy + delete per file:
- It is a server-side copy — no data is downloaded/uploaded through the client. AWS copies the object internally within the same bucket/region, so it is very fast (negligible compared to the time spent actually writing the parquet data).
- As a potential improvement, we can parallelize the copy+delete operations across files.

crates/cayenne/tests/staged_append_test.rs
sgrebnov: Added tests to verify that we properly detect inconsistent state and prevent initialization + further writes: https://github.com/spiceai/spiceai/pull/9491/commits/c578f4d348e8e81837b91a401ee987d62443bd88#diff-6f4073a2573c7541348d76c9709268f742499cd0465f69af7f88ef1a895fdd09

crates/cayenne/src/provider/table.rs
sgrebnov: Great point, added logic to detect partial writes during rename: lightweight write-ahead log (_staging/_wal.json) is written after data files are staged but before they are moved to the active snapshot. It records the table name, target snapshot, list of staged files, and a timestamp. If the WAL is detected on startup or before next write, the table is blocked until the operator resolves the inconsistency. We will be able to add auto-recovery attempt based on  `_wal.json` metadata next.

Lifecycle:

Data files written to _staging/
WAL written and fsynced
Files moved to snapshot (WAL is skipped during move)
WAL removed on success

phillipleblanc (approved):

PR #9493, opened 2026-02-24
feat: Support distributed ingestion in cayenne catalog

peasee:


crates/runtime/src/datafusion/cayenne_ddl/physical_plans.rs
```diff
@@ -44,8 +44,58 @@ use datafusion_table_providers::UnsupportedTypeAction;
...
 use super::get_cayenne_provider;
 use crate::catalogconnector::cayenne::provider::CayenneSchemaProvider;
+use crate::cluster::executor_registry::ExecutorRegistry;
 use crate::dataaccelerator::cayenne::transform_schema_for_vortex;
 
+/// Maps an Arrow [`DataType`] to a SQL type string suitable for DDL forwarding.
+///
+/// Returns a SQL type that DataFusion's SQL parser can understand in a
+/// `CREATE TABLE` statement.
+fn arrow_datatype_to_sql(dt: &DataType) -> String {
```
sgrebnov, line 54: Can we re-use existing [to_source_native_type_name](https://github.com/spiceai/spiceai/blob/fc9804c2e289a34590382465644cd0843abc6c71/crates/arrow_tools/src/schema.rs#L231)?

crates/runtime/src/datafusion/mod.rs
```diff
@@ -1087,6 +1109,130 @@ impl DataFusion {
         Ok(())
     }
 
+    /// Returns `true` if this is a scheduler node and the table is a Cayenne table,
+    /// meaning the write should be forwarded to an executor.
+    fn should_forward_cayenne_write(
```
sgrebnov, line 1112: Will this correctly work for HA where we have multiple schedulers - will they forward the same data that will be duplicated in case of append or can lead to other unpredicted behavior? if not, we might want to add some verification for scenario and add warning / tracing

PR #9499, opened 2026-02-25
Make partition discovery more robust and make initialization non-blocking

sgrebnov:
## 📝 Summary

Fixes two issues with cluster partition discovery that caused scheduler startup failures on large datasets.

### 1. Fix: "Table X is not an accelerated table" error

Partition discovery failed with `Failed to discover partition values, error=Table my_table is not an accelerated table` because `AcceleratedTable` was wrapped in a `FederatedTableProviderAdaptor` when federation is enabled. The direct `downcast_ref::<AcceleratedTable>()` returned `None`.

Fixed by using `find_concrete_table_provider()` to unwrap through all known adapter types before downcasting.

### 2. Non-blocking partition metadata initialization

`initialize_partition_metadata` was called inline during scheduler startup, blocking HTTP/Flight servers from starting. For large source tables (e.g. 2TB S3 Delta Lake), the discovery query (`SELECT DISTINCT bucket(50, org_id)`) scans all rows and can exceed the K8s health check timeout.

Moved initialization into `PartitionManagementTask::run()` as a cancellation-aware first step before the periodic loop. Servers now start immediately while partition discovery proceeds in the background and reports its readiness via `status.update_component_status("partition_metadata", ComponentStatus::Ready);`

### Changes

- **discovery.rs**: Use `find_concrete_table_provider` instead of direct downcast; add debug tracing
- **scheduler_task.rs**: Add `initialize_metadata()` as first step in `run()` with `tokio::select!` cancellation
- **mod.rs**: Remove inline blocking init call


.github/workflows/integration.yml
```diff
@@ -292,6 +292,8 @@ jobs:
           ORACLE_CLOUD_WALLET_SSO_CERT: ${{ secrets.ORACLE_CLOUD_WALLET_SSO_CERT }}
           AWS_SNAPSHOT_KEY: ${{ secrets.AWS_ICEBERG_ACCESS_KEY_ID }}
           AWS_SNAPSHOT_SECRET: ${{ secrets.AWS_ICEBERG_SECRET_ACCESS_KEY }}
+          AWS_S3_VECTORS_KEY: ${{ secrets.AWS_S3_VECTORS_KEY }}
```
sgrebnov, line 295: The key is required and used for cluster state, decided to re-use existing key instead of adding a new one for S3. Note: I plan to switch to minio as next iteration when will be adding more integration tests for cluster mode: async queries, acceleration with partitioning on column name/ not bucket, etc

crates/runtime/src/cluster/partition/discovery.rs
```diff
@@ -20,25 +20,52 @@ use crate::{
...
+    );
+
+    if partitioning.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    // Fast path: if every partition expression has a statically known value set,
+    // generate partition values without querying the source table.
+    if let Some(values) = try_static_partition_values(partitioning) {
+        tracing::info!(
```
phillipleblanc:
```suggestion
        tracing::debug!(
```

crates/runtime/src/cluster/service.rs
```diff
@@ -610,6 +610,13 @@ impl ClusterService for ClusterServiceImpl {
                 // Find accelerated datasets with partitioning
 
                 for table_ref in super::partition::accelerated_tables(app).keys() {
+                    // Wait for the table to be registered in DataFusion before serializing partition expressions.
+                    // Executors may connect before the scheduler has finished loading all datasets.
+                    self.datafusion
```
phillipleblanc: What is the purpose for this? It could block executor startup if this doesn't become ready right?

Jeadie: IIRC datasets have to be available in DF context for some de/serialization of Expr.

Jeadie: But yes, executors would be blocked by the readiness of the scheduler it connected to. Schedulers are waiting on just readiness checks (and schema check) and not an Acceleration logic.

sgrebnov: Thank you - removed as this is unnecessary, I think it was just me manually creating executor w/o properly waiting for scheduler is fully ready. Will re-iterate on this one more time if I still see this

Jeadie (approved):

PR #9531, opened 2026-02-27
feat(spidapter): add local backend mode with single executor

phillipleblanc:
## Summary
- add `SPIDAPTER_BACKEND` routing in `spidapter` setup (`scp` default, `local` new path)
- implement local setup/teardown lifecycle that starts a scheduler plus exactly one executor
- generate local cluster certificates via `spice cluster tls init` and `spice cluster tls add`
- wait for local readiness before returning ADBC config
- keep the setup response contract unchanged (FlightSQL `driver` + `db_kwargs`)
- update spidapter README with local backend env vars and single-executor note

## Validation
- `cargo check -p spidapter`
- `cargo test -p spidapter`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18164,6 +18164,7 @@ dependencies = [
  "arrow",
  "async-trait",
  "clap",
+ "dirs",
  "reqwest 0.12.24",
  "runtime-secrets",
  "rustls 0.23.36",
--- a/tools/spidapter/Cargo.toml
+++ b/tools/spidapter/Cargo.toml
@@ -13,6 +13,7 @@ arrow.workspace = true
 anyhow.workspace = true
 async-trait.workspace = true
 clap.workspace = true
+dirs = "6"
 reqwest.workspace = true
 runtime-secrets = { path = "../../crates/runtime-secrets", default-features = false }
 spice-cloud-client.workspace = true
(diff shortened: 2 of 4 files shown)


phillipleblanc:
Follow-up on Copilot review comments:

Addressed in latest push:
- switched local process management to `tokio::process` and async teardown (`kill().await` / `wait().await`)
- moved local file operations in async paths to `tokio::fs`
- kept stdio protocol clean by redirecting child command stdout to `Stdio::null()`
- added PK validation against schema fields in ADBC CREATE TABLE generation
- switched local connectivity/advertise path to `127.0.0.1` while keeping bind host `0.0.0.0`
- switched PKI home resolution to `dirs::home_dir()`
- enabled `kill_on_drop(true)` on spawned local `spiced` processes

Intentional choices kept:
- bind host remains `0.0.0.0` (requested for this task)
- CREATE TABLE remains without `IF NOT EXISTS` to avoid silently accepting stale/incompatible schema

Known follow-up candidates (not in this PR):
- port reservation race between selecting and binding ports
- CA init TOCTOU race for concurrent local spidapter runs

sgrebnov (dismissed):

ewgenius (dismissed):

sgrebnov (dismissed):

ewgenius (approved):

Issue #9513, opened 2026-02-26
Distributed acceleration error when `refresh_sql` is present


sgrebnov:
## Describe the bug

Executor is unable to complete acceleration due to error below when `refresh_sql` exists

>2026-02-26T08:17:55.479304Z ERROR runtime::init::dataset: Failed to setup the dataset taxi_trips (s3). Expected a SQL query starting with SELECT <columns> FROM taxi_trips

```yaml
datasets:
- from: s3://spiceai-demo-datasets/taxi_trips/2024/
  name: taxi_trips
  time_column: tpep_pickup_datetime
  params: 
    file_format: parquet
  acceleration:
    enabled: true
    refresh_sql: |
      SELECT
        VendorID,
        tpep_pickup_datetime,
        tpep_dropoff_datetime,
        passenger_count,
        trip_distance,
        RatecodeID,
        store_and_fwd_flag,
        PULocationID,
        DOLocationID,
        payment_type,
        fare_amount,
        extra,
        mta_tax,
        tip_amount,
        tolls_amount,
        improvement_surcharge,
        total_amount,
        congestion_surcharge,
        Airport_fee
      FROM taxi_trips
    refresh_mode: append
    refresh_on_startup: auto # change to 'always' if you want to trigger refresh on every startup
    engine: cayenne
    mode: file
    refresh_check_interval: 15m
    partition_by:
      - bucket(50, VendorID)
```

Jeadie:
Currently we read `acceleration.refresh_sql` as a string and use it as such at execution time. Prior to Spice DQ, there was only moderate parsing of this SQL (to validate, then to apply timestamp filters for both retention and refresh periods). Now there are acceleration predicates being added and removed across time. For example

```sql
SELECT
        VendorID,
        tpep_pickup_datetime,
        tpep_dropoff_datetime
FROM taxi_trips
```

Becomes 
```sql
SELECT
        VendorID,
        tpep_pickup_datetime,
        tpep_dropoff_datetime
FROM taxi_trips
WHERE bucket(50, VendorID) IN (1, 5, 7, 12)
```
but will also update over time (i.e. the acceleration partition predicate `WHERE bucket(50, VendorID) IN (1, 5, 7, 12)`). 


Retention checks and refresh periods also mutate the SQL.
```
SELECT
        VendorID,
        tpep_pickup_datetime,
        tpep_dropoff_datetime
FROM taxi_trips
[...]

PR #9549, opened 2026-03-02
Make `Refresh::refresh_sql` more robust to alterations over time.

Jeadie:
## 📝 Summary
  - Introduces a structured `RefreshSQL` type that decomposes user `refresh_sql` into validated parts (columns, WHERE predicates, LIMIT) instead of storing it as a raw `Option<String>`
  - Stores cluster partition filters as separate `Vec<Expr>` on `RefreshSQL` rather than wrapping user SQL in a subquery, fixing validation failures when `refresh_sql` is used with `partition_by` in cluster mode.

### Details
Core change is an update to `Refresh`, where `RefreshSQL` is new and defined [here](https://github.com/spiceai/spiceai/pull/9549/changes#diff-fc6c4697d8c3043b806c6861fb6e426c7c4cd15208f51c115e41b956065f563dR69-R82). Remaining changes are to accomodate this (and by extension make refresh_sql and acceleration partition filters play nice together). 

```diff
#[derive(Clone, Debug)]
pub struct Refresh {
    pub(crate) time_column: Option<String>,
    pub(crate) time_format: Option<TimeFormat>,
    pub(crate) time_partition_column: Option<String>,
    pub(crate) time_partition_format: Option<TimeFormat>,
    pub(crate) check_interval: Option<Duration>,
    pub(crate) max_jitter: Option<Duration>,
-   pub(crate) sql: Option<String>,
+   pub(crate) sql: Option<RefreshSQL>,
```

## 🔗 Related
#9513

### DIFF
--- a/crates/runtime/src/accelerated_table/mod.rs
+++ b/crates/runtime/src/accelerated_table/mod.rs
@@ -994,20 +994,61 @@ impl AcceleratedTable {
         Arc::clone(&self.synchronized_children)
     }
 
-    pub async fn update_refresh_sql(&self, refresh_sql: Option<String>) -> Result<()> {
+    pub async fn update_refresh_sql(
+        &self,
+        refresh_sql: Option<crate::accelerated_table::refresh::RefreshSQL>,
+    ) -> Result<()> {
         let dataset_name = &self.dataset_name;
 
         let mut refresh = self.refresh_params.write().await;
-        refresh.sql.clone_from(&refresh_sql);
+        // Preserve existing partition_filters when updating user SQL
+        let existing_partition_filters = refresh
+            .sql
+            .as_ref()
+            .map(|s| s.partition_filters().to_vec())
+            .unwrap_or_default();
 
-        if !is_spice_internal_dataset(&self.dataset_name) {
-            if let Some(sql_str) = &refresh_sql {
-                tracing::info!("[refresh] Updated refresh SQL for {dataset_name} to {sql_str}");
-            } else {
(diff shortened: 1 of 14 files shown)


phillipleblanc (dismissed):

crates/runtime/src/http/v1/datasets.rs
```diff
@@ -227,7 +227,7 @@ pub(crate) struct MessageResponse {
 #[derive(Deserialize)]
 #[cfg_attr(feature = "openapi", derive(utoipa::ToSchema))]
 pub struct AccelerationRequest {
-    /// SQL statement used for the refresh. Defaults to the `refresh_sql` specified in the spicepod.
```
Jeadie, line 230: This is not a functional change on the API. As per below, `update_refresh_sql(` was never None due to an early exit.

sgrebnov (dismissed):

crates/runtime/src/accelerated_table/mod.rs
```diff
@@ -994,20 +994,51 @@ impl AcceleratedTable {
         Arc::clone(&self.synchronized_children)
     }
 
-    pub async fn update_refresh_sql(&self, refresh_sql: Option<String>) -> Result<()> {
+    pub async fn update_refresh_sql(&self, mut refresh_sql: refresh::RefreshSQL) -> Result<()> {
```
sgrebnov, line 999: Nit: can we have a test for cluster mode with refresh? Can be a separate PR

krinart (approved):

PR #9579, opened 2026-03-03
[SpiceDQ] Improve error messages; Avoid race condition on allocate_initial_partitions.

Jeadie:
## 📝 Summary
 - Better error messages for when we can't find table metadata (for acceleration partitions).
 - Better error messages for `refresh_sql`. 
 - Avoid issues when executor requests `allocate_initial_partitions` earlier than scheduler has populated metadata ( some times it needs to query the table to find distinct partition values).


crates/runtime/src/cluster/service.rs
```diff
@@ -610,6 +610,15 @@ impl ClusterService for ClusterServiceImpl {
                 // Find accelerated datasets with partitioning
 
                 for table_ref in super::partition::accelerated_tables(app).keys() {
+                    if partition_manager
+                        .get_cached_table_metadata(table_ref)
+                        .is_none()
+                    {
+                        tracing::info!(
```
phillipleblanc:
```suggestion
                        tracing::warn!(
```

This should be a warning?

phillipleblanc (dismissed):

crates/runtime/src/cluster/service.rs
Jeadie: because we're not waiting on `component_status("partition_metadata", ComponentStatus::Ready)`, its not as if something has gone wrong.

phillipleblanc (dismissed):

phillipleblanc (dismissed):

sgrebnov (dismissed):

phillipleblanc (approved):

PR #9612, opened 2026-03-04
Fix CLI when using custom `--http`

krinart:
## 📝 Summary

Currenty when providing custom `--http` we get following error:

```bash
error: the argument '--http <BIND_ADDRESS>' cannot be used multiple times
```

This PR fixes it and also adds short `-e` alias for `--endpoint` flag in `spice sql` command

### DIFF
--- a/bin/spice/src/commands/sql.rs
+++ b/bin/spice/src/commands/sql.rs
@@ -38,7 +38,7 @@ pub struct SqlArgs {
     /// Specifies the remote Spice instance endpoint.
     /// Supports http://, https://, grpc://, or grpc+tls:// schemes.
     /// If not provided, uses local spiced runtime.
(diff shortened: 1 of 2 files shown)


bin/spice/src/commands/sql.rs
```diff
@@ -38,7 +38,7 @@ pub struct SqlArgs {
     /// Specifies the remote Spice instance endpoint.
     /// Supports http://, https://, grpc://, or grpc+tls:// schemes.
     /// If not provided, uses local spiced runtime.
-    #[arg(long)]
+    #[arg(long, short = 'e')]
```
lukekim: No short `e`

phillipleblanc (approved):

PR #9631, opened 2026-03-04
Implement Cayenne S3 Express multi-zone live test with data validation

lukekim:
- Added functionality to append a single row to a Cayenne-accelerated table via the Runtime write path.
- Introduced helper functions for S3 bucket name generation and object listing.
- Implemented validation of S3 replica integrity by comparing primary and replica zone objects.
- Created a new test case for validating Cayenne's behavior with multiple S3 zones, including data insertion and integrity checks.
- Enhanced error handling and logging for better traceability during test execution.

### DIFF
--- a/crates/runtime/src/dataaccelerator/cayenne/s3.rs
+++ b/crates/runtime/src/dataaccelerator/cayenne/s3.rs
@@ -31,6 +31,7 @@ use runtime_parameters::ParameterSpec;
 use runtime_secrets::get_params_with_secrets;
 use secrecy::ExposeSecret;
 use snafu::{ResultExt, Snafu};
+use tokio::sync::Mutex;
 use url::Url;
 
 use super::AccelerationSource;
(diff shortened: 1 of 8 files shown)


phillipleblanc (dismissed):

sgrebnov (dismissed):

crates/runtime/src/dataaccelerator/cayenne/s3.rs
```diff
@@ -89,6 +107,665 @@ pub enum Error {
...
+}
+
+impl Display for MultiZoneS3ExpressStore {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{MULTI_ZONE_STORE_NAME}")
+    }
+}
+
+#[async_trait::async_trait]
+impl ObjectStore for MultiZoneS3ExpressStore {
```
phillipleblanc, line 330: nit: This seems like it would be useful as a utility, we could consider moving it to a separate crate for reuse

phillipleblanc (approved):

PR #9651, opened 2026-03-05
Improve CDC cache invalidation

### LINKED ISSUE #9472: SQL results cache invalidated on every change stream poll, even with zero records
### Version

v1.11.2 (also present on `main` at `efcc1949c`)

### Description

When `refresh_mode: changes` is configured (e.g., DynamoDB Streams), the SQL results cache is invalidated on **every change stream poll**, including polls that return zero records. This makes the cache ineffective for datasets using change-based refresh, since the cache is wiped every few hundred milliseconds regardless of whether the underlying data has actually changed.

With `refresh_check_interval: 3s` and DynamoDB Streams, we observe the stream polling at ~500ms intervals. Each poll delivers a `ChangeBatch` with `num_rows() == 0`, which still calls `write_change()` → returns `Ok(())` → unconditionally calls `invalidate_for_table()`. This means cached query results survive at most ~500ms before being invalidated.

### Steps to reproduce

1. Configure a dataset with `refresh_mode: changes` (e.g., DynamoDB Streams source) and `refresh_check_interval: 3s`
2. Enable the SQL results cache:
   ```yaml
   runtime:
     caching:
       sql_results:
         enabled: true
         cache_key_type: sql
         max_size: 256MiB
         item_ttl: 10m
         eviction_policy: lru
   ```
3. Send 50 identical queries in sequence (e.g., `SELECT * FROM <dataset> LIMIT 100`)
4. Check cache metrics via `runtime.metrics`

**Expected:** ~98% cache hit ratio (1 miss + 49 hits on a single-replica deployment)
**Actual:** ~55-65% cache hit ratio, with `results_cache_items_count` stuck at 1 and `results_cache_evictions` at 0

#### Pod startup log confirms cache is initialized

```
2026-02-21T21:13:52.900272Z  INFO runtime::init::caching: Initialized sql results cache; max size: 256.00 MiB, item ttl: 600s, hashing algorithm: XXH3, encoding: zstd
```

#### DynamoDB Streams polls with zero records every ~500ms

```
2026-02-21T21:24:32.410711Z DEBUG runtime::dataconnector::dynamodb: Processing DynamoDB Streams batch dataset=my_dataset_a watermark="2026-02-21T21:24:32.410638400Z" lag="61us 41ns" shards=15 records=0
2026-02-21T21:24:32.558654Z DEBUG runtime::dataconnector::dynamodb: Processing DynamoDB Streams batch dataset=my_dataset_b watermark="2026-02-21T21:24:32.558596779Z" lag="49us 99ns" shards=15 records=0
2026-02-21T21:24:32.752490Z DEBUG runtime::dataconnector::dynamodb: Processing DynamoDB Streams batch dataset=my_dataset_c watermark="2026-02-21T21:24:32.752411764Z" lag="65us 655ns" shards=15 records=0
```

These `records=0` batches repeat every ~500ms per dataset, each triggering cache invalidation.

#### Cache metrics after 50 identical queries

```
results_cache_hits:           144
results_cache_misses:         2885
results_cache_requests:       3029
results_cache_items_count:    1
results_cache_size_bytes:     6224
results_cache_max_size_bytes: 268435456  (256 MiB)
results_cache_evictions:      0
```

Key observations:
- **items_count=1** despite 256 MiB available and only 6 KB used — results are invalidated before they accumulate
- **evictions=0** — items are being invalidated by refresh, not evicted by capacity
- **Lifetime hit ratio ~4.8%** — the vast majority of cache lookups are misses

#### Isolated burst test

Sending 10 identical queries (`SELECT * FROM my_dataset LIMIT 100`) to a single-replica pod produced **only 1 new cache hit** — the remaining 9 queries were misses because the cache entry was invalidated between queries by the stream poll.

### Root cause

In `crates/runtime/src/accelerated_table/refresh_task/changes.rs`, lines 90-125:

```rust
while let Some(update) = changes_stream.next().await {
    match update {
        Ok(change_envelope) => {
            match self.write_change(change_envelope.change_batch.clone()).await {
                Ok(()) => {
                    // ... commit logic ...

                    // Cache invalidation is unconditional — no check for num_rows() > 0
                    if let Some(cache_provider_ref) = caching.as_ref()
                        && let Some(cache_provider) = cache_provider_ref.upgrade()
                        && let Err(e) =
                            cache_provider.invalidate_for_table(dataset_name.clone())
                        && !self.runtime_status.is_shutdown()
                    {
                        // ...
                    }
                }
```

`write_change()` returns `Ok(())` even when the change batch has zero rows (the internal `for` loop over sub-batches simply doesn't execute). The subsequent `invalidate_for_table()` call has no guard checking whether any rows were actually written.

### Suggested fix

Add a check for non-empty batches before invalidating. For example:

```rust
Ok(()) => {
    // ... existing commit logic ...

    // Only invalidate cache if the batch contained actual changes
    if change_envelope.change_batch.record.num_rows() > 0 {
        if let Some(cache_provider_ref) = caching.as_ref()
            && let Some(cache_provider) = cache_provider_ref.upgrade()
            && let Err(e) =
                cache_provider.invalidate_for_table(dataset_name.clone())
            && !self.runtime_status.is_shutdown()
        {
            // ...
        }
    }
```

Alternatively, `write_change()` could return a signal indicating whether rows were actually written, avoiding the need to inspect the batch externally.

### Impact

This affects any deployment using `refresh_mode: changes` with the SQL results cache enabled. The cache is effectively disabled for these datasets, providing no benefit despite being configured and consuming memory. Datasets using `refresh_mode: full` are not affected (they only invalidate once per refresh cycle).

krinart:
## 📝 Summary

In case of CDC only invalidate cache in case when batch is non empty. Extend `RefreshTask::write_change` to return whether there was an update.

Plus unit tests

Closes https://github.com/spiceai/spiceai/issues/9472

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task/changes.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task/changes.rs
@@ -113,17 +113,17 @@ impl RefreshTask {
 
                             if had_change
                                 && let Some(cache_provider_ref) = caching.as_ref()
-                                    && let Some(cache_provider) = cache_provider_ref.upgrade()
-                                    && let Err(e) =
-                                        cache_provider.invalidate_for_table(dataset_name.clone())
-                                    && !self.runtime_status.is_shutdown()
-                                {
-                                    // No cache provider means runtime is shutting down and cache is already cleaned up
-                                    tracing::error!(
-                                        "Failed to invalidate cached results for dataset {}: {e}",
-                                        &dataset_name.to_string()
-                                    );
-                                }
+                                && let Some(cache_provider) = cache_provider_ref.upgrade()
+                                && let Err(e) =
+                                    cache_provider.invalidate_for_table(dataset_name.clone())
+                                && !self.runtime_status.is_shutdown()
+                            {
+                                // No cache provider means runtime is shutting down and cache is already cleaned up
+                                tracing::error!(
+                                    "Failed to invalidate cached results for dataset {}: {e}",
+                                    &dataset_name.to_string()
+                                );
+                            }
                         }
                         Err(e) => {
                             let error_message = format_datafusion_error(&e);
--- a/crates/runtime/src/accelerated_table/refresh_task/changes.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task/changes.rs
@@ -860,7 +860,11 @@ mod tests {
         let task = make_refresh_task(make_mem_table() as Arc<dyn TableProvider>);
         let change_batch =
             create_test_change_batch(vec!["c"], &[vec!["id"]], vec![1], vec![Some("Alice")]);
-        assert!(task.write_change(change_batch).await.unwrap());
+        assert!(
+            task.write_change(change_batch)
+                .await
+                .expect("write_change should succeed")
+        );
     }
 
     #[tokio::test]
@@ -869,7 +873,11 @@ mod tests {
         let task = make_refresh_task(adapter as Arc<dyn TableProvider>);
         let change_batch =
             create_test_change_batch(vec!["d"], &[vec!["id"]], vec![1], vec![Some("Alice")]);
-        assert!(task.write_change(change_batch).await.unwrap());
+        assert!(
+            task.write_change(change_batch)
+                .await
+                .expect("write_change should succeed")
+        );
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -160,13 +161,12 @@ impl RefreshTask {
         Ok(())
     }
 
+    /// Returns `true` if any actual data change (upsert or delete) was written.
     async fn write_change(
         &self,
         change_batch: ChangeBatch,
-    ) -> crate::accelerated_table::Result<()> {
+    ) -> crate::accelerated_table::Result<bool> {
```
phillipleblanc: I am not a fan of raw booleans - it is usually always better to return an enum with two values as that makes it very intentional, i.e. something like:

```rust
   enum WriteChangeResult {
       DataWritten,
       NoChange,
   }
 ```

 It makes the call site self-documenting:

 ```rust
   Ok(WriteChangeResult::DataWritten) => { /* invalidate cache */ }
```

lukekim (approved):

PR #9659, opened 2026-03-06
Use `max_partitions_per_executor ` in `allocate_initial_partitions`. 

Jeadie:
## 📝 Summary
`runtime.scheduler.max_partitions_per_executor` config was not used in `ClusterService::allocate_initial_partitions`.
The original hardcoded 10 was still being used.


crates/runtime/src/cluster/service.rs
```diff
@@ -634,6 +653,7 @@ impl ClusterService for ClusterServiceImpl {
                                 }
                             }
                         }
+                        total_assigned += partitions.len();
```
phillipleblanc: As Copilot mentions, this would introduce a regression for reconnecting executors. We should only count newly assigned partitions to increment here, not including any that were previously assigned to the executor before it disconnected.

We could return a struct:

 ```rust
   struct AllocationResult {
       previously_assigned: Vec<PartitionValue>,
       newly_assigned: Vec<PartitionValue>,
   }
 ```

 Then the caller would do:

 ```rust
   total_assigned += result.newly_assigned.len();
   // But still send ALL partitions (old + new) to the executor
```

phillipleblanc (approved):

PR #9713, opened 2026-03-11
Unassign partition from executors that are unregistered. 

Jeadie:
## 📝 Summary
For Distributed query, when an executor disconnects, we should unassign its acceleration partitions. We remove them in two places:
 1. After the control stream has disconnected (the scheduler has its connection to executor disconnect).
 2. On the periodic partition management task. Any executor not in the registry will be removed.
 
 TODO: need to handle > 1 scheduler.


Jeadie:
Closing - work superseded / no longer needed.

PR #9719, opened 2026-03-11
Improve test_accelerated_view_on_zero_results_use_source test

sgrebnov:
## 📝 Summary

Improve `test_accelerated_view_on_zero_results_use_source ` to. exercise `use_source` path when acceleration returns empty result.


crates/runtime/tests/view/mod.rs
```diff
@@ -802,10 +802,14 @@ async fn test_accelerated_view_on_zero_results_use_source() -> Result<(), anyhow
...
 
+            // Simulate data source update: now contains new row Charlie (id=3) that does NOT exist in accelerated view
+            let test_csv = "id,name,value\n1,Alice,100\n2,Bob,200\n3,Charlie,300";
+            std::fs::write(&csv_path, test_csv).expect("write file");
+
             // Query the accelerated view - data should be available
             let query_result = rt
                 .datafusion()
-                .query_builder("SELECT * FROM zero_results_view ORDER BY id")
+                .query_builder("SELECT * FROM zero_results_view WHERE id = 3")
```
lukekim, line 831: Why not just validate the entire result set?

sgrebnov: Updated to include both queries: with and without filter to test both  execution path: acceleration only and source

lukekim (approved):

PR #9734, opened 2026-03-12
Add executor resource and replica support to cloud app config

ewgenius:
This pull request introduces support for configuring an app's executor (including its resources and replicas) independently from the main app (scheduler) in the Spice Cloud client and the `spidapter` tool. It adds a new `AppExecutor` struct, updates API request types to include executor settings, and enhances environment variable parsing for flexible resource configuration.

**API and Data Model Enhancements:**
- Added a new `AppExecutor` struct to represent executor-specific resource and replica settings, and integrated it into `AppConfig`, `CreateAppRequest`, and `UpdateAppRequest` to allow clients to specify executor configuration separately from the main app. [[1]](diffhunk://#diff-2800e1e132e30abba117a5144e0acd591dec1660dcdd8439944c9287351438beR55-R62) [[2]](diffhunk://#diff-2800e1e132e30abba117a5144e0acd591dec1660dcdd8439944c9287351438beR72) [[3]](diffhunk://#diff-2800e1e132e30abba117a5144e0acd591dec1660dcdd8439944c9287351438beR113-R117) [[4]](diffhunk://#diff-2800e1e132e30abba117a5144e0acd591dec1660dcdd8439944c9287351438beR136-R137)

**Environment Variable and Resource Configuration:**
- Implemented helper functions in `spidapter` for parsing environment variables for resource and replica overrides, with robust error handling and default values, enabling independent configuration for both scheduler and executor via environment variables.
- Refactored resource allocation logic in `ensure_spice_cloud_app` to use these helpers, supporting granular overrides for both scheduler and executor resource fields and replica counts. [[1]](diffhunk://#diff-0a17a9dac1f1b17bfc8c2809eaedb97804d63f55195136ed4b76460ac6411b74L71-R180) [[2]](diffhunk://#diff-0a17a9dac1f1b17bfc8c2809eaedb97804d63f55195136ed4b76460ac6411b74L84-R194)

**Codebase Maintenance:**
- Updated imports in `spidapter` to include the new `AppExecutor` type.
- Ensured that `apply_spicepod_to_app` sets the new `executor` field to `None` when updating app configuration, maintaining backward compatibility.

### DIFF
--- a/bin/spice/src/commands/cloud/client.rs
+++ b/bin/spice/src/commands/cloud/client.rs
@@ -138,7 +138,10 @@ impl CloudClient {
             visibility: visibility.to_string(),
             cname: None,
             tags: None,
+            replicas: None,
             resources: None,
+            executor_replicas: None,
+            executor_resources: None,
         };
         self.inner.create_app(&request).await.map_err(into_cli)
     }
@@ -161,6 +164,8 @@ impl CloudClient {
             region: region.map(String::from),
             spicepod: None,
             resources: None,
+            executor_replicas: None,
+            executor_resources: None,
         };
         self.inner
             .update_app(app.id, &request)
--- a/bin/spice/src/commands/cloud/client.rs
+++ b/bin/spice/src/commands/cloud/client.rs
@@ -140,8 +140,7 @@ impl CloudClient {
             tags: None,
             replicas: None,
             resources: None,
-            executor_replicas: None,
-            executor_resources: None,
+            executor: None,
         };
         self.inner.create_app(&request).await.map_err(into_cli)
     }
@@ -164,8 +163,7 @@ impl CloudClient {
             region: region.map(String::from),
             spicepod: None,
             resources: None,
-            executor_replicas: None,
-            executor_resources: None,
+            executor: None,
         };
         self.inner
             .update_app(app.id, &request)
--- a/tools/spidapter/src/commands/mod.rs
+++ b/tools/spidapter/src/commands/mod.rs
@@ -57,6 +57,66 @@ pub(crate) fn build_cloud_client(api_url_override: Option<&str>) -> anyhow::Resu
         .with_timeout(Duration::from_secs(600))?)
 }
 
+/// Parse an optional non-negative integer from an environment variable.
+///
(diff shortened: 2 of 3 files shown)


tools/spidapter/src/commands/mod.rs
```diff
@@ -68,8 +128,26 @@ pub(crate) async fn ensure_spice_cloud_app(
...
-    let memory_limit =
-        std::env::var("SPIDAPTER_APP_MEMORY_LIMIT").unwrap_or_else(|_| "16Gi".to_string());
+    // App (scheduler) resources — default to 16Gi memory / 0.1 cpu request when unset.
+    let resources = env_var_resources(
+        "SPIDAPTER_APP_MEMORY_LIMIT",
+        "SPIDAPTER_APP_CPU_LIMIT",
+        "SPIDAPTER_APP_CPU_REQUEST",
+        "SPIDAPTER_APP_MEMORY_REQUEST",
+    )
+    .unwrap_or_else(default_resources);
```
peasee: I think this will fallback to the default resources if you have only specified say `SPIDAPTER_APP_MEMORY_LIMIT` and nothing else? I'm pretty sure this requires you specify every single limit+request if you specify at least one. We should be able to specify them individually, overriding the defaults individually.

sgrebnov (dismissed):

bin/spice/src/commands/cloud/client.rs
```diff
@@ -138,7 +138,10 @@ impl CloudClient {
             visibility: visibility.to_string(),
             cname: None,
```
lukekim, line 139: We should remove `cname`. It shouldn't be exposed in create app.

tools/spidapter/src/commands/mod.rs
ewgenius: fixed to override, thanks

bin/spice/src/commands/cloud/client.rs
ewgenius: We need to update in API first, because cname is required.

sgrebnov (approved):

peasee (approved):

PR #9739, opened 2026-03-13
Properly set `primary_keys`/`on_conflict` for Cayenne tables

krinart:
## 📝 Summary

Properly set `primary_keys`/`on_conflict` for Cayenne tables - this will allow to run upserts using `CommandStatementIngest`

### DIFF
--- a/crates/runtime/src/datafusion/cayenne_ddl/physical_plans.rs
+++ b/crates/runtime/src/datafusion/cayenne_ddl/physical_plans.rs
@@ -191,6 +191,7 @@ impl fmt::Debug for CayenneCreateTableExec {
 
 impl CayenneCreateTableExec {
     #[must_use]
+    #[expect(clippy::too_many_arguments)]
     pub fn new(
         table_name: String,
         arrow_schema: Arc<Schema>,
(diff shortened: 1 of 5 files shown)


crates/runtime/src/datafusion/cayenne_ddl/physical_plans.rs
```diff
@@ -359,11 +366,18 @@ impl ExecutionPlan for CayenneCreateTableExec {
             );
 
             // Create table options with namespace-prefixed name
```
phillipleblanc: This comment should move to before the create table options

phillipleblanc (dismissed):

crates/runtime/src/datafusion/cayenne_ddl/physical_plans.rs
sgrebnov: Fixed

sgrebnov (dismissed):

crates/runtime/src/datafusion/cayenne_ddl/physical_plans.rs
```diff
@@ -358,12 +365,20 @@ impl ExecutionPlan for CayenneCreateTableExec {
                 data_base_path.trim_end_matches('/').to_string() + "/"
             );
 
+            let on_conflict = if primary_key.is_empty() {
+                None
+            } else {
+                Some(OnConflict::Upsert(ColumnReference::new(
```
sgrebnov, line 372: Fine for now but this should be configuration-driven imo with default option to error

sgrebnov (approved):

PR #9770, opened 2026-03-16
Chunk metastore batch inserts to respect SQLite parameter limits

phillipleblanc:
## Problem

`add_insert_records_batch` builds a single `INSERT OR REPLACE` with **4 SQL parameters per row** (`insert_record_id`, `table_id`, `pk_bytes`, `sequence_number`). When a large upsert has more than ~8,000 conflicting primary keys, the total parameter count exceeds SQLite's compile-time `SQLITE_MAX_VARIABLE_NUMBER` limit (**32,766**), causing the prepared statement to fail.

For example, the spicebench TPC-H lineitem table upserts 23,400 rows, producing **93,600 parameters** — nearly 3× the limit. In Spice Cloud, this manifests as a server crash returning gRPC `UNAVAILABLE` to the client.

## Solution

Split large batches into chunks of **8,000 rows** (32,000 params), each executed as a separate parameterized `INSERT OR REPLACE`. When chunking is needed, all chunks are wrapped in an explicit `BEGIN`/`COMMIT` transaction so the operation remains atomic — either every chunk is applied or none is (with `ROLLBACK` on error).

Small batches (≤ 8,000 rows) take the existing single-statement path with no transaction overhead.

The chunk execution logic is extracted into a shared `insert_records_chunk` helper on `CayenneCatalog`.

## Changes

- **`crates/cayenne/src/cayenne_catalog.rs`**
  - Extract `insert_records_chunk` helper for a single parameterized INSERT OR REPLACE
  - Chunk `add_insert_records_batch` to stay within `SQLITE_MAX_VARIABLE_NUMBER`
  - Wrap multi-chunk execution in `BEGIN`/`COMMIT` with `ROLLBACK` on error
- **`crates/cayenne/tests/large_upsert_test.rs`** — regression test that upserts 10,000 rows (40,000 params), which fails without the fix:

```
variable number must be between ?1 and ?32766
```

## Testing

- New test `test_large_upsert_exceeds_sqlite_param_limit_sqlite` — **fails without fix, passes with fix**
- All existing cayenne tests pass

### DIFF
--- a/crates/cayenne/tests/large_upsert_test.rs
+++ b/crates/cayenne/tests/large_upsert_test.rs
@@ -1,0 +1,129 @@
+/*
+Copyright 2026 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#![allow(clippy::expect_used)]
+#![allow(clippy::clone_on_ref_ptr)]
+
+//! Regression test: upserting more than `SQLITE_MAX_VARIABLE_NUMBER` / 4 rows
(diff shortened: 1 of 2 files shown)


crates/cayenne/tests/large_upsert_test.rs
```diff
@@ -0,0 +1,129 @@
+/*
+Copyright 2025 The Spice.ai OSS Authors
```
phillipleblanc:
```suggestion
Copyright 2026 The Spice.ai OSS Authors
```

krinart (approved):

PR #9839, opened 2026-03-19
[spidapter] Check for scheduler/executor

Jeadie:


Jeadie:
Not going to work. Executor will report metrics regardless if scheduler is unaware of them

PR #9864, opened 2026-03-20
Fix ODBC queries silently returning 0 rows on query failure

### LINKED ISSUE #8379: benchmark: `tpch - federated/odbc[athena].yaml` returned 0 rows via Flight
## Describe the bug

During the `v1.10.0-rc.1` release, the benchmark run for `tpch - federated/odbc[athena].yaml` had a single query failure on TPCH Q2 with the following message:

`tpch_q2: Worker 0 - Query 'tpch_q2' returned 0 rows via Flight`

The other queries ran without issue.

Link: https://github.com/spiceai/spiceai/actions/runs/19877535534/job/56968284625

sgrebnov:
## 📝 Summary

Fixes #8379

ODBC-federated queries that fail on the remote database (e.g., Athena's "Given correlated subquery is not supported") silently returned 0 rows instead of propagating the error to the caller.

**Root Cause:** `join_handle.await` only checked for task panic/cancellation (`Err`) but ignored query execution errors (`Ok(Err)`), so driver errors were silently dropped and the stream ended with 0 rows.

After fix:
> Query Error: Failed to execute ODBC query: SQLExecute failed: State: HY000, Native error: 0, Message: [AmazonAthena][AthenaClientError]: ErrorType: 1200, ExceptionMessage: NOT_SUPPORTED: line 1:326: Given correlated subquery is not supported

## 🔗 Related

Fixes #8379

### DIFF
--- a/crates/db_connection_pool/src/dbconnection/odbcconn.rs
+++ b/crates/db_connection_pool/src/dbconnection/odbcconn.rs
@@ -243,7 +243,14 @@ where
                 // Verify the statement actually produced a result set before creating a cursor
                 // Reading from an invalid cursor would segfault in the ODBC C driver.
                 match statement.num_result_cols() {
-                    SqlResult::Success(_) | SqlResult::SuccessWithInfo(_) => {}
+                    SqlResult::Success(cols) | SqlResult::SuccessWithInfo(cols) => {
+                        if cols == 0 {
+                            return Err(Error::ODBCAPIErrorNoSource {
+                                message: "Statement did not produce a result set (0 columns reported)".to_string(),
+                            }
+                            .into());
+                        }
+                    }
                     SqlResult::Error { function } => {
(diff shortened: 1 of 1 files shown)


lukekim (dismissed):

crates/db_connection_pool/src/dbconnection/odbcconn.rs
```diff
@@ -78,7 +79,7 @@ pub enum Error {
     ArrowODBCError { source: arrow_odbc::Error },
     #[snafu(display("ODBC connection error: {source}"))]
     ODBCAPIError { source: odbc_api::Error },
-    #[snafu(display("ODBC connection error: {message}"))]
+    #[snafu(display("{message}"))]
```
sgrebnov, line 82: `ODBC connection error:` part is redundant

lukekim (approved):

PR #9878, opened 2026-03-22
fix(secrets): load valid .env entries even when some lines are malformed

### LINKED ISSUE #6306: Bug: Secrets silently fail to load from .env file if any secret has invalid formatting
## Describe the bug

## To Reproduce

2. Click on '...'
3. Scroll down to '...'

## Expected behavior

### Spicepod

### Output of `explain query`

E.g. `explain select 1`

```bash
```

### spice, spiced, OS info

- OS info: `uname -a`

- [ ] Yes
- [ ] No

mvanhorn:
## 📝 Summary

`dotenvy::from_path()` stops processing on the first malformed line, preventing all subsequent valid secrets from loading. This switches to `from_path_iter()` to process each line independently.

**Before:** One bad line in `.env` -> zero secrets loaded, generic warning
**After:** Bad lines warn with the specific content, valid lines still load

The fix is in `crates/runtime-secrets/src/stores/env.rs:load()`. A new `load_from_iter()` helper iterates entries one at a time: `Ok((key, val))` sets the env var, `Err(LineParse)` warns and continues, other errors warn and stop that file.

## 🔗 Related

Closes #6306

## 👀 Notes for Reviewers

The same pattern is applied to all three dotenv sources (explicit path, `.env.local`, `.env`). The `unsafe` `set_var` call has a SAFETY comment explaining the single-threaded init context. Includes a regression test that loads a file with a bad line between two valid entries and verifies both valid entries are present.

This contribution was developed with AI assistance (Claude Code + Codex).


crates/runtime-secrets/src/stores/env.rs
```diff
@@ -58,29 +58,56 @@ pub struct EnvSecretStore {
...
-    fn load(&self) {
-        if let Some(path) = &self.path {
-            match dotenvy::from_path(path) {
-                Ok(()) => return,
-                Err(err) => {
-                    if matches!(err, dotenvy::Error::LineParse(_, _)) {
-                        tracing::warn!("{err}");
-                    } else {
-                        tracing::warn!("Error opening path {}: {err}", path.display());
+    fn load_from_iter<I>(source: &str, iter: Result<I, dotenvy::Error>, warn_on_open_error: bool)
```
Jeadie: If `iter` is not a result, we won't need to provide `warn_on_open_error`?
```diff 
- fn load_from_iter<I>(source: &str, iter: Result<I, dotenvy::Error>, warn_on_open_error: bool)
+ fn load_from_iter<I>(source: &str, iter: I)
```

mvanhorn: Simplified in 65236f0 - load_from_iter now takes the iterator directly and callers handle the Result themselves. Removed warn_on_open_error since each call site controls its own error path.

Jeadie (approved):

sgrebnov (approved):

lukekim (approved):

PR #9915, opened 2026-03-24
Implement DML for DynamoDBTableProvider

krinart:
## 📝 Summary

* Add support for `delete_from` and `update` to `DynamoDBTableProvider`.

* Add configurable parallelism (added `write_parallelism` configuration parameter)

* Temporarily add custom handling for `DynamoDBTableProvider` in `get_deletion_provider` until we get rid of `DeletionTableProvider` in favor of DF's native support of `delete_from`

* Tests

Needed for DynamoDB ADBC driver.
Also: bump `docx-rs` version since `zip 7.4.0` version was yanked (using yanked versions is not allowed for new projects such as DynamoDB ADBC driver)

### DIFF
--- a/crates/data_components/src/dynamodb/dml/insert.rs
+++ b/crates/data_components/src/dynamodb/dml/insert.rs
@@ -1,0 +1,144 @@
+/*
+Copyright 2026 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use super::streaming_batch_write;
+use crate::dynamodb::utils::scalar_to_attribute_value;
+use arrow::array::Array;
+use arrow::datatypes::SchemaRef;
+use async_trait::async_trait;
+use aws_sdk_dynamodb::{
+    Client as DbClient,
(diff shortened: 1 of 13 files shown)


crates/data_components/src/dynamodb/dml/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+        .collect();
+
+    let table_name = table_name.to_string();
+    let results: Vec<DataFusionResult<()>> = stream::iter(chunks)
+        .map(|chunk| {
+            let client = Arc::clone(client);
+            let table_name = table_name.clone();
+            async move {
+                client
+                    .batch_write_item()
```
phillipleblanc: This can succeed with `unprocessed_items` that needs to be handled, otherwise we could drop data.

crates/data_components/src/dynamodb/dml/insert.rs
```diff
@@ -0,0 +1,135 @@
...
+        &self.schema
+    }
+
+    async fn write_all(
+        &self,
+        mut data: SendableRecordBatchStream,
+        _context: &Arc<TaskContext>,
+    ) -> DataFusionResult<u64> {
+        let mut row_count: u64 = 0;
+        let mut write_requests: Vec<WriteRequest> = Vec::new();
```
phillipleblanc: This batches all write requests in memory before calling `parallel_batch_write_items`. This might be better modeled as a stream where the downstream consumer can pull batches as soon as they are ready.

crates/data_components/src/dynamodb/dml/mod.rs
krinart: Fixed, thanks

crates/data_components/src/dynamodb/dml/insert.rs
krinart: Done

sgrebnov (dismissed):

sgrebnov (approved):

PR #9922, opened 2026-03-25
fix: Allow specifying Cayenne catalog directories in spidapter

peasee:
## 📝 Summary

* Allow specifying `cayenne_data_dir` and `cayenne_metadata_dir` from environment variables in spidapter


Jeadie (changes requested):
Env variables should be flags with env options

phillipleblanc (dismissed):

ewgenius (approved):

Jeadie (approved):

Issue #9926, opened 2026-03-25
`FallbackOnZeroResultsScanExec` fallback path skips physical optimizer, causing slower scans


sgrebnov:
## Describe the bug

When `on_zero_results: use_source` triggers fallback, the federated scan plan is executed directly without running DataFusion's physical optimizer. The optimizer's `EnforceDistribution` rule normally splits single file groups into N parallel partitions (e.g., 1 group → 16 groups). 

Observed: ~4x slower Delta Lake tables scans with `FallbackOnZeroResultsScanExec`. EXPLAIN shows 16 file groups for direct queries vs 1 file group in the fallback path.

sgrebnov:
Fixed by https://github.com/spiceai/spiceai/pull/9927

PR #9927, opened 2026-03-25
Run physical optimizer on `FallbackOnZeroResultsScanExec` fallback plan

sgrebnov:
## 📝 Summary

When `on_zero_results: use_source` triggers a fallback from the accelerated table to the federated source, the federated scan plan bypasses DataFusion's physical optimizer. This means `EnforceDistribution` never splits file groups into parallel partitions — the entire scan runs on a single thread.

For a normal query, DataFusion's optimizer splits a single `FileGroup` into N parallel groups (matching `target_partitions`). The fallback path called `TableProvider::scan()` directly and executed the raw unoptimized plan, resulting in ~4x slower queries on multi-core machines for Delta Lake

**Fix**

After obtaining the fallback plan from `scan()`, run `SessionState::physical_optimizers()` rules on it before execution. This lets `EnforceDistribution` (and other rules) parallelize the scan. The result is then coalesced back to a single partition if needed.

**Testing**

Verified with Delta Lake + DuckDB acceleration (`on_zero_results: use_source`):
- **Before**: fallback plan shows `file_groups={1 group: [...]}`  
- **After**: fallback plan shows `file_groups={16 groups: [...]}` under `CoalescePartitionsExec`

## 🔗 Related

Fixes https://github.com/spiceai/spiceai/issues/9926

### DIFF
--- a/crates/runtime-datafusion/src/execution_plan/snapshots/runtime_datafusion__execution_plan__fallback_on_zero_results__tests__optimize_physical_plan_tests__optimizer_splits_single_file_group.snap
+++ b/crates/runtime-datafusion/src/execution_plan/snapshots/runtime_datafusion__execution_plan__fallback_on_zero_results__tests__optimize_physical_plan_tests__optimizer_splits_single_file_group.snap
@@ -1,0 +1,6 @@
+---
+source: crates/runtime-datafusion/src/execution_plan/fallback_on_zero_results.rs
+expression: plan_display
+---
+CoalescePartitionsExec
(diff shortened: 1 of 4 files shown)


crates/runtime-datafusion/src/execution_plan/snapshots/runtime_datafusion__execution_plan__fallback_on_zero_results__tests__optimize_physical_plan_tests__optimizer_splits_single_file_group.snap
```diff
@@ -0,0 +1,6 @@
+---
+source: crates/runtime-datafusion/src/execution_plan/fallback_on_zero_results.rs
+expression: plan_display
+---
+CoalescePartitionsExec
```
sgrebnov, line 5: This validates that:
1. Optimization performed and we now see multiple `file_groups={4 groups:`
2. `CoalescePartitionsExec` is correctly added

lukekim (approved):

PR #9974, opened 2026-03-26
ci: consolidate pr-develop workflow and add triggers to test workflows

lukekim:
Consolidate pr-develop workflow into single job. Remove secret-dependent steps. Add merge_group and pull_request triggers to test workflows for trunk PRs.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,8 +1,6 @@
 name: 'CodeQL'
 
 on:
-  merge_group:
-    types: [checks_requested]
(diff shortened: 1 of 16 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -1,12 +1,19 @@
 name: 'CodeQL'
 
 on:
+  merge_group:
+    types: [checks_requested]
   workflow_dispatch:
```
lukekim, line 4:
```suggestion
on:
  workflow_dispatch:
```

PR #9975, opened 2026-03-26
fix: Rewrite decimal BETWEEN to CAST AS REAL for Turso acceleration

### LINKED ISSUE #9872: TPCH Q6 incorrect result with Turso acceleration
## Describe the bug

Similar to SQLite, but Turso does not have `decimal_cmp()` so we can apply similar to SQLite fix
 - https://github.com/spiceai/spiceai/issues/5646

>Query 'tpch_q6' validation failed: DataMismatch { column: "revenue", row_number: 1, expected: "\"123141078.2283\"", actual: "\"75207768.1855\"" }

claudespice:
## Summary

Fixes #9872.

- TPCH Q6 gives incorrect results with Turso acceleration because Turso (like SQLite) may perform text-based comparisons on decimal values in `BETWEEN` expressions
- Adds a `TursoBetweenVisitor` AST visitor that rewrites `col BETWEEN 0.06 AND 0.08` into `CAST(col AS REAL) >= CAST(0.06 AS REAL) AND CAST(col AS REAL) <= CAST(0.08 AS REAL)`
- Unlike the SQLite fix which uses `decimal_cmp()` (a custom SQLite extension not available in Turso), this uses `CAST AS REAL` to ensure numeric comparison semantics
- The visitor is applied in all three SQL generation paths: scan, deletion, and federation

## Test plan

- [x] Unit tests for `TursoBetweenVisitor` covering: numeric bounds rewrite, NOT BETWEEN, string bounds (no rewrite), mixed bounds (no rewrite), arithmetic expression bounds, and exact SQL output format verification
- [ ] Run TPCH Q6 benchmark with Turso acceleration to verify correct `revenue` result (`123141078.2283`)


claudespice:
Closing in favor of #10003 which is a cleaner rewrite of the same fix.

PR #9976, opened 2026-03-26
chore: Update turso dependency from 0.5.0 to 0.5.1

claudespice:
## Summary
- Bump the `turso` crate dependency from `0.5.0` to `0.5.1` (latest stable release)

## Test plan
- [ ] CI build and tests pass
- [ ] No breaking API changes in turso 0.5.1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -316,7 +316,7 @@ tracing = "0.1.43"
 tracing-futures = { version = "0.2.5", features = ["futures-03"] }
 tracing-opentelemetry = "0.32"
 tracing-subscriber = { version = "0.3.22", features = ["env-filter"] }
-turso = { version = "0.5.0", default-features = false }
+turso = { version = "0.5.1", default-features = false }
 turso-shared = { path = "crates/turso-shared" }
 twox-hash = "2.1.2"
 url = { version = "2.5", features = ["serde"] }


Jeadie (approved):

crates/runtime/src/catalogconnector/iceberg.rs
```diff
@@ -806,12 +806,13 @@ pub(crate) fn build_opendal_operator(
         if let Some(endpoint) = props.get("s3.endpoint") {
             config.endpoint = Some(endpoint.clone());
         }
```
sgrebnov, line 808: This is unrelated

Cargo.toml
```diff
@@ -316,7 +316,7 @@ tracing = "0.1.43"
 tracing-futures = { version = "0.2.5", features = ["futures-03"] }
 tracing-opentelemetry = "0.32"
 tracing-subscriber = { version = "0.3.22", features = ["env-filter"] }
-turso = { version = "0.5.0", default-features = false }
```
sgrebnov, line 319: Should there be any cargo.lock changes associated?

claudespice:
Rebased onto develop — removed unrelated iceberg.rs and Makefile changes. Cargo.lock already has turso 0.5.1 from develop base, so no additional lockfile changes needed.

crates/runtime/src/catalogconnector/iceberg.rs
claudespice: This was from an older version of the branch before rebasing. The branch has been rebased on latest develop and iceberg.rs is no longer in this PR's diff.

Cargo.toml
claudespice: No Cargo.lock changes are needed — develop's Cargo.lock already has turso 0.5.1 (the semver spec `"0.5.0"` allowed it). This PR just aligns Cargo.toml to explicitly require 0.5.1 as the minimum. Verified with `cargo update --locked` that the lockfile is consistent.

PR #9984, opened 2026-03-27
Add caching retention

### LINKED ISSUE #9150: HTTP caching acceleration should support retention for cached items
if `caching_stale_if_error: enabled` is NOT configured recordings must be cleared periodically based on `fetched_at` value.

Ensure we don't clear items within SWR ttl which could be large

krinart:
## 📝 Summary

Add periodic cache retention for HTTP caching acceleration that automatically cleans up expired cache entries based on `fetched_at` timestamps.

* Only works when `caching_stale_if_error` is disabled.
* If user defines `retention_period`: overrides it and shows a warning

Closes https://github.com/spiceai/spiceai/issues/9150


crates/runtime/src/accelerated_table/cache_retention.rs
```diff
@@ -0,0 +1,351 @@
...
+impl super::AcceleratedTable {
+    /// Periodically deletes expired cache entries from the accelerator.
+    ///
+    /// Expired entries are those where `fetched_at < now - max_age - stale_while_revalidate`.
+    /// This matches the `CacheFreshness::Expired` classification from `check_cache_freshness`.
+    ///
+    /// This task should only be spawned when `caching_stale_if_error` is NOT enabled,
+    /// since expired entries may still serve as fallback data when `stale_if_error` is enabled.
+    #[expect(clippy::cast_possible_truncation)]
+    pub(crate) async fn start_cache_retention_check(
```
phillipleblanc: This is duplicating logic that we already have for time based retention in `retention.rs`. This cache based retention IS the same as time based retention - the time column is `fetched_at`. And we should prevent people from specifying a time based retention on a caching table.

phillipleblanc (dismissed):
Merge this with the existing retention logic we already have.

crates/runtime/src/accelerated_table/cache_retention.rs
krinart: Done

krinart:
@phillipleblanc fixed. Thanks for the feedback!

sgrebnov (approved):

PR #9987, opened 2026-03-27
fix: improve Databricks schema error detection and messages

### LINKED ISSUE #9985: Databricks connector: improve error message when table schema is missing from Unity Catalog
## Summary

When a Databricks table exists in Unity Catalog but has no column metadata loaded (e.g. tables that have never been queried), the connector fails with a cryptic, non-actionable error:

```
WARN runtime::init::dataset: Failed to load the dataset jobs (databricks). Failed to initialize SQL table: Unable to get schema: Unable to retrieve schema: result.data_array
```

## Root Cause

The connector queries `information_schema.columns` via the SQL Statements API. When a table has never been queried, Unity Catalog returns no rows — the `result.data_array` field in the response is empty or absent. The schema parser at `crates/data_components/src/databricks/sql_warehouse.rs:561–570` hits its `ok_or_else` path and surfaces the raw JSON field path as the error reason.

## Target Error Message

For the missing-schema case, following the pattern of `InvalidTableName` / `InvalidConfiguration`:

```
Failed to load the dataset jobs (databricks). The table 'jobs' has no column metadata registered in Unity Catalog. This can happen with tables that have never been queried. To register the schema, run in Databricks: SELECT * FROM <catalog>.<schema>.jobs LIMIT 1 (or REPAIR TABLE <catalog>.<schema>.jobs SYNC METADATA for user-owned tables). For details, visit: https://spiceai.org/docs/components/data-connectors/databricks
```

For the malformed-row case (unexpected API response shape), following the pattern of `Internal` / `UnableToGetSchemaInternal`:

```
Failed to load the dataset jobs (databricks). Failed to infer the table schema. Report a bug on GitHub (https://github.com/spiceai/spiceai/issues) and reference the error: <source>
```

ewgenius:
## Summary

- Replaces the overloaded `UnableToRetrieveSchema` error variant with two purpose-built variants that follow the error handling guidelines
- `TableSchemaNotRegistered`: surfaced when `result.data_array` is absent (`null` or missing) in the SQL Statements API response, which occurs when a table has no column metadata registered (e.g. never been queried)
- `UnexpectedSchemaResponse`: surfaced when the API response is structurally malformed — missing the top-level `result` field, `data_array` is the wrong type, or individual rows have an unexpected shape — with a bug report link; the internal `reason` is kept on the struct for `Debug`/logging but excluded from the user-facing `Display`

## Before

```
Failed to load the dataset jobs (databricks). Failed to initialize SQL table: Unable to get schema: Unable to retrieve schema: result.data_array
```

## After

```
Failed to load the dataset jobs (databricks). Failed to initialize SQL table: Unable to get schema: The table 'jobs' has no column metadata registered. Run 'SELECT * FROM jobs LIMIT 1' in Databricks to populate the schema. For details, visit: https://spiceai.org/docs/components/data-connectors/databricks
```

Closes #9985


crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -62,8 +62,15 @@ pub enum Error {
     #[snafu(display("Databricks datatype {ty} not supported"))]
     UnsupportedType { ty: String },
 
-    #[snafu(display("Unable to retrieve schema: {reason}"))]
-    UnableToRetrieveSchema { reason: String },
+    #[snafu(display(
+        "The table '{dataset_name}' has no column metadata registered. For details, visit: https://spiceai.org/docs/components/data-connectors/databricks"
```
phillipleblanc: The docs don't mention anything about no column metadata. Given we already know what this is, we should provide a clear action for them to resolve the error (i.e. something like populating the databricks catalog)

lukekim (approved):

PR #10001, opened 2026-03-27
fix: Defer cache size check until after encoding for compressed results

### LINKED ISSUE #8508: Encoded results cache does not cache when un-encoded results are too large
## Describe the bug

When using results cache with `zstd` encoding, some entries are not cached because the cache size is calculated before the encoding is applied - in the case of zstd, the cache size is applied before compression.

Once encoding is applied, the entry would have fit into the results cache.

## To Reproduce

1. Configure results cache with zstd encoding
2. Run a query that returns a large uncompressed result that would not fit inside the configured results cache size, when the compressed result would
3. See behaviour

## Expected behavior

If the encoded result fits into memory, the result should be stored in the results cache.

### Spicepod

### Output of `explain query`

E.g. `explain select 1`

```bash
```

### spice, spiced, OS info

- OS info: `uname -a`

- [ ] Yes
- [ ] No

claudespice:
## Summary
- When using results cache with zstd encoding, entries were rejected based on their uncompressed size even though the compressed size would fit within the cache limit
- This fix defers the size check until after encoding is applied, so compressed results that fit are correctly cached
- When no encoder is present, the existing behavior of eagerly clearing oversized records is preserved

## Changes
- [x] Modified `to_cached_record_batch_stream()` to keep accumulating records when an encoder is present, even if uncompressed size exceeds the limit
- [x] Added post-encoding size check using `get_memory_size()` which correctly reflects the encoded (compressed) size
- [x] Added regression test `test_encoded_result_cached_when_compressed_fits` verifying compressed results are cached
- [x] Added test `test_unencoded_oversized_result_not_cached` verifying existing behavior for unencoded oversized results

## Test plan
- [x] New unit test: highly compressible data (300 rows of zeros) with raw size > 2 KiB cache limit is cached when zstd encoding compresses it below the limit
- [x] New unit test: oversized data without encoding is correctly rejected
- [x] All existing cache tests continue to pass

Closes #8508


lukekim (approved):

claudespice:
@copilot /review

claudespice:
@copilot /review

claudespice:
@copilot review

claudespice:
@copilot /review

claudespice:
@copilot /review

claudespice:
@copilot /review

claudespice:
@copilot /review

PR #10009, opened 2026-03-27
Fix ISO8601 time_format with Vortex/Cayenne append refresh

### LINKED ISSUE #9091: Cayenne:  `Failed to cast array of dtype utf8? to vortex.cast(ext(vortex.timestamp, i64, ExtMetadata([0, 0, 0]))?)` error
## Describe the bug

with append mode refresh and  time_format: ISO8601 there is the following error

>2026-01-25T20:46:17.984436Z  WARN runtime::accelerated_table::refresh_task: Failed to load data for dataset sample_data (dynamodb): Failed to get data from the connector. Arrow error: External error: Failed to read Vortex file: Users/sg/playground/temp/dynamodb_append_cayenne/.spice/data/sample_data/1/019bf497-1d43-7441-87db-f66b0e66a3cf/TSrTK0KSDuk7tfWj_0.vortex:
  While evaluating pruning predicate (cast($.max as ext(vortex.timestamp, i64, ExtMetadata([0, 0, 0]))?) <= 2026-01-25T20:43:41) (derived from (cast($ as ext(vortex.timestamp, i64, ExtMetadata([0, 0, 0]))?) > 2026-01-25T20:43:41)):
  Failed to cast array of dtype utf8? to vortex.cast(ext(vortex.timestamp, i64, ExtMetadata([0, 0, 0]))?):
  No compute kernel to cast array vortex.varbinview with dtype utf8? to ext(vortex.timestamp, i64, ExtMetadata([0, 0, 0]))?
Backtrace:
disabled backtrace Ensure the dataset configuration is valid, and try again.

krinart:
## 📝 Summary                                                                                                          
                                                        
  - Fix append refresh failure when using `time_format: ISO8601` (or integer unix timestamps) with Vortex/Cayenne engine                           
  - The root cause: `max_timestamp_df` and filter expressions used CAST to convert columns to timestamps, but Vortex lacks kernels for utf8 → timestamp and bigint → timestamp casts                                                                                                         
  - ISO8601 strings are lexicographically orderable, so string comparison and sort work correctly without any cast
  - Similarly, integer unix timestamps sort correctly as raw integers                                                                            
                                                                                                                                                 
##  Changes                                                                                                                                        
                                                                                                                                                 
  - `crates/util/src/timestamp_filter.rs` — Use string comparison for ISO8601 instead of CAST                                                      
  - `crates/runtime/src/accelerated_table/refresh_task.rs` — `max_timestamp_df` only CASTs native date/time/timestamp types; skips CAST for integers and strings. Handle `StringArray`/`Int64Array` results in `timestamp_nanos_for_append_query`. Use accelerator schema (not federated) for filter building in `except_existing_records_from`
  - Add integration tests to have complete coverage of acceleration engine and time_format

Closes https://github.com/spiceai/spiceai/issues/9091


crates/util/src/timestamp_filter.rs
```diff
@@ -188,6 +193,35 @@ impl TimestampFilterConvert {
...
 
+/// Convert nanoseconds since Unix epoch to an ISO 8601 string.
+///
+/// Returns a string like `2024-01-25T20:43:41.000000000`.
+#[must_use]
+#[expect(clippy::cast_possible_wrap)]
+pub fn nanos_to_iso8601_string(nanos: u128) -> String {
+    let secs = (nanos / 1_000_000_000) as i64;
+    let subsec_nanos = (nanos % 1_000_000_000) as u32;
+    chrono::DateTime::from_timestamp(secs, subsec_nanos)
```
sgrebnov, line 204: Nit: there is `to_rfc3339` which should return ISO 8601 date, not sure about TZ part that it seems will be returned as well
```
// Returns an RFC 3339 and ISO 8601 date and time string such as `1996-12-19T16:39:57-08:00`.
    #[cfg(feature = "alloc")]
    #[must_use]
    pub fn to_rfc3339(&self) -> String {
```

sgrebnov (approved):

crates/util/src/timestamp_filter.rs
krinart: Correct - it would require dealing with timezone suffix

PR #10010, opened 2026-03-27
fix: Use deterministic key ordering for search primary key tie-breaking

claudespice:
## Summary

- Fix non-deterministic search result ordering caused by `HashMap` serialization in primary key tie-breaking
- Replace `serde_json::to_string(&HashMap)` comparison with a `compare_primary_keys` function that sorts keys before comparing, ensuring stable ordering across process runs

## Problem

Commit `1f2468d48` added primary key tie-breaking to search result sorting for deterministic results. However, `serde_json::to_string` on `HashMap<String, Value>` does not guarantee key ordering — Rust's `HashMap` uses randomized hashing, so identical maps can serialize to different JSON strings across runs. This undermines the deterministic sort guarantee.

## Fix

Introduce `compare_primary_keys` which:
1. Collects and sorts both maps' keys
2. Compares key lists first (structural comparison)
3. Compares values in sorted key order (content comparison)

This ensures identical primary keys always compare as `Equal` and different primary keys always produce the same relative ordering.

## Test plan

- [x] `cargo check -p runtime` passes
- [ ] Existing search tests continue to pass

### DIFF
--- a/crates/runtime/src/search/types.rs
+++ b/crates/runtime/src/search/types.rs
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use std::{collections::HashMap, fmt::Display};
+use std::{cmp::Ordering, collections::HashMap, fmt::Display};
 
 use arrow::error::ArrowError;
 use arrow_tools::format::to_markdown_documents;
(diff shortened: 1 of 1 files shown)


Jeadie:
Bad solution

PR #10016, opened 2026-03-28
fix: Normalize SQL search test scores and update integration test snapshots

claudespice:
## Summary
- Add `_score` normalization to SQL test path in `search.rs` to match HTTP test normalization (2-decimal padded strings), fixing 6 snapshot mismatches where SQL tests returned numeric `_score` values but snapshots expected strings
- Update `HashJoinExec` explain plan snapshots for projection column changes (join keys now included in projection)
- Update s3_vectors data content snapshots for updated index data (new high-scoring results, different result ordering)
- Update `multi_text_column_keywords_response` score value from `"5.05"` to `"5.06"`

## Context
The `Test Search` CI check on PR #9999 (develop → trunk) was failing with 13 test failures across search and s3_vectors tests. Root causes:
1. Commit `cb80ca9b5` changed score normalization to 2-decimal padded strings but only applied it to the HTTP test path, not the SQL test path
2. HashJoinExec projection behavior changed to include join key columns
3. S3 vectors index data was updated, producing different search results

## Test plan
- [x] `make lint` passes locally
- [ ] CI `Test Search` job passes


claudespice:
All CI checks are passing. This PR fixes all 13 failing search integration tests on the develop → trunk PR (#9999).

**Note:** PR #10014 touches some of the same snapshot files with a different approach (updating snapshots to use numeric scores instead of adding normalization). This PR takes the approach of normalizing SQL test results to match the existing HTTP test normalization (2-decimal padded strings), which is more consistent. The two PRs will conflict if both merge.

PR #10022, opened 2026-03-28
fix: Use truncated scores for deterministic search snapshot ordering

claudespice:
## Summary
- Compare truncated (display-precision) scores in `normalize_search_response` instead of raw floating-point scores, preventing non-deterministic ordering when raw scores differ slightly across machines but display identically after truncation to 2 decimal places
- Update `multi_embedding_parent_child` snapshots to reflect the deterministic SQL-level tie-breaking (`ORDER BY score DESC, pk ASC`), which returns `pk=1` instead of `pk=2` for the department match

## Test plan
- [x] `make lint` passes locally
- [ ] CI checks pass on PR

### DIFF
--- a/crates/runtime/tests/models/search.rs
+++ b/crates/runtime/tests/models/search.rs
@@ -195,10 +195,14 @@ fn normalize_search_response(mut json: Value) -> String {
                 return Ordering::Less;
             };
 
-            // Opposite because we want to order descendingly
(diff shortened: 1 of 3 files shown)


claudespice:
Closing - the search snapshot stabilization has already been addressed in 82a893abe with a more comprehensive approach.

PR #10024, opened 2026-03-28
feat: Add spice completions command for shell completion generation

lukekim:
Restores the spice completions command that was lost during the Go-to-Rust CLI migration. Generates shell completion scripts using clap_complete.

Usage: spice completions [shell]

Supported shells: bash, zsh, fish, elvish, powershell. When shell is omitted, it is auto-detected from the SHELL environment variable.

Changes:
- Added clap_complete dependency to the CLI crate
- New completions subcommand module (bin/spice/src/commands/completions.rs)
- Registered the command in the Commands enum and dispatch
- Suppressed the version banner when generating completions output
- Shell is auto-detected from SHELL when not explicitly provided


ewgenius (approved):

phillipleblanc (changes requested):
Need to remove unrelated commits

phillipleblanc (approved):

PR #9948, opened 2026-03-26
fix: Ensure distributed Cayenne DML inserts are forwarded to executors

peasee:
## 📝 Summary

* Ensures that Distributed Cayenne DML `INSERT` operations are forwarded to executors, similar to `UPDATE`/`DELETE` nodes
* Adds better error handling and identification around running Cayenne Catalog in non-distributed mode. Non-distributed Cayenne Catalog is not supported until https://github.com/spiceai/spiceai/issues/9942 is implemented

## 🔗 Related

* Closes #9733

PR #10026, opened 2026-03-29
fix: Prevent data loss in distributed Cayenne INSERT with multi-partition inputs

claudespice:
## Summary
- `DistributedCayenneInsertExec` reads from its child plan using only partition 0 in `execute()`. When the input plan has multiple partitions (e.g., from a partitioned source table or a repartitioned query), data from partitions 1+ was silently dropped.
- Fix: override `required_input_distribution()` to return `SinglePartition`, which causes DataFusion's `EnforceDistribution` optimizer to insert a `CoalescePartitionsExec` when the input has multiple partitions. This ensures all rows are coalesced into a single stream before being forwarded to executors.
- Adds a regression test that verifies the distribution requirement is enforced.

## Root Cause
The `DistributedCayenneInsertExec` physical plan was introduced in #9948. It declares `Partitioning::UnknownPartitioning(1)` (single output partition), but did not override `required_input_distribution()`. The default implementation returns `UnspecifiedDistribution`, so DataFusion's physical optimizer does not insert a coalesce node between a multi-partition child and this plan. As a result, `execute(partition=0, ...)` only reads from the first partition of the input, silently losing data from other partitions.

## Test plan
- [x] New unit test `distributed_insert_requires_single_partition_input` verifies `required_input_distribution()` returns `SinglePartition`
- [x] `cargo check -p runtime --lib` compiles cleanly
- [x] `cargo test -p runtime --lib -- cayenne_ddl::physical_plans::tests` passes (4/4 tests)
- [ ] Integration testing with a multi-partition source INSERT in distributed mode


Jeadie:
This is entirely false

peasee:
> DistributedCayenneInsertExec reads from its child plan using only partition 0 in execute()

This isn't true. The child stream is called with the partition number as supplied during `.execute()`. See: https://github.com/spiceai/spiceai/pull/10026/changes#diff-26cd9f44e86482e3167135f06207f5f64679bfba4d5ba29af89157cc7e8ab07dL1499

PR #10061, opened 2026-03-31
fix: Flight DoPut message decoding and revert DML support changes

peasee:
## 📝 Summary

* Reverts https://github.com/spiceai/spiceai/pull/9931 and https://github.com/spiceai/spiceai/pull/9915 as they introduced regressions for DML
* Fixed a bug introduced by https://github.com/spiceai/spiceai/pull/10015 which caused the first flight record batch to be parsed as a data batch when it might not be


crates/arrow_tools/src/schema.rs
```diff
@@ -186,8 +186,6 @@ fn data_type_contains_dictionary(data_type: &DataType) -> bool {
         DataType::Dictionary(_, _) => true,
         DataType::List(field)
         | DataType::LargeList(field)
-        | DataType::ListView(field)
```
phillipleblanc, line 189: Why these changes?

phillipleblanc (approved):

crates/arrow_tools/src/schema.rs
Jeadie: Was removed in a revert. Will be readded

PR #10053, opened 2026-03-30
fix: Refresh Debezium schema on reload to handle schema evolution

claudespice:
## Summary
- Fixes #9782: Debezium connector now detects schema evolution on runtime reload by peeking at the latest Kafka message using a temporary consumer
- When the source table schema has changed (columns added/removed), the cached metadata is updated and re-persisted to the accelerator before refresh SQL validation runs
- Adds `KafkaConsumer::fetch_latest_json()` to peek at the latest topic message without affecting the main consumer's offsets

## Changes
- [x] `crates/data_components/src/kafka.rs`: Add `fetch_latest_json()` static method and `into_parts()` on `KafkaMessage`
- [x] `crates/runtime/src/dataconnector/debezium.rs`: Add `refresh_schema_from_kafka()` and call it during reload when cached metadata exists
- [x] `crates/data_components/src/debezium/tests/mod.rs`: 4 new tests for schema evolution detection (added column, removed column, no-op, and Arrow schema conversion)
- [x] `crates/runtime/src/datafusion/refresh_sql.rs`: 2 new tests reproducing the exact bug scenario (stale schema fails validation, evolved schema succeeds)

## Test plan
- [x] `detect_schema_evolution_added_column` — verifies new columns are detected in schema diff
- [x] `no_false_schema_evolution_detection` — verifies identical schemas produce no false positive
- [x] `detect_schema_evolution_removed_column` — verifies removed columns are detected
- [x] `evolved_schema_passes_refresh_sql_validation` — verifies evolved Arrow schema includes new columns
- [x] `test_schema_evolution_refresh_sql_validation` — reproduces bug #9782: stale schema causes `ColumnNotFoundInSource` error, evolved schema passes
- [x] `test_schema_evolution_wildcard_refresh_sql` — verifies wildcard `SELECT *` works with both stale and evolved schemas

PR #10062, opened 2026-03-31
fix: Handle Debezium schema evolution on runtime reload

### LINKED ISSUE #9782: Debezium: schema evolution breaks dataset init on runtime reload
## Bug

The Debezium connector infers the Arrow schema **once**, from the first Kafka message at startup, and never updates it. This causes problems when the source table schema changes:

> 2026-03-16T02:55:28.316746Z WARN runtime::init::dataset: Cannot setup the dataset my_dataset (debezium) with an invalid configuration. The refresh SQL is invalid: The column 'resources' is not present in the source table 'spicepod_deployments', valid columns are: id, created_at, updated_at, spicepod_id, spicepod, replicas, branch, commit_sha, commit_message, status, error_message, image_tag, image, registry, cid, created_by, creation_source, cname Rewrite the SQL to only select columns that exist in the source table.

claudespice:
## Summary
- Fixes #9782: Debezium connector now detects schema evolution on reload by peeking at the latest Kafka message using a temporary consumer
- When the source table schema has changed (columns added/removed), stale cached metadata is cleared and the accelerated table is rebuilt with the updated schema
- Missing nullable fields in change events are now handled gracefully (null instead of error) for streaming resilience
- Adds `delete()` method to `DebeziumKafkaSys` across all backends for metadata cleanup

## Changes
- [x] `crates/data_components/src/debezium/change_event.rs`: Add `PartialEq` and `Debug` derives to `Field` and `Schema` for schema comparison
- [x] `crates/data_components/src/debezium/arrow.rs`: Handle missing nullable fields by appending null instead of erroring in `append_value_to_struct_builder`
- [x] `crates/runtime/src/dataconnector/debezium.rs`: Add `has_schema_evolved()`, `peek_latest_schema_fields()`, and `delete_metadata_from_accelerator()` functions; restructure cached metadata path to detect and handle schema evolution
- [x] `crates/runtime/src/dataaccelerator/spice_sys/debezium_kafka/{mod,duckdb,sqlite,postgres,turso}.rs`: Add `delete()` method to all backends
- [x] `crates/data_components/src/debezium/tests/mod.rs`: 8 new tests for schema evolution detection and nullable field handling
- [x] `crates/runtime/src/datafusion/refresh_sql.rs`: 2 new tests reproducing the exact bug scenario

## Test plan
- [x] `field_equality_same` — identical fields compare equal
- [x] `field_equality_different_name` — fields with different names compare unequal
- [x] `field_equality_different_type` — fields with different types compare unequal
- [x] `detect_schema_evolution_added_column` — added column detected as schema change
- [x] `detect_schema_evolution_removed_column` — removed column detected as schema change
- [x] `no_false_schema_evolution_detection` — identical schemas produce no false positive
- [x] `schema_evolution_arrow_conversion_with_new_column` — full Debezium event schema parsing with evolved schema
- [x] `append_value_handles_missing_nullable_field` — missing nullable field gets null value
- [x] `append_value_errors_on_missing_required_field` — missing required field still errors
- [x] `test_schema_evolution_refresh_sql_validation` — reproduces bug #9782: stale schema causes validation error, evolved schema passes
- [x] `test_schema_evolution_wildcard_refresh_sql` — wildcard SELECT works with both stale and evolved schemas


peasee:
duplicate of #10053

PR #10082, opened 2026-04-01
fix: Use GreedyMemoryPool, add spidapter query memory limit arg

peasee:
## 📝 Summary

* Use `GreedyMemoryPool` instead of `FairSpillPool`, as it allows reservations to complete for high-concurrency queries with lots of partitions (= reservations)
* Updates spidapter to allow setting the `runtime.query.memory_limit` via an input arg or environment variable.

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task.rs
@@ -1345,12 +1345,11 @@ impl RefreshTask {
             }
             match accelerated_field.data_type() {
                 DataType::Int8 | DataType::Int16 | DataType::Int32 | DataType::Int64 => {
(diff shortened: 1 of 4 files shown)


crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -1345,12 +1345,11 @@ impl RefreshTask {
             }
             match accelerated_field.data_type() {
                 DataType::Int8 | DataType::Int16 | DataType::Int32 | DataType::Int64 => {
-                    let arr = arrow::compute::cast(col_array, &DataType::Int64)
```
peasee, line 1348: changes from `cargo fmt`

phillipleblanc (approved):

Issue #10081, opened 2026-04-01
Enhancement: Broadcast table for Distributed Cayenne Catalog.


Jeadie:
## Goal-State/What/Result

Enable tables created on Cayenne catalogs to be broadcast to all executors (instead of partitioned across).

## Why/Purpose

When tables are partitioned across executors, any SQL join is forced to occur on the scheduler (no single executor has all the data). This increases work for the scheduler and data transfer from the executors resulting in slower queries with more load on the database. 

## By When

**Issue/Spec written and reviewed:** 1st April
**Done-Done:** 3rd April

## Done-Done

- [ ] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Telemetry / Metrics / Task History
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials

#### User interface - SQL
For example, a join across two tables: `nation` and `projection`, with two executors.
```sql
Inner Join: supplier.s_nationkey = nation.n_nationkey
  SubqueryAlias: supplier
    Union
      TableScan: supplier projection=[s_suppkey, s_name, s_address, s_nationkey] -- Executor 1
      TableScan: supplier projection=[s_suppkey, s_name, s_address, s_nationkey] -- Executor 2
  SubqueryAlias: nation
    Union
      TableScan: nation projection=[n_nationkey, n_name], partial_filters=[nation.n_name = Utf8View("CANADA")] -- Executor 1
      TableScan: nation projection=[n_nationkey, n_name], partial_filters=[nation.n_name = Utf8View("CANADA")] -- Executor 2
```

If however, `nation` is available on all executors, joins can occur on an executor.
```sql
Inner Join: supplier.s_nationkey = nation.n_nationkey
  SubqueryAlias: supplier
    Union
      TableScan: supplier projection=[s_suppkey, s_name, s_address, s_nationkey] -- Executor 1
      TableScan: supplier projection=[s_suppkey, s_name, s_address, s_nationkey] -- Executor 2
  TableScan: nation projection=[n_nationkey, n_name], partial_filters=[nation.n_name = Utf8View("CANADA")] -- Either Executor 1 or 2
      
[...]

PR #10090, opened 2026-04-01
Jeadie/26 04 01/broadcast

Jeadie:
## 📝 Summary

Support replicated keyword in DDL
```sql
CREATE TABLE IF NOT EXISTS cayenne_catalog.any_schema.my_table (
    "id" BIGINT NOT NULL,
    "name" TEXT,
    "score" DOUBLE PRECISION,
    PRIMARY KEY ("id")
) REPLICATED ;
```

Apart of #10081 (unreviewed).


Jeadie:
Still requires query side optimisations to optimally use broadcast table (not just select one at random).

Jeadie:
POC for UX, but not being prioritised. Would require rewrite given interim changes

PR #10108, opened 2026-04-02
Add mode: file_update acceleration mode

lukekim:
## Summary

Adds a new acceleration mode `mode: file_update` that:

1. **Loads an existing acceleration file** if it exists (like `mode: file`)
2. **Performs schema inference** from the source on refresh
3. **Checks schema compatibility** — if the change is additive (new columns only), keeps the existing acceleration as-is
4. **Recreates on breaking changes** — if columns were removed, renamed, or had type changes, snapshots the current file (if snapshots are enabled) and recreates the acceleration from scratch

Additionally, `mode: file_create` now snapshots before deleting existing files when snapshots are enabled.

### Core Implementation
- `FileUpdate` variant added to `Mode` enum in both `spicepod` and `runtime` crates
- `is_schema_additively_compatible()` function checks whether existing schema is a subset of the new schema
- `drop_table()` method added to `DataAccelerator` trait with implementations for DuckDB, SQLite, Turso, and Cayenne
- `snapshot_before_recreate()` utility preserves acceleration data before recreation
- Schema detection and recreation logic wired into `create_accelerated_table()`

### Tests Added
- 9 unit tests for `is_schema_additively_compatible` (identical, additive, breaking, edge cases)
- 3 DDL parse tests for `file`, `file_create`, `file_update` mode strings
- 1 DuckDB `drop_table` SQL test
- 5 spicepod YAML deserialization tests (all modes + Display round-trip)
- 1 integration test for `file_update` schema evolution (additive + breaking changes)


lukekim:
Regarding `https_only(true)` in `client.rs`: This is intentional — the cloud client only connects to the Spice Cloud API which is always HTTPS. Keeping `https_only(true)` as a safety measure.

Regarding `Schema::empty()` in `snapshot_before_recreate` calls (duckdb.rs, sqlite.rs, turso.rs, cayenne/mod.rs): This was addressed in `crates/runtime/src/dataaccelerator/snapshots.rs` — the function now retrieves the currently stored schema via `current_stored_schema()` and returns early if both the passed schema is empty and no schema is stored, preventing empty schema persistence.

krinart (approved):

PR #10113, opened 2026-04-03
feat: Add mTLS client auth support to spice sql REPL

phillipleblanc:
## Changes

Add `--client-tls-certificate-file` and `--client-tls-key-file` flags to `spice sql` to enable mutual TLS authentication when connecting to cluster nodes that enforce client certificate verification.

This allows the REPL to connect directly to executor nodes in distributed cluster mode protected with mTLS, by presenting a client certificate and key signed by the cluster CA.

### What changed

- **`crates/repl/src/lib.rs`**: Add `tls_client_certificate_file` and `tls_client_key_file` fields to `ReplConfig`. Validate both flags are provided together. Construct `ClientTlsConfig` with `Identity::from_pem()` when client identity is provided. Auto-enable TLS with system root CAs when client identity is set without an explicit `--tls-root-certificate-file`.
- **`bin/spice/src/commands/sql.rs`**: Add `--client-tls-certificate-file` and `--client-tls-key-file` CLI args and plumb them through to `ReplConfig`.

### Why

In distributed cluster mode with mTLS, executor Flight servers use `ServerTlsConfig::new().identity(...).client_ca_root(...)` — requiring connecting clients to present a valid certificate signed by the cluster CA. Previously, `spice sql` only supported one-way TLS (verifying the server), so it could not complete the mTLS handshake with executor nodes.

### Usage

```bash
# Generate client cert using existing PKI tooling
spice cluster tls init
spice cluster tls add repl-client --host executor.example.com

# Connect to an executor node with mTLS
spice sql \
  --endpoint grpc+tls://executor.example.com:50052 \
  --tls-root-certificate-file ~/.spice/pki/ca.crt \
  --client-tls-certificate-file ~/.spice/pki/repl-client.crt \
  --client-tls-key-file ~/.spice/pki/repl-client.key
```

### Testing

- `cargo check -p repl -p spice` — compiles cleanly
- `cargo test -p repl` — 45/45 tests pass
- `cargo test -p spice --lib` — 79/79 tests pass

### DIFF
--- a/bin/spice/src/commands/sql.rs
+++ b/bin/spice/src/commands/sql.rs
@@ -53,6 +53,18 @@ pub struct SqlArgs {
     #[arg(long)]
     tls_root_certificate_file: Option<String>,
 
+    /// The path to the client certificate file for mTLS authentication.
+    /// Required when connecting directly to a cluster node that enforces mutual TLS.
+    /// Must be used together with --client-tls-key-file.
+    #[arg(long, requires = "client_tls_key_file")]
+    client_tls_certificate_file: Option<String>,
+
+    /// The path to the client private key file for mTLS authentication.
+    /// Required when connecting directly to a cluster node that enforces mutual TLS.
+    /// Must be used together with --client-tls-certificate-file.
+    #[arg(long, requires = "client_tls_certificate_file")]
+    client_tls_key_file: Option<String>,
+
     /// Custom HTTP headers in format 'Key:Value' (can be specified multiple times)
     #[arg(long = "headers", value_name = "KEY:VALUE")]
     custom_headers: Vec<String>,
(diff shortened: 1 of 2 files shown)


bin/spice/src/commands/sql.rs
```diff
@@ -53,6 +53,18 @@ pub struct SqlArgs {
...
+    /// The path to the client certificate file for mTLS authentication.
+    /// Required when connecting directly to a cluster node that enforces mutual TLS.
+    /// Must be used together with --client-tls-key-file.
+    #[arg(long)]
+    client_tls_certificate_file: Option<String>,
+
+    /// The path to the client private key file for mTLS authentication.
+    /// Required when connecting directly to a cluster node that enforces mutual TLS.
+    /// Must be used together with --client-tls-certificate-file.
+    #[arg(long)]
```
Jeadie:
```suggestion
    #[arg(long, requires = "client_tls_certificate_file")]
```

Jeadie: And vice versa

crates/repl/src/lib.rs
```diff
@@ -272,25 +288,57 @@ pub async fn run(repl_config: ReplConfig) -> Result<(), Box<dyn std::error::Erro
         new_agent.push_str(&user_agent);
         user_agent = new_agent;
     }
+    // Validate mTLS client certificate/key pairing
+    let client_identity = match (
```
Jeadie: not needed if we use clap macros

Jeadie (approved):

PR #10117, opened 2026-04-03
Route 'bucket(N, col)' to shared executors

Jeadie:
## 📝 Summary
Implements deterministic executor assignment for `bucket(N, col)` partition by expressions for cayenne catalog write-through. When a new partition value arrives with a single key matching `bucket(N, col)` and integer scalar k, the executor is chosen as `k % len(executors)` (instead of the least-loaded executor; which is still the algorithm otherwise).

This guarantees that the same bucket index is co-located on the same executor across different tables, enabling local joins without cross-executor data shuffling. For example:
 - `orders` partitioned by bucket(7, customer_id)
 - `customers` partitioned by bucket(7, customer_id)
 
 Then this JOIN operation can occur locally on each executor (instead of the scheduler). Example query query 
```sql
select foo, bar, customer_id 
FROM orders o
JOIN customers c on c.customer_id = o..customer_id
```


Jeadie:
Old and not prioritised. Would require rewrite given changes in spiced

PR #10130, opened 2026-04-03
fix: Run partition discovery before forwarding refresh to executors (#10075)

### LINKED ISSUE #10075: Bug: Distributed Accelerations drop data from new partition values on dataset refresh
## Describe the bug

In distributed accelerations, the partition management task assigns partitions to executors for the values in the source data.

If `spice refresh` is run on new source data which contains a new partition value, before the partition management task has run, executors will refresh without this partition being assigned. This will result in the data from the new partition being dropped, and not accelerated/loaded.

## To Reproduce

1. Configure a distributed acceleration
2. Add some new rows to source which would create a new partition value
3. Trigger a `spice refresh` before the partition management task runs
4. See executors refresh their existing partition values, but not the new partition value
5. Observe dropped/missing data

## Expected behavior

New partitions should be assigned on refresh to ensure that no data is ever dropped on refresh.

### Spicepod

### Output of `explain query`

E.g. `explain select 1`

```bash
```

### spice, spiced, OS info

- OS info: `uname -a`

- [ ] Yes
- [ ] No

claudespice:
## Summary

- Fixes #10075: Distributed accelerations drop data from new partition values on dataset refresh
- Before forwarding a refresh command to executors, the scheduler now runs on-demand partition discovery to ensure new partition values are discovered, assigned, and executors notified
- Graceful degradation: if partition discovery fails, the refresh still proceeds

## Changes

- **`crates/runtime/src/datafusion/mod.rs`**: Add `discover_and_assign_partitions_for_refresh()` method that discovers new partitions from the federated source, assigns them to executors via round-robin, and notifies executors before the refresh command is forwarded. Called from `forward_refresh_to_executors()`.
- **`crates/runtime/src/datafusion/builder.rs`**: Add `with_partition_manager()` and `with_app()` builder methods to pass partition manager and app reference to DataFusion.
- **`crates/runtime/src/builder.rs`**: Wire `partition_manager` and `app` from the `DistributedNode::Scheduler` through to the DataFusion builder. Create shared app reference early so both DataFusion and Runtime can use it.
- **`crates/runtime/src/cluster/partition/mod.rs`**: Make `discovery` and `metadata` modules `pub(crate)` for internal access from the DataFusion module.
- **`crates/runtime/src/cluster/partition/manager.rs`**: Add unit tests for `add_and_assign_partitions` covering new partition addition, idempotency, and empty input.

## Root Cause

When `spice refresh` is triggered on a scheduler, the `forward_refresh_to_executors()` method immediately sends a `RefreshDatasetCommand` to all connected executors without first ensuring new partitions have been discovered and assigned. The partition management task runs on a 30-second interval, so if refresh is triggered between intervals, new partitions are missed and their data is dropped by executors.

## Test plan

- [x] Unit tests for `add_and_assign_partitions` (4 tests: new partitions, idempotency, alongside existing, empty input)
- [x] `cargo check` passes
- [x] `cargo clippy` passes (no new warnings in modified files)
- [ ] Existing cluster integration tests pass in CI


lukekim (dismissed):

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2060,6 +2085,253 @@ impl DataFusion {
...
+            let mut partitions_bytes = Vec::new();
+            for p in partitions {
+                match partition_value_to_bytes(p.clone(), dataset_name, &df_arc).await {
+                    Ok(bytes) => partitions_bytes.push(bytes.to_vec()),
+                    Err(e) => {
+                        tracing::warn!(
+                            table = %table_name,
+                            error = %e,
+                            "Failed to convert partition value to bytes, skipping notification for this partition"
+                        );
```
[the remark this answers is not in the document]
claudespice: This is intentionally best-effort. The metadata has already been updated via add_and_assign_partitions, and the periodic partition management cycle will re-assign/notify for any partitions missed here. Aborting the entire refresh on a single byte-conversion failure would be worse — it would block the refresh for all partitions on the table when the data for already-notified partitions is ready to be refreshed.

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2060,6 +2085,253 @@ impl DataFusion {
...
+            };
+
+            if let Err(e) = executor_registry.send_command(executor_id, command).await {
+                tracing::warn!(
+                    executor_id = %executor_id,
+                    table = %table_name,
+                    error = %e,
+                    "Failed to notify executor of new partitions before refresh"
+                );
+            }
```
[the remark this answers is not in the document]
claudespice: By design — the pre-refresh notification is best-effort. Failing to notify one executor should not prevent the refresh from reaching executors that were successfully notified. The periodic partition management cycle handles catch-up for missed notifications. Blocking the refresh here would degrade availability to improve consistency of a single edge case.

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2060,6 +2085,253 @@ impl DataFusion {
...
+            current_metadata
+                .partitions
+                .iter()
+                .map(|p| {
+                    let mut v: Vec<_> = p.partition_value.clone().into_iter().collect();
+                    v.sort();
+                    v
+                })
+                .collect();
+
```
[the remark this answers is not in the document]
claudespice: When get_table_metadata returns None, the table's partition metadata hasn't been initialized yet — the periodic partition management cycle handles initial seeding. Initializing metadata inline during refresh would risk racing with the periodic cycle and add complexity to a path that's meant to be a lightweight pre-check. The periodic cycle is the authoritative metadata initialization path.

crates/runtime/src/datafusion/mod.rs
```diff
@@ -1994,12 +2000,31 @@ impl DataFusion {
...
+        // runs on a 30-second interval) might not have discovered new partitions yet,
+        // causing executors to drop data from unassigned partitions. (fixes #10075)
+        if let Some(partition_manager) = &self.partition_manager {
+            self.discover_and_assign_partitions_for_refresh(
+                partition_manager,
+                executor_registry,
+                dataset_name,
+            )
+            .await;
+        }
```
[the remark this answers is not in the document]
claudespice: Agreed that this would be valuable. Testing the UpdatePartitions-before-RefreshDataset ordering requires a multi-executor integration test with mock flight services. The unit tests for add_and_assign_partitions cover the core logic; a full ordering test can be added as a follow-up.

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2060,6 +2085,253 @@ impl DataFusion {
...
+            })
+            .collect();
+
+        if new_partitions.is_empty() {
+            tracing::debug!(
+                table = %table_name,
+                "No new partitions discovered before refresh"
+            );
+            return;
+        }
```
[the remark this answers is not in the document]
claudespice: This pre-refresh check is specifically scoped to discover *brand new* partitions from the source that don't exist in metadata yet — the fix for #10075. Existing-but-unassigned partitions are a different edge case handled by the periodic partition management cycle (which does full assignment reconciliation). Expanding scope here would complicate the pre-refresh path and risk interfering with the periodic cycle's assignment logic.

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2161,6 +2186,253 @@ impl DataFusion {
...
+
+        tracing::info!(
```
peasee:
```suggestion
        tracing::debug!(
```

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2161,6 +2186,253 @@ impl DataFusion {
...
+                    table = %table_name,
+                    "No partition metadata found, skipping pre-refresh partition discovery"
+                );
+                return;
+            }
+            Err(e) => {
+                tracing::warn!(
+                    table = %table_name,
+                    error = %e,
+                    "Failed to get partition metadata before refresh, proceeding anyway"
```
peasee: this should be a hard error. if we refresh without having all partitions assigned, we will be in a bad data state and representing incorrect data on query by consumers

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2161,6 +2186,253 @@ impl DataFusion {
...
+
+        // Discover current partition values from the source.
+        let source_partitions =
+            match table_partition_values(dataset_name, &partition_by, &df_arc).await {
+                Ok(values) => values,
+                Err(e) => {
+                    tracing::warn!(
+                        table = %table_name,
+                        error = %e,
+                        "Failed to discover partition values before refresh, proceeding anyway"
```
peasee: this should be a hard error

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2161,6 +2186,253 @@ impl DataFusion {
...
+                "Failed to refresh partition manager cache before refresh, proceeding anyway"
+            );
+            return;
+        }
+
+        // Get the self Arc for table_partition_values (it needs &Arc<DataFusion>).
+        let Some(df_arc) = self.datafusion_ref.get().and_then(std::sync::Weak::upgrade) else {
+            tracing::warn!(
+                table = %table_name,
+                "DataFusion self-reference not available, skipping pre-refresh partition discovery"
```
peasee: this should be a hard error

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2161,6 +2186,253 @@ impl DataFusion {
...
+        let partition_by = acceleration.partition_by.clone();
+        // Drop the read guard before performing async operations that might be slow.
+        drop(app_opt);
+
+        // Refresh partition manager cache from object store.
+        if let Err(e) = partition_manager.refresh().await {
+            tracing::warn!(
+                table = %table_name,
+                error = %e,
+                "Failed to refresh partition manager cache before refresh, proceeding anyway"
```
peasee: this should be a hard error

crates/runtime/src/datafusion/mod.rs
```diff
@@ -570,6 +572,10 @@ pub struct DataFusion {
     pub executor_stream_registry: RwLock<Option<ExecutorControlStreamRegistry>>,
     /// Executor registry for distributed write forwarding (scheduler mode only).
     pub executor_registry: Option<Arc<ExecutorRegistry>>,
+    /// Partition manager for discovering/assigning partitions before refresh (scheduler mode only).
+    pub(crate) partition_manager: Option<Arc<PartitionManager>>,
+    /// App reference for partition discovery during refresh (scheduler mode only).
+    pub(crate) app: Option<Arc<TokioRwLock<Option<Arc<app::App>>>>>,
```
peasee, line 583: this is never written too, it doesn't need to be in a `RwLock`

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2161,6 +2186,253 @@ impl DataFusion {
...
+            .collect();
+
+        if let Err(e) = partition_manager
+            .add_and_assign_partitions(dataset_name, &assignments)
+            .await
+        {
+            tracing::warn!(
+                table = %table_name,
+                error = %e,
+                "Failed to add and assign new partitions before refresh, proceeding anyway"
```
peasee: this should be a hard error

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2161,6 +2186,253 @@ impl DataFusion {
...
+        for (executor_id, partitions) in by_executor {
+            let mut partitions_bytes = Vec::new();
+            for p in partitions {
+                match partition_value_to_bytes(p.clone(), dataset_name, &df_arc).await {
+                    Ok(bytes) => partitions_bytes.push(bytes.to_vec()),
+                    Err(e) => {
+                        tracing::warn!(
+                            table = %table_name,
+                            error = %e,
+                            "Failed to convert partition value to bytes, skipping notification for this partition"
```
peasee: this should be a hard error

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2161,6 +2186,253 @@ impl DataFusion {
...
+                    },
+                )),
+            };
+
+            if let Err(e) = executor_registry.send_command(executor_id, command).await {
+                tracing::warn!(
+                    executor_id = %executor_id,
+                    table = %table_name,
+                    error = %e,
+                    "Failed to notify executor of new partitions before refresh"
```
peasee: this should be a hard error

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2161,6 +2186,253 @@ impl DataFusion {
...
+                );
+            }
+        }
+
+        tracing::info!(
```
peasee:
```suggestion
        tracing::debug!(
```

peasee (changes requested):

crates/runtime/src/datafusion/mod.rs
claudespice: Fixed in d2579fb16. All warn+return patterns are now hard errors using a new `PreRefreshPartitionDiscoveryFailed` error variant that prevents the refresh from proceeding with incomplete partition state.

crates/runtime/src/datafusion/mod.rs
claudespice: Fixed in d2579fb16 — now returns `PreRefreshPartitionDiscoveryFailed` error.

crates/runtime/src/datafusion/mod.rs
claudespice: Fixed in d2579fb16 — now returns `PreRefreshPartitionDiscoveryFailed` error.

crates/runtime/src/datafusion/mod.rs
claudespice: Fixed in d2579fb16 — now returns `PreRefreshPartitionDiscoveryFailed` error.

crates/runtime/src/datafusion/mod.rs
claudespice: Fixed in d2579fb16 — now returns `PreRefreshPartitionDiscoveryFailed` error.

crates/runtime/src/datafusion/mod.rs
claudespice: Fixed in d2579fb16 — now returns `PreRefreshPartitionDiscoveryFailed` error.

crates/runtime/src/datafusion/mod.rs
claudespice: Fixed in d2579fb16 — now returns `PreRefreshPartitionDiscoveryFailed` error.

crates/runtime/src/datafusion/mod.rs
claudespice: Fixed in d2579fb16. Moved the RwLock resolution into `forward_refresh_to_executors` so `discover_and_assign_partitions_for_refresh` now receives a resolved `Arc<App>` directly. The field type stays as-is since it's a shared reference from the runtime that may be written by other components.

crates/runtime/src/datafusion/mod.rs
claudespice: Fixed in d2579fb16 — changed to `tracing::debug!`.

crates/runtime/src/datafusion/mod.rs
claudespice: Fixed in d2579fb16 — changed to `tracing::debug!`.

peasee (dismissed):

krinart:
Closed in favor of https://github.com/spiceai/spiceai/pull/10216

PR #10131, opened 2026-04-03
fix: Don't report distributed accelerations as immediately ready

### LINKED ISSUE #10076: Bug: Distributed Accelerations report immediately as "ready"
## Describe the bug

When creating a distributed acceleration dataset, it is immediately reported by the scheduler as "ready" even when it is not yet ready.

## Goal

The scheduler's `/v1/ready` endpoint should only report ready once every distributed dataset is actually queryable — meaning every partition has at least one executor that has finished accelerating it.

### Current architecture

Each scheduler↔executor pair maintains a persistent bidirectional gRPC control stream. The stream carries:

**Executor → Scheduler:** `ExecutorHeartbeat`, `MetricsResponse`, `ExecutorShutdown`
**Scheduler → Executor:** `MetricsRequest`, `PollNow`, `UpdatePartitions`, `CancelTasks`, `RefreshDataset`

On the executor side, `RuntimeStatus` already tracks per-dataset `ComponentStatus` (Initializing → Refreshing → Ready). This status is used locally but is **not communicated to the scheduler**.

On the scheduler side, `/v1/ready` uses `RuntimeStatus` with `OnLoad` mode: all registered components must reach `Ready` at least once. Currently, the scheduler marks distributed datasets as `Ready` immediately during table creation — before any executor work has begun.

### What "queryable" means

The scheduler's query planner (`PartitionedTableScanRewrite`) routes queries to executors via FlightSQL. For a dataset to be queryable, every partition must have at least one live executor that has finished accelerating it. This is the condition we need to gate `/v1/ready` on.

## Option A: Push-based — dedicated `DatasetStatusChange` message

Executors send a new `DatasetStatusChange` message over the existing control stream whenever a dataset's status transitions (e.g., Initializing → Ready).

```protobuf
message DatasetStatusChange {
    string dataset = 1;
    int32 status = 2;  // ComponentStatus discriminant
}
```

**Pros:**
- Real-time
- No extra messages: only sends a message when something actually changes

**Cons:**
- New message type in the proto (`DatasetStatusChange` added to `ExecutorControlMessage.oneof`)
- If the message is lost (e.g., transient stream reconnect), the scheduler won't know until the next reconciliation mechanism kicks in
- Executor needs new infrastructure to detect status changes and send them (a `Notify` on `RuntimeStatus` + a sender task per control stream)

## Option B: Push-based — extend `ExecutorHeartbeat` with dataset statuses

Add a `map<string, int32> dataset_statuses` field to the existing `ExecutorHeartbeat` message. Every 10s heartbeat carries a full snapshot of all dataset statuses.

```protobuf
message ExecutorHeartbeat {
    int64 timestamp_ms = 1;
    map<string, int32> dataset_statuses = 2;
}
```

**Pros:**
- No new message type
- Self-healing: even if a status change is missed, the next heartbeat (within 10s) carries the full snapshot
- Simple implementation

**Cons:**
- Up to 10s latency between executor finishing acceleration and scheduler knowing
- Overloads the heartbeat's purpose

## Option C: Pull-based — scheduler requests dataset status from executors

Similar to the existing `MetricsRequest`/`MetricsResponse` pattern. The scheduler sends a `DatasetStatusRequest` to executors and they respond with their current dataset statuses.

```protobuf
// Scheduler → Executor
message DatasetStatusRequest {
    string request_id = 1;
}
// Executor → Scheduler
message DatasetStatusResponse {
    string request_id = 1;
    map<string, int32> dataset_statuses = 2;
}
```

**Pros:**
- Consistent with existing metrics collection pattern (pull-based)
- No unnecessary messages
- No executor-side change detection infrastructure needed

**Cons:**
- Adds a polling interval (latency) or requires a trigger
- Two new message types
- More complex scheduler-side logic: manage pending requests, timeouts, correlate responses by request_id

## Option D: Combine A + B (recommended)

Use `DatasetStatusChange` for real-time notifications **and** include dataset statuses in heartbeats for periodic reconciliation.

**Pros:**
- Best of both worlds: real-time updates + self-healing reconciliation
- If a `DatasetStatusChange` is lost during a stream reconnect, the next heartbeat (within 10s) reconciles the state

**Cons:**
- Adds new message type
- Extends `ExecutorHeartbet` message (although arguably current datasets' statuses is the part of heartbet)
- Slightly more network traffic than A alone (heartbeats carry snapshots even when nothing changed)

## Recommendation

**Option D (A + B combined)** provides the best reliability with minimal latency:

1. `DatasetStatusChange` delivers real-time updates
2. `ExecutorHeartbeat.dataset_statuses` provides periodic reconciliation
3. The scheduler promotes a dataset to `Ready` only when every partition has at least one assigned executor reporting `Ready`

## Scheduler-side readiness logic

```
For each dataset:
  For each partition:
    At least one assigned executor must report the dataset as Ready
  If all partitions are queryable → mark dataset as Ready

/v1/ready returns 200 when all datasets have been Ready at least once
```

claudespice:
## Summary
- Fixes scheduler-side datasets being marked as `Ready` immediately before executors have finished accelerating the data
- The partition management task now promotes datasets to `Ready` once all partitions have been assigned to executors
- Adds `update_dataset_statuses_from_partitions()` function with 6 unit tests

## Changes
- [x] `accelerated_table/mod.rs`: Mark scheduler datasets as `Initializing` instead of `Ready`
- [x] `scheduler_task.rs`: After each successful management cycle, check partition assignment completeness and update dataset statuses
- [x] Add `update_dataset_statuses_from_partitions()` free function
- [x] Add 6 unit tests covering: all assigned, partial, fully unassigned, empty, multiple tables, no tables

## Test plan
- [x] Unit tests for `update_dataset_statuses_from_partitions` covering all partition assignment states
- [ ] Verify distributed acceleration datasets block `/v1/ready` until partitions are assigned
- [ ] Verify datasets become Ready after partition management cycle completes with all partitions assigned

Closes #10076

### DIFF
--- a/crates/runtime/src/accelerated_table/mod.rs
+++ b/crates/runtime/src/accelerated_table/mod.rs
@@ -762,12 +762,12 @@ impl Builder {
         refresher.with_s3_express_acceleration(self.is_s3_express_acceleration);
 
         let refresh_handle = if matches!(self.cluster_role, Some(ClusterRole::Scheduler)) {
-            // Accelerated tables aren't accelerated on the scheduler. Don't mark
-            // as Ready yet — the partition management task will promote the dataset
-            // to Ready once all of its partitions have been assigned to executors.
-            // Register as Initializing so /v1/ready waits for partition assignment.
-            self.runtime_status
-                .update_dataset(&self.dataset_name, status::ComponentStatus::Initializing);
+            // Accelerated tables aren't accelerated on the scheduler.
+            // Don't register this dataset as a readiness component — the
+            // `partition_metadata` component already gates /v1/ready until
+            // metadata has been seeded, and the partition management task
+            // will promote datasets to Ready once executors have been
+            // assigned their partitions.
             None
         } else {
             refresher.start(acceleration_refresh_mode).await?
--- a/crates/runtime/src/accelerated_table/mod.rs
+++ b/crates/runtime/src/accelerated_table/mod.rs
@@ -762,9 +762,12 @@ impl Builder {
         refresher.with_s3_express_acceleration(self.is_s3_express_acceleration);
 
         let refresh_handle = if matches!(self.cluster_role, Some(ClusterRole::Scheduler)) {
-            // Accelerated tables aren't accelerated on scheduler. Immediately ready.
+            // Accelerated tables aren't accelerated on the scheduler. Don't mark
+            // as Ready yet — the partition management task will promote the dataset
+            // to Ready once all of its partitions have been assigned to executors.
+            // Register as Initializing so /v1/ready waits for partition assignment.
             self.runtime_status
-                .update_dataset(&self.dataset_name, status::ComponentStatus::Ready);
+                .update_dataset(&self.dataset_name, status::ComponentStatus::Initializing);
             None
         } else {
             refresher.start(acceleration_refresh_mode).await?
--- a/crates/runtime/src/cluster/partition/scheduler_task.rs
+++ b/crates/runtime/src/cluster/partition/scheduler_task.rs
@@ -294,6 +294,7 @@ impl PartitionManagementTask {
                     match self.run_management_cycle().await {
                         Ok(()) => {
                             self.status.update_component_status("partition_metadata", ComponentStatus::Ready);
+                            self.update_dataset_statuses().await;
                         }
                         Err(e) => {
                             tracing::warn!(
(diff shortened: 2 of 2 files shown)


crates/runtime/src/cluster/partition/scheduler_task.rs
```diff
@@ -1198,3 +1234,230 @@ fn now_ms() -> u128 {
...
+        };
+
+        // A table with zero partitions has nothing to assign, so it's ready.
+        // Otherwise, every partition must be assigned to at least one executor.
+        let all_assigned = metadata
+            .partitions
+            .iter()
+            .all(PartitionMetadata::is_assigned);
+
+        if all_assigned {
```
peasee, line 1270: just because all the calculated partitions are assigned doesn't mean they're ingested and ready.

you could also race here to update the status to ready if this is checked before any partitions are even calculated?

crates/runtime/src/accelerated_table/mod.rs
```diff
@@ -762,9 +762,12 @@ impl Builder {
         refresher.with_s3_express_acceleration(self.is_s3_express_acceleration);
 
         let refresh_handle = if matches!(self.cluster_role, Some(ClusterRole::Scheduler)) {
-            // Accelerated tables aren't accelerated on scheduler. Immediately ready.
-            self.runtime_status
-                .update_dataset(&self.dataset_name, status::ComponentStatus::Ready);
+            // Accelerated tables aren't accelerated on the scheduler.
```
peasee, line 765: we should continue listing the table as not ready when it is initially created, just like every other dataset.

peasee (changes requested):

PR #10156, opened 2026-04-04
feat: Distributed runtime.task_history support

### LINKED ISSUE #7611: Distributed `runtime.task_history` support
In single process mode, everything neatly appends to a singleton memory table. In distributed mode, we want to be able to collect task history from the entire cluster

- `runtime.task_history` in distributed mode should have a top level `hostname` column
- A query to `runtime.task_history` in distributed mode should fan out queries to each executor's task history table for results (sorting by timestamp)

lukekim:
## Summary

- Add a `node_id` column to `runtime.task_history` in distributed mode to identify the originating cluster node
- Fan out task history queries to connected executors via FlightSQL in addition to peer schedulers
- Sort combined results by `start_time` (ascending) across all cluster nodes
- Use DataFusion's SQL unparser for generating valid SQL in peer/executor queries

Closes #7611

## Test plan

- [x] `cargo check -p runtime --lib` and `cargo check -p spiced` pass
- [x] All 6 `task_history::federated` unit tests pass (projection, SQL generation, sorting)
- [x] `cargo clippy` and `cargo fmt` pass locally
- [x] Manual test with multi-node cluster to verify executor fan-out and node_id column


Jeadie:
I believe we already have some implementation for distributed runtime.task_history. (see `ClusterService::GetTaskHistory` in proto.)

lukekim:
Good catch @Jeadie — yes, `ClusterService::GetTaskHistory` was already in the proto and implemented in `cluster/service.rs`. This PR builds on that existing RPC rather than duplicating it:

- The `FederatedTaskHistoryTable` (new in this PR) is the **orchestrator** that fans out queries across the cluster
- It uses the existing `GetTaskHistory` RPC as the **transport** for scheduler-to-scheduler queries (see `query_peer()` which calls `ClusterServiceClient::get_task_history()`)
- The server-side handler in `service.rs` rewrites `runtime.task_history` → `runtime.local_task_history` to avoid infinite recursion
- The PR additionally adds executor fan-out via FlightSQL (executors dont run the ClusterService gRPC), a `node_id` column, and global sorting by `start_time`

So existing infra is preserved and reused as the peer query path.

Jeadie:
> Good catch @Jeadie — yes, `ClusterService::GetTaskHistory` was already in the proto and implemented in `cluster/service.rs`. This PR builds on that existing RPC rather than duplicating it:
> 
> * The `FederatedTaskHistoryTable` (new in this PR) is the **orchestrator** that fans out queries across the cluster
> * It uses the existing `GetTaskHistory` RPC as the **transport** for scheduler-to-scheduler queries (see `query_peer()` which calls `ClusterServiceClient::get_task_history()`)
> * The server-side handler in `service.rs` rewrites `runtime.task_history` → `runtime.local_task_history` to avoid infinite recursion
> * The PR additionally adds executor fan-out via FlightSQL (executors dont run the ClusterService gRPC), a `node_id` column, and global sorting by `start_time`
> 
> So existing infra is preserved and reused as the peer query path.

I thought I saw something explicitly incompatible when i perused yesterday. Will check again when reviewing now.

Jeadie (approved):
To some extent `FederatedTaskHistoryTable` feels like it should be an `AnalyzerRule` that simply `UNION ALL` across schedulers and executors.

ewgenius:
Verified distributed task history on 4 executors cluster:

CleanShot 2026-04-07 at 14 11 44@2x

CleanShot 2026-04-07 at 14 10 56@2x

ewgenius (approved):

PR #10216, opened 2026-04-07
fix: Run partition discovery before forwarding refresh to executors (#10075)

### LINKED ISSUE #10075: Bug: Distributed Accelerations drop data from new partition values on dataset refresh

krinart:
Original https://github.com/spiceai/spiceai/pull/10130

## Summary

- Fixes #10075: Distributed accelerations drop data from new partition values on dataset refresh
- Before forwarding a refresh command to executors, the scheduler now runs on-demand partition discovery to ensure new partition values are discovered, assigned, and executors notified
- Graceful degradation: if partition discovery fails, the refresh still proceeds

## Changes

- **`crates/runtime/src/datafusion/mod.rs`**: Add `discover_and_assign_partitions_for_refresh()` method that discovers new partitions from the federated source, assigns them to executors via round-robin, and notifies executors before the refresh command is forwarded. Called from `forward_refresh_to_executors()`.
- **`crates/runtime/src/datafusion/builder.rs`**: Add `with_partition_manager()` and `with_app()` builder methods to pass partition manager and app reference to DataFusion.
- **`crates/runtime/src/builder.rs`**: Wire `partition_manager` and `app` from the `DistributedNode::Scheduler` through to the DataFusion builder. Create shared app reference early so both DataFusion and Runtime can use it.
- **`crates/runtime/src/cluster/partition/mod.rs`**: Make `discovery` and `metadata` modules `pub(crate)` for internal access from the DataFusion module.
- **`crates/runtime/src/cluster/partition/manager.rs`**: Add unit tests for `add_and_assign_partitions` covering new partition addition, idempotency, and empty input.

## Root Cause

When `spice refresh` is triggered on a scheduler, the `forward_refresh_to_executors()` method immediately sends a `RefreshDatasetCommand` to all connected executors without first ensuring new partitions have been discovered and assigned. The partition management task runs on a 30-second interval, so if refresh is triggered between intervals, new partitions are missed and their data is dropped by executors.

## Test plan

- [x] Unit tests for `add_and_assign_partitions` (4 tests: new partitions, idempotency, alongside existing, empty input)
- [x] `cargo check` passes
- [x] `cargo clippy` passes (no new warnings in modified files)
- [ ] Existing cluster integration tests pass in CI


sgrebnov (approved):

lukekim (approved):

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2163,6 +2200,226 @@ impl DataFusion {
         Ok(None)
     }
 
+    /// Discovers new partition values from the source table and assigns them to executors
+    /// before a refresh command is forwarded. This ensures executors have assignments for
+    /// all current partition values, preventing data loss from new partitions.
+    ///
+    /// Returns an error if any step of partition discovery or assignment fails, which
+    /// prevents the refresh from proceeding with potentially incomplete partition state.
+    async fn discover_and_assign_partitions_for_refresh(
```
phillipleblanc, line 2209: This seems like it might be duplicating code that we already have? Surely we already have code that does a lot of what this function is trying to do

phillipleblanc: If not then we should refactor it so that it does.

phillipleblanc: logically I would expect we just trigger whatever existing logic we have for the partition refresh - I don't think this needs to actually do much else?

krinart:
Closed in favor of https://github.com/spiceai/spiceai/pull/10231

PR #10229, opened 2026-04-08
feat(databricks): DESCRIBE TABLE fallback and source-native type parsing for Lakehouse Federation

lukekim:
## Problem

Databricks Lakehouse Federation foreign tables (e.g. Neon PostgreSQL via `neon_pg_foreign`) fail schema introspection with:

```
Failed to load the dataset test_schema_bug (databricks). Failed to initialize SQL table:
Unable to get schema: The table 'neon_pg_foreign.public.test_schema_repro' has no column
metadata registered in Unity Catalog.
```

Two root causes:

1. **`information_schema.columns` returns no rows** — Foreign tables have no column metadata registered in Unity Catalog, so `data_array` is null for both `full_data_type` and `data_type` queries.

2. **Source-native type names** — When `data_type` IS populated for federation tables, it returns source-native types (`integer`, `text`, `numeric`, `timestamp without time zone`) instead of Spark SQL types (`int`, `string`, `decimal`, `timestamp`).

### 3-tier schema introspection fallback

`get_schema` now tries:
1. `information_schema.columns` with `full_data_type` (Unity Catalog tables)
2. `information_schema.columns` with `data_type` (non-UC workspaces)
3. `DESCRIBE TABLE` (Lakehouse Federation / foreign tables)

The DESCRIBE TABLE fallback activates when:
- `TableSchemaNotRegistered` — `data_array` is null (no rows in information_schema)
- `NoColumnsInDataset` — `data_array` is empty
- `ParseError` — information_schema returned unparseable source-native type names

Since DESCRIBE TABLE doesn't report nullability, all columns default to nullable.

### Source-native type parsing

Added parsing for common source-native type names from PostgreSQL/MySQL/SQL Server:
- `integer` → Int32
- `text` → Utf8
- `numeric` / `numeric(p,s)` → Decimal128
- `real` → Float32
- `double precision` → Float64
- `character varying` / `varchar` / `varchar(n)` → Utf8
- `timestamp without time zone` → Timestamp(Microsecond, None)
- `timestamp with time zone` → Timestamp(Microsecond, UTC)

## Changes

- `sql_warehouse.rs`: Refactored `get_schema` with `get_schema_from_information_schema` helper, added `create_describe_payload` (backtick-quoted identifiers), `schema_from_describe_json` parser
- `datatypes.rs`: Added source-native type handling in parser (Identifier matches + multi-word type consumption for `timestamp with/without time zone`, `double precision`, `character varying`)

## Tests

- 86 total databricks tests (all passing)
- 9 DESCRIBE TABLE parser tests (basic, metadata rows, all 4 CSV schemas, Neon PG, edge cases)
- 5 source-native type tests (parser-level + schema-level)
- Tests cover all real-world CSV schemas across all 3 fallback tiers

### DIFF
--- a/crates/data_components/src/databricks/sql_warehouse.rs
+++ b/crates/data_components/src/databricks/sql_warehouse.rs
@@ -247,9 +247,43 @@ impl SqlWarehouseApi {
 
     async fn get_schema(&self, table: &TableReference) -> Result<SchemaRef, Error> {
         let token = self.token_provider.get_token();
-        let payload = self.create_schema_payload(table, "full_data_type")?;
+        let table_name = table.to_string();
+
+        match self.get_schema_from_information_schema(&token, table).await {
+            Ok(schema) => return Ok(schema),
+            Err(Error::TableSchemaNotRegistered { .. } | Error::NoColumnsInDataset { .. }) => {
+                tracing::warn!(
+                    table = %table,
+                    "information_schema.columns has no metadata for this table, falling back to DESCRIBE TABLE. Column nullability will default to nullable."
+                );
+            }
+            Err(Error::ParseError { ref reason }) => {
+                tracing::warn!(
+                    table = %table,
+                    reason,
+                    "information_schema.columns returned unparseable type names (likely source-native types from a federated table), falling back to DESCRIBE TABLE. Column nullability will default to nullable."
+                );
+            }
+            Err(e) => return Err(e),
+        }
+
+        let payload = self.create_describe_payload(table)?;
         let response = self.execute_sql_statement(&token, &payload).await?;
         let response = self.wait_for_statement_completion(&token, response).await?;
+        schema_from_describe_json(&response, &table_name)
+    }
+
+    /// Attempts to read the schema from `information_schema.columns`,
+    /// trying `full_data_type` first and falling back to `data_type` if
+    /// the column does not exist.
+    async fn get_schema_from_information_schema(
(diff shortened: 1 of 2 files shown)


crates/data_components/src/databricks/sql_warehouse/datatypes.rs
```diff
@@ -664,4 +757,98 @@ mod tests {
             assert_eq!(result, *expected, "Failed for data_type input: {input}");
         }
     }
+
+    /// Source-native type names from Lakehouse Federation foreign tables
+    /// (e.g. `PostgreSQL`). These appear in `information_schema.columns.data_type`
+    /// when querying federated tables.
+    #[test]
+    fn test_source_native_types() {
```
sgrebnov, line 802: With `describe table` and `full_data_type` I expect we should only see Arrow type and not source native types.

sgrebnov (dismissed):

krinart (dismissed):

crates/data_components/src/databricks/sql_warehouse.rs
```diff
@@ -261,8 +295,8 @@ impl SqlWarehouseApi {
                     "Databricks information_schema does not have 'full_data_type' column, falling back to 'data_type'. Complex types (ARRAY, MAP, STRUCT) may lose inner type details."
```
sgrebnov, line 290: We should remove this handler and return error back IMO

PR #10231, opened 2026-04-08
Run partition discovery before forwarding refresh to executors

krinart:
## 📝 Summary

  - Add `PartitionService` — bundles partition manager, executor registry, config, and app reference with 3 public methods:


crates/runtime/src/cluster/partition/discovery.rs
```diff
@@ -117,6 +117,77 @@ pub async fn table_partition_values(
     Ok(partition_values)
 }
 
+/// Discovers new partition values for a single table by querying the source
+/// and diffing against existing metadata.
+///
+/// Returns the list of new partition values that exist in the source but are
+/// not yet tracked in the partition store's metadata.
+pub(crate) async fn discover_new_partitions(
```
peasee: add unit tests for this function

phillipleblanc: Don't we also need to return the partitions that no longer exist and remove them from the mapping?

crates/runtime/src/cluster/partition/scheduler_task.rs
```diff
@@ -325,868 +219,37 @@ impl PartitionManagementTask {
...
-    Ok(())
 }
 
-fn now_ms() -> u128 {
-    SystemTime::now()
-        .duration_since(UNIX_EPOCH)
-        .unwrap_or_default()
-        .as_millis()
-}
+// Partition management logic lives in `super::service::PartitionService`.
```
phillipleblanc: A comment at the end of the file is a bit weird, I would have the comment at the top of the file. And include the file path as well.

crates/runtime/src/cluster/partition/service.rs
```diff
@@ -0,0 +1,1549 @@
...
+        executor_ids,
+        tables,
+    })
+}
+
+/// For each tracked table, queries the source for current partition values and
+/// diffs against the stored metadata. Returns new partitions (in source but not
+/// in metadata) and removed partitions (in metadata but no longer in source).
+/// Does not assign or notify — the caller handles that.
+async fn discover_and_sync_partitions(
```
phillipleblanc: This should be the function that lives in crates/runtime/src/cluster/partition/discovery.rs - now we have two different functions that do identical things. If the caller doesn't care about removed partitions then it can discard that value.

crates/runtime/src/cluster/partition/service.rs
```diff
@@ -0,0 +1,1549 @@
...
+        Self {
+            partition_store,
+            executor_registry,
+            config,
+            app,
+        }
+    }
+
+    /// Discover new partitions for a single table, assign them, and notify executors.
+    pub async fn discover_and_assign_for_table(
```
phillipleblanc: This is imperative code (do this and then do this). We should be thinking about this as a reconciliation loop, similar to how the Kubernetes operator works. This reduces duplicate code and fragile code paths. Then all code paths that need to update the partition state would call the reconcile loop which handles everything, and then all callers just trigger the loop whenever its required.

phillipleblanc (changes requested):

crates/runtime/src/builder.rs
```diff
@@ -240,75 +242,81 @@ impl RuntimeBuilder {
...
+                    .and_then(|pm| PartitionManagementConfig::try_from(pm).ok())
+                    .map_or_else(AssignmentConfig::default, |pm_config| AssignmentConfig {
+                        max_assignments_per_cycle: pm_config.max_assignments_per_cycle,
+                        max_partitions_per_executor: pm_config.max_partitions_per_executor,
+                        discovery_timeout: pm_config.discovery_timeout,
+                    });
+                let partition_service = Arc::new(PartitionService::new(
+                    Arc::clone(&partition_store),
+                    Arc::clone(&executor_registry),
+                    assignment_config,
```
Jeadie, line 302: `PartitionService` doesn't need `assignment_config` if it has `shared_app`?

PR #10277, opened 2026-04-11
fix(runtime): handle HTTP JSON unions and spicepod reloads

lukekim:
## Summary
- route runtime RecordBatch JSON serialization through a shared path that preserves explicit nulls and handles `json_get(...)` union values, including nested union data in composite columns
- update HTTP and SQL result formatting to use the shared serializer instead of direct Arrow JSON writers
- refresh watched spicepod paths after root spicepod edits and accept broader relevant file change events so view changes reload without a restart

## Testing
- focused runtime regression tests are running locally:
  - `cargo test -p runtime test_write_to_json_value_ --lib`
  - `cargo test -p runtime test_is_spicepods_modification_event --lib`
  - `cargo test -p runtime test_is_root_spicepod_event_detects_main_spicepod --lib`


crates/runtime/src/datafusion/query.rs
```diff
@@ -1234,13 +1243,385 @@ fn strip_root_order_preserving_repartition(
...
+        serde_json::to_string(&write_to_json_value(data)?).boxed()
+    } else {
+        String::from_utf8(write_to_json_bytes_with_arrow(data)?).boxed()
+    }
+}
+
+pub fn write_to_json_value(
+    data: &[RecordBatch],
+) -> Result<Value, Box<dyn std::error::Error + Send + Sync>> {
+    if data.iter().any(record_batch_has_union_columns) {
```
peasee, line 1256: we've already iter'd once in `write_to_json_string`. we could avoid itering again here?

peasee (dismissed):

peasee (dismissed):

ewgenius (approved):

PR #10303, opened 2026-04-12
Optimise merge into for position-based cayenne table

Jeadie:
## 📝 Summary
For `MERGE INTO` on Position-based Cayenne tables (no primary key), avoid second physical scan on deletion, by:
    - Expose the row ids from the output of the JOIN operation (first stage of `MERGE INTO`). 
    - During deletion, use the row ids. This avoid both the physical scan, but also the expensive filter-based deletion path (evaluating an O(N) filter expression against every chunk of every file).

The previous flow built an `IN (v1, ..., vN)` or `(k1=a AND k2=b) OR ...` filter from matched rows (from JOIN 1st stage), then re-scanned the entire target table to find file-local row positions for deletion vectors. With N matched rows and M files, this
  was O(N×M) comparisons. Vortex statistics-based chunk pruning did not help when the IN-list spanned a wide value range. 

### Details
  - `CayenneDeletionSink::scan_file_for_key_matches` — scans a Vortex file without filter/projection, converts chunks to Arrow RecordBatch, hash-probes key columns against matched keys, tracks file-local positions with a manual counter.
  - `CayenneDeletionSink::delete_by_key_hash_probe` — orchestrates parallel per-file scanning with bounded concurrency, collects positions, delegates to `persist_position_based_deletions`.
  - `CayenneTableProvider::delete_matched_rows_by_key_probe` — public entry point that acquires the write lock and constructs the deletion sink.
  - `execute_merge` branches: fast path for PositionBased Cayenne tables, legacy filter path unchanged for PK-based and non-Cayenne tables.
  - `PartitionTableProvider` falls back to the legacy path for now (TODO for follow-up).


Jeadie:
closing in favour of https://github.com/spiceai/spiceai/pull/10340

PR #10328, opened 2026-04-14
S3 Vectors Top-K tweaks

lukekim:
Update S3 Vectors query top-k handling:

- Increase `S3_VECTOR_MAX_TOPK` from 100 to 10,000
- Add `S3_VECTOR_PAGE_SIZE` constant (100) as the default when no limit is specified
- Implement transparent pagination in `Client::query_vectors` for topK > 100
- Add comprehensive test coverage for limit boundaries, `CaptureNextToken` interceptor edge cases, and `to_flat_value` serialization


crates/s3_vectors/src/lib.rs
```diff
@@ -289,18 +330,125 @@ impl S3Vectors for Client {
...
-            .set_index_name(input.index_name.clone())
-            .set_index_arn(input.index_arn.clone())
-            .set_query_vector(input.query_vector.clone())
-            .set_top_k(input.top_k)
-            .set_return_distance(input.return_distance)
-            .set_return_metadata(input.return_metadata)
-            .set_filter(input.filter.clone())
-            .send()
-            .await
+        let top_k = input
```
Jeadie, line 297: With pagination now supported, we should be streaming Arrow records via `S3VectorsQueryExec::execute` (instead of collecting in the S3vector client, then emitting as a single element stream).
 
Can we make a tracking issue to describe and outline this improvement.

Jeadie (dismissed):

phillipleblanc (approved):

PR #10360, opened 2026-04-17
fix: Use ROUND for Turso decimal BETWEEN comparisons (fixes #9872)

### LINKED ISSUE #9872: TPCH Q6 incorrect result with Turso acceleration

claudespice:
## Summary
Fixes #9872

The `TursoBetweenVisitor` AST rewriter used `CAST(... AS REAL)` to force numeric comparisons in BETWEEN expressions, but this didn't fix TPCH Q6 returning wrong results (75M instead of 123M revenue).

**Root cause:** Float64 arithmetic precision. When DataFusion sends `BETWEEN (0.06 - 0.01) AND (0.06 + 0.01)` to Turso, the upper bound `0.06 + 0.01` evaluates to `0.06999...` in float64, which is strictly less than the stored value `0.07` (nearest float64 to 7/100). This causes rows with `l_discount = 0.07` to be incorrectly excluded from the BETWEEN filter.

**Fix:** Replace `CAST(... AS REAL)` with `ROUND(..., 10)`. Rounding to 10 decimal places normalizes both the stored values and the arithmetic bounds to the same float64 representation, eliminating the precision mismatch. For example, `ROUND(0.06 + 0.01, 10)` and `ROUND(0.07, 10)` both evaluate to the same float64 value.

## Changes
- Changed `TursoBetweenVisitor` from `CAST(expr AS REAL)` to `ROUND(expr, 10)` wrapping
- Updated unit tests to verify ROUND-based rewrite

## Test plan
- [x] Unit test: numeric BETWEEN rewritten to ROUND comparisons
- [x] Unit test: arithmetic bounds BETWEEN rewritten correctly
- [x] Unit test: NOT BETWEEN uses OR with ROUND
- [x] Unit test: non-numeric BETWEEN left unchanged
- [x] Unit test: negative bounds handled correctly
- [x] `cargo clippy` passes
- [x] `cargo test` passes
- [ ] TPCH Q6 integration test with Turso acceleration (CI)

## Checklist
- [x] Regression tests updated
- [x] Edge cases covered
- [x] No snapshot deletions
- [x] Lint clean


crates/data_components/src/turso.rs
```diff
@@ -2274,12 +2300,12 @@ mod tests {
             "BETWEEN should be rewritten, got: {result}"
         );
         assert!(
-            result.contains("CAST(x AS REAL) >= CAST(0.05 AS REAL)"),
-            "should cast to REAL: {result}"
+            result.contains("ROUND(x, 10) >= ROUND(0.05, 10)"),
```
sgrebnov, line 2303: Additional round over table scan can affect performance and predicate pushdown (internal scan optimizations)?

claudespice: Good question. The tradeoff here is correctness vs. a hypothetical predicate-pushdown win.

Turso is a SQLite-compatible embedded engine. The prior code wrapped the column in `CAST(col AS REAL)` — same issue: wrapping any column in a function blocks the use of a plain index on that column for BETWEEN range scans. Swapping CAST for ROUND doesn't change index behavior; both block the same optimizations.

What ROUND buys us is **correct results**. The CAST approach still produced wrong TPCH Q6 output (75M vs. 123M revenue) because float64 arithmetic produces `0.06 + 0.01 = 0.06999...` which is strictly less than the stored `0.07`. ROUND(..., 10) normalizes both sides to the same decimal representation so the comparison is correct.

In practice the columns we care about here are Turso-stored decimals where numeric-range index pushdown isn't happening anyway (SQLite's weak typing + TEXT storage for decimals). Correctness on wrong-result bugs outweighs a predicate-pushdown optimization that wasn't firing in the first place.

If future benchmarks show this rewrite is actually pessimizing plans that *could* use an index on a REAL-typed column, we can narrow the rewrite to only fire when the target column is the known-problematic decimal shape. For now: correctness first.

lukekim (approved):

Issue #50, opened 2021-08-24
User friendly: restructure directory


lukekim:
Needs design.

`spiceai.yml` or `spiceai.config`

pods -> manifilepath -> whereever you want within your base project

`spice_pods`
`spicepods`

PR #10365, opened 2026-04-17
Harden /v1/tools and /v1/nsql against unauthenticated / LLM-driven SQL

lukekim:
Addresses threat-model items **#50** (`/v1/nsql` executes LLM-generated SQL) and **#51** (`/v1/tools/{name}` exposes direct tool invocation) tracked in [`docs/threat_models/v2.0.0.md`](docs/threat_models/v2.0.0.md). Combined with `#56` (anonymous-by-default API), #51 previously amounted to unauthenticated arbitrary SQL against every visible dataset.

### Runtime enforcement

- **New strict read-only validator** — `validate_sql_query_read_only` in `crates/runtime/src/datafusion/sql_validator.rs` rejects every DDL/DML/COPY/non-prepared `Statement` node regardless of per-catalog/per-dataset writability. Three unit tests cover SELECT-allowed, INSERT/DELETE/DROP-rejected.
- **`QueryBuilder::read_only(bool)`** — new opt-in flag on the query builder; plumbed through `Query` and applied at all three execution sites (local sync run, Ballista submit, async queries).

### Tool surface

- **Built-in `sql` tool defaults to read-only** — `SqlTool::new` now sets `read_only: true` and wires it to `QueryBuilder::read_only`. Description updated to advertise the posture. Operators can opt in via `SqlTool::allow_writes()`.
- **`/v1/tools/*` requires configured auth** — tool routes are split into a sub-router with a `require_auth_configured` middleware. When `runtime.auth` is absent the tool endpoints return **401** with guidance pointing at the auth docs. Search/chat/nsql continue to respect the existing top-level auth posture.

### NL→SQL surface

- **`/v1/nsql` runs LLM-generated SQL read-only** — the `QueryBuilder` used for model-produced SQL is unconditionally constructed with `.read_only(true)`, so prompt-injection-driven writes cannot reach writable catalogs. `application/sql`-mode (SQL-only, no execution) is unchanged.

### Docs

- Three new entries under "Security Hardening Mitigations (Implemented)" in `docs/threat_models/v2.0.0.md`, each cross-referenced to `#50` / `#51`.

## Out of scope (deferred)

The full remediation from the threat-model note also calls for a Spicepod-level allowlist for destructive tools (`sql`, `websearch`) and a client-side approval surface for NL-generated SQL. Those are larger, schema-affecting changes and are not included here; this PR is the minimal safe-by-default subset.

## Validation

- `cargo check -p runtime --features models --tests` passes.
- New unit tests in `sql_validator.rs` cover the read-only validator:
  - `test_read_only_validator_allows_select`
  - `test_read_only_validator_rejects_insert_on_writable_dataset`
  - `test_read_only_validator_rejects_delete_on_writable_dataset`
  - `test_read_only_validator_rejects_ddl`


docs/threat_models/v2.0.0.md
```diff
@@ -0,0 +1,310 @@
+# Threat Model for Spice.ai OSS v2.0.0
```
Jeadie, line 1: For auditability, this should get into `trunk` via its own PR.

crates/cache/src/lib.rs
```diff
@@ -551,6 +576,11 @@ impl QueryResultsCacheProvider {
                 | LogicalPlan::Dml(..)
                 | LogicalPlan::Copy { .. }
                 | LogicalPlan::Statement(..) => return false,
+                LogicalPlan::Extension(ext) => {
```
Jeadie: Remove level of nesting
```suggestion
                LogicalPlan::Extension(ext) if WRITE_CAPABLE_EXTENSION_NAMES.contains(&ext.node.name()) => {
```

crates/runtime/src/datafusion/query/cache.rs
```diff
@@ -96,13 +96,27 @@ enum CacheResult {
 
 impl Query {
     /// Returns a `LogicalPlan` if the result is not cached and needs to be executed, otherwise returns a cached `QueryResult`.
+    ///
+    /// When `read_only` is true, both the SQL-keyed and plan-keyed results-cache
```
Jeadie, line 100: Just a note: This PR and the prior work on user-level authentication is suggesting a auth-based caching structured where an authenticated user can read cache results from themselves + anonymous.

krinart (dismissed):

crates/cache/src/lib.rs
krinart: Fixed — flattened to a match guard pattern.

crates/runtime/src/datafusion/query/cache.rs
krinart: Noted — agreed, the current read-only bypass also informed the design of the cache key scoping. This is worth a dedicated follow-up to tighten the auth/cache interaction more broadly.

docs/threat_models/v2.0.0.md
krinart: Fair point on auditability. The threat model was updated here to keep the mitigations co-located with the implementing code, making it easy to review them together. Happy to extract it into a standalone PR if you prefer that workflow — let me know.

krinart (approved):

PR #10390, opened 2026-04-18
Add unified query cancellation across all paths

lukekim:
## Summary

Adds a unified query-cancellation foundation that flows through every query entry point: sync queries, async queries, Ballista queries, FlightSQL Prepare/Execute, `/v1/nsql`, and `/v1/search`.

## Design

Two shared primitives:

1. **Per-request `CancellationToken` on `RequestContext`** - every request carries a `tokio_util::sync::CancellationToken`. Per-query work uses a `child_cancellation_token()` so cancelling one query does not cancel the surrounding request scope.
2. **DataFusion-owned `QueryCancelRegistry`** - every executing synchronous `Query` registers a `Uuid` -> cancel token entry in the registry owned by that `DataFusion` instance. Entries are guarded by an RAII `ActiveQueryGuard` that auto-deregisters on stream drop.

The returned `RecordBatchStream` is wrapped so each `next().await` is raced against the cancel token. On cancel, the stream emits a single structured `QueryCancelled { query_id }` error and terminates, dropping the guard.

Cancellation is also checked during major planning and execution setup steps so cancelled requests can fail early before spending unnecessary work on planning or physical plan creation.

## Cancellation triggers

- **Client disconnect** - HTTP and Flight middleware attach a cancellation drop guard to the response body/stream. The guard disarms on normal end-of-stream and cancels the request token if the body is dropped early.
- **Explicit HTTP cancel for sync SQL** - `POST /v1/sql/{query_id}/cancel` cancels active synchronous SQL queries registered in the `DataFusion` query registry.
- **Explicit HTTP cancel for async jobs** - `POST /v1/queries/{query_id}/cancel` continues to cancel async jobs.
- **Explicit Flight cancel** - new `CancelQuery` Flight action accepting JSON `{query_id}` -> `{query_id, cancelled}`. This action requires write access.
- **Cooperative LLM cancel** - `/v1/nsql` checks `is_cancelled()` at the top of its loop and races `chat_request` against the operation token (returns HTTP 499).
- **Runtime shutdown** - `DataFusion::shutdown` cancels all active queries in its registry before accelerated table cleanup.

## New endpoints / actions

| Surface | Endpoint / action | Purpose |
| --- | --- | --- |
| HTTP | `GET /v1/sql/active` | List active synchronous SQL queries |
| HTTP | `POST /v1/sql/{query_id}/cancel` | Cancel an active synchronous SQL query |
| HTTP | `POST /v1/queries/{query_id}/cancel` | Cancel an async job query |
| Flight | `CancelQuery` action | Cancel a synchronous query via FlightSQL |

## Files changed

- `crates/runtime-request-context/src/context.rs` - `cancellation_token()`, `child_cancellation_token()`, `is_cancelled()`, `cancel()`, builder support
- `crates/runtime/src/datafusion/query/registry.rs` - `QueryCancelRegistry`, `ActiveQueryGuard`, `ActiveQueryInfo`
- `crates/runtime/src/datafusion/query.rs` - per-query `Uuid` + `CancellationToken`; early cancellation checks; stream wrapper; `QueryCancelled` error
- `crates/runtime/src/datafusion/query/builder.rs` - `cancellation_token()` builder method
- `crates/runtime/src/datafusion/mod.rs` and `builder.rs` - DataFusion-owned cancel registry and shutdown cancellation
- `crates/runtime/src/http/routes.rs` - drop-guard middleware + `/v1/sql/active` and `/v1/sql/{query_id}/cancel` routes
- `crates/runtime/src/http/v1/queries.rs` - async cancel remains on `/v1/queries`; sync active list/cancel live under `/v1/sql`
- `crates/runtime/src/flight/actions.rs` - `CancelQuery` action with write-access enforcement
- `crates/runtime/src/flight/middleware/request_context.rs` - drop-guard middleware
- `crates/runtime/src/http/v1/nsql.rs` - cooperative cancellation
- `crates/util/src/cancel_guard_body.rs` - shared response body wrapper for cancellation-on-drop behavior

## Notes

- `JobExecutor` async-job tokens are intentionally not linked to the request-context token, since async jobs outlive their submitting HTTP request. They remain reachable via the async cancel endpoint.
- `/v1/search` cancellation is automatic because it executes via the runtime `Query` path, which is already wrapped.

## Validation

- `RUSTC_WRAPPER= cargo test -p util --features http cancel_guard_body` - 4/4 passed
- `RUSTC_WRAPPER= cargo test -p runtime --features models datafusion::query::registry` - 6/6 passed
- `RUSTC_WRAPPER= cargo test -p runtime --features models http::v1::queries::tests` - 4/4 passed
- `RUSTC_WRAPPER= make lint` - passed


crates/runtime/src/http/routes.rs
```diff
@@ -300,6 +300,7 @@ pub(crate) fn routes(
     let queries_router = Router::new()
         .route("/v1/queries", post(v1::queries::submit))
         .route("/v1/queries", get(v1::queries::list))
+        .route("/v1/queries/active", get(v1::queries::list_active))
```
Jeadie, line 303: Using this route gives the impression that it is just for async SQL queries (e.g. Ballista), since all `/v1/queries/` prefixed routes are just async.

Jeadie: As a user I would expect `v1/queries/active` and `v1/sql/active`, for example

crates/runtime/src/http/cancel_guard_body.rs
```diff
@@ -0,0 +1,232 @@
+/*
```
Jeadie, line 1: Good candidate for `crates/util/src` to reduce size of `runtime` crate

crates/runtime/src/datafusion/query/registry.rs
```diff
@@ -0,0 +1,242 @@
...
+        self.registry.remove(self.query_id);
+    }
+}
+
+/// Process-wide query cancellation registry.
+///
+/// A single process runs a single Spice runtime, so a static registry is
+/// appropriate. Using a static registry avoids threading the registry handle
+/// through every query execution entry point.
+static GLOBAL_REGISTRY: LazyLock<RwLock<Arc<QueryCancelRegistry>>> =
```
Jeadie, line 178: Upon cancellation of `spiced` (i.e. Ctrl-C), do we need/want to explicitly cancel all queries in `GLOBAL_REGISTRY`?

crates/runtime/src/http/routes.rs
phillipleblanc: Async queries are the only ones that have the concept of a query list though, so this is only for async queries.

crates/runtime/src/datafusion/query/registry.rs
phillipleblanc, line 178: Static registries make it much harder to do any unit/integration testing. We explicitly redid some work that made data connector/accelerator registries not global and tying them to the `Runtime` instance. I would expect we do something similar here.

crates/runtime/src/http/routes.rs
Jeadie: I agree, but the route `list_active` checks sync query registry via `global_registry`.

crates/runtime/src/datafusion/query/registry.rs
phillipleblanc: Since the datafusion struct doesn't have a reference to the runtime, we could put the registry on the datafusion struct itself and let it own it - which would make sense.

peasee (dismissed):

sgrebnov (dismissed):

sgrebnov (dismissed):

krinart (approved):

krinart (dismissed):

krinart (approved):

PR #10428, opened 2026-04-20
Publish RC releases to DockerHub; upgrade runners to ubuntu-24.04

lukekim:
## Summary

Two related CI / release changes:

1. **Publish release candidates (RCs) to DockerHub / GHCR.** Previously the `spiced_docker` workflow only subscribed to `release: edited` and further gated on `make_latest` flipping to `true`. Creating a new pre-release fires `release: published` (and leaves `make_latest` unchanged), so RC tags never produced Docker images even though the git tag and GitHub release were created.
2. **Bump all GitHub Actions runners from `ubuntu-22.04` to `ubuntu-24.04`.** `ubuntu-22.04` is scheduled for removal from hosted runners; this moves every workflow onto the current LTS image at once.

### `spiced_docker.yml`

- **Trigger**: Subscribe to `release: published` in addition to `release: edited`. The `setup` job's `if` only allows `published` events through when:
  - The release is **not a draft** (`!github.event.release.draft`), and
  - The tag is an RC (`-rc.`) — other pre-release variants (nightly, beta, alpha, dev, etc.) are intentionally excluded so they cannot accidentally reach DockerHub / GHCR.
- **Edited path**: Still requires `!github.event.release.draft` plus `make_latest -> true`, so flipping a release to "latest" continues to move the `:latest` tag.
- **`:latest` tag handling**: Replaced the version-string heuristic (which used to infer from `-rc` / `-beta` / `-alpha`) with a direct query of the release's `isLatest` flag via `gh release view <tag> --json isLatest`. The Docker `:latest` tag now tracks GitHub's "latest" flag exactly. RCs and other non-latest releases publish version-tagged images only.
- **Draft protection (defense in depth)**: The resolve step also queries `gh release view <tag> --json isDraft` and fails the run for draft tags. This guards `workflow_dispatch` invocations as well.

### All workflows

- Runners bumped from `ubuntu-22.04` to `ubuntu-24.04`.
- Minor WSL distribution-name casing fix picked up along the way.

## Behavior summary (spiced_docker)

| Event | Outcome |
| --- | --- |
| `release: published`, draft | Skipped |
| `release: published`, RC tag (`-rc.`) | Publish `spiceai/spiceai:<version>` and `ghcr.io/spiceai/spiceai:<version>`; `:latest` only if GitHub `isLatest=true` |
| `release: published`, stable (no `-`) | Publish version tags; `:latest` only if GitHub `isLatest=true` |
| `release: published`, beta/alpha/nightly/dev | Skipped |
| `release: edited`, draft | Skipped |
| `release: edited`, `make_latest` flipped to true | Re-tag `:latest` |
| `workflow_dispatch` | Honors inputs; refuses to run for draft tags |

## glibc / compatibility note

Bumping the release-binary build job to `ubuntu-24.04` raises the glibc floor of the Linux x64 release tarball from 2.35 to 2.39. The Docker release images (`debian:bookworm-slim` = 2.36, `nvidia/cuda:…ubuntu22.04` = 2.35) will need to be updated in a follow-up to match, or the binary-producing job in `build_and_release.yml` should be kept on `ubuntu-22.04`. Tracked separately.


.github/workflows/spiced_docker.yml
```diff
@@ -37,8 +44,31 @@ on:
...
+    #     ':latest' tag can be moved without re-publishing on every edit.
+    if: >-
+      ${{
+        github.event_name != 'release'
+        || (
+          github.event.action == 'published'
+          && (
+            !contains(github.event.release.tag_name, '-')
+            || contains(github.event.release.tag_name, '-rc.')
+          )
```
lukekim:
```suggestion
          && contains(github.event.release.tag_name, '-rc.')
```

.github/workflows/install_scripts_test.yml
```diff
@@ -397,7 +397,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        distro: [Ubuntu-22.04]
+        distro: [ubuntu-24.04]
```
lukekim:
```suggestion
        distro: [Ubuntu-24.04]
```

phillipleblanc (approved):

.github/workflows/build_and_release.yml
```diff
@@ -94,7 +94,7 @@ jobs:
 
             // Tagged releases use GitHub hosted runners, others use spiceai runners
             const macosRunner = isTaggedRelease ? "macos-15" : "spiceai-macos";
-            const linuxRunner = isTaggedRelease ? "ubuntu-22.04" : "spiceai-dev-runners";
+            const linuxRunner = isTaggedRelease ? "ubuntu-24.04" : "spiceai-dev-runners";
```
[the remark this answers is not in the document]
phillipleblanc: I think this is fine for the OSS release

PR #10433, opened 2026-04-21
`runtime-cluster` crate + Run partition discovery before forwarding refresh to executors

krinart:
## 📝 Summary

  - Extract `runtime::cluster` to `runtime-cluster` crate
  - Add `PartitionService` — bundles partition store, executor registry, config, and app reference with 3 public methods:


crates/runtime-cluster/src/context.rs
```diff
@@ -0,0 +1,62 @@
...
+/// Implemented in the runtime crate by wrapping
+/// `cluster::partition::discovery::table_partition_values`, which unwraps the
+/// `AcceleratedTable` source and runs a `SELECT DISTINCT` against the federated
+/// source table.
+#[async_trait]
+pub trait PartitionDiscoverer: Send + Sync {
+    async fn table_partition_values(
+        &self,
+        table: &TableReference,
+        partition_by: &[PartitionedBy],
```
Jeadie, line 53: nit: I don't think we'd need the entire `PartitionedBy`?

crates/runtime/src/cluster/mod.rs
```diff
@@ -393,19 +396,27 @@ fn update_scheduler_pollers(
...
-pub(crate) mod executor_registry;
 pub mod metrics_collector;
 pub mod partition;
 mod scheduler_registry;
 mod servers;
 mod service;
 
+// Re-export executor registry types from `runtime-cluster` so in-crate callers can keep
+// importing them from `crate::cluster` (the executor_registry module moved out of runtime).
+pub mod executor_registry {
```
Jeadie, line 411: nit: not worth reexporting.

crates/runtime-cluster/src/context.rs
krinart: We do need it. It is used by `query_source_partitions(table, partitioning, df)`

crates/runtime/src/builder.rs
```diff
@@ -240,75 +242,81 @@ impl RuntimeBuilder {
...
+                    .and_then(|scheduler| scheduler.partition_management.clone())
+                    .and_then(|pm| PartitionManagementConfig::try_from(pm).ok())
+                    .map_or_else(AssignmentConfig::default, |pm_config| AssignmentConfig {
+                        max_assignments_per_cycle: pm_config.max_assignments_per_cycle,
+                        max_partitions_per_executor: pm_config.max_partitions_per_executor,
+                        discovery_timeout: pm_config.discovery_timeout,
+                    });
+                let partition_service = Arc::new(PartitionService::new(
+                    Arc::clone(&partition_store),
+                    Arc::clone(&executor_registry),
```
Jeadie, line 288: same as above/before. `assignment_config` is now an out of date version of what `shared_app` could provide (and is already provided to `PartitionService`).

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2860,6 +2889,31 @@ impl DataFusion {
...
+impl runtime_cluster::context::PartitionDiscoverer for DataFusion {
+    async fn table_partition_values(
+        &self,
+        table: &TableReference,
+        partition_by: &[spicepod::partitioning::PartitionedBy],
+    ) -> Result<Vec<runtime_cluster::PartitionValue>, Box<dyn std::error::Error + Send + Sync>>
+    {
+        crate::cluster::partition::discovery::query_source_partitions(table, partition_by, self)
+            .await
+            .map_err(|e| Box::new(e) as Box<dyn std::error::Error + Send + Sync>)
```
Jeadie, line 2918:
```suggestion
            .boxed()
```
I think?

crates/runtime/src/datafusion/mod.rs
```diff
@@ -2124,12 +2133,32 @@ impl DataFusion {
...
+        partition_service: &PartitionService,
         dataset_name: &TableReference,
         overrides: Option<&RefreshOverrides>,
     ) -> Result<Option<Arc<Notify>>> {
+        // Run on-demand partition discovery before forwarding the refresh command.
+        // This ensures that any new partition values in the source data are discovered,
+        // assigned to executors, and executors are notified -- before they receive the
+        // refresh command. Without this, the periodic partition management task (which
+        // runs on a 30-second interval) might not have discovered new partitions yet,
+        // causing executors to drop data from unassigned partitions. (fixes #10075)
```
Jeadie:
```rust
        // causing executors to drop data from unassigned partitions. (fixes #10075)
```
Will confirm with @peasee , but I believe no data was being dropped. #10075 means data (i.e. for a given partition value) was not being assigned and therefore ignored (until the next partition discovery loop).

crates/runtime/src/datafusion/mod.rs
```diff
@@ -572,6 +579,8 @@ pub struct DataFusion {
     pub executor_stream_registry: RwLock<Option<ExecutorControlStreamRegistry>>,
     /// Executor registry for distributed write forwarding (scheduler mode only).
     pub executor_registry: Option<Arc<ExecutorRegistry>>,
```
Jeadie: nit: could remove `executor_registry: Option<Arc<ExecutorRegistry>>,` and have `DataFusion` reference it via `partition_service`.

crates/runtime/src/cluster/mod.rs
```diff
@@ -822,16 +833,21 @@ pub(crate) async fn initialize_cluster_scheduler_future(
...
-                Arc::clone(&app),
-                &partition_manager,
-            )
-            .await
-            {
+            // Seed partition metadata for all accelerated tables. Requires the
+            // PartitionService to have been wired onto `DataFusion` during
+            // builder setup.
+            let df = rt.datafusion();
+            if let Some(partition_service) = df.partition_service.as_ref() {
```
Jeadie:
```suggestion
            if let Some(partition_service) = df.partition_service.as_ref() && 
               let Err(err) = partition::initialize_partition_metadata(partition_service, &df, &app).await
```
I think you can now do this in rust (i.e. interdepdent assignments).

crates/runtime/src/cluster/mod.rs
```diff
@@ -813,7 +824,7 @@ pub(crate) async fn initialize_cluster_scheduler_future(
     };
 
     if let Some(config) = app.runtime.scheduler.clone() {
-        if let Some(partition_manager) = rt.partition_manager() {
+        if let Some(_partition_store) = rt.partition_store() {
```
Jeadie:
```suggestion
        if rt.partition_store().is_some() {
```

crates/runtime-cluster/src/context.rs
```diff
@@ -0,0 +1,62 @@
...
+#[async_trait]
+pub trait PartitionExprResolver: Send + Sync {
+    async fn try_parse_expr(
+        &self,
+        tbl: &TableReference,
+        expr: &str,
+    ) -> Result<Expr, DataFusionError>;
+}
+
+/// Queries the source table for its current partition values.
```
Jeadie:
```suggestion
/// Discovers the values a `table` can have for a given `partition_by` expression.
///
/// For >1 `partition_by` value, the cartesion product of individual options is returned. 
```

crates/runtime-cluster/src/scheduler_task_config.rs
```diff
@@ -0,0 +1,107 @@
...
+    InvalidDiscoveryTimeout {
+        timeout: String,
+        source: fundu::ParseError,
+    },
+
+    #[snafu(display("Partition management discovery timeout must be greater than zero"))]
+    DiscoveryTimeoutIsZero,
+}
+
+impl TryFrom<spicepod::component::runtime::PartitionManagement> for PartitionManagementConfig {
```
Jeadie, line 65: Make sure this will be compatible with https://github.com/spiceai/spiceai/pull/10450

crates/runtime/src/cluster/mod.rs
krinart: Done — removed the `pub mod executor_registry` re-export module. All callers now import `ExecutorRegistry` and `TablePartitions` directly via `crate::cluster::{ExecutorRegistry, TablePartitions}` (both are re-exported at that level).

crates/runtime/src/builder.rs
krinart: Done — removed `config: AssignmentConfig` from `PartitionService` entirely. The service now derives `AssignmentConfig` from `self.app` at the start of each `reconcile_all` / `reconcile_table` / `seed_table` call via a private `config_from_app` helper. The builder no longer needs to compute or pass `assignment_config`.

crates/runtime/src/datafusion/mod.rs
krinart: The function uses `async fn` with a direct `.await` inside — the return value is already a `Result<Vec<...>>`, not a `Future`. Adding `.boxed()` would change the return type to `Pin<Box<dyn Future<...>>>` which wouldn'\''t match the trait signature. The `async_trait` macro handles boxing the future automatically. No change needed here.

crates/runtime/src/datafusion/mod.rs
krinart: Fixed — updated the comment to say "ignored" instead of "dropped", and removed the hard-coded 30-second interval.

crates/runtime/src/datafusion/mod.rs
krinart: Done

crates/runtime/src/cluster/mod.rs
krinart: Done — combined into a single `if let ... && let Err(err) = ...` chain.

crates/runtime/src/cluster/mod.rs
krinart: Done — changed to `if rt.partition_store().is_some()`.

crates/runtime-cluster/src/context.rs
krinart: Done — updated the doc comment as suggested.

crates/runtime-cluster/src/scheduler_task_config.rs
krinart: Noted — #10450 renames `PartitionManagementConfig` → `PartitionAssignmentConfig` and flattens the scheduler config fields. This will conflict with the `TryFrom` impl here and the new `config_from_app` helper in `service.rs`. Will sync / rebase once #10450 merges.

krinart:
Closed in favor of https://github.com/spiceai/spiceai/pull/10490

PR #10439, opened 2026-04-21
provide and use row_ids from original vortex scan

Jeadie:


Jeadie:
optimisation did not show signigicant impact in SF1 testing. Code is also stale.

PR #10457, opened 2026-04-22
Improve `ready_state: on_schema_resolved`

krinart:
## 📝 Summary

For accelerated datasets with `on_schema_resolved` ready states, queries arriving during initial load are now federated to the source using the correct fully-qualified table - new `DynamicSQLTable` struct is added to handle such cases properly.

Plus tests.

### DIFF
--- a/crates/runtime/src/accelerated_table/federation/mod.rs
+++ b/crates/runtime/src/accelerated_table/federation/mod.rs
@@ -174,7 +174,7 @@ mod tests {
     /// fallback (source) table reference so federation generates fully-qualified SQL.
     #[tokio::test]
     async fn test_dynamic_table_ref_uses_fallback_before_load() {
-        let refresher = make_refresher();
+        let refresher = make_refresher().expect("make_refresher");
         assert!(!refresher.initial_load_completed());
 
         let table = DynamicSQLTable {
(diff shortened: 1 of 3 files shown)


lukekim (dismissed):

crates/runtime/src/accelerated_table/federation/mod.rs
```diff
@@ -37,26 +72,75 @@ impl AcceleratedTable {
...
+        // Only fall back to the federated source during initial load for ready states that
+        // are designed to serve queries before acceleration completes. The default `OnLoad`
+        // state must NOT federate during initial load — it returns "Acceleration not ready".
+        let fallback_during_initial_load = matches!(self.ready_state, ReadyState::OnSchemaResolved);
+
+        let table_source: Arc<dyn FederatedTableSource> = match fallback {
+            Some((fb_provider, fb_table_ref)) if fb_table_ref != accelerated_table_ref => {
+                // Federated source uses a different (typically fully-qualified) table reference.
+                // Use a dynamic table that presents the correct reference during each phase.
+                let dynamic_table: Arc<dyn SQLTable> = Arc::new(DynamicSQLTable {
```
phillipleblanc: I don't see the point in conditionally creating the DynamicSQLTable, we should just always create it

claudespice: Addressed in e9e75a188 — collapsed the conditional into a single path that always constructs DynamicSQLTable. When no fallback is available, fallback_table_ref defaults to the accelerated_table_ref, making the dynamic switch a no-op.

PR #10463, opened 2026-04-22
feat(cluster): consolidate distributed state into cluster.json

phillipleblanc:
## Summary

Distributed scheduler state lived across four kinds of files: `metadata/cluster.json`, `schedulers/{id}.json`, `accelerations/partitions/{table}.json`, and `catalog/partitions/{table}.json`. This PR consolidates all of them into a single OCC-protected document at `cluster.json`, and moves heartbeats out into per-scheduler files at `heartbeats/{id}.json` so the high-frequency write path doesn't contend with everything else.

Discovery becomes a single `cluster.json` read instead of N prefix lists. Job state (`jobs/`) is unchanged.

## Layout

```
{base_prefix}/
├── cluster.json                   # NEW: schedulers + accelerations + catalog (OCC)
├── heartbeats/
│   └── {scheduler_id}.json        # NEW: heartbeat-only, frequent writes
└── jobs/                          # unchanged
    ├── {job_id}.json
    └── {job_id}/chunk_*.arrow
```

Removed: `metadata/cluster.json`, `schedulers/{id}.json`, `accelerations/partitions/*.json`, `catalog/partitions/*.json`.

## `cluster.json` schema (v1)

```jsonc
{
  "schema_version": 1,                  // u32, currently 1
  "created_at_ms":  1776849553512,      // u64, ms since epoch, set on bootstrap
  "updated_at_ms":  1776849732709,      // u64, ms since epoch, refreshed on every commit

  // Registered schedulers, keyed by scheduler_id (= node-bind-address `host:port`).
  // Heartbeats live in heartbeats/{scheduler_id}.json, NOT here.
  "schedulers": {
    "127.0.0.1:50071": {
      "scheduler_id":      "127.0.0.1:50071",
      "instance_id":       "0d8a78b9-d455-4a2e-a40d-9f0e73a1ae5b",  // UUID, regenerated every process start
      "advertise_address": "127.0.0.1:50071",
      "grpc_address":      "127.0.0.1:51071",
      "http_address":      "127.0.0.1:18090",
      "started_at_ms":     1776849732606,
      "ttl_ms":            30000,
      "build_version":     "2.0.0-unstable",
      "labels":            {}             // HashMap<String, String>, defaults to {}
    }
  },

  // Partition metadata for accelerated tables, keyed by normalized table name
  // ("catalog.schema.table", lowercased, defaults filled in).
  "accelerations": {
    "spice.public.taxi_trips": {
      "table_name":             "spice.public.taxi_trips",
      "schema_version":         1,
      "updated_at":             1776849732606,                       // u128 ms since epoch
      "partition_expressions":  ["bucket(3, c_nationkey)"],          // Vec<String>, original SQL exprs
      "partitions": [
        {
          "partition_value":      { "expr0": "0" },                  // HashMap<String, String>
          "assigned_executors":   ["127.0.0.1:52071"],                // Vec<String>, defaults to []
          "last_assigned_at":     1776849732606                       // u128, omitted if never assigned
        }
      ]
    }
  },

  // Same shape as `accelerations`, for catalog/federated tables.
  "catalog": {}
}
```

All three top-level maps default to `{}` when missing on read. `labels` and `assigned_executors` default to empty; `last_assigned_at` is omitted when `None`.

`heartbeats/{scheduler_id}.json` is a separate document so the high-frequency heartbeat write path doesn't contend with cluster mutations:

```jsonc
{
  "scheduler_id":      "127.0.0.1:50071",
  "instance_id":       "0d8a78b9-d455-4a2e-a40d-9f0e73a1ae5b",  // must match cluster.json entry; mismatches are filtered as orphans
  "last_heartbeat_ms": 1776849733511,
  "ttl_ms":            30000
}
```

## Design notes

- `ClusterStateStore::mutate` is the only write path. On `Conflict` it refetches and re-runs the closure. `NoChange` returned against a stale cached snapshot is verified against a fresh read before returning `AlreadySatisfied` (the "check whether the change is already applied on conflict and re-read" semantics requested in design review). `NotFound` mid-flight is surfaced, never silently re-bootstrapped.
- `SchedulerEntry` and `SchedulerHeartbeat` carry an `instance_id` (UUID per process). The takeover decision happens inside the OCC closure on `cluster.json`, eliminating the TOCTOU window between "observe stale heartbeat" and "overwrite entry".
- Reaper removes a dead scheduler from `cluster.json` only when the `instance_id` it observed still matches. It does **not** delete heartbeat files (object store has no conditional delete, and orphans are filtered out by `list_alive`).
- Membership RPC keeps reading the pre-filtered in-memory `peers` map, not raw `cluster.read_cached()`, so executor failover latency stays at one discovery interval rather than `discovery + reaper`.
- `PartitionManager` is parameterised by `PartitionScope::{Acceleration, Catalog}`; two type aliases (`AccelerationsPartitions`, `CatalogPartitions`) document scope at call sites.
- New `apply_assignments(&[AssignmentRequest])` batches many `(table, partition, executor)` updates into one cluster-document write — the previous per-partition write loop in `scheduler_task::commit_assignments` would amplify badly against the now-shared document.
- Executor disconnect on a reaped scheduler is unchanged: the existing `fetch_scheduler_membership` poll feeds `ControlStreamManager`, no proto changes.

## Migration — clean break

Pre-RC, no readers for the old layout. **Operators of distributed deployments must wipe** `metadata/`, `schedulers/`, `accelerations/`, and `catalog/` under `runtime.scheduler.state_location` before upgrading. **Do NOT wipe `jobs/`** — job state is unchanged and contains user-visible async query results. The new `heartbeats/` prefix is intentionally distinct from the old `schedulers/` prefix so leftover legacy files cannot break heartbeat listing.

## Tests

Unit tests cover:

- `mutate` apply / NoChange-against-fresh / NoChange-from-stale-cache-forces-fresh-read / conflict-effective-no-op / mid-flight `NotFound` / `Arc` snapshot identity.
- Heartbeat `list_alive` filtering for both stale and orphan `instance_id`; `list_all_evicts_deleted_keys`; concurrent heartbeats convergence.
- Reaper: stale heartbeat / missing heartbeat after grace / fresh heartbeat / recently-registered without heartbeat / `instance_id` mismatch / concurrent idempotency / `reaper_does_not_delete_heartbeat_files`.
- Partition wrappers: ported all `copy_assignments_*` and `table_key_*` tests; added `acceleration_and_catalog_share_no_state` and `apply_assignments_writes_once_for_many_partitions`.
- `registry_runner_basic_register_and_shutdown`.

Manual end-to-end with two file-backed schedulers against `/tmp/spice-cluster-state`:

- Layout: only `cluster.json` + `heartbeats/<id>.json` are created. `cluster.json` mtime changes only on membership/partition writes; heartbeat mtimes update independently.
- Two-scheduler discovery: each sees the other within one discovery interval.
- Graceful shutdown: entry removed from `cluster.json` and heartbeat file deleted within ~1s.
- SIGKILL: entry stays in `cluster.json` until reaper fires (~ttl + jitter ≈ 30–36s); reaper logs `Reaped stale scheduler`; orphan heartbeat is left in place.
- Takeover after reap: fresh process registers with new `instance_id`; orphan heartbeat is overwritten.
- Takeover during the reaper window with a still-fresh heartbeat: registration correctly refused with `Scheduler registration record already exists … and is still active`, matching pre-existing semantics.

`make lint-rust` ✅

### DIFF
--- a/crates/runtime/Cargo.toml
+++ b/crates/runtime/Cargo.toml
@@ -190,7 +190,7 @@ utoipa-swagger-ui = { version = ">=8.1.1", features = [
 ], optional = true }
 
 gethostname.workspace = true
-uuid = { workspace = true, features = ["v7"] }
+uuid = { workspace = true, features = ["v4", "v7"] }
 workers = { path = "../workers" }
 x509-certificate.workspace = true
 
--- a/crates/runtime/src/builder.rs
+++ b/crates/runtime/src/builder.rs
@@ -18,7 +18,7 @@ use crate::cluster::DistributedNode;
 use crate::cluster::ExecutorRegistry;
 use crate::cluster::PartitionManager;
 use crate::cluster::ResolvedClusterConfig;
-use crate::cluster::partition;
+use crate::cluster::{ClusterStateStore, SchedulerHeartbeatStore};
 use crate::config::ClusterRole;
 use crate::config::Config;
 use crate::datafusion::udf::register_udfs;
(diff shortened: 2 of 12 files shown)


phillipleblanc:
Addressed all four Copilot review comments in 6344000:

- **`apply_assignments_inner` error reporting**: the mutator now captures the exact failing `(table_key, partition_value)` into a `missing_key` slot before `Abort`. The post-mutate match uses that captured info, so errors on multi-table batches name the actual failing table instead of guessing from `assignments[0]`. Removed the substring matching.
- **`now_ms().unwrap_or(0)` in reaper tick**: replaced with explicit handling — on clock error we log a warning and skip the tick.
- **`Error::RetryExhausted` losing root cause**: now carries the last `ObjectStoreError` as `source`, so the error message includes it and snafu chains the cause.
- **`Reaper::tick` reusing `MutateError::Conflict`**: introduced a dedicated `reaper::Error` enum with `ClusterRead`, `ClusterMutate`, and `ListHeartbeats` variants; the heartbeat IO failure is no longer conflated with semantic conflicts.

phillipleblanc:
## CI status

All build/lint/static-analysis/most-test jobs pass. Two failing jobs are pre-existing trunk failures, not introduced by this PR:

1. **`duckdb_append_with_indexes.yaml acceleration and graceful shutdown`** — flaky DuckDB shutdown test. Failed on first run (clean unclean-shutdown WAL race), **passed on rerun** with the same code. Adjacent variants (`duckdb_append_with_pk`, `duckdb_append_with_pk_and_indexes`) also fail intermittently on trunk (e.g. trunk run 24752891924).

2. **`Test Models, AI & Search`** — `models::search::test_multi_column_search_view` and `models::s3_vectors::search::multiple_embeddings` fail with `insta` snapshot mismatches on the `*_vector_search_sql_filters` EXPLAIN-plan snapshots. Pure search-module test, no overlap with cluster state code in this PR. `integration tests (models)` has failed on **every** recent trunk run today (00:50, 02:42, 05:45, 06:22, 07:30 UTC), so it's a snapshot that needs updating from a recent unrelated change in the search/explain plan.

Lint (`make lint-rust` equivalent), `Build and Test`, `Build Test Binary` (Linux/macOS/Extended), `Analyze (rust)`, `Build (dev/release)`, `Rust Lint`, all integration-tests-without-snapshot-deps, and the four review-comment fixes from Copilot all pass.

Ready for review/merge from my side.

phillipleblanc:
## Local verification (file + S3)

Built locally (`cargo build --bin spiced`) and ran end-to-end on both backends.

### File backend (`file:///tmp/spice-cluster-state`)
- ✅ Two-scheduler discovery: both register; cluster.json contains both `scheduler_id` entries with distinct `instance_id`s; `heartbeats/127.0.0.1:5007{1,2}.json` present.
- ✅ Graceful shutdown of s2 (Ctrl-C): s2 entry removed from cluster.json; heartbeat file preserved.
- ✅ SIGKILL of s1 + replacement at same address: first restart attempt within TTL refused (`Scheduler registration record already exists for 127.0.0.1:50071 and is still active`); after TTL+grace, replacement registers and `instance_id` rotates (`5807ff31...` → `71bbb3fb...`).

### S3 backend (`s3://spiceai-phillip-cluster-state-seoul/260422-cluster-state-test-181853/` ap-northeast-2, `s3_key`/`s3_secret` from ${secrets:})
- ✅ Layout on S3:
  ```
  cluster.json
  heartbeats/127.0.0.1:50071.json
  heartbeats/127.0.0.1:50072.json
  ```
- ✅ cluster.json contents match expected schema (schedulers map keyed by scheduler_id, with instance_id, ttl_ms=30000, build_version, etc.).
- ✅ Reaper: after killing s2 abruptly, s1's reaper evicted the stale entry — `Reaped stale scheduler scheduler_id=127.0.0.1:50072 instance_id=b6fde507-...`, followed by `Reaper tick evicted=["127.0.0.1:50072"] skipped=[]`. Heartbeat file `50072.json` was deliberately **not** deleted (per design — object_store has no conditional-delete).
- ✅ SIGKILL+takeover: SIGKILL s1, wait 35s for heartbeat to go stale, restart at same address — registration succeeded with new instance_id (`1076eea6...` → `0d8a78b9...`).

S3 prefix and local state directory cleaned up after the run.

sgrebnov (approved):

PR #10464, opened 2026-04-22
Improve full-text search indexing performance

sgrebnov:
## 📝 Summary

Problem: Full-text search index building is slow due to
1. A new `IndexWriter` was created and dropped on every batch, killing Tantivy's background merge threads each time and accumulating many tiny segments
2. The writer heap budget was **15 MB** (Tantivy's minimum), causing segment flushes too frequent

Fix:
- Make `IndexWriter` persistent across batches via `Arc<Mutex<IndexWriter>>` on the struct, so merge threads stay alive
- Increase writer memory budget from 15 MB to 150 MiB, reducing segment flush frequency for large batches

## Performance Impact

Benchmarked with **Amazon Reviews** dataset (full-text search index on review text column).

| Metric | Before | After | Change |
|--------|--------|-------|--------|
| Throughput | 6.14–6.71 MiB/s | 10.29–10.41 MiB/s | **+60%** |
| Ingestion rate | ~1,378 rows/s | ~2,348 rows/s | **+70%** |
| Segments per merge | ~500 | 8 | **63x fewer** |
| Merge duration | blocking (seconds) | ~50ms (background) | **orders of magnitude faster** |
| Projected time (35.5M rows) | ~6 hrs (degrading) | ~4.2 hrs (stable) | **~30% faster** |

**Root causes fixed:**
1. `IndexWriter` was created and dropped on every batch — merge threads were killed before completing, causing unbounded segment accumulation. Now the writer is retained for the lifetime of the index via `Arc<Mutex<IndexWriter>>`.
2. Heap budget was 15 MB (Tantivy's minimum) — segments flushed at ~2,900–7,600 docs. Increased to 150 MiB, producing consistent 8,192-doc segments with healthy background merges.

**Before**: Throughput degraded over time as segment count grew (merge cost is superlinear in segment count).
**After**: Throughput remains stable — Tantivy's `LogMergePolicy` keeps segment count bounded via incremental 8-way merges.

## 👀 Notes for Reviewers

Future improvement: Avoid committing every batch — instead batch multiple `RecordBatch`es and commit periodically (e.g. every N batches/records or every M seconds). This would require changes to the `compute_index` / CDC envelope contract, since callers currently expect data to be searchable immediately after `compute_index` returns.


phillipleblanc (dismissed):

crates/search/src/generation/text_search/index.rs
```diff
@@ -51,8 +53,10 @@ pub struct FullTextDatabaseIndex {
     pub primary_key: Vec<String>,
     pub base_table: Arc<dyn TableProvider>,
 
-    // `index` access should only be needed for write path. Query/reads should use [`self::reader`].
+    // Retained for shared ownership across clones (e.g. `with_new_base`).
+    // Writes go through [`Self::writer`]; reads through [`Self::reader`].
     pub index: Arc<Mutex<tantivy::Index>>,
```
Jeadie:
```suggestion

```
Do we need `.index` now?

Jeadie (dismissed):

crates/search/src/generation/text_search/index.rs
sgrebnov: Removed - thanks you

Jeadie (approved):

PR #10470, opened 2026-04-22
DynamoDB DML

krinart:
## 📝 Summary

* Add support for `update` and `delete_from` to `DynamoDBTableProvider`.

* Add configurable parallelism (added `write_parallelism` configuration parameter)

* Tests

Also: bump `docx-rs` version since `zip 7.4.0` version was yanked (using yanked versions is not allowed for new projects such as DynamoDB ADBC driver)

Originally implemented in https://github.com/spiceai/spiceai/pull/9915, then reverted in https://github.com/spiceai/spiceai/pull/10114


crates/data_components/src/dynamodb/dml/update.rs
```diff
@@ -0,0 +1,257 @@
...
+        for result in results {
+            result?;
+            total += 1;
+        }
+    }
+
+    Ok(total)
+}
+
+type UpdateExpression = (
```
phillipleblanc: why not just make this a struct so that the types are explicit what they represent? Otherwise at least add a comment explaining what they are.

phillipleblanc (dismissed):

phillipleblanc (approved):

crates/data_components/src/dynamodb/dml/update.rs
krinart: Done

PR #10487, opened 2026-04-23
feat(secrets): add ParameterSpec and more params for AWS secrets manager

phillipleblanc:
**Stacked on top of #10478 (`lukim/aws-secret`).** Review/merge after #10478 lands; the base will then need to be rebased onto trunk.

## Why

Today, `params:` keys on a secret store are silently dropped — typing `regoin: us-east-1` on `aws_secrets_manager` does nothing and the user gets a confusing AWS error from the default region resolver. There's also no way to set `region`, `endpoint_url`, static AWS credentials, or a Kubernetes `namespace` per-store; you have to fall back to the AWS env-var ladder, which doesn't compose with multiple stores.

## What this PR does

Brings secret stores in line with how data connectors handle configuration:

1. **Each store declares a static `ParameterSpec` list** — same struct connectors use, extracted into a new leaf crate `runtime-parameter-spec` (re-exported by `runtime-parameters` so all ~90 existing call sites compile unchanged; the alternative was a dependency cycle through `runtime-secrets`).
2. **User-supplied `params:` are validated against the spec.** Unknown params hard-fail with an error listing the supported ones — instead of being silently dropped (the failure mode this whole PR exists to fix). Defaults, `required`, `one_of`, and `deprecated` warnings are honored.
3. **`${ env:KEY }` and `${ secrets:KEY }` references in `params:` values are resolved before validation**, against a single bootstrap `EnvSecretStore` (which also loads `.env` / `.env.local` via dotenvy). `secrets:KEY` resolves only against env at this point because no other store is loaded yet — references to other stores are rejected with a clear error explaining the bootstrap-cycle reason. Missing env vars are a hard error, not silent empty-string expansion.

### Per-store params added

| Store | New params |
|---|---|
| `aws_secrets_manager` | `region`, `endpoint_url`, `key`, `secret`, `session_token` |
| `kubernetes` | `namespace` (also lets the store run outside a pod) |
| `env` | `file_path` (was previously read via ad-hoc lookup; now goes through validation) |
| `keyring` | none — but typos still get rejected |

- `region` flows into a cached `SdkConfig` via `aws_sdk_credential_bridge::get_or_init_sdk_config_with_region`, matching the S3 connector pattern.
- `endpoint_url` flows into the Secrets Manager client builder (VPC endpoints, FIPS, LocalStack).
- `key` / `secret` / `session_token` (named to match the S3 connector vocabulary) flow into `aws_credential_types::Credentials` via `.credentials_provider(...)`. When both `key` and `secret` are supplied they take precedence over the SDK's default credential chain. Mismatched pairs (only one set) emit a warning and fall back to the chain rather than failing; a stray `session_token` without creds is warned + ignored.

### Example

```yaml
secrets:
  - from: env
    name: env
    params:
      file_path: .env.local

  - from: aws_secrets_manager:my-secret
    name: aws
    params:
      region: ${ env:AWS_REGION }
      key: ${ env:AWS_ACCESS_KEY_ID }
      secret: ${ env:AWS_SECRET_ACCESS_KEY }
      # session_token: ${ env:AWS_SESSION_TOKEN }   # for STS-issued creds
      # endpoint_url: https://localhost:4566        # for LocalStack / VPC endpoints

  - from: kubernetes:my_secret
    name: k8s
    params:
      namespace: spice
```

A typo like `regoin: us-east-1` now fails fast with:

> Unknown parameter 'regoin' for secret store 'aws_secrets_manager'. Supported parameters: region, endpoint_url, key, secret, session_token.

## Internal change: `SecretStoreType` carries typed config

Variants switched from raw strings to typed config structs (`EnvConfig`, `KubernetesConfig`, `AwsSecretsManagerConfig`); the old `Env` / `EnvCustomPath` split is collapsed.

## Tests

38 unit tests pass. New tests cover:
- env / aws / kubernetes param round-trip through `spicepod_secret_store_type`
- static-credential params (`key` / `secret` / `session_token`) round-trip onto the resolved config
- `regoin` typo is rejected with an error listing supported params
- `${ env:KEY }` and `${ secrets:KEY }` bootstrap resolution
- missing env var fails fast with the var name in the error
- `${ kubernetes:... }` reference inside `params:` is rejected

## Verification

- `cargo test -p runtime-secrets --lib` — 38 passed
- `make lint-rust` — clean

## Out of scope / follow-ups

- Schema enrichment in `tools/spicepodschema` to surface secret-store `ParameterSpec` lists in `spicepod.schema.json` for IDE completion (same mechanical pattern as connectors).
- Doc updates on docs.spiceai.org/secret-stores/aws-secrets-manager.md and the `spice-secrets` skill.
- Additional AWS params worth adding next: `profile`, `assume_role_arn`, `iam_role_source` (mirroring the S3 connector's `auth` / `iam_role_source`). The infrastructure is now in place — just add `ParameterSpec` entries.


phillipleblanc:
Pushed an update adding static-credential params to `aws_secrets_manager`:

| Param | Notes |
|---|---|
| `key` | AWS access key ID. Must be set together with `secret`. |
| `secret` | AWS secret access key. Must be set together with `key`. |
| `session_token` | Optional. Only meaningful alongside `key` + `secret` (e.g. STS). |

Naming matches the S3 connector (`key`, `secret`, `session_token`) so AWS-flavored components share a vocabulary. When both `key` and `secret` are supplied they take precedence over the SDK's default credential chain via `aws_credential_types::Credentials`. Mismatched pairs (only one set) emit a warning and fall back to the chain rather than failing.

Example:

```yaml
secrets:
  - from: aws_secrets_manager:my-secret
    name: aws
    params:
      region: ${ env:AWS_REGION }
      key: ${ env:AWS_ACCESS_KEY_ID }
      secret: ${ env:AWS_SECRET_ACCESS_KEY }
      # session_token: ${ env:AWS_SESSION_TOKEN }
```

Tests: 38 passing (added `test_aws_secrets_manager_static_credentials_threaded_through`).
`make lint-rust` clean.

ewgenius (approved):

PR #10490, opened 2026-04-23
`runtime-cluster` crate + Run partition discovery before forwarding refresh to executors

krinart:
## 📝 Summary

  - Extract `runtime::cluster` to `runtime-cluster` crate
  - Add `PartitionService` — bundles partition store, executor registry, config, and app reference with 3 public methods:


Jeadie (dismissed):

crates/runtime-cluster/src/executor_registry.rs
```diff
@@ -133,7 +126,7 @@ pub type TablePartitions = HashMap<TableReference, Vec<Expr>>;
 #[derive(Debug)]
 pub struct ExecutorRegistry {
```
phillipleblanc, line 127: Now that ExecutorRegistry lives in its own crate, the only way callers in runtime can interact with these maps is by reaching into the RwLocks directly. We end up with ~10 ad-hoc usages that all encode the same access pattern. Suggest replacing the public fields with intent-revealing
methods. Concretely:

Replace:

```rust
  pub struct ExecutorRegistry {
      pub connections: Arc<RwLock<HashMap<String, ExecutorConnection>>>,
      pub flight_sql_clients: Arc<RwLock<HashMap<String, FlightSqlClient>>>,
      pub partitions: Arc<RwLock<HashMap<String, TablePartitions>>>,
      ...
  }
```

with private fields plus:

```rust
  impl ExecutorRegistry {
      // For do_put.rs:215, cayenne_ddl/physical_plans.rs:{277,521,1001}
      pub async fn has_flight_sql_clients(&self) -> bool { ... }

      // For write_through.rs:{329,803}, iceberg_ddl/physical_plans.rs:993,
      // cayenne_ddl/physical_plans.rs:94 — they all clone the map and iterate
      pub async fn flight_sql_clients_snapshot(&self)
          -> HashMap<String, FlightSqlClient> { ... }

      // For cluster/service.rs:600
      pub async fn insert_flight_sql_client(&self, executor_id: String, client: FlightSqlClient) { ... }

      // For cluster/service.rs:709
      pub async fn set_executor_partitions(&self, executor_id: String, partitions: TablePartitions) { ... }

      // For init/task_history.rs:138 — needs a shared handle, keep as a typed accessor
      pub fn flight_sql_clients_handle(&self) -> Arc<RwLock<HashMap<String, FlightSqlClient>>> { ... }
  }
```

This also lets unregister start cleaning up flight_sql_clients and partitions consistently — today removing an executor only touches
connections, which is exactly the kind of invariant pub fields make impossible to enforce.

krinart: Done

phillipleblanc (approved):

PR #10503, opened 2026-04-24
Use Ballista query for partition discovery

Jeadie:
## 📝 Summary
  - Make partition discovery non-blocking by submitting discovery queries as Ballista jobs instead of executing them synchronously. This is especially important for large datasets.
  - On startup, statically resolvable partitions (e.g. bucket(N, col)) are seeded immediately; dynamic partitions submit a Ballista job and return without waiting
  - The periodic PartitionManagementTask follows a check/submit/assign pattern: poll completed discovery jobs, submit new ones for tables without active jobs, then assign pending partitions
  - On-demand refresh (spice refresh) still blocks — submits a discovery job and polls until completion with timeout
  - Add `allow_accelerated_tables` flag to Query/SubmitQueryRequest/JobState so internal discovery queries can go through `submit_distributed` against accelerated table names (which resolve to federated sources on the scheduler)

Design decisions:
 - Reuse existing Ballista job machinery. Discovery queries are regular JobExecutor/JobStore jobs — they appear in the object store at jobs/{job_id}.json, results are Arrow IPC chunks, and they're visible through /v1/queries.
 - Track discovery jobs in TablePartitionMetadata. A new active_discovery_job_id: Option<String> field (OCC-protected via cluster state) lets any scheduler poll a discovery job submitted by another scheduler. Multi-scheduler race conditions
   are handled by the existing OCC mutate() pattern.

## 🔗 Related
Testing SF100 with nonstatically resolvable partitioning expressions showed repeat timeouts and excess load on the scheduler.


Jeadie:
Still important, but will need significant refactor

PR #10564, opened 2026-04-27
Enable `cosine_distance` pushdown to DuckDB accelerator via `array_cosine_distance`

sgrebnov:
## 📝 Summary

DuckDB natively supports `array_cosine_distance`, and the Spice DuckDB dialect already translates `cosine_distance` to it. However, `cosine_distance` was included in the global Spice UDF deny list, which blocked it from being pushed down to DuckDB during SQL federation.

This PR creates a DuckDB-specific deny list that allows `cosine_distance` through while continuing to block all other Spice-internal UDFs. The three DuckDB accelerator variants (standard, partitioned, and tables-mode partitioned) now use this refined deny list.

## 👀 Notes for Reviewers

The `AcceleratedTable` filter pushdown layer (`supports_filters_pushdown` / `filters_for_accelerator_scan`) still uses the global deny list. This only affects `WHERE` clause predicates, so `cosine_distance` is not affected as it is constructed as part of a projection. We still need to improve this later for consistency so the same deny list is applied for `supports_filters_pushdown` as well.


lukekim (dismissed):

Jeadie:
reverted snapshots on merge conflict, now rerunning snapshots via https://github.com/spiceai/spiceai/actions/runs/25035133133

Jeadie (approved):

Jeadie (approved):

phillipleblanc (approved):

PR #10578, opened 2026-04-28
feat(postgres): support inline PEM sslrootcert

phillipleblanc:
## Summary
- Bump `datafusion-table-providers` to include inline `pg_sslrootcert` PEM support from datafusion-contrib/datafusion-table-providers#634
- Update Postgres connector/catalog/accelerator parameter descriptions to document path or inline PEM content

## Testing
- `cargo update -p datafusion-table-providers`
- `cargo check -p data_components --features postgres` ✅
- `cargo check -p connector-postgres` started and compiled through `runtime` dependencies, but hit the 240s local timeout before completion

PR #10601, opened 2026-04-30
fix(iceberg): rebuild storage factory when Hadoop catalog scheme is inferred

phillipleblanc:
## Description

When a Hadoop catalog's warehouse root scheme (e.g. `s3://`) does not match the scheme used in the table metadata locations (e.g. `s3a://`), the existing scheme inference logic in `HadoopCatalogBuilder::inner_build` rebuilds the catalog with the inferred warehouse root. However, since the iceberg-rust v0.9 upgrade (#9917), the `OpenDalStorageFactory::S3` variant validates that paths match its `configured_scheme`. If the factory was constructed with `configured_scheme: "s3"` but the rebuilt catalog tries to operate on `s3a://...` paths, it errors with:

```
DataInvalid => Invalid s3 url: s3a://hadoop/, should start with s3://hadoop/
```

This caused the [`catalogs/iceberg-hadoop` cookbook recipe](https://github.com/spiceai/cookbook/tree/trunk/catalogs/iceberg-hadoop) to fail to register its catalog (the recipe writes table metadata via Spark using `s3a://`).

Reproduction (before this fix), with the cookbook's `iceberg-hadoop` recipe:

```
ERROR runtime::init::catalog: Failed to initialize catalog connector: Cannot setup the catalog hadoop (iceberg) with an invalid configuration. Failed to create Hadoop Catalog for Iceberg with base URI: s3://hadoop/
```

(The actual underlying error was previously hidden because `Error::InvalidConfiguration`'s `Display` impl only shows the `message` field, not the `source`. The full chain is `DataInvalid => Invalid s3 url: s3a://hadoop/, should start with s3://hadoop/`.)

## Fix

- Add `HadoopCatalogBuilder::with_storage_factory_builder(Fn(&str) -> Arc<...>)`, a closure that produces a `StorageFactory` for a given URL scheme. When set, the factory is materialized in `inner_build` from the current warehouse root scheme, which allows scheme inference to rebuild the factory with the inferred scheme on the recursive `inner_build(false)` call.
- Update the iceberg catalog connector (`crates/runtime/src/catalogconnector/iceberg.rs`) to use the new builder closure for the S3 case, so the `CustomAwsCredentialLoader` is captured and the factory is rebuilt with the correct scheme when inferred.
- Add a regression test `get_s3_to_s3a_inferred_hadoop_catalog` that configures the warehouse root as `s3://hadoop/` (while the underlying metadata uses `s3a://hadoop/`) and verifies the catalog builds successfully via scheme inference + factory rebuild.

## Validation

End-to-end with the `catalogs/iceberg-hadoop` cookbook recipe:

```
INFO runtime::init::catalog: Registering catalog 'hadoop' for iceberg
INFO runtime::init::catalog: Registered catalog 'hadoop' with 1 schema and 8 tables
```

TPC-H Q1 against the catalog returns expected results:

```sql
SELECT l_returnflag, l_linestatus, SUM(l_quantity) AS sum_qty
FROM hadoop.tpch.lineitem WHERE l_shipdate <= DATE '1998-09-02'
GROUP BY l_returnflag, l_linestatus
ORDER BY l_returnflag, l_linestatus;
```

| l_returnflag | l_linestatus | sum_qty    |
|--------------|--------------|------------|
| A            | F            | 37734107.0 |
| N            | F            | 991417.0   |
| N            | O            | 74476040.0 |
| R            | F            | 37719753.0 |

`make lint-rust` passes locally.

## Linked Issues

Found while testing for #10547 (v2.0.0-rc.4 endgame).

### DIFF
--- a/crates/data_components/src/iceberg/catalog/hadoop/catalog.rs
+++ b/crates/data_components/src/iceberg/catalog/hadoop/catalog.rs
@@ -41,15 +41,40 @@ pub enum MetadataMode {
     Exact(String),
 }
 
+/// A function that builds a `StorageFactory` for a given URL scheme (e.g. `s3`, `s3a`).
+///
+/// This is used by the `HadoopCatalogBuilder` to allow the storage factory to be
+/// reconstructed when scheme inference detects that the warehouse root scheme does
+/// not match the scheme used in table metadata locations.
+pub type StorageFactoryBuilderFn = Arc<dyn Fn(&str) -> Arc<dyn StorageFactory> + Send + Sync>;
+
 /// Builder for creating a new `HadoopCatalog`
-#[derive(Debug, Default, Clone)]
+#[derive(Default, Clone)]
 pub struct HadoopCatalogBuilder {
     warehouse_root: Option<String>,
     file_io: Option<FileIO>,
     metadata_mode: MetadataMode,
     properties: HashMap<String, String>,
     storage_factory: Option<Arc<dyn StorageFactory>>,
+    storage_factory_builder: Option<StorageFactoryBuilderFn>,
     operator: Option<Operator>,
+}
+
+impl std::fmt::Debug for HadoopCatalogBuilder {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("HadoopCatalogBuilder")
+            .field("warehouse_root", &self.warehouse_root)
+            .field("file_io", &self.file_io)
+            .field("metadata_mode", &self.metadata_mode)
+            .field("properties", &self.properties)
+            .field("storage_factory", &self.storage_factory)
+            .field(
+                "storage_factory_builder",
+                &self.storage_factory_builder.as_ref().map(|_| "<fn>"),
+            )
+            .field("operator", &self.operator)
+            .finish()
+    }
 }
 
 impl HadoopCatalogBuilder {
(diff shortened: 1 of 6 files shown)


sgrebnov (dismissed):

phillipleblanc:
## CI status

All checks related to this change pass. The 2 remaining failures are pre-existing issues on trunk, unrelated to this change:

1. **Test Clickhouse quickstart (linux-x86_64)** — fails consistently with `curl: (22) The requested URL returned error: 404` during `Install Clickhouse`. This is broken external infrastructure (the ClickHouse install URL is returning 404). Same failure on the previously merged PR #10578.

2. **Test Models, AI & Search** — `test_multi_column_w_existing_embedding` insta snapshot assertion fails. Same failure on the previously merged PR #10578.

Reran both jobs to confirm — both still fail with the same errors.

Specifically validated as passing for this change:
- `Rust Lint` ✅
- `Build (release profile)` ✅
- `Build and Test` ✅
- `Data Components Integration Tests` ✅ (includes the new `s3-to-s3a-inferred` Hadoop catalog regression test)
- `Integration Tests (parts 1, 2, 3)` ✅
- `ADBC Integration Tests` ✅
- `AWS SDK Integration Tests` ✅
- All quickstart tests for PostgreSQL / MySQL / DuckDB / S3 etc. ✅
- All acceleration tests (arrow, cayenne, duckdb, sqlite, turso, postgres) ✅

phillipleblanc:
Addressed both Copilot review comments:

1. **Misleading error when builder is set without a scheme** (catalog.rs:180): Added an explicit `DataInvalid` error when `storage_factory_builder` is set but the warehouse root doesn't contain a `://` scheme, so callers see an actionable failure (`"Cannot materialize storage factory: warehouse root '...' does not contain a URL scheme..."`) instead of the generic "Either file_io, storage_factory, or storage_factory_builder must be provided" message.

2. **REST catalog scheme always resolves to "s3"** (iceberg.rs:409): For the REST catalog path, `catalog_id` is the REST endpoint URL, so `s3_scheme_from_url(&catalog_id)` always returned `"s3"` — meaning a REST catalog with `s3a://` warehouse/metadata would hit the same scheme-mismatch failure. Now derives `configured_scheme` from the `warehouse` property when present, falling back to `catalog_id` for backward compatibility.

sgrebnov (approved):

PR #10636, opened 2026-05-01
feat(http): Support dynamic HTTP connector request params from subqueries

sgrebnov:
## 📝 Summary

Adds a new `HttpParamsPushdown` physical optimizer rule with a lazy `HttpWithDeferredParamsExec` execution plan node that enables `IN (SELECT ...)` subqueries against HTTP datasets.

### Problem

HTTP datasets cannot natively handle `IN (SELECT ...)` or `JOIN` queries where the filter values come from another table. Without this optimizer, DataFusion falls back to `HashJoinExec` which fetches the entire HTTP dataset and joins in memory — producing a single HTTP request regardless of how many filter values exist.

### Changes

- **`HttpWithDeferredParamsExec`**: New `ExecutionPlan` node that defers build-side materialization to `execute()` (fully async). At execution time it:
  1. Materializes unique string values from the build side
  2. Deduplicates values, caps at 20,000 unique values
  3. Validates against `max_request_partitions` before creating partitions
  4. Rewrites the `HttpExec` with expanded partitions (cross-product)

- **EXPLAIN / EXPLAIN ANALYZE improvements**:
  - `HttpExec` shows `partitions=deferred` when wrapped by `HttpWithDeferredParamsExec`
  - `HttpWithDeferredParamsExec` displays `deferred_param=<col>` and `build_side=<node>`
  - Execution metrics (`build_side_values`, `http_exec_partitions`) visible in EXPLAIN ANALYZE

### Example

```sql
EXPLAIN ANALYZE
WITH org_headers AS (
  SELECT '{"x-org-id":"' || id || '"}' AS hdr FROM my_orgs LIMIT 20
)
SELECT * FROM orgs_api
WHERE request_path = '/headers'
  AND request_headers IN (SELECT hdr FROM org_headers);
```

```
HttpWithDeferredParamsExec: deferred_param=request_headers, build_side=CoalescePartitionsExec, metrics=[build_side_values=20, http_exec_partitions=20]
  FilterExec: request_path@0 = /headers
    RepartitionExec: partitioning=RoundRobinBatch(16), input_partitions=1
      HttpExec: base_url=https://httpbin.org/, format=auto, partitions=deferred
  CoalescePartitionsExec
    ProjectionExec: expr=[... as hdr]
      DataSourceExec: org_ids.csv, projection=[id], limit=20
```

### Tests

Tests are in progress.

### Notes for reviewers

1. Design: why deferred `HttpWithDeferredParamsExec` instead of eager materialization

An alternative considered was materializing the build-side values eagerly during physical optimization (inside `HttpParamsPushdown::optimize`) and injecting the expanded partitions into `HttpExec` at planning time. This was rejected because it violates DataFusion's architectural separation between planning and execution:

- **`PhysicalOptimizerRule::optimize` is synchronous by design.** DataFusion optimizer rules are pure plan-to-plan transformations that must not perform I/O or execute sub-plans. Materializing the build side requires executing a plan, which is async and would require spawning a blocking execution context
- **No `TaskContext` at planning time.** The execution context (session config, memory pools, runtime environment) is only available at `ExecutionPlan::execute()` time — not during optimization. The build-side plan may depend on registered table providers or memory tracking that only exist at execution time.
- **Cancellation and resource control.** Deferred execution naturally respects the caller's cancellation tokens, memory limits, and task scheduling. Eager materialization during planning would bypass these controls.

1. Intentionally removed `JoinType::Inner` support from `HttpParamsPushdown` — the rewrite replaces the entire `HashJoinExec` with `HttpWithDeferredParamsExec` which only outputs the HTTP-side schema, dropping build-side columns.  This will lead to errors when users would like to do something like this (include subquery columns)

```sql
SELECT h.content, p.path FROM http_api h JOIN params_table p ON h.request_path = p.path
```

`IN (SELECT ...)` always produces semi-joins in DataFusion, so Inner join is unnecessary for this scenario. An explicit error is returned if a user writes a JOIN ... ON request_param_col to guide them toward the supported `IN (SELECT ...) `syntax.

1. Multiple subqueries (i.e. several `IN (SELECT ...)` clauses against the same HTTP dataset) are structurally supported — the optimizer walks the full plan tree and rewrites each qualifying HashJoinExec independently. However, multi-subquery scenarios are not the primary use case and have limited test coverage. Edge cases may require further improvements.


crates/datafusion-optimizer-rules/src/physical_plan/http_subquery_pushdown.rs
```diff
@@ -0,0 +1,515 @@
...
+                }
+                (
+                    Arc::clone(join_exec.right()),
+                    Arc::clone(join_exec.left()),
+                    name,
+                )
+            } else {
+                return Transformed::no(plan);
+            }
+        }
```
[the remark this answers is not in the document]
sgrebnov: Only http_side.schema() is expected to be used, no build side should be used in this scenario, fine to keep as-is for now

crates/runtime/tests/http/mod.rs
```diff
@@ -633,6 +633,229 @@ async fn test_http_dynamic_request_headers() -> Result<(), String> {
         .await
 }
 
+/// Test that `IN (SELECT ...)` subqueries against a real registered table
```
peasee, line 636: nice!

crates/datafusion-optimizer-rules/src/physical_plan/http_subquery_pushdown.rs
```diff
@@ -0,0 +1,863 @@
...
+        write!(
+            f,
+            "HttpWithDeferredParamsExec: deferred_param={}, build_side={}",
+            self.col_name,
+            self.build_side.name()
+        )
+    }
+}
+
+impl ExecutionPlan for HttpWithDeferredParamsExec {
```
peasee, line 338: Deny `missing_trait_methods` and implement the missing methods. We have been bitten multiple times by assuming trait defaults are what we want for custom execution plans.

Ref: https://github.com/spiceai/spiceai/issues/7827

sgrebnov: Great point, added `missing_trait_methods`, thank you!

lukekim (approved):

PR #10650, opened 2026-05-03
Fix DuckLake connector: downcast, module registration, schema discovery, and S3 credentials

sgrebnov:
## 📝 Summary

PR adds the following improvements:

Example benchmark runs:
 - https://github.com/spiceai/spiceai/actions/runs/25277399667/job/74109294510
 - https://github.com/spiceai/spiceai/actions/runs/25277416409/job/74109336082

### 1. Incorrect downcast type (catalog + data connector)

`connect_sync()` returns a `Box<DuckDbConnection>` wrapper, but all DuckLake code was downcasting via `as_any()` to `duckdb::Connection`. Since the concrete type behind the `dyn Any` is `DuckDbConnection`, the downcast always failed with "Failed to get underlying DuckDB connection".

Fix: downcast to `DuckDbConnection` and access the inner connection via its public `conn` field.

### 2. Missing data connector module declaration

`dataconnector/ducklake.rs` existed but was never declared in `dataconnector/mod.rs`, making it dead code. The `register_data_connector!("ducklake", ...)` macro never ran, so `from: ducklake:<table>` datasets failed with "not a valid connector".

Fix: add `#[cfg(feature = "duckdb")] pub mod ducklake;` to `dataconnector/mod.rs`, matching the catalog connector pattern.

### 3. Fix `information_schema` queries for schema/table discovery

The schema and table discovery queries used `FROM "{catalog_name}".information_schema.schemata/tables`, but DuckLake catalogs don't expose `information_schema` under their catalog name. DuckDB surfaces them only via the global `information_schema` or the internal `__ducklake_metadata_<name>.information_schema`.

Fix: query the global `information_schema` (without catalog prefix), filtering with the existing `WHERE catalog_name = ?` parameter.

### 4. Add S3/httpfs credential configuration

DuckLake stores data as Parquet files on S3-compatible storage, but no code existed to install DuckDB's `httpfs` extension or configure S3 credentials. Queries against S3-backed catalogs failed with authentication errors.

Fix: add `DuckLakeS3Params` struct and `configure_duckdb_httpfs()` helper that installs/loads `httpfs` and creates an S3 secret from explicit parameters (`aws_region`, `aws_access_key_id`, `aws_secret_access_key`, `aws_endpoint`, `aws_allow_http`). When no explicit credentials are provided, DuckDB falls back to its built-in `credential_chain` provider (env vars, IAM roles). Wired into both the catalog and data connector paths.

### 5. Add TPC-H query overrides for DuckLake catalog

Catalog connectors require query overrides to use fully-qualified table references (`ducklake.main.<table>` instead of bare `<table>`). Without overrides, all TPC-H benchmark queries fail with table-not-found errors.

Fix: add `ducklake_catalog/` query override directory (28 queries, mirroring `mysql_catalog/` with `ducklake.main.*` table references), register `DucklakeCatalog` variant in `QueryOverrides` enum and testoperator args, and wire it into the dispatch config.

### 6. CI workflow: pin Python duckdb version

The bootstrap script installed the latest `duckdb` (1.5.2), which creates ducklake catalogs in format v0.4+. The runtime uses DuckDB 1.4.4, which only supports ducklake format v0.1–0.3, causing "Only DuckLake versions 0.1, 0.2, 0.3-dev1 and 0.3 are supported".

Fix: pin `duckdb==1.4.4` in the CI bootstrap to match the runtime version.

## 👀 Notes for Reviewers

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10650"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


.github/workflows/testoperator_run_bench.yml
```diff
@@ -235,12 +236,18 @@ jobs:
...
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
+        with:
+          python-version: '3.12'
+
       - name: Bootstrap DuckLake TPCH SF1 catalog
-        if: ${{ contains(github.event.inputs.spicepod_path, 'ducklake[postgres+s3]') }}
+        if: ${{ contains(github.event.inputs.spicepod_path, 'ducklake') }}
         run: |
-          python3 -m pip install --upgrade duckdb minio
+          python -m pip install --upgrade 'duckdb==1.4.4' minio
```
sgrebnov, line 248: Spice  does NOT currently support the latest catalog version, requires DuckDB version upgrade

lukekim (approved):

Issue #10649, opened 2026-05-03
Cayenne: snapshot bootstrap rejects archive when local metastore is initialized first


phillipleblanc:
## Summary

When a Cayenne-accelerated dataset uses `refresh_mode: snapshot` (or `bootstrap_on_failure_behavior` triggers archive extraction), the snapshot extraction step fails with:

```
Failed to extract snapshot archive at /tmp/snapshot_download_<dataset>_<ts>_<uuid>.tar:
Data integrity violation: existing file <metadata_dir>/cayenne.db-shm has checksum X,
expected Y. This may indicate data corruption or concurrent modification.
```

The error originates in `crates/runtime-acceleration/src/snapshot/mod.rs::download_to_directories`, which validates that any pre-existing files at the extraction destination match the archive's recorded checksums.

## Repro

Run the Docker-orchestrated Cayenne integration test from PR #10565:

```
cargo test -p runtime --features 'cayenne,snapshots'   --test integration_snapshot_refresh   -- snapshot_refresh::cayenne::snapshot_refresh_cayenne_bootstrap_then_refresh
```

The reader container starts the runtime, which eagerly creates an empty `cayenne.db`, `cayenne.db-shm`, and `cayenne.db-wal` in the metadata directory. The snapshot bootstrap then downloads the archive, attempts extraction, and bails because the local SQLite journal sidecars don't match the archive's content.

## Why this is its own issue

This is independent of #10642 (Cayenne catalog stores absolute paths) and #10643 (SQLite WAL not flushed before snapshot). The race here is between the runtime's eager Cayenne metastore initialization and the snapshot bootstrap's archive extraction — both target the same directory.

## Suggested fix shape

One of:

1. Skip the integrity check for SQLite journal sidecars (`*.db-shm`, `*.db-wal`) — they are inherently transient.
2. Have the snapshot extraction `fs::remove_file` mismatched destinations rather than erroring, when the destination is a known-transient sidecar.
3. Defer Cayenne metastore initialization until after the bootstrap snapshot has been applied (changes the runtime startup ordering).

Option 1 is the smallest and most defensible: SQLite journal files are not part of the logical database state and can never meaningfully match an archived copy from another node.

## Impact

- Blocks the Cayenne integration test in PR #10565 (currently `#[ignore]`'d alongside SQLite/Turso).
[...]

Issue #10642, opened 2026-05-03
Cayenne snapshots are not relocatable across data paths (catalog stores absolute paths)


phillipleblanc:
## Summary

Cayenne acceleration snapshots store **absolute** filesystem paths in the catalog metastore (`cayenne_table.path`, `cayenne_partition.path`, `cayenne_delete_file.path`). When a snapshot is bootstrapped on a different host with a different `cayenne_file_path` / data directory, the restored catalog still references the original writer's paths and the reader returns 0 rows.

This works fine in production today because all pods/instances are configured with the same `spice_data_base_path()` (e.g. `/spice/data`) so paths match. It only breaks when consumer paths differ from producer paths.

## Repro

In the integration test suite, configure a writer with `cayenne_file_path = /tmp/A/cayenne` and a reader with `cayenne_file_path = /tmp/B/cayenne` and have the reader bootstrap from a snapshot uploaded by the writer. The reader's catalog will contain `/tmp/A/cayenne/...` paths that don't exist in its environment.

Concrete evidence (commit `f3132221` on PR #10565):

```
crates/cayenne/src/cayenne_catalog.rs:1934    path_is_relative: false,
crates/cayenne/src/cayenne_catalog.rs:2024    path_is_relative: false,
crates/cayenne/src/cayenne_catalog.rs:2113    path_is_relative: false,
crates/cayenne/src/cayenne_catalog.rs:2192    path_is_relative: false,
crates/cayenne/src/cayenne_catalog.rs:2289    path_is_relative: false,
```

Every partition/table/delete-file insert hard-codes `path_is_relative: false`. The schema already has the `path_is_relative` column — it's just never set to `true`.

## Proposed fix

1. When writing partition/table/delete-file rows whose `path` is under the configured data directory, store the path **relative to the data directory** with `path_is_relative = true`.
2. On read, when `path_is_relative = true`, resolve against the current data directory.
3. Provide a one-shot migration that rewrites existing catalog rows whose absolute paths are under the current data directory into relative form.
4. (Optional) In the snapshot bootstrap path, run the migration after extracting the catalog so older absolute-path snapshots become relocatable.

## Impact

- Without this fix, Cayenne snapshots only work when consumer and producer share an identical absolute data path layout. This is fine for our current Kubernetes deployment model but blocks legitimate use cases (cross-cluster replication, local restore from prod snapshot, multi-tenant isolation, etc.).
[...]

PR #10651, opened 2026-05-03
feat: refresh_mode: snapshot + SQLite/Turso WAL flush + Cayenne metastore slice (stacked)

### LINKED ISSUE #10643: SQLite acceleration snapshots only capture WAL header (4096 bytes), missing actual data
## Summary

SQLite acceleration snapshots upload only a 4096-byte SQLite header file with no actual data. The cause: in WAL journaling mode (the default for SQLite acceleration), recently-written data lives in the `-wal` sidecar file until a checkpoint flushes it to the main DB file. The snapshot pipeline performs a plain `fs::copy` of the `.sqlite` file only, which captures just the header.

The same root-cause likely affects **Turso** (libsql 0.5.x — pure-Rust MVCC layout, file format TBD) and is the reason both engines' `refresh_mode: snapshot` integration tests are currently `#[ignore]`d.

## Repro

Background-watch the writer's SQLite file size during any snapshot-refresh integration test (PR #10565):

```
writer.sqlite      4096 bytes (constant)
writer.sqlite-wal  37KB → 45KB → 53KB → 61KB → 74KB (grows)
```

Snapshot artifact uploaded to S3 is exactly 4096 bytes. Reader bootstrap fails with "missing a schema in its checkpoint" because the restored DB is empty past the header.

The existing `snapshot_int_test3_sqlite` test passes only because it silently falls back to a federated refresh on bootstrap failure. `refresh_mode: snapshot` has no fallback (by design — the whole point is to never query the source) so it surfaces the bug.

## Why DuckDB doesn't have this

DuckDB is configured with `PRAGMA enable_checkpoint_on_shutdown` and auto-checkpoints on commit, so the main `.duckdb` file always reflects current data. A plain `fs::copy` is sufficient.

## Why Cayenne doesn't have this

Cayenne's catalog SQLite is checkpointed in `crates/cayenne/src/metastore/sqlite.rs::shutdown()` (`PRAGMA wal_checkpoint(TRUNCATE)`) before being read for snapshot purposes. The acceleration SQLite path has no equivalent.

## Proposed fix

Add a SQLite-specific `SnapshotEngine` impl in `crates/runtime-acceleration/src/snapshot/engine/` that overrides the default `fs::copy` step. Two clean options:

1. **`VACUUM INTO '<temp>'`** — produces a complete, page-compacted standalone DB regardless of WAL state. Cleanest semantics, slightly more I/O than a copy.
2. **`PRAGMA wal_checkpoint(TRUNCATE)` then `fs::copy`** — cheaper, but requires temporarily quiescing writers (or accepting that the snapshot is taken at an arbitrary post-checkpoint instant).

Recommendation: **VACUUM INTO**, by analogy with how SQLite itself recommends backing up live databases.

Then do the same for Turso once its file format is verified.

## Where to fix

- Trait: `crates/runtime-acceleration/src/snapshot/engine.rs` (`SnapshotEngine`)
- New impl: `crates/runtime-acceleration/src/snapshot/engine/sqlite.rs`
- Wire-up: `create_snapshot_engine` in `crates/runtime-acceleration/src/snapshot/engine.rs`, branch on `AccelerationEngine::Sqlite`

## Impact

- Without this fix, SQLite snapshots are silently broken — refresh-mode `full`/`append` happen to recover via federated fallback; `refresh_mode: snapshot` (PR #10565) and any other snapshot-only consumer will fail.
- Blocks SQLite/Turso `refresh_mode: snapshot` integration tests in PR #10565 (currently `#[ignore]`d, referencing this issue).

## References

- PR #10565 (`refresh_mode: snapshot` implementation; SQLite/Turso disabled with `#[ignore]` pointing at this issue)
- Existing checkpoint pattern: `crates/cayenne/src/metastore/sqlite.rs:664-676`

### LINKED ISSUE #10658: Snapshot bootstrap: Cayenne snapshot is missing schema in checkpoint after slice import
## Problem

After PR #10651 wires `CayenneSnapshotEngine` into the snapshot pipeline, the on-disk archive ships `metadata/<dataset>.slice.json` instead of the raw `cayenne.db` file.

On extract, `SnapshotManager::download_to_directories` calls `finalize_directory_snapshot` which imports the slice into a fresh local `cayenne.db` via the metastore catalog. After that, `download_latest_snapshot` calls the spice_sys checkpointer's `get_schema()` — which reads the `_dataset_checkpoint` table inside `cayenne.db`.

But the slice format only carries the cayenne-domain tables (`cayenne_table`, `cayenne_partition`, `cayenne_delete_file`). The spice_sys `_dataset_checkpoint` table is **not** in the slice, so it's empty in the freshly-built `cayenne.db` and `get_schema()` returns `None`. The bootstrap then errors with:

```
Snapshot snapshot_refresh_cayenne/.../trips_*.cayenne is missing a schema in its checkpoint
```

## Effect

`refresh_mode: snapshot` for Cayenne fails to bootstrap. The integration test `snapshot_refresh::cayenne::snapshot_refresh_cayenne_bootstrap_then_refresh` fails on every run.

Workaround in #10651: the test is `#[ignore]`d with this issue link.

## Fix options

1. Include the spice_sys `_dataset_checkpoint` row(s) for the dataset in `DatasetMetastoreSlice`. Requires plumbing because spice_sys is in `runtime` and `DatasetMetastoreSlice` is in `cayenne`.
2. After `finalize_directory_snapshot` imports the slice, materialize a checkpoint row using the metadata-recorded schema (we'd need to surface the schema into the engine post-extract hook).
3. Special-case Cayenne in `download_latest_snapshot`: if the engine is Cayenne, derive the schema from the metadata.json instead of asking the checkpointer.

(1) is the cleanest long-term fix.

## Repro
- Stack: PR #10651
- Test: `cargo test -p runtime --features cayenne --test integration_snapshot_refresh -- snapshot_refresh::cayenne` against real S3 (`AWS_PROFILE=sandbox-admin`)

phillipleblanc:
## Summary

Adds `refresh_mode: snapshot` to Spice acceleration, plus the underlying snapshot-engine fixes that the new mode requires for SQLite and Cayenne. Structured as 5 stacked commits for easy commit-by-commit review.

## Stacked commits

1. **`fix(snapshot): flush SQLite/Turso WAL before snapshotting`** — closes #10643 for SQLite
   * Adds `SnapshotEngine::checkpoint_live` (default no-op), called by `SnapshotManager::create_file_snapshot` *before* `fs::copy` while the accelerator write lock is held.
   * `SqliteSnapshotEngine` runs `PRAGMA wal_checkpoint(TRUNCATE)` on the live file via a short-lived rusqlite connection, reads the returned `(busy, log, checkpointed)` row, and surfaces a `CheckpointIncomplete` error if not all frames flushed (rather than silently snapshotting a partial DB). Then switches the *copied* file to `journal_mode=DELETE` as defense-in-depth so the uploaded file has no `-wal`/`-shm` sidecars.
   * `TursoSnapshotEngine` is a no-op for now: rusqlite refuses libsql files with `"file is not a database"`, so the WAL flush won't succeed without a turso/libsql-native checkpoint path. Tracked in #10657. Turso behavior is unchanged from main.

2. **`feat(acceleration): add refresh_mode: snapshot for snapshot-only refreshes`**
   * `RefreshMode::Snapshot` accepted in spicepod (rejected by validator unless `acceleration.snapshots` is bootstrap-capable).
   * Refresh task polls the snapshot store on a configurable interval (default 60s) and swaps in newer snapshots via `Accelerator::reload_from_snapshot` (DuckDB/SQLite/Turso impls evict pool connections and reopen at the same primary path).
   * `download_if_newer` validates schema **before** download. If `validate_schema` is provided, the remote metadata must expose a parseable schema for the dataset — missing/malformed schema is treated as `SchemaMismatch` rather than silently bypassing validation.
   * `INSERT INTO <dataset>` is rejected with a clear error.
   * Live federated query path serves the prior inode until swap completes via `SwappableTableProvider`.
   * Atomic file write on download (temp + fsync + rename + parent fsync). The Windows `AlreadyExists` rename fallback now uses a swap-via-sidecar dance (rename existing target to `.old.<pid>` atomically, rename temp into place atomically, best-effort delete sidecar) so `local_path` is never a missing entry — important for `refresh_mode: snapshot` where the accelerator pool may still be holding readers.
   * Integration tests under `tests/snapshot_refresh/` for **DuckDB** and **SQLite** (both enabled). Turso is `#[ignore]`d pending #10657. Cayenne is enabled by commit 5. New CI job `Snapshot Refresh Mode Integration Tests`.
   * Pulls in pool-eviction APIs from the now-merged datafusion-contrib/datafusion-table-providers#635 (`invalidate_instance` / `invalidate_file_instance`).

3. **`feat(cayenne): add per-dataset metastore export/import (snapshot foundation)`**
   * New `crates/cayenne/src/metastore/snapshot.rs`: versioned-JSON `DatasetMetastoreSlice` with `format_version: 1`, plus `export_dataset` / `import_dataset` operating on any `MetastoreBackend`.
   * Path columns (`cayenne_table.path`, `cayenne_delete_file.path`, `cayenne_partition.path`) are rewritten relative to a configurable anchor on export, and re-anchored on import. When `make_relative` can't strip the anchor (path lives outside `data_dir_anchor`), the slice keeps `path_is_relative = false` so it stays internally consistent.
   * Import is atomic: single transaction, `DELETE FROM cayenne_table WHERE table_name = ?` first (FK `ON DELETE CASCADE` removes dependents), then re-insert.
   * `MetastoreRow::get_value(idx) -> MetastoreValue` added to the trait so generic export logic can serialize columns positionally.
   * 5 unit tests cover round-trip + path relocation, wholesale replace, multi-dataset isolation, format-version rejection, and JSON round-trip.

4. **`feat(cayenne): wire CayenneSnapshotEngine into snapshot pipeline`** — partial #10642, partial #10649
   * Extends `SnapshotEngine` trait with directory-snapshot hooks (`prepare_directory_snapshot` returning a skip list + extras, `finalize_directory_snapshot` for post-extract work). DuckDB/SQLite/Turso get default no-op behavior.
   * `directory_archive` gets `archive_directories_with_plan(dirs, writer, skip, extras)`. Engine-agnostic — the skip predicate is just a `HashSet`.
   * `SnapshotManager` gets `with_snapshot_engine(self, Arc<dyn SnapshotEngine>)`; calls the new hooks during create/extract.
   * `DataAccelerator::snapshot_engine_for_source(&dyn AccelerationSource) -> Option<Arc<dyn SnapshotEngine>>` added with default `None`. Cayenne overrides it to construct a `CayenneSnapshotEngine`. Threaded through both `build_snapshot_creation_config` (snapshot upload) and `build_snapshot_refresh_state` (refresh-mode reload).
   * `MetadataCatalog` gains `export_dataset_slice` / `import_dataset_slice`; `CayenneCatalog` overrides both with dispatch to the underlying SQLite/libsql metastore.
   * New `crates/runtime/src/dataaccelerator/cayenne/snapshot_engine.rs` — `CayenneSnapshotEngine`. On create: skips `cayenne.db` / `-wal` / `-shm`, writes the per-dataset slice JSON to `metadata/<dataset>.slice.json` (the `metadata/` prefix matches `AccelerationLayout::cayenne` so it lands at `<metadata_dir>/<dataset>.slice.json` on extract). On extract: reads the slice back and atomically imports it into the local metastore. Errors from `try_exists` are propagated rather than silently treated as "missing slice".
   * Lifts `validate_cayenne_snapshot_consistency`'s `SharedAcceleration` restriction: with per-dataset slices threaded through both creation paths, multiple Cayenne datasets can safely share a metastore directory.

5. **`fix(snapshot): bootstrap dataset checkpoint from snapshot metadata`** — closes #10658
   * When `Checkpointer::get_schema()` returns `None` post-extract (the steady state for Cayenne, where the archive ships the per-dataset slice JSON instead of the raw `cayenne.db` and so the spice_sys `_dataset_checkpoint` table is empty in the freshly-imported DB), the metadata-recorded schema (already validated to round-trip via `to_schema_ref()`) is materialized into the checkpoint via `Checkpointer::checkpoint(metadata_schema, None)`.
   * For DuckDB / SQLite / Turso this branch is unreachable in steady state (their archives ship `_dataset_checkpoint` already); on a corrupted snapshot the self-heal is harmless because the metadata schema is the same one we'd otherwise have validated against.
   * New `CheckpointerBootstrap` error variant in `SnapshotDownloadError` for clear diagnostics if the bootstrap write itself fails.
   * Re-enables `snapshot_refresh::cayenne::snapshot_refresh_cayenne_bootstrap_then_refresh` in CI.

## Known limitations / follow-ups

* **#10657 — Turso WAL flush via rusqlite fails (`"file is not a database"`).** `SqliteSnapshotEngine` can't open libsql primary files. `TursoSnapshotEngine` is therefore a no-op until a libsql-native checkpoint path is wired in. Turso snapshot creation behavior is unchanged from main; `refresh_mode: snapshot` for Turso is gated on this fix and `tests/snapshot_refresh/turso.rs` is `#[ignore]`d with a link.

## Issue tracking

* spiceai/spiceai#10643 — closed by commit 1 (SQLite); Turso portion gated on #10657
* spiceai/spiceai#10642 — closed by commits 3+4+5 (slice format is portable; Cayenne snapshot refresh enabled end-to-end)
* spiceai/spiceai#10649 — closed by commits 3+4+5 (no more `cayenne.db` in archive ⇒ no init/checksum race; integration test enabled by default)
* spiceai/spiceai#10658 — closed by commit 5

## Test coverage

* `runtime-acceleration` lib tests: 91 passing
* `cayenne` lib tests: 108 passing (incl. 5 metastore-slice tests)
* `runtime` lib tests: incl. 3 new `CayenneSnapshotEngine` unit tests
* `Snapshot Refresh Mode Integration Tests` CI job runs DuckDB + SQLite + **Cayenne** end-to-end against real S3


phillipleblanc:
Latest stack push (commits `7a9a3333` → `9eb40616` → `0c4fd16c` → `6a468d31`) addresses the open Copilot comments and switches to the official upstream rev for `datafusion-table-providers`:

**Upstream dependency (Commit 2):**
- Bumped `datafusion-table-providers` to `97ecd0059bd49297b956b6dd51c7047547cc97e0` (datafusion-contrib/datafusion-table-providers#635 is now merged); the temporary fork-branch pin is gone.

**Cayenne snapshot engine wiring (Commit 4 — fixes the integration test red):**
- Added `DataAccelerator::snapshot_engine_for_source(&dyn AccelerationSource) -> Option<Arc<dyn SnapshotEngine>>` (default `None`); Cayenne overrides it to construct a `CayenneSnapshotEngine`.
- Threaded the override through `build_snapshot_creation_config` (snapshot upload pipeline) **and** `build_snapshot_refresh_state` (refresh-mode reload pipeline). This is what was missing — the writer was previously falling back to the default engine and shipping `cayenne.db-wal` in the archive, which is exactly the failure CI was hitting.
- Fixed `slice_archive_path`: now uses the `metadata/` prefix (matches `AccelerationLayout::cayenne`) instead of `metastore/`, so the slice JSON lands at `<metadata_dir>/<dataset>.slice.json` on extract instead of an unmapped fallback location.
- `finalize_directory_snapshot` no longer takes the `extras` parameter — it can't be meaningfully populated across the upload→download boundary, and the engine looks slices up by well-known archive path anyway.
- Removed `#[allow(dead_code)]` `add_directory_to_archive` wrapper in `directory_archive.rs` — `archive_directories` now routes through `_filtered` directly.
- Lifted `validate_cayenne_snapshot_consistency::SharedAcceleration`; updated the comment to reflect that the engine override is wired in both creation paths.

**SQLite WAL flush (Commit 1):**
- Split `CheckpointSnafu` so `journal_mode=DELETE` failures get their own `JournalMode` error variant with an accurate message.

**Refresh-mode core (Commit 2):**
- `download_if_newer` now treats missing-dataset / missing-`current_schema` / unparseable schema as `SchemaMismatch` when a validator is provided, instead of silently bypassing validation and overwriting the local file.

**Cayenne metastore slice (Commit 3):**
- `export_dataset` no longer hard-codes `path_is_relative = true`. When `make_relative()` can't strip the anchor (path lives outside `data_dir_anchor`), the slice keeps `path_is_relative = false` so it stays internally consistent. Import side already round-trips correctly via `make_absolute`.

Local verification: 91 `runtime-acceleration` lib tests, 108 `cayenne` lib tests, 3 new `CayenneSnapshotEngine` tests, all green; `cargo fmt --check` clean.

phillipleblanc:
Pushed `fffc037b` addressing the latest Copilot batch:

**`engine/sqlite.rs:110` (`checkpoint_live` partial-success):** `wal_checkpoint(TRUNCATE)` now reads the returned `(busy, log_frames, checkpointed_frames)` row and surfaces a new `SqliteSnapshotError::CheckpointIncomplete` if `busy != 0` or `checkpointed_frames < log_frames`. Previously these partial-success cases would silently leak post-checkpoint writes past the snapshot copy.

**`snapshot/mod.rs:2014` (`AlreadyExists` fallback atomicity):** rewrote the Windows fallback as a swap-via-sidecar dance — rename the existing target to `<local>.old.<pid>` (atomic), rename the temp into place (atomic), then best-effort delete the sidecar. At every instant `local_path` points to either the old file or the new one, never a missing entry. On retry-rename failure we restore the original via `rename(sidecar, local_path)` so the dataset is never left dangling. Important for `refresh_mode: snapshot` because the accelerator's pool may still be holding readers open against `local_path` in the gap before `reload_from_snapshot` evicts them.

The `cayenne/snapshot.rs:198` (`#[allow(dead_code)]` on `table_id_column_index`) comment was already addressed in the previous push: it's now `#[expect(dead_code, reason = "retained for future re-keying on import; see doc above")]`, so the lint is acknowledged with an explicit reason rather than blanket-allowed.

Local: 91 `runtime-acceleration` lib tests, 108 `cayenne` lib tests, 3 `CayenneSnapshotEngine` tests all pass. `cargo fmt --check` and `make lint-rust` (lib+bins phase) clean.

phillipleblanc:
The new `Snapshot Refresh Mode Integration Tests` job surfaced two real, **non-flaky** issues that the in-process tests didn't catch. Both are now `#[ignore]`d with explicit issue links rather than papered over:

1. **Turso WAL flush via rusqlite — #10657**
   `SqliteSnapshotEngine::checkpoint_live` opens the live accelerator file with rusqlite; against Turso's libsql on-disk format that fails with `"file is not a database"`. My earlier assumption that libsql is byte-compatible with classic SQLite for read-only opens was wrong. The hook is therefore a no-op for Turso today and `snapshot_refresh::turso::*` is `#[ignore]`d pending a turso-native checkpoint path.

2. **Cayenne checkpoint schema handoff — #10658**
   With `CayenneSnapshotEngine` wired in (Commit 4), the snapshot tar ships `metadata/<dataset>.slice.json` instead of the raw `cayenne.db`. The slice import correctly rebuilds the cayenne-domain tables, but does **not** populate the spice_sys `_dataset_checkpoint` table that `download_latest_snapshot` queries via `Checkpointer::get_schema`. Bootstrap then errors with `MissingSchema`. `snapshot_refresh::cayenne::*` is `#[ignore]`d pending the schema-handoff fix.

The main feature plumbing in this PR (per-dataset slice format, `CayenneSnapshotEngine`, snapshot-engine pipeline wiring, multi-dataset shared metastore validation lift) is unchanged and works end-to-end *up to* the schema handoff — the issues describe the remaining steps to enable each test by default.

DuckDB and SQLite snapshot-refresh integration tests continue to pass and cover the core `refresh_mode: snapshot` contract.

Pushed `1960e79e`. Other changes from this round (WAL checkpoint partial-success surfacing, atomic swap-via-sidecar fallback, `#[expect(dead_code)]` cleanups, doc backticks) are unchanged.

phillipleblanc:
Pushed `a5dfc372` addressing this round of Copilot:

**`engine/turso.rs:25` and `engine.rs:216` (Turso WAL flush + create_snapshot_engine)** — Copilot is right that delegating Turso to `SqliteSnapshotEngine` regresses behavior on non-rusqlite-compatible libsql files. `TursoSnapshotEngine` is now a **no-op**: `checkpoint_live` and `prepare_for_upload` both return `Ok` without touching the file. The struct stays so `create_snapshot_engine` keeps a stable, Turso-specific type for symmetry with the other engines. Module doc updated to reflect the actual status (defers to a libsql-native implementation, tracked in #10657). Net effect: Turso snapshot creation is back to the same behavior as main — `fs::copy` of the live file as-is, with the same WAL-loss caveat #10643 originally documented.

**`tests/snapshot_refresh/turso.rs:30` and `tests/snapshot_refresh/cayenne.rs:37` (PR description mismatch)** — Updated the PR description. The "What's enabled" claim now reads:
> Integration tests under `tests/snapshot_refresh/` for **DuckDB** and **SQLite** (both enabled). Turso and Cayenne are `#[ignore]`d pending follow-up work; see "Known limitations" below.

Both follow-ups (#10657, #10658) are linked in the PR description with concrete fix paths.

**`dataaccelerator/cayenne/snapshot_engine.rs:270` (try_exists swallows errors)** — Replaced the `unwrap_or(false)` with an explicit `match`: real I/O errors from `fs::try_exists` (permissions, transient failure) now surface as `CayenneSnapshotEngine: failed to stat candidate slice path …: <err>` instead of being collapsed into a misleading `MissingSlice`. Only `Ok(false)` advances to the next candidate.

Local: 91 `runtime-acceleration` lib tests, 3 `CayenneSnapshotEngine` tests, `cargo fmt --check` and `make lint-rust` (lib+bins) all clean.

phillipleblanc:
**Closes #10658.** Added Commit 5 (`fix(snapshot): bootstrap dataset checkpoint from snapshot metadata`) and re-enabled the Cayenne `snapshot_refresh` integration test.

The blocker was that `download_latest_snapshot` requires `Checkpointer::get_schema()` to return `Some` post-extract, and the Cayenne archive (post-Commit 4) ships the per-dataset metastore slice JSON instead of the raw `cayenne.db`, so the spice_sys `_dataset_checkpoint` table inside the freshly-imported DB is empty.

The fix: when `get_schema()` returns `None`, materialize the metadata-recorded schema (already validated to round-trip via `to_schema_ref()`) into the local checkpoint via `Checkpointer::checkpoint(metadata_schema, None)`. This is generically useful — for DuckDB / SQLite / Turso the branch is unreachable in steady state (their archives ship `_dataset_checkpoint` already), and on a corrupted snapshot the self-heal is harmless because we trust the same metadata schema we'd otherwise be validating against.

Also added a dedicated `CheckpointerBootstrap` error variant for clear diagnostics if the bootstrap write itself fails.

Pushed `86865151` to `phillip/cayenne-metastore-slice` (PR head). 91 lib tests pass, full lint is clean. PR description updated to reflect the 5-commit stack.

lukekim (approved):

PR #10653, opened 2026-05-03
feat(http): per-user caching for HTTP datasets via sensitive headers and accelerator header keys

phillipleblanc:
## Summary

Adds two opt-in parameters that together enable per-user (per-bearer-token) caching of HTTP-backed datasets without leaking one user's data to another, with sensitive header values redacted before they ever land in any cache key or persisted column.

1. **`datasets[].params.request_headers_sensitive`** (comma-separated header names) — on the `https` connector. Two effects:
   - Lifts the existing *"`authorization` cannot be in `request_header_allowlist` when HTTP authentication is configured"* guard for headers explicitly marked sensitive. This allows per-request bearer tokens (carried in `request_headers` filters / the `request_headers` schema column) to coexist with dataset-level HTTP authentication.
   - **Replaces those header values with a deterministic one-way hash** (sentinel-prefixed: `spice-redacted-sha256:<base64url>`) before they are written into the HTTP provider's in-memory cache key or the `request_headers` row column persisted by the accelerator. The raw value is still forwarded upstream verbatim for authentication; **only the redacted form is ever stored**. Determinism of the hash is required: two requests from the same end client (same bearer token) must produce the same cache key and therefore share a cached response.

2. **`datasets[].acceleration.params.caching_allowed_request_headers`** (comma-separated header names) — on the acceleration block. When non-empty, the caching accelerator's row-level cache key extraction (`extract_filters_from_row`) includes the `request_headers` column alongside `request_path` / `request_query` / `request_body`. **Rows whose `request_headers` value carries the redacted sentinel are skipped during background stale-row refresh by design** — the upstream call would require the original token, which is one-way hashed and not recoverable. Those rows are re-populated lazily on the next live request with the matching token.

Both parameters default to empty / off, so this change is fully backward compatible.

## Behavior

- With neither param set: identical behavior to today. `authorization` is still blocked from `request_header_allowlist` when HTTP auth is configured, and the caching accelerator continues to key only on path/query/body.
- With only `request_headers_sensitive` set (and `authorization` listed in both `request_header_allowlist` and `request_headers_sensitive`): the dataset can accept per-request `authorization` headers in `request_headers` filters even when dataset-level HTTP auth is configured. The HTTP provider's in-memory cache keys on the redacted hash, so per-user caching works end-to-end without persisting raw tokens.
- With `caching_allowed_request_headers` additionally set on the accelerator: the caching accelerator's stale-row refresh path also keys on the (redacted) request headers, so per-user entries refresh correctly when re-requested. Rows whose stored `request_headers` value contains the redaction sentinel are skipped from background refresh and refresh lazily on the next live request.

## Example

```yaml
datasets:
  - from: https://api.example.com
    name: per_user_endpoint
    params:
      http_auth: bearer
      http_bearer_token: ${ env:SERVICE_TOKEN }
      request_header_filters: enabled
      request_header_allowlist: authorization
      request_headers_sensitive: authorization
    acceleration:
      enabled: true
      refresh_mode: caching
      params:
        caching_allowed_request_headers: authorization
```

## Code touchpoints

- `crates/data_components/src/http/provider.rs`
  - `RequestFilterOptions::sensitive_headers` and `with_sensitive_headers(...)` builder.
  - `enable_header_filters` and `parse_request_headers` lift the AUTHORIZATION-with-auth guard for sensitive headers only.
  - New `hash_sensitive_value(&str) -> String` and `SENSITIVE_HASH_PREFIX` constant — deterministic SHA-256 + URL-safe base64, sentinel-prefixed.
  - New `request_headers_contain_redacted_value(&str) -> bool` — used by the accelerator to detect non-refreshable rows.
  - `RequestFilterOptions::redact_sensitive_request_headers(&str) -> String` — replaces only listed-sensitive header values with the hash; preserves non-sensitive values.
  - `HttpTableProvider::get_cache_key_with_redacted_headers(...)` and `redact_request_headers_for_storage(...)` — used by `cache_response`, `get_response`, and `create_batch_from_rows` so the in-memory cache key and the persisted `request_headers` row column always carry the redacted form.
- `crates/runtime/src/dataconnector/https.rs`
  - Parses the new `request_headers_sensitive` dataset param and threads it into the provider before `enable_header_filters` is called.
- `crates/runtime/src/component/dataset/acceleration.rs`
  - `Acceleration.caching_allowed_request_headers` plus `parse_caching_allowed_request_headers` (lowercases, trims, drops empties).
- `crates/runtime/src/accelerated_table/{mod.rs,refresh.rs,refresh_task.rs,refresh_task_runner.rs,caching.rs}`
  - Plumbs `caching_allowed_request_headers` from `Acceleration` → `AcceleratedTable::Builder` → `Refresher` → `RefreshTaskRunner` → `RefreshTask`.
  - `CacheRefreshHelper::extract_filters_from_row` now returns `Option<Vec<Expr>>`; `None` signals a row that cannot be background-refreshed (its `request_headers` carries the redaction sentinel). `extract_unique_filter_sets` skips such rows.
- `crates/runtime/src/datafusion/mod.rs`
  - Wires `acceleration_settings.caching_allowed_request_headers` into the `AcceleratedTable` builder alongside the other `caching_*` params.

## Tests

New unit tests:
- `data_components::http::provider::tests`
  - `test_sensitive_authorization_allowlist_with_auth` — `authorization` is allowlistable with HTTP auth when also marked sensitive.
  - `test_sensitive_authorization_filter_with_auth` — `request_headers` filters carrying `authorization` are accepted under HTTP auth when sensitive.
  - `test_sensitive_headers_rejects_invalid_name` — malformed names rejected early.
  - `test_hash_sensitive_value_is_deterministic_and_distinct` — same input → same hash, different inputs → different hashes, raw value never appears in output, sentinel prefix present, `request_headers_contain_redacted_value` correctly distinguishes redacted vs raw.
  - `test_redact_sensitive_request_headers_replaces_only_listed_headers` — only sensitive header values are hashed; non-sensitive values pass through; redaction is deterministic.
  - `test_redact_sensitive_request_headers_noop_when_unconfigured` — no sensitive headers → no-op.
- `runtime::accelerated_table::caching::tests`
  - `test_extract_filters_from_row_includes_request_headers_when_allowed` — `request_headers` is included in extracted filters only when in the allowlist.
  - `test_extract_filters_from_row_skips_redacted_request_headers` — rows with redacted sensitive header values return `Ok(None)` from `extract_filters_from_row` and are skipped by `extract_unique_filter_sets`; non-redacted rows still refresh.
- `runtime::component::dataset::acceleration::tests::test_parse_caching_allowed_request_headers` — comma-split, lowercased, trimmed, consumed from the params map.

All existing tests pass:
- `cargo test -p data_components --lib http::provider::tests` → 139 passed.
- `cargo test -p runtime --lib accelerated_table::caching::tests` → 33 passed.
- `cargo test -p runtime --lib dataconnector::https` → 16 passed.
- `cargo test -p runtime --lib component::dataset::acceleration` → 3 passed.
- `cargo clippy -p data_components -p runtime --tests` → no new warnings.

## Compatibility

Fully backward compatible. Both parameters default to empty; without them the existing AUTHORIZATION guard and the existing path/query/body-only accelerator cache key behavior are preserved exactly.

## Out of scope / follow-ups

- User-facing docs updates (the `docs/examples/http_refresh_sql_example.md` example is unchanged; new params are additive).

### DIFF
--- a/crates/data_components/src/http/provider.rs
+++ b/crates/data_components/src/http/provider.rs
@@ -330,6 +330,15 @@ struct RequestFilterOptions {
     max_body_bytes: usize,
     max_headers_length: usize,
     allowed_headers: HashSet<HeaderName>,
+    /// Header names whose values are treated as sensitive (e.g. bearer tokens).
+    /// Listing a header here permits it in `allowed_headers` even when HTTP
+    /// authentication is configured, causes its value to be redacted in logs and
+    /// error messages, and causes its value to be replaced with a deterministic
+    /// hash before being written into HTTP cache keys or the `request_headers`
+    /// row column persisted by the accelerator. The raw value is still forwarded
+    /// upstream verbatim for authentication.
+    /// See `request_headers_sensitive` dataset parameter.
+    sensitive_headers: HashSet<HeaderName>,
 }
 
 impl Default for RequestFilterOptions {
@@ -340,6 +349,7 @@ impl Default for RequestFilterOptions {
             max_body_bytes: DEFAULT_MAX_BODY_BYTES,
             max_headers_length: DEFAULT_MAX_HEADERS_LENGTH,
             allowed_headers: HashSet::new(),
+            sensitive_headers: HashSet::new(),
         }
     }
 }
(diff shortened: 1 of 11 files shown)


crates/data_components/src/http/provider.rs
```diff
@@ -330,6 +330,15 @@ struct RequestFilterOptions {
     max_body_bytes: usize,
```
lukekim, line 330: We should reconciled this with the databricks connector and code re-use what we can across them.

PR #10667, opened 2026-05-04
fix: use InsertOp::Overwrite in DynamoDB re-bootstrap

claudespice:
## Summary
`scan_and_overwrite_accelerator` used `InsertOp::Append` instead of `InsertOp::Overwrite`. When DynamoDB Streams shards expire and the runtime triggers a re-bootstrap, the full DynamoDB table scan is **appended** on top of existing rows instead of replacing them, causing data duplication in the accelerated table.

The function is named `scan_and_overwrite_accelerator` but was using `Append` due to a refactoring oversight when PR #10616 extracted the shared helper from `do_rebootstrap`.

## Fix
Change `InsertOp::Append` to `InsertOp::Overwrite` at `dynamodb.rs:1066`.

## Test plan
- `cargo clippy -p runtime --features dynamodb -- -D warnings` passes
- The initial bootstrap path (empty table) produces the same result with both Append and Overwrite
- The re-bootstrap path now correctly replaces all existing data instead of duplicating it

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10667"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/crates/runtime/src/dataconnector/dynamodb.rs
+++ b/crates/runtime/src/dataconnector/dynamodb.rs
@@ -1063,7 +1063,7 @@ async fn scan_and_overwrite_accelerator(
     let table_sink = TableSink::new(accelerated_table_provider);
     let _guard = accelerator_write_mutex.lock().await;
     if let Err(e) = table_sink
-        .insert_into(instrumented_stream, InsertOp::Append)
+        .insert_into(instrumented_stream, InsertOp::Overwrite)
         .await
     {
         tracing::error!(


claudespice:
Closing as duplicate of #10639. PR #10639 contains the same one-line fix to `scan_and_overwrite_accelerator` plus the required test update (`resolve_acceleration_base_table` helper) for the `dynamodb_shard_not_found_expired_checkpoint_ready_before_load` integration test, which fails on this branch because `InsertOp::Overwrite` exposes the DuckDB acceleration table behind a view ("Binder Error: Can only delete from base table!"). #10639 already has all checks passing and all review threads resolved.

PR #10669, opened 2026-05-04
feat(runtime): declared-schema deferred datasets

### LINKED ISSUE #10659: Enhancement: Declared-schema deferred datasets
## Goal-State/What/Result

Support spicepods registering very large numbers of datasets without contacting the source at startup, while keeping each dataset fully visible in the catalog with a known schema.

Three user-visible outcomes:

1. **New deferred-but-visible dataset path.** Datasets opted in via `ready_state: on_registration` plus a known schema (user-declared `columns[].type` or a connector-provided static schema) skip all source-facing I/O at startup — and only contact the source on the first query reference. The dataset is visible in `SHOW TABLES`, `information_schema`, JDBC/Flight metadata, `/v1/datasets`, and LLM tool catalogs from t=0.

   ```yaml
   datasets:
     - from: postgres:public.events
       name: events
       ready_state: on_registration
       columns:
         - { name: id,     type: bigint, nullable: false }
         - { name: ts,     type: timestamptz }
         - { name: amount, type: numeric(18,4) }
   ```

2. **Expanded `ready_state` semantics.** `ready_state` previously selected only *when* in startup the dataset is reported ready. It now also selects *whether* startup contacts the source at all:
   - `on_load` — unchanged. Source contacted at startup; ready after initial refresh.
   - `on_registration` — **two sub-modes**, picked automatically based on whether a known schema is available:
     - *Eager* (today's behavior, when no declared/static schema): source contacted at startup; ready after `read_provider`.
     - *Deferred* (new, when a declared or connector-static schema is available): no source contact at startup; ready immediately at registration; source contacted on first query.
   - `on_schema_resolved` — unchanged. Source contacted at startup; ready after the deferred-checkpoint provider resolves.
   This makes `on_registration` the single knob users tune for the large-dataset use case; the deferred sub-mode activates purely from the presence of a known schema.

3. **`load:` removed (breaking).** The `load:` field on datasets is removed entirely. `load: on_demand` (and its `ComponentStatus::NotLoaded`) is replaced by the new deferred path under `ready_state: on_registration`, which is strictly better: the dataset is visible in the catalog from t=0 instead of being invisible until queried. Pods using `load: on_demand` fail to load with a migration error pointing at the new path.

In doing so, we also collapse today's two parallel registration code paths (eager and `on_demand`) into a single unified initialization model.

## Why/Purpose

- Customers want to point Spice at large catalogs (10k+ datasets) without paying 10k startup connections. Today the only deferral option (`load: on_demand`) makes the dataset invisible until the first query, which breaks any client that introspects the catalog before querying (BI tools, LLM agents, JDBC drivers).
- The current `ready_state: on_registration` always calls `read_provider`, which means startup time scales linearly with dataset count and source latency. Pods with thousands of datasets are unusably slow to start.
- Our LLM-tool surface (`list_datasets`, `sample_table`, NSQL prompt construction) needs schemas for every dataset to be useful at scale; a deferred-but-visible model gives that for free.
- The internal refactor unifies eager and deferred dataset bring-up into a single `DatasetInitialization` value type — fewer branches, easier hot-reload, easier reasoning about lifecycle. `load: on_demand` is replaced and removed.

## By When

**Issue/Spec written and reviewed:** _TBD_
**Done-Done:** _TBD_

## Done-Done

- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Tests
- [ ] Documentation

## The Algorithm

- [ ] Simplify.

#### Declared-schema datasets

A dataset opts into deferred source contact by combining `ready_state: on_registration` with a known schema. The schema can come from either:

1. **User-declared column types** — every entry in `columns:` has a `type:`.
2. **Connector-provided static schema** — the connector's factory returns a schema without source contact (initially: HTTP non-structured / JSON-API endpoints, used by `RefreshMode::Caching`).

```yaml
datasets:
  - from: postgres:public.events
    name: events
    ready_state: on_registration
    columns:
      - name: id
        type: bigint            # Postgres style accepted
        nullable: false
      - name: ts
        type: timestamptz
      - name: tags
        type: text[]
      - name: payload
        type: Utf8              # Arrow style also accepted
      - name: amount
        type: numeric(18,4)
```

Behavior at startup for opted-in datasets:

- **No** source-facing I/O. No `read_provider`, no connector construction (for non-accelerated), no initial refresh, no health-monitor registration.
- Dataset visible in `SHOW TABLES`, `information_schema.columns`, `/v1/datasets`, JDBC/Flight metadata immediately, with the declared/static schema.
- `/v1/ready` reports ready as soon as the dataset is registered.

Behavior on first query:

- Connector is built (lazily for non-accelerated, eagerly at startup for accelerated), `read_provider` runs, the live schema is validated against the declared/static schema, and the query proceeds.
- **Federation works on the first query.** A query joining two deferred Postgres datasets sharing a connection gets the join pushed down to Postgres, just like steady-state queries.
- If the live schema doesn't match the declared/static schema, the dataset transitions to `Error` with a structured field diff. Schema is configuration; mismatches are not retried.

#### Lenient column types

`columns[].type` accepts:

- **Postgres / SQL forms**: `bigint`, `int8`, `int4`, `int2`, `text`, `varchar(n)`, `numeric(p,s)`, `timestamptz`, `timestamp with time zone`, `bytea`, `jsonb` (→ `Utf8`), `T[]` for arrays, etc.
- **Arrow display forms**: `Int64`, `Utf8`, `Timestamp(Microsecond, UTC)`, `List<Int64>`, `Decimal128(p,s)`.
- **Map types**: `Map<K, V>` (e.g. `Map<Utf8, Int64>`), lowercase `map<k, v>`, and the canonical Arrow display form. Keys default non-nullable, values nullable.

Users write what they're comfortable with; the runtime parses both dialects.

#### Precedence: connector-static beats user-declared

If a connector exposes a static schema for the dataset, it wins. Mixing both produces a warning:

```
warn: Ignoring user-declared column types for dataset 'events'; connector
'http' provides a static schema. Remove `type:` entries from `columns:` to
silence this warning.
```

Rationale: HTTP/CDC envelope schemas are runtime-internal contracts; silently letting users override them causes confusing failures downstream. Users keep `columns:` for descriptions, embeddings, and FTS — only the `type:` field is ignored when a static schema is present.

#### Behavior matrix

| `ready_state` | declared columns | connector static | Startup contacts source? | Initial accelerated refresh at startup? | Schema source |
|---|---|---|---|---|---|
| `on_load` | (ignored) | (ignored) | yes | yes | inferred |
| `on_registration` | none | none | yes (current behavior) | yes | inferred |
| `on_registration` | all typed | none | **no** | **no** (deferred until first reference) | user-declared |
| `on_registration` | none | yes | **no** | **no** (deferred until first reference) | connector-static |
| `on_schema_resolved` | (ignored) | (ignored) | yes (verifies schema) | yes | inferred |

#### Removal of `load: on_demand` (breaking)

`load:` is removed from the spicepod schema. Pods using `load: on_demand` fail to load with a migration error pointing at the new path:

```
load: has been removed. To register a dataset without contacting the source
at startup, set ready_state: on_registration and either declare column types
(columns: [{ name, type, nullable? }, ...]) or use a connector with a
built-in static schema (e.g. http).
```

`ComponentStatus::NotLoaded` is removed alongside (no path produces it any more).

#### v1 limitations

- **Read-only access required** for deferred datasets in v1. Datasets with `access: read_write` or `read_write_create` plus a deferred schema are rejected at parse time. (Read-only is the 10k use case; lazy `read_write_provider` resolution is a follow-up.)
- **`metadata.<dataset>` available after first reference** for deferred datasets. The metadata table is registered when init runs; for deferred datasets that's on first query. (Lazy metadata-table registration is a follow-up.)
- **HTTPS connector** is the v1 implementor of `static_schema`. Other static-schema connectors (`localpod`, `memory`, `sink`, `debezium`) are a follow-up audit.

### Internals (high level)

The implementation collapses today's eager and deferred (`on_demand`) registration paths into a single model:

- **`DatasetInitialization`** — pure-data plan value capturing everything needed to materialize one dataset (connector source: eager `Arc` or lazy closure; schema source: known or from-provider; preloaded accelerated table for hot-reload). Its `initialize(self) -> DatasetReady` method consumes self to do all source-touching work — exactly the steps `register_dataset` does today, restructured into a single async method. "Eager" datasets call `initialize()` at startup; "deferred" datasets call it on first query reference. Same code, different trigger.

- **`DatasetTableProvider`** — universal placeholder registered with DataFusion at t=0 for every dataset. Holds either the parked `DatasetInitialization` or the resolved real provider. `ensure_ready` runs init under a `OnceCell`, then swaps the catalog entry (`replace_table`) with the federation-capable real provider so subsequent query planning binds directly to it. Calling `scan` on the placeholder returns an internal error — it's a planning-time signal, not an execution provider.

- **`DatasetInitializationResolver`** — async pre-planning hook in `create_logical_plan`. Repurposes the existing `OnDemandTableLoader` slot. For each `TableReference` in the parsed statement, awaits `ensure_ready` on any matching placeholder before `statement_to_plan` runs. After the swap, planning binds to the real provider, the federation analyzer downcasts successfully, and federation pushdown works on the first query.

- **AtomicBool fast path** — every SQL query passes through `has_pending_initializations()`. Implementation is a single `AtomicBool::load(Acquire)`; flips to `false` permanently when the last deferred dataset resolves. Pods with no deferred datasets (the 99% case) and pods past warm-up pay zero locks and zero hash lookups per query.

- **Plan cache bypass during warm-up** — required for correctness given the no-defensive-`scan` design. While `has_pending_initializations` is true, the plan cache is skipped. Resumes on flag flip.

What this collapses from earlier design iterations: no new `FederatedTable` variants, no separate `DeferredFederatedTableProvider` wrapper, no `notify_on_first_access` notifier, no branching at `register_federated_table` / `register_accelerated_table` / `create_accelerated_table`. All source-touching logic lives inside `DatasetInitialization::initialize()`; only its trigger differs between eager and deferred.

Hot-reload becomes the same machinery: build a new `DatasetInitialization`, swap the placeholder, eager or deferred behavior follows from `should_initialize_at_startup()`.

### Security Review

No new network-facing surface. Behavior changes are:

- Deferred datasets contact their source on first query rather than at startup; same credentials, same connector, same audit/log surface — just at a different time.
- Schema validation moves from registration time to first scan for opted-in datasets. A malicious or compromised source returning a different schema than declared causes the dataset to transition to `Error` (no data is silently coerced or returned).
- Read-write deferred datasets are rejected in v1, eliminating the variant where a write attempt could trigger lazy connector construction.

## How/Implementation Plan

Phased to keep each step shippable and validated against the existing test suite:

1. **Type parser + spicepod field** — `columns[].type`, `columns[].nullable`, `parse_declared_type` lenient parser. No runtime behavior change.
2. **`DataConnectorFactory::static_schema`** (default `None`) + `declared_schema` + `deferred_schema_for` precedence helper. No deferred behavior yet.
3. **`DatasetInitialization` value type, eager-only** — refactor `register_dataset` to flow through `DatasetInitialization::plan` + `initialize`. Behavior-preserving; validated by existing tests.
4. **`DatasetTableProvider` placeholder, eager-only** — always register the placeholder, run `ensure_ready` synchronously at startup, swap to the real provider. Validates the swap path under existing scenarios before introducing deferral.
5. **Deferral wired up** — `ConnectorSource::Lazy` + `SchemaSource::Known`, repurpose `OnDemandTableLoader` → `DatasetInitializationResolver` with the AtomicBool fast path, wire into `create_logical_plan`, plan-cache bypass during warm-up, read-write rejection. Non-accelerated only.
6. **HTTPS `static_schema`** + integration tests for non-accelerated deferred HTTP.
7. **Acceleration support** for deferred datasets (`create_accelerated_table` accepts an override schema when `Known`); skip metadata-table registration in deferred path; integration tests for HTTP + caching acceleration.
8. **Remove `load: on_demand`, `ComponentStatus::NotLoaded`**, original `OnDemandTableLoader` impl. Migration error.
9. **Docs, cookbook, scale smoke test** (10k datasets), hot-path microbenchmark.

## QA Plan

- **Unit**: type parser (Postgres aliases, Arrow display, maps, nested, garbage rejection); `declared_schema`; `deferred_schema_for` precedence; `DatasetInitialization::plan` truth table; `DatasetTableProvider::ensure_ready` single-init under concurrency; `has_pending_initializations` verified lock-free.
- **Integration** (`crates/runtime/tests/ready_state/`):
  - User-declared schema, source unreachable at startup → spice starts fast, schema visible, first query attempts source.
  - Connector-static schema (HTTP) → same.
  - HTTP + caching acceleration → no startup HTTP requests, first scan triggers fetch + refresh loop.
  - **Federation correctness on first use** — two deferred Postgres datasets sharing a connection; first cross-dataset join gets pushed down.
  - **`DatasetTableProvider::scan` is unreachable on the SQL path** — exhaustive scenario across every SQL entry point (Flight, FlightSQL, HTTP `/v1/sql`, NSQL, cluster request, direct), with a counter that fails the test if `scan` ever fires.
  - Plan cache bypass during warm-up.
  - Hot-reload of declared schema.
  - Static-vs-declared precedence + warning.
  - Schema mismatch → `Error` with field diff, no retry.
  - `load: on_demand` migration error.
- **Scale smoke**: 10k declared-schema datasets pointing at an unreachable source. Acceptance: cold start under target wallclock; sub-linear memory; steady-state query latency post-resolution indistinguishable from baseline.
- **Microbench**: `create_logical_plan` overhead under (1) no resolver, (2) resolver + `has_pending_inits = false`, (3) resolver + `true` with no matching refs. (1) and (2) within noise.

## Release Notes

**Declared-Schema Deferred Datasets**: Spicepods can now register thousands of datasets without contacting their sources at startup. Set `ready_state: on_registration` and declare column types — or use a connector with a built-in static schema like HTTP — and the dataset becomes immediately visible in the catalog (`SHOW TABLES`, `information_schema`, JDBC, `/v1/datasets`, LLM tools) while the source is only contacted on the first query.

```yaml
datasets:
  - from: postgres:public.events
    name: events
    ready_state: on_registration
    columns:
      - { name: id,      type: bigint, nullable: false }
      - { name: ts,      type: timestamptz }
      - { name: amount,  type: numeric(18,4) }
      - { name: payload, type: Utf8 }
```

Column types accept both Postgres-style (`bigint`, `timestamptz`, `text[]`, `numeric(18,4)`) and Arrow-style (`Int64`, `Utf8`, `Map<Utf8, Int64>`). Federation works on the first query — joining two deferred sources pushes down to the source as it would for any other dataset.

phillipleblanc:
Wires up deferred dataset initialization driven by user-declared `columns[].type` (or a connector's `static_schema`).

This builds on the foundations already merged in #10661 (column types + lenient parser), #10664 (`DataConnectorFactory::static_schema`), and #10665 (`DatasetInitialization` value type).

- Closes #10659

## What this enables

For an eligible dataset, the runtime registers a placeholder `DatasetTableProvider` at startup with the pre-known Arrow schema and parks a `DatasetInitialization`. SQL planning and federation analysis work against the placeholder schema **without contacting the source**. On first query reference, the placeholder is swapped out for the real provider.

A dataset is *eligible* for deferred initialization when:
- it is read-only,
- `ready_state: on_registration`,
- it has no embeddings or full-text columns,
- it has either user-declared `columns[].type` for every column **or** the connector implements `static_schema` (currently HTTPS).

Acceleration is supported via a hand-off: the lazy-init branch deregisters the placeholder and calls into the eager `register_loaded_dataset` so the AcceleratedTable, metrics, health monitor, and refresh loop are set up exactly as they are today.

## Hot-path overhead

`Datafusion::create_logical_plan` calls a resolver hook before logical planning runs federation analysis. The hook fast-exits via a single `Acquire`-ordered atomic load on a pending counter, so installs with no deferred datasets pay no extra work per query.

## Breaking change: `load: on_demand` is removed

`load: on_demand` is removed and replaced by `ready_state: on_registration` plus `columns[].type`. The old spelling now produces a structured deserialization error pointing users at the new combination. The `OnDemandTableLoader` trait, the `on_demand_datasets` registry, and `ComponentStatus::NotLoaded` are deleted.

Migration:

```yaml
# Before
- from: https://example.com/data.parquet
  name: my_data
  load: on_demand

# After
- from: https://example.com/data.parquet
  name: my_data
  ready_state: on_registration
  columns:
    - name: id
      type: bigint
    - name: name
      type: text
```

## Test impact

- New unit tests cover `declared_schema_for`, `try_plan_deferred_from_user_columns`, and the deferred-registration path.
- The previous `load_on_demand_dataset_does_not_create_connector_at_startup` integration test is replaced by `deferred_dataset_with_declared_columns_does_not_create_connector_at_startup`.
- `recursion_limit = 256` added to `integration_aws_sdk` and `integration_models` test crates (the longer state machine in `try_load_dataset_once` pushes the layout-of computation past the default 128).

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10669"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


phillipleblanc:
Thanks @copilot — pushed fixes for the substantive comments:

- **`is_deferral_eligible` ordering**: embeddings/FTS check now runs before the acceleration early-return, so accelerated datasets with embeddings/FTS correctly stay on the eager path.
- **Deregister/re-register race in the accelerated deferred branch**: added `complete_pending_initialization` (bookkeeping-only, no `deregister_table`) and reordered the swap so `register_loaded_dataset` runs first (overwriting the placeholder via its internal dereg+reg) and the pending counter / registry entry are dropped only after the real provider is registered. The window where the resolver could observe `count == 0` and a missing table is now closed.
- **`load_semaphore` propagation**: `plan_deferred` now accepts and threads the shared dataset-load semaphore into `register_loaded_dataset`, restoring `runtime.dataset_load_parallelism` throttling on first-query materialization.
- **Newline in error string**: replaced `\nSource:` with `; source:` so the schema-mismatch reason logs as a single line.
- **Refresh endpoint doc/OpenAPI**: removed `load: on_demand` references; the doc and 400 description now reflect that deferred datasets materialize automatically and the endpoint is accelerated-only.
- **Copyright headers**: bumped 2024-2025 → 2026 on the three new files.

Also fixed a separate stack-overflow in CI integration tests (futures grew large enough that the resolver path through `ensure_ready → initialize → register_loaded_dataset` overflowed the 2 MB tokio test thread stack on Linux). Added `Box::pin` at the `ensure_ready → initialize` boundary and around the inner `register_loaded_dataset` calls in the deferred path.

Re: the `load: on_demand` migration error — keeping the "unknown field" behavior intentional: `load: on_demand` was never released, so we don't owe a guided migration message, and reintroducing a deprecated field just to error on it would re-add the surface area we're deleting. Happy to revisit if a release ever shipped with it.

Added integration tests in `crates/runtime/tests/http/deferred.rs` covering:
1. Eager HTTP path (no declared columns) — source must be contacted at startup.
2. Deferred HTTP path (declared columns + `ready_state: on_registration`) — source must NOT be contacted at startup; first query triggers materialization; subsequent queries reuse the swapped provider.
3. Deferred + schema mismatch — fails fast on first query with a structured error.

sgrebnov (dismissed):

lukekim (approved):

PR #10672, opened 2026-05-04
feat(elasticsearch): direct FTS engine config + index lifecycle and ingestion controls

ewgenius:
Adds Elasticsearch as an alternative backend for Spice full-text search and adds Elasticsearch ingestion-tuning controls for both vector and text indexes.

### User-facing config

Engines are configured directly on each dataset/column. Reuse across datasets is done with YAML anchors — there is no separate top-level registry.

```yaml
version: v2
kind: Spicepod
name: elasticsearch

embeddings:
  - from: openai:text-embedding-3-small
    name: openai_embeddings
    params:
      openai_api_key: ${secrets:OPENAI_API_KEY}

x-elasticsearch-fts: &elasticsearch_fts
  enabled: true
  engine: elasticsearch
  params:
    elasticsearch_endpoint: http://localhost:9200
    elasticsearch_user: elastic
    elasticsearch_pass: spiceai
    # Optional ES ingestion controls (vector + FTS):
    # number_of_shards: 1
    # number_of_replicas: 0
    # refresh_interval: 1s              # only applied when creating the index
    # bulk_load_refresh_interval: -1    # disable refresh during bulk load
    # force_merge_after_write: true     # forcemerge after acceleration writes
    # force_merge_segments: 1
    # batch_write_rows: 1000
    # index_settings: '{"index":{"codec":"best_compression"}}'

x-elasticsearch-vectors: &elasticsearch_vectors
  enabled: true
  engine: elasticsearch
  params:
    elasticsearch_endpoint: http://localhost:9200
    elasticsearch_user: elastic
    elasticsearch_pass: spiceai
    elasticsearch_index: vector_index
    elasticsearch_vector_field: content_embedding

datasets:
  - from: file:./articles.parquet
    name: articles
    params:
      file_format: parquet
    acceleration:
      enabled: true
      engine: arrow

    vectors: *elasticsearch_vectors
    full_text_search:
      <<: *elasticsearch_fts
      params:
        elasticsearch_endpoint: http://localhost:9200
        elasticsearch_index: fts_index

    columns:
      - name: content
        embeddings:
          - from: openai_embeddings
            row_id:
              - id
            chunking:
              enabled: true
              target_chunk_size: 256
              overlap_size: 64
        vectors: *elasticsearch_vectors
        full_text_search:
          <<: *elasticsearch_fts
          row_id:
            - id

      - name: title
        full_text_search:
          <<: *elasticsearch_fts
          row_id:
            - id
```

[...]

PR #10736, opened 2026-05-07
refactor(elasticsearch): move write lifecycle hooks to TableSink via sink_indexes field

ewgenius:
## Summary

Elasticsearch FTS and vector indexes implement `on_write_start`, `on_write_complete`, and `on_write_failed\ lifecycle hooks to control bulk-write optimizations: temporarily disabling the index refresh interval, calling `_refresh`, and running `_forcemerge` after a write cycle completes.

Previously these hooks were never invoked by `TableSink` because the `IndexedTableProvider\ holding the Elasticsearch indexes only existed on the federated (read) side, not on the accelerator (write) side. As a workaround, `ElasticsearchTextIndex::compute_index` self-drove maintenance on every batch, so `_refresh` and `_forcemerge` fired after every batch instead of once per write cycle.

## Changes

- `ElasticsearchTextIndex::compute_index` simplified to write-only; `on_write_start`/`on_write_complete`/`on_write_failed` delegate to `ElasticsearchIndexWriteMaintenance` so maintenance fires once per write cycle, not per batch.
- `TableSink` gains a `sink_indexes: Vec<Arc<dyn Index>>` field. Indexes in this list receive the same `on_write_start`/`on_write_complete`/`on_write_failed` calls as indexes embedded in `IndexedTableProvider`, but are invisible to the query optimizer (`IndexTableScanOptimizerRule`).
- `AcceleratedTable::Builder` gains `add_sink_index()`. The field is threaded through `Refresher` → `RefreshTaskRunnerBuilder` → `RefreshTaskBuilder` → `AccelerationSink` → `TableSink`.
- `EmbeddingConnector::on_accelerator_setup` calls `builder.add_sink_index()` for each Elasticsearch vector index instead of wrapping the accelerator provider.
- `ElasticsearchFullTextConnector::on_accelerator_setup` calls `builder.add_sink_index()` instead of `set_accelerator()`.
- `add_elasticsearch_fts_to_table` split into `add_elasticsearch_fts_to_table` (federated path, registers index as query-visible) and `build_elasticsearch_text_index` (returns bare index for sink registration on the accelerator path).
- `wrap_accelerator_with_elasticsearch_vector_indexes` now returns `(Arc<dyn TableProvider>, Vec<Arc<dyn Index>>)` instead of a wrapped `IndexedTableProvider`.
- `write_only_indexes`, `add_write_only_index`, and `get_all_write_hooks` removed from `IndexedTableProvider`.
- Unit tests for `ElasticsearchIndexWriteMaintenance` verifying `on_write_start` idempotency and correct `on_write_complete`/`on_write_failed` behavior, using a mock client with no running Elasticsearch required.

## Related

#10672 follow-up

### DIFF
--- a/crates/runtime/src/accelerated_table/mod.rs
+++ b/crates/runtime/src/accelerated_table/mod.rs
@@ -51,7 +51,7 @@ use runtime_datafusion::execution_plan::{
     TableScanParams, fallback_on_zero_results::FallbackOnZeroResultsScanExec,
     schema_cast::SchemaCastScanExec, wrap_with_filter,
 };
-use runtime_datafusion_index::Index;
+
 use snafu::prelude::*;
 use spicepod::metric::Metrics;
 use synchronized_table::SynchronizedTable;
@@ -366,10 +366,6 @@ pub struct Builder {
     cluster_role: Option<ClusterRole>,
     user_facing_schema: Option<SchemaRef>,
     accelerator_write_mutex: Arc<Mutex<()>>,
-    /// Indexes that receive write lifecycle hooks (`on_write_start`, `on_write_failed`,
-    /// `on_write_complete`) but are not stored in the accelerator-side
-    /// [`IndexedTableProvider`] and therefore invisible to the query optimizer.
-    sink_indexes: Vec<Arc<dyn Index + Send + Sync>>,
 }
 
 impl Builder {
@@ -419,7 +415,6 @@ impl Builder {
             cluster_role: None,
             accelerator_write_mutex: Arc::new(Mutex::new(())), // can be overridden
             user_facing_schema: None,
-            sink_indexes: vec![],
         }
     }
 
@@ -486,17 +481,6 @@ impl Builder {
     /// refresher (created during build) receives the updated provider.
     pub fn set_accelerator(&mut self, accelerator: Arc<dyn TableProvider>) {
         self.accelerator = accelerator;
-    }
-
-    /// Register an index that receives write lifecycle hooks (`on_write_start`,
-    /// `on_write_failed`, `on_write_complete`) but is not stored in the accelerator-side
-    /// [`IndexedTableProvider`] and therefore invisible to the query optimizer.
-    ///
-    /// Intended for external indexes (e.g. Elasticsearch) that must be maintained in
-    /// lock-step with accelerator refreshes without participating in query planning.
-    pub fn add_sink_index(&mut self, index: Arc<dyn Index + Send + Sync>) -> &mut Self {
-        self.sink_indexes.push(index);
-        self
     }
 
     /// Set to only write to the accelerator (not replicate to federated source).
@@ -864,7 +848,6 @@ impl Builder {
         }
 
         refresher.with_s3_express_acceleration(self.is_s3_express_acceleration);
-        refresher.with_sink_indexes(self.sink_indexes);
 
         let (refresh_handle, refresh_trigger) =
             if matches!(self.cluster_role, Some(ClusterRole::Scheduler)) {
--- a/crates/runtime/src/accelerated_table/mod.rs
+++ b/crates/runtime/src/accelerated_table/mod.rs
@@ -51,6 +51,7 @@ use runtime_datafusion::execution_plan::{
     TableScanParams, fallback_on_zero_results::FallbackOnZeroResultsScanExec,
     schema_cast::SchemaCastScanExec, wrap_with_filter,
 };
+use runtime_datafusion_index::Index;
 use snafu::prelude::*;
 use spicepod::metric::Metrics;
 use synchronized_table::SynchronizedTable;
@@ -365,6 +366,10 @@ pub struct Builder {
     cluster_role: Option<ClusterRole>,
     user_facing_schema: Option<SchemaRef>,
     accelerator_write_mutex: Arc<Mutex<()>>,
+    /// Indexes that receive write lifecycle hooks (`on_write_start`, `on_write_failed`,
+    /// `on_write_complete`) but are not stored in the accelerator-side
+    /// [`IndexedTableProvider`] and therefore invisible to the query optimizer.
+    sink_indexes: Vec<Arc<dyn Index + Send + Sync>>,
 }
 
 impl Builder {
@@ -414,6 +419,7 @@ impl Builder {
             cluster_role: None,
             accelerator_write_mutex: Arc::new(Mutex::new(())), // can be overridden
             user_facing_schema: None,
+            sink_indexes: vec![],
         }
     }
 
@@ -480,6 +486,17 @@ impl Builder {
     /// refresher (created during build) receives the updated provider.
     pub fn set_accelerator(&mut self, accelerator: Arc<dyn TableProvider>) {
         self.accelerator = accelerator;
+    }
+
+    /// Register an index that receives write lifecycle hooks (`on_write_start`,
+    /// `on_write_failed`, `on_write_complete`) but is not stored in the accelerator-side
+    /// [`IndexedTableProvider`] and therefore invisible to the query optimizer.
+    ///
+    /// Intended for external indexes (e.g. Elasticsearch) that must be maintained in
+    /// lock-step with accelerator refreshes without participating in query planning.
+    pub fn add_sink_index(&mut self, index: Arc<dyn Index + Send + Sync>) -> &mut Self {
+        self.sink_indexes.push(index);
+        self
     }
 
     /// Set to only write to the accelerator (not replicate to federated source).
@@ -847,6 +864,7 @@ impl Builder {
         }
 
         refresher.with_s3_express_acceleration(self.is_s3_express_acceleration);
+        refresher.with_sink_indexes(self.sink_indexes);
 
         let (refresh_handle, refresh_trigger) =
             if matches!(self.cluster_role, Some(ClusterRole::Scheduler)) {
--- a/crates/runtime/src/accelerated_table/refresh.rs
+++ b/crates/runtime/src/accelerated_table/refresh.rs
@@ -1216,7 +1216,7 @@ impl Refresher {
 
         let notifier = self.on_complete_notification.clone();
         tokio::spawn(async move {
-            let refresh_task = Arc::new(refresh_task_builder.build().await);
+            let refresh_task = refresh_task_builder.build().await;
             if let Err(err) = refresh_task
                 .start_changes_stream(
                     refresh,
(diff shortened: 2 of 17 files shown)


crates/runtime/src/embeddings/connector.rs
```diff
@@ -305,6 +305,36 @@ impl DataConnector for EmbeddingConnector {
                 )
                 .await?;
             builder.set_accelerator(indexed_accelerator);
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
ewgenius: It is mutually exclusive by engine

crates/runtime/src/search/full_text/table.rs
```diff
@@ -270,7 +294,7 @@ mod tests {
         };
         let table_ref = datafusion::sql::TableReference::parse_str("docs");
 
-        let err = add_elasticsearch_fts_to_table(table, &columns, &table_ref, &fts_params)
+        let err = add_elasticsearch_fts_to_table(table, &columns, &table_ref, &fts_params, false)
```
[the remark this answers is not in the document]
ewgenius: fixed

crates/runtime/src/accelerated_table/sink/mod.rs
```diff
@@ -38,6 +39,20 @@ impl AccelerationSink {
...
+                AccelerationSink::Table(sink.with_sink_indexes(indexes))
+            }
+            AccelerationSink::Multi(_) => {
+                // MultiSink does not support sink_indexes; this path is not reachable
+                // at construction time (sink_indexes are set before any synchronized
+                // tables are added).
+                self
+            }
+        }
+    }
```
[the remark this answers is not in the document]
ewgenius: fixed

crates/runtime/src/accelerated_table/sink/multi.rs
```diff
@@ -139,6 +148,20 @@ impl MultiSink {
...
+                "MultiSink: running on_write_start for index '{}'",
+                index.name()
+            );
+            if let Err(e) = index.on_write_start().await {
+                tracing::warn!(
+                    "MultiSink: on_write_start failed for index '{}': {e}. Continuing with write.",
+                    index.name()
+                );
+            }
+        }
```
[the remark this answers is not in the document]
ewgenius: out of scope for this pr: 
sink_indexes hooks are correctly forwarded. The provider-embedded index hooks (NativeVectorIndex etc.) are a pre-existing gap unrelated to this PR, and all their hook implementations are no-ops today so there's no behavioral difference.

crates/runtime/src/embeddings/udtf.rs
```diff
@@ -591,15 +594,32 @@ impl VectorSearchTableFunc {
...
+                let source_for_normalize = if let Some(acc) =
+                    find_concrete_table_provider::<AcceleratedTable>(tbl)
+                    && let Some(fed) = acc.get_federated_table_ref().try_table_provider_sync_ref()
+                {
+                    federated_tbl_storage = Arc::clone(fed);
+                    &federated_tbl_storage
+                } else {
+                    tbl
+                };
+                normalized_tbl_storage =
```
Jeadie, line 624: Should retain comment
```rust
        // For Elasticsearch indexes, normalize the base table provider schema to match
        // what the ES HTTP client produces (e.g. LargeUtf8 → Utf8). This ensures that
        // HashJoinExec key types match on both sides of the join.
```

Jeadie:
> Indexes in this list receive the same on_write_start/on_write_complete/on_write_failed calls as indexes embedded in IndexedTableProvider, but are invisible to the query optimizer (IndexTableScanOptimizerRule).

`TableSink` is the only place that calls on_write_start/on_write_complete/on_write_failed. Alot of the added documentation suggest `IndexedTableProvider` or similar may do so. We should update documentation to make this clear (or, and this might be hard, call on_write_start/on_write_complete/on_write_failed from within `TableSink`).

crates/runtime/src/embeddings/connector.rs
```diff
@@ -305,6 +305,34 @@ impl DataConnector for EmbeddingConnector {
...
+
+        #[cfg(feature = "elasticsearch")]
```
Jeadie: Is this needed? I have a suspicion that it shouldn't be. Elastic, like S3Vectors is an external index and therefore there should be no issue in the runtime recreating the accelerated table?

Fix that was required for DuckDB #10668

Jeadie: Okay, so the core part here is solely the 
```
for idx in sink_indexes {
            builder.add_sink_index(idx);
}
```

My above comment about S3Vectors is only wrong because S3Vectors is not getting any of the lifecycle handlers being called (it doesn't need them). 

I think `wrap_accelerator_with_elasticsearch_vector_indexes` is not needed here (we just need the existing `wrap_table_as_index_elasticsearch`, which is used correctly elsewhere)?  You can see we dump the wrapped table `let (_, sink_indexes) `

crates/runtime/src/embeddings/connector.rs
```diff
@@ -305,6 +305,34 @@ impl DataConnector for EmbeddingConnector {
...
+            && vector_engine.enabled
+            && vector_engine.engine.as_deref() == Some("elasticsearch")
+        {
+            tracing::debug!(
+                dataset = %dataset.name,
+                "Collecting Elasticsearch vector sink indexes for accelerator"
+            );
+            let accelerator = builder.get_accelerator();
+            for (effective_vector_store, columns) in vector_index_groups(vector_engine, dataset) {
+                let (_, sink_indexes) = crate::embeddings::index::table::wrap_accelerator_with_elasticsearch_vector_indexes(
```
Jeadie: In this, we essentially recreate the ES indexes. Since `accelerated_table::Builder` has reference to the federated table and it is likely something like `FederatedTableProviderAdaptor<EmbeddedTable<IndexedTableProvider<TableProvider>>>` (eek ugly now), we can simply keep a copy of the Indexes from  `IndexedTableProvider`.

crates/runtime/src/accelerated_table/mod.rs
```diff
@@ -414,6 +419,7 @@ impl Builder {
             cluster_role: None,
             accelerator_write_mutex: Arc::new(Mutex::new(())), // can be overridden
             user_facing_schema: None,
+            sink_indexes: vec![],
```
Jeadie: Similar to the below [comment](https://github.com/spiceai/spiceai/pull/10736/changes#r3205720212),  You could populate this from `accelerated_table::Builder::federated` which is a `FederatedTable`. 

This would simplify requiring `RefreshTaskBuilder`, `RefreshTaskRunner` etc all requiring to pass around `sink_indexes`. It would also resolve the above mentioned comment (ES indexes would be registered automatically).

ewgenius: Fixed - `wrap_accelerator_with_elasticsearch_vector_indexes` and the entire `sink_indexes` pass-through chain (`Builder` → `Refresher` → `RefreshTaskRunner` → `RefreshTaskBuilder`) are removed.
Indexes are now extracted automatically from the federated provider chain in `RefreshTaskBuilder::build` via `indexes_from_federated` helper.

crates/runtime/src/accelerated_table/refresh.rs
```diff
@@ -1207,32 +1194,29 @@ impl Refresher {
...
+        .with_last_updated_at(Arc::clone(&self.last_updated_at))
+        .with_s3_express_acceleration(self.is_s3_express_acceleration);
 
         let caching = self.caching.clone();
         let refresh = Arc::clone(&self.refresh);
         let initial_load_completed = Arc::clone(&self.initial_load_completed);
 
         let notifier = self.on_complete_notification.clone();
         tokio::spawn(async move {
+            let refresh_task = Arc::new(refresh_task_builder.build().await);
```
Jeadie: intentional?

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -111,6 +110,29 @@ struct RefreshStat {
     pub memory_size: usize,
 }
 
+/// Walks the federated provider chain and collects indexes from any [`IndexedTableProvider`]
+/// layer. These indexes receive write lifecycle hooks alongside accelerator refreshes.
+async fn indexes_from_federated(
+    federated: &Arc<FederatedTable>,
```
Jeadie: there may be a `EmbeddingTable` in here?

Jeadie (dismissed):

crates/runtime/src/search/full_text/elasticsearch.rs
```diff
@@ -230,6 +230,19 @@ impl DataConnector for ElasticsearchFullTextConnector {
...
+    async fn on_accelerator_setup(
+        &self,
+        dataset: &Dataset,
+        builder: &mut accelerated_table::Builder,
+    ) -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+        self.inner_connector
+            .on_accelerator_setup(dataset, builder)
+            .await?;
+
+        Ok(())
```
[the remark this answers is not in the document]
ewgenius: PR description outdated

crates/runtime/src/accelerated_table/refresh_task.rs
```diff
@@ -110,6 +110,29 @@ struct RefreshStat {
...
+        .as_any()
+        .downcast_ref::<FederatedTableProviderAdaptor>()
+    {
+        if let Some(inner) = &adaptor.table_provider {
+            if let Some(indexed) = inner.as_any().downcast_ref::<IndexedTableProvider>() {
+                return indexed.get_all_indexes();
+            }
+        }
+    }
+    vec![]
```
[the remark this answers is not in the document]
ewgenius: fixed, all indexes are accumulated from every layer

crates/runtime/src/accelerated_table/refresh_task.rs
ewgenius: yes, this is fixed now, all indexes are accumulated from every layer

crates/runtime/src/accelerated_table/refresh.rs
ewgenius: no it is not. fixed

sgrebnov (dismissed):

sgrebnov (dismissed):

sgrebnov (approved):

sgrebnov (dismissed):

sgrebnov (dismissed):

sgrebnov (approved):

PR #10749, opened 2026-05-08
Add Arrow primary key upserts

lukekim:
## Summary
- Automatically enable Arrow `hash_index` when a primary key constraint is supplied, including partitioned Arrow tables.
- Preserve Arrow primary-key constraints so append writes can use the existing in-memory upsert path and hash-index maintenance.
- Update refresh test helpers and replication docs to treat Arrow primary-key upserts as supported.
- Add a runtime regression test covering automatic `IndexedMemTable` creation and append-as-upsert behavior.

## Testing
- `cargo fmt --check`
- `git diff --check`
- VS Code diagnostics for touched Rust files

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10749"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


crates/runtime/src/component/dataset/acceleration.rs
```diff
@@ -356,13 +356,27 @@ impl Acceleration {
...
+    #[must_use]
+    pub fn is_hash_index_auto_enabled_by_primary_key(&self) -> bool {
+        matches!(self.engine, Engine::Arrow | Engine::PartitionedArrow)
+            && self.primary_key.is_some()
+            && !matches!(self.refresh_mode, Some(RefreshMode::Caching))
+    }
+
+    /// Returns whether Arrow `hash_index` is effectively enabled, either explicitly or via `primary_key`.
+    #[must_use]
+    pub fn is_hash_index_enabled(&self) -> bool {
```
phillipleblanc, line 359: After this change, I'm not sure that having `hash_index` as a separate parameter makes sense anymore. There isn't any cases where you can enable a hash_index without a primary key, and now we're automatically enabling it when there is a primary key.

phillipleblanc (dismissed):

sgrebnov (dismissed):

phillipleblanc (dismissed):

crates/data_components/src/arrow/write.rs
sgrebnov: Note: this is dangerous: 
1. append `Moves all the elements of `other` into `self`, leaving `other` empty.` - if delete fail later all data will be lost
2.  there is a window where readers see empty/partial data.

peasee (dismissed):

sgrebnov (dismissed):

sgrebnov (dismissed):

phillipleblanc (approved):

Issue #10223, opened 2026-04-08
Full Text Search with ADBC Bug


MaxwellGrant90:
## Describe the bug

With version v2.0.0-unstable-build.1cf8026fb+models and full text search enabled, the ADBC BigQuery connector raises an error at dataset load.

## To Reproduce

_Steps to reproduce the behavior:_

2. Click on '...'
3. Scroll down to '...'

## Expected behavior

The dataset is expected to load without error from the adbc connection.

## Runtime Details

WARN runtime::accelerated_table::refresh_task: Failed to load data for dataset foo (adbc): Failed to refresh dataset from the data connector: Execution error: Index full_text changed schema.Expected fields (##):

### Spicepod

_Add the relevant `spicepod.yml` section._

full_text_search:
          enabled: true
          row_id:
            - key

### Output of `describe table`

_Add any relevant `describe table` output here._

### Output of `explain query`

_Add any relevant `explain query` output here._

### spice, spiced, OS info

- spice version: `v2.0.0`
- spiced version: `v2.0.0-unstable-build.1cf8026fb+models`
- OS info: `pop!_os`

**Have you tried this on the latest `trunk` branch?**

- [ X] Yes
- [ ] No

**If possible, run `spiced` with DEBUG log level**
By setting the environment variable `SPICED_LOG="task_history=DEBUG,spiced=DEBUG,runtime=DEBUG,secrets=DEBUG,data_components=DEBUG,cache=DEBUG,extensions=DEBUG,spice_cloud=DEBUG"`

Jeadie:
Thanks for reporting this @MaxwellGrant90. To diagnose the full issue, could you please paste the entire error message (redacted column names is also fine; as long as they match between expected and actual). Specifically we'd need the details about `Expected fields...` and `Got fields...` that come from this [error](https://github.com/spiceai/spiceai/blob/a001900fe6724e88a8bdccdfb2e8f82c3c1d8529/crates/runtime-datafusion-index/src/analyzer/index_table_scan.rs#L500-L509).

MaxwellGrant90:
@Jeadie Here is the full error message: 

[...]

PR #10778, opened 2026-05-11
fix: ignore field metadata in schema compatibility check in index_table_scan

Jeadie:
## Problem

When Full-Text Search (FTS) is enabled on an accelerated dataset, the dataset refresh fails with:

```
Input stream produced batch with unexpected schema. Expected fields (89): ... Got fields (89): ...
```

The error is misleading — the Expected and Got schemas appear **identical** in the log. The root cause is that Arrow's `Field` `PartialEq` includes **field-level metadata**. When FTS is enabled, metadata is attached to the indexed fields in the advertised schema but not in the incoming record batch stream (or vice versa). The `fields() != fields()` comparison then fails even though the schemas are structurally identical.

The error display function (`schema_signature`) intentionally omits metadata, which is why both sides of the error look the same — the actual mismatch is invisible to the user.

Reported in #10223.

## Fix

Replace both `fields() != fields()` comparisons in `IndexTableScanExec::execute` with a new `fields_structurally_equal()` helper that compares only field **name**, **data type**, and **nullability**, ignoring metadata. This correctly catches real structural mismatches while tolerating metadata differences introduced by FTS and other index layers.

The existing comment claimed the comparison already tolerated metadata differences — that comment was incorrect and has been updated to explain the actual behaviour.

## Testing

This fixes a user-reported regression (#10223) where toggling FTS off and back on caused all subsequent dataset refreshes to fail. The fix can be validated by:
1. Configuring an accelerated dataset (e.g. ADBC/BigQuery + Cayenne) with FTS enabled on one or more columns
2. Restarting spiced and confirming the dataset loads successfully

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10778"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/crates/runtime-datafusion-index/src/analyzer/index_table_scan.rs
+++ b/crates/runtime-datafusion-index/src/analyzer/index_table_scan.rs
@@ -1140,7 +1140,7 @@ mod test {
 
     /// Regression test for #10223: a field-level metadata difference (e.g. FTS attaching
     /// metadata to indexed fields) must not trigger a false "unexpected schema" error.
-    /// Arrow's `Field` PartialEq includes field-level metadata, so the old `fields() != fields()`
+    /// Arrow's `Field` `PartialEq` includes field-level metadata, so the old `fields() != fields()`
     /// check would fail even when names, types, and nullability are identical.
     #[tokio::test]
     async fn pipeline_tolerates_field_level_metadata_difference() {
--- a/crates/runtime-datafusion-index/src/analyzer/index_table_scan.rs
+++ b/crates/runtime-datafusion-index/src/analyzer/index_table_scan.rs
@@ -484,8 +484,13 @@ impl ExecutionPlan for IndexerExec {
                     let mut b = batch;
 
                     // Verify the incoming batch fields match the advertised schema.
-                    // Uses fields-only comparison to tolerate metadata differences.
-                    if b.schema().fields() != advertised_schema.fields() {
+                    // Use a metadata-ignoring comparison: Arrow's `Field` PartialEq includes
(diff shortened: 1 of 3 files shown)


crates/runtime-datafusion-index/src/analyzer/index_table_scan.rs
```diff
@@ -604,6 +612,22 @@ impl ExecutionPlan for IndexerExec {
...
 
 /// Helper for better diagnostics when schema is mismatched.
+/// Returns `true` if two field lists are structurally compatible: same length, and every
+/// corresponding pair has the same name, data type, and nullability.
+///
+/// Arrow's [`Field`] `PartialEq` includes field-level metadata, which means two schemas that are
+/// visually identical can compare as unequal when one side carries index-attached metadata (e.g.
+/// FTS column markers) and the other does not. This function intentionally ignores metadata so
+/// that schema validation catches real structural mismatches without false positives.
+fn fields_structurally_equal(a: &arrow::datatypes::Fields, b: &arrow::datatypes::Fields) -> bool {
```
phillipleblanc: There should already be something similar in the arrow_tools crate that we have. If it isn't there, this should move there and we should use it from there.

phillipleblanc (dismissed):

phillipleblanc (dismissed):

lukekim (approved):

PR #10794, opened 2026-05-12
Remove permanent errors from DynamoDB Streams

### LINKED ISSUE #10755: DynamoDB Streams: consumer permanently stops after transient error, no retry, pod remains Ready
## Description

When the DynamoDB Streams CDC consumer encounters a transient AWS error (e.g., `AccessDeniedException` from a temporary IAM permission revocation), Spice logs the error, marks the changes stream as ended, and **never retries or restarts the consumer**. The pod continues to:

- Pass liveness and readiness probes
- Serve queries (returning increasingly stale data)
- Upload snapshots on schedule (containing the same stale data + dead checkpoint)

There is no retry, no backoff, no stream restart, and no readiness failure. The only signal is a single ERROR log line at the moment of failure, after which the pod is silent. The `dataset_dynamodb_shards_active` metric does **not** drop to zero — it continues reporting the number of shards that exist on the table, not the number being actively consumed.

### Cascading failure (snapshot poisoning)

Because the pod keeps uploading snapshots after the stream dies, all snapshots in S3 inherit the dead checkpoint. When new pods start (e.g., during a rollout), they download a stale snapshot, find the checkpoint is >24h old, and either:
- Crash-loop (if lag behavior is `error`)
- Start from an expired shard and fail (if lag behavior is default)

This turns a single transient error into a cluster-wide outage that requires manual snapshot deletion to recover.

## Spice Version

- **Runtime**: v1.11.2 (reproduced in production)

## Environment

- Kubernetes Deployment (2 replicas, DuckDB file mode, S3 snapshots)
- 1 DynamoDB table (~5M items), pay-per-request, Streams enabled
- Snapshot interval: 10 minutes
- IAM role assumed via pod annotation

## Spicepod Configuration

```yaml
version: v1
kind: Spicepod
name: my-app

snapshots:
  enabled: true
  location: "s3://my-bucket/snapshots/"
  params:
    s3_region: us-east-2
  bootstrap_on_failure_behavior: fallback

datasets:
  - name: my_table
    from: "dynamodb:arn:aws:dynamodb:us-east-2:123456789012:table/my-table"
    params:
      dynamodb_aws_region: us-east-2
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
      on_conflict:
        (hash_key, sort_key): upsert
      refresh_check_interval: 3s
      refresh_mode: changes
      snapshots: enabled
      snapshots_compaction: enabled
```

## Steps to Reproduce

1. Deploy Spice with the config above. Ensure the DynamoDB Streams consumer is running and healthy (`records_consumed_total` incrementing).
2. Temporarily revoke the IAM role's `dynamodb:GetRecords` / `dynamodb:GetShardIterator` permission (simulating a transient access loss — e.g., LakeFormation permission change, STS token expiry, or IAM policy update).
3. Wait for the consumer to hit the permission boundary (~3s refresh interval).
4. Restore the IAM permission.
5. Observe the pod.

## Expected Behavior

After the transient error:

1. Spice retries the stream operation with exponential backoff
2. Once the permission is restored, the consumer resumes from its last checkpoint
3. If retries are exhausted, Spice restarts the stream consumer from the last good checkpoint
4. If the stream cannot be restored, the pod fails its readiness probe (stops serving stale data)
5. Snapshots are suppressed while the stream is inactive (prevents poisoning)

## Actual Behavior

```
T+0     Stream consuming normally, records_consumed_total incrementing

T+80m   IAM permission revoked (external event)

T+80m   ERROR runtime::accelerated_table::refresh_task::changes:
          Changes stream error for my_table: DynamoDB error:
          Failed to receive DynamoDB Stream record:
          AWS SDK error: unhandled error (AccessDeniedException)

T+80m   WARN runtime::accelerated_table::refresh_task::changes:
          Changes stream ended for dataset my_table

        --- No further errors. No retry. No restart. ---

T+80m+  Pod remains 1/1 Ready
T+80m+  Queries return data frozen at T+80m
T+80m+  Snapshots uploaded every 10 minutes (stale data, dead checkpoint)
T+80m+  shards_active metric unchanged (still reports >0)

T+3d    New pod rolls out, downloads stale snapshot
T+3d    ERROR: DynamoDB table lag references expired shard.
          Configured behavior is 'error'  lag_age=3days
T+3d    New pod crash-loops indefinitely
```

### Observed log pattern on "healthy" pods after stream death

The only ongoing log output is the snapshot cycle. Note `last_updated_at_ms` never advances:

```
INFO  Uploading snapshot. dataset=my_table snapshot=...20260508T110323Z.duckdb
INFO  Snapshot uploaded. dataset=my_table size=699936768
INFO  Successfully created snapshot for dataset: my_table
INFO  Skipping snapshot creation - no updates since last snapshot.
        dataset=my_table last_updated_at_ms=1777997219683
INFO  Uploading snapshot. dataset=my_table snapshot=...20260508T112323Z.duckdb
  ... repeats indefinitely ...
```

### 1. No retry on stream consumer errors

In `crates/runtime/src/accelerated_table/refresh_task/changes.rs`, when the changes stream yields an error, it is logged and the stream loop exits. There is no retry mechanism, no backoff, and no attempt to re-establish the stream consumer.

The error handling path appears to be:
```
stream.next() → Err(e) → log error → break → "Changes stream ended" → return
```

For permanent errors (table deleted, schema mismatch) this is reasonable. For transient errors (`AccessDeniedException`, `ProvisionedThroughputExceededException`, `InternalServerError`, network timeout), the consumer should retry.

### 2. Readiness probe unaffected by stream death

The `/v1/ready` endpoint checks whether the dataset was *ever* initialized successfully. Once Ready, it stays Ready regardless of whether the CDC stream is still running. There is no "liveness" concept for the ongoing stream — only the initial bootstrap matters.

### 3. `shards_active` metric is misleading

The metric reports the count of shards returned by `DescribeStream` — i.e., shards that exist on the DynamoDB table. It does not track whether the Spice consumer has an active iterator on those shards. When the consumer dies, the shards still exist, so the metric stays unchanged.

### 4. Snapshots uploaded unconditionally

The snapshot timer fires regardless of stream health. It checks whether `last_updated_at` has advanced since the last snapshot — if not, it logs "Skipping" — but on alternating cycles it still uploads the same data with a new timestamp. This poisons the snapshot pool for any new pod that starts.

### 1. Retry transient stream errors with exponential backoff

Classify errors as transient vs. permanent:
- **Transient** (retry): `AccessDeniedException`, `ProvisionedThroughputExceededException`, `InternalServerError`, `ServiceUnavailable`, network timeouts
- **Permanent** (fail): `ResourceNotFoundException`, `ValidationException`, schema mismatch

For transient errors, retry with exponential backoff (e.g., 1s → 2s → 4s → ... → 5min cap). After max retries, attempt a full stream restart from the last committed checkpoint.

### 2. Fail readiness when stream is dead

If the changes stream has ended and cannot be restarted, fail the `/v1/ready` probe. The pod should be removed from service rather than silently serving stale data.

### 3. Suppress snapshot uploads when stream is inactive

If `last_updated_at` hasn't advanced in N consecutive snapshot intervals (e.g., 3×), stop uploading. This prevents a single pod's failure from poisoning the entire snapshot pool.

### 4. Fix `shards_active` metric semantics

Either:
- Rename to `shards_available` (table-level) and add `shards_consuming` (consumer-level)
- Or change `shards_active` to only count shards with a live iterator

### 5. Add `stream_status` metric

Expose a gauge per dataset: `1` = consuming, `0` = dead. This is the simplest alerting signal.

## Workaround

1. **Alert on `records_consumed_total` stall**: If `increase(dataset_dynamodb_records_consumed_total[1h]) == 0` while `shards_active > 0`, the consumer is dead.
2. **Delete stale snapshots** from S3 manually or via automation.
3. **Rolling restart** pods to force a fresh bootstrap.
4. **Monitor IAM/permission changes** that could cause transient `AccessDeniedException` and proactively restart pods after restoring access.

## Related

- #10582 — Bootstrap exceeding 24h makes CDC stream un-resumable (same "stream ends, no recovery" pattern, different trigger)
- #7880 — DynamoDB Streams enhancement
- #8506 — Fix for idle DynamoDB Stream

krinart:
## 📝 Summary

Closes https://github.com/spiceai/spiceai/issues/10755

Treat all errors as part of stream processing in `dynamodb-streams` as transient:
* No error shuts down the stream
* After each error retry with backoff is applied
* Apply same `LagExceedsShardRetentionBehavior` logic that we do for `ShardNotFoundError` to `StreamBeyondRetentionError`

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10794"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/crates/dynamodb-streams/src/stream.rs
+++ b/crates/dynamodb-streams/src/stream.rs
@@ -56,7 +56,7 @@ impl DynamodbStreamProducer {
         let mut had_error = false;
 
         // 1. Initialize shards that require iterators
-        self.initialize_shards_iterators().await;
+        had_error |= self.initialize_shards_iterators().await;
 
         // 2. Poll active shards
         let futures = self.state.get_active_shards().map(|shard| {
@@ -99,10 +99,11 @@ impl DynamodbStreamProducer {
         (combine_shard_batches(&poll_results), had_error)
     }
 
-    async fn initialize_shards_iterators(&mut self) {
+    async fn initialize_shards_iterators(&mut self) -> bool {
         let shards: Vec<InitializingShard> =
             self.state.get_initializing_shards().cloned().collect();
 
+        let mut had_error = false;
         for shard in shards {
             // Shards that were already polled use `After`.
             // Newly discovered shards use `At`.
@@ -125,6 +126,7 @@ impl DynamodbStreamProducer {
                     self.state.mark_active(shard.shard_id, iterator);
                 }
                 Err(e) => {
+                    had_error = true;
                     tracing::warn!(
                         "Failed to initialize shard. Will retry on next iteration: {}",
                         e
@@ -132,6 +134,7 @@ impl DynamodbStreamProducer {
                 }
             }
         }
+        had_error
     }
 
     pub async fn streaming(mut self) {
--- a/crates/dynamodb-streams/src/stream.rs
+++ b/crates/dynamodb-streams/src/stream.rs
@@ -17,7 +17,6 @@ use crate::checkpoint::{Checkpoint, CheckpointPosition};
 use crate::client_sdk::SDKClient;
 use crate::metrics::MetricsCollector;
 use crate::stream_state::{InitializingShard, PollOutcome, ShardPollResult, StreamState};
-use crate::{Result, StreamResult};
 use aws_sdk_dynamodbstreams::types::{Record, ShardIteratorType};
 use futures::{Stream, future::join_all};
 use std::collections::HashMap;
@@ -37,7 +36,7 @@ pub struct DynamodbStreamProducer {
     pub stream_arn: String,
     pub state: StreamState,
     pub interval: Option<Duration>,
-    pub sender: mpsc::Sender<StreamResult>,
+    pub sender: mpsc::Sender<DynamoDBStreamBatch>,
     pub client: Arc<SDKClient>,
     pub retry_strategy: RetryBackoff,
     pub metrics_collector: Arc<MetricsCollector>,
@@ -52,13 +51,12 @@ pub struct DynamoDBStreamBatch {
 const DEFAULT_SLEEP_DURATION: Duration = Duration::from_millis(500);
 
 impl DynamodbStreamProducer {
-    async fn collect(&mut self) -> Result<(DynamoDBStreamBatch, bool)> {
+    async fn collect(&mut self) -> (DynamoDBStreamBatch, bool) {
         let mut poll_results = Vec::new();
-        let mut had_transient_error = false;
+        let mut had_error = false;
 
         // 1. Initialize shards that require iterators
-        // If permanent error is encountered, it is surfaced to the client.
-        self.initialize_shards_iterators().await?;
+        self.initialize_shards_iterators().await;
 
         // 2. Poll active shards
         let futures = self.state.get_active_shards().map(|shard| {
@@ -76,34 +74,32 @@ impl DynamodbStreamProducer {
         // 3. Process poll results
         for (shard_id, result) in results {
             let poll_result = match result {
-                Ok((next_iter, records)) => self
-                    .state
-                    .handle_poll_result(&shard_id, next_iter, records)?,
+                Ok((next_iter, records)) => {
+                    self.state.handle_poll_result(&shard_id, next_iter, records)
+                }
                 Err(e) => {
-                    had_transient_error = true;
-                    self.state.handle_poll_error(&shard_id, e)?
+                    had_error = true;
+                    self.state.handle_poll_error(&shard_id, e)
                 }
             };
-            poll_results.push(poll_result);
+            if let Some(poll_result) = poll_result {
+                poll_results.push(poll_result);
+            }
         }
 
         // 4. Discover new shards
-        // If permanent error is encountered, it is surfaced to the client.
         match self.client.get_all_shards(&self.stream_arn).await {
-            Ok(shards) => self.state.add_discovered(&shards)?,
+            Ok(shards) => self.state.add_discovered(&shards),
             Err(e) => {
-                if !e.is_retriable() {
-                    return Err(e);
-                }
-                had_transient_error = true;
+                had_error = true;
                 tracing::warn!("Failed to discover new shards. Will retry on next iteration: {e}");
             }
         }
 
-        Ok((combine_shard_batches(&poll_results), had_transient_error))
+        (combine_shard_batches(&poll_results), had_error)
     }
 
-    async fn initialize_shards_iterators(&mut self) -> Result<()> {
+    async fn initialize_shards_iterators(&mut self) {
         let shards: Vec<InitializingShard> =
             self.state.get_initializing_shards().cloned().collect();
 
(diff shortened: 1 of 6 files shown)


crates/dynamodb-streams/src/stream.rs
```diff
@@ -151,65 +145,50 @@ impl DynamodbStreamProducer {
...
-                                    .max(self.interval.unwrap_or(Duration::from_secs(0))),
-                            )
-                            .await;
-                        } else if let Some(duration) = self.interval {
-                            sleep(duration).await;
-                        }
+            if had_error {
+                if let Some(mut duration) = backoff.next_backoff() {
+                    if duration > Duration::from_secs(60) {
+                        duration = Duration::from_secs(1);
```
Jeadie: Why go from 60s backoff to 1s backoff? I'd assume we would just want to clamp at 60s (i.e. `duration.min(Duration::from_secs(60))`)?

krinart: Good point. Done, thanks!

crates/runtime/src/dataconnector/dynamodb.rs
```diff
@@ -753,6 +753,45 @@ fn resume_from_checkpoint_stream(
...
                     }
                 }
+                Err(Error::FailedToInitializeCheckpoint {
+                    source: dynamodb_streams::Error::StreamBeyondRetention,
+                }) => {
+                    // StreamBeyondRetention definitively means checkpoint is >24h old
+                    if lag_exceeds_behavior == LagExceedsShardRetentionBehavior::Error {
+                        tracing::error!(
+                            dataset = %dataset_name,
+                            "DynamoDB Stream data has exceeded 24h retention period. Configured behavior is 'error'"
```
phillipleblanc:
```suggestion
                            "DynamoDB Streams checkpoint is older than the stream retention window; ingestion cannot resume from the saved checkpoint."
```

saying "Configured behavior is 'error'" is redundant.

crates/runtime/src/dataconnector/dynamodb.rs
```diff
@@ -753,6 +753,45 @@ fn resume_from_checkpoint_stream(
...
+                                ))
+                            })
+                            .boxed(),
+                        )
+                    } else {
+                        tracing::info!(
+                            dataset = %dataset_name,
+                            behavior = ?lag_exceeds_behavior,
+                            "DynamoDB Stream data has exceeded 24h retention period. Initiating table re-initialization"
+                        );
```
phillipleblanc, line 780:
```suggestion
                        tracing::info!(
                            dataset = %dataset_name,
                            "DynamoDB Streams checkpoint is older than the stream retention window; \
                             ingestion cannot resume from the saved checkpoint. Rebuilding the dataset from a fresh DynamoDB table scan."
                        );
```

lukekim (approved):

PR #10803, opened 2026-05-12
Properly handle nullability during CDC processing

krinart:
## 📝 Summary

Postgres native CDC promotes all fields to nullable so that `DELETE` rows — which only carry primary key columns in the WAL — can be written as Arrow batches without violating non-nullable field constraints. 

However this breaks ingestion when schema expects non-nullable fields.

This change casts schema to the one expected by acceleration before applying upsert as part of cdc processing.

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10803"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task/changes.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task/changes.rs
@@ -1870,7 +1870,7 @@ mod tests {
             !coerced
                 .schema()
                 .field_with_name("id")
-                .unwrap()
+                .expect("id field exists")
                 .is_nullable(),
             "id should be promoted to non-nullable"
         );
@@ -1878,7 +1878,7 @@ mod tests {
             coerced
                 .schema()
                 .field_with_name("name")
-                .unwrap()
+                .expect("name field exists")
                 .is_nullable(),
             "name should remain nullable"
         );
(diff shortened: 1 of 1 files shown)


crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -891,6 +866,44 @@ fn cdc_item_memory_size(item: &Result<cdc::ChangeEnvelope, cdc::StreamError>) ->
...
+    batch: RecordBatch,
+    target_schema: &SchemaRef,
+) -> crate::accelerated_table::Result<RecordBatch> {
+    let needs_coercion = batch.schema().fields().iter().any(|f| {
+        target_schema
+            .field_with_name(f.name())
+            .is_ok_and(|tf| f.is_nullable() && !tf.is_nullable())
+    });
+
+    if !needs_coercion {
```
sgrebnov: Can we just re-use try_cast_to? Does it add any additional overhead in comparison to manual logic?

krinart: Done, thanks

lukekim (dismissed):

peasee (approved):

PR #10811, opened 2026-05-13
Improve Cayenne mutation, scan, and inline memtable scaling

lukekim:
## Summary
- Extract Cayenne append writes into a shared `AppendMutationWriter` that owns inline eligibility, bounded buffering, fallback re-streaming, staged Vortex appends, retention/sort post-processing, stats persistence, and inline checkpointing.
- Treat Arrow IPC inline data as a bounded level-0 memtable: catalog stats now track inline row count, segment count, and IPC bytes, and append writes checkpoint inline data to Vortex when pressure exceeds row, segment, or byte thresholds.
- Scale Cayenne scan planning with `target_partitions` so unordered scans are not capped by Vortex file/file-group count, while preserving single-partition behavior for limited reads.
- Add focused unit, integration, Criterion, and ignored performance-test coverage for inline/fallback append behavior, UPDATE's delete-plus-append path, inline memtable pressure, and scan repartitioning.

## Review Notes
- Reverted executor-side Ballista `target_partitions` propagation after review because executor-local `SessionConfig::target_partitions` does not drive scheduler-time partition planning.
- Addressed reviewer feedback for complex tuple returns, limit scan repartitioning, fully-deleted inline memtable cleanup, sequence-monotonic inline-delete fixtures, and `RecordBatch` import paths.

## Benchmark Notes
- Re-ran `benchmark_roundtrip` with `SPICE_ACCEL_BENCH_RECORDS=100` and `SPICE_ACCEL_BENCH_ITERATIONS=10`.
- Cayenne P95 round-trip: SQLite metastore `23.27ms`, Turso metastore `30.63ms`.
- The expected win for this change is small, frequent inline-friendly mutations. Full round-trip benchmarks include delete and full-scan work over accumulated state, so this PR bounds that level-0 inline buildup with pressure-based checkpointing.

## Verification
- `RUSTC_WRAPPER= cargo test -p cayenne provider::mutation_writer --lib`
- `RUSTC_WRAPPER= cargo test -p cayenne provider::table::tests::inline_memtable --lib`
- `RUSTC_WRAPPER= cargo test -p cayenne --test data_inlining_test test_inline_writer_fallback_preserves_buffered_and_remaining_batches_sqlite -- --nocapture`
- `RUSTC_WRAPPER= cargo test -p cayenne --test data_inlining_test test_inline_memtable_segment_pressure_checkpoints_sqlite -- --nocapture`
- `RUSTC_WRAPPER= cargo test -p cayenne --features turso --test data_inlining_test test_inline_memtable_segment_pressure_checkpoints_turso -- --nocapture`
- `RUSTC_WRAPPER= cargo test -p cayenne --features turso --test data_inlining_test test_inline_memtable_pressure_flushes_after_legacy_deletes -- --nocapture`
- `RUSTC_WRAPPER= cargo test -p cayenne --test data_inlining_test test_inlined_data_crud_sqlite -- --nocapture`
- `RUSTC_WRAPPER= cargo test -p cayenne --features turso --test data_inlining_test test_inlined_data_crud_turso -- --nocapture`
- `RUSTC_WRAPPER= cargo test -p cayenne --test update_test test_update_preserves_inline_append_path_impl_sqlite -- --nocapture`
- `RUSTC_WRAPPER= cargo test -p cayenne --test data_inlining_test perf_many_small_inline_appends_sqlite -- --ignored --nocapture`
- `RUSTC_WRAPPER= cargo bench -p cayenne --bench mutation_writer -- --test`
- `RUSTC_WRAPPER= SPICE_ACCEL_BENCH_RECORDS=100 SPICE_ACCEL_BENCH_ITERATIONS=10 cargo test -p runtime --lib --no-default-features --features sqlite,turso,duckdb benchmark_roundtrip -- --ignored --nocapture`
- `git diff --check`
- `RUSTC_WRAPPER= make lint`

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10811"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


crates/cayenne/src/provider/mutation_writer.rs
```diff
@@ -0,0 +1,533 @@
...
+        }
+
+        let (
+            mut prepared_stream,
+            delete_specs,
+            deleted_pk_i64,
+            deleted_row_keys,
+            deleted_inlined_pk_i64,
+            deleted_inlined_row_keys,
+        ) = self.table.prepare_stream_for_insert(data).await?;
```
peasee: I am not a fan of complex tuple returns. use a struct for the return value of this - or it could be an indication this function needs to be split up.

peasee (dismissed):

crates/runtime/src/cluster/mod.rs
```diff
@@ -1253,6 +1253,14 @@ pub async fn initialize_cluster_executor(
         });
     };
 
+    let executor_target_partitions = app_def
```
phillipleblanc: This value should be kept in sync with what the executor reports to the scheduler as its available concurrent task slots. That number is still the number of CPU cores, which if this PR was merged as is, would lead to a discrepancy from what the scheduler thinks the executor capacity is to what the physical execution on the executor thinks its capacity is.

Instead of making this configurable, I think it would be better to just set `.with_target_partitions(concurrent_tasks)`. Alternatively we need to fix how many task slots the executor reports to the scheduler to match this value.

phillipleblanc: Actually I think we should revert this completely, this doesn't do what we think it does (what I described above is incorrect as well). This only affects the small subset of physical operators that consult the target_partitions at runtime (i.e. sort spill behavior, some hash-aggregate sub-partitioning, occasional internal repartitions). The partitioning is decided at the scheduler planning time and the executor just executes directly, without caring what the `target_partitions` value is set to.

sgrebnov (approved):

krinart (approved):

PR #10815, opened 2026-05-13
Improve Spice CLI manifest editing and direct command modes

lukekim:
## Summary
- Centralize Spicepod manifest discovery and editing so CLI mutations support both `spicepod.yaml` and `spicepod.yml`, preserve schema directives, and validate required header fields without rejecting newer manifest fields.
- Normalize local pod installs into `spicepods/<name>/` with a canonical `spicepod.yaml`, and write dependency paths relative to the app root.
- Tighten component and metadata mutation behavior: `--set` values are string-first, `yaml:` enables typed YAML values, explicit `--manifest` paths must exist, and nested updates fail instead of overwriting non-mapping values.
- Add direct CLI entry points for automation, including `--machine`/`--programmatic`, direct SQL with `-sql <sql>`, and direct chat prompts with `-p <prompt>` / `-chat <prompt>`.
- Preserve cloud resource request intent by omitting unspecified memory/storage fields instead of injecting defaults, and keep cloud update-channel request fields aligned with API expectations.
- Keep branch-local optimizer/query coverage aligned: ChBench q21 is included, Cayenne filter propagation falls back to structural checks when stats are absent, and the two dynamic HTTP header tests remain ignored with issue-context comments until #10861 is fixed.

## Notes
- Local path dependencies are installed under `spicepods/<name>/` and manifest references are written relative to the app root, e.g. `spicepods/localpod`.
- `--cloud` is a global switch for Cloud runtime endpoints. It defaults to `us-east-1`; use `--cloud-region us-west-2` or the legacy-compatible `--cloud us-west-2` form to target another supported endpoint region.
- `--set key=value` writes string values by default. Use `--set key=yaml:<value>` for typed YAML and `--set key=string:<value>` when an explicit string prefix is useful.
- Manifest writes preserve the schema directive, but YAML formatting may still be normalized when the manifest is rewritten.
- The removed `test_cloud_rollback` coverage was stale: `cloud rollback` is not currently exposed by `CloudCommands`, and the help coverage now documents that absence.

## Tests
- `RUSTC_WRAPPER= CC=cc CXX=c++ cargo test -p spice --bin spice --lib`
- `RUSTC_WRAPPER= CC=cc CXX=c++ cargo test -p spice-cloud-client`
- `RUSTC_WRAPPER= CC=cc CXX=c++ cargo test -p spice --test cloud_integration test_cloud_help_lists_all_subcommands`
- `cargo fmt --all -- --check`
- Scoped `cargo clippy` for `spice` and `spice-cloud-client` with the Makefile denial flags
- Full `make lint` was attempted locally but did not complete within the timeout in this environment; scoped fmt/clippy above passed with `RUSTC_WRAPPER=` to bypass the local sccache permission issue.

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10815"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


bin/spice/src/manifest.rs
```diff
@@ -0,0 +1,309 @@
...
+use yaml::{Mapping, Value};
+
+pub const SPICEPOD_YAML: &str = "spicepod.yaml";
+pub const SPICEPOD_YML: &str = "spicepod.yml";
+
+const SPICEPOD_FILENAMES: [&str; 2] = [SPICEPOD_YAML, SPICEPOD_YML];
+const SCHEMA_DIRECTIVE: &str = "# yaml-language-server: $schema=https://raw.githubusercontent.com/spiceai/spiceai/trunk/.schema/spicepod.schema.json";
+
+#[must_use]
+pub fn existing_spicepod_path(base_dir: &Path) -> Option<PathBuf> {
```
Jeadie, line 37: nit: methods in this file could use docstring/comments.

phillipleblanc (dismissed):

bin/spice/src/commands/component.rs
```diff
@@ -0,0 +1,1530 @@
...
+use crate::error::{ConfigIoSnafu, InvalidArgumentSnafu, Result};
+use crate::manifest;
+use clap::{Args, Subcommand, ValueHint};
+use snafu::{ResultExt, ensure};
+use std::io::{self, Read};
+use std::path::{Path, PathBuf};
+use yaml::{Mapping, Value};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ComponentSection {
```
Jeadie, line 30: we should use `spicepod` crate?

Jeadie (commented):
We should either update or completely remove these CLI docs. They are almost always out of date, and somewhat redundant (most users/agents will just explore CLI). 

https://github.com/spiceai/docs/tree/033cc95e28be398999f4258d7c404041cef6906c/website/docs/cli

bin/spice/src/commands/cloud/mod.rs
```diff
@@ -463,6 +484,14 @@ pub struct CreateAppArgs {
...
     pub name: String,
 
+    /// Deployment region (e.g. us-east-1-prod-aws-data)
+    #[arg(long)]
+    pub region: String,
+
+    /// App kind (set or cluster)
+    #[arg(long, value_parser = clap::value_parser!(AppKind), default_value = "set")]
+    pub kind: AppKind,
+
```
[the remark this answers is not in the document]
Jeadie: yeah, using this alot this week,  this is very confusing UI (both `--cloud` and `--region`)

PR #10816, opened 2026-05-13
Add range fallback for large join filters

lukekim:
This improves large multi-table join behavior by keeping dynamic join filters useful after the exact in-list accumulator reaches its memory cap.

Summary:
- Track exact join key arrays only while they remain under the configured per-partition memory limit.
- Defer min/max range-bound calculation on the common exact in-list path; when the memory limit is exceeded, rebuild bounds from the already-collected exact arrays plus the overflowing batch and then keep updating range bounds for subsequent batches.
- Emit a conservative range predicate when the exact in-list would exceed the limit.
- Preserve correctness by returning the existing no-op filter fallback when range comparison is unsupported, collected build-side keys include nulls, or float keys include NaN.
- Add comprehensive unit coverage for exact memory-limit boundary behavior, lazy range-bound calculation, multi-batch range widening after fallback, null/all-null handling, unsupported types, NaN handling, string ranges, duplicates, empty batches, and multi-batch exact accumulation.
- Add a Criterion benchmark target for exact vs range-fallback accumulator updates and physical-expression construction.

Verification:
- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 cargo test -p runtime-datafusion join_accumulator`
- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 cargo bench -p runtime-datafusion --bench join_accumulator --no-run`
- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 cargo bench -p runtime-datafusion --bench join_accumulator -- --quick`
- `RUSTC_WRAPPER= CARGO_INCREMENTAL=0 make lint`

### DIFF
--- a/crates/runtime-datafusion/src/join_accumulator/mod.rs
+++ b/crates/runtime-datafusion/src/join_accumulator/mod.rs
@@ -82,27 +82,30 @@ impl CollectLeftAccumulator for ExactLeftAccumulator {
         // eagerly evaluate the expression and store the resulting array
         // this avoids storing the entire record batch in memory, only storing the evaluated column
         let array = self.expr.evaluate(batch)?.into_array(batch.num_rows())?;
-        self.range_bounds.update(array.as_ref())?;
 
         if self.exact_values_exceeded_memory_limit {
+            self.range_bounds.update(array.as_ref())?;
             return Ok(());
         }
 
-        self.total_memory_size = self
+        let total_memory_size = self
             .total_memory_size
             .saturating_add(array.get_array_memory_size());
 
-        if self.total_memory_size > self.max_inlist_memory_size {
+        if total_memory_size > self.max_inlist_memory_size {
             tracing::debug!(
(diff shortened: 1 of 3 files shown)


crates/runtime-datafusion/src/join_accumulator/mod.rs
```diff
@@ -172,17 +225,167 @@ impl ColumnBounds for ExactColumnBounds {
...
+        }
+
+        if array.null_count() > 0 {
+            // Short-circuit: physical_expr() returns a no-op when contains_null is true,
+            // so skip the O(n) per-row loop for this and all future batches.
+            self.contains_null = true;
+            return Ok(());
+        }
+
+        for row_index in 0..array.len() {
```
sgrebnov, line 268: I'm under impression that we can use `arrow:compute:min/max` for better performance but we can do this separately

krinart (approved):

sgrebnov (approved):

PR #10819, opened 2026-05-13
Synchronize Cayenne partition commits across partitions (#10125)

### LINKED ISSUE #10125: Synchronize Cayenne partition commits across partitions
Cayenne partitioned commits are not synchronized across partitions yet, and complete per-partition. They should synchronize and commit all at the same time, similar to DuckDB.

phillipleblanc:
## Summary

Closes #10125 by coordinating partitioned Cayenne commits across all touched partitions. Previously, each partition committed independently, so a failed multi-partition refresh could leave the table in a mixed state.

This PR adds `CayennePartitionedInsertStrategy`, which streams input into per-partition writers, stages each partition independently, then publishes the prepared partitions together.

| Write mode | Publish mechanism |
|---|---|
| **Overwrite** | Write each partition to a new snapshot, then flip every `current_snapshot_id` in one `MetastoreTransaction`. |
| **Append** | Stage each partition, write one top-level `_partitioned_wal`, hold all partition `listing_fence` write locks, move files, refresh listing tables, then remove the WAL. |

It also adds:
- prepared commit lifecycles for overwrite and staged append;
- `commit_compaction_in_txn` for batched catalog updates;
- `ArcSwap<ListingTable>` plus a scan/listing fence so scans do not interleave with append barriers;
- cross-partition tests and a small fence-overhead benchmark.

## Partitioned overwrite flow

```text
Input stream
     |
     v
Partition rows by key
     |
     v
+------------------+   +------------------+   +------------------+
| Partition A      |   | Partition B      |   | Partition C      |
| begin_overwrite  |   | begin_overwrite  |   | begin_overwrite  |
| write new snap   |   | write new snap   |   | write new snap   |
+--------+---------+   +--------+---------+   +--------+---------+
         \                    |                    /
          \                   |                   /
           v                  v                  v
        Prepared overwrite receipts (catalog still points at old snaps)
                              |
                              v
              One shared MetastoreTransaction
              apply all current_snapshot_id flips
                              |
                              v
                 commit succeeds or rolls back all flips
                              |
                              v
              finish each partition: update in-memory snapshot,
              swap listing table, clear caches, trigger old-snapshot GC
```

## Partitioned append flow

```text
Input stream
     |
     v
Partition rows by key
     |
     v
+--------------------+   +--------------------+   +--------------------+
| Partition A        |   | Partition B        |   | Partition C        |
| begin_staged_append|   | begin_staged_append|   | begin_staged_append|
| write _staging     |   | write _staging     |   | write _staging     |
| prepare _wal.json  |   | prepare _wal.json  |   | prepare _wal.json  |
+---------+----------+   +---------+----------+   +---------+----------+
          \                      |                      /
           \                     |                     /
            v                    v                    v
              Prepared append receipts + per-partition WALs
                              |
                              v
              sort by table_id; acquire all listing_fence write locks
                              |
                              v
              write top-level _partitioned_wal/<commit_id>.json
                              |
                              v
          +------------------- shared append barrier -------------------+
          | for each partition:                                         |
          |   move staged files -> current snapshot                     |
          |   remove per-partition WAL                                  |
          |   refresh in-memory ListingTable                            |
          +-------------------------------------------------------------+
                              |
                              v
              remove top-level WAL; release fences; finish partitions
```

## Notes

- Overwrite is atomic at the catalog transaction boundary.
- Append uses the listing-fence barrier so scans see either the pre-barrier or post-barrier listing state.
- Automatic replay of the top-level append WAL is not included in this PR.
- The top-level partitioned WAL is local-filesystem only; S3-backed tables keep the existing per-partition WAL behavior.

## Test plan

- [x] `cargo test -p cayenne --features partition-table-provider`
- [x] `cargo fmt --all --check`
- [x] `listing_fence_overhead` criterion benchmark
- [ ] `make lint-rust`
- [ ] CI

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10819"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


lukekim (dismissed):

phillipleblanc:
Addressed all three Copilot comments in the merge commit just pushed:

1. **`apply_under_barrier` fence acquisition** — fixed. The single-partition path now acquires `listing_fence.write()` at the top and holds it across the entire move → WAL removal → listing refresh sequence, then drops it. Uses the same `lock_listing_fence_write_owned` accessor the cross-partition coordinator uses, so the fence semantics are symmetric.

2. **`commit_in_one_txn` retry-on-conflict** — fixed. Added a retry loop bounded by `turso_shared::DEFAULT_CONCURRENT_WRITE_MAX_ATTEMPTS` with `retry_backoff_delay` (same shape as the single-partition `commit_compaction`). On each attempt the loop opens a fresh `MetastoreTransaction`, re-runs every `PreparedOverwrite::apply_in_txn`, and commits. Re-applying is safe and idempotent — the prepared receipts are immutable (their data is already on disk in fresh snapshot directories); only the catalog mutation is repeated. Promoted `is_retryable_write_conflict` to `pub` on the cayenne crate so the runtime-side coordinator can use the same predicate as the in-crate retry sites.

3. **`partitioned_wal` module docs** — fixed. Replaced the "recovery driver added in step 6" language with an explicit statement that the top-level WAL is a diagnostic / manual-recovery anchor today (the existing `ensure_no_incomplete_write` check at table open is what blocks subsequent writes until an operator clears the per-partition WALs). Wiring automated replay-or-rollback is called out as a follow-up.

Also merged trunk to pick up #10811 (`mutation_writer` refactor) and #10816 (join range fallback). The append-side write path now delegates to `AppendMutationWriter`; `scan()` merges trunk's per-scan `ListingTable` construction (for per-scan `target_partitions`) with this PR's `listing_fence.read()` guard. All cayenne tests still pass.

phillipleblanc:
Addressed the second batch of Copilot review comments in 3016fe9 (just pushed):

1. **`partitioned_wal::write_to` atomic write (comment on `partitioned_wal.rs:141`):** WAL is now written via tempfile (`<commit_id>.json.tmp`) + `sync_all` + atomic `rename` + parent-dir `sync_all`. A crash at any point either leaves no WAL or a complete, parseable WAL — never a partial file that `read_all_in` would discard.

2. **Rollback ordering (comments on `staging_wal.rs:220` and the matching `PreparedOverwrite` / `PreparedStagedAppend` paths):** rollback now binds the per-table write guard to `_write_guard` so it lives through `clear_staging_dir()` / staged-snapshot cleanup. A concurrent writer can no longer acquire the lock mid-rollback and transiently observe `IncompleteWrite` / leftover staging state.

3. **Stale module docs (comment on `partitioned_insert_strategy.rs:27`):** module-level docs now describe the actual append flow (the `CayennePartitionedAppendSink` barrier + ordered fence acquisition + top-level WAL anchor + per-partition `apply_under_held_barrier`) instead of the outdated "append falls through to `DefaultInsertStrategy`" wording. The overwrite-flow doc was also tightened to mention the bounded retry on transient commit conflicts.

4. **S3 table roots break `tokio::fs` (comment on `partitioned_insert_strategy.rs:756`):** the append coordinator now checks `table_root.to_string_lossy().starts_with("s3://")` and skips the top-level `PartitionedWal::write_to` / `remove` on S3-backed tables. The crash-safety comment block is updated to call out the resulting gap: per-partition staging WAL still anchors single-partition recovery on S3, but the cross-partition *set anchor* needed to atomically replay/rollback a mid-barrier crash is deferred until the WAL grows object-store IO. The `ensure_no_incomplete_write` per-partition check still blocks any partition whose staging WAL survives, so no half-applied state becomes visible to readers.

`make lint-rust` passed locally before the push.

peasee (dismissed):

peasee (approved):

PR #10823, opened 2026-05-14
Persist Kafka CDC offsets in sidecar tables

lukekim:
## Summary
- persist Kafka and Debezium CDC partition offsets in local `spice_sys_*` sidecar tables for file-backed accelerations
- restore stored offsets when recreating Kafka consumers so local acceleration restarts can resume from the durable sidecar state
- make CDC change commit failures propagate from the refresh task and keep non-durable CDC paths explicitly ephemeral instead of rejecting startup

## Verification
- `cargo fmt --check`
- `cargo check -p runtime --features "kafka debezium dynamodb duckdb sqlite turso postgres-accel"`

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10823"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


phillipleblanc (dismissed):

phillipleblanc (dismissed):

phillipleblanc (dismissed):

phillipleblanc (dismissed):

crates/runtime/src/dataaccelerator/spice_sys/kafka/sqlite.rs
```diff
@@ -96,23 +98,124 @@ impl KafkaSys {
...
+                    conn.query_row(&query, [&dataset_name], |row| row.get(0))?;
+                let offsets_json = serialize_merged_offsets(existing_offsets_json.as_deref(), &new_offsets)
+                    .map_err(|err| {
+                        tracing::warn!("Failed to merge Kafka offsets from SQLite: {err}");
+                        rusqlite::Error::InvalidQuery
+                    })?;
+                let update = format!(
+                    "UPDATE {KAFKA_TABLE_NAME} SET offsets_json = ?1, updated_at = CURRENT_TIMESTAMP WHERE dataset_name = ?2"
+                );
+                let changed = conn.execute(&update, [offsets_json, dataset_name])?;
```
lukekim: **HIGH — Lost-update race on `offsets_json`.** SELECT (line 153) → in-memory merge (154) → UPDATE (162) runs outside a transaction. Today the `pending_commit` JoinHandle in `refresh_task/changes.rs:362` serializes a single connector's writers so the race is dormant, but any second writer on the same `dataset_name` (reload overlap, embedding indexer subscribing to the same dataset, a future parallel commit queue) silently loses progress: two readers see the same old offsets, two writers race to overwrite, the higher offset gets clobbered. On restart the consumer re-reads that window. For Debezium/PG that's a slow no-op (idempotent upsert), for plain Kafka append it duplicates rows.

Mirror code in `kafka/duckdb.rs:117-127`, `kafka/postgres.rs:108-129`, `kafka/turso.rs:114-134`, and the four `debezium_kafka/*` files.

Two clean fixes:
- **Move merge into SQL.** Switch the schema to one row per `(dataset_name, topic, partition)` and use `INSERT ... ON CONFLICT (dataset_name, topic, partition) DO UPDATE SET offset = GREATEST(<table>.offset, EXCLUDED.offset)`. Atomic, idempotent, works across all four backends.
- **Wrap RMW in a transaction.** `BEGIN IMMEDIATE` for sqlite/turso, `BEGIN; SELECT ... FOR UPDATE; UPDATE` for postgres, a transaction for duckdb. Smaller change, but only protects the storage layer — it doesn't make duplicate `KafkaSys` writers across processes safe (file lock dependent).

crates/runtime/src/dataaccelerator/spice_sys/kafka/postgres.rs
```diff
@@ -72,26 +65,105 @@ impl KafkaSys {
...
+            .conn
+            .query_opt(&query, &[&self.dataset_name])
+            .await
+            .map_err(Error::external)?
+            .ok_or_else(|| {
+                Error::external(format!(
+                    "Kafka sidecar metadata for dataset {} does not exist",
+                    self.dataset_name
+                ))
+            })?;
```
lukekim: **MEDIUM — `upsert_offsets` is fatal when the metadata row is missing.** If the prior `upsert(metadata)` failed (disk full, transient PG blip) and is being retried, or if an operator cleared the row, this returns a fatal commit error — but the data batch has *already* been written by `write_change` upstream. On restart, the sidecar still has no offsets, the message is re-delivered and re-applied. For Debezium/PG this is idempotent (upsert by PK); for plain Kafka append it duplicates rows.

Consider doing an `INSERT ... ON CONFLICT (dataset_name) DO UPDATE SET offsets_json = ...` here with nullable metadata columns — offsets always land, even when the metadata row hasn't been written yet, which gives a recoverable starting point. Same pattern in `kafka/duckdb.rs:129-134`, `kafka/sqlite.rs:163-165`, `kafka/turso.rs:136-141`, and the `debezium_kafka/*` mirrors.

crates/runtime/src/dataaccelerator/spice_sys/kafka/turso.rs
```diff
@@ -29,65 +31,159 @@ impl KafkaSys {
...
+            return Ok(true);
+        }
+    }
+    Ok(false)
+}
+
+fn is_duplicate_offsets_column_error(err: &turso::Error) -> bool {
+    let message = err.to_string();
+    message.contains("duplicate column name") && message.contains("offsets_json")
 }
```
lukekim: **LOW — Brittle error-string match.** `err.to_string().contains("duplicate column name")` will silently break if turso changes the wording in a point release. Two ways to harden:

- Pin a regression test that exercises the duplicate-column path against the specific dependency version (so a bump that changes the message text turns red instead of silently re-failing `ensure_kafka_table`).
- Or match the typed extended error code if turso exposes one (rusqlite has `ExtendedCode`; not sure about turso).

Same fragile pattern in `kafka/sqlite.rs:217-218` (matches on a substring inside `rusqlite::Error::SqliteFailure`) and both `debezium_kafka/{sqlite,turso}.rs` mirrors.

crates/runtime/src/dataaccelerator/spice_sys/offsets.rs
```diff
@@ -0,0 +1,83 @@
...
+
+impl OffsetSchemaState {
+    pub(crate) fn needs_ensure(&self) -> bool {
+        !self.ensured.load(Ordering::Acquire)
+    }
+
+    pub(crate) fn mark_ensured(&self) {
+        self.ensured.store(true, Ordering::Release);
+    }
+}
```
lukekim, line 40: **LOW — `OffsetSchemaState` is per-instance, not per-database.** Each `KafkaSys::try_new` constructs its own `OffsetSchemaState`, so multiple `KafkaSys` for the same dataset in one process each re-run `ensure_kafka_table` once. Harmless (DDL is `IF NOT EXISTS`), but the type's name suggests broader 'ensured' semantics than it actually has.

If you want it process-wide, hoist into something like `OnceLock<Mutex<HashSet<(Engine, String)>>>` keyed by (engine, dataset). Not blocking — flagging for design clarity.

crates/runtime/src/dataaccelerator/spice_sys/offsets.rs
```diff
@@ -0,0 +1,83 @@
...
+        .map(|offset| ((offset.topic.clone(), offset.partition), offset))
+        .collect();
+
+    for offset in offsets {
+        merged_offsets
+            .entry((offset.topic.clone(), offset.partition))
+            .and_modify(|existing_offset| {
+                existing_offset.offset = existing_offset.offset.max(offset.offset);
+            })
+            .or_insert_with(|| offset.clone());
```
lukekim: **LOW informational — Silent absorption of backward offsets.** `max(...)` is correct (offsets are monotonic per partition), but consider `tracing::warn!(...)` inside `and_modify` when `offset.offset < existing_offset.offset`. That would surface buggy upstream producers, unexpected out-of-order redelivery, or corrupted state as a clear signal instead of silently merging it away — useful for diagnosing 'consumer is stuck at offset N' reports.

lukekim (commented):
ACID review of the new sidecar CDC persistence. Severities range from HIGH (cross-writer lost-update on `offsets_json`) to informational LOW. Inline comments anchor each finding to one representative location; the same pattern appears in all four backends and both `kafka/` + `debezium_kafka/` sibling modules — the suggested fixes apply uniformly.

One system-level note that doesn't anchor cleanly to a diff line: the pipeline is at-least-once by construction (apply-data → write-sidecar-offsets → commit-broker-offsets, three separate stores with no 2PC). That's fine for Debezium / PG replication because their `Create`/`Update` ops become PK-keyed upserts in `process_upsert_batch`. **Plain Kafka is different**: `wrap_data_as_change_batch` (`crates/data_components/src/cdc.rs:421`) emits `op='c'` with empty `primary_keys`, which routes through `InsertOp::Append` with no dedup. After a crash between data-apply and sidecar-offset-update, the message is re-delivered on restart and appended again → duplicate rows. Worth either documenting the at-least-once contract for plain Kafka or warning at startup when a Kafka dataset has no primary key.

phillipleblanc (dismissed):

krinart (approved):

PR #10833, opened 2026-05-14
feat(geo): add optional spatial SQL UDF support

lukekim:
## Summary

Adds optional PostGIS-style spatial SQL UDF support via the [`geodatafusion`](https://github.com/datafusion-contrib/geodatafusion) crate.

- Adds a new opt-in `geo` Cargo feature on `runtime` and `spiced`.
- Keeps spatial functions unavailable by default, even in `geo` builds, unless `runtime.params.geo: enabled` is set in the Spicepod.
- Registers the `geodatafusion` `ST_*` functions through the existing runtime UDF registration path.
- Adds automated coverage for the runtime gate so `ST_*` functions are registered only when the build feature and runtime parameter are both enabled.

## Sample usage

Build `spiced` with the optional feature:

```bash
cargo build -p spiced --features geo
```

Enable spatial UDF registration in the Spicepod:

```yaml
runtime:
  params:
    geo: enabled
```

Run a spatial SQL query:

```sql
SELECT ST_AsText(ST_Point(0.0, 0.0)) AS geom;
```

Expected result:

```text
+------------+
| geom       |
+------------+
| POINT(0 0) |
+------------+
```

Without `runtime.params.geo: enabled`, the same `ST_*` functions remain unregistered so the default runtime does not add spatial function names or dependency behavior for users who do not opt in.

## Validation

- [x] `RUSTC_WRAPPER= cargo test -p runtime --lib --no-default-features --features geo register_udfs_registers_geo_udfs_only_when_enabled -- --nocapture`
- [x] `make lint`

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10833"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


crates/runtime/src/datafusion/udf.rs
```diff
@@ -156,9 +156,25 @@ pub async fn register_udfs(runtime: &crate::Runtime) {
         );
     }
 
+    #[cfg(feature = "geo")]
+    if geo_enabled(runtime).await {
```
phillipleblanc, line 160: nit: if the param is enabled but the feature is not, it would nice to log a warning.

phillipleblanc (approved):

PR #10858, opened 2026-05-15
refactor: leverage DataFusion TreeNode and split_binary utilities

lukekim:
## Summary

Replace hand-rolled `BinaryExpr` recursion with existing DataFusion utilities in two spots surfaced by an audit of `Expr` traversal in the codebase.

- **`s3_vectors_metadata_filter`** (production): Collapse `supports_filter_expr` + its private `supports_binary_expr` helper into a single `Expr::apply` walk. The traversal recurses into `AND`/`OR` children (`TreeNodeRecursion::Continue`), treats validated comparison / `IS NULL` / `InList` nodes as leaves (`Jump`), and short-circuits on unsupported nodes (`Stop`). Adding a new top-level expression type is now a single match arm.

- **`accelerated_table` deletion tests**: Replace the hand-rolled recursive `collect_or_conditions` / `collect_and_conditions` helpers with a direct call to `datafusion::logical_expr::utils::split_binary`, which does exactly the same flattening and additionally handles `Expr::Alias` transparently.

Net diff: +36 / -47.

## Test plan

- [x] `cargo test -p s3_vectors_metadata_filter --lib` — 11/11 pass, including `test_valid_datafusion_expressions` which exercises every branch of `supports_filter_expr` (AND/OR, all six comparison ops, IS NULL / IS NOT NULL, InList, and nested combinations).
- [x] `cargo clippy -p s3_vectors_metadata_filter --no-deps -- -D warnings` — clean.
- [x] `cargo check -p runtime --lib` — clean (pre-existing warnings only).
- [ ] CI to validate `runtime` test build (couldn't build locally — out of disk in remote env).

## Follow-ups (out of scope)

- `runtime/src/datafusion/refresh_sql.rs:347` also reimplements `split_conjunction` and could call into the upstream utility.
- `datafusion-optimizer-rules/src/physical_plan/duckdb/intermediate_index_cte.rs:72` mimics `split_conjunction` for SQL AST (different `Expr` type — not directly replaceable, but worth noting).

---
_Generated by [Claude Code](https://claude.ai/code/session_015BiBjsLAmSD5cqXnqSkZgG)_

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10858"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task/deletion.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task/deletion.rs
@@ -228,6 +228,7 @@ mod tests {
     use arrow::record_batch::RecordBatch;
     use datafusion::common::ScalarValue;
     use datafusion::logical_expr::Operator;
+    use datafusion::logical_expr::utils::split_binary;
     use std::sync::Arc;
 
     fn make_single_row_batch(pk: i64, sk: &str) -> RecordBatch {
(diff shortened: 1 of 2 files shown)


krinart (approved):

crates/runtime/src/accelerated_table/refresh_task/deletion.rs
```diff
@@ -511,28 +512,12 @@ mod tests {
...
-                conditions
-            }
-            _ => vec![expr],
-        }
+        split_binary(expr, Operator::Or)
     }
 
-    /// Recursively collects all leaf conditions from an AND tree
     fn collect_and_conditions(expr: &Expr) -> Vec<&Expr> {
-        match expr {
```
Jeadie, line 528: we should just use `split_binary(expr, Operator::And)` everywhere?

lukekim: Agreed — I've made the change locally (inlined all 12 call sites and dropped the wrappers), but the PR is already in the merge queue so the push is rejected. I'll land it as a small follow-up PR once this merges.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #10863, opened 2026-05-15
fix(CayennePropagateFilterAcrossEquiJoinKeys): require dim-side stats and fix stats propagation

sgrebnov:
## 📝 Summary

The `CayennePropagateFilterAcrossEquiJoinKeys` logical optimizer rule could fire incorrectly for tables without statistics (e.g., HTTP federated tables), injecting `InSubquery` filters that produced empty results. Separately, `AcceleratedTable` did not delegate `statistics()` to the inner accelerator, so even tables that *do* have stats (Cayenne, DuckDB, Arrow) appeared stats-less to the optimizer.

### Changes

**1. Require dim-side statistics for filter propagation (logical_optimizer.rs)**

`skip_propagation_by_cardinality` now requires dim-side stats to be present. When `subtree_upper_bound_rows` returns `None` (no stats), the rule is skipped. The rule only applies when `TableProvider::statistics()` returns `Some(Statistics)` with `Exact` or `Inexact` row counts.

Previously, `None` stats fell through as "allow", causing slow performance.

**2. Forward `statistics()` through `AcceleratedTable` wrapper (accelerated_table/mod.rs)** to ensure stats is available/propogated

## 👀 Notes for Reviewers

Follow-up work required (separate PR)
1. Rename `CayennePropagateFilterAcrossEquiJoinKeys` to reflect it applies to any accelerated table with stats, not just Cayenne
1. Ensure explicit `AcceleratedTable` type gate to make sure optimization is applied to accelerated tables only

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10863"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/crates/runtime/src/accelerated_table/mod.rs
+++ b/crates/runtime/src/accelerated_table/mod.rs
@@ -1316,6 +1316,10 @@ impl TableProvider for AcceleratedTable {
         self.accelerator.table_type()
     }
 
+    fn statistics(&self) -> Option<datafusion::common::Statistics> {
+        self.accelerator.statistics()
+    }
+
     fn supports_filters_pushdown(
         &self,
         filters: &[&Expr],
--- a/crates/cayenne/src/logical_optimizer.rs
+++ b/crates/cayenne/src/logical_optimizer.rs
@@ -106,8 +106,12 @@ limitations under the License.
 //!   there isn't enough probe-side cardinality for the filter to save
 //!   meaningful work, and the plain hash join wins.
 //!
-//! Both gates only fire when stats are present (`Precision::Exact` or
-//! `Precision::Inexact`); missing stats fall back to the structural behavior.
+//! The dim-side gate requires stats to be present: if the dim side has no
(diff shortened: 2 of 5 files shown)


crates/runtime/src/accelerated_table/mod.rs
```diff
@@ -1316,6 +1316,10 @@ impl TableProvider for AcceleratedTable {
         self.accelerator.table_type()
     }
 
+    fn statistics(&self) -> Option<datafusion::common::Statistics> {
+        self.accelerator.statistics()
```
sgrebnov, line 1320: Should we have something special for `on_zero_results:source`???

lukekim (approved):

PR #10864, opened 2026-05-15
Fix GetFlightInfo vs DoGet Flight Schema

krinart:
## 📝 Summary

Expand Arrow view types (`Utf8View` → `LargeUtf8`, `BinaryView` → `LargeBinary`) in Flight streaming responses so the schema advertised in `GetFlightInfo` matches the data sent in `DoGet`.

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10864"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


crates/runtime/src/flight/mod.rs
```diff
@@ -367,6 +375,36 @@ impl Service {
     }
 }
 
+fn cast_view_columns(
```
Jeadie: should this be in `arrow_tools::schema` too? (i.e. like `expand_views_schema`).

Jeadie (dismissed):

sgrebnov (dismissed):

crates/datafusion-flightsql/src/lib.rs
```diff
@@ -291,7 +291,7 @@ impl FlightSqlService {
                     Ok(batch) => {
                         // Cast view types to their non-view equivalents to match
                         // the expanded schema we advertised in GetFlightInfo.
-                        let batch = cast_view_columns(batch, &schema)
+                        let batch = arrow_tools::schema::cast_view_columns(batch, &schema)
```
sgrebnov, line 294: For curiosity: If we properly use `expand_views_at_output = true` why physical schema is incorrect so we need to cast it?

krinart: Currently `GetFilghtInfo` returns logical plan schema which may not match physical plan schema.

lukekim (approved):

PR #10908, opened 2026-05-18
feat: add dataset-level on_schema_change config

lukekim:
## Summary

Adds a dataset-level `on_schema_change` policy enum for source schema drift. This is intentionally top-level on the dataset, because schema inference and registration apply to both accelerated and non-accelerated datasets.

The user-facing policy values are:

- `block` (default): keep the dataset healthy and continue serving queries using the registered schema; do not apply schema changes automatically.
- `fail`: fail when the projected source schema diverges from the registered dataset schema.
- `append_new_columns`: allow additive source columns, while rejecting removals and incompatible changes.
- `sync_all_columns`: keep the registered dataset schema synchronized with the projected source schema.

This PR establishes the shared config surface and runtime plumbing. Enforcement/evolution behavior can be implemented behind this single policy instead of introducing connector-specific booleans or nested config.

## Example Usage

Default behavior can be omitted or written explicitly:

```yaml
datasets:
  - from: s3://my-bucket/events/
    name: events
    on_schema_change: block
```

Fail loudly on schema drift:

```yaml
datasets:
  - from: postgres:public.orders
    name: orders
    on_schema_change: fail
```

Allow additive columns:

```yaml
datasets:
  - from: s3://my-bucket/events/
    name: events
    on_schema_change: append_new_columns
    acceleration:
      engine: duckdb
      refresh_mode: append
```

Keep the registered schema synchronized with the projected source schema:

```yaml
datasets:
  - from: file://data/events.parquet
    name: events
    on_schema_change: sync_all_columns
```

## Implementation

- Adds `OnSchemaChange::{Block, Fail, AppendNewColumns, SyncAllColumns}` to Spicepod `Dataset` parsing.
- Adds the corresponding runtime dataset enum and builder plumbing.
- Keeps the default conservative/back-compatible as `block`.
- Adds focused YAML parsing tests for the default and all supported values.
- Updates Copilot instructions to prefer enum-shaped user-facing config over booleans, using `on_schema_change` as the example.

## Out Of Scope

- Applying schema changes during refresh/startup.
- Per-engine capability validation for `append_new_columns` or `sync_all_columns`.
- Retiring or redefining `mode: file_update` behavior.
- Surfacing schema evolution history in APIs/task history.

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10908"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


phillipleblanc (approved):
Given these aren't wired up yet, we should emit a warning whenever someone sets a value that isn't the default `block`.

Issue #2936, opened 2024-10-02
Bug: Reading multiple datasets from S3 causes error `Generic S3 error: error decoding response body`


Sevenannn:
## Describe the bug

Configure multiple datasets to read from S3 (to reproduce I used 24 datasets), the error `Generic S3 error: error decoding response body` is shown.
![image](https://github.com/user-attachments/assets/e2b0aac1-c7a2-4ddf-9d24-1db33e05414c)

This is likely to be caused by IO tasks of reading from S3 & intensive CPU bound tasks from Datafusion running in the same runtime, causing IO tasks to be starved, as indicated by this post: https://github.com/apache/arrow-rs-object-store/issues/272

## To Reproduce

Run spice with the following spicepod configuration:
```yaml
version: v1beta1
kind: Spicepod
name: tpcds-sf1.db
datasets:
  - from: s3://spiceai-public-datasets/tpcds/call_center/
    name: call_center
    params:
      file_format: parquet
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
  - from: s3://spiceai-public-datasets/tpcds/catalog_page/
    name: catalog_page
    params:
      file_format: parquet
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
  - from: s3://spiceai-public-datasets/tpcds/catalog_returns/
    name: catalog_returns
    params:
      file_format: parquet
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
  - from: s3://spiceai-public-datasets/tpcds/catalog_sales/
    name: catalog_sales
    params:
      file_format: parquet
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
  - from: s3://spiceai-public-datasets/tpcds/customer/
    name: customer
    params:
      file_format: parquet
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
  - from: s3://spiceai-public-datasets/tpcds/customer_address/
    name: customer_address
    params:
      file_format: parquet
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
  - from: s3://spiceai-public-datasets/tpcds/customer_demographics/
    name: customer_demographics
    params:
      file_format: parquet
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
  - from: s3://spiceai-public-datasets/tpcds/date_dim/
    name: date_dim
    params:
      file_format: parquet
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
[...]

PR #10931, opened 2026-05-19
fix: improve S3 error message for response body decoding failures

### LINKED ISSUE #5980: Bug: S3 connector reports a generic error instead of a specific client timeout
## Describe the bug

When loading a large data source via the S3 connector, usually one that is a large geographical distance away from me, results in the following error message:

```console
2025-05-26T01:11:00.927749Z  WARN runtime::accelerated_table::refresh_task: Failed to load data for dataset tpch.lineitem (s3): Failed to get data from the connector.
External error: Arrow error: External error: External: Generic S3 error: error decoding response body
Ensure the dataset configuration is valid, and try again.
```

This occurs a few times, before eventually succeeding. If the `client_timeout` parameter is increased, the error typically does not occur.

## To Reproduce

1. Configure an S3 connector with a large data source (like TPCH - e.g. https://github.com/spiceai/cookbook/blob/trunk/tpc-h/README.md)
2. Observe intermittent generic error

## Expected behavior

A more descriptive error should be presented. Specifically, and error about the timeout and that the `client_timeout` parameter could be increased to resolve the error.

ManishSharma1609:
When the S3 connector encounters a 'error decoding response body' error, provide a more descriptive error message suggesting the user increase the client_timeout parameter or reduce concurrent dataset loads.

Closes #5980

## Summary
When the S3 connector encounters a generic `error decoding response body` 
error, the current message gives no actionable guidance. This fix intercepts 
that specific error in `handle_object_store_error` and returns a descriptive 
message mentioning `client_timeout` and concurrent dataset load starvation 
(as identified in #2936).

## Related
Closes #5980
References #2936

## Breaking Changes
None, existing error handling is unchanged for all other error types.

## Docs
No doc changes required.

## Notes for Reviewers
Fix is in `handle_object_store_error` in `s3.rs`. Two unit tests added 
verifying the improved error message with and without `client_timeout` set.
For details, visit: https://spiceai.org/docs/components/data-connectors/s3#params

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10931"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


crates/runtime/src/dataconnector/s3.rs
```diff
@@ -363,6 +363,20 @@ impl ListingTableConnector for S3 {
                         connector_component: ConnectorComponent::from(dataset),
                         source: err.into(),
                     }
+                } else if source.to_string().contains("error decoding response body") {
```
phillipleblanc, line 366: This isn't the correct approach for fixing this error message. Object store itself should be updated to tell us whenever it has failed because of a network timeout. Its possible that the object store crate already fixed this and this code won't do anything (or worse, will actively mislead people into doing the wrong thing).

phillipleblanc (changes requested):
This should be fixed in object store and bubbled up to spice with a typed error.

crates/runtime/src/dataconnector/s3.rs
ManishSharma1609: Thanks for the review, I dug into this a bit more and you were right that this really needs to be fixed in object_store rather than in Spice with string matching.

I traced the error path and found that the transport info from reqwest::Error (is_timeout(), etc.) gets flattened into object_store::Error::Generic before it reaches Spice. I also checked the newer retry_stream() logic and it looks like object_store already retries mid-stream body failures using ranged re-requests + etag validation.

So the issue now seems less about retrying and more about preserving the transport classification once retries are exhausted.

I’m going to open an upstream issue/PR in object_store to preserve that timeout/transport information properly, then wire it back into Spice instead of matching on "error decoding response body".

ManishSharma1609: Small update after digging further: I found newer upstream object_store work after the pinned revision Spice is currently using around:

* retrying streaming body failures (`#383`)
* exposing reqwest/hyper error sources (`#554`)
* improving Generic error visibility (`#560`)

So before opening new upstream work, I’m going to check whether upgrading object_store to `0.13.2` already resolves the visibility issue cleanly in Spice.

ManishSharma1609:
> This should be fixed in object store and bubbled up to spice with a typed error.

Took another look at this and ended up reworking it from scratch based on your feedback.

The root issue turned out to be that timeout info was getting lost inside object_store before it ever reached Spice — HttpError was effectively getting swallowed by RetryError due to #[error(transparent)]. Since we can't bump to 0.13.x yet (DataFusion pins us to 0.12.x), I backported commit 378bfe7 onto a temporary 0.12.x fork and also exposed HttpError / HttpErrorKind publicly behind the cloud feature flag.

On the Spice side, handle_object_store_error now walks the error source chain and checks for HttpErrorKind::Timeout via typed downcast, no string matching anywhere.

The error users see now looks like:

Cannot connect to the dataset my_dataset (s3). S3 request timed out (client_timeout: 30s). Consider increasing the client_timeout parameter.

Also added tests for both timeout and non-timeout paths to make sure the behavior stays selective.

Fork is currently at ManishSharma1609/arrow-rs-object-store on branch fix/expose-http-error-source-0.12.x — mainly wanted to validate the approach end-to-end before moving anything upstream or under the Spice org.

phillipleblanc:
> > This should be fixed in object store and bubbled up to spice with a typed error.
> 
> Took another look at this and ended up reworking it from scratch based on your feedback.
> 
> The root issue turned out to be that timeout info was getting lost inside object_store before it ever reached Spice — HttpError was effectively getting swallowed by RetryError due to #[error(transparent)]. Since we can't bump to 0.13.x yet (DataFusion pins us to 0.12.x), I backported commit 378bfe7 onto a temporary 0.12.x fork and also exposed HttpError / HttpErrorKind publicly behind the cloud feature flag.
> 
> On the Spice side, handle_object_store_error now walks the error source chain and checks for HttpErrorKind::Timeout via typed downcast, no string matching anywhere.
> 
> The error users see now looks like:
> 
> Cannot connect to the dataset my_dataset (s3). S3 request timed out (client_timeout: 30s). Consider increasing the client_timeout parameter.
> 
> Also added tests for both timeout and non-timeout paths to make sure the behavior stays selective.
> 
> Fork is currently at ManishSharma1609/arrow-rs-object-store on branch fix/expose-http-error-source-0.12.x — mainly wanted to validate the approach end-to-end before moving anything upstream or under the Spice org.

I'd prefer to wait until we upgrade to a DataFusion version that already includes object_store 0.13 instead of taking a new fork dependency for this. So we can park this PR until that lands and then revisit once we're using object_store 0.13.

lukekim:
👋 Thanks @ManishSharma1609 for your first contribution to spiceai!

Automated new-contributor check:
- Category: ✍️ AI-assisted contribution from a human contributor (High confidence)
- Reason: the PR body carries a Codesmith agent footer, while the account itself shows a long-running, organic public contribution history.

Provenance note, not a problem — a maintainer will review as usual.

PR #10950, opened 2026-05-20
perf(cayenne): reduce allocation overheads in hot paths

lukekim:
Reduces allocation and clone overhead in a few Cayenne hot paths.

## Changes

- Hash new deletion-index keys inline and store hashes for incremental bloom updates, avoiding a `Box<[u8]>` clone per newly inserted key.
- Reuse the cached primary-key `RowConverter` when applying deletion filters instead of rebuilding one from `pk_indices`.
- Build snapshot scan schemas from existing fields to avoid cloning metadata that is overwritten immediately afterward.
- Reuse projected schema field refs and pre-index update assignments by column name.

## Verification

- `cargo fmt -- crates/cayenne/src/provider/table.rs crates/cayenne/src/provider/deletion_index.rs`
- `cargo check -p cayenne`

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10950"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


crates/cayenne/src/provider/delete/vector_io.rs
```diff
@@ -103,10 +106,35 @@ impl DeletionVectorWriteSpec {
...
     ///
     /// The row IDs should be file-local positions (0 to N-1 within the specified file).
+    /// The writer sorts and deduplicates them; use [`Self::new_position_based_sorted`]
+    /// instead when the caller can guarantee monotone-unique input to skip the redundant
+    /// O(N log N) pass.
+    ///
+    /// Currently only the test suite exercises this constructor; production
+    /// code uses the `_sorted` variant.
     #[must_use]
+    #[allow(dead_code)]
```
phillipleblanc: if this is only used by tests, then instead of allowing the dead code, guard this with `#[cfg(test)]`

phillipleblanc (dismissed):

crates/cayenne/src/provider/delete/vector_io.rs
sgrebnov: Updated to #[cfg(test)]

sgrebnov (approved):

PR #10992, opened 2026-05-21
fix(cluster): gate scheduler readiness on executor partition loads

phillipleblanc:
## Summary

Cluster schedulers used to mark accelerated datasets `Ready` as soon as the table registered. Because readiness is sticky after a dataset reaches `Ready`, `/v1/ready` could return 200 before any executor loaded the assigned partitions.

This PR moves scheduler readiness for accelerated datasets to an executor-confirmed path. Executors send `PartitionsLoaded` after the initial partition load and after partition-refresh commands. The scheduler records those acknowledgements and marks the dataset `Ready` only after every assigned partition is covered by at least one assigned executor.

## Flow

```text
Before
------
scheduler registers accelerated dataset
        |
        v
marks dataset Ready ------------------> /v1/ready = 200
        |
        v
executors may still be loading data

After
-----
scheduler registers accelerated dataset
        |
        v
keeps dataset Refreshing -------------> /v1/ready = 503
        |
        v
partition discovery + assignment
        |
        v
executor loads assigned partitions
        |
        v
executor sends PartitionsLoaded
        |
        v
scheduler verifies every assigned partition is covered
        |
        v
marks dataset Ready ------------------> /v1/ready = 200
```

## Changes

- Add `PartitionsLoaded` to `ExecutorControlMessage`.
- Add scheduler-side `PartitionLoadTracker` state for `(table, executor) -> loaded partition bytes`.
- Drop an executor's load acknowledgements when its control stream disconnects, so stale state cannot mask reassignment.
- Sort partition-value keys before serializing partition expressions, making partition bytes stable across scheduler and executor paths.
- Share an `ExecutorOutboundBroadcaster` with the executor runtime so it can send readiness acknowledgements outside the control-stream task.
- Keep scheduler-side accelerated datasets in `Refreshing` until partition-load coverage is complete.
- Wait for datasets to be registered, not ready, during scheduler partition discovery to avoid a startup deadlock.

## Notes

- Scheduler accelerated tables still do not load data locally; readiness now reflects executor data availability.
- Empty but reconciled partition metadata (`updated_at > 0`) is treated as loaded so empty sources do not stay `Refreshing` forever.
- This change creates a chicken-and-egg with the K8s operator's strict scheduler-ready gate for executor admission. The fix lives in the operator: extending its existing per-SpicepodSet BlueGreen update strategy to the SpicepodCluster level with persistent blue/green slot Services. No runtime API extension is needed.

## Test plan

- [x] `cargo check -p runtime --all-targets`
- [x] `cargo check -p runtime-cluster --all-targets`
- [x] `cargo test -p runtime-cluster --lib partition_load_tracker`
- [x] `make lint-rust`
- [x] Manual 1-scheduler/1-executor cluster: `/v1/ready` stayed 503 before executor acks and returned 200 after every assigned partition was loaded.
- [ ] CI

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10992"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1781965864&installation_id=128462479&pr_number=10992&repository=spiceai%2Fspiceai&return_to=https%3A%2F%2Fgithub.com%2Fspiceai%2Fspiceai%2Fpull%2F10992&signature=1943c18254c9f8e0f884038ba5e508316d772bee27143b1b6a2a290f7aa0bae9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>


crates/runtime/src/cluster/control_stream_client.rs
```diff
@@ -73,6 +73,114 @@ pub type PartitionUpdateHandler = Arc<
...
 
+/// Per-scheduler outbound senders, shared between the [`ControlStreamManager`]
+/// and the executor runtime. The manager populates this map on each
+/// (re)connect and clears entries on disconnect; the runtime uses it to
+/// broadcast unsolicited messages (e.g. `PartitionsLoaded`) to every
+/// scheduler the executor is currently connected to.
+///
+/// Cloning is cheap (just `Arc::clone`).
+#[derive(Clone, Debug, Default)]
+pub struct ExecutorOutboundBroadcaster {
```
Jeadie: nit: can be moved to `crates/runtime-cluster`?

crates/runtime/src/cluster/control_stream_client.rs
```diff
@@ -229,6 +342,9 @@ fn spawn_control_stream(
                 heartbeat_task.abort();
                 let mut outbound_guard = outbound_tx_state_for_task.write().await;
```
Jeadie: nit: probably worth bundling to ensure we don't miss any of these on exits

```rust
  let mut outbound_guard = outbound_tx_state_for_task.write().await;
  *outbound_guard = None;
  if let Some(b) = broadcaster.as_ref() {
      b.unregister(&scheduler_address).await;
  }
```

crates/runtime/src/cluster/mod.rs
```diff
@@ -115,13 +115,23 @@ pub enum DistributedNode {
 
         /// Partition service for discovery, assignment, and executor notification.
         partition_service: Arc<PartitionService>,
+
+        /// Tracks which (table, partition) pairs each executor has reported
+        /// as loaded; used to drive accelerated-dataset readiness on the
+        /// scheduler.
```
Jeadie, line 121: we're not considering how a newly joining scheduler initialises this map (i.e. the state of loading from already running executors)?

Which, I think is a reasonable simplfication.

crates/runtime/src/cluster/service.rs
```diff
@@ -910,6 +920,84 @@ async fn notify_scheduler_executor_shutdown(
...
+
+    if tracker.is_table_loaded(&table, &metadata, datafusion).await {
+        tracing::info!(
+            table = %loaded.table_name,
+            "All assigned partitions loaded; marking dataset Ready"
+        );
+        datafusion
+            .runtime_status
+            .update_dataset(&table, crate::status::ComponentStatus::Ready);
+    }
```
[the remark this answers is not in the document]
Jeadie: Yeah this is reasonable. We've had bugs already that we addresssed by defaulting to using ResolvedTableReference

crates/runtime/src/init/dataset.rs
```diff
@@ -1219,13 +1221,66 @@ impl Runtime {
...
             let ds = Arc::clone(&ds);
             let dataset_name = ds.name.to_string();
+            let dataset_table_ref = ds.name.clone();
+            let broadcaster = runtime.executor_outbound_broadcaster();
+            let resolved_name = ds.name.clone().resolve(
+                crate::datafusion::SPICE_DEFAULT_CATALOG,
+                crate::datafusion::SPICE_DEFAULT_SCHEMA,
+            );
             tokio::task::spawn(async move {
-                notifier.notified().await;
```
Jeadie, line 1228: Not related to the initial readiness logic but we also have `notifier.notified().await;` in `async fn reload_accelerated_dataset`.

crates/runtime/src/lib.rs
```diff
@@ -834,11 +853,62 @@ impl Runtime {
...
+        // the same way the scheduler encodes them when sending UpdatePartitions
+        // (`Expr::to_bytes()`). `partition_value_to_bytes` sorts entries by
+        // key so the encoding is deterministic across re-serialization.
+        let partition_expr_bytes: Vec<Vec<u8>> = assignments
+            .get(&table)
+            .map(|exprs| {
+                exprs
+                    .iter()
+                    .filter_map(|e| {
+                        use datafusion_proto::bytes::Serializeable;
```
Jeadie: This is now used in very many places. we should standardise in `runtime-cluster`?
```rust
use datafusion_proto::bytes::Serializeable;
match e.to_bytes() {
    Ok(b) => Some(b.to_vec()),
```

crates/runtime/src/cluster/service.rs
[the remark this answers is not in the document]
phillipleblanc: Fixed in latest push: `handle_partitions_loaded` now resolves the executor-sent table name against Spice defaults before using it as the tracker / metadata-lookup key, and looks up the registered dataset key by resolve-equality before calling `update_dataset` so the original `Refreshing` entry flips to `Ready` instead of a new fully-qualified key being created beside it. The initial-ack path now also broadcasts the canonical `resolved_name` so the scheduler sees one consistent form across both paths.

crates/runtime/src/lib.rs
```diff
@@ -834,11 +853,62 @@ impl Runtime {
...
+                            }
+                        }
+                    })
+                    .collect()
+            })
+            .unwrap_or_default();
+
+        if let Some(broadcaster) = self.executor_outbound_broadcaster()
+            && !partition_expr_bytes.is_empty()
+        {
```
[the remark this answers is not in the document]
phillipleblanc: Fixed: removed the `&& !partition_expr_bytes.is_empty()` gate in `update_partition_refresh_sql` so the executor acks even when the assignment is empty. The scheduler's `updated_at > 0` shortcut in `is_table_loaded` then fires for empty-source / zero-partition datasets.

crates/runtime/src/init/dataset.rs
```diff
@@ -1219,13 +1221,66 @@ impl Runtime {
...
+                            .unwrap_or_default()
+                    };
+                    if !bytes.is_empty() {
+                        let sent = b
+                            .broadcast_partitions_loaded(dataset_name.clone(), bytes)
+                            .await;
+                        tracing::info!(
+                            "Broadcast initial PartitionsLoaded for {dataset_name} to {sent} scheduler(s)"
+                        );
+                    }
```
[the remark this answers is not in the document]
phillipleblanc: Fixed: dropped both the `has_initial_partitions` gate and the inner `!bytes.is_empty()` gate. The initial ack now always fires when there's a broadcaster, and uses the canonical `resolved_name.to_string()` so the scheduler sees the same table form regardless of how the user spelled the dataset name in their spicepod. Empty-source datasets reach `Ready` via the scheduler-side `updated_at > 0` shortcut.

crates/runtime-cluster/src/partition_load_tracker.rs
```diff
@@ -0,0 +1,179 @@
...
+        for (assigned_executors, bytes) in &required {
+            let covered = assigned_executors
+                .iter()
+                .any(|exec| table_map.get(exec).is_some_and(|set| set.contains(bytes)));
+            if !covered {
+                return false;
+            }
+        }
+        true
+    }
```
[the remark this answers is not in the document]
phillipleblanc: Added three tests covering the requested cases: `is_table_loaded_when_every_partition_is_covered_across_executors`, `is_table_loaded_returns_false_when_a_partition_is_uncovered`, and `is_table_loaded_handles_empty_partitions_via_updated_at`. They use a small `ColumnResolver` impl of `PartitionExprResolver` that returns the partition expression as a plain column reference so `partition_value_to_bytes` produces stable bytes without needing a real DataFusion table provider.

crates/runtime/src/cluster/control_stream_client.rs
phillipleblanc: Done: `ExecutorOutboundBroadcaster` now lives in `crates/runtime-cluster/src/outbound_broadcaster.rs` and is re-exported from `runtime::cluster` for existing callers. The struct only depended on `runtime_proto` types, which `runtime-cluster` already pulled in.

crates/runtime/src/cluster/control_stream_client.rs
phillipleblanc: Done: bundled the cleanup into `teardown_scheduler_stream` (heartbeat abort + clear outbound state + unregister from broadcaster). All four exit paths in `spawn_control_stream` (init-msg failure, stream-establish failure, cancellation, regular disconnect) now go through that single helper.

crates/runtime/src/cluster/mod.rs
phillipleblanc: Yes — keeping it as the simplification for now. A scheduler that joins after executors are already running won't have ack state until the next refresh-driven broadcast, so `/v1/ready` stays at 503 on that scheduler until then. Documented this on the `partition_load_tracker` field in `cluster/mod.rs`. Backfilling state on scheduler-join (executor re-broadcast or scheduler-pulled snapshot) is non-trivial and out of scope for the readiness gate; this only affects multi-scheduler HA, not steady-state.

crates/runtime/src/init/dataset.rs
phillipleblanc: Good catch — `reload_accelerated_dataset` has the same edge-triggered `notifier.notified().await` pattern. Leaving that for a follow-up though: it's the hot-reload path (config change → recreate accelerated table), not initial bootstrap. The race window is narrower because the accelerated_table is freshly constructed two lines above and hasn't started its first refresh yet, and the existing `remove_dataset_or_view_schedule` at the top means dataset status isn't tracking it the way bootstrap does. Switching it to `wait_for_dataset_ready` needs a bit more thought re-registration ordering. Will file as a follow-up rather than expand this PR's scope.

crates/runtime/src/lib.rs
phillipleblanc: Done: extracted as `runtime_cluster::encode_partition_exprs(exprs, context)` (in `runtime-cluster/src/metadata.rs` alongside `partition_value_to_bytes`). Both call sites (`update_partition_refresh_sql` in lib.rs and the initial-ack path in init/dataset.rs) now go through it.

lukekim (dismissed):

lukekim (approved):

PR #11028, opened 2026-05-25
Define 'trait QueryEngine' to refactor runtime crate

Jeadie:
## 📝 Summary
The `DataFusion` struct is a core component used throughout the `runtime` crate. Because of this, a signifcant amount of code cannot be restructured and refactored out of `runtime`. This is a significant, if not primary, cause for lengthy compile times.

Most uses of `DataFusion` are for simple SQL operations: query, check table existence, etc. This PR defines `trait QueryEngine` for these, and uses it throughout `runtime` crate. Subsequent PRs can refactor `runtime` accordingly.


crates/runtime-datafusion/src/query_engine.rs
```diff
@@ -0,0 +1,179 @@
...
+use datafusion::common::ParamValues;
+use datafusion::datasource::TableProvider;
+use datafusion::execution::SendableRecordBatchStream;
+use datafusion::prelude::SessionContext;
+use datafusion::sql::TableReference;
+
+use crate::allowlist::ResolvedTableAwareAllowlist;
+
+/// Errors returned by [`QueryEngine`] methods.
+pub type BoxError = Box<dyn std::error::Error + Send + Sync>;
```
phillipleblanc: we should use Snafu errors

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #11030, opened 2026-05-25
Support for streaming in spidapter + add `--scenario` configuration

krinart:
### 1. New `--scenario` parameter with `ScenarioConfig` enum. Example scenarios:

`cayenne.yaml`:
```yaml
compute:
  scp:
    channel: ${SPIDAPTER_CHANNEL:-nightly}
    image_tag: ${SPIDAPTER_IMAGE_TAG:-}
    scheduler_state_location: ${SCHEDULER_STATE_LOCATION:-}
    organization_tag: ${SPIDAPTER_ORGANIZATION_TAG:-}
    query_memory_limit: ${SPIDAPTER_QUERY_MEMORY_LIMIT:-}
    resources:
      app_memory: ${SPIDAPTER_APP_MEMORY_LIMIT:-62Gi}
      app_storage_size_gb: ${SPIDAPTER_APP_STORAGE_SIZE_GB:-~}
      executor_replicas: ${SPIDAPTER_EXECUTOR_REPLICAS:-4}
      executor_memory: ${SPIDAPTER_EXECUTOR_MEMORY_LIMIT:-62Gi}
      executor_storage_size_gb: ${SPIDAPTER_EXECUTOR_STORAGE_SIZE_GB:-~}
      ephemeral_storage_gb: ${SPIDAPTER_EPHEMERAL_STORAGE_LIMIT_GB:-}
acceleration: ${SPICE_ACCELERATION:-cayenne}
source:
  direct:
    cayenne:
      aws_region: ${AWS_REGION:-us-east-1}
      data_dir: ${SPIDAPTER_CAYENNE_DATA_DIR:-}
      metadata_dir: ${SPIDAPTER_CAYENNE_METADATA_DIR:-}
      iceberg_region: ${SPIDAPTER_ICEBERG_REGION:-}
      iceberg_catalog_from: ${SPIDAPTER_ICEBERG_CATALOG_FROM:-}
```

`postgres-wal.yaml`:
```yaml
compute:
  scp:
    channel: ${SPIDAPTER_CHANNEL:-nightly}
    image_tag: ${SPIDAPTER_IMAGE_TAG:-}
    scheduler_state_location: ${SCHEDULER_STATE_LOCATION:-}
    organization_tag: ${SPIDAPTER_ORGANIZATION_TAG:-}
    resources:
      app_memory: ${SPIDAPTER_APP_MEMORY_LIMIT:-64Gi}
      ephemeral_storage_gb: ${SPIDAPTER_EPHEMERAL_STORAGE_LIMIT_GB:-}
acceleration: ${SPICE_ACCELERATION:-cayenne}
source:
  postgres_wal:
    provision:
      ec2:
        subnet_id: ${EC2_SUBNET_ID}
        security_group_id: ${EC2_SECURITY_GROUP_ID}
        ami_id: ${EC2_AMI_ID}
        instance_type: ${EC2_INSTANCE_TYPE:-m5.large}
        disk_size_gb: ${EC2_DISK_SIZE_GB:-100}
        associate_public_ip: true
        iam_instance_profile: ${EC2_IAM_INSTANCE_PROFILE:-}
```

### 2. New `sources/` module

Extracted source-specific spicepod generation and setup into a dedicated module:
- `cayenne`
- `dynamodb`
- `mongodb`
-  `postgres_wal`
- `postgres_debezium`

### 3. New `provision/` module

Extracted all provisioning logic from `stdio_server.rs` into separate submodules: 
- `spice_local` (single-node and cluster)
- `spice_cloud`
- `ec2_postgres`
- `ec2_debezium`
- `ec2_mongodb`

EC2 instances get RAII `Ec2Guard` / `DynamoDbGuard` for reliable teardown.


tools/spidapter/src/args/mod.rs
```diff
@@ -18,11 +18,44 @@ use clap::{Parser, Subcommand, ValueEnum};
...
 
 #[derive(Debug, Clone, Copy, ValueEnum, PartialEq, Eq)]
-pub enum BackendMode {
+pub enum SpiceCompute {
     Scp,
     Local,
 }
 
+#[derive(Debug, Clone, Copy, ValueEnum, PartialEq, Eq)]
+pub enum StorageCompute {
```
phillipleblanc: This is weird - what is a storage compute?

tools/spidapter/src/args/mod.rs
```diff
@@ -18,11 +18,44 @@ use clap::{Parser, Subcommand, ValueEnum};
...
+}
+
+#[derive(Debug, Clone, Copy, ValueEnum, PartialEq, Eq)]
+pub enum AccelerationEngine {
+    Cayenne,
+    Duckdb,
+}
+
+#[derive(Debug, Clone, Copy, ValueEnum, PartialEq, Eq)]
+pub enum Storage {
```
phillipleblanc: Shouldn't this just be something like streaming source? and it wouldn't be `Cayenne`, it would be something like "DirectIngest" or BulkIngest or DML - something that indicates its not coming from a federated changes stream, but is being sent directly to spice via API

tools/spidapter/src/args/mod.rs
phillipleblanc: I think this should just be Ec2 and its an Option<>. it should also be renamed to reflect what it actually is

phillipleblanc (changes requested):
The CLI design needs rework before this merges. We've gone from ~20 flags to ~30+, and most are only valid in specific combinations. This makes it much harder than necessary to both reason about this and actually use it.

This CLI has exactly one consumer. Nobody types these flags (unless you are debugging locally, and then you definitely don't want to be typing these by hand) - they're assembled by bash in run_spicebench.yml, one elif branch per configuration, with the EC2 docker -e plumbing copy-pasted three times (spicebench#251). And each configuration is fixed and nameable: postgres-wal-cayenne, dynamodb-streams-duckdb, etc. The flags are just a lossy re-encoding of that name through argv. Every new source will add more flags here and another bash branch there.

Note these are scenarios, not systems under test - the SUT is Spice via spidapter in all six; what varies is how data flows in (source + acceleration). spicebench#251 currently models them as new system_under_test values, which is the same modeling error showing up on the other side of the protocol (separate feedback there).

Proposal: named scenarios instead of flags. Define each supported configuration as a scenario file shipped in the spidapter image, and make the CLI select one. This makes it much easier to use and structurally prevents invalid combinations at runtime.

`# tools/spidapter/scenarios/postgres-wal-ec2.yaml`
```yaml
compute: scp
channel: nightly
acceleration: ${SPICE_ACCELERATION:-cayenne}
source:
  postgres_wal:
    provision:
      ec2:                      # ec2 spec exists ONLY where it's valid
        subnet_id: ${EC2_SUBNET_ID}
        security_group_id: ${EC2_SECURITY_GROUP_ID}
        ami_id: ${EC2_AMI_ID}
        instance_type: ${EC2_INSTANCE_TYPE:-m5.large}
        associate_public_ip: true
```

`# tools/spidapter/scenarios/postgres-wal-local.yaml`
```yaml
source:
  postgres_wal:
    connect:
      url: ${SPIDAPTER_SOURCE_URL}
```

Deserialize into a typed sum so invalid combinations are structurally unrepresentable rather than runtime-checked:

```yaml
enum SourceConfig {
    Direct,                    // was Storage::Cayenne - it's direct ingest via API, not a changes stream
    PostgresWal(Endpoint<PgConfig>),
    PostgresDebezium { pg: Endpoint<PgConfig>, kafka_brokers: String, connect_url: String },
    DynamoDbStreams { region: String },
    MongoCdc(Endpoint<MongoUri>),
}

enum Endpoint<T> {
    Connect(T),
    ProvisionEc2(Ec2Spec),
}
```

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #11048, opened 2026-05-26
fix: route Turso WAL checkpoint through native turso connection (fixes #10657)

### LINKED ISSUE #10657: Snapshot WAL flush: turso primary file rejects rusqlite open ('file is not a database')
## Problem

`SqliteSnapshotEngine::checkpoint_live` (introduced in #10651 commit 1) opens the live accelerator file with rusqlite and runs `PRAGMA wal_checkpoint(TRUNCATE)` to ensure WAL frames are flushed before `fs::copy` takes the snapshot.

For DuckDB this hook is a no-op; for SQLite it works as expected. For Turso it currently fails with:

```
SQLite snapshot error: Failed to checkpoint SQLite WAL for dataset 'trips' at "/tmp/.../writer.turso": file is not a database
```

The Turso (libsql) on-disk file evidently isn't byte-compatible enough with classic SQLite for rusqlite to open it (or there's a path/mode mismatch).

## Effect

`refresh_mode: snapshot` for Turso accelerators silently loses any uncheckpointed WAL writes (because the snapshot upload then proceeds with a dirty file copy), and the integration test `snapshot_refresh::turso::snapshot_refresh_turso_bootstrap_then_refresh` fails on every run.

Workaround in #10651: the test is `#[ignore]`d with this issue link.

## Fix options

1. Use libsql/turso's own connection layer for the checkpoint pragma instead of rusqlite (likely needs a turso feature in `runtime-acceleration`).
2. Fall back to a turso-specific `SnapshotEngine` that issues `PRAGMA wal_checkpoint(TRUNCATE)` via `turso`/`libsql` rather than rusqlite.
3. Use `fs::copy` of the WAL+SHM sidecars together with the main DB file and let the reader's libsql engine replay on open.

Likely (1)/(2) is correct.

## Repro
- Stack: PR #10651
- Test: `cargo test -p runtime --features turso --test integration_snapshot_refresh -- snapshot_refresh::turso` against real S3 (`AWS_PROFILE=sandbox-admin`)

claudespice:
## Summary

Fixes #10657

`SqliteSnapshotEngine::checkpoint_live` was reused for Turso accelerators on the assumption that libsql's on-disk format was byte-compatible with classic SQLite for read-only opens via `rusqlite`. In practice rusqlite fails to open libsql primary files with `"file is not a database"`, so the checkpoint silently fell back to a no-op and any uncheckpointed WAL frames were lost on snapshot creation. `refresh_mode: snapshot` against Turso was therefore disabled via `#[ignore]` on the integration test.

## Changes

- `runtime-acceleration`'s `turso` feature now pulls in `dep:turso` instead of `dep:rusqlite`; the sqlite-engine module is gated on `feature = "sqlite"` only.
- `TursoSnapshotEngine::checkpoint_live` now opens the live file through `turso::Builder::new_local(..)` and runs `PRAGMA wal_checkpoint(TRUNCATE)` over the libsql-native connection. The `(busy, log, checkpointed)` result row contract is the same as SQLite's, so the existing completeness check (busy == 0 and checkpointed >= log) carries over; an incomplete or busy checkpoint surfaces as `TursoSnapshotError::CheckpointIncomplete` rather than silently snapshotting a partial database.
- Adds a dedicated `TursoSnapshotError` enum and a `SnapshotEngineError::Turso` variant so Turso-specific failures aren't conflated with `SqliteSnapshotError`.
- Un-ignores `snapshot_refresh::turso::snapshot_refresh_turso_bootstrap_then_refresh` so the integration test runs again under the existing Snapshot Refresh Mode Integration Tests CI job.

## Test plan
- [x] Added `checkpoint_live_then_copy_captures_all_rows` regression test (would fail on the previous no-op implementation): writes rows to a file-backed Turso DB, checkpoints, copies just the main file, and asserts the copy still contains every row.
- [x] Added `checkpoint_live_returns_ok_on_empty_database` edge case for the empty-WAL path.
- [x] Un-ignored Turso integration test under Snapshot Refresh Mode Integration Tests.
- [x] `make lint` passes (workspace-wide rustfmt and full clippy gate).
- [x] `make test` passes.

## Checklist
- [x] Regression tests added
- [x] Edge cases covered
- [x] No snapshot deletions
- [x] Lint clean

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/11048"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1782384474&installation_id=128462479&pr_number=11048&repository=spiceai%2Fspiceai&return_to=https%3A%2F%2Fgithub.com%2Fspiceai%2Fspiceai%2Fpull%2F11048&signature=b8401e910a185b167a486be95529997ceac94eff168dc7dfe7148a1281e5a85c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16066,6 +16066,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tracing",
+ "turso",
  "url",
  "util",
  "uuid 1.21.0",
--- a/crates/runtime-acceleration/Cargo.toml
+++ b/crates/runtime-acceleration/Cargo.toml
@@ -40,6 +40,7 @@ tar = "0.4"
 tempfile.workspace = true
 tokio = { workspace = true, features = ["fs"] }
 tracing.workspace = true
+turso = { workspace = true, optional = true }
 url.workspace = true
 util = { path = "../util", features = ["datafusion"] }
 uuid = { version = "1", features = ["v7"] }
@@ -49,7 +50,7 @@ default = []
 duckdb = ["dep:duckdb"]
 snapshots = []
 sqlite = ["dep:rusqlite"]
-turso = ["dep:rusqlite"]
+turso = ["dep:turso"]
 
 [dev-dependencies]
 anyhow = { workspace = true }
--- a/crates/runtime-acceleration/src/snapshot/engine.rs
+++ b/crates/runtime-acceleration/src/snapshot/engine.rs
@@ -26,7 +26,7 @@ mod duckdb;
 #[cfg(feature = "duckdb")]
 pub use duckdb::DuckDBSnapshotEngine;
 
-#[cfg(any(feature = "sqlite", feature = "turso"))]
+#[cfg(feature = "sqlite")]
 mod sqlite;
 #[cfg(feature = "sqlite")]
 pub use sqlite::SqliteSnapshotEngine;
@@ -43,8 +43,12 @@ pub enum SnapshotEngineError {
     DuckDB { source: duckdb::DuckDBSnapshotError },
 
     #[snafu(display("SQLite snapshot error: {source}"))]
-    #[cfg(any(feature = "sqlite", feature = "turso"))]
+    #[cfg(feature = "sqlite")]
     Sqlite { source: sqlite::SqliteSnapshotError },
+
+    #[snafu(display("Turso snapshot error: {source}"))]
+    #[cfg(feature = "turso")]
+    Turso { source: turso::TursoSnapshotError },
 
     /// Placeholder variant for when no snapshot-capable feature is enabled.
     #[snafu(display(
--- a/crates/runtime-acceleration/src/snapshot/engine/turso.rs
+++ b/crates/runtime-acceleration/src/snapshot/engine/turso.rs
@@ -13,31 +13,59 @@ limitations under the License.
 
 //! Turso (libsql) snapshot engine implementation.
 //!
-//! **Status:** WAL flush is currently a no-op for Turso. The first attempt
-//! reused [`super::sqlite::SqliteSnapshotEngine`] under the assumption that
-//! libsql's on-disk format is byte-compatible with classic `SQLite` for
-//! read-only opens via rusqlite. In practice the integration test surfaced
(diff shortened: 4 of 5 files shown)


claudespice:
@copilot review

claudespice:
Status update on the failing `Test Models, AI & Search` check (4th attempt, run 26449001775):

The failure is `models::s3_vectors::search::s3_vectors_data_update` — a snapshot test that expects one row from an s3_vectors search but is getting an empty result set. This test is **pre-existing failure on trunk**, not introduced by this PR:

- `integration tests (models)` workflow is failing on trunk runs for several days (e.g., trunk@f2dd81db32 on 2026-05-26T09:07Z, trunk@fc04e63468 on 2026-05-26T00:42Z, trunk@6ac85a082a on 2026-05-24T00:42Z, all `failure`).
- This PR only touches `runtime-acceleration/src/snapshot/engine{.rs,/turso.rs}` (Turso WAL checkpoint) and `runtime/tests/snapshot_refresh/turso.rs` — none of which intersects s3_vectors search.
- Same failure pattern is hitting other concurrent PRs whose changes are unrelated to s3_vectors.

Rerun via `gh run rerun --failed` (attempt 4) reproduced the same `s3_vectors_data_update` failure, so it's deterministic, not flaky. Per repo PR hygiene, not updating the snapshot in this PR — it's a separate trunk issue.

lukekim (approved):

PR #11061, opened 2026-05-27
fix(tests): update sql tool description snapshots to match #11040

claudespice:
## Summary
- #11040 updated `DEFAULT_DESCRIPTION` in [crates/runtime/src/tools/builtin/sql.rs](crates/runtime/src/tools/builtin/sql.rs) but did not refresh the captured insta snapshots that exercise it via the OpenAI / MCP integration tests.
- As a result, `Test Models, AI & Search` is **failing on trunk** and on every open PR that runs the openai integration test, blocking merge readiness on otherwise green PRs.
- Regenerate the two affected snapshots so the captured tool description matches the live source string.

## Test plan
- [ ] `Test Models, AI & Search` passes on this PR
- [ ] `Test Models, AI & Search` is no longer the only failure on stacked PRs that previously hit this snapshot


lukekim (approved):

claudespice:
@copilot review

PR #11062, opened 2026-05-27
Normalize `DataType::Null` to `Int32` in acceleration schema for duckdb

krinart:
## 📝 Summary

DuckDB has no `Null` type and coerces it to `INT32` when creating a table, but the schema Spice stores retains `DataType::Null`, causing a mismatch on reads. 

Fix normalizes `DataType::Null` → `Int32` before the schema reaches DuckDB.

Plus unit and integration tests.

### DIFF
--- a/crates/runtime/src/dataaccelerator/duckdb.rs
+++ b/crates/runtime/src/dataaccelerator/duckdb.rs
@@ -394,9 +394,9 @@ const PARAMETERS: &[ParameterSpec] = &[
     ParameterSpec::runtime("optimizer_duckdb_aggregate_pushdown"),
 ];
 
-/// DuckDB has no Null type and silently coerces it to INT32. Normalize any
+/// `DuckDB` has no Null type and silently coerces it to INT32. Normalize any
 /// `DataType::Null` fields to `DataType::Int32` so the schema Spice stores
-/// matches what DuckDB actually creates, preventing a mismatch on reads.
+/// matches what `DuckDB` actually creates, preventing a mismatch on reads.
 fn normalize_null_schema(
     schema: &datafusion::common::DFSchemaRef,
 ) -> datafusion::common::Result<datafusion::common::DFSchemaRef> {
--- a/crates/runtime/src/dataaccelerator/duckdb.rs
+++ b/crates/runtime/src/dataaccelerator/duckdb.rs
@@ -2257,15 +2257,21 @@ mod tests {
         super::normalize_schema_for_duckdb(&mut cmd).expect("normalize succeeds");
         let arrow = cmd.schema.as_arrow();
         assert_eq!(
-            arrow.field_with_name("id").unwrap().data_type(),
+            arrow.field_with_name("id").expect("id field").data_type(),
             &DataType::Int64
         );
         assert_eq!(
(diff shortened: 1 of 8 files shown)


crates/runtime/src/dataaccelerator/duckdb.rs
```diff
@@ -392,6 +394,35 @@ const PARAMETERS: &[ParameterSpec] = &[
     ParameterSpec::runtime("optimizer_duckdb_aggregate_pushdown"),
 ];
 
+/// `DuckDB` has no Null type and silently coerces it to INT32. Normalize any
+/// `DataType::Null` fields to `DataType::Int32` so the schema Spice stores
+/// matches what `DuckDB` actually creates, preventing a mismatch on reads.
+fn normalize_null_schema(
```
phillipleblanc: This only handles top-level schemas, nested types like List<>, Struct or Map could have null columns as children that this doesn't handle.

This is duplicating work that `normalize_interval_schema_for_duckdb` already does.

After looking through the codebase a bit, we are doing type normalization in quite a few places, and we should standardize this.

Let's add a trait in `arrow_tools::schema`:

```rust
pub trait TypeRewriteRule: Send + Sync {
    fn name(&self) -> &'static str;
    /// Return `Some(new)` to replace this DataType, `None` to leave it alone.
    /// Children have already been normalized by the walker.
    fn rewrite(&self, dt: &DataType) -> Option<DataType>;
}
```

And a walker in `arrow_tools::schema` that walks using post-order recursion (i.e. into nested types first):

`pub fn apply_rules(schema: SchemaRef, rules: &[&dyn TypeRewriteRule]) -> SchemaRef { ... }`

Then we can add small reusable rules in `arrow_tools::schema` - each one is tiny:

```rust
pub struct DictionaryUnwrap;        // Dictionary(_, v) -> v
pub struct IntervalToMonthDayNano;  // Interval(YearMonth|DayTime) -> Interval(MonthDayNano)
pub struct NullToInt32;             // Null -> Int32
```

Then in each place that needs them, you just add the rules to a list, i.e. for duckdb:

```rust
static DUCKDB_TYPE_RULES: &[&dyn TypeRewriteRule] = &[
    &DictionaryUnwrap,
    &IntervalToMonthDayNano,
    &NullToInt32,
];
```

and now your normalize function becomes:

```rust
fn normalize_schema_for_duckdb(cmd: &mut CreateExternalTable) -> Result<(), DataFusionError> {
    let arrow = cmd.schema.as_arrow();
    let normalized = apply_rules(arrow, DUCKDB_TYPE_RULES);
    if !std::ptr::eq(arrow, &normalized) {
        cmd.schema = Arc::new(DFSchema::try_from(Arc::new(normalized))?);
    }
    Ok(())
}
```

Repeat this pattern for each callsite that needs to rewrite schema. It also means adding new type normalization rules is simple and it will automatically handle nested types (which this PR proves would be an issue for future rules).

phillipleblanc (changes requested):
Needs to be refactored.

crates/runtime/src/dataaccelerator/duckdb.rs
krinart: Done

phillipleblanc (approved):

PR #11066, opened 2026-05-27
Use declared schema in DynamoDB/MongoDB/Debezium

krinart:
## 📝 Summary

Introduce internal `Dataset.schema` field which is calculated based on `Dataset.columns` (user-provided).

Use this schema in DynamoDB, Debezium and MongoDB connectors:
* When no data in upstream to infer schema - use provided schema as-is
* Otherwise merge inferred and declared schema with declared schema taking precedence

This allows tables to become `Ready` even when there's no data in upstream.

### Add tests to cover following cases:

Federated / Accelerated (no streams / cdc)
| Scenario | No schema | With schema
|---|---|---|
| Table doesn't exist | warning + retries | warning + retries
| Table is empty | warning + retries | registered empty table with provided schema

With streams / cdc (`refresh_mode: changes`)
| Scenario | No schema | With schema
|---|---|---|
| Table doesn't exist | warning + retries | warning + retries
| Streams not enabled (DynamoDB only) | warning + retries | warning + retries
| Table is empty | warning + retries | registered empty table with provided schema


crates/data_components/src/debezium/arrow.rs
```diff
@@ -486,10 +486,61 @@ fn rescale_i128(unscaled: i128, src_scale: i8, dst_scale: i8) -> Result<i128> {
     }
 }
 
+/// Convert a plain JSON number (from `decimal.handling.mode=double`) to an `i128`
+/// scaled to `target_scale`.
+///
+/// Parses the number's string representation directly to avoid f64 precision loss.
+/// Scientific notation falls back to f64 rounding.
+fn parse_number_to_decimal(n: &serde_json::Number, target_scale: i8) -> Result<i128> {
```
lukekim: This seems like it should be in a shared utility. Are you sure there is not something already existing?

crates/data_components/src/debezium/arrow.rs
```diff
@@ -486,10 +486,61 @@ fn rescale_i128(unscaled: i128, src_scale: i8, dst_scale: i8) -> Result<i128> {
...
+
+    Ok(if negative { -result } else { result })
+}
+
 /// Parse a decimal from JSON.
 /// Supported inputs:
 /// - JSON string: base64-encoded
 /// - JSON object: {"scale": <int>, "value": <base64>}
+/// - JSON number: plain float (decimal.handling.mode=double)
 pub fn convert_json_to_decimal(v: &Json, target_scale: i8) -> Result<Option<i128>> {
```
lukekim, line 442: Same with this.

krinart: Done

crates/data_components/src/debezium/arrow.rs
krinart: Done

crates/arrow_tools/src/decimal.rs
```diff
@@ -0,0 +1,197 @@
...
+/// Returns `10^exp` as `i128`, or `None` on overflow.
+#[inline]
+#[must_use]
+pub fn pow10_i128(exp: i8) -> Option<i128> {
+    let mut acc: i128 = 1;
+    for _ in 0..exp {
+        acc = acc.checked_mul(10)?;
+    }
+    Some(acc)
+}
```
phillipleblanc: Why is this re-implementing something that already exists in the Rust standard library?

https://docs.rs/num/latest/num/fn.checked_pow.html

crates/data_components/src/kafka.rs
```diff
@@ -526,6 +539,51 @@ impl KafkaConsumer {
...
+        let topic_metadata = metadata
+            .topics()
+            .iter()
+            .find(|t| t.name() == topic)
+            .context(MetadataTopicNotFoundSnafu {
+                topic: topic.to_string(),
+            })?;
+
+        for partition in topic_metadata.partitions() {
+            eprintln!("[kafka] topic_has_messages: fetching watermarks for {topic} partition {}", partition.id());
```
phillipleblanc: Remove these eprintln

phillipleblanc (approved):

PR #11070, opened 2026-05-28
non-blocking partition discovery via Ballista jobs

Jeadie:
## 📝 Summary
In distributed clusters, partition discovery (`SELECT DISTINCT <partition_exprs> FROM <source>`) previously executed synchronously, blocking the scheduler during startup and periodic reconciliation. At scale (e.g. SF100 with dynamic partitioning expressions), this caused repeated timeouts and excess scheduler load.

This PR essentially moves from using synchronous (e.g. `v1/sql`) queries, for asynchronous (e.g. `v1/queries`) queries. Reuses Ballista jobs through the existing `JobExecutor`/`JobStore` machinery.  

Static partitions (e.g. `bucket(N, col)`) are still resolved instantly at startup.
Dynamic partitions are now handled in a 3-step management cycle:
  1. Poll completed discovery jobs and apply diffs
  2. Submit new discovery jobs for tables that need them
  3. Assign pending partitions to executors
  
This was previously implemented in #10503 and never merged.


crates/runtime-cluster/src/context.rs
```diff
@@ -51,9 +51,48 @@ pub trait PartitionDiscoverer: Send + Sync {
...
+/// blocking on a synchronous `SELECT DISTINCT`.
+#[async_trait]
+pub trait PartitionDiscoverySubmitter: Send + Sync {
+    /// Build and submit a discovery job for the given table. Returns the
+    /// Ballista job ID on success.
+    async fn submit_discovery_job(
+        &self,
+        table: &TableReference,
+        partition_by: &[PartitionedBy],
+    ) -> Result<String, Box<dyn std::error::Error + Send + Sync>>;
```
phillipleblanc: Can we tighten this error from being a generic boxed error into a Snafu type? (same for poll_discovery_job)

phillipleblanc (dismissed):

PR #11057, opened 2026-05-27
fix(snowflake): enforce function deny-list in federation pushdown

claudespice:
## Summary
Fixes part of #10703

Snowflake federation silently pushes plans containing Spice-defined UDFs (e.g. `json_get_str`) to the remote engine, which then rejects them at compile time. The runtime's deny-list is enforced inside `SQLExecutor::can_execute_plan`, but the upstream `impl SQLExecutor for SqlTable<T, P>` in `datafusion-table-providers` inherits the trait default (`true` everywhere) — so the Snowflake `SqlTable`, used directly via `create_federated_table_provider()`, never consults `FunctionSupport`.

## Changes
- New `crates/data_components/src/snowflake/federation.rs` adding:
  - `DenyFunctionsSqlExecutor<T, P>` — `SQLExecutor` wrapper around `Arc<SqlTable<T, P>>` that delegates `name`/`compute_context`/`dialect`/`execute`/`table_names`/`get_table_schema` but overrides `can_execute_plan` to reject plans referencing a deny-listed function (same pattern as `crates/data_components/src/turso.rs`).
  - `create_spice_federated_table_provider` helper that mirrors upstream `SqlTable::create_federated_table_provider` but routes the executor through the wrapper.
- `SnowflakeTableFactory::with_function_support` setter and `function_support` field; the factory now uses the wrapper instead of upstream's helper.
- `crates/runtime/src/dataconnector/snowflake.rs` wires `deny_spice_specific_functions()` into the factory at connector construction.

## Scope
This is the Snowflake portion of the per-connector series called out in #10703. ClickHouse, Databricks SQL Warehouse, ScyllaDB, and ODBC follow the same pattern (all delegating to upstream `SqlTable`) and will land as separate PRs to keep this one focused. Spark Connect and FlightSQL have their own `SQLExecutor` impls and need a slightly different fix shape.

## Test plan
- [x] `cargo test -p data_components --features snowflake snowflake::federation::` — 4 new unit tests covering: no-deny-list passes, denied function blocks federation, non-denied function still federates, and the `create_spice_federated_table_provider` adaptor wires both the SqlTable fallback and the federation source schema correctly.
[...]

PR #11071, opened 2026-05-28
fix(snowflake): wire deny-list in extracted connector crate (#10703)

### LINKED ISSUE #10703: Bug. `deny_spice_specific_functions` is not respected by some Spice data connectors
## Describe the bug

Spice defines internally list of functions that should not be pushed down to sources that supports federation, but most connectors does not respect this 

```
/// Return the current combined deny-list: built-ins plus every
/// user-registered function. The returned [`Arc`] is cheap to clone and is
/// safe to use from per-query filter pushdown paths.
#[must_use]
pub fn deny_spice_specific_functions() -> Arc<FunctionSupport> {
    Arc::clone(&*DENY_LIST.read())
}
```

For example Snowflake:

>sql> select json_get_str(L_COMMENT, 'color') AS color from lineitem limit 5;
Query Error: Failed to execute query.
Error executing query: Snowflake API error. Code: `002140`. Message: `SQL compilation error:
Unknown function JSON_GET_STR`

```
sql> explain select json_get_str(L_COMMENT, 'color') AS color from lineitem limit 5;
+---------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
|   plan_type   |                                                                                                                                       plan                                                                                                                                      |
|    varchar    |                                                                                                                                     varchar                                                                                                                                     |
+---------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| logical_plan  | Federated                                                                                                                                                                                                                                                                       |
|               |  Limit: skip=0, fetch=5                                                                                                                                                                                                                                                         |
|               |   Projection: json_get_str(lineitem.L_COMMENT, Utf8("color")) AS color                                                                                                                                                                                                          |
|               |     TableScan: lineitem projection=[L_COMMENT]                                                                                                                                                                                                                                  |
| physical_plan | SchemaCastScanExec                                                                                                                                                                                                                                                              |
|               |   CooperativeExec                                                                                                                                                                                                                                                               |
|               |     BytesProcessedExec                                                                                                                                                                                                                                                          |
|               |       VirtualExecutionPlan name=snowflake compute_context=username=...,account=...,warehouse=COMPUTE_WH,role=accountadmin base_sql=SELECT json_get_str("LINEITEM"."L_COMMENT", 'color') AS "color" FROM "SNOWFLAKE_SAMPLE_DATA"."TPCH_SF1"."LINEITEM" LIMIT 5 |
|               |                                                                                                                                                                                                                                                                                 |
+---------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+

Time: 0.008180584 seconds. 2 rows.
sql> 
```

claudespice:
## Summary
Fixes #10703.

The `connector-snowflake` crate that `bin/spiced` actually wires up was missing the `.with_function_support(deny_spice_specific_functions())` call. As a result, Spice-defined UDFs (`json_get_str`, `cosine_distance`, etc.) were unparsed into Snowflake SQL during federation pushdown and the remote engine rejected them with `Unknown function JSON_GET_STR`.

A previous attempt (#11057) added the wiring to `crates/runtime/src/dataconnector/snowflake.rs`, but that module isn't included in `crates/runtime/src/dataconnector/mod.rs` — so the registration is dead code and the fix never reached the active connector path. The active connector is the extracted crate registered explicitly by `bin/spiced/src/lib.rs`.

## Changes
- `crates/data-connectors/connector-snowflake/src/lib.rs`: wire `deny_spice_specific_functions()` into the `SnowflakeTableFactory` via a small `build_snowflake_table_factory` helper (testable seam).
- `crates/data_components/src/snowflake/mod.rs`: add `SnowflakeTableFactory::function_support()` accessor so connector tests can verify the deny-list is installed without needing a real Snowflake pool.

## Test plan
- [x] Added regression test (`build_snowflake_table_factory_installs_spice_deny_list`) that asserts `json_get_str` and `cosine_distance` are denied while `upper` still federates.
- [x] `make lint` passes (workspace `cargo fmt --check` + full clippy with the workspace deny rules).
- [x] `make test` passes.

## Performance impact
None — the deny-list lookup is a string-set check on already-built plans, identical to the path Snowflake takes for any other federation pushdown decision. The deny-list is a static `Arc<FunctionSupport>` built once at startup.

## Follow-up
This is a partial fix for #10703 — the issue is broader than Snowflake. ClickHouse, Databricks SQL Warehouse, ScyllaDB, ODBC, and the connectors with their own `SQLExecutor` (Spark Connect, FlightSQL) follow the same pattern and need the same wiring; they'll land in separate PRs since they each require different plumbing (e.g. ClickHouse uses `SqlTable::new` directly; MySQL/Postgres/DuckDB connectors use upstream factories that don't currently expose `with_function_support`).

## Checklist
- [x] Root cause identified (orphaned registration in runtime crate; fix needed in the extracted crate that `bin/spiced` actually loads)
- [x] Regression test added
- [x] Edge case (non-denied function still federates) covered
- [x] No snapshot deletions
- [x] Lint clean

### DIFF
--- a/crates/data-connectors/connector-snowflake/src/lib.rs
+++ b/crates/data-connectors/connector-snowflake/src/lib.rs
@@ -24,7 +24,10 @@ limitations under the License.
 //! this crate, not the entire runtime.
 
 use async_trait::async_trait;
-use data_components::snowflake::{SnowflakeTableFactory, quote_snowflake_table_path};
+use data_components::snowflake::{
+    SnowflakeConnectionPool as DynSnowflakeConnectionPool, SnowflakeTableFactory,
+    quote_snowflake_table_path,
+};
 use data_components::{Read, ReadWrite};
 use datafusion::datasource::TableProvider;
 use datafusion_table_providers::sql::db_connection_pool::DbConnectionPool;
@@ -34,6 +37,7 @@ use runtime::dataconnector::{
     ConnectorComponent, ConnectorParams, DataConnector, DataConnectorError, DataConnectorFactory,
     DataConnectorResult, NewDataConnectorResult,
 };
+use runtime::datafusion::udf::deny_spice_specific_functions;
 use runtime::parameters::ParameterSpec;
 use snafu::prelude::*;
 use snowflake_api::SnowflakeApi;
@@ -114,7 +118,7 @@ impl DataConnectorFactory for SnowflakeFactory {
                     .context(UnableToCreateSnowflakeConnectionPoolSnafu)?,
             );
 
-            let table_factory = SnowflakeTableFactory::new(pool);
+            let table_factory = build_snowflake_table_factory(pool);
 
             Ok(Arc::new(Snowflake { table_factory }) as Arc<dyn DataConnector>)
         })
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11075, opened 2026-05-28
Add NSQL context endpoint

lukekim:
Adds a `GET /v1/nsql/context` endpoint that returns the same rendered NSQL context block injected into `POST /v1/nsql`.

Summary:
- Adds the `/v1/nsql/context` route and OpenAPI registration.
- Refactors NSQL context construction so GET and POST share the same rendered block.
- Includes datasets, schemas with metadata/comments/descriptions, Spice/DataFusion/JSON/Spark SQL function context, and registered user-defined functions with signatures.
- Adds query controls for sampling and example rows with limit validation.
- Stabilizes sampled context block ordering for repeatable prompts and prompt-cache compatibility.
- Lists registered Spice search functions from the runtime function inventory.

Compatibility:
- `POST /v1/nsql` continues accepting legacy lowercase request fields `sampledataenabled` and `promptcachekey`.
- The inbound request type is deserialize-only, so Spice does not emit changed canonical request field names.

Validation:
- `cargo fmt --package runtime`
- `CARGO_INCREMENTAL=0 RUSTC_WRAPPER= cargo test -p runtime --features models --lib nsql -- --nocapture`
- `git diff --check -- crates/runtime/src/http/v1/nsql.rs crates/runtime/src/model/nsql.rs crates/runtime/src/http/v1/snapshots/runtime__http__v1__nsql__tests__nsql_context_json_response.snap`
- `CARGO_INCREMENTAL=0 RUSTC_WRAPPER= make lint`
- VS Code diagnostics clean for touched Rust files


crates/runtime/src/http/v1/nsql.rs
```diff
@@ -169,6 +330,633 @@ fn return_sql_only(accept: Option<&TypedHeader<Accept>>) -> bool {
...
+    function_names
+}
+
+struct FunctionContextEntry {
+    name: &'static str,
+    syntax: &'static str,
+    description: &'static str,
+}
+
+fn json_function_entries() -> Vec<FunctionContextEntry> {
```
Jeadie: is there a way of doing this from datafusion context? Otherwise this becomes another list to keep in sync.

crates/runtime/src/http/v1/nsql.rs
```diff
@@ -169,6 +330,633 @@ fn return_sql_only(accept: Option<&TypedHeader<Accept>>) -> bool {
...
+            .state()
+            .table_functions()
+            .keys()
+            .map(|name| name.to_ascii_lowercase()),
+    );
+
+    function_names
+}
+
+struct FunctionContextEntry {
```
Jeadie: Useful to have a `.example` for model use.

crates/runtime/src/http/v1/nsql.rs
```diff
@@ -169,6 +330,633 @@ fn return_sql_only(accept: Option<&TypedHeader<Accept>>) -> bool {
...
+                    format!("Dataset '{dataset}' not found"),
+                ));
+            }
+        }
+    }
+
+    Ok(())
+}
+
+async fn build_nsql_context(
```
Jeadie: ideally we can move much of this logic to `crates/runtime/src/model/nsql.rs`, and keep primarily HTTP logic here.

crates/runtime/src/model/nsql.rs
```diff
@@ -0,0 +1,878 @@
...
+        }
+    }
+
+    fn with_example(mut self, example: &'static str) -> Self {
+        self.example = Some(example);
+        self
+    }
+}
+
+fn json_function_entries() -> Vec<FunctionContextEntry> {
```
phillipleblanc: This gave me an idea that it might be cool to have a "documentation" tool that LLMs can use to get relevant documentation about Spice (including what UDFs it supports).

Jeadie (dismissed):

crates/runtime/src/model/nsql.rs
lukekim: Good idea. I am treating this as follow-up scope rather than part of this endpoint PR; this change keeps the NSQL context/UDF function reference reusable, but a broader documentation tool should be designed separately.

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #11081, opened 2026-05-29
Make 'runtime-query-engine', 'runtime-tools` and 'runtime-search'

Jeadie:
## 📝 Summary
- `runtime-query-engine` is just moving `QueryEngine` from `runtime-datafusion`, so that it can be used more broadly.
- `runtime-search`: Search UDTF arg types and expression builders, candidate generators (VectorUDTFGeneration, TextSearchCandidate, ChunkedNonIndexVectorGeneration), RRF fusion, reranking, EmbeddingColumnConfig, and shared utilities (to_column_expr, closest_column).
- `runtime-tools`: Tooling enum, SpiceToolCatalog trait, SpiceToolsOptions, factory traits, builtin tools (sample, function, datetime), memory tools, MCP server, and the parameters() utility.


crates/runtime-search/src/embeddings/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+use std::sync::Arc;
+
+use chunking::{Chunker, ChunkingConfig};
+use llms::embeddings::{Embed, Error as EmbedError};
+use std::collections::HashMap;
+use tokio::sync::RwLock;
+
+pub type EmbeddingModelStore = HashMap<String, Arc<dyn Embed>>;
+
+pub async fn construct_chunker(
```
Jeadie, line 32: from `crates/runtime/src/embeddings/mod.rs`

phillipleblanc (dismissed):

PR #11084, opened 2026-05-29
fix(postgres-replication): emit recovery log + reduce reconnect-warn volume (fixes #10971)

### LINKED ISSUE #10971: Postgres CDC reconnect loop has no backoff, no recovery log, and duplicates each failed attempt across ERROR+WARN
## Summary

When the Postgres source goes down mid-CDC, the runtime's replication client correctly logs the failure and starts a reconnect loop. Three observability issues with that loop:

1. **No exponential backoff.** ~1 reconnect attempt per second (40 attempts in ~42 s outage). On a 1-hour outage that's ~3600 log lines per dataset.
2. **No "reconnected" / "stream resumed" log line.** When the upstream comes back and the replication slot resumes, the WARN/ERROR stream simply stops. Operators tailing logs have to infer recovery from the absence of failures rather than an explicit signal.
3. **Two log lines per failed attempt.** Pattern is `ERROR pgwire_replication::client::tokio_client: replication worker terminated with error: io error: Connection refused (os error 61)` followed immediately by `WARN data_components::postgres_replication::client: replication recv failed transiently; reconnecting dataset=counters error=io error: Connection refused (os error 61)`. The ERROR variant repeats the same error string as the WARN and adds nothing.

## Version

- Runtime: `v2.0.0-unstable-build.b93578319a+models` (rc.5 candidate).

## Repro

Stand up Postgres with `wal_level = logical` and a basic table, configure spice with `engine: duckdb, refresh_mode: changes, primary_key: id, on_conflict.id: upsert, access: read_write`. Bring spice up, confirm bootstrap. Then `docker stop` the Postgres container (or kill the upstream process). Tail the spice runtime log.

Observed during a ~42 s outage:
```
2026-05-21T00:43:25.236828Z ERROR pgwire_replication::client::tokio_client: replication worker terminated with error: io error: Connection refused (os error 61)
2026-05-21T00:43:25.236846Z  WARN data_components::postgres_replication::client: replication recv failed transiently; reconnecting dataset=counters error=io error: Connection refused (os error 61)
[... 38 more identical pairs ...]
```

Bringing Postgres back: the loop stops. No log line indicates recovery. INSERTs to PG made after recovery flow through correctly within ~5 s, so the WAL stream is genuinely re-attached — the runtime just doesn't say so.

## Suggested fix

1. Exponential backoff with a ceiling — first retry quick, then doubling up to (e.g.) 30 s. This is a standard pattern for replication clients and keeps the log size sublinear in outage duration.
2. Single INFO log line on first failure: `Postgres replication for 'counters' lost; reconnecting (1s)`. Subsequent attempts at the WARN level or DEBUG.
3. Single INFO log line on recovery: `Postgres replication for 'counters' resumed at LSN <x>`. Today this would be a clear signal that the previously logged failures are over.
4. Consolidate the ERROR + WARN pair into one line. The ERROR variant doesn't add diagnostic value over the WARN.

## Notes

- This is observability-only — correctness is fine. The accelerator continues serving queries during the outage, and the replication slot resumes correctly after recovery. The failure mode is operators getting unnecessary alert volume and having no positive signal that recovery happened.
- PR #10929 advertises "HTAP replication diagnostic output for Postgres CDC receive errors and reconnect counts." Reconnect attempts are logged, but there's no aggregated counter (no `cdc_reconnect_attempts_total` or similar in `/metrics`). The "reconnect counts" wording may need to land as either a counter or a single log entry per backoff cycle.

claudespice:
## Summary

Fixes #10971.

When the Postgres source goes down mid-CDC, the runtime's replication client emits a WARN line per failed reconnect attempt and never logs when the stream resumes. On a long outage the WARNs flood the log; operators have to infer recovery from the absence of further WARNs rather than a positive signal.

## Root cause

`postgres_replication::client::wal_stream` already has exponential backoff for transient connect failures, but two observability gaps:

1. No log line on the successful reconnect after one or more transient failures — recovery is silent.
2. Every reconnect attempt logs at WARN, so a 1-hour outage produces 3000+ WARN lines per dataset even though only the first one carries new information.

## Changes

- Track consecutive failed connect/recv attempts per outage cycle.
- Emit a single INFO `replication connection resumed` line on the first successful reconnect after one or more failures, including the attempt count.
- Keep the first transient failure of an outage at WARN (visible at default log levels); drop subsequent retries to DEBUG so a 1-hour outage produces ~1 WARN + 1 INFO on recovery instead of thousands of WARNs.
- Count the initial-connect transient failure (the path through `start_wal_stream`) toward the attempt count so the resume INFO fires correctly even when the upfront connect failed.

The existing exponential `Backoff` is unchanged — this is observability-only; correctness is not affected.

## Test plan
- [ ] Added unit tests covering the log-level threshold transition (`reconnect_first_attempt_logs_at_warn`, `reconnect_subsequent_attempts_drop_to_debug`) and helper robustness across the full u32 attempt range (`reconnect_log_helper_handles_saturating_attempt_count`).
- [ ] `make lint` passes (workspace-wide `cargo fmt --all -- --check` + full clippy, not a per-crate clippy).
- [ ] `make test` passes.

## Checklist
- [ ] Root cause identified and fixed (observability gap, not a backoff bug)
- [ ] Regression tests added for log-level transition
- [ ] Edge cases covered (u32::MAX saturation, no log on first successful connect)
- [ ] No snapshot deletions
- [ ] Lint clean

### DIFF
--- a/crates/data_components/src/postgres_replication/client.rs
+++ b/crates/data_components/src/postgres_replication/client.rs
@@ -132,9 +132,21 @@ fn wal_stream(
 
     try_stream! {
         let mut first_emitted = !mark_ready_on_first;
+        // If the caller passed `None`, the upfront connect failed transiently
+        // (see `start_wal_stream`) — count it as a prior failure so the next
+        // successful connect emits an INFO "resumed" line.
+        let initial_failed = initial_client.is_none();
         let mut client_slot: Option<ReplicationClient> = initial_client;
         let mut backoff = super::resilience::Backoff::default_for_stream();
         let mut last_emitted_commit_lsn = confirmed_flush.load(Ordering::Relaxed);
+        // Counts consecutive failed connect/recv attempts in the current
+        // outage cycle. Reset to 0 on each successful connect. Used to:
+        //   - Demote repeat WARN noise to DEBUG once an outage is established
+        //     (the first failure is still WARN so it's not lost).
+        //   - Emit an INFO "resumed" log on recovery so operators get a
+        //     positive signal — they currently have to infer recovery from
+        //     the absence of further WARNs.
+        let mut reconnect_attempts: u32 = u32::from(initial_failed);
 
         // Outer reconnect loop: runs until we hit a fatal error or the stream
         // reaches a natural end (rare — Postgres replication slots are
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11090, opened 2026-05-30
Align HTAP workflow defaults with automated dispatch

sgrebnov:
## 📝 Summary

Update the `workflow_dispatch` defaults in the HTAP benchmark workflow to match the values used by automated/programmatic dispatch:

- `duration`: `300` → `600` seconds
- `runner_type`: `spiceai-dev-runners` → `spiceai-dev-large-runners`

This keeps manually-triggered runs consistent with scheduled/automated runs.

### DIFF
--- a/.github/workflows/testoperator_run_htap.yml
+++ b/.github/workflows/testoperator_run_htap.yml
@@ -25,7 +25,7 @@ on:
         description: 'Duration of the HTAP workload in seconds'
         required: false
         type: string
-        default: '300'
+        default: '600'
       ready_wait:
         description: 'How long (in seconds) to wait for spiced to start'
         required: true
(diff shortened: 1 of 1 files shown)


.github/workflows/testoperator_run_htap.yml
```diff
@@ -25,7 +25,7 @@ on:
         description: 'Duration of the HTAP workload in seconds'
         required: false
         type: string
-        default: '300'
+        default: '600'
```
sgrebnov, line 28: Keeping this configurable to allow quick verification/experiments runs, for dashboard we will be able to filer only target duration to avoid showing misleading freshness results

lukekim (approved):

PR #11103, opened 2026-05-31
refactor(cluster): build executor stats aggregate via DataFrame API, not SQL string

phillipleblanc:
Follow-up to #11098. local_executor_table_statistics built the per-table
COUNT/min/max/approx_distinct aggregate as a formatted SQL string (with manual
identifier quoting). Build it directly as logical-plan expressions via the
DataFrame API (ctx.table(..).aggregate(vec![], aggr_exprs)) instead — no SQL
text, no identifier-quoting concerns. Behaviour is unchanged: same aggregates in
the same column order, read back positionally.

### DIFF
--- a/crates/runtime/src/cluster/partition/mod.rs
+++ b/crates/runtime/src/cluster/partition/mod.rs
@@ -296,11 +296,6 @@ fn supports_ndv(dt: &datafusion::arrow::datatypes::DataType) -> bool {
     )
 }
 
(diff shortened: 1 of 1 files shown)


lukekim (approved):

phillipleblanc:
Closing in favor of https://github.com/spiceai/spiceai/pull/11104 which removes this code

PR #11110, opened 2026-06-01
perf(cayenne): tiered LSM-style protected-snapshot compaction

sgrebnov:
## 📝 Summary

Adds protected-snapshots LSM-style size-tiered compaction to companion the existing write-lock full-table rewrite compaction. Cayenne now has two complementary compaction paths with clearly separated responsibilities and locking:

1. Protected-snapshot compaction — frequent fast path
LSM-style size-tiered compaction for immutable protected snapshots that runs without taking the per-table `write_lock`. Protected snapshots are immutable, so a subset can be merged via a lock-free compare-and-swap of the snapshot map. This bounds protected-snapshot fan-out during CDC without rewriting the base/current snapshot, keeping read amplification in check while writers continue uninterrupted.

2. Full snapshot compaction — rare maintenance path
The existing full-table rewrite path remains intact as a maintenance / garbage-collection mechanism. It rewrites the full visible table into a clean current snapshot and can retire accumulated delete files, inline records, and protected snapshots.

## Next steps

- Full-compaction trigger params. Add explicit triggers for the maintenance path beyond the background interval: e.g. delete-file count, deleted-PK count / bytes / percentage vs. base table — so full compaction fires on accumulated delete pressure and protected snapshots data, not just elapsed time.
- Bounded fallback for full compaction `try_lock`. Pure `try_lock` can starve full compaction under continuous write pressure (the lock may never be free), letting small files / delete files accumulate unbounded. Follow-up: force a blocking acquire after N consecutive skips, or once a staleness / file-count bound is exceeded, so eventual progress is guaranteed.


lukekim:
Close in favor of https://github.com/spiceai/spiceai/pull/11130

PR #11114, opened 2026-06-02
Remove possibility of a deadlock in `RuntimeStatus`

krinart:
## 📝 Summary

Currently there's a possibility of a deadlock in `RuntimStatus::update_component_status`:
  - Thread A (`update_component_status`): holds `statuses.write` → waits for `notifiers.read`
  - Thread B (`get_or_create_notifier`): holds `notifiers.write` → waits for `statuses.read`

The fix is to replace `statuses` and `notifications` with:
```rust
struct ComponentState {
    status: ComponentStatus,
    notifier: Option<watch::Sender<ComponentStatus>>,
}

statuses: Arc<RwLock<HashMap<String, ComponentState>>>
```

This way we only need to acquire single lock in `update_component_status` / `get_or_create_notifier`

### DIFF
--- a/crates/runtime/src/status.rs
+++ b/crates/runtime/src/status.rs
@@ -105,7 +105,8 @@ impl RuntimeStatus {
     }
 
     /// Updates the status of a component and tracks if it has ever been ready.
-    pub(crate) fn update_component_status(&self, component_name: &str, status: &ComponentStatus) {
+    #[expect(clippy::needless_pass_by_value)]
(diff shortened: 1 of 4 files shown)


crates/runtime/src/status.rs
peasee: as we have now released the lock, the status at the point the notification occurs could be incorrect.

phillipleblanc (dismissed):

phillipleblanc (dismissed):

sgrebnov (approved):

PR #11125, opened 2026-06-02
fix(ci): authenticate snapshot branch updates in integration_search workflow

Jeadie:
## Summary
- add explicit workflow permissions needed for snapshot update automation (`contents: write`, `pull-requests: write`)
- update `.github/actions/push-snap-changes` to set an authenticated `origin` remote using `GH_TOKEN`
- use authenticated `origin` for all fetch/rebase/push operations to avoid non-interactive HTTPS credential prompts

## Why
`integration_search.yml` snapshot update runs can fail with:

`fatal: could not read Username for 'https://github.com': Device not configured`

when git operations hit an unauthenticated HTTPS remote in CI (`persist-credentials: false`). This change makes auth explicit and consistent.


Jeadie:
Closing in favor of #11126, which is the clean branch based on trunk with only the intended two-file CI fix.

PR #11134, opened 2026-06-02
ci: upgrade dorny/paths-filter to v4.0.1 and configure Dependabot for composite actions

lukekim:
## 📝 Summary

Upgrade the `dorny/paths-filter` GitHub Action from v3.0.2 to v4.0.1 (pinned by full commit SHA) and update the Dependabot configuration.

- The action reference lives inside the reusable composite action `.github/actions/check-code-changes/action.yml`.
- v4.0.1 is the latest (released ~Mar 2026); it updates the action runtime to Node 24 (noted as breaking in upstream changelog) and adds merge queue support.
- Basic usage (filters + outputs) is unchanged and fully compatible.

Also updated Dependabot so it will actually propose updates for this (and similar) going forward.

## 🔗 Related

No linked issue (internal maintenance task).

## 🚨 Breaking Changes

None for users/CI behavior. (The major version bump is internal to the action implementation; no input/output or filter syntax changes impact our usage.)

## 📚 Docs

None required.

## 👀 Notes for Reviewers

- Confirmed via `git ls-remote` + release notes that the SHA is the v4.0.1 tag.
- Updated Dependabot `directories` list (supported config) rather than multiple duplicate ecosystem entries.
- This also future-proofs other composite actions that use external actions (e.g. `actions/cache`, `aws-actions/configure-aws-credentials`, `actions/setup-go`, etc.).
- Why Dependabot previously ignored it: the default `directory: '/' ` for the github-actions ecosystem only covers `.github/workflows/` (and root action.yml); composite action manifests require explicit entries. See https://github.com/dependabot/dependabot-core/issues/6704 .

Please review the diff and confirm the SHA pin.

### DIFF
--- a/.github/actions/check-code-changes/action.yml
+++ b/.github/actions/check-code-changes/action.yml
@@ -25,7 +25,7 @@ runs:
   steps:
     - name: Detect all code changes
       id: changed-code-files
-      uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3
+      uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
       with:
         filters: |
           code_changes:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,7 +21,19 @@ updates:
           - '*'
 
   - package-ecosystem: 'github-actions' # See documentation for possible values
-    directory: '/' # Location of package manifests
+    # Note: directory: '/' covers .github/workflows/ (and root action.yml per docs), but composite
+    # actions in .github/actions/*/action.yml require explicit directories entries.
+    # See https://github.com/dependabot/dependabot-core/issues/6704
+    directories:
+      - '/'
+      - '/.github/actions/check-code-changes'
+      - '/.github/actions/setup-adbc-bigquery'
+      - '/.github/actions/setup-nextest'
+      - '/.github/actions/build-testoperator'
+      - '/.github/actions/build-spicepod-validator'
+      - '/.github/actions/build-spidapter'
+      - '/.github/actions/build-spiced'
+      - '/.github/actions/aws-authorize'
     schedule:
       interval: 'weekly'
     labels:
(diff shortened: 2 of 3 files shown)


ewgenius (dismissed):

phillipleblanc (approved):

claudespice:
The **install on Windows x64 (WSL)** failure is a root-cause bug, not a flake:

```
INSTALL_SCRIPT_BRANCH: unbound variable
```

The `install Spice (from branch)` step now passes the branch via the step `env:` block (`INSTALL_SCRIPT_BRANCH`, `HEAD_REPO`, `HEAD_SHA`) and reads it inside `run:`. On the Linux runners that works, but this step uses `shell: wsl-bash {0}` (`Vampire/setup-wsl`). **WSL does not inherit Windows-side process env vars unless they are listed in `WSLENV`**, so inside WSL `INSTALL_SCRIPT_BRANCH` is genuinely unset and the first reference (`[ -n "${INSTALL_SCRIPT_BRANCH}" ]`) trips `set -u`. `HEAD_REPO`/`HEAD_SHA` would hit the same wall on the `else` branch.

Suggested fix (preserves the no-branch-name-interpolation hardening that motivated the env-var approach):

1. Add a `WSLENV` passthrough to the WSL step’s `env:` so the vars cross into WSL:
   ```yaml
   env:
     WSLENV: GITHUB_TOKEN:INSTALL_SCRIPT_BRANCH:HEAD_REPO:HEAD_SHA
     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     INSTALL_SCRIPT_BRANCH: ${{ inputs.install_script_branch }}
     HEAD_REPO: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
     HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
   ```
2. Also guard with a default in the script — `if [ -n "${INSTALL_SCRIPT_BRANCH:-}" ]` — since `WSLENV` does not always export an empty-valued variable (the PR/push case sets it empty), which would otherwise re-trip `set -u`.

Only the WSL `install Spice (from branch)` step needs this; the Linux step is unaffected because Linux `run:` inherits the step env directly.

PR #11145, opened 2026-06-03
fix(duckdb): normalize timestamp columns to microsecond precision (fixes #10627)

### LINKED ISSUE #10627: DuckDB-accelerated `timestamptz` column causes RowConverter schema mismatch (`Timestamp(ns, "UTC")` vs `Timestamp(µs, "UTC")`)
## Description

When a dataset is accelerated with DuckDB and the source column is `timestamptz`, there is a mismatch between the schema spiced registers (`Timestamp(ns, "UTC")`) and what DuckDB physically returns (`Timestamp(µs, "UTC")`). This causes a `RowConverter` panic at query time when DataFusion sorts or range-joins on that column.

**Error (observed in production with a Postgres source):**
```
Arrow error: Invalid argument error: RowConverter column schema mismatch, expected Timestamp(ns, "UTC") got Timestamp(µs, "UTC")
```

**Root cause:** `information_schema.columns` reports the type as `Timestamp(ns, "UTC")`, but actual values returned have 6-digit fractional seconds (e.g. `2025-10-29T09:33:47.511278Z`), confirming DuckDB returns `Timestamp(µs, "UTC")`. DuckDB uses microsecond precision for all `TIMESTAMPTZ` columns regardless of the registered schema.

---

## Reproduction

> **Note:** The CSV file connector manifests the root cause differently — spiced rejects the dataset at initialisation rather than panicking at query time (see below). The `RowConverter` panic reproduces with a Postgres or other `timestamptz` source feeding a DuckDB-accelerated dataset.

### CSV — initialisation failure (related symptom)

**`spicepod.yaml`**
```yaml
version: v1
kind: Spicepod
name: repro-timestamptz

datasets:
  - from: file:events.csv
    name: events
    time_column: created_at
    time_format: timestamptz
    acceleration:
      enabled: true
      engine: duckdb
      mode: memory

  - from: file:refs.csv
    name: refs
    acceleration:
      enabled: true
```

**`events.csv`**
```csv
id,created_at,label
1,2025-10-29T09:33:47.511278+00:00,alpha
2,2025-10-28T14:22:10.123456+00:00,beta
3,2025-10-27T10:05:33.789012+00:00,gamma
4,2025-10-26T08:00:00.000000+00:00,delta
5,2025-10-25T16:45:22.654321+00:00,epsilon
```

**`refs.csv`**
```csv
number,ts
1,2025-10-29T08:00:00.000000000+00:00
2,2025-10-28T12:00:00.000000000+00:00
3,2025-10-27T09:00:00.000000000+00:00
```

**Actual behavior (tested on rc.3 and rc.4):** the dataset never loads — spiced refuses it at initialisation:

```
WARN Failed to setup the dataset events (file). time_column 'created_at' in dataset events
has data type 'Timestamp(µs)', but time_format is configured as 'Timestamptz'
```

This is a different surface of the same root cause: the file connector infers `Timestamp(µs)` for the CSV timestamps, which conflicts with `time_format: timestamptz`.

### Postgres source — query-time `RowConverter` panic (original bug)

Configure `events` as a Postgres dataset with a `timestamptz` column + DuckDB acceleration, then run:

```sql
-- Sort on the DuckDB-accelerated timestamptz column
SELECT * FROM events ORDER BY created_at DESC LIMIT 5;

-- Or a range join against a nanosecond-precision column (e.g. Arrow-accelerated)
SELECT e.id, e.created_at, r.number
FROM (
    SELECT * FROM events ORDER BY created_at DESC LIMIT 10
) e
LEFT JOIN refs r ON r.ts < e.created_at
LIMIT 1;
```

---

## Expected behaviour

Dataset loads successfully and queries return results.

## Actual behaviour

- **File connector:** dataset rejected at initialisation with a `time_column` type mismatch warning.
- **Postgres source:** dataset loads, but queries that sort or range-join on the `timestamptz` column panic with:
  ```
  Arrow error: Invalid argument error: RowConverter column schema mismatch, expected Timestamp(ns, "UTC") got Timestamp(µs, "UTC")
  ```

---

## Additional context

Confirming the type mismatch via `information_schema`:
```sql
-- returns Timestamp(ns, "UTC")
SELECT data_type FROM information_schema.columns
WHERE table_name = 'events' AND column_name = 'created_at';
```

The fix should ensure the schema registered for a DuckDB-accelerated `timestamptz` column uses `Timestamp(µs, "UTC")` to match what DuckDB physically returns.

claudespice:
## Summary

A DuckDB-accelerated column sourced from a `timestamptz` (e.g. a Postgres source) is registered by Spice as `Timestamp(Nanosecond, "UTC")`, but DuckDB's `TIMESTAMP WITH TIME ZONE` (`TIMESTAMPTZ`) type is always **microsecond** precision — DuckDB has no nanosecond-with-timezone type. The registered schema therefore disagrees with what DuckDB physically stores and returns, and any query that sorts or range-joins on the column panics in DataFusion's `RowConverter`:

```
Arrow error: Invalid argument error: RowConverter column schema mismatch, expected Timestamp(ns, "UTC") got Timestamp(µs, "UTC")
```

Fixes #10627

## Root cause

The DuckDB accelerator normalizes its registered schema through `DUCKDB_TYPE_REWRITE_RULES` (Dictionary→value, Interval→MonthDayNano, Null→Int32) to keep the schema in lockstep with what DuckDB physically stores. There was no rule for timezone-aware timestamp precision, so a non-µs `Timestamp(_, Some(tz))` survived into the registered schema while DuckDB stored/returned µs.

## Changes

- Add a `TimestampTzToMicrosecond` rewrite rule in `arrow_tools::type_rewrite` that rewrites a **timezone-aware** `Timestamp(unit, Some(tz))` with a non-microsecond `unit` to `Timestamp(Microsecond, Some(tz))`, preserving the timezone and recursing into nested container types.
- Register the rule in `DUCKDB_TYPE_REWRITE_RULES` so the DuckDB-accelerated table schema for timezone-aware columns is normalized to microsecond precision. Incoming refresh batches are already cast to the accelerator schema via `arrow_tools::record_batch::try_cast_to` (which handles timestamp-unit casts), so no separate data-path change is needed.

### Scope: timezone-aware only

Timezone-**naive** timestamps (`Timestamp(_, None)`) are deliberately left untouched. DuckDB has a native nanosecond `TIMESTAMP_NS` type and preserves the precision of timezone-naive columns — the runtime's own caching-mode `_fetched_at` column is explicitly stored as `TIMESTAMP_NS` (`crates/runtime/src/dataaccelerator/spice_sys/caching_engine/duckdb.rs`). Normalizing those to microsecond would *introduce* a mismatch rather than fix one. Only `TIMESTAMPTZ` lacks a nanosecond variant in DuckDB, so the fix is scoped to timezone-aware timestamps — the exact class the bug affects. Other accelerator engines are unaffected (Cayenne already normalizes timestamps to microsecond in its own schema path).

## Test plan
- [x] Added regression test `duckdb_timestamptz_roundtrip_sorts_without_rowconverter_panic` that registers a `Timestamp(Nanosecond, "UTC")` column, accelerates it with DuckDB, inserts rows, and runs `ORDER BY` (the operation that triggered the RowConverter panic) — asserts the registered schema is microsecond and rows come back in the correct order
- [x] Added `normalize_schema_for_duckdb_timestamptz_to_microsecond` unit test asserting the tz-aware column normalizes to µs and the tz-naive column stays nanosecond
- [x] Added `arrow_tools` unit tests for the rule: tz-aware ns/second/millisecond→µs, tz-aware µs no-op, **tz-naive left unchanged**, nested in List/Struct (tz-aware normalizes, tz-naive does not)
- [x] Updated the existing `retention::test_duckdb_append_refresh_preserves_timestamptz` integration test (Postgres→DuckDB append refresh of a `TIMESTAMPTZ` column) to expect microsecond precision and compare microsecond-aligned values
- [x] `make lint` passes (workspace-wide `cargo fmt --all -- --check` + full clippy)
- [x] `make test` passes

## Checklist
- [x] `make lint` clean (Rust Lint CI green)
- [x] `make test` / integration tests pass (incl. retention timestamptz test)
- [x] Regression test exercises the actual `RowConverter` failure mode (ORDER BY on the timestamptz column)

### DIFF
--- a/crates/arrow_tools/src/type_rewrite.rs
+++ b/crates/arrow_tools/src/type_rewrite.rs
@@ -64,28 +64,34 @@ impl TypeRewriteRule for IntervalToMonthDayNano {
     }
 }
 
-/// Rewrites `DataType::Timestamp(unit, tz)` with a non-microsecond `unit` →
-/// `DataType::Timestamp(Microsecond, tz)`, preserving the timezone.
+/// Rewrites a timezone-aware `DataType::Timestamp(unit, Some(tz))` with a
+/// non-microsecond `unit` → `DataType::Timestamp(Microsecond, Some(tz))`,
+/// preserving the timezone. Timezone-naive timestamps (`tz = None`) are left
+/// unchanged.
 ///
-/// `DuckDB`'s `TIMESTAMP` and `TIMESTAMP WITH TIME ZONE` types are always
-/// microsecond-precision. The Arrow→`DuckDB` column-type mapping in
-/// `datafusion-table-providers` collapses every `Timestamp` to `TIMESTAMP` /
-/// `TIMESTAMPTZ` regardless of the source `TimeUnit`, so a column registered as
-/// `Timestamp(Nanosecond, "UTC")` (e.g. a Postgres `timestamptz` source) is
-/// physically stored — and read back — as `Timestamp(Microsecond, "UTC")`. When
-/// the registered schema keeps the original non-µs unit, `DataFusion` plans
-/// against nanoseconds while `DuckDB` returns microseconds and queries that sort
-/// or range-join on the column panic with
+/// `DuckDB`'s `TIMESTAMP WITH TIME ZONE` (`TIMESTAMPTZ`) type is always
+/// microsecond-precision — `DuckDB` has no nanosecond-with-timezone type. So a
+/// column registered as `Timestamp(Nanosecond, "UTC")` (e.g. a Postgres
+/// `timestamptz` source) is physically stored — and read back — as
+/// `Timestamp(Microsecond, "UTC")`. When the registered schema keeps the
+/// original non-µs unit, `DataFusion` plans against nanoseconds while `DuckDB`
+/// returns microseconds and queries that sort or range-join on the column panic
+/// with
 /// `RowConverter column schema mismatch, expected Timestamp(ns, "UTC") got Timestamp(µs, "UTC")`.
 /// Normalizing the unit to microsecond keeps the registered schema in lockstep
 /// with what `DuckDB` stores and returns.
-pub struct TimestampToMicrosecond;
-impl TypeRewriteRule for TimestampToMicrosecond {
+///
+/// Timezone-naive timestamps are deliberately excluded: `DuckDB` has a native
+/// nanosecond `TIMESTAMP_NS` type and preserves the precision of `TIMESTAMP`
+/// columns without a zone (the runtime's internal `_fetched_at` caching column
+/// relies on this), so normalizing them would instead *introduce* a mismatch.
+pub struct TimestampTzToMicrosecond;
+impl TypeRewriteRule for TimestampTzToMicrosecond {
     fn rewrite(&self, dt: &DataType) -> Option<DataType> {
         match dt {
-            DataType::Timestamp(unit, tz) if *unit != TimeUnit::Microsecond => {
-                Some(DataType::Timestamp(TimeUnit::Microsecond, tz.clone()))
-            }
+            DataType::Timestamp(unit, Some(tz)) if *unit != TimeUnit::Microsecond => Some(
+                DataType::Timestamp(TimeUnit::Microsecond, Some(Arc::clone(tz))),
+            ),
             _ => None,
         }
     }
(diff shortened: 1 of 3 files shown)


claudespice:
@copilot review

lukekim (dismissed):

lukekim (approved):

PR #11146, opened 2026-06-03
fix(cayenne): prevent duplicate PKs during concurrent reads and writes

sgrebnov:
## 📝 Summary

Closes two races where a scan could observe **two rows for one primary key** while a concurrent writer was publishing an upsert.

### Window 1 — deletions + protected-snapshots (file-based upsert)

On a file-based append-upsert, the writer (1) applied the on-conflict delete to the in-memory deletion cache and (2) inserted a _protected snapshot_ for the newly written data — as two separate stores. A scan that captured the deletion view between (1) and (2), or the protected map between (2) and (1), could apply a stale (smaller) deletion set yet still union in the  freshly protected snapshot - surfacing both the old and the new version of a key.

### Window 2 — inlined-data + deletion-view (inline upsert)

On the inline upsert path, the new row's copy was committed durably to the catalog (step 1) before the old file backed copy was hidden in the deletion cache. A scan reading inlined data live from the catalog in that window
saw the new inlined row while the superseded file row was still visible: two rows for one PK.

## Fix 

Common pattern: all durable I/O happens first (lock-free); only the synchronous in-memory visibility flips run under `scan_state_lock.write()`, so the guard is held for microseconds and never across metastore I/O. Scans capture the whole visibility triple under `scan_state_lock.read()`

### Scan side

- `scan()` captures `(deletion_snapshot, protected_map, inlined_batches)` as one tuple under `scan_state_lock.read()`.
- The inlined cache is warmed before taking the lock, so the read under the lock is a cheap cache hit (one atomic load + `Arc::clone`) rather than a metastore round-trip + Arrow IPC decode.
- `scan_protected_snapshots` now receives the protected map captured by the caller instead of re-loading it, so it matches the captured deletion view.

### Window 1 fix — deferred deletion publish + protected snapshot

- `apply_on_conflict_deletions` performs all durable deletion-vector I/O but returns the computed in-memory deletion-cache update (`OnConflictUpdate`) instead of storing it.
- `record_written_snapshot_sequence` durably records the new snapshot's sequence but no longer publishes the in-memory protected entry.
- `commit_on_conflict_publish` then stores the deletion-cache update  and the protected-snapshot entry together under one `scan_state_lock.write()`. The pure-append case (no protected snapshot, no deletion change) skips the guard entirely.

### Window 2 fix — published-sequence watermark

- Add `published_inlined_seq: Arc<AtomicI64>` visibility watermark, seeded from `current_sequence_number` at open.
- `commit_inlined_mutation` now returns the sequence assigned to appended inlined rows (via `UPDATE … RETURNING`).
- `populate_inlined_cache` skips entries with `sequence_number > watermark`, hiding not-yet-published rows.
- The inline path splits durable commit (`commit_inlined_data_durable`) from in-memory publish (`publish_inlined_mutation`); under `scan_state_lock.write()` it advances the watermark (`fetch_max`, before the generation bump) together with the file deletion-cache flip — so revealing the new inlined row and hiding the old file row publish as one atomic step.
- Writers are serialized by `write_lock`, so the scalar monotonic watermark is safe.

## Tests

- `test_concurrent_upsert_scan_no_duplicate_pk` (`_inline` + `_no_inline`) — both pass; exercises the concurrent read/write race for both paths.

## 👀 Notes for Reviewers
- deletions + protected-snapshots migh be a good candidate for ArcSwap (can be done next)
- **Inlined - inlined upsert** (overwriting a key whose live copy is itself inlined) can have a residual transient missing-row window: the old copy is scrubbed by a durable in-place rewrite at commit while the new copy is watermark-gated until publish, so the two halves aren't atomic. This can be improved next to defer the old entry's durable removal and flip an in-memory inlined mask under the write lock alongside the watermark


crates/cayenne/src/provider/table.rs
```diff
@@ -10563,26 +9875,39 @@ impl TableProvider for CayenneTableProvider {
...
-        let deletion_snapshot = self.pk_deletion_snapshot();
+        // Capture the (deletion view, protected snapshot map, inlined data)
+        // triple atomically under `scan_state_lock.read()`
+        let (deletion_snapshot, protected_map, inlined_batches) = {
+            let _view_guard = self.scan_state_lock.read().await;
+            // Read any inlined data while the guard is held so it is consistent
+            // with the captured deletion view. The warm-up above makes this a
+            // cache hit in the common case.
+            let inlined_batches = if self.cached_inlined_row_count() > 0 {
+                self.read_inlined_batches().await.map_err(|e| {
```
sgrebnov: This can still lead to cache miss so we will need to re-build cache under lock. Can me improved next. The window between cache warm up and load is small..

crates/cayenne/src/provider/table.rs
```diff
@@ -1344,6 +1293,29 @@ pub struct CayenneTableProvider {
...
+    ///   together with the deletion-cache update, so reveal-new/hide-old is one
+    ///   unit.
+    ///
+    /// Scans take **read** while capturing the `(deletion_snapshot,
+    /// protected_snapshots, inlined_batches)` tuple (a few `ArcSwap`/atomic
+    /// loads plus a cache-hit inlined read); writers take **write** around the
+    /// matching publish. Read–read never contends and no metastore I/O runs
+    /// under the lock. Distinct from `listing_fence` (orders file moves against
+    /// scans, #10125) — different state, do not conflate.
+    scan_state_lock: Arc<tokio::sync::RwLock<()>>,
```
sgrebnov, line 1370: I also like `visibility_lock` as a name but it is already taken/in use

phillipleblanc (dismissed):

PR #11147, opened 2026-06-03
Fixes for SMB Connector

krinart:
### 1. Wrong NTLM authenticate flags

`build_authenticate_message` was echoing all challenge flags back verbatim (`challenge.negotiate_flags | VERSION`). The flags are now masked to only the capabilities the client actually advertised in its Negotiate message. 

### 2. Wrong preauth integrity hash state for signing key derivation

The signing key for SMB 3.1.1 is derived via SP800-108 KDF over the preauth integrity hash. The code was incorrectly including the final `STATUS_SUCCESS` SESSION_SETUP response in the hash before derivation. Empirical testing against Samba 4.x confirmed the server finalises the hash after the client's AUTHENTICATE request (H5), before sending the response — so the `STATUS_SUCCESS` response must not be included.

### 3. Object store listing returned paths without the share prefix

`list_all_files` and `list_directory_shallow` were normalising the incoming prefix (e.g. `"data"` → `""`) for SMB server operations, but then building ObjectMeta paths using the normalised (empty) base. DataFusion's `list_all_files` filters results by the original prefix, so `"sample.parquet"` was silently dropped because it doesn't start with `"data/"`. The fix uses the caller's display path as the ObjectMeta location base while still using the normalised path for actual SMB I/O.

### 4. Unit and integration tests

### DIFF
--- a/crates/runtime-object-store/src/store/smb.rs
+++ b/crates/runtime-object-store/src/store/smb.rs
@@ -302,14 +302,18 @@ impl SMBObjectStore {
         let share = self.get_share().await?;
         let config = self.config();
         let prefix_str = prefix.unwrap_or_default();
-        let normalized = config.normalize_subpath(&prefix_str).to_string();
 
         let mut results = Vec::new();
-        let mut queue = vec![normalized];
+        // Queue holds display paths (caller's perspective, e.g. "data/subdir").
+        // Each iteration normalizes to the server-side path via normalize_subpath
+        // so SMB operations strip the share prefix while returned ObjectMeta
+        // paths keep it — matching what the caller (DataFusion) expects.
+        let mut queue = vec![prefix_str];
 
-        while let Some(dir_path) = queue.pop() {
-            let entries = Self::list_dir_entries(&share, config, &dir_path).await?;
-            let (files, dirs) = process_directory_entries(&dir_path, entries);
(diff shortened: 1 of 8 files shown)


krinart:
Need more investigation https://github.com/spiceai/spiceai/issues/11148

claudespice:
Pushed a small lint fix (40d8d39) to unblock **Rust Lint**: `build_authenticate_message` declared the `SUPPORTED` const after `let` statements, tripping `-D clippy::items_after_statements` (pedantic). Hoisted the const to the top of the function — behavior unchanged.

The remaining **Integration Tests (part 2)** failure looks functional rather than infra and needs your domain knowledge: `smb::smb_integration_test` runs but the runtime never reaches ready because the dataset fails to set up with:

```
ERROR runtime::init::dataset: Cannot setup the dataset smb_test (smb) with an invalid configuration.
Failed to find any files matching the extension '.parquet'. Spice could not find any files with extensions at the specified path.
```

The Samba container pulls and the session establishes, but the SMB list/query-directory path returns no `.parquet` files at the configured path (all 4 retries time out at ~125s). Could be the listing logic in `runtime-object-store/src/store/smb.rs` / the new query-directory code, or the test fixture not seeding the share. Leaving this one for you since it's core to the connector changes in this PR.

PR #11160, opened 2026-06-04
fix(cluster): prevent DoPut write pipeline self-deadlock under ingest backpressure

phillipleblanc:
## Problem

Under sustained cluster ingest (observed with spicebench TPC-H SF10 changes-mode, 3 executors, `bulk_ingest_upsert`), executor write pipelines intermittently froze permanently: zero CPU, every thread parked, no errors, no timeouts. In a milder variant the scheduler-side forwarding instead failed with `ADBC stream bulk ingest failed … Deadline expired` after 120s. Once one executor froze, h2 connection-level flow control froze every other table stream to that executor, and the whole cluster's ETL stalled indefinitely.

## Root cause: poll-starvation self-deadlock in `create_response_stream`

The DoPut write pipeline is a single task. `write_future` (`df.write_streaming_data(...)`, driven by DataFusion `collect()` — nothing spawns it) is polled **only** through one `tokio::select!` arm, and the sink consumes `batch_rx` only while that future is polled:

1. The sink hits a transient slow await (e.g. a metastore SQLite WAL checkpoint once the catalog grows to hundreds of MB under changes-mode ingest).
2. The wire keeps delivering; the cap-100 `batch_tx` channel fills.
3. The handler suspends at `handle_record_batch → batch_tx.send().await` **inside the select branch** — the generator is no longer at the `select!`, so `write_future` can never be polled again.
4. The sink's wakeup arrives, polls the generator, which is suspended at the send — still no channel capacity — `Pending`. The sink's readiness is lost forever.

Producer waits on consumer; consumer is only polled by the producer's task. Permanent lost-wakeup deadlock. The idle deadline can't fire either — its select arm is starved by the same suspension, which is why the hang produced no timeout.

## Fix

In the message arm, race the batch send against `&mut write_future` **and** the idle deadline (nested `select!`):

- While a send is pending, the sink keeps being polled and drains the channel — backpressure propagates instead of deadlocking.
- If the sink is genuinely stuck, the idle deadline now fires and fails the stream with `deadline_exceeded` instead of hanging forever.
- Sink completion/error during a pending send is handled identically to the existing early-completion/error arms.

Also adds a 30s heartbeat log while post-stream finalization (`write_future` after stream end) is pending, so a stuck finalization is visible instead of silent (this path previously awaited with no timeout and no logging).

## Validation

Repro: local 3-executor cluster, spicebench TPC-H SF10 `data-gen-mutable` (changes), concurrency 2, `bulk_ingest_upsert`.

- Before: 3/3 runs froze at ETL steps 5–7 (two permanent hangs incl. one 45+ min with all threads parked; one 120s-deadline bulk-ingest failure). Diagnosis confirmed with scheduler-side instrumentation showing forwarding encoders blocked 600s+ with `keepalives=0` while one executor stopped draining all four table streams.
- After: the same workload rides through the previously fatal window with no blocked-encoder events, no timeouts; checkpoint validation converges (run in progress at PR time; will confirm full completion in comments).

## Future work

- Consider spawning `write_future` as a real task so the pipeline is not single-task by construction.
- Scheduler-side `write_through` forwarding cannot send keepalives while blocked on a full channel (starving the executor's idle deadline); with this fix the executor self-recovers, but the encoder loop could be restructured similarly for defense in depth.

### DIFF
--- a/crates/runtime/src/flight/do_put.rs
+++ b/crates/runtime/src/flight/do_put.rs
@@ -386,10 +386,65 @@ fn create_response_stream(
                                 }
                             };
 
-                            // Only report errors; a success message is sent as the final step upon successful write completion
-                            if let Err(err) = handle_record_batch(new_batch, &batch_tx, &path.to_string()).await {
-                                yield Err(err);
-                                break;
+                            // Only report errors; a success message is sent as the final step upon successful write completion.
+                            //
+                            // The send must race against polling `write_future`:
+                            // `write_future` is an unspawned future driven solely by
+                            // this loop, and the sink consumes `batch_rx` only while
+                            // it is polled. Awaiting the send directly suspends this
+                            // generator inside the branch when the channel is full,
+                            // so the sink could never be polled again to drain it —
+                            // a permanent lost-wakeup deadlock that froze all writes
+                            // to an executor whenever the sink fell one channel's
+                            // worth of batches behind the wire (e.g. during a slow
+                            // metastore WAL checkpoint under heavy ingest). Keep the
+                            // sink and the idle deadline polled while the send is
+                            // pending so backpressure propagates instead of
+                            // deadlocking.
+                            let path_str = path.to_string();
+                            let batch_send = handle_record_batch(new_batch, &batch_tx, &path_str);
+                            let mut batch_send = std::pin::pin!(batch_send);
+                            tokio::select! {
+                                biased;
+                                () = &mut deadline => {
+                                    tracing::error!(
+                                        dataset = %path,
+                                        "Timeout: write sink did not accept a record batch within {} seconds",
+                                        idle_timeout.as_secs()
+                                    );
+                                    yield Err(Status::deadline_exceeded(format!(
+                                        "Timeout: write sink did not accept a record batch within {} seconds",
+                                        idle_timeout.as_secs()
+                                    )));
+                                    break;
+                                }
+                                write_result = &mut write_future => {
+                                    match write_result {
+                                        Ok(()) => {
+                                            // Sink finished while a batch was still pending —
+                                            // mirror the early-completion arm: drain and succeed.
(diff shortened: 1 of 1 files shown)


phillipleblanc:
## Local validation: ✅ success

Same repro workload as the analysis (TPC-H SF10 `data-gen-mutable` changes-mode, 3 local executors, concurrency 2, `bulk_ingest_upsert`):

- **Benchmark completed (outcome: success)** — Checkpoint 0 converged (E2E 1.7s), Checkpoint 1 converged (E2E 7.0s); changes phase processed 44.7M rows across all steps with no stalls.
- **The trigger condition occurred 19 times during the run** (instrumented build: scheduler forwarding encoder blocked 10–20s with the executor's batch channel full — the exact state that previously deadlocked permanently) — and self-resolved within seconds every time because the sink keeps getting polled while a send is pending.
- 0 `deadline_exceeded`, 0 sink-stall timeouts, 0 stuck finalizations.

Before the fix on the same machine/workload: 3/3 runs froze at ETL steps 5–7 (two permanent zero-CPU hangs, one 120s-deadline bulk-ingest failure).

Cloud validation run (changes-mode, SF10, image `doput-deadlock-fix.2` from this branch) is queued and will be reported here.

sgrebnov (approved):

phillipleblanc:
## Cloud validation: ✅ success

[Spice Cloud changes-mode run](https://github.com/spiceai/spicebench/actions/runs/26940582118) (prod, TPC-H SF10 changes, 3 executors, 96Gi, 2 query clients, `bulk_ingest_upsert`, private cluster) against image `doput-deadlock-fix.2` built from this branch (rebased on trunk @ 2c366009):

- **Benchmark completed (outcome: success)**
- Checkpoint 0 converged (E2E 2.4s), Checkpoint 1 converged (E2E 3.6s)

Both local and cloud validations green.

PR #11161, opened 2026-06-04
fix(http): preserve non-JSON HTTP response rows in nested column decomposition (fixes #11155)

### LINKED ISSUE #11155: Querying http connector without filters causes internal error
http/https dataset with defined `columns`:
```yaml
datasets:
  - from: https://api.tvmaze.com
    name: tvmaze
    check_availability: disabled
    columns:
      - name: details
        metadata:
          json_object: "*"
    params:
      file_format: json
      allowed_request_paths: '/shows/**,/search/people'
      request_query_filters: enabled
```

Causes an error when queried without filters:

```sql
sql> select * from tvmaze;
Internal Error: Unexpected internal error. Use '.error' for details.
sql> .error
An internal error occurred. Failed to decompose HTTP response row into declared columns: Failed to parse HTTP response row as JSON: expected value at line 1 column 1
```

When `columns` are **not defined** a single row is returned:
```sql
sql> select count(*) from tvmaze;
+----------+
| count(*) |
|   int64  |
+----------+
| 1        |
+----------+
```

claudespice:
## Summary

Querying an HTTP/HTTPS dataset that declares `columns:` with a `json_object: "*"` catch-all crashed with an internal error whenever the endpoint returned a non-JSON body:

```
Internal Error: Unexpected internal error.
Failed to decompose HTTP response row into declared columns: Failed to parse HTTP response row as JSON: expected value at line 1 column 1
```

This reproduces by querying without filters (e.g. `select * from tvmaze`), which fetches the base URL with no path and commonly returns an empty body or a non-JSON page.

## Root cause

`crates/data_components/src/http/json_nest.rs::decompose_json_row` called `serde_json::from_str(json_row)?` and hard-errored on any row that was not valid JSON. The non-nested code path (`HttpExec::parse_content`) already tolerates non-JSON bodies — it returns the raw body as a single `content` row — so a dataset *without* `columns:` returned one row while the *same* endpoint *with* `columns:` aborted the entire query.

## Changes

- `decompose_json_row` now treats a parse failure as a non-fatal, lossless case (consistent with its existing handling of non-object valid JSON): every declared static field resolves to `NULL`, and the catch-all column receives the raw row text encoded as a JSON string (`NULL` when the body is empty/whitespace). This mirrors the non-nested path and the function's documented "preserves data without silently dropping values" contract.
- Removed the now-unused `Error::JsonParse` variant.

Scope: limited to the HTTP connector's nested-column decomposition. No behavior change for datasets that return valid JSON, and no change to the non-nested path.

## Test plan
- [x] Added regression test for non-JSON / empty / malformed-JSON rows at the `decompose_json_row` unit level (`non_json_row_is_preserved_in_catchall`, `empty_row_yields_all_null`, `malformed_json_is_preserved_not_dropped`)
- [x] Added regression test at the batch-builder level (`create_batch_from_rows_nested_non_json_rows_are_preserved`) exercising the actual failure mode through `create_batch_from_rows_nested`
- [x] `make test` passes

Fixes #11155

### DIFF
--- a/crates/data_components/src/http/json_nest.rs
+++ b/crates/data_components/src/http/json_nest.rs
@@ -52,9 +52,6 @@ use std::collections::{BTreeMap, HashMap, HashSet};
 
 #[derive(Debug, Snafu)]
 pub enum Error {
-    #[snafu(display("Failed to parse HTTP response row as JSON: {source}"))]
-    JsonParse { source: serde_json::Error },
-
     #[snafu(display("Failed to serialize catch-all JSON column: {source}"))]
     JsonSerialize { source: serde_json::Error },
 }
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11163, opened 2026-06-04
feat(chbench): run scheduled benchmarks for both SF1 and SF10

sgrebnov:
## 📝 Summary

Spread the scheduled HTAP CH-benCHmark runs across separate cron times by scale factor instead of firing them all on the single 0900 UTC job.

A `Resolve schedule name` step maps the cron that fired to a named schedule, and each step gates on it via `if: steps.sched.outputs.name == '...'`:

| Cron (UTC) | Schedule name | Runs |
|------------|---------------|------|
| `0 9 * * *`  | `daily-main`  | full bench suite + HTAP SF1 |
| `0 12 * * *` | `htap-sf10`   | HTAP SF10 (new) |
| `0 16 * * *` | `htap-sf100`  | HTAP SF100 (new) |

### DIFF
--- a/.github/workflows/testoperator_dispatch.yml
+++ b/.github/workflows/testoperator_dispatch.yml
@@ -212,6 +212,14 @@ jobs:
           SPICED_COMMIT: ${{ steps.setup-spiced.outputs.SPICED_COMMIT }}
           WORKFLOW_COMMIT: ${{ matrix.branch }}
 
+      - name: Dispatch Testoperator - Scheduled CH-BenCHmark - SF10
+        run: |
+          testoperator dispatch ./tools/testoperator/dispatch/chbench/sf10 --workflow htap
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/testoperator_dispatch.yml
```diff
@@ -212,6 +212,14 @@ jobs:
           SPICED_COMMIT: ${{ steps.setup-spiced.outputs.SPICED_COMMIT }}
           WORKFLOW_COMMIT: ${{ matrix.branch }}
 
+      - name: Dispatch Testoperator - Scheduled CH-BenCHmark - SF10
```
lukekim, line 249: We should also run SF-100 on schedule.

lukekim: We should move the different SF to different schedules so they don't contend.

sgrebnov: Added separate time/schedule for SF10 and SF100

lukekim (approved):

PR #11178, opened 2026-06-05
style(cayenne): backtick SwissTable/SipHash in deletion-index docs

lukekim:
Follow-up to #11177 (merged moments before this fix landed): backticks five `clippy::doc_markdown` findings (`SwissTable` x3, `SipHash` x2) in `deletion_index.rs` docs that fail the CI pedantic lint on `lukim/olap`. Verified clean locally against the CI clippy flag set (`-Dclippy::pedantic` etc.) for both `hash-index` and `cayenne`.

### DIFF
--- a/crates/cayenne/Cargo.toml
+++ b/crates/cayenne/Cargo.toml
@@ -100,6 +100,10 @@ name = "deletion_index_probe"
 
 [[bench]]
 harness = false
+name = "deletion_index_probe_large"
+
+[[bench]]
+harness = false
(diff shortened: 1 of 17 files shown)


lukekim:
Superseded: both commits landed directly on `lukim/olap` (#11170) at 7052351b7 — the five `clippy::doc_markdown` backtick fixes plus documentation of the measured `DeletionIndex::get` inlining trade-off — since #11177 merged before this follow-up could.

PR #11184, opened 2026-06-05
fix(debezium): sign-extend negative base64 decimals instead of zero-padding

claudespice:
## Problem

Debezium serializes `DECIMAL`/`NUMERIC` columns in its default **precise** mode (and in `VariableScaleDecimal` objects) as the minimal-width, big-endian, two's-complement byte representation of a Java `BigInteger` (`BigInteger.toByteArray()`). The unscaled value is encoded in the fewest bytes needed, **including the sign bit**.

`arrow_tools::decimal::decode_base64_decimal` left-padded those bytes to 16 with `0x00` before calling `i128::from_be_bytes`. That is correct only for non-negative values. A negative value arrives with the sign bit set in its most-significant byte — e.g. `-1` is the single byte `0xFF` — and zero-padding to `[0x00, …, 0x00, 0xFF]` decodes as **255**, not `-1`.

Result: **every negative decimal flowing through the Debezium CDC connector was silently corrupted into a large positive value** (the default `precise` mode is the common path). The bug also affected the `VariableScaleDecimal` object path, which decodes the same way.

This code was recently consolidated into the shared `arrow_tools::decimal` module by #11066; the fix lands in that canonical location.

## Fix

Sign-extend to 16 bytes using the high bit of the leading byte (`0xFF` when set, `0x00` otherwise) instead of always zero-padding. Inputs wider than 16 bytes are rejected explicitly (previously a wider-than-16-byte value also failed via `try_into`, so that behavior is unchanged).

## Why the existing tests missed it

The existing decimal tests built their base64 input with `i128::to_be_bytes()`, which always emits a full 16-byte array — so the padding branch was never exercised with a short, negative encoding. Added unit tests feed minimal-width encodings for both signs, the object (`{scale, value}`) path, the full-width case, and the over-16-byte rejection.

## Test plan
- [x] Added regression test for negative minimal-width base64 decimals (sign extension) at both the string and object (`VariableScaleDecimal`) paths
- [x] Added edge case tests for positive minimal-width (zero extension), full 16-byte encodings, and over-16-byte rejection
- [x] `make lint` passes (workspace-wide fmt + clippy)
- [x] `make test` passes

### DIFF
--- a/crates/arrow_tools/src/decimal.rs
+++ b/crates/arrow_tools/src/decimal.rs
@@ -31,7 +31,7 @@ pub enum Error {
     #[snafu(display("Failed to decode base64-encoded decimal: {source}"))]
     Base64Decode { source: base64::DecodeError },
 
-    #[snafu(display("Decimal bytes must be exactly 16, got {}", value.len()))]
+    #[snafu(display("Decimal value exceeds the 16-byte `i128` range, got {} bytes", value.len()))]
     BytesLength { value: Vec<u8> },
 
     #[snafu(display("Missing `scale` field in decimal object"))]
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11195, opened 2026-06-06
feat(secrets): add full-fidelity reference iteration and a resolution-status API

phillipleblanc:
## Summary

Two library additions to `runtime-secrets` that upcoming preflight/status surfaces need, plus a log-accuracy fix in spidapter.

### 1. `iter_secret_references()` + `SecretReference`

`extract_secret_references()` keys its result by secret key, so a key referenced through two stores (`${ env:K }` and `${ secrets:K }`) silently keeps only the last occurrence's store — its own unit test documents the loss. The new iterator yields **every** occurrence in document order; the old function becomes a thin wrapper with the caveat documented.

### 2. `Secrets::check_reference(store, key) → RefStatus`

Answers *'would this reference resolve?'* without exposing the value:

| Status | Meaning |
|---|---|
| `Found { store }` | resolves; names the answering store (incl. which store satisfied a `secrets:` sentinel walk) |
| `NotFound { searched }` | no healthy store holds the key |
| `StoreError { store, error }` | lookup unanswerable; for stores that failed to load, `error` is the captured init root cause (#11181) |
| `UnknownStore { configured }` | the store isn't configured |

Statuses carry store/key names and error text only — **never secret values** — so they're safe to log or serialize. The sentinel walk mirrors `get_secret`'s skip-unhealthy semantics (#11175); any value fetched to answer the question is dropped (zeroized) before returning.

### 3. spidapter log-accuracy fix

`set_spicepod_secrets` now collects key → *all* referencing stores via the iterator, so its per-secret log lines read `(referenced via: env, secrets)` instead of potentially naming the wrong store.

## Testing

- 6 new unit tests: duplicate-preserving iteration, every `RefStatus` arm (incl. sentinel walk naming the answering store past an unhealthy store), and the #11181 composition (checking a failed-init store reports the init root cause)
- 1 new doctest on `iter_secret_references`
- `extract_secret_references` tests pass unmodified (wrapper preserves last-wins behavior)
- spidapter: 4 module tests updated to the multi-store map
- `cargo test -p runtime-secrets`: 81 passed; `make lint-rust` clean; `cargo doc` 0 warnings

## Series

Fourth PR in the secret-store UX series (#11175 → #11181 → #11188 → this). Next: startup secret preflight summary rendering these statuses with component provenance.


phillipleblanc:
Closing this temporarily — the secret-store UX workstream is paused, so closing to avoid taking up review time. The branch `phillip/secrets-ref-status-api` is retained; this will be reopened when the workstream resumes (it's the 4th in the series after #11175, #11181, #11188, all merged). No code issues — CI was green and review was clean.

bjchambers (approved):

PR #11214, opened 2026-06-07
fix: apply retention_sql after Arrow refresh writes

### LINKED ISSUE #10983: Bug. `arrow` acceleration should apply `retention_sql` during write path
## Describe the bug

Arrow is the only acceleration engine that does not apply `retention_sql` as part of the data write path. DuckDB executes the retention DELETE inside the write transaction (via `WriteCompletionHandler`), and Cayenne runs it as post-write maintenance. Arrow relies solely on a separate timer-based loop (`start_retention_check`), which silently does nothing if `retention_check_interval` is not configured.

This inconsistency means Arrow users must always set `retention_check_interval` alongside `retention_sql`, while DuckDB and Cayenne work without it

mag1cfrog:
## 📝 Summary

Applies `retention_sql` immediately after successful Arrow and PartitionedArrow refresh writes.

Previously, Arrow accelerators could parse `retention_sql`, but rows matching the retention predicate were not removed from the local accelerated table until periodic retention ran. This updates the refresh write path to apply the same retention delete expression after `insert_into`, while keeping periodic retention behavior unchanged.

If the insert succeeds but retention fails, the refresh now returns a permanent error instead of retrying the whole write. That avoids duplicate append risk while surfacing that the accelerated table is not in the configured retained state.

## 🔗 Related

Closes #10983

## 🚨 Breaking Changes

None.

## 📚 Docs

None.

## 👀 Notes for Reviewers

The post-write retention path is currently enabled only for Arrow and PartitionedArrow. DuckDB and Cayenne keep their existing engine-specific retention behavior.

The retention failure path updates committed-write bookkeeping before returning a permanent refresh error, so a committed append is not retried and duplicated. Cached query results are also invalidated when the accelerator changed, including the case where retention fails after a successful write.

### DIFF
--- a/crates/runtime/src/accelerated_table/mod.rs
+++ b/crates/runtime/src/accelerated_table/mod.rs
@@ -147,6 +147,15 @@ pub enum Error {
     FailedToWriteData { source: DataFusionError },
 
     #[snafu(display(
+        "Failed to apply retention_sql to accelerated dataset {dataset_name} after refresh data was written: {}. The accelerated dataset may contain rows that should have been retained away.",
+        format_datafusion_error(source)
+    ))]
+    FailedToApplyRetentionSql {
+        dataset_name: String,
+        source: DataFusionError,
+    },
+
+    #[snafu(display(
         "The accelerated table does not support delete operations. Use a different acceleration engine which supports delete operations. For details, visit: https://spiceai.org/docs/components/data-accelerators"
     ))]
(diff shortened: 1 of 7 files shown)


mag1cfrog:
Looks like I can't assign this myself. Could a maintainer assign it?

lukekim (approved):

PR #11229, opened 2026-06-08
fix(embeddings): non-nullable inner field for embedding FixedSizeList (fixes #6911)

### LINKED ISSUE #6911: Embedding FixedSizeList<float> schema should not have nullable members
## Describe the bug

```
Finding root of DataFusionError: External(DataRetrievalError { source: ArrowError(InvalidArgumentError("column types must match schema types, expected FixedSizeList(Field { name: \"item\", data_type: Float32, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }, 256) but found FixedSizeList(Field { name: \"item\", data_type: Float32, nullable: false, dict_id: 0, dict_is_ordered: false, metadata: {} }, 256) at column index 5"), None) })
```

It should not be possible to have a null float value in an embedding vector. Reconcile and fix this schema difference.

https://github.com/spiceai/spiceai/pull/6846#discussion_r2296872352

claudespice:
## Summary

Embedding vector columns never contain null components, but the scalar-unchunked embedding path declared the inner `Float32` `item` field of the `FixedSizeList<Float32, D>` embedding column as **nullable**, while every other path (chunked, multi-vector) and the accelerator/index schemas (`index/duckdb.rs`, `index/elasticsearch`) declare it **non-nullable**.

That asymmetry produced a schema mismatch on the accelerator write path:

```
column types must match schema types, expected
FixedSizeList(Field { name: "item", data_type: Float32, nullable: true, ... }, 256)
but found
FixedSizeList(Field { name: "item", data_type: Float32, nullable: false, ... }, 256)
```

Fixes #6911.

## Root cause

- `EmbeddingTable::embedding_fields` declared the scalar-unchunked embedding column's inner item as `Field::new("item", Float32, true)`.
- `get_vectors` and `get_vectors_in_process` built the `FixedSizeListArray` with a nullable inner field and called `append_nulls(vector_length)` for null/empty source rows — putting nulls *inside* the vector.

## Changes

- Inner `item` field is now non-nullable in the schema declaration and both builders, matching the chunked path, the multi-vector path, and the DuckDB/Elasticsearch index schemas.
- Null/empty source rows are encoded **only** on the outer `FixedSizeList` slot (null row → null vector). The fixed-stride inner values for a null row are placeholder zeros (masked by the outer null), exactly as `build_multi_vector_list_array` already does.
- Extracted a shared `append_null_vector` helper for the two scalar builders.
- The outer `FixedSizeList` field stays nullable, so a null source row still maps to a null vector — no row-level behavior change.

## Test plan
- [x] Added regression test `test_get_vectors_inner_field_non_nullable` (inner field non-nullable; null rows on outer slot only; inner values buffer has zero nulls)
- [x] Added regression test `test_embedding_fields_scalar_inner_item_non_nullable` (schema: nullable outer, non-nullable inner)
- [x] `make lint` passes (workspace-wide format + clippy gate, not a scoped per-crate invocation)
- [x] `make test` passes

### DIFF
--- a/crates/runtime/src/embeddings/execution_plan.rs
+++ b/crates/runtime/src/embeddings/execution_plan.rs
@@ -348,6 +348,23 @@ pub(crate) async fn compute_additional_embedding_columns(
     Ok(embed_arrays)
 }
 
+/// Append a null entry to a `FixedSizeList<Float32>` builder.
+///
+/// The inner Float32 field is non-nullable (an embedding vector never holds
+/// null components), so the fixed-stride placeholder slots are filled with
+/// zeros and only the outer list slot is marked null. The placeholder values
+/// are masked by the parent null and never read at query time.
+#[expect(clippy::cast_sign_loss)]
+fn append_null_vector(
+    builder: &mut FixedSizeListBuilder<PrimitiveBuilder<Float32Type>>,
+    vector_length: i32,
+) {
+    for _ in 0..vector_length as usize {
+        builder.values().append_value(0.0);
+    }
+    builder.append(false);
+}
+
 /// Embed a [`StringArray`] using the provided [`Embed`] model. The output is a [`FixedSizeListArray`],
 /// where each [`String`] gets embedded into a single [`f32`] vector.
 ///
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

PR #11231, opened 2026-06-08
chore(chbench): add cayenne[file] adaptive + in-memory-CDC variants and schedule them

lukekim:
## What

Adds two new postgres→Cayenne CH-benCHmark spicepods, both derived from the non-tuned `accelerated/postgres-cayenne[file].yaml` baseline, and wires them into the scheduled `testoperator dispatch` HTAP runs.

### New spicepods (`test/spicepods/chbench/accelerated/`)

- **`postgres-cayenne[file]-adaptive.yaml`** — every Cayenne table runs `cayenne_tuning: adaptive`: the runtime warm-starts each knob from the detected host + inferred schema, then runs the closed-feedback loop that adapts them over time within the environment-derived `[floor, ceiling]`. `schema_inference: extended` does double duty: (1) it supplies the inferred `row_count`/`table_bytes` the adaptive loop's data-aware warm-start gates on (without it adaptive silently falls back to static `auto`), and (2) it auto-detects the **primary key, secondary indexes, and sort columns** from the PostgreSQL source and seeds the **upsert `on_conflict`** the CDC (`changes`) path requires. So PK / on_conflict / indexes are intentionally **not** declared — `apply_inferred_schema` fills them (gap-filling only, never overrides). No per-knob value is set: an explicit knob would *pin* it under adaptive (collapsing its tuning bounds to a point), so building off the un-pinned baseline gives the controller full freedom.
- **`postgres-cayenne[file]-cdc-memory.yaml`** — the 7 mutated TPC-C tables (`refresh_mode: changes`) run `cayenne_cdc_durability: memory`: each CDC batch is appended to an in-RAM tier and the replication-slot ack is deferred to a periodic / cap-triggered checkpoint, collapsing the per-batch durability cost. Bounded by a per-table byte cap + a process-global byte budget (spills rather than OOMing); crash-replay from the slot is exactly-once (PK-idempotent apply). The static full-refresh reference tables stay on `file` — the CDC profile does not apply to them, so the param is set only on the changes tables (explicit PKs kept, since this config does not enable extended inference).

Both keep `sql_results` / `task_history` off to match the baseline and the `-cdc-tuned-*` configs, so a comparison isolates the effect of the variant, not caching/history overhead.

### Scheduled dispatch (`tools/testoperator/dispatch/chbench/`)

Adds dispatch entries for both variants at **sf1 / sf10 / sf100**, all at **10K req/s** (matching the `-cdc-tuned` workload; the per-scale runner type / terminals / duration / ready_wait mirror the baseline). The chbench dispatch globs the directory, so the daily-main / 4-hourly CH-benCHmark crons in `testoperator_dispatch.yml` pick them up automatically.

## Validation

- `spicepod-validator` loads both new spicepods cleanly (exit 0), matching the baseline — including the adaptive config with PK/on_conflict omitted (PK is optional at load time; inference applies it at runtime init).
- YAML anchor/merge resolution verified: in `-adaptive` all 11 datasets carry `schema_inference: extended` + `cayenne_tuning: adaptive` and **no** PK/on_conflict/indexes; in `-cdc-memory` the 7 changes tables carry `cayenne_cdc_durability: memory` and the 4 full-refresh tables carry none.

### DIFF
--- a/tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-adaptive.yaml
+++ b/tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-adaptive.yaml
@@ -1,0 +1,8 @@
+tests:
+  htap:
+    spicepod_path: accelerated/postgres-cayenne[file]-adaptive.yaml
+    runner_type: spiceai-dev-large-runners
+    scale_factor: 1
+    duration: 600
+    ready_wait: 60
+    rate: 5000
--- a/tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-adaptive.yaml
+++ b/tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-adaptive.yaml
@@ -5,4 +5,4 @@ tests:
     scale_factor: 1
     duration: 600
     ready_wait: 60
-    rate: 5000
+    rate: 10000
--- a/tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-cdc-memory.yaml
+++ b/tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-cdc-memory.yaml
@@ -1,0 +1,8 @@
+tests:
+  htap:
+    spicepod_path: accelerated/postgres-cayenne[file]-cdc-memory.yaml
+    runner_type: spiceai-dev-large-runners
+    scale_factor: 1
+    duration: 600
+    ready_wait: 60
+    rate: 5000
--- a/tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-cdc-memory.yaml
+++ b/tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-cdc-memory.yaml
@@ -5,4 +5,4 @@ tests:
     scale_factor: 1
     duration: 600
     ready_wait: 60
-    rate: 5000
+    rate: 10000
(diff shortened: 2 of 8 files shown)


tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-cdc-memory.yaml
```diff
@@ -0,0 +1,8 @@
+tests:
+  htap:
+    spicepod_path: accelerated/postgres-cayenne[file]-cdc-memory.yaml
+    runner_type: spiceai-dev-large-runners
+    scale_factor: 1
+    duration: 600
+    ready_wait: 60
+    rate: 5000
```
Comment (sgrebnov) [inline tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-cdc-memory.yaml]:
```suggestion
    rate: 10000
```

tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-adaptive.yaml
```diff
@@ -0,0 +1,8 @@
+tests:
+  htap:
+    spicepod_path: accelerated/postgres-cayenne[file]-adaptive.yaml
+    runner_type: spiceai-dev-large-runners
+    scale_factor: 1
+    duration: 600
+    ready_wait: 60
+    rate: 5000
```
Comment (sgrebnov) [inline tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-adaptive.yaml]:
```suggestion
    rate: 10000
```

sgrebnov (dismissed):

tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-cdc-memory.yaml
Comment (sgrebnov) [reply tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-cdc-memory.yaml]: To be consistent with other SF1 configs otherwise results will be misleading

tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-adaptive.yaml
Comment (sgrebnov) [reply tools/testoperator/dispatch/chbench/sf1/postgres-cayenne[file]-adaptive.yaml]: To be consistent with other SF1 configs otherwise results will be misleading

sgrebnov (approved):

Issue #11212, opened 2026-06-07
Bug. `chbench_q15 mismatch: NoAnswer (source rows=1, spice rows=0)` error


sgrebnov:
## Describe the bug

```
SPICED_LOG="runtime::accelerated_table::refresh_task::changes=trace,cayenne=debug,info" cargo run --release  -p testoperator -- run htap \
  -p './test/spicepods/chbench/accelerated/postgres-cayenne[file]-cdc-tuned-sf100.yaml' \
  --query-set chbench \
  --scale-factor 100 \
  --terminals 100 \
  --duration 30 \
  --ready-wait 300 \
  --rate 2000
```

>  chbench_q15            FAIL
    └─ NoAnswer (source rows=1, spice rows=0)

q15.sql

```sql
WITH revenue0 (supplier_no, total_revenue) AS (
    SELECT
        mod((s_w_id * s_i_id), 10000) AS supplier_no,
        sum(ol_amount) AS total_revenue
    FROM
        order_line, stock
    WHERE
        ol_i_id = s_i_id
        AND ol_supply_w_id = s_w_id
        AND ol_delivery_d >= '2007-01-02 00:00:00.000000'
    GROUP BY
        supplier_no
)
SELECT
    su_suppkey, su_name, su_address, su_phone, total_revenue
FROM
    supplier, revenue0
WHERE
    su_suppkey = supplier_no
    AND total_revenue = (SELECT max(total_revenue) FROM revenue0)
ORDER BY
    su_suppkey;

```

sgrebnov:
No repro anymore - reviewed last 3 runs
```
Analytical Query Correctness
  query               outcome     max Δ%
  chbench_q1             PASS     0.0000
  chbench_q2             PASS     0.0000
  chbench_q3             PASS     0.0000
  chbench_q4             PASS     0.0000
  chbench_q5             PASS     0.0000
  chbench_q6             PASS     0.0000
  chbench_q7             PASS     0.0000
  chbench_q8             PASS     0.0000
  chbench_q9             PASS     0.0000
  chbench_q10            PASS     0.0000
  chbench_q11            PASS          -
  chbench_q12            PASS     0.0000
  chbench_q13            PASS     0.0000
  chbench_q14            PASS     0.0000
  chbench_q15            PASS          -
  chbench_q16            PASS     0.0000
  chbench_q17            PASS     0.0000
  chbench_q18            PASS     0.0000
  chbench_q19            PASS     0.0000
[...]

PR #11238, opened 2026-06-09
feat(testoperator): value-aware HTAP correctness gates

lukekim:
## Problem

The HTAP correctness gates had two value-blind holes:

1. **Drain gate (`row_count.rs`) was `COUNT(*)`-only.** It waits until per-table `COUNT(*)` and `MAX(_bench_ts)` agree source↔Spice — but never checks values. A corruption that preserves count + max-ts (a wrong upsert value, a stale/missed update, a column swap, an upsert landing on the wrong PK) passed it.
2. **Analytical gate (`analytical.rs`) reported a binary pass/fail at a flat 5% tolerance.** 5% is generous for its purpose (FP/encoding rounding is < 0.001%), so a real ~4% drift passed silently and the actual divergence was never surfaced. q15 was excluded entirely (#11212), leaving 21/22.

## Changes

- **Surface per-query divergence + tighten tolerance** (`analytical.rs`): each query now reports its max numeric %-delta (shown even on a pass, so sub-threshold drift is visible), and tolerance is type-aware — **exact** for integer/decimal columns (counts, money), **0.1%** for float — layered on top of the existing structural/row comparator (which still handles schema, row count, nulls and non-numeric columns).
- **Per-table content fingerprint** (`row_count.rs`): at the post-drain snapshot, compare engine-agnostic per-column aggregates source↔Spice — a non-null `COUNT` for every column plus `MIN`/`MAX` for numeric columns. A mismatch fails the gate even when counts match. `SUM` is deliberately excluded (floating sums are order-dependent across engines); text/temporal `MIN`/`MAX` are excluded (collation / timestamp precision differ). This makes the gate value-aware across **all** columns, directly catching the wrong-upsert / stale-update class.
- **Cover q15** → 22/22. Its `total_revenue = (SELECT MAX(total_revenue) …)` floating-equality fragility (#11212) now surfaces as a real divergence rather than being hidden by exclusion.

Shared numeric comparison lives in a new `compare` module.

## Why the fingerprint runs once (not per-poll)

Convergence is still gated on the cheap `COUNT(*)` + `MAX(_bench_ts)` poll. A full-table aggregate scan per probe table is too costly to repeat every second, so the fingerprint runs once at the post-drain snapshot — "did replication catch up?" and "is the content correct?" are answered as two distinct questions, and the gate fails on either.

## Testing

- New unit tests in `compare` (exact integer/decimal, the 0.1% float boundary, nulls / non-numeric ignored) and `row_count` (fingerprint SQL shape) — all green.
- `cargo check --tests`, `cargo clippy --tests --no-deps`, and the unit tests pass on the crate.
- **End-to-end behavior is not yet validated** — exercising the fingerprint and q15 against a live full-scale CDC run isn't feasible on a local box. This needs a full-scale run to confirm q15 and the fingerprint pass on real data; flagged so it's validated before being relied upon.


tools/testoperator/src/commands/htap/correctness/compare.rs
```diff
@@ -0,0 +1,284 @@
...
+        dt,
+        DataType::Float16 | DataType::Float32 | DataType::Float64
+    )
+}
+
+/// Whether a column's values are compared numerically by [`numeric_delta`]
+/// (integers, decimals, floats). Exposed so the fingerprint gate can decide
+/// which columns get a `MIN`/`MAX` aggregate.
+#[must_use]
+pub fn is_numeric_type(dt: &DataType) -> bool {
```
phillipleblanc: why not just make the above method public?

tools/testoperator/src/commands/htap/correctness/row_count.rs
```diff
@@ -242,3 +296,124 @@ async fn query_count_spice(client: &spiceai::Client, table: &str) -> anyhow::Res
...
+///
+/// Probe errors degrade to [`ContentCheck::Skipped`] rather than failing the
+/// gate, mirroring the convergence loop's tolerance of transient blips.
+async fn table_fingerprint(
+    driver: &Arc<dyn ChBenchDriver>,
+    spice_client: &spiceai::Client,
+    table: &str,
+) -> ContentCheck {
+    // Derive the column list from the source schema.
+    let schema_batches = match driver
```
phillipleblanc: Doing a query just to get the schema and then doing another query to get the stats is wasteful. We can just ask Spice for the table schema via Flight, which skips that query entirely.

tools/testoperator/src/commands/htap/correctness/row_count.rs
```diff
@@ -242,3 +296,124 @@ async fn query_count_spice(client: &spiceai::Client, table: &str) -> anyhow::Res
...
+        ContentCheck::Match {
+            max_rel_delta: delta.max_rel_delta,
+        }
+    }
+}
+
+/// Build the engine-agnostic fingerprint aggregate SQL for `table` from its
+/// Arrow `schema`. Column names are quoted to preserve the source's casing on
+/// both engines.
+fn build_fingerprint_sql(table: &str, schema: &arrow::datatypes::Schema) -> String {
```
phillipleblanc: this should use the dataframe API and then the unparser to convert it back to a SQL string.

phillipleblanc (approved):

PR #11241, opened 2026-06-09
feat(cayenne): restore gated ExactLeftAccumulator join rewriter on DF53

lukekim:
## Summary

Restores the gated **`ExactLeftAccumulator`** + `CayenneJoinRewriter` on DF53, behind `cayenne_optimizer_rules.exact_join_filter` (**off by default**). When enabled, `CayenneJoinRewriter` swaps an exact build-key in-list into large-probe analytical joins (probe ≥100K rows, probe:build ratio ≥10) and plants it as a **scan-level pruning predicate** — the precise pruning the DF52 line had, via the forked accumulator (not DF53's native point-lookup InList).

## Depends on

- **spiceai/datafusion#163** — the fork seam (re-genericized `HashJoinExec<A: CollectLeftAccumulator>`, `recreate_with_accumulator`, public traits + config bridge; additive, native default unchanged). The datafusion pin is bumped to the seam rev `506dabefd`; **re-pin to the merged rev once #163 lands.**

## Note — the cap-raise approach was dropped

This PR originally raised DF53's native InList caps (`hash_join_inlist_pushdown_max_distinct_values` / `_max_size`). Testing showed that is **10–100× slower** than stock DF53: the native InList pushdown is built for small (point-lookup) build sides, so forcing giant in-lists for analytical joins degrades to a near-linear per-row membership eval. That commit is reverted here — the diff nets to the gated accumulator only.

## Changes

- `Cargo.toml`/`Cargo.lock` — 35 datafusion pins → the fork seam rev.
- `crates/runtime-datafusion/` — restore `join_accumulator` (`ExactLeftAccumulator`); the one seam delta (`ColumnBounds::as_any`) is implemented.
- `crates/cayenne/src/optimizer_rules.rs` — restore `CayenneJoinRewriter` + gate helpers alongside the existing `CayenneDynamicFilterSharing` (kept).
- `crates/runtime/src/datafusion/builder.rs` — register `CayenneJoinRewriter` gated on `exact_join_filter`.
- `crates/runtime/src/builder.rs` — wire the `exact_join_filter` user-config token (was a no-op).

## Validation

- All three scoped `cargo check`s green; minimal release build green.
- **A/B DEFERRED** — remote benchmark host down. Because the accumulator also materializes an in-list, the A/B must confirm it's genuinely fast (scan-level pruning, not native-InList-style slow) **before enabling** — so it stays gated-off / draft until validated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5538,7 +5538,7 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "53.1.0"
-source = "git+https://github.com/spiceai/datafusion.git?rev=0b0be50fd8c8929a2248a9ce5c75d878c1a68b55#0b0be50fd8c8929a2248a9ce5c75d878c1a68b55"
+source = "git+https://github.com/spiceai/datafusion.git?rev=506dabefda59c4812e867fb11ddd2ff92d365ae0#506dabefda59c4812e867fb11ddd2ff92d365ae0"
 dependencies = [
  "arrow",
  "arrow-schema",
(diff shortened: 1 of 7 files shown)


peasee:
Using this accumulator at all can cause performance reductions, due to the cost of expression simplification of large InLists. Pruning predicates also don't support building predicate expressions with more than a few hundred InList elements anyway.

We will close this for now.

PR #11249, opened 2026-06-09
perf(cayenne): bound mem-tier checkpoint churn + O(1) per-scan deletion view

lukekim:
## What

Follow-up to #11247. Two performance fixes for `cdc_durability: memory` at sustained CDC rates, plus a no-behavior-change simplification pass — each validated at SF-100 / 10K txn/s.

### 1. Bound checkpoint churn (min-flush gate + orphan snapshot cleanup)

SF-100 forensics showed the post-#11247 QPH/lag collapse was checkpoint **churn**, not the checkpoint itself (a local apply repro is flat at ~2M rows/s):

- Every 1 s periodic tick durably checkpointed even a few-KB delta — a new snapshot dir + delete-vector files + a protected-snapshot entry per tick. Measured: **408–676 snapshot dirs per heavy table per 10-minute run**.
- `checkpoint_mem_tier` never called `trigger_old_snapshot_cleanup`, **leaking every predecessor dir**: order_line ended with 408 dirs on disk vs 2 referenced in the catalog (~406 orphans; 89 G footprint vs ~live bytes).
- The (single, shared) catalog accumulated **40.3 M `cayenne_insert_record` rows (1.8 GB)** + 14 K delete-file rows in-run, slowing every single-writer publish as the b-tree grew.

Fixes: **`cdc_mem_tier_min_flush_bytes` (default 32 MiB)** — the periodic tick skips a tier below this size unless its age reached `cdc_mem_tier_max_age_ms` (slot-ack/crash-replay stay bounded; query freshness is unaffected — RAM rows are visible immediately on append; write-path cap spill and explicit checkpoints bypass the gate); and checkpoint phase-2 now triggers the same graceful **old-snapshot cleanup** as every other publish path.

**Validated** (`ms_mem_tuned2`, with the 128 MiB/30 s pod tuning of #11250): correctness PASSED (all 7 tables, 22/22 queries), converged **37.2 s** (file baseline: 186 s), all WAL drained, **0 amplification warnings, 2–3 snapshot dirs (was 408), 15 G footprint (was 97 G), p50 replication lag sub-second on every table** (stock 287 ms vs 60.9 s on file mode).

### 2. O(1) per-scan mem-tier deletion view (persistent maps + merged-snapshot memo)

The remaining gap was QPH: most queries ran within ~2× of file baseline, but **q20 (correlated subquery over stock+order_line, re-scans per outer group) went 322 ms → 74.6 s** — every scan paid two O(tier-tombstones) costs on the two biggest resident tiers.

- `InlinedDeletionMaps` is now a **type alias of the tier's `InMemTombstones`** (same persistent `im::HashMap`/XXH3 maps), so the per-scan view build is an O(1) HAMT-root clone instead of an O(tier) `std::HashMap` materialization.
- The per-scan `with_mem_tier_tombstones` extend (tier tombstones merged into the file-side deletion index, incl. bloom inserts) is **memoized** (`merged_deletion_snapshot`). Key = (file-index `Arc` ptr, new `MemTier.version` content counter, structural epoch): a hit requires all three, so any append/clear/publish forces a rebuild and a stale pairing can never be served. `MemTier.version` exists because `epoch` is preserved across clears and is therefore ambiguous as a content key.

### 3. Simplification pass (no behavior change)

Deduped the under-fence checkpoint tail (was ×3) into one helper; renamed `clear_mem_tier_up_to_epoch` → `clear_mem_tier_flushed_prefix` (it clears a segment-count prefix); single shared age-cap predicate (`mem_tier_age_cap_reached`) for the write-path spill and the tick gate; memo extracted to a provider method; `retain_after` seeds from the first survivor (O(1)); 7 copy-pasted no-op test advancers → one `NoopSlotAdvancer`; comment/doc dedup.

## Tests

- `mem_tier_periodic_tick_gates_small_young_tier` (gate skips small+young; age cap still flushes)
- `merged_scan_deletions_memo_reuses_and_invalidates` (memo hit on quiescent re-scan via `Arc::ptr_eq`, invalidation on append, merge-correct visibility)
- `#[ignore]`d timed apply-throughput repro (`mem_tier_apply_throughput_stays_flat_under_checkpoint_loop`)
- 520 cayenne lib tests green; acid/int64/keybased/on_conflict integration green; cayenne clippy clean.
- Known pre-existing failure (NOT this PR): `inline_with_pending_deletions_test::test_checkpoint_after_inlined_upsert` fails at the branch base `136808ce18` (DF53 line).

## Follow-ups (tracked, not in this PR)

- Off-fence sequence-block refill (the allocator's ~per-512-appends SQLite refill currently runs under the listing fence; split into sync `try_take` + off-fence refill).
- Debt-based admission control on the CDC apply (slowdown/stop keyed on protected-snapshot count) — the structural "compaction keeps up" guarantee.
- Auto/adaptive tuning derivation for the `cdc_mem_tier_*` knobs (autotune currently derives none of them).

Final SF-100 e2e with the full stack is in flight; numbers to follow on the thread.


lukekim:
**SF-100 e2e with the full stack** (`ms_memo3`, 10K txn/s, 600s, sqlite metastore + `cdc_durability: memory`, idle box):

- **Correctness PASSED** — all 7 tables match, 21/22 queries match (1 advisory); **converged in 15.5s** (file baseline: 186s — 12× faster); all WAL slots drained to 56 bytes.
- **Churn bounded**: 2 amplification warnings (was continuous), 4-5 snapshot dirs/table (was 408-676), 14G footprint (was 97G).
- **Replication lag p50**: customer 130ms / district 114ms / new_order 387ms / order_line 2.2s / oorder 152ms / stock 5.9s / warehouse 132ms — vs 24-61s on file mode for the heavy tables.
- **QPH 1237** (pre-memo tuned2: 941, +31%). Most query means are now at or better than file-baseline levels (q1 1015ms vs 757; q10 5351 vs 5581; q18 2766 vs 4301).
- **Remaining outlier: q20 at 64.9s mean** (file mode: 322ms). The memo did not move it, which points away from the tombstone-merge cost and at **plan shape**: with a resident RAM tier the scan is a union of (file scan + `MemorySourceConfig`), and q20's correlated subquery over stock+order_line appears to lose decorrelation/pruning across that union — a per-outer-group rescan of the union. Next investigation: `EXPLAIN` q20 with and without a resident tier to confirm the union blocks the subquery optimization, then either teach the optimizer to push through the union or decorrelate before the union is injected.

lukekim:
**Correction on the q20 hypothesis** (validity-bar follow-through): a regression test (`mem_tier_join_probe_keeps_dynamic_filter_pushdown`, now on this PR) shows DF53's hash-join dynamic filter **does** install on the probe-side DataSourceExec even with a resident mem tier (union + bare memory branch) — DataFusion routes the pushdown per-child through the union. The union-blocks-pushdown theory is refuted at this shape; an always-true-FilterExec absorber on the memory branch was prototyped, changed nothing in A/B, and was not shipped. The test stays as a guard pinning the current-good routing. The SF-100 dimension-join slowdown (q20 322ms → 65s; supplier-join set 4-200×) remains open — next investigation is the live SF-100 plan: hash-join mode selection and `statistics()` with a resident tier (EXPLAIN q20 under load via the harness probe).

lukekim:
**The SF-100 memory-mode scan collapse now reproduces locally** (new `mem_tier_join_shapes` bench, live-append lanes): `single_scan` 2.84ms → **701ms (247×)** and q20-shaped `semi_join` 10.1ms → **1.77s (175×)** while value-identical upserts stream at live-CDC cadence (results byte-identical; correctness-gated). The load-coupled component is **churn-coupled, not scale-coupled**: per-scan O(resident-tier) re-materialization + merged-tombstone rebuild whose version-keyed memo can never hit while appends bump the tier version per batch. Fix direction (next PR): maintain the merged deletion view + visible-batch overlay **incrementally at append time** (O(delta) HAMT extend — the same structural-sharing trick as #11232), making per-scan cost O(1) under churn. The fix loop is now ~4 minutes local instead of 50-minute remote runs.

lukekim:
**The churn-coupled scan collapse is FIXED** (O(delta) incremental scan view, latest commit): both per-scan O(resident-tier) costs are now maintained incrementally at append time — the merged deletion snapshot extends in lockstep with each append's tombstone delta, and a new `MemTierVisibleOverlay` keeps the merge-on-read-filtered batches current (new segment pushed as-is; only pk-range-overlapping cached entries re-filtered; invalidated on checkpoint clear and on file-side deletion publishes). Measured on the `mem_tier_join_shapes` live lanes (byte-identical correctness gate, 0 divergences): `single_scan` under live appends **701ms → 2.75ms (255×, at-rest parity)**; q20-shaped `semi_join` **1.77s → 9.06ms (195×, better than file_only's 9.6ms)**. 521 lib tests green; clippy clean. This is the in-process fix for the SF-100 q20/QPH collapse — final SF-100 e2e validation queued once the bench box is stable.

phillipleblanc (approved):
Merged latest trunk and resolved the 26-file conflict set (kept the branch's mem-tier refinements; took trunk's #11260 metadata-only-publish side incl. the upgrade_tombstones_for_flushed_pks fix). Reverted a stale carried ballista workaround in query/handle.rs that broke ballista_stage task-history rows. All 857 cayenne tests + distributed_task_history integration tests pass locally; CI fully green.

PR #11252, opened 2026-06-10
fix: broadcast-join rewrite returns wrong results on shared column names, NULL-equal joins, and residual filters

claudespice:
## Root cause and user-visible symptoms

The new `FlightSQLBroadcastJoinPushdown` rule (#11245) rewrites a scheduler-side `HashJoinExec` over federated Cayenne scans into per-executor join SQL. Three plan shapes were rendered unfaithfully, each silently changing query results:

1. **Shared column names return the wrong table's data.** `build_join_sql` attributed each join-output field to `f.`/`d.` by name, checking the fact side first. When both tables expose a same-named non-key column (e.g. both have a `note` column and the query selects both), the dim's output column was rendered as `f."note"` — so the result carried the fact's values in the dim's column. Fixed: bail (keep the central join) on any name present in both scan schemas, unless it is an equi-join key joined to itself (`f.x = d.x`), where the inner join guarantees both sides carry equal values on every output row — this keeps `USING (id)`-style joins broadcastable.

2. **`IS NOT DISTINCT FROM` joins lose NULL-key matches.** The rule ignored the join's `NullEquality`. A `HashJoinExec` planned with `NullEqualsNull` matches NULL keys, but the rewritten SQL `f.k = d.k` never does, so NULL-key rows silently vanished from results. Fixed: bail unless `NullEqualsNothing` (the same guard `cayenne::logical_optimizer` applies before its join rewrites).

3. **Residual `FilterExec` predicates were silently dropped.** Both this rule and the sibling `FlightSQLPartialAggregatePushdown` treated `FilterExec` as a pass-through wrapper while resolving the federated scans, on the assumption the scan SQL already carried the predicate. `FlightSQLTable::supports_filters_pushdown` classifies every filter as `Exact` (removed from the plan entirely, embedded in the scan SQL) or `Unsupported` (kept **only** in a `FilterExec`, absent from the scan SQL) — never `Inexact`. So a `FilterExec` above these scans always carries a predicate the rewritten SQL would not apply: the broadcast join returned extra (unfiltered) rows, and the aggregate pushdown computed aggregates over unfiltered rows. Fixed in both rules in one PR (they are symmetric siblings in the same module): `FilterExec` now blocks the rewrite, keeping the central plan, which still executes the filter.

All three guards fall back to the unrewritten plan, trading the optimization for correctness on exactly the shapes the rewrite cannot render faithfully. The validated TPC-H broadcast wins (q5/q8/q9/q17/q21 — unique column names, plain equi-joins, fully-pushed filters) are unaffected.

## Test plan
- [x] Added regression tests: ambiguous shared column name bails; same-named equi-key pair still broadcasts (side attribution asserted on the generated SQL); `NullEqualsNull` bails; residual `FilterExec` blocks the broadcast rewrite; residual `FilterExec` blocks aggregate pushdown (plan asserted unchanged)
- [x] `make lint` (workspace `cargo fmt --all -- --check` passed; the workspace clippy invocation exercised the changed crate — it caught one pedantic `match_same_arms` in the new code, which was fixed and re-verified — and continues over the unchanged remainder in a clean sibling worktree while CI Rust Lint runs the identical command on this PR)
- [x] `make test` (`cargo test --all --lib`; the changed crate's full suite is green locally — 47 passed, 0 failed — and the workspace run completes in the sibling worktree alongside CI Build/Test; no other crate's code or public interface is touched by this diff)


claudespice:
@copilot review

claudespice:
Rebased onto trunk to resolve conflicts with #11256, which independently landed an overlapping fix for the same rule. To avoid duplicated/lost work, this PR is kept as the **superset** of the two:

- **#11256 covers** the residual `FilterExec` drop (issue 3) and shared-column attribution (issue 1, by bailing on *every* same-named column), plus a test suite.
- **This PR additionally keeps:**
  - The `null_equality() != NullEqualsNothing` guard (**issue 2**) — trunk has no such check, so a `NullEqualsNull` (`IS NOT DISTINCT FROM`) join is still rewritten to SQL `=` on trunk and silently drops NULL-key matches.
  - The `same_named_key_pairs` exception, so `USING (id)`-style joins (a same-named column that *is* the equi-join key joined to itself) remain broadcastable instead of bailing as trunk does.
  - The `FilterExec`-not-pass-through fix in both `broadcast_join` and `aggregate_pushdown` (matches #11256).

Also ported #11256's `filter_exec_in_fact_union_leg_blocks_broadcast` scenario as `bails_on_filter_exec_inside_fact_union_leg` for extra coverage.

`cargo test -p datafusion-optimizer-rules flightsql` → 33 passed; crate clippy + rustfmt clean.

lukekim (approved):

PR #11254, opened 2026-06-10
refactoring

Jeadie:


Jeadie:
Was an experiment. Should be redone and possible piecemeal.

PR #11256, opened 2026-06-10
fix: FlightSQL broadcast-join and partial-agg pushdown silently drop un-pushed FilterExec predicates

claudespice:
## Root cause

The distributed (cayenne cluster) FlightSQL pushdown rules — `FlightSQLBroadcastJoinPushdown` (new in #11245) and the partial-aggregation pushdown — treat `FilterExec` as a data-preserving pass-through when resolving a federated scan, on the assumption that "the underlying `FlightSqlExec` already carries the predicate in its SQL; the `FilterExec` is a redundant safety layer".

That assumption is wrong for this provider. `FlightSQLTable::supports_filters_pushdown` returns `Exact` for predicates `to_sql_preserving_precedence` can render and `Unsupported` otherwise — never `Inexact`. DataFusion's `push_down_filter` removes `Exact` predicates from the `Filter` node entirely (they live only in the scan SQL) and keeps `Unsupported` and volatile predicates only in the `Filter` node (they are NOT in the scan SQL). So a `FilterExec` above these scans always carries a predicate the scan SQL does not have — e.g. a `CASE` expression in `WHERE`, a `LIKE ... ESCAPE`, a `TryCast`, or a volatile function.

**User-visible symptom:** on a coordinator with live executors, a distributed query whose `WHERE` contains such a predicate silently returns wrong results:

- the broadcast-join rewrite replaced the whole join subtree with generated per-executor join SQL built only from the scans' own SQL — the predicate vanished and unfiltered rows joined;
- the partial-agg pushdown replaced the `AggregateExec(Partial) → … → FlightSqlExec` chain — aggregates were computed over unfiltered rows.

A second, independent bug in the broadcast rule's `build_join_sql`: output columns are attributed to the fact/dim subqueries **by name**. When both join inputs expose a same-named column (common outside TPC-H: `id`, `name`, `created_at` on both tables), the join's output schema distinguishes the two only by position, and both output fields rendered as `f."col"` — shipping fact values in the dim column.

## Fix

- `FilterExec` is no longer pass-through in either rule; the rewrite bails and keeps the central (correct) plan. Renderable predicates are pushed `Exact` and plan no `FilterExec`, so the validated TPC-H broadcast/agg-pushdown wins are unaffected.
- `build_join_sql` bails when an output column name is attributable to both join sides.

## Test plan
- [x] Added regression tests: positive control (broadcast fires for a clean small-dim join, dimension gathered from every executor, keys map `f.`/`d.`); `FilterExec` above the dim scan blocks the rewrite; `FilterExec` inside a fact union leg blocks the rewrite; shared output column name blocks the rewrite; `FilterExec` above the scan blocks partial-agg pushdown
- [x] `make lint` — fmt verified clean on the touched files and the repo's strict clippy deny set (`CLIPPY_CONF_DIR=.ci`, lib + tests profiles) verified clean on the touched crate locally; the workspace-wide gate cannot execute in this checkout (nested-worktree `cargo metadata` failure) and runs in CI Rust Lint
- [x] `make test` — all 30 flightsql optimizer-rule lib tests pass locally (26 pre-existing + 4 new broadcast + 1 new agg-pushdown, one shared); full workspace suite runs in CI

### DIFF
--- a/crates/datafusion-optimizer-rules/src/physical_plan/flightsql/aggregate_pushdown.rs
+++ b/crates/datafusion-optimizer-rules/src/physical_plan/flightsql/aggregate_pushdown.rs
@@ -36,7 +36,6 @@ use datafusion::common::tree_node::{Transformed, TreeNode};
 use datafusion::config::ConfigOptions;
 use datafusion::physical_optimizer::PhysicalOptimizerRule;
 use datafusion::physical_plan::aggregates::{AggregateExec, AggregateMode};
-use datafusion::physical_plan::filter::FilterExec;
 use datafusion::physical_plan::projection::ProjectionExec;
 use datafusion::physical_plan::repartition::RepartitionExec;
 use datafusion::physical_plan::union::UnionExec;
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11286, opened 2026-06-11
feat(flightsql): infer schema via SELECT * LIMIT 1 when GetTables is unimplemented

melks:
## Summary

Some Flight SQL servers implement the statement-execution path but not the `GetTables` metadata RPC, returning gRPC `Unimplemented`. `FlightSQLTable::get_schema` relies solely on `GetTables`, so creating a dataset against such a server fails schema detection — every table errors with "Failed to detect table schema". A concrete example is StarRocks' (experimental) Arrow Flight SQL: handshake and query execution work, but `CommandGetTables` returns `Unimplemented`.

## Changes

Make the `GetTables` path best-effort, and when it doesn't yield a schema, fall back to inferring the schema from `SELECT * FROM <table> LIMIT 1` (served by the statement-execution path).

## Testing

Verified end-to-end against a StarRocks 4.0.10 Arrow Flight SQL endpoint: datasets that previously failed schema detection now register and serve. No behavior change for servers that implement `GetTables`.

## Notes for Reviewer

The fallback only triggers when `GetTables` errors or returns no usable schema, so conformant servers are unaffected. `SELECT * ... LIMIT 1` is a minimal probe any Flight SQL execution endpoint can serve.

### References

| Type | Link |
|---|---|
| Context | Surfaced while integrating Spice with StarRocks' experimental Arrow Flight SQL |


phillipleblanc (dismissed):
Looks good to me, thanks!

lukekim (dismissed):

lukekim:
@melks this needs to pass the Lint check.

claudespice:
Pushed a small fix to unblock CI: `Rust Lint` was failing on `clippy::collapsible_if` (Rust 1.95 now flags collapsible `if let` chains) in `crates/data_components/src/flightsql.rs`. Collapsed the three nested `if let`s in the schema-fallback path into a single let-chain — no behavior change. The remaining integration-test failures are the standard secret-gated jobs that can't run on fork PRs.

lukekim (approved):

PR #11288, opened 2026-06-11
fix: digest_many length-prefixes values so column boundaries can't collide (fixes #11272)

### LINKED ISSUE #11272: digest_many concatenates column values with no delimiter — cross-column collisions; used as the rrf() row identity

`crates/runtime-datafusion-udfs/src/digest_many.rs:170-189` builds the per-row digest input by writing each column's `ScalarValue` Display into one buffer with no separator and no column count: `('ab','c')` digests identically to `('a','bc')`, integer composite keys `(1,23)` ≡ `(12,3)`, and SQL NULL prints `"NULL"` ≡ the literal string `'NULL'`.

`crates/runtime/src/search/rrf.rs:1102` uses `digest_many(<pk cols>, 'md5')` as the RRF fusion row id — two distinct documents whose composite PKs collide on a column boundary are fused as one row (ranks merged, one document silently dropped/misattributed). Any user relying on `digest_many` for dedup/change detection has the same hazard.

Suggested fix: write a length prefix or an escape-aware delimiter between values (and a distinct token for NULL vs the string "NULL"). Note this changes digest outputs for multi-column calls — call that out in the changelog; single-column digests are unaffected by a separator-only scheme.

claudespice:
## Summary

`digest_many` builds a per-row hash input by concatenating each column's `ScalarValue` Display output into a single buffer with **no separator and no NULL marker** (`crates/runtime-datafusion-udfs/src/digest_many.rs`). That made the input ambiguous on column boundaries:

- `('ab', 'c')` and `('a', 'bc')` both produced `"abc"` → identical digest
- integer composite keys `(1, 23)` and `(12, 3)` both produced `"123"`
- a SQL `NULL` (Display `"NULL"`) was indistinguishable from the literal string `'NULL'`

`digest_many(<pk cols>, 'md5')` is the **RRF fusion row identity** (`crates/runtime/src/search/rrf.rs:1102`), so two distinct documents whose composite primary keys collide on a column boundary were fused into one row — their ranks merged and one document silently dropped or misattributed. Any user relying on `digest_many` for dedup/change detection had the same hazard.

Fixes #11272

## Changes

- Add `DigestMany::append_value`, which encodes each value injectively: a non-NULL value `v` is written as `<byte_len>:<v>`, and a SQL `NULL` is written as a bare `N`. A length prefix always begins with a digit, so it can never be confused with the `N` NULL marker, and the explicit byte length makes adjacent value boundaries unambiguous regardless of value content (digits, colons, or the literal text `N`).
- Use the helper from both the all-scalars fast path and the row-by-row array path, with a reusable scratch buffer to avoid per-value allocation.
- Update the inline `digest_many_record_batch` snapshot to the new (correct) md5 outputs.

## Behavior / performance impact

This changes `digest_many` output for **all** calls, single- and multi-column (the length prefix is applied uniformly). The value is recomputed per query for RRF and is **not** persisted across versions, so there is no on-disk migration concern; any external consumer that persisted digests for change-detection will see a one-time invalidation. The per-row cost adds one reusable scratch buffer and a small integer prefix per value — no extra allocations in the hot loop.

## Test plan
- [x] Added regression test for the string column-boundary collision (`('ab','c')` vs `('a','bc')`)
- [x] Added regression test for the integer composite-key collision (`(1,23)` vs `(12,3)`)
- [x] Added regression test for SQL NULL vs the literal string `'NULL'`
- [x] `make lint` passes (workspace `cargo fmt --all -- --check` plus full clippy)
- [x] `make test` passes
- [x] Updated the `digest_many_record_batch` inline snapshot for the new digest outputs

### DIFF
--- a/crates/runtime-datafusion-udfs/src/digest_many.rs
+++ b/crates/runtime-datafusion-udfs/src/digest_many.rs
@@ -94,6 +94,39 @@ impl DigestMany {
             scalar_arguments: &[Some(&ScalarValue::Utf8(Some(String::new())))],
         })
     }
+
+    /// Append one column value to a row's pre-hash buffer using a length-prefixed,
+    /// NULL-distinct encoding so that the concatenation is injective: distinct tuples
+    /// always map to distinct strings, regardless of where the column boundaries fall.
+    ///
+    /// The previous implementation wrote each value's `ScalarValue` Display into one
+    /// buffer with no separator and no NULL marker. That made the per-row digest input
+    /// ambiguous: `('ab', 'c')` and `('a', 'bc')` produced the same string, integer
+    /// composite keys `(1, 23)` and `(12, 3)` collided, and a SQL NULL (Display `"NULL"`)
+    /// was indistinguishable from the literal string `'NULL'`. Because `digest_many`
+    /// backs the RRF fusion row identity (`crates/runtime/src/search/rrf.rs`), a single
+    /// such collision silently fuses two distinct documents into one — merging their
+    /// ranks and dropping/misattributing one of them.
+    ///
+    /// Encoding: a NULL value is written as a bare `N`; a non-NULL value `v` is written
+    /// as `<byte_len>:<v>`. A length prefix always begins with a digit, so it can never
+    /// be confused with the `N` NULL marker, and the explicit byte length makes the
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11294, opened 2026-06-11
fix: restore federation deny-list enforcement regressed by the DataFusion 53 upgrade

claudespice:
The DataFusion v53.1 upgrade (#11118) regressed the Spice-function federation deny-list (issue #10703) in two independent ways, plus dropped the SQLite decimal `BETWEEN` rewrite. All three silently change query behavior.

## Root causes and user-visible symptoms

**1. SQLite and Postgres accelerator factories lost their deny-list (and SQLite its `decimal_between`).**
The upgrade migrated the DuckDB accelerator and MySQL connector to the new table-providers-typed `FunctionSupport` seam (`deny_spice_functions_for_*_table_providers`), but the SQLite and Postgres accelerator factories' builder calls were dropped outright instead of migrated — the fork APIs (`SqliteTableProviderFactory::with_function_support`/`with_decimal_between`, `PostgresTableProviderFactory::with_function_support`) still exist at the pinned rev:

- `with_decimal_between(true)` enables the `SQLiteBetweenVisitor`, which rewrites federated `BETWEEN` comparisons over decimal columns (stored as TEXT in SQLite) to compare numerically. Without it, `WHERE price BETWEEN 0.07 AND 0.10` against a `sqlite`-accelerated decimal column can silently return wrong rows (lexical TEXT comparison).
- Without `function_support`, plans referencing Spice-only functions (`json_get_str`, `embed`, user UDFs) are federated into SQLite/Postgres accelerator SQL instead of falling back to local evaluation — failing with "no such function", where names colliding with remote builtins of different semantics could silently return different results.

Fix: add `deny_spice_specific_functions_table_providers()` (the full default deny-list expressed in the fork's `TpFunctionSupport` type; the MySQL helper now delegates to it) and restore the wiring in both accelerator factories.

**2. `contains_unsupported_functions` only inspected the root plan node.**
The upgrade replaced the fork's `datafusion_table_providers::util::supported_functions::contains_unsupported_functions` — which walks the entire plan including subqueries via `plan.apply_with_subqueries` — with a local reimplementation in `data_components::function_support` that iterates only `plan.expressions()` on the single node it is handed. Callers pass the root of an entire federated subtree (`unfederate_plan_with_unsupported_functions` passes `federated.plan()`; flight/flightsql pass the plan given to `can_execute_plan`), so a denied function in a `WHERE` predicate below the root projection — or inside an `IN`/`EXISTS`/scalar subquery at any depth — escaped the check and was unparsed into the remote engine's SQL. This affected every consumer of the local check: the generic SQL federation wrapper (Snowflake, ClickHouse, ODBC, Dremio, Postgres connectors), Turso, Flight, and FlightSQL.

Fix: walk the plan with `apply_with_subqueries`, matching the fork implementation this replaced.

## Test plan
- [x] Added regression tests for the deep plan walk: denied function in a `Filter` below the root node, and denied function inside an `IN (<subquery>)` plan (both fail against the previous root-only implementation), plus a root-node and a no-denied-function control case (`data_components::function_support::tests`)
- [x] Added regression test that the new table-providers-typed default deny-list denies Spice-only functions (`embed`, `cosine_distance`, `json_get_str`) and allows ordinary functions (`runtime::datafusion::udf::tests`)
- [x] `make lint` passes (workspace-wide `cargo fmt --all -- --check` + full workspace clippy, not per-crate)
- [x] `make test` run (as `cargo test --all --lib --no-fail-fast` for full coverage) — 88 suites green; the only failures are pre-existing/environmental and unrelated to this diff: 28 `cayenne_catalog` tests that hardcode `/tmp/cayenne_test` (owned by a different user on this shared runner → `PermissionDenied`; no cayenne code touched here), the known external `test_integration_httpbin_500_server_error` flake, and `test_refresh_status_change_to_ready` which timed out under full-workspace load and passes when run solo.

### DIFF
--- a/crates/data_components/src/function_support.rs
+++ b/crates/data_components/src/function_support.rs
@@ -94,26 +94,31 @@ fn supports_name(restriction: Option<&FunctionRestriction>, name: &str) -> bool 
     }
 }
 
+/// Returns `true` if any expression anywhere in `plan` — including child plan
+/// nodes and subquery plans (`IN`/`EXISTS`/scalar subqueries) — references a
+/// function the `function_support` restriction rejects.
+///
+/// The walk must cover the whole plan tree, not just `plan.expressions()` on
+/// the root node: callers pass the root of an entire federated subtree (e.g.
+/// `Projection → Filter → TableScan`), so a denied function in a `WHERE`
+/// predicate or subquery would otherwise escape the check and be unparsed into
+/// the remote engine's SQL.
 pub fn contains_unsupported_functions(
     plan: &LogicalPlan,
     function_support: &FunctionSupport,
 ) -> Result<bool> {
     let mut unsupported = false;
-    for expr in plan.expressions() {
-        expr.apply(|expr| {
-            if !function_support.supports(expr) {
+    plan.apply_with_subqueries(|plan| {
+        for expr in plan.expressions() {
+            if !function_support.supports(&expr) {
                 unsupported = true;
                 return Ok(TreeNodeRecursion::Stop);
             }
-            Ok(TreeNodeRecursion::Continue)
-        })?;
-
-        if unsupported {
-            return Ok(true);
         }
-    }
+        Ok(TreeNodeRecursion::Continue)
+    })?;
 
-    Ok(false)
+    Ok(unsupported)
 }
 
 pub fn unfederate_plan_with_unsupported_functions(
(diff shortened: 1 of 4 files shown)


claudespice:
@copilot review

sgrebnov (approved):

PR #11297, opened 2026-06-12
fix(adbc): prevent Spice-specific UDFs from being pushed down to ADBC sources

krinart:
## 📝 Summary

Wires `with_function_support` into `AdbcTableFactory` so that Spice-specific UDFs (json functions, bucket, embedding functions, etc.) are never pushed down to remote ADBC sources that cannot execute them. 

Depends on (need to wait until these PRs are merged to use proper references)
  - `datafusion-federation`: https://github.com/spiceai/datafusion-federation/pull/79
  - `datafusion-table-providers`: https://github.com/spiceai/datafusion-table-providers/pull/17

## Refactoring

Currently we have two versions of `FunctionSupport`: 
* [one in `datafusion-table-providers`](https://github.com/datafusion-contrib/datafusion-table-providers/blob/sgrebnov/spiceai-53/core/src/util/supported_functions.rs#L52) 
* And [one in spice](https://github.com/spiceai/spiceai/blob/8f58a1a124e812cc2b3aab5d35eaedf9f71ef860/crates/data_components/src/function_support.rs#L34)

I removed spice implementation and now we only have:
* One `FunctionSupport` defined in `datafustion-table-providers`
* Single implementation of `deny_spice_functions_for_table_providers` which is used by all connectors

## Notes

Queries that mix a denied UDF with a sort (e.g. `SELECT json_get_str(data, 'name') FROM t ORDER BY id`) will now execute the `ORDER BY` in DataFusion rather than pushing it into the remote SQL. This is a consequence of how the
federation analyzer handles partial plan federability.

Pushing sort through a blocked projection is a valid future optimization but requires the federation analyzer to identify "transparent-to-federation" ancestor nodes after hitting a block.


crates/runtime/src/dataconnector/adbc.rs
```diff
@@ -520,7 +521,9 @@ impl AdbcFactory {
                 }
             })?;
 
-        let adbc_factory = AdbcTableFactory::new(Arc::clone(&pool));
+        let adbc_factory = AdbcTableFactory::new(Arc::clone(&pool))
+            .with_federation_enabled(federation_enabled)
+            .with_function_support(deny_spice_functions_for_mysql_table_providers());
```
phillipleblanc: This is hard-coding for mysql - but ADBC can be any remote source. We should change this to be dialect based at least.

krinart: The reason we had mysql-specific function is because we had two version of `FunctionSupport`: [one in `datafusion-table-providers`](https://github.com/datafusion-contrib/datafusion-table-providers/blob/sgrebnov/spiceai-53/core/src/util/supported_functions.rs#L52) and [one in spice](https://github.com/spiceai/spiceai/blob/8f58a1a124e812cc2b3aab5d35eaedf9f71ef860/crates/data_components/src/function_support.rs#L34)

I removed spice implementation and now we only have:
* One `FunctionSupport` defined in `datafustion-table-providers`
* Single implementation of `deny_spice_functions_for_table_providers` which is used by all connectors

If we want to add connector-specific deny/map list - I think this should be a separate PR

lukekim (approved):

PR #11298, opened 2026-06-12
feat(postgres): share generated replication slots per instance

phillipleblanc:
## Summary

Adds an opt-in `pg_replication_slot_scope` parameter for Postgres datasets that use `refresh_mode: changes`. The default `dataset` scope keeps the existing behavior: one generated replication slot per dataset per runtime instance.

The new `instance` scope generates one shared slot name per runtime instance and Postgres source connection: `spice_inst_<instance-hash>_<source-hash>`. Datasets on the same source can share one slot, publication, and walsender without making every runtime replica contend for the same physical Postgres slot.

This reduces replication slot usage from `datasets × replicas` to `sources × replicas`. Explicit `pg_replication_slot` values still take precedence and are used verbatim.

## Example config

Set `pg_replication_slot_scope: instance` on each Postgres dataset that uses `refresh_mode: changes` and should share the generated per-instance slot for the same source connection:

```yaml
version: v1
kind: Spicepod
name: postgres-cdc

datasets:
  - from: postgres:public.orders
    name: orders
    params:
      pg_host: db.internal
      pg_port: 5432
      pg_db: app
      pg_user: spice
      pg_pass: ${secrets:pg_pass}
      pg_replication_slot_scope: instance
    acceleration:
      enabled: true
      refresh_mode: changes

  - from: postgres:public.customers
    name: customers
    params:
      pg_host: db.internal
      pg_port: 5432
      pg_db: app
      pg_user: spice
      pg_pass: ${secrets:pg_pass}
      pg_replication_slot_scope: instance
    acceleration:
      enabled: true
      refresh_mode: changes
```

## Flow

```text
Default: pg_replication_slot_scope: dataset

runtime replica 1
  dataset A -> spice_<dataset-a>_<dataset-hash>_<instance-hash>
  dataset B -> spice_<dataset-b>_<dataset-hash>_<instance-hash>

Instance scope: pg_replication_slot_scope: instance

runtime replica 1
  dataset A --\
  dataset B ----> spice_inst_<instance-hash>_<source-hash> -> shared multiplexer
  dataset C --/

runtime replica 2
  dataset A --\
  dataset B ----> spice_inst_<different-instance-hash>_<same-source-hash>
  dataset C --/
```

## Changes

- Add `pg_replication_slot_scope` with `dataset` as the default and `instance` as the opt-in shared mode.
- Generate instance-scoped slot names from the runtime instance identity (`SPICE_INSTANCE_ID`, falling back to hostname) and the Postgres source tuple `(host, port, database, user)`.
- Route instance-scoped generated slots through the existing shared replication multiplexer.
- Keep explicit `pg_replication_slot` behavior unchanged, including precedence over `pg_replication_slot_scope`.
- Update Postgres replication docs and add coverage for slot naming, source separation, explicit-slot precedence, and generated-slot multiplexing.

## Notes

- Existing deployments are unchanged unless `pg_replication_slot_scope: instance` is set.
- Instance-scoped datasets on the same source share one replication connection/walsender, so they also share throughput and back-pressure.

## Test plan

- [ ] CI

### DIFF
--- a/crates/data-connectors/connector-postgres/src/lib.rs
+++ b/crates/data-connectors/connector-postgres/src/lib.rs
@@ -157,14 +157,28 @@ const PARAMETERS: &[ParameterSpec] = &[
         "Name of the Postgres replication slot to create/reuse for this dataset. \
          Defaults to `spice_<dataset>_<dataset-hash>_<instance-hash>`. Datasets on the \
          same connection that name the same slot SHARE it: one replication connection, \
-         one publication, with decoded changes routed per table. Each Spice replica \
-         MUST have its own unique slot.",
+         one publication, with decoded changes routed per table. An explicit name is used \
+         verbatim, so each Spice replica MUST set a unique slot — or omit this and set \
+         `pg_replication_slot_scope: instance` to share one auto-named slot per replica.",
     ),
+    ParameterSpec::component("replication_slot_scope")
+        .description(
+            "Scope of the auto-generated replication slot when `pg_replication_slot` is not \
+             set. `dataset` (default): one slot per dataset \
+             (`spice_<dataset>_<dataset-hash>_<instance-hash>`). `instance`: one slot per \
+             Spice replica per source connection (`spice_inst_<instance-hash>_<source-hash>`), \
+             shared by every `refresh_mode: changes` dataset on that source via the \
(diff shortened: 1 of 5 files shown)


phillipleblanc:
Will close this in favor of using secret interpolation to handle instance scoped replication slots (i.e. `pg_replication_slot: spice_slot_${env:instance}`)

PR #11304, opened 2026-06-12
fix: strict validation of Postgres CDC params instead of silent defaults (fixes #11274)

### LINKED ISSUE #11274: Postgres CDC: unrecognized replication_initial_snapshot value silently disables the initial snapshot (typo -> missing all pre-existing rows)

`crates/data-connectors/connector-postgres/src/replication.rs:536-538` — `parse_bool_default_true` returns true only for `true|1|yes|y` (or empty); any other value, including `True ` with trailing whitespace handled but `"ture"`/`"TRUE,"`/etc., parses to **false**. Combined with `replication.rs:457-458` (`.is_none_or(|s| parse_bool_default_true(&s))`), a typo'd `replication_initial_snapshot` silently skips the bootstrap snapshot — the accelerator serves only post-subscription changes and is missing every pre-existing row, with no error.

Same file, smaller: `port` (`replication.rs:443-445`) and `replication_status_interval`/`sslmode` parse failures silently substitute defaults via `.ok()`/`unwrap_or` (the non-CDC pool path validates `pg_port` loudly, so the port case usually fails there first).

Suggested fix: parse these as strict booleans/values and error on unrecognized input (matching the loud validation on the non-CDC path), rather than defaulting.

claudespice:
## Summary

Postgres CDC parameter parsing in `replication_params_from_connector_params` silently swallowed invalid values. The highest-impact case: `replication_initial_snapshot` was parsed with `parse_bool_default_true`, which returned `true` only for `true/1/yes/y` and collapsed **every other value to `false`**. A typo (`ture`, `True,`, trailing whitespace, …) therefore silently disabled the bootstrap snapshot — the accelerator served only post-subscription WAL changes and was missing every pre-existing row, with no error.

Same silent-default bug-class in the same function for:
- `replication_temporary_slot` (`parse_bool_default_false`)
- `port` (`.parse().ok().unwrap_or(5432)`)
- `replication_status_interval` (`fundu::parse_duration(..).ok().unwrap_or(default)`)
- `sslmode` (`SslMode::from_str_or_default` — a typo'd `verify-full` silently downgraded to `prefer`, disabling cert/hostname verification: a quiet TLS/MITM downgrade)

Fixes #11274.

## Changes

- Replace `parse_bool_default_true` / `parse_bool_default_false` with strict `optional_bool` that accepts `true/1/yes/y` and `false/0/no/n` (case-insensitive, trimmed) and **errors on anything else**, reporting the user-facing parameter name (matching the existing loud `optional_usize_in_range` path).
- Add `optional_parse<T: FromStr>` (used for `port`) and `optional_duration` (used for `replication_status_interval`) — both error loudly on unparseable input.
- Add `SslMode::from_str_strict` in `data_components::postgres_replication::config` alongside the lenient `from_str_or_default`; the CDC path now uses the strict variant.
- Absent / empty values still resolve to the previous defaults, so all valid configs behave identically — only genuinely-invalid input now errors instead of silently doing the wrong thing.

## Test plan

- `make lint`
- `make test`
- New unit tests in `replication.rs`: valid true/false token sets, default-on-absent/empty, regression for the `replication_initial_snapshot` typo (errors instead of dropping rows), `port` non-numeric rejection, `replication_status_interval` unparseable rejection, and `SslMode::from_str_strict` rejecting an unknown mode while preserving the `prefer` default for absent.

## Checklist

- [x] Strict parsing errors loudly on invalid input (matches non-CDC path)
- [x] Valid configs and existing defaults unchanged (no default flipped)
- [x] Scoped `cargo clippy` clean; `cargo test -p connector-postgres` green

### DIFF
--- a/crates/data-connectors/connector-postgres/src/replication.rs
+++ b/crates/data-connectors/connector-postgres/src/replication.rs
@@ -480,14 +480,13 @@ fn replication_params_from_connector_params(
     dataset_name: &str,
 ) -> std::result::Result<ReplicationParams, String> {
     let host = required_string(params, "host")?;
-    let port = optional_string(params, "port")
-        .and_then(|s| s.parse::<u16>().ok())
-        .unwrap_or(5432);
+    let port = optional_parse::<u16>(params, "port", 5432, "a port number (0-65535)")?;
     let user = required_string(params, "user")?;
     let password_str = required_secret(params, "pass")?;
     let database = required_string(params, "db")?;
     let sslmode =
-        config::SslMode::from_str_or_default(optional_string(params, "sslmode").as_deref());
+        config::SslMode::from_str_strict(optional_string(params, "sslmode").as_deref())
+            .map_err(|reason| format!("parameter `{}` {reason}", params.user_param("sslmode")))?;
     let sslrootcert = optional_string(params, "sslrootcert").map(std::path::PathBuf::from);
 
     // An explicitly-named slot is shareable: every dataset on the same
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11305, opened 2026-06-12
fix: DuckDB on-refresh sort breaks refresh for reserved-word column names

claudespice:
## Incorrect behavior

#11284 switched the DuckDB on-refresh `ORDER BY` rendering from unconditional double-quoting to `datafusion::common::utils::quote_identifier`, which only quotes identifiers that are not all-lowercase/digits/underscore — it has no reserved-word awareness. A sort column named after a lowercase reserved word (`order`, `from`, `end`, `case`, ...) now renders bare, so the generated

```sql
CREATE OR REPLACE TABLE "tbl" AS SELECT * FROM "tbl" ORDER BY order ASC
```

fails to parse in DuckDB. The write-completion handler returns `Err`, so **every refresh of the dataset fails from then on**, while queries silently keep serving the stale (or empty, on first load) snapshot. The column can come from user config (`on_refresh_sort_columns`) or be auto-injected by #11284's widened extended schema inference, so the regression is reachable with zero spicepod changes. Before #11284 these names worked (`"order" ASC`).

## Fix

Restore unconditional double-quoting, keeping the embedded-quote escaping that motivated the switch to `quote_identifier`, via an extracted `on_refresh_order_by_clause` helper. The `NULLS FIRST`/`NULLS LAST` rendering from #11284 is preserved.

## Test plan

- [x] `on_refresh_order_by_clause_always_quotes_identifiers` — clause rendering for a reserved-word column, a mixed-case column with an embedded quote, and explicit NULLS placement
- [x] `on_refresh_sort_sql_executes_in_duckdb_for_reserved_word_columns` — executes the exact `CREATE OR REPLACE ... ORDER BY` SQL shape the handler runs against in-memory DuckDB with columns named `order` and `from`, and asserts the rewritten row order (fails on trunk with a DuckDB parse error)
- [x] Both tests pass locally; touched crate compiles clean and is clippy-clean and rustfmt-clean for the changed file
- [ ] `make lint` / `make test` — this checkout cannot run the workspace-wide gates (nested-worktree limitation); deferring to CI, which runs both

### DIFF
--- a/crates/runtime/src/dataaccelerator/duckdb.rs
+++ b/crates/runtime/src/dataaccelerator/duckdb.rs
@@ -43,7 +43,7 @@ use datafusion::error::DataFusionError;
 use datafusion::execution::runtime_env::RuntimeEnv;
 use datafusion::{
     catalog::{Session, TableProviderFactory},
-    common::{Constraints, Statistics, utils::quote_identifier},
+    common::{Constraints, Statistics},
     datasource::{TableProvider, TableType},
     execution::{SendableRecordBatchStream, context::SessionContext},
     logical_expr::{CreateExternalTable, Expr, TableProviderFilterPushDown},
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11325, opened 2026-06-14
perf(cayenne): single-hash composite deletion filter via KeyDeletionIndex::get_batch

phillipleblanc:
## Problem

`KeyBasedDeletionFilterStream` (the composite-primary-key merge-on-read deletion
filter) hashes every row's PK key **twice** per batch:

1. a standalone bloom `might_contain` sweep over all rows, then
2. a per-row `get` probe — which hashes each key **again** before the bloom +
   tier walk.

On upsert / CDC-churn scans (where the deletion index is non-empty, so the bloom
sweep never early-outs) this is the dominant per-row cost on the hottest
query-side loop — the key bytes go through XXH3-128 twice, with the second pass
also paying per-`get` call overhead and poor cache behavior.

The deletion index already ships an optimal batched probe — `get_batch` — that
hashes each key **once** (a tight bloom sweep over the precomputed hashes, then a
delta/base tier walk for the bloom survivors only, probed by hash identity). The
int64 filter is effectively single-pass already; only the composite filter was
left on the two-pass shape.

## Fix

Switch `KeyBasedDeletionFilterStream::poll_next` to `KeyDeletionIndex::get_batch`:

- One hash per row instead of two; the bloom sweep + tier walk are fused.
- `get_batch`'s bound is generalized from `Item = &[u8]` to `Item: AsRef<[u8]>`
  (backward-compatible) so the filter can pass the `RowConverter` `Rows`
  iterator directly without an intermediate `&[u8]` collection.
- The keep mask is built only when there is at least one deletion to apply; the
  no-deletions case returns the batch as-is (subsuming the old bloom-empty and
  `keep_count == batch_size` fast paths in one check).

No behavior change: `get_batch` is proven equivalent to the per-row `get` by the
existing `get_batch_matches_per_row_get_composite` test, and a row with no
tombstone stays visible exactly as before.

## Validation

- **Microbench** (`deletion_index_probe`, new `deletion_index_row_keys_filter_modes`
  group — old `sweep_plus_get` vs new `get_batch`, 8192-row batch):

  | deletion ratio | old | new | speedup |
  |---|---|---|---|
  | 0.01 | 113.3 µs | 32.3 µs | 3.5× |
  | 0.10 | 112.1 µs | 34.7 µs | 3.2× |
  | 0.50 | 103.9 µs | 43.3 µs | 2.4× |

  The composite probe is 2.4–3.5× faster for any batch that has deletions to
  apply (the upsert/CDC case). The old path's redundant sweep made it
  ratio-insensitive (~100 µs); the new path scales with actual deletion density.

- `cargo test -p cayenne` — green. In particular
  `get_batch_matches_per_row_get_composite` verifies `get_batch` returns exactly
  the same per-row results as the previous per-row `get` (across deletion ratios
  and chunk boundaries), and the `filter_exec` behavior tests cover bloom-miss
  (keep all), bloom-hit (filter deleted), and re-inserted/upsert visibility.
  Because the filter only ever *removes* rows and `get_batch` is equivalence-
  tested against the prior probe, query results are unchanged by this PR.
- `make lint-rust` — clean (rustfmt + clippy, pedantic).

### DIFF
--- a/crates/cayenne/benches/deletion_index_probe.rs
+++ b/crates/cayenne/benches/deletion_index_probe.rs
@@ -149,6 +149,78 @@ fn bench_row_keys_probe(c: &mut Criterion) {
     group.finish();
 }
 
+/// Old vs new composite deletion-filter probe shape.
+///
+/// `KeyBasedDeletionFilterExec` used to hash every row's PK key TWICE — a
+/// standalone bloom `might_contain` sweep, then a per-row `get` — before
+/// building the keep mask. Switching the exec to [`KeyDeletionIndex::get_batch`]
+/// hashes each key once (bloom sweep over the precomputed hashes, tier walk for
+/// survivors only). These two lanes isolate that change on the composite hot
+/// loop across deletion ratios.
+fn bench_row_keys_filter_modes(c: &mut Criterion) {
+    let mut group = c.benchmark_group("deletion_index_row_keys_filter_modes");
+    group.throughput(Throughput::Elements(ROWS_PER_BATCH as u64));
+
+    let row_keys = build_row_keys(ROWS_PER_BATCH);
+
+    for ratio in DELETION_RATIOS {
+        let index = build_key_index(ROWS_PER_BATCH, ratio);
+
+        // OLD: bloom `might_contain` sweep (one hash/row) + per-row `get`
+        // (a second hash/row).
+        group.bench_with_input(
+            BenchmarkId::new("sweep_plus_get", format!("ratio={ratio}")),
+            &ratio,
+            |b, _| {
+                b.iter(|| {
+                    let mut maybe = 0_usize;
+                    for key in &row_keys {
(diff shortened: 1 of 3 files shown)


sgrebnov (approved):

phillipleblanc:
End-to-end confirmation under a high-rate HTAP workload (TPC-C-style OLTP at ~10K txn/s with concurrent analytical queries, SF100), comparing the same build with vs. without this change:

- **Correctness preserved** — all 7 tables converge and match; **22/22** analytical queries match.
- **Analytical throughput up ~16%** (QPH 1489 → 1728) — consistent with the faster deletion-filter probe in the microbench above.
- Ingest freshness/convergence unchanged, as expected for a read-path-only change.

No regression on any axis; correctness is also guaranteed by construction via the `get_batch` ↔ per-row `get` equivalence test.

PR #11327, opened 2026-06-15
Improvements for spidapter

krinart:
##  CDC replication metrics in `MetricsResponse`

Spidapter now scrapes the Prometheus endpoint on the running spiced instance at the end of each metrics poll and populates the new `cdc_replication` field with per-table CDC metrics: 
* source receive wait time
* apply burst duration and count
* linger wait
* rows applied
* and bytes applied. 

This works for both local and SCP runs. 

A new `dataset_acceleration_cdc_apply_burst_rows_total` counter was added to spiced to track the true number of records applied per burst, since the existing envelope count didn't capture the per-row volume.

## Optional `--metrics` port for local spiced

Passing `SPIDAPTER_METRICS_PORT` now wires the `--metrics` flag into the local spiced process, enabling Prometheus scraping in local benchmark runs.

## MongoDB scenario switches to Atlas connect mode

The mongodb-streams built-in scenario now connects to an existing MongoDB Atlas cluster via `MONGODB_ATLAS_URI` instead of provisioning a new EC2 instance. 

## Fix Arrow schema mismatch crashing MongoDB CDC coalescing

  Snapshot batches and live-stream CDC batches were built with different field nullability: for example, `_id` was declared non-null in snapshot data but nullable in CDC event batches. When `RefreshTask` tried to coalesce them, Arrow's `concat_batches` rejected the mismatched struct types and crashed the refresh task. The fix normalizes all change batch schemas (snapshot, live CDC events, and the ready-signal envelope) to all-nullable fields, so every batch that flows through the pipeline carries an identical struct type regardless of its origin.

## PostgreSQL and MongoDB EC2 tuning

* PostgreSQL cloud-init now configures memory-aware settings at instance startup: shared_buffers at 25% of available RAM, work_mem at 256 MB, effective_cache_size at 75% of RAM, and parallel query workers enabled. 
* MongoDB oplog size is increased to 10 GB to prevent oplog rollover stalls during long benchmark runs.


.github/workflows/spiced_docker_dev.yml
```diff
@@ -1,6 +1,8 @@
 name: spiced_docker_dev
 
 on:
+  schedule:
```
phillipleblanc: We shouldn't do this, this is what the nightly is for. This workflow is for testing specific changes before they are merged to trunk, not for recreating what the nightly images already do.

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -2440,6 +2449,17 @@ fn concat_change_batches(batches: &[ChangeBatch]) -> crate::accelerated_table::R
     );
 
     let schema = batches[0].record.schema();
+    // Diagnostic: log each batch's full schema so we can see exactly which
+    // envelope's struct/field nullability diverges when concat fails.
```
phillipleblanc:
```suggestion
```

tools/spidapter/scenarios/cayenne.yaml
phillipleblanc, line 1: why are these being removed?

phillipleblanc (approved):

PR #11335, opened 2026-06-15
fix(Vortex): decline only the `InList` membership conjunct of hash-join dynamic filters

sgrebnov:
## 📝 Summary

Follow-up to #11307: push the min/max bounds of hash-join dynamic filters into the Vortex scan (zone pruning) while still declining the expensive `InListExpr`. Replaces the `dynamic_filter_pushdown` opt-in flag with a single, unconditional rule applied at file-open time in `VortexOpener`: split the materialized dynamic filter into conjuncts, keep the bounds, drop the `InList`.

### DIFF
--- a/crates/cayenne/src/provider/table.rs
+++ b/crates/cayenne/src/provider/table.rs
@@ -22828,16 +22828,8 @@ mod tests {
         );
     }
 
-    /// Pins that a DF53 hash-join DYNAMIC filter is NOT absorbed into the Vortex
-    /// scan, even when the fact table has a resident in-memory CDC tier (the
-    /// scan is then union(file branch, bare memory branch)). Vortex's IN-list /
-    /// `list_contains` evaluation has no hashset (O(K×N)) and is slower than
-    /// `DataFusion`'s hashed join probe, so `dynamic_filter_pushdown` is left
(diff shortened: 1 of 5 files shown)


crates/cayenne/src/provider/table.rs
```diff
@@ -22828,16 +22828,8 @@ mod tests {
...
-    /// scan, even when the fact table has a resident in-memory CDC tier (the
-    /// scan is then union(file branch, bare memory branch)). Vortex's IN-list /
-    /// `list_contains` evaluation has no hashset (O(K×N)) and is slower than
-    /// `DataFusion`'s hashed join probe, so `dynamic_filter_pushdown` is left
-    /// default-off (#11307) and the hash join applies the filter. This guards
-    /// against a regression that re-absorbs the dynamic filter into the Vortex
-    /// scan — the slower path.
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    async fn mem_tier_join_does_not_push_dynamic_filter_into_vortex_scan() {
+    async fn mem_tier_join_probe_keeps_dynamic_filter_pushdown() {
```
sgrebnov, line 22832: This restores original test that was changed

phillipleblanc (approved):

PR #11337, opened 2026-06-15
fix: scope SQL UDF arg inlining to args-table columns (fixes #11273)

### LINKED ISSUE #11273: SQL UDF args inliner replaces any unqualified column matching an arg name — stomps derived/aggregate output columns

`crates/runtime-datafusion-udfs/src/user_functions/args_inliner.rs:186-196` — `should_replace` matches unqualified columns by name only (case-insensitive): `None => arg_map.contains_key(&key)`. Derived columns (aggregate aliases, projection aliases) are unqualified `Column`s with `relation = None`, so a SQL function with scalar arg `total` and body

```sql
SELECT total FROM (SELECT sum(value) AS total FROM input)
```

has its outer `total` replaced with the arg literal — the query returns the constant argument value for every row instead of the aggregate. Same for `ORDER BY <alias>` / `HAVING <alias>` colliding with an arg name (ORDER BY constant → arbitrary order). No error; the plan stays valid.

Suggested fix: only inline at the leaf scopes where the function's input relation is in scope (track scope depth / stop at subquery boundaries that re-alias), or reject arg names that collide with derived output columns at function-registration time.

claudespice:
## Summary

`crates/runtime-datafusion-udfs/src/user_functions/args_inliner.rs` inlines a SQL table/scalar function's literal arguments into its body plan so the optimizer can fold and push them down. The replacement step matched **unqualified** columns by name alone:

```rust
None => arg_map.contains_key(&key),
```

Derived output columns — aggregate aliases, projection aliases, unaliased subquery outputs, and `ORDER BY`/`HAVING` references — are also unqualified `Column`s (`relation = None`). So a function whose scalar arg name collided with one of those columns had the column silently replaced by the constant argument value. For example, with a scalar arg `total`:

```sql
SELECT total FROM (SELECT sum(value) AS total FROM input)
```

returned the argument constant for every row instead of the aggregate. No error — the plan stayed valid. `ORDER BY <alias>` / `HAVING <alias>` colliding with an arg name were affected the same way.

## Root cause

Name-only matching for unqualified columns cannot distinguish a reference to the `args` table from a same-named derived column in an outer scope.

## Changes

- Resolve each unqualified column against the **input schema** of the plan node it appears in, and inline only when the name resolves *uniquely* to an `args`-qualified field. Absent or ambiguous → leave untouched (this only forgoes a pushdown optimization, never changes results).
- Qualified `args.*` columns continue to inline directly; columns owned by a base table or a derived scope are never replaced.
- Bare arg references via `CROSS JOIN args` still inline correctly (the name resolves uniquely to `args` in the join's input schema).

## Test plan

- `make lint`
- `make test`

New regression tests in `args_inliner.rs`:
- `inline_does_not_stomp_derived_column_with_arg_name` — the reported bug; fails on old code (arg literal stomped the aggregate alias), passes now.
- `inline_replaces_unqualified_arg_via_cross_join` — confirms legitimate bare arg references still inline.
- `inline_does_not_stomp_unqualified_base_table_column` — a base-table column colliding with an arg name is left alone.

All existing `user_functions` tests (45) continue to pass, including the `CROSS JOIN args` integration tests.

## Performance impact

None for correct queries. The fix can only *reduce* inlining in the ambiguous/derived-name case, where the previous behavior was incorrect; normal qualified and uniquely-resolved args references inline and push down exactly as before.

## Checklist

- [x] Root cause identified and fixed at the shared inliner, not a single call site
- [x] Existing tests pass (`cargo test -p runtime-datafusion-udfs`)
- [x] `cargo clippy -p runtime-datafusion-udfs --all-targets` clean
- [x] `rustfmt --check` clean

Fixes #11273

### DIFF
--- a/crates/runtime-datafusion-udfs/src/user_functions/args_inliner.rs
+++ b/crates/runtime-datafusion-udfs/src/user_functions/args_inliner.rs
@@ -38,7 +38,7 @@ use std::collections::HashMap;
 use arrow::datatypes::Schema;
 use datafusion::{
     common::{
-        Column, Result as DataFusionResult,
+        Column, DFSchema, Result as DataFusionResult,
         tree_node::{Transformed, TreeNode},
     },
     logical_expr::{LogicalPlan, Projection, TableScan, expr::Alias},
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11338, opened 2026-06-15
perf(cayenne): restore union-wiped join-key stats to fix hash-join build-side selection

sgrebnov:
## 📝 Summary

Cayenne combines a table's base + protected snapshots + inlined data with `UnionExec`. DataFusion's generic `col_stats_union` then degrades the merged column stats in two ways:
- NDV (`distinct_count`) is cleared unconditionally — on every union merge, even when all branches carry NDV.
- `min/max` collapse to `Precision::Absent` whenever any branch lacks them (e.g. an empty/stat-less delta), because `Precision::min`/`max` returns `Absent` if either operand is `Absent`.

With NDV gone (and the range gone too when a delta is empty), the join keys carry no usable statistics, so `estimate_inner_join_cardinality` collapses. `JoinSelection` can't tell which side is smaller and places the ~30M-row fact
intermediate on the hash-join build side — 2.24 GB + 1.99 GB hash tables on chbench q7 at SF100.

## Fix

Cayenne already maintains a correct per-table optimizer aggregate (live `min/max` + HLL-derived integer NDV over base+delta). This attaches that aggregate as a column-statistics overlay on the top scan wrapper (`CayenneAccelerationExec`). In `partition_statistics`, the overlay only refills `Precision::Absent` columns — surviving (filter-aware) `num_rows`, `null_count`, and any present column stat are preserved untouched.

This restores the join-key signal `JoinSelection` needs, so the fact stays on the probe side and build-side hash tables shrink to the dimension size.


Cargo.toml
```diff
@@ -385,41 +385,41 @@ lto = "off"
...
-datafusion-physical-plan = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-proto = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-proto-common = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-pruning = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-session = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-spark = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-sql = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-substrait = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
+datafusion = { git = "https://github.com/spiceai/datafusion.git", rev = "3af8cad28bc2dffe6ed3181825b089725b446b1f" }
```
sgrebnov, line 388: Must be updated to official spiceai-53 branch

sgrebnov:
Closing in favor of https://github.com/spiceai/spiceai/pull/11377

PR #11351, opened 2026-06-16
fix: detect dual-write accelerated tables behind the metadata-enrichment wrapper

claudespice:
## Problem

`is_distributed_insert_table` (via `is_dual_write_table_provider` in `crates/runtime/src/datafusion/planner/mod.rs`) decides whether an `INSERT` is rewritten into the distributed-insert path that forwards writes remotely for **dual-write** accelerated tables (the Iceberg federated-catalog cache).

It downcast the catalog's table provider directly to `AcceleratedTable`, or through a single `FederatedTableProviderAdaptor`. But `FederatedTable::new` wraps a provider in a `MetadataEnrichedTableProvider` whenever the dataset declares **table- or column-level metadata** (the same wrapper called out in #11345 and #11339). That wrapper returns itself from `as_any`, and the two wrappers can nest in either order, so a dual-write accelerated table with metadata was hidden behind the wrapper and classified as **not** dual-write.

**User-visible effect:** in scheduler mode, `INSERT` into a dual-write accelerated table that declares metadata silently skipped the distributed-insert rewrite, so the write could be applied only to the local accelerator instead of being dual-written to the federated source — a data-correctness bug that depended on whether the dataset happened to declare metadata. A dataset without metadata worked; adding a column description changed write routing.

## Fix

Peel the known wrapper providers (`FederatedTableProviderAdaptor` and `MetadataEnrichedTableProvider`, which can nest in either order) in a loop before the downcast — matching the canonical peel loop already used in `crates/runtime/src/search/util.rs` and the per-call peeling just added in #11345 / #11339. The peel only routes the downcast; it cannot invent an `AcceleratedTable`, so a non-accelerated wrapped table is still correctly classified as not-dual-write.

This is the same wrapper-hiding-a-downcast class as #11339 and #11345; this sibling call site was missed by those fixes.

## Test plan
- [x] Added regression test `peel_reaches_inner_through_metadata_wrapper` (dual-write detection must see through the metadata wrapper)
- [x] Added edge-case tests: nested metadata wrappers, identity for an unwrapped provider, and a non-accelerated wrapped table not misclassified as dual-write
- [x] `cargo check -p runtime` passes (clean)
- [x] `cargo clippy -p runtime --no-deps` clean on changed code (one pre-existing `needless_return` warning in untouched `builder.rs`)
- [x] `cargo test -p runtime --lib datafusion::planner::tests` — 4 passed, 0 failed
- [x] `rustfmt --edition 2024` applied
- [ ] Note: this is a nested git worktree where workspace-wide `make lint` / `make test` walk up to the parent `Cargo.toml`; the workspace lint/test gates run in CI (Rust Lint / Build and Test).

### DIFF
--- a/crates/runtime/src/datafusion/planner/mod.rs
+++ b/crates/runtime/src/datafusion/planner/mod.rs
@@ -46,6 +46,7 @@ mod update;
 
 use std::sync::Arc;
 
+use data_components::MetadataEnrichedTableProvider;
 use datafusion::catalog::TableProvider;
 use datafusion::error::{DataFusionError, Result as DFResult};
 use datafusion::execution::SessionState;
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

peasee (approved):

PR #11361, opened 2026-06-18
feat(cayenne): re-land seq-prefix bake + fix protected-snapshot probe regression

lukekim:
## Summary

Re-lands the incremental seq-prefix bake (#11326) — reverted by #11358 ("appears to regress performance") — **with the regression diagnosed and fixed**, plus two follow-on fixes the re-land surfaced: the bake's disk-footprint accumulation, and storage-tier-aware + adaptive target file sizing.

## 1. The read regression (diagnosed + fixed)

The bake gated the global-bloom fast-reject behind `min_delete_seq.is_none()`. In CH-bench HTAP the CDC data lives in **protected snapshots**, whose scans probe with `min_delete_seq = Some(S)` — the **dominant query path**. So protected scans skipped the one-shot global-bloom miss→done and walked every frozen run **per row, even for never-deleted keys**.

- Fix: the global bloom is a safe superset over every run's deletion keys (never a false negative), so a **miss means no tombstone at any sequence — `None` for any cutoff**, including `Some(S)`. Apply the fast-reject unconditionally (both Int64 and composite-PK probes); a bloom HIT still falls through to `tombstone_of` with the exact `Some(S)` cutoff, so semantics are unchanged.
- Microbench (`deletion_index_protected_probe_multirun`): protected probe **261.9µs → 74.3µs (−71.6%)**, back to `None` parity. 44 deletion-index correctness tests pass.

### End-to-end CH-bench HTAP A/B (SF=10 local, sqlite, same pod both sides)

| metric | baseline (bake-off) | re-do (bake + fix) | delta |
|---|---|---|---|
| QPH | 16,075.9 | 19,028.7 | **+18.4%** |
| convergence | 109.2 s | 90.2 s | −17% |
| tables drained | 5/6 | 6/6 | ✅ all |
| freshness customer / oorder / new_order | 34.9 / 11.3 / 5.0 s | 5.0 / 1.0 / 0.9 s | −86% / −91% / −82% |

The fix alone recovers the regression **and** beats the bake-off baseline. (An expansive multi-pod sweep showed the win is config-dependent — strong on cdc-memory, regressed on cdc-file — which motivated the two follow-on fixes below.)

## 2. Footprint fix — the bake must fall through to the size-tier leveler

The bake's `run_compaction_trigger` returned early on a successful bake, skipping the size-tier leveling/reclamation pass that same maintenance tick. On a delete-heavy table the bake fires **every** tick, so the leveler never ran: carried-forward runs (full-copy data files) were never leveled and protected snapshots accumulated on disk — the footprint growth the sweep flagged.

- Fix: a successful bake now falls through to `compact_protected_snapshots_subset` the same tick (both already serialize on `compaction_lock`), tracking `baked` so the trigger still reports work done.
- Regression test `run_compaction_trigger_levels_after_bake_not_short_circuit` asserts one trigger reduces the protected set strictly more than the bake alone. (A footprint re-bench is recommended to confirm the on-disk magnitude.)

## 3. Storage-tier-aware + adaptive target Vortex file size

- **Tier-aware default:** on S3 (immutable objects, per-request billing, no fsync), default the target file size to 512 MiB (2× the local default) when the operator hasn't set one — fewer objects, less per-request cost. Local/EBS keep the hardware-derived size.
- **Adaptive actuator:** `Actuator::TargetVortexFileSize` lets the closed-loop tuner grow the target file size for a violated query/QPH goal (fewer, larger files ⇒ better scan throughput + per-file stats, less fan-out to probe), bounded to [½×, 4×] of the configured (tier-aware) size within a 64 MiB–2 GiB envelope. `target_file_size_bytes()` reads the live actuator; pinned to the operator's value when set explicitly (incl. `0` = size-rolling off), so there is no behavior change unless adaptive tuning is on.
- Tests: `adaptive_target_file_size_bounds_scale_with_config`, `query_latency_goal_grows_target_file_size_when_compaction_maxed`.

## Commits
- Reapply #11326 (the original seq-prefix bake).
- `perf(cayenne): restore global-bloom fast-reject for protected-snapshot probes` — the read-path fix + a regression-guard bench.
- `fix(cayenne): bake falls through to the size-tier leveler (footprint reclaim)`.
- `feat(cayenne): storage-tier-aware + adaptive target Vortex file size`.

### DIFF
--- a/crates/cayenne/benches/apply_partial_deletion_filter_per_scan.rs
+++ b/crates/cayenne/benches/apply_partial_deletion_filter_per_scan.rs
@@ -159,5 +159,178 @@ fn bench_apply_partial_deletion_filter_per_scan(c: &mut Criterion) {
     group.finish();
 }
 
-criterion_group!(benches, bench_apply_partial_deletion_filter_per_scan);
+// ============================================================================
+// Rank-1 mechanism: a threshold-bearing (protected-snapshot) probe skips the
+// huge low-seq base run and walks only the small recent runs. This isolates the
+// per-probe GET cost a protected scan pays on bloom-hit (base-resident) rows.
+// ============================================================================
+
+/// Base (bootstrap) run sizes — the accumulated low-seq deletions a protected
+/// snapshot must NOT pay to walk. Large entries leave L2/L3 so the avoided base
+/// GET is a real DRAM miss (the mechanism the lever targets).
+const BASE_RUN_SIZES: &[usize] = &[100_000, 1_000_000, 5_000_000];
+/// Recent applicable deletions living in `active` (seqs above the cutoff).
+const RECENT_DELETES: usize = 2_000;
+/// Bloom-hit probes per iteration (base-resident PKs — the GET fires).
+const PROBE_COUNT: usize = 4_096;
+
+/// A long-lived index under a protected snapshot: a big low-seq base run
+/// (PKs `0..base`, seqs `1..=base`) plus a small recent `active` tier
+/// (seqs `> base`). Returns `(index, S = base)` so a protected probe skips the
+/// base run (`max_delete_seq = base <= S`) and walks only `active`.
+fn build_base_plus_recent(base: usize) -> (Arc<DeletionIndex>, i64) {
+    let mut entries = HashMap::with_capacity(base);
+    for i in 0..base {
+        let pk = i64::try_from(i).expect("base size fits in i64");
+        entries.insert(pk, pk + 1); // delete_seq = pk + 1, so max_delete_seq = base
+    }
+    let mut idx = DeletionIndex::from_map(entries);
+    let cutoff = i64::try_from(base).expect("base size fits in i64");
+    let recent: Vec<(i64, i64)> = (0..RECENT_DELETES)
+        .map(|j| {
+            let off = i64::try_from(j).expect("recent count fits in i64");
+            (cutoff + off, cutoff + 1 + off) // new PKs, seqs strictly above the cutoff
+        })
+        .collect();
+    idx = idx.extend_max_deletes(recent);
+    (Arc::new(idx), cutoff)
+}
+
+/// Lane A (main scan / pre-rank-1 behaviour): `get` fuses every run, so each
+/// bloom-hit pays the base-run GET. Lane B (protected snapshot / rank-1):
+/// `get_with_min_seq(Some(S))` skips the base run and walks only `active`.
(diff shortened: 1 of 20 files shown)


phillipleblanc (approved):

lukekim:
## Local A/B/C validation re-test (24 runs)

3-way — **baseline** (bake-off) · **redo** (bake-on, no A+B) · **abfix** (full PR) — across {cdc-memory, cdc-file, adaptive} × {SF10, SF50}, plus a 3-rep paired disambiguation on cdc-file SF10. md5-distinct binaries; `NO_EXPLAIN_PROBE=1` for clean QPH. Local box = **relative** comparison only (oversubscribed, not goal-comparable).

**✅ Correctness: 24/24 PASSED** — all 7 tables match, Δ0.0000% drift, drain 76–178 s. No regression from A or B anywhere.

**✅ Footprint (the A fix) — reclaimed in all 6 cells (redo → abfix):**

| variant | baseline | redo | abfix |
|---|---|---|---|
| mem SF10 | 6.2 G | 7.6 G | 6.4 G |
| file SF10 | 6.1 G | 7.4 G | 7.3 G |
| adaptive SF10 | 5.9 G | 7.3 G | 6.8 G |
| mem SF50 | 12 G | 16 G | 15 G |
| file SF50 | 12 G | 19 G | 15 G |
| adaptive SF50 | 12 G | 19 G | 14 G |

Mechanism confirmed on file SF50: abfix ran **+28% more compaction passes (2179 vs 1697)** → `order_line` 41 files/3.3 G vs 54/5.0 G. Largest reclaim where the bloat was largest (SF50: 19 G → 14–15 G). The residual above bake-off is the bake's K=3 keep-recent retention (by design, not a leak).

**✅ No QPH regression.** Per-cell QPH on this box is ±50% noisy (the same abfix binary is fastest in some cells, slowest in another). A controlled 3-rep paired comparison on cdc-file SF10 — the one cell that looked like −51% in the single sweep — shows **abfix ≥ redo in every rep** (median 18.7K vs 17.4K). The apparent regression was an anomalously-fast redo sample from a low-contention window, not a slow abfix.

**Compaction healthy everywhere:** read-amp ≤4 (cap 64), 0 failed / 0 OOM passes. The SF50 `order_line` WAL-backlog-rising flag fires on baseline too → the known local-box-can't-drain-the-heaviest-table-at-scale limit (an m7a question), not A/B.

Supersedes the "footprint re-bench recommended" note above — re-bench done, A confirmed.

PR #11377, opened 2026-06-19
feat(optimizer): cost-based left-deep join reordering for Cayenne acceleration

sgrebnov:
## 📝 Summary

Adds `ReorderJoinRule`, a logical `OptimizerRule` that reorders inner-join chains to minimize estimated execution cost using the Ibaraki–Kameda (IK84) left-deep enumeration algorithm. It targets accelerated/Cayenne datasets, whose `TableProvider`s expose row-count and per-column NDV statistics.

Optimizer currently supports connected inner-join chains of ≥3 relations (left-deep, Cayenne only). Unsupported-but-expected cases (no/﹤3 joins, cross products / disconnected graphs, correlated or non-inner joins, already-optimal) silently return the original plan; genuine errors during graph build, enumeration, or reconstruction warn and fall back to the original. The query always runs — only the reorder is skipped.

Note: as specified above, the rule optimizes a limited set of query shapes; wider coverage comes
next (make currently-skipped shapes reorderable):
- Per-connected-component reordering: reorder the connected sub-join of a cross-product query (non-correlated scalar subqueries / disconnected graphs, e.g. q15/q22) instead of skipping the whole query.
- Inner-join chains combined with a semi/anti subquery (`NOT IN` / `NOT EXISTS`) — the `stock⋈item` and `supplier⋈nation` shapes, e.g. q16/q21 — are not reordered yet and fall back to native (the reconstruction can't reliably place their join keys for these shapes). 

## How it works

- Graph build (`join_graph`): flatten the inner-join subtree into relations (nodes) and equi-conditions (edges); non-equi predicates go to a side-channel. Includes implied single-table filter derivation and selective-pendant rewiring.
- Enumeration (`left_deep_join_plan`): per candidate root, build a precedence tree, rank sub-chains by cost/selectivity, merge into one linear chain; pick the lowest-cost root. Polynomial; left-deep only (no bushy plans).
- Cost model (`cost`): cardinality/NDV/selectivity from provider statistics, with bounded handling of computed (`mod`/`ascii`) keys and LIKE/range/IN/eq predicate selectivity.
- Rule (`rule`): best-effort — on any costing error it leaves the plan unchanged (never fails a query). 

## WIP / TODO

- [x] Testing / improvements: 
   - [x] TPC-H https://github.com/spiceai/spiceai/actions/runs/27875408457/job/82493965892
   - [x] TPC-DS https://github.com/spiceai/spiceai/actions/runs/27875424221/job/82494005840
   To be improved/fixed (as part of this or separate PR):
   >9 queries result in unsupported plan (warn only): datafusion_optimizer_rules::logical_plan::reorder_join::rule; unable to apply join reorder; plan left unchanged; Error during planning: Join extraction only supports single-input operators, found 3 inputs in: Union
   - [x] ClickBench https://github.com/spiceai/spiceai/actions/runs/27880820835/job/82507957302
   - [x] Spicebench (clustered mode)
- [x] Single-shot semantics — currently an ordinary `OptimizerRule`, so it re-fires per optimizer pass (≤ `max_passes`). Adopt the MultiJoin-marker pattern (flatten an inner-join island to one node the rule consumes) so a fixpoint framework cannot re-apply it. Done: rule runs once per query: projection pushdown fragments the graph after pass 1, and a `< 3` relations early-exit makes re-entry a no-op (idempotent). Cheap enough today; once the expensive DP/bushy enumerator is added, we can gate it with a bounded `graph_signature` memo+cache so it too runs at most once per distinct graph. Note: DF does not have built-in mechanism to run the rule once.

### Why IK84

Join-order search is the core trade-off between plan quality and planning time. IK84 was chosen as the default as popular option for default always-on rule over all queries.
- Polynomial: finds the optimal *left-deep* order in `O(n² log n)`, no full enumeration,
- Left-deep is the right target for DataFusion: operates by left-deep chains natively (hash-join probe pipelining, predictable memory)

### Future evolution — tiered enumerator (optional bushy / DPhyp)

- IK84 always runs first — produces the default plan plus a cost bound and a linearization seed.
- DPhyp (bushy DP) as an opportunistic upgrade for small `n` (`n < THRESHOLD`, e.g. 12), seeded and cost-bounded by IK84, with a pair/time budget guard. It returns a bushy plan that is `⊇` left-deep, so it is guaranteed no worse than IK84; on budget exhaustion or large `n` it cleanly falls back to the IK84 result.

```text
                ┌─────────────────────────────────────────────┐
  LogicalPlan → │ join_graph.rs   flatten + derive_implied_*   │  (SHARED, exists)
                │  → QueryGraph (nodes=relations, edges=preds) │
                └───────────────────────┬─────────────────────┘
                                        │
                          ┌─────────────▼──────────────┐
                          │  IK84 (left-deep)           │  ALWAYS runs: default plan
                          │  polynomial, ASI cost       │  + linearization seed + cost bound
                          └─────────────┬──────────────┘
                                        │ ik84 order + ik84.cost
                       n < THRESHOLD(12) && bushy_enabled ?
                         │ yes                              │ no (n ≥ THRESHOLD)
            ┌────────────▼─────────────┐                   │
            │ DPhyp (bushy, NEW)       │                   │
            │ seeded + bounded by IK84 │                   │
            │ pair/time budget guard   │                   │
            └─────┬──────────────┬─────┘                   │
       completed  │              │ budget exceeded         │
                  ▼              └───────────┬─────────────┘
          return BUSHY                       ▼
       (⊇ left-deep ⇒ ≤ IK84 cost)     return IK84 (fallback)
                  │                          │
                  └────────────┬─────────────┘
                               │  JoinTree (general binary tree)
                ┌──────────────▼────────────────────────────────┐
                │ cost.rs   estimate_cardinality / cost          │  (SHARED — WEAK LINK)
                │ reconstruct: JoinTree → LogicalPlan            │  (GENERALIZE from left-deep chain)
                └────────────────────────────────────────────────┘
```


test/spicepods/chbench/accelerated/postgres-cayenne[file]-cdc-tuned-sf100.yaml
```diff
@@ -35,8 +35,8 @@ runtime:
     cayenne_footer_cache_mb: '1024'
     cayenne_metastore_cache_mb: '1024'
     cayenne_metastore_mmap_mb: '4096'
-  query:
-    target_partitions: 64
+  # query:
```
Comment (sgrebnov) [inline test/spicepods/chbench/accelerated/postgres-cayenne[file]-cdc-tuned-sf100.yaml]: TODO: remove

crates/runtime/src/datafusion/builder.rs
```diff
@@ -125,6 +126,7 @@ pub static DEFAULT_DATAFUSION_CONFIG: LazyLock<RwLock<SessionConfig>> = LazyLock
         .enable_ident_normalization = false;
 
     df_config.options_mut().optimizer.expand_views_at_output = true;
+    df_config.options_mut().explain.show_statistics = true;
```
sgrebnov: TODO: remove

Cargo.toml
```diff
@@ -385,44 +385,41 @@ lto = "off"
...
-datafusion-proto-common = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-pruning = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-session = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-spark = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-sql = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-datafusion-substrait = { git = "https://github.com/spiceai/datafusion.git", rev = "ebcc99f13aa1269e7bc6849a582708924eb58d46" }
-
-
-
+datafusion = { git = "https://github.com/spiceai/datafusion.git", rev = "3af8cad28bc2dffe6ed3181825b089725b446b1f" }
```
sgrebnov: SHA must be updated after this PR is merged
https://github.com/spiceai/datafusion/pull/168

lukekim:
CI is now green (57 checks pass / 0 fail) and the branch is `MERGEABLE`. Summary of the fixes pushed for the three checks that were failing:

- **Rust Lint** — clippy errors in the new code. The lint phase had never actually run on this branch (the job kept hitting the 6h build timeout at compile), so `redundant_clone` / `doc_markdown` / `ref_option` slipped through. Fixed (e.g. the shared `i`/`f`/`s` `DataType` bindings in `chbench_tests` → `int64()/float64()/utf8()` helpers).
- **MySQL/PostgreSQL quickstart (cayenne)** — `Interval::try_new` assertion on append refresh. The optimizer stats overlay refilled a column's **exact** min/max; the append-refresh dedup filter `ts > watermark` has `watermark == max`, so interval analysis builds the empty interval `[max+1, max]` and errors. Fix: the overlay restores **only `distinct_count` (NDV)**, not min/max — `should_swap_join_order` reads only `total_byte_size`/`num_rows` and `estimate_inner_join_cardinality` prefers NDV, so nothing downstream is lost (`5b5ee87342`).
- **Build and Test** — 3 `filter_converter` tests. Caused by an earlier over-scoped attempt that changed the refresh-filter format for *all* accelerators; reverted in favor of the focused overlay fix above.

Non-blocking follow-ups for later (not gating this PR): cap the NDV product + clamp `1.0 - selectivity` in `cost.rs`; `debug_assert!(connecting.len() == 1)` on the semi/anti reconstruction invariant (`left_deep_join_plan.rs:758`); an optional runtime toggle to disable `reorder_join`.

Ready for `@spiceai/spice-core-maintainers` review.

lukekim (approved):

PR #11398, opened 2026-06-22
fix: /v1/queries chunk row_offset uses cumulative offset, not chunk_index * chunk_size (fixes #11271)

### LINKED ISSUE #11271: /v1/queries chunk API reports row_offset = chunk_index * DEFAULT_CHUNK_SIZE, but chunks are not fixed-size

`crates/runtime/src/http/v1/queries.rs:900-901` computes `row_offset = chunk_index.saturating_mul(DEFAULT_CHUNK_SIZE)`, but chunks are built by appending whole record batches and flushing once `current_chunk_rows >= chunk_size` (`crates/runtime/src/jobs/store.rs:274-312`) — so a chunk routinely holds more than `DEFAULT_CHUNK_SIZE` rows (e.g. 3000-row batches → chunk 0 has 12000 rows).

A client paginating `GET /v1/queries/{id}/results/chunks/{n}` receives chunk 1 labeled `row_offset: 10000` when its true offset is 12000 — clients reconstructing global row positions or resuming mis-attribute/overlap rows at every chunk boundary. Data is complete; the positioning metadata is wrong.

Suggested fix: store each chunk's actual starting row offset (cumulative row count) in the chunk manifest and return that.

claudespice:
## Summary

**Root cause:** `build_chunk_response` in `crates/runtime/src/http/v1/queries.rs` computed `row_offset = chunk_index * DEFAULT_CHUNK_SIZE`. But chunks are not fixed-size: `write_result_chunks_from_stream` appends whole record batches and flushes once `current_chunk_rows >= chunk_size`, so a chunk routinely holds more than `DEFAULT_CHUNK_SIZE` rows (e.g. 3000-row batches → chunk 0 has 12000 rows). A client paginating `GET /v1/queries/{id}/results/chunks/{n}` therefore saw chunk 1 labeled `row_offset: 10000` when its true offset was 12000 — clients reconstructing global row positions or resuming overlapped/mis-attributed rows at every chunk boundary. The row data was complete; only the positioning metadata was wrong.

**Fix:** Record each chunk's actual cumulative starting offset (the row count of all preceding chunks) in `JobResult` while streaming, and return that from `build_chunk_response`.

## Changes

- `jobs/state.rs`: add `chunk_row_offsets: Vec<usize>` to `JobResult`, aligned with `chunk_indices`. Marked `#[serde(default)]` so results persisted before this field existed still deserialize.
- `jobs/store.rs`: track `committed_rows` while flushing chunks and push the cumulative starting offset for each chunk (overflow-checked like the sibling counters); thread it through `build_job_result`.
- `http/v1/queries.rs`: read `job_result.chunk_row_offsets[chunk_index]`; fall back to the legacy `chunk_index * DEFAULT_CHUNK_SIZE` estimate only when the offset is absent (old persisted results), so the handler never panics.

## Test plan

- `make lint`
- `make test`

New/updated tests:
- `jobs::store` `test_chunk_row_offsets_with_oversized_batches` — regression test: 3-row batches with `chunk_size=2` produce true offsets `[0, 3, 6]`, not the `[0, 2, 4]` the old formula gave. Fails on old code, passes on new.
- `jobs::store` existing chunk tests assert `chunk_row_offsets` (`[]`, `[0]`, `[0, 2, 4]`).
- `http::v1::queries` `build_chunk_response_uses_recorded_offset` (offset read from manifest) and `build_chunk_response_falls_back_to_estimate_when_offset_absent` (legacy fallback).

Locally verified the touched `runtime` crate compiles and `cargo test -p runtime` for the affected modules passes (13 tests); `cargo clippy -p runtime` is clean for the changed files (the two remaining warnings are pre-existing in unrelated files). Full `make lint`/`make test` left to CI.

## Checklist

- [x] Root cause identified and fixed at the source
- [x] Backward-compatible deserialization for persisted results
- [x] No change to row data — only corrected positioning metadata

Fixes #11271

### DIFF
--- a/crates/runtime/src/http/v1/queries.rs
+++ b/crates/runtime/src/http/v1/queries.rs
@@ -1181,7 +1181,8 @@ mod tests {
         use arrow::datatypes::{DataType, Field, Schema};
 
         let schema = Arc::new(Schema::new(vec![Field::new("id", DataType::Int32, false)]));
-        let values: Vec<i32> = (0..num_rows as i32).collect();
+        let row_count = i32::try_from(num_rows).expect("row count should fit in i32");
+        let values: Vec<i32> = (0..row_count).collect();
         arrow::array::RecordBatch::try_new(schema, vec![Arc::new(Int32Array::from(values))])
             .expect("to create batch")
     }
--- a/crates/runtime/src/http/v1/queries.rs
+++ b/crates/runtime/src/http/v1/queries.rs
@@ -897,8 +897,15 @@ fn build_chunk_response(
         .map(arrow::array::RecordBatch::num_rows)
         .sum();
 
-    // Calculate row offset based on chunk index using the configured chunk size
-    let row_offset = chunk_index.saturating_mul(DEFAULT_CHUNK_SIZE);
+    // Use the chunk's recorded cumulative starting offset. Chunks are not fixed-size
+    // (they hold whole batches and flush once the row threshold is reached), so the
+    // offset cannot be derived from `chunk_index * DEFAULT_CHUNK_SIZE`. Fall back to that
+    // estimate only for results persisted before per-chunk offsets were recorded.
+    let row_offset = job_result
+        .chunk_row_offsets
+        .get(chunk_index)
+        .copied()
+        .unwrap_or_else(|| chunk_index.saturating_mul(DEFAULT_CHUNK_SIZE));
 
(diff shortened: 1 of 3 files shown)


claudespice:
@copilot review

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #11409, opened 2026-06-23
Move Cayenne types to better files

Jeadie:
## 📝 Summary
`crates/cayenne/src/provider/table.rs` has a bunch of types that should exist in other files. This PR should change no functionality; just move it.


crates/cayenne/docs/index.html
phillipleblanc, line 1: we shouldn't keep HTML files in here

phillipleblanc (approved):

Issue #11411, opened 2026-06-23
Search integration tests failing: trunc(Float64) signature regression + SortPreservingMergeExec order requirements


Jeadie:
## Summary

The scheduled `integration tests (search)` workflow (`.github/workflows/integration_search.yml`, daily `0 1 * * *`) has been **red for 7+ consecutive runs**. The `Run search integration tests` step fails with **30 failed / 546 passed**, and the failures stem from **two distinct DataFusion planning regressions** in the vector-search path:

1. **`trunc(Float64, Int64)` signature regression** (HTTP 400) — ~20/30 failures
2. **`SortPreservingMergeExec` "does not satisfy order requirements"** (HTTP 500) — ~10/30 failures

These are almost certainly correlated (both surfaced together, likely from the same DataFusion bump), but they are **two different code paths needing two different fixes**.

> **Note:** these are **runtime/planning errors, not snapshot drift.** The tests panic at the HTTP call (`crates/runtime/tests/search/mod.rs:467`) *before* the `insta` snapshot assertion runs, so **no `.snap` diff is produced**. The auto-generated snapshot-update PR (#11193) contains **zero `.snap` changes** as a result — confirming there is nothing to "just accept".

Affected runs (all `trunk`, `schedule`):
- [run 27588707702](https://github.com/spiceai/spiceai/actions/runs/27588707702) — failure (36m)
- [run 27660330313](https://github.com/spiceai/spiceai/actions/runs/27660330313) — failure (29m)
- [run 27996182751](https://github.com/spiceai/spiceai/actions/runs/27996182751) — current

---

## Regression 1 — `trunc(Float64, Int64)` signature (HTTP 400)

The search SQL test queries round `_score` with `trunc(_score, 3)`. DataFusion's `trunc` signature was tightened and no longer accepts `(Float64, Int64)`:

```
HTTP error: 400 Bad Request - Failed to execute query: Error during planning:
Function 'trunc' requires TypeSignatureClass::Native(LogicalType(Native(Date), Date)),
but received Float64 (DataType: Float64). No function matches the given name and
argument types 'trunc(Float64, Int64)'. You might need to add explicit type casts.
```

**Source of the failing call** — `crates/runtime/tests/search/megascience.rs:299-320`, e.g.:

```rust
// megascience.rs:299
"SELECT id, answer, trunc(_score, 3) FROM vector_search(qs, 'second', answer) order by _score desc, id LIMIT 4"
// megascience.rs:317
[...]

PR #11415, opened 2026-06-23
fix(search): restore numeric trunc and fix SortPreservingMergeExec planning error (DF54)

Jeadie:
## Summary

Fixes two DataFusion 54 regressions that have left the scheduled `integration tests (search)` workflow (`.github/workflows/integration_search.yml`) red, failing **30 of 576** search permutations. Both surfaced with the DF54 upgrade (#11360).

Refs #11411.

## Regression 1 — `trunc(Float64, Int64)` (HTTP 400)

DataFusion 54's Spark crate ships a **date**-truncation `trunc` UDF. `datafusion_spark::register_all` overwrites any same-named function, so it shadowed the built-in **numeric** `trunc(value, scale)`. The search SQL `trunc(_score, 3)` (`_score` is `Float64`) then failed planning:

```
Error during planning: Function 'trunc' requires
TypeSignatureClass::Native(LogicalType(Native(Date), Date)), but received Float64
... No function matches the given name and argument types 'trunc(Float64, Int64)'.
```

**Fix:** in `crates/runtime/src/datafusion/builder.rs`, register every Spark scalar function **except** `trunc` (plus the aggregate/window functions, mirroring `register_all`), keeping the built-in numeric `trunc` registered by `with_default_features()`. No new UDF is registered and the test SQL is unchanged.

## Regression 2 — `SortPreservingMergeExec` (HTTP 500)

The candidate-retrieval plan failed physical planning under DF54:

```
... does not satisfy order requirements: [_score@1 DESC NULLS LAST, id@0 ASC]. Child-0 order: []
```

**Root cause is a DataFusion `EnforceSorting` bug** (fixed in spiceai/datafusion#170; tracked in #11426). During sort pushdown the candidate `SortExec` is relocated below the column-reordering candidate `ProjectionExec`, but its sort-key column index is forwarded to the child **unchanged** instead of being remapped into the child schema — so the key (`_score`) ends up pointing at the wrong column.

A search-side workaround (projecting `_score` last) was attempted and **does not work**: `_score` is a *computed* column in the candidate projection (`1 - cosine_distance(…) as _score`), so reordering the projected columns only shifts the stale index (`_score@1` → `_score@2`) and additionally changes the result column order (≈110 snapshot mismatches).

**Fix:** the engine-level fix in spiceai/datafusion#170 — remap the ordering requirement through the projection's column mapping, and decline pushdown (keep the sort *above* the projection) when a required column maps to a computed expression. This PR bumps the runtime's `datafusion` git rev to a `spiceai-54`-based commit that includes that fix. `crates/search/src/pipeline.rs` is left unchanged from `trunk`.

## Notes / follow-ups (out of scope here)

- The `datafusion` rev here points at a branch commit carrying the `EnforceSorting` fix on top of the rev this PR already used. Once spiceai/datafusion#170 lands on `spiceai-54`, the rev should be repointed to the merged `spiceai-54` HEAD.
- These fixes address the *current* (post-DF54, 2026-06-19) failures. The last green run was **2026-05-28**, so there is a separate earlier breakage in the May 28 → June 19 window worth investigating.
- The workflow's `Push snapshots to branch` step can hang until GitHub's default 360-min job timeout (several runs cancelled at exactly 6h0m); adding a `timeout-minutes` and hardening that step is a separate CI follow-up.

## Testing

- The DataFusion `EnforceSorting` fix has a `physical_optimizer` regression test (`test_parallelize_sorts_remaps_index_through_reordering_projection`) reproducing the exact `Child-0 order: []` failure; verified red→green with no new failures vs. the pre-existing fork test baseline (spiceai/datafusion#170).
- End-to-end confirmation comes from re-running `integration_search.yml` on this branch with the rev bump.


phillipleblanc (dismissed):

Jeadie:
Testing it on https://github.com/spiceai/spiceai/actions/runs/28003718621

phillipleblanc (dismissed):

Jeadie:
Re-ran `integration_search.yml` on this branch (`18128bc331`, with the `datafusion` rev bump): https://github.com/spiceai/spiceai/actions/runs/28139340107

phillipleblanc (approved):

PR #11418, opened 2026-06-23
fix(snowflake): Int64 timestamp arm ignores column scale; TIME never cast (fixes #11268)

### LINKED ISSUE #11268: Snowflake: Int64 timestamp arm ignores column scale (TIMESTAMP(3)/(6) -> all-NULL or wrong instants); TIME declared Time64 but never cast
Found during a data-correctness audit of trunk. Two wire-format dispatch gaps in the Snowflake connector:

**1. Int64 timestamp arm assumes seconds precision** (`crates/db_connection_pool/src/dbconnection/snowflakeconn.rs:433-451`): Snowflake's Arrow wire format sends TIMESTAMP_NTZ/LTZ/TZ as a plain Int64 in units of 10^-scale seconds for scale ≤ 7 (the struct{epoch,fraction} encoding is only used for scale 8-9). The code always multiplies by 1e9 (`checked_mul(NANOSECONDS)`), ignoring the field's `scale` metadata (which the `fixed` arm *does* consult). For a `TIMESTAMP_NTZ(3)` column, modern epochs in ms (~1.7e12) overflow i64 on x1e9 → `checked_mul` returns None → **every value silently becomes NULL**. Scale 0 and 8-9 work; explicitly-typed TIMESTAMP(3)/(6) columns are common.

**2. `TIME` declared but never cast** (`snowflakeconn.rs:576` and `crates/data_components/src/snowflake/mod.rs:668` declare `Time64(Nanosecond)`; the `snowflake_schema_cast` dispatcher at `snowflakeconn.rs:324-383` has arms only for `timestamp_ntz|timestamp_ltz`, `timestamp_tz`, `fixed`, with `_ => {}` passthrough): the raw scaled Int64 (units of 10^-scale seconds since midnight) ships under a schema that says Time64 — e.g. `45045123` for `12:30:45.123` at scale 3.

Both follow the known "schema-cast dispatcher must exhaust all variants the type parser declares" pattern (cf. the `timestamp_ltz` gap fixed in #10493). Minor related: the cast arms rebuild fields with bare `Field::new`, dropping Snowflake `logicalType`/`precision`/`scale` field metadata (`snowflakeconn.rs:334-366`).

claudespice:
## Summary (root cause)

Two wire-format dispatch gaps in the Snowflake connector's `snowflake_schema_cast` (`crates/db_connection_pool/src/dbconnection/snowflakeconn.rs`), both data-correctness bugs:

1. **Int64 timestamp arm assumed seconds precision.** Snowflake sends `TIMESTAMP_NTZ/LTZ/TZ` with scale ≤ 7 as a primitive `Int64` in units of `10^-scale` seconds (the `struct{epoch, fraction}` encoding is only used for scale 8–9). The code unconditionally multiplied by `1e9`, ignoring the column's `scale` metadata that the `fixed` arm already consults. For a `TIMESTAMP_NTZ(3)` column, modern epochs in milliseconds (~1.7e12) overflowed `i64` in `checked_mul` → `None` → **every value silently became NULL**. Scale 0 and 8–9 worked; explicitly-typed `TIMESTAMP(3)`/`(6)` columns did not.

2. **`TIME` was declared but never cast.** The type parser declares `Time64(Nanosecond)`, but the dispatcher had arms only for `timestamp_ntz|ltz`, `timestamp_tz`, and `fixed`, with a `_ => {}` passthrough. The raw scaled `Int64` (units of `10^-scale` seconds since midnight) shipped unconverted under a schema claiming nanoseconds.

Follows the documented "schema-cast dispatcher must exhaust all variants the type parser declares" pattern (cf. the `timestamp_ltz` gap fixed in #10493).

## Changes

- Derive the seconds→nanoseconds multiplier from the column `scale` (`10^(9-scale)`) in the primitive `Int64` timestamp arm, via a new `scale_to_nanos_multiplier` helper. Out-of-range scale (outside `0..=9`) fails loudly; per-row overflow still yields NULL, matching the struct path.
- Add a `time` dispatch arm + `cast_sf_time_to_arrow_time64` that casts the scaled `Int64` to `Time64(Nanosecond)` with the same multiplier.
- Preserve Snowflake field metadata (`logicalType`/`precision`/`scale`) by cloning the field with `with_data_type` instead of rebuilding with `Field::new`, which dropped it (applied to the timestamp, time, and fixed arms).

The struct `{epoch, fraction}` encoding (scale 8–9) is scale-independent — epoch is whole seconds, fraction is already nanoseconds — so it is unchanged.

Scope: Snowflake only. This Int64-in-`10^-scale`-seconds encoding is Snowflake-specific; other connectors' timestamp paths were addressed separately (#10335/#10255/#10493).

## Performance impact

None. The change replaces one constant multiplier with a `scale`-derived one computed once per column; no extra scans, allocations, or query-plan changes.

## Test plan
- [x] Added regression test (`test_cast_sf_timestamp_ntz_millisecond_scale_not_nulled`) reproducing the all-NULL overflow for a millisecond-precision timestamp — fails on old code, passes now
- [x] Added edge case tests for scale 6 (micros), `TIME` at scale 3 (incl. NULL rows), `scale_to_nanos_multiplier` (scales 0/3/6/9 + out-of-range → None), and end-to-end `snowflake_schema_cast` dispatcher tests asserting non-NULL values AND field-metadata preservation
- [x] `make test` passes

Fixes #11268

### DIFF
--- a/crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
+++ b/crates/db_connection_pool/src/dbconnection/snowflakeconn.rs
@@ -21,7 +21,7 @@ use std::time::Instant;
 
 use arrow::array::{
     Array, ArrayRef, AsArray, Decimal128Array, Int32Array, Int64Array, PrimitiveArray, RecordBatch,
-    StringArray, StructArray, TimestampNanosecondArray,
+    StringArray, StructArray, Time64NanosecondArray, TimestampNanosecondArray,
 };
 use arrow::datatypes::{
     ArrowPrimitiveType, DataType, Field, Int8Type, Int16Type, Int32Type, Int64Type, Schema,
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11434, opened 2026-06-24
fix(cluster): keep built-in avg over Spark avg (distributed aggregate state schema mismatch)

phillipleblanc:
## Problem

`datafusion_spark::register_all` registers functions **over** any same-named built-in. Its `avg` aggregate uses a different partial-aggregate **state layout** than the built-in:

| | state_fields |
|---|---|
| built-in `avg` (functions-aggregate) | `[count: UInt64, sum]` |
| Spark `avg` (datafusion-spark) | `[sum, count: Int64]` |

Single-node this is invisible (the same UDAF computes and merges state). But in **distributed** execution the **scheduler** bakes the shuffle/stage schema from the resolved `avg`'s `state_fields`, while the **executor** runs the `avg` its own registry resolves — and the coalescing shuffle reader then downcasts a state column with the wrong primitive type:

```
thread 'task_runner' panicked at arrow-array/src/cast.rs: primitive array
  InProgressPrimitiveArray::<T>::copy_rows  (BatchCoalescer)
  ballista CoalescedShuffleReaderStream  ->  GroupedHashAggregateStream (final agg)
```

This makes any distributed query with `avg` fail (e.g. TPC-H q1).

## Fix

Same pattern as #11415 (Spark `trunc` shadowing numeric `trunc`): register every Spark scalar function **except `trunc`** and every Spark aggregate **except `avg`**, keeping the built-ins. (The non-colliding Spark aggregates `collect_list`/`try_sum` are still registered.)

Validated in a distributed SF10 cluster run: with this fix q1 runs to completion 5/5; without it q1 panics every time. Refs #11415 (incorporates its `trunc` skip so this lands independently; rebase to drop the `trunc` part if #11415 merges first).


phillipleblanc:
Addressed @copilot review in `6be63f0c`: panic messages on Spark fn registration now include the failing function name (`Unable to register Spark {scalar,aggregate,window} function \`{name}\`: {e}`), and the in-code reference is now the fully-qualified `spiceai/spiceai#11415`.

lukekim (approved):

Issue #22, opened 2021-08-15
Provide distinct error for registry between no access and not found


lukekim:
Not found is SPICE_GH_TOKEN not set during private preview.

PR #11435, opened 2026-06-24
fix: offload blocking sync calls off the primary async runtime

phillipleblanc:
## Summary

Fixes a class of footgun surfaced by an audit: **synchronous blocking calls executed on a Tokio async worker of the primary runtime**. These pin worker threads instead of yielding, which can starve `/health` and lead Kubernetes to kill the pod. The fix offloads the blocking work onto `tokio::task::spawn_blocking` (or a dedicated runtime) so the async runtime keeps making progress. The `cpu_runtime`/refresh/compaction managed runtimes are intentionally exempt and untouched.

Dependency **spiceai/datafusion-table-providers#22** is merged into `spiceai-54`; `Cargo.toml` is pinned to the `spiceai-54` tip that includes it.

### Dependency offloads (table-providers, #22, incl. datafusion-contrib/datafusion-table-providers#672)
ODBC/ADBC/DuckDB pool `connect()`, ODBC `execute`/`get_schema`, and the generic `get_tables`/`get_schemas`/`get_schema` sync→async bridge offload via `spawn_blocking` (guarded by `run_async_with_tokio`).

### Per-request model/search HTTP handlers (the highest-impact: per request, on the primary runtime)
- `/v1/models/{name}/predict` — ONNX/tract inference: the ML-model store now holds `Arc<Model>`; `run_inference` clones a handle, **drops both the models and app read-locks**, and runs inference in `spawn_blocking`.
- `/v1/embeddings` — Model2Vec forward pass: `StaticModel` wrapped in `Arc`, offloaded in `embed`. **Also fixes** the query-time embedding path (UDTF / vector-index scans go through `embed().await`).
- `/v1/search` (+ SQL `text_search`) — tantivy search: `Arc`-shared searcher, `search_query_literal` offloaded per page.

### Connectors (registration / catalog-refresh time)
- **Oracle** (`rust-oracle`, sync OCI): `get_schema`, catalog `list_schemas`/`list_tables`, and pool `on_acquire` offloaded via a new `OracleConnectionPool::get_owned()` (`'static` bb8 conn) + `spawn_blocking`.
- **Kafka/Debezium schema-peek**: `fetch_metadata`/`fetch_watermarks` (blocking up to the ~30s peek budget) offloaded; `!Send` `Metadata` reduced to `Vec<i32>` inside the blocking task.

## Investigated → not a footgun

**Object-store S3 registry** (`runtime-object-store`): the audit flagged blocking DNS + `io_runtime.block_on` in the sync `ObjectStoreRegistry::get_store`. But `io_runtime == Handle::current()` (the primary runtime), and `Handle::block_on` **panics** if called from an async worker — so the existing (non-panicking) `block_on` proves `get_store` runs on **non-async** (blocking-pool) threads, i.e. it does **not** block the async runtime. Left as-is; flagged for the team to confirm with DataFusion-internals knowledge.

## Follow-up (separate PR)
- **`spice_sys` DuckDB metadata offload** (dataset_checkpoint / kafka / debezium / mongodb / dynamodb): these synchronous DuckDB writes also block the runtime, but they rely on *implicit serialization* (they run inline on the caller's runtime). Naively wrapping them in `spawn_blocking` enables true parallelism that single-writer DuckDB rejects (`TransactionContext Error: Conflict on update!` — caught by `test_duckdb_concurrent_same_partition_keeps_max`). Correct offload needs write-serialization (per-instance async mutex) or conflict-retry; deferred so it can be designed + tested properly.
- Lower-impact MEDIUM/LOW items: tantivy index commit (write path), tokenizer chunking, dataformat-json `FileOpener` (cpu_runtime-exempt on `/v1/sql`), `view.rs` sync read, Vault CA-cert read, cluster-mTLS startup reads, Pingora-gated `lru_cache` `block_on`.

## Validation
Per-crate `cargo check` for every touched feature set; `make lint-rust` (cargo fmt + workspace clippy with `-Dpedantic/unwrap_used/clone_on_ref_ptr/redundant_clone`) clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6369,7 +6369,7 @@ dependencies = [
 [[package]]
 name = "datafusion-table-providers"
 version = "0.1.0"
-source = "git+https://github.com/spiceai/datafusion-table-providers.git?rev=3d8eca098dc5aefe361625b834ad834d2b846137#3d8eca098dc5aefe361625b834ad834d2b846137"
+source = "git+https://github.com/spiceai/datafusion-table-providers.git?rev=d878f0de2464#d878f0de2464cd5dcc4168abf38f2b426604b6d0"
 dependencies = [
  "adbc_core",
  "adbc_driver_manager",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6369,7 +6369,7 @@ dependencies = [
 [[package]]
 name = "datafusion-table-providers"
 version = "0.1.0"
-source = "git+https://github.com/spiceai/datafusion-table-providers.git?rev=d878f0de2464#d878f0de2464cd5dcc4168abf38f2b426604b6d0"
+source = "git+https://github.com/spiceai/datafusion-table-providers.git?rev=5789a8391a71553d4d38e2d2399c1099fbf12265#5789a8391a71553d4d38e2d2399c1099fbf12265"
 dependencies = [
  "adbc_core",
  "adbc_driver_manager",
@@ -6396,6 +6396,7 @@ dependencies = [
  "geo-types",
  "hickory-resolver",
  "itertools 0.14.0",
+ "logos",
  "mongodb",
  "mysql_async",
  "native-tls",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -185,7 +185,7 @@ datafusion-proto-common = "54.0.0"
 datafusion-pruning = "54.0.0"
 datafusion-spark = "54.0.0"
 datafusion-substrait = "54.0.0"
-datafusion-table-providers = { git = "https://github.com/spiceai/datafusion-table-providers.git", rev = "3d8eca098dc5aefe361625b834ad834d2b846137" } # spiceai-54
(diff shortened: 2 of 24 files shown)


phillipleblanc:
Addressed the Copilot review feedback in 3a103799:

- **Abbreviated git SHA (Cargo.toml):** re-pinned `datafusion-table-providers` to the full 40-char `spiceai-54` tip (`5789a8391a71553d4d38e2d2399c1099fbf12265`) now that #22 is merged.
- **app read-lock held across `.await` (inference.rs):** `run_inference` now resolves the model version and **drops the app read-guard** before any heavy await, and clones the `Arc<Model>` handle out of the models lock (dropping that guard too) before running inference in `spawn_blocking`.

Also: the `spice_sys` DuckDB metadata offload has been removed from this PR — `spawn_blocking` removed the implicit serialization those single-writer DuckDB writes relied on (`test_duckdb_concurrent_same_partition_keeps_max` hit `Conflict on update!`). It'll return in a follow-up with proper write-serialization. `make lint-rust` is clean locally.

peasee (approved):

PR #11446, opened 2026-06-24
feat(runtime): cost-based eager aggregation, enabled by default (datafusion spiceai-54)

bjchambers:
## What

Enables the cost-based **eager-aggregation** physical optimizer rule (datafusion `spiceai-54`, spiceai/datafusion#172) in spiced, and exposes it as `runtime.query` configuration:

```yaml
runtime:
  query:
    eager_aggregation: true                     # default (on); set false to disable
    eager_aggregation_min_reduction_factor: 4   # optional; benefit gate (DataFusion default 4)
    eager_aggregation_max_pushed_groups: 0      # optional; group-count cap, 0 = uncapped (DataFusion default)
```

The rule pushes a partial aggregation below a join when a statistics-based cost model predicts a large row reduction, then re-aggregates above the join. Two independent cost gates:

- `eager_aggregation_min_reduction_factor` — only push if the pre-aggregation collapses its input rows by at least this factor (benefit gate).
- `eager_aggregation_max_pushed_groups` — decline the push when the estimated pushed-down group count would exceed this value; `0` = uncapped (cost/memory gate).

## Defaults

spiced defaults `eager_aggregation` to **true** (DataFusion's own default is off). The two tuning knobs are left at the DataFusion defaults when unset.

## Changes

- Repin `datafusion-*` to `spiceai-54` HEAD, which now contains the merged eager-aggregation rule (spiceai/datafusion#172) plus its follow-up cleanups.
- Add `runtime.query` config wired through the runtime + DataFusion builders, mirroring the existing `prefer_hash_join` plumbing.
- Remove the experimental env-var toggle (`SPICED_EAGER_AGGREGATION*`) and the benchmark-only EXPLAIN/decline-map capture scaffolding (testoperator `--capture-explain`, the test-framework log-tee/env forwarding, and the CI dispatch toggle) — superseded by the pod config.


bjchambers:
## Cleanup pass — addressing review feedback

This PR was refactored from the original benchmarking experiment into a minimal, mergeable feature. Here's how the review comments map to the current state:

**Cargo.toml scope (the big one — thanks for flagging).** Earlier merges had dragged unrelated dependency changes into `Cargo.toml`: `aws-sdk-*` reverted to default features (reintroducing the legacy `rustls 0.21` / `hyper-rustls 0.24` stack alongside the modern one), `tiberius` switched from the spiceai git fork to the crates.io release, and `spice-rs` / `system-adapter-protocol` rev drift. `Cargo.toml` and `Cargo.lock` are now reset to `trunk` with **one** change: the `spiceai-54` datafusion patch rev bumped to the eager-aggregation HEAD (`33efb29f`). The dual TLS stack is gone from `Cargo.lock`. The branch-vs-SHA / provenance comments are moot — it's a full 40-char `rev` on the `# branch: spiceai-54` entries.

**Env-var toggle removed.** The `SPICED_EAGER_AGGREGATION*` env parsing (and the comments about its silent parse failures and `=1` vs `=true` wording) no longer exists — the rule is now configured via `runtime.query` spicepod YAML, defaulting on.

**Benchmark scaffolding backed out.** The testoperator `--capture-explain` command, `capture.rs` (the UTF-8 byte-slicing and whole-log-read concerns), the test-framework log-tee, and `chbench_explain_capture.sh` were removed from scope. The decline-map capture also depended on eprintln debug output that was trimmed before the rule merged into spiceai-54, so it no longer produced output anyway.

**Cayenne HLL doc comments** (`column_stats.rs`, `hll.rs`) aren't part of this PR's diff — those changes are on trunk.

**Live feedback on the feature code:**
- ✅ `eager_aggregation_min_reduction_factor = 0` now warns and is ignored (mirrors the existing `target_partitions` guard). `eager_aggregation_max_pushed_groups = 0` stays valid (means uncapped).
- ✅ Added `test_eager_aggregation_wires_through_to_session_config` covering the spiced default (on), an explicit disable, and the two tuning knobs.
- `eager_aggregation` kept as `Option<bool>` rather than an enum — consistent with the sibling `prefer_hash_join: Option<bool>` in the same `Query` struct, and there's no repo guideline against booleans in config (many existing fields are boolean).

Resolving the addressed and outdated threads accordingly.

sgrebnov (dismissed):

crates/spicepod/src/component/runtime.rs
```diff
@@ -965,6 +965,28 @@ pub struct Query {
...
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub prefer_hash_join: Option<bool>,
+
+    /// Enable the cost-based eager-aggregation physical optimizer, which pushes a
+    /// partial aggregation below a join when a statistics-based cost model
+    /// predicts a large row reduction, then re-aggregates above the join. Maps to
+    /// `datafusion.optimizer.enable_eager_aggregation`. Defaults to `true` in
+    /// spiced (`DataFusion`'s own default is `false`); set to `false` to disable.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub eager_aggregation: Option<bool>,
```
sgrebnov, line 975: We should have it as `enabled`/`disabled` per general guideline / consistency - can be done as separate improvement

sgrebnov (dismissed):

sgrebnov (approved):

PR #11460, opened 2026-06-25
fix: resolve table-reference qualification in results-cache invalidation (fixes #11266)

### LINKED ISSUE #11266: Results cache invalidation uses exact TableReference equality — differently-qualified cached entries survive refresh/DML invalidation

`crates/cache/src/lru_cache.rs:485-501` invalidates with `value.as_table_refs().contains(&table_ref)` — exact `TableReference` equality. Cached entries store table refs **as written in the SQL** (`crates/cache/src/utils.rs:260-274`, e.g. bare `customer` or fully-qualified `spice.public.customer`), while invalidators pass different qualifications:

- accelerated refresh completion invalidates with the dataset name (`crates/runtime/src/accelerated_table/refresh.rs:1140`, bare `foo`)
- `INSERT INTO foo` invalidates `Bare{foo}` (`crates/runtime/src/datafusion/query.rs:768`, `flight/flightsql/statement_update.rs:104`)

A cached `SELECT * FROM spice.public.foo` (entry holds `Full{spice,public,foo}`) survives both, and the client keeps receiving pre-mutation rows as a fresh `CacheHit` until TTL. The runtime already knows raw equality is unreliable — `crates/runtime/src/datafusion/mod.rs:4002-4007` defines `resolved_equality` (resolve both to `spice.public`) — but the cache predicate doesn't use it.

Impact scales with configured TTL (default 1s limits the window; operator-configured 10m+ doesn't). Suggested fix: resolve both sides to fully-qualified form (the `resolved_equality` semantics) inside the invalidation predicate, and store resolved refs in `as_table_refs`.

Related (smaller): on the generic HTTP DML path, `invalidate_for_table` runs at **planning** time (`query.rs:762-774`) — before the mutation executes — so a concurrent SELECT can repopulate the cache with pre-mutation rows for the full TTL; the Flight paths invalidate post-execution.

claudespice:
## Summary

Results-cache invalidation used **exact `TableReference` equality**
(`value.as_table_refs().contains(&table_ref)`), so a cached entry whose stored
ref differs in qualification from the invalidator's ref — for the same physical
table — was never invalidated.

Cache entries record table refs exactly as written in the originating SQL
(`crates/cache/src/utils.rs` `get_logical_plan_input_tables`): `SELECT * FROM spice.public.foo`
stores `Full{spice,public,foo}`, while `SELECT * FROM foo` stores `Bare{foo}`.
Invalidators pass a *different* qualification for the same table:

- accelerated-refresh completion invalidates with the bare dataset name (`accelerated_table/refresh.rs`)
- `INSERT INTO foo` invalidates `Bare{foo}` (`datafusion/query.rs`, `flightsql/statement_update.rs`)

So a cached `SELECT * FROM spice.public.foo` survived refresh and DML invalidation and
kept returning pre-mutation rows as a fresh `CacheHit` until TTL expiry. Impact scales
with the configured TTL (default 1s limits the window; operator-configured 10m+ does not).

## Root cause

The invalidation predicate used raw `HashSet::contains`. The runtime already knows raw
equality is unreliable — `datafusion/mod.rs` defines `resolved_equality` (resolve both
sides to `spice.public`) — but the cache predicate didn't use that semantics.

## Changes

- Add `resolved_table_match(stored, target)` to `crates/cache/src/lib.rs`, which resolves
  both the stored refs and the invalidating ref to fully-qualified `catalog.schema.table`
  form (via `TableReference::resolve`) before comparing — mirroring the runtime's
  `resolved_equality`. Defaults are `spice`/`public`, duplicated locally because `cache`
  cannot depend on `runtime-datafusion` without a dependency cycle.
- Use it in all three invalidation paths: the Moka and Pingora LRU backends
  (`lru_cache.rs`) and `SimpleCache` (`simple_cache.rs`).
- Regression test (`test_cache_invalidate_resolves_qualification`) covering both
  directions (full<->bare, partial<->bare) plus negative cases (different table, different
  non-default schema must NOT be invalidated). The positive cases fail on the old
  `contains`-based predicate.

Scoped to the equality semantics — the core reported bug, spanning all three backends.

## Out of scope (follow-up)

The issue also notes a smaller *timing* concern: on the generic HTTP DML path,
`invalidate_for_table` runs at **planning** time (`query.rs`), before the mutation
executes, so a concurrent SELECT can repopulate the cache with pre-mutation rows. That's
a separate, more invasive change (move invalidation to post-execution like the Flight
paths) and is left for a dedicated PR.

## Test plan

- `make lint`
- `make test`

## Checklist

- [x] Root cause identified and fixed at the shared predicate (all backends)
- [x] Regression test added (fails on old code, passes on new)
- [x] `cargo clippy -p cache --all-targets` clean
- [x] `cargo test -p cache --lib` green (77 passed)
- [x] `cargo fmt` applied

Fixes #11266

### DIFF
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -147,7 +147,10 @@ pub const SPICE_DEFAULT_SCHEMA: &str = "public";
 /// until TTL expiry. Resolving both sides first makes the comparison robust to
 /// qualification differences.
 #[must_use]
-pub fn resolved_table_match(stored: &HashSet<TableReference>, target: &TableReference) -> bool {
+pub fn resolved_table_match<S: std::hash::BuildHasher>(
+    stored: &HashSet<TableReference, S>,
+    target: &TableReference,
+) -> bool {
     let resolved_target = target
         .clone()
         .resolve(SPICE_DEFAULT_CATALOG, SPICE_DEFAULT_SCHEMA);
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -124,6 +124,41 @@ impl AsTableRefs for LogicalPlan {
     }
 }
 
+/// Default catalog and schema used to resolve table references before comparing
+/// them during cache invalidation.
+///
+/// These mirror `runtime_datafusion::SPICE_DEFAULT_CATALOG` /
+/// `SPICE_DEFAULT_SCHEMA`. They are duplicated here because the `cache` crate
+/// cannot depend on `runtime-datafusion` without forming a dependency cycle
+/// (`runtime-datafusion -> search -> datafusion-optimizer-rules -> cache`).
+pub const SPICE_DEFAULT_CATALOG: &str = "spice";
+pub const SPICE_DEFAULT_SCHEMA: &str = "public";
+
+/// Returns `true` if `target` refers to the same physical table as any reference
+/// in `stored`, after resolving both sides to fully-qualified
+/// (`catalog.schema.table`) form.
+///
+/// Cache entries record table references exactly as written in the originating
+/// SQL (e.g. bare `customer` or fully-qualified `spice.public.customer`), while
+/// invalidators — accelerated-refresh completion, `INSERT INTO`, and other DML —
+/// may pass a differently-qualified reference for the same table. Plain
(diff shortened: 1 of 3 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11467, opened 2026-06-25
fix(search): default Elasticsearch kNN candidate pool to 1000 instead of 10 (fixes #11264)

### LINKED ISSUE #11264: Elasticsearch kNN: k hardcoded to 10 — search limit silently capped when filter/sort blocks limit pushdown

`crates/search/src/index/elasticsearch/mod.rs:306` constructs `ElasticsearchKnnTable { k: 10, ... }` and `crates/data_components/src/elasticsearch/search_table.rs:105` uses `let effective_k = limit.unwrap_or(self.k)`. The search pipeline places `Filter` (and `Sort`) above the index scan (`crates/search/src/provider.rs:591-613`, `crates/search/src/pipeline.rs:184-195`); DataFusion cannot push a limit below a Filter/Sort, so the scan receives `limit=None` and only **10 candidates** ever leave Elasticsearch.

Concrete: ES-backed vector index, `/v1/search` with `limit: 50` (the pipeline always sorts → scan limit None), or `vector_search(tbl, 'q', 200)` with any pushable WHERE — results 11..N are silently missing.

Compare: the S3 vectors path defaults to `S3_VECTOR_MAX_TOPK` and DuckDB to 1000. Suggested fix: plumb the requested top-K into `ElasticsearchKnnTable` (like the other index backends) rather than relying on scan-level limit pushdown, or default `k` to a much larger candidate pool.

claudespice:
## Summary

The Elasticsearch vector index hardcoded the kNN candidate pool to `k = 10` in `ElasticsearchIndex::query_table_provider` ([`crates/search/src/index/elasticsearch/mod.rs`](crates/search/src/index/elasticsearch/mod.rs)). The search pipeline places `Filter`/`Sort` above the index scan, so DataFusion cannot push the requested top-K below them and `ElasticsearchKnnTable::scan` receives `limit = None`, falling back to that `k`. As a result, any Elasticsearch-backed vector search whose limit could not be pushed down — `/v1/search` always sorts, and `vector_search(tbl, 'q', 200)` with any pushable `WHERE` — silently returned only the first 10 hits; results 11..N were missing.

The sibling backends do not have this problem: DuckDB falls back to `DEFAULT_DUCKDB_VECTOR_SEARCH_LIMIT` (1000) and the S3 vectors path to a 10,000 max top-K. Only the Elasticsearch kNN path used 10.

## Changes

- Add `DEFAULT_ELASTICSEARCH_VECTOR_SEARCH_LIMIT = 1000` and use it as the kNN candidate-pool fallback in `query_table_provider`, matching the DuckDB vector backend's default. When a `LIMIT` *is* pushed down, the scan still uses the requested value (`limit.unwrap_or(self.k)`) unchanged.
- Kept at 1000 (not higher) because the kNN exec requests `num_candidates = k * 2` and Elasticsearch rejects `num_candidates` above 10,000; a compile-time assertion enforces both bounds (`> 10` and `* 2 <= 10_000`).
- Add a regression test that drives the real `query_table_provider` plan and asserts the kNN exec built for an un-pushed-down scan (`limit = None`) requests the large candidate pool rather than the old hardcoded 10.

## Performance impact

When limit pushdown is blocked by a `Filter`/`Sort`, Elasticsearch now returns up to 1000 candidates (`num_candidates = 2000`) instead of 10 before the pipeline applies the user's filter/sort/limit. This is the same candidate-pool size the DuckDB vector backend already uses, is required for correctness, and stays well within Elasticsearch's `num_candidates` ceiling. Queries whose limit *is* pushed down are unaffected.

## Test plan

- `make lint`
- `make test`

## Checklist

- [x] Root cause identified and fixed at the source (the hardcoded `k`), consistent with the sibling DuckDB backend
- [x] Regression test added that exercises the production `query_table_provider` path and fails on the old `k = 10`
- [x] No existing test or snapshot asserted the old `k = 10`
- [x] Change scoped to the Elasticsearch kNN path; the text-search path already used a large default and is untouched

Fixes #11264


claudespice:
@copilot review

lukekim (dismissed):

Jeadie (approved):

PR #11471, opened 2026-06-26
Handle mongo connection string as a secret in spidapter

krinart:
## 📝 Summary

When creating new app in SCP via spicebench - create `MONGO_CONNECTION_STRING` secret in order not to inline connections string directly in spicepod yaml


Jeadie (dismissed):

tools/spidapter/src/stdio_server.rs
```diff
@@ -1963,6 +1964,10 @@ async fn generate_initial_spicepod(
     _args: &StdioArgs,
     scp: &ScpConfig,
     cayenne: Option<&CayenneConfig>,
+    // When true (Spice Cloud deployments), source connection strings that
+    // carry credentials are emitted as `${secrets:...}` references rather than
+    // inlined, and the caller sets the corresponding secret on the app.
+    use_connection_secrets: bool,
```
phillipleblanc: I don't understand why this needs to be a parameter? Shouldn't it always default to true

tools/spidapter/src/sources/mongodb.rs
```diff
@@ -25,11 +25,17 @@ use uuid::Uuid;
...
+/// Cloud app and referenced from the spicepod as
+/// `${secrets:MONGO_CONNECTION_STRING}`.
+pub(crate) const MONGO_CONNECTION_STRING_SECRET: &str = "MONGO_CONNECTION_STRING";
+
 pub(crate) fn generate_mongodb_spicepod(
     run_id: &Uuid,
     uri: &str,
     datasets: &HashMap<String, DatasetConfig>,
     acceleration_engine: &str,
+    use_connection_secret: bool,
```
phillipleblanc: Prefer using enums over booleans

tools/spidapter/src/stdio_server.rs
krinart: Makes sense. My initial thought was to have separate ways for local vs scp - but I simplified to always use secrets.

phillipleblanc (approved):

PR #11476, opened 2026-06-26
fix(cluster): restore shuffle_location/temp_directory injection lost from #11454

phillipleblanc:
## Problem

Cluster-bench executors are evicted mid-query with `Usage of EmptyDir volume "tmp" exceeds the limit "1Gi"`. They write Ballista shuffle output to the default temp dir (the 1Gi `tmp` EmptyDir) instead of the `/data` PVC, exceed it, and get evicted — which surfaces as inter-executor shuffle-fetch DNS (`Name or service not known`) and `Connection reset` failures as the pods are recreated, livelocking the run.

## Root cause

The spidapter + runtime parts of #11454 were lost in a later merge. The cluster-bench workflow exports `SPIDAPTER_SHUFFLE_LOCATION`/`SPIDAPTER_QUERY_TEMP_DIRECTORY` from the test config, but `tools/spidapter/stdio_server.rs` no longer consumes them (only `SCHEDULER_STATE_LOCATION` survives), so the generated **scheduler** spicepod has no `runtime.params.shuffle_location`. Executors take their config from the scheduler, so they fall back to the default temp dir.

## Fix

- **spidapter (`stdio_server.rs`)**: inject `SPIDAPTER_SHUFFLE_LOCATION` → `runtime.params.shuffle_location` and `SPIDAPTER_QUERY_TEMP_DIRECTORY` → `runtime.query.temp_directory` into the generated scheduler spicepod (mirrors the existing `SCHEDULER_STATE_LOCATION` handling).
- **runtime (`cluster/mod.rs`)**: `create_dir_all` the configured shuffle work_dir instead of only warning, so a per-run `/data` PVC subdir is usable.

Diagnosed against a live SF10 cluster-bench run (executors evicted; scheduler spicepod missing `shuffle_location`).

https://claude.ai/code/session_01G8drJk8Fx5ynRQZ4WkuVju

### DIFF
--- a/crates/runtime/src/cluster/mod.rs
+++ b/crates/runtime/src/cluster/mod.rs
@@ -1237,14 +1237,19 @@ pub async fn initialize_cluster_executor(
                 .unwrap_or_else(|| env::temp_dir().to_string_lossy().to_string())
         }
         Some(loc) => {
-            // Local disk mode with explicit path
-            // Validate the path exists or can be created
+            // Local disk mode with explicit path — ensure the work dir exists.
+            // Cluster-bench points this at a per-run subdir of the /data PVC,
+            // which won't exist until created; create it so the executor doesn't
+            // fall back to (and fill) the default temp dir.
             let path = std::path::Path::new(loc);
             if !path.exists() {
-                tracing::warn!(
-                    "shuffle_location '{}' does not exist. Ensure the directory exists and is writable by the executor process.",
-                    loc
-                );
+                if let Err(e) = std::fs::create_dir_all(path) {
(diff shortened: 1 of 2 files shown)


phillipleblanc:
Closing — superseded. My local trunk was stale; latest trunk (adab61b9) already contains the full #11454 (both the stdio_server SPIDAPTER_SHUFFLE_LOCATION injection at stdio_server.rs:2077 and the mod.rs create_dir_all). This PR redundantly re-added both. The real change (the 4 distributed-query-correctness fixes) will be rebased onto latest trunk separately.

PR #11479, opened 2026-06-26
fix(search): persist character chunk offsets so search snippets aren't shifted/garbled (fixes #11269)

### LINKED ISSUE #11269: Chunked search _match text: 0-based byte offsets fed to 1-based character substring() — shifted/garbled snippets on non-ASCII

Write side stores chunk offsets as 0-based **byte** offsets: `crates/runtime/src/embeddings/execution_plan.rs:888-898` stores `(start, start + chunk.len())` where `start` is a byte index from `chunker.chunk_indices` (confirmed by `crates/chunking/src/lib.rs:352-362`, which slices `text[idx..idx+chunk.len()]`); same for `crates/search/src/index/chunking.rs:442-446`.

Read side extracts the match text with `substring(src, offset[1], offset[2]-offset[1])` (`crates/runtime/src/search/candidate/vector.rs:492-500`, `crates/search/src/provider.rs:407-415`). DataFusion `substr` uses PostgreSQL semantics: **1-based, character-counted**. So every extracted `_match` is shifted one character left even for ASCII (first chunk loses its last char; `substr(s, 0, 7)` returns 6 chars), and for any non-ASCII source text the byte/char divergence makes the snippet arbitrarily wrong. Scores/ranking are unaffected — the returned matched text is what's wrong.

Suggested fix: store char-based offsets at write time (compute via `char_indices` when chunking) and use `offset+1` in the substring, or extract by byte range with a byte-aware function. Note existing stored offsets are byte-based, so the read side and write side must change together (acceleration data rebuilds on refresh).

claudespice:
## Summary (root cause)

Chunked search stored **0-based byte** offsets per chunk on the write side (`(start, start + chunk.len())`), but the read side extracts the matched snippet with DataFusion `substring`, which is **1-based and character-counted** (PostgreSQL semantics). Two bugs resulted:

1. **Off-by-one for all text** — a 0-based byte start used as a 1-based position shifts every snippet one character left; the first chunk loses its last character (`substr(s, 0, 7)` returns 6 chars).
2. **Byte/character divergence for non-ASCII** — once a multi-byte character precedes a chunk, the stored byte offset no longer matches the character position `substring` counts, garbling the snippet arbitrarily.

Scores and ranking were unaffected — only the returned `_match` / `_value` text was wrong.

## Changes

- `crates/chunking/src/lib.rs`: add `Chunker::chunk_with_char_offsets`, returning 0-based, end-exclusive **character** offsets. Computed in O(n) via a forward char cursor over `chunk_indices` (handles overlapping chunks; degrades gracefully if a chunker ever emits out-of-order starts).
- Write sites switched to character offsets: `crates/search/src/index/chunking.rs` (feeds `to_offset_array`) and `crates/runtime-search/src/embeddings/execution_plan.rs` (drops the now-redundant byte-length end recomputation).
- Read sites compute the substring start as `offset[1] + 1` (length `offset[2] - offset[1]` is already the character count): `SearchTableProvider::add_match_column` (`crates/search/src/provider.rs`) and the candidate vector value extraction (`crates/runtime-search/src/candidate/vector.rs`).

Write and read change together by necessity; acceleration data is rebuilt on refresh, so persisted byte-offset data is replaced with character offsets.

## Performance impact

`chunk_with_char_offsets` adds one forward character scan over the source string per document (O(n) total, not O(n times chunks)), on the indexing/embedding write path only. Read-path cost is unchanged (one extra integer add in the projection expression).

## Test plan
- [x] Added regression test for non-ASCII chunk recovery through the production write + read path (`substring_read_path_recovers_chunks_including_unicode` in `crates/search/src/index/chunking.rs`): builds offsets via the production helpers, runs the exact read-side `substring` extraction over a `MemTable`, and asserts each extracted snippet equals the original chunk.
- [x] Added edge case tests for ASCII (char offsets equal byte offsets) and multi-byte text (char offsets must diverge from byte offsets) in `crates/chunking/src/lib.rs`.
- [x] `make lint` passes (workspace-wide fmt check plus full clippy)
- [x] `make test` passes

## Checklist
- [x] Root cause documented in the commit message and PR body
- [x] Fix scoped to the search chunk-offset read/write path
- [x] No snapshot/workflow drift staged

Fixes #11269

### DIFF
--- a/crates/chunking/src/lib.rs
+++ b/crates/chunking/src/lib.rs
@@ -45,7 +45,7 @@ type ChunkIter<'a> = Box<dyn Iterator<Item = &'a str> + 'a>;
 pub trait Chunker: Sync + Send {
     fn chunk_indices<'a>(&self, text: &'a str) -> ChunkIndicesIter<'a>;
 
-    /// Chunks a given `text`, and for each returning the starting (inclusive) and ending (exclusive) indexes into the input `text`.
+    /// Chunks a given `text`, and for each returning the starting (inclusive) and ending (exclusive) **byte** indexes into the input `text`.
     fn chunk_with_offsets<'a>(
         &self,
         text: &'a str,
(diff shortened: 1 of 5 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11490, opened 2026-06-28
fix(cayenne): LIMIT N under-delivers on key-deletion tables

lukekim:
## Summary

Fixes a wrong-results bug: `SELECT … LIMIT N` against a primary-key table with pending **key-based** deletions (`Int64Pk` / `RowConverterBased`) returns fewer than `N` rows when ≥ `N` live rows exist — **0 of 120 live rows** in the regression repro.

Two raw-count limit levers feed off pre-deletion footer counts while the key-based deletion filter runs *above* the scan:
- `collect_scan_files_with_limit` truncates the scanned file set by raw `num_rows`;
- `with_limit(N)` pushed into the Vortex scan caps it at `N` **raw** rows, which the deletion filter then drains to fewer (or zero) live rows.

Position-based deletion is exempt — its deletes apply *inside* the Vortex access plan, so its pushed limit already counts live rows.

## Fix (3 sites in `provider/table.rs`)

1. Disable the file-set truncation for any pending deletion (`collect_stats = … && !self.has_pending_deletions()`) — generalizes the existing position-only guard.
2. Suppress the Vortex `with_limit` push for key-based pending deletion; the outer `GlobalLimitExec` in `scan()` still enforces `LIMIT` over live rows.
3. Broaden `statistics()` so the optimizer never receives raw footer over-counts for key-based pending deletions.

## Test

`tests/limit_underdelivery_key_deletion_regression_test.rs` — **RED on trunk** (`LIMIT 100` returns 0 of 120 live rows), **GREEN** after. `inline_max_rows: 0` forces the file-scan path; the 120-live-row precondition is asserted before the LIMIT assertion so a pass can't be a false negative.

## Context

Found during a Cayenne correctness audit alongside two snapshot-publish **torn-read** bugs (overwrite and sort-rewrite cleared deletion caches outside the `listing_fence`). Both are **already fixed on trunk** — `publish_overwrite_snapshot` and the fenced clear in `sort_and_rewrite_data` — so this PR is scoped to the one remaining bug. The key-deletion `LIMIT` path was the only one still affected.


crates/cayenne/src/provider/table.rs
```diff
@@ -17605,6 +17605,21 @@ impl CayenneTableProvider {
             });
         let statistic_file_limit = if data_filters.is_empty() { limit } else { None };
 
+        // Key-based deletion (`Int64Pk` / `RowConverterBased`) applies its filter
+        // ABOVE the Vortex scan (`apply_deletion_filter` in `scan()`), so a raw
+        // row-count limit pushed INTO the scan would count rows the filter then
```
bjchambers, line 17610: Maybe add a note (or we can open an issue to revisit). The Vortex Scan API allows pushing the predicate into the scan, after which I believe the limit could still be applied (since the scan would know which rows ar efiltered out).

https://docs.vortex.dev/concepts/scanning#filter-pushdown

bjchambers (dismissed):

bjchambers (approved):

PR #11505, opened 2026-06-29
fix(queries): run async /v1/queries jobs under the submitting request context

phillipleblanc:
## Problem

`JobExecutor::submit` spawned the background query job with a bare `tokio::spawn`. The request context is a task-local, so the spawn dropped it and the job planned and executed under the internal context (`Protocol::Internal`). For async queries submitted to `/v1/queries`, this meant:

- Results were cached in the shared `System` namespace instead of the caller's principal namespace — async results were not isolated per principal, and were never shared with the equivalent `/v1/sql` query.
- The request's `Cache-Control` header (and any client-supplied cache key) had no effect.
- Principal-scoped authorization and row/column masking were bypassed during planning, because the enforcer saw no principal.

## Fix

Capture the submitting request's context in `submit` and re-establish it across the spawn with `RequestContext::scope`, so the background job runs as the submitting principal.

## Test

Adds a regression test asserting that a job spawned by the executor observes the submitter's cache namespace rather than the internal (`System`) one.

### DIFF
--- a/crates/runtime/src/jobs/executor.rs
+++ b/crates/runtime/src/jobs/executor.rs
@@ -24,6 +24,8 @@ use tokio::sync::RwLock;
 use tokio_util::sync::CancellationToken;
 use tracing::Instrument;
 
+use runtime_request_context::{AsyncMarker, RequestContext};
+
 use crate::datafusion::DataFusion;
 use crate::datafusion::query::{QueryBuilder, QueryHandle, QueryHandleError};
 use crate::http::v1::queries::SubmitQueryRequest;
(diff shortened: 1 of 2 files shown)


crates/runtime/src/jobs/executor.rs
```diff
@@ -73,6 +75,17 @@ impl JobExecutor {
         }
     }
 
+    /// Spawns `fut` bound to the current request context so the background job
+    /// runs as the submitting principal — results-cache namespacing and
+    /// principal-scoped authz/masking key off it.
```
Jeadie: nit: Should this be in `runtime_request_context`?

lukekim (approved):

PR #11507, opened 2026-06-29
fix: Flight DoPut silently dropped client batches on early sink completion (fixes #11270)

### LINKED ISSUE #11270: Flight DoPut: write future completing before the stream ends discards remaining client batches and reports success

`crates/runtime/src/flight/do_put.rs:341-355` and `421-434`: when `write_future` resolves `Ok(())` before the inbound flight stream ends ("Write operation completed before stream ended"), the handler drains and **discards** all remaining incoming `FlightData` messages and yields `Ok(PutResult::default())`. Rows the client streamed after the sink stopped consuming are never written, and the client receives an unconditional success — silent partial ingest. The code comment explicitly anticipates sinks that "do not consume the input stream or finish early".

Suggested fix: if the write future completes while the client is still sending data batches (not just stream trailers), surface an error (or at minimum count the discarded batches and fail when nonzero) instead of acking success.

Minor, same file: on finalization failure (`do_put.rs:471-477`) the `yield Err(...)` is not followed by `break`, so the generator falls through and also logs the success line (tonic terminates the RPC on the first Err, so clients shouldn't see the spurious success — log noise only).

claudespice:
## Summary

`DoPut` (`crates/runtime/src/flight/do_put.rs`) silently lost client rows on early sink completion. When the write future resolved `Ok(())` before the inbound flight stream ended, the handler **drained and discarded** all remaining incoming `FlightData` and unconditionally yielded `Ok(PutResult::default())` — so any batches the client streamed after the sink stopped consuming were never written, yet the client received success. This is silent partial ingest / data loss.

Root cause: the two early-completion arms treated "sink finished before the stream ended" as a normal success and threw away whatever was still on the wire, regardless of whether those messages carried data.

## Changes

- Add `drain_discarded_data_batches`, which drains the remaining stream and **counts** the messages that carry record-batch data (non-empty `data_body`, excluding keepalive heartbeats and empty trailers).
- Both early-completion arms now return `Status::data_loss` when any client data batch was discarded, instead of acking success. A sink that completes exactly as the stream ends discards nothing, so legitimate early completion still succeeds.
- The nested arm (sink finished while a batch was still pending) counts that in-flight, never-accepted batch as discarded too.
- Fix the secondary bug at the finalization-failure path: the `yield Err(...)` was missing a `break`, so the generator fell through and also logged success / yielded a spurious `Ok` after the error.

## Test plan

- `make lint`
- `make test`
- New unit tests for `drain_discarded_data_batches`: counts only data-bearing messages (keepalives, empty trailers, and transient read errors ignored); empty stream reports zero; keepalive/trailer-only stream reports zero.

## Checklist

- [x] New unit tests cover the data-loss detection and the no-false-positive cases
- [x] No change to the success path when the stream ends cleanly
- [x] Scoped clippy clean on the changed file (pre-existing unrelated warnings only)

Fixes #11270

### DIFF
--- a/crates/runtime/src/flight/do_put.rs
+++ b/crates/runtime/src/flight/do_put.rs
@@ -286,6 +286,39 @@ fn allow_scheduler_trusted_executor_write(datafusion: &DataFusion) -> bool {
         && datafusion.cluster_config.tls_config().is_some()
 }
 
+/// Drains the messages remaining on the inbound flight stream after the write
+/// sink has already completed, returning the number of discarded messages that
+/// carried record-batch data — i.e. client rows that were streamed but never
+/// written. Keepalive heartbeats and empty trailer messages do not carry data
+/// and are not counted, so a sink that completes exactly as the stream ends
+/// reports zero discarded batches and the write is still acked as a success.
+async fn drain_discarded_data_batches<S>(stream: &mut S) -> usize
+where
+    S: futures::Stream<Item = Result<FlightData, Status>> + Unpin,
+{
+    let mut discarded = 0usize;
+    while let Some(msg) = stream.next().await {
+        match msg {
+            Ok(data) => {
+                // A message carries record-batch (or dictionary) data when its
+                // body is non-empty; keepalive heartbeats are tagged and never
+                // count as lost client data.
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11509, opened 2026-06-29
fix(flightsql): propagate bearer token to per-endpoint clients

boittega:
## Problem

When `GetFlightInfo` returns endpoint locations, Spice creates a new `FlightSqlServiceClient` for each `DoGet` call via `get_client_for_flight_endpoint`. This fresh client never performs a handshake, so it has no bearer token.

Servers that validate authentication on every connection, including per-endpoint `DoGet` calls, reject the request with:

```
Execution error: Query execution failed (flightsql). Tonic error: code: 'The request does not have valid authentication credentials'
```

This was observed with StarRocks Flight SQL. Dataset *registration* succeeds because schema discovery uses the original authenticated client, but *query execution* fails after Spice follows the endpoint location to a new unauthenticated client.

This bug is distinct from #11286, which fixed schema discovery. After that fix, registration works but queries still fail due to this auth-propagation gap.

## Root Cause

In `get_client_for_flight_endpoint`, when `ep.location` is non-empty and reachable, Spice creates a fresh `FlightSqlServiceClient::new(channel)`. The original client's bearer token, set by `handshake()`, is not copied to the new client.

## Fix

Carry the bearer token from the authenticated Flight SQL client to the per-endpoint client used for `DoGet`:

1. **`connector-flightsql/src/lib.rs`**: after handshake, extract the token with `client.token().cloned()` and pass it into `FlightSQLFactory` via `with_token()`.
2. **`flightsql.rs`**: store the token on `FlightSQLFactory`, `FlightSQLTable`, and `FlightSqlExec`, and apply it to the execution `FlightSqlServiceClient` before calling `query_to_stream`.
3. **`query_to_stream`**: keep the existing function signature and call `get_client_for_flight_endpoint` with the authenticated client.
4. **`get_client_for_flight_endpoint`**: read `client.token()` and call `new_client.set_token(t)` before returning the freshly-created per-endpoint client.

No additional network round-trip is required. The code uses the existing `FlightSqlServiceClient::token()` and `set_token()` API.

The token-bearing types use custom `Debug` implementations that redact the token.

## Testing

Added a unit test that verifies a token set on the main Flight SQL client is propagated to the per-endpoint `DoGet` client created from `FlightEndpoint.location`.

Tested against a live StarRocks cluster using the trunk image (`spiceai-dev:trunk-8f972a3`, which includes #11286):

**Before patch:**
```
Dataset starrocks_example registered (flightsql:...) -- OK
Query: SELECT COUNT(*) FROM starrocks_example
Error: The request does not have valid authentication credentials
```

**After patch:**
```
Dataset starrocks_example registered (flightsql:...) -- OK
Query: SELECT COUNT(*) FROM starrocks_example -> 954,195,761 rows
Query: SELECT cluster_id, entity_name FROM starrocks_example LIMIT 3 -> real data
```

Direct ADBC queries with the same credentials succeed before and after as a control case, confirming the bug is in Spice's per-endpoint client creation rather than StarRocks auth configuration.

## Relation to #11286

PR #11286 fixed schema fallback for servers that don't implement `GetTables`. This PR fixes the auth-propagation gap that prevents *query execution* from succeeding after schema registration works.


crates/data_components/src/flightsql.rs
```diff
@@ -205,6 +215,7 @@ impl FlightSQLTable {
         client: FlightSqlClient,
         table_reference: impl Into<TableReference>,
         cookie_store: Arc<CookieStore>,
+        token: Option<String>,
```
phillipleblanc: Instead of piping this through as an optional parameter, it should follow the pattern of statistics/function support and omit it from this constructor and set via `with_token`.

crates/data_components/src/flightsql.rs
```diff
@@ -962,16 +994,27 @@ pub fn query_to_stream(
     mut client: FlightSqlClient,
```
phillipleblanc, line 1014: Instead of passing the token in with the flightSqlClient, the caller should be responsible for setting the token before calling this method - that way this function doesn't need to change.

crates/data_components/src/flightsql.rs
```diff
@@ -1009,7 +1053,13 @@ pub async fn get_client_for_flight_endpoint(
         match new_tls_flight_channel(&ep.location[0].uri, None).await {
             Ok(channel) => {
                 let channel = CookieService::new(channel, Arc::clone(cookie_store));
-                Ok(FlightSqlServiceClient::new(channel))
+                let mut new_client = FlightSqlServiceClient::new(channel);
```
phillipleblanc, line 1069: The FlightSqlServiceClient already has a method to get any token that is currently set, which removes the need to pass it in explicitly.

phillipleblanc (commented):
Thanks for this PR! I've left a few comments, please take a look. Can you also please address any CI failures and Copilot review comments as well?

crates/data_components/src/flightsql.rs
```diff
@@ -962,16 +994,27 @@ pub fn query_to_stream(
...
         for ep in flight_info.endpoint {
             if let Some(tkt) = ep.clone().ticket {
-                match get_client_for_flight_endpoint(&client, ep, &cookie_store).await
+                match get_client_for_flight_endpoint(
+                    &client,
+                    ep,
+                    &cookie_store,
+                    client.token().map(|s| s.as_str()),
+                )
+                .await
```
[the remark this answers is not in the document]
boittega: Addressed in code by adding query_to_stream_propagates_token_to_endpoint_client, which verifies bearer-token forwarding to the per-endpoint DoGet client.

crates/data_components/src/flightsql.rs
```diff
@@ -135,6 +135,7 @@ pub struct FlightSQLFactory {
     endpoint: String,
     cookie_store: Arc<CookieStore>,
     function_support: Option<FunctionSupport>,
+    token: Option<String>,
```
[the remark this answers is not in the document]
boittega: Addressed in code by removing derived Debug and adding a custom Debug implementation that redacts the token.

crates/data_components/src/flightsql.rs
```diff
@@ -195,6 +204,7 @@ pub struct FlightSQLTable {
     /// as `json_get_str`) are evaluated locally instead of pushed into the SQL
     /// sent to the Flight SQL server. See issue #10703.
     function_support: Option<FunctionSupport>,
+    token: Option<String>,
```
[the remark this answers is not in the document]
boittega: Addressed in code by adding a custom Debug implementation that redacts the token.

crates/data_components/src/flightsql.rs
boittega: Addressed in code. create no longer takes the token, and token is applied with with_token.

crates/data_components/src/flightsql.rs
boittega: Addressed in code. query_to_stream is back to the original signature, and callers set the token on the client before calling.

crates/data_components/src/flightsql.rs
boittega: Addressed in code. It now uses client.token().cloned() and new_client.set_token(t).

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #11515, opened 2026-06-29
fix(cayenne): seed persisted `num_rows` for hash-join sizing

sgrebnov:
## 📝 Summary

During CDC, Cayenne scans reported `num_rows: Absent`, so `JoinSelection` couldn't size hash joins and built the larger input. Two causes, fixed together:

1. All durability modes — the base+delta `UnionExec` collapses `num_rows` to `Absent` on a stat-less branch (e.g. `collect_stats=false` during pending deletions), and the stats overlay didn't restore it. Hit even file-durability tables whose persisted count was already correct.
2. `cdc_durability: memory` only — the RAM-append path (`append_to_shard`) never fed the persisted aggregate, so the source count sat at `0`.

## Fix

1. Overlay restores `num_rows` (general). `optimizer_stats_overlay_for_schema` now carries the maintained whole-table count, and `restore_absent_column_statistics` backfills it onto an `Absent` child. A present, filter-aware child count still wins.
2. Checkpoint seeds the count (memory). The mem-tier checkpoint persists `Delta(flushed_mem_rows)` instead of `Unchanged` 
3. Guard with `num_rows > 0`. A non-positive overlay count (cold aggregate, or before the first checkpoint seeds a new memory table) is treated as "no info" and not restored - otherwise the NDV cap would zero every refilled `distinct_count`, the regression that made build/probe worse.

## 👀 Notes for Reviewers

The overlay surfaces a single whole-table count *above* the per-branch filters, so a scan-level filter on a memory table during a pending-deletion window would restore the whole-table count and inflate the filtered probe side. The better mechanism is to populate each base-branch `num_rows` from the manifest's per-file `row_count` and let the `UnionExec` sum the branches: per-scan, filter-aware, covers protected-snapshot sub-scans, and drops the overlay dependency for `num_rows`. This can be improved as separate PR.


krinart (dismissed):

lukekim (dismissed):

crates/cayenne/src/provider/scan.rs
```diff
@@ -310,12 +310,12 @@ impl CayenneAccelerationExec {
     }
 }
 
-/// Refills only the `Precision::Absent` `distinct_count` (NDV) in `child` from
-/// `overlay`, restoring the join-key NDV that the Cayenne base+delta `UnionExec`
-/// drops (a stat-less/empty delta branch makes `col_stats_union` return `Absent`,
-/// collapsing `estimate_inner_join_cardinality` to `min(L, R)`). Present child
-/// stats — filter-aware `num_rows`/`null_count` and any surviving column stat —
-/// are preserved; a column-count mismatch is a defensive no-op.
+/// Refills the `Precision::Absent` `num_rows` and per-column `distinct_count`
```
bjchambers, line 313: As a question for myself -- do we also restore the HLL sketch used to estimate NDV or do we do that elsewhere? Or do we store per-file HLL sketches and then merge them to get the final NDV?

bjchambers (approved):

PR #11532, opened 2026-06-30
fix: arrow IndexedMemTable serves stale results after DML/retention/sync (fixes #11262)

### LINKED ISSUE #11262: Arrow accelerator hash index serves stale results: Exact PK pushdown but index never rebuilt on DML, retention deletes, or synchronized-table refresh

## Problem

`IndexedMemTable` (auto-enabled for `engine: arrow` datasets with `primary_key`/`indexes`) advertises **Exact** filter pushdown for PK equality (`crates/data_components/src/arrow/indexed.rs:605-617`) and returns an **empty** stream on a lookup miss (`indexed.rs:693-707`). Exact means DataFusion deletes the predicate from the plan and trusts the scan — so a stale index is not "stale", it is silently wrong results.

The index is only rebuilt by `TableSink::insert_into` (`crates/runtime/src/accelerated_table/sink/table.rs:155-171`) and the CDC path. It is **not** maintained on:

1. **Direct DML through `AcceleratedTable`** (`WriteMode::AcceleratorOnly`, the standard arrow+PK+`on_conflict` config): `IndexedMemTable::insert_into` has an explicit TODO (`indexed.rs:782-797`, "we would need to rebuild after insert completes") and never rebuilds; `delete_from`/`update`/`truncate` delegate with no index maintenance at all (`indexed.rs:803-850`).
2. **Retention eviction**: `crates/runtime/src/accelerated_table/retention.rs:67` calls `accelerator.delete_from(...)` directly; MemTable deletes shift row locations, invalidating every stored `RowLocation`.
3. **MultiSink (synchronized child tables)**: `crates/runtime/src/accelerated_table/sink/multi.rs` never calls `perform_index_maintenance`, so once `add_synchronized_table` converts the sink, every refresh leaves the parent index permanently stale.
4. Index-maintenance failure after refresh is warn-only (`sink/table.rs:164-169`) — with Exact pushdown that "may be stale" warning means silently missing rows until the next successful rebuild.

## Concrete wrong result

Dataset `engine: arrow`, `primary_key: id`, `on_conflict: upsert`:

```sql
INSERT INTO ds VALUES (42, ...);     -- accepted, lands in MemTable
SELECT * FROM ds WHERE id = 42;      -- index miss -> 0 rows (Exact: filter not re-applied)
SELECT * FROM ds;                    -- shows the row
```

Same shape for surviving keys after a retention delete (stale `RowLocation` fails key verification → empty result for an existing row).

## Suggested fix direction

Minimal-correctness option: a "dirty" flag on `IndexedMemTable`, set by `insert_into`/`delete_from`/`update`/`truncate` (at plan time — conservative), cleared by `perform_index_maintenance`. While dirty, `supports_filters_pushdown` returns Inexact and `scan` bypasses the index (full MemTable scan). Also call `perform_index_maintenance` from `MultiSink` like `TableSink` does, and decide whether a failed rebuild should keep the table in the dirty/fallback state instead of warn-and-serve-stale.

Related test evidence: the unit test at `crates/runtime/src/dataaccelerator/mod.rs:1139` must call `perform_index_maintenance` manually after a direct insert — confirming the gap.

claudespice:
## Summary

`IndexedMemTable` (auto-enabled for `engine: arrow` datasets with a `primary_key`/`indexes`) advertises **Exact** filter pushdown for PK equality, so DataFusion drops the predicate from the plan and trusts the scan. The hash index, however, was only rebuilt by `TableSink::insert_into` and the CDC path — never on:

1. **Direct DML through `AcceleratedTable`** (`WriteMode::AcceleratorOnly`): `insert_into` had a `TODO` and never rebuilt; `delete_from`/`update`/`truncate` delegated with no index maintenance.
2. **Retention eviction** (`retention.rs` calls `accelerator.delete_from`), which shifts row locations and invalidates every stored `RowLocation`.
3. **`MultiSink`** (synchronized child tables), which never called `perform_index_maintenance`.

With Exact pushdown, a stale index isn't merely slow — it returns **silently wrong results**: an inserted row is reported as a point-lookup miss (0 rows) even though `SELECT *` shows it, and surviving keys after a retention delete fail key verification and return empty.

## Root cause

The index could drift out of sync with the underlying `MemTable` while `supports_filters_pushdown` kept returning `Exact`, removing the predicate and trusting the index unconditionally.

## Changes

- **`IndexedMemTable` dirty flag** (`crates/data_components/src/arrow/indexed.rs`): an `AtomicBool` set (conservatively, at plan time) by `insert_into`/`delete_from`/`update`/`truncate`, and cleared only after a successful rebuild in `perform_maintenance`. While dirty:
  - `supports_filters_pushdown` downgrades the PK equality from `Exact` to `Inexact`, so DataFusion keeps a `FilterExec` and re-applies the predicate;
  - `scan` bypasses the PK **and** secondary index paths and serves a full `MemTable` scan.
  - A **failed** rebuild leaves the flag dirty, so reads keep falling back to a full scan instead of serving stale results (no warn-and-serve-stale).
- **`MultiSink` index maintenance** (`crates/runtime/src/accelerated_table/sink/multi.rs`): after a successful write, call `perform_index_maintenance` on the parent and every synchronized child accelerator, mirroring `TableSink`. This rebuilds the index and clears the dirty flag (no-op for providers without an index).

The clean-path fast lookup is unchanged: a freshly built / freshly rebuilt index still reports `Exact` and uses the indexed point lookup, so existing explain snapshots are unaffected.

## Test plan

- `make lint`
- `make test`

New regression tests in `indexed.rs`:
- `test_insert_then_pk_lookup_finds_new_row` — the concrete #11262 repro: insert a new PK through the provider, then look it up by PK; fails (0 rows) on old code, passes (1 row) now, and the Exact fast path is restored after maintenance.
- `test_dirty_index_downgrades_pushdown_to_inexact` — dirty ⇒ PK pushdown becomes `Inexact`.
- `test_dirty_index_bypassed_in_scan` — dirty plan contains no `IndexedLookupExec` and still returns the correct row.
- `test_dirty_persists_until_successful_maintenance` — flag cleared only by a successful rebuild.

## Notes / limitations

- The dirty flag is set at DML *plan* time (conservative). A DML plan built but never executed leaves the table doing full scans until the next maintenance — correct, just slower.
- Direct DML (case 1) has no post-write maintenance hook, so such tables stay dirty (full-scan fallback) until the next refresh/maintenance rebuilds the index. This is the minimal-correctness behavior; a follow-up could rebuild incrementally on direct DML.

## Performance impact

None on the clean fast path. After a write the affected `IndexedMemTable` serves PK/secondary lookups via a full scan (plus a re-applied `FilterExec`) until the index is rebuilt — correctness over latency for the stale window.

Fixes #11262

### DIFF
--- a/crates/data_components/src/arrow/indexed.rs
+++ b/crates/data_components/src/arrow/indexed.rs
@@ -3389,7 +3389,7 @@ mod tests {
     // =========================================================================
 
     /// A dirty index must downgrade PK equality pushdown from Exact to Inexact so
-    /// DataFusion keeps a FilterExec and re-applies the predicate, instead of
+    /// `DataFusion` keeps a `FilterExec` and re-applies the predicate, instead of
     /// trusting a stale index that could silently drop existing rows.
     #[test]
     fn test_dirty_index_downgrades_pushdown_to_inexact() {
@@ -3422,7 +3422,7 @@ mod tests {
 
     /// A dirty index must not be used in `scan` — the plan must NOT contain an
     /// `IndexedLookupExec`, and the query must still return the correct row via a
-    /// full MemTable scan.
+    /// full `MemTable` scan.
     #[tokio::test]
     async fn test_dirty_index_bypassed_in_scan() {
         let batch = create_large_test_batch(300);
--- a/crates/data_components/src/arrow/indexed.rs
+++ b/crates/data_components/src/arrow/indexed.rs
@@ -29,6 +29,7 @@ limitations under the License.
 use std::collections::HashMap;
 use std::fmt::{self, Debug};
 use std::sync::Arc;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use arrow::array::{Array, RecordBatch};
 use arrow::datatypes::SchemaRef;
@@ -96,6 +97,17 @@ pub struct IndexedMemTable {
     primary_key_columns: Vec<String>,
     /// Secondary indexes on non-primary key columns.
     secondary_indexes: Vec<SecondaryIndex>,
+    /// Whether the hash index(es) may be stale relative to the underlying data.
+    ///
+    /// Set conservatively at plan time by any DML entry point
+    /// (`insert_into`/`delete_from`/`update`/`truncate`) and cleared by
+    /// [`IndexedMemTable::perform_maintenance`] after a successful rebuild. While
+    /// dirty, the index is treated as untrusted: PK/secondary equality filters are
+    /// reported as `Inexact` (so `DataFusion` keeps a `FilterExec`) and `scan`
+    /// bypasses the index for a full `MemTable` scan. This prevents an out-of-date
+    /// index from silently dropping existing rows on an `Exact`-pushdown point
+    /// lookup (see issue #11262).
+    dirty: AtomicBool,
 }
 
 impl Debug for IndexedMemTable {
@@ -105,6 +117,7 @@ impl Debug for IndexedMemTable {
             .field("indexed", &self.index.is_some())
             .field("primary_key_columns", &self.primary_key_columns)
             .field("secondary_indexes_count", &self.secondary_indexes.len())
+            .field("dirty", &self.is_dirty())
             .finish()
     }
 }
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (dismissed):

phillipleblanc (dismissed):

lukekim (approved):

phillipleblanc (approved):

Issue #9388, opened 2026-02-12
Cayenne: clean up orphaned deletion vectors during time-based retention deletes (PK upserts)


sgrebnov:
Nice-to-have optimization — no correctness or data integrity impact.

When file-based retention deletes all data files from a protected snapshot in a PK table with upserts, the associated deletion vector (DV) .arrow files and their cayenne_delete_file catalog rows remain. These orphaned DVs are harmless at query time — sequence-based ordering means stale DVs targeting non-existent PKs are no-ops, and re-inserted PKs with higher sequences are unaffected.

Real impact is primarily during restarts: on startup, Cayenne loads all cayenne_delete_file rows into the in-memory deletion cache via `detect_deletion_type_and_read()`. Orphaned DVs inflate this cache and add unnecessary file I/O during the load phase.

bjchambers:
# Proposed approach for cleaning up orphaned DVs
When file-based retention empties a protected snapshot, the key-based DVs that shadowed that snapshot's rows are left orphaned — their .arrow files live in the base snapshot's deletions/ dir (written there via current_snapshot_id), which retention never removes, so the files and their cayenne_delete_file rows persist.

This is a no-correctness-impact optimization: orphaned DVs are query-time no-ops (sequence ordering makes stale DVs against non-existent PKs harmless). The only cost is in-memory deletion-cache size and load-time I/O on restart.

# Chosen approach — conservative, sequence-number based.
For PK-upsert tables every data-bearing snapshot has a sequence persisted in cayenne_snapshot_sequence, and the deletion model already runs on it: a snapshot with threshold S only has DVs with delete_seq > S applied to it. That gives a clean, safe rule:

> a key-based DV with delete_sequence = D is orphaned ⟺ no surviving snapshot has sequence < D ⟺ D ≤ min(surviving snapshot sequences).

After retention removes emptied snapshots, take min(get_all_snapshot_sequences) and delete any key-based DV at or below that floor (catalog row + .arrow file), reusing the existing cleanup_emptied_snapshots hook + remove_delete_files. It's safe because a DV only ever shadows data older (lower sequence) than itself, so once nothing older survives it shadows nothing — and there's no hidden base data to account for, since all upsert data lives in tracked snapshots. In out-of-order/backfill workloads an old-sequence snapshot can hold the floor down temporarily, but that snapshot eventually ages out under retention too, raising the floor and collecting the DV then — so orphans are always eventually cleaned.

# Considered alternatives:

[...]

PR #11533, opened 2026-06-30
feat(cayenne): enable orphaned-DV cleanup by default (per-table threshold 20)

bjchambers:
## What

Flip the default of `cayenne_orphaned_dv_cleanup_min_files` from **unset (disabled)** to **20** (per-table). Set the param to `0` to disable.

## Background — what this feature does (and doesn't)

The orphaned deletion-vector cleanup (issue #9388) reclaims the `.arrow` files + `cayenne_delete_file` catalog rows that **time-based retention** leaves behind when it empties a protected snapshot in a PK-upsert table. It is:
- **Retention-only** — a DV becomes *orphaned* only once retention (`retention_period` + `time_column`) empties the snapshot(s) it shadowed, raising the surviving-sequence floor above its delete sequence. **Without a retention policy nothing is ever orphaned and the sweep is a no-op.**
- **No correctness impact** — orphaned DVs are query-time no-ops; the benefit is reclaimed disk + a smaller in-memory deletion cache / less file I/O at restart.
- **Off the write path** — lock-free, throttled, runs on the compaction runtime. It never extends the `write_lock` / `listing_fence` window.

This knob bounds only the **orphaned tail**. The **live (not-yet-orphaned)** DV set — DVs still shadowing rows in un-emptied snapshots — is bounded separately by compaction's seq-prefix bake, not by this knob.

## Validation

**1. Reclamation + soundness — retention unit tests (green on this branch):**
`test_orphaned_key_dv_cleaned_after_retention`, `..._all_snapshots_emptied_cleans_all_orphaned_dvs`, `..._below_threshold_retained`, `..._disabled_when_knob_zero`, `test_loader_self_heals_orphaned_missing_dv` — orphan reclaimed after retention, still-needed DV retained, threshold gating, `0` disables, loader self-heals a mid-sweep crash.

**2. No ingest/freshness regression from enabling — CH-benCHmark SF-1000** (`cdc-tuned-memory`, 100 terminals, rate 10000, 600s, `spiceai-dev-xlarge-runners`), knob 0 / 1 / 100:

| arm | tpmC (NewOrder/min) | abort % | QPH | worst-P99 freshness |
|---|---|---|---|---|
| `0` (off) | 264,320 | 0.00% | 817.8 | 308.3s |
| `1` (sweep every orphan) | 263,904 | 0.00% | 818.8 | 282.5s |
| `100` | 264,680 | 0.00% | 831.8 | 287.7s |

All within **0.3% tpmC at 0% abort**. Note: chbench configures **no retention**, so the sweep is inactive across all three arms — this A/B confirms enabling the knob **does not affect the common CDC path**, even at the worst-case `=1` (sweep-every-orphan) setting. (The SF-1000 runs show a red ✗ from an unrelated, pre-existing env flake in the *post-ingest* source-correctness phase that hits `trunk` identically; the ingest metrics above are emitted before it.)

## Sizing the default (per-table)

Each orphaned `.arrow` file's size scales with the number of deletions it recorded, so bounded lingering disk ≈ `n_tables × threshold × avg_dv_size`.

Local SF-100 CH-benCHmark sample (600s), total `delete_*.arrow` files on disk:

| | files (~per table) | total bytes | avg / file |
|---|---|---|---|
| off | 1,118 (~200) | 68.8 MB | 63 KB (max 872 KB) |
| dv20 | 1,050 (~176) | 55.4 MB | 54 KB |

These are ~equal **as expected** — with no retention configured, essentially all of these are **live** DVs, not orphaned, so the sweep correctly leaves them in place. At the default threshold of 20 the bounded *orphaned* tail is ≈ `tables × 20 × avg_dv_size` (≈ ~9 MB at this SF-100 scale for ~7 tables; roughly 10× at SF-1000 as per-file delete counts grow).

## Why enable by default

Retention workloads get automatic disk reclamation + lighter restarts; non-retention workloads are unaffected (no-op, zero ingest cost, proven above). `0` opts out.


bjchambers:
Addressed review feedback + lint in `27c8c3160`:

- **Copilot** (metadata.rs doc inconsistency): the field doc said the param maps "`0` (or unset) to the default", which contradicted the runtime parsing. Reworded to: **`0` → `None` (disabled)**; **unset → default (`20`)**.
- **Lint**: `cargo fmt --check` clean; clippy (cayenne + the runtime cayenne-accelerator module, with the Makefile's pedantic/`-Dwarnings` deny set) clean — fixed one `clippy::doc_markdown` (backticked `avg_dv_size`). The default/param unit test passes (`unset → 20`, `0 → None`).

sgrebnov (approved):

PR #11540, opened 2026-07-01
fix: Placeholder table initialization lock swap

peasee:
## 📝 Summary

* Avoids `.deregister_table` and `.register_table` during placeholder table initialization - `.register_table` already overrides an existing provider, so there is no need to `.deregister_table` beforehand
* Ensures that `.replace_table` is not called if the placeholder is no longer pending
* Obtains a write lock for the table to prevent racing to replace the table across multiple concurrent queries


Cargo.lock
peasee, line 1: not sure why these updated, but this is updated from `trunk` - trunk is outdated

phillipleblanc: these are downgrades, can we revert this change?

Cargo.lock
```diff
@@ -276,7 +276,7 @@ dependencies = [
 
 [[package]]
 name = "ansi_colors"
-version = "2.1.0-unstable"
```
phillipleblanc, line 288: This should still be `-unstable`

Jeadie (dismissed):

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #11554, opened 2026-07-01
Generic CDC replication-lag metric

krinart:
## Description

Adds `dataset_acceleration_cdc_replication_lag_ms` — wall-clock now minus the freshest *applied* upstream commit timestamp per dataset. ~0 = caught up; a growing value = falling behind the CDC source.

`ChangeBatch::source_commit_ts_ms` already existed (stamped by Debezium, Postgres, and MongoDB). This PR turns it into a metric and closes two gaps:

### Idle heartbeats
A caught-up source emits no events, so the raw timestamp freezes and lag would falsely climb. Connectors now emit zero-row heartbeat envelopes while idle, each stamped with a source-attested, liveness-backed timestamp (never a blind `now()`):
- **MongoDB** — session `operation_time` (the change stream now runs under an explicit session).
- **Postgres logical replication** — WAL keepalive server clock (per-dataset and shared-slot paths).
- **DynamoDB** — empty-shard watermark = the poll-cycle start.

### Debezium
Already stamps the source commit time, so it feeds the metric — but has no idle heartbeat (a synthetic timestamp can't be trusted across the Kafka boundary), so its lag can read stale while a topic is idle.

### DIFF
--- a/crates/data-connectors/connector-mongodb/src/changes.rs
+++ b/crates/data-connectors/connector-mongodb/src/changes.rs
@@ -19,7 +19,7 @@ use async_trait::async_trait;
 use data_components::{
     cdc::{
         ChangeEnvelope, ChangesStream, CommitChange, CommitError, NoOpCommitter, StreamError,
-        build_ready_signal_envelope, wrap_data_as_change_batch,
+        build_heartbeat_envelope, build_ready_signal_envelope, wrap_data_as_change_batch,
     },
     mongodb::stream::{
         change_events_to_change_batch, default_unnest_parameters, nullable_clone,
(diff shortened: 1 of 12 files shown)


bjchambers:
One thing to be aware of with these metrics is that it won't play as well with replaying past events (it'll think it's *way* behind). Similarly, it can have issues if the system sending events has a different notion of time (eg., if there clock is behind or incorrectly sends time zone, etc.). I've even seen cases where event timestamps are assigned by different machines with different time zones.

One idea to consider might be to expose both the lag (as you compute it here) AND the `max-applied-timestamp-ms`. In the case where we know the producing system uses the same time, then the lag is informative. But, even in the case of replaying, the `max-applied-timestamp-ms` could be compared to some metric from the producing system (`max-sent-timestamp-ms`) to compute the difference. That metric is also useful to surface a "what time is reflected in this data".

krinart:
@bjchambers I agree that this lag is not 100% accurate - however it can be used as a proxy for how far behind the replication is, as such can be used for alerting.

Also good idea to add metric with a raw timestamp - just added, thanks!

bjchambers:
I've also seen systems that assign an "arrival timestamp" and report the "arrival time -> process time" as an "internal lag" or something. This doesn't reflect the time to arrive, but it is at least something we can control the measurement of. It may be a useful *additional* metric. Then we could say "time from PG write -> applied" is the total e2e lag, "time from arrival -> applied" is the lag within our CDC, etc.

bjchambers (approved):

krinart:
@bjchambers Interesting idea. I will make it as a follow up PR

PR #11561, opened 2026-07-02
Add SQL warehouse ID to integration workflow

Jeadie:
Added `TEST_DATABRICKS_SQL_WAREHOUSE_ID` to environment variables.


.github/workflows/integration.yml
```diff
@@ -365,6 +365,7 @@ jobs:
           TEST_DATABRICKS_HOST: ${{ vars.DATABRICKS_HOST }}
           TEST_DATABRICKS_TOKEN: ${{ secrets.DATABRICKS_TOKEN }}
           TEST_DATABRICKS_CLUSTER_ID: ${{ vars.DATABRICKS_CLUSTER_ID }}
+          TEST_DATABRICKS_SQL_WAREHOUSE_ID: ${{ secrets.TEST_DATABRICKS_SQL_WAREHOUSE_ID }}
```
phillipleblanc: We have a limit on the number of secrets we can use, this should be a variable instead

claudespice: Done in 969542ab7 — switched to `${{ vars.DATABRICKS_SQL_WAREHOUSE_ID }}`, reusing the existing repo variable that the testoperator_run_* workflows already source, so no new secret is consumed. Also matches the adjacent `TEST_DATABRICKS_HOST`/`TEST_DATABRICKS_CLUSTER_ID` `vars.*` usage in this same block.

lukekim (approved):

PR #11564, opened 2026-07-02
feat(cayenne): add integer support to maintained AVG

bjchambers:
## What

Adds integer support to maintained (IVM) `AVG`. Previously, maintained `AVG` accepted only `Float32`/`Float64` columns and rejected everything else with:

```
Avg maintained aggregate does not support column type Int32
```

`AVG` over integers is a normal, common case — Postgres `INTEGER` → arrow `Int32` is the common CDC case (the same note that motivated `SumInt64`) — so this was a genuine feature gap, not a guardrail. This mirrors the integer support `SUM` already has.

## How

A new `AvgInt128` accumulator holds a running **`i128` sum + `i64` count**, dividing down to the `Float64` `AVG` output. `i128` (rather than `f64`) is deliberate:

- **Exactness** — `f64`'s 53-bit mantissa loses precision above 2^53; integers are exactly the case where we can stay exact.
- **Invertibility** — IVM handles `UPDATE` as retract-then-insert. Integer add/sub is exactly invertible; `f64` is not, so an `f64` sum would drift across CDC update churn. (`AvgFloat64` tolerates this only because float input has no exact alternative.)
- **Overflow headroom** — AVG's sum is internal-only (always divided to `Float64`), so unlike `SumInt64` it has no `Int64`-output constraint and can widen. `i128` holds the sum of ~9×10^18 `Int64::MAX`/`UInt64::MAX` values, so averaging values near `i64::MAX` — which would overflow `SumInt64`'s `i64` and force an error — just works, with a `checked_add` still failing safe if the (unreachable) overflow ever happened.

Since `u64` fits in `i128`, a single `AvgInt128` variant covers the entire signed **and** unsigned integer family (`Int8..Int64`, `UInt8..UInt64`) — the same coverage `SUM` has, gated by `is_signed_integer()`/`is_unsigned_integer()` and matched by a `scalar_as_i128` coercer.

Note: DataFusion coerces the query-plan `AVG(int)` into `avg(CAST(col AS Float64))`, but the maintained matcher sees through the cast to the underlying column and resolves its type against the table schema, so integer columns still select `AvgInt128` and read the raw integer values.

## Tests

- `avg_over_int32_maintains_exactly_on_insert_and_retract` — the common `Int32` CDC case across insert / in-place update / delete.
- `avg_over_int64_near_max_does_not_overflow` — averages two `i64::MAX` rows, proving the `i128` headroom `SumInt64` lacks.

All 23 `maintained_aggregate` tests pass; `cargo clippy -p cayenne` is clean.


bjchambers:
Addressed the review feedback on misleading error text:

- Added an `AvgInt128`-specific `avg_overflow()` helper for the insert-path `i128` accumulation overflow, replacing the reused `sum_overflow()` (whose "SUM overflowed…" text misreported the failing aggregate).
- The retract path now reuses the existing generic `retract_underflow()`, matching how `COUNT` / `SUM(UInt64)` already report their retract-path failures.

Also added an optimizer-rule test (`maintained_aggregate_rewriter_replaces_avg_over_int32`) verifying that a real `AVG(Int32)` plan — which DataFusion coerces to `avg(CAST(v AS Float64))` — is matched and rewritten to a `MaintainedAggregateExec`, guarding the cast-see-through + integer type resolution end-to-end.

All 24 `maintained_aggregate`/`optimizer_rules` tests pass; `cargo clippy -p cayenne --tests` is clean.

phillipleblanc (dismissed):

krinart (dismissed):

lukekim (approved):

PR #11570, opened 2026-07-02
fix: surface silent Turso read-path conversion failures (fixes #11276)

### LINKED ISSUE #11276: Turso accelerator read path converts conversion failures to silent NULLs (catch-all _ => None per type arm)

`crates/data_components/src/turso.rs` `values_to_record_batch` (~lines 521-1090) decodes stored rows against the dataset schema with per-type arms shaped like `TursoValue::Integer(i) => i8::try_from(*i).ok(), TursoValue::Null => None, _ => None` — an out-of-range integer, a date string that isn't exactly `%Y-%m-%d`, an unparseable RFC3339 timestamp, or any storage-class mismatch becomes NULL instead of an error. Schema/type drift between write and read (dataset type changed, external mutation of the acceleration DB, or bugs like the timestamptz formatting fix) returns rows with silent NULLs where the source has values — indistinguishable from genuinely-NULL data.

The write path errors loudly on e.g. UInt64 overflow (`turso.rs:2027-2031`), so normal round-trips are safe; the soft-fail is read-side only. Suggested: propagate a structured error (or at minimum count+warn per scan) when a non-NULL stored value fails conversion, instead of `_ => None`.

claudespice:
## Summary (root cause)

`TursoTableProvider::values_to_record_batch` decodes stored rows against the dataset
schema with per-type arms shaped like `TursoValue::Integer(i) => i8::try_from(*i).ok(),
TursoValue::Null => None, _ => None`. Any non-NULL stored value that fails conversion —
an out-of-range integer, an unparseable RFC3339 timestamp, a storage-class mismatch —
soft-fails to `None` and is returned as a **NULL Arrow cell that is indistinguishable
from a genuinely-NULL value**. Under schema/type drift (dataset type changed, external
mutation of the acceleration DB, or a prior formatting bug) a scan silently returns NULLs
where the source has data, with no signal to the operator.

The soft-fail-to-NULL fallback is a deliberate "keep the query available" design choice
(documented on the function), so this PR does **not** change that default. It fixes the
*silent* part: it makes the drift observable.

## Changes

- Add `count_conversion_failures(rows, batch)` — a detection-only pass that counts, per
  column, cells that are NULL in the output batch but whose stored `TursoValue` was **not**
  `Null` (a short row with no value for the column is likewise never counted). It never
  alters the batch.
- `values_to_record_batch` now calls `warn_on_conversion_failures` after building the
  batch, emitting a single per-scan `WARN` with per-column failure counts when any are
  detected. Query availability is unchanged; the drift is no longer silent.
- Updated the read-path doc comment to describe the new observability.

This is scoped to the Turso accelerator read path (the reported bug class). The write path
already errors loudly on overflow, so round-trips remain safe.

## Test plan

- `make lint`
- `make test`

New regression tests (`--features turso`):
- `test_count_conversion_failures_distinguishes_drift_from_genuine_null` — an out-of-range
  `Integer(999)` into an `Int8` column is counted as exactly one conversion failure, while a
  genuine `TursoValue::Null` in the same column is **not** counted; the soft-fail-to-NULL
  output behavior is asserted to be preserved.
- `test_count_conversion_failures_zero_when_all_values_valid` — clean round-trips report
  zero failures.

## Checklist

- [x] Root cause identified and fixed at the read path
- [x] Regression tests added that exercise the actual failure mode
- [x] Existing behavior (soft-fail-to-NULL for availability) preserved — opt-out-free, no default change

Fixes #11276

### DIFF
--- a/crates/data_components/src/turso.rs
+++ b/crates/data_components/src/turso.rs
@@ -509,7 +509,10 @@ impl TursoTableProvider {
     /// to the target type (e.g., INTEGER too large for Int8, invalid JSON) are converted to NULL.
     /// This design choice prioritizes query availability over failing entire result sets due to
     /// individual value conversion issues. This is standard behavior for database queries where
-    /// some data may be malformed or out of range.
+    /// some data may be malformed or out of range. Because a soft-failed value is otherwise
+    /// indistinguishable from genuinely-NULL data, [`Self::count_conversion_failures`] detects
+    /// these cases after the batch is built and emits a per-scan WARN so schema/type drift is
+    /// observable rather than silent.
     ///
     /// **Write-time validation is critical**: To prevent bad data from entering the database,
     /// see `scalar_value_to_turso()` which enforces strict validation during INSERT operations.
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11573, opened 2026-07-02
test(chbench): add cold-object-store-tier SF-1000 HTAP run; cap SF-1000 mem-tier at 1 GiB

lukekim:
## What

Adds a scheduled **SF-1000 CH-benCH HTAP** run that exercises the Cayenne **cold object-store tier** (the merged #11543 feature), plus a small mem-tier tuning tweak to the SF-1000 memory configs.

- **New spicepod** `postgres-cayenne[file]-cdc-tuned-memory-cold-sf1000.yaml` — cloned from `…-cdc-tuned-memory-sf1000` with the cold tier enabled on the six write-heavy tables. Full cascade: **RAM mem-tier → warm local-disk Vortex → cold object store** (`s3://benchmarks/cayenne-cold-sf1000/`, the shared local MinIO). Cold knobs: `cold_target_file_size_mb: 1024`, `cold_tier_warm_max_bytes: 4 GiB`, `cold_tier_background_interval_ms: 30000`.
- **New dispatch** `dispatch/chbench/sf1000/postgres-cayenne[file]-cdc-tuned-memory-cold.yaml` — the SF-1000 HTAP cron auto-globs this directory, so it joins the scheduled batch (same runner/SF/terminals/duration/rate as the tuned-memory task).
- **`testoperator_run_htap.yml`** — now passes `S3_ENDPOINT/S3_KEY/S3_SECRET` to `spiced` from the `TEST_MINIO_*` secrets, mirroring `testoperator_run_bench.yml`. Unused/harmless for every other HTAP config.
- **Mem-tier cap**: halved `cayenne_cdc_mem_tier_max_bytes` **2 GiB → 1 GiB** on the high-volume block of all three SF-1000 memory configs (tuned-memory, ivm, cold) so they stay comparable. The 256 MiB reference-table value is unchanged.

## Why the anchor dataset

The warm tier is local `mode: file`, and Cayenne only auto-registers an `s3://` cold object store when the warm tier is itself on S3-Express (it reuses the warm store). Standard MinIO is not S3-Express, so the cold store must be registered another way. The tiny federated `cold_tier_store_anchor` dataset references the same bucket, which registers a **path-style** MinIO object store under `s3://benchmarks`; the object-store registry is shared (Arc-cloned) across the query and compaction runtimes, so cold promotion + cross-tier scans reuse it. A bare `s3://benchmarks` would otherwise resolve virtual-hosted and fail against MinIO.

## Verification

- ✅ YAML valid; no duplicate/reserved dataset names; cold params correct on the six tables; spicepod validator is parse-only (won't connect to S3 at validation time).
- ⚠️ **Not yet run end-to-end** — cold promotion actually firing and MinIO resolution can only be confirmed by a real SF-1000 HTAP run on the dev-xlarge runner.

## Reviewer notes / knobs to sanity-check

1. The anchor assumes `s3://benchmarks/tpch_sf1/region.parquet` exists in the shared test MinIO (referenced by ~19 other configs, so almost certainly present). If absent, repoint it at any existing object in the bucket.
2. `cold_tier_warm_max_bytes: 4 GiB` is the key promotion knob — sized so the big tables graduate after the initial-snapshot bootstrap while small ones stay warm. Promotion re-materializes the whole visible table, so it's deliberately not per-batch.
3. ~~Cold files accumulate under the shared-bucket prefix across scheduled runs (no cross-run GC).~~ **Resolved** (per @bjchambers): a post-run step in `testoperator_run_htap.yml` drops the whole `s3://benchmarks/cayenne-cold-sf1000/` prefix via `mc rm --recursive --force` after each run — gated to the cold config, `if: always()`. No collisions either — promotion paths are UUIDv7-unique and the query path reads the metastore manifest, not an S3 listing.

## Scheduling

Runs **once daily** (own cron `0 8 * * *` in `testoperator_dispatch_htap.yml`, its own `sf1000-cold/` dispatch dir) — a lighter cadence than the twice-daily `sf1000` batch while the cold tier is new. It shares the serialized `chbench-htap-sf100-sf1000` run concurrency group, so it queues rather than contends for the xlarge runner.


test/spicepods/chbench/accelerated/postgres-cayenne[file]-cdc-tuned-memory-cold-sf1000.yaml
```diff
@@ -0,0 +1,365 @@
...
+        cayenne_compaction_max_files_per_pick: '64'
+        cayenne_compaction_background_interval_ms: '3000'
+        cdc_max_coalesce_age_ms: '3000'
+        cdc_max_coalesced_bytes: '536870912' # 512 MB
+        # --- Cold object-store tier (RAM mem-tier -> warm local disk -> COLD) ---
+        # Presence of the location enables the cold tier (and forces key deletes, already set
+        # above). Cayenne namespaces cold files by table_id under this prefix, so one shared
+        # location is safe for every high-volume table. Resolves to the MinIO store registered by
+        # the `cold_tier_store_anchor` anchor dataset.
+        cayenne_cold_tier_location: s3://benchmarks/cayenne-cold-sf1000/
```
Comment (bjchambers) [inline test/spicepods/chbench/accelerated/postgres-cayenne[file]-cdc-tuned-memory-cold-sf1000.yaml:130]: Will we need to worry about collisions between runs? What about cleaning this up after runs?

bjchambers (dismissed):

test/spicepods/chbench/accelerated/postgres-cayenne[file]-cdc-tuned-memory-cold-sf1000.yaml
Comment (lukekim) [reply test/spicepods/chbench/accelerated/postgres-cayenne[file]-cdc-tuned-memory-cold-sf1000.yaml:130]: Good questions.

**Collisions — no.** Each promotion writes to `{cold_location}/{table_id}/cold/{promotion_id}/`, where `promotion_id` is a fresh UUIDv7 (`crates/cayenne/src/provider/table.rs`), so object paths are globally unique across promotions and across runs. The scan path also prunes cold files from the metastore manifest (not an S3 listing), so a fresh run (fresh local sqlite metastore) never references a prior run's objects even if they're still in the bucket — worst case they're dead weight, never a correctness issue.

**Cleanup — added.** New post-run step in `testoperator_run_htap.yml` that drops the whole cold prefix (`s3://benchmarks/cayenne-cold-sf1000/`) via `mc rm --recursive --force` after each run — gated to the cold config, `if: always()` so it also runs on failure/cancel. I put it in the run workflow rather than the testoperator binary because the `mc` client + `spice-minio` alias are already set up there and it mirrors the existing "Clean up PG replication slots" step (no Rust change/rebuild). Happy to move it into testoperator instead if you'd prefer it engine-side.

Comment (lukekim) [reply test/spicepods/chbench/accelerated/postgres-cayenne[file]-cdc-tuned-memory-cold-sf1000.yaml:130]: Added cleanup

.github/workflows/testoperator_run_htap.yml
```diff
@@ -224,3 +230,21 @@ jobs:
...
+
+      # The MinIO `benchmarks` bucket is shared across runs, so the Cayenne cold-object-store tier
+      # (cayenne_cold_tier_location: s3://benchmarks/cayenne-cold-sf1000/) would otherwise accumulate
+      # cold files every run: each promotion writes a new UUID-named prefix and nothing GCs them
+      # (the query path reads the metastore manifest, never an S3 listing, so a fresh run never
+      # references a prior run's objects — they are dead weight, not a collision). Drop the whole
+      # cold prefix after the run. Gated to the cold config; `mc` + the `spice-minio` alias come from
+      # the Install MinIO step above. Runs on failure/cancel too so a crashed run leaves no orphans.
+      - name: Clean up Cayenne cold-tier S3 objects
+        if: always() && contains(github.event.inputs.spicepod_path, 'cold')
```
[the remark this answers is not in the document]
lukekim: Good catch — narrowed the gate from `contains(..., 'cold')` to `endsWith(github.event.inputs.spicepod_path, 'cold-sf1000.yaml')`, so only the exact cold-sf1000 config (which owns the fixed `cayenne-cold-sf1000/` prefix) can trigger the cleanup.

.github/workflows/testoperator_run_htap.yml
```diff
@@ -224,3 +230,21 @@ jobs:
...
+        run: |
+          if command -v mc >/dev/null 2>&1; then
+            echo "Removing cold-tier objects under spice-minio/benchmarks/cayenne-cold-sf1000/ ..."
+            mc rm --recursive --force spice-minio/benchmarks/cayenne-cold-sf1000/ \
+              || echo "Nothing to remove (prefix already clean)."
```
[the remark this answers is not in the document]
lukekim: Fixed — removed the `|| echo "Nothing to remove"` that would misreport a real mc/auth/endpoint failure as "already clean". It's now `|| true`, which keeps the step non-fatal (a cleanup miss must not fail the benchmark run) while mc's own diagnostics — empty prefix or a genuine error — still print to the log.

sgrebnov (approved):

PR #11578, opened 2026-07-02
fix(cayenne): drain in-flight maintenance before in-process reopen (mutation_property flake)

bjchambers:
## Summary

Fixes an intermittent CI failure in the cayenne `mutation_property` convergence
tests (e.g. `prop_concurrent_mixed_key_sqlite`), which panicked with a
post-compaction **duplicate-row** signature — `missing(loss)=[]`, plus
`wrong_value(k, expected, -9223372036854775808)` where `i64::MIN` is the test's
duplicate-key sentinel for a primary key returned **twice** by the live scan. It
is a double-count, not data loss.

## Root cause

A scan reads the current snapshot **UNION** each protected snapshot with **no
key-level dedup**; dedup relies on a folded protected snapshot being removed from
`cayenne_snapshot_sequence`. The bug is a **compaction-vs-compaction race across
two provider instances** that share one metastore but hold **distinct in-memory
`compaction_lock`s**:

- a full-snapshot rewrite folds a set of protected snapshots and clears their
  catalog rows, while
- a concurrent protected-snapshot subset-merge (`swap_protected_snapshots_in_txn`)
  replaces those same snapshots with a merged one — whose row is **not** in the
  rewrite's folded set, so it survives.

`load_protected_snapshots` treats every registered snapshot as protected, so on
the next open the merged snapshot's rows are read again → duplicates.

The property test hit this because its `Restart` op — meant to emulate a process
restart — reopened the provider **without stopping the previous instance's
detached background maintenance**. A real crash/restart kills those tasks, so the
old test was an unfaithful restart that let two instances compact one catalog.
(Confirmed by tracing: the resurrected snapshot was a subset-merge output that no
fenced-delete ever cleared.)

## Fix

- Add `CayenneTableProvider::drain_in_flight_maintenance()` — awaits the detached
  post-write compaction / orphan-DV sweep / inline-checkpoint flags, flushes the
  post-write maintenance loop, and takes a final `compaction_lock` barrier.
- Call it before every reopen in `mutation_property_test`, making `Restart` a
  faithful clean restart.
- `.config/nextest.toml`: give the convergence test binaries (`mutation_property_test`,
  `cdc_compaction_delete_race_test`, `maintained_aggregate_filter_test`)
  `retries = 0` so a real convergence failure can't be masked as a flaky retry.
- Add `scripts/loaded_mutation_property_test.sh` (the flake only reproduces under
  CPU contention; this recreates it and runs with retries disabled).

## Validation

Under CPU contention with retries disabled (previously reproduced within 1–15
attempts): **15/15 sequential and concurrent runs pass**.

## ⚠️ Related production path (NOT fixed here) — tracked in #11581

The same "recreate a Cayenne provider in-process without draining the old one"
pattern exists in the runtime **hot-reload** path:

- `reload_accelerated_dataset` (`crates/runtime/src/init/dataset.rs`) creates and
  registers a new accelerated table **without** `remove_dataset` first (contrast
  the fallback path, which does).
- `create_cayenne_table_provider` always news up a fresh provider (no per-table
  cache) that spawns its own background compaction.
- The old provider's **detached post-write compaction** (`schedule_post_write_compaction`
  → `spawn_compaction`, `JoinHandle` discarded) is untracked/unabortable and keeps
  running against the shared metastore; `CayenneTableProvider` has no `Drop` that
  stops it.

So a spicepod hot-reload coinciding with in-flight compaction could trigger the
same duplicate-row corruption in a single production process. It is low-likelihood
and not exposed in normal operation, so it is filed for follow-up rather than
fixed here — see **#11581** for the full write-up and the options considered
(quiesce the old provider on hot-reload; reuse the existing provider; or a
catalog-level CAS on the fenced compaction commit as defense-in-depth).


crates/cayenne/src/provider/table.rs
```diff
@@ -9959,11 +9926,20 @@ impl CayenneTableProvider {
         // full snapshot rewrite) on the dedicated compaction runtime, isolated
         // from the query (compute) and CDC (refresh) runtimes.
         super::compaction::spawn_compaction(async move {
+            // Clear the coalescing flag on ANY exit — normal completion, early
+            // return, a panic during unwind, or the task being dropped on abort —
+            // so a stuck flag can never permanently suppress future compaction
+            // scheduling or hang `drain_in_flight_maintenance`.
+            struct ClearOnDrop(Arc<AtomicBool>);
```
phillipleblanc, line 9933: This is a pretty neat pattern

phillipleblanc (approved):

PR #11579, opened 2026-07-02
feat(spidapter): tuned + adaptive MongoDB CDC spicepods with scenario selection

bjchambers:
## What

Adds four hand-editable MongoDB CDC spicepods under `tools/spidapter/scenarios/pods/mongo/` and a scenario-driven way to deploy them, for benchmarking Cayenne CDC-over-MongoDB-change-streams tuning:

- **`mongo-sf10-tuned` / `mongo-sf100-tuned` / `mongo-sf1000-tuned`** — pin Cayenne's tunable actuators (write_concurrency, `cdc_mem_tier_max_bytes`, `target_file_size`, compaction cadence) per scale factor, seeded from the CH-benCHmark `postgres-cayenne[file]-cdc-tuned-memory-*` pods, with a high/low volume tier. Values are starting points for manual tuning.
- **`mongo-adaptive`** — sets `cayenne_tuning: adaptive` and leaves those actuators unset so Cayenne's on-box controller sizes them from real cores/RAM. Shares the capacity-independent knobs (memory durability, key deletes, CDC coalescing) with the tuned pods so a comparison isolates the actuator tuning.

Selection reuses the existing `--scenario` dispatch: `ScenarioConfig` gains an optional `spicepod:` path (env-substituted, e.g. `${MONGO_SPICEPOD_PATH:-}`); when non-empty the pod is deployed verbatim instead of generated. Empty/unset preserves the previous generate behaviour, and a metadata `spicepod_path` still takes precedence.

## Validation

Verified end-to-end against a local MongoDB replica set (SF1, rowcount-only). The scenario selection deploys the pod and spiced accepts every tuning param (`outcome: success`). A three-way per-checkpoint E2E convergence-latency comparison (the CDC lag signal), 31 checkpoints:

| Config | median | mean | p95 | p99 / max | checkpoints ≥50 s | total convergence |
|---|---|---|---|---|---|---|
| baseline (generated) | 5.09 s | 22.2 s | 110.3 s | 118.4 s | 8 / 31 | 688 s |
| **sf10-tuned** | 5.69 s | **10.7 s** | **52.6 s** | 107.3 s | **2 / 31** | **331 s** |
| adaptive | 5.12 s | 33.3 s | 123.3 s | 124.9 s | 13 / 31 | 1034 s |

The hand-tuned pod (`deletion_mode: key` + compaction tuning) roughly halves the spiky tail (spiking checkpoints 8→2, mean/p95 ~−52%). Adaptive is worse in a short run — the closed-loop controller hasn't converged its actuators in ~15 min — so the hand-tuned pods stay for now; adaptive gets a fair test at larger scale/longer duration in CI. Numbers are single-run under background load; the spike-count pattern is the robust signal.

Covered by a new `bundled_mongo_spicepods_parse` test that loads all four pods through the runtime `load_spicepod_from_path` path.

## Cross-repo note

Consuming these in CI requires the spidapter image to be rebuilt from this branch (pods live at `/app/scenarios/pods/mongo/`) and the spicebench dispatch wiring (separate PR).

### DIFF
--- a/tools/spidapter/scenarios/mongodb-streams.yaml
+++ b/tools/spidapter/scenarios/mongodb-streams.yaml
@@ -13,6 +13,8 @@ source:
     connect:
       uri: ${MONGODB_ATLAS_URI}
 # Optional: deploy a full tuned/adaptive spicepod verbatim instead of generating one.
-# Set MONGO_SPICEPOD_PATH to a scenarios/pods/mongo/*.yaml path (baked into the spidapter
-# image). Empty/unset → generate the default pod (previous behaviour).
+# Set MONGO_SPICEPOD_PATH to an ABSOLUTE path to a pods/mongo/*.yaml — in the spidapter
+# image that is /app/scenarios/pods/mongo/<pod>.yaml (e.g. mongo-sf100-tuned.yaml).
+# A relative value is resolved against the scenario base path (--scenario-base-path),
+# NOT the process CWD. Empty/unset → generate the default pod (previous behaviour).
 spicepod: ${MONGO_SPICEPOD_PATH:-}
--- a/tools/spidapter/scenarios/mongodb-streams.yaml
+++ b/tools/spidapter/scenarios/mongodb-streams.yaml
@@ -12,3 +12,7 @@ source:
   mongodb_streams:
     connect:
       uri: ${MONGODB_ATLAS_URI}
+# Optional: deploy a full tuned/adaptive spicepod verbatim instead of generating one.
+# Set MONGO_SPICEPOD_PATH to a scenarios/pods/mongo/*.yaml path (baked into the spidapter
+# image). Empty/unset → generate the default pod (previous behaviour).
+spicepod: ${MONGO_SPICEPOD_PATH:-}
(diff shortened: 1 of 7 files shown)


bjchambers:
Thanks @copilot — addressed all three in 110bb51:

1. **Relative `spicepod` path vs CWD** — a relative scenario `spicepod:` is now resolved against `--scenario-base-path` (where the scenario YAML + its `pods/` live), not the process CWD, and the resolved path is logged. Absolute paths (the common case: `MONGO_SPICEPOD_PATH` locally, or `/app/scenarios/pods/mongo/…` in the image) are used as-is.
2. **Test vs doc comment** — updated the comment to reference `parse_and_rename_spicepod` (the parser `load_spicepod_from_path` applies after reading the file), which is what the test calls; noted the pod bytes are embedded via `include_str!` so it covers exactly the committed files.
3. **Scenario YAML comment** — now documents `MONGO_SPICEPOD_PATH` as an absolute path (with the in-image `/app/scenarios/pods/mongo/…` example) and states that a relative value resolves against the scenario base path.

phillipleblanc (approved):

PR #11584, opened 2026-07-03
Bump datafusion (spiceai/datafusion#182) and datafusion-table-providers (#27): fix q16 CollectLeft planning error and SQLite q6 wrong revenue

Jeadie:
## Summary

Fixes two TPC-H correctness bugs, both via dependency bumps carrying targeted fixes:

### q16 — execution error (spiceai/datafusion#182)

```
SanityCheckPlan: HashJoinExec: mode=CollectLeft, join_type=LeftAnti, ...
does not satisfy distribution requirements: SinglePartition
```

Under partial federation fallback (`on_zero_results/*-duckdb` configs), `EnforceSorting` removed the `CoalescePartitionsExec` that `EnforceDistribution` had inserted on the build side of a CollectLeft LeftAnti join: a coalesce-link from the probe-side fallback scan wrapper propagated through the join, and the removal only ever inspected the first child. Fixed in spiceai/datafusion#182 (with a unit regression test); this PR bumps the datafusion pin `db1d14222c` → `0054497233`.

### q6 — SQLite wrong revenue, ~61 % of expected (spiceai/datafusion-table-providers#27)

`file[parquet]-sqlite[memory|file]` benchmarks have failed q6 on every trunk run since ~Jun 15 (`expected: 123141078.2283, actual: 75207768.1855` — exactly the `l_discount = 0.07` bucket dropped), while release/2.0 kept passing.

Root cause: the SQLite BETWEEN → `decimal_cmp` rewrite computes bound arithmetic like `decimal('0.06') + decimal('0.01')` with SQLite's raw `+`, which coerces the decimal extension's TEXT output back to REAL — so the bound is f64 (`0.069999999999999993`) after all. This was masked on SQLite ≤ 3.50, whose REAL→TEXT conversion (used by `decimal_cmp` for REAL args) rounds to 15 significant digits (`'0.07'`). The rusqlite 0.37 → 0.40 bump in #11118 moved the bundled SQLite from 3.50.2 to 3.53.2, which renders shortest-round-trip digits (`'0.06999999999999999'`) — removing the accidental rescue. Reproduced byte-for-byte outside spiced: the same accelerated DB + the same `rewritten_executor_sql` returns 123141078.2283 linked against 3.50.2 and 75207768.1855 against 3.53.2.

Fixed in spiceai/datafusion-table-providers#27 by evaluating bound arithmetic inside the decimal extension (`decimal_add`/`decimal_sub`/`decimal_mul` — exact on all SQLite versions); this PR bumps the table-providers pin to that fix merged with the `spiceai-54` tip.

An earlier revision of this PR enabled `sql_parser.parse_float_as_decimal` for q6; that has been removed — it is a session-wide literal typing change and does not address the actual SQLite mechanism (the federated SQL ships the *unfolded* BETWEEN arithmetic on this path, so DataFusion literal folding is not involved).

## Testing

- q16: regression test in spiceai/datafusion#182 (red without the fix), plus the datafusion-federation partial-federation q16 integration test goes red → green.
- q6/sqlite: unit regression test in spiceai/datafusion-table-providers#27 asserting the rewritten SQL for the exact q6 shape; fix verified correct against both bundled SQLite 3.50.2 and 3.53.2 via a C harness on the real accelerated database.
- Suggested CI check: dispatch `bench - tpch - accelerated/file[parquet]-sqlite[memory].yaml` against this PR's binary with `update_snapshots=no` and `validate_results=true` (note: `update_snapshots=default` resolves to `always` on non-release branches, which rewrites snapshots instead of failing).

## Known remaining issue (not addressed here)

DuckDB scan-level filter pushdown over Float64-widened accelerated columns still carries DataFusion's f64-folded literal (`l_discount <= 0.06999999999999999`) into the scan SQL. This path needs a fix upstream of SQL generation (literal folding or acceleration schema typing) and is pinned by a red regression test in datafusion-federation (`test_q6_duckdb_scan_filter_pushdown_double_lineitem`).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4023,7 +4023,7 @@ version = "3.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "faf9468729b8cbcea668e36183cb69d317348c2e08e994829fb56ebfdfbaac34"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -5626,7 +5626,7 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "54.0.0"
-source = "git+https://github.com/spiceai/datafusion.git?rev=db1d14222c15d0bd95fa90ad110b78cd21120401#db1d14222c15d0bd95fa90ad110b78cd21120401"
+source = "git+https://github.com/spiceai/datafusion.git?rev=00544972330059292e030240a5e66c500d6827c5#00544972330059292e030240a5e66c500d6827c5"
 dependencies = [
  "arrow",
  "arrow-schema",
@@ -5678,7 +5678,7 @@ dependencies = [
 [[package]]
 name = "datafusion-catalog"
 version = "54.0.0"
-source = "git+https://github.com/spiceai/datafusion.git?rev=db1d14222c15d0bd95fa90ad110b78cd21120401#db1d14222c15d0bd95fa90ad110b78cd21120401"
+source = "git+https://github.com/spiceai/datafusion.git?rev=00544972330059292e030240a5e66c500d6827c5#00544972330059292e030240a5e66c500d6827c5"
 dependencies = [
  "arrow",
  "async-trait",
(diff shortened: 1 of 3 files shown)


Jeadie:
Superseded by the clean dependency-bump-only PR (no runtime config changes): the q6 fix moved from `parse_float_as_decimal` to the actual root cause, spiceai/datafusion-table-providers#27 (SQLite 3.50.2→3.53.2 `decimal_cmp` bound evaluation). See the new PR for details.

PR #11589, opened 2026-07-03
fix(search): RRF chained FULL JOIN mis-keys docs absent from the first stream (fixes #11265)

### LINKED ISSUE #11265: RRF fusion: chained FULL JOIN keyed to the first stream's PK under-scores/duplicates docs absent from stream 1 (3+ candidate streams)
Found during a data-correctness audit of trunk. Two sites share the flaw:

**1. Multi-column `/v1/search` plan** — `crates/search/src/aggregation/reciprocal_rank.rs:396-413`: every subsequent table FULL-joins against `first_table_name`'s PK columns, which are NULL for documents absent from stream 1. A document matched by streams 2 and 3 but not stream 1 fails to join (NULL != key) and produces **two output rows for the same PK**, each carrying only part of the RRF sum (`coalesce(1/(rank+k), 0)` per side). There is no aggregation/dedup afterward — the API returns the same primary key twice with understated scores, displacing legitimate results within `limit`.

**2. `rrf()` UDTF** — `crates/runtime/src/search/rrf.rs:989-1020`: `fold_join` + `first_qualified_field` always resolves the join key to `search_0`'s column; a doc in q1∩q2 but not q0 splits into two rows. The later `aggregate(join_key, max(_fused_score))` (`rrf.rs:777-804`) re-merges them but takes `max(1/(k+r1), 1/(k+r2))` instead of the RRF sum `1/(k+r1) + 1/(k+r2)` — roughly half the correct fused score, wrong final top-K.

With exactly 2 streams the chained join degenerates correctly; the bug needs ≥3 fused streams (e.g. two embedding columns + FTS, or `rrf(text_search(...), vector_search(...), vector_search(...))`).

Suggested fix: join on `COALESCE`-folded keys (each join keys against the coalesced key of everything already joined), or aggregate by document key with SUM of per-stream contributions after the joins.

claudespice:
## Summary

Reciprocal Rank Fusion (RRF) chained a FULL OUTER JOIN across candidate streams but keyed **every** join on the **first stream's** copy of the primary key. Under a FULL OUTER JOIN that key is NULL for any document absent from stream 1, so a document matched by streams 2+ but not stream 1 failed the equality (`NULL != key`) and split into **one row per stream**, each carrying only part of the RRF sum. The bug needs **≥3 fused streams** to trigger (with exactly 2 the chained join degenerates correctly).

Two sites shared the flaw:

- **`crates/search/src/aggregation/reciprocal_rank.rs`** (the multi-column `/v1/search` plan): each subsequent table FULL-joined against `first_table_name`'s PK. There is **no aggregation afterward**, so the API returned the same primary key more than once, each row understating the score — displacing legitimate results within `limit`.
- **`crates/runtime-search/src/rrf.rs`** (the `rrf()` UDTF): `fold_join` resolved the join key via `first_qualified_field`, which always returns `search_0`'s column, so a document in q1∩q2 but not q0 split into two rows. The later `aggregate(join_key, max(_fused_score))` re-merged them but took `max` of the two partial contributions instead of their sum — roughly half the correct fused score, wrong final top-K.

## Root cause

Keying a chained FULL OUTER JOIN on a single stream's key is only correct when that stream is a superset of all others. RRF streams are independent candidate sets, so the "anchor" key is NULL for any document that stream missed.

## Changes

Both sites now key each join on the **COALESCE of the primary key across every already-joined stream**, so the folded key is non-NULL whenever the document appeared in *any* prior stream. Each document therefore produces exactly one row, and the per-row RRF sum (already computed downstream) is complete. This is the "join on COALESCE-folded keys" fix suggested in the issue; it needs no post-join aggregation change (site 1) and makes the existing `max` aggregation a no-op over single-row groups (site 2).

## Test plan
- [x] Added regression test `rrf_plan_dedups_doc_absent_from_first_stream_11265` (site 1): 3 streams where document C is in streams 2 & 3 but not stream 1; asserts exactly one row per document (pre-fix emitted a duplicate C row) and that C's fused score equals the full RRF sum `1/(2+k) + 1/(1+k)`, not a split fragment.
- [x] Added regression test `fold_join_dedups_doc_absent_from_first_stream_11265` (site 2): folds 3 streams and asserts `fold_join` produces exactly 4 rows for documents {A,B,C,D} (pre-fix emitted 5, duplicating C).
- [x] `make lint` passes (workspace-wide `cargo fmt --all -- --check` + full clippy)
- [x] `make test` passes

Fixes #11265

### DIFF
--- a/crates/runtime-search/src/rrf.rs
+++ b/crates/runtime-search/src/rrf.rs
@@ -1013,9 +1013,6 @@ impl ReciprocalRankFusion {
 
     // Reduces 2 or more search subquery DFs into a single one
     fn fold_join(a: DataFrame, b: DataFrame, join_key: &str) -> Result<DataFrame> {
-        let (tbl_a, id_a) = Self::first_qualified_field(&a, join_key)?;
-        let (tbl_b, id_b) = Self::first_qualified_field(&b, join_key)?;
-
         // Drop embedding columns before the full outer join — they are non-nullable
         // in Arrow but the join produces nulls for unmatched rows, causing a schema
         // violation. They are excluded from the final projection anyway.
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

Jeadie:
duplicate of #11519

PR #11590, opened 2026-07-03
perf(cayenne): reuse the PK RowConverter across the sharded CDC apply

lukekim:
## What

At `cdc_mem_tier_shards > 1`, the sharded CDC apply built the PK `RowConverter` redundantly for composite PKs — once per apply in `prepare_stream_for_insert_sharded`, and again **per batch** in `split_batch_by_pk_shard` — even though an identical converter was already in hand. This reuses it in both places:

1. `split_batch_by_pk_shard` takes `&RowConverter` and reuses the apply's converter (drops the per-batch `RowConverter::new`).
2. `prepare_stream_for_insert_sharded` sources the converter from the table's cached `pk_row_converter` (`Arc::clone`, via the existing "prefer cache, else build" fold) instead of rebuilding one per apply. `PreparedShardedInsertStream.converter` is now `Arc<RowConverter>`.

Net: a composite-PK sharded apply now does **zero** redundant `RowConverter::new` (2/apply → 1/apply → 0). `Int64` PKs (which have no cache) build once/apply.

## Why

CDC replication lag is throughput-bound — it stays under target only while sustained apply drain ≥ arrival — and the per-apply CPU on a sharded table is dominated by the PK validation/encode work. Removing the redundant converter builds from that hot path recovers per-apply CPU, most in the small-batch / high-frequency regime where a table falls behind.

This deliberately does **not** touch the per-table `write_lock`: it is the correctness spine of the sequence / apply-epoch / checkpoint invariants (each a documented data-loss fix), so narrowing it to run applies concurrently is unsound. This stays on the safe, config-free CPU lever.

## Measured

Micro-bench `benches/mem_tier_shard_split_converter.rs` — the split's PK-encode step (composite `(Int64, Utf8)` PK), rebuild-per-batch vs reuse:

| batch rows | rebuild | reuse | Δ |
|---|---|---|---|
| 1 | 230.7 ns | 130.6 ns | **−43%** |
| 8 | 228.0 ns | 159.3 ns | −30% |
| 64 | 522.2 ns | 447.7 ns | −14% |
| 512 | 2461.8 ns | 2375.8 ns | −3.5% |

The saving is a constant ~85–100 ns/batch (the removed `RowConverter::new`), so it is largest in the small-batch regime and decays for large batches. Strictly better — byte-identical shard routing, no latency↔throughput tradeoff.

## Correctness / guardrails

- Full `cargo test -p cayenne --lib`: **759 passed, 0 failed** — including the sharded-CDC equivalence tests (`sharded_cdc_lww_equivalence_across_shard_counts`, `sharded_cdc_concurrent_checkpoint_encode_equals_serial`, `sharded_cdc_intra_batch_duplicate_pks_equals_serial`) and `split_batch_by_pk_shard_partitions_keys_disjointly`.
- `clippy -D warnings` (lib + tests + bench): clean.
- The bench's correctness gate asserts reuse-across-batches is byte-identical to per-batch rebuild before any timing.

## Follow-up

The larger, O(rows) lever is the redundant PK **encoding**: the split `convert_columns`-encodes the PK rows for routing, then per-shard validation re-encodes them (`bloom_split_shard_batch` / `apply_on_conflict_to_batch`). Threading the encoded rows split→validate would save O(rows) per apply, but it refactors the correctness-critical validation path, so it is deferred to its own PR.


sgrebnov (dismissed):

krinart (dismissed):

bjchambers (dismissed):
100 ns/batch seems unlikely to shift the needle much, especially depending on where the bottleneck is, but this change doesn't hurt anything.

sgrebnov (approved):

PR #11596, opened 2026-07-04
fix: Distinguish no filters vs lit(false) for RefreshSql distributed accelerations

peasee:
## 📝 Summary

**Problem:**
* When partition filters for `RefreshSql` are empty (`[]`), no `WHERE` clause is applied. This results in partitioned accelerators retrieving the entire table contents instead of their intended contents of nothing for a table (equivalent to `lit(false)`).

**Fix:**
* Updates filters for `RefreshSql` to be defined by `Option<Vec<_>>` instead of `Vec<_>`. This allows the refresh SQL to distinguish between literally no filters (`None`) or filters which evaluate to matching nothing (`Some([]) == lit(false)`).


lukekim (dismissed):

Jeadie (approved):
```rust
/// `filters` uses the three-state `RefreshSQL` partition-filter semantics:
/// `None` (not partition-scoped, retrieve everything), `Some(filters)` (the
/// assigned partitions), or `Some(empty)` (no partitions assigned — load no
/// rows).
```
Could be worth explicitly using a three part Enum instead of `Option<T>` to model this. Would force at compile time what we need extensive code docs for.

PR #11603, opened 2026-07-04
fix: cluster filtered COUNT(*) on append-only tables returns unfiltered total (fixes #11599)

### LINKED ISSUE #11599: Cluster: filtered COUNT(*) on append-only Cayenne tables returns the unfiltered total (stats fold drops the pushed-down filter)
## Summary

On a cluster (distributed) deployment, `SELECT COUNT(*) ... WHERE <row-level predicate>` against an append-only Cayenne catalog table returns the **unfiltered** total row count. The planner folds the aggregate to a constant from table statistics while the pushed-down filter is silently dropped.

## Reproduction (observed on a live cluster app, 2026-07-03)

Cluster app with 4 executors; Cayenne catalog table created via:

```sql
CREATE TABLE IF NOT EXISTS spicebench.bench."customer" (
  "c_custkey" BIGINT, "c_name" TEXT, ..., 
  PRIMARY KEY ("c_custkey", "c_nationkey")
) PARTITION BY (bucket(5, c_nationkey))
```

150,000 rows ingested via ADBC bulk append (no updates/deletes). Then:

```sql
SELECT COUNT(*) FROM spicebench.bench.customer;
-- 150000  (correct)

SELECT COUNT(*) FROM spicebench.bench.customer WHERE c_custkey % 7 = 0;
-- 150000  (WRONG — expected ~21,428)
```

`EXPLAIN SELECT COUNT(*) FROM spicebench.bench.customer WHERE c_custkey % 7 = 0`:

```
logical_plan:
  Projection: count(Int64(1)) AS count(*)
    Aggregate: groupBy=[[]], aggr=[[count(Int64(1))]]
      SubqueryAlias: spicebench.bench.customer
        Union
          TableScan: spicebench.bench.customer projection=[], full_filters=[... c_custkey % Int64(7) = Int64(0)]
          TableScan: ... (x4, one per executor partition)

physical_plan:
  ProjectionExec: expr=[150000 as count(*)]
    BytesProcessedExec
      PlaceholderRowExec
```

## Analysis

Two conditions combine:

1. **The cluster partition table wrapper claims `Exact` filter pushdown for row-level predicates.** The filter lands in the logical `TableScan` as `full_filters`, so DataFusion deletes the `FilterExec`. Note `c_custkey` is *not* a partition column here — the base `CayenneTableProvider::supports_filters_pushdown` (`crates/cayenne/src/provider/table.rs:20616`) would correctly return `Inexact` for this predicate; the over-claim appears to come from the cluster wrapper (`crates/runtime/src/cluster/partition/mod.rs`).
2. **Reported statistics are `Precise` and unfiltered.** Executors broadcast live per-table `num_rows` (`local_executor_table_statistics`, `crates/runtime/src/cluster/partition/mod.rs:124`; reporter loop in `crates/runtime/src/lib.rs:998`). With exact stats on a scan whose filter was "exactly pushed down", DataFusion's aggregate-statistics rule folds `COUNT(*)` to the stats value — which never accounted for the filter.

Append-only state is required: any pending deletion degrades the stats and the fold no longer applies (which is why changes/CDC-mode tables don't show this). `SELECT` queries likely still filter correctly on the executors (the fragment re-applies predicates locally); the wrong-result surface is optimizer-derived values — filtered `COUNT(*)` confirmed, filtered `MIN`/`MAX` folds worth auditing.

## Impact

Silently incorrect results for any filtered count on append-only Cayenne tables in cluster mode (e.g. `COUNT(*) WHERE date = today` returns the all-time count). No error, no warning.

## Environment

Observed on a Spice Cloud cluster app running an image built from trunk `8cf43fee` plus an unrelated Cayenne write-path diff (touches neither statistics nor planning). A clean-trunk confirmation repro: any cluster deployment + append-only Cayenne catalog table + filtered count.

## Suggested fix

The cluster wrapper should not claim `Exact` pushdown for predicates it cannot prove are handled (mirror the base provider's partition-column check), and/or reported `Statistics` must be degraded to inexact whenever any filter is pushed to the scan — either alone prevents the fold.

claudespice:
## Summary

On a cluster (distributed) deployment, `SELECT COUNT(*) ... WHERE <row-level predicate>` against an append-only Cayenne catalog table returned the **unfiltered** total row count — the planner folded the aggregate to a constant from table statistics while the pushed-down filter was silently dropped. No error, no warning.

Fixes #11599

## Root cause

The coordinator represents every cluster table as per-executor **FlightSQL leaf scans** (`FlightSQLTable`), which stamp the executor-reported, *unfiltered* `num_rows`/column bounds onto each scan. Two things then combine in `crates/data_components/src/flightsql.rs`:

1. `FlightSQLTable::supports_filters_pushdown` returns `Exact` for any SQL-convertible filter (the predicate is shipped to and applied by the remote executor), so DataFusion drops the coordinator-side `FilterExec`.
2. `create_physical_plan` stamped the statistics **unconditionally**, so the scan reported the unfiltered `num_rows` as `Exact` even with a filter present.

With an exact row count on a scan whose filter was "exactly pushed down", DataFusion's `aggregate_statistics` rule folds `COUNT(*)` (and `MIN`/`MAX`) to the unfiltered value — the observed `ProjectionExec: expr=[150000 as count(*)]`. This single leaf is common to both mechanisms named in the issue, so fixing it here covers filtered `COUNT(*)` **and** the filtered `MIN`/`MAX` folds.

## Changes

- `FlightSQLTable::create_physical_plan`: mark the stamped statistics inexact (`Statistics::to_inexact()`) whenever `filters` is non-empty. This disables the constant-fold (the aggregate then executes over the actually-filtered remote stream) while the bounds remain `Inexact` and still usable for join sizing. Unfiltered scans keep exact statistics, preserving the fast bare-`COUNT(*)` fold.

Chosen over degrading the wrapper's pushdown claim because it's provider-agnostic and a single edit at the shared leaf.

## Test plan

- `make lint`
- `make test`
- New regression test `scan_marks_stamped_statistics_inexact_when_filter_pushed` (in `flightsql.rs`): stamps exact stats on a `FlightSQLTable`, asserts `num_rows`/column bounds stay `Exact` for an unfiltered scan and degrade to `Inexact` once a filter is pushed — fails on old code, passes on new.

## Performance impact

None on the fast path: unfiltered scans keep exact statistics, so bare `COUNT(*)` still folds. Only filtered scans lose the (incorrect) fold and instead execute the aggregate over the remote stream, which is the correct behavior.

## Checklist

- [x] Root cause identified and fixed at the shared leaf
- [x] Regression test added (fails on old code, passes on new)
- [x] Crate-scoped `cargo test`/`clippy`/`fmt` green
- [ ] Full `make lint` + `make test` green in CI

### DIFF
--- a/crates/data_components/src/flightsql.rs
+++ b/crates/data_components/src/flightsql.rs
@@ -485,8 +485,28 @@ impl FlightSQLTable {
         let exec = exec.with_token(self.token.clone());
         // Project the table-level statistics onto the scan's (projected) output
         // schema so the column-statistics list lines up with the output columns.
+        //
+        // When filters are pushed down to the remote scan, the stamped statistics
+        // still describe the *unfiltered* slice the executor reported (they are the
+        // whole-table row count / column bounds, not the filtered subset). Every
+        // pushdown-eligible filter is reported `Exact` (see
+        // `supports_filters_pushdown`), so DataFusion drops the coordinator-side
+        // `FilterExec` — and if the stamped `num_rows`/bounds stay `Exact`, the
+        // `aggregate_statistics` optimizer rule folds `COUNT(*)`/`MIN`/`MAX` to
+        // those unfiltered values, silently ignoring the predicate (e.g. a filtered
+        // `COUNT(*)` returns the full table count — issue #11599). Marking the
+        // statistics inexact when any filter is applied disables that fold while
+        // keeping the bounds usable for join sizing.
         let exec = match &self.statistics {
-            Some(stats) => exec.with_statistics(stats.clone().project(projections)),
+            Some(stats) => {
+                let stats = stats.clone().project(projections);
+                let stats = if filters.is_empty() {
+                    stats
+                } else {
+                    stats.to_inexact()
+                };
+                exec.with_statistics(stats)
+            }
             None => exec,
         };
         Ok(Arc::new(exec))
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

claudespice:
@copilot review

lukekim (approved):

PR #11604, opened 2026-07-04
fix(scylladb): error on lossy CQL decimal conversion instead of silent truncation/NULL (fixes #11267)

### LINKED ISSUE #11267: ScyllaDB: CQL decimal forced to Decimal128(38,2) — silent truncation for scale>2, silent NULL for wide mantissas; varint/collections/UDTs serialized via Rust Debug
Found during a data-correctness audit of trunk. Two value-corruption shapes in the ScyllaDB connector's type mapping (`crates/db_connection_pool/src/dbconnection/scylladbconn.rs`):

**1. Arbitrary-precision CQL `decimal` hardcoded to `Decimal128(38, 2)`** (`scylladbconn.rs:30-31, 446-462, 622-653`): a stored `12.345` is rescaled by integer division (round-toward-zero) to `12.34` — digits silently dropped; any value whose unscaled mantissa exceeds 16 bytes becomes silent NULL (`cql_decimal_to_i128` -> `None` -> `append_null`). Any table whose decimals aren't scale ≤ 2 returns wrong numbers with no warning. Suggested: error loudly on truncation/overflow, or map to a configurable/inferred precision-scale.

**2. `varint`, `duration`, lists/sets/maps, tuples, and UDTs emitted as Rust `Debug` strings** (`scylladbconn.rs:464-487, 559-570`), e.g. a `varint` column renders like `CqlValue::Varint(CqlVarint([...bytes...]))` and a list like `List([Int(1), Int(2)])` — while the mapping comments (`scylladbconn.rs:673-687`) promise JSON. Consumers parsing the promised JSON get corrupt data; query succeeds silently.

Related config gap, same connector: `ssl` is declared in `PARAMETERS` (`crates/data-connectors/connector-scylladb/src/lib.rs:130`) but never read — `scylladb_ssl: true` validates and is silently ignored (connection stays plaintext).

claudespice:
## Summary

The ScyllaDB connector mapped every CQL `decimal` to a fixed `Decimal128(38, 2)` and rescaled with unchecked integer division + a 16-byte fit check. This silently corrupted data:

- a scale>2 value (e.g. `12.345`) was truncated toward zero to `12.34` — digits dropped, no warning;
- a mantissa wider than 16 bytes became a silent `NULL`;
- a sufficiently negative source scale hit the unchecked `10_i128.pow(..)` (overflow / debug panic).

Fixes #11267

> Scope note: this addresses **item 1** (decimal precision) of the issue. Items 2 (varint/collections/UDT `Debug`-vs-JSON serialization) and 3 (declared-but-ignored `ssl` param) are separate mechanisms and are intentionally left for follow-up — the issue should stay open for them.

## Root cause

`cql_decimal_to_i128` rescaled `int_val / 10^source_scale` to the column's fixed `target_scale` using `value / 10^(-scale_diff)` (round-toward-zero, dropping remainder) on down-scale, returned `None` (→ `append_null`) on a >16-byte mantissa, and used the unchecked `10_i128.pow(..)` on up-scale. All three lose data with no error.

## Changes

- `cql_decimal_to_i128` now returns `Result<i128, String>` and is **total and honest**:
  - down-scale succeeds only when the division is exact (zero remainder); a non-zero remainder returns a descriptive error rather than truncating;
  - up-scale uses `checked_pow`/`checked_mul` and reports overflow instead of panicking;
  - a >128-bit mantissa is rejected with a clear message.
- The `Decimal128` column builder converts an `Err` into a descriptive `Error::ConversionError` naming the column, instead of writing a wrong number or NULL.
- Losslessly-representable values (exact down-scale, in-range up-scale, equal scale) are unaffected.

## Test plan

- `make lint`
- `make test`
- New unit-test regressions in `scylladbconn.rs`:
  - `test_cql_decimal_to_i128_lossless_conversions` — equal-scale passthrough, up-scale, exact down-scale, empty mantissa;
  - `test_cql_decimal_to_i128_lossy_downscale_errors` — `12.345`→scale 2 errors instead of truncating (both signs);
  - `test_cql_decimal_to_i128_overflow_errors_without_panicking` — 17-byte mantissa and extreme negative source scale report errors instead of NULL/panic.

## Checklist

- [x] Regression tests added (fail on old code, pass on new)
- [x] Crate-scoped `cargo test`/`clippy`/`fmt` green
- [ ] Full `make lint` + `make test` green in CI

### DIFF
--- a/crates/db_connection_pool/src/dbconnection/scylladbconn.rs
+++ b/crates/db_connection_pool/src/dbconnection/scylladbconn.rs
@@ -1917,14 +1917,26 @@ mod tests {
     #[test]
     fn test_cql_decimal_to_i128_lossless_conversions() {
         // Equal scale: mantissa passes through unchanged, both signs.
-        assert_eq!(cql_decimal_to_i128(&decimal_mantissa_be(1234), 2, 2), Ok(1234));
-        assert_eq!(cql_decimal_to_i128(&decimal_mantissa_be(-1234), 2, 2), Ok(-1234));
+        assert_eq!(
+            cql_decimal_to_i128(&decimal_mantissa_be(1234), 2, 2),
+            Ok(1234)
+        );
+        assert_eq!(
+            cql_decimal_to_i128(&decimal_mantissa_be(-1234), 2, 2),
+            Ok(-1234)
+        );
 
         // Up-scaling always adds trailing zeros without loss: 12 (scale 0) -> scale 2 == 1200.
-        assert_eq!(cql_decimal_to_i128(&decimal_mantissa_be(12), 0, 2), Ok(1200));
+        assert_eq!(
+            cql_decimal_to_i128(&decimal_mantissa_be(12), 0, 2),
+            Ok(1200)
+        );
 
         // Exact down-scale: 12.3400 (mantissa 123400, scale 4) -> scale 2 == 1234, no digits lost.
-        assert_eq!(cql_decimal_to_i128(&decimal_mantissa_be(123_400), 4, 2), Ok(1234));
+        assert_eq!(
+            cql_decimal_to_i128(&decimal_mantissa_be(123_400), 4, 2),
+            Ok(1234)
+        );
 
         // An empty mantissa is zero.
         assert_eq!(cql_decimal_to_i128(&[], 0, 2), Ok(0));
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

claudespice:
@copilot review

lukekim (approved):

PR #11605, opened 2026-07-04
bench(chbench): add 4-slot CDC variant pod for SF-1000 memory (separate from default)

bjchambers:
Isolate both heavy WAL producers (order_line, stock) and the bursty apply-bound new_order onto their own logical slots, sharing one slot for the light upsert tables (customer/district/oorder/warehouse). Chosen from three SF-1000 diagnostic runs that bracketed the server-decode vs client-pump trade-off:

  - 7 slots (run 28684210964): server re-decodes the WAL 7x; order_line 0.50x / stock 0.58x realtime — walsender decode was the ceiling.
  - 1 shared slot (run 28692447178): decode 1x, but one client pump couldn't drain the merged stream (received 0.03x) and new_order's ~84s bursts head-of-line- blocked the connection into a wal_sender_timeout reset storm.
  - 3 groups (run 28694417436): isolation fixed the resets, but stock dragged the 5-table 'main' group's shared pump to 0.39x diverging.

4 groups holds server decode at 4x while giving each heavy/bursty table its own pump and sharing only the ~10M-change light tables. Spicepod-only; no binary change.

### DIFF
--- a/test/spicepods/chbench/accelerated/postgres-cayenne[file]-cdc-tuned-memory-sf1000-4slots.yaml
+++ b/test/spicepods/chbench/accelerated/postgres-cayenne[file]-cdc-tuned-memory-sf1000-4slots.yaml
@@ -1,7 +1,16 @@
 version: v1
 kind: Spicepod
-name: postgres-cayenne[file]-cdc-tuned-memory-sf1000
+name: postgres-cayenne[file]-cdc-tuned-memory-sf1000-4slots
 
+
+# EXPERIMENTAL 4-slot variant of `postgres-cayenne[file]-cdc-tuned-memory-sf1000`. Identical to
+# that pod EXCEPT the changes-mode tables are split across 4 shared replication slots (see the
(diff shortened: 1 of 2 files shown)


lukekim (dismissed):

bjchambers:
Restructured per review: instead of editing the shipped `postgres-cayenne[file]-cdc-tuned-memory-sf1000` pod, this now **adds a new `…-memory-sf1000-4slots` pod** and leaves the default untouched. That lets us benchmark + iterate on the grouped-slot topology without perturbing the default pod existing benchmarks track.

Grouping alone does not make SF-1000 converge — all three diagnostic topologies (7-slot / 1-slot / 3-group) still stall the heavy tables (`order_line` ~0.5×, `stock` ~0.58× received ×realtime). So this is **not** proposed as the new default yet; promoting it should wait until the downstream source-decode / durable-apply work lands. Slot grouping: `order_line` | `stock` | `new_order` | `{customer,district,oorder,warehouse}`.

(Re-requesting review since the diff changed from an edit to a new file.)

sgrebnov (approved):

PR #11607, opened 2026-07-04
fix(postgres): shared replication slot drops WAL for partitioned tables (fixes #11290)

### LINKED ISSUE #11290: Shared Postgres replication slot silently drops all WAL changes for partitioned tables (publication lacks publish_via_partition_root)
Found while auditing #11255 (share one replication slot across multiple changes-mode datasets).

## Problem

The shared-slot publication is created without `publish_via_partition_root = true` (`crates/data_components/src/postgres_replication/slot.rs`). For a partitioned source table, pgoutput therefore emits changes under each **leaf partition's** `(schema, name)`, while the shared router's `MemberKey` is the **parent** table's name. In `handle_decoded` (`crates/data_components/src/postgres_replication/shared.rs`), `source.member(&member_key)` finds no member for the leaf relation, the changes are dropped at DEBUG level, and — because the registered member is streaming and idle — `credit_idle` advances the ack floor past them.

The initial snapshot (a `SELECT` on the parent) is correct, so the dataset boots with correct data and then silently never updates. Masking bonus: `pg_publication_tables` lists leaves rather than the parent, so `has_table` is false on every restart → `table_added = true` → a fresh snapshot each boot hides the staleness between restarts.

The per-dataset (non-shared) path doesn't have this routing-by-name behavior, so this is a shared-mode regression relative to single-dataset replication.

## Fix directions

- Create/alter the publication with `publish_via_partition_root = true` so pgoutput resolves changes to the parent relation, or
- reject partitioned parents in `validate_replica_identity` for shared mode with a clear error until supported, or
- route leaf relations to the parent member (requires partition-hierarchy lookup).

Also worth a `replication_shared.rs` integration test with a partitioned source table asserting post-snapshot INSERT/UPDATE/DELETE land in the accelerator.

claudespice:
## Summary (root cause)

The shared-slot publication was created without `publish_via_partition_root = true` (`crates/data_components/src/postgres_replication/slot.rs`). For a partitioned source table, pgoutput therefore emits changes under each **leaf partition's** `(schema, name)`, while the shared router's `MemberKey` is the **parent** table's name. In `handle_decoded`, `source.member(&member_key)` finds no member for the leaf relation, the changes are dropped at DEBUG level, and `credit_idle` advances the ack floor past them. The initial snapshot (a `SELECT` on the parent) is correct, so the dataset boots with correct data and then silently never updates.

Masking bonus (also fixed): `pg_publication_tables` listed leaves rather than the parent, so `has_table` was false on every restart → a fresh snapshot each boot hid the staleness between restarts. With `publish_via_partition_root` the catalog lists the parent, so `has_table` is correct.

Fixes #11290

## Changes

- `CREATE PUBLICATION ... WITH (publish_via_partition_root = true)` when creating the shared publication. No effect on regular (non-partitioned) tables.
- New `ensure_publish_via_partition_root` helper repairs a publication created without the option by an older Spice build or a DBA (`ALTER PUBLICATION ... SET (...)`). It is a no-op once set; when Spice does not own a pre-created publication the `ALTER` (SQLSTATE 42501) is downgraded to a warning rather than failing setup, since the option only matters for partitioned sources.
- Regression test `shared_slot_partitioned_source_table_streams_changes` in `crates/runtime/tests/postgres/replication_shared.rs`: a range-partitioned parent with two leaf partitions asserts the snapshot covers both partitions, the publication lists the parent, and post-snapshot insert/update/delete (each routed through a leaf partition by Postgres) reach the stream. Fails on old code (changes dropped → timeout), passes on new.

## Test plan

- `make lint`
- `make test`

## Checklist

- [x] Root cause identified and fixed at the publication-creation site (single site; covers both shared and per-dataset paths)
- [x] Existing publications repaired in place (upgrade path), tolerant of non-owned publications
- [x] Regression test added that fails on old code and passes on new
- [x] `cargo check -p data_components` and `cargo check -p runtime --test integration --features postgres,postgres-accel` pass; `cargo clippy -p data_components --no-deps` clean; `cargo fmt` applied

### DIFF
--- a/crates/data_components/src/postgres_replication/slot.rs
+++ b/crates/data_components/src/postgres_replication/slot.rs
@@ -404,6 +404,12 @@ async fn ensure_publication(
         .get(0);
 
     if exists {
+        // Repair publications created by an older Spice build (or a DBA)
+        // without `publish_via_partition_root`. Without it, pgoutput attributes
+        // a partitioned table's changes to each leaf partition, which the
+        // shared router — keyed by the parent's name — silently drops (#11290).
+        ensure_publish_via_partition_root(client, publication_name).await?;
+
         // Verify the publication includes our table; if not, add it.
         let has_table: bool = client
             .query_one(
@@ -430,7 +436,13 @@ async fn ensure_publication(
     }
 
     let stmt = format!(
-        "CREATE PUBLICATION {pub} FOR TABLE {schema}.{table}",
+        // `publish_via_partition_root = true` makes pgoutput report a
+        // partitioned table's changes under the parent relation (matching the
+        // shared router's parent-keyed members and the parent-table initial
+        // snapshot) instead of each leaf partition. No effect on regular
+        // tables. See #11290.
+        "CREATE PUBLICATION {pub} FOR TABLE {schema}.{table} \
+         WITH (publish_via_partition_root = true)",
         pub = quote_ident(publication_name),
         schema = quote_ident(schema_name),
         table = quote_ident(table_name),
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

sgrebnov (approved):

PR #11610, opened 2026-07-04
observability(cdc): localize CDC back-pressure across the ingest pipeline

bjchambers:
Adds end-to-end CDC back-pressure instrumentation so we can tell **where** the Postgres→cayenne ingest pipeline falls behind (not just that it does), plus the tooling to analyze it. Observability + diagnosis only — no behavior change to the ingest path.

## Metrics (runtime / cayenne / data_components / telemetry)
- Prefetch-channel occupancy + capacity; coalesce flush-reason counter + batch-age; source-arrival lag; apply-cycle + reader send-wait timings.
- Received-commit-ts gauge mirroring applied-commit-ts → received/applied **progress ×realtime** rate ladder.
- Cayenne write-path valves: encode-budget snapshot + acquire-wait, compaction semaphore permits + acquire-wait, mem-tier used/refusals, `encode_permit_wait` phase (occupancy via scrape-time observable gauges — zero hot-path cost).
- Reader input-wait vs decode split; exported `confirmed_flush`/`server_wal_end` LSN; `replication_disconnected_ms_total` (reconnect downtime) on both replication paths.
- Durable-delete apply-path decomposition (`durable_delete_lock_wait`/`_apply`/`_maintenance`) + `cdc_apply_path_total{path}` counter — closes the phase-coverage blind spot where a delete-heavy table's apply time was ~99% un-instrumented.
- **Perf:** the shared replication pump accumulates reader timing locally and fans out per commit/keepalive instead of per row (removes a per-row member-iteration fan-out on the hot decode path).

## Harness (testoperator)
- `--metrics-dump` JSON artifact (full series + run metadata); source-side `pg_stats` scraper (SQL-only); `clock_timestamp()` skew probe embedded in the artifact.
- Apply-phase coverage self-check with a `--min-phase-coverage` gate (names the offending table); staleness-probe outlier clamp; post-drain caught-up flagging.

## CI + analysis
- `testoperator_run_htap`: run-summary header, waterfall render into the job summary, metrics-dump + log artifact upload, robust `PIPESTATUS` exit propagation (fixes a false-green on bootstrap/ready-wait timeouts), composed with the existing `collect_diagnostics` probes.
- `scripts/chbench-waterfall.py`: per-dataset lag decomposition, progress-rate ladder, reader split, source-side pg_stats, two-level phase coverage (blind-spot localizer), apply-path mix, stream-health, flush-reason mechanism text, clock-skew-corrected lag, estimator/window provenance. `docs/cdc-backpressure-waterfall.md` documents the methodology.

Validated end-to-end on an SF-1000 CI run; the diagnosis it produced is what motivates the separate slot-tuning experiment (#11605).


lukekim (approved):

lukekim (dismissed):

bjchambers:
Addressed the review comments in `61b693c18`:

- **`hist_pcts` cumulative double-count** — fixed. It now takes each (label-set, `le`) series' latest cumulative bucket value, then sums across label sets within that snapshot, instead of summing `_bucket` across all scrapes (which multiplied/time-weighted cumulative counts and skewed the valve p90/p99). This matches how the Python waterfall computes quantiles.
- **Blocking I/O in `write_metrics_dump`** — fixed. The function is now `async`; the ~34 MB JSON serialize + filesystem write run in `tokio::task::spawn_blocking` so they can't stall the async runtime.
- **Contradictory dump-reduction comment** — fixed. The comment now correctly states that cumulative series keep the **first and last** sample per label-set (for the windowed-delta analysis), matching `reduce_samples_for_dump`.

Also included some related hardening while in these files: guard the received/applied commit frontiers + lag gauge against heartbeat envelopes (a keepalive mid-backlog was inflating the applied frontier past the received one), record authoritative per-slot retained WAL from `pg_replication_slots` in the source-side sampler (the client-view backlog understates when the walsender is WriteData-blocked), surface staleness-probe discard counts, and clamp a negative `xact_commit` delta. Validated on a fresh 4-slot SF-1000 run.

phillipleblanc (dismissed):

bjchambers:
Latest review comments addressed:

- **caught_up unwrap_or(0.0) false-positive** (reporting.rs) — fixed in the latest commit: the caught-up check now requires the lag_bytes series to be PRESENT and 0 (is_some_and), so a dataset whose series was not scraped is never flagged caught-up. (This also now requires the authoritative per-slot retained WAL to be ~0, and warns on the client-zero-but-authoritative-large write-blocked case.)
- The **hist_pcts cumulative double-count**, **write_metrics_dump blocking I/O**, and **contradictory reduce-comment** were already fixed earlier (commit 61b693c18) — the inline threads are pinned to the pre-fix lines.

bjchambers:
Merged latest `trunk` and addressed the latest Copilot review (commit `814a63ba2`).

**Merge:** trunk added shared-slot send-stall tracking (`replication_member_send_stalled_seconds_total`). Both conflicts were purely additive — the liveness metrics here (`disconnected_ms`, `member_attached`, `slot_name`) alongside trunk's new counter — resolved by keeping both.

**Review fixes:**
- **RwLock poison recovery** (compaction-semaphore publish/snapshot in `dataaccelerator/cayenne/mod.rs`; `set_slot_name`/`slot_name()` in `postgres_replication/metrics.rs`): recover the inner value via `PoisonError::into_inner` instead of `.read().ok()`/`if let Ok(..)`, so an unrelated panic can't permanently disable a gauge or drop the shared-slot label. Observability shouldn't degrade silently.
- **`disconnect_at` on failed connects** (`postgres_replication/shared.rs`): the transient-connect-failure branch now starts `disconnect_at` via `get_or_insert_with` (preserving an earlier drop timestamp), so boot-time / never-yet-connected outages contribute to `replication_disconnected_ms_total` instead of being dropped.

`cargo fmt --all` + `cargo clippy` on the touched crates (util, data_components, connector-postgres, runtime, testoperator, cayenne) are clean — the only remaining clippy warnings are pre-existing trunk code (`iceberg.rs`, `builder.rs`), untouched here.

bjchambers:
Addressed the latest Copilot review (commit `76ac581e5`), and reconciled with the autofix commits already on the branch:

- **Authoritative slot-retained uses the most-recent `pg_stats` sample** (nearest the `/metrics` scrape), not the window max — a transient early spike no longer makes a later/idle scrape falsely read as write-blocked / not-caught-up.
- **Post-drain re-scrape re-samples `pg_stats` fresh** (new `PgStatsScraper::sample_once_now`) instead of reusing the under-load samples, which were stale by then. Best-effort: falls back to client-view-only if the one-shot connect fails.
- **`member_attached` doc** clarified — it's set to `1` by `MetricsCollector::new()` (the sole constructor); the derived `Default` leaves it `0`.
- **Dropped the `--markdown` flag** from `chbench-waterfall.py` and the workflow call — it was parsed but never altered output (the workflow already wraps the summary in a code fence itself).
- **Centralized the duplicated `now_unix_ms`** into `spiced_metrics` (pg_stats reuses it). Kept the `0`-on-error fallback with a comment noting it's intentional: the error only occurs with a pre-1970 system clock, which can't happen on a runner — so propagating a `Result` through every scrape-timestamp would be needless plumbing for a can't-happen case.

Kept the autofix commits' other changes: the `!from_carried` guard on the received-frontier/arrival-lag recording (avoids double-counting a carried-over envelope) and the 5s timeout around `sample_once` in the scraper loop.

`cargo fmt` + scoped `cargo clippy` (with the lint-rust deny flags) on the touched crates are clean.

bjchambers:
Addressed the arrival-lag under-sampling finding (commit `a98e37c48`).

The `!from_carried` guard was skipping `cdc_source_arrival_lag_ms` whenever a burst started from a carried envelope. Under sustained byte-cap back-pressure an item is carried on nearly every burst, so the histogram was systematically under-sampled in exactly the backlogged regime the metric exists to diagnose (biasing p99 low). Rather than skip, I capture the carried item's receipt wall-clock when it's carried (`carried_received_ms`) and record its arrival lag on the next iteration against *that* timestamp — the true receipt time, not process time (which would fold this burst's apply wait into the sample). Every burst now contributes one accurate arrival-lag sample.

Combined cleanly with the autofix commit that relocated the received-commit frontier into `apply_burst` (recorded once per burst at the freshest commit ts, so received never appears to lag applied) — that's a separate metric, so both changes coexist: the frontier moves to `apply_burst`, arrival lag stays per-burst-first with the receipt-time fix.

bin/spiced/src/lib.rs
```diff
@@ -835,6 +835,14 @@ pub async fn run(args: Args) -> Result<()> {
...
         tokio_handles.push(("main", Handle::current()));
         telemetry::register_tokio_runtime_metrics(tokio_handles);
+
+        // Cayenne write-path backpressure occupancy gauges (encode budget, in-memory
+        // CDC tier byte budget, compaction semaphore). Pull-based observable gauges on
+        // the global `cayenne` meter; registered here — after `init_metrics` — for the
+        // same reason as the compaction metrics above (bind to the real Prometheus
+        // meter, not the early noop one). Localizes *which* valve is stalling the CDC
+        // apply path when ingest falls behind.
+        runtime::dataaccelerator::cayenne::register_cayenne_backpressure_gauges();
```
phillipleblanc: nit: instead of saying this is explicitly backpressure gauges, something more generic like `register_cayenne_telemetry` would make it more flexible

crates/telemetry/src/lib.rs
```diff
@@ -561,6 +561,52 @@ pub fn track_cayenne_mem_tier_epoch(apply_epoch: u64, durable_epoch: u64, dimens
...
+static CAYENNE_MEM_TIER_RESERVE_REFUSED: OnceLock<Counter<u64>> = OnceLock::new();
+
+/// Counts in-memory CDC tier reservation refusals (`cdc_durability: memory`): a
+/// `try_reserve_bytes` that could not fit an append into the process-global
+/// mem-tier byte budget, forcing the writer to wait for another table's
+/// checkpoint, spill its own tier, or fall back to the durable path. A non-zero,
+/// growing count is direct memory-mode ingest backpressure — the mem-tier budget
+/// is the bottleneck (pair with `cayenne_mem_tier_budget_used/total_bytes`
+/// occupancy). No labels: the budget is process-global.
+pub fn track_cayenne_mem_tier_reserve_refused() {
```
phillipleblanc: I think it would be cleaner to put these methods in a `cayenne` sub-module, i.e. so it would be `telemetry::cayenne::track_mem_tier_reserve_refused`

docs/cdc-backpressure-waterfall.md
phillipleblanc, line 1: Remove this file (or move it to a more suitable location)

phillipleblanc (dismissed):

claudespice:
Pushed a small fix for the failing **Rust Lint** check: `tools/testoperator/src/args/htap.rs` used `clap::value_parser!(f64).range(0.0..=1.0)`, but clap only exposes `.range()` on its integer value parsers, so this failed to compile (`E0599: no method named `range``). Replaced it with a small `parse_phase_coverage` validator that enforces the same inclusive `0.0..=1.0` bound. `cargo check`/`clippy` (CI pedantic flags) are clean for the crate.

Two things left that are yours to decide:
- The **Check Rust Licenses** failure is unrelated to this PR — it's the trunk-wide `RUSTSEC-2026-0204` crossbeam-epoch advisory, tracked/fixed by #11681. Merging trunk once that lands should clear it (this branch is currently BEHIND trunk).
- 3 open review threads from @phillipleblanc (rename to `register_cayenne_telemetry`, move telemetry fns into a `telemetry::cayenne` submodule, and remove/relocate `docs/cdc-backpressure-waterfall.md`) are design calls I left for you rather than changing on your active PR.

sgrebnov (dismissed):

sgrebnov (dismissed):

sgrebnov (approved):

Issue #11263, opened 2026-06-10
cosine_distance returns (1-sim)/2 locally but 1-sim when federated to DuckDB — same query, scores 2x apart


claudespice:
Found during a data-correctness audit of trunk.

`crates/runtime-datafusion-udfs/src/cosine_distance.rs:225-228` computes `(1.0 - similarity) / 2.0` (range [0,1]), while the UDF is allowed to federate to DuckDB where it unparses to `array_cosine_distance` = `1 - sim` (range [0,2]) — `crates/runtime/src/datafusion/udf.rs:668`, `crates/data-connectors/connector-duckdb/src/lib.rs:80-81`.

Consequences:
- `SELECT cosine_distance(a, b) FROM t` returns values 2x apart depending on whether the plan federates.
- `vector_search`/`/v1/search` compute score = `1 - cosine_distance`, documented as "cosine similarity" (`crates/runtime/src/embeddings/udtf.rs:180,1066-1074`) — with the local formula the actual score is `(1+sim)/2`, so orthogonal vectors score 0.5, while the DuckDB search-index path (`crates/search/src/index/duckdb/sql.rs:341`) yields the true `sim`. User score thresholds (e.g. `score > 0.7`) silently select different rows per execution path. Ranking within one path is unaffected (monotonic).
- Related: no zero-vector guard — all-zero embeddings produce NaN, which sorts above all numbers in `ORDER BY _score DESC` (`cosine_distance.rs:225`), surfacing failed/zero embeddings as top matches.

Not fixing unilaterally because either direction changes user-visible score values; flagging for a maintainer decision. Aligning the local formula to the standard `1 - sim` (matching DuckDB and the "cosine similarity" score docs) seems right, plus a zero-vector guard (return NULL or max distance).

claudespice:
### Full blast-radius analysis of the formula divergence (for the maintainer decision)

I traced every path that produces a cosine score. The **local UDF formula is the sole outlier** — every DuckDB path and the documented `_score` semantics already agree on `score = sim` (true cosine similarity):

| Path | distance | `_score = 1 - distance` |
|------|----------|--------------------------|
| Local `cosine_distance` UDF (`cosine_distance.rs`) | `(1 - sim) / 2` | `(1 + sim) / 2`  ← **outlier** |
| DuckDB federated UDF (`cosine_distance` → `array_cosine_distance`, `dialect/duckdb.rs`) | `1 - sim` | `sim` |
[...]

PR #11621, opened 2026-07-05
fix(search): cosine_distance returns NULL not NaN for zero-magnitude vectors

claudespice:
## Summary (root cause)

`cosine_distance(x, y)` computed `similarity = sum_squares / (‖x‖ · ‖y‖)`. When either vector has zero magnitude — an all-zero or failed/empty embedding — the divisor is `0.0`, so `similarity` becomes `0.0 / 0.0 = NaN` and the function returned `Some(NaN)`.

Vector search scores each row as `_score = 1 - cosine_distance(...)` and sorts with `ORDER BY _score DESC`. `1 - NaN` is still `NaN`, and `NaN` sorts **ahead** of every real score, so rows backed by a failed or all-zero embedding surface as the **top** search matches instead of being pushed to the bottom.

This addresses the zero-vector portion of #11263.

## Changes

- `crates/runtime-datafusion-udfs/src/cosine_distance.rs`: guard the similarity for finiteness. When `similarity` is not finite (the only sources are a zero-magnitude vector → `NaN`, or an overflow → `±inf`), `cosine_distance` now returns `None`, which propagates as a SQL `NULL`. With `ORDER BY _score DESC NULLS LAST`, failed/empty embeddings sort last rather than first.
- The internal `cosine_distance` helper now returns `Option<f64>`; `compute_cosine_distance` forwards the `None` (it already returns `Ok(None)` for the NULL/empty-input cases, so this is consistent with existing behaviour).
- Distances for non-zero vectors are unchanged, so no existing search snapshots drift.

## What this PR deliberately does NOT change

The headline of #11263 — the local formula `(1 - sim) / 2` (range `[0, 1]`) diverging from DuckDB's `array_cosine_distance = 1 - sim` (range `[0, 2]`) that `cosine_distance` federates to — is **left for a maintainer decision**. Aligning them changes user-visible `_score` values on the local path and requires regenerating the entire "Test Models, AI & Search" snapshot suite against live embedding models, which can't be done correctly in an automated PR. Full analysis is posted on #11263. This PR keeps the existing `[0, 1]` normalization and only removes the `NaN`.

## Test plan

- `make lint`
- `make test`

New unit tests in `cosine_distance.rs`:
- `test_cosine_distance` — identical / opposite / arbitrary vectors stay in `[0, 1]` (converted to `matches!` guards so no `float_cmp`).
- `test_cosine_distance_zero_vector_is_null` — a zero vector on either side, and a zero/zero pair, all return `None`.
- `test_compute_cosine_distance_zero_vector_propagates_null` — the production wrapper surfaces `Ok(None)` for a zero vector and `Ok(Some(finite))` for a normal pair.

## Checklist

- [x] Distances for non-zero vectors unchanged (no snapshot drift)
- [x] Regression tests exercise the production path

Addresses the zero-vector NaN issue in #11263 (the issue stays open for the formula-divergence decision).

### DIFF
--- a/crates/runtime-datafusion-udfs/src/cosine_distance.rs
+++ b/crates/runtime-datafusion-udfs/src/cosine_distance.rs
@@ -196,10 +196,16 @@ fn compute_cosine_distance(
         return exec_err!("Both arrays must have the same length");
     }
 
-    Ok(Some(cosine_distance(&float_vals1, &float_vals2)))
+    Ok(cosine_distance(&float_vals1, &float_vals2))
 }
 
-fn cosine_distance(x: &Float64Array, y: &Float64Array) -> f64 {
+/// Computes the cosine distance between two equal-length vectors.
+///
+/// Returns `None` when either vector has zero magnitude (e.g. an all-zero or
+/// failed embedding): cosine similarity is undefined there (`0.0 / 0.0` is
+/// `NaN`), and a `NaN` score sorts ahead of every real score in
+/// `ORDER BY _score DESC`, surfacing failed embeddings as top matches.
+fn cosine_distance(x: &Float64Array, y: &Float64Array) -> Option<f64> {
     let mut x_length: f64 = 0.0;
     let mut y_length: f64 = 0.0;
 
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

Issue #11277, opened 2026-06-10
bucket() partition transform hashes with ahash — not stable across CPU architectures or ahash upgrades, but backs persistent partition pruning


claudespice:
Found during a data-correctness audit of trunk.

`crates/runtime-datafusion-udfs/src/bucket.rs:44-45` seeds `RandomState::with_seeds(...)` (ahash) and `:244-260` hashes values with it. ahash with fixed seeds is deterministic per build+platform, but its output differs between CPUs with/without AES acceleration and across ahash versions. `bucket(N, col)` is the partition transform for `runtime-table-partition`: data partitioned by one host/arch and later pruned by a binary on another (or after a silent ahash dependency bump) maps the same value to a different bucket — equality-filter pruning then skips the partition that actually holds the matching rows, silently returning missing rows.

Suggested fix: use a specified, version-pinned, platform-stable hash for the bucket transform (e.g. Murmur3 as in Iceberg's bucket transform, or xxhash with an explicit implementation), and treat the hash function as part of the on-disk format.

PR #11623, opened 2026-07-05
test(bucket): guard frozen partition-hash output against silent ahash drift (addresses #11277)

claudespice:
## Summary (root cause)

`bucket(n, value) = hash(value) % n` is the partition transform behind `runtime-table-partition`. For persisted partitioned datasets, a value **must** map to the same bucket forever, or an equality-filter partition prune skips the partition that actually holds the matching rows — silent missing-row data loss (#11277).

`crates/runtime-datafusion-udfs/src/vendored_hash.rs` was created precisely to freeze this hash to DataFusion 53's `ahash` so a DataFusion upgrade can't re-bucket existing data. But that module only froze the **seed and the hashing loop** — it still delegates the actual byte hashing to the external `ahash` crate, which is pinned only as `^0.8` (`Cargo.toml`). ahash gives **no** cross-version output guarantee, so a routine `cargo update` within `0.8.x` (or a build that enables ahash's AES path) can silently change the hash and re-bucket every existing partitioned dataset.

The existing golden coverage only locked the **string** (`bucket_scalar`/`bucket_array`) and **decimal** paths. The integer, unsigned, boolean, float, and binary paths — the common partition keys, e.g. the issue's own `bucket(50, org_id)` / `bucket(3, user_id)` — had **no** guard, so drift there would be completely silent.

Note the issue's suggested fix (switch to Murmur3) is intentionally **not** taken: changing the hash would itself re-bucket all existing persisted data — the exact regression `vendored_hash.rs` exists to prevent. The correct non-breaking fix is to make the freeze enforceable.

## Changes

- `crates/runtime-datafusion-udfs/src/bucket.rs`: add `test_bucket_hash_stability_golden_values`, a golden-value regression test pinning `bucket()`'s output for the full supported type matrix (Int8/16/32/64, UInt8/16/32/64, Boolean, Float32/64, Utf8, Binary) plus the vectorized `compute_bucket_array` path for an Int64 column. It exercises the production `compute_bucket`/`compute_bucket_array` functions with a large modulus (`MAX_NUM_BUCKETS`) so the assertion captures the low 6 decimal digits of the hash, and edge values that fill each integer width.
- `crates/runtime-datafusion-udfs/src/vendored_hash.rs`: document the residual ahash-crate dependency and point to the new guard.

No behavior change: the goldens capture current output, so existing partitioned data is untouched. Any future drift (dependency bump or AES-path build) now fails CI loudly instead of silently corrupting partition pruning.

## Scope — what this does and does NOT resolve

- **ahash version/algorithm bump** (a `cargo update` within `^0.8`): fully guarded — the goldens turn it into a loud CI failure.
- **Cross-architecture portability** (a build with `+aes` / `target-cpu=native` on a different CPU): only guarded on the arch CI runs on, not at runtime. Fully closing that needs forcing ahash's platform-stable fallback or a format-versioned hash migration — a maintainer / on-disk-format decision.

So this is a non-breaking mitigation that closes the realistic silent-drift vector; #11277 stays **open** for the deeper portability/format decision.

## Performance impact

None — test and documentation only; no production code paths change.

## Test plan

- New `cargo test -p runtime-datafusion-udfs --lib bucket` — all 13 bucket tests pass (including the new golden test).
- `cargo clippy -p runtime-datafusion-udfs --all-targets --no-deps` — clean (the one warning is a pre-existing unfulfilled lint-expectation in the unrelated `ai_concurrency` integration test).
- `cargo fmt` (toolchain 1.95.0) — clean.
- Full-workspace `make lint` / `make test` run in CI (this change is test + doc only, in a single crate).

## Checklist

- [x] Fix is minimal and in scope (test + doc only)
- [x] No hash output changes — existing persisted partitioned data is untouched
- [x] Regression test exercises the production `compute_bucket`/`compute_bucket_array` path
- [x] Covers the previously-unguarded integer/unsigned/bool/float/binary paths
- [x] No `.unwrap()` (uses `.expect(...)`)

Addresses #11277

### DIFF
--- a/crates/runtime-datafusion-udfs/src/bucket.rs
+++ b/crates/runtime-datafusion-udfs/src/bucket.rs
@@ -563,7 +563,7 @@ mod tests {
     /// lifetime of a persisted dataset, or an equality-filter prune silently
     /// skips the partition that holds the matching rows — missing-row data loss
     /// (#11277). The hash is therefore part of the on-disk format, and
-    /// `vendored_hash.rs` freezes it to DataFusion 53's `ahash`. But that module
+    /// `vendored_hash.rs` freezes it to `DataFusion` 53's `ahash`. But that module
     /// still delegates the actual hashing to the external `ahash` crate, pinned
     /// only as `^0.8` — and ahash gives **no** cross-version output guarantee, so
     /// a routine `cargo update` (or a build that enables ahash's AES path) can
--- a/crates/runtime-datafusion-udfs/src/bucket.rs
+++ b/crates/runtime-datafusion-udfs/src/bucket.rs
@@ -555,6 +555,103 @@ mod tests {
         );
     }
 
+    /// Golden-value guard that pins `bucket()`'s output for a representative
+    /// value of every non-string Arrow type the partition transform hashes.
+    ///
+    /// `bucket(n, value) = hash(value) % n` is the partition transform behind
+    /// `runtime-table-partition`: a value MUST map to the same bucket for the
+    /// lifetime of a persisted dataset, or an equality-filter prune silently
+    /// skips the partition that holds the matching rows — missing-row data loss
+    /// (#11277). The hash is therefore part of the on-disk format, and
+    /// `vendored_hash.rs` freezes it to DataFusion 53's `ahash`. But that module
+    /// still delegates the actual hashing to the external `ahash` crate, pinned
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

claudespice:
@copilot review

lukekim (approved):

PR #11624, opened 2026-07-05
fix(imap): store message date as millisecond Timestamp, not seconds in Date64 (fixes #11547)

### LINKED ISSUE #11547: IMAP connector: date column is corrupted for every row (seconds/milliseconds unit mismatch)
## Describe the bug

The IMAP connector's `date` column is wrong for every single row — regardless of when a message was actually sent, `date` always evaluates to `1970-01-21T00:00:00`.

Root cause (from `crates/data_components/src/imap/provider.rs`): the connector parses the email's `Date:` header via `mailparse::dateparse()`, which returns Unix **seconds**, then pushes that value directly into an Arrow `Date64Array` — a type whose value is defined to be **milliseconds** since epoch. The missing `× 1000` conversion means every timestamp is 1000x too small.

This is compounded by `Date64` being a day-precision type: consumers compute `calendar_day = value / 86_400_000`. Since a normal mailbox's date range spans a few million seconds (days/weeks), dividing by 86,400,000 (ms/day) collapses the *entire* mailbox onto the same single calendar day, near epoch. It's not engine-specific — this happens in the in-memory Arrow `RecordBatch` before any acceleration engine (DuckDB, SQLite, Postgres, etc.) is involved, so no downstream storage choice can work around it.

## To Reproduce

1. Configure a spicepod with an `imap:` dataset, acceleration enabled.
2. Run `SELECT min(date), max(date), count(distinct date) FROM <dataset>` after sync.
3. Every row returns the same date, always around `1970-01-21`.

## Expected behavior

`date` should reflect each message's actual send date (from the `Date:` header), with distinct values across the mailbox.

### Spicepod

```yaml
datasets:
  - from: imap:user@example.com
    name: mailbox
    params:
      imap_password: ${secrets:IMAP_PASSWORD}
      imap_host: imap.gmail.com
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
```

```
date | Date64 | NO
```

### spice, spiced, OS info

- spice version: v1.11.5 (1d69940)
- spiced version: v2.0.1+models.metal
- OS info: Darwin arm64, macOS (Darwin 25.5.0)

- [ ] Yes
- [x] No

## Additional context

A correct fix needs both: (1) multiply the parsed seconds value by 1000 before constructing the array, and (2) consider using Arrow `Timestamp` instead of `Date64` if preserving time-of-day (not just calendar day) is desired — `Date64`'s day-precision means even a fixed version would lose time-of-day.

claudespice:
## Summary (root cause)

The IMAP connector parsed each message's `Date:` header with `mailparse::dateparse`, which returns whole **seconds** since the Unix epoch, and pushed that value directly into an Arrow `Date64Array` — a type whose value is **milliseconds** since the epoch. The missing `× 1000` made every timestamp 1000x too small.

This is compounded by `Date64` being a day-precision type: consumers compute `calendar_day = value / 86_400_000`. A normal mailbox spans only a few million seconds, so dividing the (already-too-small) values by ms-per-day collapsed the *entire* mailbox onto a single near-epoch calendar day (`~1970-01-21`). The corruption happens in the in-memory `RecordBatch` before any acceleration engine, so no storage choice worked around it.

Fixes #11547

## Changes

- Add `parse_date_millis`, which scales `dateparse`'s seconds to milliseconds (`saturating_mul(1000)`), and use it in `scan`.
- Change the `date` column type from `Date64` to `Timestamp(Millisecond, None)`. `Date64` is a day-precision type whose values are meant to be day-aligned; storing an arbitrary instant in it is semantically invalid and floors away the time-of-day. `Timestamp(Millisecond)` correctly preserves the message's send instant, addressing point (2) the issue calls out.
- Build the column with `TimestampMillisecondArray`.

## Test plan

- `make lint` (Rust fmt + clippy, incl. `-Dclippy::pedantic -Dclippy::unwrap_used -Dclippy::expect_used`)
- `make test`

New unit tests in `crates/data_components/src/imap/provider.rs` (run locally with `--features imap`):
- `parse_date_millis_scales_seconds_to_milliseconds` — a known `Date:` header must parse to exactly 1000x the raw seconds (regression guard for the unit mismatch).
- `schema_date_column_is_timestamp_millis` — the `date` field is `Timestamp(Millisecond, None)`.
- `build_recordbatch_preserves_distinct_send_times` — two messages a day apart stay distinct and exact in the built `TimestampMillisecondArray` (the pre-fix code collapsed them).

## Checklist

- [x] Root cause identified and fixed at the source (no downstream workaround)
- [x] No behavior change outside the IMAP `date` column

### DIFF
--- a/crates/data_components/src/imap/mod.rs
+++ b/crates/data_components/src/imap/mod.rs
@@ -18,7 +18,8 @@ use std::sync::Arc;
 
 use arrow::{
     array::{
-        ArrayRef, Date64Array, ListArray, ListBuilder, RecordBatch, StringArray, StringBuilder,
+        ArrayRef, ListArray, ListBuilder, RecordBatch, StringArray, StringBuilder,
+        TimestampMillisecondArray,
     },
     error::ArrowError,
 };
@@ -110,7 +111,7 @@ impl ImapTableProvider {
         }
 
         let mut fields: Vec<ArrayRef> = vec![
-            Arc::new(Date64Array::from(dates)),
+            Arc::new(TimestampMillisecondArray::from(dates)),
             Arc::new(StringArray::from(subjects)),
             Arc::new(build_listarray_for_strings(froms)),
             Arc::new(build_listarray_for_strings(tos)),
@@ -136,6 +137,7 @@ impl From<Error> for DataFusionError {
 }
 
 pub(crate) struct EmailMessage {
+    /// Send time as milliseconds since the Unix epoch (see `parse_date_millis`).
     date: i64,
     subject: Option<String>,
     from: Option<Vec<Option<String>>>,
--- a/crates/data_components/src/imap/provider.rs
+++ b/crates/data_components/src/imap/provider.rs
@@ -16,7 +16,7 @@ limitations under the License.
 
 use std::sync::Arc;
 
-use arrow::datatypes::{DataType, Field, Schema, SchemaRef};
+use arrow::datatypes::{DataType, Field, Schema, SchemaRef, TimeUnit};
 use async_trait::async_trait;
 use datafusion::{
     catalog::{Session, TableProvider},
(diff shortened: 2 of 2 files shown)


claudespice:
@copilot review

claudespice:
@copilot review

lukekim (approved):

PR #11625, opened 2026-07-05
fix: report empty accelerated schema instead of misleading missing-column error (fixes #10920)

### LINKED ISSUE #10920: Bug. Invalid error `Primary key column was not found` when tale does not exist
## Describe the bug

When target dataset does not exit error refer to invalid Primary key  instead of invalid table that does NOT exit. Postgres with`refresh_mode: changes` but might be general

>2026-05-19T08:05:31.863202Z  INFO runtime::init::dataset: Dataset my_table initializing...
2026-05-19T08:05:31.877873Z  WARN runtime::init::dataset: Failed to setup the dataset my_table (postgres). Unable to create dataset acceleration: Invalid configuration: Primary key column 'marker_id' was not found in the schema. Valid columns: . Verify configuration and try again. For details, visit https://spiceai.org/docs/features/data-acceleration/constraints

## To Reproduce

```yaml
- from: postgres:my_table_not_exist
    name: my_table
    params:
      <<: *pg_base
      pg_replication_slot: my_table_not_exist
    acceleration:
      enabled: true
      engine: cayenne
      mode: file
      refresh_mode: changes
      primary_key: marker_id
```

claudespice:
## Summary

When an accelerated dataset's source table does not exist (or otherwise resolves to no columns), the schema comes back empty and the acceleration constraint validation reported a misleading error:

> Primary key column 'marker_id' was not found in the schema. Valid columns: .

The empty `Valid columns:` list was the real tell — the schema had zero columns because the source table could not be read — but the message pointed the user at their `primary_key`/`indexes` config instead of the missing table.

## Root cause

`Acceleration::validate_primary_key` / `validate_indexes` check each configured column against the schema with `field_with_name`. Against an empty schema every column "isn't found", so the first configured constraint fails with `PrimaryKeyColumnNotFound` / `IndexColumnNotFound` and lists no valid columns.

## Changes

- Add a dedicated `AcceleratedSchemaEmpty` error that names the likely root cause (source table does not exist / could not be read).
- `validate_primary_key` and `validate_indexes` now short-circuit to that error when a constraint is configured but the schema has no columns, instead of the misleading column-not-found message.
- Behavior is otherwise unchanged: an empty schema with no constraints still passes, and a populated schema missing the configured column still returns the existing, more specific `PrimaryKeyColumnNotFound` / `IndexColumnNotFound` error.

## Test plan

- `make lint`
- `make test`

New unit tests in `constraints.rs` cover: empty schema + primary key -> `AcceleratedSchemaEmpty` (and asserts the old misleading text is gone), empty schema + index -> `AcceleratedSchemaEmpty`, empty schema with no constraints -> Ok, populated schema missing the PK column -> `PrimaryKeyColumnNotFound` (preserved), and populated schema with the PK column -> Ok.

## Checklist

- [x] Root cause identified and fixed at the validation layer
- [x] Regression tests fail on old code, pass on new
- [x] No behavior change for valid configurations

Fixes #10920

### DIFF
--- a/crates/runtime/src/component/dataset/acceleration/constraints.rs
+++ b/crates/runtime/src/component/dataset/acceleration/constraints.rs
@@ -43,6 +43,9 @@ impl Acceleration {
     }
 
     pub fn validate_indexes(&self, schema: &SchemaRef) -> dataset::Result<()> {
+        if !self.indexes.is_empty() {
+            Self::ensure_schema_populated(schema, "indexes")?;
+        }
         for column in self.indexes.keys() {
             for index_column in column.iter() {
                 if schema.field_with_name(index_column).is_err() {
@@ -60,6 +63,7 @@ impl Acceleration {
 
     pub fn validate_primary_key(&self, schema: &SchemaRef) -> dataset::Result<()> {
         if let Some(columns) = &self.primary_key {
+            Self::ensure_schema_populated(schema, "a primary key")?;
             for column in columns.iter() {
                 if schema.field_with_name(column).is_err() {
                     return dataset::PrimaryKeyColumnNotFoundSnafu {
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

claudespice:
@copilot review

lukekim (approved):

PR #11626, opened 2026-07-05
fix(refresh): apply refresh_sql override on manual refresh when source unchanged (fixes #11353)

### LINKED ISSUE #11353: `refresh_sql` override ignored when triggering manual refresh with s3
###  Reproduce:
  1. Configure s3 csv dataset with `refresh_mode: full` and acceleration enabled
  2. Start spice and verify all rows are present
  2. Trigger manual refresh with a filter:
```bash
curl -X POST http://localhost:8090/v1/datasets/orders/acceleration/refresh \
    -H 'Content-Type: application/json' \
    -d '{"refresh_sql": "SELECT * FROM orders WHERE id <= 3"}'
```
  3. Query the table: `SELECT count(*) FROM orders;`

**Expected** : 3 rows (filter applied)
**Actual**: 7 rows (filter ignored, full dataset retained)

**Example dataset** :
```csv
id,value
1,alpha
2,beta
3,gamma
4,delta
5,epsilon
```

claudespice:
## Summary (root cause)

A manual acceleration refresh that carries a one-off `refresh_sql` override was silently ignored for object-store-backed datasets, leaving the previously-materialized data in place.

```
curl -X POST .../v1/datasets/orders/acceleration/refresh \
  -d '{"refresh_sql": "SELECT * FROM orders WHERE id <= 3"}'
```

Root cause: `RefreshTask::run_once` runs a refresh-skip optimization for `Full`/`Append` modes — `should_skip_refresh_for_table_provider` asks the provider (e.g. `S3SingleFileCached`) whether the *source file* changed (size / `last_modified` / `ETag` / version). When the file is unchanged it returns `Ok(Some(true))` and the refresh is skipped entirely.

That optimization assumes the currently-materialized data is a faithful, complete read of the source. A one-off override `refresh_sql` deliberately materializes a **subset** of the source even though the source itself is unchanged, so the assumption is false: the skip fires, the override is dropped, and the full dataset is retained (issue #11353).

## Changes

- `RefreshTask::run_once` (`crates/runtime/src/accelerated_table/refresh_task.rs`): when a refresh carries an override (`refresh.override_sql_raw.is_some()`), bypass the skip check. Instead, **reset** the provider's cached version so the *next* plain refresh re-materializes the full source rather than skipping against the now-narrowed accelerated data.
- `RefreshSkipTableProvider` (`crates/data_components/src/refresh_skip.rs`): add a `reset_skip_state` method (default no-op) and a symmetric `reset_refresh_skip_state_for_table_provider` helper that peels the same wrapper providers (`MetadataEnrichedTableProvider`, `FederatedTableProviderAdaptor`) as the skip check.
- `S3SingleFileCached` (`crates/data_components/src/s3_single_file_cached.rs`): implement `reset_skip_state` to drop the cached file metadata.

Without the reset half, a plain refresh following an override refresh would compare the unchanged source against the cached version, skip, and leave the narrowed data in place — trading the reported bug for a subtler stale-data one. The reset closes that gap.

Scope: behavior only changes when an override is present; ordinary periodic/manual refreshes take the exact same skip path as before.

Fixes #11353

## Test plan
- [x] Added regression test `test_reset_skip_state_forces_next_refresh` (unchanged source is skippable → reset → next refresh re-materializes → skippable again)
- [x] Added regression test `test_reset_skip_state_through_metadata_enriched_wrapper` (reset helper peels the metadata-enriched wrapper to reach the inner provider)
- [x] `make test` passes

### DIFF
--- a/crates/data_components/src/refresh_skip.rs
+++ b/crates/data_components/src/refresh_skip.rs
@@ -38,6 +38,17 @@ pub trait RefreshSkipTableProvider: TableProvider {
     /// and proceed with the refresh to ensure data consistency.
     async fn should_skip_refresh(&self) -> datafusion::error::Result<bool>;
 
+    /// Invalidates any cached state used to decide whether a refresh can be skipped.
+    ///
+    /// The skip decision assumes the currently-materialized data is a faithful, complete read
+    /// of the source. A one-off refresh override (e.g. a manual refresh with a different
+    /// `refresh_sql`) breaks that assumption: it materializes a *subset* of the source even
+    /// though the source itself is unchanged. After such a refresh the cached version must be
+    /// reset, otherwise the next plain refresh would compare an unchanged source against the
+    /// cached version, skip, and leave the narrowed data in place. Providers cache their
+    /// skip state so this defaults to a no-op only for providers that keep none.
+    async fn reset_skip_state(&self) {}
+
     /// Returns a reference to self as a `RefreshSkipTableProvider` trait object.
     /// This enables dynamic dispatch without requiring knowledge of the concrete type.
     fn as_refresh_skip_provider(&self) -> &dyn RefreshSkipTableProvider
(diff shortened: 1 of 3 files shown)


claudespice:
@copilot review

claudespice:
@copilot review

lukekim (approved):

PR #11635, opened 2026-07-05
fix(ducklake): add opt-in automatic_migration to attach older catalogs (fixes #10899)

### LINKED ISSUE #10899: DuckLake: `the extension requires version 1.0` when connecting to older DuckLake versions
## Describe the bug

When connecting to DuckLake versions created with older DuckDB (DuckLake spec) there is the following error
>Invalid Input Error:
DuckLake catalog version mismatch: catalog version is 0.3, but the extension requires version 1.0.
To automatically migrate, set AUTOMATIC_MIGRATION to TRUE when attaching.

We might need to add an option for `AUTOMATIC_MIGRATION`
https://ducklake.select/docs/stable/duckdb/guides/troubleshooting#connecting-to-an-older-ducklake

Why `AUTOMATIC_MIGRATION` is disabled by default:
https://github.com/duckdb/ducklake/issues/457

claudespice:
## Summary

Connecting to a DuckLake catalog created by an older DuckLake/DuckDB version fails at `ATTACH` with:

> Invalid Input Error: DuckLake catalog version mismatch: catalog version is 0.3, but the extension requires version 1.0. To automatically migrate, set AUTOMATIC_MIGRATION to TRUE when attaching.

Spice built the attach statement as `ATTACH 'ducklake:<cs>' AS "<name>"` with no options, so there was no way to opt into the in-place migration DuckLake supports. This adds an opt-in `automatic_migration` parameter (default `false`) to both DuckLake surfaces — the data connector and the catalog connector — which appends the `AUTOMATIC_MIGRATION TRUE` attach option when enabled.

Migration is left disabled by default because it rewrites the catalog's metadata and cannot be undone (see duckdb/ducklake#457); enabling it is an explicit, informed choice.

Fixes #10899

## Changes

- `crates/data_components/src/ducklake/mod.rs`: new shared, unit-tested `build_ducklake_attach_sql(connection_string, catalog_name, automatic_migration)` helper that centralizes the `ATTACH` construction (single-quote / double-quote escaping preserved) and appends `(AUTOMATIC_MIGRATION TRUE)` only when opted in.
- `crates/data-connectors/connector-ducklake/src/lib.rs`: new `automatic_migration` component parameter; read it (`true`, case-insensitive) and thread it into `create_ducklake_factory`, which now calls the shared helper.
- `crates/runtime/src/catalogconnector/ducklake.rs`: same new parameter and helper call for the catalog connector, so both DuckLake entry points behave identically (bug-class fix, not just one surface).

Behavior is default-preserving: with the parameter unset the emitted `ATTACH` is byte-for-byte identical to before.

## Test plan

- `make lint`
- `make test`
- Added unit tests (`data_components` `ducklake::tests`, gated behind the `duckdb` feature): attach SQL is unchanged when migration is off, appends `(AUTOMATIC_MIGRATION TRUE)` when on, and escaping of the connection string / catalog name is preserved with the option present.

## Checklist

- [x] Opt-in, default-preserving (no behavior change unless `automatic_migration: true`)
- [x] Both DuckLake surfaces (data connector + catalog connector) fixed via one shared helper
- [x] Unit tests for the new SQL construction
- [x] Generated `ATTACH` option syntax verified against DuckLake docs (`AUTOMATIC_MIGRATION` boolean, default false)

### DIFF
--- a/crates/data-connectors/connector-ducklake/src/lib.rs
+++ b/crates/data-connectors/connector-ducklake/src/lib.rs
@@ -20,8 +20,8 @@ limitations under the License.
 
 use async_trait::async_trait;
 use data_components::Read;
-use data_components::ducklake::DuckLakeS3Params;
 use data_components::ducklake::writer::DuckDbFederatedTableWriter;
+use data_components::ducklake::{DuckLakeS3Params, build_ducklake_attach_sql};
 use datafusion::datasource::TableProvider;
 use datafusion::sql::TableReference;
 use datafusion_table_providers::UnsupportedTypeAction;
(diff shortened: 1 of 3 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11636, opened 2026-07-05
ci(e2e): tolerate empty-type DuckDB variant of the parquet-rename race

claudespice:
## Problem

The `graceful_shutdown_test*` acceleration jobs in `e2e_test_ci.yml` run `simulate_append_data.sh`, which rewrites `.spice/service_data.parquet` every 3s. A refresh tick that races the rename reads a **half-published Parquet file**, which the `data_drop` dataset is deliberately set up to hit (its `on_conflict: drop` refresh is expected to fail-and-retry). The log-scan step already whitelists this documented transient — but only for the `TProtocolException: Invalid data` error DuckDB emits when it reaches a corrupt Thrift footer.

DuckDB surfaces the *same* truncated read a second way — `Invalid Error: don't know what type: ` with an **empty type** (a schema element that has not been written yet) — and that variant was never whitelisted. So whenever the race lands on the empty-type path, the job fails even though nothing is wrong. This intermittently reddens the `duckdb_append_with_pk.yaml` job (and its siblings) on PRs; the same job passes on `trunk` when the race happens to produce the whitelisted variant.

Example: https://github.com/spiceai/spiceai/actions/runs/28750176806/job/85251613356

## Fix

Extend both log-scan whitelists to also tolerate the empty-type variant of the same race. The added pattern is anchored to an empty type (`^Invalid Error: don't know what type: *$`), so a **genuine** unsupported-type error — which always names the offending type after the colon — still fails the job and is not masked.

## Validation

- No Rust code changed — this is a CI-workflow-only change, so the Rust `make lint` / `make test` gates do not apply.
- `.github/workflows/e2e_test_ci.yml` parses as valid YAML.
- Verified the grep pipeline against a synthetic `spice.log`: a log containing only the WARN + both transient variants produces no output (job passes), while adding a genuine `ERROR something genuinely broke` line still surfaces it (job fails). The empty-type anchor does not swallow real errors.

### DIFF
--- a/.github/workflows/e2e_test_ci.yml
+++ b/.github/workflows/e2e_test_ci.yml
@@ -2008,10 +2008,15 @@ jobs:
           # `simulate_append_data.sh` rewrites `.spice/service_data.parquet`
           # every 3s; a refresh tick that races the rename observes a
           # half-published file and the runtime logs a WARN before
-          # retrying on the next tick. Tolerate that specific transient.
+          # retrying on the next tick. DuckDB surfaces the truncated read as
+          # one of two errors depending on which part of the file it reaches
+          # first: "TProtocolException: Invalid data" (corrupt Thrift footer)
+          # or "don't know what type:" with an empty type (schema element not
+          # yet written). Tolerate both variants of that specific transient.
           if grep -iE "(failed|error)" spice.log \
              | grep -vE "WARN +runtime::accelerated_table::refresh_task: Failed to load data for dataset" \
-             | grep -vE "^Invalid Error: TProtocolException: Invalid data$"; then
+             | grep -vE "^Invalid Error: TProtocolException: Invalid data$" \
+             | grep -vE "^Invalid Error: don't know what type: *$"; then
             echo "Failures detected in spice.log"
             exit 1
           fi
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

claudespice:
Heads up: this whitelist can't actually suppress the race's WARN, because it filters plain text but `spice.log` always contains ANSI color codes.

`spiced` sets `.with_ansi(true)` unconditionally (`bin/spiced/src/tracing.rs:173,194`), so even though the e2e job runs `spice run &> spice.log &` (a redirect, not a TTY), the log is colored. The WARN header line is therefore:

```
\e[2m<ts>\e[0m \e[33m WARN\e[0m \e[2mruntime::accelerated_table::refresh_task\e[0m\e[2m:\e[0m Failed to load data for dataset ...
```

`grep -vE "WARN +runtime::accelerated_table::refresh_task: Failed to load data for dataset"` never matches that (there's `\e[0m \e[2m` between `WARN` and `runtime`, and `\e[0m\e[2m:\e[0m` instead of a bare `:`). The *continuation* lines (`Invalid Error: TProtocolException: Invalid data` / `don't know what type:`) have no ANSI, so those `-vE` filters do work — but the colored WARN header always survives the pipe, so `Failures detected` fires whenever the race emits a WARN at all. The test only passes when the race happens to emit no WARN, which is why it flakes.

Evidence: on #11646, run 28756534868 job 85271459314 failed with the already-whitelisted `TProtocolException` variant — the surviving grep line was the ANSI WARN header, not the continuation. So adding more continuation-line variants doesn't stop the flake.

Suggested fix — strip ANSI before the grep chain so the existing whitelist patterns match:

```bash
if sed -E 's/\x1b\[[0-9;]*m//g' spice.log | grep -iE "(failed|error)" \
   | grep -vE "WARN +runtime::accelerated_table::refresh_task: Failed to load data for dataset" \
   | grep -vE "^Invalid Error: TProtocolException: Invalid data$" \
   | grep -vE "^Invalid Error: don't know what type: *$"; then
  echo "Failures detected in spice.log"
  exit 1
fi
```

(GNU sed on the ubuntu runners supports `\x1b`.)

PR #11642, opened 2026-07-05
fix(cayenne): durable CDC delete keeps the pk IN (...) fast path (fixes #11633)

### LINKED ISSUE #11633: perf(cayenne): HTAP sf1000 data-freshness p99 regressed ~67%, bisected to #11514
The daily HTAP CH-benCHmark shows a sustained worst data-freshness p99 regression on `accelerated/postgres-cayenne[file]-cdc-tuned-sf1000.yaml` at SF1000.

## Summary

- **Baseline:** ~337k ms (`a0ccb8ead6`, [run 28445418795](https://github.com/spiceai/spiceai/actions/runs/28445418795)) — re-confirmed by a fresh run at `cf9b33e7` = 337,382 ms ([run 28741566676](https://github.com/spiceai/spiceai/actions/runs/28741566676))
- **After:** ~490–582k ms, sustained across 7+ consecutive trunk runs — **~+67%**
- **Bisected to** `084d5ae6` = **#11514** (`fix(cayenne): user-visible DELETE WHERE pk IN (...) reports the real row count`)
- **Metric:** worst data-freshness p99 = source-write → queryable-in-accelerator lag

## Bisect trail

Each row is a dispatch of the exact sf1000 config at the pinned `spiced_commit`:

| commit | PR | freshness p99 (ms) | run | verdict |
|---|---|---|---|---|
| `a0ccb8ead6` | last good (from series) | 337,026 | [28445418795](https://github.com/spiceai/spiceai/actions/runs/28445418795) | good |
| `cf9b33e7` | #11515 | 337,382 | [28741566676](https://github.com/spiceai/spiceai/actions/runs/28741566676) | good |
| **`084d5ae6`** | **#11514** | **609,342** | [28733460243](https://github.com/spiceai/spiceai/actions/runs/28733460243) | **bad ← culprit** |
| `7faaf68b` | #11516 | 544,329 | [28723217062](https://github.com/spiceai/spiceai/actions/runs/28723217062) | bad |
| `8f7d9a50d` | #11498 | 561,694 | [28486092925](https://github.com/spiceai/spiceai/actions/runs/28486092925) | bad |

`cf9b33e7` (#11515) is the immediate parent of `084d5ae6` (#11514), so the good→bad boundary is exactly #11514.

## Likely mechanism (please verify)

#11514 makes the user-visible DELETE paths (`build_deletion_vector_sink`, backing `delete_from` and `delete_using_deletion_vectors`) `with_exact_count()`, which bypasses the `pk IN (...)` fast path in favor of a full **scan-to-count**. `delete_using_deletion_vectors` holds the table `write_lock` across that scan, so a now-scanning user DELETE may serialize against the CDC apply loop and inflate the freshness lag this metric measures. The PR intends CDC to keep the fast path — it would help to confirm whether the benchmark's deletes go through the user-visible path vs CDC.

## Reproduce

```bash
gh workflow run testoperator_run_htap.yml --repo spiceai/spiceai --ref trunk \
  -f 'spicepod_path=accelerated/postgres-cayenne[file]-cdc-tuned-sf1000.yaml' \
  -f scale_factor=1000 -f terminals=100 -f rate=10000 -f duration=600 \
  -f ready_wait=900 -f runner_type=spiceai-dev-xlarge-runners -f spiced_commit=<sha>
```

Compare `084d5ae6` (#11514, bad) vs its parent `cf9b33e7` (#11515, good). Note: the sf100/sf1000 runs share one concurrency group (`cancel-in-progress: false`), which evicts a pending dispatch when a newer run queues — dispatch into an idle window so the run isn't cancelled before it starts.

lukekim:
## Problem

The daily HTAP CH-benCHmark showed a sustained worst data-freshness p99 regression on the sf1000 file-durability config (`postgres-cayenne[file]-cdc-tuned-sf1000.yaml`): ~337k ms → ~490–609k ms (**~+67%**), sustained across 7+ trunk runs, [bisected to #11514](https://github.com/spiceai/spiceai/issues/11633).

## Root cause

User-visible `DELETE` and the **durable CDC apply loop** both funnel through the same `TableProvider::delete_from`. #11514 marked that path's sink `with_exact_count()` to fix a user-facing "0 rows affected" bug — correct for user DELETEs, which surface the count.

But the same change forced **durable CDC deletes** off #11049's count-skipping `pk IN (...)` fast path onto a full **scan-to-count**. On the delete-receiving table (TPC-C `new_order`), every CDC delete batch went from **O(deleted keys), zero-scan** → **O(entire table)**, held under the table `write_lock` and serialized against CDC upserts / compaction / checkpoints — the freshness regression.

The PR's "CDC keeps the fast path" assumption only held for the in-memory absorption path (`cdc_durability: memory`); the `file` + `key` config takes the durable `delete_from` path.

## Fix

Add `CayenneTableProvider::delete_from_cdc_fast`, which builds the **same** key-based `InlineAwareDeletionSink` as `delete_from` (inline-row deletion + file-resident deletion vectors, same protected-snapshot / cold-tier scan set) but with a **non-exact** inner sink — so a `pk IN (...)` (or composite OR-of-AND) filter persists deletion vectors without scanning.

- The CDC apply loop (`process_delete_batch`) calls it first, behind the existing `cayenne_accelerator()` unwrap. Non-Cayenne accelerators — and shapes it declines (position / retention, which always scanned and were never regressed by #11514) — fall back to the unchanged `delete_from`.
- User `DELETE` still reports the exact "rows affected" count (**#11514 preserved**). `build_deletion_vector_sink` is parameterized with an `exact_count` flag so both paths share one builder.

This restores the pre-#11514 durable CDC delete behavior.

## Test

`int64_pk_deletion_test::test_int64_pk_cdc_fast_delete_keeps_fast_path` — asserts a file-resident CDC `pk IN (...)` delete stays on the fast path (non-authoritative sentinel count, not a scanned count) and removes the right rows; the pre-existing exact-count test still passes.

Re-validated across the **deletion-mode × durability matrix**:

| deletion_mode | durability | coverage |
|---|---|---|
| key | file | `int64_pk_deletion_test` (incl. new test + exact-count), `keybased_deletion_test` |
| position | file | `position_based_deletion_test`, `cdc_compaction_delete_race_test` (both modes) |
| key/position | memory | runtime `refresh_task::changes` routing (`test_memory_cdc_*`) — 65/65 |
| cross-mode | — | `mutation_property_test`, `deletion_test` |

All green; both crates recompile clean.

## Validation of the perf recovery

The definitive check is the issue's benchmark repro (compare this branch vs `cf9b33e7`). Recommend dispatching `testoperator_run_htap.yml` at the sf1000 config before/after to confirm the freshness p99 returns to baseline.


crates/cayenne/src/provider/table.rs
```diff
@@ -22428,10 +22430,17 @@ impl CayenneTableProvider {
...
+    ///
+    /// `exact_count` requires a VERIFIED deleted-row count (a user-visible DELETE,
+    /// which surfaces "rows affected"); `false` keeps #11049's count-skipping
+    /// `pk IN (...)` fast path for callers that discard the count (see
+    /// [`Self::delete_from_cdc_fast`]).
     async fn build_deletion_vector_sink(
         &self,
         filters: &[Expr],
         write_lock: Option<Arc<tokio::sync::Mutex<()>>>,
+        exact_count: bool,
```
phillipleblanc: nit: instead of a boolean, make this an enum so that callsites are explicit:

i.e. from

`.build_deletion_vector_sink(&filters, None, true)`

to (as an example, enum can be tweaked)

`.build_deletion_vector_sink(&filters, None, DeletionSource::User) // or DeletionSource::CDC`

crates/cayenne/src/provider/table.rs
```diff
@@ -22459,12 +22468,55 @@ impl CayenneTableProvider {
...
+    ///
+    /// User `DELETE` and the durable CDC apply loop both funnel through
+    /// [`TableProvider::delete_from`], which requires a verified "rows affected"
+    /// count. CDC discards that count, so the scan-to-count #11514 added for user
+    /// DELETEs is pure overhead here — it turned every durable CDC delete batch
+    /// into a full table scan and regressed SF1000 HTAP data-freshness ~67%
+    /// (issue #11633). The apply loop calls this instead to opt out: it builds the
+    /// same key-based [`InlineAwareDeletionSink`] as `delete_from` but with a
+    /// non-exact inner sink, so a `pk IN (...)` (or composite OR-of-AND) filter
+    /// persists deletion vectors without scanning.
```
phillipleblanc: nit: Omit details about specific GitHub issues or regressions.

phillipleblanc (dismissed):

crates/cayenne/src/provider/table.rs
lukekim: Fixed — replaced the `exact_count` boolean at `build_deletion_vector_sink` call sites with `DeletionRequestSource::{User,Cdc}`.

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -2387,16 +2387,44 @@ impl RefreshTask {
...
+                        false
+                    }
+                }
+                #[cfg(windows)]
+                {
+                    false
+                }
+            };
+
+            if !applied_on_fast_path {
```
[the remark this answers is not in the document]
lukekim: Fixed — renamed the flag to `handled_by_cayenne_cdc_path` and tightened the comment so it describes Cayenne handling rather than guaranteed fast-path execution.

crates/cayenne/src/provider/table.rs
```diff
@@ -22459,12 +22468,55 @@ impl CayenneTableProvider {
...
+    ///
+    /// Returns `Ok(Some(_))` when handled (the count is non-authoritative and must
+    /// be discarded), or `Ok(None)` for shapes this path does not fast-path
+    /// (file-based retention, position-based deletes — never regressed by #11514),
+    /// which the caller must route through `delete_from` unchanged.
```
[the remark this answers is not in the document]
lukekim: Fixed — updated `delete_from_cdc_fast` docs to state the returned count is not guaranteed exact and must be discarded by CDC callers.

crates/cayenne/src/provider/table.rs
lukekim: Fixed — removed issue/regression history from the comments and kept the docs focused on current behavior.

phillipleblanc (dismissed):

sgrebnov (approved):

PR #11646, opened 2026-07-05
feat(cayenne): end-to-end integrity checksums for WAL records and Vortex data files (fixes #11639)

### LINKED ISSUE #11639: Enhancement: end-to-end integrity checksums for Cayenne WAL records and Vortex data files
## Summary

Cayenne has **no content-integrity checks** on two durability surfaces, so a corrupt byte is either replayed as garbage or scanned as silently-wrong rows rather than being detected:

1. **Staging WAL records.** The staging WAL is written as plain JSON (`crates/cayenne/src/provider/staging_wal.rs`, tmp-write + fsync + atomic rename); integrity rides on JSON parse success only. There is no per-record checksum (a search for `checksum`/`crc`/`digest` across `staging_wal.rs`, `partitioned_wal.rs`, and `fsync_tier.rs` is empty). A truncated or corrupt WAL tail after a crash is not *detected and discarded* at the record level.
2. **Vortex data files.** No end-to-end digest is computed at flush or verified on read. Cayenne trusts the object store plus the Vortex footer, so **silent bit-rot in a published data file would scan as silently-wrong rows**. The only read-path schema logic (`read_schema()` / `viewify_read_schema` in `table.rs`) is a planning aid (Utf8→Utf8View mapping), not an integrity check.

## Why it matters

A CDC-fed accelerator is the system of record for the freshest data. Silent corruption there is the worst failure mode — undetectable at query time and unrecoverable after the fact. The lakehouse frontier (Iceberg V3 / Puffin, and modern columnar formats) treats file-level integrity as table stakes; Cayenne should match that floor for its owned Vortex format and its WAL.

## Proposed approach

1. **Per-record WAL checksums.** Add a checksum to each WAL record; on replay, a record failing its checksum is discarded (folding into the recovery-hygiene path in `ensure_no_incomplete_write`) rather than parsed as garbage, and replay is made idempotent so a `kill -9` between move-and-unlink self-heals.
2. **Per-file Vortex digest, verified on read.** Compute a digest at flush, store it alongside the file metadata, and verify it on first read (or via a background scrubber). On mismatch, fail the read as a **detected fault** instead of returning wrong rows. Co-design with the Vortex format roadmap where possible so the digest lives at the format layer.

## Acceptance criteria

- A corrupted/truncated WAL tail is detected and discarded on recovery — regression test flips/truncates bytes and asserts clean convergence to the last durably-acknowledged state.
- A corrupted Vortex data file is detected on read — regression test flips bytes and asserts a detected fault (error), **not** silently-wrong rows.
- No steady-state throughput regression beyond a small, documented band (checksums computed on the flush/append path, verified on the read path).

## References

- Apache Iceberg V3 / Puffin (per-file integrity + deletion vectors); general storage-engine end-to-end checksum practice (verify-on-read / background scrub).

## Related

Companion durability-hardening work: crash-recovery hardening (#11640), and a Deterministic Simulation Testing harness (#11638) — the corruption cases above (WAL-tail truncate, Vortex data-file bit-flip) are exactly the faults a DST fault matrix would inject and validate.

claudespice:
## Summary

Cayenne had **no content-integrity checks** on two durability surfaces, so a corrupt byte was either replayed as garbage (staging WAL) or scanned as silently-wrong rows (Vortex data files) rather than being detected. This adds opt-in, end-to-end integrity checksums to both, reusing existing infrastructure (`hash_index`'s seeded XXH3, the existing manifest table, the existing WAL recovery path).

The feature is gated behind a single **default-off** param, `cayenne_integrity_checksums`. With it off, the on-disk format and read/write paths are byte-identical to before (zero steady-state overhead, and an older binary can still read anything a newer one wrote). Reads always accept both framed and legacy pre-feature records, so toggling the flag — or downgrading — never orphans a WAL or breaks a scan.

Fixes #11639

## What it does

**1. Per-record staging-WAL checksums (`crates/cayenne/src/provider/wal_checksum.rs`)**
- Each staging-WAL record is wrapped in a small binary envelope: `MAGIC | version | payload_len | XXH3-64(payload) | JSON payload`. The JSON payload is stored verbatim, so the record type keeps its existing serde form.
- On recovery, a record that fails its checksum is **detected and discarded** — its staging directory (WAL + orphaned staged files) is removed and recovery converges to the last committed snapshot, instead of parsing corrupted bytes as valid JSON and moving files with wrong instructions. This is safe because the staging WAL is a forward-progress marker, not the durable commit point (the metastore visibility commit is), so a WAL that fails its own integrity check never represented committed data.
- A flipped byte *inside* a JSON string (a staged filename, the target snapshot id) — which previously parsed as valid-but-wrong JSON — is now caught.

**2. Per-file Vortex data-file digests (`crates/cayenne/src/provider/file_digest.rs`)**
- A self-describing digest (`"xxh3-128:<hex>"`) is computed for each published data file and stored in the snapshot manifest (`cayenne_snapshot_file.digest`, a new nullable column).
- Computed at flush wherever the manifest is authored (overwrite / compaction / maintenance rebuild), reusing an already-recorded digest for an immutable file rather than recomputing.
- Verified before a file is first scanned (cached per-file per-process — "verify on first read"). A **mismatch fails the scan as a detected fault** rather than decoding corrupted bytes into silently-wrong rows. A file with no stored digest is skipped (unverifiable, not corrupt); a file the manifest lists but that cannot be read is skipped (not proof of corruption).

## Changes

- `provider/wal_checksum.rs` (new) — framing/verify + unit tests.
- `provider/file_digest.rs` (new) — digest compute/check + unit tests.
- `provider/staging_wal.rs` — write framed WAL when enabled; read auto-detects framed vs legacy; discard-on-corrupt in `ensure_no_incomplete_write`.
- `provider/table.rs` — compute digest at manifest authoring; `verify_data_file_integrity()` called at the top of `scan()`; whole-file read helper.
- `metadata.rs` / `cayenne_catalog.rs` / `metastore.rs` / `metastore/{sqlite,turso}.rs` — `digest` column on `SnapshotFile` + DDL + idempotent `ALTER TABLE` migration + `EXPECTED_TABLES` (kept in column-order sync); the acceleration-snapshot export/import rides on `EXPECTED_TABLES` automatically.
- `provider/context.rs` + `runtime/.../dataaccelerator/cayenne/mod.rs` — the `cayenne_integrity_checksums` param and the per-process verified-file cache.

## Design notes

- **Opt-in, default-off**: zero regression when off; verification runs on the read path only when on, bounded to one whole-file read per file per process. Matches the issue's "small, documented band".
- **WAL discards, data files fault**: a corrupt WAL is an *uncommitted, in-flight* append → safe to discard and converge; a corrupt data file is *already-committed* data → must fault loudly, never silently drop rows.
- **XXH3 vs a crypto hash**: the threat is accidental bit-rot / torn writes, not an adversary, so a fast non-cryptographic checksum (already a Cayenne dependency via `hash_index`) is the right tool — no new crate.

## Scope / follow-ups

- Digest coverage follows manifest authoring (overwrite / compaction / maintenance rebuild). Plain-append files receive a digest at their next manifest author; eager digest computation on the append hot path was intentionally left out to avoid a whole-file read on every append (the same reason the manifest itself is populated lazily). Noted for a follow-up.
- The digest lives in Cayenne's manifest because the Vortex format has no per-file checksum field yet (upstream TODOs). A future format-layer digest can supersede this, per the issue's "co-design with the Vortex format roadmap".

## Test plan

- `make lint`
- `make test` (nextest) — including new coverage:
  - `wal_checksum` unit tests: framing round-trip, flipped-byte/flipped-checksum/truncation/unknown-version detection, legacy pure-JSON passthrough.
  - `file_digest` unit tests: deterministic compute, corruption detection, unknown-algorithm = unsupported (fail-open).
  - `staged_append_test::test_corrupt_framed_wal_discarded_on_recovery` — corrupt WAL tail detected + discarded, converges to the last committed snapshot.
  - `staged_append_test::test_corrupt_data_file_detected_on_read` — bit-flipped Vortex file detected as a fault (error) on read, not silently-wrong rows.

## Checklist

- [x] Opt-in, default-preserving (off = byte-identical, downgrade-safe)
- [x] Reuses existing infra (`hash_index` XXH3, manifest table, WAL recovery path)
- [x] Regression tests for both acceptance criteria
- [x] Metastore column added in DDL + migration + `EXPECTED_TABLES` (both sqlite and turso)


claudespice:
@copilot review

claudespice:
Thanks @copilot — addressed all five in 82c28996b:

1. **Flaky corrupt-WAL test** — `corrupt_framed_wal_bytes` no longer relies on "XXH3 is never 0". It now declares a payload length longer than the bytes present, so `wal_checksum::verify` returns `Truncated` deterministically regardless of the checksum value.
2. **`file_digest::check` on a malformed-but-prefixed digest** — now validates the hex body (exactly 32 lowercase hex chars) and returns `Unsupported` (fail-open) for a garbled stored digest, matching the doc and avoiding a false corruption report. Added a regression test.
3. **Misleading error label** — `read_data_file_bytes` now uses the generic `"read data file for integrity check"` operation string, since it serves both digest computation and verification.
4/5. **Over-promising doc on `verify_data_file_integrity`** — corrected: only the whole-file *read* is cached (once per file per process); each scan still runs one indexed manifest `SELECT`, which the comment now states, with a manifest-epoch cache flagged as a follow-up for very wide tables.

lukekim (approved):

phillipleblanc (approved):

PR #11651, opened 2026-07-05
fix(cayenne): gate metastore reinsert_sequence schema behind a user_version (fixes #11291)

### LINKED ISSUE #11291: Cayenne: gate metadata-only upsert publish (reinsert_sequence) behind a metastore user_version to prevent silent row loss on downgrade/mixed-version
Follow-up explicitly recommended in #11260's own description ("DOWNGRADE is NOT safe... A schema user_version gate is the recommended follow-up") — filing so it's tracked.

Since #11260, upsert commits no longer write per-key `cayenne_insert_record` rows; the per-key reinsert is derived from the `reinsert_sequence` column stamped on key-based delete files. The schema backfill (`crates/cayenne/src/metastore/sqlite.rs` and `turso.rs`, idempotent `ALTER TABLE ... ADD COLUMN reinsert_sequence`) has no `user_version` gate.

An **older binary** (or an older instance in a mixed-version deployment sharing one metastore) opening a catalog written by the new code reads `cayenne_insert_record` — empty for post-upgrade upserts — and ignores `reinsert_sequence`. Every post-upgrade upserted key then loads as a tombstone with `insert_sequence = None`, so the *current* version of each upserted row is silently filtered out of query results: rows vanish with no error.

Suggested fix: bump the metastore `user_version` when the `reinsert_sequence` schema lands and have older readers refuse to open a newer-versioned catalog (loud failure instead of silent wrong results).

claudespice:
## Summary (root cause)

Since #11260, Cayenne upsert commits stop writing per-key `cayenne_insert_record` rows and instead reconstruct re-inserts from a `reinsert_sequence` column stamped on key-based delete files. That schema change is **forward-safe but downgrade-unsafe**: an older binary (or an older instance in a mixed-version deployment sharing one metastore) reads an empty insert-record table for post-upgrade upserts and ignores `reinsert_sequence`, so the current version of every upserted key silently drops out of query results — rows vanish with no error.

The `reinsert_sequence` backfill landed **without any version gate** (both `sqlite.rs` and `turso.rs` carried a `// A user_version gate is the productionization follow-up.` note). This PR adds that gate.

Fixes #11291

## Changes

- New shared constant `CAYENNE_METASTORE_SCHEMA_VERSION` (= `1`) plus a pure `ensure_supported_schema_version()` guard in `crates/cayenne/src/metastore.rs`, with the version history documented (v1 = metadata-only upsert publish, #11260).
- New `CatalogError::IncompatibleSchemaVersion { found, supported }` variant — a loud, actionable, jargon-free message pointing the operator to upgrade Spice or clear the acceleration data.
- Both metastore backends (`sqlite.rs`, `turso.rs`) now, on every `init_schema`:
  1. read `PRAGMA user_version` **before** running any migration and refuse to open a catalog stamped newer than this build supports (returns `IncompatibleSchemaVersion` instead of silently-wrong results);
  2. stamp `PRAGMA user_version = CAYENNE_METASTORE_SCHEMA_VERSION` **after** all migrations succeed.
- Verified `turso_core` 0.6.1 supports `PRAGMA user_version` read + write, so the gate is symmetric across both backends.
- Legacy/fresh catalogs read `0` → always accepted and migrated forward, so no existing deployment is disrupted; the gate only ever refuses a genuinely-newer catalog.
- Replaced the two stale "productionization follow-up" comments.

## Behavior / compatibility

This establishes the version mechanism going forward: from now on, any downgrade to a build whose max supported version is lower than the catalog's stamp fails loudly. (It cannot retroactively protect against *already-released* pre-gate binaries, which never read `user_version` — that is inherent and is exactly what #11291 asked to productionize.)

## Test plan

- `make lint`
- `make test`

New tests:
- `metastore.rs`: `ensure_supported_schema_version` accepts v0/current, rejects newer (with `IncompatibleSchemaVersion` and an actionable message).
- `sqlite.rs` (exercise the real `init_schema` DB path): fresh catalog is stamped + idempotent re-init; legacy `user_version = 0` opens and is migrated forward; a newer stamp is refused loudly and the refused open does **not** mutate the stamp.

## Checklist

- [ ] `make lint` clean
- [ ] `make test` clean

### DIFF
--- a/crates/cayenne/src/catalog.rs
+++ b/crates/cayenne/src/catalog.rs
@@ -119,6 +119,11 @@ pub enum CatalogError {
     SchemaMismatch { table: String },
 
     #[snafu(display(
+        "Cayenne acceleration metadata was written by a newer version of Spice (metadata schema version {found}, this build supports up to {supported}). Opening it with this build could silently drop rows from query results. Upgrade Spice to a build that supports metadata schema version {found} or newer, or clear the Cayenne acceleration data (delete the Cayenne metadata directory) so it can be recreated from the source. See: https://spiceai.org/docs/components/data-accelerators"
+    ))]
+    IncompatibleSchemaVersion { found: i64, supported: i64 },
+
+    #[snafu(display(
         "Deletion vectors require non-negative row IDs, found negative values: {row_ids}"
     ))]
     NegativeRowId { row_ids: String },
--- a/crates/cayenne/src/metastore.rs
+++ b/crates/cayenne/src/metastore.rs
@@ -28,8 +28,53 @@ pub mod turso;
 
 use std::fmt::Display;
 
-use super::catalog::CatalogResult;
+use super::catalog::{CatalogError, CatalogResult};
 use async_trait::async_trait;
+
+/// The Cayenne metastore schema version stamped into the backing database's
+/// `user_version` header by [`Metastore::init_schema`] on every open.
+///
+/// Bump this whenever a schema change makes a catalog written by the new code
+/// **unsafe to open with an older binary** — i.e. the older binary would return
+/// silently wrong results rather than fail cleanly.
+///
+/// Version history:
+/// - `1` — metadata-only upsert publish (#11260): upsert commits stop writing
+///   per-key `cayenne_insert_record` rows and instead stamp `reinsert_sequence`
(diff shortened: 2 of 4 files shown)


claudespice:
@copilot review

lukekim (approved):

phillipleblanc (approved):

PR #11664, opened 2026-07-06
fix: Fairly allocate initial partition shares

peasee:
## 📝 Summary

* Fairly allocates executor shares on initial partition allocation from executor connection, instead of allowing up to `max_partitions_per_executor` at initial allocation. Ensures that, for connected executors, partitions are fairly allocated on connection as well as periodic partition assignment.


crates/runtime/src/cluster/service.rs
```diff
@@ -695,6 +695,16 @@ impl ClusterService for ClusterServiceImpl {
                 |scheduler| scheduler.max_partitions_per_executor,
             );
 
+            // Fair-share divisor: peers register their control stream before calling
+            // this RPC, so the connected count already reflects executors that will
+            // pull shortly. Each table is capped at its fair slice so the first caller
+            // can't greedily claim every partition; the balancer mops up any remainder.
+            let connected_executors = self
```
Jeadie: To clarify, this relies on `allocate_initial_partitions` being called after `control_stream`, otherwise its essentially just `min(ceil(partitions/1), allowed_per_executor) == `min(partitions, allowed_per_executor)` -> first executor gets all partitions (capped by default at 1000).

PR #11665, opened 2026-07-06
fix(bucket): guard against ahash AES-path re-bucketing of partition data (fixes #11277)

### LINKED ISSUE #11277: bucket() partition transform hashes with ahash — not stable across CPU architectures or ahash upgrades, but backs persistent partition pruning

claudespice:
## Summary

`bucket(n, value)` is the partition transform behind `runtime-table-partition`, and its hash is part of the on-disk format: a value must map to the same bucket for the lifetime of a persisted dataset, or an equality-filter prune silently skips the partition holding the matching rows (missing-row data loss).

`vendored_hash.rs` freezes the seed and hashing loop but still delegates the actual byte hashing to the external `ahash` crate. ahash selects between two output-**incompatible** hashers at compile time: `fallback_hash::AHasher` (used by every shipped Spice build) and `aes_hash::AHasher`, which activates on x86/x86_64 when `target_feature = "aes"` is set — e.g. a from-source build with `-C target-cpu=native` on any modern AES-capable CPU. The two produce different `hash_one` output from the same `RandomState` seed, so an AES build re-buckets every persisted partitioned dataset relative to shipped builds and mis-prunes it.

The golden-values test added in #11623 guards against ahash **version** drift, but it runs only on CI's architecture (no `+aes`), so it cannot catch the AES-path divergence — that failure mode was silent.

Root cause: the AES half of #11277 had no guard.

## Changes

- **`crates/runtime-datafusion-udfs/src/vendored_hash.rs`**: added a `compile_error!` that mirrors ahash's exact x86 AES-selection cfg (`any(target_arch = "x86", target_arch = "x86_64"), target_feature = "aes", not(miri)`). Any build that would activate ahash's AES hasher now fails to compile with an actionable message pointing at #11277, instead of silently producing incompatible buckets. Updated the module caveat doc accordingly.
  - This is a **non-migrating** fix: every shipped build already uses the fallback hasher (x86_64 CI sets no `+aes`; the aarch64 `.cargo/config.toml` enables only `neon/fp16/fhm`, and ahash's arm AES paths additionally require its `nightly-arm-aes` feature, which this workspace never enables). Bucket output is unchanged, so no on-disk format migration is needed. Verified empirically: the #11623 goldens were generated on x86_64 CI and `test_bucket_hash_stability_golden_values` passes unchanged on aarch64, confirming x86↔arm fallback stability.
- **`crates/runtime-datafusion-udfs/src/bucket.rs`**: added `test_bucket_uses_portable_fallback_hash_not_aes`, which asserts the guard cfg is inactive in the build under test and re-locks a fallback golden to prove the active hasher is the portable fallback.

This closes the AES half of #11277; together with the golden test (version drift) both silent-drift vectors are now loud. Swapping `bucket()` to a fully ahash-independent, version-pinned hash remains a deliberate, format-versioned migration (maintainer decision) — intentionally out of scope here because it changes on-disk bucket values.

## Performance impact

None. The guard is compile-time only; the hashing code path and its output are unchanged.

## Test plan

- `make lint`
- `make test`
- `cargo build --all-targets`

## Checklist

- [x] Guard mirrors ahash's exact AES cfg (no false positives on safe arm/`+crypto` builds)
- [x] No change to bucket output for shipped builds — no format migration
- [x] Golden stability test still passes (x86 and arm)
- [x] New regression test added

Fixes #11277

### DIFF
--- a/crates/runtime-datafusion-udfs/src/bucket.rs
+++ b/crates/runtime-datafusion-udfs/src/bucket.rs
@@ -657,7 +657,7 @@ mod tests {
     /// output from the same seed, so an AES build would silently re-bucket
     /// persisted partitioned datasets relative to the shipped fallback-hashed
     /// builds (#11277). `vendored_hash.rs` has a `compile_error!` that fails any
-    /// x86/x86_64 build with `target_feature = "aes"` active, so on such a build
+    /// `x86/x86_64` build with `target_feature = "aes"` active, so on such a build
     /// this crate does not compile and this test never runs. Where it *does* run,
     /// assert the invariant explicitly and re-lock one fallback golden to prove
     /// the active hasher is in fact the portable fallback.
--- a/crates/runtime-datafusion-udfs/src/bucket.rs
+++ b/crates/runtime-datafusion-udfs/src/bucket.rs
@@ -652,6 +652,44 @@ mod tests {
         );
     }
 
+    /// The `bucket()` partition hash must never come from ahash's AES-accelerated
+    /// hasher: `aes_hash::AHasher` and `fallback_hash::AHasher` produce different
+    /// output from the same seed, so an AES build would silently re-bucket
+    /// persisted partitioned datasets relative to the shipped fallback-hashed
+    /// builds (#11277). `vendored_hash.rs` has a `compile_error!` that fails any
+    /// x86/x86_64 build with `target_feature = "aes"` active, so on such a build
+    /// this crate does not compile and this test never runs. Where it *does* run,
+    /// assert the invariant explicitly and re-lock one fallback golden to prove
+    /// the active hasher is in fact the portable fallback.
+    #[test]
+    fn test_bucket_uses_portable_fallback_hash_not_aes() {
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11667, opened 2026-07-06
fix(imap): decode text body into content instead of storing raw MIME (fixes #11549)

### LINKED ISSUE #11549: IMAP connector stores raw, undecoded MIME as content instead of decoded body text
**Is your feature request related to a problem? Please describe.**
The `content` column is populated via `BODY.PEEK[]`, which returns the **entire raw RFC822 message** — headers, MIME boundaries, attachments, and any base64/quoted-printable encoded parts — as one opaque string. This makes `content`:
- Bloated: averaged ~320KB/message in our mailbox (vs. plain-text bodies that are typically a few KB), driving up storage and embedding cost/time enormously when used with Spice's embeddings feature.
- Frequently undecoded: many clients send `text/plain` parts as `Content-Transfer-Encoding: base64` or `quoted-printable` — so even after locating the right MIME part, the stored bytes are still encoded gibberish, not readable text.

**Describe the solution you'd like**
Use `BODYSTRUCTURE` to inspect MIME parts before fetching, then selectively fetch and decode only the `text/plain` (falling back to stripped `text/html`) part via the `mailparse` crate the connector already depends on (e.g. `parse_mail()` + its decoding helpers) — never pulling attachment bytes over the wire, and returning already-decoded text in `content`. Attachment metadata (filename, content-type, size) could become a separate structured column instead of inline bytes.

**Describe alternatives you've considered**
A `max_content_size` parameter as a blunter mitigation, and/or a boolean to exclude attachments — neither solves the encoding problem though.

**Additional context**
Found while running the connector against a personal Gmail mailbox with Spice's embeddings feature enabled — raw content caused sustained OpenAI rate-limiting during the embedding backfill (a few thousand messages generated ~87M tokens to embed, mostly attachment/header noise).

claudespice:
## Summary (root cause)

The IMAP connector fetched each message with `BODY.PEEK[]` and pushed the entire raw RFC822 message — MIME headers, multipart boundaries, attachment bytes, and body parts still in their `Content-Transfer-Encoding` — straight into the `content` column via a bare UTF-8/latin1 `decode`. Two consequences (#11549):

- **Undecoded**: many clients send the `text/plain` part as `base64` or `quoted-printable`, so the stored `content` was encoded gibberish, not readable text.
- **Bloated**: headers and attachment payloads were inlined (~320KB/message in the reporter’s mailbox), driving up storage and embedding cost/rate-limiting.

Fixes #11549

## Changes

- Add `extract_text_body`, which parses the raw message with `mailparse` (already a dependency) and returns human-readable text: it walks the MIME tree preferring the decoded `text/plain` part, falls back to `text/html`, then the top-level decoded body for non-multipart messages, and finally to the raw decoded bytes if `mailparse` cannot parse the message at all — so content is never silently dropped.
- Add `find_part_body`, a recursive MIME-part search whose `get_body()` applies each part’s transfer-encoding and charset. Attachment part bytes are never returned.
- Use `extract_text_body` when populating `content` in `scan`.

Scope note: this fixes the stored-content correctness/bloat; it still fetches the whole message over the wire (`BODY.PEEK[]`) and discards attachment bytes when building `content`. `BODYSTRUCTURE`-based selective fetch and HTML tag-stripping are deliberately out of scope to avoid a new dependency and keep this focused.

## Test plan
- [x] Added regression test for a base64 `text/plain` part decoding to readable text (the reported gibberish case)
- [x] Added edge case tests for quoted-printable decode, `text/plain`-preferred-over-`text/html` in multipart/alternative, `text/html` fallback, attachment-bytes exclusion in multipart/mixed, and single-part plain
- [x] `make lint` passes (workspace-wide fmt + full clippy; the whole-workspace gate, not a scoped per-crate invocation)
- [x] `make test` passes

## Checklist
- [x] Root cause fixed at the source (no downstream workaround)
- [x] Regression tests that fail on the old raw-MIME code and pass on the new
- [x] No behavior change outside the IMAP `content` column

### DIFF
--- a/crates/data_components/src/imap/provider.rs
+++ b/crates/data_components/src/imap/provider.rs
@@ -25,7 +25,7 @@ use datafusion::{
     logical_expr::Expr,
     physical_plan::ExecutionPlan,
 };
-use mailparse::dateparse;
+use mailparse::{ParsedMail, dateparse, parse_mail};
 use snafu::prelude::*;
 
 use crate::arrow::write::MemTable;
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

claudespice:
@copilot review

lukekim (approved):

PR #11676, opened 2026-07-06
build: define clippy lints in [workspace.lints], opt in every crate

bjchambers:
## What

A scoped `cargo clippy -p <crate> --no-deps` runs at **default lint levels**, so it (and rust-analyzer) passes green while `make lint-rust` — which denies `clippy::pedantic`, `unwrap_used`, `expect_used`, `clone_on_ref_ptr` and more — fails in CI. This moves the clippy lint levels into `[workspace.lints.clippy]` in the root `Cargo.toml` and opts every workspace member in via `[lints]\nworkspace = true`, so scoped clippy and the editor surface the same violations as CI, **at edit time**.

## Changes

- **`[workspace.lints.clippy]`** (root `Cargo.toml`): mirrors the Makefile's deny/allow set. `pedantic` is denied at `priority = -1` so the `module_name_repetitions` / `large_futures` / `too_many_lines` opt-outs win. `-Dwarnings` intentionally stays **Makefile-only** — denying it here would fail ordinary `cargo build`.
- **`[workspace.lints.rust]`**: hoists `telemetry`'s `unexpected_cfgs` check-cfg for `tokio_unstable` to the workspace, so `telemetry` can inherit via `workspace = true` (a crate can't mix `workspace = true` with its own `[lints.*]` overrides).
- **116 member crates** opt in with `[lints]\nworkspace = true`. Skipped: vendored crates (`crates/vendor/*`), `crates/libnfs`, and the workspace-excluded `connector-nfs` — matching the Makefile's `--exclude`s.
- **Makefile / CI unchanged** — it keeps the explicit `-D` flags as a **backstop**: a crate that forgets to opt in still fails `make lint-rust` over `--workspace` rather than silently under-linting.
- **CLAUDE.md**: documents that new crates must opt in, with CI as the backstop.

Test-context exemptions (`allow-expect-in-tests`) already live in the root `clippy.toml`, so scoped clippy on test code isn't falsely flagged.

## Not a perfect substitute

Scoped `cargo clippy -p <crate>` still won't compile feature-gated code without `--features`, so `make lint-rust` remains the gate for the full release feature set + the `--tests` pass. This closes the *lint-level* gap, not the *feature* gap.

## Validation

- `cargo metadata` parses all 118 manifests.
- Scoped `cargo clippy -p util -p telemetry --all-targets` is clean and inherits the levels (telemetry's `tokio_unstable` cfg still resolves via the workspace rust lints; test `.expect()`s not falsely flagged).

Draft pending a full `make lint-rust` / CI pass to confirm the workspace-wide inheritance is green end-to-end.


claudespice:
Heads-up: the failing **Elasticsearch Integration Tests** check here is not caused by this PR. It is a latent \`trunk\` break — reqwest 0.13 (#11506) moved \`RequestBuilder::query\` behind a new \`query\` cargo feature that the workspace `reqwest` feature list omits. Full-workspace builds stay green via feature unification, but the ES job builds \`-p elasticsearch\` in isolation and hits `E0599: no method named `query``. This PR just surfaced it by touching every crate's `Cargo.toml` (re-triggering the path-gated job). Fix is up in #11760 (one-line: add \`query\` to the workspace reqwest features); once it merges and this branch picks up trunk, the ES check will pass.

phillipleblanc (approved):

PR #11680, opened 2026-07-06
feat(cdc): cap keys per durable delete plan + absorb fall-through metrics (#11673)

### LINKED ISSUE #11673: CDC delete path: cap keys per delete-plan execution + absorb fall-through reason counter
**Goal:** No single durable delete executes an unbounded predicate; the reason deletes skip the in-memory absorb is a printed fact.

**Context:** Composite-PK delete bursts currently build one predicate over up to 16,384 keys (a ~50 k-comparison OR-tree) and execute it as a single DELETE plan — measured at ~89 s per execution, which pegged the table's prefetch, triggered walsender timeouts, and paced entire shared slots. This path is used by **any** changes-mode source with deletes, not just Postgres. Two changes: (1) in `process_delete_batch`, chunk keys at `CDC_DELETE_SUBBATCH_MAX` (default 2,048) per `delete_from` execution — cost becomes linear and interruptible instead of monolithic; (2) count absorb fall-throughs: `runtime_cdc_delete_absorb_fallthrough_total{dataset, reason}` with `no_capability | no_advancer | inextractable_keys | budget`, plus a `keys_per_delete_burst` histogram. (2) is what tells us whether the eventual real fix (composite-key absorb) targets the right reason.

**Done when:** unit test shows an N-key delete burst executes ⌈N/cap⌉ plans with identical end state; the benchmark's delete table shows bounded burst durations; fall-through reason visible in metrics/waterfall.

**Risk:** low — chunking a batch of independent key-deletes preserves semantics (each key deleted exactly once; ordering within a burst is not load-bearing across chunks of distinct PKs).

bjchambers:
Fixes #11673.

## Problem

Composite-PK delete bursts built **one** predicate over up to 16,384 keys (a ~50k-comparison balanced OR-of-ANDs tree) and executed it as a **single** `delete_from` plan. On Cayenne that execution is superlinear (~O(N²)): measured at ~89 s per execution in production, which pegged the table's prefetch buffer, tripped walsender timeouts, and paced entire shared replication slots. This path is used by **any** changes-mode source with deletes, not only Postgres.

## Changes

**1. Cap keys per durable delete plan.** `process_delete_batch` now applies keyed durable deletes in chunks of `cdc_delete_subbatch_max` (default **2,048**) rows per `delete_from`, so an N-key burst runs as `⌈N/cap⌉` independent, interruptible plans instead of one monolithic predicate. Keys across chunks are distinct PKs, so each is deleted exactly once and cross-chunk ordering is not load-bearing; all chunks run under the single write lock held for the call, so the burst stays isolated. The cap is a standard CDC tunable:
- env `SPICE_CDC_DELETE_SUBBATCH_MAX`
- runtime param `cdc_delete_subbatch_max` (with per-dataset overlay)

**2. Absorb fall-through observability** — the diagnostic the issue calls out as *"what tells us whether the eventual composite-key absorb fix targets the right reason"*:
- `dataset_acceleration_cdc_delete_absorb_fallthrough_total{dataset, reason}` — counts delete sub-batches that could not be absorbed in-memory and fell through to a durable delete, by `reason`: `no_capability | no_advancer | inextractable_keys | budget`.
- `dataset_acceleration_cdc_keys_per_delete_burst` — histogram of the key count of every delete sub-batch, so we can see whether the cap actually binds in practice.

_(Metric names follow the existing `dataset_acceleration_cdc_*` convention of the sibling CDC metrics in the same meter, rather than the issue's `runtime_cdc_*` shorthand.)_

## Verification

**Microbenchmark** (`crates/runtime/benches/cdc_delete_burst.rs`) — composite-PK delete burst through the durable `delete_from` path on a **Cayenne** accelerator (the SF-1000 engine), monolithic (`cap = N`) vs chunked (`cap = 2048`):

| N (keys) | monolithic (1 plan) | chunked (cap=2048) | speedup |
|---|---|---|---|
| 2,048 | 184 ms | 184 ms (1 plan) | 1.0× |
| 8,192 | 2,125 ms | 744 ms (4 plans) | **2.9×** |
| 16,384 | 8,159 ms | 1,530 ms (8 plans) | **5.3×** |

Monolithic is ~O(N²) — 8× the keys → 44× the time (184 ms → 8.16 s) — and its throughput collapses from ~11 Kelem/s to ~2 Kelem/s. Chunked scales linearly and holds ~11 Kelem/s regardless of N. The win grows with burst size, matching the production ~89 s pathology. (An initial `MemTable`-based version of this bench showed only ~8%, because `MemTable`'s predicate evaluation is linear — it is the wrong engine; Cayenne is where the blowup lives.)

**Unit test** — `test_delete_burst_chunks_at_subbatch_cap`: for caps {1,3,4,10,100} over an N=10 key burst, asserts exactly `⌈N/cap⌉` `delete_from` executions (via a counting provider) and identical end state (all keys deleted) across every cap.

## Risk

Low. Chunking a batch of independent key-deletes preserves semantics (each key deleted exactly once; cross-chunk ordering of distinct PKs is not load-bearing). The in-memory absorb path and keyless-delete path are unchanged. Default cap (2,048) is conservative and back-compat via env/param.

## Follow-up

SF-1000 HTAP run dispatched on this branch to quantify the end-to-end replication-lag / freshness improvement on the composite-PK delete tables; numbers to be appended here.


lukekim (dismissed):

bjchambers:
## SF-1000 HTAP run — parity, no regression (as expected for this workload)

Ran the flagship SF-1000 config pinned to this branch's HEAD and compared to the latest trunk run of the same config.

- **Branch** `e5baae1e6` — run [28832072532](https://github.com/spiceai/spiceai/actions/runs/28832072532) (`cdc-tuned-memory-sf1000`, SF1000, 100 terminals, 600 s, rate 10000)
- **Trunk baseline** — run [28826476377](https://github.com/spiceai/spiceai/actions/runs/28826476377) (same config)

| metric | branch (`e5baae1e6`) | trunk baseline | Δ |
|---|---|---|---|
| QPH (analytical) | 1091.0 | 1124.1 | −3% |
| tpmC (NewOrder/min) | 265,456 | 264,922 | +0.2% |
| worst freshness p99 | 332 s (`order_line`) | 305 s (`order_line`) | +9% |
| **`new_order`** freshness p99 (delete-heavy) | **3,474 ms** | **3,536 ms** | −2% |
| **`new_order`** lag p99 / deletes | 4,180 ms / 2.42 M | 4,834 ms / 2.45 M | −14% |
| `order_line` lag p99 (0 deletes) | 336,864 ms | 326,298 ms | +3% |
| `stock` lag p99 (0 deletes) | 232,762 ms | 255,806 ms | −9% |

**Read:** every axis is within single-run SF-1000 variance, and by construction this change cannot affect insert/update lag. The delete path change is neutral on the top line — **which is the expected result**, because at SF-1000 steady state replication lag is **insert/update-bound**: it lives entirely on `order_line` (12 M inserts + 10.8 M updates, **0 deletes**) and `stock` (16.5 M updates, 0 deletes). The one delete-heavy table, `new_order` (2.4 M **composite-PK** deletes), stays fresh in both (p99 ~3.5 s).

So this run demonstrates **no regression**, not a headline win. The fix's value is the durable delete path itself: monolithic composite-PK delete plans are ~O(N²) and chunking makes them linear/bounded — **5.3× at 16,384 keys** on the Cayenne microbench in this PR. That is a latency-tail / resource-isolation safeguard against the ~89 s pathology when a large delete burst occurs; it is not expected to move an insert-bound benchmark.

**Observability gap → motivates a follow-up:** the new `cdc_keys_per_delete_burst` / `cdc_delete_absorb_fallthrough_total{reason}` metrics are **not** captured by the HTAP run (testoperator doesn't scrape them; the diagnostics artifact only has query explain/analyze dumps). Wiring them into the HTAP waterfall (separate PR) would let us confirm, on a live run, how large `new_order`'s delete bursts actually get and whether they hit the durable path or would benefit from composite-key in-memory absorb — the real next lever for the delete side of this workload.

crates/runtime/src/accelerated_table/refresh_task/changes.rs
```diff
@@ -5116,6 +5321,54 @@ mod tests {
         }
     }
 
+    /// Wraps a `TableProvider` and counts each `delete_from` call, delegating
```
phillipleblanc, line 5324: You should add the clippy lint "#[deny(clippy::missing_trait_methods)]" and implement all optional methods.

This is an _extremely_ common footgun with wrapping table providers. You should also make sure to update any unwrapping methods (i.e. `find_concrete_table_provider`) to understand this wrapper as well.

phillipleblanc: This should also be applied to `WriteOrderRecordingProvider` below

phillipleblanc: Ah ok, this is a test-only provider.

phillipleblanc (dismissed):

crates/runtime/src/accelerated_table/refresh_task/changes.rs
bjchambers: Agreed on the general footgun. Since it's a test-only wrapper I didn't add `#[deny(clippy::missing_trait_methods)]` (that pulls in a lot of boilerplate for a mock), but I did forward the behaviorally-meaningful optional methods — `constraints`, `supports_filters_pushdown`, and `statistics` — to the inner provider in `b50fde5c6`, so the wrapper mirrors what it wraps rather than silently reverting to trait defaults. Happy to switch to the deny-lint approach if you'd prefer it enforced.

lukekim (approved):

PR #11689, opened 2026-07-07
ci: dedupe testoperator query_overrides run steps

Jeadie:
## Summary

The `testoperator_run_*` workflows each had **two near-identical run steps** — one gated on `query_overrides == ''` and a copy gated on `query_overrides != ''` whose only real difference was an extra `--query-overrides` flag (plus a fully duplicated ~50-line `env:` block).

This merges each pair into a single step, injecting `--query-overrides` conditionally with the same `format(...) || ''` expression already used elsewhere in these files for optional flags (e.g. `--scenario-query-file`, `--random-param-set-count`).

## Files

- `testoperator_run_bench.yml`
- `testoperator_run_data_consistency.yml`
- `testoperator_run_load.yml`
- `testoperator_run_throughput.yml`

Net: **4 insertions, 244 deletions**. No behavioural change — when `query_overrides` is empty the expression collapses to a blank line-continuation (same as the existing optional-flag lines); when set it expands to `--query-overrides <value>`.


Jeadie:
Superseded by a PR targeting trunk instead of release/2.1.

Issue #11683, opened 2026-07-06
bug(cayenne): Vortex background compaction race panics `compaction-worker` during ch-bench/HTAP runs


phillipleblanc:
## Summary

During ch-bench / HTAP benchmark runs (Postgres source accelerated by Cayenne in CDC `changes` mode), the background Vortex compaction path intermittently **panics on the `compaction-worker` thread**. The release profile builds with `panic = "abort"`, so the panic terminates the whole `spiced` process rather than failing a single task. This is a race condition in the Vortex compaction path.

## Panic

```
thread 'compaction-worker' panicked at
  vortex-io/src/runtime/handle.rs:218:17:
Runtime dropped task without completing it, likely it panicked
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Accompanied on the high-write CDC tables (`order_line`, `stock`, `customer`, `oorder`) by:

```
WARN vortex_datafusion::persistent::sink: Vortex sink shard writer task failed to join: task N was cancelled
WARN cayenne::compaction: Background compaction failed: Execution error:
     Vortex sink shard writer task failed to join: task N was cancelled  table="order_line"
```

## Where it's observed

- Workflow **`htap benchmark tests`**, on the accelerated `postgres-cayenne[file]-*` configs (`cdc-tuned`, `adaptive`, `cdc-tuned-turso`, `cdc-tuned-memory`). Never on `duckdb`/`arrow`.
- Present in **16 failed runs between 2026-06-17 and 2026-07-05**, across at least three vortex fork revisions (`31c6537`, `729249d`, `ab4f0b1`) — persistent, not a recent regression.
- Representative runs:
  - https://github.com/spiceai/spiceai/actions/runs/28729499048 (2026-07-05, trunk)
  - https://github.com/spiceai/spiceai/actions/runs/28350999640 (2026-06-29, trunk)
  - https://github.com/spiceai/spiceai/actions/runs/28129470186 (2026-06-24, trunk)
  - https://github.com/spiceai/spiceai/actions/runs/27709323213 (2026-06-17, trunk)
- No CI run has `RUST_BACKTRACE` set, so no backtrace is captured and the inner task is not yet pinned.

## Leads (not a confirmed root cause)

- The panic site `vortex-io/.../runtime/handle.rs:218` fires when the runtime backing a Vortex `Task` is dropped while that task is still pending.
[...]

PR #11692, opened 2026-07-07
fix(cayenne): harden Vortex compaction writes

phillipleblanc:
## Summary

Draft PR for review while we continue validating the exact customer failure.

This has two related Cayenne/Vortex compaction hardening changes:

1. **Shutdown / teardown hardening**
   - Track only active Vortex-producing Cayenne maintenance pass bodies, not the long-lived background scheduler loops.
   - Stop new compaction-runtime passes when runtime shutdown begins.
   - Drain active compaction passes before the dedicated compaction runtime is dropped.
   - Increase the bounded compactor drain window to 120s, matching observed CH-benCH SF100 compaction durations.

2. **Compaction write-boundary schema normalization**
   - Normalize batches entering `CayenneTableProvider::write_to_snapshot()` to the stored table schema before Vortex derives/checks its writer dtype.
   - This covers the restored current-snapshot background compactor path, which builds rewrite input through `TableProvider::scan()` and can therefore emit the query read schema (for example `Utf8View`) while the Vortex writer schema is the stored schema (`Utf8`).

## Why

Refs #11683.

The CI/local reproduced failure is an active Cayenne/Vortex compaction pass interrupted during shutdown/runtime teardown, producing:

```text
Runtime dropped task without completing it, likely it panicked
```

Separately, the customer-reported issue is a background-compactor panic. The schema-normalization change is a hardening fix for a concrete read-schema/write-schema seam in the current-snapshot compactor, but the customer root cause is **not yet proven**.

## Customer-shaped CDC repro status

A local repro using the same source mechanism as the customer — Postgres CDC (`refresh_mode: changes`) into a Cayenne file accelerator — was added and run:

```text
~/code/spicepod_test/260707-cayenne-vortex-compaction-race/postgres-cdc-upsert-protected-compaction-repro.sh
```

Latest PASS artifact:

```text
~/code/spicepod_test/260707-cayenne-vortex-compaction-race/postgres-cdc-upsert-protected-compaction-203303_41339/summary.txt
```

Result:

- Postgres CDC + `on_conflict: { id: upsert }` ingested 48 rows.
- Protected-snapshot subset compaction ran and completed 11 times.
- No `Runtime dropped task`, `panicked at`, Vortex dtype assertion, or abort markers.

Important caveat: valid Cayenne/Postgres `refresh_mode: changes` requires an `on_conflict` upsert config, so this customer-shaped CDC path writes protected snapshots and did **not** exercise the append-only/current-snapshot compactor. Postgres `text`/`varchar` columns also surfaced as `LargeUtf8`, not the `Utf8`/`Utf8View` seam from the targeted unit regression.

So this PR should be read as hardening + candidate fix, not as confirmed closure of the customer issue without the exact customer panic/backtrace/config.

## Validation

Passed locally:

```text
cargo fmt --all -- --check
cargo test -p cayenne compaction_pass_tracker_drains_when_guard_drops --lib
cargo test -p cayenne background_compactor_drains_in_flight_compaction_on_drop --lib
cargo test -p cayenne background_compactor_ticks_at_interval_and_stops_on_shutdown --lib
cargo test -p cayenne current_snapshot_compaction_casts_force_view_scan_to_write_schema --lib
cargo test -p cayenne current_snapshot_compaction_casts_force_view_scan_to_write_schema --lib --release
cargo test -p cayenne force_view_read_schema_scan_emits_utf8view --lib
cargo test -p cayenne --test small_files_compaction_test sqlite -- --nocapture
cargo check -p cayenne -p runtime
cargo build -p spiced --release --features models
cargo deny check licenses
cargo deny check advisories
make lint-rust
```

CI/review follow-up:

- Rebased onto latest `trunk` (`2370a819`).
- Added required PR metadata (`area/cayenne`, assignee).
- Addressed Copilot's shutdown-timeout review by using a fixed 2-minute compaction drain window instead of `max(remaining_timeout, drain_timeout)`.

Existing unrelated warning remains:

```text
warning: fields `catalog` and `table_identifier` are never read
--> crates/runtime/src/dataconnector/iceberg.rs:110:5
```

Manual/local artifacts:

- Failing shutdown repro before fix: `harness-sf100-101519.log`
- Fixed shutdown repros: `harness-sf100-120245.log`, `harness-sf10-125523.log`, `harness-sf10-141554.log`
- Bootstrap no-repro: `bootstrap-oorder-new-order-141337.log`
- Customer-shaped CDC protected-compaction pass: `postgres-cdc-upsert-protected-compaction-203303_41339/summary.txt`
- Schema mismatch side artifacts under `~/code/spicepod_test/260707-cayenne-vortex-compaction-race`

## Review notes

This remains draft because the exact customer root cause is not confirmed. The shutdown drain and write-boundary schema normalization are intentionally separate mechanisms; please review them independently.

### DIFF
--- a/crates/cayenne/src/lib.rs
+++ b/crates/cayenne/src/lib.rs
@@ -101,11 +101,12 @@ pub use provider::{
     CayenneCdcWrite, CayenneContext, CayenneStagedAppend, CayenneTableProvider,
     CayenneTableProviderBuilder, EncodeBudgetSnapshot, PARTITIONED_WAL_DIR, PartitionedWal,
     PartitionedWalEntry, PreparedOverwrite, PreparedStagedAppend, QueryObservations, SlotAdvancer,
-    TimeRetentionFilterBuilder, cap_global_encode_concurrency, deregister_query_observations,
-    encode_budget_snapshot, global_mem_tier_total, global_mem_tier_used, global_qph,
+    TimeRetentionFilterBuilder, begin_compaction_shutdown, cap_global_encode_concurrency,
+    deregister_query_observations, drain_compaction_tasks, encode_budget_snapshot,
+    global_mem_tier_total, global_mem_tier_used, global_qph, in_flight_compaction_tasks,
     record_global_query, record_query_latency, register_query_observations,
-    set_compaction_runtime_env, set_compaction_runtime_handle, set_cpu_burstable,
-    set_global_encode_concurrency, set_global_mem_tier_bytes, set_global_memory_budget,
-    set_query_admission_governor, update_global_mem_tier_total,
+    reset_compaction_shutdown, set_compaction_runtime_env, set_compaction_runtime_handle,
+    set_cpu_burstable, set_global_encode_concurrency, set_global_mem_tier_bytes,
+    set_global_memory_budget, set_query_admission_governor, update_global_mem_tier_total,
 };
 pub use schema::transform_schema_for_vortex;
--- a/crates/cayenne/src/provider/compaction.rs
+++ b/crates/cayenne/src/provider/compaction.rs
@@ -36,6 +36,7 @@ limitations under the License.
 //! tables can't overwhelm the writer pool.
 
 use std::future::Future;
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, LazyLock, Weak};
 use std::time::{Duration, Instant};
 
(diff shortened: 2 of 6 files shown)


bjchambers:
This may make the issue with the reload of data sets and the correspond drain in the fuzz test unnecessary. I think that was hitting inconsistent behavior when the old compaction threads weren't shutdown. IIUC this should shut them down more regularly. Not sure if we want to back out that method and just rely on the "drain on shutdown" but they would feel a little cleaner.

lukekim (approved):

PR #11701, opened 2026-07-07
feat(datafusion-flightsql): back with QueryEngine + add do_exchange

Jeadie:
## What

Two changes to the standalone `datafusion-flightsql` crate to make it a more capable, reusable Flight SQL service:

### 1. Back the service with `QueryEngine` instead of `SessionContext`
`FlightSqlService` now holds `Arc<dyn QueryEngine>` (from `runtime-query-engine`) rather than a concrete `Arc<SessionContext>`, so it can be backed by any query engine implementation.

- `FlightSqlService::new(engine)` takes an `Arc<dyn QueryEngine>`.
- `FlightSqlService::from_session_context(ctx)` preserves the bare-`SessionContext` ergonomics (wraps `QuerySession`).
- All handlers migrated. They reach through `engine.session_context()` only where the trait has no equivalent: schema-from-arbitrary-SQL, catalog walks, and the write/insert path.

### 2. Implement `do_exchange` (CDC-style subscriptions)
Ported from the runtime's `do_exchange`, decoupled from runtime infrastructure:

- New `ChangeSource` trait: the live change feed is a generic `BoxStream<Result<DataUpdate, Status>>` rather than a broadcast receiver, so a host can bridge its own change-propagation mechanism.
- Streams an initial snapshot (as `op = "r"` change batches) then interleaves live updates, with the same bounded-buffer back-pressure as the runtime.
- Returns `Unimplemented` when no `ChangeSource` is configured.
- `changes_schema` is inlined (verified byte-identical to `data_components::cdc::changes_schema`) to avoid depending on the heavy `data_components` crate — no wire-format change for existing CDC clients.

## Testing
- `cargo check -p datafusion-flightsql` passes cleanly.

## Follow-ups (not in this PR)
- Tests for the migrated handlers and a `do_exchange` test with a stub `ChangeSource`.
- clippy pass.
- Runtime adoption: implement `QueryEngine` bridging and a `ChangeSource` adapter over `DataUpdateBroadcaster`, then have `runtime/src/flight` sit on top of this crate.

🤖 Draft — opening for early review of the approach.


Jeadie:
Superseded by #11703. This branch was cut from `release/2.1` and conflicts badly against `trunk`; #11703 is the same change cherry-picked onto `trunk`.

Issue #10970, opened 2026-05-21
runtime.params.cayenne_footer_cache_mb accepts invalid values silently; no startup log confirms applied value; typo'd keys not detected


phillipleblanc:
## Summary

`runtime.params.cayenne_footer_cache_mb` (added by #10929) is accepted at any value with no validation and no startup log line confirming the configured size. Operators have no signal that their config took effect.

A typo of the key name (`cayenne_footer_cach_mb`) is also silently accepted — no warning that the key was unknown.

## Version

- Runtime: `v2.0.0-unstable-build.b93578319a+models` (rc.5 candidate).

## Repro

Three variants, all accepted with no error and no log indication that the value was parsed or applied:

```yaml
runtime:
  params:
    cayenne_footer_cache_mb: "not_a_number"
```
↳ starts cleanly; no log mentioning footer cache; default behavior unchanged as far as the operator can tell.

```yaml
runtime:
  params:
    cayenne_footer_cache_mb: "-100"
```
↳ same — starts cleanly, no warning that negative is meaningless.

```yaml
runtime:
  params:
    cayenne_footer_cach_mb: "1"   # note typo: cach instead of cache
```
↳ same — accepted, no "unknown param" message; operator believes they configured a 1 MB cap when they actually configured nothing.

## Impact

- **Silent misconfiguration.** Per the spirit of #10929, the param is supposed to let operators tune the DataFusion file-metadata cache when running large Cayenne datasets. If the value is invalid or the key is misspelled, the operator runs at default sizing without knowing.
- **Production debugging burden.** When a Cayenne workload exhibits unexpected memory or latency behavior, the first question is "what did you set the footer cache to?" — and the runtime currently can't answer it from logs.

## Suggested fix

At spice startup, when `runtime.params.cayenne_footer_cache_mb` is set:
1. Parse it as a non-negative integer; reject malformed values with a startup error like `runtime.params.cayenne_footer_cache_mb: expected non-negative integer, got '-100'`.
2. Emit an INFO log: `Cayenne footer cache: 1 MiB (overridden via runtime.params.cayenne_footer_cache_mb)`.
3. Maintain a known-key list for `runtime.params` and warn on unknown keys (`runtime.params.cayenne_footer_cach_mb: unknown key; did you mean 'cayenne_footer_cache_mb'?`). This generalizes beyond just the footer-cache param.

[...]

PR #11719, opened 2026-07-08
fix(spicepod): reject unknown fields on columns[] config (fixes #10972)

### LINKED ISSUE #10972: Spicepod YAML accepts unknown fields and garbage values silently; spice validate doesn't catch them either
## Summary

Spicepod YAML parsing accepts unknown fields (and unknown values for known fields) silently. `spice validate`, which is documented to check "YAML syntax and schema," does not catch these either. Operators have no feedback when they typo a field name, mis-name a parameter, or supply a value that the runtime will silently default.

## Version

- Runtime/CLI: `v2.0.0-unstable-build.b93578319a+models` (rc.5 candidate).

## Repro examples

All of these pass `spice validate` with `OK` and start the runtime cleanly:

### A — Typo'd `columns[]` field names

```yaml
datasets:
  - from: file:data.csv
    name: items
    columns:
      - name: id
        typee: bigint       # typo: "typee" instead of "type"
        nulleble: false     # typo: "nulleble" instead of "nullable"
```

`spice validate` reports `OK coltype-typo-key`. The runtime starts; the `typee`/`nulleble` fields are silently discarded; the operator believes they declared a schema.

### B — Garbage values for `columns[].type`

```yaml
columns:
  - name: id
    type: definitely_not_a_real_type
  - name: name
    type: xyz_blob_garbage
```

`spice validate` returns `OK`. The runtime starts and reports the source's actual types via `describe`. Per #10661 the type field "has no runtime effect in this PR — they are forwarded through the spicepod model for later phases" — but the lenient parser is meant to be exercised on load; unrecognised values should at least warn.

### C — Typo'd `runtime.params` key

```yaml
runtime:
  params:
    cayenne_footer_cach_mb: "1"   # typo: "cach" instead of "cache"
```

Accepted by both `spice validate` and `spice run`. (Tracked separately in #10970 for the footer-cache flavor; the broader behavior is the same.)

### D — Invalid value for typed `runtime.params`

```yaml
runtime:
  params:
    cayenne_footer_cache_mb: "not_a_number"
```

(Also #10970.)

## Impact

- **Silent misconfiguration.** This is the same family as the connector and accelerator parameter issues — operators add a setting, the runtime accepts it, and the configured behavior never takes effect. The only way to discover the mistake today is to spot that the runtime behaves identically to default.
- **`spice validate` is misleading.** Its `--help` text promises "YAML syntax and schema" checks, but it only catches structural errors (missing top-level keys, malformed YAML). Unknown component fields are passed through. The CLI invites operators to use it as a pre-deploy gate; for the cases above, that gate is open.

## Suggested fix

Two options, not mutually exclusive:

1. **Strict-by-default field validation** at spicepod parsing. Serde-style `deny_unknown_fields` on the model structs, surfacing a clear startup error: `Unknown field 'typee' on columns[] (did you mean 'type'?)`. This needs to be opt-out for forward-compat (e.g. `--allow-unknown-fields` flag) but the default should be strict.
2. **At minimum, warn.** If breaking on unknown fields is too aggressive for shipping rc.5, log a single WARN per unknown field at parse time: `runtime.params.cayenne_footer_cach_mb: unknown key (did you mean 'cayenne_footer_cache_mb'?)`. `spice validate` should at least display these warnings and exit non-zero with a `--strict` flag.

Per-component examples that should be covered:
- `columns[]` unknown fields (this issue's A and B)
- `runtime.params.*` unknown keys (#10970)
- `acceleration.*` (less likely to be broken since it's a typed struct, but worth confirming)
- Top-level dataset fields (`on_schema_change` was unknown at the dataset level until recently — see related #10969).

## Related

- #10970 — `cayenne_footer_cache_mb` accepts invalid values silently.
- #10969 — `on_schema_change` setting is honored by the YAML parser but not by the CDC code path.
- #10661 — added `columns[].type/nullable` as a forwarded field; runtime side is Phase 2+.

claudespice:
## Summary

Spicepod dataset column configuration silently accepted unknown/typo'd fields. A column defined with `typee: bigint` or `nulleble: false` parsed cleanly — `spice validate` reported `OK` and the runtime started — while the misspelled fields were discarded and the intended schema declaration never took effect (issue #10972, case A).

**Root cause:** the three column-level config structs — `Column`, `ColumnLevelEmbeddingConfig`, and `FullTextSearchConfig` (`crates/spicepod/src/semantic.rs`) — lacked `#[serde(deny_unknown_fields)]`. Their parent `Dataset` and ~30 sibling component structs across the crate already deny unknown fields, so a typo at the dataset level (`datasets[].nam`) errored, while the same class of typo one level down under `columns[]` passed silently. That is an inconsistency, not a deliberate leniency policy.

## Changes

- Add `#[serde(deny_unknown_fields)]` to `Column`, `ColumnLevelEmbeddingConfig`, and `FullTextSearchConfig`, bringing them in line with `Dataset` and the rest of the crate. serde now rejects unrecognized keys, naming the field it did not expect.
- None of the three structs use `#[serde(flatten)]` — their `params`/`metadata` are explicit keys — so open connector/engine params and column metadata are unaffected.
- Add regression tests: valid fields (incl. the `data_type` alias) still parse; unknown fields on each of the three structs are now rejected.

## Scope

Fixes issue case **A** (column field-name typos). Deliberately out of scope:
- Cases **C**/**D** — `runtime.params` key/value typos — are a free-form `HashMap` and are tracked separately in #10970.
- Case **B** — garbage *values* for `columns[].type` — depends on the Phase 2 type-parsing work noted in #10661 (the field is currently forwarded without runtime effect).

## Test plan

- `make lint`
- `make test`

## Checklist

- [x] Existing spicepod fixtures (`test_all_v1_fixtures_load`, `test_all_v2_fixtures_load`) still load
- [x] `schemars` feature still compiles (deny_unknown_fields sets `additionalProperties: false` in the generated schema)
- [x] No behavior change to open `params`/`metadata` maps

Fixes #10972

### DIFF
--- a/crates/spicepod/src/semantic.rs
+++ b/crates/spicepod/src/semantic.rs
@@ -30,6 +30,7 @@ use crate::{
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[cfg_attr(feature = "schemars", derive(JsonSchema))]
+#[serde(deny_unknown_fields)]
 pub struct Column {
     pub name: String,
 
@@ -148,6 +149,7 @@ impl From<&str> for Column {
 ///
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[cfg_attr(feature = "schemars", derive(JsonSchema))]
+#[serde(deny_unknown_fields)]
 pub struct ColumnLevelEmbeddingConfig {
     #[serde(rename = "from", default)]
     pub model: String,
@@ -270,6 +272,7 @@ where
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 #[cfg_attr(feature = "schemars", derive(JsonSchema))]
+#[serde(deny_unknown_fields)]
 pub struct FullTextSearchConfig {
     pub enabled: bool,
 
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #11732, opened 2026-07-08
perf(cdc): deferred parsing — move Postgres CDC decode + build off the shared pump

bjchambers:
## Summary

Moves the per-event **decode + Arrow build off the single shared-slot Postgres
replication pump** (the pre-fanout serialization point) onto the parallel
per-dataset consumers, via a destination-passing `ChangeRows` trait. The pump
now only *peeks* each change message's type + relation id to route it and
buffers the raw pgoutput `Bytes`; the consumer decodes + transforms + builds the
Arrow `ChangeBatch`.

Builds on #11709 (shared-slot pump per-event overhead).

## What changed

- **`ChangeRows` trait on `ChangeEnvelope`** — no-build metadata
  (`is_empty`/`num_rows_hint`/`encoded_len`/`source_commit_ts_ms`/`is_heartbeat`)
  + a fallible `build`, blanket-impl'd for an eager `ChangeBatch` so every
  non-Postgres connector (MongoDB/Kafka/DynamoDB/Debezium/MySQL) is unchanged.
  `into_parts`/`change_batch` now return `Result` (build is fallible);
  `From<ChangeBatchError> for StreamError`.
- **Pump peeks + buffers raw** — `pgoutput::{message_type, relation_id}` read the
  5-byte header without decoding the tuple; `handle_decoded` splits into
  `handle_relation` (validate/route) + `buffer_raw_change`/`buffer_raw_truncate`
  (buffer raw `Bytes`). The consumer decodes via `PgChangeRows`/`decode_raw_changes`,
  relocating the TOAST-merge + primary-key-change split off the pump.
- **Cache the resolved `MemberHandle` in the route map** — removes the per-event
  `members` mutex lock + `(String, String)` hash.
- **`FxHashMap` for the `u32`-keyed route + txn maps** — SipHash is DoS-resistance
  for untrusted keys; unjustified for internal relation ids on the hot path.

## Correctness

- **Differential tests**: the raw-buffered path produces a byte-identical
  `ChangeBatch` to the eager path across insert / PK-changing update / delete /
  non-key update / truncate.
- **Acking unchanged**: raw bytes ride the same post-apply `SharedLsnCommitter`
  (flush floor = `min` across members); buffered-but-unapplied changes are never
  acked, so a crash or a decode failure re-streams them. A deferred build failure
  surfaces as a terminal `StreamError` on that one dataset (mirrors the eager
  path's pump-side fatal), isolated from other datasets.

## Measured (`pgoutput_decode` / `pgoutput_routing` criterion benches, per 1000 rows)

- Build off pump: sheds **58–77%** of per-event CPU.
- Decode off pump: pump route-peek **~1.6 µs** vs eager decode+build **~162 µs (binary) / ~303 µs (text)**.
- Routing hasher: SipHash **18.6 µs → FxHash 6.4 µs (~2–3×)**.

Net: the shared pre-fanout pump goes from decode+build inline to peek+route+buffer,
with all decode+build relocated to the parallel per-dataset consumers.

## Deferred follow-ups (not in this PR)

- Decode-direct fusion (eliminate the per-row `TupleData` intermediate; ~10–15% extra total CPU).
- `ChangeBatch` → accelerator bypass (a `ChangeSink` seam straight into the upsert-batch + deletion-vector; measure e2e first — the accelerator sizes post-build).
- `AckTable.entries` contention under tiny-transaction floods (profile before sharding; `recompute` needs a min-over-all snapshot).


phillipleblanc (dismissed):

bjchambers:
**Review thread dispositions** — thanks for the thorough review; resolving the open threads with a summary of how each was handled.

**Fixed in code:**
- `PgChangeRows::new` precomputes the update-count and byte length once, so `num_rows_hint()`/`encoded_len()` are O(1) (no per-call rescan of the buffered messages).
- `deliver_commit` re-checks member liveness via `ack.is_streaming(member_key)` before delivering a coalesced burst.
- Deferred Arrow builds are moved off the async worker: `ChangeEnvelope::into_parts_offloaded` runs a deferred build on `spawn_blocking` and resolves an already-materialized (eager) envelope inline — wired at all three build sites (apply loop, index wrapper, embedding wrapper).
- `CDC_APPLY_BURST_ROWS_TOTAL` is now summed per schema-group and recorded only *after* that group applies, so a `SkipRun`/`Stop` failure can't inflate throughput with rows that were never written.
- `encoded_len` is schema-aware (`max(wire_bytes, rows × per-row fixed width)`); `CDC_APPLY_BURST_BYTES`'s description and the `cdc_item_budget_bytes` helper name now make clear it's a coalescing byte-budget proxy, not an exact Arrow-memory measurement.
- `DeferredBatchConsumed`'s message no longer calls itself an "internal error" — it names the legitimate double-access-after-failed-build case and points at the preceding `DeferredBuild` error as the underlying cause.

**Deferred to a focused follow-up (non-correctness):** the stale-`RouteMap`-entry cluster — `buffer_raw_change`/`buffer_raw_truncate` keep buffering and metering a detached member until reconnect, and `PgChangeRows::new` does a second O(N) pass over the buffered frames. Correctness is preserved: the `is_streaming` gate in `deliver_commit` drops those bytes, and the replication slot replays from the member's last-applied LSN on reconnect. The safe optimization is a lock-free liveness flag on the cached `MemberHandle` (checked on the hot path) — **not** removing the route, since Postgres does not replay `Relation` messages mid-connection, so dropping a live route would risk the re-attach path. That's a member-lifecycle change to the shared pump and belongs with the pump-diagnostics work rather than this PR; tracked as a follow-up.

bjchambers:
Addressed in 48def7628: `arrow_fixed_width` now covers Float16/Duration/Interval/FixedSizeBinary (fixing the coalescing-budget under-count for those types), and the MySQL bootstrap maps a ready-signal `into_parts()` failure to a dedicated `Error::BuildReadySignal` rather than `SchemaMismatch`. Resolving.

sgrebnov (approved):

PR #11736, opened 2026-07-08
perf(cayenne): unify upsert PK-key hashing on prehashed XXH3-128 digests

bjchambers:
## What & why

The per-row upsert **on-conflict loop** hashed each incoming row's `OwnedRow`
byte string up to **three times per row** — the in-batch dedup pre-pass, the
cross-batch `incoming_keys` probe, and the `existing_keys` keyset probe — using
the standard library's **SipHash** default (`RandomState`). The deletion index
one file over had already evicted SipHash for seeded XXH3 (see the
`DeletionIndexHasher` doc in `deletion_index.rs`). Composite PKs lengthen the
`OwnedRow` encoding, so this cost sat directly on the composite-PK
delete-burst path.

This unifies that path on a **prehashed XXH3-128** scheme, mirroring
`KeyDeletionIndex`:

- Hash each row's key bytes **once** into a `u128` digest (`hash_index::hash_key_128`).
- Key all three probe maps **and** the cached keyset on that `u128` via
  `PrehashedBuildHasher` (which passes the digest's own entropy through instead
  of re-hashing). One hash pass now serves every probe.
- `u128` becomes the key identity — birthday-bound collision `< ~1e-20` at 1e9
  keys, the same trade the key-based deletion index already makes. The
  `OwnedRow` is retained in each map value where iteration needs the raw bytes
  (bloom rebuild, shard routing, deletion lists).
- `shard_of_pk` is **deliberately untouched** (changing the shard hash would
  reassign shards mid-life).

Behaviour is unobservable — the hasher choice does not change query results.

## Why not the mechanical `XxHash3BuildHasher` swap?

A drop-in `XxHash3BuildHasher` over the `OwnedRow` (the obvious "phase 1") was
measured and **rejected**: it *regresses* composite PKs up to ~3.6×, because its
hasher streams (slowly) once the length-prefixed `OwnedRow` encoding exceeds its
32-byte inline buffer. Only the prehashed-`u128` scheme delivers the win.

## Microbenchmark

`cargo bench --bench pk_key_hashing -p cayenne` — real `OwnedRow` keys, three
per-row probes per apply, hasher varied while the algorithm is held fixed.
Median wall-clock per apply:

| PK shape / rows | `siphash_default` (prior) | `xxh3_buildhasher` (rejected) | **`xxh3_prehashed_128` (this PR)** | speedup vs SipHash |
|---|---:|---:|---:|---:|
| `i64` / 100 | 5.44 µs | 4.72 µs | **1.46 µs** | **3.7×** |
| `i64` / 1 000 | 53.2 µs | 45.8 µs | **13.6 µs** | **3.9×** |
| `i64` / 10 000 | 533 µs | 498 µs | **152 µs** | **3.5×** |
| `i64+utf8` / 100 | 6.40 µs | 21.8 µs ⚠️ | **1.77 µs** | **3.6×** |
| `i64+utf8` / 1 000 | 59.6 µs | 213 µs ⚠️ | **16.4 µs** | **3.6×** |
| `i64+utf8` / 10 000 | 626 µs | 2 239 µs ⚠️ | **174 µs** | **3.6×** |

⚠️ = the rejected drop-in swap regressing the composite-PK path.

## Follow-up (documented, not adopted here)

A second bench group (`pk_key_hashing_i64_native`) evaluates specializing
integer PKs onto a cheaper integer hasher. Over **native `i64`** keys (the
`Int64Pk` identity, no collision risk), `FxHash` is faster still — ~4.4× vs
SipHash and ~1.9× beyond the u128 scheme at 10k rows. Capturing it would fork
the keyset key type by PK strategy (`i64` vs `u128`) across sharding/bloom/
persistence — larger than this change — so it's left as a follow-up. The
uniform u128 scheme already captures 3–4× safely for both shapes.

## Testing

- `cargo test -p cayenne --lib` — 843 passed.
- Upsert/on-conflict/sharded integration suites: `on_conflict_test`,
  `on_conflict_edge_cases_test`, `large_upsert_test` (exercises
  `cdc_mem_tier_shards > 1`), `int64_pk_deletion_test`, `keybased_deletion_test`,
  `overwrite_resurrection_test`, `position_based_deletion_test`,
  `position_mode_upsert_test`, `upsert_with_pending_deletions_test`,
  `mutation_model_test`, `mutation_roundtrip_test`,
  `inline_with_pending_deletions_test`, `cdc_compaction_delete_race_test` — all
  green.
- `make lint-rust-fix PACKAGES="cayenne"` clean (lib + tests).

Existing upsert/on-conflict coverage is the behaviour-preserving safety net —
the hasher choice is unobservable, so a passing suite is the correctness proof.


lukekim (dismissed):

sgrebnov (dismissed):

bjchambers:
Addressed the Copilot review feedback in `fc5059dae`:

- **Digest-identity invariant** — added `debug_assert_eq!(digest, pk_digest(&key))` to both `PkDigestSet::insert_with_digest` and `CachedPkKeyset::insert_with_digest`. The design keys these maps on a precomputed digest for speed; the assert makes a caller passing a mismatched digest fail fast in debug/test builds (instead of silently corrupting the set) at zero release cost. The existing upsert/on-conflict/sharded test suites pass with the asserts active, confirming every call site passes a matching digest.
- **Hot-path pre-sizing** — `miss_keys` (bloom-split) is now sized to `batch.num_rows()`, and `validated_keys` (shard union) to the summed per-shard kept counts, avoiding rehash/reserve on the CDC apply path.

Also merged `trunk` to resolve conflicts; full `cargo test -p cayenne` and the scoped lint gate are green locally.

sgrebnov (approved):

PR #11739, opened 2026-07-09
fix(cayenne): retain source deletion index in scan memos to close ABA hazard (fixes #11303)

### LINKED ISSUE #11303: Cayenne: merged-scan-deletions memo keyed by raw Arc pointer is ABA-unsafe; stale deletion view can resurrect deleted rows
### Summary

The per-scan merged deletion memo introduced in #11249 (`merged_scan_deletions` in `crates/cayenne/src/provider/table.rs`) keys the file-side deletion index by its raw `Arc` pointer (`Arc::as_ptr` → `file_index_ptr`). The memo does **not** retain the index it was built from, and several deletion-publish paths replace the index without bumping either of the other two key components (`tier_version`, `structural_epoch`). If the allocator reuses the freed `ArcInner` address for a later index generation — likely, since every generation is the same type/size class — the memo produces a false hit and the scan serves a **stale merged deletion view**: rows deleted by the most recent publishes become visible again. Silent wrong results, allocator-dependent, so it would surface as an unreproducible flake.

### Where (at trunk `6f259e438`)

- Memo key + hit check: `crates/cayenne/src/provider/table.rs:13083-13112` (`merged_deletion_snapshot`), key struct at `table.rs:3048-3056`, `index_ptr()` at `table.rs:3085-3090`.
- Same pointer-identity assumption on the append-path memo extension: `table.rs:13725-13737`.

### Deletion publishes that replace the index but bump no key component

Each of these does `load_full → extend/rebuild → store` on `deletion_snapshot` and bumps neither `MemTier::version` nor `inlined_structural_epoch`:

- `DELETE FROM` sink: `crates/cayenne/src/provider/delete/sink.rs:747` (row keys) and `:861` (Int64 PKs)
- `update_file_deletion_cache`: `table.rs:7850`, `:7864`
- `upgrade_tombstones_for_flushed_pks`: `table.rs:7939`
- `publish_staged_key_deletion_cache`: `table.rs:8956`, `:8969`
- `refresh_deletion_cache` / retention: `table.rs:11452`

(`publish_on_conflict_update` at `table.rs:9450-9482` does bump the structural epoch, so that path is safe.)

### Failure sequence

1. Memory-mode CDC table with resident mem-tier tombstones (so the memo path is active). A scan builds the memo from file index **gen A** at address `P`, tier version `V`, epoch `E`. The memo stores only the *merged* (new) index — gen A is not retained.
2. `DELETE FROM t WHERE …` publishes **gen B** (new address). Gen A's `ArcInner` is freed once in-flight scans finish.
3. A second deletion publish (another `DELETE`, a retention sweep, or a staged-CDC finalize) builds **gen C**. The allocator can hand gen C the freed block at `P` (same size class, LIFO reuse).
4. Next scan loads gen C: `ptr == P`, tier still `V`, epoch still `E` → memo **hit** → the scan filters with `merged(gen A ∪ tier)`, missing every tombstone added in steps 2–3 → deleted rows are returned.

The append-path extension at `table.rs:13725` has the same hazard and would then carry the stale base forward under sustained CDC (it re-keys the memo to the new tier version while keeping the stale `merged`).

### Suggested fix

Add a monotonic `u64` generation to `PkDeletionStrategyWithCache` (or to the snapshot structs), bumped on every `deletion_snapshot.store(...)`, and key the memo on that generation instead of `Arc::as_ptr`. Stores are already serialized by the per-table write lock, so a relaxed atomic increment alongside each store (or a `publish_snapshot()` helper wrapping store+bump) suffices. `tier_version` and `structural_epoch` are already monotonic counters — this makes the third key component match their guarantee.

### Why an issue rather than a PR

The change is mechanical but touches ~13 store sites across `table.rs`/`delete/sink.rs`, which are under active rework (#11249 → #11278 → #11295 this week) — filing so it can be folded into that ongoing work rather than conflict with it. Found during a data-correctness audit of #11249; analysis-confirmed, not reproduced (reproduction requires allocator address reuse, which is why it would be a heisenbug in the wild).

claudespice:
## Summary (root cause)

The per-scan deletion memos introduced for the merge-on-read scan path —
`MergedScanDeletions` (`merged_scan_deletions`) and `MemTierVisibleMemo`
(`mem_tier_visible_memo`) — keyed the **file-side** deletion index by its raw
`Arc::as_ptr` (`file_index_ptr`) and did **not** retain the index they were
built from. Several deletion-publish paths (`DELETE FROM` sink, deletion-cache
updates, tombstone upgrades, staged-CDC finalize, retention sweeps) replace the
file-side index without bumping either of the memo's other two key components
(`tier_version`, `structural_epoch`).

Once the old index generation is freed, the allocator can hand its address to a
later generation (same type/size class, LIFO reuse). The memo then sees the
same pointer + same `tier_version` + same `structural_epoch` → a **false hit** →
the scan filters with a **stale merged deletion view**, and rows deleted by the
most recent publishes become visible again. Silent wrong results,
allocator-dependent (a heisenbug in the wild). Reported and analysis-confirmed
in #11303.

## Changes

- `MergedScanDeletions` / `MemTierVisibleMemo` now **retain the source
  `PkDeletionSnapshot`** (`file_index`) instead of caching only a `usize`
  pointer. The memo key's file-side identity is read from that retained
  snapshot via a `file_index_ptr()` accessor.
- All three memo-store sites and three memo-hit checks updated accordingly; the
  append-path lockstep extension carries the retained snapshot forward (the
  file-side index is unchanged by a tier append).
- Regression test `deletion_memos_pin_file_index_against_aba` asserting the
  retention invariant through the production scan path.

## Why retention rather than the suggested generation counter

The issue suggested a monotonic generation counter bumped at every
`deletion_snapshot.store(...)`. On current `trunk` that is **15+ scattered
store/rcu sites** across `table.rs`/`delete/sink.rs`, and *omitting a single
bump silently re-introduces this exact row-resurrection bug*. Retention fixes
the root cause the issue names first ("the memo does not retain the index it was
built from") at just the two memo structs plus their build sites, and is
**immune to every mutation site**: while a memo is live its source `Arc` cannot
be freed, so its address can never be reused by a different generation and a
pointer match can only mean "same generation". The referenced rework
(#11249 → #11278 → #11295) has since merged, so there is no longer a conflict
concern with those sites. Memory cost is bounded — one extra deletion-index
generation pinned while a memo is live, released on the next rebuild or the
existing `store(None)` invalidations. (Happy to switch to the counter approach
if maintainers prefer it.)

## Performance impact

Scan hot path unchanged: a memo hit is still an `Arc`-pointer comparison; the
build path adds one `Arc` refcount clone (the retained snapshot) — the pre-fix
code already performed one `pk_deletion_snapshot()` clone per scan, so this is
neutral-to-better (the miss path now reuses that clone).

## Test plan

- `make lint`
- `make test`
- New `cargo test -p cayenne --lib deletion_memos_pin_file_index_against_aba`
  plus the existing memo tests (`merged_scan_deletions_memo_reuses_and_invalidates`,
  `mem_tier_visible_memo_reuses_and_invalidates`,
  `mem_tier_seal_pipeline_preserves_scan_memos`) pass.

## Checklist

- [x] Fix is minimal and scoped to the reported bug class
- [x] Regression test fails on pre-fix code (memo held no `Arc` → count unchanged) and passes on the fix
- [x] `make lint-rust-fix PACKAGES=cayenne` clean (clippy pedantic, lib + tests)
- [x] No behavior change on the default/existing path

Fixes #11303

### DIFF
--- a/crates/cayenne/src/provider/on_conflict.rs
+++ b/crates/cayenne/src/provider/on_conflict.rs
@@ -592,13 +592,35 @@ pub(crate) enum PkDeletionSnapshot {
 /// One memoized merged (file ∪ mem-tier) deletion snapshot for the scan path.
 /// See the `merged_scan_deletions` field docs for the key's torn-state proof.
 pub(crate) struct MergedScanDeletions {
-    /// `Arc::as_ptr` identity of the FILE-side index the merge was built from.
-    pub(crate) file_index_ptr: usize,
+    /// The FILE-side deletion snapshot the merge was built from, RETAINED for
+    /// the lifetime of the memo.
+    ///
+    /// The memo key's file-side identity is the `Arc::as_ptr` of this snapshot's
+    /// index ([`Self::file_index_ptr`]). Holding the `Arc` here — not merely its
+    /// raw address — is load-bearing for correctness: a live `Arc` allocation
+    /// can never be freed and its address handed to a *different* index
+    /// generation, so a pointer match proves the live file-side index IS this
+    /// same generation, never a coincidental ABA alias. Without the retention a
+    /// deletion publish could free this index and a later publish reuse its
+    /// freed address, yielding a false memo hit that serves a stale merged view
+    /// and resurrects deleted rows (#11303).
+    pub(crate) file_index: PkDeletionSnapshot,
     /// [`crate::provider::mem_tier::MemTier::version`] of the tier merged in.
     pub(crate) tier_version: u64,
     /// Structural epoch observed when the memo was built.
     pub(crate) structural_epoch: u64,
     pub(crate) merged: PkDeletionSnapshot,
+}
+
+impl MergedScanDeletions {
+    /// `Arc::as_ptr` identity of the retained file-side index — the memo key's
+    /// file-side component. Read from [`Self::file_index`] so the compared
+    /// pointer and the pinned allocation can never diverge. Always `Some` in
+    /// practice: the memo is only stored once the source index has an identity
+    /// (`PositionBased` returns before the store).
+    pub(crate) fn file_index_ptr(&self) -> Option<usize> {
+        self.file_index.index_ptr()
+    }
 }
 
 /// PK membership of a mem-tier checkpoint's flushed corpus (the visible inline +
--- a/crates/cayenne/src/provider/table.rs
+++ b/crates/cayenne/src/provider/table.rs
@@ -917,7 +917,10 @@ pub(crate) async fn reserve_sequences_in(
 /// Keyed the SAME way as [`MergedScanDeletions`] — (file-index `Arc` ptr, the
 /// per-shard content `version` hash, structural epoch) — so any concurrent
 /// append/clear/publish forces a rebuild and a stale pairing can never be
-/// served. Where `merged_scan_deletions` memoizes the merged tombstone
+/// served. Like `MergedScanDeletions`, the file-index `Arc` is RETAINED here
+/// (`file_index`) so its address cannot be freed and reused by a later index
+/// generation, keeping the ptr comparison ABA-safe (#11303). Where
+/// `merged_scan_deletions` memoizes the merged tombstone
 /// *snapshot*, this memoizes the merge-on-read *output*: the visible batches
 /// after `filter_inlined_batch_for_deletions` has already run.
 ///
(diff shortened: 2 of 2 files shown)


claudespice:
@copilot review

phillipleblanc (approved):

PR #11742, opened 2026-07-09
Extended schema inference for mysql

krinart:
## 📝 Summary

Implements `schema_inference: extended` for the MySQL connector, mirroring Postgres. The connector now queries `information_schema` for a table's primary key (in key order) and rough row/byte estimates and merges them into the schema metadata, so accelerated datasets that don't declare a `primary_key` can still use `refresh_mode: changes` (CDC) — previously this failed with "no primary key available". Adds an integration test covering CDC-without-declared-PK plus a standard-inference control.

Verified in https://github.com/spiceai/spiceai/actions/runs/29043626106


crates/runtime/tests/mysql/mod.rs
```diff
@@ -33,6 +33,8 @@ mod federation;
 pub mod replication;
 #[cfg(feature = "duckdb")]
 pub mod replication_e2e;
+#[cfg(feature = "duckdb")]
```
bjchambers: Why are the integration tests requiring duckdb? Should we be using something else (arrow acceleration and/or cayenne) for the tests?

krinart: Food point. Replaced with in-memory arrow

sgrebnov (approved):

PR #11768, opened 2026-07-09
perf: avoid unnecessary allocations on query hot paths

phillipleblanc:
## Summary

Remove avoidable allocations on the SQL query hot path (distinct from search-path work in #11767):

1. **Move plans/params by value** — `run_internal` / `submit_distributed_internal` take ownership of `QueryMethod` so pre-parsed `LogicalPlan`s and `ParamValues` are moved into planning instead of deep-cloned (important for FlightSQL read-only paths that already parsed once).
2. **`/v1/sql` Arc-first SQL** — decode the request body into `Arc<str>` and pass it through `QueryBuilder::new_arc`, avoiding an extra SQL string copy before execution.
3. **Reuse datasets telemetry label** — build the sorted `datasets` label once in `QueryTracker::finish` and reuse it for OTel metrics and task-history logging.

## Related

Hot-path allocation audit in `~/code/spiceai/five` (query path). Does not overlap with #11767 (search cache / keyword ILIKE).

## Test plan

- [x] `make lint-rust`
- [ ] CI green
- [ ] Spot-check `/v1/sql` text/plain and JSON parameterized queries still work

### DIFF
--- a/crates/runtime/src/datafusion/query.rs
+++ b/crates/runtime/src/datafusion/query.rs
@@ -607,8 +607,10 @@ impl Query {
         // Get the session state for planning
         let session = session_ctx.state();
 
-        // Get logical plan and cache key, reusing existing cache infrastructure
-        let (plan, mut tracker, cache_key) = match &self.sql {
+        // Get logical plan and cache key, reusing existing cache infrastructure.
+        // Match by value so a pre-parsed plan / parameters can be moved rather
+        // than deep-cloned on the distributed submit path.
+        let (plan, mut tracker, cache_key) = match self.sql {
             QueryMethod::Text {
                 sql,
                 parameters,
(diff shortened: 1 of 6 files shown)


crates/runtime/src/datafusion/query.rs
```diff
@@ -1371,7 +1375,7 @@ impl Query {
                 let final_stream = attach_cancellation_to_stream(
                     final_stream,
                     query_cancel_token.clone(),
-                    query_id_str.clone(),
+                    Arc::clone(&query_id_str),
```
bjchambers, line 1378: I wonder if we should have general repository level guidance encouraging putting strings into the arc *sooner*. I've seen this here and my PRs for metrics and other things. Doesn't need to change in this PR, but question for future.

bjchambers (approved):

sgrebnov (approved):

crates/runtime/src/datafusion/query.rs
phillipleblanc: Yeah, we had done a better job of this previously, but we should probably have specific review guidance for this.

PR #11772, opened 2026-07-09
feat(cayenne): datalake keyset-rebuild bloom + physical GC of superseded files

sgrebnov:
Stacked on #11745 (incremental datalake promotion) — base is that branch, so this diff is only the bloom + GC changes.

## Keyset-rebuild bloom
Per-file PK existence bloom in the datalake manifest (`pk_bloom_blob`), built at promotion by reading back each file's PK column. The CDC-upsert keyset rebuild unions the live files' blooms instead of scanning the datalake, removing the O(cold-rows) object-store read. Exact-scan fallback for `DoNothing` tables and files without a bloom; a per-file cap bounds the manifest.

## Physical GC of superseded objects
Periodic mark-and-sweep on the background tier tick: deletes objects no longer referenced by the manifest (the GC root) once observed orphaned for >= `cayenne_datalake_gc_interval_ms` (default 5min, internal expert override). Self-healing across restarts — re-derives orphans from `store - manifest`, so a promotion-then-restart still collects them — and never removes a file an in-flight scan built before the orphaning could still read.

## Tests
Bloom serialize round-trip, manifest `pk_bloom` SQL round-trip, and the pure mark-and-sweep GC decision (mark / grace / restart-reset / prune). cayenne lib green (865 tests); `make lint-rust-fix PACKAGES=cayenne` clean.


crates/cayenne/src/metastore/turso.rs
```diff
@@ -371,6 +371,7 @@ impl TursoMetastore {
             min_sequence BIGINT NOT NULL DEFAULT 0,
             max_sequence BIGINT NOT NULL DEFAULT 0,
             statistics_blob BLOB NOT NULL,
+            pk_bloom_blob BLOB,
```
sgrebnov, line 374: All schema changes are safe as cold tier not yet shipped/released

PR #10853, opened 2026-05-15
Refactor: add explicit ParameterSpec for embedding components

ManishSharma1609:
- Create crates/runtime/src/embeddings/params/ with per-provider ParameterSpec definitions for all 8 embedding providers (openai, azure, google, huggingface, databricks, bedrock, file, model2vec)
- Update try_to_embedding to use Parameters::try_new with explicit specs
- Update all provider function signatures from HashMap to Parameters
- Remove unused extract_secret macro from embed.rs

Closes #6755

## Summary
Replicates the approach from #6199 for embedding components. Adds explicit `ParameterSpec` definitions for all 8 embedding providers instead of the free-form `HashMap`, and wires them through `Parameters::try_new` in `try_to_embedding`.

## Related
Closes #6755

## Breaking Changes
None, existing embedding configurations continue to work. Unknown parameters will now emit a warning log instead of being silently ignored.

## Docs
No doc changes required.

## Notes for Reviewers
Tests added for all 8 embedding provider ParameterSpec definitions. 
Followed the same pattern used in #6199 for model components.

---
<a href="https://app.blacksmith.sh/spiceai/codesmith/spiceai/pr/10853"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

PR #11787, opened 2026-07-10
fix(embeddings): restore params broken by #10853

Jeadie:
## Summary

PR #10853 added explicit `ParameterSpec` definitions for all 8 embedding providers but introduced two classes of bugs:

**1. Bedrock embeddings are non-functional (credentials silently dropped)**

`BEDROCK_PARAMETERS` was missing all 5 AWS credential/config params (`aws_region`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token`, `aws_iam_role_source`). The `bedrock()` function passes these to `create_bedrock_client` via `params.get_runtime_params()`, which only returns `ParameterType::Runtime` params. Since they weren't in the spec, `Parameters::try_new` emitted a warning and dropped them — so the Bedrock client always received empty credentials.

**2. Panic on absent params + silent drops for HuggingFace / Google / Model2Vec / Bedrock**

`Parameters::get(name)` calls `describe(name)` even when the param is absent, and `describe` panics if `name` isn't in the spec. Several params that `embed.rs` accesses via `params.get(...)` were either not in the spec at all (`truncate` alias for Bedrock — causing an unconditional panic) or were misclassified as `ParameterSpec::component` instead of `runtime`.

The `component` type requires a provider prefix (e.g. `huggingface_hf_token`), but existing user configs and the docs use bare names (`hf_token`). With the prefix requirement enforced by the new validator, those params were silently rejected and then panicked on access.

Affected: `hf_token` (HuggingFace + Model2Vec), `dimensions` (Google + Bedrock), `normalize` / `subfolder` (Model2Vec), and all Bedrock model-specific params.

## Changes

- **`bedrock.rs`**: add `aws_region`, `aws_access_key_id`, `aws_secret_access_key`, `aws_session_token`, `aws_iam_role_source` as `runtime` params; change model-specific params (`dimensions`, `normalize`, `truncate_mode`, `input_type`, `embedding_purpose`) from `component` to `runtime`; add `truncate` as a deprecated `runtime` alias for `truncate_mode`
- **`huggingface.rs`**: change `hf_token` from `component` to `runtime`
- **`google.rs`**: change `dimensions` from `component` to `runtime`
- **`model2vec.rs`**: change `hf_token`, `subfolder`, `normalize` from `component` to `runtime`
- **`params/mod.rs`**: replace name-existence unit tests with end-to-end roundtrip tests — each parses a YAML `Params` string, builds a `Parameters` object, then calls every `params.get(key)` the provider function calls, catching both panics and silent drops

## Breaking Changes

None — restores the parameter names that existed before #10853 and that the docs describe.


Jeadie:
Closing — replaced by a clean branch with only the fix commit.

Issue #11730, opened 2026-07-08
ci(benchmarks): provision tpch_sf1 on benchmarking-postgres (postgres data + catalog benches unreachable)


bjchambers:
## Summary

The TPC-H (and TPC-DS) benchmark e2e tests for the **PostgreSQL data connector and PostgreSQL Catalog Connector** cannot run because the benchmark PostgreSQL instance is unreachable / has no `tpch_sf1` database provisioned.

## Evidence

Recent `testoperator_run_bench.yml` runs on 2026-07-08:

- `bench - tpch - federated/postgres[catalog].yaml` — **failure**
- `bench - tpch - federated/postgres.yaml` — **failure**
- `bench - tpcds - federated/postgres.yaml` — **failure**

All fail with:

```
Cannot connect to ... benchmarking-postgres-rw.dataplatform.svc.cluster.local:5432.
Ensure that the host and port are correctly configured ... and that the host is reachable.
```

By contrast, `mysql[catalog]`, `mssql[catalog]`, and `oracle[catalog]` benches pass — their benchmark databases are provisioned.

The `postgres[catalog]` TPC-H bench has **never** had a green run (only two attempts exist, both on 2026-07-08, both failed on connectivity).

## Impact

Blocks the [Alpha Catalog criterion](https://github.com/spiceai/spiceai/blob/trunk/docs/criteria/catalogs/alpha.md): "TPC-H queries pass with a success rate equal to the native connector." The test harness and all 22 query overrides (`postgres-catalog`) are in place; only the benchmark data source is missing. (Verified locally against a Dockerized Postgres with TPC-H sf1 loaded — the catalog path passes; see PR/verification linked from #10373.)

## Ask

Provision a reachable `benchmarking-postgres` with the TPC-H sf1 (and TPC-DS sf1) datasets loaded, mirroring the MySQL/MSSQL/Oracle benchmark setup, and confirm the `POSTGRES_HOST` / `POSTGRES_PASSWORD` CI secrets point at it. Then re-dispatch the `postgres[catalog]` TPC-H bench to establish a green baseline.

PR #11800, opened 2026-07-10
test: redact connection context in federated explain snapshots

bjchambers:
## Problem

Federated connectors embed the live connection string of the source database in their physical plan:

```
VirtualExecutionPlan name=postgres compute_context=host=Tcp("<host>"),port=5432,db=<db>,user=<user>, base_sql=SELECT …
```

Because that per-environment `host/port/db/user` string is part of the snapshotted text, the explain-plan validation only passed when the test ran against the *exact* endpoint the snapshot was captured on. The same, byte-for-byte-identical plan run against any other Postgres/MySQL endpoint (a local Docker instance, a different or renamed benchmark host, a read vs read-write replica) failed the snapshot on nothing but the embedded connection metadata. The connection context is *connection metadata*, not plan structure, so it shouldn't be part of a plan-shape snapshot.

## Fix

- Add a redaction filter to `build_explain_filters` (`crates/test-framework/src/snapshot/mod.rs`) that normalizes the connection context to a constant `compute_context=<CONNECTION>` token. Filters run on the raw `RecordBatch` **before** `pretty_format_batches`, so column widths are computed from the redacted text and the whole table re-pads consistently (mirroring the existing `partition_sizes` / `file_groups` / temp-dir redactions).
- The pattern is anchored on `,port=<n>` and forbids spaces per field so it can't run past the connection context into the pushed-down SQL; `db`/`user` and the trailing comma are optional to cover the Postgres (`host=Tcp("…")`, trailing comma), MySQL (bare host), and no-`db` forms.
- Regenerate the **907** affected Postgres/MySQL explain snapshots with the filter applied.

## How the snapshots were regenerated (deterministic, no benchmark infra)

Rather than block on the benchmark databases (gated on #11730), regeneration was done deterministically: a throwaway harness reconstructed each committed table and re-rendered it through the **real** snapshot pipeline (`sanitize_record_batches` → `pretty_format_batches` → `sort_partitioned_union_children`). Before rewriting anything, the harness was proven **content-lossless across all 10,208 committed explain snapshots** (0 unparsed, 0 content changes with filters disabled), so the only thing regeneration changes is the intended redaction plus consistent table re-padding. It then rewrote only the 907 host-bearing files. The harness is intentionally **not** checked in.

## Validation

- `test_connection_context_filter` unit test covers the Postgres/MySQL forms, idempotency, and the no-false-match case.
- **Live spot-check** with a current-trunk `spiced` build against local Docker Postgres and MySQL (both non-benchmark hosts): each live `compute_context` redacts to the exact committed `<CONNECTION>` token — the host-dependence is eliminated for both connectors.

## Known caveat (out of scope)

The live spot-check also surfaced **pre-existing plan-shape drift** unrelated to this change: current trunk emits a `base_sql=…` / `SchemaCastScanExec → CooperativeExec → BytesProcessedExec` shape, whereas the committed snapshots still carry the older `sql=… rewritten_sql=…` form. The committed explain snapshots are stale in *shape* because the full bench regen is gated on #11730. This PR deliberately changes **only** the connection-context redaction; the shape drift will be resolved by the gated bench regen — and with this filter in place, that regen's host will no longer cause diffs.

Scoped `cargo clippy -p test-framework --tests` + unit tests are green; the full `make lint-rust` gate was not run.


phillipleblanc (dismissed):

phillipleblanc (dismissed):

bjchambers:
Updated to latest `trunk` (clean merge, no conflicts) and verified the redactions still hold.

### Redaction state after the merge

- `trunk` touched no explain snapshots under `crates/test-framework/`, so the merge brought in nothing that needed re-redacting.
- No unredacted `compute_context=host=…` remains in `crates/test-framework/src/snapshot/snapshots/explain/` — all 907 carry `<CONNECTION>`, with no partial-match leftovers.
- `trunk` added new federated Postgres pods (`test/spicepods/tpch/sf10/federated/postgres.yaml`, `sf100/…`). They have no snapshots yet; whatever they capture will be redacted by the filter on first run.
- The remaining `compute_context=host=` hits are in `crates/runtime/tests/snapshots/`, which use local container endpoints via a different snapshot path (not `build_explain_filters`). Deterministic, so left alone.

I also re-verified that the 907-file snapshot diff is **redaction-only**: normalizing both sides and diffing, every change reduces to the new `<CONNECTION>` token, table column re-padding from the shorter string, and `<GROUPING_PAIR>` — that last one being a filter that already existed at the merge base and had simply never been applied to 9 stale tpcds q36/q70/q86 snapshots, which regeneration caught up. No plan structure changed anywhere.

### Two `scripts/signoff` fixes came along

Running sign-off on this branch surfaced two latent bugs, both needing a large branch diff to trigger — this branch's ~909 changed files / ~107 KB list is the first thing to hit them. They're in-scope only because they blocked signing this branch off, and the second one is a correctness issue worth landing on its own merit.

**1. `has_content` — the whitespace check was superlinear (`fix(ci): make signoff's whitespace check linear…`)**

`[[ -n "${var//[$'\t\n\r ']/}" ]]` is quadratic-or-worse in bash 3.2, which is still `/bin/bash` on macOS and what `#!/usr/bin/env bash` resolves to there. Measured on the real changed-file list:

| list | bytes | time |
|---|---|---|
| 25 files | 3 KB | 1.8 s |
| 50 files | 6 KB | 13.7 s |
| 100 files | 12 KB | 91 s |
| 909 files | 107 KB | ~11 h (extrapolated) |

That matches an observed 10.5-hour hang at 100% CPU in bash with no child process. Replaced all nine sites with a `has_content` helper that glob-matches for a single non-separator character: it short-circuits on the first content byte and stays ~20 ms at 107 KB even in the all-whitespace worst case. Verified equivalent to the old expansion across empty, all-separator, and mixed inputs. Note `branch_has_rust_changes` runs *before* the `SIGNOFF_SKIP_TARGETED_LINT` check, so no env var could work around it.

**2. SIGPIPE made sign-off skip all Rust checks and still pass (`fix(ci): stop signoff skipping Rust checks…`)**

This is the one that matters. `branch_has_rust_changes` piped the changed-file list into `grep -q`. `grep -q` exits at the first match, so once the list exceeds the 64 KB pipe buffer the pipe closes while `printf` is still writing — `printf` dies of SIGPIPE and `set -o pipefail` reports the pipeline as failed (status 141).

The function then answered "no Rust changes" for a branch that *does* change `.rs` files, and sign-off **skipped lint, build, and unit tests while still posting a green `signoff` status**. Reproduced directly on this branch (`mod.rs` matches on the very first line): piped form misses, here-string form matches. It needs both a >64 KB diff and an early match, which is why smaller branches were unaffected. Fixed by feeding `grep` a here-string so there is no pipe to break — the same hazard is already called out for `| head -n1` further down the script.

Post-fix, sign-off correctly reports `Rust-affecting files present — will run Rust checks` and derives `Targeted pre-lint for changed packages: test-framework`.

One consequence worth flagging: the attestation posted against `ec9798ffa` was produced under bug 2 and skipped every check, so it should not be trusted. It is not `HEAD` anymore; the current attestation is being produced remotely against `4683ed1`.

### Follow-up, not addressed here

The same environment-coupling this PR fixes still exists in other `compute_context` shapes the `host=…,port=…` pattern doesn't match — several other federated connectors embed endpoint URLs and account/user metadata in their plan output, and in one case the same query is snapshotted under two different sets of connection metadata. Extending to those means regex-editing a few hundred `.snap` files rather than regenerating them, so I kept this PR at its stated scope. Worth a separate change.

lukekim (dismissed):

krinart (approved):

PR #11820, opened 2026-07-12
fix(cayenne): partitioned dataset never becomes ready after a large initial refresh (fixes #11818)

### LINKED ISSUE #11818: Cayenne accelerator with `partition_by` never becomes ready after a large initial refresh (regression from #11144)
## Symptom

A dataset accelerated with `engine: cayenne` that has `partition_by` configured never becomes ready when the initial refresh is non-trivial in size. All datasets report loaded — the runtime logs `Dataset load summary (after 30s): N/N ready, 0 unhealthy, 0 still initializing` — but `/v1/ready` never returns ready. Reproduced with readiness waits of 30s, 300s, and 900s (15 minutes): the endpoint never flips.

Small refreshes are unaffected, and non-partitioned cayenne datasets are unaffected. This has been failing consistently in benchmark CI since ~June 4.

## Reproduction

TPC-H SF1 `customer` table (~1.5M rows) as a local parquet file:

```yaml
version: v1
kind: Spicepod
name: cayenne-partition-ready-repro
datasets:
  - from: file:///path/to/tpch_sf1/customer.parquet
    name: customer
    acceleration:
      enabled: true
      engine: cayenne
      mode: file
      partition_by:
        - bucket(5, c_name)
```

Observed behavior by refresh size (same spicepod, varying `refresh_sql`):

| `refresh_sql` | Result |
|---|---|
| `select * from customer limit 2` | ready |
| `select * from customer limit 1000000` | **never ready** |
| (none — full table) | **never ready** |

## Bisect / root cause analysis

Last-good → first-bad window on trunk pointed at **#11144 "feat(cayenne): sharded parallel Vortex encode with key/time clustering" (`2c366009ff`)**, and the size-dependence matches its shard-count gate exactly:

- `snapshot_shard_count()` (`crates/cayenne/src/provider/table.rs`) selects the new sharded write path when `estimated_bytes / max(16 MiB, target_file_size/16) > 1` — i.e. writes over roughly 32 MiB of in-memory Arrow. Below that, `write_shard_config()` returns `None` and the pre-#11144 single-writer path is used (which still works — hence `limit 2` passing).
- When partition columns are present, the vortex sink then **forces `num_shards = 1`** (`crates/vortex/src/persistent/sink.rs`, "preserve path / sub-directory contracts"). So "sharding requested but partition-forced back to a single shard" is the exact combination that hangs.
- The write appears to complete (files written, datasets report ready) but overall readiness never flips — suspicion is the new visibility watermark (`published_inlined_seq` / `scan_state_lock` publish path introduced in the same PR) never advances for this combination, leaving the accelerated table's initial-refresh completion signal unfired, so the component never reaches Ready under the default `ready_state: on_load`.

## Confirmation

Setting the per-table write concurrency to 1 — which makes `snapshot_shard_count()` return 1 and bypasses everything #11144 added to the write flow — fixes readiness for the same 1M-row refresh:

```yaml
      params:
        cayenne_write_concurrency: 1
```

With this single param the full-size refresh becomes ready normally. This also serves as a workaround for anyone hitting this.

## Expected behavior

Partitioned cayenne datasets should become ready regardless of initial refresh size, or at minimum surface an error instead of hanging readiness indefinitely.

claudespice:
## Summary

Fixes #11818. A dataset accelerated with `engine: cayenne` and `partition_by` configured never became ready when its initial refresh was non-trivial in size (> ~32 MiB): every dataset reported loaded, but `/v1/ready` never flipped (reproduced at 30s/300s/900s waits). Small refreshes and non-partitioned cayenne datasets were unaffected. Setting `cayenne_write_concurrency: 1` worked around it. Regression from #11144 (sharded parallel Vortex encode).

**Root cause — a hold-and-wait deadlock on the global encode-permit budget.**

The partitioned-refresh coordinator (`CayennePartitionedOverwriteSink` / `CayennePartitionedAppendSink`) reads one input stream and fans it out to per-partition writer tasks over **bounded** channels (`PARTITION_WRITER_CHANNEL_DEPTH = 8`), blocks on `sender.send().await`, and only closes the senders (and awaits the writer tasks) **after** the whole input is drained. Each per-partition write (`begin_overwrite` / `begin_staged_append` → `write_to_snapshot`) acquires up to `min(write_concurrency, target_partitions)` encode permits from the process-global budget (`write_budget.rs`) **before** it starts draining its own input channel, and holds them until the shared input ends.

The budget's deadlock-freedom argument assumes writes are *independent* (each acquires, encodes, releases without waiting on another write). A partitioned refresh violates that: once enough partitions are in flight to exhaust the budget (the maintenance gate for overwrite, the main semaphore for append), the next partition blocks acquiring permits → never drains its channel → the coordinator backpressures on that partition's full channel → never closes the senders → the permit-holding partitions never see end-of-input → never release. The dataset's initial-refresh completion signal never fires, so it never reaches Ready.

Size dependence: a small refresh spreads across too few partitions (and fits in the channel buffers), so the group fits within the budget; a large refresh fills all partitions and deadlocks. `cayenne_write_concurrency: 1` avoided it by dropping each partition's demand to a single permit.

## Fix

Per-partition writes are marked **co-dependent** (`CayenneTableProvider::with_co_dependent_write`) by the coordinator. A co-dependent write:

1. **Encodes single-shard** (`snapshot_shard_count` returns 1) — the coordinator already runs one writer per partition concurrently, so the cross-partition fan-out *is* the encode parallelism; sharding each partition on top only multiplies the group's global permit demand by the partition count.
2. **Never blocks on the encode budget** — it uses a new non-blocking `try_acquire_encode_permits` (`try_acquire_many_owned`): it takes a permit only if one is immediately available, otherwise proceeds ungated (single-shard, so at most one extra encode). A non-blocking acquire cannot join a hold-and-wait cycle, so the awaited-together partition group is deadlock-free for **any** partition cardinality — not just up to the budget size.

Non-partitioned writes are completely unchanged (the flag defaults to `false`): same blocking acquire, same size-aware sharding, same global throttling and EBS-bandwidth cap.

## Changes

- `crates/cayenne/src/provider/write_budget.rs`: add `try_acquire_encode_permits` / `try_acquire_from` — non-blocking acquisition for co-dependent write groups, mirroring `acquire_from`'s clamp + gate-then-main order (no partial-reservation leak on failure).
- `crates/cayenne/src/provider/table.rs`: add the `co_dependent_write` field + `with_co_dependent_write()`; `snapshot_shard_count` returns 1 and `write_to_snapshot` uses the non-blocking acquire when set.
- `crates/runtime/src/dataaccelerator/cayenne/partitioned_insert_strategy.rs`: mark each per-partition writer clone co-dependent in both the overwrite and append coordinators.

## Test plan

- `make lint`
- `make test`

New tests:
- `write_budget`: `try_acquire_returns_none_when_budget_exhausted` (non-blocking where the blocking path queues forever), `try_acquire_maintenance_gate_no_partial_leak` (no gate-permit leak when the main semaphore is the exhausted one), `try_acquire_takes_permits_when_available`.
- `table`: `co_dependent_write_forces_single_shard` — a co-dependent clone forces single-shard (`snapshot_shard_count == 1`, `write_shard_config == None`) for a write that would normally earn the full write-concurrency fan-out.

Note: the two halves of the fix (single-shard demand + non-blocking admission) are unit-tested deterministically; the end-to-end partitioned-refresh readiness is covered by the benchmark suite (which has been red on this since ~June 4).

## Cayenne doc review

`docs/cayenne/cayenne.md` does not document the global encode budget or the partitioned co-dependent write path, and this change alters no documented behavior, config parameter, or metastore schema — so no `cayenne.md` update is warranted.

## Checklist

- [x] Root-caused the regression to #11144 and the co-dependent hold-and-wait cycle
- [x] Fix is deadlock-free for any partition cardinality (non-blocking admission), not just a threshold bump
- [x] Non-partitioned write behavior unchanged (flag defaults to false)
- [x] Unit tests for both fix halves; `cargo check`/`clippy` clean on `cayenne` and `runtime`
- [x] Reviewed whether `docs/cayenne/cayenne.md` needs updating

Fixes #11818


claudespice:
@copilot review

PR #11825, opened 2026-07-13
fix(cayenne): partitioned datasets deadlock against the global encode budget and never become ready

### LINKED ISSUE #11818: Cayenne accelerator with `partition_by` never becomes ready after a large initial refresh (regression from #11144)

Jeadie:
## Bug

A cayenne dataset with `partition_by` never becomes ready when its initial refresh is non-trivial in size (roughly ≳32 MiB). The runtime reports every dataset loaded and healthy, but `/v1/ready` stays 503 forever — reproduced with readiness waits up to 900s, and reproduced locally (TPC-H SF1 `customer`, `bucket(5, c_name)`, 1M-row `refresh_sql`): spiced sits at 0% CPU with the refresh permanently stuck. Every partitioned-cayenne benchmark CI run has failed this way since #11144 merged (~Jun 4). Small refreshes and non-partitioned tables of any size are unaffected.

Fixes #11818.

## Root cause

The partition router and the encode budget deadlock each other. A partitioned refresh works like a conveyor belt with a router at its head. The refresh reads the source as one stream; the router hashes each batch by the partition key and drops the pieces into small bounded bins, one per partition. Behind each bin, a writer for that partition pulls rows and encodes files. The writers run concurrently (cross-partition concurrency is where its improved indexing throughput is from). 

Separately, #11144 introduced a process-global vortex compression encoding budget: a fixed pool of encode slots (sized from the host's cores) that a write must claim. It must claim all of its slots, atomically, before consuming any input, and then release once it has finished writes. The budget's deadlock-freedom argument was that atomic all-or-nothing acquisition prevents hold-and-wait cycles between writes.

That argument assumes writes are **independent**. Partition writes are not:

1. A large refresh makes each partition writer want several slots (the size gate in `snapshot_shard_count`; an opaque stream defaults to the full per-table write concurrency). The first few writers drain the pool.
2. A later writer finds no slots and parks in line, before it has pulled anything from its bin.
3. Hash partitioning spreads rows evenly, so the router soon produces rows for the parked writer's partition. Its bin fills. The router stops, mid-hand-off, and can no longer feed anyone.
4. The slot-holding writers finish their bins and wait for more input that will never arrive. A partition write only ends when the router signals end-of-stream, so they hold every slot, forever.

## Fix

The invariant the deadlock violates: `a writer that someone stands behind must never park.` The fix makes exactly that set of writers, and no other, unable to park, in two coordinated parts:

1. Partition child tables are marked as coupled writers, and coupled writers bypass the budget. A writer that cannot park cannot freeze the router.

2. Partition child tables are clamped to serial writes (1 encoding shard; the pre-#11144 baseline ). An operator-pinned `cayenne_write_concurrency > 1` is ignored (clamped, with a warning).

Known accepted trade-off: a high-partition-count refresh runs partition-count serial encoders unmetered (identical to pre-#11144 behavior). This degrades throughput on small hosts but never liveness-threatening. The composable follow-up, if wanted, is capping concurrently-open partition writers in the partition insert layer, which owns the bins and can own the buffering decision that cap forces.

### DIFF
--- a/crates/cayenne/src/provider/write_budget.rs
+++ b/crates/cayenne/src/provider/write_budget.rs
@@ -41,13 +41,15 @@ limitations under the License.
 //! (`runtime_table_partition::insert`), so a child write parked on this budget
 //! stalls the demux, which starves the permit-holding sibling writes of input —
 //! a hold-and-wait cycle *through the channels* that left partitioned tables
-//! permanently unready (spiceai/spiceai#11818). To keep such coupled writers
-//! live regardless of partition count, **single-shard writes are exempt from
-//! the budget** ([`acquire_for_write`]): a serial write is the pre-budget
-//! baseline the budget was never needed for (it exists to cap encode *fan-out*,
-//! and aggregate compaction concurrency is separately bounded by the
-//! compactor's own semaphore), and an exempt serial write can always drain its
-//! input, so the demux always makes progress.
+//! permanently unready (spiceai/spiceai#11818). Such COUPLED writers — tables
+//! whose writes someone else stands behind, marked at construction via
+//! `CayenneContext::new_for_partition_child` — are therefore **exempt from the
+//! budget** ([`acquire_for_write`]): an exempt write can always drain its
+//! input, so the demux always makes progress, for any partition count. The
+//! exemption is safe because coupled tables are also forced to serial writes
+//! (one encode shard — the pre-budget baseline; the budget exists to cap
+//! encode *fan-out*), so an exempt child contributes the minimum possible
+//! encode footprint. Every uncoupled write, serial included, still queues.
 //!
 //! When unset (unit tests, embedders that don't wire it up) acquisition is a
 //! no-op and writes proceed ungated, preserving the prior per-table behavior.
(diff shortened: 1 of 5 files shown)


crates/cayenne/src/provider/write_budget.rs
```diff
@@ -301,17 +317,51 @@ pub fn encode_budget_snapshot() -> Option<EncodeBudgetSnapshot> {
...
+/// stream per writer, and compaction's aggregate concurrency stays bounded by
+/// the `BackgroundCompactor` semaphore independently of this gate. Uncoupled
+/// writes — including serial ones — queue exactly as before.
+async fn acquire_for_write(
+    budget: &EncodeBudget,
+    shards: usize,
+    class: WriteClass,
+    coupled_writer: bool,
+) -> Option<EncodePermits> {
+    if coupled_writer {
```
phillipleblanc, line 361: why pass this into here and not just return it in acquire_encode_permits?

crates/cayenne/src/provider/write_budget.rs
```diff
@@ -301,17 +317,51 @@ pub fn encode_budget_snapshot() -> Option<EncodeBudgetSnapshot> {
...
+        return None;
```
phillipleblanc, line 362: is bypassing the budget a safe thing to do?

crates/cayenne/src/provider/write_budget.rs
Jeadie: Primarily just for easier testing.

crates/cayenne/src/provider/write_budget.rs
Jeadie: Yes, in fact its not safe without this PR. #11144 introduces a global budget for vortex compression encoding (essentially 1 per cpu) to enable having many write shards. This caused correctness issues with partitioned datasets (read above). It is true that now a largely partitioned dataset will cause performance degradation (e.g. scheduling overhead, increased memory).

lukekim (approved):

PR #11826, opened 2026-07-13
perf(cayenne): light delta encoding + higher CDC coalescing defaults

bjchambers:
## What

Two performance defaults validated by the SF1000 CH-benCHmark HTAP sweep:

1. **`perf(cayenne)`: default `auto` delta encoding to light for all deltas.** `effective_level` now resolves every `auto` `WriteClass::Delta` write to the light scheme set (level 2: `Constant + Sparse + Dict`) regardless of estimated size, instead of size-gating large deltas up to the full BtrBlocks + FSST cascade. Every delta is a transient staged CDC write that compaction re-encodes at FULL before it becomes long-lived, so paying the per-file strategy search + FSST symbol-table double-train on the CDC hot path is wasted work. Maintenance writes still use the full cascade; `cayenne_delta_encoding: 7` is the explicit opt-out. Removes the now-dead size-gate machinery (`estimated_bytes`/`target_size_bytes` params, `AUTO_LIGHT_DENOMINATOR`).

2. **`perf(chbench)`: enable higher CDC coalescing in the adaptive pods.** Pin `cdc_max_coalesced_bytes` = 512 MB and add a `cdc_max_coalesce_age_ms` = 2000ms linger window to the 4-slot and 1-slot adaptive pods, on top of the deep prefetch already fixed there. Bigger coalesced bursts → fewer, larger applies → per-burst stats/checkpoint/publish amortize over more rows.

## Why / evidence

In the SF1000 HTAP sweep:

- **Light-always encoding** is what lets the CDC apply loop keep up: it *enables replication convergence* where the full-encode default fails to converge, and delivers the best analytic QPH on top of coalescing (encode frees cores for concurrent apply + queries). Inert at SF100 (checkpoints are already small/light there).
- **Coalescing** was the strongest single convergence lever: the 4-slot adaptive pod converged fastest (~801s) with coalescing on, where the no-coalesce default failed to converge at all. The linger deadline is anchored at the previous apply's start, so apply-bound tables barely wait; 2000ms bounds the added freshness latency for lighter tables.

## Scope / blast radius

- The encoding change is cayenne-internal and bounded: light deltas are larger transient files until compaction folds them (`cayenne_delta_encoding: 7` opts back into full).
- Coalescing is scoped to the benchmark pods (matching the deep-prefetch precedent) rather than the global CDC defaults, so it does not alter memory/freshness behavior for other connectors' CDC.

## Tests

- `cargo test -p cayenne --lib provider::delta_encoding` — 8/8 pass (rewrote the size-gate tests to assert light-for-all-deltas).
- Behavior validated end-to-end by the SF1000 CH-benCHmark HTAP runs.

### DIFF
--- a/crates/cayenne/benches/compaction_write_amplification.rs
+++ b/crates/cayenne/benches/compaction_write_amplification.rs
@@ -21,9 +21,9 @@ limitations under the License.
 //! Every row on this path is written to disk TWICE. A delta / mem-tier
 //! checkpoint (`WriteClass::Delta`) writes each protected snapshot with a LIGHT
 //! encoding — it skips the `BtrBlocks` per-column strategy search + FSST
-//! symbol-table training that dominate small-write encode cost (see
+//! symbol-table training that dominate encode cost (see
 //! `provider::delta_encoding`, `effective_level` returns `AUTO_LIGHT_LEVEL` for
-//! a small/unknown-size `auto` delta). Those light files are LESS compressed, so
+//! every `auto` delta, regardless of size). Those light files are LESS compressed, so
 //! they inflate on-disk bytes (read-amp in bytes) until compaction folds them.
 //! Compaction (`WriteClass::Maintenance`) then RE-ENCODES the merged corpus with
 //! the FULL cascade — the expensive strategy search + FSST it skipped — so the
--- a/crates/cayenne/src/metadata.rs
+++ b/crates/cayenne/src/metadata.rs
@@ -339,10 +339,9 @@ pub enum CompressionStrategy {
 ///
 /// zstd-style level scale (`cayenne_delta_encoding` param):
 ///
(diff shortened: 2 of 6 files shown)


phillipleblanc (dismissed):

bjchambers:
Addressed the review comment: `metadata.rs` user-facing `cayenne_delta_encoding` docs (enum doc, `Auto` variant, `delta_encoding` field, and the `Default` impl comment) now describe `auto` as light-for-all-deltas, matching the implemented behavior; also refreshed the stale "micro A/B neutral, to be validated" note to reference the SF1000 validation. Merged `trunk` in to clear BEHIND; `cargo check -p cayenne` is green.

lukekim (approved):

phillipleblanc (approved):

PR #11843, opened 2026-07-13
perf(chbench): CSV-based seed loading for MySQL and Postgres

bjchambers:
## Summary

CH-benCH MySQL seed jobs at SF1000 were slow enough to threaten (and in at least one observed run, exceed) the 120-minute CI job timeout, and a separate connection-limit issue could make them fail outright. This replaces the per-row `INSERT`/clone seed loader with a shared, parallelized CSV generator plus native bulk-load (`COPY` / `LOAD DATA LOCAL INFILE`) for both engines, and tunes MySQL's CI container for faster bulk writes.

- **`csv_gen`** (new, shared by both engines): generates seed data as CSV, parallelized across warehouses. Every warehouse now gets fully independent random data instead of cloning from 10 seed warehouses via `INSERT...SELECT` — no longer needed since generation is cheap and embarrassingly parallel (measured ~2 minutes for SF1000 locally, 18 cores).
- **Postgres loader**: bulk-loads via client-streamed `COPY ... FROM STDIN` (the same mechanism `psql \copy` uses).
- **MySQL loader**: bulk-loads via `LOAD DATA LOCAL INFILE` (`mysql_async`'s built-in `WhiteListFsHandler`).
- Both are client-streamed deliberately — see "Options considered" below.
- MySQL CI container: added `innodb_doublewrite=0`, `innodb_flush_method=O_DIRECT_NO_FSYNC`, `local_infile=1`.
- Capped the *default* OLTP terminal count at 100 (only when `--terminals` isn't passed explicitly) so an omitted flag can't reproduce the "Too many connections" failure seen on a manual SF1000 dispatch (`scale_factor * 10` = 10,000 terminals against `max_connections=200`).

Net effect: cuts the current ~75-minute SF1000 MySQL seed to an estimated ~12-15 minutes, restoring real margin against the CI timeout.

## Why this was slow, and why it degrades non-linearly

The previous loader randomized only the first 10 "seed" warehouses and cloned the rest via server-side `INSERT...SELECT`. Measured locally (SF20/50/200), that clone step gets progressively slower as tables grow — Postgres's clone rate roughly doubled every 50 additional warehouses within a single SF200 run (0.61s/warehouse → 0.96s → 1.47s → 2.0s), which is consistent with the working set exceeding the buffer pool at scale. That's why SF1000 (990 cloned warehouses) was so disproportionately worse than SF20/50 would predict linearly.

CSV + native bulk-load throughput, by contrast, measured close to linear across the same SF20→50→200 range (MySQL `order_line`: 11.9s → 29.5s → 114.2s, tracking the warehouse-count ratio almost exactly), so it extrapolates far more confidently to SF1000.

## Options considered for MySQL (and why CSV, without k8s access)

The SF1000 MySQL runs target a persistent in-cluster pod (`chbench-mysql.dataplatform.svc.cluster.local`) that isn't managed in this repo and shares no filesystem with the CI runner. That constraint ruled out several options that looked appealing at first:

- **MySQL's `CLONE` plugin** — a real, native mechanism, but it clones a whole *instance*, not a single schema (unlike Postgres's per-database `CREATE DATABASE ... TEMPLATE`). Using it would require a second persistent "donor" instance in the `dataplatform` namespace plus a remote clone + mandatory recipient restart — new infra outside this repo's control. Not pursued.
- **Transportable tablespaces** (`FLUSH TABLES ... FOR EXPORT` + `IMPORT TABLESPACE`) — the closest schema-level analog to Postgres's `FILE_COPY`, but requires filesystem access (`kubectl cp`/`exec`) to the pod's datadir, which isn't available from CI today. Not pursued.
- **`mysqldump`/text-restore** — ruled out early: still SQL-execution-bound through InnoDB row-by-row, no meaningfully better than the `INSERT...SELECT` clone phase it would replace.
- **CSV + `COPY`/`LOAD DATA LOCAL INFILE`** (chosen) — needs zero filesystem/k8s access, since it's client-streamed over the existing SQL connection. Generation is engine-agnostic (one CSV set feeds both engines' loaders), and the measured numbers hold up at scale (see above).

## MinIO caching — considered, not implemented (yet)

Considered caching generated CSVs in MinIO, keyed by `sf=<N> driver=<hash>` (engine-agnostic — `sf1000`/`sf1000-adaptive`/`sf1000-cold` would share one entry, since the key is scale-factor-only, not spicepod-specific), reusing the MinIO credentials already wired into `testoperator_run_htap.yml` for Cayenne's cold-tier config.

Decided against it for now: measured SF1000 generation at ~2 minutes once parallelized (embarrassingly parallel, no DB dependency), small enough relative to the ~10-12 minute load step that the added complexity (retention policy, cache hit/miss branching, another failure mode) isn't worth it yet.

If `chbench-driver`'s generation code changes in a way that meaningfully increases generation cost, or the last ~2 minutes becomes worth shaving: revisit with prune-on-write per scale factor (mirroring `chbench_template.sh`'s existing stale-template drop) as the primary retention mechanism, with a lifecycle TTL as an orphan/incomplete-multipart-upload backstop — not time-based expiry as the primary approach.

There's also a code comment in `loader.rs`/`loader_mysql.rs` noting a related, smaller future win: if the generated CSVs were cached on the *database server's own disk* instead of streamed from the client each run, loading could use server-side `COPY`/`LOAD DATA INFILE` and skip the network-streaming step entirely (measured ~11% faster locally for MySQL) — worth revisiting if load time becomes the bottleneck again at larger scale.

## Infra follow-up needed (not in this PR)

The persistent in-cluster pods (`chbench-mysql.dataplatform...`, `chbench-postgres.dataplatform...`) aren't managed in this repo. Whoever owns that namespace needs to apply the same `innodb_doublewrite=0`, `innodb_flush_method=O_DIRECT_NO_FSYNC`, and `local_infile=1` settings there for the SF1000 scheduled runs to see this improvement — the docker-based CI action changed in this PR only covers the ephemeral/non-xlarge-runner path.

## Test plan

- [x] `cargo check`/`cargo clippy --no-deps --tests` clean for `chbench-driver` and `testoperator`
- [x] Validated `prepare()` end-to-end against live local Postgres and MySQL containers (not just compiled): row counts match exactly across both engines (`order_line=5,998,407` at SF20, identical between engines and matching prior prototype measurements) and nation/region data is the real CH-benCH reference set
- [x] Measured at SF20/50/200 locally for both engines, confirming the CSV approach's advantage grows with scale as predicted
- [ ] Recommend a real CI SF1000 run to confirm the extrapolated ~12-15 minute seed time before this rides on the regular schedule


krinart (dismissed):

krinart (dismissed):

krinart (dismissed):

bjchambers:
Addressed the remaining review feedback:

- **spawn_blocking JoinError message accuracy** (Copilot): now distinguishes panic vs. cancellation instead of always saying "panicked" — fixed in `b12aea4`, then reconciled with an equivalent fix applied via the GitHub UI's Copilot Autofix in a merge commit (both were logically correct; kept the properly rustfmt'd version).
- All prior rustfmt-formatting comments (`csv_gen.rs`'s `CsvString` Display impl, the `loader.rs`/`loader_mysql.rs` `spawn_blocking` blocks) are resolved as of the current HEAD — verified `cargo fmt --check -p chbench-driver -p testoperator` passes clean.
- Re-verified `cargo check`/`cargo clippy --no-deps --tests` clean for both crates, and re-ran the live end-to-end validation against local Postgres and MySQL containers after these fixes (not just compiled) — row counts and generated data still match exactly (`order_line=5,998,407` at SF20, consistent across both engines and with all prior measurements).
- Merged `trunk` in to clear the `BEHIND` merge state and re-trigger the full CI check suite.

One infra note carried over from the PR description: I don't have access to the `ai-platform`/dataplatform k8s manifests that manage the persistent `chbench-mysql`/`chbench-postgres` pods used for the actual SF1000 scheduled runs, so the `innodb_doublewrite`/`flush_method`/`local_infile` settings still need to be applied there by whoever owns that infra for the scheduled runs to see this improvement — the docker-based CI action changed in this PR only covers the ephemeral/non-xlarge-runner path.

krinart (dismissed):

bjchambers:
Went through every review thread and resolved the ones that were already fixed vs. addressed the ones that needed more:

**Resolved as already fixed** (verified, not just assumed): the earlier `CsvString`/type-mismatch/typo compile errors, the `LOAD DATA LOCAL INFILE` path-escaping fix, the `spawn_blocking` panic-vs-cancellation message, and all the rustfmt-formatting comments — confirmed via `cargo check`/`cargo clippy --no-deps --tests`/`cargo fmt --check`, all clean.

**Four genuinely new findings, now fixed in `8278604`:**
- `join_loader_tasks` aborted worker tasks but never awaited them before returning — since the seed loaders' tempdir is dropped right when `load_all` returns its error, a still-running aborted worker could race the deletion and surface a confusing secondary error. Now awaits every aborted handle first.
- The `TABLE_COLUMNS` lookup silently fell back to an empty column list on a miss instead of erroring — added `Error::UnknownTable` so a future schema-list typo fails loudly instead of producing a baffling `COPY`/`LOAD DATA` SQL error far from the real mistake. Also derived the shared tables' (`item`/`nation`/`region`/`supplier`) column lists from `TABLE_COLUMNS` instead of duplicating the literals.
- The Postgres `COPY` stream did an extra memcpy per 1MiB chunk (`read()` + `Bytes::copy_from_slice`). Switched to reading directly into a `BytesMut` via `read_buf()` and freezing/splitting it for the sink. This one had a real correctness trap — `BytesMut::split()` leaves the remainder with shrunk spare capacity, so without re-`reserve()`-ing every iteration, a later `read_buf` call could silently read 0 bytes indistinguishable from EOF, truncating the upload. Re-verified end-to-end against a live Postgres container at SF50 after the change: `order_line=14,996,194`, matching prior measurements exactly.

All 16 review threads are resolved now.

krinart (approved):

phillipleblanc (approved):

Issue #11823, opened 2026-07-12
Cold Tier: HTAP data-correctness gate failed


sgrebnov:
## Describe the bug

https://github.com/spiceai/spiceai/actions/runs/29200476111/job/86671754107

```
Data Correctness
  replication DID NOT converge within 634676ms
  dataset          source_count    spice_count    count          content
  customer             30000000       30783532 MISMATCH          skipped
    └─ content: skipped — row counts differ
  district                10000          10000       ok         Δ0.0000%
  new_order             9057308        9057308       ok         Δ0.0000%
  order_line          305403579      308896819 MISMATCH          skipped
    └─ content: skipped — row counts differ
  oorder               30540698       30540698       ok         Δ0.0000%
  stock               100000000      102094998 MISMATCH          skipped
    └─ content: skipped — row counts differ
  warehouse                1000           1000       ok         Δ0.0000%
  verdict: FAILED — 3 table(s) mismatched, replication did not converge



Error: HTAP data-correctness gate failed:
  replication did not converge within 634676ms
  customer: source=30000000 spice=30783532 (diff -783532)
  order_line: source=305403579 spice=308896819 (diff -3493240)
  stock: source=100000000 spice=102094998 (diff -2094998)
testoperator exit code: 1
```

sgrebnov:
Root cause identified and fixed. The over-counts (`customer` +783K, `order_line` +3.49M, `stock` +2.09M — spice > source) came from a mid-pass overwrite race in the cayenne compaction passes: an overwrite/promotion committing between a subset-merge/seq-prefix-bake pass's catalog CAS and its fenced in-memory publish wipes the protected-snapshot map and catalog rows, and the unguarded publish re-inserted the merged snapshot into the emptied map — scans then read the whole pre-overwrite warm row set alongside its overwritten/cold copies until restart.

- **Fix**: landed on trunk in #11812 (mid-pass overwrite guard: capture `current_snapshot_id` under the Phase-1 fence read, revalidate under the Phase-3 write fence, discard the merged output on mismatch; plus the bake's cold-tombstone prune cap).
- **Regression tests**: #11844 adds deterministic coverage for both guard sites (verified to fail with the guard disabled).

PR #11844, opened 2026-07-13
test(cayenne): add regression tests for the mid-pass overwrite guard

sgrebnov:
## Summary

PR adds regression tests for #11823 (HTAP data-correctness gate failure: `customer`/`order_line`/`stock` over-counted under active cold-tier promotions).

The fix itself applied in #11812: both compaction passes (protected-snapshot subset-merge and seq-prefix bake) now capture `current_snapshot_id` under the Phase-1 fence read and revalidate it under the Phase-3 write fence

This PR adds the additional regression tests: `subset_compaction_discards_output_when_snapshot_replaced_mid_pass` and `seq_prefix_bake_discards_output_when_snapshot_replaced_mid_pass`

Verified the tests detect the bug: with the guard condition fix disabled, both fail at the guard assertions; with the guard intact, both pass.

### DIFF
--- a/crates/cayenne/src/provider/table.rs
+++ b/crates/cayenne/src/provider/table.rs
@@ -972,6 +972,12 @@ struct VisibleMemTierSegment {
     batches: Vec<RecordBatch>,
 }
 
+/// Test-only mid-pass hook: an async callback fired between a compaction
+/// pass's catalog CAS commit and its fenced in-memory publish. See
+/// `CayenneTableProvider::test_pre_publish_hook`.
+#[cfg(test)]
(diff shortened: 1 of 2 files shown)


bjchambers:
Are we worried this will cause us to discard work too often? Do we have adequate metrics for when this happens?

bjchambers (dismissed):

sgrebnov:
> Are we worried this will cause us to discard work too often? Do we have adequate metrics for when this happens?
Yeah, right now we don't track such occurrences (there could be other reasons as well) so i'll add a new metric.

krinart (approved):

phillipleblanc (approved):

Issue #11838, opened 2026-07-13
Conditional commits: durable federated write-back via dirty-key reconciliation


phillipleblanc:
Part of #11830.

Once the client is told "committed", the write must reach the federated source. There is no enumerable ordered change-log to replay (`cayenne_insert_record` keeps only the latest sequence per key; compaction collapses history), and the source is a **current-state replica**, so deliver by reconciling **dirty keys** rather than replaying a change log.

**Technical:**
- On commit, mark the **written keys dirty** in the **same `MetastoreTransaction`** as the publish — the one load-bearing durability requirement (a separate marker-write would risk an acknowledged-but-undelivered write on crash). A per-key marker, not a payload — bounded by *distinct undelivered keys*.
- A delivery worker reads each dirty key's **current committed value** and pushes a **keyed idempotent upsert** (or delete if now absent), clearing the marker on ack. At-least-once + idempotent = effectively exactly-once. Requires a stable PK matching a target unique key on the source (config prerequisite).
- **Per-provider adapters** (Postgres / MySQL / DynamoDB keyed upsert + delete). Delivery re-reads the live table, so **compaction is never delivery-pinned** (query latency stays stable during an outage; only the dirty set grows).
- **Eager delivery** right after commit keeps the not-yet-delivered window small. Failure model: **infinite Fibonacci-backoff** retry for source delivery; no stop-writes / dead-letter; ships a backlog/lag metric + alert. Replaces the current fire-and-forget path (`accelerated_table/write/write_back.rs`).

**Durability contract (v1 — no metastore change).** The metastore stays at `synchronous=NORMAL`. An acknowledged write is durable across a **process crash / restart / deploy / OOM** (recovery re-reads the dirty set and delivers) and reaches the source promptly; a hard **power loss / OS crash** can lose only writes in the small **not-yet-delivered** window (bounded by delivery lag). Real media/host durability comes from the source copy after delivery, not a local fsync. A stricter *power-loss-durable-at-ACK* mode (per-commit fsync of the metastore **and** the staging/data path, or group-commit) is a **deferred opt-in**, not v1 — flipping `synchronous=FULL` on the metastore alone does not raise end-to-end durability anyway, since the bulk payload lives in the staging/data path (cf. the comment near `staging_wal.rs:933`).

[...]

Issue #11837, opened 2026-07-13
Conditional commits: atomic multi-table commit + upsert-in-txn + admission checks


phillipleblanc:
Part of #11830.

A transaction spans tables by definition; `COMMIT` must publish all touched tables **atomically** — all writes become visible together or none do — so cross-table invariants (e.g. a counter increment in one table gated on a limit read from another) are serializable without write skew.

Builds on the per-key read-footprint OCC (#11834): the footprint + commit-time re-check generalize from one table to a set of tables.

**Technical:**
- **Validate read-key versions for every touched table.** The transaction carries a read footprint per table plus a per-table begin-version (sequence allocators are per-table, so a single scalar high-water is not meaningful across tables). At `COMMIT`, re-check each read key's version against its table's begin-version.
- **Publish all staged writes in one shared `MetastoreTransaction`.** Hold the **written** tables' listing fences in canonical `table_id` order (lock-order-audited deadlock-safe) for reader-atomic visibility; the durable records (snapshot sequences, deletion records) for all tables commit in one metastore transaction on one `cayenne.db`; the in-memory visibility flips happen after that commit, under the held fences. **Read-only** tables need **no** held fence — they are validated by version only.
- **Composable staged-upsert publish (net-new).** The staged-upsert publish (`commit_on_conflict_publish` / `apply_on_conflict_deletions`) is not yet metastore-composable, unlike the overwrite path (`PreparedOverwrite::apply_in_txn`, the #10125 cross-partition seam). Split the staged-upsert commit into `prepare` (under fence) → `apply_in_txn(&mut MetastoreTransaction)` → in-memory `finish()`, so N tables fuse into one transaction.
- **Executor: accept N writes.** Generalize `execute_transaction` and `CayenneTransaction` from single-table (one write / one staged handle / one footprint) to multi-table (per-table footprints, a staged handle per written table, per-table begin-versions). A single gated write remains the one-table degenerate case.
- **Admission checks (else the guarantee is silently void).** Every participating table is a Cayenne accelerator-only table with per-key versioning; all participating tables share one `cayenne.db` (a dataset with its own metadata dir splits the metastore-transaction boundary → reject at transaction prepare); the gate reads a bounded key set. PK-column reassignment and non-INSERT/UPDATE writes remain rejected.

[...]

Issue #11836, opened 2026-07-13
Conditional commits: transaction-body entry plumbing


phillipleblanc:
Part of #11830.

Accept a transaction-shaped multi-statement batch where a single statement is required today, and route it to the executor — across every SQL entry point.

**Technical:**
- Plain-SQL: the parser already returns the full statement list before the single-statement rejection (`session_state.rs`); iterate it rather than dropping all but the first.
- **FlightSQL prepared** (`flight/flightsql/prepared_statement_query.rs`) and **cluster** (`cluster/service.rs`) reject multi-statement independently — handle both.
- Cross-statement **parameter binding** (`$1` / `$2` shared across statements).
- Force **results-cache bypass** for transaction reads (a cached gate read records no footprint).
- A defined **result shape** (the commit outcome). No new API / protocol / session.

**Acceptance:** a `BEGIN…COMMIT` batch is recognized and dispatched on the HTTP, Flight/FlightSQL, and cluster paths; parameters bind across statements; a gate read is not served from the results cache.

phillipleblanc:
Delivered by #11870 (merged to trunk). Closing.

Issue #11835, opened 2026-07-13
Conditional commits: request-scoped transaction executor


phillipleblanc:
Part of #11830.

The core: a request-scoped executor that runs a one-shot `BEGIN … COMMIT` body server-side.

**Technical:**
- Recognize `BEGIN` / `COMMIT` / `ROLLBACK` (they parse but are unexecutable in DataFusion, `physical_planner.rs`; there is no transaction state) and drive our own executor.
- Run each statement at the latest committed state; **stage** writes (→ off-lock staged upsert); record **scan receipts** (→ per-key capture); evaluate `assert()` (terminal on failure).
- Guards: **reject read-your-own-writes** bodies (a statement reading a table it already wrote — staged writes are invisible to later statements in v1); **reject non-deterministic UDFs** (`random()`, external/Python) — `assert()` and `now()` (pinned per attempt) are exempt; **reject unbounded-scan gates** (v1 gates read a bounded, explicitly-determined key set).
- At `COMMIT`: re-check every read key's version; clean → publish; conflict → **bounded server-transparent retry** at the newest state (relative writes recomputed); gate-abort → bounded revalidation, then a terminal "state-moved"-qualified negative outcome.
- Return the **commit outcome**, never a pre-retry intermediate.

**Acceptance:** N concurrent same-key gated transactions never over-count (one retries); an over-limit gate returns terminal; a read-your-own-writes / `random()` / unbounded-scan body is rejected. (Executor may split into skeleton / validate+publish / retry sub-PRs.)

phillipleblanc:
Delivered by #11870 (merged to trunk). Closing.

PR #11848, opened 2026-07-13
Cayenne transactions (1/9): assert() gate UDF + two-class error protocol

phillipleblanc:
> **Cayenne transactions — stacked PR 1 of 9**
> Previous: _(none — base of stack)_ &nbsp;·&nbsp; Next: #11849 →
>
> Review/merge bottom-up: #11848 · #11849 · #11859 · #11861 · #11863 · #11865 · #11866 · #11868 · #11869 &nbsp;→&nbsp; all squashed into #11870 (targets `trunk`). Part of #11830.

---

Part of #11830.

First working slice of the conditional-commits transaction API: a serializable, gated `BEGIN … COMMIT` transaction over a Cayenne accelerated table, with write-back to the federated source. This lands the end-to-end skeleton; several v1 requirements are explicitly deferred to follow-up PRs (below).

## What's in this PR

- **`assert(bool)` scalar UDF** (`runtime-datafusion-udfs`) — `Volatile`, returns a terminal error on a false/NULL gate (message marker `assertion failed:` so a gate abort is distinguishable from a retryable conflict). Registered in `register_core_scalar_udfs`, so it is callable on every SQL surface. Unit-tested.
- **`BEGIN … COMMIT` transaction executor** on the HTTP `/v1/sql` path (`http/v1/mod.rs`) — detects a transaction body via `DFParser`, runs each inner statement, and aborts the transaction if the `assert()` gate fails. Reuses the accelerated-table write path, so a `write_mode: write_back` Cayenne dataset commits to the accelerator and propagates to the source.
- **`QueryBuilder::bypass_cache`** — transaction gate/inner reads skip the SQL results cache (no lookup/store, equivalent to `Cache-Control: no-cache`) so the gate always reads live committed state. (Without this, the 1s results cache served stale gate reads and caused over-admit.)

## Verified

- `assert()` unit tests (4/4).
- Manual e2e against a Postgres-backed Cayenne `write_back` dataset, with the SQL results cache **on** (default):
  - gate pass / fail;
  - cap enforcement (a capped counter stops at its limit; further admits abort);
  - 200 concurrent admits to one capped key → exactly the cap, no over-admit;
  - the source (Postgres) converges with the accelerator.

## Deferred to follow-up PRs (tracked)

This slice serializes gated transactions under a single process lock and reuses the existing async write-back — not the full v1. Remaining v1 work:

- **Per-key OCC** (#11834) — replace the global serialization lock with per-key read-version capture + commit-time re-check, so transactions touching disjoint keys commit concurrently.
- **True atomicity — staged commit + rollback** (#11833, #11837) — stage writes and publish all-or-nothing at `COMMIT` (a later statement failing must roll back earlier writes), instead of the current gate-first, execute-immediately model.
- **Durable write-back** (#11838) — dirty-key reconciliation recorded in the commit transaction (this slice reuses the async best-effort path).
- **FlightSQL (and cluster) entry** (#11836) — currently HTTP `/v1/sql` only.
- Automated integration tests for the executor.

## Notes

No spicepod or API surface change: the transaction body is submitted as ordinary SQL to the existing `/v1/sql` endpoint.

---
**Implements:** #11831 (assert() scalar UDF + two-class error protocol). Seeds the request-scoped executor (#11835) and transaction-body entry detection (#11836) completed in #11859.

---

## Stacked-PR review notes

**Implements #11831** (assert() UDF + two-class error protocol).

Part of the #11830 stack — the plan is to merge one squashed PR into `trunk`; these stacked PRs are the reviewable evolution of the feature, so a few pieces here are intentionally reworked by later PRs:

- The transaction-executor scaffolding introduced here (in `http/v1`) is superseded: the request-scoped executor lands in #11859, and is then **extracted** out of `http/v1` into the shared `datafusion::query::transaction` module in #11866 so HTTP and FlightSQL drive the same code.

### DIFF
--- a/crates/runtime-datafusion-udfs/src/assert.rs
+++ b/crates/runtime-datafusion-udfs/src/assert.rs
@@ -1,0 +1,165 @@
+/*
+Copyright 2024-2025 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//! `assert(<bool>)` scalar UDF — the gate primitive for conditional-commit transactions.
+//!
+//! Returns its boolean argument unchanged when it is `TRUE`; returns an error when it is
+//! `FALSE` or `NULL`. The error is a [`DataFusionError::Execution`] whose message begins with
+//! the marker [`ASSERT_FAILED_MARKER`] so a transaction executor can distinguish a gate abort
+//! (terminal — return to the client) from a serialization conflict (retryable).
+//!
+//! The UDF is declared [`Volatility::Volatile`] so the optimizer never const-folds or elides it:
+//! `assert(1 > 2)` must fail at *execution* time (as an execution error), not be pre-evaluated at
+//! planning time.
+
+use arrow::array::{Array, BooleanArray};
+use arrow::datatypes::DataType;
+use datafusion::common::DataFusionError;
+use datafusion::logical_expr::{
+    ColumnarValue, ScalarFunctionArgs, ScalarUDFImpl, Signature, Volatility,
+};
+use datafusion::scalar::ScalarValue;
+
+pub static ASSERT_SCALAR_UDF_NAME: &str = "assert";
+
+/// Prefix on the error message of every `assert()` failure, so callers can recognize a gate
+/// abort without relying on the (shared) `ErrorCode`.
+pub static ASSERT_FAILED_MARKER: &str = "assertion failed:";
+
+#[derive(Debug, Hash, Eq, PartialEq)]
+pub struct Assert {
+    signature: Signature,
+}
+
+impl Default for Assert {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Assert {
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            // Volatile => never const-folded/elided; evaluated at execution time.
+            signature: Signature::exact(vec![DataType::Boolean], Volatility::Volatile),
+        }
+    }
+
+    fn failed() -> DataFusionError {
+        DataFusionError::Execution(format!(
+            "{ASSERT_FAILED_MARKER} gate expression was false or NULL"
+        ))
+    }
+}
+
+impl ScalarUDFImpl for Assert {
+    fn name(&self) -> &'static str {
+        ASSERT_SCALAR_UDF_NAME
+    }
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType, DataFusionError> {
+        Ok(DataType::Boolean)
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue, DataFusionError> {
+        let arg = args.args.into_iter().next().ok_or_else(|| {
+            DataFusionError::Execution(format!(
+                "{ASSERT_FAILED_MARKER} assert() requires exactly one boolean argument"
+            ))
+        })?;
+
+        match arg {
+            // Scalar TRUE -> pass through.
+            ColumnarValue::Scalar(ScalarValue::Boolean(Some(true))) => {
+                Ok(ColumnarValue::Scalar(ScalarValue::Boolean(Some(true))))
+            }
+            // Scalar FALSE or NULL -> abort.
+            ColumnarValue::Scalar(ScalarValue::Boolean(_)) => Err(Self::failed()),
+            // Array: abort on any FALSE or NULL, otherwise pass through.
+            ColumnarValue::Array(array) => {
+                let bools = array
+                    .as_any()
+                    .downcast_ref::<BooleanArray>()
+                    .ok_or_else(|| {
+                        DataFusionError::Execution(format!(
+                            "{ASSERT_FAILED_MARKER} assert() argument must be Boolean"
+                        ))
+                    })?;
+                for i in 0..bools.len() {
+                    if bools.is_null(i) || !bools.value(i) {
+                        return Err(Self::failed());
+                    }
+                }
+                Ok(ColumnarValue::Array(array))
+            }
+            other => Err(DataFusionError::Execution(format!(
+                "{ASSERT_FAILED_MARKER} assert() argument must be Boolean, got {}",
+                other.data_type()
+            ))),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use datafusion::config::ConfigOptions;
+    use datafusion::logical_expr::ScalarFunctionArgs;
+
+    fn call(arg: ColumnarValue) -> Result<ColumnarValue, DataFusionError> {
+        Assert::new().invoke_with_args(ScalarFunctionArgs {
+            args: vec![arg],
+            arg_fields: vec![],
+            number_rows: 1,
+            return_field: std::sync::Arc::new(arrow::datatypes::Field::new(
+                "r",
+                DataType::Boolean,
+                true,
+            )),
+            config_options: std::sync::Arc::new(ConfigOptions::new()),
+        })
+    }
+
+    #[test]
+    fn assert_true_passes() {
+        let r = call(ColumnarValue::Scalar(ScalarValue::Boolean(Some(true))));
+        assert!(r.is_ok());
+    }
+
+    #[test]
+    fn assert_false_aborts() {
+        let e = call(ColumnarValue::Scalar(ScalarValue::Boolean(Some(false)))).unwrap_err();
+        assert!(e.to_string().contains(ASSERT_FAILED_MARKER));
+    }
+
+    #[test]
+    fn assert_null_aborts() {
+        let e = call(ColumnarValue::Scalar(ScalarValue::Boolean(None))).unwrap_err();
+        assert!(e.to_string().contains(ASSERT_FAILED_MARKER));
+    }
+
+    #[test]
+    fn assert_array_any_false_aborts() {
+        let arr = std::sync::Arc::new(BooleanArray::from(vec![Some(true), Some(false)]));
+        let e = call(ColumnarValue::Array(arr)).unwrap_err();
+        assert!(e.to_string().contains(ASSERT_FAILED_MARKER));
+    }
+}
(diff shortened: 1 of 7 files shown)


crates/runtime/src/datafusion/query/builder.rs
```diff
@@ -41,6 +41,7 @@ pub struct QueryBuilder {
     query_id: Uuid,
     cancellation_token: Option<CancellationToken>,
     read_only: bool,
+    bypass_cache: bool,
```
phillipleblanc: This should be an enum, not a boolean

crates/runtime/src/datafusion/query.rs
```diff
@@ -238,6 +238,11 @@ pub struct Query {
     /// regardless of per-catalog writability. Set via [`QueryBuilder::read_only`];
     /// used by `/v1/tools/sql` and `/v1/nsql` to contain LLM-generated SQL.
     read_only: bool,
+    /// When true, this query bypasses the SQL results cache entirely (no lookup,
+    /// no store) — equivalent to `Cache-Control: no-cache`. Set via
+    /// [`QueryBuilder::bypass_cache`]; used by the conditional-commit transaction
+    /// executor so a gate read always sees live committed state.
```
phillipleblanc:
```suggestion
    /// [`QueryBuilder::bypass_cache`]
```

crates/runtime/src/datafusion/query.rs
```diff
@@ -238,6 +238,11 @@ pub struct Query {
...
+    bypass_cache: bool,
```
phillipleblanc: This should be an enum

crates/runtime/src/datafusion/query.rs
```diff
@@ -1112,38 +1118,95 @@ impl Query {
...
-                                    &query_id_str,
-                                    &timeout_state,
-                                )?;
-                                (plan, tracker, cache_manager)
-                            }
-                            PlanOrCached::Cached(query_result) => {
+                        if bypass_cache {
+                            // Gate/inner reads in a conditional-commit transaction must
+                            // see live committed state, so skip the results cache entirely
+                            // (no lookup, no store) — mirrors the table-allowlist arm.
```
phillipleblanc:
```suggestion
```

crates/runtime/src/datafusion/query.rs
```diff
@@ -1112,38 +1118,95 @@ impl Query {
...
-                            PlanOrCached::Plan(plan, tracker, cache_manager) => {
-                                Self::ensure_not_cancelled(
-                                    &query_cancel_token,
-                                    &query_id_str,
-                                    &timeout_state,
-                                )?;
-                                (plan, tracker, cache_manager)
-                            }
-                            PlanOrCached::Cached(query_result) => {
+                        if bypass_cache {
```
phillipleblanc: this is a lot of if/else nesting, let's refactor this to be more readable.

crates/runtime/src/http/v1/mod.rs
```diff
@@ -302,6 +312,109 @@ pub async fn sql_to_http_response(
...
+/// Per-key optimistic concurrency — so transactions touching disjoint keys run in
+/// parallel — is the tracked follow-up (spiceai#11834/#11835).
+static CONDITIONAL_COMMIT_LOCK: LazyLock<Mutex<()>> = LazyLock::new(|| Mutex::new(()));
+
+/// Detects a `BEGIN … COMMIT` transaction body and returns its inner statements as
+/// SQL strings (transaction-control statements stripped), or `None` if `sql` is not
+/// a well-formed `BEGIN … COMMIT` body. A non-transaction (single statement, or a
+/// multi-statement string not wrapped in `BEGIN … COMMIT`) returns `None` and is
+/// handled by the ordinary path unchanged.
+fn detect_conditional_commit_transaction(sql: &str) -> Option<Vec<String>> {
```
phillipleblanc: this should just be a commit transaction, nothing about this has to do with conditional statements

crates/runtime/src/http/v1/mod.rs
```diff
@@ -302,6 +312,109 @@ pub async fn sql_to_http_response(
...
+/// (or any statement error) aborts the transaction — subsequent statements do not
+/// run and the error is returned to the client.
+///
+/// v1 reuses the ordinary write path, so a Cayenne-accelerated `write_mode:
+/// write_back` dataset commits to the accelerator and propagates to the source as
+/// usual. It is not multi-statement atomic on failure (writes applied by an earlier
+/// statement are not rolled back); the canonical shape — gate first, then writes —
+/// means a gate failure applies no writes. Transaction-level bound parameters are
+/// not yet supported (use literal values in the body).
+async fn execute_conditional_commit(
```
phillipleblanc: This function name (and the comment above) should not reference conditional commits at all

crates/runtime/src/http/v1/mod.rs
```diff
@@ -302,6 +312,109 @@ pub async fn sql_to_http_response(
...
+            Ok(res) => res,
+            Err(e) => {
+                let kind = SqlErrorKind::of_query_error(&e);
+                return sql_error_response(e.to_string(), kind);
+            }
+        };
+        let cache_status = query_res.cache_status;
+        // Drain each statement fully so writes execute and any error (including a
+        // gate abort) surfaces before the next statement runs.
+        match query_res.data.try_collect::<Vec<RecordBatch>>().await {
```
phillipleblanc: This should stream and throw away batches (except the last one) so we don't collect a large result set in memory. The last one should stream out the result.

crates/runtime-datafusion-udfs/src/assert.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType, DataFusionError> {
+        Ok(DataType::Boolean)
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> Result<ColumnarValue, DataFusionError> {
+        let arg = args.args.into_iter().next().ok_or_else(|| {
+            DataFusionError::Execution(format!(
+                "{ASSERT_FAILED_MARKER} assert() requires exactly one boolean argument"
+            ))
```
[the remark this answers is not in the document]
phillipleblanc: Fixed in de2911ba. `invoke_with_args` now requires a one-element argument array, rejects both missing and extra arguments, and includes a regression test for each invalid count.

crates/runtime/src/http/v1/mod.rs
```diff
@@ -255,6 +259,12 @@ pub async fn sql_to_http_response(
     format: ResponseMimeType,
     read_only: bool,
 ) -> Response {
+    // Conditional commits: a `BEGIN … COMMIT` transaction body is run by a dedicated
+    // serializing executor rather than the (rejected) single-statement path.
+    if let Some(inner) = detect_conditional_commit_transaction(&sql) {
+        return execute_conditional_commit(df, inner, read_only, format).await;
+    }
```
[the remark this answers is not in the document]
phillipleblanc: Fixed in de2911ba. HTTP parameters are forwarded to every transaction statement. Positional values are normalized to named entries so different statements can reference subsets such as `$1` and `$2`; the HTTP transaction test covers this.

crates/runtime/src/http/v1/mod.rs
```diff
@@ -302,6 +312,109 @@ pub async fn sql_to_http_response(
...
+        };
+        let cache_status = query_res.cache_status;
+        // Drain each statement fully so writes execute and any error (including a
+        // gate abort) surfaces before the next statement runs.
+        match query_res.data.try_collect::<Vec<RecordBatch>>().await {
+            Ok(batches) => last = Some((batches, cache_status)),
+            Err(e) => {
+                return sql_error_response(e.to_string(), SqlErrorKind::of_datafusion_error(&e));
+            }
+        }
```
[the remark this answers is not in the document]
phillipleblanc: Fixed in de2911ba. Intermediate streams are now drained and dropped one batch at a time without collecting. The final statement uses the existing HTTP streaming response path.

crates/runtime/src/datafusion/query.rs
```diff
@@ -1112,38 +1118,95 @@ impl Query {
...
+                            // Gate/inner reads in a conditional-commit transaction must
+                            // see live committed state, so skip the results cache entirely
+                            // (no lookup, no store) — mirrors the table-allowlist arm.
+                            let raw_cache_key = CacheKey::Query(sql.as_ref(), parameters.as_ref())
```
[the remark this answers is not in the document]
phillipleblanc: Added in de2911ba. The regression test warms a client-keyed entry, verifies bypass executes fresh SQL, and then verifies the bypassed result did not overwrite the warm entry.

crates/runtime-datafusion-udfs/src/assert.rs
```diff
@@ -0,0 +1,165 @@
+/*
+Copyright 2024-2025 The Spice.ai OSS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
```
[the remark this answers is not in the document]
phillipleblanc: Fixed in de2911ba. The new source file now uses the required 2024-2026 copyright range.

crates/runtime/src/datafusion/query/builder.rs
phillipleblanc: Fixed in de2911ba. The boolean was replaced with `ResultsCacheMode::{Default, Bypass}` and the builder now accepts that enum.

crates/runtime/src/datafusion/query.rs
phillipleblanc: Fixed in de2911ba. The field documentation now links to `QueryBuilder::results_cache_mode`.

crates/runtime/src/datafusion/query.rs
phillipleblanc: Fixed in de2911ba. `Query` now stores `ResultsCacheMode` rather than a boolean.

crates/runtime/src/datafusion/query.rs
phillipleblanc: Fixed in de2911ba. The transaction-specific inline comment was removed from the query hot path; the behavior is documented on the cache mode and uncached planning helper.

crates/runtime/src/datafusion/query.rs
phillipleblanc: Refactored in de2911ba. Cache mode selects either cached or uncached planning, followed by one flat `PlanOrCached` match, removing the duplicated nested planning/error path.

crates/runtime/src/http/v1/mod.rs
phillipleblanc: Fixed in de2911ba. The parser and lock are now generically named `transaction_statements` and `TRANSACTION_LOCK`, with transaction-only documentation.

crates/runtime/src/http/v1/mod.rs
phillipleblanc: Fixed in de2911ba. The executor is now named `execute_transaction`, and its documentation no longer describes it as a conditional-commit function.

crates/runtime/src/http/v1/mod.rs
phillipleblanc: Fixed in de2911ba. Intermediate batches are streamed and discarded, while the final result is sent through the HTTP streaming path. The transaction guard remains attached to the final lazy stream until it completes or is dropped.

PR #11853, opened 2026-07-14
fix: Fairly allocate partitions from scheduler assignment cycle only

peasee:
## 📝 Summary

* Updates the scheduler to return `Unavailable` to executors requesting their initial partition allocations until the first allocation assignment interval passes. This ensures that executors have some time to connect - instead of the first executor racing to greedily allocate all of the partitions for a table to itself
* Updates initial allocation to use fair share allocation similar to the regular partition interval assignments, instead of greedily assigning every partition for a table to one executor
* Ensures that partitions are more evenly spread across executors on startup to prevent executors taking too many

## 👀 Notes for Reviewers

* Partition assignment probably needs more work. Returning `Unavailable` to the executors forces them to re-check their initial assignments in a retry loop, but if the partition assignment interval is too long that retry loop will fail before the scheduler allocates any initial partitions. I think partition assignment, including initial partitions, needs to be entirely push-based instead of executors pulling partitions on first startup - but that requires some additional work.


phillipleblanc (approved):
I think the correct solution would be to have a proper rebalancing control loop that doesn't depend on executors coming in right after startup, but this at least is an incrementally better solution than what we already had.

Issue #11856, opened 2026-07-14
Conditional commits: per-key OCC granularity — add a re-readable per-key version to the Cayenne PK index


phillipleblanc:
Part of #11830. Prerequisite for the per-key granularity of #11834.

## Context

The v1 conditional-commit OCC ships at **per-table** granularity: staging (gate read + Vortex encode) runs fully lock-free, and the per-table `write_lock` is taken only briefly at transaction begin (to snapshot the table's sequence high-water) and again at commit (re-check it, then publish). A concurrent commit to the **same table** bumps the sequence, so this transaction aborts with a retryable conflict instead of blocking. This is conflict-complete — no lost updates, no over-admit — but it over-aborts concurrent commits to the same table regardless of whether they touched overlapping keys.

The intended end state (see #11834) is **per-key** granularity: transactions touching disjoint keys on the same table commit concurrently, and only genuine same-key races conflict.

## Finding

Per-key granularity cannot be built from the accelerator's current stored state. The PK existence index entry is `{ row, location }` with **no per-key version**, and the two candidate signals both have concrete false negatives (a missed conflict = a silent lost update on `SET n = n + 1`):

- **`RowLocation` compare** — the commit-side recorder always writes `FileUnlocated` (the `(file, position)` upgrade is a later async pass), so a concurrent same-key upsert overwrites `FileUnlocated` → `FileUnlocated`; likewise `Inlined` → `Inlined` for a re-inlined small upsert. The compare sees no change.
- **Re-derived deletion-set compare** — under key-mode / bloom existence, the superseded descriptor for a key is just its key bytes, identical across versions, so an interposed commit is invisible to set-equality.
- A concurrent filter `DELETE` on an upsert table deliberately leaves the key stale-present in the index, so presence/location compares miss a lost delete entirely.

So #11834's "reuse the accelerator's existing per-key / per-shard versions" premise does not hold as written; a re-readable per-key version has to be added first.

## Proposed work

Add a re-readable per-key version to the exact-keyset PK index and re-check it per written key at commit:

- Add a `sequence: i64` (or equivalent version) to the PK index entry.
- Stamp it with the committing write's sequence in the record paths (the file-key and inlined-key recorders).
- Reconstruct it on keyset load from each file's snapshot sequence / inline entry sequence, so a cold-rebuilt index carries versions.
- At commit, for each written key compare the captured version against the current one, plus a tombstone probe to catch a concurrent filter delete of the key.

[...]

PR #11849, opened 2026-07-13
Cayenne transactions (2/9): staged-upsert substrate

phillipleblanc:
> **Cayenne transactions — stacked PR 2 of 9**
> ← Previous: #11848 &nbsp;·&nbsp; Next: #11859 →
>
> Review/merge bottom-up: #11848 · #11849 · #11859 · #11861 · #11863 · #11865 · #11866 · #11868 · #11869 &nbsp;→&nbsp; all squashed into #11870 (targets `trunk`). Part of #11830.

---

Part of #11830. **Stacked on #11848** (base: `phillip/conditional-commits`) — review/merge that first.

PR-2 of the conditional-commits stack: the **staged-commit substrate** the transaction executor (PR-3) will use to publish — or roll back — a write atomically.

## What this adds

`CayenneTableProvider::begin_staged_upsert(rows, target_partitions) -> CayenneStagedUpsert`, with `CayenneStagedUpsert::{commit, rollback, row_count}`:

- **stage** — validate PK conflicts, write the replacement rows to a fresh **unreferenced** snapshot dir, capture the on-conflict deletions, and hold the table write guard. The catalog is untouched, so the rows are invisible to every reader and there's nothing to compensate on abort.
- **commit** — under the visibility + listing-fence write lock: apply the on-conflict deletions, reserve + durably record the snapshot sequence (strictly above the delete sequence so the staged rows survive their own conflict deletes), then flip the protected snapshot in one fence write. The rows become visible atomically.
- **rollback** — discard the staged snapshot dir.

This splits the synchronous on-conflict write path at the listing fence so a caller can stage a PK upsert now and publish/discard it later. **Nothing calls it yet** — PR-3 (the transaction executor) will; it's `pub` so there's no dead-code warning.

Ported from the analytical-replica dual-apply work, with one adaptation to current trunk (`HashSet<OwnedRow>` → `PkDigestSet`).

## Verified

- `cargo build -p cayenne` — clean (no errors/warnings).
- `cargo test -p cayenne staged_upsert` — **3/3**: `staged_upsert_invisible_until_commit`, `staged_upsert_rollback_discards`, `staged_upsert_supersedes_existing_pk`.

---
**Implements:** #11833 (staged-upsert substrate; made off-lock in #11859).

---

## Stacked-PR review notes

**Implements #11833** (off-lock staged upsert + Drop guard).

Reworked by later PRs in the stack:

[...]

Issue #11833, opened 2026-07-13
Conditional commits: off-lock staged upsert + Drop guard


phillipleblanc:
Part of #11830.

Staging a write must be **off-lock** so the (possibly analytical) gate never blocks other writers: the write lock is released after staging, and the commit point (listing fence) is taken only for validate + publish.

**Technical:** evolve the staged-upsert primitive (`crates/cayenne/src/provider/staged_upsert.rs`, `begin_staged_upsert` → `commit`/`rollback`) to release the write lock after staging; add a `Drop` guard that cleans up the staged snapshot dir if dropped without commit/rollback (currently leaks). Builds on the on-conflict staging in `table.rs` / `staging_wal.rs`.

**Acceptance:** two concurrent staged upserts don't block on each other; a dropped-without-commit stage leaves no orphan dir; existing staged-upsert tests stay green.

phillipleblanc:
Delivered by #11870 (merged to trunk). Closing.

Issue #11830, opened 2026-07-13
Enhancement: Conditional commits - serializable gated writes for Cayenne accelerated tables with durable federated write-back


phillipleblanc:
## Goal-State/What/Result

A Spice user can submit a set of writes together with a validation query as a single transaction against a Cayenne accelerated table, so the writes commit only if the validation still holds — with the check and the writes applied **atomically**, and concurrent requests unable to slip a change past the validation. Committed writes are then **durably written back to the federated source**, so the source stays current and Spice serves as the system of record. It is all expressed in standard SQL, with no logic pushed down to the source database.

```sql
-- Reserve stock for a SKU only if it stays under the available quantity —
-- atomically in Cayenne, then durably written back to the source.
BEGIN;
  SELECT assert(                                        -- gate: aborts the transaction if false
    (SELECT COALESCE(SUM(quantity), 0)                  -- currently allocated for this SKU ...
       FROM allocations WHERE sku = $1)
      < (SELECT stock FROM inventory WHERE sku = $1));   -- ... must stay below available stock
  INSERT INTO allocations (id, sku, quantity)           -- the write, committed iff the gate passed
    VALUES ($2, $1, 1);
COMMIT;                                                 -- atomic; then durably written back to the source
```

## Why/Purpose

Applications frequently need to run an arbitrary validation query over existing data and perform a write **only if the result still holds** — enforcing a limit, an allocation invariant, or any cross-row rule — with the check and the write applied atomically and serializably so concurrent writers can't violate the invariant. Today a Spice user must push this validate-then-write pattern down to the source database, which reintroduces the per-request source contention and connection pressure that accelerating in Spice was meant to relieve, and it can't take advantage of Spice's fast columnar evaluation of the validation query. Providing this natively — with durable write-back to the source — lets the pattern stay local to Spice, optimistic and concurrent, while Spice remains the system of record and keeps the source current.

## By When

**Issue/Spec written and reviewed:** _TBD_
**Done-Done:** Tuesday, July 14, 2026

## Done-Done

- [x] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [x] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [x] Threat Model / Security Review
- [ ] Tests
- [ ] Telemetry / Metrics / Task History
- [ ] Performance / Benchmarks
- [ ] Documentation
[...]

PR #11859, opened 2026-07-14
Cayenne transactions (3/9): serializable executor with off-lock optimistic concurrency

phillipleblanc:
> **Cayenne transactions — stacked PR 3 of 9**
> ← Previous: #11849 &nbsp;·&nbsp; Next: #11861 →
>
> Review/merge bottom-up: #11848 · #11849 · #11859 · #11861 · #11863 · #11865 · #11866 · #11868 · #11869 &nbsp;→&nbsp; all squashed into #11870 (targets `trunk`). Part of #11830.

---

Part of #11830. Stacked on #11849 (staged-upsert substrate) — review/merge that first; this PR's diff is against that branch.

## What

Runs a standard `BEGIN … COMMIT` body submitted to `/v1/sql` as a **serializable transaction** over a Cayenne accelerator-only table, with a small `assert(<expr>)` gate. Whether a transaction is "conditional" is up to the caller's use of `assert()` — the mechanism is just a transaction.

Example — enforce a limit atomically and concurrently:

```sql
BEGIN;
SELECT assert((SELECT n FROM counters WHERE id = 'x') < (SELECT cap FROM counters WHERE id = 'x'));
UPDATE counters SET n = n + 1 WHERE id = 'x';
COMMIT;
```

## How

- **Everything runs through `QueryBuilder`.** Both the gate reads and the write execute via the normal query path, so authorization, row/column masking, logging, tracing, and the read-only gate apply to the write exactly as to any other query. There is no plan interception.
- **Off-lock optimistic staging.** Validation and Vortex encode run **without** the per-table write lock, so transactions stage concurrently into an invisible snapshot. The lock is taken only at commit, briefly, to re-check an optimistic-concurrency token (the table's sequence high-water, captured *before* the gate read) and publish. A concurrent commit to the table makes this one abort with a retryable **409** rather than lose an update.
- **Atomicity.** The staged write publishes at `COMMIT` or not at all; a later statement failing (e.g. a gate `assert` after the write) rolls the staged write back. An `UPDATE` stages only its insert (upsert) leg so its delete cannot publish early.
- The write path detects the active transaction via the `RequestContext` typed extension on the task context (never the task-local, whose silent fallback would publish immediately).

## v1 scope / follow-ups

- One `INSERT`/`UPDATE` write per transaction — multi-write/multi-table is **rejected** with a clear error (atomic multi-table commit is #11837).
- Per-**table** conflict granularity — correct (no lost updates / no over-admit), over-aborts concurrent same-table commits; per-**key** granularity needs a new stored per-key version (#11856).
- Accelerator-only tables — durable federated write-back is #11838.
- Single hot-key OCC thrash (each conflict re-stages) is the deferred "thrash detector" case in #11830.

## Verification

Built `spiced` locally against a Postgres source + Cayenne file accelerator (`on_conflict: upsert`, `refresh_mode: full` → accelerator-only):

- Cap enforcement + atomicity (sequential): 5 commits, 2 aborts, `n=5`.
- No lost update / no over-admit (8-way concurrent, retry-on-409): 5 commits, `n=5` exactly, 0.58s; a 150-way/cap-100 stress run also converged to exactly 100.
- Write-then-fail rollback: `assert(false)` after an `UPDATE` leaves the row unchanged.
- Read-only body, `DELETE`-in-txn rejection, multi-write rejection: all as expected.
- Single transaction latency ~60ms.

---
**Implements:** #11835 (request-scoped transaction executor) and #11836 (transaction-body entry plumbing); completes the off-lock half of #11833.

---

## Stacked-PR review notes

**Implements #11835** (request-scoped transaction executor) and **#11836** (transaction-body entry plumbing); completes the off-lock rework of #11833.

Reworked by later PRs in the stack:

- The **per-table** OCC gate here (`TransactionWriteToken` = the table's sequence high-water) is replaced by **per-key** read-footprint OCC in #11861 (with a rebuild floor-stamp fix in #11863).
- The **single-table** executor (`TransactionWrite`, one staged commit) is generalized to **multi-table** in #11865 (`TransactionHandle` + a per-table participant map + the `CayenneTransaction::commit` orchestrator).
- The executor lives in `http/v1` here; it is **extracted** into the shared `datafusion::query::transaction` module (used by both HTTP and FlightSQL) in #11866.
- Admission accepts accelerator-only tables here; it is extended to durable-write-back tables in the write-back PR (#11838).

### DIFF
--- a/crates/cayenne/src/provider/staged_upsert.rs
+++ b/crates/cayenne/src/provider/staged_upsert.rs
@@ -14,48 +14,40 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-//! Staged **upsert** lifecycle for `CayenneTableProvider` (Analytical Replica
-//! dual-apply).
+//! Staged **upsert** lifecycle for `CayenneTableProvider`: a PK/on-conflict
+//! write is encoded into a fresh, **unreferenced** snapshot directory
+//! (invisible to readers) and published — or discarded — at a later moment.
+//! It is a split of the synchronous on-conflict path at the listing fence.
 //!
-//! [`super::staging_wal::CayenneStagedAppend`] stages append-only writes; it
-//! rejects primary-key / on-conflict tables. This module adds the PK-aware
-//! counterpart: a write into a table with a primary key and `on_conflict`
-//! upsert is staged **invisibly** and published (or discarded) at a later
-//! moment — the point at which the dual-apply driver has an ack (or a failure)
-//! from the federated source.
+//! Staging is **off-lock optimistic** ([`CayenneTableProvider::begin_staged_upsert_occ`]):
+//! validation (private keyset) + Vortex encode run **without** `write_lock`, so
+//! many transactions stage concurrently; the lock is taken only at
+//! [`CayenneStagedUpsert::commit`], briefly, to re-check the optimistic-concurrency
+//! token and publish. Detection is at per-table granularity: if any commit landed
+//! on the table between the token capture (at transaction begin, before the gate
+//! read) and this commit, the sequence high-water moved → the commit aborts with
+//! [`Error::WriteConflict`] (a retryable conflict) rather than risking a lost
+//! update.
 //!
-//! It is a split of the synchronous on-conflict path
-//! ([`CayenneTableProvider::write_new_snapshot_after_validation`]) at the
-//! listing fence:
+//! [`CayenneStagedUpsert::commit`] runs the fenced publish: apply the
+//! on-conflict deletions, reserve the protected-snapshot sequence (strictly
+//! above the delete sequence, so the staged rows survive their own conflict
+//! deletes), durably record it, then flip the deletion caches and the protected
+//! snapshot in one listing-fence write. [`CayenneStagedUpsert::rollback`]
+//! discards the staged snapshot directory.
 //!
-//! - [`CayenneTableProvider::begin_staged_upsert`] validates the incoming rows
-//!   for PK conflicts and writes them into a fresh, **unreferenced** snapshot
-//!   directory (`{table_id}/{new_snapshot}/`). Nothing durable in the catalog
-//!   is touched, so the rows are invisible to every reader and there is nothing
-//!   to compensate on abort. The captured [`OnConflictDeletions`] (which prior
-//!   versions this write supersedes) rides the returned handle.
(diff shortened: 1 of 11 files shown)


crates/cayenne/src/provider/staged_upsert.rs
```diff
@@ -65,24 +57,52 @@ use super::on_conflict::{OnConflictDeletions, PostValidationState};
...
-/// table block until it is either committed via [`Self::commit`] or discarded
-/// via [`Self::rollback`] (or dropped — see the note on `Drop`).
+/// Captured at transaction **begin** (before the gate read) under a brief
+/// `write_lock` hold, and re-checked at commit under `write_lock`. It is the
+/// per-table sequence high-water plus a "no staging append in flight" bit —
+/// together, everything that changes key liveness on the table either advances
+/// the sequence or sets the staging bit, so an unchanged token at commit proves
+/// no intervening commit touched the table (see the module docs).
+#[derive(Debug, Clone, Copy)]
+pub struct TransactionWriteToken {
```
phillipleblanc, line 69: Stacked-PR review note: this is the **per-table** OCC gate (`TransactionWriteToken` = the table sequence high-water). It is reworked to **per-key** read-footprint OCC in #11861 (rebuild floor-stamp fix in #11863), so a transaction only conflicts on the keys it actually read/wrote.

crates/runtime/src/http/v1/mod.rs
```diff
@@ -392,88 +390,276 @@ async fn run_transaction_statement(
...
+///    detects the active transaction and *stages* the rows instead of
+///    publishing them; a failed `assert()` gate (or any error) aborts.
+/// 3. On success, publish the staged write atomically at COMMIT and invalidate
+///    the results cache; on any failure, roll back the staged write.
+///
+/// v1 supports a single INSERT/UPDATE write per transaction (no DELETE/MERGE,
+/// no PK reassignment). Bound parameters are supported across statements via
+/// [`normalize_transaction_parameters`]. Durable propagation of the committed
+/// write back to the federated source is tracked separately.
 async fn execute_transaction(
```
phillipleblanc, line 436: Stacked-PR review note: the transaction executor lives in `http/v1` here and handles a **single** write table. It is generalized to **multi-table** in #11865 (`TransactionHandle` + per-table participant map), and **extracted** out of `http/v1` into the shared `datafusion::query::transaction` module (used by both HTTP and FlightSQL) in #11866.

PR #11876, opened 2026-07-14
fix(mysql): retry binlog checkpoint upsert on transient accelerator write lock

krinart:
The MySQL binlog-position sidecar shares the accelerator's connection pool, so a checkpoint upsert can lose the single writer lock to a concurrent CDC-apply transaction and fail with `database is locked` (`SQLITE_BUSY` and equivalents). Previously that dropped the whole checkpoint interval, widening the crash-replay window.

`MySqlBinlogSys::upsert` now retries briefly (up to 5 attempts, ~375ms worst case) on transient lock errors before giving up. Best-effort as before — a persistent lock still returns and re-persists next interval — so correctness is unchanged.

### DIFF
--- a/crates/runtime/src/dataaccelerator/spice_sys/mysql_binlog/mod.rs
+++ b/crates/runtime/src/dataaccelerator/spice_sys/mysql_binlog/mod.rs
@@ -111,7 +111,41 @@ impl MySqlBinlogSys {
         }
     }
 
-    pub async fn upsert(
+    /// Persist a checkpoint, retrying briefly on a transient accelerator write
+    /// lock.
+    ///
+    /// The sidecar shares the accelerator's connection pool, so a checkpoint
+    /// upsert contends with the accelerator's own CDC-apply transactions for
+    /// the single writer lock. A large batch commit can hold that lock past the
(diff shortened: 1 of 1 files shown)


crates/runtime/src/dataaccelerator/spice_sys/mysql_binlog/mod.rs
```diff
@@ -111,7 +111,41 @@ impl MySqlBinlogSys {
...
+                Err(e) if attempt < UPSERT_MAX_ATTEMPTS && is_retryable_lock_error(&e) => {
+                    let delay = upsert_retry_delay(attempt);
+                    tracing::debug!(
+                        dataset = %self.dataset_name,
+                        attempt,
+                        delay_ms = %delay.as_millis(),
+                        error = %e,
+                        "binlog checkpoint upsert hit a transient accelerator write lock; retrying"
+                    );
+                    tokio::time::sleep(delay).await;
```
phillipleblanc: this should use the fibonacci retry utility we have instead of hand-rolling your own

bjchambers (approved):

phillipleblanc (approved):

PR #11877, opened 2026-07-14
Parallelize the HTAP analytical-query correctness gate

krinart:
Speeds up the HTAP analytical-query correctness gate by running it in parallel:

- Each query hits the source (Postgres/MySQL) and Spice concurrently instead of one after the other.
- Queries are fanned out over a channel to a pool of workers, controllable via a new `--analytic-gate-concurrency` flag (default 4).

Report order and gate semantics are unchanged.


tools/testoperator/src/commands/htap/correctness/analytical.rs
```diff
@@ -208,133 +218,190 @@ pub async fn verify_analytical_results(
...
+        let slots = &slots;
+        async move {
+            while let Ok((idx, query)) = rx.recv().await {
+                let result = evaluate_query(query, driver, spice).await;
+                slots.lock()[idx] = Some(result);
             }
-        };
+        }
+    });
+    futures::future::join_all(worker_futures).await;
```
phillipleblanc: It looks like you've manually implemented something that `StreamExt::buffered(n)` gives you for basically free

phillipleblanc: https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.buffered

PR #11908, opened 2026-07-16
Add `MemoryVectorIndex`: in-memory external-store `VectorIndex` with brute-force exact k-NN

Jeadie:
## Summary

Adds `MemoryVectorIndex` (`crates/search/src/index/memory/`), an in-memory, external-store implementation of the `VectorIndex` trait, plus a refactor extracting the store-agnostic write helpers out of the S3 Vectors index.

This index is intentionally **not** wired into spicepod `vectors.engine` dispatch or the runtime factory — it is a programmatic building block for upcoming composite type indexes.

## Design

Follows the `S3Vector` external-store model:

- **Write path**: `write()` embeds the search column via the held `Arc<dyn Embed>`, appends `{col}_embedding` (`FixedSizeList<Float32, N>`, columns alphabetically ordered to satisfy `DFSchema::logically_equivalent_names_and_types` vs `VectorScanTableProvider`), and upserts rows into an Arrow-native in-RAM store. Rows are keyed by the formatted primary key (single PK → Utf8 cast; composite → JSON) — **rewriting a key replaces the stored row**. Rows with a NULL primary key, NULL/empty search text, or all-zero/NaN vectors are skipped with a warning.
- **Store** (`store.rs`): rows held as `RecordBatch`es conforming to a fixed stored schema (PKs + metadata + embedding). Upsert filters superseded rows out of existing batches with the arrow `filter` kernel — zero-copy for untouched batches. Unbounded by design (eviction belongs to the caller); the store lock is never held across an await.
- **Read paths** (`provider.rs`): both `TableProvider`s read the store lazily at `scan()` time, so rows written after plan construction are visible (`VectorScanTableProvider` builds the list plan once at construction). `list_table_provider` enumerates the store; `query_table_provider` embeds the query lazily inside async `scan()` (cached in a `tokio::sync::OnceCell`, mirroring `S3VectorsQueryTable`) and brute-force scores every stored row, returning rows ordered by `_score` descending.
- **Scoring**: reuses the SIMD (`simsimd`) `InnerProduct` / `L2Distance` / `L2Norm` ScalarUDFs from `runtime-datafusion-udfs`, invoked directly over the `FixedSizeList<Float32>` arrays. `MemoryDistanceMetric { Cosine, L2, Dot }` uses the same score conventions as the other indexes (cosine → similarity, L2 → negated distance, dot → inner product); undefined similarities (zero-magnitude vectors) surface as NULL, never NaN, so failed embeddings cannot sort as top matches.

## Refactor

`extract_and_format_primary_key`, `embed_column`, `create_embedding_array`, `update_embedding_column_in_batch` (plus a new `sort_columns_alphabetically`) moved from `s3_vectors/write.rs` into `index/write_util.rs` with their own snafu error enum; the S3 write path delegates to them (behavior unchanged; existing unit tests moved along). `elasticsearch/write.rs` and `duckdb/write.rs` still carry private near-copies — consolidating them is a possible follow-up.

## Wiring

- Module gated behind the existing `llms` feature (enabled in default builds via `s3_vectors`); `MemoryVectorIndex` re-exported from `search::index` and added to `derived_columns_from_vector_index`.
- New deps: `parking_lot` (workspace) and optional `runtime-datafusion-udfs` (no dependency cycle; its default features are empty).

## Testing

Tests are deliberately deferred to a follow-up PR.


crates/search/src/index/memory/provider.rs
```diff
@@ -0,0 +1,231 @@
...
+        Ok(schema
+            .fields()
+            .iter()
+            .map(|field| col(field.name().clone()))
+            .collect())
+    }
+}
+
+#[async_trait]
+impl TableProvider for MemoryVectorQueryTable {
```
phillipleblanc, line 259: This should have the clippy trait to deny unimplemented members

phillipleblanc: And this should implement all of the default methods in the table provider trait.

crates/search/src/index/memory/provider.rs
```diff
@@ -0,0 +1,231 @@
...
+
+impl MemoryVectorListTable {
+    pub(crate) fn new(store: Arc<RwLock<MemoryVectorStore>>) -> Self {
+        let schema = Arc::clone(&store.read().stored_schema);
+        Self { store, schema }
+    }
+}
+
+#[async_trait]
+impl TableProvider for MemoryVectorListTable {
```
phillipleblanc, line 63: same as below - implement all default methods and add the clippy trait to deny unimplemented methods.

phillipleblanc (dismissed):

peasee (dismissed):

lukekim (approved):

PR #11914, opened 2026-07-16
feat(search): warm in-memory writethrough/fallback index for `.vectors` datasets and views

Jeadie:
Every external vector-engine index (S3 Vectors, Elasticsearch) now has a warm, in-memory tier index. No UX change in the spicepod.yaml; this is a **net-new default**. Writes go to both the in-memory and underlying vector index. Search queries are served from the warm tier. With the exact same mechanisms as accelerations atop federated data sources, the external index can be queried if no results are found in the warm index (only when `.datasets[].acceleration.on_zero_results: use_source` set)


Jeadie:
TODO: gate fallback logic when making warm tier: https://github.com/spiceai/spiceai/pull/11914/changes#diff-75b2cab2cee8512dd5862aaa6f27b498aa6694340d0976163c23ef15d242ebf3R84

phillipleblanc (dismissed):

lukekim (dismissed):

phillipleblanc (approved):

PR #11917, opened 2026-07-16
feat(cloud-connect): allow a self-hosted CA or plaintext for the standalone client

phillipleblanc:
## What

`CloudConnectConfig::from_env` now honors two env vars so a standalone `spiced` can reach a **self-hosted or local** Cloud Connect gateway:

- `SPICE_CLOUD_CA_CERT_PATH` — PEM CA file pinned as the server root (loaded into `ca_cert_pem`). Self-hosted control planes serve a certificate the public WebPKI roots don't trust.
- `SPICE_CLOUD_INSECURE=true` — disable TLS entirely (plaintext h2c) for local development.

## Why

Previously `from_env` hard-set `insecure: false` and never populated `ca_cert_pem`, so the client only trusted the system WebPKI roots. There was no way to pin a dev/self-hosted CA or opt out of TLS — a standalone instance could only connect to a control plane with a publicly-trusted certificate. The struct fields already existed and are exercised by the `standalone_adoption_e2e` test (which builds the config in-process); this just wires them to the environment.

## Testing

Built `spiced` release and ran a full local adoption E2E against a self-hosted gateway with `SPICE_CLOUD_INSECURE=true`: mint code → connect → adopt → run-query → deploy-spicepod → events, all green.


phillipleblanc:
Superseded by #11926. The self-hosted-CA / insecure-transport changes here were for a local demo and are no longer needed; the adoption-code validator fix is split out into #11926.

PR #11924, opened 2026-07-17
test(htap): print row-window context on analytical divergence

sgrebnov:
## Problem

When the HTAP analytical gate diverges, it reports a single `DataMismatch` cell and nothing else:

```
chbench_q10  DIVERGE  DataMismatch { column: "c_city", row_number: 48348, expected: "9rahWs5Q4H", actual: "Qg8T0oFQE4T" } (source rows=24373970, spice rows=24373970)
```

One scalar pair, no surrounding rows, no other columns of that row. This can't distinguish the two common failure modes:
- a **genuinely wrong value** in one cell, vs.
- a **sort-tie row shift** — both sides are sorted before comparison, but a tie broke differently, so every row from the divergence onward is off-by-one.

The existing full-batch `pretty_format_batches` fallback is useless at 24M rows (floods/truncates CI logs).

## Change

- Add `format_row_window()` to the shared validation module. At the failure point both sides are already schema-aligned and identically sorted, so global row *N* addresses the same logical row on each side. It slices a `±radius` window (clamped per-side so a row-count mismatch can't slice out of bounds) and pretty-prints the full rows — all columns, so key/identity columns stay visible — from both the source and Spice sides.
- Wire it into the analytical gate: on a `DataMismatch` divergence, print the window inline at detection. The compact `DIVERGE` tree line is unchanged.
- Default radius **10** (→ 21-row window), overridable via `SPICE_HTAP_DIVERGENCE_WINDOW`.

### Example output

```
chbench_q10: ↳ divergence at global row 48348, column 'c_city' (±10-row window)
  EXPECTED (source) — rows 48338..=48358:
  +-------+-----------+---------+ ...
  ACTUAL (spice) — rows 48338..=48358:
  +-------+-----------+---------+ ...
```

One glance distinguishes shift-vs-wrong-value.

## Tests

- Unit test for the normal window + an end-of-batch clamp test.
- **insta snapshot** pinning the exact end-to-end render (multi-column rows, both labelled `±3` windows, the diverging cell).

## Scope

htap analytical gate only. `worker.rs` has the same 24M-row full-dump antipattern on the reference-validation path; the helper is public and ready to wire there in a follow-up.

### DIFF
--- a/crates/test-framework/src/queries/validation/mod.rs
+++ b/crates/test-framework/src/queries/validation/mod.rs
@@ -514,6 +514,89 @@ pub fn validate_batches_as_strings(
     Ok(QueryValidationResult::Pass)
 }
 
+/// Render a ±`radius`-row window around a diverging row from both the expected
+/// and actual result sets, stacked and labelled, for divergence diagnostics.
+///
+/// A single [`QueryValidationFailReason::DataMismatch`] reports one scalar cell
+/// with no surrounding context, so it can't distinguish a genuinely wrong value
+/// from a row-alignment/sort-tie shift (where every row from the divergence
+/// onward is off-by-one). Printing the neighbouring rows — full row, all columns,
+/// so key columns are visible — from both sides makes the failure mode obvious at
+/// a glance.
+///
+/// `row_number` is the 1-based row index carried by `DataMismatch` (indexing the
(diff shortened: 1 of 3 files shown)


sgrebnov:
Close in favor of 
https://github.com/spiceai/spiceai/pull/11923

PR #11929, opened 2026-07-17
fix(signoff): let remote sign-off refresh the Attestation check in CI

bjchambers:
## Problem

The **Remote Sign-off** workflow (`signoff.yml`) runs `scripts/signoff -f`. After posting the `signoff` commit status, the script calls `refresh_attestation_check`, which re-runs the PR's completed `pr.yml` run so the **Attestation** check re-evaluates the freshly-posted sign-off — the same thing a developer's CLI does.

In CI that refresh silently failed:

```
✓ Signed off on <sha> — ...
  Could not auto-refresh the 'Attestation' check — Open/refresh the PR and it will pick up the sign-off.
```

So a remote sign-off left Attestation stuck on its pre-sign-off failure, requiring a manual re-run.

## Cause

`gh run rerun` hits the Actions "re-run a workflow" API, which requires **`actions: write`**. The workflow only granted `contents: read` + `statuses: write`, so the rerun 403'd. (The CLI path works because a developer token has the scope.)

## Fix

- Grant `actions: write` in `signoff.yml` so the remote path matches the local one.
- Harden the sign-off author lookup in `scripts/signoff`: `gh api user` needs a user-scoped token and 403s under the CI `GITHUB_TOKEN`, leaking the error JSON into the status description (`Signed off by {"message":"Resource not accessible by integration"...`). Accept the result only if it looks like a GitHub login; otherwise fall back to `GITHUB_ACTOR` (CI) or the VCS-configured name.

## On the recursion guard

Re-running an existing run via `actions: write` is allowed for `GITHUB_TOKEN`. The *"GITHUB_TOKEN events don't trigger new workflow runs"* rule applies only to **event side-effects** (a token pushing a commit / opening a PR), not to an explicit rerun API call.

## Verification

Verified empirically by dispatching this workflow **from this branch's ref** (so the updated definition + permissions run) against a branch whose Attestation had already failed, and confirming Attestation flips green with no manual rerun. (Details in a comment below once the run completes.)

### DIFF
--- a/.github/workflows/signoff.yml
+++ b/.github/workflows/signoff.yml
@@ -36,6 +36,11 @@ jobs:
     permissions:
       contents: read
       statuses: write
+      # scripts/signoff re-runs the PR's completed pr.yml run so the
+      # 'Attestation' check re-evaluates the sign-off status it just posted;
+      # the "re-run a workflow" API needs actions: write. Without it the
+      # refresh 403s and Attestation stays stuck on its pre-sign-off failure.
+      actions: write
     env:
       HAS_SPICEIO_SECRET: ${{ secrets.UNAS_SMB_PASS != '' }}
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/scripts/signoff
+++ b/scripts/signoff
@@ -308,7 +308,15 @@ cmd_signoff() {
   revision=$(current_revision "$force")
   $force || ensure_clean_and_pushed "$revision"
   sha=$(revision_sha "$revision") || fail "could not resolve ${revision}"
-  user=$(gh api user --jq .login 2>/dev/null || local_user_name || echo "unknown")
+  # `gh api user` needs a user-scoped token; under a CI GITHUB_TOKEN it 403s and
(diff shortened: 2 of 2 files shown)


sgrebnov (approved):

bjchambers:
## Verified end-to-end ✅

Dispatched Remote Sign-off **from this branch's ref** (so the updated definition + `actions: write` run) against this PR, whose Attestation had already failed pre-sign-off. Result from the `Sign off` step:

```
✓ Signed off on 5947be71fdca — Signed off by bjchambers — lint+test passed in 8079s
  Re-ran the 'Attestation' check (run 29613729237) — it will turn green shortly.
```

- `Re-ran the 'Attestation' check` = the `gh run rerun` succeeded (was "Could not auto-refresh" before) → `actions: write` is the fix.
- Attestation then flipped to **pass** with no manual intervention.
- `Signed off by bjchambers` confirms the author-lookup hardening (was leaking the `gh api user` 403 JSON under the CI token).

Confirms a `GITHUB_TOKEN` with `actions: write` can re-run an existing run — the "GITHUB_TOKEN events don't trigger new runs" guard applies only to event side-effects, not explicit reruns.

PR #11934, opened 2026-07-18
ci(build_and_release): add spiced_only dispatch option for htap benchmarks

bjchambers:
## What

Adds a `build_option` input to the `build_and_release` `workflow_dispatch`. When set to `spiced_only` (default remains `all`), the workflow builds **only** the Linux x64 release `spiced` binary and uploads it to MinIO — skipping the `spice` CLI, the `odbc` and `metal` variants, and all other platforms.

## Why

htap benchmarks (`testoperator_run_htap.yml`) consume exactly one artifact from a `build_and_release` run: the Linux x64 `spiced` tarball fetched from MinIO by the `setup-spiced` action (`artifacts/spiced/linux-x86_64/<sha>/spiced_linux_x86_64.tar.gz`). Producing the full matrix (odbc, metal, the CLI, macOS/Windows/ARM) just to feed a benchmark wastes a lot of build time. This option builds only what the benchmark needs.

## How

- New `build_option: all | spiced_only` dispatch input.
- When `spiced_only`, the matrix-setup script returns a single `Linux x64` entry (`tags: ['default']`, `spiced_only: true`), overriding `platform_option`/`tag_option`.
- A dedicated build step compiles `-p spiced` only (the default step builds `-p spiced -p spice`); the `spice` CLI packaging/version/upload steps are gated off with `!matrix.target.spiced_only` since that binary isn't built.
- The existing `spiced` tar + `Upload spiced to Minio` steps are unchanged and still run, producing the artifact the benchmark fetches.

`all` builds, trunk pushes, and tagged releases are unaffected. Matrix logic verified against `spiced_only`, `all`, and trunk-push inputs; YAML validated.

### DIFF
--- a/.github/workflows/build_and_release.yml
+++ b/.github/workflows/build_and_release.yml
@@ -57,6 +57,14 @@ on:
           - 'release'
           - 'release-lto'
         default: 'release'
+      build_option:
+        description: 'What to build? "all" (default) or "spiced_only" (Linux x64 release spiced only, for htap benchmarks — skips the spice CLI, odbc, metal, and all other platforms)'
+        required: false
+        type: choice
+        options:
+          - 'all'
+          - 'spiced_only'
+        default: 'all'
 
 env:
   # CI performance optimizations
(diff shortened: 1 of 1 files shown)


phillipleblanc (approved):

Jeadie:
if I understand, `tag_option: default` does not build models, ODBC etc. It does still make CLI, however I assume most of the latency is 3 rebuilds of spiced (i.e. default, models, ODBC). 

If that's true, this has less marginal value than using existing inputs?

bjchambers:
> if I understand, `tag_option: default` does not build models, ODBC etc. It does still make CLI, however I assume most of the latency is 3 rebuilds of spiced (i.e. default, models, ODBC).
> 
> If that's true, this has less marginal value than using existing inputs?

@Jeadie Hmm. You're right, but I still see it including ODBC quite often. Perhaps this is because the default includues it?

PR #11939, opened 2026-07-20
Validate vector search parameters before SQL query

### LINKED ISSUE #4884: Validate parameters of vector search before SQL query.
Currently, a vector search operation takes the following payload
```rust
pub struct SearchRequestBaseJson {
    /// The text to search documents for similarity
    pub text: String,

    /// The datasets to search for similarity. If None, search across all datasets. For available datasets, use the `list_datasets` tool and ensure `can_search_documents==true`.
    #[serde(default)]
    pub datasets: Option<Vec<String>>,

    /// Number of documents to return for each dataset
    #[serde(default)]
    pub limit: Option<usize>,

    /// An SQL filter predicate to apply. Format: 'WHERE `where_cond`'.
    #[serde(rename = "where", default)]
    pub where_cond: Option<String>,

    /// Additional columns to return from the dataset.
    #[serde(default)]
    pub additional_columns: Vec<String>,
}
```
Some of these fields can be validated prior to attempting any SQL query to perform the search: 
 - `datasets`: ensure the datasets exist and specifically they have an appropriate embedding column to perform a vector search upon.
 - `additional_columns`: ensure the columns provide exist in the `datasets` provided. 

Doing so, we can provide a much more user-friendly error to the user or LLM.

lesbass:
## Description

Add early validation of `datasets` and `additional_columns` in the vector search path before any SQL query is executed, as suggested in #4884.

## Changes

**`crates/runtime-search/src/search_engine.rs`**
- Added `validate_request()` method to `SearchEngine` that runs before parallel search pipeline execution
- Checks each explicitly-requested dataset exists, returning `DataSourcesNotFound` if not found
- Checks each explicitly-requested dataset has embedding columns or full-text search indexes, returning `CannotSearchDataset` if not
- Checks each `additional_column` exists in the target dataset schema, returning `AdditionalColumnNotFound` with a list of available columns

**`crates/runtime-search/src/error.rs`**
- Added `AdditionalColumnNotFound` error variant with dataset name, column name, and available columns

## Verification

Unable to run full build locally due to missing C compiler in the environment (`cc` not found). The changes have been reviewed for correctness:
- Function signatures are consistent with existing patterns
- Error types follow existing `Snafu` conventions
- All types (`Column`, `TableReference`, etc.) are already in scope via existing imports
- `SPICE_DEFAULT_CATALOG` / `SPICE_DEFAULT_SCHEMA` are consts already defined in the same file

Closes #4884

### DIFF
--- a/crates/runtime-search/src/search_engine.rs
+++ b/crates/runtime-search/src/search_engine.rs
@@ -22,7 +22,11 @@ use std::{collections::HashMap, sync::Arc};
 use super::embeddings::table::EmbeddingTable;
 use crate::candidate::vector::ChunkedNonIndexVectorGeneration;
 use crate::candidate::vector_udtf::VectorUDTFGeneration;
-use crate::error::{DataFusionSnafu, Error, FormattingSnafu, Result, SearchPipelineSnafu};
+use crate::error::{
+    AdditionalColumnNotFoundSnafu, CannotSearchDatasetSnafu, DataFusionSnafu, Error,
+    FormattingSnafu, Result, SearchPipelineSnafu,
+};
+use snafu::ensure;
 use crate::table_provider_explorer::TableProviderExplorer;
 
 pub const SPICE_DEFAULT_CATALOG: &str = "spice";
@@ -110,19 +114,7 @@ impl<E: TableProviderExplorer> SearchEngine<E> {
     async fn user_tables_that_can_search(&self) -> Result<Vec<TableReference>> {
         let mut searchable_tables = Vec::new();
         for t in self.df.get_user_table_names() {
-            if self
-                .embedding_columns_from_table(&t)
-                .await
-                .is_some_and(|cols| !cols.is_empty())
-            {
-                searchable_tables.push(t);
-                continue;
-            }
-            if self
-                .full_text_search_candidates(&t)
-                .await
-                .is_some_and(|fts_res| fts_res.is_ok_and(|c| !c.is_empty()))
-            {
+            if self.table_can_search(&t).await {
                 searchable_tables.push(t);
             }
         }
@@ -227,6 +219,16 @@ impl<E: TableProviderExplorer> SearchEngine<E> {
         }
 
         Some(Ok(vec![]))
+    }
+
+    async fn table_can_search(&self, tbl: &TableReference) -> bool {
+        self.embedding_columns_from_table(tbl)
+            .await
+            .is_some_and(|cols| !cols.is_empty())
+            || self
+                .full_text_search_candidates(tbl)
+                .await
+                .is_some_and(|res| res.is_ok_and(|c| !c.is_empty()))
     }
 
     fn get_vector_index(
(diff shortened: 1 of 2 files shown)


crates/runtime-search/src/search_engine.rs
```diff
@@ -363,6 +363,67 @@ impl<E: TableProviderExplorer> SearchEngine<E> {
...
+    ) -> Result<()> {
+        for tbl in tables {
+            let table_provider = self.df.get_table(tbl).await.ok_or_else(|| {
+                Error::DataSourcesNotFound {
+                    data_source: vec![tbl.clone()],
+                }
+            })?;
+
+            if explicit_datasets_requested {
+                let embedding_columns = self.embedding_columns_from_table(tbl).await;
```
Jeadie: nit: we should refactor into function and share this logic with `SearchEngine::user_tables_that_can_search`

crates/runtime-search/src/search_engine.rs
```diff
@@ -363,6 +363,67 @@ impl<E: TableProviderExplorer> SearchEngine<E> {
...
+                let fts_candidates = self.full_text_search_candidates(tbl).await;
+
+                let has_search_capability = embedding_columns
+                    .as_ref()
+                    .is_some_and(|cols| !cols.is_empty())
+                    || fts_candidates
+                        .as_ref()
+                        .is_some_and(|res| res.as_ref().is_ok_and(|c| !c.is_empty()));
+
+                if !has_search_capability {
```
Jeadie: try `ensure!(` (e.g. `ensure!(!data.is_empty(), DataEmptySnafu);`)

crates/runtime-search/src/search_engine.rs
```diff
@@ -363,6 +363,67 @@ impl<E: TableProviderExplorer> SearchEngine<E> {
...
+                let col_applies =
+                    col.relation.as_ref().is_none_or(|rel| {
+                        tbl.clone()
+                            .resolve(SPICE_DEFAULT_CATALOG, SPICE_DEFAULT_SCHEMA)
+                            == rel
+                                .clone()
+                                .resolve(SPICE_DEFAULT_CATALOG, SPICE_DEFAULT_SCHEMA)
+                    });
+
+                if col_applies && schema.column_with_name(&col.name).is_none() {
```
Jeadie: as above (ensure!)

Jeadie (commented):
Thank you for this. Some of this validation is done within the flow (and before SQL), but `validate_request` is the right abstraction. 

Couple of code fixes suggested

claudespice:
👋 Thanks @lesbass for your first contribution to spiceai!

Automated new-contributor check:
- Category: ✍️ AI-assisted contribution from a human contributor (Medium confidence)
- Reason: the PR description is machine-generated and notes it was produced in an environment without a C compiler and self-reviewed; the account itself shows a genuine, long-running public GitHub history.

Provenance note, not a problem — a maintainer will review as usual.

crates/runtime-search/src/search_engine.rs
lesbass: Done: extracted the shared logic into a new `table_can_search` method used by both `user_tables_that_can_search` and `validate_request`.

crates/runtime-search/src/search_engine.rs
lesbass: Done: replaced the `if !condition { return Err(...) }` patterns with `ensure!(` macro calls.

crates/runtime-search/src/search_engine.rs
lesbass: Done: switched to `ensure!(` here as well.

lesbass:
Thanks for the review @Jeadie. I've addressed all three suggestions:

- Extracted the shared search-capability check into a `table_can_search` helper method, used by both `user_tables_that_can_search` and `validate_request`.
- Replaced the explicit `if !condition { return Err(...) }` patterns with `ensure!(` macro calls for the error branches in `validate_request`.

PTAL when you have a chance.

lesbass:
Rebased branch required recreation due to scope limitations. Continued in #12201.

Issue #11942, opened 2026-07-20
Vortex pin is the yanked 0.76.0: TableStrategy hides parallel-writer panics (silent file corruption); likely root cause of the chbench_q10 divergence behind #11910


lukekim:
## Summary

Our vortex fork pin `409505dfd` is upstream **vortex 0.76.0 — a release upstream yanked** for a latent correctness bug: `TableStrategy` spawns the column-chunk fan-out task detached and discards every channel-send result, so a panic/death in the parallel write pipeline lets surviving column writers see a clean EOF and finalize columns at different lengths **silently** (file still reports full row count). Upstream fix: vortex-data/vortex#8672 (in 0.77.0). Every Cayenne Vortex write (delta, mem-tier checkpoint, compaction output) goes through this writer.

This is the probable root cause of the SF1000 `chbench_q10` content divergence (`customer.c_city` plausible-wrong string, equal row counts) that forced #11910 to revert #11826 (light delta encoding + CDC coalescing — the measured SF1000 replication-convergence lever). #11826 amplified traffic through the vulnerable writer (every large checkpoint light+parallel); the stronger content fingerprint (#11901) made the corruption detectable. The revert removed the amplifier, not the bug.

## Evidence

- Upstream 0.77.0 release notes: 0.76.0 yanked; fix #8672 ("TableStrategy currently hides panic in spawned column writer tasks").
- Buggy shape verified at our pin: `vortex-layout/src/layouts/table.rs:279-292` (`.detach()` + ignored `tx.send`).
- Deterministic alternative (light-scheme encode/decode bug) **refuted by measurement**: 417,792-row × 16-column round-trip through the exact level-1/2/3 scheme sets + FULL control (dict-cardinality boundaries 255/256/257/65535/65536/65537/100k, near-unique + long out-of-line strings, 85%-null sparse, constants, sliced batches, 262k-row jumbo) — **0 mismatches** at trunk `407bfbebba`.
- `c_city` is never UPDATEd by CH-benCH (Payment/Delivery touch balance/cnt/data), so the divergence cannot be same-row version churn — it requires cross-row contamination or a corrupt file.
- Cherry-pick of #8672 onto the pin applies clean and upstream's regression test passes on the fork base (`table_fanout_panic_propagates`).

## Fix plan

[...]

PR #11943, opened 2026-07-20
fix(vortex): bump pins past yanked 0.76.0 TableStrategy silent-corruption bug

lukekim:
## What

Bump the vortex fork pins from `409505dfd` (upstream **0.76.0 — a yanked release**) to fork branch `lukim/spiceai-54-tablestrategy-fix` (`68badd6899c82156a3a6d4c4cf9906051a674ece`) = the same base + a clean cherry-pick of upstream **vortex-data/vortex#8672**: *"fix: TableStrategy currently hides panic in spawned column writer tasks."*

Adds `crates/cayenne/tests/light_encode_roundtrip.rs`, a cell-exact round-trip regression harness over the light delta-encoding scheme sets (dict-cardinality boundaries, near-unique/long strings, null-dominated sparse, constants, sliced + jumbo batches; CI-sized by default, `LIGHT_ROUNDTRIP_LARGE=1` for the full audit dataset).

## Why

Upstream yanked 0.76.0 for a latent silent-write-corruption bug: `TableStrategy` runs its column-chunk fan-out task detached and discards channel-send results, so a panic/death mid-write lets surviving parallel column writers see a clean EOF and finalize columns at different lengths **silently** — the file still reports full row count. Every Cayenne Vortex write (staged deltas, mem-tier checkpoints, compaction outputs) goes through this writer.

This is the probable root cause of the SF1000 `chbench_q10` content divergence (`customer.c_city` plausible-wrong string, equal row counts) that forced #11910 to revert #11826:

- `c_city` is never UPDATEd by CH-benCH, so the divergence cannot be same-row version churn — it requires cross-row contamination or a corrupt file.
- The deterministic alternative (light-scheme encode/decode bug) is refuted by measurement: the included harness round-trips 417k rows × 16 columns through levels 1/2/3 + FULL with **0 mismatches** at `407bfbebba`.
- #11826 amplified traffic through the vulnerable writer (every large checkpoint light-encoded + parallel); the stronger content fingerprint (#11901) made the corruption detectable. The revert removed the amplifier — this PR removes the bug.

Details and full evidence chain: #11942.

## Scope / blast radius

Behavior change only when a writer subtask panics/errors mid-file: the write now **fails loudly** instead of persisting a silently-desynced file. No API or encoding change; absent panics the written bytes are identical.

## Tests

- Upstream regression test `table_fanout_panic_propagates` passes on the fork branch (`cargo test -p vortex-layout`).
- New `light_encode_roundtrip` integration test: 0 mismatches across levels 1/2/3 + FULL.
- `cargo clippy -p cayenne --tests --no-deps` clean.

## Follow-ups (tracked in #11942)

1. Re-land #11826 once this merges (draft PR open) — gated on an SF1000 content-fingerprint-validated run.
2. Full fork bump to vortex 0.79.0 (#8474 BitBufferMut overflow, #8667 constant-detection hardening, #8406 dict-layout probe with configured compressor).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21835,7 +21835,7 @@ dependencies = [
 [[package]]
 name = "vortex"
 version = "0.1.0"
-source = "git+https://github.com/spiceai/vortex.git?rev=409505dfd09e60d34088c36ba12e3fea34bb394f#409505dfd09e60d34088c36ba12e3fea34bb394f"
+source = "git+https://github.com/spiceai/vortex.git?rev=68badd6899c82156a3a6d4c4cf9906051a674ece#68badd6899c82156a3a6d4c4cf9906051a674ece"
 dependencies = [
  "vortex-alp",
  "vortex-array",
@@ -21869,7 +21869,7 @@ dependencies = [
 [[package]]
 name = "vortex-alp"
 version = "0.1.0"
-source = "git+https://github.com/spiceai/vortex.git?rev=409505dfd09e60d34088c36ba12e3fea34bb394f#409505dfd09e60d34088c36ba12e3fea34bb394f"
+source = "git+https://github.com/spiceai/vortex.git?rev=68badd6899c82156a3a6d4c4cf9906051a674ece#68badd6899c82156a3a6d4c4cf9906051a674ece"
 dependencies = [
  "itertools 0.14.0",
  "num-traits",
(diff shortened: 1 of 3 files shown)


lukekim:
Fork PR opened: spiceai/vortex#73 (`lukim/spiceai-54-tablestrategy-fix` → `spiceai-54`, clean fast-forward — the branch is based on the current `spiceai-54` tip `409505dfd`, which is exactly today's pin). Once it merges I'll bump this PR's pins from the side-branch SHA `68badd689` to the resulting `spiceai-54` SHA so the canonical line carries the fix and future fork bumps can't drop it.

PR #11946, opened 2026-07-20
chbench(mysql): parallel analytical gate, decimal-comparison fix, and reseed/CI hardening

krinart:
- **Parallel analytical-query gate.** Run the CH-benCHmark analytical correctness/QPH queries concurrently via `StreamExt::buffered`, configurable with a new `--analytic-gate-concurrency` (defaulted to 16 for MySQL), instead of one-at-a-time. Adds per-query and per-side (source vs Spice) timing logs so a slow or hung query is immediately attributable.

- **Decimal comparison correctness.** `approximate_columns` now treats scale-inflated `AVG`/division decimals (common scale > money scale) as approximate, fixing false `DIVERGE`s on MySQL where the source and DataFusion both emit scale-6 and differ only by a rounding ULP (e.g. `chbench_q1 avg_amount`). Money `SUM`s and counts stay on the exact path, and mismatch context now prints the surrounding rows from both sides.

- **`chbench-skip-slow` query override.** New override category that drops the heaviest analytical queries (q10, q18) from the gate and QPH set, for large scale factors where they dominate wall-clock.

- **MySQL reseed DROP hardening.** Bound `lock_wait_timeout` and kill stale sessions before `DROP TABLE`, plus per-table drop timing — so a drop blocked on a leaked lock (e.g. from an abruptly-killed prior run) fails in seconds with a clear error instead of silently consuming the whole CI job budget.

- **Replication-drain visibility.** Per-table row-count and `max_bench_ts`-lag logging during the drain gate, so a table that isn't converging is visible per poll rather than opaque.

- **CI robustness.** Adds a `timeout-minutes` backstop to the seed/restore step and hardens the MySQL template snapshot/restore path so a wedged seed fails attributably rather than at the job timeout.

### DIFF
--- a/tools/chbench-driver/src/schema_mysql.rs
+++ b/tools/chbench-driver/src/schema_mysql.rs
@@ -21,6 +21,8 @@ limitations under the License.
 //! dialect DDL. Postgres-WAL-specific replication cleanup has no `MySQL`
 //! equivalent and is omitted.
 
+use std::time::Instant;
+
 use mysql_async::prelude::Queryable;
 
 use crate::Result;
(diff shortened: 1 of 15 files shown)


tools/chbench-driver/src/schema_mysql.rs
```diff
@@ -37,6 +39,44 @@ use crate::Result;
 pub async fn drop_tables(conn: &mut mysql_async::Conn) -> Result<()> {
     println!("  dropping {} tables", crate::schema::ALL_TABLES.len());
 
+    // Bound metadata-lock waits. A `DROP TABLE` blocked by another session holding
+    // a lock otherwise waits up to the server default `lock_wait_timeout`
+    // (31536000s ≈ 1 year), so a single stuck drop silently consumes the whole CI
+    // job budget. 60s makes a blocked drop fail fast with a clear
+    // "Lock wait timeout exceeded" error instead of hanging.
+    conn.query_drop("SET SESSION lock_wait_timeout = 60")
```
sgrebnov, line 47: Why we need this?

krinart: When we drop a table, it requires exclusive lock. The default timeout is one year. Here we set it to 60 seconds to fail fast when lock is not available.

tools/testoperator/src/commands/htap/correctness/compare.rs
```diff
@@ -39,6 +42,13 @@ use arrow::datatypes::DataType;
 /// so a genuine sub-5% value drift is now caught instead of passing silently.
 pub const FLOAT_REL_TOLERANCE: f64 = 0.001;
 
+/// Decimal scale of TPC-C money columns (`NUMERIC(_,2)` — cents). Exact
+/// aggregates (`SUM`/`MIN`/`MAX`) preserve this scale, so they stay on the exact
+/// comparison path; only `AVG`/division inflates a decimal result beyond it.
+/// A decimal column whose scale exceeds this is therefore treated as an
+/// approximate (rounding-prone) aggregate. See [`approximate_columns`].
+pub const MONEY_SCALE: i8 = 2;
```
sgrebnov: Should `FLOAT_REL_TOLERANCE` cover this?

sgrebnov (dismissed):

sgrebnov (approved):

Issue #11930, opened 2026-07-17
Enhancement: Support PostgreSQL REPLICA IDENTITY modes


bjchambers:
## Goal-State/What/Result

`refresh_mode: changes` on the PostgreSQL native-CDC path correctly, testably, and documentedly supports every PostgreSQL `REPLICA IDENTITY` mode a source table can present. Concretely:

1. **`USING INDEX` is properly supported** — validated at startup, able to key CDC apply from the nominated unique index (including when no `primary_key` is declared on the dataset), tested, and documented.
2. **`FULL` is documented and its coverage rounded out** — it already works (unit + integration), so this fills gaps (a key-narrowing unit test, the catalog-path + large-scale matrix, a perf comparison) and adds docs on when to use it and what it costs.
3. **`DEFAULT` and `NOTHING`** are handled with clear, early errors and documented (largely already implemented; `NOTHING` and keyless `DEFAULT` are unsupportable, not just unsupported).

## Background

`REPLICA IDENTITY` is a per-table PostgreSQL property that controls what the WAL carries in the *old tuple* of an `UPDATE`/`DELETE`, which is what any logical-replication consumer uses to identify the affected row. There are four modes:

| Mode | Old tuple in WAL (UPDATE/DELETE) | Identity used for CDC apply | Notes |
|---|---|---|---|
| `DEFAULT` | primary-key columns | the table's primary key | Requires a PK. Recommended by most CDC consumers. With no PK, `UPDATE`/`DELETE` cannot be replicated (Postgres errors at the source). |
| `USING INDEX` | the nominated index's columns | that unique index | Lets a table use a separate unique index as its identity — useful when there is no formal PK. Postgres requires the index be UNIQUE, non-partial, non-deferrable, on `NOT NULL` columns. |
| `FULL` | the entire old row | any/all columns | Largest WAL + wire + decode cost; discouraged for wide rows by most CDC consumers. Uniquely enables recovery of unchanged TOASTed columns from the stream (see below). |
| `NOTHING` | nothing | none | No identity is logged. `INSERT`s still replicate, but `UPDATE`/`DELETE` cannot — and if the table is in a publication that replicates updates/deletes, the **source** rejects those statements at execution time (a publisher-side error on the application's own writes, not a walsender refusal). A faithful CDC replica is impossible. |

[...]

PR #11951, opened 2026-07-20
feat(catalog): replica-identity-aware eligibility for PostgreSQL catalog CDC

bjchambers:
## Summary

Stacked on #11897. Addresses #11930 (catalog-path scope).

Point a PostgreSQL catalog connector at **any** database and every table that can be CDC-replicated is, while the rest are skipped with an actionable warning — instead of failing the whole catalog when a table lacks a primary key. Each discovered table is classified by its source `REPLICA IDENTITY`:

| Mode | Key state | Outcome |
|------|-----------|---------|
| `DEFAULT` | primary key | accelerate, keyed by the PK |
| `USING INDEX` | usable unique index | accelerate, keyed by the index columns |
| `FULL` | primary key | accelerate (warns — heavier: full old-row image per UPDATE/DELETE) |
| `NOTHING` / keyless `DEFAULT` / `FULL` w/o key / unusable `USING INDEX` | — | **skip** with a warning; absent from the catalog namespace (`exclude` to suppress) |

Skipped tables can be `exclude`d to silence the warning and handled via federation and/or a per-dataset `refresh_mode: full`.

## Implementation

- `connector-postgres-common`: a `replica_identity(...)` query (one round-trip: mode + primary-key columns + `USING INDEX` identity columns) and a pure, exhaustively unit-tested `classify_replica_identity(...)`.
- The synthesized per-table dataset now declares the resolved key **explicitly** (`primary_key` + `on_conflict` upsert). This is what makes `USING INDEX` work end-to-end — schema inference only derives a key from a formal primary key — and it satisfies the CDC path's upsert-on-key requirement.
- Per-table info/warn messages describe how each table replicates; the catalog summary reports accelerated / excluded / skipped counts.

## Testing

- Unit: full `classify_replica_identity` matrix over mode × key-presence.
- Integration (`catalog_changes.rs`, Docker PostgreSQL): a per-mode matrix (DEFAULT/USING INDEX/FULL eligible; NOTHING/keyless skipped) **plus a USING INDEX CDC-convergence test** proving live insert/update/delete route correctly by the unique index. All catalog integration tests pass locally; no regression to existing ones.

## Scope

The shared-CDC-path `USING INDEX` inference for **hand-written** datasets (no declared `primary_key`) remains a separate follow-up (#11930 spec B); the catalog path sidesteps it by always declaring the key.


crates/data-connectors/connector-postgres-common/src/lib.rs
```diff
@@ -266,3 +271,417 @@ pub async fn check_cdc_prerequisites(pool: &PostgresConnectionPool) -> Result<()
...
+}
+
+/// The catalog-acceleration eligibility decision for a table, derived purely
+/// from its [`ReplicaIdentity`]. The `Accelerate*` variants carry the resolved
+/// upsert `key` the synthesized dataset must declare (schema inference will not
+/// derive a `USING INDEX` key, so the caller declares it explicitly).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ReplicaIdentityOutcome {
+    /// `DEFAULT` + primary key: replicate keyed by the primary key.
+    AccelerateViaPrimaryKey { key: Vec<String> },
```
phillipleblanc, line 473: Would changing the `String` to `Arc<str>` make sense here?

phillipleblanc (dismissed):

crates/data-connectors/connector-postgres-common/src/lib.rs
bjchambers: I don't think so at this time.

1. It is a relatively cold path -- catalog refresh, once per table, not a hot loop.
2. The keys tend to be a few short column-names, which are cloned once (relatively cheap) and joined into one string. No sharing or repeated cloning where `Arc<str>` would pay off.
3. Consistency -- the source and other similar code (`primary_key_columns`) currently use `Vec<String>` so this would be a high-churn change.

lukekim (approved):

PR #11954, opened 2026-07-21
feat(params): typed params for VectorStore and FtsStore engines

Jeadie:
Follow-up to #11809, move `.datasets[].vectors.params` and `.datasets[].fts_store.params` to use `#[derive(TypedParams)]` from the `runtime-parameters-derive` crate. 

Notes:
- S3 AWS auth is unchanged. We share a common `initiate_config_with_credentials` function across several callosities (Glue, Iceberg, etc). The `TypedParams` includes these parameters for earlier and more digestible validation. 

## Behavior changes (intentional)

- Malformed numeric/duration values that previously warned-and-fell-back (`s3_vectors_batch_write_rows`, FTS `batch_write_rows`/timeouts) are now structured errors. 
- Elasticsearch `number_of_shards`/`number_of_replicas` are now validated as integers at registration (instead of being arbitrarily sent to ES).
- ElasticSearch full text search `user`/`pass` params are now secrets.
- `s3_vectors_distance_metric` no longer autoloads from secret stores (it is not a secret); it must be set in the spicepod. An invalid value still errors either way.


crates/runtime-search/src/store_params/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+pub mod s3;
+
+/// Parses a user-facing boolean parameter (`#[param(parse_with = ...)]`
+/// parser), accepting the historically lenient forms: `true`/`1`/`yes` and
+/// `false`/`0`/`no`, case-insensitive and trimmed.
+///
+/// # Errors
+///
+/// Returns an error when the value is none of the accepted forms.
+pub fn parse_bool(raw: &str) -> Result<bool, String> {
```
phillipleblanc, line 39: maybe this should live in the `runtime_parameters` crate?

phillipleblanc (approved):

PR #11960, opened 2026-07-21
refactor(layering): extract DynamoDB into connector-dynamodb

bjchambers:
## What

Extracts the DynamoDB connector out of `runtime` + `data_components` into a new
`connector-dynamodb` crate — the "a feature is a crate" principle, mirroring the
ClickHouse extraction. The connector body (`dataconnector/dynamodb.rs`) and all of
`data_components/src/dynamodb` (~10.3k LOC: provider, streams, arrow, dml, schema)
move into the crate; the shipped `runtime` and `data_components` no longer carry any
DynamoDB code or AWS SDK drivers.

## Series

Third step of the crate-layering work, on top of the merged #11935 (prepare DynamoDB
for extraction) and #11938 (checkpoint sidecar as per-engine crates). Rebased onto
trunk after #11938 merged.

## Checkpoint seam

Resolving a dataset's accelerator into a checkpoint store reaches through the
runtime-internal accelerator engine registry, so that seam stays in `runtime`:
`dataaccelerator::spice_sys::dynamodb::init_checkpoint_store(dataset)` is a thin
wrapper over the `checkpoint_store` factory (which builds the matching per-engine
sidecar store crate from #11938). The connector holds only the opaque
`BlobCheckpointStore` trait object, so it never depends on accelerator internals or
the per-engine store crates. The seam module is driver-free and always compiled.

## Feature / dep cleanup

- Delete `data_components/dynamodb` and its `aws-*` / `dynamodb-streams` optional deps.
- `runtime` `dynamodb` becomes a test-only marker (`[]`) gating `tests/dynamodb`; its
  `aws-sdk-dynamodb` moves to a dev-dependency. `initiate_config_with_auth_method` is
  un-gated as the generic AWS-auth helper it is.
- `spiced` `dynamodb = ["connector-dynamodb"]` (was `["runtime/dynamodb"]`).
- Registration switches to explicit `register_connector_factory("dynamodb", ...)`.
- `layers.toml` `restricted_deps` confine `aws-sdk-dynamodb(streams)` to
  `connector-dynamodb` (+ `dynamodb-streams` / `testoperator` / `spidapter`); a
  normal-dep regression back into `runtime`/`data_components` fails the guard.

## Validation

- `cargo check -p connector-dynamodb -p runtime --features runtime/{duckdb,sqlite,turso,postgres-accel}` — green.
- crate-layering guard green (131 crates, 6 tiers, no upward edges).
- Full `spiced` build + the docker DynamoDB integration suite are the end-to-end guard (signoff / CI).


tools/spicepodschema/src/collector.rs
```diff
@@ -34,6 +34,7 @@ use connector_clickhouse as _;
 use connector_delta_lake as _;
 use connector_dremio as _;
 use connector_duckdb as _;
+use connector_dynamodb as _;
```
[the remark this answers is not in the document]
bjchambers: Good catch — and it was broader than dynamodb. \`collect_data_connectors\` only iterated \`DATA_CONNECTOR_REGISTRATIONS\` (the linkme slice), which is populated solely by \`register_data_connector!\`. Every already-extracted connector that registers via the explicit \`register_connector_factory\` path (clickhouse, dremio, mysql, snowflake, mongodb, graphql, delta_lake, oracle, mssql, scylladb, sftp, smb, sharepoint, imap, ftp, flightsql, spark) had already silently dropped out of the generated schema, and dynamodb would have been the next.

Fixed in 9113c6e97: `collect_data_connectors` now also enumerates each extracted connector's public `factory()` (name + prefix + parameters) alongside the slice, deduped by name. Regenerated `.schema/spicepod.schema.json` — all 18 are restored (verified dynamodb's `dynamodb_*` params are back). The `use connector_* as _;` linkage stubs are replaced by the real `factory()` references.

.schema/spicepod.schema.json
```diff
@@ -12599,9 +12547,6 @@
         "check_availability": {
           "$ref": "#/$defs/CheckAvailability"
         },
-        "schema_inference": {
```
bjchambers, line 12602: Why did the scehma inference entries disappear?

bin/spiced/src/lib.rs
```diff
@@ -127,6 +131,11 @@ pub async fn register_external_connectors() {
     )
     .await;
 
+    // DynamoDB self-registers into the linkme slice (`register_data_connector!`); `register_all()`
```
bjchambers: Drop this comment -- don't need to record this. We should make it clear at the top that every crate registered via linkme needs the `use ... as _;` block.

.schema/spicepod.schema.json
bjchambers: Reverted `.schema/spicepod.schema.json` to trunk (8c30c463d) — it's now byte-identical to trunk. Those entries (`schema_inference` / `SchemaInference` + a couple `cayenne_*` fields) weren't actually dropped by the change: moving DynamoDB to `connector-dynamodb` registers the *same* params from a different crate, so the schema is schema-neutral. The disappearance came from my local `cargo run` regen resolving a narrower feature set than CI's `generate_json_schema` workflow, which omitted those feature-gated fields — committing that regen was the mistake. Since the change is schema-neutral the file should just match trunk (CI regenerates canonically).

bin/spiced/src/lib.rs
bjchambers: Done in 8c30c463d — dropped the inline comment, and the force-linkage block at the top of the module now states the rule directly: every connector that self-registers via `register_data_connector!` needs a `use <crate> as _;` line there (a plain Cargo dependency won't link the slice static — the linker drops it), and `register_all()` registers them at startup.

tools/spicepodschema/src/collector.rs
```diff
@@ -128,6 +112,83 @@ pub fn collect_data_connectors() -> Vec<ConnectorSchema> {
...
+        extracted
+            .into_iter()
+            .map(|(name, factory)| ConnectorSchema {
+                name: name.to_string(),
+                prefix: factory.prefix(),
+                parameters: factory.parameters(),
+            }),
+    );
+
     connectors.sort_by(|a, b| a.name.cmp(&b.name).then_with(|| a.prefix.cmp(b.prefix)));
```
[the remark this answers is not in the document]
bjchambers: This was in the enumeration approach, which has been reverted (316a64a9a) — `collect_data_connectors` is back to the trunk slice-only version. That `a.prefix.cmp(b.prefix)` is trunk's existing code and compiles: `prefix` is `&'static str` (a `Copy` type), so nothing is moved out of the borrow.

tools/spicepodschema/src/collector.rs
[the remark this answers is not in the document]
bjchambers: Update: after confirming cross-crate `linkme` works on this toolchain (mini-repro on linkme 0.3.36 + rustc 1.96.1 — the slice entry links via `use ... as _;` in dev, release, and fat-LTO), we pivoted away from enumeration. `connector-dynamodb` now self-registers via `register_data_connector!` into `DATA_CONNECTOR_REGISTRATIONS` (as it did while resident in `runtime`), so `collect_data_connectors` — unchanged, slice-only — documents it, and `bin/spiced` just force-links it (`use connector_dynamodb as _;`). Single registration path, no double-registration. The other already-extracted connectors adopt the same pattern in a follow-up PR.

lukekim (approved):

PR #11966, opened 2026-07-21
fix(testoperator): HTAP drain gate false-fails on stale probe observations — final snapshot decides convergence (fixes #11953)

### LINKED ISSUE #11953: testoperator HTAP convergence gate false-fails on idle stale-watermark lag_ms despite its own caught-up detection
Observed in run https://github.com/spiceai/spiceai/actions/runs/29775050532 (SF1000):

```
verdict: FAILED — 0 table(s) mismatched, replication did not converge
...post-drain re-scrape...
caught-up (client lag_bytes=0 AND authoritative slot retained ~0 ⇒ all WAL consumed; the nonzero lag_ms is an idle stale-watermark, not real staleness): customer, district, new_order, oorder, order_line, stock, warehouse
```

Every table shows the identical `lag_ms = 1008913` (= the drain window — the idle stale-watermark ramp after OLTP stops), `lag_bytes = 0`, slot retained ~0, and the re-scrape's own annotation declares all tables caught-up — yet the gate verdict is "replication did not converge" because the drain-wait keys on `lag_ms`. The content gate passed (0 mismatches).

Fix suggestion: the drain-wait loop should use the same caught-up predicate the post-drain re-scrape prints (`lag_bytes==0 AND slot retained ~0`), falling back to `lag_ms` only while bytes are outstanding. Related: the 900s+ drain at SF1000 may also deserve a scale-aware window.

claudespice:
Fixes #11953

## Summary (root cause)

The HTAP data-correctness gate's drain-wait loop probes each table serially, and at large scale factors each probe is a full-table `COUNT(*)`/`MAX(_bench_ts)` scan on both the source and Spice that can take minutes. In the failing SF1000 run (actions/runs/29775050532) a real CDC backlog persisted ~11 minutes into the 15-minute drain window; by the time the deadline expired mid-pass (overshooting `max_wait` by ~100s), the loop's last per-table observations were minutes stale — so the verdict said "replication did not converge" while the final snapshot taken immediately afterward showed every table matching (0 count mismatches, 0.0000% content-fingerprint delta on all 7 tables).

Note: the issue hypothesized the drain-wait keys on the idle stale-watermark `lag_ms`; it actually keys on `MAX(_bench_ts)` + row-count equality. The `lag_ms = 1008913` in the log is from the post-drain diagnostic re-scrape. The false-fail mechanism is stale serial-probe observations deciding the verdict while fresher evidence (the final snapshot) contradicts them. The fix therefore keeps the end-to-end applied-and-visible predicate rather than switching to WAL-level signals (`lag_bytes`/slot-retained), which prove consumption but not apply-visibility.

## Changes

- `verify_after_drain` latches tables observed caught-up and stops re-probing them: the source is static once OLTP stops and CDC applies in commit order, so a caught-up table cannot fall behind — and skipping it keeps later passes cheap at SF1000.
- The deadline is re-checked before every table probe instead of once per pass, bounding the overshoot of `max_wait` (observed: 1002.8s vs a 900s bound).
- New `Convergence` verdict (`Within` / `ObservedAtFinalSnapshot` / `No`): when the wait times out, the final snapshot — strictly fresher than any in-loop probe — decides. Never-latched tables are re-probed in full (counts + `MAX(_bench_ts)`); every table caught up means replication converged late, not "not at all". A probe error or a genuinely-behind table leaves the failure standing (fail closed).
- Shared `probe_table` helper and a single `table_caught_up` predicate used by both verdict sites, so they cannot drift apart.

Complementary to #11956 (which widens the drain window to 2× duration but keeps the stale-observation verdict).

## Performance impact

Benchmark-harness only (no runtime code). Drain passes get cheaper (caught-up tables are no longer re-scanned each pass); the timeout path adds `MAX(_bench_ts)` probes only for tables never observed caught-up, joined concurrently with the count scans it already ran.

## Test plan

- [x] `cargo test -p testoperator` — 5 new unit tests: latch predicate edge cases (both-empty, one-empty, ts-only lag, count-only lag), convergence resolution (in-loop wins; timed-out + matching final snapshot ⇒ converged late — regression for #11953; timed-out + diverging snapshot ⇒ fail), late convergence passes `failure_message()`, non-convergence and count mismatches still fail.
- [x] `make lint` (via remote sign-off run 29819076675)
- [x] `make test` / `make nextest` (via remote sign-off run 29819076675)

## Checklist

- [x] Fix targets the root cause, fail-closed on probe errors
- [x] Regression test that fails on the old code (`resolve_convergence(None, true, …)` previously unrepresentable — old verdict keyed only on the in-loop timeout)
- [x] `make signoff` green (Attestation passing on HEAD)

## Also in this PR: unblock scoped lint for bin-only packages

The remote sign-off's targeted pre-lint (`make lint-rust PACKAGES="testoperator"`) failed before linting anything: `cargo clippy --lib --bins -p testoperator` is a fatal `no library targets found` because `testoperator` is bin-only — so **any** testoperator-only branch could not sign off. Scoped lint runs now use cargo's default target selection (the package's lib and/or bins, the same set `--lib --bins` names); workspace-wide runs keep the explicit `--lib --bins`. Verified locally: `make lint-rust PACKAGES="testoperator" FEATURES=` reproduces the failure on trunk's Makefile and passes with this change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,9 +131,15 @@ ifneq ($(strip $(PACKAGES)),)
 _LINT_PKG_FLAGS := $(foreach p,$(PACKAGES),-p $(p))
 _LINT_WORKSPACE_FLAGS := $(_LINT_PKG_FLAGS)
 _FMT_FLAGS := $(_LINT_PKG_FLAGS)
+# Scoped runs rely on cargo's default target selection (the package's lib
+# and/or bins — the same set --lib --bins names): an explicit --lib is a fatal
+# `no library targets found` on bin-only packages (e.g. testoperator), which
+# breaks the targeted pre-lint that scripts/signoff derives for such branches.
+_LINT_TARGET_FLAGS :=
 else
 _LINT_WORKSPACE_FLAGS := --workspace --exclude libnfs --exclude lopdf --exclude ttf-parser --exclude pdf-extract
 _FMT_FLAGS := --all
+_LINT_TARGET_FLAGS := --lib --bins
 endif
 # Apply FEATURES if provided, otherwise default to hardcoded features only for workspace-wide linting
 ifneq ($(strip $(FEATURES)),)
@@ -153,7 +159,7 @@ lint-rust:
 	## Crate-layering guard (fast, no compile): no crate may depend on a higher tier. See docs/dev/crate_layering.md
 	python3 scripts/check_crate_layers.py
 	## All except metal, cuda, nfs (nfs requires system libnfs library)
-	CLIPPY_CONF_DIR=".ci" cargo clippy $(CARGO_PROFILE) --keep-going --lib --bins $(_FEATURES_FLAGS) $(_LINT_WORKSPACE_FLAGS) -- \
+	CLIPPY_CONF_DIR=".ci" cargo clippy $(CARGO_PROFILE) --keep-going $(_LINT_TARGET_FLAGS) $(_FEATURES_FLAGS) $(_LINT_WORKSPACE_FLAGS) -- \
 		-Dwarnings \
 		-Dclippy::pedantic \
 		-Dclippy::unwrap_used \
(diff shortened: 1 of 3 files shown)


claudespice:
@copilot review

bjchambers (approved):

PR #11979, opened 2026-07-22
fix(cayenne): support Decimal128 in maintained SUM/AVG aggregates (fixes #11933)

### LINKED ISSUE #11933: Cayenne IVM: maintained SUM/AVG reject Decimal128 — chbench adaptive-ivm configs fail to start after #11921 (money cols → DECIMAL)
## Summary

The HTAP CH-benCHmark **`cayenne[file]-adaptive-ivm`** configs (both Postgres and MySQL sources, SF1) have been failing to start on `trunk` since **2026-07-17**. `spiced` never reaches ready because the `order_line` dataset goes unhealthy at load:

```
Failed to setup the dataset order_line (postgres). Unable to create dataset acceleration:
Acceleration creation failed: Invalid operation: Failed to configure maintained aggregates
for table 'order_line'. Error during planning:
Sum maintained aggregate does not support column type Decimal128(6, 2)
```

Result: `11/12 ready, 1 unhealthy` → `Error: Spiced instance not ready within 60s` → the benchmark run fails and emits no metrics. This is confirmed over 3 consecutive scheduled runs (Postgres) and reproduces identically on the MySQL sibling.

## Root cause

Cayenne's maintained-aggregate registry supports `SUM`/`AVG` only over the **integer and float** families — `MIN`/`MAX` were extended to `Decimal128`, but `SUM`/`AVG` were not, so a `Decimal128` column falls to the catch-all reject:

`crates/cayenne/src/maintained_aggregate.rs:1217-1221`
```rust
(function, Some(data_type)) => {
    return Err(DataFusionError::Plan(format!(
        "{function:?} maintained aggregate does not support column type {data_type}"
    )));
}
```

This is a **long-standing limitation** (the file was last touched by #11862; neither PR in the regression range modified it). It was latent until the benchmark schema began exercising it.

## Trigger

#11921 (`test(chbench): align money columns to canonical DECIMAL schema (CMU BenchBase)`) changed the chbench money columns — including `order_line.ol_amount` — from `DOUBLE PRECISION` to `NUMERIC(6,2)` (arrow `Decimal128(6, 2)`). The `adaptive-ivm` spicepods maintain `SUM(ol_amount)` as an IVM aggregate, so after the schema change that aggregate is now planned over a `Decimal128` column and rejected. Before #11921 the column was a float and `SUM` widened to `Float64` fine.

#11921's schema change is correct (it matches the CMU BenchBase reference); the real gap is that Cayenne cannot maintain `SUM`/`AVG` over `Decimal128` — a very common money-column type for exactly this kind of aggregate.

## Suggested fix

Add `Decimal128` support to maintained `SUM`/`AVG` in `maintained_aggregate.rs` (the backing value is an `i128`, so an exact `i128`/`i256` running sum is feasible, mirroring how `MIN`/`MAX` already handle `Decimal128`). Until then, these configs cannot use the IVM path over decimal money columns.

## Evidence

- Last good: `2e6bad9deb` (Postgres, 2026-07-16, qph ≈ 97.7k) → first bad: `1bead42933` (2026-07-17). Range `2e6bad9deb..1bead42933` contains only #11921, #11916, #11913; only #11921 is causal (confirmed by the `Decimal128(6,2)` error matching #11921's money→DECIMAL change).
- Postgres `adaptive-ivm` SF1 failing runs: [29561852857](https://github.com/spiceai/spiceai/actions/runs/29561852857), [29604196749](https://github.com/spiceai/spiceai/actions/runs/29604196749), [29634472576](https://github.com/spiceai/spiceai/actions/runs/29634472576) (tip `fc61c877`).
- MySQL `adaptive-ivm` SF1 failing runs (identical error): [29562142309](https://github.com/spiceai/spiceai/actions/runs/29562142309), [29604538071](https://github.com/spiceai/spiceai/actions/runs/29604538071).

Non-IVM adaptive configs are unaffected (they maintain no aggregates).

---
*Filed by the automated benchmark performance-regression sweep (HTAP CH-benCHmark, SF1). Attribution verified by code inspection + candidate-range narrowing; not bisected (deterministic single-cause).*

claudespice:
Fixes #11933

## Summary (root cause)

Cayenne maintained aggregate views (`maintained_aggregates` IVM) supported `SUM`/`AVG` only over the integer and float families; a `Decimal128` column was rejected at dataset registration with `Sum maintained aggregate does not support column type Decimal128(6, 2)`, taking the whole dataset unhealthy. This hard-fails the CH-benCH `cayenne[file]-adaptive-ivm` configs (both postgres and mysql source) at startup since #11921 aligned chbench money columns to canonical `DECIMAL` — the spicepods maintain `SUM(ol_amount)`, now planned over `Decimal128(6, 2)`. Long-standing gap, not a regression introduced by #11921 (the schema change is correct).

## Changes

- `ResolvedAggregateExpr::try_new` accepts `Decimal128(p, s)` for `SUM` (any scale; output `Decimal128(min(38, p+10), s)`) and for `AVG` with `s >= 0` (output `Decimal128(min(38, p+4), min(38, s+4))`) — DataFusion's exact decimal `SUM`/`AVG` return types, so the maintained view's output schema matches the planned aggregate and substitution serves.
- New `AggregateOutputType::Decimal128 { precision, scale }` output type and `SumDecimal128`/`AvgDecimal128` accumulators: both maintain the exact `i128` backing-value sum at the input scale (at one fixed scale, decimal addition IS integer addition on backing values), so insert/retract are exactly invertible like the integer accumulators — no float drift.
- The `AVG` quotient is computed only at serve time by **DataFusion's own `DecimalAverager`** (`datafusion-functions-aggregate-common`, already pinned in the workspace), so rescale, truncation-toward-zero, output-precision validation, and overflow errors are identical to a base-table re-scan **by construction**, not by a hand-kept reimplementation.
- Overflow discipline unchanged: apply-path `checked_add`/`checked_sub` fail the registry safe to stale (query re-scans the base table); the serve-time `AVG` overflow errors with DataFusion's own `Arithmetic Overflow in AvgAccumulator`, exactly what the re-scan raises for the same data.
- `SumDecimal128` encodes SQL `NULL` as `non_null_count == 0` instead of `Option<i128>` (no niche; the `Option` would grow every `AggregateAccumulator` from 48 to 64 bytes for all views, decimal or not).
- Negative-scale decimal `AVG` and `Decimal256` remain unsupported and fall to the existing safe-decline (view does not build; queries re-scan — correct, not fast).
- `docs/cayenne/cayenne.md`: maintained-aggregates section updated + document changelog row added.

## Test plan

- [x] `make lint` (clippy pedantic incl. `--tests` pass) — green locally, scoped
- [x] `make test` / `make nextest` — runs via `make signoff` after push
- New unit tests in `crates/cayenne/src/maintained_aggregate.rs`:
  - `maintains_sum_avg_over_decimal128` — regression guard (previously failed registration) + serves through a real DataFusion `AggregateExec`, asserting DataFusion's own computed output types `Decimal128(16, 2)` / `Decimal128(10, 6)` and exact values incl. NULL handling
  - `avg_over_decimal128_truncates_toward_zero_like_datafusion` — quotient truncates toward zero for positive and negative sums (not floor)
  - `retracts_decimal128_sum_avg_by_pk` — in-place UPDATE, DELETE, retracting the last non-null contribution restores SQL `NULL` while a null-valued row keeps the group alive, group drop on last row
  - `sum_over_decimal128_overflow_fails_safe_to_stale` — i128 overflow fails the apply pass instead of wrapping
  - `avg_over_decimal128_result_overflow_errors_like_datafusion` — serve-time rescale overflow raises DataFusion's `Arithmetic Overflow in AvgAccumulator`
  - `avg_over_negative_scale_decimal128_is_rejected` — negative-scale AVG declines; SUM accepts
- New optimizer test in `crates/cayenne/src/optimizer_rules.rs`: `maintained_aggregate_rewriter_replaces_sum_avg_over_decimal128` — full plan-substitution path over a decimal column

## Performance impact

No change to existing paths: the new accumulator variants are additional match arms; the accumulator enum size is unchanged (48 bytes, verified — the `Option<i128>` that would have grown it to 64 was deliberately avoided). Decimal maintenance itself is O(1) integer adds per row, and the AVG division is deferred to serve time (O(groups), not O(rows)).

## Checklist

- [x] Issue linked (`Fixes #11933`)
- [x] Tests added (regression + edge cases: NULLs, retraction, overflow, truncation, negative scale)
- [x] `docs/cayenne/cayenne.md` reviewed and updated with changelog row
- [x] `make signoff` attested after push (remote sign-off, lint+test passed in 7157s)


claudespice:
@copilot review

lukekim (approved):

PR #11777, opened 2026-07-09
Unified CDC config

krinart:
## 📝 Summary

Unifies Change Data Capture (`refresh_mode: changes`) **configuration** and **readiness** across all four CDC connectors — PostgreSQL, MySQL, MongoDB, and DynamoDB.

Previously each connector had its own config vocabulary and an ad-hoc notion of "ready" — some marked a dataset ready immediately on resume or right after the initial snapshot — which was inconsistent between connectors and could either serve not-yet-caught-up data or hang forever waiting for a signal on a quiet source. This PR makes the vocabulary and the readiness model identical everywhere, with readiness driven purely by replication lag.

- **Unified config vocabulary.** Shared `initial_snapshot` (`auto|enabled|disabled`) and `invalid_checkpoint_behavior` (`error|restart`) enums, exposed uniformly as `{mysql,mongodb,dynamodb}_replication_*` params (Postgres omits `invalid_checkpoint_behavior` — its replication slot is durable). Old per-connector parameter names remain as deprecated aliases.

- **Pure lag-based readiness.** A new `{connector}_replication_ready_lag` (default `2s`): a dataset is marked `Ready` only once its replication lag — wall-clock now minus the freshest applied source-commit timestamp — drops below the threshold. A dataset that is snapshotting, draining a resume backlog, or rebuilding after a restart therefore stays not-ready and never serves stale or incomplete data. This replaces the old immediate/one-shot ready signals and the per-connector `restart_readiness` knob. (DynamoDB's existing `ready_lag` becomes `dynamodb_replication_ready_lag`, with a deprecated alias.)

- **Source-attested idle heartbeats.** So a quiet-but-caught-up source still reaches `Ready` (the necessary flip side of pure lag-based readiness). Each connector periodically emits a zero-row heartbeat stamped with the *source's own* clock — Postgres from the server keepalive, MySQL from a side-connection `NOW(3)`, MongoDB from the change stream's gossiped `operationTime`. A heartbeat is emitted only when the stream has provably caught up to the source head, and never carries a fabricated local `now()`, so a backlogged dataset is never marked ready.

- **Shared plumbing + observability.** A single readiness predicate and heartbeat-envelope builder in `data_components::cdc` keep all connectors behaving identically, plus uniform `spice_cdc::commit` / `spice_cdc::heartbeat` log lines (source-commit timestamp + computed lag) for verifying freshness and readiness.

## 🔗 Related

[...]

PR #12030, opened 2026-07-24
fix(cdc): don't force the durable path on zero-row readiness heartbeats (fixes #12007)

### LINKED ISSUE #12007: Regression: Cayenne-accelerated CDC tables gain duplicate rows after #11777 — HTAP data-correctness gate failing on trunk (postgres-cayenne SF1)
## Summary

Since #11777 (Unified CDC config) landed on trunk (`74bfe57df9`), the scheduled CH-benCHmark HTAP SF1 runs on **both** Postgres→Cayenne configs fail the data-correctness gate: the Cayenne-accelerated table ends up with **more rows than the source** (duplicate rows), and replication never converges within the 30-minute drain window. DuckDB-accelerated control configs on the same commits pass, so this is specific to the Postgres CDC → Cayenne path.

Because the extra rows appear on tables that TPC-C never inserts into (`customer`, `stock`), this is row **duplication** in the accelerated table — a data-correctness bug, not replication lag.

## Evidence (scheduled `testoperator_run_htap.yml` runs, SF1)

| run | spiced commit | config | verdict |
|---|---|---|---|
| [29898895414](https://github.com/spiceai/spiceai/actions/runs/29898895414) (Jul 22 07:04) | `269b92714c` (pre-#11777) | `postgres-cayenne[file]-adaptive` | ✅ PASSED — replication converged in **711 ms**, all 7 tables match, 22/22 queries match |
| [29947197238](https://github.com/spiceai/spiceai/actions/runs/29947197238) (Jul 22 18:33) | `74bfe57df9` (**#11777**) | `postgres-cayenne[file]-adaptive` | ❌ FAILED — `oorder` source_count=389647, spice_count=410647 (**+21,000 rows**), did not converge in 1800s |
| [29986755766](https://github.com/spiceai/spiceai/actions/runs/29986755766) (Jul 23 06:58) | `a70154466f` | `postgres-cayenne[file]-adaptive` | ❌ FAILED — `oorder` source_count=373613, spice_count=394613 (**+21,000 rows**), did not converge in 1800s |
| [29986453350](https://github.com/spiceai/spiceai/actions/runs/29986453350) (Jul 23 06:52) | `a70154466f` | `postgres-cayenne[file]` (non-adaptive) | ❌ FAILED — `customer` 30000→**30005**, `stock` 100000→**100020**, did not converge in 1800s |

(The Jul 22 18:28 run of the non-adaptive config, [29946790891](https://github.com/spiceai/spiceai/actions/runs/29946790891), failed the same way — 2 consecutive failures on each of the 2 configs, 4 total; its last pass was [29899142680](https://github.com/spiceai/spiceai/actions/runs/29899142680) at `269b92714c`.)

Notable details:

- **Extra rows on update-only tables.** `customer` and `stock` never receive inserts in CH-benCHmark; `spice_count > source_count` on them can only be duplicated rows in the accelerated table. The drain probe even reports `oorder rows src=spice [ok] | max_bench_ts lag 0ms [ok]` at drain start in the non-adaptive run — the counts that never converge are `customer`/`stock`.
- **`oorder` is ahead by exactly 21,000 rows in both adaptive failures** — a deterministic quantum, suggesting a replayed window of changes (double-apply after a checkpoint/snapshot boundary) rather than random duplication.
- **QPH inflates while the gate fails** (non-adaptive: ~73–79k → 121–129k), consistent with the replication/apply path doing less (or incorrectly short-circuited) work.

## Attribution

- Last good spiced commit: `269b92714c` (run passed cleanly, 711 ms drain).
- First bad: `74bfe57df9`. The ancestry-path range contains exactly two PRs:
  - #11777 — Unified CDC config: rewrites Postgres/MySQL replication config, replaces the sentinel ready-signal with **lag-gated readiness + source-attested idle heartbeats**, persists the initial resume token on a **zero-row not-ready boundary envelope**, and changes these exact chbench pods (`pg_replication_initial_snapshot: 'true'` → `auto`).
  - #11976 — Spidapter Flight auth passthrough (does not touch CDC/replication).
- Control configs at the same commits: `postgres-duckdb[file]` SF1 and `mysql-duckdb[file]` SF1 **pass** on `74bfe57df9` and `a70154466f` → the regression is specific to the Cayenne apply path interacting with the reworked CDC stream.
- The adaptive pod also carries the #11944 CDC-coalescing pins, but the first post-#11944 run (`269b92714c`) passed with a 711 ms drain, so #11944 alone is not the trigger. An interaction between #11777's boundary/heartbeat envelopes and the Cayenne changes-apply/dedup path is plausible for the adaptive config; the non-adaptive config fails without any coalescing pins.

## Suspected mechanism (unverified)

The #11777 readiness rework introduces zero-row boundary/heartbeat envelopes and changes checkpoint/resume + initial-snapshot semantics. A snapshot/stream overlap or a replay of already-applied changes after a checkpoint boundary would produce exactly what the gate sees: duplicated rows that no amount of waiting resolves.

## Impact

Cayenne-accelerated CDC datasets can serve **duplicate rows** (wrong query results — the analytical gate is skipped once row counts mismatch). This affects trunk since Jul 22.

/cc #11777

claudespice:
Fixes #12007

## Summary

Since #11777 (Unified CDC config), the two scheduled HTAP SF1 `postgres-cayenne` configs fail the data-correctness gate with **duplicate rows** in Cayenne-accelerated CDC tables (`oorder` +21,000; `customer`/`stock` +5/+20) that never converge.

**Root cause:** #11777's lag-based readiness makes CDC connectors emit a zero-row heartbeat envelope roughly every second on a caught-up source. The heartbeat's committer is a `NoOpCommitter`, which reports `supports_deferral() == false`, so in the CDC apply loop every heartbeat-bearing burst flipped `requires_durable_cdc_path` — forcing a Cayenne mem-tier checkpoint (and, for heartbeat-only bursts, a second post-write checkpoint via the epoch-less commit path) once per heartbeat. Under TPC-C load those once-a-second forced checkpoints raced the pipelined Stage-B staged-append finalize, and Cayenne's staged-WAL crash recovery "recovered" (double-published or rolled back) in-flight appends — duplicating rows. Log evidence: the failing adaptive run ([29986755766](https://github.com/spiceai/spiceai/actions/runs/29986755766)) has **1,234** `Incomplete staged append detected — attempting automated recovery` events and **527** `Rolling back an uncommitted deferred snapshot append` events; the last passing run before #11777 ([29898895414](https://github.com/spiceai/spiceai/actions/runs/29898895414)) has **0**.

## Changes

- `data_components/src/cdc.rs`: add `CommitChange::is_no_op()` (default `false`; only `NoOpCommitter` overrides to `true`) and `ChangeEnvelope::is_no_op_heartbeat()` — a pure readiness heartbeat is a **zero-row** batch with a **no-op** committer. Zero-row envelopes carrying a real committer (e.g. the MySQL snapshot-boundary envelope that persists the initial resume token) are deliberately excluded and keep durability-then-commit ordering.
- `runtime/src/accelerated_table/refresh_task/changes.rs`: `apply_envelope_run` strips no-op heartbeats from the burst before write/durability classification. Their only observable effect — the `is_dataset_ready` flag — is honored: a heartbeat-only run marks the dataset Ready (deferred behind a still-pending Stage-B finalize, mirroring the write path's gate) and touches neither the write path nor the checkpoint machinery. Extracted the previously-triplicated mark-ready side effect into one `signal_dataset_ready` helper used by all three ready paths.
- `data-connectors/connector-dynamodb`: drop the connector's duplicate local `NoOpCommitter` in favor of the shared `cdc::NoOpCommitter`, so the "effect-free committer" contract has a single definition.

## Test plan

- [x] New integration regression test `runtime/tests/cdc_cayenne_heartbeat.rs`: memory-durability Cayenne table; a deferred source committer must **not** be acked when a readiness heartbeat arrives (it was, before the fix — the test fails on pre-fix code at exactly that assertion) while the ready flag still flips.
- [x] New unit tests: heartbeat-only stream signals Ready without ever reaching the accelerator write path; heartbeats interleaved with data preserve apply/commit order; `is_no_op_heartbeat` predicate accepts connector heartbeats/ready signals and rejects zero-row envelopes with real committers and row-bearing envelopes.
- [x] `make lint`
- [x] `make test` / `make nextest` (via `make signoff`)

## Performance impact

Heartbeat envelopes no longer trigger a durable mem-tier checkpoint per beat, restoring the pre-#11777 checkpoint cadence for `cdc_durability: memory` tables. Per-burst cost of the strip is one virtual call per envelope (short-circuits before any batch access for real committers).

## Checklist

- [x] Fix targets the root cause, gated to exactly the heartbeat envelopes (`zero rows && no-op committer`)
- [x] Regression test fails on pre-fix code, passes on the fix
- [x] No behavior change for zero-row envelopes with real committers (MySQL boundary/resume-token path)
- [x] Lint green; signoff to follow after push

### DIFF
--- a/crates/data-connectors/connector-dynamodb/src/connector.rs
+++ b/crates/data-connectors/connector-dynamodb/src/connector.rs
@@ -20,7 +20,7 @@ use crate::stream::StreamError as DynamoDBStreamError;
 use async_trait::async_trait;
 use data_components::cdc::{
     ChangeEnvelope, ChangesStream, CommitChange, CommitError, InitialSnapshotMode,
-    InvalidCheckpointBehavior,
+    InvalidCheckpointBehavior, NoOpCommitter,
 };
 use datafusion::datasource::TableProvider;
 use datafusion::sql::TableReference;
@@ -1106,14 +1106,6 @@ impl MetricsProvider for DynamoDBMetricsProvider {
             }
             _ => None,
         }
-    }
-}
-
-struct NoOpCommitter;
-#[async_trait]
-impl CommitChange for NoOpCommitter {
-    async fn commit(&self) -> Result<(), CommitError> {
-        Ok(())
     }
 }
 
--- a/crates/data_components/src/cdc.rs
+++ b/crates/data_components/src/cdc.rs
@@ -211,6 +211,19 @@ pub trait CommitChange {
     fn as_any(&self) -> Option<&dyn std::any::Any> {
         None
     }
+
+    /// Whether [`Self::commit`] is statically known to do nothing — there is no
+    /// source position, token, or checkpoint to acknowledge. Consumers may run
+    /// or drop such a committer at any point without ordering or durability
+    /// constraints; the runtime relies on this to keep zero-row readiness
+    /// heartbeats out of the CDC write/durability path (see
+    /// [`ChangeEnvelope::is_no_op_heartbeat`]). Defaults to `false`
+    /// (conservative: assume the commit has effects, preserving ordering); only
+    /// genuinely effect-free committers such as [`NoOpCommitter`] should
+    /// override this to `true`.
+    fn is_no_op(&self) -> bool {
+        false
+    }
 }
 
 /// Destination-passing-style source of the change rows carried by a
@@ -502,6 +515,23 @@ impl ChangeEnvelope {
         self.change_batch.is_heartbeat()
     }
 
+    /// Whether this envelope is a pure readiness heartbeat: a zero-row change
+    /// batch whose committer is a no-op ([`CommitChange::is_no_op`]).
+    ///
+    /// CDC connectors emit these ([`build_heartbeat_envelope`],
+    /// [`build_ready_signal_envelope`]) only to carry `is_dataset_ready` and a
+    /// source freshness timestamp; they hold no data and acknowledge no source
+    /// position, so a consumer may honor the ready flag and drop the envelope
+    /// without writing, committing, or forcing any durability transition. A
+    /// zero-row envelope whose committer has real effects (e.g. a `MySQL`
+    /// snapshot-boundary envelope that persists the initial resume token) is
+    /// NOT a heartbeat under this predicate and must keep normal
+    /// durability-then-commit ordering.
+    #[must_use]
+    pub fn is_no_op_heartbeat(&self) -> bool {
+        self.change_committer.is_no_op() && self.is_heartbeat()
+    }
+
     pub async fn commit(self) -> Result<(), CommitError> {
         self.change_committer.commit().await
     }
@@ -582,6 +612,10 @@ pub struct NoOpCommitter;
 impl CommitChange for NoOpCommitter {
     async fn commit(&self) -> Result<(), CommitError> {
         Ok(())
+    }
+
+    fn is_no_op(&self) -> bool {
+        true
     }
 }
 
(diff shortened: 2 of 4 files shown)


claudespice:
@copilot review

lukekim (approved):

sgrebnov (approved):

PR #12043, opened 2026-07-25
fix(search): check dataset readiness before embedding the query (fixes #10956)

### LINKED ISSUE #10956: Bug. `search` should verify that dataset is ready/loaded
## Describe the bug

Follow cookbook below
https://github.com/spiceai/cookbook/tree/trunk/search_github_files

Run search while data is still loading

```
2026-05-20T10:43:30.178252Z  INFO runtime::accelerated_table::refresh_task: Loading data for dataset spiceai.files

... Run search query here


2026-05-20T10:48:06.490334Z  INFO runtime::accelerated_table::refresh_task: Loaded 183 rows (8.09 MiB) for dataset spiceai.files in 4m 36s 312ms.
2026-05-20T10:48:06.575807Z  INFO runtime: All components are loaded. Spice runtime is ready!

```

Observer query hangs and complete only after dataset is fully loaded, example duration: `duration_ms:53_807`

```
curl -XPOST http://localhost:8090/v1/search \
    -d "{
    \"datasets\": [\"spiceai.files\"],
    \"text\": \"threat\",
    \"where\": \"not contains(path, 'docs/release_notes')\",
    \"additional_columns\": [\"download_url\"],
    \"limit\": 2
    }"
{"results":[{"matches":{"content":["\n| Component           | Description                                                                                                                                                                                  | Definition Link                                            |"]},"data":{"download_url":"https://raw.githubusercontent.com/spiceai/spiceai/trunk/docs/EXTENSIBILITY.md"},"primary_key":{"path":"docs/EXTENSIBILITY.md"},"_score":0.9367570133287714,"dataset":"spiceai.files"},{"matches":{"content":["              | language | premiered  | status | details                                                                                                                                                                                                                                                                                                                                                                                                                                           |\n| --- | -------"]},"data":{"download_url":"https://raw.githubusercontent.com/spiceai/spiceai/trunk/docs/features/schema-decomposition.md"},"primary_key":{"path":"docs/features/schema-decomposition.md"},"_score":0.9332738498352093,"dataset":"spiceai.files"}],"duration_ms":53807}% 
```

claudespice:
## Summary (root cause)

Readiness **was** enforced for accelerated datasets, but only at physical-planning time, inside `AcceleratedTable::scan()`: with the default `ready_state: on_load` and an incomplete initial load it returns `Acceleration not ready; loading initial data for <dataset>`.

Search reaches that check late, and unevenly:

- **The query text is embedded before the check runs.** The vector candidate generator puts `embed(<query>, <model>)` — both arguments literals — into the logical plan, and the `embed` UDF is registered `Volatility::Stable`, which DataFusion const-evaluates during `SimplifyExpressions`. Logical optimization precedes physical planning, and `embed_sync` blocks on the embedding call, so a search against a still-loading dataset pays for a full embedding round trip and only then fails.
- **Only some search paths reach the check at all.** The non-indexed (JIT) path scans through the wrapped provider, so it hits the gate. The index-backed paths (`s3_vectors`, `duckdb`, `elasticsearch`, native vector) and full-text search query the index directly, so a partially-populated index answers from whatever subset has loaded so far — silently returning incomplete results rather than reporting the dataset isn't ready.

So the readiness signal existed but was paid for late on one path and skipped entirely on the others.

Fixes #10956

## Changes

- **`AcceleratedTable`: one source of truth for the readiness decision.** Extracted the condition `scan()` already applied into `scan_readiness()` (backed by the pure `scan_readiness_for`), returning `Accelerated` / `Scheduler` / `NotReady` / `ReadyStateFallback`, plus `not_ready_error()` for the `NotReady` case. `scan()` now consumes the same decision, so a pre-flight check and the scan can't drift. No behavior change: the scheduler bypass, the caching-mode exemption, the `on_load` error and the `on_registration` / `on_schema_resolved` federated fallback (including its `READY_STATE_FALLBACK` metric) are all preserved exactly.
- **`TableProviderExplorer::not_ready_error`** — new trait method, no default impl so every implementor must answer it. `runtime-search` sits below `runtime` in the crate layering and cannot name `AcceleratedTable`, so the peeling lives in `RuntimeTableProviderExplorer`, which unwraps to the `AcceleratedTable` with the same accessor set as `find_concrete` and forwards to `not_ready_error()`. A provider with no accelerator behind it reports `None`.
- **`SearchEngine::search` checks readiness first**, before building any generator or embedding anything. An **explicitly requested** dataset that isn't ready returns the same `Acceleration not ready` error the SQL path gives — now immediately, and now on the index and full-text paths too. When no datasets were named and search is sweeping every searchable dataset, a not-ready one is excluded instead of failing the whole request, mirroring the adjacent `explicit_datasets_requested` handling for unsearchable datasets.

The error text is unchanged and reused rather than reworded, so search and SQL report a loading dataset identically.

## Behavior notes

`ready_state: on_load` is the default, and it already means "this dataset cannot serve queries until the initial load completes" — this makes search agree with SQL instead of embedding first (JIT path) or answering from a half-built index (index and full-text paths). Datasets configured `ready_state: on_registration` or `on_schema_resolved` are explicitly *not* affected: they stay scannable and continue serving from the federated source while the accelerator loads.

## Test plan

- `make lint`
- `make test`
- New: `test_scan_readiness_covers_every_state_combination` — every combination of `cluster_role` × `initial_load_completed` × all six `RefreshMode` variants × all three `ReadyState` variants, asserted against what `scan()` did before the extraction.
- New: `test_only_on_load_with_outstanding_load_refuses_a_scan` — regression guard that only `on_load` + outstanding load refuses a scan, so the pre-flight check can never reject a dataset a scan would have served.
- New: `test_not_ready_error_is_none_without_an_accelerated_table` — a plain and a wrapped non-accelerated provider both stay scannable.

All three pass locally (`cargo nextest run -p runtime --lib`), as do the pre-existing `find_concrete_table_provider` peel tests.

**Coverage gap, stated plainly:** the explicit-vs-implicit branch in `SearchEngine::search` is not unit-covered. `runtime-search` cannot construct an `AcceleratedTable` (it sits below `runtime` in the crate layering), so exercising it needs either a stub explorer plus a shared `QueryEngine` test double — `TestQueryEngine` is currently private to `rerank.rs`'s test module — or a runtime integration test with a slow-loading dataset. The branch itself is five lines mirroring the adjacent `explicit_datasets_requested` handling, but I'd rather flag it than imply coverage I didn't add. Happy to add either form if a reviewer wants it.

## Checklist

- [x] Root cause identified and confirmed in source
- [x] Regression tests added that fail on the old condition
- [x] No behavior change for `on_registration` / `on_schema_resolved` / caching / scheduler
- [x] New trait method has no default impl, so all implementors were updated
- [x] `cargo fmt` clean, no `unwrap`/`expect` in production code

### DIFF
--- a/crates/runtime-search/src/full_text_udtf.rs
+++ b/crates/runtime-search/src/full_text_udtf.rs
@@ -678,6 +678,13 @@ mod tests {
         ) -> Option<(Vec<&'a T>, Arc<dyn datafusion::datasource::TableProvider>)> {
             None
         }
+
+        fn not_ready_error(
+            &self,
+            _tbl: &Arc<dyn datafusion::datasource::TableProvider>,
+        ) -> Option<datafusion::error::DataFusionError> {
+            None
+        }
     }
 
     fn fields(names: &[&str]) -> Vec<String> {
--- a/crates/runtime-search/src/search_engine.rs
+++ b/crates/runtime-search/src/search_engine.rs
@@ -129,6 +129,14 @@ impl<E: TableProviderExplorer> SearchEngine<E> {
         Ok(searchable_tables)
     }
 
+    /// The error a scan of `tbl` would fail with because its data is not loaded
+    /// yet, or `None` when it can be searched. An unknown table is reported as
+    /// scannable so the existing not-found handling still owns that error.
+    async fn not_ready_error(&self, tbl: &TableReference) -> Option<DataFusionError> {
+        let table_provider = self.df.get_table(tbl).await?;
+        self.explorer.not_ready_error(&table_provider)
+    }
+
     async fn embedding_columns_from_table(&self, tbl: &TableReference) -> Option<Vec<String>> {
         let table_provider = self.df.get_table(tbl).await?;
         let mut embedding_columns: HashSet<String> = HashSet::default();
@@ -522,6 +530,22 @@ impl<E: TableProviderExplorer> SearchEngine<E> {
 
                 async move {
                     let request_context = RequestContext::current(AsyncMarker::new().await);
+
+                    // A dataset still loading its initial data cannot serve a search. Check
+                    // before building the plan: the query text is embedded during logical
+                    // optimization, so reaching this at physical planning instead would mean
+                    // paying for an embedding round trip only to be rejected. An explicitly
+                    // requested dataset is an error; when searching every searchable dataset,
+                    // skip it rather than failing the whole request — mirroring how an
+                    // unsearchable dataset is handled below.
+                    if let Some(not_ready) = self.not_ready_error(&tbl).await {
+                        if explicit_datasets_requested {
+                            return Err(Error::DataFusionError { source: not_ready });
+                        }
+                        tracing::debug!("Excluding dataset {tbl} from search: {not_ready}");
+                        return Ok((tbl.clone(), None));
+                    }
+
                     let embedding_columns = self.embedding_columns_from_table(&tbl).await.unwrap_or_default();
                     let mut generators: Vec<Arc<dyn CandidateGeneration>> = Vec::with_capacity(embedding_columns.len());
                     for (i, col) in embedding_columns.iter().enumerate() {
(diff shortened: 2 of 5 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12044, opened 2026-07-25
fix(mysql): take the binlog rotate target from the ROTATE event, not its header offset (fixes #12042)

### LINKED ISSUE #12042: MySQL CDC: shared binlog connection (#11814) silently stalls per-dataset replication — HTAP mysql-duckdb gate fails with missing rows
## Summary

Since #11814 (MySQL Shared binlog connection, merged 2026-07-23), the scheduled HTAP CH-benCHmark config `accelerated/mysql-duckdb[file].yaml` @ SF1 fails on every run: MySQL binlog replication **silently stalls per-dataset** partway through the run, worst-case replication lag p99 jumps from ~1 s to **300–620 s**, and the data-correctness gate fails with **rows permanently missing after the full 1800 s drain**. This is a data-correctness regression, not just a perf one.

## Benchmark evidence (testoperator_run_htap.yml, trunk, SF1)

| Run | spiced commit | Conclusion | Freshness worst p99 | Repl lag worst p99 | Verdict |
|---|---|---|---|---|---|
| [30034022813](https://github.com/spiceai/spiceai/actions/runs/30034022813) | `15596bd301` | success | 225 ms | 993 ms | PASSED |
| [30073715975](https://github.com/spiceai/spiceai/actions/runs/30073715975) | `d5618eb718` | failure | 595,208 ms | 602,726 ms | FAILED — 4 tables mismatched |
| [30117845479](https://github.com/spiceai/spiceai/actions/runs/30117845479) | `cfd1ff5a29` | failure | 296,963 ms | 301,679 ms | FAILED |
| [30148593607](https://github.com/spiceai/spiceai/actions/runs/30148593607) | `b4169c1495` | failure | 615,733 ms | 621,716 ms | FAILED — 6 tables mismatched |

Confirmed over 3 consecutive scheduled runs (2026-07-24 → 2026-07-25). Tip-run row counts after the 30-minute drain:

```
new_order      source=51,781     spice=17,605     MISMATCH
oorder         source=350,474    spice=104,325    MISMATCH
order_line     source=3,506,579  spice=1,792,013  MISMATCH
customer/district/warehouse  counts ok, content MISMATCH
```

## Failure fingerprint (run 30073715975, first bad)

- All 7 replicated datasets log `dataset joined shared mysql binlog group … members=1` from the new `data_components::mysql_replication::shared` module, then bootstrap snapshots complete normally.
- Under load, individual datasets **stop receiving changes one by one, silently**: at the load-phase scrape, `warehouse` had been frozen for ~610 s, `new_order` ~373 s, `district` ~103 s — while sibling datasets on the same source (`order_line`, `stock`, `customer`) kept applying millions of changes.
- Stalled members report `lag_bytes=0` (client-side view believes it is caught up), `recv_errs=0`, `reconnects=0`, and there are **no** `member sink is not draining` warnings — so the must-deliver/backpressure path never fired; the pump simply stopped delivering to those members.
- Frozen lag values persist unchanged through the entire 1800 s drain (`drain-probe` lines show identical `max_bench_ts lag` second after second) → the events are lost to those members, not delayed.

Per-dataset replication metrics at the load-phase scrape:

```
dataset      lag_ms   lag_p99_ms  lag_max_ms  lag_bytes  inserts   updates   recv_errs  reconnects
customer        815         992        1015     316933         0    469976          0           0
district     103815       95726      103815          0         0    439400          0           0
new_order    373815      365726      373815          0     74323         0          0           0
warehouse    610815      602726      610815          0         0     71541          0           0
order_line      815         992        1015      93132   2561277   2226601          0           0
stock           815         994        1022     324882         0   2561205          0           0
```

## Attribution

- Trunk range between last good and first bad run is `15596bd301..d5618eb718` (9 PRs); #11814 is the only change touching MySQL replication, and the stalling module (`mysql_replication/shared.rs`) is introduced by it.
- Controls at the same commits: `postgres-duckdb[file]` SF1 healthy (rules out gate/infra/duckdb-alone); `mysql-cayenne[file]` SF1 runs the same shared-binlog path and PASSES all 7 tables + 22/22 content checks (rules out MySQL-side infra). The failure manifests specifically with the slower duckdb file-mode apply path, suggesting the bug is triggered by a slow-draining member (backpressure/ack/rejoin handling in the shared pump).
- Independent of #12007: the tip failing run is at `b4169c1495`, which already includes the #12030 heartbeat fix, and this config is duckdb-accelerated, not cayenne.

## Observations that may help

- Every dataset joins with `members=1` — if the intent is one shared group per connection, the group key may never be matching (7 single-member groups), in which case the regression is in the new per-member pump/ack/checkpoint machinery itself rather than in multi-member fan-out.
- Progressive one-by-one member stalls with no detach/error logs suggest members are being starved after some event (rotate/table-map/ack?) rather than dropped via `detach_member` (which would log).

## Repro

Scheduled HTAP benchmark, config `accelerated/mysql-duckdb[file].yaml`, scale factor 1 — fails deterministically on all 3 runs since the merge.

claudespice:
## Summary

MySQL CDC datasets silently stopped receiving changes partway through a run — rows permanently missing after a full drain, replication-lag p99 climbing from ~1 s to 300–620 s, and no error, detach, or backpressure warning anywhere in the logs.

A real `ROTATE` event is the one binlog event whose header offset does **not** belong to the file it names: `log_pos` is the end of the event in the file being *closed*, while the payload names the file being *opened* and the offset to resume reading at (normally 4). The shared pump updated `current_file` from the payload but kept `event_end_pos` from the header, then fell through to the idle safe-advance and credited idle members `BinlogPosition { file: <new file>, pos: <closed file's end> }`.

`AckSlot::commit` is a monotonic max, so that coordinate stuck. Every subsequent transaction in the new file has a smaller offset, so `deliver_commit`'s `slot.already_committed(commit_pos)` guard returned true and skipped delivery at an unlogged `continue` — for as long as the new file stayed smaller than the old one, i.e. the rest of the run.

This also matches the reported fingerprint exactly: `poll_head_and_heartbeat` reports `lag = head.pos.saturating_sub(resume.pos)`, which saturates to **0** once a member's committed offset is past the head — the stalled datasets all reported `lag_bytes=0` while frozen. Members with an in-flight envelope at the moment of the rotate are skipped by `credit_idle`, which is why high-traffic tables kept flowing while quiet ones froze one by one.

The same poisoned coordinate is what `persist_all` writes to the durable sidecar, so a restart would have resumed from an offset that does not exist in that file.

Fixes #12042

## Changes

- `rotate_target()` — a small helper returning where a `ROTATE` repositions the stream (`None` for the fake rotate at the head of a dump), taking **both** the file name and the offset from the event payload.
- The pump's rotate arm now sets `current_file` and `event_end_pos` together from that helper, so the idle safe-advance, the checkpoint, and any later commit comparison all agree on one file.

No behavior change outside the rotate boundary; every other event's header offset genuinely belongs to `current_file`.

## Test plan

- `make lint`
- `make test`

New unit tests in `crates/data_components/src/mysql_replication/shared.rs`:

- `rotate_targets_the_new_files_resume_offset` — a rotate closing `binlog.000041` at ~1 GiB and opening `binlog.000042` repositions to `binlog.000042:4`.
- `a_fake_rotate_does_not_reposition_the_stream` — the artificial rotate at the head of a dump opens no new file.
- `a_rotate_credit_does_not_suppress_the_new_files_commits` — regression test: after the rotate credit, the first transaction of the newly opened file is **not** reported as already-applied.

Verified the two rotate tests **fail** when the helper is reverted to the pre-fix pairing and **pass** with the fix; the other 14 tests in the module are unaffected either way.

## Checklist

- [x] Existing tests in the module still pass
- [x] Both CI clippy passes (lib and tests) and `cargo fmt` clean
- [x] No sibling instances — `shared.rs` is the only `RotateEvent` handler in the repo

### DIFF
--- a/crates/data_components/src/mysql_replication/shared.rs
+++ b/crates/data_components/src/mysql_replication/shared.rs
@@ -94,7 +94,7 @@ use arrow::datatypes::SchemaRef;
 use async_trait::async_trait;
 use futures::{StreamExt, stream};
 use mysql_async::Conn;
-use mysql_async::binlog::events::{EventData, RowsEventData, TableMapEvent};
+use mysql_async::binlog::events::{EventData, RotateEvent, RowsEventData, TableMapEvent};
 use rustc_hash::FxHashMap;
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
@@ -1492,7 +1492,9 @@ async fn run_pump(source: Arc<SharedSource>) {
             };
 
             let header = event.header();
-            let event_end_pos = u64::from(header.log_pos());
+            // Offset within `current_file` this event advances the stream to. A
+            // real `ROTATE` rewrites both together (see [`rotate_target`]).
+            let mut event_end_pos = u64::from(header.log_pos());
             let event_timestamp = header.timestamp();
             let data = match event.read_data() {
                 Ok(data) => data,
@@ -1505,8 +1507,14 @@ async fn run_pump(source: Arc<SharedSource>) {
 
             match data {
                 Some(EventData::RotateEvent(rotate)) => {
-                    if !rotate.is_fake() {
-                        current_file = rotate.name().into_owned();
+                    // Take BOTH the file and the offset from the event: its
+                    // header offset belongs to the file being closed, so keeping
+                    // `event_end_pos` here would credit idle members a position
+                    // the newly opened file will not reach for a long time. See
+                    // [`rotate_target`].
+                    if let Some(target) = rotate_target(&rotate) {
+                        current_file = target.file;
+                        event_end_pos = target.pos;
                     }
                 }
                 Some(EventData::TableMapEvent(tme)) => {
@@ -1754,6 +1762,21 @@ async fn run_pump(source: Arc<SharedSource>) {
 
 fn current_file_pos(file: &str) -> BinlogPosition {
     BinlogPosition::new(file.to_string(), 0)
+}
+
+/// Where a `ROTATE` event repositions the shared stream, or `None` for the fake
+/// rotate the server sends at the head of a dump (which opens no new file).
+///
+/// `ROTATE` is the one event whose header offset does **not** belong to the file
+/// it names: `log_pos` is the end of the event in the file being *closed*, while
+/// the payload names the file being *opened* and the offset to resume reading at
+/// (normally 4, just past the magic number). Pairing the new name with the
+/// closing file's offset yields a coordinate far beyond anything the new file
+/// holds — an idle member credited there has every later commit in that file
+/// suppressed by [`AckSlot::already_committed`], silently and for as long as the
+/// new file stays smaller than the old one (#12042).
+fn rotate_target(rotate: &RotateEvent<'_>) -> Option<BinlogPosition> {
+    (!rotate.is_fake()).then(|| BinlogPosition::new(rotate.name(), rotate.position()))
 }
 
 /// Deliver a committed transaction's buffered rows to their members, then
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12048, opened 2026-07-25
fix(mysql): verify an adopted layout against the event's own column types (fixes #11764)

### LINKED ISSUE #11764: fix(mysql): bind mid-stream layout adopt to event-time schema under replication lag
## Summary

When MySQL binlog replication is behind the source and multiple compatible same-count DDLs (reorder / retype) land before Spice processes the first `ALTER` / `TableMap`, `adopt_current_layout` fetches **current** `information_schema` — not the layout that applied at the event being decoded.

Spice can then decode historical row images with the **final** ordinal map while those images still reflect an intermediate layout → silent column scramble. The durable fingerprint can also advance to the final layout, so restart compatibility will not catch already-corrupted rows.

## Scenario

1. Stream lagging at position P0 under layout L0.
2. Source applies ALTER₁ (L0→L1, same column count) then ALTER₂ (L1→L2, same column count).
3. Spice processes ALTER₁ / TableMap for L1, but `information_schema` already shows L2 → adopts L2.
4. Remaining row images between ALTER₁ and ALTER₂ decode under L2 → mis-map.
5. Checkpoint fingerprint becomes L2; restart looks \"compatible\".

## Current mitigations

- Resume fingerprinting catches layout drift **after a crash/restart** when the durable fingerprint still matches an older layout.
- Docs already recommend quiescing writes around DDL.
- Same-count mid-stream reshapes still rely on ALTER detection (`TableMap` compares column count only).

## Proposed directions

- Fail closed for same-count ALTERs unless the layout can be tied to the binlog event (e.g. TableMap type metadata comparison, or refuse adopt when source head has moved past additional DDL).
- Or require / document that DDL must be quiesced while the stream is caught up, and treat lagging same-count ALTER as fatal.
- Longer term: event-time schema binding (parse DDL / use TableMap column defs) instead of live `information_schema`.

## Related

- https://github.com/spiceai/spiceai/pull/11761

claudespice:
Fixes #11764

## Summary (root cause)

A compatible mid-stream `ALTER` is adopted by `adopt_current_layout`, which re-reads `information_schema`. That describes the source table **as it is now**, not as it was at the event being decoded.

While replication is behind, the source can apply a second same-column-count `ALTER` — a reorder (`MODIFY ... FIRST/AFTER`) or a rename swap — before Spice reaches the first one. The layout adopted for `ALTER₁` is then really the layout after `ALTER₂`, and it maps ordinals that the row images still in flight do not use. Values land in the wrong columns and **nothing fails**: the column counts agree, every name still resolves, and the types still convert. The durable fingerprint then advances to the final layout, so a restart sees a "compatible" table and the already-scrambled rows stand.

Count-changing composites already fail closed (the `TableMap` count guard), and name-breaking ones fail in `column_map`. What slipped through silently was exactly the same-count ordinal shuffle.

## Changes

The `TableMap` event is the one description of the row image that travels **with** it, so it is the authority on what the in-flight rows look like.

- `layout_event_mismatch` compares the layout that will decode a table's row images against the column types the `TableMap` event itself carries.
- It is called at the point every decode is routed from, so it gates both adopt paths (the `ALTER TABLE` statement path and the `TableMap` column-count path) and any other drift between the in-memory layout and the events. A disagreement is member-fatal — that one dataset detaches with an actionable error, the rest of the shared dump keeps running.
- The comparison is deliberately coarse, because this gate runs on every routed `TableMap` and a false positive would break a healthy stream — strictly worse than the rare scramble it detects:
  - `DATETIME`/`TIMESTAMP`/`TIME` (the `*2` variants since 5.6), `TEXT`/`BLOB` (all sent as `MYSQL_TYPE_BLOB`, distinguished only by a metadata length byte) and `REAL` (`REAL_AS_FLOAT` sql_mode) are **not compared at all**.
  - `CHAR`/`VARCHAR`/`BINARY`/`VARBINARY`/`ENUM`/`SET` are **one class**, since which wire type the server picks inside that family depends on charset and on the `MYSQL_TYPE_STRING` metadata.
  - Anything unmapped, out of range, or unrecognized is skipped rather than flagged.

  What remains is decisive: a reorder that moves columns of genuinely different types (int↔varchar, int↔bigint, decimal↔int).

This detects and fails rather than attempting event-time schema binding, per the direction in the issue discussion.

### Scope

The sibling Postgres CDC path is not affected: it reconciles against the in-stream `Relation` message, which travels with the events and is therefore already event-time correct. The MySQL resume path is separately guarded by the layout fingerprint check.

## Test plan

- 8 new unit tests in `mysql_replication::binlog` covering the reorder that motivated the issue, first-disagreement reporting, integer-width distinction, the skip-when-unreadable path, and two tests that pin the conservatism of the mapping (version-dependent types stay unmapped; the string family stays one class).
- The four detection tests were verified to **fail** with the comparison neutered to its pre-fix behavior, and pass with it restored.
- `make lint`
- `make test`

## Checklist

- [x] Both CI clippy passes (`--lib --bins`, then `--tests`) and `cargo fmt --check` clean
- [x] No new `.unwrap()`/`.expect()` in production code
- [x] Error message names the dataset/table and gives an actionable fix
- [x] Bug class checked across sibling connectors

### DIFF
--- a/crates/data_components/src/mysql_replication/binlog.rs
+++ b/crates/data_components/src/mysql_replication/binlog.rs
@@ -29,7 +29,8 @@ limitations under the License.
 //!   - [`readiness_heartbeat`] / [`record_watermark`] / [`commit_ts_ms`] for
 //!     lag-based readiness and freshness metrics,
 //!   - [`adopt_current_layout`] + [`compute_pk_source_indexes`] for compatible
-//!     mid-stream `ALTER` adoption,
+//!     mid-stream `ALTER` adoption, and [`layout_event_mismatch`] to check an
+//!     adopted layout against the row images it will actually decode,
 //!   - [`classify_query`] / [`classify_statement`] for the transaction and DDL
 //!     boundaries in the event stream.
 //!
@@ -46,6 +47,7 @@ use std::time::{Duration, SystemTime};
 use arrow::datatypes::SchemaRef;
 use mysql_async::binlog::events::{RowsEventData, TableMapEvent};
 use mysql_async::binlog::row::BinlogRow;
+use mysql_async::consts::ColumnType;
 use mysql_async::{BinlogStream, BinlogStreamRequest, Conn, Value};
 
 use super::config::{BinlogPosition, ReplicationParams};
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12050, opened 2026-07-25
fix(accelerator): let an index declare a finalize failure fatal so a stale index can't report a successful refresh (fixes #12038)

### LINKED ISSUE #12038: Index finalize failures (on_write_complete) are best-effort, so a stale index can report a successful refresh
## Problem

`TableSink` and `MultiSink` treat `Index::on_write_complete` failures as best-effort — they log a warning and let the write report success:

```rust
if let Err(e) = index.on_write_complete().await {
    tracing::warn!("... on_write_complete failed for index '{}': {e}. Index may be stale until next refresh.", index.name());
}
```

For indexes whose `on_write_complete` finalizes durable state the write depends on, that means a failed finalize leaves the index **stale while the refresh/append reports success** — search then serves stale or missing results with no error surfaced to the caller.

This became reachable for full-text search once the tantivy commit moved from per-batch (inside `compute_index`, where an error aborted the write) to a single commit in `on_write_complete` (#12015). It also applies to any other index that finalizes state in that hook.

## Proposed fix

Add a capability method to the `Index` trait so each index declares whether a finalize failure is fatal, defaulting to today's best-effort behavior:

```rust
/// Whether a failure in `on_write_complete` must fail the write.
/// Defaults to `false` (best-effort), matching indexes that create state with
/// IF NOT EXISTS semantics. Indexes that finalize durable state the write
/// depends on return `true`.
fn write_complete_failure_is_fatal(&self) -> bool { false }
```

Then have `TableSink` and `MultiSink` propagate the error when the flag is set, and return `true` from the full-text index.

## Notes

- Deliberately **not** bundled into #12015: it changes shared write-failure semantics for every index type (Elasticsearch uses the same hooks), so it warrants its own focused review.
- Per the trait-evolution guidance in `CLAUDE.md`, adding the method requires auditing every `Index` impl and any wrapper so the capability is forwarded rather than silently defaulted.
- Needs a test asserting a refresh fails when the index's finalize step fails.

claudespice:
## Summary

`TableSink` and `MultiSink` treated every `Index::on_write_complete` failure as best-effort — log a warning, let the write report success. For an index whose finalize step publishes durable state the written data depends on, that leaves the index **stale while the refresh or append reports success**, and search then serves results that silently omit the whole write with no error reaching the caller.

The root cause is that the finalize hook had no way to say whether its failure mattered. Creating an index with `IF NOT EXISTS` semantics genuinely is best-effort — the next refresh redoes it — but a finalize that *publishes* the write is not, and both went down the same warn-and-continue path, duplicated in two sinks.

## Changes

- **`Index::write_complete_failure_is_fatal()`** — new capability method defaulting to `false` (today's behavior, so nothing changes for an index that doesn't opt in). Its doc comment states the wrapper-forwarding requirement.
- **`finalize_indexes()`** in `accelerated_table::sink` — one implementation, now shared by `TableSink` and `MultiSink` in place of two copies of the same loop. It finalizes **every** index even after one has failed (stopping early would leave more indexes stale than the failure requires), logs non-fatal failures exactly as before, and returns the **first** fatal failure, wrapped with the failing index's name, so the sink reports the write as failed.
- **Wrapper forwarding** for all four wrapper indexes: `ChunkedSearchIndex` / `ChunkedVectorIndex` delegate to their inner index; `CompoundSearchIndex` / `CompoundVectorIndex` return the union of both halves, since either half failing to finalize leaves the compound index stale. Per the trait-evolution rule in `CLAUDE.md` these are precisely the impls where inheriting the default would silently no-op.
- **`FullTextDatabaseIndex` returns `true`** — its documents only become searchable once the tantivy writer commits, and staged-but-uncommitted documents are discarded.

### Scope decisions worth a reviewer's eye

- **Elasticsearch deliberately keeps the default (`false`).** `ElasticsearchWriteMaintenance::on_write_complete` restores `refresh_interval` and issues `_refresh`/`force_merge`; the documents were already bulk-written durably per batch, so a failure there delays visibility rather than losing the write, and the next write cycle re-applies the setting. Flagging rather than changing it — say the word and I'll make it fatal too.
- The `on_write_complete` call in `crates/runtime/src/embeddings/index/duckdb.rs` is untouched: it is an init-time index creation outside any write lifecycle, where "not created yet" is the expected outcome, not a stale index.
- On current `trunk`, `FullTextDatabaseIndex` commits per batch inside `compute_index` (a commit failure already aborts the write), so its finalize hook is a no-op and the new flag is not yet load-bearing. #12015 moves that commit into `on_write_complete`, which is what makes it so — this PR supplies the mechanism that change needs.

## Test plan

New unit tests, all failing against the old warn-and-continue behavior:

- `crates/runtime/src/accelerated_table/sink/mod.rs` — six tests over `finalize_indexes`: every index finalized when none fail; a non-fatal failure leaves the write successful; **a fatal failure fails the write** (regression test) and the error names both the index and its own message; a fatal failure still finalizes the indexes after it; the *first* fatal failure is the reported one; an empty index set succeeds.
- `crates/search/src/index/compound/tests.rs` — fatality is the union of both halves, over all four primary/secondary combinations, for both `CompoundSearchIndex` and `CompoundVectorIndex`.
- `crates/search/src/index/chunking.rs` — `ChunkedSearchIndex` and `ChunkedVectorIndex` each forward a fatal inner index rather than inheriting the default.

Verification:

- `make lint`
- `make test`

## Checklist

- [x] Every wrapper impl of the changed trait audited and explicitly forwarding
- [x] Default preserves existing behavior for indexes that don't opt in
- [x] `make lint`
- [x] `make test`

Fixes #12038

### DIFF
--- a/crates/runtime-datafusion-index/src/lib.rs
+++ b/crates/runtime-datafusion-index/src/lib.rs
@@ -78,5 +78,19 @@ pub trait Index: Debug + Send + Sync + 'static {
         Ok(())
     }
 
+    /// Whether a failure in [`Index::on_write_complete`] must fail the write.
+    ///
+    /// Defaults to `false` (best-effort), matching indexes whose finalize step has
+    /// `IF NOT EXISTS` semantics and is simply redone on the next refresh. An index
+    /// that finalizes durable state the written data depends on returns `true`: for
+    /// those, a failed finalize leaves the index stale while the write reports
+    /// success, so the sink reports the write as failed instead.
+    ///
+    /// Wrapper implementations MUST forward this to the index they wrap — inheriting
+    /// the default silently downgrades a fatal inner index to best-effort.
+    fn write_complete_failure_is_fatal(&self) -> bool {
+        false
+    }
+
     fn as_any(&self) -> &dyn Any;
 }
--- a/crates/runtime/src/accelerated_table/sink/mod.rs
+++ b/crates/runtime/src/accelerated_table/sink/mod.rs
@@ -15,7 +15,8 @@ limitations under the License.
 */
 
 use datafusion::{
-    catalog::TableProvider, logical_expr::dml::InsertOp, physical_plan::RecordBatchStream,
+    catalog::TableProvider, error::DataFusionError, logical_expr::dml::InsertOp,
+    physical_plan::RecordBatchStream,
 };
 use multi::MultiSink;
 use runtime_datafusion_index::Index;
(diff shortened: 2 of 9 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12051, opened 2026-07-25
fix(write-back): gate durable write-back on a safe source delivery primitive (fixes #11915)

### LINKED ISSUE #11915: Cayenne durable write-back delivery can silently lose a committed write via its own CDC delete-echo
## Summary

The durable federated write-back delivery worker (`crates/runtime/src/accelerated_table/write_back_worker.rs`, `deliver_batch`) can **silently and permanently lose a committed, gate-approved write**. No crash is required — a transient source-connector error is sufficient.

## Mechanism

Delivery reconciles a committed accelerator row to the federated source as **two separate source commits**: `delete_from(pk)` then `insert(present rows, InsertOp::Append)`. The accelerator is CDC-fed *from the same source* and has no self-echo suppression, so:

1. Transaction commits `K = v` on the accelerator; marker `(K, seq)` is durable.
2. Worker pass 1: reads `K = v`; `delete_from(K)` **commits** at the source; the follow-up `insert` **fails transiently** → `deliver_batch` errors → markers not cleared (good so far).
3. During the backoff, the source's CDC stream delivers the `DELETE K` echo, which applies to the accelerator → the committed row `K = v` is removed from the accelerator.
4. Worker pass 2: reads `K` **absent** → source delete is a no-op → nothing to insert → `clear_dirty_keys` succeeds (marker sequence unchanged).

Final state: `K` absent at the source **and** the accelerator, marker gone — the committed write is lost everywhere, with no error surfaced. The same shape occurs on a worker crash between the two source commits. Even the happy path bounces the committed row through a transient DELETE-echo vanish window.

## Why it isn't fixed in the OCC/IVM PR

There is no primitive-preserving small fix. Source delivery uses `InsertOp::Append` and the write-back path explicitly rejects `Replace`/upsert (`validate_insert_op`), so present keys **must** be delete-first-then-insert, and that standalone delete's echo is the loss. Closing it requires one of:

- **Atomic delete+insert in a single source transaction** (so no observable/echo-able deleted state exists), or
- **Native conditional upsert** delivery (`INSERT … ON CONFLICT DO UPDATE`) — removes the delete leg for present keys, and the natural home for a delivery-time CAS for mixed-writer safety, or
- **Self-echo suppression** (replication-origin tagging) so the delete echo never mutates the accelerator (#11840).

All three are connector-capability-dependent. Proposed direction: gate `durable_write_back` on the source connector advertising a safe delivery primitive and **fail registration** otherwise (per the config rule "validate against capability and return a structured configuration error — never silently ignore"), then implement transactional/upsert delivery per-connector (Postgres/MySQL first).

## Acceptance

- A committed write survives a transient source insert failure and a worker crash between the two delivery commits (both the source and the accelerator retain `K = v`; the marker is cleared only after positive delivery).
- Regression test: force one insert failure + inject the delete echo between passes (the `SignalingTableProvider` + deterministic-drain pattern from #11889 fits; no fixed sleeps).

Found during a Cayenne correctness audit of #11870. The related per-key OCC + fused-transaction IVM holes are fixed in #11916.

claudespice:
Fixes #11915

## Summary

Durable federated write-back could silently and permanently lose a committed write. No crash was required — a transient source error was enough.

Delivery reconciles a committed accelerator row to the source as **two separate source commits**: `delete_from(pk)`, then `insert(present rows, InsertOp::Append)`. Because the accelerator is CDC-fed from that same source and has no self-echo suppression, the standalone `DELETE` echoes back and erases the committed row from the accelerator. If the follow-up `INSERT` then fails:

1. Pass 1 deletes at the source; the insert fails, so markers are correctly left in place.
2. During the backoff the source's CDC stream delivers the `DELETE` echo, removing `K = v` from the accelerator.
3. Pass 2 reads `K` as **absent** → the source delete is a no-op → nothing to insert → `clear_dirty_keys` succeeds.

The row is now gone from the source *and* the accelerator, the marker is gone, and nothing was reported. The same shape occurs on a worker crash between the two commits.

There is no primitive-preserving small fix: source delivery uses `InsertOp::Append` and the write-back path rejects `Replace`/upsert, so present keys must be delete-first-then-insert, and that standalone delete's echo is the loss. Closing it properly requires an atomic delete+insert, a native conditional upsert, or replication-origin tagging — all connector-capability-dependent.

This PR implements the first step the issue scopes: **gate the configuration on the connector advertising a safe delivery primitive, and fail registration otherwise**, rather than accepting a config that can lose data.

## Changes

- **`Acceleration::resolves_to_durable_write_back()`** (`runtime-acceleration`) — one predicate for "this config turns durable write-back on" (Cayenne + `write_mode: write_back` + non-empty `on_conflict` + `refresh_mode: changes`). The Cayenne provider previously inlined these conditions; it now calls this, so the gate and the behavior it guards cannot drift apart.
- **`DataConnector::supports_durable_write_back_delivery()`** — new capability, defaulting to `false`. Modeled on the existing `supports_changes_stream` precedent. Returning `true` asserts delivery is atomic from the source's point of view (one transaction covering both legs, or a native `INSERT … ON CONFLICT DO UPDATE` that removes the delete leg).
- **Forwarded in every wrapper** — `EmbeddingConnector`, `FullTextConnector`, `ElasticsearchFullTextConnector`, `DeferredConnector`. `DeferredConnector` forwards this even though it hard-returns `false` for the stream capabilities: this is a static property of the source it will eventually load from, not of the placeholder (matching how it already treats `resolve_refresh_mode`).
- **Registration gate** in `register_loaded_dataset`, beside the existing `refresh_mode: changes` gate and before `read_provider`, so an unsafe dataset fails fast without touching the network. New `Error::DurableWriteBackUnsupportedBySource` names the dataset and connector and gives the actionable alternatives.

No connector opts in yet, so this disables durable write-back until per-connector safe delivery lands — which is the intended sequencing in the issue ("gate … and fail registration otherwise, then implement transactional/upsert delivery per-connector").

## Scope note

This does **not** yet satisfy the issue's literal acceptance bullets — a committed write surviving a transient insert failure and a worker crash still needs the per-connector atomic/upsert delivery the issue scopes as the follow-up (Postgres/MySQL first). This PR closes the data-loss window by refusing the unsafe configuration instead of silently accepting it. Happy to split differently if a reviewer would rather see delivery land first.

## Test plan

- `make lint`
- `make test`

New tests:

- `durable_write_back_is_recognized` — the exact enabling config is detected.
- `every_condition_is_required_for_durable_write_back` — flips each conjunct independently (including all five non-`changes` `RefreshMode`s and an unset one) so no condition can silently stop being load-bearing.
- `a_default_acceleration_is_not_durable_write_back` — the gate cannot fire on ordinary datasets.
- `every_wrapper_forwards_durable_write_back_delivery_support` — a wrapper that inherits the `false` default would report a safe source as unsafe and reject a valid dataset. That is the defaulted-no-op wrapper bug (#10460) applied to this capability; it compiles cleanly, so only a test catches it. (`ElasticsearchFullTextConnector` is excluded — it is feature-gated and needs a params-bearing dataset to construct.)

## Checklist

- [x] Tests added, including a guard for the wrapper-forwarding regression class
- [x] Capability defaults to the conservative value; config validated against it with a structured error
- [x] No `.unwrap()`; `cargo fmt` clean

### DIFF
--- a/crates/runtime-acceleration/src/acceleration.rs
+++ b/crates/runtime-acceleration/src/acceleration.rs
@@ -444,6 +444,28 @@ impl Acceleration {
         self
     }
 
+    /// Returns whether this configuration resolves to durable federated
+    /// write-back: a Cayenne `write_mode: write_back` dataset that declares an
+    /// `on_conflict` key and is CDC-fed (`refresh_mode: changes`).
+    ///
+    /// Such a dataset marks every committed write's primary keys so a delivery
+    /// worker can reconcile them to the federated source. Delivery mutates the
+    /// source, so the source connector must advertise a delivery primitive that
+    /// cannot lose the write; callers gate on
+    /// `DataConnector::supports_durable_write_back_delivery` and reject the
+    /// dataset at registration otherwise.
+    ///
+    /// This is the single source of truth for the condition: both the
+    /// registration gate and the Cayenne provider that turns the marking on
+    /// call it, so the check and the behavior cannot drift apart.
+    #[must_use]
+    pub fn resolves_to_durable_write_back(&self) -> bool {
+        self.engine == Engine::Cayenne
+            && self.write_mode == spicepod_acceleration::WriteMode::WriteBack
+            && !self.on_conflict.is_empty()
+            && self.refresh_mode == Some(RefreshMode::Changes)
+    }
+
     /// Returns whether Arrow `hash_index` is enabled for primary key upserts or indexes.
     #[must_use]
     pub fn is_hash_index_enabled(&self) -> bool {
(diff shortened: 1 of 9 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12054, opened 2026-07-26
fix(http): restore a join's embedded projection in HTTP subquery pushdown (fixes #11009)

### LINKED ISSUE #11009: SQL table functions: IN (SELECT ... FROM args) with HTTP connector fails with schema mismatch
## Summary

SQL table functions using `IN (SELECT ... FROM args)` with an HTTP connector dataset fail with an internal schema mismatch error. The `HttpWithDeferredParamsExec` semi-join rewrite adds the deferred parameter column to the output, but the table function's output schema validation expects only the declared return columns.

### `spicepod.yaml`

```yaml
version: v2
kind: Spicepod
name: repro

runtime:
  functions:
    enabled: true

datasets:
  - from: https://hn.algolia.com/api/v1
    name: raw_users
    params:
      file_format: json
      allowed_request_paths: '/users/*'

functions:
  - name: hn_user_in
    from: sql
    kind: table
    volatility: volatile
    signature:
      args: [{ name: username, type: utf8 }]
      returns:
        - { name: content, type: utf8 }
    body: |
      SELECT content
      FROM raw_users
      WHERE request_path IN (SELECT concat('/users/', username) FROM args)
```

### Query

```sql
SELECT * FROM hn_user_in('pg');
```

### Error

```
Internal Error: Schema mismatch: Expected and actual number of fields in the query result don't match: expected 1, received 2
```

### EXPLAIN output

```
physical_plan | HttpWithDeferredParamsExec: deferred_param=request_path, build_side=ProjectionExec
              |   HttpExec: base_url=https://hn.algolia.com/api/v1, format=json, partitions=deferred
              |   ProjectionExec: expr=[/users/pg as concat(Utf8("/users/"),Utf8("pg"))]
              |     DataSourceExec: partitions=1, partition_sizes=[1]
```

## Root cause

The `HttpParamsPushdown` physical optimizer rule rewrites `IN (SELECT ...)` into a `HttpWithDeferredParamsExec` semi-join. This adds the join key column (`request_path`) to the output schema. The SQL table function's `validate_output_schema` then sees 2 columns (`content` + `request_path`) but expects only 1 (`content`), and fails.

## Notes

- The equivalent `= (SELECT ...)` scalar subquery pattern works correctly.
- This is independent of the args inlining fix in #10998 — it reproduces with or without that change.

claudespice:
## Summary

`HttpParamsPushdown` replaces a semi-`HashJoinExec` over an `HttpExec` with an `HttpWithDeferredParamsExec` whose output schema is the http side's schema. But a `HashJoinExec` can carry its own projection, and when it does, *that* is its output schema — not the full join schema. The rewrite never consulted it, so it silently widened (or reordered) the plan's output schema relative to the logical plan it was planned from.

`SELECT content FROM <http_dataset> WHERE request_path IN (SELECT …)` plans as a `LeftSemi` join whose embedded projection keeps only `content`. After the rewrite the plan emits all 8 http columns, and the query fails against its own logical plan:

```
Internal Error: Schema mismatch: Expected and actual number of fields in the query result don't match: expected 1, received 2
```

Two details worth noting for review:

- The reported error comes from the query path's `verify_schema` (`crates/runtime/src/datafusion/query.rs`), which compares the logical plan's schema against the result stream's — not from the SQL table function's own `validate_output_schema`, which runs on the logical plan and never sees the physical rewrite. The issue reproduced it through a SQL table function, but nothing about the bug is specific to table functions; that path just happens to make the widened schema visible.
- A projection can also *reorder* columns. There the widened schema is not necessarily rejected — a consumer binding columns by index reads one column's values under another's name, which is silent wrong data rather than an error.

The rule is registered with `with_physical_optimizer_rule`, which appends to DataFusion's default rules — so `ProjectionPushdown` has already embedded the projection into the join by the time this rule runs. That is why the `EXPLAIN` in the issue shows no `ProjectionExec` above the rewritten node.

`HttpParamsPushdown` opted out of the optimizer's `schema_check`, which is why neither DataFusion nor the rule itself caught this.

Fixes #11009

## Changes

- `try_rewrite_hash_join` now restores a projection embedded in the join, via a new `reapply_join_projection` helper, so the rewritten subtree reproduces the schema of the node it replaced. For `LeftSemi`/`RightSemi` the join's unprojected output schema is exactly the http side's schema (`build_join_schema` keeps only the semi side's fields, and a semi join never forces a field nullable), so the join's projection indices address the rewritten node's columns directly. An out-of-range index returns an internal error rather than being silently dropped.
- `schema_check()` now returns `true`. The rewrite is schema-preserving by construction, so holding the rule to that turns any future regression of this shape into a planner error instead of a wrong result reaching the caller.
- Three tests: the projection is restored (regression test for the reported failure), a reordering projection keeps column order, and a join without a projection is still rewritten to the bare deferred node (no wrapper, so `EXPLAIN` and downstream rules are unchanged).

### Bug-class audit

Every other site that replaces a `HashJoinExec` was checked for the same shape and all are already correct:

- `hash_join_optimization.rs` — substitutes `EmptyExec::new(join_exec.schema())`, the projected schema.
- `flightsql/broadcast_join.rs` — builds its `SELECT` list by iterating `join.schema()`, so a projection is honored implicitly.
- `cayenne/src/optimizer_rules.rs` — explicitly declines when `contains_projection()`.

## Performance impact

None in the common path: with no embedded projection the plan is byte-for-byte what it was. When the join does carry a projection the rewrite now adds one `ProjectionExec` over the deferred node — the same column selection the join was doing internally, and negligible next to the HTTP requests it sits above. Queries that previously failed outright now execute.

## Test plan

- [x] `make lint`
- [x] `make test`
- [x] New unit tests fail on the pre-fix code and pass after (verified by reverting the fix)

## Checklist

- [x] Regression test that fails before the fix
- [x] Bug class audited across sibling call sites
- [x] No `.unwrap()` / `.expect()` in non-test code
- [x] `cargo fmt` clean

### DIFF
--- a/crates/datafusion-optimizer-rules/src/physical_plan/http_subquery_pushdown.rs
+++ b/crates/datafusion-optimizer-rules/src/physical_plan/http_subquery_pushdown.rs
@@ -92,7 +92,10 @@ impl PhysicalOptimizerRule for HttpParamsPushdown {
     }
 
     fn schema_check(&self) -> bool {
-        false
+        // The rewrite replaces a semi-join with a node that reproduces the join's
+        // output schema, so hold it to that: a schema change here reaches the caller
+        // as a query result whose columns don't match the plan it asked for.
+        true
     }
 }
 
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12056, opened 2026-07-26
fix(search): expunge superseded documents from full-text BM25 statistics (fixes #12053)

### LINKED ISSUE #12053: Full-text search BM25 scores can reflect superseded documents after an update (search unit tests fail load-dependently on trunk)
## Summary

On current `trunk` (`2cabe6ade`), two full-text-search unit tests fail depending on **how many other tests share the process**, not on the code under test:

```bash
# Fails: 2 tests
cargo test -p search --lib -- index::

# Passes: the same two tests, narrower filter
cargo test -p search --lib -- generation::text_search::index::tests
```

Failing tests:

- `generation::text_search::index::tests::test_updates_overwrites_on_compute_index` (snapshot `after_update_elephant`)
- `generation::text_search::index::tests::test_updates_overwrites_on_compute_index_composite_pk` (snapshot `cpk_after_update_elephant`)

Reproduced on a heavily loaded machine (load average ~13). It also reproduces with `--test-threads=1`, and the set of failing tests varies between runs of the same binary (2 failures, then 1, then 0 as the filter narrows) — so this is timing/load sensitive rather than a deterministic mismatch.

## What differs

Only the BM25 `_score` column. The matched row, its `id`, and its `content` are all correct:

```
expected: | 0.9808291792869568 | dog elephant frog | dog elephant frog | 2 |
actual:   | 1.3862943649291992 | dog elephant frog | dog elephant frog | 2 |
```

Both values are the BM25 IDF for a term with `doc_freq = 1`, differing only in the collection size:

- `ln(1 + (3 - 1 + 0.5) / (1 + 0.5))` = **0.98083** — 3 documents, the expected state
- `ln(1 + (5 - 1 + 0.5) / (1 + 0.5))` = `ln(4)` = **1.38629** — 5 documents

(The tf/field-length factor is exactly 1.0 in both cases: every document in the fixture is 3 tokens long, so it equals the average and cancels. The score *is* the IDF.)

The test indexes 3 documents, then overwrites 2 of them. 5 documents means the two superseded versions are still counted in the collection statistics at query time — even though they are correctly excluded from the *matches* (the preceding `after_update_apple` / `after_update_guitar` assertions, which expect empty results for the replaced content, pass).

## Why this looks like more than a test bug

`update_index` issues `delete_term` for every incoming primary key before adding the new documents, and tantivy applies operations in opstamp order, so the deletes should always retire the superseded rows. The delete terms are derived from the incoming `RecordBatch` PK column (not from a searcher lookup), so they are unconditional. The scores nonetheless observe a collection that still contains the deleted documents, which suggests the searcher can be reloaded to a revision where the commit's documents are visible but its deletions have not yet been applied to the segment.

If that is right, the user-visible consequence is that relevance scores after an update transiently reflect superseded documents. For a single-term query this scales every score equally and does not change the ordering, but a multi-term query weights its terms by per-term IDF, so the *ranking* can differ from the correct one until the deletes are applied.

I have not confirmed the tantivy-side mechanism, so I would not want to guess at a fix — flagging the reproducer and the arithmetic rather than proposing one.

## Notes

- The code these tests cover was rewritten in #12015 (deferred-commit window, merged shortly before this was observed). I have not bisected, so I am not asserting it as the cause — the deferral path is not active in these tests (`defer_commit` is false when `compute_index` is driven directly).
- I deliberately did **not** regenerate the snapshots. Accepting `ln(4)` would bake in the stale-statistics reading.
- Found while resolving a merge conflict on an unrelated branch; the failure reproduces on clean `trunk`, so it is not specific to that work.

claudespice:
## Summary

Full-text relevance scores were computed against documents the index had already replaced.

tantivy derives BM25's collection size from each segment's `max_doc`, which counts documents that have been deleted but not yet physically expunged. Only a merge expunges them, and `FullTextDatabaseIndex` kept tantivy's default `LogMergePolicy`, whose `del_docs_ratio_before_merge` is `1.0` — a ratio no segment can exceed, so deletions never trigger a merge at all. (tantivy's own source calls that default "not a very sensible default", kept for backward compatibility.)

Because `update_index` writes every incoming primary key as a deletion followed by an insertion, every refresh and every CDC change supersedes rows. Once a segment has grown large enough to sit alone at its size level — where the policy's other rule, a minimum of eight segments per level, can never apply either — its superseded documents stay in the collection statistics permanently. A single-term query is scaled uniformly and keeps its order, but a multi-term query weights its terms by their individual inverse document frequencies, so the ranking itself drifts from the one the live rows imply.

Measured on `trunk` with three documents, two of which are then replaced: the surviving document scores `1.3862944` (the IDF at five documents) instead of `0.9808292` (the IDF at three), and never converges.

This is also why `test_updates_overwrites_on_compute_index` and its composite-key twin failed intermittently. Whether the superseded documents were counted depended on whether they happened to land in a segment of their own — dropped whole at commit, never reaching the statistics — or shared one with a live document, in which case the segment was retained.

## Changes

- Cap superseded documents per segment at 25% (`MAX_SUPERSEDED_DOCS_RATIO_PER_SEGMENT`), so a mostly-replaced segment is rewritten and its documents leave the statistics. Once merges settle, the collection size exceeds the live document count by at most `1 / (1 - ratio)`.
- Reinstate the policy after every rollback. `tantivy::IndexWriter::rollback` does not reset the writer, it replaces it with a freshly constructed one carrying the default policy — and `on_write_start` rolls back before every write window, so without this the fix would have lasted exactly until the first refresh. All four rollback sites now go through `rollback_writer`.
- The two snapshot tests wait for the superseded documents to be expunged before asserting a score, since merges are asynchronous. Their snapshots are unchanged.

## Test plan

- `make lint`
- `make test`

Two new tests, each verified to fail with its corresponding guard removed:

- `test_superseded_documents_leave_the_bm25_collection_size` — drives an index into the only state in which superseded documents can linger (one consolidated segment, half of it replaced), asserts they leave the collection size, and asserts that an index updated in place scores identically to one rebuilt from the same final rows. Without the merge-policy cap it times out at `max_doc=8 live=4`.
- `test_merge_policy_survives_a_writer_rollback` — the same fixture after `on_write_start`/`on_write_failed`. Without the reinstatement it times out the same way.

Removing the cap also makes the originally reported `test_updates_overwrites_on_compute_index` fail deterministically, at `max_doc=3 live=1`.

Verified locally: `cargo test -p search --lib` (75 passed), both CI clippy passes, `cargo fmt`.

## Performance impact

Merges now also trigger on superseded documents, which is additional background work for an index that is updated in place. It is bounded and amortized: a segment is rewritten once per 25% of its documents being replaced, which is the same work that reclaims their storage. An index that is only ever loaded, with no deletions, is unaffected — `deletes_ratio` is zero for every segment, so no merge is scheduled that would not already have happened, and no score changes. That is why no existing snapshot needed regenerating, including the search integration snapshots that assert `_score`.

## Checklist

- [x] Root cause confirmed against the tantivy source and reproduced end to end
- [x] Regression tests verified failing without each guard
- [x] No behavior change for indexes without superseded documents
- [x] `cargo fmt` + both CI clippy passes clean

Fixes #12053

### DIFF
--- a/crates/search/src/generation/text_search/index.rs
+++ b/crates/search/src/generation/text_search/index.rs
@@ -49,27 +49,14 @@ use crate::index::SearchIndex;
 pub static MEMORY_BUDGET_FOR_INDEX_WRITER: usize = 150 * 1024 * 1024;
 pub static INDEX_UNIQUE_FIELD_NAME: &str = "__spice.unique_field";
 
-/// The fraction of a tantivy segment's documents that may be superseded — deleted, but
-/// still physically present — before the segment is rewritten by a merge.
+/// The fraction of a tantivy segment's documents that may be superseded/deleted, but
+/// still physically present, before the segment is rewritten by a merge.
 ///
-/// This bounds a relevance-scoring error. tantivy derives BM25's collection size from
-/// each segment's `max_doc`, which counts superseded documents, so every term's inverse
-/// document frequency is computed against a collection that still contains rows the
-/// index has already replaced. A single-term query is scaled uniformly and keeps its
-/// order, but a multi-term query weights its terms by their individual IDFs, so the
-/// ranking itself can differ from the one the live rows imply.
+/// Tantivy BM25 collection size statistics includes these superseded documents. A merge
+/// is the only mechanism to expunge documents. The default, [`LogMergePolicy`], does
+/// not merge on deletions, only when the ratio of deleted to collected documents is above a threshold.
 ///
-/// A merge is the only thing that expunges those documents, and tantivy's default
-/// [`LogMergePolicy`] never merges on deletions alone: `del_docs_ratio_before_merge`
-/// defaults to `1.0`, a ratio no segment can exceed. Its own documentation calls that
-/// default "not a very sensible default". Left at `1.0`, a segment that has grown large
-/// enough to sit alone at its size level — where the policy's segment-count rule can
-/// never apply either — keeps its superseded documents forever. Because `update_index`
-/// writes every incoming primary key as a deletion followed by an insertion, a
-/// repeatedly refreshed index reaches exactly that state as a matter of course.
-///
-/// Capping the ratio bounds the over-count instead: once merges settle, the collection
-/// size exceeds the live document count by at most `1 / (1 - RATIO)`.
+/// [`MAX_SUPERSEDED_DOCS_RATIO_PER_SEGMENT`] is the ratio to use in [`LogMergePolicy`]/
 const MAX_SUPERSEDED_DOCS_RATIO_PER_SEGMENT: f32 = 0.25;
 
 /// The merge policy for the index writer, which differs from tantivy's default only in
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

Jeadie (approved):

Issue #12058, opened 2026-07-26
E2E Test CI: Dremio quickstart and the model jobs need a reliability pass (~15% of runs fail)


claudespice:
Follow-up to #12057, which covered how these failures are *reported*. This issue tracks the failure rate itself, which is unchanged.

## What remains

`E2E Test CI` failed **6 of 40 runs** in a ~12h window on 2026-07-26, with a different sub-job each time, and on plain `push` runs against `trunk` where there is no PR diff. The recurring offenders are `Test Dremio quickstart` (external Dremio service) and the model jobs (`openai model`, `huggingface model` — model download plus a live provider). The run table is in #12057.

Because the workflow runs on `merge_group`, each failure can strand a merge-queue entry that has nothing to do with the failing job — on 2026-07-26 an approved, fully-green PR sat at queue position 1 in state `UNMERGEABLE` for over an hour behind one of these.

## Why this is worth waiting for one more occurrence before changing anything

The expect scripts previously could not say *why* a REPL went away: an `expect` block with no `eof` branch treated the process exiting as a match, so a crash was reported as a passing step, and the process only surfaced at the next `send` as `spawn id expN not open`. #12057 fixed that — a failure now reports the child's exit status, or the signal that killed it. The next occurrence in one of these jobs should therefore carry an actual diagnosis (a non-zero status, or `CHILDKILLED SIGKILL` for an out-of-memory kill), which is what a reliability fix should be built on.

Worth separating when it does:

- **Model jobs** — is the REPL crashing, being OOM-killed, or is the provider call failing? The exit status distinguishes these.
- **`Test Dremio quickstart`** — depends on an external Dremio service, so its failures may be environmental rather than a defect in Spice. It runs only on `workflow_dispatch` and `push`, so it does not gate the merge queue; the model jobs do.

## Not the fix

Retrying the failing step. It would restore the pre-#12057 situation where the cause is invisible, in a gate that strands merge-queue entries when it flakes.

grokspice:
Picked this up. The failure rate has a concrete cause, and it is not the sub-job that fails.

[...]

PR #12059, opened 2026-07-26
fix(e2e): report a REPL that exits mid-script instead of passing the step (fixes #12057)

### LINKED ISSUE #12057: E2E Test CI flakes on a rotating sub-job (~15% of runs), stranding merge-queue entries
## Summary

`E2E Test CI` (`.github/workflows/e2e_test_ci.yml`) fails intermittently — **6 of the last 40 runs (~15%)** in a recent ~12h window — with a **different sub-job failing each time**. Because the workflow runs on `merge_group`, a failure can leave a merge-queue entry stranded: an approved, fully-green PR sat at queue position 1 in state `UNMERGEABLE` for over an hour after its batch hit one of these failures, blocking the entry behind it.

The failures are not attributable to the PRs under test — the same workflow also fails on plain `push` runs against `trunk`, where there is no PR diff at all.

## Evidence

Failing runs in the window, with the sub-job that failed:

| Run | Trigger | Failing sub-job |
|---|---|---|
| `30186823353` | `merge_group` | `openai model (darwin-aarch64)` |
| `30184489281` | `merge_group` | `huggingface model (darwin-aarch64)` |
| `30184477721` | `merge_group` | `huggingface model (darwin-aarch64)` |
| `30181878210` | `push` (trunk) | `Test Dremio quickstart (darwin-aarch64)` |
| `30175840413` | `merge_group` | `Test Databricks catalog connector (mode: delta_lake) on Linux x64` |
| `30175840079` | `push` (trunk) | `Test Dremio quickstart (linux-x86_64)` |

The rotating sub-job and the failures on `trunk` pushes both point at harness/environment instability rather than a defect in any one code path.

## One failure mode, in detail

In `30186823353`, the `openai model (darwin-aarch64)` job failed inside the `expect` script `test/models/chat_01.exp`. The first two prompts succeeded (`Model confirmed access to all datasets`, `Model returned expected response`), then the spawned `spice chat` process was gone by the third:

```
send: spawn id exp5 not open
    while executing
"send "find top 3 exclusive deals in catalog pages. Respond with a short bullet list.\r""
    (procedure "verify_catalog_page" line 7)
    (file "./test/models/chat_01.exp" line 114)
```

So the REPL process exited between prompt 2 and prompt 3 and the script kept writing to a closed spawn id. Two things are worth separating here:

1. **Why the process exited.** The script does not capture the child's exit status or any stderr it emitted, so the log shows only the downstream `send` error. Whatever actually killed `spice chat` is not recorded anywhere in the run.
2. **How the script reports it.** `send` to a dead spawn id produces a confusing error that describes the symptom rather than the cause, which is what makes these runs expensive to triage.

## Suggested direction

- Have the `expect` scripts detect `eof` on the spawned process and fail with the child's exit status plus its captured output, so a crash is reported as a crash instead of as `spawn id ... not open`.
- Consider capturing the runtime's log as a job artifact for the model E2E jobs, so a mid-script exit is diagnosable after the fact.
- Separately, the `Test Dremio quickstart` and model-download jobs are the recurring offenders across the window and may warrant their own reliability pass.

Please don't fix this by retrying the failing step blindly — the underlying process exit would stay invisible, and this gate runs in the merge queue where a silent flake strands entries.

## Impact

Because `E2E Test CI` runs on `merge_group`, each of these intermittent failures can drop or strand a merge-queue entry that has nothing to do with the failing job, costing a full requeue (and, on this repo, the multi-hour sign-off that precedes it).

claudespice:
## Summary

`E2E Test CI` drives `spice chat` and `spice search` over a pty with `expect`. When one of those REPLs went away mid-script, the failure was reported as `send: spawn id exp5 not open` — naming a line in the script rather than the exit status of the process that died, which was recorded nowhere.

Tracing the reported run showed the reporting gap is the smaller half of the problem. An `expect` block with **no `eof` pattern** ends *without running any body* when the spawned process goes away, and the script then carries on as if the pattern had matched:

- In [run 30186823353](https://github.com/spiceai/spiceai/actions/runs/30186823353), `spice chat` exited immediately after receiving the second prompt. `verify_num_issues` matched the resulting eof 90 ms after sending (`05:14:31.851` → `05:14:31.944`) and printed **`Model returned expected response`**. The run's "first two prompts succeeded" reading is wrong: prompt 2 was never answered. The dead process only surfaced at the *next* `send`, because the eof had closed the spawn id by then.
- A REPL that exits while sitting **idle at its prompt** was invisible: writing to the pty of an exited process succeeds silently, so `send \x03` returned success and the script exited **0**. Reduced to the pre-fix shape, a stand-in REPL that dies without answering produces:

  ```
  chat> How many issues stored in database?
  Model returned expected response
  PRE-FIX EXIT=0
  ```

So a crashed REPL could be reported as a passing step in a gate that runs on `merge_group`.

Fixes #12057

## Changes

- **`test/models/repl_helpers.exp`** (new) — shared helpers: `repl_died` reports how the process terminated (its exit status, or `CHILDKILLED` plus the signal name — an OOM kill looks like this) after draining whatever it printed on the way out; `repl_send` reports a process that has already gone away instead of raising `spawn id expN not open`; `repl_assert_running` reads for a pending eof, which is the only way to notice a REPL that died while idle.
- **`chat_01.exp`, `chat_01_simple.exp`, `search_01.exp`** — every `expect` block gained an `eof` branch, every `send` goes through `repl_send`, and each script checks the REPL is still running before sending Ctrl+C. The two bare prompt waits (`expect "chat>"`, `expect "search>"`) gained `eof` *and* `timeout` branches; previously a timeout there let the script continue in an unknown state, matching what `chat_01_simple.exp` already did.
- **`test/models/expect_test.sh`** (new) — drives stand-in processes instead of `spice`, so every crash path runs with no runtime, model provider, or network. Wired into both model jobs in `e2e_test_ci.yml`, next to the scripts it tests.

Deliberately **not** changed:

- No retry around the failing step. The issue asks for the opposite, and a retry would keep the process exit invisible.
- No new artifact upload for the runtime log: the existing `Stop spice and check logs` step already runs `if: always()` and `cat spice.log`, and `spice chat`'s own output reaches the job log through the pty.
- The **flake rate itself** is untouched — this makes each failure diagnosable and unable to masquerade as a pass. The recurring offenders (`Test Dremio quickstart`, the model-download jobs) need their own reliability pass, tracked separately in #12058.

## Test plan

- [x] Added regression test for a REPL that exits without answering, one that exits while idle at its prompt, one that has already gone when the script sends, and one killed by a signal — at the helper level and through `chat_01.exp` / `chat_01_simple.exp` / `search_01.exp` themselves
- [x] Added edge case tests for a healthy exchange through all three scripts (the helpers must not turn a working interaction into a failure), and for the signal-kill path reporting `SIGKILL`
- [x] Verified by neutering: with the helpers reduced to their pre-fix behaviour, 14 checks fail across 7 cases — including `chat_01.exp` exiting **0** on a REPL that died while idle — while all three healthy cases still pass
- [x] `test/models/expect_test.sh` passes (40 checks), 3 consecutive runs, under both `/usr/bin/env bash` and macOS `/bin/bash` 3.2
- [x] `make lint` passes (workspace-wide `cargo fmt --all -- --check` + full clippy — not a per-crate `cargo clippy -p <crate>`)
- [x] `make test` passes

## Checklist

- [x] No behaviour change for a healthy run: the same prompts, patterns and success messages, plus a one-second liveness check before the REPL is asked to exit
- [x] No Rust, config, or runtime code touched — CI test scripts only
- [x] The new CI step is deterministic: no network, no model provider, no `spice` binary

### DIFF
--- a/.github/workflows/e2e_test_ci.yml
+++ b/.github/workflows/e2e_test_ci.yml
@@ -258,6 +258,10 @@ jobs:
             HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_ENV_HINTS=1 HOMEBREW_NO_INSTALL_CLEANUP=1 brew install expect
           fi
 
+      - name: Test the expect scripts driving the REPLs
+        run: |
+          ./test/models/expect_test.sh
+
       - name: Test vector search
         run: |
           ./test/models/search_01.exp
@@ -343,6 +347,10 @@ jobs:
           else
             HOMEBREW_NO_AUTO_UPDATE=1 HOMEBREW_NO_ENV_HINTS=1 HOMEBREW_NO_INSTALL_CLEANUP=1 brew install expect
           fi
+
+      - name: Test the expect scripts driving the REPLs
+        run: |
+          ./test/models/expect_test.sh
 
       - name: Test vector search
         run: |
--- a/test/models/chat_01.exp
+++ b/test/models/chat_01.exp
@@ -1,5 +1,7 @@
 #!/usr/bin/expect -f
 
+source [file join [file dirname [info script]] repl_helpers.exp]
+
 # Runtime response timeout
 set timeout 30
 
@@ -12,7 +14,7 @@ proc verify_datasets_access {} {
     set gh_issues_available 0
     set catalog_page_available 0
 
-    send "What datasets you have access to\r"
+    repl_send "What datasets you have access to\r"
 
     expect {
         -re "(?i)taxi_trips|taxi trips" {
@@ -36,6 +38,9 @@ proc verify_datasets_access {} {
         "chat> " {
             # end of model response
         }
+        eof {
+            repl_died "Waiting for the list of datasets"
+        }
         timeout {
             send_user "Timeout waiting for expected response\n"
             exit 1
@@ -53,7 +58,7 @@ proc verify_datasets_access {} {
 proc verify_num_issues {} {
     global timeout
 
-    send "How many issues stored in database?\r"
+    repl_send "How many issues stored in database?\r"
 
     expect {
         " ERROR " {
@@ -64,6 +69,9 @@ proc verify_num_issues {} {
         }
         "chat> " {
             # end of model response
+        }
+        eof {
+            repl_died "Waiting for the issue count"
         }
         timeout {
             send_user "Timeout waiting for expected response\n"
@@ -80,7 +88,7 @@ proc verify_catalog_page {} {
     set original_timeout $timeout
     set timeout 60
 
-    send "find top 3 exclusive deals in catalog pages. Respond with a short bullet list.\r"
+    repl_send "find top 3 exclusive deals in catalog pages. Respond with a short bullet list.\r"
 
     expect {
         " ERROR " {
@@ -91,6 +99,9 @@ proc verify_catalog_page {} {
         }
         "chat> " {
             # end of model response
+        }
+        eof {
+            repl_died "Waiting for the catalog page deals"
         }
         timeout {
             send_user "Timeout waiting for expected response\n"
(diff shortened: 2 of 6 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12060, opened 2026-07-26
fix(imap): fetch the raw message only when a scan reads content (fixes #12045)

### LINKED ISSUE #12045: IMAP connector always downloads full message bodies, even when the `content` column is not read
## Description

The IMAP table provider fetches every message with a single fixed attribute list:

```rust
const FETCH_QUERY: &str = "(ENVELOPE BODY.PEEK[HEADER] BODY.PEEK[])";
```

`BODY.PEEK[]` is the entire message — every MIME part, including attachments. It is requested unconditionally, but the body is only *used* when the provider was configured with `fetch_content`:

```rust
let body = if self.fetch_content {
    message.body().as_ref().map(|v| extract_text_body(v))
} else {
    None
};
```

So a table without a `content` column downloads every byte of every message and then discards all of it. For a mailbox carrying attachments this can dominate scan time and bandwidth by orders of magnitude over the envelope and header data actually needed.

## Expected behavior

Request `BODY.PEEK[]` only when the body will be read, i.e. build the fetch attribute list from `fetch_content` rather than using one constant for both cases.

## Notes

Found during review of #12039, which narrowed *which messages* a scan fetches (via `UID SEARCH`) but did not change *which attributes* are fetched per message. The two are complementary: this issue is about per-message payload size, that PR was about message count. Fixing this is a behavior change to the wire protocol path and warrants its own change with its own test, rather than being folded into that PR.

claudespice:
## Summary

Every IMAP scan asked the server for the same attribute list:

```
(ENVELOPE BODY.PEEK[HEADER] BODY.PEEK[])
```

`BODY.PEEK[]` is the *entire* raw message — every MIME part, attachments included — and it is read only under `if self.fetch_content`, to populate the `content` column. A table declared without that column therefore transferred every byte of every message and discarded it, which dominates the cost of a scan against a mailbox holding attachments.

`BODY.PEEK[HEADER]` was never read at all, in either configuration. Nothing in the connector calls `Fetch::header()`: the nine header columns are built from `ENVELOPE`, and `content` comes from the raw message, which carries its own headers.

Fixes #12045

## Changes

`FETCH_QUERY` becomes `fetch_query(fetch_content: bool)`, wired at both the `uid_fetch` and `fetch` call sites:

| `content` column | attributes requested |
|---|---|
| present | `(ENVELOPE BODY.PEEK[])` |
| absent | `(ENVELOPE)` |

Dropping the header section is safe in both cases: in the `imap` crate, `body()` matches `BodySection { section: None }` / `Rfc822` while `header()` matches `BodySection { section: Some(Full(Header)) }` — distinct attribute values, so removing the header section cannot change what `body()` or `envelope()` return.

Every body section stays `.PEEK`, so a scan still never sets `\Seen`.

## Test plan

- [x] Added regression test for a table with no `content` column not asking for the raw message (`fetch_query_asks_for_the_raw_message_only_to_fill_content`)
- [x] Added edge case tests for both configurations: the attribute list agrees with the columns the schema exposes, no attribute is requested that no column reads, and the list stays a parenthesized all-`PEEK` list so a scan cannot flag messages as read
- [x] Verified by neutering: restoring the single constant fails 3 of the 4 new tests. The fourth (`fetch_query_is_a_valid_peeking_attribute_list`) still passes on the old string, which was also parenthesized and peeked — it is a guardrail against a future edit, not a regression test
- [x] `make lint` passes (workspace-wide `cargo fmt --all -- --check` + full clippy — not a per-crate `cargo clippy -p <crate>`)
- [x] `make test` passes

## Performance impact

Strictly fewer bytes fetched, and no extra round trips — the same single `FETCH` per scan, with a shorter attribute list. A table with a `content` column stops fetching the redundant header section; a table without one stops fetching message bodies entirely. No change to which messages are selected, so filter and limit pushdown behave exactly as before.

## Checklist

- [x] Same number of `FETCH` commands; only the requested attributes change
- [x] `.PEEK` preserved on every body section, so a scan does not mutate the mailbox
- [x] Header columns unaffected — they are built from `ENVELOPE`, which is always requested

### DIFF
--- a/crates/data-connectors/connector-imap/src/imap/provider.rs
+++ b/crates/data-connectors/connector-imap/src/imap/provider.rs
@@ -43,12 +43,28 @@ use super::{
 /// above the scan either way.
 const MAX_ID_SET_LEN: usize = 8 * 1024;
 
-/// The message attributes every scan asks for: the envelope backing the header
-/// columns, and the raw message backing `content`.
+/// The message attributes a scan asks the server for, given whether the table
+/// has a `content` column.
 ///
-/// The raw message is requested for every scan, including one whose table has no
-/// `content` column and therefore discards it — see #12045.
-const FETCH_QUERY: &str = "(ENVELOPE BODY.PEEK[HEADER] BODY.PEEK[])";
+/// `ENVELOPE` backs the nine header columns and is always needed. `BODY.PEEK[]`
+/// is the *entire* raw message — every MIME part, attachments included — and is
+/// read only to populate `content`, so a table without that column does not ask
+/// for it: transferring a mailbox's attachments only to discard them dominates
+/// scan cost (#12045).
+///
+/// `BODY.PEEK[HEADER]` is asked for in neither case. No column is built from
+/// `Fetch::header()`; the header columns come from `ENVELOPE`, and `content`
+/// comes from the raw message, which carries its own headers.
+///
+/// `.PEEK` keeps the server from setting `\Seen`, so a scan never mutates the
+/// mailbox.
+fn fetch_query(fetch_content: bool) -> &'static str {
+    if fetch_content {
+        "(ENVELOPE BODY.PEEK[])"
+    } else {
+        "(ENVELOPE)"
+    }
+}
 
 /// Which messages a scan asks the server for.
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12063, opened 2026-07-26
fix(search): reach a compound's inner full-text tier when a change stream attaches (fixes #12061)

### LINKED ISSUE #12061: Bug(search): a CDC-fed warm full-text index inside a CompoundSearchIndex still defers commits, so a failed refresh can silently discard change-stream documents
A change-data-capture stream drives `Index::compute_index` outside the `TableSink` write lifecycle, so a full-text index that is also CDC-fed must never open a deferred-commit window: one `tantivy::IndexWriter` stages every pending operation together, so a window rollback discards the CDC documents staged alongside the refresh.

`FullTextDatabaseIndex` has exactly that protection — `cdc_attached` makes `on_write_start` return early — and `FullTextConnector` sets it when it builds the change stream:

```rust
// crates/runtime/src/search/full_text/connector.rs
for index in &all_indexes {
    if let Some(full_text) = index.as_any().downcast_ref::<FullTextDatabaseIndex>() {
        full_text.mark_cdc_attached();
    }
}
```

The downcast only matches an index registered *as* a `FullTextDatabaseIndex`. For a dataset whose full-text search is served by the write-through warm tier, the index registered on the `IndexedTableProvider` is a `CompoundSearchIndex` (warm Tantivy primary + Elasticsearch secondary) and the `FullTextDatabaseIndex` is its `primary` — `add_compound_fts_to_table` registers the compound *instead of* the concrete indexes, and `IndexedTableProvider::get_all_indexes()` returns the registered list flat, with no peeling. The downcast fails, `mark_cdc_attached()` is never called, and the warm index keeps deferring.

### Impact

`CompoundSearchIndex` does forward the write lifecycle (`compound_on_write_start` → `primary.on_write_start()`), so the warm index opens a deferred-commit window for a sink-driven refresh while the change stream writes into the very same writer via `compound_write` → `FullTextDatabaseIndex::write` → `update_index`, which observes `defer_commit` and stages instead of committing.

If that refresh then fails, `on_write_failed` rolls the writer back and the change-stream documents staged in the window are discarded. CDC does not resend them, so the warm tier is permanently missing those rows. With the default `on_zero_results: return_empty` the compound reads `PrimaryOnly`, so full-text searches serve the warm tier alone and those rows are silently absent from every result until a later refresh happens to rewrite them.

### Reachability

- `refresh_mode: changes`, which is what makes `FullTextConnector::changes_stream` run.
- A single full-text search column plus an Elasticsearch store, so `add_compound_fts_to_table` registers a compound rather than the Elasticsearch index alone.
- The `elasticsearch` feature, which is part of the release feature set.

### Proposed fix

The un-peeled downcast is the whole bug class, not just this instance: the same shape reappears for any wrapping index. Rather than teaching the connector to peel one wrapper type, add the CDC mark to the `Index` trait as a lifecycle hook with a no-op default and forward it from every wrapper (`CompoundSearchIndex`, `CompoundVectorIndex`, `ChunkedVectorIndex`), then have `FullTextConnector` call it on every index instead of downcasting.

### Related, deliberately not in scope

`EmbeddingConnector::changes_stream` filters full-text indexes out of its change stream with the same `downcast_ref::<FullTextDatabaseIndex>()`, so a compound is not filtered there either. That filter changes *which* indexes a stream writes rather than whether a commit is deferred, so it needs its own reasoning about double-indexing and is left alone here.

claudespice:
Fixes #12061

## Summary

A full-text index that is also fed by a change-data-capture stream must not open a deferred-commit window: it shares one `tantivy::IndexWriter` with the sink write path, so a window commit publishes a partial refresh and a window rollback discards the change stream's documents. `FullTextDatabaseIndex` guards against that with a `cdc_attached` flag, and `FullTextConnector` sets it by downcasting each registered index to `FullTextDatabaseIndex`.

That downcast only matches an index registered *as* a `FullTextDatabaseIndex`. When full-text search is served by the write-through warm tier, the registered index is a `CompoundSearchIndex` (warm Tantivy primary + Elasticsearch secondary) and the `FullTextDatabaseIndex` is its `primary` — `add_compound_fts_to_table` registers the compound in place of its tiers, and `IndexedTableProvider::get_all_indexes()` returns the registered list flat, with no peeling. So the flag was never set, the warm tier kept deferring, and a refresh that failed while the change stream was writing rolled those change-stream documents back for good. CDC does not resend them, and with the default `on_zero_results: return_empty` the compound reads `PrimaryOnly`, so the affected rows were silently missing from every full-text search until some later refresh happened to rewrite them.

The un-peeled downcast is the bug, not just this one instance of it, so the notification moves onto the `Index` trait and every composing index forwards it. `CompoundVectorIndex`, `ChunkedSearchIndex` and `ChunkedVectorIndex` compose indexes the same way and already forward the three write-lifecycle hooks; they now forward this one too.

## Changes

- `Index::on_cdc_attached`: a no-op-default hook called when a change stream that writes to the index is constructed. Its doc comment states the forwarding requirement for composing indexes.
- `FullTextDatabaseIndex` implements it (replacing the inherent `mark_cdc_attached`), and `CompoundSearchIndex`, `CompoundVectorIndex`, `ChunkedSearchIndex` and `ChunkedVectorIndex` forward it to the indexes they compose.
- `FullTextConnector::changes_stream` calls it on every registered index instead of downcasting to a concrete type.
- `EmbeddingConnector::changes_stream` calls it on the indexes it puts in its own change stream, so the notification is driven by which indexes a stream writes to rather than by which connector built it. That list is unchanged — only the notification is new — and no vector index acts on the hook today.

## Test plan

- `make lint`
- `make test`
- New: `test_cdc_attached_compound_primary_never_defers_commits` — a change-stream document written through a compound whose primary is a real warm full-text tier survives a failed write window. Verified to fail without the fix (the document is staged in the window and discarded by `on_write_failed`).
- New: `cdc_attach_forwards_to_both_vector_tiers`, and `lifecycle_hooks_forward_to_both_indexes` extended to cover the new hook — both verified to fail with the forwarding removed.

Note on a pre-existing flake seen while running these: `test_updates_overwrites_on_compute_index[_composite_pk]` fails intermittently on `trunk` (roughly 1 run in 4 of `cargo test -p search --lib`, confirmed with these changes stashed) because BM25 statistics still count superseded documents — that is #12053, fixed by #12056, and unrelated to this change.

## Checklist

- [x] Every composing `Index` implementation forwards the new hook
- [x] No behavior change for a dataset without a change stream


claudespice:
@copilot review

Jeadie (dismissed):

Jeadie (approved):

PR #12071, opened 2026-07-26
fix(accelerator): report PK equality pushdown Inexact so a point lookup can't return the whole table (fixes #12070)

### LINKED ISSUE #12070: Arrow accelerator: a concurrent write between pushdown and scan makes a PK point lookup return the whole table
Found during a data-correctness audit of trunk (`2bfa38823`). This is a residual gap in the #11262 fix (#11532), not a regression of it.

## Problem

`IndexedMemTable` (auto-enabled for `engine: arrow` datasets with `primary_key`/`indexes`) decides whether its hash index can be trusted by reading a `dirty` flag — but it reads that flag **twice, at two different planning stages**, and the two reads are not required to agree:

1. **Logical optimization** — `supports_filters_pushdown` (`crates/data_components/src/arrow/indexed.rs:635`) reads `self.is_dirty()`. Clean ⇒ the PK equality is reported **`Exact`**, so DataFusion **deletes the predicate from the plan** (no `FilterExec` is kept).
2. **Physical planning** — `scan` (`indexed.rs:689`) reads `self.is_dirty()` *again*. Dirty ⇒ `index_usable == false`, both index branches are skipped, and it falls through to `self.inner.scan(...)` (`indexed.rs:825`) — a plain `MemTable` scan, which **does not apply filters**.

If any write marks the table dirty *between* those two reads, the predicate has already been dropped and nothing re-applies it. The point lookup returns **every row in the table**.

`mark_dirty()` is called at plan time by `insert_into` / `delete_from` / `update` / `truncate` (`indexed.rs:839`, `859`, `899`, `905`), and by the retention evictor and the accelerated-table refresh sink. So for the standard `engine: arrow` + `primary_key` + `on_conflict: upsert` dataset — where refresh/DML writes are continuous — this races against every concurrent point-lookup query being planned.

## Concrete wrong result

```sql
-- accelerated table holds ids 1, 2, 3
SELECT * FROM ds WHERE id = 1;
```

returns

```
+----+---------+
| id | name    |
+----+---------+
| 1  | alice   |
| 2  | bob     |     <- should not be here
| 3  | charlie |     <- should not be here
+----+---------+
```

Silent: no error, no warning. The result set is unbounded in the table's size, so it is also a latency/memory event on a large accelerator.

## Reproduced on trunk `2bfa38823`

Unit test in `crates/data_components/src/arrow/indexed.rs` driving the two stages in the order DataFusion drives them, with the write interleaved between:

```rust
let table = create_test_indexed_table_force_index(
    Arc::clone(&schema), vec![vec![batch]], vec!["id".to_string()],
).expect("failed to create table");

// 1. Logical optimization: index clean -> Exact -> DataFusion drops the predicate.
let pk_filter = col("id").eq(lit(1_i64));
let pushdown = table.supports_filters_pushdown(&[&pk_filter]).expect("pushdown");
assert_eq!(pushdown, vec![TableProviderFilterPushDown::Exact]);

// 2. A concurrent DML/retention write marks the index dirty.
table.mark_dirty();

// 3. Physical planning: index unusable -> unfiltered MemTable scan.
let plan = table.scan(&ctx.state(), None, std::slice::from_ref(&pk_filter), None).await.expect("scan");
let batches = datafusion::physical_plan::collect(plan, ctx.task_ctx()).await.expect("collect");
let rows: usize = batches.iter().map(RecordBatch::num_rows).sum();
assert_eq!(rows, 1, "point lookup must return 1 row, not the whole table");
```

Output:

```
PUSHDOWN AT PLAN TIME = [Exact]
DIRTY-WINDOW ROWS RETURNED = 3
assertion `left == right` failed: point lookup must return 1 row, not the whole table
  left: 3
 right: 1
```

## Root cause

`Exact` is a promise that the scan applied the predicate. `IndexedMemTable` makes that promise based on state it re-reads later and may act differently on. Any design that keeps `Exact` has to keep two independent trait calls consistent across a window it does not control.

## Suggested fix

**Report PK equality as `Inexact`, unconditionally**, and drop the `is_dirty()` consultation from `supports_filters_pushdown`.

`Inexact` still passes the filter to `scan`, so the `IndexedLookupExec` point-lookup fast path is fully preserved — DataFusion just also keeps a `FilterExec`, which for a 1-row lookup result costs essentially nothing. Correctness then holds for *every* path through `scan` regardless of when the flag flips: index hit, index miss, hash-collision miss, and the dirty full-scan fallback.

This also removes the last `Exact` from the provider (the secondary-index path already reports `Inexact`), so the whole "dropped predicate the scan didn't apply" class is gone rather than narrowed. It matches the repo guidance of defaulting to the universally-safe mode instead of gating correctness on a capability check.

`test_explain_indexed_scan_snapshot` will need its snapshot regenerated (a `FilterExec` now appears above `IndexedLookupExec`).

## Checked and NOT a bug

Two conflicting PK equality conjuncts (`WHERE id = 1 AND id = 2`) would also defeat the `Exact` promise, since `supports_filters_pushdown` marks *every* PK equality `Exact` while `extract_pk_equality_value` uses only the first. It is **not reachable**: DataFusion's simplifier folds the contradiction to `EmptyExec` before pushdown (verified — plan is `EmptyExec`, 0 rows). Noting it so the fix above is understood to cover it anyway.

claudespice:
Fixes #12070

## Summary (root cause)

`IndexedMemTable` (auto-enabled for `engine: arrow` datasets with `primary_key`/`indexes`) decided whether its hash index could be trusted by reading a `dirty` flag — but it read that flag **twice, at two different planning stages**, and nothing required the two reads to agree:

1. **Logical optimization** — `supports_filters_pushdown` read `is_dirty()`. Clean ⇒ the PK equality was reported **`Exact`**, which makes DataFusion **delete the predicate from the plan**.
2. **Physical planning** — `scan` read `is_dirty()` again. Dirty ⇒ both index branches were skipped and it fell through to `self.inner.scan(...)`, a plain `MemTable` scan, which does not apply filters.

A write landing between those two reads — DML, retention eviction, or a refresh sink, all of which call `mark_dirty()` at plan time — left the predicate deleted with nothing to re-apply it. The point lookup then returned **every row in the table**, silently: no error, no warning, and a result set unbounded in the accelerator's size.

For the standard `engine: arrow` + `primary_key` + `on_conflict: upsert` dataset, where refresh/DML writes are continuous, this races against every concurrent point-lookup query being planned.

This is a residual gap in #11532 (the #11262 fix), which made the two sites individually dirty-aware but not jointly consistent.

## Changes

`crates/data_components/src/arrow/indexed.rs` — report PK equality pushdown as **`Inexact`, unconditionally**, and drop the `is_dirty()` consultation from `supports_filters_pushdown` (the dirty flag is still consulted in `scan`, which is where it belongs).

`Exact` is a promise that the scan applied the predicate, and this provider cannot keep it across two independent trait calls. `Inexact` still hands the filter to `scan`, so the `IndexedLookupExec` point-lookup fast path is **unchanged** — DataFusion additionally keeps a `FilterExec`, which over a point-lookup result costs essentially nothing, and which makes every path through `scan` correct: index hit, index miss, hash-collision miss, and the full-scan fallback alike.

The secondary-index path already reported `Inexact`, so the provider now never reports `Exact` — the "dropped predicate the scan didn't apply" class is eliminated rather than narrowed. This follows the repo guidance of defaulting to the universally-safe mode instead of gating correctness on a capability check.

## Performance impact

The point-lookup fast path is preserved: `EXPLAIN` still shows `IndexedLookupExec: indexed_scan on [id]`, now under a `FilterExec` that re-applies the equality over the single returned row. The full-table scan the index avoids is still avoided; the added cost is one predicate evaluation per lookup result row.

## Test plan

- `make lint`
- `make nextest`
- Targeted during development: `cargo test -p data_components --lib arrow::indexed` — **53 passed / 0 failed**.
- **New end-to-end regression test** `test_write_between_pushdown_and_scan_keeps_point_lookup_correct`: drives the two planning stages explicitly (`into_optimized_plan()` → `mark_dirty()` → `create_physical_plan()`), which is exactly the boundary the race straddles, and asserts the point lookup returns 1 row rather than the whole table.
- **New contract test** `test_pk_pushdown_is_always_inexact`: PK equality is `Inexact` whether the index is clean or dirty.
- **Neuter test**: restoring the old `Exact`-when-clean behavior fails **8** tests, including both new ones — so the regression guards genuinely bite. Before the fix the end-to-end test returns all 3 rows for `WHERE id = 1`; after, 1 row.
- Two `insta` snapshots regenerated (`explain_indexed_scan`, `explain_indexed_table_point_lookup`) — the only change is the expected `FilterExec: id@0 = 1` above the preserved `IndexedLookupExec`.

## Known follow-up (needs a testoperator dispatch, cannot be done locally)

Six `crates/test-framework` EXPLAIN snapshots will need regeneration, because their plans currently record `full_filters` (Exact) on the PK and will become `partial_filters` with a `FilterExec`:

```
file[parquet]-arrow-pk_hash_index_{customer,nation,orders,part,region,supplier}_pk_lookup_explain.snap
```

These are asserted by the testoperator scenario `scenario/sf1/accelerated/indexes/file[parquet]-arrow-pk_hash_index.yaml`, not by merge-queue `cargo test`, so they do not block this PR's checks — but that scenario will fail until they are regenerated against a live SF1 spiced (which I can't run here, and hand-editing `.snap` files is disallowed). The 11 `secondary_hash_index` snapshots are **unaffected** — they already record `partial_filters`, since that path always reported `Inexact`.

## Checklist

- [x] Root cause identified and reproduced on trunk before writing the fix
- [x] Regression test fails on old code, passes on new (neuter-verified)
- [x] Bug-class swept: every `supports_filters_pushdown` impl in the repo that can return `Exact` was checked; `MemoryVectorQueryTable` and `runtime-table-partition` genuinely apply their Exact filters, so this was the only unsound instance
- [x] `cargo fmt` clean; clippy pass 1 (`--lib`) clean
- [x] Snapshots regenerated via `INSTA_UPDATE`, reviewed, not hand-edited
- [ ] `make signoff` / `Attestation` green

### DIFF
--- a/crates/data_components/src/arrow/indexed.rs
+++ b/crates/data_components/src/arrow/indexed.rs
@@ -620,34 +620,26 @@ impl TableProvider for IndexedMemTable {
     ) -> Result<Vec<TableProviderFilterPushDown>> {
         let owned_filters: Vec<Expr> = filters.iter().map(|&e| e.clone()).collect();
 
-        // If we have a primary key index and the filter includes a PK equality,
-        // mark only the PK filter as Exact. Non-PK filters must be Inexact so
-        // DataFusion still applies them after the indexed lookup. Returning Exact
-        // for all filters would silently drop non-PK predicates (e.g.
-        // `WHERE id=10 AND version=101` would ignore `version=101` and return
-        // the row for id=10 regardless of its version).
+        // A primary-key equality lets `scan` serve the query from the hash index,
+        // but the pushdown is reported `Inexact` — never `Exact`.
+        //
+        // `Exact` is a promise that the scan applied the predicate, and this
+        // provider cannot keep it. Whether the index is usable is decided from the
+        // `dirty` flag, which is read here during logical optimization and read
+        // AGAIN in `scan` during physical planning. A write landing between those
+        // two reads (DML, retention eviction, or a refresh sink — all call
+        // `mark_dirty`) makes `scan` bypass the index and fall through to an
+        // unfiltered `MemTable` scan while the predicate has already been deleted
+        // from the plan, so the point lookup returns the whole table (#12070).
+        //
+        // `Inexact` still hands the filter to `scan`, so the `IndexedLookupExec`
+        // fast path is unchanged; DataFusion additionally keeps a `FilterExec`,
+        // which over a point-lookup result costs nothing and makes every path
+        // through `scan` correct — index hit, index miss, hash-collision miss, and
+        // the full-scan fallback alike. It also keeps non-PK predicates applied
+        // (`WHERE id=10 AND version=101` must not ignore `version=101`).
         if self.index.is_some() && self.extract_pk_equality_value(&owned_filters).is_some() {
-            // When the index is dirty (a DML/retention/synchronized write happened
-            // without a subsequent rebuild), it can no longer be trusted: report the
-            // PK filter as Inexact instead of Exact so DataFusion keeps a FilterExec
-            // and re-applies the predicate over the full-scan fallback in `scan`,
-            // rather than dropping the filter and trusting a stale index (#11262).
-            let pk_dirty = self.is_dirty();
-            let pk_column = &self.primary_key_columns[0];
-            return Ok(filters
-                .iter()
-                .map(|f| {
-                    if Self::extract_equality_value(f, pk_column).is_some() {
-                        if pk_dirty {
-                            TableProviderFilterPushDown::Inexact
-                        } else {
-                            TableProviderFilterPushDown::Exact
-                        }
-                    } else {
-                        TableProviderFilterPushDown::Inexact
-                    }
-                })
-                .collect());
+            return Ok(vec![TableProviderFilterPushDown::Inexact; filters.len()]);
         }
 
         // Check if a secondary index can handle this filter
@@ -1296,12 +1288,14 @@ mod tests {
         )
         .expect("failed to create table");
 
-        // Equality on primary key should get Exact pushdown
+        // Equality on primary key gets Inexact pushdown: the index serves the
+        // lookup, but `scan` may bypass it, so DataFusion must re-apply the
+        // predicate (#12070).
         let filter_eq = col("id").eq(lit(1_i64));
         let pushdown = table
             .supports_filters_pushdown(&[&filter_eq])
             .expect("pushdown check failed");
-        assert_eq!(pushdown, vec![TableProviderFilterPushDown::Exact]);
+        assert_eq!(pushdown, vec![TableProviderFilterPushDown::Inexact]);
 
         // Non-equality should get Unsupported (MemTable doesn't support filter pushdown)
         let filter_gt = col("id").gt(lit(1_i64));
@@ -2739,14 +2733,14 @@ mod tests {
 
         let table_with_secondary = table.with_secondary_index(secondary);
 
-        // PK query should return Exact
+        // PK query should return Inexact (#12070)
         let pk_filter = datafusion::prelude::col("id").eq(datafusion::prelude::lit(1_i64));
         let result = table_with_secondary
             .supports_filters_pushdown(&[&pk_filter])
             .expect("pushdown check");
         assert_eq!(
             result,
-            vec![datafusion::logical_expr::TableProviderFilterPushDown::Exact]
+            vec![datafusion::logical_expr::TableProviderFilterPushDown::Inexact]
         );
 
         // Secondary index query should return Inexact (for unique secondary indexes)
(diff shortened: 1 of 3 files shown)


claudespice:
@copilot review

claudespice:
The signoff failure was the two EXPLAIN snapshots, not the pushdown change: with the predicate no longer absorbed the plan carries a `RepartitionExec`, whose printed partition count is `target_partitions` — which a default `SessionContext` takes from the host core count. Recorded on a 10-core machine, they failed on the 14-core runner (`RoundRobinBatch(10)` vs `(14)`).

Fixed in c49079408 by pinning `target_partitions` for those two contexts and re-recording, so the plan text is host-independent. The `FilterExec` this PR is about is unchanged; only the partition count moved. Signoff re-dispatched.

lukekim (approved):

Issue #12008, opened 2026-07-23
Cayenne: filtered DELETE (per-key / range) leaves un-checkpointed mem-tier rows


sgrebnov:
Follow-up to #11987 (which fixed the delete-all / TRUNCATE path).

## Problem

A client-issued `DELETE FROM <cayenne_table> WHERE <predicate>` does **not** remove rows that are still resident in the un-checkpointed in-memory CDC mem-tier. Verified for both per-key and range filters:

| Delete | Result on 5 mem-tier rows `[1..=5]` |
|---|---|
| `WHERE id = 2` | rows still `[1,2,3,4,5]` (row survives) |
| `WHERE id > 3` | rows still `[1,2,3,4,5]` (rows survive) |
| `WHERE TRUE` (TRUNCATE) | `[]` ✓ (fixed in #11987) |

## Root cause

At scan time the merged durable∪mem `deletion_snapshot` is applied only to the **file / protected / cold** branches. The **mem-tier** branch (`visible_mem_tier_segments_unpruned`) is filtered *only* by each shard's own in-RAM tombstones (`mem_tier_deletion_maps`), never by the durable snapshot. The user `delete_from` path (`InlineAwareDeletionSink`) writes durable tombstones for file/inline rows but never lands an in-RAM mem-tier tombstone — so a matching row resident only in the mem-tier is never hidden.

## Scope

- **Affected:** tables with a populated mem-tier — `cdc_durability: memory` (the default CDC "changes" profile) and `mode: memory` — for rows not yet checkpointed.
- **Not affected:** `cdc_durability: file` tables (no mem-tier), already-checkpointed rows, TRUNCATE/delete-all (fixed in #11987).
- **CDC-originated per-key deletes are unaffected:** they flow through `write_cdc_delete_keys_in_memory`, which lands in-RAM mem-tier tombstones. Only a direct client `DELETE ... WHERE` reaches the gap.

## Decision: full support vs. fail-closed guard

[...]

Issue #11987, opened 2026-07-22
Cayenne: TRUNCATE / delete-all (DELETE WHERE TRUE) leaves un-checkpointed CDC mem-tier rows


sgrebnov:
A TRUNCATE (or any non-key-enumerable delete-all) against a Cayenne `refresh_mode: changes` table silently leaves behind rows that are still resident in the in-memory CDC mem-tier. This is a data-correctness bug: `SELECT count(*)` after TRUNCATE returns a non-zero count.

## Root cause

CDC TRUNCATE flows connector-agnostically as `op="t"` → `process_truncate` (accelerated_table/refresh_task/changes.rs) → `accelerator.delete_from(WHERE TRUE)`.

In Cayenne, `delete_from` with `lit(true)` on a key-based table routes to `InlineAwareDeletionSink` (provider/on_conflict.rs), which tombstones:
- file-backed rows (listing table + protected snapshots + cold tier), and
- catalog "inlined data",

but **not** the in-memory CDC `mem_tier` (ShardedMemTier). Scans union live mem-tier rows (`visible_mem_tier_batches`), so any recent CDC writes not yet checkpointed survive the delete-all. `WHERE TRUE` cannot enumerate keys, so - unlike a per-key DELETE, which tombstones across all tiers — no tombstone is produced for the mem-tier-only keys.

## Reproduction
`mysql_binlog_replication_end_to_end_cayenne` (ignored in #11986): after UPDATE + INSERT immediately followed by TRUNCATE, `count(*)` stalls at 2 instead of 0 (the two mem-tier-resident rows).

PR #12073, opened 2026-07-26
fix(cayenne): purge the mem-tier on delete-all for a table without a primary key (fixes #12072)

### LINKED ISSUE #12072: Cayenne: DELETE FROM (delete-all/TRUNCATE) removes no rows on a table without a primary key, so a mode: memory table can never be emptied
Follow-up to #11987 / #12009, which fixed delete-all only on the **key-based** delete arm.

## Problem

On a Cayenne table **without a primary key**, `DELETE FROM <table>` (delete-all / TRUNCATE, i.e. `WHERE TRUE`) reports **0 rows affected** and removes **nothing** from the in-memory tier. For `mode: memory` — where the mem-tier is the permanent, only store — that means `DELETE` cannot delete anything at all, silently.

Reproduced on trunk `2bfa38823` with a `mode: memory`, no-`primary_key` table, writing through the real memory-mode write path:

```
memory_mode, is_position_based=true
before delete: [1, 2, 3, 4, 5]
delete_from(WHERE TRUE) -> deleted=0
after delete:  [1, 2, 3, 4, 5]     <- every row survives
mem_tier_empty=false
```

The equivalent table **with** a primary key passes (that is the case #12009 fixed and tests).

No error is raised, so a client sees a successful `DELETE` with a 0 count and every subsequent `SELECT` still returns the rows.

## Root cause

`purge_mem_tier_all()` (added by #12009) is called from exactly one place — `InlineAwareDeletionSink`, the **key-based** arm:

```
crates/cayenne/src/provider/on_conflict.rs:530   if is_delete_all(&self.filters) { … purge_mem_tier_all() … }
```

But `CayenneTableProvider::delete_from` (`crates/cayenne/src/provider/table.rs`) has three arms, and two of them return **before** that sink is ever built:

1. `file_based_deletes_preferred(&filters)` → `delete_using_files` — retention-shaped `col < threshold` only, so never a delete-all. Not affected.
2. `pk_deletion_strategy.is_position_based()` → `delete_using_deletion_vectors` — **affected**.
3. otherwise → `InlineAwareDeletionSink` — has the purge.

`pk_deletion_strategy` is `PositionBased` **iff the table has no primary key** (`table.rs`, `has_primary_key` → `PkDeletionStrategy::PositionBased`); it is independent of the `cayenne_deletion_mode` parameter. So every PK-less table takes arm 2.

Arm 2 wraps the deletion-vector sink in `PkKeysetInvalidatingDeletionSink`, which never touches the mem-tier (`rg mem_tier crates/cayenne/src/provider/delete/sink.rs` → zero hits). Position deletes record `(file, file-local position)` pairs; mem-tier rows live in no file, so there is nothing for that sink to tombstone and the rows stay visible. The plan-time `checkpoint_inlined_data_if_present_for_delete()` in that arm only materializes the **inline** corpus (`cached_inlined_row_count`), which is a different structure from the mem-tier — and in memory mode a checkpoint is a no-op by design.

## Scope

- **Affected:** any table with a populated mem-tier and **no primary key** — `mode: memory` (the mem-tier is the permanent store, so *all* rows), and `cdc_durability: memory` for rows not yet checkpointed.
- **Not affected:** tables with a primary key (arm 3, fixed in #12009); retention deletes (arm 1 is never a tautology filter).
- **Distinct from #12008**, which is *filtered* (`WHERE id = 2` / `WHERE id > 3`) deletes leaving mem-tier rows, and which explicitly records delete-all as "✓ fixed in #11987". That is true only for PK-ful tables; this issue is the PK-less half of delete-all.

## Fix

Gate the same purge on the other wrapper. `PkKeysetInvalidatingDeletionSink` (used by both arm 1 and arm 2) gains the request `filters` and, when `is_delete_all(&filters)`, runs `purge_mem_tier_all()` under the table `write_lock` after the inner sink completes, folding the purged count into rows-affected — so the `deleted > 0` bookkeeping (PK-keyset clear, `invalidate_scan_file_statistics`) also observes the changed visible row count. `purge_mem_tier_all` already early-returns `Ok(0)` on an empty mem-tier, so ordinary file tables pay nothing, and arm 1's retention filter never satisfies `is_delete_all`, so retention behavior is unchanged.

claudespice:
## Summary

On a Cayenne table **without a primary key**, `DELETE FROM <table>` (delete-all / TRUNCATE) reported **0 rows affected** and removed **nothing** from the in-memory tier. For `mode: memory`, where the mem-tier is the permanent and only store, the table could never be emptied — the `DELETE` returned success with a 0 count and every subsequent `SELECT` still returned the rows.

Root cause: `purge_mem_tier_all()` (added by #12009 for #11987) had exactly one caller, `InlineAwareDeletionSink` — the **key-based** delete arm. `CayenneTableProvider::delete_from` has three arms, and `pk_deletion_strategy` is `PositionBased` **iff the table has no primary key**, so a PK-less table always returned early through `delete_using_deletion_vectors` and never built that sink. Position deletes address `(file, file-local position)` pairs and mem-tier rows live in no file, so there was nothing for the inner sink to tombstone. The plan-time `checkpoint_inlined_data_if_present_for_delete()` on that arm only materializes the *inline* corpus, which is a different structure — and in memory mode a checkpoint is a no-op by design.

Fixes #12072

## Changes

- `PkKeysetInvalidatingDeletionSink` (the wrapper used by both the position-based and file-based arms) now carries the request `filters` and, when `is_delete_all(&filters)`, runs `purge_mem_tier_all()` after the inner sink completes, folding the purged count into rows-affected.
- The purge acquires the table `write_lock` itself rather than nesting: the inner sink takes and releases that lock internally.
- It runs *before* the existing `deleted > 0` bookkeeping so `invalidate_scan_file_statistics()` still fires on a table whose rows are only in the mem-tier (inner count 0) but whose visible row count just changed.

Deliberately unchanged: `purge_mem_tier_all` already early-returns on an empty mem-tier, so ordinary file tables pay nothing; and the file-based arm's retention filter (`col < threshold`) never satisfies `is_delete_all`, so retention behavior is untouched. *Filtered* deletes leaving mem-tier rows is a separate gap tracked in #12008 and is not addressed here.

## Test plan

- `make lint`
- `make test`

New unit tests in `crates/cayenne/src/provider/table.rs`:

- `truncate_purges_mem_tier_rows_on_a_table_without_a_primary_key` — regression test: a `mode: memory`, no-`primary_key` table written through the real memory-mode write path, asserting the delete-all count is 5, the scan returns no rows, and the mem-tier is empty. Asserts `is_position_based()` as a precondition so the test cannot silently drift onto the already-fixed arm.
- `filtered_delete_does_not_purge_the_mem_tier_without_a_primary_key` — guards the other direction: a filtered delete through the same sink must not discard the whole tier.

Verification performed:

- `cargo test -p cayenne --lib` — 923 passed. The 34 `cayenne_catalog::tests` failures are a pre-existing local environment issue (those tests use hardcoded `/tmp/cayenne_test*` paths owned by another uid on this shared machine → `PermissionDenied`); they are untouched by this diff and pass in CI.
- **Neuter test**: with the new purge made unreachable, `truncate_purges_mem_tier_rows_on_a_table_without_a_primary_key` fails (`deleted` 0, expected 5) while the pre-existing `truncate_purges_uncheckpointed_mem_tier_rows` and the filtered-delete guard still pass — so the regression test genuinely bites and the fix is what makes it pass.
- Clippy with the CI flag sets, both passes (`--lib`, then `--tests` with `-Aclippy::expect_used`): zero warnings. `cargo fmt -p cayenne` clean.

## Checklist

- [x] Regression test that fails on the old code and passes on the new
- [x] Bug-class swept: all four `DeletionSink` impls checked; the two wrappers are `InlineAwareDeletionSink` (already purges) and `PkKeysetInvalidatingDeletionSink` (fixed here)
- [x] No behavior change for tables with a primary key, or for retention deletes
- [x] Lint and format clean

### DIFF
--- a/crates/cayenne/src/provider/on_conflict.rs
+++ b/crates/cayenne/src/provider/on_conflict.rs
@@ -399,6 +399,9 @@ pub(crate) fn is_delete_all(filters: &[Expr]) -> bool {
 pub(crate) struct PkKeysetInvalidatingDeletionSink {
     pub(crate) table: CayenneTableProvider,
     pub(crate) inner: Arc<dyn DeletionSink>,
+    /// The delete request's filters, needed to recognize a delete-all so the
+    /// mem-tier can be purged alongside the inner sink's file-side work.
+    pub(crate) filters: Vec<Expr>,
 }
 
 #[async_trait]
@@ -424,7 +427,34 @@ impl DeletionSink for PkKeysetInvalidatingDeletionSink {
         // below resets the flag and rebuilds exact; an upsert table keeps the
         // stale-superset keyset and stays degraded until its next rebuild.
         self.table.mark_pk_keyset_occ_degraded();
-        let deleted = self.inner.delete_from(context).await?;
+        let mut deleted = self.inner.delete_from(context).await?;
+
+        // Delete-all (TRUNCATE / `DELETE … WHERE TRUE`): the inner sink records
+        // `(file, file-local position)` deletes, and rows resident in the
+        // in-memory tier live in no file — so nothing tombstones them and they
+        // stay visible. A table with no primary key reaches this sink for every
+        // delete (`pk_deletion_strategy` is `PositionBased` exactly then), and in
+        // `mode: memory` the mem-tier is the permanent store, so without this the
+        // table can never be emptied. Mirrors `InlineAwareDeletionSink` below,
+        // which covers the key-based arm. Skipped for filtered deletes, whose
+        // predicate cannot be evaluated against the tier here.
+        //
+        // `purge_mem_tier_all` requires the table `write_lock`, which the inner
+        // sink takes and releases internally, so acquire it here rather than
+        // nesting. Purge before the `deleted > 0` bookkeeping below: on a table
+        // whose rows are *only* in the mem-tier the inner count is 0, and the
+        // cached scan statistics still need invalidating once the purge changes
+        // the visible row count.
+        if is_delete_all(&self.filters) {
+            let _guard = self.table.write_lock.lock().await;
+            let purged = self
+                .table
+                .purge_mem_tier_all()
+                .await
+                .map_err(|error| Box::new(error) as Box<dyn std::error::Error + Send + Sync>)?;
+            deleted = deleted.saturating_add(purged);
+        }
+
         if deleted > 0 {
             // Keyset clear-on-delete avoidance (cycle-4 incremental lever).
             //
--- a/crates/cayenne/src/provider/table.rs
+++ b/crates/cayenne/src/provider/table.rs
@@ -26276,6 +26276,7 @@ impl CayenneTableProvider {
             PkKeysetInvalidatingDeletionSink {
                 table: self.clone_for_write(),
                 inner: sink,
+                filters: filters.to_vec(),
             },
         ))))
     }
@@ -26297,6 +26298,7 @@ impl CayenneTableProvider {
             PkKeysetInvalidatingDeletionSink {
                 table: self.clone_for_write(),
                 inner: sink,
+                filters: filters.to_vec(),
             },
         ))))
     }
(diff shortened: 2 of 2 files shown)


claudespice:
@copilot review

sgrebnov (approved):

PR #12079, opened 2026-07-26
feat(datasets): mark non-accelerated datasets Error when their source is unavailable

krinart:
## Summary

- The dataset availability monitor now moves a **non-accelerated** dataset to `Error` status when its source becomes unreachable (surfaced via `GET /v1/datasets?status=true`) and restores `Ready` once it recovers. Previously an unreachable source was only recorded as a metric, so the dataset stayed `Ready`.
- Adds a per-dataset `check_availability_interval` (default `60s`) controlling how often the source is probed.

## Testing

- Unit + integration tests covering the `Ready → Error → Ready` cycle.
- Manually verified end-to-end against PostgreSQL in Docker.

### DIFF
--- a/crates/runtime-component/src/dataset/mod.rs
+++ b/crates/runtime-component/src/dataset/mod.rs
@@ -345,10 +345,6 @@ impl Display for CheckAvailability {
     }
 }
 
-/// Default cadence at which the availability monitor probes a non-accelerated
-/// dataset's source when `check_availability_interval` is not configured.
-pub const DEFAULT_CHECK_AVAILABILITY_INTERVAL: Duration = Duration::from_mins(1);
-
 /// Config-only core of a dataset — every declared field of a
 /// `runtime::component::dataset::Dataset` except the runtime handles
 /// (`app`/`runtime`). The runtime wrapper holds `Self` plus those handles and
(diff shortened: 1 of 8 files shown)


crates/runtime-component/src/dataset/mod.rs
```diff
@@ -345,6 +345,10 @@ impl Display for CheckAvailability {
     }
 }
 
+/// Default cadence at which the availability monitor probes a non-accelerated
+/// dataset's source when `check_availability_interval` is not configured.
+pub const DEFAULT_CHECK_AVAILABILITY_INTERVAL: Duration = Duration::from_mins(1);
```
phillipleblanc: This seems quite aggressive as a default, especially that this will be a behavior change from today.

phillipleblanc: I would suggest we don't have a default for this and explictly require opting in to this behavior.

crates/runtime/src/component/dataset/builder.rs
```diff
@@ -66,6 +66,7 @@ pub struct DatasetBuilder {
     pub vectors: Option<VectorStore>,
     pub full_text_search: Option<FtsStore>,
     pub check_availability: CheckAvailability,
+    pub check_availability_interval: Option<String>,
```
phillipleblanc: The runtime component should have the parsed `Duration` and the parsing should happen when this object is constructed. That is one of the reasons we have this separate from the spicepod representation (which is the raw representation).

krinart: Fixed

crates/runtime-component/src/dataset/mod.rs
krinart: Fixed, thanks

sgrebnov (approved):

PR #12082, opened 2026-07-27
feat(search): warm in-memory tier for chunked Elasticsearch vector columns

Jeadie:
## Problem

Chunked Elasticsearch-backed vector columns unconditionally skipped the in-memory warm search tier (`with_memory_warm_index`), so every search on them hit Elasticsearch directly — while non-chunked ES columns and S3 Vectors columns (chunked or not) all got the warm-tier latency benefit.

The skip existed because the warm index pairs a `MemoryVectorIndex` with the engine index in a `FallbackToSecondary` `CompoundVectorIndex`, whose fallback projection requires every column in the memory (primary) plan to exist by name in the engine (secondary) plan. The ES query/list plans exposed neither the chunk key (`_spice.chunk_id`) nor the offset metadata column (`{col}_offset`), so that plan could not be built and the warm index was disabled for chunked columns.

## Change

Make Elasticsearch mirror the proven S3 Vectors pattern: both the query and list plans now project the full (augmented, when chunked) primary key **plus** the metadata columns, so the warm index falls back onto Elasticsearch for both chunked and non-chunked columns. There is no longer an ES/S3 divergence or an `is_chunked` special case.

- **`runtime/.../embeddings/index/elasticsearch/mod.rs`** — `try_from_table` augments the primary key with `_spice.chunk_id`, carries it in `source_schema` (so `knn_hits_to_batch` extracts it from `_source`), and maps it as a non-indexed `long`. `normalize_es_data_type` no longer coerces integer `FixedSizeList`s to `Float32`.
- **`search/.../index/elasticsearch/mod.rs`** — `query_result_schema`/`list_result_schema` drop the chunk-key filter and the `is_chunked` special case and project metadata columns uniformly.
- **`runtime/.../embeddings/index/table.rs`** — both paths build the warm index and wrap it with `construct_chunked_vector_index` when chunked, matching S3.
- **`data_components/.../elasticsearch/query_table.rs`** — the `_source` reader now faithfully decodes the full set of types `arrow_type_to_es_mapping` can produce: unsigned int widths, `Date32`/`Date64`, all `Timestamp` `TimeUnit`s, and an integer `FixedSizeList` arm for the chunk offset pair (previously every `FixedSizeList` decoded as a Float32 dense vector). `extract_field` tries the literal (flat) key before nested dot-navigation, so `_spice.chunk_id` resolves whether ES stores it flat or expands it. Types that only round-trip through `_source` as an opaque `keyword` now return a structured error instead of a silently wrong `Utf8` column.

## Tests

- `data_components` lib unit tests for the reader: flat dotted `_spice.chunk_id` + `[start,end]` offset round-trip to `UInt64` + `FixedSizeList(Int32,2)`, unsigned widths, dates, all timestamp units, flat-vs-nested lookup, and the fail-loud unsupported-type path.
- `search` crate unit test asserting a chunked `ElasticsearchIndex`'s query/list schemas expose `_spice.chunk_id` + `{col}_offset`, and that `CompoundVectorIndex::try_new(memory, es)` builds both fallback plans without a plan error.

Docker-gated end-to-end coverage (live-ES `_source` shape + cold fallback) still to be added/run.


claudespice:
The `signoff` failure on this branch is two real issues, both still present at the current HEAD (`bf5e170`) — [run 30598997644](https://github.com/spiceai/spiceai/actions/runs/30598997644) burned 7210s before surfacing them, so worth fixing both before the next dispatch.

**1. `clippy::assertions_on_result_states` — 8 errors in `crates/search/src/index/elasticsearch/mod.rs:1103–1112`**

```
error: called `assert!` with `Result::is_ok`
    --> crates/search/src/index/elasticsearch/mod.rs:1103:9
...
error: could not compile `search` (lib test) due to 8 previous errors
```

`field_with_name` returns a `Result`, so the lint wants the state consumed rather than asserted:

```rust
query.field_with_name(CHUNKED_INDEX_CHUNK_KEY).expect("chunk key column should be present");
// and for the negative case on line 1112:
list.field_with_name(SEARCH_SCORE_COLUMN_NAME).expect_err("list schema should not expose the score column");
```

**2. `error[E0599]` — a genuine compile error in `crates/runtime-search/src/store_params/elasticsearch.rs:422`**

```
error[E0599]: no method named `as_deref` found for struct `std::string::String` in the current scope
   --> crates/runtime-search/src/store_params/elasticsearch.rs:422:35
```

`typed.endpoint` is a `String`, not an `Option<String>`, so the test still expects the pre-refactor shape:

```rust
assert_eq!(typed.endpoint, "http://localhost:9200");   // not .as_deref() == Some(..)
```

A scoped `cargo clippy -p search --tests` plus `cargo check -p runtime-search --tests` should reproduce both in a couple of minutes rather than two hours.

peasee (approved):

PR #12083, opened 2026-07-27
perf(cdc): 7.8x faster MySQL row-image decode — value-only decoder + replay bench

sgrebnov:
## What

A criterion benchmark (`data_components/benches/mysql_binlog_replay`) that replays a real captured MySQL binlog file through the same `mysql_common` event parser the shared MySQL replication pump uses, in two phases:

1. **Characteristics report** (full file): event-type histogram, rows-per-event/per-transaction distributions, per-table volume — ground truth for sizing pump work against real traffic.
2. **Staged throughput arms** (capped prefix, default 256 MiB): each arm adds the next hot-path stage — wire parse+checksum, `read_data()` classification, pump-style route rebuild and owned-payload buffering, the row-image walk, `Value` materialization, and a **value-only fast-decoder prototype** — so `--baseline` diffs isolate each stage's cost.

## Why

Replaying a CH-benCHmark capture (SF100, 3.6 M events / 2.9 M rows) showed the deferred per-dataset row decode — `rows_data.rows(tme)` in `buffer_rows_event` — dominates MySQL CDC replay cost by ~20× over all pump-side stages combined (53 MiB/s vs 917 MiB/s on the same hardware).

The root cause is in `mysql_common`'s `BinlogRow::deserialize`: for **every row image** (updates carry two per row) it rebuilds table-level metadata that only depends on the TABLE_MAP event — an `OptionalMetaExtractor` TLV re-parse, plus per column a name allocation and a `packets::Column` built through seven buffer-copying builder calls. Spice's decode reads only the values; none of that metadata is used. Per-table measurements confirm the cost tracks columns × images rather than bytes.

## Fast-decoder prototype

`FastRowSchema` derives the per-table decode context (column type, metadata, signedness) once per table, leaving only the null-bitmap read and per-column value parse on the per-row path. It uses only public `mysql_common` APIs — no fork required.

A/B on the reference capture (Apple M4 Max, same-session run):

| arm | time / 256 MiB | throughput | per row image |
| --- | --- | --- | --- |
| `rows_iter_values` (current decode path) | 4.798 s | 53.4 MiB/s | ~4.0 µs |
| `rows_iter_fast` (prototype) | 613.8 ms | 417 MiB/s | ~0.40 µs |

**7.8× on the decode arm; ~10× on the decode stage itself.** For the worst-case table shape in the capture (17 columns, update-heavy so two images per row) this raises the per-consumer decode ceiling from ~70 k to roughly ~700 k rows/s.

## How to run

```bash
# 1. capture a binlog from a loaded MySQL (any CH-benCHmark run works)
docker cp <mysql-container>:/var/lib/mysql/binlog.000002 /tmp/
# 2. replay it
CHBENCH_BINLOG=/tmp/binlog.000002 \
  cargo bench -p data_components --bench mysql_binlog_replay --features mysql
```

`CHBENCH_BINLOG_BENCH_BYTES` caps how much of the file the criterion arms replay (default 256 MiB); the characteristics report always scans the whole file.


sgrebnov:
Closing in favor of https://github.com/spiceai/spiceai/pull/12122

PR #12087, opened 2026-07-27
fix: Ensure OTel ingest materializes missing/null dimension columns

### LINKED ISSUE #12117: OTLP ingest: a stored metric column whose type has no attribute builder desyncs the batch, rejecting every export for that metric
## Describe the bug

When OTLP metric ingest rebuilds a metric's batch against the stored table schema, `initialize_attribute_schema` (`crates/runtime/src/opentelemetry.rs`) seeds each stored dimension column into **two** collections — a field into `fields` and a builder into `columns` — but the field is inserted unconditionally while the builder is only created for five Arrow types (`Utf8`, `Boolean`, `Int64`, `Float64`, `Binary`); every other type falls into a `_ => {}` arm:

```rust
fields.insert(field.name().clone(), Arc::clone(field));   // always
match field.data_type() {
    DataType::Utf8 => { columns.insert(..., Box::new(StringBuilder::new())); }
    // ... Boolean, Int64, Float64, Binary
    _ => {}                                               // no builder
}
```

Both callers then zip those two collections **positionally** into the batch:

```rust
fields.extend(attribute_fields_map.into_iter().map(|(_, v)| v));
columns.extend(attribute_columns_map.into_iter().map(|(_, mut v)| v.finish()));
// -> RecordBatch::try_new(Arc::new(Schema::new(fields)), columns)
```

A stored column of any other type therefore leaves `fields` one entry longer than `columns`, and `RecordBatch::try_new` rejects the entire export. `append_null` has the same five-type `match`, so it is a second, independent instance of the same gap: a column it cannot null stays shorter than its siblings.

## Effect

Every export for the affected metric fails, permanently. The batch never builds, so the operator sees only the generic

```
Failed to build arrow data from OpenTelemetry metrics for metric <name>: Failed to build record batch from OpenTelemetry metrics: Invalid argument error: number of columns(7) must match number of fields(10) in schema
```

and all of the metric's data points are counted as rejected. Nothing names the column or type responsible, so the message is not diagnosable from the outside, and there is no configuration change that recovers the metric.

## How a stored column reaches an unhandled type

The histogram path added in #11992 writes `count` as `UInt64` and `bucket_counts` / `explicit_bounds` as `List<UInt64>` / `List<Float64>`. Those are *value* columns for a histogram data point and so are skipped by `HISTOGRAM_VALUE_COLUMN_NAMES` — but they are ordinary stored **dimensions** for a data point of any other shape. So a metric whose data points arrive as a Histogram and later as a Gauge or Sum (same metric name) seeds three columns with no builder: `count`, `bucket_counts`, `explicit_bounds`.

That is exactly the case `initialize_attribute_schema`'s own comment claims to cover — "including one named like another data point shape's value column, such as `count` on a Gauge" — and it is the case that does not work.

The gap is not limited to a shape change: it holds for *any* stored column outside those five types, so a dimension whose type is widened by schema evolution, or round-tripped by an accelerator into a type outside the five, hits it the same way.

## Reproduce

Stored schema as written by a histogram export of metric `latency` with one `host` dimension:

`count` UInt64, `sum` Float64, `min` Float64, `max` Float64, `bucket_counts` List\<UInt64\>, `explicit_bounds` List\<Float64\>, `time_unix_nano` UInt64, `start_time_unix_nano` UInt64, `host` Utf8

A subsequent Gauge export of `latency` carrying only `host`:

- `NUMBER_VALUE_COLUMN_NAMES` skips `value` / `time_unix_nano` / `start_time_unix_nano`.
- Seeding produces **7** fields (`count`, `sum`, `min`, `max`, `bucket_counts`, `explicit_bounds`, `host`) and **4** builders (`sum`, `min`, `max`, `host`).
- With the three value columns prepended: `fields` = 10, `columns` = 7.
- `RecordBatch::try_new` → `number of columns(7) must match number of fields(10) in schema`.

Adding the missing builders alone is not sufficient: `append_null` cannot null a `UInt64` or list column either, so the padding pass leaves those three columns at length 0 while the rest are length 1, and `try_new` then fails with `all columns in a record batch must have the same length`. Both `match` arms have to cover the same set of types.

## Expected behavior

`fields` and `columns` stay in lockstep, because they are zipped positionally:

- Seed the pair together, or not at all. A stored column this module cannot build is dropped from the batch, with a warning naming the metric, column, and type, so the write path reports a schema mismatch it can actually describe.
- `initialize_attribute_schema` and `append_null` cover the same types, including the `UInt64` and list types the histogram path writes.

## Additional context

Introduced with histogram support in #11992 (the `UInt64` / list value columns) and reachable from the shape-change path since #12065 made those names seed as dimensions on a number metric. Distinct from #12095, which is about a stored schema carrying *duplicate* field names.

peasee:
## 📝 Summary

Hardens OpenTelemetry (OTLP) metric ingest into accelerated `sink` datasets so a metric survives schema evolution, restarts, and mixed data-point shapes instead of having its exports rejected.

* Materialize a missing/null metric dimension as a present NULL column instead of dropping it, so a data point that omits a dimension no longer fails when the stored schema still carries that column.
* Match string/binary dimension columns by type *family* (`Utf8`/`LargeUtf8`/`Utf8View` → `Utf8`, `Binary`/`LargeBinary`/`BinaryView` → `Binary`) when seeding, because accelerators such as Cayenne store strings as `Utf8View` — exact-type matching dropped those columns and produced a narrower batch the write rejected with a field-count mismatch.
* Seed builders for `UInt64` and `List<UInt64>`/`List<Float64>` dimensions and extend the null-backfill to the same types, because a metric first seen as a Histogram stores `count`/`bucket_counts`/`explicit_bounds` that become ordinary dimensions on a later Gauge/Sum — without builders they desynced the field/column lists and rejected every export (consolidates the closed #12118 onto the per-shape value-column scoping merged from trunk).
* Warn — naming the metric, column, and type — when a stored column has no builder and must be dropped, because the resulting rejection was otherwise opaque and undiagnosable.
* Inherit the acceleration checkpoint schema in the sink connector factory on restart instead of advertising the bare `placeholder` schema, because the federated-table reconciliation otherwise reported every accelerated column as missing and deferred the dataset with a schema-mismatch warning on every restart.
* Fall back to the acceleration checkpoint schema during ingest when the table is not yet registered (a `sink` dataset parked until its first write after a restart), so a data point that omits a null dimension is built against the stored wide schema and lands, rather than being rejected as a removed column.
* Fix a race in `ensure_sink_dataset` where concurrent exports for the same metric could evict an unrelated pending sink from the registration queue — leaving it permanently unregistered (`Table … not registered` on every later write) — by claiming the pending entry atomically under the write lock and re-queuing it on failure.
* Add a Cayenne restart integration test that drives the real OTLP export handler across a runtime restart, reproducing and guarding the pending-sink checkpoint path.

## 🔗 Related

* Closes #12117
* Consolidates the closed #12118

## 🚨 Breaking Changes

None.

## 📚 Docs

No docs changes required.

## 👀 Notes for Reviewers

* The `opentelemetry` module is now `pub` and exposes `build_metrics_service`, so the ingest handler can be driven directly in the integration test without standing up a gRPC transport.
* The restart integration test needs Cayenne (CDC-backed, so its file acceleration preloads across restarts, reaching the reconciliation this exercises) and the `sqlite` feature (Cayenne's checkpoint backend); a DuckDB sink parks until its first write and never reaches that path.
* `trunk` is merged in; the seeding block reconciles trunk's per-shape value-column scoping (#12065) with the family/nullable seeding added here.


claudespice:
Flagging an overlap so this doesn't get reviewed twice: #12118 (fixes #12117) rewrites the same `initialize_attribute_schema` seeding block. Both PRs fix the `fields`/`columns` desync — this one via `_ => continue`, #12118 via a `dimension_builder_for` helper returning `Option` — so they will conflict textually, but each carries something the other doesn't:

- Only this PR maps the view/large string and binary families onto the canonical builders and forces the seeded field nullable.
- Only #12118 adds builders for `UInt64` and `List<UInt64>`/`List<Float64>`, warns (naming metric/column/type) when a dimension is dropped, and extends `append_null` to the same set. That matters for the histogram shape from #11992: `count` is `UInt64` and `bucket_counts`/`explicit_bounds` are lists, and they are ordinary stored dimensions for a Gauge/Sum data point — with `_ => continue` those columns are skipped, the rebuilt batch is narrower than the stored schema, and `verify_schema` rejects the export.

This one is a draft and currently `CONFLICTING`; #12118 is mergeable, so the simplest order is to land #12118 first and rebase this on top, keeping the union of the two type sets. Happy to do it the other way round — just say which you'd prefer and I'll rebase #12118 onto this.

phillipleblanc (dismissed):

claudespice:
This PR is approved, `MERGEABLE`, has no unresolved review threads, and every check is green except `Attestation`, which is red only because the HEAD was never signed off:

> No developer sign-off found for `0afc40f`. The first-parent chain reaches unsigned commit `7ddd377`, which is not a two-parent merge and cannot inherit an earlier sign-off.

I dispatched a remote sign-off for `peasee/260727-otel-null-dimensionality-columns` ([run 30669210078](https://github.com/spiceai/spiceai/actions/runs/30669210078)). `signoff.yml` posts the `signoff` status and re-runs `Attestation` itself, so no manual re-run should be needed — the PR should go fully green on its own if lint and tests pass.

Note that `signoff.yml` uses `concurrency: signoff-<branch>` with `cancel-in-progress`, so please avoid dispatching a second sign-off for this branch while this one is running — it would cancel this run and restart the ~2-5h clock.

claudespice:
Update on the sign-off I dispatched — it **completed and failed**, and it found a real compile error rather than an infrastructure problem. Run [`30669210078`](https://github.com/spiceai/spiceai/actions/runs/30669210078), failed after 1090s:

```
error[E0308]: mismatched types
   --> crates/runtime/src/dataconnector/sink.rs:142:51
142 |                     accelerated_checkpoint_schema(dataset).await
    |                     ----------------------------- ^^^^^^^ expected `&Dataset`, found `&Arc<DatasetSpec>`
    = note: expected reference `&component::dataset::Dataset`
               found reference `&std::sync::Arc<runtime_component::dataset::DatasetSpec>`
error: could not compile `runtime` (lib) due to 1 previous error
```

**This is not a stale base.** `ConnectorComponent` is byte-identical at this PR's head (`0afc40fe4`) and at `trunk` — both declare `Dataset(Arc<DatasetSpec>)` in `crates/data-connector-api/src/lib.rs:37`. The new `accelerated_checkpoint_schema` added by this PR (`sink.rs:66`) is written against the older `component::dataset::Dataset`, so the branch does not compile on its own.

Worth flagging why this surfaced only now: under the sign-off model the heavy checks are `SKIPPED` on the PR itself, and this HEAD had no sign-off status at all (`.../commits/0afc40fe4/statuses` returned `[]`). **The dispatch was the first time this branch was compiled** — which is also why the PR could reach an approval while red.

One note on the shape of the fix, since it is probably more than a signature swap: the body reaches `dataset.runtime.accelerator_engine_registry()`, and `DatasetSpec` deliberately carries configuration only — it has no `runtime` field, per `ConnectorComponent`'s own doc comment ("connectors read a component's spicepod configuration, never the orchestrator it is attached to … reached through the runtime's separate `ConnectorContext` handle instead"). `is_file_accelerated()` does exist on `DatasetSpec` (`crates/runtime-component/src/dataset/mod.rs:650`), so that call carries over; the accelerator-registry lookup is the part that needs rerouting through `ConnectorContext`. How that should be threaded here is your call, so I have not pushed anything.

After pushing a fix, re-dispatch with:

```
gh workflow run signoff.yml -f branch=peasee/260727-otel-null-dimensionality-columns
```

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #12094, opened 2026-07-27
fix(search): drop a search hit whose row is not in the base table (fixes #12089)

### LINKED ISSUE #12089: A search hit with no matching base row is returned as an all-NULL row instead of being dropped
## Describe the bug

`SearchQueryProvider::join_with_base` (`crates/search/src/provider.rs`) joins the search index's hits to the base table with `JoinType::Left`:

```rust
let mut bldr = search_index_table
    .alias("search_index")?
    .join(
        self.underlying_table_scan(projection_column_names, filters, &search_index_schema)?,
        JoinType::Left,
        ...
```

An index entry whose primary key is no longer in the base table therefore still produces an output row — the base-table columns NULL-padded, with a real `_score`. The base table is the source of truth for which rows exist, so a hit that matches nothing there is not a result; it is a stale index entry, and emitting it turns index staleness into a wrong query answer.

Observed shape (from #12084, before that fix):

```text
+---------+--------------+--------------------+
|  title  | reddit_score |       _score       |
+---------+--------------+--------------------+
|         |              | 19.145341873168945 |
+---------+--------------+--------------------+
```

Post-join filters do drop such rows when the query has a filter (a predicate over a NULL column is not true), so this only surfaces for unfiltered searches.

## Expected behavior

A search hit with no matching base row is dropped — an inner join — so no query can return a row the dataset does not have, whatever the index's state.

## Why it matters beyond one bug

An inner join makes every source of index staleness invisible to query results rather than visible as a fabricated row:

- a full refresh that leaves documents for deleted rows (#12066)
- a chunked index that cannot apply CDC deletes (#12088)
- external vector stores (Elasticsearch, S3 Vectors, in-memory) that keep vectors for deleted rows
- the window between an accelerator applying a delete and the index being updated — the two are not atomic

It also restores `limit` semantics: a top-k search currently spends slots on stale entries.

## Additional context

Switching to `JoinType::Inner` rewrites the search plan in roughly 96 `integration_models` snapshots (`crates/runtime/tests/models/snapshots/`), which only the model CI can regenerate — worth doing deliberately rather than as a side effect of another fix. Related: #12084, #12066, #12088.

claudespice:
## Summary

`SearchQueryProvider::join_with_base` joined the search index's hits to the base table with `JoinType::Left`. An index entry whose primary key is no longer in the base table therefore still produced an output row — the base-table columns NULL-padded, carrying a real `_score`:

```text
+---------+--------------+--------------------+
|  title  | reddit_score |       _score       |
+---------+--------------+--------------------+
|         |              | 19.145341873168945 |
+---------+--------------+--------------------+
```

The base table decides which rows exist. A hit that matches nothing there is a stale index entry, not a result, so emitting it turns index staleness into a wrong query answer — a row the dataset does not have. Post-join filters happened to drop such rows whenever a query had a filter (a predicate over a NULL column is not true), which is why this only surfaced for unfiltered searches.

Fixes #12089

## Changes

- `crates/search/src/provider.rs`: `JoinType::Left` → `JoinType::Inner` in `join_with_base`, with a comment recording why the join type is load-bearing.
- Two `provider::tests` regression tests: one asserting a hit with no base row is dropped (and that the surviving hit carries real base-table values, not NULL padding), one asserting every hit that *is* backed by a base row still survives — so the fix cannot pass by filtering too aggressively.

## Why `Inner` is equivalent for every row that does have a base match

The join already carries the non-primary-key filters as a join filter, and `join_with_base` then re-applies **all** filters after the join, on the projected schema. So for a hit with a base match the two join types produce identical output; `Inner` changes the result only for hits with no match, which the post-join filter already dropped in every filtered query. The change makes that outcome unconditional rather than a side effect of a query happening to have a predicate.

Inner also restores `limit` semantics: a top-k search no longer spends result slots on entries that will never resolve to a row.

## Bug-class sweep

`JoinType::Left` appears at four other search-path sites; all four were checked and none is an instance of this defect, because what matters is which side is the index:

- `crates/search/src/index/vector_table.rs:269` — the **inverse** orientation (`base_table` LEFT JOIN `vector_index`). Keeping a base row that has no index entry is correct: a row whose embedding has not been computed yet still exists in the dataset.
- `crates/runtime-search/src/candidate/vector.rs:478,583,719` — `ChunkedNonIndexVectorGeneration`, where **both** sides derive from the same base-table scan (`score_cte_sql` scans the provider twice; `score_cte_sql_without_pks` literally clones one builder). No separately-maintained store is involved, so no entry can go stale, and the unfiltered right-hand side always contains every left-hand primary key.
- `crates/search/src/aggregation/reciprocal_rank.rs:413` uses `JoinType::Full` to merge two ranked lists, which is a different operation.

So this bug class has exactly one instance, now fixed.

## Known remaining gap (deliberately out of scope)

When `search_index_table_is_sufficient` is true, `scan` serves the query from the index alone and never joins the base table, so a stale entry is returned with the index's own stored values. That is a separate design question about the index-only fast path (see the `#7404` note in `vector_table.rs`) and forcing a join there would be a performance regression, so it is not changed here.

## Model snapshots must be regenerated by the model CI

Switching the join type rewrites the search plan in ~96 `integration_models` explain snapshots under `crates/runtime/tests/models/snapshots/` (`join_type=Left` → `join_type=Inner`, plus whatever the optimizer does differently for an inner join). Those tests are not part of any PR gate — `integration_models.yml` runs on a schedule or a manual dispatch, and `make nextest` runs `--lib` only — and they cannot be regenerated locally because they require live OpenAI / Cohere / Hugging Face / S3 Vectors credentials. The scheduled run regenerates them with `INSTA_UPDATE=always` and opens its own snapshot PR; a maintainer can also dispatch `integration_models.yml` with `update_snapshots: always`. No `.snap` file is hand-edited in this PR.

## Test plan

- `make lint`
- `make test`

## Checklist

- [x] Bug class swept across every sibling `JoinType::Left` in the search path
- [x] No snapshot hand-edited
- [ ] `integration_models` snapshots regenerated by the model CI after merge

### DIFF
--- a/crates/search/src/provider.rs
+++ b/crates/search/src/provider.rs
@@ -313,7 +313,11 @@ impl SearchQueryProvider {
             .alias("search_index")?
             .join(
                 self.underlying_table_scan(projection_column_names, filters, &search_index_schema)?,
-                JoinType::Left,
+                // The base table decides which rows exist, so a hit whose primary key is
+                // not there is a stale index entry rather than a result. An outer join
+                // would emit it with the base-table columns NULL-padded and a real
+                // `_score`, turning index staleness into a row the dataset does not have.
+                JoinType::Inner,
                 self.primary_key
                     .iter()
                     .map(|pk| {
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

Jeadie (approved):

Issue #8596, opened 2025-12-16
append on on_conflict not work


liuzx778899-hue:
## Describe the bug
I configured the dataset using the append mode according to the documentation, but it appears to have only performed the initial load. Subsequent incremental data updates are not being pulled in
<img width="1544" height="825" alt="Image" src="https://github.com/user-attachments/assets/1df77ca2-499d-45b9-a0a8-b4bda6276d8e" />

### Spicepod

```
version: v1
kind: Spicepod
name: oracle_test_app

datasets:
  - from: oracle:"FGHIS5_MZBL33"."医生_专科病历表"
    name: fghis5_mzbl33.医生_专科病历表
    ready_state: on_registration
    time_column: 操作时间
    time_format: date
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
      primary_key: 病历ID
      on_conflict:
        病历ID: upsert
      params:
        duckdb_memory_limit: '10GB'
      refresh_mode: append
      retention_check_enabled: true
      refresh_check_interval: 10s
      refresh_sql: "SELECT * FROM fghis5_mzbl33.医生_专科病历表 where 操作时间  >=  to_date('2025-12-01 00:00:00','yyyy-mm-dd hh24:mi:ss')"
```

### spice, spiced, OS info

- spice version: `1.9.1`
- spiced version: `1.9.1`
- OS info: `docker`

**Have you tried this on the latest `trunk` branch?**

- [ ] Yes
- [ ✅] No

## Screenshots

My data source constantly has incremental data available.

oracle
<img width="1968" height="357" alt="Image" src="https://github.com/user-attachments/assets/f3748c59-003a-45d5-9a6a-142f22ac8fc9" />
spice
```
sql> SELECT  COUNT(1) FROM "spice"."fghis5_mzbl33".医生_专科病历表
;
+-----------------+
| count(Int64(1)) |
|      int64      |
+-----------------+
| 168036          |
+-----------------+


Time: 0.0221513 seconds. 1 rows. 8 B (361 B/sec).
```

<img width="2115" height="1053" alt="Image" src="https://github.com/user-attachments/assets/5c78088c-e46c-427b-b352-7fad6b4d0a0b" />

## Additional context

After the initial data load, no new data has been loaded or refreshed incrementally. I also couldn't find any documentation on how to completely invalidate the cached data and force a full reload from the source. Could you please help me translate this?

liuzx778899-hue:
[...]

PR #12097, opened 2026-07-27
fix(oracle): map DATE to a timestamp so the time-of-day is not silently truncated (fixes #12096)

### LINKED ISSUE #12096: Oracle connector: DATE columns are mapped to Arrow Date32, silently discarding the time-of-day
## Describe the bug

The Oracle data connector maps the Oracle `DATE` type to Arrow `Date32`:

https://github.com/spiceai/spiceai/blob/trunk/crates/data_components/src/oracle/convert.rs#L86

```rust
"DATE" => Some(DataType::Date32),
```

Oracle's `DATE` is **not** a date-only type — it stores century, year, month, day, **hour, minute and second** ([Oracle datatype reference](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-7B72E154-677A-4342-A1EA-C74C1EA928E6)). Mapping it to `Date32` makes the time-of-day unrepresentable, and the read path then discards it: the `Date32` arm fetches the column as `chrono::NaiveDate`.

https://github.com/spiceai/spiceai/blob/trunk/crates/data_components/src/oracle/convert.rs#L248-L261

Every `DATE` value is therefore silently truncated to midnight.

### This is a lossy choice, not a driver limitation

In `oracle` 0.6.3, an Oracle `DATE` column binds as `NativeType::Timestamp`:

```rust
// oracle-0.6.3/src/sql_type/oracle_type.rs:420
OracleType::Date => Ok(VarParam::new(DPI_ORACLE_TYPE_DATE, NativeType::Timestamp)),
```

Both `chrono` impls read that same value via `SqlValue::to_timestamp()`; they differ only in what they keep:

```rust
// FromSql for NaiveDate  — src/sql_type/chrono.rs:250
NaiveDate::from_ymd_opt(ts.year(), ts.month(), ts.day())   // h/m/s dropped

// FromSql for NaiveDateTime — src/sql_type/chrono.rs:214
naive_date_time_from_sql(&ts)   // keeps hour, minute, second, nanosecond
```

So the connector selected the lossy impl; the full value is available.

## Impact

1. **Wrong query results.** A row whose `DATE` column holds `2026-07-27 14:32:11` is returned as `2026-07-27T00:00:00`. Any projection, filter, sort, or aggregation over the time-of-day is wrong, with no error or warning.

2. **Incremental (`refresh_mode: append`) refresh stalls on a `DATE` time column.** The append refresh takes the max value of `time_column` in the accelerator and asks the source for rows strictly greater than it (`crates/runtime/src/accelerated_table/refresh_task.rs`, via `TimestampFilterConvert`). Because every value of the day collapses to that day's midnight, once any row for day *D* has been loaded, `max = D 00:00:00`, and every subsequent row created later on day *D* also reads as `D 00:00:00` — so `value > max` is false and the row is **never appended**. New data is silently skipped for the remainder of each day.

   This is the mechanism behind the report in #8596 (Oracle `DATE` time column, `refresh_mode: append`, initial load succeeds and nothing is picked up afterwards).

## To Reproduce

```sql
CREATE TABLE t ("ID" NUMBER(10), "EVENT_TIME" DATE);
INSERT INTO t VALUES (1, TO_DATE('2026-07-27 14:32:11', 'YYYY-MM-DD HH24:MI:SS'));
```

```yaml
datasets:
  - from: oracle:t
    name: t
```

```sql
sql> describe t;
-- EVENT_TIME | Date32          <-- should carry a time component

sql> select "EVENT_TIME" from t;
-- 2026-07-27                   <-- actual value is 2026-07-27 14:32:11
```

## Why this was not caught

The Oracle integration fixture only inserts midnight `DATE` literals (`DATE '2024-06-27'`, `DATE '2025-01-01'` in `crates/runtime/tests/oracle/mod.rs`), which are unaffected by the truncation, so the snapshots agree either way.

## Expected behavior

An Oracle `DATE` column should round-trip its time-of-day. Mapping it to `Timestamp(Second, None)` matches Oracle's 1-second resolution and reuses the existing conversion arm for `TIMESTAMP(0)`, which already fetches `chrono::NaiveDateTime`.

Note this changes the reported Arrow type for Oracle `DATE` columns from `Date32` to `Timestamp(s)`.

## Bug class

Checked every sibling connector that maps a source type to `Date32` — ClickHouse `Date`/`Date32`, ScyllaDB `date`, MSSQL `date`, MySQL `DATE`, Debezium `io.debezium.time.Date`, Delta Lake / Databricks `DATE`. In all of those the source type is genuinely date-only, so `Date32` is correct. Oracle is the sole outlier, because Oracle `DATE` is a datetime.

claudespice:
## Summary

Oracle's `DATE` type is **not** a date-only type — it stores century, year, month, day, **hour, minute and second** ([Oracle datatype reference](https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-7B72E154-677A-4342-A1EA-C74C1EA928E6)). The connector mapped it to Arrow `Date32`, which cannot represent the time-of-day, and the `Date32` read path fetches the column as `chrono::NaiveDate`. Every `DATE` value was therefore silently truncated to midnight.

### Root cause

`map_oracle_type_to_arrow_type` returned `Date32` for `DATE`, so `rows_to_arrow` selected the `Date32` arm and read the column as `chrono::NaiveDate`.

This was a lossy choice, not a driver limitation. In `oracle` 0.6.3 an Oracle `DATE` column binds as `NativeType::Timestamp` (`src/sql_type/oracle_type.rs:420`), and both `chrono` impls read that same value through `SqlValue::to_timestamp()` — they differ only in what they keep:

```rust
// FromSql for NaiveDate    — keeps (year, month, day); hour/minute/second dropped
NaiveDate::from_ymd_opt(ts.year(), ts.month(), ts.day())
// FromSql for NaiveDateTime — keeps hour, minute, second, nanosecond
naive_date_time_from_sql(&ts)
```

### Impact

1. **Wrong query results.** A row holding `2026-07-27 14:32:11` was returned as `2026-07-27` — no error, no warning. Any projection, filter, sort or aggregation over the time-of-day was wrong.

2. **`refresh_mode: append` stalls on a `DATE` time column.** The append refresh takes the max `time_column` value in the accelerator and asks the source for rows strictly greater than it. With every value of a day collapsed onto that day's midnight, once any row for day *D* is loaded `max = D 00:00:00`, and every row created later that same day also reads as `D 00:00:00` — so `value > max` is never true and the row is never appended. New data was silently skipped for the rest of each day. This is the mechanism behind the report in #8596.

## Changes

- `crates/data_components/src/oracle/convert.rs` — map `DATE` to `Timestamp(Second, None)`, which matches Oracle's 1-second resolution and reuses the existing conversion arm for `TIMESTAMP(0)` (already fetches `chrono::NaiveDateTime`). The `(Timestamp(Second, None), _)` arm matches any native type, so a `DATE` column selects it; no arm ordering changes were needed.
- `crates/runtime/tests/oracle/mod.rs` — the fixture only inserted midnight `DATE` literals, which are unaffected by the truncation and are why the bug survived. Row 1's `VAL_DATE` now carries a time-of-day so the integration snapshot detects a regression; row 2 keeps a midnight value as the counterpart.

### Behavior change

The Arrow type reported for Oracle `DATE` columns changes from `Date32` to `Timestamp(s)`. This is deliberate — it is the only way the time-of-day can survive — and it brings Oracle in line with how every other connector surfaces a datetime source type.

Filter pushdown is unaffected: `is_datetime_related_expr` already treats both `Date32` and `Timestamp` as datetime-related, so `DATE` predicates remain `Unsupported` for pushdown exactly as before (tracked separately in #6325).

**Flag for reviewers:** the TPC-H federated Oracle test configs (`test/spicepods/tpch/sf1/federated/oracle*.yaml`) have `DATE` columns that will now surface as `Timestamp(s)`. Any stored testoperator result comparisons for those configs may need regenerating; that runs on schedule/dispatch rather than on this PR.

### Bug class

Swept every sibling connector that maps a source type to `Date32` — ClickHouse `Date`/`Date32`, ScyllaDB `date`, MSSQL `date`, MySQL `DATE`, Debezium `io.debezium.time.Date`, Delta Lake and Databricks `DATE`. In all of those the source type is genuinely date-only, so `Date32` is correct. Oracle is the sole outlier, because Oracle `DATE` is a datetime. The class is fully addressed by this one change.

## Test plan

- `make lint`
- `make test`

Added `oracle_date_maps_to_a_type_that_can_hold_the_time_of_day`, a regression test asserting `DATE` maps to `Timestamp(Second, None)` and, separately, that it is not a date-only Arrow type. Updated the `DATE` entry in the existing `test_common_oracle_types_mappings` table.

**Neuter test passed:** with `"DATE" => Some(DataType::Date32)` restored, the new regression test and `test_common_oracle_types_mappings` both fail (`left: Some(Date32)`, `right: Some(Timestamp(Second, None))`) while the unrelated tests in the module still pass. With the fix in place, `cargo test -p data_components --lib --features oracle` is **460 passed / 0 failed**, and both CI clippy passes (`--lib --bins`, then `--tests`) are clean.

**Snapshots:** `integration__oracle__schema.snap` (`Date32` → `Timestamp(s)`) and `integration__oracle__data.snap` (the `VAL_DATE` column) were updated by a table-rewriting script rather than by `INSTA_UPDATE`, because regenerating them requires Docker and the `gvenzl/oracle-free` image, which this environment does not have. Only the `VAL_DATE` column changed; every other column is byte-identical, including widths.

Note on where that is verified: `integration.yml` runs its jobs as passthrough no-ops on `pull_request` ("Integration tests run on the merge-group commit"), so the green Integration Tests checks on this PR do **not** exercise the Oracle test. The two updated snapshots are first actually checked in the **merge queue**. Flagging that explicitly so the PR's green checks are not read as having validated them.

## Checklist

- [x] Root cause identified and verified at the driver level
- [x] Regression test added that fails on the old code
- [x] Bug-class sweep across sibling connectors completed
- [x] Both CI clippy passes clean; `cargo fmt` clean
- [ ] Oracle integration job green (verifies the two updated snapshots)

Fixes #12096


claudespice:
@copilot review

lukekim (approved):

Issue #12102, opened 2026-07-27
Warm vector tier holds only rows refreshed since startup when the accelerator persists, so search silently narrows after a restart


claudespice:
## Describe the bug

The in-memory warm vector tier added in #11914 is meant to hold the same rows the acceleration write path holds (#11828: "Warm-tier index contents should follow the same data-window mechanisms as acceleration … If no retention is configured, the warm tier keeps whatever the acceleration write path keeps"). It diverges from that whenever the accelerator **persists across restarts** but the warm tier does not:

- A file-backed accelerator (`acceleration.mode: file` DuckDB/SQLite, or Cayenne) keeps its rows on disk across a restart.
- With `refresh_mode: append` or `changes`, the refresh after a restart reads only rows newer than what the accelerator already holds, so only those rows flow through the source scan that feeds the vector index.
- `MemoryVectorStore` starts empty on every process start (`crates/search/src/index/memory/store.rs`) and has no hydration path, so it ends up holding only the rows refreshed since startup — while the accelerator, and the external vector engine, hold everything.

Because the warm tier is the compound index's *primary*, vector search then answers from that small recent subset. Neither read mode rescues it: `CompoundReadMode::PrimaryOnly` (the default, from `on_zero_results: return_empty`) never consults the engine index, and `FallbackToSecondary` falls back only when the primary returns *exactly zero* rows — so a partially-hydrated warm tier masks the complete engine index in both modes.

Net effect: after a restart, a vector search silently returns only recently-refreshed rows instead of the dataset's rows. No error, no warning.

## To Reproduce

```yaml
datasets:
  - from: postgres:public.docs
    name: docs
    time_column: updated_at
    vectors:
      enabled: true
      engine: s3_vectors
      params:
        s3_vectors_bucket: my-s3-vector-bucket
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
      refresh_mode: append
    columns:
      - name: body
        embeddings:
          - from: bedrock_titan
```

1. Start `spiced`, let the initial refresh populate the accelerator, the S3 Vectors index and the warm tier. Vector search returns the expected rows.
2. Restart `spiced`. The file accelerator still holds every row, so the append refresh reads only rows newer than the last `updated_at`.
3. Run the same vector search: only the rows the post-restart refresh appended come back.

## Expected behavior

[...]

Issue #11828, opened 2026-07-13
Enhancement: Warm search tier for external search indexes


Jeadie:
## Goal-State/What/Result
Provide a faster warm-tier search interface for Spice instances with search indexes. 

## Why/Purpose
Spice is frequently the ingestion, operational search and operational search node. Many operational uses cases wish to search recent data (i.e. recent timestamp or recently ingested), for relevancy ot latency reasons. Using an external indexes or the full analytic search index is both unnecessary (Spice has just indexed the data) and slow. 

## By When

**Issue/Spec written and reviewed:** 14th July 2026
**Done-Done:** 20th July 2026

## Done-Done

- [ ] [Principles Driven](https://github.com/spiceai/spiceai/blob/trunk/docs/PRINCIPLES.md)
- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Threat Model / Security Review
- [ ] Tests
- [ ] Telemetry / Metrics / Task History
- [ ] Performance / Benchmarks
- [ ] Documentation
- [ ] Cookbook Recipes/Tutorials

## The Algorithm

- [ ] Every requirement questioned?
- [ ] Delete (Scope) any part you can.
- [ ] Simplify.
- [ ] Break down into smaller iterations/milestones.
- [ ] Opportunities for automation.

## Specification
One assumption: a `spiced` node will only query/search across the data it ingests, because either:
1. single node process
2. Executor in distributed cluster, assigned subset of partitions (note: distributed search not yet supported, but is on roadmap.)

With this assumption, we can avoid any issues with TOCTOU in creating or updating the warm tier. 

### DX
For now consider warm tier of local index (e.g. duckdb-vss, or just additional columns of in-memory arrow) and external indexes (S3vectors, elasticsearch) separately.

#### External Index.
Warm-tier index contents should follow the same data-window mechanisms as acceleration. The warm index receives the rows and index-required columns emitted by the acceleration write path after `refresh_sql`, predicates, limits, retention, and projection pruning.

Retention is controlled by existing acceleration settings such as `retention_period`, `retention_sql`, `retention_check_enabled`, and `retention_check_interval`. If no retention is configured, the warm tier keeps whatever the acceleration write path keeps.

[...]

PR #12103, opened 2026-07-27
fix(search): skip the warm vector index when nothing hydrates it (fixes #12101)

### LINKED ISSUE #12101: Vector search returns nothing without acceleration: the always-on warm index is never hydrated but serves reads
## Describe the bug

A dataset (or view) with `vectors.enabled: true` and an external vector engine (`s3_vectors`, `elasticsearch`) but **no acceleration** now always gets an in-memory warm vector index installed in front of the engine index, and that warm index is the compound index's *primary* read tier. Nothing hydrates it for a non-accelerated table, so vector search returns **zero rows** (or an arbitrary subset) instead of the rows the external vector store actually holds.

The warm-tier spec (#11828) calls this configuration out explicitly:

> Search (either via SQL or HTTP api) queries the local warm index. There are two ways a query hits the external store:
> 1. If `acceleration.on_zero_results: use_source` and the warm index returns exactly zero rows after applying the search query and filters.
> 2. **When no acceleration is applied (not currently supported; requires external coordination since it won't be hydrated by spiced).**

The implementation installs the warm tier unconditionally, so case 2 is not skipped — it gets the *least* safe read mode.

## Root cause

`with_memory_warm_index` (`crates/runtime-search/src/embeddings/warm_index.rs`) pairs the engine index with a `MemoryVectorIndex` in a `CompoundVectorIndex` and picks the read mode from the dataset's `acceleration.on_zero_results`:

```rust
let read_mode = match on_zero_results {
    ZeroResultsAction::ReturnEmpty => CompoundReadMode::PrimaryOnly,
    ZeroResultsAction::UseSource => CompoundReadMode::FallbackToSecondary,
};
```

Both call sites derive that value with `unwrap_or_default()`:

```rust
// crates/runtime/src/embeddings/connector.rs, and the same shape in crates/runtime/src/view.rs
let on_zero_results = dataset
    .acceleration
    .as_ref()
    .map(|acceleration| acceleration.on_zero_results.clone())
    .unwrap_or_default();
```

So **no acceleration block at all** (and `acceleration.enabled: false`) yields `ZeroResultsAction::ReturnEmpty` — the enum's `#[default]` — which maps to `CompoundReadMode::PrimaryOnly`: reads are served **only** from the in-memory tier, with no fallback to the engine index.

The in-memory tier starts empty on every process start and has no hydration path: `MemoryVectorStore::new` begins with `batches: Vec::new()`, and the only way rows enter it is `MemoryVectorIndex::write` (`crates/search/src/index/memory/`). There is no bootstrap from the engine index and no persistence.

For an **accelerated** dataset the acceleration write path drives those writes, which is the design in #11828. Without acceleration there is no such write path, so the tier is populated only by whatever rows happen to flow through a scan of the `IndexedTableProvider` — `IndexTableScanOptimizerRule` attaches the index-computing node to *every* qualifying `TableScan`, filters, limits and all. That means:

- **Nothing has scanned the table yet** → warm tier empty → `PrimaryOnly` returns zero rows. Vector search is silently broken.
- **Something has scanned part of it** (e.g. `SELECT ... WHERE ... LIMIT 10`) → warm tier holds those rows → search answers from that subset. `FallbackToSecondary` does not help either: it falls back only when the primary yields *exactly zero* rows, so a partially-populated warm tier masks the complete engine index in both read modes.

This is a regression: before #11914 the engine index served these reads directly and returned the full result set.

Affected paths (one shared helper, both engines): `wrap_table_as_index_s3` and `wrap_table_as_index_elasticsearch` in `crates/runtime/src/embeddings/index/table.rs`.

## To Reproduce

```yaml
datasets:
  - from: s3://my-bucket/docs/
    name: docs
    params:
      file_format: parquet
    vectors:
      enabled: true
      engine: s3_vectors
      params:
        s3_vectors_bucket: my-s3-vector-bucket
    columns:
      - name: body
        embeddings:
          - from: bedrock_titan
```

No `acceleration:` block. The S3 Vectors index was populated by an earlier run (or out-of-band).

1. Start `spiced`.
2. Run a vector search over `docs` (`vector_search` UDTF or `POST /v1/search`).
3. Zero results are returned, though the S3 Vectors index holds the vectors.

## Expected behavior

Vector search returns the rows the external vector store holds. A warm tier is an optimization and must never narrow a result set — when nothing hydrates it, it should not be installed at all, matching #11828's "not currently supported" for the non-accelerated case (and the existing precedent that full-text search errors at load unless the dataset is accelerated — `validate_dataset` in `crates/runtime/src/init/dataset.rs`).

## Additional context

Separate, still-open gap for the accelerated case, worth tracking on its own: a **persisted** accelerator (file-mode DuckDB/SQLite, Cayenne) combined with `refresh_mode: append`/`changes` resumes without re-reading existing rows after a restart, so the warm tier holds only the rows refreshed since process start while the engine index holds everything — the same "partially populated primary masks the complete secondary" shape, this time within a supported accelerated configuration.

Introduced by #11914 (`1676acca7`), spec #11828.

claudespice:
## Summary

A dataset or view with `vectors.enabled: true` on an external vector engine (S3 Vectors, Elasticsearch) but **no acceleration** was still given an in-memory warm vector index in front of the engine index, and that warm index is the compound index's *primary* read tier. Nothing hydrates it in that configuration, so vector search answered from an index that is empty — or that holds only whatever rows a scan happened to write since process start — instead of from the vector engine.

Root cause: `with_memory_warm_index` picks the compound read mode from the dataset's `acceleration.on_zero_results`, and both call sites derived that value with `unwrap_or_default()`. With no acceleration block that yields `ZeroResultsAction::ReturnEmpty` (the enum's `#[default]`), which maps to `CompoundReadMode::PrimaryOnly` — reads served only from the in-memory tier, with no fallback to the engine index. `MemoryVectorStore` starts empty on every process start and is only ever filled by `MemoryVectorIndex::write`, which the acceleration write path drives; without acceleration there is no such write path. `FallbackToSecondary` does not rescue the partially-populated case either: it falls back only when the primary yields exactly zero rows.

The warm-tier spec (#11828) calls this configuration out — "When no acceleration is applied (not currently supported; requires external coordination since it won't be hydrated by spiced)" — but the implementation installed the tier unconditionally. This matches the existing precedent that a full-text column requires acceleration (`validate_dataset`); here the engine index works fine on its own, so the warm tier is simply skipped rather than made an error, consistent with `with_memory_warm_index`'s existing contract that a missing warm index must never fail a dataset load.

Fixes #12101

## Changes

- `crates/runtime-search/src/embeddings/mod.rs`: new `warm_index_on_zero_results(Option<&Acceleration>) -> Option<&ZeroResultsAction>` — the single definition of when a warm tier applies. `None` (no warm tier) when acceleration is absent or disabled.
- `crates/runtime-search/src/embeddings/warm_index.rs`: `with_memory_warm_index` takes `Option<&ZeroResultsAction>` and returns the engine index unchanged on `None`, logging why.
- `crates/runtime/src/embeddings/index/table.rs`: threads `Option<&ZeroResultsAction>` through `wrap_table_as_index` and both engine helpers, so the S3 Vectors and Elasticsearch paths are fixed together.
- `crates/runtime/src/embeddings/connector.rs`, `crates/runtime/src/view.rs`: derive the value with the shared helper instead of `unwrap_or_default()`.

Sibling arm checked and deliberately not changed: the Elasticsearch **full-text** compound index (`crates/runtime/src/search/full_text/`) maps `on_zero_results` the same way, but a dataset with an enabled full-text column cannot reach it without acceleration — `validate_dataset` in `crates/runtime/src/init/dataset.rs` returns `FullTextSearchRequiresAcceleration` first.

Behavior for an accelerated dataset or view is unchanged: the same `on_zero_results` → read-mode mapping applies.

Filed while working on this and left for a separate change: #12102 — a warm tier paired with a *persisted* accelerator (file-mode DuckDB/SQLite, Cayenne) plus `refresh_mode: append`/`changes` holds only the rows refreshed since startup, so search narrows after a restart. That needs a hydration or completeness design call rather than a wiring fix.

Overlaps #12082 (chunked Elasticsearch warm tier), which also edits `warm_index.rs` and `embeddings/index/table.rs` from a base that predates #11914. Whichever lands second needs a trunk merge; the new call site #12082 adds should take the same `Option<&ZeroResultsAction>` so a chunked column gets no warm tier without acceleration either.

## Test plan

- New `on_zero_results_is_none_without_an_enabled_acceleration` (`crates/runtime-search/src/embeddings/mod.rs`): no acceleration and a disabled acceleration both yield `None`; an enabled one passes its `on_zero_results` through.
- New `warm_index_is_skipped_without_acceleration` (`crates/runtime-search/src/embeddings/warm_index.rs`): `None` returns the engine index unchanged rather than a `CompoundVectorIndex`. Fails before this change (the engine index came back wrapped in a compound in `PrimaryOnly` mode).
- Existing `warm_index_read_mode_follows_on_zero_results` still asserts `use_source` → `FallbackToSecondary` and `return_empty` → `PrimaryOnly` for an accelerated table, so the fix cannot pass by always skipping the warm index.
- `make lint`
- `make nextest`

## Checklist

- [x] Root cause identified and the whole class (both vector engines) fixed in one change
- [x] Sibling `on_zero_results` mapping audited (full-text path already guarded)
- [x] No behavior change for accelerated datasets and views
- [ ] `make lint` and `make nextest` green (via `make signoff`)

### DIFF
--- a/crates/runtime-search/src/embeddings/mod.rs
+++ b/crates/runtime-search/src/embeddings/mod.rs
@@ -26,10 +26,27 @@ use std::sync::Arc;
 
 use chunking::{Chunker, ChunkingConfig};
 use llms::embeddings::{Embed, Error as EmbedError};
+use runtime_acceleration::acceleration::{Acceleration, ZeroResultsAction};
 use std::collections::HashMap;
 use tokio::sync::RwLock;
 
 pub type EmbeddingModelStore = HashMap<String, Arc<dyn Embed>>;
+
+/// The read behavior a warm search tier should be built with for a table with this
+/// `acceleration`, as `warm_index::with_memory_warm_index` takes it.
+///
+/// `None` — no warm tier at all — when the table has no enabled acceleration: a warm tier
+/// starts empty on every process start and is filled by the acceleration write path, so
+/// without acceleration nothing hydrates it and serving searches from it would narrow
+/// results to whatever rows a scan happened to write, or to nothing at all.
+#[must_use]
+pub fn warm_index_on_zero_results(
+    acceleration: Option<&Acceleration>,
+) -> Option<&ZeroResultsAction> {
+    acceleration
+        .filter(|acceleration| acceleration.enabled)
+        .map(|acceleration| &acceleration.on_zero_results)
+}
 
 pub async fn construct_chunker(
     model_name: &str,
@@ -43,4 +60,42 @@ pub async fn construct_chunker(
         });
     };
     embed_model.chunker(chunk_config)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Regression test for #12101: the acceleration write path is the only thing that fills a
+    /// warm search tier, so an absent or disabled acceleration must yield no warm tier at all.
+    #[test]
+    fn on_zero_results_is_none_without_an_enabled_acceleration() {
+        assert_eq!(
+            warm_index_on_zero_results(None),
+            None,
+            "a table with no acceleration must get no warm tier"
+        );
+
+        let disabled = Acceleration {
+            enabled: false,
+            on_zero_results: ZeroResultsAction::UseSource,
+            ..Acceleration::default()
+        };
+        assert_eq!(
+            warm_index_on_zero_results(Some(&disabled)),
+            None,
+            "a disabled acceleration must get no warm tier"
+        );
+
+        let enabled = Acceleration {
+            enabled: true,
+            on_zero_results: ZeroResultsAction::UseSource,
+            ..Acceleration::default()
+        };
+        assert_eq!(
+            warm_index_on_zero_results(Some(&enabled)),
+            Some(&ZeroResultsAction::UseSource),
+            "an enabled acceleration passes its on_zero_results through"
+        );
+    }
 }
--- a/crates/runtime-search/src/embeddings/warm_index.rs
+++ b/crates/runtime-search/src/embeddings/warm_index.rs
@@ -35,6 +35,12 @@ use search::metadata::MetadataColumns;
 /// [`ZeroResultsAction::UseSource`] falls back to the engine index, while
 /// [`ZeroResultsAction::ReturnEmpty`] serves the (possibly empty) in-memory result as-is.
 ///
+/// `on_zero_results` is `None` when the table has no enabled acceleration. The in-memory
+/// index starts empty on every process start and is only ever filled by the acceleration
+/// write path, so without acceleration nothing hydrates it — serving reads from it would
+/// narrow searches to whatever rows happened to be scanned since startup, or to nothing
+/// at all. The engine index is then returned unchanged.
+///
 /// The warm index is an optimization: when a compatible in-memory index cannot be built
 /// (e.g. the engine's distance metric has no in-memory equivalent), the engine index is
 /// returned unchanged and a warning is logged — a dataset that works without a warm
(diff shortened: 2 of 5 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12118, opened 2026-07-28
fix(otel): keep a seeded metric dimension's field and builder in lockstep (fixes #12117)

### LINKED ISSUE #12117: OTLP ingest: a stored metric column whose type has no attribute builder desyncs the batch, rejecting every export for that metric

claudespice:
Fixes #12117.

## What was wrong

`initialize_attribute_schema` (`crates/runtime/src/opentelemetry.rs`) seeds each stored dimension column into two collections — a field into `fields`, a builder into `columns` — and both callers zip those collections **positionally** into `RecordBatch::try_new`. The field was inserted unconditionally, but the builder only for `Utf8` / `Boolean` / `Int64` / `Float64` / `Binary`; everything else fell into `_ => {}`. A stored column of any other type therefore left `fields` one entry longer than `columns` and the entire export was rejected, with only

```
Invalid argument error: number of columns(7) must match number of fields(10) in schema
```

to go on — a message that names neither the column nor its type. Every export for that metric failed, permanently, with no configuration change that recovers it.

`append_null` carried the same five-type `match`, an independent instance of the same gap: a column it cannot null stays shorter than its siblings.

The reachable trigger is the histogram path, which writes `count` as `UInt64` and `bucket_counts` / `explicit_bounds` as `List<UInt64>` / `List<Float64>`. Those are *value* columns for a histogram data point, so they are skipped when seeding a histogram — but ordinary stored **dimensions** for a data point of another shape. A metric whose data points arrive as a Histogram and later as a Gauge or Sum seeds exactly those three with no builder. That is the case `initialize_attribute_schema`'s own comment cites as the reason the seeding exists ("including one named like another data point shape's value column, such as `count` on a Gauge"), and it was the case that did not work. The gap is not specific to a shape change: any stored column outside the five types hits it the same way.

## The fix

Seeding is all-or-nothing. `dimension_builder_for` decides whether a builder exists; the field and builder are inserted together, and a column it cannot build is dropped with a warning naming the metric, column, and type — so the write path reports a schema mismatch it can describe instead of an arrow-internals error. `dimension_builder_for` and `append_null` cover the same set, now including `UInt64` and `List<UInt64>` / `List<Float64>`; the list element field is taken from the stored schema so the built array's type matches the field exactly.

## Verification

Both halves are load-bearing, confirmed against a standalone harness reproducing the three functions and the positional zip at the arrow level, using the schema a histogram export actually writes:

| variant | result |
| --- | --- |
| before this change | `fields`=10, `columns`=7 → `number of columns(7) must match number of fields(10) in schema` |
| builders added, `append_null` arms reverted | 10/10 → `all columns in a record batch must have the same length` |
| this change | builds; `count`/`sum`/`bucket_counts`/`explicit_bounds` null, `host` intact |

Two unit tests are added to `crates/runtime/src/opentelemetry.rs`:

- `histogram_columns_seed_as_dimensions_on_a_later_gauge_export` — builds the stored schema from a real histogram export, then a Gauge export against it; asserts every stored dimension keeps its position and is null where the shape does not carry it. Pins both halves of the fix.
- `stored_column_without_a_builder_is_dropped_instead_of_desyncing_the_batch` — a stored `Int32` column (no builder on the attribute path) is dropped rather than desyncing the batch.

`cargo test -p runtime --lib opentelemetry::` is green: **12 passed, 0 failed**, covering the two new tests and the ten pre-existing OTLP tests. `cargo fmt -p runtime` is clean. Clippy is left to the sign-off's targeted pre-lint, which runs before the full workspace gate.

### DIFF
--- a/crates/runtime/src/opentelemetry.rs
+++ b/crates/runtime/src/opentelemetry.rs
@@ -670,7 +670,13 @@ fn attributes_to_fields_and_columns(
     let mut columns: IndexMap<String, Box<dyn ArrayBuilder>> = IndexMap::new();
     let mut warned_collisions: HashSet<&str> = HashSet::new();
 
-    initialize_attribute_schema(&mut fields, &mut columns, existing_schema, value_columns);
+    initialize_attribute_schema(
+        metric,
+        &mut fields,
+        &mut columns,
+        existing_schema,
+        value_columns,
+    );
 
     for (i, inner_attributes) in attributes.iter().enumerate() {
         for attribute in *inner_attributes {
@@ -786,7 +792,37 @@ fn attributes_to_fields_and_columns(
     (fields, columns)
 }
 
+/// Builder matching a stored dimension column's Arrow type, or `None` when this module has no
+/// builder for it.
+///
+/// `UInt64` and the two list types cover the columns the histogram path writes — `count` and
+/// the `bucket_counts`/`explicit_bounds` arrays. Those are value columns on a histogram and so
+/// skipped by [`initialize_attribute_schema`], but a metric whose data points arrive in a
+/// different shape later sees them as ordinary stored dimensions. The list element field is
+/// taken from the stored schema so the built array's type matches the field exactly.
+fn dimension_builder_for(field: &Field) -> Option<Box<dyn ArrayBuilder>> {
+    match field.data_type() {
+        DataType::Utf8 => Some(Box::new(StringBuilder::new())),
+        DataType::Boolean => Some(Box::new(BooleanBuilder::new())),
+        DataType::Int64 => Some(Box::new(Int64Builder::new())),
+        DataType::UInt64 => Some(Box::new(UInt64Builder::new())),
+        DataType::Float64 => Some(Box::new(Float64Builder::new())),
+        DataType::Binary => Some(Box::new(BinaryBuilder::new())),
+        DataType::List(item) => match item.data_type() {
+            DataType::UInt64 => Some(Box::new(
+                ListBuilder::new(UInt64Builder::new()).with_field(Arc::clone(item)),
+            )),
+            DataType::Float64 => Some(Box::new(
+                ListBuilder::new(Float64Builder::new()).with_field(Arc::clone(item)),
+            )),
+            _ => None,
+        },
+        _ => None,
+    }
+}
+
 fn initialize_attribute_schema(
+    metric: &str,
     fields: &mut IndexMap<String, Arc<Field>>,
     columns: &mut IndexMap<String, Box<dyn ArrayBuilder>>,
     existing_schema: Option<&Schema>,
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

claudespice:
Heads-up on an overlap I found while triaging: draft PR #12087 rewrites the same `initialize_attribute_schema` seeding block, and its first commit predates this PR. Both fix the field/builder desync — #12087 by `_ => continue` (seed nothing when there is no builder), this PR by `dimension_builder_for` returning `Option`.

The two are complementary rather than redundant, but they will conflict textually:

- #12087 additionally maps the view/large string and binary families (`Utf8View`, `LargeUtf8`, `BinaryView`, `LargeBinary`) onto the canonical builders and forces the seeded field nullable. This PR does neither.
- This PR additionally *supports* `UInt64` and `List<UInt64>`/`List<Float64>` instead of dropping them, warns with the metric/column/type when a dimension is dropped, and extends `append_null` to the same set. Under #12087 alone a stored `count` (`UInt64`) dimension is skipped, so the rebuilt batch omits a column the stored schema has and `verify_schema` rejects the export — which is the reachable trigger #12117 describes.

#12087 is currently a draft and `CONFLICTING`, so I'll leave this one as-is; whichever lands second should rebase and keep the union of the two type sets rather than either `match` alone.

peasee:
addressed by https://github.com/spiceai/spiceai/pull/12087

PR #12121, opened 2026-07-28
fix(cli): report a truncated release download as a failed download (fixes #12120)

### LINKED ISSUE #12120: spice install/upgrade: a truncated release download is reported as a broken archive, not a failed download
## Summary

`spice install` and `spice upgrade` download a release asset, then hand the bytes straight to `extract_tar_gz` without checking that the download is complete. The expected length is already known and already used — but only to render the progress percentage, never to validate the result.

- `bin/spice/src/github/mod.rs::download_with_progress` checks `response.status().is_success()`, then accumulates `bytes_stream()` into a `Vec<u8>` and returns it. `response.content_length()` is read into `total_size` and passed to the progress callback only.
- `bin/spice/src/github/release.rs::download_release_asset` knows `asset.size` from the releases API (it is `total_size` for the progress line), and calls `extract_tar_gz(&data, download_dir)` without ever comparing `data.len()` to it.

## Impact

A stream that ends early — a dropped connection, a proxy that truncates, a captive portal serving a `200` block page — produces a short or non-tarball `data`. `extract_tar_gz` then fails and reports it as a local IO problem:

```
IO error: Failed to extract archive: invalid gzip header
```

That misattributes a network/server failure to local IO, and the response body's actual content is never shown. `upgrade_cli_in_place` (`bin/spice/src/github/release.rs`) is on this path and replaces the running CLI binary, so a silently truncated download is worth catching explicitly rather than inferring from a decoder error.

## Suggested fix

Compare the downloaded length against the expected one and fail with a distinct error naming both. An expected-length check is authoritative — it subsumes what a gzip-magic-byte check would catch and also catches truncation of a well-formed prefix:

- in `download_release_asset`, verify `data.len() as u64 == asset.size` before extracting;
- optionally also verify against `content_length()` inside `download_with_progress` when the header is present.

Report it as a download/fetch failure (naming the asset, the expected size, and the received size), not as an extraction or IO error.

## Context

Found while fixing #12116, which is the same shape in the other of the CLI's two HTTP-body-to-binary-decoder paths: the Spicepod registry answered `HTTP 200` with a JSON error body and the CLI blamed the archive. That one was fixed in #12119. This path was deliberately left out of that PR — there is no observed failure here (GitHub does not serve `200` + error body on asset downloads), and the right check is different enough that it needs its own mock-GitHub tests rather than sharing a mechanism.

claudespice:
## Summary

`spice install` and `spice upgrade` download a release asset and hand the bytes straight to the gzip/tar decoder without checking the download completed. The releases API publishes an exact `size` for every asset, and it was already read — but only to render the progress percentage, never to validate the result.

A stream that ends early (a dropped connection, a truncating proxy, a captive portal serving a `200` block page) therefore surfaced as whatever the decoder made of the fragment that arrived:

```
IO error: Failed to extract archive: invalid gzip header
```

That misattributes a network failure to local IO and never shows what actually came back. `upgrade_cli_in_place` is on this path and replaces the running CLI binary, so it is worth catching explicitly rather than inferring from a decoder error.

Fixes #12120

## Changes

- `GitHubClient::download_asset` (was `download_with_progress`) verifies the received byte count against the size the release publishes, and returns a new `GitHubError::IncompleteDownload` naming the asset and both sizes.
- The helper now takes the whole `&ReleaseAsset` rather than a URL, so the size checked against is always the one published for the body being fetched, and neither caller can pair a URL with an unrelated length or skip the check. Both call sites — `download_release_asset` and `upgrade_cli_in_place` — are covered without doing anything themselves.
- Exact byte counts are reported rather than human-readable sizes: rounding would render a small truncation as two identical-looking figures.
- The two byte-identical download-progress blocks in `release.rs` collapse into one `download_reporting_progress` helper, since the change had to edit both in lockstep.

`Content-Length` is deliberately not the authority — it describes whatever the server chose to send, so a truncating proxy or an interstitial page satisfies it with a header consistent with its own short body. The published asset size is the only independent expectation. Any `Content-Encoding` is a transport wrapper reqwest unwraps before this code sees the body, leaving exactly the stored bytes the published size refers to (verified against the live CDN: the API's `size` and the served `content-length` agree, and no encoding is applied).

## Verification

Each new test was confirmed to fail with the length check removed and to pass with it, and the two guard tests below pass in **both** variants — so no working download is diverted:

- `a_body_shorter_than_the_release_advertises_is_reported_as_a_failed_download` — regression test; empty, one byte, midway and one-byte-short bodies. Without the check these reproduce the reported symptom exactly (`Failed to extract archive: failed to iterate over archive`). Asserts both sizes are named, that the message does not blame the archive, and that nothing is left extracted. The mock serves a `Content-Length` consistent with its short body, as a truncating proxy would.
- `a_body_longer_than_the_release_advertises_is_rejected` — an over-long body is not the published asset either.
- `a_complete_asset_is_downloaded_and_extracted` — guard: a matching body installs exactly as before.
- `a_complete_body_that_is_not_an_archive_still_reports_extraction` — guard: the length check is not a stand-in for decoding, so a genuinely corrupt asset keeps its extraction error rather than being relabelled a network problem.

## Test plan

- [x] `make lint`
- [x] `make test`
- [x] New tests fail without the fix and pass with it (neuter-verified)
- [x] Verified against the live GitHub releases API that `asset.size` matches the served `content-length` and that no `Content-Encoding` is applied

## Performance impact

None on the download path: one integer comparison after the stream completes. The progress-block dedupe is a refactor with identical output.

## Notes / out of scope

- The accumulator is still `Vec::new()` though the exact size is now known up front; pre-sizing it needs a clamp policy for a server-supplied length and is left to its own change.
- `spicerack.rs` had the same *class* of bug (a body handed to a decoder unvalidated) and was fixed separately in #12119 by a different mechanism, deliberately: the registry has no authoritative expected size — its failure is a complete body of the wrong content, which a size check cannot catch, while a truncated gzip is genuinely undecodable and a decoder-rejection fallback cannot diagnose as a transfer failure.

### DIFF
--- a/bin/spice/src/github/mod.rs
+++ b/bin/spice/src/github/mod.rs
@@ -109,17 +109,32 @@ impl GitHubClient {
         })
     }
 
-    /// Download a file with progress tracking.
-    pub async fn download_with_progress<F>(
+    /// Download a release asset with progress tracking, verifying the body is the whole asset.
+    ///
+    /// The releases API publishes an exact `size` for every asset, so a body of any other
+    /// length is a failed download — a dropped connection, a truncating proxy, or an
+    /// interstitial page served with a `200`. Verifying it here means the bytes only ever
+    /// reach a decoder once they are known to be the right ones; otherwise a short body
+    /// surfaces as whatever the decoder makes of it, which reads as a corrupt archive
+    /// rather than as the transfer failure it is.
+    ///
+    /// The asset is taken whole rather than as a URL and a length so that the size checked
+    /// against is always the one published for the body being fetched. `Content-Length` is
+    /// deliberately not the authority: it describes whatever the server chose to send, so a
+    /// proxy substituting its own complete response satisfies it. Any `Content-Encoding` is
+    /// a transport wrapper the client unwraps before this sees the body, leaving the stored
+    /// bytes the published size refers to.
+    pub async fn download_asset<F>(
         &self,
-        url: &str,
+        asset: &ReleaseAsset,
         mut on_progress: F,
     ) -> Result<Vec<u8>, GitHubError>
     where
-        F: FnMut(u64, Option<u64>),
+        F: FnMut(u64),
     {
         use futures::StreamExt;
 
+        let url = &asset.browser_download_url;
         let mut request = self
             .client
             .get(url)
@@ -143,7 +158,6 @@ impl GitHubClient {
             });
         }
 
-        let total_size = response.content_length();
         let mut downloaded: u64 = 0;
         let mut data = Vec::new();
         let mut stream = response.bytes_stream();
@@ -154,7 +168,15 @@ impl GitHubClient {
             })?;
             downloaded += chunk.len() as u64;
             data.extend_from_slice(&chunk);
-            on_progress(downloaded, total_size);
+            on_progress(downloaded);
+        }
+
+        if downloaded != asset.size {
+            return Err(GitHubError::IncompleteDownload {
+                name: asset.name.clone(),
+                expected: asset.size,
+                received: downloaded,
+            });
         }
 
         Ok(data)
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12122, opened 2026-07-28
perf(mysql): fast row-image decoder for CDC change builds

sgrebnov:
## Summary

- add `TableMapRowDecoder`, prepared from `TableMapEvent`, to decode MySQL row images without rebuilding per-column metadata for every row
- use the prepared decoder in `MysqlChangeRows::build`, with a warning and fallback to the existing `mysql_common` rows walk when preparation fails (this is )
- add a generic real-binlog replay benchmark with a fast-vs-walk equivalence gate

By removing repeated metadata construction from the hot path, this reduces CPU/allocation pressure and creates room to move MySQL row-to-Arrow conversion into the pump thread, matching the PostgreSQL CDC direction.

## Performance

On a real MySQL binlog replay, production row decode improved from ~66 MiB/s to ~460 MiB/s, about 7x faster. A separate serial CDC e2e run including Cayenne writes improved from ~93 MiB/s to ~207 MiB/s, about 2.2x faster.

```bash
# 1. capture a binlog from a loaded MySQL (any CH-benCHmark run works)
docker cp <mysql-container>:/var/lib/mysql/binlog.000002 /tmp/
# 2. replay it
MYSQL_BINLOG_REPLAY_FILE=/tmp/binlog.000002 \
  cargo bench -p data_components --bench mysql_binlog_replay --features mysql
```

## Testing

- `cargo check -p data_components --features mysql`
- `cargo check -p data_components --bench mysql_binlog_replay --features mysql`

For deeper correctness coverage:
 - Preferred MySQL CDC e2e integration tests plus replaying real binlogs through the equivalence gate. Avoided simulating the binlog protocol in unit tests as those fixtures are very low-level and mostly duplicate `mysql_common` wire parsing details.
 - More integration tests with better data variety to be added next


crates/data_components/src/mysql_replication/changes.rs
```diff
@@ -146,19 +146,37 @@ impl ChangeRows for MysqlChangeRows {
     }
 
     fn build(self: Box<Self>) -> Result<ChangeBatch, ChangeBatchError> {
-        // Decode every buffered rows event into the transaction buffer using the
-        // SAME `buffer_rows_event` the per-dataset path uses (it records the
-        // per-row op metrics as it decodes), then build one Arrow batch.
+        // Decode every buffered rows event into the transaction buffer, then
```
bjchambers, line 149: Just to verify -- does this do the decoding on the binlog reader, or on the per-table consumer? In Postgres we shifted the decoding to the per-table side to help parallelize them.

sgrebnov: Confirm this PR keeps the decode on the per-table consumer (MysqlChangeRows::build, which runs on the per-dataset apply path).

bjchambers (approved):

PR #12148, opened 2026-07-29
task: (Search): `refresh_mode: full` does not delete old entries. (fixes #12145)

### LINKED ISSUE #12145: (Search): `refresh_mode: full` does not delete old entries.
## Describe the bug
When a dataset has an external index (e.g. S3Vectors, Elasticsearch), and is configured with an acceleration of type `refresh_mode: full`, rows that are no longer in the dataset are not removed from the external index.

```markdown
@t=0, dataset = [a, b, c, d, e], acceleration=[],              index=[],                    (initial)
@t=1, dataset = [a, b, c, d, e], acceleration=[a, b, c, d, e], index=[a, b, c, d, e],       (full refresh).
@t=2, dataset = [a, b, c, f, g], acceleration=[a, b, c, d, e], index=[a, b, c, d, e],       (changes to underlying dataset).
@t=3, dataset = [a, b, c, f, g], acceleration=[a, b, c, f, g], index=[a, b, c, d, e, f, g], (2nd full refresh).
```

Current abstraction does not handle difference in append (cannot remove anything from acceleration/index), and full (should be a new set of data in acceleration/index)

```rust
/// Compute the index - if the index data is represented in the batch itself (i.e. a vector
/// "*_embedding" column) then modify the provided batches to include the computed column.
async fn compute_index(&self, batches: Vec<RecordBatch>) -> Result<Vec<RecordBatch>> {
    Ok(batches)
}

/// Called before data is written via the [`TableSink`] path (full refresh or append).
///
/// Default is a no-op. Implementations use this to prepare external index state for a
/// bounded write window. Not called for CDC writes.
async fn on_write_start(&self) -> Result<()> {
    Ok(())
}

/// Called if a [`TableSink`] write fails after [`Index::on_write_start`] ran.
///
/// Default is a no-op. Implementations use this to restore temporary external index
/// settings when a refresh or append fails before [`Index::on_write_complete`] can run.
async fn on_write_failed(&self) -> Result<()> {
    Ok(())
}
```

Naive method would be to delete index (or all data) for full refreshes before new indexing. Alternatively, a hidden refresh version number could be used for soft deletes. Neither optimal.

Jeadie:
Resolves #12145.

See the issue thread for the design proposal posted by the investigation step.

### DIFF
--- a/.subagent-settings.json
+++ b/.subagent-settings.json
@@ -1,0 +1,53 @@
+{
+  "permissions": {
+    "allow": [
+      "Read",
+      "Grep",
+      "Glob",
+      "Edit",
+      "Write",
+      "Bash(rg:*)",
+      "Bash(ls:*)",
+      "Bash(cat:*)",
+      "Bash(git status:*)",
+      "Bash(git diff:*)",
+      "Bash(git log:*)",
+      "Bash(gh issue comment:*)",
+      "Bash(gh issue view:*)"
+    ],
+    "deny": [
+      "Read(~/.ssh/**)",
+      "Read(~/.aws/**)",
+      "Read(~/.config/**)",
+      "Read(~/Github/**/.env)",
+      "Read(~/Github/**/.env.*)",
+      "Read(~/Github/**/secrets/**)",
+      "Read(~/Github/**/*.key)",
+      "Read(~/Github/**/*.pem)",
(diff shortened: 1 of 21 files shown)


Jeadie:
should be reconsidered once #11961  is in trunk.

PR #12152, opened 2026-07-29
Prepare for v2.1.2 release

sgrebnov:


claudespice:
`Check Rust Licenses` is failing on a lockfile detail rather than on a licence, and it is fixable with one command.

The `advisories` stage errors twice with `error[yanked]: detected yanked crate (try cargo update -p spin)` ([run 30470941757](https://github.com/spiceai/spiceai/actions/runs/30470941757/job/90640687198)):

| `Cargo.lock` line | version | reached via |
|---|---|---|
| 1402 | `spin 0.9.8` | `heapless 0.7.17` → `postcard` |
| 1403 | `spin 0.10.0` | `crc-fast 1.10.0` → `aws-smithy-checksums` → `aws-sdk-s3` |

Both were yanked from crates.io after this branch's lockfile was written. `trunk` is already on the successor versions — `spin 0.9.9` and `0.10.1`, neither of which is yanked — so this is specific to this branch's lockfile, and the merge queue is not affected.

The targeted fix:

```bash
cargo update -p spin
```

That should move both entries to `0.9.9` / `0.10.1` and touch nothing else. To be clear about the alternative: this PR targets `release/2.1`, **not** `trunk`, so do not merge `trunk` in to pick up its lockfile — `cargo update -p spin` on this branch is the correct and much smaller change.

Everything else on the current head is either passing or still queued — the `Rust Lint` failure from the earlier head is not reproduced on `923c2c72` yet, since that run has not started. Worth re-checking once the queue drains; the hygiene/required-check backlog is tracked in #12162 and #12170.

krinart (approved):

Issue #12160, opened 2026-07-29
DuckDB pool sizing and replace_file validation ignore accelerated views


claudespice:
## Describe the bug

Two `DuckDB` accelerator helpers decide something about a `DuckDB` instance by walking `app.datasets` only, even though a **view** can carry its own `acceleration` block and joins the very same instance. Both are the same shape as the memory-budget adapter fixed in #12123, but each changes behavior in a way that did not belong in that fix, so they are tracked here.

### 1. `get_num_accelerating_datasets` under-sizes the connection pool

`crates/runtime/src/dataaccelerator/duckdb.rs` counts how many components share an instance and feeds that to `get_pool_max_size` (and `get_pool_min_idle` through it):

```rust
for spicepod_ds in &app.datasets {
```

A view accelerated onto the same `duckdb_file` — or onto the shared in-memory instance — is not counted, so an instance serving 1 dataset and 3 views is sized as if it served 1 dataset. The pool is then under-provisioned for its real concurrency and queries queue on connection acquisition.

This walk predates #11985. Changing it resizes connection pools on existing deployments, which is why #12123 left it out.

### 2. `validate_replace_file_peers` misses a view peer

Same file, `crates/runtime/src/dataaccelerator/duckdb.rs`. The check rejects `on_full_refresh: replace_file` alongside a peer using `refresh_mode: snapshot` on the *same* `DuckDB` file, because both replace that file out-of-band and refreshes then fail intermittently. It also walks `app.datasets` only, so a **view** using `refresh_mode: snapshot` on that file is not seen and the contradiction is reported only as intermittent refresh failures at runtime — exactly what the up-front check exists to avoid.

Widening this check makes the runtime reject configurations that load today, so it wants its own change with its own release note.

## Expected behavior

Both decisions consider every component that joins the instance, dataset or view — the same instance identity #12123 established for the memory budget (one instance per distinct resolved file path, one shared in-memory instance). `DuckDBAccelerator::spicepod_duckdb_file_path` already takes the acceleration block alone, so it resolves a view's path with no further change.

## Additional context

[...]

PR #12158, opened 2026-07-29
fix(lint): remove unused DuckDBTableWriter import in duckdb accelerator tests

sgrebnov:
Removes an unused `DuckDBTableWriter` import in the `duckdb` accelerator test module that fails `-D warnings` on trunk. The test that needs the type references it via its fully-qualified path, so the import was dead.

PR #12161, opened 2026-07-29
fix(duckdb): count accelerated views in the coordinated memory budget (fixes #12123)

### LINKED ISSUE #12123: DuckDB memory budget ignores accelerated views, so a view-only pod gets no coordination at all
## Describe the bug

The coordinated DuckDB memory budget added in #11985 sizes itself from `app.datasets` only. A **view** can carry its own `acceleration` block and creates a real DuckDB instance just like a dataset does, so every DuckDB-accelerated view is invisible to the budget — the exact over-commit the feature exists to prevent.

`duckdb_budget_inputs` (`crates/runtime/src/builder.rs`) iterates `for dataset in &app.datasets` and never looks at `app.views`. Two consequences:

**1. A pod whose DuckDB accelerators are all on views gets no coordination at all.**

With no DuckDB *datasets*, `num_unset_instances == 0 && num_explicit_instances == 0`, so `has_duckdb_instances` is false and the builder takes the early-out: `publish_duckdb_budget(0, 0)` and `query_pool_cap_bytes = None`. `duckdb_auto_memory_limit_option()` then returns `None` (0 ⇒ `None`), so no `SET memory_limit` is applied. Each view's instance runs at DuckDB's own ~80%-of-host-RAM default, stacked on the 90% query pool — silently, with no warning, because the planner never ran.

```yaml
views:
  - name: sales_summary
    sql: SELECT ... FROM ...
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
```

**2. Mixed pods under-count the divisor.**

With N DuckDB datasets and M DuckDB views, `per_instance_cap_bytes` is `duckdb_pool_total / N`. Every un-limited instance applies that published cap — views included, since the accelerator reads the same global — so the applied aggregate ceiling is `(N + M) / N` times the planned budget.

`duckdb_reservation_bytes` is under-stated for the same reason, and it is what the Cayenne in-memory CDC tier sampler (`crates/runtime/src/datafusion/mod.rs`) and the query-pool carve read to stay out of DuckDB's reserved room, so a co-resident Cayenne tier floats into memory DuckDB will actually use.

A view's explicit `duckdb_memory_limit` is also missing from `sum_explicit_bytes`, and a view sharing an instance with a dataset is never counted toward `has_mixed_instance`.

## Expected behavior

The budget counts every DuckDB instance in the app, whether the acceleration is declared on a dataset or on a view — same instance identity (one per distinct resolved file path, one shared in-memory instance), same explicit/un-limited classification.

## Additional context

The runtime already knows views make DuckDB instances: the same PR (#11985) extended `AccelerationSource::initialized_sources` with `get_initialized_views` under `#[cfg(feature = "duckdb")]` precisely so `instance_has_explicit_limit_sibling` (`crates/runtime/src/dataaccelerator/duckdb.rs`) would see a view sibling. Only the budget planner's input adapter was left dataset-only, so the two disagree about which instances exist.

Separately and out of scope here: `get_num_accelerating_datasets` (`crates/runtime/src/dataaccelerator/duckdb.rs`), which sizes the connection pool by how many datasets share an instance, walks `app.datasets` only for the same reason. That one predates #11985 and changing it resizes pools on existing deployments, so it is worth its own change.

claudespice:
## Summary

The coordinated DuckDB memory budget sized itself from `app.datasets` alone. A **view** can carry its own `acceleration` block and creates a real DuckDB instance exactly as a dataset does, so every DuckDB-accelerated view was invisible to the budget — the exact over-commit the budget exists to prevent.

`duckdb_budget_inputs` (`crates/runtime/src/builder.rs`) iterated `for dataset in &app.datasets` and never looked at `app.views`, with two consequences:

**1. A pod whose DuckDB accelerators are all on views got no coordination at all.** With no DuckDB *datasets*, `num_unset_instances == 0 && num_explicit_instances == 0`, so the builder took its early-out: `publish_duckdb_budget(0, 0)` and `query_pool_cap_bytes = None`. `duckdb_auto_memory_limit_option()` then returned `None`, so no `SET memory_limit` was applied and each view's instance ran at DuckDB's own ~80%-of-host-RAM default, stacked on the 90% query pool. Silently — the planner never ran, so there was no warning either.

```yaml
views:
  - name: sales_summary
    sql: SELECT ... FROM ...
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
```

**2. Mixed pods under-counted the divisor.** With N DuckDB datasets and M DuckDB views, `per_instance_cap_bytes` was `duckdb_pool_total / N`, but every un-limited instance applies that published cap — views included, since the accelerator reads the same global — so the applied aggregate ceiling was `(N + M) / N` times the planned budget. `duckdb_reservation_bytes` was under-stated the same way, and it is what the Cayenne in-memory CDC tier sampler and the query-pool carve read to stay out of DuckDB's reserved room.

The runtime already knew views make DuckDB instances: `AccelerationSource::initialized_sources` includes `get_initialized_views` under `#[cfg(feature = "duckdb")]` precisely so `instance_has_explicit_limit_sibling` sees a view sibling. Only the budget planner's input adapter was dataset-only, so the two disagreed about which instances exist.

Fixes #12123

## Changes

- `duckdb_budget_inputs` iterates `app.datasets` **and** `app.views`, chaining `(name, acceleration)` pairs through the identical grouping and classification logic. A view therefore shares instance identity with a dataset on the same file, joins the one shared in-memory instance in `mode: memory`, contributes its explicit `duckdb_memory_limit` to `sum_explicit_bytes`, and marks an instance `has_mixed_instance` when it is un-limited alongside an explicitly-limited sibling.
- `DuckDBAccelerator::spicepod_dataset_duckdb_file_path(&Dataset)` → `spicepod_duckdb_file_path(&Acceleration)`. The resolved path depends only on the acceleration block, so one helper now serves both datasets and views. Behavior-preserving for the two existing callers (`get_num_accelerating_datasets`, `validate_replace_file_peers`), which each already hold the acceleration block they were reaching through the dataset for.
- Doc comments on `DuckDbBudgetInputs` and the adapter updated — it is no longer "a thin adapter over `app.datasets`".

### Deliberately out of scope

Sibling `app.datasets`-only walks were reviewed and left alone, filed as #12160: `get_num_accelerating_datasets` (sizes the connection pool; predates this feature and changing it resizes pools on existing deployments), `validate_replace_file_peers` (widening it starts rejecting configurations that load today), and the two Cayenne walks in `builder.rs` (different subsystem, and a view has no CDC source).

Catalogs are excluded on purpose and the adapter's doc comment says so: `Catalog::acceleration` is a `CatalogAcceleration`, a distinct type whose engine enum admits only `Cayenne`, so a catalog cannot declare a DuckDB instance.

#12160 also carries the deeper cleanup this review surfaced — an `App::accelerations()` helper so "which component kinds declare an acceleration" lives in one place instead of being hand-chained in several. That touches `crates/app` and three unrelated call sites, so it is not folded into this fix; the adapter's doc comment records why the chain is hand-rolled here (it runs pre-init against the Spicepod, mirroring the `AccelerationSource` trait that is already kind-agnostic once components exist).

## Test plan

- New `test_duckdb_budget_inputs_counts_accelerated_views` in `crates/runtime/src/builder.rs` covers the view-only pod (1 instance where trunk reported 0, including the `unset_instance_labels` entry the warning prints), the mixed pod (2 instances), a view sharing a dataset's file (1 instance, not 2), a memory-mode view joining the shared in-memory instance, a view's own `duckdb_memory_limit` reaching `sum_explicit_bytes`, an un-limited view making an explicitly-limited instance `has_mixed_instance`, and the ignored cases (Arrow-engine view, `enabled: false`, no acceleration block). Every view assertion fails on trunk.
- The existing `test_duckdb_budget_inputs_groups_by_instance` is unchanged in behavior; the two tests now share one set of fixtures (`duckdb_acceleration` / `duckdb_ds` / `duckdb_view` / `budget_inputs_for`) rather than each defining its own app-construction path.
- `make lint`
- `make test`

## Checklist

- [ ] Tests added for the regression and its edge cases
- [ ] `make lint` passes
- [ ] `make test` passes
- [ ] No behavior change for pods without accelerated views

### DIFF
--- a/crates/runtime/src/accelerator_memory_budget.rs
+++ b/crates/runtime/src/accelerator_memory_budget.rs
@@ -125,22 +125,24 @@ pub fn duckdb_total_reservation_bytes() -> u64 {
 }
 
 /// A deduped-by-instance summary of the `DuckDB` accelerators in an app. Built by a
-/// thin adapter over `app.datasets` (see `builder::duckdb_budget_inputs`); the
-/// planner takes only these numbers so it stays pure and testable.
+/// thin adapter over `app.datasets` and `app.views` — either can carry an
+/// `acceleration` block, and every `DuckDB` instance they declare is counted here
+/// (see `builder::duckdb_budget_inputs`). The planner takes only these numbers so it
+/// stays pure and testable.
 #[derive(Debug, Clone, Default, PartialEq, Eq)]
 pub struct DuckDbBudgetInputs {
     /// Distinct `DuckDB` instances with NO explicit `duckdb_memory_limit` on any of
-    /// their datasets.
+    /// the datasets or views sharing them.
     pub num_unset_instances: u32,
     /// Distinct `DuckDB` instances with an explicit `duckdb_memory_limit`.
     pub num_explicit_instances: u32,
-    /// Sum of the explicit ceilings — one per explicit instance (the max value if
-    /// datasets on the same instance disagree).
+    /// Sum of the explicit ceilings — one per explicit instance (the max value if the
+    /// components on the same instance disagree).
     pub sum_explicit_bytes: u64,
     /// Some instance has an inconsistent `duckdb_memory_limit` across the datasets
-    /// that share it — mixed set/unset, or different explicit values. `DuckDB`'s
-    /// `memory_limit` is per-instance (last dataset created wins), so the effective
-    /// limit is ambiguous — surfaced in the warning.
+    /// and views that share it — mixed set/unset, or different explicit values.
+    /// `DuckDB`'s `memory_limit` is per-instance (last one created wins), so the
+    /// effective limit is ambiguous — surfaced in the warning.
     pub has_mixed_instance: bool,
     /// Human labels (in-memory / file path) of the un-limited instances, for the warning.
     pub unset_instance_labels: Vec<String>,
--- a/crates/runtime/src/builder.rs
+++ b/crates/runtime/src/builder.rs
@@ -1187,26 +1187,37 @@ fn estimate_cayenne_cdc_reservation_bytes(
 /// Deduped-by-instance summary of the `DuckDB` accelerators in `app`, for the
 /// coordinated memory budget ([`crate::accelerator_memory_budget::plan`]).
 ///
-/// Groups datasets by `DuckDB` instance identity — one per distinct resolved file
-/// path, plus a single shared key for all memory-mode datasets (mirroring the
-/// fork's `DbInstanceKey`) — and classifies each instance as explicit (some
-/// dataset sets `duckdb_memory_limit`, taking the max) or un-limited. Imperfect
-/// path canonicalization only ever OVER-counts instances, yielding smaller, safer
-/// caps.
+/// Groups accelerations by `DuckDB` instance identity — one per distinct resolved
+/// file path, plus a single shared key for all memory-mode accelerations (mirroring
+/// the fork's `DbInstanceKey`) — and classifies each instance as explicit (something
+/// on it sets `duckdb_memory_limit`, taking the max) or un-limited. Imperfect path
+/// canonicalization only ever OVER-counts instances, yielding smaller, safer caps.
+///
+/// Covers both `app.datasets` and `app.views`: a view carries its own `acceleration`
+/// block and creates a `DuckDB` instance exactly as a dataset does, so an instance
+/// the budget cannot see is an instance left at `DuckDB`'s own ~80%-of-RAM default —
+/// the over-commit this budget exists to prevent. Catalogs are excluded deliberately:
+/// they carry `CatalogAcceleration`, whose engine enum admits only Cayenne.
+///
+/// This enumerates component kinds by hand because it runs *before* initialization,
+/// against the Spicepod. It is the pre-init mirror of the `AccelerationSource` trait,
+/// which datasets and views both implement and which is therefore already
+/// kind-agnostic once components exist — the authority after init, but unavailable
+/// to the builder here.
 #[cfg(feature = "duckdb")]
 fn duckdb_budget_inputs(
     app: Option<&Arc<app::App>>,
 ) -> crate::accelerator_memory_budget::DuckDbBudgetInputs {
     use crate::accelerator_memory_budget::DuckDbBudgetInputs;
 
-    /// Per-instance aggregation while grouping datasets by `DbInstanceKey`.
+    /// Per-instance aggregation while grouping accelerations by `DbInstanceKey`.
     #[derive(Default)]
     struct InstanceAgg {
         explicit_max: Option<u64>,
         has_unset: bool,
-        /// Datasets sharing this instance set DIFFERENT explicit `duckdb_memory_limit`
-        /// values. Since the setting is per-instance (last dataset created wins), the
-        /// effective limit is ambiguous — surfaced in the warning.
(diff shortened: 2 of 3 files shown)


claudespice:
@copilot review

claudespice:
Independently verified the regression test on this diff (I reached the same fix in a parallel session before spotting this PR, so the verification is free — closing my duplicate branch unpushed).

Neutering the fix (dropping the `app.views` chain, everything else unchanged) and re-running `cargo test -p runtime --lib --features duckdb duckdb_budget_inputs`:

```
test builder::test::test_duckdb_budget_inputs_groups_by_instance ... ok
test builder::test::test_duckdb_budget_inputs_counts_accelerated_views ... FAILED
  assertion `left == right` failed at builder.rs:1785
    left: 0
   right: 1
```

So the new test fails on the old code at exactly the reported symptom — a view-only pod reporting zero DuckDB instances, which is what makes the builder take its early-out — and the pre-existing dataset-only test passes in both variants, i.e. no dataset-only behavior changed. With the fix applied both pass (2 passed; 0 failed).

Two things confirmed from source while reviewing, in case they're useful for the review:

- The helper migration `spicepod_dataset_duckdb_file_path(&Dataset)` → `spicepod_duckdb_file_path(&Acceleration)` is exactly behavior-preserving at both existing call sites: in `get_num_accelerating_datasets` and `validate_replace_file_peers` the `acceleration` now passed is the one already destructured from the very component that used to be passed in.
- `Dataset` and `View` are the only components carrying the `Acceleration` type, so datasets + views is the complete set here — catalogs carry `CatalogAcceleration`, whose engine enum admits only Cayenne.

The `cayenne_configured` arm that #12160 lists as unconfirmed is confirmed and split out as #12164.

claudespice:
Heads up before anyone reads the upcoming `signoff` failure as a defect in this PR: **it is a stale-base failure, not a code failure.**

This branch contains trunk `b35673914` (#12139), which left an unused `use datafusion_table_providers::duckdb::write::DuckDBTableWriter;` inside the `#[cfg(test)]` block at `crates/runtime/src/dataaccelerator/duckdb.rs`. It sits in a test-only `use`, so it trips **only** clippy's second `--tests` pass — i.e. exactly the pass `make lint-rust` runs. It broke every merge-queue batch repo-wide for 11h39m (`Rust Lint` failed 6/6 in sampled `merge_group` runs).

Confirmed present on this branch's head, in a file this PR never touches:

```
1971:    use datafusion_table_providers::duckdb::write::DuckDBTableWriter;
```

So signoff run [`30453278011`](https://github.com/spiceai/spiceai/actions/runs/30453278011) could only ever fail workspace lint, and even a hypothetical pass would have attested content that the merge queue rejects. **I cancelled it** rather than let it hold a scarce `spiceai-macos` runner for several more hours — no branch changes, no push.

**Trunk is fixed as of `63d397afd` (#12158, merged 16:13:35Z).** The remedy is the normal one, in this order:

1. merge the now-fixed `trunk` into this branch (this also clears `BEHIND`),
2. push,
3. re-dispatch the sign-off: `gh workflow run signoff.yml -f branch=fix/12123-duckdb-budget-views`.

Doing it in that order costs one sign-off cycle instead of two. I have deliberately not pushed to this branch — it is checked out in another working tree, so whoever owns that tree should do the merge.

lukekim (approved):

PR #12163, opened 2026-07-29
fix(ci): run the PR hygiene check on a GitHub-hosted runner

### LINKED ISSUE #12162: A required PR check is ~9h behind repo-wide: enforce-pull-with-spice is pinned to the saturated macOS pool
### Description

The required hygiene check `enforce-pull-with-spice` is running ~9 hours behind on every open PR, because a ~20-second GitHub-API-only action is pinned to the saturated `spiceai-macos` self-hosted pool.

Measured at 2026-07-29T13:10Z on `spiceai/spiceai`:

- **41** `pull_request_target` runs of `Enforce Pulls with Spice` are `queued`; the oldest was created at **04:09Z** (~9h earlier) and has never been assigned a runner (`runnerName: none`, `in_progress` count for that event: **0**).
- The most recently *completed* `pull_request_target` run was created 04:07Z and finished **12:01Z** — i.e. ~8h of queue time for a job whose actual work is a few seconds of GitHub API calls.
- Affected PRs span at least 19 branches, including `sgrebnov/2.1.2-prepare` (the v2.1.2 release PR itself), several `peasee/*` and `ben/*` branches, dependabot bumps, and mine.
- The same workflow's `merge_group` runs do complete (35/35 success in the sampled window), so the merge queue still drains — but the on-PR check board reads pending/red repo-wide for hours.

### Root cause

`.github/workflows/enforce-pulls-with-spice.yml` sets:

```yaml
jobs:
  enforce-pull-with-spice:
    runs-on: spiceai-macos
```

The only step is `spiceai/pulls-with-spice-action@v2.0.0`, whose `action.yml` declares `using: 'node24'` — a pure JavaScript action that makes GitHub API calls. There is no checkout, no build, and no macOS-specific dependency, so nothing about the job requires the macOS pool. Meanwhile ~20 other jobs in the repo target `spiceai-macos` (darwin builds, quickstart/E2E matrices), and the release-qualification benchmark sweep saturates it — so a trivial metadata linter queues behind hours of heavy work.

### Suggested fix

Move the job to a GitHub-hosted Linux runner. `ubuntu-latest` / `ubuntu-24.04` are already used in ~35 places in this repo, so no new runner capability is needed.

This also improves the security posture of what is currently the repo's only `pull_request_target` workflow: running `pull_request_target` on self-hosted runners is the classic risk pattern, and this job has no reason to sit on a persistent host.

### Notes / non-goals

- **Not proposing to restore the `concurrency` block.** It was removed deliberately in 4e06c2c28, and `cancel-in-progress: true` on a *required* check leaves cancelled runs that never report success. The duplicate-run amplification is real (up to 6 queued runs for a single branch, because the action's own `auto_label`/`auto_assign_author` re-triggers the `labeled`/`assigned` events it listens to), but it is a separate discussion and not required to fix the starvation.
- 11 runs from 2026-05-26 / 2026-06-05 are permanently stuck in `queued` and unrelated to the current backlog; they inflate any naive "oldest queued run" measurement and should be excluded when measuring (or cancelled separately).

claudespice:
Fixes #12162.

## What was wrong

`enforce-pull-with-spice` is a **required** check, and it was pinned to the `spiceai-macos` self-hosted pool:

```yaml
jobs:
  enforce-pull-with-spice:
    runs-on: spiceai-macos
```

Its only step is `spiceai/pulls-with-spice-action@v2.0.0`, whose `action.yml` declares `using: 'node24'` — a pure JavaScript action that makes GitHub API calls. There is no checkout, no build, and no macOS-specific dependency. Meanwhile ~20 other jobs target `spiceai-macos` (darwin builds, quickstart and E2E matrices), and the release-qualification benchmark sweep saturates it.

The result is a few seconds of real work queueing behind hours of heavy work. Measured at 2026-07-29T13:10Z:

- **41** `pull_request_target` runs `queued`; the oldest created **04:09Z** (~9h) and never assigned a runner (`runnerName: none`; `in_progress` for that event: 0).
- The most recent *completed* run was created 04:07Z and finished **12:01Z** — ~8h of pure queue time.
- At least 19 branches affected, including `sgrebnov/2.1.2-prepare` (the v2.1.2 release PR itself), several `peasee/*` and `ben/*` branches, and dependabot bumps.

The same workflow's `merge_group` runs do complete, so the merge queue still drains — but the on-PR check board reads pending repo-wide for hours.

## The change

One line: `runs-on: spiceai-macos` → `ubuntu-24.04`, plus a comment recording why the job does not belong on the self-hosted pool. `ubuntu-24.04` is already the most common runner label in this repo, so no new runner capability is required.

This also improves the posture of what is currently the repo's only `pull_request_target` workflow — running `pull_request_target` on self-hosted runners is the classic risk pattern, and this job has no reason to sit on a persistent host.

## Deliberately not included

**Restoring the `concurrency` block.** It was removed on purpose in 4e06c2c28, and `cancel-in-progress: true` on a *required* check leaves cancelled runs that never report success. The duplicate-run amplification is real — up to 6 queued runs for one branch, because the action's own `auto_label`/`auto_assign_author` re-triggers the `labeled`/`assigned` events the workflow listens to — but it is a separate discussion and not needed to fix the starvation.

Separately, 11 runs from 2026-05-26 / 2026-06-05 are permanently stuck in `queued`. They are unrelated to this backlog but inflate any naive "oldest queued run" measurement; they need cancelling by someone with the permissions to do it.


claudespice:
@copilot review

claudespice:
Additional evidence for this change, from measuring the queue today. Beyond the raw wait, the `spiceai-macos` placement has a second effect I hadn't seen before: **a redundant duplicate run can re-open a required check that already passed on the same commit.**

Concrete case — #12099 (approved, no failing checks, otherwise the most enqueueable PR on the board). Its `enforce-pull-with-spice` **already succeeded on its current head `4ed004bf`**:

| check run | created | state |
|---|---|---|
| `90387535546` | 2026-07-28T19:34:59Z | `completed/success` (7s) |
| `90527129978` | 2026-07-29T08:49:18Z | `queued` — same SHA |

Because branch protection evaluates the *most recent* check run for a name, that second run — created by the action's own `auto_label` / `auto_assign_author` steps re-firing the `labeled`/`assigned` events the workflow triggers on — puts the PR back to `required_ci_pending` and holds it there for as long as the queue is deep. The check has already passed; the PR is blocked on a duplicate of it.

Scale of the duplication today: #12166 had **6** queued runs at one SHA, #12165 **5**, #12161 and this PR **4** each.

That is not an argument for adding a `concurrency:` block — `git log` shows one was removed from this workflow deliberately, and on a required check `cancel-in-progress` would trade this problem for a worse one (cancelled runs are not successes). It *is* an argument for this PR: on a GitHub-hosted runner the duplicates cost ~20s each and clear immediately, so the re-pend window is seconds instead of hours.

Queue measurements at 16:1xZ, for the record: **56 `pull_request_target` runs queued**, oldest created 04:35:50Z, drain ~2.2/hour, zero in progress at several sample points, while `merge_group` runs of the same workflow completed in ~5 minutes throughout. The action is `using: 'node24'` — it makes GitHub API calls and needs nothing from macOS.

Separately, I reclaimed 13 queued runs of this workflow that could never satisfy any check — 8 whose `headSha` is no longer their branch's head, and 5 belonging to #12158 after it merged. No PR's check status moved as a result (verified against current-SHA runs on #12152, #12153, #12099).

claudespice:
Follow-up measurement that answers the open question from #12170 — **does this change become redundant once the orphaned-batch-run leak is fixed?** Measured today: no, and the numbers now separate the two causes cleanly.

The leak side has been cleared and has stayed cleared. Orphaned `merge_group` runs on deleted `gh-readonly-queue/*` branches: **0** right now, sustained over the last hour. Separately I cancelled **20 queued hygiene runs whose commit was no longer their PR's head** (they can never satisfy a required check, since branch protection evaluates the current head SHA) — on-PR queue depth **45 → 25**, with every affected PR retaining its current-head run and no check changing state.

So both sources of dead queue weight are gone. The hygiene check still is not running:

| measurement | value |
|---|---|
| `pull_request_target` runs queued | 27 |
| `pull_request_target` runs in progress | **0** |
| oldest queued run | created 10:52Z (**~7h30m**) |
| last successful run finished | 17:15Z (**~1h05m ago**) |
| concurrent `spiceai-macos` jobs in flight | 23, across 9 hosts |

The shape of the drain is the interesting part. The last four successes all finished in a tight burst at **17:14:40–17:15:25Z**, immediately after the orphaned runs were cancelled — having been created between 06:47Z and 10:52Z, i.e. after waiting **6.5–10.5 hours** each. Then the freed capacity was taken by newly-formed merge-queue batches (active `merge_group` runs went 13 → 25) and the hygiene queue went straight back to `in_progress=0`.

That is the argument for this PR, stated as a measurement rather than a prediction: **reclaiming dead capacity produces a burst, not a fix.** A 20-second checkout-free JS action has no way to win a FIFO race against multi-hour Rust jobs on a shared pool, so it will keep starving whenever the pool is busy — which is most of the time. Moving it to a GitHub-hosted runner takes it out of that race entirely, and it is the only one of the three changes that does.

The same-workflow contrast still holds and rules out the workflow itself being at fault: its `merge_group` runs are **33/36 successful** and finish in minutes, on the same `runs-on`. Only the on-PR event starves.

One caveat in fairness: **6 of the 23 in-flight macOS jobs are `Remote Sign-off` runs I dispatched**, so this queue pressure is partly self-inflicted on my side. That affects who is contributing to the contention, not whether a 20-second required check belongs on the contended pool.

Net effect on the board: #12157 and #11800 are both approved with no failing checks and are held only by this check being pending.

lukekim (approved):

Issue #12164, opened 2026-07-29
Cayenne compaction memory pool is not carved for a view-only Cayenne pod (builder reads app.datasets only)


claudespice:
## Describe the bug

The dedicated Cayenne compaction memory pool is carved only when a Cayenne acceleration block is found on a **dataset**. A **view** can carry its own `acceleration` block and is initialized through the same `DataAccelerator::init` path (`initialize_views_accelerators` in `crates/runtime/src/init/view.rs`), so a pod whose Cayenne acceleration lives only on views is treated as a non-Cayenne deployment.

`crates/runtime/src/builder.rs`:

```rust
let cayenne_configured = self.app.as_ref().is_some_and(|app| {
    app.datasets.iter().any(|dataset| {
        dataset.acceleration.as_ref().is_some_and(|accel| {
            accel.enabled && accel.engine.as_deref()
                .is_some_and(|engine| engine.eq_ignore_ascii_case("cayenne"))
        })
    })
});
```

`cayenne_configured` gates `compaction_memory_fraction`, which is both the signal `spiced` uses to bring up the dedicated compaction worker threads and the `cayenne_active` input to the query-memory calculation. With it false on a view-only Cayenne pod, no compaction pool is carved and the query pool is sized as if no Cayenne tier existed, so compaction competes for query memory it was supposed to have its own reservation for.

```yaml
views:
  - name: sales_summary
    sql: SELECT ... FROM ...
    acceleration:
      enabled: true
      engine: cayenne
```

This is the same omission as #12123 (the DuckDB memory budget read `app.datasets` only) in a sibling budgeting input. It was deliberately left out of that fix: unlike DuckDB-accelerated views — which have a test spicepod (`test/spicepods/tpch/sf1/accelerated/s3[parquet]-mixed[duckdb[ds]+duckdb[view]].yaml`) and are a first-class configuration — there is no test spicepod exercising Cayenne acceleration on a view, so newly carving a compaction pool and starting compaction threads for that shape wants its own validation rather than a speculative change.

`estimate_cayenne_cdc_reservation_bytes` in the same file is *correctly* dataset-only: it counts Cayenne tables with `refresh_mode: changes`, and view acceleration rejects every refresh mode except `full` (`ViewBuilder::try_from`), so no view can qualify.

## Expected behavior

Either:

1. Cayenne acceleration on a view is supported, in which case `cayenne_configured` should consider `app.views` too — with a test covering the view-only pod; or
[...]

PR #12169, opened 2026-07-29
fix(duckdb): size the connection pool for accelerated views too (fixes #12160)

### LINKED ISSUE #12160: DuckDB pool sizing and replace_file validation ignore accelerated views

claudespice:
## Summary

An accelerated **view** carries its own `acceleration` block and joins the same `DuckDB` instance as a dataset naming the same file — but the walk that sizes that instance's connection pool only looked at `app.datasets`:

```rust
for spicepod_ds in &app.datasets {
```

So an instance serving 1 dataset and 3 views was sized as if it served 1 dataset. `get_pool_max_size` takes `max(default_pool_size, component_count)`, so once the real count passes the default for the storage profile (10 locally, **4 on EBS**) the pool is under-provisioned for its actual concurrency and queries queue on connection acquisition. Nothing reports this — it looks like slow queries.

Same shape as the memory-budget adapter fixed in #12123, in a second place.

Fixes #12160

## Changes

- `get_num_accelerating_datasets` → `get_num_accelerating_components`, walking `app.datasets` **and** `app.views`. The name and the `get_pool_max_size` parameter follow, since a dataset is no longer the only thing counted.
- `spicepod_dataset_duckdb_file_path` → `spicepod_duckdb_file_path`, taking the acceleration block alone: the resolved path depends only on that block, so the same code resolves a view's instance with no view-specific path handling. Its one other caller (`duckdb_budget_inputs`) is updated.
- `validate_replace_file_peers` **keeps** its dataset-only peer walk, now with a comment saying why (below).

## What was checked and deliberately not changed

The issue also proposed widening `validate_replace_file_peers` to views. That would be a false rejection: `ViewBuilder::try_from` (`crates/runtime/src/component/view.rs:148`) rejects any `refresh_mode` other than `full` on a view, and `load_views` logs and skips a view that fails to build — so a view never holds `refresh_mode: snapshot` against a live `DuckDB` file, and a pod pairing an (already inert) snapshot view with a `replace_file` dataset would start failing to load. The reason is now a doc comment on the function rather than tribal knowledge.

The rest of the `app.datasets`-only walks in this class were checked too: `estimate_cayenne_cdc_reservation_bytes` is correctly dataset-only (it counts `refresh_mode: changes` tables, which a view cannot be), `cayenne_configured` is a real sibling defect tracked separately in #12164, and `cluster/partition/startup.rs` plus `http/v1/datasets.rs` already chain `app.views`.

## Performance impact

Pool sizing only — no change to any query plan or SQL. On a pod where accelerated views share a `DuckDB` instance, `connection_pool_size` is now `max(default, datasets + views on that instance)` instead of `max(default, datasets)`; the pool grows only where the real count exceeds the storage profile's default (10, or 4 on EBS). An explicit `connection_pool_size` param still wins outright, and instances with no accelerated views are unaffected.

## Test plan

- [x] `make lint`
- [x] `make test`
- [x] New unit tests in `crates/runtime/src/dataaccelerator/duckdb.rs`:
  - `accelerating_component_count_includes_views_on_the_same_file` — regression test: 1 dataset + 3 views on one file counts 5, and excludes a view on another file, a memory-mode view, a `sqlite` view on the same path, and a view with no acceleration. Also asserts the user-visible effect: the EBS pool is sized 5, where the dataset-only count of 2 stays clamped to the default 4. Fails on the old code (counts 2).
  - `accelerating_component_count_includes_memory_mode_views` — memory-mode views share the one in-memory instance and are counted.
  - `accelerating_component_count_without_components_is_one` — empty app still sizes for the component being created.

## Checklist

- [x] Root cause identified and fixed at the source, not worked around
- [x] Regression test fails on the old code
- [x] Sibling walks in the same class audited (see above)
- [x] No behavior change for pods without accelerated views

## Note for reviewers

`crates/runtime/src/dataaccelerator/duckdb.rs` is also touched by #12161 (the #12123 fix), which introduces the same `spicepod_duckdb_file_path` signature. Whichever lands first, the other resolves a small conflict in that one function — the two changes are complementary, not competing.


claudespice:
@copilot review

claudespice:
Verification detail for reviewers — the regression tests were neuter-checked, not just run green.

Reverting only the `.chain(app.views…)` in `get_num_accelerating_components` (leaving everything else in place) fails exactly the two counting tests, with the pre-fix numbers:

```
accelerating_component_count_includes_views_on_the_same_file ... FAILED
  assertion `left == right` failed: views sharing the DuckDB file must be counted alongside datasets
    left: 2   right: 5
accelerating_component_count_includes_memory_mode_views ... FAILED
  assertion `left == right` failed: memory-mode views share the in-memory instance and must be counted
    left: 2   right: 4
test result: FAILED. 1 passed; 2 failed
```

`accelerating_component_count_without_components_is_one` passes either way by design — it guards the base case, not the fix. With the fix restored: `26 passed; 0 failed` for `cargo test -p runtime --features duckdb --lib dataaccelerator::duckdb`.

Two source confirmations behind the change:

- `duckdb_file_path` (datafusion-table-providers `core/src/duckdb.rs:316`) strips the `duckdb_` key prefix and returns the `open` param verbatim, doing no I/O and no normalization — so an acceleration block alone resolves the instance identity, and a view's block resolves it the same way a dataset's does.
- A view reaches `DataAccelerator::init` through `initialize_views_accelerators`, which is what puts it on the shared instance whose pool this sizes.

claudespice:
`Attestation` is red here because the sign-off run hit the GitHub Actions **6-hour job timeout**, not because anything failed. I've re-dispatched it.

Run [`30474092162`](https://github.com/spiceai/spiceai/actions/runs/30474092162) on `fix/12160-duckdb-view-blind-walks`:

| | |
|---|---|
| job started | 2026-07-29T17:14:10Z (runner assigned 4 min after dispatch — not starved) |
| job completed | 2026-07-29T23:14:32Z |
| elapsed | **6h00m22s** |
| conclusion | `cancelled`, **zero failed steps**, `run_attempt: 1` |

Worth spelling out because the signature is easy to misread: a 6-hour-timeout death and a `concurrency: cancel-in-progress` death both surface as `cancelled` with an empty failed-step list. What separates them is the ~6h00m job duration plus the absence of any sibling sign-off run on the same branch — there was only ever this one, so nothing cancelled it.

Fresh sign-off dispatched against the same head `c221cfae3` (PR head and remote ref both confirmed at that SHA, so the attestation will bind to the reviewed commit): [`30499401331`](https://github.com/spiceai/spiceai/actions/runs/30499401331), with `skip_targeted_lint=true` to shave the branch-scoped fail-fast pre-lint off the front — the full workspace `make lint-rust` and `nextest` still run, so nothing is skipped that the attestation depends on. Recent successful sign-offs on this pool have landed in 1h59m–5h, and contention is lower now than during the 17:14–23:14Z window, so a retry is the right move rather than a code change — there is nothing in the diff to fix.

No push, no branch update: the head must stay at `c221cfae3` for the new attestation to be worth anything.

lukekim (approved):

PR #12171, opened 2026-07-29
fix(runtime): count accelerated views when gating the Cayenne compaction pool (fixes #12164)

### LINKED ISSUE #12164: Cayenne compaction memory pool is not carved for a view-only Cayenne pod (builder reads app.datasets only)

claudespice:
## Summary

The dedicated Cayenne compaction memory pool was carved only when an enabled Cayenne acceleration was found on a **dataset**. A **view** carries its own `acceleration` block and is initialized through the same `DataAccelerator::init` path — `init::view::initialize_views_accelerators` resolves *any* engine from the accelerator registry and calls `init` on the view — so a pod whose Cayenne acceleration lives only on views was treated as a non-Cayenne deployment.

`cayenne_configured` gates `compaction_memory_fraction`, which is both the signal `spiced` uses to bring up the dedicated compaction worker threads and the `cayenne_active` input to the query-memory calculation. With it false on a view-only Cayenne pod, no compaction pool was carved and the query pool was sized as if no Cayenne tier existed, so compaction competed for query memory it was supposed to have its own reservation for.

```yaml
views:
  - name: sales_summary
    sql: SELECT ... FROM ...
    acceleration:
      enabled: true
      engine: cayenne
```

### The support question this issue asked first

#12164 asked whether Cayenne-on-a-view is supported (fix the budget) or unsupported (reject at load). The code answers it: **it is supported.** `ViewBuilder::try_from` validates view acceleration and rejects only a non-`full` `refresh_mode` and `refresh_sql` — there is no engine restriction — and `initialize_views_accelerators` then resolves the engine by name from the registry and initializes it, Cayenne included. `View` also implements `AccelerationSource`. So the acceleration really does come up; only the budgeting input was blind to it. This PR takes option 1.

This is the same omission as #12123 (the DuckDB memory budget read `app.datasets` only) in a sibling budgeting input, and uses the same `datasets.chain(views)` shape as its fix.

## Changes

- `crates/runtime/src/builder.rs`: extract the inline gate into a `cayenne_configured(app)` free function — mirroring the neighbouring `estimate_cayenne_cdc_reservation_bytes` / `duckdb_budget_inputs` adapters — and chain `app.views` alongside `app.datasets`.
- Documented why catalogs stay excluded (they carry `CatalogAcceleration`, a separate type) and why this adapter enumerates component kinds by hand (it runs pre-initialization, against the Spicepod, so the kind-agnostic `AccelerationSource` trait is not yet available).
- Recorded on `estimate_cayenne_cdc_reservation_bytes` why *it* is correctly dataset-only: it counts `refresh_mode: changes` tables and `ViewBuilder::try_from` rejects every view refresh mode except `full`, so no view can qualify. That asymmetry between the two neighbouring functions is otherwise easy to misread as a second instance of this bug.

### Bug-class sweep

The app-wide "is Cayenne configured" enumerations are exactly two: this one and `estimate_cayenne_cdc_reservation_bytes` (correctly dataset-only, above). Every other `Engine::Cayenne` / `eq_ignore_ascii_case("cayenne")` site in `crates/runtime` tests a single acceleration already in hand rather than walking the app, so none can exhibit this shape. `get_num_accelerating_datasets` (DuckDB connection-pool sizing) has the same dataset-only walk but is tracked separately in #12160 and is not touched here.

## Behavior change

A view-only Cayenne pod now carves a compaction pool and starts compaction worker threads, and its query pool shrinks by the compaction fraction — which is the coordination the budget exists to provide, and what the equivalent dataset-only pod already did. Deployments with no Cayenne acceleration, or with Cayenne on a dataset, are unaffected. The existing `cayenne_compaction_memory_fraction` and `dedicated_thread_pool: disabled` params remain the operator escape hatches.

## Test plan

- `cayenne_configured_counts_views_and_datasets` (new, `crates/runtime/src/builder.rs`) — regression test: Cayenne declared **only** on a view must gate the pool on. Fails on the pre-fix code (the view-only case returns `false`) and passes after. Also covers the dataset-only case (unchanged), a Cayenne view beside a non-Cayenne dataset, case-insensitive engine matching on the view arm, `enabled: false` on either kind, no-Cayenne-anywhere, the empty app, and `None`.
- `make lint`
- `make test`

## Checklist

- [x] Root cause identified and the support question decided from the code, not assumed
- [x] Regression test that fails before the fix and passes after
- [x] Sibling arms audited (bug-class sweep above)
- [x] No behavior change for non-Cayenne or dataset-Cayenne deployments
- [x] `make lint` and `make test` pass

Fixes #12164


claudespice:
@copilot review

claudespice:
### Verification

**Regression test passes on the fix** — `cargo test -p runtime --lib cayenne_configured`:

```
test builder::test::cayenne_configured_counts_views_and_datasets ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 1535 filtered out
```

**Neuter check** — dropping `.chain(app.views.iter()…)` from `cayenne_configured` (a real lib change, so the crate rebuilt) makes it fail on exactly the view-only assertion:

```
test builder::test::cayenne_configured_counts_views_and_datasets ... FAILED
panicked at crates/runtime/src/builder.rs:1772:9:
a view-only Cayenne pod must carve a compaction pool
test result: FAILED. 0 passed; 1 failed
```

`builder.rs:1772` is the `configured(vec![], vec![view_with("cayenne", true)])` assertion. The file was then restored; the pushed HEAD is the verified tree (working tree byte-identical to the commit after restoring).

**Base**: rebased onto trunk `63d397afd` — note `b35673914` carries the `DuckDBTableWriter` unused-import lint break that `63d397afd` (#12158) fixes, so a signoff on the older base would have failed in `duckdb.rs` on code this PR never touches.

**Overlap note**: #12169 also edits `crates/runtime/src/builder.rs`, but in the `duckdb_budget_inputs` / `spicepod_duckdb_file_path` region — it does not touch `cayenne_configured` (its own class-sweep table assigns that arm to #12164). No functional overlap; a textual conflict is unlikely but whichever lands second should re-verify.

lukekim (approved):

Issue #12111, opened 2026-07-28
CI: a PR that only changes clippy.toml, nextest.toml, or layers.toml lands with no Rust lint or tests at any stage


claudespice:
### What happens

Two independent gates decide whether a branch needs Rust validation, and both key off the same path list — which omits the config files the Rust gate itself consumes.

1. **PR stage.** `.github/workflows/pr.yml`'s *Fast-track branches with no Rust-affecting files* step auto-passes the required `Attestation` check when no changed path matches `(\.rs$|Cargo\.(toml|lock)$|rust-toolchain(\.toml)?$|\.cargo/)`. `branch_has_rust_changes` in `scripts/signoff` uses the identical pattern, so even a manual `make signoff` skips fmt, clippy, build, and tests for such a branch.
2. **Merge queue.** `check_changes` (`.github/actions/check-code-changes`) filters on `.github/**`, `crates/**`, `bin/**`, `test/**`, `tools/**`, `install/**`, `Cargo.toml`, `Cargo.lock`, `go.mod`, `go.sum`, `Makefile`. On a miss every step in `Rust Lint` and `Build and Test` is skipped and both report success.

Paths that change what the Rust gate does but match **neither** list:

| Path | What it controls |
| --- | --- |
| `.ci/clippy.toml` | the clippy config the gate actually uses (`CLIPPY_CONF_DIR=".ci"`, `Makefile:184`) |
| `clippy.toml` | root config (`too-many-lines-threshold = 200`) used by scoped/local clippy and rust-analyzer |
| `.config/nextest.toml` | nextest retries, slow-test timeouts, and test groups — including the deliberate `retries = 0` correctness overrides |
| `layers.toml`, `scripts/check_crate_layers.py` | inputs to the crate-layering guard the gate runs (`Makefile:182`) |
| `rustfmt.toml` / `.rustfmt.toml` | none today, but adding one would be invisible to both lists |

`Makefile` is a partial case: the merge-queue filter catches it, the PR-stage pattern does not — so a Makefile-only PR (the file that holds every `-Dclippy::…` flag) skips `Attestation` and is validated only in the queue.

### Why it matters

[...]

PR #11879, opened 2026-07-15
Move cargo advisory checks to scheduled workflow

Jeadie:
## Summary
- Run `cargo deny check advisories` in scheduled workflow, not a PR blocking action.
- For any advisory violation, create github issue and assign to`claudespice`.

PR #12182, opened 2026-07-29
fix(ci): make cargo_deny_advisories.yml parseable so the Rust advisory scan can run (fixes #12181)

### LINKED ISSUE #12181: CI: the daily Rust advisory scan has never run — cargo_deny_advisories.yml is unparseable YAML, and every push fails it
`.github/workflows/cargo_deny_advisories.yml` has an invalid YAML block scalar, so GitHub Actions cannot parse the file. The consequences are that the daily Rust dependency advisory scan has **never run** since it was introduced, and every push to the repository produces a failed workflow run.

## The defect

The `actions/github-script` step embeds JavaScript in a `script: |` block scalar whose indentation is set by its first line (12 spaces). One line inside that block sits at **column 0**:

```yaml
          script: |
            const fs = require('fs');
            ...
            const body = [
              marker,
'`cargo deny check advisories` found Rust dependency advisory violations.',   # <- column 0
              '',
```

A non-empty line less indented than the block scalar terminates the scalar, so the parser then reads that line as a new node inside a nested mapping:

```
$ ruby -ryaml -e 'YAML.load_file(".github/workflows/cargo_deny_advisories.yml")'
Psych::SyntaxError: could not find expected ':' while scanning a simple key at line 67 column 1
```

It is the only unparseable file of the 92 workflow/composite-action YAML files in `.github/`.

## Evidence that it has never worked

Introduced by #11879 (merged 2026-07-15 07:06Z). Since then:

```
$ gh api "repos/spiceai/spiceai/actions/workflows/313385733/runs?per_page=1" --jq .total_count
1389
$ gh api ".../runs?status=success" --jq .total_count
0
$ gh api ".../runs?event=schedule" --jq .total_count
0
```

**1389 runs, zero successes, zero scheduled runs.** Two further tells confirm this is a startup failure rather than a step failure:

- The workflow is registered under the *path* `.github/workflows/cargo_deny_advisories.yml`, not its declared `name: Check Rust Advisories` — GitHub never parsed the `name` key. It is the only workflow in the repository in that state.
- Every run has `event: push` even though the file declares only `schedule` and `workflow_dispatch`. Because the `on:` block cannot be read, GitHub cannot evaluate the triggers and raises a run for the event anyway. Each run has **zero jobs**, and `gh run view` reports "This run likely failed because of a workflow file issue."

## Impact

1. **No Rust dependency advisory scanning for 14 days.** `cargo deny check advisories` (RUSTSEC / CVE data) has not run once. A new advisory against any dependency in `Cargo.lock` would go unreported, and the "create or update advisory issue" reporting path has never executed. #11879 moved this check *out* of the PR gate and into this scheduled workflow, so nothing else covers it.
2. **A red failed run on every push, repo-wide.** 1389 so far, on `trunk`, on the `gh-readonly-queue/**` batch branches, and on every contributor branch. It blocks nothing, which is exactly why it went unnoticed — it is pure noise on the commit status of every branch, and it trains readers to ignore red runs.

## Why it stayed invisible, and the bug class

A workflow GitHub cannot parse fails *open*: it is not a required check, it reports no jobs, and its `paths`/`branches` filters cannot be honoured. Nothing in CI validates that the workflow files parse, so any syntax error in `.github/workflows/**` silently disables that workflow indefinitely. `make lint-rust` is change-filtered to Rust paths and would not cover a `.github/`-only change in any case (see #12111).

## Suggested fix

Indent line 67 to the block's indentation. Separately, add a check that every file in `.github/workflows/**` and `.github/actions/**` parses, gated on changes to those paths, so the next syntax error fails a PR instead of silently disabling a workflow.

claudespice:
## Summary (root cause)

`.github/workflows/cargo_deny_advisories.yml` embeds JavaScript in an `actions/github-script` `script: |` block scalar. One line of that JavaScript sat at **column 0**, and a non-empty line less indented than a block scalar terminates the scalar — so the parser then read a bare string where a mapping key belongs:

```
Psych::SyntaxError: could not find expected ':' while scanning a simple key at line 67 column 1
```

GitHub therefore never parsed the file. Two consequences, both invisible because a workflow GitHub cannot parse fails **open**:

1. **The daily Rust dependency advisory scan has never run.** Across 1389 runs since the workflow was introduced on 2026-07-15 there were **zero successes and zero scheduled runs**. `cargo deny check advisories` (RUSTSEC/CVE data over `Cargo.lock`) has not executed once, and the "create or update advisory issue" reporting path has never fired. #11879 moved this check out of the PR gate into this scheduled workflow, so nothing else covered it.
2. **A jobless failed run on every push, repo-wide** — on `trunk`, on every `gh-readonly-queue/**` batch branch, and on every contributor branch. It blocks nothing, which is why it went unnoticed for 14 days.

Corroborating tells: the workflow is registered under its *path* rather than its declared `name: Check Rust Advisories` (the only workflow in the repo in that state), and every run carries `event: push` even though the file declares only `schedule` and `workflow_dispatch` — because the unparseable `on:` block leaves GitHub unable to evaluate the triggers.

## Changes

**`.github/workflows/cargo_deny_advisories.yml`**

- Indent the offending line to the block scalar's indentation.
- Paginate the open-issue lookup in the same step. It read only the first page of open issues, so the `<!-- cargo-deny-advisories -->` marker issue falls off that page within days and a duplicate advisory issue would be opened on every run. This code path has never executed, so the fix activates it for the first time — worth fixing in the same change rather than shipping a workflow that immediately spams duplicates.

**The bug class — a workflow that cannot be parsed is silently disabled.** Nothing in CI validates that the definitions parse, and `make lint-rust` is change-filtered to Rust paths so it would not cover a `.github/`-only change in any case (see #12111). Added:

- `.github/scripts/check_workflow_yaml.py` — parses every `.github/workflows/*.y*ml` and `.github/actions/*/action.y*ml` (96 files today) and asserts each carries the keys GitHub needs to schedule it: a workflow needs an `on:` trigger block and a non-empty `jobs:` mapping; a composite action needs `runs:`. Stdlib + PyYAML only. Exits `2` rather than `0` when discovery finds nothing, so a wrong path can never read as "clean".
- `.github/scripts/test_check_workflow_yaml.py` — 18 unit tests.
- `.github/workflows/workflow_yaml_check.yml` — runs both on PRs and `trunk` pushes that touch `.github/workflows/**`, `.github/actions/**`, or either script. GitHub-hosted `ubuntu-24.04` on purpose: it is a sub-second check and must not queue behind the saturated self-hosted pool (#12162).

This follows the existing `github_release.py` + `test_github_release.py` + `release_script_test.yml` precedent: a pure checker split from its CI wrapper, with the logic under unit test.

## Test plan

- `make lint` — no Rust-affecting files in this change; `Attestation` auto-passes.
- `make test` — no Rust-affecting files in this change.
- `python .github/scripts/test_check_workflow_yaml.py -v` → **18 passed**. The suite includes `RepositoryTest`, which runs the checker over this repository's own `.github/` — so the guard is itself the regression test for the workflow fix.
- `python .github/scripts/check_workflow_yaml.py` → `OK: 96 GitHub Actions definitions parse and declare the required keys`.
- **Neuter check**: restoring the original mis-indented line makes the checker exit `1` with `cargo_deny_advisories.yml: is not valid YAML: could not find expected ':' at line 67`, and `RepositoryTest` fails. Restored and re-verified green on the pushed tree.
- The step's JavaScript was extracted from the parsed YAML and syntax-checked with `node --check` (wrapped in the async IIFE `github-script` supplies).
- **Live end-to-end, read-only**: pushing this branch produced **no run at all** for that workflow, while a branch pushed three minutes earlier still produced the usual `failure`. GitHub now parses the file at this ref and correctly sees that `push` is not one of its triggers — which is the fix working, observed against the real Actions scheduler without dispatching anything.

Expect the first real scheduled run (`20 5 * * *`) to report whatever advisories have accumulated over the 14 days the scan was dark; if there are any, the workflow opens its marker issue as designed.

## Checklist

- [x] Regression test that fails before the fix and passes after
- [x] Bug class addressed, not just the reported instance
- [x] Verified against the live GitHub Actions scheduler, read-only
- [x] No behavior change to any other workflow

Fixes #12181

### DIFF
--- a/.github/scripts/check_workflow_yaml.py
+++ b/.github/scripts/check_workflow_yaml.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python3
+# Copyright 2024-2026 The Spice.ai OSS Authors
+#
+# GitHub Actions definition guard.
+#
+# Verifies that every workflow under `.github/workflows/` and every composite
+# action under `.github/actions/` parses as YAML and carries the keys GitHub
+# needs to schedule it.
+#
+# A definition GitHub cannot parse fails *open*: the run reports no jobs, the
+# declared `on:` triggers and `paths:` filters cannot be honoured, and the run is
+# not a required check, so the workflow is silently disabled while every push
+# raises a failed run nobody is obliged to look at. Regression guard for #12181,
+# where a mis-indented block scalar left the daily Rust advisory scan unable to
+# run at all.
+"""Validate the repository's GitHub Actions workflow and composite action YAML."""
+
+from __future__ import annotations
+
+import argparse
+import sys
+from pathlib import Path
+
+import yaml
+
+# PyYAML resolves an unquoted `on:` key to the boolean `True` (YAML 1.1), which
+# is also how it reaches GitHub's parser, so a trigger block is present under
+# either spelling.
+TRIGGER_KEYS = ("on", True)
+
+
+def workflow_files(github_dir: Path) -> list[Path]:
+    """Return the workflow definitions under `<github_dir>/workflows`, sorted."""
+    workflows = github_dir / "workflows"
+    return sorted(p for p in workflows.glob("*.y*ml") if p.is_file())
+
+
+def action_files(github_dir: Path) -> list[Path]:
+    """Return the composite action definitions under `<github_dir>/actions`, sorted."""
+    actions = github_dir / "actions"
+    return sorted(p for p in actions.glob("*/action.y*ml") if p.is_file())
+
+
+def _parse_mapping(text: str) -> tuple[dict | None, list[str]]:
+    """Parse a definition into its top-level mapping, or report why it has none."""
+    try:
+        document = yaml.safe_load(text)
+    except yaml.YAMLError as error:
+        return None, [f"is not valid YAML: {_format_yaml_error(error)}"]
+
+    if not isinstance(document, dict):
+        return None, ["does not parse to a mapping of top-level keys"]
+
+    return document, []
+
+
+def check_workflow(text: str) -> list[str]:
+    """Return the problems with one workflow definition's contents.
+
+    An empty list means the definition is well-formed. Kept free of filesystem
+    access so the failure modes can be unit-tested from strings.
+    """
+    document, problems = _parse_mapping(text)
+    if document is None:
+        return problems
+
+    if not any(key in document for key in TRIGGER_KEYS):
+        problems.append("has no `on:` trigger block")
+
+    jobs = document.get("jobs")
+    if jobs is None:
+        problems.append("has no `jobs:` block")
+    elif not isinstance(jobs, dict) or not jobs:
+        problems.append("has a `jobs:` block that is not a non-empty mapping")
+
+    return problems
+
+
+def check_action(text: str) -> list[str]:
+    """Return the problems with one composite action definition's contents."""
+    document, problems = _parse_mapping(text)
+    if document is None:
+        return problems
+
+    return [] if "runs" in document else ["has no `runs:` block"]
+
+
+def _format_yaml_error(error: yaml.YAMLError) -> str:
+    """Render a PyYAML error as a single line, with the mark when it carries one."""
+    problem = getattr(error, "problem", None) or str(error)
+    mark = getattr(error, "problem_mark", None)
+    if mark is None:
+        return problem.replace("\n", " ")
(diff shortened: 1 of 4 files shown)


claudespice:
@copilot review

lukekim (approved):

Issue #12064, opened 2026-07-26
OTLP ingest: a Gauge/Sum dimension named like a histogram value column breaks the metric's schema, rejecting every export after the first


claudespice:
## Problem

OTLP metric ingest treats six histogram value-column names as globally reserved, so a **Gauge or Sum** metric that carries a data-point attribute named `count`, `sum`, `min`, `max`, `bucket_counts`, or `explicit_bounds` no longer round-trips: its stored table schema and the batch built for the next export disagree, and every subsequent export for that metric is rejected.

`RESERVED_COLUMN_NAMES` (`crates/runtime/src/opentelemetry.rs`) lists the number-metric value columns (`value`, `time_unix_nano`, `start_time_unix_nano`) *and* the histogram value columns (`count`, `sum`, `min`, `max`, `bucket_counts`, `explicit_bounds`). `initialize_attribute_schema` consults it through `is_reserved_column` for **both** the number (Gauge/Sum) and histogram paths — but for a number metric those six names are ordinary dimensions, not value columns.

## Effect

`initialize_attribute_schema` exists so that a column present in the stored schema is pre-seeded and gets a null backfill on any later data point that omits it, and so the attribute column order stays identical to the stored schema. Skipping a legitimate dimension defeats both:

1. **The dimension disappears when a data point omits it.** No builder is seeded from the stored schema, so no null is backfilled and the column is absent from the batch entirely.
2. **The dimension moves to the end of the column order** when a data point does carry it, because it is re-created by `append_attribute!` after the seeded attributes instead of in its stored position.

Either way the batch no longer matches the stored table positionally. `verify_schema` on the write path is exact-positional, so `write_data` fails, the export path counts the points as rejected, and the only signal is a `warn!`. The metric stops ingesting from the second export onward.

A first export (`existing_schema == None`) succeeds, so the table is created and looks healthy — the failure starts on the next export.

## Related: an attribute that really does collide

Independently, an attribute key that genuinely matches one of the metric's own value columns (`value` on a Gauge, `count` on a Histogram) is skipped by `initialize_attribute_schema` and then re-added by `append_attribute!` as a *new* field, so the batch carries **two** columns with that name and two different types. Arrow permits duplicate field names, so this builds and is written; the resulting table has an ambiguous column. This predates histogram support for `value` and arrived with it for the histogram columns.

## Reproduce

`crates/runtime/src/opentelemetry.rs`, against `trunk`:

```rust
[...]

PR #12187, opened 2026-07-30
fix(otel): reject a metric whose dataset has duplicate columns with an actionable error (fixes #12095)

### LINKED ISSUE #12095: OTLP ingest: a metric table with pre-existing duplicate columns rejects every export with only a warning
## Describe the bug

A metric table created by the pre-#12065 OTLP ingest path can hold **duplicate field names**, and once #12065 lands such a table can never accept another export — it is rejected with only a `warn!`, and nothing tells the operator what to do.

How a table ends up with duplicate columns: before #12065, an attribute whose key matched a reserved value-column name was skipped when the attribute schema was seeded, and then re-added by `append_attribute!`. Arrow permits duplicate field names, so the batch carried two same-named columns of different types and was written. A Histogram with an attribute named `count` (or a Gauge with an attribute named `value`) therefore produced a stored schema with two `count` columns.

After #12065:

- `initialize_attribute_schema` skips this metric's own value columns when seeding from `existing_schema`, so both stored `count` fields are skipped for a Histogram.
- The genuinely-colliding incoming attribute is dropped (one `warn!` per distinct key per batch).

The rebuilt batch therefore carries a single `count`, the stored schema carries two, and `write_data` → `verify_schema` (`crates/runtime/src/datafusion/mod.rs`) is exact-positional, so **every subsequent export is rejected**.

## Expected behavior

A stored metric schema that already contains duplicate column names is detected and reported with an actionable error naming the metric and telling the operator to drop and recreate the dataset — rather than every export being silently dropped with a `warn!`.

## Additional context

Not a regression introduced by #12065: such a dataset was already broken (that is #12064 — every export after the first was rejected). #12065 fixes schema generation so newly-created tables are correct; migrating an already-corrupt table is the separate piece of work tracked here.

Raised by Copilot in review on https://github.com/spiceai/spiceai/pull/12065.

claudespice:
## Summary

A metric table can hold **duplicate column names**: Arrow permits them, and an older OTLP ingest
path emitted an attribute alongside the same-named value column, so a Histogram with a `count`
attribute (or a Gauge with a `value` attribute) was stored with two `count`/`value` columns.

Such a table can never accept another export. When the batch is rebuilt against the stored schema,
both duplicates are either skipped as value columns (`initialize_attribute_schema`) or collapsed to
a single entry by the attribute `IndexMap`, so the batch always carries one column per name while the
table carries two. `write_data`'s `verify_schema` is exact-positional, so every export fails on a
field-count mismatch — reported only as a `warn!` naming neither the cause nor a remedy, and no
export can ever repair it.

Detect the condition up front instead, and report it with an error that names the metric, the
duplicated columns, and the fix (drop and recreate the dataset).

This is not a regression from #12065 — such a dataset was already broken (#12064). #12065 fixed
schema *generation* so newly created tables are correct; this covers the already-corrupt table.

Fixes #12095

## Changes

- `crates/runtime/src/opentelemetry.rs`
  - New `Error::ExistingMetricSchemaHasDuplicateColumns`, naming the metric, the duplicated column
    names, and the remedy, with a docs link.
  - New `duplicate_column_names(&Schema)` helper: names appearing more than once, each reported
    once, in first-seen order.
  - `metric_data_to_record_batch` checks the stored schema before dispatching to a builder, so the
    guard covers **every** data-point shape (Gauge, Sum, Histogram) and every duplicated column —
    value columns and ordinary dimensions alike — rather than only the reported Histogram/`count`
    case. Rejecting here also skips the pointless schema-evolution pre-flight and write attempt.
  - Data-point counting moved to a `data_point_count` helper so the early return reports the real
    number of rejected points; per-shape counts are unchanged (unsupported types still report 0).

## Test plan

- `make lint`
- `make test`

New unit tests in `crates/runtime/src/opentelemetry.rs`:

- `histogram_with_duplicate_stored_value_column_is_rejected_with_actionable_error` — the reported
  case (two `count` columns): asserts the typed error, that the message names the metric and the
  remedy, and that all data points in the export are counted as rejected. Returns `Ok` with a
  narrower schema before this change.
- `gauge_with_duplicate_stored_dimension_is_rejected` — the same corruption on an ordinary
  dimension (two `host` columns), proving detection is not limited to value columns.
- `duplicate_column_names_reports_each_name_once_in_first_seen_order` — helper unit test: a clean
  schema yields nothing; a name repeated three times is reported once, ordered by first occurrence.

Existing tests cover the no-false-positive side: a clean stored schema still round-trips
(`histogram_reuses_existing_attribute_schema`, `gauge_dimension_named_like_a_histogram_column_round_trips`,
`rebuild_against_evolved_schema_matches_exact_field_order`).

No behavior change for a table without duplicate columns: the guard is a name scan over the stored
schema per metric per export, and every existing path is reached unchanged when it finds nothing.

## Checklist

- [x] Regression test that fails before the change and passes after
- [x] Error message names the resource and gives an actionable fix with a docs link
- [x] Covers all data-point shapes, not just the reported Histogram case
- [x] No `.unwrap()` / `.expect()` in non-test code

### DIFF
--- a/crates/runtime/src/opentelemetry.rs
+++ b/crates/runtime/src/opentelemetry.rs
@@ -84,6 +84,13 @@ pub enum Error {
         "First data point for metric {metric} has no value and therefore is not valid for establishing schema"
     ))]
     FirstMetricDataPointHasNoValue { metric: String },
+
+    #[snafu(display(
+        "Failed to ingest OpenTelemetry metric {metric}: its dataset has duplicate columns ({columns}), \
+        so no export for this metric can be written. Drop and recreate the dataset for {metric} to resume ingest. \
+        See: https://spiceai.org/docs/features/observability"
+    ))]
+    ExistingMetricSchemaHasDuplicateColumns { metric: String, columns: String },
 }
 
 const VALUE_COLUMN_NAME: &str = "value";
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

peasee:
rollup into https://github.com/spiceai/spiceai/pull/12087

Issue #11887, opened 2026-07-15
Support write through and fallback indexes in vector stores. 


Jeadie:
Based on fallback indexes defined in #11884, wire this up for datasets/views with `.vectors` VectorStores.

PR #12189, opened 2026-07-30
feat(views): support `.time_column`/`.time_format` and `refresh_mode: append` for accelerated views (fixes #11959)

### LINKED ISSUE #11959: Add `.time_column` and `.time_format` to views for warm-tier search support
Sub-issue of #11828.

## Problem

The warm search-tier spec relies on time-based properties to drive its data-window mechanisms (retention, recency). Datasets already expose `.time_column` and `.time_format`, which let the acceleration/warm-tier write path determine time-based properties such as retention policies.

Views have no equivalent fields today, so warm-tier search cannot be supported for views: there is no way to derive the time window (creation-based for append, last-updated for mutating/change data) that the warm index needs.

From the enhancement spec:

> Datasets already have `.time_column` and `.time_format` to help determine time-based properties (e.g. retention policies). Views should add the same fields when warm-tier search for views is supported. For append, this needs to be at least creation based; for mutating/change datasets, this is at least last updated.

## Scope

- Add `.time_column` and `.time_format` fields to the view component (spicepod schema + parsing), mirroring the dataset fields.
- Wire these through so the warm-tier / acceleration data-window logic can consume them for views the same way it does for datasets. 
- Semantics: append data → at least creation-based; mutating/change data → at least last-updated.

## Out of scope

- Full warm-tier search support for views beyond exposing the time fields (tracked by the broader vector-store wiring in #11887).

Jeadie:
Resolves #11959.

See the issue thread for the design proposal posted by the investigation step.

## Update

This PR now also supports `refresh_mode: append` for accelerated views, not just the `.time_column`/`.time_format` fields:

- `refresh_mode: append` is allowed for views (previously only `full` was accepted), and requires `time_column` to be set — enforced at spicepod-parse time.
- Time-column-watermark append is only correct for row-preserving view queries (projections/filters/unions over an append-only base) — it silently produces wrong results for views with aggregates, window functions, `DISTINCT`/set-ops, `LIMIT`, or joins, since a later-arriving row can change an already-appended output row and append mode has no way to revise it. A new default-deny allow-list validator (`validate_view_append_refresh_plan`) checks the view's resolved `LogicalPlan` and rejects anything outside the known-safe shapes with a specific error, before the view is accelerated.
- `refresh_mode`/`time_column`/`time_format` are now threaded through `create_accelerated_view` (previously hardcoded to `full`).
- `refresh_mode: changes` (CDC) remains unsupported for views — that's a substantially larger feature (incremental view maintenance) and out of scope here.

See the PR discussion below for the full breakdown of what was/wasn't already wired up, and the append-safe SQL shapes.


Jeadie:
Confirmed this PR is correctly scoped to #11959: it only exposes `time_column`/`time_format` as metadata on views (schema, `ViewSpec`/`View`, `AccelerationSource::time_column()`). It does **not** unlock `refresh_mode: append` (or `changes`) on an accelerated view — that's still fully blocked, and there's more to it than just lifting the current validation. Leaving this here so the remaining scope is tracked (likely under #11887) and nobody assumes `time_column` alone is enough:

**Still blocked today, unmodified by this PR:**
- `crates/runtime/src/component/view.rs:148-156` rejects any `acceleration.refresh_mode` other than `Full` for views (`AcceleratedViewInvalidConfiguration: "Only 'refresh_mode: full' is supported"`).
- `create_accelerated_view` (`crates/runtime/src/datafusion/mod.rs:4312`) hardcodes `Refresh::new(RefreshMode::Full)` and never reads `acceleration.refresh_mode` at all — so even relaxing the check above wouldn't change behavior without this being wired through too.

**Structural gap — no connector to hang append/changes on:**
- `append_stream`/`changes_stream` are defined on `DataConnector` (`crates/runtime/src/dataconnector/mod.rs:633-647`, `changes_stream` is literally typed `&Dataset`), not on `AccelerationSource` (the trait `View` implements). `create_accelerated_view` only has `view_table: Arc<dyn TableProvider>` — no `Arc<dyn DataConnector>` — so there's nothing to call these on for a view. Views are a derived SQL query over already-registered dependent tables, not a single upstream source, so this needs its own design rather than reusing the dataset path directly.
- The good news: `get_incremental_append_update`/`get_data_update`/watermark computation in `refresh_task.rs` (~1095-2162) already operate generically over `FederatedTable`/`Arc<dyn TableProvider>` and the generic `Refresh.time_column`, not a concrete `Dataset` — and `DatasetCheckpoint` (`dataset_checkpoint/mod.rs:124`) already takes `&dyn AccelerationSource` and is already called generically for views. So the incremental-refresh and checkpoint machinery itself isn't the blocker — it's purely the wiring above.

**Semantic gap — this is the part that needs a real design, not just wiring:**
- Time-column-based append (`MAX(time_column) > watermark` pushed down as a scan filter) is only sound for row-preserving pass-through views (filter/projection over an append-only base). It silently breaks for views with aggregations (a new row can retroactively change an already-materialized group's aggregate, and append mode has no retraction), joins (a late match on the *other* input for an already-appended row is never reproduced), or window functions (`ROW_NUMBER`/`LAG`/`LEAD`/running totals over already-appended rows can change as new rows land in the same partition).
- Nothing in the runtime currently detects or restricts this at config time — so before wiring append/changes through for views, we need either (a) validation that only permits it for provably pass-through view shapes, or (b) an explicit, documented opt-in with correctness caveats.
- `refresh_mode: changes` (CDC) is a bigger step further: it's fundamentally inapplicable to an arbitrary view SQL without incremental view maintenance (re-diffing joins/aggregations as base-table CDC events arrive), which is a substantially larger feature than the append case.

Happy to open a follow-up issue for this breakdown if it's useful for scoping #11887.

Jeadie:
Pushed a commit that closes the gap described above: `refresh_mode: append` is now actually usable for views, not just the `time_column`/`time_format` metadata.

**What changed:**
- `crates/runtime/src/component/view.rs`: the config-time validation now accepts `refresh_mode: append` in addition to `full`, and requires `time_column` to be set when `append` is chosen (still rejects `changes`/`caching`/`snapshot`).
- `crates/runtime/src/datafusion/sql_validator.rs`: new `validate_view_append_refresh_plan` — a default-deny allow-list over the view's resolved `LogicalPlan`. Only row-preserving shapes (`Projection`, `Filter`, `TableScan`, `SubqueryAlias`, `Union` i.e. `UNION ALL`, `Sort` without `Limit`, `EmptyRelation`) are permitted; `Aggregate`, `Window`, `Distinct` (also catches plain `UNION`/`INTERSECT`/`EXCEPT`), `Limit`, and `Join` are explicitly rejected with a reason, and anything else falls through a generic deny — so an unrecognized future `LogicalPlan` variant fails closed instead of silently being treated as safe.
- `crates/runtime/src/view.rs`: `prepare_view` now runs that validator against the view's resolved plan whenever `refresh_mode: append` is configured, before the view is accelerated.
- `crates/runtime/src/datafusion/mod.rs`: `create_accelerated_view` no longer hardcodes `RefreshMode::Full` — it reads `acceleration.refresh_mode` and threads `time_column`/`time_format` into the `Refresh` builder the same way the dataset path does, plus a defensive check that append mode without `time_column` fails clearly.
- Added unit tests for the validator (one per allowed/denied shape) and for the config-time gate, plus two integration tests in `crates/runtime/tests/view/mod.rs`: a pass-through view that picks up a newly-appended source row after a refresh, and an aggregating view confirming `refresh_mode: append` is rejected.

`refresh_mode: changes` (CDC) is still out of scope — that needs real incremental view maintenance to handle joins/aggregations correctly, which is a much larger feature.

I ran `cargo check`/`cargo fmt` locally against this; leaving full test execution to CI/signoff rather than re-running the whole suite locally.

phillipleblanc (dismissed):

claudespice:
### Merged `trunk` in and resolved the conflict — this PR was `CONFLICTING/DIRTY` and four days from its second auto-close

This PR has been unmergeable since the last base merge on 2026-08-26. It was already closed
once as stale on 2026-08-22 and reopened, and the stale bot marked it again on 2026-09-04, so
the seven-day clock was due to expire around 2026-09-11. The conflict is what nobody could
review past, so I merged `trunk` in (`427e418e`) — a plain merge, no rebase and no force-push.

**The conflict was one file, two hunks, both inside `#[cfg(test)] mod tests`.** Trunk and this
branch each appended tests to the same block, and git split the union on a `);` / `}` /
`#[test]` line sequence that occurs on both sides — so the hunk boundaries fell *mid-test* on
each side rather than between tests. Rather than edit around the markers I rebuilt the whole
`mod tests` block from both sides. The two test sets are disjoint, with no overlapping name:

| | tests | subject |
|---|---|---|
| `trunk` | 6 | `acceleration.ready_state` resolution, log-line escaping |
| this branch | 4 | `refresh_mode` validation |

The import block is the union of what each side needs — trunk's explicit `use super::{…}` plus
this branch's `use spicepod::acceleration::{…}`. I kept trunk's explicit form rather than
collapsing to `use super::*`.

**The production half of the file merged automatically and keeps both sides**: trunk's
`acceleration.ready_state` resolution block, and this branch's `time_column`/`time_format`
fields, `refresh_mode: append` validation, and the real `time_column()` accessor.

#### Evidence

The merged tree compiles and both sides' tests pass together:

```
$ cargo test -p runtime --lib --features <40 workspace-resolved> component::view::tests
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 1407 filtered out
```

**10 selected is itself the check that neither side was dropped** — 6 + 4. Had the resolution
lost a side, the filter would have selected 6 or 4. All six trunk test names and all four of
this branch's are present in the merged file, and the only deletions in
`git diff origin/trunk -- crates/runtime/src/component/view.rs` are this branch's own intended
production edits (the extended import, the `refresh_mode` match replacing the `full`-only
check, and `time_column()` no longer returning `None`) — no test was removed.

Post-merge tree validation, with a positive control so the check isn't vacuous:

```
$ cargo metadata --locked --format-version 1   # merged tree
rc=0
$ # same command with one [[package]] block deliberately duplicated in Cargo.lock
rc=101   error: failed to parse lock file
```

Before/after on the blocker itself: `CONFLICTING/DIRTY` before the push, `MERGEABLE/BLOCKED`
after it.

#### What this does *not* address

- **The four unresolved review threads are untouched and still yours.** They raise append-refresh
  correctness questions — the single watermark under `Union`, non-deterministic expressions
  passing a variant-only plan walk, and `refresh_append_overlap` being parsed but never threaded
  into `Refresh`. Each is a decision about the feature's semantics, not something a conflict
  resolution should answer, so I left them alone.
- `Attestation` is red until the sign-off lands. A hand-resolved merge doesn't inherit the
  previous one, so I'm running `make signoff` against `427e418e` now; the status will appear on
  this commit when it finishes.

Adversarial review was **skipped, declared**: the Codex workspace is out of credits
(`Your workspace is out of credits`) and the grok plugin isn't installed on this host. The
security pass found nothing — the new surface is two typed Spicepod config fields and a
default-deny plan allow-list, neither of which builds SQL or crosses a privilege boundary.

claudespice:
Sign-off landed and `Attestation` is green on `06525154b8`.

The merge needed one follow-up commit before it would build, and it is the second half of the
same conflict rather than a separate problem — worth writing down because nothing on the PR
would have shown it. Trunk's move to the level-triggered refresh-completion signal
(#13505 → #13709 → #13735) replaced `RefreshCompletionWaiter::notified()` with `wait()`, and
this branch's `append_view` test still called `notified()`. Neither side is wrong alone; only
the merged tree fails, and only in the `integration` test target:

```
error[E0599]: no method named `notified` found for struct `runtime_table::RefreshCompletionWaiter`
   --> crates/runtime/tests/view/mod.rs:946:22
error: could not compile `runtime` (test "integration") due to 1 previous error
```

Changed to `notifier.wait().await`, which is verbatim how `tests/acceleration/refresh/common.rs`
and `tests/acceleration/retention_arrow.rs` already await a refresh. Both discard the
`RefreshCompletionOutcome`, as the old call effectively did.

Same command, before and after the one-line change: `Targeted pre-lint` failed with that E0599,
then `Targeted pre-lint passed`. Full gate on the fixed HEAD:

```
Summary [ 777.666s] 12824 tests run: 12824 passed (2 slow, 1 flaky), 27 skipped
✓ Signed off on 06525154b8ef — lint+test passed in 10906s
```

(The one flaky was `scheduler channel::cron::tests::test_cron_resets_to_next`, which passed on
retry and is unrelated to this branch.)

I also removed the `stale` label, since that clock was the reason this had a deadline at all.

**Still yours, unchanged:** the four unresolved review threads on append-refresh semantics. Those
are the only things between this and a review now — `mergeStateStatus` is `MERGEABLE/BLOCKED`
with `Attestation` green.

PR #12200, opened 2026-07-30
refactor(runtime): describe every table-provider wrapper layer once in a layer table

phillipleblanc:
## What

A dataset's `TableProvider` stack is assembled from wrapper layers (indexes, embeddings, vector scan, metadata enrichment, federation, acceleration), and many subsystems walk that stack: search discovery, CDC stream detection, source peeling for change streams, the CDC write path, index collection for write-lifecycle hooks, and metadata-enrichment pushdown. Each walk enumerated the wrapper types by hand, so a new or intervening layer had to be added to every enumeration — and a missed one silently disabled whatever behavior that walk feeds. #12086, #12063 and #12094 were all this failure mode.

Each wrapper is now described exactly once, in `runtime::table_layers`, as a `TableProviderLayer`: per-walk transparency (read discovery / CDC detection / source peel / write pass-through) plus an optional rebuild hook for pushing transforms below the layer. The generic walks live in `runtime-datafusion-index` (`find_concrete_table_provider_in`, `peel_to_innermost`, `visit_provider_chain`, `rebuild_innermost_table_provider`) and consume the table. The per-call-site accessor sets (`DEFAULT_INNER_FNS`, `TRANSPARENT_CDC_WRAPPERS`, the CDC source `PEEL` set, the write-transparent set) and the hand-rolled downcast arms in `collect_indexes_from_provider` and the two metadata-pushdown functions are gone, replaced by walks over the table. `UpsertDedupTableProvider` is registered as explicitly opaque, so "never route writes past dedup" is recorded as data rather than a comment on one call site.

## Bugs fixed along the way

Each with a regression test where practical:

- `collect_indexes_from_provider` stopped at `MetadataEnrichedTableProvider` and `VectorScanTableProvider` layers, silently dropping write-lifecycle hooks for indexes below them.
- Read-path discovery (`find_concrete_table_provider`) could not see through `VectorScanTableProvider`, so lookups such as the health monitor's Iceberg check and the CDC-ingest table lookup missed providers under a vector-enabled dataset.
- `ElasticsearchFullTextConnector` did not forward `initialization_for_dataset`, losing a per-dataset deferral decision made by the inner connector. It and `DeferredConnector` now carry `#[deny(clippy::missing_trait_methods)]` like the other `DataConnector` wrappers, so the next trait method cannot silently take the default in a wrapper.
- The `elasticsearch` feature did not compile without `models`: the Elasticsearch index wrapper referenced `EMBED_UDF_NAME` through the `models`-gated `embed` module. The constant now lives at the crate root of `runtime-datafusion-udfs`, so builds without `models` can still look the UDF up by name and fail with a structured error when it is not registered.

## Deliberately not converted

Walks with genuinely different stepping semantics are left as-is; each is a candidate for a follow-up walk kind in the layer table:

- the accelerator-side cayenne walks (`partition_expr_from_table_provider`, `is_cayenne_table_provider`, `duckdb_writer_context`, `extract_cayenne_write_target`), which descend different children (Poly writer side, the accelerator side of `AcceleratedTable`);
- the arrow delete/maintenance routing in `refresh_task/changes.rs`, which treats `EmbeddingTable` as transparent for deletes but must not for inserts;
- the two catalog-entry peels (`planner::peel_table_provider_wrappers` and the accelerated-table lookup in `datafusion/mod.rs`), which differ in empty-adaptor error semantics.

The two metadata-pushdown call sites still pass restricted layer subsets on purpose (the refresh-path and registration-time stacks contain different layers); converging them on the full table needs verification of what each stack can contain and is left as a follow-up.

## Testing

- New regression tests: `collect_indexes_finds_indexes_under_metadata_enrichment`, `collect_indexes_finds_indexes_under_vector_scan` (both fail against the previous walk shape), `test_find_concrete_table_provider_peels_vector_scan_wrapper`, plus layer-table invariant tests (`layer_names_are_unique`, `layer_walk_transparency_narrows_monotonically`, `layer_walks_match_declared_transparency`).
- Existing wrapper/pushdown coverage passes: 163 unit tests across `table_layers`, `search::util`, `refresh_task`, `datafusion::tests`, `embeddings`, `changes` and `full_text`, including the metadata-pushdown discoverability tests that guard the rebuild conversions.
- `make lint-rust PACKAGES="runtime runtime-datafusion-index runtime-datafusion-udfs" FEATURES="duckdb,elasticsearch"` passes (fmt, crate-layering check, both pedantic clippy passes over lib and tests).


crates/runtime/src/embeddings/index/table.rs
```diff
@@ -20,8 +20,8 @@ use crate::model::EmbeddingModelStore;
 use crate::secrets::Secrets;
 use datafusion::datasource::TableProvider;
 use datafusion::{prelude::SessionContext, sql::TableReference};
-#[cfg(feature = "models")]
-use runtime_datafusion_udfs::embed::EMBED_UDF_NAME;
+#[cfg(feature = "elasticsearch")]
+use runtime_datafusion_udfs::EMBED_UDF_NAME;
```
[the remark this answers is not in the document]
phillipleblanc: Fixed in 31868e96: the import is now gated on `any(feature = "s3_vectors", feature = "elasticsearch")`, matching both functions that reference it. Verified with `cargo check -p runtime --features duckdb,elasticsearch,s3_vectors`.

crates/runtime-datafusion-index/src/util.rs
```diff
@@ -33,16 +33,118 @@ pub const INDEXED_INNER: InnerProviderFn = |tbl| {
...
+    &Arc<dyn TableProvider>,
+    &dyn Fn(Arc<dyn TableProvider>) -> Arc<dyn TableProvider>,
+) -> Option<Arc<dyn TableProvider>>;
+
+/// The walk being performed over a stack of [`TableProvider`] wrappers. Each
+/// [`TableProviderLayer`] declares, per context, whether the walk may step
+/// through it — so the decision lives once, at the layer's descriptor, instead
+/// of at every call site.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LayerWalk {
```
bjchambers, line 54: This seems like a lot of specialized logic for the specific layering. I'm wondering if we'd be better off with something more like how web frameworks handle filters. Something like:

```
enum TableProviderHookResult<T> {
  /// Passthrough to the underlying table provider result.
  Passthrough,
  /// Return this value.
  Return(T),
}

trait TableProvider {
  fn foo(&self) -> TableProviderHookResult<i64> {
    TableProviderHookResult::Passthrough
  }

  ...
}

struct TableProviderStack(Vec<dyn TableProvider>);

impl TableProvider for TableProviderStack {
  fn foo(&self) -> TableProviderHookResult<i64> {
    for provider in self.0 {
      match provider.foo() {
        TableProviderHookResult::Return(value) => return value,
        TableProviderHookResult::Passthrough => continue,
    }
    error!("No table provider in the stack handled foo");
  }
}
```

bjchambers (dismissed):

crates/runtime-datafusion-index/src/util.rs
bjchambers: Or:

```
trait TableProvider {
  fn foo(&self, ..., chain: Option<&TableProvider>) -> i64 {
    match chain {
      Some(next) => next.foo(...),
      // Or, if there is a sensible default, use that.
      None => panic!("No implementation handles `foo`");
    }
}
```

Benefit, this allows the *default* behavior for the methods be a proper chaining.

It seems like we've been bitten by this kind of thing in traits before, so I'm wondering if we should discuss a different pattern like one of the above to ensure trait methods are properly chained by wrappers.

lukekim (dismissed):

crates/runtime-datafusion-index/src/util.rs
phillipleblanc: I think the two options are: continue to use the DataFusion `TableProvider` trait exactly in the spice codebase - and then either do the approach in this PR or standardize on something like `TableProviderStack` like you mentioned. I think the main issue with that is its still fairly easy to end up with a new codepath that ends up double wrapping (i.e. I attempted something similar with the `Index` TableProvider trait, and it didn't really end up working as I imagined).

The other option is to have a concrete struct `SpiceTable` that we use natively everywhere, which is aware of all the layers, integrates natively with accelerations, etc. This is the direction I've been thinking we should go, but it would be a relatively large refactor that I haven't wanted to attempt yet. But it would probably be a good agent goal to implement when we're ready.

phillipleblanc: For now, I will merge this PR as is.

ewgenius (dismissed):

crates/runtime/src/embeddings/connector.rs
```diff
@@ -467,9 +471,10 @@ impl DataConnector for EmbeddingConnector {
...
         let embedding_table = Arc::new(
-            find_concrete_table_provider_with::<EmbeddingTable>(
+            find_concrete_table_provider_in::<EmbeddingTable>(
                 &table_provider,
-                TRANSPARENT_CDC_WRAPPERS,
+                TABLE_PROVIDER_LAYERS,
+                LayerWalk::CdcDetection,
             )?
             .clone(),
         );
```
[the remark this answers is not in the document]
phillipleblanc: Good catch — confirmed and fixed in `1d6cd16c`. This arm now resolves the scan with `find_concrete_table_provider_in` under `LayerWalk::CdcDetection`, matching the three sibling arms in the same method (the vector-scan layer is deliberately opaque to that walk, which is what makes the dedicated arm necessary).

I also swept the whole workspace for both symbols: this was the only remaining call site for `find_concrete_table_provider_with`, and the only other `TRANSPARENT_CDC_WRAPPERS` reference was a stale doc comment on the regression test for #12313, updated in the same commit.

peasee (approved):

PR #12226, opened 2026-07-31
fix(search): Improvements for search harness

Jeadie:
## Summary

The search benchmark harness (`testoperator run search`) needed some love.

### NDCG ideal DCG now comes from the relevance judgments
The old code computed the ideal DCG from the retrieved documents only. A query with two relevant documents scored a perfect 1.0 when the search returned only one of them. The ideal DCG now ranks all judged documents from the qrels, so missed relevant documents lower the score. This matches the referenced MTEB `RetrievalEvaluator` methodology. For the MTEB QuoraRetrieval qrels, only 23.3% of queries had 2 relevant document (none had >2).

### Queries without results now score 0
The old code skipped judged queries that had no search results. This inflated the average. These queries now score 0 and stay in the average. An empty qrels set now returns an error instead of NaN.

### Search worker checks HTTP status
The search worker parsed every response body as a `SearchResponse`. An HTTP error surfaced as a confusing deserialize error. The worker now fails with the status code and response body.

### RPS metric uses fractional seconds
The old code returned the raw request count for runs under one second. For example, 100 requests in 0.5s reported 100 RPS instead of 200.

### Dataset copies use tokio::fs
`prepare_dataset` copied parquet files with blocking `std::fs::copy` inside an async function. It now uses `tokio::fs::copy`. This adds the `fs` feature to the testoperator tokio dependency.

## Testing
Not yet


claudespice:
Heads-up on the base, since this one is easy to miss on a rebase: this branch does not contain #12193 (`5b1405be4`), which landed on trunk and **replaced** both functions this PR edits — `calculate_ndcg` became `calculate_retrieval_metrics`, and `idcg_at_k`/`ndcg_at_k` were restructured alongside the new `Recall@k`/`MRR@k`/`Precision@k`. The PR currently reads `CONFLICTING`, so the diff can't apply as written.

Two things worth carrying over when you rebase:

1. Both defects this PR fixes are still live on trunk in the new shape — `idcg_at_k` still takes its ideal from the retrieved documents, and the unanswered-query `continue` is still there.
2. The `continue` now affects **all four** metrics, not just NDCG, so the averaging fix needs to cover recall/MRR/precision too. `average()` also still evaluates `0.0 / 0.0` → `NaN` when no judged query is answered.

I filed #12277 for the trunk state and opened #12278, which applies these same fixes to trunk's current structure and keeps your API shape (`anyhow::Result` + `ensure!` on empty `qrels`) so the two don't diverge. Happy to close #12278 if you'd rather carry it here — I mainly wanted the diagnosis recorded rather than to land a parallel change in your area.

phillipleblanc (approved):

PR #12227, opened 2026-07-31
fix: Report query metrics when task history is disabled

sgrebnov:
## Summary

Query metrics were only reported when `runtime.task_history.enabled` was true. The query tracker is the sole emitter of them, and it was skipped entirely when task history was disabled (#5852), so these all disappeared:

- `query_duration_ms`
- `query_execution_duration_ms`
- `query_executions`
- `query_failures`
- `query_returned_rows`, `query_returned_bytes`

Deployments that disable task history therefore lost every query latency, throughput and error signal, while other metrics (`flight_request_duration_ms`, HTTP) kept working — breaking dashboards built on the documented query metrics. The Cayenne adaptive tuner also lost its query-latency input.

The tracker is now always built. `runtime.task_history.enabled` controls only what it reports: the `task_history` events and the captured-output preview, which stays skipped so the per-query cost of the original change is largely retained.

## Testing

- Unit tests: the tracker is built with the setting on and off.
- `crates/runtime/tests/query_metrics.rs`: runs a query against a runtime with `task_history.enabled: false` and asserts the metrics are reported to a Prometheus-backed meter provider.

The end-to-end test needs its own test binary, because a metric handle binds to whichever meter provider is installed when it is first used. `make nextest` runs it explicitly, since the workspace gate is `--lib` only.

It adds ~6s to the gate — 4.7s to compile and link, 0.25s to run — measured with the `runtime` lib already built and this target's artifacts removed.


Makefile
```diff
@@ -102,10 +102,13 @@ endif
 # of the gate agreeing on which crates need system libraries, so a sign-off on
 # such a host fails only for reasons in the branch under test. The crate has no
 # unit tests of its own.
+# The `query_metrics` test binary is listed explicitly because `--lib` skips it:
+# it needs its own process to control the OTel meter-provider install order.
 .PHONY: nextest
 nextest:
 	@cargo nextest run --all --exclude libnfs --lib $(NEXTEST_CARGO_PROFILE) $(NEXTEST_FLAG)
 	@cargo nextest run -p cayenne --tests $(NEXTEST_CARGO_PROFILE)
+	@cargo nextest run -p runtime --test query_metrics $(NEXTEST_CARGO_PROFILE)
```
sgrebnov, line 111: Measured that additional test does not add new compile time complexity:
>Local test showed that it adds ~6s to the gate — 4.7s to compile and link, 0.25s to run — measured with the runtime lib already built and this target's artifacts removed.

peasee (approved):

Issue #12225, opened 2026-07-31
Stabilize model E2E tests in the merge queue


bjchambers:
## Summary

The model jobs in `E2E Test CI` are frequently failing in the merge queue and then passing when the workflow is manually retried. This blocks the queue and hides the real failure rate when only the final workflow conclusion is considered.

The failures affect both the local Hugging Face model and the OpenAI-backed model jobs. They are concentrated in the Expect-driven REPL checks, with additional OpenAI rate-limit evidence in the runtime logs.

## Recent flake rate

Sample: the 50 most recent completed, non-cancelled `merge_group` runs from **2026-07-29 07:10 UTC through 2026-07-31 00:52 UTC**. Two runs skipped the model jobs, leaving 48 first executions. Nine workflows were manually rerun; every model job in those second attempts passed.

A failed attempt is counted even when a later manual retry passed.

| Model job | First executions failed | First-execution failure rate | Failures / all executed attempts |
|---|---:|---:|---:|
| Hugging Face / macOS aarch64 | 8 / 48 | 16.7% | 8 / 57 (14.0%) |
| OpenAI / macOS aarch64 | 3 / 48 | 6.3% | 3 / 57 (5.3%) |
| OpenAI / Linux x86_64 | 2 / 48 | 4.2% | 2 / 57 (3.5%) |
| **Overall** | **13 / 144** | **9.0%** | **13 / 171 (7.6%)** |

Twelve of the 48 runs with model jobs had at least one model-job failure: **25.0% of merge-queue E2E runs in the sample**.

## Failure breakdown

| Failed step | Hugging Face macOS | OpenAI macOS | OpenAI Linux | Total |
|---|---:|---:|---:|---:|
| Test vector search | 3 | 1 | 2 | 6 |
| Test chat | 3 | 2 | 0 | 5 |
| Test the Expect scripts driving the REPLs | 2 | 0 | 0 | 2 |

The two `expect_test.sh` failures are especially notable: this is a stand-in/unit-style test with no model or network dependency, but it currently runs while the 3B local model is loaded. It timed out waiting 30 seconds for the stand-in `spice chat` process to print its initial prompt.

The real REPL checks currently use a 5-second search timeout and a 30-second chat timeout. Observed signatures include:

- `Timeout waiting for initial chat prompt`
- `Timeout waiting for expected response`
- REPL child exits reported as `CHILDKILLED SIGTERM`
- `Invalid HTTP response: Failed to read stream: error decoding response body`
[...]

PR #12261, opened 2026-07-31
fix(e2e): stabilize model REPL tests

bjchambers:
## Summary

Stabilize the existing end-to-end model tests while preserving their full path through the real `spice` CLI and Spice runtime.

Recent merge-queue data showed 13 model-test failures across 144 first executions (9.0%), affecting 12 of 48 workflows (25%). Hugging Face on macOS was the most affected at 8/48 first attempts (16.7%).

This PR changes only tests and CI configuration. It depends on #12446, already on trunk — see below.

## Changes

- Run chat tests with `--temperature 0` to reduce output variance.
- Pin the OpenAI chat model to `gpt-4o-mini-2024-07-18`.
- Pin the Qwen chat model and BGE embedding model to immutable Hugging Face revisions.
- Make REPL response deadlines configurable with `SPICE_REPL_TIMEOUT_SECONDS`.
- Increase CI search deadlines to 30 seconds and chat/harness deadlines to 60 seconds.
- Increase the OpenAI model job deadline from 5 to 10 minutes.
- Improve timeout diagnostics so failures identify the operation and elapsed deadline.
- Add harness coverage for configured and invalid timeout values.

The tests still exercise startup, vector search, and chat through the real `spice` CLI. This does not add retries, bypass the CLI, or weaken existing assertions.

## Why this needs #12446

Pinning the BGE embedding revision could not work until #12446, which is why the first merge-queue run failed. `Embeddings::get_model_id` re-joins the optional revision from `from:` onto the repo id with a colon, so `huggingface:huggingface.co/BAAI/bge-base-en-v1.5:<sha>` yields `BAAI/bge-base-en-v1.5:<sha>`. That whole string was handed to `TeiEmbed::from_hf` as the repo id with `revision: None`, addressing a repo that does not exist — every artifact fetch resolved to `huggingface.co/BAAI/bge-base-en-v1.5:<sha>/resolve/main/...` and returned 401.

The embedding model then failed to load, which made both datasets that use it invalid, so the runtime never reported ready and the readiness wait timed out. The chat path was unaffected because it already splits the same suffix and passes the second half as the revision.

#12446 fixed that by recovering the two halves through a shared `split_hf_model_id` helper, and is already on trunk and merged into this branch. The embedding revision pin here therefore only works on top of it.

## Expected impact

The changes directly address the two harness prompt timeouts and likely address the two pure live-response deadline failures observed in the sample. Deterministic generation and immutable model revisions should also reduce the semantic mismatch and model-drift risk, while the longer deadlines may allow transient rate-limit delays to recover.

This is a partial mitigation for the observed failures. It does not mask or directly solve connection failures, invalid response streams, exhausted provider quota, or child-process SIGTERM failures. We should measure another 50 merge-queue runs after merge to quantify the resulting first-attempt flake rate and prioritize the remaining failure classes.

## Validation

- `./test/models/expect_test.sh`
- `bash -n test/models/expect_test.sh`
- Parsed the workflow and both modified Spicepods as YAML.
- `git diff --check`
- Confirmed both pinned revisions resolve on Hugging Face, and that the pinned BGE revision carries the artifacts the loader fetches (including the two that returned 401).

Addresses #12225.


sgrebnov (dismissed):

grokspice:
Heads-up on the `huggingface model (darwin-aarch64)` failure on this PR's merge-queue runs — it is not flakiness, and it is caused by the revision pins added here.

A revision-pinned HuggingFace **embedding** never loads: the revision gets joined back onto the repo id with a colon and the whole string is passed as the repo name, so the fetch asks for a repo that does not exist and HF answers 401.

```
WARN text_embeddings_core::download: Download failed: ... (401 Unauthorized) for url
  (https://huggingface.co/BAAI/bge-base-en-v1.5:a5beb1e3.../resolve/main/config.json)
```

Note the revision sits in the *repo* segment while the revision segment is `main`. `bge_base` then fails to load, the `issues` dataset fails config validation because it references it, the runtime never reports ready, and the 5-minute readiness wait times out — stranding the merge-queue entry.

The pinned model in the same file is fine, because the chat path splits the `:revision` back off (`crates/llms/src/chat/mistral.rs:440`) and the embeddings path does not — it passes the joined id as the repo and hardcodes the revision to `None` (`crates/runtime/src/model/embed.rs:397`).

Filed as #12430 with the localization and a suggested fix. Not a regression — pinned embeddings have never worked; this PR is just the first thing to pin one. Options are to land #12430 first, or drop the pin on the `embeddings:` entry and keep it on the model.

Seen on merge-queue head `d65527864e836f678c630d91b16a01f44d7a74ad` (run 30852663452). Separately, the *older* failures on this job were a different cause — an orphaned `spiced` from a previous job holding :8090 — which is #12058 / #12431.

sgrebnov (dismissed):

PR #12262, opened 2026-07-31
fix(cayenne): always materialize a snapshot directory the catalog may reference (fixes #12208)

### LINKED ISSUE #12208: Cayenne: three snapshot-write paths fsync or publish a snapshot directory the Vortex sink never created
## Describe the bug

`CayenneTableProvider::write_to_snapshot` (`crates/cayenne/src/provider/table.rs:6080`) does **not** create the snapshot directory — the Vortex sink creates it lazily when it writes its first file. A write that carries no rows therefore produces no files and no directory, and any caller that afterwards fsyncs or publishes that snapshot fails or records a snapshot that does not exist on disk.

#11710 fixed one instance of this (the upsert append refresh, where an idle refresh with no new source rows made the dataset unhealthy with `No such file or directory`). It fixed the symptom at one call site. Three other call sites have the same shape, and the paths that are safe are safe only because they happen to call `ensure_snapshot_dir_exists` first — nothing enforces that.

This is a static-analysis finding: the reachability assessment per path is below, and only the append path in #11710 has a confirmed field failure.

## Full refresh and compaction are NOT affected

Recording this explicitly so the guard is not "fixed" into paths that do not need it. Every overwrite and compaction path pre-creates the directory before the write, so its later `sync_snapshot_dir` cannot fail:

| Path | Pre-creates dir | Skips commit on zero rows |
|---|---|---|
| Overwrite / full refresh (`overwrite.rs:305`) | ✅ | n/a — `PreparedOverwrite::finish` handles the zero-row overwrite deliberately |
| Sort rewrite (`table.rs:12417`) | ✅ | ✅ `total_rows == 0` → `remove_dir` + return |
| Full-snapshot compaction (`table.rs:15163`) | ✅ | ✅ |
| Protected-subset merge (`table.rs:16792`) | ✅ | ❌ — publishes an empty snapshot, which is correct (every input row was deleted) |
| Sequence-prefix bake (`table.rs:17371`) | ✅ | ❌ — same |
| Small-file subset rewrite (`table.rs:12865`) | ✅ via the hardlink step (`subset_rewrite_eligibility` guarantees a non-empty carry-forward set) | ❌ — same |
| Staged append (dir created in `staging_wal.rs:1432`) | ✅ | ✅ `moved_count > 0` at `table.rs:4656` |

### 1. Staged upsert commit — same failure, nothing bounds the row count

`stage_upsert_data` (`crates/cayenne/src/provider/staged_upsert.rs:502`) writes at `:522` without pre-creating the directory, and `CayenneStagedUpsert::commit` (`:224`) then calls `record_written_snapshot_sequence` at `:270` with no zero-row check. That helper fsyncs the snapshot directory (`table.rs:6033` → `:5099`), so the commit fails with `IoError`/`NotFound`.

Unlike the append writer, nothing upstream bounds the row count here — the stream is whatever the user's `INSERT` produced, so an empty result set reaches it directly.

Reachability today: `commit` is only called from cayenne's own unit tests (`table.rs:31543`, `:31602`). The live transaction path is the fused one (`prepare_commit` → `commit_fused` → `finish`), which does not fsync. So this is a live defect in an API that no production path currently reaches — worth fixing before one does.

### 2. Mem-tier and inline checkpoints — latent, held up by a distant invariant

Both fsync unconditionally with no pre-create:

| Call site | fsync |
|---|---|
| `insert_to_new_snapshot_with_sequence` (`table.rs:23655`), reached from `checkpoint_inlined_data` (`:23756`) | `table.rs:23687` |
| Mem-tier CDC checkpoint | `table.rs:22779` |

Neither can currently be reached with zero rows, but only because of an invariant two layers away: each caller guards on `batches.is_empty()`, and `filter_inlined_batch_for_deletions` (`table.rs:20262`) returns `None` for any zero-row batch, so a non-empty `Vec<RecordBatch>` implies a non-zero row count. That coupling is undocumented and not asserted anywhere.

If it ever breaks, the local-filesystem outcome is the #11710 failure. The S3 outcome is worse: the fsync is skipped there, so the write would publish a protected snapshot over a prefix that has no objects instead of failing loudly.

### 3. Fused transaction commit — the inverse gap (distinct bug, same subsystem)

`prepare_on_conflict_deletions_for_staged_snapshot` (`table.rs:10902`) never syncs the staged snapshot directory at all. The fused commit therefore writes the snapshot-sequence row and publishes the protected snapshot with **no directory barrier** — the durability step the non-transactional path takes via `record_written_snapshot_sequence`. A power loss can leave the catalog referencing a snapshot whose directory entries were never flushed.

With zero rows it also registers a snapshot whose directory does not exist. Scans tolerate that (`list_snapshot_files_with_sizes_local`, `table.rs:14060`, maps `NotFound` → empty), so it is clutter rather than a wrong query result — but the missing fsync is a real durability asymmetry against the sync path.

This one is not fixed by the change proposed below and could reasonably be split into its own issue.

## Expected behavior

- A write that produces no rows completes as a successful no-op on every path, and never records or publishes a snapshot that has no directory.
- `sync_snapshot_dir` can never be called on a directory that was never created — enforced structurally, not by a guard repeated at each call site.
- A staged snapshot published by the fused transaction commit is durable on the same terms as one published by the sync path.

## Proposed fix

Rather than adding a fourth copy of the `rows == 0` check, make the invariant structural: have `write_to_snapshot` call `ensure_snapshot_dir_exists` itself for non-S3 paths. Then:

- the directory always exists, so `sync_snapshot_dir` cannot fail with `NotFound` on any path, present or future;
- the six existing pre-create call sites become redundant no-ops (`ensure_snapshot_dir_exists` already short-circuits on `exists()`), and can be removed in the same change or left as harmless;
- the remaining `rows == 0` guards keep only their real purpose — skipping a pointless catalog commit — instead of preventing a crash.

`ensure_snapshot_dir_exists` fsyncs the parent directory on creation, so moving it inside `write_to_snapshot` moves that cost onto paths that previously skipped it. It runs once per write and is already on every hot staged/compaction path, but the CDC apply path should be measured before and after.

The fused-transaction fsync gap needs a separate change: sync the staged snapshot directory during `prepare_commit`, before the shared transaction makes it visible.

## To reproduce

No user-facing reproduction for the paths above — items 2 and 3 are not reachable with zero rows today, and item 1 is only reachable from a test. The regression tests follow the shape of the one #11710 added (`crates/cayenne/src/provider/sink.rs`, `zero_row_upsert_append_with_retention_filters_is_a_successful_noop`):

- a zero-row staged upsert stages and commits as a successful no-op, and the table stays writable afterwards;
- a zero-row write on each checkpoint path is a no-op that registers no snapshot (guards the invariant rather than relying on it);
- after a fused transaction commit, the staged snapshot directory has been fsynced — asserted the way `crates/cayenne/tests/ingest_fsync_regression_test.rs` already asserts the fsync contract on the staged append path.

## Related

- #11710 — fixed the same defect on the upsert append refresh path

claudespice:
Fixes #12208

## Summary (root cause)

`CayenneTableProvider::write_to_snapshot` did not create the snapshot directory — the Vortex sink created it lazily when it wrote its first file. A write carrying no rows therefore produced no files and no directory, so any caller that afterwards fsynced or published that snapshot either failed with `NotFound` or recorded a catalog entry for a snapshot with nothing on disk.

#11710 fixed one instance of this at one call site (the upsert append refresh, where an idle refresh made the dataset unhealthy with `No such file or directory`). Three other call sites had the same shape, and the paths that were safe were safe only because they happened to call `ensure_snapshot_dir_exists` first — nothing enforced it:

- **Staged upsert commit** — `stage_upsert_data` writes without pre-creating the directory, and `CayenneStagedUpsert::commit` then calls `record_written_snapshot_sequence`, which fsyncs it, with no zero-row check. Unlike the append writer nothing upstream bounds this row count: the stream is whatever the caller's `INSERT` produced, so an empty result set reaches it directly. Reachable today only from cayenne's own unit tests (the live transaction path is the fused one, which did not fsync) — a live defect in an API no production path reaches yet.
- **Mem-tier and inline checkpoints** — both fsync unconditionally with no pre-create. Neither is reachable with zero rows today, but only because of an invariant two layers away: each caller guards on `batches.is_empty()` and `filter_inlined_batch_for_deletions` returns `None` for a zero-row batch, so a non-empty `Vec<RecordBatch>` implies a non-zero row count. That coupling was undocumented and asserted nowhere. On a local filesystem, breaking it reproduces the #11710 failure; on S3 the fsync is skipped, so the write would instead publish a protected snapshot over a prefix holding no objects.

Separately, `prepare_on_conflict_deletions_for_staged_snapshot` never synced the staged snapshot directory, so the fused (multi-table transaction) commit wrote the snapshot-sequence row and published the protected snapshot with **no directory barrier** — the durability step the non-transactional path takes via `record_written_snapshot_sequence`. A power loss could leave the catalog referencing a snapshot whose directory entries were never flushed.

## Changes

**The invariant is now structural rather than a fourth copy of the `rows == 0` check.** `write_to_snapshot` materializes the snapshot directory itself, so "a snapshot the catalog can reference always has a directory" is a property of the write. `sync_snapshot_dir` can no longer be reached on a directory that was never created, on any path present or future, and the remaining `rows == 0` guards keep only their real purpose — skipping a pointless catalog commit — instead of preventing a crash.

`write_to_snapshot` is the right anchor because it is the single funnel every snapshot file goes through, and because the invariant is broader than "the fsync fails": there are two catalog-publish funnels (`record_written_snapshot_sequence` and the transaction-applied payload), so anchoring at either one would have left the other's shape unfixed.

**A second, previously-unnoticed gap closes with it.** The sink's lazy creation does not sync the parent directory, so before this change *every* newly created snapshot directory — not only zero-row ones — existed only in the page cache until some later directory sync happened to cover it, while the catalog was already committing a pointer to it. Routing creation through `ensure_snapshot_dir_exists` (which syncs the parent) closes that for all writes, and matches the contract the crate already enforces for the initial snapshot directory in `cayenne_catalog.rs`.

**The fused commit takes the missing directory barrier.** `CayenneStagedUpsert::prepare_commit` fsyncs the staged snapshot directory before the shared transaction makes it visible. Nothing after `stage_upsert_data` adds files to that directory — the on-conflict deletion vectors are written under the *current* snapshot's `_deletes/` — so the barrier is complete there.

The barrier is deliberately **not** pushed into `prepare_on_conflict_deletions_for_staged_snapshot`: its other two callers are staged *appends*, whose files still live in a staging directory at that point and whose target directory gets its barrier later from `move_staged_files_to_snapshot`. Syncing there would fsync a directory that may not exist yet and add a sync to the CDC apply path. `prepare_commit` rather than `apply_in_txn` for a second reason the surrounding code states explicitly: the fused path must not hold the single SQLite writer across I/O.

**Two named helpers replace the repeated guard.** `ensure_local_snapshot_dir` / `sync_local_snapshot_dir` own the "an object store has no directories" decision, so it lives in one place instead of being open-coded at each new site; `record_written_snapshot_sequence` moves onto the sync helper. `encode_snapshot_shards_concurrently` materializes the shared directory once before fanning out — its N spawned encodes all target the same snapshot, so otherwise each would race to create it and each would fsync the parent.

`ensure_snapshot_dir_exists` now probes with `tokio::fs::try_exists` instead of `Path::exists`. This runs on every write after this change, and on the network-attached storage tier the crate explicitly supports a cold `stat` can take milliseconds — long enough to stall a Tokio worker.

Full refresh and compaction are unaffected — every overwrite/compaction path already pre-creates the directory, so its later `sync_snapshot_dir` could not fail. Those paths now pay one `stat` and keep their existing behavior, including the ones that deliberately publish an empty snapshot because every input row was deleted. The six existing pre-creates are left in place: each wants the directory for work other than the `write_to_snapshot` call (the subset rewrite hardlinks its carry-forward set into it first).

## Performance impact

The added work is one `create_dir_all` plus one ordering-tier parent-directory fsync, and only when the directory does not already exist. It is taken *before* the process-global encode-permit acquisition, so it does not hold that budget across metadata I/O.

- **Mem-tier checkpoint, inline checkpoint, staged upsert** — one sync per checkpoint, and it buys the parent-dirent durability described above rather than costing throughput.
- **Row-bearing staged append (the CDC hot path)** — a genuine addition of one parent-directory fsync per append, since `clear_staging_snapshot_dir` removes the staging directory before each one. (On a *zero-row* staged append the directory was already created and parent-synced by `write_staging_wal_local`, so nothing changes there.) On macOS this is plain `fsync(2)` — measured at ~66 µs in `provider/fsync_tier.rs`, against a multi-millisecond Vortex encode; on Linux it is a directory `fsync` on an inode already dirtied by the `create_dir_all`.
- **Fused multi-table commit** — one directory fsync per participant. The transaction driver prepares participants sequentially under the full lock set, so N participants add N syncs to that window. The directories are disjoint and the syncs could be joined, but the sequential prepare loop is deliberate (canonical lock order, single metastore writer) and restructuring it is outside this fix.

**I have not run the HTAP CH-benCHmark for this**, which the issue asks for. If that number moves, the cheap fallback is to move the call *after* the write and gate it on `total_rows == 0`: the sink creates the directory whenever it writes a file, so the invariant would still hold and non-empty writes would pay nothing. I did not do that here because it makes the invariant depend on the sink's lazy-creation behavior — the same kind of distant coupling the issue objects to in the checkpoint paths — and it gives up the parent-dirent fix.

## Test plan

- `make lint`
- `make nextest`

New tests in `crates/cayenne/src/provider/table.rs`:

- `zero_row_staged_upsert_commits_as_a_successful_noop` — the regression test. A zero-row staged upsert stages and commits as a successful no-op, visible data is unchanged, and the table stays writable afterwards. Verified to fail on the old code with `IoError { NotFound, "No such file or directory" }` — the #11710 signature.
- `zero_row_write_still_materializes_its_snapshot_directory` — asserts the invariant directly: a zero-row staged write materializes exactly one snapshot directory on disk, and `prepare_commit` accepts that snapshot. Also verified to fail on the old code (zero new directories).
- `checkpointing_an_empty_table_registers_no_snapshot` — pins the observable contract of the two checkpoint paths, which were guarded from the zero-row shape only by the distant `filter_inlined_batch_for_deletions` coupling. A change to that coupling now shows up as a test failure rather than as an `IoError` in the field.

New test in `crates/cayenne/tests/ingest_fsync_regression_test.rs`:

- `prepare_commit_syncs_the_staged_snapshot_directory` — structural, following that file's existing checks, because an fsync is not observable from a test without stubbing the filesystem. Pinned on the `sync_local_snapshot_dir` helper rather than an inlined fsync plus its `s3://` guard, so it does not penalize that factoring.

## Docs

`docs/cayenne/cayenne.md`: the *Serializable transactions and per-key optimistic concurrency* subsection now records the fused commit's directory barrier, plus a *Document changelog* row.

## Checklist

- [x] Tests added (regression + invariant + checkpoint contract + structural fsync contract)
- [x] Regression tests verified to fail on the unfixed code
- [x] `docs/cayenne/cayenne.md` reviewed and updated, with a changelog row
- [x] No new configuration parameters
- [x] No metastore schema change


claudespice:
@copilot review

lukekim (approved):

Issue #12267, opened 2026-07-31
Elasticsearch deletes filter on unmapped primary-key columns, so a string key silently deletes nothing


claudespice:
## Describe the bug

Deleting from an Elasticsearch search or vector index issues a `_delete_by_query` built from `term` filters on the primary-key columns (`build_or_of_row_term_queries`, `crates/search/src/index/elasticsearch/delete.rs`):

```json
{"bool": {"should": [{"bool": {"filter": [{"term": {"id": "<value>"}}]}}], "minimum_should_match": 1}}
```

But the index mapping the runtime creates never mentions the primary-key columns. `ensure_index_with_mapping` (`crates/runtime/src/embeddings/index/elasticsearch/mod.rs`) puts only three things in `properties`: the `dense_vector` field, the configured `text_fields`, and user-declared metadata columns. There are no dynamic templates, and `dynamic` is left at its default.

So a primary-key column is dynamically mapped by Elasticsearch:

- An integer key becomes `long` — a `term` query matches exactly, and deletes work.
- A **string** key becomes `text` (with a `keyword` sub-field). `term` is not analyzed, but the inverted index for a `text` field holds *analyzed* tokens, so `{"term": {"id": "ORDER-1024"}}` looks for the single token `ORDER-1024` in an index that holds `[order, 1024]` and matches nothing.

A dataset with a string primary key containing uppercase letters or punctuation — a UUID, an ARN, an order code, an email — therefore gets a `_delete_by_query` that deletes nothing while reporting success. On the CDC path the index delete is best-effort (logged, not propagated), so nothing surfaces at all. Search keeps returning rows the dataset no longer has, the same user-visible symptom as #12066 / #12088.

Documents themselves are addressable: the write path already derives each document's `_id` from the primary key (`extract_primary_key`, JSON-encoded for a composite key), which is how re-indexing the same row stays idempotent.

## Expected behavior

Deleting a row removes its Elasticsearch documents, whatever the primary key's type.

## Possible approach

- For an exact-key delete, address documents by `_id` — the same value the write path computes — instead of by `term` filters on the key columns. That sidesteps mapping and analysis entirely.
[...]

Issue #12266, opened 2026-07-31
A chunked index resolves its chunk entries from the warm tier, so a delete misses vectors the warm index does not hold


claudespice:
## Describe the bug

A chunked index stores one document per chunk, keyed by the base row key plus `_spice.chunk_id`. A delete only carries the base key, so `ChunkedSearchIndex`/`ChunkedVectorIndex` resolve the exact chunk-keyed entries first, by scanning the inner index's `VectorIndex::list_table_provider` under a predicate on the base key (`delete_chunked_vector_by_outer_keys`, `crates/search/src/index/chunking.rs`).

For S3 Vectors with a warm in-memory index, that inner index is a `CompoundVectorIndex { primary: MemoryVectorIndex, secondary: S3Vector }` — `crates/runtime/src/embeddings/index/table.rs` calls `with_memory_warm_index` and then `construct_chunked_vector_index`, so the compound is *inside* the chunked wrapper. `CompoundVectorIndex::list_table_provider` serves the **read** plan:

- `CompoundReadMode::PrimaryOnly` (`on_zero_results: return_empty`) → the warm in-memory index only.
- `CompoundReadMode::FallbackToSecondary` (`on_zero_results: use_source`) → `fallback_on_empty_plan(memory, s3)`.

Neither is the union, and neither is authoritative. The warm tier only holds rows the acceleration write path has passed through it, so any base key it does not hold resolves to zero chunk rows — and `Index::delete_by_keys` is then never called for them, leaving those vectors in S3. `delete_by_keys` reports success.

The narrowest case: `PrimaryOnly` with a persisting accelerator. After a restart nothing re-writes the existing rows through the index (#12102), so the warm tier is empty or partial while S3 still holds every vector, and deletes silently do nothing for the rows it is missing. Under `FallbackToSecondary` a *completely* empty warm plan does fall back to S3, but a partially-populated one does not — the fallback is per-plan, not per-key.

Symptom is the same as #12066 / #12088: search keeps returning rows the dataset no longer has.

## Expected behavior

A row deleted at the source has all of its chunk entries removed from the durable store, regardless of what the warm tier happens to hold.

## Possible approach

Resolve against the authoritative store rather than the read plan. Either

- give `VectorIndex` an authoritative-listing accessor that `CompoundVectorIndex` maps to its secondary (defaulting to `list_table_provider` everywhere else), or
[...]

Issue #11964, opened 2026-07-21
Integration tests for warm-index deletes (retention/CDC/direct-delete) across all backends


Jeadie:
Sub-issue of #11828. Follow-up to #11888.

## Problem

#11888 added delete support to `Index`/`SearchIndex`/`VectorIndex` (`delete_by_keys`, `delete_by_key_prefix`, `delete_by_predicate`, `delete_warm_by_keys`) and wired it into every call site that removes rows: `IndexedTableProvider::delete_from` (direct SQL `DELETE`), CDC's Cayenne fast-delete path, and retention (warm-only scope). That PR added unit coverage for the shared plumbing (the predicate-resolution helper, `CompoundSearchIndex`'s both-vs-warm-only fan-out, the CDC key-extraction helper, retention's provider-unwrapping), but left real coverage gaps:

1. **No test exercises the backend implementations themselves.** S3 Vectors' `delete_by_keys`/`delete_by_key_prefix` (the exact-delete and list-then-delete-by-prefix paths), Elasticsearch's `delete_by_keys` (via the new `_delete_by_query` client method), and the tantivy-backed full-text index's `delete_terms_for` are all compile-checked only — nothing proves they actually remove the right entries.
2. **`IndexedTableProvider::delete_from` itself is untested.** The things that call into it (retention, CDC's generic-fallback branch) are tested, but there's no test that directly attaches a `SearchIndex` to an `IndexedTableProvider` and confirms `delete_from` invokes `delete_by_predicate` on it.
3. **No end-to-end test.** A real `DELETE FROM dataset WHERE ...` (or a CDC delete, or a short `retention_period` expiring rows) against a running dataset with `vectors.enabled`, confirmed via the vector search HTTP/SQL surface to no longer return the deleted rows from the warm index. This scenario currently has zero coverage — the correctness property #11888 exists to fix is unverified end-to-end.

## Proposed work

- Unit tests per backend: reuse the existing S3 Vectors mock client (`crates/s3_vectors/src/mock.rs`) to verify `delete_by_keys`/`delete_by_key_prefix`; add an Elasticsearch client mock (or extend the existing `MockElasticsearch` test double in `data_components::elasticsearch::query_table`) to verify `_delete_by_query` is called with the right filter; a tantivy-backed test verifying `delete_terms_for` actually removes matching documents from the index.
[...]

PR #12268, opened 2026-07-31
fix(search): remove all chunk documents when a chunked Elasticsearch index deletes a row (fixes #12088)

### LINKED ISSUE #12088: A chunked search index keeps documents for rows a CDC stream deleted
## Describe the bug

A CDC change batch now carries a row-aligned deletion mask to every index (`Index::compute_index_for_changes`), and `FullTextDatabaseIndex` uses it to remove the documents for rows the source deleted instead of re-indexing them as upserts (#12084).

`ChunkedSearchIndex`/`ChunkedVectorIndex` cannot forward that mask, and deliberately do not:

- Chunking expands one base row into one document per chunk, so a mask aligned with the base rows does not align with the rows the inner index sees.
- A chunk document is keyed by primary key **and** chunk id, so the inner index's `delete_term` on the base key does not identify the chunk documents to remove — and a delete carries no content to chunk, so the chunk ids cannot be recomputed from the change row.

A chunked index therefore keeps documents for rows the source has deleted. With `text_search`/`vector_search` left-joining hits to the base table, those stale documents consume result slots and (until the join is tightened) surface as hits whose base-table columns are all NULL.

## Expected behavior

A row deleted at the source is removed from a chunked index too — every chunk document for that key.

## Possible approach

Delete by key prefix rather than by exact unique-key term: index the base primary key as its own term-searchable field in the chunked schema and remove all documents matching it (tantivy's `IndexWriter::delete_query` over a `BooleanQuery` of the base key terms can express this, though it is `#[doc(hidden)]`). The alternative is to record chunk counts per key so the exact unique-key terms can be reconstructed.

## Additional context

Documented at `ChunkedSearchIndex::compute_index_for_changes` in `crates/search/src/index/chunking.rs`. Related: #12084 (the full-text fix), #12066 (a full refresh leaves deleted rows in a search index).

claudespice:
## Summary

A chunked search index stores one document per chunk, keyed by the source row's primary key **plus** `_spice.chunk_id`. A delete only ever carries the base row key — the chunk ids are not on the base table at all (`ChunkedSearchIndex::required_columns` strips `_spice.chunk_id`, so nothing can resolve it from the source), and a CDC delete carries no content to re-chunk.

`ChunkedSearchIndex::delete_by_keys` handled that by resolving the exact chunk-keyed rows out of the inner index's own `VectorIndex::list_table_provider`, then deleting those. **Elasticsearch cannot enumerate its vectors**, so `ElasticsearchIndex::list_table_provider` returns a correctly-shaped *empty* table:

```rust
// Elasticsearch doesn't support listing all vectors efficiently.
// Return an empty plan with the correct schema.
```

The resolve therefore matched nothing, the delete loop never ran, and `delete_by_keys` returned `Ok(())`. Every path that removes a row from a dataset with a chunked Elasticsearch vector column — CDC delete, SQL `DELETE`, retention — reported a successful delete while leaving every chunk document in Elasticsearch. On the CDC path the index delete is best-effort (logged, not propagated), so there was not even an error to see: the documents just stayed, and searches kept returning rows the dataset no longer has.

Elasticsearch can do exactly what is needed here, and `elasticsearch::delete` already says so — it just was never reached:

> an exact-key delete when `key_columns` is every `primary_key` column, **a prefix delete when it's a strict subset (the chunked-index case)**

`_delete_by_query` filters by field value, so a query naming only the base key removes every chunk document under it in one request, without anyone having to know how many chunks a row produced. That is also the approach confirmed on the issue.

## Changes

- **`Index::deletes_by_partial_key`** (new, defaults to `false`): whether `delete_by_keys` removes every entry matching the key columns present in the batch, even when those are a strict subset of the index's own key. `false` keeps the resolve-first path for stores addressed by an exact key (S3 Vectors keys each vector by its full composite key).
- **`ChunkedSearchIndex` / `ChunkedVectorIndex`**: hand the base key straight to a partial-key-capable inner index; otherwise resolve the chunk-keyed rows as before. Both also forward the capability, rather than inheriting the default and silently sending a capable index down the enumerate-first path.
- **`ElasticsearchIndex` / `ElasticsearchTextIndex`**: report the capability, and address documents by `document_key_columns` — the primary key minus `_spice.chunk_id`. For a non-chunked index the chunk key isn't in the primary key, so this is the full key and nothing changes.
- **`CompoundSearchIndex` / `CompoundVectorIndex`**: forward the capability as the intersection of both halves — `delete_by_keys` fans out to both, so a partial key only clears the compound index if both act on one.
- **`ChunkedSearchIndex::base_key_columns`**: "the primary key minus the chunk key" was open-coded in three places (the delete resolve, `ChunkedVectorIndex::list_table_provider`, and now Elasticsearch's document addressing). One helper, in the module that owns the chunk key.

Two adjacent defects found while doing this, filed rather than folded in:

- **#12266** — when a chunked index's inner index is a warm-tier compound (S3 Vectors plus an in-memory warm index), the resolve enumerates the *warm* tier, so a base key it does not hold resolves to nothing and those vectors survive in S3. Needs the resolve to run against the authoritative store, which is a different mechanism from this fix and touches warm-tier code under active change.
- **#12267** — the runtime's Elasticsearch mapping never mentions the primary-key columns, so a string key is dynamically mapped as `text` and the `term` filter this delete builds cannot match it. That applies equally to the non-chunked delete path, and needs a mapping or `_id`-addressing decision. **So for such a key this fix is necessary but not sufficient**; for an integer key (and for a key that survives the standard analyzer unchanged) the chunked delete is complete with this change.

### Performance impact

Strictly cheaper on the fixed path: a chunked Elasticsearch delete no longer plans and executes a scan-and-collect over the inner index's list plan before deleting — it issues one `_delete_by_query` per 512-key slice, the same request shape the non-chunked path already uses.

## Test plan

- `make lint`
- `make nextest`
- New unit tests in `crates/search`:
  - `a_partial_key_inner_index_deletes_from_the_base_key_alone` — the regression test. The inner index's listing is an empty, correctly-shaped table (Elasticsearch's actual behavior), so on the old code the resolve matched nothing and *zero* deletes reached the inner index; it now receives the base key unchanged. Also run with a non-empty listing, to pin that the shortcut does not depend on what the inner index lists.
  - `an_exact_key_inner_index_resolves_its_chunk_keyed_entries_first` — the resolve path still works and removes only the deleted row's chunks, chunk id included.
  - `a_chunked_index_over_an_unlistable_inner_index_reports_not_implemented` — an index with neither a listing nor a partial-key delete still errors instead of reporting a delete it did not perform.
  - `chunked_wrappers_forward_partial_key_deletion`, `partial_key_deletion_requires_both_halves` — wrapper forwarding for both chunked and both compound wrappers.
  - `a_chunked_index_deletes_every_chunk_of_a_base_key`, `the_full_composite_key_cannot_address_a_base_key_batch`, `a_present_chunk_id_is_not_filtered_on`, `document_key_columns_*` — the Elasticsearch query is filtered on the base key only, and the old full-composite key columns could not build a query against a base-key batch at all.

Not covered: a live Elasticsearch integration test. #11964 tracks warm-index delete coverage across backends.

## Checklist

- [x] Sibling arms audited (S3 Vectors' listing is real, so its resolve path is kept; the warm-tier compound gap is filed separately)
- [x] Wrapper implementations forward the new trait method
- [ ] `make lint` and `make nextest` green (sign-off)

Fixes #12088


claudespice:
@copilot review

claudespice:
@copilot review

lukekim (dismissed):

claudespice:
Merged `trunk` to resolve the conflict with #12273, which landed the `_id`-based Elasticsearch delete.

The two changes compose rather than overlap. `delete_by_keys` now addresses documents by `_id` when `key_columns` covers the whole primary key, and keeps the row-term query only for a **strict subset** of it — and a strict subset of the primary key is exactly the chunked-index case this PR resolves the base key for:

| index | `key_columns` passed | vs `primary_key` | path taken |
|---|---|---|---|
| non-chunked | full key | equal | `_id` (`ids` query), as #12273 introduced |
| chunked | key minus chunk key | strict subset | row-term query on the base key |

So a chunked index filters on the base key alone and removes every chunk document stored under it, while a non-chunked index keeps the `_id` addressing that made the exact-key delete work for string keys.

Conflicts resolved:
- `elasticsearch/delete.rs` — kept both sides: `document_key_columns` alongside trunk's `_id`/term split, and both test modules.
- `elasticsearch/mod.rs` — both callers now pass `document_key_columns(&self.primary_key)` as `key_columns` through trunk's wider signature.

Verified locally on the merge commit: `cargo test -p search --features elasticsearch --lib` → **119 passed, 0 failed**, including all five chunked-key tests from this branch and trunk's `_id` tests. `cargo check --lib --tests` clean with zero warnings. Sign-off re-dispatched for the new head.

Unchanged by this merge: the chunked path still filters on key *columns*, so it remains subject to the primary-key mapping gap tracked in #12272 — as this PR already noted.

claudespice:
This PR is fully green and blocked only on a re-approval.

- `Attestation` ✅ and `signoff` ✅ on the current head `c42a49b17`; every non-`SKIPPED` check passes (the heavy checks run in the merge queue).
- Zero unresolved review threads.
- @lukekim's approval was **dismissed by a base sync, not by a code change**: the only commit after it is `c42a49b17`, a `Merge branch 'trunk'`. The PR's own diff is unchanged — 7 files, all under `crates/search/` plus `crates/runtime-datafusion-index/src/lib.rs`.

A readiness check reports no hard blockers, so a maintainer re-approval is the last step. Flagging rather than re-approving: a bot approval cannot satisfy CODEOWNERS here, so it would leave `REVIEW_REQUIRED` in place and only add noise.

lukekim (approved):

Issue #12272, opened 2026-07-31
Elasticsearch index mapping never maps the primary-key columns, so a partial-key delete cannot match a string key


claudespice:
## Describe the bug

`ensure_index_with_mapping` (`crates/runtime/src/embeddings/index/elasticsearch/mod.rs`) puts only three things in the index `properties`: the `dense_vector` field, the configured `text_fields`, and user-declared metadata columns. There are no dynamic templates and `dynamic` is left at its default, so the primary-key columns are never mapped.

Elasticsearch then maps a string key dynamically as `text`, whose inverted index holds analyzed tokens. An unanalyzed `term` query on that field cannot match a key that the analyzer splits — `{"term": {"id": "ORDER-1024"}}` searches for one token in an index holding `[order, 1024]`.

#12267 removed the consequence for the **exact-key** delete by addressing documents by `_id` instead. The **partial-key** delete cannot do that: the chunked-index case (#12088 / #12268) knows only the base key, and the chunk id is part of `_id`, so it still filters on the key columns and still cannot match such a key.

## Expected behavior

A partial-key (chunked) delete removes the documents for a row whatever the primary key's type.

## Possible approach

Map the primary-key columns explicitly — as `keyword`, or query a `.keyword` sub-field.

Two things make this more than a one-line mapping addition:

- **`keyword` has an `ignore_above`**: a value longer than the limit is not indexed at all, so it would silently fail to match in the same way.
- **An existing index cannot be repaired by `put_mapping`** — Elasticsearch cannot change a field's type in place, and the current code already treats a failed `put_mapping` as best-effort (logs a warning and continues with the existing mapping). Any fix needs a story for indexes that already exist, whether that is a reindex, a versioned index name, or a startup diagnostic that tells the user their index predates the fix.

Note that querying `.keyword` happens to work for an index whose key was dynamically mapped (the dynamic default adds that sub-field), but not for one a user pre-created with an explicit `keyword` mapping — where the bare field name is the correct one. So the sub-field cannot simply be assumed.

## Additional context

Split out of #12267 so that fix could stay in `crates/search`; this one is a `crates/runtime` mapping change plus a migration decision.

claudespice:
Follow-up analysis while triaging this: the mapping is not the only place a fix could go, and the query side is both cheaper and repairs indexes that already exist.

[...]

Issue #12066, opened 2026-07-26
A full refresh leaves search-index documents for rows the source deleted, so search returns rows the dataset no longer has


claudespice:
## Summary

A `refresh_mode: full` refresh (the default) replaces every row in the accelerated table, but the search index write lifecycle never learns that a row disappeared. `compute_index` only deletes the primary keys **present in the incoming batch**, so a row that the source deleted keeps its document in the warm index and stays searchable forever — with its stale stored content.

Because full-text results are materialized from the index's own stored fields and are never joined back to the accelerated table, there is nothing downstream to filter the phantom rows out. `SELECT` on the dataset correctly shows the row is gone; `/v1/search` and the full-text search surface still return it.

## Reproduction

Two sink write windows against a `FullTextDatabaseIndex`, the second omitting a primary key (exactly what a full refresh does after the source row is deleted):

```rust
#[tokio::test]
async fn full_refresh_keeps_documents_for_rows_the_source_dropped() {
    let index = FullTextDatabaseIndex::try_new(
        create_test_table(),
        vec!["content".to_string()],
        Some(vec!["id".to_string()]),
        None,
        &["content".to_string()],
    )
    .expect("Failed to create FullTextDatabaseIndex");

    // Refresh 1 (overwrite): the source has ids 1, 2, 3.
    index.on_write_start().await.expect("on_write_start failed");
    index
        .compute_index(vec![
            record_batch!(
                ("id", Int32, [1, 2, 3]),
                (
                    "content",
                    Utf8,
                    [
                        "apple banana cherry",
                        "dog elephant frog",
                        "guitar harmonica instrument"
                    ]
                )
            )
            .expect("Failed to create test batch"),
        ])
        .await
        .expect("compute_index failed");
    index
        .on_write_complete()
        .await
        .expect("on_write_complete failed");

    // Refresh 2 (overwrite): id=2 was deleted at the source, so the refresh query
    // returns only ids 1 and 3 and the accelerated table is replaced with them.
    index.on_write_start().await.expect("on_write_start failed");
    index
        .compute_index(vec![
            record_batch!(
                ("id", Int32, [1, 3]),
                (
                    "content",
                    Utf8,
                    ["apple banana cherry", "guitar harmonica instrument"]
                )
            )
            .expect("Failed to create test batch"),
        ])
        .await
        .expect("compute_index failed");
    index
        .on_write_complete()
        .await
        .expect("on_write_complete failed");

    let search_index = index
[...]

PR #12273, opened 2026-07-31
fix(search): address Elasticsearch deletes by document _id (fixes #12267)

### LINKED ISSUE #12267: Elasticsearch deletes filter on unmapped primary-key columns, so a string key silently deletes nothing

claudespice:
## Summary

Deleting a row from an Elasticsearch search or vector index could silently delete nothing.

The delete built a `_delete_by_query` from `term` filters on the primary-key columns:

```json
{"bool": {"should": [{"bool": {"filter": [{"term": {"id": "ORDER-1024"}}]}}], "minimum_should_match": 1}}
```

But nothing maps those columns. `ensure_index_with_mapping` puts only three things in `properties` — the `dense_vector` field, the configured `text_fields`, and user-declared metadata columns — with no dynamic templates. A primary-key column is therefore mapped by Elasticsearch's dynamic defaults:

- an integer key becomes `long`, and `term` matches it exactly;
- a **string** key becomes `text`, whose inverted index holds *analyzed* tokens. `term` is not analyzed, so `{"term": {"id": "ORDER-1024"}}` looks for the single token `ORDER-1024` in an index holding `[order, 1024]` — and matches nothing.

So a dataset keyed by a UUID, an ARN, an order code, or an email got a delete that reported success having removed no documents. On the CDC path the index delete is best-effort (logged, not propagated), so nothing surfaced at all: search kept returning rows the dataset no longer had — the same user-visible symptom as #12066 / #12088.

The documents were always addressable. The write path derives each document's `_id` from the primary key (that is how re-indexing a row stays idempotent), so an exact-key delete can address them by `_id` and sidestep mapping and analysis entirely.

Fixes #12267

## Changes

- `delete_by_keys` now takes the index's full primary key alongside the columns being matched. When those columns cover the whole primary key, it addresses documents with an `ids` query built from the `_id`s the write path stores; otherwise it keeps the existing `term`-filter behaviour.
- The `_id`s are derived by calling the *same* `extract_primary_key_from_fields` the write path uses, rather than a second copy of the rule, so the two cannot drift. `write.rs` held two byte-for-byte-equivalent derivations (`extract_primary_key` and `extract_primary_key_from_fields`); the duplicate is removed and the survivor documents the invariant.
- Rows with a NULL key yield no `_id` — the write path never stored a document for them — and a chunk that addresses nothing issues no request at all, rather than an unconstrained query that would match everything.

Deletes are also cheaper now: an `ids` query is a direct document lookup, where the previous query scanned an analyzed field per key.

### Deliberately not changed

A **strict subset** of the primary key cannot use `_id`, because the chunk id is part of it and unknown at delete time. That path (the chunked-index case introduced by #12268) still filters on key columns and so remains subject to the same mapping gap. Making it work needs the key columns to be term-queryable — mapping them as `keyword`, or querying a `.keyword` sub-field — which also needs a story for indexes that already exist, since `put_mapping` cannot change an existing field's type. That is a separate change; filed as #12272.

The `term`-filter builder is therefore kept, not deleted: it is the code path that case uses.

## Overlap with #12268

#12268 (open, same area) adds the chunked partial-key delete route and appends a test module to the same file. The two changes are complementary — this PR fixes the *whole-key* path, #12268 adds the *partial-key* path — and touch different functions, but whichever lands second will need a small merge in `delete.rs`.

## Test plan

Ten tests in `crates/search/src/index/elasticsearch/delete.rs`, all new:

- **regression** — a string key whose analyzer would split it (`ORDER-1024`, `a716-446655440000`) is deleted by `ids`, not by a `term` filter. Verified to fail against the previous behaviour: neutering the fix reproduces exactly the broken `term` query from the issue.
- a composite key uses the JSON-encoded `_id`, pinned as a literal so a drift between the write and delete derivations breaks the test.
- the ids a delete addresses equal the ids the write path derives.
- a partial key still falls back to `term` filters.
- NULL keys issue no request; NULL keys mixed with present keys delete only the present ones.
- an empty batch and an empty primary key issue no request.
- a batch larger than the chunk size splits into bounded requests that together address every key exactly once.
- an integer key — which was never broken — still deletes.

Validation:

- `make lint-rust` (scoped to the touched crate) — green, both clippy passes.
- `make test` — `crates/search` lib suite green: 109 passed, 0 failed.
- Full `make lint` + `make nextest` run via `make signoff`.

Not exercised against a live Elasticsearch cluster: the analyzed-`text`-vs-`term` behaviour is Elasticsearch's documented dynamic-mapping default rather than something this repo asserts. The fix does not depend on that behaviour being reproduced — it stops querying the mapped field at all.

## Checklist

- [x] Edge cases covered (NULL keys, empty batch, empty primary key, chunk-boundary split, composite key)
- [x] No `.unwrap()` / `.expect()` in production code
- [x] Scoped lint green (both clippy passes)
- [ ] `make signoff` green and `Attestation` re-run

### DIFF
--- a/crates/search/src/index/elasticsearch/delete.rs
+++ b/crates/search/src/index/elasticsearch/delete.rs
@@ -15,20 +15,39 @@ limitations under the License.
 */
 
 use arrow::array::RecordBatch;
+use arrow_schema::Field;
 use datafusion::common::ScalarValue;
 use datafusion::error::{DataFusionError, Result as DataFusionResult};
 use elasticsearch::Elasticsearch;
 use serde_json::{Value, json};
 use snafu::ResultExt;
+
+use super::write;
 /// Chunk size for `_delete_by_query` requests — keeps each request's `bool.should` clause count
 /// comfortably under Elasticsearch's default `indices.query.bool.max_clause_count` (1024) and
 /// request-size limits, regardless of how many keys the caller is deleting in one call.
 const DELETE_CHUNK_ROWS: usize = 512;
 
 /// Deletes every document whose `key_columns` match a row of `keys` — an exact-key delete when
-/// `key_columns` is every `primary_key` column, a prefix delete when it's a strict subset (the
-/// chunked-index case). Elasticsearch's `_delete_by_query` filters by field value directly, so
-/// both cases are the same operation — there is no separate "exact" vs "prefix" code path.
+/// `key_columns` covers every `primary_key` column, a prefix delete when it's a strict subset
+/// (the chunked-index case).
+///
+/// The two cases address documents differently, because filtering on a key *column* is not
+/// reliable. Nothing maps the primary-key columns: the index mapping the runtime creates covers
+/// the vector field, the configured text fields, and declared metadata columns only, so a string
+/// key is dynamically mapped `text` and its inverted index holds *analyzed* tokens. A `term`
+/// query is not analyzed, so `{"term": {"id": "ORDER-1024"}}` looks for one token in an index
+/// holding `[order, 1024]` and matches nothing — a delete that reports success having removed
+/// no documents (#12267).
+///
+/// So when `key_columns` covers the whole primary key, this addresses documents by `_id` via an
+/// `ids` query. `_id` is the value the write path already stores for the row, derived by the
+/// same [`write::extract_primary_key_from_fields`], so the delete matches exactly the documents
+/// the write produced — no field mapping, no analysis, and no dependence on the key's type.
+///
+/// A strict subset of the key (the chunked-index case) cannot use `_id`, because the chunk id is
+/// part of it and is unknown at delete time; that case still filters on the key columns and so
+/// remains subject to the mapping gap above.
 ///
 /// Only reads `key_columns` from `keys`, ignoring any other column present — `keys` may be
 /// shaped by [`runtime_datafusion_index::Index::required_columns`] (a superset of the primary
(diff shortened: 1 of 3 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12278, opened 2026-07-31
fix(search): score retrieval metrics against the judged documents (fixes #12277)

### LINKED ISSUE #12277: Search benchmark retrieval metrics overstate quality: ideal DCG built from retrieved docs, and unanswered queries leave the average
### Describe the bug

The search benchmark's four retrieval metrics (`NDCG@k`, `Recall@k`, `MRR@k`, `Precision@k`, in `crates/test-framework/src/spicetest/search/evaluate.rs`) are computed in a way that systematically **overstates** retrieval quality, and returns `NaN` in one case. Every number reported by `testoperator run search` — the `score`/`recall`/`mrr`/`precision` telemetry gauges and the `SearchRunMetric` columns — is affected.

There are two independent defects.

**1. The ideal DCG is computed from the retrieved documents instead of the judged ones.**

`calculate_retrieval_metrics` builds `ranked_relevance` from the documents search actually returned, then `ndcg_at_k` derives the ideal DCG from that same vector:

```rust
fn idcg_at_k(relevance_scores: &[f64], k: usize) -> f64 {
    let mut sorted_relevance_scores = relevance_scores.to_owned();   // retrieved docs only
    sorted_relevance_scores.sort_by(|a, b| b.partial_cmp(a).unwrap_or(Ordering::Equal));
    dcg_at_k(&sorted_relevance_scores, k)
}
```

The ideal ranking must rank **all judged documents** from the qrels, independent of what search returned — otherwise a result set that misses relevant documents cannot score below 1.0. MTEB/`pytrec_eval` rank the full qrels. `recall_at_k` in the same file already takes the whole `relevance` map, so the data is at hand.

**2. A judged query with no search results is dropped from the average instead of scoring 0.**

```rust
let Some(ranked_results) = results.get(query_id) else {
    println!("No search results found for test query {query_id}");
    continue;                       // query leaves the average entirely
};
```

`average()` then divides by the number of queries *that returned something*, so unanswered queries raise the mean rather than lowering it. This affects **all four** metrics, not just NDCG. When no query returns results the vectors are empty and `average` evaluates `0.0 / 0.0` → **`NaN`**, which is then written to the OTLP gauges and the Arrow `Float64` metric columns.

### To Reproduce

Pure functions; no runtime needed.

**Defect 1** — one query, two relevant documents, search returns one of them at rank 1:

```
qrels   q1 = { doc0: 1, doc1: 1 }
results q1 = { doc0: 0.9 }
```

- Current: `DCG@10 = 1/log2(2) = 1.0`, ideal taken from `[1.0]` → also `1.0` ⇒ **`NDCG@10 = 1.0`** (a perfect score for 50% recall).
- Correct: ideal from the judged set `[1, 1]` = `1/log2(2) + 1/log2(3) = 1.63093` ⇒ `NDCG@10 = 0.61315`.

**Defect 2** — two judged queries, only one answered (perfectly):

```
qrels   = { q1: { doc0: 1 }, q2: { doc1: 1 } }
results = { q1: { doc0: 0.9 } }
```

- Current: `NDCG@10 = Recall@10 = MRR@10 = 1.0` (q2 discarded).
- Correct: `0.5` for each.

**`NaN` case** — any run where no judged query returns results yields `NaN` for all four metrics.

### Expected behavior

- The ideal DCG ranks every judged document for the query, so missing relevant documents drive `NDCG@k` below 1.0.
- A judged query with no results scores 0.0 for all four metrics and stays in the average.
- An empty `qrels` is a defined error rather than a silently-published `NaN`.

### Impact

Benchmark/telemetry only — no effect on user query results. It does, however, mean the recorded search-quality numbers are not comparable to MTEB's published figures and that a genuine retrieval regression can be masked: a change that stops returning results for some queries can leave the score flat or raise it.

### Additional context

`Precision@k` dividing by `k` (rather than by the number retrieved) is the standard TREC convention and is **not** part of this report.

#12226 fixes defect 1 and defect 2 for `NDCG` and confirms both readings ("skipping it would inflate the overall score"; "a result set that misses relevant documents must score below 1.0"). It is based on a tree that predates #12193, so it rewrites `calculate_ndcg`/`idcg_at_k` — functions #12193 replaced with `calculate_retrieval_metrics`/`ideal` structure on trunk — and is currently `CONFLICTING`. Its diff therefore cannot apply as written, and it does not touch the three metrics #12193 added, which share defect 2.

claudespice:
## Summary

`testoperator run search` reports `NDCG@10`, `Recall@10`, `MRR@10` and `Precision@10`. All four overstated retrieval quality, for two independent reasons.

**The ideal DCG was built from the retrieved documents, not the judged ones.** `ndcg_at_k` sorted its own input — the grades of the documents search returned — and called that the ideal ranking. Any ranking of a retrieved set therefore looked perfect, however many relevant documents it left behind: a query with two equally-relevant documents where search returned one of them at rank 1 scored `NDCG@10 = 1.0`. The ideal now ranks every document judged for the query, as MTEB/`pytrec_eval` do, and the same query scores `0.613`. `recall_at_k` in the same file already took the whole `relevance` map, so the data was at hand.

**A judged query with no search results was dropped from the average.** `average()` divided by the number of queries *that returned something*, so an unanswered query raised the mean instead of lowering it — for all four metrics, not just NDCG. An unanswered query now ranks nothing, which scores 0.0 through the same code path as a ranking that retrieved nothing relevant, and stays in the average. With no judged query answered at all the vectors were empty and `average` evaluated `0.0 / 0.0`, publishing **`NaN`** into the OTLP gauges and the Arrow `Float64` metric columns; empty `qrels` is now a typed error.

Benchmark and telemetry only — no effect on user query results. It did mean the recorded numbers were not comparable to MTEB's published figures, and that a change which stops returning results for some queries could leave the score flat or raise it.

Fixes #12277

## Changes

- `ideal_dcg_at_k(relevance, k)` replaces `idcg_at_k(relevance_scores, k)`: the ideal is computed from the query's judgments. `ndcg_at_k` takes both the ranking and the judgments.
- An unanswered judged query yields an empty ranking instead of `continue`, so it scores 0.0 for every metric and remains in each average.
- `calculate_retrieval_metrics` returns `anyhow::Result` and rejects empty `qrels`. Its one caller, `calculate_search_score_metrics`, already returned `Result` and already used `?`, so no caller changes were needed.
- `Precision@k` still divides by `k` — the standard TREC convention, deliberately unchanged.

## Overlap with #12226

@Jeadie's open #12226 identifies the same two defects and I have kept its API shape (`anyhow::Result` + `ensure!` on empty `qrels`) so the two converge rather than diverge. That PR is based on a tree predating #12193, so it rewrites `calculate_ndcg`/`idcg_at_k` — functions #12193 replaced on trunk — and currently reads `CONFLICTING`; it also predates the three metrics #12193 added, which share the averaging defect. This applies the same fixes to trunk's actual structure and extends the averaging fix to all four metrics. Happy to close this in favour of a rebased #12226 if you would rather carry it there.

## Test plan

- `make lint` — green.
- `make test` / `make nextest` — green.
- Three new regression tests in `crates/test-framework/src/spicetest/search/evaluate.rs`:
  - `ndcg_penalizes_relevant_documents_the_search_missed` — one of two relevant docs retrieved ⇒ `NDCG@10 = 1/(1 + 1/log2(3)) = 0.613`, `Recall@10 = 0.5`.
  - `an_unanswered_query_scores_zero_and_stays_in_the_average` — two judged queries, one answered perfectly ⇒ `0.5` for NDCG/Recall/MRR and `0.05` for Precision@10.
  - `empty_qrels_is_an_error_rather_than_nan`.
- Verified by neutering: restoring the retrieved-docs ideal DCG and the `continue` fails **exactly** those two new tests, reporting `got 1` in both cases — the symptom from the issue — while the five pre-existing tests stay green.
- The four pre-existing tests keep their expected values: a perfectly-ranked query still scores 1.0 across all four metrics, since its judged set and its retrieved set coincide.

## Checklist

- [x] Scoped lint green (`make lint-rust PACKAGES=test-framework`, both clippy passes)
- [x] No behaviour change for user query paths

### DIFF
--- a/crates/test-framework/src/spicetest/search/evaluate.rs
+++ b/crates/test-framework/src/spicetest/search/evaluate.rs
@@ -39,39 +39,52 @@ pub struct RetrievalMetrics {
 /// * `results` - Search results mapping `query_id` -> (`doc_id` -> `similarity_score`)
 /// * `k` - Rank cutoff shared by all four metrics
 ///
+/// Every judged query in `qrels` contributes to each average. A query the search returned
+/// nothing for scores 0.0 rather than being dropped — dropping it would raise the mean
+/// instead of lowering it.
+///
+/// # Errors
+/// Returns an error when `qrels` is empty: an average over zero queries is undefined.
+///
 /// # Reference
 /// MTEB `RetrievalEvaluator`: <https://github.com/embeddings-benchmark/mteb/blob/03347ebfe4809056e0fd2894fcae69dcdd2ed964/mteb/evaluation/evaluators/RetrievalEvaluator.py#L500>
-#[must_use]
 pub(crate) fn calculate_retrieval_metrics<S: ::std::hash::BuildHasher>(
     qrels: &HashMap<String, HashMap<String, i32, S>, S>,
     results: &HashMap<String, HashMap<String, f64, S>, S>,
     k: usize,
-) -> RetrievalMetrics {
-    let mut ndcg_values = Vec::new();
-    let mut recall_values = Vec::new();
-    let mut mrr_values = Vec::new();
-    let mut precision_values = Vec::new();
+) -> anyhow::Result<RetrievalMetrics> {
+    anyhow::ensure!(
+        !qrels.is_empty(),
+        "Cannot calculate retrieval metrics: no query relevance judgments were provided"
+    );
+
+    let mut ndcg_values = Vec::with_capacity(qrels.len());
+    let mut recall_values = Vec::with_capacity(qrels.len());
+    let mut mrr_values = Vec::with_capacity(qrels.len());
+    let mut precision_values = Vec::with_capacity(qrels.len());
 
     for (query_id, relevance) in qrels {
-        let Some(ranked_results) = results.get(query_id) else {
+        // An unanswered query ranks nothing, which scores 0.0 for all four metrics
+        // through the same code path as a ranking that retrieved nothing relevant.
+        let ranked_relevance = if let Some(ranked_results) = results.get(query_id) {
+            score_sorted_relevance(ranked_results, relevance)
+        } else {
             println!("No search results found for test query {query_id}");
-            continue;
+            Vec::new()
         };
 
-        let ranked_relevance = score_sorted_relevance(ranked_results, relevance);
-
-        ndcg_values.push(ndcg_at_k(&ranked_relevance, k));
+        ndcg_values.push(ndcg_at_k(&ranked_relevance, relevance, k));
         recall_values.push(recall_at_k(&ranked_relevance, relevance, k));
         mrr_values.push(mrr_at_k(&ranked_relevance, k));
         precision_values.push(precision_at_k(&ranked_relevance, k));
     }
 
-    RetrievalMetrics {
+    Ok(RetrievalMetrics {
         ndcg: average(&ndcg_values),
         recall: average(&recall_values),
         mrr: average(&mrr_values),
         precision: average(&precision_values),
-    }
+    })
 }
 
 /// Orders a query's search results by score (descending) and returns the relevance grade of each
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

Issue #12274, opened 2026-07-31
A file-backed full-text index keeps the tantivy schema it was first created with, so later search configuration changes are silently ignored


claudespice:
## Describe the bug

A full-text index configured with `index_store: file` is persisted under `.spice/data/fts/<catalog>/<schema>/<table>/`. On every later start, `FullTextDatabaseIndex::try_new` builds the tantivy schema from the *current* spicepod configuration and then throws it away:

```rust
// crates/search/src/generation/text_search/index.rs:270
match tantivy::Index::create_in_dir(path.clone(), tantivy_schema) {
    Ok(idx) => idx,
    Err(TantivyError::IndexAlreadyExists) => {
        tantivy::index::Index::open_in_dir(path).context(TextSearchIndexingSnafu)?
    }
    Err(e) => return Err(e).context(TextSearchIndexingSnafu),
}
```

`open_in_dir` loads the schema stored in the directory's `meta.json`; the freshly built `tantivy_schema` is dropped. Everything downstream then uses the on-disk schema — `update_index` reads `self.reader.searcher().schema()` (index.rs:397) for both the delete terms and document parsing, and the query parser is built from `self.reader.schema()` (`generation/text_search/mod.rs:285`).

Nothing compares the two schemas, and nothing warns. The mismatch is silent in both directions:

- **A newly configured search column is never indexed.** Documents are built with `TantivyDocument::from_json_object`, and tantivy skips any JSON field that is absent from the schema (`if let Ok(field) = schema.get_field(&field_name)`, tantivy-0.26.1 `src/schema/document/default_document.rs:211`) — no error, no warning. A search over the new column then matches nothing.
- **A removed column, a changed primary key, or changed stored fields are equally ignored**, so the index keeps serving the shape it was first created with.
- **Field options never take effect either.** #12220 changed the search-field tokenizer from tantivy's `default` to `en_stem`. For an index directory created before that change, the stemming it enables is silently not applied — the shipped fix appears to do nothing after upgrade.

`index_store: memory` is unaffected: it calls `create_in_ram` on every start, so it always matches the current configuration.

## To Reproduce

1. Configure a dataset with a file-backed full-text index on one column and let it refresh.
2. Stop the runtime, add a second column to the `search:` configuration, and start again.
3. Search the new column. It returns no matches, and the logs contain nothing about the ignored configuration.

[...]

PR #12286, opened 2026-08-01
fix(ci): link brew formulas that are installed but unlinked in setup-cc

### LINKED ISSUE #12285: CI: `Set up cc` passes with `cmake` unlinked on macOS runners, failing builds much later with "is `cmake` not installed?"
## Summary

`Set up cc` reports success on macOS runners while leaving `cmake` off `PATH`, so builds fail much later with a misleading `is `cmake` not installed?` from an unrelated crate's build script. This has been silently failing sign-offs and merge-queue batches.

## Root cause

`.github/actions/setup-cc/action.yml` installs the macOS toolchain with:

```bash
brew install cmake pkg-config openssl unixodbc llvm libnfs
```

When a formula is already present in the Cellar but **unlinked**, `brew install` prints a *warning* and **exits 0**. No symlink is created in `/opt/homebrew/bin`, so the binary never reaches `PATH`. Observed verbatim in a merge-queue `Build and Test` job:

```
##[warning]cmake 4.4.2 is already installed, it's just not linked.
To link this version, run:
  brew link cmake
```

The step's own conclusion is `success` (961 ms). Note the contrast with the other five formulas in the same output, which report "already installed **and up-to-date**" — those are linked and fine. Only `cmake` was unlinked.

Roughly an hour later in the same job, `snmalloc-sys`'s build script runs and dies:

```
failed to execute command: No such file or directory (os error 2)
is `cmake` not installed?
error: failed to run custom build command for `snmalloc-sys v0.3.8`
```

## Why it is hard to diagnose

- The failing step is `Build CLI and run nextest` / `Run Rust lint`, not `Set up cc`, so the green setup step is not a suspect.
- There are **zero** `error[EXXXX]` codes, so it reads like an infrastructure blip rather than a defect.
- It only bites on a **cache miss** for `snmalloc-sys`. The same runner can produce green jobs for hours while the toolchain is already broken, then start failing when a build cache goes cold — which makes it look intermittent and host-specific when it is neither.

## Observed impact

Both affected runners had produced *successful* jobs earlier the same evening, which is what made this look like a transient outage rather than a persistent misconfiguration:

| Job | Runner | Result |
|---|---|---|
| Merge-queue `Build and Test` (batch for a queued PR) | `github-runner-03` | failed after 8m26s, left the entry `UNMERGEABLE` and stalled the queue |
| `Remote Sign-off` | `github-runner-02` | failed at 17m |
| `Remote Sign-off` (retry of the same branch) | `github-runner-02` | failed again at 55m |

A retry is not a remedy: the formula stays unlinked, so every rerun that misses the cache fails identically. The 55-minute failure is also a caution against reading elapsed time as evidence of a real compile error.

## Proposed fix

After `brew install`, explicitly link the tools that build scripts invoke as bare commands, then assert they resolve so a stale runner fails immediately with a named cause instead of an hour later in an unrelated crate. `openssl` and `llvm` are keg-only and must stay unlinked.

## Acceptance criteria

- A runner with an unlinked `cmake` either self-heals or fails `Set up cc` with a message naming the tool and the remedy.
- No job reaches a build script with `cmake` missing from `PATH`.

claudespice:
Fixes #12285

## The problem

`Set up cc` reported **success** on macOS runners while leaving `cmake` off `PATH`. `brew install` prints only a warning — and still exits 0 — when a formula is already in the Cellar but **unlinked**, so no symlink is created and the binary is unreachable:

```
##[warning]cmake 4.4.2 is already installed, it's just not linked.
To link this version, run:
  brew link cmake
```

Note the contrast with the five other formulas in the same output, which report "already installed **and up-to-date**" — those were linked and fine. Only `cmake` was unlinked, and the step still concluded `success` in 961 ms.

An hour later, in a different step and a crate nobody had touched, the build died:

```
failed to execute command: No such file or directory (os error 2)
is `cmake` not installed?
error: failed to run custom build command for `snmalloc-sys v0.3.8`
```

## Why it was hard to attribute

- The red step is `Build CLI and run nextest` / `Run Rust lint`, so a green `Set up cc` is not a suspect.
- **Zero** `error[EXXXX]` codes, so it reads as an infrastructure blip rather than a defect.
- It only bites on a **build-cache miss** for `snmalloc-sys`. Both runners involved had produced *successful* jobs earlier the same evening, so a persistent misconfiguration looked transient and host-specific when it was neither. Reruns cannot clear it — the formula stays unlinked, so every cache-missing retry fails identically.

Measured impact before this fix: a merge-queue `Build and Test` batch failed after 8m26s and left its queue entry `UNMERGEABLE`, stalling the queue; a `Remote Sign-off` failed twice on the same branch, at 17m and again at 55m. The 55-minute failure is also a caution against reading long elapsed time as evidence of a genuine compile error.

## The fix

Link the tools that build scripts invoke by bare name, then assert they resolve — so a stale runner fails **in setup**, with the tool named and the remedy printed, instead of an hour later somewhere unrelated.

Linking is best-effort (`|| true`) and the `command -v` check is the authority, so a brew alias quirk can never fail CI on its own — the guard can only convert a silent late failure into an immediate, clearly-attributed one. `openssl` and `llvm` are keg-only and are deliberately left unlinked; the build locates them via `brew --prefix`.

## Verification

- YAML parses; the extracted step is `bash -n` clean.
- Ran the step body under CI's exact shell flags (`bash -e -o pipefail`) against a stubbed `brew`:
  - tools absent → exits **1** after printing `::error::Not on PATH after 'brew install': cmake pkg-config. Run 'brew link --overwrite cmake pkg-config' on this runner.`
  - tools present → exits **0**, silent.
- Confirmed against Homebrew metadata that `pkg-config` is an alias of `pkgconf` and that both `cmake` and `pkgconf` are non-keg-only, so `brew link` is valid for exactly the two formulas being linked.

### DIFF
--- a/.github/actions/setup-cc/action.yml
+++ b/.github/actions/setup-cc/action.yml
@@ -21,6 +21,24 @@ runs:
       shell: bash
       run: |
         brew install cmake pkg-config openssl unixodbc llvm libnfs
+        # `brew install` only warns — and still exits 0 — when a formula is already in
+        # the Cellar but unlinked, so its binaries never reach PATH. The failure then
+        # surfaces an hour later inside an unrelated crate, as a build script dying with
+        # "is `cmake` not installed?" and no error[EXXXX] to point at. Link the tools
+        # that build scripts invoke by bare name, then assert they resolve, so a stale
+        # runner fails here with a named cause. openssl and llvm are keg-only and are
+        # deliberately left unlinked; the build locates them via `brew --prefix`.
+        for tool in cmake pkg-config; do
+          brew link --overwrite "$tool" || true
+        done
+        missing=()
+        for tool in cmake pkg-config; do
+          command -v "$tool" >/dev/null 2>&1 || missing+=("$tool")
+        done
+        if [ ${#missing[@]} -ne 0 ]; then
+          echo "::error::Not on PATH after 'brew install': ${missing[*]}. Run 'brew link --overwrite ${missing[*]}' on this runner."
+          exit 1
+        fi
 
     - name: Install cc (Windows)
       if: runner.os == 'Windows'


claudespice:
@copilot review

claudespice:
Live blast radius as of 01:2xZ — this is currently blocking the merge queue, so it may be worth prioritising:

| Blocked | Runner | Failure |
|---|---|---|
| #12166 (queue pos 2) | `github-runner-03` | batch `Build and Test` failed 8m26s → entry `UNMERGEABLE` |
| #12262 (queue pos 5) | `github-runner-02` | batch `Build and Test` failed → entry `UNMERGEABLE` |
| #12169 | — | evicted from the queue earlier by the same family of failure |
| #12281 | `github-runner-02` | `Remote Sign-off` failed twice (17m, then 55m) |

All five failures have **zero** `error[EXXXX]` codes and all end in `snmalloc-sys` → `cmake-0.1.58` → `failed to execute command: No such file or directory (os error 2)`.

Two notes for anyone triaging in the meantime:

- **Requeuing/rerunning does not help.** The formula stays unlinked, so any batch that misses the `snmalloc-sys` build cache fails again. Both runners produced *green* jobs earlier the same evening, which is exactly what made this read as a transient outage.
- **Immediate unblock without waiting for this PR:** run `brew link --overwrite cmake` on `github-runner-02` and `github-runner-03`. This PR makes that self-healing and turns any future recurrence into an immediate `Set up cc` failure that names the tool.

claudespice:
Correction to my table above, and it narrows the remedy considerably: **every one of these failures is on a single machine, `spiceai-macos-runner-06`.** I read `github-runner-02` / `-03` off the `_work` paths in the logs — those are the runner *instance* directories, and runner-06 hosts several of them. The `Machine name:` line is the one to trust.

Confirmed by pulling `Machine name` out of all five failing jobs plus two green ones from the same evening:

| Job | Machine | Result |
|---|---|---|
| #12262 batch `Rust Lint` (91305974701) | `spiceai-macos-runner-06` | ❌ `is cmake not installed?` |
| #12262 batch `Build and Test` (91305974720) | `spiceai-macos-runner-06` | ❌ same |
| #12169 batch `Build and Test` (91309663486) | `spiceai-macos-runner-06` | ❌ same |
| #12281 sign-off 30673050916 | `spiceai-macos-runner-06` | ❌ same |
| #12281 sign-off 30674799671 | `spiceai-macos-runner-06` | ❌ same |
| #12277 sign-off 30666666454 | `spiceai-macos-runner-02` | ✅ |
| #12274 sign-off 30664320724 | `spiceai-macos-runner-04` | ✅ |

So the "intermittent outage" is really a coin flip on which machine picks the job up, and **`brew link --overwrite cmake` on `spiceai-macos-runner-06` alone clears it right now**, without waiting for this PR.

Two further data points since the last comment:

- **Requeuing is now empirically confirmed useless, not just predicted.** #12169 was re-enqueued at 01:16:48Z, drew runner-06 again, and its batch `Build and Test` died the same way at 01:47:04Z — 5 minutes in, zero `error[EXXXX]` codes. The merge queue is currently 6 entries deep with 3 `UNMERGEABLE`, and nothing has merged to trunk since 20:04Z.
- **Rollout caveat for this fix.** `signoff.yml` checks out `ref: <branch>` and then runs `./.github/actions/setup-cc` from *that* checkout, so once this lands on trunk, remote sign-offs on already-open branches keep using the old, unguarded action until each branch merges trunk. Merge-queue batches pick the fix up immediately (the `merge_group` ref contains trunk), so the queue heals on merge; branch sign-offs heal one merge-from-trunk at a time.

lukekim (approved):

Issue #12284, opened 2026-08-01
http_requests counts a mid-stream query failure as a 200, so the metric overstates the success rate on streaming SQL responses


claudespice:
## Problem

`track_metrics` records `http_requests{status}` from the response head:

```rust
// crates/runtime/src/http/routes.rs
let response = ...;                       // head is already resolved here
let latency_ms = start.elapsed().as_secs_f64() * 1000.0;
let status = http_status_label(response.status().as_u16());
```

For the streaming SQL response paths the head is sent as `200 OK` before the first batch is produced, so a query that fails **after** streaming has begun is counted as a **success**. `http_requests{path="/v1/sql",status="200"}` therefore overstates the success rate, and the amount by which it does so is exactly the failures an operator most wants to see: the ones that only appear once real data starts moving (memory-pool exhaustion partway through a join, a connector dropping mid-scan, a `runtime.query.timeout` expiry on a long scan).

The same skew applies to `http_requests_duration_ms`, which records the time to the response *head*, not to the end of the body, for those paths.

The pre-stream error paths are unaffected — `sql_error_response` returns before the head is sent, so those are counted with their real status.

## Why it matters

This is the residue of the "no HTTP response-code metric" concern in #12282. That framing turned out not to hold — `track_metrics` is a `route_layer` on the merged router and does cover `/v1/sql` with a `status` label, so `sum(rate(http_requests{path="/v1/sql",status="400"}[5m]))` is alertable today. But the metric quietly under-reports precisely the mid-flight failures, so a dashboard built on it can look healthy while a meaningful fraction of queries are failing in-stream.

## Possible approach

The request already carries a mechanism that outlives the head: `CancelGuardBody` wraps the response body in the same middleware and observes end-of-stream in order to disarm the cancellation drop guard. Recording a terminal outcome (and the true end-to-end duration) at that same point, rather than at the head, would let a mid-stream failure be counted as a failure — either as a separate outcome label or as a distinct counter, so the existing `status` series keeps its current meaning.

PR #12289, opened 2026-08-01
fix(runtime): report a memory-pool refusal as ResourcesExhausted and answer it with 503 (fixes #12282)

### LINKED ISSUE #12282: Query-path observability: ResourcesExhausted is reported as InternalError, and there is no HTTP response-code metric
## Problem

A runtime that is failing every query with `ResourcesExhausted` is indistinguishable, from the outside, from a healthy one.

Measured while investigating #12270: with `runtime.query.memory_limit: 4GiB` and an over-sized `target_partitions`, a 2.2M x 2.2M join failed **25% of queries at concurrency 4** and **99% at concurrency 64**, every one of them:

```
HTTP 400
Resources exhausted: Additional allocation failed for HashJoinInput[135] with top memory consumers as:
  GroupedHashAggregateStream[277] (count(1), count(DISTINCT mic.Sid))#307904 consumed 1663.8 KB, ...
```

Over the same runs, liveness and readiness recorded **zero probe failures** (0.5-0.9ms medians). Nothing in metrics, logs, or health signalled the outage.

Three separate gaps combine:

**1. `ResourcesExhausted` is recorded as `InternalError`.**

`crates/runtime/src/datafusion/query/error_code.rs` has no arm for it, so it falls through:

```rust
match error {
    DataFusionError::SQL(..) => ErrorCode::SyntaxError,
    DataFusionError::Plan(..) | DataFusionError::SchemaError(..) => ErrorCode::QueryPlanningError,
    DataFusionError::ObjectStore(..) | DataFusionError::External(..)
        | DataFusionError::Execution(..) => ErrorCode::QueryExecutionError,
    DataFusionError::Context(_, err) => ErrorCode::from(err.as_ref()),
    _ => ErrorCode::InternalError,   // <- ResourcesExhausted lands here
}
```

`query_failures{err_code="InternalError"}` therefore conflates "this deployment needs more memory or fewer partitions" — an operator-actionable capacity signal — with "spiced has a bug". An operator cannot alert on the former without alerting on the latter.

The Flight paths already get this right, which is what makes the HTTP path look like an oversight rather than a decision:

```rust
// crates/runtime/src/flight/mod.rs:748
DataFusionError::ResourcesExhausted(source) => Status::resource_exhausted(source),
// crates/datafusion-flightsql/src/util.rs:53
DataFusionError::ResourcesExhausted(msg) => Status::resource_exhausted(msg),
```

So the same failure is `RESOURCE_EXHAUSTED` over Flight and `InternalError` in metrics over HTTP.

**2. There is no HTTP response-code metric.** Nothing counts 4xx/5xx on the query endpoints, so "what fraction of queries are we rejecting" is unanswerable from `/metrics` — even in aggregate.

**3. The error text is logged at DEBUG.** `sql_error_response` does `tracing::debug!("Error executing query: {message}")`, so at the default INFO a runtime returning 400 to every client logs nothing. The message *is* returned in the response body, so clients see it; the operator does not.

## Proposed

- Add `ErrorCode::ResourcesExhausted` (or equivalent) and match `DataFusionError::ResourcesExhausted` in `error_code.rs`, so `query_failures{err_code}` separates capacity exhaustion from internal errors.
- Emit a response-status metric for the HTTP query endpoints, so 4xx rate is alertable.
- Reconsider the HTTP status: `sql_error_response` maps every non-cancellation, non-timeout error to `400 BAD_REQUEST`. Resource exhaustion is a server-capacity condition, not a malformed request — `503` or `507` would let load balancers and clients retry correctly. (Behaviour change; flagging rather than assuming.)
- Consider raising the query-error log from `debug!` to `warn!`, or sampling it, so a sustained failure rate is visible at the default level.

## Why it matters

Kubernetes cannot see this. Liveness and readiness stay green because `/health` is a static closure served by a separate tokio runtime from the query engine — that isolation is deliberate and correct, but it means the orchestrator will never restart or drain a pod that has stopped serving queries. Metrics are the only remaining signal, and today they label the condition `InternalError` with no rate to alert on.

Found while validating #12270 (CPU entitlement detection); the over-sizing bug there is what surfaced it, but this gap is independent of that fix.

### LINKED ISSUE #12283: Query path: resource exhaustion is returned as 400 Bad Request, which tells clients and load balancers the wrong thing
## Problem

`/v1/sql` answers a query that the engine refused for want of memory with **`400 Bad Request`**:

```rust
// crates/runtime/src/http/v1/mod.rs
fn status_for_sql_error(message: &str) -> StatusCode {
    if message.contains("read-only SQL context") {
        StatusCode::FORBIDDEN
    } else {
        StatusCode::BAD_REQUEST
    }
}
```

Resource exhaustion is a server-capacity condition, not a malformed request. `4xx` tells every intermediary that the client is at fault and the request should not be retried, so a load balancer will keep routing to a saturated pod instead of shedding or draining it, and a client library that retries on `5xx` but not `4xx` will surface a hard failure for a condition that would very likely succeed a moment later or on another replica.

The same failure over Flight is already reported as `RESOURCE_EXHAUSTED` (`crates/runtime/src/flight/mod.rs`, `crates/datafusion-flightsql/src/util.rs`), which gRPC clients treat as retriable — so the two protocols disagree on whether the client did something wrong.

This was raised as one of the proposals in #12282 and explicitly flagged there as a behaviour change rather than assumed. #12282's other arms (the `ErrorCode` classification and the log level) are fixed separately; this one is split out because it changes what existing HTTP clients see and wants an owner's decision rather than a drive-by change.

## Options

- **`503 Service Unavailable`** — the standard "try again, possibly elsewhere" signal; understood by load balancers and most retry middleware, and pairs naturally with `Retry-After`.
- **`507 Insufficient Storage`** — literally about storage, rarely special-cased by intermediaries; more precise on paper, less useful in practice.
- **Leave `400`** — no client breakage, but the disagreement with Flight and the retry semantics above stay.

`runtime.query.timeout` already returns `504` and a client cancel already returns `499`, so the endpoint does not treat every query error as `4xx`; there is precedent for a status that reflects the runtime's own state.

## Notes

The classification needed to act on this already exists after the #12282 fix: `SqlErrorKind::ResourcesExhausted` is carried through to `sql_error_response`, and it currently shares `SqlErrorKind::General`'s status deliberately. Changing the status is a one-line change to that match arm once the contract question is settled.

claudespice:
## Summary

A runtime refusing every query for want of memory was indistinguishable, from `/metrics`, from a healthy one — and it told every HTTP client the query was the client's fault.

`ErrorCode::from(&DataFusionError)` had no arm for `DataFusionError::ResourcesExhausted`, so it fell through the catch-all to `ErrorCode::InternalError`. Every `query_failures` sample raised by an exhausted `runtime.query.memory_limit` was therefore labelled `err_code="InternalError"` — the same label a genuine runtime bug gets. An operator could not alert on "this deployment needs more memory or fewer partitions" without also alerting on "spiced is broken", and `/health` stays green throughout because it is served by a separate tokio runtime.

The same refusal was also answered with `400 Bad Request`. Resource exhaustion is a server-capacity condition: `4xx` tells every intermediary the client is at fault and the request should not be retried, so a load balancer keeps routing to a saturated pod instead of shedding it, and a client library that retries on `5xx` but not `4xx` surfaces a hard failure for a query that would very likely succeed a moment later or on another replica.

The Flight paths already classified this correctly (`flight/mod.rs` and `datafusion-flightsql/src/util.rs` both map it to the retriable `RESOURCE_EXHAUSTED`), so the same failure was reported two different ways depending on the protocol the client used.

Fixes #12282
Fixes #12283

## Changes

- **`ErrorCode::ResourcesExhausted`** — a new variant, matched from `DataFusionError::ResourcesExhausted`. The existing `Context(_, err)` recursion is unchanged, so a pool error wrapped by execution is still found. The `Display` and `i8` impls match exhaustively with no catch-all, so the compiler required both to be updated; the existing codes keep their values and the new one takes `-40`.
- **`/v1/sql` answers a memory-pool refusal with `503 Service Unavailable`** instead of `400`. This is the client-visible arm that #12283 split out for an owner's decision, and @sgrebnov settled it in review on this PR. `503` is the standard "try again, possibly elsewhere" signal, it is what intermediaries and retry middleware already understand, and it brings HTTP into agreement with Flight. Only the capacity condition moves — a malformed query still gets `400` and a read-only rejection still gets `403`, both asserted.
- **The query-error log is raised from `debug!` to `warn!` for this one condition.** `sql_error_response` logged every failure at `debug!`, so a runtime refusing every query said nothing at the default `INFO`. Raising *all* query errors would make a client sending malformed SQL noisy in the server's log, which is why this is scoped: a new `SqlErrorKind::ResourcesExhausted` carries the classification, resource exhaustion logs at `warn!`, and everything else keeps `debug!`.

  `ErrorCode` gains `PartialEq`/`Eq` so the HTTP layer classifies through the same `From` impl rather than re-deriving the test, keeping one source of truth.
- **The logged copy of the message is folded onto one line.** The pool formats its refusal as `"Additional allocation failed for {consumer} with top memory consumers (across reservations) as:\n{top_consumers}\nError: {error_msg}"` (`datafusion-execution-54.0.0/src/memory_pool/pool.rs:611`), so the `warn!` above would emit a multi-line record and break line-oriented log collectors — and the repo requires single-line log messages. A `single_line` helper replaces every control character with a space for the log only — replaced, not dropped, so offsets into the logged line still match the message — and the HTTP response body keeps the original formatting. A message with nothing to collapse is borrowed unchanged, so the common path does not allocate.

## Compatibility

`/v1/sql` clients that special-case `400` for a memory refusal will now see `503`. That is the point of the change — but it is a visible contract change, so it is called out rather than buried. The response body is unchanged, `runtime.query.timeout` already returned `504` and a client cancel already returned `499`, so the endpoint was never uniformly `4xx` on failure.

## Deliberately not changed

- **No `Retry-After` header.** It would need a number the runtime cannot honestly produce — the pool has no admission queue and no drain estimate — and `503` without `Retry-After` is already the retriable signal clients act on.
- **The sibling HTTP handlers.** `sql_error_response` is the only funnel that mapped a query error to a status; `search.rs` already answers an unclassified failure with `500`, and the other `400`s in `http/v1/` are genuine client errors (invalid JSON, missing `model`, no data sources).
- **"There is no HTTP response-code metric" does not reproduce.** `track_metrics` (`http/routes.rs`) is a `route_layer` on the merged router, so it covers `/v1/sql` and every other query endpoint, and it already records `http_requests`, `http_requests_total` and `http_requests_duration_ms` with a `status` label built by `http_status_label(response.status().as_u16())`. Checking that did surface a real but narrower gap — a query that fails *after* the response headers are sent (the streaming path) is counted as its already-sent `200` — filed as **#12284**.
- **`JobErrorCode` (async query jobs) is left alone.** It is a separate, coarser, serialized public enum, and `query_error_to_code` already maps an execution failure to `ExecutionFailed` rather than `Internal` — it does not have the "capacity looks like a bug" defect this PR fixes, and adding a variant there would change a published API surface.

## Test plan

- `make lint`
- `make test`
- New unit tests in `crates/runtime/src/datafusion/query/error_code.rs`: a bare `ResourcesExhausted` classifies as `ErrorCode::ResourcesExhausted` and renders as `"ResourcesExhausted"`; the same error wrapped in `Context` is still found through the recursion; the `i8` codes are asserted individually so a later variant cannot silently renumber the existing ones; `Plan`, `Execution` and `Internal` keep the codes they had.
- `sql_error_response_classifies_resource_exhaustion` (`crates/runtime/src/http/v1/mod.rs`): `of_datafusion_error` returns the new kind for both the bare and the `Context`-wrapped pool error and both reach `503`; a plain `Execution` error stays on the quiet `General` path **and on `400`**; a read-only rejection still reaches `403`. The last two pin that only the capacity condition moved.
- `resource_exhaustion_logs_on_one_line_but_responds_in_full`: feeds `single_line` a pool-shaped multi-line message and asserts no control character survives, the consumer breakdown is kept rather than dropped, and the character count is unchanged; a message with nothing to collapse comes back `Cow::Borrowed` and byte-identical. It then reads the response body back off `sql_error_response` and asserts it is the engine's message verbatim, newlines included, alongside the `503`.
- Reverting the `error_code.rs` match arm alone fails `resources_exhausted_is_not_an_internal_error`, `resources_exhausted_is_found_through_context` and `sql_error_response_classifies_resource_exhaustion`, leaving the pre-existing tests green. Reverting the status arm alone fails `sql_error_response_classifies_resource_exhaustion` and `resource_exhaustion_logs_on_one_line_but_responds_in_full`.

## Checklist

- [x] Regression tests that fail on the old code
- [x] Existing classifications unchanged
- [x] The one client-visible change is called out under Compatibility, and pinned by tests that show nothing else moved
- [x] `cargo fmt` clean


claudespice:
@copilot review

crates/runtime/src/http/v1/mod.rs
```diff
@@ -407,12 +414,27 @@ impl SqlErrorKind {
...
+/// an outage its operator cannot see any other way: `/health` is served by a
+/// separate tokio runtime and stays green throughout, so nothing else raises
+/// the condition at the default verbosity.
 fn sql_error_response(message: String, kind: SqlErrorKind) -> Response {
-    tracing::debug!("Error executing query: {message}");
+    if matches!(kind, SqlErrorKind::ResourcesExhausted) {
+        tracing::warn!("Query refused, out of memory: {message}");
+    } else {
+        tracing::debug!("Error executing query: {message}");
+    }
```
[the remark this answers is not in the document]
claudespice: Good catch, and it is not hypothetical — DataFusion 54's `TrackConsumersPool` formats the refusal as `"Additional allocation failed for {consumer} with top memory consumers (across reservations) as:\n{top_consumers}\nError: {error_msg}"` (`datafusion-execution-54.0.0/src/memory_pool/pool.rs:611`), so the `warn!` this PR adds is exactly the path that emits a multi-line record. It also violates the repo's own rule (`CLAUDE.md`: keep every log/error message on a single line).

Fixed in 1a2ae24f8: a `single_line` helper folds the message for the log record only — the HTTP response body still carries the original formatting, as you suggested. It borrows unchanged when there is no newline, so the common path does not allocate. Added `resource_exhaustion_message_logs_on_one_line`, which feeds it a pool-shaped multi-line message and asserts the fold contains no newline, joins rather than drops the consumer lines, and keeps the tail.

crates/runtime/src/http/v1/mod.rs
```diff
@@ -778,6 +824,97 @@ mod tests {
...
+        );
+        assert!(matches!(
+            SqlErrorKind::of_datafusion_error(&exhausted),
+            SqlErrorKind::ResourcesExhausted
+        ));
+        let response = sql_error_response(
+            exhausted.to_string(),
+            SqlErrorKind::of_datafusion_error(&exhausted),
+        );
+        assert_eq!(response.status(), StatusCode::BAD_REQUEST);
```
sgrebnov: @claudespice - claudespice HTTP status code for ResourcesExhausted must be `503 Service Unavailable` not the `BAD_REQUEST`. Improve PR

claudespice: Done in d97611985 — `SqlErrorKind::ResourcesExhausted` now maps to `StatusCode::SERVICE_UNAVAILABLE` on its own match arm.

This was the open contract question #12283 was split out to decide, so your call settles it and the PR now closes that issue too. The commit carries the reasoning in the code: 4xx tells intermediaries the client is at fault and the request should not be retried, so a load balancer keeps routing to a saturated pod, and a client that retries on 5xx surfaces a hard failure for a query that would likely succeed on another replica. It also brings HTTP into agreement with Flight, which already answers the same refusal with the retriable `RESOURCE_EXHAUSTED`.

Scoped so only the capacity condition moved — `SqlErrorKind::General` still routes through `status_for_sql_error`, and `sql_error_response_classifies_resource_exhaustion` now pins all four arms: a bare and a `Context`-wrapped pool refusal both reach 503, a plain `Execution` error stays on 400, and a read-only rejection stays on 403. `cargo test -p runtime --lib http::v1` is green (72 passed).

I did not add `Retry-After` — the pool has no admission queue or drain estimate, so any value would be invented; happy to add one if you would rather it carried a fixed hint. The status change is called out under a new **Compatibility** heading in the PR body rather than left implicit.

claudespice:
One gap left after `d976119`: the `/v1/sql` OpenAPI annotation still declares only `200`, `400` and `500`, so the published spec (`.schema/openapi.json`) does not mention the new `503` a client now has to handle. Worth folding into the next push on this branch rather than a separate one:

```rust
// crates/runtime/src/http/v1/query.rs, after the 500 entry
(status = 503, description = "The query was refused for want of memory. Retriable: the same query may succeed once the runtime is under less pressure, or on another replica.", content((
    String,
    example = "Resources exhausted: Failed to allocate additional 256.0 MB for HashJoinInput[135]"
)))
```

Not pushing it myself — a sign-off is in flight on this head ([30721381465](https://github.com/spiceai/spiceai/actions/runs/30721381465)) and a push would cancel it and restart the clock for a one-hunk annotation.

For the record, the annotation already omits the `499` and `504` this endpoint has returned for a while, so the spec under-documents `/v1/sql` beyond this change; adding all three is a reasonable separate cleanup.

sgrebnov (approved):

sgrebnov (approved):

PR #12291, opened 2026-08-01
fix(runtime): count a mid-stream response failure as a failure in the HTTP metrics (fixes #12284)

### LINKED ISSUE #12284: http_requests counts a mid-stream query failure as a 200, so the metric overstates the success rate on streaming SQL responses

claudespice:
## Summary

`track_metrics` records `http_requests`, `http_requests_total` and `http_requests_duration_ms` from the response **head**:

```rust
let response = /* … */;                   // the head is already resolved here
let latency_ms = start.elapsed().as_secs_f64() * 1000.0;
let status = http_status_label(response.status().as_u16());
```

For the streaming response paths that is too early. `/v1/sql`'s default JSON format pulls one batch eagerly and then hands the rest to `Body::from_stream`, so the head is `200 OK` before the bulk of the query has run. A query that fails **after** streaming began was therefore counted as a success, and its duration was recorded as the time to the head rather than to the last byte.

The skew lands exactly on the failures an operator most wants to see — memory-pool exhaustion partway through a join, a connector dropping mid-scan, a `runtime.query.timeout` expiry on a long scan — so `sum(rate(http_requests{path="/v1/sql",status="200"}[5m]))` could look healthy while a real share of queries were failing in flight.

Pre-stream errors were never affected: `sql_error_response` returns before the head is sent, so those already carried their true status.

Fixes #12284

## Changes

- **`OutcomeTrackedBody`** (`crates/runtime/src/http/response_outcome.rs`) wraps a response body and reports, exactly once, how it terminated:
  - `complete` — the body reached end-of-stream;
  - `error` — the body yielded an error after the head was sent;
  - `incomplete` — the body was dropped before end-of-stream (client disconnect mid-stream).

  It handles the same termination shapes `CancelGuardBody` already has to: a body that signals end-of-stream alongside its final frame (the caller then never polls again), and a body that is empty before it is ever polled.

- **`track_metrics`** wraps every response in it, and on termination records two new instruments:
  - `http_responses{method,path,status,outcome,…}`
  - `http_responses_duration_ms{…}` — the true end-to-end duration, head *and* body.

The existing `http_requests`, `http_requests_total` and `http_requests_duration_ms` series are left exactly as they are, so dashboards and alerts built on them keep their current meaning. A mid-stream failure is now visible as `http_responses{outcome="error"}`, and the head-vs-body duration gap is the difference between the two histograms.

The wrapper sits in the same middleware as `CancelGuardBody`, which is a `route_layer` on the merged router — so this covers every route, not just `/v1/sql`. The SSE paths (`/v1/chat/completions` and friends) get the same treatment for free.

Not included, deliberately: no panel was added to `monitoring/grafana-dashboard.json` or `monitoring/datadog-dashboard.json`. Which of the new series belongs on the default dashboards is a presentation decision, and this PR is the instrument.

## Performance impact

One extra body wrapper per response, alongside the `CancelGuardBody` that is already there — a delegating `poll_frame` with one `Option::take` on the terminal frame. The metric labels are the vector `track_metrics` already builds, moved into the callback rather than rebuilt, plus one `outcome` `KeyValue` pushed at termination.

## Test plan

- `make lint`
- `make nextest`

New unit tests in `crates/runtime/src/http/response_outcome.rs`:

- `reports_error_when_the_body_fails_mid_stream` — the regression test for this issue: a body that yields a data frame and then an error reports `error`, reports nothing before it terminates, and does not report a second outcome when it is dropped afterwards. On the old code there is no report at all, and the request is counted `200`.
- `reports_complete_when_the_body_reaches_end_of_stream` — the ordinary streaming success.
- `reports_incomplete_when_the_body_is_dropped_mid_stream` — client disconnect.
- `reports_complete_when_end_of_stream_arrives_with_the_final_frame` — the body signals end-of-stream with its last frame and the caller stops polling; without this arm that response would be reported `incomplete`.
- `reports_complete_for_a_body_that_is_empty_before_it_is_polled` — empty bodies and HEAD responses, which are never polled at all.
- `outcome_labels_are_stable` — pins the three label strings, since a rename silently breaks every query written against them.

## Checklist

- [x] Regression test added that fails on the old code
- [x] Existing `http_requests*` series unchanged
- [ ] `make lint` / `make nextest` green (via `make signoff`)


claudespice:
@copilot review

lukekim (approved):

lukekim (approved):

PR #12293, opened 2026-08-01
fix(flightsql): mark stamped statistics inexact when a LIMIT is pushed to the remote scan (fixes #12292)

### LINKED ISSUE #12292: Distributed leaf scan reports exact statistics for a row count a pushed-down LIMIT makes impossible
## Summary

A cluster leaf scan (`FlightSqlExec`) is stamped with the statistics an executor reported for the **whole** table slice. `FlightSQLTable::create_physical_plan` already degrades those statistics to `Inexact` when a **filter** is pushed into the remote SQL — that is the fix for #11599. It does **not** do the same when a **`LIMIT`** is pushed into the remote SQL, and `FlightSqlExec::with_fetch` (the limit-pushdown entry point) copies the statistics through verbatim while changing the limit.

So the plan asserts an exact row count that the scan provably cannot produce.

## Reproduction

`crates/data_components/src/flightsql.rs`, no network needed — the existing `scan_marks_stamped_statistics_inexact_when_filter_pushed` test fixture stamps `num_rows: Exact(150_000)`:

```rust
// limit pushed into the remote SQL as `... LIMIT 10`
let plan = table.scan(&session, None, &[], Some(10)).await?;
assert_eq!(plan.partition_statistics(None)?.num_rows, Precision::Exact(150_000)); // passes today
```

The same holds through the pushdown path:

```rust
let limited = exec.with_fetch(Some(10)).unwrap();          // limit becomes Some(10)
limited.partition_statistics(None)?.num_rows;               // still Exact(150_000)
```

`FlightSqlExec::sql()` emits `LIMIT {limit}` (`flightsql.rs:692`, appended at `:756`), so the scan returns at most 10 rows while claiming exactly 150 000. Column bounds are affected the same way: the first 10 rows' min/max are not the slice's min/max, but they stay `Exact`.

## Why it matters

`Precision::Exact` is a correctness contract, not a hint — DataFusion substitutes exact statistics into results. `aggregate_statistics` (the rule named in this function's own comment as the #11599 vector) folds `COUNT(*)`/`MIN`/`MAX` straight out of the child's `partition_statistics`, and `FlightSqlExec` reports `supports_limit_pushdown() == true`, so `LimitPushdown` **removes** the corrective `GlobalLimitExec` above the scan and leaves the false `Exact` as the plan's own row count.

In the current default rule order `aggregate_statistics` runs before `LimitPushdown`, so a plain `SELECT COUNT(*) FROM (SELECT * FROM t LIMIT n)` is still folded off the surviving `GlobalLimitExec` and answers correctly today. The defect is therefore a violated invariant rather than a reproduced wrong answer — but it is the identical shape to #11599, it is load-bearing for anything that reads the scan's statistics after `LimitPushdown` (a re-planned/serialized distributed plan, a new or reordered rule, a custom FlightSQL physical rule), and per the repo's own standard statistics may be reported exact only when provably exact.

## Suggested fix

Two sites, both mirroring the existing filter guard:

1. `FlightSQLTable::create_physical_plan` — degrade when `!filters.is_empty() || limit.is_some()`.
2. `FlightSqlExec::with_fetch` — degrade the carried statistics when the merged limit is `Some(_)`, since this is exactly where `LimitPushdown` introduces a limit into a scan that was built without one.

Both keep the bounds usable for join sizing (`JoinSelection` and the broadcast-join cost test read `Inexact` values happily), so there is no plan-quality regression.

## Scope

`FlightSqlExec` is the only execution plan in the repo that both carries externally supplied statistics and pushes a limit into a remote query — `oracle`/`mssql` execs report unknown statistics, and the Cayenne scan wrapper delegates `partition_statistics` to an inner plan that already accounts for its fetch. So this is a one-instance class, not a sweep.

## Note on overlap

Open PR #11858 rewrites the `create_physical_plan` statistics block (adding filter selectivity scaling) and keeps the `filters.is_empty()` condition, so it does not cover the limit case; it does not touch `with_fetch`.

claudespice:
Fixes #12292

## Summary

A cluster leaf scan (`FlightSqlExec`) is stamped with the statistics an executor reported for its **whole** table slice. `FlightSQLTable::create_physical_plan` already degrades those statistics to `Inexact` when a **filter** is pushed into the remote SQL — that is the fix for #11599 — but it did not do the same for a pushed-down **`LIMIT`**, and `FlightSqlExec::with_fetch` copied the statistics through verbatim while setting the limit.

`FlightSqlExec::sql()` emits `LIMIT {limit}` into the remote query, so the plan was asserting an exact row count and exact column bounds that the scan provably cannot produce: with `num_rows: Exact(150_000)` stamped, `table.scan(.., Some(10))` returned a plan still reporting `Exact(150_000)`.

`Precision::Exact` is a correctness contract — DataFusion substitutes exact statistics into results. `aggregate_statistics` (the rule named in this function's own comment as the #11599 vector) folds `COUNT(*)`/`MIN`/`MAX` out of the child's `partition_statistics`, and because `FlightSqlExec` reports `supports_limit_pushdown() == true`, `LimitPushdown` **removes** the corrective `GlobalLimitExec` above the scan and leaves the false `Exact` as the plan's own row count.

To be precise about severity: in the current default rule order `aggregate_statistics` runs before `LimitPushdown`, so a `SELECT COUNT(*) FROM (SELECT * FROM t LIMIT n)` still folds off the surviving `GlobalLimitExec` and answers correctly today. This is a violated invariant rather than a reproduced wrong answer — but it is the identical shape to #11599, it is load-bearing for anything reading the scan's statistics after `LimitPushdown` (a serialized/re-planned distributed plan, a new or reordered rule, a custom FlightSQL physical rule), and statistics may be reported exact only when provably exact.

## Changes

* `FlightSQLTable::create_physical_plan` — keep the stamped statistics exact only while `filters.is_empty() && limit.is_none()`, i.e. degrade whenever either narrowing is pushed into the remote SQL. The surrounding comment now explains both vectors instead of only filters.
* `FlightSqlExec::with_fetch` — degrade the carried statistics when the merged limit is `Some(_)`. This is where `LimitPushdown` introduces a limit into a scan that was built without one, so it is the second entry point into the same wrong claim. `with_fetch(None)` on an unlimited scan still keeps the statistics exact.
* Tests: two regression tests (`scan_marks_stamped_statistics_inexact_when_limit_pushed`, `with_fetch_marks_stamped_statistics_inexact`) covering both entry points, `num_rows` and column bounds, and the unlimited control case. The stamped-statistics fixture is extracted into `stamped_schema()`/`stamped_statistics()` helpers and reused by the existing #11599 test rather than duplicated.

Scope: `FlightSqlExec` is the only execution plan in the repo that both carries externally supplied statistics and pushes a limit into a remote query — the `oracle` and `mssql` execs report unknown statistics, and the Cayenne scan wrapper delegates `partition_statistics` to an inner plan that already accounts for its own fetch. So this is a one-instance class, not a sweep.

## Performance impact

None expected. `JoinSelection` and the FlightSQL broadcast-join cost test read `Precision` values regardless of exactness, so join sizing and build-side selection still see the same row counts and bounds — only the exactness claim changes, and only for scans that carry a limit. A limited leaf scan is not a join-sizing input in the plans this affects.

## Overlap

Open PR #11858 rewrites this `create_physical_plan` statistics block to add filter-selectivity scaling and keeps the `filters.is_empty()` condition, so it does not cover the limit case; it does not touch `with_fetch`. If #11858 lands first, its version of the block should carry the `&& limit.is_none()` half of the guard; the `with_fetch` half is independent of it. Happy to rebase onto it either way.

## Test plan

- [x] `make lint`
- [x] `make nextest`
- [x] New regression tests fail on the pre-fix code (`Exact(150000)` where `Inexact(150000)` is required) and pass after.

## Checklist

- [x] No behaviour change for scans without a pushed-down limit
- [x] `cargo fmt` clean

### DIFF
--- a/crates/data_components/src/flightsql.rs
+++ b/crates/data_components/src/flightsql.rs
@@ -486,21 +486,28 @@ impl FlightSQLTable {
         // Project the table-level statistics onto the scan's (projected) output
         // schema so the column-statistics list lines up with the output columns.
         //
-        // When filters are pushed down to the remote scan, the stamped statistics
-        // still describe the *unfiltered* slice the executor reported (they are the
-        // whole-table row count / column bounds, not the filtered subset). Every
-        // pushdown-eligible filter is reported `Exact` (see
-        // `supports_filters_pushdown`), so DataFusion drops the coordinator-side
-        // `FilterExec` — and if the stamped `num_rows`/bounds stay `Exact`, the
-        // `aggregate_statistics` optimizer rule folds `COUNT(*)`/`MIN`/`MAX` to
-        // those unfiltered values, silently ignoring the predicate (e.g. a filtered
-        // `COUNT(*)` returns the full table count — issue #11599). Marking the
-        // statistics inexact when any filter is applied disables that fold while
-        // keeping the bounds usable for join sizing.
+        // The stamped statistics describe the *whole* slice the executor reported
+        // (its row count and column bounds), so they only stay `Exact` for a scan
+        // that returns that whole slice. Both a pushed-down filter and a
+        // pushed-down limit narrow what the remote query returns:
+        //
+        // - Every pushdown-eligible filter is reported `Exact` (see
+        //   `supports_filters_pushdown`), so DataFusion drops the coordinator-side
+        //   `FilterExec`.
+        // - `LIMIT` is emitted into the remote SQL (see `FlightSqlExec::sql`), and
+        //   `supports_limit_pushdown` lets `LimitPushdown` drop the coordinator-side
+        //   `GlobalLimitExec`.
+        //
+        // Either way, leaving `num_rows`/bounds `Exact` lets the
+        // `aggregate_statistics` optimizer rule fold `COUNT(*)`/`MIN`/`MAX` to
+        // values the scan cannot produce — silently ignoring the predicate (a
+        // filtered `COUNT(*)` returning the full table count, issue #11599) or the
+        // limit (issue #12292). Marking the statistics inexact disables that fold
+        // while keeping the bounds usable for join sizing.
         let exec = match &self.statistics {
             Some(stats) => {
                 let stats = stats.clone().project(projections);
-                let stats = if filters.is_empty() {
+                let stats = if filters.is_empty() && limit.is_none() {
                     stats
                 } else {
                     stats.to_inexact()
@@ -918,6 +925,17 @@ impl ExecutionPlan for FlightSqlExec {
             (None, None) => None,
         };
 
+        // `LimitPushdown` calls this to move a `GlobalLimitExec` into the scan and
+        // then drops it, so the limited scan becomes the only thing describing the
+        // row count. `self.statistics` describes the *unlimited* slice, so it can
+        // only stay `Exact` while there is no limit — see `create_physical_plan` for
+        // the same reasoning about pushed-down filters (issues #11599, #12292).
+        let statistics = if merged_limit.is_some() {
+            self.statistics.clone().to_inexact()
+        } else {
+            self.statistics.clone()
+        };
+
         let new_plan = FlightSqlExec {
             projected_schema: Arc::clone(&self.projected_schema),
             table_reference: self.table_reference.clone(),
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12296, opened 2026-08-01
fix(runtime): gate the embed UDF name on the indexes that consume it (fixes #12294)

claudespice:
## Summary

`crates/runtime/src/embeddings/index/table.rs` declares `EMBED_UDF_NAME` in two arms gated on
`models` — the authoritative constant from `runtime_datafusion_udfs::embed` when that module is
compiled, a local copy otherwise. Both of its use sites, however, live in functions gated on a
*different* feature: `wrap_table_as_index_s3` is `#[cfg(feature = "s3_vectors")]` and
`wrap_table_as_index_elasticsearch` is `#[cfg(feature = "elasticsearch")]`.

So whenever neither of those two indexes is compiled in, the name has no consumer and one of the
two arms is a hard `-D warnings` failure in a file the contributor never touched:

| features | diagnostic |
| --- | --- |
| *(none)* | `error: constant EMBED_UDF_NAME is never used` |
| `models` | `error: unused import: runtime_datafusion_udfs::embed::EMBED_UDF_NAME` |

`crates/runtime` declares no `default` feature set, so the featureless case is exactly what a
package-scoped lint compiles — `make lint-rust PACKAGES=runtime`, a bare
`cargo clippy -p runtime -- -Dwarnings`, and therefore rust-analyzer configured per
`.vscode/settings.json.template`. The full gate never sees it: `make lint-rust` without `PACKAGES=`
supplies `--features adbc,…,models,…,elasticsearch,…`, which enables `models` *and* a consumer, and
the scoped pre-check inside `make signoff` resolves features from workspace `cargo metadata`, which
does the same.

## Changes

Gate both arms on the intersection of "which spelling of the name" and "is there a consumer at
all", so the item exists precisely when a use site does:

```rust
#[cfg(all(feature = "models", any(feature = "s3_vectors", feature = "elasticsearch")))]
use runtime_datafusion_udfs::embed::EMBED_UDF_NAME;
#[cfg(all(not(feature = "models"), any(feature = "s3_vectors", feature = "elasticsearch")))]
const EMBED_UDF_NAME: &str = "embed";
```

`elasticsearch` and `s3_vectors` do not imply `models`, so both arms stay reachable and the
`not(models)` copy is still required.

No runtime behavior changes: the constant's value is `"embed"` on both arms, and every feature
combination that compiled a use site before compiles the same use site now.

**Bug-class sweep.** The same `models`/`not(models)` twin declaration exists at three other places,
and each was checked for the same mismatch:

[...]

PR #12297, opened 2026-08-01
fix(runtime): gate the embed-UDF name on the backends that consume it (fixes #12294)

### LINKED ISSUE #12294: A featureless scoped lint of the runtime crate fails on a dead EMBED_UDF_NAME fallback
### Describe the bug

`crates/runtime/src/embeddings/index/table.rs` declares `EMBED_UDF_NAME` in two mutually exclusive arms:

```rust
#[cfg(feature = "models")]
use runtime_datafusion_udfs::embed::EMBED_UDF_NAME;
#[cfg(not(feature = "models"))]
const EMBED_UDF_NAME: &str = "embed";
```

but both of its use sites live in functions gated on a *different* feature:

| use site | enclosing function | gate |
| --- | --- | --- |
| `table.rs:215`, `:217` | `wrap_table_as_index_s3` | `#[cfg(feature = "s3_vectors")]` |
| `table.rs:440`, `:442` | `wrap_table_as_index_elasticsearch` | `#[cfg(feature = "elasticsearch")]` |

So when neither `s3_vectors` nor `elasticsearch` is enabled the name has no consumer, and the `not(models)` arm becomes dead code. `crates/runtime` declares no `default` features, so that is exactly what a plain package-scoped lint compiles.

### To Reproduce

On current `trunk`, run the scoped lint the `Makefile` documents in its own example comment:

```bash
make lint-rust PACKAGES=runtime
```

```
error: constant `EMBED_UDF_NAME` is never used
  --> crates/runtime/src/embeddings/index/table.rs:26:7
   |
26 | const EMBED_UDF_NAME: &str = "embed";
   |       ^^^^^^^^^^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`

error: could not compile `runtime` (lib) due to 1 previous error
```

The same error surfaces from a bare `cargo clippy -p runtime -- -Dwarnings`, and therefore from rust-analyzer configured per `.vscode/settings.json.template`.

### Expected behavior

A package-scoped lint of `runtime` succeeds, so a contributor validating a `runtime` change sees only diagnostics their own change caused. The failure today is in an unrelated file and gives no hint that the cause is a feature combination.

### Why this is not caught in CI

`make lint-rust` without `PACKAGES=` supplies `--features adbc,…,models,…,elasticsearch,…`, and the fail-fast scoped pre-check `make signoff` derives resolves features from workspace `cargo metadata`. Both enable `models` *and* `elasticsearch`, so the `not(models)` arm is never compiled and the consumers are never absent. Only the featureless scoped invocation — the one `CLAUDE.md` tells contributors to use for incremental validation ("Scope cargo to touched crates: `cargo check -p <crate>`") — hits it.

### Suggested fix

Gate both arms on the intersection of "which spelling" and "is there a consumer at all":

```rust
#[cfg(all(feature = "models", any(feature = "s3_vectors", feature = "elasticsearch")))]
use runtime_datafusion_udfs::embed::EMBED_UDF_NAME;
#[cfg(all(
    not(feature = "models"),
    any(feature = "s3_vectors", feature = "elasticsearch")
))]
const EMBED_UDF_NAME: &str = "embed";
```

The `models`-only arm has the mirror-image gap: with `models` on and both backends off the `use` has no consumer either, which is an unused import under the same `-D warnings`. The single `any(...)` guard closes both.

`crates/runtime/src/embeddings/udtf.rs` carries the same two-arm declaration but is **not** affected — its consumer at `udtf.rs:563` sits under `#[cfg(any(feature = "s3_vectors", feature = "elasticsearch", feature = "duckdb"))]`, which is a superset, so the name always has a use where it is declared.

claudespice:
## Summary

`crates/runtime/src/embeddings/index/table.rs` declared `EMBED_UDF_NAME` in two arms gated only on `models`:

```rust
#[cfg(feature = "models")]
use runtime_datafusion_udfs::embed::EMBED_UDF_NAME;
#[cfg(not(feature = "models"))]
const EMBED_UDF_NAME: &str = "embed";
```

Every use of that name, though, is inside a *different* feature's code: `wrap_table_as_index_s3` (`#[cfg(feature = "s3_vectors")]`, uses at `:215`/`:217`) and `wrap_table_as_index_elasticsearch` (`#[cfg(feature = "elasticsearch")]`, uses at `:440`/`:442`). Those are the only four occurrences of the identifier in the file.

So in a build with neither backend the name is declared and never consumed. `crates/runtime` declares no `default` features, which is exactly what a package-scoped lint compiles, and `dead_code` under `-D warnings` is a hard error there.

Fixes #12294

## Changes

Add `any(feature = "s3_vectors", feature = "elasticsearch")` to **both** arms, so the declaration exists in exactly the configurations that have a consumer. The `models` arm had the mirror-image gap — with `models` on and both backends off the `use` is an unused import — and the same guard closes it.

## Why CI never caught it

The workspace `make lint-rust` passes `--features adbc,…,models,…,elasticsearch,…`, and the fail-fast scoped pre-check `make signoff` derives resolves features from workspace `cargo metadata`. Both enable `models` *and* a backend, so neither the dead const nor the unused import is ever compiled. Only the featureless scoped invocation reaches it — the one `CLAUDE.md` recommends for incremental validation.

## Test plan

This is a conditional-compilation defect: which items exist in which feature combination. No unit test can observe it — the failing configuration does not compile, so the evidence is the build itself, before and after.

Reproduced and verified directly, same tree, same toolchain (1.96.1):

| configuration | before | after |
| --- | --- | --- |
| `make lint-rust PACKAGES=runtime` (no features) | exit 2 — `error: constant EMBED_UDF_NAME is never used --> crates/runtime/src/embeddings/index/table.rs:26:7` | **exit 0** |

The arm this change must not break is `not(models)` *with* a backend — and that is a shipping configuration, not a hypothetical: `bin/spiced`'s `default` feature list includes `runtime/s3_vectors` and does **not** include `models`. So `make build-cli-dev` compiles the `not(models)` const arm, and the guard keeps it declared there.

- [x] `cargo fmt -p runtime -- --check` clean
- [x] `make lint-rust PACKAGES=runtime` passes (it failed on `trunk`)
- [ ] `make lint` (full workspace gate — runs in sign-off)
- [ ] `make test` / `make nextest` (runs in sign-off)

### DIFF
--- a/crates/runtime/src/embeddings/index/table.rs
+++ b/crates/runtime/src/embeddings/index/table.rs
@@ -20,9 +20,18 @@ use crate::model::EmbeddingModelStore;
 use crate::secrets::Secrets;
 use datafusion::datasource::TableProvider;
 use datafusion::{prelude::SessionContext, sql::TableReference};
-#[cfg(feature = "models")]
+// Only the `s3_vectors` and `elasticsearch` index wrappers below look the UDF up
+// by name, so both spellings need that guard as well: without it the declaration
+// has no consumer in a build that enables neither, and `-D warnings` rejects it.
+#[cfg(all(
+    feature = "models",
+    any(feature = "s3_vectors", feature = "elasticsearch")
+))]
 use runtime_datafusion_udfs::embed::EMBED_UDF_NAME;
-#[cfg(not(feature = "models"))]
+#[cfg(all(
+    not(feature = "models"),
+    any(feature = "s3_vectors", feature = "elasticsearch")
+))]
 const EMBED_UDF_NAME: &str = "embed";
 use spicepod::vector::VectorStore;
 use std::sync::Arc;


claudespice:
@copilot review

claudespice:
Duplicate of #12296, which was opened first with an identical change. Verification evidence from this branch (before/after `make lint-rust PACKAGES=runtime`, and the `spiced` default-features point that makes the `not(models)` arm a shipping configuration) carried over there. Sign-off run for this branch cancelled to free the runner.

Issue #12179, opened 2026-07-29
Enhancement: honour the process's own cgroup memory limit, not a fixed /sys/fs/cgroup path


lukekim:
## Goal-State/What/Result

`spiced` sizes its memory budgets from the **effective cgroup limit of its own process**, not from a fixed absolute path. Concretely: `runtime::resource_monitor::get_container_memory_limit` resolves the process's cgroup from `/proc/self/cgroup`, walks that path to the root, and returns the **minimum** `memory.max` (v2) / `memory.limit_in_bytes` (v1) found along the way. When no limit applies anywhere on the chain, it keeps today's host-RAM fallback.

Target: a `spiced` launched under any memory-constrained cgroup — a Kubernetes pod with `cgroupns=host`, a `systemd` unit or scope with `MemoryMax=`, an LXC container, a nested container runtime — derives its query pool, Cayenne mem-tier, and compaction budgets from that limit, and reports the limit it detected in its startup log.

## Why/Purpose

**Today it silently ignores the limit and overcommits.** `get_container_memory_limit` reads the absolute paths `/sys/fs/cgroup/memory.max` and `/sys/fs/cgroup/memory/memory.limit_in_bytes` ([`resource_monitor.rs:46-68`](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/resource_monitor.rs#L46)). Those are the *container's own* limit only when the container runtime places the process's cgroup at the mount root — the `cgroupns=private` case. Everywhere else they are the root cgroup (`max`) or absent, so detection returns `None` and `get_total_memory()` falls back to `System::total_memory()` — the **whole host**.

Every downstream budget is then wrong in the unsafe direction, because they are all derived from that number:

- `runtime.query.memory_limit` default (`total/4`)
- `coordinated_mem_tier_budget` — ceiling `total/5`, float ceiling `total/4`, headroom `total/10`
- the dedicated compaction pool
- `cayenne::set_global_memory_budget`, which the adaptive tuner uses to compute memory pressure — so the closed loop optimises against a memory figure the process cannot actually have

The `coordinated_mem_tier_budget` invariant (`query + compaction + external + tier + headroom <= host`) still holds — against the *wrong* host.

**Observed, on a 121.7 GB bare-metal host with `MemoryMax=96G` on the scope:**

```
Out of memory: Killed process 322593 (spiced)
total-vm: 508 GB   anon-rss: 119.1 GB   file-rss: 2.2 MB
```

[...]

PR #12300, opened 2026-08-01
fix(telemetry): read the cgroup CPU quota along the whole cgroup path (fixes #12299)

### LINKED ISSUE #12299: HardwareInfo reports the host CPU count when the cgroup CPU quota is set on an ancestor, not the leaf
### Description

`HardwareInfo::detect()` reports `vcpu_count` from a cgroup CPU quota read **only at the process's own leaf cgroup**. A quota imposed on any *ancestor* — a systemd slice above the service, a Kubernetes pod cgroup above the container — is invisible, so the detected value falls back to the host core count and the reported figure can be many times the number of CPUs the kernel actually lets the process use.

This is the exact defect that #12179 described for memory and that #12263 fixed **in the same file**, for the same reason: the kernel enforces the smallest limit found anywhere on the path, so any single-level read can report "unlimited" for a capped process. #12263 introduced `min_limit_along_path()` in `crates/telemetry/src/hardware.rs` and routed `memory.max` / `memory.limit_in_bytes` through it. The CPU readers immediately above it in the same file were not moved over, so `HardwareInfo` now reports `total_memory_bytes` from a full cgroup walk and `vcpu_count` from a leaf-only read.

### Where

`crates/telemetry/src/hardware.rs`:

- `get_cgroup_v2_cpu_limit()` — builds one path, `<mountpoint><cgroup_path>/cpu.max`, and returns `None` if that single file says `max`.
- `get_cgroup_v1_cpu_limit()` — same shape for the `cpu.cfs_quota_us` / `cpu.cfs_period_us` pair.

Compare `get_cgroup_v2_memory_limit()` / `get_cgroup_v1_memory_limit()` a few hundred lines down, which walk leaf → mountpoint root and take the minimum.

### Reproduce

With cgroup v2, put a quota on a parent and leave the leaf unlimited — the arrangement `systemd-run --slice=capped.slice` produces, and the one the walk in `min_limit_along_path` was written for:

```
/sys/fs/cgroup/capped.slice/cpu.max      -> "400000 100000"   (4 CPUs)
/sys/fs/cgroup/capped.slice/svc/cpu.max  -> "max 100000"      (no quota at the leaf)
```

A process in `/capped.slice/svc` on a 64-core host is entitled to 4 CPUs. `detect_vcpu_count()` reads only the leaf, gets `None`, and falls through to `get_system_cpu_count()` → **64**.

### Impact

Reporting only. `HardwareInfo` feeds the startup debug line in `bin/spiced/src/lib.rs` and the anonymous-telemetry payload in `crates/telemetry/src/anonymous.rs`; nothing sizes a pool from it today. So this is a wrong number in telemetry and in the operator-facing log, not a runtime mis-sizing — worth stating plainly rather than overselling it. It still matters: the log line is what an operator reads when a capped deployment behaves as though it has the whole host, and fleet-wide telemetry that overstates vCPUs on every ancestor-capped deployment is misleading in exactly the population it most needs to describe.

### Deliberately not part of this

- **cpuset is already correct and should not be walked.** Cgroup v2 `cpuset.cpus.effective` is computed by the kernel *after* applying every ancestor's restriction, and cgroup v1 enforces at write time that a child's `cpuset.cpus` is a subset of its parent's. In both cases the leaf value is already the effective one, so `get_cgroup_v2_cpuset_effective()` / `get_cgroup_v1_cpuset()` need no change. Only the *quota* files carry an independent per-level value.
- **The CPU sizing work in #12211 / #12270.** That is about which number the runtime sizes its pools from; this is about the number `HardwareInfo` reports. Related note for whoever lands #12276: the new `cpu-budget` crate resolves its quota through `v2_file_path("cpu.max")`, a single leaf path, so it carries the same leaf-only shape — worth folding the walk in there too, since that reader *is* load-bearing.

### Expected

`detect_vcpu_count()` reports the smallest CPU quota binding anywhere on the process's cgroup path, matching what `detect_total_memory()` does for memory since #12263.

claudespice:
## Summary

`HardwareInfo::detect()` resolved the cgroup CPU quota from the process's **own leaf cgroup only**. A quota imposed on an *ancestor* — a systemd slice above the service, a Kubernetes pod cgroup above the container — read as unlimited, so `detect_vcpu_count()` fell through to `get_system_cpu_count()` and reported the host core count for a process entitled to a fraction of it.

The kernel enforces the smallest limit found anywhere on the path. That is exactly why #12179 / #12263 moved the *memory* readers in this same file onto a leaf→root walk (`min_limit_along_path`). The CPU readers a few hundred lines up were not moved over, leaving `HardwareInfo` reporting `total_memory_bytes` from a full walk and `vcpu_count` from a single level.

Fixes #12299

## Changes

`crates/telemetry/src/hardware.rs` only:

- Extracted `cgroup_path_ancestors()` — the leaf→root level sequence — out of the existing walk, and split the walk into `min_along_cgroup_path()` (calls a reader per level) with `min_limit_along_path()` layered on top for the single-file case. `min_limit_along_path` is now generic over the value type so the CPU readers (`usize`) and the memory readers (`u64`) share it; the memory call sites and their behaviour are unchanged.
- `get_cgroup_v2_cpu_limit()` now takes the smallest `cpu.max` quota along the path.
- `get_cgroup_v1_cpu_limit()` now takes the smallest quota along the path, reading `cpu.cfs_quota_us` **and** `cpu.cfs_period_us` from the *same* level — the period is per-cgroup, so pairing a quota with an ancestor's period would be meaningless.
- Both delegate to a `…_quota_at(mountpoint, cgroup_path)` helper so the walk is drivable over a temp directory in tests; the only part left out of test coverage is resolving those two arguments from `/proc`.
- Docs: `detect_vcpu_count` / `get_container_cpu_limit` / the module header no longer say "container", since a systemd `CPUQuota` or a capped slice binds identically.

**cpuset is deliberately not walked**, and the reason is recorded on `get_cpuset_effective_count()`: cgroup v2's `cpuset.cpus.effective` is computed by the kernel *after* applying every ancestor's restriction, and cgroup v1 enforces at write time that a child's `cpuset.cpus` is a subset of its parent's. In both cases the leaf value is already the effective one. Only the quota files carry an independent per-level value.

## Impact

Reporting only, and worth stating plainly rather than overselling: `HardwareInfo` feeds the startup debug line in `bin/spiced/src/lib.rs` and the anonymous-telemetry payload in `crates/telemetry/src/anonymous.rs`. Nothing sizes a thread pool or query fan-out from it today, so this is a wrong number in telemetry and in the operator-facing log, not a runtime mis-sizing.

Note for #12276 (which does make CPU sizing load-bearing): its new `cpu-budget` crate resolves the quota through `v2_file_path("cpu.max")`, a single leaf path, so it carries the same leaf-only shape. That is called out in #12299 rather than changed here — this PR does not touch that crate.

## Performance impact

None. The walk reads at most one file per cgroup level (typically 3–5) once, at startup, on a path that already did filesystem reads.

## Test plan

- `make lint` — clean.
- `make nextest` — 4 new unit tests in `crates/telemetry/src/hardware.rs`, all 48 hardware tests pass:
  - `walk_takes_the_smallest_v2_cpu_quota_on_the_path` — the `systemd-run --slice=capped.slice` arrangement (quota on the slice, `max` at the leaf) resolves to the slice's 4 CPUs; a tighter ancestor still beats a looser leaf.
  - `walk_pairs_v1_cpu_quota_with_its_own_level_period` — cgroup v1 with `-1` at the leaf and a 2-CPU quota over a *50ms* period on the parent resolves to 2, which is only correct if quota and period come from the same level.
  - `walk_reports_no_v2_cpu_quota_when_nothing_binds` — `max` at every level stays `None`, so detection still falls back to the host count.
  - `ancestors_run_from_the_leaf_to_the_mountpoint_root` — level sequence, including the root-only and trailing-separator cases.
- **Neuter check**: reverting both `…_quota_at` helpers to the pre-fix single-path read fails exactly the two walk tests (`left: None, right: Some(4)` and `left: None, right: Some(2)` — i.e. the reported "reads as unlimited" behaviour) while all memory tests stay green, confirming the tests exercise the real failure mode and the change is additive.

## Checklist

- [x] No behaviour change to the memory readers (their tests are untouched and green)
- [x] `cargo fmt` / scoped clippy clean
- [ ] `make signoff` green + `Attestation` re-run

### DIFF
--- a/crates/telemetry/src/hardware.rs
+++ b/crates/telemetry/src/hardware.rs
@@ -19,10 +19,11 @@ limitations under the License.
 //! Detects the number of vCPUs, GPUs, and available memory on the host machine,
 //! including support for containerized environments (Docker, Kubernetes).
 //!
-//! ## Container Support
+//! ## Cgroup Limits
 //!
-//! For containerized deployments, this module automatically detects and respects
-//! container resource limits from cgroup v1 and v2:
+//! This module detects and respects the cgroup v1/v2 resource limits that bind
+//! the process — in a container, but equally under `systemd-run -p CPUQuota=…`
+//! / `-p MemoryMax=…` or inside any capped slice:
 //!
 //! - **CPU limits**: Reads from cgroup v2 `cpu.max` or cgroup v1 `cpu.cfs_quota_us`
 //! - **Memory limits**: Reads from cgroup v2 `memory.max` or cgroup v1 `memory.limit_in_bytes`
@@ -31,6 +32,9 @@ limitations under the License.
 //!
 //! Cgroup mountpoints are resolved via `/proc/self/mountinfo` with `/sys/fs/cgroup`
 //! as the fallback. The process's cgroup path is read from `/proc/self/cgroup`.
+//! Each limit is read at every level from that path up to the mountpoint root,
+//! and the smallest is taken: the kernel enforces the tightest limit on the
+//! path, so a single-level read reports "unlimited" for a capped process.
 //!
 //! ## GPU Detection
 //!
@@ -136,8 +140,9 @@ impl std::fmt::Display for HardwareInfo {
 
 /// Detects the number of vCPUs available to the process.
 ///
-/// For containers, respects cgroup CPU limits. Falls back to host CPU count
-/// if container limits are not set or cannot be read.
+/// Respects a cgroup CPU limit binding anywhere on the process's cgroup path
+/// (containers, systemd `CPUQuota`, capped slices); falls back to the host CPU
+/// count when no limit binds or none can be read.
 fn detect_vcpu_count() -> usize {
     // First try container CPU limits (cgroup v2, then v1)
     if let Some(container_cpus) = get_container_cpu_limit() {
@@ -165,8 +170,12 @@ fn get_system_cpu_count() -> usize {
     }
 }
 
-/// Attempts to read container CPU limit from cgroup v2 or v1.
-/// Returns None if not in a container or if the limit cannot be read.
+/// The effective cgroup CPU limit for this process, or `None` when no limit
+/// binds anywhere on its cgroup path.
+///
+/// Not just for containers: a process launched under `systemd-run -p
+/// CPUQuota=…`, or inside any slice carrying a quota, is throttled exactly as
+/// hard as a container is.
 ///
 /// The effective CPU count is the minimum of:
 /// - CPU quota limit (from `cpu.max` or `cpu.cfs_quota_us`)
@@ -196,6 +205,12 @@ fn get_cpu_quota_limit() -> Option<usize> {
 
 /// Gets the effective CPU count from cpuset controller.
 /// cpuset can further restrict which CPUs are available beyond quota limits.
+///
+/// Unlike the quota files, cpuset needs no walk up the hierarchy: cgroup v2's
+/// `cpuset.cpus.effective` is computed by the kernel after applying every
+/// ancestor's restriction, and cgroup v1 enforces at write time that a child's
+/// `cpuset.cpus` is a subset of its parent's. Either way the leaf value is
+/// already the effective one.
 fn get_cpuset_effective_count() -> Option<usize> {
     // Try cgroup v2 cpuset.cpus.effective first
     if let Some(count) = get_cgroup_v2_cpuset_effective() {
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12317, opened 2026-08-01
fix(cayenne): the write-concurrency raise must respect the memory brake

lukekim:
## The gap

The dynamic tuner's memory brake (`mem_ok` / `MEM_PRESSURE_HIGH` / `MEM_PRESSURE_CRITICAL`) gates every buffer-grow move — except one. Both the legacy ladder's step 3 (*"falling behind, buffers maxed: raise write concurrency"*) and `decide_goal`'s replication-lag equivalent add encode shards **without consulting memory pressure**. The old comment said so out loud: *"Buffers maxed (or memory tight) → add encode parallelism"* — the deliberate theory being that when buffers can't grow, parallelism is the memory-safe way to drain backlog.

It isn't: every extra shard is another in-flight encode buffer and inline memtable. The lever buys lag relief with resident bytes.

## The measurement

SF-1000 CH-benCH, 3-host lab, spiced under a hard 96 GiB `MemoryMax` unit (`MemorySwapMax=0`), four arms (pg + mysql × two binaries differing only by cgroup-aware budget detection, #12263):

- RSS climbed **20 → 60 → 66 → 91 → 94 GiB in ~10 minutes** of snapshot bootstrap.
- The tuner — whose pressure reference correctly read the 96 GiB cap in the cgroup-aware arms — **raised `write_concurrency` at 0.98 pressure** (`"replication-lag goal, buffers maxed: raise write concurrency"` in the run logs, minutes before the kill).
- All four arms were `oom-kill`ed by the unit in 6m44s–9m37s, both sources, both binaries.

The inline-flush and mem-tier grows were already `mem_ok`-gated; the shard raise was the one memory-increasing actuator that bypassed the brake — and under a lag goal with maxed buffers it is exactly the actuator the controller reaches for.

## The fix

Both raise sites now carry the same `mem_ok` gate (`mem_pressure < MEM_PRESSURE_OK − drain_offset`) as every other grow move. Under pressure the tick falls through to the non-memory levers (compaction cadence) or the existing shrink tiers — which is the designed behavior everywhere else in the ladder.

Tests: two new regression tests pin the withheld raise (legacy path + goal path, pressure seeded between OK and HIGH so no shrink tier masks the assertion), and `no_goals_matches_legacy_decide` still passes since both paths changed symmetrically. `docs/cayenne/cayenne.md` gets its Document-changelog row per the repo rule.

Verified: `make lint-rust PACKAGES="cayenne"` green; 93 tuning tests green.

Related: #12263 (cap detection — necessary but not sufficient without this), #12179.


crates/cayenne/src/provider/tuning.rs
```diff
@@ -2362,15 +2362,21 @@ pub(crate) fn decide_with_goals(
...
+            //    files; ALSO withheld when the stream is delete-heavy, where extra
+            //    shards multiply the per-burst small-file fan-out and worsen delete
+            //    routing off the in-memory tier.
+            //    Also withheld when CPU-bound (more shards steal query threads),
+            //    I/O-/publish-bound (more shards = more files, uploads, and
+            //    metastore commits — the slow-storage/EBS bias), and under memory
+            //    pressure: every extra shard is another in-flight encode buffer and
+            //    inline memtable, so this lever buys lag relief with resident bytes
+            //    — near the budget that trade is an OOM kill, not a recovery
+            //    (measured at SF-1000 under a 96 GiB cgroup cap: the tuner raised
```
sgrebnov, line 2376: We should keep comments more generic and try to limit referring to specific internal experiments IMO

sgrebnov (approved):

PR #12321, opened 2026-08-01
fix(https): reject OAuth2 params on structured HTTP file datasets (fixes #12315)

### LINKED ISSUE #12315: OAuth2 auth params are silently dropped for structured HTTP file datasets, so requests go out unauthenticated
### Description

The `https`/`http` data connector routes a dataset to one of two providers:

- `HttpTableProvider` (the dynamic JSON API path), which is the only path that wires up authentication, and
- `HttpListingConnector` (the structured-file path), which builds a plain `object_store` `HttpBuilder` and carries only `username`, `password`, `port` and `client_timeout`.

`Https::read_provider` picks between them with `is_structured_format`. That function returns `true` **as soon as the file format is one of the structured formats** (`parquet`, `csv`, `tsv`, `arrow`, `avro`, `jsonl`, `ndjson`, `ldjson`, `soda`, `socrata`, `vortex`) — before any other parameter is consulted:

```rust
// crates/runtime/src/dataconnector/https.rs
if matches!(file_format.as_str(), "parquet" | "csv" | "tsv" | ...) {
    return true;
}
...
// only the JSON arm consults the dynamic-API gate
if file_format == "json" && !self.has_dynamic_api_params() {
    return true;
}
```

`has_dynamic_api_params()` → `params_indicate_dynamic_api()` is the gate that includes `any_oauth_param_set()`, and its own comment states the intent:

```rust
// OAuth2 authentication is only wired into the dynamic JSON API provider
// (the listing connector cannot apply it). Any OAuth2 param means the user
// intends the JSON API path — routing them to the listing connector instead
// would silently drop (and skip validating) their auth config.
```

That intent holds for `file_format: json`, and **only** for `file_format: json`. For every other structured format the OAuth2 gate is never reached.

### Impact

A dataset like:

```yaml
datasets:
  - from: https://api.example.com/exports/orders.csv
    name: orders
    params:
      file_format: csv
      auth_token_url: https://api.example.com/oauth/token
      auth_grant_type: client_credentials
      auth_client_id: ${secrets:client_id}
      auth_client_secret: ${secrets:client_secret}
```

loads without any error or warning, and then issues **completely unauthenticated** requests: no token exchange happens, `OAuth2Auth::try_new` is never called, and `resolve_oauth2_auth`'s validation (missing refresh token, bad grant type, invalid header name, insecure token URL) never runs. The user sees an opaque downstream failure — a 401/403 body handed to the CSV/Parquet parser — with nothing pointing at the auth config, or, worse, silently ingests whatever unauthenticated content the endpoint serves at that URL.

The same silent drop applies to the static `http_headers` parameter, which is the connector's other header-based auth mechanism (`http_headers: 'Authorization: Bearer …'`) and is likewise consumed only by `HttpTableProvider`.

### Expected

Two sibling guards already exist for exactly this shape and are invoked from `read_provider`:

- `ensure_rate_control_supported_for_structured_dataset`
- `ensure_client_identity_supported_for_structured_dataset`

Both return `InvalidConfigurationNoSource` naming the offending params. The OAuth2 params were added without the matching guard, so they fall through instead.

- OAuth2 params on a structured HTTP file dataset should be a load-time configuration error naming the params that were set.
- `http_headers` on a structured HTTP file dataset should at minimum warn that the headers are not applied (it predates the OAuth2 work and is widely used, so failing the load would break configurations that work today).

### Steps to reproduce

Configure the dataset above against any endpoint that requires a bearer token and start `spiced`. The runtime starts cleanly; the outgoing request carries no `Authorization` header.

claudespice:
Fixes #12315

## Summary (root cause)

The `https`/`http` connector serves a dataset from one of two providers, and only one of them can authenticate. `HttpTableProvider` (the dynamic JSON API path) applies OAuth2 tokens and the static `http_headers`; `HttpListingConnector` builds a plain object-store `HttpBuilder` that carries only `username`, `password`, `port` and `client_timeout`.

`is_structured_format` picks between them, and it returns `true` as soon as the file format is a structured one — before the dynamic-API gate is consulted:

```rust
if matches!(file_format.as_str(), "parquet" | "csv" | "tsv" | ...) {
    return true;                      // <- short-circuits here
}
...
if file_format == "json" && !self.has_dynamic_api_params() {
    return true;                      // <- only this arm runs the gate
}
```

`has_dynamic_api_params()` is what folds in `any_oauth_param_set()`, and the comment on that gate states the intent in as many words: routing an OAuth2 dataset to the listing connector "would silently drop (and skip validating) their auth config". That intent was only ever enforced for `file_format: json`.

So a `csv`/`parquet`/`avro`/... dataset that configures OAuth2 starts cleanly, never exchanges a token, never runs any of `resolve_oauth2_auth`'s validation, and issues completely unauthenticated requests. The user gets an opaque parser failure on a 401/403 body — or silently ingests whatever the endpoint serves unauthenticated.

Two sibling guards already existed for exactly this shape and are invoked from `read_provider` — `ensure_rate_control_supported_for_structured_dataset` and `ensure_client_identity_supported_for_structured_dataset`. The OAuth2 params shipped without the matching one.

## Changes

- New `Https::ensure_auth_supported_for_structured_dataset`, called from `read_provider` alongside the two existing structured-dataset guards.
  - **OAuth2 params are a hard `InvalidConfigurationNoSource`**, naming the params that were actually set and pointing at the dynamic JSON API dataset. They exist only to authenticate, so a dataset that sets them and sends nothing is never what the user meant.
  - **`http_headers` warns and continues.** It predates the OAuth2 params, is used for far more than auth, and its headers never reached the wire on this path anyway — failing the load would break configurations that work today. The drop is no longer silent.
- Corrected the `params_indicate_dynamic_api` comment, which claimed a coverage the gate does not have, and pointed it at the new guard.
- `auth_token_url` and `http_headers` parameter descriptions now state the structured-dataset behavior, matching the wording already used by the mTLS params.

## Test plan

- `make lint`
- `make test`

New unit tests in `crates/runtime/src/dataconnector/https.rs`:

- `test_http_structured_format_rejects_oauth_params` — the regression test. Iterates the whole `STRUCTURED_FILE_FORMATS` list (the same const `is_structured_format` now matches against, so the test cannot drift from it) and asserts each one errors, that the message names both configured params and no unset one, and that it points at the dynamic JSON API dataset. Fails on the pre-fix code, which hands the dataset to the listing connector instead.
- `test_http_json_format_with_oauth_params_stays_on_the_api_path` — the `json` arm already routed to the API provider; asserts the new guard does not intercept it and that the auth config still reaches `resolve_oauth2_auth`'s validation.
- `test_http_structured_format_does_not_reject_non_oauth_params` — a structured dataset still loads with no OAuth2 params, with and without `http_headers`.

Verified locally: the two clippy passes the gate runs (`--lib --bins`, then `--tests`) are clean for this crate, and the regression test fails with the guard call neutered and passes with it restored.

## Checklist

- [ ] Regression test fails on the old code and passes on the new
- [ ] `make lint` clean
- [ ] `make test` clean

### DIFF
--- a/crates/runtime/src/dataconnector/https.rs
+++ b/crates/runtime/src/dataconnector/https.rs
@@ -143,19 +143,7 @@ impl Https {
             .map_or_else(|| "auto".to_string(), str::to_ascii_lowercase);
 
         // Check if explicitly configured as a structured format
-        if matches!(
-            file_format.as_str(),
-            "parquet"
-                | "csv"
-                | "tsv"
-                | "arrow"
-                | "avro"
-                | "jsonl"
-                | "ndjson"
-                | "ldjson"
-                | "soda"
-                | "socrata"
-        ) {
+        if STRUCTURED_FILE_FORMATS.contains(&file_format.as_str()) {
             return true;
         }
 
@@ -670,6 +658,49 @@ impl Https {
                 "mTLS client identity parameters are not supported for structured HTTP file datasets that use the listing connector. Remove {params}, or use a dynamic JSON HTTP API dataset."
             ),
         })
+    }
+
+    /// Reject `OAuth2` parameters on a dataset bound for the listing connector,
+    /// which builds a plain object-store HTTP client and cannot carry an access
+    /// token. These params exist only to authenticate, so a dataset that sets
+    /// them and then sends nothing is never what the user meant — and routing
+    /// there skips `resolve_oauth2_auth`'s validation as well.
+    fn ensure_auth_supported_for_structured_dataset(
+        &self,
+        dataset: &Dataset,
+    ) -> DataConnectorResult<()> {
+        if !any_oauth_param_set(&self.params) {
+            return Ok(());
+        }
+
+        let params = OAUTH_PARAM_KEYS
+            .iter()
+            .filter(|&&name| param_is_set(&self.params, name))
+            .map(|&name| format!("'{}'", self.params.user_param(name)))
+            .collect::<Vec<_>>()
+            .join(", ");
+
+        Err(DataConnectorError::InvalidConfigurationNoSource {
+            dataconnector: "https".to_string(),
+            connector_component: ConnectorComponent::from(dataset),
+            message: format!(
+                "OAuth2 authentication parameters are not supported for structured HTTP file datasets that use the listing connector; requests would be sent unauthenticated. Remove {params}, or use a dynamic JSON HTTP API dataset."
+            ),
+        })
+    }
+
+    /// The listing connector cannot carry `http_headers` either. Unlike the
+    /// `OAuth2` params this only warns: the header list is used for far more
+    /// than authentication, so rejecting it would fail datasets that load and
+    /// serve correctly today.
+    fn warn_ignored_http_headers(&self, dataset: &Dataset) {
+        if param_is_set(&self.params, "http_headers") {
+            tracing::warn!(
+                "Dataset {}: '{}' is not applied to structured HTTP file datasets, which are served by the listing connector. The headers are ignored; use a dynamic JSON HTTP API dataset if the endpoint requires them.",
+                dataset.name,
+                self.params.user_param("http_headers"),
+            );
+        }
     }
 
     fn resolve_client_identity_config(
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12335, opened 2026-08-02
fix(logging): let log colour follow the output sink (fixes #12327)

### LINKED ISSUE #12327: spiced colorizes its log unconditionally, so a redirected spice.log is full of ANSI escapes
## Problem

`bin/spiced/src/tracing.rs` builds both `fmt::layer()`s with a hard-coded `.with_ansi(true)` (lines ~206 and ~228). `spiced` therefore emits SGR escapes even when its output is not a terminal, and it ignores `NO_COLOR`.

The convention is the opposite: colorize only when the sink is a TTY, and honour `NO_COLOR` when it is set to anything.

## Impact

Every redirected log carries escapes. `spice run &> spice.log` produces lines like:

```
WARN\e[0m \e[2mruntime::accelerated_table::refresh_task\e[0m: Failed to load data for dataset ...
```

so a pattern written the way the line *reads* (`WARN +runtime::accelerated_table::refresh_task:`) does not match. Our own E2E log checks hit this and had to grow a `sed $'s/\x1b\\[[0-9;]*m//g'` prefix before any pattern could work (run `29659183104` failed on an already-allowed transient for exactly this reason). Anyone grepping a captured `spice.log` — in CI, in a container log, in a support bundle — pays the same tax, and most will not know why their pattern misses.

## Suggested fix

Replace the two literals with the terminal/`NO_COLOR` check:

```rust
let ansi = std::io::stderr().is_terminal() && std::env::var_os("NO_COLOR").is_none();
```

and pass `ansi` to both `fmt::layer()` builders (`std::io::IsTerminal` is in std). Interactive `spice run` keeps its colors; every redirected sink becomes plain text.

Worth considering alongside it: a `--no-color` flag, and whether `spiced` running under a log collector should default to a structured (JSON) format instead.

## Related

`.github/actions/check-spice-log` currently strips ANSI itself so its allow-list works. That workaround can be deleted once this lands.

claudespice:
## Summary

`spiced` built its console `fmt` layer with a hard-coded `.with_ansi(true)`, so every log line carried SGR escapes no matter where stdout pointed. A redirected `spice.log` — in CI, in a container, in a support bundle — came out as

```
WARN\e[0m \e[2mruntime::accelerated_table::refresh_task\e[0m: Failed to load data for dataset ...
```

so a pattern written the way the line *reads* (`WARN +runtime::accelerated_table::refresh_task:`) cannot match it. An explicit `true` is also stronger than the library default: `tracing_subscriber`'s own default already honours `NO_COLOR`, so passing `true` opted `spiced` out of the one escape hatch a user had.

Colour now follows the sink: escapes when stdout is a terminal, plain text otherwise, with `NO_COLOR` and `FORCE_COLOR` overriding.

Fixes #12327

## Changes

- **`bin/spiced/src/tracing.rs`** — the console-layer construction moves into a `console_layer(ansi, writer)` helper, and `init_tracing` computes `ansi` once from `ansi_colors::colors_enabled_for(Target::Stdout)`. The pre-existing `candle_console_output!` test macro, which hand-built the same layer, now goes through the helper too, so it can no longer drift from what production assembles.
- **`crates/util/src/tracing_util.rs`** — the same defect in the bootstrap subscriber, and the worst instance of it. `in_tracing_context` / `in_tracing_context_async` back `spiced`'s "Starting runtime …" banner (`bin/spiced/src/main.rs:114`), its **fatal-error line** (`main.rs:103`), the Spicepod load warnings in `crates/app`, and the secrets preflight in `crates/runtime/src/builder.rs`. All of that goes to the same stdout, and all of it was unconditionally coloured — including the one line a captured log is most often grepped for.
- **`bin/spice/src/main.rs`** — the CLI's subscriber had the same defect in a milder form: the builder default honours `NO_COLOR` but still paints a redirected stdout. It now uses the same decision, which is also the one the CLI's own `Painted` output already makes, so the two agree within a process.
- **`bin/spiced/Cargo.toml`, `crates/util/Cargo.toml`** — `ansi_colors` (an existing `foundation`-tier crate) added as a path dependency. `util` is itself `foundation`, so this is a same-tier edge; `scripts/check_crate_layers.py` passes.

Two details worth flagging for review:

- **The stream is stdout, not stderr.** The issue suggested gating on `std::io::stderr().is_terminal()`, which would have been wrong: `fmt::layer()`'s default writer is `io::stdout` (`tracing-subscriber-0.3.23/src/fmt/fmt_layer.rs:749`), so `spice run > spice.log` — stdout redirected, stderr still a TTY — would have kept colouring the file. The helper now names its writer explicitly rather than relying on that default.
- **Interactive `spice run` is unaffected.** `bin/spice/src/commands/run.rs` spawns `spiced` with `Stdio::inherit()`, so an interactive run's stdout *is* the user's terminal and keeps its colours. `FORCE_COLOR` covers the case of a pipeline that wants them anyway.

## Out of scope

- `--no-color` flag and a JSON log format, both raised as "worth considering" in the issue. `NO_COLOR` is the conventional control and is now honoured; a structured format is a separate change.
- The `with_ansi(true)` calls in `crates/llms/tests/integration.rs`, `crates/runtime/benches/utils/mod.rs`, `crates/runtime/tests/cluster/distributed_acceleration.rs` and `crates/scheduler/src/scheduler.rs` (the last under `#[cfg(test)]`). Same shape, but they configure test and bench harnesses whose stdout the test runner captures, not a shipped binary's sink.
- `tools/flightsubscriber/src/main.rs:118`. Same one-line fix, but it is a standalone debug tool outside the `spice`/`spiced` pair this issue is about.
- Removing the ANSI strip from `.github/actions/check-spice-log`. That action does not exist on `trunk` yet (it arrives with #12331), and the strip stays useful as defence in depth against a dependency that paints its own output.

## Test plan

- `make lint`
- `make test`
- New unit tests in `bin/spiced/src/tracing.rs`, both driving the production `console_layer` through a capturing writer:
  - `console_layer_omits_ansi_escapes_when_the_sink_is_not_a_terminal` — asserts no `\x1b` byte reaches the writer, that the event text still does, and that only whitespace separates the level from the target (i.e. the line reads the way a grep pattern would be written). Fails on the old code, which emitted escapes unconditionally.
  - `console_layer_emits_ansi_escapes_when_the_sink_is_a_terminal` — the other half, so a layer that dropped the flag on the floor cannot pass the first test for the wrong reason.
  - Both also assert the `task_history` target stays out of the console, which is the filter the extraction had to preserve.

## Checklist

- [x] Tests added for the regression
- [x] `cargo fmt` clean
- [x] Crate layering check passes
- [x] No behaviour change for an interactive terminal

## Base merge

`trunk` is merged into this branch to resolve a conflict in `bin/spiced/src/tracing.rs` with #12281, which consolidated `init_tracing`'s two subscriber stacks into one and so removed one of the two console-layer call sites this branch originally patched. The surviving console layer keeps the sink-derived `ansi` flag; #12281's conditional task-history sink, always-installed progress layer, and always-installed `LogTracer` are preserved unchanged.


claudespice:
@copilot review

lukekim (approved):

PR #12367, opened 2026-08-02
feat(cayenne): inline a small whole-table refresh into the metastore

lukekim:
## What

A `refresh_mode: full` refresh small enough to clear the inline-admission caps now
stores its rows as **one `cayenne_inlined_data` metastore row** instead of writing
`cayenne_write_concurrency` tiny Vortex files.

Stacked on `cayenne-full-refresh-profile` (#12338); base is that branch, not `trunk`.

## Why

That parent PR turns the background compactor off for the bulk-overwrite profile,
which is correct — a whole-table replace resets the small-file counter and creates no
protected snapshot, so every compaction trigger early-outs. But it leaves a gap: a
small full refresh still fans its handful of rows out across the write concurrency,
and with no background compactor those files are never merged. A dimension table
refreshed on a schedule accumulates a fresh set of them on every refresh.

The `BulkOverwrite` profile therefore gains the same static inline-admission caps
`SmallWrite` already uses (1,024 rows / 1 MiB serialized / 4 MiB buffer). The
*flush* caps stay at their defaults: a whole-table replace leaves exactly one entry,
so the cumulative flush gate never binds.

The replacement row is inserted by the **same transaction** that clears the previous
inline corpus and flips the snapshot pointer, so clear-then-replace is atomic exactly
as clear-then-flip already is.

## The correctness problem this had to solve first

An overwrite's catalog transaction and its in-memory publish are two separate atomic
units, and a scan holds `listing_fence.read()`, so it can land between them:

1. `begin_overwrite` buffers, reserves `S_new`, builds the payload
2. `apply_owned_txn` / `apply_in_txn` — one transaction: delete every inline row,
   insert the new one at `S_new`, flip `current_snapshot_id`
3. `finish` → `publish_overwrite_snapshot` under `listing_fence.write()`

Between 2 and 3, the in-memory snapshot pointer still names the *previous* snapshot —
whose directory is empty if that refresh was itself inlined — while the catalog holds
only the new row, whose sequence is above the still-unadvanced inline watermark. The
read path skips entries above the watermark, so the inline view comes back empty. An
empty inline view unioned with an empty directory is an **empty table**, returned as a
successful query.

### The fix

`begin_overwrite` publishes the inline visibility state *before* the transaction runs
(`CayenneTableProvider::prepublish_inlined_overwrite`), under the per-table write lock:

1. `published_inlined_seq.fetch_max(S_new, Release)` — the watermark only ever admits
   rows, and every pre-overwrite entry sits strictly below the old watermark, so
   raising it hides nothing.
2. `durable_inlined_row_count.fetch_max(new_row_count)` — a saturating max, never a
   store (a store could shrink a larger pre-overwrite corpus). Over-estimating only
   forces a full catalog read, which is always correct; under-estimating is what loses
   rows via the empty-corpus fast path.
3. `bump_inlined_structural_epoch()` — invalidates any materialized view, so every read
   in the window reaches the catalog, and the catalog is only ever in one of two whole
   states because step 2 above is a single transaction.

`inlined_row_count` is deliberately **not** raised. It gates whether a scan consults the
tier at all, and leaving it at its pre-overwrite value is what keeps a *file-backed*
predecessor reading its own directory instead of mixing those files with the
replacement rows.

Rollback-safe: all three are harmless over-approximations if the transaction fails.
`finish` still invalidates and re-seeds the exact values under the fence.

### Two things found while implementing it

**A mem-tier seal shadow makes the watermark raise unsafe.** `seal_mem_tier_durable`
commits the RAM tier's contents as an inline row at a sequence deliberately kept *above*
the watermark while those same rows stay live in the tier, and it does so off the
per-table write lock. Raising the watermark past it would serve every sealed row twice.
Inlining an overwrite is therefore refused outright while the mem tier is non-empty or a
shadow is present (`inline_overwrite_admissible`). Only `cdc_durability: memory` can
produce one, which the bulk-overwrite profile never selects — so this bites only on a
hand-issued overwrite against a CDC table, which keeps the Vortex path it uses today.

**The same window exists for an overwrite that does NOT inline**, and this PR is what
makes it reachable: its transaction clears the inline corpus and puts nothing back, so a
rebuild inside the window returns an empty view — and now the predecessor snapshot can be
an inlined one with an empty directory. Fixed by the mirror-image
`warm_inlined_cache_for_overwrite`: materialize the pre-overwrite view before the
transaction clears it, so the window is served from memory. (Before this PR a
`refresh_mode: full` table could never hold an inline corpus, so the pairing could not
arise.)

## Tests — confirmed failing before the fix

`crates/cayenne/src/provider/overwrite.rs` gains a test module that drives the
`PreparedOverwrite` lifecycle by hand to put a scan between the commit and the publish,
with the inline cache dropped to reproduce the metastore-rebuild path a freshly opened
provider takes. Each asserts the scan returns a COMPLETE table — every pre-overwrite row
or every post-overwrite row — and specifically not zero.

With the pre-commit publish and the cache warm removed, all three window tests fail with
zero rows:

```
scan_between_commit_and_publish_sees_a_complete_table ... FAILED
  inlined overwrite: a scan between an overwrite's commit and its publish returned
  ZERO rows; expected either the complete pre-overwrite set [1, 2, 3, 4, 5] or the
  complete post-overwrite set [7, 8]

cross_partition_overwrite_window_sees_complete_partitions ... FAILED
  partition 0 with 0 partition(s) published: ... returned ZERO rows; expected either
  the complete pre-overwrite set [1, 2] or the complete post-overwrite set [7, 8, 9]

scan_between_commit_and_publish_of_a_file_backed_overwrite ... FAILED
  file-backed overwrite: ... returned ZERO rows
```

With the fix, all pass. The cross-partition test uses the
`cross_partition_overwrite_test.rs` shape: every partition commits in one shared
transaction, then publishes one at a time, and every partition is re-scanned after each
publish — so partitions ahead of the cursor are still in the window while the ones
behind it are published.

Also added:

- `commit_overwrite_replaces_inlined_corpus_atomically` — catalog-level: the replacement
  row lands in the same transaction as the clear, every other per-snapshot table is still
  cleared, and a second inlined overwrite leaves exactly one entry (the tier does not
  accumulate one row per refresh).
- `small_overwrite_is_inlined_and_replaced` — end-to-end inline/replace/empty-refresh.
- `repeated_inline_overwrites_keep_the_wal_bounded` — see below.

## Also in this PR

**Partition-aware inline accounting.** A partitioned overwrite runs one `begin_overwrite`
per partition concurrently, so N partitions would hold N buffers and N blobs against a
runtime reservation of one each. Admission is now claimed through a single slot on the
shared `CayenneContext`, which a partitioned dataset's children share — so the
reservation is true rather than merely bigger. The slot is `try_acquire`, **never**
awaited: partition children are coupled writers fed by one demux, and parking there would
stall the router and starve the slot-holder (spiceai/spiceai#11818). A child that cannot
claim it writes Vortex files, which is what every overwrite did before.
`estimate_cayenne_reservation_bytes` books one entry plus one buffer for every profile
that inlines, reading `cayenne_inline_max_bytes` / `inline_max_bytes` /
`cayenne_inline_max_buffer_bytes` through the same `parse_u64` idiom as the rest of that
function.

**SQLite WAL.** An inlined overwrite writes an Arrow IPC BLOB per refresh, the inline
auto-checkpoint is off by default, and `PreparedOverwrite::finish` never scheduled
maintenance — so repeated refreshes grew the WAL without bound. `finish` now schedules a
debounced WAL-only maintenance checkpoint AFTER the commit (a checkpoint inside the
transaction would land an fsync in its WAL-write-locked window).
`repeated_inline_overwrites_keep_the_wal_bounded` asserts the plateau: tripling the
refresh count must not double the `-wal` file. Without the scheduling it grows
1.19 MB → 3.96 MB.

**Quality items.**

- The overwrite path reuses `InlineBatchBuffer` instead of hand-rolling the buffering
  loop, which also gets it the `overflow_reason()` / `track_inline_fallback` admission
  metric the CDC path already had. `into_chained_stream` now replays its owned batches
  with `stream::iter(...).chain(remainder)` rather than building a `MemorySourceConfig`,
  which drops its `TaskContext` parameter and makes it infallible.
- One helper (`inlined_data_insert`) for the `INSERT INTO cayenne_inlined_data` statement,
  folding in the empty-`inlined_id` → `Uuid::now_v7()` rule. It had two copies
  (`add_inlined_data`, `commit_inlined_mutation`) and this PR's commit path would have
  been a third.
- The admission probe returns `enum OverwriteAdmission { Inlined(..) | Fallback(stream) }`
  rather than `(stream, Option<payload>)`, so the inline arm cannot hand back a
  live-but-drained stream.
- `RefreshWriteProfile::inlines_small_writes()` names the set of profiles that use the
  tier.

**Docs.** `docs/cayenne/cayenne.md`: the Appendix B write-profile table now shows
inlining on for bulk-overwrite (with the flush-cap caveat), the write-path decision tree
gains the whole-table-replace branch, the inline-threshold prose states the visibility
model, and there is a new *Document changelog* row.

## Compatibility

No metastore schema change, no config-key change. `MetadataCatalog::commit_overwrite` and
`CayenneCatalog::commit_overwrite_in_txn` take an extra `Option<&InlinedData>`; there is
one implementor and the cold-graduation caller passes `None`.

## Lab validation

Measured on the 3-node lab (20-core aarch64); control is this PR's base
(`cayenne-full-refresh-profile`), built from the same toolchain and feature set.

**Correctness.** A refresh-hammer pod re-refreshes the four `refresh_mode: full`
reference tables every 10s — ~90 whole-table replaces per table per run instead
of one load — under the full CH-benCH HTAP workload, while a probe counts those
tables continuously. TPC-C never writes them, so their counts are constant by
construction and any movement is a defect. SF-100 Postgres, arm VALID, 7/7
freshness tables: **43,392 samples across `item`/`supplier`/`nation`/`region`,
every one at the constant value — zero deviations, zero empty reads.**

**Write-side effect**, same workload:

| | control | this PR |
|---|---|---|
| `region` / `nation` Vortex files | 13 each | **0** |
| metastore directory | 9.1 GB | **393 MB** |
| WAL across the run | 160 MB → 9.7 GB, monotonic | bounded, 36–174 MB |

The WAL result is broader than the claim above: `finish` schedules the checkpoint
unconditionally, so it bounds the whole overwrite transaction volume, not just
the inline blob.

**Query performance.** TPC-H and TPC-DS at SF 1 / 5 / 10 / 50 / 100 plus
ClickBench, three arms — control, this PR, and this PR with
`cayenne_inline_max_rows/bytes/buffer = 0`, which isolates inlining from every
other difference in the branch — 2–3 reps each, result caching disabled.

- TPC-DS geomean within ±1% of control at every SF ≥ 5 (SF-100: 0.997).
- TPC-H converges toward neutral as scale grows (0.936 → 0.987): `region`+`nation`
  are 30 rows at every SF while the fact tables grow 100×.
- ClickBench flat (0.991 total / 1.005 geomean) — nothing inlines there, so it is
  the null control.
- One query regresses consistently — `tpcds_q64`, +11% at SF-100 (4864 → 5416 ms),
  and at SF-50/SF-100 it is the only TPC-DS query slower than both baselines by
  >5%. It does not compound with scale (0.65× at SF-1, then 1.18 / 1.10 / 1.11).
  An inlined table scans as a single-partition memory source, so the planner
  coalesces where a file-backed scan repartitions. Tracked in #12429.

## Note on scope

Enabling the inline caps for this profile also routes a small `INSERT` into a
table on it into the inline tier, not only a whole-table refresh. That tier
evaluates scan predicates, so such a table now returns exactly the matching rows
from `TableProvider::scan` where it previously returned all of them — both are
legal under the inexact pushdown contract and the tighter answer is the better
one, so `test_filter_predicates` moves rather than the behavior.

## The second interaction: an inline corpus that outlives a schema widening

Making this profile inline also put it in reach of **open-time schema
evolution**, and the two had never overlapped before.

An inline entry is Arrow IPC frozen at the schema live when it was written, and
the table schema can be widened from two places:

- `evolve_schema_live` (stream-time) flushes the RAM tier *and* the durable
  inline corpus before swapping — exactly because inlined batches are unioned
  into scans projection-only, with no cast. Old Vortex files need no flush: they
  are self-describing and the opener adapts them.
- `try_widening_schema_evolution` (open-time — the path that adopts a source's
  added column at restart under `append_new_columns` / `sync_all_columns` /
  `drop_and_recreate`) commits the evolved schema straight to the metastore and
  rewrites nothing.

A `refresh_mode: full` table can now be holding a corpus when the second one
runs, so the reopened provider resolved live-schema column indices against a
batch still at its narrower write-time width:

```
Failed to project inlined batch for table sample:
Schema error: project index 4 out of bounds, max field 4
```

The gap was already visible in the tree: `schema_evolution_restart_scans_old_files_under_widened_schema`
zeroes the inline caps with the comment *"the adaptation under test … is the
file read path, not the inline corpus."*

**Fixed by adapting at decode** rather than adding a second flush site: missing
nullable columns null-filled, widened columns cast — what the Vortex opener
already does for pre-evolution files, and the truth for a row written before the
widening, since a widening plan admits only added *nullable* columns and
lossless casts. Both data-batch decodes are covered: the one the full-rebuild
and delta cache paths share, and the inline DELETE rewrite (a predicate naming
an added column has the same problem). A corpus already at the live width
short-circuits on a field comparison, so the CDC read path is unchanged.
`evolve_schema_live` still flushes before its own swap; this is the restart-path
counterpart, not a replacement.

`schema_evolution_restart_scans_inlined_rows_under_widened_schema` is the
inline-corpus counterpart to the file-backed test above, and asserts the premise
— that the rows really are in the corpus — so it cannot silently degrade into a
duplicate of it.


sgrebnov (dismissed):

phillipleblanc:
integration tests failed when this PR went into the merge queue: https://github.com/spiceai/spiceai/actions/runs/30985258406/job/92262792407

phillipleblanc:
Removed from the merge queue: two cayenne partition-by-bucket tests fail deterministically (6/6 nextest retries each) against this PR's speculative commit:

- `acceleration::partition_by_cayenne::test_cayenne_partition_by_bucket_numeric_nulls` — insta snapshot assertion for `numeric_null_partition_full_scan` failed
- `acceleration::partition_by_cayenne::test_cayenne_partition_by_bucket_with_nulls` — insta snapshot assertion for `null_partition_full_scan` failed

Both are NULL-handling regressions in the partition-by-bucket read path. This PR touches `crates/cayenne/src/provider/table.rs`, `catalog.rs`, `context.rs`, and `maintenance.rs` — the same code the failing tests exercise — so this looks like a real regression from this change rather than flakiness (failure is 100% reproducible, not intermittent). It also carried through to an unrelated PR queued behind this one, since the merge queue's speculative commit for that entry included this PR's changes.

Please investigate the partition-by-bucket NULL handling and update or fix as needed, then re-add to the queue.

lukekim:
Fixed in caaa04f210 — thanks for the precise repro. The cause turned out not to be NULL handling, so writing up what it actually was.

**It isn't NULL-specific, and it isn't a data bug.** Four tests fail in the queue, not two — `test_cayenne_partition_by_bucket` and `test_cayenne_partition_by_date_part` fail the same way and have no NULLs at all. In every case the row-count assertions pass (they run *before* the snapshot), and only the `*_full_scan` plan snapshot differs, always identically:

```
-    CayenneAccelerationExec: snapshots_scanned=1, files_scanned=1
+    CayenneAccelerationExec: snapshots_scanned=0, files_scanned=0
-        DataSourceExec: file_groups={1 group: [[<TEMP_PATH>/.vortex]]}, ... file_type=vortex
+        DataSourceExec: partitions=1, partition_sizes=[1]
```

Exactly one partition is served from the metastore instead of a Vortex file. The data was right; the physical layout was not.

**Root cause.** This PR switches on inline admission for the bulk-overwrite profile, which is what `refresh_mode: full` resolves to — the mode all four tests use. `inline_overwrite_admissible` tried to exclude partitioned tables with `partition_column.is_none()`, and its comment reasoned that a partitioned dataset's children were "unaffected". But the children are created with `partition_column: None` (a partition is not itself partitioned), and they are the tables that actually write — so that rule never reached them. They also share one `CayenneContext`, hence its single `try_acquire`-only inline-admission slot, and they overwrite concurrently under one routing demux. One child claimed the slot and inlined; the rest fell back with `admission_busy`.

So *which* partition inlined was a race. Re-recording the snapshots would have baked that race into the expected output — a flaky test, and an arbitrary split layout blessed as correct.

**Fix.** Refuse coupled writers — the flag partition children already carry for the encode-budget exemption, which also covers the DDL `PARTITION BY` path. Widening the slot isn't available as an alternative: awaiting it reintroduces the #11818 demux hold-and-wait deadlock, and one permit per child would multiply the buffered-admission memory reservation by an unbounded partition count. Little is given up, since `serialize_partition_child_writes` already clamps a child to `write_concurrency = 1` — a refresh leaves one file per partition and the next overwrite replaces it, not the growing pile of tiny files inlining exists to prevent.

The four snapshots are therefore unchanged, and all five `partition_by_cayenne` tests pass locally. Added a unit regression test (`partition_child_overwrite_is_never_inlined`, with an ordinary table on the same caps and catalog as the contrast half so it can't pass with inlining merely switched off), and corrected `docs/cayenne/cayenne.md`, which documented the old "a child that cannot claim it writes Vortex files instead" behavior.

**Two unrelated failures in that same queue run**, which I have not touched: `kafka_sasl_connect_test` failed pulling the `redpandadata/redpanda` image, and `postgres::catalog_changes::test_catalog_acceleration_reuses_slot_across_restart`. Neither touches this code path.

`make signoff` is running now; I'll re-add to the queue once it's green.

phillipleblanc (dismissed):

sgrebnov (dismissed):

phillipleblanc:
Removed from the merge queue: the merge-queue run for this PR (commit `fe404695`) failed `Integration Tests (part 1)` with reproducible failures in the schema-evolution suite:

- `schema_evolution::test_drop_recreate_full_csv_cayenne`
- `schema_evolution::test_schema_evolution_widening_cayenne_sync_all_columns`

Both fail the same way — a warning is logged first:
```
WARN cayenne::provider::table: Failed to deserialize serialized table statistics: Assertion failed error: array_stats.len() != struct_fields.nfields(): 4 != 5
```
followed by a query execution error:
```
ERROR ... Failed to execute query: Execution error: Failed to project inlined batch for table sample: Schema error: project index 4 out of bounds, max field 4
```

This looks like a stats/schema mismatch surfaced by the schema-evolution + whole-table-refresh path this PR changes (`publish_overwrite_snapshot_fenced`, `prepublish_inlined_overwrite`, and the new inline-row-count/stats republishing on overwrite). I isolated it to this PR specifically: the immediately preceding queue entry (base + the two PRs ahead of this one, no other cayenne changes) ran the same integration suite clean, and this commit's run (same base + this PR's diff) fails deterministically (reproduced 3x and 2x respectively within the same job).

Leaving the queue running rather than blocking on it — feel free to re-add once this is addressed. Full logs: the failed check run is linked from this PR's checks tab.

lukekim:
Fixed in b8216347fd — thanks for isolating it to this PR, that was exactly right.

**Root cause: an inline corpus that outlived a schema change.**

An inline entry is Arrow IPC frozen at the schema that was live when it was written. There are two places the table schema can be widened, and only one of them accounted for that:

- `evolve_schema_live` (stream-time) flushes the RAM tier *and* the durable inline corpus before it swaps, precisely because "inlined IPC batches decode with the stored schema and are unioned into scans projection-only (no cast)". Old Vortex files are fine — they're self-describing and the opener adapts them.
- `try_widening_schema_evolution` (open-time — the path that adopts a source's added column at restart) commits the evolved schema straight to the metastore and rewrites nothing.

So the reopened provider resolved live-schema column indices against a corpus still at its narrower write-time width. Hence `project index 4 out of bounds, max field 4`, and the companion `array_stats.len() != struct_fields.nfields(): 4 != 5` from the persisted stats blob.

**Why this PR made it reachable.** Both failing tests are `refresh_mode: full` on Cayenne file mode with a handful of rows. Before this PR that profile never inlined, so a table that could evolve at restart could never be holding a corpus — the two features had no overlap. Turning on inline admission for bulk-overwrite created it. The gap was already visible in the tree: `schema_evolution_restart_scans_old_files_under_widened_schema` explicitly zeroes the inline caps with the comment *"the adaptation under test … is the file read path, not the inline corpus."*

**Fix.** Adapt at decode rather than adding a second flush site — missing nullable columns null-filled, widened columns cast. That's the treatment the Vortex opener already gives pre-evolution files, and it's the truth for a row written before the widening, since `classify` admits only added *nullable* columns and lossless casts into a widening plan. Applied at both data-batch decodes: the one the full-rebuild and delta cache paths share, and the inline DELETE rewrite (a predicate naming an added column has the same problem). A corpus already at the live width short-circuits on a field comparison, so the CDC read path is untouched. `evolve_schema_live` still flushes before its own swap — this is the restart-path counterpart, not a replacement.

**Verification.**

- `schema_evolution::test_drop_recreate_full_csv_cayenne` — the test from your log — **passes locally**, all four phases, with zero occurrences of either the error or the stats warning.
- `test_schema_evolution_widening_cayenne_sync_all_columns` I could not run here (no Docker on this machine). It is the same mechanism with different widths — its log line is `project index 3 out of bounds` on `cham` — and the new unit test below covers exactly that 3→4 shape.
- New `schema_evolution_restart_scans_inlined_rows_under_widened_schema`: the inline-corpus counterpart to the file-backed test above, asserting the premise (rows really are in the corpus) so it can't silently degrade into a duplicate of it.

**Also in that queue run:** `Rust Lint` failed too, but on `make: cargo: No such file or directory` between the `--lib` and `--tests` clippy passes — a runner PATH flake, not code. Nothing to fix there.

Two small things I picked up while in here, both on this PR's own code: a failed inline-replacement insert reported as `FailedToSetCurrentSnapshot` when the snapshot update had already succeeded (now `InvalidOperation`), and the doc claim that a partitioned dataset "does not inline at all" — the coupled-writer refusal gates the *overwrite* path only, so children stay eligible for ordinary CDC inlining.

Branch is synced with `trunk`. `make signoff` is running; I'll re-add to the queue once it's green.

sgrebnov (approved):

PR #12369, opened 2026-08-02
fix(search): read the Elasticsearch delete response instead of assuming it applied (fixes #12364)

### LINKED ISSUE #12364: Elasticsearch delete ignores the _delete_by_query response, so a partially applied delete reports success
## Describe the bug

`delete_by_keys` in `crates/search/src/index/elasticsearch/delete.rs` discards the `_delete_by_query` response body:

```rust
client
    .delete_by_query(es_index, &or_of_row_clauses(chunk))
    .await
    .boxed()
```

A `2xx` from `_delete_by_query` does not mean every document was deleted. The response reports `deleted`, `version_conflicts`, `noops`, and a `failures` array, and Elasticsearch returns `200` with a populated `failures` array for per-document failures — the request succeeded, some deletes did not. `version_conflicts` is the common one: `_delete_by_query` snapshots the index at the start of the request and skips a document whose version changed since, which a concurrent write cycle over the same rows produces.

So a delete can leave documents behind and report `Ok(())`. That is the same observable symptom as #12267 and #12272 — search keeps returning rows the dataset no longer has — reached through a path those fixes do not touch.

## Expected behavior

A delete that Elasticsearch reports as partially applied is not reported as a success.

## Possible approach

Read the response: treat a non-empty `failures` array as an error naming the first few failures, and decide deliberately what `version_conflicts` should do — `conflicts=proceed` plus a retry is one option, surfacing it is another. Note that the callers of `Index::delete_by_keys` log its error rather than propagate it (`crates/runtime-datafusion-index/src/provider.rs`, `crates/runtime/src/accelerated_table/refresh_task/changes.rs`), so an error here is currently a loud log rather than a failed refresh — worth weighing when choosing between erroring and retrying.

`ElasticsearchIndex::write` should be checked for the same shape: `bulk_index` likewise returns `200` with per-item errors.

## Additional context

Raised by an adversarial review of the #12272 fix. Related: #12267 (exact-key delete addressing), #12272 (key-column mapping), #12363 (a Spice-owned base-key field).

claudespice:
## Summary

`delete_by_keys` discarded the `_delete_by_query` response body, so a request Elasticsearch answered `200` for was reported as a completed delete no matter what it actually removed.

A `2xx` there only means the request ran. The body is what says whether the documents are gone:

- **`failures`** carries per-document rejections — the request succeeded, those deletes did not.
- **`version_conflicts`** counts documents whose version moved between the snapshot the request scans and the delete itself, which a concurrent write cycle over the same rows produces. The default `conflicts=abort` stops the whole operation at the first one.
- **`canceled`** is set when the task was cancelled through the task API, which stops it short of the documents it had matched.
- **`timed_out`** marks a search phase that did not finish, so the scan never reached every matching document.

Each leaves documents the caller asked to remove still in the index while the delete reports success — the same observable symptom as #12267 and #12272 (search keeps returning rows the dataset no longer has), reached through a path those addressing fixes do not touch. The write path already inspects its `_bulk` response for exactly this shape (`inspect_bulk_response`); the delete path was the asymmetric sibling.

Fixes #12364

## Changes

- `ensure_delete_applied` reads each `_delete_by_query` response and errors on any of the four conditions above. A body carrying no `failures` array, or no `total`/`deleted` counts, is also an error: it does not report what was removed, so it is not evidence the delete applied — the same treatment the write path gives a `_bulk` response with no `errors` field.
- **The counters are the backstop.** After the named checks, `deleted < total` is an error on its own, so a partial delete whose cause this code has no check for still cannot report success.
- **Both addressing paths issue their requests through one `issue_chunked_deletes` helper**, so neither can grow a request whose response nobody reads — the defect being fixed here was one call site not checking.
- **A failure is reported as its status and cause type only.** Elasticsearch names the document by the `_id` the write path derives from the row's primary key, and the failure's free-form `reason` quotes it, so stringifying the entry wholesale would copy primary keys (an email address, an account id) into every log line a routine version conflict produces. Callers of `delete_by_keys` log this error, so it is operational output.

### Deliberate choices

- **The delete stops at the first request that did not apply**, rather than continuing and reporting at the end. Continuing would remove more documents but report the last slice's verdict for all of them; stopping matches the write path, and the error says the index still holds rows the dataset does not.
- **No retry and no `conflicts=proceed`.** Both are real options for the version-conflict case and both change what the delete does; this PR only stops a partial delete being reported as a complete one. `conflicts=proceed` in particular would skip the conflicted documents by design, which is the behaviour being reported as a bug.
- **This is a louder error, not a failed refresh.** Both callers of `Index::delete_by_keys` log the error and continue by design (`crates/runtime-datafusion-index/src/provider.rs`, `crates/runtime/src/accelerated_table/refresh_task/changes.rs`), because the authoritative row is already gone by then. The change turns a silent partial delete into an ERROR line naming the index.

### Base

Based on `fix/12272-elasticsearch-key-mapping` (#12366), not `trunk`. That PR splits `delete_by_keys` into the `_id` and term-filter paths this change instruments, and adds the `RecordingClient` mapping harness these tests extend; on `trunk` there is one call site and no term path, so a `trunk`-based fix would leave the second path unchecked and conflict on merge. `pull_request` workflows filter to `trunk`/`release-*`/`feature-*` bases, so CI beyond `enforce-pull-with-spice` does not run here — it runs once this retargets to `trunk` after #12366 merges.

## Test plan

- [x] `make lint`
- [x] `make nextest`
- [x] `cargo test -p search --lib --features elasticsearch index::elasticsearch` — 49 passing (43 in the delete module, 10 of them new; `elasticsearch` is not a default feature of `search`, so the feature is explicit)
- [x] Four deterministic neuters, each failing only the tests that pin it: dropping the response check → 10 fail; removing the counter backstop → 2 fail; removing the cancellation check → 1 fail; stringifying the whole failure entry → the key-redaction test fails
- [x] The mock now answers with the response shape Elasticsearch actually returns (`total`/`deleted`/`version_conflicts`/`failures`), so every pre-existing test exercises the new reader

## Checklist

- [x] Regression tests that fail without the fix
- [x] Error messages name the index and say what is still in it, with a docs link
- [x] No `.unwrap()`/`.expect()` in production code
- [x] `cargo fmt` + `cargo clippy -p search --lib/--tests --features elasticsearch` clean
- [x] Bug-class sweep: the two `_bulk` write sites already inspect their responses; no other caller in `crates/` issues `_delete_by_query`

## Deferred

- **#12370** — the `_bulk` write path reports a document failure by stringifying Elasticsearch's whole `error` object, whose `reason` quotes the document `_id` (the row's primary key). Same class as the redaction applied here, in a file this PR does not touch, so it is filed rather than folded in.

### DIFF
--- a/crates/search/src/index/elasticsearch/delete.rs
+++ b/crates/search/src/index/elasticsearch/delete.rs
@@ -465,13 +465,21 @@ fn ensure_delete_applied(es_index: &str, response: &Value) -> DataFusionResult<(
         ));
     }
 
-    if let Some(reason) = response.get("canceled").and_then(Value::as_str) {
-        return Err(delete_not_applied(
-            es_index,
-            &format!(
-                "the delete was cancelled before it finished ('{reason}'), so Elasticsearch stopped short of the documents it had matched"
-            ),
-        ));
+    if let Some(canceled) = response.get("canceled") {
+        if canceled.as_bool() == Some(true) {
+            return Err(delete_not_applied(
+                es_index,
+                "the delete was cancelled before it finished, so Elasticsearch stopped short of the documents it had matched",
+            ));
+        }
+        if let Some(reason) = canceled.as_str() {
+            return Err(delete_not_applied(
+                es_index,
+                &format!(
+                    "the delete was cancelled before it finished ('{reason}'), so Elasticsearch stopped short of the documents it had matched"
+                ),
+            ));
+        }
     }
 
     let version_conflicts = response
--- a/crates/search/src/index/elasticsearch/delete.rs
+++ b/crates/search/src/index/elasticsearch/delete.rs
@@ -107,15 +107,7 @@ pub async fn delete_by_keys(
     let term_fields = resolve_term_fields(client, es_index, key_columns).await?;
     let row_clauses = build_row_term_clauses(es_index, &term_fields, keys)?;
 
-    for chunk in row_clauses.chunks(DELETE_CHUNK_ROWS) {
-        client
-            .delete_by_query(es_index, &or_of_row_clauses(chunk))
-            .await
-            .boxed()
-            .map_err(DataFusionError::External)?;
-    }
-
-    Ok(())
+    issue_chunked_deletes(client, es_index, &row_clauses, or_of_row_clauses).await
 }
 
 /// Deletes the documents `keys` names by the `_id` the write path derived for each row.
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

Jeadie (approved):

Jeadie:
Closing as https://github.com/spiceai/spiceai/pull/12366 is closed and would need to be reworked.

Issue #11112, opened 2026-06-01
Test Models, AI & Search failing on trunk: search-metadata snapshots drifted by #11062 (Null→Int32 normalization)


claudespice:
## Summary

The `Test Models, AI & Search` integration job (`integration_models.yml`) is failing on **trunk** and on every PR that runs it. Multiple search-metadata snapshot assertions fail:

- `models::s3_vectors::search::metadata_columns` — snapshot `s3vectors_metadata_basic_response`
- `models::search::test_text_search_metadata` — snapshot `text_search_metadata_basic_response`
- `models::s3_vectors::search::with_chunking_metadata`
- `models::s3_vectors::search::with_chunking`

Non-snapshot search tests in the same suite (`test_text_search`, `test_rrf_search`, `openai_test_search`, etc.) pass, so this is isolated to the metadata-bearing snapshots.

## Likely cause

Bisecting trunk `integration_models.yml` runs, the last green run was commit `b7ba8c4f` (2026-06-01 05:12Z). The next commit, `3bf467c7` — **#11062 "Normalize `DataType::Null` to `Int32` in acceleration schema for duckdb"** — is where the job started failing; every trunk run since has been red.

That PR changed `crates/arrow_tools/src/type_rewrite.rs`, `crates/arrow_tools/src/schema.rs`, and the duckdb/sqlite accelerators, but updated no search/integration_models snapshots. The affected search tests run over accelerated tables, so normalizing `Null` → `Int32` changes the metadata-column schema/values that land in the search response snapshots.

It slipped through pre-merge because `Test Models, AI & Search` is secret-gated and does not run on every PR (and never on fork PRs).

## Suggested fix

Re-run the search metadata snapshots with `cargo insta test --review` against the normalized schema and confirm the new `Int32` metadata typing is the intended output, then commit the updated `*_metadata_basic_response` snapshots. If `Null` → `Int32` normalization should **not** apply to these search/metadata paths, scope the normalization to exclude them instead.

## Impact

Blocks merge readiness for any PR that triggers `integration_models.yml`, including the auto-generated snapshot-refresh PRs (e.g. #10873) and the develop→trunk merge PR (#11111).

claudespice:
> [!WARNING]
[...]

PR #12389, opened 2026-08-03
fix(search): preserve field names with capitals and '.'

Jeadie:
## Summary
Several places in search introduced bad `col(name)` that does not preserve capitalisation or periods correctly.


peasee (approved):

grokspice:
Cross-checking this against a bug-class sweep for the same `col(<name>)` → `Column::from_qualified_name` split: this PR covers `index/memory/provider.rs`, `index/elasticsearch/mod.rs` and `provider.rs`, and those were the sites causing the `models::s3_vectors::search` failures in the secret-gated models suite (`No field named message.body_embedding` / `No field named _spice.search_field` — see the re-triage on #11112).

Two files outside this PR have the same shape, both reached with index primary keys that can be `_spice.chunk_id`:

- `crates/runtime-datafusion-index/src/delete.rs:74` and `:117` — the projection and the equality predicate in `resolve_keys_matching_predicate` / `build_key_match_predicate`. The projection is taken off a plan scanned as relation `t`, so a dotted key resolves against a non-existent relation `_spice` and a delete on a chunked index resolves no rows.
- `crates/runtime-search/src/candidate/vector.rs:668` and `:686` — the two group-by lists in `search_late_interaction`, whose line 638 already uses `ident` for the projection of the same `pks`.

Filed as #12462 and fixed in #12464, which touches only those two files — no overlap with this PR (or with #12411 / #12410). Happy to fold it in here instead if you would rather it land as one change; just say so and I will close #12464.

PR #12431, opened 2026-08-04
fix(ci): stop the spiced child E2E cleanup was leaking, scoped to this runner (fixes #12058)

### LINKED ISSUE #12058: E2E Test CI: Dremio quickstart and the model jobs need a reliability pass (~15% of runs fail)

grokspice:
## Summary

`E2E Test CI` fails on roughly 1 in 7 runs, on a different sub-job each time, and each failure can strand an unrelated merge-queue entry. The randomness is because **the job that fails is not the job that caused it**.

Two facts combine:

1. **`spice run` is not the runtime.** The CLI spawns `spiced` as a child (`get_run_cmd` in [bin/spice/src/context.rs](bin/spice/src/context.rs) builds `Command::new(spiced_path)`), and `spiced` is what binds :8090 and :50051. Every cleanup in the workflow did `killall spice` / `pkill -x spice`. Both match exactly, so neither ever matched `spiced`: cleanup reaped the CLI and left the runtime holding the ports.

2. **The self-hosted runners are reused, and several instances share one machine.** A runner is named `<machine>-NN` and works out of `/opt/github-runner-NN/_work`, so `spiceai-macos-runner-04` hosts `-01`, `-02`, `-03`… concurrently, sharing one network namespace.

So an orphaned `spiced` is inherited by whatever runs next on that machine, and its `spice run` dies with:

```
ERROR spiced: Unable to start Spice Runtime servers: Unable to start HTTP server:
  Unable to bind to address: Address already in use (os error 48)
```

The runtime then never reports ready — and the readiness wait was
`while [[ "$(curl -s .../v1/ready)" != "ready" ]]; do sleep 1; done`, which prints nothing. So the entire diagnosis available to whoever triaged it was one line:

```
##[error]The action 'Wait for Spice runtime is ready' has timed out after 5 minutes.
```

That is the "no diagnosis" #12058 asked to fix before changing anything. Observed on run 30798700215 (`huggingface model (darwin-aarch64)`, machine `spiceai-macos-runner-04`).

## Changes

**`.github/actions/stop-spice`** (new) — stops the runtime this job started:

- signals **both** `spice` and `spiced`, SIGTERM then SIGKILL after a grace period;
- kills **by PID, filtered to processes whose cwd is under `$RUNNER_WORKSPACE`**. A name-wide `killall spiced` would have been worse than the leak — on a shared machine it would reap a *concurrent* job's runtime. When a PID's cwd cannot be determined it is left alone: leaking costs the next job a retry, killing a concurrent job destroys work in flight;
- waits for the ports, and distinguishes "free" from "held by someone who isn't us" — the latter being the shared-machine collision rather than a leak.

**`.github/actions/wait-for-spice-ready`** (new) — replaces the silent poll. Reports progress while waiting, and on timeout dumps the last response body, whether `spiced` is alive, what holds the port, and `spice.log`, calling out `Address already in use` explicitly with both of its causes.

**Applied across the class**, not just the reported job: 18 readiness waits and 11 cleanups in `e2e_test_ci.yml`, plus `.github/actions/check-spice-log`, which had the same `killall spice`. The two graceful-shutdown steps now use the shared stopper too — they were SIGTERMing the CLI, so the shutdown path under test never ran; they pass a grace longer than their step timeout so a runtime that does not exit cleanly still trips the timeout rather than being SIGKILLed into looking healthy.

## Test plan

- `scripts/test_e2e_spice_lifecycle.sh` — 49 unit tests, no processes started and no ports bound (stubbed `pgrep`/`kill`/`lsof`/`curl`). Run in CI by `.github/workflows/e2e_lifecycle_scripts_test.yml` on ubuntu-24.04 and macos-latest.
- Verified as genuine regression tests against both failure modes: reverting to `spice`-only fails 9 of them, and removing the workspace scoping fails 5 (including "the other instance survives intact").
- `make lint` and `make test` — no Rust, Cargo, or build inputs are touched, so this branch is outside `RUST_AFFECTING_PATH_PATTERN` and `Attestation` auto-passes.
- YAML validated for all changed workflow and action files.

## Not fixed here

The port collision *between concurrent runner instances on one machine* is real and separate: every E2E job hardcodes :8090, so two jobs on one machine contend by design regardless of leaks. That needs an infra decision (per-instance ports, or a concurrency group) rather than a workflow edit. This PR makes that case **self-identifying** instead of anonymous — `stop-spice` reports a port held by a PID that is not ours, and the readiness diagnosis names it.

Fixes #12058

### DIFF
--- a/.github/actions/check-spice-log/action.yml
+++ b/.github/actions/check-spice-log/action.yml
@@ -15,13 +15,21 @@ inputs:
 runs:
   using: 'composite'
   steps:
-    - name: Stop spice and check logs
+    # Stops the `spiced` child as well as the `spice` CLI. `killall spice` here
+    # used to leave the runtime holding :8090 for the next job on the same
+    # self-hosted runner (#12058).
+    - name: Stop spice
+      uses: ./.github/actions/stop-spice
+      with:
+        working-directory: ${{ inputs.working-directory }}
+        print-log: 'false'
+
+    - name: Check logs
       shell: bash
       working-directory: ${{ inputs.working-directory }}
       env:
         ALLOW: ${{ inputs.allow }}
       run: |
-        killall spice || true
         cat spice.log
 
         # `spiced` logs with_ansi(true), so an allow pattern such as
--- a/.github/actions/stop-spice/action.yml
+++ b/.github/actions/stop-spice/action.yml
@@ -1,0 +1,41 @@
+name: 'Stop Spice'
+description: |
+  Stops the Spice runtime started by `spice run` — both the `spice` CLI and the
+  `spiced` child that actually binds the ports — and waits for those ports to be
+  released, so a reused self-hosted runner does not hand an orphaned listener to
+  the next job (#12058).
+inputs:
+  working-directory:
+    description: 'Directory the runtime was started from (where spice.log lives)'
+    required: false
+    default: '.'
+  ports:
+    description: 'Space-separated TCP ports to wait for release'
+    required: false
+    default: '8090 50051'
+  print-log:
+    description: 'Print spice.log after stopping'
+    required: false
+    default: 'true'
+runs:
+  using: 'composite'
+  steps:
+    - name: Stop spice
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      env:
+        SPICE_STOP_PORTS: ${{ inputs.ports }}
+      # Invoked through `bash` so the step does not depend on the script's
+      # executable bit surviving a checkout on every platform.
+      run: bash "${GITHUB_ACTION_PATH}/stop_spice.sh"
+
+    - name: Print spice.log
+      if: ${{ inputs.print-log == 'true' }}
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      run: |
+        if [ -f spice.log ]; then
+          cat spice.log
+        else
+          echo "no spice.log in $(pwd)"
+        fi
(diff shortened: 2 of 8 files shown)


grokspice:
@copilot review

lukekim (approved):

PR #12435, opened 2026-08-04
feat(cayenne): drive incremental vacuum from the maintenance tick

lukekim:
## What was wrong

`SqliteAutoVacuum::Incremental` has been selectable since the mode was
introduced, and selecting it did nothing.

`PRAGMA auto_vacuum = INCREMENTAL` only *marks* freed pages as reclaimable;
reclaiming them requires an explicit `PRAGMA incremental_vacuum`, and nothing
ever issued one — the string appeared in a doc comment and nowhere else in the
tree. So the mode behaved like `None` plus per-commit bookkeeping, and an
operator who chose it to bound the metastore's footprint got a file that still
only ever plateaued. The mode's own doc says reclamation "must be driven off the
hot path, never inside a CDC burst"; this is that driver.

## Where it runs, and why there

On the background maintenance tick, on the same dedicated connection as the WAL
checkpoint and for the same reason: the pragma holds the write lock while it
relocates pages to the end of the file, so it must never run on the hot commit
path.

Ordering is load-bearing. In WAL mode the relocation is written as WAL frames
and the main database file only shrinks once a checkpoint copies them back, so
the vacuum runs immediately **before** the pass's checkpoint. After it, every
truncation would wait a whole tick.

## Keeping it free for everyone who doesn't want it

- **Gates on the database's ACTUAL `auto_vacuum` mode, not the configured one.**
  The pragma only takes effect on a fresh file, so a deployment that switched the
  config over an existing metastore is still in mode `NONE` — where
  `incremental_vacuum` is silently a no-op. Reading the real mode stops the driver
  burning a lock acquisition per tick forever on such a file.
- **Reads `freelist_count` first** and returns without taking the write lock when
  there is nothing to reclaim, which is the steady state once a freelist drains.
- **Bounded** by `cayenne_metastore_incremental_vacuum_pages` (default 256 = 1 MiB
  at the 4 KiB page size), so a large freelist drains over successive ticks
  instead of holding the write lock proportional to its whole length. `0` disables
  reclamation without changing the database's mode.

`None` remains the default and the recommendation — freed pages are reused, the
file plateaus, and the write path pays nothing. This is for deployments that would
rather have the space back.

## Tests

- `..._reclaims_freelist_and_shrinks_the_file` — the property that matters: under
  INCREMENTAL, after freeing a large run of pages, the vacuum reclaims them and the
  **main DB file shrinks** once the following checkpoint lands.
- `..._is_a_noop_in_the_default_mode` — `None` reclaims nothing, as documented.
- `..._gates_on_the_databases_actual_mode` — a file created `NONE` does not become
  reclaimable just because the config later says `INCREMENTAL`.
- `..._pages_zero_disables_the_driver`.

All 35 `metastore` tests pass.

## Observability

`cayenne_metastore_incremental_vacuum_ms` and
`cayenne_metastore_vacuum_pages_total`, emitted only for a pass that reclaimed
something, so the histogram describes real reclamation rather than being diluted
by no-op ticks.

## Compatibility

No schema change. One new runtime param; the default behaviour of every existing
deployment is unchanged, since they are all in `auto_vacuum: none`.


crates/cayenne/src/metastore/sqlite.rs
```diff
@@ -1542,6 +1560,73 @@ impl MetastoreBackend for SqliteMetastore {
...
+        //
+        // Ordering note for the caller: this belongs BEFORE the checkpoint in a
+        // maintenance pass. In WAL mode the relocation is written as WAL frames
+        // and the main DB file only shrinks when a checkpoint copies them back,
+        // so vacuuming after the checkpoint would defer the actual truncation by
+        // a whole tick.
+        let max_pages = sqlite_metastore_config().incremental_vacuum_pages;
+        if max_pages == 0 {
+            return Ok(0);
+        }
```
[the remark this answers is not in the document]
lukekim: Fixed — `incremental_vacuum` now returns early unless the configured `auto_vacuum` mode is `Incremental` (and `incremental_vacuum_pages > 0`). The existing gate on the DB's actual mode is unchanged, so a config flip over an existing NONE/FULL file still does not pretend reclamation is available.

crates/telemetry/src/lib.rs
```diff
@@ -1874,6 +1874,37 @@ pub mod cayenne {
             .record(duration.as_secs_f64() * 1000.0, dimensions);
     }
 
+    static METASTORE_INCREMENTAL_VACUUM_MS: OnceLock<Histogram<f64>> = OnceLock::new();
+    static METASTORE_VACUUM_PAGES: OnceLock<Counter<u64>> = OnceLock::new();
+
+    /// Records one off-hot-path incremental-vacuum pass over the metastore
+    /// freelist: how long it held the write lock, and how many pages it returned
+    /// to the filesystem. Only emitted for a pass that reclaimed something, so
```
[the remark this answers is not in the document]
lukekim: Fixed — the doc now describes the metric as the caller-supplied wall-clock duration of the pass, matching what is recorded.

crates/telemetry/src/lib.rs
```diff
@@ -1874,6 +1874,37 @@ pub mod cayenne {
...
+                    )
+                    .with_unit("ms")
+                    .with_boundaries(CONTENTION_MS_HISTOGRAM_BUCKETS.to_vec())
+                    .build()
+            })
+            .record(duration.as_secs_f64() * 1000.0, &[]);
+        METASTORE_VACUUM_PAGES
+            .get_or_init(|| {
+                operational_meter()
+                    .u64_counter("cayenne_metastore_vacuum_pages_total")
```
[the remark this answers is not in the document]
lukekim: Fixed — renamed the counter to `cayenne_metastore_incremental_vacuum_pages_total` so it pairs with `cayenne_metastore_incremental_vacuum_ms`.

crates/cayenne/src/metastore/sqlite.rs
```diff
@@ -1542,6 +1560,78 @@ impl MetastoreBackend for SqliteMetastore {
...
+        let max_pages = cfg.incremental_vacuum_pages;
+        let Some(pool) = self.pool.get() else {
+            return Ok(0);
+        };
+        let conn = &pool.checkpoint_conn;
+        let guard = conn.lock().await;
+
+        let start = std::time::Instant::now();
+        let reclaimed = guard
+            .call(move |conn| {
```
bjchambers: I feel like this should probably be in a separate method that gets `.call(helper_method)` rather than inline.

crates/cayenne/src/metastore/sqlite.rs
```diff
@@ -1542,6 +1560,78 @@ impl MetastoreBackend for SqliteMetastore {
...
+                // Gate on the DB's ACTUAL mode, not the configured one. The
+                // `auto_vacuum` pragma only takes effect on a fresh database, so
+                // a deployment that switched the config over an existing file is
+                // still in mode 0 — and `incremental_vacuum` there is silently a
+                // no-op. Reading the real mode keeps this from burning a lock
+                // acquisition per tick forever on such a database.
+                //
+                // 0 = NONE, 1 = FULL, 2 = INCREMENTAL. FULL already reclaims at
+                // commit time, so it needs nothing here either.
+                let mode: i64 = conn.query_row("PRAGMA auto_vacuum", [], |row| row.get(0))?;
```
bjchambers: Should we just remember whether this is enabled, rather than requerying it under the lock every time? Are we confident the query_row is fast enough to be called every time?

crates/cayenne/src/metastore/sqlite.rs
```diff
@@ -1542,6 +1560,78 @@ impl MetastoreBackend for SqliteMetastore {
...
+                    conn.query_row("PRAGMA freelist_count", [], |row| row.get(0))?;
+                if free_before <= 0 {
+                    return Ok(0);
+                }
+                // Bounded: reclaims at most `max_pages`, leaving the rest for
+                // later ticks rather than holding the write lock proportional to
+                // the whole freelist.
+                conn.execute_batch(&format!("PRAGMA incremental_vacuum({max_pages})"))?;
+                let free_after: i64 =
+                    conn.query_row("PRAGMA freelist_count", [], |row| row.get(0))?;
```
bjchambers: Makes sense. I've noticed that ArcSwap and `im` data structures often shift the cost to drop the `im` structure to the last thread that swaps it out. We could potentially have a similar mechanism for our own allocation -- put them in a list to be dropped in chunks by this background task. I suspect it's not a major performance problem at the current time (it shows up in some profiles and microbenchmarks at sf100, but doesn't seem to be the bottleneck for ingestion).

bjchambers (dismissed):

sgrebnov (dismissed):

crates/cayenne/src/metastore/sqlite.rs
lukekim: Fixed — freelist reclamation is now `reclaim_freelist_pages`, invoked via `.call(move |conn| reclaim_freelist_pages(conn, max_pages))`. The mode probe is a separate `read_auto_vacuum_mode` helper used the same way on the first tick.

crates/cayenne/src/metastore/sqlite.rs
lukekim: Fixed — the live `PRAGMA auto_vacuum` mode is probed once under the checkpoint lock and cached on the metastore (`OnceLock<bool>`). The mode is fixed at file creation, so after a non-INCREMENTAL result subsequent ticks return 0 without taking the lock at all. When INCREMENTAL, each tick only runs the freelist reclaim (no re-query of the mode). `freelist_count` still runs each tick because that value changes.

crates/cayenne/src/metastore/sqlite.rs
lukekim: Agreed — deferred bulk-drop of retired `ArcSwap`/`im` snapshots on this maintenance tick is a good follow-up if drop cost ever becomes a real bottleneck (sf100 profiles suggest it is not today). Out of scope for this PR, which only drives SQLite freelist reclamation. Happy to track it separately if useful.

sgrebnov (approved):

Issue #12442, opened 2026-08-04
Catalog connector load failures are reported twice (catalogs::LOAD_ERROR double-counted, status updated twice)


grokspice:
The catalog analogue of #12365, which tracks the same shape on the dataset path.

When `load_catalog_connector` fails with `UnknownCatalogConnector`, the failure is reported
**twice per attempt**:

- `load_catalog_connector` increments `metrics::catalogs::LOAD_ERROR` and calls
  `status.update_catalog(..., error_with_message(...))` on the unknown-provider branch, then
  returns the error;
- `load_catalog` receives it and does both again in its `Err` arm.

So one misconfigured catalog contributes 2 to `catalogs::LOAD_ERROR` per attempt, and the
component status is written twice with the same message.

### Expected behavior

Reporting is caller-owned: `load_catalog_connector` raises the error, and `load_catalog`
alone counts it and updates the status — matching the direction #12365 takes for datasets.

### Additional context

Found while fixing #12417. That fix stops the *repetition* — the failure is now classified
permanent, so the metric no longer climbs for the life of the process — but the intra-attempt
double count is a separate defect and was deliberately left out of scope there.

claudespice:
Sequencing note, from verifying this shape on the dataset side (#12365): **on trunk today the
double report here is not reachable, so this issue only becomes a live defect once #12469
lands.**

The reporting in `load_catalog_connector` sits in the `create_new_connector(...) == None` arm.
That arm cannot be reached, because the parameter builder resolves the factory from the *same*
registry first and fails through the `?` above it:

- `ConnectorParamsBuilder::build`'s catalog arm looks up `CATALOG_CONNECTOR_FACTORY_REGISTRY`
  and returns `DataConnectorError::InvalidConnectorType` when the provider is absent.
- [`load_catalog_connector`](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/init/catalog.rs#L102)
  applies that with `?` before ever calling `create_new_connector`.
- [`create_new_connector`](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/catalogconnector/mod.rs#L141)
  consults the same registry by the same name, so it returns `Some` whenever the builder
  succeeded.

Empirically, an unregistered provider yields `Failed to initialize catalog connector: ... The
connector '<name>' is not a valid connector.` — the `?` path — not `UnknownCatalogConnector`.
So an unknown catalog provider is reported **once** today, not twice.

[...]

Issue #12365, opened 2026-08-02
Dataset connector load failures are reported twice (LOAD_ERROR double-counted, status updated twice)


claudespice:
### Describe the bug

A dataset whose connector fails to load increments `metrics::datasets::LOAD_ERROR` **twice**, updates the component status twice, and emits two log lines for the single failure.

`Runtime::load_dataset_connector` already reports the failure on its error path — it updates the dataset status, increments `LOAD_ERROR`, and logs — before returning `Err`:

https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/init/dataset.rs#L348

Its caller then repeats all three on any `Err` it receives:

https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/init/dataset.rs#L488-L498

So one connector-creation failure produces `LOAD_ERROR += 2`, two `update_dataset(error)` calls, and two log lines.

### To Reproduce

Load a spicepod with a dataset whose connector cannot be created (for example an unparseable connector parameter), then read `LOAD_ERROR` — it advances by 2 per failing dataset rather than 1.

### Expected behavior

One failure reports once: a single `LOAD_ERROR` increment, a single status update, and a single log line at the appropriate level.

### Additional context

Both sites predate #12345 — that PR classifies permanent configuration failures and returns `PermanentDatasetFailure`, but leaves the reporting structure untouched, so the duplication is unchanged by it. Raised by review on that PR: https://github.com/spiceai/spiceai/pull/12345#discussion_r3699875593

The natural fix is to own the reporting in one place — either report only in `load_dataset_connector` and have the caller skip its block for errors that came from there, or strip the reporting out of `load_dataset_connector` and let the caller do it. Note that a permanent failure currently logs ERROR in the callee and WARN in the caller, so whichever site wins must preserve the ERROR level for permanent failures.

PR #12443, opened 2026-08-04
fix(runtime): stop retrying a permanent configuration failure on the catalog and dataset load paths (fixes #12417)

### LINKED ISSUE #12416: A dataset parameter configuration error is classified transient, so it retries forever at WARN
### Describe the bug

#12345 stopped the runtime retrying a dataset configuration failure no retry can clear, but
its classifier only recognises failures whose boxed source is a `DataConnectorError`:

https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/init/dataset.rs#L1816

Parameter validation raises a different type. `Parameters::try_new` rejects an out-of-vocabulary
`one_of` value and a missing required parameter with
`runtime_parameters::Error::InvalidConfigurationNoSource`, which
`get_dataconnector_from_dataset` boxes into `Error::UnableToInitializeDataConnector`. The
`downcast_ref::<DataConnectorError>()` then fails, so `is_permanent_dataset_failure` returns
`false` and the failure is treated as transient.

The consequences for a plain typo in a parameter value, or an omitted required parameter:

- `load_dataset` retries it with unbounded backoff for the life of the process, rebuilding
  the table provider on every attempt.
- Each attempt increments `datasets::LOAD_ERROR`.
- It is logged at `WARN`, so the one thing an operator has to fix does not appear at `ERROR`.

### Expected behavior

A configuration error that is a pure function of the Spicepod is classified permanent
whatever crate raised it: retried zero times, counted once, logged at `ERROR`.

### Additional context

Found by adversarial review of #12365 (which made the reporting caller-owned and therefore
made the level a function of this classification). Related permanent-but-unclassified
failures on the same load path: `AcceleratedTableInvalidChanges` and
`DurableWriteBackUnsupportedBySource` both return only a warning, with no metric and no
permanent wrapper.

### LINKED ISSUE #12417: Catalog loads retry a permanent configuration failure forever, counting catalogs::LOAD_ERROR on every attempt
### Describe the bug

`load_catalog` retries with `max_retries(None)` and classifies *every* connector-build failure
as transient, so a catalog configuration error no retry can clear is retried for the life of
the process:

https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/init/catalog.rs#L53
https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/init/catalog.rs#L66

Every failure `load_catalog_connector` can return arrives at that `RetryError::transient`:
an unregistered provider, a missing required parameter, an invalid parameter value. Each
attempt updates the component status, increments `catalogs::LOAD_ERROR` and logs a warning,
so a single misconfigured catalog drives the metric up indefinitely.

Only failures raised later, by `register_catalog`, are classified permanent — and only when
the source downcasts to a `catalogconnector::Error` that reports itself a configuration error:

https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/init/catalog.rs#L72

### Expected behavior

The catalog load path classifies a permanent configuration failure the way the dataset path
does after #12345: stop retrying, count it once, and log it at `ERROR` so an operator sees
that only a Spicepod change will clear it.

### Additional context

Found by adversarial review of #12365. This is the catalog analogue of #12345, which fixed
exactly this shape for datasets; the catalog path was not covered by that change. The dataset
counterpart for parameter-validation errors specifically is tracked separately.

grokspice:
Fixes #12416
Fixes #12417

## Why

Both component load paths retry with unbounded Fibonacci backoff and stop only when the
failure is classified permanent. A failure that is a pure function of the Spicepod — a
typo'd parameter value, a missing required parameter, a provider name this build does not
register — can never be cleared by retrying, so leaving it in the transient arm means the
runtime rebuilds it for the life of the process, increments `LOAD_ERROR` on **every**
attempt, and reports it at `WARN` when the one thing an operator must act on belongs at
`ERROR`.

Two gaps of the same shape, which is why they are one PR:

**Datasets (#12416).** #12345 classified connector-creation failures by downcasting the
boxed source to `DataConnectorError`. But `ConnectorParamsBuilder` validates parameters
*before* any connector exists, so an out-of-vocabulary `one_of` value or a missing required
parameter raises `runtime_parameters::Error` instead. The downcast fails, and a plain
Spicepod typo is retried forever.

**Catalogs (#12417).** Nothing classified `load_catalog_connector` at all — every failure it
can raise reached `RetryError::transient`, including `UnknownCatalogConnector`. Only
failures raised later by `register_catalog` were classified, and only for
`catalogconnector::Error`.

## Changes

- `is_permanent_catalog_failure` in `init/catalog.rs`, the catalog analogue of
  `is_permanent_dataset_failure`, applied to the previously unclassified
  `load_catalog_connector` arm.
- The existing inline `matches!` at the `register_catalog` site now routes through that same
  helper, so the two sites cannot drift apart on what "permanent" means. Behaviour there is
  unchanged for `catalogconnector::Error`; it additionally picks up parameter errors.
- Both paths recognise `runtime_parameters::Error::InvalidConfigurationNoSource`, matched by
  **variant name** rather than by type, so a future retriable variant cannot silently
  inherit "permanent" from that arm.

Anything unclassified stays retriable — a source that is merely unreachable still recovers on
its own, and a catalog registration failure (`UnableToLoadCatalogConnector`) is deliberately
left transient. Both directions are covered by tests.

## Scope

Deliberately not included, so this stays a classification fix:

- `metrics::catalogs::LOAD_ERROR` is incremented **twice per attempt** for
  `UnknownCatalogConnector` — once inside `load_catalog_connector` and again in
  `load_catalog` — and the component status is likewise written twice. This PR stops the
  *repetition* (the metric no longer climbs forever) but not the intra-attempt double count.
  That is the catalog analogue of #12365 and is filed separately as #12442.

## Test plan

- [x] Added regression test for a dataset parameter-validation failure being classified permanent (#12416)
- [x] Added regression tests for an unregistered catalog provider and a catalog parameter failure being classified permanent (#12417)
- [x] Added edge case tests for the fail-open direction: an unrecognised boxed error, an unreachable source, and a registration failure all stay retriable
- [x] Added a guard test that `catalogconnector::Error` configuration errors still classify permanent after the `matches!` → helper refactor
- [x] `make lint` passes (workspace-wide `cargo fmt --all -- --check` + full clippy, not a per-crate invocation)
- [x] `make test` passes

> **Verification note.** There is no Rust toolchain on the machine this was written on, so
> `make lint` / `make test` were **not** run locally — they run through **Remote Sign-off**,
> which executes the same `make lint-rust` + `make build-cli` + `nextest` gate and is what
> attests the head commit. I am flagging that explicitly rather than implying a local run.
> The two boxes above are checked on that basis; treat a red sign-off as the real verdict.

## Review gate

- `/security-review`: no findings. The diff adds no input parsing, SQL, filesystem, network,
  deserialization or auth surface. The only output change is severity-only — a message
  already logged at `WARN`, and already written to the component status, now logs at `ERROR`
  when the failure is permanent, with identical content.
- Cross-model adversarial review: **skipped** — the companion is a Node script and this
  machine has no Node.js runtime.


grokspice:
@copilot review

lukekim (approved):

Issue #12428, opened 2026-08-03
A cancelled sign-off can revoke a signoff=success that had already been posted


grokspice:
`.github/workflows/signoff.yml` gained a `Correct the sign-off status when this run is cancelled` step in #12425, which rewrites `signoff` to `pending` whenever the job ends in `cancelled()`. That is right for the case it was written for — a run killed by `cancel-in-progress` whose dying `run_checks` posted `signoff=failure` — but it writes unconditionally, and cancellation is not always that case.

If the job is cancelled *after* the `Sign off` step already posted `signoff=success` (killed during a later step, e.g. `Show sccache stats`), the handler overwrites a sign-off that genuinely completed. The commit's Attestation then goes red and the branch is pushed back through a 1–4 hour re-dispatch for checks that already passed. The same applies if another run has posted `success` for the same SHA by the time the cancelled run tears down.

### Expected

A cancelled run corrects only the bogus `failure`/`error` its own death wrote, and leaves `success` alone.

### Fix

Read the current status before writing, and rewrite nothing but `failure`/`error`:

```bash
current="$(gh api "repos/${REPO}/commits/${sha}/status" \
  --jq 'first(.statuses[]? | select(.context=="signoff") | .state) // ""' 2>/dev/null)" || current=""

case "${current}" in
  failure | error) ;;
  *) echo "signoff on ${sha} is '${current:-unset}', not a failure this run posted — leaving it alone."; exit 0 ;;
esac
```

Reading it as combined-status-first-entry-for-context matches how `scripts/signoff` already reads the same status, so the workflow and the script agree on the current state.

Raised by the Copilot review on #12425; that PR was already approved and in the merge queue when it was confirmed, so it is split out here rather than dequeuing an approved PR.

PR #12448, opened 2026-08-04
fix(runtime): fail the write when an index cannot prepare its write window (fixes #12421)

### LINKED ISSUE #12421: A failed Index::on_write_start does not stop the write, so an index that could not prepare its window is written to anyway
## Summary

Both sinks treat a failed `Index::on_write_start` as advisory: they log a warning and run the write anyway.

`crates/runtime/src/accelerated_table/sink/table.rs`:

```rust
if let Err(e) = index.on_write_start(write_window).await {
    tracing::warn!(
        "TableSink: on_write_start failed for index '{}': {e}. Continuing with write.",
        index.name()
    );
}
```

`crates/runtime/src/accelerated_table/sink/multi.rs` does the same.

## Impact

The hook is where an index prepares the state the write depends on, so a failure there is not cosmetic:

- The full-text index rolls its writer back in `on_write_start` to discard operations a previously abandoned window left staged. If that rollback fails the sink still writes, and those stale operations are swept into this window's commit.
- Since #12066 the same hook stages the `ReplaceAll` clear. A failure means the clear did not happen, so the refresh proceeds as an append and leaves documents for rows the source dropped — the exact defect #12066 fixes, minus the error the user could act on.
- The accelerated table is written either way, so the row data and the index state diverge with only a `WARN` to say so.

The corresponding *completion* hook already models this properly: `Index::write_complete_failure_is_fatal` lets an index declare that a failed finalize must fail the write (#12038). The start hook has no equivalent.

## Suggested fix

Mirror the finalize design: give the trait a way to say a failed start is fatal (either a `write_start_failure_is_fatal` flag or an error classification), and in both sinks return before `collect()` when a fatal start fails — calling `on_write_failed` only on the indexes whose start actually succeeded, since an implementation whose start failed partway owns its own cleanup (the convention `compound_on_write_start` already follows).

Defaulting the flag to `false` keeps today's best-effort behavior for indexes where a failed prepare really is advisory (the Elasticsearch `refresh_interval` tuning is a genuine example — the write is still correct without it).

grokspice:
## Summary

Both acceleration sinks treated a failed `Index::on_write_start` as advisory — they logged a warning and ran the write anyway. That hook is where an index prepares the state the write depends on, so the accelerated table was written while the index could not reflect it: row data and index state diverge, with only a `WARN` to say so.

`FullTextDatabaseIndex::on_write_start` is the live case. It rolls the tantivy writer back to discard operations an earlier abandoned window left staged, and its own comment states that a rollback failure is fatal for the window, because those stale operations would otherwise be swept into this window's commit. It returns the error to say so — and the sink dropped it.

The completion hook already models this correctly: `Index::write_complete_failure_is_fatal` lets an index declare that a failed finalize must fail the write (#12038). The start hook had no equivalent.

## Changes

- `Index::write_start_failure_is_fatal()` on the trait, defaulting to `false`, so every index that is best-effort today keeps exactly today's behavior.
- `prepare_indexes` in `accelerated_table/sink/mod.rs`, mirroring `finalize_indexes`. Both sinks route through it and return before any data is written when a fatal start fails. `TableSink` returns before `collect()`; `MultiSink` returns before any task is spawned.
- On a fatal failure, only the indexes that actually started are rolled back via `on_write_failed`. The index whose own start failed owns its cleanup, and the ones after it were never started — the convention `compound_on_write_start` already follows.
- `rollback_indexes` extracted alongside it; the two existing open-coded `on_write_failed` loops in the sinks now call it.
- `FullTextDatabaseIndex` declares its start fatal. Both chunked wrappers and both compound wrappers forward the flag (a wrapper inheriting the default silently downgrades a fatal inner index). The Elasticsearch indexes keep the default — their start only tunes `refresh_interval` and the write is correct without it.

### One adjacent fix, called out deliberately

`ElasticsearchIndexWriteMaintenance::on_write_start` sets `write_cycle_active` before its fallible calls but only cleared `refresh_interval_overridden` on error. That was invisible while every start failure was still followed by a write and an `on_write_complete` to clear it. Now that a start failure can abandon the write, a left-open cycle would make every later write's start short-circuit as a second-batch no-op and silently drop the tuning — so its failure paths now clear both flags. It is in this PR rather than a follow-up because it is the hole this change would otherwise open.

## Performance impact

None. The added trait method is a `bool` returned once per write, and the write path is otherwise unchanged: same hooks, same order, same number of calls.

## Test plan

- `make lint`
- `make test`

New tests:
- `sink/mod.rs` — `prepare_indexes`: every index starts when none fail; a best-effort failure leaves the write successful and still starts the rest; **a fatal failure fails the write** (regression test — fails on old code, which had no error path at all); the error names the index and survives `find_datafusion_root`; a fatal failure rolls back exactly the started indexes and never starts the ones after it; an empty set succeeds.
- `chunking.rs` / `compound/tests.rs` — the new flag forwards through both chunked wrappers and is the union of both compound halves, and asserts it is **not** conflated with the finalize flag.
- `text_search/index.rs` — the full-text index declares its start fatal.
- `elasticsearch/mod.rs` — a failed start leaves no write cycle open, so the next write still applies the override.

## Checklist

- [x] Behavior is opt-in: the trait default preserves today's best-effort handling
- [x] Regression test fails on the old code
- [x] Every wrapper `Index` impl forwards the new flag
- [x] No `.unwrap()` added

Fixes #12421


grokspice:
@copilot review

grokspice:
Status-history note, because this PR's `signoff` history reads worse than the code deserves. All four entries below are on the **same** commit, `b1fc7141e` (pushed 07:12; nothing has been pushed since):

```
12:32:49 success — Signed off by grokspice via remote — lint+test passed in 18902s
13:40:42 failure — Sign-off checks failed after 12444s
13:44:09 pending — Remote sign-off in progress
14:17:33 pending — Remote sign-off in progress
```

So this commit **has** passed a full remote lint+test (5h15m) once already. A later run then overwrote that `success` with a `failure`, and two further dispatches — [30915240698](https://github.com/spiceai/spiceai/actions/runs/30915240698) (branch-keyed, 13:43) and [30917976000](https://github.com/spiceai/spiceai/actions/runs/30917976000) (`-f pr_number=12448`, 14:15) — are still in flight against it, in different concurrency groups because they used different input forms.

I have not re-dispatched, and no one should: a new dispatch would evict one of the two runs already working on this commit.

Related, since it is the same machinery this PR's own fix lives near:

- The overwrite of a genuine `success` is the #12428 hazard. My #12432 narrows it, but only for *cancelled* runs — a run that reaches a real failure verdict on a commit another run already signed off still clobbers it.
- The duplicate-dispatch grouping is #12472, fixed by #12474 (approved). Today it cost PR #12469 a 5.9-hour run that died at the `timeout-minutes: 353` ceiling; evidence is on #12474.

Nothing here suggests a defect in this diff. If the in-flight runs land green, this is just noise; if one fails, the 12:32 success is the reason to suspect infrastructure before the code.

lukekim (approved):

grokspice:
Sign-off run 30942941645 posted `signoff=failure`, but nothing in this PR failed: the job hit its 353-minute budget (`The action 'Sign off' has timed out after 353 minutes`) with the workspace tests still passing — the log shows PASS up to 1452/1858 and contains no compile or test failure. The "Sign-off checks failed after 21195s" description is the budget, not a verdict.

Re-dispatched sign-off as run 30967077807. Filed #12518 for the reporting gap: neither `correct-cancelled` (gated on `cancelled()`, and a timeout is a *failure*) nor `clear-pending` (which deliberately leaves a `failure` alone) repairs a budget expiry.

grokspice:
The re-dispatched sign-off (run [30967077807](https://github.com/spiceai/spiceai/actions/runs/30967077807)) hit the same 353-minute budget, so this PR now has two consecutive timeouts and I am stopping rather than dispatching a third ~6-hour run.

Neither failure is about the code in this PR:

- Run 30942941645: reached the test phase, `PASS` up to 1452/1858, no compile error and no failing test anywhere in the 1.1 MB log — then the budget expired.
- Run 30967077807: **zero** `PASS` lines. It spent all 353 minutes in the build phase and never started testing, i.e. it got *less* far than the first attempt.

The status text "Sign-off checks failed after 21201s" is the budget restated, not a verdict — tracked as #12518, which neither `correct-cancelled` (gated on `cancelled()`) nor `clear-pending` (which deliberately leaves a `failure` alone) repairs.

The pool had 9+ concurrent sign-off runs across several actors while these ran, and a runtime-touching change needs the full workspace build plus `nextest`. Retrying made it worse, so this looks like capacity rather than anything actionable here.

What a human can decide:

- Re-dispatch when the self-hosted pool is quiet (`gh workflow run signoff.yml -f pr_number=12448`), or
- raise the sign-off budget / improve build-cache reuse per #12518 first, so the retry can actually finish.

The PR itself is approved with no unresolved review threads; the only thing missing is a `signoff=success` that fits in the budget.

PR #12488, opened 2026-08-04
fix(tests): assert the search cache status instead of wall-clock timings (fixes #12487)

### LINKED ISSUE #12487: Search cache tests assert on wall-clock timing: one is flaky, and the bypass test can never fail
# Problem

The two search-results-cache tests in `crates/runtime/tests/models/search.rs` both assert on wall-clock durations. One is flaky as a result, and the other cannot fail at all — so the cache-bypass path is effectively untested.

## 1. `test_search_with_cache` is flaky (wall-clock ratio)

`search.rs:2186` compares a **single, cold** request against the median of 10 warm cached requests and requires a 10x speedup:

```rust
assert!(duration_cached * 10 < duration, "Cache did not improve performance by an order of magnitude. First: {duration:?}, Second: {duration_cached:?}");
```

`duration` is the very first request the process makes, so it carries one-time costs that have nothing to do with the results cache: TCP connect, embedding-model warm-up, and the first scan of the vector index. The assertion therefore passes only when that one cold sample happens to land ≥10x above the cached median.

It failed in the nightly trunk run [30865794719](https://github.com/spiceai/spiceai/actions/runs/30865794719) (2026-08-04):

```
thread 'models::search::test_search_with_cache' panicked at crates/runtime/tests/models/search.rs:2186:13:
Cache did not improve performance by an order of magnitude. First: 204.048541ms, Second: 56.111958ms
```

It passed on retry, so it is the `1 flaky` in that run's `576 tests run: 573 passed (1 flaky), 3 failed` summary. Nothing about the cache changed between the two attempts — only how loaded the runner was.

## 2. `test_search_with_cache_bypass` can never fail

`search.rs:2249` is a tautology:

```rust
assert!(duration >= duration_cached*0.7 || duration <= duration_cached*1.3,
    "Cache bypass did not return similar performance. First: {duration:?}, Second: {duration_cached:?}");
```

For this to fail, both disjuncts must be false, i.e. `duration < 0.7 * duration_cached` **and** `duration > 1.3 * duration_cached`. Since `duration_cached > 0` and `0.7 * duration_cached < 1.3 * duration_cached`, no pair of values satisfies both. The assertion holds for every possible input.

The intent reads as "the two requests took a similar time" (a `&&` of a ±30% band), but as written the test asserts nothing, so a regression that started serving `Cache-Control: no-cache` requests out of the cache would not be caught here.

# Root cause

Both tests infer cache behaviour from timing, even though the runtime already reports the disposition directly. `/v1/search` sets a `Search-Results-Cache-Status` response header (`crates/runtime/src/http/v1/search.rs:164`) from `CacheStatus::to_header_string()` (`crates/cache/src/result/mod.rs:37`), whose values are `HIT`, `MISS`, `BYPASS`, `STALE`. The tests discard response headers because their helper (`run_search_test` → `http_post`) returns only the body.

# Proposed solution

Assert on `Search-Results-Cache-Status` instead of on elapsed time:

- `test_search_with_cache` — first request reports `MISS`, repeats report `HIT`.
- `test_search_with_cache_bypass` — populate the cache with a normal request, then assert a request carrying `Cache-Control: no-cache` still reports `BYPASS`, which is the property the tautology was reaching for and is strictly stronger than a timing band.

Note that the bypass path deliberately still *writes* to the cache (`CacheControl::NoCache` calls `wrap_cache_to_result` just like the miss arm, `crates/runtime-search/src/search_engine.rs:447`), so "bypass ⇒ a later normal request misses" would be an incorrect assertion. Bypass means "never served from cache", not "does not populate it".

This makes both tests deterministic and independent of runner load, and keeps the existing response-body snapshot assertions as-is.

# Scope and non-goals

- Test-only change. No runtime, cache, or search behaviour changes.
- The same single-sample `duration_cached * 10 < duration` shape also appears in three chat/embedding cache tests — `crates/runtime/tests/models/hf.rs:419`, `local.rs:143`, and `openai.rs:442`. Those endpoints emit no cache-status header, so they need a different remedy and are deliberately left out of scope here.

grokspice:
Fixes #12487

## Summary

Both search results-cache integration tests inferred cache behaviour from wall-clock timings. One of them flaked in nightly CI; the other could never fail.

**Root cause.** `/v1/search` already reports the cache disposition on every response — `Search-Results-Cache-Status`, set from `CacheStatus::to_header_string()` with values `MISS`, `HIT`, `BYPASS`, `STALE` — but the tests discarded response headers and timed the requests instead.

- `test_search_with_cache` compared a **single cold** request against the median of 10 warm cached requests and demanded a 10x speedup. That first request carries one-time costs unrelated to the results cache (TCP connect, embedding-model warm-up, first scan of the vector index), so the assertion passed only when that one sample happened to land 10x above the cached median. It failed in nightly run [30865794719](https://github.com/spiceai/spiceai/actions/runs/30865794719) at `First: 204.048541ms, Second: 56.111958ms` and passed on retry — the `1 flaky` in that run's `576 tests run: 573 passed (1 flaky), 3 failed` summary.
- `test_search_with_cache_bypass` asserted `duration >= duration_cached*0.7 || duration <= duration_cached*1.3`. For that to fail, `duration` would have to be both below `0.7 * duration_cached` and above `1.3 * duration_cached`, so it holds for every possible input. The bypass path had no effective coverage at all.

## Changes

- Add `search_cache_status`, a `/v1/search` POST that returns the `Search-Results-Cache-Status` value, so the cache tests can assert on what the runtime reports rather than on how loaded the runner is.
- Add `search_until_cache_hit`, a **deadlined** poll (`CACHE_HIT_WAIT`, 5s, 100ms interval) for that `HIT`. This is load-bearing rather than defensive: the cache write is detached. `wrap_cache_to_result` hands it to a `tokio::spawn` — deliberately, since awaiting it inline would deadlock against the response stream it has to drain first — so `put_raw_key` can land *after* the caller already holds its response. Holding a response is not a happens-before for the insert, so asserting `HIT` on the immediately-following request would have been its own race, just a different one from the timing flake. It mirrors the existing `utils::wait_until_true` idiom, but keeps the last status seen and propagates request errors rather than collapsing both into `false` — so a cache that never populates fails on a real status rather than retrying forever.
- `test_search_with_cache` now asserts that repeating a request already in the cache reports `HIT`.
- `test_search_with_cache_bypass` now warms the cache with an ordinary request, confirms the entry is live (`HIT`), and then asserts that the same request carrying `Cache-Control: no-cache` still reports `BYPASS` — the property the tautology was reaching for, and one that only means something once there is a warm entry the bypass could wrongly be served from.
- Raise that test's `item_ttl` from 10s to 60s, comfortably clear of `CACHE_HIT_WAIT`. It tests the bypass, not expiry, so an entry ageing out mid-test is purely a flake source. No assertion depends on expiry.
- `std::time::Instant` is now used for the poll deadline rather than for asserting on elapsed time.

Both tests keep every existing `insta` response-body snapshot, and all four snapshot names are unchanged, so the cached and bypassed response bodies are still compared against the uncached one.

Note that `CacheControl::NoCache` deliberately still *writes* to the cache (`search_with_cache` calls `wrap_cache_to_result` on the bypass arm exactly as on the miss arm), so "bypass implies a later ordinary request misses" would be an incorrect assertion. Bypass means "never served from the cache", not "not stored in it"; the test asserts only the former.

### On what the timing assertion was worth

The 10x margin is not preserved, because it was measuring cold-start cost rather than the cache. A `HIT` is the property the test was reaching for and it is exact, so the replacement is deterministic where the old one was probabilistic. What is no longer covered is "a cache hit is also *fast*" — a hit that became slow would now pass. That is a throughput question rather than a correctness one, and it belongs in a benchmark, not in a secret-gated integration test that retries on failure.

## Test plan

- `make lint`
- `make nextest`
- The changed tests live in the secret-gated `integration_models` suite, which needs live HuggingFace/S3 credentials and so runs in the `Test Models, AI & Search` workflow rather than on this PR. `make lint` still compiles and clippies that test target (its `--tests` pass enables the `models` feature these tests are gated behind), so the change is build- and lint-verified here; the behavioural confirmation is the next nightly run of that workflow.

## Checklist

- [x] Test-only change — no runtime, cache, or search behaviour is modified
- [x] All four existing `insta` snapshots retained, with unchanged names, and none regenerated
- [x] No new timing- or load-dependent assertions
- [x] The bypass assertion matches the implementation's actual semantics (bypass reads through, still writes)
- [x] The detached cache write is accounted for, so the new assertions are not themselves racy

## Known, bounded gap

The `HIT` is asserted on a different request from the one whose body is snapshotted, so a *transient* bad hit could slip between them. Coupling the two would mean returning the header from `run_search_test`, which every other call site in this file also uses; that seemed worse than the gap. Persistent bad cached data still fails the retained `with_cache_post_cache` snapshot, and the snapshot request runs milliseconds after a confirmed live entry under a 30s TTL.

Separately, a `HIT` that became *slow* would now pass. That contract is not preserved, for the reason in the section above.

The poll is bounded by a deadline it checks between attempts, not by a per-request timeout: `Client::new()` sets none, so a request that stalls indefinitely would still stall the test. That is true of every other HTTP call in this suite and is left alone rather than fixed here.

## Review notes

An adversarial pass on the first draft caught a real defect: that draft asserted `HIT` on the request immediately following the cache-populating one, which races the detached write described above. That is what the deadlined poll exists for. The same pass independently confirmed the two claims this change rests on — that the old bypass assertion is unsatisfiable, and that `CacheControl::NoCache` still writes, so the tempting "bypass implies a later miss" assertion would have been wrong.

## Out of scope

The same single-sample `duration_cached * 10 < duration` shape also appears in three chat/embedding cache tests — `crates/runtime/tests/models/hf.rs:419`, `local.rs:143`, `openai.rs:442`. Those endpoints emit no cache-status header, so they need a different remedy; recorded in #12487's non-goals rather than fixed here.


grokspice:
@copilot review

lukekim (approved):

Issue #12499, opened 2026-08-04
Append dedupe is set subtraction, not multiset: legitimate duplicate rows are dropped, and the comparison set is unbounded


grokspice:
# Problem

The append-refresh dedupe in `except_existing_records_from` / `filter_records` (`crates/runtime/src/accelerated_table/refresh_task.rs`) has two defects that are independent of any one time-column type:

1. **It is set subtraction, not multiset subtraction.** `filter_records` drops an incoming row if *any* existing row compares `Equal`, so it imposes cross-refresh `DISTINCT` semantics on a table that has bag semantics unless a key says otherwise. A legitimate second identical row is silently dropped.

2. **The comparison set is fully materialised and compared row-by-row.** The accelerator side is `collect()`ed into memory with every column projected, then every incoming row is compared against every existing row.

Neither is introduced by #12492 — both already apply to any dataset with a nonzero `append_overlap`, which is what makes the comparison set non-empty today. #12492 makes the path unconditional for a `Date32` time column, which is what motivates filing this.

# 1. Multiplicity is lost

- Accelerator holds `(day=D, id=1, payload="x")` once.
- The source legitimately grows to hold that same row twice.
- The refresh re-fetches both copies.
- Both compare `Equal` to the single existing row, so **both** are dropped. The new occurrence is lost.

The same false positive happens through a `refresh_sql` projection, where the distinguishing column never reaches the comparison:

- Existing source row `(D, 1, hidden="A")`, newly appended row `(D, 1, hidden="B")`.
- The accelerator materialises only `SELECT day, id`.
- Both project to `(D, 1)`, so the new row is dropped.

(If the update stream contains a column *absent* from the accelerator, `filter_records` fails loudly at `index_of` instead — the silent case is specifically a column projected out before the update stream is built.)

# 2. Memory and comparison cost

The dedupe query is collected in full, and `filter_records` is a nested loop over (incoming × existing). For a dataset appending ~100M rows/day, the boundary window is:

- ~800 MB of value buffers for even a `Date32 + Int32` schema, and ~10 GB at 100 bytes/row, before execution overhead.
- ~100M rows read from the source plus ~100M from the accelerator per refresh.
- A worst-case comparison count that is quadratic in the window size.

This is an availability risk on high-volume append datasets, and it scales with the window rather than with the number of new rows.

# Suggested direction

[...]

PR #12500, opened 2026-08-04
fix(runtime): compare an append high-water mark inclusively on a day-granular time column (fixes #12492)

### LINKED ISSUE #12492: append refresh stalls on a Date32/Date64 time column: day-granular high-water mark with a strict Gt filter skips same-day rows
# Problem

`refresh_mode: append` stalls permanently on any dataset whose `time_column` is a **day-granular Arrow type** (`Date32` or `Date64`): rows arriving for the latest already-loaded day are never picked up, and are lost for good once the source moves on to the next day.

This is the connector-independent core of #8596. That report's Oracle repro was caused by Oracle `DATE` being mapped to `Date32` and is fixed by #12097 (merged 2026-07-30). The defect below is **independent of Oracle** and still live on `trunk` for any genuinely `Date32`/`Date64` time column — it has never had its own issue. Verified by code reading on `trunk` @ `96a0f891d`; **not reproduced** (needs an append dataset with a `Date32` time column).

# Root cause

Append refresh takes its high-water mark from the *accelerator* and filters the *source* strictly greater than it:

- `timestamp_nanos_for_append_query` (`crates/runtime/src/accelerated_table/refresh_task.rs`) computes `max(time_column)` via `max_timestamp_df`, which `CAST`s to `Timestamp(ns)`.
- `get_full_or_incremental_append_update` pushes the filter at [`refresh_task.rs:1141`](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/accelerated_table/refresh_task.rs#L1141):

```rust
filters.push(converter.convert(timestamp, Operator::Gt));
```

For a `Date32` column this comparison is day-granular. `data_type_to_timestamp_format` maps `DataType::Date32 => TimestampFormat::Date` ([`crates/util/src/timestamp_filter.rs:127`](https://github.com/spiceai/spiceai/blob/trunk/crates/util/src/timestamp_filter.rs#L127)), and `convert_timestamp_expr` emits `CAST(col AS Timestamp(ns)) > lit(nanos)` ([`timestamp_filter.rs:71`](https://github.com/spiceai/spiceai/blob/trunk/crates/util/src/timestamp_filter.rs#L71)).

Every `Date32` value casts to **midnight**. So once any row for day *D* is loaded, the high-water mark is *D*-midnight, and every other row of day *D* also casts to *D*-midnight — `> D-midnight` excludes all of them. They are skipped on every subsequent refresh, and permanently lost once the source advances past *D*.

`refresh_sql` is **not** involved: `get_data` (`crates/runtime/src/dataconnector/mod.rs`) ANDs the filter onto the parsed plan correctly even with a custom refresh SQL. Please don't re-chase that.

## `Date64` has the same defect but not the same format

`data_type_to_timestamp_format` sends `Date64 | Time32 | Time64 => TimestampFormat::Timestamp` ([`timestamp_filter.rs:122-123`](https://github.com/spiceai/spiceai/blob/trunk/crates/util/src/timestamp_filter.rs#L122-L123)), and only `Date32` to `TimestampFormat::Date`. But Arrow requires `Date64` values to be exact multiples of `86_400_000` ms, so a `Date64` time column is *also* midnight-aligned and *also* stalls — while being indistinguishable from a genuinely sub-day `Time32`/`Time64` column by `TimestampFormat` alone.

# Proposed solution, and the trap in it

Use `GtEq` for day-granular time columns, relying on the existing exact-row `StructArray` dedupe in `except_existing_records_from` to drop the rows that come back.

Two constraints on the implementation:

1. **Decide the operator from the Arrow `DataType`, not from `TimestampFormat`.** Widening the `TimestampFormat::Date` arm misses `Date64`, and it cannot be extended to cover `Date64` without also loosening `Time32`/`Time64`, which are legitimately sub-day. Day-granular ⇒ `GtEq`, threaded to the call sites.

2. **Widen both `Gt` sites in the same commit.** The dedupe's own query also uses `Operator::Gt`, at [`refresh_task.rs:1892`](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/accelerated_table/refresh_task.rs#L1892):

```rust
.filter(filter_converter.convert(value, Operator::Gt))
```

Widening only line 1141 means the already-loaded same-day rows never enter the comparison set, so the re-fetched rows pass dedupe and are **inserted again**. That turns a stall into silent duplication, which is worse than the bug.

Note that the third `Gt` in that file (line 1144) is the `refresh_data_window` wall-clock window, not a high-water mark, and should be left alone.

# Impact

Silent data loss on append refresh for any `Date32`/`Date64` time column, on every connector and accelerator — the filter is built in the shared append path. Severity is the loss, not the stall: rows skipped for day *D* are never revisited.

# Scope and non-goals

- Does not touch `refresh_sql` handling, `on_conflict` (the #8596 title blames it; nothing in the append path reads it), or the Oracle type mapping already fixed by #12097.
- Sub-day `Time32`/`Time64` and `Timestamp(_, tz)` columns must keep strict `Gt` — they have the precision to make it correct.
- The blast radius is the shared append path for every connector, and the dedupe is the only thing that makes `GtEq` safe, so this wants integration coverage on a real append dataset rather than unit tests alone.

grokspice:
## Summary

`refresh_mode: append` silently loses data on any dataset whose `time_column` is a day-granular Arrow type. Rows arriving for the latest already-loaded day are never picked up, and are gone for good once the source advances to the next day.

**Root cause.** The append high-water mark is read back out of the *accelerator* — `timestamp_nanos_for_append_query` takes `max(time_column)` via `max_timestamp_df`, which `CAST`s to `Timestamp(ns)` — and the *source* is then filtered strictly greater than it. Every `Date32` value casts to **midnight**, so once any row of day *D* is loaded the mark is *D*-midnight, and every other row of day *D* casts to that same instant. `> D-midnight` excludes all of them, on that refresh and on every one after it.

The fix compares inclusively on a day-granular column and lets the existing exact-row dedupe in `except_existing_records_from` drop the rows that come back.

**Both high-water sites move together.** The dedupe builds its own comparison set with the same filter (`refresh_task.rs`, the `.filter(...)` in `except_existing_records_from`). Widening only the source side would leave the already-loaded same-day rows out of that set, so the re-fetched rows would pass the dedupe and be **appended a second time** — a stall turned into silent duplication, which is worse than the bug. Both are widened in this commit; a comment at each site points at the other.

The third `Gt` in that file is the `refresh_data_window` wall-clock window, not a high-water mark, and is deliberately left strict — it still goes through the unchanged `convert()`.

## Changes

- `crates/util/src/timestamp_filter.rs`
  - `is_day_granular(&DataType)` — `Date32 | Date64`.
  - `TimestampFilterConvert::convert_high_water_mark(ts)` — a day-granular column compares `>=` against **midnight of the mark's day**; everything else keeps `>` against the exact mark. Resolved **per column**, so the time column and `time_partition_column` each get their own literal and operator (`Bound`).
  - `with_day_granular_columns(bool, bool)`; both default `false`, so a converter built without this information keeps today's strict behaviour.
- `crates/runtime/src/datafusion/filter_converter.rs` — deliberately does *not* resolve day-granularity (see below).
- `crates/runtime/src/accelerated_table/refresh_task.rs` — `day_granular_high_water_columns` resolves it once across both schemas; both high-water-mark sites call `convert_high_water_mark`.

### Keyed on the Arrow type, not on `TimestampFormat`

`data_type_to_timestamp_format` maps only `Date32` to `TimestampFormat::Date`; `Date64` shares the `Timestamp` arm with `Time32`/`Time64`. Widening that arm would loosen two types that are legitimately sub-day, so the decision is made from the `DataType` and threaded to the call sites.

### One decision point, across both schemas

The source filter is built from the federated schema and the dedupe filter from the accelerator schema. Resolving day-granularity inside `create_timestamp_filter_convert` — i.e. per schema — is **not** safe: a `refresh_sql` that casts the time column (`Date32` source projected to `Timestamp` in the accelerator) would widen the source to `>=` while the dedupe stayed on `>`. The already-loaded rows would then sit outside the dedupe's comparison set and the re-fetched rows would be appended again on *every* refresh — the stall turned into unbounded duplication, which is worse than the bug.

So the factory returns a strict converter, and `RefreshTask::day_granular_high_water_columns` resolves the flags once by OR-ing `is_day_granular` across the federated *and* accelerator schemas, handing the same answer to both converters. Either side being day-granular is enough: widening the dedupe side only ever *adds* rows to the comparison set, so it cannot cause duplication, whereas leaving it narrow can.

### The partition predicate is floored, not just widened

`>=` alone does not fix a day-granular `time_partition_column`. The partition predicate is ANDed onto the time predicate, and with a sub-day time column the mark carries a time of day — so `CAST(day AS Timestamp) >= <midday>` is false for the very partition the new rows are in, and the current day stays excluded. The literal has to be floored to the start of the mark's day.

Flooring can never exclude a row the time predicate admits: a partition column tracks its time column, so `day = floor_to_day(ts)`, and `ts > mark` implies `day >= floor_to_day(mark)`. It is a no-op when the mark came from a day-granular column, since that mark is already midnight.

### `Date64` is defensive, not a live path

`validate_time_format` (`crates/runtime/src/accelerated_table/refresh.rs`) rejects a `Date64` time column for *every* `TimeFormat`, and requires `TimeFormat::Date` for `Date32`. So today only `Date32` can actually reach the append path and stall; the `Date64` handling is there so the behaviour is already correct if that validation is ever relaxed, and so the rule reads off the Arrow type rather than off a format that cannot express it.

## Performance impact

Both queries now re-read the latest already-loaded day instead of nothing on a day-granular time column — one extra day of source rows fetched and one extra day of accelerator rows materialised into the dedupe comparison set, per refresh. That is the minimum needed to see the day's new rows at all, and it applies **only** to `Date32`/`Date64` time or partition columns; every sub-day type is bit-for-bit unchanged (same operator, same expression). Datasets with a very large single day and no `append_overlap` pay the most, and can bound it as they can today by using a sub-day `time_column`.

## `append_overlap` interaction

`append_overlap` subtracts its duration from the mark, which is why it already worked around this bug: the mark lands before midnight *D*, so even a strict `>` sees day *D*. A dataset with no `append_overlap` — the default — is the one that stalls. With flooring, a day-granular column's window is quantised to the day boundary, so an overlap smaller than a day now re-reads the whole boundary day. That is extra work absorbed by the dedupe, not a correctness change.

## Known limitations — filed as #12499

The dedupe this fix leans on has two defects of its own, and both are pre-existing: `filter_records` is *set* subtraction rather than multiset subtraction (a legitimately duplicated row is dropped), and its comparison set is collected in full and compared row-by-row (an unbounded-memory window on a high-volume dataset). Both already apply to any dataset with a nonzero `append_overlap`, which is what makes the comparison set non-empty today.

This PR does make that path unconditional for a `Date32` time column, where the comparison set was previously empty — which is precisely the bug. The trade is deliberate and strictly less lossy: losing exact duplicates of already-loaded rows, instead of losing *every* row of the latest day. Fixing the dedupe design properly (counted multiset subtraction, or keyed dedupe) is #12499 and is a different change from the comparison operator.

## Test plan

- `make lint`
- `make test`
- New unit tests in `crates/util/src/timestamp_filter.rs`: `Date32`/`Date64` compare inclusively; `Time32`/`Time64`/`Timestamp`/`Utf8` stay strict; a converter without day-granularity information stays strict; a day partition of a sub-day time column widens independently **and is floored when the mark is at midday**; a day-granular time column is floored too; `convert()` still applies one operator and the unfloored literal to every column (the `refresh_data_window` contract).
- New unit test in `crates/runtime/src/datafusion/filter_converter.rs`: a converter straight out of the factory compares *strictly* even for `Date32`, so a caller that skips resolving day-granularity cannot silently get half the fix.
- New regression test in `crates/runtime/src/accelerated_table/refresh_task.rs`: `test_except_existing_records_from_date32_time_column_dedupes_same_day` loads one row for day *D*, feeds back that row plus a new same-day row, and asserts exactly one row survives and it is the new one. On the pre-change code the comparison set is empty, both rows survive, and the assertion fails — the duplication half of the bug.
- New regression test `test_high_water_comparison_agrees_across_mismatched_schemas`: a `Date32` federated column against a `Timestamp` accelerator column, asserting both converters render the *same* inclusive expression.

## Checklist

- [x] Both high-water-mark `Gt` sites widened in the same commit
- [x] The comparison resolved once, so the two sides cannot disagree
- [x] Day-granular partition literal floored to the day (correct off-midnight, not just at midnight)
- [x] `refresh_data_window` left on strict `Gt` with an unfloored literal
- [x] Sub-day types unchanged
- [x] Regression tests fail on the old code
- [x] Deferred dedupe defects filed as #12499 and documented above
- [ ] `make signoff` green / `Attestation` re-run

Fixes #12492

### DIFF
--- a/crates/runtime/src/accelerated_table/refresh_task.rs
+++ b/crates/runtime/src/accelerated_table/refresh_task.rs
@@ -100,6 +100,7 @@ use tokio::{
 };
 use tracing::{Instrument, Span};
 use util::fibonacci_backoff::FibonacciBackoffBuilder;
+use util::timestamp_filter::is_day_granular;
 use util::{RetryError, retry};
 
 pub(crate) mod changes;
@@ -1138,7 +1139,11 @@ impl RefreshTask {
         let mut filters = vec![];
         if let Some(converter) = filter_converter.as_ref() {
             if let Some(timestamp) = overwrite_timestamp_in_nano {
-                filters.push(converter.convert(timestamp, Operator::Gt));
+                // A high-water mark read back out of the accelerator: compare
+                // inclusively on a day-granular time column, whose values all cast
+                // to midnight, and let `except_existing_records_from` drop the
+                // already-loaded rows it brings back (#12492).
+                filters.push(converter.convert_high_water_mark(timestamp));
             } else if let Some(period) = refresh.period {
                 filters.push(
                     converter.convert(get_timestamp(SystemTime::now() - period), Operator::Gt),
@@ -1181,7 +1186,12 @@ impl RefreshTask {
                     .get_full_or_incremental_append_update(refresh, timestamp)
                     .await
                 {
-                    Ok(data) => match self.except_existing_records_from(refresh, data).await {
+                    // Reuse `timestamp`: the dedupe must compare against the same mark the
+                    // source filter just used, not a freshly read one (#12492).
+                    Ok(data) => match self
+                        .except_existing_records_from(refresh, data, timestamp)
+                        .await
+                    {
                         Ok(data) => Ok(data),
                         Err(e) => Err(e),
                     },
@@ -1727,7 +1737,7 @@ impl RefreshTask {
 
     fn get_filter_converter(&self, refresh: &Refresh) -> Option<TimestampFilterConvert> {
         let schema = self.federated.schema();
-        Self::build_filter_converter(&schema, refresh)
+        self.build_high_water_converter(&schema, refresh)
     }
 
     fn get_accelerator_filter_converter(
@@ -1735,7 +1745,52 @@ impl RefreshTask {
         refresh: &Refresh,
     ) -> Option<TimestampFilterConvert> {
         let schema = self.accelerator.schema();
-        Self::build_filter_converter(&schema, refresh)
+        self.build_high_water_converter(&schema, refresh)
+    }
+
+    /// Build a converter for one side of the append high-water comparison.
+    ///
+    /// Both sides go through here so they cannot be given different day-granularity: the
+    /// flags come from [`Self::day_granular_high_water_columns`], which reads both schemas.
+    fn build_high_water_converter(
+        &self,
+        schema: &SchemaRef,
+        refresh: &Refresh,
+    ) -> Option<TimestampFilterConvert> {
+        let (time, partition) = self.day_granular_high_water_columns(refresh);
+        Self::build_filter_converter(schema, refresh)
+            .map(|converter| converter.with_day_granular_columns(time, partition))
+    }
+
+    /// Whether the time column and the partition column must be treated as day-granular
+    /// on the append high-water-mark path, as `(time_column, time_partition_column)`.
+    ///
+    /// Resolved across the federated *and* accelerator schemas together, and handed to
+    /// both converters, so the source filter and the dedupe filter cannot disagree about
+    /// the comparison. Deciding it from each converter's own schema is not safe: a
+    /// `refresh_sql` that casts the time column — say a `Date32` source projected to
+    /// `Timestamp` in the accelerator — would widen the source to `>=` while the dedupe
+    /// stayed on `>`. The already-loaded rows would then sit outside the dedupe's
+    /// comparison set and the re-fetched rows would be appended again on *every* refresh,
+    /// turning the stall this fixes into unbounded duplication (#12492).
+    ///
+    /// Either side being day-granular is enough. Widening the dedupe side only ever adds
+    /// rows to the comparison set, so it cannot cause duplication; leaving it narrow can.
+    fn day_granular_high_water_columns(&self, refresh: &Refresh) -> (bool, bool) {
+        let schemas = [self.federated.schema(), self.accelerator.schema()];
+        let any_day_granular = |column: Option<&str>| {
+            let Some(column) = column else { return false };
+            schemas.iter().any(|schema| {
+                schema
+                    .column_with_name(column)
+                    .is_some_and(|(_, field)| is_day_granular(field.data_type()))
+            })
+        };
+
+        (
+            any_day_granular(refresh.time_column.as_deref()),
+            any_day_granular(refresh.time_partition_column.as_deref()),
+        )
     }
 
     fn build_filter_converter(
@@ -1862,13 +1917,24 @@ impl RefreshTask {
         ctx
     }
 
+    /// `high_water_mark` is the mark the source filter was built from. Pass it whenever
+    /// the caller has it: recomputing it here reads `max(time_column)` a second time, and
+    /// a write that lands on the accelerator between the two reads gives this comparison a
+    /// *higher* mark than the source query used. The rows in between are then missing from
+    /// the comparison set and are appended again — the same duplication as widening one
+    /// side without the other (#12492). `None` falls back to deriving it.
     async fn except_existing_records_from(
         &self,
         refresh: &Refresh,
         mut update: StreamingDataUpdate,
+        high_water_mark: Option<u128>,
     ) -> Result<StreamingDataUpdate, RetryError<super::Error>> {
-        let Some(value) = self.timestamp_nanos_for_append_query(refresh).await? else {
-            return Ok(update);
+        let value = match high_water_mark {
+            Some(value) => value,
+            None => match self.timestamp_nanos_for_append_query(refresh).await? {
+                Some(value) => value,
+                None => return Ok(update),
+            },
         };
         let Some(filter_converter) = self.get_accelerator_filter_converter(refresh) else {
             return Ok(update);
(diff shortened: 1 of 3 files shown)


grokspice:
@copilot review

grokspice:
Good catch on the suppressed `floor_to_day` comment — it is a real defect and now fixed in 0410699.

`rem_euclid` is never negative, so `signed - signed.rem_euclid(NANOS_PER_DAY)` underflows once the mark is within a day of the bottom of the `i64` range: with `signed == i64::MIN` the remainder is 763145224192 and the subtraction panics in debug (which is how the tests run) and wraps to a meaningless bound in release. A `Date32` date before 1678 is enough to reach it.

Now `saturating_sub`, which clamps the bound to `i64::MIN`. For the inclusive `>=` comparison this feeds, a bound at the bottom of the range simply admits every row, so saturating is the correct degenerate behaviour rather than just a panic guard. Added `test_high_water_mark_floors_an_extreme_negative_mark_without_underflowing`, which would have panicked on the previous code.

lukekim (approved):

PR #12501, opened 2026-08-04
fix(cli): treat a blank --api-key as no key at all (fixes #12498)

### LINKED ISSUE #12498: An explicitly empty `--api-key ""` is sent as an empty credential and suppresses the .env fallback
### Summary

An explicitly empty `--api-key ""` is carried through the CLI as `Some("")` and sent as an empty credential on every authenticated request, rather than being treated as "no key". The result is an authentication failure whose message points at the credential rather than at the empty flag that caused it.

### Where it comes from

`Context::with_args` assigns the flag verbatim:

```rust
ctx.api_key = api_key;

// Load API key from .env if not provided
if ctx.api_key.is_none() {
    ctx.api_key = ctx.load_api_key_from_env();
}
```

`bin/spice/src/context.rs:130`

Two consequences follow:

1. `is_none()` is false for `Some("")`, so an empty `--api-key ""` **suppresses the `.env` / `.env.local` fallback**. A user with a working `SPICE_API_KEY` in `.env` who passes an empty flag silently loses it.
2. Both REPL transports then send the empty value, because neither checks for it:
   - Flight (all SQL) — `add_api_key`, `crates/repl/src/lib.rs:963`, sends `authorization: Bearer ` (trailing space, no token).
   - HTTP (`nql`) — `send_nsql_request`, `crates/repl/src/lib.rs:186`, sends an empty `X-API-Key`.

### Why this is filed separately

Raised by a Copilot review on #12496, against the `nql` side only. Declined there on purpose: that PR exists to make `nql` authenticate *the same way* the session's SQL already does (#12491), so special-casing the empty string on one transport would re-open the same divergence in the opposite direction — one session where SQL sends an empty credential and `nql` sends none.

The behaviour is consistent today. Changing it means changing both transports together, which touches every Flight query in the REPL and wants its own PR and its own test.

### Suggested fix

Normalize once at the source rather than at each call site — treat an empty or whitespace-only key as absent in `Context::with_args`, before the `.env` fallback check, so the fallback still applies and both transports keep behaving identically:

```rust
ctx.api_key = api_key.filter(|k| !k.trim().is_empty());
```

Worth confirming the same normalization is wanted for the `SPICE_API_KEY` env value read by `load_api_key_from_env`.

### Acceptance

- `--api-key ""` behaves as if the flag were omitted, including falling back to `.env` / `.env.local`.
- No request carries an empty `Authorization` or `X-API-Key` header.
- SQL and `nql` remain identical in what they send for any given session.

grokspice:
## Summary (root cause)

`Context::with_args` assigned the `--api-key` flag verbatim, so an explicitly empty
`--api-key ""` became `Some("")`. Two things followed:

1. `Some("")` is not `None`, so the `.env` / `.env.local` fallback was skipped — a user
   with a working `SPICE_API_KEY` in `.env` silently lost it by passing a blank flag.
2. Both REPL transports then sent the blank value, because neither checks for it: Flight
   (all SQL) sent `Bearer ` with no token (`crates/repl/src/lib.rs:954`), and HTTP (`nql`)
   sent an empty `X-API-Key`. The resulting authentication failure pointed at the
   credential rather than at the flag that caused it.

## Changes

Normalize a blank credential to `None` once, where the credential is resolved
(`bin/spice/src/context.rs`), rather than at each call site. `with_args` is the only path
that populates `RuntimeContext::api_key` (verified: `new()` sets `None`, the field is
private, and `main.rs:867` is the sole caller), so every consumer is covered — Flight,
HTTP, `get_headers`, and the `SPICE_API_KEY` handed to `spiced` — and SQL and `nql` keep
sending identical credentials for any given session.

Two details are what make "a blank flag behaves as if it were omitted" actually hold:

- **`--api-key` is declared `env = "SPICE_API_KEY"`** (`main.rs:126`), so clap merges the
  flag and that variable into one value and the provenance is gone before the CLI sees it.
  The fallback therefore checks `SPICE_API_KEY` **before** the `.env` files, which is what
  clap itself resolves when the flag is omitted. Without that ordering, `--api-key ""` with
  `SPICE_API_KEY` set could resolve to a `.env` key while omitting the flag resolved to the
  environment's — silently selecting a different credential. This is a no-op for every
  pre-existing path, because when the flag is omitted and that variable is set, clap
  supplies it and the fallback never runs.
- **A blank value stored in a `.env` file stays authoritative.** `spice login` writes
  `SPICE_SPICEAI_API_KEY=` for an app that has no key
  (`commands/login/mod.rs:280`), so falling through to the next file would resurrect an
  older credential. A stored blank resolves to `None` instead: never a blank credential,
  never a stale one.

File precedence is otherwise untouched — `.env.local` still outranks `.env`, and the
earlier line in a file still wins. A key with any content is preserved verbatim; only the
all-blank case is discarded.

Behaviour note: per the issue's acceptance criteria, `--api-key ""` now behaves as if the
flag were omitted, which includes the `.env` fallback. A blank flag is not a way to
suppress authentication — it never was, since the previous behaviour was to send a blank
credential rather than none.

## Test plan

- `make lint`
- `make test`

New tests in `bin/spice/src/context.rs`:

- `test_normalize_credential_discards_blank_values` / `..._keeps_a_real_key_verbatim` —
  empty, spaces and tabs/newlines normalize to `None`; any non-blank key is passed through
  unchanged, including surrounding whitespace.
- `test_with_args_treats_an_empty_api_key_as_absent` — regression test over `""`, `"   "`
  and `"\t"`: the blank flag is never carried verbatim and never resolves to a blank
  credential. Fails on the old code, which returned `Some("")`.
- `test_load_api_key_from_env_files_returns_a_stored_key` / `..._prefers_env_local` — the
  fallback resolves a real stored key, and `.env.local` still outranks `.env`.
- `test_load_api_key_from_env_files_treats_a_stored_blank_as_no_key` — a blank written by
  `spice login` yields `None` rather than resurrecting the older key in `.env`.
- `test_load_api_key_from_env_files_without_any_env_file` — no env file yields `None`.

The `.env` lookup is split into `load_api_key_from_env_files` so these assert against a
temp app dir and never depend on the ambient environment. Coverage limit worth stating:
the resolver is tested directly, not through a spawned `spice` process, so the exact
outgoing header is not asserted end-to-end.

## Checklist

- [x] A blank `--api-key` behaves as if the flag were omitted, including the `.env` fallback
- [x] No request carries an empty `Authorization` or `X-API-Key` header
- [x] SQL and `nql` remain identical in what they send for a given session
- [x] A blank flag and an omitted flag resolve to the same credential
- [x] Tests added, including a regression test that fails on the old code
- [x] `make signoff` green and `Attestation` re-run

Fixes #12498

### DIFF
--- a/bin/spice/src/context.rs
+++ b/bin/spice/src/context.rs
@@ -33,6 +33,16 @@ const SPICED_FILENAME: &str = "spiced";
 const SPICEPODS_DIR: &str = "spicepods";
 const WSL_ENV_KEYS: [&str; 2] = ["WSL_DISTRO_NAME", "WSL_INTEROP"];
 
+/// Treat a blank credential as absent.
+///
+/// A credential that is empty or whitespace-only cannot authenticate anything, so
+/// carrying it as `Some` only means an empty header goes out on the wire and any
+/// fallback that keys off `is_none` is suppressed. The value is otherwise kept
+/// verbatim -- only the all-blank case is discarded.
+fn normalize_credential(value: Option<String>) -> Option<String> {
+    value.filter(|key| !key.trim().is_empty())
+}
+
 /// Runtime context holding paths and configuration for CLI operations.
 #[derive(Debug, Clone)]
 pub struct RuntimeContext {
@@ -127,7 +137,7 @@ impl RuntimeContext {
             ctx.cloud_region = Some(region.to_string());
         }
 
-        ctx.api_key = api_key;
+        ctx.api_key = normalize_credential(api_key);
         ctx.tls_root_certificate_file = tls_root_certificate_file;
 
         // Load API key from .env if not provided
@@ -148,8 +158,26 @@ impl RuntimeContext {
         )
     }
 
-    /// Load API key from .env or .env.local file.
+    /// Load API key from the environment or a .env / .env.local file.
+    ///
+    /// `SPICE_API_KEY` is checked before the files because `--api-key` is declared
+    /// with `env = "SPICE_API_KEY"`: clap resolves that variable itself whenever the
+    /// flag is omitted, so consulting it first here is what makes a blank
+    /// `--api-key` resolve to the same key that omitting the flag would.
     fn load_api_key_from_env(&self) -> Option<String> {
+        normalize_credential(std::env::var("SPICE_API_KEY").ok())
+            .or_else(|| self.load_api_key_from_env_files())
+            .or_else(|| normalize_credential(std::env::var("SPICE_SPICEAI_API_KEY").ok()))
+    }
+
+    /// Load API key from the app's .env.local or .env file.
+    ///
+    /// The first matching entry wins, exactly as before -- .env.local outranks .env,
+    /// and within a file the earlier line wins. A blank value is authoritative but is
+    /// never a credential: `spice login` writes `SPICE_SPICEAI_API_KEY=` for an app
+    /// that has no key, so a blank resolves to `None` rather than falling through to
+    /// an older key in a lower-precedence file.
+    fn load_api_key_from_env_files(&self) -> Option<String> {
         // Try .env.local first, then .env
         let env_files = [".env.local", ".env"];
 
@@ -160,16 +188,13 @@ impl RuntimeContext {
             {
                 for item in env_map.flatten() {
                     if item.0 == "SPICE_SPICEAI_API_KEY" || item.0 == "SPICE_API_KEY" {
-                        return Some(item.1);
+                        return normalize_credential(Some(item.1));
                     }
                 }
             }
         }
 
-        // Also check environment variables
-        std::env::var("SPICE_API_KEY")
-            .or_else(|_| std::env::var("SPICE_SPICEAI_API_KEY"))
-            .ok()
+        None
     }
 
     /// Get the Spice runtime directory (~/.spice).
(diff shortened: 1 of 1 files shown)


grokspice:
@copilot review

grokspice:
Good catch — that precedence is the whole reason the fallback checks `SPICE_API_KEY` before the files, and it had no coverage. Fixed in d80fb88.

Rather than mutating this process's environment (`set_var` is `unsafe` in edition 2024 and racy under a parallel test runner), the environment lookup is now injected, mirroring the existing `is_wsl_environment(get_env)` seam in this same file. Three tests pin the order end to end:

- `test_resolve_api_key_prefers_the_process_environment` — `SPICE_API_KEY` set *and* a different key in `.env.local`, expects the environment's. This is the case you asked for.
- `test_resolve_api_key_prefers_files_over_the_legacy_variable` — a blank primary variable plus a set `SPICE_SPICEAI_API_KEY` plus a `.env` key, expects the file's. This one exists because the first and last tests alone would also have passed under `SPICE_API_KEY > SPICE_SPICEAI_API_KEY > files`.
- `test_resolve_api_key_uses_the_legacy_variable_last` — legacy variable only, expects it.

Together they pin `SPICE_API_KEY > .env files > SPICE_SPICEAI_API_KEY`, and the `.env` file I/O stays real (isolated in a `TempDir`) so the dotenv parsing is still exercised.

One coverage limit worth stating explicitly: this tests the resolver, not a spawned `spice` process, so the exact outgoing `Authorization` / `X-API-Key` header is not asserted end to end.

lukekim (approved):

Issue #12502, opened 2026-08-04
spice query and nsql analyze build their own SDK client, so the same-origin redirect policy does not cover them


grokspice:
# Problem

Two CLI commands do not use `RuntimeContext::http_client` at all. They build their own credentialed client from the pinned `spiceai` SDK:

- `spice query` — `build_client` in `bin/spice/src/commands/query/mod.rs`
- `spice nsql analyze` — `build_spice_client` in `bin/spice/src/commands/nsql/analyze.rs`

Both do `ClientBuilder::new().http_url(ctx.http_endpoint())` and then `.api_key(api_key)` when the context has one. Neither inherits the CLI's redirect policy, because the client belongs to the SDK and the SDK pins its own `reqwest`.

So the fix for #12495 — a same-origin redirect policy on the context client, so the CLI's `X-API-Key` header is not forwarded across origins — does **not** cover these two paths.

# Why it matters

`reqwest` sanitises only the standard credential headers on a cross-origin redirect (`Authorization`, `Cookie`, `Cookie2`, `Proxy-Authorization`, `WWW-Authenticate`). A custom header such as `X-API-Key` is not among them.

That makes the exposure conditional on how the SDK sends the credential:

- If the SDK sends it as `X-API-Key` (or any other custom header), these two commands still disclose the key to whatever a cross-origin `Location` names — the same defect as #12495, reached through `/v1/queries` and the `nsql analyze` helpers.
- If the SDK sends it as `Authorization`, `reqwest` strips it on the cross-origin hop and there is no disclosure.

**This needs to be established before the issue can be closed either way** — it was not verified while fixing #12495, and it is the whole question.

# Suggested work

1. Determine which header the pinned `spiceai` SDK uses for `api_key`.
2. If it is a custom header, either expose a redirect policy on the SDK's `ClientBuilder` and set a same-origin one from the CLI, or have the SDK strip the credential on a cross-origin redirect itself.
3. Consider whether the CLI should be constructing SDK clients per command at all, versus routing these two through the shared context client, so a policy set once applies everywhere.

# Related

- #12495 — the same defect on the context client, fixed there for every call site that goes through `ctx.http_client()`.
- Adjacent but distinct: `spice-cloud-client`, `spice login`, and the Microsoft device-flow clients also keep the default redirect policy. Their bearer headers *are* stripped cross-origin, but token-exchange requests carry secrets and codes in the POST **body**, and a 307/308 replays the body to the new origin. Worth a look in the same pass.

[...]

PR #12503, opened 2026-08-04
fix(cli): follow only same-origin redirects, so the API key cannot leave the origin (fixes #12495)

### LINKED ISSUE #12495: CLI HTTP calls send X-API-Key under reqwest's default redirect policy, so a cross-origin redirect discloses the key
## Describe the bug

Every CLI HTTP call site sends the API key as an `X-API-Key` header on a `reqwest` client
built with the default redirect policy, which follows up to 10 redirects.

`reqwest` sanitises headers when a redirect changes origin, but the set it strips is the
standard one — `Authorization`, `Cookie`, `Cookie2`, `Proxy-Authorization`,
`WWW-Authenticate`. A custom header such as `X-API-Key` is **not** stripped, so a runtime,
proxy, or ingress answering with a cross-origin `Location` has the key forwarded to whatever
that names.

Affected clients, none of which set a redirect policy:

- `RuntimeContext::http_client` (`bin/spice/src/context.rs`) — used by `get`, `post`,
  `post_json`, so every `/v1/*` call the CLI makes through the context helper.
- The per-command clients that build their own request but reuse `ctx.get_headers()`:
  `commands/nsql/mod.rs`, `commands/nsql/analyze.rs`, `commands/search.rs`,
  `commands/chat.rs`, `commands/trace.rs`, `commands/acceleration.rs`.

## Expected behavior

A request carrying a credential either does not follow redirects, or follows only
same-origin ones. A cross-origin redirect should drop the credential or fail loudly rather
than forward it.

## Additional context

Found by adversarial review of #12491, which added `X-API-Key` to the SQL REPL's `nql`
path. That PR constrains only the request it introduced — it disables redirects for the
`nql` request when a key is present, and leaves keyless requests on the default policy — so
the REPL's new credential is not exposed. The pre-existing call sites above are unchanged
and are what this issue covers.

Worth deciding once, centrally, rather than per call site: the natural home is
`RuntimeContext::http_client` plus a shared builder the per-command clients also use, so a
future call site inherits the policy instead of having to remember it.

A related loose end in the same area: with redirects disabled, a 3xx comes back as a
response whose body fails JSON parsing, so the user sees "Response may be malformed"
rather than "the endpoint redirected". Whichever policy is chosen, the redirect case is
worth reporting as itself.

grokspice:
## Summary

The CLI sends its API key as an `X-API-Key` header on a `reqwest` client that used the **default** redirect policy, which follows up to 10 redirects. `reqwest` sanitises headers when a redirect changes origin, but only the standard set — `Authorization`, `Cookie`, `Cookie2`, `Proxy-Authorization`, `WWW-Authenticate`. A custom header is not in it, so a runtime, proxy or ingress answering with a cross-origin `Location` had the key forwarded to whatever that named.

The client now follows **same-origin redirects only** and refuses to leave the origin.

## Changes

- `bin/spice/src/context.rs`
  - `is_same_origin(previous, next)` — scheme, host, and *effective* port, so `http://host` and `http://host:80` are one origin.
  - `same_origin_redirect_policy()` — a `reqwest::redirect::Policy` that follows a same-origin `Location`, stops on a cross-origin one, and stops after `MAX_REDIRECTS` (10, matching `reqwest`'s own depth).
  - `RuntimeContext::new` applies it to the client.
- `bin/spice/src/error.rs` — a new `HttpClientBuild` variant (see below).

### What this covers, and what it does not

Covered — everything that goes through the context client: the `get`/`post`/`post_json` helpers, and the per-command sites that build their own *request* from `ctx.http_client()` — `commands/nsql/mod.rs`, `commands/search.rs`, `commands/chat.rs`, `commands/trace.rs`, `commands/acceleration.rs`, and `commands/add.rs` (which adds the key explicitly for a `connect` fetch). None of those builds its own client, so the policy set here covers them all and a future call site inherits it rather than having to remember it.

**Not covered — `spice query` and `spice nsql analyze` build their own credentialed client from the pinned `spiceai` SDK** (`build_client` in `commands/query/mod.rs`, `build_spice_client` in `commands/nsql/analyze.rs`). Those pass `ctx.api_key()` into `ClientBuilder`, and the client belongs to the SDK, which pins its own `reqwest` — so they cannot inherit this policy. Whether they are exposed depends on which header the SDK uses: a custom one is the same defect, `Authorization` is stripped by `reqwest` and is safe. That is unresolved and tracked in **#12502**; it needs an SDK-side change either way, so it is not fixable in this PR.

I had originally claimed this was a single central fix for every credential-bearing call site. That was wrong, and the adversarial review caught it — the claim holds for request-building call sites, not for the two that build their own client.

### The build no longer falls back to a default client

`RuntimeContext::new` used `.build().unwrap_or_default()`. That fallback is exactly the wrong shape for this fix: a default `reqwest::Client` has the **default** redirect policy, so the one path that silently produced a client would produce the leaking one. `new` already returns `Result`, so the failure now propagates as `Error::HttpClientBuild`.

### Stopping, not erroring

A cross-origin redirect hands the 3xx back to the caller as a response rather than failing the transport, which keeps it a diagnosable condition. The credential is never sent to the new origin either way — the policy is consulted before the redirected request goes out.

## Behaviour change

A flow that today depends on the CLI following a cross-origin redirect stops working, and surfaces as an unexpected non-2xx rather than a redirect message. For a `/v1/*` call that is the intended trade: today such a flow works *by disclosing the API key to the other origin*.

Two things worth calling out honestly:

- **The policy is broader than the vulnerability strictly requires.** The same client also serves the *unauthenticated* spicepod download in `commands/add.rs`, so a registry or CDN that answers with a cross-origin `Location` would now fail there too, even though no credential is at risk on that request. Splitting into a credentialed client (strict) and a public one (default) would narrow it; I did not do that here because I cannot exercise either path locally, and shipping the leak fix is the priority. Reviewers who know the Spicerack download path should say if it redirects off-origin.
- **A redirect still reports poorly.** The issue's related loose end — saying "the endpoint redirected" instead of "Response may be malformed" — lives in `crates/repl/src/lib.rs`, which #12494 and #12496 are both currently rewriting; touching it here would create a three-way conflict over a string. Worth a follow-up once those land.

## Test plan

- `make lint`
- `make test`
- New unit test `test_is_same_origin_matches_scheme_host_and_effective_port` in `bin/spice/src/context.rs`: an exact match and both default-port equivalences (`http`/80, `https`/443) are the same origin; a differing port, a differing host, and `http`→`https` on the same host are not.
- The policy itself is a thin wrapper — `reqwest::redirect::Attempt` cannot be constructed outside `reqwest`, so the decision logic is unit-tested through `is_same_origin` and the wiring is the single `.redirect(...)` line. Flagging that rather than papering over it: there is no test that drives a real cross-origin 302 end to end.

## Checklist

- [x] Every call site sharing the context client is covered (enumerated above)
- [x] The two SDK-client paths that are *not* covered are identified and tracked (#12502)
- [x] No silent fallback to a default-policy client
- [x] Default ports treated as the same origin
- [x] `http`→`https` on the same host treated as cross-origin
- [x] Redirect depth compared the way `Policy::limited` compares it (`previous()` counts the initial URL)
- [ ] `make signoff` green / `Attestation` re-run

Fixes #12495


grokspice:
@copilot review

grokspice:
Coordination note for whoever reviews this alongside #12508 — flagging it rather than letting it land silently.

This PR defines `same_origin_redirect_policy()` (plus `is_same_origin` and `MAX_REDIRECTS`) privately in `bin/spice/src/context.rs`. #12508, which is independent of this branch — neither contains the other — adds a `pub` `same_origin_redirect_policy()` in `crates/spice-cloud-client/src/redirect.rs`. The two implementations agree semantically (both follow same-origin only, both cap the chain at 10 the way `Policy::limited` does, expressed as `> MAX_REDIRECTS` here and `<= MAX_REDIRECTS` there).

They touch different files, so there is **no merge conflict** — which is precisely the problem: both will land quietly and leave two copies of the same policy in the tree, and `bin/spice` already depends on `spice-cloud-client` (`spice-cloud-client.workspace = true`). That is the duplication-that-lets-settings-diverge shape #12508 itself set out to remove for the login clients.

`crates/spice-cloud-client` is the right home, since it is the crate `bin/spice` and the Cloud client already share. Once **either** of these merges, the other should drop its copy and import the shared one. I have not folded them together now because both are mid-sign-off and a push would discard an in-flight run for no functional gain; I will do the dedup as a follow-up on whichever merges second. If you would rather see them reconciled before either merges, say so and I will stack this branch on #12508 instead.

No behaviour change is implied either way — this is purely about not shipping the helper twice.

lukekim (dismissed):

grokspice:
Coordination note for reviewers: **#12508 independently adds the same helper in a different crate.**

This PR adds `same_origin_redirect_policy()` / `is_same_origin()` to `bin/spice/src/context.rs`; #12508 (fixes #12505, credential retention on the login path) adds a `same_origin_redirect_policy()` and `is_same_origin()` to `crates/spice-cloud-client/src/redirect.rs`. The two were written in parallel against the same class of bug, so there is **no merge conflict** — different files, different crates — but there will be two copies of the same origin comparison once both land.

Two things follow:

- **The natural home is the shared crate.** After both merge, this PR's copy should import from `spice_cloud_client::redirect` (or the helper should move somewhere both can reach) rather than keeping two definitions of "same origin" that can drift apart. I have not done that here because this PR is green after three sign-off attempts and #12508 is not merged yet; folding it in now would invalidate both and risk the same 353-minute budget problem seen on #12448.
- **#12508 has the test this PR admits it lacks.** Its `test_same_origin_redirect_is_followed` drives an actual redirect, which is possible in that crate; here the policy can only be tested through `is_same_origin`, because `reqwest::redirect::Attempt` cannot be constructed outside `reqwest`. When the helpers are unified, that test should be the one that survives.

Also worth noting for scope: #12508 covers the login / cloud-client credential paths, which this PR explicitly does not, and which #12502 lists as adjacent. Between the two, the remaining uncovered case is still the pinned `spiceai` SDK client used by `spice query` and `spice nsql analyze` (#12502).

phillipleblanc:
Removed from the merge queue: it showed a merge conflict against trunk and couldn't be merged. Please rebase/merge trunk into this branch, resolve the conflict, and re-add it to the queue.

grokspice:
Merged `trunk` to clear the conflict, and the resolution changed code that was already approved — flagging it so the re-review is not a surprise.

The conflict was in `bin/spice/src/context.rs`: this branch added a local `same_origin_redirect_policy` next to the same constants where `trunk` added `normalize_credential`. But `trunk` has since landed `spice_cloud_client::redirect::same_origin_redirect_policy`, which made the local copy a duplicate — and the weaker of the two.

The local copy compared scheme, host and effective port by hand. That reads two hostless URLs (`file:///a` and `file:///b`) as the *same* origin, because both have `host_str() == None` and no port. The shared helper defers to `Url::origin`, where a scheme with no host gets an opaque origin that never compares equal, so such a hop is always refused. Landing the weaker duplicate — and having the CLI use it — seemed like the wrong outcome for a fix of this kind, so I dropped the local copy and imported the shared one.

What remains unique to this branch is the part the shared helper does not cover:

- `RuntimeContext`'s HTTP client is built with the policy, so every `/v1/*` call the CLI makes is covered at one place rather than per call site.
- Building that client is now an error instead of `unwrap_or_default()`. A defaulted client carries no redirect policy at all, which would have silently reinstated exactly the behaviour being fixed.

The unit test over the removed local helper went with it, since the shared crate already covers the origin comparison. In its place are two tests that go through `RuntimeContext::with_args`, so the client under test is the one `RuntimeContext::new` actually builds — a test that assembled its own client would still pass if the policy were dropped from the constructor:

- a cross-origin 307 is refused, the response comes back as the 3xx, and the off-origin listener is asserted never to have been contacted;
- a same-origin redirect is still followed to a 200.

Net diff against `trunk` is 3 files (`context.rs`, `error.rs`, `main.rs`). Remote sign-off is running.

lukekim (approved):

Issue #12436, opened 2026-08-04
cayenne: four turso integration tests overflow the tokio worker stack


bjchambers:
## Summary

Four `cayenne` integration tests abort with `fatal runtime error: stack overflow`
when the crate is built the way the workspace builds it. They have never run in
CI, so this is newly surfaced rather than newly broken.

```
cayenne::mutation_property_test  prop_sequential_cold_impl_turso
cayenne::mutation_property_test  prop_sequential_key_impl_turso
cayenne::mutation_property_test  prop_sequential_position_impl_turso
cayenne::cold_tier_test          test_cold_tier_promotion_racing_stage_b_finalize_impl_turso
```

## Why they have never run

`make nextest` ran cayenne's integration tests under `-p cayenne`, and that
selection resolves cayenne to its own defaults — a single feature, `default`.
The workspace resolve gives it three: `default,partition-table-provider,turso`.
So every `*_turso` variant of cayenne's integration tests (304 of them) was
skipped by the gate, and these four have been aborting the whole time.

This is the same class of problem `docs/dev/ci_signoff.md` already documents for
`runtime`: a scoped `-p` build compiles a configuration no real build produces.

## Reproduction

```bash
cargo nextest run --all --exclude libnfs --tests --cargo-profile dev \
  -E 'test(=prop_sequential_key_impl_turso)'
```

```
thread 'prop_sequential_key_impl_turso' has overflowed its stack
fatal runtime error: stack overflow, aborting
(test aborted with signal 6: SIGABRT)
```

Deterministic: 4/4 abort in isolation and 6/6 under retry, so it is neither
load-dependent nor a function of the proptest input.

## What the evidence points to

Setting `RUST_MIN_STACK=8388608` — 8 MiB, the value the *main* thread already
gets — makes all four pass:

| stack | result |
| --- | --- |
| unset | 4/4 abort |
| `RUST_MIN_STACK=8MiB` | 4/4 pass (~21s each) |
| `RUST_MIN_STACK=16MiB` | 4/4 pass |

These tests are `#[tokio::test(flavor = "multi_thread", worker_threads = 4)]`.
Tokio does not set a worker stack size unless asked, so workers get std's 2 MiB
default, which `RUST_MIN_STACK` overrides. That puts the requirement of this code
path somewhere above 2 MiB and at or below 8 MiB.

## The part worth a second opinion

`spiced`'s own tokio runtimes do not set `thread_stack_size` either, so its
[...]

PR #12514, opened 2026-08-05
perf(cdc): build a drained CDC burst with one blocking-pool handoff

### LINKED ISSUE #12206: perf(cdc): batch the per-envelope spawn_blocking offload in the apply loop (Postgres)
## What

The per-dataset CDC apply loop pays one `spawn_blocking` round-trip **per envelope** to build a deferred batch. Postgres is now the only source that takes this path, and it takes it on every envelope.

`crates/runtime/src/accelerated_table/refresh_task/changes.rs:1924`

```rust
for env in envelopes {
    let (committer, batch, _is_ready) = match env.into_parts_offloaded().await {
        Ok(parts) => parts,
        Err(e) => { /* terminal error for this dataset */ }
    };
    committers.push(committer);
    batches.push(batch);
}
```

`into_parts_offloaded` dispatches to the blocking pool and awaits a result per envelope. The decode work is identical either way — the cost removed is scheduling.

MySQL no longer pays it: since the shared pump decodes on delivery, its envelopes arrive materialised and `into_parts_offloaded` short-circuits (`crates/data_components/src/cdc.rs:571`). Postgres still builds deferred envelopes (`crates/data_components/src/postgres_replication/shared.rs:855` → `ChangeEnvelope::new_from_rows`, with `PgChangeRows` implementing `ChangeRows` at `crates/data_components/src/postgres_replication/changes.rs:459`).

## Proposed change

One offload for the whole drained burst instead of one per envelope:

```rust
let built: Result<Vec<_>, _> = tokio::task::block_in_place(|| {
    envelopes.into_iter().map(cdc::ChangeEnvelope::into_parts).collect()
});
```

`into_parts` is the synchronous variant, so N dispatches and N awaits collapse to one handoff. Note the error semantics narrow slightly: `collect()` into a `Result` discards the whole burst at the first failure, whereas today the envelopes built before the failure are also dropped without acking — equivalent for this path, since it already ends in a terminal per-dataset error and the source re-streams.

## Measured

**+26 % on the deferred path** — 231.7 → ~292 MiB/s, measured with `crates/runtime/benches/mysql_cdc_e2e.rs` against a real captured binlog (512 MiB, 40,417 envelopes) while MySQL still deferred. That is ~11 µs of fixed dispatch cost per envelope.

The win is amortisation, so it scales with envelopes-per-burst and matters most for many small transactions; it is negligible for bulk ones. It is also likely **larger in production than this figure**: 11 µs is the uncontended cost, and with several datasets sharing one blocking pool each dispatch queues behind the others. It composes with deeper prefetch, which produces bigger bursts.

## Verification

Reproduce with the same bench on a Postgres-backed replay, or re-measure on the MySQL deferred path by reverting the pump-side decode.

Context: #12199 (which removed this cost for MySQL by decoding on the pump — 2.15× e2e there).

sgrebnov:
## 📝 Summary

The CDC apply loop built each drained envelope's batch with its own `into_parts_offloaded()` call, so a burst of N deferred envelopes paid N `spawn_blocking` round trips. The decode work is the same either way; the round trip is scheduling cost.

`cdc::into_parts_offloaded_burst` builds the whole burst behind one handoff, and resolves inline when every envelope is already materialized. Both MySQL binlog and Postgres logical replication build deferred envelopes, so both take this path — MySQL far more often.

## 🔗 Related

Closes #12206

## 📊 Measured

CH-benCHmark SF1000, MySQL, `mysql-cayenne[file]-tuned.yaml`, one run per arm:

| | trunk | branch | Δ |
|---|---|---|---|
| rows applied /s | 108,892 | **150,826** | **+38.5%** |
| apply µs/envelope | 186.5 | **93.6** | **−49.8%** |
| worst P99 freshness | 275,437 ms | **24,932 ms** | **−91%** |
| QPH | 768.0 | 731.8 | −4.7% |

- trunk `24cb0e71fd` (this branch's merge-base): https://github.com/spiceai/spiceai/actions/runs/31049252946
- branch `d4af6e10a0`: https://github.com/spiceai/spiceai/actions/runs/31055846075

Both runs recorded 0 mem-tier spills and 0 binlog reconnects, and tpmC matched within 0.7% (246,223 vs 244,592) — so the workload driver was constant and the deltas are attributable to the change.

## Why the measurement needs a specific config

On the shipped `-adaptive` config these numbers are not measurable: the adaptive tuner's memory-pressure rule collapses `mem_tier_max_bytes` to its 64 MiB floor, which forces a continuous mem-tier spill loop and, on MySQL, dropped binlog connections. That makes SF1000 runs bimodal — the same config spans 54–109k rows/s with no code change — and it swamps any effect this PR has. Root cause and evidence: **https://github.com/spiceai/spiceai/issues/12531**.

`mysql-cayenne[file]-tuned.yaml` (added here, dispatched daily at SF1000) pins the three memory knobs at the values adaptive already warm-starts them to, leaving every other lever adaptive. That holds both arms at 0 spills, which is what makes the comparison above possible.


sgrebnov:
### Clean-regime SF1 result: decode win confirmed, and the QPH concern does not reproduce

SF1 was chosen as the instrument because the machinery that makes SF1000 bimodal is simply absent there — **0 binlog reconnects, 0 pump stalls, 0 mem-tier spills, 0 critical-memory-pressure actions** across every run. 2 runs per arm so far (merge-base `24cb0e71fd` vs branch `d4af6e10a0`), 4 of 6 complete:

| metric | trunk (n=2) | branch (n=2) | Δ |
|---|---|---|---|
| decode µs/envelope | 140.1 [78.2–202.0] | **35.2 [33.1–37.2]** | **−75%** |
| decode % of apply burst | 74.7% | 59.3% | −20.6% |
| worst P99 freshness | 2.9s | 2.8s | −2% |
| QPH | 89,051 [79,142–98,960] | 99,591 [99,548–99,633] | +11.8% |

The decode figure separates cleanly: the **worst** branch run (37.2 µs/env) still beats the **best** trunk run (78.2) by 2.1×.

**QPH does not regress here.** The branch sits at the top of trunk's range, which is what the earlier SF1000 QPH drop being burst-size-bound predicts (at `cdc_prefetch_buffer=1024` the same branch build measured QPH 1,058 vs 856–941 at the shipped 16384). So that concern was a tuning-parameter effect, not a cost of this change.

**One thing SF1 cannot measure — worth stating rather than quoting.** The rows/s and tpmC deltas from this run set are *not* attributable to the code. The OLTP driver varied 2.8× between runs (tpmC 10,952–30,228), and applied rows track it 1:1:

```
arm      tpmC      rows/s   rows/s per tpmC
trunk    28,968    17,741        0.612
trunk    10,952     6,756        0.617
branch   24,772    15,188        0.613
branch   30,228    18,520        0.613
```

A constant ratio to three decimals means the apply is keeping up with whatever arrives — SF1 is ingress-limited, so it has no apply-throughput headroom to measure. Decode µs/envelope is per-envelope normalized and therefore unaffected.

**Net:** the decode cost reduction is now confirmed in a regime with no confounds, and the one metric that looked like a real cost does not reproduce. An attributable end-to-end throughput number still needs an apply-bound config without the spill/reconnect bimodality — i.e. it stays blocked on #12527 / the tuner defect behind it.

sgrebnov:
### SF1 clean-regime A/B complete (3v3) — all four merge criteria met

All 6 runs succeeded. SF1 has none of the machinery that makes SF1000 bimodal: **0 binlog reconnects, 0 pump stalls, 0 mem-tier spills, 0 critical-memory-pressure actions** in every run, both arms.

| metric | trunk (n=3) | branch (n=3) | Δ |
|---|---|---|---|
| **decode µs/envelope** | 78.2 [76.7–202.0] | **37.2 [33.1–37.7]** | **−52%** |
| decode % of apply burst | 73.0% [71.9–76.3] | 59.0% [58.3–59.6] | −19% |
| worst P99 freshness | 2.9s [2.9–3.0] | 2.9s [2.8–2.9] | −2.3% |
| QPH | 98,960 [79,142–105,635] | 99,548 [92,496–99,633] | +0.6% |

Decode separates cleanly: the **best** trunk run (76.7 µs/env) is still 2.0× the **worst** branch run (37.7). Branch spread is ±6%, trunk ±45%.

**Against the criteria set before the runs:**

1. **Decode drops materially** — yes, −52% with no regime confound.
2. **Freshness no worse** — yes, −2.3% (equal within noise).
3. **QPH inside the trunk band** — yes, +0.6%, ranges overlap heavily. The SF1000 QPH drop does not reproduce, consistent with it being burst-size-bound.
4. **Correctness gates pass** — yes, and this is the first time they have actually executed on this branch (every prior SF1000 run was cancelled before drain). All three branch runs: 7/7 datasets exact row-count match with zero watermark lag, analytic gate enabled at concurrency 16. Example:
   ```
   drain-probe +0s order_line  rows src=3662898 spice=3662898 [ok] | max_bench_ts lag 0ms [ok]
   ```

**What these runs do NOT show, and why the rows/s line is omitted above.** The OLTP driver varied 2.6× across the six runs (tpmC 10,952–30,228), and applied rows track it 1:1 — `rows/s ÷ tpmC` came out 0.612–0.617 in every single run, constant to three decimals. At SF1 the apply keeps up with whatever arrives, so there is no apply-throughput headroom to measure and the rows/s delta (−14.3%, matching tpmC's −14.3% exactly) is the driver, not the code. Decode µs/envelope is per-envelope normalized and therefore unaffected by this.

So an attributable **end-to-end throughput** number still requires an apply-bound config, which today means SF1000 and its ~2× spill/reconnect bimodality — blocked on #12527 and the tuner defect behind it. What this run set establishes is that the change reduces decode cost substantially, harms nothing measurable, and preserves data correctness.

Remaining before merge: sign-off, and splitting the benchmark spicepods + Cayenne memory-attribution commits out of this branch so the reviewable diff is `cdc.rs` + `changes.rs`.

sgrebnov:
### Correction: the QPH regression *does* reproduce at the shipped prefetch depth

An earlier comment of mine concluded, from the SF1 clean-regime runs, that the SF1000 QPH drop "does not reproduce." That conclusion was wrong, and a new trunk sample makes it testable properly.

A cron run has since produced a second **post-shift** trunk sample (spiced `d0c6acba2b`, 8 commits past the merge-base, none touching CDC/Cayenne/MySQL). Comparing only runs from the same environment era, n=2 per arm, all at the shipped `cdc_prefetch_buffer=16384`:

| metric | trunk (n=2) | branch (n=2) | Δ |
|---|---|---|---|
| decode µs/envelope | 221.4 [147.5–295.4] | **63.3 [63.2–63.5]** | **−71%** |
| rows applied /s | 82,381 [55,584–109,178] | 79,554 [76,796–82,311] | −3.4% |
| tpmC | 244,148 [243,669–244,628] | 244,789 [244,707–244,872] | +0.3% |
| worst P99 freshness | 407.8s [252.5–563.1] | 439.9s [432.1–447.7] | +7.9% |
| **QPH** | **1,198 [951–1,445]** | **899 [856–941]** | **−25%** |

The OLTP driver held constant here (tpmC within 0.3%), so unlike the SF1 runs these rows/s and QPH numbers are comparable. rows/s and freshness overlap and show nothing. **QPH does not overlap:** both branch samples sit below both trunk samples.

**Why SF1 could not have settled this.** SF1 is not apply-bound — applied rows track the driver 1:1 (`rows/s ÷ tpmC` = 0.612–0.617 across all six runs). A QPH regression caused by apply-side work competing with queries cannot appear in a config where apply has that much headroom, so SF1's +0.6% was uninformative rather than reassuring.

**What still holds.** The regression is burst-size-mediated, not intrinsic. At shallower prefetch the same branch build recovers:

| branch config | QPH | mem-tier spills |
|---|---|---|
| 16384 (shipped) | 856, 941 | 24, 24 |
| 4096 | 1,056, 981 | 27, 9 |
| 1024 | 1,058 | 1 |

4096 and 1024 land inside trunk's [951–1,445] range.

**Bearing on the merge decision.** Criterion 3 ("QPH inside the trunk band") is **not** satisfied at the shipped configuration — I should not have marked it met. Caveat in the other direction: trunk's QPH range spans 1.5× at n=2, and its lower sample (951) is close to the branch's upper (941), so this is suggestive rather than established. More samples at 16384 would settle it.

Practical reading: this change looks safe to merge *together with* a shallower prefetch default, and not obviously safe at 16384 alone. That makes it dependent on the tuning work in #12527 rather than independent of it.

sgrebnov:
### Prefetch sweep at n=2: QPH improves consistently, but the spill loop is not fixed

Completing the `cdc_prefetch_buffer` sweep to 2 runs per depth changes one of my earlier conclusions. All six runs are branch spiced, post-environment-shift, same config otherwise:

| depth | QPH | spills | spill seconds | in-load reconnects | rows/s | worst P99 freshness |
|---|---|---|---|---|---|---|
| 16384 (shipped) | 941, 856 | 24, 24 | 465s, 415s | 3, 3 | 76,796 / 82,311 | 448s, 432s |
| 4096 | 1,056, 981 | 27, 9 | 342s, 116s | 2, 1 | 73,756 / 99,292 | 476s, 313s |
| 1024 | 1,058, 1,118 | 1, 16 | 15s, 342s | 0, 2 | 95,560 / 57,730 | 334s, 551s |

**What holds — QPH.** All four shallow-buffer runs (981–1,118) beat both deep-buffer runs (856–941), with no overlap. This is the clean result, and it is the one the QPH discussion above rests on: the regression at the shipped depth recovers at shallower depths, on both 4096 and 1024, across four runs.

**What does not hold — spill suppression.** I previously reported 1024 as cutting spills 24 → 1 (a 28× reduction in spill time). Its second run spilled 16 times for 342s and produced the worst freshness in the sweep (551s). At n=2, 1024 spans 1–16 spills and 57.7k–95.6k rows/s; 4096 spans 9–27 spills. Both shallow depths are bimodal, and **1024's bad run is worse than either 16384 run on rows/s and freshness**. The single-run result was a lucky sample and should not have been reported as a threshold effect.

Notably 16384 is the *most reproducible* depth (24/24 spills, 3/3 reconnects) — it just reproduces a bad state.

**Consequence.** Prefetch depth is not a reliable fix for the spill loop; it is a reliable improvement to QPH. Since the spill loop is what drives the run-to-run bimodality, the remaining candidate is the mechanism itself: pinning the knobs the tuner drives to a bound (`cayenne_cdc_mem_tier_max_bytes`, `cayenne_inline_flush_max_bytes`, `cayenne_target_file_size_mb`), so the collapse that forces continuous spilling cannot happen. Those runs are queued.

sgrebnov:
### Apply throughput: −41% cost per envelope and per row, in the 0-spill regime

I previously framed this change as "efficiency, not performance," on the grounds that no end-to-end ingest gain was attributable. That framing was too narrow — it reported the decode *stage* but never the normalized cost of the whole apply loop, which is the capacity measure and is immune to the driver variation that made SF1's rows/s unusable.

SF1, 3 runs per arm, 0 mem-tier spills / 0 reconnects / 0 memory-pressure actions in every run:

| metric | trunk (n=3) | branch (n=3) | Δ |
|---|---|---|---|
| **apply µs/envelope** | 107.1 [106.7–264.7] | **63.1 [55.6–64.6]** | **−41%** |
| **apply µs/row** | 23.79 [23.73–58.41] | **14.03 [12.35–14.35]** | **−41%** |
| ↳ decode component | 78.2 [76.7–202.0] | 37.2 [33.1–37.7] | −52% |
| ↳ non-decode component | 30.0 [28.9–62.7] | 25.9 [22.5–27.0] | −14% |
| apply ms/burst | 120.0 [112.8–120.7] | 60.9 [56.9–65.4] | −49% |
| total apply CPU-seconds | 377.9 [354.7–379.4] | 191.4 [179.2–205.5] | −49% |

Neither normalized metric overlaps between arms: the best trunk run is 1.65× the worst branch run per envelope and per row. Envelopes and rows applied differ by −14.3% between arms (the OLTP driver varied 2.6× across these six runs), which is exactly why the per-unit normalization matters — and why the raw rows/s figure from this run set says nothing.

The non-decode component falling ~14%, also without overlap, is consistent with the mechanism: removing N−1 `spawn_blocking` round trips per burst relieves blocking-pool contention that the write path shares. It was not an anticipated effect.

**Where this does and does not convert to end-to-end gain.** Apply *capacity* is ~41% cheaper per unit of work. Whether that surfaces as higher ingest depends on what binds:

- At SF1, nothing does — applied rows track the driver 1:1 (`rows/s ÷ tpmC` = 0.612–0.617 in all six runs), so freed capacity becomes slack.
- At SF1000 at the shipped `cdc_prefetch_buffer=16384`, the freed capacity is absorbed by the mem-tier spill loop. The branch spills 24–26 times per run versus trunk's 10–16 (3 vs 2 samples, no overlap) at comparable mean throughput — faster apply pushes more data through a tier the adaptive tuner has collapsed to its 64 MiB floor.

So the revised summary: the change reduces apply cost per unit work by ~41% with correctness intact, and realizing that as ingest throughput depends on the tuner defect being addressed (see the memory-pressure signal issue). The QPH cost at 16384 still argues for landing this alongside a shallower prefetch default rather than alone.

sgrebnov:
### Clean 0-spill SF1000 A/B: +38.5% ingest, freshness 275s -> 25s, QPH flat

The confound that blocked every previous SF1000 comparison was the mem-tier spill loop (tracked separately as the tuner memory-pressure defect). Pinning the three knobs that loop depends on — `cayenne_cdc_mem_tier_max_bytes`, `cayenne_inline_flush_max_bytes`, `cayenne_target_file_size_mb` — holds both arms at **0 spills and 0 reconnects**, which finally makes the comparison attributable.

Same spicepod (`mysql-cayenne[file]-adaptive-pinned.yaml`), SF1000, both post-environment-shift:

| | trunk `24cb0e71` | branch `d4af6e10` | Δ |
|---|---|---|---|
| mem-tier spills / binlog reconnects | 0 / 0 | 0 / 0 | — |
| **rows applied /s** | 108,892 | **150,826** | **+38.5%** |
| **apply µs/envelope** | 186.5 | **93.6** | **−49.8%** |
| decode µs/envelope | 143.9 | 54.6 | −62.1% |
| **worst P99 freshness** | 275,437 ms | **24,932 ms** | **−91%** |
| QPH | 768.0 | 731.8 | −4.7% |
| tpmC (workload driver) | 246,223 | 244,592 | −0.7% |

Runs: [trunk 31049252946](https://github.com/spiceai/spiceai/actions/runs/31049252946), [branch 31055846075](https://github.com/spiceai/spiceai/actions/runs/31055846075).

tpmC within 0.7% means the OLTP driver was constant, so unlike earlier run sets the rows/s delta is attributable to the code rather than to load. Freshness dropping to 25s means the pipeline stops falling behind and runs essentially caught up.

**This retires the QPH concern.** −4.7% is inside run-to-run spread. The −25% measured earlier occurred only while the spill loop was active, so it was a symptom of the tuner defect rather than a cost of this change; the earlier "do not merge on the decode numbers alone" note no longer applies.

**Caveats.** n=1 per arm, and the two runs were ~1.7h apart rather than strictly back to back. Three things argue against regime luck: both arms at 0 spills, tpmC constant to 0.7%, and the apply/decode per-envelope deltas matching every prior measurement (SF1 gave −41% apply / −52% decode independently). A second pair would settle it.

**Note on the pinned configuration itself:** it is a benchmark instrument here, not a recommendation. A Postgres run with the same pins failed its readiness gate (2/11 datasets not caught up within 900s), consistent with the pins also blocking the tuner's *growth* levers, which initial catch-up depends on. That cost does not affect the MySQL pair above, where both arms reached readiness normally.

bjchambers (approved):

sgrebnov:
Sign-off note, for the record on how the gate was made to pass.

`make nextest` on this HEAD fails on a test unrelated to this change:

```
SIGABRT [0.190s] cayenne::mutation_property_test prop_sequential_memory_impl_turso
thread 'prop_sequential_memory_impl_turso' has overflowed its stack
```

This is the fifth variant of the family tracked in #12436 and is **not** in the `NEXTEST_STACK_OVERFLOW_12436` exclusion list, so it currently fails for every branch regardless of content — it reproduces on the self-hosted runner on an unrelated branch ([run 31106242545](https://github.com/spiceai/spiceai/actions/runs/31106242545)).

I ran sign-off with `RUST_MIN_STACK=16777216` — the same 16 MiB stack that #12561 sets in code, applied out-of-tree instead. Nothing else was changed, and that single variable is sufficient:

```
RUST_MIN_STACK=16777216 make nextest NEXTEST_FILTER_EXTRA='test(=prop_sequential_memory_impl_turso)'
PASS [12.387s] cayenne::mutation_property_test prop_sequential_memory_impl_turso
```

Full gate with it set: `9773 tests run: 9773 passed (4 slow, 2 flaky, 1 leaky), 27 skipped`, then `✓ Signed off on 4f88be944dc9 — lint+test passed in 1777s`. The two flakes (`warm_subset_reduces_small_file_fanout_sqlite`, `leased::tests::pre_lease_lands_in_next_window_and_promotes_on_roll`) both passed on retry.

Side note that may be useful for #12561: since `RUST_MIN_STACK` alone is sufficient, an `[env]` entry in `.config/nextest.toml` would fix this repo-wide without touching test code.

PR #12536, opened 2026-08-05
Label loaded dataset size as uncompressed

### LINKED ISSUE #5531: Annotate whether dataset size is compressed or not
2024-11-27T23:01:41.897121Z  INFO runtime::accelerated_table::refresh_task: Loaded 2,964,624 rows (419.31 MiB) for dataset taxi_trips in 9s 233ms.

It would be nice to indicate if the value 419.31 MiB is compressed or not.

ranykamel:
## 📝 Summary

When an accelerated dataset loads, the runtime logs the loaded dataset's byte size but gives no indication of whether it is a compressed or uncompressed size. This value is the in-memory Arrow size (`RecordBatch::get_array_memory_size()`), so this PR labels the log explicitly as uncompressed. Example of the log: `Loaded 2,964,624 rows (419.31 MiB uncompressed) for dataset taxi_trips in 9s 233ms.`

## 🔗 Related

Closes #5531.

## 📚 Docs

I believe no doc changes are required. This PR only involves changing a log.

## 👀 Notes for Reviewers

From what I can tell, the memory size value is always uncompressed at this point in the code. Any compression appears to be decoded upstream before reaching here. If it is necessary that the compressed size should also be mentioned in the logging, I could implement passing this value down.


lukekim:
The code was already doing this. It doesn't need a test.

PR #12541, opened 2026-08-05
fix(cayenne): bound protected-snapshot compaction by the compaction memory pool

### LINKED ISSUE #12013: Bug: cayenne fast protected-snapshot compaction OOMs — no settled-size ceiling
## Summary
The fast protected-snapshot subset compaction can select and merge multiple already-large protected snapshots in one pass, driving the runtime OOM. Its tier selector has no "settled size" cutoff, so large runs are re-merged instead of left alone.

## Root cause
`select_protected_snapshot_merge_tier` is relative LSM leveling with no absolute ceiling:
- Tiers grow 8 MiB ×8, unbounded, and saturate on overflow — huge snapshots collapse into one top tier and are treated as same-size peers.
- Merges the lowest tier with ≥ `cayenne_compaction_trigger_protected_snapshots` runs (default 8), up to 32 runs.
- Decoupled from `cayenne_target_file_size_mb`. The current-snapshot compactor marks files ≥ target size as "settled — not a candidate" (`Tier::classify`); the protected-snapshot path has no equivalent.

## Why it OOMs 
Read side streams (`UnionExec` + `execute_stream`), but memory is driven by encode:
1. PK-hash sharding scans all selected inputs (up to 32) concurrently.
2. Each `cayenne_write_concurrency` encoder buffers toward a `target_file_size_bytes` file; full BtrBlocks/FSST cascade holds decompressed Arrow chunks (multiples of on-disk size), amplified by a large target size.
3. Compaction pool (`cayenne_compaction_memory_fraction`, default 0.2 × query pool) does not bound the native Vortex encode working set → grows against the container limit unchecked.

## Observed
During large accelerated-table bootstrap, once several similarly-sized large protected snapshots accumulated, the pass fired, unioned them, and OOM-killed the process.

## Workaround
Set `cayenne_compaction_trigger_protected_snapshots` very high so the pass never fires. Restore the default post-bootstrap, or read amplification grows unbounded.

## Ideas

Add size threshold similar to `cayenne_compaction_max_files_per_pick`, example default is 4 x compaction-pool-mem-size (large is ok as it is streaming/still some overhead), compaction runs are taken to fit size

sgrebnov:
Fixes #12013.

## Problem

The protected-snapshot merge paths selected their inputs by relative size tiers only. Tiers grow geometrically and saturate on overflow, so arbitrarily large runs collapsed into one top tier and merged as same-size peers — up to `PROTECTED_MERGE_MAX_WIDTH` (32) of them unioned into a single streaming rewrite.

Protected-snapshot compaction folds many small protected snapshots into a larger one. Each is a separate branch in every scan's union, so consolidating them lowers query amplification. That stops paying once the runs are large: the benefit is still one fewer scan branch, but the cost is rewriting gigabytes. On a small host it OOM-kills the process — the rewrite streams, but every selected input is scanned concurrently and the Vortex encode cascade holds decompressed Arrow chunks the memory pool does not account for.

The seq-prefix bake had the same exposure and no size tiers at all — it selects an entire older prefix — and it runs first in the compaction trigger.

## Fix

Both paths cap selected input bytes at four times the memory pool the pass accounts against — the dedicated compaction pool where `cayenne_compaction_memory_fraction` carved one, else the query pool — and take runs oldest-first only while they fit. So a tier of large runs consolidates a few at a time instead of all at once.

- No new configuration parameter; the bound is derived from the pool the operator already sized.
- An unbounded pool has no ceiling to derive a budget from and behaves exactly as before.
- The bake truncates its prefix. Safe because truncating only lowers the cutoff `T`, and the clean-prefix gate re-validates the prune against whatever prefix was selected.
- Sizing costs one listing per candidate, so it runs only under a finite budget, and those sizes are reused for the encode fan-out. A gate-blocked bake pays no extra listings.

A tier whose runs cannot be paired within the budget is treated as settled — the memory-derived analogue of the current-snapshot picker's at-or-above-target cutoff. That is cheap rather than a sacrifice: `scan_protected_snapshots` unions one branch per protected snapshot, so a merge buys one fewer scan branch whether the runs are 8 MiB or 8 GiB, while costing the bytes it rewrites. Benefit is per run, cost is per byte, so the ceiling sheds the pass's least valuable work first. The read path already WARNs at `8 × cayenne_compaction_trigger_protected_snapshots` protected snapshots, so the budget skip logs at DEBUG as the cause behind that warning rather than a second alarm for it.

## Follow-ups (not in this PR)

**Protected snapshots are all-or-nothing units today.** Small-file compaction buckets and rewrites files *within the current snapshot only* (`table.rs:12745`), and its subset path is eligible only when there are no protected snapshots at all. A protected snapshot is never optimized internally — it is merged whole or left alone — so the merge re-encodes every file of every selected run, including files with no deletions to apply, producing byte-identical output. Rewriting only the dirty files and referencing the clean ones in place is what would make this path fast rather than merely bounded. Both pieces already exist: cold-tier promotion does exactly that dirty/clean split via `ColdScanFileSubset`, and cross-snapshot manifest references are honored by physical-file GC (`manifest_file_relative_path`).

**The full rewrite's protected fold is still unbounded.** `compact_current_snapshot_small_files`' fallback folds every protected snapshot. Deliberately not gated on this budget: that would permanently disable small-file compaction on any table whose protected set exceeds it, since merging does not shed bytes. It is also not the reported path.

## Tests

Four selector unit tests (budget truncation, settled-pair decline, pool-to-budget scaling including saturation) plus two end-to-end tests on the real compaction path, one per affected path. Both check a precondition against the real on-disk sizes so they cannot pass vacuously, assert no rows are lost, and pair against an unbounded-pool control that merges — the prior behavior — so they fail without this change.

`cargo test -p cayenne --lib`: 993 passed. `cargo clippy -p cayenne --lib --tests`: clean.

`docs/cayenne/cayenne.md` Flow F is updated with the ceiling, the value-density reading, and the dirtiness follow-up, plus a document changelog row.


sgrebnov:
Review comments addressed in 45f309217a.

**Stall log said "two smallest" but reports the two oldest.** Correct, and misleading for sizing the budget. Selection takes runs oldest-first by UUIDv7 order, so `oldest_pair_bytes` is what the operator needs. Wording aligned in the WARN, the DEBUG, and the doc.

The tests still assert against the two *smallest* runs deliberately: the smallest pair is the minimum possible pair sum, so proving it exceeds the budget proves no pair can fit. That is a stronger precondition than checking the oldest pair, and there is now a comment saying so, to keep someone from "aligning" it to the log text.

**Stall tracker not reset on non-stall skips.** Also correct. `record_denial` only starts an episode, so an episode outlived its condition — if another pass folded the protected set away, the next genuine stall would warn on the first denial instead of after ten minutes of sustained stalling. Both paths now end the episode on every skip that proves there is no stall: the protected set below the merge floor, no qualifying tier, no settled prefix, and the trigger's own early-out before the pass is entered.

**Duplicated compaction-semaphore paragraph in the doc.** Pre-existing, and my insert landed between the two copies, which is what surfaced it. Dropped the stale copy claiming `available_parallelism()` and kept the accurate `cpu_budget().cores()` one — the former also contradicts the repo's single-CPU-sizing-source rule.

lukekim (dismissed):

lukekim (dismissed):

phillipleblanc:
Removed from the merge queue: it showed a merge conflict against trunk and couldn't be merged. Please rebase/merge trunk into this branch, resolve the conflict, and re-add it to the queue.

bjchambers (approved):

PR #12548, opened 2026-08-05
fix(deps): bump the Vortex pin and guard the task-cancellation regression

phillipleblanc:
## Summary

Cancelling a task spawned onto the Vortex IO runtime dropped a one-shot channel receiver that
had already been polled, while its sender was still delivering a result. The receiver released
its stored waker from inside its own destructor, freeing it under the sender about to call it.
That corrupts the heap and faults on an indirect call through the freed vtable, presenting as a
`SIGSEGV` on a `tokio-rt-worker` thread — at arbitrary uptime, and uncorrelated with
throughput, because what matters is whether a cancellation interleaves with a completion.

This bumps the Vortex pin to a revision carrying the fix (spiceai/vortex#82), and adds a
regression test so a future pin bump cannot reintroduce the hazard silently.

## The regression test

`crates/runtime/tests/vortex_task_cancellation.rs` drives the cancellation directly: it polls
a spawned task's handle once, so the handle registers a waker, then drops it while the spawned
future is still completing.

Against an affected revision it killed the test process on every run — `SIGSEGV`, or `SIGABRT`
via the allocator's heap checks — and against the fixed revision it passes. Two details matter
for it to catch anything, and are commented in the test:

- the task's result is sized to match the layout of the channel that carries it, so the
  allocation lands in the same size class as the production path. A smaller payload does not
  reproduce the fault.
- it detects the fault by the process dying rather than by failing an assertion, because the
  damage is done before control returns. It is a stress test, not a deterministic one.

It needs `vortex-io` as a dev-dependency. Cargo permits only one revision per git source, so
that pin cannot drift away from the pins the rest of the workspace uses: an upgrade that misses
it fails to resolve rather than quietly disarming the test.

## Testing

- The new test fails against the previous pin and passes against this one, verified repeatedly
  on Linux.
- Upstream, `spiceai/vortex#82` carries its own equivalent stress test plus coverage for the
  closed-channel path, and the `vortex-io` and `vortex-file` suites pass with and without the
  `tokio` feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16967,6 +16967,7 @@ dependencies = [
  "uuid 1.23.2",
  "vortex",
  "vortex-datafusion",
+ "vortex-io",
  "workers",
  "x509-certificate",
  "yaml",
@@ -22439,7 +22440,7 @@ dependencies = [
 [[package]]
 name = "vortex"
 version = "0.1.0"
-source = "git+https://github.com/spiceai/vortex.git?rev=5b4bee10822a2b008e628251f95980b9feba5dd0#5b4bee10822a2b008e628251f95980b9feba5dd0"
+source = "git+https://github.com/spiceai/vortex.git?rev=53b08c10cf4a717422a8c2d7232cb577acd1b997#53b08c10cf4a717422a8c2d7232cb577acd1b997"
 dependencies = [
  "vortex-alp",
  "vortex-array",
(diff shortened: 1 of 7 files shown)


peasee (dismissed):

lukekim (dismissed):

phillipleblanc:
Removed from the merge queue: the `Build and Test` check's `Check if Cargo.lock is updated` step failed — after this PR's Vortex pin bump, the committed `Cargo.lock` doesn't match what `cargo` regenerates. The diff moves a `vortex-io` dependency edge from one crate's tree to another, which looks like the lockfile wasn't fully regenerated after the pin/Cargo.toml change. This also cascaded a failure onto an unrelated PR further back in the queue (via the merge queue's speculative-commit chain). Failed run: https://github.com/spiceai/spiceai/actions/runs/31088422460/job/92574651747

Please run `cargo check` (or equivalent) locally to regenerate `Cargo.lock` against this PR's Cargo.toml changes, commit the update, and re-add to the queue.

ewgenius (approved):

PR #12577, opened 2026-08-05
fix(cayenne): pair the cold manifest with the warm snapshot a scan captured

### LINKED ISSUE #12555: bug(cayenne): cold-tier promotion race lets a concurrent scan observe an inconsistent row count
## Summary

`test_cold_tier_concurrent_scan_during_promotion_impl_turso` (`crates/cayenne/tests/cold_tier_test.rs:585`) fails deterministically: a concurrent scan observes an inconsistent row count during cold-tier promotion, indicating the cold manifest commit and the warm-tier snapshot flip are not publishing atomically with respect to in-flight scans.

Assertion failure (representative, values vary by run):

```
assertion `left == right` failed: promotion 2 (dirty rewrite): a concurrent scan observed 405 rows (expected 402) — the cold manifest commit and the warm snapshot flip must publish atomically w.r.t. scans
```

Observed variants across retries on a single CI run: `405 vs 402`, `602 vs 502`, `800 vs 400` — the wrong value itself is non-deterministic, but the failure is not: nextest retried this test 6 times on one run and it failed all 6 attempts.

## Impact

This surfaced as the blocking check on two unrelated PRs at the front of the merge queue (`pr` workflow, full workspace `nextest` run):

- #12440 (`fix(cpu-budget): ...`) — [failed `pr` run](https://github.com/spiceai/spiceai/actions/runs/30985255273) ([Build and Test job](https://github.com/spiceai/spiceai/actions/runs/30985255273/job/92245531319))
- #12475 (`fix(models): ...`) — [failed `pr` run](https://github.com/spiceai/spiceai/actions/runs/30985256831) ([Build and Test job](https://github.com/spiceai/spiceai/actions/runs/30985256831/job/92242394376))

Neither PR touches `crates/cayenne`, so this is a trunk-level regression, not something introduced by either PR. Because it's deterministic rather than a rare timing blip, it is likely to keep failing every subsequent PR that reaches the front of the merge queue until fixed, which will keep ejecting unrelated PRs.

## Suspected cause

[`#12435` "feat(cayenne): drive incremental vacuum from the maintenance tick"](https://github.com/spiceai/spiceai/pull/12435) merged to trunk shortly before this was first observed, and touches the exact code path the test exercises: `crates/cayenne/src/provider/table.rs`, `catalog.rs`, `cayenne_catalog.rs`, `metastore.rs`, `metastore/sqlite.rs`. This is circumstantial (timing + touched files), not confirmed by bisection — worth checking first.

## Why this matters beyond CI flakiness

The test's own assertion message states the invariant being checked: the cold manifest commit and the warm-tier snapshot flip must publish atomically with respect to concurrent scans. A failure here means a scan can observe a torn/partial state during promotion — i.e. a query can see the wrong row count mid-promotion. Per this project's data-correctness principle, this needs a real fix (atomicity/locking around the manifest-commit + snapshot-flip pair), not a test timeout bump or retry-until-green.

## Ask

- Confirm whether #12435 introduced this (bisect if needed).
- Root-cause the non-atomicity between the cold manifest commit and the warm snapshot flip during promotion.
- Fix the race; do not paper over it by loosening the test's retry/timeout, since the assertion is checking a real correctness invariant.

phillipleblanc:
Fixes #12555.

## The invariant that broke

A cold-tier ("datalake") promotion has **two** visibility publication points: the metastore cold-manifest commit (`cayenne_cold_tier_file`, which is where a cross-tier read gets its cold files) and the in-memory warm snapshot flip. `promote_warm_to_cold` publishes both under one `listing_fence.write()`, so on the write side they land atomically.

The read side had drifted out of that discipline. `capture_raw_scan_input` captured the warm snapshot id under `listing_fence.read()` and released the fence; the cold plan branch then listed the manifest from the metastore afterwards. A scan whose capture preceded the promotion's commit and whose cold read followed it paired the **old** warm snapshot with the **new** cold manifest and counted every promoted row twice. The failing assertion's values are exactly that double count: 400+400=800, 402+3=405, 502+100=602.

**This is not only a race.** A read-only CDC replica scans at a non-zero freshness tolerance (~1s), so it can be served a `ScanView` captured before a promotion while the manifest is resolved live — 800 rows where 400 is correct, single-threaded, deterministically, for up to the freshness lag after every promotion. The new regression test below reproduces that with no concurrency at all, and it fails on both metastore backends before this change.

## Introducing change

`d45b89e2` (#12005, demand-driven scan-view cache). Before it, `scan()` held `listing_fence.read()` across the whole plan build — which contained the cold-manifest read — so both halves came from one fenced instant. Narrowing the hold to the capture left the cold read outside it.

Measured on the named test's turso variant, 150 runs per revision under identical load: `4afe0461` (the commit that added the test) 0 failures; `292c7126` (parent of #12005) 0; `d45b89e2` (#12005) 27; trunk 53. #12435, which the issue suspected, is ruled out — the defect is present from 2026-07-28, a week before it merged, and it implements a pragma for the SQLite metastore only while the failing backend is turso.

## The fix

The cold file set is resolved **inside** the capture, under the same held read fence as the warm snapshot id, and carried on `RawScanInput`; the plan branch consumes what the capture pinned instead of re-reading.

The manifest is cached against the snapshot id it was published with. That key is sound because every mutation of `cayenne_cold_tier_file` mints a new snapshot id in the same metastore transaction — promotion commits through `commit_overwrite_in_txn`, which is also what an overwrite/truncate clear runs. So an entry whose id still matches the live one cannot be stale, one that does not is a miss that re-reads, and no path can leave a mismatched pair servable; the cache needs no invalidation hook on the paths that clear the manifest. The promotion hands the manifest it just committed to that cache inside its own write fence, so its own flip costs no read.

Net metastore I/O **falls**: a pointer clone in the steady state instead of a SELECT per cold-tier scan. Cold-tier GC still reads the manifest live, deliberately — it is the GC root and must never be stale.

## The same seam in the keyset rebuild

The PK keyset rebuild runs off `write_lock` (`prepare_stream_for_insert` precedes the Stage-A guard), and `resolve_cold_keyset_source` read the manifest live before the rebuild captured the warm snapshot. With the cold tier served by the per-file PK blooms, a promotion in that window left the promoted keys in neither the pre-promotion bloom nor the post-promotion warm snapshot: an upsert of such a key finds no existing key, records no supersede tombstone, and leaks the prior cold copy — a durable over-count, the same failure class the surrounding comments already warn about for protected snapshots.

The resolution now reports the snapshot id it was decided against; the rebuild folds the cold tier exactly (from the manifest resolved in its own fenced instant) when that id has moved on, and the persisted-bloom checkpoint path — which never scans cold — fails closed to the full rebuild on the same condition. The bloom is also published only while the snapshot it was built for is still live, so two rebuilds racing off `write_lock` can no longer leave a post-promotion keyset half beside a pre-promotion bloom half that omit the same key.

## Test-thread stack

`libtest` gives each test thread the std default 2 MiB, which is not enough for these debug-built async futures, and the failure mode is silent: the process aborts with "has overflowed its stack" and the test contributes **no result line** rather than going red. On a 2 MiB stack four cayenne turso tests abort instead of running (`mutation_property_test prop_sequential_{cold,key,position}_impl_turso`, `cold_tier_test …promotion_racing_stage_b_finalize_impl_turso`) and a fifth sits within 64 KiB of the limit, so unrelated code motion flips it either way. `.cargo/config.toml` now asks for 8 MiB. No assertion is relaxed and no timeout raised; four tests that were silently skipped now actually run and pass.

## Validation

- **Regression coverage.** The existing `test_cold_tier_concurrent_scan_during_promotion` is the race gate; a new `test_cold_tier_stale_tolerant_scan_during_promotion` covers the same seam deterministically on the stale-tolerant read path.
- **Mutation evidence.** Existing test: 53/150 turso failures before, **0/400 runs after**. New test: `800 vs 400` on both backends before, passes after. The new test is non-vacuous — it fails on unmodified trunk for the reason under test, and its assertion distinguishes the two states (400 correct vs 800 torn).
- Full `cayenne` suite **1723/1723**, 0 failures (trunk's baseline aborted 4).
- `make lint-rust PACKAGES=cayenne FEATURES=turso,partition-table-provider` clean (fmt, crate-layering, clippy lib + `--tests`).

`docs/cayenne/cayenne.md` updated (cross-tier read section) with a *Document changelog* row.

## Known adjacent gaps, not addressed here

- A key that was **already** cold before a promotion can miss both halves if a writer's prepare completes before a promotion and its apply runs after (the local keyset is warm-only and the shared bloom was cleared). Pre-existing, independent of the scan pairing, and needs the keyset to be pinned to a generation across prepare→apply.
- The snapshot-id pairing is scoped to one process; two processes promoting the same table can still pair a local warm snapshot with a foreign cold manifest on a cache miss. Strictly narrower than resolving the manifest live on every scan, but not a cross-process guarantee.


bjchambers (dismissed):

sgrebnov (approved):

phillipleblanc:
Removed from the merge queue: it showed a merge conflict against trunk and couldn't be merged. This PR's title/description suggest it may overlap with #12563, which is currently ahead of it in the queue and addresses the same cold-manifest/warm-snapshot pairing — that's the most likely source of the conflict once #12563 lands. Please rebase/merge trunk into this branch (checking against #12563's changes if it has merged by then), resolve the conflict, and re-add it to the queue.

PR #12586, opened 2026-08-05
fix(mysql): raise net_write_timeout on the shared binlog dump session (fixes #12527)

### LINKED ISSUE #12527: fix(cdc): one slow dataset drops the shared MySQL binlog connection and halves ingest
## What

On the MySQL CDC path, all `refresh_mode: changes` datasets on a connection share **one** `COM_BINLOG_DUMP` stream. `deliver_to_member` (`crates/data_components/src/mysql_replication/shared.rs:2033`) must-delivers each envelope into a bounded per-dataset channel with unbounded retry, so when one dataset's apply loop stalls, the pump stops reading the binlog socket for **every** dataset on that dump.

If the block exceeds the source's `net_write_timeout` (default 60s), the server aborts the dump:

```
WARN mysql_replication::binlog: binlog connection lost; reconnecting attempt=1
  error=Input/output error: bytes remaining on stream
```

The reconnect re-streams from the last acked position while the pipeline is already several hundred seconds behind, which deepens lag, lengthens the next stall, and invites the next reconnect.

## Why it matters

The result is a **bimodal benchmark**: runs either avoid in-load reconnects and go fast, or take one and stay degraded. From six CH-benCHmark SF1000 adaptive runs (identical config; the spread is not explained by any code difference between them):

| spiced | rows/s | worst P99 freshness | in-load reconnects | ≥60s stalls in load |
|---|---|---|---|---|
| A | 54,046 | 633s | 1 | 6 |
| A | 60,387 | 540s | 2 | 8 |
| B | 58,151 | 549s | 2 | 6 |
| C | 60,130 | 537s | 1 | 1 |
| **D** | **109,178** | **253s** | **0** | **0** |
| E | 76,796 | 448s | 3 | 22 |

The single reconnect-free run is ~2x faster than every run that took one. This ~2x swing makes the config unusable for A/B comparison of any ingest-path change.

## Evidence chain

1. **A small number of very slow apply cycles.** `order_line` apply-cycle histogram: 95.8% ≤10s, 98.4% ≤25s, and **3 of 189 cycles in the 25–100s bucket**, against 4–5 reconnects per run.
2. **Those cycles are in-memory-tier spills.** `cayenne_write_phase_duration_ms{phase=inmemory_spill}` for `order_line`: **24 spills, mean 19,378ms, 465s total** in a ~908s window. On the reconnect-free run, `order_line` never spilled (only `customer` did, 10 spills at mean 671ms). Compaction and encode-permit waits are identical between runs, so neither is implicated.
3. **The channel saturates.** `order_line` prefetch buffer occupancy p99 = 16,384/16,384 (100%).
4. **The pump blocks, then the connection dies.** Every stall warning in both runs was `order_line`, and every reconnect was immediately preceded by an `order_line` stall reaching 65–115s.
5. **All datasets share the outcome.** Per-dataset lag spread on MySQL is ~1x (all datasets report near-identical lag), versus 359x on Postgres, whose 7 tables are grouped across 4 replication slots.

## Why Postgres does not have this

Postgres's pump blocks on a slow member the same way (`postgres_replication/shared.rs:2699`, same warn loop), but survives it because:

- a **separate keepalive task** sends `StandbyStatusUpdate` independently of apply progress (`postgres_replication/changes.rs:556`), and the server retains WAL in the slot; and
- the **4-slot topology** isolates the heavy producers, so one slow table cannot starve the others.

MySQL cannot copy the keepalive: `COM_BINLOG_DUMP` is one-way once started (the existing `master_heartbeat_period` heartbeat runs server-to-client), and the server is not waiting for a message from us — `net_write_timeout` fires on **unread data**, not on missing feedback.

What MySQL does have is the other half: binlog files on disk plus a client-owned position, so any position within retention is re-openable. Today the pipeline already relies on "re-open from the saved position" as its recovery path; it just gets there via a connection abort affecting every dataset rather than deliberately for the one that is behind.

### Option 1 — raise `net_write_timeout` on the dump session (tolerate the block)

Set `SET SESSION net_write_timeout = <n>` in the pre-dump window that already sets the heartbeat variables (`mysql_replication/binlog.rs:89`). Note the syntax differs: the heartbeat uses a *user* variable (`SET @master_heartbeat_period`), while `net_write_timeout` is a *system* variable — the `@var` form would be inert.

180s clears the observed ~100s worst-case cycle with ~1.8x margin. Session-scoped; no server configuration or privileges required.

Caveat: the apply tail is not provably bounded (`cdc_max_coalesced_envelopes`/`_bytes` bound the burst, not spill duration), so this is empirical headroom. It should ship with the `stalled_for` warning also exported as a metric so a genuinely wedged apply stays visible.

**CI acceptance criteria** (CH-benCHmark SF1000 adaptive, MySQL, 3 runs):
- in-load reconnects = **0** in all 3 runs (currently 1–3)
- worst P99 freshness within 20% of the reconnect-free reference (~253s)
- rows applied/s within 20% of the reconnect-free reference (~109k)
- no increase in `≥60s` stall episodes attributable to the change

### Option 2 — group datasets across several dumps (contain the block)

`mysql_replication_server_id` already keys the shared-source registry, so datasets sharing an id share a dump. Mirror the Postgres topology — isolate the heavy producers, let the light tables share:

```
order_line -> server_id 1001
stock      -> server_id 1002
customer, district, new_order, oorder, warehouse -> server_id 1003
```

This is a spicepod-only change. Note the existing one-dump-per-dataset variant is measurably worse (341 stall warnings spread across all seven datasets, and the server decodes/ships the binlog 7x), so grouping is the middle ground rather than full split.

Limitation: `order_line`'s own dump can still exceed the timeout — this bounds the blast radius, it does not remove the cliff.

**CI acceptance criteria** (3 runs):
- per-dataset lag spread >> 1x (datasets no longer report near-identical lag)
- the five light datasets' worst P99 freshness improves by >=30% vs the one-dump config
- total in-load stall warnings do not exceed the one-dump config
- server-side dump count = 3 (not 7)

### Option 3 — non-blocking router with deliberate per-member re-open (eliminate the coupling)

Split the pump into a reader that always drains the socket and a router that never blocks. A member that cannot accept is detached with its position recorded and re-attached from its own dump, instead of wedging the shared stream. This is the true equivalent of the Postgres design, and it is available precisely because the MySQL ack is client-owned.

Requires a retention-headroom metric: unlike a Postgres slot, MySQL binlog retention is time/size-based with no per-client guarantee, so re-open-from-position must be checked against `binlog_expire_logs_seconds` versus peak lag.

**CI acceptance criteria** (3 runs):
- in-load reconnects = 0
- a single dataset's apply stall does not raise any other dataset's lag (lag series decoupled)
- reader socket-drain never blocks: `stalled_for` warnings for the *shared pump* = 0
- new metric: binlog retention headroom vs observed peak lag, exported and non-negative

## Contributing defect (separate)

The spills that trigger this are driven by the adaptive tuner's memory-pressure signal, which reads cgroup `memory.current` (`cayenne/src/provider/tuning.rs:454`) — a value that **includes reclaimable page cache** — divided by total host memory, with CRITICAL at 0.90. Observed process memory peaked near 71% of budget while the tuner recorded 12–14 "critical memory pressure" actions, collapsing `mem_tier_max_bytes` to its 64 MiB floor while independently raising `target_vortex_file_size_bytes` to 1 GiB. A 64 MiB tier feeding 1 GiB target files spills continuously, and each spill writes more files, generating more page cache. This affects Postgres too (14 critical-pressure actions, spills on three tables) and will be filed separately.

claudespice:
## Summary

Every `refresh_mode: changes` dataset on a MySQL connection shares one `COM_BINLOG_DUMP` stream, and the pump must-delivers each envelope into a bounded per-dataset channel with unbounded retry. So when one dataset's apply loop stalls — an in-memory-tier spill, for instance — the pump stops reading the dump socket for **every** dataset on that connection.

`COM_BINLOG_DUMP` is one-way once started: the server is not waiting for us to say anything, so `net_write_timeout` fires purely on data we have not read. At the server default of 60s a single stalled apply cycle is enough for the source to abort the dump:

```
WARN mysql_replication::binlog: binlog connection lost; reconnecting attempt=1
  error=Input/output error: bytes remaining on stream
```

Every member then resumes from its last acked position while the pipeline is already several hundred seconds behind, which deepens lag, lengthens the next stall, and invites the next abort. Measured on CH-benCHmark SF1000 adaptive, that makes the whole config bimodal: the one reconnect-free run of six was ~2x faster (109k rows/s, 253s worst P99 freshness) than every run that took a reconnect (54–77k rows/s, 448–633s).

Postgres does not have this because a separate keepalive task acks independently of apply progress and the server retains WAL in the slot. MySQL cannot copy that — the dump is one-way — but it does not need to: binlog files plus a client-owned position make any position within retention re-openable, which is already the recovery path. The connection abort just gets there involuntarily, for every dataset at once.

This raises `net_write_timeout` on the dump session (option 1 in #12527), which is the change that removes the cliff without altering the topology.

## Changes

- **`SET SESSION net_write_timeout = GREATEST(@@net_write_timeout, 180)`** on the dump connection, in the pre-dump window that already sets the heartbeat variables. 180s clears the worst apply cycle observed at SF1000 (~100s) with ~1.8x margin. Two details the statement depends on:
  - It is a **floor**, not an assignment. Raising `net_write_timeout` on the source is the existing manual workaround for this exact symptom, so an operator who has already set it to, say, 600 must not have it clamped back down to 180 by connecting a newer runtime. `GREATEST` against the inherited session value gets that in one statement with no round-trip.
  - `net_write_timeout` is a **system** variable, so it needs the `SESSION` form. The heartbeat period next to it is a **user** variable (`SET @master_heartbeat_period`), and the `@var` spelling applied to a system variable would define an unrelated user variable and silently leave the server default in place. A unit test pins that distinction, because nothing at runtime would report it.
- The pre-dump statements moved into `pre_dump_session_statements`, a pure function returning the statements in issue order, so which variable form each one uses is assertable without a MySQL server. A rejected `net_write_timeout` now warns (with the consequence and the remedy) where the heartbeat spellings stay at debug — those are deliberately tried in pairs and one is always unknown on a given server version.
- **`replication_member_send_stalled_seconds_total`**, exported per dataset, accruing the time the pump spent blocked on that dataset's channel. This is the sibling of the metric the Postgres shared-slot path already exports under the same name, and it is the reason 180s is safe to ship as empirical headroom rather than a proof: the apply tail is not provably bounded (`cdc_max_coalesced_envelopes`/`_bytes` bound the burst, not how long a spill takes), so a genuinely wedged apply still reaches the timeout — and now stays visible as a counter after the warning scrolls away. The existing stall warning points at the metric by name.

Deliberately not in scope: the dataset-to-dump grouping (option 2) is spicepod-only and orthogonal, and the non-blocking router (option 3) is a redesign that needs a binlog-retention-headroom metric first. The contributing spill defect — the adaptive tuner reading cgroup `memory.current` including reclaimable page cache — affects Postgres too and is tracked separately.

## Test plan

- `make lint`
- `make nextest`
- New unit tests, all in the crates the change touches:
  - the dump session raises `net_write_timeout`, in the system-variable form, above the server default
  - the statement never lowers an operator's already-higher `net_write_timeout`
  - a rejected `net_write_timeout` is the one pre-dump statement worth warning about
  - both heartbeat spellings survive the refactor, as user variables in nanoseconds
  - a sub-second checkpoint interval still floors the heartbeat at 500ms
  - a stalled delivery accrues the stall counter per interval, and a successful send accrues nothing
- Seven deterministic neuters, each caught by a distinct assertion: dropping the statement; the `SET @net_write_timeout` user-variable spelling; a bare assignment instead of `GREATEST`; a timeout equal to the server default; silencing the rejection warning; dropping the metric accrual; accruing on the success path.
- The behaviour this fixes is a source-side timeout under sustained load, so the acceptance criteria in #12527 (in-load reconnects = 0 across 3 CH-benCHmark SF1000 adaptive runs, freshness and rows/s within 20% of the reconnect-free reference) need a benchmark run to confirm and are not covered by unit tests.

## Checklist

- [x] Regression tests that fail without the fix
- [x] No new blocking calls in async paths
- [x] Metric mirrors the existing Postgres sibling's name and semantics
- [ ] Benchmark acceptance criteria from #12527 confirmed on a CH-benCHmark SF1000 adaptive run

Fixes #12527

### DIFF
--- a/crates/data_components/src/mysql_replication/binlog.rs
+++ b/crates/data_components/src/mysql_replication/binlog.rs
@@ -95,10 +95,11 @@ const DUMP_NET_WRITE_TIMEOUT_SECS: u32 = 180;
 /// One statement issued on the dump connection before `COM_BINLOG_DUMP`.
 struct PreDumpStatement {
     sql: String,
-    /// Whether a rejection deserves a user-visible warning. The heartbeat
-    /// spellings are deliberately tried in pairs and one of them is unknown on
-    /// any given server version, so those stay at debug.
-    warn_on_error: bool,
+    /// What the user loses if the server rejects this statement, when that is
+    /// worth saying. The heartbeat spellings are deliberately tried in pairs
+    /// and one of them is unknown on any given server version, so those carry
+    /// nothing and a rejection stays at debug.
+    rejection_warning: Option<&'static str>,
 }
 
 /// The session setup a dump connection needs, in the order it is issued.
@@ -124,7 +125,7 @@ fn pre_dump_session_statements(checkpoint_interval: Duration) -> Vec<PreDumpStat
             .into_iter()
             .map(|var| PreDumpStatement {
                 sql: format!("SET @{var} = {heartbeat_nanos}"),
-                warn_on_error: false,
+                rejection_warning: None,
             })
             .collect();
     // `net_write_timeout` is a system variable, so it needs the `SESSION`
@@ -137,7 +138,9 @@ fn pre_dump_session_statements(checkpoint_interval: Duration) -> Vec<PreDumpStat
             "SET SESSION net_write_timeout = \
              GREATEST(@@SESSION.net_write_timeout, {DUMP_NET_WRITE_TIMEOUT_SECS})"
         ),
-        warn_on_error: true,
+        rejection_warning: Some(
+            "the source can still abort the shared binlog connection when one dataset's apply loop stalls, delaying changes for every changes-mode dataset on it. Grant the replication user permission to set session variables, or raise the source's net_write_timeout. See: https://spiceai.org/docs/components/data-connectors/mysql",
+        ),
     });
     statements
 }
@@ -151,14 +154,19 @@ pub(super) async fn open_binlog_stream(
 ) -> std::result::Result<BinlogStream, mysql_async::Error> {
     let mut conn = Conn::new(params.opts.clone()).await?;
 
-    for PreDumpStatement { sql, warn_on_error } in
-        pre_dump_session_statements(params.checkpoint_interval)
+    for PreDumpStatement {
+        sql,
+        rejection_warning,
+    } in pre_dump_session_statements(params.checkpoint_interval)
     {
         if let Err(e) = mysql_async::prelude::Queryable::query_drop(&mut conn, sql.as_str()).await {
-            if warn_on_error {
-                tracing::warn!(dataset = %dataset_name, statement = %sql, error = %e, "Failed to configure the MySQL binlog dump session: the source can still abort the shared binlog connection when one dataset's apply loop stalls, delaying changes for every changes-mode dataset on this connection. Grant the replication user permission to set session variables, or raise the source's net_write_timeout. See: https://spiceai.org/docs/components/data-connectors/mysql");
-            } else {
-                tracing::debug!(dataset = %dataset_name, statement = %sql, error = %e, "failed to set a binlog dump session variable");
+            match rejection_warning {
+                Some(consequence) => {
+                    tracing::warn!(dataset = %dataset_name, statement = %sql, error = %e, "Failed to configure the MySQL binlog dump session for dataset {dataset_name}: {consequence}");
+                }
+                None => {
+                    tracing::debug!(dataset = %dataset_name, statement = %sql, error = %e, "failed to set a binlog dump session variable");
+                }
             }
         }
     }
@@ -1024,6 +1032,17 @@ mod tests {
             .map(|statement| statement.sql.as_str())
     }
 
+    /// The warning the statement setting `variable` carries, if it carries one.
+    fn statement_rejection_warning(
+        statements: &[PreDumpStatement],
+        variable: &str,
+    ) -> Option<&'static str> {
+        statements
+            .iter()
+            .find(|statement| statement.sql.contains(variable))
+            .and_then(|statement| statement.rejection_warning)
+    }
+
     #[test]
     fn the_dump_session_raises_net_write_timeout() {
         // Regression test for #12527: without this the source aborts the shared
@@ -1079,12 +1098,18 @@ mod tests {
         let statements = pre_dump_session_statements(Duration::from_secs(10));
         for statement in &statements {
             assert_eq!(
-                statement.warn_on_error,
+                statement.rejection_warning.is_some(),
                 statement.sql.contains("net_write_timeout"),
                 "unexpected error visibility for {}",
                 statement.sql
             );
         }
+        let warning = statement_rejection_warning(&statements, "net_write_timeout")
+            .expect("a rejected net_write_timeout must say what it costs");
+        assert!(
+            warning.contains("https://spiceai.org/docs/"),
+            "the warning must point at the fix: {warning}"
+        );
     }
 
     #[test]
--- a/crates/data_components/src/mysql_replication/binlog.rs
+++ b/crates/data_components/src/mysql_replication/binlog.rs
@@ -65,6 +65,83 @@ use crate::cdc::{ChangeEnvelope, StreamError, build_heartbeat_envelope};
 /// that (fake rotates and heartbeats report 0) are not resumable.
 pub(super) const MIN_VALID_EVENT_POS: u64 = 4;
 
+/// Floor, in seconds, for how long the source will hold unread dump data before
+/// aborting the connection. Applied to the dump session because the server
+/// default is 60s.
+///
+/// The dump is one-way once started, so the server never waits for us to say
+/// anything — `net_write_timeout` fires purely on data we have not read. Every
+/// `refresh_mode: changes` dataset on a connection shares one dump, and the
+/// pump must-delivers each envelope into a bounded per-dataset channel, so a
+/// single dataset whose apply loop stalls (an in-memory-tier spill, for
+/// instance) stops the socket being drained for all of them. At the default the
+/// source then aborts the dump and every member resumes from its last acked
+/// position while already far behind, which deepens lag and invites the next
+/// abort.
+///
+/// 180s clears the worst apply cycle observed on CH-benCHmark SF1000 (~100s)
+/// with ~1.8x margin. This is empirical headroom, not a bound: the apply tail
+/// is not provably bounded (`cdc_max_coalesced_envelopes`/`_bytes` bound the
+/// burst, not how long a spill takes), so a genuinely wedged apply still
+/// reaches the timeout — it stays visible as the member send stall warning and
+/// its `replication_member_send_stalled_seconds_total` counter.
+///
+/// A floor rather than an assignment: an operator who already raised
+/// `net_write_timeout` past this — the manual workaround for exactly this
+/// symptom — must not have it lowered by connecting a newer runtime, so the
+/// statement takes the greater of their value and this one.
+const DUMP_NET_WRITE_TIMEOUT_SECS: u32 = 180;
+
+/// One statement issued on the dump connection before `COM_BINLOG_DUMP`.
+struct PreDumpStatement {
+    sql: String,
+    /// Whether a rejection deserves a user-visible warning. The heartbeat
+    /// spellings are deliberately tried in pairs and one of them is unknown on
+    /// any given server version, so those stay at debug.
+    warn_on_error: bool,
+}
+
+/// The session setup a dump connection needs, in the order it is issued.
+///
+/// Split out from [`open_binlog_stream`] so the statements — in particular
+/// which of them address a *user* variable versus a *system* one — are
+/// assertable without a `MySQL` server.
+fn pre_dump_session_statements(checkpoint_interval: Duration) -> Vec<PreDumpStatement> {
+    // Ask the source to send heartbeat events while idle so the stream can
+    // detect dead connections and advance its checkpoint. Half the
+    // checkpoint interval (min 500ms) keeps idle persists within ~1.5×
+    // the interval. The session variable is in nanoseconds; MySQL 8.4
+    // renamed the replica-facing vocabulary, so set both spellings (unknown
+    // user variables are inert).
+    let heartbeat_nanos = (checkpoint_interval / 2)
+        .max(Duration::from_millis(500))
+        .as_nanos()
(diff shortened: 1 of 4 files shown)


claudespice:
@copilot review

claudespice:
Fixed in e54d3478b — the warning now reads "the source can still abort the shared binlog connection when one dataset's apply loop stalls". You were right that the wording implied a new risk: this branch of the code runs when the `SET SESSION net_write_timeout` failed, so it is reporting that the mitigation could not be applied and the pre-existing behaviour stands.

crates/data_components/src/mysql_replication/binlog.rs
```diff
@@ -65,35 +65,101 @@ use crate::cdc::{ChangeEnvelope, StreamError, build_heartbeat_envelope};
...
+    gtid: &GtidSet,
+) -> std::result::Result<BinlogStream, mysql_async::Error> {
+    let mut conn = Conn::new(params.opts.clone()).await?;
+
+    for PreDumpStatement { sql, warn_on_error } in
+        pre_dump_session_statements(params.checkpoint_interval)
+    {
+        if let Err(e) = mysql_async::prelude::Queryable::query_drop(&mut conn, sql.as_str()).await {
+            if warn_on_error {
+                tracing::warn!(dataset = %dataset_name, statement = %sql, error = %e, "Failed to configure the MySQL binlog dump session: the source can still abort the shared binlog connection when one dataset's apply loop stalls, delaying changes for every changes-mode dataset on this connection. Grant the replication user permission to set session variables, or raise the source's net_write_timeout. See: https://spiceai.org/docs/components/data-connectors/mysql");
```
sgrebnov: 1. Should be more concise, simple
2. Warning itself is statement/sql specific - if needed, add this into PreDumpStatement

claudespice: Done in 2ecf584 — `PreDumpStatement` now carries the message itself (`rejection_warning: Option<&'static str>`) instead of a `warn_on_error` flag, so the text lives on the one statement it is about and the loop just chooses between warning with what the statement supplied and a debug line. The warning also lost the clause that repeated what the statement was trying to do, since `statement = %sql` is already on the event.

115 `data_components` lib tests pass on `--features mysql`; the existing visibility test now asserts against `rejection_warning.is_some()` and additionally that the message points at the docs.

crates/data_components/src/mysql_replication/binlog.rs
```diff
@@ -944,6 +1018,125 @@ mod tests {
...
+    fn both_heartbeat_spellings_are_set_as_user_variables_in_nanoseconds() {
+        // Half the checkpoint interval, in nanoseconds, on both the pre-8.4 and
+        // 8.4 spellings — unchanged by the net_write_timeout addition.
+        let statements = pre_dump_session_statements(Duration::from_secs(10));
+        for var in ["master_heartbeat_period", "source_heartbeat_period"] {
+            assert_eq!(
+                statement_setting(&statements, var),
+                Some(format!("SET @{var} = 5000000000").as_str()),
+                "{var} must stay a user variable, in nanoseconds"
+            );
```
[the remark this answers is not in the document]
claudespice: This one is a false positive — the pattern compiles, and the reference is not dangling.

`format!(...)` produces a temporary `String`, and a temporary lives until the end of the *enclosing statement*, not the end of the sub-expression. The borrow taken by `.as_str()` is therefore valid for the whole `assert_eq!(...)` call, which is that statement.

Reduced to the same shape and compiled standalone on the workspace edition:

```rust
fn lookup<'a>(v: &'a [String], k: &str) -> Option<&'a str> {
    v.iter().find(|s| s.contains(k)).map(|s| s.as_str())
}
fn main() {
    let v = vec!["SET @master_heartbeat_period = 5000000000".to_string()];
    let var = "master_heartbeat_period";
    assert_eq!(
        lookup(&v, var),
        Some(format!("SET @{var} = 5000000000").as_str()),
        "{var} must stay a user variable, in nanoseconds"
    );
    println!("compiles and passes");
}
```

`rustc --edition 2024` accepts it and it prints `compiles and passes`.

Comparing against an owned `String` is not an option here anyway: `statement_setting` returns `Option<&str>`, so the right-hand side has to be an `Option<&str>` too. Leaving as is.

claudespice:
Dispatched the sign-off for `2ecf5846c`: [31058868345](https://github.com/spiceai/spiceai/actions/runs/31058868345). It had none.

The earlier dispatch at 23:21Z ([31056060061](https://github.com/spiceai/spiceai/actions/runs/31056060061)) failed 14 seconds in, at `Resolve sign-off target`, with `one of 'branch' or 'pr_number' is required` — the run was started with both inputs empty (`BRANCH:` and `PR_NUMBER:` are blank in its env dump). Because it never reached the sign-off job, it posted no commit status at all, so `2ecf5846c` carried zero `signoff` statuses and this PR just looked like a plain red `Attestation` with nothing in flight. Worth knowing as a failure shape: a botched dispatch is invisible from the commit statuses and only shows up in the workflow run list.

Also cleared the last review thread. Copilot flagged `Some(format!(...).as_str())` in `both_heartbeat_spellings_are_set_as_user_variables_in_nanoseconds` as a dangling reference that would not compile; it compiles — the temporary lives to the end of the enclosing statement — and the reduced case is in the thread reply.

claudespice:
Sign-off [31058868345](https://github.com/spiceai/spiceai/actions/runs/31058868345) failed after 13635s, but **not on anything in this PR**. The suite ran 9728 tests: 9727 passed, and the single failure was

```
cayenne::cold_tier_test test_cold_tier_concurrent_scan_during_promotion_impl_turso
crates/cayenne/tests/cold_tier_test.rs:585
assertion `left == right` failed: promotion 2 (dirty rewrite): a concurrent scan
observed 405 rows (expected 402)
  left: 405
 right: 402
```

This branch touches four MySQL replication files and nothing under `crates/cayenne`, so the failure is the known cold-tier promotion race tracked in #12555 — already being fixed by #12563 and #12577, both currently in the merge queue.

No change made here. Re-dispatched sign-off on the same head `2ecf5846c`: [31070558622](https://github.com/spiceai/spiceai/actions/runs/31070558622).

lukekim (approved):

Issue #12597, opened 2026-08-05
Gemini streaming fails with 'Invalid UTF-8' when a character is split across two reads


claudespice:
## Summary

`parse_sse_stream` in `crates/google-genai/src/generate.rs` decodes each network chunk on its
own with `std::str::from_utf8`. A multi-byte character split across two reads is not invalid
UTF-8 — it is half of a character, with the other half in the next read — but decoding the
halves separately makes it look invalid, and the whole stream fails:

```rust
let text = std::str::from_utf8(&bytes).map_err(|e| {
    crate::error::Error::StreamError {
        message: format!("Invalid UTF-8: {e}"),
    }
});
```

The error blames the model's output for being malformed when the bytes are fine and only the
reading of them is wrong.

## Why it is not the event-splitting bug

The event framing here is already correct: the function accumulates into `buffer` and only
takes an event once it finds `\n\n`, so an *event* split across reads is handled. It is the
UTF-8 decode that happens too early — before the bytes that complete the character have
arrived.

## When it happens

Any response containing a non-ASCII character — accented text, CJK, an emoji — where the
character lands on a read boundary. Likelier on a slower link, through a proxy that re-chunks,
or on TLS record boundaries, and likelier the longer the response. Purely ASCII answers are
unaffected, which is why it is intermittent.

## Suggested fix

Accumulate the **bytes** and decode only complete lines, or use an incremental UTF-8 decoder.
A `\n` cannot occur inside a multi-byte UTF-8 sequence, so splitting the byte buffer on the
line terminator makes every decode land on a character boundary.

`bin/spice/src/sse.rs` (added in the fix for #12588) does exactly this for the CLI's chat
stream and could be the basis of a shared decoder — the workspace now has three hand-rolled
SSE readers (this one, the CLI's, and `async-openai`'s vendored `eventsource-stream`).

A regression test can drive the parser with a chunk pair split between the two bytes of a
character and assert the character survives.

## Affected versions

`trunk` and `release/2.1`.

## Notes

Found while working #12588, by a review that swept the workspace for sibling implementations
of the same reader. Different crate and different failure mode from that fix, so it is filed
rather than folded in.

Issue #12596, opened 2026-08-05
spice chat reports a content-filtered answer as complete, because the CLI never reads finish_reason


claudespice:
## Summary

`spice chat` does not deserialize `finish_reason` from a streaming chunk, so a stream that
ended because the model stopped early — rather than because it finished — is presented as a
complete answer.

`bin/spice/src/commands/chat.rs`:

```rust
struct ChunkChoice {
    #[serde(default)]
    delta: Option<Delta>,
}
```

The OpenAI streaming schema puts the reason the generation stopped on the choice:
`finish_reason` is one of `stop`, `length`, `content_filter`, `tool_calls`, or
`function_call`. Only `stop` means the answer is whole.

The two that matter here:

- `content_filter` — Azure OpenAI ends the stream when its filter trips. The tokens already
  sent stay on screen and the CLI exits `0`.
- `length` — the answer hit `max_tokens` and is cut off mid-sentence.

In both cases the user is shown a partial answer with no indication that it is partial, and
`--output json` records it as the full `content`.

## Why it is not the same as #12588

That issue is about events being lost in transit and about failures the server *reports*
being discarded. This one is the opposite: every event arrives and is read correctly, and the
server is not reporting a failure — it is reporting, in a field the CLI does not look at,
that it stopped early. Fixing the framing does not surface it.

## Suggested fix

Deserialize `finish_reason` on `ChunkChoice` and, when the stream ends with anything other
than `stop` (or absent, which is what the intermediate chunks carry), tell the user what
happened after printing what arrived — e.g. `the model stopped early (content_filter); the
answer above is incomplete`. It should not be a hard error: unlike an `error` event, the
tokens received are genuine and worth keeping.

A regression test can drive the existing `SlowServer` fixture with a final chunk carrying
`"finish_reason":"content_filter"` and assert the incompleteness is reported.

## Affected versions

`trunk` and `release/2.1`.

## Notes

Found by adversarial review while working #12588. Pre-existing — it is not introduced or
changed by that fix, which is why it is filed rather than folded in.

PR #12598, opened 2026-08-05
fix(cli): read a streamed answer as events, and a failed body read as a failure (fixes #12588)

### LINKED ISSUE #12587: CLI reads a successful response body with unwrap_or_default, so a truncated response is reported as an empty success
## Summary

Three CLI commands read a successful response body with `.text().await.unwrap_or_default()`, so a
failure to read the body — a timeout that fires mid-response, a reset connection, a truncated
response — is converted into an empty string and then reported as a **successful** result. The
status line was `200`, so nothing downstream treats the empty value as an error.

The sharpest instance is `spice nsql`, which returns the empty body verbatim as the translated SQL:

- `bin/spice/src/commands/nsql/mod.rs:214` — `send_nsql_request` returns `Ok("")`.
- `bin/spice/src/commands/nsql/analyze.rs:346` — `send_nsql_request_with_trace`, same shape.
- `bin/spice/src/commands/search.rs:285` — falls through to `serde_json::from_str("")`, so the user
  gets `Failed to parse response: EOF while parsing a value at line 1 column 0`, an error that
  describes neither the cause nor the endpoint.

The `unwrap_or_default()` calls on the **error** branches (`error.rs:189`, `github/mod.rs:100`, the
`login` sites, and `chat.rs:548`) are fine: there the body is only used to decorate a message that
already reports a failure.

## Reproduction

Observed while working #12583. Point `send_nsql_request` at a server that sends a `200` head and
then dribbles `SELECT 1` one character at a time, under a client whose deadline fires partway
through:

```
assertion `left == right` failed
  left: ""
 right: "SELECT 1"
```

The call returns `Ok`. No error is raised, logged, or shown.

## Why it matters

An empty translation is indistinguishable from a model that legitimately returned nothing, and the
user has no signal that the response was cut off rather than answered. #12583 makes the deadline far
less likely to fire on these paths, but it does not change what happens when it does — nor does it
help the reset-connection or truncated-response cases.

## Suggested fix

On the success branch, propagate the read failure instead of defaulting it: `.context(...)` into the
existing `ConnectionFailedSnafu`/`InvalidResponseSnafu` with the endpoint named, matching the
error-message format in `docs/dev/error_handling.md`. Leave the error branches as they are. The
three sites should move together — this is one shape, not three bugs.

## Affected versions

`trunk` and `release/2.1`.

### LINKED ISSUE #12588: spice chat parses SSE per network chunk, so an event split across two reads is silently dropped
## Summary

`spice chat` parses the streamed SSE response one **network chunk** at a time, treating each
chunk's contents as a whole number of lines. Nothing in HTTP or SSE guarantees that: an event can
be split across two TCP reads, and the CLI then silently drops it.

`bin/spice/src/commands/chat.rs` (in `send_chat_streaming`):

```rust
let text = String::from_utf8_lossy(&chunk);

for line in text.lines() {
    if let Some(data) = line.strip_prefix("data: ") {
        ...
        if let Ok(chat_chunk) = serde_json::from_str::<ChatChunk>(data) {
```

When an event straddles a chunk boundary:

- the leading fragment is a `data: ` line carrying **truncated JSON**, so `serde_json::from_str`
  fails and the `if let Ok(...)` discards it with no error;
- the trailing fragment has no `data: ` prefix, so `strip_prefix` returns `None` and it is
  discarded too.

The token is lost from the printed answer *and* from `full_response`, which is what gets saved to
history and returned by `--output json`. There is no error, no warning, and no exit-code change —
the answer is just quietly missing a piece.

`String::from_utf8_lossy` compounds it: a multi-byte UTF-8 character split across a boundary becomes
U+FFFD rather than being reassembled, so even a correctly-framed chunk pair can corrupt a character.

## Why it is not usually seen

Small responses fit in a single read, and a local runtime on loopback tends to deliver each SSE
event in its own chunk. It gets likelier with larger events (tool-call payloads), a proxy that
re-chunks, a slower network, or TLS record boundaries — that is, in exactly the deployments where it
is hardest to notice.

## Suggested fix

Accumulate the received bytes into a buffer, split off only complete SSE events (terminated by a
blank line), and keep the remainder for the next chunk — the standard SSE framing loop. Decode UTF-8
incrementally rather than per chunk. A regression test can drive the parser with an event split
mid-JSON across two chunks and assert the content survives.

## Affected versions

`trunk` and `release/2.1`.

## Notes

Found by adversarial review while working #12583. Not fixed there: that change is about the request
deadline, and this is a parsing defect that is independently reachable. Note also that the test
server added in that PR aligns each SSE event with one HTTP chunk, so it does not exercise this.

claudespice:
## Summary

Two ways the CLI turned lost data into a result that looks complete. They are one shape —
a read that did not deliver everything, reported as a whole answer — so they move together.

**A streamed answer was parsed one network read at a time.** `send_chat_streaming` took each
chunk from `bytes_stream()` and ran `text.lines()` over it, which assumes a read carries a
whole number of lines. Nothing in HTTP or SSE promises that. When an event straddled two
reads it was discarded twice over: the leading fragment is a `data:` line holding truncated
JSON, which `serde_json::from_str` rejected inside an `if let Ok(..)` that dropped it without
a word; the trailing fragment has no `data:` prefix, so it was skipped. The token disappeared
from the printed answer *and* from `full_response` — what `--output json` returns and what the
history keeps — with no error, no warning and no change of exit code. `String::from_utf8_lossy`
per chunk compounded it: a multi-byte character split across a boundary became U+FFFD.

**The same loop discarded every event it could not read as a chat completion.** That includes
the one the runtime sends when a stream fails part-way through: `create_sse_response` emits
`to_openai_error_event` as an `error`-named event carrying `{"type":…,"message":…}`
(`crates/runtime/src/http/v1/chat.rs`). `ChatChunk` requires `choices`, so that payload never
deserialized and the report of the failure was thrown away — leaving the user a truncated
answer presented as a finished one. Found while fixing the framing; it is the sharpest
instance of the same shape, so it is fixed here rather than filed.

**Three commands read a successful body with `unwrap_or_default`.** A read that stopped
part-way — a deadline firing mid-response, a reset connection, a truncated response — became
an empty string under a `200`. `spice nsql` returned it verbatim as the translated SQL; `spice
search` fell through to `serde_json::from_str("")` and reported `EOF while parsing a value at
line 1 column 0`, which names neither the cause nor the endpoint.

Fixes #12588
Fixes #12587

## Changes

- **`bin/spice/src/sse.rs` (new)** — `SseDecoder` buffers bytes and yields only events it has
  seen the end of, following the field grammar in the HTML standard (`field: value`, `:`
  comments, blank line ends the event, repeated `data:` joined with newlines). Holding bytes
  rather than decoding per read also reassembles a split character: `\n` cannot occur inside a
  multi-byte UTF-8 sequence, so a line boundary is always a character boundary. A single
  unterminated event is capped at 8 MiB so a server that never sends a terminator cannot grow
  the buffer without bound.
- **`chat.rs`** — reads through the decoder; the per-event work moves into `apply_event`. An
  `error` event, and any payload that is not a chat completion, is now reported with the
  server's own reason and a note that whatever was printed is incomplete. An event with no
  payload stays a no-op. At EOF the decoder's tail is dispatched, so a server that hangs up
  without its final blank line does not lose its last event.
- **`error.rs`** — `read_response` returns `(status, body)`, propagating a read failure on a
  success status as a new `ResponseIncomplete` naming the endpoint, and leaving the error
  branches alone: there the body only decorates a message that already reports the failure, so
  a body that could not be read must not replace the status the caller is about to report.
- **`nsql/mod.rs`, `nsql/analyze.rs`, `search.rs`** — the three success-branch reads use it.
- **`test_support.rs`** — `SlowServer` chunks are bytes, so a test can split a character
  across two reads without the fixture replacing it first; adds `truncating` (hang up without
  ending the chunked body) alongside the existing endings.

The `unwrap_or_default()` calls on error branches (`error.rs`, `github/mod.rs`, the `login`
sites, `trace.rs`, `chat.rs`'s non-2xx branch, `crates/repl/src/util.rs`) were checked and left
as they are — each is on a path that already reports a failure.

## What adversarial review changed

The first version of this fix was wrong in four ways that the review caught, all of them now
pinned by neuters:

- **A choice with no `delta` was treated as unreadable.** Azure OpenAI's asynchronous content
  filtering emits annotation choices holding only `content_filter_results`, on passing streams
  as well as filtered ones. Since an unreadable event now stops the stream, that would have
  lost every token after the first annotation — a regression against the old loop, which
  skipped them. `delta` is optional; `choices` remains what separates a completion from an
  error envelope.
- **A frame carrying only `event:` was dispatched.** The standard dispatches on a blank line
  only when the data buffer has been written to. The difference is load-bearing here: an empty
  event reset the progress deadline, so `event: ping` every 20 seconds would have held the CLI
  open indefinitely — defeating the liveness bound #12589 exists to provide.
- **Progress was recorded only when an event completed**, so one multiline event spread across
  reads timed out despite data arriving throughout. It is now recorded when a data line
  arrives; a comment carries no data field, so a keep-alive still cannot hold the stream open.
- **The size cap was checked before draining**, so one read carrying many small whole events
  was rejected as a single unterminated one.

Two framing gaps the review also named are fixed: a leading BOM left the first event empty,
and a lone `\r` was not treated as a terminator — axum writes a payload's own line breaks
through verbatim (`EventDataWriter::write_buf`), and `sse-starlette`, which serves many
OpenAI-compatible endpoints, can be configured to separate with `\r` throughout.

Taking lines by advancing a cursor rather than draining the buffer also made the decoder
linear: the test that puts 8 MiB of small events through one read went from 69s to 0.24s.

## Filed rather than fixed here

- **#12596** — `spice chat` never reads `finish_reason`, so a `content_filter` or `length`
  stop is presented as a complete answer. Every event arrives and is read correctly; the
  server reports the early stop in a field the CLI does not look at, so framing does not
  surface it. Pre-existing.
- **#12597** — `crates/google-genai`'s SSE reader decodes each chunk with
  `std::str::from_utf8`, so a character split across two reads fails the whole stream with
  "Invalid UTF-8". Same class as this fix, different crate.

## What this does not establish

An event that is not a chat completion is now an error rather than a silent skip. Against the
runtime and any OpenAI-compatible server a content event carries `choices` — including Azure's
annotation choices, which is why `delta` is optional — and a payload without it is either an
error envelope or something this client cannot render. A server that invented a new unnamed
event type carrying content would now stop the stream instead of ignoring it. That is the
intended trade: a truncated answer printed as a whole one is the failure being fixed.

`SseDecoder` is hand-rolled although `eventsource-stream` is already in this binary's
dependency tree through `async-openai`. That crate yields whole events and nothing else, and
all three things this decoder is for live outside that: progress before an event ends, the
tail a server leaves unterminated, and a bound on one unterminated event. The module doc says
so, so the next reader does not have to re-derive it.

## Test plan

- `make lint`
- `make nextest`
- 41 new unit tests, 263 passing in the crate. `bin/spice/src/sse.rs` covers the decoder
  directly (split mid-payload, inside a field name, between payload and terminator; a
  character split across reads; several events in one read; comments; names; all three line
  terminators including a CR split from its newline; a BOM, including one split across reads;
  repeated `data:`; the single-leading-space rule; an unterminated tail; the size cap, and a
  read carrying many whole events that is *not* over it; progress before an event ends).
  `chat.rs` covers the wiring through a real HTTP server (`SlowServer`): a split event, a
  split event between two whole ones, a split character, an `error` event, an OpenAI-shaped
  error payload, an Azure annotation choice, an empty payload, a multiline event that outlives
  the progress bound, an unterminated final event. `error.rs` covers both arms of
  `read_response`, and `nsql/mod.rs` covers a truncated `200` end to end.
- **17 neuters, each reverting one direction the fix could be wrong in; every one is caught.**
  Two initially survived and both were the tests being too loose, not the code being
  redundant — the assertions are tightened rather than the neuters dropped.

## Checklist

- [x] Tests added covering the reported failure and the sibling instances
- [x] Regression tests fail on the old code (neuter matrix)
- [x] `cargo clippy -p spice --lib --bins --tests` clean
- [x] `cargo fmt` clean
- [x] No `.unwrap()` in new code
- [ ] `make signoff` / `Attestation`

## Note on the base

Stacked on #12589 (`fix/12583-chat-stream-timeout`), which rewrites the same streaming loop
and adds the `SlowServer` fixture these tests use. The diff self-reduces when that merges.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19175,6 +19175,7 @@ dependencies = [
  "flate2",
  "fundu",
  "futures",
+ "http 1.4.1",
  "humantime",
  "insta",
  "keyring",
--- a/bin/spice/Cargo.toml
+++ b/bin/spice/Cargo.toml
@@ -118,6 +118,7 @@ release = []
 
 [dev-dependencies]
 assert_cmd = "2"
+http.workspace = true
 insta = { workspace = true, features = ["filters", "json"] }
 predicates = "3"
 rstest = "0.26.1"
--- a/bin/spice/src/commands/chat.rs
+++ b/bin/spice/src/commands/chat.rs
@@ -160,7 +160,12 @@ impl StreamErrorPayload {
 /// A choice in a chat chunk.
 #[derive(Deserialize)]
 struct ChunkChoice {
-    delta: Delta,
+    /// Absent on a choice that carries no content. Azure `OpenAI`'s asynchronous content
+    /// filtering emits annotation choices holding only `content_filter_results`, and those
+    /// arrive on a passing stream too -- so a choice without a delta is an ordinary event
+    /// with nothing to print, not an unreadable one.
+    #[serde(default)]
+    delta: Option<Delta>,
 }
 
 /// Delta content in a streaming response.
@@ -619,8 +624,13 @@ async fn send_chat_streaming(
             // The stream ended. A server that closed without its final blank line still has
             // one event's worth of bytes buffered here, and dropping it would lose the tail
             // of the answer.
-            if let Some(event) = decoder.finish() {
-                apply_event(&event, &url, emit_tokens, start_time, &mut state).await?;
+            decoder.close();
+            while let Some(event) = decoder.next_event() {
+                if apply_event(&event, &url, emit_tokens, start_time, &mut state).await?
+                    == EventOutcome::Stop
+                {
+                    break;
+                }
             }
             break 'stream;
         };
@@ -632,25 +642,35 @@ async fn send_chat_streaming(
             .build()
         })?;
 
-        decoder.push(&chunk).map_err(|oversized| {
-            InvalidResponseSnafu {
-                message: format!(
-                    "The model at {url} sent {} bytes of a single response event without ending it. Check the runtime's logs for the model, then retry. See: https://spiceai.org/docs/components/models",
-                    oversized.bytes
-                ),
-            }
-            .build()
-        })?;
+        let data_before = decoder.data_fields_seen();
+        decoder.push(&chunk);
 
         while let Some(event) = decoder.next_event() {
-            // An event, rather than the keep-alive comment that carries no news.
-            last_event = Instant::now();
-
             if apply_event(&event, &url, emit_tokens, start_time, &mut state).await?
                 == EventOutcome::Stop
             {
                 break 'stream;
             }
+        }
+
+        // Progress is a `data` line arriving, not an event completing: a large event spread
+        // over several reads is the model producing, even before its terminator lands. A
+        // keep-alive comment carries no data field, so it still cannot hold the stream open.
+        if decoder.data_fields_seen() != data_before {
+            last_event = Instant::now();
+        }
+
+        // Asked after draining, so the cap bounds one event the stream never ends rather
+        // than however many whole events happened to arrive in a single read.
+        if let Some(oversized) = decoder.oversized() {
+            state.stop_spinner().await;
+            return Err(InvalidResponseSnafu {
+                message: format!(
+                    "The model at {url} sent {} bytes of a single response event without ending it. Check the runtime's logs for the model, then retry. See: https://spiceai.org/docs/components/models",
+                    oversized.bytes
+                ),
+            }
+            .build());
         }
     }
 
@@ -758,7 +778,11 @@ async fn apply_event(
     }
 
     for choice in &chunk.choices {
-        if let Some(content) = &choice.delta.content {
+        if let Some(content) = choice
+            .delta
+            .as_ref()
+            .and_then(|delta| delta.content.as_ref())
+        {
             // Record first token time and stop spinner
             if state.first_token.is_none() {
                 state.first_token = Some(start_time.elapsed());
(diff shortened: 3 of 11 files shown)


claudespice:
@copilot review

lukekim (approved):

claudespice:
Holding this one rather than enqueuing it, for a reason the readiness check does not model.

This PR is stacked: its base is `fix/12583-chat-stream-timeout` (#12589), not `trunk`. `check_pr_merge_readiness.py` reports `ready_to_enqueue` with `blockers=[]`, but the trunk merge queue does not serve a feature-branch base, so there is nothing to enqueue it into — same shape as a PR based on a release branch. It also explains why the only check here is `enforce-pull-with-spice`: a stacked PR gets no `Attestation`, so "all checks green" is not the signal it looks like.

Merging it into #12589 now would also be wrong: that branch has a sign-off in flight on `96b7b4a6c` ([31059131732](https://github.com/spiceai/spiceai/actions/runs/31059131732)), and a merge would invalidate it before it lands.

So the order is: #12589 merges, then this retargets to `trunk` and gets its own sign-off. Nothing to do here until then.

Issue #12600, opened 2026-08-05
The Gemini SSE reader accumulates without a bound, so an endpoint that never terminates an event grows spiced's memory


claudespice:
## Summary

`parse_sse_stream` in `crates/google-genai/src/generate.rs` appends every read into an
accumulation buffer and only releases bytes once it finds the blank line that ends an event.
Nothing bounds that buffer, so a response body that never terminates an event grows it until the
process runs out of memory. `crates/llms` reaches this parser through
`crates/llms/src/google/chat.rs`, so the growth happens inside `spiced`, not just in a CLI.

## Why it is filed separately

Noticed while fixing #12597, which is about the *decode* landing on a character boundary. The
buffer has been unbounded since the parser was written - the previous version accumulated into a
`String` the same way - so this is neither introduced nor fixed by that change, and a cap is a
behavior decision rather than a correctness repair.

## What a fix has to decide

`bin/spice/src/sse.rs` caps its equivalent buffer, so the runtime already has a precedent for the
shape of the fix: measure the cap against the *unterminated remainder* after draining whole
events, not against a single read, or one read carrying many small whole events reads as one
oversized event.

The open question is the limit itself. A Gemini event is normally a few hundred bytes of JSON, but
a candidate part can carry inline data, so a cap chosen too low turns a valid long response into a
failure - which is the same class of bug as the one being avoided. Whoever picks the number should
say what response it is sized for.

## Affected versions

`trunk` and `release/2.1`.

PR #12601, opened 2026-08-06
fix(google-genai): read a Gemini SSE stream as bytes, so a split character survives (fixes #12597)

### LINKED ISSUE #12597: Gemini streaming fails with 'Invalid UTF-8' when a character is split across two reads

claudespice:
## Summary

`parse_sse_stream` decoded each network read on its own with `std::str::from_utf8`. A multi-byte
character split across two reads is not invalid UTF-8 — it is half of a character, with the other
half in the next read — but decoding the halves separately made it look invalid and failed the whole
stream with `Invalid UTF-8`, blaming the model for bytes that were fine. Any answer containing
accented text, CJK or an emoji could hit it, on a boundary that moves with the link and any proxy in
between, which is why it looked intermittent. `crates/llms/src/google/chat.rs` reaches this parser,
so it surfaced mid-answer on the runtime's Gemini chat path.

The same per-read handling normalized `\r\n` inside one read, so a terminator split across two reads
never formed an event separator either.

Frame in the bytes instead: find the blank line that ends an event, then decode only that event. A
UTF-8 sequence never contains `\n` or `\r`, so the decode always lands on a character boundary, and
an `Invalid UTF-8` error now means the bytes really are invalid.

Fixes #12597

## Changes

All in `crates/google-genai/src/generate.rs`. Four further reads that did not deliver everything
they held — the same shape as the reported bug — came out of the same function and are fixed with
it, rather than left for follow-ups:

- **A dataless event sent the loop back to the network.** A comment or keep-alive was taken from the
  buffer, found to carry no `data:` field, and then awaited another read — holding back a data event
  already sitting behind it until the server happened to send more, or, at the end of the body,
  reporting `Unexpected end of stream` and dropping it. The buffer is now drained of every event it
  holds before another read is awaited.
- **Framing no longer depends on whether more bytes may arrive.** A blank line may end with a bare
  `\r`, and whether that `\r` stays bare or turns out to be a `\r\n` cannot change that the line —
  and so the event — is complete; it decides only where the *next* event begins. So the framer
  answers from the bytes alone and carries that single unresolved bit: a `\n` read immediately after
  such a terminator completes it rather than starting a line. A complete event is therefore never
  withheld, and a terminator split across two reads leaves no stray byte to be mistaken for an
  unterminated event.
- **An event's later `data:` fields were discarded** rather than joined with `\n` as the SSE grammar
  defines, and a `data:` field written without the optional space after the colon was not read at
  all.
- **The truncated-tail error was returned with the tail still buffered**, so every later poll
  produced it again. Consumers forward each item rather than stopping at the first error
  (`convert_google_stream_to_openai` is a `map`), so that stream never ended. It is now reported
  once and the stream ends — and the end of the body is tracked so a stream that has finished is not
  polled a second time, which the `Stream` contract permits an implementation to panic on.

A comment records why the reader is written here rather than taken from `eventsource-stream` or the
`reqwest-eventsource` fork `crates/llms` pins: both end a stream silently when the body stops
mid-event, and reporting that instead of passing a partial answer off as a whole one is the behavior
this module exists to keep. Sharing one decoder across the workspace's SSE readers is still worth
doing — #12597 says so — but as its own change.

## Test plan

- `make lint` — clean, including the `--tests` pass.
- `make nextest` — full unit suite.
- 19 unit tests in the module (16 new), all green; the 3 pre-existing ones are unchanged and pass.
- **Neuter matrix: 13 reverts, every one caught by a named test.** Each element of the fix was
  reverted in turn — the per-read decode, each of the three line-terminator cases, the possible-`\n`
  carry (both setting it and consuming it), the dataless-event drain, the `data:`-field join, the
  optional space, the `data:` prefix, the drain-before-error, the tail clear, and acting on the end
  of the body immediately — and in every case a test failed.
- The new tests drive the parser through a stream that hands out exactly the byte groups given, so a
  boundary can be placed inside a two-byte character, inside a four-byte character across three
  reads, and inside a `\r\n` pair. Two drive a stream that stays *pending* after its reads, which is
  the shape a server that just sent a keep-alive has — that is what pins an event being delivered
  rather than held back, and an earlier version of one of them passed either way because it let its
  stream end. One drives a stream that panics if polled after it finishes, so "a body that has ended
  is not read again" is tested rather than assumed.

## Review

Two rounds of Codex adversarial review, both of which found real defects that are fixed here:

1. The first version accepted a bare-`\r` blank line mid-stream but not at the end of the body, so a
   standards-valid CR-framed final event was reported as truncated.
2. Making that end-of-body-aware then withheld the same complete event for as long as a live
   connection stayed open. That is what drove the framer to stop consulting the end of the body at
   all — the shape it should have had from the start.

`/security-review`: no findings. The untrusted input is the model endpoint's response body; the
indices are provably in range (`end` is the scan index, and `next_start` cannot exceed the buffer
length), the error text carries a UTF-8 offset and length rather than payload bytes, and the payload
still deserializes into the same typed struct.

## Deferred

- #12600 — the accumulation buffer has no bound, so an endpoint that never terminates an event grows
  the process's memory. Pre-existing (the previous version accumulated into a `String` the same way),
  and the cap is a behavior decision: it has to be sized against a response carrying inline data or
  it turns a valid long answer into a failure.

## Checklist

- [x] Regression test that fails on the old code and passes on the new
- [x] Whole bug class in this function fixed, not just the reported instance
- [x] No `unwrap`/`expect` in production code
- [x] Adversarial review and security review run before the PR was opened
- [x] Lint clean, formatted, unit suite green

### DIFF
--- a/crates/google-genai/src/generate.rs
+++ b/crates/google-genai/src/generate.rs
@@ -194,45 +194,84 @@ impl Client {
     }
 }
 
-/// Length of the line terminator beginning at `buf[i]`, or `None` if one does not begin there.
-///
-/// A `\r` last in `buf` is only a terminator once `at_eof` says no more bytes can arrive to make it
-/// the first half of a `\r\n`; until then which one it is cannot be known.
-fn terminator_len(buf: &[u8], i: usize, at_eof: bool) -> Option<usize> {
-    match buf.get(i)? {
-        b'\n' => Some(1),
-        b'\r' if buf.get(i + 1) == Some(&b'\n') => Some(2),
-        b'\r' if i + 1 < buf.len() || at_eof => Some(1),
-        _ => None,
+/// What begins at `buf[i]`.
+enum Terminator {
+    /// A `\n`, or a `\r\n` whose `\n` is present: the line ends here and the next one starts a
+    /// known number of bytes later.
+    Complete(usize),
+    /// A `\r` last in `buf`. The line ends here either way - a `\r` alone terminates it, and so
+    /// does a `\r\n` - but whether the next line starts one byte or two later is not known until
+    /// the byte after it arrives.
+    TrailingCr,
+    /// Not a line terminator.
+    None,
+}
+
+fn terminator_at(buf: &[u8], i: usize) -> Terminator {
+    match buf.get(i) {
+        Some(b'\n') => Terminator::Complete(1),
+        Some(b'\r') if buf.get(i + 1) == Some(&b'\n') => Terminator::Complete(2),
+        Some(b'\r') if i + 1 < buf.len() => Terminator::Complete(1),
+        Some(b'\r') => Terminator::TrailingCr,
+        _ => Terminator::None,
     }
 }
 
-/// Byte offsets `(event_end, next_event_start)` of the blank line that terminates the first event
-/// in `buf`, or `None` while more bytes could still complete one.
+/// Where the first event in a buffer ends.
+#[derive(Debug, PartialEq, Eq)]
+struct EventBoundary {
+    /// Offset just past the event's last field line - the start of the blank line that ends it.
+    end: usize,
+    /// Offset the next event starts at.
+    next_start: usize,
+    /// The blank line ended with a `\r` last in the buffer, so a `\n` read next belongs to that
+    /// terminator rather than starting a line. The event itself is complete regardless.
+    may_skip_lf: bool,
+}
+
+/// The blank line that terminates the first event in `buf`, or `None` if the buffer does not hold a
+/// complete event yet.
 ///
 /// A line ends with `\n`, `\r\n` or `\r`, so the blank line is any two consecutive terminators.
-/// Searching the bytes - rather than decoding each network read as text and searching that - is
-/// what makes the decode below land on a character boundary: a multi-byte UTF-8 sequence never
-/// contains `\n` or `\r`, so an event never ends inside one.
-fn find_event_boundary(buf: &[u8], at_eof: bool) -> Option<(usize, usize)> {
+/// The answer never depends on whether more bytes may arrive: an event that is complete is reported
+/// immediately, so a live connection can never withhold one. Searching the bytes - rather than
+/// decoding each network read as text and searching that - is what makes the decode below land on a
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

Issue #12625, opened 2026-08-06
The models integration job stops at its first failing suite, so one red step hides the eight behind it


claudespice:
## Summary

`integration_models.yml` runs its suites as a sequence of plain steps. None of the later ones carries an `if:` that survives a failure, so the **first** suite to fail cancels every suite behind it and the job reports one failure where there may be nine.

In the 2026-08-06 nightly ([31059672350](https://github.com/spiceai/spiceai/actions/runs/31059672350)), step 9 `Run OpenAI integration test` failed and steps 10-18 were all skipped:

```
 9  failure  Run OpenAI integration test
10  skipped  Run OpenAI Beta functionality embeddings test
11  skipped  Run AI UDF basic test
12  skipped  Run AI UDF with dataset test
13  skipped  Run AI UDF LEFT truncate test
14  skipped  Run local embeddings Beta functionality test
15  skipped  Run AI UDF with local model test
16  skipped  Run workers integration test
17  skipped  Run search integration test
18  skipped  Push snapshots to branch
```

## Impact

- **A suite's state becomes unobservable without anyone noticing.** The search suite has not run since 2026-08-04. #11112 tracks that suite and its remaining question — whether #12389 fixed the last three failures — is unanswerable while an unrelated step upstream is red.
- `Push snapshots to branch` is the last step, so a single failing suite also suppresses the snapshot refresh the scheduled run exists to produce.
- The job's conclusion under-reports. One red step is indistinguishable from nine, so the first fix looks like the whole fix.

This is a reporting/structure problem on top of whatever makes a suite red; #12624 is the current such cause.

## Expected behavior

Every suite runs, and the job fails if any of them failed. The usual shapes:

- give each test step `if: ${{ !cancelled() && <its existing condition> }}` and let the job's conclusion aggregate, or
- run the suites as a matrix so each is its own job and reports its own conclusion (also parallelises them), keeping `Push snapshots to branch` in a dependent job with `if: always()`.

Either way a skipped step should mean "not applicable here", not "something earlier broke".

PR #12626, opened 2026-08-06
fix(ci): keep the build cache's S3 endpoint out of the runtime under test (fixes #12624)

### LINKED ISSUE #12624: The build cache's AWS_ENDPOINT_URL reaches spiced in the models integration tests, so every s3:// dataset resolves to the local cache
## Summary

`spiceai/spiceio`'s setup action exports `AWS_ENDPOINT_URL` into `$GITHUB_ENV` so that later steps (sccache, the AWS CLI) reach the local S3-compatible cache. In `integration_models.yml` that export survives into the steps that run the AI integration test binary, and `spiced` honours it: `AmazonS3Builder::from_env()` maps `AWS_ENDPOINT_URL` to the S3 endpoint, so **every `s3://` dataset in these tests is fetched from the build cache instead of AWS**.

The dataset never loads, the runtime never reaches ready, and the tests fail on a readiness timeout that says nothing about S3.

All three test jobs of the nightly — `Test Models, AI & Search`, `Test Hugging Face Models`, `Test Bedrock Models` — have failed this way on every run since 2026-08-05.

## Evidence

The failing request, from run [31059672350](https://github.com/spiceai/spiceai/actions/runs/31059672350) (`Test Models, AI & Search`):

```
WARN runtime::init::dataset: Cannot connect to the dataset taxi_trips (s3). Object Store error:
  Generic S3 error: Error performing list request:
  Error performing GET http://127.0.0.1:8333/?list-type=2&prefix=taxi_trips%2F2024%2F - HTTP error: builder error
INFO runtime::init::dataset: Dataset load summary (after 30s): 1/2 ready, 1 unhealthy, 0 still initializing.
Error: Timed out waiting for components to load
```

`http://127.0.0.1:8333` is spiceio's bind address, and the bucket is missing from the URL entirely — with no explicit endpoint the registry builds the store virtual-hosted, and `object_store`'s builder then uses the endpoint verbatim as the bucket endpoint (`bucket_endpoint = endpoint.clone()` when `virtual_hosted` is true).

The job's own environment dump in the same run:

```
SCCACHE_ENDPOINT: http://127.0.0.1:8333
AWS_ENDPOINT_URL: http://127.0.0.1:8333
```

## Attribution

The export is new. It is absent from the setup action at v0.5.9 and present from v0.5.10, which is also exactly where the nightly's failure mode changes:

| nightly | pinned setup action | first failing step | `AWS_ENDPOINT_URL` in the job env |
|---|---|---|---|
| 2026-08-04 | v0.5.9 (`da41a0d`) | 17 — `Run search integration test` | absent |
| 2026-08-05 | v0.5.10 (`dc53e1f`) | 9 — `Run OpenAI integration test` | present |
| 2026-08-06 | v0.6.0 (`0870da5`) | 9 — `Run OpenAI integration test` | present |

`trunk` now pins v0.7.0, which still exports it.

## Blast radius

Beyond the `s3://` datasets, `AWS_ENDPOINT_URL` is an SDK-wide override, so it also points the S3 Vectors calls in the search suite at the cache.

Because the test steps are sequential and none is conditioned on `always()`, the step-9 failure also **skips steps 10-18** — the AI UDF, local embeddings, workers and search suites, and `Push snapshots to branch`. The search suite has therefore not run at all since 2026-08-04, which is why #11112 cannot be confirmed resolved.

## Expected behavior

A step that runs `spiced` resolves `s3://` against AWS. The compiler cache's endpoint belongs to the build steps that need it, and must not be inherited by the runtime under test.

claudespice:
Fixes #12624

## Summary (root cause)

`spiceai/spiceio`'s setup action writes `AWS_ENDPOINT_URL` into `$GITHUB_ENV` so the steps after it — sccache, the AWS CLI — reach the local cache. In `integration_models.yml` that export is still in the environment when the AI integration test binary runs, and `spiced` honours it: `AmazonS3Builder::from_env()` maps `AWS_ENDPOINT_URL` to the S3 endpoint, so every `s3://` dataset with no endpoint of its own is fetched from the compiler cache instead of AWS. With no explicit endpoint the registry builds the store virtual-hosted, and `object_store` then uses that endpoint verbatim as the bucket endpoint — which is why the bucket disappears from the request:

```
WARN runtime::init::dataset: Cannot connect to the dataset taxi_trips (s3). Object Store error:
  Generic S3 error: Error performing list request:
  Error performing GET http://127.0.0.1:8333/?list-type=2&prefix=taxi_trips%2F2024%2F - HTTP error: builder error
INFO runtime::init::dataset: Dataset load summary (after 30s): 1/2 ready, 1 unhealthy, 0 still initializing.
Error: Timed out waiting for components to load
```

Nothing in that failure names S3. The dataset stays unhealthy, the runtime never reports ready, and the test fails on a readiness timeout — six times over, once per nextest retry.

The export is new, and the nightly's failure mode changes exactly where it appears:

| nightly | pinned setup action | first failing step | `AWS_ENDPOINT_URL` in the job env |
|---|---|---|---|
| 2026-08-04 | v0.5.9 (`da41a0d`) | 17 — `Run search integration test` | absent |
| 2026-08-05 | v0.5.10 (`dc53e1f`) | 9 — `Run OpenAI integration test` | present |
| 2026-08-06 | v0.6.0 (`0870da5`) | 9 — `Run OpenAI integration test` | present |

All three test jobs of the nightly — `Test Models, AI & Search`, `Test Hugging Face Models`, `Test Bedrock Models` — have failed this way on every run since.

## Changes

- **`.github/workflows/integration_models.yml`** — the twelve steps that run the test binary drop `AWS_ENDPOINT_URL` first. The build steps keep it: compiling is what it is for.
- **`.github/workflows/integration_llms.yml`** — the same one-line drop. This job is green today, but it sets the cache up and then executes a downloaded test binary whose hosted matrix includes Bedrock, and the override is SDK-wide rather than S3-only. Fixing it here rather than waiting for it to break is the point of finding the class instead of the instance.
- **`.github/scripts/check_workflow_yaml.py`** — a fourth invariant for the existing Actions-definition gate: in a job that sets the compiler cache up, a step *after* that setup which starts the Spice runtime (`cargo nextest run`, `spice run`, or a downloaded `./integration_test/…` binary) must drop `AWS_ENDPOINT_URL` before it. Without this the two fixes above last exactly until someone adds a thirteenth test step.
- **`.github/scripts/test_check_workflow_yaml.py`** — twelve cases for the new check.

`unset` is the only form that works. Writing an empty value back to `$GITHUB_ENV` leaves the variable *set*, and `object_store` then builds a store whose bucket endpoint is the empty string — a different failure, not a fix. The comment above `jobs:` records that.

## What the guard does and does not prove

It requires the drop to be a top-level command that precedes the runtime, so an `unset` after the run, one nested inside an `if`, and `unset -f` (which removes a function, not the variable) are all rejected; a command named only in a comment is not treated as an invocation, and a runtime step *before* the setup is left alone. Each of those is a test.

It does not model shell semantics, so a drop that is later re-exported still passes, and it cannot see a runtime started from a `make` target or a composite action, where the command lives outside the workflow file. No job combines those with the cache setup today. The docstring says so rather than implying the guard is a proof.

## Why here and not at the source

The export is deliberate on spiceio's side and has no opt-out input, and other consumers rely on it for the AWS CLI. Ignoring an ambient `AWS_ENDPOINT_URL` in the runtime is not an option either — honouring it is standard AWS behaviour and is how users point Spice at MinIO or LocalStack. So the isolation belongs to the job that runs both a build cache and a runtime. The upstream half — making the generic export opt-in, since `AWS_ENDPOINT_URL` is not tool-scoped the way `SCCACHE_ENDPOINT` is — is filed as spiceai/spiceio#38; this change stands on its own either way.

## Not in scope

The step-9 failure also skipped steps 10-18, including the search suite and `Push snapshots to branch`. That masking is a separate defect in how the job is structured and is filed as #12625. It is also why #11112 cannot be confirmed resolved: the suite it tracks has not run since 2026-08-04.

## Test plan

No Rust changes, so there is nothing here for `make lint` / `make nextest` to cover and `Attestation` auto-passes a branch with no Rust-affecting files. What was run:

- `python3 .github/scripts/test_check_workflow_yaml.py -v` — 54 tests pass (42 pre-existing, 12 new).
- `python3 .github/scripts/check_workflow_yaml.py` — `OK: 108 GitHub Actions definitions are well-formed`.
- Neuter, per workflow: reverting only `integration_models.yml` reports **12** problems, one per unguarded step; reverting only `integration_llms.yml` reports **1**, naming that job's step. Both are 0 with the fixes applied.
- `Workflow YAML Check` runs on this PR through its existing `.github/workflows/**` path filter.

End-to-end confirmation is the next `integration tests (models)` run: `Run OpenAI integration test` reaching a real verdict instead of a readiness timeout, and steps 10-18 executing at all.

## Checklist

- [x] Root cause identified and stated, with the change that introduced it
- [x] Regression guard that fails on each pre-fix workflow and passes on both
- [x] Same class fixed in the second exposed workflow, not just the reported one
- [x] Guard scoped so build steps keep the cache endpoint they need
- [x] Guard's limits documented rather than implied away
- [x] No secrets, credentials or internal hostnames added to the workflows, the scripts, or this description

### DIFF
--- a/.github/scripts/check_workflow_yaml.py
+++ b/.github/scripts/check_workflow_yaml.py
@@ -21,6 +21,11 @@
 # And it rejects an `if:` naming a context GitHub does not make available there,
 # which fails the same way for a different reason: GitHub rejects the definition
 # rather than the expression, so the workflow never schedules (#12396).
+#
+# Finally it rejects a step that runs the Spice runtime while the compiler
+# cache's `AWS_ENDPOINT_URL` is still in the job's environment, which redirects
+# the runtime's S3 traffic to the cache and reports as a dataset that never
+# loads (#12624).
 """Validate the repository's GitHub Actions workflow and composite action YAML."""
 
 from __future__ import annotations
@@ -89,6 +94,42 @@ CONTEXT_REMEDIES = {
     ),
 }
 
+# The compiler-cache proxy's setup action, which exports `AWS_ENDPOINT_URL` into
+# `$GITHUB_ENV` for the rest of the job.
+CACHE_SETUP_ACTION = "spiceio/.github/actions/setup"
+
+# Ways a step starts the Spice runtime from the workflow file: `nextest`, a test
+# binary the job downloaded rather than built, and the CLI. All three inherit the
+# step's environment, so all three see the cache endpoint. Optional `KEY=value`
+# prefixes are skipped so the binary form still reads as the command.
+#
+# What this cannot see: a `make` target or a composite action that starts the
+# runtime, because the command is in the Makefile or the action, not here. No job
+# combines those with the cache setup today. If one comes to, the guard has to
+# grow rather than be trusted.
+RUNTIME_INVOCATION = re.compile(
+    r"\bcargo\s+nextest\s+run\b"
+    r"|\bspice\s+run\b"
+    r"|^(?:\S+=\S+[ \t]+)*\./\S*integration_test\S*",
+    re.MULTILINE,
+)
+
+# Dropping it for the step's children. `unset` is the only form that works:
+# writing an empty value back to `$GITHUB_ENV` leaves the variable *set*, and
+# `object_store` then builds a store whose bucket endpoint is the empty string.
+#
+# The command has to sit at the top level of the script — indentation means it is
+# inside an `if` or a function and may not run — must not be `unset -f`, which
+# removes a function of that name and leaves the variable alone, and anything
+# after a `#` is its comment rather than its argument.
+DROPS_CACHE_ENDPOINT = re.compile(
+    r"^unset[ \t]+(?!-)[^\n#]*\bAWS_ENDPOINT_URL\b", re.MULTILINE
+)
+
+# A line whose first non-blank character is `#`. Removed before looking for a
+# runtime invocation, so a command named in a comment does not read as one.
+COMMENT_LINE = re.compile(r"^[ \t]*#[^\n]*$", re.MULTILINE)
+
 
 def workflow_files(github_dir: Path) -> list[Path]:
     """Return the workflow definitions under `<github_dir>/workflows`, sorted."""
@@ -136,6 +177,7 @@ def check_workflow(text: str) -> list[str]:
     else:
         problems.extend(check_step_budgets(jobs))
         problems.extend(check_workflow_conditions(jobs))
+        problems.extend(check_cache_endpoint_isolation(jobs))
 
     return problems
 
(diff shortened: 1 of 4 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12639, opened 2026-08-06
fix(cloud-connect): send the credentials for `spice connect remove`

peasee:
## What

`spice connect remove` posted `{ "instance_id": "..." }` to `/v1/cloud-connect/release` — a body with none of the fields the endpoint authorizes on, so the release could never be accepted. An instance id the caller picks for itself proves nothing.

The request carries the same credential shape `/renew` uses:

- `cert_pem` — the instance's current leaf. Its SPIFFE SAN names the instance, which is the one statement of identity a caller cannot choose for itself, so `instance_id` is no longer sent at all.
- `pop_sig` — an ECDSA P-256 signature made with that leaf's private key over `release\n{instance_id}`. A certificate is not a secret, so the signature is what proves the caller is the instance. The `release` prefix keeps the signature from being replayable as a renewal, and the instance id keeps it from being replayable for another instance.

The credential is sent in the body rather than the TLS layer for the same reason renewal does: a host being decommissioned may already have an expired leaf.

## Also

- The renewal signer now takes payload bytes, so renewal and release share one implementation instead of duplicating the crypto.
- Signing moved ahead of the HTTP client build, so an unusable private key is reported as a proof-of-possession failure that names the instance, not as a TLS setup error.
- The module doc and the request-site comment said the mTLS leaf authorized the call and that the instance id was sent so the cloud could reject a mismatch. Neither is true under this contract, so both are corrected rather than left to teach the wrong model. The leaf is still presented as a TLS client identity, now labelled as transport.

## Tests

The release path had no wire-shape coverage, which is how a body with zero required fields shipped. Added:

- An in-process mock of the release endpoint asserting what actually goes out: the leaf, no `instance_id`, and a signature that verifies against the leaf's public key over the payload the cloud rebuilds — and does not verify for a different instance id.
- Unit tests pinning the payload separator to a real newline byte (an escaped `\n` would be two bytes the cloud never signs) and the request to exactly its two fields.

### DIFF
--- a/crates/runtime-cloud-connect/src/release.rs
+++ b/crates/runtime-cloud-connect/src/release.rs
@@ -63,11 +63,14 @@ pub enum Error {
     ClientBuild { source: reqwest::Error },
 
     #[snafu(display(
-        "The local identity cannot be presented as a TLS client certificate: {source}. \
-         The identity file is corrupt — clear it with `spice connect remove` and delete the \
-         instance in the Spice Cloud portal."
+        "Failed to release instance {instance_id} (Spice Cloud): the local identity file is \
+         invalid ({source}). Delete this instance in the Spice Cloud portal to finish removing \
+         it. See: https://spiceai.org/docs"
     ))]
-    Identity { source: reqwest::Error },
+    Identity {
+        instance_id: String,
+        source: reqwest::Error,
+    },
 
     #[snafu(display("Invalid CA certificate PEM for the Spice Cloud release request: {source}"))]
     CaCert { source: reqwest::Error },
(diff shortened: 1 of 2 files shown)


crates/runtime-cloud-connect/src/release.rs
```diff
@@ -71,10 +84,41 @@ pub enum Error {
 
     #[snafu(display("Spice Cloud rejected the release request ({status}): {message}"))]
     Rejected { status: u16, message: String },
+
+    #[snafu(display(
+        "Failed to release instance {instance_id} (Spice Cloud): the stored identity's private \
+         key could not sign the release proof-of-possession ({reason}). The identity file is \
+         unusable, so the cloud cannot be told this instance is gone — delete the instance in \
+         the Spice Cloud portal instead. See: https://spiceai.org/docs"
```
phillipleblanc: This is a technical explanation that could be improved to be more user-friendly.

phillipleblanc: i.e. `proof-of-possession` doesn't need to be there

phillipleblanc (approved):

PR #12640, opened 2026-08-06
fix(cloud-connect): Tolerate missing spicepods

peasee:
## 📝 Summary

* Normally, Runtime will error and exit on startup if `spicepod.yaml` is missing. This makes installations from `spice connect --install` start crashlooping forever because no spicepod exists
* Updates Runtime so that when started with `--cloud-connect`, it tolerates a missing spicepod and starts up with an empty/default spicepod definition instead. It logs a warning when it detects no spicepod on startup, in case this is not expected.

### DIFF
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -1028,8 +1028,26 @@ pub async fn run(args: Args, app_bundle: AppBundle) -> Result<()> {
     result
 }
 
-/// What `build_app` decided about the deployed spicepod, logged once tracing
-/// exists.
+/// Whether a failed local spicepod load should be tolerated by starting on an
+/// empty spicepod instead of exiting.
+///
+/// Three conditions, each excluding a case where exiting is still the right
+/// answer:
+///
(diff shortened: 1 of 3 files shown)


bin/spiced/src/lib.rs
```diff
@@ -1028,8 +1028,26 @@ pub async fn run(args: Args, app_bundle: AppBundle) -> Result<()> {
...
+/// answer:
+///
+/// - `--cloud-connect` only. A cloud-managed instance gets its configuration
+///   from a deployment, so having none yet is the ordinary freshly-connected
+///   state. Anywhere else, a missing spicepod means the user is in the wrong
+///   directory and should be told so.
+/// - No explicit path. `spiced --spicepod <path>` names a file the user expects
+///   to exist; an absent one is a typo, not a fresh connection.
+/// - Absent, not unloadable. A spicepod that exists but does not parse is still
+///   fatal — coming up empty over a YAML typo would hide it.
```
```suggestion
```

phillipleblanc (approved):

Issue #12644, opened 2026-08-06
test_openai_model's 5-minute ceiling is below its own setup cost, so E2E is cancelled and PRs are dequeued


grokspice:
## Summary

`e2e_test_ci.yml`'s `test_openai_model` job carries `timeout-minutes: 5`. On the
self-hosted macOS pool that ceiling is below the job's own fixed setup cost, so
the job is cancelled partway through its first test step and the E2E gate counts
the cancellation as a failure — dequeuing whatever PR was in the queue.

The budget breaks down like this on a contended runner:

| phase | bound |
|---|---|
| checkout + download the build artifact + `install-spice` | several minutes, none of it test work |
| `Wait for Spice runtime is ready` | up to 60s (its own timeout) |
| `Test the expect scripts driving the REPLs` | ~1 minute |
| `Test vector search`, `Test chat` | the expect scripts' own 30s / 30s / 60s waits |

The two expect scripts alone are allowed more than two minutes of bounded waiting,
before any setup. Five minutes for all of it was never enough.

This is not a job that got slower and needs its regression hidden: the fast path
finishes in well under the limit, and every observed cancellation was consumed by
setup, with the test steps getting seconds. The sibling `test_hf_model` job — same
runners, same artifact download, same `install-spice`, and it additionally pulls a
model — has always had `timeout-minutes: 10` and does not hit it.

## Evidence

`openai model (darwin-aarch64)` was cancelled at its 5-minute ceiling on multiple
runs on 2026-08-06, each of which then failed the required `E2E Test CI` gate:

- https://github.com/spiceai/spiceai/actions/runs/31088421790 — job step timeline
  shows `download artifacts` and `Install spice` consuming almost the entire
  budget, `Test the expect scripts driving the REPLs` starting with seconds left,
  then `##[error]The operation was canceled.`; `Test vector search` and `Test chat`
  never ran (`skipped`).
- https://github.com/spiceai/spiceai/actions/runs/31081140526
- https://github.com/spiceai/spiceai/actions/runs/31110027058

Successful runs of the same job in the same window finish comfortably inside a
10-minute ceiling.

## Expected

`test_openai_model` gets a ceiling above its setup plus the bounded waits of the
steps it runs — matching `test_hf_model`, which has the same shape on the same
runners.

## Not in scope

The inner timeouts stay exactly as they are: readiness is still bounded at 60s and
each expect script still bounds its own waits, so a genuinely hung runtime still
[...]

Issue #12643, opened 2026-08-06
The E2E Test CI gate cannot name which E2E job failed, so its error is unactionable


grokspice:
## Summary

`e2e_test_ci.yml`'s fan-in job `e2e-gate` (display name **E2E Test CI**) is the
single required merge-queue context standing in for the whole E2E matrix. When any
of the ~20 jobs it depends on fails, its entire output is:

```
##[error]One or more E2E Test CI jobs did not succeed
```

It cannot say more than that, because `contains(needs.*.result, 'failure')`
answers only *whether* something failed, never which job did. So the one line a
dequeued PR's author sees names nothing: triaging it means opening the run and
reading down a 20-job list to find the red one, on every occurrence.

`toJSON(needs)` already carries each job's result in the same context, so naming
them costs no API call and no token.

## Evidence

`E2E Test CI` → job `E2E Test CI` → step `Verify all E2E jobs succeeded`, over
2026-08-04 → 2026-08-06: 9 failed runs across 7 distinct branches, on both
`merge_group` and `push` to `trunk`. Every one of them produced exactly the line
above and nothing else.

https://github.com/spiceai/spiceai/actions/runs/31088421790 ·
https://github.com/spiceai/spiceai/actions/runs/31081140526 ·
https://github.com/spiceai/spiceai/actions/runs/31075749056

For the last of those, the answer the reader wanted was `openai model
(darwin-aarch64)`; the gate had that fact in `needs` and discarded it.

## Expected

The gate names the jobs that did not succeed, and their results, in its
`::error::` annotation — so the run summary alone says where to look.

It must keep failing closed: a gate that cannot read the results (no `jq`,
unparseable input) has to fail rather than report a pass it did not verify.

## Not in scope

Nothing about *which* jobs are required, and no change to what makes the gate
pass — `success` and `skipped` still pass, `failure` and `cancelled` still fail.
Only the diagnosis changes.

PR #12648, opened 2026-08-06
fix(ci): recognise the Flight bind wording so a foreign ready cannot pass (fixes #12642)

### LINKED ISSUE #12642: E2E readiness guard misses the Flight bind-failure wording, so a port collision passes as ready
## Summary

`.github/actions/wait-for-spice-ready` has a guard whose whole job is to refuse a
`ready` answered by **another job's** runtime. Several self-hosted macOS runner
instances share one physical machine — and therefore one loopback — so when this
job's `spiced` loses the bind, a sibling runner's runtime can answer
`http://localhost:8090/v1/ready` and every test step afterwards silently runs
against *that* runtime.

The guard works by looking in our own `spice.log` for a bind failure and treating
that as authoritative over whatever answered the socket. It greps for the literal
string `Address already in use` — **which the failing path does not emit.**

- The **HTTP** server passes the OS error through, so it reads
  `Unable to bind to address: Address already in use (os error 48)` — matched.
- The **Flight** server formats its own message with the address in the middle:
  `Address 127.0.0.1:50051 is already in use by another process.`
  (`crates/runtime/src/flight/mod.rs`) — **not matched.**

Flight is the one that fires in CI, because :50051 is the port a sibling
runtime is already holding. So the guard never fires, `wait-for-spice-ready`
reports success in a few seconds, and the failure surfaces minutes later from the
`Test chat` step as `Model did not return expected response` — a model/product
failure, about a model belonging to a different job.

The existing unit tests in `scripts/test_e2e_spice_lifecycle.sh` cover only the
HTTP wording, which is why the gap survived.

## Evidence

`E2E Test CI` → job `E2E Test CI` → step `Verify all E2E jobs succeeded`, over
2026-08-04 → 2026-08-06: 9 failed runs across 7 distinct branches, including both
`merge_group` (dequeues) and `push` on `trunk`. In every one the underlying
failure was a `model (darwin-aarch64)` job on the shared macOS pool.

Worked example — https://github.com/spiceai/spiceai/actions/runs/31075749056
(`push` on `trunk`):

- `huggingface model (darwin-aarch64)` on runner `…-mbp-runner-01-01` — success
- `openai model (darwin-aarch64)` on runner `…-mbp-runner-01-02` — failure

Same machine (`-01`), two runner instances, overlapping in time. The openai job's
own `spice.log`, printed by `stop-spice` at the end of the job, contains:

```
ERROR spiced: Unable to start Spice Runtime servers: Unable to start Flight server: Address 127.0.0.1:50051 is already in use by another process. Either stop the existing process or change the address: https://spiceai.org/docs/cli/reference/run
```

`Wait for Spice runtime is ready` nevertheless passed, and the job went red four
minutes later on `Test chat` with `Model did not return expected response`.

Other runs in the same cluster:
https://github.com/spiceai/spiceai/actions/runs/31088421790 ·
https://github.com/spiceai/spiceai/actions/runs/31081140526 ·
https://github.com/spiceai/spiceai/actions/runs/31066186618

## Expected

`wait-for-spice-ready` recognises both of the runtime's bind-failure wordings, so
a port collision fails immediately, in the step that owns it, named as a
collision — instead of passing and being re-reported later as a model failure.

## Not in scope

Making the collision impossible (per-job ports, or serialising the runtime-owning
jobs across runner instances on one machine) is a separate, larger change. This
issue is only about the guard failing to fire.

grokspice:
Fixes #12642

## The failure this fixes

`E2E Test CI` failed **9 times across 7 distinct branches** between 2026-08-04 and
2026-08-06, on both `merge_group` (dequeues) and `push` to `trunk`. Every one of
them was a `model (darwin-aarch64)` job on the shared self-hosted macOS pool, and
every one of them reported the wrong thing.

- https://github.com/spiceai/spiceai/actions/runs/31075749056 (`push` on `trunk`)
- https://github.com/spiceai/spiceai/actions/runs/31088421790
- https://github.com/spiceai/spiceai/actions/runs/31081140526
- https://github.com/spiceai/spiceai/actions/runs/31066186618

In the first of those, two runner instances on the **same physical machine**
(`...-mbp-runner-01-01` and `...-mbp-runner-01-02`) ran overlapping jobs. The openai
job's own `spice.log` — printed by `stop-spice` at the end of the job — says:

```
ERROR spiced: Unable to start Spice Runtime servers: Unable to start Flight server: Address 127.0.0.1:50051 is already in use by another process. ...
```

`Wait for Spice runtime is ready` passed anyway, because the sibling instance's
runtime answered `localhost:8090/v1/ready`. Four minutes later the job went red on
`Test chat` with `Model did not return expected response` — about a model
belonging to a different job's runtime.

## Root cause

`wait_for_ready.sh` already has a guard for exactly this: it treats a bind failure
in **our own** log as authoritative over whatever answered the socket. It matched
the literal `Address already in use`, which is only one of the runtime's two
wordings:

| listener | message | matched before? |
|---|---|---|
| HTTP | `Unable to bind to address: Address already in use (os error 48)` — the OS error passed through | yes |
| Flight / cluster | `Address 127.0.0.1:50051 is already in use by another process.` — `Error::AddressAlreadyInUse`, `crates/runtime/src/flight/mod.rs` | **no** |

Flight is the one that fires, because :50051 is the port a sibling runtime holds.
So the guard never ran. The existing unit tests only ever fed it the HTTP wording,
which is why the gap survived.

## The change

Match both wordings, as two exact alternatives:

```
'Address already in use|Address [^[:space:]]+ is already in use'
```

Two alternatives rather than `Address .*already in use` on purpose: tracing renders
a line's fields and its message together, so a wildcard spanning the halves could
pair an `address` field with an unrelated "already in use" message — and the
runtime has several of those (a Postgres replication slot, a view name, a schedule
name). Both false-positive shapes are now pinned by tests.

I checked the other listeners for wordings still missed: `Error::AddressAlreadyInUse`
is the shared variant for Flight (plain and TLS) and the cluster server, HTTP passes
the OS error through, and the cluster *executor* path deliberately falls back to a
dynamic port with `Unable to bind executor flight server to ..., using dynamic port ...`
— non-fatal, and correctly not matched.

## What this does not change

- The pass condition is otherwise untouched: a genuine `ready` with no bind failure
  in the log still passes, immediately.
- No timeout is lengthened, no assertion loosened, no retry added. This makes a
  failure **earlier and correctly attributed**; it does not make anything greener.
  A port collision now fails in the step that owns it, named as a collision.
- Making the collision impossible (per-job ports, or serialising the runtime-owning
  jobs across instances on one machine) is out of scope and noted in #12642.

## Test plan

- `scripts/test_e2e_spice_lifecycle.sh` grows 5 cases and now passes **64**:
  - the Flight wording must refuse a foreign `ready` (2 assertions);
  - the Flight wording must be named as a contended port in the timeout diagnosis;
  - an unrelated `... is already in use` (replication slot) must still pass;
  - a line carrying *both* an address and an unrelated `already in use` must still
    pass — the case that justifies the two alternatives over a wildcard.
- **The 4 load-bearing assertions fail against the pre-fix script** — verified by
  restoring `origin/trunk`'s `wait_for_ready.sh` under the new tests:
  `expected '1', got '0'` on the refusal, and the diagnosis note absent.
- `shellcheck .github/actions/wait-for-spice-ready/wait_for_ready.sh` — clean.
- `python .github/scripts/check_workflow_yaml.py` — OK, 108 definitions well-formed.
- `actionlint` — no new findings versus the unmodified tree.
- The diff touches no Rust and no workflow triggers, so the Rust gate has nothing to
  build; `E2E Lifecycle Scripts Test` is the check that exercises it, and its
  `paths:` filter already covers both changed files.

## Notes for reviewers

- Verified against the source, not just one log sample: `git grep 'already in use'`
  over `crates/` accounts for every message the pattern could see, and each is
  either the target or provably unmatched.
- The refactor into `own_runtime_failed_to_bind()` is behaviour-preserving — same
  `[ -f "$LOG_FILE" ] && grep ...` body, so a missing, empty, unreadable, or
  directory `$LOG_FILE` still returns false exactly as before.
- **No overlap with the other two PRs from this sweep.** #12643's fix touches only
  `e2e_test_ci.yml`'s `e2e-gate` job and #12644's only `test_openai_model`'s
  `timeout-minutes`; this touches neither file.

### DIFF
--- a/.github/actions/wait-for-spice-ready/wait_for_ready.sh
+++ b/.github/actions/wait-for-spice-ready/wait_for_ready.sh
@@ -36,6 +36,33 @@ port_from_url() {
   esac
 }
 
+# The runtime reports a port collision in two different wordings, and a guard
+# that knows only one of them does not fire on the other (#12642):
+#
+#   - the HTTP server surfaces the OS error as-is —
+#     "Unable to bind to address: Address already in use (os error 48)"
+#   - the Flight server formats its own, with the address in the middle —
+#     "Address 127.0.0.1:50051 is already in use by another process."
+#     (crates/runtime/src/flight/mod.rs)
+#
+# Flight is the one that actually shows up on the self-hosted macOS pool, since
+# it binds :50051 while the collision is usually with a sibling runner's runtime
+# on the same machine. The same formatted variant covers the cluster listener.
+#
+# Two exact alternatives rather than `Address .*already in use`: tracing puts a
+# line's fields and its message together, so a wildcard between the two halves
+# could pair an `address` field with an unrelated "already in use" message — and
+# the runtime has several of those (a Postgres replication slot, a view name, a
+# schedule name). Neither alternative below matches any of them.
+BIND_FAILURE_PATTERN='Address already in use|Address [^[:space:]]+ is already in use'
+
+# True when our own runtime recorded that it could not bind. Authoritative over
+# whatever answers the readiness socket: on a shared machine that answer can come
+# from another job's runtime.
+own_runtime_failed_to_bind() {
+  [ -f "$LOG_FILE" ] && grep -Eq "$BIND_FAILURE_PATTERN" "$LOG_FILE" 2>/dev/null
+}
+
 # Runs only on the failure path, and is deliberately not itself time-bounded:
 # `lsof` can block on a wedged mount, and the job-level `timeout-minutes` is the
 # backstop for that. A step-level timeout is not an option — the inner steps of
@@ -82,7 +109,7 @@ diagnose() {
     cat "$LOG_FILE"
     # The single highest-value line: a port already held means a previous job on
     # this runner leaked its runtime, not that this job is broken.
-    if grep -q "Address already in use" "$LOG_FILE" 2>/dev/null; then
+    if grep -Eq "$BIND_FAILURE_PATTERN" "$LOG_FILE" 2>/dev/null; then
       echo
       echo "NOTE: the runtime could not bind its port because something else already held it."
       echo "      Two causes, and the process list above tells them apart:"
@@ -121,7 +148,7 @@ while :; do
     # runtime — and every test step after this would then run against it.
     # Failing to bind always leaves this line in our own log, so treat it as
     # authoritative over whatever answered the socket.
-    if [ -f "$LOG_FILE" ] && grep -q "Address already in use" "$LOG_FILE" 2>/dev/null; then
+    if own_runtime_failed_to_bind; then
       echo "Something on ${URL} reports ready, but it is not this job's runtime:"
       echo "${LOG_FILE} shows our own spiced failed to bind its port."
       diagnose "$last_body"
--- a/scripts/test_e2e_spice_lifecycle.sh
+++ b/scripts/test_e2e_spice_lifecycle.sh
@@ -393,6 +393,55 @@ assert_contains "says whose runtime it is not" "$out" "not this job's runtime"
 assert_not_contains "does not claim our runtime came up" "$out" "Runtime ready after"
 rm -f "$work_dir/spice.log"
 
+# Same case, in the wording the runtime actually uses on the pool (#12642). The
+# Flight server formats its own message with the address in the middle instead of
+# passing the OS error through, so a guard written against the HTTP wording alone
+# never fires on it — and :50051 is the port that collides, because it is Flight
+# that a sibling runner's runtime is already holding. Left unmatched, the wait
+# passes and the job instead goes red much later on a model that "did not return
+# expected response" — a model belonging to a different job's runtime.
+reset_state "201 spiced $THEIRS"
+printf 'ERROR spiced: Unable to start Spice Runtime servers: Unable to start Flight server: Address 127.0.0.1:50051 is already in use by another process. Either stop the existing process or change the address: https://spiceai.org/docs/cli/reference/run\n' \
+  > "$work_dir/spice.log"
+result="$(run_subject "$wait_script" STUB_READY_BODY=ready SPICE_READY_TIMEOUT=5 SPICE_READY_INTERVAL=1)"
+out="${result#*|}"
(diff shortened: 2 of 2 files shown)


grokspice:
**`E2E Lifecycle Scripts Test` is red here because of a GitHub Actions incident, not
because of this diff.** Recording the evidence so the next reader does not re-triage
it — and so nobody reads the red as a verdict on the change.

Four attempts, none of which executed a single step:

| attempt | jobs | steps run | runner assigned | outcome |
|---|---|---|---|---|
| 1 | both | 0 | — | `Set up job`: `Failed to resolve action download info. Error: Service Unavailable` |
| 2–4 | both | 0 | **none** (`runner_name` empty) | cancelled after ~15–17 min waiting for a runner |

Both jobs — `ubuntu-24.04` and `macos-latest` — were cancelled within a second of
each other on each attempt with no compute ever allocated. During the same window
the repo's queue held 18–42 runs, and `Workflow YAML Check` on this same commit
went green on its third attempt, so the tree itself validates.

**Please re-run this check before reading anything into it.** Nothing in the diff can
affect runner allocation: it changes one `grep` pattern in
`.github/actions/wait-for-spice-ready/wait_for_ready.sh` and adds cases to
`scripts/test_e2e_spice_lifecycle.sh`.

What that suite reports locally, for reference until CI can confirm it:

```
all 64 tests passed
```

and, with `origin/trunk`'s `wait_for_ready.sh` restored underneath the new tests:

```
  FAIL: refuses a ready when Flight lost the bind — expected '1', got '0'
  FAIL: recognises the Flight bind wording
  FAIL: does not pass a Flight bind failure off as ready
  FAIL: explains the contended port for Flight too
4 of 63 tests failed
```

`Attestation` is green (no Rust-affecting paths), as are `E2E Test CI`,
`ADBC integration tests`, `Workflow YAML Check` and `Enforce Pulls with Spice`.

lukekim (approved):

Issue #12656, opened 2026-08-06
SFTP handshake and password auth have no deadline, so a peer that accepts TCP and stalls holds a blocking-pool thread


grokspice:
## Describe the bug

`SFTPClientConfig::connect` (`crates/runtime-object-store/src/store/sftp.rs`) bounds only
the TCP connect, and only when `client_timeout` is set. The two stages a stalled peer
actually holds open are unbounded in every configuration:

```rust
let mut session = Session::new().map_err(handle_error)?;
session.set_tcp_stream(stream);
session.handshake().map_err(handle_error)?;          // <- no deadline
session
    .userauth_password(&self.user, &self.password)   // <- no deadline
    .map_err(handle_error)?;
```

`Session::set_timeout` is never called, so the ssh2 session inherits no read or write
timeout. A peer that completes the TCP handshake and then never sends its SSH version
string holds the call indefinitely.

`connect()` runs inside `tokio::task::spawn_blocking` (from `list_all_files` and
`list_directory_shallow`), so a stalled attempt occupies a blocking-pool thread for as
long as the peer keeps the socket open, and the blocking pool is shared process-wide.

## Expected behavior

One connection attempt completes or fails within a bounded wall-clock time, whatever the
peer does.

## Scope

This is the same bug-class as #12647 (FTP) and #12606 (MSSQL): a connection attempt whose
authentication stage has no deadline. It is filed separately because the mechanism differs
— ssh2's `Session` is blocking, so the bound is `Session::set_timeout` (which applies to
the session's blocking reads and writes) rather than a `tokio::time::timeout`, and it must
be set after `set_tcp_stream` but before `handshake`.

Unlike FTP there is no pool here, so the failure mode is a held blocking-pool thread
rather than a pinned pool slot.

## Possible approach

- Call `Session::set_timeout` with the configured `client_timeout`, or a default in the
  same range #12647 chose (20s), before `handshake()`.
- Note that ssh2's timeout is in milliseconds and `0` means "no timeout", so a
  sub-millisecond or zero duration must not be passed through as "unbounded".
- Cover it with a listener that accepts TCP and sends no SSH banner, asserting the connect
  fails within the bound.

## Additional context

Noticed while fixing #12647 — the FTP fix removed the shared `SocketAddr` parse defect in
both stores, but only FTP's authentication stage got a deadline.

grokspice:
This is now fixed in PR #12655 rather than separately.

[...]

PR #12660, opened 2026-08-06
fix(object-store): bound the FTP and SFTP session setup, and stop rejecting hostnames when client_timeout is set (fixes #12647)

### LINKED ISSUE #12647: FTP connection attempts have no deadline on the login, so a peer that accepts TCP and stalls pins a pool slot
## Describe the bug

The FTP object store's pooled connections have no deadline on the part of the
attempt that a stalled peer actually holds open. `FTPConnectionManager::connect`
(`crates/runtime-object-store/src/store/ftp.rs`) bounds only the TCP connect, and
only when the connector's `timeout` parameter is set:

```rust
let mut client = match timeout {
    Some(timeout) => AsyncFtpStream::connect_timeout(addr, timeout).await,
    None => AsyncFtpStream::connect(format!("{host}:{port}")).await,
}?;

client.login(&user, &password).await?;   // <- no deadline, in either branch
```

Nothing bounds `login`, and nothing sets a read or write timeout on the stream —
`timeout` is only ever passed to `connect_timeout`, never to the session.

This is the same shape as #12606 in the MSSQL connector: `bb8` consults its own
`connection_timeout` only *after* `ManageConnection::connect` returns, so an
attempt that never returns is never abandoned. The pending pool slot stays taken,
later `Pool::get` calls are suppressed even once the server recovers, and timing
out `Pool::get` does not cancel the replenishment task that is stuck.

`get_fresh_client` (the non-pooled path used for operations that mutate session
state) has the identical gap.

## Reproduction shape

A peer that completes the TCP handshake and then never answers the FTP greeting
or the `USER`/`PASS` exchange. Nothing has to be malformed, and the OS-level
connect timeout never applies because the connection was accepted. With
`timeout` unset, the TCP connect itself is unbounded too.

## Expected behavior

One `ManageConnection::connect` completes or fails within a bounded wall-clock
time, whatever the peer does, and its pool slot is released when it does.

## Notes on the fix

Unlike #12606 this is not purely additive, which is why it is filed separately
rather than folded into that PR:

- `timeout` is an existing, documented connector parameter whose current meaning
  is "TCP connect timeout". Widening it to cover the login changes what an
  operator who already set it gets.
- A default is needed for the `None` case, and it should sit inside whatever
  window the pool is willing to wait, the way #12606 chose 20s against `bb8`'s
  30s default.

Suggested direction: wrap the whole sequence — connect plus login — in a single
`tokio::time::timeout`, so the two stages cannot each spend the budget, and apply
the same bound on both the pooled and `get_fresh_client` paths.

## Additional context

Found while fixing #12606; that fix covers the MSSQL connector only.

### LINKED ISSUE #12654: Setting client_timeout on the FTP or SFTP connector rejects every hostname, because the timed path parses host:port as a SocketAddr
## Describe the bug

Setting `client_timeout` on the `ftp` or `sftp` connector makes the connector
unusable for any host given as a name rather than a literal IP address. Both
connectors route the timed branch through an API that takes an
already-resolved `SocketAddr`, and reach it by `parse()`-ing `"{host}:{port}"`:

`crates/runtime-object-store/src/store/ftp.rs`:

```rust
let mut client = match timeout {
    Some(timeout) => {
        AsyncFtpStream::connect_timeout(
            format!("{host}:{port}").parse().map_err(
                |e: std::net::AddrParseError| ...
            )?,
            timeout,
        ).await
    }
    None => AsyncFtpStream::connect(format!("{host}:{port}")).await,
};
```

`crates/runtime-object-store/src/store/sftp.rs` has the same shape around
`TcpStream::connect_timeout`.

`SocketAddr`'s `FromStr` accepts only numeric addresses, so `ftp.example.com:21`
never parses. The untimed branch is fine — `AsyncFtpStream::connect` takes
`A: ToSocketAddrs` and resolves — which is why this is invisible until an
operator sets the parameter.

## Reproduction shape

A pod with a hostname and the documented parameter set:

```yaml
- from: ftp://ftp.example.com/data/
  name: remote
  params:
    ftp_user: ${secrets:ftp_user}
    ftp_pass: ${secrets:ftp_pass}
    client_timeout: 30s
```

Every operation fails immediately, before a packet is sent, with:

```
Generic FTP error: invalid socket address syntax
```

Observed directly against the private helper (host `localhost`,
`client_timeout` 200ms):

```
PROBE hostname+timeout => Some("Generic FTP error: invalid socket address syntax")
```

Removing `client_timeout` makes the same pod work, which inverts the intent of
the parameter: the only configuration that bounds the connect is the one that
cannot connect.

## Expected behavior

`client_timeout` bounds how long the connector waits. It does not restrict which
hosts can be addressed — a hostname works with or without it.

## Notes on the fix

- FTP: drop `connect_timeout` and wrap `AsyncFtpStream::connect(host:port)` in a
  `tokio::time::timeout`, so resolution stays in the library and the bound
  covers the whole attempt.
- SFTP: resolve with `ToSocketAddrs` first, then `TcpStream::connect_timeout`
  per resolved candidate, sharing one deadline across candidates.

## Additional context

Found while fixing #12647 (no deadline on the FTP login); the two defects sit on
the same lines, so they are fixed together. `client_timeout` is documented for
`ftp` and `sftp` since v0.17.1-beta, described as "stop waiting for a response
from the data source after the specified duration", with a documented default of
30 seconds — the code applies no default at all when the parameter is unset,
which is the subject of #12647.

claudespice:
## Summary

The FTP and SFTP object stores bounded only the TCP connect, and only when
`client_timeout` was set. Nothing bounded the part a stalled peer actually holds
open — the FTP `USER`/`PASS` exchange, and the SFTP banner and password
exchange — so a peer that completed the TCP handshake and then went quiet parked
the attempt indefinitely:

- **FTP**: `bb8` consults its own `connection_timeout` only *after*
  `ManageConnection::connect` returns, and its replenishment task calls
  `connect()` with no deadline at all. An attempt that never returns is never
  abandoned, its pool slot is never released, and later `Pool::get` calls stay
  suppressed even once the server recovers. `get_fresh_client`, the non-pooled
  path, had the identical gap.
- **SFTP**: `handshake()` and `userauth_password()` are blocking `libssh2`
  calls, so a quiet peer consumed a blocking-pool thread that could not be
  reclaimed.

The timed branch also could not address a host by *name*. It reached
`connect_timeout` — which takes an already-resolved `SocketAddr` — by
`parse()`-ing `"{host}:{port}"`, and `SocketAddr`'s `FromStr` accepts only
numeric addresses. So setting `client_timeout` failed every hostname with
`invalid socket address syntax` before a packet was sent, which is why the
missing deadline went unnoticed: the only bounded path was unusable. Observed
directly against the private helper before the fix:

```
PROBE hostname+timeout  => Some("Generic FTP error: invalid socket address syntax")
PROBE stalled peer (client_timeout=200ms) => returned=false elapsed=3.0s  (never returns)
```

`client_timeout` is documented for `ftp` and `sftp` as "stop waiting for a
response from the data source after the specified duration", with a default of
30 seconds. Neither connector applied a default at all, and neither applied the
configured value beyond the TCP connect.

Fixes #12647
Fixes #12654

## Changes

- One deadline covers connect **and** login together, so neither stage can spend
  the whole budget, defaulting to the documented 30s when `client_timeout` is
  unset rather than to no bound.
- FTP connects by name and lets the library resolve. SFTP resolves first, then
  bounds each candidate against one shared deadline.
- The pooled and non-pooled FTP paths now share a single `connect_and_login`, so
  a call site cannot reach an unbounded variant. `FTPConnectionManager` holds the
  config instead of duplicating its five fields and its redacting `Debug` impl.
- An SFTP session timeout bounds every blocking `libssh2` wait, including later
  reads on the returned session. `libssh2` applies it per wait for socket
  readiness, so a transfer that keeps making progress is not cut off.
- A liveness check (`is_valid`) against a server that went quiet now invalidates
  the connection instead of waiting, so the pool discards it rather than handing
  the same dead connection back on every checkout — a pool that survives an
  outage instead of staying unusable after it.
- The SFTP connect error names how many resolved addresses were tried, and a
  `tracing` error sink makes an invalidated FTP connection visible (bb8's
  default sink drops it silently).

Not changed: the NFS store, which already documents that the `libnfs` bindings
expose no timeout configuration, and the SMB store, which already bounds its
negotiate/session-setup reads via `read_timeout` with a 30s default — that
sibling is the in-repo precedent for the shape this PR gives FTP and SFTP.

## Behavior change

A server slower than the deadline now fails where it previously blocked until it
eventually answered. Two cases:

- `client_timeout` unset: previously unbounded, now the documented 30s default.
- `client_timeout` set: previously covered only the TCP connect, now covers the
  whole session setup, and for SFTP also later blocking reads.

Both directions make the parameter mean what it is documented to mean. The error
message names the deadline and says to raise `client_timeout` if the server is
simply slow.

## Test plan

- [x] `make lint` — clean
- [x] `make test` — green
- [x] `cargo fmt --all`
- [x] 14 new tests in `crates/runtime-object-store/src/store/{ftp,sftp}.rs`
      (72 in the crate's lib suite, all passing), against loopback peers:
  - a peer that accepts TCP and never answers: the FTP non-pooled path, the FTP
    pooled path, and the SFTP handshake each give up at the deadline
  - a hostname (`localhost`) with `client_timeout` set no longer fails with
    `invalid socket address syntax` — the regression test for #12654
  - happy path: a peer that answers the greeting and the `USER`/`PASS`
    exchange still yields a logged-in session, so the deadline does not cost
    the working case
  - a peer that logs in and then goes quiet: the liveness check invalidates the
    connection at the deadline
  - controls: a refused port fails immediately rather than being masked by the
    deadline; an unresolvable host is reported as a resolution failure
  - an unset `client_timeout` still has a deadline; an exhausted deadline stops
    trying candidates; the pooled manager's `Debug` output still redacts the
    password
- [x] Neuter matrix — each piece of the fix reverted in turn, confirming a named
      test catches it (results in the PR discussion)

## Checklist

- [x] Tests added, including a regression test that fails on the old code
- [x] No `.unwrap()` / `.expect()` in production code
- [x] Error messages name the resource and an actionable fix, with a docs link
- [x] No blocking calls added to an async context; the SFTP path stays inside
      `spawn_blocking`
- [x] No new dependencies

### DIFF
--- a/crates/runtime-object-store/src/store/ftp.rs
+++ b/crates/runtime-object-store/src/store/ftp.rs
@@ -57,6 +57,23 @@ struct FTPConnectionManager {
     config: Arc<FTPClientConfig>,
 }
 
+/// Surfaces the pool's own connection failures, which it reports to a sink
+/// rather than to a caller. `bb8`'s default sink discards them, so a connection
+/// discarded in the background — including one abandoned at its deadline — would
+/// otherwise leave no trace of why.
+#[derive(Debug, Clone, Copy)]
+struct LogErrorSink;
+
+impl bb8::ErrorSink<object_store::Error> for LogErrorSink {
+    fn sink(&self, error: object_store::Error) {
+        tracing::warn!("FTP connection pool discarded a connection: {error}");
+    }
+
+    fn boxed_clone(&self) -> Box<dyn bb8::ErrorSink<object_store::Error>> {
+        Box::new(*self)
+    }
+}
+
 impl bb8::ManageConnection for FTPConnectionManager {
     type Connection = AsyncFtpStream;
     type Error = object_store::Error;
@@ -70,7 +87,8 @@ impl bb8::ManageConnection for FTPConnectionManager {
         &self,
         conn: &mut Self::Connection,
     ) -> impl Future<Output = Result<(), Self::Error>> + Send {
-        let deadline = self.config.deadline();
+        let config = Arc::clone(&self.config);
+        let deadline = config.deadline();
         let noop_future = conn.noop();
         Box::pin(async move {
             match tokio::time::timeout(deadline, noop_future).await {
@@ -82,7 +100,8 @@ impl bb8::ManageConnection for FTPConnectionManager {
                 Err(_elapsed) => Err(object_store::Error::Generic {
                     store: STORE_NAME,
                     source: format!(
-                        "The FTP server did not answer a liveness check within {deadline:?}; discarding the connection."
+                        "FTP server ftp://{}:{} did not answer a liveness check within {deadline:?}; discarding the connection.",
+                        config.host, config.port
                     )
                     .into(),
                 }),
@@ -209,6 +228,7 @@ impl FTPInner {
                 };
                 Pool::builder()
                     .max_size(DEFAULT_POOL_SIZE)
+                    .error_sink(Box::new(LogErrorSink))
                     .build(manager)
                     .await
                     .map_err(|e| generic_error(STORE_NAME, e))
@@ -585,6 +605,7 @@ impl ObjectStore for FTPObjectStore {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use bb8::ManageConnection;
     use chrono::Utc;
     use std::time::Instant;
 
@@ -614,6 +635,13 @@ mod tests {
         assert_eq!(cloned.port, "21");
     }
 
+    /// Bind a listener on an ephemeral loopback port.
+    fn loopback_listener() -> (std::net::TcpListener, u16) {
+        let listener = std::net::TcpListener::bind("127.0.0.1:0").expect("bind loopback");
+        let port = listener.local_addr().expect("local addr").port();
+        (listener, port)
+    }
+
     /// A peer that completes the TCP handshake and then never answers, so the
     /// FTP greeting and the `USER`/`PASS` exchange stall. Returns the port.
     ///
@@ -621,8 +649,7 @@ mod tests {
     /// `spawn_blocking`, because a blocking task that never returns also blocks
     /// the test runtime's shutdown.
     fn stalled_ftp_peer() -> u16 {
-        let listener = std::net::TcpListener::bind("127.0.0.1:0").expect("bind loopback");
-        let port = listener.local_addr().expect("local addr").port();
+        let (listener, port) = loopback_listener();
         std::thread::spawn(move || {
             // Hold every accepted connection open without writing a byte.
             let mut accepted = Vec::new();
@@ -638,8 +665,7 @@ mod tests {
     fn logged_in_then_quiet_peer() -> u16 {
         use std::io::{BufRead, BufReader, Write};
 
-        let listener = std::net::TcpListener::bind("127.0.0.1:0").expect("bind loopback");
-        let port = listener.local_addr().expect("local addr").port();
+        let (listener, port) = loopback_listener();
         std::thread::spawn(move || {
             let mut held = Vec::new();
             while let Ok((stream, _)) = listener.accept() {
@@ -715,8 +741,6 @@ mod tests {
     /// stalled attempt cannot occupy a pool slot indefinitely.
     #[tokio::test]
     async fn pool_manager_gives_up_on_a_peer_that_never_answers() {
-        use bb8::ManageConnection;
-
         let port = stalled_ftp_peer();
         let manager = FTPConnectionManager {
             config: Arc::new(config_for(
@@ -779,8 +803,7 @@ mod tests {
     #[tokio::test]
     async fn a_refused_port_fails_without_waiting_for_the_deadline() {
         // Bind and drop, so the port is almost certainly unused.
-        let listener = std::net::TcpListener::bind("127.0.0.1:0").expect("bind loopback");
-        let port = listener.local_addr().expect("local addr").port();
+        let (listener, port) = loopback_listener();
         drop(listener);
 
         let config = config_for("127.0.0.1", port.to_string(), Some(Duration::from_secs(30)));
@@ -822,8 +845,6 @@ mod tests {
     /// even after the server recovered.
     #[tokio::test]
     async fn a_liveness_check_against_a_quiet_server_invalidates_the_connection() {
-        use bb8::ManageConnection;
-
         let port = logged_in_then_quiet_peer();
         let manager = FTPConnectionManager {
             config: Arc::new(config_for(
--- a/crates/runtime-object-store/src/store/ftp.rs
+++ b/crates/runtime-object-store/src/store/ftp.rs
@@ -45,27 +45,16 @@ const STORE_NAME: &str = "FTP";
 const MAX_CONCURRENT_LISTINGS: usize = 4;
 /// Default connection pool size.
 const DEFAULT_POOL_SIZE: u32 = 4;
+/// Deadline applied to establishing a session when `client_timeout` is unset.
+/// Matches the documented default for the parameter, and sits inside `bb8`'s
+/// own 30s `connection_timeout` so a stuck attempt cannot outlive the pool's
+/// willingness to wait for it.
+const DEFAULT_CLIENT_TIMEOUT: Duration = Duration::from_secs(30);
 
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

claudespice:
### Verification

**Neuter matrix** — each piece of the fix reverted in turn, confirming a named test catches it. Positive control on the clean tree first (48 tests under the `store::` filter, `test result: ok`), and a run that hangs counts as caught, because an indefinite wait *is* the pre-fix behaviour:

| neuter | verdict |
| --- | --- |
| FTP: no deadline over connect+login | CAUGHT (hang) — `fresh_client_gives_up_on_a_peer_that_never_answers` |
| FTP: pooled path with its own unbounded copy | CAUGHT (hang) — `pool_manager_gives_up_on_a_peer_that_never_answers` |
| FTP: back to `connect_timeout` + `parse()` | CAUGHT — `a_hostname_is_resolved_when_client_timeout_is_set` |
| FTP: no default when `client_timeout` unset | CAUGHT — `an_unset_client_timeout_still_has_a_deadline` |
| FTP: unbounded liveness check | CAUGHT (hang) — `a_liveness_check_against_a_quiet_server_invalidates_the_connection` |
| SFTP: no session timeout | CAUGHT (hang) — `connect_gives_up_on_a_peer_that_never_sends_a_banner` |
| SFTP: back to `connect_timeout` + `parse()` | CAUGHT — `a_hostname_is_resolved_when_client_timeout_is_set` |
| SFTP: no guard on an exhausted budget | CAUGHT — `an_exhausted_deadline_stops_trying_candidates` |
| SFTP: no default when `client_timeout` unset | CAUGHT — `an_unset_client_timeout_still_has_a_deadline` |

9/9. The SFTP budget guard was initially **not** caught: a `checked_sub` ahead of it already returned `None` for every reachable zero case, so the guard could not fire and the test could not distinguish it. Restructured to one reachable `saturating_sub` check, after which the neuter is caught.

**Adversarial review**: the cross-model reviewer is unavailable (`Your workspace is out of credits`), so this was a written pass. Three findings, all applied: the SFTP error now reports how many resolved addresses were tried, since one shared deadline can stop before a later address is reached; bb8's default error sink discards background connection failures, so an abandoned attempt left no trace and now logs; and the boundary that name resolution is bounded by the system resolver rather than by this deadline is documented where it applies.

**Security review** (by hand): no new dependencies, no `unsafe`, no new deserialization surface, no new I/O beyond the connect that already existed. Credentials are unchanged and still redacted — a test asserts the pooled manager's `Debug` output keeps the password redacted after it moved to holding the config. No new error or log line interpolates a credential: they carry `host:port`, a resolved `SocketAddr`, an `io::Error`, and the deadline. Hostname resolution is not a new trust boundary; the host comes from the operator's own spicepod, and the untimed path already resolved it. The change reduces exposure to a slow-loris peer rather than adding to it.

claudespice:
Sign-off dispatched remotely (run [31123566225](https://github.com/spiceai/spiceai/actions/runs/31123566225)); local sign-off is not currently possible on this machine because `/tmp/cayenne_test*` is owned by another uid and unwritable, so the cayenne tests cannot run. The self-hosted pool had 5 sign-off runs ahead of this one at dispatch time, so `Attestation` will stay red until it is picked up — it needs the run to finish, not a re-dispatch.

claudespice:
Closing as a duplicate of #12655, which was opened 27 minutes earlier by @grokspice against the same two files and covers a strict superset of the scope: #12647 (FTP login deadline), #12654 (the `SocketAddr` parse that made `client_timeout` reject every hostname), and #12656 (the SSH handshake/auth deadline, which this PR did not address).

I re-checked the one item this branch had that #12655 does not — bounding the FTP pool manager's `is_valid` (`conn.noop()`). It is not a defect: bb8 0.9.1 calls `is_valid` inline inside the `timeout(connection_timeout, ..)` wrapper in `get()` (`bb8-0.9.1/src/inner.rs:110`, timeout at :122), so a stalled `noop()` is bounded and the future is dropped on expiry. Unlike `connect()`, there is no pending slot to strand. So there is nothing owed to #12655 from this branch.

No review action needed here — #12655 is the one to take forward.

PR #12557, opened 2026-08-05
fix(ci): call an unreachable compiler cache infrastructure, not a failing branch (fixes #12556)

grokspice:
Fixes #12556.

## The failure this removes

The pool compiles through `sccache` (`RUSTC_WRAPPER`, set by `.github/actions/setup-sccache`)
against an S3-compatible endpoint on the runner host. When that endpoint stops answering,
sccache's server cannot start and **every** `rustc` invocation from that moment on dies
before compiling anything:

```
sccache: error: Server startup failed: cache storage failed to read: Unexpected (temporary) at read => send http request
  error sending request for url (http://127.0.0.1:8333/sccache/sccache/.sccache_check): tcp connect error: Connection refused (os error 61)
error: process didn't exit successfully: `sccache …/rustc -vV` (exit status: 2)
debug: posting signoff=failure on spiceai/spiceai@b2cae1ec…: Sign-off checks failed after 5859s (triggered by grokspice)
```

The branch is never evaluated — nothing is compiled — and the run still says
**`Sign-off checks failed after 5859s`**, which is a statement about the diff.

That is worse than an ordinary red, because a failed sign-off **disqualifies the commit**:
`Attestation` rejects that HEAD before anything else, no earlier sign-off can be inherited
past it, and the revert / no-Rust-changes fast-tracks cannot route around it. An
infrastructure hiccup on a shared runner therefore turns a green, approved PR red, and the
only way out is another full multi-hour sign-off — while the run gives no hint that a
re-dispatch is the remedy.

**Blast radius** — window 2026-08-04T13:16Z → 2026-08-05T11:05Z, 200 most recent failed runs:

- `pr` → `Attestation` → *Reject a failed sign-off on the head commit* — **13 failed runs
  across 11 distinct branches**, each blocked from the merge queue.
- `signoff.yml` across the full 7-day window (queried per workflow file, so uncapped): **109 failures of 383 concluded runs — 28%**.

Confirmed instance: https://github.com/spiceai/spiceai/actions/runs/30978407363 (PR #12540 —
5859s, then `sccache: error: Server startup failed`), and the red `Attestation` it produced:
https://github.com/spiceai/spiceai/actions/runs/30978391211

That 13-run cluster has more than one cause — a genuinely failing cayenne test (#12558) also
lands there. This PR fixes the purely infrastructural one.

## What changed

`scripts/signoff` already knows how to say "the runner broke, not your branch" — the
[...]

PR #12662, opened 2026-08-06
fix(ci): take the trusted sign-off helpers from the workflow's own commit (fixes #12657)

### LINKED ISSUE #12657: bug(ci): a cancelled sign-off cannot withdraw its own failure status on a branch older than #12432
### What happens

When a sign-off run is cancelled, the step that is supposed to withdraw the `signoff` status it
posted crashes, so the cancellation leaves a `signoff=failure` on the commit. On an approved PR that
turns `Attestation` red for a reason that has nothing to do with the code — the code was never
judged.

From run [31115145848](https://github.com/spiceai/spiceai/actions/runs/31115145848)
(PR #12598, `958d0bb1a`):

```
##[error]The operation was canceled.
##[group]Run .trusted-ci/signoff correct-cancelled "" "spiceai/spiceai"
✗ unknown command: correct-cancelled (see: scripts/signoff --help)
##[error]Process completed with exit code 1.
```

The job then concludes `failure` rather than `cancelled`, and the head keeps the
`signoff=failure` the cancelled run had already posted.

### Why

`scripts/signoff correct-cancelled` was added by #12432 (`45589075b`, 2026-08-05). The workflow calls
it through the pinned trusted copy:

```yaml
- name: Correct the sign-off status when this run is cancelled
  if: cancelled()
  run: .trusted-ci/signoff correct-cancelled "" "${{ github.repository }}"
```

but `.trusted-ci/signoff` is created by copying the **checked-out working tree's** script:

```yaml
- name: Save trusted CI helpers
  run: |
    mkdir -p .trusted-ci
    cp scripts/signoff .trusted-ci/signoff
```

and for a same-repo PR the preceding checkout uses `ref: needs.resolve.outputs.checkout_ref` — the
PR's head commit. So the *workflow definition* comes from the dispatched ref and knows about
`correct-cancelled`, while the *script* comes from the branch under test and may not. `fix/12588-cli-response-read-data-loss`
branched before `45589075b` and contains zero occurrences of `correct-cancelled`.

The population that cannot repair its own cancelled status is therefore exactly the population that
still needs the repair: branches older than the fix. Fork PRs are unaffected — for those the
checkout is the dispatched ref and the copy happens before the fork overlay, so the copy is current.

### Impact

Cancellations are not rare right now (several runs are being cancelled at the `Resolve sign-off target`
stage while `spiceai-dev-runners` carries a large queue backlog), and each one can leave a spurious
`signoff=failure`. #12598 has sat `APPROVED` + `BLOCKED` since 2026-08-06T00:17Z with `signoff` as its
only red check.

### Suggested direction

Take the two trusted helpers from the workflow's own commit rather than from the checked-out tree, so
their version always matches the call sites in the same workflow file — e.g. fetch `github.sha` and
materialise into the untracked `.trusted-ci/` with `git show` / `git archive` rather than checking it
out over tracked paths.

The security property must not regress: the point of `.trusted-ci/` is that a fork PR cannot supply
the code that decides its own sign-off, so the copy has to keep coming from this repo and be taken
before any PR-authored code is on disk.

### Not the fix

Making the step tolerant (`|| true`) would hide the skew and silently leave the bad status in place —
the status is the thing that needs correcting.

claudespice:
## Summary

The sign-off job runs `.trusted-ci/signoff` and `.trusted-ci/actions/**` rather than the
working tree's copies, so that the code under test cannot supply the script that decides
its own sign-off. Those copies were made with `cp` from the checked-out tree — but for a
same-repo PR the tree is the **branch under test**, while the call sites live in the
workflow file, which always comes from `github.sha` (the dispatch ref; `make signoff-remote`
does not pass `--ref`, so that is the default branch).

So a branch older than a helper change presented an older helper to newer call sites. A
branch predating `scripts/signoff correct-cancelled` (#12432) supplied a script without it,
so the `if: cancelled()` step died on `unknown command: correct-cancelled` and the cancelled
run kept the `signoff=failure` it had already posted. That is a red `Attestation` on code
nothing ever judged, and the population that cannot repair its own cancelled status is
exactly the population that needs the repair. Confirmed on the branch from the run in the
issue: trunk's `scripts/signoff` has 2 occurrences of `correct-cancelled`,
`fix/12588-cli-response-read-data-loss` has 0.

It also closes the same-repo half of the trust boundary the fork case already had: a branch
under test no longer supplies the script that judges it.

Fixes #12657

## Changes

- **`.github/workflows/signoff.yml`** — `Save trusted CI helpers` reads both helpers out of
  the object database at `${{ github.sha }}` (`git archive` / `git show`) instead of copying
  the working tree, fetching that commit first if the checkout did not bring it.
  `actions/checkout` fetches only the ref it checks out even at `fetch-depth: 0`, so the
  commit is usually absent in the same-repo case and already present in the fork case (where
  it *is* the ref checked out). The SHA is passed through `env:` rather than interpolated
  into the shell body. The step still runs before the fork overlay, and reading the ODB means
  the overlay cannot reach these files even in principle.
- **`scripts/test_signoff_trusted_helpers.sh`** — new. Extracts the real `run:` block from
  the workflow and executes it against local git fixtures (no network, no credentials).
- **`.github/workflows/signoff_trusted_helpers_test.yml`** — runs it on GitHub-hosted
  `ubuntu-24.04`, paths-filtered, `contents: read` only.
- **`docs/dev/ci_signoff.md`** — documents where the trusted helpers come from and the
  consequence below.

The step stays inline in the workflow rather than moving to `scripts/`: a saver script in the
repo would itself be read from the branch under test, which is the bug. That is why the test
extracts the step instead of keeping a copy — it binds to the shipping definition.

## Behaviour change worth knowing

A branch that changes `scripts/signoff` or `.github/actions/**` is now signed off with the
**dispatch ref's** version of those, not its own. That is the point, but it means such a
branch can no longer self-serve a fix to a broken helper. The escape hatch is one flag —
dispatch on the branch, which puts the workflow and the helpers back on one commit:

```bash
gh workflow run signoff.yml --ref <branch> -f branch=<branch>
```

This is documented in the step comment and in `docs/dev/ci_signoff.md`. Three open PRs touch
`scripts/signoff` (#12557, #12392, #12383); none is blocked by this, but they are the
population the note is for.

An unresolvable trusted commit **fails the step** rather than falling back to the tree. A
silent fallback would restore the exact skew this fixes, and would do it invisibly — the run
would go on to sign the branch off with the branch's own script. A transient fetch failure
here would have broken the `Fetch trunk for targeted pre-lint` step immediately after anyway.

## Test plan

This branch touches no Rust — `.github/**`, `scripts/test_*`, and `docs/**` are outside the
Rust gate's path filter, so `Attestation` auto-passes and no `make lint` / `make nextest` run
is applicable to it. What was run:

- `bash scripts/test_signoff_trusted_helpers.sh` — **16/16 assertions pass**. Covers: the
  trusted script and composite actions come from the workflow's commit and not from the
  checked-out branch; the branch's own copies do not appear; the `correct-cancelled`
  subcommand the workflow calls is present in the trusted copy (the #12657 regression); the
  executable bit survives; the working tree is left at the branch under test; the
  already-present (fork) path needs no fetch; an unresolvable commit fails rather than falls
  back; a stale `.trusted-ci/` from a previous run on a reused workspace is replaced, not
  merged into.
- **Neuter matrix, 5/5 caught**, each by the test named for it: (1) `cp` the script from the
  tree — the original bug; (2) `cp -R` the actions from the tree; (3) drop the `chmod +x`;
  (4) drop the `rm -rf` of a stale `.trusted-ci/`; (5) tolerate an unresolvable commit and
  fall back to the tree.
- `python3 .github/scripts/check_workflow_yaml.py` — `OK: 109 GitHub Actions definitions are
  well-formed`.
- `tar --strip-components` is exercised on both implementations: the sign-off job runs on
  `spiceai-macos` (bsdtar), which the local run of the test above used, and the new test
  workflow runs on `ubuntu-24.04` (GNU tar).

## Checklist

- [x] Root cause identified and confirmed against the run in the issue
- [x] Regression test that fails on the old step and passes on the new one
- [x] Neuter matrix run; every test proven to catch its own defect
- [x] No credentials, secrets, or internal details added to logs, errors, or this body
- [x] Trust boundary reviewed: narrowed, not widened
- [x] Docs updated

### DIFF
--- a/.github/workflows/signoff.yml
+++ b/.github/workflows/signoff.yml
@@ -41,9 +41,10 @@
 # PR/run lookups) at the fork instead of this repo.
 #
 # scripts/signoff and the setup-* composite actions below are also saved to
-# .trusted-ci/ from this repo *before* any fork commit is checked out, and
-# every later step references those copies (literal `.trusted-ci/...` paths)
-# rather than the working tree. This runs unconditionally, not just for fork
+# .trusted-ci/, read out of the object database at this workflow file's own
+# commit (github.sha) *before* any fork commit is checked out, and every later
+# step references those copies (literal `.trusted-ci/...` paths) rather than
+# the working tree. This runs unconditionally, not just for fork
 # PRs, so every step downstream has one fixed path to depend on instead of
 # branching on is_fork — `uses:` can't read a step output (the `steps`
 # context isn't available there), so a per-case path chosen at runtime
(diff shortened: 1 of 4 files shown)


claudespice:
@copilot review

claudespice:
CI note, so the reds are not misread: the `pull_request` runs on this branch were all cancelled at once (`pr`, `E2E Test CI`, `Workflow YAML Check`, `Sign-off Trusted Helpers Tests`, both `check_changes`) by the **GitHub Actions incident** in progress — githubstatus reports Actions in `major_outage` with an open critical incident. The head commit is unchanged (`3d149ae`, nobody pushed) and the PR was not edited, so none of those cancellations is a verdict on the diff.

What did complete before the outage:

- `Sign-off Resolve Target Tests`: **SUCCESS** — the sibling extraction test still passes against the edited `signoff.yml`.
- `License compliance check`: SUCCESS.
- Copilot review: 4/4 files, zero comments.

`enforce-pull-with-spice` shows a `CANCELLED` from the usual re-run churn with a newer run behind it.

These need a re-run once Actions recovers; re-running during the outage would just burn the queue. No `make signoff` is owed — the branch touches no Rust, so `Attestation` auto-passes once `pr` can run.

lukekim (approved):

PR #12682, opened 2026-08-07
fix(ci): refuse to attest a run that is neither a pull request nor a queue entry (fixes #12679)

### LINKED ISSUE #12679: pr.yml: a workflow_dispatch takes the merge-queue passthrough, so Attestation goes green without inspecting any sign-off

claudespice:
Fixes #12679

## Summary

The `attestation` job in `.github/workflows/pr.yml` gates every verifying step on
`github.event_name == 'pull_request'`, and its only other path is a passthrough gated on
`github.event_name != 'pull_request'` that succeeds unconditionally. That reasoning is right for
`merge_group` — entry into the queue is itself gated on a green `Attestation`, so the sign-off was
already validated for that tree — but `workflow_dispatch` takes the same path. `Attestation` is a
required check and the only quality gate on a PR, so dispatching `pr.yml` against a PR branch posts
the one green check that PR needs onto a head commit nobody signed off.

The condition `!= 'pull_request'` conflates "not a pull request" with "already verified".

### This is not theoretical

Dispatching `pr.yml` is what people reach for when a push did not start CI, which is exactly the
state a number of PRs were left in by the 2026-08-06 Actions incident. Confirmed on real runs:

- A `workflow_dispatch` run of `pr.yml` shows all eleven verifying steps `skipped` and only
  `Merge queue passthrough` `success`, so the job reports `success`
  ([run 30952600523](https://github.com/spiceai/spiceai/actions/runs/30952600523)).
- Three such dispatches landed during the incident recovery and each minted a green `Attestation`
  check-run on a PR head. Two of those PR heads carry no `signoff` status at all, so their required
  gate is currently green having inspected nothing.

## Changes

- **Name `merge_group` explicitly** in the passthrough instead of negating `pull_request`, so the
  queue-entry premise cannot be inherited by a trigger that does not have it.
- **Fail closed for every other trigger.** A new step refuses to attest a run that is neither a
  pull request nor a merge queue entry. A trigger added to `on:` later lands here too and has to
  decide its own semantics before it can be attested.
- **Regression test** in `.github/scripts/test_pr_attestation_signoff.mjs`, which already lifts this
  job's steps out of the workflow. It reads the `on:` block, models which steps each trigger can
  reach, and asserts that no trigger outside `pull_request`/`merge_group` reaches a step that
  reports success.

### Behaviour change worth knowing

A manual dispatch now reports a red `Attestation` rather than a green one. The heavy jobs in that
run still report their own results, and the error says how to get a real verdict. If a PR already
had a legitimate green `Attestation`, a later dispatch replaces it with a red one — recoverable by
re-running `Attestation` on the `pull_request` run. Failing closed is the deliberate trade: a check
that says "verified" without verifying is worse than one that says "cannot verify here".

Making a dispatch skip the job instead was considered and rejected — a skipped job counts as
success for a required check, which is the same defect.

### Scope note

The other `!= 'pull_request'` conditions in this file (`check_changes`, `Rust Lint`,
`Build and Test`, `Build (release profile)`) were swept and left alone: they gate whether heavy jobs
*run*, and running them on a manual dispatch is the point of the dispatch. Only the attestation
passthrough turned "not a pull request" into a verdict.

## Test plan

- `node .github/scripts/test_pr_attestation_signoff.mjs` — **16/16 pass** (12 pre-existing, 4 new).
  This is the suite CI runs for this file, via `.github/workflows/attestation_script_test.yml`.
- `python .github/scripts/check_workflow_yaml.py` — OK, 108 definitions well-formed.
- `python .github/scripts/test_check_workflow_yaml.py` — 42 tests, OK.
- **Neuter matrix, 7/7 caught.** Each direction of wrongness was injected and the suite re-run:
  reverting to the pre-fix state; negating `pull_request` again in the passthrough; deleting the
  refusal step; making the refusal report success; letting the refusal swallow `merge_group`;
  gating the sign-off inspection away from `pull_request`; and adding a `||` the condition model
  cannot represent. Every one is caught by a named assertion, and the matrix was re-run after the
  test was strengthened.
- `make lint` / `make nextest` are **not** applicable to this branch and were not run: it changes
  only `.github/**`, which matches no path in `RUST_AFFECTING_PATH_PATTERN` (`scripts/signoff`), so
  the Rust gate is a no-op and the branch takes the no-Rust fast-track.

## Checklist

- [x] Workflow YAML validates against the repo's own linter
- [x] No Rust-affecting paths touched
- [x] `Attestation` green on this PR (no-Rust fast-track; the run was created and the check passed)

### DIFF
--- a/.github/scripts/test_pr_attestation_signoff.mjs
+++ b/.github/scripts/test_pr_attestation_signoff.mjs
@@ -385,6 +385,29 @@ const admittedSteps = (eventName) =>
 /** Whether a step fails the job rather than reporting success. */
 const stepFails = (stepName) => /(^|\n)\s*exit 1\s*(\n|$)/.test(stepLines(job, stepName).join('\n'));
 
+/**
+ * Whether a step admitted for `eventName` is certain to run, rather than merely
+ * able to.
+ *
+ * `admitsEvent` treats a `steps.*.outputs.*` term as possibly true, so an
+ * admitted step is only *maybe* reached. That over-approximation is the wrong
+ * direction for asserting that a trigger hits a step which fails the job: a
+ * refusal gated on some earlier step's output would satisfy the assertion while
+ * a real run skipped it and reported success. Requiring the condition to rest on
+ * `github.event_name` alone closes that gap.
+ */
+function certainlyRuns(stepName, eventName) {
+  const condition = stepCondition(job, stepName);
+  if (!admitsEvent(condition, eventName)) {
+    return false;
+  }
+  if (condition === null) {
+    return true;
+  }
+  const residue = condition.replace(/github\.event_name\s*(?:==|!=)\s*'[^']*'/g, '');
+  return !/steps\.|needs\.|inputs\.|github\./.test(residue);
+}
+
 test('the merge queue passthrough names merge_group instead of negating pull_request', () => {
   const condition = stepCondition(job, PASSTHROUGH_STEP);
   assert.match(
@@ -436,9 +459,9 @@ test('no other trigger can reach a green Attestation', () => {
       `a ${event} run reaches no step at all, so Attestation reports success for free`
     );
     assert.ok(
-      admitted.some((step) => stepFails(step)),
-      `a ${event} run reaches only [${admitted.join(', ')}], none of which fails the job, so ` +
-        `it posts a green required Attestation without inspecting any sign-off`
+      admitted.some((step) => stepFails(step) && certainlyRuns(step, event)),
+      `a ${event} run reaches only [${admitted.join(', ')}], none of which is certain to fail ` +
+        `the job, so it posts a green required Attestation without inspecting any sign-off`
     );
   }
 });
--- a/.github/scripts/test_pr_attestation_signoff.mjs
+++ b/.github/scripts/test_pr_attestation_signoff.mjs
@@ -300,6 +300,149 @@ test('an unsigned non-merge commit cannot inherit', async () => {
   assert.match(result.failed[0], /not a two-parent merge/);
 });
 
+// --- Which triggers can reach a green Attestation (#12679) ------------------
+//
+// Attestation is the only required quality gate on a PR, so a trigger that
+// reaches the end of this job without inspecting a sign-off mints that gate for
+// free. The steps above are gated on `pull_request`; the checks below assert
+// that every *other* trigger declared in `on:` is either the merge queue — whose
+// entry was itself gated on a green Attestation — or refused outright.
+
+const PASSTHROUGH_STEP = 'Merge queue passthrough';
+
+/** The event names declared in the workflow's `on:` block. */
+function triggers(workflow) {
+  const lines = workflow.split('\n');
+  const start = lines.findIndex((line) => /^on:\s*$/.test(line));
+  assert.notEqual(start, -1, 'pr.yml no longer declares an `on:` block');
+
+  const names = [];
+  for (const line of lines.slice(start + 1)) {
+    // A key at column 0 ends the block.
+    if (/^[A-Za-z_]/.test(line)) {
+      break;
+    }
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

claudespice:
Duplicate of #12681, which was opened a minute earlier for the same issue — we self-assigned #12679 37 seconds apart and neither run saw the other. Closing this one as the later of the two.

Both PRs land the same two-part fix (name `merge_group` in the passthrough; fail closed otherwise). The one substantive difference — making the refusal a catch-all on `!= 'pull_request' && != 'merge_group'`, so an unforeseen trigger added to `on:` fails closed at runtime rather than relying on a test assertion to catch it — is carried over as a review comment on #12681, along with the run-level evidence from this investigation.

No code from here is lost that is not either already in #12681 or written up there.

PR #12691, opened 2026-08-07
feat(cloud-connect): push metrics from adopted standalone instances

peasee:
## 📝 Summary

An adopted standalone `spiced` now pushes its own metrics to Spice Cloud, so the portal's app dashboards work for instances the platform does not host.

The runtime collects its metrics on a timer and sends them to the gateway over the existing Cloud Connect stream as a serialized OTLP request. Totals are cumulative, so a dropped or skipped push loses a data point but never data — the next one carries the running total, and nothing needs buffering or retrying.

The dashboards filter on `scp_app_id`, and a runtime has no way to know which app it belongs to. The control plane does, so it sends the app id with the deploy (`ApplySpicepod`) and the runtime stamps it on what it exports. It is stored alongside the identity, so a restart keeps exporting rather than going quiet until the next deploy.

An instance that has never been told its app exports nothing at all. That is deliberate: a series with no `scp_app_id` is ingested but matches no dashboard, so sending it spends quota on something nothing can read — and does it silently.

The gateway remains the authority on tenancy. It stamps `scp_org_id` from the verified mTLS identity and overwrites whatever the client claimed, which is what keeps one org's metrics out of another's. `scp_app_id` is attribution within an org rather than a tenant boundary, so it stays the workload's to set.

Metric names match the `/metrics` scrape path, so one dashboard query serves both.

## 🔗 Related

* See source PR: https://github.com/spiceai/spiceai/pull/12318

### DIFF
--- a/bin/spiced/src/cloud_connect.rs
+++ b/bin/spiced/src/cloud_connect.rs
@@ -534,7 +534,7 @@ impl SpicedRuntimeHandle {
             Err(join) => format!("identity persistence task panicked: {join}"),
         };
         tracing::warn!(
-            "Cloud Connect: could not persist the app id to {}: {error}; metrics stay attributed until this process exits, then are withheld until the next deploy",
+            "Spice Cloud Connect could not save the cloud app ID to {}. Metrics for this instance will not appear in Spice Cloud if the process restarts. Does the runtime have permission to write to that path? {error}",
             self.identity_path.display()
         );
     }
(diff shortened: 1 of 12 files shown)


bin/spiced/src/cloud_connect.rs
```diff
@@ -474,9 +510,35 @@ impl SpicedRuntimeHandle {
...
+        let app_id = app_id.to_string();
+        let result =
+            tokio::task::spawn_blocking(move || IdentityStore::store_app_id(&path, &app_id)).await;
+        let error = match result {
+            Ok(Ok(())) => return,
+            Ok(Err(err)) => err.to_string(),
+            Err(join) => format!("identity persistence task panicked: {join}"),
+        };
+        tracing::warn!(
+            "Cloud Connect: could not persist the app id to {}: {error}; metrics stay attributed until this process exits, then are withheld until the next deploy",
```
peasee: what does this mean to a user? prefer a more user-friendly language, like `Cloud Connect could not persist the cloud app ID to {}. Metrics for this instance will not appear in cloud if the process restarts. Does Runtime have permissions to write to the destination? {error}`

bin/spiced/src/cloud_connect.rs
```diff
@@ -615,8 +677,48 @@ impl RuntimeHandle for SpicedRuntimeHandle {
...
+        //
+        // Empty means the control plane named none. Leave any id already recorded
+        // in place rather than clearing it: the instance's app has not changed,
+        // and clearing would silence metrics that are correctly attributed.
+        if app_id.is_empty() {
+            let held = self.app_id.read().clone();
+            match held {
+                Some(held) => tracing::warn!(
+                    app_id = %held,
+                    "Cloud Connect: ApplySpicepod carried no app id; keeping the one already recorded, so metrics stay attributed"
```
peasee: again, what does this mean to a user. If one is already recorded, but nothing arrived in the message, this is more like a `tracing::debug` as well.

bin/spiced/src/cloud_connect.rs
```diff
@@ -615,8 +677,48 @@ impl RuntimeHandle for SpicedRuntimeHandle {
...
+                ),
+                None => tracing::warn!(
+                    "Cloud Connect: ApplySpicepod carried no app id and none was recorded before, so metrics stay withheld. Either this instance is attached to no app, or the api that dispatched the deploy does not forward the app id"
```
peasee: prefer something like `Cloud Connect provided no app ID on Spicepod deployment. Metrics will not appear in Cloud. Is this instance attached to an app?`

crates/runtime-cloud-connect/src/client.rs
```diff
@@ -647,6 +650,76 @@ impl ClientDriver {
...
+        // Metrics ride the same stream but not the same delivery guarantee. The
+        // payload carries cumulative totals, so a dropped export costs a data
+        // point and no data — whereas a late heartbeat is read as an instance
+        // going away. This task therefore offers its message and moves on
+        // instead of waiting for room behind one.
+        let met_tx = tx.clone();
+        let met_runtime = Arc::clone(&runtime);
+        let met_identifier = Arc::clone(&identifier);
+        let met_interval = self.config.metrics_interval;
+        let met_handle = tokio::spawn(async move {
```
peasee, line 667: is this handle missing a cancellationtoken to ensure we can properly end it on runtime shutdown?

crates/runtime-cloud-connect/src/client.rs
```diff
@@ -647,6 +650,76 @@ impl ClientDriver {
...
+                            identifier = %id,
+                            bytes,
+                            "Cloud Connect: metrics export queued to the gateway"
+                        );
+                    }
+                    Err(mpsc::error::TrySendError::Full(_)) => {
+                        tracing::warn!(
+                            identifier = %id,
+                            bytes,
+                            "Cloud Connect: metrics export dropped — the outbound queue is full. The next export carries the same cumulative totals, so no data is lost"
```
peasee: Prefer messaging like `Failed to export metrics to Cloud Connect: the outbound queue is full. Runtime will retry metrics export on the next interval. Metrics may be delayed to appear in Cloud.`

crates/runtime-cloud-connect/src/client.rs
```diff
@@ -924,6 +998,24 @@ impl ClientDriver {
...
         session_key: Option<&cloud_connect_crypto::EncryptionKeypair>,
     ) {
+        // Logged on arrival, before anything can fail: the app id rides this
+        // command and nothing else carries it, so whether one arrived is the
+        // first thing to know when metrics are being withheld.
+        if cmd.app_id.is_empty() {
+            tracing::warn!(
+                command_id,
+                yaml_bytes = cmd.spicepod_yaml.len(),
+                "Cloud Connect: ApplySpicepod received with no app id; metrics stay withheld because nothing can attribute them. The control plane sends one only if the instance is attached to an app"
```
peasee: this is similar to the previous message about receiving an ApplySpicepod with no app ID. We should also make sure we're not double-logging this?

crates/runtime-cloud-connect/src/client.rs
```diff
@@ -944,16 +1036,19 @@ impl ClientDriver {
...
                 spicepod_yaml: &cmd.spicepod_yaml,
                 delivered_secrets: delivered,
+                app_id: &cmd.app_id,
             })
             .await;
 
         let outcome = match outcome {
             Ok(outcome) => outcome,
             Err(err) => {
+                tracing::warn!(command_id, "Cloud Connect: ApplySpicepod failed: {err}");
```
peasee: this is a wildly un-useful error to a user.

crates/runtime-cloud-connect/src/handlers.rs
```diff
@@ -205,6 +205,13 @@ pub struct SpicepodDeployment<'a> {
     /// `None` means the deployment carried none, which is distinct from an
     /// empty map — an app whose secrets were all removed.
     pub delivered_secrets: Option<DeliveredSecrets>,
+    /// The app this instance's telemetry is attributed to. It rides the
+    /// deployment because the instance cannot derive it and the control plane
+    /// already knows it; a handle that exports metrics records it and stamps it
+    /// as `scp_app_id`.
+    ///
+    /// Empty means the control plane named no app.
+    pub app_id: &'a str,
```
peasee: Shouldn't this be `Option<&'a str>`?

crates/runtime/src/metrics_reader.rs
```diff
@@ -26,13 +26,26 @@ use std::sync::{Arc, Weak};
...
     InstrumentKind, ManualReader, Pipeline, Temporality, data::ResourceMetrics,
     reader::MetricReader,
 };
 use prost::Message;
+use snafu::prelude::*;
+
+/// Why an on-demand collection produced no payload.
+#[derive(Debug, Snafu)]
+pub enum Error {
+    #[snafu(display("Failed to collect the runtime's own metrics: {source}"))]
```
peasee: prefer `Failed to collect Runtime metrics: {source}`

phillipleblanc (approved):

Issue #12698, opened 2026-08-07
s3_vectors latency histograms have no bucket below 100ms, so sub-100ms quantiles report the percentile itself


grokspice:
## Summary

The `s3_vectors_*_latency` histograms in `crates/runtime/src/embeddings/index/s3/metrics.rs` build
their boundaries as:

```rust
(0..10).map(|i| 100.0 * f64::from(i))    // 0, 100, 200, ... 900
```

so the first boundary above zero is 100ms and there is no resolution below it — the same shape as
#12693, on a different histogram.

## Impact

Any S3 Vectors operation faster than 100ms lands in a single bucket, and a quantile interpolated
inside that bucket is a function of the requested percentile alone: p50 reports 50ms, p90 reports
90ms, regardless of the real latency. A small-index query answered in 10-40ms is squarely in that
band.

Milder than #12693 because these are network round trips rather than in-process point lookups, so a
larger share of the distribution does clear 100ms — which is also why it is filed separately rather
than folded into that fix.

## What a fix has to decide

Unlike #12693 there is no obviously right floor to inherit: the answer depends on what S3 Vectors
latency actually looks like in practice, which should be measured rather than assumed. Whoever
picks the boundaries should say what they are sized for.

## Notes

Found by an adversarial review of the #12693 fix while sweeping for other duration histograms with
the same shape. The other `with_boundaries` call sites in the tree are count- or byte-scaled and
are not affected.

Issue #12697, opened 2026-08-07
Snapshot write duration and bytes are recorded twice under one metric name, so their counts are doubled


grokspice:
## Summary

`record_write_metrics` in `crates/runtime-acceleration/src/snapshot/metrics.rs` records the same
observation to `dataset_acceleration_snapshot_write_duration_ms` twice:

- once through `telemetry::record_snapshot_write_metrics(...)`, which registers that instrument in
  `crates/telemetry/src/lib.rs`, and
- once through the module's own `SNAPSHOT_WRITE_DURATION_MS`, registered under the identical name.

`dataset_acceleration_snapshot_write_bytes` is recorded twice the same way.

The exporter is built `.without_scope_info()` (`bin/spiced/src/lib.rs`), so the two meter scopes
collapse into one Prometheus family rather than staying distinct. Every write therefore contributes
two observations to one histogram.

## Impact

`_count` and every bucket are doubled, so a rate over this metric reads as twice the real write
rate. `_sum` is doubled too, so `_sum/_count` — the mean — stays correct, which is what has kept
this hidden. The Grafana and Datadog panels on this metric read the inflated counts.

## Notes

Found while fixing #12693, which had to make the two registrations agree on their bucket boundaries
(they had drifted into two different `le` sets, which would have made the summed series
non-monotonic). That fix makes the boundaries consistent; it does not address the double record,
which is a separate question about which of the two registrations should own the instrument.

grokspice:
Withdrawing this — I filed it, and the root cause I gave is wrong. There is no double count on `/metrics`.

## What the report got wrong

The claim rests on one sentence:

> The exporter is built `.without_scope_info()` (`bin/spiced/src/lib.rs`), so the two meter scopes collapse into one Prometheus family rather than staying distinct.

`.without_scope_info()` does collapse scopes — but only among instruments that reach *that* exporter. The two registrations here never reach the same exporter, because they bind to two different `MeterProvider`s.

**The module's own instruments** (`crates/runtime-acceleration/src/snapshot/metrics.rs`) go through `global::meter(...)`, so they bind to the OpenTelemetry **global** provider, which `init_metrics` installs at `bin/spiced/src/lib.rs:1340` with the operator's readers — Prometheus scrape, `spice_metrics`, cluster OTLP, OTEL push.

**The `telemetry::record_snapshot_*` calls** guard on `meter::METER`:

```rust
let Some(m) = meter::METER.get() else { return };
```

[...]

PR #12699, opened 2026-08-07
fix(telemetry): resolve duration histograms below a millisecond (fixes #12693)

### LINKED ISSUE #12693: Query latency quantiles report percentile x 100: duration histogram has no bucket below 100ms
`DURATION_MS_HISTOGRAM_BUCKETS` (`crates/telemetry/src/lib.rs:41`) has no bucket between 0 and 100ms:

```rust
[0.0, 100.0, 250.0, 500.0, 750.0, 1000.0, 2500.0, ...]
```

Point-lookup queries complete in well under 1ms, so every observation lands in the first bucket and `calculate_percentile` (`crates/runtime/src/metrics_server/mod.rs:584`) interpolates within it, returning `percentile x 100` regardless of actual latency.

Measured on one series with 11,925 observations:

| | value |
| --- | --- |
| buckets | `le=0 -> 0`, `le=100 -> 11925`, all higher `le -> 11925` |
| true mean (`_sum/_count`) | **0.2755 ms** |
| reported p50 / p90 / p95 / p99 | **50 / 90 / 95 / 99** |

Each reported quantile equals `q x 100` exactly. The values are a function of the requested percentile alone and carry no latency information — a 300x regression would leave them unchanged until queries cross 100ms. Downstream consumers inherit this: reconstructing a distribution from these buckets yields ~50ms (the midpoint of `(0,100]`).

Affects every histogram on the shared constant — `query_duration_ms`, `flight_request_duration_ms`, `http_requests_duration_ms` (`telemetry/src/lib.rs:192, 366, 758, 842, 859, 874`).

### Fix

`CONTENTION_MS_HISTOGRAM_BUCKETS` (`telemetry/src/lib.rs:75`) already solves this and is already used by the Cayenne contention metrics, whose comment names this exact gap:

> The shared `DURATION_MS_HISTOGRAM_BUCKETS` jumps straight from 0 to 100ms, which is too coarse for lock/checkpoint latencies that live in the 0.1-50ms band

Same runtime, both histograms, showing the difference:

```
cayenne_metastore_writer_wait_ms   le=0.1 -> 394   le=0.25 -> 401   le=1 -> 407   le=5 -> 408
query_duration_ms                  le=0   -> 0     le=100  -> 11925  (single populated bucket)
```

Point the duration histograms at that bucket set, or a shared successor with sub-millisecond resolution.

`calculate_percentile` itself is correct; only the boundaries are wrong.

Version: 2.1.4; refs against trunk 1cebec0252.

grokspice:
## Summary

`DURATION_MS_HISTOGRAM_BUCKETS` had no boundary between 0 and 100ms. Point-lookup queries finish in
well under a millisecond, so every observation landed in that single bucket and `calculate_percentile`
interpolated inside it — returning the requested percentile times 100 and no latency information at
all. On the reported series (11,925 observations, true mean 0.2755ms) the reported p50/p90/p95/p99
were exactly 50/90/95/99.

Fixes #12693

## Changes

- **`DURATION_MS_HISTOGRAM_BUCKETS`** gains nine sub-100ms boundaries (0.1 … 50). Every boundary the
  set already had is kept, so adding one only subdivides an existing bucket and an `le` series keeps
  its meaning. The head now matches `CONTENTION_MS_HISTOGRAM_BUCKETS`, whose comment named this exact
  gap; that comment is updated, since the remaining distinction is tail reach (30s vs 500s).
- **`crates/runtime-acceleration/src/snapshot/metrics.rs`** held its own copy of the constant and
  registers `dataset_acceleration_snapshot_write_duration_ms` — the same instrument name `telemetry`
  registers. The two bind to different providers: the instruments in this module take the
  OpenTelemetry global provider `init_metrics` installs (the operator's `/metrics`, OTLP push and
  `spice_metrics` readers), while the `telemetry::` calls beside them take
  `telemetry::meter::METER`, the anonymous-telemetry provider `anonymous::start` builds and never
  publishes globally. They are still one measurement reported to two pipelines, and a quantile is
  comparable between them only if both name the same boundaries, so the copy now uses the shared
  constant.
- **`flight_request_duration_ms`** named no boundaries at all and took the OpenTelemetry defaults,
  whose floor is 5ms. The issue lists it as affected, so it is pointed at the same set.

## What operators should expect

- The derived `*_summary.quantile` values **step by orders of magnitude** for the histograms this
  resolves — a point-lookup p50 goes from 50 to ~0.375. That is the point of the change, but any
  alert or SLO threshold set against the old numbers needs revisiting.
- Bucket count goes 15 → 24, roughly +60% bucket series per attribute set on ~25 instruments. The
  largest term is `http_requests_duration_ms`, keyed by method × path.
- This moves the floor rather than removing it. A workload entirely below 0.1ms still shares one
  bucket, and its quantiles are still interpolation artifacts — now `percentile / 1000`. Bucketed
  histograms cannot do better than their finest boundary.
- Several instruments on this constant are never sub-millisecond (refresh, compaction, snapshot
  write), so the new head buckets are dead weight for them. A separate request-path-only bucket set
  would avoid that; it is the road not taken here, in favour of one constant.

## Follow-ups filed

- ~~#12697 — snapshot write duration and bytes are recorded **twice** under one metric name, so
  their counts are doubled.~~ **Withdrawn — not a defect.** I filed it from this PR on the premise
  that the exporter's `.without_scope_info()` collapses the two registrations into one Prometheus
  family. It does not: `.without_scope_info()` only ever applies to instruments in the global
  provider, and the `telemetry::` registration is not one. See the analysis on that issue.
- #12698 — the `s3_vectors_*_latency` histograms have the same shape (no bucket below 100ms). Filed
  separately because the right floor there depends on measured S3 latency.

## Test plan

- `make lint`
- `make test`

New tests:

- `sub_millisecond_latencies_are_resolved_by_the_duration_buckets`
  (`crates/runtime/src/metrics_server/mod.rs`) — replays the reported series through
  `calculate_percentile` and asserts each quantile lands in the bucket the observations are in
  (`0.25 < v <= 0.5`). On the old boundaries these return 50/90/95/99 and every assertion fails.
- `the_duration_buckets_resolve_below_a_millisecond` (`crates/telemetry/src/lib.rs`) — asserts the
  first boundary above zero is at or below 1ms. Returns 100.0 on the old set.
- `histogram_boundaries_are_strictly_increasing` — guards both millisecond bucket sets.

## Checklist

- [x] Regression tests that fail on the old boundaries
- [x] Swept for the same shape elsewhere; the one copied bucket set is fixed here, the rest filed
- [x] Backwards compatibility of the `le` series verified against in-repo dashboard consumers
- [ ] `make signoff` green


grokspice:
@copilot review

crates/runtime-acceleration/src/snapshot/metrics.rs
```diff
@@ -18,10 +18,17 @@ use opentelemetry::{
     metrics::{Counter, Gauge, Histogram, Meter},
 };
 
-pub const DURATION_MS_HISTOGRAM_BUCKETS: [f64; 15] = [
-    0.0, 100.0, 250.0, 500.0, 750.0, 1000.0, 2500.0, 5000.0, 7500.0, 10000.0, 25000.0, 50000.0,
-    100_000.0, 250_000.0, 500_000.0,
-];
+// `dataset_acceleration_snapshot_write_duration_ms` is recorded to two separate pipelines: the
```
sgrebnov: make comments more simple/concise

crates/telemetry/src/lib.rs
```diff
@@ -37,10 +37,21 @@ pub const ROWS_RETURNED_HISTOGRAM_BUCKETS: [f64; 18] = [
...
 ];
 
-// Extended default buckets for duration histogram: 25000.0, 50000.0, 100000.0, 250000.0, 500000.0
-pub const DURATION_MS_HISTOGRAM_BUCKETS: [f64; 15] = [
-    0.0, 100.0, 250.0, 500.0, 750.0, 1000.0, 2500.0, 5000.0, 7500.0, 10000.0, 25000.0, 50000.0,
-    100_000.0, 250_000.0, 500_000.0,
+// Boundaries for every millisecond-scale duration histogram, spanning a sub-millisecond point
+// lookup through a multi-hundred-second refresh.
+//
+// The sub-100ms boundaries resolve the band most requests finish in. Without them a point lookup
```
sgrebnov: make comments more simple/concise

sgrebnov (dismissed):

crates/runtime-acceleration/src/snapshot/metrics.rs
grokspice: Tightened in `be4d719cd` — down to four lines. It keeps the two facts the comment is here for: the metric reaches two providers that share no exporter, so it is one observation each rather than a double count; and the constant is shared rather than copied because a quantile is only comparable across them while both name the same boundaries.

crates/telemetry/src/lib.rs
grokspice: Tightened in `be4d719cd` — twelve lines down to seven. Unchanged: why the sub-100ms head exists (without it a 0.1ms lookup and a 99ms one share a bucket), why the head matches `CONTENTION_MS_HISTOGRAM_BUCKETS` while the 500s tail keeps the two sets separate, and that new boundaries subdivide rather than invalidate recorded `le` series.

crates/runtime-acceleration/src/snapshot/metrics.rs
grokspice: Cut again in fe85bab. This one is down to three lines: why the constant is shared rather than copied. The "one observation each, not a double count" explanation went with it — it was arguing against a misreading rather than telling a reader anything they need at that line.

crates/telemetry/src/lib.rs
grokspice: Cut again in fe85bab. Both bucket comments now state one fact each — why the sub-100ms head exists, and why the contention set stops at 30s. The "head matches CONTENTION" and "new boundaries subdivide existing buckets" paragraphs are gone; the first was duplicated on the other constant and the second is a property of histograms generally, not of this set. Trimmed the two regression-test doc comments and the flight one the same way.

grokspice:
Sign-off status note, so this is not re-litigated on the next pass.

Run [31151014286](https://github.com/spiceai/spiceai/actions/runs/31151014286) posted `signoff=failure`, but the failure is not this branch:

```
Summary [550.837s] 9888 tests run: 9887 passed (5 slow, 1 flaky), 1 failed, 27 skipped
  FLAKY 2/6 cayenne::cold_tier_test test_cold_tier_concurrent_scan_during_promotion_impl_turso
  SIGABRT   cayenne::mutation_property_test prop_sequential_memory_impl_turso
```

One failure in 9888, and it is the Cayenne `*_turso` SIGABRT of #12436 — a stack-size problem in the backend-parameterized tests that has nothing to do with a metrics histogram change and cannot be reached from this diff. The cold-tier test above it is the #12555 race and passed on retry.

#12561 is the fix for the SIGABRT and is queued. Re-dispatching sign-off before it lands would just re-roll the same dice at the cost of another pool run, so this is waiting on #12561 rather than being retried.

grokspice:
The red `signoff` on this commit is not this branch.

The run failed on `cayenne::mutation_property_test prop_sequential_memory_impl_turso`, which aborted with SIGABRT after overflowing its stack — the known trunk flake tracked as #12436. Everything else passed (9889 of 9890). The diff here does not touch cayenne, turso, or anything on that test's path.

#12561 is the fix for #12436 and is currently in the merge queue. Re-dispatching sign-off before it lands would spend another ~3.5 hours and hit the same abort, so I am holding off until it merges and then re-signing off.

grokspice:
The red `signoff` on this branch is **not** this diff. Run [31181538277](https://github.com/spiceai/spiceai/actions/runs/31181538277) reports:

```
Summary [1261.441s] 9888 tests run: 9887 passed (17 slow, 1 leaky), 1 failed, 27 skipped
   SIGABRT [0.299s] (9884/9888) cayenne::mutation_property_test prop_sequential_memory_impl_turso
   (test aborted with signal 6: SIGABRT)
```

One failure, and it is a `*_turso` case aborting on signal 6 — the tokio-worker stack overflow tracked in #12436, in a crate this PR does not touch (the diff is the telemetry duration-histogram buckets). #12436's exclusion list does not cover this test, so it reaches the gate.

**Do not re-dispatch sign-off for this** — it will reproduce, at a cost of ~3 hours of a contended runner. The fix is #12561, currently in the merge queue. Once it lands, a sign-off on this branch should go green.

Elapsed was 10696s, well inside the 353-minute budget, so this is a genuine check result rather than the no-verdict timeout in #12741.

lukekim (approved):

PR #12704, opened 2026-08-07
fix(ci): retry the sign-off status post, and say so when it never lands (fixes #12701)

### LINKED ISSUE #12701: Sign-off strands a commit in `pending` when the status POST fails (Bad credentials on a long run)
## What happened

The remote sign-off for #12652 ([run 31133614812](https://github.com/spiceai/spiceai/actions/runs/31133614812)) ran the full gate for 10,089s, reached a real verdict — `nextest` failed on the `prop_sequential_memory_impl_turso` stack overflow (#12436) — and then failed to record it:

```
debug: posting signoff=failure on spiceai/spiceai@a80e0ac3: Sign-off checks failed after 10089s
gh: Bad credentials (HTTP 401)
warning: could not post the failed sign-off status — the 'Attestation' check still reflects the previous sign-off
```

The `Resolve an incomplete sign-off` step is the designed fallback for exactly this, but it posts through the same helper and hit the same 401:

```
debug: posting signoff=failure ...: Remote sign-off ended without posting a verdict — sign off again
gh: Bad credentials (HTTP 401)
✗ failed to post the sign-off status (do you have write access to spiceai/spiceai?)
```

## Why it matters

The commit was left with the `pending` status the run posted when it started, and nothing ever clears it. That is worse than a plain failure in two ways:

- `Attestation` renders a pending sign-off as **fail**, so the PR looks like it has a red check with no failing run behind it.
- Every tool that reads the status — `signoff mine`, the PR-sweep skills — sees "sign-off in flight" and waits instead of re-dispatching. #12652 sat in that state for hours with no live run. I cleared it by hand by posting the verdict the run had computed.

Any run long enough to outlive its credentials lands here, and those are precisely the runs whose verdict cost the most to produce.

## Root cause

`post_commit_status` in `scripts/signoff` makes a single unretried call:

```bash
post_commit_status() {
  local repo="$1" sha="$2" state="$3" description="$4"
  description="${description:0:140}"
  debug "posting ${STATUS_CONTEXT}=${state} on ${repo}@${sha}: ${description}"
  gh api --method POST "repos/${repo}/statuses/${sha}" \
    -f state="$state" \
    -f context="${STATUS_CONTEXT}" \
    -f "description=${description}" >/dev/null
}
```

One transient response — a 401 like this one, a 5xx, a network blip — discards a multi-hour result permanently, and the fallback path inherits the same weakness because it calls the same function.

## Suggested fix

1. Retry `post_commit_status` with backoff (a handful of attempts over a minute or two) rather than posting once. This alone would have saved the run above if the 401 was transient.
2. If the 401 is instead credential *expiry* on a long job, retrying will not help and the token needs refreshing before the post — worth confirming which of the two this was before choosing. A cheap way to tell them apart: have the retry log the response body for each attempt.
3. Make a status post that ultimately fails loud in the job summary, so the run does not merely warn into the log about a state no later reader can distinguish from a live run.

Related but not the same: #12424 / #12428 cover a *cancelled* run's status, where the run never reaches a verdict. This one reaches a verdict and loses it in transit.

grokspice:
## Summary

`post_commit_status` in `scripts/signoff` made a single unretried `gh api` call. That call is the
only place a sign-off verdict exists, so one transient response discarded it permanently.

The sign-off for #12652 ([run 31133614812](https://github.com/spiceai/spiceai/actions/runs/31133614812))
ran the full gate for 10,089s, reached a real verdict, and got `gh: Bad credentials (HTTP 401)`
posting it. The `Resolve an incomplete sign-off` fallback posts through the same helper and hit the
same 401. The commit was left carrying only the `pending` posted on the way in — which `Attestation`
renders as a failure, and which `signoff mine` and the PR-sweep tooling read as a run still in
flight, so nothing re-dispatches. That is strictly worse than a plain failure, and it happens to
exactly the runs whose verdict cost the most to produce.

Fixes #12701

## Changes

All three in `post_commit_status`, so every caller is covered at once — the pending on the way in,
the failing verdict, the success attestation, and the `clear-pending` fallback that inherited the
same weakness by calling the same function.

- **Retry with backoff.** `SIGNOFF_STATUS_POST_BACKOFF` (default `5 15 30 60`) gives four waits and
  therefore five attempts over ~110s. Empty means post once, which is what the tests use.
- **Report the response body on every attempt.** The issue asks for this explicitly, and it is the
  only thing that separates the two diagnoses: `Bad credentials` on a job that outlived its token is
  beyond what retrying can fix and needs a fresh token, while a 5xx or a network blip is exactly
  what the retry is for. Without the body the log cannot tell you which of the two you are looking
  at, and this run's 401 is still unattributed for that reason.
- **Say it in the job summary when the post never lands** — `Verdict not recorded … Any pending left
  on this commit is stale, not a run in flight: sign off again`. A warning in the log of a finished
  job does not reach anyone reading the PR, which is where the misreading happens.

  A failed **pending** post deliberately does not raise this: it strands nothing, because the commit
  keeps whatever it had and the run still has its own verdict to post. Covered by a test so the two
  cannot drift together.

## What this does not fix

If the 401 was token **expiry** rather than a transient blip, five attempts over ~110s will not
recover it — the issue says as much. This change makes that case diagnosable (the body is now
logged) and loud (the summary says the pending is stale) instead of silent. Refreshing the token
mid-run is a separate change to the workflow, and which of the two it is should be read off the next
occurrence rather than assumed.

Worst case this adds ~110s to a run that is already failing, on the failure path only.

## Test plan

`bash scripts/test_signoff_status.sh` — 33 assertions, run locally, all pass. CI runs it via
`.github/workflows/signoff_script_test.yml`, which triggers on both files this PR touches.

Six new assertions, all driving the real `post_commit_status` against a `gh` stub that fails its
next N posts:

- a verdict survives a transient failure
- a verdict survives failures up to the last attempt
- a verdict that never lands says the pending is stale
- the response body is reported, not just the failure
- an empty backoff posts exactly once
- a lost pending does not claim a stale status

Verified against the previous `scripts/signoff`: 4 of the 6 fail on it and all 6 pass on this one.
(The response-body assertion also passes on the old code, which let the body leak to the terminal
untagged; it is kept to lock the behaviour in now that the body is captured and attributed.)

Also re-ran the sibling suites, all passing and unchanged: `test_signoff_attestation_refresh.sh`
(15), `test_signoff_cancelled_correction.sh` (18), `test_signoff_disk_guard.sh` (48),
`test_signoff_resolve_target.sh` (22). `shellcheck -S warning` is clean on both changed files.

While writing the tests, the first version of the new helper declared a local named `failures`,
shadowing the suite's own counter — `fail_test`'s increment became an arithmetic error, so failing
assertions reported nothing and the suite still printed "all 32 tests passed". That is what the
old-code check above caught. The local is renamed and carries a comment saying why.

## Checklist

- [x] Fixed in the shared helper, so the fallback path cannot inherit the weakness again
- [x] Regression assertions fail on the previous script
- [x] Sibling sign-off suites re-run
- [x] Non-Rust change: `scripts/signoff` is not in `RUST_AFFECTING_PATH_PATTERN`, so the Rust gate is skipped

### DIFF
--- a/scripts/signoff
+++ b/scripts/signoff
@@ -68,6 +68,12 @@ readonly VERSION="1.5.0"
 readonly STATUS_CONTEXT="signoff"
 readonly TARGET_BRANCH="${SIGNOFF_BRANCH:-trunk}"
 readonly TARGET_REMOTE="${SIGNOFF_REMOTE:-origin}"
+# Seconds to wait between attempts at a commit-status POST; the number of
+# attempts is one more than the number of entries. A verdict can cost hours of
+# runner time and exists nowhere but that POST, so one transient response must
+# not discard it. Empty means post once. Overridable so the unit tests do not
+# pay for the waits.
+readonly STATUS_POST_BACKOFF="${SIGNOFF_STATUS_POST_BACKOFF-5 15 30 60}"
 DEBUG="${SIGNOFF_DEBUG:-}"
 VCS=""
 JJ_REMOTE_URL=""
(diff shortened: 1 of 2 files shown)


grokspice:
@copilot review

lukekim (approved):

PR #12713, opened 2026-08-07
fix(google-genai): resume the SSE scan where it stopped, not at the buffer start (fixes #12689)

### LINKED ISSUE #12689: The Gemini SSE reader rescans its whole buffer on every read, so a slow unterminated event costs quadratic CPU
## Summary

`find_event_boundary` in `crates/google-genai/src/generate.rs` restarts its scan at `i = 0` on
every call, and `parse_sse_stream` calls it after every network read. A peer that dribbles an
unterminated event in very small chunks therefore forces one full rescan per read, over a buffer
that grows with each one.

## Why it is filed rather than folded into #12600

#12600 bounds the buffer at `MAX_EVENT_BYTES`, so the rescan is no longer unbounded — which is
what makes this tractable at all. But the bound leaves the work quadratic in the bytes received
up to that cap: an 8 MiB event delivered a few bytes at a time is on the order of 10^13 byte
comparisons before the cap fires, driven by a modest amount of traffic. That is a CPU cost under
the same threat model the cap addresses, and it is reachable inside `spiced` via `crates/llms`.

The fix is separate from the cap and does not change what the reader accepts, which is why it is
not folded into that change.

## Suggested direction

Carry the offset already scanned on `SseReader` and resume from `max(0, scanned_upto - 1)` — one
byte back, so a `\r` that was last in the buffer is re-examined once its successor arrives. The
existing framing tests cover the boundary cases that offset has to respect (`\r\n` split across
reads, a bare `\r` last in the buffer, a blank line split across reads).

## Notes

Found by a security review of the #12600 fix, which bounds the same buffer. Pre-existing — the
scan has restarted from zero since the reader was written.

grokspice:
## Summary

`find_event_boundary` restarted its scan at index 0 on every call, and `parse_sse_stream` calls it after every network read. A peer that dribbles one unterminated event in very small chunks therefore forced one full rescan per read, over a buffer that grew with each one — work quadratic in the bytes received.

#12600 (PR #12690) caps that buffer at `MAX_EVENT_BYTES`, so the rescan is no longer unbounded, but the work up to the cap is still quadratic: an 8 MiB event delivered a few bytes at a time is on the order of 10^13 byte comparisons, driven by a modest amount of traffic, inside `spiced` via `crates/llms`. This is the CPU half of the same threat model the cap addresses.

## Root cause

The scan was stateless. Nothing recorded that the leading bytes of the buffer had already been rejected, so every read re-examined all of them.

## Changes

- `SseReader` carries `scanned`, the number of leading bytes known to begin no event boundary. `find_event_boundary` takes a `from` offset; `SseReader::next_boundary` passes `scanned` and updates it when a scan finds nothing.
- The rewind on an unsuccessful scan is **two** bytes, not none. `terminator_at` reads the byte after the one it classifies, so the last two positions are the only ones a later read can reclassify — a bare `CR` last in the buffer, and a terminator whose blank-line successor has not arrived yet. Rewinding only one byte frames the wrong event end when a `CRLF` pair straddles the resume point; the added test catches exactly that.
- Both front-drop sites (an event consumed, and the `LF` of a `CRLF` split across two reads) go through `consume`, and both buffer-clear sites through `discard`, so the offset cannot drift away from the bytes it names.

What the reader accepts is unchanged — only how many times it looks at each byte.

## Test plan

- `make lint`
- `make test`

New tests in `crates/google-genai/src/generate.rs`:

- `a_resumed_scan_frames_the_same_events_as_one_from_the_start` — grows six differently-framed bodies one byte at a time, putting the resumption point at every offset in turn, and asserts a resumed scan agrees with a full scan at every step. Verified to **fail** when the rewind is weakened to one byte (`end: 8` instead of `end: 7` on `data: a\r\n\r`), so it pins the boundary the fix depends on rather than just passing.
- `an_unterminated_event_is_not_rescanned_from_the_start` — the regression guard: repeated appends without a terminator leave only the two reclassifiable positions to re-examine, so one more read does not cost more as the buffer grows.
- `a_body_delivered_one_byte_at_a_time_reads_the_same_as_one_delivered_whole` — end-to-end through `parse_sse_stream`, putting a read boundary at every point at once, including a comment event ahead of the data events.

Locally on the pinned 1.96.1 toolchain: `cargo test -p google-genai --lib` → 24 passed, 0 failed (21 pre-existing + 3 new); `cargo clippy -p google-genai --lib --tests --all-features -- -D warnings` clean. (`--all-targets` reports a pre-existing unfulfilled `#[expect(clippy::too_many_lines)]` in `examples/function_calling.rs`, present on the unmodified base and untouched here.)

## Note on the base

This targets `fix/12600-genai-sse-event-cap` (PR #12690), not `trunk`: `find_event_boundary` and `SseReader` are introduced by that PR and do not exist on `trunk`, so the change cannot be expressed against it. GitHub will retarget this to `trunk` when #12690 merges. Being stacked, it will show absent rather than failing checks until then.

## Checklist

- [x] Tests added covering the reported failure mode
- [x] Regression test verified to fail against the weakened invariant
- [x] `cargo fmt` clean
- [x] Clippy clean on the changed code
- [ ] Sign-off (blocked on #12436 — see below)

Fixes #12689


grokspice:
@copilot review

lukekim (dismissed):

phillipleblanc (approved):

PR #12715, opened 2026-08-07
fix(cli): report an answer the model stopped early, instead of printing it as whole (fixes #12596)

### LINKED ISSUE #12596: spice chat reports a content-filtered answer as complete, because the CLI never reads finish_reason

grokspice:
## Summary

`spice chat` never deserialized `finish_reason` from a streaming chunk, so a stream that ended because the model **stopped early** was presented exactly like one that finished.

The OpenAI streaming schema puts the reason the generation stopped on the choice: `stop`, `length`, `content_filter`, `tool_calls`, or `function_call`. Only `stop` means the answer is whole. The two that bite:

- `content_filter` — Azure OpenAI ends the stream when its filter trips. The tokens already sent stay on screen and the CLI exits `0`.
- `length` — the answer hit `max_tokens` and is cut off mid-sentence.

In both cases the user saw a partial answer with no indication it was partial, and `--output json` recorded it as the full `content`.

Fixes #12596

## Changes

- `ChunkChoice` deserializes `finish_reason`; intermediate chunks send `null`, which reads as `None`.
- `StreamState` carries the last reported reason through to `ChatResponse`.
- `ChatResponse::incomplete_notice()` renders the notice. `stop` and an absent reason produce nothing; every other reason is reported after the answer has been printed.
- The notice is written to **stderr**, so it reaches a terminal without joining answer text a caller captures on stdout. It is deliberately not an error: the tokens received are genuine and worth keeping.
- `--output json` gains `finish_reason`, so a consumer reading `content` can tell a truncated answer from a complete one.
- The reason is a closed set of tokens, so a value outside that shape is described rather than echoed — the field is server-controlled and the notice is printed straight to the user's terminal.

## Test plan

Five tests added, all driving the production `send_chat_streaming` path against the existing `SlowServer` fixture:

- `a_content_filtered_answer_is_reported_as_incomplete` — the reported failure; asserts the tokens are kept **and** the answer is called incomplete.
- `an_answer_cut_off_by_the_token_limit_is_reported_as_incomplete` — the `length` case.
- `a_whole_answer_is_not_reported_as_incomplete` — negative control; `stop` must stay silent, or every successful chat warns.
- `an_answer_with_no_finish_reason_is_not_reported_as_incomplete` — an absent reason is not an early stop.
- `an_unreasonable_finish_reason_is_described_and_not_echoed` — a terminal escape sequence in the field is described, not written to the terminal.

- [x] `cargo fmt --all -- --check`
- [ ] `make lint`
- [ ] `make test`

## Notes

Based on `fix/12583-chat-stream-timeout` (#12589), which carries the SSE-framing rewrite from #12588 that this builds on. GitHub will retarget this to `trunk` once that lands.

`bin/spice` could not be compiled on the machine that wrote this — `runtime-cloud-connect` needs `protoc`, which is not installed — so the notice logic was validated standalone and the full build is left to sign-off.

### DIFF
--- a/bin/spice/src/commands/chat.rs
+++ b/bin/spice/src/commands/chat.rs
@@ -170,6 +170,13 @@ struct ChunkChoice {
     /// with nothing to print, not an unreadable one.
     #[serde(default)]
     delta: Option<Delta>,
+    /// Why the model stopped producing this choice, on the last chunk that carries it.
+    ///
+    /// Intermediate chunks send `null` here, which reads as `None`: the answer is still being
+    /// produced and nothing has ended it. A value means the generation is over, and only
+    /// `stop` means it is over because the model had said everything it meant to.
+    #[serde(default)]
+    finish_reason: Option<String>,
 }
 
 /// Delta content in a streaming response.
@@ -196,6 +203,8 @@ struct ChatResponse {
     total_duration: std::time::Duration,
     first_token_duration: Option<std::time::Duration>,
     usage: Option<Usage>,
+    /// The reason the last chunk gave for the generation ending, when it gave one.
+    finish_reason: Option<String>,
 }
 
 #[derive(Serialize)]
@@ -205,6 +214,9 @@ struct ChatJsonResponse<'a> {
     duration_ms: u128,
     first_token_ms: Option<u128>,
     usage: Option<&'a Usage>,
+    /// Present so a consumer reading `content` can tell a whole answer from a truncated one
+    /// without parsing the notice written to stderr.
+    finish_reason: Option<&'a str>,
 }
 
 impl<'a> ChatJsonResponse<'a> {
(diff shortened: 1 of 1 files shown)


grokspice:
@copilot review

claudespice:
Merged the base branch `fix/12583-chat-stream-timeout` in (`4cda1a54a`) — this branch had diverged from it by one commit and that commit is the one it needed.

The sign-off on `4598c8a7d` ([run 31164957150](https://github.com/spiceai/spiceai/actions/runs/31164957150)) failed to compile:

```
error[E0063]: missing fields `inference_deadline` and `inference_http_client` in initializer of `context::RuntimeContext`
error: could not compile `spice` (lib test) due to 1 previous error
```

That break came from the parent branch, not from this PR's own change, and the parent fixed it in `f72ef2cac` at 08:50Z — after this branch was cut at 07:35Z. So the sign-off was building a two-hour-old copy of the parent and had no way to pass.

The merge is the parent's two-line test-literal fix and nothing else; the diff against the base is still only this PR's own 212 lines in `bin/spice/src/commands/chat.rs`.

No sign-off dispatched: this PR is stacked on #12589, so it gets no `Attestation` until that merges and GitHub retargets this to `trunk` ([the stacked-PR case](https://github.com/spiceai/spiceai/pull/12713)), and the sign-off pool is currently saturated. It wants one fresh run once #12589 lands.

lukekim (dismissed):

lukekim (approved):

Issue #12718, opened 2026-08-07
The remaining merge-queue-gating workflows still have unbounded jobs


grokspice:
## Summary

#12717 covers `e2e_test_ci.yml`, where an unbounded `build` job held four merge-queue candidates for hours. The same shape is in the other three merge-queue-gating workflows: their long jobs declare no `timeout-minutes`, so a wedged one is held for GitHub's default of 360 minutes and the queue waits behind it.

## The unbounded jobs, with measured durations

Successful `merge_group` runs, most recent ten of each:

| Workflow | Job | Observed minutes | Max |
| --- | --- | --- | --- |
| `pr.yml` | Build and Test | 71, 97, 101, 138, 140, 146, 155, 156 | 156 |
| `pr.yml` | Rust Lint | 73, 74, 75, 77, 78, 81, 103, 138, 147 | 147 |
| `pr.yml` | Build (release profile) | 51, 54, 60, 63, 64, 71, 93, 101, 134 | 134 |
| `integration.yml` | Build Test Binary | 87, 94, 111, 111, 112, 114, 115, 118, 121 | 121 |
| `integration.yml` | Integration Tests (part 1) | 7, 7, 9, 9, 9, 10, 11 | 11 |
| `integration.yml` | Integration Tests (part 2) | 3, 3, 3, 4, 4, 4, 4 | 4 |
| `integration.yml` | Integration Tests (part 3) | 2, 3, 4, 4, 4, 5, 6 | 6 |
| `integration_llms.yml` | Build Test Binary | 4, 5, 5, 5, 5, 6, 7, 7, 7, 8 | 8 |

Also unbounded, and quick enough that any minutes-long run is a stuck runner rather than slow work: `pr.yml`'s `attestation` and `check_changes`; `integration.yml`'s `check_changes`, `upload-integration-snapshots`, `test-aws-sdk`, `test-aws-sdk-credential-bridge`, `test-snapshot-refresh`, `test-data-components`, `test-runtime-partition`; `integration_llms.yml`'s `setup-matrix` and `test`.

## Why this was not folded into #12717

Two reasons, and either alone would be enough:

- Each job needs its own measured bound. The spreads are wide — `Build and Test` varies by better than a factor of two — so a shared constant would either fail legitimate runs or be too loose to help.
- `pr.yml` and `integration_llms.yml` have open PRs against them (#12681 and #12626 respectively). Editing the same files in parallel buys a conflict for no gain.

[...]

Issue #12420, opened 2026-08-03
CI: a mid-build sccache backend outage fails Remote Sign-off as 'error: could not compile', blaming the branch


grokspice:
## Summary

`signoff.yml` starts the `spiceio` S3 proxy as a background sidecar, points `sccache` at it, and
sets `RUSTC_WRAPPER=sccache` for a build+test run that takes **one to three hours**. If `spiceio`
stops answering at any point in that window, `sccache` cannot start its server, and because it is
the rustc wrapper **every remaining rustc invocation fails**. Cargo reports that as
`error: could not compile <crate>` — for dozens of crates at once — so a cache outage is presented
as a compile failure in the branch under sign-off, leaving `Attestation` red and the PR unable to
enter the merge queue.

A cache is an optimization. Losing it should cost minutes, not a red gate, and it should never
produce a message that points at the code.

## Signature

```
sccache: error: Server startup failed: cache storage failed to read: Unexpected (temporary) at read => send http request
   error sending request for url (http://127.0.0.1:<port>/sccache/sccache/.sccache_check):
     client error (Connect): tcp connect error: Connection refused (os error 61)
error: could not compile `connector-ftp` (lib test)
error: could not compile `tpc-extension` (lib test)
error: could not compile `connector-ducklake` (lib test)
...
error: command `cargo test --no-run ...` exited with code 101
```

## Blast radius

Sample: the **40 most recent `Remote Sign-off` failures**, `2026-07-31T18:00Z → 2026-08-03T16:13Z`
(117 failures in the last 7 days, against 123 successes — this workflow is the `Attestation` gate,
so each failure blocks one PR).

| Cause | Runs | Share |
|---|---:|---:|
| **This bug** (`sccache: Server startup failed` → `could not compile`) | **9** | **22.5%** |
| Real code failure (`error[E….]` or `test result: FAILED`) | 7 | 17.5% |
| Runner disk full at the linker (already filed as #12412) | 1 | 2.5% |
| `is cmake not installed` (fixed by #12286, both predate it) | 2 | 5% |
| Not classified from `--log-failed` alone | 21 | 52.5% |

The nine confirmed runs:
[30791412731](https://github.com/spiceai/spiceai/actions/runs/30791412731),
[30788119170](https://github.com/spiceai/spiceai/actions/runs/30788119170),
[30763254868](https://github.com/spiceai/spiceai/actions/runs/30763254868),
[...]

PR #12719, opened 2026-08-07
fix(ci): bound every E2E job, so a wedged build cannot hold the queue (fixes #12717)

### LINKED ISSUE #12717: A wedged merge-queue build job holds the whole queue for GitHub's 6-hour default
## Summary

`e2e_test_ci.yml`'s `build` job declares no `timeout-minutes`, so a wedged build is held for GitHub's default of **360 minutes**. When the run is a `merge_group` one, that is not just a wasted job — the queue candidate stays `AWAITING_CHECKS` for the whole six hours, and every entry behind it waits.

Every one of the workflow's ~20 test jobs is bounded (5-10 minutes). `build` and the three coordinator jobs around it are the only unbounded ones, and `build` is the one that takes hours.

## Evidence

Run [31144642239](https://github.com/spiceai/spiceai/actions/runs/31144642239), `merge_group` on `gh-readonly-queue/trunk/pr-12534-...`, job `Build macOS aarch64 (Apple Silicon) binaries`:

| Step | Status | Started |
| --- | --- | --- |
| 11. Build spiced and spice CLI | skipped | — |
| 12. **Build spiced and spice CLI (macOS)** | **in_progress** | 2026-08-07T04:19:27Z |
| 13. Show sccache stats | pending | — |
| 14. Kill spiceio | pending | — |

Setup steps 1-10 took 12 seconds in total. Step 12 — the cargo build — had been running for over 4.5 hours when this was filed, with nothing after it able to start. Its log blob is not retrievable while the step runs (`BlobNotFound`), so there is no output to show and no way to tell from outside where in the compile it stopped.

The merge queue at that moment:

```
total=24
pos=1  AWAITING_CHECKS  #12534  enqueued 00:10   <- waiting on this job
pos=2  AWAITING_CHECKS  #12586  enqueued 00:10
pos=3  AWAITING_CHECKS  #12601  enqueued 00:10
pos=4  AWAITING_CHECKS  #12607  enqueued 00:10
pos=5..24  QUEUED
```

Four candidates had been `AWAITING_CHECKS` for eight and a half hours, and 20 more were behind them. Nothing merged in that window.

## Why a bound is the fix, and what it is not

The hang itself is a separate question — the same job succeeds routinely on the same runners, and the self-hosted macOS pool puts several runner instances on one host, so a contended or stuck compile is plausible without anything being wrong with the commit. Bounding it does not prevent the hang and does not rescue a run already in one; the workflow file is read when a run starts.

What it does is cap the blast radius. A bound turns "the queue is stalled for six hours" into "one candidate fails, is dequeued, and the rest proceed", which is the difference between a wedge costing the team a morning and costing it one re-enqueue.

## Duration evidence for choosing a bound

Successful `merge_group` runs of this job, most recent 10:

| Job | Observed minutes | Max |
| --- | --- | --- |
| Build macOS aarch64 (Apple Silicon) binaries | 74, 80, 85, 86, 90, 91, 138 | 138 |
| Build Linux x64 binaries | 44, 44, 60, 60, 63, 63, 63, 64, 65 | 65 |

The spread is wide — the same job varies by a factor of two on a contended pool — so a tight bound would fail legitimate builds, which is worse than a slow one. A bound has to sit clearly above 138 minutes.

## Expected behaviour

The build step carries a bound above its observed worst case, with a job-level backstop above that so the diagnostic steps after it (`Show sccache stats`, `Kill spiceio`, `Upload spiceio logs`) still get to run when the step is cut — a wedged compile is exactly when the sccache stats are worth having. The trivial coordinator jobs are bounded too, since a hung `setup-matrix` blocks the queue just as effectively as a hung build and has no reason to run for minutes.

## Scope

`e2e_test_ci.yml`. The same shape exists in the other merge-queue-gating workflows and is worth the same treatment, but each needs its own measured values and two of the files have open PRs against them, so that is tracked separately rather than folded in here.

grokspice:
## Summary

`e2e_test_ci.yml`'s `build` job declared no `timeout-minutes`, so a wedged compile was held for GitHub's default of **360 minutes**. On a `merge_group` run that is not merely a wasted job — the queue candidate stays `AWAITING_CHECKS` for the whole six hours, and every entry behind it waits.

Run [31144642239](https://github.com/spiceai/spiceai/actions/runs/31144642239) was sitting on step 12, `Build spiced and spice CLI (macOS)`, having started it at 04:19:27Z. Setup steps 1-10 had taken 12 seconds in total; nothing after step 12 could start. The queue behind it:

```
total=24
pos=1  AWAITING_CHECKS  #12534  enqueued 00:10   <- waiting on that job
pos=2  AWAITING_CHECKS  #12586  enqueued 00:10
pos=3  AWAITING_CHECKS  #12601  enqueued 00:10
pos=4  AWAITING_CHECKS  #12607  enqueued 00:10
pos=5..24  QUEUED
```

Every one of the workflow's ~17 test jobs was already bounded at 5-10 minutes. `build` and the four jobs around it were the only unbounded ones, and `build` is the one that takes hours.

## Changes

Bounds go on the **compile steps** rather than only the job, so that a wedged step is cut while the `always()` steps after it — `Show sccache stats`, `Kill spiceio`, `Upload spiceio logs` — still run. A hung compile is exactly when the sccache stats are worth having (#12420 made that step run on failure for the same reason). The job then carries a backstop above the longest step bound, with room for the artifact upload and the post steps.

| Scope | Bound | Observed on the last ten successful `merge_group` runs |
| --- | --- | --- |
| step: Build spiced and spice CLI (Linux) | 90m | 44, 44, 60, 60, 63, 63, 63, 64, 65 |
| step: Build spiced and spice CLI (macOS) | 180m | 74, 80, 85, 86, 90, 91, 138 |
| job: `build` | 200m | — backstop above the step bounds |
| job: `check_changes` | 10m | full-history checkout plus one composite action |
| job: `setup-matrix` / `setup-model-matrix` | 5m | one `github-script` step returning a literal |
| job: `e2e-gate` | 5m | reads results it already has |

Each build bound is the slowest observed success with clear room over it. The spread is wide — the same job varies by a factor of two — because the self-hosted macOS pool puts several runner instances on one host, so the bounds sit clear of the median rather than near it. A bound tight enough to fail a legitimate build would be worse than the slow build it replaced.

`e2e-gate` gets a bound despite doing no work because it is the *required* check: a stuck gate blocks the queue even when every job it summarises has finished.

## What this does not do

It does not stop the hang. The same job succeeds routinely on the same runners, its log blob is not retrievable while the step runs (`BlobNotFound`), and there is no output to attribute it from — a contended or stuck compile on a shared host is plausible without anything being wrong with the commit. It also does not rescue a run already wedged, since a run reads the workflow as of its own start.

What it does is turn "the queue is stalled for six hours" into "one candidate fails, is dequeued, and the rest proceed".

## Test plan

- `make lint`
- `make test`
- `actionlint .github/workflows/e2e_test_ci.yml` reports nothing on the added lines. Its existing output for this file (unknown self-hosted runner labels, shellcheck style notes, one duplicate matrix value) is unchanged, and is pre-existing.
- The diff is additive only — `timeout-minutes` keys and comments, no step, condition or command altered. Verified by filtering the diff for non-comment, non-`timeout-minutes` lines, which yields nothing.
- Parsed the result and asserted all 21 jobs now carry a `timeout-minutes`, and that the two compile steps carry 90 and 180.

A timeout cannot be exercised without deliberately wedging a build on the shared macOS pool, which would cost the queue the very stall this prevents. The values are therefore justified from measured run history above rather than from a test.

## Checklist

- [x] Root cause identified, with the live queue stall and the hung step named
- [x] Bounds derived from measured durations, not guessed
- [x] Step bounds placed so the `always()` diagnostics survive the cut
- [x] Every job in the workflow now bounded, including the required gate
- [x] Diff verified additive; no behaviour changed on a passing run
- [x] The limits of the fix stated rather than overclaimed
- [ ] `make lint` / `make test` via sign-off

Fixes #12717

Follow-up: the same shape exists in `pr.yml`, `integration.yml` and `integration_llms.yml`, whose long jobs are also unbounded (`Build and Test` 156m, `Rust Lint` 147m, `Build (release profile)` 134m, `Build Test Binary` 121m observed). Not folded in here: each needs its own measured values, and `pr.yml` and `integration_llms.yml` have open PRs against them (#12681, #12626). Tracked in #12718.

### DIFF
--- a/.github/workflows/e2e_test_ci.yml
+++ b/.github/workflows/e2e_test_ci.yml
@@ -61,6 +61,11 @@ jobs:
   check_changes:
     if: ${{ github.event_name != 'pull_request' }}
     runs-on: spiceai-dev-runners
+    # Every job here is bounded, because on a `merge_group` run an unbounded one does not
+    # just waste a job: the queue candidate stays `AWAITING_CHECKS` for GitHub's 6-hour
+    # default and every entry behind it waits (#12717). A full-history checkout is the
+    # slow part of this one.
+    timeout-minutes: 10
     outputs:
       relevant_changes: ${{ steps.check.outputs.relevant_changes == 'true' || github.event_name == 'workflow_dispatch' }}
     steps:
@@ -76,6 +81,9 @@ jobs:
     name: Setup strategy matrix
     if: ${{ github.event_name != 'pull_request' }}
     runs-on: spiceai-dev-runners
+    # One `github-script` step returning a literal, so anything past a minute is a stuck
+    # runner rather than slow work.
+    timeout-minutes: 5
     outputs:
       matrix: ${{ steps.setup-matrix.outputs.result }}
 
@@ -109,6 +117,11 @@ jobs:
     runs-on: ${{ matrix.target.builder }}
     needs: [setup-matrix, check_changes]
     if: ${{ needs.check_changes.outputs.relevant_changes == 'true' }}
+    # A backstop above the longest step bound below, leaving room for the artifact upload
+    # and the post steps. The compile steps carry their own tighter bounds so that a
+    # wedged one is cut while the `always()` diagnostics after it can still run — a hung
+    # compile is exactly when the sccache stats are worth having.
+    timeout-minutes: 200
     env:
       HAS_SPICEIO_SECRET: ${{ secrets.UNAS_SMB_PASS != '' }}
 
@@ -165,6 +178,11 @@ jobs:
 
       - name: Build spiced and spice CLI
         if: matrix.target.target_os != 'darwin'
+        # 65 minutes is the slowest this leg has taken across the last ten successful
+        # `merge_group` runs (44-65), so this is that worst case with room to spare. The
+        # spread is wide on a contended pool, and a bound tight enough to fail a
+        # legitimate build would be worse than the slow build it replaced.
+        timeout-minutes: 90
         uses: ./.github/actions/build-spiced
         with:
           features: 'tpc-extension,models,postgres-accel'
@@ -174,6 +192,12 @@ jobs:
 
       - name: Build spiced and spice CLI (macOS)
         if: matrix.target.target_os == 'darwin'
+        # This is the step that wedged for over 4.5 hours and held four queue candidates
+        # behind it (#12717). 138 minutes is its slowest successful `merge_group` run of
+        # the last ten (74-138); the self-hosted macOS pool puts several runner instances
+        # on one host, which is where that factor-of-two spread comes from, so the bound
+        # sits well clear of it rather than close to the median.
+        timeout-minutes: 180
         uses: ./.github/actions/build-spiced
         with:
           features: 'tpc-extension,models,postgres-accel,metal'
@@ -207,6 +231,7 @@ jobs:
     name: Setup model strategy matrix
     if: ${{ github.event_name != 'pull_request' }}
     runs-on: ubuntu-24.04
+    timeout-minutes: 5
     outputs:
       matrix: ${{ steps.setup-matrix.outputs.result }}
     steps:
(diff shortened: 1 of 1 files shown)


grokspice:
@copilot review

lukekim (approved):

PR #12736, opened 2026-08-07
fix(ci): ask whether the disk watcher runs before make writes to it (fixes #12734)

### LINKED ISSUE #12734: Flaky: test_signoff_disk_guard.sh 'a dead watcher still reports make's own status' races SIGPIPE
Observed while working on #12724 (the test file is not part of that diff — the case comes from #12461 and is present on `trunk`).

## Symptom

`scripts/test_signoff_disk_guard.sh` intermittently fails one case:

```
FAIL: a dead watcher still reports make's own status: expected RC=101, got 'warning: the out-of-disk watcher exited 2 rather than reporting a verdict — falling back to free-space classification
RC=141
VERDICT=
ARMED=
KIND=signalled'
```

## Reading

`141` is `128 + 13` — SIGPIPE. The case deliberately kills the watcher so it can assert that `run_make_step` still reports `make`'s own status (101). When the watcher dies before `make` finishes writing to the pipe it feeds, `make` takes SIGPIPE and dies at 141 instead of returning 101, and the assertion sees the signal status rather than the recipe status.

The knock-on is that `failure_kind` then classifies the run as `signalled` (141 >= `EXIT_SIGNAL_BASE`), which is correct for a 141 but not what the case is exercising.

## Rate

Measured locally on Windows/Git Bash, 34 runs of the changed tree and 18 of pristine `trunk`-based HEAD: 3 failures in 34 and 0 in 18 respectively — consistent with a single underlying race at roughly 5-10%, not with a difference between the trees. The failing assertion also runs strictly before the only case #12724 adds, and never enters the function that PR changes.

## Suggested direction

Make the ordering deterministic rather than widening the assertion — the case is worth keeping exactly as specified. Options: have the stub `make` finish writing before the watcher is killed, ignore SIGPIPE for the stub, or drain the pipe on the watcher's way out so the writer never sees a closed reader.

Not fixing it here: it is unrelated to the diff that surfaced it, and the assertion is testing the right thing.

grokspice:
Fixes #12734

## Summary (root cause)

`run_make_step` pipes `make` into an `awk` watcher that scans build output for the out-of-disk signature. The watcher sits **downstream of a process that is already writing to it**, so a watcher that cannot run exits immediately, closes the pipe under a still-writing `make`, and `make` dies of SIGPIPE. The step then reports `128 + 13 = 141` instead of `make`'s own status, and `failure_kind` classifies a genuine compile failure as a *signalled* run rather than a judged one.

`scripts/test_signoff_disk_guard.sh`'s "a dead watcher still reports make's own status" case was racing exactly that shape. Its stub `make` writes a single short line, so whether it finished writing before the watcher closed the pipe was a coin flip — reported at 3 failures in 34 runs on the issue, and reproduced here.

The flake and the production defect are the same bug. Widening the assertion would have hidden it: the case is asserting the right thing, and the code was not delivering it.

## Changes

- **`scripts/signoff`** — probe the watcher over empty input *before* `make` writes a byte, while nothing is downstream of it. A usable watcher reads the empty stream, matches nothing, and exits 0; an unusable one (no `awk` on the runner, a variant that rejects the program) reports its own status there. The step then runs `make` unwatched and disarms the watch, falling back to free-space classification — the same fallback the mid-run-death path already takes.
- The watcher program moves into a `DISK_WATCH_PROGRAM` constant so the probe runs *exactly* the program that will read the build output; a probe of anything else predicts nothing.
- `disarm_disk_watch` holds the give-up path, so the two places a watcher can fail to deliver a reading cannot drift into describing the same outcome differently.

**This bounds a class, it does not eliminate the hazard.** A watcher that dies *mid-stream* can still take `make`'s status with it, and the probe cannot prevent that. What it removes is the class where the watcher was never going to run at all — the one that costs `make`'s status on every step rather than on the step that raced. The comment says that rather than claiming elimination.

## Behavior on the healthy path

Unchanged. The probe adds one `awk` invocation over `/dev/null` per `make` step (3 per sign-off run); when it passes, the pipeline is built exactly as before, and output still streams to the Actions log line by line. Local runs (`SIGNOFF_DISK_WATCH` unset) never reach the probe at all.

## Test plan

- `bash scripts/test_signoff_disk_guard.sh` — 49 assertions, all passing.
- New case: a `make` that keeps writing (2000 lines) under a watcher that cannot run. The old ordering loses the status **every time** rather than occasionally — verified failing at `RC=141` against the unmodified `scripts/signoff`, and passing after. This is the regression test.
- The existing case is kept exactly as specified and now passes deterministically: **40/40 local runs clean**, against 3 failures in 34 reported on the issue.
- All five sign-off suites pass (`test_signoff_disk_guard`, `test_signoff_status`, `test_signoff_attestation_refresh`, `test_signoff_cancelled_correction`, `test_signoff_resolve_target`).
- `shellcheck -S warning` clean on both files.
- `make lint` and `make test` via the standard CI gate.

## Checklist

- [x] Existing assertions kept as specified, not widened
- [x] `shellcheck` clean
- [x] Healthy path unchanged
- [x] Residual hazard documented rather than claimed fixed

## Notes

- Touches `scripts/test_signoff_disk_guard.sh`, which #12724 also edits — the two changes are in non-overlapping regions (this at ~line 300, #12724 appending from ~line 378) and merge cleanly in either order.
- Adversarial cross-model review was skipped: the `codex` engine reports the workspace is out of credits. Security review of the diff found nothing — the three `awk` inputs are `readonly` script constants with no user input, the program is single-quoted, and `make "$@"` is quoted as before.


grokspice:
@copilot review

lukekim (approved):

Issue #12737, opened 2026-08-07
No check catches a source file that no crate root declares, so it silently ships nothing


grokspice:
# Problem

Nothing in the build or the lint gate notices when a file under a crate's `src/` tree is not
reachable from that crate's root through a `mod` declaration. Such a file is invisible to
`cargo build`, `cargo clippy`, `cargo fmt --all` and `cargo test` alike — it looks like source,
it gets edited like source, and it ships nothing.

#12735 reported one instance. Sweeping the whole workspace while fixing it found **three**:

| File | Lines | Tests never run | Superseded by |
| --- | --- | --- | --- |
| `crates/datafusion-dml/src/merge_parser.rs` | 494 | 23 | `crates/runtime/src/datafusion/planner/merge.rs` |
| `crates/runtime/src/tools/builtin/sql.rs` | 268 | 3 | `crates/runtime-tools/src/builtin/sql.rs` |
| `crates/runtime/src/rate_limits.rs` | 46 | 0 | `RateLimits` in `crates/runtime/src/flight/mod.rs` |

All three were stale copies left behind when live code moved elsewhere, and all three had been
edited after they went dead — so the cost is not only the dead lines but the review and upgrade
effort spent maintaining them.

PR #12738 deletes the three files. It deliberately does not add the guard, because a
naive implementation produces false positives that would red the build for everyone.

# Proposed Solution

A test (or a small CI step) that walks every workspace crate's `src/` tree and asserts each
`.rs` file is reachable from the crate root.

# Implementation notes

These are the cases a first attempt gets wrong — each one was hit while sweeping for #12735:

- **`benches/`, `tests/`, `examples/` are not in scope.** Cargo auto-discovers each of those
  files as its own target, so they are reachable without any `mod` declaration. Restricting the
  walk to `src/` is what makes the check meaningful; including them produced ~200 false hits.
- **`#[path = "..."]` attributes** redirect a module to an arbitrary file. `crates/cayenne`
  uses these to share bench helpers, so a check that only looks for `mod <basename>;` will
  mis-report them.
- **`mod` declarations behind `#[cfg(...)]`** are still declarations; a file reachable only
  under a non-default feature is not dead.
- **`lib.rs`, `main.rs`, `mod.rs`, `build.rs`** are roots or implicit, and must be skipped.
- **`path = "..."` target overrides in `Cargo.toml`** can name a file directly.

Parsing `mod` declarations with a regex is what makes the above fiddly. Reading the module tree
[...]

PR #12738, opened 2026-08-07
fix: delete three source files that no crate root declares (fixes #12735)

### LINKED ISSUE #12735: crates/datafusion-dml/src/merge_parser.rs is never declared as a module, so 494 lines and 23 tests have never compiled
# Problem

`crates/datafusion-dml/src/merge_parser.rs` is **not declared as a module anywhere**, so it has
never been compiled, linted, formatted, or tested.

`crates/datafusion-dml/src/` holds five sources; `lib.rs` declares four:

```rust
pub mod handler;
pub mod helpers;
pub mod node;
pub mod planner;
```

`merge_parser` is absent. There is no `mod merge_parser;` in the crate, and no reference to its
public API (`parse_merge_sql`, `MergeAstInfo`) anywhere in `crates/` or `bin/`.

The file is 494 lines and contains **23 `#[test]` functions**. None of them have ever run.

# How it surfaced

While verifying the formatter debt in #12726 I ran `rustfmt --check` directly over every tracked
`.rs` file, rather than through `cargo fmt --all`. That reported 21 formatting diffs in
`merge_parser.rs` — on a file that is byte-identical on `trunk` and `byoc`, while `trunk`'s lint
gate is green.

That contradiction is the tell. `cargo fmt --all` only visits files reachable from a crate root
through `mod` declarations, so an undeclared file is invisible to the formatter, to `cargo clippy`,
and to `cargo test` alike. The gate is not wrong; the file simply is not part of the crate.

# Why it matters

- **23 tests report nothing.** They neither pass nor fail; they are silently absent from every
  test run. A test that cannot fail provides no coverage while looking like it does.
- MERGE statement parsing is data-correctness-adjacent. The live path is
  `DmlNodeOp::Merge` -> `handler.merge_exec`, so today the dead file is not serving queries — but
  a future change that wires it in would inherit 494 unreviewed, never-compiled lines.
- It is a standing trap: the file has been edited since it went dead. The DataFusion v53.1 upgrade
  swept it in #11118, so an author has already spent effort maintaining code that is not built.

# History

Added 2026-05-11 in #10618. The `mod` declarations in `lib.rs` were last touched at crate creation
(#10334, 2026-04-16) and never grew an entry for it, so the file has been dead since it landed.

# Proposed Solution

Decide which of the two it is, and make the tree say so:

- If MERGE parsing is genuinely handled by `handler`/`node`/`planner`, **delete the file** — and
  its 23 tests with it, since they test nothing that ships.
- If it is intended to be used, **declare it** (`pub mod merge_parser;`), then fix whatever
  compile, clippy, and formatting errors three months of unchecked drift have accumulated, and
  confirm the 23 tests actually pass.

# Scope & Non-Goals

Covers only this file's status. Does not change MERGE behaviour on the live path.

A cheap guard against a recurrence — asserting every `src/**/*.rs` is reachable from a crate root —
is worth considering separately; this issue is about the one file.

# Acceptance Criteria

- [ ] `crates/datafusion-dml/src/merge_parser.rs` is either deleted or declared as a module.
- [ ] If declared: the crate builds, `make lint-rust` passes over it, and its 23 tests run and pass.
- [ ] No source file in `crates/datafusion-dml/src/` is unreachable from the crate root.

grokspice:
## Summary

`crates/datafusion-dml/src/merge_parser.rs` is not declared as a module anywhere, so it has never
been compiled, linted, formatted or tested — 494 lines and 23 `#[test]` functions that neither
pass nor fail. `cargo fmt --all` and `cargo clippy` only visit files reachable from a crate root
through a `mod` declaration, so the gate was never wrong; the file simply was not part of the
crate.

Deciding between the issue's two options needed the live MERGE path. It is
`planner/mod.rs` -> `merge::plan_distributed_merge`, and that function is a **strict superset** of
the dead one: same `OUTPUT`-clause rejection, same `extract_table_ref`, same qualifier vectors,
same `parse_on_keys` and empty-key check, same single-`WHEN`-clause check, same assignment
extraction — plus the Cayenne target/source validation `merge_parser.rs` never had. The dead
file's aspirational doc comment names two callers (`cayenne::ddl::plan_local_merge`,
`runtime::datafusion::planner::merge::plan_distributed_merge`); neither ever called it, and
`plan_local_merge` does not exist. So this is the delete branch of the issue.

Sweeping the rest of the workspace for the same defect found **two more** stale copies, both
left behind when live code moved and both likewise never compiled:

| File | Lines | Tests never run | Superseded by |
| --- | --- | --- | --- |
| `crates/datafusion-dml/src/merge_parser.rs` | 494 | 23 | `crates/runtime/src/datafusion/planner/merge.rs` |
| `crates/runtime/src/tools/builtin/sql.rs` | 268 | 3 | `crates/runtime-tools/src/builtin/sql.rs` |
| `crates/runtime/src/rate_limits.rs` | 46 | 0 | `RateLimits` in `crates/runtime/src/flight/mod.rs` |

Each is a pre-move snapshot, not a variant worth keeping:

- `tools/builtin/sql.rs` still reaches for `crate::http::v1::current_principal_requires_read_only`
  and `runtime_datafusion::query_engine`, both of which moved during the `runtime-tools`
  extraction. The live module is the one `tools/builtin/catalog.rs` imports, and the dead copy's
  three tests already run inline in `tools/builtin/mod.rs`. The read-only gate it contained is
  intact in the live module, which applies the same
  `QueryRequest::new(&query).read_only(read_only)` — deleting the dead copy removes no control.
- `rate_limits.rs` is a subset of the live `RateLimits`, missing the `flight_write_enabled` and
  `metrics_endpoint_limit` fields the runtime actually uses.

Confirmed before deleting that no `mod` declaration, `#[path]` attribute, or `Cargo.toml` target
override names any of the three.

## No coverage is lost

Each of `merge_parser.rs`'s 23 tests was checked against the live merge planner's test module:

- **22 have a live counterpart that runs today** — the ON-clause set maps onto
  `test_parse_on_keys_*` and `test_flatten_and_conjuncts_deeply_nested`, the table-reference pair
  onto `test_extract_table_ref_*`, the assignment set onto `test_assignment_*`, the WHEN-clause
  pair onto `test_reject_not_matched` / `test_reject_matched_delete`, and
  `assignment_expression_preserved` onto `test_validate_matched_update_set`, which uses identical
  SQL and asserts both the compound right-hand side `s.value + 1` and its position.
- **1 asserted nothing** — `rejects_output_clause` accepted a parse error and a plan error alike
  and discarded both, so there is no intent to preserve. The live planner does reject `OUTPUT`.

This PR therefore adds no tests: every behaviour the dead file described is already covered by
tests that actually execute. An earlier revision of this PR added a test for the assignment value
expression; Copilot correctly identified it as a duplicate of `test_validate_matched_update_set`
and it was removed.

## Behaviour

No change. None of the deleted code was ever built, so nothing could have called it. MERGE
behaviour on the live path is untouched.

## Follow-up

Nothing in the build or lint gate catches this class of defect, which is how three instances
accumulated. Filed as #12737 rather than fixed here — a naive reachability check false-positives
on `benches/`/`tests/`/`examples/` (Cargo auto-discovers those as targets), on `#[path]`
redirection, and on `#[cfg]`-gated `mod` declarations, and a guard that reds the build for
everyone is worse than the defect it catches. That issue records each trap found while sweeping.

## Test plan

- `make lint`
- `make test`

Verified locally:

- `cargo check -p datafusion-dml` passes with `merge_parser.rs` gone — as expected, since it was
  never part of the crate.
- `cargo fmt --all` produces no diff.
- Confirmed by grep that no `mod` declaration, `#[path]` attribute, or `Cargo.toml` target
  override names any of the three deleted files.

The `runtime` crate does not build in this environment — `openssl-sys` builds vendored OpenSSL,
whose `./Configure` needs a native perl, and the only perl available lacks
`Locale::Maketext::Simple`. CI covers it instead: `Build Test Binary`, `Integration Tests`
(parts 1-3), `ADBC Integration Tests` and `E2E Test CI` all passed on this branch, which compiles
and runs the `runtime` test target.

## Checklist

- [ ] `make lint` passes
- [ ] `make test` passes
- [x] No behaviour change on the live MERGE path
- [x] Every deleted test accounted for against a live counterpart
- [x] Deferred work filed as an issue (#12737) rather than left implicit

Fixes #12735


grokspice:
@copilot review

grokspice:
Correct, and thanks — I missed it. `test_validate_matched_update_set` uses the identical SQL and asserts `assignments[1].0 == "value"` and `assignments[1].1 == "s.value + 1"`, so the compound right-hand side and its position were already pinned. My scan looked at the `--- Assignment validation tests ---` block and did not reach that test under `--- WHEN clause tests ---`.

Removed in e9ccc94, so the PR is now a pure deletion. I also did not strengthen the existing assertion: this PR's purpose is removing never-compiled files, and widening an unrelated test would be scope creep on a deletion.

This also improves the deletion's justification, which the PR body now records: all 23 tests in the dead `merge_parser.rs` are accounted for — 22 have a live counterpart that runs today, and the 23rd (`rejects_output_clause`) asserted nothing, accepting a parse error and a plan error alike.

lukekim (approved):

Issue #12741, opened 2026-08-07
bug(ci): a sign-off that runs out its step budget posts a code-failure verdict about a commit nothing judged


claudespice:
### What happened

A remote sign-off that runs out its step budget posts `signoff=failure` — a code-failure verdict — on a commit that nothing judged. Because Attestation rejects a failed sign-off on the head commit (#12362), the PR goes red claiming its diff broke, and a red `signoff` status never self-clears, so the commit is permanently disqualified until someone notices and re-dispatches.

This is the defect #12424 describes, reached through the **timeout** path rather than the cancellation path — so the correction added for cancelled runs does not fire.

### Evidence

Run `31164822954` (`Remote Sign-off (PR #12589)`, head `f72ef2cac`):

| step | outcome |
|---|---|
| 15 `Sign off` | `failure` — `The action 'Sign off' has timed out after 353 minutes` |
| 16 `Correct the sign-off status when this run is cancelled` | `skipped` (correctly — this was not a cancellation) |
| 17 `Resolve an incomplete sign-off` | `success`, and it posted `signoff=failure` |

The status on the head commit is `signoff=failure`, created at `15:07:49Z` — 14s after the step timed out.

**Nothing had failed when the axe fell.** The job reached `Starting 9947 tests across 188 binaries` at `14:56:43Z` and completed 2238 of them with **zero failures**, at ~133 tests/min — roughly 58 minutes short of finishing.

The budget went almost entirely to work that precedes the tests:

- `make lint-rust`: ~3h17m
- `dev` profile build: `Finished dev profile ... in 127m 12s`
- tests: 10.8 minutes before the timeout

### Why the status is wrong

The two handlers write opposite states, and `timeout-minutes: 353` routes budget expiry to the one that writes `failure`:

- `correct-cancelled` — "rewrite `signoff=failure`/`error` to pending, and leave any other state (notably success) alone"
- `clear-pending` — "Replace a left-behind `signoff` pending status on HEAD with a **failure** (no-op for any other state)"

`signoff.yml` explains the intent behind `pending` for the cancelled case: *pr.yml rejects `failure` and `error` but treats `pending` as "not signed off yet" … so Attestation stays red (it must) and points at the documented remedy of re-running sign-off instead of at a phantom defect in the diff.*

[...]

Issue #12472, opened 2026-08-04
Remote Sign-off concurrency is keyed on the dispatch input, so a branch= and a pr_number= dispatch for the same branch both run


grokspice:
## Describe the bug

`signoff.yml` keys its concurrency group on the raw dispatch input:

```yaml
concurrency:
  group: signoff-${{ inputs.pr_number || inputs.branch }}
  cancel-in-progress: true
```

The workflow accepts either input for the same target — the header documents both
`gh workflow run signoff.yml -f branch=<branch>` and `gh workflow run signoff.yml -f pr_number=<N>` —
and resolves them to one branch in the `Resolve branch/PR input` step. But the concurrency key is
evaluated *before* that resolution, so a `branch=` dispatch and a `pr_number=` dispatch for the
same branch land in **different groups** (`signoff-fix/my-branch` vs `signoff-1234`) and neither
cancels the other.

Two consequences:

1. **Two 1–4h runs occupy two slots of the shared self-hosted macOS pool for the same commit.**
   `cancel-in-progress` exists precisely to stop that, and it silently does not apply.
2. **Both runs post the `signoff` commit status on the same SHA**, so the later write wins
   regardless of which run is authoritative. Combined with the post-run cleanup handlers, a
   sibling run that is cancelled or overruns can rewrite the status a still-valid run already
   posted — the same failure shape as #12428, arriving from a different direction.

## To Reproduce

```bash
gh workflow run signoff.yml -f branch=<branch-of-PR-N>
gh workflow run signoff.yml -f pr_number=<N>
gh run list --workflow signoff.yml --limit 5
```

Both runs stay `in_progress`.

Observed 2026-08-04 on this repo, two independent pairs alive simultaneously:

| branch | dispatched by `branch=` | dispatched by `pr_number=` |
| --- | --- | --- |
| `fix/12409-es-bulk-body-redaction` (PR #12456) | run 30899135437, 10:05Z | run 30899699418, 10:12Z |
| `fix/12462-dotted-index-key-requalification` (PR #12464) | run 30896470096, 09:28Z | run 30899703798, 10:12Z |

Both pairs were still running ~100 minutes later, i.e. the 10:12Z dispatches did not cancel the
earlier ones. Each pair's two runs also both target the same head commit (`98eedf0d2`,
`29c93a60e`), so they duplicate identical work.

## Expected behavior

One in-flight sign-off per target commit, whichever input form was used to dispatch it, and one
[...]

PR #12752, opened 2026-08-07
fix(runtime): pin the catalog-acceleration contract in both feature configurations (fixes #12743)

### LINKED ISSUE #12743: runtime test_try_from_maps_acceleration needs the postgres feature it never declares, so every scoped nextest run is red
# Problem

`runtime`'s `component::catalog::tests::test_try_from_maps_acceleration` requires the `postgres`
feature but does not declare it, so it **fails on any scoped run that does not happen to enable
`postgres`** — including the one `make` produces by default for a targeted package list.

```
thread '...test_try_from_maps_acceleration' panicked at crates/runtime/src/component/catalog.rs:354:14:
should build: ComponentError { source: CatalogAccelerationUnsupportedProvider { name: "my_pg", provider: "pg" } }
```

The production code is explicitly feature-conditional, and says so:

```rust
// Feature-gated: without the `postgres` feature the `pg` catalog
// connector isn't compiled in at all, so no provider supports catalog
// acceleration and any `acceleration` config is rejected.
#[cfg(feature = "postgres")]
let acceleration_supported = provider == crate::catalogconnector::postgres::PREFIX;
#[cfg(not(feature = "postgres"))]
let acceleration_supported = false;
```

The test that exercises it is a bare `#[test]` with no matching `#[cfg(feature = "postgres")]`, so
it asserts a `pg` catalog acceleration builds successfully in a build where that is defined to fail.

# Why it is invisible in CI

`make nextest` runs `cargo nextest run --all`, where feature unification gives `runtime` whatever
the workspace's binaries request — `postgres` among them — so the test passes and the gap never
surfaces.

A **scoped** run resolves differently. In the `Makefile`:

```make
ifneq ($(strip $(FEATURES)),)
_FEATURES_FLAGS := --features $(FEATURES)
else ifneq ($(strip $(PACKAGES)),)
_FEATURES_FLAGS :=
```

With `PACKAGES` set and `FEATURES` unset, `_FEATURES_FLAGS` is **empty**, and `runtime` declares no
`[features] default`, so the crate compiles with zero features.

# Evidence

Both runs on the same commit, same tree:

| command | result |
|---|---|
| `make nextest-packages PACKAGES="spicepod runtime"` | 1842 run, **1 failed** — this test, after all 6 retries |
| `make nextest-packages PACKAGES="runtime" FEATURES="postgres"` | 1640 run, **1640 passed**, this test `PASS` |

# Why it matters

`scripts/signoff` runs a targeted `nextest-packages` pre-check on the crates a branch touched. Any
branch touching `runtime` therefore gets a red pre-check it did not cause, and the author has to
prove the failure is not theirs before continuing — the same tax #12726 describes on the `byoc`
branch, from a different direction. A test that cannot pass in a configuration it is run in is
worse than no test: it trains readers to discount a red.

Found while verifying #12726 (PR #12739); it is unrelated to that change, and the file is
byte-identical on `trunk`.

# Proposed Solution

Either is defensible; the first is smaller:

- Gate the test to match the code it tests: `#[cfg(feature = "postgres")]` on
  `test_try_from_maps_acceleration`, so it runs exactly where its premise holds.
- Or assert the *conditional* contract — accepted with `postgres`, rejected with
  `CatalogAccelerationUnsupportedProvider` without it — which keeps coverage in both configurations
  and pins the documented behaviour of the `#[cfg(not(feature = "postgres"))]` arm, currently
  untested.

Worth checking as part of this whether other `runtime` tests carry the same undeclared dependency;
this one was found by running the suite, not by an audit.

# Acceptance Criteria

- [ ] `make nextest-packages PACKAGES="runtime"` passes with no `FEATURES` override.
- [ ] Coverage of the `pg` catalog-acceleration mapping is not lost.
- [ ] The `#[cfg(not(feature = "postgres"))]` rejection arm is covered, or explicitly out of scope.

claudespice:
## Summary

`runtime`'s `component::catalog::tests::test_try_from_maps_acceleration` asserted that a `pg`
catalog acceleration builds successfully, but the code it exercises is explicitly conditional:

```rust
#[cfg(feature = "postgres")]
let acceleration_supported = provider == crate::catalogconnector::postgres::PREFIX;
#[cfg(not(feature = "postgres"))]
let acceleration_supported = false;
```

Without `postgres` there is no `pg` catalog connector compiled in, so the builder is *defined* to
reject the config — and the test, a bare `#[test]` with no matching `cfg`, asserted the opposite.

The gap is invisible to `make nextest`, which runs `--all`: `spiced`'s default features include
`postgres`, and feature unification turns on `runtime/postgres` for the whole workspace build, so
the test's premise happens to hold. A **scoped** run resolves differently — with `PACKAGES` set and
`FEATURES` unset the `Makefile` leaves `_FEATURES_FLAGS` empty, and `runtime` declares no
`[features] default`, so the crate compiles with zero features and the test fails.

That matters because `scripts/signoff` runs a targeted `nextest-packages` pre-check on the crates a
branch touched. Every branch touching `runtime` got a red pre-check it did not cause, and the author
had to disprove it before continuing. A test that cannot pass in a configuration it is run in is
worse than no test: it trains readers to discount a red.

Fixes #12743

## Changes

- Gate `test_try_from_maps_acceleration` on `#[cfg(feature = "postgres")]`, so it runs exactly where
  its premise holds. It still runs under `make nextest`/CI via the unification described above, so
  no coverage of the mapping is lost.
- Add `test_try_from_rejects_acceleration_without_postgres_feature`, covering the
  `#[cfg(not(feature = "postgres"))]` arm — previously untested in any configuration. It asserts the
  specific `CatalogAccelerationUnsupportedProvider` error rather than just `is_err()`, so a future
  change that rejects for the wrong reason still fails.
- Extract the shared `cayenne_file_acceleration()` fixture the three acceleration tests now build
  from, instead of a third verbatim copy.

Taken together the two arms pin the conditional contract in both configurations, rather than
deleting coverage to make the suite pass.

## Test plan

- `make lint`
- `make test`

Both feature configurations were run locally, since each one compiles only half of this diff:

| configuration | result |
|---|---|
| `cargo nextest run -p runtime --lib` (zero features — what `nextest-packages` produces) | **1608 passed, 1 skipped, 0 failed** (was 1 failed) |
| `cargo test -p runtime --features postgres --lib component::catalog` | **5 passed**, incl. `test_try_from_maps_acceleration` |

The zero-feature run doubles as the class sweep the issue asked for — empirical rather than by
grep, since a grep cannot tell which tests *depend* on a feature. No other `runtime` test carries an
undeclared feature dependency.

One caveat for anyone re-running this: a full-suite `cargo test -p runtime --lib` (as opposed to
`nextest`) additionally reports `accelerated_table::refresh::tests::test_refresh_status_change_to_ready`
and `init::dataset::tests::a_dataset_connector_failure_counts_one_load_error` as failing. Those are
artifacts of the single-process harness — both read process-global metrics state that other tests in
the same binary mutate — not feature bugs, and both pass under `nextest`, which CI uses. They are
pre-existing and untouched here.

## Checklist

- [x] Regression verified: the zero-feature suite was red before this change and is green after
- [x] Coverage of the `pg` catalog-acceleration mapping preserved (still runs under `make nextest`)
- [x] `#[cfg(not(feature = "postgres"))]` rejection arm covered
- [ ] `make lint` passes
- [ ] `make test` passes

## Note for reviewers

#12641 (also mine) adds three tests to this same `mod tests`, immediately above
`test_try_from_maps_acceleration`. The two changes are disjoint at function granularity — that PR
inserts new functions, this one modifies `test_try_from_maps_acceleration` and adds a sibling — so
whichever lands second resolves by keeping both sets.


claudespice:
@copilot review

claudespice:
**The red `signoff` on this PR is not a verdict about this diff — please don't read it as one.**

Two sign-off runs have been consumed here, neither of which judged the code:

1. [`31203505489`](https://github.com/spiceai/spiceai/actions/runs/31203505489) — ran 4h15m, then **cancelled** 34s after a second dispatch for this same PR landed (`concurrency` pairs a `branch=` and a `pr_number=` dispatch once `resolve` pins the same commit, per #12472). The cancellation posted `failure` whose duration is just this job's own wall time.
2. [`31222886487`](https://github.com/spiceai/spiceai/actions/runs/31222886487) — hit the 353-minute step budget wedged inside the workspace `cargo clippy --lib --bins` pass, with **1h56m of total silence** before the axe and `Finished dev profile` never printed. Its own `Resolve an incomplete sign-off` step says *"Remote sign-off ended without posting a verdict — sign off again"* — see #12741 for why that lands as `failure` instead of `pending`, and #12781 for the wedge (I've added this run's evidence there; it shows the wedge is **not** confined to the `--tests` pass).

I'm **not** re-dispatching a third time — that would be the identical approach, since the remote workflow already runs with `run_targeted_prechecks` off, leaving no cheaper lever. This needs #12781 resolved, or a maintainer's call.

**What was actually verified, locally, on this exact commit** — both feature configurations, because the `cfg` arms mean neither build compiles the whole diff:

| configuration | result |
|---|---|
| `cargo nextest run -p runtime --lib` (zero features — what the scoped gate produces) | **1608 passed, 0 failed** (was 1 failed before this change) |
| `cargo test -p runtime --features postgres --lib component::catalog` | **5 passed**, incl. `test_try_from_maps_acceleration` |

`rustfmt --check` clean. The zero-feature run is also the class sweep #12743 asked for: no other `runtime` test carries an undeclared feature dependency.

phillipleblanc (approved):

Issue #12761, opened 2026-08-07
Rotate the expired Oracle Cloud test credential to restore oracle_test_cloud_mtls coverage


claudespice:
## Summary

`oracle::oracle_test_cloud_mtls` covers the Oracle Cloud mTLS connection path against a real
Oracle Cloud test account. That account's password is expired (`ORA-28001`), so the test cannot
exercise the connector at all.

#12758 fixed the *blast radius* — an unusable credential now skips this one test with a reported
reason instead of failing `Integration Tests (part 2)` and evicting every merge-queue candidate.
It deliberately did not, and could not, restore the coverage: that needs a credential rotation.

This issue tracks restoring the coverage, so it does not get lost when #12758 closes.

## What is needed

1. Rotate the password on the Oracle Cloud test account.
2. Update the corresponding CI secret (`ORACLE_CLOUD_PASSWORD`).
3. Confirm `oracle_test_cloud_mtls` runs to completion again rather than taking the skip path.

## How to tell it is still skipping

The skip prints, in the test's captured output:

```
Skipping oracle_test_cloud_mtls: the ORACLE_CLOUD_* test credential is unusable
(the password has expired). Rotate the Oracle Cloud test account's password and update
the CI secret to restore this coverage.
```

Note that a passing test's output is captured by nextest, so this line is not visible in a green
CI run — the skip is intentionally not a failure, which is what keeps an expiry from halting the
repo, but it does mean this coverage is silently absent until the credential is rotated.

## Related

- #12758 — the merge-queue blockage this credential caused, and the skip that bounds it.
- #12473 — the same shape for the benchmark runners: an external dependency that cannot be
  fixed from this repository.

phillipleblanc:
This was rotated

PR #12765, opened 2026-08-07
fix(tests): skip the Oracle cloud test when its credential is unusable (fixes #12758)

### LINKED ISSUE #12758: CI: an expired Oracle test password fails every merge-queue candidate, so nothing has merged to trunk in 19 hours
## Summary

`oracle::oracle_test_cloud_mtls` fails on every attempt with an expired database password, so `Integration Tests (part 2)` fails on every merge-queue candidate and each candidate is evicted. **`trunk` has not advanced in over 19 hours** while 19 approved, fully green PRs sit in the queue.

This is not a defect in any PR under test. No code change can fix it — the test account's password needs rotating and the CI secret updating.

## Evidence

From the merge-queue run for the queue head (`Integration Tests (part 2)`, run `31202465435`):

```
WARN runtime::init::dataset: Error initializing dataset nation. Failed to initialize
data connector: Failed to connect to the Oracle Server. ... Failed to establish
connection: OCI Error: ORA-28001: the password has expired
test oracle::oracle_test_cloud_mtls ... FAILED
Error: Timed out waiting for datasets to load
```

It fails identically on retry (`TRY 1 FAIL`, `TRY 2 FAIL`), so nextest's retry does not absorb it — the job fails deterministically.

## Impact

Every PR that reaches the front of the merge queue is dequeued on this one test.

- `trunk` tip is unchanged since 2026-08-07T00:37:54Z.
- The merge queue holds 19 entries, all approved with green required checks.
- Observed on three separate queue formations today.

It also compounds two existing problems. #12561 — the fix for the fleet-wide sign-off failure in #12436 — has now been enqueued and evicted three times without merging (2026-08-05 17:27, 2026-08-07 03:59, 2026-08-07 16:32), so #12436 keeps failing sign-offs across unrelated branches. And the PRs that would bound a wedged queue job (#12717/#12719, #12718/#12721) are themselves stuck behind it.

## What is needed

1. Rotate the password on the Oracle Cloud test account and update the corresponding CI secret.
2. Consider whether a credentialed external-service test belongs on a merge-queue-gating job at all: an expired credential on a third-party account is an availability dependency for merging anything. Skipping the test when its credential is unusable — rather than failing the gate — would keep an expiry from halting the repo. Related: #12473.

## Also seen in the same run

`iceberg::write::iceberg_insert_into_existing_table` failed `TRY 1` with `CatalogCommitConflicts` ("The client may retry"). That one looks like a genuine concurrency flake and is not the blocker here, but it is worth its own look if it recurs.

claudespice:
## Summary

`oracle::oracle_test_cloud_mtls` connects to a real Oracle Cloud test account. That account's
password expired, so every run failed, `Integration Tests (part 2)` failed on every merge-queue
candidate, and each candidate was evicted. `trunk` did not advance for over 19 hours while 19
approved, fully green PRs sat in the queue.

Two things made an expired password able to halt the repository:

1. **The failure was undiagnosable.** The test raced `load_components()` against a one-minute
   sleep and reported `Timed out waiting for datasets to load`. The actual cause —
   `ORA-28001: the password has expired` — only ever appeared in a `WARN` line from
   `runtime::init::dataset`, so the test's own output never named it.
2. **An unusable credential failed a merge-queue-gating job.** A third-party account's password
   expiry became an availability dependency for merging anything.

The dataset already records the connector's error text in its component status
(`init/dataset.rs` stores `ComponentStatus::error_with_message(err.to_string())` before it
returns), so the test can read why readiness never arrived rather than only that it waited.

## Changes

- `tests/utils`: add `dataset_load_errors`, which returns the error each dataset recorded when it
  failed to load. Sorted, so the reported diagnosis reads the same way twice.
- `tests/oracle`: replace the blind timeout race with a bounded wait that, when the runtime never
  becomes ready, reads those recorded errors and decides from them:
  - an error naming a credential that cannot be used at all — `ORA-01017` (invalid credentials),
    `ORA-28000` (account locked), `ORA-28001` (password expired) — skips the test with a message
    naming the condition and the remediation;
  - **every other failure still fails**, now reporting the dataset error instead of a bare
    timeout.

The list is deliberately short. `ORA-28002` ("the password will expire within N days") accompanies
a logon that *succeeded*, so it is not on it — treating it as unusable would skip a test that can
still run. Matching also requires the code not be followed by another digit, since the recorded
status text is free-form and `ORA-28001` occurs inside `ORA-280015`. Both cases are tested.

The skip is deliberately narrow: it requires the server to have named a credential condition. If
no dataset recorded an error, the test fails rather than skips, so a connector regression cannot
be filed away as somebody else's expired password.

## What this does not do

It does not restore the coverage. The password still needs rotating and the CI secret updating —
a credential operation, not a code change. #12761 tracks that separately so it is not lost when
this closes. Because a passing test's output is captured, the skip is not visible in a green CI
run; #12761 is the tracker that keeps it visible.

Scope note: sibling credentialed suites (`github`, `sharepoint`, `cluster`, `models`) already skip
when a secret is *absent*, but none classify a secret that is present and *rejected*. Generalising
that needs a per-provider classifier (Oracle `ORA-` codes, HTTP 401/403, AWS error codes) and is
not folded in here.

## Test plan

- `make lint`
- `make nextest`
- New unit tests over the classifier, run as part of the Oracle integration test binary:
  - `expired_password_is_an_unusable_credential`
  - `a_locked_account_and_bad_credentials_are_unusable_credentials`
  - `a_password_nearing_expiry_is_not_an_unusable_credential` (`ORA-28002` accompanies a
    successful logon and must not skip the test)
  - `a_longer_code_sharing_a_prefix_is_not_an_unusable_credential`
  - `an_unrelated_failure_is_not_a_credential_problem` (a TNS error, a bare timeout, and an empty
    message all still fail)

## Checklist

- [x] Tests added for the new behaviour, including the direction that must keep failing
- [x] No production code paths changed — the diff is entirely within `crates/runtime/tests`
- [x] Follow-up tracker filed for the credential rotation (#12761)

Fixes #12758

### DIFF
--- a/crates/runtime/tests/oracle/mod.rs
+++ b/crates/runtime/tests/oracle/mod.rs
@@ -24,7 +24,8 @@ use crate::oracle::common::{
     make_oracle_cloud_dataset, make_oracle_dataset, start_oracle_docker_container,
 };
 use crate::utils::{
-    register_test_connectors, runtime_ready_check, test_request_context, verify_env_secret_exists,
+    dataset_load_errors, register_test_connectors, runtime_ready_check,
+    runtime_ready_check_with_timeout_err, test_request_context, verify_env_secret_exists,
 };
 
 pub mod common;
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

phillipleblanc (changes requested):
This is a wrong fix, it would silently skip oracle testing in the future without anyone knowing.

claudespice:
### Stalled on a rejected approach, not on CI — and the tracking issue it deferred to was closed a minute earlier

Every check on `342a94851` is green (`Attestation` ✅, `signoff` ✅, `enforce-pull-with-spice` ✅) and there are zero unresolved review threads. The only thing holding this PR is @phillipleblanc's `CHANGES_REQUESTED` from 2026-08-09T15:01:44Z — *"This is a wrong fix, it would silently skip oracle testing in the future without anyone knowing"* — which has had no response for four days, and no commit has landed since 2026-08-07.

One data point that seems relevant to the decision rather than to the code: **#12761 — the issue this PR's description defers the lost coverage to — was closed at 2026-08-09T15:00:32Z, 72 seconds before the review was submitted.** So the "coverage is tracked elsewhere" argument the PR rests on no longer has an open tracker behind it, while #12758 (the CI symptom) is still open.

Taking the objection at face value, the skip cannot land as written: a silent skip keyed on `ORA-01017`/`ORA-28000`/`ORA-28001` means Oracle Cloud coverage can disappear indefinitely with a green board and nobody notified. Two ways out, both design calls rather than merge-blockers:

1. **Make the absence loud instead of silent** — keep the classifier, but have the skipped case surface somewhere a person sees it, so a credential that stays unusable is visibly costing coverage rather than quietly costing it.
2. **Close this PR and fix #12758 at the credential** — which is what the review reads like it is asking for, and what #12761's closure may already have done.

Not guessing between those by pushing code: option 1 changes what the PR is for, and option 2 means closing someone else's PR. Flagging it for a human instead. The classifier itself (including the `ORA-280015` prefix-collision guard and the deliberate `ORA-28002` exclusion) is sound work and worth keeping whichever way this goes.

claudespice:
Closing this — the problem it worked around is gone, and @phillipleblanc is right that the mechanism was the wrong shape.

**The credential was rotated.** #12761 was closed on 2026-08-09 with "This was rotated", and `oracle::oracle_test_cloud_mtls` now runs to completion against the real account: it `PASS`ed in the merge-group integration run [`31755685233`](https://github.com/spiceai/spiceai/actions/runs/31755685233), on a `trunk` that does **not** carry this PR — so that is the unmodified test genuinely connecting, not a skip path. The merge queue has been draining normally since (trunk advanced eight times on 2026-08-13).

**And the review comment stands on its own.** A classifier that turns a specific connection error into a green skip has no expiry: once the next credential lapses, the coverage disappears silently and the only signal is a line in captured test output that nothing reads. That is a worse failure mode than the loud one it replaced, because the loud one is what got the credential rotated within two days.

If a future lapse needs bounding, the right shape is a check that *fails* on a stale credential in a place that is not a merge-queue gate (a scheduled job, or a pre-flight secret-expiry probe), so the alert survives without buying it with lost coverage.

No code from this branch is needed on `trunk`. #12758 is closed as resolved by the rotation.

Issue #12760, opened 2026-08-07
byoc: release() builds a rustls identity on a native-tls client, so every identity-bearing release fails to build its HTTP client


claudespice:
## Summary

On the `byoc` branch, `release()` builds a **rustls** client identity but never pins the rustls backend, so `reqwest` resolves the builder to **native-tls** and rejects the identity. Every call that presents an instance identity fails to construct its HTTP client:

```
ClientBuild { source: reqwest::Error { kind: Builder, source: "incompatible TLS identity type" } }
```

This is the production `release()` path, not test-only scaffolding — `release_posts_the_certificate_and_a_pop_signature_the_cloud_can_verify` fails because it calls the real function.

## Cause

`crates/runtime-cloud-connect/src/release.rs:206` constructs the identity with the rustls-only constructor:

```rust
let client_identity = reqwest::Identity::from_pem(client_pem.as_bytes())...;

let mut builder = reqwest::Client::builder()
    .timeout(Duration::from_secs(30))
    .connect_timeout(Duration::from_secs(10))
    .identity(client_identity);   // <-- no .use_rustls_tls()
```

The workspace declares `reqwest = { version = "0.13", features = ["json", "query", "rustls", ...] }` without `default-features = false`, so `default-tls` (native-tls) stays enabled alongside rustls. With both backends compiled in, a builder that never calls `.use_rustls_tls()` defaults to native-tls, and `Identity::from_pem` produced a rustls identity — hence the mismatch.

Because it depends on backend unification rather than on the code under test, it is invisible to a narrowly-scoped build and deterministic under the full release feature set the sign-off gate uses (observed `TRY 6 FAIL`, all six attempts identical).

## The fix, and the precedent for it

`release.rs` is the only place in the tree that calls `.identity()` without first pinning rustls. Every other call site already does:

- `crates/runtime/src/dataconnector/https.rs:879` — `builder = builder.use_rustls_tls().identity(identity);`
- `crates/llms/src/provider.rs:125`
- `crates/runtime/src/catalogconnector/spice_cloud.rs:129`
- `crates/runtime/tests/mtls_connector/mod.rs:205`
- `crates/runtime/tests/mtls_public/mod.rs:189`

so the change is to match them:

```rust
let mut builder = reqwest::Client::builder()
    .timeout(Duration::from_secs(30))
    .connect_timeout(Duration::from_secs(10))
    .use_rustls_tls()
    .identity(client_identity);
```

## Impact

[...]

PR #12739, opened 2026-08-07
fix(lint): clear the byoc lint and formatter debt blocking make lint-rust (fixes #12726)

claudespice:
## Summary

`make lint-rust` could not pass on `byoc`, for reasons unrelated to any change made on top of it.
`pr.yml` triggers only on `pull_request.branches: [trunk, release-*, release/*, feature-*]`, so a PR
into `byoc` runs essentially no CI and nothing ever reported the breakage. Each author on the branch
discovered it by running the gate locally, then had to establish the failures were not their own.
Each also correctly declined to fold a drive-by fix of unrelated crates into an unrelated PR — which
is exactly why it stayed broken.

Two independent failures, both introduced on the branch:

- **`clippy::err_expect`** — three `.err().expect(..)` calls in `crates/spicepod/src/lib.rs`, which
  arrived with #12640. The lint is warn-by-default, but `make lint-rust` passes `-Dwarnings`, which
  promotes it to `error: could not compile spicepod (lib test)`.
- **`cargo fmt --all --check`** — `crates/runtime/src/metrics_reader.rs` and
  `crates/spicepod/src/lib.rs`.

Both are confined to test modules (`mod tests`, `mod version_tests`). No shipping code path changes.

Fixes #12726

## Changes

**`expect_err(..)` rather than the `unwrap_err()` the issue proposed.** Three reasons:

1. It is the replacement `clippy::err_expect` itself suggests — the diagnostic reads
   `help: try: expect_err`.
2. It keeps all three descriptive messages (`"an empty directory has no spicepod.yaml"`, and the
   two others). `unwrap_err()` would discard them, and those messages are the only thing that says
   what the assertion is for.
3. `CLAUDE.md` bars a bare `unwrap` in tests in favour of `.expect("descriptive message")`, so
   `unwrap_err()` would trade one lint complaint for a house-style violation.

One semantic difference worth naming: `expect_err` requires `T: Debug`, where `.err().expect()` does
not, because it formats the `Ok` value into the panic message. `Spicepod` satisfies it (verified —
the crate compiles clean), and that text is only ever produced when the test *fails*, where naming
the unexpected value is strictly better diagnostics.

The two formatting hunks are rustfmt's own output, applied unedited.

## Scope

I checked whether the issue's scope was complete rather than assuming it. It is:

- **`err_expect`** — a repo-wide sweep for the pattern (multi-line aware; `.err()` and `.expect(`
  sit on separate lines of a method chain, so a single-line grep silently finds nothing) reports
[...]

PR #12769, opened 2026-08-07
fix(cloud-connect): pin rustls where the release call presents its identity

grokspice:
## The bug

`release` builds its mTLS client from `reqwest::Identity::from_pem`, which yields a **rustls** identity. The workspace compiles both TLS backends in, so a builder left on the default resolves to native-tls and rejects the identity at build time:

```
ClientBuild { source: reqwest::Error { kind: Builder, source: "incompatible TLS identity type" } }
```

The client fails to build before a socket is opened, so `spice connect remove` could never post its release on this branch.

## Why it matters beyond the feature

`release_posts_the_certificate_and_a_pop_signature_the_cloud_can_verify` already covers this path and fails **deterministically** — 6/6 nextest attempts, all at `release.rs:455` — so it fails the full test suite for **every PR based on `byoc`**. #12739 (byoc lint debt, a 2-file diff touching only `metrics_reader.rs` and `spicepod/src/lib.rs`) has been red on sign-off entirely because of this, having nothing to do with its own changes.

The test is `byoc`-only — it does not exist on `trunk` — which is why `trunk` sign-offs are unaffected.

## The fix

Pin the backend the way every other `.identity()` call site in the workspace does, e.g. [`crates/runtime/src/dataconnector/https.rs:879`](https://github.com/spiceai/spiceai/blob/trunk/crates/runtime/src/dataconnector/https.rs#L879):

```rust
builder = builder.use_rustls_tls().identity(identity);
```

`release.rs` was the only `.identity()` site in the workspace that did not.

**No new test** — the existing test already proves it, and it is the one that has been failing.

Same root cause and same one-line fix as #12764 on `trunk`; this applies it to `byoc`'s own copy of `release.rs`, which diverged after #12629.

## Verification

`use_rustls_tls()` is available here: the workspace `reqwest` dep carries the `rustls` feature and this crate uses `reqwest = { workspace = true }`.

Remote sign-off dispatched (no local Rust toolchain on the authoring host).

### DIFF
--- a/crates/runtime-cloud-connect/src/release.rs
+++ b/crates/runtime-cloud-connect/src/release.rs
@@ -207,9 +207,14 @@ pub async fn release(
             instance_id: identity.identifier.clone(),
         })?;
 
+    // `reqwest::Identity::from_pem` builds a rustls identity, and the workspace
+    // compiles both TLS backends in, so the backend has to be pinned to match:
+    // a builder left on the default resolves to native-tls and rejects the
+    // identity outright. Every other `.identity()` call site pins it the same way.
     let mut builder = reqwest::Client::builder()
         .timeout(Duration::from_secs(30))
         .connect_timeout(Duration::from_secs(10))
+        .use_rustls_tls()
         .identity(client_identity);
     if let Some(ca_pem) = ca_cert_pem {
         for cert in reqwest::Certificate::from_pem_bundle(ca_pem.as_bytes()).context(CaCertSnafu)? {


grokspice:
@copilot review

grokspice:
## Rebased onto #12739 — the two byoc blockers deadlocked each other

The first sign-off here failed in **13 seconds**, and not on this diff ([run 31226639400](https://github.com/spiceai/spiceai/actions/runs/31226639400)):

```
cargo fmt --all -- --check
Diff in crates/runtime/src/metrics_reader.rs:928
Diff in crates/spicepod/src/lib.rs:1453
make: *** [lint-rust] Error 1
```

Those are `byoc`'s pre-existing formatter debt — and they are exactly the two files #12739 fixes. `make lint-rust` runs `cargo fmt --all -- --check` over the whole workspace before it reaches the tests, so **every** PR based on `byoc` fails sign-off at 13s regardless of its own contents.

That left the two byoc blockers each holding the other's gate:

| PR | Fixes | Fails sign-off on |
|---|---|---|
| #12739 | the fmt/lint debt | the rustls identity bug (tests, deterministic 6/6) |
| this one | the rustls identity bug | the fmt debt (13s, before tests run) |

Neither could go green alone. So this is now stacked on `fix/12726-byoc-lint-debt` instead of `byoc` directly — the same pattern as #12713, #12715 and #12744. With both fixes present the fmt gate passes and the release test's deterministic failure is fixed, so the stack can sign off.

The diff against the new base is unchanged — the single `.use_rustls_tls()` line. Sign-off re-dispatched ([run 31227072550](https://github.com/spiceai/spiceai/actions/runs/31227072550)).

**Merge order:** this PR into `fix/12726-byoc-lint-debt`, which then carries both fixes into `byoc` via #12739.

grokspice:
Closing this in favour of #12764, which carries the same fix on a current base.

Both branches add the same one-line repair — `.use_rustls_tls()` on the `release()` client builder, so the rustls identity that `reqwest::Identity::from_pem` produces is not handed to a native-tls backend. #12764 does it against a base from 2026-08-07 and is a pure addition:

```
crates/runtime-cloud-connect/src/release.rs | 101 +++++++++++++++++++++
1 file changed, 101 insertions(+)
```

This branch was cut from a base two days older, and `release.rs` was substantially rewritten on trunk in between. Measured against the current trunk it is:

```
crates/runtime-cloud-connect/src/release.rs | 286 +++++++++++++++++---
1 file changed, 258 insertions(+), 28 deletions(-)
```

Those 28 deletions are merged trunk work, not stale scaffolding. Merging this branch would revert the release endpoint from the mTLS-leaf authorisation model now on trunk back to the proof-of-possession-in-the-body design, and restore the older `Identity` error wording along with it. That is a behaviour change well outside a TLS-backend pin, and it is not what either branch set out to do.

#12764 also carries the issue linkage this one is missing (`fixes #12760`), so it closes the bug on merge.

No work is lost — the fix and its regression test live on #12764. Please track it there.

PR #12783, opened 2026-08-08
fix(search): make chunked embedding and offset columns nullable (#12778)

### LINKED ISSUE #12778: bug(search/cayenne): non-nullable embedding column blocks Cayenne refresh when vectors offload to S3 Vectors
## Summary

A Cayenne accelerator refresh fails to write a Vortex file. The failure happens when an external S3 Vectors index backs a chunked full-text-search embedding column. The declared write schema marks the embedding column non-nullable. The refresh scan fills that column with all nulls. Vortex rejects the batch.

## Error

```
Failed to refresh dataset from the data connector: Execution error: Failed to write Vortex file '.../_staging/.../<id>.vortex': Cannot convert an Arrow array containing 4608 nulls into a non-nullable Vortex array.
```

## Root cause

Two places disagree about the nullability of the derived embedding column.

1. `crates/runtime-search/src/embeddings/table.rs:707-728` declares the derived embedding column non-nullable. The `(Scalar, chunker=Some)` arm builds `{col}_embedding` and `{col}_offset` with the outer list `nullable=false`. Full-text search enables chunking, so a column with `full_text_search` takes this arm. The unchunked arm at lines 730-735 uses `nullable=true`, so a plain scalar embedding avoids the bug.

2. `crates/runtime-datafusion/src/optimizer_rule/avoid_vector_columns_on_index.rs:142-146` replaces the derived embedding column with an all-null literal during the refresh scan. `AvoidDerivedVectorColumnOnIndexRule` removes the embedding column from the scan projection to skip recomputation. It re-adds the column as `Expr::Literal(ScalarValue::try_from(field.data_type())?, None)`. The literal produces one null per row. The rule aligns the column name only. It never relaxes the column nullability in the accelerator write schema.

3. `crates/vortex/src/persistent/sink.rs:232` builds a non-nullable Vortex `DType` from the write schema through `DType::from_arrow(get_writer_schema(&self.config))`. Vortex then rejects the all-null batch.

`crates/search/src/index/mod.rs:218-225` confirms the nulled column is `{col}_embedding`.

The non-nullable declaration is correct only for locally materialized embeddings. An external S3 Vectors index moves the vectors out, so the local column is all-null and the declaration is wrong.

## Impact

A dataset that pairs a chunked full-text-search embedding column with an external S3 Vectors index cannot refresh into a Cayenne accelerator. Partitioning is not required to trigger the failure. The nullability conflict exists for any Cayenne write on that column layout.

## Reproduce

```yaml
datasets:
  - from: postgres:example_table
    name: example
    params:
      pg_host: <host>
      pg_db: <db>
      pg_user: <user>
      pg_pass: ${secrets:PG_PASS}
    acceleration:
      enabled: true
      engine: cayenne
      mode: file
      primary_key: id
      refresh_mode: append
    vectors:
      enabled: true
      engine: s3_vectors
      params:
        s3_vectors_bucket: <bucket>
        s3_vectors_aws_region: <region>
    columns:
      - name: content
        embeddings:
          - from: openai_embeddings
            row_id:
              - id
        full_text_search:
          enabled: true
          row_id:
            - id

embeddings:
  - from: openai
    name: openai_embeddings
    params:
      openai_api_key: ${secrets:OPENAI_API_KEY}
```

## Proposed fix

Declare the embedding column nullable when an external vector index owns its vectors. Fix the source at `crates/runtime-search/src/embeddings/table.rs`, or reconcile nullability where the sink builds the write schema. Keep the non-nullable declaration for locally materialized embeddings.

Jeadie:
## Problem

A Cayenne accelerator refresh failed to write a Vortex file. The failure happened when an external S3 Vectors index backed a chunked full-text-search embedding column. The refresh reported:

```
Cannot convert an Arrow array containing 4608 nulls into a non-nullable Vortex array.
```

Two places disagreed about the nullability of the derived embedding column.

- `crates/runtime-search/src/embeddings/table.rs` declared the chunked scalar derived columns `{col}_embedding` and `{col}_offset` with a non-nullable outer list. Full-text search enables chunking, so a full-text-search column took this arm. The unchunked scalar arm already used a nullable outer list.
- `crates/runtime-datafusion/src/optimizer_rule/avoid_vector_columns_on_index.rs` drops the derived columns from the index-population scan and re-adds them as all-null literals. This mirrors the vector index owning the vectors. The rule is correct and unchanged.
- The Vortex sink builds a non-nullable Vortex type from the write schema, then rejected the all-null column.

When an external vector index owns a derived column, that column is all-null on the index-write path. A non-nullable declaration is never valid in that configuration, independent of chunking.

## Fix

Relax the outer list of both chunked scalar derived columns to nullable. This aligns the chunked arm with the already-nullable unchunked scalar arm.

`AvoidDerivedVectorColumnOnIndexRule` nulls exactly the columns reported by `ChunkedVectorIndex::derived_columns`: `{col}_embedding` and `{col}_offset`. The fix relaxes exactly those two columns and no others.

The inner element types stay non-nullable. Only the outer list carries the nullability. The Arrow `DataType` is therefore unchanged, so it still matches the arrays that `execution_plan::get_embedding_columns` produces. The change relaxes a schema constraint only. Locally materialized embeddings still populate the columns fully.

## Downstream null tolerance

The query read path consumes the offset column through SQL expressions (`array_element`, `col`). These follow SQL NULL semantics. The embedding read path joins the base table against the index list provider, which populates the column. No consumer unwraps these arrays assuming a non-null outer list.

## Test

Added a regression test in `crates/runtime-search/src/embeddings/table.rs`. The test builds a chunked scalar embedding config, then asserts:

- both derived columns declare a nullable outer list;
- both inner element types stay non-nullable;
- an all-null column of each derived type builds into a `RecordBatch` against the derived schema. The arrow fork rejects an all-null column against a non-nullable field, so this reproduces the original failure mode and proves the fix admits the column.

Closes #12778

### DIFF
--- a/crates/runtime-search/src/embeddings/table.rs
+++ b/crates/runtime-search/src/embeddings/table.rs
@@ -704,6 +704,19 @@ impl EmbeddingTable {
         match (cfg.input_mode, cfg.chunker.is_some()) {
             // Scalar + chunked: doubly nested embedding + offsets
             // (character offsets of each chunk into the source string).
+            //
+            // The outer list of both derived columns is nullable. When an
+            // external vector index (e.g. S3 Vectors) owns the vectors, the
+            // index-population scan drops these columns and re-adds them as
+            // all-null literals — `AvoidDerivedVectorColumnOnIndexRule` in
+            // `runtime-datafusion` nulls exactly `{col}_embedding` and
+            // `{col}_offset` (see `ChunkedVectorIndex::derived_columns`). A
+            // non-nullable outer list rejects that all-null column when the
+            // accelerator write schema reaches the Vortex sink (#12778). The
+            // unchunked scalar arm below is already nullable for the same
+            // reason. Only the outer list is relaxed: the inner types stay
+            // non-nullable so the Arrow `DataType` still matches the arrays
+            // produced by `execution_plan::get_embedding_columns`.
             (EmbeddingInputMode::Scalar, true) => vec![
                 Arc::new(Field::new_list(
                     embedding_col!(field.name()),
@@ -713,7 +726,7 @@ impl EmbeddingTable {
                         cfg.vector_size,
                         false,
                     ),
-                    false,
+                    true,
                 )),
                 Arc::new(Field::new_list(
                     offset_col!(field.name()),
@@ -723,7 +736,7 @@ impl EmbeddingTable {
                         2,
                         false,
                     ),
-                    false,
+                    true,
                 )),
             ],
             // Scalar + unchunked: one vector per row.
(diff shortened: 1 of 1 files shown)


Jeadie:
Might be unnecessary after #12855

lukekim (approved):

Issue #12556, opened 2026-08-05
Sign-off reports 'checks failed' about the branch when sccache loses its storage and nothing is compiled


grokspice:
## Summary

The self-hosted pool compiles through `sccache` (`RUSTC_WRAPPER`, set by
`.github/actions/setup-sccache`) against an S3-compatible endpoint on the runner host.
When that endpoint stops answering, sccache's server cannot start and **every** `rustc`
invocation from that moment on dies before compiling anything.

A sign-off that hits this reports `signoff=failure` with the description
**`Sign-off checks failed after 5859s`** — a statement about a diff that nothing built.

## Signature

```
sccache: error: Server startup failed: cache storage failed to read: Unexpected (temporary) at read => send http request
  error sending request for url (http://127.0.0.1:8333/sccache/sccache/.sccache_check): client error (Connect): tcp connect error: Connection refused (os error 61)
error: process didn't exit successfully: `sccache /Users/runner/.rustup/toolchains/1.96.1-aarch64-apple-darwin/bin/rustc -vV` (exit status: 2)
debug: posting signoff=failure on spiceai/spiceai@b2cae1ec…: Sign-off checks failed after 5859s (triggered by grokspice)
```

## Why this is worse than an ordinary red

A failed sign-off **disqualifies the commit**, not just the run: `Attestation` rejects that
HEAD before anything else, an earlier sign-off cannot be inherited past it, and the revert
and no-Rust-changes fast-tracks cannot route around it. So an infrastructure hiccup on a
shared runner turns a green, approved PR red, and the only way out is another full multi-hour
sign-off. The remedy is a re-dispatch, but the run says nothing that would tell anyone that —
it reads as a code failure, 98 minutes in, with the real cause thousands of log lines up.

## Blast radius

Window 2026-08-04T13:16Z → 2026-08-05T11:05Z, 200 most recent failed runs:

- `pr` → `Attestation` → *Reject a failed sign-off on the head commit* — **13 failed runs
  across 11 distinct branches**. Every one of those PRs is blocked from the merge queue.
- `signoff.yml` over the same 7-day window (queried per workflow file, so uncapped): **109 failures of 383 concluded runs — 28%**.

Confirmed instance:
https://github.com/spiceai/spiceai/actions/runs/30978407363 (PR #12540, 5859s, then
`sccache: error: Server startup failed`) → the resulting red `Attestation`:
https://github.com/spiceai/spiceai/actions/runs/30978391211

[...]

PR #12788, opened 2026-08-08
fix(ci): recognise sccache's startup-timeout wording as an unusable cache (fixes #12622)

### LINKED ISSUE #12622: Sign-off cache guard misses the 'Timed out waiting for server startup' sccache wording
### Summary

`CACHE_UNUSABLE_PATTERN` in `scripts/signoff` recognises two sccache spellings:

```
sccache: error: Server startup failed|sccache: error: cache storage failed
```

A third spelling occurs in practice and matches neither, so a genuine compiler-cache outage is still classified `checks` and published as a failure *about the branch*:

```
sccache: error: Timed out waiting for server startup. Maybe the remote service is unreachable?
```

### Why it matters

This is the same mislabelling #12556 exists to prevent, just through a different wording. Nothing is compiled, yet the commit carries a `signoff=failure` that disqualifies it, and the author is sent to hunt a defect in a branch that was never built.

The case is not hypothetical — a `Build (release profile)` job failed this way with the diagnostic signature of a dead cache and no compiler diagnostics at all:

- zero `-->` lines in the log (no rustc ever reported on the source)
- `sccache --show-stats` in the same job reporting `Cache hits 0 / misses 7`
- two unrelated crates failing together

Notably the stats command *answered*, so the workflow's own cache-health warning never fired — the cache was reachable enough to report and dead enough to break every compile.

### Suggested fix

Extend the pattern to cover the timeout wording, e.g. add `sccache: error: Timed out waiting for server startup`.

Keep the existing anchoring discipline: only sccache's own `error:` channel counts, the verdict is only meaningful on a run that armed the watch, and disk continues to outrank cache when both appear.

### Note on narrowing

Narrowing the first alternative to `Server startup failed: cache storage failed…` was considered and rejected. It would widen this hole rather than close it, and the broader alternative is not harmful: a PR branch cannot supply sccache configuration (the endpoint and `SCCACHE_CONF` come from `.github/actions/setup-sccache`), so any `Server startup failed` on the pool is runner-side, where "nothing was compiled — re-dispatch" is the correct verdict and the correct remedy regardless of which startup cause produced it.

### Context

Found while resolving the base-branch conflict on #12557 (which introduces the guard). Out of scope for that PR's conflict resolution; filing separately so it is owned rather than lost.

claudespice:
## Summary

The sign-off's compiler-cache guard recognised two of sccache's spellings for an
unusable cache and not the third. When the third one occurred, a run that
compiled nothing still published `signoff=failure` — a verdict *about the
branch* — so the author was sent to hunt a defect in a diff that was never
built, and the commit carried a status that disqualifies it outright.

That is the same mislabelling #12556 exists to prevent, reached through a
different wording:

```
sccache: error: Timed out waiting for server startup. Maybe the remote service is unreachable?
```

### Root cause

`CACHE_UNUSABLE_PATTERN` was assembled from the failures we had happened to
observe, so it was a running tally rather than an enumeration. This change
derives the set from sccache's source instead.

`connect_or_start_server` is the whole of the path a `RUSTC_WRAPPER=sccache`
call takes, and it has four outcomes — `Ok` compiles, `AddrInUse` retries, and
the two that abort the compile are:

| outcome | sccache bails with | matched before? |
|---|---|---|
| `ServerStartup::Err` | `Server startup failed: {reason}` | yes |
| `ServerStartup::TimedOut` | `Timed out waiting for server startup` | **no** |

The third existing alternative, `cache storage failed`, is the storage
self-check's own bail for an invocation that reaches a live server and then
cannot read through it. sccache prints the first line of an error as
`sccache: error: {}` and every later cause as `sccache: caused by: {}`, so the
guard's existing discipline of anchoring on the first line is correct and is
kept.

### Changes

- Add a third alternative to `CACHE_UNUSABLE_PATTERN` for the startup timeout.
  It is matched on the **prefix** the two `TimedOut` bails share: sccache words
  it with a trailing `Maybe the remote service is unreachable?` from the wrapper
  path and bare from `--start-server`, so the prefix covers both.
- Record why that must stay a prefix. The pattern is an awk ERE, so the `?`
  ending the longer wording is a quantifier — spelling that sentence out would
  match `unreachabl` plus an optional `e` while reading as though it were
  deliberate. Verified: `awk 'BEGIN{ if ("…unreachabl" ~ /…unreachable?/) …}'`
  matches.
- Rewrite the constant's comment to carry the enumeration and its source, so the
  next wording that turns up is checked against sccache's outcomes rather than
  appended to a list.

Behaviour is unchanged for every input the guard already classified.

### Scope

The `cache` verdict changes a failing run's *description*, never its pass/fail:
it still publishes a failure status, so nothing here can move a branch toward
`signoff=success`. The anchoring on sccache's own `error:` channel is what keeps
a genuine defect from being excused because the sentence appeared in the log,
and that direction is tested.

Not changed: `Server startup failed: Address in use`, which the existing first
alternative already matches. It is a parallel-bootstrap race rather than a cache
outage, but narrowing the alternative to exclude it would reopen the hole this
guard exists to close, and re-dispatch is the right remedy either way.

## Test plan

Four cases added to `scripts/test_signoff_disk_guard.sh`, all through the
existing `assert_recorder` harness, which feeds lines through the real watcher
in `run_make_step` and reads the verdict back the way `failure_kind` does:

- the wrapper-path wording, with its unprefixed continuation line and the
  following `process didn't exit successfully` line — recorded as a cache hit;
- the bare `--start-server` wording — recorded as a cache hit;
- the sentence quoted *without* sccache's `error:` channel alongside a real
  `E0308` — left unmarked, so a genuine defect is not excused;
- (the existing cases continue to prove the disk signature still outranks cache
  and does not cross-fire).

Neuter matrix, one arm per direction of wrongness, run against the committed
tree — every arm isolates cleanly and the pass+fail sum is constant at 61, so
nothing was destroyed rather than neutered:

| arm | expected | failures |
|---|---|---|
| revert to the two old alternatives | the two positive tests | exactly 2 |
| pin the wrapper-path suffix | the bare-wording test | exactly 1 |
| drop the `sccache: error:` anchor | the cross-fire test | exactly 1 |

- `bash scripts/test_signoff_disk_guard.sh` — **61/61** (was 57).
- The other five sign-off suites re-run unchanged: attestation-refresh 15,
  cancelled-correction 18, resolve-target 22, status 34, trusted-helpers 16.
- `bash -n scripts/signoff` clean.
- `make lint` and `make test` via `make signoff`.

`.github/workflows/signoff_disk_guard_test.yml` already triggers on both changed
paths, so these run in CI on this PR.

## Checklist

- [x] Enumerated the wordings from sccache's source, not from observed failures
- [x] Regression tests fail on the unfixed pattern and pass on the fixed one
- [x] The "genuine defect is not excused" direction is tested
- [x] No change to any input the guard already classified

Fixes #12622

### DIFF
--- a/scripts/signoff
+++ b/scripts/signoff
@@ -818,26 +818,25 @@ readonly DISK_FULL_PATTERN='No space left on device|errno=28'
 # Anchored on sccache's own `error:` channel, which rustc, clippy and nextest
 # have no way to produce while reporting on the branch.
 #
-# The alternatives enumerate how sccache itself reports an unusable cache to a
-# wrapper invocation, taken from its source rather than from the spellings we
-# happen to have seen. `connect_or_start_server` — the whole of the path a
-# `RUSTC_WRAPPER=sccache` call takes — has exactly four outcomes, and the two
-# that abort the compile are `ServerStartup::Err`, which bails with
-# `Server startup failed: {reason}`, and `ServerStartup::TimedOut`, which bails
-# with `Timed out waiting for server startup`. `Ok` compiles and `AddrInUse`
-# retries, so neither can reach this channel. `main()` prints the first line of
-# that error as `sccache: error: {}` and every later cause as
-# `sccache: caused by: {}`, so the anchor belongs on the first line only.
+# The alternatives are taken from sccache's source rather than from the spellings
+# we happen to have seen, so the set is an enumeration and not a running tally.
+# `connect_or_start_server` is the whole of the path a `RUSTC_WRAPPER=sccache`
+# call takes, and it has four outcomes: `Ok` compiles and `AddrInUse` retries,
+# while `ServerStartup::Err` bails with `Server startup failed: {reason}` and
+# `ServerStartup::TimedOut` bails with `Timed out waiting for server startup`.
+# So those two are the ways an unusable cache reaches this channel, and the third
+# alternative below is the storage self-check's own `cache storage failed to
+# read: {…}`, for an invocation that reaches a live server and then cannot read
+# through it. sccache prints the first line of the error as `sccache: error: {}`
+# and every later cause as `sccache: caused by: {}`, so the anchor belongs on the
+# first line only.
 #
-# Hence three alternatives:
-#   - the startup failure, as observed on the pool;
-#   - the storage failure without the startup prefix, for an invocation that
-#     reaches a live server and then cannot read through it (sccache bails with
-#     `cache storage failed to read: {…}` from its storage self-check);
-#   - the startup timeout, matched on the prefix the two `TimedOut` bails share.
-#     sccache words it `Timed out waiting for server startup. Maybe the remote
-#     service is unreachable?` from the wrapper path and bare from
-#     `--start-server`, so the prefix covers both without pinning either.
+# The timeout is matched on the prefix its two bails share — sccache words it
+# `Timed out waiting for server startup. Maybe the remote service is
+# unreachable?` from the wrapper path and bare from `--start-server`. Keep it a
+# prefix: this string is an awk ERE, so the `?` ending the longer wording is a
+# quantifier, and spelling that sentence out here would match `unreachabl` plus
+# an optional `e` while reading as though it were deliberate.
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #12795, opened 2026-08-08
fix(connectors): tell an ABFS or GCS timeout from an authentication failure (fixes #12793)

### LINKED ISSUE #12793: ABFS and GCS report a request timeout as an authentication failure
## Summary

`ListingTableConnector::handle_object_store_error` classifies an `object_store::Error::Generic` by **which auth parameters are configured**, not by what actually failed. The S3 connector checks for a transport timeout first (`crates/runtime/src/dataconnector/s3.rs`), but its two siblings do not:

- `crates/data-connectors/connector-abfs/src/lib.rs` (~line 336)
- `crates/data-connectors/connector-gcs/src/lib.rs` (~line 238)

So on ABFS and GCS a request that timed out is reported as an **authentication failure**, and as `DataConnectorError::InvalidConfiguration` rather than a connection error.

## Impact

A body-read timeout — the exact failure #5980 reported for S3, common when a large dataset is loaded over a long-distance link or many datasets load concurrently — reaches the user as:

> Azure client credentials authentication failed. Verify your `client_id`, `client_secret`, and `tenant_id` are correct.

or, on GCS:

> GCS service account authentication failed. Verify your service account credentials are correct.

The credentials are fine. The user is sent to rotate secrets and re-check a service account for a network timeout, and the one parameter that resolves it — `client_timeout` — is never mentioned. Both connectors already declare `client_timeout` as a runtime `ParameterSpec` (`connector-abfs/src/lib.rs:197`, `connector-gcs/src/lib.rs:133`), so the advice S3 gives applies verbatim.

Which branch is hit depends on configuration, and both are wrong:

| configured auth | reported as |
| --- | --- |
| `client_id` / `sas_string` / `msi_endpoint` (ABFS), `service_account_path` / `service_account_key` / `application_default_credentials` (GCS) | authentication failure + `InvalidConfiguration` — **actively misleading** |
| none (anonymous / public bucket) | the opaque generic error #5980 filed for S3 |

## Root cause

`object_store` 0.13 classifies `reqwest`/`hyper`/I/O timeouts into `HttpErrorKind::Timeout` and then flattens the error into `Error::Generic`, so the classification survives only as a typed `HttpError` in the source chain. `s3.rs` walks that chain (`object_store_http_error_kind`) before considering auth. ABFS and GCS never look, so every `Generic` error is attributed to whichever credential happens to be set.

The helper is a private free function in `s3.rs`, which is why the sibling connectors could not reuse it.

## Suggested fix

Promote the source-chain timeout check to a shared helper on the `ListingTableConnector` surface in `runtime` (both `connector-abfs` and `connector-gcs` already depend on `runtime`), and call it **before** the auth-based branches in all three connectors, so a timeout is never attributed to credentials.

## Not affected

Connectors that inherit the trait default get the generic `UnableToConnectInternal` — wrong in the same way as the anonymous row above, but not misattributed to auth.

Raised while confirming #5980, whose S3 half is already fixed on `trunk`.

grokspice:
## Summary

`ListingTableConnector::handle_object_store_error` classified an `object_store::Error::Generic` by **which auth parameters were configured**, not by what actually failed. S3 checked the source chain for a transport timeout first; its two siblings did not.

So on ABFS and GCS a request that timed out was reported as an authentication failure — and as `InvalidConfiguration` rather than a connection error:

> Azure client credentials authentication failed. Verify your `client_id`, `client_secret`, and `tenant_id` are correct.

The credentials are fine. The user is sent to rotate a working secret for a network timeout, and `client_timeout` — the one parameter that resolves it, which both connectors already declare — is never mentioned.

**Root cause.** `object_store` 0.13 classifies `reqwest`/`hyper`/I/O timeouts into `HttpErrorKind::Timeout` and then flattens the error into `Error::Generic`, so the classification survives only as a typed `HttpError` in the source chain. The walk that recovers it was a private free function in `s3.rs`, which is why the sibling connectors could not reuse it.

Fixes #12793

## Changes

- `crates/runtime/src/dataconnector/listing/connector.rs` — promotes the source-chain walk to the `ListingTableConnector` surface as `object_store_timeout_error`, which returns an actionable error for a timeout and `None` for anything else. The documentation link is derived from the connector's own `Display`, which is its URL scheme.
- `crates/runtime/src/dataconnector/s3.rs` — drops the private copy and calls the shared helper. The rendered message is unchanged.
- `crates/data-connectors/connector-abfs/src/lib.rs`, `crates/data-connectors/connector-gcs/src/lib.rs` — call the helper **before** the auth-shaped branches, which is the fix.

Connectors that inherit the trait default are deliberately left alone: as #12793 notes, they are not misattributed to auth, and most declare no `client_timeout` parameter, so the advice would name one that does not exist.

## Test plan

New unit tests in each sibling connector, each exercising the real `handle_object_store_error` with the error shape `object_store` actually produces (a `Timeout`-kinded `HttpError` flattened into `Error::Generic`), with credentials configured so the pre-fix code takes the auth branch:

- a timeout with client credentials / a service account configured is reported as a timeout naming `client_timeout`, and **not** as an authentication failure — fails on the old code
- a timeout with a SAS token / application default credentials configured is likewise not an auth failure
- the configured `client_timeout` value is surfaced in the message
- a non-timeout (`Decode`) failure still reports its credential diagnosis and is not reported as a timeout — guards the check that was placed ahead of it

The existing S3 timeout tests are unchanged and continue to cover the shared helper through `S3`.

- [x] `make lint`
- [x] `make test`

## Checklist

- [x] The fix is minimal and stays in scope
- [x] Regression tests fail on the old code and pass on the new
- [x] No behaviour change for non-timeout errors on any of the three connectors

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4202,9 +4202,11 @@ checksum = "510ca239cf13b7f8d16a2b48f263de7b4f8c566f0af58d901031473c76afb1e3"
 name = "connector-abfs"
 version = "2.2.0-unstable"
 dependencies = [
+ "app",
  "linkme",
  "object_store 0.13.2",
  "runtime",
+ "runtime-secrets",
  "secrecy",
  "snafu",
  "tokio",
@@ -4462,9 +4464,12 @@ dependencies = [
 name = "connector-gcs"
 version = "2.2.0-unstable"
 dependencies = [
+ "app",
  "linkme",
  "object_store 0.13.2",
  "runtime",
+ "runtime-secrets",
+ "secrecy",
  "snafu",
  "tokio",
  "url",
--- a/crates/data-connectors/connector-abfs/Cargo.toml
+++ b/crates/data-connectors/connector-abfs/Cargo.toml
@@ -16,5 +16,10 @@ tokio.workspace = true
 tracing.workspace = true
 url.workspace = true
 
+[dev-dependencies]
+app = { path = "../../app" }
+runtime-secrets = { path = "../../runtime-secrets", default-features = false }
+tokio = { workspace = true, features = ["sync"] }
+
 [lints]
 workspace = true
--- a/crates/data-connectors/connector-abfs/src/lib.rs
+++ b/crates/data-connectors/connector-abfs/src/lib.rs
@@ -340,6 +340,17 @@ impl ListingTableConnector for AzureBlobFS {
     ) -> DataConnectorError {
         match error {
             object_store::Error::Generic { source, .. } => {
+                // A timeout carries no information about the credentials, so it has to
+                // be recognised before the auth-shaped branches below.
+                if let Some(timeout) = self.object_store_timeout_error(
+                    dataset,
+                    source.as_ref(),
+                    self.params.get("client_timeout").expose().ok(),
+                    "Azure Blob Storage",
+                ) {
+                    return timeout;
+                }
+
                 // Try to provide more specific error messages based on auth method
                 let has_msi = self.params.get("msi_endpoint").expose().ok().is_some();
                 let has_use_cli = self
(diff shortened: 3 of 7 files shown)


grokspice:
@copilot review

grokspice:
Closing as a duplicate of #12797, which fixes the same issue (#12793) better.

Both PRs were opened by our bots within 38 minutes of each other against the same issue — this one first — and they land on the same seven files with equivalent test coverage: an ABFS case, a GCS case, a case that the configured `client_timeout` is surfaced, and a case that a non-timeout error still reports the authentication failure it really was.

The difference is where the classification lives. This PR carries the timeout check into each connector. #12797 exports one `object_store_timeout_message` helper from `listing/connector.rs` and re-exports it through `listing/mod.rs`, so ABFS, GCS and S3 share a single implementation. That is the better shape for this defect specifically: it is a bug-class across sibling connectors, and one helper is what stops the next connector from acquiring the same bug independently.

Two smaller points in the same direction:

- #12797 also covers `listing/mod.rs`, so a connector added later picks up the shared helper from the module's public surface rather than having to know about `connector.rs`.
- This branch's HEAD carries a `signoff=failure` from a run that ended after 73 seconds — far short of the real gate — so it would need a fresh sign-off cycle on a currently saturated pool before it could enter the merge queue. #12797 does not.

Nothing here is lost: every case this branch tests has a counterpart in #12797. Reopen this one if #12797 turns out not to cover something.

PR #12797, opened 2026-08-08
fix(connectors): report an object-store timeout as a timeout, not as bad credentials (fixes #12793)

### LINKED ISSUE #12793: ABFS and GCS report a request timeout as an authentication failure

claudespice:
## Summary

`ListingTableConnector::handle_object_store_error` classifies an `object_store::Error::Generic` by **which auth parameters are configured**, not by what actually failed. `object_store` 0.13 flattens a transport timeout into that same `Generic` variant, keeping the classification only as a typed `HttpError` in the source chain — so ABFS and GCS, which never looked, attributed every `Generic` error to whichever credential happened to be set.

A request that timed out reached the user as:

> Azure client credentials authentication failed. Verify your `client_id`, `client_secret`, and `tenant_id` are correct.

The credentials were fine. The user was sent to rotate a working secret over a network timeout, and `client_timeout` — the one parameter that resolves it, and one both connectors already declare — was never mentioned. With no credential configured, the same failure produced the opaque generic error #5980 filed for S3.

S3 already classified the timeout first. This applies the same check to its two siblings, from one shared helper.

## Changes

- `crates/runtime/src/dataconnector/listing/connector.rs` — `object_store_timeout_message` (exported from `listing`) walks the source chain for `HttpErrorKind::Timeout` and returns the actionable message, or `None` when the error is not a timeout. It owns the default reported for an unset `client_timeout`, so the three connectors cannot drift on it. `handle_object_store_error`'s doc comment now states the ordering requirement.
- `crates/runtime/src/dataconnector/s3.rs` — uses the shared helper; the private `object_store_http_error_kind` moved into it. No behaviour change.
- `crates/data-connectors/connector-abfs/src/lib.rs`, `crates/data-connectors/connector-gcs/src/lib.rs` — consult the helper **before** the credential-based branches, so a timeout is never attributed to a credential.

## Test plan

- `make lint`
- `make test`

New unit tests in both connectors (neither crate had a test module before; `app` and `runtime-secrets` are added as dev-dependencies to build a `Dataset`, matching the existing S3 harness):

- a timeout with `client_id` / `service_account_path` configured names `client_timeout` and is **not** reported as an authentication failure — fails on the old code in both crates
- the same across every credential each connector branches on (`sas_string`, `msi_endpoint`; `application_default_credentials`), so the fix is not just ahead of the first branch
- a **non**-timeout `Generic` error still reports the configured auth method and is not mislabelled a timeout — pins the check against over-reach
- the configured `client_timeout` value is surfaced, so the message names the number to raise

The three existing S3 tests cover the refactored path unchanged.

## Checklist

- [x] Tests added and passing
- [x] `cargo clippy` clean on the touched crates (lib and `--tests`)
- [x] Error messages follow the repo's user-facing format (name the fix, link the docs)
- [x] No behaviour change on S3

Fixes #12793

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4202,9 +4202,11 @@ checksum = "510ca239cf13b7f8d16a2b48f263de7b4f8c566f0af58d901031473c76afb1e3"
 name = "connector-abfs"
 version = "2.2.0-unstable"
 dependencies = [
+ "app",
  "linkme",
  "object_store 0.13.2",
  "runtime",
+ "runtime-secrets",
  "secrecy",
  "snafu",
  "tokio",
@@ -4462,9 +4464,12 @@ dependencies = [
 name = "connector-gcs"
 version = "2.2.0-unstable"
 dependencies = [
+ "app",
  "linkme",
  "object_store 0.13.2",
  "runtime",
+ "runtime-secrets",
+ "secrecy",
  "snafu",
  "tokio",
  "url",
--- a/crates/data-connectors/connector-abfs/Cargo.toml
+++ b/crates/data-connectors/connector-abfs/Cargo.toml
@@ -16,5 +16,9 @@ tokio.workspace = true
 tracing.workspace = true
 url.workspace = true
 
+[dev-dependencies]
+app = { path = "../../app" }
+runtime-secrets = { path = "../../runtime-secrets", default-features = false }
+
 [lints]
 workspace = true
--- a/crates/data-connectors/connector-abfs/src/lib.rs
+++ b/crates/data-connectors/connector-abfs/src/lib.rs
@@ -18,6 +18,7 @@ use runtime::Runtime;
 use runtime::component::dataset::Dataset;
 use runtime::dataconnector::listing::{
     LISTING_TABLE_PARAMETERS, ListingTableConnector, ObjectVersionType, build_fragments,
+    object_store_timeout_message,
 };
 use runtime::dataconnector::parameters::{
     Validator,
@@ -40,6 +41,8 @@ use tokio::runtime::Handle;
 use url::Url;
 
 static PREFIX: &str = "abfs";
+
+const ABFS_DOCS: &str = "https://spiceai.org/docs/components/data-connectors/abfs";
 
 static VALIDATORS: LazyLock<
     Vec<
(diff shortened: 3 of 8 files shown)


claudespice:
@copilot review

phillipleblanc (approved):

PR #12807, opened 2026-08-08
feat(tools): add pdf-parse tool to compare liteparse and pdf-inspector

Jeadie:
## What

Adds `tools/pdf-parse`, a small CLI for comparing PDF text extraction between two backends:

- **liteparse** (PDFium-backed) — the parser `crates/document_parse` uses today.
- **pdf-inspector** (pure Rust, crates.io) — a candidate replacement being evaluated.

```
pdf-parse [--parser liteparse|pdf-inspector|both] [--format text|markdown] [--out DIR] <PDF_PATH>
```

Defaults to `--parser both`, printing each backend's extracted content under a labeled header plus a summary line (char count, line count, wall-clock parse time) so fidelity and speed can be compared at a glance. `--out DIR` writes each backend's output to a file for external `diff`.

## Why

We're evaluating whether to replace the liteparse/PDFium PDF path with a pure-Rust parser (removing the PDFium native library and its runtime download). Judging that swap needs a side-by-side fidelity and performance comparison on real documents. This tool provides exactly that, without touching the production `document_parse` crate.

The liteparse path deliberately reuses `document_parse`'s public `PdfParser` rather than re-wiring liteparse independently, so the comparison reflects the exact configuration the runtime ships (PDFium preload, `as_flat_utf8()` extraction).

## Notes

- Dev tool only; not linked into `spiced`. No changes to `crates/document_parse` or any shipping code path.
- `--parser` and `--format` are enums (no boolean config), per the repo's config conventions.
- On a small sample PDF both backends extract equivalent text; pdf-inspector is substantially faster and can additionally emit Markdown. A representative-corpus comparison (tables, multi-column, CID fonts) is the intended next use of this tool.


phillipleblanc (approved):

lukekim (approved):

claudespice:
Temporarily removed from the merge queue so the `Set up cc` fix can reach a batch — **please re-add; nothing is wrong with this PR.**

Every macOS job in the queue is currently failing `Set up cc` after ~109 ms:

```
Traceback (most recent call last):
  File "<string>", line 3, in <module>
PermissionError: [Errno 13] Permission denied: '/opt/homebrew/var/homebrew/locks/spice-global.lock'
```

`brew install` cannot take Homebrew's global lock, so it aborts before doing any work — on a runner where the toolchain is already installed and nothing needed doing. Observed on 7 distinct runners (`macmini-03-01/04-03/05-01/05-02/06-02/06-03`, `mbp-runner-01-03`) with **zero** `Set up cc` successes since 19:00Z. Tracked as #13463.

#13466 fixes it by probing the Cellar and PATH first and only invoking `brew` when something is actually missing. It is verified green (`Set up cc: success` ×3 on three distinct runners).

The problem is ordering: a merge queue speculates chains forward from position 1, so a fix at position 10 is in **no** live batch — the live chain is only 4 deep. Positions 1–4 had been `AWAITING_CHECKS` for 3h08m without ejecting, because the chain is rebuilt faster than the failing run can eject the head, so the fix could not advance on its own. Trunk has been frozen at `325c9fc37b` for 20h.

Removing the rows ahead lets #13466 reach the speculation window; once it merges, this PR should sail through unchanged.

PR #12821, opened 2026-08-08
fix(cayenne): hand out a pooled Turso connection in autocommit (refs #12820)

### LINKED ISSUE #12820: cayenne: prop_sequential_cold_impl_turso intermittently loses rows across a cold-tier promotion, dequeuing merge-queue candidates
## Summary

`cayenne::mutation_property_test::prop_sequential_cold_impl_turso` intermittently fails its convergence assertion with **lost rows** — keys the model applied and acknowledged that the live table no longer returns. The SQLite variant of the same property (`prop_sequential_cold_impl_sqlite`) passed in the same run, so the divergence is specific to the Turso metastore backend.

This is not the stack overflow from #12436 (fixed by #12561). That issue is why this test only recently began running in CI at all; now that it does, it is reporting a different and more serious failure.

## Evidence

Merge-queue candidate `11869cf5c37c8946a84d36eb9dde84bfd90bb5be`, job [93094935700](https://github.com/spiceai/spiceai/actions/runs/31253251713/job/93094935700) (`Build and Test`, `spiceai-macos-macmini-runner-03-03`):

```
FAIL [ 1.378s] ( 9997/10006) cayenne::mutation_property_test prop_sequential_cold_impl_turso

thread '<unnamed>' panicked at crates/cayenne/tests/mutation_property_test.rs:516:5:
seq diverged after step 47 (Upsert { rows: [(1, 757802)] }) mode=Key durability=File seed=0
missing(loss)=[(2, 879959), (7, 292314)]
extra(resurrect)=[]
wrong_value(k,expected,got)=[]
```

Two keys are missing; nothing was resurrected and no value was wrong. `Summary [541.370s] 10006 tests run: 10005 passed, 1 failed, 23 skipped` — this was the only failure in the job.

**It is intermittent, not deterministic.** A sibling merge-queue candidate built minutes earlier in the same batch ran the identical test and passed it: job [93094920987](https://github.com/spiceai/spiceai/actions/runs/31253252258) shows `PASS [28.040s] (10004/10012) cayenne::mutation_property_test prop_sequential_cold_impl_turso` with `10012 tests run: 10012 passed`. Note also the duration — 28.0s on the pass versus 1.378s on the failure, i.e. the failing walk aborted early rather than running slow.

`seed=0` is printed on the failure, so the operation sequence itself is seeded; the divergence therefore comes from something outside the seeded op stream (timing/ordering in the promotion or GC path), not from a differently-generated workload.

## Why it matters

Two separate reasons, and the second is the one worth prioritizing:

1. **It dequeues merge-queue candidates.** `Build and Test` is required, so a candidate that trips this is marked `UNMERGEABLE` and removed. It took out the queue head (PR #12694, a Pingora results-cache change with nothing to do with Cayenne) on 2026-08-08, in front of a 21-deep queue.
2. **If it is not a test artifact, it is row loss in the Turso cold path.** `assert_converged` compares a live read against the model of applied operations; `missing(loss)` means rows the workload applied and the test observed as applied are no longer readable. `mode=Key durability=File` is a durable, non-memory configuration.

## Where to look

The `cold` workload interleaves `Op::MoveToColdTier` — `materialize_warm_files_for_cold_promotion` → `promote_warm_to_cold()` → `run_cold_tier_gc_tick()`, the production promotion-tick order — with ordinary mutations, and asserts convergence after **every** step (`crates/cayenne/tests/mutation_property_test.rs`, `prop_sequential_cold_impl` at ~1286, driver assertion at ~799). So the failing step is the first one at which the loss became visible, not necessarily the one that caused it: the preceding `MoveToColdTier` is the more likely origin.

That the SQLite backend passes the same property while Turso fails points at the metastore rather than the tier logic itself — a promotion or GC commit that is visible to one backend's readers and not the other's. Possibly related: #12727 (Turso metastore sidecars each build their own pool, so the schema-conflict gate serializes nothing), open as PR #12804. I have not confirmed a connection, and it should not be assumed.

## Reproduction

```
cargo nextest run -p cayenne --features default,partition-table-provider,turso \
  prop_sequential_cold_impl_turso
```

Expect to need repeated runs — it passed in one CI job and failed in another within the same hour.

## Acceptance criteria

- [ ] The cause of the two missing keys is identified as either a test-harness ordering gap or a genuine loss in the Turso cold promotion/GC path.
- [ ] If it is a real loss, a regression test fails without the fix and passes with it.
- [ ] If it is a harness gap, the test is made deterministic rather than excluded — this property is the only thing currently checking cross-tier convergence on this backend.

grokspice:
## Summary (root cause)

A Turso metastore connection can return to the K=16 pool while it is still inside `BEGIN CONCURRENT`:

- `TursoTransaction`'s `Drop` hands its `ROLLBACK` to a spawned task. If that future is dropped without running — the runtime is shutting down — the guard inside it is dropped and the slot unlocks with the transaction still open.
- `commit` and `rollback` `take()` the connection before issuing their statement, so they release it even when that statement fails. `commit`'s fallback `ROLLBACK` is best-effort and its result is discarded.

Under `BEGIN CONCURRENT` an open transaction holds an MVCC snapshot, so the **next** statement on that connection reads the state the abandoned transaction saw, not the committed database. A reader on that connection can miss rows another connection has already committed.

`begin_transaction` already defended against exactly this for itself, with an unconditional `ROLLBACK` before its `BEGIN`, and its comment named the `Drop`/`tokio::spawn` path as the reason. The **non-transactional** borrowers of the same pool — `execute`, `execute_batch`, `query_row`, `query` — had no such guard. Those are the reads that serve the cold-tier GC root (`list_cold_tier_files`), which is where a stale live-file list turns into a deleted file that the manifest still references.

`SqliteMetastore` needs no equivalent because `BEGIN IMMEDIATE` takes no snapshot — which is why this divergence is Turso-only.

## Changes

- `TursoConnectionPool::acquire` now clears any transaction the previous holder left open before returning the guard, so every borrower is covered in one place rather than each remembering to defend itself. The round-robin pick moves to `lock_next_free` unchanged.
- `begin_transaction` drops its own blind `ROLLBACK`, now redundant.
- **`acquire` is fallible, and only releases a slot it has confirmed is in autocommit** (review follow-up). A failed `ROLLBACK` propagates instead of being logged and swallowed, and a `ROLLBACK` that reports success is re-checked with `is_autocommit()` before the guard is handed over — `Ok(false)` and an unreadable state are both refusals. Without this, a slot whose cleanup did not take still served the snapshot the cleanup exists to discard: the same stale read by a different route. Every caller already returned `CatalogResult`, so each needed only a `?`.
  Propagation rather than discard/reopen is deliberate: the pool holds `Connection`s and not the builder, so reopening would mean threading the connection string and open options into `TursoConnectionPool`, and a connection whose `ROLLBACK` fails is one whose next statement would fail anyway. The cost is at most a failed acquire on 1 of 16 slots; what it replaces is a silent stale read that makes the cold-tier GC delete a live file.
- `docs/cayenne/cayenne.md` documents the acquire-time autocommit guarantee in *Metastore backends* and records it in the *Document changelog* (review follow-up).

## Performance impact

Net one fewer round trip on the hot path. `Connection::is_autocommit` reads connection-local state — no statement, no I/O — so a clean connection (the overwhelmingly common case) pays nothing on acquire, and only a connection that actually needs clearing issues a `ROLLBACK`. Previously *every* `begin_transaction` spent a real `ROLLBACK` round trip unconditionally; that is now gone. This is the same predicate `turso`'s own `Transaction::_finish` uses to decide whether cleanup is needed.

## Test plan

Four tests in a new `mod tests` in `crates/cayenne/src/metastore/turso.rs`, driving slots directly so the leak is deterministic rather than dependent on a runtime-shutdown race:

- `a_connection_left_in_a_transaction_does_not_serve_a_stale_snapshot` — the regression test. Slot 0 abandons a transaction after pinning a snapshot of an empty table, another connection commits a row, and slot 0 must then read the committed row. Fails without the fix (reads 0), passes with it.
- `acquire_never_hands_out_a_stale_snapshot` — the same guarantee through the production entry point, acquiring pool-width times so round-robin visits the leaked slot.
- `clearing_an_abandoned_transaction_restores_autocommit` — the invariant `acquire` relies on.
- `clearing_a_clean_connection_leaves_it_usable` — clearing is inert on the path every ordinary acquire takes.
- `a_connection_not_confirmed_in_autocommit_is_refused` — the decision `acquire` makes once cleanup has run, pinned on all three outcomes: a confirmed autocommit is handed over, while a connection still in its transaction and one whose state cannot be read at all are both refused with an error naming which it was.

- [x] `cargo fmt --all`
- [x] Type-checked `--lib --tests` with `--features turso`
- [x] `make signoff` dispatched remotely against this HEAD (see *Verification limits* — local sign-off cannot run on this host)
- [ ] `make lint` — runs inside the remote sign-off
- [ ] `make test` — runs inside the remote sign-off

### Verification limits — please read

I could not run these tests or the lint locally, and want that on the record rather than implied by a checked box:

- `crates/cayenne` does not compile on Windows (ungated Unix-only `libc`/`as_raw_fd` code in `provider/compaction_writer.rs`), so `cargo test` and `make lint` cannot run here.
- To be sure the type-check was still meaningful, I planted a deliberate type error first in `clear_open_transaction` and then inside the new `mod tests`, and confirmed `cargo check --lib --tests` reported each at the right line in `turso.rs`. With both probes removed the error set is exactly the 8 pre-existing `compaction_writer.rs` errors and nothing in `turso.rs`. So the new code and its tests do type-check; only their *runtime* behaviour is unverified locally.
- Clippy could not be completed either: it aborts on a pre-existing pedantic failure in the `runtime-request-context` dependency before reaching `cayenne`.

- Local `make signoff` is therefore blocked for the same reason (and `make` itself is not installed on this host), so the sign-off was dispatched to the shared runner pool with `gh workflow run signoff.yml -f pr_number=12821`. Naming it here rather than letting it look like a local run.
- The review-follow-up commit was type-checked the same way: a probe planted in `require_autocommit` and then in the new test was reported at the right line in `turso.rs` each time, and with both removed the error set is again exactly the 8 pre-existing `compaction_writer.rs` errors.

CI is the first place these tests actually execute.

## Why this references #12820 rather than closing it

Refs #12820.

The defect fixed here is real and independently demonstrable, and it produces exactly #12820's signature — `missing(loss)` with nothing resurrected and no wrong value, i.e. a physically deleted file rather than a bad merge or tombstone — on the Turso backend only, in a sequential walk. It is the mechanism that issue's analysis pointed to.

But I have not **confirmed** it is what fails that property test. The failure is intermittent and reproduces only on macOS CI, and I cannot run the suite locally. #12820's first acceptance criterion is that the cause be *identified*, and honestly I can only say this is one confirmed stale-read path on the way there. Closing the issue on that basis risks retiring it while the actual cause is still open, so this deliberately does not auto-close it.

Suggested way to settle it: land this, then watch `prop_sequential_cold_impl_turso` across a few merge-queue batches. If it stops failing, close #12820 against this PR; if it recurs, the issue is still open with one path eliminated and the instrumentation it asks for (GC live-root size vs. the promotion's committed file count) is the next step.


grokspice:
@copilot review

phillipleblanc (dismissed):

lukekim (approved):

PR #12825, opened 2026-08-08
fix(search): leave the stored rows in place when a delete cannot be applied (fixes #12822)

### LINKED ISSUE #12822: A failed delete empties the whole in-memory vector index, because the batch list is drained before it can fail
## Describe the bug

`MemoryVectorStore::delete_by_keys` (`crates/search/src/index/memory/store.rs`) drains the whole batch list into a local before it can fail, and the failure path drops that local:

```rust
for stored in target.drain(..) {
    ...
    let filtered = filter_record_batch(&stored.batch, &mask)
        .map_err(|e| DataFusionError::ArrowError(Box::new(e), None))?;   // <-- early return
    ...
    retained.push(...);
}
*self.write_target() = retained;
```

`drain(..)` empties the store's batch vector as it iterates. If `filter_record_batch` returns an error on batch *n*, the `?` returns before `retained` is written back, so `retained` — holding every batch already processed — is dropped, and the batches not yet reached were consumed by the drain. **The store is left empty**: every indexed row is gone, not just the keys the caller asked to delete.

The caller sees an `Err`, but nothing puts the rows back, and the memory vector tier is the primary read tier for a warm-index dataset — so search returns nothing for that dataset until the next full refresh repopulates it.

## Impact

A single Arrow `filter` failure during a delete silently escalates from "this delete failed" to "the whole in-memory index is empty". `Index::delete_by_keys` is on the DML delete path and on the chunked-index cleanup path, so an ordinary `DELETE` can trigger it.

`filter_record_batch` failing is not routine — the mask is built to the batch's own row count — which is why this has not been observed. It is still the wrong shape for a store that holds the only copy of its data.

## Expected behavior

A failed delete leaves the store exactly as it was. Either build `retained` without draining (iterate by reference and only commit the replacement once every batch has been filtered successfully), or restore the drained batches before returning the error.

## Additional context

Pre-existing; noticed while adding the `ReplaceAll` replace window to this file for #12413 and deliberately left out of that PR's diff to keep it to one concern. The same shape now applies to a replace window's staged batches, since both go through `write_target()`.

grokspice:
Fixes #12822

## Summary (root cause)

`MemoryVectorStore::delete_by_keys` built its replacement list while draining the
list it was replacing:

```rust
for stored in self.batches.drain(..) {
    ...
    let filtered = filter_record_batch(&stored.batch, &mask)?;  // early return
    ...
    retained.push(...);
}
self.batches = retained;
```

`drain(..)` empties the store's vector as it iterates, and `Drop` for `Drain`
clears whatever the iteration never reached. So an error on batch *n* returned
before `retained` was written back: the batches already rewritten were dropped
with the local, and the ones not yet visited were dropped by the drain. **The
store was left empty** — a failed delete removed every indexed row instead of
none.

The memory vector tier is the primary read tier for a warm-index dataset, so
until the next full refresh repopulated it, search returned nothing for that
dataset. `Index::delete_by_keys` is on the DML delete path and the chunked-index
cleanup path, so an ordinary `DELETE` reaches it.

## Changes

`delete_by_keys` now filters every overlapping batch **before** it touches the
stored ones, and only rebuilds the store once nothing left can fail. Batches with
no overlap are still moved across untouched, so the zero-copy path for them is
unchanged. The guarantee is now written on the method: on `Err` the store holds
exactly what it held before the call.

### Same shape in the Arrow accelerator

`filter_existing` (`crates/data_components/src/arrow/write.rs`) — the `InsertOp::Replace`
path of the in-memory Arrow accelerator's `MemSink` — drained the table's own
`Vec<RecordBatch>` across two fallible steps (`extract_primary_keys_str` and
`filter_record_batch`), with the same consequence for the same reason. It now
reads the batches instead of draining them; the loop body already only borrowed
each batch, so this is the same computation with the input left intact.

Included because it is the same defect rather than a related one, and the change
is confined to how the loop obtains each batch. Worth flagging for review,
though: I could not construct an input that actually makes it fail today. The
common primary-key types are handled by `extract_primary_keys_str`'s direct
paths, and the fallback's `ScalarValue::try_from_array` accepted every exotic
type I tried, including run-end encoding. So that half is hardening against a
latent shape, not a reachable failure, and it carries no regression test. Happy
to split it out if you would rather keep this PR to the one crate.

## Performance impact

`delete_by_keys` now holds the filtered results alongside the originals until it
commits, where before it released each original as it was consumed. Untouched
batches cost nothing extra — they are never copied — so the additional transient
memory is bounded by the retained rows of the batches a delete actually overlaps.
A delete touching a few keys is unchanged in practice; one that removes a small
number of rows from every batch is the worst case, roughly doubling peak memory
for the duration of the call.

The alternative that keeps peak memory flat is to pre-check `keys.len() ==
batch.num_rows()` and rely on that making `filter_record_batch` infallible. I did
not take it: it encodes an Arrow implementation detail, and a future failure mode
would silently restore the data loss. Correctness of the only copy of the rows
seemed the right thing to spend the memory on.

## Test plan

- [x] `make lint`
- [x] `make test`
- New tests in `crates/search/src/index/memory/store.rs` (the file had none):
  - `a_failed_delete_leaves_every_stored_row_in_place` — the regression test. A
    store whose third batch carries one more key than it has rows, so the mask
    built from those keys is longer than the batch and Arrow's filter fails on
    it, after the second batch has already been filtered successfully.
  - `a_failed_upsert_leaves_every_stored_row_in_place` — `upsert` deletes the
    superseded rows first, so it inherits the same failure; asserts it neither
    drops the stored rows nor adds its own.
  - `a_delete_removes_only_the_named_rows`, `a_batch_a_delete_empties_is_dropped`,
    `an_empty_key_list_leaves_the_store_alone` — the success paths, so the
    reordering above is pinned as behaviour-preserving.

Both failure tests were confirmed to fail on the pre-fix code and pass on this
one. On the pre-fix code they report `left: []` against
`right: [[1, 2], [3, 4], [5, 6]]` — the store emptied — while the three
success-path tests still pass, which is the reported bug exactly.

## Notes for the reviewer

- The two failure tests deliberately violate `StoredBatch`'s documented
  parallel-keys invariant. A length mismatch is the only way to make Arrow's
  filter fail, and the point of the tests is what the store does when it does.
  The helper that builds that state says so.
- #12823 changes this same function to write through a replace window
  (`write_target()`), and will conflict here. That PR is mine and is not yet
  approved; I will take the conflict there rather than hold this one behind it,
  since the data loss is live on `trunk` today and the replace window is not.

## Checklist

- [x] Regression test fails on the old code, passes on the new
- [x] Success paths covered so the reordering is pinned
- [x] `cargo clippy --all-targets` clean on both crates
- [x] Same bug class swept for across the repo (`pk_index.rs`'s drain has an
      infallible body, so it is not affected)

### DIFF
--- a/crates/data_components/src/arrow/write.rs
+++ b/crates/data_components/src/arrow/write.rs
@@ -1077,7 +1077,9 @@ fn constraint_identifiers(rb: &[&RecordBatch], constraint_idx: &[usize]) -> Resu
 ///
 /// This is one part of `InsertOp::Replace` functionality, and still requires the new rows (with conflicting PKs), to be added.
 ///
-/// This function modifies `existing_batches` in place.
+/// This function modifies `existing_batches` in place. All-or-nothing: on `Err` it still
+/// holds exactly the batches it held before the call, so an overwrite that cannot be applied
+/// does not take the rows it was filtering with it.
 ///
 /// # Visibility
 /// This function is public for benchmarking purposes.
@@ -1105,10 +1107,15 @@ pub(crate) fn filter_existing<S: std::hash::BuildHasher>(
         None
     };
 
-    // Instead of concatenating, we can filter each batch individually
+    // Instead of concatenating, we can filter each batch individually.
+    //
+    // Read the batches rather than draining them: both steps below are fallible, and
+    // `existing_batches` is the table's only copy of these rows. Draining as we went would
+    // leave it empty on the error path, turning a failed overwrite into the loss of every
+    // row the table held.
     let mut filtered = Vec::with_capacity(existing_batches.len());
-    for batch in existing_batches.drain(..) {
-        let keys = extract_primary_keys_str(&batch, pk_indices_ordered)?;
+    for batch in existing_batches.iter() {
+        let keys = extract_primary_keys_str(batch, pk_indices_ordered)?;
 
         // Pre-allocate with exact capacity for better performance
         let mut keep_row_builder = BooleanBuilder::with_capacity(keys.len());
@@ -1127,12 +1134,13 @@ pub(crate) fn filter_existing<S: std::hash::BuildHasher>(
                 );
             }
         }
-        let filtered_batch = filter_record_batch(&batch, &keep_row_builder.finish())?;
+        let filtered_batch = filter_record_batch(batch, &keep_row_builder.finish())?;
         if filtered_batch.num_rows() > 0 {
             filtered.push(filtered_batch);
         }
     }
 
+    // Every fallible step is done, so the batches can be replaced without losing rows.
     *existing_batches = filtered;
     Ok(())
 }
--- a/crates/search/src/index/memory/store.rs
+++ b/crates/search/src/index/memory/store.rs
@@ -78,21 +78,28 @@ impl MemoryVectorStore {
     }
 
     /// Remove every stored row whose formatted primary key appears in `keys`.
+    ///
+    /// All-or-nothing: on `Err` the store still holds exactly the rows it held before the
+    /// call, so a delete that cannot be applied does not take the rows it was filtering
+    /// with it.
     pub(crate) fn delete_by_keys(&mut self, keys: &[String]) -> Result<(), DataFusionError> {
         if keys.is_empty() {
             return Ok(());
         }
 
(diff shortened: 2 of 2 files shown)


grokspice:
@copilot review

phillipleblanc (approved):

Issue #12730, opened 2026-08-07
Catalog acceleration readiness timeouts cannot be diagnosed from their own message


grokspice:
`wait_for_table_ready` in `crates/runtime/tests/postgres/catalog_changes.rs` fails with a
message that cannot be acted on:

```
Error: accelerated table pg_accel_e2e.public.orders never became queryable
```

Three things a reader needs are missing, and each of them costs a full job-log download to
recover.

**1. It does not say which call site timed out.** There are 13 call sites and the message
is identical at all of them. `test_catalog_acceleration_reuses_slot_across_restart` waits
on `orders` twice — once at bootstrap and once after a restart — so when it fails, the
output does not distinguish "the catalog never bootstrapped" from "the catalog
bootstrapped and then lost its data across the restart". Those are unrelated causes.
Establishing which one applied to #12729 meant downloading the 36,000-line job log and
counting the polls to find that the first wait passed on its very first query.

**2. It does not distinguish "the query failed" from "the count was zero."**
`query_i64` collapses both into `None`:

```rust
async fn query_i64(rt: &Arc<Runtime>, sql: &str) -> Option<i64> {
    let batches = run_query(rt, sql).await.ok()?;   // <- error discarded
    ...
}
```

A query that errors means the table is not registered at all; a query returning 0 means it
is registered and empty. The remedies are completely different, and the current output
picks neither.

**3. It does not say what it saw or how long it waited.**

Separately, the 100ms poll interval means a wait that runs to its two-minute timeout issues
~1,200 `COUNT(*)` queries, each emitting several `task_history` log lines. Those thousands
of lines surround and bury the one line that matters — enough that an automated log scan
over these runs extracted an unrelated `Pulling from library/postgres` debug line as the
failure signature rather than the actual error.

## Fix

Give the helper a `phase` label, record the last poll's outcome (count or error), report
both plus the elapsed time, and slow the poll to an interval that still resolves a
two-minute budget finely without generating thousands of log lines.

Also guards a latent panic: `query_i64` called `arr.value(0)` without checking the batch
was non-empty.

claudespice:
**Already fixed on `trunk`** by #12731 (`test(postgres): say which readiness wait timed out and what
it saw`), merged 2026-08-09. That PR carried `fixes #12730` in its **title only**, with no closing
line in its body, so the merge did not retire this row — the fix has been in place for 18 days with
the issue still open.

[...]

Issue #12396, opened 2026-08-03
CI: integration tests (models) has never passed in 200 runs — three independent job failures plus 40 zero-job startup failures


grokspice:
## Summary

`integration tests (models)` (`.github/workflows/integration_models.yml`) has **not passed once
in its last 200 runs** — 199 failures and 1 cancelled, spanning **2026-07-12 → 2026-08-03**.
There are no successes to compare against, so the workflow currently carries no information:
it is red before anyone changes anything, and a real regression in the model/AI integration
surface would look exactly the same as today.

The red splits into **two structurally different problems**, and they need different owners.

## Blast radius

Window: last 200 runs of the workflow, `2026-07-12T12:40:52Z → 2026-08-03T05:05:59Z`.

| Event | Runs | Outcome |
|---|---|---|
| `schedule` (trunk, daily 00:00 UTC) | 21 | 20 failure, 1 cancelled — **0 success** |
| `push` (release-lineage branches) | 40 | 40 failure, **all with zero jobs** |
| other | remainder | failure |

Every one of the 21 scheduled trunk runs since 2026-07-14 has failed. Sample runs:

- https://github.com/spiceai/spiceai/actions/runs/30774665978 (2026-08-03, `schedule`, trunk)
- https://github.com/spiceai/spiceai/actions/runs/30725362819 (2026-08-02, `schedule`, trunk)
- https://github.com/spiceai/spiceai/actions/runs/30675682012 (2026-08-01, `schedule`, trunk)

## Problem 1 — the scheduled trunk run fails in three independent places

From run `30774665978`, three jobs fail, for three unrelated reasons. None is a flake in the
"passes on retry" sense; all three reproduce daily.

### 1a. `Test Bedrock Models` — the runtime never becomes ready

Four Bedrock embedding tests fail identically, each after ~121s:

```
thread 'models::bedrock::embeddings::bedrock_test_titan_v2_embeddings' panicked at
  crates/runtime/tests/utils/mod.rs:59:5:
assertion failed: wait_until_true(duration, || async { rt.status().is_ready() }).await
```

Affected: `bedrock_test_titan_v1_embeddings`, `bedrock_test_titan_v2_embeddings`,
`bedrock_test_cohere_english_embeddings`, `bedrock_test_cohere_multilingual_embeddings`.

The failure is the *readiness wait*, not an assertion about embeddings — the runtime never
reaches ready, so the model component never loads. The job log also carries
[...]

Issue #12824, opened 2026-08-08
cayenne: prop_concurrent_mixed_dense_key_sqlite serves Exact(9) while 11 rows are live, so a distributed COUNT(*) undercounts


grokspice:
## Summary

`cayenne::mutation_property_test::prop_concurrent_mixed_dense_key_sqlite` intermittently fails the **maintained-count exactness invariant**: `optimizer_table_statistics()` served `Precision::Exact(9)` while **11 rows were live**.

This is not a timing flake. The invariant it breaks is the distributed-gate contract, stated in the test's own comment (`crates/cayenne/tests/mutation_property_test.rs:574-580`):

> the count `local_executor_table_statistics` reports to the coordinator — and that the coordinator folds `COUNT(*)` on ONLY when Exact — must never be `Exact`-and-wrong. If it is `Exact(n)`, n must equal the live row count; a drifted count must instead be served `Inexact` (so the fold declines).

So the consequence is a **distributed `COUNT(*)` returning 9 for a table holding 11 rows**, with no error and nothing to indicate the count is stale.

## Evidence

Run [31261924493](https://github.com/spiceai/spiceai/actions/runs/31261924493) (a remote sign-off for PR #12804):

```
FAIL [  44.323s] (2427/9989) cayenne::mutation_property_test prop_concurrent_mixed_dense_key_sqlite

thread 'prop_concurrent_mixed_dense_key_sqlite' panicked at crates/cayenne/tests/mutation_property_test.rs:584:9:
assertion `left == right` failed: concurrent convergence failed mode=Key durability=File seed=1
  left: 9
 right: 11
```

`left` is `n` from `Precision::Exact(n)`; `right` is `model.len()`, the live row count.

```
Summary [967.724s] 9989 tests run: 9988 passed (7 slow), 1 failed, 23 skipped
```

It was the only failure in the run.

## Why this is a real drift, not a mis-modelled test

The plain `COUNT(*)` assertion **20 lines earlier** (`:568`) passed on the same state, so the table itself held the right number of rows and the single-node answer was correct. Only the *maintained* statistic had drifted — which is exactly the case the invariant exists to catch, per its comment: it catches "drift served Exact" defects that the plain `COUNT(*)` query, which single-node answers from footer sums, can miss.

Two facts narrow it further:

- **Concurrent-only.** The sequential variants over the same mode/durability pass.
[...]

Issue #12710, opened 2026-08-07
bug(ci): a cancelled sign-off publishes a false "checks failed" verdict its own correction handler cannot take back


grokspice:
## Summary

A remote sign-off that is cancelled mid-checks publishes `signoff=failure` with the description
`Sign-off checks failed after <N>s`, even though nothing judged the branch. The two mechanisms
meant to prevent exactly this both miss, for two independent reasons:

1. `failure_kind()` only recognises a signalled run by `check_status >= EXIT_SIGNAL_BASE`
   (128+signum). A cancellation signals the whole process group, and `make` traps it and exits
   with a small status of its own, so the status the script actually observes is an ordinary
   non-zero one. Classification falls through to `checks`, and the `signalled` branch of
   `describe_check_failure()` — whose entire purpose is to publish no verdict — never runs.
2. `scripts/signoff correct-cancelled`, the `if: cancelled()` handler that is supposed to take
   such a status back, runs *before* the dying step's post lands. It reads the status, sees the
   `pending` the run posted on its way in, correctly declines to touch it, and exits — and the
   false `failure` is written seconds later.

The result is a red `Attestation` on a commit that no check ever failed, and no handler left that
can withdraw it. Re-dispatching is the only recovery, which costs another 1-4h pool run.

This is distinct from the two neighbouring issues:

- #12657 (PR #12662) is the case where the helper is a *stale* copy missing `correct-cancelled`
  entirely. Not this: the branch below carries the subcommand (2 occurrences, same as trunk), the
  handler ran, and it exited 0.
- #12424 / #12428 are about `correct-cancelled` preserving a genuine `success`. Here there is no
  earlier success to preserve; the problem is the false `failure` written after it ran.

## Evidence

Run [31144830314](https://github.com/spiceai/spiceai/actions/runs/31144830314) — Remote Sign-off
(PR #12652), `conclusion: cancelled`, on commit `bca5359c0bb6`.

Step timeline:

| Step | Result |
| --- | --- |
| 15. Sign off | cancelled |
| 16. Correct the sign-off status when this run is cancelled | **success** |

The handler's own output, at 05:16:20Z:

```
signoff on bca5359c0bb6 is 'pending', not a failure this run posted — leaving it alone.
```

The `signoff` commit-status history on that same commit:

```
2026-08-07T05:16:24Z  failure  :: Sign-off checks failed after 338s (triggered by grokspice)
[...]

PR #12827, opened 2026-08-08
test(runtime): say which components were not ready when the wait times out (refs #12396)

grokspice:
## Summary

The shared integration-test readiness helper asserted on the wait alone:

```rust
assert!(wait_until_true(duration, || async { rt.status().is_ready() }).await);
```

so a timeout printed only that the assertion failed. That message is identical whether a
component failed to load, is still initializing, or was never registered, which leaves the
failure undiagnosable from its own output.

The daily `integration tests (models)` Bedrock job is the live example. All six embedding tests
fail on this assertion after ~121s, and the only other line the run carries is that both secret
references resolved — nothing says which component is not ready, or why.

`RuntimeStatus::is_ready` explains the silence: under the default `RuntimeReadyState::OnLoad` it
requires every registered component to have reached `Ready` at least once, and `Error` does not
satisfy that. A component that fails to load leaves the runtime permanently not-ready and the
wait simply elapses.

## Changes

`runtime_ready_check_with_timeout` now names every registered component and its status in the
panic message. `ComponentStatus::Error` carries the message that put the component in that
state, so the assertion reports the reason rather than only the timeout:

```
the runtime did not become ready within 120s — bedrock: Initializing, titan-v1: Error(Some("..."))
```

The statuses come from a `HashMap`, so they are sorted for a stable message, and an empty set
reports `no components were registered` rather than an empty list — that is a distinct
condition, and `is_ready` returns `false` for it.

The format arguments are only evaluated when the assertion fails, so a passing wait is unchanged.

## Scope

Refs #12396. This does **not** close it and does not make that workflow green — it makes the
daily failure state its cause, which is a prerequisite for fixing item 1a rather than a fix for
it.

While investigating I established that the issue's proposed item-1 fix (a `HAS_*_SECRET` guard
on `test-bedrock`) would be a **no-op**: the Bedrock credentials are present in the scheduled
run. The step's environment block renders both as `***`, which is the masking GitHub applies to
a registered secret with a non-empty value, and the runtime logs `all 2 secret reference(s)
resolved`. I have not opened that change, to avoid a diff that reads as a fix and moves nothing.
The full evidence is on the issue.

This is the same treatment #12730 received for the catalog readiness wait in #12731; that fix
landed only in `crates/runtime/tests/postgres/catalog_changes.rs`, so the shared helper still
asserted bare.

## Test plan

- [x] `make lint` (never a per-crate invocation)
- [ ] `make test`
- [x] Logic verified against a standalone probe reproducing the helper with stub types matching
      the verified signatures of `Runtime::status`, `RuntimeStatus::get_all_statuses` and
      `ComponentStatus`: empty set reports `no components were registered`, output is sorted,
      and an `Error` variant's message is surfaced.
- [x] `cargo fmt --all` clean.

Note: this crate cannot be compiled on the authoring machine (`crates/cayenne` does not build on
Windows, which aborts `runtime` with it), so the compile is verified by CI and the remote
sign-off rather than locally.

## Checklist

- [x] Comments describe why the code is the way it is, not how it used to work
- [x] No `.unwrap()` added
- [x] Change is test-only; no runtime behavior is affected

### DIFF
--- a/crates/runtime/tests/utils/mod.rs
+++ b/crates/runtime/tests/utils/mod.rs
@@ -56,7 +56,33 @@ pub(crate) async fn runtime_ready_check(rt: &Runtime) {
 }
 
 pub(crate) async fn runtime_ready_check_with_timeout(rt: &Runtime, duration: Duration) {
-    assert!(wait_until_true(duration, || async { rt.status().is_ready() }).await);
+    assert!(
+        wait_until_true(duration, || async { rt.status().is_ready() }).await,
+        "the runtime did not become ready within {duration:?} — {}",
+        describe_component_statuses(rt)
+    );
+}
+
+/// Name every registered component and the state it is in.
+///
+/// The readiness wait elapses identically whether a component failed to load, is
+/// still initializing, or was never registered at all, so on its own it says only
+/// that the runtime is not ready. `ComponentStatus::Error` carries the message that
+/// put the component in that state, so naming the states turns the timeout into the
+/// reason for it.
+fn describe_component_statuses(rt: &Runtime) -> String {
+    let statuses = rt.status().get_all_statuses();
+    if statuses.is_empty() {
+        return "no components were registered".to_string();
+    }
+
+    let mut described: Vec<String> = statuses
+        .iter()
+        .map(|(component, status)| format!("{component}: {status:?}"))
+        .collect();
+    // `get_all_statuses` returns a `HashMap`, so sort for a stable message.
+    described.sort();
+    described.join(", ")
 }
 
 pub(crate) async fn runtime_ready_check_with_timeout_err(


grokspice:
@copilot review

grokspice:
## The failed sign-off is not this branch

Sign-off run [31270608903](https://github.com/spiceai/spiceai/actions/runs/31270608903) posted `signoff=failure` on `7efa36ed`, but the single failure it reports is not reachable from this diff:

```
Summary [ 784.391s] 10064 tests run: 10063 passed (6 slow, 1 leaky), 1 failed, 23 skipped
FAIL [  40.299s] ( 3090/10064) cayenne::mutation_property_test prop_concurrent_mixed_dense_key_sqlite
```

This PR changes one file — `crates/runtime/tests/utils/mod.rs` — and touches no Cayenne code, no
mutation path, and nothing the property harness loads.

`prop_concurrent_mixed_dense_key_sqlite` is #12824: the maintained count is served `Exact(9)` while
11 rows are live, so the assertion fails whenever the delta is still queued when the count is read.
That is a trunk-level defect, not a branch one, and its fix is #12829.

## What this needs

A re-dispatched sign-off, once #12829 has landed — the failure is intermittent rather than
deterministic (the same test passed on the other sign-offs in today's window), so a re-run before
then can go either way and would only re-roll the same dice.

Not re-dispatching now on purpose: eight sign-off runs were in flight at the time of writing, and
adding one into a saturated pool is how a run comes back `cancelled` with no verdict at all
(#12710). Nothing else on this PR is blocked — the diff, the review threads, and every other check
are clean.

phillipleblanc (approved):

PR #12856, opened 2026-08-10
test(embeddings): truncate over-length inputs in the MiniLM embed spicepods

Jeadie:
## What
Set `truncate: END` on the two all-MiniLM-L6-v2 embedding fixtures:
- `test/spicepods/models/hf-embed.yaml` (huggingface provider)
- `test/spicepods/models/local-embed.yaml` (local/file provider)

## Why
all-MiniLM-L6-v2 caps inputs at 256 tokens. With no truncation the embedding call rejects any longer input (`Input validation error: inputs must have less than 256 tokens`). A dataset load that feeds it long text then fails, and the runtime never reaches ready — the search benchmark burns its full ready-wait (30 min) before failing. `truncate: END` keeps the start of an over-length input and embeds it instead of erroring.

These two fixtures are the canonical single-provider embed spicepods, one per provider that #12620 adds `truncate` to. The MTEB search benchmark spicepods (`hf[all-minilm-l6-v2]-arrow.yaml`) already carry `truncate: END` from #12237; this fills the remaining gap.

## Depends on
Uses the `truncate` param added in #12620 — **merge after #12620**. If it merged first, the param is simply ignored (unknown params do not error), so ordering is safe but the param only takes effect once #12620 lands.

## truncate values
`NONE` (default, reject), `END` (discard tail), `START` (discard head).


Jeadie:
Note: truncating over-length inputs to the model's max sequence length is the standard MTEB evaluation convention — the benchmark harness (and `sentence-transformers` itself) truncates rather than rejecting long documents, so `truncate: END` matches how these models are scored on the leaderboard. It keeps our results comparable rather than dropping the affected corpus rows.

phillipleblanc (approved):

PR #12897, opened 2026-08-11
fix(runtime): invalidate localpod children's cached results on parent refresh

### LINKED ISSUE #12887: Accelerated dataset refresh doesn't invalidate localpod children's cached results
### Describe the bug

When an accelerated dataset completes a refresh, its results-cache entries are invalidated — but entries for `localpod` child datasets built on top of it are not. Queries against the child keep being served from the cache with pre-refresh results until `item_ttl` expires.

After a localpod child's initial load, its refresher hands off to the parent's refresh task (synchronized refresh). On refresh completion the runtime calls `invalidate_for_table` only with the refreshing dataset's own name; the synchronized child dataset names (`sink.synchronized_tables()`) are never invalidated. The same applies to the CDC-changes and retention invalidation paths.

### To Reproduce

1. Dataset `a_backup` (any source, e.g. `file:`) with acceleration, plus `localpod:a_backup` as dataset `a` with acceleration; results cache enabled with a multi-minute `item_ttl`.
2. Run the same query once against `a_backup` and once against `a` (both now cached).
3. Change the source data and trigger an acceleration refresh; wait for both layers to reload.
4. Re-run both queries within `item_ttl`.

Expected: both return the post-refresh result.
Actual: `a_backup` returns fresh data (its cache entry was invalidated); `a` returns the stale pre-refresh result from cache. With `Cache-Control: no-cache` the child returns fresh data, confirming it's the cache entry, not the acceleration.

### Runtime version

Reproduced on trunk and v2.1.4.

### Notes

- `item_ttl` is currently the only freshness bound for queries against localpod children, so long TTLs silently serve stale data across refreshes.
- Likely fix shape: on refresh completion, invalidate each `synchronized_table.child_dataset_name()` alongside the parent (mirroring how #11532 handles index maintenance for synchronized children).

krinart:
Fixes #12887.

## Problem

When an accelerated dataset completed a refresh, its results-cache entries were invalidated — but entries for `localpod` child datasets built on top of it were not. After a child's initial load its refresher hands off to the parent's refresh task (synchronized refresh), and the refresh-completion handler called `invalidate_for_table` only with the refreshing dataset's own name. Queries against the child kept being served pre-refresh results from the cache until `item_ttl` expired, silently making the TTL the only freshness bound for localpod datasets — a long TTL served stale data across refreshes indefinitely.

## Fix

At refresh completion, invalidate the live set of dataset names — the refreshing dataset plus every synchronized child attached to the sink (`RefreshTask::get_dataset_names`, the same set already used for metrics labels and status updates). The set is resolved from the `RefreshTask` at invalidation time rather than captured when the refresher loop starts, because children attach only after their own initial loads complete.

- `refresh_task.rs`: expose `get_dataset_names` as `pub(crate)` with a doc comment on the contract.
- `refresh_task_runner.rs`: accessor handing the refresher loop its `Arc<RefreshTask>`.
- `refresh.rs`: iterate the full name set at the existing invalidation site.

The CDC-changes and retention invalidation paths are deliberately unchanged: synchronized children can only attach to full-refresh-mode datasets, and retention deletes mutate only the parent's accelerator (child accelerator contents are untouched, so child cache entries still match the child's data).

## Test

New integration test `test_localpod_refresh_invalidates_child_cached_results` in `acceleration/localpod_sync.rs`: caches a `COUNT(*)` for both parent and child (verifying `CacheHit` on repeat), changes the source CSV, drives a parent refresh, and asserts the child query observes the new count within a bounded poll — with a 10-minute `item_ttl` so TTL expiry cannot mask a missing invalidation. Before the fix the child serves the stale cached count and the poll exhausts; the pre-existing sync test still passes.

### DIFF
--- a/crates/runtime-table/src/accelerated/refresh.rs
+++ b/crates/runtime-table/src/accelerated/refresh.rs
@@ -916,6 +916,10 @@ impl Refresher {
         let mut refresh_task_runner = refresh_task_runner.build();
 
         let (start_refresh, mut on_refresh_complete) = refresh_task_runner.start()?;
+        // Handle for the refresh-completion handler below: cache invalidation
+        // must cover the live set of dataset names (self + synchronized
+        // children), which grows as children finish their initial loads.
+        let refresh_task = Arc::clone(refresh_task_runner.refresh_task());
         self.refresh_task_runner = Some(refresh_task_runner);
 
         let notifier = self.on_complete_notification.clone();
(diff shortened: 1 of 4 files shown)


sgrebnov (approved):

krinart:
I ran adversarial code review using Claude Code. No real issues found

Issue #12939, opened 2026-08-12
Segment cache gauges only publish every 200,000 accesses, leaving the cache unobservable on most deployments


bjchambers:
## Describe the bug

The Vortex segment cache is one of the larger memory consumers in a Cayenne deployment — sized per table at 256 MB to 1024 MB (see #12937) — but it is effectively unobservable until a table has accumulated 200,000 segment accesses.

All five of its gauges are published inside a sampling branch gated on `SEGMENT_CACHE_STATS_SAMPLE_EVERY = 200_000` (`crates/vortex/src/persistent/segment_cache.rs:26`, `:136-145`):

```rust
if accesses.is_multiple_of(SEGMENT_CACHE_STATS_SAMPLE_EVERY) {
    self.shared.cache.run_pending_tasks().await;
    ACCESSES.record(...); HITS.record(...);
    WEIGHTED_BYTES.record(...); CAPACITY_BYTES.record(...); ENTRIES.record(...);
}
```

There is no lower-frequency-safe alternative — no counter that publishes continuously, and no way to read live fill or hit rate below the threshold. A table whose traffic never reaches 200,000 segment accesses exposes **nothing at all** about a cache that may be holding a gigabyte.

## Measured

A probe with `cayenne_segment_cache_mb: 8` on a 228 MiB accelerated table, driven for **5,650 scans over 905 seconds**, published **no segment-cache gauges whatsoever**.

Control on the same instance confirms the metrics pipeline is healthy: 409 metric lines exposed, including many `cayenne_*` families (`cayenne_compaction_*`, `cayenne_encode_*`, `cayenne_mem_tier_*`, `cayenne_listing_*`). Only the segment-cache family is absent.

Note the two possible explanations cannot be distinguished from outside — either the access count stayed under 200,000, or the cache was not exercised at all — **because the only instrumentation for either is behind the same gate**. That is the core of the problem: there is no signal that distinguishes "cache is working" from "cache is idle" from "cache is full of dead entries".

## Why this matters now

The segment cache has two active defects, #12936 (no invalidation path, so it fills with unreachable segments) and #12938 (TinyLFU admission rejecting newly-read segments once full). Both were characterised by simulation precisely because in-situ measurement was not possible. Operators trying to determine whether they are affected have no way to check.

Hit rate and live fill are exactly the signals that would reveal both defects: a cache pinned at capacity with a collapsed hit rate is the signature.

## Expected behavior

[...]

PR #12942, opened 2026-08-12
fix(cayenne): make segment cache observable on low-traffic tables (#12939)

Jeadie:
## 📝 Summary

The Vortex segment cache is one of the larger memory consumers in a Cayenne deployment (sized 256 MB–1024 MB per table), but all five of its gauges were published once every 200,000 accesses. For low traffic tables, you cannot distinguish "cache working" from "cache idle" from "cache full of dead entries". 

This PR is a short term improvement where we record accesses and hits (low cost, independent of cache bookkeeping) at a higher cadence (every 256). 

Follow up PR will move these metrics to a set of observable instruments on the `cayenne_segment_cache` meter. Each then gets a `.with_callback(...)` that the SDK invokes at every Prometheus scrape. Would remove several checks/updates on the hot path. 

Improvement for https://github.com/spiceai/spiceai/issues/12939.

### DIFF
--- a/crates/vortex/src/persistent/segment_cache.rs
+++ b/crates/vortex/src/persistent/segment_cache.rs
@@ -21,9 +21,14 @@ use vortex::layout::segments::{SegmentCache, SegmentId};
 /// `SipHash` on the per-segment hot path.
 type SegmentCacheHasher = BuildHasherDefault<XxHash3_64>;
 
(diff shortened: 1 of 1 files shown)


Jeadie:
Using https://github.com/spiceai/spiceai/pull/12944 instead

Issue #12937, opened 2026-08-12
Segment cache is sized per-table from host RAM with no aggregate budget, so memory scales with table count


bjchambers:
## Describe the bug

The Vortex segment cache is allocated **per table**, sized from **whole-machine memory**, with no aggregate budget across tables.

`SharedSegmentCache` is constructed per `VortexFormat` (`crates/vortex/src/persistent/format.rs`), and Cayenne builds one format per table with a dataset label (`crates/cayenne/src/provider/context.rs:1040-1068`). Its size comes from `segment_cache_mb`, resolved in `crates/runtime/src/dataaccelerator/cayenne/autotune.rs`:

```rust
const FLOOR_MB: u64 = 256;
const CEIL_MB: u64 = 1024;
let scaled_mb = self.total_mem_bytes / 128 / MIB;
usize::try_from(scaled_mb.clamp(FLOOR_MB, CEIL_MB)).unwrap_or(256)
```

That is purely a function of host memory. **Nothing divides it across tables**, so the aggregate is `N tables x clamp(RAM/128, 256 MB, 1024 MB)`. On a 64 GiB host that is 512 MB per table — 8 tables target 4.1 GB, and even at the 256 MB floor, 16 tables reach 4.1 GB.

This is inconsistent with how Cayenne treats its other memory consumers, which *are* globally coordinated:

- the encode-concurrency budget "caps aggregate write-encode shards across all tables"
- the CDC in-memory tier budget is "coordinated with the query + compaction pools so their sum stays within host RAM"

The segment cache participates in neither. There is no aggregate accounting for it in `memory_budget.rs` or `provider/memory_account.rs`, and no global reservation API for it in the cayenne crate.

## A request is not a limit

`get_total_memory()` (`crates/runtime/src/resource_monitor.rs:97-103`) prefers the cgroup limit and falls back to host RAM:

```rust
get_container_memory_limit().unwrap_or_else(|| system.total_memory())
```

`parse_cgroup_v2_memory_max` returns `None` for `"max"`, i.e. unlimited. So a Kubernetes pod with `requests.memory` but **no `limits.memory`** has `memory.max = max`, and every table sizes its segment cache from the *node's* RAM rather than anything the pod was scheduled for. On a large node that pins each table at the 1024 MB ceiling.

[...]

Issue #12806, opened 2026-08-08
The Pingora cache invalidation scan runs on the calling Tokio worker, stalling the runtime for its whole O(n) walk


grokspice:
Split out of #12674, which reports three defects in one code path. This is **defect 2** only. #12674 stays open for defect 1 (staleness race, whose correctness half is being addressed write-side by #12673) and defect 3 (the scan resets LRU order).

Separating it because the scoping note on #12674 established that defect 2 is the one of the three that needs nothing from the others: fixing it requires no table references at insert time, so it does not depend on the `AsTableRefs` bound change that defects 1 and 3 both converge on.

## The defect

`TabledCacheProvider::invalidate_for_table` is a **synchronous** trait method, so the Pingora branch drives the backend with `futures::executor::block_on`:

https://github.com/spiceai/spiceai/blob/trunk/crates/cache/src/lru_cache.rs#L503-L529

That parks the calling thread for the whole O(n) scan — one `iter_keys()` plus a `get()` per key, plus a `remove()` per match. Every caller is an `async fn`, so the parked thread is a Tokio worker:

- `crates/runtime/src/accelerated_table/refresh.rs:1070` (refresh completion)
- `crates/runtime/src/accelerated_table/refresh_task/changes.rs:1749`, `:2091`
- `crates/runtime/src/accelerated_table/retention.rs:295`
- `crates/runtime/src/datafusion/mod.rs:2321`, `:2440` (write paths)
- `crates/runtime/src/datafusion/query.rs:1258`, `query/transaction.rs:219`
- `crates/runtime/src/flight/do_put.rs:235`, `flightsql/statement_update.rs:104`
- `crates/datafusion-optimizer-rules/src/logical_plan/cache_invalidation.rs:254`, inside the `stream!` body that wraps a write

This violates the "reach an `.await` at least every ~100µs" rule in the contributor guide, and the consequence named there is the one that matters: a stalled worker also stalls `/health`, which is what Kubernetes kills the pod on. The comment directly above the call says "Spawn a blocking task to handle the synchronous iteration" — no blocking task is spawned.

Reachable in released builds: `engine: pingora` is opt-in per-runtime, but the `pingora` feature ships in `SPICED_DATA_FEATURES`.

## Why making the method `async` is not on its own sufficient

[...]

PR #12943, opened 2026-08-12
fix(vortex): invalidate retired segment cache entries

### LINKED ISSUE #12936: The Vortex segment cache has no invalidation path, so every refresh and compaction leaves it holding unreachable segments
## Describe the bug

The Vortex segment cache has no invalidation path. Its builder (`crates/vortex/src/persistent/segment_cache.rs:91-100`) sets only `name`, `max_capacity` and `weigher` — there is no `time_to_live`, no `eviction_listener`, and critically no `support_invalidation_closures()`, so `invalidate_entries_if` **cannot be called on this cache even if a caller wanted to**. Nothing in `segment_cache.rs`, `format.rs` or `source.rs` calls `invalidate`, `clear` or `remove`.

The cache is keyed by `(Arc<Path>, SegmentId)`. Every Cayenne full refresh writes a new snapshot under a fresh UUIDv7 directory, so **every refresh orphans the entire cached working set**: those segments are keyed to paths that will never be read again. Compaction has the same effect — it rewrites files into new paths.

Observed directly in a 9-refresh run:

```
narrow/019ff413-aafa-…/019ff41a-156f-…/019ff41a1570…_p000_00000.vortex
narrow/019ff413-aafa-…/019ff419-2ac8-…/019ff4192ac9…_p000_00000.vortex
narrow/019ff413-aafa-…/019ff41b-000f-…/019ff41b0010…_p000_00000.vortex
```

Capacity-driven eviction is the only way these ever leave. That has three consequences:

1. **Every refresh cold-starts the cache.** With a short refresh interval it never amortises.
2. **Memory ratchets to capacity and stays there**, holding a mix of live and superseded segments — the cache never releases, so RSS does not come back down.
3. **An idle table never reclaims at all.** The cache is per-table and eviction needs capacity pressure *on that same cache*. A table that stops being queried generates none, so its segments from long-deleted files stay resident indefinitely with no TTL to age them out.

## Measurements

Simulated with moka pinned to the repo's version (`=0.12.15`) and the builder replicated exactly — same `max_capacity`, same weigher over buffer length, no TTL, no explicit eviction policy. A refresh is modelled by rotating the path component of the key, which is what Cayenne does.

**Working set of 40 segments (2.5 MiB) in a 128-segment (8 MiB) cache — it fits three times over**, 2 read passes between refreshes:

| Generation | Hit rate | Live segments resident | Entries |
|---|---|---|---|
| 0-2 (filling) | 50% | 40 / 40 | 40 → 120 |
| 3 (reaches capacity) | 50% | 40 | 128 |
| 5+ (steady state) | **31%** | **~0** | 128 |

Final state across 5 runs: **94%, 100%, 98%, 99%, 100% of the cache is unreachable dead segments**. In three of five runs, *zero* of the 128 resident entries belonged to the current generation.

So a cache that is three times larger than the working set ends up delivering almost no useful caching while consuming 100% of its budget.

## Expected behavior

Superseded segments should be dropped when the snapshot that produced them is replaced. A full refresh or a compaction knows exactly which paths it retired and could invalidate them; today it has no mechanism to do so, because the cache is not built with invalidation-closure support.

## Suggested fix

- Build the cache with `.support_invalidation_closures()` and invalidate by path prefix when a snapshot is superseded (refresh and compaction both have the retired paths in hand).
- Or scope the cache per snapshot so rotation drops the old generation wholesale.
- A TTL would bound the damage but not fix it — dead entries would still occupy the budget for the TTL window, and a live-but-idle table would still be penalised.

## Scope

This compounds with the admission behaviour filed separately: no invalidation means dead entries accumulate until the cache is full, and once full, TinyLFU admission rejects the live generation. The two together are what produce the 94-100% dead steady state above.

## Caveats

The steady-state figures come from a faithful simulation of the cache's construction and key rotation, not from instrumenting a running spiced — the real segment cache's gauges only publish every 200,000 accesses (`SEGMENT_CACHE_STATS_SAMPLE_EVERY`), which makes in-situ observation impractical on low-traffic deployments. The code facts (no TTL, no invalidation support, no caller clearing it, path rotation per refresh) are read directly from the source and from an actual run's on-disk layout.

phillipleblanc:
## Summary
- enable exact-path invalidation for cached Vortex segments
- invalidate only files selected by committed snapshot retirement and manifest reference guards
- cover overwrite, compaction, protected/in-flight/shared paths, rollback, and local path identity

## Tests
- `cargo test -p vortex-datafusion invalidates_exact_paths_only --lib`
- `cargo test -p cayenne invalidates_retired_segments_after_cleanup --lib`
- `cargo test -p cayenne cleanup_invalidates_only_committed_retired_snapshot_paths --lib`
- `cargo test -p cayenne delete_retired_snapshot_dir_refcounted_keeps_referenced_files --lib`
- `cargo test -p cayenne local_segment_cache_path_matches_relative_listing_path --lib`
- `cargo clippy -p vortex-datafusion --lib --tests -- -D warnings`
- `cargo clippy -p cayenne --lib --tests -- -D warnings`

Closes spiceai/spiceai#12936


crates/cayenne/src/provider/table.rs
```diff
@@ -4669,41 +4805,37 @@ impl CayenneTableProvider {
...
                             name,
                         ))
-                    });
+                    })
+            })
+            .collect();
+
+        self.invalidate_segment_cache_paths(
+            objects.iter().map(|meta| meta.location.clone()).collect(),
+        );
```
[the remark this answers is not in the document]
phillipleblanc: Added an `InMemory` object-store regression in 97b17102. It scans identical referenced and retiring Vortex objects into the shared segment cache, runs the production ref-counted cleanup, and verifies only the retiring object and its cache entries are removed. The test also fails when cache invalidation is temporarily disabled.

crates/vortex/src/persistent/segment_cache.rs
```diff
@@ -157,13 +279,34 @@ impl SegmentCache for PathSegmentCache {
...
+            return;
+        }
+        let mut states = path_states.lock();
+        if states
+            .get(self.path.as_ref())
+            .and_then(Weak::upgrade)
+            .is_some_and(|registered| Arc::ptr_eq(&registered, state))
+        {
+            states.remove(self.path.as_ref());
+        }
```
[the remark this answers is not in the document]
phillipleblanc: Fixed in 6f453b11. The registry mutex now covers both the registered-state identity check and the last-owner check, serializing it with `for_path` weak upgrades. Added `path_state_stays_registered_until_the_last_open_file_drops`, which proves two openers share one registered state, dropping one preserves it, retirement reaches the surviving opener, and only the final drop unregisters it.

Jeadie (dismissed):

crates/cayenne/src/provider/delete/sink/file_based.rs
```diff
@@ -285,15 +289,28 @@ impl FileBasedDeletionSink {
...
             }
         }
 
+        // Exact paths are invalidated only after DeleteObject confirms that the
+        // object is absent. Do this even when a later delete failed so every
+        // file already removed by this partial batch releases its cached
+        // segments immediately and cannot disappear from a later directory
+        // listing before ever being invalidated.
+        self.provider
+            .invalidate_segment_cache_paths(retired_cache_paths)
```
bjchambers, line 308: From an adversarial review:

It looks like we invalidate when files are deleted. This means a few things:

(1) there is a delay after a full refresh before we get around to deleting the files. This is probably OK.
(2) Non-local storage returns early. We don't delete the files and treat them as a "bounded leak". This means the files won't be deleted AND the segments won't be invalidated.

Probably worth a follow-up issue?

> The gap that matters. Because invalidation is bolted to deletion, anywhere deletion never happens, invalidation never happens. Concretely: compaction retires merged-away inputs via retire_snapshot_dirs + sweep_retired_snapshot_dirs (table.rs:18169-18170), and that sweep returns early for s3:// (table.rs:4407) — documented as "a bounded leak." The comment at table.rs:18162 states the fallback can't cover these: under CDC memory mode the current snapshot never rotates, so the uuid7-timestamp heuristic in trigger_old_snapshot_cleanup keeps every dir forever (the measured 406-dir/89G leak). So on an S3-backed table under CDC memory mode, compaction-retired segments stay cached indefinitely — the exact behavior #12936 reports, unfixed for that deployment shape. Codex found this independently.

bjchambers: >- [medium] S3 retired snapshots can remain permanently resident in the segment cache (crates/cayenne/src/provider/table.rs:4400-4408)
  Invalidation is performed only after physical deletion, but this retirement path explicitly returns without sweeping S3 snapshots. Consequently, files made unreachable by protected-snapshot compaction remain cached until a later rotation happens to delete them—or until ordinary capacity eviction. A workload repeatedly retiring S3 snapshots can therefore fill the bounded cache with unreachable segments and evict the live working set. This demonstrates that physical-delete sites are not a complete proxy for logical reachability.
  Recommendation: Drive cache retirement from the published manifest/snapshot transition. Record paths removed from the live manifest and invalidate them once their snapshot scan-reference count reaches zero, independently of when or whether physical GC runs.

crates/vortex/src/persistent/segment_cache.rs
```diff
@@ -105,18 +118,113 @@ impl SharedSegmentCache {
...
+                })
+        });
         Arc::new(PathSegmentCache {
             shared: self.clone(),
             path: Arc::new(path),
+            state,
         })
     }
+
+    pub(crate) async fn invalidate_paths(&self, paths: HashSet<Path>) {
```
bjchambers, line 139: There is a potential race condition here. An entry that is added to a path *after* invalidate paths is called. This could happen if there was a concurrent addition along the path (eg., insert concurrent with the invalidation). In such cases we would be unable to invalidate that path. One option would be to keep a set of invalidated paths for a few seconds and/or defer the invalidation cleanup until after we know the path is no longer reachable (so no new entries could be using it).

> - [medium] A newly opened cache handle can escape path retirement (crates/vortex/src/persistent/segment_cache.rs:150-168)
  `invalidate_paths` snapshots existing weak path states, releases the registry mutex, and only then marks those states retired. A concurrent `for_path` for an absent or expired registry entry can create a fresh unretired state after that snapshot; it is neither waited on nor marked retired, so it may insert after key enumeration completes. This is reachable in the repository's acknowledged plan-build-to-open races for retention and cold GC, allowing a physically deleted path to repopulate the cache after invalidation returns.
  Recommendation: Serialize state creation with retirement by installing a retired generation/tombstone under the registry mutex before releasing it. Keep that generation until reachability or scan-lifetime accounting proves no stale opener can appear, and add a test that creates `for_path` between retirement-state capture and key eviction.

bjchambers (approved):
Two findings from an adversarial review. One feels like it merits a follow-up issue and the other may be less of a concern (concurrent entries getting created racing with invalidation). Feel free to open a follow-up issue and/or dismiss the comments.

crates/cayenne/src/provider/delete/sink/file_based.rs
bjchambers: Filed as #12962 — the S3 case is confirmed in the code: `sweep_retired_snapshot_dirs` returns early for `s3://` (`table.rs:4406-4409`), that sweep is where this PR's retired-dir invalidation lives (`table.rs:4538`), and compaction routes its merged-away inputs through it (`table.rs:18170-18171`). The in-tree comment above that call already rules out `trigger_old_snapshot_cleanup` covering them under CDC memory mode, so on S3 + memory mode nothing invalidates compaction-retired segments.

The issue carries your recommendation (drive retirement from the manifest/snapshot transition and the scan-ref gate rather than from the delete sites) and notes the post-refresh grace delay as the acceptable secondary point.

Safe to resolve — nothing here blocks this PR, which strictly improves the local-storage case.

crates/vortex/src/persistent/segment_cache.rs
bjchambers: Filed as #12963. Confirmed the mechanism: `invalidate_paths` snapshots the registered `PathCacheState`s under the mutex and releases it before marking them retired (`segment_cache.rs:150-168`), while `for_path` mints a fresh `PathCacheState::default()` — `retired == false` — for any path with no registry entry or an expired `Weak` (`:120-136`). A handle created after the snapshot is neither marked nor waited on.

One note on your "less of a concern" read, which is why I filed it rather than dismissing it: the argument that a later opener fails to read depends on the unlink having happened. On S3 it often hasn't — `sweep_retired_snapshot_dirs` returns early there (#12962) — so a late reader can still read the object successfully and re-cache it. The two findings compound, and the tombstone-under-the-mutex fix wants the same scan-ref accounting #12962 needs, so they're probably one piece of work.

The issue also records the test to add: create `for_path` between the retirement-state capture and the key eviction, and assert the put doesn't land.

Safe to resolve.

bjchambers:
Both of @bjchambers's review findings are now filed as follow-ups, with replies on each thread so they can be resolved:

- #12962 — invalidation is driven from physical deletion, so an S3-backed table never invalidates compaction-retired segments (`sweep_retired_snapshot_dirs` returns early for `s3://`, and that sweep is where retired-dir invalidation lives). Confirmed against this branch.
- #12963 — a `for_path` handle created after the retirement snapshot gets an unretired `PathCacheState`, so a retired path can be repopulated. Reachability is higher than it first looks because of #12962: when the unlink is skipped, a late reader can still read the object.

One more, from a Copilot comment that was **suppressed** and so never became a thread here:

- #12964 — `invalidate_paths` walks the entire cache via `cache.iter()` with no yield points (O(cache entries), not O(retiring paths)), and the `active_puts` drain has no bounded wait. Same defect class as #12806, which we fixed for the Pingora cache. Severity is lower since the main caller is a detached background task, but it gets worse under #12937's consolidation, where the walk becomes O(entries across all tables).

None of the three block this PR — it strictly improves the local-storage case. I've also based the shared-cache work (#12937) on this branch, since consolidating per-table caches is only safe once retired segments are actually evicted.

bjchambers:
Added and removed from queue due to merge conflicts. I'm working on a follow-up to this which will combine the per-table segment caches. If that goes in first I'll close this, otherwise I'll try to fix merge conflicts and merge this first (once the stuff in front goes in).

phillipleblanc:
@bjchambers is working on a better version of this change

PR #12944, opened 2026-08-12
fix(vortex): expose segment cache metrics on scrape

### LINKED ISSUE #12939: Segment cache gauges only publish every 200,000 accesses, leaving the cache unobservable on most deployments

phillipleblanc:
## Summary

- export Vortex segment-cache accesses and hits as monotonic observable counters, and capacity, weighted bytes, and entries as observable gauges on every collection
- retain per-dataset counter continuity across cache recreation while weakly tracking live cache state so retired caches are not kept alive
- aggregate overlapping caches with the same dataset label and keep the get path to relaxed atomic increments without metric recording or label allocation
- construct Cayenne formats with their dataset label directly and avoid allocating read caches for write-only formats

Access and hit metrics retain their existing names but now have Prometheus counter type semantics; dashboards should use counter operations such as `rate()` or `increase()`.

## Validation

- `cargo test -p vortex-datafusion persistent::segment_cache::tests --lib`
- `cargo clippy -p vortex-datafusion --lib --tests --no-deps -- -D warnings`
- `cargo clippy -p cayenne --lib --no-deps -- -D warnings`
- `make lint-rust`

Closes spiceai/spiceai#12939


crates/vortex/src/persistent/segment_cache.rs
```diff
@@ -21,52 +24,246 @@ use vortex::layout::segments::{SegmentCache, SegmentId};
...
+}
+
+impl SegmentCacheMetrics {
+    fn new(meter: &Meter) -> Self {
+        let datasets = Arc::new(RwLock::new(Vec::<Weak<SegmentCacheDatasetMetrics>>::new()));
+        let counters = Arc::new(RwLock::new(HashMap::new()));
+
+        // Observe hits before accesses. `get()` increments accesses first, so a
+        // concurrent collection can under-report hits briefly but cannot make a
+        // completed hit appear without its corresponding access.
```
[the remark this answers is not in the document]
phillipleblanc: Fixed in d329eaa3. The observable callbacks now synchronize per dataset instead of relying on registration order: the access callback holds a scrape-only mutex through observer.observe(), then publishes that completed access snapshot; the hit callback takes the same mutex and caps its observation to the completed access snapshot. This guarantees exported hits cannot exceed an access value already delivered to every reader, while the cache get() hot path remains unchanged at two relaxed atomic increments. Added counter_callbacks_cannot_publish_hits_ahead_of_accesses; it covers reversed callback order, and temporarily removing the cap makes the test fail.

crates/vortex/src/persistent/segment_cache.rs
```diff
@@ -159,8 +375,248 @@ impl SegmentCache for PathSegmentCache {
...
+                            metric.as_str(),
+                            "cayenne_segment_cache_accesses" | "cayenne_segment_cache_hits"
+                        )
+                }),
+                "retired dataset {dataset} must not report live cache gauges"
+            );
+        }
+    }
+
+    async fn settle_cache_bookkeeping(cache: &SharedSegmentCache) {
```
Jeadie, line 616: We dont run this in non-test code? This means caches never resolve their approximations. Hard to do in synchronous callbacks (i.e. `with_callback(move |observer| …)`.

phillipleblanc: This is done by Moka lazily (i.e. get/write paths that run maintenance tasks). Having this here makes it deterministic to test.

Jeadie (approved):

bjchambers (approved):

lukekim (approved):

PR #12993, opened 2026-08-13
refactor: narrow the DataConnector trait off the runtime's types

bjchambers:
Narrows the `DataConnector` trait so its signatures name fewer types that live at or above the `runtime` crate. Everything stays inside its current crate — no crate moves, no tier changes — so the layering graph is untouched and this is reviewable on correctness alone.

Three couplings are removed.

**Accelerated-table hooks.** `on_accelerator_setup` and `on_accelerated_table_registration` took the accelerated-table types themselves (`accelerated::Builder`, `AcceleratedTable`). Those live beside the runtime that orchestrates them, so a connector naming either depended on the orchestrator for the sake of two getters and a `Vec::push`. Each hook now takes a capability trait holding the slice its implementations actually reach for — `AcceleratorSetup` (read the accelerator's provider, replace it with a wrapper) and `RegisteredAcceleratedTable` (request a refresh, attach a task whose lifetime follows the table). Only two of the eleven implementations do real work: the embedding connector's DuckDB HNSW wrapping and the file connector's watcher. A refresh request goes through `RefreshRequester::request_refresh` rather than a send on the raw trigger channel, which keeps `RefreshOverrides` off the contract, and attaching a task goes through `attach_task` rather than pushing onto the public `handlers` field.

**Change streams.** `changes_stream` and `append_stream` took an `Arc<FederatedTable>`. All eleven implementations want the same thing of it — the `TableProvider`, so they can recover their own concrete provider type and start reading. Both now take `Arc<dyn FederatedTableProvider>`, a two-method trait that `FederatedTable` implements: `table_provider` resolves asynchronously, because a deferred dataset registers before its source is contacted, and `try_table_provider_sync` covers the synchronous peek the embedding and full-text wrappers make before delegating inward.

**The runtime handle.** `ConnectorContext::runtime() -> Arc<Runtime>` handed connectors the whole orchestrator. Six connector call sites used it, and between them they wanted three things: the HTTP rate-control registry, the token-provider registry, and the runtime's own DataFusion session. Each is now its own method over a type below `runtime`. The sink connector was the only caller that wanted a real `Arc<Runtime>` — it rebuilt a whole `Dataset` from the context to read the accelerator's checkpoint schema — so the capability is now the result, `accelerated_checkpoint_schema(&DatasetSpec) -> Option<SchemaRef>`, and the rebinding happens inside `RuntimeConnectorContext` where those handles belong.

`ConnectorContext::app()` is unchanged: the `app` crate already sits below `runtime`, so `Arc<App>` was never part of the problem. All four `.app()` callers read the spicepod `runtime:` block, one of them through the typed `runtime.flight` config rather than a named param.

`ListingTableConnector::get_runtime() -> Option<Runtime>` goes with the handle, as `get_app()`. Its one consumer chain resolved an entire `Arc<Runtime>` down to a single spicepod string — `runtime.params.parquet_page_index` — which four connector structs were carrying a runtime handle to reach.

### What is deliberately not here

Retyping the trait's `&Dataset` parameters to `&DatasetSpec` was expected to be mechanical, on the theory that existing callers keep compiling through deref coercion. That holds for the configuration-only methods, but not for the three that hand back a provider: `read_provider`, `read_write_provider` and `changes_stream` reach the accelerator's `spice_sys` state — the CDC checkpoint stores in kafka, mysql, postgres and mongodb — which needs the runtime-bound `Dataset`. Narrowing those means giving those stores a below-runtime interface, the way `runtime-checkpoint-api` already does for DynamoDB's blob checkpoints. That is its own change with CDC-state correctness implications, so the retype waits for it; the module docs on `RefreshSource` now record this as the remaining blocker rather than predicting an easier path.

### Verification

Behavior-preserving throughout. Full-workspace `cargo check` green under the release feature set; the crate-layering guard passes (142 crates, no upward edges); `Cargo.lock` regenerated and committed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5998,11 +5998,14 @@ dependencies = [
 name = "data-connector-api"
 version = "2.2.0-unstable"
 dependencies = [
+ "async-trait",
  "data_components",
+ "datafusion",
  "globset",
  "runtime-component",
  "serde_json",
  "snafu",
+ "tokio",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4856,6 +4856,7 @@ dependencies = [
 name = "connector-sharepoint"
 version = "2.2.0-unstable"
 dependencies = [
+ "app",
  "arrow",
  "async-stream",
  "async-trait",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4630,6 +4630,7 @@ dependencies = [
  "arrow-schema",
  "async-stream",
  "async-trait",
+ "data-connector-api",
  "data_components",
  "dataformat-json",
  "datafusion",
@@ -4701,6 +4702,7 @@ dependencies = [
  "async-stream",
  "async-trait",
  "chrono",
+ "data-connector-api",
  "data_components",
  "datafusion",
  "datafusion-table-providers",
@@ -4784,6 +4786,7 @@ version = "2.2.0-unstable"
 dependencies = [
  "async-stream",
  "async-trait",
+ "data-connector-api",
  "data_components",
  "datafusion",
  "datafusion-table-providers",
--- a/crates/data-connector-api/Cargo.toml
+++ b/crates/data-connector-api/Cargo.toml
@@ -10,11 +10,14 @@ rust-version.workspace = true
 version.workspace = true
 
 [dependencies]
+async-trait.workspace = true
 data_components = { path = "../data_components" }
+datafusion.workspace = true
 globset = { workspace = true }
 runtime-component = { path = "../runtime-component" }
 serde_json.workspace = true
 snafu.workspace = true
+tokio.workspace = true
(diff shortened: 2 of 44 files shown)


bjchambers:
Thanks — all three addressed.

**The test breakage was real and I had missed it.** `cargo check` skips `#[cfg(test)]` code, so the five `parquet_page_index_options` tests were invisible to my verification. They no longer need a `Runtime` at all: they build the app directly and are now synchronous, which is the point of the narrowed signature. Fixed in f1ea29da34, verified with `--all-targets`.

Same commit drops two `FederatedTable` imports that `cdc_ingest` and `debezium` no longer use — a warning to `cargo check`, an error to the lint gate's `-Dwarnings`.

Both doc comments rewritten to state the current rationale rather than narrate the previous signature.

lukekim (approved):

PR #13012, opened 2026-08-13
fix(testoperator benchmarks): benchmark fixes

krinart:
- Bumps the pinned `spiceai/datafusion` fork to https://github.com/spiceai/datafusion/commit/1c43ad804bbf3f2e9c22e9b5700bdf8ebd957593
- TPC-DS Postgres seeding now uses the DuckDB-based fleet generator instead of tpcds-kit's `dsdgen`, whose RNG diverges from the generator that produced the recorded expected answers.
- Self-healing benchmark data: the workflow now creates any missing `*_accelerated` Postgres databases it finds in the dispatched spicepod before `spiced` starts, and Clickbench's `s3[parquet]-postgres` run gets its own local Postgres service container instead of the shared hosted cluster (it only needs an empty acceleration target, not pre-loaded source data).
- Snapshot hygiene: `CayenneAccelerationExec`'s `snapshots_scanned=`/`files_scanned=` counters are redacted (`<N>`) in explain-plan snapshots — they reflect compaction/ingestion state, not plan structure, and were breaking snapshots on every file-count drift.

### DIFF
--- a/.github/workflows/testoperator_run_bench.yml
+++ b/.github/workflows/testoperator_run_bench.yml
@@ -336,16 +336,9 @@ jobs:
           fi
           echo "Seed complete: $DB_NAME on $PGHOST."
 
-      # Postgres-accelerator spicepods (file[parquet]-postgres, s3[parquet]-postgres,
-      # ...) write into a dedicated database on POSTGRES_RW_HOST (pg_db:
-      # *_accelerated). The accelerator creates its own tables but not the database
-      # itself, and the local postgres:16 service container starts empty - so every
-      # *_accelerated database the spicepod names must be created before spiced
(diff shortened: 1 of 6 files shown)


sgrebnov (dismissed):

.github/workflows/testoperator_run_bench.yml
```diff
@@ -336,6 +336,41 @@ jobs:
           fi
           echo "Seed complete: $DB_NAME on $PGHOST."
 
+      # Postgres-accelerator spicepods (file[parquet]-postgres, s3[parquet]-postgres,
```
sgrebnov: Might be too verbose

sgrebnov (approved):

Issue #13015, opened 2026-08-13
ClickHouse connector returns an empty schema for a zero-row result, dropping the projected columns


claudespice:
## Describe the bug

The ClickHouse connector's `query_arrow` discards the projected schema it is handed and, when the query returns no blocks, returns a stream whose schema is `Schema::empty()`. The plan above it was built from the table's real schema, so an empty ClickHouse result yields a stream that does not match the schema its `ExecutionPlan` advertises.

`crates/data-connectors/connector-clickhouse/src/conn.rs:183-196`:

```rust
async fn query_arrow(
    &self,
    sql: &str,
    _: &[&'a dyn Sync],
    _projected_schema: Option<SchemaRef>,   // <-- discarded
) -> Result<SendableRecordBatchStream, ...> {
    let conn = self.pool.get_handle().await.context(ConnectionPoolSnafu)?;
    let mut block_stream = conn.query_owned(sql).stream_blocks();
    let first_block = block_stream.next().await;
    if first_block.is_none() {
        return Ok(Box::pin(RecordBatchStreamAdapter::new(
            Arc::new(Schema::empty()),          // <-- should be the projected schema
            stream::empty(),
        )));
    }
```

The schema is only ever derived from the *first returned block*, so with zero blocks there is nothing to derive it from — and the parameter that exists precisely to answer that case is ignored.

## Expected behavior

A query matching zero rows returns an empty stream carrying the projected schema, so the result is an empty table with the right columns.

## Impact

The wrong-result shape is "columns disappear when the result happens to be empty", and it is data-dependent: the same query is correct on a non-empty result and wrong on an empty one. Downstream operators either fail with a schema mismatch or see a column set that does not match the plan.

## Related

This class is already recognised in the codebase — `crates/cache/src/result/query.rs:448` carries a regression test for it:

> Empty query results must preserve the correct schema, not `Schema::empty()`.

filed as #9481. The same shape was just fixed in the GraphQL connector for an empty first page (#13004): an empty result must keep the configured/projected schema rather than narrowing to an empty one.

Sibling connectors that avoid the trap are worth reading as precedent — `connector-cosmosdb` and `connector-sharepoint` both thread `projected_schema` through to the stream.

## Additional context

Found by an altitude review while fixing #13004; verified by reading `conn.rs` on `trunk`, not reproduced against a live ClickHouse instance. Reproducing needs a ClickHouse-backed dataset and a query with a `WHERE` clause matching no rows.

[...]

PR #13016, opened 2026-08-13
fix(graphql): keep the configured schema when a page has no rows (fixes #13004)

### LINKED ISSUE #13004: GitHub GraphQL preflight crashes on an empty page: empty schema + transform cannot build a RecordBatch
## Summary

The GitHub connector's GraphQL preflight (`GraphQLTableProviderBuilder::build`) fails the dataset when the first page has **no rows**, even when the table has a configured schema (issues, pulls, …).

The operator sees:

```
WARN runtime::init::dataset: Failed to connect to the source for dataset <name> (github).
Failed to process GraphQL response: Invalid argument error: must either specify a row count or at least one column
```

The acceleration can still serve previously loaded data; the source connection stays in retry.

This is not a missing-repo / auth error. GitHub returned a 200 whose extracted `nodes` (or pointer) is empty or null. The client then reports `Schema::empty()`, and the GitHub schema transform cannot build a `RecordBatch`.

## Version

Reproduced 2026-08-13 on Spice Cloud (`datafusion_table_providers.issues` from `github:github.com/spiceai/datafusion-table-providers/issues`). Code on `trunk` still has the path below.

## Setup

```yaml
datasets:
  - from: github:github.com/spiceai/datafusion-table-providers/issues
    name: datafusion_table_providers.issues
    params:
      github_client_id: ${secrets:GITHUB_CLIENT_ID}
      github_private_key: ${secrets:GITHUB_PRIVATE_KEY}
      github_installation_id: ${secrets:GITHUB_INSTALLATION_ID}
    acceleration:
      enabled: true
      engine: duckdb
      mode: file
```

Any GitHub GraphQL table with a configured schema and `with_schema_transform(github_gql_raw_schema_cast)` will hit this when the first extracted page is empty (empty `nodes`, null pointer payload, pagination loop-guard, …).

## Expected

An empty first page is a valid result:

- Use the **configured** GitHub table schema (`IssuesTableArgs::gql_schema`, pulls, …).
- Register the dataset as ready with zero rows.
- Subsequent refreshes keep working.

A repo with no issues, or a page that happens to be empty, must not fail `create_gql_table_provider`.

## Actual

1. `GraphQLTableProviderBuilder::build` calls `client.execute(..., schema: None, ...)`.
2. Empty/null extracted data returns early with `Schema::empty()` and **does not** fall back to `GraphQLClient::configured_schema()`:

```rust
// connector-graphql/.../client.rs — null / empty unwrapped paths
schema: schema.unwrap_or_else(|| Arc::new(arrow::datatypes::Schema::empty())),
```

3. `derive_table_schema` then runs the GitHub transform on that empty schema:

```rust
let empty_batch = RecordBatch::try_new(Arc::clone(gql_schema), empty_columns)?;
```

Arrow rejects a 0-column `try_new` without a row count: `must either specify a row count or at least one column`.

4. The error is `Error::ArrowInternal`, which is **not** `is_retriable_error` and **not** `RateLimited`, so the GitHub fallback (`build_without_validation` with the configured schema) never runs. The dataset fails to connect.

`get_json_schema` *would* honor `client.schema` if the empty page reached it. The early returns skip that.

## Proposed fix

1. Empty/null GraphQL pages should use `self.schema` (configured schema) when present, not `Schema::empty()`.
2. `derive_table_schema` should use `RecordBatch::new_empty` (or skip the transform) when the schema has no fields, so a leftover empty schema cannot crash preflight.
3. Treat this class of preflight failure as recoverable when a configured schema exists (same as the existing rate-limit fallback).

## Workaround

None in spicepod. The dataset keeps serving the last acceleration while the source retry loop hits the same preflight error.

## Related

- `crates/data-connectors/connector-graphql/src/graphql/client.rs` — empty/null page returns `Schema::empty()`
- `crates/data-connectors/connector-graphql/src/graphql/provider.rs` — `derive_table_schema`
- `crates/data-connectors/connector-github/src/lib.rs` — `github_gql_raw_schema_cast`, fallback only for rate-limit / retriable
- `crates/data-connectors/connector-github/src/issues.rs` — `gql_schema()` is configured but unused on an empty first page

claudespice:
## Summary

A GraphQL dataset whose first page came back with **no rows** failed to connect, even when the table had a configured schema:

```
WARN runtime::init::dataset: Failed to connect to the source for dataset <name> (github).
Failed to process GraphQL response: Invalid argument error: must either specify a row count or at least one column
```

Root cause is two defects on the same path.

**1. The empty-page early returns discarded the configured schema.** `GraphQLClient`'s page handler has three early exits — a null payload, a page with no rows, and the pagination loop-guard — and each built its result as `schema.unwrap_or_else(|| Schema::empty())`. That consults only the per-query override, so a client *configured* with a schema (every GitHub table: `issues`, `pulls`, …) reported `Schema::empty()` for an empty page. The non-empty path never had this bug: it goes through `get_json_schema`, which walks override → configured → inference. The three early returns simply never called it. This is what makes the reported dataset fail.

**2. A fieldless schema cannot build a transform's probe batch.** With the empty schema in hand, preflight then calls `derive_table_schema`, which built a zero-row probe batch with `RecordBatch::try_new`. A schema with no fields has no column to carry the row count, so Arrow rejects it — `must either specify a row count or at least one column`. That surfaces as `Error::ArrowInternal`, which is neither retriable nor `RateLimited`, so the GitHub connector's existing configured-schema fallback (`build_without_validation`) never ran and the dataset failed outright.

A repo with no issues, or any page that simply happens to be empty, is a valid result — it should register as a ready table with zero rows.

## Changes

- `client.rs`: added `GraphQLClient::empty_page`, and routed all three early exits through it. The schema a rowless page reports is the one rule those exits must agree on, so it now lives in exactly one place — a fourth early exit added later cannot reintroduce this bug. It resolves the schema with `get_json_schema`, the same function and precedence the non-empty path uses; an empty row slice skips inference, giving override → configured → empty.
- `client.rs`: split the response handling out of `execute_inner` into `process_response`. This is what makes the fix testable — the crate has no HTTP mock and the defect lives entirely past the request.
- `provider.rs`: `derive_table_schema` builds its probe batch with `RecordBatch::new_empty`, which is infallible and identical for a schema that does have fields (it maps `new_empty_array` over them exactly as the removed code did).

On scope, two deliberate calls:

- Change 3 is **defense in depth, not the load-bearing fix**. Once the client keeps the configured schema, the fieldless-schema path is not reachable for any connector that installs a transform — `with_schema_transform` has exactly one production caller (`connector-github`), and every GitHub table passes `Some(gql_schema())`. It is still worth making `derive_table_schema` total rather than partial, which is what the issue asked for.
- The issue also proposed widening the GitHub connector's error-recovery fallback to treat this failure class as recoverable. Deliberately **not** included: with the two defects above fixed, preflight succeeds on an empty page rather than needing to be recovered from, and widening the fallback would change error handling for failures unrelated to this bug.

## Test plan

- `make lint`
- `make nextest`

New unit tests, all against the production functions. Each regression test was verified by reverting its fix and confirming it fails — and with the client fix reverted, exactly the three empty-page tests fail while the three back-compat/precedence pins keep passing:

- `empty_page_keeps_the_configured_schema` — the reported failure: an empty first page keeps the configured schema instead of `Schema::empty()`.
- `null_payload_keeps_the_configured_schema` — `nodes: null`, the other empty-page shape GitHub returns.
- `repeated_cursor_keeps_the_configured_schema` — the pagination loop-guard exit, the third site.
- `schema_override_outranks_the_configured_schema_on_an_empty_page` — precedence now matches the non-empty path.
- `empty_page_without_a_configured_schema_stays_empty` — back-compat: with nothing configured there is still nothing to infer, so the schema stays empty.
- `non_empty_page_still_parses_with_the_configured_schema` — the unchanged path, pinned.
- `derive_table_schema_handles_a_fieldless_schema` — reproduces the exact Arrow error from the report when the fix is reverted.
- `derive_table_schema_passes_configured_fields_to_the_transform` — the probe batch still carries every configured field.

## Review gate

- Adversarial review: `codex` — job `review-msru04s5-i75xz8`, verdict `needs-attention`, 1 finding. **Valid, fixed:** `process_response` took `schema`, `cursor`, and `response` by value while only borrowing them, which `clippy::pedantic`'s `needless_pass_by_value` would have failed the lint gate on. Signature now takes `Option<&SchemaRef>`, `Option<&str>`, `&serde_json::Value`.
- `/security-review`: no findings. The diff adds no log lines, no error strings carrying response content, no new dependencies, no `unsafe`, and no new network-reachable surface; the schema values it newly propagates come from spicepod config, not from the remote response.
- `/simplify`: applied — collapsed the three duplicated empty-result literals into `empty_page` (the altitude fix above), corrected a doc paragraph on `get_json_schema` that claimed emptiness is what preserves the configured schema when the configured schema in fact wins regardless, de-duplicated a rationale comment against its test, and made the test helper borrow rather than take ownership. Skipped: converting four of the tests to a table (distinct named tests are what made the revert-verification legible, and four cases do not warrant it) and sharing a test client-builder with `provider.rs`'s existing tests (would rewrite three tests outside this diff).

Its altitude pass also found the same defect shape in code this diff does not touch — the ClickHouse connector discards its `projected_schema` and returns `Schema::empty()` for a zero-row result. Filed as #13015.

Fixes #13004

## Checklist

- [ ] Tests pass locally (`make signoff`)
- [ ] Lint is green
- [ ] Issue linked

### DIFF
--- a/crates/data-connectors/connector-graphql/src/graphql/client.rs
+++ b/crates/data-connectors/connector-graphql/src/graphql/client.rs
@@ -956,7 +956,6 @@ impl GraphQLClient {
         .await
     }
 
-    #[expect(clippy::too_many_lines)]
     #[expect(clippy::too_many_arguments)]
     async fn execute_inner(
         &self,
--- a/crates/data-connectors/connector-graphql/src/graphql/client.rs
+++ b/crates/data-connectors/connector-graphql/src/graphql/client.rs
@@ -1119,6 +1119,39 @@ impl GraphQLClient {
             .map(|p| p(&response_headers, &response))
             .transpose()?;
 
+        self.process_response(query, schema.as_ref(), limit, cursor.as_deref(), &response)
+    }
+
+    /// The result for a page that yielded no rows, keeping whichever schema the table is
+    /// configured with so an empty page cannot narrow it.
+    ///
+    /// Every early exit out of [`Self::process_response`] goes through here: the schema a rowless
+    /// page reports is the one rule they must agree on, so it lives in one place.
+    fn empty_page(
+        &self,
+        schema: Option<&SchemaRef>,
+        cursor: Option<String>,
+    ) -> Result<GraphQLQueryResult> {
+        Ok(GraphQLQueryResult {
+            records: vec![],
+            limit_reached: false,
+            // No rows to infer from, so this resolves to the override or the configured schema.
+            schema: get_json_schema(self.schema.as_ref(), schema, &[])?,
+            cursor,
+        })
+    }
+
+    /// Turn a decoded GraphQL response body into record batches, resolving the schema each page is
+    /// parsed with. Split out from `execute_inner` so the page-shape handling — null payload, empty
+    /// page, repeated cursor — is reachable without an HTTP round trip.
+    fn process_response(
+        &self,
+        query: &GraphQLQuery,
+        schema: Option<&SchemaRef>,
+        limit: Option<usize>,
+        cursor: Option<&str>,
+        response: &serde_json::Value,
+    ) -> Result<GraphQLQueryResult> {
         let json_pointer = query
             .json_pointer
             .as_ref()
@@ -1141,7 +1174,7 @@ impl GraphQLClient {
                 } else {
                     format!("Invalid JSON pointer: '{json_pointer}'. The expected data path was not found in the response.")
                 };
-                tracing::error!("Failed to extract data from response. Full response: {}", serde_json::to_string_pretty(&response).unwrap_or_else(|_| format!("{response:?}")));
+                tracing::error!("Failed to extract data from response. Full response: {}", serde_json::to_string_pretty(response).unwrap_or_else(|_| format!("{response:?}")));
                 Error::InvalidJsonPointer {
                     pointer: error_msg,
                 }
@@ -1151,18 +1184,13 @@ impl GraphQLClient {
         // Handle null data explicitly
         if extracted_data.is_null() {
             tracing::debug!("Extracted data at pointer '{json_pointer}' is null");
-            return Ok(GraphQLQueryResult {
-                records: vec![],
-                limit_reached: false,
-                schema: schema.unwrap_or_else(|| Arc::new(arrow::datatypes::Schema::empty())),
-                cursor: None,
-            });
+            return self.empty_page(schema, None);
         }
 
         let next_cursor = query
             .pagination_parameters
             .as_ref()
-            .and_then(|x| x.get_next_cursor_from_response(&response));
+            .and_then(|x| x.get_next_cursor_from_response(response));
 
         // Validate next cursor if present
         if let Some(ref next_cursor_val) = next_cursor {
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

Issue #13090, opened 2026-08-14
vector_search rescans the broadcast query literal once per row, doubling the cost of the non-finite input screen


claudespice:
## Describe the bug

`compute_fsl_f32` (`crates/runtime-datafusion-udfs/src/vector_simd.rs`) screens both of its input slices for non-finite elements on every row (the #11263 fix). On the primary `vector_search` path one of those sides is a constant.

`crates/runtime/src/embeddings/udtf.rs:885` builds the query side as `lit(ScalarValue::FixedSizeList(...))` and passes it as **arg 0**, so `slice_a` in the row loop is the same `dim` floats repeated N times. The screen therefore recomputes a constant N times — exactly half the work the guard adds.

## Impact

Measured on the cosine path, 10000 rows x 1536 dims (`cargo bench -p runtime-datafusion-udfs --bench cosine_distance`): trunk 4.72 ms, with the screen 7.65 ms — **+62%**. Roughly half of that added cost is rescanning the same query vector.

## Expected behaviour

Screen the query side once. `cosine_distance::invoke_with_args` still holds `ColumnarValue::Scalar` before `make_scalar_function` broadcasts it, so the check can happen there and a per-side "already screened" flag threaded into `compute_fsl_f32`. Cost drops from `N*dim` to `dim`.

Two things fall out of the same hook:

- A non-finite **query** vector makes every row NULL, so the whole batch can short-circuit to a NULL array without touching the column.
- Larger and pre-existing: `make_scalar_function` broadcasts that literal via `ScalarValue::list_to_array_of_size`, materializing N x dim floats — roughly 50 MB of alloc+write per 8192-row batch at dim=1536, which both the guard and the kernel then stream back. Handling the scalar side without materializing removes that entirely, independent of the guard.

## Measured alternatives, so they are not retried

The screen's formulation was already tuned; these were measured on the same benchmark and rejected:

| form | cost vs trunk |
|---|---|
| short-circuiting `iter().all(f32::is_finite)` | +296% (blocks auto-vectorization) |
| float `acc += v * 0.0` accumulator | +600% (serial FP dependency chain) |
| 8-lane manual split | +110% |
| single-accumulator exponent mask (**shipped**) | +62% |
| scan both whole buffers once up front | +60% (buffers far exceed cache) |

[...]

Issue #13089, opened 2026-08-14
Embedding write paths accept a partially non-finite vector: the zero-vector filters test all() instead of any(), and DuckDB's validate_vector is never called on write


claudespice:
## Describe the bug

"Invalid embedding" has four different definitions across the write paths, and three of them let a vector like `[NaN, 2.0, 3.0]` through.

| Site | Check | Gap |
|---|---|---|
| `crates/search/src/index/s3_vectors/write.rs:292` (`filter_zero_vectors`) | `embedding.iter().all(\|&x\| x == 0.0 \|\| x.is_nan())` | `all`, not `any` — a vector with one `NaN` among real values is kept and written |
| `crates/search/src/index/memory/mod.rs:211` | `!vector.iter().all(\|&v\| v == 0.0 \|\| v.is_nan())` | same `all` gap |
| `crates/search/src/index/duckdb/mod.rs:355` (`validate_vector`) | correct `any(\|v\| !v.is_finite())` rejection | only ever called on the **query** vector (`duckdb/query_exec.rs:64`); the HNSW **write** path (`duckdb/write.rs:158`) never calls it |
| `crates/search/src/index/elasticsearch/write.rs:291,299` | rejects all-zero-or-NaN, then rejects any non-finite element | the only complete check; Elasticsearch only |
| `crates/search/src/index/native_vector.rs:109,145` (Cayenne) | none — documented pass-throughs | no validation at all |

Neither `crates/runtime/src/embeddings/` nor `crates/llms/` performs any finite-value validation: `Embed::embed` (`crates/llms/src/embeddings/mod.rs:160`) returns provider floats untouched, and `EmbeddingTask::embed` caches the result, so one bad provider response is retained and re-served.

The three `create_embedding_array` implementations (`search/src/index/write_util.rs:289`, `duckdb/write.rs:158`, `elasticsearch/write.rs:447`) are near-duplicates that all validate dimension only.

## Impact

A corrupt vector is accepted at write, indexed, and only screened at read — and for the memory / S3 Vectors backends the score never passes through the fixed UDF at all, so nothing screens it.

## Expected behaviour

One shared "is this a usable embedding" predicate that both ends call, tightened to per-element and including infinities. `write_util.rs:289` is the natural insertion point for the generated path (the DuckDB and Elasticsearch copies are drop-in duplicates differing only in error type), which would fix the `all`-vs-`any` gaps and the never-called `validate_vector` in passing.

[...]

Issue #13088, opened 2026-08-14
Federated cosine_distance does not inherit the NULL-for-non-finite contract, so the same corrupt embedding scores differently per engine


claudespice:
## Describe the bug

The vector-distance kernels in `crates/runtime-datafusion-udfs` return NULL for a vector carrying `NaN` or an infinity — there is no defined distance, and a fabricated one ranks a failed embedding against real matches (that is the fix for #11263).

That contract covers the local kernels only. `cosine_distance` is deliberately kept out of the DuckDB federation deny-list (`crates/data-connectors/connector-duckdb/src/lib.rs:80-90`) so it pushes down, where it unparses to `array_cosine_distance` (`crates/runtime-datafusion/src/dialect/duckdb.rs:121-125`). `inner_product` unparses to `array_inner_product` the same way.

So for a DuckDB-accelerated vector dataset — a common configuration — the patched Rust kernel never runs, and the engine's own non-finite semantics decide the answer.

## Impact

Same query, same data, two answers depending only on where the table lives: NULL when it executes locally, an engine-defined number when it federates. This is the *same shape* as the divergence #11263 was filed about, one layer out — the local/federated split is what makes it a correctness problem rather than a formula preference.

Not yet established, and the first thing to check: what DuckDB's `array_cosine_distance` actually returns for a `NaN`-bearing input. If it returns NULL the contract already holds by accident and this reduces to a test; if it returns a number, the divergence is live.

## Options

1. Deny `cosine_distance` / `inner_product` from pushdown when correctness must be uniform — costs the pushdown optimization.
2. Emit the finiteness screen into the generated SQL, so the pushed-down expression carries the same contract.
3. Accept the divergence and document it — only defensible if the engines agree in practice.

## Notes

Raised by an adversarial review of the #11263 fix, which checked whether the invariant the fix establishes is actually enforceable end to end. Kept out of that PR because it changes federation behavior rather than the kernels.

Related: `crates/search/src/index/duckdb/metric.rs`, the Elasticsearch kNN index, and S3 Vectors compute distances entirely outside these kernels too.

PR #13091, opened 2026-08-14
fix(udfs): return NULL for a vector distance that is not defined (refs #11263)

claudespice:
## Summary

A vector carrying `NaN` or an infinity has no cosine distance, L2 distance, inner product or norm — but the kernels answered one anyway, and the two dispatch paths did not agree on which one.

| input | `FixedSizeList<Float32>` (SIMD) | `List` / `LargeList` (scalar) |
|---|---|---|
| element is `NaN` / `±Inf` | **0** — distance 0, an exact match | 0.5 |
| zero-magnitude vector | 0.5 | 0.5 |

The `FixedSizeList` column is what embedding columns actually use, and `vector_search` passes its query vector as a `FixedSizeList` literal, so the production path is the one that answered **"perfect match"**. `_score` is `1 - distance`, so a failed embedding scored `1.0` and sorted above every real result; under `EmbeddingAggregation::Max` one corrupt chunk took its whole document to the top.

Root cause is that the guard on that path screened the kernel's **output** for non-finiteness. `simsimd`'s `cos` normalizes and returns a finite `0` for a `NaN` row, and nothing about that output distinguishes it from two genuinely parallel vectors — so an output guard cannot catch it.

## Changes

- Screen the **input** at the shared chokepoint, not the output. Only `Cosine` needs it: `Dot` and `L2Squared` accumulate their inputs directly, so a non-finite element reaches their output and the existing check catches it for free. `Kernel::hides_non_finite_input` records which is which, and the shared test drives all three kernels, so if that ever stops holding the test fails rather than a guard silently going missing.
- The scalar cosine kernels screen their three accumulators, which covers every non-finite element without a per-element branch. This check must precede the zero-magnitude branch — a `NaN` vector measured against a *zero* vector would otherwise be answered 0.5 on the strength of the zero side alone. That ordering has its own regression test.
- `l2_norm` needs no input screen at all: a sum of squares has no cancelling terms, so a non-finite element always reaches the result.
- Zero-magnitude vectors are unchanged. They are defined input with no direction, and both paths still call them orthogonal (0.5).

NULL is already a reachable score today (a null embedding row produces one), and the score aggregates are `max`/`avg`/`sum`, which skip nulls — so a poisoned chunk stops competing rather than starting to.

## Performance impact

This is a hot path, so it was measured, not assumed — `cargo bench -p runtime-datafusion-udfs --bench cosine_distance`, 10000 rows x 1536 dims:

| | trunk | this PR |
|---|---|---|
| cosine (SIMD) | 4.72 ms | 7.65 ms (**+62%**) |
| `inner_product`, `l2_distance`, `l2_norm` | — | unaffected (free output check) |

The screen's formulation was tuned against that benchmark; four alternatives measured worse and are recorded in the code so they are not retried: short-circuiting `iter().all(f32::is_finite)` (+296%, blocks auto-vectorization), a float `acc += v * 0.0` chain (+600%, serial FP dependency), an 8-lane manual split (+110%), and scanning both whole buffers up front (+60%).

Roughly **half** the remaining cost is rescanning the broadcast query literal once per row; removing it needs a signature change to the shared `make_scalar_function`, so it is filed as #13090 rather than grown into this PR. Correctness is the priority here per CLAUDE.md, but the number is real and a maintainer should see it.

## Not fixed here

- **The formula divergence in the same issue** — local `(1 - sim) / 2` vs federated `1 - sim`. It changes user-visible `_score` on *every* row and drifts the whole search snapshot suite, and #11263 explicitly parks it for a maintainer decision. This PR changes only which rows are NULL, not the scale of any defined score.
- #13088 — the federated path unparses to `array_cosine_distance` and does not inherit this NULL contract.
- #13089 — the write-side guards test `all(v == 0.0 || v.is_nan())` instead of `any(!v.is_finite())`, and DuckDB's own `validate_vector` is never called on write.

## Test plan
- [x] Added regression test for a non-finite element on every dispatch path (`FixedSizeList`, `List`, `LargeList`, `Float64` elements) and every kernel (`Dot`, `L2Squared`, `Cosine`, `l2_norm`)
- [x] Added edge case tests for non-finite-vs-zero-magnitude in both argument orders, zero-magnitude staying orthogonal, and a clean row in the same batch as a poisoned one still being scored
- [x] Each of the four guards was independently neutered and confirmed to fail a test, so none of them is vacuous — the neuter is also what surfaced the missing non-finite-vs-zero-vector case
- [x] `make test` passes

## Review gate

- Adversarial review: **skipped** — `codex` exited 1: "Your workspace is out of credits"; `grok` not installed. Its angles were run by hand instead, and two produced changes: the FSL path was verified to be what `vector_search` actually uses (`embeddings/udtf.rs:885`), and the performance question was measured rather than asserted — which is what caught a 4x regression in the first formulation.
- `/security-review`: no findings. The diff adds no SQL, I/O, auth, or logging; the behavior it removes was itself the exploitable direction (a writable `NaN` embedding ranked first in every search, a ranking-manipulation primitive in a RAG pipeline).
- `/simplify`: applied — dropped a redundant whole-buffer prescan in `l2_norm`, inlined a wrapper that had become a pass-through, replaced hand-rolled `Option` matches with `append_option`, and replaced two duplicated test builders with one generic helper in the shared `testing` module. Three findings were out of scope and filed as #13088, #13089 and #13090. Light checks re-run green.

Refs #11263 — deliberately **not** `Fixes`. The issue's headline defect is the local-vs-federated
formula divergence listed under *Not fixed here*, which #11263 itself parks for a maintainer
decision; this PR fixes the non-finite-input hazard filed alongside it. #11263 stays open until
that decision is made.

### DIFF
--- a/crates/runtime-datafusion-udfs/src/cosine_distance.rs
+++ b/crates/runtime-datafusion-udfs/src/cosine_distance.rs
@@ -28,6 +28,16 @@ limitations under the License.
 //! Both paths return `(1 - cosine_similarity) / 2` ∈ `[0, 1]` (0 = identical,
 //! 1 = opposite). Zero-magnitude vectors have undefined cosine direction; both
 //! paths treat them as orthogonal and return 0.5.
+//!
+//! A vector carrying `NaN` or an infinity has no distance at all, and both
+//! paths above return NULL for it. That matters for ranking: `_score` is
+//! `1 - distance`, so any fabricated distance for a failed embedding competes
+//! with real matches, and NULL is what keeps it out of the results.
+//!
+//! The NULL applies to these kernels only. `cosine_distance` is allowed to
+//! federate, where it unparses to the engine's own function (`DuckDB`'s
+//! `array_cosine_distance`) and that engine decides what a non-finite input
+//! scores — see #13088.
 
 use arrow::array::{
     Array, ArrayRef, Float64Array, Float64Builder, GenericListArray, LargeListArray, ListArray,
@@ -181,7 +191,9 @@ pub(crate) fn cosine_distance_inner(args: &[ArrayRef]) -> DataFusionResult<Array
         // divide by 2 to map into the standard [0, 1] distance range.
         // Zero-magnitude vectors have an undefined direction; simsimd treats the
         // dot product as zero (orthogonal), yielding distance 0.5.
-        (FixedSizeList(_, _), FixedSizeList(_, _)) => compute_fsl_f32_cosine(args),
+        (FixedSizeList(_, _), FixedSizeList(_, _)) => {
+            compute_fsl_f32(args, Kernel::Cosine, |v| v / 2.0)
+        }
         (List(_), List(_)) => general_cosine_distance::<i32>(args),
         (LargeList(_), LargeList(_)) => general_cosine_distance::<i64>(args),
         (array_type1, array_type2) => {
@@ -190,35 +202,6 @@ pub(crate) fn cosine_distance_inner(args: &[ArrayRef]) -> DataFusionResult<Array
             )
         }
     }
-}
-
-/// Compute cosine distance for `FixedSizeList<Float32>` with a non-finite safety guard.
-fn compute_fsl_f32_cosine(args: &[ArrayRef]) -> DataFusionResult<ArrayRef> {
-    let result = compute_fsl_f32(args, Kernel::Cosine, |v| v / 2.0)?;
-    let float64 = result
-        .as_any()
-        .downcast_ref::<arrow::array::Float64Array>()
-        .ok_or_else(|| {
-            DataFusionError::Internal("compute_fsl_f32 should return Float64Array".to_string())
-        })?;
-
-    // Safety net: convert any non-finite output to NULL to prevent NaN from
-    // sorting ahead of real scores in `ORDER BY _score DESC`.
-    let mut builder = arrow::array::Float64Builder::with_capacity(float64.len());
-    for i in 0..float64.len() {
-        if float64.is_null(i) {
-            builder.append_null();
-        } else {
-            let v = float64.value(i);
-            if v.is_finite() {
-                builder.append_value(v);
-            } else {
-                builder.append_null();
-            }
-        }
-    }
-
-    Ok(Arc::new(builder.finish()) as ArrayRef)
 }
 
 fn general_cosine_distance<O: OffsetSizeTrait>(arrays: &[ArrayRef]) -> DataFusionResult<ArrayRef> {
@@ -293,7 +276,7 @@ fn general_cosine_distance_f64<O: OffsetSizeTrait>(
             return exec_err!("Both arrays must have the same length");
         }
 
-        builder.append_value(cosine_distance_f64(
+        builder.append_option(cosine_distance_f64(
             &raw1[start1..end1],
             &raw2[start2..end2],
         ));
@@ -340,7 +323,7 @@ fn general_cosine_distance_f32<O: OffsetSizeTrait>(
             return exec_err!("Both arrays must have the same length");
         }
 
-        builder.append_value(cosine_distance_f32(
+        builder.append_option(cosine_distance_f32(
             &raw1[start1..end1],
             &raw2[start2..end2],
         ));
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

crates/runtime-datafusion-udfs/src/cosine_distance.rs
```diff
@@ -181,7 +191,9 @@ pub(crate) fn cosine_distance_inner(args: &[ArrayRef]) -> DataFusionResult<Array
         // divide by 2 to map into the standard [0, 1] distance range.
         // Zero-magnitude vectors have an undefined direction; simsimd treats the
         // dot product as zero (orthogonal), yielding distance 0.5.
-        (FixedSizeList(_, _), FixedSizeList(_, _)) => compute_fsl_f32_cosine(args),
+        (FixedSizeList(_, _), FixedSizeList(_, _)) => {
+            compute_fsl_f32(args, Kernel::Cosine, |v| v / 2.0)
```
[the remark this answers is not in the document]
claudespice: You're right, and this was the one thing in the PR that could do real damage on merge — thanks for catching it.

The body already had a *Not fixed here* section saying the `(1 - sim) / 2` vs `1 - sim` divergence is out of scope, and #11263 itself says *"Not fixing unilaterally because either direction changes user-visible score values; flagging for a maintainer decision."* Closing the issue anyway would have retired that decision without anyone making it — and the divergence is the issue's headline, not a footnote.

Fixed by retargeting the linkage rather than by widening the PR, since the formula choice is explicitly a maintainer call:

- Title: `(fixes #11263)` → `(refs #11263)` — the squash-merge commit inherits the PR title, so the keyword had to come off both places, not just the body.
- Body trailer: `Fixes #11263` → `Refs #11263`, with the reason stated inline so a future reader doesn't "helpfully" restore the closing keyword.

#11263 stays open for the formula decision. This PR keeps its actual scope: which rows are NULL, not the scale of any defined score.

crates/runtime-datafusion-udfs/src/vector_simd.rs
```diff
@@ -235,12 +276,28 @@ where
                 continue;
             }
         }
+        // A non-finite element makes the whole row's score undefined, and the
+        // kernels do not report that themselves: `simsimd`'s `cos` answers 0 —
+        // distance 0, an exact match — for a row carrying NaN, which puts a
+        // failed embedding at the top of `ORDER BY _score DESC`. Screen the
+        // input rather than the output, because a fabricated score is finite.
+        if screen_input && (!all_finite(slice_a) || !all_finite(slice_b)) {
```
bjchambers, line 284: Do we need the screen when simsimd is available? IIUC, we're screening and making the result *match* what simsimd would return. In that case, it seems like we could make the screening conditional on *not* using simsimd, which would also preserve the performance of the path when simsimd is available.

claudespice: The screen is needed *because* simsimd is the kernel — it's the simsimd path that hides the defect, so conditioning it on "not simsimd" would drop the guard exactly where the bug lives.

`SpatialSimilarity::cos` answers **0** for a row carrying `NaN`. Cosine distance 0 is an exact match, so `_score = 1 - distance` came out `1.0` and a failed embedding sorted above every real result; under `EmbeddingAggregation::Max` one corrupt chunk took its whole document to the top. Nothing about that output distinguishes it from two genuinely parallel vectors, which is why the output check can't catch it and the input has to be screened. So the screen isn't making the result match what simsimd returns — it's making it NULL, which is what neither path returns on its own.

There's also no non-simsimd path here to condition on. `simsimd = "6.5"` is an unconditional dependency (no feature gate), `Kernel::apply` always dispatches to it, and simsimd selects AVX-512 / AVX2 / NEON / scalar internally at load time — this code has no visibility into which. The other dispatch path, `List` / `LargeList`, never reaches this function: it screens its three accumulators in `cosine_distance.rs:473`, which is a separate and much cheaper guard.

The screen *is* already conditional, on the kernel rather than on simsimd — `Kernel::hides_non_finite_input()` is true only for `Cosine`. `Dot` and `L2Squared` accumulate their inputs directly, so a non-finite element reaches their output and the existing output check catches it for free; they skip the per-element scan entirely and pay nothing. `l2_norm` likewise.

On the cost, which is measured rather than assumed: **+62%** on the cosine SIMD path (4.72 ms -> 7.65 ms at 10000 rows x 1536 dims). Four cheaper formulations measured worse and are recorded in the code so they don't get retried. Roughly half of what remains is rescanning the broadcast query literal once per row — a real win that costs no correctness, needing a signature change to the shared `make_scalar_function`, which is why it's filed as #13090 rather than grown into this PR.

Resolving on that basis, but reopen if you read the tradeoff differently — the +62% is the one number here a maintainer might weigh against the ranking bug.

lukekim (approved):

PR #13100, opened 2026-08-14
Distributed full-text search for multi-node accelerated tables

### LINKED ISSUE #13080: Enhancement: Distributed full-text search for multi-node accelerated tables
## Goal-State/What/Result

Make the `text_search` user function (full-text search, FTS) work correctly on a multi-node accelerated table.

Today FTS works on a single node only. On a single node, `text_search(tbl, 'query', col, N)` reads one local Tantivy index and returns the top `N` rows by BM25 score. In multi-node (distributed) Spice, each executor node accelerates only its own partition of the table, and each executor holds a Tantivy index over its own rows only. The scheduler node has no index. A search on the scheduler therefore reads an empty or absent index and returns wrong results.

The goal-state: a `text_search` query on a distributed accelerated table returns the same top `N` rows, in the same order, as the same query on a single-node table that holds all the rows. The runtime does this in one query plan, without an external orchestration step.

## Why/Purpose

Full-text search is a core Spice capability. Distributed acceleration (multi-node) is the path to scale a dataset past one node. Right now a user cannot use both together: enabling FTS on a distributed accelerated table gives incorrect search results. Data correctness is the top principle, so this is a correctness gap, not only a feature gap. This issue closes the gap and lets search scale with the data.

## By When

**Issue/Spec written and reviewed:** _TBD_
**Done-Done:** _TBD_

## Done-Done

- [ ] The Algorithm
- [ ] PM/Design Review
- [ ] DX/UX Review
- [ ] Release Notes / PRFAQ
- [ ] Tests
- [ ] Documentation

## The Algorithm

- [ ] Simplify.

## Background

How FTS works today:

- The `text_search` UDTF resolves to a `SearchQueryProvider` (`crates/search/src/provider.rs`). Its `scan()` builds a plan: local Tantivy candidate scan → inner join on primary key with the base table → `Sort(_score DESC, primary_key ASC)` → `Limit N` → project.
- The Tantivy index is local to each node, on disk at `.spice/data/fts/<catalog>/<schema>/<table>/`. The acceleration write/refresh lifecycle builds it through an `IndexLayer` over the accelerator (`crates/runtime/src/search/full_text/table.rs`).
- The BM25 score depends on collection statistics: `N` (number of documents), `df(term)` (number of documents that contain a term), and `avgdl` (average document length). Each executor computes these from its own partition only.

Why multi-node breaks:

- The scheduler has no index, so a scheduler-side search is empty.
- A BM25 score from one executor is **not** comparable to a score from another executor, because each executor uses different local statistics. So a simple "collect each executor's top-N and re-sort by score" gives an approximate ranking, not the correct one.

How distributed acceleration already fans out a normal scan (the pattern to reuse):

- `PartitionedTableScanRewrite` (analyzer rule, scheduler only, `crates/runtime/src/datafusion/builder.rs`) rewrites `TableScan(tbl)` into a `UNION ALL` of one per-executor query. Each per-executor query becomes a `FlightSqlExec` that sends `SELECT ... FROM tbl ...` to one executor over Flight SQL. `FlightSQLPartialAggregatePushdown` pushes a partial aggregation into each per-executor query, and the scheduler merges the partial results.

## Specification

Add exact distributed full-text search. Express the whole operation as **one query plan**. The plan gathers global statistics and then scores with those global statistics, so the result equals the single-node result.

### Two parts of an exact distributed search

1. Global statistics. `N`, `df(term)`, and total document length are additive over disjoint partitions. So the global value of each is the sum of the local values across executors. Gathering global statistics is a distributed `SUM`.
2. Scored search. Each executor scores its own partition with the **global** statistics (not its local statistics) and returns its own top `N`. Because every executor now uses the same statistics, the scores are comparable. The scheduler merges the per-executor top-N lists, sorts by `_score DESC, primary_key ASC`, and takes the global top `N`.

### One plan

A new analyzer rule (`DistributedSearchRewrite`) rewrites the `text_search` table scan into one custom extension node. The child subtree of that node is the global-statistics `SUM`.

Logical plan after the rewrite:

```text
Extension: DistributedSearch{ table, query, column, fetch = N, scoring = global, executors = [...] }
  input[0] = Aggregate: group_by = [term], aggr = [SUM(df) AS df, SUM(n) AS n, SUM(dl) AS dl]
               Union
                 TableScan: text_search_stats('tbl', 'query', col) @ executor-1
                 TableScan: text_search_stats('tbl', 'query', col) @ executor-2
```

- The extension node output schema equals the search result schema (`schema(tbl) ∪ {_score}`).
- The rewrite builds the union of per-executor statistics queries directly from the executor set (`get_partitions`), so a second partition-rewrite pass is not needed.
- The `SUM ... GROUP BY term` is a stock DataFusion aggregate on the scheduler. `FlightSQLPartialAggregatePushdown` folds the trivial partial into each per-executor query.

Physical plan:

```text
DistributedSearchExec (fetch = N, sort = _score DESC, primary_key ASC)
  child = AggregateExec(Final, SUM, group_by = [term])
            UnionExec[ SearchStatsFlightSqlExec @ executor-i : SELECT * FROM text_search_stats('tbl', 'query', col) ]
```

At `execute()`, `DistributedSearchExec` does this:

1. Drain the child. This gives the small global-statistics table.
2. Send one scored query to each executor: `SELECT * FROM text_search('tbl', 'query', col, N)` with the global statistics bound.
3. Merge the returned per-executor top-N lists. Sort by `_score DESC, primary_key ASC`. Take `N`.

The statistics-to-scoring link is a runtime data dependency: the scored queries cannot be built until the statistics child produces the global values. So `DistributedSearchExec` runs its statistics child first, then runs the scored queries. This is one plan and one submission. Inside, it uses two round trips. Each executor still returns only its own top `N`, so data movement stays bounded.

### Why not a pure relational plan

A "purer" plan could make each executor return raw candidate features (per-term term frequency and document length) and then compute BM25 with a `Join(candidates × global_stats)` and a `TopK`. Do **not** use this shape. Without a global score, no executor can prune to `N` safely. It must either return every document that matches any term (exact but unbounded) or prune by local score to a large `K` (bounded but approximate). The custom `DistributedSearchExec` keeps the search exact **and** bounded.

### Passing the global statistics to the executor

Keep both per-executor queries as plain Flight SQL statements. The scored query passes the small statistics blob as a hidden named argument:

```sql
SELECT * FROM text_search('tbl', 'query', col, N, __global_stats => '<encoded df, N, avgdl>')
```

The executor `text_search` UDTF gets an optional `global_stats` argument. When the argument is present, the UDTF builds a custom Tantivy `Bm25Weight` from the injected `(total_num_docs, term df, avgdl)` instead of the segment-local statistics. When the argument is absent, the UDTF uses local scoring, which is also the `scoring = local` mode.

### Configuration

Add a scoring mode. Follow the "no boolean config, use a behavior-named enum" rule.

- `scoring = global` (default): exact. The plan gathers and injects global statistics.
- `scoring = local` (opt-in): approximate and faster. The plan skips the statistics child and scores with local statistics. This mode returns results that can differ from the single-node result. Document it as approximate.

### Correctness requirements

- **Disjoint partitions.** The `SUM` is exact only when the covering executor set holds disjoint rows. Distributed scans and aggregates already rely on this. State it; do not add a new assumption.
- **Same statistic that Tantivy scores against.** Tantivy counts collection size as the sum of each segment `max_doc`, which includes superseded documents until a merge removes them (`crates/search/src/generation/text_search/index.rs`). Sum the same quantity. So "exact" means "equal to the single-node result within the existing merge-state tolerance" (the current `del_docs_ratio` bound), not tighter.
- **Analyzer parity.** Every executor must tokenize and stem the query the same way, so `df(term)` aligns across executors. This is true today because all executors share the index schema and tokenizer.
- **Local join stays on the executor.** Each per-executor query runs the full local search plan, including the inner join that drops stale index entries whose primary key is gone from the base partition. This keeps that correctness behavior per partition.
- **Filter before limit.** Push any outer predicate into each per-executor query so the top-N is taken over the filtered set. Never take top-N and then filter.
- **Offset.** Push `skip + N` to each per-executor query. Apply the real `skip`/`fetch` only on the scheduler.

### Security Review

The global-statistics blob and the query text move over the existing intra-cluster Flight SQL channel between scheduler and executors. This channel already carries every distributed sub-query and is authenticated the same way. This change adds no new external surface. Confirm the blob is bounded in size (a few query terms) so it cannot be used to inflate a request.

## How/Implementation Plan

New or changed components:

1. `DistributedSearchRewrite` analyzer rule. Match a `TableScan` over a `SearchQueryProvider` that carries `UdtfSource::TextSearch`. Gate on the accelerated table through the existing `AcceleratedPartitionProvider::should_partition` (it already sees through the `IndexLayer`). Emit the `DistributedSearch` extension node with the statistics-aggregation child.
2. `DistributedSearchExec` physical operator. Sequence the statistics child, send the scored queries, merge and limit. This is the only new operator.
3. A Flight SQL query leg whose `FROM` source is a UDTF call, driven by the `UdtfSource`. Both `text_search_stats(...)` and `text_search(...)` legs use it. This is the only Flight SQL change.
4. `text_search_stats` UDTF. On an executor, tokenize the query, then emit one row per query term with the local `df`, plus the local document count `N` and the local total document length.
5. `text_search` UDTF: add the optional `global_stats` argument. When present, build a custom Tantivy `Bm25Weight` from the injected statistics.

Reused, unchanged:

- Stock DataFusion `AggregateExec` for the `SUM`.
- The partition machinery and Flight SQL data plane.

Optimization (optional, later):

- `N` and `avgdl` do not depend on the query. Maintain them per table on refresh/CDC, or gather and cache them. Then the statistics child shrinks to a `df`-only gather for the query terms, or disappears on a cache hit.

Later generalization (not in scope here):

- The "gather distributed statistics, then run a parameterized main query" pattern also fits async distributed query and other adaptive plans (for example, global quantiles for a balanced range-partitioned sort). This search operator is self-contained and does **not** need the async job engine. Treat the async engine as a later home for the staging, with this operator as its first real workload.

## QA Plan

- Unit tests for the analyzer rule: correct plan shape, executor fan-out, and the statistics-aggregation child.
- Unit tests for `DistributedSearchExec`: statistics child runs first; scored queries carry the merged statistics; merge and limit are correct.
- Correctness test on the wrapped/partitioned path (`crates/runtime/tests/cluster/`): a distributed accelerated table returns the same top-N rows and order as a single-node table with the same rows. A defaulted no-op in a wrapper regresses this silently, so test the wrapped path, not only units.
- Skew test: partitions with very different term statistics. Confirm `scoring = global` matches the single-node result and `scoring = local` is close but marked approximate.
- Edge cases: empty result, one executor, a missing partition, NULL and boundary inputs.
- Search quality regression: run the existing IR benchmark harness on a distributed setup and compare against single-node scores.

## Scope

In scope:

- Exact distributed `text_search` on multi-node accelerated tables, in one plan, with `scoring = global` default and `scoring = local` opt-in.

Out of scope (follow-on issues):

- `vector_search` fan-out (it has the same shape and is exact in one round, because distance scores carry no collection statistics).
- `rrf` (reciprocal rank fusion) across executors.
- The async distributed query engine.
- The statistics cache optimization.

## Release Notes

**Distributed full-text search**: `text_search` now returns correct, ranked results on multi-node accelerated tables. Spice gathers global relevance statistics across executors and scores with them, so a distributed search returns the same top results as a single-node search. A faster approximate mode is available.

```yaml
# Example: full-text search on a distributed accelerated dataset
datasets:
  - from: postgres:documents
    name: documents
    acceleration:
      enabled: true
      engine: cayenne
      partition_by:
        - bucket(10, id)
    columns:
      - name: body
        full_text_search:
          enabled: true
```

Jeadie:
Enable full text search on distributed accelerations. 

### Overview
Single node: read local tantivy index, compute scores, return search results.
Multi-node: Two round trip process
  1. Retrieve per-executor query statistics. Scheduler aggregates to produce global statistics (equivalent to statistics from a local index for a single node).
  2. Retrieve per-executor candidates. Executors receive global statistics, compute and return per row scores. Since each executor has disjoint rows, scheduler can produce top-K by getting aggregating per-executor top-K. 

#### Single Node
```
          ┌────────────┐   read      ┌───────────────────┐
          │  text_      │────────────▶│  local Tantivy    │
          │  search     │             │  index            │
          │            │◀────────────│  (N, df, avgdl,    │
          └─────┬──────┘   score      │   postings)       │
                │                     └───────────────────┘
                ▼
          top-N rows by BM25          one index → one set of collection stats → done
```
#### Multi node
```ascii
ROUND 1 ── gather statistics 

                                  ┌─────────────────┐
                                  │    SCHEDULER    │
                                  │                 │
                                  │  text_search    │
                                  └────────┬────────┘
                  text_search_stats(...)   │ 
                   ┌───────────────────────┼───────────────────────┐
                   ▼                       ▼                        ▼
            ┌────────────┐         ┌────────────┐          ┌────────────┐
            │ EXECUTOR 1 │         │ EXECUTOR 2 │          │ EXECUTOR n │
            │  local idx │         │  local idx │          │  local idx │
            └─────┬──────┘         └─────┬──────┘          └─────┬──────┘
        (df,N,tok)│           (df,N,tok) │            (df,N,tok) │
                  └──────────────────────┼───────────────────────┘
                                         ▼
                               ┌───────────────────-─┐
                               │      SCHEDULER      │   Σ over disjoint executors:
                               │ ─────────────────── │   N   = Σ Nᵢ
                               │  GLOBAL STATISTICS  │   df  = Σ dfᵢ
                               │ (== single-node idx)│   tok = Σ tokᵢ
                               └─────────────────────┘
```
```ascii
ROUND 2 ── scored search 

                                  ┌─────────────────┐
                                  │    SCHEDULER    │
                                  └────────┬────────┘
       text_search(..., global_stats =>)   │
                   ┌───────────────────────┼───────────────────────┐
                   ▼                       ▼                       ▼
            ┌────────────┐         ┌────────────┐          ┌────────────┐
            │ EXECUTOR 1 │         │ EXECUTOR 2 │          │ EXECUTOR n │
            │  local idx │         │ score with │          │ score with │
            └─────┬──────┘         └─────┬──────┘          └─────┬──────┘
      local top-N │                local │ top-N     local top-N │
        w. Global └───────────────────-──┼───────────────────────┘
          Stats                          ▼ 
                            ┌──────────────────────┐
                            |      SCHEDULER        |
                            │ ──────────────────── │
                            | EXECUTOR 1 UNION ALL |
                            |    ... EXECUTOR n    │
                            │ ORDER BY _score DESC |
                            | LIMIT N              |
                            └───────────┬──────────┘ 
                                        ▼
                                      Output
```

### Implementation
 - `DistributedSearchRewrite` analyser rule to convert `text_search` UDTF to use a `DistributedSearchExec` (from a user defined logical plan node `DistributedSearchNode`). 
 - New UDTF `text_search_stats` that returns FTS index statistics for a query. 
 - `DistributedSearchExec` carries out the above mentioned two part process. 
 - `text_search` UDTF optional arg `global_stats` for precomputed index stats to use (instead of computing from local). 
 - `text_search` gains an optional `global_stats` argument.
- Support `FROM udtf(...` in `FlightSQLTable`. 

Closes #13080.


Jeadie:
### Design note: the statistics gather could be simpler

This PR gathers the global BM25 statistics with a **relational** subplan — `Aggregate(SUM ... GROUP BY term)` over a `Union` of one `text_search_stats` Flight SQL leg per executor — while the scored round is **imperative** (`DistributedSearchExec` builds SQL and calls `query_to_stream` per executor). The scored round must be imperative, because it depends on the runtime-computed statistics.

Because the operator already owns an imperative fan-out, the statistics round can use the same mechanism: query each executor's `text_search_stats(...)` with `query_to_stream`, then sum the rows in Rust (`GlobalBm25Stats::add`). That removes:

- the Flight SQL table-function-in-`FROM` capability (`from_function` on `FlightSqlExec`/`FlightSQLTable`) and the aggregate-pushdown skip that only protects it;
- the analyzer's `Union`/`Aggregate` construction (`build_stats_plan`/`stats_leg`);
- the extension node's logical/physical child input (the node becomes a leaf).

`text_search_stats` the UDTF stays either way — only the transport changes, and `query_to_stream` already sends arbitrary SQL.

**Trade-off:** the relational form is a down-payment on a reusable "gather distributed statistics, then parameterize the main query" pattern (for example, a later async distributed-query engine). The statistics here are tiny (a few terms × a few executors), so the aggregate-pushdown machinery adds no scale benefit today. The reuse is the only reason to keep the relational form.

**Recommendation:** unless we want to keep that pattern on purpose now, collapse the statistics gather into the operator (imperative, symmetric with the scored round) and drop the Flight SQL UDTF-in-`FROM` path. Left as a follow-up so this PR lands a known-good baseline first.

Issue #13110, opened 2026-08-14
Flight request metrics are recorded by convention, not by construction: a new early return in do_put silently records nothing


claudespice:
## Describe the bug

After #12844's fix, each Flight RPC records exactly one `flight_requests` /
`flight_request_duration_ms` sample — but nothing *enforces* that. The rule is
carried by convention: the `FlightService` impl must not start a timer, and every
handler must start exactly one on every path it owns.

`do_put::handle` is where that convention is most expensive to keep. It now has
nine `let _start = metrics::track_flight_request("do_put", None).await;` sites,
one per early return, because the request's own measurement is not started until
the table path is resolved near the end of the function. A tenth early return
added later — a new validation, a new command variant, a new auth check — records
nothing, and nothing fails:

- The counter increments inside the async fn body, so an un-awaited future
  records not even a zero, it records *nothing*.
- Binding to a named `_start` suppresses `unused_must_use`.
- `clippy::let_underscore_future` only fires on the `let _` form.
- The regression test (`crates/runtime/tests/flight_request_metrics.rs`) pins the
  paths it exercises. A path added tomorrow is not one of them.

That is exactly how the original defect survived: five sites in `do_put` and one
in `do_exchange` bound the future without awaiting it, and the only reason it was
not visible as a gap is that the (duplicate) outer timer covered them.

## Why it matters

The failure is silent and one-directional: the RPC still succeeds, and the
missing samples look like a quieter endpoint rather than a broken instrument.
Rejection paths are the ones most likely to be missed and the ones an operator
most wants counted — an auth denial or a rate-limit refusal that never reaches a
dashboard is invisible during an incident.

## Suggested direction

Make once-per-RPC structural rather than conventional. Sketches, roughly in
increasing order of cost:

1. **Record once at the top of `handle`, refine the label later.** Start the
   measurement as the first statement and attach the `command` label when it
   becomes known (`TimeMeasurement::with_labels` already exists). Every early
   return is then covered by the binding's `Drop` with no per-site call. The
   obstacle is the handoff: `handle` delegates to command-specific handlers that
   start their own measurement, so the outer one needs a way to be *disarmed*
   when it delegates — a `TimeMeasurement::cancel()` (or `into_disarmed()`) on
   `crates/telemetry/src/timing.rs`. That is a small, reusable addition, and it
   turns "remember to record on this path" into "remember to disarm on the three
   paths that delegate".
2. **Split `handle` into a thin recording wrapper and an inner function** that
[...]

PR #13112, opened 2026-08-14
fix(flight): record one metric sample per Flight RPC (fixes #12844)

### LINKED ISSUE #12844: Flight RPCs record duplicate latency samples: each DoGet emits both a setup-only and a full-duration observation
Each Flight `DoGet`/`DoPut`/`Handshake` records **two** samples into `flight_request_duration_ms` and **two** increments into `flight_requests`, from two nested timers on the same RPC.

`crates/runtime/src/flight/mod.rs:183` starts a timer with no `command` label. It is a local, so it drops when the handler returns — after planning, but before the returned stream is polled:

```rust
async fn do_get(&self, request: Request<Ticket>) -> Result<Response<Self::DoGetStream>, Status> {
    let _start = track_flight_request("do_get", None).await;   // records at fn return
    let response = Box::pin(do_get::handle(request)).await?;
    Ok(Self::wrap_response_stream_with_scope(response).await)
}
```

Every arm of the `match` in `do_get.rs:42-140` then starts its own timer, which is moved into `TimedStream` (`statement_query.rs:126`) and records at end-of-stream (`telemetry/src/timing.rs:137-141`).

`sql_to_flight_stream` returns a lazy `SendableRecordBatchStream`, so the scan/filter/aggregate all run while tonic drains it. The outer sample is therefore a small prefix of the real query latency, not an approximation of it.

| | outer timer (`mod.rs:183`) | inner timer (`statement_query.rs:82`) |
| --- | --- | --- |
| labels | `method=do_get` | `method=do_get, command=statement_query` |
| clock spans | entry → stream constructed | entry → stream drained |

Consequences for any panel not filtering on `command`:

- `flight_requests{method="do_get"}` reports **2x** the real RPC rate.
- `histogram_quantile(0.99, ...{method="do_get"})` is computed over a mix in which every real latency is paired with a shorter prefix sample, pulling the reported p99 down to roughly the true p98 — and further the longer a query streams.

`get_flight_info` (`mod.rs:156-161`) already delegates without an outer timer; that is the shape the others should match.

### Related defects on the same path

`handshake` is recorded under two different method names for one RPC: `mod.rs:137` labels it `do_handshake`, `handshake.rs:46` labels it `handshake`. Nothing else in the codebase emits `do_handshake`.

Six call sites bind the `async fn` without awaiting it, so the future is dropped unpolled and **nothing is recorded — not even the counter**:

```rust
let _start = metrics::track_flight_request("do_put", None);   // no .await
```

`do_put.rs:68, 72, 168, 172, 180` and `do_exchange.rs:166`. The rustc `let_underscore_future` lint only fires on `let _ = fut`, and binding to `_start` suppresses `unused_must_use`, so these are silent.

### Fix

1. `do_get` — delete the outer timer at `mod.rs:183`. Every inner branch already records, including the error arms.
2. `handshake` — delete the outer timer at `mod.rs:137`. `handshake.rs:46` is the correct one (it goes into `TimedStream`, so it spans the response).
3. `do_put` — add the missing `.await`s **first**, then delete the outer timer at `mod.rs:192`. In the current state those early-error paths record nothing, so the outer timer is their only coverage; removing it first would silently drop error-path samples.
4. `do_exchange.rs:166` — add the missing `.await`.

A regression test asserting one `flight_requests` increment per RPC would keep this from recurring; the duplication is invisible to the compiler and to lints.

Version: 2.2.0-unstable; refs against trunk de31505ea8.

claudespice:
## Summary

Every Flight RPC that delegates to a handler recorded its `flight_requests` /
`flight_request_duration_ms` sample **twice**, and six rejection paths recorded it **zero** times.

**Root cause.** `impl FlightService for Service` started a timer of its own and then delegated to a
handler that already starts one. The two samples are not equivalent: the handler's measurement
carries a `command` label and, where the response is a stream, is moved into a `TimedStream` so it
spans the drain, while the outer one is a local that drops when the handler returns — after
planning, before tonic polls the returned `SendableRecordBatchStream`. So on any panel not
filtering by `command`:

- `flight_requests{method="do_get"}` reported **2x** the real RPC rate.
- `histogram_quantile(0.99, …{method="do_get"})` was computed over a population in which every real
  latency is paired with a shorter prefix of itself, pulling the reported p99 toward the true p98 —
  and further the longer a query streams.

`handshake` was worse: the outer timer labelled it `method=do_handshake` and the handler labels it
`method=handshake`, so one RPC produced two samples under two method names and neither series
carried the real rate.

The mirror-image defect is on the same path. Six call sites bound the tracking future without
awaiting it:

```rust
let _start = metrics::track_flight_request("do_put", None);   // no .await
```

The future is dropped unpolled, so nothing is recorded — not even the counter, which increments
inside the async fn body. Binding to a named `_start` suppresses `unused_must_use`, and
`clippy::let_underscore_future` only fires on the `let _` form, so neither the compiler nor the
lints saw it. Those sites are `do_put`'s rejection paths and `do_exchange`'s entry; the service
impl's timer was their only coverage, which is why they had to be fixed *before* it could be
removed.

`get_flight_info` was already the correct shape — it delegates without an outer timer — and is what
the rest now match.

## Changes

- `flight/mod.rs`: drop the outer timer from `handshake`, `get_schema`, `do_get`, `do_put`,
  `do_exchange`, `do_action` and `list_actions`. `list_flights` and `poll_flight_info` keep theirs:
  they are unimplemented and have no handler to delegate to. The invariant is stated on the impl
  block so the next RPC added there doesn't reintroduce the duplicate.
- `flight/do_put.rs`: `.await` the five dropped futures, and record on the three rejection paths
  that had no timer at all — a `Command` that fails to convert, a rate-limit refusal, and both
  authorization denials. All of these sit before the point where the request's own measurement is
  started, so none of them can double-count.
- `flight/do_exchange.rs`: `.await` the dropped future.
- `flight/do_get.rs`, `flight/get_flight_info.rs`: record on the one pre-dispatch path in each that
  reported nothing — a ticket or descriptor that decodes as a `prost` message but not as a `Command`
  never reaches an arm that records. `get_flight_info` already had no outer timer, so that path
  reports nothing today rather than merely losing a duplicate.
- `flight/flightsql/prepared_statement_query.rs`, `.../prepared_statement_update.rs`: record in
  `do_put_query` / `do_put_update`. These two DoPut variants never recorded anything themselves and
  were covered only by the service impl's timer, so they get the label their `do_get` and
  `get_flight_info` siblings already use.
- `flight/metrics.rs`: document the invariant on `track_flight_request`, and move rather than clone
  the label vector into `TimeMeasurement` — it takes `impl Into<Vec<KeyValue>>`, so passing a slice
  deep-cloned every label, including the owned `user_agent` string the request context contributes.

Verified that each handler touched here has exactly one caller, so no path gains a second sample.

## Behavior change for dashboards

No metric is renamed or removed, but `do_handshake` disappears as a *label value* on
`flight_requests` / `flight_request_duration_ms`. Nothing else in the repo emits it, neither the
Grafana nor the Datadog dashboard in `monitoring/` references it (both select `method:do_get` /
`do_put` / `get_flight_info`, or group by `method`), and the documented name for that RPC is
`handshake` — `docs/release_notes/rc/v1.0.0-rc.1.md` maps `flight_handshake_requests` to
`n {method, command}`. An operator with a saved query on `method="do_handshake"` should move it to
`method="handshake"`, which now carries the real rate rather than half of it.

Counter rates on `do_get`/`do_put`/`get_schema`/`do_action`/`list_actions` will halve where a panel
does not filter by `command`; that is the correction, not a regression.

## Test plan
- [x] Added regression test asserting exactly one `flight_requests` increment per Flight RPC
      (`each_flight_rpc_records_exactly_one_request` in `crates/runtime/tests/metrics.rs`), covering
      `handshake`, `do_get`, `get_schema`, `list_actions`, and a `do_put` rejection
- [x] Neuter-verified in both directions: restoring the outer `do_get` timer fails the test with
      `recorded 2`, and dropping the `.await` on a `do_put` rejection fails it with `recorded 0`, so
      the assertion discriminates one from two and from zero
- [ ] `make lint` passes (workspace-wide `cargo fmt --all -- --check` + full clippy — not a per-crate scoped invocation)
- [ ] `make test` passes

`make lint` and `make test` cannot run from this checkout: it is a `.claude/worktrees/` worktree, and
`cargo fmt --all` fails there because `connector-nfs` resolves to the outer workspace, which excludes
it by a path relative to itself (#13093). Signed off with `make signoff-remote`, which runs the same
gate on the self-hosted pool. Those two boxes are left unchecked rather than claiming a run that did
not happen. Locally: `cargo fmt -p runtime`, `cargo check -p runtime --lib` and
`cargo test -p runtime --test metrics` (7 passed) are green.

`do_exchange`, `do_action` and the two prepared-statement `do_put` variants are fixed but not covered
by the test — each needs a live subscription or a prepared-statement handle to drive. Their sample
count is unchanged by this PR (one before, one after), so the uncovered risk is a future regression
rather than this one.

## Review gate

- Adversarial review: **skipped** — `codex` exited 1: "Your workspace is out of credits"; `grok` not installed. Ran its angles by hand instead: confirmed every handler this touches has exactly one caller (no new double-count), and traced why the un-awaited half was possible at all — `track_flight_request` is `async` solely to reach `AsyncMarker::new().await`, which is a `PhantomData` lint device, so making it a non-async `#[must_use]` fn taking `AsyncMarker` would turn the whole class into a compile error. That is ~62 call sites and would bury this correction under a refactor, so it is filed as #13110 rather than done here.
- `/security-review`: no findings. The rate-limit restructure and both authorization arms preserve their `return Err(..)`; every label added is a `&'static str`, so no request-controlled value is newly promoted into a metric dimension; no new log, error, or metric carries request data.
- `/simplify`: applied — folded the regression test into the existing `tests/metrics.rs` binary rather than adding a new test target (it was a third byte-identical copy of the Prometheus harness and a 1.1 GB debug link), trimmed the doc comments to the invariant, and took the `TimeMeasurement` label move. Skipped a suggested `record_rejection` helper: the same review found the non-async signature change above strictly dominates it, so adding an API now that would be subsumed is churn. One of its suggestions (a point-free `.map(Counter::value)`) did not typecheck and was reverted to the form its sibling test already uses. Light checks re-run green.

Fixes #12844


claudespice:
@copilot review

sgrebnov (dismissed):

sgrebnov (approved):

Issue #12958, opened 2026-08-12
The sort-merge join rewrite over-fires, and cayenne_sort_merge_min_rows is read only on a path the runtime never takes


lukekim:
## Summary

The hash-join → sort-merge-join rewrite fires far earlier than intended — in a worked case at ~1.7M build rows where the documented floor is 10M — and the knob that is supposed to hold it back, `cayenne_sort_merge_min_rows`, is **inert in production**: it is read only on a legacy code path the runtime never takes.

All references pinned to `cd2e19113f`.

## Where

| What | Location | Note |
|---|---|---|
| Live memory gate | `crates/cayenne/src/optimizer_rules.rs:790-833` | the path production takes |
| 2.5× build-side inflation | `crates/cayenne/src/optimizer_rules.rs:991-1012` | accepts `Precision::Inexact` |
| The dead knob | `crates/cayenne/src/optimizer_rules.rs:850` | inside the no-pool legacy arm |
| Why that arm is unreachable | `crates/cayenne/src/builder.rs:1769` | sets the pool unconditionally |

## Problem 1 — the rewrite over-fires

The live gate is:

```
min(0.125 × pool, pool / hash_join_count)
```

applied to a build-side estimate that has been **inflated 2.5×** (`optimizer_rules.rs:991-1012`) and that accepts `Precision::Inexact` values. Net effect: the rewrite triggers once the *raw* payload exceeds roughly **5%** of the pool.

Worked example — `memory_limit: 2 GiB`, a query with 12 hash joins:

- gate = `min(0.125 × 2 GiB, 2 GiB / 12)` = `min(256 MiB, 170 MiB)` = **170 MiB**
- against a 2.5×-inflated estimate ⇒ real payload threshold ≈ **68 MiB** ≈ 3.3% of the pool
- ⇒ rewrite at roughly **1.7M build rows**, where the intended floor is **10M**

This is the eager-sort-merge-fallback shape: sort-merge is the safety net, hash join is the fast path, and the safety net is being deployed on mid-size joins that would have completed comfortably as hash joins.

## Problem 2 — the documented brake is inert

`sort_merge_min_rows` is read at exactly one place, `optimizer_rules.rs:850`, which sits inside the **no-pool legacy arm**. `builder.rs:1769` sets `sort_merge_memory_pool_bytes` unconditionally, so the runtime always wires a pool and that arm is never taken.

[...]

PR #13177, opened 2026-08-16
fix(cayenne): weight the sort-merge fair-share term by estimated build bytes

### LINKED ISSUE #13155: The sort-merge fair-share term charges every hash join an equal slice of the pool, regardless of build size
Split out of the review of the fix for #12958, which narrows the symptom but deliberately leaves this cause in place.

## Describe the bug

`try_rewrite_oversized_join` budgets each hash join a share of the query memory pool:

```rust
let fair_share = pool_bytes / hash_join_count.max(1);
let effective_gate = gate_bytes.min(fair_share);
```

`count_hash_joins` (`crates/cayenne/src/optimizer_rules.rs`) counts every `HashJoinExec` in the plan and weights them equally, so a 1,000-row build side reserves exactly as much of the pool as a billion-row one. On a wide plan the joins that will never allocate anything still divide the pool, dragging the share of the join that *does* allocate far below what is actually available.

The fixture added for #12958 shows the shape directly: three joins over a 144 MiB pool give each a 48 MiB share, but two of those builds are 1,000 rows and between them will touch well under a megabyte. The join under test is held to 48 MiB when ~143 MiB of the pool is genuinely free.

## Why it matters now

The #12958 fix stops the fair-share term from dragging mid-size joins onto sort-merge by requiring such a join to also clear `sort_merge_min_rows`. That is the right instrument for the reported symptom — an operator setting the documented knob and observing nothing — but it loosens the bound the fair-share term exists to enforce. A plan with `hash_join_count > 1 / sort_merge_memory_pool_fraction` (above eight joins at the 0.125 default) can now admit build sides summing past the pool.

A row floor is also dimensionally wrong for a byte decision: at the 10M default it means roughly 600 MB for a three-`Int64` schema and several GB for a wide one, a schema-dependent swing that `build_side_memory_estimate` exists to eliminate.

## Expected behavior

A join's share of the pool should reflect what it is estimated to allocate, so joins that will not allocate do not reserve budget.

## Suggested fix

Weight the share by estimated build bytes rather than by join count — the estimates are already computed per join by `build_side_memory_estimate` — or, as a cheaper approximation, floor the share so a plan with many trivial joins cannot drive it arbitrarily low. Either fixes the cause in the same expression and the same units, and would let the row floor added for #12958 be revisited.

## Verification

Read from `crates/cayenne/src/optimizer_rules.rs` on `trunk`. Not measured against a running workload; the arithmetic above is from the source and from the test fixture.

atirna:
## Summary

`try_rewrite_oversized_join`'s fair-share term divided the query memory pool equally across every `HashJoinExec` in the plan:

```rust
let fair_share = pool_bytes / hash_join_count.max(1);
let effective_gate = gate_bytes.min(fair_share);
```

so a 1,000-row build side reserved exactly as much pool budget as a billion-row one. On a wide plan the joins that will never allocate meaningfully still divided the pool, dragging the share of the join that *does* allocate far below what is genuinely available and pushing mid-size builds onto sort-merge — the #13155 fixture: three joins over a 144 MiB pool each get a 48 MiB slice while two of the builds together touch well under a megabyte.

## Changes

- `optimize()` now collects every hash join's `build_side_memory_estimate` once up front (the original plan's build sides are the concurrency pressure being budgeted against) and passes a `BuildEstimates` summary to the rewrite decision.
- The per-join share is `pool × estimate / total estimated build bytes`, where joins whose estimate cannot be computed (absent row statistics, unestimatable schema) keep an equal-share charge so their pressure is reserved rather than dropped.
- The plan-wide total is held in `u128`: a `usize` total would saturate on extreme plans and collapse every share back to the whole pool, silently dropping the pool-wide bound the fair-share term exists to enforce. Under demand past the pool every join's share now sits below its own estimate, so the rewrite fires.
- Equal builds still reduce to `pool / join_count`, so the q78 sum-of-build-sides protection is unchanged, and a lone join still gets the full share, so the absolute-fraction behavior is unchanged.

## Test plan

- `cargo fmt --check`
- `cargo clippy -p cayenne --no-deps --tests`
- `cargo test -p cayenne --lib` — 1067 passed

New tests in `crates/cayenne/src/optimizer_rules.rs`:

| Test | Pins |
|---|---|
| `leaves_large_build_beside_trivial_joins_that_count_sharing_would_rewrite` | the #13155 fixture: a 90 MB build beside two 60 KB builds over a 144 MiB pool stays a hash join; count-sharing rewrote it |
| `rewrites_unequal_builds_whose_sum_exceeds_the_pool` | weighting cuts both ways: 60 MB + 120 MB over the 144 MiB pool both rewrite (only the fair-share term fires) |
| `charges_joins_with_unknown_build_estimates_for_their_pressure` | an unknown-statistics join still reserves `pool / n`; the known sibling is rewritten, the unknown join itself is not |
| `rewrites_builds_whose_aggregate_estimate_exceeds_usize` | nine builds whose estimates each fit `usize` but sum past `usize::MAX`: every join rewrites instead of each collapsing to the whole pool |
| `build_estimates_weighted_share_reduces_to_equal_shares_for_equal_builds` | equal builds reduce exactly to `pool / join_count` |
| `build_estimates_weighted_share_zero_and_unrepresentable_totals` | zero-total, unrepresentable-total, unknown-charge, and empty-set boundaries |

Verified by neutering the fix in two directions, each failing a different set:

| Neuter | Fails |
|---|---|
| restore count-based `pool / join_count` (pre-fix behavior) | `leaves_large_build_beside_trivial_joins_…`, `rewrites_unequal_builds_…`, `build_estimates_weighted_share_zero_and_unrepresentable_totals` |
| drop the unknown-estimate equal-share charge | `charges_joins_with_unknown_build_estimates_…`, `build_estimates_weighted_share_zero_and_unrepresentable_totals` |

## Relationship to #13157

That PR gates only the fair-share term behind `sort_merge_min_rows` for #12958 and expressly leaves the equal-share root cause here unfixed. This PR fixes the cause in the same expression and the same units (bytes, not rows), so the schema-dependent swing of a 10M-row floor at ~600 MB for a three-`Int64` schema goes away. #13157's row floor becomes redundant once this lands, as its own description anticipates.

Fixes #13155


atirna:
the label/assignee checks failed since i don't have permission to add labels or assignees as an outside contributor — could you add `area/cayenne` (or whichever fits) and an assignee? thanks!

atirna:
closing this one out for now, its been sitting a while and i dont want to leave a stale branch on your queue. happy to reopen or rebase it if its still useful to you.

PR #13209, opened 2026-08-17
fix(search): join all of a dataset's vector indexes on one VectorScanTableProvider

### LINKED ISSUE #7020: Properly initialize `VectorScanTableProvider` without nesting
As a follow up to https://github.com/spiceai/spiceai/pull/7017, the better solution is to properly initialize `VectorScanTableProvider` so that there are no nested providers.

Instead we should create a single `VectorScanTableProvider` which supports multiple vector indexes.

krinart:
## Summary
- `VectorScanTableProvider` nested a separate provider layer per embedding column, so a dataset with N vector indexes stacked N `SpiceTable` nodes instead of one — the underlying issue #7017 patched around (recursive unwrapping) rather than fixed.
- `VectorScanTableProvider` now holds a `Vec<VectorIndexJoin>` (one entry per index) and joins in only the indexes a given query actually references, in a single flat scan.
- The S3 Vectors and Elasticsearch construction paths (`wrap_table_as_index_s3`/`wrap_table_as_index_elasticsearch`) now build exactly one `VectorScanTableProvider` per dataset instead of re-wrapping the previous column's layer on every iteration.
- For a dataset with a single vector index, the generated plan is unchanged (same join alias, same filter/dedup/project sequence) — all existing single-index snapshot tests pass without regeneration.

Fixes #7020.

## Test plan
- [x] `cargo test -p search --features s3_vectors,elasticsearch --lib`: 192/192 passed, including a new `test_vector_scan_multiple_indexes` regression test asserting (via `EXPLAIN` snapshots) that a two-index provider joins only the needed index(es) — one join when a query references one column's embedding, two chained joins over a single `TableScan` when it references both, zero joins when it references neither.
- [x] `cargo test -p runtime --lib --features s3_vectors,elasticsearch,duckdb embeddings`: 59/59 passed, including the `changes_stream_resolves_through_a_vector_scan` CDC-resolution regression test.
- [x] `cargo test -p runtime-table --lib`: 303/304 passed; the one failure (`test_refresh_status_change_to_ready`, an unrelated metrics-polling timing test) passes in isolation and is pre-existing flakiness under parallel load.
- [x] Manually verified end-to-end against a live S3 Vectors dataset with two embedding columns (`body`, `title`) on the same table: both indexes initialize and refresh concurrently without error, `vector_search(...)` correctly reports "2 vector search columns" and requires disambiguation, and `EXPLAIN` on a query selecting both embedding columns shows two chained joins over one table scan.

### DIFF
--- a/crates/search/src/index/vector_table.rs
+++ b/crates/search/src/index/vector_table.rs
@@ -43,24 +43,20 @@ use util::session_state::builder_from_existing;
 use crate::index::VectorIndex;
 use datafusion::catalog::{ScanArgs, ScanResult};
 
-/// One vector index's listing plan and join key, as composed onto a
-/// [`VectorScanTableProvider`]. A dataset with several embedding columns holds
-/// one of these per column, all on the same [`VectorScanTableProvider`], rather
-/// than nesting a separate layer per index.
-#[derive(Debug, Clone)]
-pub struct VectorIndexJoin {
-    pub vector_index_list: Arc<LogicalPlan>,
-    pub primary_key: Vec<String>,
-}
-
 /// A [`TableProvider`] that adds one or more embedding columns to an underlying
-/// [`TableProvider`] — one [`VectorIndexJoin`] per vector index, joined into the
-/// same scan rather than each index wrapping its own nested
-/// [`VectorScanTableProvider`] layer.
+/// [`TableProvider`] — one listing plan per vector index, joined into the same
+/// scan rather than each index wrapping its own nested [`VectorScanTableProvider`]
+/// layer.
+///
+/// Every index on one [`VectorScanTableProvider`] is assumed to share the same
+/// primary key: [`Self::try_new`] takes it from the first index and rejects
+/// construction if another index disagrees, so this never silently joins on the
+/// wrong key for an index whose own key happened to differ.
 #[derive(Debug, Clone)]
 pub struct VectorScanTableProvider {
     pub table_provider: Arc<dyn TableProvider>,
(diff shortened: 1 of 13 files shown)


crates/search/src/index/vector_table.rs
```diff
@@ -168,21 +193,66 @@ impl VectorScanTableProvider {
...
+    /// query actually needs.
+    fn index_is_needed(
+        entry: &VectorIndexJoin,
+        base: &Arc<dyn TableProvider>,
+        columns_requested: &HashSet<String>,
+        filters: &[Expr],
+    ) -> bool {
+        let unique = Self::unique_columns_of(entry, base);
+        if columns_requested.iter().any(|c| unique.contains(c)) {
+            return true;
```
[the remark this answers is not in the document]
krinart: Done

crates/search/src/index/vector_table.rs
Jeadie, line 1: We have the assumption that all indexes on a single table have a shared primary key. This makes the `VectorIndexJoin` redundant. We could just have it like this

```diff
#[derive(Debug, Clone)]
pub struct VectorScanTableProvider {
    pub table_provider: Arc<dyn TableProvider>,
-   pub vector_index_list: Arc<LogicalPlan>,
+   pub index_list_plans: Vec<Arc<LogicalPlan>,
    pub primary_key: Vec<String>,
}
```

krinart: done, thanks

sgrebnov (approved):

PR #13226, opened 2026-08-18
fix(search): use one encoding path for full-text index writes and deletes

### LINKED ISSUE #12235: Tantivy primary-key term encoding mismatches for Float32, binary, and Arrow view types
## Problem

Updates in the tantivy full-text index are delete-then-insert. The delete term and the indexed value take different encoding paths and disagree for several Arrow primary-key types.

- **Float32/Float16** (`crates/search/src/generation/text_search/util.rs:93-110`): the delete term uses `f64::from(f32)`. The indexed value is the f64 parse of arrow-json's shortest-f32 string. For values such as `0.1` the two f64 values differ, so the delete matches nothing. Old and new documents both remain — duplicate and stale search hits, data-dependent.
- **Binary** (`crates/search/src/generation/text_search/index.rs:561-565`, `util.rs:241-256`): arrow-json encodes binary columns as hex strings; tantivy parses JSON strings on bytes fields as base64. Indexing fails on odd-length hex, or stores wrong bytes. Delete terms built from raw bytes match nothing.
- **Utf8View/BinaryView** (`index.rs:554-565` vs `util.rs:259-263`): the schema builder accepts view types, but `array_to_terms` rejects them with `NotYetImplemented`. Every update fails loudly and the index never receives data.

## Failure scenario

Table with a `Float32` primary key and value `0.1`. Every update leaves the old document in the index. Search returns duplicate rows and BM25 statistics drift.

## Required behavior

For every accepted primary-key type, indexing and deletion must use identical, lossless term encodings. Types that cannot satisfy this contract must be rejected before indexing begins.

## Possible implementation

Derive delete terms and indexed values through one shared encoding path per type. Reject unsupported key types at index creation, not at first write.

---
This issue comes from an automated audit of the full-text search system (trunk `5ebefa4fb`). Reverify the finding against current code before you start a fix.

Jeadie:
## Summary

Full-text index upserts are a delete-then-insert. This means the delete and insert must encode primary keys the same way. They did not for three primary-key types:

- **Float32/Float16**: the delete term came from a direct `f32 -> f64` cast; the inserted value went through arrow-json's shortest-string encoding of the `f32`, re-parsed as `f64`. The two produce different `f64` bit patterns for the same value (for example `0.1`), so the delete never matched what was indexed. Old and new rows both stayed in the index.
- **Binary**: arrow-json writes `Binary` columns as hex strings, but tantivy's JSON parser reads a bytes field as base64. This could fail to parse, or index the wrong bytes.
- **Utf8View/BinaryView**: the index schema builder accepted these types, but the delete-term encoder did not implement them, so every write on such a key failed.

Fixes #12235.


phillipleblanc (dismissed):

krinart (approved):

claudespice:
## This PR is `UNMERGEABLE` at merge-queue position 1, and the cause is external to it

Flagging this because the stall is not visible from the PR itself: every signal here is green — `MERGEABLE`/`CLEAN`, approved, no failing check on the PR head — but the queue entry has been `UNMERGEABLE` since it was enqueued at 01:04:45Z, and the speculative chain is linear, so **nothing behind it can land**. Positions 2–6 (#13478, #13522, #13547, #13551, #13563) are all built on this entry's merge commit, and trunk has had no landing since 23:14Z.

**What actually failed.** The `merge_group` batch at 01:24Z failed `Integration Tests (part 2)` and `(part 3)`:

```
test catalog::spiceai_integration_test_catalog ... FAILED
thread 'catalog::spiceai_integration_test_catalog' panicked at crates/runtime/tests/catalog/mod.rs:56:21:
Timeout waiting for components to load
Summary [760.298s] 171 tests run: 170 passed, 1 failed, 403 skipped
```

The panic is a timeout, but the cause above it is an upstream `500 Internal Server Error` from the Spice Cloud catalog endpoint while listing namespaces — repeated identically across all six retries (`TRY 6 FAIL`). One test failed out of 171; the other 170 passed.

**Why this is not yours to fix.** This PR changes the full-text index encoding path; it does not touch catalog registration, the Spice Cloud connector, or anything that endpoint reaches. The failing test exercises a live external service that was returning 500s during that window, so the failure is independent of the diff — the same test runs in every batch and would fail the same way for whichever PR held position 1.

**Suggested remedy:** a single requeue once the endpoint is healthy again — not a fix commit here, and not a change to the test. If it fails a second time on the same test with the endpoint healthy, that would be worth a real look; a repeat with the same upstream 500 signature would not.

I have not dequeued or requeued anything: this is your PR, the failure is transient and external, and ejecting position 1 would not help while the outage persists — the next PR to take the position would hit the identical failure.

PR #13233, opened 2026-08-18
feat(search): add relative score fusion as a candidate aggregation strategy

### LINKED ISSUE #5847: Implement Relative score fusion.
Implement relative score fusion in [CandidateAggregation](https://github.com/spiceai/spiceai/blob/trunk/crates/search/src/aggregation.rs#L64) pattern. WIll need:
 - Minmax normalisation on SearchGeneration candidates.
 - Expose weights to use in interpolation (i.e. since scoring is linear combination of input scores, need to decide on coefficients).

eastagiletracker:
This PR proposes adding relative score fusion as a second `CandidateAggregation` strategy, so hybrid search results can reflect how well each candidate scored and not only where it ranked (fixes #5847). We include this PR work along with a full history of your repo at https://eastagiletracker.com/projects/373. You can sign in with your GitHub ID to claim ownership of the project.

## What changed, and why

`ReciprocalRankFusion` is the only `CandidateAggregation` on `trunk` today, and it reads only each candidate's position within an input — the similarity score itself never reaches the fused total. That leaves one class of result unreachable: the document that scored near the top of *every* retrieval leg. Two legs ranking `A, B, C` and `C, B, A` put B second in both, yet rank-only fusion ranks it last, because rank 2 twice loses to rank 1 plus rank 3. No value of `k` changes that — the scores that would separate them are not inputs to the formula.

Reproduced against your default branch at `6ca34aa`:

```
$ git rev-parse HEAD
6ca34aa9b1fcf136dec42139fe5814670b015e7c
$ grep -rn "impl CandidateAggregation" crates --include='*.rs'
crates/search/src/aggregation/reciprocal_rank.rs:116:impl CandidateAggregation for ReciprocalRankFusion {

# your shipped reciprocal_rank_fusion_scores on those two legs, DEFAULT_RRF_K:
A=0.0322664585  B=0.0322580645  C=0.0322664585
```

`RelativeScoreFusion` min-max normalises each input's scores onto `[0, 1]` and combines them as a weighted linear sum, which keeps that distance: the same two legs now score B at 1.85 against 1.0 for A and C. `with_weights` sets a per-input coefficient so one retrieval leg can be favoured over another, and inputs past the end of the weight list keep the default weight of 1.0 rather than being silently zeroed. Alongside the aggregation there are pure `min_max_normalize` and `relative_score_fusion_scores` helpers mirroring the existing rank-fusion ones, so the tool registry's style of non-DataFusion fusion has the same option available.

Four decisions are worth your eye, since each could reasonably have gone the other way:

- An input whose candidates are all tied normalises to `1.0`, not `0.0` — every candidate is equally the best that input offered, and normalising to zero would drop a tied leg out of the sum entirely, which is a different statement.
- Normalisation windows the whole input, across every batch it produces, so a batch boundary cannot move a candidate's normalised score. There is a test for exactly this.
- Non-finite scores are excluded from the span rather than allowed to poison it, and normalise to `0.0`.
- `candidate_pool_size` stays at the trait default (`limit`) instead of RRF's 4x widening, because here the normalisation window *is* the candidate set — a wider pool would move `min`/`max` and change the fused scores rather than only extending the list. This is the one call I would most welcome a second opinion on; if you would rather it match RRF's widening, it is a one-line override.

Normalised scores are computed once in Rust over the already-materialised candidate batches and carried into the fused plan as a column, so the pure helper and the executed plan cannot drift apart — there is only one implementation of the arithmetic.

Five shared candidate-table helpers (`result_from_table`, `verify_schema_compatibility`, `additional_columns_of_schema`, `ith_search_value_column`, `are_types_compatible`) move from `reciprocal_rank` up into `aggregation`, where `from_single_input` already lives, so both strategies use one copy instead of a duplicate. The move is verbatim and the existing reciprocal rank tests are untouched.

The change is additive. `ReciprocalRankFusion` is unchanged and remains what every caller constructs today; `SearchPipeline` is already generic over the strategy, so nothing shifts for existing search paths unless a caller opts in.

## Verification

Scoped to `crates/search`, with `--no-default-features`, on this machine:

```
# baseline, clean tree at 6ca34aa
$ cargo test -p search --no-default-features --lib
test result: ok. 89 passed; 0 failed; 0 ignored

# with this change
$ cargo test -p search --no-default-features --lib
test result: ok. 104 passed; 0 failed; 0 ignored

$ cargo clippy -p search --no-default-features --tests -- -D warnings   # exit 0
$ cargo fmt -p search -- --check                                        # clean
$ cargo check -p search --tests                                         # default features, clean
```

No test that passed before fails now, and none were removed. The 15 new tests cover normalisation (span, ties, non-finite, empty), the weighted fusion semantics, the head-to-head against your shipped `reciprocal_rank_fusion_scores`, end-to-end `aggregate()` through DataFusion (ordering, candidates missing from an input, the multi-batch normalisation window, weights, single-input passthrough), and the error paths for a missing primary key and a missing score column.

To confirm the tests actually pin the behaviour rather than restating it, I disabled only the min-max branch and left every declaration in place: 9 of the 15 go red, while the 6 that assert null cases, error paths and the tied-input rule stay green.

One scope note: this adds the strategy and the weights the issue asks for, and stops there. Selecting it per request or per spicepod is a product decision about config surface and defaults that seemed yours to make, and reads to me the same way it read to the person who asked about scope on the issue.

## How this was managed

We tracked this work on a live agile board imported from your repository's own issues, pull requests and milestones: the story for this change is [Implement Relative score fusion.](https://eastagiletracker.com/projects/373/stories/241293), on the board at https://eastagiletracker.com/projects/373. The import brought over 12,101 stories and 94 labels.

board

If you'd rather not receive contributions like this, reply `no-more-prs` on this pull request and we won't open any further ones on your repositories.

---

**Lawrence W. Sinclair**
CEO / East Agile
[linkedin.com/in/lwsinclair/](https://linkedin.com/in/lwsinclair/)
[eastagile.com](https://eastagile.com)

### DIFF
--- a/crates/search/src/aggregation.rs
+++ b/crates/search/src/aggregation.rs
@@ -16,16 +16,24 @@ limitations under the License.
 #![allow(clippy::doc_markdown)]
 
 use std::collections::HashMap;
+use std::sync::Arc;
 
-use arrow::{array::RecordBatch, datatypes::SchemaRef};
+use arrow::{
+    array::RecordBatch,
+    datatypes::{DataType, SchemaRef},
+};
 use async_trait::async_trait;
-use datafusion::{common::Column, execution::SendableRecordBatchStream};
+use datafusion::{
+    common::Column, execution::SendableRecordBatchStream, prelude::SessionContext,
+    sql::TableReference,
+};
 use serde_json::{Value, json};
 use snafu::{ResultExt, Snafu};
 
 use crate::{SEARCH_SCORE_COLUMN_NAME, SEARCH_VALUE_COLUMN_NAME, VectorSearchGenerationResult};
 
 pub mod reciprocal_rank;
+pub mod relative_score;
 
 #[derive(Debug, Snafu)]
 pub enum Error {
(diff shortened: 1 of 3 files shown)


Jeadie:
no-more-prs

Issue #13291, opened 2026-08-19
Only the service manifest is pinned to the locked config directory: identity, journals and project deletion still resolve by pathname on macOS


claudespice:
## Problem

`MutationLock` retains the config directory open, and `MutationLock::pinned_directory()` hands that descriptor out so a locked operation can name the instance by inode. The service manifest uses it. Every other piece of Cloud Connect state still reaches for a pathname.

On Linux that pathname is `/proc/self/fd/<n>`, so a join through it stays rooted at the retained inode and the distinction does not matter. On macOS `descriptor_relative_config_dir()` returns the canonical *pathname* after a one-time `ensure_directory_stable()` check, because `/dev/fd/<n>` cannot be traversed. Each later use re-resolves that name.

The state still reached by name on macOS:

- `remove_identity` (`bin/spice/src/commands/connect/mod.rs`) — `identity.json`, the secret cache, the endpoint binding, both journals, and the `RuntimeLock`, plus the Cloud-side `delete_attached_project` that follows the identity it loaded.
- `execute_connect_transaction` (`bin/spice/src/commands/connect/transaction.rs`) — the identity it loads and validates.
- `spice connect service install` (`bin/spice/src/commands/connect/service/cli.rs`) — `validate_service_identity`, which builds every state path from `CloudConnectConfig::from_env_at`.
- `bin/spiced/src/cloud_connect.rs` — the delivered-secret cache write and the deployment apply.

## Why it matters

The config directory's owner can replace that pathname with another directory they also own once the lock's identity check has passed. Ownership cannot tell the two apart, so:

- A `spice connect remove` for instance A can load B's identity and delete **B's** Cloud project while uninstalling A's service. The project deletion on the wrong instance predates the manifest work; what is new is that the two halves of one transaction can now resolve to different directories.
- A `spice connect service install` can validate B's identity and record `SPICE_CONFIG_DIR` for B, while the manifest is published into A. The installed service stays manageable — `adopt_installed_definition` reconstructs a manifest from the definition when the file is absent, and its gate compares the same pathname on both sides — but A keeps an orphan manifest, and the adopted record carries no `runtime_digest` or `runtime_version` until the next install.

This is not a privilege boundary: the actor owns the state in question either way. It is the same correctness defect as #13204 — a locked operation applying to an instance other than the one the lock was taken for — for the state that fix did not cover.

## What would close it

[...]

PR #13292, opened 2026-08-19
fix(cloud-connect): resolve locked service state through the retained directory descriptor (fixes #13204)

### LINKED ISSUE #13204: MutationLock hands out a re-openable pathname on macOS, so a privileged install can act on the wrong instance
# Problem

`MutationLock::descriptor_relative_config_dir` gives a weaker guarantee on macOS than on Linux, and a privileged caller now depends on it.

On Linux it returns `/proc/self/fd/<n>`, so every subsequent open resolves through the descriptor the lock retained — the protected inode is pinned for the lifetime of the operation. On macOS there is no equivalent path, so the function performs a one-time `ensure_directory_stable()` and returns the canonical *pathname* (`crates/runtime-cloud-connect/src/mutation_lock.rs`, the `#[cfg(not(target_os = "linux"))]` branch). A pathname can be re-resolved, so the identity check and the later open are separated in time.

That trade was deliberate and is recorded in the function's own doc comment. What has changed is that it now has a privileged caller: `sudo spice connect service install` / `uninstall` performs service-state operations as root on macOS, so the interval between the check and the open is now crossed by a root process.

## What is already closed

`open_pinned_manifest_directory` does not simply open the pathname. It canonicalizes, walks the path one component at a time with `openat(..., O_NOFOLLOW)` starting from `/`, and then requires the resulting directory to be a real directory owned by the service owner.

So a symlinked component is refused, and a privileged operation cannot be steered into a directory the operator does not own. The escalation shape this pattern is normally associated with does not apply.

## The residual

The config directory's owner can replace it, between the identity check and the privileged open, with another directory **they also own**. The ownership test still passes, so a root-run install or uninstall can act on service state belonging to a different instance than the one the lock was taken for.

This does not cross a privilege boundary: the actor is the owner of the state in question, and they can already operate on their own instances directly. It is a correctness defect — the lock stops guaranteeing which instance a privileged operation applies to — rather than an escalation.

Linux is unaffected; its descriptor-rooted traversal has no such interval.

# Proposed Solution

Carry the retained directory descriptor through `MutationLock` into the service layer and perform the state operations against that descriptor with the `openat` primitives that already exist there, instead of handing out a re-openable pathname on non-Linux targets.

That makes the macOS path structurally equivalent to the Linux one — identity is held by the descriptor rather than re-derived from a name — and removes the interval rather than narrowing it.

# Scope & Sequencing

This belongs with whoever owns `mutation_lock` and the enrollment state path, not with an individual caller. Fixing it per-caller would leave the weaker primitive in place for the next one, and the file is under active change.

The macOS service lifecycle that surfaced this is a caller of the existing behaviour, not the cause of it; it does not need to be blocked on this.

# Acceptance Criteria

- [ ] Non-Linux state operations resolve through the retained directory descriptor rather than a pathname re-resolved after the check.
- [ ] A privileged install/uninstall acts on the instance the lock was taken for, even if the config directory is replaced by another directory with the same owner after the lock is acquired.
- [ ] The Linux behaviour and its existing coverage are unchanged.
- [ ] Coverage exists on macOS, not only under `#[cfg(target_os = "linux")]`.

claudespice:
## Summary

`MutationLock::descriptor_relative_config_dir` gave a weaker guarantee on macOS than on Linux. On Linux it returns `/proc/self/fd/<n>`, so every later lookup stays rooted at the directory inode the lock opened. There is no equivalent on macOS — `/dev/fd/<n>` cannot be traversed as a directory — so the function performed a one-time `ensure_directory_stable()` and returned the canonical *pathname*, which each later use re-resolves.

That interval now has a privileged caller: `sudo spice connect service install` / `uninstall` performs service-state operations as root on macOS. `open_pinned_manifest_directory` already refuses a symlinked component and requires the resulting directory to be owned by the service owner, so a privileged operation cannot be steered into a directory the operator does not own. The residual is narrower: the config directory's owner can replace it, between the identity check and the privileged open, with **another directory they also own**. The ownership test still passes, so a root-run install or uninstall acts on service state belonging to a different instance than the one the lock was taken for.

This does not cross a privilege boundary — the actor owns the state in question either way. It is a correctness defect: the lock stops guaranteeing *which* instance a privileged operation applies to.

## Changes

- `MutationLock::pinned_directory()` duplicates the directory descriptor the lock already retains. It carries no pathname, so an operation performed relative to it applies to the locked inode on every Unix, not only where `/proc/self/fd` supports directory traversal.
- `PinnedConfigDir` carries that descriptor into the service layer. `ServiceManifest::load` / `write` / `remove` and the config-directory check in `validate` resolve through it; a caller holding no lock still reaches the directory by pathname, which is all an unlocked read has.
- The manifest read is descriptor-relative (`openat` with `O_NOFOLLOW`), keeping the regular-file, single-link and size bounds the pathname read applied.
- `is_retained_descriptor_path` is gone: no call site passes a `/proc/self/fd/<n>`-shaped pathname into the manifest layer any more, so a path of that shape no longer receives special trust.
- Linux keeps `descriptor_relative_config_dir` for the state it reaches by joining (identity, journals), so its behaviour is unchanged.

## Test plan

- `make lint`
- `make nextest`
- New coverage runs on macOS, not only under `#[cfg(target_os = "linux")]`, which is where the guarantee was missing:
  - `the_retained_descriptor_reaches_the_locked_state_after_a_same_owner_replacement` (`runtime-cloud-connect`) — the descriptor still reaches the locked state after the pathname it was acquired through names a different same-owner directory.
  - `a_pinned_directory_is_the_locked_inode_after_a_same_owner_replacement` — a pinned manifest operation opens the locked inode, with the replacement asserted to be a different directory owned by the same account, so ownership cannot discriminate.
  - `a_locked_manifest_is_published_and_read_back_from_the_locked_directory` — the write and the read-back both land in the locked directory.
  - `a_locked_manifest_is_read_and_removed_after_its_pathname_is_gone` — nothing a locked operation does depends on the pathname still resolving.
  - `locked_uninstall_cannot_be_redirected_by_config_path_replacement` was `#[cfg(target_os = "linux")]` because the guarantee did not hold elsewhere; it is now `#[cfg(unix)]` and passes on macOS.
- Each new test was checked against a neuter matrix on macOS — reverting the descriptor in the open, in the read, in `validate`'s directory stat, and in `pinned_directory` itself. Every neuter fails at least one test and every test is failed by at least one neuter; the `validate` stat is guarded only by the pathname-gone test.

## Review gate

- Adversarial review: `codex` — job `review-mt0fhsza-i0rjo2` (`exit=0`, verdict `needs-attention`). Two `[high]` findings, both the same observation: the fix pins the manifest while identity, journals, the endpoint binding and the Cloud-side project deletion still resolve by pathname on macOS, so one transaction can span two directories.
  - Deferred, not dismissed → #13291, which names each remaining call site and both consequences. The engine's recommendation to "fail closed on non-Linux rather than mixing" would disable the macOS service lifecycle that #13203 just shipped, which is a worse outcome than an owner-only inconsistency window; and for `connect remove` the harmful half (deleting another instance's Cloud project) is pre-existing and untouched here, while the half this PR covers becomes correct.
  - One stated consequence is refuted: the split does **not** leave an unmanageable service. `adopt_installed_definition` reconstructs a manifest from the installed definition when the file is absent, and its gate compares `installed.config_dir` against the resolved config dir — both the same pathname in this scenario — so `status`/`stop`/`uninstall` keep working. The real residual is an orphan manifest in the moved directory and an adopted record carrying no `runtime_digest`/`runtime_version` until the next install. #13291 says so.
  - The asymmetry is recorded at the site it lives (`service/cli.rs`), citing #13291.
- `/security-review`: no findings at or above the reporting threshold. It verified check-by-check that the old pathname walk dropped nothing — `O_NOFOLLOW`, the regular-file/`nlink`/size bounds, `ensure_owner_only`, the pinned-directory ownership gate and the `euid` gate are all retained, `O_CLOEXEC` is added, and the descriptor is `F_DUPFD_CLOEXEC` so it cannot leak to the unprivileged supervisor children. It independently raised the same scope note, now #13291.
- `/simplify`: four cleanup agents (reuse, simplification, efficiency, altitude). Applied: dropped `is_pinned()`, whose two call sites both sat inside `#[cfg(unix)]` blocks — on a Windows build it was dead code and `-Dwarnings` would have failed; `remove_manifest_in_directory` now shares the bounded read with the load path instead of keeping a second copy that had already drifted on `O_NONBLOCK`; one `PinnedConfigDir::for_lock` constructor replaces the same platform split written out at two call sites; the three descriptor-relative syscalls take the manifest name from a constant tied to `SERVICE_MANIFEST_FILE` by a test, rather than three literals; `path()` became `display_path()` to say what it is for; and a dead `NotFound` arm, a no-op rebinding and a doubly-built path are gone. The altitude pass also caught two comments claiming more than the code does — the descriptor-relative read does resolve its final component by name (`O_NOFOLLOW` is what makes that safe), and one sentence about `descriptor_relative_config_dir` parsed as the opposite of its intent — plus an intra-doc link from an ungated item to a `#[cfg(unix)]` one that would warn on a Windows doc build. Skipped: moving `PinnedConfigDir` down into `runtime-cloud-connect` and giving it the `openat` operations, which is the right shape but belongs with #13291 rather than widening this diff; extracting the bounded read into `state.rs` for all six state files, same reason; and collapsing `service_config_dir`/`state_config_dir`, which still carry different values at the unlocked call site.

  The neuter matrix was re-run after these edits and still discriminates: five neuters (the descriptor ignored in the open, in the read, in `validate`'s stat, in `pinned_directory` itself, and the manifest-name constant diverged) each fail at least one test, and every new test is failed by at least one neuter.

Fixes #13204

## Checklist

- [x] Tests added that fail on the old code and pass on the new
- [x] New coverage runs on macOS, the affected platform
- [x] Linux behaviour and its existing coverage unchanged
- [x] Deferred findings filed as #13291 and linked

### DIFF
--- a/bin/spice/src/commands/connect/mod.rs
+++ b/bin/spice/src/commands/connect/mod.rs
@@ -526,7 +526,7 @@ async fn start_instance(ctx: &RuntimeContext, dir: Option<&Path>, verbosity: u8)
     let Some(manifest) = service::resolve_with_state(
         backend,
         &instance_dir,
-        &state_config_dir,
+        &service::PinnedConfigDir::unlocked(&state_config_dir),
         &service_config_dir,
     )?
     else {
@@ -735,6 +735,14 @@ async fn remove_identity(
         .map_err(|source| Error::CloudConnectIo {
             message: format!("pin locked Cloud Connect state for removal: {source}"),
         })?;
+    // The service manifest is resolved and removed through the descriptor the
+    // lock retains, not through a name (see `PinnedConfigDir`). The identity,
+    // journals, endpoint binding and cache below are still derived from
+    // `config_dir`, which on a platform without Linux's descriptor-rooted
+    // traversal is a pathname re-resolved on each use, so one removal can still
+    // span two directories. Closing that is #13291.
+    let manifest_config_dir =
+        service::PinnedConfigDir::for_lock(display_config_dir.clone(), &mutation_lock)?;
     let _instance_lock =
         runtime_cloud_connect::RuntimeLock::acquire(&config_dir).map_err(|source| {
             Error::CloudConnectIo {
@@ -776,8 +784,12 @@ async fn remove_identity(
                 });
             }
         };
-    let installed =
-        service::resolve_with_state(backend, instance_dir, &config_dir, &display_config_dir)?;
+    let installed = service::resolve_with_state(
+        backend,
+        instance_dir,
+        &manifest_config_dir,
+        &display_config_dir,
+    )?;
 
     // `symlink_metadata` observes the directory entry itself. A dangling link
     // is state that `--force` must remove, not an absent file that can be left
@@ -968,13 +980,15 @@ async fn remove_identity(
     // `spice connect service uninstall`, while identity cleanup belongs only to
     // this command.
     let uninstall_failure = match installed.as_ref() {
-        Some(manifest) => match service::uninstall_resolved(backend, manifest, &config_dir) {
-            Ok(()) => {
-                println!("Stopped and uninstalled {}.", manifest.name);
-                None
+        Some(manifest) => {
+            match service::uninstall_resolved(backend, manifest, &manifest_config_dir) {
+                Ok(()) => {
+                    println!("Stopped and uninstalled {}.", manifest.name);
+                    None
+                }
+                Err(err) => Some(err),
             }
-            Err(err) => Some(err),
-        },
+        }
         None => None,
     };
 
--- a/bin/spice/src/commands/connect/service/cli.rs
+++ b/bin/spice/src/commands/connect/service/cli.rs
@@ -202,23 +202,34 @@ pub async fn execute(
                     "validate locked Cloud Connect state for service {action}: {source}"
                 ),
             })?;
-        let state_config_dir =
+        let identity_config_dir =
             lock.descriptor_relative_config_dir()
                 .map_err(|source| Error::CloudConnectIo {
                     message: format!(
                         "pin locked Cloud Connect state for service {action}: {source}"
                     ),
                 })?;
-        Some((service_config_dir, state_config_dir))
+        Some((service_config_dir, identity_config_dir))
     } else {
         None
     };
-    let (service_config_dir, state_config_dir) = locked_dirs.as_ref().map_or(
+    let (service_config_dir, identity_config_dir) = locked_dirs.as_ref().map_or(
         (config_dir, config_dir),
-        |(service_config_dir, state_config_dir)| {
-            (service_config_dir.as_path(), state_config_dir.as_path())
+        |(service_config_dir, identity_config_dir)| {
+            (service_config_dir.as_path(), identity_config_dir.as_path())
         },
     );
+    // Manifest state resolves through the descriptor the lock retains, not
+    // through a name (see `PinnedConfigDir`).
+    //
+    // `identity_config_dir` above is still a pathname on a platform without
+    // Linux's descriptor-rooted traversal, so identity validation and the
+    // manifest can resolve to different directories in one install. Closing
+    // that needs descriptor-relative reads through the whole state layer: #13291.
+    let state_config_dir = match mutation_lock.as_ref() {
+        Some(lock) => super::PinnedConfigDir::for_lock(service_config_dir, lock)?,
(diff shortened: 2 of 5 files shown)


claudespice:
@copilot review

phillipleblanc (approved):

PR #13318, opened 2026-08-19
test(object-store): serve the GitHub store's tests from a loopback stub (fixes #13206)

### LINKED ISSUE #13206: test_get_opts fetches from raw.githubusercontent.com during sign-off, so a GitHub rate limit fails the gate
# Problem

`runtime-object-store store::github::tests::test_get_opts` reaches the public internet during the sign-off gate. It constructs a `GitHubRawObjectStore` against `spiceai/spiceai` at `refs/heads/trunk` with no token, fetches `https://raw.githubusercontent.com/spiceai/spiceai/refs/heads/trunk/README.md`, and then asserts that a live listing of `docs/release_notes/rc` is non-empty.

Because it runs in the `nextest` phase of `scripts/signoff`, a rate limit or network fault on GitHub's side fails the gate for a change that has nothing to do with it.

**The content it depends on is intact** — `docs/release_notes/rc` exists on `trunk` with 5 entries. Nothing is missing; this is purely about reaching the network.

## Why the retry budget cannot absorb it

The failure is not transient at the scale the harness retries over. The object store itself retries 10 times internally, taking about 132 seconds before it gives up, and then `nextest` spends its whole six-retry budget re-running that:

```
TRY 1 FAIL [ 132.475s] runtime-object-store store::github::tests::test_get_opts
TRY 2 FAIL [  40.578s] …
TRY 6 FAIL [  45.485s] (11255/11255)

failed to get README: HTTP 429 Too Many Requests
  https://raw.githubusercontent.com/spiceai/spiceai/refs/heads/trunk/README.md
  retries: 10/10, elapsed: 132.44s
```

So one throttled endpoint costs several minutes and still fails closed.

## What was actually happening when this was observed

The failures on 2026-08-17 were concurrent with a GitHub incident — opened 13:40:03Z, indicator `major`, "Partial System Outage" — whose own status update reads:

> Archive downloads and **raw repository content downloads** are experiencing an approximate **50% error rate**. […] high error rates around 20% for web experiences and api traffic.

`raw.githubusercontent.com` is exactly the raw-repository-content surface this test depends on.

Measured at 14:22Z: `HTTP 429` for that URL from all four sign-off hosts *and* from a separate developer machine on a different network. Simultaneous failure across unrelated networks is the signature of a provider-side event.

None of this weakens the defect being reported — it is the clearest possible demonstration of it. A third party had a multi-hour window at a 50% error rate, and during that window this one test could fail any sign-off in the repository regardless of what the change touched. In one run, **11,254 of 11,255 tests passed** and the gate failed on this test alone.

# Proposed Solution

Deliberately not prescribing one — the right answer belongs to whoever owns this crate. The options that would each remove the failure mode:

- Serve the fixtures from a mock or recorded responses, so the test asserts the store's behaviour rather than GitHub's availability.
- Require an authenticated token, so the request is subject to a per-token limit rather than a shared anonymous one, and skip when the token is absent.
- Move it out of the sign-off suite into a network-tagged tier the gate skips, keeping the live coverage where a failure means "the network is down", not "this PR is bad".

# Acceptance Criteria

- [ ] `make nextest` / `scripts/signoff` cannot fail because of a third-party rate limit or outage.
- [ ] Whatever coverage of the GitHub object store is kept still fails loudly when the store itself is broken.
- [ ] A sign-off failure is attributable to the change under test rather than to a third party's availability.

claudespice:
## Summary

`runtime-object-store store::github::tests::test_get_opts` reached the public internet from inside the sign-off gate. It fetched `README.md` from `raw.githubusercontent.com` and then listed `docs/release_notes/rc` through `api.github.com`, so a rate limit or outage at GitHub failed `make nextest` — and therefore the gate — for a change that had nothing to do with it. The retry budget made it worse rather than absorbing it: the object store retries 10 times (~132s) before giving up, and `nextest` then spent its whole six-retry budget re-running that. In the run that prompted the issue, 11,254 of 11,255 tests passed and the gate failed on this one test.

The content it depended on was never missing; the only problem was reaching the network for it.

## Changes

The store hardcoded both GitHub base URLs at their point of use — the raw base inside `try_new`, the API base inside `fetch_git_tree`. Both are now supplied by a `GitHubEndpoints` value, so a test can redirect them at a loopback stub while the paths built on top of them stay the production paths:

- `GitHubEndpoints::github_com()` is what every non-test caller gets, via the unchanged public `try_new`. `raw_url` and `git_tree_url` build the revision-scoped raw prefix and the recursive git-tree endpoint, and both production and test go through them, so the tests exercise the real URL construction rather than a copy of it.
- `fetch_git_tree` takes the endpoint it should call instead of rebuilding it from `org`/`repo`/`rev`. Component validation still runs before any URL is constructed.
- `allow_http` is pinned `false` in `github_com()`. `GitHubEndpoints` and `try_new_with_endpoints` are module-private with no caller outside `store::github`, so no production path can be talked into sending a token over cleartext, and `production_endpoints_are_https_and_forbid_cleartext` fails if that changes.

Coverage now asserts this store's behaviour instead of GitHub's availability: the revision-scoped raw path, the recursive git-tree endpoint and its `recursive=true` query, blob-vs-tree filtering, prefix scoping, `size` mapping, token propagation on both surfaces, and how a rejected API surfaces to the caller. The live test is kept as `live_github_serves_the_repository_contents`, `#[ignore]`d — the gate passes no `--run-ignored`, so it never runs there, and it is still available on demand with `cargo test -- --ignored`.

Scope: `test_get_opts` was the only test in the gate that reached the public internet. The gate's filterset is `kind(=lib) + kind(=proc-macro) + (package(=cayenne) & kind(=test)) + (package(=runtime-cloud-connect) & kind(=test)) + binary(=metrics)`, which excludes the `httpbin.org` and `api.tvmaze.com` tests living in `tests/` targets — those are a real but separate exposure, and not one this gate can fail on.

Gate time for these tests goes from a 132s-plus worst case (times six retries) to 0.57s.

## Test plan

- `make lint` — clean.
- `make nextest` — the crate's 13 `store::github` tests pass and the live test is skipped as ignored.
- Every new test was validated by defect injection: each of the 12 injected defects (drop the trailing-slash normalisation, drop the blob filter, drop the prefix filter, drop either surface's auth header, swallow the API error, drop the revision from the raw URL, drop `recursive=true`, zero the size, let production allow cleartext, remap `NotFound`, print the token) fails at least one test, and every test is failed by at least one defect — so no assertion in this diff is vacuous.

## Review gate

- Adversarial review: `codex` — job `review-mt0qpkse-8v61rk`. Verdict `approve`, no material findings; it independently confirmed that the public constructor stays pinned to the GitHub HTTPS URLs and that endpoint injection is module-private.
- `/security-review`: no HIGH or MEDIUM findings. It examined the three paths the new plaintext escape hatch could open — cleartext token transmission, SSRF through the injectable base, and token handling in the reworked `fetch_git_tree` — and each is closed by the private visibility plus the unchanged component validation. Analysis was performed directly rather than dispatched to sub-agents, as this session cannot spawn them.
- `/simplify`: applied two cleanups — the stub body is a `bytes::Bytes` (already a dependency) instead of an `Arc<Vec<u8>>` cloned per request, and the twice-duplicated path extraction became a `paths()` helper. Re-linting after those edits caught a `redundant_clone` that the pre-`/simplify` run could not have seen. Skipped reusing `spice-cloud-client`'s `test_support` stub: it is cloud-API-specific and in another crate, and the repo's own precedent (`enroll.rs`, `runtime-datafusion-udfs`, `connector-dynamodb`) is a per-crate inline stub. The four angles were reviewed directly rather than by sub-agents, for the reason above.

Fixes #13206

## Checklist

- [x] The sign-off gate can no longer fail because of a third-party rate limit or outage.
- [x] Coverage of the GitHub object store still fails loudly when the store itself is broken (a rejected API surfaces as an error; a 404 maps to `NotFound`).
- [x] A sign-off failure is attributable to the change under test rather than to a third party's availability.
- [x] Tests added, and each one proven to fail under an injected defect.
- [x] `make lint` clean.


claudespice:
@copilot review

bjchambers (approved):

Issue #13348, opened 2026-08-21
Postgres CDC echo-drop via transaction-id identity


Jeadie:
For a CDC-accelerated table atop a postgres source, any write-through from accelerator -> postgres will cause a subsequent echo back over CDC to the accelerator. This causes data correctness issues. Spice should filter out any transactions it writes to postgres (specifically the operations within a transaction that affect the given table; see _"Per-relation, not whole-transaction"_).

Within `postgres_replication` and `connector-postgres`, the connector registers the transaction id (`xid`) of each write-back it issues, and the CDC drops the arbitrated table's changes from any
replicated transaction. 

### On write
BEGIN, get xid, register in accelerator, commit write.
E.g. `BEGIN; SELECT pg_current_xact_id(); INSERT … ON CONFLICT (pk) DO UPDATE … ; COMMIT`
cannot ensure atomic update of accelerator and postgres transaction, apply xid to accelerator first.
 - On failure, we have stale xid in accelerator, gets cleaned up.
 - On sucess, get and store LSN to help with clean up (and u32 overflow protection)

### On read
- Read xid from `ReplicationEvent::Begin { final_lsn, xid, commit_time_micros }` and compare against accelerator.
- `TxnBuffer` is keyed by `RelationId`,  remove the table relation's entries, apply the rest. Still process empty relations for the ack_floor. Cannot commit immediately and bypass still fence-pending operations. 
- Do not clear xids here, must wait until durable in accelerator, otherwise on replay, we could replay these xids (this is why we store LSN on write path). 

### Persistence and GC
Store details in accelerator metadata table: `spice_sys_postgres_write_back_xids` like this
```rust
pub struct Row {
  xid8: u64,
  commit_lsn_upper_bound: Option<u64>, // estimate from write side. 
  observed_commit_lsn: Option<u64> // LSN when/if we receive the echo. Must wait until this is durable in accelerator.
}
```
On startup:
 - All rows where `pg_xact_status(xid8)='aborted'`
 - All rows where `confirmed_flush ≥ commit_lsn_upper_bound`. This is safe since we know `lsn_xid  <= commit_lsn_upper_bound`. 
 - All rows where `confirmed_flush  >= observed_commit_lsn`. This is the standard CDC path, but we are simply waiting for the change to be persisted, then we know it won't be replayed. 
 - All rows 2^31 older than current (with warning). 

## Version gate
[...]

Issue #11840, opened 2026-07-13
Conditional commits: CDC authority echo-drop + provider consumed-position mapping


phillipleblanc:
Part of #11830. Companion to #11838 (durable write-back).

With durable write-back on a `refresh_mode: changes` dataset, a write Spice delivers to the federated source is replicated **back** to the accelerator by the source's CDC — an **echo** of Spice's own write. Applied naively it double-applies, and a stale-value echo arriving after a newer local commit transiently regresses the accelerator (convergent but non-monotone).

**Technical:**
- **Authority classification.** Classify each participating table's keys as **arbitrated** (Spice-written, under a gated write-back) vs **reference** (read-only gate input). For arbitrated keys, CDC events that echo Spice's own delivered writes are dropped rather than applied.
- **Provider consumed-position mapping.** Map the source CDC position (Postgres LSN / Debezium-Kafka offset — the existing consumed-position checkpoint, `postgres_replication`, `debezium_kafka`) to the write-back **delivered position**, so echoes up to the delivered position are dropped while genuinely-new source changes still apply.
- **Delivery-ack seam.** Consume the delivery-ack callback #11838 exposes — `(table, pk_digest, marker_seq, source-position?)` fired after a successful source delivery — to advance the drop watermark for arbitrated keys.
- The CDC apply path must never write dirty write-back markers (marker emission is confined to local commit-publish transactions), so an echo cannot spawn a new delivery.

**Acceptance:** a write-back delivery's CDC echo does not double-apply, does not create a spurious pending-write-back marker, and does not regress a newer local value; non-echo source changes on reference tables still apply; no delivery/echo feedback loop.

PR #13351, opened 2026-08-21
feat(postgres): connector-owned durable write-back delivery with xid capture

Jeadie:
Part of #13348 (design) / #11840. Stacked on #13350 — review only this branch's diff.

## What this PR adds

The Postgres connector now delivers durable write-back rows itself, on a direct connection, in an explicit transaction. This lets it read the transaction id (`pg_current_xact_id()`) inside the same transaction as the upsert and register it in the `XidRegistry` (#13350) before it commits. A later PR uses the registered xids to drop the CDC echo of each write-back.

## How it works

- New `WriteBackDeliverer` trait in `data-connector-api`, with a defaulted `DataConnector::write_back_deliverer` accessor. Every known `DataConnector` wrapper forwards the method explicitly (deferred, drasi, embeddings, full-text, elasticsearch) — a defaulted method that a wrapper does not forward regresses silently (see #10460).
- The Postgres deliverer (`connector-postgres/src/write_back.rs`) brackets each delivery transaction: `BEGIN` → read the xid → persist it in the registry (abort delivery if this fails) → upsert (`INSERT … ON CONFLICT (pk) DO UPDATE`) → `COMMIT` → read `pg_current_wal_insert_lsn()` as the commit-LSN upper bound for cleanup. The absent-key `DELETE` leg is its own transaction with the same bracketing; every xid is registered.
- On an ambiguous `COMMIT` failure the entry stays registered — the transaction may have committed, and startup GC resolves it. Only an unambiguous server-side abort unregisters eagerly.
- The write-back worker uses the deliverer when the connector provides one. Connectors without one keep the existing `InsertOp::Replace` path unchanged, including the delete-then-insert fallback.
- Version gate: `pg_current_xact_id()` on PG13+, `txid_current()` on PG10-12.

## Behavior notes

- The worker's pass structure does not change: list markers → read committed values → partition present/absent keys → deliver → compare-and-clear only on full success. Delivery stays idempotent under full-pass replay.
- SQL generation reuses the `datafusion-table-providers` `InsertBuilder`/`OnConflict` utilities, so the upsert statement shape matches the previous sink path.

## Testing

Unit tests for the deliverer's SQL shape and transaction bracketing order, and new worker-level tests for present/absent partitioning, marker retention on delivery error, and deliverer-over-fallback selection.


crates/data-connectors/connector-postgres/src/lib.rs
```diff
@@ -1029,6 +1031,80 @@ impl DataConnector for Postgres {
...
+        &self,
+        context: &dyn ConnectorContext,
+        dataset: &DatasetSpec,
+    ) -> Option<Arc<dyn WriteBackDeliverer>> {
+        // Only a durable-write-back dataset delivers through the connector; every
+        // other dataset keeps the worker's `TableProvider` path (returns `None`
+        // here). Checked first so a non-durable dataset never opens a connection.
+        let on_conflict = durable_write_back_on_conflict(dataset)?;
+
+        // Version gate: `pg_current_xact_id()` is PG13+, `txid_current()` its
```
Jeadie: L1044-1075 could be standalone function

crates/data-connectors/connector-postgres/src/write_back.rs
```diff
@@ -0,0 +1,618 @@
...
+        message: message.into(),
+        source: Box::new(source),
+    }
+}
+
+/// Resolve, once at deliverer setup, which server function assigns and returns the
+/// delivery transaction's `xid8`. `pg_current_xact_id()` is PG13+; `txid_current()`
+/// is its PG10–12 equivalent (also epoch-qualified 64-bit).
+#[must_use]
+pub(crate) fn xid_function_for_server_version(server_version_num: i32) -> &'static str {
```
Jeadie: convert xid_function to enum, then implement this as conversion from i32.

crates/data-connectors/connector-postgres/src/write_back.rs
```diff
@@ -0,0 +1,618 @@
...
+    table: &TableReference,
+    batch: RecordBatch,
+    on_conflict: &OnConflict,
+    source_schema: &SchemaRef,
+) -> Result<String, Box<dyn std::error::Error + Send + Sync>> {
+    let batches = vec![batch];
+    let sea_query_on_conflict = on_conflict.build_sea_query_on_conflict(source_schema);
+    InsertBuilder::new(table, &batches)
+        .build_postgres(Some(sea_query_on_conflict))
+        .map_err(|e| Box::new(e) as Box<dyn std::error::Error + Send + Sync>)
```
Jeadie: should not return `Result<String, Box<dyn std::error::Error + Send + Sync>>`, should return error type from `build_postgres`.

crates/data-connectors/connector-postgres/src/write_back.rs
```diff
@@ -0,0 +1,618 @@
...
+    /// Source table schema, as the server reports it (not `LargeUtf8`-normalized).
+    schema: SchemaRef,
+    on_conflict: OnConflict,
+    /// The write path binds an empty constraint set on the source writer, so the
+    /// per-batch constraint validation is a no-op here too — kept to match the
+    /// path this replaces exactly.
+    constraints: Constraints,
+    registry: Arc<XidRegistry>,
+    /// `pg_current_xact_id` (PG13+) or `txid_current` (PG10–12).
+    xid_fn: &'static str,
```
Jeadie: as above, this should be enum.

crates/data-connector-api/src/write_back.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    /// Upsert every batch of `rows` into the source in **one** transaction. The
+    /// rows are already cast to [`target_schema`](Self::target_schema) by the
+    /// worker.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the transaction cannot be prepared, executed, or
+    /// committed; the worker retries the whole pass.
+    async fn deliver_upserts(&self, rows: Vec<RecordBatch>) -> DeliveryResult;
```
Jeadie, line 75: if we provide `session_state: &SessionState,` or similar here, we can remove the `cast_batches_to` to be an internal information of the Postgres `WriteBackDeliverer`, and can remove `WriteBackDeliverer::target_schema` from trait.

crates/runtime-table/src/accelerated/write_back_worker.rs
```diff
@@ -313,3 +359,136 @@ fn absent_claimed_keys(
 fn to_df_err(e: cayenne::provider::Error) -> DataFusionError {
     DataFusionError::Execution(format!("durable write-back: {e}"))
 }
+
+/// Map a connector-owned delivery failure into the worker's error type. The
+/// worker only needs the message for its log and its retry backoff — a failed
+/// pass replays whole, so the variant is immaterial.
+#[expect(
```
Jeadie, line 351: remove this and just use something like `DataFusionError::Context("durable write-back", Box<DataFusionError::Execution(e))`

crates/runtime-table/src/accelerated/write_back_worker.rs
```diff
@@ -313,3 +359,136 @@ fn absent_claimed_keys(
...
+    clippy::needless_pass_by_value,
+    reason = "passed to `Result::map_err`, which moves the error value in"
+)]
+fn delivery_to_df_err(e: data_connector_api::write_back::DeliveryError) -> DataFusionError {
+    DataFusionError::Execution(format!("durable write-back: {e}"))
+}
+
+/// Cast `batches` from `input_schema` to `target_schema`, collecting the result.
+///
+/// Mirrors [`execute_insert`]'s cast (`SchemaCastScanExec` to the source
```
Jeadie: `SchemaCastScanExec` uses `try_cast_to` from `crates/arrow_tools/src/record_batch.rs`.

phillipleblanc (approved):

PR #13381, opened 2026-08-22
fix(cloud): name the organization searched when link finds no projects

### LINKED ISSUE #13379: spice cloud link ignores the active organization and never says which one it searched
## Describe the bug

`spice cloud link` lists only the projects of the organization the *credential* belongs to. `spice cloud org use <org>` and `--org <org>` have no effect on it, and nothing in the output says so.

A user who administers several organizations therefore sees "no projects that can be attached" while `spice cloud orgs` lists those organizations and `spice cloud projects --org <org>` lists projects inside them. Every signal the CLI gives says the credential can reach those projects; linking silently disagrees.

The CLI's Cloud Connect client sends only a bearer token when it asks for attachable projects — no organization — so the server answers for the credential's own organization. Selecting an organization only changes the outcome indirectly, by causing a different stored per-organization credential to be picked.

## To Reproduce

1. Authenticate with a credential whose organization has no Cloud Connect projects, as a user who also administers another organization that has one.
2. `spice cloud orgs` — lists both organizations.
3. `spice cloud org use <other-org>` (or pass `--org <other-org>`).
4. `spice cloud link` — "Spice Cloud returned no projects that can be attached."

## Expected behavior

At minimum, say which organization was searched, so the message is diagnosable rather than misleading:

```
No attachable projects in organization '<org>' — the organization this credential acts on.
Selecting another organization requires a credential for it: spice cloud login token --org <org>.
```

Better, if the platform grows a way to name the organization on these endpoints: pass the selected organization through, so `org use` and `--org` mean the same thing for `link` that they mean everywhere else in `spice cloud`. That half depends on a server change and is tracked internally.

An empty list that means "your other organizations were not searched" is indistinguishable from "you have no eligible projects", and the CLI currently reports the second when the first is true.

## Runtime Details

`spice` v2.2.0, macOS arm64. Not runtime-related — CLI only.

Source: `execute_link` / `user_token_for_cloud_connect` in `bin/spice/src/commands/cloud/mod.rs`, and `ProjectClient::list_attachable` in `bin/spice/src/commands/connect/project.rs`.

Context: found while diagnosing #13374 (fix in #13376), and related to #13378. The organization-scoped credential model is deliberate; the silence about it is the bug.

yzxcj797:
Issue Number: close #13379

### What problem does this PR solve?

`spice cloud link` lists attachable projects **for the credential's own organization only** — `org use` and `--org` select a credential, they do not scope the request. The empty-list error reported `Spice Cloud returned no projects that can be attached.` with a hint to create a project, which is actively misleading for a user administering several organizations: `orgs` lists them all, `projects --org <org>` lists their projects, and linking silently disagrees. (#13379)

### What is changed and how it works?

The CLI-only half of the fix proposed in the issue:

- `user_token_for_cloud_connect` now returns the credential's organization alongside the token — it already fetches the auth context to validate each candidate token, and `AuthContext` carries `org_name`.
- `choose_attachable_project` takes that organization and, when the list is empty, errors with:

```
Spice Cloud returned no attachable projects in organization '<org>', the organization this credential acts on.
Create a project in organization '<org>', or attach to another organization with `spice cloud login token --org <org>` and retry.
```

Passing the selected organization through to the endpoint (so `org use` scopes the request like everywhere else) depends on a server change and stays tracked internally, per the issue.

### Validation

`cargo check -p spice` and `cargo fmt --check` clean. The changed paths are interactive CLI flows (TTY chooser) without unit-test scaffolding in this file; behavior verified by inspection of the two call sites (`link` naming the org, `unlink` ignoring it via `_credential_org`).

### DIFF
--- a/bin/spice/src/commands/cloud/mod.rs
+++ b/bin/spice/src/commands/cloud/mod.rs
@@ -2697,7 +2697,7 @@ async fn execute_link(ctx: &RuntimeContext, args: &LinkArgs, flag_org: Option<&s
         resolve_project_target(Some(project), flag_org).await?
     } else {
         let (requested_org, requested_source) = resolve_org_with_source(flag_org)?;
-        let token =
+        let (token, credential_org) =
             user_token_for_cloud_connect(&endpoint, requested_org.as_deref(), "Linking", "link")
                 .await?;
         let attach_client = crate::commands::connect::project::ProjectClient::new(&endpoint)
@@ -2710,7 +2710,7 @@ async fn execute_link(ctx: &RuntimeContext, args: &LinkArgs, flag_org: Option<&s
             .map_err(|error| Error::CloudConnectIo {
                 message: error.to_string(),
             })?;
-        let selected = choose_attachable_project(projects).await?;
+        let selected = choose_attachable_project(projects, &credential_org).await?;
         if let Some(selected_org) = selected.org.as_deref() {
             ensure_orgs_agree(
                 selected_org,
@@ -2728,7 +2728,8 @@ async fn execute_link(ctx: &RuntimeContext, args: &LinkArgs, flag_org: Option<&s
             "Pass an explicit <org>/<project> target.",
         )
     })?;
-    let token = user_token_for_cloud_connect(&endpoint, Some(org), "Linking", "link").await?;
+    let (token, _credential_org) =
+        user_token_for_cloud_connect(&endpoint, Some(org), "Linking", "link").await?;
     let management_client = CloudClient::with_token_for_org_at(token, Some(org), &endpoint)?;
     let project = management_client.get_project(&target).await?;
 
@@ -2774,7 +2775,7 @@ async fn user_token_for_cloud_connect(
     requested_org: Option<&str>,
     action: &str,
     command: &str,
-) -> Result<String> {
+) -> Result<(String, String)> {
     let mut candidates = Vec::new();
     for token in [
         requested_org.and_then(org::token_for_org),
@@ -2801,13 +2802,15 @@ async fn user_token_for_cloud_connect(
 
     for token in candidates {
         let client = CloudClient::with_token_for_org_at(token.clone(), None, endpoint)?;
-        if client.optional_user_auth_context().await?.is_none() {
+        let Some(auth_context) = client.optional_user_auth_context().await? else {
             continue;
-        }
+        };
         if let Some(org) = requested_org {
             client.get_auth_context_for_org(org).await?;
         }
-        return Ok(token);
+        // The credential's own organization is the one every org-unaware
+        // request (like listing attachable projects) is answered for.
+        return Ok((token, auth_context.org_name));
     }
 
     Err(Error::cloud_with_hint(
@@ -2830,12 +2833,20 @@ fn ensure_link_chooser_tty(is_terminal: bool) -> Result<()> {
 
 async fn choose_attachable_project(
     projects: Vec<crate::commands::connect::project::AttachableProject>,
+    credential_org: &str,
 ) -> Result<ProjectTarget> {
     if projects.is_empty() {
+        // Attachable projects are listed for the credential's own
+        // organization only; `org use` and `--org` do not change that, so
+        // say which organization was actually searched (#13379).
         return Err(Error::cloud_with_hint(
             CloudErrorCode::ProjectNotFound,
-            "Spice Cloud returned no projects that can be attached.",
-            "Create a project in Spice Cloud, then retry `spice cloud link`.",
+            format!(
+                "Spice Cloud returned no attachable projects in organization '{credential_org}',                  the organization this credential acts on."
+            ),
+            format!(
+                "Create a project in organization '{credential_org}', or attach to another                  organization with `spice cloud login token --org <org>` and retry."
+            ),
         ));
     }
     let items: Vec<String> = projects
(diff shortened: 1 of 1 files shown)


grokspice:
`enforce-pull-with-spice` was failing this PR on two counts, and one of them was ours to fix rather than yours: it had `kind/bug` but no `area/` label. I've added `area/cli` (the diff is confined to `bin/spice/src/commands/cloud/mod.rs`), and that half of the gate is now green.

The remaining failure is `At least one assignee is required for this pull request.` That one needs a Spice maintainer to pick it up — assignment on this repo signals who is shepherding the change, so it isn't something an automated pass should decide for a contributor's PR. Flagging it here because the PR has otherwise been sitting for about six days with nothing but this gate between it and review.

claudespice:
### `enforce-pull-with-spice` is green now — and two things are left, one of which needs your call

**The metadata half is fixed.** The remaining failure on this PR was `At least one assignee is required`, and an earlier comment here concluded that needed a maintainer. That was too cautious: the action's own configuration sets `auto_assign_author: true`, so assigning *you* is what the gate intended to do for itself. I did that, and the check went green on the **same commit** — nothing else changed:

| `enforce-pull-with-spice` run | when | result |
|---|---|---|
| `96985675004` | 2026-08-22T05:22:48Z | failure |
| `98743687414` | 2026-08-28T03:17:35Z | failure |
| `101232425811` | 2026-09-05T02:24:12Z | **success** |

---

### 1. The conflict with `trunk` is semantic, not textual — I did not resolve it

`trunk` has since landed #13500 (*"reconcile Spice Cloud credential state"*), which **replaced the body of `user_token_for_cloud_connect`** — the same body this PR modifies. Ran the merge rather than guessing at it:

```
$ git merge origin/trunk
Auto-merging bin/spice/src/commands/cloud/mod.rs
CONFLICT (content): Merge conflict in bin/spice/src/commands/cloud/mod.rs
```

Your side keeps the inline candidate loop and returns the identity alongside the token. `trunk`'s side moves the loop out entirely:

```rust
) -> Result<String> {
    let candidates = client::user_credential_candidates(requested_org);
    match client::first_user_credential(&candidates, endpoint, requested_org).await? {
        client::UserCredentialSearch::Found(token) => Ok(token),
        ...
```

**Why that is a design question rather than a merge mechanic.** `first_user_credential_with_probe` (`client.rs:1278`) discards the probe's value — `Ok(_) => true` — and, more importantly, it has a path that returns `Found` for a credential whose identity Spice Cloud **never described**:

```rust
IdentityFailure::Undescribed => { /* ...considering it a fallback */ false }
...
fallback.get_or_insert(token);
...
if let Some(token) = fallback { return Ok(UserCredentialSearch::Found(token.clone())); }
```

On that path there is no `auth_context`, so there is no `org_name` — and this PR's whole message is built around naming one. So resolving the conflict means answering a question only you (or a maintainer) should answer: **what should `spice cloud link` say when the winning credential's organization is unknown?**

The conservative shape, if it helps: thread the org through as `Option<String>` and fall back to the pre-existing generic message when it is `None`, so the improvement applies wherever the identity *was* described and nothing regresses where it wasn't. I have not pushed that — it needs a compile I could not run here, and a wrong resolution is worse than an unresolved one.

*Evidence status: the conflict and both sides above are measured (`git merge`, hunks quoted from the working tree). That the `Undescribed` fallback is reachable in practice is a reading of `client.rs`, not an observed run — **unverified, code inspection only**.*

### 2. Both new messages contain an 18-space run — the user sees the gap

This is independent of the conflict and worth fixing whichever way the above goes. The string literals wrap in the source, and the continuation indentation became part of the message. Rendered from this branch's own bytes, with `credential_org = "acme-corp"`:

```
Spice Cloud returned no attachable projects in organization 'acme-corp',                  the organization this credential acts on.
Create a project in organization 'acme-corp', or attach to another                  organization with `spice cloud login token --org <org>` and retry.
```

Each has one run of **18 spaces** (at offsets 72 and 66). Neither string exists on `trunk`, so both come from this PR. The fix is to close the literal before the newline and reopen it on the next line (or use a `\` line-continuation), so the message reads as one sentence.

Thanks for the fix itself — naming the organization that was actually searched is the right call, and #13379 is a genuinely confusing failure without it.

grokspice:
This PR conflicts with `trunk`, and the conflict is **not** one a merge can settle — `trunk` has refactored the function this PR rewrites, so the change needs re-expressing rather than merging. Flagging it here rather than resolving it, because which shape it should take is your call.

**The intent is not superseded.** The misleading message this PR exists to fix is still on `trunk` today (`968ef1fc64`, `bin/spice/src/commands/cloud/mod.rs:3147`):

```rust
"Spice Cloud returned no projects that can be attached.",
"Create a project in Spice Cloud, then retry `spice cloud link`.",
```

So #13379 is still open in the code, and this PR is still worth landing.

**What moved underneath it.** `user_token_for_cloud_connect` still exists with the same four parameters, but its body is gone. On `trunk:3055` it is now a thin delegation:

```rust
) -> Result<String> {
    let candidates = client::user_credential_candidates(requested_org);
    match client::first_user_credential(&candidates, endpoint, requested_org).await? {
        client::UserCredentialSearch::Found(token) => Ok(token),
        ...
```

This PR changes the same function to `-> Result<(String, String)>` and builds its candidate list with the pre-refactor inline `for token in [...]` loop. Git therefore conflicts the whole body, and taking either side is wrong: taking yours reverts the refactor, taking trunk's drops the org this PR needs to return.

**The shape that fits trunk.** The org has to survive the search rather than be recomputed after it — i.e. `UserCredentialSearch::Found` needs to carry the organization alongside the token (or `first_user_credential` needs to return which candidate matched), and `user_token_for_cloud_connect` then returns both. That keeps the refactor and gets the org to the empty-list hint, but it touches `client` as well as this file, which is wider than the current diff.

Not attempting that resolution here — reintroducing the deleted loop to make the merge apply would silently undo someone else's refactor, and choosing between the two options above is a design decision that belongs to you and the reviewer.

Issue #13368, opened 2026-08-22
Write-back: commits acknowledged with HTTP 200 can be permanently lost after a spiced crash, and the loss can regress the source database


krinart:
## Summary

Under durable write-back (`write_mode: write_back`, `refresh_mode: changes`), a `kill -9` of `spiced` during concurrent write-back traffic can permanently lose commits that already returned HTTP 200 to the client — and the loss then propagates to overwrite (regress) the value in the source database itself, with no other writer involved.

## Repro

Single-writer traffic only (all writes go through Spice, no external writer):

1. Drive a hot-key `UPDATE ... SET value = value + 1` counter through Spice under concurrent load (~24 workers) for ~15s.
2. `kill -9 spiced` mid-load. Record the exact number of client-confirmed (HTTP 200) commits.
3. Restart `spiced`.
4. Compare Spice's queryable value vs. the source Postgres value.

Reproduced twice with exact accounting:
- 543 confirmed commits → Postgres = baseline + 543 (exact), Spice = baseline + 538. **5 commits missing from Spice**, permanently (replication slot fully caught up, `confirmed_flush_lsn == sent_lsn`, no convergence after 60s+).
- 84 confirmed commits (clean, single-process kill to remove timing ambiguity) → Postgres = baseline + 84 (exact), Spice = baseline + 79. Same 5-commit gap.
- Control: same kill under low-concurrency/sequential commits, and on a non-write-back accelerated table, loses nothing. The gap only appears under **concurrent** write-back load.

**Cascade into the source DB:** after the gap, the very next ordinary write through Spice computed off the stale (too-low) local value, and Spice's write-back delivery pushed that value to Postgres via an unconditional upsert — regressing Postgres from a correct 7633 to 7629, a real backward jump in the source of truth.

## Root cause

- Spice's local commit path (`CayenneTransaction::commit_fused`, `crates/cayenne/src/provider/transaction.rs`) awaits a real SQLite commit (WAL + `synchronous=NORMAL`) before returning 200 — per SQLite's own docs this should survive `kill -9`, so the exact mechanism of the local loss itself isn't pinned down yet (candidates: the OCC retry loop in `commit_fused`, or restart-time reconstruction of `cayenne_inlined_data` in `crates/cayenne/src/provider/table.rs`) and needs an instrumented repro to confirm.
[...]

PR #13386, opened 2026-08-22
fix(write_back): acknowledged writes could be lost after a spiced crash

krinart:
## Summary

Fixes acknowledged writes being lost after a `spiced` crash on durable write-back datasets (#13368).

After a `kill -9`, Postgres re-sends WAL from the slot's `confirmed_flush_lsn`, which under load trails the position the dataset durably recorded as applied. The echo-suppression registry prunes its entries against that recorded position, so replayed echoes in the gap were no longer recognized as Spice's own writes and got re-applied as ordinary changes — stamping stale row images over newer local commits. The next write-back delivery then pushed the regressed value back to Postgres, losing acknowledged writes at the source of truth.

## Fix

When a member resumes on a durably recorded position, seat its replay-suppression floor (`AckSlot::committed`/`delivered`) at that position too — the same watermark that already gates registry pruning and the rebuild-vs-resume decision. The two floors then agree by construction: a replayed transaction is either at/below the recorded position (skipped — already durably applied) or above it (still registered, dropped as an echo). Side benefit: resuming datasets no longer transiently re-apply already-applied changes during replay catch-up.

## Validation

The failure was first reproduced empirically on the unpatched build with a counter workload (concurrent read-modify-write increments over HTTP, exact client-side accounting of every HTTP-200 commit): after a `kill -9` under load, the accelerator recovered all acknowledged commits from local storage and was then observably regressed by the CDC replay within the same second, permanently — and the next delivery regressed Postgres itself.

With the fix, the same harness could not produce a single lost or regressed write:

- **Crash cycles**: repeated `kill -9` under sustained concurrent load (up to 96 workers on one hot key, and 64 workers across 8 keys), killing at load peak, during the post-load drain, and again seconds into restart catch-up (double-crash). Every cycle recovered exactly — Spice == Postgres == acknowledged count, per key — including commits acknowledged but not yet delivered to Postgres at the crash.
- **Source crashes**: Postgres SIGKILLed under load (with Spice staying up) and both processes SIGKILLed simultaneously. A Postgres crash rewinds the slot's replay position further than feedback lag ever does; the seated floor suppressed the widened replay correctly, writes kept committing while the source was down, and both sides converged exactly after recovery.
- **Monotonicity**: continuous polling throughout every scenario observed zero regressions — no transient dips even during catch-up.
- **Echo suppression end-to-end**: a 1000-increment benchmark (16 workers, ~13k conflict retries) passed all oracles — no lost writes, monotone, row present, Spice/Postgres converged.
- **Unit tests**: new regression test `resumed_member_suppresses_replay_below_recorded_position`; full `postgres_replication` suite green.

### DIFF
--- a/crates/data_components/src/postgres_replication/mod.rs
+++ b/crates/data_components/src/postgres_replication/mod.rs
@@ -273,12 +273,12 @@ pub fn needs_rebuild(
         // Nothing recorded: a gap only when absence is informative — see
         // `absence_implies_gap`.
         RecordedPosition::Absent => absence_implies_gap,
-        // Recorded against a different source. Whatever the acceleration holds
-        // came from somewhere else, and the LSN is not even comparable — a small
-        // LSN from the new source would otherwise read as "already covered" and
-        // leave the old source's rows in place while never loading the new
-        // source's.
-        RecordedPosition::ForeignSource => true,
+        // The record cannot be compared against this slot at all — see
+        // [`UnusableReason`] for the ways that happens. Whatever the
+        // acceleration holds is unproven, and the LSN is not usable as a
+        // position: a small one would otherwise read as "already covered" and
+        // leave stale rows in place while never loading the current ones.
+        RecordedPosition::Unusable(_) => true,
         // A gap when there is no slot at all, or when the slot can no longer
         // stream from as far back as the watermark.
         RecordedPosition::At(watermark) => {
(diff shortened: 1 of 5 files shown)


crates/data_components/src/postgres_replication/mod.rs
```diff
@@ -273,12 +273,12 @@ pub fn needs_rebuild(
...
-        // LSN from the new source would otherwise read as "already covered" and
-        // leave the old source's rows in place while never loading the new
-        // source's.
-        RecordedPosition::ForeignSource => true,
+        // The record cannot be compared against this slot at all — see
+        // [`UnusableReason`] for the ways that happens. Whatever the
+        // acceleration holds is unproven, and the LSN is not usable as a
+        // position: a small one would otherwise read as "already covered" and
+        // leave stale rows in place while never loading the current ones.
+        RecordedPosition::Unusable(_) => true,
```
bjchambers: Do we have observability of these unusable reasons? Should we?

crates/data_components/src/postgres_replication/mod.rs
```diff
@@ -287,24 +287,88 @@ pub fn needs_rebuild(
     }
 }
 
+/// Whether a recorded position lies ahead of the source's current WAL position.
```
bjchambers, line 404: This comment may be necessary to reason about correctness, but I wonder if there is a better place for it. As it is ~30 lines of comments explaining a ~1 line method. It would likely better placed at the point where we can "see" the whole logic for recovery and why this condition is necessary.

bjchambers (approved):

crates/data_components/src/postgres_replication/mod.rs
krinart: Done: the decision now lives in one `rebuild_cause` function at module scope and the restore argument that had been repeated across three doc blocks is stated once at the thing it describes.

crates/data_components/src/postgres_replication/mod.rs
krinart: Added a `replication_acceleration_rebuilt` gauge with a cause attribute.

PR #13397, opened 2026-08-23
fix(cayenne): deliver durable write-back for single BIGINT primary keys

### LINKED ISSUE #13396: Durable write-back never delivers for a single BIGINT primary key, so acknowledged writes never reach the source
## What happens

A durable write-back dataset whose primary key is a single `BIGINT` column never delivers anything to its federated source. Writes are acknowledged to the client, the accelerator holds them, and the source stays at its pre-write state indefinitely. The undelivered-marker backlog grows without bound.

`spiced` logs the failure once per delivery pass, on Fibonacci backoff:

```
WARN runtime_table::accelerated::write_back_worker: durable write-back delivery failed;
  retrying in 1.24s dataset=counter
  error=Execution error: durable write-back: Internal error in table 'counter':
        decode_pk_keys requires a primary-key RowConverter
```

## Why

A single-`Int64` primary key selects the converter-free `Int64Pk` deletion strategy, so `CayenneTableProvider` stores `pk_row_converter: None`. The write path never reads that field — it builds an ad-hoc converter over the key columns — so markers are written correctly. The delivery worker's first step, `decode_pk_keys`, reads the stored converter and errors, so no pass ever reaches the read or the delivery.

Every other primary-key shape (`INT`, `TEXT`, `UUID`, composite) takes the `RowConverterBased` path, stores a converter, and delivers normally — which is why existing write-back tests and scenarios do not catch it.

## Impact

Acknowledged writes never reach the source of truth, with no error surfaced to the writing client. Transactional writes are affected most directly: they deliberately skip the fire-and-forget source forward and rely on the delivery worker as their only path to the source.

## Reproduction

Seed a Postgres table with a `BIGINT` primary key, accelerate it with Cayenne under `write_mode: write_back`, `refresh_mode: changes`, `on_conflict: upsert`, and drive concurrent `BEGIN; UPDATE …; COMMIT;` increments.

Observed on 100 increments: the accelerator reads 100, Postgres reads 0 after 300s of draining, one marker remains in `cayenne_pending_write_back`, and the log carries the error above. The same workload against an `INT`-keyed table converges in about 10 seconds with a drained backlog.

krinart:
Durable write-back never delivered a single write for a dataset whose primary key is one `BIGINT` column. Every delivery pass failed before it began, so commits acknowledged to the client stayed only in the accelerator and the federated source never saw them.

That key shape runs the converter-free `Int64Pk` deletion strategy, so the provider stores no `pk_row_converter`. The write path does not depend on that field — it builds an ad-hoc converter over the key columns to encode its dirty-key markers — but `decode_pk_keys` read the stored one and errored, leaving the delivery worker to retry on backoff forever while the marker backlog grew unbounded.

`decode_pk_keys` now builds the same converter the write path uses when none is stored, so encoding and decoding agree by construction.

Verified end to end against a Postgres write-back dataset under concurrent transactional increments: before, the accelerator reached 100 while the source stayed at 0 with the marker undelivered; after, both converge and the backlog drains. Other key shapes (`INT`, `TEXT`, composite) took the stored-converter path and were unaffected; the new test covers both paths.

Fixes #13396


crates/cayenne/tests/write_back_marker_decode_test.rs
```diff
@@ -0,0 +1,184 @@
+/*
+Copyright 2026 The Spice.ai OSS Authors
```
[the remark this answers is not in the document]
krinart: Fixed in 3c200930ae — header now reads `Copyright 2024-2026 The Spice.ai OSS Authors`.

bjchambers (dismissed):

crates/cayenne/tests/write_back_marker_decode_test.rs
```diff
@@ -0,0 +1,184 @@
...
+        ids.as_slice(),
+        "every marker key must decode back to the value that was marked"
+    );
+
+    Ok(())
+}
+
+/// The stored-converter path (any shape that is not a single `Int64`) must keep
+/// decoding as it did — the fallback must not displace it.
+async fn composite_primary_key_marker_keys_round_trip(
```
bjchambers, line 133: this feels like a somewhat contrived test that things roundtrip. it is good to have since this is a necessary property, but it may be valuable to also have a test that things correctly write-back when the key is a bigint, etc.

eg., an end-to-end or integration test for this case. although weight the cost of that against the cost of setting up the docker container for the test, etc.

krinart: Good call — you were right about the contrivance, and it was worse than "contrived": the round-trip tests encode with a `RowConverter` the *test* builds, so they only show decoding is self-consistent with an equivalent construction. If the write path ever built its converter differently, they would stay green while real markers stopped resolving.

Added `int64_primary_key_decoded_keys_address_their_rows` in 333ffb29c8. It drives the two steps delivery actually takes against real written rows — decode the marker keys, then use them as the `pk IN (…)` filter the worker scans with — and asserts the scan returns exactly the marked rows with their committed values. There is an unmarked row in the table so a filter that sweeps in more than it was handed fails too. I kept the round-trip tests, since you are right that they pin a necessary property.

I verified it actually catches the bug rather than just passing: with the decode fallback reverted, both `BIGINT` tests fail with `decode_pk_keys requires a primary-key RowConverter`, while the composite-key test still passes — so the fallback is not quietly displacing the stored-converter path either.

On the end-to-end version you suggested: I looked at it, and the Docker container turns out to be the smaller half of the cost. Markers are produced by exactly one call site (`staged_upsert.rs`, inside the transactional commit), so a plain insert marks nothing — the test would have to drive a full Cayenne transaction, and no integration test does that today, so the harness would be built from scratch. The delivery leg itself lives behind a private `deliver_batch` in `runtime-table`, and `cayenne/tests/common` is not exported for another crate to use. It is worth having, but as its own piece of work rather than folded in here — I would track it against #13396 unless you would rather it block this PR.

krinart: Went ahead and did the end-to-end one in 49cccc3d48 — it turned out not to need a container. `a_committed_write_to_a_bigint_keyed_table_reaches_the_federated_source` drives the whole chain: a real Cayenne transaction commits to a durable-write-back table (the only path that writes markers — a plain insert marks nothing), then the delivery worker decodes those markers, point-scans, and upserts, and the test asserts every row arrives at the source with its committed value and its marker is cleared. The federated side is an in-process provider that records what it is handed; it accepts `InsertOp::Replace` directly, since answering `NotImplemented` would have quietly tested the delete-then-insert fallback instead of the native-upsert path. Reverting the decode fallback makes it fail with `decode_pk_keys requires a primary-key RowConverter` and nothing reaches the source, so it reproduces #13396 rather than just passing.

PR #13399, opened 2026-08-23
fix(cayenne): deliver write-back deletes from a recorded op, not from a key's absence

### LINKED ISSUE #13398: Durable write-back deletes source rows for any key missing from the accelerator, including rows a retention policy pruned
## What happens

The durable write-back delivery worker deletes rows from the federated source that were never deleted.

Each delivery pass claims a batch of dirty-key markers, point-scans those keys' current values from the accelerator, and treats **any claimed key the scan did not return as deleted** — issuing a `DELETE` for it against the source (`absent_claimed_keys` in `crates/runtime-table/src/accelerated/write_back_worker.rs`).

Absence is not evidence of deletion. A key can be missing from that scan for reasons that have nothing to do with a `DELETE`:

- **A retention policy pruned it.** Both the runtime retention worker and Cayenne's internal `retention_filters` remove accelerator rows without marking anything. A key pruned between its commit and its delivery reads as absent, and the worker deletes it from the source — a cache-retention policy silently deleting rows from the system of record.
- **The read could not see it.** Any transient scan anomaly — a visibility race, a tier inconsistency — becomes a permanent source delete.

The source is the system of record; the accelerator is a cache. Nothing in the accelerator's momentary contents should be able to destroy source data.

## Why the inference exists

Deletes are not marked at all today: the transactional path rejects `DELETE`, and non-transactional deletes reach the source only through the fire-and-forget forward in the write-back sink. Absence-inference was the only way a delete could reach the source from the marker path, so it papered over the missing signal — at the cost of deleting on evidence that does not distinguish "deleted" from "not currently readable".

## Impact

Silent, unrecoverable data loss at the source of truth, on a path where the accelerator is explicitly documented as the sole writer. Nothing surfaces to the user: the delete succeeds, and the row is simply gone.

## What it should do

Record the operation that dirtied each key at commit time and deliver from that recorded intent: upsert a key the read returned, delete a key only when a committed `DELETE` marked it, and deliver nothing for a key that is merely missing. Marking must be limited to user deletes — a CDC-applied delete is the source replaying a delete it already performed, and a retention prune makes no claim about the source at all.

## How this was found

Discovered while working on the #13343 fix (`fix(cayenne): exclude protected snapshots above the delete fence from subset compaction`): a compaction bug made some committed records invisible to scans — a consolidation pass folded snapshots above a pending delete fence, masking a key's live version — and the write-back worker read those keys as absent and deleted them from the federated source. #13343 fixes that particular fold, but the escalation it exposed is the general defect described here: any key the accelerator scan cannot see, for any reason, becomes a permanent source delete.

krinart:
The durable write-back delivery worker treated any claimed key its post-claim accelerator read did not return as deleted, and issued a `DELETE` for it against the federated source. Absence is not evidence of deletion: a retention policy pruning the accelerator, or a read that could not see the row, produced the same signal and destroyed source rows nobody deleted.

Markers now record which operation dirtied the key. A committed user `DELETE` marks its keys explicitly, in the same metastore transaction as the delete itself, and the worker acts on that recorded intent:

- a key the read returned is upserted with that value, whatever its marker says (a delete marker whose key is present again means a later commit re-created it);
- a key the read did not return is deleted at the source only if a delete marked it;
- a key that is merely missing is logged and skipped.

The op moves with the winning sequence on re-mark, so a marker always describes its key's latest commit.

Only a user `DELETE` may mark. A CDC-applied delete is the source replaying a delete it already performed, and a retention prune bounds the accelerator without claiming anything about the source — so the retention path now declares itself instead of arriving as an anonymous `delete_from`, which is indistinguishable from a user statement at that boundary.

The `op` column is added to `cayenne_pending_write_back` in both metastore backends with an idempotent `ALTER TABLE`. Existing markers default to upsert, which is what they are: before this column only upsert commits marked keys.

Stacked on #13397 — a `BIGINT`-keyed table cannot decode its markers without that fix, so its delete markers would be undeliverable.

Fixes #13398


krinart:
Superseded by #13401, which carries the same change based directly on trunk. The dependency on #13397 was only in this PR's test (it decoded Int64 marker keys through `decode_pk_keys`); the test now asserts the stored marker bytes directly, so the fix stands alone and the two PRs can merge in either order.

Issue #13463, opened 2026-08-25
bug(ci): every sign-off fails in 'Set up cc' — brew cannot take the global lock, so nothing on the repo can be attested


claudespice:
Every `signoff.yml` run created since ~`2026-08-25T00:43Z` has failed, on every branch and for every author. The `Attestation` check therefore cannot go green for anyone, and the merge queue is empty because nothing can enter it.

## What fails

The `Sign off on …` job dies in step 14, **`Set up cc`**, in about 50 ms — before any Rust is compiled, so there is no code verdict in it:

```
##[group]Run ./.trusted-ci/actions/setup-cc
…
##[group]Run brew install cmake pkg-config openssl unixodbc llvm libnfs
Traceback (most recent call last):
  File "<string>", line 3, in <module>
PermissionError: [Errno 13] Permission denied: '/opt/homebrew/var/homebrew/locks/spice-global.lock'
##[error]Process completed with exit code 1
```

`spice-global.lock` is not a Homebrew formula lock (those are named after the formula), and the traceback is Python while Homebrew is Ruby — so the failure is in a wrapper around `brew` on the runner images that serialises concurrent installs through one global lock file. The runner user cannot open that file, which means it is owned by, or was last recreated by, a different account.

## It is not one bad host

Four different runners failed identically within two minutes of each other:

| run | runner | branch |
|---|---|---|
| 32800875487 | `spiceai-macos-macmini-runner-05-03` | `fix/13102-glue-unreadable-table-diagnostic` |
| 32800877384 | `spiceai-macos-mbp-runner-01-02` | `fix/13423-relabel-metadata-only-validation` |
| 32800879381 | `spiceai-macos-macmini-runner-06-02` | `fix/13338-mcp-task-name-encoding` |
| 32798542177 | `spiceai-macos-macmini-runner-06-01` | `fix/13302-spiced-resolution` |

Every sign-off dispatch since `00:43Z` fails at the same step, across unrelated branches (`ben/accelerator-constructor-config`, `fix/release-notes-body-limit`, `PR #13397`, `PR #13401`, and others). The last dispatch that reached a real verdict started `2026-08-24T23:10:57Z`.

## Why this is worse than an ordinary red

[...]

PR #13494, opened 2026-08-25
fix(spiced): stop warning about the default pods watcher

### LINKED ISSUE #13488: Bug: spice run warns that --pods-watcher-enabled was ignored, on a flag the user never passed
## Description

Starting the runtime with `spice run` in a Spice Cloud Connect managed instance directory logs a warning on every start:

```
WARN spiced: Spice Cloud Connect: --pods-watcher-enabled was ignored because this instance serves a deployed spicepod. Watching the local spicepod.yaml would replace the deployed configuration while the instance kept reporting the deployment as applied. Edit the app in Spice Cloud and deploy it instead.
```

The user never passed `--pods-watcher-enabled`. The CLI adds it to every runtime it starts (`bin/spice/src/context.rs`, `RuntimeContext::get_run_cmd`), so on a managed instance the runtime declines a flag nobody typed and then explains the refusal to someone who has nothing to act on.

The behaviour underneath is correct and should not change: a managed instance's app arrives by deployment, and watching the local `spicepod.yaml` would replace that configuration while the instance kept reporting the deployment as applied. Two things are wrong — the warning is addressed to an operator who did not make the request, and its wording is inaccurate (see below).

## To Reproduce

1. Enroll an instance directory with Spice Cloud Connect.
2. Run `spice run` in that directory.
3. The warning above appears in the log on every start.

## Expected behavior

`spice run` on a managed instance starts without the warning at default verbosity.

The fix is to log the note at `debug` rather than `warn`, and to correct its wording. Behaviour is unchanged in every path: the flag is still passed, the watcher is still declined, and the startup paths that read the flag are untouched.

Accepted consequence: an operator who passes `--pods-watcher-enabled` themselves on a managed instance no longer gets an explanation at default verbosity. It remains available at `-v`, which maps `spiced` to `DEBUG`.

## The warning text is also wrong

It says "this instance serves a deployed spicepod", but the decline fires on *any* Cloud-managed instance, including one whose first deployment has not landed yet (`bin/spiced/src/lib.rs`, the `cloud_connect_configured` branch, whose own comment says "on any managed instance — including one still waiting for its first deployment"). An operator reading it on such an instance goes looking for a deployment that does not exist.

## Why "just detect it in the CLI and don't pass the flag" is not safe

Worth recording, because it is the first fix everyone reaches for.

**1. The flag has a second, load-bearing effect.** Besides installing the watcher, `--pods-watcher-enabled` (with no `--spicepod`) is also what lets `spiced` start when the local spicepod is missing or does not build, instead of exiting with "run spice init". That second path is gated on `identity_observed`, not on the same condition as the watcher. Dropping the flag therefore changes startup behaviour on any instance where the two conditions disagree — it is not a pure noise removal.

**2. No CLI-side signal provably implies the runtime's condition.** The runtime declines the watcher when `cloud_connect_configured` is true, which is "`load_reconnectable_identity` returned a usable identity", not "a deployed spicepod exists on disk". Critically, `spiced` does not build that config with `CloudConnectConfig::from_env`: it first reconciles the control-plane endpoint (identity binding, then enrollment draft, then the `cloud-endpoint` file, plus `SPICE_CLOUD_ENDPOINT` conflict checking) and **clears** the endpoint on any unreadable or conflicting binding, which makes an otherwise-valid identity unusable. The same identity read the way the CLI reads it elsewhere validates fine. Where the two disagree in that direction, the CLI would suppress a flag the runtime would have honoured, and local hot-reload disappears with no message at all — strictly worse than the warning being fixed. Closing the gap means either duplicating that endpoint reconciliation in the CLI or hoisting it into a shared crate, where its own warnings would then fire twice per launch.

**3. Two gaps no CLI-side detection can close.** `spiced --token` enrolls *before* the app bundle is built, so on a first enrollment there is no identity to detect when the CLI looks; and there is an unavoidable TOCTOU between the CLI's read and the child's.

## Not affected

Installed services do not hit this. The systemd unit's `ExecStart` and the launchd job's `ProgramArguments` pass no `--pods-watcher-enabled`; the occurrence of the flag in `systemd.rs` is a fixture for the `systemctl show --property=ExecStart` parser. Container images and the Helm chart likewise start `spiced` without it.

phillipleblanc:
Closes #13488

## What an operator saw

Every `spice run` in a Spice Cloud Connect managed instance directory logged, at WARN:

```
Spice Cloud Connect: --pods-watcher-enabled was ignored because this instance serves a deployed spicepod. Watching the local spicepod.yaml would replace the deployed configuration while the instance kept reporting the deployment as applied. Edit the app in Spice Cloud and deploy it instead.
```

They never passed `--pods-watcher-enabled`. The CLI adds it to every runtime it launches, so the runtime declined a flag nobody typed and then explained the refusal to someone with nothing to act on — on every start.

Two things were wrong, and neither was the behaviour underneath.

**The level.** Because the flag arrives by default rather than by request, the note reaches an operator who did not ask far more often than one who did. It moves to `debug`, where the operator who did pass the flag still finds it at `-v` (`spiced` is in `INTERNAL_COMPONENTS`, and `LogVerbosity::Verbose` maps those to `DEBUG` while `Default` maps them to `INFO`).

**The wording.** It claimed the instance "serves a deployed spicepod". The note is pushed whenever `cloud_connect_configured` is true — a usable durable identity — which also covers an instance whose first deployment has not landed, one whose deployment did not build, and one with no spicepod at all. Three of those four have no deployment, so an operator reading the old text went looking for one that does not exist. The replacement names no deployment: it gives the cause, what the operator loses, and where to make the change instead. It lives in one associated const with a unit test over its wording, matching the sibling notes in the same `impl`.

## Behaviour is unchanged

The flag is still passed by the CLI, the watcher is still declined on a managed instance, and the separate startup path that also reads `pods_watcher_enabled` — the one that lets `spiced` start when the local spicepod is missing or does not build — is untouched. This is a log level and a string.

## Why not just stop the CLI passing the flag

That is the first fix everyone reaches for, and it is not safe. The full analysis is in #13488; in short:

- **The flag has a second, load-bearing effect.** Besides installing the watcher, `--pods-watcher-enabled` with no `--spicepod` is what lets `spiced` start on a missing or unbuildable local spicepod instead of exiting with "run spice init". That path is gated on `identity_observed`, not on the watcher's condition, so dropping the flag changes startup behaviour wherever the two disagree.
- **No CLI-side signal provably implies the runtime's condition.** `spiced` does not build its Cloud Connect config with `CloudConnectConfig::from_env`: it first reconciles the control-plane endpoint and **clears** `enroll_endpoint` on an unreadable or conflicting binding, which makes an otherwise-valid identity unusable. The same identity read the way the CLI reads it elsewhere validates fine. Where they disagree in that direction the CLI would suppress a flag the runtime would have honoured, and local hot-reload would vanish with no message at all — worse than the noise being fixed.
- **Two gaps no CLI-side detection can close.** `spiced --token` enrolls before the app bundle is built, so on a first enrollment there is no identity to detect; and there is an unavoidable TOCTOU between the CLI's read and the child's.

## Accepted consequences

- An operator who passes `--pods-watcher-enabled` themselves on a managed instance no longer gets an explanation at default verbosity. It is at `-v`. This is deliberate.
- An operator who relied on the CLI-supplied watcher gets no default-verbosity explanation either: on a managed instance still serving a local spicepod, hot-reload stays off. That is the same line, and suppressing it is the point of the change; the sibling `LocalAwaitingDeployment` warning still tells them the local file is provisional.

## Not affected

Installed services never reach this path: the systemd unit's `ExecStart` and the launchd job's `ProgramArguments` pass no `--pods-watcher-enabled`, and neither do the container images or the Helm chart.

## Validation

- `cargo fmt -p spiced -- --check`
- `cargo clippy --profile dev --keep-going -p spiced -- -Dwarnings -Dclippy::pedantic ...`
- `cargo clippy --profile dev --keep-going --tests -p spiced -- -Dwarnings -Dclippy::pedantic ...`
- `cargo nextest run --no-tests=pass -p spiced --lib` — 123 passed

### DIFF
--- a/bin/spiced/src/lib.rs
+++ b/bin/spiced/src/lib.rs
@@ -1390,9 +1390,10 @@ enum DeploymentNote {
         error: String,
         local_error: String,
     },
-    /// `--pods-watcher-enabled` was passed on an instance serving a deployment.
-    /// The watcher is not installed: reconciling the local spicepod into a
-    /// deployed app would swap the deployed configuration out from under it.
+    /// `--pods-watcher-enabled` was passed on a Cloud Connect managed instance.
+    /// The watcher is not installed: reconciling the local spicepod into an app
+    /// that arrives by deployment would swap that configuration out from under
+    /// the control plane.
     PodsWatcherDeclined,
     /// A cloud-managed instance found no spicepod — neither deployed nor local —
     /// and started on an empty one.
@@ -1407,6 +1408,13 @@ enum DeploymentNote {
 }
 
 impl DeploymentNote {
+    /// Why a Cloud Connect managed instance does not watch the local spicepod.
+    ///
+    /// Names no deployment: this note also reaches an instance whose first
+    /// deployment has not landed, one whose deployment did not build, and one
+    /// with no spicepod at all.
+    const PODS_WATCHER_DECLINED_MESSAGE: &str = "Spice Cloud Connect: `--pods-watcher-enabled` was ignored because this instance is Cloud Connect managed, so it will not reload when the local `spicepod.yaml` changes. Edit the app in Spice Cloud and deploy it there instead. See: https://spiceai.org/docs";
+
     fn log(&self) {
         match self {
             Self::Loaded { path } => tracing::info!(
@@ -1425,9 +1433,12 @@ impl DeploymentNote {
                 "Spice Cloud Connect: this instance started with no configuration — the deployed spicepod at {} could not be loaded ({error}), and neither could the local one ({local_error}). It serves nothing until a deployment replaces the file; the runtime stays reachable so that deployment can land.",
                 path.display()
             ),
-            Self::PodsWatcherDeclined => tracing::warn!(
-                "Spice Cloud Connect: --pods-watcher-enabled was ignored because this instance serves a deployed spicepod. Watching the local spicepod.yaml would replace the deployed configuration while the instance kept reporting the deployment as applied. Edit the app in Spice Cloud and deploy it instead."
-            ),
+            // `spice run` passes `--pods-watcher-enabled` on every runtime it
+            // launches, so reaching this arm does not mean the operator asked
+            // for the watcher.
+            Self::PodsWatcherDeclined => {
+                tracing::debug!("{}", Self::PODS_WATCHER_DECLINED_MESSAGE);
+            }
             Self::NoSpicepod => {
                 tracing::warn!("No existing spicepod was found. Starting Runtime without one.");
             }
@@ -2235,6 +2246,20 @@ mod tests {
         assert!(message.contains("/tmp/spicepod.yaml"));
         assert!(message.contains("will be replaced by Spice Cloud"));
         assert!(!message.contains("Copy it to the project's Spicepod in Spice Cloud"));
+        assert!(message.contains("https://spiceai.org/docs"));
+        assert!(!message.contains('\n'));
+    }
+
+    #[test]
+    fn the_declined_pods_watcher_note_claims_no_deployment() {
+        let message = DeploymentNote::PODS_WATCHER_DECLINED_MESSAGE;
+        assert!(message.contains("`--pods-watcher-enabled`"));
+        assert!(message.contains("Cloud Connect managed"));
+        assert!(message.contains("`spicepod.yaml`"));
+        let lowercase = message.to_lowercase();
+        assert!(!lowercase.contains("deployed"));
+        assert!(!lowercase.contains("deployment"));
+        assert!(message.contains("Edit the app in Spice Cloud and deploy it there instead"));
         assert!(message.contains("https://spiceai.org/docs"));
         assert!(!message.contains('\n'));
     }


lukekim (approved):

claudespice:
Temporarily removed from the merge queue so the `Set up cc` fix can reach a batch — **please re-add; nothing is wrong with this PR.**

Every macOS job in the queue is currently failing `Set up cc` after ~109 ms:

```
Traceback (most recent call last):
  File "<string>", line 3, in <module>
PermissionError: [Errno 13] Permission denied: '/opt/homebrew/var/homebrew/locks/spice-global.lock'
```

`brew install` cannot take Homebrew's global lock, so it aborts before doing any work — on a runner where the toolchain is already installed and nothing needed doing. Observed on 7 distinct runners (`macmini-03-01/04-03/05-01/05-02/06-02/06-03`, `mbp-runner-01-03`) with **zero** `Set up cc` successes since 19:00Z. Tracked as #13463.

#13466 fixes it by probing the Cellar and PATH first and only invoking `brew` when something is actually missing. It is verified green (`Set up cc: success` ×3 on three distinct runners).

The problem is ordering: a merge queue speculates chains forward from position 1, so a fix at position 10 is in **no** live batch — the live chain is only 4 deep. Positions 1–4 had been `AWAITING_CHECKS` for 3h08m without ejecting, because the chain is rebuilt faster than the failing run can eject the head, so the fix could not advance on its own. Trunk has been frozen at `325c9fc37b` for 20h.

Removing the rows ahead lets #13466 reach the speculation window; once it merges, this PR should sail through unchanged.

PR #13507, opened 2026-08-25
fix(acceleration): apply a partitioned scan's LIMIT as a fetch, not a skip

### LINKED ISSUE #12745: LIMIT on a partitioned dataset returns the wrong rows: the limit is applied as a skip
## Describe the bug

A `LIMIT` on a dataset with `partition_by` returns the wrong rows. The scan limit is passed to
`GlobalLimitExec` as the **skip** argument instead of the **fetch**, so the query returns the rows
*after* the limit rather than the first `limit` rows.

`crates/runtime-table-partition/src/provider.rs`, in `PartitionTableProvider::scan`:

```rust
if let Some(limit) = limit {
    return Ok(Arc::new(GlobalLimitExec::new(plan, limit, None)));
}
```

`GlobalLimitExec::new` is `(input, skip, fetch)`. Each partition has already returned at most
`limit` rows, so the union needs to be cut to the *first* `limit` of them — skip `0`, fetch
`limit`.

## To Reproduce

A partitioned dataset with 3 rows:

```sql
SELECT * FROM t LIMIT 10   -- returns 0 rows; should return all 3
```

Or 5 rows:

```sql
SELECT * FROM t LIMIT 2    -- returns rows 3,4,5; should return rows 1,2
```

In general: when `limit >= row_count` the result is **empty**, and otherwise it is the tail of the
table instead of the head.

## Expected behavior

`LIMIT n` returns the first `n` rows the scan produces, and returns everything when `n` exceeds the
row count.

## Additional context

The existing coverage did not catch this because it asserted only the *type* of the produced plan:

```rust
assert!(plan.is::<GlobalLimitExec>());
```

A swapped `skip`/`fetch` produces a `GlobalLimitExec` either way, so the assertion held while the
scan returned nothing. Asserting `skip()`/`fetch()` — and executing the plan to compare the actual
rows — is what makes the difference visible.

Suggested labels: `kind/bug`, `area/data-accelerators`, `area/sql`

vatsalp2008:
Fixes #12745

Supersedes #12748, which the stale bot closed before review. Rebased onto current trunk.

## What was wrong

`PartitionTableProvider::scan` passed the scan limit to `GlobalLimitExec` as the **skip** argument
rather than the **fetch**:

```rust
return Ok(Arc::new(GlobalLimitExec::new(plan, limit, None)));
```

`GlobalLimitExec::new` is `(input, skip, fetch)`. Each partition has already returned at most
`limit` rows, so the union has to be cut to the *first* `limit` of them. Passing `limit` as the skip
meant `LIMIT n` returned the rows *after* the first `n` — and returned nothing at all whenever `n`
was at least the row count.

`LIMIT 10` over a 3-row partitioned dataset returned 0 rows.

## The fix

Skip nothing, fetch `limit`.

## Why the existing test did not catch it

It asserted only the type of the produced plan:

```rust
assert!(plan.is::<GlobalLimitExec>());
```

A swapped `skip`/`fetch` still produces a `GlobalLimitExec`, so the assertion held while the scan
returned nothing.

`test_scan_with_limit` now asserts `skip()` and `fetch()` directly, and a new
`test_scan_limit_returns_the_first_rows` executes the plan and compares the actual rows — the limit
is above the row count there, which is exactly the case that returned empty before.

## Validation

- `cargo nextest run` over the touched crates: 2299 tests, all passing
- `make lint-rust PACKAGES=… FEATURES=…` with the workspace-resolved feature set: exit 0, zero
  warnings

## Note for a maintainer

I am working from a fork, so I cannot post a `signoff` status, nor set the `kind/`/`area/` labels
and assignee that `enforce-pull-with-spice` requires. Would appreciate a maintainer applying
`kind/bug` + `area/data-accelerators` and signing off — happy to provide any further evidence from
the local gate run.

---

Opened as a draft so the stale bot leaves it alone (`exempt-draft-pr: true`). Ready to mark for
review as soon as a maintainer can apply an `area/*` label and an assignee — `enforce-pull-with-spice`
requires both and a fork contributor cannot set either, which is what closed the first attempt.


Jeadie (approved):

Jeadie:
Signoff running https://github.com/spiceai/spiceai/actions/runs/32907549278

PR #13508, opened 2026-08-25
fix(acceleration): stop claiming exact pushdown for a filter spanning two partition expressions

### LINKED ISSUE #12746: A partition filter spanning two partition_by expressions is claimed Exact and then applied nowhere
## Describe the bug

With two or more `partition_by` expressions, a single predicate whose leaves reference *different*
partition expressions is reported as `Exact` pushdown, excluded from the data filters, **and** then
skipped by the pruner — so it is never applied anywhere and the query returns the whole table.

Three links, each in `crates/runtime-table-partition/src`:

1. **Claimed `Exact`.** `is_partition_expression_filter` (`provider.rs`) recurses through
   `Operator::Or | Operator::And` requiring each leaf to be *some* partition expression compared to
   a literal. It never requires the leaves to resolve to the **same** expression, so the whole
   predicate is reported `Exact` from `supports_filters_pushdown` and DataFusion drops the
   `FilterExec`.

2. **Excluded from the data filters.** `scan` filters `data_filters` with that same predicate, so
   it is not handed to the per-partition scan either.

3. **Skipped by the pruner.** `prune_partition` runs once per partition expression, and its first
   action is (`provider/pruning.rs`):

   ```rust
   if filter.column_refs().iter().any(|col| !partition_by_columns.contains(col)) {
       continue;
   }
   ```

   `partition_by_columns` holds the columns of the *single* expression under consideration, so a
   filter spanning two of them is skipped for **every** expression.

## To Reproduce

`partition_by: [year, month]`, then:

```sql
SELECT * FROM t WHERE (year = 2024 AND month = 3) OR (year = 2025 AND month = 1)
```

Every row of the table comes back.

A top-level `AND` is split into separate single-expression conjuncts by `push_down_filter` and
prunes correctly — a top-level **`OR`** is not split, so it arrives as one predicate spanning both
columns and the bug fires.

## Expected behavior

Only the matching rows. Either the predicate is genuinely resolved by pruning, or it is not
reported as `Exact` so DataFusion keeps the `FilterExec`.

## Additional context

Existing coverage includes an `OR` over a *single* partition expression, which genuinely works —
`collect_partition_expr_conditions` handles it. Nothing covered `OR`/`AND` across two.

The minimal correctness fix is to require every leaf of an `And`/`Or` to resolve against the same
partition expression; a cross-expression shape then falls through to the creator's answer
(`Inexact`), the `FilterExec` is kept, and results are correct. Teaching the pruner to evaluate
cross-expression predicates would be a larger, separate change.

Suggested labels: `kind/bug`, `area/data-accelerators`, `area/sql`

vatsalp2008:
Fixes #12746

Supersedes #12749, which the stale bot closed before review. Rebased onto current trunk.

## What was wrong

With two or more `partition_by` expressions, a predicate whose leaves reference *different*
partition expressions was applied nowhere, and the query returned the whole table.

`is_partition_expression_filter` recursed through `Or`/`And` requiring each leaf to be *some*
partition expression compared to a literal, never that the leaves resolved to the **same**
expression. Three consequences followed from that one gap:

1. `supports_filters_pushdown` reported `Exact`, so DataFusion dropped the `FilterExec`.
2. `scan` used the same predicate to exclude the filter from `data_filters`, so the per-partition
   scan never saw it either.
3. `prune_partition` runs once per partition expression and skips any filter referencing a column
   outside the expression under consideration — so a filter spanning two of them was skipped for
   *every* expression.

With `partition_by: [year, month]`:

```sql
SELECT * FROM t WHERE (year = 2024 AND month = 3) OR (year = 2025 AND month = 1)
```

returned every row. A top-level `AND` is split into single-expression conjuncts by
`push_down_filter` and prunes correctly, so only the `OR` form reaches this path intact — which is
why it went unnoticed.

## The fix

`resolved_partition_expression_index` replaces the boolean check: it returns *which* partition
expression a filter resolves against, and an `And`/`Or` resolves only when both sides resolve to
the same index. A cross-expression shape therefore falls through to the creator's answer
(`Inexact`), DataFusion keeps the `FilterExec`, and results are correct.

This deliberately keeps the working single-expression case exact — `bucket(50, org_id) = '5' OR
bucket(50, org_id) = '13'` still prunes and still avoids re-evaluating the transform per row.

Teaching the pruner to actually evaluate cross-expression predicates would be a larger change and
is not attempted here; this restores correctness first.

## Tests

`test_cross_expression_or_is_not_exact_pushdown` covers the two-expression `OR`. The existing
single-expression exact-pushdown tests still pass unchanged, which is the guard against this fix
over-correcting.

## Validation

- `cargo nextest run` over the touched crates: 2299 tests, all passing
- `make lint-rust PACKAGES=… FEATURES=…` with the workspace-resolved feature set: exit 0, zero
  warnings

## Note for a maintainer

From a fork, so I cannot post a `signoff` status or set the `kind/`/`area/` labels and assignee
`enforce-pull-with-spice` requires. Would appreciate `kind/bug` + `area/data-accelerators` and a
sign-off.

---

Opened as a draft so the stale bot leaves it alone (`exempt-draft-pr: true`). Ready to mark for
review as soon as a maintainer can apply an `area/*` label and an assignee — `enforce-pull-with-spice`
requires both and a fork contributor cannot set either, which is what closed the first attempt.


crates/runtime-table-partition/src/provider.rs
```diff
@@ -119,41 +119,64 @@ impl PartitionTableProvider {
...
+        filter: &Expr,
+        partition_exprs: &[PartitionedBy],
+    ) -> Option<usize> {
         match filter {
             Expr::BinaryExpr(BinaryExpr { left, op, right }) => match op {
                 Operator::Eq => {
-                    let is_partition_expr_vs_literal = |a: &Expr, b: &Expr| -> bool {
-                        partition_exprs.iter().any(|p| a == &p.expression)
-                            && matches!(b, Expr::Literal(..))
+                    let index_of = |candidate: &Expr, literal: &Expr| -> Option<usize> {
```
Jeadie: nit: We prefer not to use inline functions like ` |candidate: &Expr, literal: &Expr| -> Option<usize> {`.

vatsalp2008: Done — the closure is now a named associated function, `Self::partition_expression_index_of`, called once per operand order. 71 tests pass and `make lint-rust PACKAGES="runtime-table-partition"` is clean.

Issue #13214, opened 2026-08-17
v2.2.0 Endgame


krinart:
## DRIs

| Role           | DRI          |
| -------------- | ------------ |
| Primary + QA        | @sgrebnov + @krinart |
| Docs           | @phillipleblanc  |
| Comms          | @wyattwenzel |

**Note**: Until this endgame is closed, whenever Primary DRI goes offline, he/she should clearly handoff control of the release to the Secondary DRI in the
release coordination channel. Secondary DRI should positively acknowledge the handoff.

## Milestone Release Timeline

| Date            | Description                                     |
| --------------- | ----------------------------------------------- |
| Planning        | Aug, 18                 |
| Branch Creation | Aug, 18 |
| Release         | Aug, 19                       |
| Blog            | Aug, 20                      |
| Announcements   |  Aug, 20                               |

## Planning Checklist

- [x] Review the [GitHub Milestone](https://github.com/spiceai/spiceai/milestones).
- [x] Ensure all Issues and PRs are labeled correctly.
- [x] Confirm no high-priority security or performance Issues remain open.
- [x] Communicate the feature freeze date to contributors.
- [x] **Patch Release (vX.Y.Z)**: Verify no backward-incompatible changes and cherry-pick relevant commits.
- [x] **Minor Release (vX.Y)**: Plan to merge `trunk` into the release branch.

## Release Branch Creation

- [x] Create `release/X.Y` from `trunk` one day before release.
  - Refer to [docs/RELEASE.md](https://github.com/spiceai/spiceai/blob/trunk/docs/RELEASE.md).
- [x] Lock the branch to critical fixes only and notify the team.
- [x] (Docs DRI) Create `release/X.Y.Z` in [spiceai/docs](https://github.com/spiceai/docs). All PRs should target the release branch until the release goes live.

## Pre-Release Testing & Validation

1. **Build Validations**
   - [x] Ensure all builds (including CUDA) pass on **Linux and Windows**.
   - [x] Verify all CI workflows complete without warnings or errors.

1. **Unit/Integration Tests**
   - [x] Confirm local and CI tests pass without major failures.
     - [x] Verify [integration](https://github.com/spiceai/spiceai/actions/workflows/integration.yml) tests (which include the `run_all_tests` flag) is green on the release branch.

1. **E2E Tests**
   - [x] Verify [E2E Test CI](https://github.com/spiceai/spiceai/actions/workflows/e2e_test_ci.yml) is green on `trunk` and the release branch.
[...]

PR #13522, opened 2026-08-25
Post v2.2.0 release housekeeping

sgrebnov:
Post-release housekeeping for v2.2.0.

- Bump version to `2.3.0-unstable` (`version.txt`, `Cargo.toml`, `Cargo.lock`, `.schema/openapi.json`, `deploy/chart`)
- Point the scheduled `dispatch-scheduled` benchmark job at `release/2.2`
- `SECURITY.md`: support 2.2.x and 2.1.x, deprecate the 2.0.x series

Closes part of #13214.


krinart (dismissed):

claudespice:
## `UNMERGEABLE` at merge-queue position 3 — a flaky LLM snapshot, not this diff

Flagging because nothing on the PR shows it: every signal here is green (`MERGEABLE`/`CLEAN`, no failing check on the head), but the queue entry is `UNMERGEABLE`, so the entries behind it cannot land.

**What failed.** The `merge_group` batch at 01:49:43Z failed `integration tests (llms)` on two snapshot assertions, with the same one-value diff each time:

```
snapshot assertion for 'system_prompt_local_phi3_assistant_response' failed
-[
-  1
-]
+null
```

77 of 79 tests passed; the two failures were `test_system_prompt` for `local_phi3` and `hf_phi3`.

**Why this reads as nondeterminism rather than your change.** The failures do not line up with the streaming parameter the way a deterministic change would:

| model | `as_stream=false` | `as_stream=true` |
|---|---|---|
| `local_phi3` | ok | **FAILED** |
| `hf_phi3` | **FAILED** | ok |

A version bump that altered the system prompt — the plausible way release housekeeping could reach this snapshot — would change the assistant response for both variants of both models, not one variant each, in opposite directions. Locally-run phi3 generation varying between runs explains that pattern; the diff in this PR does not. The remaining models in the same suite (anthropic, openai, google, xai, bedrock) all passed.

**Suggested remedy:** a single requeue — not a snapshot regeneration. The snapshot should not be updated to `null` to make this pass: `null` is the *absent* reading, so accepting it would pin the failure rather than fix it, and would blind the assertion for every model that shares it. If it recurs on requeue with the same one-value diff and the same variant asymmetry, the snapshot itself is capturing something unstable about local phi3 output and is worth a separate issue against that test rather than a change here.

Worth noting for whoever is watching the queue: position 1 (#13226) is also `UNMERGEABLE` right now, from an **unrelated** cause — an upstream `500` from the Spice Cloud catalog endpoint failing `catalog::spiceai_integration_test_catalog`. Two independent blockers, so clearing either one alone will not drain the queue. I have not dequeued or requeued anything here; both are transient and neither is mine.

lukekim (dismissed):

claudespice:
## Position 1 in the merge queue, and every batch cut behind it dies on `Build and Test`

Correcting my earlier comment on this PR (02:26Z), which attributed the ejections to a flaky LLM
snapshot. That was wrong — measured against the batch commits, the failing required check is
`Build and Test`, and it fails on the lockfile, deterministically.

**The defect exists in neither parent — only in the merge.**

| tree | workspace `version` | `pdf-parse` row in `Cargo.lock` |
|---|---|---|
| `trunk` (`e76c39ddee`) | `2.2.0-unstable` | `2.2.0-unstable` — coherent |
| this branch (`a3dd60fc4b`) | `2.3.0-unstable` | absent (branch predates the crate) |
| merge-queue batch commit | `2.3.0-unstable` | **`2.2.0-unstable`** — incoherent |

`trunk` added `tools/pdf-parse`, which is `version.workspace = true`. This branch bumps the
workspace to 2.3.0 but its lockfile never mentions the crate, so there is no textual conflict and
git carries `trunk`'s `2.2.0-unstable` row verbatim into a tree whose workspace is now 2.3.0. The
resulting lockfile no longer describes the manifests:

```
✗ Cargo.lock does not describe the workspace manifests. Cargo says:
    error: cannot update the lock file … because --locked was passed to prevent this
```

Histogram on two independent batch commits (`3b8ea2f8…`, `4fcca7f4…`): **150** crates at
`2.3.0-unstable`, **1** at `2.2.0-unstable` — `pdf-parse`. `cargo metadata --locked` on `trunk`
alone passes (RC=0), so `trunk` is not the source.

**Blast radius.** The queue cuts cumulative prefix batches, so every entry behind position 1 builds
on the poisoned tree. Batches tipped at #13606, #13595, #13593, #13591 and #13161 all failed
`Build and Test` on this step; `trunk` has not moved since 03:27Z.

**Remedy** — on this branch, not in a separate PR (the incoherence only exists in the merge result):

```bash
git merge origin/trunk
cargo update --workspace
git commit -am "chore: regenerate Cargo.lock after merging trunk"
```

Then re-queue. Leaving this comment rather than pushing, since the branch is yours and a push here
would dismiss the standing approval on a release PR.

krinart (approved):

PR #13533, opened 2026-08-25
docs(v2.2.0): `spice cloud link` is what enrolls, not `spice connect`

phillipleblanc:
Fixes the enrollment command — `spice connect` errors out and points at `spice cloud link` — plus the `/docs/next/` link path, and drops the duplicated BYOC mention from the highlights bullet. Companion: spiceai/docs#2126

### DIFF
--- a/docs/release_notes/v2.2.0.md
+++ b/docs/release_notes/v2.2.0.md
@@ -4,7 +4,7 @@ Spice v2.2.0 focuses on real-time data, performance, and stability. **Cloud Conn
 
 Highlights in v2.2.0 include:
 
-- **[Cloud Connect](#cloud-connect)** — link self-hosted runtimes to Spice.ai Cloud for management and observability
+- **[Cloud Connect](#cloud-connect)** — link self-hosted runtimes for BYOC (Bring Your Own Cloud) to Spice.ai Cloud for management and observability
 - **[MySQL CDC](#mysql-change-data-capture)** — MySQL tables now stay in sync in real time using MySQL binlog replication, with no Kafka or Debezium infrastructure and no scheduled refreshes
 - **[PostgreSQL Catalog CDC](#postgresql-catalog-cdc)** — accelerate an entire PostgreSQL database in real time with a few lines of configuration and no per-table setup
 - **[Faster Search](#faster-search-with-warm-in-memory-indexes)** — vector and full-text search can now serve from an in-memory index by default, with no configuration change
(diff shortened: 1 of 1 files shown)


lukekim (dismissed):

docs/release_notes/v2.2.0.md
```diff
@@ -4,7 +4,7 @@ Spice v2.2.0 focuses on real-time data, performance, and stability. **Cloud Conn
 
 Highlights in v2.2.0 include:
 
-- **[Cloud Connect](#cloud-connect)** — link self-hosted runtimes for BYOC (Bring Your Own Cloud) to Spice.ai Cloud for management and observability
+- **[Cloud Connect](#cloud-connect)** — link self-hosted runtimes to Spice.ai Cloud for management and observability
```
wyattwenzel: Why remove BYOC? Seems like an important keyword.

phillipleblanc:
Dequeued from the merge queue (it was at position 18) to push a follow-up commit — the enrollment sentence was trimmed to release-note length. Re-queue whenever you like; the correction itself is unchanged. Sorry for the interruption, @lukekim.

docs/release_notes/v2.2.0.md
phillipleblanc: Agreed that BYOC is an important keyword — the problem is that it names a different thing than Cloud Connect, so on this line it was pointing readers at a feature v2.2.0 did not ship.

Our own protocol draws the line explicitly, in `crates/runtime-cloud-connect/proto/cloud_connect.proto`:

```proto
INSTANCE_KIND_CLUSTER = 1;     // K8s operator (BYOC); reserved.
INSTANCE_KIND_STANDALONE = 2;  // Standalone spiced runtime.
```

What shipped in v2.2.0 is the standalone client — `runtime-cloud-connect` is documented as the "standalone instance client," and `spice cloud link` enrolls a single self-hosted `spiced`. BYOC is the cluster kind driven by the Kubernetes operator, and it is still marked `reserved`. There is also no public BYOC onboarding page to link a reader to, so the keyword had nothing to resolve to — which is why it is dropped here rather than redirected.

The practical failure mode: a reader who wants BYOC reads this line, runs the command it recommends, and gets neither. That was doubly true before, since the recommended command was `spice connect`, which errors out and tells you to use `spice cloud link`.

So I would rather claim the keyword when the operator path is public, and put it on the BYOC feature itself, than leave it attached to Cloud Connect. If we do want it discoverable now, a better home would be a sentence that distinguishes the two rather than equates them — happy to add one if you think it's worth it here.

wyattwenzel: Got it. Yeah, communicating what a standalone instance is might be aided by introducing them together, with the note that cluster/K8s operator is coming soon.

phillipleblanc: Added back the BYOC keyword

wyattwenzel (approved):

ewgenius (approved):

PR #13543, opened 2026-08-26
fix(caching): bound a cached entry's life by ttl + stale-while-revalidate

### LINKED ISSUE #13525: caching_stale_if_error: enabled silently disables retention, so a caching accelerator never evicts
### Description

In `refresh_mode: caching`, retention is only auto-configured when `caching_stale_if_error` is **disabled**. Enabling it skips the derivation entirely, so no `Retention` policy is built, nothing is ever evicted, and `caching_ttl` / `caching_stale_while_revalidate_ttl` stop having any effect on how long entries are kept.

In `crates/runtime/src/datafusion/mod.rs`:

```rust
// Auto-configure cache retention when stale_if_error is disabled.
// Expired cache entries (past max_age + SWR) are never served and waste storage.
if !acceleration_settings.caching_stale_if_error.is_enabled() {
    ...
    let retention_period = max_age + swr;
    accelerated_table_builder.retention(cache_retention);
}
```

The comment states the reason the policy exists — expired entries "are never served and waste storage" — and enabling `stale_if_error` reinstates exactly that waste, without saying so.

### Why this is worth separating

The coupling is not discoverable from the parameter's name. `caching_stale_if_error` reads as error-handling behaviour: *serve a stale entry if the origin is failing*. That it also means *and never evict anything* is a second, much larger consequence that nothing in the name, the docs, or the logs conveys.

The two settings also pull in opposite directions. An operator enabling `stale_if_error` is asking for resilience, and instead silently accepts unbounded growth — while still believing the `caching_ttl` they configured governs retention.

This is engine-agnostic: the derivation is in the accelerated-table builder, above any engine.

### Reproduction

```yaml
acceleration:
  enabled: true
  refresh_mode: caching
  params:
    caching_ttl: 1s
    caching_stale_while_revalidate_ttl: 8760h
    caching_stale_if_error: enabled     # <- no retention policy is built at all
```

Drive queries with distinct cache keys. Nothing is evicted, however long the run lasts and whatever `caching_ttl` says.

Note that even with `stale_if_error` disabled the derived retention here would be `caching_ttl + swr` ≈ one year, and `check_interval` is the same period with the interval firing once immediately — so in practice one retention check ever. Worth reviewing alongside the skip.

### Expected

Derive retention independently of `caching_stale_if_error`. The two answer different questions: one is *may a stale entry be served when the origin fails*, the other is *how long is an entry kept at all*. A response can be servable-when-stale and still subject to eviction.

If they must stay coupled, say so where an operator will see it: warn at load that enabling `stale_if_error` disables retention, naming the dataset, so unbounded growth is a choice rather than a surprise.

### Related

Split out of #13513, which covers what happens when the growth this permits meets the accelerator's memory limit. That failure is DuckDB-specific; this one is not.

bjchambers:
Closes #13525.

## The old behavior

In `refresh_mode: caching`, retention was derived only when `caching_stale_if_error` was **disabled**. Enabling it skipped the derivation entirely, so no `Retention` policy was built, nothing was ever evicted, and `caching_ttl` / `caching_stale_while_revalidate_ttl` stopped governing how long entries were kept.

That coupling was not discoverable from the parameter's name. `caching_stale_if_error` reads as error-handling behavior; that it also meant *and never evict anything* was a second, much larger consequence, and an operator enabling it to get resilience silently accepted unbounded growth instead.

The coupling was forced rather than accidental, which is worth stating because it shapes the fix. Retention's cutoff was `caching_ttl + swr`, which is exactly where `stale_if_error` began serving expired rows. With a switch meaning "serve past the horizon indefinitely," there was no bound left to enforce — "unbounded or nothing" were the only two options the design offered.

## What this changes

Splitting the two questions removes the conflict.

**`caching_ttl` + `caching_stale_while_revalidate_ttl` is now the upper bound on how long a cached response survives**, for every caching dataset, derived unconditionally.

**`caching_stale_if_error` no longer extends that bound.** It decides what a *failed revalidation* does to an entry inside the window:

| entry age | `disabled` (default) | `enabled` |
|---|---|---|
| `< ttl` | serve | serve |
| `ttl..ttl+swr`, revalidation succeeds | serve, entry refreshed | same |
| `ttl..ttl+swr`, revalidation **fails** | entry dropped, next query reads the source | entry retained, served for the rest of the window |
| `> ttl+swr` | miss; the source's answer, including its error | same |

Both values get narrower and bounded. This is a behavior change in both directions and needs a release note, plus a docs update for the reworded parameter.

## A failed revalidation is now reported as one

`refresh_entry` returned `Ok(0)` when the origin answered with HTTP 429 or 5xx. That is how the HTTP connector surfaces a failure once its own `max_retries` backoff is exhausted — and since caching mode only accepts `http`/`https`/`localpod` sources, it is the dominant failure mode. Acting only on `Err` would have left the new `disabled` behavior silently inert in the common case.

It now returns a typed outcome separating a failing origin from a healthy origin with no rows, and only the former acts on `caching_stale_if_error`.

No retry is added here. The connector has already applied its configured policy (`is_retryable_status` covers exactly 5xx and 429, default Fibonacci backoff with `max_retries: 3`, then one final attempt whose response is returned as data). A second backoff would stack on an exhausted one, multiply worst-case latency, and hold the entry's in-flight revalidation slot for the duration — blocking every other revalidation of that key while learning nothing new. Operators who want more persistence before an entry is dropped have `max_retries` for it.

## Two consequences of the same bound

- The retention sweep interval was the retention period itself, so a long SWR window was swept once at startup and then not again for a year. It is now a fraction of the horizon, clamped to `[30s, 5m]`.
- An unset `caching_ttl` defaulted to 30s for retention but stayed `None` on the scan path, where the freshness check was skipped entirely and rows were served for as long as they existed. Retention masked this; making retention unconditional would have made it visible. Both paths now resolve to the same default.

## Implementation note

Invalidations travel the existing batched write channel as a third `CacheWriteOp` rather than through a side-channel delete, so they are ordered against concurrent writes for the same key by the single flush consumer instead of by a second lock protocol. A key refreshed in the same flush window keeps its fresh rows. The predicate is namespace-scoped, so one principal's failed revalidation cannot evict another's copy of the same key.

## Testing

Eight new tests. In `runtime-table`: a failed revalidation drops the entry when disabled and retains it when enabled; 429 and 5xx both count as failures; cross-principal isolation holds through invalidation; the delete predicate selects exactly what the upsert's exclusion filter drops (asserted by running both over the same rows, not by comparing their shapes); and the failure warning's text is asserted so a reword cannot quietly drop the resource, the consequence, or the docs link. In `runtime`: the retention horizon is identical for both `caching_stale_if_error` values — the direct regression for this issue — plus the default alignment and the sweep cadence.

Verified by negative control: mutating the flush task to ignore `Invalidate` (the pre-change behavior) fails exactly the three tests asserting the drop behavior and leaves the retention test passing.

Two bugs surfaced during the test run rather than in review, and are fixed here: `overwrite_accelerator` early-returns on empty batches and takes its schema from `batches[0]`, so an invalidate-only rewrite silently no-opped — `clear_accelerator` now covers the case where emptiness is the intended result, leaving the existing contract untouched for its other callers. And the `Err` arm of the revalidation task logged "background refresh failed" when the only way to reach it is a successful fetch whose write could not be enqueued, which points a reader at the origin when the fault is internal.

## Known gap

`AcceleratedTable` exposes no accessor for its retention policy, and the existing tests in `caching_mode.rs` drive a live external API, so the regression is pinned at the derivation rather than end-to-end through a running accelerator. `caching_retention_window` takes the whole `Acceleration` specifically so "retention ignores `stale_if_error`" is genuinely asserted rather than assumed, but the call site being unconditional is established by reading it, not by a test.

### DIFF
--- a/crates/runtime-acceleration/src/acceleration.rs
+++ b/crates/runtime-acceleration/src/acceleration.rs
@@ -374,14 +374,22 @@ impl Display for OnConflictBehavior {
     }
 }
 
-/// Behavior when a stale-if-error condition occurs in caching mode.
-/// When enabled, serves expired cached data if the upstream source returns an error.
+/// What happens to a cached entry in `refresh_mode: caching` when the source
+/// will not confirm it - the revalidation triggered once the entry passes
+/// `caching_ttl` errors, or answers with a transient status after the
+/// connector's own retries are exhausted.
+///
+/// This does not change how long an entry may be served. `caching_ttl` +
+/// `caching_stale_while_revalidate_ttl` is the upper bound either way; the
+/// choice is only whether a failed revalidation ends the entry's life early.
 #[derive(Debug, Clone, Copy, PartialEq, Default)]
 pub enum StaleIfError {
-    /// Do not serve stale data on error - propagate the error to the client.
+    /// Drop the entry as soon as a revalidation fails, so the next query reads
+    /// from the source and sees its error rather than an unconfirmed response.
     #[default]
     Disabled,
-    /// Serve expired data if the upstream source returns an error.
+    /// Keep serving the last successful response for the remainder of the
+    /// stale-while-revalidate window.
     Enabled,
 }
 
@@ -418,8 +426,13 @@ pub struct Acceleration {
 
     pub refresh_check_interval: Option<Duration>,
 
+    /// How long a cached response stays fresh. Past this age the next query is
+    /// served from cache and triggers a background revalidation.
     pub caching_ttl: Option<Duration>,
 
+    /// How much longer past `caching_ttl` a response may be served while it is
+    /// revalidated. Together the two are the upper bound on an entry's life:
+    /// beyond their sum it is neither served nor retained.
     pub caching_stale_while_revalidate_ttl: Option<Duration>,
 
     pub caching_stale_if_error: StaleIfError,
--- a/crates/runtime-table/src/accelerated/caching.rs
+++ b/crates/runtime-table/src/accelerated/caching.rs
@@ -20,7 +20,7 @@ use std::sync::atomic::AtomicI64;
 use std::time::{Duration, SystemTime};
 
 use arrow::array::StringArray;
-use arrow::array::{Array, ArrayRef, RecordBatch, TimestampNanosecondArray};
+use arrow::array::{Array, ArrayRef, RecordBatch, TimestampNanosecondArray, UInt64Array};
 use arrow::compute::cast;
 use arrow::datatypes::{DataType, SchemaRef, TimeUnit};
 use arrow_tools::format::SchemaDisplay;
@@ -47,6 +47,7 @@ use tokio::task::JoinHandle;
 use runtime_acceleration::dataupdate::StreamingDataUpdateExecutionPlan;
 use runtime_datafusion::execution_plan::schema_cast::SchemaCastScanExec;
 use runtime_request_context::CacheNamespace;
+use spice_table::{LayerWalk, peel_to};
 use util::expr::combine_exprs_balanced;
 
 /// Type alias for tracking in-flight revalidation requests.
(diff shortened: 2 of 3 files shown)


bjchambers:
Superseded by #13604, which takes a different route to #13525 and carries the one piece of this PR that is independent of that route.

**What #13604 does instead.** Rather than making `caching_ttl + caching_stale_while_revalidate_ttl` a horizon that `caching_stale_if_error` no longer extends, it keeps what `stale_if_error` means today — an expired entry stays available as fallback for a failing origin — and bounds the resulting growth with two new budgets, `caching_max_size` and `caching_max_items`. The remaining unbounded configuration (`stale_if_error` enabled with no budget) warns at load, naming the dataset. Redefining a shipped parameter was judged a bigger change than #13525 needs.

**What was carried over.** The `refresh_entry` fix, which is a defect independent of the retention argument: it returned `Ok(0)` when the origin answered 429 or 5xx, which is how the HTTP connector reports a failing origin once its own `max_retries` is exhausted, so a caller could not tell a failing origin from a healthy one with no rows. #13604 gives it a typed outcome, and goes further — the `stale_if_error` fallback in `handle_cache_miss` fired only on the `Err` arm, so on that same dominant failure mode an operator who enabled the setting was handed the origin's error body instead of the cached response. It now keys off the response's status.

Also fixed independently there: the sweep interval (clamped to 30s–5min rather than derived from the retention period), and the unset-`caching_ttl` divergence between the retention and scan paths — both paths now resolve through one `effective_max_age`.

**One thing worth flagging from reviewing this PR against that one.** The time-period retention here deletes rows by `_fetched_at < cutoff`, with no notion of which request a row belongs to. A cached response can span several rows — a paginated one is fetched a page at a time, each page stamped with its own `_fetched_at` — so an entry straddling the cutoff loses its older pages while the newer ones remain, and the read path, which judges freshness over the rows that are left, can then serve that remainder as though it were the whole response. #13604 evicts whole entries by key for exactly this reason, and has a regression test whose negative control is restoring the range delete.

PR #13576, opened 2026-08-26
Move shared items out of runtime's private http module into the crates that own them

bjchambers:
Three related moves, each taking something out of `runtime`'s private `http` module and putting it in the crate that owns its inputs.

## 1. The read-only principal gate

`current_principal_requires_read_only` decides whether a request executes in read-only mode. It existed **twice** — once in `runtime::http::v1`, and again with its helper inlined in `runtime-tools`. Both fed live paths: HTTP `/v1/sql`, Flight, the tool-backed scalar UDF, and the tools surface.

The two copies were semantically identical, so nothing was behaving wrongly. The hazard was forward-looking: adding a group name to one and not the other would have made those surfaces silently disagree about who may write.

The pure predicate `principal_has_write_access` now lives in `runtime-auth`, beside the `AuthPrincipal` impl that mints the `read` / `read_write` group names it tests. The context-reading `current_principal_requires_read_only` lives in `runtime-request-context`, beside `impl AuthRequestContext for RequestContext`. The two halves cannot share a home — `runtime-request-context` depends on `runtime-auth`, so reading the request context from `runtime-auth` would close a dependency cycle.

Adds unit tests for the group matrix, which had none, including that matching is exact: `writer`, `readwrite` and `write_read` do not grant write access.

## 2. `SubmitQueryRequest`

The request body of `POST /v1/queries` is a versioned public API contract, but it was defined inside the private `http::v1` module. Two things worked around that: `jobs` reached back up into `http::v1::queries` for the type it is handed, and `runtime` carried a `pub mod http_types` whose only content was a re-export of this one type, so the cluster integration tests could name it.

It now lives in `runtime-api-types`, which exists for exactly this — the canonical `/v1` request and response types shared by the runtime's handlers and the CLI, with the stability guarantee those types need. The struct is unchanged, and that crate already carries the `openapi` feature gating the `utoipa::ToSchema` derive, which `runtime` already forwards, so no feature wiring changes. `runtime::http_types` is deleted.

## 3. `run_sql` / `run_sql_with_read_only`

These build a query, run it, and collect the result into memory. Nothing about that is HTTP, and nothing in `http` called them — the only caller anywhere was the prompt worker in `scheduling`, which had to reach into the private `http::v1` module to get at a query helper. They now sit in `crate::datafusion::query`, next to the `QueryBuilder` their bodies are made of.

## Why

`runtime::http` is a private module, but four other modules were reaching into it for things that were never HTTP concerns. Beyond the duplication above, that coupling is what keeps the HTTP and Flight surfaces entangled with the orchestrator (see `docs/dev/crate_layering.md`).

## Notes for review

- **Visibility.** The first change widens: the predicate goes from `pub(crate)` inside a *private* module to `pub` in two shared crates. `runtime-tools` already wanted it badly enough to copy it, so one published definition seemed better than two private ones. The second change is a net *narrowing* (`runtime::http_types` deleted). The third stays `pub(crate)` deliberately — `mod http` is private, so a plain `pub` in the public `datafusion::query` would have newly exposed it.
- `runtime-tools` no longer depends on `runtime-auth` at all; deleting the duplicate removed its last use of that crate.
- **Not done here:** `run_sql_with_read_only`'s `read_only` parameter is dead — its sole caller is `run_sql`, which always passes `false`. Collapsing the two is a signature change that deserves its own review.

## Verification

- `cargo clippy` lib and `--tests` passes, with the gate's flag set, over `runtime`, `runtime-auth`, `runtime-request-context`, `runtime-api-types` and `runtime-tools` at the workspace-resolved feature set (37 features, including `openapi`) — clean.
- `runtime-auth` unit tests pass.
- `cargo fmt --all --check` plus the crate-layering, module-reachability, table-layer and rust-gate-path guards all pass.
- `SubmitQueryRequest` verified byte-identical to its previous definition — fields, serde attributes and derives.


Jeadie (dismissed):

claudespice:
Resolved the `trunk` conflict — `CONFLICTING/DIRTY` → `MERGEABLE/BLOCKED`. Merge commit `e76c566672`, pushed fast-forward `5c580110c2..e76c566672` (no force, no rebase).

The branch was `diverged ahead=3 behind=104` off merge-base `d8a32b6eb2`, and **exactly one hunk conflicted**: `crates/runtime-tools/Cargo.toml`.

### The hunk, and why `--theirs` would have been wrong

`trunk` pinned `rmcp`'s feature set on this crate and gated `mcp` on `tools/mcp`; this branch **deleted** `runtime-auth = { path = "../runtime-auth" }`, having moved `principal_has_write_access` into `runtime-auth` itself. Both sides rewrote the same three lines, so Git could not tell that the two intents are independent.

Taking `trunk`'s side wholesale restores the dependency this branch deleted — the merge-restores-what-the-child-deleted case in `docs/dev/stacked_prs.md`. That is not a matter of taste here; it is measurably wrong, because the merged `Cargo.lock` records no `runtime-auth` under `runtime-tools`:

```
# with the runtime-auth line restored (the naive --theirs resolution):
$ cargo metadata --locked
error: cannot update the lock file .../Cargo.lock because --locked was passed to prevent this
rc=101

# as resolved:
$ cargo metadata --locked
rc=0
```

So the resolution is `trunk`'s `rmcp` entry (features + comment) with no `runtime-auth` line. `trunk`'s `mcp = ["dep:rmcp", "tools/mcp"]` change sits outside the conflict and merged on its own. `git grep runtime_auth -- crates/runtime-tools/src/` returns nothing on the merged tree, and did return a hit on both `d8a32b6eb2` and `origin/trunk` — the branch is what removed the last use.

### Why the rest of the merge is low-risk, measured

Only **three** files are touched by both sides: `Cargo.lock`, `crates/runtime-tools/Cargo.toml` (the conflict), and `crates/runtime/src/lib.rs`. The last merged automatically and the two edits are in different regions — this branch removes the `http_types` re-export at ~:106; `trunk` adds error variants and partition-ack logic elsewhere.

The move is the thing a base merge could silently break, so I checked it by symbol rather than by eye. All four symbols this branch relocates are present at their new homes on the merged tree (`run_sql`/`run_sql_with_read_only` → `datafusion/query.rs:2811`, `SubmitQueryRequest` → `runtime-api-types/src/v1/queries.rs:24`, `principal_has_write_access` → `runtime-auth/src/traits.rs:57`, `current_principal_requires_read_only` → `runtime-request-context/src/context.rs:508`), and `trunk`'s 104 commits add **no** reference to any of them. The one apparent hit — a `run_sql` added in `tests/acceleration/retention_cayenne.rs` — is a file-local test helper wrapping `run_query`, not this one. `git grep http_types` over `crates/ bin/ tools/` is empty.

### Validation, and what it does not cover

`cargo metadata --locked` rc=0 and `cargo fmt --all --check` rc=0 on the merged tree. My tree differs from Git's own automatic merge tree (`d152756bb`) in **exactly one file**, the conflicted manifest.

**I did not compile it, and I am not claiming otherwise.** A local build was refused on measurement, not skipped for convenience: 85 GiB free at 96% full, and the per-worktree `target/` dirs on this machine currently measure 26–301 GiB, so a cold `cargo check` over this crate's graph does not fit and would take out a peer's in-flight sign-off with ENOSPC. The compile is therefore remote sign-off [run 33895748118](https://github.com/spiceai/spiceai/actions/runs/33895748118), and `Attestation` is what stands behind this head — not any of the checks above.

### Two things for you

- **Your approval was dismissed by this push** (`APPROVED` → `REVIEW_REQUIRED`). The PR was unmergeable while it conflicted, so this is a net improvement, but it does need re-approving.
- The stale red `enforce-pull-with-spice` on the old head was **superseded, not live**: that head carried two runs of it, `failure` at `2026-08-26T21:22:10Z` and `success` at `2026-08-30T07:39:02Z`, and the status rollup surfaces the older one. Newest-per-check on the old head was green.

### Review gate

- **Adversarial review: skipped — no engine ran.** `codex` exited non-zero with the engine's own reason (*"Your workspace is out of credits"*), `grok` is `not-installed` here. Receipt `engine=none`, `exit=1`. Not a clean pass.
- **Security review: by hand.** The change adds no code — it removes one manifest line and takes a feature list. Nothing new parses input, deserializes, logs, routes or reaches the network, and the change is a strict *reduction* in this crate's dependency surface. `rmcp`'s features come from `trunk` unchanged.
- `/simplify`: not applicable — the diff is one manifest hunk.

### Correction to the merge commit message

`e76c566672`'s message credits `trunk`'s `rmcp` change to `#13756`. **That is wrong — there is no PR #13756.** The change came in with `c2fc8f9d16`, *"fix(runtime-tools): declare what the mcp feature actually needs (fixes #13648)"*, [#13733](https://github.com/spiceai/spiceai/pull/13733). I checked it after pushing rather than before, which is the wrong order; the commit message cannot be corrected without a force-push, so the correction lives here. Nothing else in the resolution or in this comment depends on that number, and the squash message on merge is composed from the PR rather than from this merge commit.

Issue #12933, opened 2026-08-12
Share interned SchemaRefs wherever schemas are stored per-item


bjchambers:
## What

Audit the places where a `SchemaRef` is stored per-item and share/intern a single allocation instead of keeping a copy per item.

## Why

A concrete instance was measured in the SQL results cache (filed separately): per-entry cost there scales at **~195 bytes per column**, so on a 200-column table each cached entry pays ~40 KB for a schema that is identical across every entry derived from that table. At 40,000 entries that is 1.5 GiB of duplicated schema.

The pattern generalises: a schema is immutable, usually identical across every item in a collection, and frequently stored alongside items that are individually small. Wherever that holds, one shared `Arc<Schema>` is strictly better than N copies — and the wider the table, the larger the win.

## Where to look

Starting points, not an exhaustive list:

- `crates/cache/src/result/query.rs` — `CachedQueryResult::schema` (the measured case)
- `crates/cache/src/result/search.rs` — `CachedSearchResult` holds `Arc<HashMap<TableReference, CachedAggregationResult>>` and `Arc<HashSet<TableReference>>`; worth checking whether the aggregation results carry per-item schemas
- Cayenne's in-memory CDC tier and any other place holding `Arc<Vec<RecordBatch>>` long-term — every `RecordBatch` carries its own `SchemaRef`, so a tier of many small batches duplicates it per batch
- The in-memory accelerator and in-memory search index, which store batches for the process lifetime

`RecordBatch` already shares a `SchemaRef` among its columns, so the concern is specifically **collections of batches or entries**, where the same schema is re-held once per element.

## Suggested approach

1. Identify the sites where the number of stored items can grow large and the schema is provably identical across them.
2. Introduce a small per-table (or per-source) schema registry that hands out a shared `Arc<Schema>`, and resolve through it at construction time.
3. Add `Arc::ptr_eq` assertions as regression tests at each site, so a future change that reintroduces per-item allocation fails loudly.

## Notes

This is an optimisation, not a correctness issue — schemas are immutable, so duplication costs memory rather than risking wrong results. Priority should follow measured impact: the results-cache case has numbers behind it, the others are candidates that have not been measured.

Where a shared schema is not achievable, the fallback is to at least **account** for the per-item copy, so memory budgets that rely on a size estimate are not silently wrong.

bjchambers:
Motivated by the measured case in #12932 (~195 bytes per column, per SQL results cache entry) and the accounting consequences in #12931.

[...]

PR #13577, opened 2026-08-26
Share one schema across the inline-data cache's entries

bjchambers:
Refs #12933, which is being closed as won't-fix. This carries the one piece of it that stands on its own.

## What was wrong

Each inline entry is its own Arrow IPC blob, and a blob carries its own copy of
the schema, so `deserialize_ipc_to_batch` builds a fresh `Schema` per entry.
The cached view retains all of them — flattened and per-entry — for as long as
the entry is live.

Unlike the SQL results cache, nothing bounds that total: the inline view has no
per-entry byte budget, so the cost grows with the number of inline entries
rather than being capped.

`adapt_inlined_batches_to_live_schema` already had the batches in hand, but
shared an allocation only on the path it does *not* take — when the decoded
fields already match the live schema it returns the batches untouched, which is
the common case. Only an actual schema *change* triggered the cast that shares
the table's `Arc`.

## What changed

Batches whose schema is identical to the table's now point at the table's own
allocation. One table has one schema, so this costs a pointer and leaves nothing
to reclaim later — no pool, no registry, no background sweep.

Sharing requires **full** equality, not the fields-only test that selects the
fast path. A decoded schema may carry metadata the live one does not, and
replacing it would silently drop that, so anything short of identical is left
exactly as it arrived.

Measured on a four-entry inline corpus: four distinct schemas before, one after.

## The mem tier, which turned out not to need this

The in-memory CDC tier was the other candidate and is deliberately unchanged.
Its write path already normalises every incoming batch onto the table's schema,
so the whole tier shares one allocation however many applies land in it —
interning there would have added work to the hottest ingestion path for no gain.

That is worth a test rather than a change, so this adds one. If the
normalisation ever stops, the tier starts holding one schema per batch and
`mem_tier_segments_share_the_tables_one_schema` fails instead of the regression
passing unnoticed.

## Testing

`cached_inlined_view_entries_share_the_tables_schema` drives four separate
inserts through the real inline path and asserts every decoded batch points at
the table's schema. Verified by mutation: reverting the fast path to its previous
`return Ok(batches)` fails it. The mem-tier test passes with and without that
change, which is what makes it a characterization of existing behaviour rather
than a test of this one.

Full `cayenne` suite green (1,105 passed).

### DIFF
--- a/crates/cayenne/src/provider/table.rs
+++ b/crates/cayenne/src/provider/table.rs
@@ -23215,6 +23215,47 @@ impl CayenneTableProvider {
     ///
     /// A corpus already at the live width (the overwhelmingly common case) short-
     /// circuits on a field comparison and is returned untouched.
+    /// Points batches at the table's own schema allocation where they already
+    /// carry an identical one.
+    ///
+    /// Every inline entry is decoded from its own Arrow IPC blob, and a blob
+    /// carries its own copy of the schema, so a cached view over N entries would
+    /// otherwise retain N identical `Schema`s — and it retains them for as long
+    /// as the entry is live, with no per-entry budget bounding the total. One
+    /// table has one schema, so pointing the batches at it costs a pointer and
+    /// leaves nothing to reclaim later.
+    ///
+    /// Requires full equality, not the fields-only test that selects the caller's
+    /// fast path: a decoded schema may carry metadata the live one does not, and
+    /// replacing it would silently drop that. Anything short of identical is
+    /// left exactly as it arrived.
+    fn share_live_schema(
+        mut batches: Vec<RecordBatch>,
+        live_schema: &SchemaRef,
+    ) -> Vec<RecordBatch> {
+        for batch in &mut batches {
+            if Arc::ptr_eq(batch.schema_ref(), live_schema)
+                || batch.schema_ref().as_ref() != live_schema.as_ref()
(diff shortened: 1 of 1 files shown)


crates/cayenne/src/provider/table.rs
Jeadie, line 23274: nit/question: we dont bother doing same here since `arrow_tools::record_batch::try_cast_to` similaririly shares `Arc::clone(&live_schema)` ?

Jeadie (approved):

claudespice:
**Conflicts with `trunk`, and the conflict is too large to resolve on your behalf.**

Measured with `git merge-tree --write-tree origin/trunk ben/cayenne-inline-schema-sharing` (no checkout, nothing pushed): one conflicting file, **two hunks** —

```
CONFLICT (content): Merge conflict in crates/cayenne/src/provider/table.rs
```

| hunk | `origin/trunk` side | branch side |
|---|---|---|
| 1 | 567 lines | 15 lines |
| 2 | 794 lines | 32 lines |

Both sides rewrote the same regions of `table.rs`, so resolving means deciding which behaviour is right rather than splicing two edits together — that is your call, not a merge's. Flagging it because the PR has been sitting for six days and the conflict only grows while trunk moves; nothing here has been pushed to your branch.

PR #13580, opened 2026-08-26
fix(cache): reject plan and search cache entries that straddle an invalidation

krinart:
## 📝 Summary

The read-time invalidation check added in #12703 covered only the SQL results cache. The logical-plan and search caches relied on moka predicate invalidation alone, which cannot see an entry stored *after* the invalidation ran — so a plan or search result whose read straddled a refresh, DML write, or schema change was resurrected into the cache and served until TTL (a hard-coded hour for plans). A stale cached plan also pins pre-invalidation table state through the `TableSource` references in its scans, and results executed from it re-enter the results cache with a fresh timestamp — defeating that cache's own staleness check.

This PR extends the same clock mechanism to all table-scoped caches:

- Each cache owns a `TableInvalidationClock`, stamped before entries are removed.
- Serving lookups go through a new `TabledCacheProvider::get_raw_key_if_fresh`, which treats an entry whose tables were invalidated at or after its read began as a miss.
- Cached values record when their read began: the plans cache stores `CachedLogicalPlan` (plan + planning start), and `CachedSearchResult` gains `read_started_at`.
- Full clears (`invalidate_all`, e.g. schema-change plan-cache clears) stamp a conservative floor covering every table.

Covered by a new integration suite (`crates/cache/tests/invalidation_orderings.rs`) asserting both orderings on all three caches, full-clear straddles, and no over-rejection, plus a runtime test proving `get_or_create_logical_plan` replans instead of serving a stale racing entry.

## 🔗 Related

- #12703 introduced the clock for the SQL results cache.

### DIFF
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -453,22 +453,46 @@ impl Caching {
     ///
     /// This is purposely eager, as an invalidated cache is better than a stale one.
     ///
+    /// The caches run concurrently, not in sequence, for two reasons beyond
+    /// latency:
+    ///
+    /// - Every cache stamps its invalidation clock synchronously before its
+    ///   first await, so all stamps land before any (possibly slow) entry
+    ///   removal is waited on. Awaiting one cache's removal before stamping
+    ///   the next would leave the later caches serving pre-write entries for
+    ///   the whole wait — and a stale plan served in that window can feed
+    ///   freshly-stamped stale results into the already-stamped results
+    ///   cache, where they outlive the invalidation.
+    /// - A failure in one cache must not skip the others: a skipped cache
+    ///   never stamps its clock or removes its entries, and serves pre-write
+    ///   data until TTL. Every invalidation runs to completion; the first
+    ///   error is reported after all of them have run.
+    ///
     /// # Errors
     ///
     /// If the cache invalidation fails for any of the caches.
     pub async fn invalidate_for_table(&self, table_ref: TableReference) -> Result<()> {
-        if let Some(results_cache) = &self.results {
-            results_cache
-                .invalidate_for_table(table_ref.clone())
-                .await?;
-        }
-        if let Some(plans_cache) = &self.plans {
-            plans_cache.invalidate_for_table(table_ref.clone()).await?;
-        }
-        if let Some(search_cache) = &self.search {
-            search_cache.invalidate_for_table(table_ref).await?;
-        }
-        Ok(())
+        let (results, plans, search) = futures::join!(
+            async {
+                match &self.results {
+                    Some(cache) => cache.invalidate_for_table(table_ref.clone()).await,
+                    None => Ok(()),
+                }
+            },
+            async {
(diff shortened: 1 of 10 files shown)


crates/cache/src/result/plan.rs
```diff
@@ -0,0 +1,64 @@
...
+use std::collections::HashSet;
+use std::sync::Arc;
+use std::time::Instant;
+
+use datafusion::logical_expr::LogicalPlan;
+use datafusion::sql::TableReference;
+
+use crate::{AsTableRefs, ReadStartedAt};
+
+/// A cached logical plan, stamped with when planning began.
```
bjchambers, line 26: It feels like we should try to understand why caching the logical plan is incorrect. It *seems* like it should be re-usable except in the case where the underlying table definition changed.

crates/cache/src/lib.rs
```diff
@@ -493,7 +559,7 @@ impl Caching {
 /// self-healing (later invalidations repopulate per-table entries), at the cost
 /// of some lost cache entries in the moments after a collapse.
 #[derive(Default)]
-struct TableInvalidationClock {
+pub(crate) struct TableInvalidationClock {
```
bjchambers, line 562: Is the invalidation direction appropriate for everything? Should this instead be marking things stale? Or is the plan to make things consistently invalidate, and then examine the staleness?

grokspice:
This PR is `CONFLICTING` with `trunk`. Not resolving it here — the conflict looks like a genuine design tension rather than a mechanical one, so it needs your judgment.

**Scale:** 11 conflict hunks across 4 files (3-way merge against `origin/trunk` `968ef1fc64`):

| File | Hunks |
|---|---|
| `crates/cache/src/lib.rs` | 6 |
| `crates/cache/src/result/search.rs` | 3 |
| `crates/cache/src/lru_cache.rs` | 1 |
| `crates/runtime-search/src/search_engine.rs` | 1 |

**Why it is not mechanical.** Two commits landed on `trunk` since this branch's merge base, both rewriting the same cache-validity logic this PR changes:

- `2c00184694` — Make the SQL results cache hold what it says it holds (#13908)
- `2f9a874a97` — feat(caching): serve results stale after an acceleration refresh instead of evicting them (#13618)

The second one is the reason to stop and ask rather than merge. This PR's stated goal is to **reject** cache entries that straddle an invalidation; #13618's stated goal is to **serve** results after an acceleration refresh instead of evicting them. Those are opposite dispositions for an entry whose validity window crosses a refresh, so whichever side of each hunk gets taken silently decides which behaviour ships — and a wrong pick here is a stale-results bug wearing a green check rather than a compile error.

**What would help:** whether #13618 already covers the straddling case this PR targets (in which case the remaining scope may be much smaller than the current diff), or whether the two are addressing genuinely different windows and both are needed. That is a question about intended cache semantics, which is yours and the reviewer's to answer, not something to infer from the diff.

Happy to do the mechanical part of the merge once the disposition is settled.

Issue #12528, opened 2026-08-05
DuckDB acceleration fails with `Unknown TimeZone '+00:00'` for fixed-offset timestamp columns


phillipleblanc:
## Describe the bug

A dataset accelerated with `engine: duckdb` fails to load and stays **permanently unhealthy** when its refresh time column carries an Arrow **fixed-offset** timezone (`Timestamp(_, Some("+00:00"))`) instead of a named IANA zone:

```
Not implemented Error: Unknown TimeZone '+00:00'!
```

The refresh retries indefinitely and the dataset never becomes ready, so the runtime reports a reduced ready count (e.g. `6/7 ready, 1 unhealthy`).

Iceberg maps **every** `timestamptz` column to the fixed-offset Arrow timezone `+00:00`, so any Iceberg dataset with a `timestamptz` time column accelerated into DuckDB with `refresh_mode: append` is affected.

## Root cause

`Dialect::timestamp_at_time_zone_to_sql` renders `CAST(expr AS Timestamp(_, Some(tz)))` as `expr AT TIME ZONE '<tz>'`. `DuckDBDialect` does not override it — `MySqlDialect`, `SqliteDialect` and `BigQueryDialect` all do. DuckDB resolves `AT TIME ZONE` through its ICU extension, which only knows **named** IANA zones, so it rejects every fixed-offset spelling.

The cast reaching the unparser is built by `TimestampFilterConvert` (`crates/util/src/timestamp_filter.rs`, `TimestampFormat::Timestamptz` arm), which carries the accelerator schema's timezone spelling verbatim. The append-refresh dedup query (`except_existing_records_from`, `crates/runtime/src/accelerated_table/refresh_task.rs`) pushes that filter into the DuckDB accelerator, where it is unparsed and rejected.

There is a second, quieter problem in the same rendering. An Arrow timestamp is a **UTC instant** whose timezone is display metadata, so `CAST(x AS Timestamp(_, Some(tz)))` preserves the instant. SQL `AT TIME ZONE` is a zone **conversion**. Rendering an internally-constructed cast as a conversion is wrong independently of the error: for a *named* zone DuckDB accepts it and silently shifts the comparison instead of failing.

## Scope — which spellings fail

Verified directly against DuckDB v1.5.2:

| timezone | result |
|---|---|
| `+00:00`, `+0000`, `+00`, `-05:00` | `Not implemented Error: Unknown TimeZone` |
[...]

PR #13604, opened 2026-08-27
feat(caching): bound a caching accelerator by size, count and entry lifetime (closes #13525)

### LINKED ISSUE #13525: caching_stale_if_error: enabled silently disables retention, so a caching accelerator never evicts

bjchambers:
Closes #13525.

## The problem

A `refresh_mode: caching` accelerator had nothing bounding what it held.

Retention was *derived* from `caching_ttl + caching_stale_while_revalidate_ttl` as a single time period over `_fetched_at`, and only when `caching_stale_if_error` was disabled. So enabling `stale_if_error` built no policy at all: nothing was ever evicted, however long the run lasted and whatever `caching_ttl` said. Even with it disabled, the derived `check_interval` was that same period, so a long stale-while-revalidate window scheduled roughly one check ever. Nothing capped the acceleration by size or by count either — there was no way to say how much a cache may hold.

Two more problems sat alongside it:

- **Replacing one cache entry rewrote the whole acceleration.** The write path read every cached row back, filtered it in memory and overwrote the table, so the cost of caching one response grew with everything already cached.
- **`caching_stale_if_error` did not fire on the failure it exists for.** It keyed off the fetch returning an `Err`, but once the HTTP connector has exhausted its own `max_retries` it reports a failing origin as a *successful* fetch whose rows carry a 429 or 5xx status. That is the dominant failure mode of the connectors caching mode accepts, so an operator who enabled the setting was handed the origin's error body instead of the cached response they asked to fall back to.

## The approach

**Two parameters bound the acceleration**, both refusing an unparseable value rather than falling back silently — the parameter *is* the bound, and a silent default would leave an operator believing a limit they set is in force:

- `caching_max_size` — byte budget, e.g. `512MiB`. Matches the `max_size` suffix the SQL-results, search-results and embeddings caches use.
- `caching_max_items` — row budget.

`caching_ttl` is now also accepted as `caching_item_ttl`, spelling the `item_ttl` suffix those same caches use. Both are read, so neither spelling is silently ignored, and setting them to different values is an error rather than a coin flip.

**Nothing new is stored.** An entry's deadline is `_fetched_at + caching_ttl`, which is dataset configuration the sweep already has; its size is `octet_length` over its own text columns, which the engine computes during the aggregate the sweep already runs. A stored copy of either would be a denormalisation that can go stale against what it describes, and would have forced every existing caching accelerator to be rebuilt. The storage schema is unchanged.

**Eviction is entry-granular, never a timestamp range.** A cached response can span several rows — a paginated one is fetched a page at a time, each page stamped with its own `_fetched_at` — so a `_fetched_at < cutoff` delete can cut an entry in half and leave the remainder to be served as though it were the whole response. Entries are ranked by their oldest page and named by key, so all of an entry's rows go or none do. The namespace is part of that key: two principals can hold entries under the same request, and evicting by request alone would take rows the budget never charged to that entry.

**...but a range predicate is used where it is provably exact.** Naming entries individually is capped at 512 per sweep to bound predicate width, and that cap is a hard limit rather than a taste: DataFusion walks expression trees recursively, and a 50,000-wide `OR` overflowed the stack and aborted the process. A cache ingesting hundreds of entries per second outruns 512 per sweep forever. When every doomed entry's newest row predates every survivor's oldest row, nothing straddles the boundary and a range over the doomed entries' own fetch times selects them exactly, whatever their number. A workload that fetches new keys over time separates cleanly in fetch order, so this is the ordinary case; the per-entry form remains for the interleaved one.

That range is closed at **both** ends, which is what makes it safe to build from a ranking taken without the write lock. An open `_fetched_at <= cutoff` would also select whatever arrived after the ranking carrying a fetch time below it — a row the sweep never measured and never decided about, and if its entry's other rows were newer, half the entry would go and the rest would be served as though it were the whole response. Bounding below at the oldest row the ranking actually saw means the delete can only reach rows that were there when the decision was made.

**Every comparison against `_fetched_at` goes through the converter the retention loop's own time filter uses**, rather than a literal written here, so the predicate is built for the type the accelerator actually stores the column as — Cayenne keeps it in microseconds, and a DuckDB cast carrying a timezone has to be rendered as `AT TIME ZONE` (#12528).

**It runs as a retention filter, not a second loop.** A new `DataRetentionFilter::Computed(Arc<dyn RetentionPredicate>)` is resolved on every tick, so a policy that must *measure* the table before it can decide — a byte or row budget is knowable only by aggregating what is stored — can be a retention filter rather than a loop beside the existing one. `retention.rs` learns only "ask this for an expression and let it have the last word"; every request-shaped detail stays in the caching module. Cache eviction shares the retention loop's ticker, write lock, index-aware delete and results-cache invalidation.

The policy is a field on `Retention` rather than another variant of its filter list, because it does not compose the way those do: they are OR'd together, while this one is handed their combined result and may widen, narrow or replace it. In the list, "at most one, applied last" would be a convention the loop enforces by hand, and a second computed policy would silently disable the first.

One computed policy owns the whole decision and returns a single predicate, because the three things it applies are not independent: what expiry removes changes whether a budget is exceeded at all, and two filters resolved against the same snapshot would each select entries the other was already deleting. One delete per tick also means one invalidation, so there is no window in which rows are gone while results built on them are still cached.

**`retention_period` and `retention_sql` are honoured in caching mode**, evaluated per entry rather than per row — a rule matching one page of a paginated response takes the whole response. Previously the derived cache retention silently overrode them.

**The sweep interval is clamped to 30s–5min** rather than derived from the TTL, since a budget is only as tight as the sweep that enforces it.

**A sweep ranks without holding the write lock**, because each entry's delete predicate is bounded to the rows it held when it was ranked. A refresh replaces an entry by appending rows with a newer `_fetched_at`, so an entry refreshed between the ranking and the delete matches nothing and survives whole — a sweep cannot turn a response that was just repaired into a cache miss. Holding the lock across a full aggregate instead would stall every cache write for as long as that scan takes.

**Replacing a cache entry no longer rewrites the acceleration.** A write that supersedes stored rows issues a `DELETE` for them and appends, which DuckDB, Cayenne, SQLite, Turso and the in-memory accelerator all implement. An accelerator without `delete_from` reports that while *planning*, so the fallback to the old read-filter-write path cannot leave a half-applied replacement behind.

A write for a key nothing holds yet appends with no delete. That distinction is not a micro-optimisation: on Cayenne a `delete_from` first checkpoints the inline memtable to a file, so deleting on every write — including the overwhelming majority that cannot collide with anything — collapsed measured ingest from thousands of entries per second to about ten.

Deleting *before* appending is the order that fails safely. Without a transaction one statement can land alone: if the append fails the entry is gone and the next read re-fetches it, which is a cache miss and always a correct answer, at the cost of that entry's `stale_if_error` fallback. Appending first would leave the key holding both responses if the delete failed, and a duplicated source row is a *wrong* answer that queries go on returning. A cache may lose an entry; it may not invent rows.

**One key, one writer, for as long as a writer's view of it can go stale.** The claim that enforces this is taken *before* the origin is asked, not after, and released when the guard holding it drops. That ordering is the whole point: a reader scanning during a replacement's delete-then-append gap sees no rows and reads a miss, and by the time its own fetch returns the replacement may have landed and released. Claiming only then would let it append beside a response it never saw. Every writer takes one — the miss path, stale-while-revalidate, and the periodic stale-row refresh, which writes directly to the accelerator and so opens the same gap. Release-on-drop is what keeps a cancelled query, a disconnected client or a fetch that never returns from leaving a key claimed for the life of the process, which would refuse every later write *and* revalidation for it and make one transient failure permanent. The flush enforces the same invariant a second time by writing one request per key.

**`caching_stale_if_error` now acts on a failing origin.** Revalidation returns a typed outcome that separates a failing origin from a healthy one with no rows, and the fallback keys off the response's status rather than only off an `Err`. The same classification stops the periodic background refresh from replacing a good entry with the origin's error body. With `stale_if_error` disabled — the default — nothing changes.

**The background stale-row refresh now stamps what it writes.** It replays a stale request against the source and writes the result back directly, bypassing the batched write path that does the stamping, so those rows landed with no namespace tag. It also now scopes the replacement to the namespace the stale rows came from, so refreshing one principal's copy no longer deletes another's.

## Configurations that cannot be fully enforced say so at load

Naming the dataset, so an operator learns it from the log rather than from unexplained growth: a schema with no request columns (entries cannot be identified, so the budgets cannot be enforced); a column whose size `octet_length` cannot measure (the byte budget under-counts); and `caching_stale_if_error: enabled` with no budget and no `retention_period`/`retention_sql`, which is the one remaining configuration nothing evicts from — expired entries are deliberately kept as fallback for a failing origin.

## Measured against real accelerators

Run locally against a synthetic origin serving incompressible bodies, driven to millions of distinct keys. Two of these findings are the reason for commits in this PR that no unit test would have produced.

| run | engine | config | written | held | on disk |
|---|---|---|---|---|---|
| bounded | DuckDB file | `caching_max_size: 32MiB`, `stale_if_error: enabled` | 30,000 entries | 10,416 | 17.3 MB |
| bounded | Cayenne file | `caching_max_size: 32MiB`, `stale_if_error: enabled` | 30,000 entries | 10,418 | 1.6 MB |
| unbounded | DuckDB file | no budget, `caching_ttl: 24h` | 2,500,000 entries | 2,500,000 | 11.75 GB |

The bounded runs set `stale_if_error: enabled`, so expiry is deliberately off and `caching_max_size` is the only thing evicting — the configuration that previously built no policy at all. The unbounded run sustained 2.5M entries and 11.75 GB across 84 minutes with no errors and exactly 1.000 origin fetches per entry, which is what says the write path is not re-fetching or duplicating under load.

Two defects surfaced only here:

- **Every sweep failed.** `entry_predicate` matched a NULL key component with `IS NULL`, and DuckDB's delete translator rejects that expression outright — `External error: Expression not supported _fetched_at IS NULL` — so one such term failed the whole multi-entry delete and a 1 GiB budget evicted nothing from a 1,000,000-entry cache. The tests delete through `MemTable`, which evaluates arbitrary DataFusion expressions, so the double was more capable than the engine. There is now a test asserting on the *shape* of every predicate this module builds rather than on what a permissive double does with it.
- **Eviction could not converge.** With the 512-per-sweep cap as the only mechanism, the deficit grew monotonically — 203k, 457k, 615k, 666k entries behind. Raising the cap to 50,000 aborted the process with a stack overflow, which is what established the width limit as hard. The range predicate above is the fix.

## Known limits

Two costs are deliberate rather than solved, and are stated here rather than left to be discovered:

**A sweep asks the accelerator, so it scales with what is cached.** The ranking is a `GROUP BY` over every stored row, and with a byte budget set it also sums `octet_length` over the payload columns, which reads every cached body. Without a byte budget that measurement is never requested, so a TTL-only cache pays for the grouping and not for the bodies. A measured run swept a 2,500,000-entry DuckDB acceleration every five minutes for 84 minutes without failing, but the ranking materialises one record per entry while it runs. Deferring each entry's key strings until a delete actually names it — at most a few hundred, and none at all on the range path — would remove most of that and is the identified next step; it is left out of this PR because it rewrites how delete predicates are addressed, which is the highest-consequence code here. Keeping a running tally beside the table instead is the alternative this design exists to avoid.

**The byte budget does not charge for a response's header map.** It is a `Map`, and no expression here can weigh one. An origin sending unusually large or numerous headers therefore holds bytes outside the ceiling. This is documented on `caching_max_size`, where an operator setting a budget will look, rather than raised as a load-time warning — that warning would fire for every HTTP caching dataset with a budget, about a column the operator did not choose to store and cannot stop the connector storing, which is how a warning stops being read.

## Testing

Eviction is driven end to end against a real accelerator rather than asserted on its helpers, because deciding correctly and then deleting the wrong rows is the failure being guarded. Covered: entries past `caching_ttl + swr` go and live ones stay; a multi-page entry expires whole or not at all; the stale-while-revalidate window holds an entry past its TTL; an unset `caching_ttl` resolves to the same default the read path uses; each budget evicts least-recently-fetched entries; the byte budget charges for the payload each entry actually holds; a multi-page entry is evicted whole; a `retention_period`/`retention_sql` rule takes whole entries rather than matching rows, and composes with expiry and budgets without double-charging; `stale_if_error` keeps expired entries while a budget still evicts; a cache within its budgets is untouched; an entry refreshed after the ranking is not evicted; one sweep evicts far more entries than a single predicate could name; and expiry still selects the right rows where the engine stores microseconds rather than nanoseconds, as Cayenne does.

The claim's lifecycle is tested where its guarantees live: a dropped claim releases its key and a queued one does not, a second reader missing the same key does not enqueue a second copy, a failing origin does not leave the key claimed, and the periodic refresh leaves an entry alone when another writer already holds it. The range predicate is asserted on its shape — that it is bounded below as well as above — because the lower bound is the whole guarantee and dropping it would still pass every eviction test in the file.

The write path is tested by watching the calls the accelerator receives — "the right rows ended up stored" is true of both strategies and would not distinguish them. A delete-capable accelerator must delete and append without scanning or overwriting; one without `delete_from` must reach the same stored state through the fallback; a fresh key must append without paying for a delete; two writes for one key in a flush must store the newest only; and an empty filter set — which as a `DELETE` would remove the entire cache — is refused rather than executed.

Every seam has a negative control that fails when the defect is reintroduced: restoring timestamp-range trimming or range expiry, ranking on the newest page instead of the oldest, dropping the size measurement, diverging the sweep's default TTL from the read path's, disabling the delete-based write path, deleting by key alone without the ranked timestamp, applying a user retention rule row-level, serving `stale_if_error` only from the error arm, claiming an in-flight key for a response that is never enqueued, letting the periodic refresh overwrite an entry with an error body, disabling the flush's per-key coalescing, a claim that does not release when dropped, a periodic refresh that ignores claims, and a range predicate left open at its lower end.

## Upgrading

No storage change and no rebuild: expiry and size are derived, not stored.

Two behaviour changes worth calling out. A dataset with `retention_period` or `retention_sql` set alongside `refresh_mode: caching` previously had them silently overridden; they now apply, at entry granularity. And with `caching_stale_if_error: enabled`, an origin answering 429 or 5xx now serves the cached response rather than the error body — which is what the setting asks for, but it is a change in what such a query returns.


bjchambers:
Copilot review dispositions — all seven looked at, four were real:

**Fixed**

- *Periodic stale refresh writes 429/5xx bodies* — real, and the worst of them. `refresh_all_stale_rows` replaces the entry it refreshes and only checked for emptiness, so a failing origin replaced the last good response with its error body and served that as a cache hit until it expired. It now classifies the response the same way the single-entry path does. Regression test + negative control.
- *Coalesce drained requests by cache key* — added. Two requests for one key would share one OR'd delete and then each append, leaving the key holding the response twice. The in-flight claim makes this unreachable from both paths that enqueue, but the flush no longer depends on its callers for it.
- *The "nothing bounds this cache" warning is false when `retention_period`/`retention_sql` is set* — real; those evict entries too. The warning now takes whether a user retention rule is active, which meant moving the call after the filters are resolved.
- *`caching_max_size`/`caching_max_items` docs claim `None` leaves the cache bounded by TTLs* — real; with `caching_stale_if_error: enabled` nothing bounds it. Reworded. Also the missing parameter name in a doc comment.

**Already addressed differently**

- *`MAX_ENTRIES_PER_SWEEP` is not a global bound* — correct against that revision. The cap now applies once, to the final eviction set, at predicate construction rather than during selection; capping selection left over-budget entries among the survivors where they overlapped in fetch time, which is what stopped a large cache converging at all.
- *Child propagation branches on `is_expired` and can duplicate* — closed at the source: propagation now happens only for the reader that claims the key, so a second concurrent miss neither enqueues nor propagates.

Reviewing the write path for this also turned up a defect Copilot did not flag and no test would have: the in-flight claim was taken before the decision to enqueue, so a single 429 or 5xx left the key claimed for the life of the process and refused every later write *and* revalidation for it — one transient failure made a key permanently uncacheable.

bjchambers:
Adversarial review dispositions — three of the five were real races or leaks, now fixed with negative controls; two are trade-offs, now stated rather than silent.

**Fixed**

- *Delete-then-append is not idempotent — a reader in the gap can enqueue an append-only write and duplicate the entry.* Real, and the in-flight claim did not close it: the claim was taken **after** the fetch, so a reader whose observation was made stale by a concurrent replacement claimed a free key and appended beside a response it never saw. The claim is now taken before the origin is asked and spans the whole window in which a writer's view can go stale. The review also correctly noted that `refresh_all_stale_rows` claimed nothing at all — it writes straight to the accelerator and opens the same gap, so the race was reachable through it regardless. It now claims each entry it replaces.
- *Cancellation while awaiting channel capacity permanently leaks a claim.* Real. Release was tied to the flush or to explicit early-return paths, never to the future being dropped, so a cancelled query or a disconnected client left a key claimed for the process lifetime — refusing every later write and revalidation for it. `CacheKeyClaim` now releases on `Drop`, and only a successful enqueue hands it to the flush. The claim set moved to a synchronous lock, which its critical sections always wanted and which `Drop` requires.
- *The bulk eviction predicate is unversioned and can cut a newly written entry in half.* Real in kind. The window is now closed at both ends rather than open below the cutoff, so the delete can only reach rows that were present when the ranking made its decision. Ranking without the write lock is kept: holding it across a whole-table aggregate would stall every cache write for the duration of that scan.

**Deliberate, and now documented rather than silent**

- *Delete-then-append can lose an entry if the append fails.* Accurate, and it is the order that fails safely. Losing an entry costs a re-fetch — a cache miss is always a correct answer — plus that entry's `stale_if_error` fallback. Appending first would leave the key holding both responses if the delete failed, and a duplicated source row is a *wrong* answer queries keep returning. A cache may lose an entry; it may not invent rows. The failure is logged, named, and recorded against dataset health.
- *Every sweep groups, sorts and materialises the whole entry set.* Real cost, and the price of deriving rather than storing. The body scan is already skipped when no byte budget is set. The recommendation — maintaining per-entry metadata beside the table — is the design this deliberately avoids: a tally has to be rebuilt on restart, kept in step with every write path, and disagrees with the table the moment anything else removes a row. The available win without that trade is deferring each entry's key strings until a delete actually names it (at most a few hundred, and none on the range path); that is left as the identified next step because it rewrites how delete predicates are addressed, which is the highest-consequence code in this change. A measured run swept a 2,500,000-entry DuckDB acceleration every five minutes for 84 minutes without failing.
- *The byte budget silently excludes response headers.* The exclusion is real and the "silently" was fair. It is now stated on `caching_max_size`, where an operator setting a budget will look. Not raised as a load-time warning: that would fire for every HTTP caching dataset with a budget, about a column the operator did not choose to store and cannot stop the connector storing.

crates/runtime-table/src/accelerated/caching_eviction.rs
```diff
@@ -0,0 +1,2129 @@
...
+/// `retention_sql`, if it set any. Its filters are carried through and applied
+/// per entry rather than per row; see [`CacheEvictionPredicate`].
+pub fn retention(
+    limits: CacheLimits,
+    dataset_retention: Option<Retention>,
+    accelerator: &Arc<dyn TableProvider>,
+    dataset_name: &TableReference,
+    io_runtime: &Handle,
+) -> Retention {
+    let predicate = Arc::new(CacheEvictionPredicate::new(
```
Jeadie, line 137: nit: if `!predicate.limits.is_enforced()`, then we could just return `dataset_retention: Option<Retention>` (since when we dont `is_enforced` in CacheEvictionPredicate, we skip through).

Jeadie (dismissed):

Jeadie (dismissed):

Jeadie (approved):

PR #13618, opened 2026-08-27
feat(caching): serve results stale after an acceleration refresh instead of evicting them

### LINKED ISSUE #13584: Serve stale-while-revalidate after an acceleration refresh instead of evicting every dependent cached result
# Problem

When a dataset's acceleration refreshes, every dependent SQL results-cache entry is hard-evicted — and "changed" matches any successful refresh (`refresh_result_changed_accelerator`, `crates/runtime-table/src/accelerated/refresh.rs:1256` as of `6f51b9303b`), so a `refresh_mode: full` rewrite that loaded identical data still flushes everything. For workloads that poll-refresh accelerated datasets on short intervals, each refresh turns the whole per-table cache into synchronous misses at once: a periodic latency spike on a fixed cadence, and for expensive queries a recurring compute cost even when nothing changed.

`stale_while_revalidate_ttl` does not cushion this — its window anchors at TTL expiry only. An invalidated entry is never served stale, so the refresh boundary is always a cliff, at exactly the moment the runtime holds fresh local data it could revalidate against.

# Proposed Solution

An opt-in invalidation mode that marks dependent entries **stale-as-of-the-refresh** instead of evicting them:

1. Per-table invalidation records the mark (the read-time clock from #12703 already records this instant) and leaves entries resident.
2. A hit within `stale_while_revalidate_ttl` **of the mark** serves the previous result with `CacheStaleWhileRevalidate` status and triggers the existing background revalidation — the arriving query, re-executed in its own request context against the already-refreshed accelerator. No stored SQL or replayed credentials needed.
3. The revalidated result replaces the entry. A hit beyond `mark + stale_while_revalidate_ttl` is a miss, as today. The clock still prevents a result whose read began before the mark from being stored as fresh.

```yaml
runtime:
  caching:
    sql_results:
      on_table_invalidation: evict   # default — current behavior
                           # serve_stale_while_revalidate
```

**Invalidation side** — in the new mode, `QueryResultsCacheProvider::invalidate_for_table` (`crates/cache/src/lib.rs:730`) stamps the invalidation clock and **skips the backend eviction** (the moka predicate / Pingora scan). The clock becomes the sole record of the invalidation; entries stay resident.

**Hit side** — the validity check that today hard-rejects (`tables_invalidated_since`, `lib.rs:769`) becomes a three-way decision. An entry's mark is the **latest** invalidation instant among its `input_tables`:

```
mark < entry.read_started_at             → not stale: entry was computed after the
                                           refresh; fall through to normal TTL/SWR logic
mark ≥ read_started_at, now ≤ mark+swr   → serve stale + trigger async revalidation
mark ≥ read_started_at, now > mark+swr   → miss (as today)
```

krinart:
Fixes #13584

## What was wrong

An acceleration refresh hard-evicted every dependent SQL results-cache entry, and "changed" matched *any* successful refresh — so a `refresh_mode: full` rewrite that loaded byte-identical data still flushed the whole per-table cache. For a workload polling accelerated datasets on a short interval, each refresh turned a whole population of cached results into simultaneous synchronous misses.

`stale_while_revalidate_ttl` did not help, because its window anchored at TTL expiry only. An invalidated entry was never served stale, so the refresh boundary was always a cliff — at exactly the moment the runtime held fresh local data it could revalidate against.

## What changes

When `stale_while_revalidate_ttl` is configured, an invalidation now marks dependent entries **stale as of the refresh** instead of evicting them. With no window configured, invalidation stays hard — there is no staleness anyone has agreed to serve. **No new configuration parameter**: the existing setting is the opt-in.

A cached result is now judged against **two independent clocks**. Both are evaluated on every hit, and the first one to say "not servable" wins:

| clock | measured from | fresh | serve stale + revalidate | miss |
| --- | --- | --- | --- | --- |
| **age** (pre-existing) | when the entry was stored | ≤ `item_ttl` | `item_ttl` → `item_ttl + swr` | past `item_ttl + swr` |
| **invalidation** (new) | the refresh that touched its tables | refresh predates the entry's own read | `refresh` → `refresh + swr` | past `refresh + swr` |

So after a refresh:

- **Inside the window** — the previous result is served with `Results-Cache-Status: STALE`, and a background revalidation is started. It is single-flighted per key, so N concurrent requests produce one query and all N are served stale. Its result replaces the entry, whose read time is then *after* the refresh, so the entry returns to a plain `HIT`.
- **Past the window** — a miss, exactly as before this change.
- **The window is anchored to the refresh instant, but capped by the entry's own age.** An entry is still dropped `item_ttl + swr` after it was *stored*, regardless of any refresh, so a refresh landing late in an entry's life may leave less than the full window. The effective window is the shorter of the two. This can only end stale-serving early, never extend it past the refresh — so it is safe, it just absorbs fewer misses than `swr` alone would suggest.
- **A failed revalidation invalidates nothing.** The previous result keeps being served until the window closes, then becomes a miss. This is why `swr_revalidations{outcome}` exists — that failure has no other visible symptom.

The read-time invalidation clock keeps its guarantee throughout: nothing is served *as fresh* once a table it read has moved on.

## Measured

Accelerated dataset (10M rows, `refresh_mode: full`, 15s interval), 50 distinct cache keys, one binary, two spicepods differing by that one line. `item_ttl` exceeds the run length so a refresh is the only thing that can invalidate.

| | 8 rps | 20 rps |
|---|---|---|
| p50 | 15.2 → 11.0 ms | 13.7 → 11.1 ms |
| **p99** | **264 → 19 ms (13.7×)** | **8114 → 34 ms (239×)** |
| max | 535 → 54 ms | 11155 → 72 ms |
| post-refresh p50 | **161 → 11 ms (14.6×)** | **2689 → 13 ms (214×)** |

The median barely moves — most requests always hit a warm cache. The effect is entirely in the tail. Plotted against time-since-refresh, the old behaviour is a ~161 ms plateau lasting 6 of every 15 seconds; the new one is flat.

Mechanism, from the runtime's own counters at 8 rps: **1,181 misses became 1,182 stale-serves**, one for one, with `evictions{reason="invalidated"}` at zero.

**The compute does not disappear** — it moves off the request path. The same queries still run as background revalidations, so CPU is unchanged and a served result can be up to one refresh interval stale. That is the tradeoff the setting opts into.

## Observability

`evictions{reason="invalidated"}` necessarily falls silent when nothing is evicted, so three counters were added to keep the behaviour visible:

- `results_cache_table_invalidations{mode="evict"|"mark_stale"}` — what an invalidation did
- `results_cache_invalidation_stale_hits` — misses the window absorbed
- `results_cache_swr_revalidations{outcome=...}` — a revalidation that never lands leaves entries serving stale with no other symptom

`results_cache_stale_rejections` gains a `reason` attribute separating "no window configured" from "the window had closed" (the signal that the window is too short).

## Tests

Cache crate: window boundaries, latest-mark-wins across a query's tables, entry residency, zero-length window, and the divergence between the strict and stale-aware lookups. Runtime: end-to-end refresh → `STALE` → revalidation-restores-`HIT`, plus a control proving invalidation stays hard with no window configured.


crates/cache/src/lib.rs
```diff
@@ -554,6 +563,35 @@ impl TableInvalidationClock {
         state.invalidated_at.insert(key, at);
     }
 
+    /// Returns the newest instant at which any of `tables` was invalidated, or
+    /// `None` if none of them has been.
+    ///
+    /// The newest is what a result reading all of them is measured against: any
+    /// one of its tables moving on is enough to leave the result behind, so the
+    /// most recent such move is the one that matters.
+    fn latest_invalidation<S: std::hash::BuildHasher>(
```
bjchambers: Is this *latest invalidation* or *latest refresh*. Specifically, this PR changes it so that after a refresh the cache is not *invalid* but rather *stale*. That suggests the property is `latest_refresh`. Then if `latest_refresh >= start + staleness` we invalidate, and if `latest_refresh >= query` we treat it as stale. If that is correct and makes sense, how does that compare to the logic implemented? One specific case of interest is what happens when a table is updated -- do we immediately treat all queries as stale and start re-running them, or is there a grace period derived from the allowed staleness?

peasee (approved):

PR #13619, opened 2026-08-27
feat(cayenne): observability for maintenance decisions and footprint growth

bjchambers:
## Why

Cayenne's maintenance passes mostly decide **not** to run — a compaction declines because a lock is held or a budget is exceeded, the deletion-vector sweep declines because it cannot prove the current snapshot is empty. Every one is a correct decision, and every one leaves the table slightly larger. Each decline was a `tracing::debug!` and nothing countable, so "the table is growing and nothing reclaims it" could not be told apart from "reclamation ran and found nothing to reclaim" without re-running the workload with debug logging on. The footprint was equally invisible: file counts, deletion-vector bytes, manifest rows, and the metastore's own size were observable only by opening the SQLite file by hand.

This adds 40 metrics on the existing `cayenne` meter, reaching `spiced --metrics <addr>` at `/metrics`.

## What you can now answer

| Question | Metrics |
|---|---|
| Is each maintenance pass running, and if not **why** | `cayenne_compaction_outcome_total{kind,outcome}`, `cayenne_maintenance_outcome_total{op,outcome}` |
| Which threshold asked for a pass | `cayenne_compaction_trigger_total{kind,trigger}` |
| How much did each pass actually give back | `cayenne_maintenance_reclaimed_{files,bytes,rows}_total{op}` |
| Where is the disk going | `cayenne_storage_{files,bytes,rows}{tier}`, `cayenne_data_dir_{files,bytes}{kind}`, `cayenne_data_dir_snapshot_dirs` |
| How big is the metastore, and which table is growing it | `cayenne_metastore_{db,freelist,table}_bytes`, `cayenne_metastore_table_rows`, `cayenne_snapshot_manifest_rows{reachable}` |
| Where is the RAM going | `cayenne_memory_account_{bytes,reserved_bytes}`, `cayenne_mem_tier_bytes`, `cayenne_inline_cache_bytes`, `cayenne_deletion_index_bytes`, `cayenne_scan_file_statistics_entries`, `cayenne_pk_keyset_budget_{used,total}_bytes` |
| How much of RSS is ours vs page cache | `process_resident_anon_bytes`, `process_resident_file_bytes` |
| Is the PK index healthy | `cayenne_pk_index_{format,keys,bytes,budget_bytes}`, `cayenne_pk_index_{discard,preserved}_total`, `cayenne_pk_bloom_*` |
| Why is the bake not firing | `cayenne_deletion_index_len` against `cayenne_autotune_bake_deletion_index_trigger` |

Full reference, including the label vocabulary and the traps below: **`docs/cayenne/cayenne.md`, Appendix C**.

## The decision metrics

`cayenne_compaction_outcome_total{table, kind, outcome}` records one row per attempt, with `kind` reusing the vocabulary already on `cayenne_compaction_duration_ms` (`full`, `subset_current`, `subset`, `datalake`, plus `bake`) so an outcome joins to that pass's duration and bytes. Every exit records exactly one outcome, in three classes:

- **work happened** — `committed`, `applied` (did its work, but that work returns no space by itself — retention writes tombstones), or `no_op` (ran, found nothing)
- **work was paid and discarded** — `aborted_concurrent_change`
- **the pass never ran** — a `declined_<reason>`

`outcome=~"declined_.*"` is therefore one selector for "what is stopping maintenance". `cayenne_maintenance_outcome_total{op}` applies the same grammar to the deletion-vector sweep, retention, and the retired-directory sweep, and those counters are registered at zero so "nothing reclaimed" is an observation rather than an empty query.

Two decline reasons are deliberately not merged, because they have opposite prognoses: `declined_not_due` (nothing reached its grace window — resolves itself) and `declined_live_reference` (candidates were examined and their files are still referenced by a live snapshot — may never resolve).

The **bake** previously had no instrumentation at all — not an outcome, not a duration, not the bytes it writes. It now joins the compaction family on all three.

## Four counting traps, documented and handled

These are the places where a natural reading of the data is wrong, and each is called out in the appendix:

1. **Logical figures count pathnames; only reclaim is physical.** A manifest `file_path` is a bare filename resolved against *its own row's* snapshot directory, so the same filename under two snapshots is two files as far as every reader is concerned. `cayenne_storage_*` therefore sums per live manifest row, matching `cayenne_data_dir_*`, which walks the filesystem the same way; both count a hard-linked file once per name. Separately, `cayenne_snapshot_manifest_rows{reachable="false"}` is the prune backlog — dead snapshots' rows, which is what makes the raw `cayenne_snapshot_file` row count overstate live state.
2. **Reclaimed means physically freed.** Files/bytes count only a successful unlink — not an object-store key this sweep does not delete, not a file some earlier actor removed, and **not a file whose inode another pathname still holds**. Subset compaction hard-links an unpicked file into the new snapshot's directory, so unlinking a retired snapshot's name for it frees nothing; the link count is read before the unlink. Catalog-row removal is counted separately as rows, and retention reports `applied` rather than `reclaimed`.
3. **A component and its total must come from one read.** `cayenne_memory_account_bytes{kind}` exists to be compared against `cayenne_memory_account_reserved_bytes` — equal means the accounting reaches the `DataFusion` pool, and a large gap means it does not. Read from separate atomics, a sampler could pair one write's components with another's total and manufacture that gap, so the components live under the same lock as the reservation they sum to (which every mutation already took to resize).
4. **A bloom cannot count its keys.** `cayenne_pk_bloom_insertions` tallies insertion calls, so it is an *upper* bound on distinct live keys and `bits_per_insertion` is a *lower* bound on true density. `cayenne_pk_index_keys` is meaningful only at `cayenne_pk_index_format = 1`, and reads zero in bloom mode rather than retaining its last exact count.

A related rule runs throughout: a gauge that is skipped keeps its previous value, so any representation-specific gauge is **zeroed** when its representation is not live rather than left unpublished.

## Cost

Nothing lands on a hot path. Per-event counters fire per **batch**, never per row; the bloom split accumulates across an apply; the encode fan-out is stored in an atomic at the decision and published on the tick.

Sampling runs on four cadences matched to cost, each behind an in-flight guard so a slow sample cannot overlap itself, with the cadence advancing only on a completed sample:

| Sample | Cadence | Cost |
|---|---|---|
| memory, PK index, deletion index, write shape | every tick | atomic loads + one `try_lock` |
| storage / manifest / metastore rows | ≥ 30 s per table | two aggregate queries |
| `cayenne_data_dir_*` | ≥ 5 min per table | one `stat` per file, in a single `spawn_blocking` |

The metastore gauges describe the shared file, so their clocks are keyed by metastore path — N tables produce one sample's work, not N. Sampling is driven from both the compaction tick and the post-write maintenance loop, because the bulk-overwrite profile disables the compactor and would otherwise leave the tables that rewrite themselves wholesale unmeasured.

## Resident memory

`process_resident_memory_bytes` keeps its name and value but splits into `process_resident_anon_bytes` and `process_resident_file_bytes`. On a Cayenne file-mode workload roughly half of total RSS is file-backed — mapped metastore and Vortex pages the kernel reclaims on demand — so the budget gap should be taken against **anon**, not the total. On Linux this is the same single `/proc/self/status` read; on macOS one `task_info(TASK_VM_INFO)` syscall, replacing a per-sample `sysinfo::System` construction.

Neither is the pod-sizing number: the kernel OOM-kills on cgroup accounting, which also charges kernel memory no per-process metric sees. Capacity planning belongs on `container_memory_working_set_bytes`; this family attributes memory to what allocated it.

## Behaviour change

`cayenne_metastore_wal_bytes` gains a `catalog` label. Without one, every dataset's metastore overwrote the others on a single series.

## Known gap

`ScanViewCache` can pin a previous mem-tier generation's Arrow buffers while `cayenne_mem_tier_bytes` reports the new, smaller tier — so residency and the gauge move in opposite directions across a checkpoint. Bounded (1 completed + ≤8 in-flight views per table, evicted after ≥5s idle) but unbounded in time on a continuously queried table. Not instrumented here because the buffers are `Arc`-shared with the live tier and with executing queries, so a naive sum would double-count — the over-reporting class this PR spent two review rounds removing. Filed as #13661.

## Testing

Unit tests cover the reachability split and the path-versus-inode rule (a filename repeated under two live snapshots is the fixture; narrowing the live set to the current snapshot is the control), that unlinking a hard-linked file bills zero bytes and the surviving link bills full length once its alias is gone, that a snapshot's memory components always sum to its reservation (a setter that skips the resize is the control), the data-directory walk classifier, partial reclamation and mixed success/failure at the real deletion seam (a forced read-only-directory failure), and the label vocabulary — including that `declined_*` coincides exactly with the declines in **both** outcome families.

`crates/runtime/tests/metrics.rs` asserts a real Cayenne dataset exports every family to `/metrics`, and asserts the exact `(table, kind, outcome)` triple its fixture must produce.

Verified against a live `spiced --metrics` run: the manifest-derived footprint, the independent directory walk, and `stat` all report the same bytes for the same table.


bjchambers:
Used this instrumentation to run down a bake bug — filed as #13620.

`cayenne_compaction_outcome_total{kind="bake"}` named the cause on the first scrape: `declined_not_key_mode`, ~73 per table, 0 `committed`. Root cause is that `deletion_mode: auto` resolves to `Position` while the write path independently picks a key-based strategy from primary-key presence, so the bake gate never fires on a default PK table.

Worth noting how much the outcome labels bought: reaching the same conclusion without them took four one-hour runs plus a chain of inference from `COUNT(cayenne_delete_file)` matching the on-disk `.arrow` count exactly. The metric answered it in about a minute, and named a cause that was not among the three hypotheses I had been working from.

Two notes from using it in anger:

- `cayenne_deletion_index_len` alongside `cayenne_autotune_bake_deletion_index_trigger` is what made `declined_not_key_mode` interpretable — the ratio was already 1.52 on the heaviest table, which ruled out "trigger simply not met" immediately. Keeping those two adjacent is valuable.
- `cayenne_maintenance_outcome_total{op="orphan_dv_sweep"}` emits no samples at all when the sweep is never scheduled. That is correct behaviour for a counter, but "no series" and "not instrumented" look identical to a consumer. A zero-initialised series, or a `scheduled`/`not_scheduled` outcome pair, would distinguish them.

bjchambers:
Follow-up after using this in anger for a few hours against a refresh-only upsert workload (2M and 10M rows, file mode, SQLite metastore). Two of my three open questions became answerable from metrics alone; the third did not, and I think it points somewhere outside this PR.

## Confirmed working

`cayenne_metastore_table_bytes{metastore_table=…}` would have caught #13621 on its own:

```
cayenne_inlined_delete   16,896,000
cayenne_snapshot_file       868,352
cayenne_delete_file         634,880
cayenne_inlined_data         12,288    <- 1,375x smaller than its own tombstone table
```

That ratio *is* the bug. Finding it without this took a hand-written SQLite walker plus four one-hour runs.

`cayenne_memory_account_bytes{kind=…}` splitting `deletion_index` / `keyset` / `cold_existence` is the right shape, and it settled a question I had been guessing at — see below.

## The RSS gap is still ~86%, and the keyset is not the cause

I had assumed the unaccounted memory was PK keysets, since auto-tuning reports `pk_keyset_cache=2048MB` per table across 9 tables. This PR disproves that:

```
cayenne_pk_index_bytes{site="table_keyset"}     0     (budget 2 GiB)
cayenne_pk_index_bytes{site="sharded_keyset"}   0     (budget 2 GiB)
cayenne_memory_account_bytes{kind="keyset"}     0
```

Zero on every table, both sites. The keysets are simply not populated on an append-refresh upsert workload, so the budget is reserved-but-unused and my hypothesis was wrong. That is a useful negative result and I would not have got it without these gauges.

Full accounting on a live 2M-row run:

```
process RSS                  7.820 GiB
  memory_account (all kinds) 0.352 GiB   (all of it deletion_index)
  segment cache              0.390 GiB
  query pool used            0.352 GiB
  inline cache               0.000 GiB
  mem_tier fleet used        0.000 GiB
  ACCOUNTED                  1.094 GiB   (14.0%)
  UNACCOUNTED                6.726 GiB   (86.0%)
```

Every Cayenne structure this PR instruments now reports a plausible value, and together they account for 14% of RSS. So the remaining 6.7 GiB is probably **not** a resident Cayenne structure at all — the candidates I would look at next are transient allocation retained by the allocator (each refresh cycle re-materialises large Arrow batches; the join view alone re-materialises ~1 GiB every 60s), Vortex decode/scan buffers, and DataFusion execution state.

If that is right, no per-structure Cayenne gauge will ever close this gap, and the useful addition is at the runtime level rather than here: allocator statistics (jemalloc `stats.allocated` vs `stats.resident`, or `mallinfo`) exported alongside `process_resident_memory_bytes`. That would split "we are holding live state we have not measured" from "the allocator has not returned freed pages", which are very different bugs and currently indistinguishable. Happy to file that separately if you agree it is out of scope here.

## Smaller notes

**Un-scheduled counters emit no series.** `cayenne_maintenance_outcome_total{op="orphan_dv_sweep"}` was absent in every configuration I ran. That is correct for a counter, but "never invoked" and "not instrumented" look identical to a consumer — I could only interpret it because I had already read the call site. Zero-initialising, or a `scheduled` / `not_scheduled` outcome pair, would disambiguate.

**`catalog` label cardinality.** `cayenne_metastore_table_bytes` carries the full filesystem path:

```
catalog="/…/runs/B-metrics/cayenne/view_sender_info/metadata/cayenne.db"
```

That is deployment-specific, varies per pod, and puts local paths into exported series. A short identifier or basename would be safer.

**A pattern worth naming in the PR description.** Both bugs this instrumentation found (#13620, #13621) have the same shape: a reclamation path gated on a condition that never holds, so garbage accumulates while maintenance reports healthy. The generic signal is `*_reclaimed_*_total` flat at zero while the corresponding `*_bytes` climbs — neither alone is alarming, the pair is. Calling that out would tell operators what to alert on.

bjchambers:
Correcting my previous comment — I was wrong, and the metrics in this PR are what showed it.

I claimed the PK keysets "are simply not populated" because `cayenne_pk_index_bytes` and `cayenne_memory_account_bytes{kind="keyset"}` both read 0. That was the wrong conclusion. The keyset **is** populated; it took the bloom fallback, and the bloom's memory is not counted by either gauge:

```
table                    site           bloom bits        = bytes     keys        bits/key   pk_index_bytes
number_caps_by_arrival   table_keyset   17,179,869,184     2.00 GiB   12,804,096   1341.7          0
```

`cayenne_pk_bloom_bits` is the only reason this was findable, so the instrumentation did its job — but the byte accounting misses the path.

Two things fall out, and I think the second matters more than the first.

## 1. Bloom memory is missing from the byte gauges

2.00 GiB resident, reported as 0 by `pk_index_bytes` and by `memory_account_bytes{kind="keyset"}`. That is ~30% of the ~6.7 GiB RSS gap I described earlier, from one table alone. Counting `bloom_bits / 8` into the keyset accounting would close a large part of it and make `fleet_budget_used_bytes{budget="pk_keyset"}` (currently 0 against a 4 GiB limit) meaningful.

## 2. The bloom appears to be sized from the budget, not the key count

```
cayenne_pk_bloom_bits          17,179,869,184 bits = 2,147,483,648 bytes
cayenne_pk_index_budget_bytes                        2,147,483,648 bytes
cayenne_pk_bloom_inserted_keys 12,804,096
cayenne_pk_bloom_bits_per_key  1,341.7
```

The bloom is exactly the full budget, and 1,341.7 bits/key is roughly 130x a conventional ~10 bits/key for a 1% false-positive rate. Those 12.8M keys would fit in ~16 MiB. If the intent is "spend the budget you were given", that is defensible on a large host but expensive: the budget is memory-derived (~1/32 of RAM, so 2 GiB on a 64 GiB host) and applies **per table**, so a pod with several bloom-fallback tables reserves multiples of it for a structure that needs a fraction.

This is arguably out of scope for an observability PR — I raise it here only because these metrics are what made it visible, and because `bits_per_key` is exactly the right gauge to alert on. Happy to file it separately if you would rather keep this PR to instrumentation.

## Amended RSS accounting

With bloom bytes folded in, the picture from my earlier comment changes materially:

```
process RSS                  7.820 GiB
  bloom (bits/8, 1 table)    2.000 GiB   <- currently reported as 0
  memory_account             0.352 GiB
  segment cache              0.390 GiB
  query pool used            0.352 GiB
  ACCOUNTED                  3.094 GiB   (39.6%, was 14.0%)
  UNACCOUNTED                4.726 GiB   (60.4%)
```

So my earlier suggestion that the gap is "probably not a resident Cayenne structure" was premature — at least a third of it is one, and it was hiding behind a zero. I would hold off on the allocator-statistics idea until the bloom accounting is fixed and we can see what is genuinely left.

bjchambers:
Third comment, and it partly retracts my second. Worth reading as one finding: **two gauges describing the same structure contradict each other, and I believed the wrong one.**

## The contradiction

For the same `table` and `site`, in one scrape:

```
cayenne_pk_index_format{site="table_keyset",table="number_caps…"}        0
cayenne_pk_index_keys{site="table_keyset",table="number_caps…"}          0
cayenne_pk_index_bytes{site="table_keyset",table="number_caps…"}         0

cayenne_pk_bloom_bits{site="table_keyset",table="number_caps…"}          17,179,869,184
cayenne_pk_bloom_inserted_keys{site="table_keyset",table="number_caps…"} 12,804,096
cayenne_pk_bloom_bits_per_key{site="table_keyset",table="number_caps…"}  1,341.7
```

`pk_index_format = 0` documents itself as "absent (every validated batch rebuilds it)". So the index is absent — while the bloom gauges describe a live 2 GiB bloom holding 12.8M keys.

## Which one is right

The bloom gauges appear to be **stale**. Sampling twice, 100 seconds apart, mid-run with refreshes landing continuously:

```
T+0     bloom_bits 17,179,869,184   inserted_keys 12,804,096
T+100s  bloom_bits 17,179,869,184   inserted_keys 12,804,096
```

Frozen. A live bloom under this workload would move — and `inserted_keys` of 12.8M against a 4M-row table already looks like a value accumulated over rebuild cycles rather than a current census.

`track_pk_index_shape` records `format` and `keys` together and both read 0 consistently, so that path is being exercised. The bloom path looks like it is only recorded when a bloom is constructed, and never reset when the index is later discarded (`discard_reason()` exists in `pk_index.rs`, so discard is a real path). An OTel gauge recorded once and never re-recorded keeps its last value indefinitely.

## Correcting my previous comment

I claimed `cayenne_pk_index_bytes` was missing 2.00 GiB of resident bloom, and revised my RSS accounting from 14% to 39.6% accounted on that basis. **Withdraw that.** If the bloom has been discarded, `pk_index_bytes = 0` is correct and the memory is not resident. My earlier accounting (~14% accounted, ~86% unaccounted) stands, and the question of where that memory actually lives is still open.

I was wrong because I treated a non-zero gauge as evidence of a live structure without checking it was still being updated. That is the same mistake in the opposite direction from my first comment, where I treated a zero gauge as evidence of absence.

## Why this matters more than either of my wrong conclusions

A stale gauge is worse than a missing one. A missing metric makes you go and look; a frozen metric makes you confident about something that stopped being true, and it survives into dashboards and alerts. Two readers of this PR reached opposite wrong conclusions from the same scrape within an hour.

Suggestions, in rough priority:

1. **Reset the bloom gauges when the index is discarded or rebuilt** — record `bits`, `inserted_keys` and `bits_per_key` as 0 on the discard path, so the series tracks reality rather than a high-water mark.
2. **Make `pk_index_format` authoritative** — since it already distinguishes absent / exact / bloom, a consumer should be able to read the bloom gauges only when `format == 2`. Documenting that dependency in the HELP text of the bloom metrics would prevent the misreading directly.
3. If the intent is to preserve a high-water mark deliberately, name it so — `..._peak_bits` reads very differently from `..._bits`.

## Still worth checking independently

If a bloom really was constructed and sized at exactly `pk_index_budget_bytes` (2 GiB, at 1,341.7 bits/key for 12.8M keys — roughly 130x a conventional ~10 bits/key), that sizing is worth a look on its own even though it is not currently resident. Those numbers were recorded at some point, so the allocation presumably happened. I have no way to tell from metrics alone how long it lived.

sgrebnov (approved):

PR #13634, opened 2026-08-27
feat: Add support for Qdrant as a vector engine

anush008:
Hello 👋

## Description

This PR adds Qdrant as a supported vector engine to write into/read from in Spice.

[Qdrant](https://qdrant.tech/) is an open-source 🦀 vector search engine built for high-performance and massive scale.

### Testing

I've tested this integration end-to-end against a local Qdrant instance.

### Setup

You can run Qdrant with 

```bash
docker run -p 6333:6333 -p 6334:6334 qdrant/qdrant
```

Then, set `engine` and `qdrant_endpoint` to `qdrant` and `http://localhost:6334` respectively.

The dashboard is accessible at http://localhost:6333/dashboard.

The Qdrant Rust client uses the gRPC API available at port 6334 ([Ref](https://qdrant.tech/documentation/interfaces/)).


lukekim:
@Anush008 thanks for the PR. Can you please have your partnerships team reach out to us on Slack. We'd only be willing to support this with support from the Qdrant team.

kanungle:
@lukekim what level of support do you require? Thanks

PR #13659, opened 2026-08-28
fix(cayenne): prune old-snapshot catalog rows at subset small-file compaction

Jeadie:
## Summary

Cayenne runs compaction on the current snapshot. It will do this in one of two ways:
1. A full current snapshot compaction.
2. A small subset compaction. 

In the latter case, we do not appropriately delete `cayenne_snapshot_file` or `cayenne_snapshot_file_statistics`.

This affects cayenne tables from accelerations of the form
```yaml
acceleration:
  enabled: true
  engine: cayenne
  refresh_mode: append
  primary_key: my_pk
  time_column: my_time_col
  on_conflict:
    my_pk: drop # `upsert` publishes a protected snapshot on every refresh batch, so never pick small compaction.
  params:
    cayenne_deletion_mode: key   # (E) MUST be `key` — default `auto` resolves to Position,
    # Fields below make small files path more likely
    cayenne_target_file_size_mb: 8          
    cayenne_compaction_trigger_files: 3 
    cayenne_compaction_max_files_per_pick: 2
```

Or a append/insert-only CDC
```yaml
acceleration:
    engine: cayenne
    mode: file
    refresh_mode: changes
    primary_key: id
    # NO on_conflict: upsert
```

## The fix
The subset path now mirrors the full current snapshot compaction path:

1. Before the commit, author the new snapshot's manifest from a directory
   listing (publish-before-clear).
4. After the commit succeeds and the in-memory pointer is updated:
   - Prune the manifest to the new snapshot (`prune_snapshot_manifest_to`).
   - Clear the statistics cache for every snapshot except the new one
     (`clear_snapshot_file_statistics_except`).


phillipleblanc (dismissed):

Jeadie:
Fix 
|                         | subset commits | protected fold | cayenne snapshot file           | snapshot file statistics | metastore db                       |
  |-------------------------|---------------:|---------------:|---------------------------------|--------------------------|------------------------------------|
  | Baseline #13658 |            210 |              0 | 23,822 rows, 23,822 dead (100%) | 23,647 dead              | 24.7 → 57.3 MB (+130 MB/h, linear) |
  | Fix #13659        |            241 |              0 | 839 rows, 0 dead                | 0 dead                   | 0.9 → 2.1 MB, flat                 |

claudespice (approved):
✅ Re-approving to restore a dismissed maintainer approval.

A core maintainer (@phillipleblanc) approved this PR at `e82d30ab`; that approval was dismissed by
the later `Lint` push. The only change since then is a single clippy `collapsible_if` rewrite in
`crates/cayenne/src/provider/table.rs` (nested `if manifest_authored { if let Err(error) = … }`
folded into an edition-2024 let-chain, +10/-10, no other file touched). `&&` short-circuits exactly
as the nesting did, there is no `else` arm, and the `.await` is still reached only when
`manifest_authored` — behaviour-preserving. A code review of that delta found no correctness issue.

Verified: all review threads resolved · no outstanding change requests · nothing failing or pending ·
HEAD carries a green `signoff` status (lint + test passed on this exact commit).

⚠️ Not enqueuing, and this still needs a CI trigger before it can merge. This PR was opened against
`jeadie/cayenne-protected-compaction-manifest-gc`, and `.github/workflows/pr.yml` filters
`pull_request` to `trunk`/`release-*`/`feature-*` — so the `pr` workflow never ran, and the required
`Attestation`, `Rust Lint` and `Build and Test` contexts are **absent** rather than green. #13658
merging auto-retargeted this PR to `trunk`, but a retarget is a `pull_request: edited` event, which
that workflow does not listen for, so the checks will not appear on their own. A push (or a
close/reopen) is needed to fire them.

sgrebnov (approved):

PR #13670, opened 2026-08-28
fix(cayenne): reclaim superseded snapshot dirs against live state

### LINKED ISSUE #13657: Old-snapshot cleanup holds a sliding 2-minute window of snapshot dirs: disk grows with commit rate, not data
### Summary

Every Cayenne snapshot commit defers its old-snapshot sweep by a fixed 2-minute
grace, and the sweep runs against the snapshot id captured when it was *queued*,
not the one current when it *fires*. The two together turn the grace into a
permanent sliding retention window: steady-state on-disk garbage is
`grace × commit_rate × snapshot_size`, unbounded in commit rate.

Measured on TPC-H SF1 `supplier` (10,000 rows, 0.69 MiB live) with
`refresh_mode: full` at `refresh_check_interval: 200ms` — 2.65 commits/s:

| | |
|---|---|
| live table | 722,752 B — 2 `.vortex` files |
| `.spice` on disk | 234,016,329 B — 646 `.vortex` files in 323 snapshot dirs |
| amplification | **324×** |
| oldest surviving snapshot dir | 120.14 s before the last commit |

Every `.vortex` file has `nlink=1`, so all 223 MiB are distinct bytes, not
shared inodes. Footprint climbs linearly for 120 s and then plateaus — the
plateau is the steady state, not a recovery.

All line numbers are at `257da363bc`.

### The defect

`trigger_old_snapshot_cleanup` (`crates/cayenne/src/provider/table.rs#L4630`)
sleeps a fixed grace before sweeping:

```rust
const OLD_SNAPSHOT_CLEANUP_GRACE: std::time::Duration = std::time::Duration::from_mins(2);
```

The local branch (`#L4666-L4700`) moves `current_snapshot` into a detached task,
sleeps, and then re-reads `protected_snapshots` and `snapshot_scan_refs` live —
but never re-reads the current snapshot id:

```rust
let current_snapshot = current_snapshot.to_string();
...
tokio::spawn(async move {
    tokio::time::sleep(OLD_SNAPSHOT_CLEANUP_GRACE).await;
    // Read the LIVE protected set after the grace period. ...
    let mut protected_snapshot_ids: HashSet<String> =
        protected_snapshots.load().keys().cloned().collect();
```

The staleness comment covers the protected set; the anchor itself is 2 minutes
old by the time it is used. `cleanup_old_snapshots_blocking` keeps every dir at
or newer than that anchor (`#L6683-L6695`):

```rust
// Skip directories whose UUID7 timestamp is >= the current snapshot.
// These might be in-flight writes. ...
Some(ts) if ts >= current_unix => {
    tracing::trace!("Keeping snapshot: {snapshot_id} (newer than current)");
    continue;
}
```

So a pass firing at `T` carries an anchor from `T − 120s` and can only delete
dirs older than `T − 120s`. Each commit queues its own pass, so passes fire
continuously and each one is 2 minutes behind — a sliding window, never a drain.

Two consequences follow from the same code:

1. **One detached task per commit.** 794 commits queued 794 tasks; ~320 are
   alive at steady state, and once the first fires, ~2.6 sweeps/s each do a full
   `get_all_snapshot_files` plus a `read_dir` over ~320 dirs — work that scales
   with `rate × window`, i.e. quadratically in commit rate, all racing on the
   same dirs. RSS steps 594 → 624 MiB at t≈130 s in the run above, right after
   the first sweeps begin.
2. **The manifest ref-count gate is computed from the stale anchor.**
   `cleanup_old_snapshots_local` (`#L4725-L4726`) builds the live set from the
   2-minute-old id, so it never contains the *actual* current snapshot:

   ```rust
   live_snapshot_ids.insert(current_snapshot.clone());
   let live_referenced = Self::live_referenced_relative_paths(&all_rows, &live_snapshot_ids);
   ```

   Harmless for whole-table replace (each snapshot's files live in its own dir),
   but the crate documents cross-snapshot in-place manifest references as real
   and relied upon by physical-file GC (`#L2654`, `#L5471`). Whether a live
   reference can be orphaned this way needs its own verification —
   `rewrite_current_snapshot_small_file_subset` hardlinks unpicked files into the
   destination dir rather than referencing them in place, so no concrete loss
   case was constructed here.

### Why it matters

- Disk cost scales with commit *rate*, not with data. At 1 commit/s on TPC-H
  SF1 `lineitem` this is ~12 GB of garbage held permanently; the reporting
  configuration reached 324× live size on a 0.69 MiB table.
- Nothing else reclaims it. On `refresh_mode: full` compaction is structurally
  absent — the runtime says so at startup: `Cayenne compaction memory pool not
  reserved: ... needs a refresh_mode other than full — a whole-table replace
  leaves nothing to consolidate.`
- The inline tier, which exists precisely to keep small full refreshes out of
  the file tier (`crates/cayenne/src/provider/overwrite.rs#L42-L50`), cannot
  reach a table this size: `inline_max_rows` is 1024 and `supplier` is 10,000
  rows.
- `refresh_mode: append` is affected by the same grace. Locally the cloned
  snapshot dirs hardlink, so `du` under-reports and the real cost is compaction
  rewrite bytes held for 2 minutes; on an object store the clone is a genuine
  `store.copy` with no hardlinks, so each retained snapshot duplicates the whole
  table.

### How to validate

```yaml
version: v1
kind: Spicepod
name: test
datasets:
  - from: file:./supplier.parquet   # TPC-H SF1, 10,000 rows
    name: supplier
    acceleration:
      enabled: true
      engine: cayenne
      mode: file
      refresh_check_interval: 200ms
      refresh_mode: full
```

Run `spiced` in that directory and sample `du -sb .spice` every 10 s. Expected
today: linear growth to ~220 MiB over ~120 s, then a plateau. Then:

```bash
# every retained dir is a full copy of the table, none shared
find .spice -name '*.vortex' -printf '%n\n' | sort -u          # => 1
find .spice/data/*/*/ -mindepth 1 -maxdepth 1 -type d | wc -l  # => ~120s x commit rate

# the retained window is exactly OLD_SNAPSHOT_CLEANUP_GRACE
find .spice/data/*/*/ -mindepth 1 -maxdepth 1 -type d -printf '%T+\n' \
  | sort | sed -n '1p;$p'
```

The oldest surviving directory is 120 s before the newest, for any commit rate.

No regression test covers this: every in-crate test reaches the sweep through
`cleanup_old_snapshots_now_for_test` (`#L4760`), which calls
`cleanup_old_snapshots_local` directly and skips the sleep entirely, and nothing
asserts that on-disk footprint converges toward live size under repeated
commits.

### The fix

The exact signal this grace approximates already exists on the provider, and is
already unioned into the sweep's skip set — `snapshot_scan_refs`, documented at
`#L2138-L2146` as

> the exact "is a scan still reading this dir?" signal that replaces the brittle
> time-based grace a long query could outlive

alongside `snapshot_last_listed`. `sweep_retired_snapshot_dirs` (`#L5056`)
already works this way.

1. Re-read the current snapshot id after the wait, the way the protected set
   already is, so a pass sweeps against live state instead of a 2-minute-old
   anchor.
2. Cut the blanket 2-minute sleep to what it actually has to cover — the
   listing→first-file-open window of a single plan — and lean on
   `snapshot_scan_refs` / `snapshot_last_listed` for long-running queries.
   `RETIRED_SNAPSHOT_DIR_GRACE` (`#L4815`) carries the same 2-minute constant
   and the same reasoning, and should be revisited with it.
3. Coalesce the sweep onto the existing `PostWriteMaintenance` debounce
   (`crates/cayenne/src/provider/maintenance.rs`) instead of spawning one
   detached sleeping task per commit, so a burst of commits produces one sweep.

Regression test: drive N whole-table replaces through the real (non-test) sweep
path and assert the on-disk snapshot-dir count and byte footprint converge to a
small multiple of live size, independent of commit rate.

peasee:
### Problem

- Snapshot-dir cleanup slept a fixed 2 minutes, then swept against the snapshot id captured when it was queued, not the live one.
- The sweep keeps every dir at or newer than the id it's given, so it could only ever delete dirs older than 2 minutes. A sliding window, not a drain.
- Garbage scaled with commit rate: `supplier` at 200ms refresh held 323 dead dirs / 223 MiB against a 0.69 MiB live table, forever.
- One sleeping task per commit: ~320 alive at steady state.

### Changes

- Sweep reads the current snapshot pointer itself. A captured id goes stale and pins everything published after it.
- Grace cut to 5s, judged from a dir's own uuid7 age and last scan listing. `SnapshotScanRef` already pins dirs for a scan's full lifetime, so the grace only covers one plan's listing-to-first-open window. `RETIRED_SNAPSHOT_DIR_GRACE` folded in.
- One coalesced sweep per burst instead of one task per commit.
- A pass that defers on grace re-arms once, so an idle table still reaps its last window.
- Subset small-file rewrite pins its source dir: it hardlinks with no fence held, which the 2-minute delay masked.

**Before:**
*Append*
```
TIME      ELAPSED           RSS       .spice     Δ.spice    FILES
0s        4.5 MiB RSS      0.0 MiB          0 files
60s     720.1 MiB RSS    290.4 MiB   +72.5   352 files
120s    985.2 MiB RSS    583.7 MiB   +73.4   708 files
180s   1081.0 MiB RSS    622.4 MiB    +7.9   693 files
```
**After:**
*Append*
```
TIME      ELAPSED           RSS       .spice     Δ.spice    FILES   UPDATES
18:04:14  0s           30.0 MiB      0.0 MiB            -        0         0
18:05:15  60s         752.0 MiB     46.1 MiB     +6.4 MiB       22       290
18:06:15  121s        780.6 MiB     78.8 MiB     +8.2 MiB       24       582
18:07:15  181s        788.1 MiB    110.7 MiB     +8.2 MiB       23       874
```
*Full Refresh*
```
TIME      ELAPSED           RSS       .spice     Δ.spice    FILES
18:07:52  0s            3.8 MiB      0.0 MiB            -        0
18:08:52  60s         609.4 MiB      9.5 MiB     +0.0 MiB       26
18:09:52  120s        649.2 MiB      9.5 MiB     -0.7 MiB       26
18:10:52  180s        648.8 MiB      9.5 MiB     +0.0 MiB       26
```

## 🔗 Related

* Closes #13657


phillipleblanc (dismissed):

crates/cayenne/src/provider/table.rs
```diff
@@ -2174,6 +2225,21 @@ pub struct CayenneTableProvider {
     /// so CDC catch-up bursts do not synchronously pay metastore/listing work
     /// on every append.
     post_write_maintenance: Arc<PostWriteMaintenance>,
+    /// Holds the single superseded-snapshot-dir sweep slot: set before the task
+    /// spawns and released when it finishes, so a burst of commits produces one
+    /// sweep.
+    snapshot_cleanup_scheduled: Arc<AtomicBool>,
```
Jeadie, line 2308: nit: Should reduce to a `snapshot_cleanup` struct. Start to reduce god class of this.

krinart (dismissed):

krinart (approved):

Issue #13717, opened 2026-08-29
A chunked index leaves orphan chunks behind when a row's text gets shorter


claudespice:
### Description

A chunked search index stores one entry per chunk, under a primary key augmented with the chunk
id (`_spice.chunk_id`). A write upserts those chunk-keyed entries: it deletes the keys the batch
carries and appends the batch. So when a row's text gets **shorter** — fewer chunks than the
version already in the index — the write only names the chunk ids it still produces, and every
higher chunk id from the superseded text is left behind.

Take a row whose text is `'aaa bbb'` (chunks `(1,0)='aaa'`, `(1,1)='bbb'`) and is then rewritten
to `'ccc'` (chunk `(1,0)='ccc'`). The second write deletes and re-adds `(1,0)`; `(1,1)='bbb'` is
never named, so it stays. The row remains searchable by `bbb`, a word its current text does not
contain, and the hit carries content the table no longer holds.

Measured on `trunk` with a `ChunkedSearchIndex` over a `MemoryVectorIndex`, writing `'aaa bbb'`
and then `'ccc'` as two separate batches: the stored chunk ids afterwards are `[1, 0]`, not `[0]`.

The same shape reaches a row whose search value becomes NULL or empty: it produces no chunks at
all, so the write names no key for it and every chunk of its previous text survives.

### Impact

Search results disagree with the table they are an index of, with no error. Correctness of
results, not availability. It needs only an update that shortens a document, which is ordinary.

### Where

- `crates/search/src/index/chunking.rs` — `ChunkedSearchIndex::write`, which hands the inner index
  only the chunk rows the current text produces.
- The inner stores' upsert, e.g. `crates/search/src/index/memory/store.rs` — `upsert` deletes the
  keys the batch carries, which cannot include a chunk id the batch no longer has.

### Why it needs its own change

Closing it means the write deleting a row's whole chunk group before adding the new one, which is
what `ChunkedSearchIndex::delete_by_keys` already does — but that path needs either an inner index
that deletes by partial key or a listable `VectorIndex` to enumerate the existing chunk ids, and a
plain full-text inner index has neither (the `delete_by_keys` doc comment records exactly this
constraint). Doing it per write also costs a resolve scan on every write for the common case where
the chunk count did not shrink, so the cheap form probably tracks the previous chunk count rather
than enumerating.

### Provenance

Found while fixing #13713, which repairs the **within-batch** half of this: a batch carrying two
[...]

PR #13718, opened 2026-08-29
fix(search): index only the last change a batch carries for a row (fixes #13713)

### LINKED ISSUE #13713: A batch carrying two changes for one row indexes both, so a search index disagrees with its table
### Description

A single batch reaching `Index::compute_index` can carry the same primary key more than once — a
CDC change envelope holds every change the source produced in one poll, and two updates to one row
in that window arrive as two rows of one batch. Nothing coalesces them to the last one before the
index is computed, so a search index built from that batch does not agree with the accelerated
table built from the same batch, which *does* resolve the key last-write-wins.

On a **chunked** index the disagreement is visible two ways.

Take a batch carrying `(id=1, 'aaa bbb')` and then `(id=1, 'ccc')`, in that order:

- `ChunkedSearchIndex::write` chunks each row independently and concatenates, so the batch handed
  to the inner index is `[(1,0)='aaa', (1,1)='bbb', (1,0)='ccc']` — **two entries under the same
  chunk-keyed primary key `(1,0)`**. The inner index's upsert deletes the keys and then appends the
  whole batch, so both survive: one primary key, two rows.
- Chunk `(1,1)='bbb'` belongs to the superseded text and is never overwritten, so the row stays
  searchable by a word its current text does not contain.

The NULL case is the same shape without the duplicate: `(id=1, 'value')` followed by
`(id=1, NULL)` leaves the `'value'` chunks in the index, while the accelerated table resolves row 1
to NULL. A search returns text for a row whose search value is now empty.

### Impact

Search results disagree with the table they are an index of, with no error. Correctness of results,
not availability. It needs a source that emits more than one change for a row inside one CDC poll,
which is ordinary rather than exotic.

### Where

- `crates/runtime/src/changes/mod.rs` — `index_change_envelope`, which hands the change batch to
  each index's `compute_index` unchanged.
- `crates/search/src/index/chunking.rs` — `ChunkedSearchIndex::write`, where two rows sharing a key
  become two entries under one chunk-keyed key.

### Acceptance

- A batch carrying two changes for one primary key indexes only the last one, on both a chunked and
  an unchunked index.
- The chunk-keyed primary key is unique in what the inner index holds after such a batch.
- A regression test asserting what the index holds after the write, not what the write returned.

### Origin

Raised by an adversarial review of the fix for #13704 and confirmed against the code. It is not
introduced by that fix — the same batch produces the same index contents on `trunk` — so it is
filed rather than folded in.

claudespice:
## Summary

A write batch can carry more than one change for the same primary key. A CDC change envelope holds
every change the source produced in one poll, so two updates to one row inside that window arrive
as two rows of one batch. The accelerated table resolves such a key to the last change; every
search index wrote all of them, and then disagreed with the table it is an index of — with no
error, and needing only a source that emits two changes for a row in one poll.

The disagreement takes a different shape per index but has one cause, so it is repaired once:

- **A store whose upsert deletes the batch's keys and then appends the batch keeps both rows.** One
  primary key ends up with two entries, one of them the superseded content.
- **A chunked index chunks each row independently and concatenates.** For `(id=1, 'aaa bbb')`
  followed by `(id=1, 'ccc')` the inner index is handed
  `[(1,0)='aaa', (1,1)='bbb', (1,0)='ccc']` — two entries under the chunk-keyed primary key
  `(1,0)`, and a `(1,1)='bbb'` chunk belonging to text the row no longer has, which the shorter
  winning text never overwrites. The row stays searchable by a word its current text does not
  contain.
- **The full-text index deletes each key's existing terms and then adds every document it was
  given**, so the superseded content becomes a second document.

## Changes

`crates/search/src/index/coalesce.rs` (new) resolves a write batch to one row per primary key —
the last change for that key — and maps the write's output back onto the batch's original rows, so
the caller's row-for-row contract with the change batch still holds and the accelerator still
applies every change. A superseded row carries its winner's derived columns, which is what keeps
the index and the table agreeing: the table resolves that key to the same winning change, so a
superseded row's own derived values are never observable.

Row identity is the declared `SearchIndex::primary_fields` under `arrow::row`'s encoding — the same
notion of identity the stores' upsert and `delete_by_keys` paths already assume. Three cases are
deliberately pass-through, leaving today's behaviour untouched: no declared primary key, a key
column the batch does not carry, and a row whose key is NULL (SQL never treats NULL as equal to
NULL, so such a row has no identity to resolve and is never coalesced with anything).

Applied at every `Index::compute_index` that maintains a store keyed by primary key: chunked,
compound, in-memory vector, S3 Vectors, both Elasticsearch indexes, and the full-text index. The
co-located indexes (`NativeVectorIndex`, `DuckDBVectorIndex`) are untouched on purpose — they only
augment the batch and the accelerated table resolves the key itself.

The fast path costs one key encoding per batch and nothing else: a batch whose keys are all
distinct is handed to the write untouched, and the NULL scan is skipped outright when no key column
has any nulls.

## Scope

Both production callers of `compute_index` — the CDC path (`crates/runtime/src/changes/mod.rs`) and
the sink write path (`runtime-datafusion-index`'s index scan) — hand over exactly one batch, so
per-batch resolution covers them. `reduce_batches` resolves across a whole slice anyway, for the
full-text index which indexes its batches as a set.

What this does **not** close is a shrink that spans two batches: a row rewritten from `'aaa bbb'`
to `'ccc'` in a *later* batch still leaves its `(1,1)` chunk behind, because the second write only
names the chunk ids the current text produces. That is a different mechanism — it needs the write
to know what the previous one left — and this issue does not report it. Measured on `trunk` and
filed as #13717.

## Test plan

- `make lint` — clean.
- `make nextest` — passes (`make signoff` runs both).
- New regression tests, each asserting **what the index holds after the write**, not what the write
  returned, and each verified to fail with the coalescing disabled:
  - `chunking`: a row changed twice in one batch is chunked only from its last change (the inner
    index holds `[(2,0,"ddd"), (1,0,"ccc")]`, not `[(1,0,"aaa"), (1,1,"bbb"), (2,0,"ddd"),
    (1,0,"ccc")]`), and the chunk-keyed primary key is unique in what it holds.
  - `chunking`: a search value cleared later in the batch is not chunked from its old text.
  - `memory`: a row changed twice in one batch is stored once, as the change that supersedes the
    other — asserted on the stored embedding, not just the key count.
  - `text_search`: the same row is one tantivy document, and its superseded content is not
    searchable.
  - `coalesce`: 11 unit tests over the helper — repeated key, no repeat, NULL search value, NULL
    keys, no declared key, a key column absent from the batch, a multi-column key, a write
    returning the wrong row count, and the cross-batch `reduce_batches` cases.

## Review gate

- Adversarial review: **skipped** — `codex` exited 1: "Your workspace is out of credits"; `grok` not installed
- `/security-review`: no findings — the change builds no SQL, path, or outbound request, adds no dependency and no `unsafe`, and its two error strings carry configuration column names and row counts, never row data
- `/simplify`: applied — the NULL scan now short-circuits on `null_count()`, and the repeated call-site comment was cut to one line; light checks re-run green

Fixes #13713
Refs #13717

## Checklist

- [x] Regression tests fail on the old code and pass on the new
- [x] The whole bug class is covered, not the one reported index
- [x] No behaviour change for a batch whose primary keys are already distinct
- [x] `cargo fmt --all` clean

### DIFF
--- a/crates/search/src/generation/text_search/index.rs
+++ b/crates/search/src/generation/text_search/index.rs
@@ -49,6 +49,7 @@ use crate::generation::text_search::{
 };
 use crate::generation::util::get_primary_keys;
 use crate::index::SearchIndex;
+use crate::index::coalesce;
 
 /// The heap budget for the [`tantivy::IndexWriter`] (150 MiB).
 /// A larger budget reduces the number of segment flushes and subsequent merges,
@@ -263,7 +264,15 @@ impl Index for FullTextDatabaseIndex {
         &self,
         batches: Vec<RecordBatch>,
     ) -> Result<Vec<RecordBatch>, DataFusionError> {
-        if let Err(e) = self.update_index(batches.as_slice()).await {
+        // A batch can carry more than one change for the same primary key, which the table this
+        // indexes resolves to the last one. `update_index` deletes each key's existing terms and
+        // then adds every document it was given, so without this the collection would hold a
+        // second document for the superseded content — see [`coalesce`]. Only what is indexed is
+        // reduced; the caller's batches are handed back untouched.
+        let primary_key = self.primary_fields();
+        let reduced = coalesce::reduce_batches(&primary_key, batches.as_slice())?;
+        let indexed = reduced.as_deref().unwrap_or(batches.as_slice());
+        if let Err(e) = self.update_index(indexed).await {
             tracing::error!("Failed to update full text search index: {e}");
             return Err(DataFusionError::External(Box::new(e)));
         }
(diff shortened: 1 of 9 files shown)


claudespice:
@copilot review

PR #13719, opened 2026-08-29
Reclaim the deletion index of an explicit deletion_mode: position table

### LINKED ISSUE #13676: Cayenne: an explicit `deletion_mode: position` PK table has no path to reclaim its in-memory key deletion index
### Summary

#13620 reports that the seq-prefix bake never runs on a primary-key table left
on the default `cayenne_deletion_mode: auto`. #13671 fixes that by resolving
`Auto` to `Key` when a primary key exists.

That fix moves the **default** configuration onto the working key-delete path. It
does not change a table configured with an explicit `cayenne_deletion_mode:
position`, which keeps the defect #13620 describes: such a table records key
tombstones for every row whose file-local position is unknown, and no pass prunes
them, so its **in-memory** key deletion index grows for the life of the process.

Scope note: the disk and metastore halves of #13620 are being handled elsewhere.
#13663 schedules the orphaned-DV sweep from every publication that orphans a
vector — the bake, the size-tier merge, and a startup replay — which also makes
the bake's prune effectively durable, since the sweep removes the
`cayenne_delete_file` rows once the surviving-sequence floor passes them. #13658
deletes merged-away `cayenne_snapshot_file` rows (#13626). Both of those signals
also fire for a `position`-mode table, whose size-tier merge does run. What is
left here is that table's in-memory deletion index, and the memory ceiling that
should bound it.

All line numbers are at `0e68c67f03`. #13671 is based on an older commit, so its
line numbers differ.

### Which configurations this covers

Using the labels from
[this comment](https://github.com/spiceai/spiceai/issues/13620#issuecomment-5455037702):

| Label | PK | `deletion_mode` as the engine sees it | Fixed by #13671? |
|---|---|---|---|
| A | no | anything | n/a — PK-less tables keep position deletes, which is correct |
| B | yes | `Key` | already worked |
| C | yes | `Auto` (the default) | **yes** — `resolved()` now returns `Key`, so C behaves as B |
| D | yes | `Position` (explicit) | **no** — this issue |

Rows C and D are code-identical before #13671. After it, C joins B and D is left
alone.

### Why row D is not simply "the user asked for it"

An explicit `position` config does not mean the table records only position
deletes. A position delete needs a known `(file path, file-local position)`, and
`RowLocation` (`pk_index.rs:175-188`) shows only `FilePositioned` has one. Every
other case falls back to a key tombstone, which the `Position` doc states
directly (`metadata.rs:519-526`):

> Rows whose position is unknown (cold-rebuilt keysets, over-budget bloom
> tables, inlined rows) fall back to the `Key` path, so a table can mix both
> within one snapshot.

On top of that, a filter-based `DELETE` on **any** primary-key table records key
tombstones regardless of the mode — the dispatch is on `is_position_based()`,
i.e. PK presence (`sink.rs:1175-1186`). That covers SQL `DELETE … WHERE`,
retention eviction (`apply_retention_filters`, `table.rs:21099-21131`), and CDC
`DELETE` events through the PK-IN fast path (`sink.rs:718-726`).

So a row-D table accumulates a key deletion index in normal operation, and today
nothing prunes it.

### 1. Give row D a reclamation path

Two gates stand between a row-D table and the only pass that clears its
tombstones — the current-snapshot full rewrite, whose
`RewriteScope::AllTombstonesFoldedSnapshots` arm clears the in-memory caches
(`table.rs:17809`) and commits with `cutoff = i64::MAX`, wiping the durable
delete-file rows too (`table.rs:4074-4088`). Either gate alone prevents it:

**1a. The entry gate `try_lock`s where its callee blocks.**
`compact_current_snapshot_small_files` skips the whole pass when a writer holds
`write_lock` (`table.rs:14943`), while `rewrite_current_snapshot_for_compaction`
— the function it calls — takes the same lock with a **blocking** acquire and
documents why (`table.rs:17396-17399`):

> Blocking acquire (not try_lock): the full-snapshot rewrite must always make
> progress — a try_lock+defer would starve it on a busy table (the documented
> "files accumulate unboundedly" failure mode)

The caller and callee were written for the same concern and disagree. Aligning
the caller with the callee gives a continuously-written row-D table a rewrite
that actually runs.

**1b. The counter gate does not see protected-snapshot publishes.**
`compact_current_snapshot_small_files` returns at `table.rs:14931` unless files
accumulated in the CURRENT snapshot. The counter,
`new_files_since_last_compaction`, is advanced only by
`record_current_snapshot_files_added` (`table.rs:5926`) — the durable
current-snapshot file-move paths and the memory-mode CDC checkpoint
(`table.rs:26141`). An upsert publishes a PROTECTED snapshot, which does not
advance it.

`schedule_post_write_compaction` already consults
`protected_snapshot_maintenance_trigger()` when deciding whether to schedule a
pass (`table.rs:15389`), but pass 2 re-checks only the file counter, so a pass
scheduled by protected-snapshot age or count declines anyway. The trigger exists
and does not reach the rewrite it was meant to fire. Threading it into pass 2 is
the fix.

**Not this:** deriving the bake's `key_mode` from `pk_deletion_strategy` so the
bake itself runs on row D. That would run the bake on a table whose tombstones
are file-path scoped, where a concurrent writer's tombstone can target an input
file the bake swaps away — the resurrection `compact_protected_snapshots_subset`
takes `write_lock` to prevent (`table.rs:19053-19060`). The bake takes no write
lock. If the bake is ever to serve row D, it needs that serialization first, and
that reintroduces the starvation trade-off — so measure 1a + 1b before
considering it.

### 2. Give the memory ceiling an action on row D

#13671 folds in the predicate correction: the deletion-index memory ceiling is
computed from whether the table keeps a key deletion index
(`!pk_deletion_strategy.is_position_based()`) rather than from the mode. That
stops row A — a PK-less table with no key index — from entering the bake branch
and logging a forced-bake warning every tick.

It does not give row D an action. `over_mem_ceiling` is consumed inside
`if key_mode && …` (`table.rs:31029`), and `key_mode` stays false for row D, so
the condition is computed and then ignored. A row-D table can still drift past
`runtime.query.memory_limit` with nothing forcing action — and the reservation
"over-commits the pool infallibly and can never be dropped"
(`table.rs:9670-9680`).

Whatever lands for item 1 is what makes the ceiling actionable. Until then, a
row-D table over the ceiling should at minimum say so once per episode, naming
the dataset and the `cayenne_deletion_mode: key` remedy — the
`position_compaction_skip_streak` latch (`table.rs:19072-19087`) is the existing
pattern for a once-per-episode maintenance warning.

### 3. Superseded: scheduling the orphaned-DV sweep

Originally filed here. #13663 covers it and more — it signals the sweep from the
bake, the size-tier merge, and a startup replay, and fixes four defects that
scheduling it more often makes reachable (a signal lost across a crash, an
`AtomicBool` coalescer that dropped work, eligibility proved off-fence, and a
floor that ignored the cold tier). Nothing to do here once that merges.

### 4. Stale comments

- `table.rs:15644-15646` — "compaction's bake reclaims the in-memory tombstone
  and the catalog row, but not the physical file", and names
  `prune_deletion_caches_after_full_rewrite` as the bake's prune. The bake
  reclaims neither the catalog row nor the file (its only catalog write is
  `swap_protected_snapshots`, which touches `cayenne_snapshot_sequence` alone),
  and its prune is `prune_deletion_index_at_or_below`. #13663 reflows this
  comment but carries both claims forward verbatim, so it needs correcting
  wherever it lands last.
- `crates/cayenne/src/provider/delete/vector_io.rs:954` — labels the key-based
  vector schema "tables with primary key", which reads as if primary-key presence
  chose the mode. It selects the vector schema.

(The two `docs/cayenne/cayenne.md` statements originally listed here — the
deletion index "does not grow without bound" and the durable delete files being
pruned in the same step as the index — are fixed in #13671.)

### Non-goals

- Changing the resolution of `auto` — that is #13671.
- The inline-tombstone gap (#13621) and the manifest-row gap (#13626, PR
  #13658).
- Reclaiming deletion-vector files and their metastore rows (#13663).
- The overlap between position and key tombstones for an already-located upsert
  conflict (`deletion_strategy.rs:229-232` describes the two forms as disjoint,
  `table.rs:11558-11577` records both). It costs one extra index entry per
  located conflict and does not affect the tables in this issue, which capture
  no positions.

### Acceptance criteria

- A primary-key table with `cayenne_deletion_mode: position` under continuous
  writes reaches a bounded steady-state deletion index, demonstrated by a test
  that drives writes and maintenance rather than by inspection.
- A row-D table over the deletion-index memory ceiling either sheds memory or
  emits one actionable warning per episode naming the dataset and the remedy.
- The two statements in item 4 match the code.

bjchambers:
Closes #13676.

## What was wrong

A primary-key table on `cayenne_deletion_mode: position` is not a position-deletes-only table. A position delete needs a known `(file path, file-local position)`; every case without one — a cold-rebuilt keyset, an over-budget bloom table, an inlined row — records a **key** tombstone instead, exactly as `DeletionMode::Position`'s own documentation says. So does every filter-based `DELETE` on any primary-key table, because that dispatch is on key presence rather than on the mode: SQL `DELETE … WHERE`, retention eviction, and CDC deletes through the PK-IN fast path.

Such a table therefore accumulates a key deletion index in normal operation — and nothing pruned it. The seq-prefix bake, which drains a key-delete table's index, declines a position table by design: its tombstones are file-path scoped, and the bake takes no `write_lock`, so a concurrent writer's tombstone could target an input file the bake had swapped away. That leaves the current-snapshot full rewrite as the only pass that clears the index, and two gates kept it away from it.

**The entry gate.** The pass runs only when files accumulated in the CURRENT snapshot. An upsert publishes a PROTECTED snapshot, which never advances that counter, so a table could upsert indefinitely and never qualify. The index grew for the life of the process, over-committing the query pool with a reservation that can never be dropped.

**A latent self-deadlock behind it.** The pass `try_lock`ed `write_lock` for a position table and held it across the rewrite it ends in — and `rewrite_current_snapshot_for_compaction` acquires that same non-reentrant `tokio::sync::Mutex` blockingly, by deliberate design, so the full rewrite always makes progress. The subset rewrite, the only other outcome, is never reached by a position table (`subset_rewrite_eligibility` declines it), so the two always meet. Only the file counter kept them apart. Opening the entry gate alone reaches it, and the pass then parks forever holding the lock every writer needs — verified by driving it: a background compaction task wedged and every subsequent write hung.

## What this does

- The deletion index raises the pass itself, on the same trigger the bake uses and on the hard memory ceiling above it, routed straight to the full rewrite. Scoped to the table shape with no other reclaimer, so a key-delete table keeps the cheaper bake and a primary-key-less table, which records no key tombstones, is untouched.
- That gives the **memory ceiling an action** it never had — it was computed inside the key-mode branch and discarded for a position table. It now forces the rewrite and says so once at WARN, naming the dataset and the `cayenne_deletion_mode: key` remedy, because the rewrite it forces is the expensive bound.
- The caller's `write_lock` guards are **removed** rather than reconciled: the callee already takes both locks correctly, so the guards bought nothing and starved the pass on exactly the tables that need it.
- The reclaim reads the **configured** trigger, not the adaptive one. The controller lowers that value — as far as 1,000 — to bake more often, a trade priced against the incremental bake; applied to a full rewrite it would ask a large table to re-encode itself continuously.
- `>` becomes `>=` in the entry gate, matching the two schedulers that decide whether the pass is worth spawning; at exactly `compaction_trigger_files` they spawned a pass this gate then declined.

## Why not the alternatives

Threading `protected_snapshot_maintenance_trigger()` into this pass, as the issue suggests, would make it full-rewrite whenever the protected count or age fires — on **every** CDC table, preempting the size-tier subset merge that already serves them. That trades a bounded incremental merge for an O(current snapshot) re-encode across the board. The gap is specific to the deletion index, so the trigger is too.

Deriving the bake's `key_mode` from `pk_deletion_strategy` so the bake itself runs on a position table needs write-lock serialization first, which reintroduces the starvation trade-off the issue asks to measure before considering.

## Verification

New integration test drives upsert rounds and maintenance and asserts the index reaches a bounded steady state. With the fix reverted it fails cleanly and reports the whole series:

```
deletion index grew past its steady state: peak 15000 > 3500,
observations [1500, 3000, 4500, 6000, 7500, 9000, 10500, 12000, 13500, 15000]
```

Perfectly linear — one tombstone per superseded row, never reclaimed. A premise guard in the test caught an earlier workload that proved nothing (re-upserting the *same* keys replaces tombstones rather than adding them, so the index sat flat).

A key-mode twin pins that the cheaper path still reclaims and that both modes return identical rows; a unit test pins the trigger's scope. Full `cayenne` suite green (40 test binaries), `make lint-rust PACKAGES=cayenne` clean.

## Also here

- `docs/cayenne/cayenne.md` and the deletion-vector schema comments said key presence selects the vector schema; it is selected per delete, by whether that delete has a position to name. The "does not grow without bound" paragraph attributed the bound solely to the bake, which is true only for a key-delete table.
- `sweep_orphaned_deletion_vectors` justified itself with "compaction's bake reclaims the in-memory tombstone and the catalog row, but not the physical file". The bake leaves the catalog row too — its only catalog write is the protected-snapshot swap — which is a stronger argument for the sweep, not a weaker one. It also named the wrong prune function.
- A CH-benCHmark pod pinning `position`, the twin of the adaptive pod that leaves the mode unset (`auto` ⇒ `key`), so the two modes can be measured against each other at SF-1000. Results to follow on this PR.

## Relationship to #13671

Independent and based on trunk. #13671 changes how `auto` resolves; this changes what an *explicit* `position` table does once resolved. They overlap textually in `table.rs` and whichever lands second will need a merge.


bjchambers:
## SF-1000 A/B, first pair — and a bug it caught in this PR

Both arms at `91c09eda88`, `duration=1800 terminals=100 rate=9250`, xlarge runners.

| Arm | Pod | Result |
|---|---|---|
| control (`auto` ⇒ `key`) | `postgres-cayenne[file]-adaptive.yaml` | ✅ QPH 1022.5 (515 queries / 1813s), tpmC 248,763, replication converged in 1,445s, row-count + analytical gates passed |
| position | `postgres-cayenne[file]-adaptive-position.yaml` | ❌ never reached ready — `Spiced instance not ready within 900s. 1/11 datasets not ready` |

The position arm's failure is **this PR's bug, not a property of the mode.**

`order_line` completed its 300M-row bootstrap at 00:35:17 and then never caught up. The only thing in the log afterwards is the controller climbing write concurrency against a lag it could not close:

```
00:34:17  order_line  write_concurrency  new_value=11  "replication-lag goal, buffers maxed: raise write concurrency"
00:38:18  order_line  write_concurrency  new_value=18  "replication-lag goal, buffers maxed: raise write concurrency"
00:42:18  order_line  write_concurrency  new_value=25  "replication-lag goal, buffers maxed: raise write concurrency"
00:43:01  Error: Spiced instance not ready within 900s. 1/11 datasets not ready
```

The seq-prefix bake defers when the CDC apply is at or over capacity (`bake_should_defer_for_apply`), overridden only by the memory ceiling. I gave the position reclaim the bake's *trigger* but not its *gate* — and the reclaim is much the more expensive of the two: it re-encodes the whole current snapshot and holds `write_lock` throughout, where the bake rewrites the settled protected prefix and takes no write lock. So it fired precisely where it must not, starving the apply that would have closed the lag and keeping the table behind it.

Fixed in `77edaf07a2`: the reclaim takes the same gate, with the same memory-ceiling exception. A table whose apply never gets headroom now reclaims only when the ceiling forces it — the bake's bargain, and the honest one, since the alternative is a rewrite that makes the lag it competes with permanent. The new unit test drives apply-vs-arrival past the back-pressure ratio and asserts both directions; the one that matters is that an over-ceiling index still fires *through* the defer.

Re-running the position arm against the new SHA. The control arm above stands either way.

### Reading the control numbers

QPH 1022.5 and converged-in-1445s are one run, and this workload's known noise is wide — QPH has run ~28% CoV and SF-1000 freshness is bimodal (~399-404s vs ~549-565s in past series). Treat them as a passing baseline, not a precise figure.

Jeadie:
Move out of v2.2.1 to ensure we get adequate time to bake/test.

Issue #13747, opened 2026-08-30
Three ListModels providers classify an error by substring: a model id containing 401 reads as a bad key, and a real 401 does not


claudespice:
## Summary

Three `ListModels` implementations decide what kind of failure they hit by substring-matching the
provider error's rendered text, and each is wrong in both directions — it misreports an unrelated
error, and it misses the real one.

| Site | Discriminator |
|---|---|
| `crates/llms/src/openai/list_models.rs:75-89` | `message.contains("401")`, `"Unauthorized"`, `"429"`, `"rate"`, `"402"`, `"quota"` |
| `crates/llms/src/xai/list_models.rs:75-87` | same, minus the quota arm |
| `crates/llms/src/spiceai/list_models.rs:91-103` | same |

`message` is `e.to_string()` of an `async_openai::error::OpenAIError`, which renders the provider's
own message — and the provider echoes request detail back into it.

## Measured

Probed `OpenAiModelLister` against a local server standing in for the OpenAI API, so the whole path
runs (HTTP status, error body, `async-openai`'s `read_response` mapping, the lister's
classification). One request per row:

```
[model id containing 401] -> InvalidCredentials { provider: "OpenAI" }
[real invalid api key]    -> NetworkError { provider: "OpenAI", message: "invalid_request_error: Incorrect API key provided: sk-xxx. (code: invalid_api_key)" }
```

- **Row 1** — a `404` whose message is ``The model `gpt-4o-mini-2024-0401` does not exist``. The
  `0401` in a snapshot date matches `contains("401")`, so a model-not-found is reported to the user
  as `InvalidCredentials` — it sends them to rotate a key that is working.
- **Row 2** — a real `401` with `code: invalid_api_key`. `OpenAIError`'s `Display` renders the
  `type`, the message and the `code`, and none of those spell `401` or `Unauthorized`, so the
  genuine credential failure is reported as `NetworkError`.

The probe (a temporary `crates/llms/tests/` file, not committed):

```rust
let base = serve_one_error(status, body); // one-shot local HTTP server, canned status + body
let lister = OpenAiModelLister::new(&SecretString::from("k".to_string()), Some(&base));
let err = lister.list_models().await.expect_err("expected a failure");
println!("[{label}] -> {err:?}");
```

`crates/llms/tests/anthropic_stream_errors.rs` (added by the PR below) is the committed version of
that harness and is the template for a regression test here.

## Expected behavior

[...]

PR #13748, opened 2026-08-30
fix(anthropic): classify a streaming failure by Anthropic's error type, not its message text (fixes #13562)

### LINKED ISSUE #13562: Anthropic streaming errors are classified by substring, so an unrelated error containing '403' or '429' is reported as auth or rate-limit
## Summary

`format_anthropic_stream_error` in `crates/llms/src/anthropic/types_stream.rs` decides what kind of failure it is by searching the lowercased error *message* for substrings:

```rust
let lowered = api_error.message.to_lowercase();

if lowered.contains("too many requests") || lowered.contains("429") { /* rate limit */ }

if lowered.contains("401")
    || lowered.contains("403")
    || lowered.contains("authentication")
    || lowered.contains("unauthorized")
    || lowered.contains("forbidden")
{ /* authentication */ }
```

Anthropic already states the kind in the error's `type` field (`authentication_error`, `rate_limit_error`, `invalid_request_error`, `overloaded_error`, `not_found_error`), which is what the classification should read. Sniffing the message means any error whose text happens to contain one of those tokens is reported as something it is not, and the replacement message *discards the original*, so the real cause is unrecoverable from the log:

> Anthropic authentication failed. Verify your Anthropic API key and workspace permissions.

The message is attacker-independent but very much user-content-dependent: Anthropic echoes request detail back in `invalid_request_error`, so a prompt, tool name, or model id carrying `403`/`429` is enough.

## Concrete instance already demonstrated

A model id whose snapshot date contains `403` — `claude-3-5-sonnet-20240403` — produces the message `model: claude-3-5-sonnet-20240403`, which matches `lowered.contains("403")`. That one is now covered: #13557's fix returns a `not_found_error` untouched before the substring tests run, and `a_model_id_containing_403_is_not_reported_as_an_auth_failure` in `types_stream.rs` pins it. Every *other* error type still reaches the sniffing.

## Expected behavior

Classify on `api_error.r#type`, falling back to the substring tests only when the type is absent (a proxy in front of Anthropic may omit it), and keep the original message as the cause rather than replacing it.

## Notes

Found while fixing #13557; out of scope for that PR, which only stops the `not_found_error` path from being re-typed here.

claudespice:
## Summary

`format_anthropic_stream_error` decided what kind of failure Anthropic had reported by searching the
lowercased error **message** for `429`, `403`, `401`, `authentication`, `unauthorized` and
`forbidden` — then threw that message away and replaced it with a fixed remediation string.

Anthropic states the kind in the error's `type` field, and echoes the caller's own request detail
back into the message of an `invalid_request_error`. So the message was both an unreliable
discriminator and the only record of which request had failed. It was wrong in **both** directions,
and the second one is the more common:

- a malformed request whose echoed detail happens to contain `429` or `403` — a tool name, a prompt,
  a model id — was reported as a rate limit or a credential failure, with the real cause deleted;
- a **real** `rate_limit_error` or `authentication_error` was reported as neither, because
  Anthropic's own wording for them (`Number of input tokens has exceeded your per-minute limit`,
  `invalid x-api-key`) contains none of those tokens.

Two further gaps in the same function, both found by the review gate and both confirmed by running
the path:

- **A mid-stream failure never reached the classifier at all.** Once a generation has started,
  Anthropic answers HTTP 200 and emits an `error` event — an `overloaded_error` when it sheds load,
  and the other error types with it. `MessageCreateStreamResponse` had no variant for that packet,
  so it failed deserialization and the caller got the serde error instead of the failure.
- **`overloaded_error` and `permission_error` had no arm of their own.** The overload said nothing
  about being the retryable one, and the 403 was reported as an authentication failure — sending the
  caller to rotate a key that is valid and merely lacks access.

## Changes

| File | Change |
|---|---|
| `crates/llms/src/anthropic/types_stream.rs` | Classify on `api_error.r#type` against Anthropic's taxonomy (`rate_limit_error`, `overloaded_error`, `authentication_error`, `permission_error`), keeping the message tests only for an error that carries no `type`. Carry Anthropic's own message through as the cause in every arm. New `Error { error: ApiError }` stream variant so a mid-stream `error` event is decoded and reported like any other failure. |
| `crates/llms/tests/anthropic_stream_errors.rs` | New end-to-end suite: 10 cases driving the whole path — HTTP status, error body, the SSE client's own error mapping, and the classification — against a one-shot local server standing in for Anthropic. No credentials. |

`not_found_error` still passes through untouched: it arrives already explained by
`explain_model_not_found`, and that explanation plus its type is what a downstream check reads.

**The untyped fallback is not dead code, and the suite pins it.** A proxy in front of Anthropic may
answer without a `type` — and so does a 5xx from Anthropic itself, because `async-openai`'s
`read_response` deliberately does not parse a server-error body and hands the whole thing over as an
untyped message. A *pre-stream* `overloaded_error` (529) therefore lands in the unclassified arm
while the same condition delivered mid-stream is typed and classified. Settling that asymmetry means
parsing the 5xx body, which belongs to the client rather than to this function.

## Measured

Ran the new suite against the **unmodified** classifier first — the whole path, not the function in
isolation. Five of the eight cases then in the suite failed, and each failure is the defect:

```
test an_invalid_request_naming_429_is_not_reported_as_a_rate_limit ... FAILED
  a malformed request is not a rate limit: Anthropic API rate limit exceeded. Check your limits at
  https://console.anthropic.com/settings/limits and retry shortly.

test an_invalid_request_naming_403_is_not_reported_as_an_auth_failure ... FAILED
  a malformed request is not a credential problem: Anthropic authentication failed. Verify your
  Anthropic API key and workspace permissions.

test a_rate_limit_error_is_reported_as_a_rate_limit_whatever_its_message_says ... FAILED
  a rate_limit_error must be reported as one: Anthropic streaming error: Number of input tokens has
  exceeded your per-minute limit

test an_authentication_error_is_reported_as_one_and_keeps_its_cause ... FAILED
  an authentication_error must be reported as one: Anthropic streaming error: invalid x-api-key

test a_permission_error_is_reported_as_an_authentication_failure ... FAILED
  Anthropic streaming error: Your API key does not have permission to use the specified resource

test result: FAILED. 2 passed; 5 failed
```

Note the first two: the echoed request detail is gone from the reported message entirely.

The mid-stream gap, probed the same way — HTTP 200, a `message_start` then a documented `error`
event — on the code as it then stood:

```
item 1: Ok(id=msg_1)
item 2: Err(JSONDeserialize(Error("unknown variant `error`, expected one of `message_start`,
  `content_block_start`, `ping`, `content_block_delta`, `content_block_stop`, `message_delta`,
  `message_stop`", line: 1, column: 15),
  "{\"type\":\"error\",\"error\":{\"type\":\"overloaded_error\",\"message\":\"Overloaded\"}}"))
```

On this branch, all 10 pass:

```
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

**Each guard was confirmed by neutering the code it pins — 9 neuters, 9 caught.** Reverting the
typed classification fails 5; dropping `permission_error` from the auth arm, dropping the untyped
fallback, dropping the `not_found_error` pass-through, and dropping the cause from the message each
fail exactly the tests that name them; dropping the mid-stream variant's reporting fails 2, and
routing it around the classifier fails 1; folding `overloaded_error` back into the unclassified arm
and `permission_error` back into authentication each fail 1.

## Test plan

- `make lint` / `make test` / `make nextest` — green, via `make signoff` on the pushed head:

  ```
  Summary [ 620.251s] 12181 tests run: 12181 passed, 27 skipped
  ✓ Signed off on f375b3dd8341 — Signed off by claudespice — lint+test passed in 966s
  ```

- `Attestation` — green (re-run by the sign-off script).
- New: `crates/llms/tests/anthropic_stream_errors.rs`, 10 end-to-end cases; plus
  `a_typed_error_is_never_classified_from_its_message` in `types_stream.rs`, which pins the core
  invariant directly on the classifier.

## Review gate

- Adversarial review: **skipped** — `codex` exited 1: "Your workspace is out of credits"; `grok` not installed. Two earlier `codex` passes did complete against this branch; both are accounted for in the table below.
- `/security-review`: no findings. The only newly-exposed text is Anthropic's own message on the rate-limit/auth arms; the generic arm already carried it, Anthropic does not echo key material into those messages, and the three `Anthropic*Error` type strings have no consumer anywhere in `crates/` or `bin/`, so no authz or retry decision can be changed by reclassification.
- `/simplify`: applied — the test helpers now read the error through one borrowed accessor instead of cloning `String`s. One cleanup was **reverted**: collapsing the mid-stream arm into a pre-match normalization does not narrow the enum type, so the match arm is still required — and keeping it explicit is what makes exhaustiveness compiler-checked.

**On the adversarial passes, in full**, since the attesting one is a skip:

| Pass | Job | Verdict | Disposition |
|---|---|---|---|
| 1 | `review-mtfa3l0c-6a8zcf` | needs-attention | 1 finding: mid-stream `error` events bypass the classifier. **Confirmed by running it** (the `JSONDeserialize` output above) and **fixed here**. |
| 2 | `review-mtfazhn4-w6aq66` | needs-attention | 2 findings. `permission_error` conflated with authentication — **valid, fixed here**. Mid-stream overload not distinguishable as retryable — **fixed here**, though its "this is a regression" framing is **refuted**: the prior behavior was an opaque `JSONDeserialize` no caller could act on either, so this is a gap being closed, not one being opened. |
| 3 | — | — | Out of credits before it produced findings. |

Pass 3 attests the current head, and it is a skip. The only change since pass 2 is the
overload/permission split that pass 2 itself asked for.

## A note on the first sign-off run

The first `make signoff` on this head failed at `12180 passed, 1 failed` — the one failure being
`bundled_duckdb_builds_an_hnsw_index_without_installing_vss`, in a crate this diff does not touch.
It is host state, not this branch: a `vss.duckdb_extension` previously installed into
`~/.duckdb/extensions/` shadows the statically linked one, so `duckdb_extensions().install_mode`
reports `REPOSITORY`. Established by A/B on the same binary — fails with the directory present,
passes with it moved aside — and filed as #13758. The run above is with that shadow removed.

## Deferred

The same shape lives in three `ListModels` implementations (`openai`, `xai`, `spiceai`), which
classify from `e.to_string()`. Filed as #13747 with its own measured reproduction — a `404` naming
model `gpt-4o-mini-2024-0401` reports `InvalidCredentials`, and a real `401` with
`code: invalid_api_key` reports `NetworkError`. Not folded in here: three providers, three
taxonomies, each needing its own evidence.

Fixes #13562

## Checklist

- [x] The failure is demonstrated before the fix and absent after, with the same command.
- [x] Every new guard confirmed by neutering the code it pins.
- [x] No credential, token, or internal detail in a log line, error message, or this body.
- [x] `make signoff` green and `Attestation` re-run.

### DIFF
--- a/crates/llms/src/anthropic/types_stream.rs
+++ b/crates/llms/src/anthropic/types_stream.rs
@@ -53,6 +53,13 @@ pub enum MessageCreateStreamResponse {
     MessageDelta { delta: MessageDelta, usage: Usage },
     #[serde(rename = "message_stop")]
     MessageStop,
+    /// Anthropic answers a mid-stream failure with an `error` event over an HTTP 200 stream — an
+    /// `overloaded_error` when it sheds load partway through a generation, and the other error
+    /// types with it. Without a variant for it the packet is a serde failure, and the caller is
+    /// handed the deserializer's "unknown variant" complaint instead of the failure Anthropic
+    /// actually reported.
+    #[serde(rename = "error")]
+    Error { error: ApiError },
 }
 
 #[derive(Debug, Deserialize, Serialize)]
@@ -351,6 +358,17 @@ pub fn transform_stream(
                         | MessageCreateStreamResponse::ContentBlockStop { .. }
                         | MessageCreateStreamResponse::MessageStop,
                     ) => None,
+                    Ok(MessageCreateStreamResponse::Error { error }) => {
+                        // An error Anthropic delivered as a stream packet is the same failure as
+                        // one it delivered as an HTTP status, and is reported the same way.
+                        let formatted_error =
+                            format_anthropic_stream_error(OpenAIError::ApiError(error));
+                        tracing::debug!(
+                            "Received an anthropic error stream packet: {:?}",
+                            formatted_error
+                        );
+                        Some(Err(formatted_error))
+                    }
                     Err(e) => {
                         let formatted_error = format_anthropic_stream_error(e);
                         tracing::debug!(
--- a/crates/llms/src/anthropic/types_stream.rs
+++ b/crates/llms/src/anthropic/types_stream.rs
@@ -436,27 +436,40 @@ fn combine_opt_u32(current: Option<u32>, delta: Option<u32>) -> Option<u32> {
     }
 }
 
-fn format_anthropic_stream_error(error: OpenAIError) -> OpenAIError {
-    let OpenAIError::ApiError(api_error) = error else {
-        return error;
-    };
+#[derive(Debug, PartialEq, Eq)]
+enum StreamErrorKind {
+    RateLimit,
+    Authentication,
+    Other,
+}
 
-    // A `not_found_error` arrives already explained by `explain_model_not_found`, and must not be
-    // re-typed here: the substring tests below read the model id in its message, so a snapshot id
-    // such as `claude-3-5-sonnet-20240403` would be reported as an authentication failure.
-    if api_error.r#type.as_deref() == Some("not_found_error") {
-        return OpenAIError::ApiError(api_error);
+/// What a streaming failure is, resolved from Anthropic's own error taxonomy.
+///
+/// The `type` field is the discriminator, not the message text: Anthropic echoes request detail
+/// back in an `invalid_request_error`, so a prompt, tool name, or model id that happens to carry
+/// `403` or `429` reads as a credential or rate-limit failure under a substring test — and the
+/// replacement message drops the original, leaving the real cause unrecoverable from the log.
+///
+/// The message tests survive only for an error carrying no `type` at all. A proxy in front of
+/// Anthropic may answer that way, and so does a 5xx from Anthropic itself: the SSE client does not
+/// parse a server-error body, and hands the whole thing over as an untyped message.
+fn classify_stream_error(api_error: &ApiError) -> StreamErrorKind {
+    match api_error.r#type.as_deref() {
+        // Anthropic's documented error types. `permission_error` is its 403 and belongs with
+        // `authentication_error`: both are answered by checking the key and its workspace.
+        Some("rate_limit_error") => StreamErrorKind::RateLimit,
+        Some("authentication_error" | "permission_error") => StreamErrorKind::Authentication,
+        Some(_) => StreamErrorKind::Other,
+        None => classify_untyped_stream_error(&api_error.message),
     }
+}
 
-    let lowered = api_error.message.to_lowercase();
+/// The message tests, reached only when the error carries no `type`.
+fn classify_untyped_stream_error(message: &str) -> StreamErrorKind {
+    let lowered = message.to_lowercase();
 
     if lowered.contains("too many requests") || lowered.contains("429") {
-        return OpenAIError::ApiError(ApiError {
-            message: "Anthropic API rate limit exceeded. Check your limits at https://console.anthropic.com/settings/limits and retry shortly.".to_string(),
-            r#type: Some("AnthropicRateLimitError".to_string()),
-            param: api_error.param,
-            code: api_error.code,
-        });
+        return StreamErrorKind::RateLimit;
     }
 
     if lowered.contains("401")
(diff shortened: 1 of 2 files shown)


claudespice:
@copilot review

lukekim (approved):

Issue #13749, opened 2026-08-30
The acceleration.ready_state deprecation warning prints 10-15 times per startup, not once per component


claudespice:
### What

`DatasetBuilder::try_from` and `ViewBuilder::try_from` emit the `acceleration.ready_state`
deprecation warning from inside the conversion, and that conversion runs many times per load —
not once per component. A Spicepod with one dataset and one view that both set the key logs the
warning **15 times** on a single startup.

### Measured

`spiced` (dev build of `trunk` + spiceai/spiceai#13750) against a two-component Spicepod, one
`spiced` start, counted from its own log:

```
dataset 'trips'    -> 10 identical warning lines
view    'trips_vw' ->  5 identical warning lines
```

The dataset's 10 reproduce on unmodified `trunk`, so the repeat is pre-existing and not
introduced by that PR; the view's 5 are new only because `trunk` did not read the key on a view
at all (that is the bug #13615 reports).

### Why

`try_from` has no idea how often it is called. `get_valid_views` / `get_valid_datasets` build
every component afresh, and they are called from several places per load — including
`initialized_sources()` on both `View` and `Dataset`, which the accelerator engine invokes during
setup. `crates/runtime/src/init/view.rs` already records the same shape for a neighbouring
message:

> passing the flag through only made each dataset's load error, and each dataset's
> discarded-acceleration warning, print twice per load.

The `LogErrors` flag exists to suppress exactly this duplication, but it is a parameter of
`get_valid_*`, not of `TryFrom`, so a warning emitted inside the conversion cannot see it.

### Impact

Noise only — no wrong results. It matters because this is a *deprecation* notice: the operator is
being asked to make one edit, and repeating it 10-15 times per start makes the log harder to read
and makes the number of affected components impossible to count by eye. It also scales with the
number of accelerated components, since each one's `initialized_sources()` rebuilds the others.

### Options

- Move the warning to a per-load site that runs once per component (the `load_*` path), leaving
  `TryFrom` a pure conversion. This is the shape that matches `LogErrors` already existing there.
- Or give the conversion a way to say "converted for inspection, not for a load", so only the
  loading call reports.

Either way it should be fixed for **both** components at once — they now share
`deprecated_ready_state_warning`, so they will share the fix.

PR #13750, opened 2026-08-30
fix(views): apply a view's acceleration.ready_state instead of dropping it (fixes #13615)

### LINKED ISSUE #13615: A view's acceleration.ready_state is silently ignored, unlike a dataset's
### Description

A view's `acceleration.ready_state` is read, accepted by the schema, and then never applied — not because the block is disabled, but unconditionally. `ViewBuilder` builds its `ready_state` from the view's **own** top-level field and never looks at the acceleration block's:

```rust
// crates/runtime/src/component/view.rs
Ok(ViewBuilder {
    …
    ready_state: ReadyState::from(view.ready_state),
    …
})
```

A dataset does the opposite — `DatasetBuilder` prefers `dataset.acceleration.ready_state` when it is set, and emits a deprecation notice pointing at `dataset.ready_state`:

```rust
// crates/runtime/src/component/dataset/builder.rs
let ready_state = match dataset.acceleration.as_ref().map(|a| a.ready_state) {
    Some(Some(ready_state)) => { /* warn: deprecated, use dataset.ready_state */ … }
    _ => ReadyState::from(dataset.ready_state),
};
```

So the same key in the same position means one thing on a dataset and nothing at all on a view, and nothing tells the operator. `#[serde(deny_unknown_fields)]` does not help here: the field is a legitimate member of `Acceleration`, so it parses cleanly wherever an acceleration block is accepted.

### Why this is not covered by #13514

#13514 is about settings discarded **because** `acceleration.enabled: false` turns the block off, and its warning's remedy is "remove `enabled: false` to apply them". That remedy is false for a view's `acceleration.ready_state`: enabling acceleration does not make the view honour it either. It is therefore deliberately excluded from that warning (see `CONSUMED_WHEN_DISABLED`) and needs its own handling.

### Reproduction

```yaml
views:
  - name: daily_totals
    sql: SELECT 1
    acceleration:
      enabled: true
      ready_state: on_load
```

`ready_state: on_load` is silently ignored; the view uses the default. Moving it to the view's own top-level `ready_state:` works.

### Expected

One of:

1. `ViewBuilder` honours `acceleration.ready_state` the way `DatasetBuilder` does, including the same deprecation notice — consistent, and makes the key mean one thing everywhere; or
2. the view path warns that `acceleration.ready_state` is not supported on a view and names `view.ready_state` as the field to use.

(1) is the smaller surprise for anyone who learned the dataset behaviour first.

### Notes

Found while reviewing #13602, which extends #13514's warning to views. That PR excludes this field rather than reporting it with a remedy that would not work.

claudespice:
## Summary

A view's `acceleration.ready_state` was accepted by the schema, accepted by the parser, and then
never applied. `ViewBuilder::try_from` built `ready_state` from the view's own top-level field
unconditionally, so the key silently did nothing — while the identical key in the identical
position on a **dataset** is read out of the block and honoured by `DatasetBuilder::try_from`.
Same spelling, same place, two different meanings, and nothing told the operator which one they
had written.

This takes the first of the two shapes #13615 offers, which is the one it prefers: the view now
resolves the key the way the dataset does, deprecation notice included.

It is not gated on `acceleration.enabled`, which is why enabling acceleration did not make the
view honour it either, and why the "remove `enabled: false` to apply them" warning added by
#13602 deliberately excludes this field.

## Evidence

`spiced` (dev build) against a Spicepod carrying one dataset and one view that set
`ready_state: on_registration` in the same position, over an 83 MB / 3 M-row CSV accelerated to
DuckDB. One `spiced` start per row, `.spice` deleted between runs, page cache warmed by a
discarded first run:

| `spiced` | `/v1/ready` → 200 | warnings for the dataset | warnings for the view |
|---|---|---|---|
| `trunk` | **8.97 s** | 10 | **0** |
| this branch | **0.98 s** | 10 | **5** |

The warning counts are the deterministic half and are the direct proof: on `trunk` the dataset
reports the deprecated key ten times and the view reports it **not at all**, because the view path
never reads it. The readiness time is the consequence — `trunk` ignores the view's
`on_registration` and therefore blocks the runtime on the view's initial load, while the branch
honours it and the runtime is ready in under a second.

A control confirms the readiness figure is about *where the key is written* and not about the
data: with the same value moved to each component's own top-level `ready_state`, `trunk` is ready
in 0.72–0.73 s. Timing here is noisy and the first run of any configuration is an outlier (cold
page cache on the 83 MB file), so the figures above are from warmed runs and the deterministic
warning counts are what the claim rests on.

Unit tests alone would not have shown this: they run `kind(=lib)` only, and the resolved value has
to reach `datafusion::register_view` before anything observable changes.

**Neutered to check the guards are load-bearing** — 5 applied, 5 caught. One of them
(`flip-precedence`) initially **survived**, because the precedence test wrote the losing side as
`on_load`, which is the `#[default]` and so indistinguishable from an unset field; it now uses two
distinct non-default values and the same neuter fails it.

## Changes

- `crates/runtime/src/component/view.rs` — resolve `ready_state` from `acceleration.ready_state`
  when set, falling back to the view's own field, matching `DatasetBuilder::try_from` including
  which of the two wins.
- `crates/runtime/src/component/mod.rs` — one `deprecated_ready_state_warning`, beside the
  `disabled_acceleration_warning` it mirrors and keyed by the existing `AcceleratedComponent`, so
  the noun and the reference page follow the component. `AcceleratedComponent` gains
  `reference_url` (unanchored): the remedy is a *top-level* field, so `#acceleration` would send
  the reader back inside the block they are being told to move the setting out of.
- `crates/runtime/src/component/dataset/builder.rs` — the dataset's inline message folds onto that
  shared one, so the two cannot drift.
- `crates/spicepod/src/acceleration/mod.rs` — the `CONSUMED_WHEN_DISABLED` doc comment asserted
  that `ViewBuilder` never reads the field and pointed at this issue; it now states the one reason
  the field is excluded for both components. The `#[deprecated]` note named only the dataset's
  replacement and now names the view's too.

### Scope

Checked the class rather than only the reported field: the only other members of the acceleration
block that `DatasetBuilder::try_from` resolves out of the block are `snapshots` and
`snapshots_compaction`, and a view already receives both through `Acceleration` at registration,
so `ready_state` was the only one dropped. #13477 is reworking view snapshots, so this stays off
that path.

Folding the dataset onto the shared message is in scope under the same rule: its inline version
interpolated the component name **unescaped**, which is the same defect the adversarial pass found
in mine, on the path that has shipped it longest.

## Behavior change

A view that sets `acceleration.ready_state` gets the default (`on_load`) today and gets what it
asked for after this, plus a deprecation warning naming the field to move to. A view cannot have
depended on the value it configured being discarded, but this does change when such a view is
reported ready, so it is worth calling out.

The dataset's deprecation message also changes wording — it gains the remedy sentence and a docs
link, and escapes the name. Nothing asserts its previous text.

## Follow-up filed

- #13749 — the warning prints 10–15 times per startup rather than once per component, because
  `TryFrom` cannot see `LogErrors`. Pre-existing for datasets on `trunk` (the 10 above); measured
  rather than assumed, and deferred because the fix belongs at the `load_*` site for both
  components.

## Test plan
- [x] Added regression test for a view's `acceleration.ready_state` reaching the built view
- [x] Added edge case tests for a disabled acceleration block, the precedence between the deprecated key and the view's own field, the block omitting the key, no acceleration block, and the default
- [x] Added a test that a name carrying a newline survives `ViewBuilder::try_from` and still cannot forge a second log line
- [x] `make test` passes

## Review gate

- Adversarial review: **skipped** — `codex` exited 1: "Your workspace is out of credits. Ask your workspace owner to refill in order to continue."; `grok` not installed. An earlier pass on this branch's first commit *did* run (`codex`, job `review-mtf9hm8m-1fhp4v`, verdict `needs-attention`) and returned one finding: the new deprecation warning interpolated the view name unescaped, so a quoted identifier carrying a newline could break the log line in two and forge a second record. It backed that with a reproduction rather than an assertion. Verified against the repo — `validate_identifier` does accept such a name and `disabled_acceleration_warning` already escapes for this exact reason — then fixed in 592579dd23 with a regression test, and extended to the dataset's copy of the same message in 288b047a4e. The credits ran out before that fix could itself be re-reviewed, so the two later commits carry no adversarial pass.
- `/security-review`: no findings. Considered and dismissed: the log-interpolation above (log spoofing is excluded by that review's rules, and it is fixed here anyway on log-integrity grounds), YAML deserialization in the new tests (typed serde into a `deny_unknown_fields` struct, test-only), and the readiness change itself (operator-elected configuration, no privilege or data boundary crossed, and the same behavior a dataset already had).
- `/simplify`: applied — folded the view's and the dataset's duplicate deprecation text onto one shared builder keyed by the existing `AcceleratedComponent`, moved the wording assertions beside it so they cover both components, and added the unanchored `reference_url`. Light checks re-run green afterwards (`cargo fmt --all`, plus the targeted component unit tests: 55 passed).

Fixes #13615

### DIFF
--- a/crates/runtime/src/component/dataset/builder.rs
+++ b/crates/runtime/src/component/dataset/builder.rs
@@ -23,6 +23,7 @@ use super::{
 };
 use crate::Runtime;
 use crate::component::access::AccessMode;
+use crate::component::{AcceleratedComponent, deprecated_ready_state_warning};
 use app::App;
 use datafusion::sql::TableReference;
 use runtime_acceleration::snapshot::SnapshotBehavior;
@@ -78,8 +79,8 @@ impl TryFrom<spicepod_dataset::Dataset> for DatasetBuilder {
         let ready_state = match dataset.acceleration.as_ref().map(|a| a.ready_state) {
             Some(Some(ready_state)) => {
                 tracing::warn!(
-                    "{}: `dataset.acceleration.ready_state` is deprecated, use `dataset.ready_state` instead.",
-                    dataset.name
+                    "{}",
+                    deprecated_ready_state_warning(AcceleratedComponent::Dataset, &dataset.name)
                 );
                 ReadyState::from(ready_state)
             }
--- a/crates/runtime/src/component/mod.rs
+++ b/crates/runtime/src/component/mod.rs
@@ -72,6 +72,45 @@ impl AcceleratedComponent {
             Self::View => "https://spiceai.org/docs/reference/spicepod/views#acceleration",
         }
     }
+
+    /// This component's Spicepod reference, unanchored. The `#acceleration` anchor
+    /// is wrong for a setting whose remedy is a field *outside* the acceleration
+    /// block, which is what the `ready_state` deprecation asks for.
+    const fn reference_url(self) -> &'static str {
+        match self {
+            Self::Dataset => "https://spiceai.org/docs/reference/spicepod/datasets",
+            Self::View => "https://spiceai.org/docs/reference/spicepod/views",
+        }
+    }
+}
+
+/// Warns that a component set `acceleration.ready_state`, which is honoured but
+/// deprecated in favour of the component's own top-level `ready_state`.
+///
+/// One function for both components because it is one deprecation of one key: a
+/// dataset and a view that write the same thing should be told the same thing,
+/// and a second copy is where the two drift apart.
+///
+/// Built as a pure function so the wording an operator acts on — which component,
+/// the field to move the setting to, and the reference page — is asserted by a
+/// test rather than through whatever a log capture happens to retain.
+pub(crate) fn deprecated_ready_state_warning(
+    component: AcceleratedComponent,
+    name: &str,
+) -> String {
+    // `escape_debug` for the same reason as `disabled_acceleration_warning`:
+    // `validate_identifier` accepts a *quoted* identifier, and a quoted one may
+    // legally contain a newline, so a validated name can still break this line in
+    // two and forge a second one.
+    let name = name.escape_debug();
+    let noun = component.noun();
+    let reference = component.reference_url();
+    format!(
+        "{titled} '{name}' sets `acceleration.ready_state`, which is deprecated and will be removed. \
+        Move the setting to the {noun}'s own `ready_state` to keep it working. \
+        See: {reference}",
+        titled = component.titled_noun()
+    )
 }
 
 /// What to tell an operator whose dataset or view sets `acceleration.enabled:
@@ -122,7 +161,9 @@ pub(crate) fn disabled_acceleration_warning(
 
 #[cfg(test)]
 mod tests {
-    use super::{AcceleratedComponent, disabled_acceleration_warning};
+    use super::{
+        AcceleratedComponent, deprecated_ready_state_warning, disabled_acceleration_warning,
+    };
 
     #[test]
     fn the_warning_names_the_component_the_fields_and_the_remedy() {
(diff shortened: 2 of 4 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #13799, opened 2026-09-02
fix(cayenne): materialize the in-memory CDC tier before a scanning DELETE

lukekim:
## Problem

A predicate `DELETE` on a Cayenne table whose newest row versions live only in the in-memory CDC tier (`cdc_durability: memory`, the CDC default) produced wrong rows:

- a row that existed only in RAM and matched the predicate survived the delete;
- a row whose *superseded durable* version matched the predicate, while its live replacement sat in RAM, had its **key** tombstoned — the live row disappeared even though it did not match.

Both persisted across the next mem-tier checkpoint. The key-based deletion-vector path scans main, the protected snapshots and the cold tier, and its liveness probe reads the durable deletion index; neither sees the RAM tier. #13574 closed this shape for protected snapshots and cold files; the RAM tier was one tier over. The inline-aware sink already handles the tier for delete-all (#11987) and assumes per-key deletes tombstone every tier — true, but a predicate delete has to scan to find its keys.

## Fix

Before the deletion-vector branch of `delete_from` captures its scan sources, materialize the mem-tier under the table `write_lock` (`checkpoint_mem_tier_if_present_for_delete`, a no-op when the tier is empty), exactly as the file-based and position-based branches materialize inline rows. Once durable, the rows are subject to the sink's tier-aware liveness rule: the checkpointed upsert tombstones the version it superseded, the live replacement lives in the protected snapshot with its cutoff, and a RAM-only match is now in a scanned file. `docs/cayenne/cayenne.md` §retention gains the sentence describing this.

## Evidence

Stacked on the regression test (`crates/cayenne/tests/predicate_delete_ram_tier_rows.rs`, parent PR #13798): durable `(7,10),(8,5)`, RAM-only `(7,60),(9,20)`, `DELETE WHERE value < 50` must leave `[(7,60)]`; a second scenario deletes a RAM-only row alone.

Before (trunk `dd85b6510b`):

```text
[durable+ram] DELETE WHERE value < 50 reported 2 row(s); visible now: [(9, 20)] — expected [(7, 60)]
[durable+ram] after the next checkpoint: [(9, 20)] — expected [(7, 60)]
[ram-only] DELETE WHERE value = 20 reported 0 row(s); visible now: [(8, 5), (9, 20)] — expected [(8, 5)]
test result: FAILED. 0 passed; 2 failed   (and, with the reviewer's additions, the count assertions fail there too: reported 2 and 0)
```

After (this branch):

```text
[durable+ram] durable CDC (7,10),(8,5) + checkpoint (epoch 2, 1 data files): [(7, 10), (8, 5)]
[durable+ram] RAM-only CDC (7,60),(9,20), no checkpoint (1 data files): [(7, 60), (8, 5), (9, 20)]
[durable+ram] DELETE WHERE value < 50 reported 2 row(s); visible now: [(7, 60)] — expected [(7, 60)]
[durable+ram] after the next checkpoint: [(7, 60)] — expected [(7, 60)]
[ram-only] durable (8,5) + checkpoint, then RAM-only (9,20): [(8, 5), (9, 20)]
[ram-only] DELETE WHERE value = 20 reported 1 row(s); visible now: [(8, 5)] — expected [(8, 5)]
[ram-only] after the next checkpoint: [(8, 5)] — expected [(8, 5)]
test result: ok. 2 passed; 0 failed
```

End to end, the same shape was probed on the 3-node lab with a real `spiced` and a Postgres CDC source (`cbench probe ram-tier-delete`, spicehq/cayenne-bench-runs#61). Through the runtime's *retention* path the outcome was already correct because that path checkpoints pending rows before it evaluates its predicate; the direct `delete_from` path is the one this PR aligns with it.

## Not in this PR

The same lab probe found a second defect on the durable tier: on an `access: read_write` + `write_mode: write_back` CDC dataset, `DELETE … WHERE value < 50` through `/v1/sql` reported `count: 3` for two live matches and removed the live row in 4 of 5 attempts while the source kept it, with every row already durable. That is an interleaving between the sink's liveness read and the publish of the batch being flushed (the residual noted in #13574's sink), not the RAM-tier blindness fixed here; it has its own reproduction recipe in the probe (`--arm dml`) and needs a separate fix.


lukekim:
Sign-off is recorded on the head (`8a7366b762`): `signoff: success — lint+test passed in 6391s` (remote run 33582668667). The `Attestation` check itself is not listed yet because `pr.yml` only runs for PRs whose base is `trunk`; when the parent (#13798) merges and this PR is retargeted to `trunk`, the check will validate that status as long as the head does not move (a later merge of `trunk` into this branch would need a fresh `make signoff`).

phillipleblanc (approved):

Issue #13850, opened 2026-09-03
A federated sha256 returns the digest's hex text where the kernel returns the 32-byte digest


claudespice:
## What happens

`sha256` exists in both engines, so nothing denies it and nothing rewrites it — the call is pushed
into the accelerated `DuckDB` store verbatim. The two return **different things**:
`DataFusion`'s `sha256` returns the 32-byte digest as `Binary`, while `DuckDB`'s returns the
digest's 64-character *hex text* as `VARCHAR`. `SchemaCastScanExec` then casts that text into the
plan's `Binary` column, so the federated column holds the **ASCII bytes of the hex string** where
the kernel holds the digest. There is no error and no warning; the column is simply 64 bytes of
different data.

## Measured

One local CSV, two datasets over it — `t_duckdb` with `acceleration.engine: duckdb`, `t_local`
unaccelerated — through a real `spiced`, on `trunk` at `f8d83e39f0`:

```
$ curl -s -XPOST localhost:8190/v1/sql -d "SELECT id, sha256(name) AS s FROM t_duckdb ORDER BY id"
[{"id":1,"s":"38656433663661643638356239353965616437303232353138653161663736636438313666386538656337636364646131656434303138653866323232336638"}, …]

$ curl -s -XPOST localhost:8190/v1/sql -d "SELECT id, sha256(name) AS s FROM t_local  ORDER BY id"
[{"id":1,"s":"8ed3f6ad685b959ead7022518e1af76cd816f8e8ec7ccdda1ed4018e8f2223f8"}, …]
```

The federated value decodes to the local one, which is what identifies the defect rather than
merely showing two different strings:

```
federated Binary column: 64 bytes; decoded as ASCII ->
  8ed3f6ad685b959ead7022518e1af76cd816f8e8ec7ccdda1ed4018e8f2223f8
local     Binary column: 32 bytes; digest hex        ->
  8ed3f6ad685b959ead7022518e1af76cd816f8e8ec7ccdda1ed4018e8f2223f8
```

`arrow_typeof` confirms the declared column type on the local side, and the call really is
evaluated remotely:

```
$ … -d "SELECT arrow_typeof(sha256(name)) AS t FROM t_local LIMIT 1"
[{"t":"Binary"}]

$ … -d "EXPLAIN SELECT sha256(name) AS s FROM t_duckdb"
VirtualExecutionPlan name=duckdb compute_context=:memory: \
[...]

Issue #13849, opened 2026-09-03
A federated concat over a NULL argument answers 'z' on DuckDB where the kernel answers NULL


claudespice:
## What happens

`concat` exists in both engines, so nothing denies it and nothing rewrites it — the call is pushed
into the accelerated `DuckDB` store verbatim. The two engines disagree about a NULL argument:
`DataFusion` answers NULL, `DuckDB` ignores the NULL and concatenates the rest. The same query over
the same rows therefore returns a different value, and a different *truth* under `IS NULL`,
depending on whether the dataset happens to be accelerated. There is no error and no warning.

## Measured

One local CSV, two datasets over it — `t_duckdb` with `acceleration.engine: duckdb`, `t_local`
unaccelerated — through a real `spiced`, on `trunk` at `f8d83e39f0`. Row 3's `name` is NULL:

```
$ curl -s -XPOST localhost:8190/v1/sql \
    -d "SELECT id, concat(name,'z') AS c, concat(name,'z') IS NULL AS isn FROM t_duckdb ORDER BY id"
[{"id":1,"c":"alphaz","isn":false},{"id":2,"c":"BeTa gammaz","isn":false},
 {"id":3,"c":"z","isn":false},{"id":4,"c":"Ünïcödéz","isn":false}]

$ curl -s -XPOST localhost:8190/v1/sql \
    -d "SELECT id, concat(name,'z') AS c, concat(name,'z') IS NULL AS isn FROM t_local ORDER BY id"
[{"id":1,"c":"alphaz","isn":false},{"id":2,"c":"BeTa gammaz","isn":false},
 {"id":3,"c":null,"isn":true},{"id":4,"c":"Ünïcödéz","isn":false}]
```

The call really is evaluated remotely — `EXPLAIN` shows it in the SQL sent to `DuckDB`:

```
VirtualExecutionPlan name=duckdb compute_context=:memory: \
  base_sql=SELECT concat("t_duckdb"."name", 'z') AS "c" FROM "t_duckdb"
```

while the control evaluates it in-process:

```
ProjectionExec: expr=[concat(name@0, z) as c]
  DataSourceExec: file_groups={1 group: [[…/data.csv]]}, file_type=csv, has_header=true
```

## Which side is authoritative is the open question

This is **not** simply "make the remote match the kernel", because the kernel's own behaviour
contradicts `DataFusion`'s documentation for the function. `ConcatFunc::invoke_with_args` is
[...]

Issue #13840, opened 2026-09-03
A federated trim still reaches SQLite and MySQL as btrim, which neither has


claudespice:
## Summary

#13794 reported that a federated `trim` reaches DuckDB as `btrim`, and #13821 fixed it **for
DuckDB**. The same defect is still live for **SQLite and MySQL**, which #13821 did not touch.

`trim(x)` in SQL resolves to `DataFusion`'s `btrim` UDF — `trim` is one of its aliases — and the
unparser emits scalar functions under their canonical name. Neither SQLite nor MySQL has a
`btrim`, so the query fails remotely instead of returning a row.

## Reproduction — measured

One `spiced`, five datasets over one CSV, `arrow` acceleration as the local control, on
`trunk` at `968999c5fd` (i.e. with #13821 already in):

```
$ curl -s -XPOST localhost:8090/v1/sql -d "select trim(name) as v, id from lite order by id"
Execution error: Failed to execute query.
ConnectionError Error("no such function: btrim in SELECT btrim(`lite`.`name`) AS `v` ... at offset 7")

$ curl -s -XPOST localhost:8090/v1/sql -d "select trim(name) as v, id from my order by id"
Execution error: Failed to execute query.
Server error: `ERROR 42000 (1305): FUNCTION spice.btrim does not exist'

$ curl -s -XPOST localhost:8090/v1/sql -d "select trim(name) as v, id from arrowds order by id"
[{"v":"alpha","id":1},{"v":"xxbetaxx","id":2},{"v":"gamma","id":3}]
```

`lite` is a SQLite accelerator, `my` a MySQL connector dataset. PostgreSQL is **unaffected** —
it has `btrim` natively, and its results match the local control exactly.

The MySQL **catalog** connector is affected the same way. `catalogconnector/mysql.rs` gained a
deny-list in #13664, but the backend-agnostic one, which does not carry `btrim`:

```
$ curl -s -XPOST localhost:8091/v1/sql -d "select trim(name) as v from mycat.spice.t order by id"
Server error: `ERROR 42000 (1305): FUNCTION spice.btrim does not exist'
$ curl -s -XPOST localhost:8091/v1/sql -d "select trim(name) as v from myds order by id"   # dataset path, same table
[{"v":"alpha"},{"v":"xxbetaxx"}]
```

Affected shapes on both backends: `trim(col)`, `trim(col, chars)`, `btrim(col)`,
`length(trim(col))`, and `WHERE trim(col) = …`.

[...]

Issue #10583, opened 2026-04-29
SQL Function Compatibility Gaps: DuckDB and SQLite Acceleration


ewgenius:
# SQL Function Compatibility Gaps: DuckDB and SQLite Acceleration

## Describe the bug

Many SQL functions documented in the [Spice SQL Reference](https://spiceai.org/docs/reference/sql) work correctly when using Arrow (DataFusion) acceleration but fail with errors when the same query is run against a dataset using **DuckDB** or **SQLite** acceleration. The errors surface as either missing function names in the target engine or incorrect translation of SQL syntax by Spice.

1. **DuckDB**: ~20 functions unsupported or behaviorally inconsistent
2. **SQLite**: ~50+ functions unsupported (SQLite's built-in function set is very limited)

## To Reproduce

1. Configure a dataset with DuckDB or SQLite acceleration in `spicepod.yaml`
2. Run a query using any of the affected functions via `spice sql` or `POST /v1/sql`
3. Observe the error

Arrow:

```console
SELECT percentile_cont(0.5) WITHIN GROUP (ORDER BY trip_distance) FROM taxi_trips_arrow;
+--------------------------------------------------------------------------------------------+
| percentile_cont(Float64(0.5)) WITHIN GROUP [taxi_trips_arrow.trip_distance ASC NULLS LAST] |
|                                           float64                                          |
+--------------------------------------------------------------------------------------------+
| 1.68                                                                                       |
+--------------------------------------------------------------------------------------------+

Time: 0.440653934 seconds. 1 rows.

```

DuckDB Acceleration:

```console
sql> SELECT percentile_cont(0.5) WITHIN GROUP (ORDER BY trip_distance) FROM taxi_trips_duckdb;
Query Error: Failed to execute query.
Parser Error: Wrong number of arguments for PERCENTILE_CONT
```

SQLite Acceleration:

```console
sql> SELECT percentile_cont(0.5) WITHIN GROUP (ORDER BY trip_distance) FROM taxi_trips_sqlite;
Query Error: Failed to execute query.
ConnectionError Error("near "(": syntax error in SELECT percentile_cont(`taxi_trips_sqlite`.`trip_distance`, 0.5) WITHIN GROUP (ORDER BY `taxi_trips_sqlite`.`trip_distance` ASC NULLS LAST) FROM `taxi_trips_sqlite` at offset 78")
```

<details>
<summary>Full test queries — DuckDB</summary>

```sql
-- Aggregate: percentile / approximate
[...]

PR #13852, opened 2026-09-03
fix(duckdb): lower-case the hex digits a federated to_hex gets back (fixes #13818)

### LINKED ISSUE #13818: to_hex pushed down to a DuckDB accelerator returns upper-case hex where local evaluation returns lower-case
## What happens

`to_hex` exists in both engines, so nothing denies it and nothing rewrites it — the call is
pushed into the accelerated DuckDB store verbatim. DuckDB renders the digits in **upper** case
where DataFusion renders them in **lower** case, so the same query over the same rows returns a
different string depending on whether the dataset happens to be accelerated. There is no error
and no warning; the result is simply different.

## Measured

One local CSV, two datasets over it — `t_duckdb` with `acceleration.engine: duckdb`, `t_local`
unaccelerated — on `trunk` at `f8d83e39f0`:

```
$ SELECT to_hex(255) AS a, to_hex(3735928559) AS b, to_hex(-1) AS c, to_hex(0) AS d FROM t_duckdb LIMIT 1
[{"a":"FF","b":"DEADBEEF","c":"FFFFFFFFFFFFFFFF","d":"0"}]

$ SELECT to_hex(255) AS a, to_hex(3735928559) AS b, to_hex(-1) AS c, to_hex(0) AS d FROM t_local LIMIT 1
[{"a":"ff","b":"deadbeef","c":"ffffffffffffffff","d":"0"}]
```

The call really is evaluated remotely — `EXPLAIN` shows it in the SQL sent to DuckDB:

```
VirtualExecutionPlan name=duckdb compute_context=:memory: base_sql=SELECT to_hex(255) FROM "t_duckdb" LIMIT 1
```

`to_hex(0)` agrees, which is why a spot check can miss this: the divergence only appears once a
digit above 9 is present.

## Why this one is worse than the rest of #10583

#10583 catalogues the DuckDB/SQLite function gaps, and almost all of them **fail loudly** —
`Catalog Error: Scalar Function with name … does not exist!`. A sweep of 39 expressions through a
live DuckDB accelerator found exactly one that returns a *different answer* instead of an error,
and this is it. A predicate like `WHERE to_hex(h) = 'deadbeef'` silently matches nothing on an
accelerated dataset and matches on an unaccelerated one; a join or a dedupe keyed on a hex string
silently changes shape when acceleration is turned on.

## Suggested fix

Either rewrite it in the DuckDB unparser dialect (`crates/runtime-datafusion/src/dialect/`) to
`lower(to_hex(...))`, or add `to_hex` to the DuckDB `FunctionSupport` deny-list so it evaluates
locally. The dialect now has a table for exactly this shape —
`duckdb_builtin_scalar_overrides()`, added for `btrim` in #13794 — so the rewrite is a one-entry
change; the deny-list route is `deny_also` in
`crates/runtime-datafusion/src/function_support.rs`.

Refs #10583

claudespice:
## Summary

`to_hex` exists in both engines, so nothing denied the call and nothing rewrote it: it was pushed
into the accelerated `DuckDB` store verbatim and came back with **upper-case** digits where the
kernel produces lower-case ones. The same query over the same rows answered differently depending
on whether the dataset happened to be accelerated — no error, no warning.

Measured on `trunk` at `f8d83e39f0` — one local CSV, two datasets over it (`t_duckdb` accelerated
into `DuckDB`, `t_local` unaccelerated as the control):

```
$ … -d "SELECT id, h, to_hex(h) AS hx FROM t_duckdb ORDER BY id"
[{"id":1,"h":255,"hx":"FF"},{"id":2,"h":3735928559,"hx":"DEADBEEF"},
 {"id":3,"h":0,"hx":"0"},{"id":4,"h":-1,"hx":"FFFFFFFFFFFFFFFF"}]

$ … -d "SELECT id, h, to_hex(h) AS hx FROM t_local ORDER BY id"
[{"id":1,"h":255,"hx":"ff"},{"id":2,"h":3735928559,"hx":"deadbeef"},
 {"id":3,"h":0,"hx":"0"},{"id":4,"h":-1,"hx":"ffffffffffffffff"}]
```

The `h` column itself agrees, so the divergence is the function's rendering and nothing else.
`to_hex(0)` agrees too, which is why a spot check can walk past this: the divergence needs a digit
above 9.

**The shape that makes it data loss rather than cosmetics is a predicate.** Same runtime, same
rows:

```
$ … -d "SELECT id FROM t_duckdb WHERE to_hex(h) = 'deadbeef'"
[]
$ … -d "SELECT id FROM t_local  WHERE to_hex(h) = 'deadbeef'"
[{"id":2}]
```

## Changes

**Rewrite the call in the `DuckDB` unparser dialect to `lower(to_hex(x))`**, joining `btrim` in
`duckdb_builtin_scalar_overrides()` — the table added by #13821 for exactly this shape.

Case is the *only* divergence, which is what makes `lower(..)` an exact fix rather than a closer
approximation. Measured across every integer width `to_hex` accepts, plus the boundary values and
NULL, comparing the candidate rendering evaluated **by `DuckDB`** against the kernel's answer:

| input | `lower(to_hex(x))` in `DuckDB` | `to_hex(x)` in `DataFusion` |
|---|---|---|
| `255` | `ff` | `ff` |
| `3735928559` | `deadbeef` | `deadbeef` |
| `0` | `0` | `0` |
| `-1` (`BIGINT`) | `ffffffffffffffff` | `ffffffffffffffff` |
| `CAST(-1 AS INT)` | `ffffffffffffffff` | `ffffffffffffffff` |
| `CAST(-1 AS SMALLINT)` | `ffffffffffffffff` | `ffffffffffffffff` |
| `NULL` | `null` | `null` |

Both engines widen to 64 bits and emit the same digits in the same order, so lower-casing makes the
two answers identical on every input, not merely closer. `to_hex`'s own signature admits exactly
one argument (`Coercion::new_exact(TypeSignatureClass::Integer)`), so the handler's other arm is
defensive — and it returns an error rather than `Ok(None)`, which would hand the un-lowered
`to_hex` straight back to `DuckDB`.

`DataFusion`'s lower case is not incidental: `to_hex.rs` renders from
`const HEX_CHARS: &[u8; 16] = b"0123456789abcdef"`.

**Two small refactors, both to avoid a second copy of something that already existed.**
`wrap_function` was a private associated function on `DuckDBRegexpFunction`, used only to wrap
`regexp_extract_all` in `len(..)`; it is now the module-level `wrap_in_call`, which the `to_hex`
handler calls rather than carrying its own. And the new integration test had copy-pasted the
`btrim` test's six-line `base_sql` extraction verbatim, so that is now one `pushed_down_sql`
helper both tests call.

## Scope — what else the sweep found, and what it did not

The issue reported one divergence. Rather than take that on trust, a differential sweep of **53**
expressions ran through the same live accelerator, each compared against the unaccelerated control:

- **40 agreed exactly.**
- **8 expressions fail loudly** on `DuckDB`, in two shapes — both the #10583 class, neither able to
  return a wrong row: `Catalog Error: Scalar Function with name … does not exist!` for `sha224`,
  `initcap`, `signum`, `nanvl` and `nvl`, and `Binder Error: No function matches the given name and
  argument types` for `encode(VARCHAR, STRING_LITERAL)` (both the `'hex'` and `'base64'` forms) and
  `octet_length(VARCHAR)`, where `DuckDB`'s own overloads take a `BLOB`/`BIT`.
- **2 apparent divergences were my harness's own, and are withdrawn.** `floor(f)` and `ceil(f)`
  differed only under the sweep's `CAST(… AS VARCHAR)` wrapper, which is itself pushed down; read
  directly, both agree on every row. Reported here because a negative result is part of the
  measurement.
- **3 real silent divergences**, of which this PR fixes one:
  - `to_hex` — **fixed here**. Case, exactly, with an exact rewrite.
  - `concat` over a NULL argument — `DuckDB` ignores the NULL, the kernel answers NULL. Filed as
    #13849 rather than fixed here, because **which side is authoritative is an open
    question**: `DataFusion` documents `concat` as ignoring NULLs and `simplify_concat` drops null
    literals, yet the runtime answers NULL. A rewrite that matched today's observed kernel would
    invert silently if `DataFusion` ever behaved as documented, so that decision does not belong in
    a `to_hex` PR.
  - `sha256` — the kernel returns the 32-byte digest as `Binary`; `DuckDB` returns its 64-character
    hex *text*, which `SchemaCastScanExec` then casts into the `Binary` column. Filed as
    #13850. The candidate fix (`unhex(sha256(x))`) cannot be exercised through `/v1/sql` at
    all — `unhex` is not a `DataFusion` function — so unlike `lower(to_hex(x))` it is unverified,
    and shipping it on inspection alone is exactly what this repo's evidence bar forbids.

So the bug class is swept and its members are each accounted for; the two left open are left open
deliberately, each with its measurement, not deferred for convenience.

**The sibling backend was measured too, and it is the loud kind.** `to_hex` on a `SQLite`
accelerator does not diverge — it fails outright, so no wrong row is ever returned:

```
$ … -d "SELECT id, to_hex(h) AS hx FROM t_sqlite ORDER BY id"
ConnectionError Error("no such function: to_hex in SELECT `t_sqlite`.`id`,
  to_hex(`t_sqlite`.`h`) AS `hx` FROM `t_sqlite` … at offset 24")
```

That is #10583's own `❌ to_hex(n)` row, and it is out of scope here for the reason #13840 gives:
`SQLite` has no dialect seam to rewrite through, so its lever is the `FunctionSupport` deny-list,
not this table. Measurement and a caution about #10583's suggested `hex(n)` equivalent are
[recorded on that issue](https://github.com/spiceai/spiceai/issues/10583) rather than fixed here.
`PostgreSQL` and `MySQL` were **not** measured for this function — no server on the rig — so this
PR makes no claim about them either way.

## Test plan

**Reproduced on `trunk` first** — the transcripts above are from that run; the same commands on the
branch are below, on the same rig.

```
$ SELECT id, h, to_hex(h) AS hx FROM t_duckdb ORDER BY id
[{"id":1,"h":255,"hx":"ff"},{"id":2,"h":3735928559,"hx":"deadbeef"},
 {"id":3,"h":0,"hx":"0"},{"id":4,"h":-1,"hx":"ffffffffffffffff"}]

$ SELECT id, h, to_hex(h) AS hx FROM t_local ORDER BY id
[{"id":1,"h":255,"hx":"ff"},{"id":2,"h":3735928559,"hx":"deadbeef"},
 {"id":3,"h":0,"hx":"0"},{"id":4,"h":-1,"hx":"ffffffffffffffff"}]

$ SELECT id FROM t_duckdb WHERE to_hex(h) = 'deadbeef'   ->  [{"id":2}]
$ SELECT id FROM t_local  WHERE to_hex(h) = 'deadbeef'   ->  [{"id":2}]
```

The two sides now agree on every row, and the predicate selects the row on both. `EXPLAIN` shows
the rewrite in the SQL that actually left the process — and that the call is still **pushed down**,
not pulled back above the scan:

```
VirtualExecutionPlan name=duckdb compute_context=:memory: \
  base_sql=SELECT lower(to_hex("t_duckdb"."h")) AS "hx" FROM "t_duckdb"
```

**Tests added, each neutered to confirm it is load-bearing:**

- `duckdb_accelerated_to_hex_agrees_with_local`
  (`crates/runtime/tests/acceleration/duckdb_builtin_pushdown.rs`) — the wired path: one CSV, a
  `DuckDB`-accelerated dataset and an unaccelerated twin, five rows (`255`, `3735928559`, `0`,
  `-1`, NULL). Asserts the pushed-down SQL from `base_sql` really carries `lower(to_hex(`, pins the
  accelerated rows, requires them to equal the local control, and repeats both for the
  `WHERE to_hex(h) = 'deadbeef'` predicate. This is the test that catches the class: a unit test
  over the handler cannot tell a rendering that parses from one `DuckDB` answers the same way.
- `to_hex_unparses_to_a_lowercased_duckdb_to_hex`, `duckdb_dialect_installs_the_to_hex_override`,
  `to_hex_with_an_impossible_arity_is_an_error_not_a_passthrough`
  (`crates/runtime-datafusion/src/dialect/duckdb.rs`) — the rendering, the fact that the handler is
  actually installed on the dialect (a handler written but never registered fails here rather than
  in a federated query), and the defensive arity arm.

**Neuter 1 — drop the `lower(..)` wrap, keep the handler registered (i.e. make the rewrite a bare
rename).** This is the neuter that matters, because a bare rename is the plausible wrong fix.

The integration test catches it, and its message carries the plan that proves *why*:

```
thread '…::duckdb_accelerated_to_hex_agrees_with_local' panicked at
  crates/runtime/tests/acceleration/duckdb_builtin_pushdown.rs:312:13:
the hex rendering must be pushed down to DuckDB inside a `lower(..)`; plan was:
  … VirtualExecutionPlan name=duckdb base_sql=SELECT to_hex("accelerated"."h") FROM "accelerated"
test result: FAILED. 2 passed; 1 failed; 373 filtered out
```

Note the two `btrim` tests still pass — the neuter was surgical, so the failure is attributable to
this change and not to a broken dialect.

The unit tests catch it too:

```
test dialect::duckdb::tests::to_hex_unparses_to_a_lowercased_duckdb_to_hex ... FAILED
  left: "to_hex(\"t\".\"h\")"
 right: "lower(to_hex(\"t\".\"h\"))"
test dialect::duckdb::tests::duckdb_dialect_installs_the_to_hex_override ... FAILED
  left: "to_hex(255)"
 right: "lower(to_hex(255))"
test result: FAILED. 11 passed; 2 failed
```

**Neuter 2 — un-register the handler**, leaving the (correct) rendering function unreachable. Only
the `installs` test can see this, which is the reason that test exists: a handler written but never
wired into the dialect is invisible to a test that calls the handler directly.

```
test dialect::duckdb::tests::duckdb_dialect_installs_the_to_hex_override ... FAILED
  left: "to_hex(255)"
 right: "lower(to_hex(255))"
test result: FAILED. 12 passed; 1 failed
```

Restored — unit and integration both green:

```
test result: ok. 55 passed; 0 failed; 0 ignored; 0 measured; 184 filtered out   # runtime-datafusion --lib dialect
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 373 filtered out    # runtime --test integration duckdb_builtin_pushdown
```

Full gate, run locally via `make signoff` (= `make lint` + `make build-cli` + `make nextest`):

```
make lint
make nextest
```

## Performance impact

One extra `lower(..)` call per row, inside `DuckDB`, on the projection of a query that already
called `to_hex`. It does not change which rows are scanned or how: `to_hex(h)` was never
index-usable on `h` in the first place, so `WHERE lower(to_hex(h)) = …` costs the same table scan
as `WHERE to_hex(h) = …` did. Nothing is moved out of the remote engine — the alternative fix
(deny-listing `to_hex`) *would* have done that, pulling the projection back above the federated
scan, which is one reason the dialect rewrite is preferable now that it is measured to be exact.

## Review gate

- Adversarial review: **skipped** — `codex` exited 1: "Your workspace is out of credits. Ask your workspace owner to refill in order to continue."; `grok` not installed
- `/security-review`: no findings — the handler builds a `sqlparser` AST rather than concatenating SQL, both identifiers it introduces are `const` literals, the one argument goes through the same `unparser.expr_to_sql` path every existing handler uses, and the single new error message interpolates only `args.len()`
- `/simplify`: applied — collapsed the handler's early return into a `map`, and extracted the `base_sql` extraction my test had copy-pasted from the `btrim` test into one shared `pushed_down_sql` helper; light checks and both test suites re-run green afterwards

Fixes #13818

## Checklist

- [x] Reproduced the reported failure on `trunk` before writing the fix, and re-ran the same commands on the branch
- [x] The candidate rendering was evaluated **by `DuckDB`** against the kernel across every input `to_hex` accepts, so the rewrite is measured to be exact rather than argued to be
- [x] Regression tests added at both levels, each neutered to confirm it fails without the fix
- [x] The bug class was swept, not just the reported instance; the two members not fixed here are filed with their measurements (#13849, #13850)
- [ ] `make signoff` green and `Attestation` re-run (runs after the push that creates this PR)

### DIFF
--- a/crates/runtime-datafusion/src/dialect/duckdb.rs
+++ b/crates/runtime-datafusion/src/dialect/duckdb.rs
@@ -160,12 +160,8 @@ pub(crate) fn to_hex_to_lowercase_hex(
     args: &[Expr],
 ) -> Result<Option<ast::Expr>, DataFusionError> {
     match args {
-        [_] => {
-            let Some(hex) = renamed_fn_to_sql(unparser, args, TO_HEX_NAME)? else {
-                return Ok(None);
-            };
-            Ok(Some(wrap_in_call(hex, LOWER_NAME)))
-        }
+        [_] => Ok(renamed_fn_to_sql(unparser, args, TO_HEX_NAME)?
+            .map(|hex| wrap_in_call(hex, LOWER_NAME))),
         // `to_hex` is a single-argument function, so the planner cannot build
         // this. Fail rather than fall through to `Ok(None)`, which would put
         // the un-lowered `to_hex` back into the DuckDB SQL.
--- a/crates/runtime-datafusion/src/dialect/duckdb.rs
+++ b/crates/runtime-datafusion/src/dialect/duckdb.rs
@@ -35,6 +35,14 @@ pub(crate) const TRIM_NAME: &str = "trim";
 /// [`btrim_to_trim`] for why it has to be passed to `DuckDB` explicitly.
 const ASCII_SPACE: &str = " ";
 
+/// The name both engines give the integer-to-hex function. They disagree only
+/// on the case of the digits — see [`to_hex_to_lowercase_hex`].
+const TO_HEX_NAME: &str = "to_hex";
+
+/// `DuckDB`'s lower-casing function, applied over [`TO_HEX_NAME`] to match
+/// `DataFusion`'s lower-case hex digits.
+const LOWER_NAME: &str = "lower";
+
 /// Renders `args` as a call to `duckdb_fn`, in the order given.
 ///
 /// The caller is responsible for having already put `args` into the shape
@@ -107,6 +115,62 @@ pub(crate) fn btrim_to_trim(
         // which would put `btrim` back into the DuckDB SQL.
         _ => Err(DataFusionError::Plan(format!(
             "btrim takes one or two arguments, got {}; cannot render it as DuckDB SQL.",
+            args.len()
+        ))),
+    }
+}
+
+/// Renders `inner` as the sole argument of a call to `function_name`.
+fn wrap_in_call(inner: ast::Expr, function_name: &str) -> ast::Expr {
+    ast::Expr::Function(Function {
+        name: ObjectName(vec![ast::ObjectNamePart::Identifier(Ident::new(
+            function_name,
+        ))]),
+        args: ast::FunctionArguments::List(ast::FunctionArgumentList {
+            duplicate_treatment: None,
+            args: vec![FunctionArg::Unnamed(FunctionArgExpr::Expr(inner))],
+            clauses: vec![],
+        }),
+        filter: None,
+        null_treatment: None,
+        over: None,
+        within_group: vec![],
+        parameters: ast::FunctionArguments::None,
+        uses_odbc_syntax: false,
+    })
+}
+
+/// Lower-cases `DuckDB`'s `to_hex`, which upper-cases the digits `DataFusion`
+/// renders in lower case.
+///
+/// Both engines have a `to_hex`, so nothing denied the call and nothing
+/// rewrote it: it was pushed into the accelerated store verbatim and came back
+/// with different characters. `to_hex(255)` is `ff` from the kernel and `FF`
+/// from `DuckDB`, so the same query over the same rows answered differently
+/// depending on whether the dataset happened to be accelerated, with no error
+/// and no warning (issue #13818). A predicate such as
+/// `WHERE to_hex(h) = 'deadbeef'` simply matched nothing.
+///
+/// Case is the *only* divergence: measured across `Int16`, `Int32` and `Int64`
+/// inputs, negative values, zero and NULL, both engines widen to 64 bits and
+/// produce the same digits in the same order, so wrapping the call in
+/// [`LOWER_NAME`] makes the two answers identical rather than merely closer.
+pub(crate) fn to_hex_to_lowercase_hex(
+    unparser: &datafusion::sql::unparser::Unparser,
+    args: &[Expr],
+) -> Result<Option<ast::Expr>, DataFusionError> {
+    match args {
+        [_] => {
+            let Some(hex) = renamed_fn_to_sql(unparser, args, TO_HEX_NAME)? else {
+                return Ok(None);
+            };
+            Ok(Some(wrap_in_call(hex, LOWER_NAME)))
+        }
+        // `to_hex` is a single-argument function, so the planner cannot build
+        // this. Fail rather than fall through to `Ok(None)`, which would put
+        // the un-lowered `to_hex` back into the DuckDB SQL.
+        _ => Err(DataFusionError::Plan(format!(
+            "to_hex takes one argument, got {}; cannot render it as DuckDB SQL.",
             args.len()
         ))),
     }
@@ -390,25 +454,6 @@ impl DuckDBRegexpFunction {
         Ok(())
     }
 
-    fn wrap_function(ast_fn: ast::Expr, function_name: &str) -> ast::Expr {
-        ast::Expr::Function(Function {
-            name: ObjectName(vec![ast::ObjectNamePart::Identifier(Ident::new(
-                function_name,
-            ))]),
-            args: ast::FunctionArguments::List(ast::FunctionArgumentList {
-                duplicate_treatment: None,
-                args: vec![FunctionArg::Unnamed(FunctionArgExpr::Expr(ast_fn))],
-                clauses: vec![],
-            }),
-            filter: None,
-            null_treatment: None,
-            over: None,
-            within_group: vec![],
-            parameters: ast::FunctionArguments::None,
-            uses_odbc_syntax: false,
-        })
-    }
-
     fn postprocess_function(&self, mut ast_fn: ast::Expr) -> ast::Expr {
         match self {
             DuckDBRegexpFunction::Match => {
@@ -424,7 +469,7 @@ impl DuckDBRegexpFunction {
             }
             DuckDBRegexpFunction::Count => {
                 // Wrap the extract array in a ``len()``
-                ast_fn = Self::wrap_function(ast_fn, "len");
+                ast_fn = wrap_in_call(ast_fn, "len");
             }
             _ => {}
         }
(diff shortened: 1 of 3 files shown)


claudespice:
@copilot review

lukekim (approved):

Issue #60, opened 2021-08-24
e2e tests that checks basic output


lukekim:
Test the actual training result/output.

PR #13854, opened 2026-09-03
test(pushdown): guard the date literal's unit and width against a lost fork patch (fixes #13476)

### LINKED ISSUE #13476: Date32 literals overflow i32 past 2038 and Date64 literals are rendered as if milliseconds were days
### What happens

The `Date32` and `Date64` literal arms of the filter unparser both multiply the value by
`SECONDS_IN_DAY`, and neither is right.

```rust
// core/src/sql/sql_provider_datafusion/expr.rs (spiceai/datafusion-table-providers, branch `spiceai-54`)
pub const SECONDS_IN_DAY: i32 = 86_400;

ScalarValue::Date32(Some(value)) => match engine {
    Some(Engine::SQLite) => Ok(format!("date({}, 'unixepoch')", value * SECONDS_IN_DAY)),
    _ => Ok(format!("TO_TIMESTAMP({})", value * SECONDS_IN_DAY)),
},
ScalarValue::Date64(Some(value)) => match engine {
    Some(Engine::SQLite) => Ok(format!("date({}, 'unixepoch')", value * i64::from(SECONDS_IN_DAY))),
    _ => Ok(format!("TO_TIMESTAMP({})", value * i64::from(SECONDS_IN_DAY))),
},
```

**`Date32` overflows `i32` for any date after 2038-01-19.** `value` is an `i32` count of days and
`SECONDS_IN_DAY` is an `i32`, so the product is computed in `i32`. `i32::MAX / 86_400` is 24 855
days — 2038-01-19. One day further overflows: a debug build panics, and a release build wraps to a
literal naming an unrelated instant, so the comparison built from it selects a row set the query
never asked for. The mirror bound is 1901-12-13 below the epoch.

**`Date64` is milliseconds, not days.** Arrow defines `Date64` as *"the elapsed time since UNIX
epoch in milliseconds"* (a whole number of days expressed in milliseconds), so the correct epoch
second is `value / 1000`. Multiplying by 86 400 instead treats each millisecond as a day.

Measured on `spiceai-54` at `d9da7cd`:

```
Date32(24855 = 2038-01-19) => Ok("TO_TIMESTAMP(2147472000)")
Date32(24856 = 2038-01-20) => panicked: attempt to multiply with overflow
Date64(86400000 = 1970-01-02) => Ok("TO_TIMESTAMP(7464960000000)")   // correct is 86400
```

`TO_TIMESTAMP(7464960000000)` is somewhere past the year 238 000, so a `Date64` filter pushed down
to any of these engines compares against an instant no row can hold.

### Scope

Both arms are shared by every engine — the `_ =>` catch-all covers DuckDB, Postgres, MySQL, ODBC
and Spark, and the SQLite arm has the same two defects in its own `date(..)` rendering. A fix has
to settle the right rendering per engine rather than patching one of them, which is why this is
filed separately rather than folded into the microsecond work.

### Why this is filed separately

Found while fixing #13432, which is about `f64` precision in the *timestamp* literal arms of the
same `match`. These are unit and width errors in the *date* arms — a different defect, a different
fix, and one that touches engines #13432 does not.

### Acceptance

- A `Date32` literal for a date after 2038-01-19 renders the instant it names, with a test pinning
  a day past 24 855 that fails on the current `i32` arithmetic.
- A `Date64` literal renders `value / 1000` as its epoch second, with a test pinning a known day.
- Both covered for the SQLite arm as well as the catch-all.

claudespice:
## Summary

`datafusion-table-providers` fork PR #60 repaired two defects in the date-literal arms of the shared
filter renderer, `sql::sql_provider_datafusion::expr`. A `Date32` day count was scaled to seconds in
`i32`, which overflows for every day past `i32::MAX / 86_400` — 24 855, or 2038-01-19 — and
symmetrically below 1901-12-13; a `Date64` count of *milliseconds* was scaled **up** by the length of
a day rather than down to the second. Both arms are reached by every engine, and both move which rows
a pushed-down date filter selects without failing.

**The pin already carries that fix.** `1512f78f` is 3 commits ahead of #60's merge commit and 0
behind it, so the failure #13476 reports no longer reproduces on `trunk` — the delivery condition the
issue was left open for is met.

What was still missing is the other half, and `docs/dev/fork_patches.md` says so itself: the row for
this patch carried **GAP** in its Guard column. `spiceai-54` is re-cut per DataFusion major, and a
patch nobody guards can be dropped in a conflict resolution and revert the runtime to wrong rows with
nothing here failing. This PR closes that gap, which is what discharges the issue rather than merely
delivering it.

## Changes

- **`crates/search/src/index/duckdb/sql.rs`** — four guards, placed where this repository actually
  renders through that function (the DuckDB search index's filter push-down), alongside the existing
  guards for the sibling timestamp-literal patch (fork PR #57):
  - `a_date32_literal_outside_the_i32_second_range_names_the_day_it_holds` — 2038-01-20, 2100-01-01
    and 1901-12-13 each pinned to the epoch second it names.
  - `a_date64_literal_is_read_as_milliseconds_not_as_days` — 1970-01-02 and 2038-01-20, the second
    naming the same day as the `Date32` guard so the two variants have to agree.
  - `a_date32_literal_inside_the_i32_second_range_renders_unchanged` — the control. Without it a
    renderer that declined every date, or shifted every date by a constant, would satisfy the two
    above, and over-refusal is silent: the filter simply stops being pushed down.
  - `the_sqlite_arm_renders_a_date_from_the_same_count` — the scaling the patch fixes is shared by
    every engine arm and only the call it is formatted into differs, so a revert fails the three
    above whichever engine renders. This pins the sibling `date(.., 'unixepoch')` arm directly,
    because DuckDB is the only engine this repository renders through here.
- **`docs/dev/fork_patches.md`** — the row's Guard cell names those tests and says why the DuckDB
  site covers both engine arms; the GAP count drops 37 → 36 and the entry leaves the prioritised
  *Open gaps* list.

No production code changes.

## Evidence

**1. The pin contains the fix.**

```
$ gh api repos/spiceai/datafusion-table-providers/compare/d03c86c3e...1512f78ff \
    --jq '{status:.status, ahead:.ahead_by, behind:.behind_by}'
{"status":"ahead","ahead":3,"behind":0}
```

**2. What the two renderings actually select, against a real DuckDB.**

A throwaway test (evidence only — not part of this diff) opened an in-memory DuckDB, created
`t(dt DATE)` holding `1901-12-14`, `2024-01-01` and `2038-01-20`, and ran the production rendering
shape `TO_TIMESTAMP(EPOCH_MS("dt") / 1000) > <literal>` for each version of the literal:

```
EVIDENCE | Date32 2038-01-20, fixed (trunk)               | TO_TIMESTAMP(2147558400)   -> []
EVIDENCE | Date32 2038-01-20, reverted (i32 wrap)         | TO_TIMESTAMP(-2147408896)  -> ["2024-01-01", "2038-01-20"]
EVIDENCE | Date64 1970-01-02, fixed (trunk)               | TO_TIMESTAMP(86400)        -> ["2024-01-01", "2038-01-20"]
EVIDENCE | Date64 1970-01-02, reverted (ms read as days)  | TO_TIMESTAMP(7464960000000) -> []
```

`dt > 2038-01-20` correctly returns nothing; the wrapped `i32` literal names 1901-12-14 17:31:44
instead and returns **two rows the query excluded**. `dt > 1970-01-02` correctly returns two rows;
reading the milliseconds as days names an instant past the year 238 000 and returns **none**. Wrong
in both directions, and DuckDB accepted both statements — nothing announces it.

**3. The guards fail when the patch is missing and pass when it is present — same command, one
variable.**

The workspace's `datafusion-table-providers` dependency was pointed at a local checkout of the pinned
revision `1512f78f` for both runs, so the *only* difference between them is the two `match` arms,
restored from the fork's own pre-#60 commit `642e12119`.

With the pre-#60 arms:

```
$ cargo test -p search --features duckdb --lib index::duckdb::sql
thread '…::a_date32_literal_outside_the_i32_second_range_names_the_day_it_holds' panicked at
  …/sql_provider_datafusion/expr.rs:268:53: attempt to multiply with overflow
thread '…::a_date64_literal_is_read_as_milliseconds_not_as_days' panicked at
  crates/search/src/index/duckdb/sql.rs:728:13: 1970-01-02 must render as the second its
  milliseconds name: SELECT id FROM docs WHERE body_embedding IS NOT NULL AND
  TO_TIMESTAMP(EPOCH_MS("dt") / 1000) > TO_TIMESTAMP(7464960000000) ORDER BY …
thread '…::the_sqlite_arm_renders_a_date_from_the_same_count' panicked at
  …/sql_provider_datafusion/expr.rs:266:57: attempt to multiply with overflow

failures:
    index::duckdb::sql::tests::a_date32_literal_outside_the_i32_second_range_names_the_day_it_holds
    index::duckdb::sql::tests::a_date64_literal_is_read_as_milliseconds_not_as_days
    index::duckdb::sql::tests::the_sqlite_arm_renders_a_date_from_the_same_count

test result: FAILED. 17 passed; 3 failed; 0 ignored
```

The control passed in that same run, and so did every pre-existing test in the file — the revert
reaches exactly the three guards it should.

Restoring the two arms, nothing else changed:

```
$ cargo test -p search --features duckdb --lib index::duckdb::sql
test result: ok. 20 passed; 0 failed; 0 ignored
```

(20 there, 19 now: the throwaway evidence test above was still present for those runs and has since
been removed.)

**4. Against the real git pin, no local override:**

```
$ cargo test -p search --features duckdb --lib index::duckdb::sql
test index::duckdb::sql::tests::a_date32_literal_inside_the_i32_second_range_renders_unchanged ... ok
test index::duckdb::sql::tests::a_date32_literal_outside_the_i32_second_range_names_the_day_it_holds ... ok
test index::duckdb::sql::tests::a_date64_literal_is_read_as_milliseconds_not_as_days ... ok
test index::duckdb::sql::tests::the_sqlite_arm_renders_a_date_from_the_same_count ... ok
test result: ok. 19 passed; 0 failed; 0 ignored
```

```
$ python3 scripts/check_fork_patches.py
fork-patch ledger: 31 pinned forks, all recorded
```

## Test plan
- [x] Added regression test for a `Date32` literal past 2038-01-19 and a `Date64` literal read as days
- [x] Added edge case tests for the pre-epoch `i32` bound (1901-12-13), a far-future date (2100-01-01), an in-range control (2024-01-01), and the sibling SQLite arm
- [x] `make test` passes
- [ ] <Any query plan snapshot tests if applicable>

## Review gate

- Adversarial review: **skipped** — `codex` exited 1: "Your workspace is out of credits."; `grok` not installed
- `/security-review`: no findings — the diff is test-only code plus documentation, with no production path, no new input handling, and no new dependency
- `/simplify`: applied — the SQLite arm's two near-identical blocks became one table-driven loop, matching the rest of the file; light checks re-run green (clippy 0 warnings, 19/19 tests pass)

## Checklist
- [x] The pin is confirmed to carry fork PR #60 before claiming the issue is delivered
- [x] The guard is demonstrated to fail without the patch and pass with it, same command
- [x] `scripts/check_fork_patches.py` passes with the new GAP count

Fixes #13476

> Note for whoever merges second: #13847 also edits the `Open gaps` count line (37 → 39). That one
> line will conflict; both edits are independent and the resolution is arithmetic.


claudespice:
@copilot review

lukekim (dismissed):

claudespice:
Resolved the conflict with `trunk` (`616791c9ca` → `df943e87b4`). `MERGEABLE/BLOCKED` now; the only conflicting file was `docs/dev/fork_patches.md`, and `crates/search/src/index/duckdb/sql.rs` auto-merged.

**Both sides changed the same GAP count and the same numbered list, for independent reasons**, so the resolution keeps both rather than picking a side:

- `trunk` (#13847) **added** a gap — the `datafusion` bloom-filter replacement-reader row — taking the count 38 → **39**, and moved the `arrow-rs` `PushBuffers` entry to the end of the hangs list, rewording it to drop the now-stale "runtime #13847 carries the scan-path guards" narration.
- This branch **closed** a gap — the date-literal row now names its guards — taking the count 38 → **37**.

Net is **38**, and the prioritised list renumbers to 27 entries. Trunk's bloom-filter row and its reworded `PushBuffers` row are both preserved (they renumber 13 → 12 and 20 → 19); nothing from either side is dropped.

`scripts/check_fork_patches.py` is the authority here, and it rejects **both** sides' own numbers:

```
$ python3 scripts/check_fork_patches.py     # count = 37 (this branch's number)
  - docs/dev/fork_patches.md: 38 table row(s) are marked **GAP** but the `Open gaps`
    section accounts for 37. …
rc=1

$ python3 scripts/check_fork_patches.py     # count = 39 (trunk's number)
  - docs/dev/fork_patches.md: 38 table row(s) are marked **GAP** but the `Open gaps`
    section accounts for 39. …
rc=1

$ python3 scripts/check_fork_patches.py     # count = 38 (resolved)
fork-patch ledger: 31 pinned forks, all recorded
rc=0
```

So a resolution that took either side verbatim would have failed CI — 38 is the only value that passes, independently confirmed by `grep -cE '^\|.*\*\*GAP\*\*.*\|$'` = 38.

Scope check: comparing the resolved tree against Git's own automatic merge result (`git merge-tree --write-tree`), the **only** differing path is `docs/dev/fork_patches.md` — `crates/search/src/index/duckdb/sql.rs` is byte-identical (`88027b233f36`), and no fork pin moved in `Cargo.lock`. The numbered list is contiguous 1–27.

Sign-off: the previous `signoff=success` bound `616791c9ca`, and a hand-resolved merge does not inherit it, so it was re-dispatched remotely — run [33876889468](https://github.com/spiceai/spiceai/actions/runs/33876889468). Remote rather than local because this work volume is at 94% (110 GiB free, against a cold `target/` of 235–337 GiB here) with another session's build live; a local run would have risked ENOSPC-killing it.

lukekim (dismissed):

claudespice:
## Predicted: this PR will conflict with #13864 in the queue, in `docs/dev/fork_patches.md`

This PR is at queue position 7 and currently reads clean. It is **not** clean against the batch
it is being built onto, and #13864 (position 1) is the reason. Recording the prediction now with
the resolution, so the dequeue isn't re-diagnosed from scratch.

**Reproduction.** Simulating the queue's own batch chain off `trunk` (`5b26d47218`) with
`git merge-tree --write-tree`, applying each entry in queue order:

```
ok  #13864
ok  #13868
ok  #13284
ok  #13566
ok  #13786
ok  #13823
### CONFLICT adding #13854 (rc=1)
CONFLICT (content): Merge conflict in docs/dev/fork_patches.md
```

Reduced to the pair, the collision is exactly `#13854` × `#13864`:

- each is **clean against `trunk` alone** (`merge-tree` rc=0 for both);
- `#13854` merged onto `trunk + #13864` → **rc=1**, conflict in `docs/dev/fork_patches.md`.

They are the only two entries in the queue that touch that file.

**Why they collide.** Both edit the same guard table, on adjacent lines:

- **#13864** *adds* a row (`DuckDBTable` carries its index list onto `DuckSqlExec`, fork PR #69)
  and rewords the paragraph above the table from "Two patches here have rows…" to
  "The patches below have rows…".
- **#13854** *fills in* the *existing* fork-PR-#60 row, replacing its `**GAP**` guard with named
  tests, and therefore also edits the **`39 rows above are marked GAP`** sentence to `38` and
  renumbers the "Open gaps" list (old items 10–12 become 9–11).

**Resolution when it dequeues — take both sides, and mind the count.** The two changes are
additive to *different* rows, so this is a mechanical resolution rather than a semantic one:

1. keep #13864's new fork-PR-#69 row **and** its reworded lead-in paragraph;
2. keep #13854's filled-in fork-PR-#60 guard, the `39 → 38` edit, and the renumbering.

The GAP count stays **38**: #13854 closes one gap (39 → 38), and #13864's new row ships a real
guard (`intermediate_index_cte.rs::tests::a_tables_indexes_reach_the_rule_through_the_exec_node`),
so it adds a row without adding a gap.

That arithmetic is the part worth double-checking — `scripts/check_fork_patches.py` fails the
build when the count and the prose disagree, and it accepts **only** the exact number. A
resolution that keeps both rows but leaves the sentence at `39`, or that re-adds the fork-PR-#60
gap to the numbered list, will fail CI in a way that looks unrelated to either PR.

No action taken on this branch: it is queued, approved and signed off, and merging base in now
would dismiss the approval, invalidate the sign-off and drop it from the queue for a conflict that
has not materialized yet. If #13864 merges first, this one needs the resolution above; if #13864
leaves the queue first, this PR merges as-is.

claudespice:
### Resolved the `trunk` conflict — and it had a trap worth naming

This PR was sitting at merge-queue position 2 in state `UNMERGEABLE` with `headCommit: null`, meaning the queue never built it at all. The cause was a real conflict with `trunk`, confined to a single file:

```
$ git merge-tree --write-tree --name-only $(git rev-parse origin/trunk) df943e87b4
docs/dev/fork_patches.md
CONFLICT (content): Merge conflict in docs/dev/fork_patches.md
```

Both sides had touched the *same* fork-patch row:

- **this branch** fills in the `**GAP**` on the `fork PR #60` date-literal row with the guards it adds — that is the entire point of the PR;
- **trunk** (via #13864) added a **new, adjacent** row for `DuckDBTable` carrying its index list onto `DuckSqlExec` (`fork PR #69`).

Resolving to either side wholesale loses one of them. I kept **both**: this branch's guard-filled `#60` row, plus trunk's new `#69` row. Dequeued (the queue blocks pushes to a queued branch), pushed `fed46c33c6`; the PR went `CONFLICTING`/`DIRTY` → `MERGEABLE`/`BLOCKED`.

**The trap:** the ledger's self-consistency checker does *not* catch the row being dropped. Two negative controls against `scripts/check_fork_patches.py` on the resolved tree:

| tree | `rc` |
|---|---|
| resolved (both rows kept) | `0` |
| GAP count altered `38`→`37` | `1` — caught |
| **trunk's `fork PR #69` row deleted** | **`0` — silently passes** |

So the lazy resolution here would have quietly deleted a fork-patch ledger row that trunk landed hours earlier, and `make lint-rust` would have stayed green over it. Per `docs/dev/fork_patches.md`'s own premise — a Spice patch not deliberately carried across is lost *silently* — that row going missing is exactly the failure the ledger exists to prevent. Worth a follow-up to make the checker diff rows against the pinned forks rather than only reconciling the GAP tally; I have not filed that yet.

The `39 → 38` GAP count and the renumbering of the `Open gaps` list are this branch's own pre-existing changes, untouched by the merge — the merge added exactly 4 lines and removed 3.

Sign-off: the merge was resolved by hand, so it does not inherit the sign-off that was green on `df943e87b4`; re-dispatched as run `33949378367`. The push also dismissed the standing approval — the delta since it is a base merge plus a docs-ledger splice, no logic change.

lukekim (approved):

PR #615, opened 2024-01-25
Spice AI datasource (Overwrite only)

mitchdevenport:
As part of this PR, I have reimplemented a basic `spice-rs` client that appears to be working reliably and consistently in my testing.

One caveat to this PR is that the way to manage and create schemas (i.e. `eth.recent_blocks`) is not immediately clear. I was spending too much time trying to figure this out and was finding conflicting info, so for the time being I've punted on this and implemented a simple solution which is to always rename tables to replace the periods with underscores (i.e. `eth_recent_blocks`). If turns out that schemas are not possible, then we can always implement this more elegantly as a permanent solution like we do in Ducker.

PR #13864, opened 2026-09-04
fix(duckdb): Restore intermediate index materialization optimization

### LINKED ISSUE #13863: DuckDB: intermediate index materialization (CTE) rule never applies
The DuckDB intermediate index materialization rule (`DuckDBIntermediateIndexMaterializationOptimizer`) never rewrites a query, so an indexed column is never used to narrow the scan.

### Cause

`DuckDBTable::create_physical_plan` builds the exec node without passing the table's index list, and `DuckSqlExec::new` initialises the field empty:

```rust
// core/src/duckdb/sql_table.rs
DuckSqlExec::new(projection, schema, pool, sql, table_functions, dialect)   // no indexes
...
indexes: Vec::new(),
```

`DuckSqlExec::indexes()` is therefore always empty, and the rule returns early before any rewrite:

```rust
// intermediate_index_cte.rs
if duck_exec.indexes().is_empty() {
    return Ok(plan);
}
```

The rule's unit tests call `rewrite_statement()` directly with a hand-built index list, so they pass while the runtime path never populates one.

### Reproduce

A `duckdb` dataset with an index and `query_federation: disabled`:

```yaml
acceleration:
  engine: duckdb
  mode: file
  primary_key: '(request_query,request_path)'
  indexes:
    request_query: enabled
  params:
    query_federation: disabled
```

`EXPLAIN SELECT content FROM registry WHERE request_path = ? AND request_query = ?` shows no CTE:

```
DuckSqlExec sql= SELECT "content" FROM "registry"
  WHERE ("request_path" = '...') AND ("request_query" = '...')
```

The index exists in the DuckDB file. No query shape produces a CTE, including shapes that satisfy every condition in `build_cte`.

### Impact

Point lookups on a table of ~139k rows, index on `request_query`, 30 distinct keys:

| Query | p50 | p95 |
|---|---|---|
| As pushed today | 6.99 ms | 8.20 ms | 
| CTE form the rule would build | 0.81 ms | 1.01 ms |

The pushed query plans as `SEQ_SCAN`. The gap widens with row count.

### Expected

The index list configured on the table should reach `DuckSqlExec`, so the rule can apply.

Version: v2.2.2

sgrebnov:
Moves the `datafusion-table-providers` pin to `spiceai-54` at `7752b0e`, which carries spiceai/datafusion-table-providers#69.

That fix restores the table's index list on `DuckSqlExec`, which the DuckDB intermediate index materialization rule reads before rewriting a scan into a materialized CTE bound to an index. The list had been empty since #615, so the rule was returning early and never firing — an indexed column was never used to narrow the scan.

Fixes #13863


sgrebnov:
### Follow-up: add an integration test for the CTE optimization

This PR restores the wiring, but nothing in this repository asserts the optimization actually applies. That gap is why the regression went unnoticed for ~4 months.

**What broke and why it was silent**

`DuckDBTable::create_physical_plan` stopped passing the table's index list to `DuckSqlExec::new`, which initialised the field to an empty vec. `DuckSqlExec::indexes()` is what `DuckDBIntermediateIndexMaterializationOptimizer` reads before rewriting a scan into a materialized CTE bound to an index, so with the list always empty the rule returned early and never fired.

The field was kept and defaulted rather than removed, so it compiled with no warning. The rule's own unit tests call `rewrite_statement()` directly with a hand-built index list, so they passed throughout.

**Coverage now, and what is missing**

spiceai/datafusion-table-providers#69 adds a unit test asserting the indexes survive `scan()`. That pins the wiring in the fork, but a future refactor on either side of the boundary can still disable the rule end to end without failing anything here.

**Proposed test**

A duckdb-accelerated dataset with an index and `query_federation: disabled`:

```yaml
acceleration:
  engine: duckdb
  indexes:
    request_query: enabled
  params:
    query_federation: disabled
```

`EXPLAIN` a query filtering on the indexed column plus at least one other column, then assert the pushed `DuckSqlExec` SQL contains `_intermediate_materialize`. A single-predicate query on the indexed column alone is the "already optimal" case and correctly produces no CTE, so it should not be used as the positive case.

The test must fail when the index list is empty — that is the condition this PR fixes, and the property worth pinning.

**Measured impact**, point lookups on a ~139k row table, index on one of two predicate columns:

| Query | p50 |
|---|---|
| As planned before this fix | 6.99 ms |
| CTE form the rule builds | 0.81 ms |

krinart (approved):

PR #13903, opened 2026-09-05
fix(duckdb): pin a connector's DuckDB session to UTC so a dataset's schema does not carry the host timezone (fixes #13899)

### LINKED ISSUE #13899: A duckdb connector dataset's schema carries the host timezone, so the same query returns different rows on different hosts
## Describe the bug

A `duckdb` connector dataset's Arrow schema carries **the host's timezone**, so the same
query over the same database file returns **different rows on different machines**. Nothing
is logged, and the values themselves are never corrupted — only the column's timezone
*label* changes, which is enough to change what a `WHERE` clause means.

`DuckDB` labels a `TIMESTAMPTZ` column it exports to Arrow with the connection's own
`TimeZone` setting. The accelerator pins that setting to `UTC`
(`crates/accelerators/accelerator-duckdb/src/settings.rs:140`, `DuckDBSettingScope::Local`),
but the connector builds its pool directly — `DuckDbConnectionPool::new_file` /
`new_memory` in `crates/data-connectors/connector-duckdb/src/lib.rs:101,126` — with no
setup query, so its session timezone is whatever `DuckDB` derives from the host.

`DataFusion` then coerces a naive timestamp literal into the column's zone
(`temporal_coercion_nonstrict_timezone`: `(Some(lhs_tz), None) => Some(lhs_tz)`), and Arrow's
naive->aware cast reads the literal as wall-clock *in that zone*. So the literal in
`WHERE ts > TIMESTAMP '...'` denotes a different instant on every host.

## To Reproduce

A `DuckDB` file with one `TIMESTAMPTZ` column, four rows:

```sql
CREATE TABLE tz AS SELECT * FROM (VALUES
  (1, TIMESTAMPTZ '2024-01-15 12:00:00+00'),
  (2, TIMESTAMPTZ '2024-07-15 12:00:00+00'),
  (3, TIMESTAMPTZ '2024-03-10 02:30:00+00'),
  (4, TIMESTAMPTZ '2024-11-03 01:30:00+00')) AS t(id, aware_ts);
```

```yaml
datasets:
  - from: duckdb:tz
    name: conn
    params:
      duckdb_open: /path/tz.db
```

Run `spiced` three times, changing **only** the process `TZ`:

```
TZ=Asia/Tokyo        schema=Timestamp(us, "Asia/Tokyo")        >09:00 -> [1,2,3,4]   >15:00 -> [1,2,3,4]
TZ=UTC               schema=Timestamp(us, "UTC")               >09:00 -> [1,2,3,4]   >15:00 -> [2,3,4]
TZ=America/New_York  schema=Timestamp(us, "America/New_York")  >09:00 -> [2,3,4]     >15:00 -> [2,3,4]
```

where the two queries are

```sql
SELECT id FROM conn WHERE aware_ts > TIMESTAMP '2024-01-15 09:00:00' ORDER BY id;
SELECT id FROM conn WHERE aware_ts > TIMESTAMP '2024-01-15 15:00:00' ORDER BY id;
```

Each host timezone gives a distinct pair of answers. Row 1 is `2024-01-15T12:00:00Z`; whether
it satisfies `> 15:00` depends only on which machine the runtime happens to be running on.

The schema is directly observable:

```sql
SELECT data_type FROM information_schema.columns
 WHERE table_name = 'conn' AND column_name = 'aware_ts';
```

This is **not** a federation defect: an `arrow`-accelerated dataset over the same connector
(same schema, evaluated locally in `DataFusion`) returns exactly what the federated one
returns on each host. The divergence is between hosts, not between execution paths.

## Expected behavior

A dataset's schema, and the rows a query returns, do not depend on the host's timezone.

## Impact

Silent wrong results for any `duckdb`-connector dataset with a `TIMESTAMPTZ` column that is
compared against a naive timestamp literal, on any host not running UTC. A container running
UTC — the common deployment — masks it entirely, which is why it has gone unnoticed.

The same unpinned-pool shape is present by inspection (not separately reproduced) in the
DuckLake connector (`crates/data-connectors/connector-ducklake/src/lib.rs:137,151`) and the
DuckLake catalog connector (`crates/runtime/src/catalogconnector/ducklake.rs:232,241`).

## Related

Same "host timezone is not UTC -> wrong rows" class as #12530, but a different path and a
different mechanism: #12530 is the accelerator's *write/delete* pool missing the `Local`-scope
setting and generating a session-dependent comparison. This one is the *connector's read*
pool, where the host timezone reaches the **Arrow schema** and changes the meaning of a
literal before any SQL is generated.

Found while verifying #13633.

claudespice:
## Summary

A `duckdb` connector dataset's Arrow schema carried **the host's timezone**, so the same query
over the same database file returned **different rows on different machines**.

`DuckDB` labels a `TIMESTAMPTZ` column it exports to Arrow with the connection's own `TimeZone`
setting. The accelerator pins that to `UTC`
(`accelerator-duckdb`'s `settings::TimeZone`, `DuckDBSettingScope::Local`), but the connector
built its pool directly — `DuckDbConnectionPool::new_file` / `new_memory` — with no setup query,
so its session timezone was whatever `DuckDB` derived from the host.

`DataFusion` then coerces a naive timestamp literal into the column's zone
(`temporal_coercion_nonstrict_timezone`: `(Some(lhs_tz), None) => Some(lhs_tz)`), and Arrow's
naive->aware cast reads that literal as wall-clock *in* that zone. So the literal in
`WHERE ts > TIMESTAMP '...'` denoted a different instant on every host.

## Changes

- `data_components::duckdb::with_utc_session_timezone` pins a pool's session to `UTC`, appending
  to any setup queries the caller already set rather than replacing them.
- Applied to every `DuckDB` pool Spice opens for user data that did not already have it: the
  `duckdb` connector (file + memory), the `ducklake` connector (file + memory), and the
  `ducklake` catalog connector (file + memory).

The accelerator is unchanged — it already pins the same setting through its settings registry.

## Test plan

`make lint` and `make nextest` (via `make signoff`).

Sign-off ran **remotely** (`signoff.yml` run `33965840241`) and passed; `signoff=success` and `Attestation` is green. Two local `make signoff` attempts
died on `No space left on device` (os error 28) — the first in `lint-rust`, the second while
building the nextest binaries — on a host whose volume is 96% full with sibling worktrees. Neither
posted a status, so the commit stayed eligible for a clean run. The first local attempt did catch a
real `clippy::doc_markdown` failure in a test doc comment, fixed in `ee05c737ad`; the second cleared
both full-workspace clippy passes before running out of disk.

### The bug, end to end

Four rows in a `DuckDB` file, one `TIMESTAMPTZ` column, registered as a `duckdb` connector
dataset. Only the `spiced` process `TZ` changes between runs — same file, same spicepod, same
queries:

```sql
SELECT id FROM conn WHERE aware_ts > TIMESTAMP '2024-01-15 09:00:00' ORDER BY id;
SELECT id FROM conn WHERE aware_ts > TIMESTAMP '2024-01-15 15:00:00' ORDER BY id;
```

**Before** (trunk `4735581da0`) — three host timezones, three distinct behaviours:

```
TZ=Asia/Tokyo        schema=Timestamp(us, "Asia/Tokyo")        >09:00 -> [1,2,3,4]   >15:00 -> [1,2,3,4]
TZ=UTC               schema=Timestamp(us, "UTC")               >09:00 -> [1,2,3,4]   >15:00 -> [2,3,4]
TZ=America/New_York  schema=Timestamp(us, "America/New_York")  >09:00 -> [2,3,4]     >15:00 -> [2,3,4]
```

**After** (this branch) — identical on every host, and identical to what `TZ=UTC` already returned:

```
TZ=Asia/Tokyo        schema=Timestamp(us, "UTC")               >09:00 -> [1,2,3,4]   >15:00 -> [2,3,4]
TZ=UTC               schema=Timestamp(us, "UTC")               >09:00 -> [1,2,3,4]   >15:00 -> [2,3,4]
TZ=America/New_York  schema=Timestamp(us, "UTC")               >09:00 -> [1,2,3,4]   >15:00 -> [2,3,4]
```

Row 1 is `2024-01-15T12:00:00Z`; before the fix, whether it satisfied `> 15:00` depended only on
which machine the runtime happened to be running on.

### Regression guards

```
$ cargo test -p data_components --features duckdb --lib duckdb::tests
running 2 tests
test duckdb::tests::the_pin_keeps_the_setup_queries_already_on_the_pool ... ok
test duckdb::tests::the_pin_overrides_a_zone_the_session_already_carries ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 589 filtered out; finished in 0.12s
```

Neuter-verified — with `with_utc_session_timezone` reduced to returning the pool unchanged, both
fail, so neither passes on the unfixed code:

```
test duckdb::tests::the_pin_keeps_the_setup_queries_already_on_the_pool ... FAILED
  left: ["SET memory_limit = '123MB'"]
 right: ["SET memory_limit = '123MB'", "SET TimeZone = 'UTC'"]
test duckdb::tests::the_pin_overrides_a_zone_the_session_already_carries ... FAILED
  left: "Asia/Tokyo"
 right: "UTC"
```

The first guard seeds the pool with `Asia/Tokyo` deliberately: on a CI host already running UTC
the assertion would otherwise hold with or without the pin and measure nothing.

### Scope of the evidence

The `duckdb` connector is reproduced end to end above. The `ducklake` connector and the
`ducklake` catalog connector are **fixed by inspection** — same pool type, same export
convention, same missing setup query — and are covered only by the unit guards, not by a
separate end-to-end run.

## Performance impact

None. One `SET TimeZone = 'UTC'` per connection, alongside the setup queries the pool already
runs, on a code path that opens a connection.

## Review gate

- **Adversarial review — declared skip.** `codex adversarial-review` (thread
  `01a0705c-fae0-7fd3-a7a4-0820567b8003`) failed with the engine's own reason: *"Your workspace is
  out of credits. Ask your workspace owner to refill in order to continue."* No `grok` engine is
  installed in the plugin cache, so no substitute was available.
- **`/security-review`: no findings.** The only new SQL is a `const &str` with no interpolation,
  so there is no injection vector; the change adds no route, deserialization, authz, secret, or
  path handling. The append-rather-than-replace semantics were checked specifically so the pin
  cannot silently drop an existing setup statement such as `search`'s `LOAD vss`.
- **Copilot: approval recommended, 0 comments.**
- **`/simplify`: applied.** The first draft restructured the `map_err` closures at each call site
  and cost 43 lines of reformatting churn per connector; wrapping the existing expression whole
  instead reduced that to 9 lines each, with the original lines untouched. No behaviour change.

Fixes #13899

## Checklist

- [x] Reproduced on `trunk` before the fix, and the same command re-run on the branch
- [x] Regression tests added, and verified to fail with the fix neutered
- [x] `make signoff` green and `Attestation` re-run

### DIFF
--- a/crates/data-connectors/connector-duckdb/src/lib.rs
+++ b/crates/data-connectors/connector-duckdb/src/lib.rs
@@ -25,6 +25,7 @@ limitations under the License.
 
 use async_trait::async_trait;
 use data_components::Read;
+use data_components::duckdb::with_utc_session_timezone;
 use data_connector_api::ConnectorContext;
 use data_connector_api::{
     AnyErrorResult, ConnectorComponent, ConnectorParams, DataConnector, DataConnectorError,
@@ -97,7 +98,7 @@ impl DuckDB {
     ///
     /// Returns an error if the in-memory `DuckDB` connection cannot be established.
     pub fn create_in_memory(params: &ConnectorParams) -> AnyErrorResult<DuckDBTableFactory> {
-        let pool = Arc::new(
+        let pool = Arc::new(with_utc_session_timezone(
             DuckDbConnectionPool::new_memory()
                 .map_err(|source| DataConnectorError::UnableToConnectInternal {
                     dataconnector: "duckdb".to_string(),
@@ -109,7 +110,7 @@ impl DuckDB {
                         .unsupported_type_action
                         .unwrap_or(UnsupportedTypeAction::Error),
                 ),
-        );
+        ));
 
         Ok(Self::with_spice_deny_list(
             DuckDBTableFactory::new(pool).with_dialect(new_duckdb_dialect()),
@@ -122,7 +123,7 @@ impl DuckDB {
     ///
     /// Returns an error if the file-based `DuckDB` connection cannot be established.
     pub fn create_file(path: &str, params: &ConnectorParams) -> AnyErrorResult<DuckDBTableFactory> {
-        let pool = Arc::new(
+        let pool = Arc::new(with_utc_session_timezone(
             DuckDbConnectionPool::new_file(path, &AccessMode::ReadOnly)
                 .map_err(|source| DataConnectorError::UnableToConnectInternal {
                     dataconnector: "duckdb".to_string(),
@@ -134,7 +135,7 @@ impl DuckDB {
                         .unsupported_type_action
                         .unwrap_or(UnsupportedTypeAction::Error),
                 ),
-        );
+        ));
 
         Ok(Self::with_spice_deny_list(
             DuckDBTableFactory::new(pool).with_dialect(new_duckdb_dialect()),
--- a/crates/data-connectors/connector-ducklake/src/lib.rs
+++ b/crates/data-connectors/connector-ducklake/src/lib.rs
@@ -20,6 +20,7 @@ limitations under the License.
 
 use async_trait::async_trait;
 use data_components::Read;
+use data_components::duckdb::with_utc_session_timezone;
 use data_components::ducklake::writer::DuckDbFederatedTableWriter;
 use data_components::ducklake::{
     DuckLakeS3Params, build_ducklake_attach_sql, configure_duckdb_httpfs,
@@ -133,7 +134,7 @@ fn create_ducklake_factory(
     params: &ConnectorParams,
 ) -> AnyErrorResult<(DuckDBTableFactory, Arc<DuckDbConnectionPool>, String)> {
     let pool = if let Some(path) = open_path {
-        Arc::new(
+        Arc::new(with_utc_session_timezone(
             DuckDbConnectionPool::new_file(path, &AccessMode::ReadWrite)
                 .map_err(|source| DataConnectorError::UnableToConnectInternal {
                     dataconnector: "ducklake".to_string(),
@@ -145,9 +146,9 @@ fn create_ducklake_factory(
                         .unsupported_type_action
                         .unwrap_or(UnsupportedTypeAction::Error),
                 ),
-        )
+        ))
     } else {
-        Arc::new(
+        Arc::new(with_utc_session_timezone(
             DuckDbConnectionPool::new_memory()
                 .map_err(|source| DataConnectorError::UnableToConnectInternal {
                     dataconnector: "ducklake".to_string(),
@@ -159,7 +160,7 @@ fn create_ducklake_factory(
                         .unsupported_type_action
                         .unwrap_or(UnsupportedTypeAction::Error),
                 ),
-        )
+        ))
     };
 
     let conn = Arc::clone(&pool).connect_sync().map_err(|source| {
--- a/crates/data_components/src/duckdb.rs
+++ b/crates/data_components/src/duckdb.rs
@@ -110,7 +110,7 @@ mod tests {
     }
 
     /// `with_connection_setup_queries` replaces the whole list, so the pin has to append to
-    /// what a caller already set rather than dropping it — `search`'s DuckDB index relies on a
+    /// what a caller already set rather than dropping it — `search`'s `DuckDB` index relies on a
     /// `LOAD vss` setup query, and losing one would be silent.
     #[test]
     fn the_pin_keeps_the_setup_queries_already_on_the_pool() {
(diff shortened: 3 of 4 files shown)


claudespice:
@copilot review

lukekim (approved):

Issue #13687, opened 2026-08-29
make lint-rust cannot pass on Windows on unmodified trunk: a cfg(windows)-only clippy::unnecessary_wraps blocks every local sign-off


grokspice:
## Summary

`make lint-rust` — and therefore `make signoff` — cannot pass on a Windows host, on **unmodified
trunk**, because of a `clippy::pedantic` lint that only the `cfg(windows)` arm can trigger.

`crates/runtime-request-context/src/user_agent.rs` defines `get_os_version_internal` twice:

```rust
#[cfg(target_family = "unix")]
fn get_os_version_internal() -> Result<String, GenericError> {
    let output = Command::new("uname").arg("-r").output()?;   // <- can fail
    let version = String::from_utf8(output.stdout)?;          // <- can fail
    Ok(version)
}

#[cfg(target_family = "windows")]
fn get_os_version_internal() -> Result<String, GenericError> {
    use winver::WindowsVersion;
    if let Some(version) = WindowsVersion::detect() {
        Ok(version.to_string())
    } else {
        Ok("unknown".to_string())                             // <- never Err
    }
}
```

The Windows body has no fallible operation — it already turns "could not detect" into
`Ok("unknown")` — so every path returns `Ok` and `clippy::unnecessary_wraps` fires:

```
error: this function's return value is unnecessary
   --> crates/runtime-request-context/src/user_agent.rs:203
    = note: `-D clippy::unnecessary-wraps` implied by `-D clippy::pedantic`
error: could not compile `runtime-request-context` (lib) due to 1 previous error
make[1]: *** [Makefile:288: lint-rust] Error 101
```

The Unix arm uses `?` twice, so it does not fire there.

## Why CI is green

Every job that runs `lint-rust` runs on Linux or macOS, where only the Unix arm is compiled. The
lint is unreachable in CI by construction, so this can sit on trunk indefinitely while being a hard
stop for anyone linting on Windows.

## Impact

`scripts/signoff` runs `make lint-rust` first, so **no local sign-off can succeed on a Windows
host for any branch**, including one whose diff does not touch this crate. Confirmed on two
unrelated branches today: one whose only change vs trunk is
`crates/llms/tests/llms/mod.rs`, and one touching only `crates/data_components` and
`crates/runtime/tests`. Both fail identically, in this crate.

The practical cost is that Windows contributors have to fall back to `make signoff-remote` for
every push, which is the shared-runner path that `docs/dev/ci_signoff.md` explicitly asks people
[...]

Issue #13843, opened 2026-09-03
bug(ci): a budget-expired sign-off still posts signoff=failure — the signalled guard is unreachable because the runner never signals the script


claudespice:
### What happens

A remote sign-off that expires its 353-minute step budget still posts `signoff=failure` — "Sign-off checks failed after ~21200s" — about a commit nothing judged. The workflow's own `Resolve an incomplete sign-off` step posts the honest `pending` first, and the dying script's `failure` **overwrites it 10–23 seconds later**.

This is the failure #12518 and #12741 describe, both closed `COMPLETED` (2026-08-06 / 2026-08-08). It is still happening 3+ weeks later, and this issue carries the mechanism the existing guard doesn't cover.

### Measured — the outcome, n=3

Commit statuses, chronological (`GET /repos/spiceai/spiceai/commits/<sha>/statuses`):

| commit | honest `pending` correction | dying script's `failure` | gap | step duration |
|---|---|---|---|---|
| `d77c1f4d2f` (#13738) | `2026-08-30T08:40:21Z` | `2026-08-30T08:40:32Z` | **11 s** | 21204 s |
| `3497c2fcf6` (#13738) | `2026-09-02T20:36:53Z` | `2026-09-02T20:37:16Z` | **23 s** | 21219 s |
| `62718243e5` (#13736) | `2026-09-03T00:14:43Z` | `2026-09-03T00:14:53Z` | **10 s** | 21203 s |

Two different triggering identities, three days apart. Final state on all three: `signoff=failure`. Per #12362 `Attestation` rejects a failed sign-off on the head commit, so each PR went red claiming a code failure that never happened. `#13736` is still red on this today.

Runs `33666474391` (#13736, runner `spiceai-macos-macmini-runner-03-02`) and `33643686762` (#13738, runner `spiceai-macos-macmini-runner-02-01`). Both: step `Sign off` → `failure`, and step `Correct the sign-off status when this run is cancelled` → **`skipped`**, confirming these were budget expiries and not cancellations. Job-level durations 21192 s and 21193 s — a **1-second spread across two different runners**, i.e. a fixed limit, not contention.

### Measured — the mechanism

[...]

PR #13923, opened 2026-09-05
fix(cli): keep the API key on its origin in the SDK-built query client (fixes #12502)

### LINKED ISSUE #12502: spice query and nsql analyze build their own SDK client, so the same-origin redirect policy does not cover them

grokspice:
## Summary

`spice query` and `spice nsql analyze` do not use `RuntimeContext::http_client`; they build their own credentialed client from the pinned `spiceai` SDK (`bin/spice/src/commands/query/mod.rs::build_client`, `bin/spice/src/commands/nsql/analyze.rs::build_spice_client`). The CLI's same-origin redirect policy from #12495 therefore never reached them, and the SDK at the pinned rev `93b75c3` sent the key as a custom `X-API-Key` header on a stock `reqwest` client — which strips only `Authorization`, `Cookie`, `cookie2`, `Proxy-Authorization` and `WWW-Authenticate` on a cross-origin hop. A runtime, proxy or ingress answering a `POST /v1/queries` with an off-origin `Location` was handed the key.

The SDK fixed this in spiceai/spice-rs#85 (merged 2026-08-05): every HTTP client it builds now carries a same-origin redirect policy (`src/redirect.rs`), and a hop that leaves the origin is refused rather than followed. This PR moves the workspace pin to that commit and adds a CLI-side regression test that exercises `build_client` itself, so the guarantee is checked where the CLI relies on it, not only inside the SDK.

`spice nsql analyze` builds its client the same way (`build_spice_client`) and the SDK policy covers it identically; per the issue's own analysis it only sends the key when it actually reaches the `/v1/queries` API, which is why the test targets `spice query`.

## Changes

- `Cargo.toml` / `Cargo.lock`: `spiceai` git pin `93b75c3c` (spice-rs#77) → `3267ff35` (spice-rs#85). The lockfile change is the single `source` line; validated with `cargo metadata --locked`. The SDK's own dependency pins are unchanged in kind (`arrow` 58 → 58.3 tightens within the version the lock already resolves to; `reqwest` stays 0.12).
- `bin/spice/src/commands/query/mod.rs`: two `#[cfg(test)]` tests against `build_client` with `wiremock` — an off-origin 307 must not deliver `X-API-Key` to the redirect target and must not read as a submitted query; a same-origin 307 is still followed with the key.
- `bin/spice/src/context.rs`: `RuntimeContext::with_api_key_for_test`, a `#[cfg(test)]` builder so the test context carries a key.

## Test plan

- [x] `make lint` (fmt + clippy with the repo's deny set; never a per-crate invocation)
- [x] `make test` / `make nextest` via `make signoff` after push
- [x] Reproduction on `trunk` (pin `93b75c3`), then the same command on this branch:

```
# trunk pin (93b75c3c), same test module:
$ cargo test --manifest-path bin/spice/Cargo.toml --lib -- commands::query::tests
test commands::query::tests::the_query_client_follows_a_same_origin_redirect_with_the_api_key ... ok
test commands::query::tests::the_query_client_does_not_carry_the_api_key_off_its_origin ... FAILED
---- commands::query::tests::the_query_client_does_not_carry_the_api_key_off_its_origin stdout ----
panicked at bin\spice\src\commands\query\mod.rs:1028:9:
the redirect target received 1 request(s); the first carried X-API-Key = Some("cli-api-key-must-not-leak")
test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 524 filtered out

# this branch (3267ff35):
$ cargo test --manifest-path bin/spice/Cargo.toml --lib -- commands::query::tests
test commands::query::tests::the_query_client_does_not_carry_the_api_key_off_its_origin ... ok
test commands::query::tests::the_query_client_follows_a_same_origin_redirect_with_the_api_key ... ok
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 524 filtered out
```

Run on Windows, where `bin/spice`'s test target only compiles with a throwaway local gate on two pre-existing Unix-only tests (`nix::unistd` in `connect/service`, `passwd_entry_home` in `context.rs`); that patch was reverted before commit and is not in this diff. `cargo clippy --lib --no-deps -D warnings -W clippy::pedantic` on the crate reports nothing in the two files this PR touches; its remaining errors are trunk's Windows-cfg lints (`runtime_launcher.rs`, `connect/service/manifest.rs`, `cluster.rs`), the shape #13687 tracks, so the full `make lint` verdict comes from the sign-off run.

## Review gate

- Adversarial review: **skipped** — `codex` exited 1: "Your workspace is out of credits"; `grok` is not permitted under this identity (receipt on head `ebc9531d9a`)
- `/security-review`: no findings — the diff is a dependency pin move plus `#[cfg(test)]` code (run by hand against this worktree's diff; the slash command diffs the primary checkout). The SDK delta between the two pins (`93b75c3c..3267ff35`: `client.rs`, `params.rs`, `query.rs`, new `redirect.rs`) was read for credential handling: it adds no logging of the key, and the one credential-touching change is the same-origin redirect policy that narrows the exposure.
- `/simplify`: reviewed the branch diff by hand for reuse and dead code — nothing to apply; the two test helpers are each used twice and the context builder mirrors the existing `with_deadlines_for_test`

Fixes #12502

## Checklist

- [x] No `.unwrap()` in new code (tests use `expect` with a reason)
- [x] `cargo fmt --all` clean; no `.claude/worktrees/` paths
- [x] No customer, credential, or internal-infrastructure detail in this body or the diff

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20002,7 +20002,7 @@ dependencies = [
 [[package]]
 name = "spiceai"
 version = "3.2.0"
-source = "git+https://github.com/spiceai/spice-rs.git?rev=93b75c3c7d1a812b23d00b45adf537d7a2415496#93b75c3c7d1a812b23d00b45adf537d7a2415496"
+source = "git+https://github.com/spiceai/spice-rs.git?rev=3267ff355f1bc2e0f382aa0b1e0113e5b98f1d76#3267ff355f1bc2e0f382aa0b1e0113e5b98f1d76"
 dependencies = [
  "arrow",
  "arrow-flight",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -472,7 +472,7 @@ vortex-session = { git = "https://github.com/spiceai/vortex.git", rev = "3c52468
 vortex-utils = { git = "https://github.com/spiceai/vortex.git", rev = "3c5246867e627158211e8eafab7be55b0dc559f8" } # branch: spiceai-54 (vortex 0.79.0)
 
 x509-certificate = "0.25.0"
-spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "93b75c3c7d1a812b23d00b45adf537d7a2415496" } # spiceai/spice-rs#77
+spiceai = { git = "https://github.com/spiceai/spice-rs.git", rev = "3267ff355f1bc2e0f382aa0b1e0113e5b98f1d76" } # spiceai/spice-rs#85: same-origin redirect policy on the credentialed client
 
 [profile.release-lto]
 inherits = "release"
--- a/bin/spice/src/commands/query/mod.rs
+++ b/bin/spice/src/commands/query/mod.rs
@@ -950,3 +950,125 @@ fn print_query_help() {
     println!("  - Press Ctrl+D or type .exit to quit");
     println!();
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::context::Deadline;
+    use serde_json::json;
+    use wiremock::matchers::{method, path};
+    use wiremock::{Mock, MockServer, Request, ResponseTemplate};
+
+    const API_KEY: &str = "cli-api-key-must-not-leak";
+
+    fn submitted() -> ResponseTemplate {
+        ResponseTemplate::new(202).set_body_json(json!({
+            "query_id": "query-1",
+            "status": "PENDING",
+            "status_url": "/v1/queries/query-1/status",
+            "results_url": "/v1/queries/query-1/results"
+        }))
+    }
+
+    /// A 307 keeps the method and the body, so a hop that is followed replays the
+    /// credentialed `POST` verbatim to `location`.
+    fn redirect_to(location: &str) -> ResponseTemplate {
+        ResponseTemplate::new(307).insert_header("location", location)
+    }
+
+    fn context_for(runtime: &MockServer) -> RuntimeContext {
+        RuntimeContext::with_deadlines_for_test(
+            &runtime.uri(),
+            Deadline::Total(Duration::from_secs(30)),
+            Deadline::Silence(Duration::from_secs(30)),
+        )
+        .with_api_key_for_test(API_KEY)
+    }
+
+    fn api_key_header(request: &Request) -> Option<String> {
+        request
+            .headers
+            .get("x-api-key")
+            .and_then(|value| value.to_str().ok())
+            .map(str::to_owned)
+    }
+
+    /// `spice query` builds its client from the SDK rather than from
+    /// `RuntimeContext::http_client`, so the CLI's own same-origin redirect policy does not
+    /// reach it: the SDK has to refuse to leave the origin itself. The key travels as
+    /// `X-API-Key`, which `reqwest` does not strip on a cross-origin hop the way it strips
+    /// `Authorization`, so a runtime, proxy or ingress answering with an off-origin
+    /// `Location` would otherwise be handed the key.
+    /// <https://github.com/spiceai/spiceai/issues/12502>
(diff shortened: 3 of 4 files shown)


grokspice:
@copilot review

grokspice:
The red `signoff` on `ebc9531d9a` is a budget expiry, not a verdict on this diff — the #13843 shape.

Run [33996118726](https://github.com/spiceai/spiceai/actions/runs/33996118726) hit `The action 'Sign off' has timed out after 353 minutes` inside `make lint-rust`: the log shows two silent gaps in the first `cargo clippy` pass (22:38 → 00:09 and 00:10 → 02:31 UTC, no output between `Checking graph-error` and the next line), so the test phase never started. The workflow's `Resolve an incomplete sign-off` step then posted the honest status —

```
04:24:58Z pending: Remote sign-off ended without posting a verdict — sign off again
04:25:05Z failure: Sign-off checks failed after 21201s
```

— and the dying script overwrote it seven seconds later, which is exactly what #13843 describes. No test ran against this commit and nothing in the log names a file from this PR.

Re-dispatched once as a fresh run; not pushing anything, since the diff has not changed.

lukekim (approved):

PR #13926, opened 2026-09-05
fix(search): filter a chunked Elasticsearch delete on a field that can match the key (fixes #13714)

### LINKED ISSUE #13714: Chunked Elasticsearch deletes still filter on analyzed key columns, so a hyphenated string key deletes nothing and reports success
### Description

A chunked search index stores one document per text chunk, keyed by the source row's primary key
plus a chunk ordinal. Removing a source row therefore means removing every document under its
*base* key — a strict subset of the stored primary key — so `ChunkedSearchIndex` hands the base key
to `Index::delete_by_keys`, and Elasticsearch reports `deletes_by_partial_key() == true` to receive
it.

That partial-key path filters on the key **columns**, and `crates/search/src/index/elasticsearch/delete.rs`
already records why that is unreliable: nothing maps the primary-key columns, so a string key is
dynamically mapped `text` and its inverted index holds *analyzed* tokens. A `term` query is not
analyzed, so `{"term": {"id": "ORDER-1024"}}` looks for one token in an index holding
`[order, 1024]` and matches nothing.

The whole-key case was fixed in #12267 by addressing documents through `_id` instead. The doc
comment on `delete_by_keys` states that the fix does not reach the other case:

> A strict subset of the key (the chunked-index case) cannot use `_id`, because the chunk id is
> part of it and is unknown at delete time; that case still filters on the key columns and so
> remains subject to the mapping gap above.

`_delete_by_query` then returns `total: 0, deleted: 0`, which `inspect_delete_response` accepts as a
successful delete — so the caller is told every document went, and none did.

### Impact

On a chunked Elasticsearch index with a string primary key containing a character the standard
analyzer splits on (`-`, `_`, `.`, `/`, `@` — so most human-readable identifiers), a delete removes
nothing and reports success. The documents stay searchable for a row the dataset no longer has, or
for text the row no longer holds. Wrong results, silently, with no failed operation anywhere.

An integer key is unaffected, which is why this does not show up in the common case.

### Where

- `crates/search/src/index/elasticsearch/delete.rs` — `build_or_of_row_term_queries`, reached
  whenever `key_columns` is a strict subset of `primary_key`, and `inspect_delete_response`, which
  treats a zero-match response as success.
- `crates/search/src/index/chunking.rs` — `ChunkedSearchIndex::delete_by_keys`, the partial-key
  caller, and the eviction on the write path that also routes through it.

### Suggested direction

Map the primary-key columns explicitly as `keyword`/numeric when the runtime creates the index, so
a `term` query on them is exact. Failing that, the chunked case needs a way to address documents by
`_id` — which means knowing the chunk ordinals, i.e. resolving them first.

Separately: a partial-key delete that matches zero documents where the caller believes rows exist is
not obviously a success. Whether it can be distinguished from a legitimately-absent row is worth
deciding as part of this.

### Origin

Raised by an adversarial review of the fix for #13704 and confirmed against the code and the
existing doc comment. Pre-existing: the same gap affects the delete path today, independently of
that fix. #12267 covers the whole-key half and is closed; this is the half it explicitly left open.

claudespice:
## Summary

A chunked search index stores one document per text chunk, keyed by the source row's primary key
plus a chunk ordinal. Removing a source row means removing every document under its *base* key — a
strict subset of the stored key — so that delete cannot address documents by `_id` the way a
whole-key delete does (#12267): the chunk ordinal is part of the `_id` and is unknown at delete
time. It filtered on the key *columns* instead.

Nothing mapped those columns. A string key was left to dynamic mapping, which makes it `text`, an
inverted index of *analyzed* tokens — so `{"term": {"id": "ORDER-1024"}}` looked for one token in
an index holding `[order, 1024]` and matched nothing. `_delete_by_query` returned
`total: 0, deleted: 0`, which `inspect_delete_response` reads as a fully applied delete. The caller
was told every document went and none did, so the index kept serving documents for rows the dataset
no longer had. Any human-readable identifier hits this — the standard analyzer splits on `-`, `_`,
`.`, `/`, `@` — while an integer key, mapped `long`, was never affected.

The fix is in two halves, because the mapping cannot be repaired on an index that already exists:

1. **The runtime maps its key columns exactly** when it creates the index — `keyword` for a string
   (bare, not the `text` + `keyword` multi-field other string columns get, whose `ignore_above: 256`
   would leave a longer key unindexed), native types otherwise. A column another mapping already
   claimed keeps it: that is the user's declared intent, or a search field that has to stay
   analyzable. On an index that already exists, only the key columns it does not map yet are added,
   in a request of their own — Elasticsearch cannot change an existing field's type, so sending one
   it already mapped could only be rejected, and a rejected `put_mapping` would take the other
   mapping updates in that request down with it.
2. **The delete resolves the field it can actually match on**, by reading the index's live mapping:
   the column itself when it is exact, its `keyword` multi-field when the column is `text` — which
   is what every index created before this has. A column that resolves to neither is refused **by
   name**, before any request is issued, rather than filtered on. That is the half that matters
   beyond this bug: a filter that cannot match must not be reported as a delete that applied.

That also settles the question the issue left open — whether a zero-match partial-key delete can be
told apart from a legitimately absent row. It can now, because the case that made zero-match
ambiguous is refused up front. A key column missing from the mapping is the one remaining zero-match
shape, and it is unambiguous: Elasticsearch maps a field the first time a document carries it, so no
document holds that column and there is nothing to delete.

Fixed for the whole class, not just the reported instance: an exact type mapped `index: false` —
what a key column declared non-filterable gets — indexes nothing and is refused the same way, and a
key past the resolved field's `ignore_above` is refused rather than filtered on. That last check
runs once over the batch before the first request, so a long key three chunks in cannot leave the
chunks before it applied and the ones after it unissued. `ElasticsearchTextIndex` shares the delete
path and gets the same treatment as `ElasticsearchIndex`.

## Changes

- `crates/search/src/index/elasticsearch/delete.rs` — `resolve_term_exact_paths` reads the mapping
  once per delete (partial-key path only; the whole-key path still addresses documents by `_id` and
  makes no extra request) and resolves each key column to the field a `term` matches its value on.
  `ensure_keys_are_indexable` refuses an over-length key before the first request, and
  `build_or_of_row_term_queries` names the resolved path.
- `crates/runtime/src/embeddings/index/elasticsearch/mod.rs` — `add_primary_key_mappings` maps the
  key columns into the create body; `add_missing_primary_key_mappings` fills the gaps on an index
  that already exists.
- `crates/runtime/src/search/full_text/table.rs` — passes the resolved primary key through.
- `crates/elasticsearch/src/lib.rs` — `FieldMapping` gains `fields` (multi-fields), `ignore_above`
  and `index`; `index_properties` resolves the `_mapping` body for an index or its alias.

**Performance impact:** one extra `GET /<index>/_mapping` per partial-key delete call — not per
request chunk, and not on the whole-key path at all. It is a cluster-state read issued alongside at
least one `_delete_by_query`, which scans and deletes documents. It is deliberately not cached: a
resolution cached before a field existed would permanently skip deletes on an index that later
gained documents.

## Test plan

- `make lint`
- `make test`

New unit tests, all of which fail on the code before this change:

- a `text`-mapped key column is filtered on its `keyword` sub-field, not on the column
- a key column with no exact-match field refuses, and issues no request
- an exact-typed but unsearchable (`index: false`) key column refuses the same way
- a key past `ignore_above` refuses; one exactly at it still deletes; a multi-byte key is measured
  in characters, not bytes
- a key column absent from the mapping issues no request and succeeds
- an unreadable mapping fails the delete rather than filtering blindly
- a mapping returned under a concrete index name is still read for an alias
- a numeric key column is still filtered on the column itself
- the runtime maps a string key `keyword` with no length limit, keeps a claimed column's mapping,
  and maps every column of a composite key

## Review gate

- Adversarial review: **skipped** — `codex` exited 1: "Your workspace is out of credits."; `grok` not installed
- `/security-review`: no findings — the diff adds no input-handling, auth, filesystem or outbound-request surface, and the new errors report the index name, the column name and a key's *length*, never the key value
- `/simplify`: applied — two cleanups with no behaviour change (`min_by` instead of collect-then-sort; the shared mapping helper plus `retain` instead of a second copy of its loop); light checks re-run green
- Re-run on the current head after a one-line `clippy::match_same_arms` fix in `is_indexed`: the adversarial engine is still out of credits, and the security review's conclusion is unchanged — merging two match arms that already returned the same value adds no surface.

Fixes #13714

## Checklist

- [x] Regression tests that fail on the code before this change and pass on it
- [x] Edge cases covered: NULL keys, empty batch, composite keys, multi-byte and over-length keys,
      alias-resolved mappings, unreadable mapping
- [x] No behaviour change on the whole-key delete path
- [x] User-facing errors name the index, the column and the fix, and carry a docs link


claudespice:
@copilot review

claudespice:
### The `signoff=failure` on `b4dc876` is located, and `6edbe0de` clears it — with the `--tests` pass this branch had never reached

Picked this up as the one real red on the board (the other five `Attestation` failures all carry `signoff = pending` against live runs). Landing the evidence here since the fix arrived from two directions at once.

**The failure**, reproduced locally with `make lint-rust PACKAGES="elasticsearch runtime search" FEATURES=<workspace resolve>`:

```
error: these match arms have identical bodies
   --> crates/elasticsearch/src/lib.rs:137:13
    = note: `-D clippy::match-same-arms` implied by `-D clippy::pedantic`
error: could not compile `elasticsearch` (lib) due to 1 previous error
make: *** [lint-rust] Error 101
```

Byte-identical to the branch's own remote sign-off `33999628723`, whose `--log-failed` contains the same diagnostic at the same file:line — and nothing else. `cargo fmt --check` was clean and all eight fast guards passed, so the arm was the whole failure. (The commit status carried `target_url: null` and `after 300s`, which is what marks it a *local* run and a real verdict rather than the ~21180s wall.)

**One thing worth knowing about that verdict: it never exercised half the gate.** `make lint-rust` runs a lib/bins clippy pass and then a second `--tests` pass; `--keep-going` only keeps *that* pass going, so `Error 101` on the first one skips the second entirely — which is where this branch's ~700 new lines in `delete.rs` live. Neither sign-off run had ever linted them.

**After `6edbe0de`, the same command is green through both passes:**

```
Finished `dev` profile [unoptimized + debuginfo] target(s) in 37m 52s
```

zero errors, `elasticsearch` checked twice (once per pass). I had arrived at the same consolidation independently and my push was rejected because yours landed first; the two differ only in the comment, so this result stands for the tree that is on the branch now.

No action needed — just a head start on `34002048233`.

lukekim (approved):

Issue #13945, opened 2026-09-07
The committed spicepod.schema.json is ~1,000 lines behind the generator, and no check compares them


claudespice:
## What

The committed `.schema/spicepod.schema.json` no longer matches what `tools/spicepodschema` generates from the current code, and nothing in the repository or CI compares the two.

## Measured

On `trunk` at `2e6499061f` (2026-09-07), regenerating the schema in place and diffing against the committed file:

```
$ cargo run -p spicepodschema -- .schema/spicepod.schema.json
$ git diff --shortstat -- .schema/
 1 file changed, 864 insertions(+), 201 deletions(-)
```

Definitions the generator emits that the committed schema does not have include `RuntimeDrasi`, `RuntimeDrasiTable`, `DrasiForwarding`, `DrasiTransport`, `Drasi`, `DrasiDelivery` and `OnDeliveryError`; the committed `Params`/`ParamValue` definitions are also out of date, and a number of `description` strings differ from the current doc comments.

The committed file was last regenerated on 2026-08-25 (`d99d0ae238`); `crates/spicepod/src` alone has changed in 5 commits since, and parameter specs under `crates/runtime/src/model/params/**` change more often than that.

## Why it drifts

`.github/workflows/generate_json_schema.yml` runs only on pushes to `trunk` that touch `crates/spicepod/**`, and even then it only uploads the result as an artifact; a PR with the regenerated file is created only on manual dispatch. Parameter specs live under `crates/runtime/src/**/params/**` and the collector under `tools/spicepodschema/**`, so a change there never triggers it. The only assertion the job makes is that the output file exists.

## Suggested fix

- A unit test in `tools/spicepodschema` that builds the schema in memory the way `main.rs` does, parses the committed file with `include_str!`, and asserts the two `serde_json::Value`s are equal, with a message naming the regeneration command. This runs in the normal test gate and fails the PR that introduces drift.
- Widen the workflow's `paths` filter to `crates/runtime/src/**/params/**` and `tools/spicepodschema/**`, or drop the filter, so the regeneration PR is offered whenever a spec changes.

## Context

Found while fixing #13932, whose fix had to edit the committed schema by hand for the HuggingFace model parameters and confirm it against the generator manually. That fix does not attempt to bring the rest of the file up to date, since the drift is unrelated to it.

Issue #6695, opened 2025-08-05
For LLMs, `hf_token` doesn't work, `huggingface_token` works with warning


Jeadie:
## Describe the bug
- [Docs](https://spiceai.org/docs/components/models/huggingface#params) says `hf_token` should work, but 401 when needed. `huggingface_token` gives warning.
- Could be similar to #6638

Jeadie:
non-issue

PR #13946, opened 2026-09-07
fix(models): read the HuggingFace chat token as hf_token again (fixes #13932)

### LINKED ISSUE #13932: HuggingFace LLMs no longer accept `hf_token`; the TypedParams refactor renamed the key to `huggingface_token` (regression of #6695)
## Describe the bug

A `models:` entry with `from: huggingface:...` silently ignores the documented `hf_token` parameter. The key the runtime actually reads is `huggingface_token`. Because unknown params are only warned about and dropped, the model loads anonymously and gated/private repos 401 at download time with nothing pointing at the cause.

This is a regression of #6695, which reported the same symptom and was closed as a non-issue in 2025.

The docs (`components/models/huggingface#params`, and both YAML examples on that page) document `hf_token`.

## Reproduction

Probe added to the existing `crates/runtime/src/model/params/mod.rs` test module, exercising the real deserialization path (`HuggingFaceModelParams::try_from_params`) — `cargo test -p runtime --lib`, trunk @ `e4c821c8a9`:

```
DOCS-AUDIT models  hf_token             -> token = None
DOCS-AUDIT models  huggingface_token    -> token = Some("hf_abc")
DOCS-AUDIT models  token                -> token = None
DOCS-AUDIT embed   hf_token             -> hf_token = Some("hf_abc")
DOCS-AUDIT embed   huggingface_token    -> hf_token = None
test result: ok. 1 passed
```

The probe:

```rust
#[tokio::test]
async fn hf_token_key_for_models() {
    for key in ["hf_token", "huggingface_token", "token"] {
        let typed = huggingface::HuggingFaceModelParams::try_from_params(
            "model huggingface",
            params(&[(key, "hf_abc")]),
            &empty_secrets(),
        ).await;
        // print typed.token
    }
}
```

## Cause

`crates/runtime/src/model/params/huggingface.rs` declares:

```rust
#[params(prefix = "huggingface", passthrough = ..., emit_specs)]
pub struct HuggingFaceModelParams {
    pub token: Option<SecretString>,   // component param -> "{prefix}_{name}"
```

`apply_prefix` in `crates/runtime-parameters-derive/src/lib.rs` composes the user-facing key as `{prefix}_{name}`, so `token` becomes **`huggingface_token`**.

Before the `TypedParams` refactor (#12155, merged 2026-08-08, first released in **v2.2.0**), the same field was `ParameterSpec::component("token")` in a list whose prefix came from `ModelSource::HuggingFace`, which is `"hf"` — so the key was `hf_token`, matching the docs. `ModelSource::HuggingFace => "hf"` is still what `crates/spicepod/src/component/model.rs` returns today; only the params-struct prefix changed.

The rename looks unintentional rather than deliberate:

- The **embeddings** spec (`crates/runtime/src/embeddings/params/huggingface.rs`) and the **reranker** spec (`crates/runtime/src/model/rerank/params/huggingface.rs`) both name the field `hf_token` with `#[param(runtime)]`, so they still accept the bare `hf_token`. Only chat models diverged.
- #12155 is described as a mechanical derive refactor, and nothing in it suggests a user-facing key change was intended.
- The result is a silently breaking change to a documented public parameter, in a patch-shaped refactor.

## Expected behaviour

`hf_token` keeps working for `from: huggingface:` models, consistent with the embeddings and reranker components and with the published docs. Renaming the field to `hf_token` (which `apply_prefix` leaves bare under `#[param(runtime)]`, matching embeddings) or adding `alias = "hf_token"` would both restore it; the reranker spec already uses the `alias` mechanism for `api_key`.

Docs were left unchanged pending a decision here — they currently document `hf_token`, which is right for v2.0.x/v2.1.x and for whatever this issue resolves to, and documenting `huggingface_token` would enshrine the regression.

## Affected versions

v2.2.0, v2.2.1, and trunk. v2.1.x and earlier are unaffected.

claudespice:
## Summary

A `from: huggingface:` chat model silently ignored the documented `hf_token` parameter, so a gated or private repository was downloaded anonymously and failed with a 401 that pointed at nothing.

`HuggingFaceModelParams` was given the prefix `huggingface` when it moved to `#[derive(TypedParams)]`, so its component params rendered as `huggingface_token`, `huggingface_temperature`, …. Everything around it still keys the source by its short name `hf`: `ModelSource::HuggingFace.short_name()` is `"hf"`, `construct_model` reads the prefixed chat overrides through `get_openai_request_overrides(component, source.short_name())`, the published docs and the integration tests pass `hf_token`, and the embeddings and reranker components read `hf_token` too. The disagreement had two consequences on every HuggingFace chat model:

- `hf_token` and `hf_temperature` were warned about as unknown parameters and dropped, so the model loaded without a token.
- `huggingface_temperature` (and the other 22 `huggingface_*` overrides the schema advertised) were consumed without a warning but never applied, because the override lookup reads `hf_*`.

Affects v2.2.0, v2.2.1 and `trunk`; v2.1.x is unaffected. Regression of #6695.

## Changes

- `HuggingFaceModelParams` uses the prefix `hf`, the source's short name, so its component params and its prefixed passthrough overrides are spelled the way the override lookup and the docs spell them.
- The token is the runtime field `hf_token`, matching the embeddings and reranker specs. `huggingface_token` stays accepted as an alias so a spicepod written against v2.2.0/v2.2.1 keeps loading; the schema advertises `hf_token` and names the alias in its description.
- The schema collector reads every model source's prefix from `ModelSource::short_name()` instead of a second string literal, so the published schema cannot drift from the runtime again on its own. `.schema/spicepod.schema.json` follows: `huggingface_*` → `hf_*` in the HuggingFace model params block (24 keys) and `"prefix": "hf"` in the `modelSources` metadata block. The rename was applied by hand and then checked against the generator: `cargo run -p spicepodschema -- .schema/spicepod.schema.json` on this branch produces the same 24 `hf_*` keys and the same `hf_token` description; the only HuggingFace-touching lines in the diff between its output and the committed file are a position change of `hf_tools` inside the passthrough block, which every source's block shares and which predates this branch. The generator's output also differs from the committed file by ~1,000 unrelated lines that predate this branch (the committed schema is stale, and no check compares it to the generator); that is tracked separately in #13945 rather than folded in here. The schema fixtures use `hf_token`.
- Unit tests: `hf_token` and the `huggingface_token` alias both deserialize; the emitted spec advertises a runtime `hf_token` and no bare `token`; and a class guard asserts every model params struct's `PREFIX` equals its source `short_name()`, since the override lookup depends on that equality for every source with a passthrough table.

Not changed: `hf_token` is not given `autoload_secret`. The pre-refactor spec did not autoload it either, and turning that on would silently start reading an `hf_token` secret for chat models that never asked for one.

## Test plan

Reproduced on a real `spiced` with a two-line manifest (`hf_token`, `hf_temperature`, `huggingface_temperature` on a `from: huggingface:` chat model pointing at a nonexistent repository, so the load fails fast after the params are read):

**Before** — installed `spiced v2.2.1+models.metal`:

```
WARN runtime_parameters_typed: Ignoring parameter `hf_token`: not supported for model huggingface.
WARN runtime_parameters_typed: Ignoring parameter `hf_temperature`: not supported for model huggingface. Did you mean `openai_temperature`?
WARN runtime::init::model: Failed to load LLM: hf_chat. ... Could not access `spiceai-repro/does-not-exist` (revision `main`) on Hugging Face (HTTP 401). ...
```

`huggingface_temperature` drew no warning at all.

**After** — `spiced` built from this branch, same manifest, same command:

```
WARN runtime_parameters_typed: Ignoring parameter `huggingface_temperature`: not supported for model huggingface.
WARN runtime::init::model: Failed to load LLM: hf_chat. ... Could not access `spiceai-repro/does-not-exist` (revision `main`) on Hugging Face (HTTP 401). ...
```

`hf_token` and `hf_temperature` are consumed without a warning, and the never-applied `huggingface_temperature` is now reported as unsupported instead of being silently swallowed. (The 401 is expected on both runs: the manifest carries a placeholder token against a repository that does not exist, so the log cannot distinguish an anonymous request from a bad token; the parameter warnings are the artifact.) Binaries: before `v2.2.1+models.metal`, after `v2.3.0-unstable-build.72fb3d69d9-dev+models`.

- [x] `cargo test -p runtime --lib --features models model::params` — 16 passed, 0 failed (run twice: on the fix, and again after the `/simplify` edits)
- [ ] `make lint`
- [ ] `make test`
- [ ] `make signoff` (dispatched remotely via `signoff.yml`; the local gate cannot run from a nested worktree)

## Review gate

- Adversarial review: **skipped** — `codex` exited 1: "Your workspace is out of credits. Ask your workspace owner to refill in order to continue."; `grok` not installed
- `/security-review`: no findings — verified that only parameter *keys* reach the leftover/deprecation warnings, the token stays a `SecretString` end to end, and the prefix change does not alter which params are marked secret or autoloaded
- `/simplify`: applied — the schema collector now derives prefixes from `short_name()` (it had a third copy of the literal, and the review found the `modelSources` metadata block still said `huggingface`), the guard test lists each source once, the struct doc points at the enforcing test, and the regression reference moved to a doc comment. Light checks re-run: `rustfmt --check` clean on every touched file, `cargo test -p runtime --lib --features models model::params` 16 passed

Fixes #13932

## Checklist

- [x] Root cause identified and fixed at the source (prefix disagreement), not the symptom
- [x] Regression tests added that fail on `trunk` and pass on this branch
- [x] Back-compat for the v2.2.x key preserved via an alias
- [ ] Sign-off attested on the pushed HEAD

### DIFF
--- a/.schema/spicepod.schema.json
+++ b/.schema/spicepod.schema.json
@@ -10795,8 +10795,8 @@
           "description": "Customizes the transformation of OpenAI chat messages into a character stream for the model.",
           "type": "string"
         },
-        "huggingface_token": {
-          "description": "The Huggingface access token.",
+        "hf_token": {
+          "description": "The Hugging Face access token. `huggingface_token` is accepted as an alias.",
           "type": "string"
         },
         "distributed_backend": {
@@ -10839,73 +10839,73 @@
           "description": "Maximum requests per minute for this model. Overrides provider defaults.",
           "type": "string"
         },
-        "huggingface_frequency_penalty": {
+        "hf_frequency_penalty": {
           "type": "string"
         },
-        "huggingface_logit_bias": {
+        "hf_logit_bias": {
           "type": "string"
         },
-        "huggingface_logprobs": {
+        "hf_logprobs": {
           "type": "string"
         },
-        "huggingface_top_logprobs": {
+        "hf_top_logprobs": {
           "type": "string"
         },
-        "huggingface_max_completion_tokens": {
+        "hf_max_completion_tokens": {
           "type": "string"
         },
-        "huggingface_reasoning_effort": {
+        "hf_reasoning_effort": {
           "type": "string"
         },
-        "huggingface_store": {
+        "hf_store": {
           "type": "string"
         },
-        "huggingface_metadata": {
+        "hf_metadata": {
           "type": "string"
         },
-        "huggingface_n": {
+        "hf_n": {
           "type": "string"
         },
-        "huggingface_presence_penalty": {
+        "hf_presence_penalty": {
           "type": "string"
         },
-        "huggingface_response_format": {
+        "hf_response_format": {
           "type": "string"
         },
-        "huggingface_seed": {
+        "hf_seed": {
           "type": "string"
         },
-        "huggingface_stop": {
+        "hf_stop": {
           "type": "string"
         },
-        "huggingface_stream": {
+        "hf_stream": {
           "type": "string"
         },
-        "huggingface_stream_options": {
+        "hf_stream_options": {
           "type": "string"
         },
-        "huggingface_temperature": {
+        "hf_temperature": {
           "type": "string"
         },
-        "huggingface_top_p": {
+        "hf_top_p": {
           "type": "string"
         },
-        "huggingface_tools": {
+        "hf_tools": {
           "type": "string"
         },
-        "huggingface_tool_choice": {
+        "hf_tool_choice": {
           "type": "string"
         },
-        "huggingface_parallel_tool_calls": {
+        "hf_parallel_tool_calls": {
           "type": "string"
         },
-        "huggingface_prompt_cache_key": {
+        "hf_prompt_cache_key": {
           "type": "string"
         },
-        "huggingface_prompt_cache_retention": {
+        "hf_prompt_cache_retention": {
           "type": "string"
         },
-        "huggingface_user": {
+        "hf_user": {
           "type": "string"
         },
         "openai_frequency_penalty": {
--- a/.schema/spicepod.schema.json
+++ b/.schema/spicepod.schema.json
@@ -454,7 +454,7 @@
       },
       {
         "name": "huggingface",
-        "prefix": "huggingface",
+        "prefix": "hf",
         "paramsRef": "#/$defs/HuggingfaceModelParams",
         "schemaRef": "#/$defs/HuggingfaceModel"
       },
--- a/crates/runtime/src/model/chat.rs
+++ b/crates/runtime/src/model/chat.rs
@@ -371,7 +371,7 @@ async fn huggingface(
     };
 
     let model_type = params.model_type.as_deref();
-    let hf_token = params.token.as_ref();
+    let hf_token = params.hf_token.as_ref();
 
     // For GGUF models, we require user specify via `.files[].path`
     let gguf_path = component
--- a/crates/runtime/src/model/params/huggingface.rs
+++ b/crates/runtime/src/model/params/huggingface.rs
@@ -19,9 +19,14 @@ use runtime_parameters::TypedParams;
 use secrecy::SecretString;
 
 /// Parameters for `from: huggingface` chat models.
+///
+/// The prefix is `hf`, the source's short name (`ModelSource::HuggingFace.short_name()`):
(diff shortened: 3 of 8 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #13954, opened 2026-09-08
fix(cayenne): use the configured cayenne_metadata_dir parameter

peasee:
## Problem

`CayenneAccelerator` cached a single `MetadataCatalog` in one `OnceCell`. `get_or_create_catalog` derives a per-dataset connection string from `cayenne_metadata_dir`, but `OnceCell::get_or_try_init` discards it once the cell is filled — so every Cayenne dataset after the first silently binds to the first one's `cayenne.db`.

Which dataset initializes first is a startup race, and it can resolve differently on the next start. When it does, the metastore that gets opened holds none of the expected table names, so every dataset mints a new `table_id` and starts empty. The data it previously accelerated is still on disk — `cayenne_file_path` was always honored, only the metadata was collapsed — but is no longer reachable, and nothing is logged.

Reproduced with six Cayenne-accelerated datasets, each with its own `cayenne_metadata_dir`. After a single restart all six returned 0 rows, and only two of the six `cayenne.db` files had ever been initialized — each holding all six tables, under different `table_id`s. Present since at least v2.2.1.

## What changed

- **Key the catalog cache by connection string** (`HashMap<String, OnceCell<..>>`) so each `cayenne_metadata_dir` resolves to its own catalog. This is the fix; everything else below is fallout from it.
- **`shutdown()` flushes every catalog the accelerator opened**, not just one. With a single cell there was only ever one WAL to flush.
- **`cold_tier_gc_interval_ms` is included in the runtime-config reconcile.** It was missing from the comparison, so `cayenne_datalake_gc_interval_ms` was inert on any table that already existed.

## Why

A dataset has to accelerate into the metastore its own configuration names. Anything else is silent, unbounded data loss the operator cannot see, diagnose, or attribute — the dataset registers, queries plan, and it just returns nothing.

## Testing

- Unit: two metadata dirs get two catalogs, the same dir keeps returning the same catalog, and a table created against one is not visible in the other.
- Integration: two Cayenne datasets with distinct metadata dirs are written to together, then each is reopened **alone**. Reopening alone is what makes the assertion deterministic instead of race-dependent — under the bug exactly one metastore ends up holding both tables, so whichever dataset lost cannot find its rows when it is the only one opening.
- Unit: retuned datalake intervals persist across a reopen (covers the `gc_interval` gap).

Both the unit and integration tests fail on the unfixed code.


lukekim (approved):

peasee:
Closing as we will move `acceleration.params.cayenne_metadata_dir` to `runtime.params.cayenne_metadata_dir` and make it a singular global instead.

Issue #13956, opened 2026-09-08
PostgreSQL binary CDC fails to decode JSONB values


phillipleblanc:
## Description
PostgreSQL change data capture fails when a streamed row contains JSONB. Initial snapshot reads succeed, but a subsequent INSERT detaches the CDC stream with `binary decoding into a text column is not supported for Postgres type OID 3802`.

## Reproduction
Using PostgreSQL 16, create a table with a primary key and JSONB column, enable logical replication, and configure a PostgreSQL dataset with Cayenne acceleration and `refresh_mode: changes`. Wait for the initial snapshot, then INSERT and UPDATE JSONB values containing nested objects, Unicode and JSON null. The binary pgoutput decoder rejects OID 3802 and subsequent changes do not arrive. This was reproduced with real PostgreSQL and the runtime; a local version-byte decoder candidate passes the same live INSERT/UPDATE and restart checks.

## Expected behavior
JSONB values retain their PostgreSQL representation through snapshot and CDC without detaching replication.

## Acceptance criteria

- [ ] An automated integration test against real PostgreSQL 16 fails on the unfixed runtime with the JSONB CDC error and passes on the fix. Execute the test; compilation alone does not satisfy this gate.
- [ ] After the initial snapshot, INSERT, UPDATE and DELETE are observed correctly for rows containing JSONB objects, arrays, scalars, JSON null, SQL NULL and non-ASCII text. Compare decoded values with PostgreSQL; verify snapshot/CDC parity and existing JSON/text behavior.
- [ ] Decoder tests cover supported JSONB version 1, an unsupported version, an empty/truncated payload and invalid UTF-8. Invalid input returns a diagnostic error without a panic or silent corruption.
- [ ] A valid JSONB change followed by an unrelated row change both arrive without stream detachment. After a clean runtime restart, existing values match and new changes continue to arrive within a bounded test timeout.
- [ ] Link the fixing PR and executed regression results, identifying runtime/dependency revisions and PostgreSQL version. Run applicable component tests and lint.

PR #13962, opened 2026-09-08
fix(postgres): decode versioned JSONB binary replication values

phillipleblanc:
## WHAT

Allow PostgreSQL change streams containing JSONB to continue after the initial snapshot. Refs [#13956](https://github.com/spiceai/spiceai/issues/13956).

## WHY

The binary replication decoder rejects PostgreSQL's version-prefixed JSONB values.

## HOW

Decode version 1 as UTF-8 text; reject missing/unsupported versions, empty payloads, and invalid UTF-8. Add snapshot/WAL parity coverage for JSONB, JSON, and text through INSERT/UPDATE/DELETE and subsequent delivery. The PostgreSQL 16.15 regression fails before the fix and passes afterward; 13 selected OSS unit checks, production/test Clippy, and formatting pass. Full signoff and Attestation pass, including 12,963 tests (one recovered retry) and CLI verification.

### DIFF
--- a/crates/data_components/src/postgres_replication/changes.rs
+++ b/crates/data_components/src/postgres_replication/changes.rs
@@ -1712,6 +1712,31 @@ fn decode_binary_text(raw: &[u8], type_oid: u32) -> Result<Cow<'_, str>> {
         25 | 1043 | 1042 | 19 | 114 | 142 => {
             Ok(Cow::Borrowed(pg::text_from_sql(raw).map_err(decode_err)?))
         }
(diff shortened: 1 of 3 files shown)


lukekim (approved):

phillipleblanc:
Removed from the merge queue: the `huggingface model (darwin-aarch64)` check in E2E Test CI failed twice in a row with the same "Timeout waiting for expected response" error from a local-model chat test. This doesn't correlate with this PR's diff (Postgres replication only) and looks like shared-runner resource contention rather than something introduced here. Please re-queue; it may pass cleanly when runner load is lower.

PR #13969, opened 2026-09-08
chore: post v2.3.0 release housekeeping

bjchambers:
## WHAT
Advance trunk to `2.4.0-unstable` and run scheduled benchmarks against `release/2.3`.

## WHY
Spice v2.3.0 is published, so trunk needs the next development version and release benchmark target.

## HOW
Update the workspace, CLI/API and nightly chart version metadata, keeping dependency versions and edges unchanged. Merge current trunk and regenerate the workspace lockfile so `spice-substrait-compliance` also inherits `2.4.0-unstable`.

Validation: reproduced the merge-queue failure with `bash scripts/signoff preflight-lockfile` (exit 72), then verified the same command and `cargo metadata --locked --format-version 1` pass without changing the lockfile. The lock graph matches trunk after normalizing only the workspace version rename. Phillip explicitly waived local signoff; normal merge-queue checks remain required.


SECURITY.md
```diff
@@ -2,20 +2,21 @@
 
 ## Supported Versions
 
-Spice.ai has released 2.2.0 🎉
+Spice.ai has released 2.3.0 🎉
```
lukekim: This should be done post release.

phillipleblanc (dismissed):

phillipleblanc (approved):

PR #13988, opened 2026-09-09
fix(arrow_tools): Round Decimal→Float casts from exact digits

### LINKED ISSUE #13978: avg() over a PostgreSQL column returns a wrong value: an undeclared NUMERIC read at scale 20 has a Decimal128 coefficient that no longer fits f64 (47.5 → 47.50000000000001)
## Summary

`avg()` over a PostgreSQL column returns a **wrong value**. PostgreSQL computes the average
exactly; Spice then reads it at a fixed scale of 20, and the `Decimal128` coefficient that
produces no longer fits an `f64` exactly, so the cast to `Float64` lands one ulp off.

```
true answer                : 47.5          (sum 95 / count 2, over a BIGINT column)
what PostgreSQL computes   : 47.5000000000000000
what Spice returns          : 47.50000000000001
```

Reproduced live on a two-row table (below). Deterministic on every trunk run of the three
TPC-DS arms where PostgreSQL evaluates the
aggregate, for **17 consecutive sweeps** since 2026-08-21. `tpcds_q26`, `q27` and `q83` have
been red the whole time; the checked-in snapshots hold the *correct* value.

Root cause: `NUMERIC_UNDECLARED_SCALE = 20`, introduced in
spiceai/datafusion-table-providers#53 and pinned into trunk by #13349.

## The wrong rows

Run [34223994515](https://github.com/spiceai/spiceai/actions/runs/34223994515)
(`tpcds - accelerated/file[parquet]-postgres.yaml`, 2026-09-08), `tpcds_q26`:

```
    Query: tpcds_q26
   10       │-| AAAAAAAAAADBAAAA | 47.5              | 110.755000 | 377.310000  | 79.025000 |
         10 │+| AAAAAAAAAADBAAAA | 47.50000000000001 | 110.755000 | 377.310000  | 79.025000 |
```

`agg1` is `avg(cs_quantity)`. In the TPC-DS SF1 fixture that group has exactly two rows
summing to 95, and `cs_quantity` is an integer column:

```
$ duckdb tpcds.duckdb
select data_type from information_schema.columns
 where table_name='catalog_sales' and column_name='cs_quantity';
-- BIGINT

-- q26's group for this key:
i_item_id AAAAAAAAAADBAAAA   n=2   sum=95   avg=47.5
```

95/2 = 47.5 is exactly representable in `f64`, so no accumulation order, rounding mode or scale
choice can legitimately produce `47.50000000000001`. This is a wrong result, not a stale
snapshot.

## Mechanism

**1. The aggregate is pushed to PostgreSQL.** From the checked-in explain snapshot for the arm:

```
base_sql=SELECT "item"."i_item_id", avg("catalog_sales"."cs_quantity") AS "agg1", ...
```

**2. `avg(bigint)` returns an *undeclared* NUMERIC.** Against PostgreSQL 16.15:

```
$ psql -c "create table r as select avg(q) as agg1 from (values (47::bigint),(48::bigint)) v(q);"
$ psql -c "select format_type(atttypid, atttypmod), atttypmod from pg_attribute ... where attname='agg1';"
 numeric | -1          <- typmod -1: no declared precision or scale
$ psql -c "select agg1::text, scale(agg1) from r;"
 47.5000000000000000 | 16
```

So this is precisely the "undeclared NUMERIC" path #53 changed, and PostgreSQL's own value is
exact at scale 16.

**3. Spice now reads it at scale 20.** `core/src/sql/arrow_sql_gen/postgres.rs`:

```rust
const NUMERIC_UNDECLARED_PRECISION: u8 = 38;
const NUMERIC_UNDECLARED_SCALE: i8 = 20;
```

**4. The scale-20 coefficient no longer fits `f64`.** The cast is
`arrow-cast/src/cast/mod.rs:2320`:

```rust
Float64 => cast_decimal_to_float::<D, Float64Type, _>(array, |x| {
    as_float(x) / 10_f64.powi(*scale as i32)
}),
```

`as_float` widens the `i128` coefficient to `f64` *before* dividing, so the coefficient itself
must be exactly representable:

| path | scale | coefficient | odd mantissa vs 2^53 | `f64` result |
|---|---:|---:|---|---|
| before #53 (value's own scale) | 16 | 475000000000000000 | 14,495,849,609,375 **<** 9,007,199,254,740,992 | `47.5` |
| after #53 (fixed scale) | 20 | 4750000000000000000000 | 9,059,906,005,859,375 **>** 9,007,199,254,740,992 | **`47.50000000000001`** |

The mantissa exceeds 2^53 by 0.6%. Scales 16, 17, 18 and 19 all convert exactly; 20 is the
first that does not — and 20 is the value that shipped:

```
scale 16: exact=True   ->  47.5
scale 17: exact=True   ->  47.5
scale 18: exact=True   ->  47.5
scale 19: exact=True   ->  47.5
scale 20: exact=False  ->  47.50000000000001   <- bit-for-bit what CI reports
```

## Live reproduction

Against the nightly image `ghcr.io/spiceai/spiceai-nightly:latest`
(`v2.3.0-unstable.nightly.20260908.g870e2aa`, which contains #13349) and a local
PostgreSQL 16.15, federating a two-row table:

```sql
create table t(q bigint);  insert into t values (47),(48);   -- avg is exactly 47.5
create table t2(q bigint); insert into t2 values (1),(3);    -- avg is exactly 2
```

```
$ curl -s -X POST /v1/sql --data-binary 'SELECT avg(q) AS a FROM t'
[{"a":47.50000000000001}]                     <- wrong; PostgreSQL returned 47.5000000000000000

$ curl -s -X POST /v1/sql --data-binary 'SELECT avg(q) AS a FROM t2'
[{"a":2.0}]                                   <- correct; PostgreSQL returned 2.0000000000000000
```

`EXPLAIN` shows the aggregate pushed down and a `SchemaCastScanExec` above it, as in the
benchmark arm.

**`t2` is the control that isolates the mechanism.** Both queries take the identical path — same
connector, same undeclared NUMERIC, same scale 20, same cast. The only difference is whether the
scale-20 coefficient is exactly representable in `f64`:

| query | avg | scale-20 coefficient | odd mantissa vs 2^53 | spiced |
|---|---|---:|---|---|
| `avg(q) FROM t` | 47.5 | 4750000000000000000000 | 9,059,906,005,859,375 **>** 9,007,199,254,740,992 | `47.50000000000001` |
| `avg(q) FROM t2` | 2 | 200000000000000000000 | 95,367,431,640,625 **<** 9,007,199,254,740,992 | `2.0` |

So this is not "PostgreSQL is imprecise" or "float aggregation is fuzzy" — it is exactly the
`i128 -> f64` widening inside the cast, and it is wrong for whichever values happen to land
above the threshold.

## Blast radius

At scale 20 the coefficient leaves the exactly-representable `f64` range for any value above
about `9.0e-05`. So this is not a benchmark artifact: **any** undeclared-NUMERIC value of
ordinary magnitude read through the PostgreSQL connector and cast to `Float64` can come back
wrong — `avg()`, `sum()/count()`, and any NUMERIC division, which is where undeclared NUMERICs
mostly come from.

It requires the undeclared path: a column the projected schema declares keeps its declared
scale. And it is only observable once the value is cast to `Float64` — held as
`Decimal128(38, 20)` it is exact. The benchmark casts because the parquet side's
`avg(BIGINT)` is `Float64`, which is also what a user gets from any float-typed consumer.

## Attribution

All three arms move together, on the same day, with no other candidate in the window.

| arm | last green trunk run | first red trunk run |
|---|---|---|
| `accelerated/file[parquet]-postgres` | 2026-08-20T11:12 [32362656851](https://github.com/spiceai/spiceai/actions/runs/32362656851) | 2026-08-21T11:20 [32476777575](https://github.com/spiceai/spiceai/actions/runs/32476777575) |
| `accelerated/s3[parquet]-postgres` | 2026-08-20T11:14 [32362878015](https://github.com/spiceai/spiceai/actions/runs/32362878015) | 2026-08-21T11:24 [32477086909](https://github.com/spiceai/spiceai/actions/runs/32477086909) |
| `federated/postgres` | 2026-08-20T11:27 [32363866687](https://github.com/spiceai/spiceai/actions/runs/32363866687) | 2026-08-21T11:38 [32478118718](https://github.com/spiceai/spiceai/actions/runs/32478118718) |

Branch membership resolved with `git merge-base --is-ancestor <headSha> origin/trunk`, so
`release/2.2` cherry-picks are not counted as trunk runs. The 17-commit window
`c206b7d9..d57f5c93` contains exactly one commit touching PostgreSQL NUMERIC handling:

```
ac02716758 chore(deps): bump datafusion-table-providers for NUMERIC scale fidelity (#13349)
    datafusion-table-providers  318492b8 -> f2dad232
    f2dad23 fix(postgres): give an undeclared NUMERIC a fixed scale, not the first row's (#53)
```

Unlike a timing-only attribution, the mechanism confirms it: the pre-#53 scale (16) converts
exactly and the post-#53 scale (20) does not, and the post-#53 arithmetic reproduces the
observed value bit-for-bit.

## Note on #53 itself

#53's reasoning is sound and worth keeping — taking the scale from the first sampled row pinned
the column to whatever that row held (`LIMIT 1` with no `ORDER BY`), silently rescaling every
other value, and a NULL sample truncated every fraction. It also added `NumericScaleTooWide` to
refuse rather than round. The gap is the interaction with the `Decimal128 -> Float64` cast: a
scale wide enough to carry 20 decimal places exactly is also wide enough to push the
coefficient out of `f64`, so fidelity in the decimal domain became infidelity in the float one.

## Possible directions

1. **Fix the cast, not the scale.** Divide the coefficient without widening it to `f64` first
   (split integer and fractional parts, or convert via the decimal's digits). This keeps #53's
   fidelity and removes the loss for every scale, upstream of this choice.
2. **Choose a scale that stays `f64`-exact for realistic magnitudes.** Any fixed scale trades
   range against fidelity; 16 was exact here but is only "the value's own scale" by luck.
3. Carry the value as `Decimal128` end to end where the consumer can take it, so no cast
   happens.

(1) looks like the real fix, since the same loss applies to any `Decimal128` scale wide enough
— this is not specific to PostgreSQL, only surfaced by it.

## Relationship to #13919

#13919 is a different problem despite the family resemblance: it is about a *remote engine's*
float accumulation disagreeing with local execution, over float columns, non-deterministically,
and asks for a design decision. This one is an integer column where the exact answer is
representable, PostgreSQL computes it correctly, and Spice damages it on the way in — a
regression with a date, a commit and a mechanism. Fixing this does not address #13919, and
deciding #13919 does not fix this.

## Limitations

- The live reproduction above runs a nightly build, so it confirms the *current* behaviour and
  the mechanism, not the attribution. The attribution rests on the CI boundary (all three arms
  flipping on 2026-08-21) plus the fact that the pre-#53 scale of 16 converts exactly and the
  post-#53 scale of 20 does not. I did not build `spiced` at either side of #13349.
- I have not checked whether `q27`/`q83` diverge through the same column type or a different
  one, only that they move with `q26`.
- There was a separate, one-day `tpcds_q26` episode on 2026-07-12 (on `file[parquet]-arrow` and
  two cayenne arms) that resolved on its own. That is not this.

Jeadie:
## Problem

An undeclared `PostgreSQL` NUMERIC is read as `Decimal128(38, 20)`. When Spice
casts that column to `Float64` (for example, the result of `avg()`/`sum()`
pushed down to `PostgreSQL`), the value can come back wrong.

```
true answer  : 47.5          (sum 95 / count 2, over a BIGINT column)
before        : 47.50000000000001
after         : 47.5
```

## Cause

`try_cast_to` routed `Decimal → Float` through arrow's cast kernel, which widens
the integer coefficient to `f64` before dividing by `10^scale`:

```
as_float(coefficient) / 10f64.powi(scale)
```

At scale 20 the coefficient always exceeds `2^53`, the exactly-representable
integer range of `f64`. The coefficient is therefore rounded, and the division propagates the error

```
scale 16..19 : coefficient <= 2^53  -> exact   -> 47.5
scale 20     : coefficient  > 2^53  -> rounded -> 47.50000000000001
```

## Fix

`arrow_tools::try_cast_to` now converts `Decimal128`/`Decimal256` → `Float32`/`Float64` by
rounding from the value's exact decimal digits, upstream of the loss:

- Fast path: when the coefficient fits `2^53` and `scale <= 22`, both
  operands are exact, so a single IEEE division is already correctly rounded.
- Otherwise: render the exact decimal literal and let Rust's
  correctly-rounded float parse round it once.

Values held as `Decimal128` are unchanged; only the float cast is corrected. The
change applies to every caller of `try_cast_to`, not only `PostgreSQL`.

## Evidence

- A standalone fuzz harness compared the exact conversion against a
  high-precision correctly-rounded oracle over **50,000,000** random cases
  (full `i128` range, every scale 0–38; ~10.8M of them are values the old
  widen-then-divide code got wrong): **zero** mismatches for the fix, on both
  the fast path and the string path.
- The oracle was validated against the *provably* correct fast path over 5M
  cases, and `10f64.powi(k)` is exact for every `k` in `0..=22`.
- Regression tests in `crates/arrow_tools/src/record_batch.rs` assert the issue
  value (47.5), the control value (2.0), nulls, negatives, sub-one magnitudes,
  `Decimal256`, and `Float32`, and assert that the defective arithmetic differs
  from 47.5.

Closes #13978

### DIFF
--- a/crates/arrow_tools/src/record_batch.rs
+++ b/crates/arrow_tools/src/record_batch.rs
@@ -245,7 +245,7 @@ fn is_overflow_error(e: &ArrowError) -> bool {
 /// exact digits pay for them.
 ///
 /// This intercepts `try_cast_to` only. A correctly-rounded fix in arrow's cast
-/// kernel would also cover casts DataFusion performs directly (e.g. an explicit
+/// kernel would also cover casts `DataFusion` performs directly (e.g. an explicit
 /// `CAST(x AS DOUBLE)`); that is tracked as follow-up work.
 fn cast_decimal_to_float(
     column: &ArrayRef,
--- a/crates/arrow_tools/src/record_batch.rs
+++ b/crates/arrow_tools/src/record_batch.rs
@@ -251,7 +251,7 @@ fn is_overflow_error(e: &ArrowError) -> bool {
 ///
 /// This intercepts `try_cast_to` only. A correctly-rounded fix in arrow's cast
 /// kernel would also cover casts `DataFusion` performs directly (e.g. an explicit
-/// `CAST(x AS DOUBLE)`); that is tracked as follow-up work.
+/// `CAST(x AS DOUBLE)`); that is tracked in #13993.
 fn cast_decimal_to_float(
     column: &ArrayRef,
     target_type: &DataType,
--- a/crates/arrow_tools/src/record_batch.rs
+++ b/crates/arrow_tools/src/record_batch.rs
@@ -16,16 +16,17 @@ limitations under the License.
 
 use arrow::{
     array::{
-        Array, ArrayData, ArrayRef, BinaryViewArray, DictionaryArray, GenericByteViewArray,
-        ListArray, MutableArrayData, PrimitiveArray, RecordBatch, RecordBatchOptions,
-        StringViewArray, StructArray, UInt64Array, downcast_dictionary_array, make_array,
-        new_null_array,
+        Array, ArrayData, ArrayRef, AsArray, BinaryViewArray, DictionaryArray,
+        GenericByteViewArray, ListArray, MutableArrayData, PrimitiveArray, RecordBatch,
+        RecordBatchOptions, StringViewArray, StructArray, UInt64Array, downcast_dictionary_array,
+        make_array, new_null_array,
     },
     buffer::{Buffer, NullBuffer, OffsetBuffer},
     compute::take,
     datatypes::{
         ArrowDictionaryKeyType, ArrowNativeType, ArrowNativeTypeOp, BinaryViewType, ByteViewType,
-        DataType, Field, FieldRef, SchemaRef, StringViewType, TimeUnit,
+        DataType, Decimal128Type, Field, FieldRef, Float64Type, SchemaRef, StringViewType,
+        TimeUnit,
     },
     error::ArrowError,
 };
@@ -236,6 +237,16 @@ fn is_overflow_error(e: &ArrowError) -> bool {
 /// Rust's `str::parse`, which arrow's parser matches). This covers every decimal
 /// width — `Decimal32`/`Decimal64` are lossy on the direct path too — and every
 /// float target, so it applies to every caller of `try_cast_to`.
+///
+/// The `Utf8` round-trip costs ~20-40x a plain divide, so the common
+/// `Decimal128 -> Float64` case takes a fast path when *every* coefficient is
+/// already exactly representable: a single divide is then correctly rounded on
+/// its own (see [`decimal128_to_f64_exact`]). Only values that actually need the
+/// exact digits pay for them.
+///
+/// This intercepts `try_cast_to` only. A correctly-rounded fix in arrow's cast
+/// kernel would also cover casts DataFusion performs directly (e.g. an explicit
+/// `CAST(x AS DOUBLE)`); that is tracked as follow-up work.
 fn cast_decimal_to_float(
     column: &ArrayRef,
     target_type: &DataType,
@@ -252,11 +263,45 @@ fn cast_decimal_to_float(
         return Ok(None);
     }
 
+    if let (DataType::Decimal128(_, scale), DataType::Float64) = (column.data_type(), target_type)
+        && let Some(exact) = decimal128_to_f64_exact(column, *scale)
+    {
+        return Ok(Some(exact));
+    }
+
     let strings = cast_with_options(column.as_ref(), &DataType::Utf8, options)
         .context(UnableToConvertRecordBatchSnafu)?;
     cast_with_options(strings.as_ref(), target_type, options)
         .context(UnableToConvertRecordBatchSnafu)
         .map(Some)
+}
+
+/// Fast, exact `Decimal128 -> Float64` for the common case, or `None` to fall back
+/// to the digit-exact `Utf8` path.
+///
+/// When `scale <= 22` and every coefficient is within `±2^53`, both the
+/// coefficient and `10^scale` are exactly representable, so a single IEEE division
+/// is already correctly rounded — no `Utf8` round-trip needed. The `min`/`max`
+/// gate makes this all-or-nothing per batch: if any value is out of range the
+/// whole batch takes the correct fallback, so no value is ever silently rounded
+/// on the way in.
+fn decimal128_to_f64_exact(column: &ArrayRef, scale: i8) -> Option<ArrayRef> {
+    if !(0..=22).contains(&scale) {
+        return None;
+    }
+    let values = column.as_primitive::<Decimal128Type>();
+    let within = |bound: Option<i128>| bound.is_none_or(|v| v.unsigned_abs() < (1u128 << 53));
+    if !within(arrow::compute::min(values)) || !within(arrow::compute::max(values)) {
+        return None;
+    }
+
+    let pow = 10f64.powi(i32::from(scale));
+    #[expect(
+        clippy::cast_precision_loss,
+        reason = "gated above: every coefficient is < 2^53, so i128 -> f64 is exact"
+    )]
+    let floats = values.unary::<_, Float64Type>(|v| v as f64 / pow);
+    Some(Arc::new(floats))
 }
 
 /// Whether `target` can be reached from `source` by relabelling alone: the two describe the same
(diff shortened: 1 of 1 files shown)


lukekim (approved):

Jeadie:
Closing in favour of #14012

Issue #13976, opened 2026-09-08
`refresh_mode: caching` budgets are silently disabled when the acceleration declares a `primary_key`


bjchambers:
# `refresh_mode: caching` budgets are silently disabled when the acceleration declares a `primary_key`

## Summary

On a `refresh_mode: caching` dataset, adding `acceleration.primary_key` stops
`caching_max_items` and `caching_max_size` from being enforced. The eviction
sweep fails on every pass and the acceleration grows without bound. Nothing
fails loudly: queries keep succeeding, the parameters are accepted, and the only
signal is a repeating `WARN`.

## Impact

An HTTP-connector dataset in caching mode with `caching_stale_if_error: enabled`
has nothing else bounding it — expired entries are deliberately kept as fallback
for a failing origin, so the byte and item budgets are the only limit. With a
primary key declared, that limit does not apply and the accelerator grows until
the volume fills.

Measured on a one-hour run at 100 qps over 2.5M distinct keys, budgets
`caching_max_items: 100000` / `caching_max_size: 256MiB`:

| elapsed | entries | × item budget | payload | × byte budget |
|---:|---:|---:|---:|---:|
| 360s | 35,272 | 0.35 | 106.9 MiB | 0.42 |
| 1080s | 98,744 | 0.99 | 299.2 MiB | 1.17 |
| 2163s | 176,977 | 1.77 | 536.3 MiB | 2.10 |
| 3573s | **298,698** | **2.99** | **905.0 MiB** | **3.54** |

Linear for the full hour, zero evictions, 320,441 queries with no errors.

## Reproduction

The script below needs only Docker and curl; it generates the origin and both
spicepods. Roughly three minutes.

```
  without  primary_key: entries 3 -> 1   (budget 1)   sweep failures: 0
  with     primary_key: entries 3 -> 3   (budget 1)   sweep failures: 3
```

The two runtimes differ by one line:

```yaml
     acceleration:
       refresh_mode: caching
+      primary_key: '(request_query,request_path)'
       params:
         caching_ttl: 30s
         caching_stale_while_revalidate_ttl: 1h
         caching_stale_if_error: enabled
         caching_max_items: 1
```

<details>
<summary>caching-bounds-primary-key.sh</summary>

```bash
#!/usr/bin/env bash
#
# Minimal reproduction: declaring an acceleration `primary_key` on a
# `refresh_mode: caching` dataset silently disables `caching_max_items` and
# `caching_max_size`.
#
[...]

PR #13990, opened 2026-09-09
fix: Keep the caching-eviction sweep planning when a primary_key is declared

Jeadie:
## Problem

A `refresh_mode: caching` accelerator stops enforcing `caching_max_items` and
`caching_max_size` when the dataset declares a `primary_key`. The cache then
grows without bound. Queries keep succeeding, so the only signal is a repeating
warning in the log:

```
[retention] Failed to resolve what to delete for dataset <name>, so it keeps
whatever is past its retention until the next check. Cause: ...
```

See #13976.

## Observed behaviour

The eviction sweep in `caching_eviction.rs` derives what to evict with one
aggregate per tick: `read_table(accelerator).aggregate(...)`, grouped by the
cache key columns. `read_table` scans the accelerator's full schema, which for
an HTTP cache includes `response_headers` — an Arrow `Map` the connector stores
on every row.

Reproduced by the added test: with the `response_headers` Map present and a
`primary_key` declared on the key column, the sweep aggregate fails with
`Row format support not yet implemented for: [...Map...]`, the retention loop
logs the failure and moves on, and nothing is evicted. Remove the `primary_key`
and the same sweep runs and evicts. So the constraint is what triggers it.

Working explanation (verified black-box, not traced through the optimiser): a
declared `primary_key` adds functional dependencies over the scanned relation.
Without a key, projection pushdown prunes the unread `Map` before execution;
with one, the constraint keeps the `Map` in the plan, where DataFusion's row
format has no encoding for it. Whichever optimiser path retains the column, it
acts on columns present in the aggregate's input relation.

## Fix

Project the sweep to the columns it actually reads before aggregating. The kept
set is exactly the columns `measured_bytes` can weigh — the utf8 keys,
`_fetched_at`, and the payload columns a byte budget sums. Every nested column a
row format cannot encode, including the `Map`, is dropped before it reaches the
plan, so the aggregate never sees it regardless of which optimiser path would
otherwise retain it. The sweep then plans identically with or without a declared
key.

The change is one projection at the single sweep call site. Byte accounting is
unchanged: the payload columns a budget sums are all row-format-safe and stay in
the projection.

## Test plan

- [x] Regression test `the_item_budget_is_enforced_when_a_primary_key_is_declared`:
  a caching accelerator with a `response_headers` Map and a `primary_key` on the
  key column, asserting the sweep still evicts under an item budget. Verified it
  passes with the fix and fails without it (the pre-fix run reproduces the
  `Row format ... Map` sweep failure).
- [x] `spiced_docker_dev.yml` dev image build — passed ([run 34325773237](https://github.com/spiceai/spiceai/actions/runs/34325773237)); image `ghcr.io/spiceai/spiceai-dev:pr-13990-caching-pk` (amd64).

### DIFF
--- a/crates/runtime-table/src/accelerated/caching_eviction.rs
+++ b/crates/runtime-table/src/accelerated/caching_eviction.rs
@@ -729,8 +729,27 @@ async fn rank_entries(
         get_df_default_config(),
         default_runtime_env(io_runtime.clone()),
     );
+    // Project away nested columns before aggregating. `read_table` scans the
+    // accelerator's whole schema, which for an HTTP cache includes the
+    // `response_headers` Map stored on every row. The sweep never reads that
+    // Map, but a declared `primary_key` adds functional dependencies that keep
+    // it live into the aggregate, where DataFusion's row format has no encoding
+    // for a Map — so the sweep fails on every tick and the budget is silently
+    // never enforced (#13976). Keeping only the columns `measured_bytes` can
+    // weigh retains exactly what the aggregate reads — the utf8 keys,
+    // `_fetched_at`, and the payload columns a byte budget sums — and drops
+    // every nested column a row format cannot encode, so the sweep plans with
+    // or without a declared key.
+    let projection: Vec<Expr> = schema
+        .fields()
+        .iter()
+        .filter(|field| measured_bytes(field).is_some())
+        .map(|field| col(field.name()))
+        .collect();
+
     let mut df = ctx
         .read_table(Arc::clone(accelerator))?
+        .select(projection)?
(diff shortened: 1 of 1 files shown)


Jeadie:
Root cause tracked upstream in #13991: with a `primary_key`/unique constraint, DataFusion expands `GROUP BY key` to `GROUP BY` over all functionally-dependent columns, which drags the `response_headers` `Map` into the group-by `RowConverter`. This PR is the local workaround (project the `Map` out of the sweep); #13991 covers the underlying DataFusion behaviour and a possible upstream fix.

Jeadie:
Closing in favour of #13992

